PR #13710, opened 2022-02-17
scripts: Remove Windows build scripts

serathius:
Closes https://github.com/etcd-io/etcd/issues/13667

cc @spzala @ahrtr @jberkus

Issue #13667, opened 2022-02-04
Build on Windows fails


jeffrson:
### What happened?

Etcd and Etcdctl are extremely slow on Windows (~5 Seconds response time). This may be subject to another issue.

Now I tried to build Etcd myself, following documentation. After clone I run .\build.bat, because this way it's using Powershell (otherwise Powershell Core would be used, where the script is broken: https://github.com/etcd-io/etcd/issues/12293)

However, no build will be executed. Instead the following message is shown:
```
>powershell -ExecutionPolicy Bypass -File build.ps1
no required module provides package go.etcd.io/etcd; to add it:
        go get go.etcd.io/etcd
no required module provides package go.etcd.io/etcd/etcdctl; to add it:
        go get go.etcd.io/etcd/etcdctl
```

Now, when I do this (which is useless as "build") it fails as well:
```
>go get go.etcd.io/etcd
go: downloading go.etcd.io/etcd v3.3.27+incompatible
go: downloading github.com/coreos/etcd v3.3.27+incompatible
...
go.etcd.io/etcd imports
        github.com/coreos/etcd/etcdmain imports
        github.com/coreos/etcd/proxy/grpcproxy imports
        google.golang.org/grpc/naming: cannot find module providing package google.golang.org/grpc/naming
```

### What did you expect to happen?

Build fresh etcd/etcdctl as on Linux

### How can we reproduce it (as minimally and precisely as possible)?

try
`.\build.bat`

and
`go get go.etcd.io/etcd`

### Etcd version (please run commands below)

n/a

### Etcd configuration (command line flags or environment variables)

n/a

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

n/a

ahrtr:
You can run command `GOOS=windows GOARCH=amd64 make build` on Linux or darwin to generate Windows binaries. 

I see that the [build.bat](https://github.com/etcd-io/etcd/blob/main/build.bat) haven't been updated for 6 years, and [build.ps1](https://github.com/etcd-io/etcd/blob/main/build.ps1) haven't been updated for about 4 years. Do we need to continue to maintain these build scripts for Windows which is only on tier 3 support? cc @serathius  @ptabor @spzala

serathius:
Overall I'm against maintaining anything that we cannot automatically test and we don't have dedicated maintainer that can fix it when something breaks. Keeping this in this state breaks promise we make for users in our documentation and leads to lowering standards in other places.

[...]

PR #13952, opened 2022-04-15
[DRAFT] add powershell build scripts for native Windows development of etcd

rosskirkpat:
This PR adds new Windows build scripts as the previous script was [removed](https://github.com/etcd-io/etcd/pull/13710) due to no longer working and not being actively maintained.

I have also added a step to the `build` GH action that leverages the new scripts to build the etcd/etcdctl binaries on a Windows runner.

Status:  
- etcd and etcdctl build with the new scripts.
- I am currently working through issues with the etcd binary `version.GitSha` not being correctly set during build time even with the same linker flag values and process as linux. This appears to be a more complex problem than I anticipated and that's delayed this PR. It appears to be due to the same underlying issue that caused the original script to break. 

I wanted to get this PR up as a draft in the interim until I can figure out how to fix the linker flag.

@serathius I believe this PR will meet the requirements you stated here https://github.com/etcd-io/etcd/issues/13667#issuecomment-1044177247 once I get the linker flag item resolved.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,6 @@
 name: Build
 on: [push, pull_request]
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yaml
```diff
@@ -58,3 +59,33 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    steps:
+    # This step is required as long as the symlinks for client/v3/*_test.go -> tests/integration/clientv3/examples/*_test.go exist
+      - name: core.symlinks
+        run: |
+          git config --global core.symlinks true
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1      
+      - uses: WillAbides/setup-go-faster@v1.7.0
```
serathius, line 82: Faster is not alwasy better. We should prefer official actions that are well maintained instead of trying a fork.
Please use normal `setup-go` and create separate issue/PR with proposal to migrate to setup-go-faster so we can discuss it independently.

.github/workflows/build.yaml
```diff
@@ -58,3 +59,33 @@ jobs:
...
+        with:
+          go-version: ${{ matrix.go }}
+      - name: Continuous Integration
+        run: make.ps1 -Script ci # -Version ${{ github.ref }}
+      - uses: dominikh/staticcheck-action@v1.2.0
```
serathius, line 87: Please don't put static check as part of build workflow.

.github/workflows/build.yaml
saintdarthcoder: This should be actions/setup-go?

jmhbnz:
Closing due to lack of activity, we can always reopen in future if someone has capacity to continue the work.

Issue #13637, opened 2022-01-24
Unify testing framework


serathius:
Test flakiness and maintenance cost remains one of the larger issues for Etcd project. Proposals like https://github.com/etcd-io/etcd/issues/13167 to track flakiness and contributions to address singular tests have definitely helped however to fully resolve the problem we also need to address the root cause. Etcd projects currently maintains multiple disconnected ways to run tests (unit, integration, e2e, functional) that mostly verify the same scenarios. Testing features on different level is itself desired as it allows to  isolate failures and speed up resolution, however not in a way that is currently done in Etcd. Almost zero test scenario and test framework is reused. This means that we have multiplied test code lines without any benefit.

Goal:
* Reduce number of test code without affecting test coverage
* Simplify process of adding new integration/e2e test
* Make it easier to isolate test failures by comparing test results from different methods

I propose to unify testing by identifying common test scenarios, test framework operations and making them accessable the same way no matter which test method is used (unit, integration, e2e, functional). For example simple put&get test scenario goes through very same stages no matter if this is integration or e2e cluster: create cluster, execute put, execute get, compare results, cleanup. This test should be written once and executed on different layers, no matter if underneath framework runs a fake grpc client or starts whole process. With this we will reduce number of code lines as each test will be needed to be implemented once, hide all the differences between test types reducing knowledge needed to add new tests and improve deflaking process by comparing same test results from different methods (we can isolate test location, if both e2e and integration tests fail then feature is broken, if only one of them then the test framework)

At this moment I would be most effective to target unifying integration and e2e tests as they are most similar in both cluster setup and test scenarios. We should be able to identify minimal common interface for cluster creation and communication that would allow us to start rewriting tests. 

Plan:
* Create new `Cluster` and `Client` interfaces in `tests/framework`.
* Find common function in `tests/framework/e2e/cluster.go` and `tests/framework/integration/cluster.go` and migrate them one by one.
* Deduplicate test scenarios between e2e and integration tests
* Create common tests located in `tests/common` that will be invoked by e2e and integration.

cc @ptabor @ahrtr @spzala

ahrtr:
It's really a good proposal. I am still very new to the existing test framework & cases, but I think the high level architecture would be something like below,

![test_framework](https://user-images.githubusercontent.com/30739825/151268188-c1d97e09-4dcc-4757-a277-d538cd2f724b.png)

serathius:
[...]

PR #13953, opened 2022-04-17
tests: Migrate watch test to common framework

nic-chen:
Part of https://github.com/etcd-io/etcd/issues/13637


tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +507,51 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+	if err != nil {
+		return nil
+	}
+	err = json.Unmarshal([]byte(line), &resp)
+	if err != nil {
+		return nil
+	}
+	ch := make(chan clientv3.WatchResponse, 1)
+	ch <- resp
+	close(ch)
```
serathius, line 554: I'm worried that watch in `e2e` tests (return first watch response and close) will have different semantic meaning than in `integration` tests (provide all updates until close). This might lead to hard to debug problems when more tests are migrated.

Can you look if it is possible to make both work the same?

tests/common/watch_test.go
```diff
@@ -0,0 +1,116 @@
...
+					{ // watch 3 keys by range
+						puts:     []testutils.KV{{Key: "key1", Val: "val1"}, {Key: "key3", Val: "val3"}, {Key: "key2", Val: "val2"}},
+						watchKey: "key",
+						opts:     config.WatchOptions{Revision: 1, RangeEnd: "key3"},
+						wanted:   []testutils.KV{{Key: "key1", Val: "val1"}, {Key: "key2", Val: "val2"}},
+					},
+				}
+
+				for i, tt := range tests {
+					donec := make(chan struct{})
```
serathius, line 82: please use sync.WaitGroup to wait for goroutine to finish.

tests/common/watch_test.go
```diff
@@ -0,0 +1,116 @@
...
+					errs := make(chan error, 1)
+					go func(i int, puts []testutils.KV) {
+						for j := range puts {
+							if err := cc.Put(puts[j].Key, puts[j].Val, config.PutOptions{}); err != nil {
+								errs <- fmt.Errorf("can't not put key %q, err: %s", puts[j].Key, err)
```
serathius, line 87: I think it might be easier to just call `t.Fatal` to convey fail.

tests/common/watch_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+func TestTxn(t *testing.T) {
```
serathius, line 14: The effort is about migrating the test, replacing old test with new one. Which test are you migrating here?
Can you just remove the original code of the test so it will be easier to review and compare.

tests/framework/e2e/etcdctl.go
nic-chen: Sure, thanks for pointing it out, I will try to do this.

tests/common/watch_test.go
nic-chen: use `t.Fatal` will cause  `call to (*T).Fatalf from a non-test goroutine`, so I changed it to this..

tests/common/watch_test.go
nic-chen: When I was migrating of this part of the test, I found that the watch test actually has a big difference between `e2e` and `integration`.

For example `--interactive` is only for the command line (if I missed something, please let me know, thank you ), but many e2e tests are related to it.

Would like to know your opinion on this issue. Thanks!

nic-chen:
Hi, @serathius, please review again at your convenience. Thanks.

tests/framework/config/client.go
```diff
@@ -59,3 +59,11 @@ type LeaseOption struct {
 type UserAddOptions struct {
 	NoPassword bool
 }
+
+type WatchOptions struct {
+	Prefix      bool
+	Revision    int64
+	RangeEnd    string
+	ExecCmd     []string
```
serathius, line 67: Please don't use ExecCmd as this breaks abstraction. Goal is to extract common tests between integration and e2e, however we should watch for tests that make sense only for one environment.

tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +508,60 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+
+	ch := make(chan clientv3.WatchResponse)
+	go func() {
+		var resp clientv3.WatchResponse
+		for {
+			lines := proc.Lines()
+			for _, line := range lines {
+				err = json.Unmarshal([]byte(line), &resp)
+				if err != nil {
+					continue
```
serathius, line 553: What lines are skipped here?

tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +508,60 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+				}
+				ch <- resp
+			}
+
+			select {
+			case <-ctx.Done():
```
serathius, line 559: Won't this wait always until `ctx.Done()`? I think we should default case in select so the external for loop makes sense, however I don't internal loop will look as `proc.Lines()` will repeat lines.

tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +508,60 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+	if len(opts.ExecCmd) > 0 {
+		args = append(args, "--")
+		args = append(args, opts.ExecCmd...)
+	}
+
+	proc, err := SpawnCmd(args, nil)
+	if err != nil {
+		return nil
+	}
+	defer proc.Stop()
```
serathius, line 537: This will kill the proc even though we started separate read goroutine on line 545

tests/common/watch_test.go
```diff
@@ -0,0 +1,121 @@
...
+				}
+
+				for i, tt := range tests {
+					var wg sync.WaitGroup
+					wg.Add(1)
+					errs := make(chan error, 1)
+					ctx, cancel := context.WithCancel(context.Background())
+					go func(i int, puts []testutils.KV) {
+						defer wg.Done()
+						defer close(errs)
```
serathius, line 89: Don't think this is needed.

tests/common/watch_test.go
```diff
@@ -0,0 +1,121 @@
...
+						defer cancel()
```
serathius, line 90: Context cancel should be called next to context creation

tests/common/watch_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+					err := <-errs
+					if err != nil {
+						t.Fatal(err)
+					}
+					wg.Wait()
+
+					wch := cc.Watch(ctx, tt.watchKey, tt.opts)
+					select {
+					case wresp, ok := <-wch:
```
serathius, line 109: Should this read until channel is closed, not just once?

tests/common/watch_test.go
```diff
@@ -0,0 +1,121 @@
...
+						if ok {
```
serathius, line 110: In case that channel is closed, should we report error if we were expecting response?

ahrtr:
Please resolve all review comments and conflicts.

nic-chen:
> Please resolve all review comments and conflicts.

Sorry, I was on a long business trip during that time and missed this PR. Will pick it up again soon. thanks.

nic-chen:
Due to my habit, I merged main to this branch before, and there is no way to use rebase to deal with the `DCO` issue, so I will close this PR and resubmit a new one.

nic-chen:
> Due to my habit, I merged main to this branch before, and there is no way to use rebase to deal with the `DCO` issue, so I will close this PR and resubmit a new one.

@serathius @ahrtr  please help review the new PR https://github.com/etcd-io/etcd/pull/14345, thanks.

PR #13954, opened 2022-04-17
server/auth: protect rangePermCache with a mutex

mitake:
This PR introduces a new mutex for protecting `authStore.rangePermCache`. The field needs an explicit synchronization because it can be accessed concurrently by the state machine apply layer and RPC layer.

cc @ahrtr @ptabor @serathius

### DIFF
--- a/server/auth/range_perm_cache.go
+++ b/server/auth/range_perm_cache.go
@@ -103,34 +103,49 @@ func checkKeyPoint(lg *zap.Logger, cachedPerms *unifiedRangePermissions, key []b
 	return false
 }
 
-func (as *authStore) isRangeOpPermitted(tx AuthReadTx, userName string, key, rangeEnd []byte, permtyp authpb.Permission_Type) bool {
+func (as *authStore) isRangeOpPermitted(userName string, key, rangeEnd []byte, permtyp authpb.Permission_Type) bool {
 	// assumption: tx is Lock()ed
-	_, ok := as.rangePermCache[userName]
+	as.rangePermCacheMu.RLock()
+	defer as.rangePermCacheMu.RUnlock()
+
+	rangePerm, ok := as.rangePermCache[userName]
 	if !ok {
+		as.lg.Error(
+			"user doesn't exist",
+			zap.String("user-name", userName),
+		)
+		return false
+	}
+
+	if len(rangeEnd) == 0 {
+		return checkKeyPoint(as.lg, rangePerm, key, permtyp)
+	}
+
+	return checkKeyInterval(as.lg, rangePerm, key, rangeEnd, permtyp)
+}
+
+func (as *authStore) refreshRangePermCache(tx AuthReadTx) {
+	// Note that every authentication configuration update calls this method and it invalidates the entire
+	// rangePermCache and reconstruct it based on information of users and roles stored in the backend.
+	// This can be a costly operation.
+	as.rangePermCacheMu.Lock()
+	defer as.rangePermCacheMu.Unlock()
+
+	as.rangePermCache = make(map[string]*unifiedRangePermissions)
+
+	users := tx.UnsafeGetAllUsers()
+	for _, user := range users {
+		userName := string(user.Name)
 		perms := getMergedPerms(tx, userName)
 		if perms == nil {
 			as.lg.Error(
 				"failed to create a merged permission",
 				zap.String("user-name", userName),
 			)
-			return false
+			continue
 		}
 		as.rangePermCache[userName] = perms
 	}
-
-	if len(rangeEnd) == 0 {
-		return checkKeyPoint(as.lg, as.rangePermCache[userName], key, permtyp)
-	}
-
-	return checkKeyInterval(as.lg, as.rangePermCache[userName], key, rangeEnd, permtyp)
-}
-
-func (as *authStore) clearCachedPerm() {
-	as.rangePermCache = make(map[string]*unifiedRangePermissions)
-}
-
-func (as *authStore) invalidateCachedPerm(userName string) {
-	delete(as.rangePermCache, userName)
 }
 
 type unifiedRangePermissions struct {
--- a/server/auth/store.go
+++ b/server/auth/store.go
@@ -235,7 +235,14 @@ type authStore struct {
 	enabled   bool
 	enabledMu sync.RWMutex
 
-	rangePermCache map[string]*unifiedRangePermissions // username -> unifiedRangePermissions
+	// rangePermCache needs to be protected by rangePermCacheMu
+	// rangePermCacheMu needs to be write locked only in initialization phase or configuration changes
+	// Hot paths like Range(), needs to acquire read lock for improving performance
+	//
+	// Note that BatchTx and ReadTx cannot be a mutex for rangePermCache because they are independent resources
+	// see also: https://github.com/etcd-io/etcd/pull/13920#discussion_r849114855
+	rangePermCache   map[string]*unifiedRangePermissions // username -> unifiedRangePermissions
+	rangePermCacheMu sync.RWMutex
 
 	tokenProvider TokenProvider
 	bcryptCost    int // the algorithm cost / strength for hashing auth passwords
@@ -268,7 +275,7 @@ func (as *authStore) AuthEnable() error {
 	as.enabled = true
 	as.tokenProvider.enable()
 
-	as.rangePermCache = make(map[string]*unifiedRangePermissions)
+	as.refreshRangePermCache(tx)
 
 	as.setRevision(tx.UnsafeReadAuthRevision())
 
@@ -437,6 +444,7 @@ func (as *authStore) UserAdd(r *pb.AuthUserAddRequest) (*pb.AuthUserAddResponse,
 	tx.UnsafePutUser(newUser)
 
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
 	as.lg.Info("added a user", zap.String("user-name", r.Name))
 	return &pb.AuthUserAddResponse{}, nil
@@ -459,8 +467,8 @@ func (as *authStore) UserDelete(r *pb.AuthUserDeleteRequest) (*pb.AuthUserDelete
 	tx.UnsafeDeleteUser(r.Name)
 
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
-	as.invalidateCachedPerm(r.Name)
 	as.tokenProvider.invalidateUser(r.Name)
 
 	as.lg.Info(
@@ -500,8 +508,8 @@ func (as *authStore) UserChangePassword(r *pb.AuthUserChangePasswordRequest) (*p
 	tx.UnsafePutUser(updatedUser)
 
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
-	as.invalidateCachedPerm(r.Name)
 	as.tokenProvider.invalidateUser(r.Name)
 
 	as.lg.Info(
@@ -545,9 +553,8 @@ func (as *authStore) UserGrantRole(r *pb.AuthUserGrantRoleRequest) (*pb.AuthUser
 
 	tx.UnsafePutUser(user)
 
-	as.invalidateCachedPerm(r.User)
-
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
 	as.lg.Info(
 		"granted a role to a user",
@@ -617,9 +624,8 @@ func (as *authStore) UserRevokeRole(r *pb.AuthUserRevokeRoleRequest) (*pb.AuthUs
 
 	tx.UnsafePutUser(updatedUser)
 
-	as.invalidateCachedPerm(r.Name)
-
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
 	as.lg.Info(
 		"revoked a role from a user",
@@ -682,11 +688,8 @@ func (as *authStore) RoleRevokePermission(r *pb.AuthRoleRevokePermissionRequest)
 
 	tx.UnsafePutRole(updatedRole)
 
-	// TODO(mitake): currently single role update invalidates every cache
-	// It should be optimized.
-	as.clearCachedPerm()
-
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
 	as.lg.Info(
 		"revoked a permission on range",
@@ -734,10 +737,10 @@ func (as *authStore) RoleDelete(r *pb.AuthRoleDeleteRequest) (*pb.AuthRoleDelete
 
 		tx.UnsafePutUser(updatedUser)
 
-		as.invalidateCachedPerm(string(user.Name))
 	}
 
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
 	as.lg.Info("deleted a role", zap.String("role-name", r.Role))
 	return &pb.AuthRoleDeleteResponse{}, nil
@@ -822,11 +825,8 @@ func (as *authStore) RoleGrantPermission(r *pb.AuthRoleGrantPermissionRequest) (
 
 	tx.UnsafePutRole(role)
 
-	// TODO(mitake): currently single role update invalidates every cache
-	// It should be optimized.
-	as.clearCachedPerm()
-
 	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 
 	as.lg.Info(
 		"granted/updated a permission to a user",
@@ -871,7 +871,7 @@ func (as *authStore) isOpPermitted(userName string, revision uint64, key, rangeE
 		return nil
 	}
 
-	if as.isRangeOpPermitted(tx, userName, key, rangeEnd, permTyp) {
+	if as.isRangeOpPermitted(userName, key, rangeEnd, permTyp) {
 		return nil
 	}
 
--- a/server/auth/store_test.go
+++ b/server/auth/store_test.go
@@ -29,6 +29,7 @@ import (
 	"go.etcd.io/etcd/api/v3/authpb"
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	"go.etcd.io/etcd/pkg/v3/adt"
 	"golang.org/x/crypto/bcrypt"
 	"google.golang.org/grpc/metadata"
 )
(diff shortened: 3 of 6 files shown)


server/auth/range_perm_cache.go
```diff
@@ -105,6 +105,9 @@ func checkKeyPoint(lg *zap.Logger, cachedPerms *unifiedRangePermissions, key []b
 
 func (as *authStore) isRangeOpPermitted(tx AuthReadTx, userName string, key, rangeEnd []byte, permtyp authpb.Permission_Type) bool {
 	// assumption: tx is Lock()ed
+	as.rangePermCacheMu.Lock()
+	defer as.rangePermCacheMu.Unlock()
```
ahrtr, line 109: Locking the whole method will definitely lead to decrease of performance. 

Firstly, I was thinking a sync.RWMutex could make more sense because usually the reading cache operations are greatly overnumber the writing cache operations. It means that we need to upgrade a read lock to a write lock when needed, but it complicates the design & implementation, and [golang](https://github.com/golang/go/issues/38859) doesn't support such case either :( .  

Secondly, [isOpPermitted](https://github.com/etcd-io/etcd/blob/81491914fbfbfcc041e175beb20f2cf3f575557e/server/auth/store.go#L839) is the only case which could be called by both the applying workflow and the API. All other cases ( I mean cases which need to access the `rangePermCache`) are coming from the applying workflow. Is it possible to resolve this from a higher level? 

An unrelated comment, the naming [isRangeOpPermitted](https://github.com/etcd-io/etcd/blob/81491914fbfbfcc041e175beb20f2cf3f575557e/server/auth/store.go#L874) isn't accurate, because it handles all operation types, including `authpb.READ` and `authpb.WRITE`.

ahrtr: Another proposal: 
Firstly change the sync.Mutex to sync.RWMutex.

Secondly, the isRangeOpPermitted only reads the cache, and the cache is only updated in the following cases:

1. etcd startup & initialization
2. recovering from snapshot
3. RoleGrant and RoleRevoke

Does it make sense?

mitake: @ahrtr I think your suggestion makes sense, it will simplify the design and implementation. Let me update this PR. Thanks!

xiaods:
need fix coverage testing.

server/auth/store.go
```diff
@@ -682,11 +687,8 @@ func (as *authStore) RoleRevokePermission(r *pb.AuthRoleRevokePermissionRequest)
 
 	tx.UnsafePutRole(updatedRole)
 
-	// TODO(mitake): currently single role update invalidates every cache
```
ptabor, line 685: I think the 'cost' mentioned by the TODO still exists... now every single role change is rebuilding the cache from the scratch... that actually might be even more costly on the 'synchronous' path of apply...

server/auth/range_perm_cache.go
```diff
@@ -103,34 +103,50 @@ func checkKeyPoint(lg *zap.Logger, cachedPerms *unifiedRangePermissions, key []b
...
+	tx := as.be.ReadTx()
+	tx.Lock()
+	defer tx.Unlock()
+
 	as.rangePermCache = make(map[string]*unifiedRangePermissions)
-}
 
-func (as *authStore) invalidateCachedPerm(userName string) {
-	delete(as.rangePermCache, userName)
+	users := as.be.GetAllUsers()
```
ptabor, line 137: Aren't we afraid that this path might be costly ? 
I don't know how many user's we have in practice... but if we have series of permission edits (e.g. 
N role Adds, we will have `O(N^2)` actions to be performed on the critical apply code-path. 

What's the reasoning for proactive cache refresh vs. invalidation and `lazy` cache population ?

mitake: > What's the reasoning for proactive cache refresh vs. invalidation and lazy cache population ?

I selected proactive cache refresh just for simplicity. I know an etcd cluster which initializes permission configuration just after initialization and doesn't udpate it often so won't be so problematic even the operation becomes costly. But as you say having the costly operation in the apply path is potentially dangerous. I'll check performance of the method.

mitake: I've been fixing a functional issue of this PR, not evaluating performance. I think I'll be able to evaluate performance tomorrow...

mitake: I'm comparing the main branch and this PR's branch with 100 users and 10 roles each on my local laptop (prepared by this script: https://gist.github.com/mitake/3438b6b7fb64d3f07f3d4282951645e0).
After preparing the dummy users and roles, just creating one user by `etcdctl user add <username>`, which involves `as.refreshRangePermCache()`, didn't show significant difference on each branch.

I measured wall clock time for the function with a change like below and it was about 200 us for each call:
`22:33:43 etcd1 | {"level":"info","ts":"2022-05-11T22:33:43.811+0900","caller":"auth/range_perm_cache.go:153","msg":"elapsed time for refreshRangePermCache: ","elapsed":"195.75µs"}`

```
diff --git a/server/auth/range_perm_cache.go b/server/auth/range_perm_cache.go
index 35fcfe120..43a85b12a 100644
--- a/server/auth/range_perm_cache.go
+++ b/server/auth/range_perm_cache.go
@@ -18,6 +18,7 @@ import (
        "go.etcd.io/etcd/api/v3/authpb"
        "go.etcd.io/etcd/pkg/v3/adt"
        "go.uber.org/zap"
+       "time"
 )

 func getMergedPerms(tx AuthReadTx, userName string) *unifiedRangePermissions {
@@ -125,6 +126,7 @@ func (as *authStore) isRangeOpPermitted(userName string, key, rangeEnd []byte, p
 }

 func (as *authStore) refreshRangePermCache(tx AuthReadTx) {
+       start := time.Now()
        // Note that every authentication configuration update calls this method and it invalidates the entire
        // rangePermCache and reconstruct it based on information of users and roles stored in the backend.
        // This can be a costly operation. TODO(mitake): check how it is costly actually.
@@ -146,6 +148,10 @@ func (as *authStore) refreshRangePermCache(tx AuthReadTx) {
                }
                as.rangePermCache[userName] = perms
        }
+
+       elapsed := time.Since(start)
+       as.lg.Info("elapsed time for refreshRangePermCache: ",
+               zap.Duration("elapsed", elapsed))
 }

 type unifiedRangePermissions struct {
```

I'll evaluate with different parameters but I guess the function isn't adding significant cost.

ptabor:
@mitake Please update the unit tests: 
```
% (cd server && 'go' 'vet' './...')
stderr: # go.etcd.io/etcd/server/v3/auth
Error: stderr: vet: auth/store_mock_test.go:63:21: cannot use (*backendMock)(nil) (value of type *backendMock) as AuthBackend value in variable declaration: wrong type for method GetAllUsers (have func() []*go.etcd.io/etcd/api/v3/authpb.User, want func(tx go.etcd.io/etcd/server/v3/auth.AuthReadTx) []*go.etcd.io/etcd/api/v3/authpb.User)
stderr: # go.etcd.io/etcd/server/v3/storage/schema
Error: stderr: vet: storage/schema/auth_users_test.go:116:30: not enough arguments in call to abe2.GetAllUsers
```

server/auth/range_perm_cache.go
mitake: I did a same test with 1000 users and 10 roles for each. The cost for refreshing perm range cache was about 5ms - 10ms: https://gist.github.com/mitake/135b9b51bf05dcd2fb020e17bded5fe8
I think it's safe to conclude the overhead is acceptable. How do you think @ptabor ?

mitake:
@ptabor Updated the broken test and enhanced some unit tests, could you take a look after CIs pass?

mitake:
@ptabor @ahrtr Could you check this PR when you have time?

ptabor (approved):
Thank you. Looks good to me. 
I'm sorry for not looking on the PR earlier. It passed out of horizont of the recent PRs.

CHANGELOG/CHANGELOG-3.4.md
```diff
@@ -10,9 +10,11 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.4.18...v3.4.19) an
 
 ### etcd server
 - Fix [exclude the same alarm type activated by multiple peers](https://github.com/etcd-io/etcd/pull/13475).
+- Fix [Defrag unsets backend options](https://github.com/etcd-io/etcd/pull/13713).
+- Fix [lease leak issue due to tokenProvider isn't enabled when restoring auth store from a snapshot](https://github.com/etcd-io/etcd/pull/13206).
```
ahrtr, line 14: These changes are not made this PR. Could you rebase this PR?

server/auth/store.go
```diff
@@ -571,7 +578,11 @@ func (as *authStore) UserGet(r *pb.AuthUserGetRequest) (*pb.AuthUserGetResponse,
 }
 
 func (as *authStore) UserList(r *pb.AuthUserListRequest) (*pb.AuthUserListResponse, error) {
-	users := as.be.GetAllUsers()
+	tx := as.be.BatchTx()
```
ahrtr, line 581: Please use` as.be.ReadTx() `instead if it's readonly operation.

server/storage/schema/auth_users.go
```diff
@@ -31,10 +32,7 @@ func (atx *authBatchTx) UnsafeGetUser(username string) *authpb.User {
 	return arx.UnsafeGetUser(username)
 }
 
-func (abe *authBackend) GetAllUsers() []*authpb.User {
-	tx := abe.BatchTx()
-	tx.Lock()
-	defer tx.Unlock()
+func (abe *authBackend) GetAllUsers(tx auth.AuthReadTx) []*authpb.User {
```
ahrtr, line 35: I would suggest to remove `GetAllUsers` and users call `UnsafeGetAllUsers` directly. All methods prefixed with `Unsafe` are unsafe operations, otherwise, they are safe operations.

server/storage/schema/auth_users_test.go
```diff
@@ -113,7 +113,7 @@ func TestGetAllUsers(t *testing.T) {
 			be2 := backend.NewDefaultBackend(lg, tmpPath)
 			defer be2.Close()
 			abe2 := NewAuthBackend(lg, be2)
-			users := abe2.GetAllUsers()
+			users := abe2.GetAllUsers(abe2.BatchTx())
```
ahrtr, line 116: abe2.ReadTx()

server/auth/store_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"go.etcd.io/etcd/pkg/v3/adt"
```
ahrtr, line 21: Please move this import item into the following group starting from line 30.

server/auth/store_test.go
```diff
@@ -155,6 +157,13 @@ func TestUserAdd(t *testing.T) {
 	if err != ErrUserEmpty {
 		t.Fatal(err)
 	}
+
+	as.rangePermCacheMu.RLock()
+	defer as.rangePermCacheMu.RUnlock()
```
ahrtr, line 162: I don't understand why do we need to acquire the lock in unit test. Is there any potential race condition here? Of course, it isn't a big deal to keep it as it is. 

The same comment to some following test cases.

ahrtr:
> @ptabor @ahrtr Could you check this PR when you have time?

Sorry for the late response. Overall looks good to me, I just added some minor comments.

mitake:
@ahrtr Thanks for reviewing, I'll update based on your comments.

ahrtr:
> @ahrtr Thanks for reviewing, I'll update based on your comments.

@mitake Please ping me again once you resolve all comments, so that I will review & approve this PR in a timely manner.

server/auth/store.go
mitake: Thanks for pointing out, I mistakenly used `BatchTx`. But after changing it to `ReadTx`, I saw below error of the unit test:
```
--- FAIL: TestGetAllUsers (0.37s)
    --- FAIL: TestGetAllUsers/Empty_by_default (0.37s)
panic: do not use unsafeRange on non-keys bucket [recovered]
	panic: do not use unsafeRange on non-keys bucket
```

It's caused by this line: https://github.com/etcd-io/etcd/blob/main/server/storage/backend/read_tx.go#L85

This is the reason buckets other than the one for keys shouldn't allow unsafe range: https://github.com/etcd-io/etcd/blob/0bac6096325c20b0709a3565fbf5808aa7fbf917/server/storage/backend/batch_tx.go#L39

I'm still not sure read txns actually see anomalies of duplicated keys. But probably it's safe to use BatchTx for `AuthBackend.GetAllUsers()`. How do you think?

ahrtr: I think we might need to use [UnsafeForEach](https://github.com/etcd-io/etcd/blob/0bac6096325c20b0709a3565fbf5808aa7fbf917/server/storage/backend/read_tx.go#L53) in this case.

ahrtr: The range operation only makes sense for the key space/buckets, I think it should be the reason why the original author introduces the restriction. 

For non-key space/buckets, users are supposed to perform a single key/value lookup or ListAll operation. For ListAll, users should use [UnsafeForEach](https://github.com/etcd-io/etcd/blob/0bac6096325c20b0709a3565fbf5808aa7fbf917/server/storage/backend/read_tx.go#L53)

mitake: Yeah I see, let me examine `UnsafeForEach` and update this PR.

mitake:
@ahrtr Thanks for reviewing. I pushed a new commit for resolving your comments, could you check? If it's ok for you, I'll squash the commits.

ahrtr:
> @ahrtr Thanks for reviewing. I pushed a new commit for resolving your comments, could you check? If it's ok for you, I'll squash the commits.

Thanks @mitake for the feedback. Yes, the latest commit looks good.  Please sign-off all commits using command `git rebase --signoff HEAD~2`. There are also some test failures, please take a look. I will review this PR again once you resolve these.

server/auth/store.go
```diff
@@ -200,7 +200,6 @@ type AuthBackend interface {
 	BatchTx() AuthBatchTx
 
 	GetUser(string) *authpb.User
-	GetAllUsers() []*authpb.User
```
ahrtr, line 203: It seems that we should keep `GetAllUsers()`, and we only need to change the implementation of `(atx *authReadTx) UnsafeGetAllUsers()` as you already did.

server/auth/store.go
```diff
@@ -571,7 +577,11 @@ func (as *authStore) UserGet(r *pb.AuthUserGetRequest) (*pb.AuthUserGetResponse,
 }
 
 func (as *authStore) UserList(r *pb.AuthUserListRequest) (*pb.AuthUserListResponse, error) {
-	users := as.be.GetAllUsers()
+	tx := as.be.ReadTx()
+	tx.Lock()
+	defer tx.Unlock()
+
+	users := tx.UnsafeGetAllUsers()
```
ahrtr, line 584:
```suggestion
users := as.be.GetAllUsers()
```

I should add this comment earlier, sorry for the confusion. The readTx indeed can't see the changes in the batchTx before being committed.

server/storage/schema/auth.go
```diff
@@ -111,6 +111,7 @@ func (atx *authBatchTx) Lock() {
 
 func (atx *authBatchTx) Unlock() {
 	atx.tx.Unlock()
+	atx.tx.Commit()
```
ahrtr, line 114: It looks like a big change although it's just one line code change.

ahrtr:
> @ahrtr Thanks for checking again. The root cause of the failed integration test (`TestUserDelete`) was that read tx cannot observe some updates introduced by batch tx when a number of pending writes isn't enough and not committed. I added explicit commit like this: https://github.com/etcd-io/etcd/pull/13954/files#diff-0ed8c872c10300094c1789395835940a91a943e8ec9f244d401e97de6ad6e69cR114 Could you check the change?

Yes, a boltDB ReadTx indeed can't see the changes in a boltDB BatchTx because we need to guarantee `repeatable read`. But in etcd, we we need to guarantee that each completed backend batchtx transaction is visible to following transaction no matter they are committed or not, so we update the readBuf each time when unlocking a batchTx, please see [batch_tx.go#L310](https://github.com/etcd-io/etcd/blob/7a1cecd5fafcea03adfb11b6bed49e901cd1dcf2/server/storage/backend/batch_tx.go#L310) .  

So could you double confirm the test failures?

server/storage/schema/auth.go
mitake: Yeah, I added a comment for this line. Anyway, I revived `UnsafeGetAllUsers()`, which uses batchTx. So technically it's not needed anymore (and is safe in the previous versions). But keeping it would be good for defensive purpose so I want to keep the `Commit()` call. How do you think?

mitake:
@ahrtr Updated based on your comments, could you check?

server/storage/schema/auth.go
ahrtr: It took me about 2 hour to figure out the root cause why `ReadTx can miss some writes issued by its predecessor BatchTx.`.

When users delete something, i.e. a user, from the [Auth](https://github.com/etcd-io/etcd/blob/f94af77d4ad17effb9208808ff82ba181f9e64cd/server/storage/schema/bucket.go#L52) bucket, actually etcd just deletes the data from boltDB, and it doesn't delete the data from the read buf at all. It's exactly the reason why the following readTx may still can read the data immediately after deleting it. Once the periodically commit goroutimes commits all pending data, then readTx can see the new data. 

But for the k/v data in the [Key](https://github.com/etcd-io/etcd/blob/f94af77d4ad17effb9208808ff82ba181f9e64cd/server/storage/schema/bucket.go#L43) bucket, there isn't such issue, because each time when users delete a k/v pair, etcd just add a new k/v pair with tombstone flag set. So it updates the writeBuf at [batch_tx.go#L364](https://github.com/etcd-io/etcd/blob/f94af77d4ad17effb9208808ff82ba181f9e64cd/server/storage/backend/batch_tx.go#L364), and the readBuf at [batch_tx.go#L310](https://github.com/etcd-io/etcd/blob/f94af77d4ad17effb9208808ff82ba181f9e64cd/server/storage/backend/batch_tx.go#L310). So my previous comment [issuecomment-1169362104](https://github.com/etcd-io/etcd/pull/13954#issuecomment-1169362104) is still valid.

Actually all other buckets have such issue as well, but if we use batchTx to read data, then we will not see such issue, such as lease. 

Anyway, I am OK to keep the `atx.tx.Commit()` for now.

Let me take care of the rest of stuff later in separate PR if needed.

ahrtr (approved):

ahrtr:
> @ahrtr Updated based on your comments, could you check?

Thanks for the fix.  Overall looks good to me. 

Let me take care of all other confusion stuff later. Please see my comment in https://github.com/etcd-io/etcd/pull/13954#discussion_r912407969

mitake:
@ahrtr Thanks for checking and merging. I'll also open PRs for backporting this change to the stable releases.

server/storage/schema/auth.go
mitake: Yeah, I feel the semantics of ReadTx and BatchTx is sometime counterintuitive and needs careful coding. If you can check other usages, it's really great. I personally want to revisit the value of read/write buffer of transactions.

ahrtr: > I personally want to revisit the value of read/write buffer of transactions.

Sure.

PR #13643, opened 2022-01-25
etcdctl: Remove V2 API commands

serathius:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Part of https://github.com/etcd-io/etcd/issues/12913
cc @ptabor

PR #13963, opened 2022-04-19
Add verification consistent index is (nearly) never decreasing

ptabor:


server/storage/schema/cindex.go
```diff
@@ -57,13 +59,31 @@ func ReadConsistentIndex(tx backend.ReadTx) (uint64, uint64) {
...
 
+func UnsafeUpdateConsistentIndexForce(tx backend.BatchTx, index uint64, term uint64) {
+	unsafeUpdateConsistentIndex(tx, index, term, true)
+}
+
 func UnsafeUpdateConsistentIndex(tx backend.BatchTx, index uint64, term uint64) {
+	unsafeUpdateConsistentIndex(tx, index, term, false)
+}
+
+func unsafeUpdateConsistentIndex(tx backend.BatchTx, index uint64, term uint64, allowDecreasing bool) {
```
ahrtr, line 70: Should we call `UnsafeUpdateConsistentIndexForce` in [backup_command.go#L328](https://github.com/etcd-io/etcd/blob/4555fc39988a723992748fe36704e114fa80e2a7/etcdutl/etcdutl/backup_command.go#L328)?

ptabor: Most likely. I assumed that the bucket creation code above suggests that the consistent-index is not present...
 - either we have testing hole
 - or the copying code before is actually not copying this specific bucket, just data.

ahrtr: I am thinking why do we need to add `allowDecreasing`? Can you elaborate a reasonable & practical case which could lead to decrease of consistent_index? 

If you can't, then we can just remove `allowDecreasing`, and always perform the verification in tests.

ahrtr: Partially correct to me.

 `etcdutl restore ` is the only practical case which may lead to decrease of consistent_index, because it reset the `index` & `term` to `len(ents)` and `1` respectively, see [v3_snapshot.go#L449](https://github.com/etcd-io/etcd/blob/e02ac59cfc4e750f4b0a2a121e20db8e6775d5cd/etcdutl/snapshot/v3_snapshot.go#L449) . 

I don't think it's possible for `etcdutl backup` to decrease the consistent_index, and [UnsafeUpdateConsistentIndex](https://github.com/etcd-io/etcd/blob/e02ac59cfc4e750f4b0a2a121e20db8e6775d5cd/etcdutl/etcdutl/backup_command.go#L323-L328) is only called for v2 scenario. Since we are going to remove v2 commands ( see [13643](https://github.com/etcd-io/etcd/pull/13643) ), so we can even simplify the `etcdutl backup `. I just submitted a PR for this [13972](https://github.com/etcd-io/etcd/pull/13972) 

Suggested actions: 
1. @serathius rebases [13643](https://github.com/etcd-io/etcd/pull/13643) and manage to get it merged . We have already removed server side code for V2, so there is no reason to keep the v2 command. If users need to use v2 command, then they can get the binary from previous releases, such as 3.5 or 3.4. Note: v2 discovery is a special case for now. 
2. Afterwards, I will rebase my PR [13972](https://github.com/etcd-io/etcd/pull/13972);
3. For this PR, I think it's good now with only the minor naming [comment](https://github.com/etcd-io/etcd/pull/13963#discussion_r854671077)

ptabor: Done

server/verify/verify.go
```diff
@@ -27,8 +27,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const ENV_VERIFY = "ETCD_VERIFY"
-const ENV_VERIFY_ALL_VALUE = "all"
+const EnvVerifyStorageWalValue = "storage_wal"
```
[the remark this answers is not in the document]
ptabor: Done

server/verify/verify.go
```diff
@@ -27,8 +27,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const ENV_VERIFY = "ETCD_VERIFY"
-const ENV_VERIFY_ALL_VALUE = "all"
+const ENV_VERIFY_VALUE_STORAGE_WAL verify.VerificationType = "storage_wal"
```
ahrtr, line 30: Although golang prefers camel case instead of snake case, personally I think it should be OK if we keep it consistent.

ahrtr (approved):

PR #13965, opened 2022-04-20
Documentation: Update contributor guide and remove outdated ROADMAP.md

serathius:
cc @ptabor @spzala


Documentation/contributor-guide/logging.md
```diff
@@ -1,6 +1,6 @@
 # Logging Conventions
 
-etcd uses the [capnslog][capnslog] library for logging application output categorized into *levels*. A log message's level is determined according to these conventions:
+etcd uses the [zap][zap] library for logging application output categorized into *levels*. A log message's level is determined according to these conventions:
```
spzala, line 3: Good catch. We should also remove the contentions mentioned here, they are from capnslog. With Zap, etcd supports `debug, info, warn, error, panic, fatal. We can just list them without any details as the types are self-explanatory. Optionally, add a sentence something like "refer to the https://github.com/uber-go/zap/blob/master/logger.go to learn more about these conventions".

spzala (commented):
lgtm @serathius thanks! I have one comment suggesting another a change with Zap logging.

Documentation/contributor-guide/logging.md
serathius: ok, let me send a separate PR for that.

spzala: Ok, thanks! That's fine. I will send it after this merge as you are in mid of many other things.

spzala (approved):

PR #13966, opened 2022-04-20
Fixed potential goroutine leak due to p.Ready() receive in pkg/proxy and dependents.

VladSaioc:
Fixed potential goroutine leak due to receive operations on p.Ready().

If the server creation process via [NewServer](https://github.com/etcd-io/etcd/blob/4555fc39988a723992748fe36704e114fa80e2a7/pkg/proxy/server.go#L193) encounters an error at [line 257](https://github.com/etcd-io/etcd/blob/4555fc39988a723992748fe36704e114fa80e2a7/pkg/proxy/server.go#L257), it will fail to spawn `go s.listenAndServer` at [line 265](https://github.com/etcd-io/etcd/blob/4555fc39988a723992748fe36704e114fa80e2a7/pkg/proxy/server.go#L265), thus not closing `s.readyc` ([line 289](https://github.com/etcd-io/etcd/blob/4555fc39988a723992748fe36704e114fa80e2a7/pkg/proxy/server.go#L289)) . Any receive on `p.Ready()` for the server `p` returned by `NewServer` will block indefinitely.

The introduced `select` statement captures the error sent to `s.errc` on [line 258](https://github.com/etcd-io/etcd/blob/4555fc39988a723992748fe36704e114fa80e2a7/pkg/proxy/server.go#L258), reports  the abnormal behaviour, and unblocks the receiver thread of `p.Ready()`.

### DIFF
--- a/pkg/proxy/server_test.go
+++ b/pkg/proxy/server_test.go
@@ -73,7 +73,9 @@ func testServer(t *testing.T, scheme string, secure bool, delayTx bool) {
 		cfg.TLSInfo = tlsInfo
 	}
 	p := NewServer(cfg)
-	<-p.Ready()
+
+	waitForServer(t, p)
+
 	defer p.Close()
 
 	data1 := []byte("Hello World!")
@@ -196,7 +198,9 @@ func testServerDelayAccept(t *testing.T, secure bool) {
 		cfg.TLSInfo = tlsInfo
 	}
 	p := NewServer(cfg)
-	<-p.Ready()
+
+	waitForServer(t, p)
+
 	defer p.Close()
 
 	data := []byte("Hello World!")
@@ -246,7 +250,9 @@ func TestServer_PauseTx(t *testing.T) {
 		From:   url.URL{Scheme: scheme, Host: srcAddr},
 		To:     url.URL{Scheme: scheme, Host: dstAddr},
 	})
-	<-p.Ready()
+
+	waitForServer(t, p)
+
 	defer p.Close()
 
 	p.PauseTx()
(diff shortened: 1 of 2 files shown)


pkg/proxy/server_test.go
```diff
@@ -73,7 +73,13 @@ func testServer(t *testing.T, scheme string, secure bool, delayTx bool) {
 		cfg.TLSInfo = tlsInfo
 	}
 	p := NewServer(cfg)
-	<-p.Ready()
+
+	select {
+	case <-p.Ready():
+	case err := <-p.Error():
+		t.Fatal(err)
+	}
```
ahrtr, line 81: There are lots of duplicated code, can you wrap them into a common function or method?

VladSaioc: Done! There is a wrapper around these selects in `server_test.go` now. There is another possible approach:
A method on `(*server)`, but this requires a method definition on the `Server` interface as well. The method (tentative name `WaitUntilReady`) could either have a signature of:
`func () error`
And handling would be given as:
`if err := s.WaitUntilReady(); err != nil { [handle error] }`
or
`func (errcb func(error))`
with the handling as:
s.WaitUntilReady(func (err error) { [handle error] })`

Let me know if these approaches are more desirable.

ahrtr (approved):
LGTM

cc @serathius @spzala @ptabor

ahrtr (changes requested):
Please squash the commits. Thanks.

ahrtr:
Looks good. I will approve the change once the pipeline is green. cc @serathius @spzala @ptabor

ptabor (approved):

PR #13910, opened 2022-04-08
*: update golang.org/x/crypto

serathius:
cc @ptabor @ahrtr @spzala

PR #13969, opened 2022-04-21
*: Update golang.org/x/crypto to latest

cmurphy:
Update crypto to address CVE-2022-27191.

The CVE fix is added in 0.0.0-20220315160706-3147a52a75dd but this
change updates to latest.


ahrtr:
Thanks @cmurphy.
 
Two comments:
1. etcd doesn't use the package `golang.org/x/crypto/ssh` at all;
2. I recall @serathius tried to upgrade to the latest version in [13910](https://github.com/etcd-io/etcd/pull/13910), but it caused some pipeline failures.

I am open to discussion. But first of all, the pipeline should be green firstly.

ptabor:
```
Error: stderr: ../server/auth/store.go:33:2: missing go.sum entry for module providing package golang.org/x/crypto/bcrypt (imported by go.etcd.io/etcd/tests/v3/framework/integration); to add:
```

Please execute `scripts/fix.sh`

cmurphy:
Thanks for the reviews. @ahrtr -

> etcd doesn't use the package golang.org/x/crypto/ssh at all;

Our CVE scanner flagged this dependency. If you think this CVE doesn't affect etcd and it's not worth upgrading, we can add it to the ignore list in our scanner and close this. But you may get other similar PRs for the same reason.

ahrtr:
> Thanks for the reviews. @ahrtr -
> 
> > etcd doesn't use the package golang.org/x/crypto/ssh at all;
> 
> Our CVE scanner flagged this dependency. If you think this CVE doesn't affect etcd and it's not worth upgrading, we can add it to the ignore list in our scanner and close this. But you may get other similar PRs for the same reason.

It's OK to bump the crypto version. Please make sure the pipeline is green firstly just as I mentioned above. Since you are first time contributor, so the pipeline doesn't get triggered automatically (I have no permission to trigger it for now) ; so I suggest you to create a PR in your own forked etcd repo, please add a comment here once the PR is green.

cmurphy:
There was a merge conflict so I rebased against main, the pipeline is green on my end https://github.com/cmurphy/etcd/pull/1

ahrtr (approved):
LGTM

Thank you!

Usually bumping to the newest version (released on Apr 11, 2022) is not a good practice, but we are still in the early stage of 3.6 release train. So it's OK to me.

PR #13982, opened 2022-04-25
allocate unique port for each member in test cases

ahrtr:
Fix test failure below,
```
v3_snapshot_test.go:255: initial cluster m0=unix://localhost:8550,m1=unix://localhost:8550,m2=unix://localhost:39058 has duplicate url
```

FYI. https://github.com/etcd-io/etcd/runs/6149754823?check_suite_focus=true


tests/integration/snapshot/v3_snapshot_test.go
```diff
@@ -286,10 +286,23 @@ func restoreCluster(t *testing.T, clusterN int, dbPath string) (
...
 	for i := 0; i < n; i++ {
-		rand.Seed(int64(time.Now().Nanosecond()))
-		u, _ := url.Parse(fmt.Sprintf("unix://localhost:%d", rand.Intn(45000)))
+		u, _ := url.Parse(fmt.Sprintf("unix://localhost:%d", allocateUniquePort(2000, 45000, allocatedPorts)))
 		urls[i] = *u
 	}
 	return urls
 }
+
+func allocateUniquePort(minPort, maxPort int, allocatedPort map[int]struct{}) int {
```
serathius, line 297: Wow, just wow. How many different places we have in tests that allocate shared resources (port,unix sockets, etc)?

Instead of having this generator, can we maybe try use a `LocalListenCount` https://github.com/etcd-io/etcd/blob/main/tests/framework/integration/cluster.go#L88 or adapt https://github.com/etcd-io/etcd/blob/b00bb535f09c627c4217a9f5a22d1327c4560219/tests/framework/integration/cluster.go#L525-L530 function to return port number.

ahrtr: The new added `allocateUniquePort` is only used by snapshot related test, which depends on [embed.StartEtcd](https://github.com/etcd-io/etcd/blob/b00bb535f09c627c4217a9f5a22d1327c4560219/tests/integration/snapshot/v3_snapshot_test.go#L45-L71) to start etcd directly. 

We can consider how to unify all such resources allocation in a separate & dedicate PR.

serathius: ok, however are you sure that ports generated by `newEmbedURLs` were conflicting with itself or maybe with ports generated by `newLocalListener` ?

I want to make sure we are making a proper improvement, not just continue with band-aid solutions.

ahrtr: Yes, basically I am sure about it based on error message. The call stack is simple and clear.

ptabor: Another option for generator is to create a pool of ports guaranteed to be unused using: 
  a) creation of listeners with port 0 (https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/httptest/server.go;l=68)
  b) getting the port (https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/httptest/server.go;l=130#:~:text=129-,130,-131)
  c) closing the listener

ptabor (approved):

tests/integration/snapshot/v3_snapshot_test.go
ahrtr: Yes, it's a feasible solution to unify the port allocation. Let's address it in a separate & dedicate PR.

endocrimes, line 297: we're not guaranteed to get a freeport this way - and it's more state to track, which won't scale too well when expanding across test cases

we might wanna do something more like:

```
t.Helper()
addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
if err != nil {
	t.Failf("failed to resolve localhost: %v", err)
}

l, err := net.ListenTCP("tcp", addr)
defer l.Close()
if err != nil {
	t.Failf("failed to listen tcp: %v", err)
}
return l.Addr().(*net.TCPAddr).Port
```

endocrimes (changes requested):

tests/integration/snapshot/v3_snapshot_test.go
ahrtr: It isn't worse than current implementation. Note that this is only used for snapshot related cases. Let's unify the port allocation in a separate PR.

ahrtr: @ptabor  already proposed a similar solution above.

endocrimes: It's not worse - but it's not a marked improvement, we could use this to prove out a better implementation that can then migrate to be shared everywhere.

State tracking when a ~simpler solution would be more reliable doesn't really seem too necessary?

ahrtr: Done.

ahrtr: Updated per [discussion_r857364543](https://github.com/etcd-io/etcd/pull/13982#discussion_r857364543).  PTAL @ptabor @serathius 

With regard to unify the port allocation such as [cluster.go#L525-L530](https://github.com/etcd-io/etcd/blob/b00bb535f09c627c4217a9f5a22d1327c4560219/tests/framework/integration/cluster.go#L525-L530) , I tend to keep it as it's for now, because the existing behavior is easier to debug for test failure, because the port number is more predictable. Anyway, let's address it in a separate PR.

tests/integration/snapshot/v3_snapshot_test.go
```diff
@@ -287,8 +287,15 @@ func restoreCluster(t *testing.T, clusterN int, dbPath string) (
...
-		u, _ := url.Parse(fmt.Sprintf("unix://localhost:%d", rand.Intn(45000)))
+		l, err := net.Listen("tcp", "127.0.0.1:0")
+		if err != nil {
+			panic(fmt.Sprintf("failed to listen on a port: %v", err))
+		}
+
+		nPort := l.Addr().(*net.TCPAddr).Port
+		defer l.Close()
+
+		u, _ := url.Parse(fmt.Sprintf("unix://localhost:%d", nPort))
```
serathius, line 298: Why use `tcp` port for unix socket?

ahrtr: done.

ptabor (approved):

ahrtr:
@serathius PTAL, thanks.

tests/framework/integration/cluster.go
```diff
@@ -522,7 +522,7 @@ func isMembersEqual(membs []*pb.Member, wmembs []*pb.Member) bool {
 	return cmp.Equal(membs, wmembs, cmpopts.IgnoreFields(pb.Member{}, "ID", "PeerURLs", "ClientURLs"))
 }
 
-func newLocalListener(t testutil.TB) net.Listener {
+func NewLocalListener(t testutil.TB) net.Listener {
 	c := atomic.AddInt32(&LocalListenCount, 1)
 	// Go 1.8+ allows only numbers in port
 	addr := fmt.Sprintf("127.0.0.1:%05d%05d", c+BasePort, os.Getpid())
```
endocrimes, line 528: unrelated-ish: but nice.

endocrimes (approved):

serathius (approved):

PR #8238, opened 2017-07-10
v3rpc: Let clients establish unlimited streams

yudai:
Issue Observed
----------------

From etcd v3.2.0, custom clients can create up to 100 streams for watch and keepalive. This limitation can cause unexpected behavior to custom clinets/proxies that need to establish many streams.

Issue Detail
------------

etcd v3.2.0 updated the go-grpc library and it had a (breaking) change in it, which caps the default max streams per client to 100 (https://github.com/grpc/grpc-go/pull/1071). 

The official client and proxies do not establish many streams for now. However, this limitation can have impact for third party libraries that need to establish more streams for some reason.

Proposal
---------

In this commit, we simply set the number of streams to the max number. We may have configuration property for this as well.

```
From go-grpc v1.2.0, the number of max streams per client is set to 100
by default by the server side. This change makes it impossible
for third party proxies and custom clients to establish many streams.
```

PR #13985, opened 2022-04-25
 fix https-> ConcurrentMaxStream bug

YANGJINJUE:
when etcd enable https,many requests will blocked,because of etcd->server.go use a default ConcurrentMaxStream:250.

Fix : I set MaxConcurrentStreams: math.MaxUint32,it work good,we haved deploy the etcd-fix in production.


serathius (changes requested):
Have you tried to use other etcd versions to verify if issue introduced in v3.5.3 and not any other version? From surface I don't see any reason why v3.5.3 would have introduced this limitation. 

Have you investigated why there are so many requests blocked? I would like to first understand if there was any changes in client/server that causes us to require so many concurrent streams.

I don't think this a proper fix, as it doesn't identify the root cause, but tries to fix a symptom.

ptabor:
Unbound number of connections is likely to cause: 
  - OOM errors as each connection comes with memory overhead. 
  - depletion of networking ports (there are only 63k)

Please use 'debug' interface (pprof) or SIGSEV to check what code the connections are being stack on.

YANGJINJUE:
> 
if you disable https，it work good。the problem

dojiong:
ulimited grpc streams is first introduced at #8238, but it's not working under tls connection

ahrtr:
Comments:
1. Please update commit message using English so that all contributors can understand;
2. Please clarify why [8238](https://github.com/etcd-io/etcd/pull/8238) doesn't work when http2/tls is enabled. 

It will be helpful if gRPC experts can provide comments here. @menghanl @iamqizhao @dfawley

ptabor:
> Comments:
> 
> 1. Please update commit message using English so that all contributors can understand;

> 2. Please convince us or clarify why [8238](https://github.com/etcd-io/etcd/pull/8238) doesn't work when http2/tls is enabled.

I assume it's because of  our layout:  we have `httpserver` -> `cmux` -> `grpc`. 
`grpc` is configured to allow unlimited number of connections, but `http server` sticks to its defaults. 

What I don't understand is: 
  1. What makes the TLS special ? The code path for `insecure` is also not setting the max concurrent connections. 
     Do you mean that plane http server is: 
            a) not capping the connections at 100 ?
            b) capping, but not causing problems, as establishing new http connections is lightweight vs. expensive TLS handshake 
            c) [not really tested for plain http scenario] 

  2. Does it mean currently a `101` clients cannot establish 101 watch streams (each client 1 watch stream) ?
      I assume it's not a problem for k8s as it has O(3) api-servers, each creating a lot of watch-streams over the same connection.... but might be an issue for other etcd applications. 

My recommendation and top of Benjamin's: 
  - let's expose this as an explicit flag that applies to both http, https & grpc (with default of 100).  
  - let's have a test that shows that it works (e.g. when the flag is 5,  the 6 independent watch connections cannot be established)
  - let's have some measurement on memory impact on number of connections. E.g Prometheus graphs showing a load-test of growing number of connections versus RSS memory usage, to be able to comment on the flag the memory expections for the high values of the flag.

ptabor:
> > 1. What makes the TLS special ? The code path for `insecure` is also not setting the max concurrent connections.
> 
> The reason is etcd creates a new goroutine for gRPC server in `insecure` mode. See [serve.go#L129](https://github.com/etcd-io/etcd/blob/d849269dab5c07c1259de4441809d41135e7d0d8/server/embed/serve.go#L129)

the `serve` method is executed once per each listening port, so it's creates `O(opened ports)` go-routines, in the code-path you mentioned.   I don't see (but likely missing something) how this `few additional` goroutines could be fundamental change for `>200 connections`.

ahrtr:
> the `serve` method is executed once per each listening port, so it's creates `O(opened ports)` go-routines, in the code-path you mentioned. I don't see (but likely missing something) how this `few additional` goroutines could be fundamental change for `>200 connections`.

The point is the gRPC server can receive client requests directly when creating dedicate goroutines just as I pointed out at [serve.go#L129](https://github.com/etcd-io/etcd/blob/d849269dab5c07c1259de4441809d41135e7d0d8/server/embed/serve.go#L129); in other words, it doesn't go through the `http.Server` any more, accordingly it will not be impacted by the MaxConcurrentStreams settings of http2.

ahrtr:
@YANGJINJUE  could you update this PR per above comments?

ahrtr:
@YANGJINJUE Please update this PR per comments [issuecomment-1109346863](https://github.com/etcd-io/etcd/pull/13985#issuecomment-1109346863) and [issuecomment-1109512194](https://github.com/etcd-io/etcd/pull/13985#issuecomment-1109512194)

ahrtr:
@YANGJINJUE Please resolve the following comments:
1. Squash the commits (Refer to [git-squash](https://www.git-tower.com/learn/git/faq/git-squash) ), and change the commit message into English;
2. Add an integration or E2E test to verify the change. We can address it in a separate PR.

Battlefield-Pony:
Sorry but I have not get a clue on the `250` Connections where's the `250` comeing from?

ahrtr:
FYI.
[server.go;l=170](https://cs.opensource.google/go/x/net/+/master:http2/server.go;l=170)
[server.go;l=58](https://cs.opensource.google/go/x/net/+/master:http2/server.go;drc=1d687d428aca0546c0ca84160c8700ee521e9fb9;l=58)

leslie-tsang:
@YANGJINJUE @ahrtr Hello there, After having tested this PR and proving that it solves our [problem](https://github.com/etcd-io/etcd/issues/14077). 
> seems [issue/14065](https://github.com/etcd-io/etcd/issues/14065) can be solved as well. xD

It will be nice if this PR can be merge as soon as possible. :)

YANGJINJUE:
how to reopen the pull request

Issue #13973, opened 2022-04-22
Verification of exchanged etcd snapshots


ptabor:
We observed following test flakes: 

Test failure - flake (https://github.com/etcd-io/etcd/runs/6091231156?check_suite_focus=true):

```
    logger.go:130: 2022-04-20T08:11:00.714Z	FATAL	m0	Verification failed	{"member": "m0", "data-dir": "/tmp/TestV3WatchRestoreSnapshotUnsync501892535/002/etcd4044943490", "error": "backend.ConsistentIndex (24) must be >= last snapshot index (25)"}
```

Discussion: https://github.com/etcd-io/etcd/pull/13942#issuecomment-1103833733

What happens is that a lagging follower is downloading a snapshot from leader. 
The snapshot is marked as representing exact 'wal-index' (i.e. snapshot-index). But within the snapshot file, the cindex < snapshot-index. 

We should: 
  - understand in what situation a generated snapshot (that is produced within Apply goroutine) is not reflecting the most recent cindex and fix. 
  - Add a verification (at least in tests, i.e. ETCD_VERIFY='snapshots') that:
      - the snapshot being produced has consistent cindex with it's expected WAL-log position
      - the snapshot being received matches the expectations

ahrtr:
The issue was caused by the intermediate change (see below) in PR [13942](https://github.com/etcd-io/etcd/pull/13942).  

[server.go#L1823](https://github.com/etcd-io/etcd/blob/e02ac59cfc4e750f4b0a2a121e20db8e6775d5cd/server/etcdserver/server.go#L1823)

```
if  ar != nil  // This is not correct!!!
```

I believe we should be good with the final change in above PR,

```
if !applyV3Performed || (ar != nil && ar.err != nil) 
```

I agree that we should add a snapshot verification when a follower receive a snapshot.

PR #13988, opened 2022-04-25
Verify consistent_index in snapshot must be greater than the snapshot index

ahrtr:
Fix [13973](https://github.com/etcd-io/etcd/issues/13973)

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -41,6 +41,7 @@ import (
 	"go.etcd.io/etcd/api/v3/version"
 	"go.etcd.io/etcd/client/pkg/v3/fileutil"
 	"go.etcd.io/etcd/client/pkg/v3/types"
+	"go.etcd.io/etcd/client/pkg/v3/verify"
 	"go.etcd.io/etcd/pkg/v3/idutil"
 	"go.etcd.io/etcd/pkg/v3/pbutil"
 	"go.etcd.io/etcd/pkg/v3/runtime"
@@ -971,6 +972,7 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, apply *apply) {
 	// Eventually the new consistent_index value coming from snapshot is overwritten
 	// by the old value.
 	s.consistIndex.SetBackend(newbe)
+	verifySnapshotIndex(apply.snapshot, s.consistIndex.ConsistentIndex())
 
 	// always recover lessor before kv. When we recover the mvcc.KV it will reattach keys to its leases.
 	// If we recover mvcc.KV first, it will attach the keys to the wrong lessor before it recovers.
@@ -1065,6 +1067,14 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, apply *apply) {
 	ep.appliedi = apply.snapshot.Metadata.Index
 	ep.snapi = ep.appliedi
 	ep.confState = apply.snapshot.Metadata.ConfState
+}
+
+func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
+	verify.Verify(func() {
+		if cindex != snapshot.Metadata.Index {
+			panic(fmt.Sprintf("consistent_index(%d) isn't equal to snapshot index (%d)", cindex, snapshot.Metadata.Index))
+		}
+	})
 }
 
 func (s *EtcdServer) applyEntries(ep *etcdProgress, apply *apply) {
--- a/client/pkg/verify/verify.go
+++ b/client/pkg/verify/verify.go
@@ -37,18 +37,30 @@ func IsVerificationEnabled(verification VerificationType) bool {
 	return env == string(ENV_VERIFY_VALUE_ALL) || env == strings.ToLower(string(verification))
 }
 
-// EnableVerifications returns a function that can be used to bring the original settings.
+// EnableVerifications sets `ENV_VERIFY` and returns a function that
+// can be used to bring the original settings.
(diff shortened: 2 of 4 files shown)


server/etcdserver/server.go
```diff
@@ -1067,6 +1069,15 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, apply *apply) {
 	ep.confState = apply.snapshot.Metadata.ConfState
 }
 
+func (s *EtcdServer) verifySnapshot(snapshot raftpb.Snapshot) {
+	verify.Verify(func() {
+		cindex := s.consistIndex.ConsistentIndex()
+		if cindex < snapshot.Metadata.Index {
```
ptabor, line 1075: Do we expect cindex > snapshot.Metadata.Index ? 
I know we can recover from such state, so it's not critical...
but if this is supposed to be defensive invariant for testing, I would assume we should aim for `==`

server/etcdserver/server.go
```diff
@@ -971,6 +972,7 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, apply *apply) {
 	// Eventually the new consistent_index value coming from snapshot is overwritten
 	// by the old value.
 	s.consistIndex.SetBackend(newbe)
+	s.verifySnapshot(apply.snapshot)
```
ptabor, line 975: This is pretty specific verification in the context of recently populated s.consistencyIndex from the snapshot. 
It's generic truth that last loaded snapshot should has `index < cindex`. 

Please consider: 

```
   verifySnapshotIndex(apply.snapshot, s.consistIndex.ConsistentIndex())
```

```
func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex int) {
	verify.Verify(func() {
		if cindex != snapshot.Metadata.Index { ... }
    }
}
```

ptabor (commented):
Thank you for the fix. 

Have you considered adding such verification on the snapshot producer side, 
such that we catch it when the 'inconsistent' snapshot is being prepared to be send over the wire apart of crashing the 'receiver' ?

ahrtr:
> Thank you for the fix.
> 
> Have you considered adding such verification on the snapshot producer side, such that we catch it when the 'inconsistent' snapshot is being prepared to be send over the wire apart of crashing the 'receiver' ?

Yes, it's a good point. I just updated the code to perform the verification on both the sender and receiver sides.

server/etcdserver/server.go
ahrtr: I think it's possible in theory although I am not 100% sure for now, because the entry.Index might be smaller than the cindex. Let's work with an imaginary (& extreme) example, 
1. The leader (S1) partially replicates some log entries at indices `i ~ j`; the leader crashes;
2. Another member (S2) is elected as the new leader, but its local `applied index` & `consistent_index` is still` i-1.` You may challenge this, please look at` figure 8` at `page 8 `in [raft.pdf](https://raft.github.io/raft.pdf);
3. S2 replicates new log entries with index` i` to followers, including S1; S1's local `consistent_index` is `j` (>i),  so which will keep unchanged. 
4. S2 crashes for whatever reason, and S1 is elected as the leader again.  
5. If S1 happens to replicates snapshot to a slower follower of a new member, then the `consistent_index` in the snapshot will be greater than the `snapshot.Metadata.Index`.

ptabor: > 3.  ...S1's local `consistent_index` is `j` (>i),  so which will keep unchanged. But S1's applied index is updated to `i`.

I'm confused in this step. What do you mean by 'consistent_index' ? I believe S1 might have WAL log replicated to `j`, but I assume that the only way to change `consistent_index` is by application. 
When leader S1 crashes and is restarted it's consistent index is populated from latest  snapshot `l<=i` and the only way to advance it is through apply.

ahrtr: YES, cindex can only be advanced by applier.

S1's consistent_index and WAL LOG index were advanced to j at step 1. But S2 somehow did not received the message from S1, and it is what the "partially" means at step 1.

Although raft is much simpler than paxos, it is still very subtle and there are some corner cases.

ahrtr: You know, etcd applies the entries in a separate goroutine. It means that etcd applies entries [concurrently](https://github.com/etcd-io/etcd/blob/bb63484dcefd7339a2ccbff7ad343a3ef949c381/server/etcdserver/raft.go#L213). So it's possible etcd applies the entries successfully, but fails to replicates the data to some followers. 

So at step1, the leader (S1) might apply the entries successfullly, accordingly the consistent_index is advanced to j. But it failed to replicate the data to S2, accordingly S2's conssitent_index is still i-1.

ptabor: I still don't get it. I understand the Fig.8 story from RAFT. 

But I think that crucial aspect here is that application is always following commitment. 

So: 
 1) The S1 leader must have confirmed j) is replicated to majority of replicas and decided/WAL checkpointed `committedIndex=j`.
      After the restart the S1 follower bootstraps from snapshot & WAL and positions itself (applied/committed/consistent=j). 
  3) In 3, I don't see a code-paths that would allow 'decrease' the applied index from `j->i`. Any replication coming from S2 must be either no-change (if its `index<=committed`) and gets ignored by S1 as we already have it in our WAL log... or it's `index>committed`, so would need to wait for another 'committed' signal before application... and in this scenario `appliedIndex` would only grow as expected.

ahrtr: Resolved. PTAL

ptabor (approved):

Issue #13980, opened 2022-04-24
Improve and automate releases


serathius:
As mentioned on v3.5.3 release issue it took me 4 hours to run the release script. It required a lot of desperate patches to run and I don't expect anyone sane to do the same. 

Some of the issues encountered when running release script:
* Path from local machine included in binaries. Requires a lot of manual steps to setup an environment to avoid leaking.
* Release sources/execute scripts from different branches in hard to predict way. There are a lot of bugs present in scripts from old release branches and incompatible changes done.
 
 Changes that I needed to do to release v3.5.3 and v3.5.4 https://github.com/etcd-io/etcd/pull/13979
 
 Proposed improvements:
 * [x] Release scripts are tested in dry mode on main branch
 * [x] Release scripts are tested in dry mode on supported branches (release-3.5, release-3.4).
   * [x] Release scripts are tested on release-3.5
   * [x] Release scripts are tested on release-3.4
 * [ ] Process of building docker images is reproducible as it builds artifacts within docker image.
 * [ ] Merging commit runs release process that pushes development artifacts. 
 * [ ] Tagging git repo starts automatic release process that uploads (most/all artifacts).
 * [ ] Release process qualifies the release artifacts (runs e2e and functional tests)
 
 cc @ptabor @spzala @ahrtr

vivekpatani:
Is there any interest in generating rpms as a part of the release process? @serathius @serathius

mrueg:
One thing that might be worth to consider is using https://github.com/goreleaser/goreleaser to build binaries for github releases.

serathius:
@vivekpatani No there is no interest in expanding the scope of releases as we still need a lot of work to make quality of the release process reasonable.

@mrueg Thanks for suggestion, the main challenge is untangling the scripts used in existing release process.

protosam:
Took a look at one of the PR's and the checks look incomplete to me. The goal should probably be to ensure that is in a state where we can actually trust that what's merged is being tested against the support platforms.

I'll spend my next free cycle seeing if I can use the github workers for actions to be viable. They support Windows, Mac, and Linux. For the extra architectures supported, I'll see if I can make it happen qemu or something in github actions.

Feel free to add me as an assignee on this. Gladly willing to help.

serathius:
> Took a look at one of the PR's and the checks look incomplete to me. The goal should probably be to ensure that is in a state where we can actually trust that what's merged is being tested against the supported platforms.

[...]

PR #13992, opened 2022-04-26
tests: Test release script for release-3.5 branch

serathius:
Ref https://github.com/etcd-io/etcd/issues/13980


serathius:
cc @ptabor @ahrtr @ptabor

scripts/release.sh
```diff
@@ -63,6 +63,11 @@ main() {
   log_callout "BRANCH=${BRANCH}"
   log_callout "REPOSITORY=${REPOSITORY}"
   log_callout ""
+
+  if [ "${BRANCH}" != "main" ] && [ "${BRANCH}" != "release-3.5" ]; then
```
ahrtr, line 67: Just curious how do you or someone else release 3.4.x?

.github/workflows/release.yaml
```diff
@@ -22,3 +22,24 @@ jobs:
         Expire-Date: 0
         EOF
         DRY_RUN=true BRANCH=main ./scripts/release.sh --no-upload --no-docker-push 3.6.99
+  release35:
```
ahrtr, line 25: My immediate feeling is that probably we should add all release-3.5 related changes into branch `release-3.5` only. It's clearer to me, but the bad side is we need maintain different stuff/scripts for different releases (usually 3).

Maintaining everything only in the main branch is also an option. The good point is we don't need to maintain multiple copies (with only minor differences) ; but the bad side is we complicate the release script & config.  

It's a tradeoff. I prefer to the first option (maintain different scripts/configs for different releases).

scripts/release.sh
serathius: Goal is to give a clear signal that this script was not tested and is not expected to work with `release-3.4`. Instead of failing in random part of the execution for v3.4, goal is to give release manager signal that this script needs to be fixed.

.github/workflows/release.yaml
ptabor: Agree. If the testing scenario pre-existed in the the 3.5 branch... we would just update it when creating the 3.6 branch (main) with 3.6 specific changes.

scripts/release.sh
ptabor: Do you assume in general usage of script from 'main' against 3.5 branch ? Why this is a goal ?

One of the principles it breaks is 'build determinism'. If someone sync's to given sha on given branch and perform release, we expect the the produced artifacts to be as much identical as possible (i.e. golang version for building binaries is enforced), vs. pick part of the release logic from the 'head' branch.

ptabor (changes requested):
Marking as 'requesting changes' to mark the opens Benjamin's comment.

Issue #13991, opened 2022-04-26
etcd-dump-db not working


IamSatyaonline:
### What happened?

etcdctl tool is working fine but unable to run the etcdctl-dump-db.
below is the error we are getting.

bash-4.4$ etcd-dump-db -help
bash: etcd-dump-db: command not found

### What did you expect to happen?

 etcd-dump-db should be run to debug the dump file. Please help us to configure this command.

### How can we reproduce it (as minimally and precisely as possible)?

By running the command etcd-dump-db

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
bash-4.4$ etcd --version
etcd Version: 3.3.11
Git SHA: 2cf9e51d2
Go Version: go1.10.7
Go OS/Arch: linux/amd64


$ etcdctl version
bash-4.4$ etcdctl version
etcdctl version: 3.3.11
API version: 3.3
bash-4.4$

```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```shell
bash-4.4$
bash-4.4$ etcd-dump-db
bash: etcd-dump-db: command not found
```

spzala:
@IamSatyaonline did you install the tool? The tools, https://github.com/etcd-io/etcd/tree/main/tools, are not installed by default like etcdctl.

IamSatyaonline:
Hi @spzala 
I have installed etcd-3.4.16 and it is running inside container. is it possible to run the etcd-dump-db inside container with etcdctl client ? Could you please share the steps to install/enable etcd-dump-db tool along with etcdctl ? 
I see, etcd-dump-db is the part of 3.4.16 etcd package.

bash-4.4$ etcd --version
etcd Version: 3.4.16
Git SHA: d19fbe541
Go Version: go1.12.17
Go OS/Arch: linux/amd64

bash-4.4$ etcdctl version
etcdctl version: 3.4.16
API version: 3.4
bash-4.4$

ahrtr:
Please refer to the comment in [issuecomment-1110774613](https://github.com/etcd-io/etcd/pull/13994#issuecomment-1110774613)

You can also run the tool using command `go run ./tools/etcd-dump-db`.

[...]

Issue #13478, opened 2021-11-14
[Improvement] Use workspace mode to simplify the go.mod


ahrtr:
Currently there is lots of replace directives in each go.mod file, and most of them are some duplicated items.

In golang 1.18 to be release early next year, a new workspace mode will be added. Please see [go/issues/45713](https://github.com/golang/go/issues/45713). When etcd upgrades the golang version to 1.18, then we can add a go.work in the top level directory, and get all the required replace directives included. Then all the go.mod files do not need to define the replace directives any more. If it's accepted, then I can work on this when the golang version is upgraded to 1.18 in etcd. 

On the other hand, I don't think it's good to define lots of replace directives although they are necessary for now. I think we should remove all the replace directives in the go.mod files in all modules in future, so as to remove the build level dependencies on the files on local disk in future. If this is true and accepted, then we do not need the go.work in the future. 

I would like to get more feedback on this topic.  

cc @mitake

ptabor:
Interesting. But there is also a lot of discussion about not submitting the `go.work` file into repository as an anti-pattern,
that I would assume is what you are proposing. 

https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#preventing-files-from-being-checked-in-to-repositories

serathius:
> Interesting. But there is also a lot of discussion about not submitting the `go.work` file into repository as an anti-pattern, that I would assume is what you are proposing.

Yep it was also surprising for me. I even started working on PR on this as I was tired of cding into every package to test it (don't like to use scripts as I like to add some arguments). However apparently K8s plans to do that? 

cc @thockin seems interested in this.

PR #13994, opened 2022-04-26
Doc: add installtion instructions for tools

spzala:
Add missing install commands.

related #13991


spzala:
cc @serathius @ahrtr @ptabor

tools/etcd-dump-db/README.md
```diff
@@ -1,6 +1,10 @@
 ### etcd-dump-db
 
-etcd-dump-db inspects etcd db files.
+etcd-dump-db inspects etcd db files. Install and use the tool as described below.
+
+```
+  $ go get go.etcd.io/etcd/v3/tools/etcd-dump-db
```
ahrtr, line 6: We need to use `go install `instead of `go get` to install a command. FYI. [go1.17#go-command](https://go.dev/doc/go1.17#go-command)

tools/etcd-dump-logs/README.md
```diff
@@ -1,6 +1,10 @@
 ### etcd-dump-logs
 
-etcd-dump-logs dumps the log from data directory.
+etcd-dump-logs dumps the log from data directory. Install and use the tool as described below.
+
+```
+  $ go get go.etcd.io/etcd/v3/tools/etcd-dump-logs
```
ahrtr, line 6: same as above.

tools/etcd-dump-db/README.md
spzala: @ahrtr thanks, makes sense. I am doing changing it for current instruction in the `benchmark` tool as well. Eventually we should separately update current usage of go get in code.

ahrtr (approved):
LGTM

serathius:
My knowledge might be outdated, but last time I was installing tools like this on fresh mashing there was an issue with PATH. I think that `go install` puts binaries in `$GOPATH/bin`, which is not guaranteed to be included in PATH variable.

serathius:
> It should be developers' responsibility to configure the PATH environment.

I know, but we will still get tens of issues from people reporting that they run the commands as listed and it didn't work. Could we leave some hint/link to mitigate this?

ahrtr:
> > It should be developers' responsibility to configure the PATH environment.
> 
> I know, but we will still get tens of issues from people reporting that they run the commands as listed and it didn't work. Could we leave some hint/link to mitigate this?

Agreed. We can add a hint like below and add link to [go install command](https://pkg.go.dev/cmd/go#hdr-Compile_and_install_packages_and_dependencies).

```
Executables are installed in the directory named by the GOBIN environment variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH environment variable is not set.
```

ptabor:
We can give more actionable instruction:

```
$go list -f "{{.Target}}" ./tools/etcd-dump-db

/Users/ptab/go/bin/etcd-dump-db
```

With etcd directory customer can call: 

```
$ go run ./tools/etcd-dump-db
etcd-dump-db inspects etcd db files.

Usage: ...
```

I was hoping also for solution that does not require checking out etcd sources, like:

```
$ go {install/run}  go.etcd.io/etcd/v3/tools/etcd-dump-db@v3.5.4
go install: go.etcd.io/etcd/v3/tools/etcd-dump-db@v3.5.4 (in go.etcd.io/etcd/v3@v3.5.4):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```
but it seems to be broken since golang 1.18 (https://github.com/golang/go/issues/44840#issuecomment-1098144248). 
We might hope 1.19 will fix it, or we might need to move 'tools' to a separate module without 'replace' directives (or get rid of the 'replace' hack and implement proper validation as part of ./test.sh).

ahrtr:
> but it seems to be broken since golang 1.18 ([golang/go#44840 (comment)](https://github.com/golang/go/issues/44840#issuecomment-1098144248)).

It's broken in golang 1.17 as well. I am thinking probably we can introduce `go work ` (which is a new feature introduced i golang 1.18) in future, so as to remove the usage of `replace`. I raised an [issue](https://github.com/etcd-io/etcd/issues/13478) last year, but got closed due to stale.

ahrtr (approved):

spzala:
Thanks for the great feedback @ptabor @ahrtr @serathius Addressed comments as discussed and also added some additional overall improvements.

serathius (approved):

ptabor (approved):

Issue #14003, opened 2022-05-01
WARNING: DATA RACE on TestWatchOverlapDropConnContextCancel


ahrtr:
```

WARNING: DATA RACE
Write at 0x00c0000e0d28 by goroutine 461:
  go.etcd.io/etcd/tests/v3/framework/integration.(*Member).Close()
      /home/runner/work/etcd/etcd/tests/framework/integration/cluster.go:1131 +0x98
  go.etcd.io/etcd/tests/v3/framework/integration.(*Member).Terminate()
      /home/runner/work/etcd/etcd/tests/framework/integration/cluster.go:1255 +0x5e4
  go.etcd.io/etcd/tests/v3/framework/integration.(*Cluster).Terminate.func1()
      /home/runner/work/etcd/etcd/tests/framework/integration/cluster.go:1387 +0xb7
  go.etcd.io/etcd/tests/v3/framework/integration.(*Cluster).Terminate·dwrap·9()
      /home/runner/work/etcd/etcd/tests/framework/integration/cluster.go:1388 +0x47

Previous read at 0x00c0000e0d28 by goroutine 718:
  go.etcd.io/etcd/tests/v3/framework/integration.(*Member).Bridge()
      /home/runner/work/etcd/etcd/tests/framework/integration/cluster.go:808 +0xc4
  go.etcd.io/etcd/tests/v3/integration/clientv3.TestWatchOverlapDropConnContextCancel.func1()
      /home/runner/work/etcd/etcd/tests/integration/clientv3/watch_test.go:1049 +0x27
  go.etcd.io/etcd/tests/v3/integration/clientv3.testWatchOverlapContextCancel.func1()
      /home/runner/work/etcd/etcd/tests/integration/clientv3/watch_test.go:1082 +0x2be

Goroutine 461 (running) created at:
  go.etcd.io/etcd/tests/v3/framework/integration.(*Cluster).Terminate()
      /home/runner/work/etcd/etcd/tests/framework/integration/cluster.go:1385 +0x30c
  go.etcd.io/etcd/tests/v3/integration/clientv3.testWatchOverlapContextCancel·dwrap·93()
      /home/runner/work/etcd/etcd/tests/integration/clientv3/watch_test.go:1057 +0x58
  runtime.deferCallSave()
      /opt/hostedtoolcache/go/1.17.8/x64/src/runtime/panic.go:950 +0x81
  testing.(*common).Fatalf()
[...]

PR #14004, opened 2022-05-01
Fix data race in testWatchOverlapContextCancel #14003

eval-exec:
This PR fix data race in `testWatchOverlapContextCancel()`

See #14003

### DIFF
--- a/tests/integration/clientv3/watch_test.go
+++ b/tests/integration/clientv3/watch_test.go
@@ -1072,6 +1072,8 @@ func testWatchOverlapContextCancel(t *testing.T, f func(*integration2.Cluster)) 
 		t.Fatal(err)
 	}
 	ch := make(chan struct{}, n)
+	tCtx, cancelFunc := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


ahrtr (approved):
LGTM, thank you.

I believe this isn't the only place which has such similar issue. I am thinking probably we can review the repo, especially the test cases, and provide a better generic solution.

eval-exec:
> LGTM, thank you.
> 
> I believe this isn't the only place which has such similar issue. I am thinking probably we can review the repo, especially the test cases, and provide a better generic solution. 

I would like to work on reviewing etcd's tests cases.

spzala (approved):
lgtm Thanks @eval-exec

spzala:
> > LGTM, thank you.
> > I believe this isn't the only place which has such similar issue. I am thinking probably we can review the repo, especially the test cases, and provide a better generic solution.
> 
> I would like to work on reviewing etcd's tests cases.

That will be great. Please review, and share your findings/open issues/PR as needed for team's review. Thank you!

PR #14007, opened 2022-05-03
change ARCH to GOARCH in build.yaml

ahrtr:
`ARCH` isn't correct, instead we should use `GOARCH`. The script `scripts/build.sh` doesn't use `ARCH` at all.


serathius:
Yep, this GOARCH was replaced by ARCH unintentionally in https://github.com/etcd-io/etcd/commit/4c5efc970df6c0ebd37e11311afaf8a4391e105f.

serathius (approved):

Issue #13978, opened 2022-04-24
CodeQL is broken (again)


serathius:
Example: https://github.com/etcd-io/etcd/runs/6146418414?check_suite_focus=true
Previous issue: https://github.com/etcd-io/etcd/issues/13588
```
Run github/codeql-action/init@v1
  with:
    languages: go
    token: ***
    matrix: {
    "language": "go"
  }
    setup-python-dependencies: true
    debug: false
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/github/codeql-action/v1/node_modules/@octokit/request/dist-node/index.js:66:[2](https://github.com/etcd-io/etcd/runs/6141920349?check_suite_focus=true#step:3:2)[3](https://github.com/etcd-io/etcd/runs/6141920349?check_suite_focus=true#step:3:3)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Job.doExecute (/home/runner/work/_actions/github/codeql-action/v1/node_modules/bottleneck/light.js:[4](https://github.com/etcd-io/etcd/runs/6141920349?check_suite_focus=true#step:3:4)0[5](https://github.com/etcd-io/etcd/runs/6141920349?check_suite_focus=true#step:3:5):18) {
  status: 403,
  headers: {
    'access-control-allow-origin': '*',
    'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',
    connection: 'close',
    'content-encoding': 'gzip',
    'content-security-policy': "default-src 'none'",
    'content-type': 'application/json; charset=utf-8',
    date: 'Sat, 23 Apr 2022 18:5[6](https://github.com/etcd-io/etcd/runs/6141920349?check_suite_focus=true#step:3:6):3[7](https://github.com/etcd-io/etcd/runs/6141920349?check_suite_focus=true#step:3:7) GMT',
    'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
    server: 'GitHub.com',
    'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
[...]

PR #14010, opened 2022-05-05
github: Add necessery permissions for CodeQL

serathius:
cc @ptabor @ahrtr @spzala 
Fixes https://github.com/etcd-io/etcd/issues/13978
Ref https://github.com/github/codeql/issues/8843

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -24,6 +24,10 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
 
     strategy:
       fail-fast: false


ahrtr (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -24,6 +24,10 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
```
ahrtr, line 30: nit: `actions` and `contents` are only required for private repository. 

See the `Usage` section in https://github.com/github/codeql-action

serathius: I think permissions should reflect whole set of permissions required, even those that are given by default in certain environments.

Issue #14013, opened 2022-05-05
Asset transparency flow is faling


serathius:
Don't know what it is, just noticed that we have a broken badge on main page https://github.com/etcd-io/etcd.
Maybe someone can take a look.

https://github.com/etcd-io/etcd/actions/workflows/asset-transparency.yaml

cc @spzala @ahrtr @ptabor

ahrtr:
The reason is the service at [https://beta.transparencylog.net/](https://beta.transparencylog.net/) was shutdown on November 9th 2021, see [transparencylog/tl](https://github.com/transparencylog/tl) . It matches the time when we started to see the failure in [workflows/asset-transparency.yaml](https://github.com/etcd-io/etcd/actions/workflows/asset-transparency.yaml).

We have already provided the [SHA256SUMS](https://github.com/etcd-io/etcd/releases/download/v3.5.4/SHA256SUMS), so I think we can just remove the `workflows/asset-transparency.yaml` for now?

ahrtr:
After second thought, I think it's a useful verification because it can automatically verify each asset's content against the digest on each release. 

But current it's a post release verification, probably it makes more sense to perform pre-release verification unless we stop releasing the binaries (Instead we only release source code and container images). 

Proposed actions:
1. Remove the workflows/asset-transparency.yaml, because it is already broken and there is no immediate replacement for now.
2. Enhance the release script to perform a post-release & pre-release verification. "post-release" means the assests have already been uploaded and the SHA256SUMS has already been generated. "pre-release" means before releasing it on the github.

serathius:
sounds good

PR #14014, opened 2022-05-05
Remove the workflow/transparency.yaml

ahrtr:
First step to fix [14013](https://github.com/etcd-io/etcd/issues/14013)

The service at https://beta.transparencylog.net/ has already been shutdown on November 9th 2021, and we start to see failure after that. So there is no reason to keep the file, but we will continue to find an alternative solution.


serathius:
Please remove badge in [README.](https://github.com/etcd-io/etcd)

ahrtr:
> Please remove badge in https://github.com/etcd-io/etcd/blob/main/README.md?plain=1#L6

Thanks & Done.

serathius (approved):

Issue #13915, opened 2022-04-09
Make leases linearizable (excluding TTL)


serathius:
As etcd leases depend on real time, they cannot provide same guarantees as rest of the API. In distributed systems we cannot depend on local time of single member, so etcd mitigates this by relying on cluster leader to decide what it lease TTL and whether it expired. 

There is however there are two major issues related to current implementation of leases in etcd:  
* Members proxy requests to leader, which cannot be reliably done outside of raft. Raft provides guarantee about replication of the log, however is not expected to provide accurate information about current leader.
* API returning lease doesn't exist based on real time (TTL <= 0), before it's removal is negotiated by raft. This leads to situation where leader returns that lease doesn't exist even though it's in a middle refreshing it. for example https://github.com/etcd-io/etcd/issues/13675.

First issue means that methods that read/update lease TTL, are not guaranteed to provide accurate results/succeed. However, fixing them will require larger redesign of leases. We should revisit this in future releases.

Second issue means that leases might be reported as deleted, even though keys using them will still exist and they can be stil renewed. For v3.5 we should look into fixing this issue.

Proposal: Whether lease exist should depend on raft, making it linearizable. 
With this change we should be able to address https://github.com/etcd-io/etcd/issues/13675, and improve reliability of lease API. 

Changes:
* Leases should not be treated as deleted when TTL hits 0.
* Listing leases (LeaseLeases) method should be linealizable.

This change brings one issue that we should be discussed. Clients might not be correctly handling case where lease has ttl=0. For example client could be using ttl as a delay to wait for lease to be revoked. We should double check client code to verify if this change will have an impact and consider having some minimal TTL that is returned (for example 1 second).

cc @ptabor @ahrtr @endocrimes @spzala

ahrtr:
I think we may have a short-term a long-term plan,

Short-term plan:
The PR [pull/13690](https://github.com/etcd-io/etcd/pull/13690) follows the same way as [previous one ](https://github.com/etcd-io/etcd/pull/7015), it should can resolve the [issue](https://github.com/etcd-io/etcd/issues/13675, at least greatly mitigate it. Regarding to `ListLease`, current implementation is serializable request, we can keep it as it's for now and address it in 3.6, because it's an enhancement.

So in short the short-term plan is merge [pull/13690](https://github.com/etcd-io/etcd/pull/13690) into main and backport to 3.5.

Long-term plan:
[...]

PR #14020, opened 2022-05-08
migrate e2e & integration role_test to common

### LINKED ISSUE #13637: Unify testing framework

chaochn47:
Partially fix #13637


tests/e2e/ctl_v3_role_test.go
```diff
@@ -21,17 +21,8 @@ import (
...
 )
 
-func TestCtlV3RoleAdd(t *testing.T)      { testCtl(t, roleAddTest) }
-func TestCtlV3RootRoleGet(t *testing.T)  { testCtl(t, rootRoleGetTest) }
-func TestCtlV3RoleAddNoTLS(t *testing.T) { testCtl(t, roleAddTest, withCfg(*e2e.NewConfigNoTLS())) }
-func TestCtlV3RoleAddClientTLS(t *testing.T) {
-	testCtl(t, roleAddTest, withCfg(*e2e.NewConfigClientTLS()))
-}
-func TestCtlV3RoleAddPeerTLS(t *testing.T) { testCtl(t, roleAddTest, withCfg(*e2e.NewConfigPeerTLS())) }
 func TestCtlV3RoleAddTimeout(t *testing.T) { testCtl(t, roleAddTest, withDialTimeout(0)) }
```
chaochn47, line 24: I don't think we can easily make the client dial-timeout configurable both in e2e and integration tests. So I leave it alone.

tests/e2e/ctl_v3_role_test.go
```diff
@@ -21,17 +21,8 @@ import (
...
 
-func TestCtlV3RoleGrant(t *testing.T) { testCtl(t, roleGrantTest) }
-
 func roleAddTest(cx ctlCtx) {
```
chaochn47, line 26: Called from `TestCtlV3RoleAddTimeout`

tests/e2e/ctl_v3_role_test.go
```diff
@@ -58,94 +49,6 @@ func roleAddTest(cx ctlCtx) {
...
-	}
-}
-
-func ctlV3RoleMultiExpect(cx ctlCtx, args []string, expStr ...string) error {
-	cmdArgs := append(cx.PrefixArgs(), "role")
-	cmdArgs = append(cmdArgs, args...)
-
-	return e2e.SpawnWithExpects(cmdArgs, cx.envMap, expStr...)
-}
 func ctlV3Role(cx ctlCtx, args []string, expStr string) error {
```
chaochn47, line 52: Line 52 and below are retained only in this PR. After auth_test are migrated to common, they should be cleaned up.

tests/e2e/ctl_v3_role_test.go
ahrtr: Please add a comment here.

tests/common/role_test.go
```diff
@@ -0,0 +1,149 @@
...
+			testutils.ExecuteWithTimeout(t, 10*time.Second, func() {
+				_, err := cc.RoleAdd("root")
+				if err != nil {
+					t.Fatalf("want no error, but got (%v)", err)
+				}
+				_, err = cc.RoleAdd("root")
+				if err == nil || !strings.Contains(err.Error(), rpctypes.ErrRoleAlreadyExist.Error()) {
+					t.Fatalf("want (%v) error, but got (%v)", rpctypes.ErrRoleAlreadyExist, err)
+				}
+				_, err = cc.RoleAdd("")
```
serathius, line 70: This looks like a separate test, like `TestRoleError` was merged with other test.

chaochn47:
All comments are addressed. PTAL, thx! @ahrtr @serathius

client/v3/maintenance.go
```diff
@@ -181,13 +181,16 @@ func (m *maintenance) AlarmDisarm(ctx context.Context, am *AlarmMember) (*AlarmR
...
+		header := ar.Header
 		for _, am := range ar.Alarms {
 			dresp, derr := m.AlarmDisarm(ctx, (*AlarmMember)(am))
 			if derr != nil {
 				return nil, toErr(ctx, derr)
 			}
 			ret.Alarms = append(ret.Alarms, dresp.Alarms...)
+			header = dresp.Header
 		}
+		ret.Header = header
```
serathius, line 193:
```suggestion
		ret.Header = ar.Header
		for _, am := range ar.Alarms {
			dresp, derr := m.AlarmDisarm(ctx, (*AlarmMember)(am))
			if derr != nil {
				return nil, toErr(ctx, derr)
			}
			ret.Alarms = append(ret.Alarms, dresp.Alarms...)
			ret.Header = dresp.Header
		}
```

serathius: This looks like a reasonable fix for a bug, however I would want to make sure that it's properly reviewed. Can you send it in separate PR?

tests/common/user_test.go
```diff
@@ -301,7 +302,8 @@ func TestUserDelete(t *testing.T) {
 				if err == nil {
 					t.Fatalf("deleting a non-existent user should fail")
 				}
-				assert.Contains(t, err.Error(), "user name not found")
+				expectErrMsg := "user name not found"
```
serathius, line 305: Looks like unrelated change.

tests/framework/e2e/etcd_spawn.go
```diff
@@ -15,10 +15,27 @@
...
+
 	"go.etcd.io/etcd/pkg/v3/expect"
 	"go.uber.org/zap"
 )
 
 func SpawnCmd(args []string, envVars map[string]string) (*expect.ExpectProcess, error) {
 	return SpawnCmdWithLogger(zap.NewNop(), args, envVars)
 }
+
+// SpawnCmdWithJsonPrinter spawns an etcdctl process with provided command line arguments and environment variables
```
serathius, line 28: This function has strong assumptions about working only for `etcdctl`, this makes both name and location (`etcd_spawn.go`) not match expectations. Please consider making this a helper method for `EtcdctlV3` struct below.

tests/framework/e2e/etcd_spawn.go
```diff
@@ -15,10 +15,27 @@
...
+func SpawnCmdWithJsonPrinter(args []string, envVars map[string]string) (string, error) {
+	args = append(args, "-w", "json")
+	cmd, err := SpawnCmd(args, envVars)
+	if err != nil {
+		return "", err
+	}
+	return cmd.ExpectFunc(func(txt string) bool {
+		// lease response does not have header substring
+		// member add response does not have revision substring
+		return strings.Contains(txt, "header") || strings.Contains(txt, "revision")
```
serathius, line 39: I consider using or here a workaround for tests, however I think more important here is to find out why we have such inconsistency and consider fixing it. Could you give exact examples of `Response` protos in https://github.com/etcd-io/etcd/blob/main/api/etcdserverpb/rpc.proto that are missing some fields?

tests/framework/e2e/etcdctl.go
```diff
@@ -120,8 +121,7 @@ func (ctl *EtcdctlV3) Put(key, value string, opts config.PutOptions) error {
 }
 
 func (ctl *EtcdctlV3) Delete(key string, o config.DeleteOptions) (*clientv3.DeleteResponse, error) {
-	args := ctl.cmdArgs()
-	args = append(args, "del", key, "-w", "json")
+	args := ctl.cmdArgs("del", key)
```
serathius, line 124: Thanks for refactor, however I'm worried that this might make a breaking change for othere when they migrate. Could we move it to separate PR so it can be quickly reviewed and merged?

tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +508,100 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+	}
+	var resp clientv3.AuthRoleAddResponse
+	line, err := cmd.Expect("header")
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal([]byte(line), &resp)
+	return &resp, err
+}
+
```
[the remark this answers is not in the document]
serathius: It's a good idea, however let's try not to mix in to many changes into PR.

tests/framework/integration.go
```diff
@@ -44,8 +44,8 @@ func (e integrationRunner) NewCluster(t testing.TB, cfg config.ClusterConfig) Cl
 	var err error
 	var integrationCfg integration.ClusterConfig
 	integrationCfg.Size = cfg.ClusterSize
-	integrationCfg.ClientTLS, err = tlsInfo(t, cfg.ClientTLS)
 	integrationCfg.QuotaBackendBytes = cfg.QuotaBackendBytes
+	integrationCfg.ClientTLS, err = tlsInfo(t, cfg.ClientTLS)
```
serathius, line 48: Not sure this is related.

tests/framework/interface.go
```diff
@@ -39,6 +39,7 @@ type Member interface {
 	Stop()
 }
 
+// Client is a wrapper on top of "go.etcd.io/etcd/client/v3".Client interface
```
serathius, line 42: Not sure what what `wrapper on top of ... interface` means. I would set that this interface is meant as a subset of `clientv3.Client` interface implemented by common framework, however this is also not true as options are passed differently to be more expressive. 

Overall this is still work in progress, so I would propose that you skip adding documentation in this PR.

tests/framework/e2e/etcdctl.go
chaochn47: Make sense.

tests/framework/e2e/etcd_spawn.go
chaochn47: Yeah, let me raise a issue instead of fixing it in the tests migration PR.

tests/common/user_test.go
chaochn47: Not related.

chaochn47:
Thanks for the review. I will keep the current PR for the minimal changes of `role_test` test migration. 

The issues/bugs exposed by refactoring with etcdctl json printer should be addressed in a separate PR.

tests/framework/e2e/etcd_spawn.go
chaochn47: It's not grpc client implementation missed some fields but client v3. 

Examples:
1. `header` substring is missing because of embedding.
https://github.com/etcd-io/etcd/blob/b7be91f98b7a7a6202ed9e9bc37b692d484585a8/client/v3/lease.go#L71-L75
2. `revision` is not populated in
https://github.com/etcd-io/etcd/blob/b7be91f98b7a7a6202ed9e9bc37b692d484585a8/server/etcdserver/api/v3rpc/member.go#L108-L110

ahrtr: Thanks for raising this. 

For lease, we need some big change on this. There is an [issue](https://github.com/etcd-io/etcd/issues/13915) discussing this.  The problem mentioned here is lease response indeed has the header info, but it isn't included under the "header" section. For example,

```
$ etcdctl lease list -w json | jq
{
  "revision": 1,
  "raft_term": 2,
  "leases": []
}

$ etcdctl  get k1 -w json | jq
{
  "header": {
    "revision": 1,
    "raft_term": 2
  }
}
```

With regard to the missing "revision" in the memer related commands, I think we might need to fix it. Feel free to raise a PR to send out for review. 
```
$ etcdctl member list  -w json | jq
{
  "header": {
    "raft_term": 2
  },
  "members": [
    {
      "ID": 10276657743932975000,
      "name": "default",
      "peerURLs": [
        "http://localhost:2380"
      ],
      "clientURLs": [
        "http://localhost:2379"
      ]
    }
  ]
}
```

chaochn47: > The problem mentioned here is lease response indeed has the header info, but it isn't included under the "header" section

Exactly. I think may be a json tag added to the LeaseLeasesResponse struct should help. It's a relatively small change.

ahrtr: Technically speaking, it's a small change. But it has impact on user experience. So we should be cautious about this.  Let's track this in the issue I mentioned above.

Feel free to fix the mission revision in member related commands for now.

chaochn47: Yeah, 100% agree. I just realized it like kube-apiserver will be impacted by this small change so I dropped the drafted PR.

tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +508,79 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+// SpawnCmdWithJsonPrinter spawns an etcdctl process with provided command line arguments and environment variables
+// with json printer output and capture header line as output.
+func SpawnCmdWithJsonPrinter(args []string, envVars map[string]string) (string, error) {
+	args = append(args, "-w", "json")
+	cmd, err := SpawnCmd(args, envVars)
+	if err != nil {
+		return "", err
+	}
+	return cmd.ExpectFunc(func(txt string) bool {
+		// header substring is missing because of embedding in LeaseLeasesResponse.
```
serathius, line 582: This PR doesn't touch `LeaseLeasesResponse`, let's skip this for now.

tests/framework/e2e/etcdctl.go
```diff
@@ -507,3 +508,79 @@ func (ctl *EtcdctlV3) UserChangePass(user, newPass string) error {
...
+		// revision substring is missing because ClusterServer.header() does not fill revision in the header
+		return strings.Contains(txt, "header") || strings.Contains(txt, "revision")
+	})
+}
```
serathius, line 586:
```suggestion
func (ctl *EtcdctlV3) RoleDelete(role string) (*clientv3.AuthRoleDeleteResponse, error) {
    var resp clientv3.AuthRoleDeleteResponse
    err := ctl.spawnJsonCmd(&resp, "role", "delete",  role)
    return &resp, err
}

func (ctl *EtcdctlV3) spawnJsonCmd(output interface{}, args ...string) error {
	args = append(args, "-w", "json")
	cmd, err := SpawnCmd(append(ctl.cmdArgs(), args...), nil)
	if err != nil {
		return err
	}
	line, err := cmd.Expect("header")
	if err != nil {
		return err
	}
	err = json.Unmarshal([]byte(line), output)
	if err != nil {
		return err
	}
	return nil
}
```

`SpawnCmdWithJsonPrinter` you proposed doesn't work for any cmd, it depends directly on `-w json` parameters available only for some of the etcdctl commands. As so it should not be a function, but a method on EtcdctlV3 struct. 

I also proposed to make one additional step to include json parsing. It make sense that if we provide commands used to force json output, then we should always be able to parse json.

serathius (approved):

PR #14022, opened 2022-05-09
fix(pkg/ioutil):avoid panic in PageWriter.Write() when pageBytes is 0

secsys-go:
pkg/ioutil: add check for pageBytes when creating PageWriter

NewPageWriter() without check could return a PageWriter whose pageBytes
is equal to 0. This is fatal when buffer to write is  longer than defaultBufferBytes, 
and it will crashed in PageWriter.Write() with:
"panic: runtime error: integer divide by zero".
This check will prevent creating PageWriter in NewPageWriter()when pageBytes is 0, 
and return a nil at that time.


pkg/ioutil/pagewriter.go
```diff
@@ -41,6 +41,9 @@ type PageWriter struct {
 // NewPageWriter creates a new PageWriter. pageBytes is the number of bytes
 // to write per page. pageOffset is the starting offset of io.Writer.
 func NewPageWriter(w io.Writer, pageBytes, pageOffset int) *PageWriter {
+	if pageBytes < 1 {
+		return nil
```
ahrtr, line 45: Instead of returning `nil`, I think it makes more sense to add a [verification](https://github.com/etcd-io/etcd/blob/b7be91f98b7a7a6202ed9e9bc37b692d484585a8/client/pkg/verify/verify.go#L68), and panic when needed.

ptabor: The check is so inexpensive here, that it might be a regular assertion as well.

secsys-go: I have turned the made the comparison a regular assertion, and it will panic instead of return nil when triggered

pkg/ioutil/pagewriter_test.go
```diff
@@ -19,6 +19,20 @@ import (
...
+func TestPageWriterPageBytes(t *testing.T) {
+	// PageWriter should not be created with 0 pageBytes,
+	// otherwise, it will crash when writing.
+	pageBytes := 0
+	defaultBufferBytes = 8 * 1024
+
+	cw := &checkPageWriter{pageBytes: pageBytes, t: t}
+	w := NewPageWriter(cw, pageBytes, 0)
+	if w != nil {
+		t.Fatal("PageWriter created with pageBytes=0")
```
ahrtr, line 31: The test should be updated as well.

pkg/ioutil/pagewriter.go
ptabor: Thank you. 
Thanks to @ahrtr we have a regular function for this: 
https://github.com/etcd-io/etcd/blob/29d30ab7b01f70375aa322fdbddb31a5acf7e003/client/pkg/verify/verify.go#L76

please use it.

ahrtr: FYI. 
```
verify.Assert(pageBytes > 0, "pageBytes %d is an invalid value, it must be greater than 0", pageBytes)
```

ahrtr:
Please also rebase this PR.

ahrtr:
@secsys-go  Please update this PR per the comment.

pkg/ioutil/pagewriter.go
secsys-go: Thanks, I'll fix it

ahrtr:
Please squash the commits.

ahrtr:
Again, please squash the commits into just one commit. FYI. [git-squash](https://www.git-tower.com/learn/git/faq/git-squash)

secsys-go:
This is the squashed one:dc559c17979a1823fab7c80478c74b4db28e0f1b

ahrtr (changes requested):
Please resolve the comment and squash the commits (there are still 8 commits).

ahrtr:
Since there is no  response from the author for a long time, so I resolved it in https://github.com/etcd-io/etcd/pull/14452

PR #14024, opened 2022-05-09
[3.4] etcdserver: fix broken APIv2 with long URI

### LINKED ISSUE #14023: APIv2 doesn't support operations with URI length > 4k.
### What happened?

APIv2 doesn't support operations with URI length > 4k.

APIv2 should support URI length > 4k.

Perform CAS operation with APIv2 and value and previous value size > 2k

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.6.0-alpha.0
Git SHA: b7be91f98
Go Version: go1.17.8
Go OS/Arch: linux/amd64

$ etcdctl version
API version: 3.6
```

</details>

<details>

</details>

<details>

```console
# paste output here

# paste output here
```

</details>

### Relevant log output

_No response_

Hexta:
APIv2 doesn't support operations with URI length > 4k.
This could be relatively easy achieved by CAS operations with value size
about 2k.

This is not documented issue caused by using cmux to serve HTTP1
and HTTP2 protocols on the same port (2379 by default).
cmux has hardcoded max request size for HTTP1 matcher
and returns `false` if size is greater.
HTTP1Fast matcher from cmux validates requests only against known
HTTP methods (GET, POST, PUT, DELETE, OPTIONS,..)
and doesn't have such limit.

Fixes #14023

### DIFF
--- a/embed/serve.go
+++ b/embed/serve.go
@@ -132,7 +132,7 @@ func (sctx *serveCtx) serve(
 			Handler:  createAccessController(sctx.lg, s, httpmux),
 			ErrorLog: logger, // do not log user error
 		}
-		httpl := m.Match(cmux.HTTP1())
+		httpl := m.Match(cmux.HTTP1Fast())
 		go func() { errHandler(srvhttp.Serve(httpl)) }()
 
 		sctx.serversC <- &servers{grpc: gs, http: srvhttp}
--- a/tests/e2e/ctl_v2_test.go
+++ b/tests/e2e/ctl_v2_test.go
@@ -54,6 +54,26 @@ func testCtlV2Set(t *testing.T, cfg *etcdProcessClusterConfig, quorum bool) {
 	}
 }
 
+func TestCtlV2SetLongKey(t *testing.T) {
+	os.Setenv("ETCDCTL_API", "2")
+	defer os.Unsetenv("ETCDCTL_API")
+	defer testutil.AfterTest(t)
+
+	cfg := &configNoTLS
+
+	cfg.enableV2 = true
+	epc := setupEtcdctlTest(t, cfg, false)
+	defer func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	}()
+
+	if err := etcdctlSetNoSync(epc, strings.Repeat("X", 5000), "bar"); err != nil {
+		t.Fatalf("failed set (%v)", err)
+	}
+}
+
 func TestCtlV2Mk(t *testing.T)       { testCtlV2Mk(t, &configNoTLS, false) }
 func TestCtlV2MkQuorum(t *testing.T) { testCtlV2Mk(t, &configNoTLS, true) }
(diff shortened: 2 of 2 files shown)


serathius:
FYI V2 API was deprecated for couple of releases and is already removed on main branch, the only remaining things are for serving `/metrics` and `/version` endpoints.


If you would like to continue with a PR, for example to backport it to old releases, please add tests.

Hexta:
@serathius I've changed destination branch of the PR to release-3.4 (for version which we actually use on the production) and added a test.
Pls advise me if I did it in the wrong way.

ahrtr:
@Hexta  Do you see any real issue in your test or just by reviewing the source code? 

Note that the [cmux.HTTP1()](https://github.com/etcd-io/etcd/blob/b7be91f98b7a7a6202ed9e9bc37b692d484585a8/server/embed/serve.go#L145) is just reading the first line in the HTTP header. In the following example, it's just reading the first line `GET /examples/my-awesome-test/ HTTP/1.1`.  Also see [matchers.go#L94](https://github.com/soheilhy/cmux/blob/v0.1.5/matchers.go#L94)

```
GET /examples/my-awesome-test/ HTTP/1.1
Host: code.tutsplus.com
User-Agent: Mozilla/5.0 (Windows; U; Windows NT 6.1; en-US; rv:1.9.1.5) Gecko/20091102 Firefox/3.5.5 (.NET CLR 3.5.30729)
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-us,en;q=0.5
Accept-Encoding: gzip,deflate
Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7
Keep-Alive: 300
Connection: keep-alive
Cookie: PHPSESSID=r2t5uvjq435r4q7ib3vtdjq120
Pragma: no-cache
Cache-Control: no-cache
```

Hexta:
@ahrtr Try to do the same test with URI longer than 4096. 
Yes, I've the real issue on the production.

ahrtr:
> @ahrtr Try to do the same test with URI longer than 4096. Yes, I've the real issue on the production.

Got it. So you run into this issue when there is a long long `prevValue`?  

Just as @serathius pointed out, v2 API is deprecated, and the v2 server & etcdctl v2 commands have already been removed in main branch (3.6). Can you migrate to the v3 API? You have to perform the migration someday, it's just a matter of time.

ptabor:
+1 to migration to v3.

1. v3.4 is rather EOL. We cherry pick there security patches, but not semantical improvements. 
   I'm a little bit concerned that `Http1Fast` might bring unexpected semantical differences. To be on the safe side, it should be flag gated... but than it becomes substantial change in v3.4. 

2. If we decided to merge - we would need to update the changelog file as well for v3.4.

ptabor:
I'm sorry, I don't understand. Do you mean that you application was working correctly over V2 API and not hitting the limites, and the problem started when the traffic becomes forwarded through the gateway ?

Hexta:
@ptabor Exactly.

Issue #13989, opened 2022-04-25
Revision in response header might not reflect data in the body


lavacat:
### What happened?

Related to https://github.com/etcd-io/etcd/pull/13940

In LeaseLeases function, newHeader is called at the end and there is no synchronization.
https://github.com/etcd-io/etcd/blob/main/server/etcdserver/v3_server.go#L399

This can cause a situation when list of leases in the body doesn't include new lease reported by revision in the header.

The same is true for any other function that's using newHeader

### What did you expect to happen?

Expected Revision in the header to be the latest revision among return leases

### How can we reproduce it (as minimally and precisely as possible)?

might be hard to reproduce
try running TestLeaseGrantAndList

### Etcd version (please run commands below)

3.6 latest main branch

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

lavacat:
Not sure there is nice fix for this. If we move newHeader to the start of the function, the problem is reversed.

IMO, revision in the header should be the latest revision in the data

PR #14026, opened 2022-05-09
Fix member response header missing revision

chaochn47:
Fix issue raised in https://github.com/etcd-io/etcd/pull/14020#discussion_r868453327

Before
```
$ etcdctl member list -w json | jq .
{
  "header": {
    "raft_term": 13
  },
  "members": [
    {
      "ID": 10276657743932975000,
      "name": "default",
      "peerURLs": [
        "http://localhost:2380"
      ],
      "clientURLs": [
        "http://localhost:2379"
      ]
    }
  ]
}
```

After
```
$ etcdctl member list -w json | jq .
{
  "header": {
    "revision": 1,
    "raft_term": 14
  },
  "members": [
    {
      "ID": 10276657743932975000,
      "name": "default",
      "peerURLs": [
        "http://localhost:2380"
      ],
      "clientURLs": [
        "http://localhost:2379"
      ]
    }
  ]
}
```

### DIFF
--- a/server/etcdserver/api/v3rpc/member.go
+++ b/server/etcdserver/api/v3rpc/member.go
@@ -106,7 +106,7 @@ func (cs *ClusterServer) MemberPromote(ctx context.Context, r *pb.MemberPromoteR
 }
 
 func (cs *ClusterServer) header() *pb.ResponseHeader {
-	return &pb.ResponseHeader{ClusterId: uint64(cs.cluster.ID()), MemberId: uint64(cs.server.ID()), RaftTerm: cs.server.Term()}
+	return &pb.ResponseHeader{ClusterId: uint64(cs.cluster.ID()), MemberId: uint64(cs.server.ID()), RaftTerm: cs.server.Term(), Revision: cs.server.KV().Rev()}
 }
 
 func membersToProtoMembers(membs []*membership.Member) []*pb.Member {


ahrtr:
The change looks good to me.  

Two additional verification:
1. On client side (etcdctl), there are different printers, please make sure all of them handle this new field correctly. It should be good, just to double confirm. 
2. Please verify whether this change has any impact on the K8s.

server/etcdserver/api/v3rpc/member.go
```diff
@@ -106,7 +106,7 @@ func (cs *ClusterServer) MemberPromote(ctx context.Context, r *pb.MemberPromoteR
 }
 
 func (cs *ClusterServer) header() *pb.ResponseHeader {
-	return &pb.ResponseHeader{ClusterId: uint64(cs.cluster.ID()), MemberId: uint64(cs.server.ID()), RaftTerm: cs.server.Term()}
+	return &pb.ResponseHeader{ClusterId: uint64(cs.cluster.ID()), MemberId: uint64(cs.server.ID()), RaftTerm: cs.server.Term(), Revision: cs.server.KV().Rev()}
```
serathius, line 109: This goes into recent discussion about meaning of revision in lease response requests that was raised during community meeting https://github.com/etcd-io/etcd/issues/13989. Please remember that revision only relates to key value store, even though revision field is present in all etcd rpc responses. Current situation is not satisfactory as usage of revision is inconsistent and not properly documented.

Before we merge this PR we should make a decision of meaning of revision in all rpc responses and document it. We should look into how revision is used currently in APIs, propose changes and analyse impact.

serathius: I think you mistook `revision` for `raft index`. It's raft index that relates to all data which needs raft concensus. Revision is global version number used for key value store. Etcd implements MVCC by using btree with revision as key. 

If you see this if you look how bbolt is used, we have multiple `buckets` used to stored different values:
* key values are stored under `key` bucket with keys being revisions
* leases are stored under `lease bucket with keys being revision id
* membership info is stored under two buckets `members` and `members_removed` both use member id as key

ahrtr: Yes, you are right.  The key for the lease is lease ID. This could be the reason why member related commands do not have revision in the header. 

But all responses share the same header struct, so we might need to hide the revision field if it's 0. For example [printer_fields.go#L39](https://github.com/etcd-io/etcd/blob/main/etcdctl/ctlv3/command/printer_fields.go#L39)

ptabor: +1 to hiding in etcdctl and not exposing revision for these methods.

chaochn47: Okay. The missing revision behavior sounds like intended not to reflect unrelated mvcc global revision.

PR #14033, opened 2022-05-11
Hide the revision field when it isn't populated

ahrtr:
Refer to [discussion_r869007119](https://github.com/etcd-io/etcd/pull/14026#discussion_r869007119) and [discussion_r869044404](https://github.com/etcd-io/etcd/pull/14026#discussion_r869044404)

Before the change: 
```
$ etcdctl  member list -w fields
"ClusterID" : 14841639068965178418
"MemberID" : 10276657743932975437
"Revision" : 0
"RaftTerm" : 2
"ID" : 10276657743932975437
"Name" : "default"
"PeerURL" : "http://localhost:2380"
"ClientURL" : "http://localhost:2379"
"IsLearner" : false
```

After the change:
```
$ ./etcdctl  member list  -w fields
"ClusterID" : 14841639068965178418
"MemberID" : 10276657743932975437
"RaftTerm" : 2
"ID" : 10276657743932975437
"Name" : "default"
"PeerURL" : "http://localhost:2380"
"ClientURL" : "http://localhost:2379"
"IsLearner" : false
```

### DIFF
--- a/api/etcdserverpb/rpc.pb.go
+++ b/api/etcdserverpb/rpc.pb.go
@@ -275,7 +275,8 @@ type ResponseHeader struct {
 	ClusterId uint64 `protobuf:"varint,1,opt,name=cluster_id,json=clusterId,proto3" json:"cluster_id,omitempty"`
 	// member_id is the ID of the member which sent the response.
 	MemberId uint64 `protobuf:"varint,2,opt,name=member_id,json=memberId,proto3" json:"member_id,omitempty"`
-	// revision is the key-value store revision when the request was applied.
+	// revision is the key-value store revision when the request was applied, and it's
+	// unset (so 0) in case of calls not interacting with key-value store.
 	// For watch progress responses, the header.revision indicates progress. All future events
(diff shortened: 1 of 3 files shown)


serathius:
I think it would be better to do a holistic analysis of how different rps return revision and cli prints them to plan changes, instead of jumping on and fixing one thing.

ptabor (approved):

ptabor:
Please consider updating comments of: 

https://github.com/etcd-io/etcd/blob/153824be58dba0398999aec8a6aeb584d5af5a88/api/etcdserverpb/rpc.proto#L402

to stated that the revision is unset (so 0) in case of calls not interacting with the key-value store.

ahrtr:
> Please consider updating comments of:
> 
> https://github.com/etcd-io/etcd/blob/153824be58dba0398999aec8a6aeb584d5af5a88/api/etcdserverpb/rpc.proto#L402
> 
> to stated that the revision is unset (so 0) in case of calls not interacting with the key-value store.

thx & Done.

PR #14038, opened 2022-05-13
tests: Extract cluster test cases

serathius:
cc @ptabor @ahrtr

### DIFF
--- a/tests/common/kv_test.go
+++ b/tests/common/kv_test.go
@@ -26,32 +26,7 @@ import (
 
 func TestKVPut(t *testing.T) {
 	testRunner.BeforeTest(t)
-	tcs := []struct {
-		name   string
-		config config.ClusterConfig
-	}{
-		{
-			name:   "NoTLS",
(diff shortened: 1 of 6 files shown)


tests/common/integration_test.go
```diff
@@ -23,4 +23,10 @@ import (
 
 func init() {
 	testRunner = framework.IntegrationTestRunner
+	clusterTestCases = []testCase{
+		{
+			name:   "NoTLS",
+			config: config.ClusterConfig{ClusterSize: 1},
+		},
+	}
```
ahrtr, line 31: Does it mean you remove some cases for integration test?

serathius: Yes, maybe it was worth to mention it more.
With e2e and integration having 1to1 coverage of same tests, we can speed up tests while maintaining same coverage. 
I think this needs a broader discussion, so I will move it to separate PR.

ahrtr (approved):

PR #14041, opened 2022-05-13
migrate auth_test to common

chaochn47:
Related to #13637

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202764322183583


chaochn47:
@serathius Can I get some early feebacks if the common test framework auth implementation makes sense to you? Thanks!

ahrtr:
cc @mitake

tests/framework/e2e/etcdctl.go
```diff
@@ -39,6 +41,12 @@ func NewEtcdctl(cfg *EtcdProcessClusterConfig, endpoints []string) *EtcdctlV3 {
 	}
 }
 
+func (ctl *EtcdctlV3) WithAuth(username, password string) *EtcdctlV3 {
```
serathius, line 44: This doesn't return new client, but mutates original client and returns it.

tests/framework/e2e/etcdctl.go
```diff
@@ -309,6 +317,7 @@ func (ctl *EtcdctlV3) flags() map[string]string {
 		}
 	}
 	fmap["endpoints"] = strings.Join(ctl.endpoints, ",")
+	fmap["user"] = ctl.userName + ":" + ctl.password
```
serathius, line 320: This should be optional if userName and password is not set?

tests/framework/integration.go
```diff
@@ -380,3 +405,23 @@ func getOps(ss []string) ([]clientv3.Op, error) {
 	}
 	return ops, nil
 }
+
+type integrationAuthClient struct {
```
serathius, line 409: Please don't do anything that doesn't make sense in standard client. If `ClusterClientWithAuth` needs to return erorr, then `ClientWithAuth` should also return erorr.

tests/framework/integration/cluster.go
```diff
@@ -179,6 +179,8 @@ type Cluster struct {
 
 	mu            sync.Mutex
 	clusterClient *clientv3.Client
+
+	clusterAuthClients []*clientv3.Client
```
serathius, line 183: There is already a lot of different types of clients stored just to call `.Close()`, maybe we should just store list destructor functions? Even better would be to use `t.Cleanup(func())` for call client.Close.

tests/framework/interface.go
```diff
@@ -30,6 +30,7 @@ type testRunner interface {
 type Cluster interface {
 	Members() []Member
 	Client() Client
+	ClientWithAuth(userName, password string) Client
```
serathius, line 33: I think we should add Options argument to `Client` instead.

tests/framework/integration.go
chaochn47: > Please don't do anything that doesn't make sense in standard client.

Some auth tests checks unauthenticated errors. 

Integration test will throw an error at client creation stage with `clientv3.New` https://github.com/etcd-io/etcd/blob/7ec336f80c9463fecd6d8ef15a9f05f3e8f6287a/client/v3/client.go#L439

However, e2e test won't fail at client creation stage and fail at "got response" stage. That's why I defer the error from client creation to "got response"stage in integration test. 

Also every request from e2e test will establish a new grpc connection to the server but integration test reuses existing one. In order to use a different user name and password, integration test needs to create another client.

tests/framework/integration/cluster.go
chaochn47: > maybe we should just store list destructor functions

I think that's a good idea. Maybe we could optimize it in a follow-up PR?

tests/framework/interface.go
chaochn47: I agree. To not break other in-flight tests migration PR, can we defer to a follow-up PR?

tests/framework/e2e/etcdctl.go
chaochn47: It follows a `NewEtcdctlV3`. So I think it should be fine.

serathius: > It follows a `NewEtcdctlV3`. So I think it should be fine.

What do you mean?

tests/framework/integration.go
serathius: What you describe makes sense, however I would prefer different solution to avoid over engineering. It's true that `clientv3.New` used in integration tests will return error if auth is invalid. This is because it calls `Auth.Authenticate` https://github.com/etcd-io/etcd/blob/7ec336f80c9463fecd6d8ef15a9f05f3e8f6287a/client/v3/client.go#L259-L276 when creating client.

Could e2e tests also call `Auth.Authenticate` when creating client with username/password provided?

tests/framework/integration/cluster.go
serathius: I mean this PR could just avoid introducing `clusterAuthClients []*clientv3.Client`, by using `cleanup []func()`. We can migrate else in future.

tests/framework/interface.go
serathius: I think we can do a minimal PR targeted just to fix it it should be ok. Conflicts will still happen, let's just make sure PRs will be quickly merged and conflics are trivial to resolve.

chaochn47: SGTM

tests/framework/integration.go
chaochn47:
```
dev-dsk-chaochn-2c-a26acd76 % etcdctl auth
NAME:
	auth - Enable or disable authentication

USAGE:
	etcdctl auth <enable or disable> [flags]

API VERSION:
	3.6


COMMANDS:
	disable	Disables authentication
	enable	Enables authentication
	status	Returns authentication status

OPTIONS:
  -h, --help[=false]	help for auth

GLOBAL OPTIONS:
      --cacert=""				verify certificates of TLS-enabled secure servers using this CA bundle
      --cert=""					identify secure client using this TLS certificate file
      --command-timeout=5s			timeout for short running command (excluding dial timeout)
      --debug[=false]				enable client-side debug logging
      --dial-timeout=2s				dial timeout for client connections
  -d, --discovery-srv=""			domain name to query for SRV records describing cluster endpoints
      --discovery-srv-name=""			service name to query when using DNS discovery
      --endpoints=[127.0.0.1:2379]		gRPC endpoints
      --hex[=false]				print byte strings as hex encoded strings
      --insecure-discovery[=true]		accept insecure SRV records describing cluster endpoints
      --insecure-skip-tls-verify[=false]	skip server certificate verification (CAUTION: this option should be enabled only for testing purposes)
      --insecure-transport[=true]		disable transport security for client connections
      --keepalive-time=2s			keepalive time for client connections
      --keepalive-timeout=6s			keepalive timeout for client connections
      --key=""					identify secure client using this TLS key file
      --password=""				password for authentication (if this option is used, --user option shouldn't include password)
      --user=""					username[:password] for authentication (prompt if password is not supplied)
  -w, --write-out="simple"			set the output format (fields, json, protobuf, simple, table)
```

tests/framework/integration/cluster.go
chaochn47: okay. Make sense.

tests/framework/e2e/etcdctl.go
chaochn47: I meant `WithAuth` follows `NewEtcdctlV3`. So it should be fine that it is not mutating existing client. 

However, I accept your suggestion to pass a option to `Client()` instead. So this discussion around `WithAuth` can be dropped. 

```
func (c *e2eCluster) Client() Client {
	return e2eClient{e2e.NewEtcdctl(c.Cfg, c.EndpointsV3())}
}

func (c *e2eCluster) ClientWithAuth(userName, password string) Client {
	return e2eClient{e2e.NewEtcdctl(c.Cfg, c.EndpointsV3()).WithAuth(userName, password)}
}
```

tests/framework/integration.go
serathius: I think we can create `clientv3.Client` just for this.

chaochn47:
```
--- FAIL: TestDowngradeUpgradeClusterOf3 (41.84s)

{"level":"warn","ts":"2022-08-04T07:33:45.478Z","caller":"embed/serve.go:120","msg":"stopping grpc server due to error","error":"accept tcp 127.0.0.1:20000: use of closed network connection"}
```

ref. https://github.com/etcd-io/etcd/runs/7667198807?check_suite_focus=true

Looks like similar to https://github.com/etcd-io/etcd/issues/14275

ahrtr:
Thanks @chaochn47 for the PR. More than 2K line of code change is too big, could you try to break this PR into small ones?

server/auth/store.go
```diff
@@ -832,6 +832,8 @@ func (as *authStore) RoleGrantPermission(r *pb.AuthRoleGrantPermissionRequest) (
 		"granted/updated a permission to a user",
 		zap.String("user-name", r.Name),
 		zap.String("permission-name", authpb.Permission_Type_name[int32(r.Perm.PermType)]),
+		zap.ByteString("key", r.Perm.Key),
+		zap.ByteString("range-end", r.Perm.RangeEnd),
```
ahrtr, line 836: Please resolve this in a separate PR so as to make this one as simple & clean as possible.

tests/common/alarm_test.go
```diff
@@ -32,17 +32,18 @@ func TestAlarm(t *testing.T) {
 	defer cancel()
 	clus := testRunner.NewCluster(ctx, t, config.ClusterConfig{ClusterSize: 1, QuotaBackendBytes: int64(13 * os.Getpagesize())})
 	defer clus.Close()
+	cc := clus.MustClient(config.ClientOption{})
 	testutils.ExecuteUntil(ctx, t, func() {
 		// test small put still works
 		smallbuf := strings.Repeat("a", 64)
-		if err := clus.Client().Put("1st_test", smallbuf, config.PutOptions{}); err != nil {
+		if err := cc.Put("1st_test", smallbuf, config.PutOptions{}); err != nil {
```
ahrtr, line 39: Can you resolve this in a separate PR so as to make this one as simple & clean as possible?

ahrtr:
Sorry for the late response. Are you still working on this PR?

chaochn47:
Thanks @ahrtr for the reminder 

Yeah, I am working on the second common test framework changes mentioned above. https://github.com/etcd-io/etcd/pull/14331 but a little confused how to address the last comment raised by Merek. Will keep the PR updated.

chaochn47:
The latest code is on top of https://github.com/etcd-io/etcd/pull/14626, once it is merged, rebase will make the code diff smaller.

chaochn47:
Impacted by a flaky test case

```
=== FAIL: integration TestPeriodicCheckDetectsCorruption
    corrupt_test.go:100: 
        	Error Trace:	corrupt_test.go:100
        	Error:      	Not equal: 
        	            	expected: []*etcdserverpb.AlarmMember{(*etcdserverpb.AlarmMember)(0xc00bb8f9b0)}
        	            	actual  : []*etcdserverpb.AlarmMember{(*etcdserverpb.AlarmMember)(0xc00bb8f950)}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -2,3 +2,3 @@
        	            	  (*etcdserverpb.AlarmMember)({
        	            	-  MemberID: (uint64) 12970263741109204454,
        	            	+  MemberID: (uint64) 11662817823488512165,
        	            	   Alarm: (etcdserverpb.AlarmType) 2,
        	Test:       	TestPeriodicCheckDetectsCorruption
```

fuweid:
> Impacted by a flaky test case
> 
> 
> 
> ```
> 
> === FAIL: integration TestPeriodicCheckDetectsCorruption
> 
>     corrupt_test.go:100: 
> 
>         	Error Trace:	corrupt_test.go:100
> 
>         	Error:      	Not equal: 
> 
>         	            	expected: []*etcdserverpb.AlarmMember{(*etcdserverpb.AlarmMember)(0xc00bb8f9b0)}
> 
>         	            	actual  : []*etcdserverpb.AlarmMember{(*etcdserverpb.AlarmMember)(0xc00bb8f950)}
> 
>         	            	
> 
>         	            	Diff:
> 
>         	            	--- Expected
> 
>         	            	+++ Actual
> 
>         	            	@@ -2,3 +2,3 @@
> 
>         	            	  (*etcdserverpb.AlarmMember)({
> 
>         	            	-  MemberID: (uint64) 12970263741109204454,
> 
>         	            	+  MemberID: (uint64) 11662817823488512165,
> 
>         	            	   Alarm: (etcdserverpb.AlarmType) 2,
> 
>         	Test:       	TestPeriodicCheckDetectsCorruption
> 
> ```

Found that root cause and will file pr later.

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+		t.Fatalf("failed to create test-user %v", err)
+	}
+	_, err = cc.RoleAdd(ctx, roleName)
+	if err != nil {
+		t.Fatalf("failed to create test-role %v", err)
+	}
+	_, err = cc.UserGrantRole(ctx, userName, roleName)
+	if err != nil {
+		t.Fatalf("failed to grant role test-role to user test-user %v", err)
+	}
```
ahrtr, line 1283:
```suggestion
	require.NoError(t, err, "failed to grant role test-role to user test-user %v", err)
```

ahrtr:
This PR is too big. So it is almost impossible to carefully check each line of code change. But overall looks clear and good to me. 

Please try to use `github.com/stretchr/testify` to simplify the code. 

Suggestion for similar PR in future: Please try to migrate the cases step by step. 

I am OK if other maintainers agree to merge this PR.  So defer to @serathius and @spzala to merge or comment this PR.

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+		if len(gresp.Kvs) != 1 || string(gresp.Kvs[0].Key) != "key" || string(gresp.Kvs[0].Value) != "value" {
+			t.Fatalf("want kv pair ('key', 'value') but got %v", gresp.Kvs)
+		}
+	})
+}
+
+func TestAuthLeaseTestTimeToLiveExpired(t *testing.T) {
+	tcs := []struct {
+		name       string
+		JWTEnabled bool
```
serathius, line 685: I don't see this used.

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+				gresp, err := rootAuthClient.Get(ctx, "key", config.GetOptions{})
+				if err != nil || len(gresp.Kvs) != 0 {
+					t.Fatalf("want nil err and no kvs but got (%v) error and %d kvs", err, len(gresp.Kvs))
+				}
+			})
+		})
+	}
+}
+
+func TestAuthLeaseGrantLeases(t *testing.T) {
```
serathius, line 735: Note, there are couple of tests for leases that do not really tests authorization. They are normal lease testing scenarios with only difference that authorization is enabled. 

We should investigate if we can expand standard lease tests and treat authorization as part of cluster setup.

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+				time.Sleep(3 * time.Second)
+				tresp, err := rootAuthClient.TimeToLive(ctx, leaseID, config.LeaseOption{})
+				if err != nil {
+					t.Fatal(err)
+				}
+				if tresp.TTL != -1 {
+					t.Fatalf("want leaseID %v expired but not", leaseID)
+				}
+				gresp, err := rootAuthClient.Get(ctx, "key", config.GetOptions{})
+				if err != nil || len(gresp.Kvs) != 0 {
```
serathius, line 727: Note, no need to fix it in this PR. This checking for empty response will really benefit from rewrite to testify.

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	clus := testRunner.NewCluster(ctx, t, config.WithClusterConfig(config.ClusterConfig{ClusterSize: 1}))
+	defer clus.Close()
+	cc := testutils.MustClient(clus.Client())
+	testutils.ExecuteUntil(ctx, t, func() {
+		authEnable(ctx, cc, t)
+		rootAuthClient := testutils.MustClient(clus.Client(WithAuth("root", "root")))
+		authSetupDefaultTestUser(ctx, rootAuthClient, t)
+		if _, err := rootAuthClient.RoleGet(ctx, "test-role"); err != nil {
```
serathius, line 873: Original test validated here contents of the role, like what keys it can write and what keys it can read.

```
	expected := []string{
		"Role test-role",
		"KV Read:", "foo",
		"KV Write:", "foo",
	}
```

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+	cc := testutils.MustClient(clus.Client())
+	testutils.ExecuteUntil(ctx, t, func() {
+		authEnable(ctx, cc, t)
+		rootAuthClient := testutils.MustClient(clus.Client(WithAuth("root", "root")))
+		authSetupDefaultTestUser(ctx, rootAuthClient, t)
+		resp, err := rootAuthClient.UserGet(ctx, "test-user")
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.ElementsMatch(t, resp.Roles, []string{"test-role"})
```
serathius, line 904: Original test also validated user name in response.

tests/common/auth_test.go
```diff
@@ -0,0 +1,1303 @@
...
+		// test-user can get the information of test-user itself
+		testUserAuthClient := testutils.MustClient(clus.Client(WithAuth("test-user", "pass")))
+		resp, err = testUserAuthClient.UserGet(ctx, "test-user")
+		if err != nil {
+			t.Fatal(err)
+		}
+		assert.ElementsMatch(t, resp.Roles, []string{"test-role"})
```
serathius, line 911: ditto

serathius (commented):
Reviewed and left couple fo comments. I would highly encourage migrating one tests at a time to make review quick. This one review iteration took me over 30 minutes and we will need more iterations.

If you send a separate PR that just migrates one test, I expect that review would take 2 minutes and I good give more detailed feedback.

chaochn47:
Thanks for the review, I will split those test cases into multiple incremental PRs. 

Each PR contains 2 test cases.

tests/common/auth_test.go
chaochn47: switched to utilize `auth_utils.go`.

PR #14042, opened 2022-05-16
tests: Make common framework context aware

serathius:
cc @ahrtr

### DIFF
--- a/tests/framework/testutils/execute.go
+++ b/tests/framework/testutils/execute.go
@@ -15,6 +15,7 @@
 package testutils
 
 import (
+	"context"
 	"fmt"
 	"testing"
 	"time"
@@ -23,6 +24,14 @@ import (
 )
 
 func ExecuteWithTimeout(t *testing.T, timeout time.Duration, f func()) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	ExecuteUntil(ctx, t, f)
+}
+
+func ExecuteUntil(ctx context.Context, t *testing.T, f func()) {
+	deadline, deadlineSet := ctx.Deadline()
+	timeout := time.Until(deadline)
 	donec := make(chan struct{})
 	go func() {
 		defer close(donec)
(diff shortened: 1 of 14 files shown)


tests/framework/testutils/execute.go
```diff
@@ -23,15 +24,26 @@ import (
...
 		f()
 	}()
 
 	select {
-	case <-time.After(timeout):
-		testutil.FatalStack(t, fmt.Sprintf("test timed out after %v", timeout))
+	case <-ctx.Done():
+		msg := ctx.Err().Error()
+		if deadlineSet {
+			msg = fmt.Sprintf("test timed out after %v, err: %v", deadline, msg)
```
ahrtr, line 44: The `deadline` here is a timestamp instead a duration.

ahrtr (approved):
LGTM with a minor comment

tests/framework/testutils/execute.go
serathius: nice catch!

tests/framework/testutils/execute.go
```diff
@@ -23,15 +24,27 @@ import (
...
 		f()
 	}()
 
 	select {
-	case <-time.After(timeout):
-		testutil.FatalStack(t, fmt.Sprintf("test timed out after %v", timeout))
+	case <-ctx.Done():
+		msg := ctx.Err().Error()
+		if deadlineSet {
+			timeout := time.Until(deadline)
```
ahrtr, line 44: This might not be correct. The `timeout` here isn't the original 10 seconds any more. For simplicity, you can just remove it for now.

serathius: It will write `9m59s` or similar, but I think it's still correct. Only issue is that it will not be easy to find the source of timeout, however this is expected when we introduce hierarchical timeouts.

serathius: Ups, wanted to refactor but forgot about `time.Until` needs to be called at the top.

PR #14047, opened 2022-05-16
build script: Allow for over ride of CGO_ENABLED

harisund:
The current build.sh script hard codes the value of CGO_ENABLED to 0
This commit makes it configurable in such a way that it defaults to 0
but can be overriden via an environment variable


ahrtr:
Thanks @harisund for the PR, but we'd prefer to keep it as it's for now.

Issue #14039, opened 2022-05-13
etcd detects data corruption by default


serathius:
On of the highlighted issues in https://github.com/etcd-io/etcd/blob/main/Documentation/postmortems/v3.5-data-inconsistency.md lack of usage and reliability of etcd data consistency check.

Problems:
* Lack of adoption prevents early detection hampering etcd community ability to fix issues. 
* No hard guarantees about detecting inconsistency, undermines purpose of this feature.
* Supportability of v3.5 release is very restricted in case of new data inconsistency issues. 
* Properly fixing data inconsistency check can potentially take multiple releases as it requires a new WAL entry and larger code changes. 

## Proposal
Develop a dedicated improved to data consistency check that will be enabled by default and backported to v3.5 release. Scope of this patch will be heavy limited to avoid introducing any breaking changes or performance regressions. 

I think it's worth to invest into this change now to make sure v3.5 is in acceptable state, instead of trying to rush v3.6 release with half backed state. Due to heavy restrictions it might turn out that there is no reasonable improvement that we can make, still I think it's worth to consider.

### Current state of consistency check
* Initial check  during bootstrap of each member validates that hash for its latest revision matches hashes of all peers. Doesn't check peers that haven't seen the revision or have run compaction during check.
* Periodic check run by leader validates that hash for its latest revision matches hashes of all peers and confirms that no follower has revision newer than leader. Doesn't check slow peers that haven't seen the revision or have run compaction during check.

## Design

Goals
* Consistency check should work even when there are slow cluster members and independent on ongoing compactions.

Requirements:
* No/minimal performance impact when check is enabled. Check needs to be enabled by default, so we should have not performance impact when compared to original check that was disabled by default.
* No WAL and API changes. This will be a patch fix so we cannot make any changes to maintain compatibility.

### Implementation

Calculate and evaluate consistency on compacted revisions. Compactions are negotiated via raft, meaning they are executed on all members on the same revisions, meaning slow members don't matter. During compaction we already need to touch bbolt, meaning calculating hash at the same time should only have minimal performance cost. 

Only issue is that compactions are not done automatically by default, however Kubernetes already runs them every 5 minutes which should be enough to provide significant improvement to majority of etcd users. 

Algorithm:
* During each compaction member calculates hash on data that is being compacted and stores last N hashes to be served by HashKV method.
[...]

PR #14049, opened 2022-05-17
Calculate hash during compaction

serathius:
Implements low cost hash calculation during compaction as proposed in https://github.com/etcd-io/etcd/issues/14039

This PR was deliberately split into smaller commits to avoid any mistakes in code. Recommend to review it one by one.

To make sure that hash calculated by hashKV and scheduleCompaction is the same, I needed to change API so that it's always accompanied by revision range it was calculated on, which is change from previous API that always returned latest revision (WAT?).

I also introduced a intense tests for hash function on a first commit to make sure I will not change result of hashing function. 

This PR is big even though it's meant to be backported to v3.5, this is by design as I don't think we could make a quick hacky change and not make mistake. I included some refactors that should make code clear to understood.

cc @ahrtr @ptabor

### DIFF
--- a/server/etcdserver/api/v3rpc/maintenance.go
+++ b/server/etcdserver/api/v3rpc/maintenance.go
@@ -208,7 +208,7 @@ func (ms *maintenanceServer) HashKV(ctx context.Context, r *pb.HashKVRequest) (*
 		return nil, togRPCError(err)
 	}
 
-	resp := &pb.HashKVResponse{Header: &pb.ResponseHeader{Revision: rev}, Hash: h, CompactRevision: compactRev}
+	resp := &pb.HashKVResponse{Header: &pb.ResponseHeader{Revision: rev}, Hash: h.Hash, CompactRevision: compactRev}
 	ms.hdr.fill(resp.Header)
 	return resp, nil
 }
--- a/server/etcdserver/api/v3rpc/maintenance.go
+++ b/server/etcdserver/api/v3rpc/maintenance.go
@@ -70,20 +70,20 @@ type ClusterStatusGetter interface {
 }
 
 type maintenanceServer struct {
-	lg  *zap.Logger
-	rg  apply.RaftStatusGetter
-	kg  KVGetter
-	bg  BackendGetter
-	a   Alarmer
-	lt  LeaderTransferrer
-	hdr header
-	cs  ClusterStatusGetter
-	d   Downgrader
-	vs  serverversion.Server
+	lg     *zap.Logger
+	rg     apply.RaftStatusGetter
+	hasher mvcc.HashStorage
+	bg     BackendGetter
+	a      Alarmer
+	lt     LeaderTransferrer
+	hdr    header
+	cs     ClusterStatusGetter
+	d      Downgrader
+	vs     serverversion.Server
 }
 
 func NewMaintenanceServer(s *etcdserver.EtcdServer) pb.MaintenanceServer {
-	srv := &maintenanceServer{lg: s.Cfg.Logger, rg: s, kg: s, bg: s, a: s, lt: s, hdr: newHeader(s), cs: s, d: s, vs: etcdserver.NewServerVersionAdapter(s)}
+	srv := &maintenanceServer{lg: s.Cfg.Logger, rg: s, hasher: s.KV().HashStorage(), bg: s, a: s, lt: s, hdr: newHeader(s), cs: s, d: s, vs: etcdserver.NewServerVersionAdapter(s)}
 	if srv.lg == nil {
 		srv.lg = zap.NewNop()
 	}
@@ -193,7 +193,7 @@ func (ms *maintenanceServer) Snapshot(sr *pb.SnapshotRequest, srv pb.Maintenance
 }
 
 func (ms *maintenanceServer) Hash(ctx context.Context, r *pb.HashRequest) (*pb.HashResponse, error) {
-	h, rev, err := ms.kg.KV().Hash()
+	h, rev, err := ms.hasher.Hash()
 	if err != nil {
 		return nil, togRPCError(err)
 	}
@@ -203,7 +203,7 @@ func (ms *maintenanceServer) Hash(ctx context.Context, r *pb.HashRequest) (*pb.H
 }
 
 func (ms *maintenanceServer) HashKV(ctx context.Context, r *pb.HashKVRequest) (*pb.HashKVResponse, error) {
-	h, rev, err := ms.kg.KV().HashByRev(r.Revision)
+	h, rev, err := ms.hasher.HashByRev(r.Revision)
 	if err != nil {
 		return nil, togRPCError(err)
 	}
--- a/server/etcdserver/api/v3rpc/maintenance.go
+++ b/server/etcdserver/api/v3rpc/maintenance.go
@@ -203,12 +203,12 @@ func (ms *maintenanceServer) Hash(ctx context.Context, r *pb.HashRequest) (*pb.H
 }
 
 func (ms *maintenanceServer) HashKV(ctx context.Context, r *pb.HashKVRequest) (*pb.HashKVResponse, error) {
-	h, rev, compactRev, err := ms.kg.KV().HashByRev(r.Revision)
+	h, rev, err := ms.kg.KV().HashByRev(r.Revision)
 	if err != nil {
 		return nil, togRPCError(err)
 	}
 
-	resp := &pb.HashKVResponse{Header: &pb.ResponseHeader{Revision: rev}, Hash: h.Hash, CompactRevision: compactRev}
+	resp := &pb.HashKVResponse{Header: &pb.ResponseHeader{Revision: rev}, Hash: h.Hash, CompactRevision: h.CompactRevision}
 	ms.hdr.fill(resp.Header)
 	return resp, nil
 }
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -39,7 +39,7 @@ type corruptionMonitor struct {
 }
 
 type Hasher interface {
-	mvcc.Hasher
+	mvcc.HashStorage
 	ReqTimeout() time.Duration
 	MemberId() types.ID
 	PeerHashByRev(int64) []*peerHashKVResp
@@ -50,13 +50,13 @@ type Hasher interface {
 func NewCorruptionMonitor(lg *zap.Logger, s *EtcdServer) *corruptionMonitor {
 	return &corruptionMonitor{
 		lg:     lg,
-		hasher: hasherAdapter{s, s.KV()},
+		hasher: hasherAdapter{s, s.KV().HashStorage()},
 	}
 }
 
 type hasherAdapter struct {
 	*EtcdServer
-	mvcc.KV
+	mvcc.HashStorage
 }
 
 func (h hasherAdapter) ReqTimeout() time.Duration {
@@ -345,7 +345,7 @@ func (h *hashKVHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "error unmarshalling request", http.StatusBadRequest)
 		return
 	}
-	hash, rev, err := h.server.KV().HashByRev(req.Revision)
+	hash, rev, err := h.server.KV().HashStorage().HashByRev(req.Revision)
 	if err != nil {
 		h.lg.Warn(
 			"failed to get hashKV",
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -82,7 +82,7 @@ func (cm *corruptionMonitor) InitialCheck() error {
 		zap.Duration("timeout", cm.hasher.ReqTimeout()),
 	)
 
-	h, rev, crev, err := cm.hasher.HashByRev(0)
+	h, rev, err := cm.hasher.HashByRev(0)
 	if err != nil {
 		return fmt.Errorf("%s failed to fetch hash (%v)", cm.hasher.MemberId(), err)
 	}
@@ -94,7 +94,7 @@ func (cm *corruptionMonitor) InitialCheck() error {
 			fields := []zap.Field{
 				zap.String("local-member-id", cm.hasher.MemberId().String()),
 				zap.Int64("local-member-revision", rev),
-				zap.Int64("local-member-compact-revision", crev),
+				zap.Int64("local-member-compact-revision", h.CompactRevision),
 				zap.Uint32("local-member-hash", h.Hash),
 				zap.String("remote-peer-id", peerID.String()),
 				zap.Strings("remote-peer-endpoints", p.eps),
@@ -104,7 +104,7 @@ func (cm *corruptionMonitor) InitialCheck() error {
 			}
 
 			if h.Hash != p.resp.Hash {
-				if crev == p.resp.CompactRevision {
+				if h.CompactRevision == p.resp.CompactRevision {
 					cm.lg.Warn("found different hash values from remote peer", fields...)
 					mismatch++
 				} else {
@@ -122,7 +122,7 @@ func (cm *corruptionMonitor) InitialCheck() error {
 					"cannot fetch hash from slow remote peer",
 					zap.String("local-member-id", cm.hasher.MemberId().String()),
 					zap.Int64("local-member-revision", rev),
-					zap.Int64("local-member-compact-revision", crev),
+					zap.Int64("local-member-compact-revision", h.CompactRevision),
 					zap.Uint32("local-member-hash", h.Hash),
 					zap.String("remote-peer-id", p.id.String()),
 					zap.Strings("remote-peer-endpoints", p.eps),
@@ -133,7 +133,7 @@ func (cm *corruptionMonitor) InitialCheck() error {
 					"cannot fetch hash from remote peer; local member is behind",
 					zap.String("local-member-id", cm.hasher.MemberId().String()),
 					zap.Int64("local-member-revision", rev),
-					zap.Int64("local-member-compact-revision", crev),
+					zap.Int64("local-member-compact-revision", h.CompactRevision),
 					zap.Uint32("local-member-hash", h.Hash),
 					zap.String("remote-peer-id", p.id.String()),
 					zap.Strings("remote-peer-endpoints", p.eps),
@@ -154,7 +154,7 @@ func (cm *corruptionMonitor) InitialCheck() error {
 }
 
 func (cm *corruptionMonitor) periodicCheck() error {
-	h, rev, crev, err := cm.hasher.HashByRev(0)
+	h, rev, err := cm.hasher.HashByRev(0)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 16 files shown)


server/storage/mvcc/kvstore_compaction.go
```diff
@@ -43,24 +44,31 @@ func (s *store) scheduleCompaction(compactMainRev int64, keep map[revision]struc
 
 		tx := s.b.BatchTx()
 		tx.LockOutsideApply()
-		keys, _ := tx.UnsafeRange(schema.Key, last, end, int64(batchNum))
-		for _, key := range keys {
-			rev = bytesToRev(key)
+		keys, values := tx.UnsafeRange(schema.Key, last, end, int64(batchNum))
+		for i, _ := range keys {
```
ahrtr, line 48:
```suggestion
		for i := range keys {
```

server/storage/mvcc/kvstore.go
```diff
@@ -255,7 +232,8 @@ func (s *store) compact(trace *traceutil.Trace, rev int64) (<-chan struct{}, err
 		start := time.Now()
 		keep := s.kvindex.Compact(rev)
 		indexCompactionPauseMs.Observe(float64(time.Since(start) / time.Millisecond))
-		if !s.scheduleCompaction(rev, keep) {
+		_, err := s.scheduleCompaction(rev, keep)
+		if err != nil {
```
ahrtr, line 236:
```suggestion
		if _, err := s.scheduleCompaction(rev, keep); err != nil {
```

ahrtr:
It should have no performance loss after a quick review. Will have a deeper review when the PR is marked as ready for review.

server/storage/mvcc/kvstore.go
serathius: Previously it returned bool, whether compaction succeeded (or was interrupted by stopC). I have changed it to error as it better represents intention.

serathius:
Pushing PR for review. cc @ahrtr @ptabor 
This one only implements calculating hash, I will prepare separate for validating hashes.

server/storage/mvcc/kvstore.go
```diff
@@ -232,7 +232,7 @@ func (s *store) compact(trace *traceutil.Trace, rev int64) (<-chan struct{}, err
 		start := time.Now()
 		keep := s.kvindex.Compact(rev)
 		indexCompactionPauseMs.Observe(float64(time.Since(start) / time.Millisecond))
-		if !s.scheduleCompaction(rev, keep) {
+		if err := s.scheduleCompaction(rev, keep); err != nil {
 			s.compactBarrier(context.TODO(), ch)
```
ptabor, line 228: shell we at least log the error ?

server/storage/mvcc/kvstore_compaction.go
```diff
@@ -73,7 +74,7 @@ func (s *store) scheduleCompaction(compactMainRev int64, keep map[revision]struc
 		select {
 		case <-time.After(batchInterval):
 		case <-s.stopc:
-			return false
+			return fmt.Errorf("stopped")
```
ptabor: "compaction interrupted due to store stop signal" ?

server/storage/mvcc/kvstore_compaction.go
```diff
@@ -48,24 +52,27 @@ func (s *store) scheduleCompaction(compactMainRev int64) error {
 
 		tx := s.b.BatchTx()
 		tx.LockOutsideApply()
-		keys, _ := tx.UnsafeRange(schema.Key, last, end, int64(batchNum))
-		for _, key := range keys {
-			rev = bytesToRev(key)
+		keys, values := tx.UnsafeRange(schema.Key, last, end, int64(batchNum))
```
ptabor, line 55: Please see that `end` is defined as: 
```
binary.BigEndian.PutUint64(end, uint64(compactMainRev+1))
```

I'm afraid this logic would miss all entries created after the most recent compaction... but maybe I'm missing sth...

server/storage/mvcc/hash.go
```diff
@@ -22,7 +22,7 @@ import (
 	"go.etcd.io/etcd/server/v3/storage/schema"
 )
 
-func unsafeHashByRev(tx backend.ReadTx, lower, upper revision, keep map[revision]struct{}) (uint32, error) {
+func unsafeHashByRev(tx backend.ReadTx, lower, upper int64, keep map[revision]struct{}) (uint32, error) {
```
ptabor: I wish revision was strongly typed: 

`type Revision int64` such that it's hard to miss-type between int and Revision.

Probably for another refactoring.

server/storage/mvcc/kvstore_compaction.go
serathius: Done

server/storage/mvcc/kvstore_compaction.go
serathius: `compactMainRev` is revision that is being compacted. This is our goal, calculate hash at the same time as compaction is being done, thus not increasing number of times entry is touched. We don't care about entries after most recent compaction.

server/storage/mvcc/hash.go
serathius: That's the good point. Here I wanted to stop using `revision` type to ensure that only main revision is set. `revision` type allows setting sub revision could lead to hard to predict side effects.

ahrtr:
I will take a deep review once it's rebased to resolve the conflict.

serathius:
Done

server/storage/mvcc/kvstore_compaction.go
```diff
@@ -16,14 +16,23 @@ package mvcc
...
 
 	"go.etcd.io/etcd/server/v3/storage/schema"
 	"go.uber.org/zap"
 )
 
-func (s *store) scheduleCompaction(compactMainRev int64, keep map[revision]struct{}) bool {
+func (s *store) scheduleCompaction(compactMainRev int64) (kvHash, error) {
+	s.revMu.RLock()
+	minHashRev := s.compactMainRev
+	s.revMu.RUnlock()
```
ahrtr, line 29: No need to acquire the read lock, we can use 
```
minHashRev := atomic.LoadInt64(&s.compactMainRev)
```

server/storage/mvcc/hash.go
```diff
@@ -0,0 +1,78 @@
...
+		h.WriteKeyValue(k, v)
+		return nil
+	})
+	return h.Hash(), err
+}
+
+type kvHasher struct {
+	hash            hash.Hash32
+	compactRevision int64
+	revision        int64
```
ahrtr, line 37: rename to `curRevision` for better readability?

server/etcdserver/corrupt.go
```diff
@@ -47,30 +47,30 @@ func (s *EtcdServer) CheckInitialHashKV() error {
 		zap.String("local-member-id", s.MemberId().String()),
 		zap.Duration("timeout", s.Cfg.ReqTimeout()),
 	)
-
-	h, rev, crev, err := s.kv.HashByRev(0)
+	currentRev := s.kv.Rev()
```
ahrtr, line 50: The same as above, it doesn't make much sense to introduce an unnecessary creation & close of a concurrentReadTx.

server/etcdserver/corrupt.go
```diff
@@ -150,11 +150,12 @@ func (s *EtcdServer) monitorKVHash() {
 func (s *EtcdServer) checkHashKV() error {
 	lg := s.Logger()
 
-	h, rev, crev, err := s.kv.HashByRev(0)
+	currentRev := s.kv.Rev()
```
ahrtr, line 153: same as above

server/etcdserver/corrupt.go
```diff
@@ -337,7 +338,12 @@ func (h *hashKVHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "error unmarshalling request", http.StatusBadRequest)
 		return
 	}
-	hash, rev, compactRev, err := h.server.KV().HashByRev(req.Revision)
+	currentRev := h.server.KV().Rev()
```
ahrtr, line 341: same as above

server/etcdserver/api/v3rpc/maintenance.go
```diff
@@ -203,12 +203,16 @@ func (ms *maintenanceServer) Hash(ctx context.Context, r *pb.HashRequest) (*pb.H
 }
 
 func (ms *maintenanceServer) HashKV(ctx context.Context, r *pb.HashKVRequest) (*pb.HashKVResponse, error) {
-	h, rev, compactRev, err := ms.kg.KV().HashByRev(r.Revision)
+	rev := r.Revision
+	if rev == 0 {
+		rev = ms.kg.KV().Rev()
+	}
```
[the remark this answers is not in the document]
serathius: Goal of this change is to simplify `KV.HashByRev` method, by removing this special case that could cause become a source of bugs in future. From KV point of view, it doesn't makes sense that `0` revision has special meaning of latest. This is more of an API than storage behavior. 

I think lower level code should follow rule "require more, promise less". It's ok to a lot of promises at the API level, however it can be disruptive for lower level code. With proposed changes `HashByRev` will be called for other use cases than just API response as so we should avoid special cases as much as possible. 

This is tradeoff on clarity vs performace. For me here clarity is more important, as performance impact seems not very significant. This change would require additional `concurentReadTx` lock grap on every HashByRev request. This request is not called very often as so I think it is acceptable. 

Please let me know if it makes sense or I missed anything.

server/etcdserver/corrupt.go
serathius: This was intentional for code clarity, let's continue discussion in comment above.

server/storage/mvcc/kvstore_compaction.go
serathius: Here I wanted to follow the same locking mechanism used by `HashByRev` which uses `revMu`. From code:
```
	// revMu protects currentRev and compactMainRev.
	// Locked at end of write txn and released after write txn unlock lock.
	// Locked before locking read txn and released after locking
```

serathius: Reverted this code this was incorrect.

server/etcdserver/api/v3rpc/maintenance.go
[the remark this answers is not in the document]
serathius: This change was reimplemented so the comment no longer applies.

server/storage/mvcc/hash.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (s *hashStorage) Store(hash kvHash) {
+	s.lg.Info("storing new hash",
+		zap.Uint32("hash", hash.Hash),
+		zap.Int64("revision", hash.Revision),
+		zap.Int64("compact-revision", hash.CompactRevision),
+	)
+	s.hashMu.Lock()
+	defer s.hashMu.Unlock()
+	s.hashes = append(s.hashes, hash)
```
ahrtr, line 141: What if a member gets restarted? all hashes will get lost? Should we persist the hashes into `db`?

serathius: Fully correct solution would persist hashes into `WAL` entries, however we cannot backport this change. You are right, not persisting hashes would mean that restarts are disruptive, however I prefer to have this flaw than develop a temporary solution. 

Not every mechanism in etcd works well with restarts. Definitely consistency of KV store needs to be maintained independent of restarts, however secondary mechanism like leases will not work if etcd is frequently restarted. It's up to admin to ensure that etcd is not too frequently restarted so leases work well. I think it's reasonable to assume that etcd doesn't restart for non-critical features to work.

For this v3.5 backport I would treat correctness checking as secondary mechanism that requires etcd not to restart and for v3.6 I would look into making the mechanism fully independent.

ahrtr: Will think about this and get back later. 



> Fully correct solution would persist hashes into `WAL` entries

Not sure what do you mean? Do you mean you want to depend on WAL replaying on startup to  recover the `hashes`?

serathius: I mean that we will introduce new WAL entry with Hash. Leader would periodically calculate it hash and send it as a raft proposal. When applying followers would verify if their hash matches one in raft.

ahrtr: It can work, but looks like it doesn't make much sense. The raft is just like a network transport channel in this case instead of a consensus protocol.  The leader just transports the hash to followers. Why not use the existing peer communication mechanism? 

But it might be useful on the solution to implement merkele root?

server/storage/mvcc/hash_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+const (
+	// Use high prime
+	compactionCycle = 71
+)
+
+// Test HashByRevValue values to ensure we don't change the output which would
+// have catastrophic consequences. Expected output is just hardcoded, so please
+// regenerate it every time you change input parameters.
+func TestHashByRevValue(t *testing.T) {
```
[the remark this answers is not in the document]
ahrtr: My suggestion can simplify (instead of complicate) the implementation. Feel free to apply the following patch. You are welcome :) 

[corrupt.txt](https://github.com/etcd-io/etcd/files/8867776/corrupt.txt)

server/storage/mvcc/hash_test.go
```diff
@@ -0,0 +1,239 @@
...
+		{106915399, 603, 674},
+		{406914006, 674, 745},
+		{1882211381, 745, 816},
+		{806177088, 816, 887},
+		{664311366, 887, 958},
+		{1496914449, 958, 1029},
+		{2434525091, 1029, 1100},
+		{3988652253, 1100, 1171},
+		{1122462288, 1171, 1242},
+		{724436716, 1242, 1883},
```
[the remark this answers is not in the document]
ahrtr: OK, SG

server/storage/mvcc/hash.go
```diff
@@ -0,0 +1,148 @@
...
+	h.hash.Write(k)
+	h.hash.Write(v)
+}
+
+func (h *kvHasher) Hash() kvHash {
+	return kvHash{Hash: h.hash.Sum32(), CompactRevision: h.compactRevision, Revision: h.revision}
+}
+
+type kvHash struct {
+	Hash            uint32
```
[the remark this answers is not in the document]
serathius: Yes, this PR is already big and it just implements just calculating hash on compaction, which is technically easier. It can be merge as it has no impact on etcd functionality. Harder or at least more controversial is the part that will check the hash. That's why I will send it as separate PR.

server/storage/mvcc/hash_test.go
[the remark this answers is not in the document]
serathius: You're right. I mistook which function you meant. Your code is of course better :P

server/etcdserver/corrupt.go
```diff
@@ -206,33 +205,33 @@ func (s *EtcdServer) checkHashKV() error {
...
 				zap.Int64("follower-compact-revision", p.resp.CompactRevision),
 				zap.String("follower-peer-id", types.ID(id).String()),
 			)
 			mismatch(id)
 		}
 
 		// follower's compact revision is leader's old one, then hashes must match
-		if p.resp.CompactRevision == crev && p.resp.Hash != h {
+		if p.resp.CompactRevision == h.CompactRevision && p.resp.Hash != h.Hash {
 			lg.Warn(
```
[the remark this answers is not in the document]
serathius: You are right, `corrupt.go` code needs some improvements. However I don't want to do much in this PR.

server/storage/mvcc/hash.go
```diff
@@ -0,0 +1,148 @@
...
+		revision:        rev,
+		keep:            keep,
+	}
+}
+
+func (h *kvHasher) WriteKeyValue(k, v []byte) {
+	kr := bytesToRev(k)
+	upper := revision{main: h.revision + 1}
+	if !upper.GreaterThan(kr) {
+		return
```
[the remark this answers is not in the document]
serathius: This is optimization for hash calculation that should not be backported. We can do a follow up PR.

serathius:
@ahrtr addressed all coments, PTAL

server/storage/mvcc/hash.go
ahrtr: > For immediate hash verification on every entry we would need merkele root allowing us automatic, point-in-time recovery from data corruption.

The historical data has already been compacted, how can you perform the point-in-time recovery from data corruption? It seems that we can only recover the data from the backup of the db files.  I assume we need to deliver a separate tool or new commands in etcdctl or etcdutl to do it in future. Please let me know if I missed anything.

server/storage/mvcc/hash.go
```diff
@@ -0,0 +1,148 @@
...
+	Hash() (hash uint32, revision int64, err error)
+
+	// HashByRev computes the hash of all MVCC revisions up to a given revision.
+	HashByRev(rev int64) (hash kvHash, currentRev int64, err error)
+
+	// Store adds hash value in local cache, allowing it can be returned by HashByRev.
+	Store(kvHash)
+}
+
+type hashStorage struct {
```
ahrtr, line 98: nit: The name `hashStorage` isn't accurate. It's supposed to be just a storage for historical hash values, but obviously it can also calculate the hash values, although it dedicates to other instance to do it.

ahrtr:
Overall this PR looks good to me. 

What's your plan on 3.5?  Only backporting this PR to 3.5 doesn't make much sense to me, because it doesn't address the existing flaw we have in 3.5. Of course, we can discuss it separately.

ahrtr (approved):

serathius:
Thanks for review! Backport will be done after full funcionality is implemented and properly tested.

server/etcdserver/corrupt.go
```diff
@@ -184,15 +183,15 @@ func (s *EtcdServer) checkHashKV() error {
 		})
 	}
 
-	if h2 != h && rev2 == rev && crev == crev2 {
+	if h2 != h && h2.Revision == h.Revision && h.CompactRevision == h2.CompactRevision {
```
serathius, line 186:
```suggestion
	if h2.Hash != h.Hash && h2.Revision == h.Revision && h.CompactRevision == h2.CompactRevision {
```
I though I fixed that :(

tests/integration/clientv3/maintenance_test.go
```diff
@@ -36,6 +37,11 @@ import (
 	"go.etcd.io/etcd/server/v3/storage/mvcc"
 )
 
+const (
+	// Use high prime
+	compactionCycle = 71
```
serathius, line 42: Not needed

server/etcdserver/corrupt.go
ahrtr: There is no difference. The suggested change has the same effect as the original implementation.

server/etcdserver/corrupt.go
```diff
@@ -150,11 +149,11 @@ func (s *EtcdServer) monitorKVHash() {
 func (s *EtcdServer) checkHashKV() error {
 	lg := s.Logger()
 
-	h, rev, crev, err := s.kv.HashByRev(0)
+	h, currentRev, err := s.KV().HashStorage().HashByRev(0)
 	if err != nil {
 		return err
 	}
-	peers := s.getPeerHashKVs(rev)
+	peers := s.getPeerHashKVs(currentRev)
```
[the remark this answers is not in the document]
ahrtr: Yes, this suggested change is the key point that fixes the existing flaw we have.

server/etcdserver/corrupt.go
serathius: ok, but that was not intentional :P

PR #14011, opened 2022-05-05
Nomination for Benjamin (ahrtr@) as a project maintainer

ptabor:
Benjamin (@ahrtr) is one of the current reviewers. Benjamin has proven high level of etcd expertise
and impact on the project, including delivering fixes to data-corruption bugs in etcd,
care about test quality and overall project health.

Per discussion with existing maintainers, I nominate him for the maintainer role.
Benjamin, thank you for your outstanding contributions to the project.

@gyuho
@mitake 
@jingyih 
@jpbetz 
@serathius 
@spzala 
@hexfusion 
@wenjiaswe 
@xiang90

PR #14053, opened 2022-05-19
Governance: Use lazy consensus when needed to make decision

spzala:
In lack of supermajority, we may sometimes require to hold on to
important decisions for long time. In order to speed up, after
giving enough time for supermajority, use lazy consensus.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -42,8 +42,9 @@ below.
 Contributors who are interested in becoming a maintainer, if performing these
 responsibilities, should discuss their interest with the existing maintainers. New
 maintainers must be nominated by an existing maintainer and must be elected by a
-supermajority of maintainers. Likewise, maintainers can be removed by a supermajority
-of the maintainers and moved to emeritus status.
+supermajority of maintainers with a fallback on lazy consensus after three business weeks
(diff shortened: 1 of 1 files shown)


spzala:
cc @serathius @ptabor @ahrtr @hexfusion @mitake @xiang90 @gyuho @jpbetz @jingyih @wenjiaswe

hexfusion:
Seems fine could you give an example of this in the past as a historical reference point.

ptabor:
Last time (https://github.com/etcd-io/etcd/pull/14011) we reached 6/9 in 3 days. So we are still good. 

I think this is "preceding step" for a case community would land in a formal deadlock (hopefully not).

ptabor (approved):

ahrtr (approved):

spzala:
@hexfusion @ptabor @ahrtr thanks for the quick review! As @ptabor said this is for any future case(s), if it's ever needed, to not to run into any deadlock.

hexfusion (approved):

jingyih (commented):
LGTM

serathius (approved):

gyuho (approved):

mitake (approved):

Issue #14025, opened 2022-05-09
It's possible to get "wal: max entry size limit exceeded" with recommended values


biosvs:
### What happened?

I run clean etcd node as followed:
`etcd --max-request-bytes=10485760`

Then through the go client I `put` to key `k1` value consists of `10*1024*1024-27` bytes.

Then I stopped server and try to start it again, but it failed with error `wal: max entry size limit exceeded` (https://github.com/etcd-io/etcd/blob/main/server/storage/wal/decoder.go#L88).

### What did you expect to happen?

1. It should not be possible to write more data than etcd can read then.
2. It should be obvious how much data (key + value) in bytes I'm able to write. I set --max-request-bytes to 10mb, but in fact was able to pass 25bytes less data (2bytes to key, 10mb-27bytes for data). Maybe internal overhead should not be part of validation?
3. It should not be possible to set --max-request-bytes to value so high that etcd will allow to write more than it can read then. OR the WAL limit should be configurable.

### How can we reproduce it (as minimally and precisely as possible)?

Run clean etcd instance:
`etcd --max-request-bytes=10485760`

Run this go-code:
```
package main

import (
	"context"
	"go.etcd.io/etcd/clientv3"
)

func main() {
	length := 10*1024*1024 - 27
	b := make([]byte, length)
	for i := 0; i < length; i++ {
		b[i] = 'a'
	}

	cli, err := clientv3.New(clientv3.Config{
		Endpoints:          []string{"http://127.0.0.1:2379"},
		MaxCallSendMsgSize: length + 1024,
	})

	if err != nil {
		panic(err)
	}

	_, err = cli.Put(context.Background(), "k1", string(b))
	if err != nil {
		panic(err)
	}
}
```

### Anything else we need to know?

Let's assume you run code above, you will not be able to restart server.
But you can call `etcd snap save` while server is still running, delete all WAL files and the start server and receive saved value.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
etcd Version: 3.5.4
Git SHA: 08407ff76
Go Version: go1.18.1
Go OS/Arch: darwin/amd64

$ etcdctl version
etcdctl version: 3.5.4
API version: 3.5
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

--max-request-bytes=10485760

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
[...]

PR #14057, opened 2022-05-20
wal: removes entry size limit

algobardo:
Solves #14025

I don't fully get the reason the entry size limit was introduced.
I have added tests that show that the entry size that is currently set as maximum can be exceeded, without compromising the wal functionalities.

### DIFF
--- a/server/storage/wal/decoder.go
+++ b/server/storage/wal/decoder.go
@@ -59,11 +59,6 @@ func (d *decoder) decode(rec *walpb.Record) error {
 	return d.decodeRecord(rec)
 }
 
-// raft max message size is set to 1 MB in etcd server
-// assume projects set reasonable message size limit,
-// thus entry size should never exceed 10 MB
-const maxWALEntrySizeLimit = int64(10 * 1024 * 1024)
-
 func (d *decoder) decodeRecord(rec *walpb.Record) error {
 	if len(d.brs) == 0 {
 		return io.EOF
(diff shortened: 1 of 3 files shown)


server/storage/wal/wal_test.go
```diff
@@ -368,48 +369,56 @@ func TestSaveWithCut(t *testing.T) {
 	}
 }
 
-func TestRecover(t *testing.T) {
-	p := t.TempDir()
+func TestRecoverWithDifferentEntrySizes(t *testing.T) {
+	zeroMb := 0
+	oneMb := 1 * 1024 * 1024
+	twentyMb := 20 * 1024 * 1024
+	for _, entrySize := range []int{zeroMb, oneMb, twentyMb} {
```
ahrtr, line 376: Please use subtest instead.  There are lots of examples, such as [discovery_test.go#L66](https://github.com/etcd-io/etcd/blob/b7be91f98b7a7a6202ed9e9bc37b692d484585a8/server/etcdserver/api/v3discovery/discovery_test.go#L66)

ahrtr, line 376: Please use subtest instead.

ahrtr (changes requested):

ahrtr:
Overall looks good to me. Please fix the comments and see [issuecomment-1133780272](https://github.com/etcd-io/etcd/issues/14025#issuecomment-1133780272)

server/storage/wal/decoder.go
```diff
@@ -84,10 +79,6 @@ func (d *decoder) decodeRecord(rec *walpb.Record) error {
 	}
 
 	recBytes, padBytes := decodeFrameSize(l)
-	if recBytes >= maxWALEntrySizeLimit-padBytes {
```
ptabor, line 87: I think it was defensive protection against OOMing or consuming whole server memory in case of a corrupted frame. 
I agree with @ahrtr suggestion from: https://github.com/etcd-io/etcd/issues/14025#issuecomment-1133781965 to use: 

https://github.com/etcd-io/etcd/blob/8093fc90abe78a431c841a688c7e66fd660d20dd/server/storage/wal/wal.go#L55

And refuse configs where --max-request-bytes > [SegmentSizeBytes](https://github.com/etcd-io/etcd/blob/8093fc90abe78a431c841a688c7e66fd660d20dd/server/storage/wal/wal.go#L55)/4=16MB

Issue #13468, opened 2021-11-10
support grpc middleware for embed server


Ehco1996:
i think it's helpful for many projects, i can do this work if anyone can help with the code review

rleungx:
It's better to also support the stream interceptor.

Ehco1996:
> It's better to also support the stream interceptor.

thanks! i will try impl stream and  unary interceptor in server ServerOption

jiacai2050:
Anyone working on this?

I'm glad to help if maintainers still think this is useful.

PR #12290, opened 2020-09-12
*: Implement etcd QoS POC

tangcong:
## Draft Proposal

please see latest draft proposal: [etcd QoS Feature Design](https://docs.google.com/document/d/122e-uLqngbL9qUEHx4pWcxQj8WO4PCbItTZ-jtkt7ik/edit?usp=sharing)

## Background

As the metadata storage of kubernetes, etcd, stability is extremely important. However, at present, etcd will consume a lot of CPU, memory, and bandwidth resources in the face of some highly concurrent expensive read / write requests, resulting in high service load, even OOM, etc. Common expensive requests such as full keyspace fetch, a lot of event queries(list event), a lot of pod queries(list pods), and a lot of configmap writes.

 At the same time, etcd currently has only an extremely simple speed limit protection. When etcd's committed index is greater than the applied index threshold is greater than 5000, it will reject all requests and return the error ErrTooManyRequests. Therefore, its shortcomings are obvious, it is impossible to accurately limit the speed of expensive read / write requests, and it cannot effectively prevent the cluster from being overloaded. 

Therefore, I hope to implement a QoS feature that allows users to set the maximum qps under different paths and gRPC methods, and can automatically identify expensive read / write requests and limit their speed to prevent etcd overload and further improve etcd cluster Stability.

## Goals

● Support the configuration QoS rules based on multiple object types (such as grpcMethod, grpcMethod + request key prefix path, traffic, cpu-intensive/RangeKeyNum, latency)

● Support multiple rate limiters for different scenarios (such as token bucket filter,maxInflightRequest, leaky bucket filter), with good scalability.

● Different QoS rules support the configuration of different rate limiters, priority. if a request satisfies multiple rules, the rule with the highest priority is selected.

● The memory consumption of the QoS feature can be controlled without significantly reducing the server performance.

● Users can customize rules simply through conditional expressions. (For example, dbUsedbytes is about to exceed db quota bytes, you can limit the rate of k8s event events）

## POC Implementation/Feature

●  Implement QoS API, rules can be adjusted dynamically.
●  Support multiple rate limiter and rule type.
●  Users can customize rules simply through conditional expressions.

## Architecture/Flow

image

### add qos rule
```
etcdctl qos add rule-gRPCMethod-1 gRPCMethod Range /prefix 100 TokenBucket 1 0 ""
```
```
NAME:
	qos add - adds a qos rule into the cluster

USAGE:
[...]

PR #7215, opened 2017-01-22
embed: support user defined grpc services.

disksing:
Fixes #7200

Hi again,
My original design does not progress well, the main problem is that the code generated by gRPC does not export corresponding ServiceDesc. Please refer https://github.com/grpc/grpc-go/blob/master/examples/helloworld/helloworld/helloworld.pb.go#L123
It seems we have to setup a `func(*grpc.Server)` instead.
/cc @xiang90 @heyitsanthony

PR #14066, opened 2022-05-23
embed: add `GRPCAdditionalServerOptions` config

rleungx:
This PR introduces GRPCAdditionalServerOptions which allow changing the internal gRPC settings. Sometimes, we may register our own gRPC service into `etcd` and change the `max-request-bytes` might affect the internal etcd logic.

### DIFF
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -284,6 +284,10 @@ type Config struct {
 	// before closing a non-responsive connection. 0 to disable.
 	GRPCKeepAliveTimeout time.Duration `json:"grpc-keepalive-timeout"`
 
+	// GRPCAdditionalServerOptions is the additional server option hook
+	// for changing the default internal gRPC configuration.
+	GRPCAdditionalServerOptions []grpc.ServerOption `json:"grpc-additional-server-options"`
+
 	// SocketOpts are socket options passed to listener config.
 	SocketOpts transport.SocketOpts `json:"socket-options"`
 
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -285,7 +285,10 @@ type Config struct {
 	GRPCKeepAliveTimeout time.Duration `json:"grpc-keepalive-timeout"`
 
 	// GRPCAdditionalServerOptions is the additional server option hook
-	// for changing the default internal gRPC configuration.
+	// for changing the default internal gRPC configuration. Note these
+	// additional configurations take precedence over the existing individual
+	// configurations if present. Please refer to
+	// https://github.com/etcd-io/etcd/pull/14066#issuecomment-1248682996
 	GRPCAdditionalServerOptions []grpc.ServerOption `json:"grpc-additional-server-options"`
 
 	// SocketOpts are socket options passed to listener config.
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -761,6 +761,7 @@ func (e *Etcd) serveClients() {
 			Timeout: e.cfg.GRPCKeepAliveTimeout,
 		}))
 	}
+	gopts = append(gopts, e.cfg.GRPCAdditionalServerOptions...)
 
 	splitHTTP := false
 	for _, sctx := range e.sctxs {
--- a/tests/framework/integration/cluster.go
+++ b/tests/framework/integration/cluster.go
@@ -144,14 +144,16 @@ type ClusterConfig struct {
 	QuotaBackendBytes    int64
 	BackendBatchInterval time.Duration
 
-	MaxTxnOps              uint
-	MaxRequestBytes        uint
+	MaxTxnOps       uint
+	MaxRequestBytes uint
+
 	SnapshotCount          uint64
 	SnapshotCatchUpEntries uint64
 
-	GRPCKeepAliveMinTime  time.Duration
-	GRPCKeepAliveInterval time.Duration
-	GRPCKeepAliveTimeout  time.Duration
+	GRPCKeepAliveMinTime        time.Duration
+	GRPCKeepAliveInterval       time.Duration
+	GRPCKeepAliveTimeout        time.Duration
+	GRPCAdditionalServerOptions []grpc.ServerOption
 
 	ClientMaxCallSendMsgSize int
 	ClientMaxCallRecvMsgSize int
@@ -278,6 +280,7 @@ func (c *Cluster) mustNewMember(t testutil.TB) *Member {
 			GRPCKeepAliveMinTime:                c.Cfg.GRPCKeepAliveMinTime,
 			GRPCKeepAliveInterval:               c.Cfg.GRPCKeepAliveInterval,
 			GRPCKeepAliveTimeout:                c.Cfg.GRPCKeepAliveTimeout,
+			GRPCAdditionalServerOptions:         c.Cfg.GRPCAdditionalServerOptions,
 			ClientMaxCallSendMsgSize:            c.Cfg.ClientMaxCallSendMsgSize,
 			ClientMaxCallRecvMsgSize:            c.Cfg.ClientMaxCallRecvMsgSize,
 			UseIP:                               c.Cfg.UseIP,
@@ -603,6 +606,7 @@ type MemberConfig struct {
 	GRPCKeepAliveMinTime        time.Duration
 	GRPCKeepAliveInterval       time.Duration
 	GRPCKeepAliveTimeout        time.Duration
+	GRPCAdditionalServerOptions []grpc.ServerOption
 	ClientMaxCallSendMsgSize    int
 	ClientMaxCallRecvMsgSize    int
 	UseIP                       bool
@@ -709,6 +713,7 @@ func MustNewMember(t testutil.TB, mcfg MemberConfig) *Member {
 			Timeout: mcfg.GRPCKeepAliveTimeout,
 		}))
 	}
+	m.GRPCServerOpts = append(m.GRPCServerOpts, mcfg.GRPCAdditionalServerOptions...)
 	m.ClientMaxCallSendMsgSize = mcfg.ClientMaxCallSendMsgSize
 	m.ClientMaxCallRecvMsgSize = mcfg.ClientMaxCallRecvMsgSize
 	m.UseIP = mcfg.UseIP
--- a/tests/integration/v3_grpc_test.go
+++ b/tests/integration/v3_grpc_test.go
@@ -1923,7 +1923,71 @@ func TestV3LargeRequests(t *testing.T) {
 					t.Errorf("#%d: range expected no error, got %v", i, err)
 				}
 			}
+		})
+	}
+}
 
+// TestV3AdditionalGRPCOptions ensures that configurable GRPCAdditionalServerOptions works as intended.
+func TestV3AdditionalGRPCOptions(t *testing.T) {
+	integration.BeforeTest(t)
+	tests := []struct {
+		name            string
+		maxRequestBytes uint
+		grpcOpts        []grpc.ServerOption
+		valueSize       int
+		expectError     error
+	}{
+		{
+			name:            "requests will get a gRPC error because it's larger than gRPC MaxRecvMsgSize",
+			maxRequestBytes: 8 * 1024 * 1024,
+			grpcOpts:        nil,
+			valueSize:       9 * 1024 * 1024,
+			expectError:     status.Errorf(codes.ResourceExhausted, "grpc: received message larger than max"),
+		},
+		{
+			name:            "requests will get an etcd custom gRPC error because it's larger than MaxRequestBytes",
+			maxRequestBytes: 8 * 1024 * 1024,
+			grpcOpts:        []grpc.ServerOption{grpc.MaxRecvMsgSize(10 * 1024 * 1024)},
+			valueSize:       9 * 1024 * 1024,
+			expectError:     rpctypes.ErrGRPCRequestTooLarge,
+		},
+		{
+			name:            "requests size is smaller than MaxRequestBytes but larger than MaxRecvMsgSize",
+			maxRequestBytes: 8 * 1024 * 1024,
+			grpcOpts:        []grpc.ServerOption{grpc.MaxRecvMsgSize(4 * 1024 * 1024)},
+			valueSize:       6 * 1024 * 1024,
+			expectError:     status.Errorf(codes.ResourceExhausted, "grpc: received message larger than max"),
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			clus := integration.NewCluster(t, &integration.ClusterConfig{
+				Size:                        1,
+				MaxRequestBytes:             test.maxRequestBytes,
+				ClientMaxCallSendMsgSize:    12 * 1024 * 1024,
+				GRPCAdditionalServerOptions: test.grpcOpts,
+			})
+			defer clus.Terminate(t)
+			kvcli := integration.ToGRPC(clus.Client(0)).KV
+			reqput := &pb.PutRequest{Key: []byte("foo"), Value: make([]byte, test.valueSize)}
+			if _, err := kvcli.Put(context.TODO(), reqput); err != nil {
+				if _, ok := err.(rpctypes.EtcdError); ok {
+					if err.Error() != status.Convert(test.expectError).Message() {
+						t.Errorf("expected %v, got %v", status.Convert(test.expectError).Message(), err.Error())
+					}
+				} else if !strings.HasPrefix(err.Error(), test.expectError.Error()) {
+					t.Errorf("expected error starting with '%s', got '%s'", test.expectError.Error(), err.Error())
+				}
+			}
+			// request went through, expect large response back from server
+			if test.expectError == nil {
+				reqget := &pb.RangeRequest{Key: []byte("foo")}
+				// limit receive call size with original value + gRPC overhead bytes
+				_, err := kvcli.Range(context.TODO(), reqget, grpc.MaxCallRecvMsgSize(test.valueSize+512*1024))
+				if err != nil {
+					t.Errorf("range expected no error, got %v", err)
+				}
+			}
 		})
 	}
 }
--- a/tests/integration/clientv3/kv_test.go
+++ b/tests/integration/clientv3/kv_test.go
@@ -702,14 +702,12 @@ func TestKVLargeRequests(t *testing.T) {
 		// without proper client-side receive size limit
 		// "code = ResourceExhausted desc = grpc: received message larger than max (5242929 vs. 4194304)"
 		{
-
 			maxRequestBytesServer:  7*1024*1024 + 512*1024,
 			maxCallSendBytesClient: 7 * 1024 * 1024,
 			maxCallRecvBytesClient: 0,
 			valueSize:              5 * 1024 * 1024,
 			expectError:            nil,
 		},
-
 		{
 			maxRequestBytesServer:  10 * 1024 * 1024,
 			maxCallSendBytesClient: 100 * 1024 * 1024,


ahrtr:
Comments:
1. What do you mean by `Sometimes, we may register our own gRPC service into etcd`?  Could you explain this?
2. Why you do not expose the two items? i.e. [config.go#L132](https://github.com/etcd-io/etcd/blob/c3bc4116dcd1272c6a1318feadf5ac9739409a12/server/etcdmain/config.go#L132)

server/etcdserver/api/v3rpc/grpc.go
```diff
@@ -66,8 +62,8 @@ func Server(s *etcdserver.EtcdServer, tls *tls.Config, interceptor grpc.UnarySer
 	opts = append(opts, grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(chainUnaryInterceptors...)))
 	opts = append(opts, grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(chainStreamInterceptors...)))
 
-	opts = append(opts, grpc.MaxRecvMsgSize(int(s.Cfg.MaxRequestBytes+grpcOverheadBytes)))
-	opts = append(opts, grpc.MaxSendMsgSize(maxSendBytes))
+	opts = append(opts, grpc.MaxRecvMsgSize(s.Cfg.MaxRecvMsgSize))
```
ahrtr, line 65: What if `MaxRequestBytes` is different from `s.Cfg.MaxRecvMsgSize`?  If you configure a big value for `s.Cfg.MaxRecvMsgSize`, i.e. 10MB, while the `MaxRequestBytes` is 1 MB, then etcdserver may still reject the request.

rleungx: Do you mean we should ensure `MaxRecvMsgSize` is larger than `MaxRequestBytes`? The internal raft request is still restricted by `MaxRequestBytes`.

rleungx: Do you mean `MaxRecvMsgSize` should be larger or equal to `MaxRequestBytes`?

ahrtr: I don't see any reason to add the two parameters. 

Did you see any issue without adding MaxRecvMsgSize/MaxSendMsgSize?

rleungx: Actually, we have a feature that needs to change `MaxRecvMsgSize` to 150M. I was wondering if etcd can offer a separate config to customize it without changing `MaxRequestBytes`.  For now, when we start a cluster, it will give a warning about the `MaxRequestBytes` exceeds the recommended size.

ahrtr: A couple of comments:
1. etcd is designed to serve small metadata. Why do you need to change `MaxRecvMsgSize` to 150M? Can you elaborate your use case?
2. Have you verified your case based on this PR with `MaxRecvMsgSize` being set to 150M, while keep `--max-request-bytes` unchanged?

ahrtr: Any update on this?

rleungx: Sorry for the late reply.
> 1. etcd is designed to serve small metadata. Why do you need to change MaxRecvMsgSize to 150M? Can you elaborate your use case?

We will register our service into the embed etcd. See https://github.com/tikv/pd/blob/0d05fba64372408c5af9a78cd9b47b58c4925ca9/server/server.go#L277-L280. For example, the store heartbeat https://github.com/tikv/pd/blob/0d05fba64372408c5af9a78cd9b47b58c4925ca9/server/grpc_service.go#L570.When we need to recover our cluster, the heartbeat request will bring the meta of data, which could be large. We utilize the meta to decide which part of the data should be recovered.

> 2. Have you verified your case based on this PR with MaxRecvMsgSize being set to 150M, while keep --max-request-bytes unchanged?

not yet

ptabor: I would consider registering a hook to alter the grpc options as part of the config: 

`func additionalGrpcServerOptions() grpc.ServerOption[]`

ptabor: Actually: 
```
   grpcAdditionalServerOptions grpc.ServerOption[]
```
in the config would do the trick.

server/config/config.go
```diff
@@ -129,6 +129,11 @@ type ServerConfig struct {
 	// MaxRequestBytes is the maximum request size to send over raft.
 	MaxRequestBytes uint
 
+	// MaxRecvMsgSize is the max gRPC message size in bytes the server can receive.
```
ptabor, line 132: How about adding 
  `grpcAdditionalServerOptions grpc.ServerOption[]` as a generic options instead ?

ptabor (changes requested):
grpcAdditionalServerOptions grpc.ServerOption[]

ahrtr:
@ptabor  One item in my to do list is to sort out all the thresholds in etcd, including this one. Will discuss this in a separate session once my current task of refactoring lease is done.

ptabor:
> @ptabor One item in my to do list is to sort out all the thresholds in etcd, including this one. Will discuss this in a separate session once my current task of refactoring lease is done.

@ahrtr As this sound as a snowflake use-case (grpc shared between etcd and other service), I think we should not be in the business of looking at the limits... and allow programmers (only in embed server) to pass additional grpc settings, that etcd passes-through.

ahrtr:
> @ahrtr As this sound as a snowflake use-case (grpc shared between etcd and other service), I think we should not be in the business of looking at the limits... and allow programmers (only in embed server) to pass additional grpc settings, that etcd passes-through.

Overall adding a generic `grpcAdditionalServerOptions grpc.ServerOption[]` looks good to me. But it doesn't resolve the concern that it may have impact on the build-in services provided by etcd itself. 

The good side is flexibility, because users can set any additional options they want. The down side is it may have impact on the gRPC service provided by etcd itself. 

Another perspective to think about this is that we should take all services, including etcd build-in services and users registered services, under control. Once users see any issue for the embedded etcd, they will also raise issue in etcd community.  Based on the principle of equal responsibility and obligation, If we need to support such case, then we should can add more restriction on the services registered by users, instead of just providing flexibility without any restriction.

ahrtr:
The new feature of [ServiceRegister](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/embed/config.go#L309) was introduced in [pull/7215](https://github.com/etcd-io/etcd/pull/7215). 

Again, the good side of the [PR]((https://github.com/etcd-io/etcd/pull/7215)) is users can reuse the same port as etcd. But the bad side is it introduces additional unnecessary complexity to etcd. Personally I don't like the PR ( I mean 7215), and the better choice should be resolving it out of the box (such as providing a proxy/gateway in front of embedded etcd and users' gRPC service) instead of hacking etcd itself.

rleungx:
> > @ahrtr As this sound as a snowflake use-case (grpc shared between etcd and other service), I think we should not be in the business of looking at the limits... and allow programmers (only in embed server) to pass additional grpc settings, that etcd passes-through.
> 
> Overall adding a generic `grpcAdditionalServerOptions grpc.ServerOption[]` looks good to me. But it doesn't resolve the concern that it may have impact on the build-in services provided by etcd itself.
> 
> The good side is flexibility, because users can set any additional options they want. The down side is it may have impact on the gRPC service provided by etcd itself.
> 
> Another perspective to think about this is that we should take all services, including etcd build-in services and users registered services, under control. Once users see any issue for the embedded etcd, they will also raise issue in etcd community. Based on the principle of equal responsibility and obligation, If we need to support such case, then we should can add more restriction on the services registered by users, instead of just providing flexibility without any restriction. So users shouldn't set any additional options, instead they should follow all the existing options from this perspective.

IMO, no matter if it's a requirement of user-defined services, we still need to provide the ability to change the default gRPC settings.

rleungx:
Besides, supporting the custom interceptor for embed etcd is another need. See https://github.com/etcd-io/etcd/issues/13468

ahrtr:
Note that the users registered service and the etcd build-in gRPC services share the same gRPC channel. Any options will have impact on both of them. 

Please just use the existing flag `--max-request-bytes ` to change the MaxRecvSize. Do you need to change the MaxSendSize in your case?

nolouch:
@ahrtr  https://github.com/etcd-io/etcd/pull/7215  not allow users to define gRPC's setting, I think this PR is to improve it.

We have some requirements with the `embed` etcd use-case, not only `max-request-bytes` :
-  register own service in etcd, and we want to protect with those services, but etcd's service may influence it (some client directly use etcd service), so we need QoS features. But same as the [issue comment](https://github.com/etcd-io/etcd/pull/12290#issuecomment-697060195) says, it suggests this can be simply implemented as a gRPC server middleware, but we cannot use custom gRPC server middleware
- Fixed parameters bring some inconvenience because it is not extensible, we need to add supported parameters to etcd first, and then wait for it release to the stable version, and then update the dependencies. This time period is too long.

IMO, `embed etcd` users need to be responsible for changing etcd parameters, this does not affect normal etcd users(standalone process). So hopefully this PR will be approved, it is more convenient for `embed etcd` users.

server/embed/etcd.go
```diff
@@ -719,6 +719,7 @@ func (e *Etcd) serveClients() (err error) {
 			Time:    e.cfg.GRPCKeepAliveInterval,
 			Timeout: e.cfg.GRPCKeepAliveTimeout,
 		}))
+		gopts = append(gopts, e.cfg.GRPCAdditionalServerOptions...)
```
nolouch, line 722: Does it need to move out this if statement?

ptabor: +1

tests/framework/integration/cluster.go
```diff
@@ -553,10 +556,11 @@ type Member struct {
 	Server        *etcdserver.EtcdServer
 	ServerClosers []func()
 
-	GrpcServerOpts []grpc.ServerOption
-	GrpcServer     *grpc.Server
-	GrpcURL        string
-	GrpcBridge     *bridge
+	GrpcServerOpts           []grpc.ServerOption
+	GrpcAdditionalServerOpts []grpc.ServerOption
```
ptabor, line 560: Do we really need it here ? 

I think that `MustNewMember` could just propagate `MemberConfig.GrpcAdditionalServerOpts` into `GrpcServerOpts` and not store it here explicitly.

ahrtr:
> Hi @ahrtr #7215 not allow users to define gRPC's setting, I think this PR is to improve it.
> 
> We have some requirements with the `embed` etcd use-case, not only `max-request-bytes` :
> 
>     * register own service in etcd, and we want to protect with those services, but etcd's service may influence it (some client directly use etcd service), so we need `QoS features`. But same as the [issue comment](https://github.com/etcd-io/etcd/pull/12290#issuecomment-697060195) says, it suggests this can be simply implemented as a gRPC server middleware, and keep etcd as minimal as possible, but we cannot use custom gRPC server middleware
> 
>     * Fixed parameters bring some inconvenience because it is not extensible, we need to add supported parameters to etcd first, and then wait for it release to the stable version, and then update the dependencies. This time period is too long.
> 
> 
> IMO, `embed etcd` users need to be responsible for changing etcd parameters, this does not affect normal etcd users(standalone process). So hopefully this PR will be approved, it is more convenient for `embed etcd` users. ptal @ahrtr @ptabor, Thanks.

I am still not convinced that we really need to add `grpcAdditionalServerOptions grpc.ServerOption[]`. I am not saying we shouldn't add it, instead I am just trying to say that we shouldn't add any new feature without a deep understanding the use cases beforehand. So please elaborate your use cases firstly. Why do you need it, and how will use it after adding it? 

Note that there are several QoS related PRs, but none of them is merged so far. If your discussion is based on the QoS feature, then it isn't valid at the moment. 

More flexibility also means more chance to make mistake. If there is no any strong incentive or reasonable benefit, then we shouldn't add it. 

Please also make sure you well understood my previous two comments, [issuecomment-1155881593](https://github.com/etcd-io/etcd/pull/14066#issuecomment-1155881593) and [issuecomment-1155894742](https://github.com/etcd-io/etcd/pull/14066#issuecomment-1155894742)

ptabor:
@ahrtr: 
1. I'm not insisting on merging, this... just steering towards such variant rather than explosion of custom options. 

2. A use-case is linked from: https://github.com/etcd-io/etcd/pull/14066#discussion_r887444999, and seems direct consequence of allowing embed server to have custom customer's services being used. I think usage of embedded service put's a high bar on customer to have proper testing for their integrated solution. We should document this requirement more explicitly. 

3. I was hoping (but seems not possible currently) that we could perform validation of different constants/limit post-grpc server start, It's run it and obtain the effective config. Unfortunately grpc does not exposes the resolved configuration of running server publicly: 

https://github.com/grpc/grpc-go/blob/28de4866ce7440b675662abbdd5c43b476bd4dae/server.go#L124

rleungx:
@ahrtr @ptabor Thanks for your reply. https://github.com/etcd-io/etcd/pull/14066#issuecomment-1163834967 make sense to me. But for now, we indeed have two requirements:
1. support changing the `MaxRecvMsgSize` which can be replaced by changing `max-request-bytes`. It doesn't well conform to semantics but it works. I'm ok with the current usage.
2. support adding the custom gRPC interceptor for user-defined gRPC services so we can add our logic to intercept the request and do something, e.g. the QoS feature or forwarding requests. But etcd doesn't support it at the moment.

nolouch:
@ahrtr 
> Once users see any issue for the embedded etcd, they will also raise issue in etcd community. Based on the principle of equal responsibility and obligation, If we need to support such case, then we should can add more restrictions on the services registered by users, instead of just providing flexibility without any restriction. 

> I don't think it's good to just expose them all without deep understanding and full testing ourselves (the community) beforehand

I have some questions:
- Community can issue that and add the option if approve, as I said before, users need to wait a long time. on the other hand, how do restriction users not use the inappropriate option value that may affect etcd? 
- As you said, currently etcd exposes about half of 24 options, so about 12 options. Do all users really deep understand those 12 options and full testing by yourselves with all use cases?

I think there still have the same problems. so if the user wants to change the default parameters, they should understand it, and be responsible for it. that means users should test it and make sure it is suitable for themselves scenarios. As @ptabor said, we can document it. 
And I don't think it introduces unnecessary complexity to etcd, etcd still keep minimal, and it's more flexibility, let user can extend etcd (such as grpc services and middlewares).

nolouch:
BTW, If you really concert about it, I think etcd `embed model` should be more like a library, it can be as a service be registered to a custom grpc server rather than let itself as a server. what is your opinion?

ahrtr:
I am OK to revisit this PR, but it needs at least three maintainers' approval.

tests/integration/v3_grpc_test.go
```diff
@@ -1922,7 +1922,45 @@ func TestV3LargeRequests(t *testing.T) {
...
 			}
+		})
+	}
+}
 
+// TestV3AdditionalGRPCOptions ensures that configurable GRPCAdditionalServerOptions works as intended.
+func TestV3AdditionalGRPCOptions(t *testing.T) {
+	integration.BeforeTest(t)
+	tests := []struct {
+		maxRequestBytes uint
```
ahrtr, line 1933: Please add one more field `name` into the struct to clarify the purpose of the case.

tests/integration/v3_grpc_test.go
```diff
@@ -1922,7 +1922,45 @@ func TestV3LargeRequests(t *testing.T) {
...
+		grpcOpts        []grpc.ServerOption
+		valueSize       int
+		expectError     error
+	}{
+		{8 * 1024 * 1024, nil, 9 * 1024 * 1024, status.Errorf(codes.ResourceExhausted, "grpc: received message larger than max")},
+		{8 * 1024 * 1024, []grpc.ServerOption{grpc.MaxRecvMsgSize(10 * 1024 * 1024)}, 9 * 1024 * 1024, rpctypes.ErrGRPCRequestTooLarge},
+	}
+	for i, test := range tests {
+		t.Run(fmt.Sprintf("#%d", i), func(t *testing.T) {
```
ahrtr, line 1942: Use `name` to be added above as the test case name.

No need to print the `i` in the following code. The golang test lib will print the `name` automatically when fails.

ahrtr:
The `GRPCAdditionalServerOptions` may define duplicate items as the existing conf ones. It seems that `GRPCAdditionalServerOptions` overwrites the existing items, and it's basically OK. But `GRPCAdditionalServerOptions` only takes effect in the gRPC communication, but some existing items not only takes effect in the gRPC communication, but also handled by etcd, such as [v3_server.go#L687](https://github.com/etcd-io/etcd/blob/012fc51390f6a4c91ae62a94e88fdce38b3ee26b/server/etcdserver/v3_server.go#L687). 

So we need to get all these sorted out.

rleungx:
> The `GRPCAdditionalServerOptions` may define duplicate items as the existing conf ones. It seems that `GRPCAdditionalServerOptions` overwrites the existing items, and it's basically OK. But `GRPCAdditionalServerOptions` only takes effect in the gRPC communication, but some existing items not only takes effect in the gRPC communication, but also handled by etcd, such as [v3_server.go#L687](https://github.com/etcd-io/etcd/blob/012fc51390f6a4c91ae62a94e88fdce38b3ee26b/server/etcdserver/v3_server.go#L687).
> 
> So we need to get all these sorted out.

It seems only `MaxRequestBytes` will affect the etcd itself.

ahrtr:
I think it makes more sense to make the existing configuration items overwrite the items defined in `GRPCAdditionalServerOptions`. For example, if users define both `--max-request-bytes` and `[]grpc.ServerOption{grpc.MaxRecvMsgSize(...)}`, then the `--max-request-bytes` overwrites the latter.

rleungx:
> I think it makes more sense to make the existing configuration items overwrite the items defined in `GRPCAdditionalServerOptions`. For example, if users define both `--max-request-bytes` and `[]grpc.ServerOption{grpc.MaxRecvMsgSize(...)}`, then the `--max-request-bytes` overwrites the latter.

How about adding some comments to emphasize that `GRPCAdditionalServerOptions` may overwrite the existing gRPC configs? Or is there any way to tell if the config is set by users without comparing it with the default value?

ahrtr:
The existing individual configurations should take precedence over the generic configurations `grpcAdditionalServerOptions`, and we should guarantee this from code level. 

You just need to make sure the items in `grpcAdditionalServerOptions` are in the head of the option list, because the latter will overwrite the former if two items have the same key. Please see [grpc.go#L72](https://github.com/etcd-io/etcd/blob/4d57eb8d3b2190bfdba3c65f8eb93c0349fc6dcc/server/etcdserver/api/v3rpc/grpc.go#L72) and [server.go#L563-L565](https://github.com/grpc/grpc-go/blob/5b509df1e30a25b14dba119c09295924c9e17abe/server.go#L563-L565)

rleungx:
> The existing individual configurations should take precedence over the generic configurations `grpcAdditionalServerOptions`, and we should guarantee this from code level.
> 
> You just need to make sure the items in `grpcAdditionalServerOptions` are in the head of the option list, because the latter will overwrite the former if two items have the same key. Please see [grpc.go#L72](https://github.com/etcd-io/etcd/blob/4d57eb8d3b2190bfdba3c65f8eb93c0349fc6dcc/server/etcdserver/api/v3rpc/grpc.go#L72) and [server.go#L563-L565](https://github.com/grpc/grpc-go/blob/5b509df1e30a25b14dba119c09295924c9e17abe/server.go#L563-L565)

Actually, we have another need to add some new interceptors. If using this way, the options of interceptors will be overwritten I think.

ahrtr:
Previously I was thinking the existing individual configurations should take precedence over the generic configurations `grpcAdditionalServerOptions`, but it seems not a big deal, since the generic configuration will only be used by the users of embedded etcd. 

Please consider to add a verification on the configuration. If there are duplicated items, then raise an error and fail the process.

rleungx:
Do you mean the duplicated items inside `grpcAdditionalServerOptions` or between `grpcAdditionalServerOptions` and other gRPC-related configurations?

tests/integration/v3_grpc_test.go
```diff
@@ -1923,7 +1923,48 @@ func TestV3LargeRequests(t *testing.T) {
...
+		name            string
+		maxRequestBytes uint
+		grpcOpts        []grpc.ServerOption
+		valueSize       int
+		expectError     error
+	}{
+		{"requests will get a gRPC error because it's larger than gRPC MaxRecvMsgSize",
+			8 * 1024 * 1024, nil, 9 * 1024 * 1024, status.Errorf(codes.ResourceExhausted, "grpc: received message larger than max")},
+		{"requests will get an etcd custom gRPC error because it's larger than MaxRequestBytes",
+			8 * 1024 * 1024, []grpc.ServerOption{grpc.MaxRecvMsgSize(10 * 1024 * 1024)}, 9 * 1024 * 1024, rpctypes.ErrGRPCRequestTooLarge},
```
ahrtr, line 1943: Please try to make this more readable using structure something like below,
```
{
    name: "abc",
    maxRequestBytes: 8 * 1024 * 102,
   ...
}
```

tests/integration/v3_grpc_test.go
```diff
@@ -1923,7 +1923,48 @@ func TestV3LargeRequests(t *testing.T) {
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1, MaxRequestBytes: test.maxRequestBytes, ClientMaxCallSendMsgSize: 12 * 1024 * 1024, GRPCAdditionalServerOptions: test.grpcOpts})
```
ahrtr, line 1947: ditto

tests/integration/v3_grpc_test.go
```diff
@@ -1923,7 +1923,48 @@ func TestV3LargeRequests(t *testing.T) {
...
+			kvcli := integration.ToGRPC(clus.Client(0)).KV
+			reqput := &pb.PutRequest{Key: []byte("foo"), Value: make([]byte, test.valueSize)}
+			_, err := kvcli.Put(context.TODO(), reqput)
+			if _, ok := err.(rpctypes.EtcdError); ok {
+				if err.Error() != status.Convert(test.expectError).Message() {
+					t.Errorf("expected %v, got %v", status.Convert(test.expectError).Message(), err.Error())
+				}
+			} else if err != nil && !strings.HasPrefix(err.Error(), test.expectError.Error()) {
+				t.Errorf("expected error starting with '%s', got '%s'", test.expectError.Error(), err.Error())
+			}
```
ahrtr, line 1958:
```suggestion
			if _, err := kvcli.Put(context.TODO(), reqput); err != nil {
				if _, ok := err.(rpctypes.EtcdError); ok {
					if err.Error() != status.Convert(test.expectError).Message() {
						t.Errorf("expected %v, got %v", status.Convert(test.expectError).Message(), err.Error())
					}
				} else !strings.HasPrefix(err.Error(), test.expectError.Error()) {
					t.Errorf("expected error starting with '%s', got '%s'", test.expectError.Error(), err.Error())
				}
			}
```

ahrtr:
> > Please consider to add a verification on the configuration. If there are duplicated items, then raise an error and fail the process.
> 
> Do you mean the duplicated items inside `grpcAdditionalServerOptions` or between `grpcAdditionalServerOptions` and other gRPC-related configurations?

It might be a little subtle. For example, [MaxRequestBytes](https://github.com/etcd-io/etcd/blob/2fa73023a0dc65e8f6bf5e032df15bba6f8cddca/server/embed/config.go#L208) is used by both gRPC and etcdserver, and it's default value is 1.5MB. If you add `grpc.MaxRecvMsgSize` into `grpcAdditionalServerOptions`, it only affects the gRPC communication, and probably your customized interceptor, but not etcdserver. 

If the `grpc.MaxRecvMsgSize` in `grpcAdditionalServerOptions` has a smaller value than [MaxRequestBytes](https://github.com/etcd-io/etcd/blob/2fa73023a0dc65e8f6bf5e032df15bba6f8cddca/server/embed/config.go#L208), then it means the max request what the etcdserver can receive is actually smaller than expected. If it has a bigger value, then a request may be accepted by gRPC, but eventually rejected by etcdserver. The latter case might be OK.  Note that based on your current implementation, the generic configurations has higher priority than the existing individual configuration items.

Since `grpcAdditionalServerOptions` will only be used for etcd as embedded case, so the `embed user` should take the responsibility to provide proper configurations to make sure everything (including etcdserver and customized interceptors) work as expected. 

Overall looks good to me. cc @serathius @spzala @ptabor

tests/integration/v3_grpc_test.go
```diff
@@ -1923,7 +1923,64 @@ func TestV3LargeRequests(t *testing.T) {
...
+			valueSize:       9 * 1024 * 1024,
+			expectError:     status.Errorf(codes.ResourceExhausted, "grpc: received message larger than max"),
+		},
+		{
+			name:            "requests will get an etcd custom gRPC error because it's larger than MaxRequestBytes",
+			maxRequestBytes: 8 * 1024 * 1024,
+			grpcOpts:        []grpc.ServerOption{grpc.MaxRecvMsgSize(10 * 1024 * 1024)},
+			valueSize:       9 * 1024 * 1024,
+			expectError:     rpctypes.ErrGRPCRequestTooLarge,
+		},
```
ahrtr, line 1953: Please consider to add a case that the `MaxRecvMsgSize` included in `GRPCAdditionalServerOptions` is smaller than the existing individual `MaxRequestBytes`, and the request size is in-between.

ahrtr:
Please rebase this PR, and squash the commits.

tests/integration/v3_grpc_test.go
ahrtr: any update on this?

ahrtr:
The go version has already been bumped to 1.19.1 in pipeline (github workflow), so please rebase this PR, and it should can resolve the `Release` workflow failure.

ahrtr (approved):
LGTM

Thank you @rleungx 

cc @ptabor @spzala @serathius @mitake

rleungx:
@ptabor PTAL

nolouch:
Hi @serathius @ptabor  
Could you please take a look?

rleungx:
Could we move it on? @ahrtr @ptabor

rleungx:
Friendly ping @ahrtr

ahrtr:
> Friendly ping @ahrtr

Sorry for the late response. Please rebase this PR.

rleungx:
> > Friendly ping @ahrtr
> 
> Sorry for the late response. Please rebase this PR.

done

server/embed/config.go
```diff
@@ -284,6 +284,10 @@ type Config struct {
 	// before closing a non-responsive connection. 0 to disable.
 	GRPCKeepAliveTimeout time.Duration `json:"grpc-keepalive-timeout"`
 
+	// GRPCAdditionalServerOptions is the additional server option hook
+	// for changing the default internal gRPC configuration.
```
ahrtr, line 288:
```suggestion
	// GRPCAdditionalServerOptions is the additional server option hook
	// for changing the default internal gRPC configuration. Note these
	// additional configurations take precedence over the existing individual
	// configurations if present. Please refer to 
	// https://github.com/etcd-io/etcd/pull/14066#issuecomment-1248682996
```

ahrtr:
Overall looks good to me. cc @fuweid  @jmhbnz @serathius  @tjungblu

tjungblu:
same. /lgtm

server/embed/config.go
rleungx: done

ahrtr:
@rleungx FYI. https://github.com/etcd-io/etcd/pull/18015#issuecomment-2162101637

rleungx:
@ahrtr Is there anything I need to do to get this PR merged?

server/embed/config.go
```diff
@@ -284,6 +284,13 @@ type Config struct {
 	// before closing a non-responsive connection. 0 to disable.
 	GRPCKeepAliveTimeout time.Duration `json:"grpc-keepalive-timeout"`
 
+	// GRPCAdditionalServerOptions is the additional server option hook
+	// for changing the default internal gRPC configuration. Note these
+	// additional configurations take precedence over the existing individual
+	// configurations if present. Please refer to
+	// https://github.com/etcd-io/etcd/pull/14066#issuecomment-1248682996
+	GRPCAdditionalServerOptions []grpc.ServerOption `json:"grpc-additional-server-options"`
```
ahrtr, line 292: It would be better if we also add a page under the https://etcd.io/docs/v3.5/dev-guide/ to document this. Can you raise a ticket in https://github.com/etcd-io/website/issues?

ahrtr:

server/embed/config.go
rleungx: https://github.com/etcd-io/website/issues/862

ivanvc (approved):
I've been reviewing this pull request back and forth. The implementation looks good. But I see some previous discussion about whether this is a feature we want to bring or not. It seems like the latest is that there is a consensus. Maybe it would be good to hear from other team members, too.

jmhbnz (commented):
Thanks for the work on this @rleungx.

Given this is a more complex configuration option I suggest adding a commented out example in https://github.com/etcd-io/etcd/blob/main/etcd.conf.yml.sample showing how this option could be configured via configuration file if desired.

ahrtr:
> Thanks for the work on this @rleungx.
> 
> Given this is a more complex configuration option I suggest adding a commented out example in https://github.com/etcd-io/etcd/blob/main/etcd.conf.yml.sample showing how this option could be configured via configuration file if desired.

The new option is only for the embedded use case. Users won't be able to configure it through the command line or config file.

jmhbnz (approved):
Thanks for clarification @ahrtr.

LGTM

ahrtr:
Please also update [changelog-3.6](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md), which can be done in a separate PR. @rleungx

rleungx:
> Please also update [changelog-3.6](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md), which can be done in a separate PR. @rleungx

Sure. Once it is merged, I will raise another PR to update it.

okJiang:
@ahrtr Sorry to bother you, can this PR accept a pick back to 3.5? I can submit a PR for review. Any suggestions are valuable to me.

ahrtr:
> @ahrtr Sorry to bother you, can this PR accept a pick back to 3.5? I can submit a PR for review. Any suggestions are valuable to me.

We may not backport this (feature) to stable release without strong justification.

PR #14067, opened 2022-05-24
Fix the race condition between goroutine and channel on the same leases to be revoked

ahrtr:
It's a classic mistake in using the Golang channel + goroutine.  When a goroutine accesses data in the closure environment, we should make a copy of the data for the goroutine. 

See the example in [effective_go#channels](https://go.dev/doc/effective_go#channels)

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -842,33 +842,7 @@ func (s *EtcdServer) run() {
 			f := func(context.Context) { s.applyAll(&ep, &ap) }
 			sched.Schedule(f)
 		case leases := <-expiredLeaseC:
-			s.GoAttach(func() {
-				// Increases throughput of expired leases deletion process through parallelization
-				c := make(chan struct{}, maxPendingRevokes)
-				for _, lease := range leases {
-					select {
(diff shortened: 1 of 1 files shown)


serathius (approved):
Great find! Did you find it yourself or used static analysis. Would be great if we could scan the codebase for more cases like this.

Still I would prefer if we just stop using anonymous functions. Could you maybe refactor this to call a function?

ahrtr:
I just found it during code review on the lease related code. I am investigate how can we refactor lease. I have a bias for action, so submit a PR to fix it immediately when I see the issue. 

Probably some linter tools can discover such issue, and it's in my to do list. 

Just moved the code into a separate method, and it looks much better now.

ahrtr:
cc @spzala as well. This is a minor change, and it should be safe.

ptabor (approved):

ptabor:
Nice finding. Benefits of final variables in Java,

ptabor:
Neither `intelij` nor `staticcheck` is reporting this as problem:
```(cd server && .../go/bin/staticcheck ./...)  | grep server.go```

PR #13576, opened 2021-12-31
Enhance lease keep-alive command for the case when the etcd server isn't reachable

ahrtr:
Fix [issues/13573](https://github.com/etcd-io/etcd/issues/13573).

Actually there are two issues here. 

The first issue is that the command something like `etcdctl lease keep-alive 32697e065bb41204 --once ` will be blocked forever when the etcd server isn't reachable. The reason is that the client side keeps waiting for the gRPC to be ready, Please see [options.go#L29](https://github.com/etcd-io/etcd/blob/69279532f46cb7491bb60a588be534100fdc058d/client/v3/options.go#L29). 

The second issue is just like the issue [issues/13573](https://github.com/etcd-io/etcd/issues/13573) raises. When the etcd server isn't reachable, the error message is something like "`lease 32697e065bb41204 expired or revoked`", which is really confusing. 

This PR is try to fix both above issues. 

@lilic  @yudai @heyitsanthony @serathius

PR #14076, opened 2022-05-26
clientv3: Fixed clientv3.Config DialTimeout usage

zjj:
if DialTimeout > 0, grpc.WithBlock() opt shall be set

### DIFF
--- a/client/v3/client.go
+++ b/client/v3/client.go
@@ -299,7 +299,9 @@ func (c *Client) dial(creds grpccredentials.TransportCredentials, dopts ...grpc.
 	if c.cfg.DialTimeout > 0 {
 		var cancel context.CancelFunc
 		dctx, cancel = context.WithTimeout(c.ctx, c.cfg.DialTimeout)
-		defer cancel() // TODO: Is this right for cases where grpc.WithBlock() is not set on the dial options?
+		defer cancel()
+
+		opts = append(opts, grpc.WithBlock())
 	}
 	target := fmt.Sprintf("%s://%p/%s", resolver.Schema, c, authority(c.endpoints[0]))
 	conn, err := grpc.DialContext(dctx, target, opts...)


zjj:
```Golang
package main

import (
	"context"
	"fmt"
	"log"
	"time"

)

var (
	dialTimeout = 5 * time.Second

	endpoints = []string{"localhost:2379", "localhost:22379", "localhost:32379"}
)

func main() {
	cli, err := clientv3.New(clientv3.Config{
		Endpoints:   endpoints,
		DialTimeout: dialTimeout,
	})
	if err != nil {
		log.Fatal(err)
	}
	defer cli.Close()

	ch := cli.Watch(context.TODO(), "x", clientv3.WithPrefix())  // here would block
	for {
		es := <-ch
		for _, e := range es.Events {
			fmt.Println(e.Kv)
		}
	}
}
```

```Golang
ch := cli.Watch(context.TODO(), "x", clientv3.WithPrefix())  // here would block
```

ahrtr:
Do you see any real issue? 

Please note that it's a non-blocking dial by default and the connecting happens in the background, and it should be OK.  It doesn't matter whether adding the `grpc.WithBlock() `option. But your change breaks the existing test. So I'd like keep it as it's for now. Of course, we can remove the [confusing comment](https://github.com/etcd-io/etcd/blob/908faa44c4c49d783c384481369895770fc2cff5/client/v3/client.go#L302). 

Test with your change:
```
$ ./etcdctl  get k1 v1 --dial-timeout=5s
Error: context deadline exceeded
```

Test without your chanage:
```
$ ./etcdctl  get k1 v1 --dial-timeout=5s
{"level":"warn","ts":"2022-05-27T06:03:00.156+0800","logger":"etcd-client","caller":"v3/retry_interceptor.go:64","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0001d65a0/127.0.0.1:2379","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"transport: Error while dialing dial tcp 127.0.0.1:2379: connect: connection refused\""}
Error: context deadline exceeded
```

zjj:
> Do you see any real issue?
> 
> Please note that it's a non-blocking dial by default and the connecting happens in the background, and it should be OK. It doesn't matter whether adding the `grpc.WithBlock()`option. But your change breaks the existing test. So I'd like to keep it as it's for now. Of course, we can remove the [confusing comment](https://github.com/etcd-io/etcd/blob/908faa44c4c49d783c384481369895770fc2cff5/client/v3/client.go#L302).
> 
> Test with your change:
> 
> ```
> $ ./etcdctl  get k1 v1 --dial-timeout=5s
> Error: context deadline exceeded
> ```
> 
> Test without your chanage:
> 
> ```
> $ ./etcdctl  get k1 v1 --dial-timeout=5s
> {"level":"warn","ts":"2022-05-27T06:03:00.156+0800","logger":"etcd-client","caller":"v3/retry_interceptor.go:64","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0001d65a0/127.0.0.1:2379","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"transport: Error while dialing dial tcp 127.0.0.1:2379: connect: connection refused\""}
> Error: context deadline exceeded
> ```

plese try this command
./etcdctl watch a --prefix  --dial-timeout=1s without etcd running.  
maybe it should exit with error.
 

```golang
      cli, err := clientv3.New(clientv3.Config{
		Endpoints:   endpoints,
		DialTimeout: dialTimeout,
	})
	if err != nil {
		log.Fatal(err)
	}
	defer cli.Close()

	ch := cli.Watch(context.TODO(), "x", clientv3.WithPrefix())  // here would block
```

ahrtr:
Got it, actually `etcdctl lease keep-alive` will be blocked forever as well if the etcd server isn't running. See [pull/13576](https://github.com/etcd-io/etcd/pull/13576)

This PR might change the user behavior. Previously, commands something like `etcdctl lease keep-alive` and "etcdctl watch" are blocked forever until the etcd server is reachable. It means users can kick off the command firstly, and start etcd server sometime later. After your change, it doesn't work any more, because the etcd server must be started firstly. 

Anyway, please fix the test failures firstly.

zjj:
```go
  cli, err := clientv3.New(clientv3.Config{
		Endpoints:   endpoints,
		DialTimeout: dialTimeout,
	})
  if err != nil {
     log.Fatal(err)
  }

  ch := cli.Watch(context.TODO(), "x", clientv3.WithPrefix())
  for {
    select {
    case r := <-ch:
      if len(r.Events) > 0 {
       doBussiness0()
    case <-time.After(time.Second * 10):
       doBussiness1()
    }
  }
```
Above is my scenario, if there's no event from etct for 10 seconds, doBussiness1(),  if the server not readable when codes running,  it blocks at ```cli.Watch```, even if the server is reachable later,  it still blocks there.

But if server is ok at the beginning, it's still accepted while server stops at sometime after the codes enter for{} loop block.
I am not sure if there's a proper way to handle the situation that the server is not readable at the beginning.

ahrtr:
> DialOptions: []grpc.DialOption{grpc.WithBlock()} shall be assigned to Config

Yes, in your case, I think you can update your code something like below,
```
  cli, err := clientv3.New(clientv3.Config{
		Endpoints:   endpoints,
		DialTimeout: dialTimeout,
                DialOptions: []grpc.DialOption{grpc.WithBlock()},
	})
```

ahrtr:
> > DialOptions: []grpc.DialOption{grpc.WithBlock()} shall be assigned to Config
> 
> Yes, in your case, I think you can update your code something like below,
> 
> ```
>   cli, err := clientv3.New(clientv3.Config{
> 		Endpoints:   endpoints,
> 		DialTimeout: dialTimeout,
>                 DialOptions: []grpc.DialOption{grpc.WithBlock()},
> 	})
> ```

Please let me know whether this works for you or not.

zjj:
> > > DialOptions: []grpc.DialOption{grpc.WithBlock()} shall be assigned to Config
> > 
> > 
> > Yes, in your case, I think you can update your code something like below,
> > ```
> >   cli, err := clientv3.New(clientv3.Config{
> > 		Endpoints:   endpoints,
> > 		DialTimeout: dialTimeout,
> >                 DialOptions: []grpc.DialOption{grpc.WithBlock()},
> > 	})
> > ```
> 
> Please let me know whether this works for you or not.

Yes， it works.
maybe， I should not be compulsive on  the default action of ```Dialtimeout``` ：）
shall we close this pr ?

ahrtr:
Thanks for the confirmation.

PR #14078, opened 2022-05-27
Clean up the local variable cps so that it will not continue to schedule checkpoint

ahrtr:
Fix a minor issue, which was discovered during code review. 

```
// Clean up the cps, otherwise it will continue to schedule
// checkpoints even after it steps down the leader position.
```

### DIFF
--- a/server/lease/lessor.go
+++ b/server/lease/lessor.go
@@ -646,10 +646,10 @@ func (le *lessor) revokeExpiredLeases() {
 // checkpointScheduledLeases finds all scheduled lease checkpoints that are due and
 // submits them to the checkpointer to persist them to the consensus log.
 func (le *lessor) checkpointScheduledLeases() {
-	var cps []*pb.LeaseCheckpoint
-
 	// rate limit
 	for i := 0; i < leaseCheckpointRate/2; i++ {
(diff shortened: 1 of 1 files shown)


ahrtr:
cc @serathius @ptabor @spzala another minor fix

server/lease/lessor.go
```diff
@@ -669,6 +669,9 @@ func (le *lessor) checkpointScheduledLeases() {
 		if len(cps) < maxLeaseCheckpointBatchSize {
 			return
 		}
+		// Clean up the cps, otherwise it will continue to schedule
+		// checkpoints even after it steps down the leader position.
+		cps = []*pb.LeaseCheckpoint{}
```
ptabor, line 674: Is there a reason to not make `cps` a local variable within the loop, so declare it in line 660 instead of 656 ?

ahrtr: Ah:), good catch. Thanks & resolved.

spzala (approved):
lgtm Thanks @ahrtr

ahrtr:
Thanks @spzala . It's very minor change, and should be safe, so will merge it.

PR #14080, opened 2022-05-29
Fix potential deadlock between Revoke and (Grant or Checkpoint)

ahrtr:
The [Grant](https://github.com/etcd-io/etcd/blob/ce77d83ee6cd2ea812c0e1336959c3552fc51a80/server/lease/lessor.go#L271) and [Checkpoint](https://github.com/etcd-io/etcd/blob/ce77d83ee6cd2ea812c0e1336959c3552fc51a80/server/lease/lessor.go#L361) acquire the `lessor.mu` firstly, and acquire `batchTx lock` secondly when they are still holding `lessor.mu`.

But the [Revoke](https://github.com/etcd-io/etcd/blob/ce77d83ee6cd2ea812c0e1336959c3552fc51a80/server/lease/lessor.go#L321) acquires the `transaction lock `firstly, and secondly acquires the `lessor.mu` lock when it's still holding the transaction lock.  

Obviously there is a potential deadlock between Revoke and (Grant or Checkpoint).

### DIFF
--- a/server/lease/lessor.go
+++ b/server/lease/lessor.go
@@ -286,15 +286,15 @@ func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
 		revokec: make(chan struct{}),
 	}
 
+	if l.ttl < le.minLeaseTTL {
+		l.ttl = le.minLeaseTTL
+	}
+
 	le.mu.Lock()
 	defer le.mu.Unlock()
 
 	if _, ok := le.leaseMap[id]; ok {
 		return nil, ErrLeaseExists
-	}
-
-	if l.ttl < le.minLeaseTTL {
-		l.ttl = le.minLeaseTTL
 	}
 
 	if le.isPrimary() {
@@ -326,6 +326,12 @@ func (le *lessor) Revoke(id LeaseID) error {
 		le.mu.Unlock()
 		return ErrLeaseNotFound
 	}
+
+	// We shouldn't delete the lease inside the transaction lock, otherwise
+	// it may lead to deadlock with Grant or Checkpoint operations, which
+	// acquire the le.mu firstly and then the batchTx lock.
+	delete(le.leaseMap, id)
+
 	defer close(l.revokec)
 	// unlock before doing external work
 	le.mu.Unlock()
@@ -344,9 +350,6 @@ func (le *lessor) Revoke(id LeaseID) error {
 		txn.DeleteRange([]byte(key), nil)
 	}
 
-	le.mu.Lock()
-	defer le.mu.Unlock()
-	delete(le.leaseMap, l.ID)
 	// lease deletion needs to be in the same backend transaction with the
 	// kv deletion. Or we might end up with not executing the revoke or not
 	// deleting the keys if etcdserver fails in between.


ahrtr:
cc @serathius @ptabor @spzala

server/lease/lessor_test.go
```diff
@@ -150,11 +150,19 @@ func TestLeaseConcurrentKeys(t *testing.T) {
 	wg.Wait()
 }
 
+func TestLessorRevoke(t *testing.T) {
+	lessorRevoke(t, true)
+}
+
+func TestLessorRevokeWithoutRangeDeleter(t *testing.T) {
```
serathius, line 157: Problem with adding this test is fact that in any correct configuration `RangeDeleter` should only be set. `RangeDeleter` can be nil when creating Lessor and set later just to avoid circular dependencies in bootstap and allow changing reference to backend during recovery. 
We should rather work to make the

serathius, line 157: Problem with adding this test is fact that in any correct configuration `RangeDeleter` should only be set. `RangeDeleter` can be nil when creating Lessor and set later just to avoid circular dependencies in bootstap and allow changing reference to backend during recovery. We should rather work to make the

ahrtr: Yes, this is a valid comment. I rolled back the change for the test case for now. Let's address the task of `removing circular dependencies` in separate PR if needed.

serathius: SG

server/lease/lessor.go
```diff
@@ -326,6 +326,12 @@ func (le *lessor) Revoke(id LeaseID) error {
 		le.mu.Unlock()
 		return ErrLeaseNotFound
 	}
+
+	// We shouldn't delete the lease inside the transaction lock, otherwise
+	// it may lead to deadlock with Grant or Checkpoint operations, which
+	// acquire the le.mu firstly and then the batchTx lock.
```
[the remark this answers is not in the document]
serathius: I agree that there is a chance of deadlock here, however I'm worried about removing lease from leaseMap and releasing lock before db operation is executed. This might be totally unfounded worry as backend operation as flushed only once 5 seconds, however I need to think a little on it. Have you thought on it more?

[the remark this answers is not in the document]
ahrtr: It should be OK. The` le.mu` is used to protect the data structures inside `lessor`, and the backend transaction is protected by separate transaction lock.  Just as I mentioned previously, there is no reason to hold the `le.mu` during the transaction. 

Please note that during `Grant` and `Checkpoint`, lessor may also persist the lease, but both of them will require the transaction lock before persisting the lease data.  So it's safe.

serathius (approved):

server/lease/lessor.go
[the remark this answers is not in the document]
chaochn47: For my understanding, 

before the change 

```
Grant is (called Process 1)
1. acquires lessor.mu R1 (called Resource 1)
2. acquires backend tx R2
3. releases backend tx R2
4. releases lessor.mu R1

Revoke P2
1. acquires lessor.mu R1
2. releases lessor.mu R1
3. acquires backend tx R2
4. acquires lessor.mu R1
5. releases backend tx R2
6. releases lessor.mu R1
```

There is a potential deadlock for sure
- P1 is holding R1 and wait to acquire R2
- P2 is holding R2 and wait to acquire R1

Why not we move P2 step 4 to happen after step 5? Namely `delete(le.leaseMap, id)` after `txn.End()` which releases the backend transaction.

PR #14082, opened 2022-05-30
tools/benchmark: migrate cheggaaa/pb.v1 to cheggaaa/pb/v3

mikelolasagasti:
This commit also changes the format of the progress bar to use a default one
provided by the library rather than previous custom 'Booooom !' one.


spzala (commented):
@mikelolasagasti welcome and thanks for your contribution! Can you split this PR into two - one for the cheggaaa update and other one for format?. For the later, it would be great if you can also share how the new default pb look. Also, I think etcdctl/ctlv3/command/check.go should be updated for cheggaaa as well. Thanks!

mikelolasagasti:
Thanks for reviewing @spzala! 

I added missing `etcdctl/ctlv3/command/check.go` and also tidied the tests go.mod/sub. 

About a second PR for changing the format, the problem is that v3 doesn't accept .Format("yourformat") and requires using a template, so either I use the template to keep previous format or just remove and don't use anything custom for the progress bar and avoid breaking between PRs or extra work. If you still think that is required I'll do it, but I don't see the benefit.

ahrtr:
Please squash the commits.

spzala:
> Thanks for reviewing @spzala!
> 
> I added missing `etcdctl/ctlv3/command/check.go` and also tidied the tests go.mod/sub.
> 
> About a second PR for changing the format, the problem is that v3 doesn't accept .Format("yourformat") and requires using a template, so either I use the template to keep previous format or just remove and don't use anything custom for the progress bar and avoid breaking between PRs or extra work. If you still think that is required I'll do it, but I don't see the benefit.

@mikelolasagasti thank you! And I see, in that case, it's fine to keep format changes here but can you please share the output of any benchmark to see how default bar look. I agree that if new default progress bar is good then we should not keep and maintain the custom one.

mikelolasagasti:
@ahrtr squashed the commits, not sure if I documented it correctly in the commit log

@spzala I documented the change as part of the commit. Maybe you referred to the PR, if so I can change it.

spzala:
> @ahrtr squashed the commits, not sure if I documented it correctly in the commit log
> 
> @spzala I documented the change as part of the commit. Maybe you referred to the PR, if so I can change it.

@mikelolasagasti I see it in the commit message. Thanks!

spzala (approved):

ahrtr (approved):
LGTM. Thank you!

ahrtr:
cc @ptabor @serathius

serathius (dismissed):
Any reason why we are picking additional dependencies? I don't think we should care about upgrading progress bar lib, however additional dependencies are worrying.

I would not be so picky if this change impacted only `tool/benchmark`, however it impacts dependencies of `etcdctl` for which we care about. More dependencies = harder to manage, higher chance of security vuln, etc.

serathius:
> > Any reason why we are picking additional dependencies?
> 
> My understanding is this PR isn't meaning to introducing additional dependencies. It just bumps the version from `cheggaaa/pb.v1` to `cheggaaa/pb/v3`. However `cheggaaa/pb/v3` has more dependencies than `v1`, and it is exactly the additionally dependencies we see in this PR.

Dependencies of your dependencies (so called transitive dependencies) are also your dependencies. Point stands that upgrading from v1 to v3 is regression from dependency standpoint.

> The bigger concern is the repo `https://github.com/cheggaaa/pb` is out of maintenance, and it's still depending on golang 1.12. I think we should clean up all such repositories and find some better alternatives. Definitely we can address it in separate PR.

I think this is more important problem to solve. Is having nice progress bar worth picking up additional 4 dependencies. From etcd project perspective having too many dependencies is harmful, on the other hand progress bar doesn't matter. 

Please consider that those dependencies will need to be upgraded, patched for security issues, etc. It might be far from now, however someone will need to do it. Upgrading 4 dependencies doesn't seem as much work, however if we continue to make decisions like this and adding new 4 dependencies we will be overwhelmed sooner than later. Let's not leave future maintainers with this problem.

For tradeoff: upgrade progress bar library and pick up 4 new dependencies vs remove progress bar, I would rather get rid off progress bar.

spzala:
I wouldn't unnecessary like to add any new dependencies either and that was the fist thought came up when I reviewed first time. But my take here was to upgrade the version vs updating the bar. The new bar is a side outcome of newer version as discussed in this PR above. I believe progress bar is good to have in general and we shouldn't get rid of it. I am open to go with any approach here.

ptabor:
[removed the tag as the PR seems controversial] 

 - The fact that it's `etcdctl` (vs. e.g. `etcdclient`) is IMHO lowering the bar on dependencies. The dependencies are not going to be transitively propagated to all the software linking-in etcd client libraries, so there is no risk of conflict with the user's libraries.
 -  I think we should have some form of progress notion in the benchmarks and `etcdctl check` (even if it's `5/144`)
 -  I don't find this libraries as particularly warring... Probably transitive 'size' should be also indicator of problems rather than pure count. 
 - https://github.com/schollz/progressbar seems to be a little lighter and maintained. 

Summary: 
  - not strong opinion. Leaning towards keeping as is and optionally evaluating `schollz` progressbar as more up-to-date.

mikelolasagasti:
Using `github.com/schollz/progressbar/v3` also add new deps:

```
go: upgraded github.com/mattn/go-runewidth v0.0.9 => v0.0.13
go: added github.com/mitchellh/colorstring v0.0.0-20190213212951-d06e56a500db
go: added github.com/rivo/uniseg v0.2.0
go: added github.com/schollz/progressbar/v3 v3.8.6
go: upgraded golang.org/x/sys v0.0.0-20220114195835-da31bd327af9 => v0.0.0-20220128215802-99c3d69c2c27
go: upgraded golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 => v0.0.0-20210927222741-03fcf44c2211
```
Two of the deps required by `github.com/cheggaaa/pb/v3`, the change in this PR, are quite common:

- `github.com/mattn/go-isatty` is used by other tools like GitHub's `gh` cli, gopass or hugo.
- `github.com/fatih/color` is used by DigitalOcean's `doctl` cli, gopass or hashicorp's hclog.

The other two are less common:

- `github.com/rivo/uniseg` is required by `go-runewidth`, thus required in both options.
- `github.com/VividCortex/ewma`, a Exponentially Weighted Moving Average algorithms for Go, doesn't seem to be widely used.

If progress bar is not removed, both options have pros and cons, so I rather just go with the updated version of `cheggaaa/pb/`. 

If removing progress bar is decided, I can work in another PR.

Another option would be to remove the progress bar from `etcdctl`, that is something more users will use, and keep for the benchmark tool, that will be used mainly by developers.

ptabor:
Thank you. I'm fine with updating `cheggaaa/pb`, but final decision to @serathius

serathius (commented):
Looks like only alternative solution is to remove progress bar. I think long term it's important for etcd project to reduce maintenance cost by removing not needed features and focusing on the core functionality.
However dependencies don't cause too many problems now, so I don't want to block this PR when other maintainers think that dependencies are in ok state.

ptabor (approved):

PR #14084, opened 2022-06-01
mvcc: use the 'visit' to implement the 'RangeSince' 

SimFG:
To make the meaning of the RangeSince function in treeIndex clearer


server/storage/mvcc/index.go
```diff
@@ -194,14 +194,9 @@ func (ti *treeIndex) RangeSince(key, end []byte, rev int64) []revision {
...
 
-	endi := &keyIndex{key: end}
 	var revs []revision
-	ti.tree.AscendGreaterOrEqual(keyi, func(item btree.Item) bool {
-		if len(endi.key) > 0 && !item.Less(endi) {
-			return false
-		}
-		curKeyi := item.(*keyIndex)
-		revs = append(revs, curKeyi.since(ti.lg, rev)...)
+	ti.visit(key, end, func(ki *keyIndex) bool {
```
ahrtr, line 198: It will lead to deadlock, because it will acquire the read lock inside the visit method as well. We can add an `unsafeVisit` method to resolve it.

ahrtr:
The `RangeSince` interface method is NOT used by any production code at all, it just only used by one unit test case `TestIndexRangeSince`. So probably we can just remove `RangeSince`, and update the unit case `TestIndexRangeSince` to use other method, such as [Revisions](https://github.com/etcd-io/etcd/blob/08f4c340eb92338d0f266795222d9ea04cb69056/server/storage/mvcc/index.go#L109)

SimFG:
Got it. I think the lock use in some methods is confused. For example, In the `Range` method, it will be better if using the lock at the beginning the method call, and we should remove the lock in the `visit` method and add a 'get' method without the lock that it is likely the `Get` method.

SimFG:
fail to the test. How to solve it? @ahrtr

SimFG:
All tests have passed. @ahrtr

ahrtr (approved):
LGTM. Thank you.

cc @ptabor @serathius @spzala 

I also suggest to remove `RangeSince` , see [issuecomment-1143485118](https://github.com/etcd-io/etcd/pull/14084#issuecomment-1143485118), which of course can be addressed in a separate PR.

server/storage/mvcc/index.go
```diff
@@ -67,9 +67,13 @@ func (ti *treeIndex) Put(key []byte, rev revision) {
 }
 
 func (ti *treeIndex) Get(key []byte, atRev int64) (modified, created revision, ver int64, err error) {
-	keyi := &keyIndex{key: key}
 	ti.RLock()
 	defer ti.RUnlock()
+	return ti.get(key, atRev)
+}
+
+func (ti *treeIndex) get(key []byte, atRev int64) (modified, created revision, ver int64, err error) {
```
spzala, line 75: @SimFG  I have few comments: 
1) Why we need to modify `Get`?. The new `get` method can be mistakenly used in future without setting lock..we can add some doc for `get` but even then why not just use `Get` as it's being used currently and we don't worry about every caller func to `get` set lock. 
2) I would prefer to have PR stick to what's in the description, that we are getting rid of few redundant lines of code by using `visit`. 
3) Can't we just use`visit` in `RangeSince` by only change in this PR to call `visit` and remove the two lines of lock call similar to how `visit` is used in the `Revisions`?  Thanks!

spzala:
> LGTM. Thank you.
> 
> cc @ptabor @serathius @spzala
> 
> I also suggest to remove `RangeSince` , see [issuecomment-1143485118](https://github.com/etcd-io/etcd/pull/14084#issuecomment-1143485118), which of course can be addressed in a separate PR.

@ahrtr thanks! It seems okay to remove `RangeSince` and may be separately to this PR as you said to have more review thoughts on it.

server/storage/mvcc/index.go
SimFG: 1. Because if doing like this, the lock will be only used in the exposed method, like `Put`, `KeyIndex`. And the internal method will be without the lock which makes these functions clearer.
2. Yes, I agree.
3. I have thought of doing this. When using `visit` to replace `AscendGreaterOrEqual`, the lock in this method should be moved into the `if` inside, which is strange. If removing the lock and using `ti.Get` like `Revisions`, the `keyi.since` isn't safe without the lock.

server/storage/mvcc/index.go
```diff
@@ -92,9 +96,6 @@ func (ti *treeIndex) keyIndex(keyi *keyIndex) *keyIndex {
 func (ti *treeIndex) visit(key, end []byte, f func(ki *keyIndex) bool) {
```
serathius, line 96: Note about convention. In etcd code we prefix functions with `unsafe` if functions requires external caller to lock mutex. This would mean that function `Get` -> `unsafeGet` and `visit` -> `unsafeVisit`.

serathius (approved):
With function rename code is much more readable and intention of this PR clear. Great work! Thanks!

server/storage/mvcc/index.go
serathius: With function renames it was much clearer for me what this PR is changing. Your right, this PR should not have any performance impact.

SimFG:
@serathius all tests have passed.

PR #14085, opened 2022-06-01
*: fix typo "iff" to "if" in comments

kaijchen:
Fixed typo "iff" to "if" in comments.

Searched for all occurences in the project by `grep -rw`.


spzala (commented):
@kaijchen thank you but "iff" is fine - it's an abbreviation for `if and only if` so I think we are good.  Closing it but please continue contributing to the project with other improvements. Thanks!

PR #14087, opened 2022-06-04
[3.5] Backport two lease related bug fixes to 3.5

ahrtr:
Backport [pull/14067](https://github.com/etcd-io/etcd/pull/14067) and [pull/14078](https://github.com/etcd-io/etcd/pull/14078) to 3.5.

The first bug fix is to resolve the race condition between goroutine and channel on the same leases to be revoked. It's a classic mistake in using Golang channel + goroutine. Please refer to [effective_go#channels](https://go.dev/doc/effective_go#channels).

The second bug fix is to resolve the issue that etcd lessor may continue to schedule checkpoint after stepping down the leader role.


serathius (approved):
Both fixes look good for backport.

PR #14104, opened 2022-06-10
Bump OpenTelemetry dependencies to v1.7.0

lucacome:
Bumps OpenTelemetry dependencies to v1.7.0


ahrtr:
This is the probably 4th attempt to bump the OpenTelemetry.  Note that all previous attempts were failed. It could be due to the behavior change on the new gRPC version. 

It would be helpful if you get it resolved.

lucacome:
@ahrtr I noticed unfortunately 😭 but I wanted to give it a try

lucacome:
I think one of the problems is
```
go.etcd.io/etcd/raft/v3/raftpb imports
	github.com/golang/protobuf/proto tested by
	github.com/golang/protobuf/proto.test imports
	github.com/google/go-cmp/cmp imports
	github.com/google/go-cmp/cmp/internal/value loaded from github.com/google/go-cmp@v0.5.5,
	but go 1.16 would select v0.5.6

To upgrade to the versions selected by go 1.16:
	go mod tidy -go=1.16 && go mod tidy -go=1.17
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
For other options, see:
	https://golang.org/doc/modules/pruning
```

ptabor:
Closing as it's done in https://github.com/etcd-io/etcd/pull/14125

lucacome:
@ptabor it looks like only grpc was updated, not otel? 🤔

PR #14106, opened 2022-06-13
wal: remove the repeated test case

SimFG:
The TestFilePipelineFailLockFile case maybe test the line 77 in the alloc function. But the LockFile will not throw an error when the file lock is hold. And TestFilePipelineFailPreallocate and TestFilePipelineFailLockFile are exactly the same except for the comments.

### DIFF
--- a/server/storage/wal/file_pipeline.go
+++ b/server/storage/wal/file_pipeline.go
@@ -58,6 +58,7 @@ func newFilePipeline(lg *zap.Logger, dir string, fileSize int64) *filePipeline {
 
 // Open returns a fresh file for writing. Rename the file before calling
 // Open again or there will be file collisions.
+// it will 'block' if the tmp file lock is already taken.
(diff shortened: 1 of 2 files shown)


ptabor (changes requested):
Thank you for finding. Indeed this piece of code is not adding any value. 

Could you, please, lock manually the file `0.tmp` in the preamble of the test, such that: 
https://github.com/etcd-io/etcd/blob/4592af25ba3b9b9d249845d7d376fe0f7c5ba9de/server/storage/wal/file_pipeline.go#L77 would fail.

ptabor (approved):
Thank you for checking.

Optionally: Please add a comment to fp.Open(...) method that it will 'block' if the lock is already taken.

SimFG:
@ptabor I have commented in this method

serathius (approved):

PR #14107, opened 2022-06-13
Get all constant etcd versions defined in a centralized place

ahrtr:
Currently the etcd versions are defined in everything in the repo, it isn't good.  In this PR, we get all constant versions defined in a centralized places, so that they can be reused by all other places. 

cc @serathius @spzala @ptabor


ptabor (approved):
Thank you for the cleanup.

serathius (approved):

PR #14109, opened 2022-06-13
schedule: Provide logs when the fifo job panic happens

SimFG:
To make the fifo scheduler better debuggability.

### DIFF
--- a/pkg/schedule/schedule.go
+++ b/pkg/schedule/schedule.go
@@ -17,9 +17,36 @@ package schedule
 import (
 	"context"
 	"sync"
+
+	"go.etcd.io/etcd/client/pkg/v3/verify"
+
+	"go.uber.org/zap"
 )
 
-type Job func(context.Context)
+type Job interface {
+	Name() string
+	Do(context.Context)
+}
+
+type job struct {
+	name string
+	do   func(context.Context)
+}
+
+func (j job) Name() string {
+	return j.name
+}
+
+func (j job) Do(ctx context.Context) {
+	j.do(ctx)
+}
+
+func NewJob(name string, do func(ctx context.Context)) Job {
+	return job{
+		name: name,
+		do:   do,
+	}
+}
 
 // Scheduler can schedule jobs.
 type Scheduler interface {
@@ -56,14 +83,18 @@ type fifo struct {
 
 	finishCond *sync.Cond
 	donec      chan struct{}
+	lg         *zap.Logger
 }
 
 // NewFIFOScheduler returns a Scheduler that schedules jobs in FIFO
 // order sequentially
-func NewFIFOScheduler() Scheduler {
+func NewFIFOScheduler(lg *zap.Logger) Scheduler {
+	verify.Assert(lg != nil, "the logger should not be nil")
+
 	f := &fifo{
 		resume: make(chan struct{}, 1),
 		donec:  make(chan struct{}, 1),
+		lg:     lg,
 	}
 	f.finishCond = sync.NewCond(&f.mu)
 	f.ctx, f.cancel = context.WithCancel(context.Background())
@@ -125,7 +156,6 @@ func (f *fifo) Stop() {
 }
 
 func (f *fifo) run() {
-	// TODO: recover from job panic?
 	defer func() {
 		close(f.donec)
 		close(f.resume)
@@ -149,17 +179,29 @@ func (f *fifo) run() {
 				f.mu.Unlock()
 				// clean up pending jobs
 				for _, todo := range pendings {
-					todo(f.ctx)
+					f.executeJob(todo, true)
 				}
 				return
 			}
 		} else {
-			todo(f.ctx)
+			f.executeJob(todo, false)
+		}
+	}
+}
+
+func (f *fifo) executeJob(todo Job, updatedFinishedStats bool) {
+	defer func() {
+		if !updatedFinishedStats {
 			f.finishCond.L.Lock()
 			f.finished++
 			f.pendings = f.pendings[1:]
 			f.finishCond.Broadcast()
 			f.finishCond.L.Unlock()
 		}
-	}
+		if err := recover(); err != nil {
+			f.lg.Panic("execute job failed", zap.String("job", todo.Name()), zap.Any("panic", err))
+		}
+	}()
+
+	todo.Do(f.ctx)
 }
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -54,7 +54,7 @@ import (
 	"go.etcd.io/etcd/raft/v3/raftpb"
 	"go.etcd.io/etcd/server/v3/auth"
 	"go.etcd.io/etcd/server/v3/etcdserver/api"
-	"go.etcd.io/etcd/server/v3/etcdserver/api/etcdhttp/types"
+	httptypes "go.etcd.io/etcd/server/v3/etcdserver/api/etcdhttp/types"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
@@ -744,7 +744,7 @@ func (s *EtcdServer) run() {
 	}
 
 	// asynchronously accept toApply packets, dispatch progress in-order
-	sched := schedule.NewFIFOScheduler()
+	sched := schedule.NewFIFOScheduler(lg)
 
 	var (
 		smu   sync.RWMutex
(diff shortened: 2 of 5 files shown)


pkg/schedule/schedule.go
```diff
@@ -149,17 +156,29 @@ func (f *fifo) run() {
...
+		if !ctxDone {
 			f.finishCond.L.Lock()
 			f.finished++
 			f.pendings = f.pendings[1:]
 			f.finishCond.Broadcast()
 			f.finishCond.L.Unlock()
 		}
-	}
+		if err := recover(); err != nil {
+			f.lg.Warn("execute job failed", zap.Any("job", todo), zap.Any("err", err))
```
serathius, line 179: What do you expect will be printed for `job` key? Just guessing that it will be pointer to function. In most cases we pass anonymous function to fifo. 

Maybe we could add a proper name to jobs scheduled? Could be a separate PR

SimFG: I want to give some tips when the panic happens. But I don’t have a good idea. Maybe deleting the job key is better and we add a proper name to jobs in a separate pr. What do you think of this?

pkg/schedule/schedule.go
```diff
@@ -56,14 +58,20 @@ type fifo struct {
...
 	finishCond *sync.Cond
 	donec      chan struct{}
+	lg         *zap.Logger
 }
 
 // NewFIFOScheduler returns a Scheduler that schedules jobs in FIFO
 // order sequentially
-func NewFIFOScheduler() Scheduler {
+func NewFIFOScheduler(lg *zap.Logger) Scheduler {
+	if lg == nil {
```
ptabor, line 67: We decided to `verify.Assert` (client/pkg/verify/verify.go) whether the logger is prided rather than silently turn off logging if it was not provided.

pkg/schedule/schedule.go
```diff
@@ -149,17 +156,29 @@ func (f *fifo) run() {
...
 				return
 			}
 		} else {
-			todo(f.ctx)
+			f.executeJob(todo, false)
+		}
+	}
+}
+
+func (f *fifo) executeJob(todo Job, ctxDone bool) {
```
ptabor, line 169: nit: Instead of `ctxDone` I would call it `updatedFinishedStats` as we should talk about meaning of the parameter, and not why we had 'set' it.

pkg/schedule/schedule.go
```diff
@@ -149,17 +156,29 @@ func (f *fifo) run() {
...
+	defer func() {
+		if !ctxDone {
 			f.finishCond.L.Lock()
 			f.finished++
 			f.pendings = f.pendings[1:]
 			f.finishCond.Broadcast()
 			f.finishCond.L.Unlock()
 		}
-	}
+		if err := recover(); err != nil {
```
ptabor, line 178: I have a doubts about the `catching exceptions and swallowing errors` behavior.
If one of the periodic book-keeping processes within DB server fails unexpectedly, it might be caused by assertion that is supposed to detect data inconsistency or other serious problem (e.g. during compaction).   The earlier we detect it and avoid propagation (by e.g. reading checkpoint state from the other replica), the lowest chances we will make this 'broken' state authoritative and the easier to debug the problem. 

How about expanding the logging (as you did), flushing the logs, but applying this pattern: 

https://stackoverflow.com/questions/34842322/can-i-recover-from-panic-handle-the-error-then-panic-again-and-keep-the-origin

instead of the `recover` ?

We should look at this error 1 by 1 and decide whether they are recoverable or not.

ptabor (changes requested):

pkg/schedule/schedule.go
SimFG: Got it. About the `recover`, I write a demo.
image

image

So I think the recover method can be used. So I plan to add a `HandlePanic` func in the `Job` struct. When the `HandlePanic` isn't nil, it will be called when the panic happens.

pkg/schedule/schedule.go
```diff
@@ -149,17 +160,34 @@ func (f *fifo) run() {
...
+		if !updatedFinishedStats {
 			f.finishCond.L.Lock()
 			f.finished++
 			f.pendings = f.pendings[1:]
 			f.finishCond.Broadcast()
 			f.finishCond.L.Unlock()
 		}
-	}
+		if err := recover(); err != nil {
+			f.lg.Warn("execute job failed", zap.String("job", todo.Name), zap.Any("err", err))
```
ptabor, line 183: I would make it Error... not Warn, as this type of error "used" to panic the server.

pkg/schedule/schedule.go
```diff
@@ -149,17 +160,34 @@ func (f *fifo) run() {
...
+			if todo.HandlePanic != nil {
+				todo.HandlePanic(err)
+			} else {
+				panic(err)
```
ptabor, line 187: lg.Panic will trigger flushing logs, so is recommended from debuggability standpoint.

pkg/schedule/schedule.go
```diff
@@ -17,15 +17,23 @@ package schedule
...
 	"context"
 	"sync"
+
+	"go.etcd.io/etcd/client/pkg/v3/verify"
+
+	"go.uber.org/zap"
 )
 
-type Job func(context.Context)
+type Job struct {
```
ptabor, line 26: no strong opinion, but probably I would make it an interface... 

and have a:
   `NewJob("foo", func() {...} )` constructor...
   `NewRecoveringJob("boo", func() {...}, func() { ... } );

pkg/schedule/schedule.go
ptabor: Thank you. This seems good overall. 

Minor doubt I have... is what's the benefit of having separate HandlePanic(...) if Job author could actually already handle it within the Job.Do itself.

SimFG: Thanks your reading. It seems that the `HandlePanic ` method is a little redundant.

pkg/schedule/schedule.go
SimFG: Thanks for your advice.

server/etcdserver/server.go
```diff
@@ -839,7 +839,7 @@ func (s *EtcdServer) run() {
 	for {
 		select {
 		case ap := <-s.r.apply():
-			f := func(context.Context) { s.applyAll(&ep, &ap) }
+			f := schedule.NewJob("server applyAll", func(context.Context) { s.applyAll(&ep, &ap) })
```
ptabor, line 842: nit: I would avoid using 'space' in the identifiers. 

I would prefer any of:  "server.applyAll" or "server:applyAll" or "server_applyAll" as the convention for naming.   If we ever used this IDs for monitoring metric's field, it would also make things cleaner.

server/storage/mvcc/kvstore.go
```diff
@@ -202,7 +202,7 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
 	s.revMu.Lock()
 	if rev <= s.compactMainRev {
 		ch := make(chan struct{})
-		f := func(ctx context.Context) { s.compactBarrier(ctx, ch) }
+		f := schedule.NewJob("kvstore updateCompactRev compactBarrier", func(ctx context.Context) { s.compactBarrier(ctx, ch) })
```
ptabor, line 205: ditto.

server/storage/mvcc/kvstore.go
```diff
@@ -225,7 +225,7 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
 
 func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-chan struct{}, error) {
 	ch := make(chan struct{})
-	var j = func(ctx context.Context) {
+	j := schedule.NewJob("kvstore compact", func(ctx context.Context) {
```
ptabor, line 228: ditto

ptabor (approved):
Thank you. Apart of naming convention, looks good for merging.

server/etcdserver/server.go
SimFG: I have changed it.

serathius (approved):

pkg/schedule/schedule.go
```diff
@@ -149,17 +179,29 @@ func (f *fifo) run() {
...
+		if !updatedFinishedStats {
 			f.finishCond.L.Lock()
 			f.finished++
 			f.pendings = f.pendings[1:]
 			f.finishCond.Broadcast()
 			f.finishCond.L.Unlock()
 		}
-	}
+		if err := recover(); err != nil {
+			f.lg.Panic("execute job failed", zap.String("job", todo.Name()), zap.Any("err", err))
```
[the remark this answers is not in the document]
serathius: Have you tested how this log will look like?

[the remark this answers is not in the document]
SimFG: image
 the `err` can't use the zap.Error(), because the `zap.Error()` need a param of the `error` type. In this code, the type of the `err` is `Any`.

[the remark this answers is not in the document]
serathius: ack, then maybe we shouldn't use `err`. Using it could potentially pass different type under, maybe instead use `panic` key?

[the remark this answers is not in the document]
SimFG: Good idea. I have changed it.

PR #14117, opened 2022-06-15
Extend .gitignore to ignore intelij config/generated files.

ptabor:


serathius:
I'm on the other side that contributor responsibility to configure their global git file ignores https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer and not project to collect all file extensions used by all contributors.

ptabor:
- We have already `.idea` folder in the .gitignore
- I don't think global settings are the solution as different projects have different philosophy. 
   It might make sense for convenience of contributors to store some dev-environments configs (like .idea or .eclipse) project files. I worked in companies that stored them to have consistent dev experience. 
- I'm not saying it's project responsibility to collect *all possible* file extensions, but if there are a few most common tools, why not to make it convenient for contributors to not require them to think about it.

spzala (approved):
Agree with both @ptabor and @serathius :) that it's contributors responsibility in general but for common developer tools we should be okay to do it for convenience. lgtm.

PR #14120, opened 2022-06-15
server: Implement compaction hash checking

serathius:
Signed-off-by: Marek Siarkowicz <siarkowicz@google.com>

cc @ptabor @ahrtr @spzala 
Implements https://github.com/etcd-io/etcd/issues/14039


serathius:
Based on experience on etcd data inconsistency announcement I think we can launch disabled by default with strong suggestion to enable this feature. I have added 2 flags one for enabling the feature and second that allows changing frequency.

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,83 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+	cm.mux.RLock()
+	lastRevisionChecked := cm.latestRevisionChecked
+	cm.mux.RUnlock()
+	i := 0
+	for _, hash := range hashes {
+		if hash.Revision <= lastRevisionChecked {
+			return hashes[:i]
+		}
+		i++
+	}
```
ahrtr, line 328:
```suggestion
	for i, hash := range hashes {
		if hash.Revision <= lastRevisionChecked {
			return hashes[:i]
		}
	}
```

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,83 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+	return hashes[:i]
```
ahrtr, line 329:
```suggestion
	return hashes
```

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,83 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+	hashes := cm.uncheckedRevisions()
+	// Assume that revisions are ordered from largest to smallest
+	for i, hash := range hashes {
+		peers := cm.hasher.PeerHashByRev(hash.Revision)
+		if len(peers) == 0 {
+			continue
+		}
+		peersChecked := 0
+		for _, p := range peers {
+			if p.resp == nil || p.resp.CompactRevision != hash.CompactRevision {
```
ahrtr, line 269: Shouldn't `p.resp.CompactRevision != hash.CompactRevision` be an issue, because compaction operation goes through raft?

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,83 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+				cm.lg.Warn(
+					"same compact revision then hashes must match",
+					zap.Int64("revision", hash.Revision),
+					zap.Int64("leader-compact-revision", hash.CompactRevision),
+					zap.Uint32("leader-hash", hash.Hash),
+					zap.Int64("follower-compact-revision", p.resp.CompactRevision),
+					zap.Uint32("follower-hash", p.resp.Hash),
+					zap.String("follower-peer-id", types.ID(id).String()),
+				)
+				cm.hasher.TriggerCorruptAlarm(id)
```
ahrtr, line 285: Two comments here:
1. How do you know that which member's data is corrupted? I think we should depend on quorum. For example, for a cluster with 3 members, if two members' hashes are equal, but the third one is different, then we can say the third one's data is corrupted.
2. Probably we should add a metrics as well, so that users can leverage their monitoring system (such as prometheus) to generate alarm automatically.

server/etcdserver/server.go
```diff
@@ -2209,12 +2211,30 @@ func (s *EtcdServer) monitorKVHash() {
...
 			lg.Warn("failed to check hash KV", zap.Error(err))
 		}
 	}
 }
 
+func (s *EtcdServer) monitorCompactHash() {
+	if !s.Cfg.CompactHashCheckEnabled {
+		return
+	}
+	t := s.Cfg.CompactHashCheckTime
```
ahrtr, line 2224: Should we apply a default value if it's 0 or should we disable it if it's 0?

ahrtr:
I think probably we should also add an API & command to expose the compact hashes, so that users can check the values themselves, and take action accordingly. For example, for a cluster of 3 members, if two members have the same compact hashes, but the third one has a different value. Then the user may want to replace the third member. 

I also suggested to add a metrics for this case, please see https://github.com/etcd-io/etcd/pull/14120#discussion_r915542551

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,83 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+				zap.Int64("revision", hash.Revision),
+			)
+			cm.mux.Lock()
+			if hash.Revision > cm.latestRevisionChecked {
+				cm.latestRevisionChecked = hash.Revision
+			}
+			cm.mux.Unlock()
+			cm.lg.Info("finished compaction hash check", zap.Int("number-of-hashes-checked", i+1))
+			return
+		}
```
ahrtr, line 307: Should we add a warning log message for the else case? 
```
if len(peers) == peersChecked  {
    ......
} else {
    // add warning log here
}
```

server/etcdserver/corrupt.go
serathius: Request to start compaction goes through raft, however compaction itself is done asynchronously in background.

server/etcdserver/corrupt.go
serathius: Both improvements would be great, however I'm again little worried to backport too big of a feature. Previous consistency checking didn't identify corrupted member nor it had monitoring. I think those are great ideas for feature graduation, but are not needed to address etcd v3.5 supportability as proposed orginally.

server/etcdserver/server.go
serathius: Added a validation to prevent negative value.

server/etcdserver/corrupt.go
serathius: Yep

server/etcdserver/corrupt.go
serathius: Or better to say that compaction window goes through raft, it's up to member to clean up compacted keys in background.

serathius:
> I think probably we should also add an API & command to expose the compact hashes, so that users can check the values themselves, and take action accordingly. For example, for a cluster of 3 members, if two members have the same compact hashes, but the third one has a different value. Then the user may want to replace the third member.

This implementation was meant to be limited to ensure it doesn't require additional API thus is backportable. We compact hashes are exposed via normal hashKV API, however API doesn't expose which revisions hashes where calculated for. This is ok as we will take a different approach when implementing it for v3.6.

serathius:
PTAL @ahrtr

server/etcdserver/corrupt.go
ahrtr: I am not convinced.  Each member is indeed performing the compaction in background or asynchronously, but it doesn't matter.  You get the peers' hash using a parameter `hash.Revision` (line 263), so the `resp.CompactRevision` must be equal to `hash.CompactRevision`, or the resp is nil. 

When a member receives the `getPeerHashKVs` request, it just gets the cached value (`hashStorage.hashes`), which is only updated when a compaction operation finishes. The point is each member's cache data must be exactly the same or a subset of other member values. The case of not matching (KeyValueHash.Revision is the same, but CompactRevision are different) will never happen.

server/etcdserver/corrupt.go
ahrtr: I agree that we can make any additional improvement on in the main branch.  The point here is that it might not be correct to raise an alarm here (line 276), because we are not sure which member's data is corrupted.

server/etcdserver/server.go
ahrtr: OK

server/etcdserver/corrupt.go
serathius: You right, but this is true for all data inconsistency checks. We can only hope that it's not the leader.

server/etcdserver/corrupt.go
ahrtr: My point is `p.resp.CompactRevision != hash.CompactRevision` will never be `true`, no matter it's real or pathological case. If somehow it's `true`, them something must be wrong and most likely it's a bug.

ahrtr: Re-reading your code change, it seems that it's possible for `p.resp.CompactRevision != hash.CompactRevision` to be `true`. 

When there is no any compaction historical data, then the member may calculate the hash itself. In this case, the `CompactRevision` might be different. 
[hash.go#L133](https://github.com/serathius/etcd/blob/0dc860e1cf12d4a89953c82cff14e20f50237ae6/server/storage/mvcc/hash.go#L133)

serathius: Agreed, I implemented it this way to ensure backward compatibility. Just remembered that it's needed to handle upgrade. During upgrade older members will return different CompactRevision.

Does it make sense?

ahrtr:
Overall looks good to me. The benefit for release-3.5 (after backporting this and one previous related PR) is that data inconsistency/corruption detection is enhanced to leverage the side effect of compaction operation. But users can only know the result by checking the log and alarm, which may not be user-friendly. 

I agree that we should try NOT to change the API in a stable release, so it's OK just to backport what we have done to 3.5, and try to do more enhancement in 3.6 (see https://github.com/etcd-io/etcd/pull/14120#issuecomment-1177193974 ).

server/embed/config.go
```diff
@@ -759,6 +765,10 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("setting experimental-enable-lease-checkpoint-persist requires experimental-enable-lease-checkpoint")
 	}
 
+	if cfg.ExperimentalCompactHashCheckTime <= 0 {
```
ahrtr, line 768: I think we should only apply this check when `ExperimentalCompactHashCheckEnabled` is `true`. In other words, we shouldn't check `ExperimentalCompactHashCheckTime` at all if `ExperimentalCompactHashCheckEnabled` is disabled.

tests/e2e/etcd_corrupt_test.go
```diff
@@ -99,39 +99,42 @@ func corruptTest(cx ctlCtx) {
...
+
+	err = epc.Procs[0].Restart()
+	assert.NoError(t, err)
+	_, err = cc.Compact(5, config.CompactOption{})
+	assert.NoError(t, err)
+	time.Sleep(checkTime * 11 / 10)
+	alarmResponse, err := cc.AlarmList()
+	assert.NoError(t, err, "error on alarm list")
+	// TODO: Investigate why MemberID is 0?
+	assert.Equal(t, []*etcdserverpb.AlarmMember{{Alarm: etcdserverpb.AlarmType_CORRUPT, MemberID: 0}}, alarmResponse.Alarms)
```
ahrtr, line 139: My immediate feeling is it (`MemberID is 0`) might be a potential bug. The possible reason could be the leader uses an invalid memberID when activating the `AlarmType_CORRUPT` alarm.

server/etcdserver/corrupt.go
ahrtr: OK. The only concern is sometimes it wastes the follower(s)'s CPU to calculate the hash, which isn't used by the leader at all.  But it should be very low possibility to run into this situation.

server/embed/config.go
ahrtr: ack

This isn't a big deal, feel free to keep it as it's.

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,84 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+				zap.Int("number-of-peers-checked", peersChecked),
+				zap.Int64("revision", hash.Revision),
+			)
+			cm.mux.Lock()
+			if hash.Revision > cm.latestRevisionChecked {
+				cm.latestRevisionChecked = hash.Revision
+			}
+			cm.mux.Unlock()
+			cm.lg.Info("finished compaction hash check", zap.Int("number-of-hashes-checked", i+1))
+			return
```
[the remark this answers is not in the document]
serathius: Technically it could as a corruption can technically self heal. For example imagine that there PUT requests that was not reflected on all members. It means that hashes between members will not match. However, sometime later comes a DELETE request on the same key and it is reflected. That would restore the hash between members and invalidate previous corruption. If we look through cached hashes, the latest one will match, even though the old ones don't. Does it mean etcd is corrupted? In my opinion no.

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,84 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+		}
+		cm.lg.Warn("skipped revision in compaction hash check; was not able to check all peers",
+			zap.Int("number-of-peers-checked", peersChecked),
+			zap.Int("number-of-peers", len(peers)),
+			zap.Int64("revision", hash.Revision),
```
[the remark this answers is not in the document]
serathius: There is a log above that list all peers that were successfully checked.

tests/e2e/etcd_corrupt_test.go
serathius: Sure, same issue affects also other check, that is not covered by e2e tests. Will send a separate PR to fix both.

serathius:
Fixed member id in https://github.com/etcd-io/etcd/pull/14272 and rebased this PR on it. This is ok as I expect we will want to backport MemberID fix.

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
ahrtr: I tried to intentionally create such situation in which previous hashes do not match and the latest hashes match,  but one member's data is indeed corrupted, but eventually I failed to create such situation. The root cause is etcd scans the whole KEY space starting from 0x00 to the specified `compactMainRev`; Specifically, the [last](https://github.com/etcd-io/etcd/blob/5fab045242bd6415eb6ea47a6d96ef27a30c1438/server/storage/mvcc/kvstore_compaction.go#L43) isn't set to the `prevCompactRev`. 

So in summary, the case `that the latest [compactionRev, rev] hashes match, but previous ones do NOT match` will never happen.  And I think it's more safe to scan the whole KEY space instead of starting from `prevCompactRev`.

server/etcdserver/corrupt.go
```diff
@@ -241,6 +252,84 @@ func (cm *corruptionMonitor) periodicCheck() error {
...
+				cm.hasher.TriggerCorruptAlarm(uint64(p.id))
+				cm.lg.Error("failed compaction hash check",
+					zap.Int64("revision", hash.Revision),
+					zap.Int64("leader-compact-revision", hash.CompactRevision),
+					zap.Uint32("leader-hash", hash.Hash),
+					zap.Int64("follower-compact-revision", p.resp.CompactRevision),
+					zap.Uint32("follower-hash", p.resp.Hash),
+					zap.String("follower-peer-id", p.id.String()),
+				)
+				return
```
ahrtr, line 284: I suggest not to return here; instead, we should continue to compare other peers' hash, and generate multiple alrams if needed.

ahrtr:
The `ID` is displayed as hex-decimal, but it's displayed as decimal in the `ERRORS`. I think we should fix this, of course in a separate PR. 

```
$ etcdctl  endpoint status -w table --cluster  
+------------------------+------------------+---------------+-----------------+---------+----------------+-----------+------------+-----------+------------+--------------------+--------------------------------+
|        ENDPOINT        |        ID        |    VERSION    | STORAGE VERSION | DB SIZE | DB SIZE IN USE | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX |             ERRORS             |
+------------------------+------------------+---------------+-----------------+---------+----------------+-----------+------------+-----------+------------+--------------------+--------------------------------+
|  http://127.0.0.1:2379 | 8211f1d0f64f3269 | 3.6.0-alpha.0 |           3.6.0 |   25 kB |          25 kB |      true |      false |         4 |         69 |                 69 |  memberID:10501334649042878790 |
|                        |                  |               |                 |         |                |           |            |           |            |                    |                 alarm:CORRUPT  |
| http://127.0.0.1:22379 | 91bc3c398fb3c146 | 3.6.0-alpha.0 |           3.6.0 |   25 kB |          25 kB |     false |      false |         4 |         69 |                 69 |  memberID:10501334649042878790 |
|                        |                  |               |                 |         |                |           |            |           |            |                    |                 alarm:CORRUPT  |
| http://127.0.0.1:32379 | fd422379fda50e48 | 3.6.0-alpha.0 |           3.6.0 |   25 kB |          25 kB |     false |      false |         4 |         69 |                 69 |  memberID:10501334649042878790 |
|                        |                  |               |                 |         |                |           |            |           |            |                    |                 alarm:CORRUPT  |
+------------------------+------------------+---------------+-----------------+---------+----------------+-----------+------------+-----------+------------+--------------------+--------------------------------+
```

server/etcdserver/corrupt.go
serathius: This is not consistent with periodic check which only raises one alarm even though there could be multiple members corrupted.

serathius:
ping @ahrtr

server/etcdserver/corrupt.go
ahrtr: Should we at least finish comparing all peers' hashes and generate a info or error log? I think it's helpful for 3.5, at least we can tell the hashes from log.

serathius: This will heavily complicate implementation. There is nothing in those logs that users cannot get themselves.

ahrtr: Imagine a situation for 3.5, when the compact hash check raises an alarm `CORRUPT`, then users must figure out manually which member's data is corrupted. If all member's revisions are exactly the same, and the corruption is just caused by some disk I/O bit flip or disk corruption. Then users must get all members' log, and find out all the log entries below, and compare the hashes themselves. 

```
{"level":"info","ts":"2022-07-26T15:26:01.386+0800","caller":"mvcc/kvstore_compaction.go:65","msg":"finished scheduled compaction","compact-revision":9,"took":"40.078µs","hash":1087596250}

```

If we finish comparing all peers' hashes here and print a info or error log, then the users can easily figure out which member's data is corrupted. 

Actually It just a minor code change, and I don't think it complicate the implementation. 

For 3.6, we can some additional enhancement, but for 3.5 the users can only use the log to do analysis themselves.

ahrtr: Of course, we can do it in a separate PR.

ahrtr (approved):
LGTM with a minor comment.

PR #14124, opened 2022-06-16
mvcc:add ut for Revisions/CountRevisions and remove RangeSince as it …

wayblink:
…is not used

Signed-off-by: wayblink <wayasxxx@gmail.com>

### DIFF
--- a/server/storage/mvcc/index.go
+++ b/server/storage/mvcc/index.go
@@ -15,7 +15,6 @@
 package mvcc
 
 import (
-	"sort"
 	"sync"
 
 	"github.com/google/btree"
@@ -29,7 +28,6 @@ type index interface {
 	CountRevisions(key, end []byte, atRev int64) int
 	Put(key []byte, rev revision)
 	Tombstone(key []byte, rev revision) error
-	RangeSince(key, end []byte, rev int64) []revision
 	Compact(rev int64) map[revision]struct{}
 	Keep(rev int64) map[revision]struct{}
 	Equal(b index) bool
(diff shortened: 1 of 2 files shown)


server/storage/mvcc/index.go
```diff
@@ -107,6 +105,9 @@ func (ti *treeIndex) unsafeVisit(key, end []byte, f func(ki *keyIndex) bool) {
 	})
 }
 
+// Revisions returns limited number of revisions from key(including) to end(excluding)
+// at the given rev. The returned slice is sorted in the order of key.
+// There is no limit if limit <= 0.
```
ptabor, line 110: It seems also important that 'total' is not capped by the limit and reflects all the revisions.

ahrtr: Comments:
1. `including` to `included`, and `excluding` to `excluded`?  This applies to the comment for method `CountRevisions` as well.
2. adding comment something like "The second return parameter `total` isn't capped by the `limit`, and reflects all the expected revisions".

I'd defer to @spzala to correct the syntax of the comment.

ahrtr:
Overall looks good to me. Thank you.

Leave @spzala to final decision.

server/storage/mvcc/index.go
wayblink: Thanks for review, update as you commented.

server/storage/mvcc/index.go
```diff
@@ -107,6 +105,9 @@ func (ti *treeIndex) unsafeVisit(key, end []byte, f func(ki *keyIndex) bool) {
 	})
 }
 
+// Revisions returns limited number of revisions from key(included) to end(excluded)
+// at the given rev. The returned slice is sorted in the order of key. There is no limit if limit <= 0.
+// The second return parameter total isn't capped by the limit, and reflects all the expected revisions.
```
ahrtr, line 110: Minor comment: replace total with `total`.

ahrtr (approved):

spzala:
> Overall looks good to me. Thank you.
> 
> Leave @spzala to final decision.

Thanks @ahrtr Looking right now.

spzala (approved):
One comment in-line but otherwise looks good. Thanks @wayblink

server/storage/mvcc/index.go
ahrtr: @wayblink  Please update per the comment. Either replace total with `total`, or just remove the parameter name per @spzala 's comment. 

I will merge this PR once it's updated and pipeline is green.

wayblink: Updated. Thanks a lot

Issue #14112, opened 2022-06-14
Use GRPC url standard 


serathius:
Etcd code and tests depend on In particular our tests depend a lot on `unix://localhost:port` url schema, while
according to grpc standard: https://grpc.github.io/grpc/cpp/md_doc_naming.html the supported names are: `unix:path` or `unix://absolute_path` (so `unix:///tmp/foo`). This issue blocks grpc upgrade as etcd tests will not pass until we stop using `unix://localhost:port`.

Context https://github.com/etcd-io/etcd/pull/14104

First issue that we need to address is stop depending on  `unix://localhost:port` schema in tests to unblock dependency upgrade. Later we should propose a plan to match grpc standard. I expect this might require multiple releases to avoid introducing breaking changes.

serathius:
@ptabor Thanks for fixing the connection format, any followups needed or we can close the issue?

serathius:
ping

ptabor:
It's good enough for now. I don't think we have cycles to review all URL schemes proactively.

PR #14125, opened 2022-06-16
Update to grpc-1.47 (and fix the connection-string format)

ptabor:
Minimal change required to upgrade to grpc-1.47. 

https://github.com/etcd-io/etcd/issues/14112. Connecting to `unix:localhost:port` instead of never officially supported `unix://localhost:port` (https://grpc.github.io/grpc/cpp/md_doc_naming.html)

### DIFF
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -234,7 +234,7 @@ func (sctx *serveCtx) registerGateway(opts []grpc.DialOption) (*gw.ServeMux, err
 	addr := sctx.addr
 	if network := sctx.network; network == "unix" {
 		// explicitly define unix network for gRPC socket support
-		addr = fmt.Sprintf("%s://%s", network, addr)
+		addr = fmt.Sprintf("%s:%s", network, addr)
 	}
 
 	opts = append(opts, grpc.WithDefaultCallOptions([]grpc.CallOption{


api/go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/stretchr/testify v1.7.0
 	google.golang.org/genproto v0.0.0-20211118181313-81c1377c94b1
-	google.golang.org/grpc v1.41.0
+	google.golang.org/grpc v1.47.0
```
ahrtr, line 12: The title mentions 1.48, but actually 1.47

spzala: :) I guess it's a typo in the PR title as the latest stable release is 1.47. lgmt.

ptabor: Good catch. Reworded.

ahrtr:
@ptabor Overall the PR looks good to me. But I don't understand why do you add one more directory `out`, and get some go.mod and go.sum files included in it?  I guess they might be just intermediate files generated during your debug?

ptabor:
> @ptabor Overall the PR looks good to me. But I don't understand why do you add one more directory `out`, and get some go.mod and go.sum files included in it? I guess they might be just intermediate files generated during your debug?

Removed it. It was produces by intelij. I've just added to it to `.gitignore` to avoid such problems... and it has happened.

ahrtr (approved):
LGTM

Issue #14135, opened 2022-06-18
3.4 pipeline is broken


ahrtr:
I just realized that the release 3.4 pipeline wasn't green since it's first day being generated in [13135](https://github.com/etcd-io/etcd/pull/13135) on Jun 24, 2021. 

Please **code freeze release-3.4 branch** until I fix all the pipeline errors.

cc @serathius @ptabor @spzala

ahrtr:
Resolved by [14136](https://github.com/etcd-io/etcd/pull/14136)

PR #14134, opened 2022-06-18
release-3.4: updated test workflow go version to 1.16

lavacat:

Issue #14105, opened 2022-06-11
Plans for v3.4.19 release


lavacat:
As per discussion during community meeting, this issue is to estimate potential work and timeline for 3.4.19 release.

3.4.18 release was on Oct 15, 2021. Here is the [table](https://docs.google.com/spreadsheets/d/1_eH50hLwi3NtqpnyDnSQF_5bVtfYGVaRE4glY12Ppqo/) of all commits from release-3.5 branch (except 'tests: *', 'scripts: *' and 'Merge pull request *') since Oct 6, 2021. I'm assuming that fixes were backported before that.

To proceed, let's agree on:

1. What version of go to use for 3.4? I suggest 1.17 since 1.15 is EOL. We can keep go.1.12 in go.mod for compatibility. See [1] and [2]. @ahrtr @ptabor @lilic @hexfusion
2. Priority of issues to be backported. Feel free to comment in the table above. @serathius @ahrtr 

related discussions about go version:
[1] https://github.com/etcd-io/etcd/issues/12840
[2] https://github.com/etcd-io/etcd/issues/13912

ahrtr:
Thanks @lavacat. Please see my comments below.

## Go version
1.17 is definitely not an option for release-3.4. One breaking change introduced in Golang 1.15 is the deprecation of the legacy behavior of treating the CommonName field on X.509 certificates as a host name when no Subject Alternative Names are present. The workaround is to add the value `x509ignoreCN=0` to the `GODEBUG` environment variable to re-enable it again. see [go1.15#commonname](https://go.dev/doc/go1.15#commonname).  Note that `GODEBUG=x509ignoreCN=0` flag is removed in Golang 1.17. It means applications which are still use the legacy CommonName field in certificate will run into issue, unless they update their certificates. 

So Golang 1.17 isn't accepted for 3.4. Instead, both Golang 1.15 & 1.16 are OK for 3.4. Since Golang supports N-2 versions (1.16, 1.17 and 1.18 for now), so Golang 1.16 is better than 1.15. We still need to make sure there is any impact on the existing applications which are using or depending on 3.4.

## Making the pipeline green is the priority
Just I mentioned in [issuecomment-1119280894](https://github.com/etcd-io/etcd/issues/13912#issuecomment-1119280894), the top priority is to fix all the test failures and  make the pipeline green. Please let's get this done before talking about the release plan for 3.4.19.

## Backport
[...]

Issue #12840, opened 2021-04-07
etcd Go maintenance plan


hexfusion:
In the past, it has been the policy to maintain (as long as possible) the same Go version per minor release. So giving the example of `release-3.4`  we have and continue to build new releases with Go 1.12.x.  I think there are some advantages to this including shielding etcd from possibly complicated performance regressions. In the past Go version iteration was on a much slower cadence than it is today. But now Go is iterating is quite fast and managing things like CVE on EOL Go versions can become unfortunate.

The Go team maintains 2[1] versions of Go which unfortunately puts our latest versions of etcd in an undesirable condition. While all users are free to build and compile etcd with any version of go they wish they do so without the upstream test cycles and validation/blessing of maintainers.

I think it would be worthwhile to consider adopting a policy that states within the supported versions of etcd efforts will be made to align the Go version we build and ship upstream etcd with supported versions of Go (within reason). This policy would come as a burden to maintainers as a group and should require explicit responsibilities from maintainers tasked with these maintenance tasks.

One thought is having a maintainer(s) responsible for bumping Go on the supported versions of etcd. The task would be to ensure CI is aligned with expectations and as a group look at integrating performance regression checks into CI. There should be no rush to bump etcd to a new Go release version but all attempts should be made to align with upstream Kubernetes where possible. I think moving forward 3.5+ this should be more straightforward but fairly large gaps still exist in 3.3 and 3.4 releases.

Let's talk about this further at the community meeting tomorrow?

cc @ptabor @gyuho @wenjiaswe @marun @lilic @retroflexer 

[1] https://golang.org/doc/devel/release.html#policy
[2] https://github.com/etcd-io/etcd/issues/12330
[3] https://github.com/etcd-io/etcd/issues/12732

ptabor:
I think we should update the CI or build binaries using the newer version, but at the same time I think we should cease using new features/libraries to would force our clients to upgrade go version. 

I.e. if etcd-3.4.0 was compilable&testable using go-1.12,  3.4.99 should remain go-1.12 compatible.

The problem would also be less significant if we have minor release more frequently (e.g. every 6m).

hexfusion:
great input, lets try to harden this.

lilic:
> The problem would also be less significant if we have minor release more frequently (e.g. every 6m).

+1 from me on this. What are the current issues around not doing those more frequently? Or is there just no need?

[...]

PR #14136, opened 2022-06-20
[3.4] Fix all the pipeline failues for release 3.4

ahrtr:
Fix [issues/14135](https://github.com/etcd-io/etcd/issues/14135)

Items resolved:
1. fix the vet error: possible misuse of reflect.SliceHeader;
2. fix the vet error: call to (*T).Fatal from a non-test goroutine;
3. bump package golang.org/x/crypto, net and sys;
4. bump boltdb from 1.3.3 to 1.3.6;
5. **remove the vendor directory**;
6. remove go 1.12.17 and 1.15.15, add go 1.16.15 into pipeline;
7. bump go version to 1.16 in go.mod;
8. fix the issue: compile: version go1.16.15 does not match go tool version go1.17.11;
9. fix data race on compactMainRev and watcherGauge;
10. fix test failure for `TestLeasingTxnOwnerGet` in `cluster_proxy` mode.


ahrtr:
cc @hexfusion @serathius @ptabor @spzala

.github/workflows/tests.yaml
```diff
@@ -17,53 +16,38 @@ jobs:
...
+        go-version: "1.16.15"
     - run: date
     - env:
         TARGET: ${{ matrix.target }}
-        GO_VERSION: ${{ matrix.go }}
       run: |
-        RACE='true'; if [[ ${GO_VERSION} == 1.15.15 ]]; then echo 'setting race off'; RACE='false'; fi
+        go version
+        echo ${GOROOT}
+        sudo unlink /usr/bin/go; sudo ln -s ${GOROOT}/bin/go /usr/bin/go; ls -lrt /usr/bin/go
```
ptabor, line 30: Please add a comment why do you relink the go version.

.github/workflows/tests.yaml
```diff
@@ -17,53 +16,38 @@ jobs:
...
+
         echo "${TARGET}"
         case "${TARGET}" in
           linux-amd64-fmt)
             GOARCH=amd64 PASSES='fmt bom dep' ./test
             ;;
           linux-amd64-integration-1-cpu)
-            GOARCH=amd64 CPU=1 PASSES='integration' RACE="${RACE}" ./test
+            GOARCH=amd64 CPU=1 PASSES='integration' RACE='false' ./test
```
ptabor, line 38: Is it intended to turn `RACE` of permanently ? 
I see you fix some of the races in the other places in this PR.

ptabor (approved):
Thank you. LGTM.

.github/workflows/tests.yaml
```diff
@@ -17,53 +16,38 @@ jobs:
...
-            GOARCH=amd64 CPU=4 PASSES='integration' RACE="${RACE}" ./test
+            GOARCH=amd64 CPU=4 PASSES='integration' RACE='false' ./test
             ;;
           linux-amd64-functional)
-            ./build && GOARCH=amd64 PASSES='functional' RACE="${RACE}" ./test
+            ./build && GOARCH=amd64 PASSES='functional' ./test
             ;;
           linux-amd64-unit)
-            GOARCH=amd64 PASSES='unit' RACE="${RACE}" ./test
+            GOARCH=amd64 PASSES='unit' RACE='false' ./test
```
serathius, line 50: I think we should have at least one test running with `RACE='true'`

.github/workflows/tests.yaml
ahrtr: Resolved

.github/workflows/tests.yaml
ahrtr: Yes, I agree. But let's add any new pipeline in a separate PR. I just bumped the go version to 1.16.15 in this PR, and keep all others unchanged. 

Previously when go version is 1.15.15, then RACE is set to false, see [tests.yaml#L47](https://github.com/etcd-io/etcd/blob/c2c9e7de019661b61f86e1743f461090c1b10771/.github/workflows/tests.yaml#L47). So I just keep it as it's.

go.mod
```diff
@@ -1,6 +1,6 @@
 module go.etcd.io/etcd
 
-go 1.12
+go 1.16
```
lavacat, line 3: I think the idea was to keep this at 1.12 https://github.com/etcd-io/etcd/issues/12840#issuecomment-815534131 CC @ptabor

ahrtr: No need to support go 1.12. It's out of support, and is subject to vulnerability. IIRC It is also not compatible with some system packages in newer version. 

Note that it's more than 1 year before when ptabor added that comment, and go 1.12 wasn't that old although it's also out of support at that time.

etcdserver/api/v2store/node_extern_test.go
```diff
@@ -104,5 +104,5 @@ func TestNodeExternClone(t *testing.T) {
 func sameSlice(a, b []*NodeExtern) bool {
 	ah := (*reflect.SliceHeader)(unsafe.Pointer(&a))
 	bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
-	return *ah == *bh
+	return ah.Data == bh.Data && ah.Len == bh.Len && ah.Cap == bh.Cap
```
lavacat, line 107: cherry pick this commit to make code same as in 3.5 https://github.com/etcd-io/etcd/pull/14134/commits/300c90a5e6f0b2ecc02841510860ffbec5caf8bc

clientv3/concurrency/election_test.go
```diff
@@ -64,14 +65,15 @@ func TestResumeElection(t *testing.T) {
...
+	errCHan := make(chan error)
 	go func() {
 		o := e.Observe(ctx)
 		respChan <- nil
 		for {
 			select {
 			case resp, ok := <-o:
 				if !ok {
-					t.Fatal("Observe() channel closed prematurely")
+					errCHan <- errors.New("observe() channel closed prematurely")
```
lavacat, line 76: cherry pick this to make code same as in 3.5 https://github.com/etcd-io/etcd/pull/14134/commits/31af5bdd96276beaab05751e944f98e1d2f677f4

ahrtr: updated.

etcdserver/api/v2store/node_extern_test.go
ahrtr: We do not necessarily backport everything from 3.5 to 3.4. 

And the previously fix isn't Semantically correct. The goal of the function `sameSlice` is to check whether the two slices are the same or not.  

If you look at go runtime's source code, you will get that a slice is actually a struct in golang, and it contains three fields. The change what I did is exactly to check the three fields. Actually the original code (`*ah ==  *bh`) is also correct to compare a struct, but go vet complains it because it's `reflect.SliceHeader`.

The fix in 3.5 * main (see below) isn't semantically correct, because [va.Pointer() ](https://pkg.go.dev/reflect#Value.Pointer)returns the first element of the slice. It's possible that two slices are different, but the first element is the same. I will submit a PR to fix it in main and 3.5. 

```
va := reflect.ValueOf(a)
vb := reflect.ValueOf(b)
return va.Len() == vb.Len() && va.Pointer() == vb.Pointer()
```

ahrtr: Correction to my previous comment. The code in 3.5 & main is also correct. The [doc](https://pkg.go.dev/reflect#Value.Pointer) mentions that `If v's Kind is Slice, the returned pointer is to the first element of the slice`. Actually it's the first element in the underneath array. The memory space for an array must be continuous, so if the first element is the same, then it means they are pointing to the same array.

go.mod
lavacat: I'm not 100% sure but I think go 1.16 keeps go compatibility [promise](https://go.dev/doc/go1.16), so its not subject to vulnerabilities. 
I also run into package issues but `go mod tidy` fixed it https://github.com/etcd-io/etcd/pull/14134/commits/ff898367250362288f5ce66fb7ed82be20210582

Either way is fine with me, but both  will work IMO.

ahrtr: Note that the [promise of compatibility](https://go.dev/doc/go1compat) is for golang language of version < 2.0. There is no guarantee on the core library and tool-chain.

ahrtr:
Plan for 3.4.19: [14105](https://github.com/etcd-io/etcd/issues/14105)

Issue #13511, opened 2021-11-29
etcdctl make-mirror crashed without any error information.


laixintao:
Hi, we have an `etcdctl make-mirror` process running for backup purpose. It runs normally until one day we did a stress test (which created high qps again our etcd server), after that, the etcdctl mirror process crashed, and can not finish the `SyncBase`.

I find no error information nor logs from etcdctl process, any suggestions for me about where can I check this? thanks.

some information about our setup:
```
systemctl cat etcd_mirror
# /etc/systemd/system/etcd_mirror.service
[Unit]
Description=Making mirror from old to new etcd cluster
Documentation=https://github.com/etcd-io/etcd

[Service]
User=root
Group=root
ExecStart=etcdctl make-mirror http://<ip-omitted>:2379 --user=xxx:yyy
Restart=always
RestartSec=20

[Install]
WantedBy=multi-user.target
```

```
systemd[1]: Started Making mirror from old to new etcd cluster.
etcdctl[232812]: 15126
etcdctl[232812]: 30299
etcdctl[232812]: 45098
etcdctl[232812]: 60294
etcdctl[232812]: 72321
etcdctl[232812]: 80344
etcdctl[232812]: 87954
etcdctl[232812]: 94798
etcdctl[232812]: 101501
etcdctl[232812]: 107981
etcdctl[232812]: 114677
etcdctl[232812]: 120958
etcdctl[232812]: 128069
etcdctl[232812]: 135138
etcdctl[232812]: Error: <nil>
```

```
etcd --version
etcd Version: 3.4.9
Git SHA: 54ba95891
Go Version: go1.12.17
Go OS/Arch: linux/amd64
```

ahrtr:
Looks like the messages are printed by your application? Note that etcd printes an error to os.Stderr when this command fails. 
```
etcdctl[232812]: 128069
etcdctl[232812]: 135138
etcdctl[232812]: Error: <nil>
```

laixintao:
hi @ahrtr thanks for replying.

This message was from systems (syslog), I think it is printed by `etcdctl make-mirror` process.

I guess this is the line of printing code. It means total keys that have been synced I guess?

https://github.com/etcd-io/etcd/blob/7572a61a39d4eaad596ab8d9364f7df9a84ff4a3/etcdctl/ctlv3/command/make_mirror_command.go#L136

laixintao:
[...]

PR #14142, opened 2022-06-22
make-mirror store mod revision to support incremental mirror 

phuc1998:
Issue: https://github.com/etcd-io/etcd/issues/13511#issuecomment-987706713

Currently make-mirror have to pass rev flag when start mirror, but by the way, we have to store revision at somewhere. To make it easier to use, we store revision at key __mod_revision and using it when start make-mirror again.

Signed-off-by: Phuc Chau chauhoangphuc432@gmail.com


ptabor:
Interesting proposal. Thank you. 

A few observations: 
  1. This code would definitely need to be opt-in (e.g. by flag deciding about the frequency). It's unacceptable to etcd to start modifying data by mirroring when it used not to do it.
  2. This implementation would not work if there are multiple mirrors of the same server, as they would keep conflicting. We would need a way to 'name' the mirrors (as part of the key name). 
  3. I think we would need to make a bigger decision about naming conventions of "etcd-metadata" naming convention when stored/exposed within etcd.

ahrtr:
Thanks @phuc1998 for the PR. 

I think it's application's responsibility to store the rev. It means that the code change should be included in your application instead of etcd. 

The rev isn't necessarily to be stored in etcd, you can save it anywhere. It's up to application.

phuc1998:
@ahrtr @ptabor thanks for your response, I am pleasure to contribute to the open source. I will make it senses and contribute later

Issue #14110, opened 2022-06-13
panic: unexpected error during txn


aphyr:
### What happened?

In a Jepsen test run of five etcd 3.5.3 nodes, with process pauses (sending processes SIGSTOP and SIGCONT), compaction (performed via the admin API), and defragmentation (via etcdctl), one etcd process crashed with the following error message:

```
[...]

PR #3534, opened 2015-09-15
etcdserver: better v3 api error handling

xiang90:
Start to handle error

PR #14149, opened 2022-06-23
server: don't panic in readonly serializable txn

lavacat:
Problem: We pass grpc context down to applier in readonly serializable txn.
This context can be cancelled for example due to timeout. This will trigger panic.

Solution: provide different error handler for readonly serializable txn.

fixes https://github.com/etcd-io/etcd/issues/14110

### DIFF
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -17,6 +17,7 @@ package txn
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"sort"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
@@ -265,7 +266,18 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
 		txnWrite.End()
 		txnWrite = kv.Write(trace)
 	}
-	applyTxn(ctx, lg, kv, lessor, txnWrite, rt, txnPath, txnResp)
+	_, err := applyTxn(ctx, lg, kv, lessor, txnWrite, rt, txnPath, txnResp)
+	if err != nil {
+		if isWrite {
+			// end txn to release locks before panic
+			txnWrite.End()
+			// When txn with write operations starts it has to be successful
+			// We don't have a way to recover state in case of write failure
+			lg.Panic("unexpected error during txn with writes", zap.Error(err))
+		} else {
+			lg.Error("unexpected error during readonly txn", zap.Error(err))
+		}
+	}
 	rev := txnWrite.Rev()
 	if len(txnWrite.Changes()) != 0 {
 		rev++
@@ -277,7 +289,7 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
 		traceutil.Field{Key: "number_of_response", Value: len(txnResp.Responses)},
 		traceutil.Field{Key: "response_revision", Value: txnResp.Header.Revision},
 	)
-	return txnResp, trace, nil
+	return txnResp, trace, err
 }
 
 // newTxnResp allocates a txn response for a txn request given a path.
@@ -311,7 +323,7 @@ func newTxnResp(rt *pb.TxnRequest, txnPath []bool) (txnResp *pb.TxnResponse, txn
 	return txnResp, txnCount
 }
 
-func applyTxn(ctx context.Context, lg *zap.Logger, kv mvcc.KV, lessor lease.Lessor, txnWrite mvcc.TxnWrite, rt *pb.TxnRequest, txnPath []bool, tresp *pb.TxnResponse) (txns int) {
+func applyTxn(ctx context.Context, lg *zap.Logger, kv mvcc.KV, lessor lease.Lessor, txnWrite mvcc.TxnWrite, rt *pb.TxnRequest, txnPath []bool, tresp *pb.TxnResponse) (txns int, err error) {
 	trace := traceutil.Get(ctx)
 	reqs := rt.Success
 	if !txnPath[0] {
@@ -328,7 +340,7 @@ func applyTxn(ctx context.Context, lg *zap.Logger, kv mvcc.KV, lessor lease.Less
 				traceutil.Field{Key: "range_end", Value: string(tv.RequestRange.RangeEnd)})
 			resp, err := Range(ctx, lg, kv, txnWrite, tv.RequestRange)
 			if err != nil {
-				lg.Panic("unexpected error during txnWrite", zap.Error(err))
+				return 0, fmt.Errorf("applyTxn: failed Range: %w", err)
 			}
 			respi.(*pb.ResponseOp_ResponseRange).ResponseRange = resp
 			trace.StopSubTrace()
@@ -339,26 +351,30 @@ func applyTxn(ctx context.Context, lg *zap.Logger, kv mvcc.KV, lessor lease.Less
 				traceutil.Field{Key: "req_size", Value: tv.RequestPut.Size()})
 			resp, _, err := Put(ctx, lg, lessor, kv, txnWrite, tv.RequestPut)
 			if err != nil {
-				lg.Panic("unexpected error during txnWrite", zap.Error(err))
+				return 0, fmt.Errorf("applyTxn: failed Put: %w", err)
 			}
 			respi.(*pb.ResponseOp_ResponsePut).ResponsePut = resp
 			trace.StopSubTrace()
 		case *pb.RequestOp_RequestDeleteRange:
 			resp, err := DeleteRange(kv, txnWrite, tv.RequestDeleteRange)
 			if err != nil {
-				lg.Panic("unexpected error during txnWrite", zap.Error(err))
+				return 0, fmt.Errorf("applyTxn: failed DeleteRange: %w", err)
 			}
 			respi.(*pb.ResponseOp_ResponseDeleteRange).ResponseDeleteRange = resp
 		case *pb.RequestOp_RequestTxn:
 			resp := respi.(*pb.ResponseOp_ResponseTxn).ResponseTxn
-			applyTxns := applyTxn(ctx, lg, kv, lessor, txnWrite, tv.RequestTxn, txnPath[1:], resp)
+			applyTxns, err := applyTxn(ctx, lg, kv, lessor, txnWrite, tv.RequestTxn, txnPath[1:], resp)
+			if err != nil {
+				// don't wrap the error. It's a recursive call and err should be already wrapped
+				return 0, err
+			}
 			txns += applyTxns + 1
 			txnPath = txnPath[applyTxns+1:]
 		default:
 			// empty union
 		}
 	}
-	return txns
+	return txns, nil
 }
 
 //---------------------------------------------------------
--- a/server/etcdserver/txn/txn_test.go
+++ b/server/etcdserver/txn/txn_test.go
@@ -1,0 +1,81 @@
+package txn
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/server/v3/lease"
+	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
+	"go.uber.org/zap/zaptest"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadonlyTxnError(t *testing.T) {
+	b, _ := betesting.NewDefaultTmpBackend(t)
+	defer betesting.Close(t, b)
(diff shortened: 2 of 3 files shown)


ahrtr:
Please sign off the commit.

server/etcdserver/api/v3rpc/grpc.go
```diff
@@ -36,7 +36,7 @@ const (
 	maxSendBytes      = math.MaxInt32
 )
 
-func Server(s *etcdserver.EtcdServer, tls *tls.Config, interceptor grpc.UnaryServerInterceptor, gopts ...grpc.ServerOption) *grpc.Server {
+func Server(s *etcdserver.EtcdServer, tls *tls.Config, interceptors []grpc.UnaryServerInterceptor, gopts ...grpc.ServerOption) *grpc.Server {
```
ahrtr, line 39: unrelated change?

server/etcdserver/txn/txn.go
```diff
@@ -311,7 +322,7 @@ func newTxnResp(rt *pb.TxnRequest, txnPath []bool) (txnResp *pb.TxnResponse, txn
 	return txnResp, txnCount
 }
 
-func applyTxn(ctx context.Context, lg *zap.Logger, kv mvcc.KV, lessor lease.Lessor, txnWrite mvcc.TxnWrite, rt *pb.TxnRequest, txnPath []bool, tresp *pb.TxnResponse) (txns int) {
+func applyTxn(ctx context.Context, lg *zap.Logger, kv mvcc.KV, lessor lease.Lessor, txnWrite mvcc.TxnWrite, rt *pb.TxnRequest, txnPath []bool, tresp *pb.TxnResponse, errHander CriticalErrHandler) (txns int, err error) {
```
ahrtr, line 325: The concern is it's hacking the implementation too much. The clearer way is to resolve it out of the box, we can just add a panic-recover, just as I mentioned in [issuecomment-1154738363](https://github.com/etcd-io/etcd/issues/14110#issuecomment-1154738363)

server/etcdserver/api/v3rpc/grpc.go
lavacat: this is to make test work. This interceptors are used only in integrations tests. There is already grpc_testing.GrpcRecorder. I've added another one. See changes in intergration/cluster.go

server/etcdserver/txn/txn.go
```diff
@@ -265,7 +265,12 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
 		txnWrite.End()
 		txnWrite = kv.Write(trace)
 	}
-	applyTxn(ctx, lg, kv, lessor, txnWrite, rt, txnPath, txnResp)
+	_, err := applyTxn(ctx, lg, kv, lessor, txnWrite, rt, txnPath, txnResp)
+	if err != nil && isWrite {
+		// When txn with write operations starts it has to be successful
+		// We don't have a way to recover state in case of write failure
+		lg.Panic("unexpected error during txnWrite", zap.Error(err))
```
lavacat, line 272: @ptabor @ahrtr I've tried passing an error handler (see outdated version of this PR) but it seams a bit cumbersome. @ahrtr suggested [recover](https://github.com/etcd-io/etcd/issues/14110#issuecomment-1154738363). But it's only going to be for readonly txn not all requests, also seams cumbersome. 

How about just panic when txn has write ops? 

For ref see, original PR when panic was introduced https://github.com/etcd-io/etcd/pull/3534/files#diff-26b2effd11719aa79af125ed4e17234ef08970839d4f83fb4d802e04614c3cd4R161

ahrtr: Looks good to me!

tests/integration/v3_grpc_test.go
```diff
@@ -1952,3 +1953,77 @@ func waitForRestart(t *testing.T, kvc pb.KVClient) {
 		t.Fatalf("timed out waiting for restart: %v", err)
 	}
 }
+
+func TestV3ReadonlyTxnCancelledContext(t *testing.T) {
```
ahrtr, line 1957: Have you considered to add a common test which is executed on both e2e and integration environments? And also use the new [interface](https://github.com/etcd-io/etcd/blob/main/tests/framework/interface.go) ?  

cc @serathius

ahrtr: It seems we can only add integration test here because we need to inject an interceptor, but we do not support that in command line (and also e2e). 

But please consider to use the new interface in https://github.com/etcd-io/etcd/blob/main/tests/framework/integration.go

tests/integration/v3_grpc_test.go
```diff
@@ -1952,3 +1953,77 @@ func waitForRestart(t *testing.T, kvc pb.KVClient) {
...
+	integration.BeforeTest(t)
+	clus := integration.NewCluster(t, &integration.ClusterConfig{
+		Size: 1,
+		// Context should be cancelled on the second check that happens inside rangeKeys
+		GrpcInterceptors: []grpc.UnaryServerInterceptor{injectMockContextForTxn(newMockContext(2))},
```
ahrtr, line 1962: The `2` here is a little tricky. 

I suggest to simplify the test case for better readability:
1. Remove the Put (line 1968), and let's test the `readonly transaction` only;
2. always return a closed channel in `(ctx *mockCtx) Done();`

tests/integration/v3_grpc_test.go
```diff
@@ -1952,3 +1953,77 @@ func waitForRestart(t *testing.T, kvc pb.KVClient) {
...
+	_, err := kvc.Put(context.TODO(), pr)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	txnget := &pb.RequestOp{Request: &pb.RequestOp_RequestRange{RequestRange: &pb.RangeRequest{Key: []byte("abc")}}}
+	txn := &pb.TxnRequest{Success: []*pb.RequestOp{txnget}}
+	_, err = kvc.Txn(context.TODO(), txn)
+	if err == nil || !strings.Contains(err.Error(), "range context cancelled: mock context error") {
+		t.Fatal(err)
```
ahrtr, line 1977: It seems that the `err` has nothing to do with `applyTxn`.  Note that it's returned by the API instead of the applying workflow.

tests/integration/v3_grpc_test.go
serathius: One problem, new interface is not available in v3.5, so using it now would make backport much harder.

tests/integration/v3_grpc_test.go
ptabor: Agree that counting calls on context to trigger reaction is not maintainable. 
Thinking about alternatives, how about intercepting zap logs, using zap.hooks: 

https://stackoverflow.com/questions/64803634/how-to-add-a-hook-into-a-zap-logger

Still not perfect, but allows to e.g. disable context when specific transaction is debug-logged.

server/etcdserver/txn/txn.go
ptabor: I have small preference towards apply workflow being extremely strict and deterministic and any unexpected error causing premature exit from function (even in RO) code is something that should lead to investigation and fixing. 

But I don't see any exact problem with the proposed approach.

tests/integration/v3_grpc_test.go
lavacat: @ahrtr Put is there to prevent short circuit in storeTxnRead.Range. We need some data to trigger `case <-ctx.Done():`
Always returning closed channel doesn't work because it fails early in `EtcdServer.linearizableReadNotify` and never gets to `txn.Txn`

@ptabor how do you trigger `context cancelled` inside `storeTxnRead.Range` if you don't let the first check in `EtcdServer.linearizableReadNotify` to pass? Maybe I'm missing something, but how will intercepting zap logs help? As far as I can tell only failure inside storeTxnRead.Range triggers panic, so that's what we want to model in the test.

Another approach I see is to add test only on txn.Txn level and not replicate panic as integration test. Maybe not ideal, but integrations tests don't have to cover every failure scenario.

tests/integration/v3_grpc_test.go
```diff
@@ -1952,3 +1953,77 @@ func waitForRestart(t *testing.T, kvc pb.KVClient) {
...
+
+func newMockContext(doneAfter int) context.Context {
+	return &mockCtx{
+		calledDone: 0,
+		doneAfter:  doneAfter,
+		donec:      make(chan struct{}),
+	}
+}
+
+func (*mockCtx) Deadline() (deadline time.Time, ok bool) {
```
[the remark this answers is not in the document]
lavacat: It's part of the interface https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/context/context.go;l=66

tests/integration/v3_grpc_test.go
lavacat: wrapped err in applyTxn

lavacat:
@ahrtr @ptabor updated PR with alternative version of the test that uses reflection to check the name of the caller function and cancel context for rangeKeys. 
Added more comments about test setup.
Addressed other small comments.

The only thing that isn't done is migration to new client interface ([comment](https://github.com/etcd-io/etcd/pull/14149#discussion_r915294098)). If this approach looks good, I'll try the new client for main and keep this version for 3.5

tests/integration/v3_grpc_test.go
```diff
@@ -1952,3 +1952,39 @@ func waitForRestart(t *testing.T, kvc pb.KVClient) {
...
+	// need to insert some data, otherwise rangeKeys will terminate early before checking context
+	pr := &pb.PutRequest{Key: []byte("abc"), Value: []byte("def")}
+	_, err := kvc.Put(context.TODO(), pr)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	txnget := &pb.RequestOp{Request: &pb.RequestOp_RequestRange{RequestRange: &pb.RangeRequest{Key: []byte("abc")}}}
+	txn := &pb.TxnRequest{Success: []*pb.RequestOp{txnget}}
+	_, err = kvc.Txn(context.TODO(), txn)
```
ahrtr, line 1975: I am a little concerned about the readability of the test case. Can you try the following proposal?
1. pass a context with a deadline, i.e. 5s;
2. register an unary interceptor, and just sleep a couple of seconds (5+1=6s) in the interceptor;
3. The `kvc.Txn` should return an error with something like context timeout

ahrtr: I am also thinking probably we can just add a unit test to accurately test the different behaviors:
1. For readonly transaction, it should NOT panic due to context timeout;
2. For non-readonly trancation, it should panic if the context timeout

lavacat: Deadline approach doesn't work, because it's early terminated [here](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/v3_server.go#L170) (tried it out).

Unit tests are fine with me. I've mentioned this [earlier](https://github.com/etcd-io/etcd/pull/14149#discussion_r920533239). 
> Another approach I see is to add test only on txn.Txn level and not replicate panic as integration test. Maybe not ideal, but integrations tests don't have to cover every failure scenario.

Will work on PR to convert to unit tests and remove this test.

lavacat:
@ahrtr this is ready for review with unit tests instead of integration

server/etcdserver/txn/txn_test.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+	}
+
+	assert.Panics(t, func() { Txn(ctx, zaptest.NewLogger(t), txn, false, s, &lease.FakeLessor{}) }, "Expected panic in Txn with writes")
+}
+
+func cleanupStore(s mvcc.KV, b backend.Backend, path string) {
+	s.Close()
+	b.Close()
+	os.Remove(path)
```
ahrtr, line 86: It should be unnecessary, because the golang build-in test lib can remove the directory returned by  `t.TempDir() `automatically.

server/etcdserver/txn/txn_test.go
```diff
@@ -0,0 +1,87 @@
...
+			},
+		},
+	}
+
+	assert.Panics(t, func() { Txn(ctx, zaptest.NewLogger(t), txn, false, s, &lease.FakeLessor{}) }, "Expected panic in Txn with writes")
+}
+
+func cleanupStore(s mvcc.KV, b backend.Backend, path string) {
+	s.Close()
+	b.Close()
```
ahrtr, line 85: Please use [betesting.Close(t, b)](https://github.com/etcd-io/etcd/blob/649babaf4a5add26b41b94dc0eec6738d04c89fb/server/storage/backend/testing/betesting.go#L50) instead.

ahrtr (approved):
LGTM 

Thank you @lavacat 

cc @spzala @ptabor @serathius to double confirm. This is the core applying workflow, there must be at least two approvals.

spzala (approved):
lgtm
Thanks @lavacat @ahrtr

ahrtr:
Thanks @spzala 

Great work! @lavacat

PR #14160, opened 2022-06-25
removes empty option in JSON tag

chavacava:
option can not be empty in JSON tag

Found with [`revive`](https://github.com/mgechev/revive) (rule `struct-tag`)


ahrtr:
Thank you. Please sign off the commit using command `git  commit --signoff --amend`

Probably we should add a automated check into the pipeline, would you be interested in work on this?

PR #14165, opened 2022-06-26
pkg/expect: send SIGTERM to target expect process instead of SIGKILL for `Stop()`

eval-exec:
The `SIGKILL` signal is extremely rude to a process, it can not be caught by target process.
Kill a process by `SIGTERM` is more elegent, because we can give the target process a chance to do clean ups, if the target process has a signal handler for `SIGTERM`.

https://linuxhandbook.com/sigterm-vs-sigkill/
https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -42,7 +42,7 @@ type ExpectProcess struct {
 	count int // increment whenever new line gets added
 	err   error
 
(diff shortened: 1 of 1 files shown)


ahrtr (approved):
LGTM

Thank you. @eval-exec

PR #14081, opened 2022-05-30
 https server will not be impacted by the MaxConcurrentStreams settin…

YANGJINJUE:
https server will not be impacted by the MaxConcurrentStreams settings of grpc opts

Issue #14077, opened 2022-05-27
Etcd unexpected request time consumption with HTTPs v3 API


leslie-tsang:
### What happened?

We use a API gateway APISIX to watch etcd cluster(`auth` and `TLS` enabled) resource with HTTPs request, the APISIX cluster will raise about 260 connection to etcd cluster.

In this scenario, the etcd operation with `HTTPs` request will speed too much time (more than 20s). the operation with `gRPC` protocol wouldn't spend that much time.

> We use etcdctl and etcdkeeper as `gRPC` client

## Simple conclusion
Etcd operation with `RESTful` request with TLS will spend too much time (more than 10s).
Etcd operation with `gRPC` request performing normal (within 0.5s).

Etcd use `grpc-gateway` to convert `HTTP` request to `gRPC` request, Once the `auth` is enabled, the perf of etcd cluster becomes lower, and then when TLS is enabled, the perf of the etcd cluster becomes even worse.

We had a try about etcd cluster <ins>`without TLS`</ins>, then it perf as normal one again.

### What did you expect to happen?

Etcd operation within 1s.

### How can we reproduce it (as minimally and precisely as possible)?

A `auth` and `TLS` enabled etcd cluster env can launch with [apisix-etcd-mtls](https://github.com/NGX-AI-APISIX/apisix-etcd-mtls)

### Anything else we need to know?

APISIX will generates a numbers of watch request to etcd cluster via `HTTPs`. 

> There is a quick calculation formula

Number of watch request = APISIX nodes * worker number * 13
Take 5 APISIX nodes with 4 worker per node as an example, The APISIX cluster will raise about: 5 * 4 * 13 = 260.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
etcd Version: 3.5.4
Git SHA: 08407ff76
Go Version: go1.16.15
Go OS/Arch: linux/amd64

$ etcdctl version
etcdctl version: 3.5.4
API version: 3.5
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

ETCD_CLIENT_KEY_FILE=/opt/etcd/ssl/client-key.pem
HOSTNAME=bc18fac70256
ETCD_LISTEN_CLIENT_URLS=https://0.0.0.0:2379
PWD=/opt/bitnami/etcd
OS_FLAVOUR=debian-10
ETCD_ENABLE_V2=true
ETCD_LISTEN_PEER_URLS=https://0.0.0.0:2380
[...]

Issue #14065, opened 2022-05-23
ETCD unusual long request 


Battlefield-Pony:
### What happened?

during setting up 3 node etcd cluster, it is found that if there are more than 5 apisix nodes exists, the etcd may takes longer than 20 seconds to process the request.
As picture indicated below
![image](https://user-images.githubusercontent.com/66858976/169774048-8031920a-4636-4fb4-bfa3-af37eb29eb8c.png)

### What did you expect to happen?

Usually with in this amount of data and connections, it usually wont be wait more than 1 second to process.

### How can we reproduce it (as minimally and precisely as possible)?

setting up 3 node etcd cluster, and 6 apisix nodes
try to modify a route, and it produces.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
etcd Version: 3.5.3
Git SHA: 0452feec7
Go Version: go1.16.15
Go OS/Arch: linux/amd64


$ etcdctl version
etcdctl version: 3.5.3
API version: 3.5
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here
flags:
--cert-file=/root/etcd-cert/server.crt 
--key-file=/root/etcd-cert/server.key 
--peer-auto-tls --metrics 'extensive'
--enable-pprof  
--log-level 'debug'

ENV_PARAMS:
ETCD_INITIAL_CLUSTER="node1=https://node1:2380,node2=https://node2:2380,node3=https://node3:2380"
ETCD_INITIAL_CLUSTER_TOKEN="etcd-cluster"
ETCD_INITIAL_CLUSTER_STATE="new"
ETCD_NAME="node1"
ETCD_DATA_DIR="/data/etcd/default.etcd"
ETCD_LISTEN_PEER_URLS="https://node1:2380"
ETCD_LISTEN_CLIENT_URLS="https://node1:2379,https://127.0.0.1:2379"
ETCD_INITIAL_ADVERTISE_PEER_URLS="https://node1:2380"
ETCD_ADVERTISE_CLIENT_URLS="https://node1:2379"
</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
[...]

PR #14169, opened 2022-06-27
Support configuring `MaxConcurrentStreams` for http2

ahrtr:
Continue to work on [14081](https://github.com/etcd-io/etcd/pull/14081) .

Related issues: 
* https://github.com/etcd-io/etcd/issues/14077
* https://github.com/etcd-io/etcd/issues/14065
* https://github.com/etcd-io/etcd/issues/14185

## Benchmark test result:
Note I only created some concurrent streams, and there is no any traffic. The test were performed on a Linux VM with 8GiB memory and 4 CPUs (Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz).

It seems the memory used by concurrent streams isn't very high, it's about 90MB for 1000 concurrent streams, and 108MB for 2000 concurrent streams. 

### 0 stream
```
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 1.7567744e+07
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.1481223168e+10
# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
# TYPE process_virtual_memory_max_bytes gauge
process_virtual_memory_max_bytes 1.8446744073709552e+19

# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 0.44
```
[0_stream.txt](https://github.com/etcd-io/etcd/files/8996516/0_stream.txt)

### 3 streams
```
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 1.994752e+07
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.1481223168e+10
# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
# TYPE process_virtual_memory_max_bytes gauge
process_virtual_memory_max_bytes 1.8446744073709552e+19

# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 1.55
```
[3_stream.txt](https://github.com/etcd-io/etcd/files/8996518/3_stream.txt)

### 100 streams
```
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 2.7332608e+07
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.1481485312e+10
# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
# TYPE process_virtual_memory_max_bytes gauge
process_virtual_memory_max_bytes 1.8446744073709552e+19

# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 0.69
```
[100_streams.txt](https://github.com/etcd-io/etcd/files/8996519/100_streams.txt)

### 1000 streams
```
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 9.1369472e+07
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.1552464896e+10
# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
# TYPE process_virtual_memory_max_bytes gauge
process_virtual_memory_max_bytes 1.8446744073709552e+19

# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 1.68
```
[1000_streams.txt](https://github.com/etcd-io/etcd/files/8996520/1000_streams.txt)

### 2000 streams
```
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 1.0809344e+08
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.155272704e+10
# HELP process_virtual_memory_max_bytes Maximum amount of virtual memory available in bytes.
# TYPE process_virtual_memory_max_bytes gauge
process_virtual_memory_max_bytes 1.8446744073709552e+19

# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 1.83
```
[2000_streams.txt](https://github.com/etcd-io/etcd/files/8996521/2000_streams.txt)


ahrtr:
cc @ptabor @serathius @spzala

tests/e2e/v3_curl_maxstream_test.go
```diff
@@ -0,0 +1,218 @@
...
+		expectedResponse = "Operation timed out"
+	}
+	wg.Add(int(concurrentNumber))
+	t.Logf("Running the test, MaxConcurrentStreams: %d, concurrentNumber: %d, expectedResponse: %s\n",
+		cx.cfg.MaxConcurrentStreams, concurrentNumber, expectedResponse)
+	errCh := make(chan error, concurrentNumber)
+	submitConcurrentWatch(cx, int(concurrentNumber), &wg, errCh)
+	submitRangeAfterConcurrentWatch(cx, expectedResponse)
+
+	// Step 4: check the watch errors. Note that we ony check the watch error
```
spzala, line 117: Nit, "Only" (instead of `ony`) and "close" (instead of `closed`)

spzala (commented):
@ahrtr thanks, nice work. I have few nits.

pkg/flags/uint32.go
```diff
@@ -0,0 +1,45 @@
...
+package flags
+
+import (
+	"flag"
+	"strconv"
+)
+
+type uint32Value uint32
+
+// NewUint32Value creates a uint32Value instance with the default value `v`.
```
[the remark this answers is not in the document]
ahrtr: thx & resolved.

pkg/flags/uint32.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Set parses a command line uint32 value.
+// Implements "flag.Value" interface.
+func (i *uint32Value) Set(s string) error {
+	v, err := strconv.ParseUint(s, 0, 32)
+	*i = uint32Value(v)
+	return err
+}
+
+func (i *uint32Value) String() string { return strconv.FormatUint(uint64(*i), 10) }
```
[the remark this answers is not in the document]
ahrtr: The defined type `uint32Value` implements the interface [flag.Value](https://github.com/golang/go/blob/go1.17.11/src/flag/flag.go#L297-L300) (see below as well) , so it must implement the method `String`.

```

type Value interface {
	String() string
	Set(string) error
}
```

FYI. Usually if a defined type needs to be printed by `fmt.Printf`, then it needs to implement the interface [fmt.Stringer ](https://pkg.go.dev/fmt#Stringer). 
```

type Stringer interface {
	String() [string](https://pkg.go.dev/builtin#string)
}
```

pkg/flags/uint32.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Uint32FromFlag return the uint32 value of a flag with the given name
+func Uint32FromFlag(fs *flag.FlagSet, name string) uint32 {
```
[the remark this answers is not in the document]
ahrtr: thx & done.

server/config/config.go
```diff
@@ -129,6 +129,10 @@ type ServerConfig struct {
 	// MaxRequestBytes is the maximum request size to send over raft.
 	MaxRequestBytes uint
 
+	// MaxConcurrentStreams specifies the max number of concurrent
```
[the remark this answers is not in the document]
ahrtr: thx & updated all places.

server/embed/etcd.go
```diff
@@ -336,7 +337,10 @@ func print(lg *zap.Logger, ec Config, sc config.ServerConfig, memberInitialized
 		zap.String("initial-cluster", sc.InitialPeerURLsMap.String()),
 		zap.String("initial-cluster-state", ec.ClusterState),
 		zap.String("initial-cluster-token", sc.InitialClusterToken),
-		zap.Int64("quota-size-bytes", quota),
+		zap.Int64("quota-backend-bytes", quota),
```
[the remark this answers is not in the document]
ahrtr: Yes, I should fix it in a separate PR. But it's really a minor minor change, so I just fixed it here.

tests/e2e/v3_curl_maxstream_test.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func testV3CurlMaxStream(t *testing.T, reachLimit bool, opts ...ctlOption) {
+	e2e.BeforeTest(t)
+
+	// Step 1: generate configuration for creating cluster
+	t.Log("Generating configuration for creating cluster.")
+	cx := getDefaultCtlCtx(t)
+	cx.applyOpts(opts)
+	// We must set the clusterSize to 1, otherwise different streams may
```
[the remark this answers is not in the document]
ahrtr: thx & done.

tests/e2e/v3_curl_maxstream_test.go
ahrtr: Good catch. updated.

ahrtr:
@spzala Resolved all your comments, PTAL, thx.

server/embed/config.go
```diff
@@ -205,6 +207,10 @@ type Config struct {
 	MaxTxnOps           uint   `json:"max-txn-ops"`
 	MaxRequestBytes     uint   `json:"max-request-bytes"`
 
+	// MaxConcurrentStreams specifies the maximum number of concurrent
+	// streams that a client can open at a time.
```
serathius, line 211:
```suggestion
	// MaxConcurrentStreams specifies the maximum number of concurrent
	// streams that each client may have open at a time. 
```

server/embed/config.go
```diff
@@ -59,6 +60,7 @@ const (
 	DefaultWarningApplyDuration        = 100 * time.Millisecond
 	DefaultWarningUnaryRequestDuration = 300 * time.Millisecond
 	DefaultMaxRequestBytes             = 1.5 * 1024 * 1024
+	DefaultMaxConcurrentStreams        = math.MaxUint32
```
serathius, line 63: This changes the default of at least 100 that is recommended by HTTP/2 spec. https://pkg.go.dev/golang.org/x/net/http2#Server 

Going from hundreds to MaxUint32 is pretty big difference. This opens a new possible issue of misbehaving problem, so the question is about expected user expirience:
* We leave the previous default, if user hits the limit, they will notice slow requests, learn that there is option to change it and they decide to increase it.
* We set MaxUint32 as default, misbehaving clients will cause performance issues for users possibly causing some clusters to go down. Client bug in popular application using etcd, makes this issue prominent and users figure out new safe default by themselves. 

I would prefer to separate opening new options to users from changing the defaults.  I don't think that provided examples would motivate us to change the default for everyone.

pkg/flags/uint32.go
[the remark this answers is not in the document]
spzala: Cool, thanks!

server/embed/config.go
ahrtr: It seems that the best choice for us is to use `math.MaxUint32` as the default value. Reasons:
1. The existing maxStream for grpc is already [math.MaxUint32](https://github.com/etcd-io/etcd/blob/248f072370628fa2ad7c3d73e3935ba5636c0442/server/etcdserver/api/v3rpc/grpc.go#L35), changing the value may breaking the existing behavior;
2. The [HTTP2 spec](https://datatracker.ietf.org/doc/html/rfc7540#section-6.5.2) says "no smaller than 100", but does not say that it can't be MaxUint32 (of course, this point may be a little debatable). 
```
      Initially, there is no limit to
      this value.  It is recommended that this value be no smaller than
      100, so as to not unnecessarily limit parallelism.
```
3. Actually the existing implementation of gRPC also set the default value as [math.MaxUint32](https://github.com/grpc/grpc-go/blob/5b509df1e30a25b14dba119c09295924c9e17abe/internal/transport/http2_server.go#L172).

server/embed/config.go
ahrtr: Anyway, I replaced the `a` with `each`, and kept the `can` as it's, just as I mentioned above. This is the best I can tell from my perspective as a non-native English speaker :)

server/etcdmain/help.go
```diff
@@ -81,6 +81,8 @@ Member:
     Maximum number of operations permitted in a transaction.
   --max-request-bytes '1572864'
     Maximum client request size in bytes the server will accept.
+  --max-concurrent-streams 'math.MaxUint32'
+    Maximum concurrent streams that a client can open at a time.
```
spzala, line 85: nit, let's keep it consistent  (`each` vs `a`) if you modify the PR. @ahrtr

pkg/flags/uint32_test.go
```diff
@@ -0,0 +1,111 @@
...
+			defaultVal:  105,
+			arguments:   []string{"--max-concurrent-streams", "105"},
+			expectedVal: 105,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			fs := flag.NewFlagSet("etcd", flag.ContinueOnError)
+			fs.Var(NewUint32Value(tc.defaultVal), flagName, "Maximum concurrent streams that a client can open at a time.")
```
spzala, line 103: nit, let's keep it consistent  (`each` vs `a`) if you modify the PR. @ahrtr

spzala (approved):
lgtm 
Thanks @ahrtr

server/embed/config.go
spzala: :) thanks much @ahrtr I added couple of nits related to it, but it looks good to me. As long as it's technically correct we are good.

server/etcdmain/help.go
ahrtr: thx & updated.

pkg/flags/uint32_test.go
ahrtr: thx & updated.

ahrtr:
> lgtm Thanks @ahrtr

Thanks @spzala .

cc @serathius Do you have any other comments on this? I only updated some messages (`a` vs `each`) since your last review.

serathius (approved):

server/embed/config.go
serathius: Ok, I'm getting a little confused about what bugfix vs change of default. https://github.com/etcd-io/etcd/issues/14077#issuecomment-1140057917 mentions that the current ConcurrentMaxStream for TLS enabled clusters is 250. In above comment you say to set ConcurrentMaxStream to max.MaxUint32 which is default for grpc, so why it was not used by etcd before? Are we making changes or not?

Based on that we are changing the default which would be something that I would prefer not to backport.

Can you please summarize:
* What this PR changes?
* Why do you think it's worth backporting this?

ahrtr: They are indeed good questions. The points are:

1. The default maxStream for gRPC is [math.MaxUint32](https://github.com/grpc/grpc-go/blob/5b509df1e30a25b14dba119c09295924c9e17abe/internal/transport/http2_server.go#L172), while the default value for http2 is [250](https://cs.opensource.google/go/x/net/+/master:http2/server.go;drc=1d687d428aca0546c0ca84160c8700ee521e9fb9;l=58).

2. Why does only https have such issue? If you have this question as well, please see https://github.com/etcd-io/etcd/pull/13985#issuecomment-1120047735 

3. Actually previous PR [8238](https://github.com/etcd-io/etcd/pull/8238) isn't complete, because it only changed the `maxStream` for gRPC, but not for HTTP2. But probably the [code of HTTP encapsulation gRPC](https://github.com/etcd-io/etcd/blob/248f072370628fa2ad7c3d73e3935ba5636c0442/server/embed/serve.go#L126-L207) might be added after the PR, I do not waste my time to double check that; if it's the case, then we can say the change of encapsulating gRPC using HTTP has issue.  Anyway, no matter which case is true, the issue is the max concurrent stream each client can open at a time is actually 250 instead of math.MaxUint32 over https.  It's an issue, so we need to resolve it, and accordingly backport to 3.5. 

Do you still have any concern?

ahrtr: Regarding how to the cherry pick the PR to 3.5, it's open to discussion. Probably we can just set the maxStream as `math.MaxUint32` for http2, and do not add any command line parameters for 3.5. What do you think?

ahrtr: Probably @liggitt doesn't have the full context, so let me clarify. Previously in PR [8238](https://github.com/etcd-io/etcd/pull/8238), the `grpc.MaxConcurrentStreams` was set to `math.MaxUint32` by default. But the PR did not change the [MaxConcurrentStreams for http2](https://cs.opensource.google/go/x/net/+/master:http2/server.go;l=99;drc=1d687d428aca0546c0ca84160c8700ee521e9fb9;bpv=0;bpt=1).

etcd [encapsulates the gRPC in https](https://github.com/etcd-io/etcd/blob/248f072370628fa2ad7c3d73e3935ba5636c0442/server/embed/serve.go#L166-L172), which means each gRPC request goes through httpHandler --> gRPC. Sicne the [default MaxConcurrenetStreams is 250](https://cs.opensource.google/go/x/net/+/master:http2/server.go;drc=1d687d428aca0546c0ca84160c8700ee521e9fb9;l=58) for http2, so actually the max concurrent stream each client can open at a time is actually 250 instead of math.MaxUint32 over https. 

Please let us know if you have any comments or concerns.  thx. also cc @dims

neolit123: Lgtm

nic-chen:
Thank you all for making this change, it helps a lot for Apache APISIX users!

PR #14171, opened 2022-06-28
raft: Remove the redundant `append` execution

SimFG:
Avoid extra `slice` creation of the `Entry`

### DIFF
--- a/raft/log_unstable.go
+++ b/raft/log_unstable.go
@@ -135,8 +135,7 @@ func (u *unstable) truncateAndAppend(ents []pb.Entry) {
 		// truncate to after and copy to u.entries
 		// then append
 		u.logger.Infof("truncate the unstable entries before index %d", after)
-		u.entries = append([]pb.Entry{}, u.slice(u.offset, after)...)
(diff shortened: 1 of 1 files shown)


raft/log_unstable.go
```diff
@@ -135,8 +135,7 @@ func (u *unstable) truncateAndAppend(ents []pb.Entry) {
 		// truncate to after and copy to u.entries
 		// then append
 		u.logger.Infof("truncate the unstable entries before index %d", after)
-		u.entries = append([]pb.Entry{}, u.slice(u.offset, after)...)
-		u.entries = append(u.entries, ents...)
+		u.entries = append(u.slice(u.offset, after), ents...)
```
ahrtr, line 138: There is a difference. Existing implementation will always allocate new space (actually array) for the new slice, but your code might reuse the same array.  

Although both can work, let try to keep it as it's because raft is a stable and important package.

SimFG: Yes, both can work. I think reusing the same array is safe. Because if it's unsafe, the slice will not be used. But I agree with you, I will close this pull request.

PR #14173, opened 2022-06-28
ectdctl: Add the `SilenceErrors` config For the `cobra.Command`

SimFG:
Avoid repeated printing of error messages.

before
image
after
image
related source code
```
// SilenceErrors is an option to quiet errors down stream.
SilenceErrors bool
```
link: https://github.com/spf13/cobra/blob/bba9331d4e5e5ebe2d0fdf0d9432b4b1e67d5061/command.go#L972-L975
image

### DIFF
--- a/etcdctl/ctlv3/ctl.go
+++ b/etcdctl/ctlv3/ctl.go
@@ -16,6 +16,7 @@
 package ctlv3
 
 import (
+	"os"
 	"time"
 
 	"go.etcd.io/etcd/api/v3/version"
@@ -114,7 +115,11 @@ func Start() error {
 
 func MustStart() {
 	if err := Start(); err != nil {
-		cobrautl.ExitWithError(cobrautl.ExitError, err)
+		if rootCmd.SilenceErrors {
+			cobrautl.ExitWithError(cobrautl.ExitError, err)
+		} else {
+			os.Exit(cobrautl.ExitError)
+		}
 	}
 }
 


ahrtr:
The PR can remove the duplicated error, but it's also missing the useful info `Run 'etcdctl --help' for usage.` printed by `spf13/cobra`.

It's indeed a real issue although it's minor. It seems that we need to figure out how to differentiate errors between `spf13/cobra` and etcd itself.  For the errors from `spf13/cobra` (such as command name typos in this case), we need the the useful info `Run 'etcdctl --help' for usage.`, otherwise it isn't needed.  But from another perspective, it seems that it isn't worthwhile to spend too much effort for such a small benefit. 

Anther concern is that the error messages are visible to end users, and we should be cautious about it.

A big goal is to sort out the error handling between `etcdctl/etcdurl` and `spf13/cobra`.

SimFG:
@ahrtr I have an idea. We can use `os.Exit(cobrautl.ExitError)` to replace `cobrautl.ExitWithError(cobrautl.ExitError, err)` when the `rootCmd` get error. Because the `unknown command` only occurs when the `rootCmd` finds a sub command.
image

ahrtr:
How about to update the PR something like below? 

```
func MustStart() {
	if err := Start(); err != nil {
		if rootCmd.SilenceErrors {
			cobrautl.ExitWithError(cobrautl.ExitError, err)
		} else {
			os.Exit(cobrautl.ExitError)
		}
	}
}

```

SimFG:
@ahrtr I think it's good. Have updated.

ahrtr (approved):

ptabor (approved):

spzala (approved):
This is a good improvement. Thanks @SimFG @ahrtr

Issue #14174, opened 2022-06-28
Peer member handler could return stable member list and cause new added node fails to bootstrap


chaochn47:
### What happened?

One of 3 member etcd production clusters update failed. 

The update is removing a old member, adding a new member and start the etcd process just like what is descried in https://etcd.io/docs/v3.5/op-guide/runtime-configuration/#cluster-reconfiguration-operations 

The error trace indicated the new node failed bootstrap at T1 at here
https://github.com/etcd-io/etcd/blob/7a1cecd5fafcea03adfb11b6bed49e901cd1dcf2/server/etcdserver/api/membership/cluster.go#L703-L708

From the etcd servers logging, we realized that local (new) node member count is 3 (passing from command line flags `--intial-cluster`) at T1

Leader node returned 2 member list at [peerMemberHandler](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/etcdhttp/peer.go#L105-L120) to the new node at T1. 

The above logging explained the error message. 

---

Leader node had around 200ms delay applied the `member add` configuration to the membership store at T2. 
Most likely due to the other overloaded client traffic. 

The other node which accepted this `member add` request returned this write instantly once it applied to its store at T0 while leader processes this `member add` asynchronously.  

### What did you expect to happen?

After a new member is added, all of etcd servers will return member list with the new member added when serving member list API in 
https://github.com/etcd-io/etcd/blob/7a1cecd5fafcea03adfb11b6bed49e901cd1dcf2/server/etcdserver/api/etcdhttp/peer.go#L105-L121

Otherwise, it breaks etcd strong consistency guarantee. 

### How can we reproduce it (as minimally and precisely as possible)?

Add a gofail failpoint to member add, build etcd with failpoint turned on. 

https://github.com/etcd-io/etcd/blob/7a1cecd5fafcea03adfb11b6bed49e901cd1dcf2/server/etcdserver/api/membership/cluster.go#L391-L412

Inject a sleep(0.2) delay. 

Re-do https://etcd.io/docs/v3.5/op-guide/runtime-configuration/#cluster-reconfiguration-operations 

### Anything else we need to know?

Retry in the new node provisioning should succeed and successfully bootstrap. 

However, it did reveal the issue etcd peer member handler could return stale member list. 

[...]

PR #14175, opened 2022-06-29
serve peer linearizable member list

chaochn47:
Related to https://github.com/etcd-io/etcd/issues/14174

To add a meaningful test, a new failpoint and fail case may be introduced to the functional test cases. I think it's a overkill.. 

Appreciate any comment or suggestion how to test this change!! Thanks

---
I will add another commit to update the CHANGELOG to make current one backportable.

### DIFF
--- a/server/etcdserver/api/etcdhttp/peer.go
+++ b/server/etcdserver/api/etcdhttp/peer.go
@@ -15,6 +15,7 @@
 package etcdhttp
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -44,7 +45,7 @@ func NewPeerHandler(lg *zap.Logger, s etcdserver.ServerPeerV2) http.Handler {
 
 func newPeerHandler(
 	lg *zap.Logger,
-	s etcdserver.Server,
+	s etcdserver.ServerV2,
 	raftHandler http.Handler,
 	leaseHandler http.Handler,
 	hashKVHandler http.Handler,
@@ -53,7 +54,7 @@ func newPeerHandler(
 	if lg == nil {
 		lg = zap.NewNop()
 	}
-	peerMembersHandler := newPeerMembersHandler(lg, s.Cluster())
+	peerMembersHandler := newPeerMembersHandler(lg, s)
 	peerMemberPromoteHandler := newPeerMemberPromoteHandler(lg, s)
 
 	mux := http.NewServeMux()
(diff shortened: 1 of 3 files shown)


chaochn47:
Both Integration tests and grpcproxy integration tests failed at the following 2 cases.

```
=== FAIL: integration TestDoubleClusterSizeOf1 (5.58s)
     logger.go:130: 2022-06-29T01:20:49.363Z	WARN	m0	timed out waiting for read index response (local node might have slow network)	{"member": "m0", "timeout": "5.2s"}
    logger.go:130: 2022-06-29T01:20:49.363Z	WARN	m0	failed to serve linearizable member list	{"member": "m0", "error": "etcdserver: request timed out"}
    logger.go:130: 2022-06-29T01:20:49.363Z	WARN	m0	unexpected v2 response error	{"member": "m0", "remote-addr": "@", "internal-server-error": "etcdserver: request timed out"}
    logger.go:130: 2022-06-29T01:20:49.363Z	WARN	m1	failed to unmarshal cluster response	{"member": "m1", "address": "unix://127.0.0.1:2103205737/members", "error": "json: cannot unmarshal object into Go value of type []*membership.Member"}
    cluster.go:317: failed to initialize the etcd server: cannot fetch cluster info from peer urls: could not retrieve cluster information from the given URLs

=== FAIL: integration TestIssue3699 (30.70s)
...
```

I am suspecting it's the test itself breaks. Will spend some time investigating the problem.

chaochn47:
Looks like in case of growing etcd member size from 1 to 2, Linearizable read notify will hang up which is expected. 

We may want to fail open in this case and expect up to 7s bootstrap delay in such scenario.. @ahrtr

ahrtr:
You might need to use [waitAppliedIndex](https://github.com/etcd-io/etcd/blob/7a1cecd5fafcea03adfb11b6bed49e901cd1dcf2/server/etcdserver/v3_server.go#L259).

See example at [v3_server.go#L281](https://github.com/etcd-io/etcd/blob/7a1cecd5fafcea03adfb11b6bed49e901cd1dcf2/server/etcdserver/v3_server.go#L281)

chaochn47:
> or keep it as it's for now, because when K8s or users retry and then it will work. Users may also add some sleep between adding the new node and starting the new node if they have a automated operator. 

Yeah, it sounds like a feasible solution and I also expect retry should succeed eventually. 

If this solution is picked, we should document this behavior and call this out in https://etcd.io/docs/v3.5/op-guide/runtime-configuration/#error-cases-when-adding-members.

PR #14182, opened 2022-06-29
Client: fix check for WithPrefix op

### LINKED ISSUE #14056: Client: IsOptsWithPrefix returns false even if WithPrefix() is included
### What happened?

If you use the function `IsOptsWithPrefix` with opts containing `WithPrefix()` it always returns `false` instead of true.

`IsOptsWithPrefix` should return true if `opts` contains `WithPrefix()`

```go
package main

import (
  "log"
  etcd "go.etcd.io/etcd/client/v3"
)

func main() {
  log.Printf("%v", etcd.IsOptsWithPrefix([]etcd.OpOption{etcd.WithPrefix()}))
}
```

Logs `false` instead of `true`

`IsOptsWithPrefix` creates an empty op with `NewOp` which sets a key with len 0. Therefore the len check in `WithPrefix` also is true and the flag is not set.

Not related to etcd backend, only client. Tested versions: v3.5.4, v3.6.0-alpha.0

_No response_

_No response_

### Relevant log output

_No response_

spzala:
Make sure that WithPrefix correctly set the flag, and add test.
Also, add test for WithFromKey.

fixes #14056

### DIFF
--- a/client/v3/op.go
+++ b/client/v3/op.go
@@ -389,12 +389,12 @@ func getPrefix(key []byte) []byte {
 // can return 'foo1', 'foo2', and so on.
 func WithPrefix() OpOption {
 	return func(op *Op) {
+		op.isOptsWithPrefix = true
 		if len(op.key) == 0 {
 			op.key, op.end = []byte{0}, []byte{0}
 			return
 		}
 		op.end = getPrefix(op.key)
-		op.isOptsWithPrefix = true
 	}
 }
 
--- a/client/v3/op_test.go
+++ b/client/v3/op_test.go
@@ -75,3 +75,27 @@ func TestIsSortOptionValid(t *testing.T) {
 		}
 	}
 }
+
+func TestIsOptsWithPrefix(t *testing.T) {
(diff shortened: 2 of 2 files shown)


spzala:
cc @serathius @ptabor @ahrtr

client/v3/op.go
```diff
@@ -391,6 +391,7 @@ func WithPrefix() OpOption {
 	return func(op *Op) {
 		if len(op.key) == 0 {
 			op.key, op.end = []byte{0}, []byte{0}
+			op.isOptsWithPrefix = true
```
ahrtr, line 394: minor comment: Both line 394 and 397 set `isOptsWithPrefix` as `true`, so I suggest to move 			`op.isOptsWithPrefix = true` to the beginning of the function.

spzala: @ahrtr Good catch, I missed it. Thanks!

ahrtr (approved):
LGTM

serathius (approved):

spzala:
@ahrtr @serathius thanks for the quick review! Build is green so merging.

PR #14198, opened 2022-07-07
lease: Rename `Poll` to `Peek` in the `LeaseExpiredNotifier`

SimFG:
`Poll` means that the elements in the heap will be removed.
And it is more appropriate to use `Peek` to get the top of the heap.

### DIFF
--- a/server/lease/lease_queue.go
+++ b/server/lease/lease_queue.go
@@ -96,7 +96,7 @@ func (mq *LeaseExpiredNotifier) Unregister() *LeaseWithTime {
 	return item
 }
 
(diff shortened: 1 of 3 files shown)


serathius (approved):

serathius (commented):
I did some Googling and found that `peek` is much more popular way to name this function.

SimFG:
@serathius yeah, I think it's better.

ahrtr:
Both `Peek` and `Top` are OK, but actually `Top` fits more in this case, because we only want to check the status of the top 1 entry and it's clearer. 

It isn't a big deal, so approving this PR.

ahrtr (approved):
Thank you, @SimFG

PR #14200, opened 2022-07-08
set RACE as true for linux-amd64-unit and linux-amd64-grpcproxy

ahrtr:
We should add some pipelines with RACE enabled. 

Part of https://github.com/etcd-io/etcd/issues/14105 


ahrtr:
cc  @serathius @spzala @ptabor

PR #12786, opened 2021-03-18
client: Bring back ETCD_CLIENT_DEBUG variable interpretation.

ptabor:
env ETCD_CLIENT_DEBUG supports log levels (debug, info, warn, error, dpanic, panic, fatal).

Only when set, overrides application-wide grpc logging settings.

PR #14203, opened 2022-07-08
clientv3: Fix parsing of ETCD_CLIENT_DEBUG

Jille:
It checked `err == nil` rather than `err != nil`.

@ptabor

### DIFF
--- a/client/v3/logger.go
+++ b/client/v3/logger.go
@@ -52,7 +52,7 @@ func etcdClientDebugLevel() zapcore.Level {
 	}
 	var l zapcore.Level
 	if err := l.Set(envLevel); err != nil {
-		log.Printf("Deprecated env ETCD_CLIENT_DEBUG value. Using default level: 'info'")
+		log.Printf("Invalid value for environment variable 'ETCD_CLIENT_DEBUG'. Using default level: 'info'")
 		return zapcore.InfoLevel
 	}
 	return l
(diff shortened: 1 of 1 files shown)


client/v3/logger.go
```diff
@@ -51,7 +51,7 @@ func etcdClientDebugLevel() zapcore.Level {
 		return zapcore.InfoLevel
 	}
 	var l zapcore.Level
-	if err := l.Set(envLevel); err == nil {
+	if err := l.Set(envLevel); err != nil {
```
ahrtr, line 54: The change should be good. But the message below is a little confusing. Is `ETCD_CLIENT_DEBUG` really deprecated? cc @serathius 
```
Deprecated env ETCD_CLIENT_DEBUG value. Using default level: 'info'
```

ahrtr:
Thanks @Jille for the PR, please sign off the commit,
```
git commit --signoff --amend
```

client/v3/logger.go
Jille: I parsed that sentence as the _value_ being deprecated, rather than the setting.

https://github.com/etcd-io/etcd/pull/12786 introduced this code and explicitly brought back ETCD_CLIENT_DEBUG.

ahrtr: Doesn't the following message  make more sense? @ptabor 
```
Invalid value for 'ETCD_CLIENT_DEBUG'. Using default level: 'info'
```

ptabor: Yes. That would be definitely better.

ahrtr (approved):
LGTM

Thanks both @ptabor  and @Jille

ahrtr:
@Jille Could you backport this PR to 3.5? cc  @serathius

Jille:
Sure: https://github.com/etcd-io/etcd/pull/14222

PR #14205, opened 2022-07-11
[3.4] Update release scripts for release-3.4

ahrtr:


ahrtr:
cc @serathius

serathius (approved):

serathius:
Should we consider backporting release test?

ahrtr:
> Should we consider backporting release script test?

Yes, it's a good point. Let me do it in a separate PR after merging this one.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
+name: Release
+on: [push, pull_request]
+jobs:
+  main:
```
serathius, line 4:
```suggestion
  release-3.4:
```

ahrtr: You saved me at least 30 minutes:)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
...
+        %no-protection
+        Key-Type: 1
+        Key-Length: 2048
+        Subkey-Type: 1
+        Subkey-Length: 2048
+        Name-Real: Github Action
+        Name-Email: github-action@etcd.io
+        Expire-Date: 0
+        EOF
+        DRY_RUN=true ./scripts/release.sh --no-upload --no-docker-push 3.4.99
```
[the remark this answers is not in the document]
ahrtr: Yes, I saw the difference, but I generate the BRANCH in release.sh (line 37),
```
BRANCH="release-${MINOR_VERSION}"
```

.github/workflows/release.yaml
serathius: Hmm, not sure this worked as intended as there is no longer a `Release` run in checks below. The variable here is only name of the job. It doesn't influence anything so maybe it's worth to revert it to `main` or any other name that is less confusing.

.github/workflows/release.yaml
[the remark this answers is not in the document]
serathius: ack

.github/workflows/release.yaml
ahrtr: I realized that I must merge this PR firstly, otherwise the release test pipeline will always fail, because it always clone relealse-3.4, which doesn't contain this PR yet. 

So let me merge this PR firstly.

PR #14208, opened 2022-07-11
Print clusterID, memberID and leaseID in hexdecimal

ahrtr:
ClusterID, memberID and leaseID are just identities, so it doesn't make sense to print them in decimal. Instead, we should always print them in hexadecimal. 

Please note the existing behavior isn't consistent. Sometimes etcdctl prints them in hexdecimal (such as `etcdctl endpoint status -w table`), but sometimes in decimal (such as `etctctl endpoint status -w fields`). 

## Previous output
```
wachao-a01:bin wachao$ ./etcdctl  member list -w fields
"ClusterID" : 14841639068965178418
"MemberID" : 10276657743932975437
"RaftTerm" : 5
"ID" : 10276657743932975437
"Name" : "default"
"PeerURL" : "http://localhost:2380"
"ClientURL" : "http://localhost:2379"
"IsLearner" : false

wachao-a01:bin wachao$ ./etcdctl endpoint status -w fields
"ClusterID" : 14841639068965178418
"MemberID" : 10276657743932975437
"Revision" : 3
"RaftTerm" : 5
"Version" : "3.6.0-alpha.0"
"StorageVersion" : "3.6.0"
"DBSize" : 24576
"DBSizeInUse" : 24576
"Leader" : 10276657743932975437
"IsLearner" : false
"RaftIndex" : 23
"RaftTerm" : 5
"RaftAppliedIndex" : 23
"Errors" : []
"Endpoint" : "127.0.0.1:2379"

wachao-a01:bin wachao$ ./etcdctl  lease list -w fields
"ClusterID" : 14841639068965178418
"MemberID" : 10276657743932975437
"Revision" : 3
"RaftTerm" : 8
"ID" : 7587863798256828424
"ID" : 7587863798360740612
"ID" : 7587863798360740614
"ID" : 7587863798256828422
```

## New output
```
wachao-a01:bin wachao$ ./etcdctl member list -w fields
"ClusterID" : cdf818194e3a8c32
"MemberID" : 8e9e05c52164694d
"RaftTerm" : 7
"ID" : 8e9e05c52164694d
"Name" : "default"
"PeerURL" : "http://localhost:2380"
"ClientURL" : "http://localhost:2379"
"IsLearner" : false

wachao-a01:bin wachao$ ./etcdctl endpoint status -w fields
"ClusterID" : cdf818194e3a8c32
"MemberID" : 8e9e05c52164694d
"Revision" : 3
"RaftTerm" : 7
"Version" : "3.6.0-alpha.0"
"StorageVersion" : "3.6.0"
"DBSize" : 24576
"DBSizeInUse" : 24576
"Leader" : 10276657743932975437
"IsLearner" : false
"RaftIndex" : 27
"RaftTerm" : 7
"RaftAppliedIndex" : 27
"Errors" : []
"Endpoint" : "127.0.0.1:2379"

wachao-a01:bin wachao$ ./etcdctl  lease list -w fields
"ClusterID" : cdf818194e3a8c32
"MemberID" : 8e9e05c52164694d
"Revision" : 3
"RaftTerm" : 7
"ID" : 694d81ec1b250008
"ID" : 694d81ec1b250006
```

### DIFF
--- a/etcdctl/ctlv3/command/printer_fields.go
+++ b/etcdctl/ctlv3/command/printer_fields.go
@@ -19,10 +19,14 @@ import (
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	spb "go.etcd.io/etcd/api/v3/mvccpb"
+	"go.etcd.io/etcd/client/pkg/v3/types"
 	v3 "go.etcd.io/etcd/client/v3"
 )
 
-type fieldsPrinter struct{ printer }
+type fieldsPrinter struct {
+	printer
+	isHex bool
+}
 
 func (p *fieldsPrinter) kv(pfx string, kv *spb.KeyValue) {
 	fmt.Printf("\"%sKey\" : %q\n", pfx, string(kv.Key))
@@ -30,12 +34,21 @@ func (p *fieldsPrinter) kv(pfx string, kv *spb.KeyValue) {
 	fmt.Printf("\"%sModRevision\" : %d\n", pfx, kv.ModRevision)
 	fmt.Printf("\"%sVersion\" : %d\n", pfx, kv.Version)
 	fmt.Printf("\"%sValue\" : %q\n", pfx, string(kv.Value))
-	fmt.Printf("\"%sLease\" : %d\n", pfx, kv.Lease)
+	if p.isHex {
+		fmt.Printf("\"%sLease\" : %016x\n", pfx, kv.Lease)
+	} else {
+		fmt.Printf("\"%sLease\" : %d\n", pfx, kv.Lease)
+	}
 }
 
 func (p *fieldsPrinter) hdr(h *pb.ResponseHeader) {
-	fmt.Println(`"ClusterID" :`, h.ClusterId)
-	fmt.Println(`"MemberID" :`, h.MemberId)
+	if p.isHex {
+		fmt.Println(`"ClusterID" :`, types.ID(h.ClusterId))
+		fmt.Println(`"MemberID" :`, types.ID(h.MemberId))
+	} else {
+		fmt.Println(`"ClusterID" :`, h.ClusterId)
+		fmt.Println(`"MemberID" :`, h.MemberId)
+	}
 	// Revision only makes sense for k/v responses. For other kinds of
 	// responses, i.e. MemberList, usually the revision isn't populated
 	// at all; so it would be better to hide this field in these cases.
@@ -99,7 +112,11 @@ func (p *fieldsPrinter) Watch(resp v3.WatchResponse) {
 
 func (p *fieldsPrinter) Grant(r v3.LeaseGrantResponse) {
 	p.hdr(r.ResponseHeader)
-	fmt.Println(`"ID" :`, r.ID)
+	if p.isHex {
+		fmt.Printf("\"ID\" : %016x\n", r.ID)
+	} else {
+		fmt.Println(`"ID" :`, r.ID)
+	}
 	fmt.Println(`"TTL" :`, r.TTL)
 }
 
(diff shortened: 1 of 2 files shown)


ahrtr:
cc @serathius @ptabor @spzala 

@liggitt @dims @neolit123 do you have any concerns from K8s perspective?

etcdctl/ctlv3/command/printer_fields.go
```diff
@@ -30,12 +30,12 @@ func (p *fieldsPrinter) kv(pfx string, kv *spb.KeyValue) {
 	fmt.Printf("\"%sModRevision\" : %d\n", pfx, kv.ModRevision)
 	fmt.Printf("\"%sVersion\" : %d\n", pfx, kv.Version)
 	fmt.Printf("\"%sValue\" : %q\n", pfx, string(kv.Value))
-	fmt.Printf("\"%sLease\" : %d\n", pfx, kv.Lease)
+	fmt.Printf("\"%sLease\" : %016x\n", pfx, kv.Lease)
 }
 
 func (p *fieldsPrinter) hdr(h *pb.ResponseHeader) {
-	fmt.Println(`"ClusterID" :`, h.ClusterId)
```
ptabor, line 37: Can we cast it to `types.Id` to benefit from the common stringer implementation: 

(https://github.com/etcd-io/etcd/blob/53bfdc11d8f6d6cdf0ccb986f052e9739086110b/client/pkg/types/id.go#L22-L24) ?

ptabor (changes requested):
I agree it would be better to introduce this in more backward compatible way. 
The fastest path I see: 
  1. Introduce flag in 3.6 defaulting to 'hex'
  2. Backport the PR to 3.5, but change the flag there to `decimal`. In 3.5, when the flag is unset, there should be a warning about behavioral change in 3.6
  3. Remove the flag in 3.7 (or 4).

etcdctl/ctlv3/command/printer_fields.go
ahrtr: Thanks @ptabor . It makes sense. But only `memberID` and `clusterID` are uint64, and `LeaseID` is int64. So I only cast `memberID` and `clusterID` `to types.ID`.

ahrtr:
Thanks @neolit123 and @ptabor for the feedback, which basically makes sense to me. 

Actually I thought about backward compatible, the reason I did not do it was that there is no any change on JSON output which should be the most common case for programmatically parsing the output. 

But anyway, I updated this PR and leverage the existing flag `--hex`

```
wachao-a01:bin wachao$ ./etcdctl  member list -w fields --hex
"ClusterID" : cdf818194e3a8c32
"MemberID" : 8e9e05c52164694d
"RaftTerm" : 2
"ID" : 8e9e05c52164694d
"Name" : "default"
"PeerURL" : "http://localhost:2380"
"ClientURL" : "http://localhost:2379"
"IsLearner" : false


wachao-a01:bin wachao$ ./etcdctl  lease list -w fields --hex
"ClusterID" : cdf818194e3a8c32
"MemberID" : 8e9e05c52164694d
"Revision" : 1
"RaftTerm" : 2
"ID" : 694d820437ab0a07
```

PTAL @ptabor @neolit123 @serathius  thx

spzala (commented):
@ahrtr extending the current flag that's used for json seems good approach! I have tested it my env to verify. I think we should add Changelog and also how about adding a test?  It's okay to add it in a separate PR if you want. Thanks!

ahrtr:
Thanks @spzala for the review comment, which basically makes sense to me. 

Points from my side:
1. Unfortunately, there is no any unit tests for any command so far. So I will investigate how to add unit test and add an example case firstly in a separate PR; afterwards, let's get more contributors involved to add unit tests for all commands. What do you think?
2. With regard the Changelog, I intentionally did not do it, because there are some other related changes (such as https://github.com/etcd-io/etcd/pull/14120#issuecomment-1195159939 , and the JSON output is also printing decimal for MemberID etc.) to be done. So I plan to add the ChangeLog after all the related items are done.

spzala:
> Thanks @spzala for the review comment, which basically makes sense to me.
> 
> Points from my side:
> 
>     1. Unfortunately, there is no any unit tests for any command so far. So I will investigate how to add unit test and add an example case firstly in a separate PR; afterwards, let's get more contributors involved to add unit tests for all commands. What do you think?
> 
>     2. With regard the Changelog, I intentionally did not do it, because there are some other related changes (such as [server: Implement compaction hash checking #14120 (comment)](https://github.com/etcd-io/etcd/pull/14120#issuecomment-1195159939) , and the JSON output is also printing decimal for MemberID etc.) to be done. So I plan to add the ChangeLog after all the related items are done.

@ahrtr thank you, and both the points you mentioned sounds good to me. I will be happy to share work on adding unit test coverage.

spzala (approved):

spzala:
@ahrtr considering @ptabor is away for sometime and his comments are addressed, we should be good with merging changes, if that sounds good to you?. Thanks!

ahrtr:
Thanks @spzala . 

Actually this PR should be a small and safe change because there is no any change on the JSON output so far, so let me merge it.

PR #14212, opened 2022-07-12
benchmark: Fix the panic when running `benchmark watch`

SimFG:
When executing the `go run main.go watch` in the `tools/benchmark` dir, error:
image
The default value of `watchPutRate` is `0`, we get a `Limiter` and the `limit` property of the `Limiter` object is 0 and the `burst` property is 1.
`limit := rate.NewLimiter(rate.Limit(watchPutRate), 1)` 
Each time running the code `limit.Wait(context.TODO())`, the value of the burst property will be decremented by one. Therefore, when the second `wait` is executed, the value of burst will become 0, which is less than the limit, and an error will be reported at this time.
```
limit := rate.NewLimiter(rate.Limit(watchPutRate), watchPutTotal)
	for _, cc := range clients {
		go func(c *clientv3.Client) {
			for op := range putreqc {
				if err := limit.Wait(context.TODO()); err != nil {
					panic(err)
				}
				if _, err := c.Do(context.TODO(), op); err != nil {
					panic(err)
				}
			}
		}(cc)
	}
```
```
func (lim *Limiter) Wait(ctx context.Context) (err error) {
	return lim.WaitN(ctx, 1)
}

func (lim *Limiter) WaitN(ctx context.Context, n int) (err error) {
	// The test code calls lim.wait with a fake timer generator.
	// This is the real timer generator.
	newTimer := func(d time.Duration) (<-chan time.Time, func() bool, func()) {
		timer := time.NewTimer(d)
		return timer.C, timer.Stop, func() {}
	}

	return lim.wait(ctx, n, time.Now(), newTimer)
}

func (lim *Limiter) wait(ctx context.Context, n int, now time.Time, newTimer func(d time.Duration) (<-chan time.Time, func() bool, func())) error {
	// ...

	if n > burst && limit != Inf {
		return fmt.Errorf("rate: Wait(n=%d) exceeds limiter's burst %d", n, burst)
	}
	// ...
	// Reserve
	r := lim.reserveN(now, n, waitLimit)
	// ...
}

func (lim *Limiter) reserveN(now time.Time, n int, maxFutureReserve time.Duration) Reservation {
	lim.mu.Lock()
	defer lim.mu.Unlock()

	if lim.limit == Inf {
		// ...
	} else if lim.limit == 0 {
		var ok bool
		if lim.burst >= n {
			ok = true
			lim.burst -= n
		}
		return Reservation{
			ok:        ok,
			lim:       lim,
			tokens:    lim.burst,
			timeToAct: now,
		}
	}
        // ...

	return r
}
```

### DIFF
--- a/tools/benchmark/cmd/watch.go
+++ b/tools/benchmark/cmd/watch.go
@@ -23,7 +23,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"go.etcd.io/etcd/client/v3"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/pkg/v3/report"
 
 	"github.com/cheggaaa/pb/v3"
@@ -209,7 +209,12 @@ func benchPutWatches(clients []*clientv3.Client, wk *watchedKeys) {
 		}
 	}()
 
-	limit := rate.NewLimiter(rate.Limit(watchPutRate), 1)
+	watchPutLimit := rate.Inf
+	if watchPutRate > 0 {
+		watchPutLimit = rate.Limit(watchPutRate)
+	}
+
+	limit := rate.NewLimiter(watchPutLimit, 1)
 	for _, cc := range clients {
 		go func(c *clientv3.Client) {
 			for op := range putreqc {


tools/benchmark/cmd/watch.go
```diff
@@ -209,7 +209,7 @@ func benchPutWatches(clients []*clientv3.Client, wk *watchedKeys) {
 		}
 	}()
 
-	limit := rate.NewLimiter(rate.Limit(watchPutRate), 1)
+	limit := rate.NewLimiter(rate.Limit(watchPutRate), watchPutTotal)
```
ptabor, line 212: I assume that the intent is to have the rate of `put's` being effectively unlimitted. 

Can we use `rate.Inf` constant (https://pkg.go.dev/golang.org/x/time/rate#pkg-constants) that seems to be dedicated for such scenario. Then watchPutTotal will not have really any meaning: 

```
func NewLimiter(r [Limit](https://pkg.go.dev/golang.org/x/time/rate#Limit), b [int](https://pkg.go.dev/builtin#int)) *[Limiter](https://pkg.go.dev/golang.org/x/time/rate#Limiter)

As a special case, if r == Inf (the infinite rate), b is ignored.
```

so the code would be like: 

```
watchPutLimit:=rate.Inf
if watchPutRate>0 {
  watchPutLimit=rate.Limit(watchPutRate);
}
 
limit := rate.NewLimiter(watchPutLimit, watchPutTotal)
```

SimFG: Thanks your suggestion.

ahrtr (approved):
LGTM 

Thank you!

SimFG:
@ptabor I found a small problem, this value of `burst` will guarantee the number of tokens acquired between two time intervals. When it is obtained for the first time, because the `lastTime` is infinitely small, the difference between this time and the current time is very large, and the number of tokens will go to the `burst` value at this time. If the value of `burst` is set to `total`, the `put-rate` parameter will be invalid at this time, because the number of tokens at this time is directly the total number, which will be consumed after the program runs.
```
// advance calculates and returns an updated state for lim resulting from the passage of time.
// lim is not changed.
// advance requires that lim.mu is held.
func (lim *Limiter) advance(now time.Time) (newNow time.Time, newLast time.Time, newTokens float64) {
	last := lim.last
	if now.Before(last) {
		last = now
	}

	// Calculate the new number of tokens, due to time that passed.
	elapsed := now.Sub(last)
	delta := lim.limit.tokensFromDuration(elapsed)
	tokens := lim.tokens + delta
	if burst := float64(lim.burst); tokens > burst {
		tokens = burst
	}
	return now, last, tokens
}
```

SimFG:
@serathius  Maybe I need your approval.

cameronelliott:
@SimFG Just curious whats the status of this? These changes didn't get merged? I ran into this issue, so I am trying to understand this. Thank you

ahrtr:
@SimFG  One PR is supposed to fix only one thing.  Please only fix the benchmark tool issue in this PR, and also remove the .DS_store file (feel free to get it included in the `.gitignore`).

SimFG:
> @SimFG One PR is supposed to fix only one thing. Please only fix the benchmark tool issue in this PR, and also remove the .DS_store file (feel free to get it included in the `.gitignore`).

Updated. After closing this pull request, I accidentally submitted some test code.

ahrtr:
Confirmed that it's working, thank you!  @SimFG

cameronelliott:
Amazing! thanks to @SimFG and other maintainers for making etcd great.

PR #14221, opened 2022-07-13
etcd-mixin:add etcdDatabaseHighFragmentationRatio alert

Elbehery:
This PR add an alert `etcdDatabaseHighFragmentationRatio `. 

As the size of the backend storage disk space in use is less than half of the actual allocated disk space, this alerts helps admins to know when to run compaction and defragmentation. 

cc @hexfusion @ahrtr @ptabor @wking @simonpasquier

### DIFF
--- a/contrib/mixin/mixin.libsonnet
+++ b/contrib/mixin/mixin.libsonnet
@@ -238,6 +238,21 @@
               summary: 'etcd cluster database growing very fast.',
             },
           },
+          {
+            alert: 'etcdDatabaseHighFragmentationRatio',
+            expr: |||
+              (last_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / last_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'etcd cluster "{{ $labels.%s }}": database size in use on instance {{ $labels.instance }} is {{ $value | humanizePercentage }} of the actual allocated disk space, please run defragmentation (e.g. etcdctl defrag) to retrieve the unused fragmented disk space.' % $._config.clusterLabel,
+              summary: 'etcd database size in use is less than 50% of the actual allocated storage.',
+              runbook_url: 'https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation',
+            },
+          },
         ],
       },
     ],
--- a/contrib/mixin/test.yaml
+++ b/contrib/mixin/test.yaml
@@ -139,3 +139,26 @@ tests:
             exp_annotations:
               description: 'etcd cluster "etcd": Predicting running out of disk space in the next four hours, based on write observations within the past four hours on etcd instance 10.10.10.0, please check as it might be disruptive.'
               summary: 'etcd cluster database growing very fast.'
+
+  - interval: 1m
+    input_series:
+      - series: 'etcd_mvcc_db_total_size_in_use_in_bytes{job="etcd",instance="10.10.10.0"}'
+        values: '30000+0x10'
+      - series: 'etcd_mvcc_db_total_size_in_bytes{job="etcd",instance="10.10.10.0"}'
+        values: '100000+0x10'
+      - series: 'etcd_mvcc_db_total_size_in_use_in_bytes{job="etcd",instance="10.10.10.1"}'
+        values: '70000+0x10'
+      - series: 'etcd_mvcc_db_total_size_in_bytes{job="etcd",instance="10.10.10.1"}'
+        values: '100000+0x10'
+    alert_rule_test:
+      - eval_time: 11m
+        alertname: etcdDatabaseHighFragmentationRatio
+        exp_alerts:
+          - exp_labels:
+              instance: '10.10.10.0'
+              job: etcd
+              severity: warning
+            exp_annotations:
+              description: 'etcd cluster "etcd": database size in use on instance 10.10.10.0 is 30% of the actual allocated disk space, please run defragmentation (e.g. etcdctl defrag) to retrieve the unused fragmented disk space.'
+              runbook_url: https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation
+              summary: 'etcd database size in use is less than 50% of the actual allocated storage.'


ahrtr:
Overall looks good to me.  Currently we can merge this PR, but long-term, I prefer to the [long-term plan](https://github.com/etcd-io/etcd/pull/14196#issuecomment-1183341222) proposed by @ptabor and [my comment](https://github.com/etcd-io/etcd/pull/14196#issuecomment-1183738941) on top of that.

contrib/mixin/mixin.libsonnet
```diff
@@ -238,6 +238,20 @@
...
+            alert: 'etcdBackendHighFragmentationRatio',
+            expr: |||
+              (etcd_mvcc_db_total_size_in_use_in_bytes/etcd_mvcc_db_total_size_in_bytes)*100 < 50
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'etcd cluster "{{ $labels.%s }}": database size in use on instance {{ $labels.instance }} is less than half of the actual allocated disk space, please run compaction and defragmentation procedures to retrieve the unused fragmented disk space.' % $._config.clusterLabel,
```
ahrtr, line 251: Just one minor comment, suggest to change "`please run compaction and defragmentation procedures`" to "`please run defragmentation procedure`". 

Please note that compaction is only related to the historical data (MVCC), but defragmentation is only related to `size_in_use / size_allocated`.

ahrtr (approved):

contrib/mixin/mixin.libsonnet
Elbehery: @ahrtr done 👍🏽

contrib/mixin/mixin.libsonnet
```diff
@@ -238,6 +238,20 @@
...
+            alert: 'etcdBackendHighFragmentationRatio',
+            expr: |||
+              (etcd_mvcc_db_total_size_in_use_in_bytes/etcd_mvcc_db_total_size_in_bytes)*100 < 50
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'etcd cluster "{{ $labels.%s }}": database size in use on instance {{ $labels.instance }} is less than half of the actual allocated disk space, please run defragmentation procedure to retrieve the unused fragmented disk space.' % $._config.clusterLabel,
```
wking, line 251: are there docs we can link or a runbook or something via `runbook_url` to help folks like me who will think "I have no idea how to run a defrag"?  Maybe [here][1], with some way to get the current version?  Or following [this pattern][2] so mixin consumers can slot in their preferred runbook source?  No worries if this is something that should get punted to follow-up work.

[1]: https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation
[2]: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/a54540b92fb7b2d0826c0eff741e0894943babc7/lib/add-runbook-links.libsonnet#L13-L22

ahrtr: I think we can use link below, cc @spzala 
https://etcd.io/docs/latest/op-guide/maintenance/ 

+1 to add `runbookURLPattern` pointed out above. Please feel free to address it in a followed PR.

simonpasquier, line 251:
```suggestion
              description: 'etcd cluster "{{ $labels.%s }}": database size in use on instance {{ $labels.instance }} is {{ $value | humanizePercentage }} of the actual allocated disk space, please run defragmentation procedure to retrieve the unused fragmented disk space.' % $._config.clusterLabel,
```

contrib/mixin/mixin.libsonnet
```diff
@@ -238,6 +238,20 @@
               summary: 'etcd cluster database growing very fast.',
             },
           },
+          {
+            alert: 'etcdBackendHighFragmentationRatio',
+            expr: |||
+              (etcd_mvcc_db_total_size_in_use_in_bytes/etcd_mvcc_db_total_size_in_bytes)*100 < 50
```
simonpasquier, line 244: use of min_over_time() will account for transient failed scrapes.
```suggestion
              (min_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / min_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5
```

ahrtr: Will `'for': '10m'` workaround the transient failed scrape? 

Usually both `etcd_mvcc_db_total_size_in_use_in_bytes` and `etcd_mvcc_db_total_size_in_bytes` are monotonic increasing for KEY space unless there are some compaction & de-fragmentation operations. So the latest value or `max_over_time` makes more sense than `min_over_time`?

simonpasquier: > Will `'for': '10m'` workaround the transient failed scrape?

No. In case of a failed scrape, Prometheus will insert stale markers and the alert expression will return no data right after the failed scrape which will resolve the alert. So you'll see:
* alert firing
* failed scrape -> alert resovled
* scrape ok -> alert pending
* ...
* 10 minutes elapsed -> alert firing

> Usually both `etcd_mvcc_db_total_size_in_use_in_bytes` and `etcd_mvcc_db_total_size_in_bytes` are monotonic increasing for KEY space unless there are some compaction & de-fragmentation operations. So the latest value or `max_over_time` makes more sense than `min_over_time`?

in fact `last_over_time()` would be better in this case.

ahrtr: OK, then why not just `(etcd_mvcc_db_total_size_in_use_in_bytes/etcd_mvcc_db_total_size_in_bytes)*100 < 50 `instead of `(last_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / last_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5`?

ahrtr: It's an interesting & subtle discussion :). 

I am thinking should we use `last_over_time` for all similar cases (see an example below)? Which of course can be resolved in a separate PR if needed. 
```
(etcd_mvcc_db_total_size_in_bytes/etcd_server_quota_backend_bytes)*100 > 95
```

contrib/mixin/mixin.libsonnet
```diff
@@ -238,6 +238,20 @@
               summary: 'etcd cluster database growing very fast.',
             },
           },
+          {
+            alert: 'etcdBackendHighFragmentationRatio',
+            expr: |||
+              (min_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / min_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5
```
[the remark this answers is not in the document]
ahrtr: I agree `last_over_time` is better for this case.

contrib/mixin/mixin.libsonnet
```diff
@@ -238,6 +238,20 @@
               summary: 'etcd cluster database growing very fast.',
             },
           },
+          {
+            alert: 'etcdBackendHighFragmentationRatio',
```
spzala, line 242: Thanks @Elbehery - lgtm but couple of small suggestions 1) the `summary` and `description` mixes use of database and backend, we may want to stick with using database. Also, how about renaming `etcdBackendHighFragmentationRatio` to `etcdDatabaseHighFragmentationRatio` ?  2) in-lined comment below.

contrib/mixin/mixin.libsonnet
```diff
@@ -238,6 +238,20 @@
...
+            expr: |||
+              (min_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / min_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'etcd cluster "{{ $labels.%s }}": database size in use on instance {{ $labels.instance }} is {{ $value | humanizePercentage }} of the actual allocated disk space, please run defragmentation procedure to retrieve the unused fragmented disk space.' % $._config.clusterLabel,
```
spzala, line 251: Nit, we should use just `defragmentation` as it's a single command vs a procedure. I think we don't mention defragmentation as a procedure anywhere in the doc. So may be use something like `please run defragmentation (e.g. etcdctl defrag) to retrieve the...`). If we can provide link to defrag document as @ahrtr discussed in another comment that would be good otherwise the example using etcdctl I mentioned here should be good.

spzala (commented):
Couple of in-lined comments. Thanks!

contrib/mixin/mixin.libsonnet
Elbehery: @ahrtr sure thing, will raise a fix following this one 👍🏽

contrib/mixin/mixin.libsonnet
Elbehery: @spzala thanks a lot for your review, will fix it right away 👍🏽

contrib/mixin/mixin.libsonnet
Elbehery: @ahrtr I have created https://github.com/etcd-io/etcd/pull/14237 to address your request above

contrib/mixin/mixin.libsonnet
spzala: Thank you @Elbehery

spzala (approved):

PR #14238, opened 2022-07-19
Add process to remove maintainers that don't fulfill their duties

serathius:
When asking CNCF for support due to reduced community activity we stumbled upon a problem, 
governing board was surprised that project is unmaintained if there is so many maintainers listed.

Feedback from CNCF governing board was to improve
accuracy of maintainers list so it reflects only active maintainers.
Self nominating process it not enough and we need to start enforcing
moving maintainers to emeritus status. Period of 9 months to take maternity leave into account.

Signed-off-by: Marek Siarkowicz <siarkowicz@google.com>

cc @ptabor @ahrtr @spzala @gyuho @mitake @jingyih @jpbetz @hexfusion @wenjiaswe @xiang90 @bdarnell @tbg @lilic @wilsonwang371

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -42,25 +42,30 @@ below.
 ### Nomination and retiring of maintainers
 
 [Maintainers](./MAINTAINERS) file on the `main` branch reflects the latest
-state of project maintainers. Changes to list of maintainers should be done by
-opening a pull request and CCing all the existing maintainers.
+state of project maintainers. List of existing maintainers should be kept up to
+date by existing maintainers to properly reflect community health and to gain
+better understanding of recruiting need for new maintainers. Changes to list of
+maintainers should be done by opening a pull request and CCing all the existing
+maintainers.
 
 Contributors who are interested in becoming a maintainer, if performing relevant
-responsibilities, should discuss their interest with the existing maintainers. New
-maintainers must be nominated by an existing maintainer and must be elected by a
-supermajority of maintainers with a fallback on lazy consensus after three business weeks
-inactive voting period and as long as two maintainers are on board. 
+responsibilities, should discuss their interest with the existing maintainers.
+New maintainers must be nominated by an existing maintainer and must be elected
+by a supermajority of maintainers with a fallback on lazy consensus after three
+business weeks inactive voting period and as long as two maintainers are on board.
 
 Life priorities, interests, and passions can change. Maintainers can retire and
-be removed from [maintainers](./MAINTAINERS) list and added to
-[emeritus](./README.md#etcd-emeritus-maintainers). If a maintainer needs to step
-down, they should inform other maintainers, if possible, help find someone to
-pick up the related work. At the very least, ensure the related work can be
-continued. Afterward they can remove themselves from list of existing maintainers.
+move to the [emeritus status](./README.md#etcd-emeritus-maintainers). If a
+maintainer needs to step down, they should inform other maintainers, if possible,
+help find someone to pick up the related work. At the very least, ensure the
+related work can be continued. Afterward they can remove themselves from list of
+existing maintainers.
 
 If a maintainer is not been performing their duties for period of 12 months,
-they can be removed by other active maintainers. If an emeritus maintainer wants
-to regain an active role, they can do so after renewing their contributions.
+they can be removed by other maintainers. In that case inactive maintainer will
+be first notified via an email. If situation doesn't improve, they will be
+removed. If an emeritus maintainer wants to regain an active role, they can do
+so by renewing their contributions. Active maintainers should welcome such a move.
 Retiring of other maintainers or regaining the status should require approval
 of at least two active maintainers.
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -14,7 +14,7 @@ please see [CONTRIBUTING](./CONTRIBUTING.md) guide.
 
 ## Maintainers
 
-[Maintainers](./MAINTAINERS) are first and foremost contributors that have shown they
+Maintainers are first and foremost contributors that have shown they
 are committed to the long term success of a project. Maintainership is about building
 trust with the current maintainers of the project and being a person that they can
 depend on to make decisions in the best interest of the project in a consistent manner.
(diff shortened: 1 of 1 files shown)


serathius:
Note, possibly we should do the same for reviewers, but first we need to define their responsibilities.

hexfusion:
In the context of removal what is the process for reactivation? Can we elaborate/flesh this out a bit?

Folks shift projects, jobs etc.

I can see the motivation for cleaning up the list.

jpbetz:
Yes, I feel like I'm in the same boat as @hexfusion. I'm not active on the project and so I'd like to change status to something that reflects that, and also help ensure the active maintainers are easy to find and are appropriately recognized.  I can imagine that I might return to work on the project again in the future, so having some idea how that process would work would be valuable to me, and probably quite a few others on the list.

spzala (commented):
@serathius thanks for creating this PR for discussion. My thoughts would be: 
- Maintainer(s) with 12 months of inactive period, as decided by the active maintainers, should be moved to the emeritus status after discussing this topic with them by the active maintainers. (Note, Kubernetes project use 18 months of inactive period - https://github.com/kubernetes/community/blob/master/community-membership.md#inactive-members So I would think of 12 months vs 9 months but open to other opinions) 
- If any of the emeritus maintainers wish to contribute back to the project, they will be added back to maintainers list without going through a formal process. 
@hexfusion @jpbetz

ahrtr:
Basically I am on the same page as @spzala . The point is the existing maintainers have already demonstrated great expertise and big contribution to the community, they should be always welcome if they want to contribute back to the project.

The only minor comment is we still need to clearly clarify the process on how to add any maintainer from emeritus list to the maintainer list. If someone (assuming A) in the emeritus list want to be added back into the maintainer list, then my thought is that A is supposed to follow the process something like below,
- Keep active in the community at least one month;
- A doesn’t need to be nominated by any existing maintainer; Instead A can raise an issue in the community himself/herself, and cc all existing active (all inactive maintainers should have already been moved to the emeritus list) maintainers, and list all his/her contribution at least in the past month;
- If there is no any objections in 3 working days (I believe there will be no any objections if A indeed has already been active for at least one month), then A will be added back to the maintainer list.

spzala:
Thanks @ahrtr I agree with the overall thoughts and that we need to put this in the process. We know and trust that any emeritus maintainer/reviewer who can get back to active contributions means it, and so I think we should keep the process very simple (I am open to any more ideas but this is what I think): 
- When an emeritus maintainer(s) is back to contributing to the project and wants to be back to take the maintainer/reviewer role, they will open a PR to add themself back to the MAINTAINERS file.
- Per two or more existing maintainer reviews or lazy consensus of a working week or three days (as you suggest), a current maintainer will merge the changes and make needed GitHub admin-related changes.

GOVERNANCE.md
```diff
@@ -39,18 +39,30 @@ below.
...
+inactive voting period and as long as two maintainers are on board. 
+
+Life priorities, interests, and passions can change. Maintainers can retire and
+be removed from [maintainers](./MAINTAINERS) list and added to
+[emeritus](./README.md#etcd-emeritus-maintainers). If a maintainer needs to step
+down, they should inform other maintainers, if possible, help find someone to
+pick up the related work. At the very least, ensure the related work can be
+continued. Afterward they can remove themselves from list of existing maintainers.
+
+If a maintainer is not been performing their duties for period of 12 months,
```
ahrtr, line 61: `their` to `his/her`

minor comment: I see a couple of issues related to `singular` vs `plural`:)

serathius: My intention was to use https://en.wikipedia.org/wiki/Singular_they

ahrtr (approved):
Overall looks good to me.  Thanks @serathius 

cc @spzala to double check the English syntax.

GOVERNANCE.md
```diff
@@ -39,18 +39,30 @@ below.
...
-removed by a supermajority of the maintainers and moved to emeritus status.
-
-Life priorities, interests, and passions can change. If a maintainer needs to step
-down, inform other maintainers about this intention, and if possible, help find someone
-to pick up the related work. At the very least, ensure the related work can be continued.
-Afterward, create a pull request to remove yourself from the [MAINTAINERS](./MAINTAINERS)
-file.
+inactive voting period and as long as two maintainers are on board. 
+
+Life priorities, interests, and passions can change. Maintainers can retire and
```
spzala, line 54: Small re-write suggestion to "Maintainers can retire and move to the emeritus status. instead of "Maintainers can retire and be removed from [maintainers]...."

GOVERNANCE.md
spzala, line 61: Let's also add why such a step is needed -  "If a maintainer has not been performing their duties for period of 12 months, they can be moved to  emeritus status by active maintainers to gain better understanding of recruiting need for new maintainers."

GOVERNANCE.md
```diff
@@ -39,18 +39,30 @@ below.
...
+be removed from [maintainers](./MAINTAINERS) list and added to
+[emeritus](./README.md#etcd-emeritus-maintainers). If a maintainer needs to step
+down, they should inform other maintainers, if possible, help find someone to
+pick up the related work. At the very least, ensure the related work can be
+continued. Afterward they can remove themselves from list of existing maintainers.
+
+If a maintainer is not been performing their duties for period of 12 months,
+they can be removed by other active maintainers. If an emeritus maintainer wants
+to regain an active role, they can do so after renewing their contributions.
```
spzala, line 63: A small suggestion - "If an emeritus maintainer wants
to regain an active role, they can do so by renewing their contributions. Such a move should be welcomed by active maintainers." ?

GOVERNANCE.md
spzala: Also, I think we should add something like this after the above sentence - "Before such a change, an inactive maintainer should be notified through email or slack by active maintainers considering GitHub notifications sometimes go unnoticed."

serathius: Ad 1, added motivation to maintain the list to first section.

Ad 2, I think email is enough as all maintainers are required to publish their preferred email contact in ./MAINTAINERS.

spzala: Thanks @serathius and yes, I agree with both the points - email is good enough, and we should remove IRC reference.

spzala (approved):
lgtm @serathius thanks!

serathius:
Note: with @spzala approval we have started the 3 weeks lazy consensus timeout.

serathius:
We passed the lazy consensus time (3 business weeks). Feedback from @hexfusion and @jpbetz was also incorporated, allowing emeritus maintainers to imminently regain their status when they renew their contributions.

With 3 maintainers approval we can merge the change.

ptabor:
My small after-thought (influence also by recent issue), is that maybe we should keep RAFT maintainers as is. There is no activity, as project is considered stable/done, but when help was needed they act. In particular @tbg is advising(#14370), coding(#14413) reviewing (#14538),

Issue #14232, opened 2022-07-19
Plans for v3.4.20 release


ahrtr:
We intentionally skipped back-porting some bug fixes in 3.4.19 in order to minimize the impact. Since 3.4.19 is already released and validated to be working well, so we can consider to release 3.4.20 to backport some PRs. 

## Pipelines/test
There are still some flaky test failures from time to time. As previous release, any fix to stabilize the pipeline is welcome.

* https://github.com/etcd-io/etcd/issues/14216
  * [X] https://github.com/etcd-io/etcd/pull/14274
* https://github.com/etcd-io/etcd/issues/14256
  * [X] https://github.com/etcd-io/etcd/pull/14290
* https://github.com/etcd-io/etcd/issues/14257
* https://github.com/etcd-io/etcd/issues/14259
* https://github.com/etcd-io/etcd/issues/14270
  * [X] https://github.com/etcd-io/etcd/pull/14279

## Issues/PRs that should be included in 3.4.20
* [X] https://github.com/etcd-io/etcd/pull/12762
  * [X] https://github.com/etcd-io/etcd/pull/14258
* [X] https://github.com/etcd-io/etcd/pull/14269
* [X] https://github.com/etcd-io/etcd/pull/12846
  * [X] https://github.com/etcd-io/etcd/pull/14246
* [X] https://github.com/etcd-io/etcd/pull/13932
  * [X] https://github.com/etcd-io/etcd/pull/14177
* [X] https://github.com/etcd-io/etcd/pull/13917
  * [X] https://github.com/etcd-io/etcd/pull/14236
* [X] https://github.com/etcd-io/etcd/pull/13726
  * [X] https://github.com/etcd-io/etcd/pull/14239
* [X] https://github.com/etcd-io/etcd/pull/13736
  * [X] https://github.com/etcd-io/etcd/pull/14241
* [X] https://github.com/etcd-io/etcd/pull/13508
  * [X] https://github.com/etcd-io/etcd/pull/14253
* [X] https://github.com/etcd-io/etcd/pull/14230
* [X] https://github.com/etcd-io/etcd/pull/12335
  * [X] https://github.com/etcd-io/etcd/pull/14240
* [X] https://github.com/etcd-io/etcd/pull/14169
  * [X] https://github.com/etcd-io/etcd/pull/14251
* [X] https://github.com/etcd-io/etcd/pull/13435
  * [X] https://github.com/etcd-io/etcd/pull/14254

[...]

PR #12335, opened 2020-09-25
 etcdserver: add more detailed traces on linearized reading

PierreZ:
This PR is related to #11884 

## Why

To improve debuggability of `agreement among raft nodes before linearized reading`, we added some tracing inside `linearizableReadLoop`.

This will allow us to know the timing of `s.r.ReadIndex` vs `s.applyWait.Wait(rs.Index)`.

## Results

The patch has been deployed on clusters running v3.4.9, here's some traces:

```json
{
  "level": "info",
  "ts": "2020-09-22T12:54:01.021Z",
  "caller": "traceutil/trace.go:152",
  "msg": "trace[2138445431] linearizableReadLoop",
  "detail": "",
  "duration": "855.447896ms",
  "start": "2020-09-22T12:54:00.166Z",
  "end": "2020-09-22T12:54:01.021Z",
  "steps": [
    "trace[2138445431] read index received  (duration: 824.408µs)",
    "trace[2138445431] applied index is now lower than readState.Index  (duration: 854.622058ms)"
  ]
}
```

PR #14240, opened 2022-07-19
[3.4] etcdserver: add more detailed traces on linearized reading

chaochn47:
Cherry-pick #12335

ref. #14232

@ahrtr

### DIFF
--- a/etcdserver/v3_server.go
+++ b/etcdserver/v3_server.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"strconv"
 	"time"
 
 	"go.etcd.io/etcd/auth"
@@ -710,6 +711,9 @@ func (s *EtcdServer) linearizableReadLoop() {
 			return
 		}
 
+		// as a single loop is can unlock multiple reads, it is not very useful
+		// to propagate the trace from Txn or Range.
+		trace := traceutil.New("linearizableReadLoop", s.getLogger())
 		nextnr := newNotifier()
 
 		s.readMu.Lock()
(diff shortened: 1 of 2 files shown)


etcdserver/v3_server.go
```diff
@@ -782,16 +786,26 @@ func (s *EtcdServer) linearizableReadLoop() {
...
 			continue
 		}
+		trace.Step("read index received")
 
-		if ai := s.getAppliedIndex(); ai < rs.Index {
+		index := rs.Index
+		trace.AddField(traceutil.Field{Key: "readStateIndex", Value: index})
+
+		ai := s.getAppliedIndex()
+		trace.AddField(traceutil.Field{Key: "appliedIndex", Value: strconv.FormatUint(ai, 10)})
```
ahrtr, line 795: Just a minor comment, why do we need to use `strconv.FormatUint` here, but use the original `uint64` value (line 792) above?

I understand you just  backport the PR, so this comment should be added for the original PR.

ahrtr: I suggest to make them consistent, and update it in the main branch as well.

chaochn47: I agree. logging should also be consistent with metric `proposalsApplied.Set(float64(ep.appliedi))` and etcdctl endpoint status output.

ahrtr: This is just a backporting PR, so it might be better to merge it for now, and address the comment in separate PR(s).

Are you updating this PR? If not, then let me merge this PR firstly. If you are in progress of updating this PR, then feel free to continue to do it.

ahrtr (approved):
LGTM

Thank you @chaochn47

ahrtr:
Since it's just a minor change, and there is no any change on the functionality, so merging it right now.

PR #12846, opened 2021-04-09
etcdserver: remove temp files in snap dir when etcdserver starting

pyiyun:
When etcd exits abnormally, tmp files will remain in snap dir, so clean up tmp files in snap dir when etcdserver starting.

Fixes #12837

PR #14246, opened 2022-07-20
[3.4] etcdserver,pkg: remove temp files in snap dir when etcdserver starting

vivekpatani:
- Backporting: https://github.com/etcd-io/etcd/pull/12846
- Reference: https://github.com/etcd-io/etcd/issues/14232

### DIFF
--- a/pkg/fileutil/fileutil.go
+++ b/pkg/fileutil/fileutil.go
@@ -16,6 +16,7 @@ package fileutil
 
 import (
 	"fmt"
+	"go.uber.org/zap"
 	"io"
 	"io/ioutil"
 	"os"
@@ -124,6 +125,37 @@ func CheckDirPermission(dir string, perm os.FileMode) error {
 	if dirMode != perm {
 		err = fmt.Errorf("directory %q exist, but the permission is %q. The recommended permission is %q to prevent possible unprivileged access to the data.", dir, dirInfo.Mode(), os.FileMode(PrivateDirMode))
 		return err
+	}
+	return nil
+}
+
+// RemoveMatchFile deletes file if matchFunc is true on an existing dir
+// Returns error if the dir does not exist or remove file fail
+func RemoveMatchFile(lg *zap.Logger, dir string, matchFunc func(fileName string) bool) error {
+	if lg == nil {
+		lg = zap.NewNop()
+	}
+	if !Exist(dir) {
+		return fmt.Errorf("directory %s does not exist", dir)
+	}
+	fileNames, err := ReadDir(dir)
+	if err != nil {
+		return err
+	}
+	var removeFailedFiles []string
+	for _, fileName := range fileNames {
+		if matchFunc(fileName) {
+			file := filepath.Join(dir, fileName)
+			if err = os.Remove(file); err != nil {
+				removeFailedFiles = append(removeFailedFiles, fileName)
+				lg.Error("remove file failed",
+					zap.String("file", file),
+					zap.Error(err))
+			}
+		}
+	}
+	if len(removeFailedFiles) != 0 {
+		return fmt.Errorf("remove file(s) %v error", removeFailedFiles)
 	}
 	return nil
 }
--- a/pkg/fileutil/fileutil_test.go
+++ b/pkg/fileutil/fileutil_test.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"go.uber.org/zap/zaptest"
 )
 
 func TestIsDirWriteable(t *testing.T) {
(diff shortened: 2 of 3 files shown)


vivekpatani:
@ahrtr looks like I've missed a few things, I'll have it fixed in sometime.

pkg/fileutil/fileutil.go
```diff
@@ -127,3 +128,35 @@ func CheckDirPermission(dir string, perm os.FileMode) error {
...
+	var removeFailedFiles []string
+	for _, fileName := range fileNames {
+		if matchFunc(fileName) {
+			file := filepath.Join(dir, fileName)
+			if err = os.Remove(file); err != nil {
+				removeFailedFiles = append(removeFailedFiles, fileName)
+				lg.Error("remove file failed",
+					zap.String("file", file),
+					zap.Error(err))
+				continue
```
ahrtr, line 154: A very minor comment: actually the `continue` isn't needed. Of course, it will do any harm:)

pkg/fileutil/fileutil_test.go
```diff
@@ -166,3 +168,45 @@ func TestDirPermission(t *testing.T) {
 		t.Errorf("expected error, got nil")
 	}
 }
+
+func TestRemoveMatchFile(t *testing.T) {
+	tmpdir := t.TempDir()
+	defer os.RemoveAll(tmpdir)
```
ahrtr, line 174: No need to manually remove the `tmpdir`, because it is automatically removed by the golang testing utilities.
FYI. 
* https://github.com/golang/go/blob/go1.16.15/src/testing/testing.go#L911-L915 
* https://github.com/golang/go/blob/go1.16.15/src/testing/testing.go#L957
* https://github.com/etcd-io/etcd/pull/14248

pkg/fileutil/fileutil_test.go
```diff
@@ -166,3 +168,45 @@ func TestDirPermission(t *testing.T) {
...
+	f, err := ioutil.TempFile(tmpdir, "tmp")
```
ahrtr, line 175: `io/ioutil ` has already been deprecated in golang 1.16, please use `os.CreateTemp` instead.

FYI. https://go.dev/doc/go1.16#ioutil

pkg/fileutil/fileutil_test.go
```diff
@@ -166,3 +168,45 @@ func TestDirPermission(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	f.Close()
+	f, err = ioutil.TempFile(tmpdir, "foo.tmp")
```
ahrtr, line 180: same as above

pkg/fileutil/fileutil_test.go
```diff
@@ -166,3 +168,45 @@ func TestDirPermission(t *testing.T) {
...
+	}
+	fnames, err := ReadDir(tmpdir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(fnames) != 1 {
+		t.Errorf("expected exist 1 files, got %d", len(fnames))
+	}
+
+	f, err = ioutil.TempFile(tmpdir, "tmp")
```
ahrtr, line 200: same as above

serathius:
I would be worried about making improvements/patches to cherry-picked commits. When cherry-picking we should do minimal changes to original code so it works with older branch. 

We should not improve the code during cherry-pick. Back-porting should be focused on fixing bug and not alter the code from main branch.  If there are any places for improvements, please start from main branch.

ahrtr:
> I would be worried about making improvements/patches to cherry-picked commits. When cherry-picking we should do minimal changes to original code so it works with older branch.

Actually there are already some changes on the main branch, and they are exactly what I pointed out. The `defer os.RemoveAll(tmpdir)` isn't correct, we should remove it. Other comments isn't a big deal, personally I would suggest to update it as well, but I will not insist on it.

ahrtr:
Specifically, the PR [12846](https://github.com/etcd-io/etcd/pull/12846) is already out of date in the main branch.

pkg/fileutil/fileutil_test.go
vivekpatani: addressed.

pkg/fileutil/fileutil_test.go
vivekpatani: addressed.

pkg/fileutil/fileutil_test.go
vivekpatani: addressed.

pkg/fileutil/fileutil.go
vivekpatani: addressed.

ahrtr (approved):
LGTM

Thank you @vivekpatani

ahrtr:
IT's a minor and accordingly safe change, merging...

PR #14250, opened 2022-07-21
Resolve the channel+goroutine race condition

ahrtr:
Refer to https://go.dev/doc/effective_go#channels

### DIFF
--- a/raft/raft_test.go
+++ b/raft/raft_test.go
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"reflect"
 	"strings"
+	"sync"
 	"testing"
 
 	pb "go.etcd.io/etcd/raft/v3/raftpb"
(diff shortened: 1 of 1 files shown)


serathius:
Maybe I missed it, but I didn't see any goroutines used in modified code. Pattern you used `tt := tt` is usually used in cases where in for loop there is some parallel execution. 
Examples I know
```
for _, tt := range tests {
  tt := tt
  go func(){
  ...
  }()
}
```
and 
```
for _, tt := range tests {
  tt := tt
  t.Run(tt.name, func(t *testing.T){
    t.Parallel()
    ...
  }()
}
```
I don't see those patterns in test this PR changes. Maybe I missed something.

ahrtr:
Ah:) you are right. It seems I need some sleep to fresh my head:)

PR #13508, opened 2021-11-26
Lease Checkpoints fix

serathius:
This is continuation of work on https://github.com/etcd-io/etcd/pull/13491
Main difference is that this proposes to persist remainingTTL on checkpoint making checkpoints not depend on snapshot frequency.

I also check that renew logic should not be impacted by this change. Renew already includes logic that will reset remainingTTL if needed.

PR #14253, opened 2022-07-21
[3.4] Checkpoints fix 3.4

serathius:

Backport https://github.com/etcd-io/etcd/pull/13508
cc @ahrtr


server/embed/etcd.go
```diff
@@ -0,0 +1,860 @@
+// Copyright 2016 The etcd Authors
```
ahrtr, line 1: I do not understand why do you add this file? It seems it isn't needed in release-3.4.

ahrtr: You need to update `embed/etcd.go` instead.

serathius: Ups, looks like this was added as this file was not present in v3.4 branch.

PR #14273, opened 2022-07-25
tests: fix the logic of testNonleaderElectionTimeoutRandomized in raft_paper_test.go

RobertIndie:
In the test `testNonleaderElectionTimeoutRandomized`, the possible values of election timeout randomized out should be 10-19(inclusive), but the current test only tests the possible values as 11-19(inclusive), which may be incorrect here. We need to expand the scope of the tests to ensure robustness and make the logic clearer.

Signed-off-by: Zike Yang <zike@apache.org>


serathius (approved):
LGTM

PR #14276, opened 2022-07-26
Add alarms metrics for server

### LINKED ISSUE #14234: Expose alarms as metrics
Currently the only way to know about alarms is to issue an maitenance/alarm grpc call to list currently active alarms (by using etcdctl alarm list, for example).

It would be great to have expose etcd alarms in prometheus metrics so they're easier to monitor and know about.

yuzhiquan:

etcdserver: add alarms metrics for etcd server.

Fix #14234

### DIFF
--- a/server/etcdserver/apply/apply.go
+++ b/server/etcdserver/apply/apply.go
@@ -17,7 +17,6 @@ package apply
 import (
 	"context"
 
-	"github.com/coreos/go-semver/semver"
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/membershippb"
 	"go.etcd.io/etcd/client/pkg/v3/types"
@@ -34,9 +33,10 @@ import (
 	serverstorage "go.etcd.io/etcd/server/v3/storage"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	"go.etcd.io/etcd/server/v3/storage/mvcc"
-
-	"github.com/gogo/protobuf/proto"
 	"go.uber.org/zap"
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/gogo/protobuf/proto"
 )
 
 const (
@@ -231,12 +231,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.WithLabelValues(types.ID(ar.MemberID).String(), m.Alarm.String()).Inc()
 	case pb.AlarmRequest_DEACTIVATE:
 		m := a.alarmStore.Deactivate(types.ID(ar.MemberID), ar.Alarm)
 		if m == nil {
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.WithLabelValues(types.ID(ar.MemberID).String(), m.Alarm.String()).Dec()
 	default:
 		return nil, nil
 	}
--- a/server/etcdserver/apply/metrics.go
+++ b/server/etcdserver/apply/metrics.go
@@ -1,0 +1,31 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apply
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	alarms = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "server",
+		Name:      "alarms",
+		Help:      "Alarms for every member in cluster. 1 for 'server_id' label with current ID. 2 for 'alarm_type' label with type of this alarm",
+	},
+		[]string{"server_id", "alarm_type"})
+)
+
+func init() {
+	prometheus.MustRegister(alarms)
+}
--- a/CHANGELOG/CHANGELOG-3.6.md
+++ b/CHANGELOG/CHANGELOG-3.6.md
@@ -83,6 +83,7 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 See [List of metrics](https://etcd.io/docs/latest/metrics/) for all metrics per release.
 
 - Add [`etcd_disk_defrag_inflight`](https://github.com/etcd-io/etcd/pull/13371).
+- Add [`etcd_debugging_server_alarms`](https://github.com/etcd-io/etcd/pull/14276).
 
 ### Go
 - Compile with [Go 1.17+](https://golang.org/doc/devel/release.html#go1.17)


server/etcdserver/apply/apply.go
```diff
@@ -231,12 +233,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.With(prometheus.Labels{"server_id": strconv.FormatUint(ar.MemberID, 10)}).Inc()
```
ahrtr, line 236: hex-decimal makes more sense.

server/etcdserver/apply/apply.go
```diff
@@ -231,12 +233,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
...
 	case pb.AlarmRequest_DEACTIVATE:
 		m := a.alarmStore.Deactivate(types.ID(ar.MemberID), ar.Alarm)
 		if m == nil {
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.With(prometheus.Labels{"server_id": strconv.FormatUint(ar.MemberID, 10)}).Dec()
```
ahrtr, line 243: hex-decimal makes more sense.

server/etcdserver/apply/apply.go
```diff
@@ -233,14 +232,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
-		alarms.With(prometheus.Labels{"server_id": strconv.FormatUint(ar.MemberID, 10)}).Inc()
+		alarms.With(prometheus.Labels{"server_id": fmt.Sprintf("%016x", ar.MemberID)}).Inc()
```
ahrtr, line 235: Please cast it to [types.ID](https://github.com/etcd-io/etcd/blob/53bfdc11d8f6d6cdf0ccb986f052e9739086110b/client/pkg/types/id.go#L22-L24)

server/etcdserver/apply/apply.go
```diff
@@ -233,14 +232,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
...
 	case pb.AlarmRequest_DEACTIVATE:
 		m := a.alarmStore.Deactivate(types.ID(ar.MemberID), ar.Alarm)
 		if m == nil {
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
-		alarms.With(prometheus.Labels{"server_id": strconv.FormatUint(ar.MemberID, 10)}).Dec()
+		alarms.With(prometheus.Labels{"server_id": fmt.Sprintf("%016x", ar.MemberID)}).Dec()
```
ahrtr, line 242: same as above

server/etcdserver/apply/apply.go
```diff
@@ -231,12 +232,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.With(prometheus.Labels{"server_id": strconv.FormatUint(uint64(types.ID(ar.MemberID)), 10)}).Inc()
```
ahrtr, line 235:
```suggestion
		alarms.With(prometheus.Labels{"server_id":  types.ID(ar.MemberID).String()}).Inc()
```

server/etcdserver/apply/apply.go
```diff
@@ -231,12 +232,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
...
 	case pb.AlarmRequest_DEACTIVATE:
 		m := a.alarmStore.Deactivate(types.ID(ar.MemberID), ar.Alarm)
 		if m == nil {
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.With(prometheus.Labels{"server_id": strconv.FormatUint(uint64(types.ID(ar.MemberID)), 10)}).Dec()
```
ahrtr, line 242: same as above

server/etcdserver/apply/apply.go
```diff
@@ -34,8 +35,6 @@ import (
 	serverstorage "go.etcd.io/etcd/server/v3/storage"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	"go.etcd.io/etcd/server/v3/storage/mvcc"
-
-	"github.com/gogo/protobuf/proto"
```
ahrtr, line 38: minor comment, please organize the import entries into three groups something like as below,

```
import (
    "context"

    "go.etcd.io/X"
    "go.etcd.io/Y"

    "github.com/xxx"
    "go.uber.org/zap"
)
```

ahrtr (approved):
LGTM

Thank you @yuzhiquan

yuzhiquan:
> LGTM
> 
> Thank you @yuzhiquan

My pleasure

ahrtr:
Please also add a [changelog](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md) item.

server/etcdserver/apply/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	alarms = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "server",
+		Name:      "alarms",
+		Help:      "Alarms of every member in cluster.1 for 'server_id' label with current ID.",
+	},
+		[]string{"server_id"})
```
stek29, line 26: Please also add alarm type label, with a string name of alarm

server/etcdserver/apply/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apply
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	alarms = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
```
stek29, line 21: Is there any reason why these metrics are in debugging namespace?

stek29:
also, just a nitpick: shouldn't commit message begin with `etcdserver:` too?

server/etcdserver/apply/metrics.go
yuzhiquan: I don't known, if we add new metrics, shoud it address in debugging firstly?

stek29: I'm not sure on when it's named `debugging` and when it's not
but this metric would definitely be useful when running etcd in production, and I'm not sure if it's useful for debugging.

yuzhiquan: @serathius please point out should this metrics start with `etcd_debugging`?

spzala: Yes, that's fine and it should. Any new metrics are done that way. Eventually, if it's stable, it can be moved to `etcd`.

server/etcdserver/apply/metrics.go
```diff
@@ -0,0 +1,31 @@
...
+		Subsystem: "server",
+		Name:      "alarms",
+		Help:      "Alarms of every member in cluster.1 for 'server_id' label with current ID.2 for 'alarm_type' label with type of this alarm",
```
spzala, line 24: Suggest to add space before `1` and `2` for better readability.  Also, nit, may be we should reword the first sentence to `Alarms for every member in the cluster`.

spzala (changes requested):
lgtm but please see couple of minor in-line comments and answer to the question on namespace. Thanks @yuzhiquan

spzala (approved):
lgtm 
Thanks @yuzhiquan

server/etcdserver/apply/apply.go
```diff
@@ -231,12 +232,14 @@ func (a *applierV3backend) Alarm(ar *pb.AlarmRequest) (*pb.AlarmResponse, error)
 			break
 		}
 		resp.Alarms = append(resp.Alarms, m)
+		alarms.With(prometheus.Labels{"server_id": types.ID(ar.MemberID).String(), "alarm_type": m.Alarm.String()}).Inc()
```
ahrtr, line 235: This isn't an issue, but it seems `WithLabelValues` is better than `prometheus.Labels`, because we do not need to type the hard coded labels everywhere. 

```
alarms.WithLabelValues(types.ID(ar.MemberID).String(), m.Alarm.String()).Inc()
```

I searched the repo, and majorities are using `WithLabelValues`, and only 6 places are using `prometheus.Labels`. So I suggest to use `WithLabelValues` so as to be consistent.

PR #14278, opened 2022-07-26
tests: Migrate member list tests to common framework

clarkfw:
Context #13637 

Signed-off-by: Clark <fwyongxing@gmail.com>

### DIFF
--- a/tests/common/member_test.go
+++ b/tests/common/member_test.go
@@ -1,0 +1,53 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


serathius:
I would recommend to not do everything in one PR. If you can migrate one test independently it would be faster for us to iterate.

clarkfw:
> I would recommend to not do everything in one PR. If you can migrate one test independently it would be faster for us to iterate.

sounds good. 
I've made some changes: In this PR, I'll only migrate member list tests.

tests/common/member_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2016 The etcd Authors
```
serathius, line 1:
```suggestion
// Copyright 2022 The etcd Authors
```

serathius (approved):
This PR makes me super excited. With only couple of lines the test is much more readable as we reuse so much code.

For context this is the old test:
```
func ctlV3MemberList(cx ctlCtx) error {
	cmdArgs := append(cx.PrefixArgs(), "member", "list")
	lines := make([]string, cx.cfg.ClusterSize)
	for i := range lines {
		lines[i] = "started"
	}
	return e2e.SpawnWithExpects(cmdArgs, cx.envMap, lines...)
}
```

serathius:
DCO failed after adding the suggestion commit. Can you squash the commits in PR?

Issue #14270, opened 2022-07-25
[3.4] WARNING: DATA RACE between mvcc.(*store).setupMetricsReporter and mvcc.(*store).restore


ahrtr:
I see the following error in 3.4 pipeline, 
```
==================
WARNING: DATA RACE
Read at 0x00c003404a00 by goroutine 943:
  go.etcd.io/etcd/mvcc.(*store).setupMetricsReporter.func5()
      /home/runner/work/etcd/etcd/mvcc/kvstore.go:600 +0xa9
  go.etcd.io/etcd/mvcc.glob..func9()
      /home/runner/work/etcd/etcd/mvcc/metrics.go:277 +0x95
  github.com/prometheus/client_golang/prometheus.(*valueFunc).Write()
      /home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.0.0/prometheus/value.go:72 +0x52
  github.com/prometheus/client_golang/prometheus.processMetric()
      /home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.0.0/prometheus/registry.go:594 +0x131
  github.com/prometheus/client_golang/prometheus.(*Registry).Gather()
--
  net/http.HandlerFunc.ServeHTTP()
      /opt/hostedtoolcache/go/1.16.15/x64/src/net/http/server.go:2050 +0x51
  github.com/prometheus/client_golang/prometheus/promhttp.InstrumentHandlerInFlight.func1()
      /home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.0.0/prometheus/promhttp/instrument_server.go:40 +0xf6
  net/http.HandlerFunc.ServeHTTP()
      /opt/hostedtoolcache/go/1.16.15/x64/src/net/http/server.go:2050 +0x51
  github.com/prometheus/client_golang/prometheus/promhttp.InstrumentHandlerCounter.func1()
      /home/runner/go/pkg/mod/github.com/prometheus/client_golang@v1.0.0/prometheus/promhttp/instrument_server.go:100 +0xfd
  net/http.HandlerFunc.ServeHTTP()
      /opt/hostedtoolcache/go/1.16.15/x64/src/net/http/server.go:2050 +0x51
  net/http.(*ServeMux).ServeHTTP()
      /opt/hostedtoolcache/go/1.16.15/x64/src/net/http/server.go:2429 +0xaf
  go.etcd.io/etcd/etcdserver/api/v2http.requestLogger.func1()
[...]

PR #14279, opened 2022-07-26
[3.4] clientv3/mvcc: fixed DATA RACE

SimFG:
clientv3/mvcc: fixed DATA RACE between mvcc.(*store).setupMetricsReporter and mvcc.(*store).restore
issue: #14270 

By comparing the code on the main branch, it is found that the main branch does not have this problem. But there is a small problem on the main branch, the code is as follows:
https://github.com/etcd-io/etcd/blob/a3b410cac74d6b0082ad3e40ba384b8650a809e4/server/storage/mvcc/kvstore.go#L349-L364
L362: `s.compactMainRev`
I think this needs a read lock. However, I think this simple modification like this pull request is more appropriate.
```
{
		s.revMu.Lock()
		s.currentRev = <-revc

		// keys in the range [compacted revision -N, compaction] might all be deleted due to compaction.
		// the correct revision should be set to compaction revision in the case, not the largest revision
		// we have seen.
		if s.currentRev < s.compactMainRev {
			s.currentRev = s.compactMainRev
		}

		if scheduledCompact <= s.compactMainRev {
			scheduledCompact = 0
		}
		s.revMu.Unlock()
	}
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/soheilhy/cmux v0.1.4
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.1
-	github.com/stretchr/testify v1.3.0 // indirect
+	github.com/stretchr/testify v1.3.0
 	github.com/tmc/grpc-websocket-proxy v0.0.0-20200427203606-3cfed13b9966
 	github.com/urfave/cli v1.20.0
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2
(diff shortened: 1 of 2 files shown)


SimFG:
cc @ahrtr

go.mod
```diff
@@ -35,7 +35,7 @@ require (
 	github.com/soheilhy/cmux v0.1.4
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.1
-	github.com/stretchr/testify v1.3.0 // indirect
```
ahrtr, line 38: Did you manually make this change?

SimFG: No, it's GoLand editor.

ahrtr (approved):
LGTM

Thank you @SimFG

PR #14281, opened 2022-07-27
tests: Migrate member add tests to common framework

clarkfw:
Conctext #13637

### DIFF
--- a/tests/common/member_test.go
+++ b/tests/common/member_test.go
@@ -19,6 +19,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
 )
 
@@ -50,5 +53,102 @@ func TestMemberList(t *testing.T) {
 				}
 			})
 		})
+	}
+}
+
+func TestMemberAdd(t *testing.T) {
+	testRunner.BeforeTest(t)
+
+	learnerTcs := []struct {
+		name    string
+		learner bool
+	}{
+		{
+			name:    "NotLearner",
+			learner: false,
+		},
+		{
+			name:    "Learner",
+			learner: true,
+		},
+	}
+
+	quorumTcs := []struct {
+		name                string
+		strictReconfigCheck bool
+		waitForQuorum       bool
+		expectError         bool
+	}{
+		{
+			name:                "StrictReconfigCheck/WaitForQuorum",
+			strictReconfigCheck: true,
+			waitForQuorum:       true,
+			expectError:         false,
+		},
+		{
+			name:                "StrictReconfigCheck/NoWaitForQuorum",
+			strictReconfigCheck: true,
+			waitForQuorum:       false,
+			expectError:         true,
+		},
+		{
+			name:                "DisableStrictReconfigCheck/WaitForQuorum",
+			strictReconfigCheck: false,
+			waitForQuorum:       true,
+			expectError:         false,
+		},
+		{
+			name:                "DisableStrictReconfigCheck/NoWaitForQuorum",
+			strictReconfigCheck: false,
+			waitForQuorum:       false,
+			expectError:         false,
+		},
+	}
+
+	for _, learnerTc := range learnerTcs {
+		for _, quorumTc := range quorumTcs {
+			for _, clusterTc := range clusterTestCases {
+				t.Run(learnerTc.name+"/"+quorumTc.name+"/"+clusterTc.name, func(t *testing.T) {
+					ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+					defer cancel()
+					c := clusterTc.config
+					if !quorumTc.strictReconfigCheck {
+						c.DisableStrictReconfigCheck = true
+					}
+					clus := testRunner.NewCluster(ctx, t, c)
+					defer clus.Close()
+					cc := clus.Client()
+
+					testutils.ExecuteUntil(ctx, t, func() {
+						var addResp *clientv3.MemberAddResponse
+						var err error
+						if quorumTc.waitForQuorum {
+							time.Sleep(etcdserver.HealthInterval)
+						}
+						if learnerTc.learner {
+							addResp, err = cc.MemberAddAsLearner(ctx, "newmember", []string{"http://localhost:123"})
+						} else {
+							addResp, err = cc.MemberAdd(ctx, "newmember", []string{"http://localhost:123"})
+						}
+						if quorumTc.expectError && c.ClusterSize > 1 {
+							// calling MemberAdd/MemberAddAsLearner on a single node will not fail,
+							// whether strictReconfigCheck or whether waitForQuorum
+							require.ErrorContains(t, err, "etcdserver: unhealthy cluster")
+						} else {
+							require.NoError(t, err, "MemberAdd failed")
+							if addResp.Member == nil {
+								t.Fatalf("MemberAdd failed, expected: member != nil, got: member == nil")
+							}
+							if addResp.Member.ID == 0 {
+								t.Fatalf("MemberAdd failed, expected: ID != 0, got: ID == 0")
+							}
+							if len(addResp.Member.PeerURLs) == 0 {
+								t.Fatalf("MemberAdd failed, expected: non-empty PeerURLs, got: empty PeerURLs")
+							}
+						}
+					})
+				})
+			}
+		}
 	}
 }
--- a/tests/e2e/ctl_v3_member_test.go
+++ b/tests/e2e/ctl_v3_member_test.go
@@ -50,19 +50,7 @@ func TestCtlV3MemberRemoveClientAutoTLS(t *testing.T) {
 func TestCtlV3MemberRemovePeerTLS(t *testing.T) {
 	testCtl(t, memberRemoveTest, withQuorum(), withDisableStrictReconfig(), withCfg(*e2e.NewConfigPeerTLS()))
 }
-func TestCtlV3MemberAdd(t *testing.T)      { testCtl(t, memberAddTest) }
-func TestCtlV3MemberAddNoTLS(t *testing.T) { testCtl(t, memberAddTest, withCfg(*e2e.NewConfigNoTLS())) }
-func TestCtlV3MemberAddClientTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigClientTLS()))
-}
-func TestCtlV3MemberAddClientAutoTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigClientAutoTLS()))
-}
-func TestCtlV3MemberAddPeerTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigPeerTLS()))
-}
-func TestCtlV3MemberAddForLearner(t *testing.T) { testCtl(t, memberAddForLearnerTest) }
-func TestCtlV3MemberUpdate(t *testing.T)        { testCtl(t, memberUpdateTest) }
+func TestCtlV3MemberUpdate(t *testing.T) { testCtl(t, memberUpdateTest) }
 func TestCtlV3MemberUpdateNoTLS(t *testing.T) {
 	testCtl(t, memberUpdateTest, withCfg(*e2e.NewConfigNoTLS()))
 }
@@ -176,18 +164,6 @@ func memberRemoveTest(cx ctlCtx) {
 func ctlV3MemberRemove(cx ctlCtx, ep, memberID, clusterID string) error {
 	cmdArgs := append(cx.prefixArgs([]string{ep}), "member", "remove", memberID)
 	return e2e.SpawnWithExpectWithEnv(cmdArgs, cx.envMap, fmt.Sprintf("%s removed from cluster %s", memberID, clusterID))
-}
-
-func memberAddTest(cx ctlCtx) {
-	if err := ctlV3MemberAdd(cx, fmt.Sprintf("http://localhost:%d", e2e.EtcdProcessBasePort+11), false); err != nil {
-		cx.t.Fatal(err)
-	}
-}
-
-func memberAddForLearnerTest(cx ctlCtx) {
-	if err := ctlV3MemberAdd(cx, fmt.Sprintf("http://localhost:%d", e2e.EtcdProcessBasePort+11), true); err != nil {
-		cx.t.Fatal(err)
-	}
 }
 
 func ctlV3MemberAdd(cx ctlCtx, peerURL string, isLearner bool) error {
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -76,14 +76,6 @@ func NewConfigClientTLS() *EtcdProcessClusterConfig {
 	}
 }
 
-func NewConfigClientBoth() *EtcdProcessClusterConfig {
-	return &EtcdProcessClusterConfig{
-		ClusterSize:  1,
-		ClientTLS:    ClientTLSAndNonTLS,
-		InitialToken: "new",
-	}
-}
-
 func NewConfigClientAutoTLS() *EtcdProcessClusterConfig {
 	return &EtcdProcessClusterConfig{
 		ClusterSize:     1,
--- a/tests/framework/integration.go
+++ b/tests/framework/integration.go
@@ -327,3 +327,11 @@ func (c integrationClient) Watch(ctx context.Context, key string, opts config.Wa
 
 	return c.Client.Watch(ctx, key, opOpts...)
 }
+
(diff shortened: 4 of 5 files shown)


tests/e2e/ctl_v3_member_test.go
```diff
@@ -50,17 +50,7 @@ func TestCtlV3MemberRemoveClientAutoTLS(t *testing.T) {
...
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigClientTLS()))
-}
-func TestCtlV3MemberAddClientAutoTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigClientAutoTLS()))
-}
-func TestCtlV3MemberAddPeerTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigPeerTLS()))
-}
+func TestCtlV3MemberAdd(t *testing.T)           { testCtl(t, memberAddTest) }
 func TestCtlV3MemberAddForLearner(t *testing.T) { testCtl(t, memberAddForLearnerTest) }
```
serathius, line 54: Isn't this also covered?

serathius:
Found errror:
```
member_test.go:94: could not add member, err: match not found. Set EXPECT_DEBUG for more info Err: read /dev/ptmx: input/output error, last lines:
```

tests/e2e/ctl_v3_member_test.go
clarkfw: The underlying implementation of `TestCtlV3MemberAdd` and `TestCtlV3MemberAddForLearner` use `member add` and `member add --learner` command.

I thought it would be valuable to keep them in the e2e testing, to cover the real command usage scenarios.

Do we need to keep them?

serathius: Let me know if I understood you. New framework runs e2e test, but runs etcdctl only with `--format=json`, so we never test `etcdctl member add`, but `etcdctl member add --format=json`?.

clarkfw: Oh, I've got the point. 

I thought our common framework doesn't run e2e test. Now I know it does.

I'm going to remove these redundant functions.

serathius: > I thought our common framework doesn't run e2e test. Now I know it does.

Common framework is an interface for testing that covers both e2e and integration. Tests in `tests/common` that use common framework will be automatically run as both e2e and integration tests. Write a test once and double the benefit. 

Instead of having two separate frameworks and redundant tests, we want to just have one framework with shared tests. For now we are migrating `tests/e2e` directory, however if you want you can look into `tests/integration` directory and check if some test scenarios are already covered and can be removed.

It's a good point that we don't test exact output of `etcdctl member add`, however that's much lower on our priority than testing e2e process.

clarkfw:
This clue shows that a member of a cluster starts serving client before it's connected to all peers.
```
/home/runner/work/etcd/etcd/bin/etcd-22812: {"level":"warn","ts":"2022-07-27T08:46:33.653Z","caller":"etcdserver/server.go:1351","msg":"rejecting member add request; local member has not been connected to all peers, reconfigure breaks active quorum","local-member-id":"a8f3458387986d","requested-member-add":"{ID:557cfa84711678b8 RaftAttributes:{PeerURLs:[http://240.0.0.0:65535/] IsLearner:false} Attributes:{Name: ClientURLs:[]}}","error":"etcdserver: unhealthy cluster"}
[14847](https://github.com/etcd-io/etcd/runs/7536399620?check_suite_focus=true#step:5:14848)
```

To add or to remove a member from a cluster, a client need to wait a little bit. I've reproduced it on my computer.

Xnip2022-07-27_19-40-35

To correct the test cases, I'm gonna keep trying `MemberAdd()` for a few seconds.

serathius:
Hmm, maybe we should use `clus.WaitLeader`?

serathius:
One potential problem is adding members to 1 node cluster. In that situation we grow from 1 -> 2 member cluster. This means that quorum is now 2 member, however in tests we really don't add working member (just a backhole url). This means that 1 node cluster will become unavailable.

On the other hand 3 node cluster grows to 4 members, so the quorum stays the same (2 members). Please try removing test cases that use cluster of size 1.

clarkfw:
> Hmm, maybe we should use `clus.WaitLeader()`?

Wow, this is an amazing function, it looks like what I need. I'll give it try.

clarkfw:
> One potential problem is adding members to 1 node cluster. In that situation we grow from 1 -> 2 member cluster. This means that quorum is now 2 member, however in tests we really don't add working member (just a backhole url). This means that 1 node cluster will become unavailable.
> 
> On the other hand 3 node cluster grows to 4 members, so the quorum stays the same (2 members). Please try removing test cases that use cluster of size 1.

sg.

serathius:
>>Hmm, maybe we should use clus.WaitLeader()?
>
>Wow, this is an amazing function, it looks like what I need. I'll give it try.

I think cluster create should already call it so I would look into second suggestion with skipping test cases where cluster size is equal 1.

clarkfw:
As I went through the code base, I found that `clus.WaitLeader()` is called in many integration test cases, not in any e2e test case or the process of cluster creation. 
Neither in old e2e test code nor in common framework e2e code, an equivalent function to `clus.WaitLeader()` was found. 😟

clarkfw:
> > > Hmm, maybe we should use clus.WaitLeader()?
> > 
> > 
> > Wow, this is an amazing function, it looks like what I need. I'll give it try.
> 
> I think cluster create should already call it so I would look into second suggestion with skipping test cases where cluster size is equal 1.

Skipping test cases where cluster size is equal 1 is not likely going to make the tests work. The failed test cases at the moment are the ones with cluster size > 1.

When cluster size=1, the single node accepts `MemberAdd` request immediately and responses OK, ignoring the fact that  the newly added backhole peer url would make the cluster unavailable.

serathius:
I would recommend to rebase as there was a change that greatly reduces test flakiness https://github.com/etcd-io/etcd/pull/14283

clarkfw:
Maybe we need to implement a `clus.WaitLeader()` function in e2e cluster, and expose it in Cluster interface. What do you think?
```go
type Cluster interface {
	Members() []Member
        // HERE
	WaitLeader() int
	Client() Client
	Close() error
}

type Member interface {
	Client() Client
	Start() error
	Stop()
}

type Client interface {
	Put(key, value string, opts config.PutOptions) error
	Get(key string, opts config.GetOptions) (*clientv3.GetResponse, error)
	Delete(key string, opts config.DeleteOptions) (*clientv3.DeleteResponse, error)
	Compact(rev int64, opts config.CompactOption) (*clientv3.CompactResponse, error)
	Status() ([]*clientv3.StatusResponse, error)
	HashKV(rev int64) ([]*clientv3.HashKVResponse, error)
	Health() error
	Defragment(opts config.DefragOption) error
	AlarmList() (*clientv3.AlarmResponse, error)
	AlarmDisarm(alarmMember *clientv3.AlarmMember) (*clientv3.AlarmResponse, error)

```

ahrtr:
Do you need to wait for https://github.com/etcd-io/etcd/pull/14304 to be merged and then update this PR?

clarkfw:
> Do you need to wait for #14304 to be merged and then update this PR?

Yes. This PR is blocked by #14304. Should we label this PR or close it at the moment?

ahrtr:
> > Do you need to wait for #14304 to be merged and then update this PR?
> 
> Yes. This PR is blocked by #14304. Should we label this PR or close it at the moment?

Once https://github.com/etcd-io/etcd/pull/14304 is merged, you can rebase this PR.

tests/common/member_test.go
```diff
@@ -51,3 +57,56 @@ func TestMemberList(t *testing.T) {
...
+				clus := testRunner.NewCluster(ctx, t, nc.config)
+				defer clus.Close()
+				cc := clus.Client()
+
+				testutils.ExecuteUntil(ctx, t, func() {
+					clus.WaitLeader(t)
+					var addResp *clientv3.MemberAddResponse
+					var err error
+					if tc.learner {
+						addResp, err = cc.MemberAddAsLearner("newmember", []string{blackHolePeerUrl})
```
ahrtr, line 92: Why not to use a real peerURL something like `http://localhost:xxx `instead of `http://240.0.0.0:65535`?

clarkfw: `http://240.0.0.0:65535` is used to avoid "Peer URLs already exists" error. If `http://localhost:xxx` happens to be the peer-url of a started etcd instance, `member add` operation would fail due to that error. I considered to use `http://localhost:xxx` with a unused random port such as 2777 and 15981, but I'm slightly worried about port conflict in the future.

clarkfw:
The `E2E / test (linux-386-e2e) (pull_request)` check failed due to an `unhealthy cluster` error
```
/home/runner/work/etcd/etcd/bin/etcd-22730: {"level":"warn","ts":"2022-08-14T11:45:32.476Z","caller":"etcdserver/server.go:1351","msg":"rejecting member add request; local member has not been connected to all peers, reconfigure breaks active quorum","local-member-id":"d70245c72180528a","requested-member-add":"{ID:ccf9af529e52be7e RaftAttributes:{PeerURLs:[http://240.0.0.0:65535/] IsLearner:false} Attributes:{Name: ClientURLs:[]}}","error":"etcdserver: unhealthy cluster"}
[14842](https://github.com/etcd-io/etcd/runs/7826140582?check_suite_focus=true#step:5:14843)
```


`MemberAdd` function works fine in integration test, but failed in e2e test. That's because integration cluster has `strict-reconfig-check` disabled by default and e2e cluster has `strict-reconfig-check` enabled by default. 

Therefore, in e2e cluster, `MemberAdd` function has to wait 5 seconds before e2e cluster `isConnectedFullySince`, or an `unhealthy cluster` error would occur.

https://github.com/etcd-io/etcd/blob/d05c894e854e1b8545c41ff26ac85c8b76917a7c/server/etcdserver/server.go#L1350-L1358

A PR will be opened to discuss this.

ahrtr:
Please resolve the conflict.

tests/common/member_test.go
```diff
@@ -52,3 +55,157 @@ func TestMemberList(t *testing.T) {
...
+						addResp, err = cc.MemberAdd(ctx, "newmember", []string{"http://localhost:123"})
+					}
+					checkAddResp(t, addResp, err)
+				})
+			})
+		}
+	}
+}
+
+func checkAddResp(t *testing.T, addResp *clientv3.MemberAddResponse, err error) {
```
serathius, line 198: This checks error and validates response. Could we maybe split this:
* rename `checkAddResp` to `validateMemberAdd` and remove `err` argument.
* for error checking we can use `require.NoError(err)`.

tests/common/member_test.go
```diff
@@ -52,3 +55,157 @@ func TestMemberList(t *testing.T) {
...
+		for _, nc := range clusterTestCases {
+			t.Run(tc.name+"/"+nc.name, func(t *testing.T) {
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer cancel()
+				clus := testRunner.NewCluster(ctx, t, nc.config)
+				defer clus.Close()
+				cc := clus.Client()
+
+				testutils.ExecuteUntil(ctx, t, func() {
+					clus.WaitLeader(t)
```
serathius, line 86: Do we need `WaitLeader` here? I haven't seen it used in other tests as I would expect that `NewCluster` returns me cluster after leader is established.

tests/common/member_test.go
```diff
@@ -52,3 +55,157 @@ func TestMemberList(t *testing.T) {
...
+					var addResp *clientv3.MemberAddResponse
+					var err error
+					if tc.learner {
+						addResp, err = cc.MemberAddAsLearner(ctx, "newmember", []string{"http://localhost:123"})
+					} else {
+						addResp, err = cc.MemberAdd(ctx, "newmember", []string{"http://localhost:123"})
+					}
+					if nc.config.ClusterSize == 1 {
+						checkAddResp(t, addResp, err)
```
serathius, line 95: I would expect reverse result, adding invalid (`localhost:123` will never respond) member should fail in single node cluster as you cannot establish quorum. From 2 members only 1 is alive, but quorum is 2.

tests/common/member_test.go
```diff
@@ -52,3 +55,157 @@ func TestMemberList(t *testing.T) {
...
+							t.Fatalf("expected: 'etcdserver: unhealthy cluster' error,  got: %s", err)
+						}
+					}
+				})
+			})
+		}
+	}
+}
+
+func TestMemberAdd_SleetHealthInterval(t *testing.T) {
```
serathius, line 110:
```suggestion
func TestMemberAdd_SleepHealthInterval(t *testing.T) {
```

serathius, line 110: I don't see similar test removed in this PR. Where this test has come from?

tests/common/member_test.go
clarkfw: Thanks to the suggestions! I'd love to do so , as they make the code clearer.

tests/common/member_test.go
```diff
@@ -52,3 +55,157 @@ func TestMemberList(t *testing.T) {
...
+				defer cancel()
+				clus := testRunner.NewCluster(ctx, t, nc.config)
+				defer clus.Close()
+				cc := clus.Client()
+
+				testutils.ExecuteUntil(ctx, t, func() {
+					clus.WaitLeader(t)
+					var addResp *clientv3.MemberAddResponse
+					var err error
+					time.Sleep(etcdserver.HealthInterval)
```
serathius, line 140: Why adding sleep here makes a difference?

tests/common/member_test.go
clarkfw: No. It turns out we don't need `WaitLeader` here. I'll remove them.

The first time I saw `etcdserver: unhealthy cluster` error after calling `MemberAdd`, I thought it was because all etcd members did not agree on the same leader. Therefore, I added `WaitLeader` function to the `Cluster` interface. But the `etcdserver: unhealthy cluster` error did not disappear. Finally I found it was caused by `--strict-reconfig-check`.

tests/common/member_test.go
clarkfw: This is a scenario not covered before: 

- In the first 5 seconds(HealthInterval) after the creation of a cluster(with `--strict-reconfig-check` enabled), any call to `MemberAdd` and `MemberRemove` should return an `etcdserver: unhealthy cluster` error. 
- As the first 5 seconds(HealthInterval) has passed, call to these functions should be successful.

This behaviour is controlled by the `isConnectedFullySince` function:
https://github.com/etcd-io/etcd/blob/d05c894e854e1b8545c41ff26ac85c8b76917a7c/server/etcdserver/server.go#L1350-L1358

serathius: ok, please rename/rewrite test so the reason for the behavior is known. For example `TestMemberAdd_BeforeConnectedToAllPeers`.

serathius:
Overall note, instead of having 3 different test could merge them into one table test. For me the scenario in all those tests is almost the same:
* Create cluster
* Call MemberAdd
* Check whether it was successful or not

We could just add `expectError`, `strictReconfigCheck` and `waitForQuorum` as fields on test case struct.

tests/common/member_test.go
clarkfw: Actually, adding an invalid member succeeds in the original e2e test. I also tried this in my terminal, it succeeded too. Is this a long-standing bug?🤣

clarkfw:
> Overall note, instead of having 3 different test could merge them into one table test. For me the scenario in all those tests is almost the same:
> 
> * Create cluster
> * Call MemberAdd
> * Check whether it was successful or not
> 
> We could just add `expectError`, `strictReconfigCheck` and `waitForQuorum` as fields on test case struct.

I was worried about making a single test function too large and complex. I'll try and see if I can handle this.

clarkfw:
> Overall note, instead of having 3 different test could merge them into one table test. For me the scenario in all those tests is almost the same:
> 
> * Create cluster
> * Call MemberAdd
> * Check whether it was successful or not
> 
> We could just add `expectError`, `strictReconfigCheck` and `waitForQuorum` as fields on test case struct.

I followed your suggestion and it works great! I've force pushed. Please take a look.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -50,19 +50,7 @@ func TestCtlV3MemberRemoveClientAutoTLS(t *testing.T) {
...
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigClientTLS()))
-}
-func TestCtlV3MemberAddClientAutoTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigClientAutoTLS()))
-}
-func TestCtlV3MemberAddPeerTLS(t *testing.T) {
-	testCtl(t, memberAddTest, withCfg(*e2e.NewConfigPeerTLS()))
-}
-func TestCtlV3MemberAddForLearner(t *testing.T) { testCtl(t, memberAddForLearnerTest) }
-func TestCtlV3MemberUpdate(t *testing.T)        { testCtl(t, memberUpdateTest) }
```
serathius, line 65: Should we remove MemberUpdate test?

serathius (approved):

Issue #14259, opened 2022-07-22
[3.4] panic: runtime error: index out of range [0] with length 0 [recovered]


ahrtr:
The issue was found in 3.4 pipeline.

```
--- FAIL: TestWatchReconnRequest (1.12s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
goroutine 68799 [running]:
testing.tRunner.func1.2(0xef8e00, 0xc000a222a0)
	/opt/hostedtoolcache/go/1.16.15/x64/src/testing/testing.go:1153 +0x332
testing.tRunner.func1(0xc0005dc780)
	/opt/hostedtoolcache/go/1.16.15/x64/src/testing/testing.go:1156 +0x4b6
panic(0xef8e00, 0xc000a222a0)
	/opt/hostedtoolcache/go/1.16.15/x64/src/runtime/panic.go:965 +0x1b9
go.etcd.io/etcd/clientv3/integration.putAndWatch(0xc0005dc780, 0xc0002864c0, 0x10a3890, 0x1, 0x10a3890, 0x1)
	/home/runner/work/etcd/etcd/clientv3/integration/watch_test.go:341 +0x40c
go.etcd.io/etcd/clientv3/integration.testWatchReconnRequest(0xc0005dc780, 0xc0002864c0)
	/home/runner/work/etcd/etcd/clientv3/integration/watch_test.go:220 +0x290
go.etcd.io/etcd/clientv3/integration.runWatchTest(0xc0005dc780, 0xfc87b0)
	/home/runner/work/etcd/etcd/clientv3/integration/watch_test.go:64 +0x24f
go.etcd.io/etcd/clientv3/integration.TestWatchReconnRequest(0xc0005dc780)
	/home/runner/work/etcd/etcd/clientv3/integration/watch_test.go:181 +0x37
testing.tRunner(0xc0005dc780, 0xfc86d8)
	/opt/hostedtoolcache/go/1.16.15/x64/src/testing/testing.go:1203 +0xe5
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.16.15/x64/src/testing/testing.go:1248 +0x2b3
FAIL	go.etcd.io/etcd/clientv3/integration	260.996s
FAIL
Error: Process completed with exit code 1.
```

Refer to https://github.com/etcd-io/etcd/runs/7462977558?check_suite_focus=true

zhangzerui20:
[...]

PR #14285, opened 2022-07-29
Checking cluster member count in watch_test

tjungblu:
The test should fail early on an insufficient cluster instead of
panicking during the test.

Fixes issue #14259.

### DIFF
--- a/clientv3/integration/watch_test.go
+++ b/clientv3/integration/watch_test.go
@@ -50,6 +50,10 @@ func runWatchTest(t *testing.T, f watcherTest) {
 	clus := integration.NewClusterV3(t, &integration.ClusterConfig{Size: 3})
 	defer clus.Terminate(t)
 
+	if len(clus.Members) != 3 {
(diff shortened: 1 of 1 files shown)


clientv3/integration/watch_test.go
```diff
@@ -50,6 +50,10 @@ func runWatchTest(t *testing.T, f watcherTest) {
 	clus := integration.NewClusterV3(t, &integration.ClusterConfig{Size: 3})
 	defer clus.Terminate(t)
 
+	if len(clus.Members) != 3 {
```
lavacat, line 53: I've tried stepping through this test and there is check [here](https://github.com/etcd-io/etcd/blob/9bb134233a42c0339cd7c5355dd9c2772e580944/tests/framework/integration/cluster.go#L232) 
that waits for all members to be up. I think it checks for size implicitly.

Do we really need this extra check?

ahrtr: Right, I don't think we should add the additional check either.

PR #14286, opened 2022-07-31
Fixed goroutine leak in server/etcdserver/raft_test.go

VladSaioc:
package: etcdserver

Potential for goroutine leak in [TestStopRaftWhenWaitingForApplyDone](https://github.com/etcd-io/etcd/blob/f90f93e610bb8adb95d0803926fd92c0d301105f/server/etcdserver/raft_test.go#L159). [Timeout](https://github.com/etcd-io/etcd/blob/f90f93e610bb8adb95d0803926fd92c0d301105f/server/etcdserver/raft_test.go#L173) case will lead raft node [goroutine](https://github.com/etcd-io/etcd/blob/f90f93e610bb8adb95d0803926fd92c0d301105f/server/etcdserver/raft.go#L161) to block at a [select](https://github.com/etcd-io/etcd/blob/f90f93e610bb8adb95d0803926fd92c0d301105f/server/etcdserver/raft.go#L212) statement. Stopping raft node before calling `t.Fatalf` will unblock the select and terminate the child goroutine.

### DIFF
--- a/server/etcdserver/raft_test.go
+++ b/server/etcdserver/raft_test.go
@@ -168,18 +168,24 @@ func TestStopRaftWhenWaitingForApplyDone(t *testing.T) {
 	srv := &EtcdServer{lgMu: new(sync.RWMutex), lg: zaptest.NewLogger(t), r: *r}
 	srv.r.start(nil)
 	n.readyc <- raft.Ready{}
+
+	stop := func() {
+		srv.r.stopped <- struct{}{}
+		select {
+		case <-srv.r.done:
+		case <-time.After(time.Second):
+			t.Fatalf("failed to stop raft loop")
+		}
+	}
+
 	select {
 	case <-srv.r.applyc:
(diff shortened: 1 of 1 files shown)


server/etcdserver/raft_test.go
```diff
@@ -171,6 +171,7 @@ func TestStopRaftWhenWaitingForApplyDone(t *testing.T) {
 	select {
 	case <-srv.r.applyc:
 	case <-time.After(time.Second):
+		srv.r.stop()
```
ahrtr, line 174: The `srv.r.stop()` may be blocked. I would suggest to follow the same way as line 178 ~ 183, which can be wrapped in a function.

VladSaioc: Done.

ahrtr (approved):
LGTM

Thank you @VladSaioc

Issue #13167, opened 2021-06-30
Deflake etcd tests


serathius:
If we look into tests results since we migrated Github Actions [commits on main branch](https://github.com/etcd-io/etcd/commits/main?branch=main) we get:
* 7 out of 32 failures on [1st page](https://github.com/etcd-io/etcd/commits/main?before=2fb919bef908f4cdf9a841a7d9494614566335e0+35&branch=main)
* 14 out of 31 failures on [2nd page](https://github.com/etcd-io/etcd/commits/main?before=2fb919bef908f4cdf9a841a7d9494614566335e0+70&branch=main)
* 14 out of 33 failures on [3rd page](https://github.com/etcd-io/etcd/commits/main?before=2fb919bef908f4cdf9a841a7d9494614566335e0+105&branch=main)
* 15 out of 29 failures on [4th page](https://github.com/etcd-io/etcd/commits/main?before=2fb919bef908f4cdf9a841a7d9494614566335e0+140&branch=main)
* 13 out of 25 failures on [5th page](https://github.com/etcd-io/etcd/commits/main?before=2fb919bef908f4cdf9a841a7d9494614566335e0+175&branch=main)
* 13 out of 22 failures on [6th page](https://github.com/etcd-io/etcd/commits/main?before=2fb919bef908f4cdf9a841a7d9494614566335e0+210&branch=main)
* 19 out of 33 failures on [7th page](https://github.com/etcd-io/etcd/commits/main?after=2fb919bef908f4cdf9a841a7d9494614566335e0+209&branch=main)

Where failure/success is based on green check vs red cross under commit message (commits without them means that they were not tested as they were multiple commits in one PR).

Those are all test failures on main branch, so after a PR passed tests and was approved. We can use those failures to calculate chance of any PR failing to pass tests just due to test flaking. 
```
(7 + 14 + 14+ 14 + 15 + 13 + 13 + 19) / (32 + 31 + 33 + 29 + 25 + 22 + 33) = 53%
```

Having flakyness ratio of over 50% means that average PR needs to be run 2 times, but number of failures in sequences may be much much longer, 3-5 failures in row is not something uncommon. This can be frustrating especially to new contributors,  as there is no easy way to retrigger tests (need to do an empty commit amend and push). 

## Proposal 

Etcd community should set on a test flakyness target, measure it and establish a process to fix flaky tests. 

[...]

PR #14292, opened 2022-08-01
Add test name to e2e cluster members

tjungblu:
This should aid in debugging test flakes, especially in tests where the process is restarted very often and thus changes its pid.
Now it's a lot easier to grep for different members, also when different tests fail at the same time.
The test `TestDowngradeUpgradeClusterOf3` as mentioned in #13167 is a good example for that.

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -19,13 +19,15 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"strings"
 	"testing"
 	"time"
 
-	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
+
+	"go.etcd.io/etcd/server/v3/etcdserver"
 )
 
 const EtcdProcessBasePort = 20000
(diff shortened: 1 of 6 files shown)


ahrtr (approved):
LGTM

Thank you @tjungblu 

cc @spzala  @ptabor  to double confirm

tests/framework/e2e/cluster.go
```diff
@@ -19,13 +19,15 @@ import (
...
+	"regexp"
 	"strings"
 	"testing"
 	"time"
 
-	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
+
+	"go.etcd.io/etcd/server/v3/etcdserver"
```
spzala, line 30: @tjungblu thanks, changes lgtm but I am curious why are you changing import order here and in other files? The current order is as per overall convention used in the project (i.e. third party packages follow internal ones..) and we should keep it that way unless I am missing anything.

spzala (commented):
lgtm but one inline comment @tjungblu thanks!

tests/framework/e2e/cluster.go
spzala: @tjungblu can you please look at my small comment above?. Your PR looks good otherwise and we can merge it.

spzala: I won't hold the PR due to the ordering changes in the import which seems unnecessary but can be reordered separately.

ahrtr: Usually we shouldn't change the existing code (including this case, the order of the import entries and the blank line) without explicit reason. 

But on the other hand, it seems that it isn't a big deal on the change in this case.  Either keeping it as it's for now or reverting the change works for me. 

I suggest to keep it as it's for now. 

Then this PR is good to go, and defer to @spzala to make the final decision and merge the PR.

spzala: Yup, sgtm. Thanks @ahrtr

spzala (approved):

Issue #14252, opened 2022-07-21
failed to open WAL


jasondobryk:
Hi, there is a problem on several etc clusters on windows server at the customer. process есв.ехе managed by the windows service. It looks like this: the cluster is working, but if you restart the etc service on one of the servers, we will get an error:

`{"level":"fatal","ts":"2022-07-21T13:17:36.221+0300","caller":"etcdserver/storage.go:95","msg":"failed to open WAL","error":"wal: file not found","stacktrace":"go.etcd.io/etcd/server/v3/etcdserver.readWAL\n\t/go/src/go.etcd.io/etcd/release/etcd/server/etcdserver/storage.go:95\ngo.etcd.io/etcd/server/v3/etcdserver.restartNode\n\t/go/src/go.etcd.io/etcd/release/etcd/server/etcdserver/raft.go:481\ngo.etcd.io/etcd/server/v3/etcdserver.NewServer\n\t/go/src/go.etcd.io/etcd/release/etcd/server/etcdserver/server.go:533\ngo.etcd.io/etcd/server/v3/embed.StartEtcd\n\t/go/src/go.etcd.io/etcd/release/etcd/server/embed/etcd.go:245\ngo.etcd.io/etcd/server/v3/etcdmain.startEtcd\n\t/go/src/go.etcd.io/etcd/release/etcd/server/etcdmain/etcd.go:228\ngo.etcd.io/etcd/server/v3/etcdmain.startEtcdOrProxyV2\n\t/go/src/go.etcd.io/etcd/release/etcd/server/etcdmain/etcd.go:123\ngo.etcd.io/etcd/server/v3/etcdmain.Main\n\t/go/src/go.etcd.io/etcd/release/etcd/server/etcdmain/main.go:40\nmain.main\n\t/go/src/go.etcd.io/etcd/release/etcd/server/main.go:32\nruntime.main\n\t/go/gos/go1.16.15/src/runtime/proc.go:225"}`

What did you expect to happen?
I expect a normal restart of the etcd process

How can we reproduce it?
so far we have not been able to repeat the error on our test servers

Etcd version
etcd.exe --version
etcd Version: 3.5.4

etcdctl.exe version
etcdctl version: 3.5.4

Etcd configuration of one of the servers:
name: ck11-apache2.test.local
initial-advertise-peer-urls: http://10.X.X.178:2380
listen-peer-urls: http://10.X.X.178:2380
[...]

PR #14293, opened 2022-08-01
fileutil: fixed work with wal files lock on the windows

MIVerTFT:
The current wal file lock check contains a very strange code. 

```go
//Etcd/client/pkg/fileutil/lock_windows.go

        if err == nil {
		return nil
	} else if err.Error() == errLocked.Error() {
		return ErrLocked
	} else if err != windows.ERROR_LOCK_VIOLATION {
		return err
	}
	return nil
```

This code check whether the lock is set or not by the message of the error received from the operating system. At the same time, the message with which it is compared differs from the message that the operating system returns. 

```go
//The message returned by the OS
"The process cannot access the file because another process has locked a portion of the file."
//The message expected in the code
"the process cannot access the file because another process has locked a portion of the file"
```

As a result, the check fails and the purge loop deletes the blocked wal files. I rewrote the check to check the error code, in my opinion this is a more reliable way.

Linked issue [14252](https://github.com/etcd-io/etcd/issues/14252)

### DIFF
--- a/client/pkg/fileutil/lock_windows.go
+++ b/client/pkg/fileutil/lock_windows.go
@@ -18,15 +18,12 @@
 package fileutil
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"syscall"
 
 	"golang.org/x/sys/windows"
 )
-
-var errLocked = errors.New("the process cannot access the file because another process has locked a portion of the file")
 
 func TryLockFile(path string, flag int, perm os.FileMode) (*LockedFile, error) {
 	f, err := open(path, flag, perm)
@@ -89,10 +86,8 @@ func lockFile(fd windows.Handle, flags uint32) error {
 	err := windows.LockFileEx(fd, flags|windows.LOCKFILE_EXCLUSIVE_LOCK, 0, 1, 0, &windows.Overlapped{})
 	if err == nil {
 		return nil
-	} else if err.Error() == errLocked.Error() {
+	} else if err == windows.ERROR_LOCK_VIOLATION {
 		return ErrLocked
-	} else if err != windows.ERROR_LOCK_VIOLATION {
-		return err
 	}
-	return nil
+	return err
 }


client/pkg/fileutil/lock_windows.go
```diff
@@ -89,10 +86,8 @@ func lockFile(fd windows.Handle, flags uint32) error {
 	err := windows.LockFileEx(fd, flags|windows.LOCKFILE_EXCLUSIVE_LOCK, 0, 1, 0, &windows.Overlapped{})
 	if err == nil {
 		return nil
-	} else if err.Error() == errLocked.Error() {
+	} else if err == windows.ERROR_LOCK_VIOLATION {
```
spzala, line 89: @MIVerTFT thanks! So per some reading, I think`windows.ERROR_LOCK_VIOLATION` gives the error message, `The process cannot access the file because another process has locked a portion of the file.` - please correct me if I am wrong?  In that case, `errLocked` check seems redundant and your changes makes sense.

client/pkg/fileutil/lock_windows.go
```diff
@@ -89,10 +86,8 @@ func lockFile(fd windows.Handle, flags uint32) error {
...
 		return ErrLocked
-	} else if err != windows.ERROR_LOCK_VIOLATION {
-		return err
 	}
-	return nil
+	return err
```
spzala, line 92: I am not sure why `nil` was returned originally if the `err` was not windows.ERROR_LOCK_VIOLATION  (or `errLocked` which seems redundant per your changes above). But now, that if you return `err` that might break something in some case (I don't know but just guessing). Can you please double check or may be we leave `nil` as it is considering any other cases are expected to treat as `nil` (similar to the check above for `windows.InvalidHandle`wher it returns `nil`)? Thanks!

client/pkg/fileutil/lock_windows.go
MIVerTFT: Yes `windows.ERROR_LOCK_VIOLATION` gives the same error. But the comparison operator compares the values of windows error codes cause the `windows.ERROR_LOCK_VIOLATION` is a `syscall.Errno = 33`. I think it is more reliable than comparing strings.

client/pkg/fileutil/lock_windows.go
MIVerTFT: Yes, it also seemed strange to me, but this method could previously return `nil` only in two cases 
1. When it was possible to capture the lock and `err == nil`;
2. When condition `err.Error() != errLocked.Error() && err == windows.ERROR_LOCK_VIOLATION` is met. This condition looks very strange and now leads to uncorrected behavior;

In all other cases, `err` or `ErrLocked` is returned.

client/pkg/fileutil/lock_windows.go
spzala: Thanks @MIVerTFT Let me think little more, I will be adding review comments soon.

client/pkg/fileutil/lock_windows.go
spzala: Yes, the original condition (#`3` as you mentioned above) seems strange. Your changes lgtm, but let's wait for @ahrtr review as all. Also, curios, when you see the `wal: file not found` as described in the #14252 do you see it's hitting this `lock` related error? Thanks for your patience with the review! @MIVerTFT

ahrtr: It seems that it might be related to the https://github.com/etcd-io/etcd/issues/14252. When the [lockFile](https://github.com/etcd-io/etcd/blob/ae36a577d7becbdeebf1f0fb665573b721e435f8/client/pkg/fileutil/lock_windows.go#L97) intentionally ignore some errors, then the [purge goroutine](https://github.com/etcd-io/etcd/blob/ae36a577d7becbdeebf1f0fb665573b721e435f8/client/pkg/fileutil/purge.go#L64) may delete the file.

client/pkg/fileutil/lock_windows.go
ahrtr: I do not have a windows server right now, could you try to execute the following code? 

```
func checkError() {
	var err error
	err = windows.ERROR_LOCK_VIOLATION

	fmt.Printf("windows.ERROR_LOCK_VIOLATION: %s\n", err.Error())

	fmt.Printf("err.Error() == errLocked.Error(): %t\n", err.Error() == errLocked.Error())
}
```

MIVerTFT: I tried on windows 10 and on windows server 2019 got the same result
```
c:\1>test.exe
windows.ERROR_LOCK_VIOLATION: The process cannot access the file because another process has locked a portion of the file.
err.Error() == errLocked.Error(): true
```

client/pkg/fileutil/lock_windows.go
MIVerTFT: Yes, this is exactly this issue #14252, I wrote in the message above about it.

ahrtr: This PR indeed can fix https://github.com/etcd-io/etcd/issues/14252, but I am afraid it may also change the existing behavior. 

The current implementation intentionally ignore the error if it isn't `windows.ERROR_LOCK_VIOLATION`. I suspect some filesystem doesn't support `lock`, and it might be the reason why the original author added this restriction. 

Please also see my comment https://github.com/etcd-io/etcd/issues/14252#issuecomment-1208591535

MIVerTFT: Yes, I understand, but the ERROR_LOCK_VIOLATION error is a specific windows OS error with a specific code(0x21) and specific causes. It should not occur in any other scenarios.

Issue #12385, opened 2020-10-10
watch stream return "permission denied" if token expired


cfz:
i ran into the similar problem as #8914

i'm using v3.4.13 server and v3.4.12 client.
i can reproduce the error using following steps:

on one terminal , start etcd server and enable auth and shroten the default token ttl:
```
etcd --log-level warn --logger zap --auth-token-ttl 10 & 
etcdctl user add root:root
etcdctl user grant-role root root
etcdctl auth enable
```

on the other termianal, run a the demo program:
```
cfz@cfz-desktop:~/ws/etcd-watch-demo$ cat main.go
package main

import (
        "context"
        "fmt"
        "sync"
        "time"

        "go.etcd.io/etcd/clientv3"
)

var wg sync.WaitGroup

func main() {
        cfg := clientv3.Config{
                Endpoints: []string{"http://127.0.0.1:2379"},
                Username:  "root",
                Password:  "root",
                LogConfig: nil,
        }

        cli, err := clientv3.New(cfg)
        if err != nil {
                panic(err)
        }

        watch(cli, 1)
        time.Sleep(1 * time.Second)
        watch(cli, 2)
        time.Sleep(11 * time.Second)
        watch(cli, 3)
        time.Sleep(1 * time.Second)
        watch(cli, 4)

        wg.Wait()
}

func watch(cli *clientv3.Client, num int) {
        wg.Add(1)

        go func() {
                defer wg.Done()
                for wr := range cli.Watch(context.Background(), "") {
                        if wr.Err() != nil {
                                fmt.Printf("%#v\n", wr.Err())
                        }
                }

                fmt.Printf("#%v watch existed\n", num)
        }()
}
cfz@cfz-desktop:~/ws/etcd-watch-demo$ go run main.go
&errors.errorString{s:"rpc error: code = PermissionDenied desc = etcdserver: permission denied"}
#3 watch exited
&errors.errorString{s:"rpc error: code = PermissionDenied desc = etcdserver: permission denied"}
#4 watch exited
```

as shown in the log , the 1st and 2nd watch is ok, while the 3rd and 4th report "permission denied"

meanwile, etcd server will log following:
```
{"level":"warn","ts":"2020-10-10T20:34:45.010+0800","caller":"auth/store.go:1301","msg":"invalid auth token","token":"JWourXlCoFPTxMHj.8"}
[...]

PR #14296, opened 2022-08-02
Fixed unexpected cancelled watch with WatchID=0.

kafuu-chino:
etcdserver: Fixed unexpected cancelled watch with WatchID=0.

I ran into the same problem from https://github.com/etcd-io/etcd/issues/12385, I found that the error will cause the watch with `WatchID=0` to be invalid，so I set WatchID to start at 1 to fix it.

First of all `WatchID=0` is semantic, it's `AutoWatchID`.

Secondly when token expired, make a new watch request to server side, `serverWatchStream.isWatchPermitted` returns false, then send `WatchResponse` to `serverWatchStream.ctrlStream` with `Canceled=true`, however `WatchResponse.WatchID=creq.WatchId`, default `creq.WatchId=0`, so a watch with `WatchID=0` will be unexpectedly cancelled on the server side and client can never receive messages with `WatchID=0`.

```go
if !sws.isWatchPermitted(creq) {
    wr := &pb.WatchResponse{
        Header:       sws.newResponseHeader(sws.watchStream.Rev()),
        WatchId:      creq.WatchId, // default to 0
        Canceled:     true, // will delete 0 from ids
        Created:      true,
        CancelReason: rpctypes.ErrGRPCPermissionDenied.Error(),
    }

    select {
        case sws.ctrlStream <- wr: // send to ctrlStream
            continue
        case <-sws.closec:
            return nil
    }
}

...

case c, ok := <-sws.ctrlStream: // receive from ctrlStream
    if !ok {
        return
    }

    ...

    // track id creation
    wid := mvcc.WatchID(c.WatchId)
    if c.Canceled {
        delete(ids, wid) // delete from ids and client can never receive messages
        continue
    }
```

### DIFF
--- a/client/v3/watch.go
+++ b/client/v3/watch.go
@@ -25,7 +25,6 @@ import (
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	v3rpc "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
-
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -38,6 +37,13 @@ const (
 	EventTypePut    = mvccpb.PUT
 
 	closeSendErrTimeout = 250 * time.Millisecond
+
+	// AutoWatchID is the watcher ID passed in WatchStream.Watch when no
+	// user-provided ID is available. If pass, an ID will automatically be assigned.
+	AutoWatchID = 0
+
+	// InvalidWatchID represents an invalid watch ID and prevents duplication with an existing watch.
+	InvalidWatchID = -1
 )
 
 type Event mvccpb.Event
@@ -451,7 +457,7 @@ func (w *watcher) closeStream(wgs *watchGrpcStream) {
 
 func (w *watchGrpcStream) addSubstream(resp *pb.WatchResponse, ws *watcherStream) {
 	// check watch ID for backward compatibility (<= v3.3)
-	if resp.WatchId == -1 || (resp.Canceled && resp.CancelReason != "") {
+	if resp.WatchId == InvalidWatchID || (resp.Canceled && resp.CancelReason != "") {
 		w.closeErr = v3rpc.Error(errors.New(resp.CancelReason))
 		// failed; no channel
 		close(ws.recvc)
@@ -482,7 +488,7 @@ func (w *watchGrpcStream) closeSubstream(ws *watcherStream) {
 	} else if ws.outc != nil {
 		close(ws.outc)
 	}
-	if ws.id != -1 {
+	if ws.id != InvalidWatchID {
 		delete(w.substreams, ws.id)
 		return
 	}
@@ -544,7 +550,7 @@ func (w *watchGrpcStream) run() {
 				// TODO: pass custom watch ID?
 				ws := &watcherStream{
 					initReq: *wreq,
-					id:      -1,
+					id:      InvalidWatchID,
 					outc:    outc,
 					// unbuffered so resumes won't cause repeat events
 					recvc: make(chan *WatchResponse),
@@ -690,7 +696,7 @@ func (w *watchGrpcStream) run() {
 			if len(w.substreams)+len(w.resuming) == 0 {
 				return
 			}
-			if ws.id != -1 {
+			if ws.id != InvalidWatchID {
 				// client is closing an established watch; close it on the server proactively instead of waiting
 				// to close when the next message arrives
 				cancelSet[ws.id] = struct{}{}
@@ -742,9 +748,9 @@ func (w *watchGrpcStream) dispatchEvent(pbresp *pb.WatchResponse) bool {
 		cancelReason:    pbresp.CancelReason,
 	}
 
-	// watch IDs are zero indexed, so request notify watch responses are assigned a watch ID of -1 to
+	// watch IDs are zero indexed, so request notify watch responses are assigned a watch ID of InvalidWatchID to
 	// indicate they should be broadcast.
-	if wr.IsProgressNotify() && pbresp.WatchId == -1 {
+	if wr.IsProgressNotify() && pbresp.WatchId == InvalidWatchID {
 		return w.broadcastResponse(wr)
 	}
 
@@ -899,7 +905,7 @@ func (w *watchGrpcStream) newWatchClient() (pb.Watch_WatchClient, error) {
 	w.resumec = make(chan struct{})
 	w.joinSubstreams()
 	for _, ws := range w.substreams {
-		ws.id = -1
+		ws.id = InvalidWatchID
 		w.resuming = append(w.resuming, ws)
 	}
 	// strip out nils, if any
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -24,6 +24,8 @@ import (
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	"go.etcd.io/etcd/client/pkg/v3/verify"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/auth"
 	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/server/v3/etcdserver/apply"
@@ -286,7 +288,7 @@ func (sws *serverWatchStream) recvLoop() error {
 
 				wr := &pb.WatchResponse{
 					Header:       sws.newResponseHeader(sws.watchStream.Rev()),
-					WatchId:      creq.WatchId,
+					WatchId:      clientv3.InvalidWatchID,
 					Canceled:     true,
 					Created:      true,
 					CancelReason: cancelReason,
@@ -320,7 +322,10 @@ func (sws *serverWatchStream) recvLoop() error {
 					sws.fragment[id] = true
 				}
 				sws.mu.Unlock()
+			} else {
+				id = clientv3.InvalidWatchID
 			}
+
 			wr := &pb.WatchResponse{
 				Header:   sws.newResponseHeader(wsrev),
 				WatchId:  int64(id),
@@ -357,7 +362,7 @@ func (sws *serverWatchStream) recvLoop() error {
 			if uv.ProgressRequest != nil {
 				sws.ctrlStream <- &pb.WatchResponse{
 					Header:  sws.newResponseHeader(sws.watchStream.Rev()),
-					WatchId: -1, // response is not associated with any WatchId and will be broadcast to all watch channels
+					WatchId: clientv3.InvalidWatchID, // response is not associated with any WatchId and will be broadcast to all watch channels
 				}
 			}
 		default:
@@ -481,7 +486,10 @@ func (sws *serverWatchStream) sendLoop() {
 
 			// track id creation
 			wid := mvcc.WatchID(c.WatchId)
-			if c.Canceled {
+
+			verify.Assert(!(c.Canceled && c.Created) || wid == clientv3.InvalidWatchID, "unexpected watchId: %d, wanted: %d, since both 'Canceled' and 'Created' are true", wid, clientv3.InvalidWatchID)
+
+			if c.Canceled && wid != clientv3.InvalidWatchID {
 				delete(ids, wid)
 				continue
 			}
--- a/server/proxy/grpcproxy/watch.go
+++ b/server/proxy/grpcproxy/watch.go
@@ -22,7 +22,6 @@ import (
 	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	"go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc"
-
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
@@ -238,7 +237,7 @@ func (wps *watchProxyStream) recvLoop() error {
 			if err := wps.checkPermissionForWatch(cr.Key, cr.RangeEnd); err != nil {
 				wps.watchCh <- &pb.WatchResponse{
 					Header:       &pb.ResponseHeader{},
-					WatchId:      -1,
+					WatchId:      clientv3.InvalidWatchID,
 					Created:      true,
 					Canceled:     true,
 					CancelReason: err.Error(),
@@ -258,7 +257,7 @@ func (wps *watchProxyStream) recvLoop() error {
 				filters:  v3rpc.FiltersFromRequest(cr),
 			}
 			if !w.wr.valid() {
-				w.post(&pb.WatchResponse{WatchId: -1, Created: true, Canceled: true})
+				w.post(&pb.WatchResponse{WatchId: clientv3.InvalidWatchID, Created: true, Canceled: true})
 				wps.mu.Unlock()
 				continue
 			}
--- a/server/storage/mvcc/watcher.go
+++ b/server/storage/mvcc/watcher.go
@@ -20,11 +20,8 @@ import (
 	"sync"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
 )
-
-// AutoWatchID is the watcher ID passed in WatchStream.Watch when no
-// user-provided ID is available. If pass, an ID will automatically be assigned.
-const AutoWatchID WatchID = 0
 
 var (
 	ErrWatcherNotExist    = errors.New("mvcc: watcher does not exist")
@@ -118,7 +115,7 @@ func (ws *watchStream) Watch(id WatchID, key, end []byte, startRev int64, fcs ..
 		return -1, ErrEmptyWatcherRange
 	}
 
-	if id == AutoWatchID {
+	if id == clientv3.AutoWatchID {
 		for ws.watchers[ws.nextID] != nil {
 			ws.nextID++
 		}
--- a/tests/integration/v3_auth_test.go
+++ b/tests/integration/v3_auth_test.go
@@ -531,3 +531,57 @@ func TestV3AuthWatchAndTokenExpire(t *testing.T) {
 	watchResponse = <-wChan
 	testutil.AssertNil(t, watchResponse.Err())
 }
+
+func TestV3AuthWatchErrorAndWatchId0(t *testing.T) {
+	integration.BeforeTest(t)
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
+
+	ctx, cancel := context.WithTimeout(context.TODO(), 10*time.Second)
+	defer cancel()
+
+	users := []user{
+		{
+			name:     "user1",
+			password: "user1-123",
+			role:     "role1",
+			key:      "k1",
+			end:      "k2",
+		},
+	}
+	authSetupUsers(t, integration.ToGRPC(clus.Client(0)).Auth, users)
+
+	authSetupRoot(t, integration.ToGRPC(clus.Client(0)).Auth)
+
+	c, cerr := integration.NewClient(t, clientv3.Config{Endpoints: clus.Client(0).Endpoints(), Username: "user1", Password: "user1-123"})
+	if cerr != nil {
+		t.Fatal(cerr)
+	}
+	defer c.Close()
+
+	watchStartCh, watchEndCh := make(chan interface{}), make(chan interface{})
+
+	go func() {
+		wChan := c.Watch(ctx, "k1", clientv3.WithRev(1))
+		watchStartCh <- struct{}{}
+		watchResponse := <-wChan
+		t.Logf("watch response from k1: %v", watchResponse)
+		testutil.AssertTrue(t, len(watchResponse.Events) != 0)
+		watchEndCh <- struct{}{}
+	}()
+
+	// Chan for making sure that the above goroutine invokes Watch()
+	// So the above Watch() can get watch ID = 0
+	<-watchStartCh
+
+	wChan := c.Watch(ctx, "non-allowed-key", clientv3.WithRev(1))
+	watchResponse := <-wChan
+	testutil.AssertNotNil(t, watchResponse.Err()) // permission denied
+
+	_, err := c.Put(ctx, "k1", "val")
+	if err != nil {
+		t.Fatalf("Unexpected error from Put: %v", err)
+	}
+
+	<-watchEndCh
+}
--- a/tests/integration/v3_watch_test.go
+++ b/tests/integration/v3_watch_test.go
@@ -26,6 +26,7 @@ import (
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc"
 	"go.etcd.io/etcd/tests/v3/framework/integration"
 )
@@ -390,8 +391,8 @@ func TestV3WatchWrongRange(t *testing.T) {
 		if cresp.Canceled != tt.canceled {
 			t.Fatalf("#%d: canceled %v, want %v", i, tt.canceled, cresp.Canceled)
 		}
-		if tt.canceled && cresp.WatchId != -1 {
-			t.Fatalf("#%d: canceled watch ID %d, want -1", i, cresp.WatchId)
+		if tt.canceled && cresp.WatchId != clientv3.InvalidWatchID {
+			t.Fatalf("#%d: canceled watch ID %d, want %d", i, cresp.WatchId, clientv3.InvalidWatchID)
 		}
 	}
 }


kafuu-chino:
I found some comments that said watchID starts at 0, I don't know if there are any other effects.

ahrtr:
Thanks @kafuu-chino for the PR, but I did not get what's the issue you are fixing. The PR is also causing some test failures.

kafuu-chino:
@ahrtr 
Let me explain in detail: 

1. First, `WatchCreateRequest` does not set the value of `WatchID`, so `WatchID = 0`.
```go
func (wr *watchRequest) toPB() *pb.WatchRequest {
    req := &pb.WatchCreateRequest{
        StartRevision:  wr.rev,
        Key:            []byte(wr.key),
        RangeEnd:       []byte(wr.end),
        ProgressNotify: wr.progressNotify,
        Filters:        wr.filters,
        PrevKv:         wr.prevKV,
        Fragment:       wr.fragment,
    }
    cr := &pb.WatchRequest_CreateRequest{CreateRequest: req}
    return &pb.WatchRequest{RequestUnion: cr}
}
```

2. On server side, `WatchID = 0` means the `WatchID` is automatically generated, see `AutoWatchID`. 
```go
// AutoWatchID is the watcher ID passed in WatchStream.Watch when no
// user-provided ID is available. If pass, an ID will automatically be assigned.
const AutoWatchID WatchID = 0
```

Due to `ws.nextID = 0`, so the first watch with `WatchID = 0`.
```go
if id == AutoWatchID {
    for ws.watchers[ws.nextID] != nil {
        ws.nextID++
    }
    id = ws.nextID
    ws.nextID++
}
```

3. Watch once, initialize watch with `WatchID = 0` and wait for the token to expire. Then watch again, if the token expires, `sws.isWatchPermitted` returns false, `creq.WatchId = 0`, `Canceled = true` and `Created = true` at this time, then send `WatchResponse` to `sws.ctrlStream`.
```go
if !sws.isWatchPermitted(creq) {
    wr := &pb.WatchResponse{
        Header:       sws.newResponseHeader(sws.watchStream.Rev()),
        WatchId:      creq.WatchId, // = 0
        Canceled:     true,
        Created:      true,
        CancelReason: rpctypes.ErrGRPCPermissionDenied.Error(),
    }

    select {
        case sws.ctrlStream <- wr:
            continue
        case <-sws.closec:
            return nil
    }
}
```

4. Next handle `WatchResponse` on after `Send`, if `Canceled == true`, then `delete(ids, wid)`, `wid = 0` at this time. This operation causes the server considers the watch with `WatchID = 0` to be no longer active.
```go
case c, ok := <-sws.ctrlStream:
    if !ok {
        return
    }

if err := sws.gRPCStream.Send(c); err != nil {
    if isClientCtxErr(sws.gRPCStream.Context().Err(), err) {
        sws.lg.Debug("failed to send watch control response to gRPC stream", zap.Error(err))
    } else {
        ws.lg.Warn("failed to send watch control response to gRPC stream", zap.Error(err))
        streamFailures.WithLabelValues("send", "watch").Inc()
    }
    return
}

// track id creation
wid := mvcc.WatchID(c.WatchId)
if c.Canceled {
    delete(ids, wid)
    continue
}
```

5. Client side receives `WatchResponse`, then enter switch `case pbresp.Created:`. The new watch returns an error, the old watch with `WatchID = 0` does not respond, and no longer receives messages from the server. This is not what we want, we just want this watch to fail, and we don't want to make the previous watch unresponsive on the server side.
```go
case pbresp := <-w.respc:
    if cur == nil || pbresp.Created || pbresp.Canceled {
        cur = pbresp
    } else if cur != nil && cur.WatchId == pbresp.WatchId {
    // merge new events
    cur.Events = append(cur.Events, pbresp.Events...)
    // update "Fragment" field; last response with "Fragment" == false
    cur.Fragment = pbresp.Fragment
    }

    switch {
    case pbresp.Created:
```

I just set `ws.nextID` start to 1. 

I've found some tests that will check that the WatchID is equal to 0 causing the test to fail.
```go
if wresp.WatchId != 0 {
    t.Errorf("watchId got = %d, want = 0", wresp.WatchId)
}
```

kafuu-chino:
@ahrtr 
For test itself, I think it's easy to change the WatchID to 1, but I think changing the WatchID starting value requires considering whether other logic is affected, or maybe there's a better way to fix it.

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -270,7 +270,7 @@ func (sws *serverWatchStream) recvLoop() error {
 			if !sws.isWatchPermitted(creq) {
 				wr := &pb.WatchResponse{
 					Header:       sws.newResponseHeader(sws.watchStream.Rev()),
-					WatchId:      creq.WatchId,
+					WatchId:      -1,
```
ahrtr, line 273: Probably we should check the value of `creq.WatchId`, how about the logic something like below?
```
watchId := creq.WatchId
if watchId == int64(mvcc.AutoWatchID) {
	watchId = -1
}
```

Another choice is to make watchId starting with 1 instead of 0. But we need to fix some test cases, and it also change the default behavior, accordingly might have some impact on client applications which depend on the watch. 

I manually verified that this PR indeed can fix the issue https://github.com/etcd-io/etcd/issues/12385. 

But the overall design looks weird to me. The existing watch can continue to work even the simple token expired, because the client can retry the authentication. But creating new watch stream using exactly the same client doesn't work, it's a little strange. 

It seems that we need to get the overall design clarified firstly. WDYT @mitake

kafuu-chino: I set  watchId starting with 1 at the beginning, all issues are resolved and do not conflict with `AutoWatchID` but test failed. If can fix test, I think it is better to start with 1.

I haven't found a place to set `creq.WatchId` yet, so it should always be mvcc.autowatchid?

mitake: I think it's better to use -1 as an invalid watch ID for preserving the current behavior of watch API like what current PR does. How about introducing a new const like `AutoWatchID`?

Let me check the behavior introduced by this PR in a few days. I only tested a case which don't rely on it.

ahrtr: > I haven't found a place to set creq.WatchId yet, so it should always be mvcc.autowatchid?

Yes, it's basically true. In theory, users do not necessarily use the etcd client SDK. When they construct the `wathRequest` themselves, then watchId might not be 0 anymore. But the possibility should be very low, and it isn't the recommended way. So probably we can assume that it's always `mvcc.AutoWatchID` for any `pb.WatchRequest_CreateRequest`. 

> I think it's better to use -1 as an invalid watch ID for preserving the current behavior of watch API like what current PR does.

Based on point above (watchId always be `mvcc.AutoWatchID` for any `pb.WatchRequest_CreateRequest`), it makes sense to always use `-1` as an invalid watch ID

mitake: > In theory, users do not necessarily use the etcd client SDK. When they construct the wathRequest themselves, then watchId might not be 0 anymore. But the possibility should be very low, and it isn't the recommended way.

Yeah using non standard SDK is technically possible... but I think it's good to exclude it because it's quite irregular as you point out.

ahrtr: Adding InvalidWatchID is a good idea to me. 

Regarding to adding a condition like` c.WatchID != InvalidWatchID`... it shouldn't happen otherwise it's a bug. So I think we should add a [Assert](https://github.com/etcd-io/etcd/blob/22cc6826064aabea365480e63c43b352c42eba07/client/pkg/verify/verify.go#L76).  For example [schedule.go#L92](https://github.com/etcd-io/etcd/blob/22cc6826064aabea365480e63c43b352c42eba07/pkg/schedule/schedule.go#L92)

mitake: Yeah, I think using `Assert` is a better idea 👍

kafuu-chino: I have revised it as suggested and any other suggestions for the Assert location?

kafuu-chino:
@ahrtr @mitake 
Help review it sometime.

mitake:
@kafuu-chino Could you squash the 2 commits into one, and change the commit title like this? `*: avoid closing a watch with ID 0 incorrectly`
I think the test failure isn’t related to your change, so running tests on my local machine.

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -465,7 +466,10 @@ func (sws *serverWatchStream) sendLoop() {
 
 			// track id creation
 			wid := mvcc.WatchID(c.WatchId)
-			if c.Canceled {
+
+			verify.Assert((c.Canceled && c.Created) && wid != mvcc.InvalidWatchID, "the watch response invalid")
```
ahrtr, line 470:
```suggestion
			verify.Assert((c.Canceled && c.Created) && wid != mvcc.InvalidWatchID, fmt.Sprintf("unexpected watchId: %d, wanted: %d, since both 'Canceled' and 'Created' are true", wid, mvcc.InvalidWatchID))
```

server/etcdserver/api/v3rpc/watch.go
ahrtr: > I think `WatchResponse` with `Created && Canceled` and valid ID is possible in other paths, e.g. https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L308
> Could you update the assert condition for checking `CancelReason`? @kafuu-chino

The [Watch](https://github.com/etcd-io/etcd/blob/e9c53a0875aa048b79f1844d6cd09973691d1262/server/storage/mvcc/watcher.go#L108) always returns `-1` when there is an error. Please replace all the `-1` with the const `InvalidWatchID`. cc @kafuu-chino

kafuu-chino: > I think `WatchResponse` with `Created && Canceled` and valid ID is possible in other paths, e.g. https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L308 Could you update the assert condition for checking `CancelReason`? @kafuu-chino

I think the `WatchId` shoud be `InvalidWatchID` if `err != nil`.

https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L310

For example: if we send duplicated `WatchId` from client, err will be `ErrWatcherDuplicateID`, the `WatchId` will cancel by server.

Then the `Assert` will not trigger unless there is a bug.
@mitake

kafuu-chino: @ahrtr 
I tried to replace all the -1 I found , then squashed commit.

ahrtr:
@kafuu-chino The change looks good to me. But please rebase this PR instead of merging main into your dev branch.
Firstly sync your [main branch](https://github.com/kafuu-chino/etcd) with the upstream etcd. Afterwards, execute commands something like below,
```
$ git checkout main
$ git pull
$ git checkout fix
$ git rebase -i main  # keep the "pick" in the first line, and change all the following lines' "pick" to "s"
```

kafuu-chino:
@ahrtr 
Like this?

server/storage/mvcc/watcher.go
```diff
@@ -22,9 +22,14 @@ import (
...
-// AutoWatchID is the watcher ID passed in WatchStream.Watch when no
-// user-provided ID is available. If pass, an ID will automatically be assigned.
-const AutoWatchID WatchID = 0
+const (
+	// InvalidWatchID represents an invalid watch ID and prevents duplication with an existing watch.
+	InvalidWatchID WatchID = -1
+
+	// AutoWatchID is the watcher ID passed in WatchStream.Watch when no
+	// user-provided ID is available. If pass, an ID will automatically be assigned.
+	AutoWatchID WatchID = 0
```
ahrtr, line 31: Please consider to define the two const in [clientv3](https://github.com/etcd-io/etcd/blob/main/client/v3/watch.go), so as to remove the `clientv3` dependency on `server`

client/v3/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/stretchr/testify v1.7.2
 	go.etcd.io/etcd/api/v3 v3.6.0-alpha.0
 	go.etcd.io/etcd/client/pkg/v3 v3.6.0-alpha.0
+	go.etcd.io/etcd/server/v3 v3.6.0-alpha.0
```
ahrtr, line 12: remove this after moving the two const. Try to run "`go mod tidy`"

client/v3/go.mod
```diff
@@ -42,6 +43,7 @@ require (
 replace (
 	go.etcd.io/etcd/api/v3 => ../../api
 	go.etcd.io/etcd/client/pkg/v3 => ../pkg
+	go.etcd.io/etcd/server/v3 => ../../server
```
ahrtr, line 46: remove this as well

client/v3/watch.go
```diff
@@ -24,6 +24,7 @@ import (
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	v3rpc "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
```
ahrtr, line 27: same as above

server/storage/mvcc/watcher.go
kafuu-chino: Are these two constants defined both on the client and server?

server/etcdserver/api/v3rpc/watch.go
mitake: @kafuu-chino I see, but actually I could trigger the assert in `TestKVCompact`. Let me check again. It's great if you can run the test on your side too.

kafuu-chino: @mitake I run `TestKVCompact ` on my local system but no assert triggerred, did you update my later commit？

mitake: @kafuu-chino yeah the assert is still failing like below:
```

goroutine 252 [chan receive]:
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*serverWatchStream).sendLoop.func1()
	/Users/mitake/go/src/go.etcd.io/etcd/server/etcdserver/api/v3rpc/watch.go:373 +0xac
panic({0x104c27660, 0x140036801f0})
	/usr/local/go/src/runtime/panic.go:838 +0x204
go.etcd.io/etcd/client/pkg/v3/verify.Assert(...)
	/Users/mitake/go/src/go.etcd.io/etcd/client/pkg/verify/verify.go:78
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*serverWatchStream).sendLoop(0x14000540f00)
	/Users/mitake/go/src/go.etcd.io/etcd/server/etcdserver/api/v3rpc/watch.go:473 +0xb10
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*watchServer).Watch.func1()
	/Users/mitake/go/src/go.etcd.io/etcd/server/etcdserver/api/v3rpc/watch.go:181 +0x28
created by go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*watchServer).Watch
	/Users/mitake/go/src/go.etcd.io/etcd/server/etcdserver/api/v3rpc/watch.go:180 +0x1e8

```

Anyway the assert isn't mandatory. Passing `InvalidWatchID` to delete() or looking it up from `ids` isn't harmful. I think it's ok to remove. Or we can update the condition to check `CancelReason` like `verify.Assert(c.CancelReason == "" && wid != mvcc.InvalidWatchID, "unexpected watchId: %d, wanted: %d, since CancelReason is empty", wid, mvcc.InvalidWatchID)`. How do you think?

kafuu-chino: @mitake 
I only found two places to return both `Canceled` and `Created` are true and can trigger the assert.
1. https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L271
2. https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L308

What do you think which cause assert triggerred? Or i missed something and can you help to supplement?

I agree with :
> Passing `InvalidWatchID` to delete() or looking it up from ids isn't harmful. 

Because it will not affect the normal watch.

Does `c.CancelReason == "" ` equal to `c.Canceled == flase`? I think it means the watch will not be deleted and the `WatchId` will be a normal watch id. 

In my opinion, the purpose of this assert is that when a watch is newly created and an error is encountered ( both `Canceled` and `Created` are true), do not cause other normal watches to be deleted (`WatchId` MUST BE `InvalidWatchID`), it's hard to detect when it happens.

mitake: @kafuu-chino In the case of `TestKVCompaction`, https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L308 is causing the assert. The condition is `c.Created == true && c.Canceled == false && c.WatchId == 0 && c.CancelReason == ""`.

On the second thought, the condition should be like `(c.Canceled || c.Created) && wid != mvcc.InvalidWatchID`? If it's false, the branches https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L468-L472 will be executed and unintentional operations against invalid watch ID will happen (although it's not harmful).

It's great if you can try to run the test case too: `TESTCASE=TestKVCompact TEST_OPTS="PASSES='integration'" make test`

kafuu-chino: @mitake 
The assert condition really wrong and I've got the condition backwards.

If condition like `(c.Canceled || c.Created) && wid != mvcc.InvalidWatchID`, there will be problems here: https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L342

1. `c.Created == false` and `c.Canceled == false` 
2. then `(c.Canceled || c.Created) == false`
3. and `(c.Canceled || c.Created) && wid != mvcc.InvalidWatchID` == false
4. `condition == false` and assert will be trigger

How about `!(c.Canceled && c.Created) || wid == mvcc.InvalidWatchID`？

mitake: @kafuu-chino Sorry the above my condition is very wrong... Especially if it includes `&& wid != mvcc.InvalidWatchID`, it cannot handle an invalid `wid` always. I guess I pasted a wrong version :(
I think your condition is correct. Sorry for the confusion.

mitake: Let me review the entire PR, I think I can reply sometime tomorrow.

mitake (commented):
Overall LGTM, could you rebase on the latest main branch correctly? The commit https://github.com/etcd-io/etcd/pull/14296/commits/9fba81a14531285fe69cadac7bd65aab0d01686f comes from the PR https://github.com/etcd-io/etcd/pull/14377 so it shouldn't be included in this PR.

`server/etcdserver/api/v3rpc/watch.go`, especially `serverWatchStream.recvLoop()` is hard to test with unit tests. Do you think it's possible to add integration or e2e test?

ahrtr (changes requested):
Please update per  @mitake 's comment. 

You need to rebase this PR instead of merging others branch into it.

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -270,7 +271,7 @@ func (sws *serverWatchStream) recvLoop() error {
 			if !sws.isWatchPermitted(creq) {
 				wr := &pb.WatchResponse{
 					Header:       sws.newResponseHeader(sws.watchStream.Rev()),
-					WatchId:      creq.WatchId,
+					WatchId:      int64(mvcc.InvalidWatchID),
```
ahrtr, line 274: Can we use `clientv3.InvalidWatchID` indirectly? Same to following places.

server/storage/mvcc/watcher.go
ahrtr, line 31: I suggest to remove the duplicated definition, and reuse the definitions in clientv3.

kafuu-chino:
@mitake 
That's the idea for my local test.
1. Watch a key.
2. Watch a new key and enter this code https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L271. I set the `--auth-token-ttl` of the server and wait for the token to expire.
3. Set new value of key in step 1, and check whether message from watch can be received normally.

I need to read the tests package code first and it's take some time to think test code.

server/proxy/grpcproxy/watch.go
```diff
@@ -238,7 +239,7 @@ func (wps *watchProxyStream) recvLoop() error {
 			if err := wps.checkPermissionForWatch(cr.Key, cr.RangeEnd); err != nil {
 				wps.watchCh <- &pb.WatchResponse{
 					Header:       &pb.ResponseHeader{},
-					WatchId:      -1,
+					WatchId:      int64(mvcc.InvalidWatchID),
```
ahrtr, line 242: It seems you are still using `mvcc.InvalidWatchID `instead of `clientv3.InvalidWatchID`

kafuu-chino: Sorry, I’m missing some place and I did a whole check on the commit.

mitake:
@kafuu-chino I'm writing a test case in https://github.com/etcd-io/etcd/pull/14322/commits/ff9c6612f7dff2dd730ee4be0c4ab167ba042e6f for https://github.com/etcd-io/etcd/pull/14322
For now the test case is still in progress (for some reasons the unary interceptor refreshes the token so the mechanism I'm adding isn't used), but it might also be useful for this PR too. If you want to try it as a starting point, feel free to use it.

ahrtr (approved):
LGTM

Thank you @kafuu-chino 

The only concern is that there is no new test case for the code change.  But it's OK to @mitake or @kafuu-chino  can address the test case in a separate PR.

Leave it to @mitake to approve & merge this PR.

mitake:
@kafuu-chino @ahrtr I finalized the test for my PR in https://github.com/etcd-io/etcd/pull/14322/commits/94fd1612949e5b4279e923efdec7b2de67e5d3bc I think it can be extended for covering the change in this PR. Let me think about it. Probably I'll be able to share my update sometime early next week.

mitake:
@kafuu-chino On the second thought, I think this change is essentially not limited to auth token TTL. I wrote below integration test and checked that the test can fail in the main branch and run successfully with your branch. Could you check it's a suitable test case for your change? If it's ok, could you include it in this PR? (path is `tests/integration/v3_auth_test.go`)

```
func TestV3AuthWatchErrorAndWatchId0(t *testing.T) {
	integration.BeforeTest(t)
	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
	defer clus.Terminate(t)

	ctx, cancel := context.WithTimeout(context.TODO(), 10*time.Second)
	defer cancel()

	users := []user{
		{
			name:     "user1",
			password: "user1-123",
			role:     "role1",
			key:      "k1",
			end:      "k2",
		},
	}
	authSetupUsers(t, integration.ToGRPC(clus.Client(0)).Auth, users)

	authSetupRoot(t, integration.ToGRPC(clus.Client(0)).Auth)

	c, cerr := integration.NewClient(t, clientv3.Config{Endpoints: clus.Client(0).Endpoints(), Username: "user1", Password: "user1-123"})
	if cerr != nil {
		t.Fatal(cerr)
	}
	defer c.Close()

	watchEndCh := make(chan interface{})

	go func() {
		wChan := c.Watch(ctx, "k1", clientv3.WithRev(1))
		watchResponse := <-wChan
		fmt.Printf("watch response from k1: %v\n", watchResponse)
		testutil.AssertTrue(t, len(watchResponse.Events) != 0)
		watchEndCh <- struct{}{}
	}()

	// Sleep for making sure that the above goroutine invokes Watch()
	// So the above Watch() can get watch ID = 0
	time.Sleep(1 * time.Second)

	wChan := c.Watch(ctx, "non-allowed-key", clientv3.WithRev(1))
	watchResponse := <-wChan
	testutil.AssertNotNil(t, watchResponse.Err()) // permission denied

	_, err := c.Put(ctx, "k1", "val")
	if err != nil {
		t.Fatalf("Unexpected error from Put: %v", err)
	}

	<-watchEndCh
}
```

kafuu-chino:
@mitake Ok, I'll try the test later.

kafuu-chino:
@mitake 
I think the test is fine and add it to the PR.

mitake:
@kafuu-chino Could you rebase this PR based on the latest main? My PR introduced a conflict in the integration test file, sorry.

kafuu-chino:
@mitake 
Rebase finished.

mitake (approved):
LGTM, thanks a lot @kafuu-chino ! It's great if @ahrtr can cross check the test case.

tests/integration/v3_auth_test.go
```diff
@@ -531,3 +531,56 @@ func TestV3AuthWatchAndTokenExpire(t *testing.T) {
...
+		t.Fatal(cerr)
+	}
+	defer c.Close()
+
+	watchEndCh := make(chan interface{})
+
+	go func() {
+		wChan := c.Watch(ctx, "k1", clientv3.WithRev(1))
+		watchResponse := <-wChan
+		fmt.Printf("watch response from k1: %v\n", watchResponse)
```
ahrtr, line 567:
```suggestion
		t.Logf("watch response from k1: %v\n", watchResponse)
```

tests/integration/v3_auth_test.go
```diff
@@ -531,3 +531,56 @@ func TestV3AuthWatchAndTokenExpire(t *testing.T) {
...
+		testutil.AssertTrue(t, len(watchResponse.Events) != 0)
+		watchEndCh <- struct{}{}
+	}()
+
+	// Sleep for making sure that the above goroutine invokes Watch()
+	// So the above Watch() can get watch ID = 0
+	time.Sleep(1 * time.Second)
```
ahrtr, line 574: Please use channel or `WaitGroup` to sync between goroutines instead of hard code 1s sleep.

tests/integration/v3_auth_test.go
```diff
@@ -531,3 +531,56 @@ func TestV3AuthWatchAndTokenExpire(t *testing.T) {
 	watchResponse = <-wChan
 	testutil.AssertNil(t, watchResponse.Err())
 }
+
+func TestV3AuthWatchErrorAndWatchId0(t *testing.T) {
```
ahrtr, line 535: The test case is OK, but I do not any relationship between this case and the [issue](https://github.com/etcd-io/etcd/issues/12385) this PR is trying to fix.

tests/integration/v3_auth_test.go
mitake: Ah that's true, sorry for that. Could you update it @kafuu-chino ? 🙏

tests/integration/v3_auth_test.go
ahrtr: Thanks @mitake for the explanation. 



> Actually auth token TTL expiration isn't the only one reason which can cause unexpected cancelling watch with ID=0.

I follow this issue intermittently, so I might lose some context. I agree with you on your point, but I don't think it verifies the scenario of the original issue https://github.com/etcd-io/etcd/issues/12385 ? My understand the client should can successfully Watch even when the auth token is expired, so we need to add a case to verify it. Please correct me if I am wrong.

mitake: > My understand is that the client should can successfully Watch even when the auth token is expired, so we need to add a case to verify it.

Correct, and this behavior is checked by the test case added in another PR: https://github.com/etcd-io/etcd/pull/14322 The test case is https://github.com/etcd-io/etcd/pull/14322/files#diff-c42aac07498140d89b902c1e02edb73076615172896ba431ec99319636432e1dR501 (`TestV3AuthWatchAndTokenExpire`). So the test case added in this PR (`TestV3AuthWatchErrorAndWatchId0`) focuses only on the issue happens to watch with ID=0. I think the 2 test cases are enough for covering the issues reported in https://github.com/etcd-io/etcd/issues/12385 (actually there are 2 related but independent issues). How do you think @ahrtr ?

ahrtr: Thanks @mitake for the clarification, which makes sense.

tests/integration/v3_auth_test.go
kafuu-chino: @mitake OK, I will update later.

kafuu-chino:
@mitake @ahrtr  modification of test has committed, please review.

ahrtr:
Overall look good to me. Thanks!

mitake:
@kafuu-chino please let me know if you don’t have time to backport the change to release-3.4 and 3.5. Probably I’ll be able to help.

kafuu-chino:
> @kafuu-chino please let me know if you don’t have time to backport the change to release-3.4 and 3.5. Probably I’ll be able to help.

Sorry, I just got back from my vacation, I'll handle it today.

PR #14299, opened 2022-08-02
tests: fix status responses size and nil

clarkfw:
As I migrate tests to common framework,  I found 2 problems in the following code
1. `resp` is filled with `nil` before it is appended
2. `--endpoints` argument is redundant, as `ctl.spawnJsonCmd()` has already taken care of it. With this redundant`--endpoints`, status of one endpoint appears 2 times in `epStatus`

https://github.com/etcd-io/etcd/blob/4f0e92d94ceacdd0af26e7268764a5bce7b0a3eb/tests/framework/e2e/etcdctl.go#L290-L304

### DIFF
--- a/tests/common/status_test.go
+++ b/tests/common/status_test.go
@@ -1,0 +1,58 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+func TestStatus(t *testing.T) {
+
+	testRunner.BeforeTest(t)
+
+	for _, tc := range clusterTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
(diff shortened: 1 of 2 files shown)


clarkfw:
> @clarkfw thanks for the PR. I would appreciate it if you could provide some details in the PR description for these changes and the issue(s) it solves.

I've update the comment. By the way, is it better to open an issue before committing a PR like this(bug fix)?

spzala:
> > @clarkfw thanks for the PR. I would appreciate it if you could provide some details in the PR description for these changes and the issue(s) it solves.
> 
> I've update the comment. By the way, is it better to open an issue before committing a PR like this(bug fix)?

Thank you @clarkfw No, the issue is not required to open. But, when there is no issue related to the PR, the details provided in the PR description help reviewers to understand the context quickly for the proposed changes.

clarkfw:
> > > @clarkfw thanks for the PR. I would appreciate it if you could provide some details in the PR description for these changes and the issue(s) it solves.
> > 
> > 
> > I've update the comment. By the way, is it better to open an issue before committing a PR like this(bug fix)?
> 
> Thank you @clarkfw No, the issue is not required to open. But, when there is no issue related to the PR, the details provided in the PR description help reviewers to understand the context quickly for the proposed changes.

Thanks for the detailed explanation, which makes it easier for me to participate in contribution.

ahrtr (approved):
LGTM 

Thank you @clarkfw

ahrtr:
I see that you made the same change in https://github.com/etcd-io/etcd/pull/14304, so you need to rebase that PR after this one is merged.

ahrtr:
@spzala This is only test case change, the PR is fixing the golang slice length related issue. For example, the value 10 is actually added as the 3rd element (index: 2) instead of the first element (index: 0). It's a classic error when people handle golang slice. 

```
s := make([]int, 2)
s = append(s, 10)
```

spzala:
> @spzala This is only test case change, the PR is fixing the golang slice length related issue. For example, the value 10 is actually added as the 3rd element (index: 2) instead of the first element (index: 0). It's a classic error when people handle golang slice.
> 
> ```
> s := make([]int, 2)
> s = append(s, 10)
> ```

Ok, lgtm! Thanks @ahrtr @clarkfw  I meant to review it after my initial general comment on PR description but coming back to work today from couple of days vacation/long weekend :)

clarkfw:
Thank you for taking time to review🫰

PR #14300, opened 2022-08-02
Enhance the WAL file related error

ahrtr:
This PR should be helpful to debug https://github.com/etcd-io/etcd/issues/14252

The `ErrFileNotFound` was used for for three cases:
1. There is no any WAL files (probably due to no read permission);
2. There is no WAL files matching the snapshot index;
3. The WAL file seqs do not increase continuously.

It's not good for debug when users see the `ErrFileNotFound` error,
so in this PR, a different error is returned for each case above.

### DIFF
--- a/server/storage/wal/wal.go
+++ b/server/storage/wal/wal.go
@@ -377,9 +377,12 @@ func selectWALFiles(lg *zap.Logger, dirpath string, snap walpb.Snapshot) ([]stri
 	}
 
 	nameIndex, ok := searchIndex(lg, names, snap.Index)
-	if !ok || !isValidSeq(lg, names[nameIndex:]) {
-		err = ErrFileNotFound
-		return nil, -1, err
+	if !ok {
+		return nil, -1, fmt.Errorf("wal: file not found which matches the snapshot index '%d'", snap.Index)
+	}
+
+	if !isValidSeq(lg, names[nameIndex:]) {
+		return nil, -1, fmt.Errorf("wal: file sequence numbers (starting from %d) do not increase continuously", nameIndex)
 	}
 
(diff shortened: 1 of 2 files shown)


ahrtr:
cc @spzala

spzala (approved):
@ahrtr thanks, this can be very helpful with debugging. I don't see anything rely on the return err, `ErrFileNotFound` but to be on safe side, probably good idea to update CHANGELOG around error enhancement? (e.g. Verify() in the wal.go will return different error with these changes) Thanks!

ahrtr:
> probably good idea to update CHANGELOG around error enhancement? (e.g. Verify() in the wal.go will return different error with these changes)

Thanks @spzala for the comment. I thought about updating the CHANGELOG, but didn't do it, reasons:
1. We do not have detailed document on WAL, so nobody will understand the CHANGELOG item without digging into the source code;
2. The change is a very minor change and just to enhance the error message, which is only useful for debugging related issue. 

So I'd like not to update the CHANGELOG, what do you think?

spzala:
> > probably good idea to update CHANGELOG around error enhancement? (e.g. Verify() in the wal.go will return different error with these changes)
> 
> Thanks @spzala for the comment. I thought about updating the CHANGELOG, but didn't do it, reasons:
> 
>     1. We do not have detailed document on WAL, so nobody will understand the CHANGELOG item without digging into the source code;
> 
>     2. The change is a very minor change and just to enhance the error message, which is only useful for debugging related issue.
> 
> 
> So I'd like not to update the CHANGELOG, what do you think?

@ahrtr ok, that's fair.  It's minor as you said and I don't have a hard opinion.

PR #14304, opened 2022-08-03
tests: add WaitLeader function to common framework

clarkfw:
This PR relates to #14305

### DIFF
--- a/tests/common/wait_leader_test.go
+++ b/tests/common/wait_leader_test.go
@@ -1,0 +1,79 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
+)
+
+func TestWaitLeader(t *testing.T) {
+	testRunner.BeforeTest(t)
+
+	for _, tc := range clusterTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			clus := testRunner.NewCluster(ctx, t, tc.config)
+			defer clus.Close()
+
+			leader := clus.WaitLeader(t)
+			if leader < 0 || leader >= len(clus.Members()) {
+				t.Fatalf("WaitLeader failed for the leader index (%d) is out of range, cluster member count: %d", leader, len(clus.Members()))
+			}
+		})
+	}
+}
+
+func TestWaitLeader_MemberStop(t *testing.T) {
+	testRunner.BeforeTest(t)
+	tcs := []testCase{
+		{
+			name:   "PeerTLS",
+			config: config.ClusterConfig{ClusterSize: 3, PeerTLS: config.ManualTLS},
+		},
+		{
+			name:   "PeerAutoTLS",
+			config: config.ClusterConfig{ClusterSize: 3, PeerTLS: config.AutoTLS},
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			clus := testRunner.NewCluster(ctx, t, tc.config)
+			defer clus.Close()
+
+			lead1 := clus.WaitLeader(t)
+			if lead1 < 0 || lead1 >= len(clus.Members()) {
+				t.Fatalf("WaitLeader failed for the leader index (%d) is out of range, cluster member count: %d", lead1, len(clus.Members()))
+			}
+
+			clus.Members()[lead1].Stop()
+			lead2 := clus.WaitLeader(t)
+			if lead2 < 0 || lead2 >= len(clus.Members()) {
+				t.Fatalf("WaitLeader failed for the leader index (%d) is out of range, cluster member count: %d", lead2, len(clus.Members()))
+			}
+
+			if lead1 == lead2 {
+				t.Fatalf("WaitLeader failed for the leader(index=%d) did not change as expected after a member stopped", lead1)
+			}
+		})
+	}
+}
--- a/tests/framework/e2e.go
+++ b/tests/framework/e2e.go
@@ -17,12 +17,16 @@ package framework
 import (
 	"context"
 	"os"
+	"strings"
 	"testing"
+	"time"
 
 	"go.etcd.io/etcd/client/pkg/v3/testutil"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
+
+const TickDuration = 10 * time.Millisecond
 
 type e2eRunner struct{}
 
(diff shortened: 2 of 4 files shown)


tests/framework/e2e.go
```diff
@@ -92,6 +95,67 @@ func (c *e2eCluster) Members() (ms []Member) {
...
+// or fails the test (if not established in 30s).
+func (c *e2eCluster) WaitLeader(t testing.TB) int {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	return c.WaitMembersForLeader(ctx, t, c.Members())
+}
+
+// WaitMembersForLeader waits until given members agree on the same leader,
+// and returns its 'index' in the 'membs' list
+func (c *e2eCluster) WaitMembersForLeader(ctx context.Context, t testing.TB, membs []Member) int {
```
lavacat, line 108: have you considered refactoring this and cluster.waitMembersForLeader into a common function to avoid code duplication?
Seams like you've changed some logic here and you aren't using m.Server.Lead(), is there a reason?

clarkfw: This idea of refracting has once crossed my mind. Then I realised that there is no equivalent function to `m.Server.Lead()` in e2e cluster code, or something like `m.Server.StopNotify()`.

e2e cluster has less control of details of a member. To get member ID in e2e cluster, `membs[i].Client().Status()` is the way to go to my knowledge.

I did change some logic. For example, in integration cluster, `waitMembersForLeader` select on `m.Server.StopNotify()` to detect if an endpoint has stopped; in e2e cluster, if an endpoint has stopped, the test would fail due to network error in `resp, err := membs[i].Client().Status()`. I haven't figured out how to deal with this error yet.


https://github.com/etcd-io/etcd/blob/9bb134233a42c0339cd7c5355dd9c2772e580944/tests/framework/integration/cluster.go#L455-L470

clarkfw: If you are familiar with this part, please give me more advice. It would be helpful to me🫰

lavacat:
Created amended version of your PR https://github.com/etcd-io/etcd/pull/14313 as an example. It's a draft to showcase possible solution. I'll close it out and we can keep working on this PR.

The idea is to convert WaitLeader and WaitMembersForLeader to helper functions that use Cluster interface. 
In integration/cluster.go, integrationCluster can be converted and we can use the same helper.

@serathius this might be a good use case for you to review. How do we share implementation of methods that only use Cluster interface? Instead of helper methods we can use embeddings, but that will change your original interface.

clarkfw:
> Created amended version of your PR #14313 as an example. It's a draft to showcase possible solution. I'll close it out and we can keep working on this PR.
> 
> The idea is to convert WaitLeader and WaitMembersForLeader to helper functions that use Cluster interface. In integration/cluster.go, integrationCluster can be converted and we can use the same helper.
> 
> @serathius this might be a good use case for you to review. How do we share implementation of methods that only use Cluster interface? Instead of helper methods we can use embeddings, but that will change your original interface.

It's a neat solution in a more idiomatic way. I really like it :)

serathius:
cc @ahrtr @spzala as I'm on vacations.

tests/framework/e2e/etcdctl.go
```diff
@@ -292,13 +292,13 @@ func (ctl *EtcdctlV3) Status() ([]*clientv3.StatusResponse, error) {
...
-	err := ctl.spawnJsonCmd(&epStatus, "endpoint", "status", "--endpoints", strings.Join(ctl.endpoints, ","))
+	err := ctl.spawnJsonCmd(&epStatus, "endpoint", "status")
 	if err != nil {
 		return nil, err
 	}
 	resp := make([]*clientv3.StatusResponse, len(epStatus))
-	for _, e := range epStatus {
-		resp = append(resp, e.Status)
+	for i, e := range epStatus {
+		resp[i] = e.Status
```
ahrtr, line 301: nice catch !!!

tests/framework/e2e/etcdctl.go
```diff
@@ -292,13 +292,13 @@ func (ctl *EtcdctlV3) Status() ([]*clientv3.StatusResponse, error) {
 		Endpoint string
 		Status   *clientv3.StatusResponse
 	}
-	err := ctl.spawnJsonCmd(&epStatus, "endpoint", "status", "--endpoints", strings.Join(ctl.endpoints, ","))
+	err := ctl.spawnJsonCmd(&epStatus, "endpoint", "status")
```
ahrtr, line 295: good!!!

It's unnecessary, because [(*EtcdctlV3) flags()](https://github.com/etcd-io/etcd/blob/216f0ae7e94d33bd97e51da119a674fdfc578050/tests/framework/e2e/etcdctl.go#L274) adds `"--endpoints"`

tests/framework/e2e.go
```diff
@@ -16,14 +16,17 @@ package framework
...
-	"os"
-	"testing"
-
 	"go.etcd.io/etcd/client/pkg/v3/testutil"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"os"
+	"strings"
+	"testing"
+	"time"
```
ahrtr, line 25: Please put the golang build-in packages in the first group (immediately follow `context` (line 18))

ahrtr (approved):
LGTM

Thank you @clarkfw 

cc @spzala @serathius @ptabor

spzala (changes requested):
@clarkfw thanks for your great work helping with the common framework testing. I have few inline comments.

tests/framework/e2e.go
ahrtr: Thanks @spzala. Usually we should try to implement common functions/methods which can be used by both e2e and function test. But it seems that there is no easy way to implement a common `waitMembersForLeader`, because e2e and integration depend on different solution to wait for the leader.  I think it's OK for now.

tests/framework/e2e.go
```diff
@@ -92,6 +96,72 @@ func (c *e2eCluster) Members() (ms []Member) {
...
+		leaders = make(map[uint64]struct{})
+		members = make(map[uint64]int)
+		time.Sleep(10 * TickDuration)
+	}
+	for l := range leaders {
+		if index, ok := members[l]; ok {
+			t.Logf("members agree on a leader, members:%v , leader:%v", members, l)
+			return index
+		}
+		t.Fatalf("members agree on a leader which is not one of members, members:%v , leader:%v", members, l)
```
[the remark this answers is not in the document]
ahrtr: The etcdctl get the all member's status [one by one](https://github.com/etcd-io/etcd/blob/b2726c4da396d66edc744248befb876d9cfdd036/etcdctl/ctlv3/command/ep_command.go#L198-L208), obviously it isn't an atomic operation. If there is a leader change in between, then you will get different leaders. Of course, the possibility should be low.

tests/common/wait_leader_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			clus := testRunner.NewCluster(ctx, t, tc.config)
+			defer clus.Close()
+
+			lead1 := clus.WaitLeader(t)
+			if lead1 < 0 || lead1 >= len(clus.Members()) {
+				t.Fatalf("failed to wait leader")
```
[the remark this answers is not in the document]
ahrtr: The error is unexpected, which means that it might be a potential bug. I think it's OK to just fail the test case. 

Minor comment, there is no any parameter, so `t.Fatal `should be used instead of `t.Fatalf`.  But the better way to get info included in the error message,
```
t.Fatalf("The leader index (%d) is out of range, cluster member count: %d", len(clus.Members()))
```

tests/framework/e2e.go
lavacat: @ahrtr new implementation of WaitLeader is using only framework/interface.go interfaces. So, in principle it can be moved into a shared helper function. I was trying get a POC for this, but run into some circular imports issues. If there is interest in doing this, please open a new issue.

lavacat: @spzala FYI, [commented](https://github.com/etcd-io/etcd/pull/14328#issuecomment-1211011685) about difficulty with reusing new impl in integration/cluster.go 
>one runs into circular imports when trying to call this new helper (part of 'framework' package) from framework/intergration/cluster.go (part of 'integration'), because 'framework' imports 'integration' [here](https://github.com/etcd-io/etcd/blob/main/tests/framework/integration.go#L31)
There is probably a way to get this working, but it might involve changing existing packages/interfaces, so might be not worth the effort.

ahrtr: FYI. https://github.com/etcd-io/etcd/pull/14328#issuecomment-1211394367 cc @spzala @lavacat 

I think this PR should be good to go.

tests/common/wait_leader_test.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+
+			clus.Members()[lead1].Stop()
+			lead2 := clus.WaitLeader(t)
+			if lead2 < 0 || lead2 >= len(clus.Members()) {
+				t.Fatalf("failed to wait leader")
+			}
+
+			if lead1 == lead2 {
+				t.Fatalf("leader did not change as expected after stopping memberts")
```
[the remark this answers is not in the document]
clarkfw: I've fixed the typo.

tests/common/wait_leader_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, tc := range clusterTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			clus := testRunner.NewCluster(ctx, t, tc.config)
+			defer clus.Close()
+
+			leader := clus.WaitLeader(t)
+			if leader < 0 || leader >= len(clus.Members()) {
+				t.Fatalf("failed to wait leader")
```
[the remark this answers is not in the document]
clarkfw: Good, `WaitLeader failed` is a better for the log message. I've change the log message and it contains `leader` value now.

tests/common/wait_leader_test.go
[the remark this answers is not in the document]
clarkfw: Thanks to @ahrtr for explaining very well

tests/framework/e2e.go
ahrtr: > @ahrtr new implementation of WaitLeader is using only framework/interface.go interfaces. So, in principle it can be moved into a shared helper function. I was trying get a POC for this, but run into some circular imports issues. If there is interest in doing this, please open a new issue.

@lavacat  Are you still working on the PoC? If yes, feel free to deliver a PR and we can discuss it in your PR. It's also OK to merge this PR firstly, then you can continue to work on your PoC/PR afterwards.

lavacat: No, I didn't solve circular import problem. Feel free to merge this PR

tests/framework/e2e.go
```diff
@@ -17,13 +17,17 @@ package framework
...
+	"strings"
 	"testing"
+	"time"
 
 	"go.etcd.io/etcd/client/pkg/v3/testutil"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
 
+const TickDuration = 10 * time.Millisecond
```
ahrtr, line 29: minor comment: the `TickDuration` is only used in `e2e.go`, no need to expose it (the first letter uppercase).

I am thinking probably it makes more sense to just define one public `const` to be used for both e2e and integration ( [cluster.go#L72](https://github.com/etcd-io/etcd/blob/fff5d00ccfddaf5f06fa11cad65c1ecaa8c09f87/tests/framework/integration/cluster.go#L72) ) . Of course, we can address this in a separate PR.

ahrtr:
@spzala  This PR is good to go for me. Do you have any other comments or concerns? thx

spzala (approved):
lgtm
Defer to @ahrtr Thanks @clarkfw for addressing my comments!

tests/framework/e2e.go
spzala: Thanks @ahrtr @lavacat for the discussion. sgtm!

ahrtr:
Thanks @spzala 

Great work! @clarkfw

tests/framework/e2e.go
clarkfw: It sound good👍
#14344 was opened to address this.

clarkfw:
Thanks to @ahrtr and @spzala for your help! This PR won't  have gone so well without your collaboration.

Issue #13766, opened 2022-03-08
Inconsistent revision and data occurs


liuycsd:
I'm running a 3-member etcd cluster in a testing environment,
and a k8s cluster with each kube-apiserver connecting to one etcd server via localhost.

etcd cluster: created and running with v3.5.1
k8s cluster: v1.22

I found a data inconsistent several days ago:
Some keys exists on node `x` and `y` but not on node `z`, and some on `z` but not on the other two, e.g. different pods list would be returned from different servers.
Some keys have different values between `z` and the others, and can be updated to another different value via the corresponding etcd endpoint, e.g. the kube-system/kube-controller-manager lease points to different pods on different servers and both pods can successfully update the lease via their corresponding local api-server and etcd.
While other keys, including some new created ones, are consistent.

Checking with `etcdctl endpoint status`, `raft_term`, `leader`, `raftIndex` and `raftAppliedIndex` are same, but `revision` and `dbSize` does not: `revision` on node `z` is smaller than the others (~700000 diff) and both revisions are increasing.

Checking with `etcd-dump-logs`, it seems all the 3 nodes are receiving the same raft logs.

I'm not sure how and when this happens.
But the nodes might be under loads and may run out of memory or disk spaces sometimes.
Checking and comparing db keys with `bbolt keys SNAPSHOT key`, and seaching node operating system logs for the revisions near where the difference starts, I found some slow read logs mentions those revisions and dozon of leader elections and switches during those several hours. Besides, the disk space (including the etcd data dir) of node `z` may be full and the operating system logs was cleared so I don't know what happeds exactly and I'm not sure whether this relates to the issue or not.

serathius:
cc @ptabor

ptabor:
Was this cluster originally created as 3.5.1 or it was subject to migration from 3.4.x. or 3.3.x ?

serathius:
Hey @liuycsd, can you provide etcd configuration you are running. Specifically it would be useful to know if you use `--experimental-corrupt-check-time` as would allow early detection.

liuycsd:
> Was this cluster originally created as 3.5.1 or it was subject to migration from 3.4.x. or 3.3.x ?

IIRC, it was originally created as 3.5.1.

liuycsd:
> Hey @liuycsd, can you provide etcd configuration you are running. Specifically it would be useful to know if you use `--experimental-corrupt-check-time` as would allow early detection.

Checking the running server, current config is:
[...]

Issue #11651, opened 2020-02-24
a data corruption bug in all etcd3 version when authentication is enabled


tangcong:
**What happened**:

Recently, our team(TencentCloud k8s team) encountered a serious etcd data inconsistency bug. k8s resources such as node, pods, service, and deployment were not found when you use kubectl to get resource. The cluster did not work when you deploy/update workload. 

**How to trouble-shooting it**:

the cluster status information is as follows. you can see that node-1, node-2, node-3 have same raftIndex，but node-2's revision is different from others. the number of keys per node is also inconsistent.for example, some keys are on the leader, but do not exist on the follower node. after adding the simple debug log, we found that the reason the follower node failed to apply  command  is that its auth revision is smaller than the leader.
```
[
  {
    "Endpoint": "https://node-1:2379",
    "Status": {
      "header": {
        "cluster_id": 4202817850163301400,
        "member_id": 8591302216124782000,
        "revision": 11,
        "raft_term": 5
      },
      "version": "3.4.3",
      "dbSize": 45056,
      "leader": 8717530244098866000,
      "raftIndex": 392,
      "raftTerm": 5,
      "raftAppliedIndex": 392,
      "dbSizeInUse": 32768
    }
  },
  {
    "Endpoint": "https://node-2:2379",
    "Status": {
      "header": {
        "cluster_id": 4202817850163301400,
        "member_id": 12692962784632463000,
        "revision": 1,
        "raft_term": 5
      },
      "version": "3.4.3",
      "dbSize": 49152,
      "leader": 8717530244098866000,
      "raftIndex": 392,
      "raftTerm": 5,
      "raftAppliedIndex": 392,
      "dbSizeInUse": 45056
    }
  },
  {
    "Endpoint": "https://node-3:2379",
    "Status": {
      "header": {
        "cluster_id": 4202817850163301400,
        "member_id": 8717530244098866000,
        "revision": 11,
        "raft_term": 5
      },
      "version": "3.4.3",
      "dbSize": 45056,
      "leader": 8717530244098866000,
      "raftIndex": 392,
      "raftTerm": 5,
      "raftAppliedIndex": 392,
      "dbSizeInUse": 32768
    }
  }
]
```
follower can received leader's command, we exclude cluster split brain and the implementation of raft algorithm bugs. why follower node auth revision is less than leader node? 

we add debugging log and develop a simple chaos monkey tool to reproduce it. after running for a few days, we successfully reproduced.
[...]

PR #14306, opened 2022-08-03
CHANGELOG: update version recommendation

spzala:
Update version recommendation per the current policy.

For more details, see discussions under
https://github.com/etcd-io/website/pull/601

### DIFF
--- a/CHANGELOG/CHANGELOG-4.0.md
+++ b/CHANGELOG/CHANGELOG-4.0.md
@@ -2,12 +2,7 @@
 
 Previous change logs can be found at [CHANGELOG-3.x](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.x.md).
 
-
-The minimum recommended etcd versions to run in **production** are 3.2.28+, 3.3.18+, and 3.4.2+.
-
-
 <hr>
-
 
 ## v4.0.0 (TBD)
 
--- a/CHANGELOG/README.md
+++ b/CHANGELOG/README.md
@@ -1,6 +1,6 @@
 # Production recommendation
 
-The minimum recommended etcd versions to run in **production** are 3.3.18+, 3.4.2+, v3.5.3+. Refer to the [versioning policy](https://etcd.io/docs/v3.5/op-guide/versioning/) for more details.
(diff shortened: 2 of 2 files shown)


CHANGELOG/README.md
```diff
@@ -1,6 +1,6 @@
 # Production recommendation
 
-The minimum recommended etcd versions to run in **production** are 3.3.18+, 3.4.2+, v3.5.3+. Refer to the [versioning policy](https://etcd.io/docs/v3.5/op-guide/versioning/) for more details.
+The minimum recommended etcd versions to run in **production** are 3.4.2+ and v3.5.3+. Refer to the [versioning policy](https://etcd.io/docs/v3.5/op-guide/versioning/) for more details.
```
ahrtr, line 3: There is a known data inconsistent issue (https://github.com/etcd-io/etcd/issues/11651) in both 3.4, and fxied in 3.4.8.  Although the data inconsistent issue (https://github.com/etcd-io/etcd/issues/13766) in 3.5 was fixed in 3.5.3, but there is a regression which is fixed in 3.5.4. 

So I would suggest to change the minimum recommended etcd version to 3.4.8+ and 3.5.4+.

spzala: @ahrtr ok, that sounds good. Thanks!

CHANGELOG/CHANGELOG-4.0.md
```diff
@@ -3,7 +3,7 @@
 Previous change logs can be found at [CHANGELOG-3.x](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.x.md).
 
 
-The minimum recommended etcd versions to run in **production** are 3.2.28+, 3.3.18+, and 3.4.2+.
+The minimum recommended etcd versions to run in **production** are v3.4.8+ and v3.5.4+.
```
ahrtr, line 6: I am thinking probably it makes more sense just to remove this in CHANGELOG-4.0, and keep the  recommendation in one place `CHANGELOG/README.md`

What do you think? @spzala

spzala: @ahrtr yes, good point! I am not sure why didn't we remove it when modified other files but probably it was missed. Thanks!

ahrtr (approved):
LGTM

Thanks @spzala

PR #14317, opened 2022-08-05
clientv3: fix --insecure-skip-tls-verify not working when not specify…

QianChenglong:
… --cert-path and --key-path

Fixes 14315

Signed-off-by: QianChenglong <qian_cheng_long@163.com>

### DIFF
--- a/client/v3/config.go
+++ b/client/v3/config.go
@@ -19,9 +19,10 @@ import (
 	"crypto/tls"
 	"time"
 
-	"go.etcd.io/etcd/client/pkg/v3/transport"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
+
+	"go.etcd.io/etcd/client/pkg/v3/transport"
 )
 
 type Config struct {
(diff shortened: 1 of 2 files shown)


client/v3/config.go
```diff
@@ -166,18 +167,11 @@ func newTLSConfig(scfg *SecureConfig, lg *zap.Logger) (*tls.Config, error) {
 		}
 	}
 
-	// If key/cert is not given but user wants secure connection, we
-	// should still setup an empty tls configuration for gRPC to setup
-	// secure connection.
-	if tlsCfg == nil && !scfg.InsecureTransport {
+	if tlsCfg == nil {
 		tlsCfg = &tls.Config{}
```
ahrtr, line 171: The PR is breaking the following case:
If the etcd is listening an insecure port, such as 
```
etcd --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 
```

The etcdctl still needs to set `--insecure-skip-tls-verify`, otherwise it runs into issue something like below.  Obviously it doesn't make sense,
```
{"level":"warn","ts":"2022-08-07T07:41:04.805+0800","logger":"etcd-client","caller":"v3/retry_interceptor.go:64","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc000472000/127.0.0.1:2379","method":"/etcdserverpb.Cluster/MemberList","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"transport: authentication handshake failed: EOF\""}
Error: context deadline exceeded
```

QianChenglong: fixed

ahrtr (approved):
LGTM

Thank you @QianChenglong

PR #14322, opened 2022-08-07
*: handle auth invalid token and old revision errors in watch

mitake:
Fix https://github.com/etcd-io/etcd/issues/12385

~It's still a WIP PR, please do not merge. Remaining todos:~
- [x] clean up code
- [x] write a test

### DIFF
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -224,16 +224,16 @@ func (ws *watchServer) Watch(stream pb.Watch_WatchServer) (err error) {
 	return err
 }
 
-func (sws *serverWatchStream) isWatchPermitted(wcr *pb.WatchCreateRequest) bool {
+func (sws *serverWatchStream) isWatchPermitted(wcr *pb.WatchCreateRequest) error {
 	authInfo, err := sws.ag.AuthInfoFromCtx(sws.gRPCStream.Context())
 	if err != nil {
-		return false
+		return err
 	}
 	if authInfo == nil {
 		// if auth is enabled, IsRangePermitted() can cause an error
 		authInfo = &auth.AuthInfo{}
 	}
-	return sws.ag.AuthStore().IsRangePermitted(authInfo, wcr.Key, wcr.RangeEnd) == nil
+	return sws.ag.AuthStore().IsRangePermitted(authInfo, wcr.Key, wcr.RangeEnd)
 }
 
 func (sws *serverWatchStream) recvLoop() error {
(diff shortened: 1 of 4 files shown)


ahrtr:
Thanks @mitake , just marked this PR as draft. Please feel free to mark it as ready for review once it's ready.

kafuu-chino:
I clone locally, tried a run and the problem was solved, the previous watch will also be send by `nextResume`. 

However, because the default `WatchID = 0` is returned by error, this will cause` WatchID = 0` to be deleted once in `ids`, although it will be recreated later. See https://github.com/etcd-io/etcd/pull/14296.

I also read the committed code, instead of using a string, would try passing an error code between the client and the server? 
And why not
```go
err := sws.isWatchPermitted(creq)
if err != nil {
```

mitake:
Thanks for trying this PR @kafuu-chino . 
> I also read the committed code, instead of using a string, would try passing an error code between the client and the server?

It might be simpler but the error codes don't have unique integer IDs. So keeping the type of `cancelReason` as string is good.

I'll also change the return value type of isWatchPermitted().

mitake:
@ahrtr I think this PR is ready to be reviewed, could you check when you have time?

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -224,16 +224,17 @@ func (ws *watchServer) Watch(stream pb.Watch_WatchServer) (err error) {
...
-		return false
+		return err
 	}
 	if authInfo == nil {
 		// if auth is enabled, IsRangePermitted() can cause an error
 		authInfo = &auth.AuthInfo{}
 	}
-	return sws.ag.AuthStore().IsRangePermitted(authInfo, wcr.Key, wcr.RangeEnd) == nil
+	err = sws.ag.AuthStore().IsRangePermitted(authInfo, wcr.Key, wcr.RangeEnd)
+	return err
```
ahrtr, line 237:
```suggestion
	return sws.ag.AuthStore().IsRangePermitted(authInfo, wcr.Key, wcr.RangeEnd)
```

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -267,13 +268,27 @@ func (sws *serverWatchStream) recvLoop() error {
...
+			if err != nil {
+				var cancelReason string
+				switch err {
+				case auth.ErrInvalidAuthToken:
+					cancelReason = rpctypes.ErrGRPCInvalidAuthToken.Error()
+				case auth.ErrAuthOldRevision:
+					cancelReason = rpctypes.ErrGRPCAuthOldRevision.Error()
+				default:
+					if err != auth.ErrPermissionDenied {
+						sws.lg.Panic("unexpected error code", zap.Error(err))
```
ahrtr, line 281: It seems that it's also possible to get an [ErrUserEmpty](https://github.com/etcd-io/etcd/blob/1222269ebaf7ac7f9e8dca79bf1582bba1f0de74/server/auth/store.go#L849) response.

ahrtr: It might not be good to panic etcd server.  I suggest to change it to an error log instead, and return `ErrPermissionDenied` to clients.

ahrtr:
The pipeline failure should be caused by this PR. Please fix the failures.

server/etcdserver/api/v3rpc/watch.go
mitake: Thanks, I'll follow the suggestion.

server/etcdserver/api/v3rpc/watch.go
mitake: Good catch, thanks! Let me fix.

ahrtr (approved):
LGTM

Thank you @mitake

kafuu-chino (commented):
LGTM
I think it's okay and the test is fine.
Thanks for solving my problem :)

mitake:
@ahrtr @kafuu-chino Thanks for reviewing! The failed check caused by the previous version of the PR was e2e IIRC, but I couldn't reproduce the failure on my local env, I guess it was a non deterministic issue. Let me merge this branch.

mitake:
I'll open PRs for backporting the change to stable releases later.

PR #14327, opened 2022-08-10
add .DS_Store into .gitignore

ahrtr:
When accessing directories/files through `Finder` on MacOS, then `.DS_Store` may be generated automatically. We should ignore such file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ hack/tls-setup/certs
 *.tmp
 *.bak
 .gobincache/
+.DS_Store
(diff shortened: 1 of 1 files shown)


SimFG:
maybe `**/.DS_Store` should also be added for some folder and subfloder.

ahrtr:
> maybe `**/.DS_Store` should also be added for some folder and subfloder.

It's the same pattern as `.DS_Store`.

spzala (approved):
lgtm Thanks @ahrtr

ahrtr:
thx  @spzala for the quick review.

PR #14328, opened 2022-08-10
[Draft] tests: another way to add WaitLeader function to common framework

clarkfw:
Follow up on #14304

This PR demonstrates a different approach to add `WaitLeader` to common framework.

As new tests are added to common framework, they can call `framework.WaitLeader()` function, no matter they are integration tests or e2e tests.

Furthermore, it is possible to make some changes and replace `framework.WaitLeader()` function with `Cluster.WaiterLeader()` function as below. At this moment, I'm not sure if it makes sense to do so.

```go
type Cluster interface {
	Members() []Member
        // HERE
	WaitLeader() int
	Client() Client
	Close() error
}

type Member interface {
	Client() Client
	Start() error
	Stop()
}

type Client interface {
	Put(key, value string, opts config.PutOptions) error
	Get(key string, opts config.GetOptions) (*clientv3.GetResponse, error)
	Delete(key string, opts config.DeleteOptions) (*clientv3.DeleteResponse, error)
	Compact(rev int64, opts config.CompactOption) (*clientv3.CompactResponse, error)
	Status() ([]*clientv3.StatusResponse, error)
	HashKV(rev int64) ([]*clientv3.HashKVResponse, error)
	Health() error
	Defragment(opts config.DefragOption) error
	AlarmList() (*clientv3.AlarmResponse, error)
	AlarmDisarm(alarmMember *clientv3.AlarmMember) (*clientv3.AlarmResponse, error)

```

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -14,6 +14,15 @@
 
 package framework
 
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
+)
+
 var (
 	// UnitTestRunner only runs in `--short` mode, will fail otherwise. Attempts in cluster creation will result in tests being skipped.
 	UnitTestRunner testRunner = unitRunner{}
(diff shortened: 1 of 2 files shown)


lavacat:
Nit: mark this PR as draft

lavacat:
I think we need to clarify what's the end goal.
Is it to replace [existing](https://github.com/etcd-io/etcd/blob/main/tests/framework/integration/cluster.go#L403) implementation of WaitLeader in framework/intergration/cluster.go with this one? 

Assuming that's the goal, then one runs into circular imports when trying to call this new helper (part of 'framework' package) from framework/intergration/cluster.go (part of 'integration'), because 'framework' imports 'integration' [here](https://github.com/etcd-io/etcd/blob/main/tests/framework/integration.go#L31)
There is probably a way to get this working, but it might involve changing existing packages/interfaces, so might be not worth the effort.

IMO, your original PR is good. Let's wait for that to get merged.

tests/framework/framework.go
```diff
@@ -22,3 +31,89 @@ var (
 	// IntegrationTestRunner runs etcdserver.EtcdServer in separate goroutine and uses client libraries to communicate.
 	IntegrationTestRunner = integrationRunner{}
 )
+
+const TickDuration = 10 * time.Millisecond
+
+// WaitLeader returns index of the member in c.Members() that is leader
+// or fails the test (if members don't agree on a legal leader in 30s).
+func WaitLeader(t testing.TB, c Cluster) int {
```
ahrtr, line 39: This  is against the existing design, it is supposed to be method in the `Cluster` interface instead of a function.

ahrtr:
In general,  it's a good point to use the [Status](https://github.com/etcd-io/etcd/blob/cb5f358b5fec337831373e2d3bd9b25478866e09/tests/framework/interface.go#L48) in both the integration and e2e test to implement the `WaitLeader` method. The benefit is that we can unify the implementation and accordingly make the code cleaner. 

But the bad side is to unnecessarily complicate the integration test. For e2e test, it has to perform a gRPC communication to the etcdserver to get the status and eventually find the leader; but for integration test, it can just get the leader from each member directly and no need to perform a gRPC call. 

I have no strong opinion on which approach we should use, but I'd prefer to keep it as it's for now. So let's keep the solution in the original PR https://github.com/etcd-io/etcd/pull/14304 for now.

tests/framework/framework.go
clarkfw: Yeah, this is the bad side. Removing `WaitLeader` from `Cluster` interface would make it a little counterintuitive.

clarkfw:
> 

sg. Let's the keep original PR. I'm closing this PR.

PR #14331, opened 2022-08-10
common tests framework: cluster client creation could fail with invalid auth

chaochn47:
common tests framework: cluster client creation with auth could fail

related to #14041

> More than 2K line of code change is too big, could you try to break down this PR into small ones?

This PR has 227 additions and 73 deletions. 

Signed-off-by: Chao Chen <chaochn@amazon.com>

---

In order to fail the test cluster client creation with invalid auth, I have to change the interface function signature.

### DIFF
--- a/tests/framework/e2e.go
+++ b/tests/framework/e2e.go
@@ -22,8 +22,11 @@ import (
 	"time"
 
 	"go.etcd.io/etcd/client/pkg/v3/testutil"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/integration"
+	"google.golang.org/grpc"
 )
 
 type e2eRunner struct{}
@@ -47,6 +50,7 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, cfg config.Clus
 		ClusterSize:                cfg.ClusterSize,
 		QuotaBackendBytes:          cfg.QuotaBackendBytes,
 		DisableStrictReconfigCheck: cfg.DisableStrictReconfigCheck,
+		AuthTokenOpts:              cfg.AuthToken,
 		SnapshotCount:              cfg.SnapshotCount,
 	}
 	switch cfg.ClientTLS {
@@ -78,15 +82,31 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, cfg config.Clus
 	if err != nil {
 		t.Fatalf("could not start etcd integrationCluster: %s", err)
 	}
-	return &e2eCluster{*epc}
+	return &e2eCluster{t, *epc}
 }
 
 type e2eCluster struct {
+	t testing.TB
 	e2e.EtcdProcessCluster
 }
 
-func (c *e2eCluster) Client() Client {
-	return e2eClient{e2e.NewEtcdctl(c.Cfg, c.EndpointsV3())}
+func (c *e2eCluster) Client(cfg clientv3.AuthConfig) (Client, error) {
+	etcdctl := e2e.NewEtcdctl(c.Cfg, c.EndpointsV3())
+	if !cfg.Empty() {
+		// use integration test client to validate if permissions are authorized
+		_, err := integration.NewClient(c.t, clientv3.Config{
+			Endpoints:   c.EndpointsV3(),
+			DialTimeout: 5 * time.Second,
+			DialOptions: []grpc.DialOption{grpc.WithBlock()},
+			Username:    cfg.Username,
+			Password:    cfg.Password,
+		})
+		if err != nil {
+			return nil, err
+		}
+		etcdctl = etcdctl.WithAuth(cfg.Username, cfg.Password)
+	}
+	return e2eClient{etcdctl}, nil
 }
 
 func (c *e2eCluster) Members() (ms []Member) {
@@ -107,7 +127,7 @@ func (c *e2eCluster) WaitLeader(t testing.TB) int {
 // WaitMembersForLeader waits until given members agree on the same leader,
 // and returns its 'index' in the 'membs' list
 func (c *e2eCluster) WaitMembersForLeader(ctx context.Context, t testing.TB, membs []Member) int {
-	cc := c.Client()
+	cc := MustClient(c.Client(clientv3.AuthConfig{}))
 
 	// ensure leader is up via linearizable get
 	for {
--- a/tests/framework/e2e/etcdctl.go
+++ b/tests/framework/e2e/etcdctl.go
@@ -31,6 +31,8 @@ import (
 type EtcdctlV3 struct {
 	cfg       *EtcdProcessClusterConfig
 	endpoints []string
+	userName  string
+	password  string
 }
 
 func NewEtcdctl(cfg *EtcdProcessClusterConfig, endpoints []string) *EtcdctlV3 {
@@ -38,6 +40,12 @@ func NewEtcdctl(cfg *EtcdProcessClusterConfig, endpoints []string) *EtcdctlV3 {
 		cfg:       cfg,
 		endpoints: endpoints,
 	}
+}
+
+func (ctl *EtcdctlV3) WithAuth(userName, password string) *EtcdctlV3 {
+	ctl.userName = userName
+	ctl.password = password
+	return ctl
 }
 
 func (ctl *EtcdctlV3) DowngradeEnable(ctx context.Context, version string) error {
@@ -234,6 +242,7 @@ func AddTxnResponse(resp *clientv3.TxnResponse, jsonData string) {
 		}
 	}
 }
+
 func (ctl *EtcdctlV3) MemberList(ctx context.Context) (*clientv3.MemberListResponse, error) {
 	var resp clientv3.MemberListResponse
 	err := ctl.spawnJsonCmd(ctx, &resp, "member", "list")
@@ -283,6 +292,9 @@ func (ctl *EtcdctlV3) flags() map[string]string {
 		}
 	}
 	fmap["endpoints"] = strings.Join(ctl.endpoints, ",")
+	if ctl.userName != "" && ctl.password != "" {
+		fmap["user"] = ctl.userName + ":" + ctl.password
+	}
 	return fmap
 }
 
@@ -453,6 +465,24 @@ func (ctl *EtcdctlV3) AlarmDisarm(ctx context.Context, _ *clientv3.AlarmMember) 
 	return &resp, err
 }
 
+func (ctl *EtcdctlV3) AuthEnable(ctx context.Context) (*clientv3.AuthEnableResponse, error) {
+	var resp clientv3.AuthEnableResponse
+	err := ctl.spawnJsonCmd(ctx, &resp, "auth", "enable")
+	return &resp, err
+}
+
+func (ctl *EtcdctlV3) AuthDisable(ctx context.Context) (*clientv3.AuthDisableResponse, error) {
+	var resp clientv3.AuthDisableResponse
+	err := ctl.spawnJsonCmd(ctx, &resp, "auth", "disable")
+	return &resp, err
+}
+
+func (ctl *EtcdctlV3) AuthStatus(ctx context.Context) (*clientv3.AuthStatusResponse, error) {
+	var resp clientv3.AuthStatusResponse
+	err := ctl.spawnJsonCmd(ctx, &resp, "auth", "status")
+	return &resp, err
+}
+
 func (ctl *EtcdctlV3) UserAdd(ctx context.Context, name, password string, opts config.UserAddOptions) (*clientv3.AuthUserAddResponse, error) {
 	args := ctl.cmdArgs()
 	args = append(args, "user", "add")
@@ -492,6 +522,12 @@ func (ctl *EtcdctlV3) UserAdd(ctx context.Context, name, password string, opts c
 	return &resp, err
 }
 
+func (ctl *EtcdctlV3) UserGet(ctx context.Context, name string) (*clientv3.AuthUserGetResponse, error) {
+	var resp clientv3.AuthUserGetResponse
+	err := ctl.spawnJsonCmd(ctx, &resp, "user", "get", name)
+	return &resp, err
+}
+
 func (ctl *EtcdctlV3) UserList(ctx context.Context) (*clientv3.AuthUserListResponse, error) {
 	var resp clientv3.AuthUserListResponse
 	err := ctl.spawnJsonCmd(ctx, &resp, "user", "list")
@@ -519,6 +555,18 @@ func (ctl *EtcdctlV3) UserChangePass(ctx context.Context, user, newPass string) 
 
 	_, err = cmd.ExpectWithContext(ctx, "Password updated")
 	return err
+}
+
+func (ctl *EtcdctlV3) UserGrantRole(ctx context.Context, user string, role string) (*clientv3.AuthUserGrantRoleResponse, error) {
+	var resp clientv3.AuthUserGrantRoleResponse
+	err := ctl.spawnJsonCmd(ctx, &resp, "user", "grant-role", user, role)
+	return &resp, err
+}
+
+func (ctl *EtcdctlV3) UserRevokeRole(ctx context.Context, user string, role string) (*clientv3.AuthUserRevokeRoleResponse, error) {
+	var resp clientv3.AuthUserRevokeRoleResponse
+	err := ctl.spawnJsonCmd(ctx, &resp, "user", "revoke-role", user, role)
+	return &resp, err
 }
 
 func (ctl *EtcdctlV3) RoleAdd(ctx context.Context, name string) (*clientv3.AuthRoleAddResponse, error) {
--- a/tests/framework/integration.go
+++ b/tests/framework/integration.go
@@ -48,6 +48,7 @@ func (e integrationRunner) NewCluster(ctx context.Context, t testing.TB, cfg con
 		Size:                       cfg.ClusterSize,
 		QuotaBackendBytes:          cfg.QuotaBackendBytes,
 		DisableStrictReconfigCheck: cfg.DisableStrictReconfigCheck,
+		AuthToken:                  cfg.AuthToken,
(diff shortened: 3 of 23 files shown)


chaochn47:
Does it look better or I can give another shot to split the PR? @ahrtr

tests/framework/e2e/etcdctl.go
```diff
@@ -435,6 +459,20 @@ func (ctl *EtcdctlV3) AlarmDisarm(_ *clientv3.AlarmMember) (*clientv3.AlarmRespo
 	return &resp, err
 }
 
+func (ctl *EtcdctlV3) AuthEnable() (*clientv3.AuthEnableResponse, error) {
+	return &clientv3.AuthEnableResponse{}, SpawnWithExpect(ctl.cmdArgs("auth", "enable"), "Authentication Enabled")
```
ahrtr, line 463: It doesn't make sense to return a empty `&clientv3.AuthEnableResponse{}` response. You have two choice:
1. Only return an error for the method;
2. Populate the `&clientv3.AuthEnableResponse{}` using `spawnJsonCmd`

tests/framework/config/cluster.go
```diff
@@ -27,4 +27,6 @@ type ClusterConfig struct {
 	PeerTLS           TLSConfig
 	ClientTLS         TLSConfig
 	QuotaBackendBytes int64
+	AuthToken         string
+	SnapshotCount     uint64
```
ahrtr, line 31: `SnapshotCount` unrelated change?

tests/framework/config/cluster.go
```diff
@@ -27,4 +27,6 @@ type ClusterConfig struct {
 	PeerTLS           TLSConfig
 	ClientTLS         TLSConfig
 	QuotaBackendBytes int64
+	AuthToken         string
```
ahrtr, line 30: I assume you will update some test cases to use this field in separate PR(s). Please let me know if it isn't the case.

chaochn47: yeah, this field will be used in real auth tests.

tests/framework/config/cluster.go
chaochn47: This field will be used in real auth tests to trigger frequent snapshot to validate consistency.

tests/framework/e2e/etcdctl.go
chaochn47: Good call. Will go with option 1

tests/framework/config/client.go
```diff
@@ -20,6 +20,15 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
+type ClientOption struct {
+	UserName string
+	Password string
+}
```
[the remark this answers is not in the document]
chaochn47: Nice. I am not aware of that. Will do

tests/framework/config/client.go
```diff
@@ -20,6 +20,15 @@ import (
...
+
+func (opt ClientOption) Empty() bool {
+	return opt.UserName == "" && opt.Password == ""
+}
```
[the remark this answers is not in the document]
chaochn47: Will do

chaochn47:
mixin test failure is unrelated. There is no monitoring changes. 

All the comments are addressed. PTAL, thx! @ahrtr

tests/framework/config/client.go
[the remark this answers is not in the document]
ahrtr: same as above

tests/framework/config/client.go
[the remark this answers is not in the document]
ahrtr: Why do you still keep `ClientOption`? My understanding is it should be removed, because it's duplicate to `AuthConfig`.

tests/framework/integration.go
```diff
@@ -287,12 +308,28 @@ func (c integrationClient) LeaseRevoke(id clientv3.LeaseID) (*clientv3.LeaseRevo
 	return c.Client.Revoke(c.ctx, id)
 }
 
+func (c integrationClient) AuthEnable() (*clientv3.AuthEnableResponse, error) {
+	return c.Client.AuthEnable(context.Background())
```
serathius, line 312:
```suggestion
	return c.Client.AuthEnable(c.ctx)
```

tests/framework/integration.go
```diff
@@ -287,12 +308,28 @@ func (c integrationClient) LeaseRevoke(id clientv3.LeaseID) (*clientv3.LeaseRevo
...
+}
+
+func (c integrationClient) AuthDisable() (*clientv3.AuthDisableResponse, error) {
+	return c.Client.AuthDisable(context.Background())
```
serathius, line 316: ditto

tests/framework/integration.go
```diff
@@ -287,12 +308,28 @@ func (c integrationClient) LeaseRevoke(id clientv3.LeaseID) (*clientv3.LeaseRevo
...
+}
+
+func (c integrationClient) AuthStatus() (*clientv3.AuthStatusResponse, error) {
+	return c.Client.AuthStatus(context.Background())
```
serathius, line 320: ditto

tests/framework/integration.go
```diff
@@ -287,12 +308,28 @@ func (c integrationClient) LeaseRevoke(id clientv3.LeaseID) (*clientv3.LeaseRevo
...
+}
+
 func (c integrationClient) UserAdd(name, password string, opts config.UserAddOptions) (*clientv3.AuthUserAddResponse, error) {
 	return c.Client.UserAddWithOptions(c.ctx, name, password, &clientv3.UserAddOptions{
 		NoPassword: opts.NoPassword,
 	})
 }
 
+func (c integrationClient) UserGet(name string) (*clientv3.AuthUserGetResponse, error) {
+	return c.Client.UserGet(context.Background(), name)
```
serathius, line 330: ditto

tests/framework/integration.go
```diff
@@ -306,6 +343,14 @@ func (c integrationClient) UserChangePass(user, newPass string) error {
 	return err
 }
 
+func (c integrationClient) UserGrantRole(user string, role string) (*clientv3.AuthUserGrantRoleResponse, error) {
+	return c.Client.UserGrantRole(context.Background(), user, role)
```
serathius, line 347: ditto

tests/framework/integration.go
```diff
@@ -306,6 +343,14 @@ func (c integrationClient) UserChangePass(user, newPass string) error {
...
+}
+
+func (c integrationClient) UserRevokeRole(user string, role string) (*clientv3.AuthUserRevokeRoleResponse, error) {
+	return c.Client.UserRevokeRole(context.Background(), user, role)
```
serathius, line 351: ditto

tests/framework/interface.go
```diff
@@ -30,7 +30,8 @@ type testRunner interface {
 
 type Cluster interface {
 	Members() []Member
-	Client() Client
+	Client(cfg clientv3.AuthConfig) (Client, error)
+	MustClient(cfg clientv3.AuthConfig) Client
```
serathius, line 34: There is no need to have a separate method `MustClient` for interface. We can keep the interface simple by having a normal function.

tests/framework/integration/cluster.go
```diff
@@ -1410,32 +1410,41 @@ func (c *Cluster) Endpoints() []string {
...
-		if c.Cfg.ClientTLS != nil {
-			tls, err := c.Cfg.ClientTLS.ClientConfig()
-			if err != nil {
-				return nil, err
-			}
-			cfg.TLS = tls
-		}
-		c.clusterClient, err = newClientV3(cfg)
+func (c *Cluster) ClusterClient(opts ...func(*clientv3.Config)) (client *clientv3.Client, err error) {
+	if c.clusterClient != nil && len(opts) == 0 {
```
serathius, line 1414: Adding ability to create a client from `opts` breaks assumptions needed to cache the client on `c.clusterClient`.

I think the original goal of caching client was removing need for user to call `.Close()` on client as this can be handled by Cluster.

If we allowing arguments for clients, I would prefer to remove caching altogether. Problem with closing client can be nicely done by adding it to `testing.T.Cleanup` as you nicely used in this PR above.

tests/framework/e2e/etcdctl.go
```diff
@@ -39,6 +43,23 @@ func NewEtcdctl(cfg *EtcdProcessClusterConfig, endpoints []string) *EtcdctlV3 {
 	}
 }
 
+func (ctl *EtcdctlV3) WithAuth(userName, password string) (*EtcdctlV3, error) {
```
serathius, line 46: I would prefer to avoid mixing binary and library client in the same module. This hides fact how client interface is implemented, which is important for testing. We want to explicitly know that e2e tests use etcdctl and integration tests are using library. 

The tests you are working on are an exemption to the rule. They want to use client in e2e test. This is ok, however it should be explicit.

chaochn47:
This PR is accidentally closed because I pushed to the same commit of main to the remote branch. How to re-open it?

chaochn47:
This PR is ready to review. Would you mind take a look again? @ahrtr @serathius Thanks!!

tests/framework/config/client.go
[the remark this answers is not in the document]
chaochn47: Will remove, thx!

tests/framework/integration.go
chaochn47: Done in the next PR revision

tests/framework/integration.go
chaochn47: Done in the next PR revision

tests/framework/integration.go
chaochn47: Done in the next PR revision

tests/framework/integration.go
chaochn47: Done in the next PR revision

tests/framework/interface.go
chaochn47: Done in the next PR revision

tests/framework/integration/cluster.go
chaochn47: Done in the next PR revision

tests/framework/e2e/etcdctl.go
chaochn47: Done in the next PR revision

chaochn47:
Ping again, do you think we still need this auth test migration? @serathius @ahrtr

tests/framework/integration.go
```diff
@@ -261,17 +272,85 @@ func (c integrationClient) TimeToLive(ctx context.Context, id clientv3.LeaseID,
 	return c.Client.TimeToLive(ctx, id, leaseOpts...)
 }
 
+func (c integrationClient) LeaseList(ctx context.Context) (*clientv3.LeaseLeasesResponse, error) {
```
serathius, line 275: Any reason you re-declare those functions?

tests/framework/integration.go
```diff
@@ -315,6 +394,18 @@ func getOps(ss []string) ([]clientv3.Op, error) {
 	return ops, nil
 }
 
+func (c integrationClient) MemberAdd(ctx context.Context, _ string, peerAddrs []string) (*clientv3.MemberAddResponse, error) {
```
serathius, line 397: Any reason you reorder the functions here? It only obscures git history.

tests/framework/integration/cluster.go
```diff
@@ -178,8 +178,7 @@ type Cluster struct {
 	Members       []*Member
 	LastMemberNum int
 
-	mu            sync.Mutex
-	clusterClient *clientv3.Client
+	mu sync.Mutex
```
serathius, line 181: Is the mutex still used? Can we remove it?

tests/framework/integration.go
chaochn47: It was added before Leases related PR is merged in upstream. I will rebase on latest code changes and send a new revision again. Thanks for pointing it out!!

tests/framework/integration.go
chaochn47: Oh, I wanted it to be in the same order of https://github.com/etcd-io/etcd/blob/cd9764a99fab81788d6234d3524a08c9b893c3e4/tests/framework/interface.go#L72-L79

tests/framework/integration/cluster.go
```diff
@@ -1410,32 +1405,33 @@ func (c *Cluster) Endpoints() []string {
...
-		}
-		if c.Cfg.ClientTLS != nil {
-			tls, err := c.Cfg.ClientTLS.ClientConfig()
-			if err != nil {
-				return nil, err
-			}
-			cfg.TLS = tls
-		}
-		c.clusterClient, err = newClientV3(cfg)
+func (c *Cluster) ClusterClient(opts ...func(*clientv3.Config)) (client *clientv3.Client, err error) {
```
serathius, line 1408: I don't think it's good to have inconsistency between `ClusterClient` and `Client` methods. Could update the integration. Cluster code so it also doesn't require user to call `Close`.

tests/framework/integration/cluster.go
chaochn47: It is stilled used when mutating `Members []*Member`. https://github.com/etcd-io/etcd/blob/cd9764a99fab81788d6234d3524a08c9b893c3e4/tests/framework/integration/cluster.go#L1371-L1375

Looks like I need to group it with `Members`, does it make sense to you?

tests/framework/integration/cluster.go
chaochn47: In order to achieve that, I need to register the test clean up function whenever `ClusterClient` is called like this

```
	t.Cleanup(func() {
		cc.Close()
	})
```

However, `testing.TB` is not stored in the https://github.com/etcd-io/etcd/blob/cd9764a99fab81788d6234d3524a08c9b893c3e4/tests/framework/integration/cluster.go#L177-L184

Do you think one of the implementation can be adding this to the Cluster struct? Thanks!

serathius:
For faster review I would recommend removing unrelated changes from the PR. It's over 300 lines of code, but still includes changes that muddy the water. Best if PRs do one thing and only this one thing.

chaochn47:
> For faster review I would recommend removing unrelated changes from the PR. It's over 300 lines of code, but still includes changes that muddy the water. Best if PRs do one thing and only this one thing.

tests/framework/integration.go
serathius: Please separate logic changes from code cleanups, either separate commits or PRs.

tests/framework/integration/cluster.go
serathius: Hmm, usually the mutex declaration is above the variables it protects so it's a little unnatural. Let's keep it as it is and revisit it in future cleanup.

tests/framework/integration/cluster.go
serathius: I think it's ok to add `testing.TB` to integration.Cluster.

tests/framework/integration.go
chaochn47: Make sense. Will do

tests/framework/e2e.go
```diff
@@ -78,15 +82,31 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, cfg config.Clus
...
+		_, err := integration.NewClient(c.t, clientv3.Config{
+			Endpoints:   c.EndpointsV3(),
+			DialTimeout: 5 * time.Second,
+			DialOptions: []grpc.DialOption{grpc.WithBlock()},
+			Username:    cfg.Username,
+			Password:    cfg.Password,
+		})
+		if err != nil {
+			return nil, err
+		}
```
ahrtr, line 106: Probably we do not need this. Reasons:
1. It seems not good for e2e test to depends on integration test;
2. The related test cases will get a denied response if the auth info isn't correct. So the pre-check isn't necessary.

tests/framework/interface.go
```diff
@@ -30,7 +30,7 @@ type testRunner interface {
 
 type Cluster interface {
 	Members() []Member
-	Client() Client
+	Client(cfg clientv3.AuthConfig) (Client, error)
```
ahrtr, line 33: I am thinking whether it's possible to get rid of the parameter `cfg clientv3.AuthConfig`. When the test cases setup the user & role, and enable the auth, then the client should record the authConfig automatically. 

Does this make sense?

ahrtr:
Please also rebase this PR although there is no conflict.

tests/framework/e2e.go
chaochn47: I agree it's not optimal that e2e test depending on integration test. This is an exception when auth config is not empty. 

Copying from past conversation between @serathius and me. 

From me
> Some existing auth tests checks unauthenticated errors. 

> Integration test will throw an error at client creation stage with `clientv3.New` https://github.com/etcd-io/etcd/blob/7ec336f80c9463fecd6d8ef15a9f05f3e8f6287a/client/v3/client.go#L439

> However, e2e test won't fail at client creation stage and fail at "got response" stage instead. 

> Also every request from e2e test will establish a new grpc connection to the server but integration test reuses existing one. In order to use a different user name and password, integration test needs to create another client. 

From Marek
> I think we can create `clientv3.Client` just for this.

ref. https://github.com/etcd-io/etcd/pull/14041#discussion_r874629440

Does this makes sense to you?

tests/framework/interface.go
chaochn47: I don't think client should be stateful. In general, it makes the client not concurrent safe. If multiple callers set up the user & role and enable the auth using the client, this would cause contention.

ahrtr: ack. 

It's just a little strange to provide a specific` cfg clientv3.AuthConfig` parameter, but automatically get all other parameters from the cluster.  Probably it makes more sense to resolve this using a generic way, such as change the parameter to something like `opts ...Option`.  It might be a little over engineering from another perspective. 

Let's keep it as it's for now, and probably think about it separately.

ahrtr (approved):
Overall looks good to me. 

Thank you @chaochn47

tests/framework/e2e.go
serathius: > It seems not good for e2e test to depends on integration test;

It's not tests, it's e2e tests using integration client, which is just a wrapper for clientv3 library. There is nothing wrong with e2e tests using client library for some things that are not surfaced in etcdctl. To complete migration of tests to new framework this is even required as Some tests cannot be expressed using just e2e or integration tools

> I don't think e2e and integration test must follow the same pattern or behavior

But it would make testing and reasoning about them much simpler.

serathius (approved):

PR #14344, opened 2022-08-14
tests: e2e and integration share TickDuration constant

clarkfw:
As mentioned in https://github.com/etcd-io/etcd/pull/14304#discussion_r945192946, it would be a good idea to share the same `TickDuration` constant between `integration/cluster.go` and `e2e.go`, for the benefit of less duplication.

### DIFF
--- a/tests/framework/config/cluster.go
+++ b/tests/framework/config/cluster.go
@@ -14,12 +14,16 @@
 
 package config
 
+import "time"
+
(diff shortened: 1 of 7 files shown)


ahrtr (approved):
LGTM 

Thank you @clarkfw

spzala (approved):
lgtm
Thanks @clarkfw

PR #14345, opened 2022-08-15
tests: Migrate watch test to common framework

nic-chen:
Part of https://github.com/etcd-io/etcd/issues/13637

### DIFF
--- a/tests/common/watch_test.go
+++ b/tests/common/watch_test.go
@@ -1,0 +1,82 @@
+package common
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+func TestWatch(t *testing.T) {
+	testRunner.BeforeTest(t)
+	watchTimeout := 1 * time.Second
+	for _, tc := range clusterTestCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+			defer cancel()
+			clus := testRunner.NewCluster(ctx, t, tc.config)
+
+			defer clus.Close()
+			cc := clus.Client()
+			testutils.ExecuteUntil(ctx, t, func() {
+				tests := []struct {
+					puts     []testutils.KV
+					watchKey string
+					opts     config.WatchOptions
+					wanted   []testutils.KV
+				}{
+					{ // watch by revision
+						puts:     []testutils.KV{{Key: "bar", Val: "revision_1"}, {Key: "bar", Val: "revision_2"}, {Key: "bar", Val: "revision_3"}},
+						watchKey: "bar",
+						opts:     config.WatchOptions{Revision: 3},
+						wanted:   []testutils.KV{{Key: "bar", Val: "revision_2"}, {Key: "bar", Val: "revision_3"}},
+					},
+					{ // watch 1 key
+						puts:     []testutils.KV{{Key: "sample", Val: "value"}},
+						watchKey: "sample",
+						opts:     config.WatchOptions{Revision: 1},
+						wanted:   []testutils.KV{{Key: "sample", Val: "value"}},
+					},
+					{ // watch 3 keys by prefix
+						puts:     []testutils.KV{{Key: "foo1", Val: "val1"}, {Key: "foo2", Val: "val2"}, {Key: "foo3", Val: "val3"}},
+						watchKey: "foo",
+						opts:     config.WatchOptions{Revision: 1, Prefix: true},
+						wanted:   []testutils.KV{{Key: "foo1", Val: "val1"}, {Key: "foo2", Val: "val2"}, {Key: "foo3", Val: "val3"}},
+					},
+					{ // watch 3 keys by range
+						puts:     []testutils.KV{{Key: "key1", Val: "val1"}, {Key: "key3", Val: "val3"}, {Key: "key2", Val: "val2"}},
+						watchKey: "key",
+						opts:     config.WatchOptions{Revision: 1, RangeEnd: "key3"},
+						wanted:   []testutils.KV{{Key: "key1", Val: "val1"}, {Key: "key2", Val: "val2"}},
+					},
+				}
+
+				for _, tt := range tests {
+					wCtx, wCancel := context.WithCancel(ctx)
+					wch := cc.Watch(wCtx, tt.watchKey, tt.opts)
+					if wch == nil {
+						t.Fatalf("failed to watch %s", tt.watchKey)
+					}
+
+					for j := range tt.puts {
+						if err := cc.Put(tt.puts[j].Key, tt.puts[j].Val, config.PutOptions{}); err != nil {
+							t.Fatalf("can't not put key %q, err: %s", tt.puts[j].Key, err)
+						}
+					}
+
+					kvs, err := testutils.KeyValuesFromWatchChan(wch, len(tt.wanted), watchTimeout)
+					if err != nil {
+						wCancel()
+						t.Fatalf("failed to get key-values from watch channel %s", err)
+					}
+
+					wCancel()
+					assert.Equal(t, tt.wanted, kvs)
+				}
+			})
+		})
+	}
+}
--- a/tests/framework/e2e/etcdctl.go
+++ b/tests/framework/e2e/etcdctl.go
@@ -15,6 +15,7 @@
 package e2e
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -102,6 +103,7 @@ func (ctl *EtcdctlV3) Get(key string, o config.GetOptions) (*clientv3.GetRespons
 		if err != nil {
 			return nil, err
 		}
+		defer cmd.Close()
 		_, err = cmd.Expect("Count")
 		return &resp, err
 	}
@@ -145,6 +147,7 @@ func (ctl *EtcdctlV3) Txn(compares, ifSucess, ifFail []string, o config.TxnOptio
 	if err != nil {
 		return nil, err
 	}
+	defer cmd.Close()
 	_, err = cmd.Expect("compares:")
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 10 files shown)


tests/common/watch_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+func TestWatch(t *testing.T) {
+	testRunner.BeforeTest(t)
+	tcs := []struct {
```
serathius, line 16: Why use different list of cluster setups than default?

tests/common/watch_test.go
```diff
@@ -0,0 +1,132 @@
...
+					{ // watch 3 keys by range
+						puts:     []testutils.KV{{Key: "key1", Val: "val1"}, {Key: "key3", Val: "val3"}, {Key: "key2", Val: "val2"}},
+						watchKey: "key",
+						opts:     config.WatchOptions{Revision: 1, RangeEnd: "key3"},
+						wanted:   []testutils.KV{{Key: "key1", Val: "val1"}, {Key: "key2", Val: "val2"}},
+					},
+				}
+
+				for i, tt := range tests {
+					wCtx, wCancel := context.WithCancel(context.Background())
```
serathius, line 84:
```suggestion
					wCtx, wCancel := context.WithCancel(ctx)
```

tests/common/watch_test.go
```diff
@@ -0,0 +1,132 @@
...
+					wch := cc.Watch(wCtx, tt.watchKey, tt.opts)
+					if wch == nil {
+						t.Fatalf("failed to watch %s", tt.watchKey)
+					}
+
+					errs := make(chan error, 1)
+					go func(i int, puts []testutils.KV) {
```
serathius, line 91: Does this need to be in separate goroutine? With open watch you should be able to do puts synchronously.

tests/common/watch_test.go
```diff
@@ -0,0 +1,132 @@
...
+						defer close(errs)
+						for j := range puts {
+							if err := cc.Put(puts[j].Key, puts[j].Val, config.PutOptions{}); err != nil {
+								errs <- fmt.Errorf("can't not put key %q, err: %s", puts[j].Key, err)
+								break
+							}
+						}
+						errs <- nil
+						// sleep some time for watch
+						time.Sleep(100 * time.Millisecond)
```
serathius, line 101: Is this sleep needed?

serathius: Please update comment to explain why.

tests/common/watch_test.go
```diff
@@ -0,0 +1,132 @@
...
+					}(i, tt.puts)
+
+					err := <-errs
+					if err != nil {
+						t.Fatal(err)
+					}
+
+					var kvs []testutils.KV
+				ForLoop:
```
serathius, line 110: Move this whole loop that reads KV to separate function.

tests/framework/e2e/etcdctl.go
```diff
@@ -545,9 +554,73 @@ func (ctl *EtcdctlV3) spawnJsonCmd(output interface{}, args ...string) error {
...
+				lines := proc.Lines()
+				for _, line := range lines {
+					var resp clientv3.WatchResponse
+					json.Unmarshal([]byte(line), &resp)
+					if resp.Canceled {
+						close(ch)
+						return
+					}
+					if len(resp.Events) > 0 {
+						// if the mod revision for the event is smaller than the current mod revision,
```
serathius, line 602: This looks suspicious. Why this is a problem for e2e tests and not integration tests? That would mean that etcd client library handles it but etcdctl not?

serathius:
Left some small comments, but overall looks great! Good job.

tests/common/watch_test.go
nic-chen: sure, updated.

tests/common/watch_test.go
nic-chen: fixed.

tests/common/watch_test.go
nic-chen: removed the goroutine.

tests/common/watch_test.go
nic-chen: removed.

tests/common/watch_test.go
nic-chen: done.

tests/framework/e2e/etcdctl.go
nic-chen: updated to comments. please take a look.

tests/framework/testutils/helpters.go
```diff
@@ -35,3 +38,28 @@ func KeyValuesFromGetResponse(resp *clientv3.GetResponse) (kvs []KV) {
...
+
+func KeyValuesFromWatchChan(wch clientv3.WatchChan, wantedLen int, timeout time.Duration) (kvs []KV, err error) {
+ForLoop:
+	for {
+		select {
+		case watchResp, ok := <-wch:
+			if ok {
+				kvs = append(kvs, KeyValuesFromWatchResponse(watchResp)...)
+				if len(kvs) == wantedLen {
+					break ForLoop
```
serathius, line 57:
```suggestion
					return kvs, nil
```

tests/framework/testutils/helpters.go
```diff
@@ -35,3 +38,28 @@ func KeyValuesFromGetResponse(resp *clientv3.GetResponse) (kvs []KV) {
...
+
+func KeyValuesFromWatchResponse(resp clientv3.WatchResponse) (kvs []KV) {
+	for _, event := range resp.Events {
+		kvs = append(kvs, KV{Key: string(event.Kv.Key), Val: string(event.Kv.Value)})
+	}
+	return kvs
+}
+
+func KeyValuesFromWatchChan(wch clientv3.WatchChan, wantedLen int, timeout time.Duration) (kvs []KV, err error) {
+ForLoop:
```
serathius, line 50: No need for this label if we use return instead.

tests/framework/testutils/helpters.go
nic-chen: updated.

tests/framework/testutils/helpters.go
nic-chen: removed.

tests/framework/e2e/etcdctl.go
```diff
@@ -545,9 +554,75 @@ func (ctl *EtcdctlV3) spawnJsonCmd(output interface{}, args ...string) error {
...
+				lines := proc.Lines()
+				for _, line := range lines {
+					var resp clientv3.WatchResponse
+					json.Unmarshal([]byte(line), &resp)
+					if resp.Canceled {
+						close(ch)
+						return
+					}
+					if len(resp.Events) > 0 {
+						// Since `proc.Lines()` returns all the content currently read,
```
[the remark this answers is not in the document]
nic-chen: Updated. Thanks for the suggestion. I thought about it but didn't make it happen.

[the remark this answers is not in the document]
serathius: Looks great!

serathius (approved):

tests/framework/e2e/etcdctl.go
```diff
@@ -545,9 +554,55 @@ func (ctl *EtcdctlV3) spawnJsonCmd(output interface{}, args ...string) error {
...
+	if opts.Revision != 0 {
+		args = append(args, "--rev", fmt.Sprint(opts.Revision))
+	}
+	proc, err := SpawnCmd(args, nil)
+	if err != nil {
+		return nil
+	}
+
+	ch := make(chan clientv3.WatchResponse)
+	go func() {
```
ahrtr, line 584: Have you considered to wait for the goroutine to exit when the test case finish?

nic-chen: hi, @ahrtr 
 the goroutine will exit by context cancel when the test finishes, see: https://github.com/etcd-io/etcd/pull/14345/files#diff-70546f01c62cf8d453ef453d51b64bbdc664e39613453b572120148a77600a08R76

ahrtr: It's possible (although with low possibility) that the test case finish before the goroutine exit, eventually the test might report goroutine leak errors. Our pipeline fails due to this kind of error from time to time. So for safety, we need to make sure the goroutine exits before the the test case returns/exits.

serathius: I don't think this is a problem. We should be good as long as we propagate signal to goroutine that causes it to exit. It's up to leak detection to wait for this signal to propagate. If you read the etcd test goroutine leak implementation, it gives test up to 5 seconds to cleanup it's goroutines and explicitly gives them time to execute by calling `runtime.Gosched()`.

nic-chen: I think we could merge it first. If there is a real problem, I will fix it. What do you think? @serathius @ahrtr

serathius: I think @ahrtr is worried that the problem will not just imminently show up after merging. Leak goroutine issues like this one are very subtle (for example I cannot reproduce this issue locally at all), so I understand his position.

I still think my argument holds and this should not be a problem, but I would prefer to wait for confirmation from @ahrtr

ahrtr: > If you read the etcd test goroutine leak implementation, it gives test up to 5 seconds to cleanup its goroutines and uses `runtime.Gosched()` to give goroutine time to execute and exit.

The goroutine leak checker indeed waits up to [1 second](https://github.com/etcd-io/etcd/blob/e363480089783aff0d1c26586f703d31223cef5d/client/pkg/testutil/leak.go#L117) in the `afterTest` check. 

This is just a mitigation solution instead of best practice to me. We should always make sure gracefully shutdown of all threads/goroutines in almost all cases, including production code and test code, unless the cases that we intentionally create.

Previously we also saw a situation that a test case (in test code) or the main goroutine (in production code) already finishes/exits, but a goroutine it creates still prints log, and it isn't allowed by golang. Of course, I do not see such issue in this PR.

I just searched the repo, there are lots of similar cases which create goroutines but do not gracefully shutdown them. It's OK to merge this PR for now. But I'd like to revisit all such cases afterwards in separate PR(s).  WDYT? @serathius @spzala

serathius: I agree the is a problem worth looking into.

serathius:
Merging as we agreed to follow up with separate PRs.

PR #14347, opened 2022-08-15
It is best a constant

yzcyayaya:
testutil:  It is best a constant

Because individual numbers are not constants very unfriendly

Record


spzala (commented):
@yzcyayaya thanks for the PR but since the variable is local, I think there is no need to make changes.

spzala:
@yzcyayaya thanks again, closing it considering changes are not required. Please continue helping project with work on open issues or other improvements.

PR #14348, opened 2022-08-15
embed: add integration test for distributed tracing

### LINKED ISSUE #13814: Add integration test for etcd tracing
We can follow the same way as Kubernetes did in [pull/103234](https://github.com/kubernetes/kubernetes/pull/103234). 

It's also the prerequisite to graduate/stable the experimental tracing flags. [12460](https://github.com/etcd-io/etcd/issues/12460)

VinozzZ:
To verify distributed tracing feature is correctly setup, this PR adds
an integration test for this feature.
In the process of writing the test, I discovered a goroutine leak due to
the TraceProvider not being closed. This PR fixs this issue as well

Closes #13814

### DIFF
--- a/server/embed/config_tracing.go
+++ b/server/embed/config_tracing.go
@@ -40,13 +40,19 @@ func validateTracingConfig(samplingRate int) error {
 	return nil
 }
 
-func setupTracingExporter(ctx context.Context, cfg *Config) (exporter tracesdk.SpanExporter, options []otelgrpc.Option, err error) {
-	exporter, err = otlptracegrpc.New(ctx,
+type tracingExporter struct {
+	exporter tracesdk.SpanExporter
+	opts     []otelgrpc.Option
+	shutdown func(ctx context.Context)
+}
+
+func newTracingExporter(ctx context.Context, cfg *Config) (*tracingExporter, error) {
+	exporter, err := otlptracegrpc.New(ctx,
 		otlptracegrpc.WithInsecure(),
 		otlptracegrpc.WithEndpoint(cfg.ExperimentalDistributedTracingAddress),
 	)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	res, err := resource.New(ctx,
@@ -55,7 +61,7 @@ func setupTracingExporter(ctx context.Context, cfg *Config) (exporter tracesdk.S
 		),
 	)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	if resWithIDKey := determineResourceWithIDKey(cfg.ExperimentalDistributedTracingServiceInstanceID); resWithIDKey != nil {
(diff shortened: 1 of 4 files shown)


server/embed/config_tracing.go
```diff
@@ -40,13 +40,19 @@ func validateTracingConfig(samplingRate int) error {
 	return nil
 }
 
-func setupTracingExporter(ctx context.Context, cfg *Config) (exporter tracesdk.SpanExporter, options []otelgrpc.Option, err error) {
-	exporter, err = otlptracegrpc.New(ctx,
+type tracingExporter struct {
+	exporter tracesdk.SpanExporter
+	opts     []otelgrpc.Option
+	shutdown func(ctx context.Context)
```
ahrtr, line 46: The name `shutdown` is a little strange to me, if `tracingExporter` knows what it should do on close, why not to get the `traceProvider` added to the struct directly? Two choice:
1. Add the `traceProvider` into the struct directly, and call `te.traceProvider.Shutdown(ctx)` directly on Close. 
2. Rename to `cleanup` so that the logic is agnostic to `tracingExporter`, instead it's just doing the cleanup task it's told.

server/embed/etcd.go
```diff
@@ -227,15 +227,14 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
 
 	if srvcfg.ExperimentalEnableDistributedTracing {
 		tctx := context.Background()
-		tracingExporter, opts, err := setupTracingExporter(tctx, cfg)
+		tracingExporter, err := newTracingExporter(tctx, cfg)
 		if err != nil {
 			return e, err
 		}
-		if tracingExporter == nil || len(opts) == 0 {
-			return e, fmt.Errorf("error setting up distributed tracing")
```
ahrtr, line 235: You intentionally removed this, does it mean that `tracingExporter == nil || len(opts) == 0` will never be true if the returned err is `nil`?

server/embed/config_tracing.go
VinozzZ: Thank you for the suggestion. I went with the first option

server/embed/etcd.go
VinozzZ: Yes, that's the intention. I noticed no code path would result in err == nil and either exporter nor opts would be nil

server/embed/config_tracing.go
```diff
@@ -93,7 +101,20 @@ func setupTracingExporter(ctx context.Context, cfg *Config) (exporter tracesdk.S
...
+	}, nil
+}
+
+func (te *tracingExporter) Close(ctx context.Context) {
+	if te.exporter != nil {
+		te.exporter.Shutdown(ctx)
+	}
+	if te.provider != nil {
+		te.provider.Shutdown(ctx)
+	}
```
ahrtr, line 117: I am not quite confident on this. Have you considered the order of the shutdown? In other words, is there any potential issue when shutting down the exporter firstly? I just had a quick code review on the otl SDK, it seems that the provider callbacks the exporters; if we shutdown the exporter firstly, will it cause any potential temporary issue on the provider?

Have you thought about shutting down the provider firstly? 

Please clarify this.

server/embed/etcd.go
ahrtr: ack

ptabor:
Please run `./scripts/fix.sh` to fix the `mod tidy` problems causing a Static Analysis to fail.

server/embed/config_tracing.go
VinozzZ: From my understanding, if the exporter is shutdown first, the provider will either hit a timeout or a shutdown error. It would mean that some spans that hasn't been sent to the remote may be lost. Thank you for pointing it out. I changed it to shutdown the provider first.

ahrtr:
@VinozzZ There are lots of comment changes in this PR. Did you do it intentionally or they are just updated by IDE or any tool? If it's the former (intentionally did), please do it in a separate PR. If it's the latter, please revert the changes.

VinozzZ:
> @VinozzZ There are lots of comment changes in this PR. Did you do it intentionally or they are just updated by IDE or any tool? If it's the former (intentionally did), please do it in a separate PR. If it's the latter, please revert the changes.

This is the change from running `./scripts/fix.sh` . Just curious if the script makes changes based on some configs from the local dev machine?

ahrtr (approved):
LGTM

Thank you @VinozzZ

spzala (approved):
Thanks @VinozzZ lgtm. Defer to @ahrtr

PR #14351, opened 2022-08-16
chore: log when an invalid watch request is received

spacewander:
As protobuf doesn't have required field, user may send an empty
WatchRequest by mistake. Currently, etcd will ignore the invalid request
and keep the stream opening. If we don't reject the invalid request by
closing the stream, it would be better to leave a log there.

This commit also fixes a typo in the comment.

### DIFF
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -348,7 +348,7 @@ func (sws *serverWatchStream) recvLoop() error {
 			// we probably should not shutdown the entire stream when
 			// receive an invalid command.
 			// so just do nothing instead.
-			sws.lg.Warn("invalid watch request received in gRPC stream")
+			sws.lg.Sugar().Infof("invalid watch request type %T received in gRPC stream", uv)
 			continue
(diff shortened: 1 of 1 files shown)


server/etcdserver/api/v3rpc/watch.go
```diff
@@ -346,8 +346,9 @@ func (sws *serverWatchStream) recvLoop() error {
 			}
 		default:
 			// we probably should not shutdown the entire stream when
-			// receive an valid command.
+			// receive an invalid command.
 			// so just do nothing instead.
+			sws.lg.Warn("invalid watch request received in gRPC stream")
```
SimFG, line 351: Adding the `request type` info will be more helpful to debug.

spzala, line 351: @spacewander thanks! Logging it for information purpose sounds good, and so I think the severity should be `Info` (not warning).

spacewander:
@spzala @SimFG 
Updated.

ahrtr (approved):
LGTM 

Thank you @spacewander

PR #14353, opened 2022-08-16
Improve code readability.

Comolli:
All the inflights_test code passed.

### DIFF
--- a/raft/tracker/inflights.go
+++ b/raft/tracker/inflights.go
@@ -56,11 +56,9 @@ func (in *Inflights) Add(inflight uint64) {
 	if in.Full() {
 		panic("cannot add into a Full inflights")
 	}
-	next := in.start + in.count
-	size := in.size
-	if next >= size {
-		next -= size
-	}
+
+	next := (in.start + in.count) % in.size
+
 	if next >= len(in.buffer) {
 		in.grow()
 	}


raft/tracker/inflights.go
```diff
@@ -56,11 +56,9 @@ func (in *Inflights) Add(inflight uint64) {
 	if in.Full() {
 		panic("cannot add into a Full inflights")
 	}
-	next := in.start + in.count
-	size := in.size
-	if next >= size {
-		next -= size
-	}
+
+	next := (in.start + in.count) % in.size
```
SimFG, line 60: If `in.start + in.count` is more than twice as large as `in.size`, `next` will be changed compared to before. And the `next` will decide whether to execute `grow`. Maybe some context I don't know.

spzala (commented):
@Comolli thanks but I think it's unnecessary changes to stable raft library. Please continue helping project with work on open issues or other improvements. Thanks! cc @ahrtr

spzala:
@Comolli thanks again. Closing it per discussion here. Please continue helping project with work on open issues or other improvements.

raft/tracker/inflights.go
Comolli: Thank you for your kind reply.

PR #14354, opened 2022-08-16
Added client-auto-sync-interval argument to the grpc-proxy

biosvs:
etcdmain: added client-auto-sync-interval argument to the grpc-proxy

Currently it's possible to enable auto sync in client, which connects to grpc-proxies (if --namespace is used). But grpc-proxy itself don't have an option to update endpoints list from etcd clusters.

This PR introduces an option client-auto-sync-interval for grpc-proxy, which allows to specify auto sync interval for grpc-proxy itself.

### DIFF
--- a/server/etcdmain/grpc_proxy.go
+++ b/server/etcdmain/grpc_proxy.go
@@ -57,15 +57,16 @@ import (
 )
 
 var (
-	grpcProxyListenAddr            string
-	grpcProxyMetricsListenAddr     string
-	grpcProxyEndpoints             []string
-	grpcProxyDNSCluster            string
-	grpcProxyDNSClusterServiceName string
-	grpcProxyInsecureDiscovery     bool
-	grpcProxyDataDir               string
-	grpcMaxCallSendMsgSize         int
-	grpcMaxCallRecvMsgSize         int
+	grpcProxyListenAddr                string
+	grpcProxyMetricsListenAddr         string
+	grpcProxyEndpoints                 []string
+	grpcProxyEndpointsAutoSyncInterval time.Duration
+	grpcProxyDNSCluster                string
+	grpcProxyDNSClusterServiceName     string
+	grpcProxyInsecureDiscovery         bool
+	grpcProxyDataDir                   string
+	grpcMaxCallSendMsgSize             int
+	grpcMaxCallRecvMsgSize             int
 
 	// tls for connecting to etcd
 
@@ -134,6 +135,7 @@ func newGRPCProxyStartCommand() *cobra.Command {
 	cmd.Flags().StringVar(&grpcProxyMetricsListenAddr, "metrics-addr", "", "listen for endpoint /metrics requests on an additional interface")
 	cmd.Flags().BoolVar(&grpcProxyInsecureDiscovery, "insecure-discovery", false, "accept insecure SRV records")
 	cmd.Flags().StringSliceVar(&grpcProxyEndpoints, "endpoints", []string{"127.0.0.1:2379"}, "comma separated etcd cluster endpoints")
+	cmd.Flags().DurationVar(&grpcProxyEndpointsAutoSyncInterval, "endpoints-auto-sync-interval", 0, "etcd endpoints auto sync interval (disabled by default)")
 	cmd.Flags().StringVar(&grpcProxyAdvertiseClientURL, "advertise-client-url", "127.0.0.1:23790", "advertise address to register (must be reachable by client)")
 	cmd.Flags().StringVar(&grpcProxyResolverPrefix, "resolver-prefix", "", "prefix to use for registering proxy (must be shared with other grpc-proxy members)")
 	cmd.Flags().IntVar(&grpcProxyResolverTTL, "resolver-ttl", 0, "specify TTL, in seconds, when registering proxy endpoints")
@@ -338,8 +340,9 @@ func newProxyClientCfg(lg *zap.Logger, eps []string, tls *transport.TLSInfo) (*c
 func newClientCfg(lg *zap.Logger, eps []string) (*clientv3.Config, error) {
 	// set tls if any one tls option set
 	cfg := clientv3.Config{
-		Endpoints:   eps,
-		DialTimeout: 5 * time.Second,
+		Endpoints:        eps,
+		AutoSyncInterval: grpcProxyEndpointsAutoSyncInterval,
+		DialTimeout:      5 * time.Second,
 	}
 
 	if grpcMaxCallSendMsgSize > 0 {
--- a/tests/e2e/etcd_grpcproxy_test.go
+++ b/tests/e2e/etcd_grpcproxy_test.go
@@ -1,0 +1,169 @@
+// Copyright 2017 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/pkg/v3/expect"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+func TestGrpcProxyAutoSync(t *testing.T) {
+	e2e.SkipInShortMode(t)
+
+	var (
+		node1Name      = "node1"
+		node1ClientURL = "http://localhost:12379"
+		node1PeerURL   = "http://localhost:12380"
+
+		node2Name      = "node2"
+		node2ClientURL = "http://localhost:22379"
+		node2PeerURL   = "http://localhost:22380"
+
+		proxyClientURL = "127.0.0.1:32379"
+
+		autoSyncInterval = 1 * time.Second
+	)
+
+	// Run cluster of one node
+	proc1, err := runEtcdNode(
+		node1Name, t.TempDir(),
+		node1ClientURL, node1PeerURL,
+		"new", fmt.Sprintf("%s=%s", node1Name, node1PeerURL),
+	)
+	require.NoError(t, err)
+
+	// Run grpc-proxy instance
+	proxyProc, err := e2e.SpawnCmd([]string{e2e.BinDir + "/etcd", "grpc-proxy", "start",
+		"--advertise-client-url", proxyClientURL, "--listen-addr", proxyClientURL,
+		"--endpoints", node1ClientURL,
+		"--endpoints-auto-sync-interval", autoSyncInterval.String(),
+	}, nil)
+	require.NoError(t, err)
+
+	proxyCtl := e2e.NewEtcdctl(&e2e.EtcdProcessClusterConfig{}, []string{proxyClientURL})
+	err = proxyCtl.Put("k1", "v1", config.PutOptions{})
+	require.NoError(t, err)
+
+	memberCtl := e2e.NewEtcdctl(&e2e.EtcdProcessClusterConfig{}, []string{node1ClientURL})
+	_, err = memberCtl.MemberAdd(node2Name, []string{node2PeerURL})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Run new member
+	proc2, err := runEtcdNode(
+		node2Name, t.TempDir(),
+		node2ClientURL, node2PeerURL,
+		"existing", fmt.Sprintf("%s=%s,%s=%s", node1Name, node1PeerURL, node2Name, node2PeerURL),
+	)
+	require.NoError(t, err)
+
+	// Wait for auto sync of endpoints
+	err = waitForEndpointInLog(proxyProc, node2ClientURL)
+	require.NoError(t, err)
+
+	memberList, err := memberCtl.MemberList()
+	require.NoError(t, err)
+
+	node1MemberID, err := findMemberIDByEndpoint(memberList.Members, node1ClientURL)
+	require.NoError(t, err)
+
+	// Second node could be not ready yet
+	for i := 0; i < 10; i++ {
+		_, err = memberCtl.MemberRemove(node1MemberID)
+		if err != nil && strings.Contains(err.Error(), rpctypes.ErrGRPCUnhealthy.Error()) {
+			time.Sleep(500 * time.Millisecond)
+			continue
+		}
+		break
+	}
+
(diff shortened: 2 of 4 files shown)


server/etcdmain/grpc_proxy.go
```diff
@@ -130,6 +131,7 @@ func newGRPCProxyStartCommand() *cobra.Command {
 	cmd.Flags().StringVar(&grpcProxyMetricsListenAddr, "metrics-addr", "", "listen for endpoint /metrics requests on an additional interface")
 	cmd.Flags().BoolVar(&grpcProxyInsecureDiscovery, "insecure-discovery", false, "accept insecure SRV records")
 	cmd.Flags().StringSliceVar(&grpcProxyEndpoints, "endpoints", []string{"127.0.0.1:2379"}, "comma separated etcd cluster endpoints")
+	cmd.Flags().DurationVar(&grpcProxyClientAutoSyncInterval, "client-auto-sync-interval", 0, "etcd endpoints auto sync interval (disabled by default)")
```
ahrtr, line 134: rename to `endpoints_auto_sync_interval`?

ahrtr:
Thanks @biosvs for the PR, two comments:
1. Please add a changelog item into [CHANGELOG-3.6](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md);
2. Please try to add an integration/e2e test case.

server/etcdmain/grpc_proxy.go
biosvs: Yep, looks better. Changed.

biosvs:
@ahrtr thank you for the quick review!

1. Appended changelog.
2. Wrote e2e test - I used plain commands instead of wrappers, because looks like that's the only way to run independent grpc-proxy and to change cluster members list (without discovery, which I didn't want to use to avoid over complication).

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/pkg/v3/expect"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestGrpcProxyAutoSync(t *testing.T) {
```
ahrtr, line 31: I would suggest to add a test into [gateway_test.go](https://github.com/etcd-io/etcd/blob/main/tests/e2e/gateway_test.go) and  reuse some existing code, and also reuse [ctlV3MemberAdd](https://github.com/etcd-io/etcd/blob/1851316519eea48ef350fa642e8aaa5a8ba6b5d4/tests/e2e/ctl_v3_member_test.go#L193) or [ctlV3MemberRemove](https://github.com/etcd-io/etcd/blob/1851316519eea48ef350fa642e8aaa5a8ba6b5d4/tests/e2e/ctl_v3_member_test.go#L176)

I believe it can greatly simplify the implementation.

ahrtr: Actually a better approach is to add a common test, which can be executed on both integration and e2e test environment. FYI. https://github.com/etcd-io/etcd/issues/13637

biosvs: 1. I'm not sure why grpc-proxy tests should be in the same file with the gateway tests? They are different things, and therefore it's impossible to reuse startGateway function, for example.
2. I tried to use ctl* function, but they accept ctlCtx which I don't have and can't create properly. It useless without setting up epc field of type e2e.EtcdProcessCluster, which in turn I can't use because I need to add/remove member, which is impossible with e2e.EtcdProcessCluster.
3. I really love your idea to use common framework to write tests once and then run them for both integration and e2e env. But for this particular case I have two concerns:
a. Neither e2e framework nor common framework have API for adding/removing members properly.
b. GRPC-proxy is a standalone application, I'm not sure if we at can talk about integration tests at all.

biosvs: I also thought how to write integration test for grpc-proxy, but auto sync is part of client. What I want to check is the code inside etcdmain, which cannot be tested in integration tests way (without actually ports binding, running additional processes, etc.).

So in fact my changes can be tested only in e2e tests, I guess.

ahrtr: > I'm not sure why grpc-proxy tests should be in the same file with the gateway tests? They are different things, and therefore it's impossible to reuse startGateway function, for example.

Yes, you are right. Sorry I misread the filename and content. They are indeed different things. 




> So in fact my changes can be tested only in e2e tests, I guess.

Agreed.



> I tried to use ctl* function, but they accept ctlCtx which I don't have and can't create properly. It useless without setting up epc field of type e2e.EtcdProcessCluster, which in turn I can't use because I need to add/remove member, which is impossible with e2e.EtcdProcessCluster.

There are lots of examples in the existing e2e test cases. Please try to reuse the existing utilities/methods/functions, such as [testCtl](https://github.com/etcd-io/etcd/blob/1851316519eea48ef350fa642e8aaa5a8ba6b5d4/tests/e2e/ctl_v3_test.go#L214), [getMemberList](https://github.com/etcd-io/etcd/blob/1851316519eea48ef350fa642e8aaa5a8ba6b5d4/tests/e2e/ctl_v3_member_test.go#L94), [ctlV3MemberAdd](https://github.com/etcd-io/etcd/blob/1851316519eea48ef350fa642e8aaa5a8ba6b5d4/tests/e2e/ctl_v3_member_test.go#L193), [ctlV3MemberRemove](https://github.com/etcd-io/etcd/blob/1851316519eea48ef350fa642e8aaa5a8ba6b5d4/tests/e2e/ctl_v3_member_test.go#L176), etc.

biosvs: Used e2e.EtcdctlV3, now it looks better (without custom json parser).

As for ctlV3 - I'm afraid I can't use it. Let me explain myself:
- gprc-proxy is not part of ctl utils, it's part of main etcd.
- All ctl* test function requires ctlCtx, which could be created only with e2e.EtcdProcessCluster.
- I can't use e2e.EtcdProcessCluster (neither with ctl*, nor directly), because it spawns all etcd nodes by its own. It's possible to change member list, but impossible to actually start new node or stop another.
- I also can't use e2e.EtcdProcess, because it could be created only from e2e.EtcdServerProcessConfig, which has private field `lg`, which is set only by function that creates e2e.EtcdProcessCluster, which I can't use (previous paragraph).
- At last, I found [number of examples](https://github.com/etcd-io/etcd/search?q=e2e.BinDir+etcd) in e2e tests, where nodes are spawned explicitly. So I hope for now my test fits into other e2e tests (:

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,134 @@
...
+		node2ClientURL, node2PeerURL,
+		"existing", fmt.Sprintf("%s=%s,%s=%s", node1Name, node1PeerURL, node2Name, node2PeerURL),
+	)
+	require.NoError(t, err)
+
+	// Wait for auto sync of endpoints
+	time.Sleep(autoSyncInterval + 5*time.Second)
+
+	memberList, err := memberCtl.MemberList()
+	require.NoError(t, err)
```
ahrtr, line 86: Hard coded the 5 seconds sleep seems not good. Please consider to wait until the 2 members are ready.

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	node1MemberID, err := findMemberIDByEndpoint(memberList.Members, node1ClientURL)
```
ahrtr, line 88: Check `node2ClientURL` as well?

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,134 @@
...
+	require.NoError(t, err)
+
+	// Remove node1 from member list and stop this node
+	_, err = memberCtl.MemberRemove(node1MemberID)
+	require.NoError(t, err)
+	require.NoError(t, proc1.Stop())
```
ahrtr, line 94: It seems that the data dir will not be cleaned up.

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Wait for the full stop
+	time.Sleep(time.Second)
+
+	resp, err := proxyCtl.Get("k1", config.GetOptions{})
+	kvs := testutils.KeyValuesFromGetResponse(resp)
+	assert.Equal(t, []testutils.KV{{Key: "k1", Val: "v1"}}, kvs)
+
+	require.NoError(t, proc2.Stop())
```
ahrtr, line 103: The data dir will not be cleaned up.

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,134 @@
...
+	node1MemberID, err := findMemberIDByEndpoint(memberList.Members, node1ClientURL)
+	require.NoError(t, err)
+
+	// Remove node1 from member list and stop this node
+	_, err = memberCtl.MemberRemove(node1MemberID)
+	require.NoError(t, err)
+	require.NoError(t, proc1.Stop())
+
+	// Wait for the full stop
+	time.Sleep(time.Second)
```
ahrtr, line 97: It seems unnecessary.

tests/e2e/etcd_grpcproxy_test.go
ahrtr: > * It's possible to change member list, but impossible to actually start new node or stop another

It seems like a real gap. Please consider to improve the existing e2e test framework if you have bandwidth and interested.

biosvs: BTW I found another place where adding new member would be useful: https://github.com/etcd-io/etcd/blob/main/tests/e2e/ctl_v3_snapshot_test.go#L258

Sure, I'll try to extend API.

tests/e2e/etcd_grpcproxy_test.go
biosvs: There is no any API event regarding endpoints update. But fortunately grpc-proxy dumps endpoints into log, so I used Expect to wait for the update.

tests/e2e/etcd_grpcproxy_test.go
biosvs: I need memberID only to remove this member from the cluster. Moreover, after fix above (with waiting for endpoints update) we could be sure that second node is in member list.

tests/e2e/etcd_grpcproxy_test.go
biosvs: I get data dir with t.TempDir(), which registers cleanup func that in turn removes created temporary directories.

ahrtr: ack

tests/e2e/etcd_grpcproxy_test.go
biosvs: It's not necessary here, because later we're ensuring that gprc-proxy reads value from second node.

tests/e2e/etcd_grpcproxy_test.go
biosvs: Changed all hardcoded timeouts to retries.

biosvs:
Used new Expect function with the context. And all checks finally have passed.

ahrtr (approved):
Great work! Thanks @biosvs 

I believe majority of the test code should be integrated into the existing e2e test framework, and accordingly simplify this test case and benefit other similar e2e cases. Of course, it can be addressed in separate PR. 

cc @serathius @spzala @ptabor to double check.

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	require.NoError(t, err)
+	kvs := testutils.KeyValuesFromGetResponse(resp)
+	assert.Equal(t, []testutils.KV{{Key: "k1", Val: "v1"}}, kvs)
+
+	require.NoError(t, proc2.Stop())
+	require.NoError(t, proxyProc.Stop())
+}
+
+func runEtcdNode(name, dataDir, clientURL, peerURL, clusterState, initialCluster string) (*expect.ExpectProcess, error) {
```
serathius, line 126: Could you use `tests/framework/e2e.NewEtcdServerProcess instead?

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,180 @@
...
+func waitForEndpointInLog(proxyProc *expect.ExpectProcess, endpoint string) error {
+	endpoint = strings.Replace(endpoint, "http://", "", 1)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	// ExpectFunc reads existing lines, but we're expecting for the events from the future,
+	// that's why sometimes we have to repeat ExpectFunc calls.
+	for {
+		select {
```
serathius, line 164: No need for checking `ctx.Done` here. There is no other IO operations in this loop than `ExpectFunc` which already does this for you.

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -0,0 +1,180 @@
...
+	return 0, fmt.Errorf("member not found")
+}
+
+func waitForEndpointInLog(proxyProc *expect.ExpectProcess, endpoint string) error {
+	endpoint = strings.Replace(endpoint, "http://", "", 1)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	// ExpectFunc reads existing lines, but we're expecting for the events from the future,
```
serathius, line 161: Maybe I'm missing something, but I think this statement is not true. `ExpectFunc` has an infinite for loop that only ends if we match the line or there is an error reported (expected when program exit). So it should be able to read from future events, if it doesn't this is a bug.

tests/e2e/etcd_grpcproxy_test.go
biosvs: Unfortunately I can't, described reasons above: https://github.com/etcd-io/etcd/pull/14354#discussion_r951199683

biosvs: Unfortunately I can't, described reasons above: https://github.com/etcd-io/etcd/pull/14354#discussion_r951199683

serathius: Heh, ctlV3 is terrible :(. Sorry for not noticing above comment.

tests/e2e/etcd_grpcproxy_test.go
biosvs: Yes, you're right. My mind somehow kept old interpretation of what this function does. Yesterday I changed my mind, but today forget it once again, lol.

Get rid of redundant code.

biosvs:
Is it possible to rerun check (https://github.com/etcd-io/etcd/actions/runs/2926560942)? Seems like it flaky.

serathius (approved):

biosvs:
@ahrtr how to you think, is it possible to port it to 3.5.5?

ahrtr:
> @ahrtr how to you think, is it possible to port it to 3.5.5?

Although it's an enhancement, but it's also a bug from another perspective, because the gRPC proxy may lose connection to the backend etcd cluster if there are member changes. 

I am OK to backport this, but I have no strong opinion on this. What do you think? @serathius @spzala @ptabor

biosvs:
So @ahrtr seems like there are no objections. How it's better to do? Just cherry-pick the commit into `release-3.5` branch? Or apply patch with another commit?

ahrtr:
> So @ahrtr seems like there are no objections. How it's better to do? Just cherry-pick the commit into `release-3.5` branch? Or apply patch with another commit?

Yes, please feel free to deliver a PR for this. I think you need to manually take care of the cherry-pick instead of automatically applying the patch, because there is big difference between `release-3.5 `and `main`.

PR #14357, opened 2022-08-17
clientv3: close streams after use in lessor keepAliveOnce method

amdprophet:
We use `client.KeepAliveOnce` heavily in Sensu Go. Since upgrading to Etcd 3.5 we've seen an excessive amount of logs containing `failed to receive lease keepalive request from gRPC stream` when logging at debug level. I believe this is happening as there's no attempt to close the stream after the `keepAliveOnce` method has completed. The stream is instead closed by the deferred call to cancel the context.

The log lines no longer appear after attempting to close the stream using the code in this PR.

### DIFF
--- a/client/v3/lease.go
+++ b/client/v3/lease.go
@@ -397,7 +397,7 @@ func (l *lessor) closeRequireLeader() {
 	}
 }
 
-func (l *lessor) keepAliveOnce(ctx context.Context, id LeaseID) (*LeaseKeepAliveResponse, error) {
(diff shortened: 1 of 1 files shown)


amdprophet:
@ahrtr updated

ahrtr (approved):
LGTM

Thank you @amdprophet

ahrtr:
cc @spzala

spzala (approved):
lgtm
Thanks @amdprophet

PR #14359, opened 2022-08-18
Cleanup all etcd-agent and etcd processes when the test finishes

ahrtr:
There are two cases, when interrupted by users, then forcibly kill all processes. Otherwise, gracefully terminate all processes.

cc @spzala 


spzala (approved):
Thanks @ahrtr !!

PR #14360, opened 2022-08-18
tests: unify `--strict-reconfig-check` config of common framework

clarkfw:
### Purpose
This PR is opened to discuss the different policies of `--strict-reconfig-check' between e2e test and integration test.

### What went wrong?
As described in https://github.com/etcd-io/etcd/pull/14281#issuecomment-1219263217, the `MemberAdd` function works fine in integration test but encounters an `unhealthy cluster` error in e2e test.

### Why?
Integration cluster has `--strict-reconfig-check' disabled by default:
https://github.com/etcd-io/etcd/blob/d05c894e854e1b8545c41ff26ac85c8b76917a7c/tests/framework/integration/cluster.go#L723

in the meantime, e2e cluster has `--strict-reconfig-check' enabled by default:
https://github.com/etcd-io/etcd/blob/d05c894e854e1b8545c41ff26ac85c8b76917a7c/tests/framework/e2e/cluster.go#L313-L315

therefore, in e2e cluster, functions like `MemberAdd`, `MemberRemove` and `TransferLeadership` would encounter an `unhealthy cluster` error before members `isConnectedFullySince`, which will take 5 seconds:
https://github.com/etcd-io/etcd/blob/d05c894e854e1b8545c41ff26ac85c8b76917a7c/server/etcdserver/server.go#L1350-L1358

### Solution
I've come up with some solutions. But I'm not sure if they have more hidden risks.
1. `MemberAdd` function could wait 5 seconds before all members `isConnectedFullySince`. Apparently it'll make our code difficult to maintain.
2. Expose `--strict-reconfig-check` to common framework, in order to give test cases control of whether to enable/disable `--strict-reconfig-check` in the process of cluster creation.
3. Enable or disable `--strict-reconfig-check` both in e2e cluster and integration cluster.

I tend to go with the 3rd solution. Though the 2nd solution may be required to do some specific tests in the future, we can  skip it now.

### What have I done?
I made some changes to disable `--strict-reconfig-check` in e2e cluster of common framework. At the same time, I kept it enabled as it was in the old framework, to avoid breaking existing test cases.

### DIFF
--- a/tests/e2e/ctl_v3_test.go
+++ b/tests/e2e/ctl_v3_test.go
@@ -122,9 +122,9 @@ func dialWithSchemeTest(cx ctlCtx) {
 }
 
 type ctlCtx struct {
-	t                          *testing.T
-	apiPrefix                  string
-	cfg                        e2e.EtcdProcessClusterConfig
+	t         *testing.T
+	apiPrefix string
+	cfg       e2e.EtcdProcessClusterConfig
 
 	corruptFunc                func(string) error
 	disableStrictReconfigCheck bool
--- a/tests/e2e/ctl_v3_test.go
+++ b/tests/e2e/ctl_v3_test.go
@@ -122,11 +122,12 @@ func dialWithSchemeTest(cx ctlCtx) {
 }
 
 type ctlCtx struct {
-	t                *testing.T
-	apiPrefix        string
-	cfg              e2e.EtcdProcessClusterConfig
-	corruptFunc      func(string) error
-	noStrictReconfig bool
+	t                          *testing.T
+	apiPrefix                  string
+	cfg                        e2e.EtcdProcessClusterConfig
+
+	corruptFunc                func(string) error
+	disableStrictReconfigCheck bool
 
 	epc *e2e.EtcdProcessCluster
 
@@ -185,8 +186,8 @@ func withCorruptFunc(f func(string) error) ctlOption {
 	return func(cx *ctlCtx) { cx.corruptFunc = f }
 }
 
-func withNoStrictReconfig() ctlOption {
-	return func(cx *ctlCtx) { cx.noStrictReconfig = true }
+func withDisableStrictReconfig() ctlOption {
+	return func(cx *ctlCtx) { cx.disableStrictReconfigCheck = true }
 }
 
 func withApiPrefix(p string) ctlOption {
@@ -226,7 +227,7 @@ func testCtlWithOffline(t *testing.T, testFunc func(ctlCtx), testOfflineFunc fun
 	if !ret.quorum {
 		ret.cfg = *e2e.ConfigStandalone(ret.cfg)
 	}
-	ret.cfg.NoStrictReconfig = ret.noStrictReconfig
+	ret.cfg.DisableStrictReconfigCheck = ret.disableStrictReconfigCheck
 	if ret.initialCorruptCheck {
 		ret.cfg.InitialCorruptCheck = ret.initialCorruptCheck
 	}
--- a/tests/framework/config/cluster.go
+++ b/tests/framework/config/cluster.go
@@ -27,8 +27,9 @@ const (
 )
 
 type ClusterConfig struct {
-	ClusterSize       int
-	PeerTLS           TLSConfig
-	ClientTLS         TLSConfig
-	QuotaBackendBytes int64
+	ClusterSize                int
+	PeerTLS                    TLSConfig
+	ClientTLS                  TLSConfig
+	QuotaBackendBytes          int64
+	DisableStrictReconfigCheck bool
 }
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -166,14 +166,14 @@ type EtcdProcessClusterConfig struct {
 
 	CipherSuites []string
 
-	ForceNewCluster     bool
-	InitialToken        string
-	QuotaBackendBytes   int64
-	NoStrictReconfig    bool
-	EnableV2            bool
-	InitialCorruptCheck bool
-	AuthTokenOpts       string
-	V2deprecation       string
+	ForceNewCluster            bool
+	InitialToken               string
+	QuotaBackendBytes          int64
+	DisableStrictReconfigCheck bool
+	EnableV2                   bool
+	InitialCorruptCheck        bool
+	AuthTokenOpts              string
+	V2deprecation              string
 
 	RollingStart bool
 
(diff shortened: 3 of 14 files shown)


ahrtr:
Good catch and thanks for raising this discussion.

Comments:
1. We should expose a configuration item for the `--strict-reconfig-check ` in the common config;
2. Since the item is enabled by default in production environment, so we should enable it by default as well in our test; and we can add a couple of common case to intentionally disable it. 
3. If above change breaks some existing cases, then we need to investigate whether it's expected. If yes, then we can disable the item for the specific cases, otherwise fix the issue.

clarkfw:
Thanks for your analysis!
I agree with you on enabling it by default. I'm gonna do it and see what happens.

serathius (approved):
LGTM, leaving for @ahrtr to approve and merge

tests/framework/e2e/cluster.go
```diff
@@ -310,7 +310,7 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfigs(tb testing.TB) []*
 				"--quota-backend-bytes", fmt.Sprintf("%d", cfg.QuotaBackendBytes),
 			)
 		}
-		if cfg.NoStrictReconfig {
+		if !cfg.StrictReconfig {
 			args = append(args, "--strict-reconfig-check=false")
 		}
```
ahrtr, line 315:
```suggestion
		if cfg.StrictReconfig {
			args = append(args, "--strict-reconfig-check")
		}
```

ahrtr:
> Thanks for your analysis! I agree with you on enabling it by default. I'm gonna do it and see what happens.

@clarkfw Are you going to do this in this PR or in a separate PR?

tests/framework/e2e/cluster.go
ahrtr: Please ignore this comment, which isn't correct.

clarkfw:
I'm going to do it in this PR, still working on it.

clarkfw:
A new commit was force pushed. It enables `--strict-reconfig-check` in both e2e and integration cluster. It also exposes `--strict-reconfig-check` to `ClusterConfig`.

The checks are expected to fail. Testing on my own computer shows that 15 test cases failed due to this commit. Most of them use `MemberAdd` or `MemberRemove` function.

I'm gonna push another commit to fix or discuss these failed test cases. Before that, I would like to know if changes in this commit are appropriate @ahrtr .


image

tests/e2e/ctl_v3_test.go
```diff
@@ -191,8 +191,8 @@ func withCorruptFunc(f func(string) error) ctlOption {
 	return func(cx *ctlCtx) { cx.corruptFunc = f }
 }
 
-func withNoStrictReconfig() ctlOption {
-	return func(cx *ctlCtx) { cx.noStrictReconfig = true }
+func withDisableStrictReconfig() ctlOption {
+	return func(cx *ctlCtx) { cx.disableStrictReconfigCheck = true }
 }
```
ahrtr, line 196:
```suggestion
func withStrictReconfig() ctlOption {
	return func(cx *ctlCtx) { cx.disableStrictReconfigCheck = true }
}
```

ahrtr:
@clarkfw  yes, overall the change looks good to me. Please continue to finish the task. Thanks.

I see that the e2e configuration is a bit mess. There are three ways to set the configuration.
1. Users(test cases) can set a [configuration object/struct](https://github.com/etcd-io/etcd/blob/e40828540cef4e7dbb92ed3ec65a666d87dc8519/tests/e2e/ctl_v3_test.go#L162) directly;
2. Users(test cases) can set some fields via the `ctlCtx`. For example, set the [initialCorruptCheck](https://github.com/etcd-io/etcd/blob/e40828540cef4e7dbb92ed3ec65a666d87dc8519/tests/e2e/ctl_v3_test.go#L186) in `ctlCtx`, then pass it to [cfg.InitialCorruptCheck](https://github.com/etcd-io/etcd/blob/e40828540cef4e7dbb92ed3ec65a666d87dc8519/tests/e2e/ctl_v3_test.go#L240);
3. Users(test cases) can set some fields to `EtcdProcessClusterConfig` directly, such as [withMaxConcurrentStreams](https://github.com/etcd-io/etcd/blob/e40828540cef4e7dbb92ed3ec65a666d87dc8519/tests/e2e/ctl_v3_test.go#L208) or [CorruptCheckTime](https://github.com/etcd-io/etcd/blob/e40828540cef4e7dbb92ed3ec65a666d87dc8519/tests/e2e/corrupt_test.go#L106) 

I would suggest to remove all `EtcdProcessClusterConfig` related fields from `ctlCtx`, and keep using #2 and #3 above. WDYT? @serathius @spzala @clarkfw 

Of course, it is unrelated to this PR.

serathius:
+1 to config cleanup (not needed to be addressed in this PR). For me whole usage `ctlCtx` is an anti-pattern of how tests should be written, so it would be even better if we just get rid of it :P

clarkfw:
With `--strict-reconfig-check` enabled by default, some test cases fail because they call `MemberAdd` or `MemberRemove` function.
In this commit(https://github.com/etcd-io/etcd/pull/14360/commits/93a17b9d8f64f2d9d87e6f7dc0928065ad31c476), I disabled `--strict-reconfig-check` for the failed test cases.

clarkfw:
There are still some check failures. I'm working on it.

clarkfw:
Thank God, all checks have passed.
Please take a look when you guys are available🤓

tests/framework/config/cluster.go
```diff
@@ -27,8 +27,9 @@ const (
...
 type ClusterConfig struct {
-	ClusterSize       int
-	PeerTLS           TLSConfig
-	ClientTLS         TLSConfig
-	QuotaBackendBytes int64
+	ClusterSize                int
+	PeerTLS                    TLSConfig
+	ClientTLS                  TLSConfig
+	QuotaBackendBytes          int64
+	DisableStrictReconfigCheck bool
```
ahrtr, line 34: The name `DisableStrictReconfigCheck` seems to be anti-pattern because users (test cases) need to use double negative to enable it, `not disabled == enabled`. It's also a tradeoff, please see https://github.com/etcd-io/etcd/pull/14389#discussion_r955674249 .

You just follow the same naming as the existing e2e configuration `NoStrictReconfig`, so it isn't your fault. So we can keep it as it's for now, and feel free to change it to `StrictReconfigCheck` so as to be consistent with the definition in [the production code](https://github.com/etcd-io/etcd/blob/3c79a4dc3801b347a91b114524bdeb4712573a5e/server/embed/config.go#L240) in a separate PR.

ahrtr:
Overall looks good to me. I see that you disabled the StrictReconfigCheck for some cases, did you intentionally disable them to increase the test coverage or fix the test failures?

clarkfw:
> Overall looks good to me. I see that you disabled the `StrictReconfigCheck` for some cases, did you intentionally disable them to increase the test coverage or fix the test failures?

Just to fix the test failures. After enabling `StrictReconfigCheck` these tests failed due to calling `MemberAdd` or `MemberRemove` function. I disabled `StrictReconfigCheck` for theses tests to make them run as they did before.

ahrtr (approved):
LGTM

Thank you @clarkfw

PR #14376, opened 2022-08-24
Fixed infinite loop in ExpectProcess.ExpectFunc

biosvs:
I wrote unit test (expect_test.go, TestExpectFuncTimeout) that shows how to get stuck in the infinite loop of ExpectFunc.

The problem is that this function releases lock, sleeps and reads new lines. For some talkative processes 10ms is enough to write new line.

First purpose of the diff is to show the problem and possible solution. I'm not sure about interface for Expect function (it seems like you're okay with context, but maybe you prefer time.Duration as a parameter for some reasons) and existing code - should I use timeout there or not?

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -18,6 +18,7 @@ package expect
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -96,7 +97,7 @@ func (ep *ExpectProcess) read() {
 }
 
 // ExpectFunc returns the first line satisfying the function f.
-func (ep *ExpectProcess) ExpectFunc(f func(string) bool) (string, error) {
+func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (string, error) {
 	i := 0
 
 	for {
@@ -114,7 +115,13 @@ func (ep *ExpectProcess) ExpectFunc(f func(string) bool) (string, error) {
 			break
 		}
 		ep.mu.Unlock()
-		time.Sleep(time.Millisecond * 10)
+
+		select {
+		case <-ctx.Done():
+			return "", fmt.Errorf("failed to find match string: %w", ctx.Err())
+		case <-time.After(time.Millisecond * 10):
+			// continue loop
+		}
 	}
 	ep.mu.Lock()
 	lastLinesIndex := len(ep.lines) - DEBUG_LINES_TAIL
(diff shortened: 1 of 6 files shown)


pkg/expect/expect.go
```diff
@@ -128,9 +135,14 @@ func (ep *ExpectProcess) ExpectFunc(f func(string) bool) (string, error) {
 		ep.err, lastLines)
 }
 
+// ExpectWithContext returns the first line containing the given string.
+func (ep *ExpectProcess) ExpectWithContext(ctx context.Context, s string) (string, error) {
+	return ep.ExpectFunc(ctx, func(txt string) bool { return strings.Contains(txt, s) })
+}
+
 // Expect returns the first line containing the given string.
```
serathius, line 143: Can you mark Expect as deprecated?

serathius (approved):
This change was on my plate for long time, thanks for proposing it.
Push context all the way!

pkg/expect/expect.go
```diff
@@ -96,10 +97,16 @@ func (ep *ExpectProcess) read() {
...
 // ExpectFunc returns the first line satisfying the function f.
-func (ep *ExpectProcess) ExpectFunc(f func(string) bool) (string, error) {
+func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (string, error) {
 	i := 0
 
 	for {
+		select {
+		case <-ctx.Done():
+			return "", fmt.Errorf("failed to find match string: %w", ctx.Err())
+		default:
```
serathius, line 107: We could consider merging this select with time.Sleep

biosvs: Tried to merge with time.Sleep and simplify a little bit.
Looks legitimately and units are passed (:

pkg/expect/expect.go
biosvs: Done

serathius (approved):

pkg/expect/expect.go
```diff
@@ -109,12 +110,18 @@ func (ep *ExpectProcess) ExpectFunc(f func(string) bool) (string, error) {
 				return line, nil
 			}
 		}
+		ep.mu.Unlock()
+
```
serathius, line 114: This makes reading `ep.err` outside of lock. 

It's not obvious, but it should be ok? Even if comparing interfaces is not atomic, it should work as you are checking equality to nil. Maybe I'm wrong, but I hope you thought about this :P

ahrtr (approved):
Looks great to me!

Thank you @biosvs 

Are you going to replace all the usage of the deprecated `Expect` with `ExpectFunc` or `ExpectWithContext` in separte PR(s)?

biosvs:
@ahrtr at first I want to end with pull/14354 and to try to extend e2e framework. Maybe some time later I'll return to this Expect function.
Btw, you know incomparably more about tests. I don't think it makes sense to use clean context.WithTimeout all the time. Is there a good place where context with timeout should be created? Place for helper function?

ahrtr:
> I don't think it makes sense to use clean context.WithTimeout all the time. Is there a good place where context with timeout should be created? Place for helper function?

Probably we can add a method to [ctlCtx](https://github.com/etcd-io/etcd/blob/96a2669839bd5b9afd82ba16886b74e426bc7a6c/tests/e2e/ctl_v3_test.go#L123) something like below,
```
func (cx *ctlCtx) ContextWithTimeout() (context.Context, context.CancelFunc) {
	timeout := cx.getTestTimeout()
	return context.WithTimeout(context.Background(), timeout)
}
```

Also refer to [ctl_v3_test.go#L166-L172](https://github.com/etcd-io/etcd/blob/96a2669839bd5b9afd82ba16886b74e426bc7a6c/tests/e2e/ctl_v3_test.go#L166-L172)

serathius:
I will look into adding context to common framework, it's long overdue for context.

Issue #14333, opened 2022-08-11
[test] TestBalancerUnderNetworkPartitionDelete failure


ahrtr:
Raised on the latest code on `main` branch.

```
=== FAIL: integration/clientv3/connectivity TestBalancerUnderNetworkPartitionDelete (8.12s)
    cluster.go:532: Creating listener with addr: 127.0.0.1:2109805817
......
    logger.go:130: 2022-08-10T22:44:54.878Z	WARN	m1	request stats	{"member": "m1", "start time": "2022-08-10T22:44:53.878Z", "time spent": "1.000107558s", "remote": "@", "response type": "/etcdserverpb.KV/DeleteRange", "request count": 0, "request size": 3, "response count": 0, "response size": 0, "request content": "key:\"a\" "}
    logger.go:130: 2022-08-10T22:44:54.879Z	WARN	client	retrying of unary invoker failed	{"target": "etcd-endpoints://0xc000550f00/localhost:m0", "method": "/etcdserverpb.KV/DeleteRange", "attempt": 0, "error": "rpc error: code = Unavailable desc = etcdserver: request timed out, possibly due to previous leader failure"}
    network_partition_test.go:137: Op returned error: etcdserver: request timed out, possibly due to previous leader failure
    network_partition_test.go:138: Cancelling...
    network_partition_test.go:144: #0: expected 'expected error', got 'etcdserver: request timed out, possibly due to previous leader failure'
    logger.go:130: 2022-08-10T22:44:54.993Z	INFO	m0.raft	c16a4db1d4d2aea3 is starting a new election at term 8	{"member": "m0"}
......
    logger.go:130: 2022-08-10T22:44:56.922Z	INFO	m0.raft	c16a4db1d4d2aea3 became candidate at term 21	{"member": "m0"}
Saved JUnit XML test report to /home/runner/work/etcd/etcd/linux-amd64-integration-4-cpu/junit_MTY2MDE3MTMxMwo.xml
FAIL: 'integration' failed at Wed Aug 10 22:48:23 UTC 2022
......
     logger.go:130: 2022-08-10T22:44:59.879Z	DEBUG	client	retrying of unary invoker	{"target": "etcd-endpoints://0xc000550f00/localhost:m0", "method": "/etcdserverpb.KV/DeleteRange", "attempt": 0}
    network_partition_test.go:137: Op returned error: <nil>
    network_partition_test.go:138: Cancelling...
[...]

PR #14377, opened 2022-08-24
Add ErrTimeoutDueToLeaderFail as expected error

tjungblu:
This fixes ci flakes where timeouts are expected to happen in network
partitions.

issue: https://github.com/etcd-io/etcd/issues/14333

### DIFF
--- a/tests/integration/clientv3/connectivity/network_partition_test.go
+++ b/tests/integration/clientv3/connectivity/network_partition_test.go
@@ -33,13 +33,18 @@ import (
 
 var errExpected = errors.New("expected error")
 
+func isErrorExpected(err error) bool {
+	return clientv3test.IsClientTimeout(err) || clientv3test.IsServerCtxTimeout(err) ||
+		err == rpctypes.ErrTimeout || err == rpctypes.ErrTimeoutDueToLeaderFail
+}
+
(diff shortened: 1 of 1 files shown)


tests/integration/clientv3/connectivity/network_partition_test.go
```diff
@@ -23,23 +23,29 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/grpc"
```
tjungblu, line 26: I really wonder where this import inconsistency comes from, are you guys using a different gofmt? I'm on `go version go1.18.2 linux/amd64`

serathius: Etcd haven't upgraded go to 1.18

ahrtr:
This might not be the root cause of https://github.com/etcd-io/etcd/issues/14333, because `clus.Members[0]` will not receive any message from the new leader when it loses connection with other two peers. 

But I agree that we should add `ErrTimeoutDueToLeaderFail` into the expected error list. So this PR looks good to me.

ahrtr (approved):
LGTM

Thank you @tjungblu

serathius (approved):

Issue #14384, opened 2022-08-25
[test] after failure functional tests can't access logs in /tmp/etcd-functional*


lavacat:
### What happened?

In case of functional test failure, test.sh can't access etcd logs.

/tmp/etcd-functional* dirs are already removed

> PASSES='functional' ./scripts/test.sh

--- FAIL: TestFunctional (16.32s)
    --- FAIL: TestFunctional/round:0 (1.24s)
        --- FAIL: TestFunctional/round:0/0_BLACKHOLE_PEER_PORT_TX_RX_LEADER_UNTIL_TRIGGER_SNAPSHOT (1.24s)
FAIL
FAIL: (code:1):
  % './bin/etcd-tester' '--config' './tests/functional/functional.yaml' '-test.v'
./scripts/test.sh: line 183: kill: (21716) - No such process
./scripts/test.sh: line 183: kill: (21717) - No such process
./scripts/test.sh: line 183: kill: (21718) - No such process
ETCD_TESTER_EXIT_CODE: 1

FAILED! 'tail -1000 /tmp/etcd-functional-1/etcd.log'
tail: cannot open '/tmp/etcd-functional-1/etcd.log' for reading: No such file or directory

FAILED! 'tail -1000 /tmp/etcd-functional-2/etcd.log'
tail: cannot open '/tmp/etcd-functional-2/etcd.log' for reading: No such file or directory

FAILED! 'tail -1000 /tmp/etcd-functional-3/etcd.log'
tail: cannot open '/tmp/etcd-functional-3/etcd.log' for reading: No such file or directory
--- FAIL: exit code 1

### What did you expect to happen?

/tmp/etcd-functional-* directories should be preserved to allow server log inspection.

### How can we reproduce it (as minimally and precisely as possible)?

to force test failure set 
> retries := 1
in case.go line 270

### Etcd version (please run commands below)

3.6, 3.4
haven't tested in 3.5 but probably the same.

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

lavacat:
I'll submit PR shortly, please assign this to me

PR #14387, opened 2022-08-25
functional: remove SIGQUIT_ETCD_AND_REMOVE_DATA_AND_STOP_AGENT command

lavacat:
Problem: both SIGQUIT_ETCD_AND_REMOVE_DATA_AND_STOP_AGENT and test.sh
will attempt to stop agents and remove directories.

Solution: since test.sh creates directories and starts test, it should be
responsible for cleanup.

See https://github.com/etcd-io/etcd/issues/14384

### DIFF
--- a/tests/functional/tester/cluster.go
+++ b/tests/functional/tester/cluster.go
@@ -424,28 +424,7 @@ func (clus *Cluster) broadcast(op rpcpb.Operation) error {
 		if err == nil {
 			continue
 		}
-
-		if err != nil {
-			destroyed := false
-			if op == rpcpb.Operation_SIGQUIT_ETCD_AND_REMOVE_DATA_AND_STOP_AGENT {
-				if err == io.EOF {
-					destroyed = true
-				}
-				if strings.Contains(err.Error(),
-					"rpc error: code = Unavailable desc = transport is closing") {
-					// agent server has already closed;
-					// so this error is expected
-					destroyed = true
-				}
-				if strings.Contains(err.Error(),
-					"desc = os: process already finished") {
-					destroyed = true
-				}
-			}
-			if !destroyed {
-				errs = append(errs, err.Error())
-			}
-		}
+		errs = append(errs, err.Error())
 	}
 
 	if len(errs) == 0 {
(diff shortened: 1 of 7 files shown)


ahrtr:
@lavacat please fix the fmt pipeline error

tests/functional/tester/cluster.go
```diff
@@ -424,28 +424,6 @@ func (clus *Cluster) broadcast(op rpcpb.Operation) error {
...
-					// so this error is expected
-					destroyed = true
-				}
-				if strings.Contains(err.Error(),
-					"desc = os: process already finished") {
-					destroyed = true
-				}
-			}
-			if !destroyed {
-				errs = append(errs, err.Error())
```
ahrtr, line 446: IT seems that you need to keep this line of code

```
errs = append(errs, err.Error())
```

ahrtr:
It seems that the etcd data directories will never be removed? What will happen if we restart the functional test when it fails?

lavacat:
> It seems that the etcd data directories will never be removed? What will happen if we restart the functional test when it fails?

This line in test.sh will remove data dirs at the start of the tests. 
https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L159

ahrtr:
> > It seems that the etcd data directories will never be removed? What will happen if we restart the functional test when it fails?
> 
> This line in test.sh will remove data dirs at the start of the tests. https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L159

I am thinking probably it makes more sense to stop etcd and cleanup the data if the test is successful; otherwise keep the data. Proposed changes:
1. Update the [(*Cluster) Run](https://github.com/etcd-io/etcd/blob/2fa73023a0dc65e8f6bf5e032df15bba6f8cddca/tests/functional/tester/cluster_run.go#L34) to return an `error`, and [TestFunctional](https://github.com/etcd-io/etcd/blob/2fa73023a0dc65e8f6bf5e032df15bba6f8cddca/tests/functional/cmd/etcd-tester/etcd_tester_test.go#L51) sends a `Operation_SIGQUIT_ETCD_AND_REMOVE_DATA` to the `agent` if the returned error is `nil`.
2. The test may `panic`, so you need to use `recover()` to check whether [(*Cluster) Run](https://github.com/etcd-io/etcd/blob/2fa73023a0dc65e8f6bf5e032df15bba6f8cddca/tests/functional/tester/cluster_run.go#L34) is successful or not.

ahrtr:
> This line in test.sh will remove data dirs at the start of the tests.
> https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L159

This should be just a remedy instead of a formal solution. Please consider to update this PR per https://github.com/etcd-io/etcd/pull/14387#issuecomment-1248574599

serathius (approved):

ahrtr:
@lavacat It's OK to resolve the comment in a separate PR, so that we can get this one merged firstly.  Please let me know whether you want to resolve it in this PR or in a separate PR.

ahrtr:
Since there is no response for a long time, so let me merge this PR firstly and take care of the comment in a separate PR.

ahrtr (approved):

Issue #13785, opened 2022-03-10
Track stabilization --experimental-wait-cluster-ready-timeout


serathius:
This is tracking issue that tracks graduation of this feature. Let's work towards stabilization of the feature by discussing what steps   should be taken for graduation.
Context https://github.com/etcd-io/etcd/issues/13775

ahrtr:
This flag was added in 3.6. Normally we should graduate a flag at least in the next release (>=3.7). But for this flag, since it's low risk, I am OK to graduate it in 3.6 if there is no any objections. cc @serathius @spzala @ptabor

serathius:
I have bulk created issues to graduate all experimental flag. For this flag that was added v3.6 it should be reasonable to wait for v3.7.

niconorsk:
Hi, not sure if this is the correct place to ask, but noticed one slight issue with this new flag.  We don't notify to systemd that we are ready to go which means that etcd ends up in a restart loop in this scenario.

I've fixed this in a local install by changing `startEtcd` in `etcdmain\etcd.go` to also timeout according to the config flag.

I'm ok with putting up an MR for this, but wanted to check that this was a desired change in the first place, and also suggestions to where to add tests for something like that

ahrtr:
> We don't notify to systemd that we are ready to go which means that etcd ends up in a restart loop in this scenario.

I am not sure I get your point. Do you mean systemd restarted etcd because etcd blocked on [serve.go#L105](https://github.com/etcd-io/etcd/blob/82bf79d4ce382f5d1333fb6e7b0242ac67869e8e/server/embed/serve.go#L105)?

The [PR](https://github.com/etcd-io/etcd/pull/13525) isn't cherry picked to 3.5. Did you build etcd on `main` branch yourself?

niconorsk:
I tested this with a patch on 3.3.11 because that is the old version I am using but the code in question does not to look to have changed.

Specifically, etcd blocks on [etcd.go#L208](https://github.com/etcd-io/etcd/blob/82bf79d4ce382f5d1333fb6e7b0242ac67869e8e/server/etcdmain/etcd.go#L208)

Which means that it nevers sends the message back to systemd saying it should be considered started and so systemd eventually times it out and restarts it

ahrtr:
Got it. It is a real issue to me. If other members get started too late or slow for whatever reason, then the running member may be restarted by systemd; accordingly it makes the situation even worse. 

[...]

PR #13525, opened 2021-12-08
Etcd server shouldn't wait for the ready notification infinitely on startup

ahrtr:
Fix [13524](https://github.com/etcd-io/etcd/issues/13524 ).

etcd server shouldn't wait for the ready notification infinitely on startup, otherwise, it can't even serve serializable read request. 

The simplest solution is to add a timeout. I am not sure whether we should add a flag to the command line parameters. 

cc @serathius @xiang90 @gyuho @ptabor

PR #14388, opened 2022-08-26
etcdmain: Honour ExperimentalWaitClusterReadyTimeout in startEtcd

niconorsk:
When we can't reach quorum, we were waiting forever and never sending
the systemd notify message. As a result, systemd would eventually time out
and restart the etcd process which likely would make the unhealthy cluster
in an even worse state

Improves #13785

### DIFF
--- a/tests/e2e/ctl_v3_grpc_test.go
+++ b/tests/e2e/ctl_v3_grpc_test.go
@@ -18,6 +18,7 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"testing"
@@ -148,7 +149,7 @@ func firstMatch(t *testing.T, expectLine string, logs ...e2e.LogsExpect) string 
 	match := make(chan string, len(logs))
 	for i := range logs {
 		go func(l e2e.LogsExpect) {
-			line, _ := l.Expect(expectLine)
+			line, _ := l.ExpectWithContext(context.TODO(), expectLine)
 			match <- line
 		}(logs[i])
 	}
--- a/tests/framework/e2e/etcd_process.go
+++ b/tests/framework/e2e/etcd_process.go
@@ -15,9 +15,12 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"os"
+	"testing"
+	"time"
 
 	"go.uber.org/zap"
 
@@ -48,7 +51,7 @@ type EtcdProcess interface {
 }
 
 type LogsExpect interface {
-	Expect(string) (string, error)
+	ExpectWithContext(context.Context, string) (string, error)
 	Lines() []string
 	LineCount() int
 }
@@ -178,4 +181,15 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 		ep.cfg.lg.Panic("Please grab logs before process is stopped")
 	}
 	return ep.proc
+}
+
+func AssertProcessLogs(t *testing.T, ep EtcdProcess, expectLog string) {
+	t.Helper()
+	var err error
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	_, err = ep.Logs().ExpectWithContext(ctx, expectLog)
+	if err != nil {
+		t.Fatal(err)
+	}
 }
--- a/server/etcdmain/etcd.go
+++ b/server/etcdmain/etcd.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
 	"go.etcd.io/etcd/client/pkg/v3/fileutil"
 	"go.etcd.io/etcd/client/pkg/v3/logutil"
@@ -207,6 +208,8 @@ func startEtcd(cfg *embed.Config) (<-chan struct{}, <-chan error, error) {
 	select {
 	case <-e.Server.ReadyNotify(): // wait for e.Server to join the cluster
 	case <-e.Server.StopNotify(): // publish aborted from 'ErrStopped'
+	case <-time.After(cfg.ExperimentalWaitClusterReadyTimeout):
+		e.GetLogger().Warn("startEtcd: timed out waiting for the ready notification")
 	}
 	return e.Server.StopNotify(), e.Err(), nil
 }
--- a/tests/e2e/cluster_downgrade_test.go
+++ b/tests/e2e/cluster_downgrade_test.go
@@ -61,7 +61,7 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 
 	t.Log("Downgrade enabled, validating if cluster is ready for downgrade")
 	for i := 0; i < len(epc.Procs); i++ {
-		expectLog(t, epc.Procs[i], "The server is ready to downgrade")
+		e2e.AssertProcessLogs(t, epc.Procs[i], "The server is ready to downgrade")
 		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: lastVersionStr, Server: currentVersionStr})
 	}
 	t.Log("Cluster is ready for downgrade")
@@ -73,7 +73,7 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 		startEtcd(t, epc.Procs[i], lastReleaseBinary)
 	}
 	t.Log("All members downgraded, validating downgrade")
-	expectLog(t, leader(t, epc), "the cluster has been downgraded")
+	e2e.AssertProcessLogs(t, leader(t, epc), "the cluster has been downgraded")
 	for i := 0; i < len(epc.Procs); i++ {
 		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: lastVersionStr, Server: lastVersionStr})
 	}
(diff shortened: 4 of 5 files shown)


tests/framework/e2e/etcd_process.go
```diff
@@ -179,3 +182,14 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 	}
 	return ep.proc
 }
+
+func ExpectLog(t *testing.T, ep EtcdProcess, expectLog string) {
+	t.Helper()
+	var err error
+	testutils.ExecuteWithTimeout(t, 30*time.Second, func() {
+		_, err = ep.Logs().Expect(expectLog)
```
ahrtr, line 190: `Expect` is deprecated, please use `ExpectWithContext` instead.

tests/e2e/no_quorum_notify_daemon_test.go
```diff
@@ -0,0 +1,41 @@
...
+	// Remove two members, so that only one etcd will get started
+	epc.Procs = epc.Procs[:1]
+
+	// Start the etcd cluster with only one member
+	if err := epc.Start(); err != nil {
+		t.Fatalf("Failed to start the etcd cluster: %v", err)
+	}
+
+	// Expect log message indicating systemd notify message has been sent
+	e2e.ExpectLog(t, epc.Procs[0], "notifying init daemon")
```
ahrtr, line 39: Suggest to add code something like below as well
```
e2e.ExpectLog(t, epc.Procs[0], "startEtcd: timed out waiting for the ready notification")

```

tests/framework/e2e/etcd_process.go
niconorsk: I don't think I am understanding this request.  I can't find any other uses of `ExpectWithContext` in the etcd repo and google isn't giving me much either

ahrtr: `ExpectWithContext` was just added yesterday

tests/e2e/no_quorum_notify_daemon_test.go
niconorsk: Done

tests/framework/e2e/etcd_process.go
niconorsk: Ok found it and done now

tests/framework/e2e/etcd_process.go
```diff
@@ -179,3 +183,14 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 	}
 	return ep.proc
 }
+
+func ExpectLog(t *testing.T, ep EtcdProcess, expectLog string) {
+	t.Helper()
+	var err error
+	testutils.ExecuteWithTimeout(t, 30*time.Second, func() {
+		_, err = ep.Logs().ExpectWithContext(context.Background(), expectLog)
+	})
```
ahrtr, line 192: It seems that there is no need to use `testutils.ExecuteWithTimeout` at all, you can use `ExpectWithContext` directly. You just need to create a context with Timeout, and pass it to `ExpectWithContext`

tests/e2e/no_quorum_notify_daemon_test.go
```diff
@@ -0,0 +1,43 @@
...
+	// Expect log message indicating time out waiting for quorum hit
+	e2e.ExpectLog(t, epc.Procs[0], "startEtcd: timed out waiting for the ready notification")
```
ahrtr, line 41: Minor comment: change the order of the log, so as to be consistent with the order they being generated. 
```suggestion
	// Expect log message indicating time out waiting for quorum hit
	e2e.ExpectLog(t, epc.Procs[0], "startEtcd: timed out waiting for the ready notification")
	// Expect log message indicating systemd notify message has been sent
	e2e.ExpectLog(t, epc.Procs[0], "notifying init daemon")
```

tests/framework/e2e/etcd_process.go
niconorsk: Done

tests/e2e/no_quorum_notify_daemon_test.go
niconorsk: Done

ahrtr:
Please fix the fmt errors raised in the pipeline

ahrtr (approved):
LGTM

Good job! @niconorsk Thank you!

cc @serathius @spzala @ptabor to double check.

tests/e2e/ctl_v3_grpc_test.go
```diff
@@ -148,7 +149,7 @@ func firstMatch(t *testing.T, expectLine string, logs ...e2e.LogsExpect) string
 	match := make(chan string, len(logs))
 	for i := range logs {
 		go func(l e2e.LogsExpect) {
-			line, _ := l.Expect(expectLine)
+			line, _ := l.ExpectWithContext(context.Background(), expectLine)
```
serathius, line 152:
```suggestion
			line, _ := l.ExpectWithContext(context.TODO(), expectLine)
```

tests/framework/e2e/etcd_process.go
```diff
@@ -179,3 +182,14 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 	}
 	return ep.proc
 }
+
+func ExpectLog(t *testing.T, ep EtcdProcess, expectLog string) {
+	t.Helper()
+	var err error
```
serathius, line 188: Err is not set by anyone

tests/framework/e2e/etcd_process.go
```diff
@@ -179,3 +182,14 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 	}
 	return ep.proc
 }
+
+func ExpectLog(t *testing.T, ep EtcdProcess, expectLog string) {
```
serathius, line 186: Don't like this helper function because:
* It has confusing naming `Expect` doesn't imply calling `t.Fatal` in context of e2e tests as it's used differently throughout code
* It's so short that it adds code instead of shortening it.

tests/e2e/no_quorum_notify_daemon_test.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2021 The etcd Authors
```
serathius, line 1: Would suggest to rename file to `no_quorum_ready_test.go` so we can add more than one test for ExperimentalWaitClusterReadyTimeout flag.

tests/e2e/no_quorum_notify_daemon_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// Start the etcd cluster with only one member
+	if err := epc.Start(); err != nil {
+		t.Fatalf("Failed to start the etcd cluster: %v", err)
+	}
+
+	// Expect log message indicating time out waiting for quorum hit
+	e2e.ExpectLog(t, epc.Procs[0], "startEtcd: timed out waiting for the ready notification")
+	// Expect log message indicating systemd notify message has been sent
+	e2e.ExpectLog(t, epc.Procs[0], "notifying init daemon")
```
serathius, line 41: It this sufficient signal for us to confirm that `ExperimentalWaitClusterReadyTimeout` works? I think the idea for the flag was to allow serializable requests without quorum. Do we have a test for that? If not, could you just add serializable request to this test?

ahrtr: > I think the idea for the flag was to allow serializable requests without quorum. Do we have a test for that?

Yes, we have added test for that long ago (last year?). This PR is just to enhance the https://github.com/etcd-io/etcd/pull/13525 to address https://github.com/etcd-io/etcd/issues/13785#issuecomment-1226776763

serathius: Awesome, can we have the both tests in one file? They test the same feature so it makes sense they are close to each other.

ahrtr: I thought about it, but it isn't feasible. I recall that previous tests run for both e2e and integration. But this test only run for e2e.

serathius: :( ok

tests/e2e/no_quorum_notify_daemon_test.go
niconorsk: Done

tests/framework/e2e/etcd_process.go
niconorsk: Oops. Fixed

tests/e2e/ctl_v3_grpc_test.go
niconorsk: Done

tests/framework/e2e/etcd_process.go
niconorsk: I have rename this function as AssertProcessLogs but happy to name it something that fits more with the projects naming conventions

As an argument to keeping it as a helper, it already has 4 callers in 2 different files, and in my personal experience, waiting for a log message is a sadly common way of making end-to-end tests

tests/e2e/no_quorum_notify_daemon_test.go
niconorsk: [kv_test.go#L202](https://github.com/etcd-io/etcd/blob/f00578c3cdb848fd8c4449899aebfe6bbc5dd2c6/tests/common/kv_test.go#L202)

The existing test for reference.  It got moved in to common when it was just e2e in the original MR

serathius (approved):

PR #14389, opened 2022-08-26
Cleanup some e2e test configurations

ahrtr:
See https://github.com/etcd-io/etcd/pull/14360#issuecomment-1227863196

Notes:
1. `compactPhysical` in `ctlCtx` and `withQuota` aren't used at all, they are dead code.
2. `quotaBackendBytes` in `ctlCtx` isn't used either. Instead, users (test cases) set the [QuotaBackendBytes](https://github.com/etcd-io/etcd/blob/e40828540cef4e7dbb92ed3ec65a666d87dc8519/tests/framework/e2e/cluster.go#L170) directly.
3. If we want to enable a feature (e.g. `initialCorruptCheck`) by default, then a best practice is to define the field as "`DisableXX`", such as `DisableInitialCorruptCheck bool`. Its value will be `false` by default if not set.

### DIFF
--- a/tests/e2e/ctl_v3_test.go
+++ b/tests/e2e/ctl_v3_test.go
@@ -121,12 +121,11 @@ func dialWithSchemeTest(cx ctlCtx) {
 }
 
 type ctlCtx struct {
-	t                 *testing.T
-	apiPrefix         string
-	cfg               e2e.EtcdProcessClusterConfig
-	quotaBackendBytes int64
-	corruptFunc       func(string) error
-	noStrictReconfig  bool
+	t                *testing.T
+	apiPrefix        string
+	cfg              e2e.EtcdProcessClusterConfig
+	corruptFunc      func(string) error
+	noStrictReconfig bool
 
 	epc *e2e.EtcdProcessCluster
 
@@ -143,9 +142,6 @@ type ctlCtx struct {
 
 	initialCorruptCheck bool
 
-	// for compaction
-	compactPhysical bool
-
 	// dir that was used during the test
 	dataDir string
 }
@@ -156,6 +152,7 @@ func (cx *ctlCtx) applyOpts(opts []ctlOption) {
 	for _, opt := range opts {
 		opt(cx)
 	}
+
 	cx.initialCorruptCheck = true
 }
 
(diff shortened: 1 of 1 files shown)


ahrtr:
cc @serathius @spzala @ptabor @clarkfw

tests/framework/e2e/cluster.go
```diff
@@ -165,14 +165,14 @@ type EtcdProcessClusterConfig struct {
...
-	EnableV2            bool
-	InitialCorruptCheck bool
-	AuthTokenOpts       string
-	V2deprecation       string
+	ForceNewCluster            bool
+	InitialToken               string
+	QuotaBackendBytes          int64
+	NoStrictReconfig           bool
+	EnableV2                   bool
+	DisableInitialCorruptCheck bool
```
serathius, line 173: I don't think this direction is much better. This is bad way to handle defaults, negating the configuration fields again makes it harder to track and understand.

Instead I would prefer that we separate struct definition from defaults. This way we can make the configuration struct more similar to "production" one and just have tests having different defaults.

Example of what I mean:
```
type  EtcdProcessClusterConfig struct {
...
  InitialCorruptionCheck bool
...
} 

func DefaultEtcdProcessClusterConfig() *EtcdProcessClusterConfig {
  return &EtcdProcessClusterConfig{
    InitialCorruptionCheck: true,
  }
}

func TestCorruption(t *testing.T) {
  config := DefaultEtcdProcessClusterConfig()
  config.InitialCorruptionCheck = false
  ...
}
```

Later we might decide to shorten shorten code by adding options for DefaultEtcdProcessClusterConfig, but this is separate issue.

ahrtr: > I don't think this direction is much better. This is bad way to handle defaults, negating the configuration fields again makes it harder to track and understand.
> 
> Instead I would prefer that we separate struct definition from defaults. This way we can make the configuration struct more similar to "production" one and just have tests having different defaults.

Thanks. Actually it's a tradeoff. The good side is that we can leverage golang build-in feature of zero value (e.g. `false` for bool), so that the implementation can be simplified because we need to do nothing if we just want to use the default value; while the bad side it's counterintuitive (not disable == enable), and it isn't consistent with the product code just as you pointed out.

Let me follow your suggestion in this PR.  I will rollback the change to `InitialCorruptCheck`, and we can address it in a separate PR.

ahrtr: Done.

serathius (approved):

Issue #14370, opened 2022-08-22
Durability API guarantee broken in single node cluster


hasethuraman:
I observed the possibility of data loss and I would like the community to comment / correct me otherwise.

Before explaining that, I would like to explain the happy path when user does a PUT <key, value>. I have tried to only necessary steps to focus this issue. And considered a single etcd instance.

====================================================================================
----------api thread --------------

User calls etcdctl PUT k v

It lands in v3_server.go::put function with the message about k,v

Call delegates to series of function calls and enters v3_server.go::processInternalRaftRequestOnce

It registers for a signal with wait utility against this keyid

Call delegates further to series of function calls and enters raft/node.go::stepWithWaitOption(..message..)

It wraps this message in a msgResult channel and updates its result channel; then sends this message to propc channel.

After sending it waits on msgResult.channel
----------api thread waiting --------------

On seeing a message in propc channel, raft/node.go::run(), it wakes up and sequence of calls adds the message.Entries to raftLog

Notifies the msgResult.channel

----------api thread wakes--------------
10. Upon seeing the msgResult.channel, api thread wakes and returns down the stack back to v3_server.go::processInternalRaftRequestOnce and waits for signal that it registered at step#4
----------api thread waiting --------------

In next iteration of raft/node.go::run(), it gets the entry from raftLog and add it to readyc
etcdserver/raft.go::start wakes up on seeing this entry in readyc and adds this entry to applyc channel
and synchronously writes to wal log ---------------------> wal log
etcdserver/server.go wakes up on seeing entry in applyc channel (added in step https://github.com/etcd-io/etcd/pull/12)
From step#14, the call goes through series of calls and lands in server.go::applyEntryNormal
applyEntryNormal calls applyV3.apply which will eventually puts the KV to mvcc kvstore txn kvindex
applyEntryNormal now sends the signal for this key which is basically to wake up api thread that is waiting in 7
----------api thread wakes--------------
18. User thread here wakes and sends back acknowledgement
----------user sees ok--------------

Batcher flushes the entries added to kvstore txn kvindex to database file. (also this can happen before 18 based on its timer)
====================================================================================
Here if step https://github.com/etcd-io/etcd/pull/13 thread is pre-empted and rescheduled by the underlying operating system after completing step https://github.com/etcd-io/etcd/pull/18 and when there is a power failure at the end of step 18 where after user sees error, then the kv is neither written to wal nor to database file

[...]

PR #14394, opened 2022-08-27
Fix the potential data loss for clusters with only one member

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/14370

For a cluster with only one member, the raft always send identical
unstable entries and committed entries to etcdserver, and etcd
responds to the client once it finishes (actually partially) the
applying workflow.

When the client receives the response, it doesn't mean etcd has already
successfully saved the data, including BoltDB and WAL, because:
   1. etcd commits the boltDB transaction periodically instead of on each request;
   2. etcd saves WAL entries in parallel with applying the committed entries.
Accordingly, it may run into a situation of data loss when the etcd crashes
immediately after responding to the client and before the boltDB and WAL
successfully save the data to disk.
Note that this issue can only happen for clusters with only one member.

For clusters with multiple members, it isn't an issue, because etcd will
not commit & apply the data before it being replicated to majority members.
When the client receives the response, it means the data must have been applied.
It further means the data must have been committed.
Note: for clusters with multiple members, the raft will never send identical
unstable entries and committed entries to etcdserver.

### DIFF
--- a/server/etcdserver/raft.go
+++ b/server/etcdserver/raft.go
@@ -68,8 +68,10 @@ func init() {
 type toApply struct {
 	entries  []raftpb.Entry
 	snapshot raftpb.Snapshot
-	// notifyc synchronizes etcd server applies with the raft node
-	notifyc chan struct{}
+	// snapNotifyc synchronizes etcd server applies with the raft node
+	snapNotifyc chan struct{}
+	// walNotifyc synchronizes etcd server applies with the WAL persistence
+	walNotifyc chan struct{}
 }
 
 type raftNode struct {
@@ -200,11 +202,16 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 					}
 				}
 
-				notifyc := make(chan struct{}, 1)
+				snapNotifyc := make(chan struct{}, 1)
+				var walNotifyc chan struct{}
+				if shouldWaitWALSync(rd) {
+					walNotifyc = make(chan struct{}, 1)
+				}
 				ap := toApply{
-					entries:  rd.CommittedEntries,
-					snapshot: rd.Snapshot,
-					notifyc:  notifyc,
+					entries:     rd.CommittedEntries,
+					snapshot:    rd.Snapshot,
+					snapNotifyc: snapNotifyc,
+					walNotifyc:  walNotifyc,
 				}
 
 				updateCommittedIndex(&ap, rh)
@@ -237,6 +244,12 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 				if err := r.storage.Save(rd.HardState, rd.Entries); err != nil {
 					r.lg.Fatal("failed to save Raft hard state and entries", zap.Error(err))
 				}
+
+				if walNotifyc != nil {
+					// etcdserver should wait for this notification before responding to client.
+					walNotifyc <- struct{}{}
+				}
+
 				if !raft.IsEmptyHardState(rd.HardState) {
 					proposalsCommitted.Set(float64(rd.HardState.Commit))
 				}
@@ -252,7 +265,7 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 					}
 
 					// etcdserver now claim the snapshot has been persisted onto the disk
-					notifyc <- struct{}{}
+					snapNotifyc <- struct{}{}
 
 					// gofail: var raftBeforeApplySnap struct{}
 					r.raftStorage.ApplySnapshot(rd.Snapshot)
@@ -272,7 +285,7 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 					msgs := r.processMessages(rd.Messages)
 
 					// now unblocks 'applyAll' that waits on Raft log disk writes before triggering snapshots
-					notifyc <- struct{}{}
+					snapNotifyc <- struct{}{}
 
 					// Candidate or follower needs to wait for all pending configuration
 					// changes to be applied before sending messages.
@@ -293,7 +306,7 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 						// to be in sync with scheduled config-change job
 						// (assume notifyc has cap of 1)
 						select {
-						case notifyc <- struct{}{}:
+						case snapNotifyc <- struct{}{}:
 						case <-r.stopped:
 							return
 						}
@@ -303,7 +316,7 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 					r.transport.Send(msgs)
 				} else {
 					// leader already processed 'MsgSnap' and signaled
-					notifyc <- struct{}{}
+					snapNotifyc <- struct{}{}
 				}
 
 				r.Advance()
@@ -312,6 +325,42 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 			}
 		}
 	}()
+}
+
+// For a cluster with only one member, the raft may send both the
+// unstable entries and committed entries to etcdserver, and there
+// may have overlapped log entries between them.
+//
+// etcd responds to the client once it finishes (actually partially)
+// the applying workflow. But when the client receives the response,
+// it doesn't mean etcd has already successfully saved the data,
+// including BoltDB and WAL, because:
+//    1. etcd commits the boltDB transaction periodically instead of on each request;
+//    2. etcd saves WAL entries in parallel with applying the committed entries.
+// Accordingly, it might run into a situation of data loss when the etcd crashes
+// immediately after responding to the client and before the boltDB and WAL
+// successfully save the data to disk.
+// Note that this issue can only happen for clusters with only one member.
+//
+// For clusters with multiple members, it isn't an issue, because etcd will
+// not commit & apply the data before it being replicated to majority members.
+// When the client receives the response, it means the data must have been applied.
+// It further means the data must have been committed.
+// Note: for clusters with multiple members, the raft will never send identical
+// unstable entries and committed entries to etcdserver.
+//
+// Refer to https://github.com/etcd-io/etcd/issues/14370.
+func shouldWaitWALSync(rd raft.Ready) bool {
+	if len(rd.CommittedEntries) == 0 || len(rd.Entries) == 0 {
+		return false
+	}
+
+	// Check if there is overlap between unstable and committed entries
+	// assuming that their index and term are only incrementing.
+	lastCommittedEntry := rd.CommittedEntries[len(rd.CommittedEntries)-1]
+	firstUnstableEntry := rd.Entries[0]
+	return lastCommittedEntry.Term > firstUnstableEntry.Term ||
+		(lastCommittedEntry.Term == firstUnstableEntry.Term && lastCommittedEntry.Index >= firstUnstableEntry.Index)
 }
 
 func updateCommittedIndex(ap *toApply, rh *raftReadyHandler) {
--- a/server/etcdserver/raft_test.go
+++ b/server/etcdserver/raft_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/pkg/v3/pbutil"
 	"go.etcd.io/etcd/raft/v3"
@@ -190,6 +191,52 @@ func TestStopRaftWhenWaitingForApplyDone(t *testing.T) {
 
 // TestConfigChangeBlocksApply ensures toApply blocks if committed entries contain config-change.
 func TestConfigChangeBlocksApply(t *testing.T) {
+	testEtcdserverAndRaftInteraction(t, testConfigChangeBlockApply)
+}
+
+func TestWALSyncNotBlockApply(t *testing.T) {
+	testEtcdserverAndRaftInteraction(t, testWALSyncNotBlockApply)
+}
+
+func TestWALSyncBlockApply(t *testing.T) {
+	testEtcdserverAndRaftInteraction(t, testWALSyncBlockApply)
+}
+
+func testConfigChangeBlockApply(t *testing.T, srv *EtcdServer, n *readyNode) toApply {
+	n.readyc <- raft.Ready{
+		SoftState:        &raft.SoftState{RaftState: raft.StateFollower},
+		CommittedEntries: []raftpb.Entry{{Type: raftpb.EntryConfChange}},
+	}
+	return <-srv.r.applyc
+}
+
+func testWALSyncNotBlockApply(t *testing.T, srv *EtcdServer, n *readyNode) toApply {
+	n.readyc <- raft.Ready{
+		SoftState:        &raft.SoftState{RaftState: raft.StateFollower},
+		CommittedEntries: []raftpb.Entry{{Type: raftpb.EntryConfChange}},
+	}
+	ap := <-srv.r.applyc
+	if ap.walNotifyc != nil {
+		t.Error("unexpected ap.walNotifyc, expected nil")
+	}
+	return ap
+}
+
+func testWALSyncBlockApply(t *testing.T, srv *EtcdServer, n *readyNode) toApply {
+	n.readyc <- raft.Ready{
+		SoftState:        &raft.SoftState{RaftState: raft.StateFollower},
+		Entries:          []raftpb.Entry{{Type: raftpb.EntryConfChange}},
+		CommittedEntries: []raftpb.Entry{{Type: raftpb.EntryConfChange}},
+	}
+	ap := <-srv.r.applyc
+	if ap.walNotifyc == nil {
+		t.Error("unexpected ap.walNotifyc, expected not nil")
+	}
+	assert.NotEqual(t, nil, ap.walNotifyc)
+	return ap
+}
+
+func testEtcdserverAndRaftInteraction(t *testing.T, testFunc func(*testing.T, *EtcdServer, *readyNode) toApply) {
 	n := newNopReadyNode()
 
 	r := newRaftNode(raftNodeConfig{
@@ -208,11 +255,7 @@ func TestConfigChangeBlocksApply(t *testing.T) {
 	})
 	defer srv.r.Stop()
 
-	n.readyc <- raft.Ready{
-		SoftState:        &raft.SoftState{RaftState: raft.StateFollower},
-		CommittedEntries: []raftpb.Entry{{Type: raftpb.EntryConfChange}},
-	}
-	ap := <-srv.r.applyc
+	ap := testFunc(t, srv, n)
 
 	continueC := make(chan struct{})
 	go func() {
@@ -228,7 +271,7 @@ func TestConfigChangeBlocksApply(t *testing.T) {
 	}
 
 	// finish toApply, unblock raft routine
-	<-ap.notifyc
+	<-ap.snapNotifyc
 
 	select {
 	case <-continueC:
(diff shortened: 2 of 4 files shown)


ahrtr:
cc @ptabor @serathius @spzala This might be an important fix, please take a look, thx.

Note that clusters with only one member isn't recommended in production usage in the existing official releases, including 3.5.[0-4] and 3.4.x, because it may cause data loss when etcd crashes and under high load. cc @dims @liggitt

server/etcdserver/raft.go
```diff
@@ -314,6 +328,38 @@ func (r *raftNode) start(rh *raftReadyHandler) {
...
+func shouldWaitWALSync(unstableEntries []raftpb.Entry, committedEntries []raftpb.Entry) bool {
+	if len(unstableEntries) != len(committedEntries) {
+		return false
+	}
+
+	if len(unstableEntries) == 0 {
+		return false
+	}
+
+	return reflect.DeepEqual(unstableEntries, committedEntries)
```
liggitt, line 360: Is this as expensive as it looks?

server/etcdserver/raft.go
```diff
@@ -314,6 +328,38 @@ func (r *raftNode) start(rh *raftReadyHandler) {
...
+// successfully save the data to disk.
+// Note that this issue can only happen for clusters with only one member.
+//
+// For clusters with multiple members, it isn't an issue, because etcd will
+// not commit & apply the data before it being replicated to majority members.
+// When the client receives the response, it means the data must have been applied.
+// It further means the data must have been committed.
+// Note: for clusters with multiple members, the raft will never send identical
+// unstable entries and committed entries to etcdserver.
+func shouldWaitWALSync(unstableEntries []raftpb.Entry, committedEntries []raftpb.Entry) bool {
```
liggitt, line 351: Is there a way to have a single code path that is safe regardless of whether we're in multi-server or single server mode?

server/etcdserver/raft.go
ahrtr: Actually I thought about this when writing this function.

Firstly, I added code above to do [some initial check](https://github.com/ahrtr/etcd/blob/0d8644b6141027a002781e93e3750fe997a595a0/server/etcdserver/raft.go#L352-L358). For cluster with multiple members,  the `reflect.DeepEqual` will not get executed at all.

Secondly, the `reflect.DeepEqual` indeed is a little expensive for cluster with only one member. The alternative is to check the log index, and it should be inexpensive. The applying workflow should only wait for the WAL sync if there are any overlap between unstableEntries and commitEntries.

server/etcdserver/raft.go
ahrtr: I do not get your point. 

For multi-member cluster, there is no need to wait for the WAL sync, and this function will always return `false`.

server/etcdserver/raft.go
ahrtr: Updated & thx.

server/etcdserver/raft.go
ahrtr: There are two solutions in my mind before delivering this PR.

The first solution is to enhance the existing raft protocal. The existing raft workflow commit each log immediately when it receves each proposal for clusters with only member, because it doesn't need to get confirmation from itself. Accordingly it sends both the unstable logs and committed logs to etcdserver. The solution is to send a message to etcdserver and wait for the confirmation, no matter it's single-server or multi-server. The good side of this solution is that it looks elegant. The bad side it has some impact on the performance, and it also needs to update the stable raft package. It might be what your `a single code path` means.

The second solutio is what this PR delivers. The good side is that it has little performance impact, and no impact on multi-server clusters at all. The bad side is that it complicates the applying workflow, but it should be accepted. 

Eventually I followed the second solution above for now.

server/etcdserver/raft.go
```diff
@@ -237,6 +244,12 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 				if err := r.storage.Save(rd.HardState, rd.Entries); err != nil {
 					r.lg.Fatal("failed to save Raft hard state and entries", zap.Error(err))
 				}
+
+				if walNotifyc != nil {
+					// etcdserver should wait for this notification before responding to client.
+					walNotifyc <- struct{}{}
```
serathius, line 250: I don't think this is correct. Interestingly it's not raft code that responds to client. It's apply code.

serathius: The exact code where client is notified is `s.w.Trigger`. For example https://github.com/etcd-io/etcd/blob/eb25e433cda952cfa2bf9ff9289b86fa8899f193/server/etcdserver/server.go#L1889

server/etcdserver/raft.go
```diff
@@ -314,6 +327,43 @@ func (r *raftNode) start(rh *raftReadyHandler) {
...
+// successfully save the data to disk.
+// Note that this issue can only happen for clusters with only one member.
+//
+// For clusters with multiple members, it isn't an issue, because etcd will
+// not commit & apply the data before it being replicated to majority members.
+// When the client receives the response, it means the data must have been applied.
+// It further means the data must have been committed.
+// Note: for clusters with multiple members, the raft will never send identical
+// unstable entries and committed entries to etcdserver.
+func shouldWaitWALSync(unstableEntries []raftpb.Entry, committedEntries []raftpb.Entry) bool {
```
serathius, line 350: could you just take `Ready` as argument? unstableEntries and commitedEntries have the same type `[]raftpb.Entry`, it is easy to mistake one for the other.

server/etcdserver/raft.go
```diff
@@ -314,6 +327,43 @@ func (r *raftNode) start(rh *raftReadyHandler) {
...
+	if len(unstableEntries) != len(committedEntries) {
```
serathius, line 351: Instead of checking if all entries match, we should check if there is a common entry in both.

server/etcdserver/raft.go
```diff
@@ -314,6 +327,43 @@ func (r *raftNode) start(rh *raftReadyHandler) {
...
+	if len(unstableEntries) == 0 {
+		return false
+	}
+
+	// Previously `reflect.DeepEqual` is used to compare the `unstableEntries`
+	// and `committedEntries`, but it's a little expensive, so we just compare
+	// the log index. The applying workflow should wait for the WAL sync if
+	// there is any log index overlap between them.
+	lastCommittedEntry := committedEntries[len(committedEntries)-1]
+	return lastCommittedEntry.Term >= unstableEntries[0].Term && lastCommittedEntry.Index >= unstableEntries[0].Index
```
serathius, line 364:
```suggestion
	return lastCommittedEntry.Term > unstableEntries[0].Term || (lastCommittedEntry.Term == unstableEntries[0].Term && lastCommittedEntry.Index >= unstableEntries[0].Index)
```

server/etcdserver/raft_test.go
```diff
@@ -317,3 +360,62 @@ func TestStopRaftNodeMoreThanOnce(t *testing.T) {
...
+			commitedEntries: []raftpb.Entry{{Term: 4, Index: 10, Type: raftpb.EntryNormal, Data: []byte{0x11, 0x22, 0x33}}},
+			expectedResult:  false,
+		},
+		{
+			name:            "has different term and index",
+			unstableEntries: []raftpb.Entry{{Term: 3, Index: 11, Type: raftpb.EntryNormal, Data: []byte{0x11, 0x22, 0x33}}},
+			commitedEntries: []raftpb.Entry{{Term: 4, Index: 10, Type: raftpb.EntryNormal, Data: []byte{0x11, 0x22, 0x33}}},
+			expectedResult:  false,
+		},
+		{
```
serathius, line 407: Please add case that there is a common element between entries.

server/etcdserver/raft.go
serathius: Proposed fix by @ahrtr makes sense for me.

server/etcdserver/raft.go
```diff
@@ -237,6 +244,12 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 				if err := r.storage.Save(rd.HardState, rd.Entries); err != nil {
 					r.lg.Fatal("failed to save Raft hard state and entries", zap.Error(err))
 				}
+
+				if walNotifyc != nil {
```
lavacat, line 248: I think you can remove snapNotifyc and walNotifyc and just use `r.applyc <- ap` here for 1 node cluster (instead of line 220).
What does snapNotifyc achieve? If I understand this correctly it will allow applyAll->applySnapshot to run sooner. But this is a 1 node cluster case, so applySnapshot from leader will never be triggered because there are no followers.

If my logic is correct, then you can also rename shouldWaitWALSync() to isSingleNodeCluster()

server/etcdserver/raft.go
ahrtr: Makes sense.  Thx & updated.

server/etcdserver/raft.go
ahrtr: Thx. Eventually I changed to below,
```
	if len(rd.CommittedEntries) == 0 || len(rd.Entries) == 0 {
		return false
	}
```

server/etcdserver/raft.go
ahrtr: I thought about this before, and the comment is only partially correct. Specially, it isn't correct in this case. Points:
1. The index indeed `might` go down, because unstable(uncommitted) log might be overwritten by new leader in a new term. But committed log will never be overwritten. It means the index will never go down to the committed index, in other words, the index will always be bigger than the committed index. 
2. Especially for a cluster with only one member, there is no leader change at all. So this comment doesn't stand. 

Both the original code and your suggest code work, but your suggested change has better readability, so updated per your suggestion.

server/etcdserver/raft_test.go
ahrtr: updated & thx

server/etcdserver/raft.go
ahrtr: It isn't good to mix the new patch/fix with the existing code/logic, because it may break the existing functionality and cause regression. 

Note that there may be multiple entries (including both unstable and committed entries for single-server cluster) in `Ready`, and it may also have zero entry for raft `no-op`. The current implementation triggers the responding to client when finish applying each entry, but we can't send&receive the `walNotifyc` multiple times.

The above explanation may look complicated. The point here is it's really subtle and error prone, and the best practice for this case is to create a separate channel for this requirement (waiting for WAL sync).

dims:
cc @chaochn47 @geetasg

ahrtr:
# Performance comparison 
Linux server configuration
```
MemTotal:       16423532 kB
16CPU, and each with Intel(R) Xeon(R) Gold 5120 CPU @ 2.20GHz
```

Commands:
```
./etcd  --quota-backend-bytes=4300000000  
./benchmark txn-put --endpoints="http://127.0.0.1:2379" --clients=200 --conns=200 --key-space-size=4000000000 --key-size=128 --val-size=10240  --total=200000 --rate=40000
```

# Result on one-server cluster
Note that I tried multiple times, and got stable results.

## Result on `main`
```
Summary:
  Total:	56.1029 secs.
  Slowest:	0.1439 secs.
  Fastest:	0.0021 secs.
  Average:	0.0559 secs.
  Stddev:	0.0289 secs.
  Requests/sec:	3564.8803

Response time histogram:
  0.0021 [1]	|
  0.0163 [20555]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0304 [14212]	|∎∎∎∎∎∎∎∎∎
  0.0446 [38302]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0588 [56887]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0730 [12060]	|∎∎∎∎∎∎∎∎
  0.0872 [16407]	|∎∎∎∎∎∎∎∎∎∎∎
  0.1013 [25805]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1155 [11993]	|∎∎∎∎∎∎∎∎
  0.1297 [3452]	|∎∎
  0.1439 [326]	|

Latency distribution:
  10% in 0.0161 secs.
  25% in 0.0361 secs.
  50% in 0.0506 secs.
  75% in 0.0785 secs.
  90% in 0.0981 secs.
  95% in 0.1087 secs.
  99% in 0.1203 secs.
  99.9% in 0.1360 secs.
```

## Result on branch `one_member_data_loss`
```
Summary:
  Total:	59.1221 secs.
  Slowest:	0.1435 secs.
  Fastest:	0.0128 secs.
  Average:	0.0590 secs.
  Stddev:	0.0213 secs.
  Requests/sec:	3382.8273

Response time histogram:
  0.0128 [1]	|
  0.0259 [1029]	|
  0.0390 [36291]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0520 [64848]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0651 [25016]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0782 [26937]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0912 [25053]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1043 [16544]	|∎∎∎∎∎∎∎∎∎∎
  0.1174 [3387]	|∎∎
  0.1305 [764]	|
  0.1435 [130]	|

Latency distribution:
  10% in 0.0361 secs.
  25% in 0.0417 secs.
  50% in 0.0516 secs.
  75% in 0.0754 secs.
  90% in 0.0920 secs.
  95% in 0.0965 secs.
  99% in 0.1116 secs.
  99.9% in 0.1267 secs.
```

# Summary
Overall the performance decreases by about 5.38% ((3564 - 3382) / 3382).

ahrtr:
# Result on three-server cluster
Commands:
```
$ goreman start  

$ ./benchmark txn-put --endpoints="http://127.0.0.1:2379" --clients=200 --conns=200 --key-space-size=1000000000 --key-size=128 --val-size=10240  --total=100000 --rate=40000
```

## Result on `main`
```
Summary:
  Total:	58.5007 secs.
  Slowest:	0.2334 secs.
  Fastest:	0.0135 secs.
  Average:	0.1168 secs.
  Stddev:	0.0349 secs.
  Requests/sec:	1709.3817

Response time histogram:
  0.0135 [1]	|
  0.0355 [166]	|
  0.0575 [2216]	|∎∎∎
  0.0795 [16128]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1015 [19326]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1235 [14502]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1455 [25602]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1674 [15279]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1894 [5343]	|∎∎∎∎∎∎∎∎
  0.2114 [1140]	|∎
  0.2334 [297]	|

Latency distribution:
  10% in 0.0719 secs.
  25% in 0.0857 secs.
  50% in 0.1208 secs.
  75% in 0.1420 secs.
  90% in 0.1606 secs.
  95% in 0.1727 secs.
  99% in 0.1920 secs.
  99.9% in 0.2294 secs.
```

## Result on branch `one_member_data_loss`
```
Summary:
  Total:	57.8285 secs.
  Slowest:	0.2478 secs.
  Fastest:	0.0114 secs.
  Average:	0.1155 secs.
  Stddev:	0.0338 secs.
  Requests/sec:	1729.2500

Response time histogram:
  0.0114 [1]	|
  0.0350 [48]	|
  0.0587 [2551]	|∎∎∎
  0.0823 [20336]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1060 [16035]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1296 [21643]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1532 [26055]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.1769 [11141]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.2005 [1703]	|∎∎
  0.2241 [465]	|
  0.2478 [22]	|

Latency distribution:
  10% in 0.0714 secs.
  25% in 0.0839 secs.
  50% in 0.1199 secs.
  75% in 0.1410 secs.
  90% in 0.1594 secs.
  95% in 0.1664 secs.
  99% in 0.1883 secs.
  99.9% in 0.2166 secs.
```

# Summary
Overall the performance results are the same.

ahrtr:
Points: 
1. There is no any performance impact on multi-server cluster.
2. There is slightly downgrade (about 5.38%) of the performance for single-server cluster.  Correctness takes precedence over performance. So I think the PR is accepted, and should be cherry picked to 3.5, and probably 3.4.

serathius:
Makes sense, looks like the performance regression is mostly visible in 10 percentile of latency distribution. I would expect that much lower 10%ile benefited from lack of durability. I think it's reasonable to trade latency for durability for those requests. 

I support backporting this change as it is for v3.4 and v3.5

server/etcdserver/raft.go
```diff
@@ -314,6 +327,41 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 	}()
 }
 
+// For a cluster with only one member, the raft always send identical
```
serathius, line 330: Note about comments, it should be written from a point for the issue being fixed. Issue that is fixed can be just linked.

server/etcdserver/raft.go
```diff
@@ -314,6 +327,41 @@ func (r *raftNode) start(rh *raftReadyHandler) {
...
+// unstable entries and committed entries to etcdserver.
+func shouldWaitWALSync(rd raft.Ready) bool {
+	if len(rd.CommittedEntries) == 0 || len(rd.Entries) == 0 {
+		return false
+	}
+
+	// Previously `reflect.DeepEqual` is used to compare the `unstableEntries`
+	// and `committedEntries`, but it's a little expensive, so we just compare
+	// the log index. The applying workflow should wait for the WAL sync if
+	// there is any log index overlap between them.
```
serathius, line 358:
```suggestion
	// Check if there is overlap between unstable and committed entries assuming that their index and term are only incrementing.
```

serathius (approved):
LGTM, however let's wait for more maintainers to have a look

server/etcdserver/raft.go
ahrtr: Updated & thx

server/etcdserver/raft.go
ahrtr: Good catch. Thx & updated.

ahrtr:
> LGTM, however let's wait for more maintainers to have a look

Thanks @serathius for the quick review. Please @ptabor and  @spzala take a look, thx

ahrtr:
> Makes sense, looks like the performance regression is mostly visible in 10 percentile of latency distribution. I would expect that much lower 10%ile benefited from lack of durability. I think it's reasonable to trade latency for durability for those requests.
> 
> I support backporting this change as it is for v3.4 and v3.5

ack. One more point, the faster the disk I/O, the smaller the performance downgrade. It means when the disk I/O is faster enough, then the performance downgrade should be even smaller.

spzala (approved):
@ahrtr thanks for the great work and benchmark results! We need to add an entry to changelog for this, but that can be done separately. Also, the backport approach sounds good, thanks @serathius

ahrtr:
> @ahrtr thanks for the great work and benchmark results! We need to add an entry to changelog for this, but that can be done separately. Also, the backport approach sounds good, thanks @serathius

Thanks @spzala . We need to make a decision to merge this one or https://github.com/etcd-io/etcd/pull/14400.  Either way, I will update the changelog in separate PR, and after backporting the PR.

ahrtr:
Closing this PR because we eventually merged https://github.com/etcd-io/etcd/pull/14400 .

Issue #14393, opened 2022-08-27
Coverage pipeline error on main branch


ahrtr:
Issue on the latest code on `main`

```
ERROR: Tests for following packages failed:
   go.etcd.io/etcd/tests/v3/e2e
mv: cannot stat '/home/runner/work/etcd/etcd/covdir/e2e/*': No such file or directory
[Sat Aug 27 10:22:49 UTC 2022] Collecting coverage from e2e tests with proxy ...
% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' '-tags=cov cluster_proxy' '-timeout' '30m' 'go.etcd.io/etcd/tests/v3/e2e')
stderr: # go.etcd.io/etcd/tests/v3/framework/e2e
Error: stderr: framework/e2e/cluster_proxy.go:129:33: too many arguments in call to SpawnCmdWithLogger
stderr: have (*zap.Logger, []string, nil, string)
stderr: want (*zap.Logger, []string, map[string]string)
Error: stderr: framework/e2e/etcd_process.go:107:33: too many arguments in call to SpawnCmdWithLogger
stderr: have (*zap.Logger, []string, map[string]string, string)
stderr: want (*zap.Logger, []string, map[string]string)
Error: stderr: framework/e2e/etcd_spawn.go:30:27: too many arguments in call to SpawnCmdWithLogger
stderr: have (*zap.Logger, []string, map[string]string, string)
stderr: want (*zap.Logger, []string, map[string]string)
Error: stderr: framework/e2e/etcd_spawn_cov.go:68:36: not enough arguments in call to expect.NewExpectWithEnv
stderr: have (string, []string, []string)
stderr: want (string, []string, []string, string)
FAIL: (code:2):
  % (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' '-tags=cov cluster_proxy' '-timeout' '30m' 'go.etcd.io/etcd/tests/v3/e2e')
FAIL	go.etcd.io/etcd/tests/v3/e2e [build failed]
FAIL
ERROR: Tests for following packages failed:
   go.etcd.io/etcd/tests/v3/e2e
mv: cannot stat '/home/runner/work/etcd/etcd/covdir/e2e_proxy/*': No such file or directory
```

Refer to https://github.com/etcd-io/etcd/runs/8049592009?check_suite_focus=true#logs

cc @serathius

PR #14395, opened 2022-08-29
[test] add the missing parameter 'name' into SpawnCmdWithLogger for cov test

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/14393 , which should be caused by https://github.com/etcd-io/etcd/commit/51933a7c8bc3fdb4ad60693b99cabe2df34b6127 

cc @serathius @spzala @tjungblu

### DIFF
--- a/tests/framework/e2e/etcd_spawn_cov.go
+++ b/tests/framework/e2e/etcd_spawn_cov.go
@@ -36,7 +36,7 @@ var (
 	coverDir = integration.MustAbsPath(os.Getenv("COVERDIR"))
 )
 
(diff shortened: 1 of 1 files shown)


serathius:
Ups sorry, missed that coverage also has build tags. Maybe we could compile tests in CI as part of build process?

serathius (approved):

ahrtr:
> Ups sorry, missed that coverage also has it's own build tag. Maybe we could compile tests in CI as part of build process?

Good idea to me!

ahrtr:
> > Ups sorry, missed that coverage also has it's own build tag. Maybe we could compile tests in CI as part of build process?
> 
> Good idea to me!

Let me take care of this in separate PR if you do not have bandwidth.

Issue #14045, opened 2022-05-16
etcd testing can reproduce historical data inconsistency issues


serathius:
P0 action item proposed in https://github.com/etcd-io/etcd/blob/main/Documentation/postmortems/v3.5-data-inconsistency.md 

Collecting ideas how we can improve testing to reproduce v3.5.0 data inconsistency issues in our automated testing.

xiaods:
suggest: build a testing team to guide more contribute to do testing on his environments

serathius:
@aphyr Based on recent issues you posted I expect that your are working on improving Etcd coverage with jepsen. Could you maybe shed some light on your plans so we can improve etcd official testing and qualification? 

In https://github.com/etcd-io/etcd/blob/main/Documentation/postmortems/v3.5-data-inconsistency.md you can find that current state of etcd release qualification needs a lot of work. I would like to collaborate on how we can ensure that there are no other remaining consistency issues in v3.5 release.

aphyr:
I'm afraid I can't talk too much about the present work, but I can say that the Jepsen etcd test suite has been updated and has found at least one safety issue I'd consider serious (#14092) and one spurious crash (#14110), as well as results that suggest etcd may not be performing its fsync/write dance properly (#14098). Given these findings, it might be worth running Jepsen on etcd releases regularly, or adapting some of its nemeses (e.g. kill, process, partition, disk corruption, losing un-fsynced writes) and checkers (e.g. https://github.com/jepsen-io/elle) to work with your current test suite. :-)

endocrimes:
👀 i'll update my jepsen fork and try to get those reproing to validate the setup.

serathius:
Thanks @aphyr for your recommendation. Etcd has it's own tests and framework (called `functional tests`) for failure injection testing (kill, partition etcd). However, as latest release has showed, it doesn't fulfill it's job. I think it's under invested and not maintained. 

I don't think etcd community should maintain their own framework as it's distracts from main goal of the project. Instead investing more in outdated closed off solutions, we should be able to utilize current state of the art that was build to solve similar problems.  I have some catching up to do in this area, so please free to give me suggestions about tools in this area. I will start from reading more about Jepsen and https://github.com/jepsen-io/elle. 

@endocrimes Thanks for help. I think it could be great if we could automate running jepsen tests. For example we could run it periodically once a week/day.

serathius:
[...]

PR #14398, opened 2022-08-29
Validate etcd linearizability

serathius:
First draft of linearizability tests that are able to reproduce https://github.com/etcd-io/etcd/issues/14370 within 20 seconds with 80% accuracy. Part of https://github.com/etcd-io/etcd/issues/14045

This approach uses a generic way to of verifing [linearizability](https://etcd.io/docs/v3.5/learning/api_guarantees/#isolation-level-and-consistency-of-replicas). In this proof of concept to reproduce https://github.com/etcd-io/etcd/issues/14370, however for full solution scenarios should be generated randomly based on preexisting fail points.

### DIFF
--- a/tests/framework/e2e/etcd_process.go
+++ b/tests/framework/e2e/etcd_process.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"syscall"
 	"testing"
 	"time"
 
@@ -38,12 +39,14 @@ type EtcdProcess interface {
 	EndpointsV3() []string
 	EndpointsMetrics() []string
 
+	Wait() error
 	Start(ctx context.Context) error
 	Restart(ctx context.Context) error
 	Stop() error
 	Close() error
 	Config() *EtcdServerProcessConfig
 	Logs() LogsExpect
+	Kill() error
 }
 
 type LogsExpect interface {
@@ -171,6 +174,22 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 		ep.cfg.lg.Panic("Please grab logs before process is stopped")
 	}
 	return ep.proc
+}
+
+func (ep *EtcdServerProcess) Kill() error {
+	ep.cfg.lg.Info("killing server...", zap.String("name", ep.cfg.Name))
+	return ep.proc.Signal(syscall.SIGKILL)
+}
+
+func (ep *EtcdServerProcess) Wait() error {
+	err := ep.proc.Wait()
+	if err != nil {
+		ep.cfg.lg.Error("failed to wait for server exit", zap.String("name", ep.cfg.Name))
+		return err
+	}
+	ep.cfg.lg.Info("server exited", zap.String("name", ep.cfg.Name))
+	ep.proc = nil
+	return nil
 }
 
 func AssertProcessLogs(t *testing.T, ep EtcdProcess, expectLog string) {
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -1,0 +1,87 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package linearizability
+
+import (
+	"context"
+	"time"
+
+	"github.com/anishathalye/porcupine"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
+)
+
+type recordingClient struct {
+	client clientv3.Client
+	id     int
+
+	operations []porcupine.Operation
+}
+
+func NewClient(endpoints []string, id int) (*recordingClient, error) {
+	cc, err := clientv3.New(clientv3.Config{
+		Endpoints:            endpoints,
+		Logger:               zap.NewNop(),
+		DialKeepAliveTime:    1 * time.Millisecond,
+		DialKeepAliveTimeout: 5 * time.Millisecond,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &recordingClient{
+		client:     *cc,
+		id:         id,
+		operations: []porcupine.Operation{},
+	}, nil
+}
+
+func (c *recordingClient) Close() error {
+	return c.client.Close()
+}
+
+func (c *recordingClient) Get(ctx context.Context, key string) error {
+	callTime := time.Now()
+	resp, err := c.client.Get(ctx, key)
+	returnTime := time.Now()
+	if err != nil {
+		return err
+	}
+	var readData string
+	if len(resp.Kvs) == 1 {
+		readData = string(resp.Kvs[0].Value)
+	}
+	c.operations = append(c.operations, porcupine.Operation{
+		ClientId: c.id,
+		Input:    etcdRequest{op: Get, key: key},
+		Call:     callTime.UnixNano(),
+		Output:   etcdResponse{getData: readData},
+		Return:   returnTime.UnixNano(),
+	})
+	return nil
+}
+
+func (c *recordingClient) Put(ctx context.Context, key, value string) error {
+	callTime := time.Now()
+	_, err := c.client.Put(ctx, key, value)
+	returnTime := time.Now()
+	c.operations = append(c.operations, porcupine.Operation{
+		ClientId: c.id,
+		Input:    etcdRequest{op: Put, key: key, putData: value},
+		Call:     callTime.UnixNano(),
+		Output:   etcdResponse{err: err},
+		Return:   returnTime.UnixNano(),
+	})
+	return nil
+}
--- a/tests/linearizability/failpoints.go
+++ b/tests/linearizability/failpoints.go
@@ -1,0 +1,49 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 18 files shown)


serathius:
@ahrtr @spzala I think this is ready to review. This PR provides first scenario to new framework of linerazibility test. Using this approach I was able to reproduce both data durability and data inconsistency issues. Further work should add more diverse scenarios to cover broader types of failures and allow dynamic scenarios based on random set of actions. 

Generating random scenarios should allow us to cover much broader space and should be our main goal. Current approach based on prepacked scenarios is limited to testing historical or simple scenarios. By testing a generic system property like linearizability we are no longer limited by simple `got` == `expect` validation. We need to start proactively finding issues instead of just responding to them post.

Next steps:
* Add scenario to reproduce https://github.com/etcd-io/etcd/issues/14370
  * Integrate gofail into linearizability tests 
  * Extend gofail to allow trigger multiple failpoints at once
  * Add Dynamic scenario based by triggering a random failpoint
*  Add scenario to reproduce https://github.com/etcd-io/etcd/issues/13766

tests/linearizability/model.go
```diff
@@ -0,0 +1,120 @@
...
+
+type etcdResponse struct {
+	getData string
+	err     error
+}
+
+type EtcdState struct {
+	Key          string
+	Value        string
+	FailedWrites map[string]bool
```
ahrtr, line 43: It seems that you do not use the value at all.

```suggestion
	FailedWrites map[string]struct{}
```

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/anishathalye/porcupine"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"golang.org/x/time/rate"
+)
+
+var (
```
ahrtr, line 30:
```suggestion
const (
```

tests/framework/e2e/etcd_process.go
```diff
@@ -173,6 +176,25 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 	return ep.proc
 }
 
+func (ep *EtcdServerProcess) Kill() error {
+	ep.cfg.lg.Info("killing server...", zap.String("name", ep.cfg.Name))
+	err := ep.proc.Signal(syscall.SIGKILL)
+	if err != nil {
+		return err
```
ahrtr, line 183: add an error log?

tests/framework/e2e/etcd_process.go
```diff
@@ -173,6 +176,25 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
...
+	}
+	return ep.Wait()
+}
+
+func (ep *EtcdServerProcess) Wait() error {
+	err := ep.proc.Wait()
+	if err != nil {
+		return err
```
ahrtr, line 191: add an error log?

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,172 @@
...
+func testLinearizability(t *testing.T, config e2e.EtcdProcessClusterConfig, failpoint FailpointConfig, traffic trafficConfig) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	clus, err := e2e.NewEtcdProcessCluster(ctx, t, &config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer clus.Close()
+	go func() {
+		defer cancel()
```
ahrtr, line 90: cancel is called two times? line 83 and line 90?

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,172 @@
...
+	limiter := rate.NewLimiter(rate.Limit(config.maximalQPS), 200)
+
+	startTime := time.Now()
+	wg := sync.WaitGroup{}
+	for i := 0; i < config.clientCount; i++ {
+		c, err := NewClient([]string{endpoints[i%len(endpoints)]}, i, startTime)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer c.Close()
```
ahrtr, line 144: This will not be called inside the loop.

spzala (approved):
@serathius great work! My only concern is that the`github.com/anishathalye/porcupine` seems developed/maintained by an individual, and there is not much to find out on dev history/PRs. But at the same time, we don't have other options and also that we are using it for test. So giving `porcupine` a try sounds okay. (Forking and maintaining our own copy may be too much of work specially when we don't have enough contributors so I guess we don't have that option.) I don't have any other comment besides I noticed a question from @ahrtr Thanks!

tests/linearizability/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/anishathalye/porcupine"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
+)
+
+type recordingClient struct {
+	client   clientv3.Client
+	id       int
+	baseTime time.Time
```
ahrtr, line 29: I don't think we need `baseTime`.  The [Operation.Call](https://github.com/anishathalye/porcupine/blob/7cbc8342681863b08e96432402d0588fc1defa88/model.go#L8) and [Operation.Return](https://github.com/anishathalye/porcupine/blob/7cbc8342681863b08e96432402d0588fc1defa88/model.go#L10) are just the timestamp when the operation is invoked and when it's returned, instead of duration.

tests/linearizability/client.go
```diff
@@ -0,0 +1,89 @@
...
+		return err
+	}
+	var readData string
+	if len(resp.Kvs) == 1 {
+		readData = string(resp.Kvs[0].Value)
+	}
+	c.operations = append(c.operations, porcupine.Operation{
+		ClientId: c.id,
+		Input:    etcdRequest{op: Get, key: key},
+		Call:     callTime.Sub(c.baseTime).Nanoseconds(),
```
ahrtr, line 70:
```suggestion
		Call:     callTime.Nanoseconds(),
```

tests/linearizability/client.go
```diff
@@ -0,0 +1,89 @@
...
+		Output:   etcdResponse{getData: readData},
+		Return:   returnTime.Sub(c.baseTime).Nanoseconds(),
```
ahrtr, line 72:
```suggestion
		Return:   returnTime.Nanoseconds(),
```

tests/linearizability/traffic.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+var (
+	PutGetTraffic Traffic = putGetTraffic{}
```
ahrtr, line 26: It doesn't make sense to define a global variable. We should either define a function something like `NewTraffic` or users use `putGetTraffic{}` directly.

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,183 @@
...
+			failpoint := FailpointConfig{
+				failpoint:           tc.failpoint,
+				count:               failpointTriggersCount,
+				waitBetweenTriggers: waitBetweenFailpointTriggers,
+			}
+			traffic := trafficConfig{
+				minimalQPS:  minimalQPS,
+				maximalQPS:  maximalQPS,
+				clientCount: 8,
+				traffic:     PutGetTraffic,
```
ahrtr, line 75:
```suggestion
				traffic:     putGetTraffic{},
```

tests/linearizability/failpoints.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+	"math/rand"
+
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+var (
+	KillFailpoint Failpoint = killFailpoint{}
```
ahrtr, line 25: It doesn't make sense to define a global variable. Please see the comment for `PutGetTraffic`.

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,183 @@
...
+func TestLinearizability(t *testing.T) {
+	testRunner.BeforeTest(t)
+	tcs := []struct {
+		name      string
+		failpoint Failpoint
+		config    e2e.EtcdProcessClusterConfig
+	}{
+		{
+			name:      "KillClusterOfSize1",
+			failpoint: KillFailpoint,
```
ahrtr, line 51:
```suggestion
			failpoint: killFailpoint{},
```

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,183 @@
...
+			config: e2e.EtcdProcessClusterConfig{
+				ClusterSize: 1,
+			},
+		},
+		{
+			name:      "KillClusterOfSize3",
+			failpoint: KillFailpoint,
```
ahrtr, line 58:
```suggestion
			failpoint: killFailpoint{},
```

tests/linearizability/linearizability_test.go
```diff
@@ -0,0 +1,183 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			failpoint := FailpointConfig{
+				failpoint:           tc.failpoint,
+				count:               failpointTriggersCount,
+				waitBetweenTriggers: waitBetweenFailpointTriggers,
+			}
+			traffic := trafficConfig{
+				minimalQPS:  minimalQPS,
+				maximalQPS:  maximalQPS,
+				clientCount: 8,
```
ahrtr, line 74: minor comment: please consider to make this configurable, such as adding a field `clientCount` into the struct.

ahrtr:
Overall looks good to me with some minor comments.  Great work!



> My only concern is that the`github.com/anishathalye/porcupine` seems developed/maintained by an individual, and there is not much to find out on dev history/PRs

Yes, I have the same concern, especially if we want to spend more resource/time on the linearizability test. We can discuss this separately.

tests/linearizability/traffic.go
serathius: I want to maintain a single clean place to that lists are available traffic and failpoints. I find it more clear than using structs and asking user to scan whole file where they are scattered.

tests/linearizability/linearizability_test.go
serathius: Struct `tcs` is not set in stone, we can always add it when we decide to parameterize tests based on it.

ahrtr (approved):
Overall looks good to me.

PR #14399, opened 2022-08-29
recipe: pass context as argument in RWMutex

goodbaikin:
recipe: pass context as argument in RWMutex

We have a context.Context inside the struct RWMutex. It's hardcoded as context.TODO() and reused in every function calls.
https://github.com/etcd-io/etcd/blob/77773c7798e4e89497f59ecc4bdb54c4ab2aa1d1/client/v3/experimental/recipes/rwmutex.go#L34

This shoud be able to passed by users as the Mutex in concurrency
 package does.
https://github.com/etcd-io/etcd/blob/77773c7798e4e89497f59ecc4bdb54c4ab2aa1d1/client/v3/concurrency/mutex.go#L71

### DIFF
--- a/client/v3/experimental/recipes/rwmutex.go
+++ b/client/v3/experimental/recipes/rwmutex.go
@@ -23,18 +23,17 @@ import (
 )
 
 type RWMutex struct {
-	s   *concurrency.Session
-	ctx context.Context
+	s *concurrency.Session
 
 	pfx   string
 	myKey *EphemeralKV
 }
 
 func NewRWMutex(s *concurrency.Session, prefix string) *RWMutex {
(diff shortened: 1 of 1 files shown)


goodbaikin:
I noticed that the waitDelete() in concurrency package handles ctx.Err(), while the RWMutex uses WaitEvents() which does not.

https://github.com/etcd-io/etcd/blob/cdd2b737f04812a919a5735380fdaa1f932346d0/client/v3/concurrency/key.go#L42
https://github.com/etcd-io/etcd/blob/cdd2b737f04812a919a5735380fdaa1f932346d0/client/v3/experimental/recipes/watch.go#L25

So passing ctx has almost no meaning other than Get requests.
I'll close thie issue.

PR #14404, opened 2022-08-30
Added member interaction into EtcdProcessCluster

biosvs:
As I promised in https://github.com/etcd-io/etcd/pull/14354, trying to introduce to EtcdProcessCluster functions for changing cluster member list.

As an example rewrote e2e tests from mentioned pull request.

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -28,6 +28,8 @@ import (
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	"go.etcd.io/etcd/server/v3/etcdserver"
 )
 
@@ -136,12 +138,14 @@ func ConfigStandalone(cfg EtcdProcessClusterConfig) *EtcdProcessClusterConfig {
 }
 
 type EtcdProcessCluster struct {
-	lg    *zap.Logger
-	Cfg   *EtcdProcessClusterConfig
-	Procs []EtcdProcess
+	lg      *zap.Logger
+	Cfg     *EtcdProcessClusterConfig
+	Procs   []EtcdProcess
+	nextSeq int // sequence number of the next etcd process (if it will be required)
 }
 
 type EtcdProcessClusterConfig struct {
+	Logger      *zap.Logger
 	ExecPath    string
 	DataDirPath string
 	KeepDataDir bool
@@ -205,11 +209,14 @@ func NewEtcdProcessCluster(ctx context.Context, t testing.TB, cfg *EtcdProcessCl
 func InitEtcdProcessCluster(t testing.TB, cfg *EtcdProcessClusterConfig) (*EtcdProcessCluster, error) {
 	SkipInShortMode(t)
 
-	etcdCfgs := cfg.EtcdServerProcessConfigs(t)
+	cfg.InitBaseValues(t)
+
+	etcdCfgs := cfg.EtcdAllServerProcessConfigs(t)
 	epc := &EtcdProcessCluster{
-		Cfg:   cfg,
-		lg:    zaptest.NewLogger(t),
-		Procs: make([]EtcdProcess, cfg.ClusterSize),
+		Cfg:     cfg,
+		lg:      zaptest.NewLogger(t),
+		Procs:   make([]EtcdProcess, cfg.ClusterSize),
+		nextSeq: cfg.ClusterSize,
 	}
 
 	// launch etcd processes
@@ -258,9 +265,10 @@ func (cfg *EtcdProcessClusterConfig) PeerScheme() string {
 	return peerScheme
 }
 
-func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfigs(tb testing.TB) []*EtcdServerProcessConfig {
-	lg := zaptest.NewLogger(tb)
-
+func (cfg *EtcdProcessClusterConfig) InitBaseValues(tb testing.TB) {
+	if cfg.Logger == nil {
+		cfg.Logger = zaptest.NewLogger(tb)
+	}
 	if cfg.BasePort == 0 {
 		cfg.BasePort = EtcdProcessBasePort
 	}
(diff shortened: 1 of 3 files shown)


ahrtr:
Please resolve the conflict.

biosvs:
@ahrtr done, rebased on main

tests/framework/e2e/cluster.go
```diff
@@ -516,6 +611,10 @@ func (epc *EtcdProcessCluster) Stop() (err error) {
 	return err
 }
 
+func (epc *EtcdProcessCluster) CtlClient() *EtcdctlV3 {
```
serathius, line 614:
```suggestion
func (epc *EtcdProcessCluster) Client() *EtcdctlV3 {
```

tests/framework/e2e/etcd_process.go
```diff
@@ -183,6 +184,17 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
...
+func (cfg *EtcdServerProcessConfig) SetInitialCluster(nodes []string, initialClusterState string) {
+	cfg.InitialCluster = strings.Join(nodes, ",")
+	cfg.Args = append(cfg.Args, "--initial-cluster", cfg.InitialCluster)
+	cfg.Args = append(cfg.Args, "--initial-cluster-state", initialClusterState)
+}
+
+func (cfg *EtcdServerProcessConfig) EnableDiscovery(token string, endpoints []string) {
+	cfg.Args = append(cfg.Args, fmt.Sprintf("--discovery-token=%s", token))
+	cfg.Args = append(cfg.Args, fmt.Sprintf("--discovery-endpoints=%s", strings.Join(endpoints, ",")))
+}
```
serathius, line 196: let's not implement setters for configuration structs. I think we should avoid mutating args when constructing EtcdServerProcess. One solution would be to separate args provided by caller (`extraArgs`) from args generated from configuration fields (`configArgs`).

serathius (changes requested):
Overall I'm not big fan of complicating `EtcdProcessCluster` code. From brief look at the code looks like you implement cluster resizes just for this test. I think dynamic cluster sizes over-complicate the code and are not worth it. 

Instead of complicating library used by hundreds of tests, could we make just your test simpler?

serathius:
For me problem is making list of members dynamic, which would be only used for this test. Could we maybe check if the test itself can be simpler? For testing `endpoints-auto-sync-interval` do we really need to start a whole etcd process? 

My understanding is that we are just looking for a `Resolver state updated` log in proxy. I think that just calling a `MemberAdd` on cluster should be enough.

biosvs:
Hmm, can't agree here.
Even using logs for making progress in tests make me uncomfortable, honestly. But I definitely would not say that something is working based on logs. I prefer to think that "it doesn't work until you run it explicitly and get expected result".

ahrtr:
I do not get time to have a deeper review on this PR, but my immediate feeling is it's a little over complicated. 

The gap mentioned in https://github.com/etcd-io/etcd/pull/14354#discussion_r953155400 is that the current test framework has interface [MemberAdd](https://github.com/etcd-io/etcd/blob/0e1c895108edbd7d5d8d8b37a40dc8b833b2b5c8/tests/framework/interface.go#L75), but doesn't have interface method something like `MemberLaunch` or `LaunchMember` or `StartMember` or `MemberStart`. I think we do need to support dynamically changing the member in the framework. 

Please always keep OCP (Open Close principle) in mind: Open for extension, but close for modification. When updating existing framework, please try not to change the existing logic. Instead, please try to make your new code adapt to the existing framework.

biosvs:
I'm afraid we're stuck with this PR (because we're in a kinda closed loop).

Changed e2e test cannot be rewritten with common or integration framework, because we're testing interaction through different running instances (including updates via network).

The same story is here: https://github.com/etcd-io/etcd/blob/main/tests/e2e/ctl_v3_snapshot_test.go#L258

Both these tests need actual member list changes. Both tests do it via raw commands. I tried to add these commands to e2e framework, but you think it's unnecessary. You're contributors, and if you're saying that it's easer to use raw commands for couple of tests - let it be.

tests/framework/e2e/etcd_process.go
ahrtr, line 196: These two methods are only used by `SetInitialOrDiscovery`, it seems that we can just get the code included in it directly instead of creating two separate methods.

tests/framework/e2e/cluster.go
```diff
@@ -136,12 +138,14 @@ func ConfigStandalone(cfg EtcdProcessClusterConfig) *EtcdProcessClusterConfig {
 }
 
 type EtcdProcessCluster struct {
-	lg    *zap.Logger
-	Cfg   *EtcdProcessClusterConfig
-	Procs []EtcdProcess
+	lg      *zap.Logger
+	Cfg     *EtcdProcessClusterConfig
+	Procs   []EtcdProcess
+	nextSeq int // sequence number of the next etcd process (if it will be required)
```
ahrtr, line 144: No need to add `nextSeq`, you can just use `Cfg.ClusterSize` or` len(epc.Procs)`

ahrtr:
When I tried to enhance the existing e2e test framework, eventually it's similar to this PR. But the interface definition looks clearer, I also get rid of the `nextSeq`. Please consider to enhance this PR per https://github.com/ahrtr/etcd/commit/9649fd7538cecaaabdef88aa159129629262da96

biosvs:
Maybe I didn't get the idea, but code looks broken for me:

1. Cluster state is omitted
It's necessary to pass "initial-cluster-state existing" when new member is added (I think that was the reason to add `SetInitialCluster` to EtcdServerProcessConfig).

2. nextSeq
With "epc.Cfg.ClusterSize", you'll try to reuse occupied ports in case of this sequence: 
- Start cluster of >1 nodes
- Remove member (with least ports number)
- Add member (it will try to use the same ports as the member with highest ports number)

ahrtr:
The commit is just to demonstrate the high level idea instead of a PR ready for review.

ahrtr:
> 2\. nextSeq
>         With "epc.Cfg.ClusterSize", you'll try to reuse occupied ports in case of this sequence:

The initial value for `nextSq` is `cfg.ClusterSize`, and incremental by 1 each time calling `StartNewProc`. I don't understand why it can't be removed and use `cfg.ClusterSize` or `len(epc.Procs)` directly.  You just need to update `cfg.ClusterSize` and `epc.Procs` each time after calling `StartNewProc`.

biosvs:
Let's assume that initial configuration is:
- base port 2000
- 2 nodes

Then node A has port 2000 + 1*5 = 2005, node B has port 2000 + 2*5 = 2010.

Now imaging that I'm first removing node A and then adding new node C. What will I get?

If I use cfg.ClusterSize or len(epc.Procs), I'll get port for new node = 2000 + 2*5 = 2010, which is overlapping with already running node B.
That's why I need some counter that doesn't depend on the current state of cluster. That's why `nextSq` was introduced.

ahrtr:
> Let's assume that initial configuration is:
> 
>     * base port 2000
> 
>     * 2 nodes
> 
> 
> Then node A has port 2000 + 1_5 = 2005, node B has port 2000 + 2_5 = 2010.
> 
> Now imaging that I'm first removing node A and then adding new node C. What will I get?
> 
> If I use cfg.ClusterSize or len(epc.Procs), I'll get port for new node = 2000 + 2*5 = 2010, which is overlapping with already running node B. That's why I need some counter that doesn't depend on the current state of cluster. That's why `nextSq` was introduced.

Make sense to me, thanks for the clarification.

ahrtr:
@biosvs Could you resolve the inline comments?

tests/framework/e2e/etcd_process.go
ahrtr: @biosvs  After you resolve https://github.com/etcd-io/etcd/pull/14404#discussion_r969102907, then this comment should can be resolved automatically.

ahrtr:
Continue to work on this PR in https://github.com/etcd-io/etcd/pull/14589

PR #14407, opened 2022-08-31
[Third solution] Fix the potential data loss for clusters with only one member (raft layer change)

ahrtr:
Third solution to fix https://github.com/etcd-io/etcd/issues/14370 

For a cluster with only one member, the raft always send identical
unstable entries and committed entries to etcdserver, and etcd
responds to the client once it finishes (actually partially) the
applying workflow.

When the client receives the response, it doesn't mean etcd has already
successfully saved the data, including BoltDB and WAL, because:
   1. etcd commits the boltDB transaction periodically instead of on each request;
   2. etcd saves WAL entries in parallel with applying the committed entries.
Accordingly, it may run into a situation of data loss when the etcd crashes
immediately after responding to the client and before the boltDB and WAL
successfully save the data to disk.
Note that this issue can only happen for clusters with only one member.

For clusters with multiple members, it isn't an issue, because etcd will
not commit & apply the data before it being replicated to majority members.
When the client receives the response, it means the data must have been applied.
It further means the data must have been committed.
Note: for clusters with multiple members, the raft will never send identical
unstable entries and committed entries to etcdserver.

### DIFF
--- a/raft/node.go
+++ b/raft/node.go
@@ -87,6 +87,12 @@ type Ready struct {
 	// MustSync indicates whether the HardState and Entries must be synchronously
 	// written to disk or if an asynchronous write is permissible.
 	MustSync bool
+
+	// MustSaveEntriesBeforeApply indicates whether the Entries must be
+	// synchronously saved to disk before applying the CommittedEntries.
+	// Note that Entries and CommittedEntries may have overlapped entries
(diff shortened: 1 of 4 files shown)


ahrtr:
Closing this PR as we agreed that we will not proceed with this PR.

PR #14411, opened 2022-09-01
[Fourth solution] Fix the potential data loss for clusters with only one member (raft layer change)

ahrtr:
The fourth solution to fix https://github.com/etcd-io/etcd/issues/14370

### DIFF
--- a/raft/node.go
+++ b/raft/node.go
@@ -267,6 +267,7 @@ type node struct {
 	status     chan chan Status
 
 	rn *RawNode
+	rd Ready
 }
 
 func newNode(rn *RawNode) node {
@@ -304,7 +305,6 @@ func (n *node) run() {
 	var propc chan msgWithResult
 	var readyc chan Ready
 	var advancec chan struct{}
-	var rd Ready
 
 	r := n.rn.raft
 
@@ -322,7 +322,7 @@ func (n *node) run() {
 			// handled first, but it's generally good to emit larger Readys plus
 			// it simplifies testing (by emitting less frequently and more
 			// predictably).
-			rd = n.rn.readyWithoutAccept()
+			n.rd = n.rn.readyWithoutAccept()
 			readyc = n.readyc
 		}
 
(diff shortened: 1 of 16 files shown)


raft/node.go
```diff
@@ -503,6 +503,17 @@ func (n *node) stepWithWaitOption(ctx context.Context, m pb.Message, wait bool)
 func (n *node) Ready() <-chan Ready { return n.readyc }
 
 func (n *node) Advance() {
+	ents := n.rd.Entries
+
+	// Provide a feedback that the `Entries` has already been successfully
+	// saved by the user. Note that only leader needs to do this, because
+	// a follower will not send the feedback to leader until it finishes
+	// saving the `Entries`.
+	if n.isLeader() && (len(ents) > 1 || (len(ents) == 1 && len(ents[0].Data) > 0)) {
```
ptabor, line 512: Could you, please describe why len(ents) >= 1  would be not sufficient here,
it is, what case and why we want to filter out here.

ptabor (approved):
Thank you. This change looks good to me.
The fixed tests are huge value, irregardless whether we will go with Step or the in-line approach.

raft/node.go
ahrtr: Good point, I was planning to add more comment as well. Will update soon. thx

ahrtr: // An empty entry is generated in the following two cases:
	//    1. A new leader is elected.
	//    2. enter into joint configuration.
	// We need to ignore above two cases because they are already committed. Note
	// that it's safe to commit the empty entry immediately after appending the entry
	// because it can only happen for multi-member cluster.

ahrtr: Updated.

@ptabor FYI. https://github.com/ahrtr/etcd/blob/a79941679f401186dc29e10d4e86a53590c3d9d6/raft/raft.go#L643-L645

ahrtr: Just updated the comment to avoid confusion,

```
	// An empty entry is generated in the following two cases:
	//    1. a new leader is elected.
	//    2. enter into joint configuration.
	// We need to ignore above two cases because the `pr.Match` is updated
	// immediately after appending the empty entry. Note that it's safe to do
	// so because it can only happen for multi-member cluster. Please also refer to
	// https://github.com/ahrtr/etcd/blob/a79941679f401186dc29e10d4e86a53590c3d9d6/raft/raft.go#L643-L645
```

ahrtr:
leave it to @tbg to fix the raft side in https://github.com/etcd-io/etcd/pull/14413

PR #14400, opened 2022-08-30
[Second Solution] Fix the potential data loss for clusters with only one member (simpler solution)

ahrtr:
Second solution to fix https://github.com/etcd-io/etcd/issues/14370

This solution is based on the following feedback,
1. https://github.com/etcd-io/etcd/issues/14370#issuecomment-1229185373 from @hasethuraman
2. https://github.com/etcd-io/etcd/pull/14394#discussion_r956781047 from  @lavacat
3. https://github.com/etcd-io/etcd/issues/14370#issuecomment-1230398569  from @ptabor 

I compared the performance between this PR and https://github.com/etcd-io/etcd/pull/14394 for one-member cluster , overall https://github.com/etcd-io/etcd/pull/14394 is a little better than this one (about 2.7% higher than this one). But this PR is much simpler; excluding the test and comment, this PR only has about 20 lines of code change. 

cc @serathius @spzala @ptabor @liggitt @dims

Issue #12257, opened 2020-08-25
Proposal: support fully control fsync frequency in raft


BusyJay:
# Support fully control fsync frequency in raft

## Summary

Let application decides when to sync logs and make leader aware of the safe index
to commit.

## Motivation

The current way of driving raft is a loop of following process
```
1. get ready from raft
2. persist ready
3. advance ready in raft
```

Only one ready is allowed to be processed at a time, and the leader expects all pending logs in the ready are synced to disk before handling next one.

The mechanism is simple but has several problems:
1. it can cause high fsync frequency. In the practice of TiKV, we observed the fsync frequency can reach the limit of hardware very easily, which can cause unstable latency and hurt performance. Also high fsync frequency is also expensive in the cloud.
2. sync size is unpredictable, which can waste IO when syncing small data size.

So we need a way to control the fsync frequency without hurting correctness and also get maximum performance.

## Detailed design

Since all nodes are communicating with messages, so application can stash all messages and send them out only after logs are synced. In this way, application is free to fetch and advance as many readiness as they want and sync whenever they want.

It works great for followers. However, for leaders, both Etcd and TiKV have follow the raft thesis 10.2.1 that sends out messages of leader before syncing to make leader and follower write in parallel. If leader batches up more than two readiness, it can commit an un-synced logs accidentally. For example, if leader has logs 3, 4 in the first ready, after advancing and receiving followers' ACK, it can broadcast commit to all followers in the second ready. This is because leader assumes logs are synced in the first ready. To fix the problem, we need to stop updating leader's progress until logs are synced to disk. So when leader calculates commit index only those really sync progress are considered.

It's possible quorum followers sync logs before leader, in such case, leader is also safe to consider the logs are committed, but it should not mark them ready to apply until logs are synced, otherwise it can corrupt the state between raft and application.

Note that batching changes to hard state doesn't affect the correctness. Changes to term and vote can only happen when the node is follower. So messages will not be sent out before changes are synced, which is same as before. Changes to commit can happen when the node is leader, however, commit index is not required to be store at all according to raft thesis 3.8.

To conclude, we need to make those changes to raft library:
1. Don't update leader's progress when logs are appended,
2. Add a method to allow application inform raft about the synced index, which should change its progress and trigger logs commitment.

All others are left to applications to guarantee choose different processing logic for different roles. 

## Drawbacks

It's a little complicated for applications to do the batch.

## Alternatives

[...]

PR #14413, opened 2022-09-01
raft: don't emit unstable CommittedEntries

tbg:
Fixes https://github.com/etcd-io/etcd/issues/14370.

When run in a single-voter configuration, prior to this PR
raft would emit `HardState`s that would emit a proposed `Entry`
simultaneously in `CommittedEntries` and `Entries`.

To be correct, this requires users of the raft library to enforce an
ordering between appending to the log and notifying the client about
`CommittedEntries` also present in `Entries`. This was easy to miss.

Walk back this behavior to arrive at a simpler contract: what's
emitted in `CommittedEntries` is truly committed, i.e. present
in stable storage on a quorum of voters.

This in turn pessimizes the single-voter case: rather than fully
handling an `Entry` in just one `Ready`, now two are required,
and in particular one has to do extra work to save on allocations.

We accept this as a good tradeoff, since raft primarily serves
multi-voter configurations.

----

Suggested review plan:

- Look at `raft: don't emit unstable CommittedEntries` first to get an idea of the gist of the change
- Look at `raft: directly update leader in advance` for an alternative solution that "inlines" the relevant bits of `r.Step` instead.
-  The test suites pass with either solution, so the two solutions are functionally equivalent to the degree that we're exercising them.
- Look at `raft: add BenchmarkRawNode` and at the benchmark results in `raft: benchmark results`
- If there are substantial concerns and/or comments, stop here and sort them out with me.
- Then slog through the remaining commits, which are all going to be squashed before the merge, and fix up the tests one-by-one.
- Note: `raft: always mark leader as RecentActive` was included here since otherwise the outputs of the two possible solutions differ in sometimes failing to track the leader as recently active.

### DIFF
--- a/raft/diff_test.go
+++ b/raft/diff_test.go
@@ -55,7 +55,7 @@ func mustTemp(pre, body string) string {
 }
 
 func ltoa(l *raftLog) string {
-	s := fmt.Sprintf("committed: %d\n", l.committed)
+	s := fmt.Sprintf("lastIndex: %d\n", l.lastIndex())
 	s += fmt.Sprintf("applied:  %d\n", l.applied)
 	for i, e := range l.allEntries() {
 		s += fmt.Sprintf("#%d: %+v\n", i, e)
--- a/raft/node_test.go
+++ b/raft/node_test.go
@@ -178,6 +178,10 @@ func TestNodePropose(t *testing.T) {
 func TestNodeReadIndex(t *testing.T) {
 	var msgs []raftpb.Message
 	appendStep := func(r *raft, m raftpb.Message) error {
+		if m.Type == raftpb.MsgAppResp {
+			// See (*raft).advance.
+			return nil
+		}
 		msgs = append(msgs, m)
 		return nil
 	}
--- a/raft/node_test.go
+++ b/raft/node_test.go
@@ -173,59 +173,6 @@ func TestNodePropose(t *testing.T) {
 	}
 }
 
-// TestNodeReadIndex ensures that node.ReadIndex sends the MsgReadIndex message to the underlying raft.
-// It also ensures that ReadState can be read out through ready chan.
-func TestNodeReadIndex(t *testing.T) {
-	var msgs []raftpb.Message
-	appendStep := func(r *raft, m raftpb.Message) error {
-		if m.Type == raftpb.MsgAppResp {
-			// See (*raft).advance.
-			return nil
-		}
-		msgs = append(msgs, m)
-		return nil
-	}
-	wrs := []ReadState{{Index: uint64(1), RequestCtx: []byte("somedata")}}
-
-	s := newTestMemoryStorage(withPeers(1))
-	rn := newTestRawNode(1, 10, 1, s)
-	n := newNode(rn)
-	r := rn.raft
-	r.readStates = wrs
-
-	go n.run()
-	n.Campaign(context.TODO())
-	for {
-		rd := <-n.Ready()
-		if !reflect.DeepEqual(rd.ReadStates, wrs) {
-			t.Errorf("ReadStates = %v, want %v", rd.ReadStates, wrs)
-		}
-
-		s.Append(rd.Entries)
-
-		if rd.SoftState.Lead == r.id {
-			n.Advance()
-			break
-		}
-		n.Advance()
-	}
-
-	r.step = appendStep
-	wrequestCtx := []byte("somedata2")
-	n.ReadIndex(context.TODO(), wrequestCtx)
-	n.Stop()
-
-	if len(msgs) != 1 {
-		t.Fatalf("len(msgs) = %d, want %d", len(msgs), 1)
-	}
-	if msgs[0].Type != raftpb.MsgReadIndex {
-		t.Errorf("msg type = %d, want %d", msgs[0].Type, raftpb.MsgReadIndex)
-	}
-	if !bytes.Equal(msgs[0].Entries[0].Data, wrequestCtx) {
-		t.Errorf("data = %v, want %v", msgs[0].Entries[0].Data, wrequestCtx)
-	}
-}
-
 // TestDisableProposalForwarding ensures that proposals are not forwarded to
 // the leader when DisableProposalForwarding is true.
 func TestDisableProposalForwarding(t *testing.T) {
--- a/raft/node_test.go
+++ b/raft/node_test.go
@@ -317,7 +317,7 @@ func TestNodeProposeConfig(t *testing.T) {
 func TestNodeProposeAddDuplicateNode(t *testing.T) {
 	s := newTestMemoryStorage(withPeers(1))
 	cfg := newTestConfig(1, 10, 1, s)
-	ctx, cancel, n := newNodeTestHarness(t, context.Background(), cfg)
+	ctx, cancel, n := newNodeTestHarness(context.Background(), t, cfg)
 	defer cancel()
 	n.Campaign(ctx)
 	allCommittedEntries := make([]raftpb.Entry, 0)
@@ -590,7 +590,7 @@ func TestNodeStart(t *testing.T) {
 		MaxInflightMsgs: 256,
 	}
 	n := StartNode(c, []Peer{{ID: 1}})
-	ctx, cancel, n := newNodeTestHarness(t, context.Background(), c, Peer{ID: 1})
+	ctx, cancel, n := newNodeTestHarness(context.Background(), t, c, Peer{ID: 1})
 	defer cancel()
 
 	{
@@ -740,7 +740,7 @@ func TestNodeAdvance(t *testing.T) {
 		MaxSizePerMsg:   noLimit,
 		MaxInflightMsgs: 256,
 	}
-	ctx, cancel, n := newNodeTestHarness(t, context.Background(), c)
+	ctx, cancel, n := newNodeTestHarness(context.Background(), t, c)
 	defer cancel()
 
 	n.Campaign(ctx)
@@ -895,7 +895,7 @@ func TestCommitPagination(t *testing.T) {
 	s := newTestMemoryStorage(withPeers(1))
 	cfg := newTestConfig(1, 10, 1, s)
 	cfg.MaxCommittedSizePerReady = 2048
-	ctx, cancel, n := newNodeTestHarness(t, context.Background(), cfg)
+	ctx, cancel, n := newNodeTestHarness(context.Background(), t, cfg)
 	defer cancel()
 	n.Campaign(ctx)
 
--- a/raft/node_test.go
+++ b/raft/node_test.go
@@ -479,6 +479,10 @@ func TestNodeProposeWaitDropped(t *testing.T) {
 			t.Logf("dropping message: %v", m.String())
 			return ErrProposalDropped
 		}
+		if m.Type == raftpb.MsgAppResp {
+			// This is produced by raft internally, see (*raft).advance.
+			return nil
+		}
 		msgs = append(msgs, m)
 		return nil
 	}
@@ -511,7 +515,7 @@ func TestNodeProposeWaitDropped(t *testing.T) {
 
 	n.Stop()
 	if len(msgs) != 0 {
-		t.Fatalf("len(msgs) = %d, want %d", len(msgs), 1)
+		t.Fatalf("len(msgs) = %d, want %d", len(msgs), 0)
 	}
 }
 
--- a/raft/node_test.go
+++ b/raft/node_test.go
@@ -914,15 +914,14 @@ func TestCommitPagination(t *testing.T) {
 	s := newTestMemoryStorage(withPeers(1))
 	cfg := newTestConfig(1, 10, 1, s)
 	cfg.MaxCommittedSizePerReady = 2048
-	rn, err := NewRawNode(cfg)
-	if err != nil {
-		t.Fatal(err)
-	}
-	n := newNode(rn)
-	go n.run()
-	n.Campaign(context.TODO())
+	ctx, cancel, n := newNodeTestHarness(t, context.Background(), cfg)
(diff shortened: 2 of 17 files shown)


serathius:
Rerun the e2e flake. This PR is still failing DCO, please run `git rebase origin/main --signoff` and push head with force `git push origin HEAD:raft-single-voter -f`.

raft/raft.go
```diff
@@ -635,9 +648,6 @@ func (r *raft) appendEntry(es ...pb.Entry) (accepted bool) {
 	}
 	// use latest "last" index after truncate/append
 	li = r.raftLog.append(es...)
```
nvanbenschoten, line 650: nit: no need to update `li`.

nvanbenschoten (commented):
This change makes sense to me. It feels like the "proper" solution because it simplifies the `Ready` contract and avoids the need for special-casing of single-peer deployments. After this change, entries in the `CommittedEntries` slice are always committed.

I actually made very similar changes in a prototype (https://github.com/nvanbenschoten/etcd/commit/1d1fa32ec22715f6e21a6d6486f865356c3eb4fc) intended to address https://github.com/etcd-io/etcd/issues/12257 / https://github.com/cockroachdb/cockroach/issues/17500. These kinds of extensions will be easier to make if the Raft leader doesn't immediately consider an entry to be durable upon appending to its own unstable log (in `raft.appendEntry`). Things are cleaner if, instead, the durability status is maintained either implicitly in `raft.advance` or explicitly through a different (opt-in) API if clients want the flexibility to sync their log asynchronously.

From the perspective of CRDB's use of Raft, we weren't susceptible to the bug in #14370 because log application always came after log appending in the Ready processing state machine loop. However, we were aware of this idiosyncrasy and had to work around it when acking proposals ahead of application, see: https://github.com/cockroachdb/cockroach/blob/ce55e1b46604401f6085a6ece1355970da94600a/pkg/kv/kvserver/replica_raft.go#L834-L844

The comment there is interesting because it references a second case where `Entries` and `CommittedEntries` can overlap — that being when a follower in a multi-node replication group is catching up after falling behind. In those cases, the overlapping entries are already committed, so there's no durability risk from applying CommittedEntries before appending Entries (like `etcdserver` does). However, that still means that it's possible to have applied an entry that is not durably stored locally in the Raft log. I don't know if that could cause issues for etcd or whether entry application is idempotent.

raft/raft.go
ahrtr: +1

raft/storage.go
```diff
@@ -83,6 +87,8 @@ type MemoryStorage struct {
 	snapshot  pb.Snapshot
 	// ents[i] has raft log position i+snapshot.Metadata.Index
 	ents []pb.Entry
+
+	callStats inMemStorageCallStats
```
ahrtr, line 91: I do not see where `callStats` is used. Are you planning to use it in a separate PR?

raft/node_util_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2015 The etcd Authors
```
ahrtr, line 1: 2022?

raft/storage.go
ahrtr: It seems also an unrelated change, so suggest to do it in a separate PR.

ahrtr: Got it, it's used in the Benchmark test cases.

raft/rawnode_test.go
```diff
@@ -1105,3 +1131,90 @@ func TestRawNodeConsumeReady(t *testing.T) {
...
 	}
 }
+
+func BenchmarkRawNode(b *testing.B) {
+	b.Run("single-voter", func(b *testing.B) {
+		benchmarkRawNodeImpl(b, 1)
+	})
+	b.Run("two-voters", func(b *testing.B) {
+		benchmarkRawNodeImpl(b, 1, 2)
+	})
```
ahrtr, line 1141:
```suggestion
	cases := []struct{
	    name string
	    peers []uint64
	 } {
	     {
	         name: "single-voter",
	         peers: []uint64{1},
	     },
	     {
	         name: "two-voters",
	         peers: []uint64{1, 2},
	     },
	     // You can easily add more cases here
	 }
	 
	 for _, tc := range cases {
	     b.Run(tc.name, func(b *testing.B) {
	         benchmarkRawNodeImpl(b, tc.peers)
	     }
	 }
```

raft/diff_test.go
```diff
@@ -55,7 +55,7 @@ func mustTemp(pre, body string) string {
 }
 
 func ltoa(l *raftLog) string {
-	s := fmt.Sprintf("committed: %d\n", l.committed)
+	s := fmt.Sprintf("lastIndex: %d\n", l.lastIndex())
```
ahrtr, line 58: Why do you change this?

raft/node_util_test.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+)
+
+type nodeTestHarness struct {
+	*node
+	t *testing.T
```
ahrtr, line 26:
```suggestion
	*testing.T
```

ahrtr: After this change, you can remove all the following methods (e.g. Debug(f), Error(f), Info(f), Warning(f), Fatal(f), Panic(f)

raft/node_util_test.go
```diff
@@ -0,0 +1,109 @@
...
+		n = setupNode(cfg, peers)
+	} else {
+		rn, err := NewRawNode(cfg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		nn := newNode(rn)
+		n = &nn
+	}
+	go func() {
```
ahrtr, line 98: There is a potential goroutine leak. When the test case finishes, it may still run. We need to make sure gracefully shutdown for every goroutine.

raft/diff_test.go
ahrtr: I am curious why this PR is much bigger than https://github.com/etcd-io/etcd/pull/14411, since they are almost following the same logic ? I am not sure whether it's related to this change?

raft/node_test.go
```diff
@@ -163,55 +173,6 @@ func TestNodePropose(t *testing.T) {
 	}
 }
 
-// TestNodeReadIndex ensures that node.ReadIndex sends the MsgReadIndex message to the underlying raft.
-// It also ensures that ReadState can be read out through ready chan.
-func TestNodeReadIndex(t *testing.T) {
```
ahrtr, line 168: Why did you remove this case ?

ahrtr:
The production code change looks simple & good, but the test code change is too big. This PR almost follows the same logic as https://github.com/etcd-io/etcd/pull/14411, but this PR is much bigger, even excluding the new benchmark test case. 

Another comment... Have you compared the performance as compared to the `main` branch using the [benchmark](https://github.com/etcd-io/etcd/tree/main/tools/benchmark) tool? I suggest to compare both single-node cluster and multi-node cluster.  FYI. https://github.com/etcd-io/etcd/pull/14394#issuecomment-1229606410 and https://github.com/etcd-io/etcd/pull/14394#issuecomment-1229673875

tbg:
Thanks for the review, @ahrtr. I'm not responding to the inline comments yet since you seem to have general concerns about the size of the change.

I think it comes down to the discussion in https://github.com/etcd-io/etcd/issues/14370#issuecomment-1237747454.

You're fixing `*node`, I'm fixing `raft`. So I have to update more tests.
Also, you keep the old behavior for empty entries for reasons that aren't clear to me. I think this prevents you from having to adjust more tests.

> Usually empty entries will only be generated in multi-node cluster. Since multi-node clusters do not have this issue, so I think it makes sense to keep the optimization (update the pr.Match immediately after appending it).

In my book, a special case needs a compelling reason. Special cases add complexity, and this complexity has to be made up for by some benefit. I cannot see that benefit here.

Some of the tests were pretty confused about whether they wanted to make statements about entry append or entry commit or entry apply. This is because the problem we're fixing made these things all the same in a single-node cluster in the same cycle.

I made sure to update the tests to be more robust (especially the `TestNode.*` tests are pretty annoying when they fail, prior to my improvements) and to not get too crufty. That is some extra work, but again, this PR should be judged by its changes to production code, not the corresponding updates in tests.

> 2. If you remove the optimization (mentioned above) so as to keep the behavior consistent for both empty and non-empty entry, then likely there are more subtle edge cases you need to handle, accordingly it might further complicate the raft package; and they are exactly what I have been trying to avoid.

Could you elaborate on this? Two cases = more complexity. In this PR, the empty entry is just an entry; why make it more complex? There should be fewer subtle edge cases now - namely none at all - and that is what we are after. What am I missing?

> 3. Actually https://github.com/etcd-io/etcd/pull/14411 doesn't almost change anything on the raft protocol, which I think is better. But I agree it has some impact on the direct users on RawNode.

We can agree to disagree here - *raft is the thing that produces `Ready` and the semantics of `Ready` are what need to be fixed. We could fix it in `*RawNode` but `RawNode` is intentionally a thin interface wrapper around `*raft` (similar to `*node` is just a goroutine around `*RawNode`). Spreading implementation details between them is not going to make things clearer. We shouldn't start to bolt on logic outside of `*raft` because we're scared of changing `*raft`.

ahrtr:
Thanks for the feedback. 

> Could you elaborate on this? Two cases = more complexity. In this PR, the empty entry is just an entry; why make it more complex? There should be fewer subtle edge cases now - namely none at all - and that is what we are after. What am I missing?

I just kept the legacy behavior on empty entry so as to keep the PR(14411) as simple&safe as possible, and it's the reason why the size of 14411 is much smaller than this one. I agree that regarding the empty entry as a special case isn't good, although it was consistent with the legacy behavior. 





> Some of the tests were pretty confused about whether they wanted to make statements about entry append or entry commit or entry apply. This is because the problem we're fixing made these things all the same in a single-node cluster in the same cycle.

Thanks for the clarification. 



> That is some extra work, but this PR should be judged by its changes to production code, not the corresponding updates in tests.

Partially agreed. We also need to make sure the quality of test code. The reason why I raised the concern on the size on the test code is just to understand the reason of the bigger size. Good to confirmed that one of the major reasons is due to the removal of the special case on empty entry.  

Overall looks good to me. Please,
1. Consider to resolve the inline minor comments;
2. Consider to squash the commits into a couple of commits if possible.

tbg:
Thanks @ahrtr. I'm travelling this week, so there won't be activity on this PR until next week but I plan to return to it the week after. I am also planning to validate the perf impact of this change with the etcd suite as you suggested, and also against CRDB.

raft/diff_test.go
tbg: The various tests that are using this really wanted to see what's in the raft log. That this was called "committed" was a misnomer, the tests really didn't care whether it was committed or not. It just so happened that append==commit in the single-node case prior to this change.

raft/node_util_test.go
tbg: No, we are implementing `raft.Logger` here, `*testing.T` does not implement that. It is true that some methods overlap (for example, `*testing.T` has a matching `Fatalf`) but semantically we don't necessarily want to call `t.Fatalf` just because a raft node calls `Fatalf`. I will leave this as is.

raft/node_util_test.go
tbg: Good call, fixed.

raft/node_test.go
tbg:
```
commit 0a6a94b5e7570f2876e677831964628e9df94272
Author: Tobias Grieger <tobias.b.grieger@gmail.com>
Date:   Tue Sep 6 17:26:13 2022 +0200

    remove TestNodeReadIndex

    This is tested directly at the level of `RawNode` in
    `TestRawNodeReadIndex`. `*node` is a thin wrapper around `RawNode` so
    this is sufficient.

    The reason to remove the test is that it now incurs data races
    since it's not possible to adjust the `readStates` and `step`
    fields while the node is running, and there is no primitive
    to synchronize with its goroutine. This could all be fixed
    but it's not worth it.

    Signed-off-by: Tobias Grieger <tobias.b.grieger@gmail.com>
```

tbg:
Addressed the first round of comments, see the fixup commit. I also added an (empty) commit describing the results for [this benchmark](https://github.com/etcd-io/etcd/pull/14394#issuecomment-1229606410).

TL;DR is that we're not regressing, in fact getting a little *faster*.

This is mildly surprising to me.
To be fair, I improved performance in this PR via some micro-optimizations which may result in fewer calls to `LastIndex`, `FirstIndex`, etc.
Also, the benchmark hopefully saturates the raft pipeline, meaning that even though we can't commit in a single round through raft any more we are doing lots of work in each raft cycle, i.e. we're fsyncing at about the same rate.
But I still didn't think we'd beat the main branch in latency, which seems to be happening, too, in both runs.

ahrtr:
Overall looks good to me. 

Good to see the performance enhancement below, although it is an unrelated change. 
```
func releasePendingReadIndexMessages(r *raft) {
	if len(r.pendingReadIndexMessages) == 0 {
		// Fast path for the common case to avoid a call to storage.LastIndex()
		// via committedEntryInCurrentTerm.
		return
	}
	......
}
```

Please resolve the conflict.

ahrtr:
Next steps:
1. After merging this PR, I am going to revert the PR https://github.com/etcd-io/etcd/pull/14400 on main branch;
2. Just as we discussed early, we will not backport this PR to release-3.5 and release-3.4;

Please let me know if you have any concerns or comments. cc @tbg @ptabor

ahrtr (approved):
Great work! Thank you @tbg

ahrtr:
Since there is no any more comments or concerns from other maintainers, so let me merge this PR, otherwise we need to rebase this PR to resolve conflict from time to time.

PR #14421, opened 2022-09-03
testing: fix TestOpenWithMaxIndex cleanup

### LINKED ISSUE #14332: [test] TestOpenWithMaxIndex failure
Raised on the latest code on `main` branch.

```
--- FAIL: TestOpenWithMaxIndex (0.00s)
    testing.go:967: TempDir RemoveAll cleanup: unlinkat /tmp/TestOpenWithMaxIndex2918559458/001: directory not empty
```

Refer to https://github.com/etcd-io/etcd/runs/7777195247?check_suite_focus=true

vsvastey:
A WAL object was closed by defer, however the WAL was rewritten afterwards,
so defer closed already closed WAL but not the new one. It caused a data
race between writing file and cleaning up a temporary test directory,
which led to a non-deterministic bug.

Fixes #14332

### DIFF
--- a/server/storage/wal/wal_test.go
+++ b/server/storage/wal/wal_test.go
@@ -635,24 +635,31 @@ func TestOpenForRead(t *testing.T) {
 func TestOpenWithMaxIndex(t *testing.T) {
 	p := t.TempDir()
 	// create WAL
-	w, err := Create(zaptest.NewLogger(t), p, nil)
+	w1, err := Create(zaptest.NewLogger(t), p, nil)
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer w.Close()
+	defer func() {
+		if w1 != nil {
+			w1.Close()
+		}
+	}()
 
 	es := []raftpb.Entry{{Index: uint64(math.MaxInt64)}}
-	if err = w.Save(raftpb.HardState{}, es); err != nil {
+	if err = w1.Save(raftpb.HardState{}, es); err != nil {
 		t.Fatal(err)
 	}
-	w.Close()
+	w1.Close()
+	w1 = nil
 
-	w, err = Open(zaptest.NewLogger(t), p, walpb.Snapshot{})
+	w2, err := Open(zaptest.NewLogger(t), p, walpb.Snapshot{})
 	if err != nil {
 		t.Fatal(err)
 	}
-	_, _, _, err = w.ReadAll()
-	if err == nil || err != ErrSliceOutOfRange {
+	defer w2.Close()
+
+	_, _, _, err = w2.ReadAll()
+	if err != ErrSliceOutOfRange {
 		t.Fatalf("err = %v, want ErrSliceOutOfRange", err)
 	}
 }


server/storage/wal/wal_test.go
```diff
@@ -639,7 +639,9 @@ func TestOpenWithMaxIndex(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer w.Close()
+	defer func() {
+		w.Close()
+	}()
```
ahrtr, line 644: I would suggest to move the `defer w.Close()` to the place right after the `Open` (about line 653) for better readability.

serathius, line 644: I'm not sure this changes anything, deferred function will still refer to variable `w` which is overridden in the same scope. 
Possible fixes:
* Use different scopes
* Use different names
* Copy 

```suggestion
	defer func(w *wal.WAL) {
		w.Close()
	}(w)
```

ahrtr: The proposed change looks good, but unfortunately it doesn't work. The change in this PR is working, its purpose is to make sure the `w` isn't evaluated until the function is actually called. The code you proposed will make the w to be evaluated immediately. 

Note that the w is opened two times, and also closed two times in `defer`. The reference to `w` in the first `defer` isn't valid anymore after recreating/opening the WAL. 

Please consider to update this PR per my comment above https://github.com/etcd-io/etcd/pull/14421#discussion_r962522454

serathius: > The proposed change looks good, but unfortunately it doesn't work.

I'm super confused, how both can be true :P

I think I understood the problem. I missed `w.Close()` on line 650. 
There are 2 changes I would suggest:
* add `defer w.Close()` on after line 655 (like @ahrtr suggested)
* Renaming `w` to reflect two separate opening of raft `w1`, `w2`. Sharing same name of variable that needs to be cleanup makes code unnecessarily complicated.

ahrtr: > I'm super confused, how both can be true :P


Ah:), I wanted to say your proposed change looks good in generic, but unfortunately it doesn't work in this case. :) 

I like the idea of w1 and w2 proposal, because it's clearer.

vsvastey: @ahrtr I like the idea to keep defer as close to the creation as possible. But if the Create method returns error, w might be nil and defer w.Close() will cause panic. It seems quite a common pattern create -> check error -> defer. So I believe that the current location of defer is perfect.

@serathius @ahrtr renaming to w1 w2 was the first idea that came to my mind, but there would be an issue with w1. We need to explicitly close it (line 650). After that defer w1.Close() doesn't make any sense and will Close a closed object. I'm pretty much sure that Close is idempotent at the moment, but who knows whether it will always be true.

I agree that this approach is more readable, so if you guys think that double close is not a problem, I will change the PR.

ahrtr: Does the following change make sense to you? @vsvastey If yes, please feel free to update this PR. thx
```
$ git diff
diff --git a/server/storage/wal/wal_test.go b/server/storage/wal/wal_test.go
index 5696f8f71..7135233b3 100644
--- a/server/storage/wal/wal_test.go
+++ b/server/storage/wal/wal_test.go
@@ -639,7 +639,6 @@ func TestOpenWithMaxIndex(t *testing.T) {
        if err != nil {
                t.Fatal(err)
        }
-       defer w.Close()
 
        es := []raftpb.Entry{{Index: uint64(math.MaxInt64)}}
        if err = w.Save(raftpb.HardState{}, es); err != nil {
@@ -651,6 +650,8 @@ func TestOpenWithMaxIndex(t *testing.T) {
        if err != nil {
                t.Fatal(err)
        }
+       defer w.Close()
+
        _, _, _, err = w.ReadAll()
        if err == nil || err != ErrSliceOutOfRange {
                t.Fatalf("err = %v, want ErrSliceOutOfRange", err)
```

vsvastey: @ahrtr unfortunately, if we remove the first defer and any panic happens in range 640 - 649 or for example on w.Save at 647 returns error, then we will not close the object and probably leak resources.

ahrtr: You are right. How about this? 

```
$ git diff
diff --git a/server/storage/wal/wal_test.go b/server/storage/wal/wal_test.go
index 5696f8f71..83f0dd7d5 100644
--- a/server/storage/wal/wal_test.go
+++ b/server/storage/wal/wal_test.go
@@ -635,23 +635,31 @@ func TestOpenForRead(t *testing.T) {
 func TestOpenWithMaxIndex(t *testing.T) {
        p := t.TempDir()
        // create WAL
-       w, err := Create(zaptest.NewLogger(t), p, nil)
+       w1, err := Create(zaptest.NewLogger(t), p, nil)
        if err != nil {
                t.Fatal(err)
        }
-       defer w.Close()
+       defer func() {
+               if w1 != nil {
+                       w1.Close()
+               }
+       }()
 
        es := []raftpb.Entry{{Index: uint64(math.MaxInt64)}}
-       if err = w.Save(raftpb.HardState{}, es); err != nil {
+       if err = w1.Save(raftpb.HardState{}, es); err != nil {
                t.Fatal(err)
        }
-       w.Close()
 
-       w, err = Open(zaptest.NewLogger(t), p, walpb.Snapshot{})
+       w1.Close()
+       w1 = nil
+
+       w2, err := Open(zaptest.NewLogger(t), p, walpb.Snapshot{})
        if err != nil {
                t.Fatal(err)
        }
-       _, _, _, err = w.ReadAll()
+       defer w2.Close()
+
+       _, _, _, err = w2.ReadAll()
        if err == nil || err != ErrSliceOutOfRange {
                t.Fatalf("err = %v, want ErrSliceOutOfRange", err)
        }
```

ahrtr: Please feel free to update this PR if it makes sense to you, thx

vsvastey: @ahrtr yes, I guess this solution is the best having both efficiency and readability. I've update the PR with it. Thanks.
Please note that I've also slightly changed a condition on line 662

ahrtr (approved):
LGTM

Thank you @vsvastey 

Could you please double check whether release-3.5 and relaese-3.4 have the similar issue and backport the PR if needed?

vsvastey:
@ahrtr For release-3.5 I cannot reproduce the initial issue. It is still a good idea to call Close properly, but the test doesn't fail, and temp directory keeps clean. Will you suggest to backport the change to 3.5 anyway?

ahrtr:
@vsvastey The test isn't stable, but I think it is a real issue. 

I suggest to backport to both 3.4 and 3.5. Please feel free to raise two PRs, thx

PR #14422, opened 2022-09-04
remove redundant log messsages

kkkkun:
Signed-off-by: kkkkun <scuzk373x@gmail.com>

Clean up some redundant log messages

### DIFF
--- a/pkg/osutil/interrupt_unix.go
+++ b/pkg/osutil/interrupt_unix.go
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"syscall"
 
+	"go.etcd.io/etcd/client/pkg/v3/verify"
+
 	"go.uber.org/zap"
 )
 
@@ -47,6 +49,7 @@ func RegisterInterruptHandler(h InterruptHandler) {
 
 // HandleInterrupts calls the handler functions on receiving a SIGINT or SIGTERM.
 func HandleInterrupts(lg *zap.Logger) {
+	verify.Assert(lg != nil, "the logger should not be nil")
 	notifier := make(chan os.Signal, 1)
 	signal.Notify(notifier, syscall.SIGINT, syscall.SIGTERM)
 
@@ -60,9 +63,7 @@ func HandleInterrupts(lg *zap.Logger) {
 
 		interruptExitMu.Lock()
 
-		if lg != nil {
-			lg.Info("received signal; shutting down", zap.String("signal", sig.String()))
-		}
+		lg.Info("received signal; shutting down", zap.String("signal", sig.String()))
 
 		for _, h := range ihs {
 			h()
--- a/server/etcdmain/etcd.go
+++ b/server/etcdmain/etcd.go
@@ -120,10 +120,12 @@ func startEtcdOrProxyV2(args []string) {
 			)
 		}
 	} else {
+		lg.Info(
+			"Initialize and start etcd server",
+			zap.String("data-dir", cfg.ec.Dir),
+			zap.String("dir-type", string(which)),
+		)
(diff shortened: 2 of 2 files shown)


pkg/osutil/interrupt_unix.go
```diff
@@ -60,9 +60,7 @@ func HandleInterrupts(lg *zap.Logger) {
 
 		interruptExitMu.Lock()
 
-		if lg != nil {
```
ahrtr, line 63: If the `lg` shouldn't be nil, then it would be better to add an assertion. 

FYI. https://github.com/etcd-io/etcd/blob/4d57eb8d3b2190bfdba3c65f8eb93c0349fc6dcc/pkg/schedule/schedule.go#L92

server/etcdmain/etcd.go
```diff
@@ -121,9 +121,6 @@ func startEtcdOrProxyV2(args []string) {
 		}
 	} else {
 		stopped, errc, err = startEtcd(&cfg.ec)
-		if err != nil {
-			lg.Warn("failed to start etcd", zap.Error(err))
-		}
```
ahrtr, line 126: Suggest to add one more log before starting etcd,
```
lg.Info(
			"server isn't initialized yet",
			zap.String("data-dir", cfg.ec.Dir),
			zap.String("dir-type", string(which)),
		)
```

pkg/osutil/interrupt_unix.go
kkkkun: It‘s  should not be nil because it would be checked in https://github.com/etcd-io/etcd/blob/4d57eb8d3b2190bfdba3c65f8eb93c0349fc6dcc/server/etcdmain/etcd.go#L54. And then call `HandleInterrupts` in https://github.com/etcd-io/etcd/blob/4d57eb8d3b2190bfdba3c65f8eb93c0349fc6dcc/server/etcdmain/etcd.go#L182.

ahrtr: OK, it would be better to an assertion

ahrtr:
@kkkkun Please resolve the inline minor comments, I will approve and merge this PR once it's done.

kkkkun:
> @kkkkun Please resolve the inline minor comments, I will approve and merge this PR once it's done.

Thanks.  Done！

pkg/osutil/interrupt_unix.go
```diff
@@ -60,9 +60,7 @@ func HandleInterrupts(lg *zap.Logger) {
...
-			lg.Info("received signal; shutting down", zap.String("signal", sig.String()))
-		}
+		lg.Info("received signal; shutting down", zap.String("signal", sig.String()))
```
ahrtr, line 63: Please add an assertion something like example below at the beginning of this function.

https://github.com/etcd-io/etcd/blob/4d57eb8d3b2190bfdba3c65f8eb93c0349fc6dcc/pkg/schedule/schedule.go#L92

kkkkun: https://github.com/etcd-io/etcd/pull/14422#discussion_r962517555 
Actually it's would not be nil. `lg != nil` should be historical reasons. 
But I am ok add this.

ahrtr: Please note that not only etcd may use this function, other applications may also use this function. We should clearly state the the `lg` shouldn't be nil using assertion.

kkkkun: That make sense. DONE！

ahrtr (approved):
LGTM

Thank you @kkkkun

Issue #13775, opened 2022-03-10
Introduce feature tracking


serathius:
When looking through some recent reports of data inconsistencies in etcd v3.5 I was surprised that all of them where detected manually by users. Manual detection means that users find out pretty late and they no longer have a logs from the event. This is very worrying that we are unable to effectively reproduce nor fix such cases.

I wondered why etcd doesn't have any mechanism to detect data inconsistencies. However, after more digging I found that there is a feature called "corrupt checks" just for that reason and it was implemented in v3.3 https://github.com/etcd-io/etcd/issues/7125. Weird, I haven't found it being used neither by [Kubernetes](https://github.com/kubernetes/kubernetes/search?q=%22corrupt+check%22&type=code) repo, [etcdadm](https://github.com/kubernetes-sigs/etcdadm/search?q=%22corrupt+check%22) nor any other public project using etcd. This is a big problem, how do etcd developers expect to reproduce and fix data corruption issues that users are not aware of and don't use the feature intended to detect them.

After more digging, the reason was clear, it's an experimental feature, that was meant to be graduated in v3.4 (https://github.com/etcd-io/etcd/issues/9190), but got dropped to v3.5 (https://github.com/etcd-io/etcd/issues/10893) and forgotten to be included in v3.5. Without proper feature tracing, such essential features don't get enough attention causing long term negative consequences for whole project. This doesn't just happens once, if you look through experimental flags there are a lot of features that were never graduated.

Problem:
* Over years etcd has collected a lot of experimental features that are abandoned and not tracked for the next release.
* Work of  features is not properly tracked between releases. If original owners is no longer actively contributing, work on the feature stops and issue is closed by original stale bot.
* Important etcd features have low adoption and are not known to users.

Proposal:
* Establish a basic feature graduation process and tracking for it. This process will be used to ensure that they are not dropped between releases, there are expectation that they will be moving forward and there is a plan to make the feature enabled by default. We need to agree on:
  * Stages: `experimental` -> `optional` -> `default` and also `deprecated` if we plan to remove it. 
  * Expectation about progress: `experimental` and `deprecated` should only take one release. We should also plan to make `optional` features enabled as `default`, but it could take longer.
  * Method to track work: We could have an issue per feature and use Github project board with different stages. Such issues should have a special label that would prevent them from being closed by stale bot. 
* Use the process for v3.6 release. We should go through all experimental features in the codebase, create issues for each of them and decide whether we want to graduate or deprecate them.

[...]

PR #14428, opened 2022-09-06
Add features dev and support guidelines doc

spzala:
Related https://github.com/etcd-io/etcd/issues/13775


spzala:
cc @serathius @ahrtr

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,36 @@
...
+- Any configuration flags related to the implementation of the feature are prefixed with `experimental` e.g. `--experimental-feature-name`.
+- Any variable names related to the implementation of the feature are prefixed with `Experimental` e.g. `ExperimentalFeatureName`.
+- Might be buggy. Enabling the feature may not work as expected. 
+- Support for such a feature may be dropped at any time without notice.
+- Disabled by default.
+
+### Graduation to Stable
+
+It is important that experimental features don't get stuck in that stage. An experimental feature should move to the stable stage by using the criteria provided below: 
+- The feature was enabled for at least one major release or one year, whatever comes first. This time frame should be enough to find any potential bugs in the feature.
```
ahrtr, line 18: Probably one `minor` release makes more sense here. For example, assuming we add an experimental feature `--experimental-abc` in 3.5, and it's disabled by default in 3.5.  It may be enabled by default in 3.6. Eventually we can graduate it in 3.7 if no any major issue is opened against the feature. 

If it's `major` release, then it means we can't graduate the feature until `5.0.`

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,36 @@
...
+
+A stable feature is characterized as below: 
+- Supported as part of the supported releases of etcd.
+- May be enabled by default.
+- Discontinuation of support must follow the Feature Depreciation Process. 
+
+### Feature Depreciation Process
+
+As the project evolves, sometimes a stable feature may need to be deprecated and removed. Such a situation should be handled using the steps below:
+- Provide a depreciation message in the feature usage document before a planned major release for feature depreciation. e.g. `To be deprecated in <release>`.
```
ahrtr, line 32: Same as above, Probably `minor` release makes more sense here.

Documentation/contributor-guide/features.md
spzala: @ahrtr thanks and yes agree!! That's how I had in mind too :) but mixed up. I will update it.

Documentation/contributor-guide/features.md
spzala: Yes, thanks @ahrtr

spzala: @ahrtr I think this can be both minor or minor+major, let me update changes and see if that make more sense.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,39 @@
...
+- Support for such a feature may be dropped at any time without notice.
+- Disabled by default when added initially.
+
+### Graduation to Stable
+
+It is important that experimental features don't get stuck in that stage.  An experimental feature should move to the stable stage by following the lifecycle steps below:
+- A new feature added in the latest release and disabled by default e.g. etcd v3.5 or v4.0
+- Given there are no open issues against the feature,
+  - It is enabled by default in the next release e.g. etcd v3.6 or v4.1
+  - It move to stable stage in the following release e.g. etcd v3.7 or v4.2. If release cycle takes longer, one year time frame can be used to move to the stable stage.
```
ahrtr, line 21: "move" --> "moves"?

ahrtr, line 21: In general, I agree with the two points here. But I think two minor releases is just a minimum requirement. It means it isn't necessarily to graduate an experimental feature after two minor releases (e.g., added in 3.5, and graduate in 3.7), although usually it does. 

The point is we need to have some flexibility on this.

Documentation/contributor-guide/features.md
ahrtr: ack

Documentation/contributor-guide/features.md
spzala: @ahrtr I agree that flexibility is needed. How about re-write as below for clarity that the next action is based on the review?: 
- Given there are no open issues against the feature, it is ready for review to, 
  - Enable by default in the next release e.g. etcd v3.6 or v4.1
  - Move to the stable stage in the following release e.g. etcd v3.7 or v4.2. If release cycle takes longer, one year time frame can be used for the review.

ahrtr: `it is ready for review` makes sense, since we have flexibility to approve or reject the PR.

spzala: Yup, thanks @ahrtr

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,39 @@
...
+- Any variable names related to the implementation of the feature are prefixed with `Experimental` e.g. `ExperimentalFeatureName`.
+- Might be buggy. Enabling the feature may not work as expected. 
+- Support for such a feature may be dropped at any time without notice.
+- Disabled by default when added initially.
+
+### Graduation to Stable
+
+It is important that experimental features don't get stuck in that stage.  An experimental feature should move to the stable stage by following the lifecycle steps below:
+- A new feature added in the latest release and disabled by default e.g. etcd v3.5 or v4.0
+- Given there are no open issues against the feature, it is ready for review to,
```
serathius, line 19: I think we should have more strict requirements for graduating feature than just no open issues. 

Things that should be considered when graduating (not all are required depending on feature) :
* Logs allowing for proper debugging
* Documentation
* Metrics
* Benchmarks
* Proper E2e tests
* upgrade/downgrade tests if there are changes in api, raft, wal protos

Graduation issue should have those requirements as a checklist and be approved by at least 2 maintainers that are willing to support this feature.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Experimental
+
+Any new feature should be added as an experimental feature. An experimental feature is characterized as below:
+- Associated with an issue with a clear need for such a feature.
+- Any configuration flags related to the implementation of the feature are prefixed with `experimental` e.g. `--experimental-feature-name`.
+- Any variable names related to the implementation of the feature are prefixed with `Experimental` e.g. `ExperimentalFeatureName`.
+- Might be buggy. Enabling the feature may not work as expected. 
+- Support for such a feature may be dropped at any time without notice.
+- Disabled by default when added initially.
```
serathius, line 13: Could we describe minimal requirements for feature to be accepted into etcd as experimental. Fact that issue is experimental doesn't mean that we accept any low quality code.

Proposed 
* Unit tested
* One E2e test for basic scenario

Even though etcd can reserve right to stop supporting the feature and remove it in next release, we should still ensure that we don't unneseserly bloat codebase. I think accepting features should also require support from at least 2 maintainers .

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Graduation to Stable
```
serathius, line 15: I expect that during graduation we will remove `experimental` prefix from feature flag. This would be a breaking change in flags that could be easily mitigated by not renaming old experimental flag, but by marking it as hidden and adding new one. There should not be a difference in behavior but it should bring big improvements to user experience during upgrade.

ahrtr: I am thinking can we follow the same approach as [Kubernetes](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) to manage the feature?

serathius: Feature gates are good when you don't need to do any configuration for feature to work. On multiple occasions I have added a new feature to Kubernetes and decided against feature gating. 

What I am proposing is more inline with https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli. From CLI perspective graduating a feature is a breaking change without any deprecation. We remove experimental flag and add a new one. During upgrade admins need to take a immediate action and update all graduated flags. 

I'm proposing that when feature graduates we still leave `experimental` flags for one release, giving users time to adapt.

Alternative approaches:
* Stop using `experimental` prefix and require user to opt in into experimental feature by providing separate `--experimental-feature-gate` flag like https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

ahrtr: > * Stop using `experimental` prefix and require user to opt in into experimental feature by providing separate `--experimental-feature-gate` flag like https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

This is exactly what I am thinking. 

- Adding/removing the prefix `experimental` is indeed a breaking change; 
- Keeping `experimental` flags for one more release complicates the feature graduation process.

spzala: Yes, good points both about breaking changes related to prefix, thanks. May be simple way would be to treat experimental features naming/usage in a similar way to unsafe features (I missed it in the stages earlier by we have stable, experimental and unsafe features currently). I will be pinging you for quick discussion around it.

Documentation/contributor-guide/features.md
spzala: Good points, I will make them more explicit. Thanks!

Documentation/contributor-guide/features.md
serathius: I think there are 3 separate things to address:
* Have documentation reflect the current state (usage of `experimental` prefix)
* Avoiding breaking changes in flags (not immidietly removing experimental flags upon graduation)
* Long term approach for handling feature flags that avoid problem of flag compatibility altogether. 

In this PR we should address first two points. @spzala feel free to open an issue with proposal for the third one.

Documentation/contributor-guide/features.md
lavacat: Another proposal:
Limit number of experimental features. If limit is reached, author of a new feature has to deprecate/graduate existing feature.
It might sound too much, but otherwise this work will fall onto maintainers.

spzala: Thanks @lavacat but I think we cannot limit the numbers as adding a new feature is need based. I will be opening a separate issue for future approach and we can discuss it there as needed.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+- Disabled by default when added initially
+
+### Minimum development requirements
+
+Experimental feature related development work is expected to be high quality, similar to any development work for the project. A related PR should meet the following criteria,
+- Associated with an issue providing a clear need for such a feature
+- Any configuration flags related to the implementation of the feature are prefixed with `experimental` e.g. `--experimental-feature-name`
+- Any variable names related to the implementation of the feature are prefixed with `Experimental` e.g. `ExperimentalFeatureName`
+- Provided with unit tests
+- Provided with e2e test coverage for at least the basic scenario
```
ahrtr, line 21: e2e or integration test

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+- Added a CHANGELOG entry
+- Disabled by default
+- Approved by at least two project maintainers
+
+### Graduation to Stable
+
+It is important that experimental features don't get stuck in that stage. After a release of an intiial enablement (e.g. etcd v3.5), an experimental feature should be revisited in the following release (e.g. etcd v3.6 or after six months if release takes longer) and see if it's ready for graduation to the stable stage in the future release (e.g. etcd v3.7 or v4.0). If it's ready per discussions with the project maintainers, then depending on the scope of the feature, at least the following enhancements should be consiered to be worked on by opening a PR(s) and new issue(s) for tracking as needed,
+- Fix any open issues against the feature
+- Add robust e2e tests
```
ahrtr, line 30: e2e or integration tests

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+- Enhance logs for proper debugging
+- Update metrics and benchmarks if needed
+- Upgrade/downgrade tests specially if there are changes in API, Raft, WAL protos.
+- Enable by default if appropriate
+- Modify the feature documentation and add a CHANGELOG entry to reflect that `--experimental` prefix in the feature name is expected to be dropped in the next release e.g. `Expected to be graduated and renamed to --feature-name in the future release`
+- Create a new issue for future graduation with reference to the PR discussed here. Add the release label of the upcoming release for tracking.
+
+One release with above enhancements should give a good idea on the feature's stability. Unless there are open issues that can prevent feature gradution, the feature can move to the stable stage in the next release with a PR on feature rename, doc updates, and an entry in the CHANGELOG. All the related PRs and discussions should be approved by at least two project maintainers.
```
ahrtr, line 38: typo: graduation

spzala: ah :) thanks @ahrtr

spzala: Also, as you and @serathius will notice and as I mentioned in the commit message, I have addressed feedback with some added thoughts. I think by the time, we graduate a feature to stable it should be treated as a stable feature for a release (something that I added here). And also a warning message on possible name change (i.e. removing prefix) for a release should serve our purpose on giving user heads up (something that's general done with deprecation). Thanks!

ahrtr (approved):
LGTM

Thanks @spzala 

Although removing the `--experimental` may have some impact on the client application when graduating a feature, but it is consistent with the existing behavior,  and it explicitly requires users' awareness and action so as to avoid any unexpected surprise. 

We can think about and enhance it later if needed.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+### Graduation to Stable
+
+It is important that experimental features don't get stuck in that stage. After a release of an intiial enablement (e.g. etcd v3.5), an experimental feature should be revisited in the following release (e.g. etcd v3.6 or after six months if release takes longer) and see if it's ready for graduation to the stable stage in the future release (e.g. etcd v3.7 or v4.0). If it's ready per discussions with the project maintainers, then depending on the scope of the feature, at least the following enhancements should be consiered to be worked on by opening a PR(s) and new issue(s) for tracking as needed,
+- Fix any open issues against the feature
+- Add robust e2e tests
+- Enhance logs for proper debugging
+- Update metrics and benchmarks if needed
+- Upgrade/downgrade tests specially if there are changes in API, Raft, WAL protos.
+- Enable by default if appropriate
+- Modify the feature documentation and add a CHANGELOG entry to reflect that `--experimental` prefix in the feature name is expected to be dropped in the next release e.g. `Expected to be graduated and renamed to --feature-name in the future release`
```
ahrtr, line 35: Probably we'd better to clearly state that we will make both the experimental flag and graduated flag co-exist for at least one minor release. For example, when we decide to graduate a experimental feature in 3.6, we add a new flag `--feature-name`, and do not remove the `--experimental-feature-name` until next minor, e.g. 3.7 (or major 4.0 if no next minor release) release.

Documentation/contributor-guide/features.md
serathius: For new features that expose new user facing elements (command line flags), I would always prefer e2e tests over integration.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+
+This document provides an overview of etcd features and general guidelines for adding and removing features. The project maintainers can override the guidelines per the need of the project following the project governance.
+
+The etcd features fall into three stages, experimental, stable, and unsafe.
+
+## Experimental
+
+Any new feature should be added as an experimental feature. An experimental feature is characterized as below:
+- Might be buggy due to lack of user testing. Enabling the feature may not work as expected.
+- Support for such a feature may be dropped at any time without notice
```
serathius, line 11: I think we should clarify what dropping support means. This is important to convey what users should expect when we break support. I think there are two parts for it:
* Issues related to the feature will be closed without investigation
* Feature will be removed in the next minor release.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+- Disabled by default when added initially
+
+### Minimum development requirements
+
+Experimental feature related development work is expected to be high quality, similar to any development work for the project. A related PR should meet the following criteria,
+- Associated with an issue providing a clear need for such a feature
+- Any configuration flags related to the implementation of the feature are prefixed with `experimental` e.g. `--experimental-feature-name`
+- Any variable names related to the implementation of the feature are prefixed with `Experimental` e.g. `ExperimentalFeatureName`
```
serathius, line 19: I think this only holds true for external configuration in `embed.Server` and is not propagated deeper. I think it is a bit of overkill.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+- Provided with unit tests
+- Provided with e2e test coverage for at least the basic scenario
+- Added a CHANGELOG entry
+- Disabled by default
+- Approved by at least two project maintainers
+
+### Graduation to Stable
+
+It is important that experimental features don't get stuck in that stage. After a release of an intiial enablement (e.g. etcd v3.5), an experimental feature should be revisited in the following release (e.g. etcd v3.6 or after six months if release takes longer) and see if it's ready for graduation to the stable stage in the future release (e.g. etcd v3.7 or v4.0). If it's ready per discussions with the project maintainers, then depending on the scope of the feature, at least the following enhancements should be consiered to be worked on by opening a PR(s) and new issue(s) for tracking as needed,
```
serathius, line 28:
```suggestion
It is important that experimental features don't get stuck in that stage. After a release of an initial enablement (e.g. etcd v3.5), an experimental feature should be revisited in the following release (e.g. etcd v3.6 or after six months if release takes longer) and see if it's ready for graduation to the stable stage in the future release (e.g. etcd v3.7 or v4.0). If it's ready per discussions with the project maintainers, then depending on the scope of the feature, at least the following enhancements should be consiered to be worked on by opening a PR(s) and new issue(s) for tracking as needed,
```

Documentation/contributor-guide/features.md
serathius, line 38: Note on `Unless there are open issues that can prevent feature graduation`: We don't have a good way to track issues for each feature. Instead of scattering graduation requirements into multiple issues that we need to handle, we should keep single issue for feature graduation. In the top comment there should be a checklist for each requirement with link to PR/issue that implements this.

I would change this point to say that graduating a feature should require creating an issue with list of requirements that tracks the proces. When enough requirements are fulfilled and issue get approval by 2 maintainers, then the feature can be graduated by making proper rename/documentation/changelog updates.

serathius:
I think that adding this would really help to clarify some of unwritten processes, however it's unclear for me who is the audience for this document. It describes both internal developer graduation proces and user facing support guarantees. I think it would be much clearer if we split it into two separate documents based on audience:
* User facing documentation describing feature maturity levels (experimental, graduated, deprecated).
* Developer facing description of process to move features between those levels (how to introduce new feature to experimental, what is needed to graduate or deprecate a feature).

Most important for now is to maintain and improve quality. All features should have proper test, documentation and metrics. I think we can postpone the detailing the support and deprecation policies for user facing documentation. I would skip description of experimental/mature features and instead describe the process we expect developers should go through.

lavacat:
To provide some context, I've listed all experimental features across 3.3, 3.4, 3.5 and main branches.
https://docs.google.com/spreadsheets/d/1_HleY81w5AHqIA857O6SlSzzWtPjCvHT1ZkwWUOr88Q/edit#gid=0

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,60 @@
...
+- Fix any open issues against the feature
+- Add robust e2e tests
+- Enhance logs for proper debugging
+- Update metrics and benchmarks if needed
```
lavacat, line 32: Maybe robust tests, logs, metrics and benchmarks should be part of minimum dev requirements. Otherwise this creates an expectation that somebody will pick up this work during 'graduation' phase. This will slow down graduation process. Also original author probably has the best context to add those tests/logs.

For example `experimental-compaction-batch-limit` https://github.com/oss-study/etcd-3.4.9/commit/1e213b7ab6991d67722e7e1496a93b3a7fae0366#diff-0abf7361537ed66a5b8d85cdc518a4ab2f693a1e8c929c93de3746851a5ddfbe

Technically it's just exposing hardcoded value. But to graduate it, do we need more tests/benchmarks?

spzala:
> To provide some context, I've listed all experimental features across 3.3, 3.4, 3.5 and main branches. https://docs.google.com/spreadsheets/d/1_HleY81w5AHqIA857O6SlSzzWtPjCvHT1ZkwWUOr88Q/edit#gid=0
> 
> There was only 1 experimental feature that was graduated `experimental-backend-bbolt-freelist-type` and 1 removed `experimental-enable-v2v3`

Thanks for the doc! Going forward our goal would be not to keep experimental features in that stage forever but revisit them and see if they can be graduated to stable stage or be removed as needed.

Documentation/contributor-guide/features.md
```diff
@@ -0,0 +1,50 @@
...
+    - Provide logs for proper debugging.
+    - Provide metrics and benchmarks as needed.
+    - The Feature should be disabled by default.
+    - Any configuration flags related to the implementation of the feature must be prefixed with `experimental` e.g. `--experimental-feature-name`.
+    - Add a CHANGELOG entry.
+- At least two maintainers must approve feature requirements and related code changes.
+
+### Graduating an Experimental feature to Stable (TODO - spzala)
+
+### Deprecating a feature (TODO - spzala)
```
spzala, line 50: @serathius I have restructured the doc for a better readability and easier review, and also tried addressing the review feedback. I will be addressing two TODOs separately/iteratively building on top of this initial overview and guidelines. I think that way it will be easier to review and speed up finalizing the content. Also, once developer doc is done, I will be adding user doc. Please let me know your thoughts. Thanks! cc @ahrtr

serathius: Looks great, short and precise.

serathius (approved):

Documentation/contributor-guide/features.md
spzala: Thanks @serathius

PR #14307, opened 2022-08-03
etcdctl: allow move-leader to connect to multiple endpoints

tjungblu:
Re-opening closed PR etcd-io#11775 which was originaly authored by benmoss.
Then again opened PR etcd-io#12757 which was authored by zerodayz.

The mustClientForCmd function is responsible for parsing environment
variables and flags into configuration data. A change was made in etcd-io#9382
to call Fatal if a flag is provided multiple times. This means that we
cannot call the mustClientForCmd function more than once,
since it will think that flags parsed the first time are now
being redefined and error out.

Some people have commented about this in etcd-io#8380 but I don't think
there's an open issue for it.

Signed-off-by: Thomas Jungblut <tjungblu@redhat.com>

tracked under our [BZ1918413](https://bugzilla.redhat.com/show_bug.cgi?id=1918413)

PR #14434, opened 2022-09-07
etcdctl: allow move-leader to connect to multiple endpoints

tjungblu:
Re-opening closed PR etcd-io#11775 which was originaly authored by benmoss. Then again opened PR etcd-io#12757 which was authored by zerodayz. And then a continuation of #14307.

The mustClientForCmd function is responsible for parsing environment variables and flags into configuration data. A change was made in etcd-io#9382 to call Fatal if a flag is provided multiple times. This means that we cannot call the mustClientForCmd function more than once, since it will think that flags parsed the first time are now being redefined and error out.

Some people have commented about this in etcd-io#8380 but I don't think there's an open issue for it.

### DIFF
--- a/etcdctl/ctlv3/command/move_leader_command.go
+++ b/etcdctl/ctlv3/command/move_leader_command.go
@@ -43,7 +43,8 @@ func transferLeadershipCommandFunc(cmd *cobra.Command, args []string) {
 		cobrautl.ExitWithError(cobrautl.ExitBadArgs, err)
 	}
 
-	c := mustClientFromCmd(cmd)
+	cfg := clientConfigFromCmd(cmd)
+	c := cfg.mustClient()
 	eps := c.Endpoints()
 	c.Close()
 
@@ -53,7 +54,6 @@ func transferLeadershipCommandFunc(cmd *cobra.Command, args []string) {
 	var leaderCli *clientv3.Client
 	var leaderID uint64
 	for _, ep := range eps {
-		cfg := clientConfigFromCmd(cmd)
 		cfg.endpoints = []string{ep}
 		cli := cfg.mustClient()
 		resp, serr := cli.Status(ctx, ep)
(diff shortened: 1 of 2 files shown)


etcdctl/ctlv3/command/move_leader_command.go
```diff
@@ -19,6 +19,7 @@ import (
 	"strconv"
 
 	"github.com/spf13/cobra"
+
```
ahrtr, line 22: minor comment: Did you intentionally add this blank line or it's added by IDE automatically? I would suggest to keep it as it's. Of course, it isn't a big  deal.

ahrtr (approved):
LGTM with a minor comment. 

Thank you @tjungblu 

cc @spzala @serathius @ptabor

etcdctl/ctlv3/command/move_leader_command.go
spzala: +1

spzala (approved):
@tjungblu thanks! lgtm but I think the PR/commit description is not correct, so I would suggest to at least edit PR description for future reference if someone look at it. First, there is a typo in `mustClientForCmd`, it's `mustClientCfgFromCmd`. Secondly, it's not the `mustClientCfgFromCmd` getting called multiple times resulting into flag related error as the description says. It's the `clientConfigFromCmd` that's getting called twice (directly and through `mustClientCfgFromCmd`) which the PR provides fix for.

etcdctl/ctlv3/command/move_leader_command.go
tjungblu: yep IDE, let me remove it again

tjungblu:
Thanks for the reviews, fixed the commit and removed that newline.

ahrtr:
@tjungblu  Could you please backport this PR to 3.4 as well?

tjungblu:
sure, can do.

PR #14441, opened 2022-09-08
[release-3.4] etcdctl: fix move-leader for multiple endpoints

tjungblu:
This is a backport of #14434 to 3.4. 

---

Due to a duplicate call of clientConfigFromCmd, the move-leader command would fail with "conflicting environment variable is shadowed by corresponding command-line flag". Also in scenarios where no command-line flag was supplied.

Signed-off-by: Thomas Jungblut <tjungblu@redhat.com>

PR #14445, opened 2022-09-09
etcdctl: fix move-leader for multiple endpoints

tjungblu:
Due to a duplicate call of clientConfigFromCmd, the move-leader command would fail with "conflicting environment variable is shadowed by corresponding command-line flag". Also in scenarios where no command-line flag was supplied.

---

main pick of #14434 and #14441 
thanks @ahrtr for triple checking me :)

### DIFF
--- a/tests/e2e/ctl_v3_move_leader_test.go
+++ b/tests/e2e/ctl_v3_move_leader_test.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tests/e2e/ctl_v3_move_leader_test.go
```diff
@@ -88,13 +101,12 @@ func testCtlV3MoveLeader(t *testing.T, cfg e2e.EtcdProcessClusterConfig) {
 		}
 	}
 
-	os.Setenv("ETCDCTL_API", "3")
-	defer os.Unsetenv("ETCDCTL_API")
```
tjungblu, line 92: this was spouting a warning, so I thought I'd clean this up. Let me know if I should leave it.

ahrtr: `ETCDCTL_API` is 3 by default, why the case need this? It should be legacy "issue". So it's Ok to remove it.

ahrtr:
FYI. https://github.com/etcd-io/etcd/pull/14443 

I realized that the `SetPflagsFromEnv` is a little strange, probably we may need to refactor it a little bit later.

ahrtr (approved):
LGTM

Thank you @tjungblu

serathius (approved):

PR #14453, opened 2022-09-13
fix the flaky test TestV3AuthRestartMember

ahrtr:

cc @serathius @mitake


mitake:
@ahrtr Do you want to close this PR without merging?

ahrtr:
@mitake Yes, I closed this PR intentionally, because` clus.Members[0].WaitOK(t)` uses clientv3, and it always fails due to no username:password. While 3.4 and 3.5 do not have such issue, because they are using clientv2 API.

PR #13947, opened 2022-04-15
Client: fix panics if no logger is provided to fileutil and transport.

nickexported:
Fixes panics in client pkg in transport that can happen if nil logger is provided. Also fixes panic with nil logger in fileutil, which is used by transport.

PR #14455, opened 2022-09-13
add verification on nil log

ahrtr:
There is no response on https://github.com/etcd-io/etcd/pull/13947 for a long time, so fix it in this PR.

cc  @serathius @mitake @spzala


mitake:
@ahrtr LGTM. Should the commit log format be like `client: add verification on nil log` (including the package prefix)? It seems that these days the tests don't check the commit log so it might be fine though.

mitake (approved):
LGTM

PR #14457, opened 2022-09-13
etcdserver: fix corruption check when server has just been compacted

### LINKED ISSUE #14325: Periodic corruption check doesn’t do anything after compaction
### What happened?

After turning on `--experimental-corrupt-check-time`, I’ve noticed that any time there has been a compaction and no recent write to etcd, the corruption check fails to actually run. 

The output on the leader is
```
{"level":"info","ts":"2022-08-02T17:55:38.557Z","caller":"etcdserver/corrupt.go:244","msg":"finished peer corruption check","number-of-peers-checked":0}
```

Looking at a follower, we see the requested revision has been compacted

```
{"level":"warn","ts":"2022-08-02T17:48:38.516Z","caller":"etcdserver/corrupt.go:377","msg":"failed to get hashKV","requested-revision":2592,"error":"mvcc: required revision has been compacted"}
```

I then wrote a key to etcd and observed the corruption check working correctly on the next run.

The corruption check should work regardless of whether or not a compaction has happened very recently.

- Set `--experimental-corrupt-check-time` to something very short like 1m
- Compact your etcd cluster
- Note that the next corruption check doesn't work
- Insert a key into etcd
- Note that the next corruption check does work

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.5.3
Git SHA: GitNotFound
Go Version: go1.16.2
Go OS/Arch: linux/amd64

$ etcdctl version
etcdctl version: 3.5.3
API version: 3.5
```

</details>

<details>
  ...
  --enable-v2=false
  --auto-compaction-retention=1h
  --experimental-corrupt-check-time=6h
  --experimental-initial-corrupt-check
  --log-level 'info'
  ...
</details>

<details>

Leaving these off. Happy to share more details if required. It is a healthy five node cluster, no learners, all running v3.5.3.

```console
# paste output here

# paste output here
```

</details>

### Relevant log output

_No response_

jbml:
When a key-value store corruption check happens immediately after a compaction, the revision at which the key-value store hash is computed, is the compacted revision itself.
In that case, the hash computation logic was incorrect because it returned an ErrCompacted error; this error should instead be returned when the revision at which the key-value store is hashed, is strictly lower than the compacted revision.

Fixes #14325

Signed-off-by: Jeremy Leach <44558776+jbml@users.noreply.github.com>

### DIFF
--- a/server/storage/mvcc/kvstore_test.go
+++ b/server/storage/mvcc/kvstore_test.go
@@ -542,9 +542,7 @@ type hashKVResult struct {
 func TestHashKVWhenCompacting(t *testing.T) {
 	b, tmpPath := betesting.NewDefaultTmpBackend(t)
 	s := NewStore(zaptest.NewLogger(t), b, &lease.FakeLessor{}, StoreConfig{})
-	defer s.Close()
-	defer b.Close()
-	defer os.Remove(tmpPath)
+	defer cleanup(s, b, tmpPath)
 
 	rev := 10000
 	for i := 2; i <= rev; i++ {
@@ -552,9 +550,10 @@ func TestHashKVWhenCompacting(t *testing.T) {
 	}
 
 	hashCompactc := make(chan hashKVResult, 1)
-
-	donec := make(chan struct{})
 	var wg sync.WaitGroup
+	donec := make(chan struct{})
+
+	// Call HashByRev(10000) in multiple goroutines until donec is closed
 	for i := 0; i < 10; i++ {
 		wg.Add(1)
 		go func() {
@@ -573,10 +572,12 @@ func TestHashKVWhenCompacting(t *testing.T) {
 		}()
 	}
 
+	// Check computed hashes by HashByRev are correct in a goroutine, until donec is closed
+	wg.Add(1)
 	go func() {
-		defer close(donec)
+		defer wg.Done()
 		revHash := make(map[int64]uint32)
-		for round := 0; round < 1000; round++ {
+		for {
 			r := <-hashCompactc
 			if revHash[r.compactRev] == 0 {
 				revHash[r.compactRev] = r.hash
@@ -584,17 +585,26 @@ func TestHashKVWhenCompacting(t *testing.T) {
 			if r.hash != revHash[r.compactRev] {
 				t.Errorf("Hashes differ (current %v) != (saved %v)", r.hash, revHash[r.compactRev])
 			}
+
+			select {
+			case <-donec:
+				return
+			default:
+			}
 		}
 	}()
 
-	wg.Add(1)
+	// Compact the store in a goroutine, using revision 9900 to 10000 and close donec when finished
 	go func() {
-		defer wg.Done()
+		defer close(donec)
 		for i := 100; i >= 0; i-- {
-			_, err := s.Compact(traceutil.TODO(), int64(rev-1-i))
+			_, err := s.Compact(traceutil.TODO(), int64(rev-i))
 			if err != nil {
 				t.Error(err)
 			}
+			// Wait for the compaction job to finish
+			s.fifoSched.WaitFinish(1)
+			// Leave time for calls to HashByRev to take place after each compaction
 			time.Sleep(10 * time.Millisecond)
 		}
 	}()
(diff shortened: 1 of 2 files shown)


server/storage/mvcc/kvstore_test.go
```diff
@@ -542,23 +542,19 @@ type hashKVResult struct {
 func TestHashKVWhenCompacting(t *testing.T) {
```
jbml, line 542: The reason for the refactoring of this test case, is that the test used to run compactions/HashByRev  for 1000 rounds, using hash 9899 to 9999, with no guarantee that all compactions have finished before the end of the test.

Now, the test runs for as long as it takes to complete all compactions (within a limit of 10 seconds), using revisions from 9900 to 10000. This is beneficial because the problem from the initial issue, only appears after compacting revision 10000.

jbml:
Thank you for reviewing the PR @serathius 

> * It makes `HashKV` revision semantics inconsistent with rest of API. Compaction removes etcd history up to and including  specified revision. This operation is inclusive, meaning it also removes the changes from the revision. This is why you cannot request neither `HashKV` nor `Range` or revision that was compacted. This change would make it possible to get Hash for revision that is no longer accessible rest of API.

The intention here was actually to make the HashByRev API consistent with Range, which I believe is what the author of #14325 meant in https://github.com/etcd-io/etcd/issues/14325#issuecomment-1210717339 when he/she said:

> It's a surprise to me that after compaction, the latest revision is not available to be queried at. Does that seem like the expected behavior? Based on testing with just etcdctl that doesn't seem to be the case.

I believe the following test illustrates this point. Start with a fresh etcd cluster:
```
> ./bin/etcdctl put foo bar1
> ./bin/etcdctl put foo bar2
> ./bin/etcdctl put foo bar3
> ./bin/etcdctl put foo bar4
> ./bin/etcdctl put foo bar5

>. /tools/etcd-dump-db/etcd-dump-db iterate-bucket  ~/projects/etcd/data/uniquenode.etcd key --decode
rev={main:6 sub:0}, value=[key "foo" | val "bar5" | created 2 | mod 6 | ver 5]
rev={main:5 sub:0}, value=[key "foo" | val "bar4" | created 2 | mod 5 | ver 4]
rev={main:4 sub:0}, value=[key "foo" | val "bar3" | created 2 | mod 4 | ver 3]
rev={main:3 sub:0}, value=[key "foo" | val "bar2" | created 2 | mod 3 | ver 2]
rev={main:2 sub:0}, value=[key "foo" | val "bar1" | created 2 | mod 2 | ver 1]
```
Then compact it at revision 5:
```
> ./bin/etcdctl compact 5
compacted revision 5

./tools/etcd-dump-db/etcd-dump-db iterate-bucket  ~/projects/etcd/data/uniquenode.etcd key --decode
rev={main:6 sub:0}, value=[key "foo" | val "bar5" | created 2 | mod 6 | ver 5]
rev={main:5 sub:0}, value=[key "foo" | val "bar4" | created 2 | mod 5 | ver 4]
```

Then check what revisions are accessible with Range:
```
> ./bin/etcdctl --rev=4 get foo
{"level":"warn","ts":"2022-09-18T16:19:51.102+1000","logger":"etcd-client","caller":"v3/retry_interceptor.go:64",
"msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc00000a780/127.0.0.1:2379",
"method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = OutOfRange 
desc = etcdserver: mvcc: required revision has been compacted"}
Error: etcdserver: mvcc: required revision has been compacted
> ./bin/etcdctl --rev=5 get foo
foo
bar4
> ./bin/etcdctl --rev=6 get foo
foo
bar5
```

As the version at which the compaction is done (5) is still available, it made sense to me that HashByRev raising an "required revision has been compacted" error when called on the revision, was indeed an issue.

This inconsistency is what we can see when we compare this for the HashByRev API:
https://github.com/etcd-io/etcd/blob/e61d43179291562c44225bf49a6d97c52a65fdca/server/storage/mvcc/kvstore.go#L180-L182
and this for the Range API:
https://github.com/etcd-io/etcd/blob/fff5d00ccfddaf5f06fa11cad65c1ecaa8c09f87/server/storage/mvcc/kvstore_txn.go#L74-L76
with the comparison being "<=" on one side, and "<" on the other side.

> * Impact is negligible as provided scenario is not realistic. I don't expect that in production clusters someone will compact up to the latest revision and than don't do any updates for long enough for corruption check to run. Periodic corruption check is pretty heavy operation so I also don't expect anyone to set it for 1 minute period. For example proposed period was 12h https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-4.0.md#etcd-server

Fully agree that this is a rare case.

> * Periodic check is known to be not reliable. We have already implemented a replacement. [etcd detects data corruption by default #14039](https://github.com/etcd-io/etcd/issues/14039). I think we can consider deprecating the old check when we have enough production data.

Fair enough.

jbml:
@serathius : would you suggest we simply close the PR and associated issue, please ?

serathius:
> The intention here was actually to make the HashByRev API consistent with Range

Sorry for mistaking it there. I think there is some value in making API consistent, however from top of my head I don't know if this change doesn't negatively impact any of existing mechanisms using HashKV. @ahrtr what's your opinion?

ahrtr:
> > The intention here was actually to make the HashByRev API consistent with Range
> 
> Sorry for mistaking it there. I think there is some value in making API consistent, however from top of my head I don't know if this change doesn't negatively impact any of existing mechanisms using HashKV. @ahrtr what's your opinion?

I will have a closer look at this PR and related issue late today or tomorrow morning my time.

ahrtr:
I think this is a valid PR. 

When the leader performs the [periodic check ](https://github.com/etcd-io/etcd/blob/54f9483e725f7300440fcabd9fbc4b94c81da3bd/server/etcdserver/corrupt.go#L167), it calculates the hash up to the `currentRev`. The reason why it can pass the check below is that the `rev` is 0.

https://github.com/etcd-io/etcd/blob/54f9483e725f7300440fcabd9fbc4b94c81da3bd/server/storage/mvcc/kvstore.go#L180

But when leader uses the same `currentRev` to get peers' hashes, it fails to pass the same check above, because the rev isn't 0 anymore. It doesn't make sense.  It's exactly what this PR tries to fix.

ahrtr:
Please note that when I review this PR, I found a related issue, please see https://github.com/etcd-io/etcd/pull/14510

server/storage/mvcc/kvstore_test.go
```diff
@@ -542,23 +542,19 @@ type hashKVResult struct {
...
 
 	rev := 10000
 	for i := 2; i <= rev; i++ {
 		s.Put([]byte("foo"), []byte(fmt.Sprintf("bar%d", i)), lease.NoLease)
 	}
 
 	hashCompactc := make(chan hashKVResult, 1)
-
 	donec := make(chan struct{})
-	var wg sync.WaitGroup
```
ahrtr, line 557: It seems it would better to keep this WaitGroup, because we need to make sure all goroutines have already finished their tasks before exiting the test case.

jbml: Thank you for taking a look @ahrtr 
I see what you mean, I will adjust this.

jbml: @ahrtr : I have added the WaitGroup back, and would need your approval for the test workflows to run again, please.
Thank you.

ahrtr:
Overall looks good to me, and I don't see any negative impact. 

> I don't know if this change doesn't negatively impact any of existing mechanisms using HashKV

ahrtr (approved):
LGTM

Thank you @jbml 

Leave @serathius to approve and merge this PR.

jbml:
Hello @serathius ,
Do you mind reviewing the PR when time permits please ?

ahrtr:
cc @spzala @ptabor @serathius @mitake

serathius (approved):

ahrtr:
@jbml Could you backport this PR to 3.5 and 3.4 if needed?

kkkkun:
I found this still do not backport to v3.4 and v3.5.  I would help to do this.

PR #14460, opened 2022-09-14
Integration_test: fix the flaky test TestV3AuthRestartMember

ahrtr:

cc @mitake @spzala @serathius


ahrtr:
The `WaitClientV3` reads key "/" by default, while the test case `TestV3AuthRestartMember` only grants permission on key `foo`. So the case was always failing with `permission denied`. 

It took me at least 3 hours to dig into the clientv3 package, but eventually it turns out to be a stupid mistake :) 

It should be working right now. Please take a look. @mitake

serathius (approved):
Awesome find!

PR #14476, opened 2022-09-16
Changelog: add item into 3.6 for the fix on panic in readonly serializable txn

ahrtr:
cc @spzala @serathius @mitake 

Refer to https://github.com/etcd-io/etcd/issues/14110 and https://github.com/etcd-io/etcd/pull/14149

### DIFF
--- a/CHANGELOG/CHANGELOG-3.6.md
+++ b/CHANGELOG/CHANGELOG-3.6.md
@@ -82,6 +82,7 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 - Fix [Durability API guarantee broken in single node cluster](https://github.com/etcd-io/etcd/pull/14400)
 - Fix [etcd fails to start after performing alarm list operation and then power off/on](https://github.com/etcd-io/etcd/pull/14419)
 - Fix [authentication data not loaded on member startup](https://github.com/etcd-io/etcd/pull/14358)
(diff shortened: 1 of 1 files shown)


serathius (approved):
Just a thought. Is having all bug fixes present in changelog for minor release unnecessary?

Should changelog for v3.6.0 describe changes from v3.5.0 or latest v3.5.X release? I think option 2 makes more sense. For example:
* v3.5.5 changelog lists changes from v3.5.4
* v3.6.0 changelog lists changes from v3.5.5

Approving this PR as it follows currently established process.

ahrtr:
> Just a thought. Is having all bug fixes present in changelog for minor release unnecessary?

I don't think we have hard limitation. But convention is to link to PRs instead of issues, and it looks good to me. 



> Should changelog for v3.6.0 describe changes from v3.5.0 or latest v3.5.X release? I think option 2 makes more sense.

It's a good point. I agree with you. The rules should be:
1. Each patch release should only include change against previous patch, such as 3.5.4 vs 3.5.5.
2. For the first release for each minor version, we should only describe new features.

So Proposed action:
1. Remove all bug fix items from changelog-3.6. 

What do you think? @serathius @spzala

ahrtr:
Just raised a separate PR to cleanup changelog 3.6. https://github.com/etcd-io/etcd/pull/14480 

So closing this PR.

PR #14481, opened 2022-09-16
makefile: Split fmt into separate verify commands

serathius:
cc @spzala @ahrtr

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,12 +29,12 @@ test-e2e-release: build
 
 # Static analysis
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -62,7 +62,44 @@ update-lint:
 
 .PHONY: update-fix
 update-fix:
-	./scripts/fix.sh
+	./scripts/fix.s
```
spzala, line 65: fix.sh? @serathius

spzala (commented):
Nice. One inline comment. Thanks @serathius

PR #14482, opened 2022-09-17
e2e: make maxstream test stable

fuweid:
submitConcurrentWatch use sleep 3s to wait for all the watch connections ready. When the number of connections increases, like 1000, the 3s is not enough and the test case becomes flaky.

In this commit, spawn curl process and check the ouput line with `created":true}}` to make sure that the connection has been initialized and ready to receive the events. It is reliable to test the following range request.

### DIFF
--- a/tests/e2e/v3_curl_maxstream_test.go
+++ b/tests/e2e/v3_curl_maxstream_test.go
@@ -28,6 +28,7 @@ import (
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/testutil"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
 )
 
 // NO TLS
@@ -183,19 +184,22 @@ func submitConcurrentWatch(cx ctlCtx, number int, wgDone *sync.WaitGroup, closeC
 		return nil
 	}
 
-	wgSchedule.Add(number)
-	for i := 0; i < number; i++ {
-		go func(i int) {
-			defer wgDone.Done()
+	testutils.ExecuteWithTimeout(cx.t, cx.getTestTimeout(), func() {
+		wgSchedule.Add(number)
 
-			if err := createWatchConnection(); err != nil {
-				cx.t.Fatalf("testV3CurlMaxStream watch failed: %d, error: %v", i, err)
-			}
+		for i := 0; i < number; i++ {
+			go func(i int) {
+				defer wgDone.Done()
 
-		}(i)
-	}
-	// make sure all goroutines have already been scheduled.
-	wgSchedule.Wait()
+				if err := createWatchConnection(); err != nil {
+					cx.t.Fatalf("testV3CurlMaxStream watch failed: %d, error: %v", i, err)
+				}
+			}(i)
+		}
+
+		// make sure all goroutines have already been scheduled.
+		wgSchedule.Wait()
+	})
 }
 
 func submitRangeAfterConcurrentWatch(cx ctlCtx, expectedValue string) {
--- a/tests/e2e/v3_curl_maxstream_test.go
+++ b/tests/e2e/v3_curl_maxstream_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"sync"
 	"syscall"
 	"testing"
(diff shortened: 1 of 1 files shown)


fuweid:
cc @ahrtr

fuweid:
https://github.com/etcd-io/etcd/actions/runs/3071777532/jobs/4963526956

It seems that the failure CI is not related to this change. Repush it and rerun it again.

tests/e2e/v3_curl_maxstream_test.go
```diff
@@ -162,22 +150,49 @@ func submitConcurrentWatch(cx ctlCtx, number int, wgDone *sync.WaitGroup, errCh
 	var wgSchedule sync.WaitGroup
 	wgSchedule.Add(number)
```
ahrtr, line 151: It seems that we can remove `wgSchedule`, there is no reason to keep it because once it gets the `"created":true` response, then the goroutine must have been scheduled.

tests/e2e/v3_curl_maxstream_test.go
```diff
@@ -162,22 +150,49 @@ func submitConcurrentWatch(cx ctlCtx, number int, wgDone *sync.WaitGroup, errCh
...
+			if err != nil {
+				return fmt.Errorf("failed to spawn: %w", err)
+			}
+			defer proc.Stop()
+
+			// make sure that watch request has been created
+			expectedLine := `"created":true}}`
+			_, lerr := proc.ExpectWithContext(context.TODO(), expectedLine)
+			if lerr != nil {
+				return fmt.Errorf("%v %v (expected %q). Try EXPECT_DEBUG=TRUE", args, lerr, expectedLine)
```
ahrtr, line 176: We can fail the case directly, no need to return an error.

tests/e2e/v3_curl_maxstream_test.go
```diff
@@ -162,22 +150,49 @@ func submitConcurrentWatch(cx ctlCtx, number int, wgDone *sync.WaitGroup, errCh
...
+			cluster := cx.epc
+			member := cluster.Procs[rand.Intn(cluster.Cfg.ClusterSize)]
+			curlReq := e2e.CURLReq{Endpoint: "/v3/watch", Value: string(watchData)}
+
+			args := e2e.CURLPrefixArgs(cluster.Cfg, member, "POST", curlReq)
+			proc, err := e2e.SpawnCmd(args, nil)
+			if err != nil {
+				return fmt.Errorf("failed to spawn: %w", err)
+			}
+			defer proc.Stop()
```
ahrtr, line 170: It seems not good to copy the detailed implementation from the framework, but we do not have better way for now.

tests/e2e/v3_curl_maxstream_test.go
```diff
@@ -109,31 +105,23 @@ func testV3CurlMaxStream(t *testing.T, reachLimit bool, opts ...ctlOption) {
...
-	default:
-	}
+	closeServerCh := make(chan struct{})
+	submitConcurrentWatch(cx, int(concurrentNumber), &wg, closeServerCh)
+	submitRangeAfterConcurrentWatch(cx, expectedResponse)
 
-	// Step 5: Close the cluster
+	// Step 4: Close the cluster
 	t.Log("Closing test cluster...")
+	close(closeServerCh)
```
ahrtr, line 117: It isn't correct to close the channel directly, because `submitConcurrentWatch`  might not finish the task yet.

tests/e2e/v3_curl_maxstream_test.go
fuweid: Thanks for the review! 

The `wgSchedule` is used to make sure that all the goroutines are ready. If we remove this, we need to setup curl command one by one. It might take a long time to setup the ton of connections. I think we can keep the `wgSchedule` and run case in the efficient way. WDYT?

tests/e2e/v3_curl_maxstream_test.go
fuweid: Actually, I handle it in defer. I will update it to make it clear.

tests/e2e/v3_curl_maxstream_test.go
fuweid: The `e2e.SpawnWithExpect` doesn't work well in this case. And the `SpawnWithExpect` is just a helper to write case eaiser. But the `e2e.SpawnCmd` is also exported as helper, right? I was thinking that it is ok to use it 😂

tests/e2e/v3_curl_maxstream_test.go
fuweid: If I understand correctly, the `submitConcurrentWatch` is used to increase the active streams before hiting the quota, which depends on the case. If the case wants `Range` request fails with timeout, the `submitConcurrentWatch` will hit the quota of `MaxConcurrentStream`. Otherwise, the `submitConcurrentWatch` will left one availiable stream to `Range` request. 

So, in this commit, the `submitConcurrentWatch` can make sure that all the active streams are ready before return. And the `submitRangeAfterConcurrentWatch` will verify that the `MaxConcurrentStreams` quota setting is working. After `submitRangeAfterConcurrentWatch` finish, it is ready to close the server and let all the watch connections closes.( It is not attempt to verify the watch event, and the `Range` request doesn't create event)

But the watch connection might exit somehow, like oom killer. It is unexpected. So I use a channel `closeServerCh` to mark that the server is ready to close. If the watch connection is close and the `closeServerCh` is still open, it should fail.

```go
                       // hold the connection and wait for server shutdown
			perr := proc.Close()

			// curl process will return
			select {
			case <-closeCh:
			default:
				// perr could be nil.
				return fmt.Errorf("unexpected connection close before server closes: %v", perr)
			}
			return nil
```

Does it make sense to you?

ahrtr: I missed the info that `submitRangeAfterConcurrentWatch` has already been called before closing the channel.  It makes sense.

tests/e2e/v3_curl_maxstream_test.go
ahrtr: Please consider to add a timeout, because `submitConcurrentWatch` may be blocked for a long time. Probably we can refactor the `submitConcurrentWatch` using `ExecuteUntil`. FYI. https://github.com/etcd-io/etcd/blob/5344085338a8dc57eb87f8676dea66401185b362/tests/framework/testutils/execute.go#L32

fuweid:
@ahrtr I updated about the error handle in `submitConcurrentWatch` function. Please take a look~ Thanks

tests/e2e/v3_curl_maxstream_test.go
fuweid: Thanks for the hint. Let me try it.

fuweid: Updated with testutilslExecuteWithTimeout and the timeout is from ctlCtx.timeout

ahrtr (approved):
LGTM

Great work, thank you @fuweid

fuweid:
> LGTM
> 
> 
> 
> Great work, thank you @fuweid 

thanks for the review!

PR #14483, opened 2022-09-17
Doc: Add feature graduation and deprecation policies

spzala:
Related https://github.com/etcd-io/etcd/pull/14428

### DIFF
--- a/Documentation/contributor-guide/features.md
+++ b/Documentation/contributor-guide/features.md
@@ -34,6 +34,7 @@ Any new enhancements to the etcd are typically added as an experimental feature.
 - Open an issue
     - It must provide a clear need for the proposed feature.
     - It should list development work items as checkboxes. There must be one work item towards future graduation to the stable future.
+    - Label the issue with `type/feature` and `experimental`.
     - Keep the issue open for tracking purpose until a decision is made on graduation.
 - Open a Pull Request (PR)
     - Provide unit tests. Integreation tests are also recommended as possible.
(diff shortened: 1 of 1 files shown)


spzala:
cc @ahrtr @serathius

Documentation/contributor-guide/features.md
```diff
@@ -34,6 +34,7 @@ Any new enhancements to the etcd are typically added as an experimental feature.
 - Open an issue
     - It must provide a clear need for the proposed feature.
     - It should list development work items as checkboxes. There must be one work item towards future graduation to the stable future.
+    - Label the issue with `type/feature` and `experimental`.
```
spzala, line 37: I am proposing two new labels, experimental and deprecated. They should help with tracking. cc @serathius @ahrtr

Documentation/contributor-guide/features.md
```diff
@@ -45,6 +46,38 @@ Any new enhancements to the etcd are typically added as an experimental feature.
...
+- Find an issue that was used to enable the experimental feature initially. One way to find such issues is to search for issues with `type/feature` and `experimental` labels.
+- Fix any known open issues against the feature.
+- Make sure the feature was enabled for at least one previous release. Check the PR(s) reference from the issue to see when the feature related code changes were merged.
+
+#### Provide implementation
+If an experimental feature is found ready for graduation to the stable stage, open a Pull Request (PR) with the following changes.
+- Add robust e2e tests if not already provided.
+- Add a new stable feature name identical to the experimental feature name but without the `--experimental` prefix.
+- Deprecate the experimental feature following the [feature deprecation policy](#Deprecating-a-feature).
+- Implementation must ensure that both the graduated and deprecated experimental features work as expected. Note that both these features will co-exist for the timeframe described in the feature deprecation policy.
```
ahrtr, line 64: Just both the experimental and stable flags co-exist, and actually they are actually the same feature. If users are still using the experimental flag, we should print a warning log.

Documentation/contributor-guide/features.md
```diff
@@ -45,6 +46,38 @@ Any new enhancements to the etcd are typically added as an experimental feature.
...
+- Enable the graduated feature by default if needed.
+
+At least two maintainers must approve the work. Patch releases should not be considered for graduation.
+
+### Deprecating a feature
+
+#### Experimental
+An experimental feature deprecates when it graduates to the stable stage.
```
ahrtr, line 72: An experimental goes to stable stage or removed directly. 

I don't think there is a deprecation stage for an experimental feature. Of course we can add a warning before removing it.

Documentation/contributor-guide/features.md
spzala: @ahrtr yes, and that warning would be "Deprecated. Use <the stable feature> instead."

Documentation/contributor-guide/features.md
spzala: Yes, today we just remove experiment feature when it graduates. In line what we discussed in the previous PR on not surprising users by keeping experimental feature around for one release after it graduate. So, I am proposing deprecation phase to keep features coexist (as we discussed above) with experimental as deprecated. Does that sound good? I agree that experimental feature can get removed directly without being graduated and in that case there is no deprecation. I will make that change. Thanks @ahrtr

Documentation/contributor-guide/features.md
ahrtr: ack

Documentation/contributor-guide/features.md
ahrtr: > So, I am proposing deprecation phase to keep features coexist (as we discussed above) with experimental as deprecated. Does that sound good?

Makes sense to deprecate the experimental flag by adding a comment something like "`Deprecated xxx, use xxx instead`"

spzala: Thanks @ahrtr !! Also, for the case when experimental feature don't get to graduation, we have already covered it in the previous PR `It can be removed in the next minor or major release without following the feature deprecation policy unless it graduates to the stable future.` under the [Overview section](https://github.com/etcd-io/etcd/blob/bbe848f5c46679a11d30264ddaa10247fcb7a124/Documentation/contributor-guide/features.md#overview)

serathius (approved):
Looks good

Documentation/contributor-guide/features.md
```diff
@@ -45,6 +46,38 @@ Any new enhancements to the etcd are typically added as an experimental feature.
...
 
-### Graduating an Experimental feature to Stable (TODO - spzala)
+### Graduating an Experimental feature to Stable
 
-### Deprecating a feature (TODO - spzala)
+It is important that experimental features don't get stuck in that stage. They should be revisited and moved to the stable stage following the graduation steps as described here.
+
+#### Locate graduation candidate
+Decide if an experimental feature is ready for graduation to the stable stage.
+- Find an issue that was used to enable the experimental feature initially. One way to find such issues is to search for issues with `type/feature` and `experimental` labels.
```
ahrtr, line 55:
```suggestion
- Find the issue that was used to enable the experimental feature initially. One way to find such issues is to search for issues with `type/feature` and `experimental` labels.
```

Documentation/contributor-guide/features.md
```diff
@@ -45,6 +46,38 @@ Any new enhancements to the etcd are typically added as an experimental feature.
...
+- Fix any known open issues against the feature.
+- Make sure the feature was enabled for at least one previous release. Check the PR(s) reference from the issue to see when the feature related code changes were merged.
+
+#### Provide implementation
+If an experimental feature is found ready for graduation to the stable stage, open a Pull Request (PR) with the following changes.
+- Add robust e2e tests if not already provided.
+- Add a new stable feature name identical to the experimental feature name but without the `--experimental` prefix.
```
ahrtr, line 62:
```suggestion
- Add a new stable feature flag identical to the experimental feature flag but without the `--experimental` prefix.
```

Documentation/contributor-guide/features.md
ahrtr, line 64:
```suggestion
- Implementation must ensure that both the graduated and deprecated experimental feature flags work as expected. Note that both these flags will co-exist for the timeframe described in the feature deprecation policy.
```

Documentation/contributor-guide/features.md
spzala: @ahrtr addressing it in the new update. Thanks!

ahrtr (approved):
LGTM

Thanks @spzala

PR #14486, opened 2022-09-17
Test: increase the TestKVDelete's timeout to 15s

ahrtr:
The test case `TestKVDelete` is flaky. Sometimes it just needs more time to execute the test case, so increase the timeout from 10s to 15s.

cc  @spzala


spzala (approved):
lgtm, we can watch what happens with this change. Thanks @ahrtr

PR #14488, opened 2022-09-19
Improve static analysis fixing scripts

serathius:
cc @ahrtr @spzala

### DIFF
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -13,4 +13,4 @@ jobs:
         with:
           version: v1.49.0
       - run: make verify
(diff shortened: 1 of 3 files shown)


ahrtr:
Can we combine all the fix_*.sh scripts into one? It doesn't make much sense to have 5 separate small script files, and all of have duplicated  code something like below,

```
set -e

# Top level problems with modules can lead to test_lib being not functional
go mod tidy

source ./scripts/test_lib.sh
```

serathius:
Ok, I have changed the PR so we don't split the script.

ahrtr (approved):
LGTM

PR #12580, opened 2020-12-24
Switch from dgrijalva/jwt-go to form3tech-oss/jwt-go.

dlorenc:
dgrijalva/jwt-go has been abandoned and contains several serious
security issues. Most projects are now switching to the form3tech fork.

See https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMDGRIJALVAJWTGO-596515 for
info on the issues.

PR #14489, opened 2022-09-19
Replace github.com/form3tech-oss/jwt-go with https://github.com/golang-jwt/jwt

dims:
Cherry-pick https://github.com/etcd-io/etcd/pull/13256 to release-3.5 branch

form3tech fork is outdated as well :( (introduced in https://github.com/etcd-io/etcd/pull/12580)

So pick up the changes from @Haimantika in #13256 from main to release-3.5 (At one point we were waiting for 3.6, but given that otel changes got picked to 3.5, we should do this as well.

Longer term plan was to ensure k/k only depends on `github.com/golang-jwt/jwt`, see:
- https://github.com/kubernetes/kubernetes/issues/104007
- https://github.com/kubernetes/kubernetes/issues/106422

Also note that form3tech is no longer active as well. So better to switch to something where we can get help if needed.

Signed-off-by: haimantika mitra <haimantikamitra@gmail.com>
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/server/auth/jwt.go
+++ b/server/auth/jwt.go
@@ -21,7 +21,7 @@ import (
 	"errors"
 	"time"
 
-	jwt "github.com/form3tech-oss/jwt-go"
+	"github.com/golang-jwt/jwt/v4"
 	"go.uber.org/zap"
 )
 
--- a/server/auth/options.go
+++ b/server/auth/options.go
@@ -21,7 +21,7 @@ import (
 	"io/ioutil"
 	"time"
 
-	jwt "github.com/form3tech-oss/jwt-go"
+	"github.com/golang-jwt/jwt/v4"
 )
 
 const (
--- a/server/go.mod
+++ b/server/go.mod
@@ -6,8 +6,8 @@ require (
 	github.com/coreos/go-semver v0.3.0
 	github.com/coreos/go-systemd/v22 v22.3.2
 	github.com/dustin/go-humanize v1.0.0
-	github.com/form3tech-oss/jwt-go v3.2.3+incompatible
 	github.com/gogo/protobuf v1.3.2
+	github.com/golang-jwt/jwt/v4 v4.4.2
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
 	github.com/golang/protobuf v1.5.2
 	github.com/google/btree v1.0.1
--- a/bill-of-materials.json
+++ b/bill-of-materials.json
@@ -135,15 +135,6 @@
 		]
 	},
 	{
-		"project": "github.com/form3tech-oss/jwt-go",
-		"licenses": [
-			{
-				"type": "MIT License",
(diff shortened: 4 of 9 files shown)


spzala (approved):
Makes sense. lgtm. Thanks @dims :) 
Rerunning failed tests which seems not related to these changes.

spzala:
cc @ahrtr @serathius @mitake

server/go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/coreos/go-semver v0.3.0
 	github.com/coreos/go-systemd/v22 v22.3.2
 	github.com/dustin/go-humanize v1.0.0
-	github.com/form3tech-oss/jwt-go v3.2.3+incompatible
 	github.com/gogo/protobuf v1.3.2
+	github.com/golang-jwt/jwt v3.2.2+incompatible
```
ahrtr, line 10: `github.com/golang-jwt/jwt` starts to support go mod starting from 4.4.0. We might need to use v4.4.2 instead.

server/auth/jwt.go
```diff
@@ -21,7 +21,7 @@ import (
 	"errors"
 	"time"
 
-	jwt "github.com/form3tech-oss/jwt-go"
+	jwt "github.com/golang-jwt/jwt"
```
ahrtr, line 24: After bumpping to 4.4.2, we need to import `"github.com/golang-jwt/jwt/v4"` instead. 

FYI. https://github.com/golang-jwt/jwt/blob/bfea432b1a9da383509086a2141c06dc103e82f9/cmd/jwt/main.go#L20

server/auth/options.go
```diff
@@ -21,7 +21,7 @@ import (
 	"io/ioutil"
 	"time"
 
-	jwt "github.com/form3tech-oss/jwt-go"
+	jwt "github.com/golang-jwt/jwt"
```
ahrtr, line 24: ditto

ahrtr (approved):
LGTM with minor comments.

Thank you @dims 

We can address the minor comments in a separate PR. If you do not have bandwidth, I can  take care of it later in a separate PR.

dims:
ok, should be ready now! thanks.

@ahrtr please feel free to update master to this same version.

ahrtr:
> ok, should be ready now! thanks.
> 
> @ahrtr please feel free to update master to this same version.

Sure. Thanks @dims .

Usually we keep the backporting PR as it's. Specifically I would suggest to bump to 4.4.2 in the main branch firstly, and cherry pick to 3.5 and 3.4 later.  But I won't insist on that.

Anyway, let me take care of the main branch to bumping 4.4.2. Thanks again.

dims:
thanks @ahrtr !

ahrtr:
FYI. I just delivered a PR to take care of the main branch https://github.com/etcd-io/etcd/pull/14491

PR #14479, opened 2022-09-16
etcd: modify declaring empty slices 

demoManito:
declare an empty slice to `var s []int` replace `s :=[]int{}`, https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

Issue #14164, opened 2022-06-26
Enhance the static-analysis workflow


ahrtr:
Some issues were found using [revive](https://github.com/mgechev/revive), such as,
* https://github.com/etcd-io/etcd/issues/14162
* https://github.com/etcd-io/etcd/pull/14160
* https://github.com/etcd-io/etcd/pull/14163

We should enhance the [static-analysis](https://github.com/etcd-io/etcd/blob/main/.github/workflows/static-analysis.yaml) workflow. The [revive](https://github.com/mgechev/revive) has already been integrated into [golangci-lint](https://github.com/golangci/golangci-lint). So I think we can enhance the static-analysis using the `golangci-lint `?

cc @serathius @ptabor @spzala @chavacava @ldez @jirfag @ernado

chavacava:
`revive` has more than 65 rules and, of course, there is some overlap with those of the other linters `golangci-lint` can run.

When looking for potential bugs I use the following `revive` configuration: 
```toml
ignoreGeneratedHeader = false
severity = "warning"
confidence = 0.8
errorCode = 1
warningCode = 0

[rule.constant-logical-expr]
[rule.time-equal]
[rule.unreachable-code]
[rule.struct-tag]
[rule.modifies-value-receiver]
[rule.bool-literal-in-expr]
[rule.range-val-in-closure]
[rule.range-val-address]
[rule.datarace]
[rule.string-of-int]
[rule.unconditional-recursion]
[rule.identical-branches]
```

cenkalti:
I am willing to work on this task is no one else has started yet. I can integrate `golangci-lint`. After integration is done we can talk about which linter to turn on/off and rules.

ramses:
Yesterday I finished exactly this for another project — i.e., golangci-lint for the project, then added a github action.

I've just noticed that we do not have the following:

- `.golangci.yml` configuration.
- `make` target for linting  — this and the above bullet would enable repeatable local linting.

Also, the static analysis workflow calls the `scripts/test.sh`. Instead, we could use the `golangci-lint` github-action for static analysis.

Thoughts?

cenkalti:
It's better to use Github actions instead of writing more script code.

ramses:
> It's better to use Github actions instead of writing more script code.

I'm not recommending more script code.

I am recommending LESS script code.

A configuration file for `golangci-lint` would enable local tests.

Using the golangci-lint-github-action would simplify `scripts/test.sh`.

Your thoughts?

[...]

PR #14494, opened 2022-09-20
etcd: remove redundant type conversion

demoManito:


ahrtr (approved):
LGTM

Thank you @demoManito

serathius:
Thanks for your hard work finding and fixing those issue. However I would like to repeat the comment from previous PR https://github.com/etcd-io/etcd/pull/14479#issuecomment-1249169815

There is little sense in fixing broken code conventions that cannot be verified automatically by static analysis. Fact that those mistakes were in codebase means that they have high chance to be reintroduced again. We can fix them, however not everything can be caught by code review. Cycle will just repeat. 

Better use of time would be to look into improving static analysis like in  https://github.com/etcd-io/etcd/issues/14164 or look for other issues with codebase.

serathius (approved):

PR #14505, opened 2022-09-22
etcdserve: revert the etcdserver side change for the data loss in one node cluster

ahrtr:
Since the raft side change https://github.com/etcd-io/etcd/pull/14413 has been merged, so we need to revert the etcdserver side change https://github.com/etcd-io/etcd/pull/14400.

cc @serathius @spzala

### DIFF
--- a/server/etcdserver/raft.go
+++ b/server/etcdserver/raft.go
@@ -209,14 +209,6 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 
 				updateCommittedIndex(&ap, rh)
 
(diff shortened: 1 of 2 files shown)


serathius (approved):
Have you verified that durability issue cannot be reproduced? Please check out https://github.com/etcd-io/etcd/pull/14398 which I used to reproduce it orginally.

ahrtr:
Yes, already confirmed the issue can't be reproduced using steps in https://github.com/etcd-io/etcd/issues/14370#issuecomment-1229702988

PR #14506, opened 2022-09-22
wal: Fix the `walWriteBytes` metric

SimFG:
In the `encode`, the `flush` should not add the `walWriteBytes` metric, because the write byte has counted when executing the `write`.


ahrtr (approved):
LGTM

Nice catch! Thank you @SimFG

PR #14536, opened 2022-09-29
etcdserver: get peer's hash using the same revision as the value used by leader

ahrtr:
Currently the leader uses its latest revision to get peers' hashes, it isn't correct, instead it should use the same revision which it uses to calculate the hash locally. Although it has the same results, but it isn't semantically correct. 

cc @serathius @spzala

### DIFF
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -93,18 +93,18 @@ func (cm *corruptionChecker) InitialCheck() error {
 		zap.Duration("timeout", cm.hasher.ReqTimeout()),
 	)
 
-	h, rev, err := cm.hasher.HashByRev(0)
+	h, _, err := cm.hasher.HashByRev(0)
 	if err != nil {
 		return fmt.Errorf("%s failed to fetch hash (%v)", cm.hasher.MemberId(), err)
 	}
-	peers := cm.hasher.PeerHashByRev(rev)
+	peers := cm.hasher.PeerHashByRev(h.Revision)
 	mismatch := 0
 	for _, p := range peers {
 		if p.resp != nil {
 			peerID := types.ID(p.resp.Header.MemberId)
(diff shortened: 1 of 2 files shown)


server/etcdserver/corrupt.go
```diff
@@ -212,17 +212,6 @@ func (cm *corruptionChecker) PeriodicCheck() error {
 		}
 		checkedCount++
 
-		// leader expects follower's latest revision less than or equal to leader's
-		if p.resp.Header.Revision > rev2 {
```
serathius, line 216: PR doesn't state why remove this check. It seems unrelated so it would be good to confirm.

Goal of this check is to detect cases where follower got corrupted in a way that would lead them to have higher revision number. In such cases follower might never respond to leader request for `HashByRev` follower might have already compacted this revision. To detect such cases leader verifies that follower current revision is lower than its own revision.

Why would you want to remove this check?

ahrtr: Please note that if there are new K/V data during the process, then the follower's rev may bigger than the leader.

serathius: Nope, please check order of the operations. We first request and collect hashes from peers with their latest revision, and only after that we again check local latest revision. 
Due to linearizability of etcd to consecutive requests should result in revision not decreasing.

ahrtr: In theory, it's still possible for the follower's rev to be bigger than the leader, although it's not easy to reproduce the situation in real production environment. The reason is the peers communication do not go through raft.

ahrtr: Let me take one step back. Just reverted the change, the only reason is that it's unrelated change.

ahrtr:
The PR should be ready to go. Please take a look, thx. cc  @spzala @serathius

serathius (approved):

PR #14537, opened 2022-09-29
Add a new field hash_revision into HashKVResponse

ahrtr:
Added one new field "HashRevision" into [HashKVResponse](https://github.com/etcd-io/etcd/blob/54f9483e725f7300440fcabd9fbc4b94c81da3bd/api/etcdserverpb/rpc.pb.go#L1630). The field is populated when responding to leader or client's `HashKV` requests. It's the revision up to which the returned hash is calculated. 

Two reasons why adding this field:
1. Users/clients may get inconsistent hashes (which might be expected) and accordingly take wrong action. Please see https://github.com/etcd-io/etcd/pull/14510#issuecomment-1261495210
2. The [KeyValueHash](https://github.com/etcd-io/etcd/blob/d379e6221e10c32d83dc72ce39e0b19b3f00366f/server/storage/mvcc/hash.go#L81) has three fields (`Hash`, `CompactRevision` and `Revision`), and they are an integral whole. It would be better to return all the three fields to client or other member (e.g. leader). 

cc @serathius @spzala @ptabor


serathius:
As mentioned in https://github.com/etcd-io/etcd/pull/14510#issuecomment-1261951651 adding the field itself will not improve the user expirience much. It would just change:
```
+------------------------+------------+
|        ENDPOINT        |    HASH    |
+------------------------+------------+
|  http://127.0.0.1:2379 | 12345      |
| http://127.0.0.1:22379 | 54321      |
| http://127.0.0.1:32379 | 34521      |
+------------------------+------------+
```
to 
```
+------------------------+------------+----------+
|        ENDPOINT        |    HASH    | REVISION |
+------------------------+------------+----------+
|  http://127.0.0.1:2379 | 12345      | 132      |
| http://127.0.0.1:22379 | 54321      | 133      |
| http://127.0.0.1:32379 | 34521      | 134      |
+------------------------+------------+----------+
```

If we really want to improve user experience. Let's rethink how `etcdctl endpoint hashkv -w table --cluster` should work to make it useful.

ahrtr:
The etcdctl side change is the next step I will do.

ahrtr:
> Still each member returns different hash.

This isn't a problem. The point is just to let users know the rev for each hash. Of course, we need to enhance `etcdctl` in a following PR.

scripts/etcd_version_annotations.txt
```diff
@@ -166,6 +166,7 @@ etcdserverpb.HashKVRequest.revision: ""
 etcdserverpb.HashKVResponse: "3.3"
 etcdserverpb.HashKVResponse.compact_revision: ""
 etcdserverpb.HashKVResponse.hash: ""
+etcdserverpb.HashKVResponse.hash_revision: ""
```
ptabor, line 169: I think that this field should be annotated that it was added in v3.6. 
Seems it requires a proto.option in rpc.proto file.

ptabor:
> The etcdctl side change is the next step I will do.

What's your plan for this ?  Pre-fetching revisions and double-lookup on min ?

ahrtr:
> > The etcdctl side change is the next step I will do.
> 
> What's your plan for this ? Pre-fetching revisions and double-lookup on min ?

I will add the `HashRevision` into the output, just similar to https://github.com/etcd-io/etcd/pull/14537#issuecomment-1261962349 pointed out.

scripts/etcd_version_annotations.txt
ahrtr: Thanks @ptabor for pointing out this. 

But this field isn't persisted in the WAL, nor the boltDB. Should we still annotate that it's added in v3.6? @serathius @ptabor

ahrtr: Added annotation `[(versionpb.etcd_version_field)="3.6"]` anyway although it isn't necessary for now. But it will not do any harm for now, and it is also safe in case it will be persisted in WAL in future.

ahrtr:
After second thought, I would like to adopt @ptabor 's comment above https://github.com/etcd-io/etcd/pull/14537#issuecomment-1268776992 .

The solution to enhance etcdctl (in a separate PR) would be:
1. If users explicitly set a revision such as `etcdctl endpoint hashkv -w table --cluster --rev 10000`, then etcdctl just gets the hash of each member using the provided rev.
2. Otherwise if rev isn't specified, then pre-fetching revisions and double-lookup on min(rev). 

Please let me know if you have any concern or comment.

spzala (approved):
@ahrtr thanks, lgtm but I am wondering if you think good idea to add a test coverage for the changes in the `corrupt_test.go`? Doing it separately is okay too.

ahrtr:
> @ahrtr thanks, lgtm but I am wondering if you think good idea to add a test coverage for the changes in the `corrupt_test.go`? Doing it separately is okay too.

Thanks @spzala . Yes, it makes sense to me. Will do it separately.

ahrtr:
Just rebased this PR in case there are any potential conflicts.

spzala:
> > @ahrtr thanks, lgtm but I am wondering if you think good idea to add a test coverage for the changes in the `corrupt_test.go`? Doing it separately is okay too.
> 
> Thanks @spzala . Yes, it makes sense to me. Will do it separately.

Thanks @ahrtr

PR #14538, opened 2022-09-29
raft: broadcast MsgApp on auto-leave joint config proposal

nvanbenschoten:
This commit ensures that the raft leader eagerly broadcasts a MsgApp to each follower when initiating an automatic transition out of a joint configuration. This had been missed previously, which could lead to delayed completion of an auto-transition.

cc. @tbg


tbg (approved):
Thanks!

ahrtr (approved):
Thank you @nvanbenschoten

PR #14539, opened 2022-09-29
raft: update prevHardSt on Ready accept, not advance

nvanbenschoten:
This commit updates the `RawNode`'s `prevHardSt` to the new HardState in `acceptReady` instead of on `Advance`. This aligns the handling of `prevHardSt` with that of `prevSoftSt` (and other fields like `msgs`) and simplifies the logic in `Advance`.

cc. @tbg


tbg (approved):
Thanks!

ahrtr (approved):
LGTM

Thank you @nvanbenschoten

PR #14545, opened 2022-10-03
raft: simplify auto-leave joint config on entry application logic

nvanbenschoten:
This commit simplifies the logic added in 37c7e4d to auto-leave joint configurations. It does so by making the following adjustments to the code:

- remove the `oldApplied <= r.pendingConfIndex` condition. This does not seem necessary. When a node first attempts to auto-leave a joint config, it will bump `r.pendingConfIndex` when proposing. In cases where `oldApplied >= r.pendingConfIndex`, the proposal must have already been applied. Reviewers should double check this.
- use raft.Step instead of custom proposal code. This code was already present in stepLeader, so there was no reason to duplicate it. This would have avoided bugs like the one we fixed in #14538.
- use `confChangeToMsg` to generate message, to centralize the creation of all `MsgProp{EntryConfChange}` messages.

cc. @tbg 

### DIFF
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -548,26 +548,29 @@ func (r *raft) advance(rd Ready) {
 	// new Commit index, this does not mean that we're also applying
 	// all of the new entries due to commit pagination by size.
 	if newApplied := rd.appliedCursor(); newApplied > 0 {
-		oldApplied := r.raftLog.applied
 		r.raftLog.appliedTo(newApplied)
 
-		if r.prs.Config.AutoLeave && oldApplied <= r.pendingConfIndex && newApplied >= r.pendingConfIndex && r.state == StateLeader {
+		if r.prs.Config.AutoLeave && newApplied >= r.pendingConfIndex && r.state == StateLeader {
 			// If the current (and most recent, at least for this leader's term)
 			// configuration should be auto-left, initiate that now. We use a
 			// nil Data which unmarshals into an empty ConfChangeV2 and has the
 			// benefit that appendEntry can never refuse it based on its size
 			// (which registers as zero).
-			ent := pb.Entry{
-				Type: pb.EntryConfChangeV2,
-				Data: nil,
+			m, err := confChangeToMsg(nil)
+			if err != nil {
+				panic(err)
 			}
-			// There's no way in which this proposal should be able to be rejected.
-			if !r.appendEntry(ent) {
-				panic("refused un-refusable auto-leaving ConfChangeV2")
+			// NB: this proposal can't be dropped due to size, but can be
+			// dropped if a leadership transfer is in progress. We'll keep
+			// checking this condition on each applied entry, so either the
+			// leadership transfer will succeed and the new leader will leave
+			// the joint configuration, or the leadership transfer will fail,
+			// and we will propose the config change on the next advance.
+			if err := r.Step(m); err != nil {
+				r.logger.Debugf("not initiating automatic transition out of joint configuration %s: %v", r.prs.Config, err)
+			} else {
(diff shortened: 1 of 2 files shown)


tbg (approved):

spzala (commented):
lgtm, ru-running failed tests.

serathius:
Maybe a naive question, how do we ensure that changes in raft library are backward compatible? Raft library is automatically picked up by etcd release, which has strict compatibility policies that might not be easy to enforce at library level.

Etcd supports clusters with 1 version skew to provide zero downtime upgrade. For example when during upgrade of regional clusters from v3.5.X to v3.6.X therere will be multiple states where leader and followers will run different versions. 

Are we sure that all recent changes to raft will not cause issues when running side by side with older version? In cases where leader can have either older and newer raft code.

tbg:
The changes I see happening in raft are generally local in nature, i.e. they don't change the behavior of the raft peer across RPC boundaries. There are changes that did have cross-version impacts, such as adding joint consensus. Here the strategy was to use separate mechanisms and to leave it to the users to make sure that they use the new functionality only when they  know that all peers are running the same/a compatible version.

But, to the best of my knowledge no mixed-version testing exists. We get some implicit mixed-version testing through CockroachDB, where we test adjacent releases and thus also two different raft versions, though these just happen to be the two versions these CRDB releases use, and we don't follow any particular upgrade pattern. I imagine etcd has some mixed-version testing as well which might provide more applicable testing.

raft/raft.go
```diff
@@ -548,26 +548,29 @@ func (r *raft) advance(rd Ready) {
...
-			if !r.appendEntry(ent) {
-				panic("refused un-refusable auto-leaving ConfChangeV2")
+			// NB: this proposal can't be dropped due to size, but can be
+			// dropped if a leadership transfer is in progress. We'll keep
+			// checking this condition on each applied entry, so either the
+			// leadership transfer will succeed and the new leader will leave
+			// the joint configuration, or the leadership transfer will fail,
+			// and we will propose the config change on the next advance.
+			if err := r.Step(m); err != nil {
+				r.logger.Debugf("not initiating automatic transition out of joint configuration %s: %v", r.prs.Config, err)
```
ahrtr, line 570: Shouldn't this be an error or warning? 

And also I am curious why the log at https://github.com/etcd-io/etcd/blob/e5790d204c6a3d19941cb11a1d7d9b8901fee81f/raft/raft.go#L646-L649 isn't an warning or error?

tbg: It's valid for there to be an error here if a leadership transfer is in progress, so I'm not sure this qualifies as warning (it might pop up spuriously even though everything is normal).

Re: the other snippet, this is not expected in normal operations, so I would agree with it becoming a warning. It's something the operator might conceivably want to know about.

The bottom line is, we're not very principled about log severities, but never too late to start. Personally I think we should also change `raft` to use structured events instead of logging, and then drive the logging off the events.

tbg:
Still LGTM, @ahrtr is there anything left to do here?

raft/raft.go
ahrtr: Makes sense. 

Just submitted another PR to change the log to warning.  https://github.com/etcd-io/etcd/pull/14698

ahrtr: I am curious is cockroachDB using joint consensus?

ahrtr (approved):

ahrtr:
Thank you @nvanbenschoten

tbg:
@ahrtr 

> I am curious is cockroachDB using joint consensus?

Yes, we introduced it to `etcd/raft` for that purpose.

PR #14515, opened 2022-09-24
upate:use google/btree in the genric way

spongecaptain:
mvcc: use google/btree in the genric way

As etcd bumped go1.19, using generic instead of cast interface{} to some type has better operation efficiency.

Signed-off-by: wathenjiang <wathenjiang@tencent.com>

PR #14551, opened 2022-10-05
Update changelog 3.6 to clearly requiring applications which depend on etcd 3.6 packages must be built with go version >= 1.18

ahrtr:
etcdserver starts to using generic in https://github.com/etcd-io/etcd/pull/14515, so any applications which depend on etcdserver must be built with go version >= 1.18.

cc @serathius @spzala @ptabor

### DIFF
--- a/CHANGELOG/CHANGELOG-3.6.md
+++ b/CHANGELOG/CHANGELOG-3.6.md
@@ -12,7 +12,8 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 
 - `etcd` will no longer start on data dir created by newer versions (for example etcd v3.6 will not run on v3.7+ data dir). To downgrade data dir please check out `etcdutl migrate` command.
 - `etcd` doesn't support serving client requests on the peer listen endpoints (--listen-peer-urls). See [pull/13565](https://github.com/etcd-io/etcd/pull/13565).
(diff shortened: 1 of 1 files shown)


ptabor:
How about similifying the message ?
 - any package in etcd-3.6 requires go 1.18 (as it's pretty likely we will take some 1.18 dependency soon)
  - if user want's to use go 1.17 with etcd-3.6, they still can link to older client (v3.5) that still will connect to etcd-3.6

ptabor:
There is Benjamin's answer to my comment that I missposted in the other thread: 
https://github.com/etcd-io/etcd/pull/14515#issuecomment-1268526845

spzala (commented):
@ahrtr thanks and good idea to update changelog! One comment I have is the sentence `because etcdserver starts using generic.` looks incomplete and may not be read clearly. I would suggest to modify it to `because etcdserver now use generic b-tree.`

ahrtr:
Thanks both @ptabor and @spzala for the comment. I just updated the PR per the discussion in https://github.com/etcd-io/etcd/pull/14515 to use simplified message. I added more detailed info in the commit message.

spzala (approved):
This looks better @ahrtr thanks! One nit if you want to make change is to change `3.6` to `v3.6` and `1.18` to `v1.18` as just two line above we mention versions as `etcd v3.6 will not run on v3.7`.  Keeping it consistence is a good reading experience for users.

ptabor (approved):

Issue #14477, opened 2022-09-16
[Test] TestMaintenanceSnapshotCancel failed 


ahrtr:
### Which github workflows are flaking?

test (linux-amd64-grpcproxy-integration)

### Which tests are flaking?

TestMaintenanceSnapshotCancel

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/3064755697/jobs/4948171914

### Reason for failure (if possible)

```
--- FAIL: TestMaintenanceSnapshotCancel (0.05s)
    cluster.go:533: Creating listener with addr: 127.0.0.1:2125413976
    cluster.go:533: Creating listener with addr: 127.0.0.1:2125513976
    logger.go:130: 2022-09-16T01:20:48.006Z	INFO	m0	LISTEN GRPC	{"member": "m0", "grpcAddr": "localhost:m0", "m.Name": "m0", "workdir": "/tmp/TestMaintenanceSnapshotCancel3954815318/001"}
    cluster.go:217: Launching new cluster...
    logger.go:130: 2022-09-16T01:20:48.007Z	INFO	m0	launching a member	{"member": "m0", "name": "m0", "advertise-peer-urls": ["unix://127.0.0.1:2125413976"], "listen-client-urls": ["unix://127.0.0.1:2125513976"], "grpc-url": "unix://localhost:m0"}
    logger.go:130: 2022-09-16T01:20:48.008Z	INFO	m0	opened backend db	{"member": "m0", "path": "/tmp/TestMaintenanceSnapshotCancel3954815318/002/etcd3345957309/member/snap/db", "took": "1.147304ms"}
    logger.go:130: 2022-09-16T01:20:48.011Z	INFO	m0	starting local member	{"member": "m0", "local-member-id": "17189ef9c980bc18", "cluster-id": "b21ab012ec882f7c"}
    logger.go:130: 2022-09-16T01:20:48.011Z	INFO	m0.raft	17189ef9c980bc18 switched to configuration voters=()	{"member": "m0"}
    logger.go:130: 2022-09-16T01:20:48.011Z	INFO	m0.raft	17189ef9c980bc18 became follower at term 0	{"member": "m0"}
    logger.go:130: 2022-09-16T01:20:48.011Z	INFO	m0.raft	newRaft 17189ef9c980bc18 [peers: [], term: 0, commit: 0, applied: 0, lastindex: 0, lastterm: 0]	{"member": "m0"}
--
              "Type": 0,
              "Metadata": null
            }
[...]

PR #14552, opened 2022-10-05
integration: deflaking TestMaintenanceSnapshotCancel/Timeout

fuweid:
Since http2 spec defines the receive windows's size and max size of frame in the stream, the underlayer - gRPC client can pre-read data from server even if the application layer hasn't read it yet.

And the initialized cluster has 20KiB snapshot, which can be pre-read by underlayer. We should increase the snapshot's size, just in case that the io.Copy won't return the canceled or timeout error.

Fixes: #14477


tests/integration/clientv3/maintenance_test.go
```diff
@@ -156,6 +157,24 @@ func TestMaintenanceSnapshotCancel(t *testing.T) {
 
 	// reading snapshot with canceled context should error out
 	ctx, cancel := context.WithCancel(context.Background())
+
+	// Since http2 spec defines the receive windows's size and max size of
```
ptabor, line 161: Nice. Thank you. 

Could you, please, move it to a separate helper method as it shared piece of logic, e.g.

func populateWithData(t testing.T, cluster *Cluster, numKeys int, valueSize int) {
...
}

ptabor (changes requested):

tests/integration/clientv3/maintenance_test.go
ahrtr: +1

fuweid: Sure! Updated. PTAL Thanks!

serathius (approved):

serathius:
Really awesome find!

fuweid:
Thanks!

PR #14555, opened 2022-10-07
raft: panic on self-addressed messages

nvanbenschoten:
These are nonsensical and a network implementation is not required to handle them correctly, so panic instead of sending them out.

cc. @tbg 

### DIFF
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -415,6 +415,9 @@ func (r *raft) send(m pb.Message) {
 			m.Term = r.Term
 		}
 	}
+	if m.To == r.id {
+		r.logger.Panicf("message should not be self-addressed when sending %s", m.Type)
+	}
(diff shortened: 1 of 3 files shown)


ahrtr (approved):
LGTM

Thank you @nvanbenschoten 

A quick question, do you see any real issue or you are just reviewing the source code? 

cc @tbg and @ptabor to double check.

nvanbenschoten:
Hi @ahrtr, thanks for the review!

I did not see any real issues that motivated this PR, it just seemed like a useful assertion to add because network implementations are not required to handle self-addressed messages correctly.

I came across this and a collection of other cleanups while working on a change that will address https://github.com/etcd-io/etcd/issues/12257.

tbg (approved):

PR #14556, opened 2022-10-07
client/v3: Add backoff before retry when watch stream returns unavailable

tomari:
The client retries connection without backoff when the server is gone after the watch stream is established. This results in high CPU usage in the client process. This change introduces backoff when the stream is failed and unavailable.

Signed-off-by: Hisanobu Tomari <posco.grubb@gmail.com>


client/v3/watch.go
```diff
@@ -676,6 +677,17 @@ func (w *watchGrpcStream) run() {
...
+				// retry, but backoff
+				if backoff < maxBackoff {
+					// 25% backoff factor
+					backoff = backoff + backoff/4
+					if backoff > maxBackoff {
+						backoff = maxBackoff
+					}
+				}
+				time.Sleep(backoff)
+			}
```
ahrtr, line 690: The code is duplicated to https://github.com/etcd-io/etcd/blob/6a0bbf346256960cbbe0218d6ab13443ee93e8e3/client/v3/watch.go#L1018-L1028 

Please get it included in a function so that it can be reused by both places.

tomari: Thanks for the review. The backoff function has been added.

ahrtr (approved):
LGTM

Thank you @tomari

spzala (approved):
lgtm. Thanks @tomari

ahrtr:
@tomari Could you backport the fix to 3.5 and 3.4?

PR #13865, opened 2022-04-01
Add no_proxy environment variable for localhost,127.0.0.1

ahrtr:
The e2e test `TestClusterVersion` always fails, the reason is the OS environment variables, including`http_proxy` and `https_proxy` if present, are automatically added into the command's environment variable list, see [expect.go#L58](https://github.com/etcd-io/etcd/blob/main/pkg/expect/expect.go#L58) .

Accordingly all http requests to localhost fail, the error message is something like below,
```
 (expected "\"etcdcluster\":\"3.6", got ["<!DOCTYPE html PUBLIC \"-//W3C//DTD HTML 4.01//EN\" \"http://www.w3.org/TR/html4/strict.dtd\">\r\n" "<html><head>\r\n" "<meta type=\"copyright\" content=\"Copyright (C) 1996-2016 The Squid Software Foundation and contributors\">\r\n" "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\">\r\n"

```

This is one of the reasons for the CI failures. cc @ptabor @serathius @spzala

PR #14560, opened 2022-10-07
tests: Pass only etcd related environment variables during e2e tests

serathius:
Should really clean up e2e test logs and make tests much more repeatable.
cc @ahrtr @spzala

### DIFF
--- a/tests/framework/e2e/util.go
+++ b/tests/framework/e2e/util.go
@@ -127,6 +127,10 @@ func mergeEnvVariables(envVars map[string]string) []string {
 	currVars := os.Environ()
 	for _, v := range currVars {
 		p := strings.Split(v, "=")
+		// TODO: Remove PATH when we stop using system binaries (`awk`, `echo`)
+		if !strings.HasPrefix(p[0], "ETCD_") && !strings.HasPrefix(p[0], "ETCDCTL_") && !strings.HasPrefix(p[0], "EXPECT_") && p[0] != "PATH" {
+			continue
+		}
 		if _, ok := envVars[p[0]]; !ok {
 			env = append(env, fmt.Sprintf("%s=%s", p[0], p[1]))
 		}


ahrtr:
Did you see any issue which is caused by the code you changed in this PR?

Points from my side:
1. The command `echo`, `awk` and `env` are very common, but they may locate on different directories on different platforms. For example, the `echo` is in `/bin/echo` instead of `/usr/bin/echo` on Mac. Although darwin in on tier 3 support, but it's very widely used. So I suggest to keep it as it's for now so as not to break anything. 
2. E2E test is just simulating users' real end to end scenario, and any process (including etcd) must be affected by the environment in which it's running. So it makes sense to merge all available environment variables instead of just etcd related environment variables.

serathius:
Ad 1, I think it's reasonable to require exact locations of binaries. On the other hand, I don't think e2e tests should depend on external binaries too much. 
Ad 2, I don't agree that e2e tests are to simulate user end scenario in local environment. I have never heard about a case where we recommended to user to verify their environment by running e2e tests. They are meant to simulate a generic environment with minimal assumptions to make it easy to reproduce issues. To achieve that we need to minimize chances that local environment will impact the e2e tests. 

We should prefer allowing to parameterize e2e environment instead of blind pass-through. If there is issue with particular environment we should extend existing test to also cover it.

ahrtr:
There is a similar discussion previously https://github.com/etcd-io/etcd/pull/13865

ahrtr:
If you ignore all other “unrelated” environment variables, then it means that you disable some features supported by golang and probably some other dependencies, such as `GOMEMLIMIT`.  Although we do not use the features in e2e test, but everyone can do it out of the box by setting environment variables. 

I understand that your point is to make the test reproducible. If you insist on passing only etcd related environment variables, then you'd better make it configurable. We should can control whether to pass all or onlly etcd related environment variables. 

What's your opinion? @ptabor @spzala

serathius:
> I understand that your point is to make the test reproducible. If you insist on passing only etcd related environment variables, then you'd better make it configurable. We should can control whether to pass all or onlly etcd related environment variables.

This was the direction I'm proposing, however I don't think we need to imminently make it configurable. If we need we can just change the `mergeEnvVariables` function to pick a environment that defines a set of variables.

serathius:
ping @ahrtr

ahrtr:
Please consider defining a whitelist so that we don't explode the `if` statement. 

> Did you see any issue which is caused by the code you changed in this PR?

Could you respond to this question?  Previously I saw an issue caused by http_proxy (FYI https://github.com/etcd-io/etcd/pull/13865), did you ever see any other issue which was caused by what ever local environment variable?

serathius:
> Please consider defining a whitelist so that we don't explode the if statement.

I don't expect to grow the list of prefixes. We might have a prefix for each etcd binary and for expect package.

> Did you see any issue which is caused by the code you changed in this PR?

No, I didn't see issues here, however every time I run e2e tests I'm terrified by number of environment variables on my machine that could introduce noise into testing. They also could not be ever pasted input any issue without heavy obfuscating. 

Every time I analyse e2e logs I'm always hit by how unreadable they are. I would like to start cleanup from making it clear what commands are run. There was a recent issue of e2e tests passing same flag 3 times and we never noticed because it's impossible to know from logs what is really being run.

ahrtr (approved):

PR #14561, opened 2022-10-07
Ensure that input validation between API and Apply is consistent

### LINKED ISSUE #13617: Ensure that input validation between API and Apply is consistent
Recent fixes for bugs detected by fuzzing has resulted in improved validation for Apply, for example https://github.com/etcd-io/etcd/pull/13602. Errors during Apply result in panic as raft is unable to progress. Fuzz tests detect those panics within apply function, however this not the only place that fix should be fixed.

Detecting invalid raft message during apply is very late stage. Such messages should be rejected on much earlier stage. Recent fixes has showed that we have no good mechanism to ensure consistency validating messages. 

Goal: Ensure that API rejects all requests that would be rejected during apply or caused panic.

Ideas for implementation:
* We could write a fuzzer test that will generate a lot of randomly filled structs representing API rpc. With rpcs we can push them through both validators and check if they result any inconsistencies.

cc @ptabor

samueleresca:
Fixes #13617 
The fuzz input is passed to the input validation. In case the input validation passes - the application of that request shouldn't panic.

Adding a fuzzing script and a GitHub action for executing fuzzing with a 5 min time limit

### DIFF
--- a/.github/workflows/fuzzing.yaml
+++ b/.github/workflows/fuzzing.yaml
@@ -1,0 +1,22 @@
+name: Fuzzing v3rpc
+on: [push, pull_request]
+jobs:
+  fuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+          Target: [FuzzRangeRequest, FuzzPutRequest, FuzzDeleteRangeRequest]
+    env:
+      FUZZ_TARGET: ${{matrix.Target}}
+      TARGET_PATH: ./server/etcdserver/api/v3rpc
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: "1.19.1"
+    - run: GOARCH=amd64 CPU=4 make fuzz
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        path: "${{env.TARGET_PATH}}/testdata/fuzz/${{env.FUZZ_TARGET}}/*"
--- a/.github/workflows/fuzzing.yaml
+++ b/.github/workflows/fuzzing.yaml
@@ -5,10 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
-          Target: [FuzzRangeRequest, FuzzPutRequest, FuzzDeleteRangeRequest]
     env:
-      FUZZ_TARGET: ${{matrix.Target}}
       TARGET_PATH: ./server/etcdserver/api/v3rpc
     steps:
     - uses: actions/checkout@v2
@@ -19,4 +16,4 @@ jobs:
     - uses: actions/upload-artifact@v2
       if: failure()
       with:
-        path: "${{env.TARGET_PATH}}/testdata/fuzz/${{env.FUZZ_TARGET}}/*"
+        path: "${{env.TARGET_PATH}}/testdata/fuzz/**/*"
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,10 @@ test-e2e-release: build
 test-linearizability: build
 	PASSES="linearizability" ./scripts/test.sh $(GO_TEST_FLAGS)
 
+.PHONY: fuzz
+fuzz: 
+	./scripts/fuzzing.sh
+
 # Static analysis
 
 verify: verify-gofmt verify-bom verify-lint verify-dep verify-shellcheck verify-goword verify-govet verify-license-header verify-receiver-name verify-mod-tidy verify-shellcheck verify-shellws verify-proto-annotations
--- a/scripts/fuzzing.sh
+++ b/scripts/fuzzing.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+set -e
+source ./scripts/test_lib.sh
+
+GO_CMD="go"
+fuzz_time=${FUZZ_TIME:-"300s"}
+target_path=${TARGET_PATH:-"./server/etcdserver/api/v3rpc"}
+fuzz_target=${FUZZ_TARGET:-"FuzzRangeRequest"}
+
+log_callout -e "\\nExecuting fuzzing with target ${fuzz_target} in $target_path with a timeout of $fuzz_time\\n"
+cd "$target_path"
+$GO_CMD test -fuzz "$fuzz_target" -fuzztime "$fuzz_time"
+log_success -e "\\COMPLETED: fuzzing with target $fuzz_target in $target_path \\n"
--- a/scripts/fuzzing.sh
+++ b/scripts/fuzzing.sh
@@ -5,9 +5,14 @@ source ./scripts/test_lib.sh
 GO_CMD="go"
 fuzz_time=${FUZZ_TIME:-"300s"}
 target_path=${TARGET_PATH:-"./server/etcdserver/api/v3rpc"}
-fuzz_target=${FUZZ_TARGET:-"FuzzRangeRequest"}
+TARGETS="FuzzTxnRangeRequest  FuzzTxnPutRequest  FuzzTxnDeleteRangeRequest"
 
-log_callout -e "\\nExecuting fuzzing with target ${fuzz_target} in $target_path with a timeout of $fuzz_time\\n"
-cd "$target_path"
-$GO_CMD test -fuzz "$fuzz_target" -fuzztime "$fuzz_time"
-log_success -e "\\COMPLETED: fuzzing with target $fuzz_target in $target_path \\n"
+
+for target in ${TARGETS}; do
+    log_callout -e "\\nExecuting fuzzing with target ${target} in $target_path with a timeout of $fuzz_time\\n"
+    run pushd ${target_path}
+        $GO_CMD test -fuzz ${target} -fuzztime ${fuzz_time}
+    run popd
+    log_success -e "\\COMPLETED: fuzzing with target $target in $target_path \\n"
+done
+
--- a/scripts/fuzzing.sh
+++ b/scripts/fuzzing.sh
@@ -10,8 +10,8 @@ TARGETS="FuzzTxnRangeRequest  FuzzTxnPutRequest  FuzzTxnDeleteRangeRequest"
 
 for target in ${TARGETS}; do
     log_callout -e "\\nExecuting fuzzing with target ${target} in $target_path with a timeout of $fuzz_time\\n"
-    run pushd ${target_path}
-        $GO_CMD test -fuzz ${target} -fuzztime ${fuzz_time}
+    run pushd "${target_path}"
+        $GO_CMD test -fuzz "${target}" -fuzztime "${fuzz_time}"
     run popd
     log_success -e "\\COMPLETED: fuzzing with target $target in $target_path \\n"
 done
--- a/server/etcdserver/api/v3rpc/validationfuzz_test.go
+++ b/server/etcdserver/api/v3rpc/validationfuzz_test.go
@@ -12,7 +12,7 @@ import (
 	"go.uber.org/zap/zaptest"
 )
 
-func FuzzRangeRequest(f *testing.F) {
+func FuzzTxnRangeRequest(f *testing.F) {
 	testcases := []pb.RangeRequest{
 		{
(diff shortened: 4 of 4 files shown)


server/etcdserver/api/v3rpc/validationfuzz_test.go
```diff
@@ -0,0 +1,82 @@
...
+			Limit:      limit,
+			SortOrder:  pb.RangeRequest_SortOrder(sortOrder),
+			SortTarget: pb.RangeRequest_SortTarget(sortTarget),
+		})
+
+		if errCheck != nil {
+			t.Skip("Validation not passing. Skipping the apply.")
+		}
+
+		_, _, err := txn.Txn(ctx, zaptest.NewLogger(t), request, false, s, &lease.FakeLessor{})
```
serathius, line 79: Should we check that `err is nil`?

serathius:
Looks great, would be good setup CI to run fuzz test. I think we should run:
* Short 5-10m run o every pull request to verify that fuzz works
* Long 1-2h once a day to do a proper fuzzing

What do you think? Do you need help with that? We doing the first one should be enough for now.

server/etcdserver/api/v3rpc/validationfuzz_test.go
samueleresca: Now it should be fine

samueleresca:
> I think we should run:
> 
> * Short 5-10m run o every pull request to verify that fuzz works
> * Long 1-2h once a day to do a proper fuzzing
> 
> What do you think? Do you need help with that? We doing the first one should be enough for now.

@serathius I have added the GitHub Action for fuzzing:
- Tested that on my fork repo in case of success: https://github.com/samueleresca/etcd/actions/runs/3215190941
- In case of failure: https://github.com/samueleresca/etcd/actions/runs/3215106275 (producing artifacts)

I'm proceeding by covering the others input validation methods.
 
Any other feedback is obvs welcome

scripts/fuzzing.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# To fix according to newer version of go:
+# go get golang.org/dl/gotip
+# gotip download
+# GO_CMD="gotip"
+GO_CMD="go"
+
+log_callout -e "\\nExecuting v3rpc with target $FUZZ_TARGET \\n"
+
+cd $BASE_PATH
```
serathius, line 14: This script requires providing multiple non-trivial environment variables for it to work. This script should only be called via github action, but also locally to reproduce issues. I think we should make the script independent so it can be called without to much hardship. 

Can we have the script include a good defaults for running? Or if the default cannot be provided, could we add validation to give better error message?

.github/workflows/fuzzing.yaml
```diff
@@ -0,0 +1,24 @@
...
+    env:
+      BASE_PATH: server/etcdserver/api/v3rpc
+      FUZZ_TIME: 300s
+      FUZZ_TARGET: ${{matrix.Target}}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: "1.19.1"
+    - run: date
```
serathius, line 19:
```suggestion
```
Not needed

.github/workflows/fuzzing.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - run: GOARCH=amd64 CPU=4 ./scripts/fuzzing.sh
```
serathius, line 20: Could you add a fuzz command to Makefile?

samueleresca: Updated Makefile with the `fuzz` command

scripts/fuzzing.sh
samueleresca: I updated the script:

- FUZZ_TIME now defaults to 300s
- TARGET_PATH display an error in the case that is unset
- FUZZ_TARGET display an error in the case that is unset

Let me know if you prefer another setup

.github/workflows/fuzzing.yaml
serathius: Can we call the makefile from here?

scripts/fuzzing.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -e
+source ./scripts/test_lib.sh
+
+if [[ -z "${TARGET_PATH}" ]]; then
+  log_error -e "\\nERROR: specify the TARGET_PATH environment variable. The TARGET_PATH is the path where the fuzz test is defined."
+  exit 255
+else
+  TARGET_PATH="${TARGET_PATH}"
+fi
```
serathius, line 11:
```suggestion
: ${TARGET_PATH:?ERROR: specify the TARGET_PATH environment variable. The TARGET_PATH is the path where the fuzz test is defined.}
```

scripts/fuzzing.sh
```diff
@@ -0,0 +1,33 @@
...
+  FUZZ_TARGET="${FUZZ_TARGET}"
+fi
+
+fuzz_time=${FUZZ_TIME:-"300s"}
+
+# To fix according to newer version of go:
+# go get golang.org/dl/gotip
+# gotip download
+# GO_CMD="gotip"
+GO_CMD="go"
```
serathius, line 26: We no longer use gotip.

Makefile
```diff
@@ -27,6 +27,10 @@ test-e2e: build
 test-e2e-release: build
 	PASSES="release e2e" ./scripts/test.sh
 
+.PHONY: fuzz
+fuzz:
+	PASSES="fuzz" ./scripts/fuzzing.sh
```
serathius, line 32: `PASSES="fuzz"` is not needed here.

serathius, line 32: Could `make fuzz` be self sufficient to call without additional arguments (TARGET_PATH, FUZZ_TARGET)?

serathius:
Note, you could clean up the commit history? Looks like you have been pulling other commits as PR has multiple borrowed commits from other PRs. I think it would be best if we squash everything into one commit. Please let me know if you need any help.

ahrtr:
cc @AdamKorcz  I recall that you added some fuzz test a couple of months back. Could you take a look at this PR as well? Thanks.

AdamKorcz:
@ahrtr Thanks for the tag.

@samueleresca:

1.  Do these fuzz tests replace any of the ones running on OSS-Fuzz? https://github.com/cncf/cncf-fuzzing/tree/main/projects/etcd
2. Could you also add these to the build script for OSS-Fuzz? https://github.com/cncf/cncf-fuzzing/blob/main/projects/etcd/build.sh

serathius:
Hey @AdamKorcz, this is independent fuzzing to address one of the bugs in the api we found.

Unfortunately the OSS-Fuzz was developed outside of the etcd repo, as so they:
* Were not reviewed by maintainers
* Are apparently broken https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=51781
* We cannot reliably maintain them as they are not integrated in the CI and only maintainers have access to their reports. 
* Maintainers do not have capacity to maintain external project

As so I would recommend that we migrate OSS-Fuzz code into the etcd repo. We can use this PR as the bases for the test. This should allow us to at test that the fuzz test build during CI.

server/etcdserver/api/v3rpc/validationfuzz_test.go
```diff
@@ -0,0 +1,219 @@
...
+		if err != nil {
+			t.Logf("Check: %s | Apply: %s", errCheck, err)
+			t.Skip("Application erroring.")
+		}
+	})
+}
+
+func FuzzDeleteRangeRequest(f *testing.F) {
```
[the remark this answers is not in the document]
ahrtr: Firstly, the three cases are all testing the same api `Txn`, so please rename per comments above.

Secondly, the code logic of the three cases are very similar, please consider to refactor the implementation, such as create a common function to be reused by the cases.

.github/workflows/fuzzing.yaml
```diff
@@ -0,0 +1,22 @@
+name: Fuzzing v3rpc
+on: [push, pull_request]
+jobs:
+  fuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+          Target: [FuzzRangeRequest, FuzzPutRequest, FuzzDeleteRangeRequest]
```
ahrtr, line 9: Each time when any contributor add a new fuzz case, then they need to update this file?

server/etcdserver/api/v3rpc/validationfuzz_test.go
```diff
@@ -0,0 +1,219 @@
...
+			SortTarget: pb.RangeRequest_SortTarget(sortTarget),
+		})
+
+		if errCheck != nil {
+			t.Skip("Validation not passing. Skipping the apply.")
+		}
+
+		_, _, err := txn.Txn(ctx, zaptest.NewLogger(t), request, false, s, &lease.FakeLessor{})
+		if err != nil {
+			t.Logf("Check: %s | Apply: %s", errCheck, err)
```
ahrtr, line 81: Shouldn't we fail the test?

ahrtr:
Note: Golang supports fuzzing starting from 1.18, 
https://go.dev/security/fuzz/

server/etcdserver/api/v3rpc/validationfuzz_test.go
samueleresca: The goal of the test is to verify that the `Tnx` function won't panic in case the validation check passes.  it is okay in this context if the `Tnx` handles the request with an error. In case of panic, the test should fail.

samueleresca:
> Note: Golang supports fuzzing starting from 1.18, https://go.dev/security/fuzz/

I'm already using the out-of-the-box fuzzing coming from golang

.github/workflows/fuzzing.yaml
samueleresca: Yes. Unfortunately out-of-box fuzzing requires a specific target method to start the fuzzing. In case you try to use something like `go test -fuzz Fuzz*` the returned error is:

```
testing: will not fuzz, -fuzz matches more than one fuzz test: [FuzzRangeRequest FuzzPutRequest FuzzDeleteRangeRequest]
```
Let me know if there is a better way of doing that

ahrtr: I would suggest to get all targets defined in the `scripts/fuzzing.sh ` for now, so that we can manually easily run all fuzz tests locally.

We can update the script/fuzzing.sh something like below,

```
TARGETS="FuzzRangeRequest  FuzzPutRequest  FuzzDeleteRangeRequest"
for target in ${TARGETS}; do
    run pushd ${target_path}
        go test -fuzz ${target}
    run popd
done
```

server/etcdserver/api/v3rpc/validationfuzz_test.go
ahrtr: OK.

ahrtr:
I see lots of unrelated commits in this PR, please rebase this PR.

.github/workflows/fuzzing.yaml
samueleresca: I proceeded by updating the script as requested. Note that, the fuzzing for the different targets won't run in parallel. The fuzz process will start from `FuzzRangeRequest` and finishing with `FuzzDeleteRangeRequest` with the specified timeout

scripts/fuzzing.sh
```diff
@@ -0,0 +1,18 @@
...
+TARGETS="FuzzTxnRangeRequest  FuzzTxnPutRequest  FuzzTxnDeleteRangeRequest"
+
+
+for target in ${TARGETS}; do
+    log_callout -e "\\nExecuting fuzzing with target ${target} in $target_path with a timeout of $fuzz_time\\n"
+    run pushd "${target_path}"
+        $GO_CMD test -fuzz "${target}" -fuzztime "${fuzz_time}"
+    run popd
+    log_success -e "\\COMPLETED: fuzzing with target $target in $target_path \\n"
+done
```
ahrtr, line 17: Minor comment: It seems that we only need to execute `pushd` and `popd` once outside the for loop.

Of course, it isn't a big deal, feel free to fix it in a separate PR.

ahrtr (approved):
LGTM

Thank you @samueleresca

PR #14566, opened 2022-10-09
raft:Optimized sorting alogorithm

lastdz:
Optimize insertion sort to quick sort
Time complexity is optimized from O(N^2) to O(N*log(N))

### DIFF
--- a/raft/quorum/majority.go
+++ b/raft/quorum/majority.go
@@ -112,15 +112,6 @@ func (c MajorityConfig) Slice() []uint64 {
 	return sl
 }
 
-func insertionSort(sl []uint64) {
-	a, b := 0, len(sl)
-	for i := a + 1; i < b; i++ {
-		for j := i; j > a && sl[j] < sl[j-1]; j-- {
-			sl[j], sl[j-1] = sl[j-1], sl[j]
-		}
-	}
-}
-
 // CommittedIndex computes the committed index from those supplied via the
 // provided AckedIndexer (for the active config).
 func (c MajorityConfig) CommittedIndex(l AckedIndexer) Index {
@@ -160,9 +151,10 @@ func (c MajorityConfig) CommittedIndex(l AckedIndexer) Index {
 		}
 	}
 
-	// Sort by index. Use a bespoke algorithm (copied from the stdlib's sort
-	// package) to keep srt on the stack.
-	insertionSort(srt)
+	// Sort by index.
+	sort.Slice(srt, func(i, j int) bool {
+		return srt[i] < srt[j]
+	})
 
 	// The smallest index into the array for which the value is acked by a
 	// quorum. In other words, from the end of the slice, move n/2+1 to the


ahrtr:
Please signoff the commit: `git rebase HEAD~1 --signoff`

lastdz:
> Please signoff the commit: `git rebase HEAD~1 --signoff`

ok

ahrtr (approved):
LGTM

Thank you @lastdz

serathius (changes requested):
Please don't assume that theoretical complexity always means faster. Big O complexity gives you asymptotic idea how algorithm behaves for big input. At smaller inputs it doesn't account for constant time at all. This is sorting list of members range is between 3-5. I would not be surprised if insert sort would be as fast if not faster than merge sort. 

In addition the comment you removed author mentions directly that they picked insert sort to keep the sorting on stack, and not heap. Based on that I would be against making changes.

Please provide direct benchmarks that would show performance improvement through 
micro benchmarks using real data (for example taken from running etcd cluster).

ahrtr:
Thanks @serathius for the feedback, which basically makes sense to me. 

Insertion sort seems indeed better than quicksort on small size of slice. Golang is also[ using insertion sort](https://github.com/golang/go/blob/4a4127bccc826ebb6079af3252bc6bfeaec187c4/src/sort/zsortinterface.go#L72-L75) on small size (<=12). 

[sort.Sort](https://github.com/golang/go/blob/go1.19.1/src/sort/sort.go#L42) guarantees that the slice data is always on where it's, but [sort.Slice](https://github.com/golang/go/blob/go1.19.1/src/sort/slice.go#L18) may escape the data to heap, although golang may have plan to improve this (allow contents of a Value to live on the stack).  Anyway we shouldn't use `sort.Slice` for now. 

The only reason for approving this PR is for better readability, but this PR should be updated as below. Note: golang has `IntSlice` and `sort.Ints`, but unfortunately doesn't have `Int64Slice/UInt64Slice` and `sort.Int64s/sort.UInt64s `. 

 cc  @ptabor @spzala as well. 


```
$ git diff
diff --git a/raft/quorum/majority.go b/raft/quorum/majority.go
index 12766137a..bebbdbea5 100644
--- a/raft/quorum/majority.go
+++ b/raft/quorum/majority.go
@@ -112,14 +112,12 @@ func (c MajorityConfig) Slice() []uint64 {
        return sl
 }
 
-func insertionSort(sl []uint64) {
-       a, b := 0, len(sl)
-       for i := a + 1; i < b; i++ {
-               for j := i; j > a && sl[j] < sl[j-1]; j-- {
-                       sl[j], sl[j-1] = sl[j-1], sl[j]
-               }
-       }
-}
+// UInt64Slice implements the sort.Interface.
+type UInt64Slice []uint64
+
+func (x UInt64Slice) Len() int           { return len(x) }
+func (x UInt64Slice) Less(i, j int) bool { return x[i] < x[j] }
+func (x UInt64Slice) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
 
 // CommittedIndex computes the committed index from those supplied via the
 // provided AckedIndexer (for the active config).
@@ -162,7 +160,7 @@ func (c MajorityConfig) CommittedIndex(l AckedIndexer) Index {
 
        // Sort by index. Use a bespoke algorithm (copied from the stdlib's sort
        // package) to keep srt on the stack.
-       insertionSort(srt)
+       sort.Sort(UInt64Slice(srt))
 
        // The smallest index into the array for which the value is acked by a
        // quorum. In other words, from the end of the slice, move n/2+1 to the

```

lastdz:
> Thanks @serathius for the feedback, which basically makes sense to me.
> 
> Insertion sort seems indeed better than quicksort on small size of slice. Golang is also[ using insertion sort](https://github.com/golang/go/blob/4a4127bccc826ebb6079af3252bc6bfeaec187c4/src/sort/zsortinterface.go#L72-L75) on small size (<=12).
> 
> [sort.Sort](https://github.com/golang/go/blob/go1.19.1/src/sort/sort.go#L42) guarantees that the slice data is always on where it's, but [sort.Slice](https://github.com/golang/go/blob/go1.19.1/src/sort/slice.go#L18) may escape the data to heap, although golang may have plan to improve this (allow contents of a Value to live on the stack). Anyway we shouldn't use `sort.Slice` for now.
> 
> The only reason for approving this PR is for better readability, but this PR should be updated as below. Note: golang has `IntSlice` and `sort.Ints`, but unfortunately doesn't have `Int64Slice/UInt64Slice` and `sort.Int64s/sort.UInt64s `.
> 
> cc @ptabor @spzala as well.
> 
> ```
> $ git diff
> diff --git a/raft/quorum/majority.go b/raft/quorum/majority.go
> index 12766137a..bebbdbea5 100644
> --- a/raft/quorum/majority.go
> +++ b/raft/quorum/majority.go
> @@ -112,14 +112,12 @@ func (c MajorityConfig) Slice() []uint64 {
>         return sl
>  }
>  
> -func insertionSort(sl []uint64) {
> -       a, b := 0, len(sl)
> -       for i := a + 1; i < b; i++ {
> -               for j := i; j > a && sl[j] < sl[j-1]; j-- {
> -                       sl[j], sl[j-1] = sl[j-1], sl[j]
> -               }
> -       }
> -}
> +// UInt64Slice implements the sort.Interface.
> +type UInt64Slice []uint64
> +
> +func (x UInt64Slice) Len() int           { return len(x) }
> +func (x UInt64Slice) Less(i, j int) bool { return x[i] < x[j] }
> +func (x UInt64Slice) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
>  
>  // CommittedIndex computes the committed index from those supplied via the
>  // provided AckedIndexer (for the active config).
> @@ -162,7 +160,7 @@ func (c MajorityConfig) CommittedIndex(l AckedIndexer) Index {
>  
>         // Sort by index. Use a bespoke algorithm (copied from the stdlib's sort
>         // package) to keep srt on the stack.
> -       insertionSort(srt)
> +       sort.Sort(UInt64Slice(srt))
>  
>         // The smallest index into the array for which the value is acked by a
>         // quorum. In other words, from the end of the slice, move n/2+1 to the
> ```

I make this pr because it was used `sort.Slice` before,I think it should be regulated

		func (c MajorityConfig) Slice() []uint64 {
			var sl []uint64
			for id := range c {
				sl = append(sl, id)
			}
			sort.Slice(sl, func(i, j int) bool { return sl[i] < sl[j] })
			return sl
		}

ahrtr:
Please note that `(c MajorityConfig) Slice()` isn't frequently called, while `(c MajorityConfig) CommittedIndex` gets called frequently, because leader might call it each time receiving a `MsgAppResp` from any followers. So even slight performance reduction (note  `sort.Slice` uses reflection as well) might cause non-trivial overall performance reduction. 

So you need do benchmark test on both the unit test level and system level (using the [benchmark](https://github.com/etcd-io/etcd/tree/main/tools/benchmark) tool) to compare the performance results.

PR #14524, opened 2022-09-26
Update 3.4, 3.5, and 3.6 changelogs for adding configurable cipher list to gRPC proxy

dusk125:
Since I'm backporting #14308 from 3.6 to 3.5, I need to move the changelog line to 3.5

PR #14570, opened 2022-10-10
Update changelog readme to add rules on how to add changelog items

ahrtr:
Please refer to the following discussion:
1. https://github.com/etcd-io/etcd/pull/14476#issuecomment-1249192558 
2. https://github.com/etcd-io/etcd/pull/14524#issuecomment-1267816042 

cc @spzala @serathius @ptabor @mitake

### DIFF
--- a/CHANGELOG/README.md
+++ b/CHANGELOG/README.md
@@ -1,4 +1,6 @@
-# Production recommendation
+# Change logs
+
+## Production recommendation
 
 The minimum recommended etcd versions to run in **production** are v3.4.8+ and v3.5.4+. Refer to the [versioning policy](https://etcd.io/docs/v3.5/op-guide/versioning/) for more details.
 
@@ -9,3 +11,11 @@ If etcd process is killed, occasionally some committed transactions are not refl
 Recommendation is to upgrade to v3.5.4+.
 
 If you have encountered data corruption, please follow instructions on https://etcd.io/docs/v3.5/op-guide/data_corruption/.
+
+## Change log rules
+1. Each patch release only includes changes against previous patch release.
+For example, the change log of v3.5.5 should only include items which are new to v3.5.4.
+2. For the first release (e.g. 3.4.0, 3.5.0, 3.6.0, 4.0.0 etc.) for each minor or major 
+version, it only includes changes which are new to the first release of previous minor
+or major version. For example, v3.5.0 should only include items which are new to v3.4.0,
+and v3.6.0 should only include items which are new to v3.5.0.


CHANGELOG/README.md
```diff
@@ -9,3 +11,11 @@ If etcd process is killed, occasionally some committed transactions are not refl
 Recommendation is to upgrade to v3.5.4+.
 
 If you have encountered data corruption, please follow instructions on https://etcd.io/docs/v3.5/op-guide/data_corruption/.
+
+## Change log rules
```
spzala, line 15: @ahrtr this is a good start but couple of comments:
- You haven't covered a case when a major release and previous minor supported release? e.g. 4.0 and 3.x.x 
- One thing that I worry is that not every user will upgrade to newest patch release. e.g. users moving to 3.6 from 3.4.x or 3.5.x (when few more patch releases were available after 3.5.x)? To see what's changed they need to go find the patch releases of 3.5? 
- Also, users may miss some of the new changes that's going in main branch because main branch changelog was not updated as we backported those changes that also went into the main but we only updated changelog of backported release branch right? 
Thanks!

ahrtr: Thanks @spzala for the quick response.

1. Yes, it's a good catch, just updated. 
2. Yes, users need to read all the changelogs between current version and the target version to which they plan to upgrade. We can's take care of all such scenarios. 
3. This isn't a problem. Let's work with three examples. 
    -  Example 1: Assuming a user adds a new feature to `main`, and backports to both 3.5 and 3.4, then we shouldn't update 3.6 changelog item, because 3.6.0 isn't released yet. 
    - Example 2: Assuming 3.6.0 is already released, and `release-3.6` branch is cut. We need to follow the first rule `Each patch release only includes changes against previous patch release. ` to update the changelog-3.6.  If a user adds a new feature to main, and backport to 3.5 and  3.6, then both 3.5 and 3.6 changelogs need to be updated. 
    - Example 3: Assuming 3.6.0 is already released, but `release-3.6` branch isn't cut yet. I don't think we will run into this case. Each time when we release the first major or minor version, a related branch must be cut.

ahrtr: one more comment for # 3,  usually we don't backport features. But there is no hard limitation on this. I think the same thing to changelog item. 

If all maintainers agree, we can also add changelog item for both main and previous version for some important feature.For example, @serathius  added changelog item into both 3.6 and 3.5 for https://github.com/etcd-io/etcd/issues/14039

What do you think? @spzala

spzala: @ahrtr thank you so much for the detailed answers. As you noticed, some cases if maintainers agree we will update both the patch release and main branch changelog. That's good but then there may be some items that's added to main changelog and some not, I am not sure if we (specially non developer users) can easily track it. I overall feel that keeping main branch changelog (as we used to) is more user-friendly. There will be some redundancy but users can get changes at one place over previous minor release. Also, I think of couple more scenarios: 
- There may be changes that may be only backported and not added in newer version. How user differentiate by looking at the changelog of a patch release of what's also changed in the main branch? 
- What happens when, as an example, v3.5.x3 and v3.5.x4 had changes that were only backported, where as v3.5.x2 has some changes that were also changed in the main branch. How easily users notice it?

So far, I personally like the current way of updating main changelog as that gives one place to see what's changed over previous minor or major release (not keeping patch releases in account here). If we really want to change the current way, I am fine and see how it goes but also it would be good to get some more thoughts from you and others (including etcd users if we can). Thanks!

ahrtr: Thanks @spzala for the valuable comment, which makes sense to me.


> The current way of updating main changelog gives one place to see what's changed over previous minor or major release (not keeping patch releases in account here)

This seems correct and good to me. Specifically, the main changelog (`first release of each minor or major version`) should only include:
- new features. (Even a feature is backported, we also need to add it into main changelog).
- breaking changes
- deprecations
- Bumping of dependencies. Probably we only need to list the core dependencies bumping, such as Go, gRPC.etc. WDYT? 
- Other changes, which should be added per maintainers' discussion

One thing to be clearer, I don't think the main changelog shouldn't have bug fixes. Because a bug fix must be
- either introduced in previous version and accordingly need to be backporte 
- or introduced only in main branch, so no need to backport. In this case, we don't need to list either because it (e.g. 3.6.0) hasn't be released yet.

spzala (approved):
@ahrtr thank you for working on it. I am good to give it at try. cc @serathius - if you have any comments. Thanks!

CHANGELOG/README.md
ahrtr: FYI. I just updated the second rule to reflect the discussion. Basically we are still following the existing rules.

serathius (approved):

PR #14582, opened 2022-10-12
[3.5] client/v3: Add backoff before retry when watch stream returns unavailable

tomari:
Backport #14556 to 3.5.

The client retries connection without backoff when the server is gone after the watch stream is established. This results in high CPU usage in the client process. This change introduces backoff when the stream is failed and unavailable.

Signed-off-by: Hisanobu Tomari <posco.grubb@gmail.com>


ahrtr (approved):
LGTM

Thank you @tomari

Issue #14541, opened 2022-09-30
integration TestAuthority failed due to "bind: address already in use"


ahrtr:
### Which github workflows are flaking?

linux-amd64-integration-4-cpu

### Which tests are flaking?

TestAuthority

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/3153014723/jobs/5128969677

### Reason for failure (if possible)

=== FAIL: integration TestAuthority/Size:_3,_Scenario:_"http://address[:port]" (0.00s)
    cluster.go:535: Creating listener with addr: 127.0.0.1:2134105778
    cluster.go:535: Creating listener with addr: 127.0.0.1:2134205778
    logger.go:130: 2022-09-29T16:45:35.996Z	INFO	m0	LISTEN GRPC	{"member": "m0", "grpcAddr": "127.0.0.1:33400", "m.Name": "m0", "workdir": "/tmp/TestAuthoritySize_3,_Scenario_httpaddressport382019041/001"}
    cluster.go:535: Creating listener with addr: 127.0.0.1:2134405778
    cluster.go:535: Creating listener with addr: 127.0.0.1:2134505778
    logger.go:130: 2022-09-29T16:45:35.996Z	INFO	m1	LISTEN GRPC	{"member": "m1", "grpcAddr": "127.0.0.1:33431", "m.Name": "m1", "workdir": "/tmp/TestAuthoritySize_3,_Scenario_httpaddressport382019041/001"}
    cluster.go:535: Creating listener with addr: 127.0.0.1:2134705778
    cluster.go:535: Creating listener with addr: 127.0.0.1:2134805778
    logger.go:130: 2022-09-29T16:45:35.997Z	INFO	m2	LISTEN GRPC	{"member": "m2", "grpcAddr": "127.0.0.1:33462", "m.Name": "m2", "workdir": "/tmp/TestAuthoritySize_3,_Scenario_httpaddressport382019041/001"}
    cluster.go:731: listenGRPC FAILED: listen failed on grpc socket 127.0.0.1:33462 (listen tcp 127.0.0.1:33462: bind: address already in use)

xiaods:
retry run below cmd 
```
    linux-amd64-integration-4-cpu)
      GOARCH=amd64 CPU=4 PASSES='integration' RACE='false' ./scripts/test.sh
      ;;
```
get err output:
```bash
'integration' started at Thu Oct  6 08:16:08 CST 2022
[...]

PR #14585, opened 2022-10-13
Test port allocating using net.Listen

tjungblu:
To avoid port clashes during integration tests, we let the OS give us a free port.

A more general fix for #14541

### DIFF
--- a/tests/framework/integration/cluster.go
+++ b/tests/framework/integration/cluster.go
@@ -83,9 +83,8 @@ const (
 var (
 	ElectionTicks = 10
 
-	// LocalListenCount integration test uses unique ports, counting up, to listen for each
-	// member, ensuring restarted members can listen on the same port again.
-	LocalListenCount = int32(0)
+	// UniqueCount integration test is used to set unique member ids
+	UniqueCount = int32(0)
 
 	TestTLSInfo = transport.TLSInfo{
 		KeyFile:        MustAbsPath("../fixtures/server.key.insecure"),
@@ -258,6 +257,7 @@ func (c *Cluster) ProtoMembers() []*pb.Member {
 func (c *Cluster) mustNewMember(t testutil.TB) *Member {
 	memberNumber := c.LastMemberNum
 	c.LastMemberNum++
+
 	m := MustNewMember(t,
 		MemberConfig{
 			Name:                        fmt.Sprintf("m%v", memberNumber),
@@ -369,6 +369,7 @@ func (c *Cluster) RemoveMember(t testutil.TB, cc *clientv3.Client, id uint64) er
 			}
 		}
 	}
+
 	c.Members = newMembers
 	c.WaitMembersMatch(t, c.ProtoMembers())
 	return nil
@@ -524,7 +525,7 @@ func isMembersEqual(membs []*pb.Member, wmembs []*pb.Member) bool {
 }
 
 func NewLocalListener(t testutil.TB) net.Listener {
-	c := atomic.AddInt32(&LocalListenCount, 1)
+	c := atomic.AddInt32(&UniqueCount, 1)
 	// Go 1.8+ allows only numbers in port
 	addr := fmt.Sprintf("127.0.0.1:%05d%05d", c+BasePort, os.Getpid())
 	return NewListenerWithAddr(t, addr)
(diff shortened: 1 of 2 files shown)


tests/integration/grpc_test.go
```diff
@@ -154,7 +154,7 @@ func templateEndpoints(t *testing.T, pattern string, clus *integration.Cluster)
 	for _, m := range clus.Members {
 		ent := pattern
 		if strings.Contains(ent, "%d") {
-			ent = fmt.Sprintf(ent, integration.GrpcPortNumber(m.UniqNumber, m.MemberNumber))
+			ent = fmt.Sprintf(strings.Replace(ent, "%d", "%s", -1), m.GrpcPortNumber())
```
tjungblu, line 157: That turned out a bit weird due to the change in type and since we're checking for %s below as well, this has some unexpected clashes.

tests/framework/integration/cluster.go
```diff
@@ -543,6 +544,7 @@ type Member struct {
 	config.ServerConfig
 	UniqNumber                     int
 	MemberNumber                   int
+	Port                           string
```
serathius, line 547: In what cases port is not a int? Could we just cast port number to int?

tjungblu: in the unix socket case, there it's the name (eg m0). It's weird, I also started with an integer, but almost all stdlib methods are also using string for the port - so I ended up fmt.Sprintf all over the place.

tests/integration/grpc_test.go
serathius: Why not update patterns to change `%d` to `%s`

tjungblu: unfortunately this test controls the input by which type of pattern is used:
https://github.com/etcd-io/etcd/blob/ba9f5b557f1dfca3444259c587fb751ee238b031/tests/integration/grpc_test.go#L156-L161

so when I would change it it would always fill it with m.Name.
Is it worth to refactor the whole test to make this work, wdyt?

serathius: ah, ok it's terrible :(

serathius:
```suggestion
			ent = fmt.Sprintf(ent, strconv.Atoi(m.GrpcPortNumber()))
```
What about this?

tjungblu: great idea, implemented.

tests/framework/integration/cluster.go
```diff
@@ -790,8 +802,8 @@ func (m *Member) clientScheme() string {
 
 func (m *Member) addBridge() (*bridge, error) {
 	network, host, port := m.grpcAddr()
-	grpcAddr := host + ":" + port
-	bridgeAddr := grpcAddr + "0"
+	grpcAddr := net.JoinHostPort(host, m.Port)
+	bridgeAddr := net.JoinHostPort(host, fmt.Sprintf("%s%s", port, "0"))
```
serathius, line 806: Should bridge also bind a random port instead of basing it of `m.Port`?

serathius:
FYI https://github.com/etcd-io/etcd/pull/14586

tests/framework/integration/cluster.go
tjungblu: That's a good idea, not sure that append 0 trick works all the time - I've left it for the `unix://` for the time being.

tests/framework/integration/cluster.go
```diff
@@ -790,8 +802,11 @@ func (m *Member) clientScheme() string {
...
+	if m.UseTCP {
+		bridgeAddr = net.JoinHostPort(host, "0")
+	}
```
serathius, line 809:
```suggestion
    bridgePort := fmt.Sprintf("%s%s", port, "0")
    if m.UseTCP {
		bridgeAddr = "0"
	}
	bridgeAddr := net.JoinHostPort(host, bridgePort)
```

ahrtr:
@tjungblu  Please rebase this PR.

tests/framework/integration/cluster.go
tjungblu: applied

tjungblu:
@ahrtr done.

serathius (approved):

Issue #14540, opened 2022-09-30
[Test] TestDowngradeUpgradeClusterOf3 timeout


chaochn47:
### Which github workflows are flaking?

test (linux-amd64-e2e)

### Which tests are flaking?

TestDowngradeUpgradeClusterOf3

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/3156389973/jobs/5136053128

### Reason for failure (if possible)

```
--- FAIL: TestDowngradeUpgradeClusterOf3 (32.84s)
logger.go:130: 2022-09-30T04:57:21.930Z	INFO	started server.	{"name": "TestDowngradeUpgradeClusterOf3-test-2", "pid": 10915}
logger.go:130: 2022-09-30T04:57:21.930Z	INFO	started server.	{"name": "TestDowngradeUpgradeClusterOf3-test-0", "pid": 10916}
logger.go:130: 2022-09-30T04:57:21.941Z	INFO	started server.	{"name": "TestDowngradeUpgradeClusterOf3-test-1", "pid": 10921}
cluster_downgrade_test.go:57: Cluster created
cluster_downgrade_test.go:59: etcdctl downgrade enable 3.5
cluster_downgrade_test.go:62: Downgrade enabled, validating if cluster is ready for downgrade
cluster_downgrade_test.go:64: failed to find match string: context deadline exceeded
logger.go:130: 2022-09-30T04:57:52.050Z	INFO	closing test cluster...
```

e2e test cluster logging

```
/home/runner/work/etcd/etcd/bin/etcd (TestDowngradeUpgradeClusterOf3-test-0) (10916): 
{
	"level": "info",
	"ts": "2022-09-30T04:57:25.934Z",
	"caller": "etcdserver/server.go:2083",
	"msg": "saved snapshot",
	"snapshot-index": 10
}
/home/runner/work/etcd/etcd/bin/etcd (TestDowngradeUpgradeClusterOf3-test-0) (10916): 
{
	"level": "info",
	"ts": "2022-09-30T04:57:25.935Z",
	"caller": "etcdserver/server.go:2113",
	"msg": "compacted Raft logs",
	"compact-index": 1
}
/home/runner/work/etcd/etcd/bin/etcd (TestDowngradeUpgradeClusterOf3-test-2) (10915): 
{
	"level": "error",
	"ts": "2022-09-30T04:57:29.929Z",
	"caller": "version/monitor.go:81",
[...]

PR #14587, opened 2022-10-14
Allow downgrade to proceed on same version

tjungblu:
This is to fix the downgrade/upgrade test that sometimes is stuck on the same version already downgraded to.

"fix" for #14540


serathius:
PTAL https://github.com/etcd-io/etcd/pull/14588

tjungblu:
@serathius let's take yours instead. /close.

PR #14596, opened 2022-10-17
Simplify e2e coverage code

serathius:
cc @ahrtr @spzala @ptabor

### DIFF
--- a/tests/framework/e2e/etcd_spawn_cov.go
+++ b/tests/framework/e2e/etcd_spawn_cov.go
@@ -40,9 +40,10 @@ func init() {
 
 func initBinPathCov(binDir string) binPath {
 	return binPath{
-		Etcd:    binDir + "/etcd_test",
-		Etcdctl: binDir + "/etcdctl_test",
-		Etcdutl: binDir + "/etcdutl_test",
+		Etcd:            binDir + "/etcd_test",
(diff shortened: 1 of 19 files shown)


tests/framework/e2e/etcd_spawn_cov.go
```diff
@@ -35,40 +32,21 @@ var (
 	coverDir = testutils.MustAbsPath(os.Getenv("COVERDIR"))
 )
 
-func SpawnCmdWithLogger(lg *zap.Logger, args []string, envVars map[string]string, name string) (*expect.ExpectProcess, error) {
-	cmd := args[0]
-	env := mergeEnvVariables(envVars)
-	switch {
-	case strings.HasSuffix(cmd, "/etcd"):
-		cmd = cmd + "_test"
```
ahrtr, line 43: Do you understand why the ` _test `suffix was added in the first place? 

cc @ptabor

serathius: Yes, the coverage scenario compiles separate binaries with `_test` suffix.

ahrtr: I was thinking removing the suffix "_test" may cause some issue. I just realized that you do not remove the suffix, which is hard coded in `initBinPathCov`.

PR #14500, opened 2022-09-21
Release-3.5: server/etcdmain: add configurable cipher list to gRPC proxy listener

dusk125:
This is a backport of #14308

PR #14601, opened 2022-10-19
Backport #14500 to 3.4

dusk125:
server/etcdmain: add configurable cipher list to gRPC proxy listener


ahrtr:
Backporting https://github.com/etcd-io/etcd/pull/14500

pkg/tlsutil/cipher_suites.go
```diff
@@ -16,38 +14,43 @@
...
+		if s == c.Name {
+			return c.ID, true
+		}
+	}
+	switch s {
+	case "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305":
+		return tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256, true
+	case "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305":
+		return tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256, true
+	}
```
ahrtr, line 40: Golang 1.16.15 already includes these two cipher suites, why do you process then separately?

FYI. https://github.com/golang/go/blob/go1.16.15/src/crypto/tls/cipher_suites.go#L71-L72

dusk125: This may be a unneeded copy from Sam's original PR; I don't have a specific reason to have this. I would be happy to remove it, however it does align this version with the future versions.

ahrtr: > the future versions

Do you mean future golang version? All golang future version already includes these two cipher suites. So we need to remove them from main and release-3.5 as well. 

@ptabor could you share the thought why did you add this (process these two cipher suite separately) in the first place?

ahrtr: Double checked the go library src, it includes both `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256` and `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256`, but we (etcd) also supports `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305` and `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305`. Let keep it as it's for now.

ahrtr (approved):
LGTM

Thank you @dusk125

Issue #14603, opened 2022-10-20
runtime: goroutine stack exceeds 1000000000-byte limit


stonever:
### What happened?

call function: func (b *DoubleBarrier) Leave() error, got fatal error

### What did you expect to happen?

no fatal error

### How can we reproduce it (as minimally and precisely as possible)?

not leave  barrier for a long time

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
etcd Version: 3.5.4
Git SHA: 08407ff76
Go Version: go1.16.15
Go OS/Arch: linux/amd64

$ etcdctl version
# paste output here
```
etcdctl version: 3.5.4
API version: 3.5

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here
+------------------+---------+--------+----------------------------------------------------------+---------------------------------------------------------------------------------------------------+------------+
|        ID        | STATUS  |  NAME  |                        PEER ADDRS                        |                                           CLIENT ADDRS                                            | IS LEARNER |
+------------------+---------+--------+----------------------------------------------------------+---------------------------------------------------------------------------------------------------+------------+
| 2f15c3c81b1a3934 | started | etcd-1 | http://etcd-1.etcd-headless.kline.svc.cluster.local:2380 | http://etcd-1.etcd-headless.kline.svc.cluster.local:2379,http://etcd.kline.svc.cluster.local:2379 |      false |
| 6883d19530f1b73d | started | etcd-0 | http://etcd-0.etcd-headless.kline.svc.cluster.local:2380 | http://etcd-0.etcd-headless.kline.svc.cluster.local:2379,http://etcd.kline.svc.cluster.local:2379 |      false |
| 7a0e402caf21665c | started | etcd-2 | http://etcd-2.etcd-headless.kline.svc.cluster.local:2380 | http://etcd-2.etcd-headless.kline.svc.cluster.local:2379,http://etcd.kline.svc.cluster.local:2379 |      false |
+------------------+---------+--------+----------------------------------------------------------+---------------------------------------------------------------------------------------------------+------------+

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```shell
runtime: goroutine stack exceeds 1000000000-byte limit
[...]

PR #14604, opened 2022-10-20
clientv3: fix the implementation of double barrier

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/14603

Check the waiter count before creating the ephemeral key, do not create the key if there are already too many clients. Check the count after creating the key again, if the total kvs is bigger than the expected count, then check the rev of the current key, and take action accordingly based on its rev. If its rev is in the first ${count}, then it's valid client, otherwise, it should fail.

### DIFF
--- a/client/v3/experimental/recipes/double_barrier.go
+++ b/client/v3/experimental/recipes/double_barrier.go
@@ -45,25 +45,46 @@ func NewDoubleBarrier(s *concurrency.Session, key string, count int) *DoubleBarr
 // Enter waits for "count" processes to enter the barrier then returns
 func (b *DoubleBarrier) Enter() error {
 	client := b.s.Client()
+
+	// Check the entered clients before creating the UniqueEphemeralKey,
+	// fail the request if there are already too many clients.
+	if resp1, err := b.enteredClients(client); err != nil {
+		return err
+	} else if len(resp1.Kvs) >= b.count {
+		return ErrTooManyClients
+	}
+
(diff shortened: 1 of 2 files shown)


ahrtr:
Please note that the logic is similar to the double check mechanism (`check  --> register --> check`) used in [v3discovery](https://github.com/etcd-io/etcd/blob/e5790d204c6a3d19941cb11a1d7d9b8901fee81f/server/etcdserver/api/v3discovery/discovery.go#L208-L228). 

If there are multiple clients register concurrently, then it's possible the registered client count is greater than the expected count (assuming N). In this case, we just sort all the client by createRevision, and the first N clients are valid, and all others are invalid.

ahrtr:
Please note that the logic is similar to the double check mechanism (`check  --> register --> check`) used in [v3discovery](https://github.com/etcd-io/etcd/blob/e5790d204c6a3d19941cb11a1d7d9b8901fee81f/server/etcdserver/api/v3discovery/discovery.go#L208-L228). 

If there are multiple clients register concurrently, then it's possible the registered client count is greater than the expected count (assuming N). In this case, we just sort all the client by createRevision, and the first N clients are valid, and all others are invalid. 

cc @serathius @spzala @ptabor

ahrtr:
This PR should can fix issue https://github.com/etcd-io/etcd/issues/14603, please also take a look, thx. cc @serathius @spzala @ptabor @mitake

spzala (approved):
lgtm, thanks @ahrtr

ahrtr:
Thanks @spzala

PR #14610, opened 2022-10-23
Add etcd release compatibility tests

serathius:
cc @ahrtr

### DIFF
--- a/tests/common/e2e_test.go
+++ b/tests/common/e2e_test.go
@@ -18,6 +18,7 @@
 package common
 
 import (
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
 	"go.etcd.io/etcd/tests/v3/framework"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
@@ -25,6 +26,52 @@ import (
 
 func init() {
 	testRunner = framework.E2eTestRunner
+	clusterTestCases = e2eClusterTestCases
+}
+
+func e2eClusterTestCases() []testCase {
+	tcs := []testCase{
+		{
+			name:   "NoTLS",
+			config: config.ClusterConfig{ClusterSize: 1},
+		},
+		{
+			name:   "PeerTLS",
+			config: config.ClusterConfig{ClusterSize: 3, PeerTLS: config.ManualTLS},
+		},
+		{
+			name:   "PeerAutoTLS",
+			config: config.ClusterConfig{ClusterSize: 3, PeerTLS: config.AutoTLS},
+		},
+		{
+			name:   "ClientTLS",
+			config: config.ClusterConfig{ClusterSize: 1, ClientTLS: config.ManualTLS},
+		},
+		{
+			name:   "ClientAutoTLS",
+			config: config.ClusterConfig{ClusterSize: 1, ClientTLS: config.AutoTLS},
+		},
+	}
+	if fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
+		tcs = append(tcs, testCase{
+			name: "MinorityLastVersion",
+			config: config.ClusterConfig{
+				ClusterSize: 3,
+				E2eConfig: &config.E2eClusterConfig{
+					Version: config.MinorityLastVersion,
+				},
+			},
+		}, testCase{
+			name: "QuorumLastVersion",
+			config: config.ClusterConfig{
+				ClusterSize: 3,
+				E2eConfig: &config.E2eClusterConfig{
+					Version: config.QuorumLastVersion,
+				},
+			},
+		})
+	}
+	return tcs
 }
 
 func WithAuth(userName, password string) config.ClientOption {
--- a/tests/framework/config/cluster.go
+++ b/tests/framework/config/cluster.go
@@ -14,14 +14,22 @@
 
 package config
 
-import "time"
+import (
+	"time"
+)
 
 type TLSConfig string
+type ClusterVersion string
 
 const (
 	NoTLS     TLSConfig = ""
 	AutoTLS   TLSConfig = "auto-tls"
 	ManualTLS TLSConfig = "manual-tls"
+
+	CurrentVersion      ClusterVersion = ""
+	MinorityLastVersion ClusterVersion = "minority-last-version"
+	QuorumLastVersion   ClusterVersion = "quorum-last-version"
+	LastVersion         ClusterVersion = "last-version"
 
 	TickDuration = 10 * time.Millisecond
 )
@@ -34,6 +42,12 @@ type ClusterConfig struct {
 	StrictReconfigCheck bool
 	AuthToken           string
 	SnapshotCount       int
+
+	E2eConfig *E2eClusterConfig
+}
+
+type E2eClusterConfig struct {
+	Version ClusterVersion
 }
 
 func DefaultClusterConfig() ClusterConfig {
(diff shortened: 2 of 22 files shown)


ahrtr:
It seems you are doing multiple things in one PR. Please try to break down this PR, and each PR should only do one thing. 

Such as renaming `etcd-last-release` to `etcd-previous-release`, let's get it fixed in a separate PR, so that we can get it merged as soon as possible.

tests/framework/config/cluster.go
```diff
@@ -14,15 +14,23 @@
...
 
 const (
 	NoTLS     TLSConfig = ""
 	AutoTLS   TLSConfig = "auto-tls"
 	ManualTLS TLSConfig = "manual-tls"
 
+	CurrentVersion      ClusterVersion = ""
+	MinorityLastVersion ClusterVersion = "minority-last-version"
+	QuorumLastVersion   ClusterVersion = "quorum-last-version"
+	LastVersion         ClusterVersion = "last-version"
```
ahrtr, line 32: We shouldn't have this option `LastVersion`, because it doesn't make sense to test previous version in `main` branch CI.

tests/common/e2e_test.go
```diff
@@ -17,8 +17,59 @@
...
+		},
+		{
+			name:   "ClientTLS",
+			config: config.ClusterConfig{ClusterSize: 1, ClientTLS: config.ManualTLS},
+		},
+		{
+			name:   "ClientAutoTLS",
+			config: config.ClusterConfig{ClusterSize: 1, ClientTLS: config.AutoTLS},
+		},
+	}
```
ahrtr, line 54: These cases are duplicated to the items in `integrationClusterTestCases`. Have you considered to define a `commonClusterTestCases`?

tests/framework/config/cluster.go
```diff
@@ -34,4 +42,10 @@ type ClusterConfig struct {
 	DisableStrictReconfigCheck bool
 	AuthToken                  string
 	SnapshotCount              int
+
+	E2eConfig *E2eClusterConfig
```
ahrtr, line 46: It's weird to see e2e config in the common `ClusterConfig`, which shouldn't care about the test infra/env (e2e or integration) at all.

I would suggest to move all the code from common area to `e2e` packages.

tests/framework/config/cluster.go
serathius: Upgrade tests start from LastVersion.

tests/common/e2e_test.go
serathius: I considered it, but found I don't think we should do a premature refactor and jump on every code duplication.

tests/framework/config/cluster.go
serathius: This PR maintains invariant that tests in `common` do not care about cluster setup. However this should not be misunderstood that common framework itself doesn't care about e2e configuration. Goal of common framework is to abstract cluster setup so we can write a single test that works can be used as integration, e2e or upgrade.

ahrtr: > I would suggest to move all the code from common area to e2e packages.

Have you considered this?

serathius: Not sure if I understand what you mean by "all code". I think that putting `E2eConfig` into `framework.Config` config is a good decision as it makes it explicitly part of cluster configuration. I think it's more important that maintaining `framework.Config` generic. 

If we want to create a cluster with different server binaries, it should be part of cluster config, even if we cannot setup such cluster during integration tests. That's why instrumentation tests will immidietly panic when they discover `E2eConfig` set.

As for other code added to package, I cannot move it to e2e because it would create an import cycle.

tests/common/e2e_test.go
ahrtr: Although I don't entirely agree this, it isn't a big deal.

tests/framework/config/cluster.go
serathius: > Points:
> 
> 1. Just as the comment [Add etcd release compatibility tests #14610 (comment)](https://github.com/etcd-io/etcd/pull/14610#issuecomment-1288300788) I raised previously, let's try to do one thing in one PR.  If you get the change on `e2eClusterTestCases()` in a separate PR, it might have already been approved and merged.

It doesn't make sense to move `e2eClusterTestCases` as splitting integration and e2e scenarios is not nesesery if we change implementation of this PR.

> 2. "all code" means all code related to `E2eConfig`, which means it doesn't include the code related to `e2eClusterTestCases()` mentioned above.

As mentioned above code for `E2eConfig` cannot be moved because of import cycles. 

> 3. Regarding to `it should be part of cluster config`, why not to add whatever config specific to e2e test into `EtcdProcessClusterConfig` and test cases specific to e2e into `e2e` package? It's really weird to add `E2eConfig` into the common `ClusterConfig`.

When thinking about unifying testing via common framework we need to separate test scenario from cluster setup (need to work on naming to avoid confusion). By test scenario I mean a test like "TestPutKV" it just takes a cluster, puts a key and verifies that it's there. Test scenario itself doesn't care how cluster was created. 

This brings me to testing clusters with servers having different versions. This is not a test scenario, but a cluster setup. It doesn't make sense to have a specific test scenarios just because it requires some specific e2e configuration. We can reuse all existing common test scenarios and run them against the specific e2e cluster configuration. 

You keep saying that `ClusterConfig` is common, but it's not part of common package not it was meant to be generic. For now it only has properties shared by both e2e and integration tests, but that just because we haven't migrated any test that would require a e2e specific cluster setup. 


> 4. With regard to "`However this should not be misunderstood that common framework itself doesn't care about e2e configuration`",  I don't agree this. The diagram I provided about 9 months ago is still valid. FYI. [Unify testing framework #13637 (comment)](https://github.com/etcd-io/etcd/issues/13637#issuecomment-1022721677)

Thank's for diagram, however it's not set in stone truth. Original proposal didn't set such limitations.

serathius:
ping @ahrtr @spzala @ptabor

tests/framework/config/cluster.go
ahrtr: Don't you think [ClusterConfig](https://github.com/etcd-io/etcd/blob/0dfd7264c0b7c8e02e4a7c45259d35543a02d462/tests/framework/config/cluster.go#L29) is a common struct, which is used by both e2e and integration test, and it's an anti-pattern for [integration checks something it shouldn't care about](https://github.com/etcd-io/etcd/blob/9ceec77dc20d063f38ec733e58057375c0e81c3e/tests/framework/integration.go#L48-L50) at all?

ahrtr: Please see also https://github.com/etcd-io/etcd/pull/14683

serathius: There are two parts we need to agree on:
* Should hybrid version tests be part of common tests.
* How to integrate E2e config into common tests. 

First, hybrid version clusters are expected to behave exactly like a normal cluster. This means that we can and should run exactly the same tests as we do for other tests. Common tests are just a blackbox tests that replay scenarios that should work no matter if we use client library, etcdctl, single cluster node or multi cluster cluster node. They should also work in hybrid clusters (minus new features but however we also test  protection against inconsistency during upgrade due to usage of new features). 

Alternative that you proposed to have those tests as e2e does not scale. We would again be in a situation that we need to copy each and every test just change one line in configuration. Instead of thinking of hybrid clusters as a separate special test we should think it as a normal cluster that just has a different setup.

Does it make sense?

ahrtr: See https://github.com/etcd-io/etcd/pull/14697

PR #14611, opened 2022-10-23
tests: Add linearizability tests scenario for #14370

serathius:
cc @ahrtr @spzala

### DIFF
--- a/scripts/build_lib.sh
+++ b/scripts/build_lib.sh
@@ -19,6 +19,22 @@ toggle_failpoints() {
   mode="$1"
   if command -v gofail >/dev/null 2>&1; then
     run gofail "$mode" server/etcdserver/ server/storage/backend/
+    (
+      cd ./server
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./etcdutl
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./etcdctl
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./tests
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
   elif [[ "$mode" != "disable" ]]; then
     log_error "FAILPOINTS set but gofail not found"
     exit 1
--- a/tests/linearizability/failpoints.go
+++ b/tests/linearizability/failpoints.go
@@ -15,14 +15,20 @@
 package linearizability
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 	"math/rand"
+	"net/http"
+	"net/url"
+	"time"
 
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
 
 var (
-	KillFailpoint Failpoint = killFailpoint{}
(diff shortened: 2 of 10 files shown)


ahrtr:
Please fix the static analysis error.

scripts/build_lib.sh
```diff
@@ -19,6 +19,22 @@ toggle_failpoints() {
...
+      cd ./etcdutl
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./etcdctl
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./tests
+      run go get go.etcd.io/gofail/runtime
```
ahrtr, line 36: Why do we need to get `gofail/runtime` under `tests` ?

scripts/build_lib.sh
```diff
@@ -19,6 +19,22 @@ toggle_failpoints() {
...
+      cd ./server
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./etcdutl
+      run go get go.etcd.io/gofail/runtime
+    ) || exit 2
+    (
+      cd ./etcdctl
+      run go get go.etcd.io/gofail/runtime
```
ahrtr, line 32: It isn't needed to get `gofail/runtime` under `etcdctl`.

tests/linearizability/failpoints.go
```diff
@@ -47,3 +53,51 @@ func (f killFailpoint) Trigger(ctx context.Context, clus *e2e.EtcdProcessCluster
 	}
 	return nil
 }
+
+type goFailpoint struct {
+	failpoint string
+	payload   string
+}
+
+func (f goFailpoint) Trigger(ctx context.Context, clus *e2e.EtcdProcessCluster) error {
```
ahrtr, line 62: Please add method to support deleting the `failpoints`, so that test cases can reset the environment after finished.

scripts/build_lib.sh
serathius: Because tests import server.

scripts/build_lib.sh
serathius: etcdctl import server

tests/linearizability/failpoints.go
serathius: I don't understand what you mean by reseting the environment. Every e2e tests should spin their own etcd cluster so why I should delete the failpoint.

scripts/build_lib.sh
ahrtr: ack

ahrtr: Based on my previous test, we don't need to get `gofail/runtime` under `etcdctl`. 

But it isn't a big deal.

scripts/build_lib.sh
ahrtr: ack

tests/linearizability/failpoints.go
ahrtr: You are right!

ahrtr (approved):
LGTM

Issue #14533, opened 2022-09-28
Protobuf: cleanup both golang/protobuf and gogo/protobuf


ahrtr:
### What would you like to be added?

Both `github.com/golang/protobuf` and `github.com/gogo/protobuf` are deprecated. 

- [github.com/golang/protobuf](https://github.com/golang/protobuf) is the first version of Go protocol buffer API.  **Deprecated**. 
- [github.com/gogo/protobuf](https://github.com/gogo/protobuf) is actually a fork of `github.com/golang/protobuf` with extra code generation features. **Deprecated**. 
-  [google.golang.org/protobuf](https://github.com/protocolbuffers/protobuf-go) is the second version of the Go protocol buffer API. **Active**. 

In short, **[google.golang.org/protobuf](https://github.com/protocolbuffers/protobuf-go)** is the future, so we should cleanup both `github.com/golang/protobuf` and `github.com/gogo/protobuf`, and only use `google.golang.org/protobuf`.

### Why is this needed?

Same as above

cc @ptabor  @serathius @spzala @dims @liggitt

ahrtr:
This seems like a big change, please let me know if anyone has any comments or concerns.

ajityagaty:
@ahrtr - Has it been decided to make the change? If so, can I take a swing at this?

ahrtr:
@ajityagaty Thanks for your interest on this.

There is no any feedback from other maintainers yet, but I think we must do this because both `github.com/golang/protobuf` and `github.com/gogo/protobuf` are deprected, it's just a matter of time. 

Since it's a big change, so please evaluate the impact and provide a detailed plan firstly. Afterwards, we can update the [functional_test](https://github.com/etcd-io/etcd/tree/main/tests/functional) in the first step.

ahrtr:
any updates on this? @ajityagaty

ajityagaty:
@ahrtr - Apologies! I have not been able to look at this yet as I have been overwhelmed at work. When I briefly looked at the work that is required, I realized that I am lacking context of how all components are dependent on this change. So, I think I would be better off taking a smaller piece of work.

ahrtr:
> @ahrtr - Apologies! I have not been able to look at this yet as I have been overwhelmed at work. When I briefly looked at the work that is required, I realized that I am lacking context of how all components are dependent on this change. So, I think I would be better off taking a smaller piece of work.

Thanks for the feedback. No worries.

ahrtr:
[...]

PR #14612, opened 2022-10-23
chore: commit the change generated by scripts/genproto.sh

spacewander:
TODO: ensure the generated code is up-to-date in the CI.

### DIFF
--- a/Documentation/dev-guide/apispec/swagger/rpc.swagger.json
+++ b/Documentation/dev-guide/apispec/swagger/rpc.swagger.json
@@ -36,7 +36,7 @@
             }
           },
           "default": {
-            "description": "An unexpected error response",
+            "description": "An unexpected error response.",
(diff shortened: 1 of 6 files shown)


spacewander:
This PR only regenerates some files, and the diff is mostly about changing the description. So I think the failure of test isn't caused by the change.

ptabor (approved):
Thank you. 

We should have a test that makes sure that regeneration is an no-op.

ahrtr:
> We should have a test that makes sure that regeneration is an no-op.

The reason should be that we do not set down the versions, https://github.com/etcd-io/etcd/blob/f215cd89d2e2d24def615bb6da4d0f3038ddc3f9/scripts/genproto.sh#L21-L25

We indeed to take care of it after we fix https://github.com/etcd-io/etcd/issues/14533

PR #14613, opened 2022-10-23
fix: add a warning message on startup if the server name is default

### LINKED ISSUE #13757: Duplicate names in `ETCD_INITIAL_CLUSTER` not handled correctly
### What happened?

If you don't pass a `--name` argument to your `etcd` processes, they will all have the name `default` and the cluster will operate normally.  However, when you add a member, the generated `ETCD_INITIAL_CLUSTER` variable will have multiple entries with the name "default".  When this environment variable is used, `etcd` will parse these into a mapping under a single key ("default") with multiple URLs, and create a single member.  See https://github.com/etcd-io/etcd/blob/63a1cc3fe40bace6898289dec35a9aad05163889/server/etcdserver/api/membership/cluster.go#L83-L86

This leads to the confusing error message "member count is unequal".  The documentation on https://etcd.io/docs/v3.5/op-guide/runtime-configuration/ mentions this failure, but the situation is different.

Either

a. `member add` should fail, saying it cannot generate a valid `ETCD_INITIAL_CLUSTER` due to duplicate names, or
b. `etcd` should accept duplicate names in `ETCD_INITIAL_CLUSTER` and treat them as separate members.  This can be accomplished by updating `func NewClusterFromURLsMap` as follows:

```
	c := NewCluster(lg, opts...)
	for name, urls := range urlsmap {
		for idx, _ := range urls {
			m := NewMember(name, urls[idx:idx+1], token, nil)
			[...]
```

I don't know if there's a real need to be able to specify multiple URLs for a single name.

You need three terminals, `x, `y`, and `z`:
```
x$ mkdir -p test_case/{a,b,c}/{data/member,wal}
x$ ETCD_INITIAL_CLUSTER="a=http://127.0.0.1:40000,b=http://127.0.0.1:40001" ETCD_INITIAL_CLUSTER_STATE=new etcd --name a --{initial-advertise,listen}-peer-urls=http://127.0.0.1:40000 --{advertise,listen}-client-urls=http://127.0.0.1:50000 --data-dir test_case/a/data --wal-dir test_case/a/wal
y$ ETCD_INITIAL_CLUSTER="a=http://127.0.0.1:40000,b=http://127.0.0.1:40001" ETCD_INITIAL_CLUSTER_STATE=new etcd --name b --{initial-advertise,listen}-peer-urls=http://127.0.0.1:40001 --{advertise,listen}-client-urls=http://127.0.0.1:50001 --data-dir test_case/b/data --wal-dir test_case/b/wal
[now kill both servers with Ctrl-C]
x$ etcd --listen-peer-urls=http://127.0.0.1:40000 --{advertise,listen}-client-urls=http://127.0.0.1:50000 --data-dir test_case/a/data --wal-dir test_case/a/wal
y$ etcd --listen-peer-urls=http://127.0.0.1:40001 --{advertise,listen}-client-urls=http://127.0.0.1:50001 --data-dir test_case/b/data --wal-dir test_case/b/wal
z$ ETCDCTL_ENDPOINT=http://localhost:50000 etcdctl member add c http://127.0.0.1:40002
Added member named c with ID 7b4d6e3edb76bc59 to cluster

ETCD_NAME="c"
ETCD_INITIAL_CLUSTER="default=http://127.0.0.1:40000,c=http://127.0.0.1:40002,default=http://127.0.0.1:40001"
ETCD_INITIAL_CLUSTER_STATE="existing"
z$ export ETCD_NAME="c"
z$ export ETCD_INITIAL_CLUSTER="default=http://127.0.0.1:40000,c=http://127.0.0.1:40002,default=http://127.0.0.1:40001"
z$ export ETCD_INITIAL_CLUSTER_STATE="existing"
z$ etcd --listen-peer-urls=http://127.0.0.1:40002 --{advertise,listen}-client-urls=http://127.0.0.1:50002 --data-dir test_case/c/data --wal-dir test_case/c/wal
[...]
member count is unequal
```

_No response_

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

<details>

</details>

etcd 3.5.2

### Relevant log output

_No response_

nic-chen:
fix #13757

### DIFF
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -765,6 +765,15 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("--experimental-compact-hash-check-time must be >0 (set to %v)", cfg.ExperimentalCompactHashCheckTime)
 	}
 
+	// If `--name` isn't configured, then multiple members may have the same "default" name.
+	// When adding a new member with the "default" name as well, etcd may regards its peerURL
+	// as one additional peerURL of the existing member which has the same "default" name,
+	// because each member can have multiple client or peer URLs.
+	// Please refer to https://github.com/etcd-io/etcd/issues/13757
+	if cfg.Name == DefaultName {
+		cfg.logger.Warn("the server is using default name, which might cause error that failed the startup.")
+	}
+
 	return nil
 }
 
--- a/server/etcdmain/etcd.go
+++ b/server/etcdmain/etcd.go
@@ -90,6 +90,10 @@ func startEtcdOrProxyV2(args []string) {
 		lg.Info("failed to detect default host", zap.Error(dhErr))
 	}
 
+	if cfg.ec.Name == embed.DefaultName {
+		lg.Warn("the server is using default name, which might cause error that failed the startup.")
+	}
+
 	if cfg.ec.Dir == "" {
 		cfg.ec.Dir = fmt.Sprintf("%v.etcd", cfg.ec.Name)
 		lg.Warn(
--- a/server/etcdmain/etcd.go
+++ b/server/etcdmain/etcd.go
@@ -90,10 +90,6 @@ func startEtcdOrProxyV2(args []string) {
 		lg.Info("failed to detect default host", zap.Error(dhErr))
 	}
 
-	if cfg.ec.Name == embed.DefaultName {
-		lg.Warn("the server is using default name, which might cause error that failed the startup.")
-	}
-
 	if cfg.ec.Dir == "" {
 		cfg.ec.Dir = fmt.Sprintf("%v.etcd", cfg.ec.Name)
 		lg.Warn(


server/etcdserver/api/membership/cluster.go
```diff
@@ -83,14 +83,16 @@ const (
...
-		m := NewMember(name, urls, token, nil)
-		if _, ok := c.members[m.ID]; ok {
-			return nil, fmt.Errorf("member exists with identical ID %v", m)
-		}
-		if uint64(m.ID) == raft.None {
-			return nil, fmt.Errorf("cannot use %x as member id", raft.None)
+		for idx := range urls {
+			m := NewMember(name, urls[idx:idx+1], token, nil)
+			if _, ok := c.members[m.ID]; ok {
+				return nil, fmt.Errorf("member exists with identical ID %v", m)
```
ahrtr, line 89: You might be heading in the wrong direction. Usually each member has different peerURL(s), they must have different memberID. FYI. https://github.com/etcd-io/etcd/blob/f215cd89d2e2d24def615bb6da4d0f3038ddc3f9/server/etcdserver/api/membership/member.go#L63-L76

For simplicity, we can just add a warning message on startup if `--name `isn't provided a value.

nic-chen: Thank you very much for the reminder.

server/etcdmain/etcd.go
```diff
@@ -90,6 +90,10 @@ func startEtcdOrProxyV2(args []string) {
 		lg.Info("failed to detect default host", zap.Error(dhErr))
 	}
 
+	if cfg.ec.Name == embed.DefaultName {
+		lg.Warn("the server is using default name, which might cause error that failed the startup.")
+	}
```
ahrtr, line 95: I suggest to add the check into [(*Config) Validate()](https://github.com/etcd-io/etcd/blob/9bc4a63a419cd316f6920e37ec551b87d1c7c3a7/server/embed/config.go#L665), so that we have all configuration checker in one place.

ahrtr, line 95:
```suggestion
	// If `--name` isn't configured, then multiple members may have the same "default" name. 
	// When adding a new member with the "default" name as well, etcd may regards its peerURL 
	// as one additional peerURL of the existing member which has the same "default" name, 
	// because each member can have multiple client or peer URLs. 
	// Please refer to https://github.com/etcd-io/etcd/issues/13757
	if cfg.ec.Name == embed.DefaultName {
		lg.Warn("the server is using default name, which might cause error that failed the startup.")
	}
```

nic-chen: Thanks! 
I tried to find such a place, but failed. 
So I put it here according to the content in the log `'data-dir' was empty; using default`...

nic-chen: updated.

server/embed/config.go
```diff
@@ -765,6 +765,15 @@ func (cfg *Config) Validate() error {
 		return fmt.Errorf("--experimental-compact-hash-check-time must be >0 (set to %v)", cfg.ExperimentalCompactHashCheckTime)
 	}
 
+	// If `--name` isn't configured, then multiple members may have the same "default" name.
+	// When adding a new member with the "default" name as well, etcd may regards its peerURL
+	// as one additional peerURL of the existing member which has the same "default" name,
+	// because each member can have multiple client or peer URLs.
+	// Please refer to https://github.com/etcd-io/etcd/issues/13757
+	if cfg.Name == DefaultName {
```
ahrtr, line 773:
```suggestion
	if cfg.Name == "" || cfg.Name == DefaultName {
```

ahrtr: It seems etcd fails to get started if an empty string "" is provided for `--name`, so feel free to ignore this comment.

ahrtr (approved):
LGTM

Thank you @nic-chen

server/embed/config.go
```diff
@@ -765,6 +765,15 @@ func (cfg *Config) Validate() error {
...
+		cfg.logger.Warn("the server is using default name, which might cause error that failed the startup.")
```
ahrtr, line 774: I was planning to add a comment for this, but somehow forgot to do it:(

Please change it to something like below,
```
return fmt.Errorf("it isn't recommended to use %q name, please set a value for --name. Note that etcd might run into issue when multiple members have the same default name", DefaultName)
```

ahrtr: Please ignore my previous comment. We can't return an error, because etcd may fail to get started. Let's just update the message,
```
cfg.logger.Warn("it isn't recommended to use %q name, please set a value for --name. Note that etcd might run into issue when multiple members have the same default name", DefaultName)
```

nic-chen: OK, I will update it later today or tomorrow

nic-chen: OK, I will update it later today or tomorrow

nic-chen: updated.   https://github.com/etcd-io/etcd/pull/14642
please take a look at your convenience. thanks.

Issue #14324, opened 2022-08-09
Bad Certificates should be explained, not just stated.


illeatmyhat:
### What happened?

I previously set up a 3-node cluster using bi-directional TLS, which was working before
Clients attempting to call etcd using a previously working certificate now fails with:
```
{"level":"warn","ts":"2022-08-09T10:28:48.597Z","caller":"embed/config_logging.go:169","msg":"rejected connection","remote-addr":"172.30.214.50:39334","server-name":"etcd-0.etcd","error":"remote error: tls: bad certificate"}
```

### What did you expect to happen?

I expected the logs to explain why the certificate was bad instead of stonewalling me

### How can we reproduce it (as minimally and precisely as possible)?

Create an etcd cluster and connect to it using a bad certificate. Any flavor you want. Pretend to be a clueless user who knows nothing about certificates and wonder what's wrong with your certificate

### Anything else we need to know?

The client and etcd use the same CA Issuer, hence the etcd configuration below.

<details>
<summary>Etcd Server's working peer certificate </summary>

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            ...
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=wmlserving-ca
        Validity
            Not Before: Jun 20 20:00:46 2022 GMT
            Not After : Sep 18 20:00:46 2022 GMT
        Subject: CN=*.etcd
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                RSA Public-Key: (4096 bit)
                Modulus:
                    ...
                Exponent: 65537 (0x10001)
        X509v3 extensions:
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Authority Key Identifier: 
                keyid:...

            X509v3 Subject Alternative Name: 
                DNS:localhost, DNS:etcd, DNS:*.etcd, DNS:etcd.argo-wo, DNS:*.etcd.argo-wo, DNS:etcd.argo-wo.svc, DNS:*.etcd.argo-wo.svc, DNS:etcd.argo-wo.svc.cluster.local, DNS:*.etcd.argo-wo.svc.cluster.local
    Signature Algorithm: sha256WithRSAEncryption
         ...

(Decoded using the following version of OpenSSL: OpenSSL 1.1.1b  26 Feb 2019)
```

</details>

<details>
[...]

PR #14617, opened 2022-10-24
Issue 14324  Bad Certificates should be explained, not just stated

EmilyM1:
Fixing error message for cert expiration case: https://github.com/etcd-io/etcd/issues/14324


serathius:
Please clean the commit history

PR #14623, opened 2022-10-25
concurrency: Make `waitDeletes` faster and decrease the number of using`Get`

SimFG:


SimFG:
@ahrtr @serathius PTAL when you have time.

PR #14624, opened 2022-10-25
raft: add MaxInflightBytes limit to the outgoing messages flow

pavelkalinnikov:
Raft flow control limits the message size and the number of inflight messages. However, a single large entry that exceeds the size limit can still be sent. In combination with the max messages count limit, many large messages can be sent in a row and overflow the receiver. In effect, the "max" size value acts as "target" which can be missed.

This PR adds an additional soft limit on the total size of inflight messages, which catches such situations and prevents the receiver overflow.

### DIFF
--- a/CHANGELOG/CHANGELOG-3.6.md
+++ b/CHANGELOG/CHANGELOG-3.6.md
@@ -46,6 +46,10 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 - Package `wal` was moved to `storage/wal`
 - Package `datadir` was moved to `storage/datadir`
 
+### Package `raft`
+- Send empty `MsgApp` when entry in-flight limits are exceeded. See [pull/14633](https://github.com/etcd-io/etcd/pull/14633).
+- Add [MaxInflightBytes](https://github.com/etcd-io/etcd/pull/14624) setting in `raft.Config` for better flow control of entries.
+
 ### etcd server
 
 - Add [`etcd --log-format`](https://github.com/etcd-io/etcd/pull/13339) flag to support log format.
--- a/raft/confchange/confchange.go
+++ b/raft/confchange/confchange.go
@@ -265,7 +265,7 @@ func (c Changer) initProgress(cfg *tracker.Config, prs tracker.ProgressMap, id u
 		// making the first index the better choice).
 		Next:      c.LastIndex,
 		Match:     0,
-		Inflights: tracker.NewInflights(c.Tracker.MaxInflight, 0), // TODO: set maxBytes
+		Inflights: tracker.NewInflights(c.Tracker.MaxInflight, c.Tracker.MaxInflightBytes),
 		IsLearner: isLearner,
 		// When a node is first added, we should mark it as recently active.
 		// Otherwise, CheckQuorum may cause us to step down if it is invoked
--- a/raft/confchange/confchange.go
+++ b/raft/confchange/confchange.go
@@ -265,7 +265,7 @@ func (c Changer) initProgress(cfg *tracker.Config, prs tracker.ProgressMap, id u
 		// making the first index the better choice).
 		Next:      c.LastIndex,
 		Match:     0,
-		Inflights: tracker.NewInflights(c.Tracker.MaxInflight),
+		Inflights: tracker.NewInflights(c.Tracker.MaxInflight, 0), // TODO: set maxBytes
 		IsLearner: isLearner,
 		// When a node is first added, we should mark it as recently active.
 		// Otherwise, CheckQuorum may cause us to step down if it is invoked
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -1790,11 +1790,7 @@ func (r *raft) responseToReadIndexReq(req pb.Message, readIndex uint64) pb.Messa
 // Empty payloads are never refused. This is used both for appending an empty
 // entry at a new leader's term, as well as leaving a joint configuration.
 func (r *raft) increaseUncommittedSize(ents []pb.Entry) bool {
-	var s uint64
-	for _, e := range ents {
-		s += uint64(PayloadSize(e))
-	}
-
+	s := payloadsSize(ents)
 	if r.uncommittedSize > 0 && s > 0 && r.uncommittedSize+s > r.maxUncommittedSize {
 		// If the uncommitted tail of the Raft log is empty, allow any size
 		// proposal. Otherwise, limit the size of the uncommitted tail of the
@@ -1816,12 +1812,7 @@ func (r *raft) reduceUncommittedSize(ents []pb.Entry) {
 		// Fast-path for followers, who do not track or enforce the limit.
 		return
 	}
-
-	var s uint64
-	for _, e := range ents {
-		s += uint64(PayloadSize(e))
-	}
-	if s > r.uncommittedSize {
+	if s := payloadsSize(ents); s > r.uncommittedSize {
 		// uncommittedSize may underestimate the size of the uncommitted Raft
 		// log tail but will never overestimate it. Saturate at 0 instead of
 		// allowing overflow.
@@ -1829,6 +1820,14 @@ func (r *raft) reduceUncommittedSize(ents []pb.Entry) {
 	} else {
 		r.uncommittedSize -= s
 	}
+}
+
+func payloadsSize(ents []pb.Entry) uint64 {
+	var s uint64
+	for _, e := range ents {
+		s += uint64(PayloadSize(e))
+	}
+	return s
 }
 
 func numOfPendingConf(ents []pb.Entry) int {
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -160,6 +160,16 @@ type Config struct {
 	// overflowing that sending buffer. TODO (xiangli): feedback to application to
 	// limit the proposal rate?
 	MaxInflightMsgs int
+	// MaxInflightBytes limits the number of in-flight bytes in append messages.
+	// Complements MaxInflightMsgs. Ignored if zero.
+	//
+	// This effectively bounds the bandwidth-delay product. Note that especially
+	// in high-latency deployments setting this too low can lead to a dramatic
+	// reduction in throughput. For example, with a peer that has a round-trip
+	// latency of 100ms to the leader and this setting is set to 1 MB, there is a
+	// throughput limit of 10 MB/s for this group. With RTT of 400ms, this drops
+	// to 2.5 MB/s. See Little's law to understand the maths behind.
+	MaxInflightBytes uint64
 
 	// CheckQuorum specifies if the leader should check quorum activity. Leader
 	// steps down when quorum is not active for an electionTimeout.
@@ -227,6 +237,11 @@ func (c *Config) validate() error {
 
 	if c.MaxInflightMsgs <= 0 {
 		return errors.New("max inflight messages must be greater than 0")
+	}
+	if c.MaxInflightBytes == 0 {
+		c.MaxInflightBytes = noLimit
+	} else if c.MaxInflightBytes < c.MaxSizePerMsg {
+		return errors.New("max inflight bytes must be >= max message size")
 	}
 
 	if c.Logger == nil {
@@ -332,7 +347,7 @@ func newRaft(c *Config) *raft {
 		raftLog:                   raftlog,
 		maxMsgSize:                c.MaxSizePerMsg,
 		maxUncommittedSize:        c.MaxUncommittedEntriesSize,
-		prs:                       tracker.MakeProgressTracker(c.MaxInflightMsgs, 0), // TODO: set maxBytes
+		prs:                       tracker.MakeProgressTracker(c.MaxInflightMsgs, c.MaxInflightBytes),
 		electionTimeout:           c.ElectionTick,
 		heartbeatTimeout:          c.HeartbeatTick,
 		logger:                    c.Logger,
@@ -484,8 +499,7 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 
 	// Send the actual MsgApp otherwise, and update the progress accordingly.
 	next := pr.Next // save Next for later, as the progress update can change it
-	// TODO(pavelkalinnikov): set bytes to sum(Entries[].Size())
-	if err := pr.UpdateOnEntriesSend(len(ents), 0 /* bytes */, next); err != nil {
+	if err := pr.UpdateOnEntriesSend(len(ents), payloadsSize(ents), next); err != nil {
 		r.logger.Panicf("%x: %v", r.id, err)
 	}
 	r.send(pb.Message{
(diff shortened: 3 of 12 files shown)


pavelkalinnikov:
@tbg

raft/tracker/inflights.go
```diff
@@ -22,21 +22,26 @@ package tracker
...
 
-	// buffer contains the index of the last entry
-	// inside one message.
+	size     int    // the max number of inflight messages
+	maxBytes uint64 // the max total byte size of inflight messages
+
+	// buffer contains the index of the last entry inside one message.
 	buffer []uint64
+	// size contains the corresponding byte size of the message.
+	sizes []uint64
```
pavelkalinnikov, line 35: For allocation efficiency, this should better be paired with the index in a struct, and allocated in a single slice. Also, this would be less error-prone.

pavelkalinnikov: Done.

raft/raft.go
```diff
@@ -481,8 +484,12 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 			// optimistically increase the next when in StateReplicate
 			case tracker.StateReplicate:
 				last := m.Entries[n-1].Index
+				var bytes uint64
```
tbg, line 487: Just saw this:

https://github.com/etcd-io/etcd/blob/419ee8a9c6799084a78b6188f40fd120a4d316a5/raft/raft.go#L1302-L1305

So on each heartbeat response (200ms), we'll be throwing out the oldest inflight. We kind of have to do that (or we'll be stuck if all inflight messages in fact did get dropped), but it seems like it might circumvent the protection.

Ideally, when inflights is full, the leader would send an empty append instead of freeing up a slot in inflights. An empty append is ~zero sized, so we can afford to ignore it for the purpose of inflight tracking.

To implement this, presumaby here

https://github.com/etcd-io/etcd/blob/419ee8a9c6799084a78b6188f40fd120a4d316a5/raft/raft.go#L437-L439

we would instead send an empty append instead of returning with `false`.

Btw, tried to see by which mechanism the leader would try to catch up a node that's behind. Let's say n1 is the leader, and n2 is fully caught up, but all inflight messages to n3 were dropped. There is no more activity (just regular heartbeats). What prompts the leader to re-send to n3? I thought it would be on heartbeats or heartbeat responses, but I don't see it.

pavelkalinnikov: Sounds doable, will fix.

On re-sends, would the Transport implementation do any retries?

pavelkalinnikov: So these lines will send entries (on heartbeat responses):
https://github.com/etcd-io/etcd/blob/419ee8a9c6799084a78b6188f40fd120a4d316a5/raft/raft.go#L1306-L1308

But they "optimistically" update the progress `Next`: https://github.com/etcd-io/etcd/blob/419ee8a9c6799084a78b6188f40fd120a4d316a5/raft/raft.go#L484

But it's unclear whether `Next` ever comes back if the node continues to hold leadership.

Maybe the assumption is that there is something like TCP underneath, so if we are able to heartbeat then other messages including appends will slip through the stream.

pavelkalinnikov: > There is no more activity (just regular heartbeats). What prompts the leader to re-send to n3? I thought it would be on heartbeats or heartbeat responses, but I don't see it.

My hypothesis after more reading is:
- We keep sending `MsgApp` indefinitely.
- Eventually we receive `MsgAppResp` with a reject.
- This [throws](https://github.com/etcd-io/etcd/blob/419ee8a9c6799084a78b6188f40fd120a4d316a5/raft/raft.go#L1242) the progress back, and results in retries.

pavelkalinnikov: Fix in #14633, I think the described issue is pre-existing.

raft/raft.go
```diff
@@ -160,6 +160,9 @@ type Config struct {
 	// overflowing that sending buffer. TODO (xiangli): feedback to application to
 	// limit the proposal rate?
 	MaxInflightMsgs int
+	// MaxInflightBytes limits the number of in-flight bytes in append messages.
+	// Complements MaxInflightMsgs. Ignored if zero.
```
tbg, line 164: > This effectively bounds the bandwidth-delay product. Note that especially in high-latency deployments setting this too low can lead to a dramatic reduction in throughput. For example, with a peer that has a round-trip latency of 100ms to the leader and this setting is set to 1mb, there is a throughput limit of 10mb/s for this group. With a RTT of 400ms, this drops to 2.5mb/s.

tbg (commented):
Have only skimmed this now, I think this will be close to ready when the precursor PR is in. 👍🏽

raft/raft.go
pavelkalinnikov: Done. Is this what you refer as [Little's law](https://en.wikipedia.org/wiki/Little%27s_law)?

pavelkalinnikov:
@tbg I rebased this on top of #14633 which is approved and nearly merged, so this PR is ready for review too, just ignore the first 6 commits.

raft/raft.go
tbg: I think so, where the "system" is "being inflight". "Customers" arrive (and leave) at a rate of X mb/s, they spend `latency` time in the system, and the product of those two is the "number of customers in the store" (inflight bytes).

raft/confchange/confchange.go
```diff
@@ -265,7 +265,7 @@ func (c Changer) initProgress(cfg *tracker.Config, prs tracker.ProgressMap, id u
 		// making the first index the better choice).
 		Next:      c.LastIndex,
 		Match:     0,
-		Inflights: tracker.NewInflights(c.Tracker.MaxInflight),
+		Inflights: tracker.NewInflights(c.Tracker.MaxInflight, 0), // TODO: set maxBytes
```
tbg: Not suggesting to do this here (really, not worth your time), but in the future, a neat trick could be to have a

```
const todoMaxBytes = 0
```

and reference that instead, that way it's easier to track down the usages later. Makes more sense with bigger changes of course.

raft/tracker/inflights.go
```diff
@@ -14,6 +14,12 @@
 
 package tracker
 
+// inflight describes an in-flight message.
```
tbg, line 17: More precisely, an in-flight `MsgApp`. I think that bit of clarification can be helpful to readers less in the weeds than us.

raft/tracker/inflights.go
```diff
@@ -22,37 +28,40 @@ package tracker
...
-	// number of inflights in the buffer
-	count int
 
-	// the size of the buffer
-	size int
+	count int    // number of inflight messages in the buffer
+	bytes uint64 // number of inflight bytes
+
+	size     int    // the max number of inflight messages
+	maxBytes uint64 // the max total byte size of inflight messages
```
tbg, line 36: Point out that it's not a hard max - we might overshoot.  CRDB does it all the time as you know :-)
I would also guess the behavior here is to overshoot *every time* when we can. We'd never "not send" a message because that would put us over, right? Instead we make sure to overshoot only by one message. (I am guessing, since this is the beginning of my review).

raft/tracker/inflights.go
```diff
@@ -22,37 +28,40 @@ package tracker
...
-	buffer []uint64
+	// buffer is a ring buffer containing info about all in-flight messages.
+	buffer []inflight
 }
 
-// NewInflights sets up an Inflights that allows up to 'size' inflight messages.
-func NewInflights(size int) *Inflights {
+// NewInflights sets up an Inflights that allows up to 'size' inflight messages,
+// with the total size up to maxBytes. If maxBytes is 0 then there is no byte
+// size limit. The maxBytes limit is soft, a single message may overflow it.
```
tbg, line 44: Ah, you mention it here. Maybe that is ok. But you could clarify what you mean by "a single message will overflow it". The design of the interface here is that we will always end up overflowing by one message. There's no method on `Inflights` to check whether adding a message would lead to overflow. You add the message and check after.

raft/tracker/progress.go
```diff
@@ -136,13 +136,13 @@ func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
 
 // UpdateOnEntriesSend updates the progress on the given number of consecutive
 // entries being sent in a MsgApp, appended at and after the given log index.
-func (pr *Progress) UpdateOnEntriesSend(entries int, nextIndex uint64) error {
+func (pr *Progress) UpdateOnEntriesSend(entries int, bytes, nextIndex uint64) error {
```
tbg, line 139: In case this isn't done by the end of the commit chain, update the comment for the `bytes` parameter.

tbg (approved):

tbg:
cc @ahrtr you probably want to take a look as well.

raft/tracker/inflights.go
```diff
@@ -115,7 +128,7 @@ func (in *Inflights) FreeLE(to uint64) {
 
 // Full returns true if no more messages can be sent at the moment.
 func (in *Inflights) Full() bool {
-	return in.count == in.size
+	return in.count == in.size || (in.bytes >= in.maxBytes && in.maxBytes > 0)
```
ahrtr, line 131: Minor improvement: if `maxBytes` isn't set, then the right operand `in.bytes >= in.maxBytes` will not be evaluated at all. Since it's a new field, and defaults to 0, so it's disabled (0) by default. 

```suggestion
	return in.count == in.size || (in.maxBytes > 0 && in.bytes >= in.maxBytes)
```

ahrtr:
Overall looks good to me. 

Probably we should add a check: the `MaxInflightBytes` shouldn't be smaller than [MaxSizePerMsg](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/raft/raft.go#L148).

ahrtr:
Please also add two [changelog items](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md), one for this feature, and the other for https://github.com/etcd-io/etcd/pull/14633

raft/tracker/progress.go
pavelkalinnikov: Done.

raft/raft.go
pavelkalinnikov: Added a note on the law too.

pavelkalinnikov:
> Probably we should add a check: the MaxInflightBytes shouldn't be smaller than [MaxSizePerMsg](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/raft/raft.go#L148).

I'm on the fence here, since both are soft limits, and both are used independently.

pavelkalinnikov:
Maybe we should not overload `maxSize = 0` in places other than the config. We could reset `MaxInflightBytes` to `noLimit` in the validate func, similarly to how it's done for `MaxUncommittedEntriesSize`: https://github.com/etcd-io/etcd/blob/215b799a3207bf7d250db44b597c51bafb6630f8/raft/raft.go#L228-L230

tbg:
> Maybe we should not overload maxSize = 0 in places other than the config.

Sounds good to me as a pattern to follow, and since there's precedent we should be consistent.

> Probably we should add a check: the MaxInflightBytes shouldn't be smaller than [MaxSizePerMsg](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/raft/raft.go#L148).

@pavelkalinnikov while you're here mind giving MaxSizePerMsg a similar comment to the one we wrote for MaxInflightBytes (bandwidth-delay product etc), and make sure they cross-reference each other?

I think @ahrtr has a point.

https://github.com/etcd-io/etcd/blob/215b799a3207bf7d250db44b597c51bafb6630f8/raft/raft.go#L460-L462

It is very likely a misconfiguration to have `MaxSizePerMsg >> MaxInflightBytes`; I don't see why anyone would set that in practice. It is true that `MaxInflightBytes` can overshoot, but in an ideal world it would overshoot "minimally", i.e. by exactly one entry.

The code above could be improved: rather than pulling entries until the sum is >= `MaxSizePerMsg` and then shoving that into the tracker, more "optimal" code would be this (mod bugs maybe :-) )

```go
	var maxBytes uint64
	if pr.State == tracker.StateReplicate {
		maxBytes = min(r.maxMsgSize, pr.Inflights.BytesBudget())
	} else {
		// When follower is being probed, send only one entry.
		// TODO: should we send *no* entries? Or maybe the optimistic
		// strategy of sending maxMsgSize worth of entries is actually
		// better in practice? We've seen some *very divergent* raft logs
		// in CRDB incidents so I'm inclined to be conservative here and
		// send as little data as possible.
		maxBytes = 1
	}
	var ents []pb.Entry
	var erre error
	// In a throttled StateReplicate only send empty MsgApp, to ensure progress.
	// Otherwise, if we had a full Inflights and all inflight messages were in
	// fact dropped, replication to that follower would stall. Instead, an empty
	// MsgApp will eventually reach the follower (heartbeats responses prompt the
	// leader to send an append), allowing it to be acked or rejected, both of
	// which will clear out Inflights.
	if maxBytes > 0 {
		ents, erre = r.raftLog.entries(pr.Next, r.maxMsgSize)
	}

	if len(ents) == 0 && !sendIfEmpty {
		return false
	}
```

and at that point it should be clear that it makes no sense to set `maxMsgSize` to anything larger than `MaxInflightBytes`.

CHANGELOG/CHANGELOG-3.6.md
```diff
@@ -46,6 +46,10 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 - Package `wal` was moved to `storage/wal`
 - Package `datadir` was moved to `storage/datadir`
 
+### Package `raft`
+- Send empty `MsgApp` when entry in-flight limits are exceeded. See [pull/16633](https://github.com/etcd-io/etcd/pull/14633).
```
ahrtr, line 50:
```suggestion
- Send empty `MsgApp` when entry in-flight limits are exceeded. See [pull/14633](https://github.com/etcd-io/etcd/pull/14633).
```

pavelkalinnikov:
@tbg Added a couple of checks in the config, as suggested by you and @ahrtr.

@tbg I like this bit:
```golang
if pr.State == tracker.StateReplicate {
	maxBytes = min(r.maxMsgSize, pr.Inflights.BytesBudget())
} else {
```

The `maxBytes = 1` bit in the probing state would change the current behaviour, so I would defer it to another PR / thread of thinking, and for now only do:

```golang
maxBytes := r.maxMsgSize
if pr.State == tracker.StateReplicate {
	maxBytes = min(maxBytes, pr.Inflights.BytesBudget())
}
// ...
if maxBytes > 0 {
	ents, erre = r.raftLog.entries(pr.Next, maxBytes)
}
```

Although I think it would break the edge case of `MaxSizePerMsg == 0`. Currently it results in always sending one entry, but with the `maxBytes > 0` guard we would rule it out, and never send any entries. Slightly inclined to leave as is, but will poke at the `BytesBudget()` idea in general tomorrow.

tbg:
> Although I think it would break the edge case of MaxSizePerMsg == 0

Could we map `0 -> 1` when installing the config?

> Slightly inclined to leave as is, but will poke at the BytesBudget() idea in general tomorrow.

Sounds good, let's do a follow-up soon while we remember this piece of code.

ahrtr (approved):
LGTM

Thank yo @pavelkalinnikov

PR #14627, opened 2022-10-26
raft: support asynchronous storage writes

### LINKED ISSUE #12257: Proposal: support fully control fsync frequency in raft

nvanbenschoten:
Fixes #12257.

This change adds opt-in support to raft to **perform local storage writes asynchronously** from the raft state machine handling loop.

## Summary

A new `AsyncStorageWrites` configuration instructs the raft node to write to its local storage (raft log and state machine) using a request/response message passing interface instead of the default `Ready`/`Advance` function call interface. Local storage messages can be pipelined and processed asynchronously (with respect to `Ready` iteration), facilitating reduced interference between Raft proposals and increased batching of log appends and state machine application. As a result, use of asynchronous storage writes can reduce end-to-end commit latency and increase maximum throughput.

When `AsyncStorageWrites` is enabled, the `Ready.Message` slice will include new `MsgStorageAppend` and `MsgStorageApply` messages. The messages will target a `LocalAppendThread` and a `LocalApplyThread`, respectively. Messages to the same target must be reliably processed in order. In other words, they can't be dropped (like messages over the network) and those targeted at the same thread can't be reordered. Messages to different targets can be processed in any order.

`MsgStorageAppend` carries Raft log entries to append, election votes to persist, and snapshots to apply. All writes performed in response to a `MsgStorageAppend` are expected to be durable. The message assumes the role of the Entries, HardState, and Snapshot fields in Ready.

`MsgStorageApply` carries committed entries to apply. The message assumes the role of the CommittedEntries field in Ready.

Local messages each carry one or more response messages which should be delivered after the corresponding storage write has been completed. These responses may target the same node or may target other nodes. The storage threads are not responsible for understanding the response messages, only for delivering them to the correct target after performing the storage write.

## Design Considerations

- There must be no regression for existing users that do not enable `AsyncStorageWrites`. For instance, CommittedEntries must not wait on unstable entries to be stabilized in cases where a follower is given committed entries in a MsgApp.
- Asynchronous storage work should use a message passing interface, like the rest of this library.
- The Raft leader and followers should behave symmetrically. Both should be able to use asynchronous storage writes for log appends and entry application.
- The LocalAppendThread on a follower should be able to send MsgAppResp messages directly to the leader without passing back through the raft state machine handling loop.
- The `unstable` log should remain true to its name. It should hold entries until they are stable and should not rely on an intermediate reliable cache.
- Pseudo-targets should be assigned to messages that target the local storage systems to denote required ordering guarantees.
- Code should be maximally unified across `AsyncStorageWrites=false` and `AsyncStorageWrites=true`. `AsyncStorageWrites=false` should be a special case of `AsyncStorageWrites=true` where the library hides the possibility of asynchrony.
- It should be possible to apply snapshots asynchronously, even though a snapshot touches both the Raft log state and the state machine. The library should make this easy for users to handle by delaying all committed entries until after the snapshot has applied, so snapshot application can be handled by 1) flushing the apply thread, 2) sending the `MsgStorageAppend` that contains a snapshot to the `LocalAppendThread` to be applied.

## Usage

When asynchronous storage writes is enabled, the responsibility of code using the library is different from what is presented in [raft/doc.go](https://github.com/etcd-io/etcd/blob/main/raft/doc.go) (which has been updated to include a section about async storage writes). Users still read from the `Node.Ready()` channel. However, they process the updates it contains in a different manner. Users no longer consult the HardState, Entries, and Snapshot fields (steps 1 and 3 in doc.go). They also no longer call `Node.Advance()` to indicate that they have processed all entries in the Ready (step 4 in doc.go). Instead, all local storage operations are also communicated through messages present in the Ready.Message slice.

The local storage messages come in two flavors. The first flavor is log append messages, which target a LocalAppendThread and carry Entries, HardState, and a Snapshot. The second flavor is entry application messages, which target a LocalApplyThread and carry CommittedEntries. Messages to the same target must be reliably processed in order. Messages to different targets can be processed in any order. Each local storage message carries a slice of response messages that must delivered after the corresponding storage write has been completed.

With Asynchronous Storage Writes enabled, the total state machine handling loop will look something like this:

```go
for {
	select {
	case <-s.Ticker:
		n.Tick()
	case rd := <-s.Node.Ready():
		for _, m := range rd.Messages {
			switch m.To {
			case raft.LocalAppendThread:
				toAppend <- m
			case raft.LocalApplyThread:
				toApply <-m
			default:
				sendOverNetwork(m)
			}
		}
	case <-s.done:
		return
	}
}
```

Usage of Asynchronous Storage Writes will typically also contain a pair of storage handler threads, one for log writes (append) and one for entry application to the local state machine (apply). Those will look something like:

```go
// append thread
go func() {
	for {
		select {
		case m := <-toAppend:
			saveToStorage(m.State, m.Entries, m.Snapshot)
			send(m.Responses)
		case <-s.done:
			return
		}
	}
}

// apply thread
go func() {
	for {
		select {
		case m := <-toApply:
			for _, entry := range m.CommittedEntries {
				process(entry)
				if entry.Type == raftpb.EntryConfChange {
					var cc raftpb.ConfChange
					cc.Unmarshal(entry.Data)
					s.Node.ApplyConfChange(cc)
				}
			}
			send(m.Responses)
		case <-s.done:
			return
		}
	}
}
```

## Compatibility

The library remains backwards compatible with existing users and the change does not introduce any breaking changes. Users that do not set `AsyncStorageWrites` to true in the `Config` struct will not notice a difference with this change. This is despite the fact that the existing "synchronous storage writes" interface was adapted to share a majority of the same code. For instance, `Node.Advance` has been adapted to transparently acknowledge an asynchronous log append attempt and an asynchronous state machine application attempt, internally using the same message passing mechanism introduced in this change.

The change has no cross-version compatibility concerns. All changes are local to a process and nodes using asynchronous storage writes appear to behave no differently from the outside. Clusters are free to mix nodes running with and without asynchronous storage writes.

## Performance

The bulk of the performance evaluation of this functionality thus far has been done with [rafttoy](https://github.com/nvanbenschoten/rafttoy), a benchmarking harness developed to experiment with Raft proposal pipeline optimization. The harness can be used to run single-node benchmarks or multi-node benchmarks. It supports plugable raft logs, storage engines, network transports, and pipeline implementations.

To evaluate this change, we fixed the raft log (`etcd/wal`), storage engine (`pebble`), and network transport (`grpc`). We then built (https://github.com/nvanbenschoten/rafttoy/pull/3) a pipeline implementation on top of the new asynchronous storage writes functionality and compared it against two other pipeline implementations.

The three pipeline implementations we compared were:
- **basic** (**P1**): baseline stock raft usage, similar to the code in `doc.go`
- **parallel append + early ack** (**P2**): CockroachDB's current pipeline, which includes two significant variations to the basic pipeline. The first is that it sends MsgApp messages to followers before writing to local Raft log (see [commit](https://github.com/cockroachdb/cockroach/commit/b67eb696f6a064d6a3a18517c1ed5525faa931e8) for explanation), allowing log appends to occur in parallel across replicas. The second is that it acknowledges committed log entries before applying them (see [commit](https://github.com/cockroachdb/cockroach/commit/87aaea70d68903ad04ecae1adf2910333aa07a78) for explanation).
- **async append + async apply + early ack** (**P3**): A pipelining using asynchronous storage writes with a separate append thread and a separate apply thread. Also uses the same early acknowledgement optimization from above to ack committed entries before handing them to the apply thread.

All testing was performed on a 3 node AWS cluster of m5.4xlarge instances with gp3 EBS volumes (16000 IOPS, 1GB/s throughput). The testing used an open-loop workload to increase the rate of new raft proposals until a saturation point was reached.

Throughput vs latency of Raft proposal pipeline implementations

The comparison demonstrates two different benefits of asynchronous storage writes.

The first is that it reduces end-to-end latency of proposals by 20-25%. For instance, when serving 16MB/s of write traffic, **P1**'s average latency was 13.2ms, **P2**'s average latency was 7.3ms, and **P3**'s average latency was 5.2ms. This is a reduction in average latency of **29%** from the optimized pipeline that does not use asynchronous storage writes. This matches the expectations outlined in https://github.com/cockroachdb/cockroach/issues/17500.

The second is that it increases the maximum throughput at saturation. In this test, **P1** and **P2** topped out at 30MB/s, while **P3** could push up to 52MB/s, an increase in maximum throughput of **73%**. This is because asynchronous storage writes can improve batching for both log appends and log application. In this experiment, we saw the average append batch size under saturation increase from 928 to 1542, which is a similar ratio to the increase in peak throughput. We see a similar difference for apply batch sizes.

There is more benchmarking to do. For instance, we'll need to thoroughly verify that this change does not negatively impact the performance of users of this library that do not use asynchronous storage writes.

cc. @tbg @bdarnell


ahrtr:
Just a reminder, big PR is hard to review. Please consider to deliver a design doc firstly, and breakdown the PR into small ones as possible as you can.

nvanbenschoten:
Yes, I completely agree. This is just a draft PR that includes a sequence of preparatory refactors. I intend to pull those out into separate PRs once the high-level architecture presented here reaches a consensus (no pun intended). The core of this change is really just the last three commits in the PR (two of which are small).

I'll be talking with @tbg and @bdarnell about the approach over the next few days. I'd also be happy to talk through it with you @ahrtr at your convenience, as etcd may want to adopt this interaction model as well.

nvanbenschoten:
@serathius thanks for raising the point about testing. I think there are two aspects to the risks posed by this change.

----

The **first aspect** is the risk to users who do not enable the new `AsyncStorageWrites` configuration. The risk to these users is not negligible, but it should be small and contained. This is because there is almost no visible behavior change for those users. One way to see this is that very few existing tests inside the `raft` package needed to change in response to this change, and no tests outside of the `raft` package needed to change. In fact, the only externally visible change for these users comes from https://github.com/etcd-io/etcd/pull/14627/commits/c05e48f13c76a3c9bac7a9b530726627594b8412, which parallels https://github.com/etcd-io/etcd/pull/14413 and keeps the rest of the async storage writes change focused. I think it makes sense to pull that preliminary commit into a separate PR and talk through whether we need to add more testing than the commit already adds to gain more confidence in the change.

There is also a risk that some of the internal refactors made to unify logic between the `AsyncStorageWrites: true` and `AsyncStorageWrites: false` configs are faulty. The fact that we don't see any impact on tests outside of the `raft` package gives us some confidence that no such bugs were introduced. Something else we can do to stress the change is upgrade CockroachDB to use this code (with `AsyncStorageWrites: false`) and stress its layered test suite. Together, I think that gives us good confidence in the correctness of these reactors.

There's also the performance aspect of these reactors. Are we regressing performance for users where `AsyncStorageWrites: false`? I'll need to run some benchmarks to confirm that we're not. 

----

The **second aspect** is that there are risks of bugs that would specifically affect users that do enable the new `AsyncStorageWrites` configuration. There's a higher likelihood of this, as async storage writes introduces new states and requires additional progress tracking.

The design of this change fought back against these new states resulting in undertested code by unifying the majority of the logic between the `AsyncStorageWrites: true` and `AsyncStorageWrites: false` configs. This means that the new progress tracking is exercised with `AsyncStorageWrites: false`, so it's not dead code when async storage writes are not used. Still, there are interactions that can only occur with async storage writes due to different orderings between remote responses, log write acknowledgment, and log application acknowledgment.

I think a form of randomized testing could help give us more confidence that we handle these new states correctly, regardless of the ordering of responses. I'll look into this.

raft/raft.go
```diff
@@ -423,10 +475,59 @@ func (r *raft) send(m pb.Message) {
 			m.Term = r.Term
 		}
 	}
-	if m.To == r.id {
-		r.logger.Panicf("message should not be self-addressed when sending %s", m.Type)
+	if m.Type == pb.MsgAppResp || m.Type == pb.MsgVoteResp || m.Type == pb.MsgPreVoteResp {
```
bdarnell, line 478: Pre-votes are not persisted (right? it's been a long time), so I don't think MsgPreVoteResp needs to be here.

raft/raft.go
```diff
@@ -423,10 +475,59 @@ func (r *raft) send(m pb.Message) {
...
+		// another peer. Or it may reject a vote from one peer because it has
+		// unstable log entries that indicate that the peer is behind on its
+		// log. In these cases, it is likely safe to send out the rejection
+		// response immediately without compromising safety in the presence of a
+		// server restart. However, because these rejections are rare and
+		// because the safety of such behavior has not been formally verified,
+		// we err on the side of safety and omit a `&& !m.Reject` condition
+		// above.
+		r.msgsAfterAppend = append(r.msgsAfterAppend, m)
+	} else {
```
bdarnell, line 525: This is a reordering of messages relative to the status quo: messages other than voting responses now get to cut in line and go ahead of (earlier) voting responses. That should be fine for correctness (raft is already theoretically robust against reordered messages), but preserving order of messages has proven important for performance in the past. 

This was mainly about ordering within a message type, although I think there may have been interactions between appends and snapshots (notable since only appends are affected by this change). I can't think of any problems here but just wanted to raise it as a possible concern.

raft/raft.go
```diff
@@ -551,58 +652,35 @@ func (r *raft) bcastHeartbeatWithCtx(ctx []byte) {
...
-			//  r.prs.Progress[r.id].MaybeUpdate(e.Index)
-			//  if r.maybeCommit() {
-			//  	r.bcastAppend()
-			//  }
-			_ = r.Step(pb.Message{From: r.id, Type: pb.MsgAppResp, Index: e.Index})
+		// There's no way in which this proposal should be able to be rejected.
+		if !r.appendEntry(ent) {
+			panic("refused un-refusable auto-leaving ConfChangeV2")
 		}
-		// NB: it's important for performance that this call happens after
```
bdarnell, line 589: I don't see a similar comment in the new code. Is this "important" optimization still occurring?

raft/node.go
```diff
@@ -159,6 +168,9 @@ type Node interface {
 	// commands. For example. when the last Ready contains a snapshot, the application might take
 	// a long time to apply the snapshot data. To continue receiving Ready without blocking raft
 	// progress, it can call Advance before finishing applying the last ready.
+	//
+	// NOTE: Advance must not be called when using AsyncStorageWrites. Response messages from the
```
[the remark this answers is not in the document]
nvanbenschoten: We could do this if we wanted to. The low-level `raft` object actually is entirely async now. It never consults the `asyncStorageWrites` flag. Instead, this flag is only consulted in `node.go` and `rawnode.go` when constructing, accepting, and advancing Ready structs. The question we would need to answer is whether a separate async variant of `RawNode` and `Node` would be worth the code duplication.

Even without a split interface, I think it would be worthwhile to lift this configuration from `raft` to `RawNode` to make its role clearer.

On a related note, I think we should move `newReady` from `node.go` to `rawnode.go` so that async storage writes are entirely handled in `rawnode.go`.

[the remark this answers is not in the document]
nvanbenschoten: > Even without a split interface, I think it would be worthwhile to lift this configuration from raft to RawNode to make its role clearer.

Done.

> On a related note, I think we should move newReady from node.go to rawnode.go so that async storage writes are entirely handled in rawnode.go.

I left a TODO about this that I'll address before merging.

raft/node.go
```diff
@@ -578,6 +591,30 @@ func newReady(r *raft, prevSoftSt *SoftState, prevHardSt pb.HardState) Ready {
 		rd.ReadStates = r.readStates
 	}
 	rd.MustSync = MustSync(r.hardState(), prevHardSt, len(rd.Entries))
+
+	if r.asyncStorageWrites {
```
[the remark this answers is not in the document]
nvanbenschoten: This all belongs in `RawNode`, see the previous comment thread. `newReady` is actually only called from `RawNode.readyWithoutAccept`. So there's no duplication here, it's just related logic that's split across two files. I'll fix this before any of this lands.

raft/node.go
```diff
@@ -591,3 +628,175 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+// after the committed entries in the current Ready (along with those in all
+// prior Ready structs) have been applied to the local state machine.
+func newStorageApplyRespMsg(r *raft, committedEnts []pb.Entry) pb.Message {
+	last := committedEnts[len(committedEnts)-1].Index
+	size := r.getUncommittedSize(committedEnts)
+	return pb.Message{
+		Type:  pb.MsgStorageApplyResp,
+		To:    r.id,
+		From:  LocalApplyThread,
+		Term:  None, // committed entries don't apply under a specific term
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/raft.go
```diff
@@ -31,9 +31,18 @@ import (
...
-// None is a placeholder node ID used when there is no leader.
-const None uint64 = 0
-const noLimit = math.MaxUint64
+const (
+	// None is a placeholder node ID used when there is no leader.
+	None uint64 = 0
+	// LocalAppendThread is a reference to a local thread that saves unstable
+	// log entries and snapshots to stable storage. The identifier is used as a
+	// target for MsgStorageAppend messages when AsyncStorageWrites is enabled.
+	LocalAppendThread uint64 = math.MaxUint64
```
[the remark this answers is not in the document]
nvanbenschoten: There is a 1:1 relationship between the message type and the thread pseudo-node, but I have come to like the explicitness of the thread targets. They accurately represent the meaning of the messages, permit routing solely based on the `Message.To` field, and they make it easier to talk about ordering requirements to a given target without ordering requirements across targets.

At the same time, your concern about backwards compat is valid. We could get around the issue by only preventing the use of these IDs when AsyncStorageWrites is enabled.

raft/raft.go
nvanbenschoten: Pre-votes aren't persisted, but they are based on the state of the node's log (see `r.raftLog.isUpToDate`). By including MsgPreVoteResp here, we ensure that the response is derived from stable state.

Given pre-vote's optional role in the Raft algorithm and the limited amount of state that nodes record from MsgPreVoteResp's they receive from others, I can't think of a case where responding based on unstable state would cause an issue. However, I still think it's best to keep it here unless we find that it's important not to and then go through the effort of proving its correctness.

That's similar to why I avoided adding an optimization for message rejections (see the last paragraph in this comment, which starts with `Rejected responses`).

raft/log.go
```diff
@@ -190,8 +197,9 @@ func (l *raftLog) nextCommittedEnts() (ents []pb.Entry) {
 		// See comment in hasNextCommittedEnts.
 		return nil
 	}
-	if l.committed > l.applied {
-		lo, hi := l.applied+1, l.committed+1 // [lo, hi)
+	if l.committed > l.applying {
+		lo, hi := l.applying+1, l.committed+1 // [lo, hi)
+		// TODO: handle pagination correctly.
```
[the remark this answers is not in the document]
nvanbenschoten: Yes, this is a TODO that I'll need to address before this sequence of commits can be merged.

"Correctly" means that we can now have multiple `Ready`s worth of committed entries outstanding at a time, so the memory-limiting per-Ready pagination of up to `maxNextEntsSize` committed entries will not work correctly. We need that limit to apply to all outstanding committed entries across multiple Ready structs.

raft/raft.go
nvanbenschoten: This is a good point. As you mention, we now introduce certain re-ordering across message types, but not within a message type. I haven't been able to construct a case where delaying any of the three `msgsAfterAppend` message types causes issues.

>  I think there may have been interactions between appends and snapshots

Was this between MsgApp and MsgSnap, or MsgAppResp and MsgSnap? It must have been the former, as MsgApp and MsgSnap are both sent from the leader, while MsgAppResp is sent from followers.

Also, recall that MsgSnap also uses a MsgAppResp to respond to the leader, so these responses will remain ordered.

raft/raft.go
nvanbenschoten: The optimization is still important and was still in effect, but the comment was lost. Revived and improved.

raft/node.go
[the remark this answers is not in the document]
nvanbenschoten: > I left a TODO about this that I'll address before merging.

Done.

xiang90:
@nvanbenschoten 

It seems that even with this change, we are far away from saturating the I/O and Compute resources. Do you get a chance to see why? I am asking since I want to make sure we are trying to get both the short-term and long-term direction right :)

raft/rafttest/interaction_env_handler_add_nodes.go
```diff
@@ -43,10 +44,12 @@ func (env *InteractionEnv) handleAddNodes(t *testing.T, d datadriven.TestData) e
 				arg.Scan(t, i, &snap.Metadata.Index)
 			case "content":
 				arg.Scan(t, i, &snap.Data)
+			case "async-storage-writes":
+				arg.Scan(t, i, &asyncStorageWrites)
```
pavelkalinnikov, line 48: Consider reading into a `Config` directly, like in ffee6cc4bd03f1ab8c7f328ee6b11a13b1c2ed49.

raft/util.go
```diff
@@ -70,6 +76,10 @@ func IsResponseMsg(msgt pb.MessageType) bool {
 	return isMsgInArray(msgt, isResponseMsg[:])
 }
 
+func IsLocalMsgTarget(id uint64) bool {
+	return id == LocalAppendThread || id == LocalApplyThread
```
pavelkalinnikov, line 80: Is it necessary to introduce "threads" at the API level in this design, is it not deductible from the message type? I.e. I'm thinking just a `Local` target with no indication of threads. The receiver would then determine the "thread" based on the message metadata.

Also, currently / before this change, does `Ready` ever contain messages with `To == <this-node>`? If not, could we even avoid introducing the special target value(s), and instead send regular self-messages.

pavelkalinnikov: Is it true that the requirement for the local `MsgStorageAppend` and `MsgStorageApply` messages is:
- When a `Resp` message is sent, we have the guarantee that the data in the request is persisted.
- App sends `MsgStorageAppendResp` messages back to raft in the same order `MsgStorageAppend` were requested.
- Ditto for `MsgStorageApply[Resp]` messages.

As discussed in person, this does not necessarily imply a requirement for running these requests sequentially at the app/storage side, which the "thread" semantics alludes to. For instance, storage could receive multiple appends (entries 11-20, and 21-30, and 31-40), parallelize them (if the storage engine can do so efficiently, a simple example is it has multiple disks and round robins entries). It may actually happen that latest entries are stored earlier.

Even for "apply" messages, where seemingly sequential semantics are necessary, some parallelization is possible, e.g. when batches are independent, most notably when the entire store is a CRDT.

As discussed, we probably don't want to deal with out-of-order replies at the raft side. One of the reasons: when we have an ack on a later batch, acks for earlier batches may never show up, so we would have to also add retry logic to raft. If it's possible to do out-of-order stuff, the storage implementation should do the necessary sync/reconciliation, and feed replies in the original order.

pavelkalinnikov: Similarly to wanting to separate threads, I can also imagine sometimes wanting to **not** do so. Say, the storage is built so that it's more efficient to append entries and apply some earlier entries in the same batch/thread.

A silly example: maybe the "storage" is a single KV variable (or a few of them), and instead of separating the log and KV storages the implementation just piggybacks the applied KV data in every log entry. So, the handler would squash `MsgStorageAppend` and `MsgStorageApply` into a single sequence of entries and flush it; instead of flushing entries and the variable asynchronously / separately.

nvanbenschoten:
> It seems that even with this change, we are far away from saturating the I/O and Compute resources. Do you get a chance to see why? I am asking since I want to make sure we are trying to get both the short-term and long-term direction right :)

@xiang90 I re-ran the rafttoy benchmark setup that I posted above to determine why and where we were hitting the limiting bottleneck. As you called out, ~50MB/s of Raft proposals shouldn't be saturating compute, network, or I/O on this cluster[^1]. It turns out that we were approaching disk write I/O saturation. This was due to the high degree of write amplification in [the LSM](https://github.com/nvanbenschoten/rafttoy/blob/master/storage/engine/pebble.go) that the test was using as the replicated state machine's storage engine. In other words, LSM compactions of applied entries were responsible for most disk write I/O.

I constructed a new setup using an [in-memory](https://github.com/nvanbenschoten/rafttoy/blob/master/storage/engine/mem.go) replicated state machine storage engine and [etcd's WAL](https://github.com/nvanbenschoten/rafttoy/blob/master/storage/wal/etcd.go) for the Raft log. This meant that all disk I/O came from Raft log manipulation, so there would be no write amplification in the test. I also increased the entry size to 4KB so that I could try to saturate the hardware with < 1M proposals/s.

Throughput vs  latency of Raft proposal pipeline implementations

With this setup, we can saturate the hardware. Asynchronous storage writes allow nodes to write continuously to their Raft log in a tight loop, with minimal interference between entry log writes and entries at different stages of the Raft pipeline. While the basic pipeline maxes out at 340 MB/s worth of proposals, the async storage writes pipeline maxes out at 583 MB/s.

However, we don't saturate compute or disk I/O. Instead, with async storage writes, we can saturate the 10 Gigabit network between the Raft leader and its two followers (`10Gb/s / 8 = 1.25GB/s = 625MB/s per follower`). Experimentally, 600 MB/s is about as much as I can push from the leader VM to both follower VMs concurrently using `iperf3`.

[^1]: a 3 node AWS cluster of m5.4xlarge instances with gp3 EBS volumes (16000 IOPS, 1GB/s throughput)

xiang90:
@nvanbenschoten 

Thanks for the detailed analysis. That makes a lot of sense!

I was surprised that we could only do 30MB/s before the optimization as I did some perf analysis in 2017 with a very different result. I thought there might be something significant changed.

xiang90:
@nvanbenschoten 

Awesome work, by the way! HUGE improvements.

raft/rafttest/interaction_env_handler_add_nodes.go
nvanbenschoten: Nice, I like this. Once https://github.com/etcd-io/etcd/pull/14633 lands, I'll rebase on top of it and make this change.

raft/util.go
nvanbenschoten: Thanks for writing up these details from our discussion! 

> Is it true that the requirements for handling the local MsgStorageAppend and MsgStorageApply messages are:

Yes, the requirements that you laid out at all correct. The documentation that outlines these requirements is on `Config.AsyncStorageWrites`, and then again in `doc.go`. Please let me know if there are details missing there.

> Is it necessary to introduce "threads" at the API level in this design, is it not deductible from the message type? I.e. I'm thinking just a Local target with no indication of threads. The receiver would then determine the "thread" based on the message metadata.

I think where we landed during our discussion was that it was useful to assign a target to these messages for two reasons:
1. it allows a routing component to route to remote nodes, the local node, or local storage processes by looking solely at a message's `To` field. It avoids the need for this router to deconstruct the contents of messages in order to route them.
2. splitting the local storage targets into an "appender" and an "applier" makes it easier to discuss the ordering requirements for all messages to the same target, but absence of ordering requirements between messages to different targets.

There was also some discussion about whether the use of the term "thread" was too specific and restrictive to users of the library. This library leaves the storage and network implementation to the user, so there's a reasonable argument that the details of threading should also be left out of the API. My intention here was for "thread" to be thought of abstractly, more closely to an arbitrary "thread of execution" than a specific "pthread". In fact, if we wanted to refer to a specific threading implementation, we would probably want to call these "goroutines". There are parallels here with "storage" and "network", which the library evokes for clarity and intuition, but otherwise leaves the specific details to users.

> If not, could we even avoid introducing the special target value(s), and instead send regular self-messages?

We discussed the idea of an actor model and message passing briefly last week. In that analogy, it becomes important to differentiate a node ID from local storage actors. This is because the node ID is a reference to Raft state machine (this library). It would be ambiguous it the node ID could refer to multiple actors.

> As discussed in person, this does not necessarily imply a requirement for running these requests sequentially at the app/storage side, which the "thread" semantics alludes to.

> Similarly to wanting to separate threads, I can also imagine sometimes wanting to not do so.

These are helpful examples of how this library might be used to think through. Our goal when making this change should be to ensure that users of the library are provided with enough information to adhere to the requirements of the library (e.g. `MsgStorageAppend` msgs must be acked in order), but otherwise, given the flexibility to adhere to them however they want. For instance, the library refers to an "appender" and an "applier", but users are welcome to have the same goroutine play this role. Or to play this role directly in the state machine handling loop.

The library should then strive to place as few requirements on its use as possible, except where there is a meaningful benefit to restricting its use in a way that is not overly burdensome on users. The example we discussed was the reliable delivery of local storage messages. We could certainly allow for these messages to be dropped, but supporting this would add complexity and cost to the library. Since in-process communication is usually reliable (unlike across a network), it's hard to imagine cases where this restriction is particularly imposing to users of the library, so this seems like the right trade-off to make.

raft/node.go
```diff
@@ -562,7 +575,7 @@ func (n *node) ReadIndex(ctx context.Context, rctx []byte) error {
 func newReady(r *raft, prevSoftSt *SoftState, prevHardSt pb.HardState) Ready {
 	rd := Ready{
 		Entries:          r.raftLog.nextUnstableEnts(),
-		CommittedEntries: r.raftLog.nextCommittedEnts(true /* allowUnstable */),
```
tbg: Ok, disregard my earlier comment on this then, I understand now that `true` was the right setting for the status quo (no async writes). I didn't realize (but now that I actually think about it remember) that we do emit unstable entries for application. I just spent time on that in the context of single-voter nodes.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+		// (index, term) of the last entry in rd.Entries. If rd.Entries is not empty,
+		// these will be the same. However, if rd.Entries is empty, we still want to
+		// attest that this (index, term) is correct at the current term, in case the
+		// MsgStorageAppend that contained the last entry in the unstable slice carried
+		// an earlier term and was dropped.
+		// TODO(nvanbenschoten): test this behavior in a data-driven test.
+		m.Index = r.raftLog.lastIndex()
+		m.LogTerm = r.raftLog.lastTerm()
+	}
+	if !IsEmptySnap(rd.Snapshot) {
```
tbg: You had an earlier commit that made sure that we never emit a snapshot together with "anything else", it might be worth repeating that here.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+// storage.
+func newStorageAppendRespMsg(r *raft, rd Ready) pb.Message {
+	m := pb.Message{
+		Type: pb.MsgStorageAppendResp,
+		To:   r.id,
+		From: LocalAppendThread,
+		// Dropped after term change, see below.
+		Term: r.Term,
+	}
+	if r.raftLog.hasNextOrInProgressUnstableEnts() {
```
tbg: Haven't reviewed this part yet, will need more time than I have right now.

TODO(tbg): come back.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+func newStorageApplyMsg(r *raft, rd Ready) pb.Message {
+	ents := rd.CommittedEntries
+	last := ents[len(ents)-1].Index
+	return pb.Message{
+		Type:    pb.MsgStorageApply,
+		To:      LocalApplyThread,
+		From:    r.id,
+		Term:    0, // committed entries don't apply under a specific term
+		Entries: ents,
+		Index:   last,
```
tbg: Mentioned elsewhere, but right now `Message.Index` has no comment which is no bueno. Leaves you guessing whether it's the first or last index or maybe even a one-off of the first.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+		},
+	}
+}
+
+// newStorageApplyRespMsg creates the message that should be returned to node
+// after the committed entries in the current Ready (along with those in all
+// prior Ready structs) have been applied to the local state machine.
+func newStorageApplyRespMsg(r *raft, committedEnts []pb.Entry) pb.Message {
+	last := committedEnts[len(committedEnts)-1].Index
+	size := r.getUncommittedSize(committedEnts)
```
tbg: `getUncommittedSize` had me confused for a minute, since it has this fast-path. Should we rename to `maybeGetUncommittedSize` to make it clearer that it returns 0 sometimes when we don't care to track?

raft/node.go
```diff
@@ -591,3 +628,175 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+	return pb.Message{
+		Type:  pb.MsgStorageApplyResp,
+		To:    r.id,
+		From:  LocalApplyThread,
+		Term:  None, // committed entries don't apply under a specific term
+		Index: last,
+		// NOTE: we abuse the LogTerm field to store the aggregate entry size so
```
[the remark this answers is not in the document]
tbg: I think Nathan and I chatted about this the other day and concluded that we could track this state in a slice in raft. It is sort of sad, but between the current abuse, adding a new field just for this purpose, or tracking the inflight entries inside of raft, my pick would be the tracking, though the slice for it could get pretty long if entries are tiny. Second preference the explicit field, third preference keep as is. It's just not clean and it's hard enough to keep track of what's in a `Message` already.

raft/raft.go
```diff
@@ -140,6 +151,38 @@ type Config struct {
...
+	// When true, the Ready.Message slice will include MsgStorageAppend and
+	// MsgStorageApply messages. The messages will target a LocalAppendThread
+	// and a LocalApplyThread, respectively. Messages to the same target must be
+	// reliably processed in order. In other words, they can't be dropped (like
+	// messages over the network) and those targeted at the same thread can't be
+	// reordered. Messages to different targets can be processed in any order.
+	//
+	// MsgStorageAppend carries Raft log entries to append, election votes to
+	// persist, and snapshots to apply. All writes performed in response to a
+	// MsgStorageAppend are expected to be durable. The message assumes the role
```
tbg, line 172: Isn't there the case in which all we're doing is announcing a new commit index, which "generally" doesn't need to be durable[^1]

[^1]: at least not in vanilla raft, need to page in the subtle semantics with our apply-time cfg changes again..

raft/raft.go
```diff
@@ -140,6 +151,38 @@ type Config struct {
...
+	// of the Entries, HardState, and Snapshot fields in Ready.
+	//
+	// MsgStorageApply carries committed entries to apply. The message assumes
```
tbg, line 175: Mention that these don't have to be durable.

raft/raft.go
```diff
@@ -261,15 +307,19 @@ type raft struct {
...
+	// Messages in this list must target other nodes.
 	msgs []pb.Message
-
-	// voteSelfOnAdvance is a marker that the local raft node should vote for
-	// itself upon its next call to advance. This is not meant to be the final
-	// approach towards handling self-votes, but it's a useful intermediate
-	// point to get all tests working and to write some additional tests that
-	// demonstrate possible race conditions when self-voting is asynchronous.
-	// This is replaced in a later commit.
-	voteSelfOnAdvance pb.Message
```
tbg: 👍🏽 disregard my earlier comment about this field then.

raft/raft.go
```diff
@@ -261,15 +307,19 @@ type raft struct {
...
+	// msgsAfterAppend contains the list of messages that should be sent after
+	// the accumulated unstable state (e.g. term, vote, []entry, and snapshot)
+	// has been persisted to durable storage. This includes waiting for any
+	// unstable state that is already in the process of being persisted (i.e.
+	// has already been handed out in a prior Ready struct) to complete.
+	//
+	// Messages in this list may target other nodes or may target this node.
```
tbg, line 321: may target other nodes (MsgAppResp, MsgVoteResp) or may target this node (<fill in examples; MsgAppResp too?>, MsgStorageAppendResp)

raft/raft.go
```diff
@@ -423,10 +475,59 @@ func (r *raft) send(m pb.Message) {
 			m.Term = r.Term
 		}
 	}
-	if m.To == r.id {
-		r.logger.Panicf("message should not be self-addressed when sending %s", m.Type)
+	if m.Type == pb.MsgAppResp || m.Type == pb.MsgVoteResp || m.Type == pb.MsgPreVoteResp {
+		// If async storage writes are enabled, messages added to the msgs slice
```
tbg, line 477: TODO(tbg): give this a close read

raft/util_test.go
```diff
@@ -104,3 +104,37 @@ func TestIsLocalMsg(t *testing.T) {
 		}
 	}
 }
+
+func TestIsResponseMsg(t *testing.T) {
```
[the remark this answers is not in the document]
nvanbenschoten: > Could you also add a comment at the bottom of the MessageType enum with a reminder to update is{Local,Response}Msg?

Done in https://github.com/etcd-io/etcd/pull/14723.

> Should we also unit test IsLocalMsg?

This already existed above. See `TestIsLocalMsg`.

raft/raft.go
```diff
@@ -796,7 +869,7 @@ func (r *raft) becomeLeader() {
 	// uncommitted log quota. This is because we want to preserve the
 	// behavior of allowing one entry larger than quota if the current
 	// usage is zero.
-	r.reduceUncommittedSize([]pb.Entry{emptyEnt})
+	r.uncommittedSize = 0
```
tbg, line 870: There's something fishy here. `getUncommittedSize` returns `0` when `r.uncommittedSize == 0`. So aren't we accidentally disabling the tracking here, since we won't ever add to `r.uncommittedSize` again? If so, let's make sure this is also exercised in a test.

Even if the above problem isn't actually occurring for some reason I don't understand, isn't resetting here open up the possibility of each leader amassing (say) 10mb of uncommitted log, losing leadership to a second node which adds another 10mb of uncommitted log, loses leadership back, adds another 10mb, etc, for essentially unbounded log growth? This would need a replication factor of 5 (if the log is to be "truly" uncomitted) but still.

raft/raft.go
```diff
@@ -945,6 +1023,21 @@ func (r *raft) Step(m pb.Message) error {
...
+		} else if m.Type == pb.MsgStorageAppendResp {
+			if m.Index != 0 {
+				// Don't consider the appended log entries to be stable because
+				// they may have been overwritten in the unstable log during a
+				// later term. See the comment in newStorageAppendResp for more
+				// about this race.
+				r.logger.Infof("%x [term: %d] ignored a %s message with lower term [term: %d]",
+					r.id, r.Term, m.Type, m.Term)
+			}
+			if m.Snapshot != nil {
```
tbg, line 1033: Do `m.Snapshot` messages not carry an index? You might hit the `if m.Index != 0` branch above and then *not* ignore the message.

raft/raft.go
```diff
@@ -961,6 +1054,20 @@ func (r *raft) Step(m pb.Message) error {
...
+		}
+		if m.Snapshot != nil {
+			r.appliedSnap(m.Snapshot)
+		}
+
+	case pb.MsgStorageApplyResp:
+		r.appliedTo(m.Index)
+		// NOTE: we abuse the LogTerm field to store the aggregate entry size so
+		// that we don't need to introduce a new field on Message.
+		r.reduceUncommittedSize(m.LogTerm)
```
tbg, line 1067: Anchoring a comment here because there's an active discussion about this elsewhere.

raft/rafttest/interaction_env_handler_process_append_thread.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return nil
+}
+
+// ProcessAppendThread runs processes a single message on the "append" thread of
+// the node with the given index.
+func (env *InteractionEnv) ProcessAppendThread(idx int) error {
+	n := &env.Nodes[idx]
+	if len(n.AppendWork) == 0 {
+		return fmt.Errorf("no append work to perform")
```
tbg, line 48: Is this really an error?

raft/rawnode.go
```diff
@@ -170,7 +187,10 @@ func (rn *RawNode) HasReady() bool {
 	if r.raftLog.hasNextUnstableSnapshot() {
```
tbg, line 426: Should we order this one down? After all, this is a rare case, so we don't have to run it in the common path. Checking `r.msgs`, `r.msgsAfterAppend` should probably be first, followed by hard state.

Also, `r.softState()` probably allocates on each `HasReady` call, which seems dumb. We should make that method return a value and also order it way down since it's basically not usually true.

Not really what you're after in this PR so maybe entomb this comment as a TODO instead.

raft/rawnode.go
```diff
@@ -181,8 +201,21 @@ func (rn *RawNode) HasReady() bool {
...
 // last Ready results.
-func (rn *RawNode) Advance(rd Ready) {
-	rn.raft.advance(rd)
+//
+// NOTE: Advance must not be called when using AsyncStorageWrites. Response messages from
+// the local append and apply threads take its place.
+func (rn *RawNode) Advance(_ Ready) {
+	// The actions performed by this function are encoded into stepsOnAdvance in
+	// acceptReady. In earlier versions of this library, they were computed from
+	// the provided Ready struct. Retain the unused parameter for compatability.
```
tbg, line 449: compatibility

raft/rawnode_test.go
```diff
@@ -342,6 +342,7 @@ func TestRawNodeProposeAndConfChange(t *testing.T) {
 				if len(rd.Entries) > 0 {
 					t.Fatal("expected no more entries")
 				}
+				rawNode.Advance(rd)
```
tbg, line 345: This got caused by the Ready-Ready without Advance fatal you introduced, right?

tbg (commented):
Finishing up first pass.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+func newStorageAppendRespMsg(r *raft, rd Ready) pb.Message {
+	m := pb.Message{
+		Type: pb.MsgStorageAppendResp,
+		To:   r.id,
+		From: LocalAppendThread,
+		// Dropped after term change, see below.
+		Term: r.Term,
+	}
+	if r.raftLog.hasNextOrInProgressUnstableEnts() {
+		// If the raft log has unstable entries, attach the last index and term to the
```
tbg: the last index of the append (which is really also the unstable log's last index, but good to be precise anyway).

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+		// response message. This (index, term) tuple will be handed back and consulted
+		// when the stability of those log entries is signaled to the unstable. If the
+		// (index, term) match the unstable log by the time the response is received,
```
tbg: Could you say more about what exactly it means to "match"? I think it means that we look up the term for the resp message's index and compare it with the resp message's log term. Then, additionally we compare the *term* (not log term) with the current (unstable) term, and only handle the message if the two match. There is no requirement that the unstable storage has a matching last index (it typically won't since other apps have arrived in the meantime).

tbg:
I know it's tempting but please don't rebase, I'm worried we'll lose a lot of the review state. Mind just continuing the commit history with fixup commits (`git commit --fixup :/^raft:.support.asynchronous` for example) and no reordering so I can keep my sanity? Looks like there are a slew of little fixes and edits to come.

The review is pretty large but a lot of it is cosmetic. A few TODOs remain that should be addressed before merge, I tried to point them all out via comments. I'm in favor of deferring anything that doesn't need to be done now, given the size of this PR.

The main aspect of this PR that still needs a resolution is the "To/From" field use https://github.com/etcd-io/etcd/pull/14627#discussion_r1015501972, feels like reviewers are generally skeptical of this. Maybe we should schedule sync time so you can try to convince me again (or vice versa).

I also need to understand the uncommitted size tracking better https://github.com/etcd-io/etcd/pull/14627/files/1abfc66b165d234251dd590051629428380cbcc2..1ccb57531dfedf2f6094d4adb2b9f486a6425007#diff-b9adbc46e4a317ffbb3d11a66c38d6b9af41a09170d77d87efbd96d115da452f

The LogTerm abuse also seems unsavory - would not mind this as a follow-up fix - but then we have to worry about users picking up this first pass and then having to migrate to the next, so better to keep it simple.

raft/rafttest/interaction_env_handler_process_append_thread.go
ahrtr: Note: if there is no any parameter, please use errors.New instead.

raft/rafttest/interaction_env_handler_add_nodes.go
nvanbenschoten: Done.

raft/raft_test.go
```diff
@@ -1763,6 +1768,90 @@ func testCandidateResetTerm(t *testing.T, mt pb.MessageType) {
...
+	}
+
+	// Its self-vote does not make its way to its ProgressTracker.
+	granted, _, _ := sm.prs.TallyVotes()
+	if granted != 0 {
+		t.Errorf("granted = %v, want %v", granted, 0)
+	}
+}
+
+func TestCandidateDeliversPreCandidateSelfVoteAfterBecomingCandidate(t *testing.T) {
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/raft_test.go
```diff
@@ -1763,6 +1768,90 @@ func testCandidateResetTerm(t *testing.T, mt pb.MessageType) {
...
+		t.Errorf("granted = %v, want %v", granted, 0)
+	}
+
+	// A single vote from n2 does not move n1 to the leader.
+	sm.Step(pb.Message{From: 2, To: 1, Term: sm.Term, Type: pb.MsgVoteResp})
+	if sm.state != StateCandidate {
+		t.Errorf("state = %v, want %v", sm.state, StateCandidate)
+	}
+
+	// n1 becomes the leader once its self-vote is received.
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/log_unstable.go
```diff
@@ -25,11 +35,27 @@ type unstable struct {
...
+	// entries[:offsetInProgress-offset] are being written to storage.
+	// Like offset, offsetInProgress is exclusive, meaning that it
+	// contains the index following the largest in-progress entry.
+	// Invariant: offset <= offsetInProgress
+	offsetInProgress uint64
 
 	logger Logger
 }
 
+// init initializes the unstable log.
```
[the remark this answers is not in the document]
nvanbenschoten: Inlined.

raft/log_unstable.go
```diff
@@ -72,6 +98,42 @@ func (u *unstable) maybeTerm(i uint64) (uint64, bool) {
 	return u.entries[i-u.offset].Term, true
 }
 
+// nextEntries returns the unstable entries that are not already in the process
+// of being written to storage.
+func (u *unstable) nextEntries() []pb.Entry {
```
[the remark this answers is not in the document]
nvanbenschoten: Yes it would, but I wanted to keep the nil return value for cases where there are no entries to return. This is placed directly in a `Ready.Entries`, so it's possible that users of the library (incorrectly) compare the slice against nil to check whether anything needs to be appended. This isn't the PR to change that behavior.

raft/log_unstable.go
```diff
@@ -113,34 +180,40 @@ func (u *unstable) shrinkEntriesArray() {
...
+	u.snapshotInProgress = false
 }
 
 func (u *unstable) truncateAndAppend(ents []pb.Entry) {
 	after := ents[0].Index
 	switch {
 	case after == u.offset+uint64(len(u.entries)):
-		// after is the next index in the u.entries
-		// directly append
+		// after is the next index in the u.entries directly append.
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/log_unstable.go
```diff
@@ -113,34 +180,40 @@ func (u *unstable) shrinkEntriesArray() {
...
 func (u *unstable) restore(s pb.Snapshot) {
 	u.offset = s.Metadata.Index + 1
+	u.offsetInProgress = u.offset
 	u.entries = nil
 	u.snapshot = &s
+	u.snapshotInProgress = false
 }
 
 func (u *unstable) truncateAndAppend(ents []pb.Entry) {
 	after := ents[0].Index
```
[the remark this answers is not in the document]
nvanbenschoten: Agreed. I'll save for a different PR to keep the code movement down in this one though.

raft/log_unstable.go
```diff
@@ -113,34 +180,40 @@ func (u *unstable) shrinkEntriesArray() {
...
 		u.entries = ents
+		u.offset = after
+		u.offsetInProgress = u.offset
 	default:
-		// truncate to after and copy to u.entries
-		// then append
+		// truncate to after and copy to u.entries then append.
 		u.logger.Infof("truncate the unstable entries before index %d", after)
-		u.entries = append([]pb.Entry{}, u.slice(u.offset, after)...)
+		keep := u.slice(u.offset, after)
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/log_unstable.go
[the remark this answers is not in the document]
nvanbenschoten: I added a TODO.

raft/log_unstable.go
```diff
@@ -113,34 +180,40 @@ func (u *unstable) shrinkEntriesArray() {
...
 
 func (u *unstable) truncateAndAppend(ents []pb.Entry) {
 	after := ents[0].Index
 	switch {
 	case after == u.offset+uint64(len(u.entries)):
-		// after is the next index in the u.entries
-		// directly append
+		// after is the next index in the u.entries directly append.
 		u.entries = append(u.entries, ents...)
 	case after <= u.offset:
```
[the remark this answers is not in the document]
nvanbenschoten: Is this always true? The entries are being replaced by a leader in a newer term, but I think it is possible for entries with larger terms to be replaced by entries with smaller terms.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+		//
+		//  if len(rd.Entries) > 0 { ... }
+		//
+		// To do so, we attach r.raftLog.lastIndex() and r.raftLog.lastTerm(), not the
+		// (index, term) of the last entry in rd.Entries. If rd.Entries is not empty,
+		// these will be the same. However, if rd.Entries is empty, we still want to
+		// attest that this (index, term) is correct at the current term, in case the
+		// MsgStorageAppend that contained the last entry in the unstable slice carried
+		// an earlier term and was dropped.
+		// TODO(nvanbenschoten): test this behavior in a data-driven test.
```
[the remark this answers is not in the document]
nvanbenschoten: This was done in a later commit (`raft: add data-driven test for subtle async storage write aba problem`).

raft/rawnode.go
```diff
@@ -171,7 +171,7 @@ func (rn *RawNode) HasReady() bool {
 	if r.raftLog.hasNextUnstableSnapshot() {
 		return true
 	}
-	if len(r.msgs) > 0 || r.raftLog.hasNextUnstableEnts() || r.raftLog.hasNextCommittedEnts() {
+	if len(r.msgs) > 0 || r.raftLog.hasNextUnstableEnts() || r.raftLog.hasNextCommittedEnts(true /* allowUnstable */) {
```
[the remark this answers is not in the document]
nvanbenschoten: Yes, this was an intermediate state.

raft/log.go
```diff
@@ -192,13 +192,16 @@ func (l *raftLog) hasNextUnstableEnts() bool {
...
 		// See comment in hasNextCommittedEnts.
 		return nil
 	}
-	if l.committed > l.applying {
-		lo, hi := l.applying+1, l.committed+1 // [lo, hi)
+	lo, hi := l.applying+1, l.committed+1 // [lo, hi)
+	if !allowUnstable {
+		hi = min(hi, l.unstable.offset)
+	}
+	if lo < hi {
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/rafttest/interaction_env_handler_process_append_thread.go
nvanbenschoten: Done.

raft/rawnode.go
nvanbenschoten: Done.

raft/rawnode_test.go
nvanbenschoten: Right. The fatal is actually gone in these later revisions, but I've kept the testing changes because they don't hurt.

Interestingly, with the way that we're remembering `stepsOnAdvance` on the `RawNode`, it actually is now safe to Ready-Ready without Advance. Or at least, I think it is, whereas it clearly wasn't before. Either way, the library still says it's not supported.

nvanbenschoten:
Thanks @tbg for the thorough review! I'm still working through your comments and trying to thin out the threads. Please don't bother taking a look until I ping here, at which point we can bottom out on the remaining few threads.

I've also spent portions of the past week integrating this change back into CockroachDB. I'll be posting results in https://github.com/cockroachdb/cockroach/issues/17500 later. The high-level summary is that integration of async storage writes (log appends only, not yet entry application) into cockroach does provide the same 20-40% average and tail latency improvements that we had simulated above. However, it doesn't provide the throughput improvements at the top end because cockroach becomes CPU bound earlier than rafttoy and fails to benefit much from the added opportunities for batching. That roughly matches expectations, as the latency win was always the goal here and the possibility of a throughput win in cockroach was speculative. I think it's also possible that the batching is more impactful for entry application than for log appends, so pulling that async in cockroach (at some later point) will still provide a throughput win.

raft/rawnode.go
nvanbenschoten: Done.

raft/log.go
```diff
@@ -210,7 +218,7 @@ func (l *raftLog) hasNextCommittedEnts() bool {
 		// first.
 		return false
 	}
-	return l.committed > l.applied
+	return l.committed > l.applying
```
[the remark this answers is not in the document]
nvanbenschoten: The old `applied` index had a mix of semantics. This commit didn't as much rename `applied` to `applying` as it did introduce the new `applying`, which is incremented when accepting a ready. `applied` continues to be incremented when advancing a ready.

I've gone through with find usages and confirmed that nothing was missed in this decomposition. I've also added to each fields comment to explain its use.

raft/doc.go
```diff
@@ -295,5 +383,15 @@ stale log entries:
...
 	the leader sets it back to probe.
+
+	'MsgStorageAppend' is a message from a node to its local append storage
+	thread to write entries, hard state, and/or a snapshot to stable storage.
+	The message will carry one or more responses, one of which will be a
+	'MsgStorageAppendResp' back to itself. Used with AsynchronousStorageWrites.
+
+	'MsgStorageApply' is a message from a node to its local apply storage
+	thread to apply committed entries. The message will carry one or more
+	responses, one of which will be a 'MsgStorageApplyResp' back to itself.
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/doc.go
```diff
@@ -295,5 +383,15 @@ stale log entries:
 	that the follower that sent this 'MsgUnreachable' is not reachable, often
 	indicating 'MsgApp' is lost. When follower's progress state is replicate,
 	the leader sets it back to probe.
+
+	'MsgStorageAppend' is a message from a node to its local append storage
+	thread to write entries, hard state, and/or a snapshot to stable storage.
+	The message will carry one or more responses, one of which will be a
+	'MsgStorageAppendResp' back to itself. Used with AsynchronousStorageWrites.
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/node.go
nvanbenschoten: Done.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+		From: LocalAppendThread,
+		// Dropped after term change, see below.
+		Term: r.Term,
+	}
+	if r.raftLog.hasNextOrInProgressUnstableEnts() {
+		// If the raft log has unstable entries, attach the last index and term to the
+		// response message. This (index, term) tuple will be handed back and consulted
+		// when the stability of those log entries is signaled to the unstable. If the
+		// (index, term) match the unstable log by the time the response is received,
+		// the unstable log can be truncated.
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/raft.go
nvanbenschoten: Done.

raft/node.go
```diff
@@ -57,7 +57,12 @@ type Ready struct {
 
 	// The current state of a Node to be saved to stable storage BEFORE
 	// Messages are sent.
+	//
 	// HardState will be equal to empty state if there is no update.
+	//
+	// If async storage writes are enabled, this field does not need to be acted
```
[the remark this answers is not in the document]
nvanbenschoten: I went back and forth on this. Keeping these fields in place makes the logic in `RawNode.readyWithoutAccept` and `RawNode.acceptReady` easier to write. It also preserves the ability to describe a Ready struct's contents (e.g. `DescribeReady`) without needing to describe how that Ready struct should be consumed. So I'm tempted to leave this as is, unless you feel strongly.

raft/node.go
```diff
@@ -68,18 +73,34 @@ type Ready struct {
...
+	// on immediately. It will be reflected in a MsgStorageApply message in the
+	// Messages slice.
 	CommittedEntries []pb.Entry
 
-	// Messages specifies outbound messages to be sent AFTER Entries are
-	// committed to stable storage.
+	// Messages specifies outbound messages.
+	//
+	// If async storage writes are not enabled, these messages must be sent
+	// AFTER Entries are committed to stable storage. If async storage writes
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/node.go
```diff
@@ -68,18 +73,34 @@ type Ready struct {
...
+	// are enabled, they can be sent immediately.
+	//
 	// If it contains a MsgSnap message, the application MUST report back to raft
 	// when the snapshot has been received or has failed by calling ReportSnapshot.
 	Messages []pb.Message
```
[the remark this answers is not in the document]
nvanbenschoten: Good point. Switched "synchronous"/"asynchronous" with "durable"/"non-durable".

You are correct that the field is unused with async append. Instead, the cases where `MustSync` used to be false now fall into two separate categories:
- **no HardState.Commit update**: in this case, there is no `MsgStorageAppend` message
- **only HardState.Commit update**: in this case, there is a `MsgStorageAppend` message, but it has no `Response` messages. This is nice because it means that logic that handles `MsgStorageAppend` can have logic which looks like:

```go
func appender(msg raftpb.Message) {
    var b pebble.Batch
    if len(msg.Entries) > 0 {
        b.Append(msg.Entries...)
    }
    if msg.HardState != nil {
        // NOTE: will change a bit with your comment above.
        b.SetHardState(msg.Entries...)
    }
    if len(msg.Responses) == 0 {
        b.CommitWithoutSync()
    } else {
        b.CommitWithSyncCallback(func() {
            sendMessages(msg.Responses...)
        })
    }
}
```
In a sense, we're no longer making a distinction about whether the "appender" logic needs to force an immediate sync or not. Instead, we're using the presence of response messages to decide whether to register an on-sync callback. The timing of that fsync can then be decided outside of this code.

raft/node.go
nvanbenschoten: I don't think that constraint would change anything here. This is just constructing the response message, so I'll keep it general for now.

raft/node.go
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/raft.go
nvanbenschoten: That was actually intentional. I improved the log message to make that more clear.

raft/raft.go
nvanbenschoten: Yes, the gap in tracking that you proposed is real. It's been since this tracking was introduced because we didn't want to scan the uncommitted log on leadership changes.

I've cleaned some of this logic up so that the code is more clear, but that didn't resolve the problem. Let's extract that discussion from this PR, since it's not a new concern.

raft/node.go
```diff
@@ -144,7 +152,8 @@ type Node interface {
 	Step(ctx context.Context, msg pb.Message) error
 
 	// Ready returns a channel that returns the current point-in-time state.
-	// Users of the Node must call Advance after retrieving the state returned by Ready.
+	// Users of the Node must call Advance after retrieving the state returned by Ready (unless
+	// async storage writes is enabled, in which case it should never be called).
```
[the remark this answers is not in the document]
nvanbenschoten: I hear what you're saying, but IMO that leaves room for bugs. I'll keep this as a panic to avoid misuse. It's easy enough to only call `Advance` when `!AsyncStorageWritesEnabled` as a user of this library.

raft/node.go
nvanbenschoten: Deleted to make this accounting more clear.

raft/raft.go
nvanbenschoten: Done.

raft/log.go
```diff
@@ -266,6 +284,8 @@ func (l *raftLog) stableTo(i, t uint64) { l.unstable.stableTo(i, t) }
 
 func (l *raftLog) stableSnapTo(i uint64) { l.unstable.stableSnapTo(i) }
 
+func (l *raftLog) acceptUnstable() { l.unstable.inProgress() }
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/log_unstable.go
```diff
@@ -72,6 +98,42 @@ func (u *unstable) maybeTerm(i uint64) (uint64, bool) {
...
+	}
+	return u.snapshot
+}
+
+// inProgress marks all entries and the snapshot, if any, in the unstable as
+// having begun the process of being written to storage. The entries/snapshot
+// will no longer be returned from nextEntries/nextSnapshot. However, new
+// entries/snapshots added after a call to inProgress will be returned from
+// those methods, until the next call to inProgress.
+func (u *unstable) inProgress() {
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/raft.go
nvanbenschoten: Done.

raft/node.go
nvanbenschoten: Done.

raft/node.go
```diff
@@ -591,3 +628,184 @@ func MustSync(st, prevst pb.HardState, entsnum int) bool {
...
+		//
+		// However, this replaces a correctness problem with a liveness problem. If we
+		// only attempted to truncate the unstable log when appending new entries but
+		// also occasionally dropped these responses, then quiescence of new log entries
+		// could lead to the unstable log never being truncated.
+		//
+		// To combat this, we attempt to truncate the log on all MsgStorageAppendResp
+		// messages where the unstable log is not empty, not just those associated with
+		// entry appends. This includes MsgStorageAppendResp messages associated with an
+		// updated HardState, which occur after a term change.
```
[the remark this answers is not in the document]
nvanbenschoten: Done.

raft/raft.go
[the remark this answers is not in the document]
nvanbenschoten: I agree that we don't strictly need the pseudo-targets and that clients using this library could get away without them as long as they carefully read documentation of the library which describes which message types flow in which directions and how to route based on message types.

In my opinion, the pseudo-targets are a form of documentation so that we don't need to push those concerns onto users. They also make the message-passing interface consistent. Without them, users of the library are forced to ask:
* what does it mean for a message to have no target? Who do I send it to? Why do some messages not have targets?
* is `Msg.To: r.ID` a reference to the local `etcd/raft` state machine library or a reference to a local storage component. How do I distinguish the two?
* if we're not even assigning `Msg.To: r.ID` on `Response` messages, am I also matching on `MsgStorageAppendResp` and `MsgStorageApplyResp`? If so, why do `Response` slices contain some ID-ed messages and some not? Why do I sometimes (on a leader) have `Responses: [{MsgAppResp, To: r.ID}, {MsgStorageAppendResp, To: 0}]`?
* which local messages can be reordered and which need to be acted on in order?

The question to ask here isn't "do we need these IDs?", it's "how much harder would this be to explain without them?"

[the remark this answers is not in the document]
tbg: I could write a lengthy response here making the opposite case, but I think both arguments have merit and the choice is ultimately subjective. Let's go with your preference.

xiang90:
> Thanks @tbg for the thorough review! I'm still working through your comments and trying to thin out the threads. Please don't bother taking a look until I ping here, at which point we can bottom out on the remaining few threads.

Amazing work! I assume this would also help etcd with tail latency.

tbg:
https://github.com/etcd-io/raft/pull/8

PR #14629, opened 2022-10-26
test: fix anti-pattern naming problem in config.ClusterConfig

clarkfw:
This PR tries to fix the anti-pattern naming problem(see [#14389](https://github.com/etcd-io/etcd/pull/14389#discussion_r955668262)), such as `DisableStrictReconfigCheck`(written by me🤣).

As far as I see, at least 4 structs have this problem:
1. [ctlCtx](https://github.com/etcd-io/etcd/blob/8ce81a1624e47f173d195bc335145b9923ee35e6/tests/e2e/ctl_v3_test.go#L124)
2. [config.ClusterConfig](https://github.com/etcd-io/etcd/blob/main/tests/framework/config/cluster.go)
3. [e2e.EtcdProcessClusterConfig](https://github.com/etcd-io/etcd/blob/8ce81a1624e47f173d195bc335145b9923ee35e6/tests/framework/e2e/cluster.go#L138)
4. [integration.ClusterConfig](https://github.com/etcd-io/etcd/blob/8ce81a1624e47f173d195bc335145b9923ee35e6/tests/framework/integration/cluster.go#L135)

This PR only tries to fix `(2)config.ClusterConfig`. 
`(3)e2e.EtcdProcessClusterConfig` and `(4)integration.ClusterConfig` will be fixed after this PR.

This PR changed `(2)config.ClusterConfig` with the **_Functional Options Pattern_**. I’m not sure if this is a proper solution. Please take a look and give some suggestions.

PR #14632, opened 2022-10-26
tests: add a ClusterOption interface for ClusterSize in config.Cluster

clarkfw:
As mentioned in [#14629](https://github.com/etcd-io/etcd/pull/14629#discussion_r1005582285), the code could be more concise with these changes.

From my understanding, `3` was used as the default value of `clusterSize`.

@ahrtr please have a look, thanks!


ahrtr (approved):
LGTM

Thank you @clarkfw

serathius (approved):
Nice! one additional improvement would be to make the `NewCluster` also accept configuration as options.

PR #14633, opened 2022-10-26
raft: send empty appends when replication is paused

pavelkalinnikov:
When Inflights to a particular node is full, i.e. MaxInflightMsgs for the append messages flow is saturated, it is still necessary to continue sending MsgApp to ensure progress. Currently this is achieved by "forgetting" the first in-flight message in the window, which frees up quota for one new MsgApp.

This new message is constructed in such a way that it potentially has multiple entries, or a large entry. The effect of this is that the in-flight limitations can be exceeded arbitrarily, for as long as the flow to this node continues being saturated. In particular, if a follower is stuck, the leader will keep sending entries to it.

This commit makes the MsgApp empty when Inflights is saturated, and prevents the described leakage of Entries to slow followers.

Signed-off-by: Pavel Kalinnikov <pavel@cockroachlabs.com>

### DIFF
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -1304,7 +1304,7 @@ func stepLeader(r *raft, m pb.Message) error {
 		}
 	case pb.MsgHeartbeatResp:
 		pr.RecentActive = true
-		pr.ProbeSent = false
+		pr.MsgAppFlowPaused = false
 
 		// NB: if the follower is paused (full Inflights), this will still send an
 		// empty append, allowing it to recover from situations in which all the
@@ -1349,7 +1349,7 @@ func stepLeader(r *raft, m pb.Message) error {
 		// If snapshot finish, wait for the MsgAppResp from the remote node before sending
 		// out the next MsgApp.
 		// If snapshot failure, wait for a heartbeat interval before next try
-		pr.ProbeSent = true
+		pr.MsgAppFlowPaused = true
 	case pb.MsgUnreachable:
 		// During optimistic replication, if the remote becomes unreachable,
 		// there is huge probability that a MsgApp is lost.
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -472,20 +472,9 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 	}
 
 	// Send the actual MsgApp otherwise, and update the progress accordingly.
-	// TODO(pavelkalinnikov): factor out the Progress update to a method
 	next := pr.Next // save Next for later, as the progress update can change it
-	if n := len(ents); n != 0 {
-		switch pr.State {
-		// optimistically increase the next when in StateReplicate
-		case tracker.StateReplicate:
-			last := ents[n-1].Index
-			pr.OptimisticUpdate(last)
-			pr.Inflights.Add(last)
-		case tracker.StateProbe:
-			pr.ProbeSent = true
-		default:
-			r.logger.Panicf("%x is sending append in unhandled state %s", r.id, pr.State)
-		}
+	if err := pr.UpdateOnEntriesSend(len(ents), next); err != nil {
+		r.logger.Panicf("%x: %v", r.id, err)
 	}
 	r.send(pb.Message{
 		To:      to,
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -439,7 +439,18 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 	}
 
 	term, errt := r.raftLog.term(pr.Next - 1)
-	ents, erre := r.raftLog.entries(pr.Next, r.maxMsgSize)
+	var ents []pb.Entry
+	var erre error
+	// In a throttled StateReplicate only send empty MsgApp, to ensure progress.
+	// Otherwise, if we had a full Inflights and all inflight messages were in
+	// fact dropped, replication to that follower would stall. Instead, an empty
+	// MsgApp will eventually reach the follower (heartbeats responses prompt the
+	// leader to send an append), allowing it to be acked or rejected, both of
+	// which will clear out Inflights.
+	if pr.State != tracker.StateReplicate || !pr.Inflights.Full() {
+		ents, erre = r.raftLog.entries(pr.Next, r.maxMsgSize)
+	}
+
 	if len(ents) == 0 && !sendIfEmpty {
 		return false
 	}
@@ -1295,10 +1306,10 @@ func stepLeader(r *raft, m pb.Message) error {
 		pr.RecentActive = true
 		pr.ProbeSent = false
 
-		// free one slot for the full inflights window to allow progress.
-		if pr.State == tracker.StateReplicate && pr.Inflights.Full() {
-			pr.Inflights.FreeFirstOne()
-		}
+		// NB: if the follower is paused (full Inflights), this will still send an
+		// empty append, allowing it to recover from situations in which all the
+		// messages that filled up Inflights in the first place were dropped. Note
+		// also that the outgoing heartbeat already communicated the commit index.
 		if pr.Match < r.raftLog.lastIndex() {
 			r.sendAppend(m.From)
 		}
--- a/raft/raft.go
+++ b/raft/raft.go
@@ -437,8 +437,6 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 	if pr.IsPaused() {
 		return false
 	}
-	m := pb.Message{}
-	m.To = to
 
 	term, errt := r.raftLog.term(pr.Next - 1)
 	ents, erre := r.raftLog.entries(pr.Next, r.maxMsgSize)
@@ -452,7 +450,6 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
 			return false
 		}
 
-		m.Type = pb.MsgSnap
 		snapshot, err := r.raftLog.snapshot()
 		if err != nil {
 			if err == ErrSnapshotTemporarilyUnavailable {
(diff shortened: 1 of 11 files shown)


raft/raft_flow_control_test.go
```diff
@@ -50,8 +50,8 @@ func TestMsgAppFlowControlFull(t *testing.T) {
 	for i := 0; i < 10; i++ {
 		r.Step(pb.Message{From: 1, To: 1, Type: pb.MsgProp, Entries: []pb.Entry{{Data: []byte("somedata")}}})
 		ms := r.readMessages()
-		if len(ms) != 0 {
-			t.Fatalf("#%d: len(ms) = %d, want 0", i, len(ms))
+		if len(ms) != 1 || len(ms[0].Entries) != 0 {
```
pavelkalinnikov, line 53: This means that now every proposal triggers an empty `MsgApp` when Inflights is saturated. Previously it would only be after a heartbeat response.

This is probably because `maybeSendAppend`/`sendAppend` is invoked from a few other places. We may need to restrict this, to minimize the number of empty appends.

pavelkalinnikov: Done.

pavelkalinnikov:
@tbg This is now ready for review.

raft/tracker/progress.go
```diff
@@ -59,6 +59,12 @@ type Progress struct {
 	// true, raft should pause sending replication message to this peer until
 	// ProbeSent is reset. See ProbeAcked() and IsPaused().
 	ProbeSent bool
+	// MsgAppSent is used while this follower is in StateReplicate but the flow of
+	// messages is throttled, i.e. Inflights is full. In this state we still need
+	// to send empty MsgApp to guarantee progress, but we do so only when
+	// MsgAppSent == false.
+	// This works similarly to ProbeSent, only in a different state.
+	MsgAppSent bool
```
pavelkalinnikov, line 67: We could probably use the same `MsgAppSent` here both for probe and replicate state. The technique for both cases is the same, and boils down to "probes" on heartbeats.

pavelkalinnikov: Added a commit after this one implementing this idea. Please let me know which way it looks better. I prefer re-using the same bool, cause it's the same heuristic.

raft/raft.go
```diff
@@ -437,11 +437,16 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
...
 	}
-	m := pb.Message{}
-	m.To = to
+	m := pb.Message{To: to}
 
 	term, errt := r.raftLog.term(pr.Next - 1)
-	ents, erre := r.raftLog.entries(pr.Next, r.maxMsgSize)
+	var ents []pb.Entry
+	var erre error
+	// In a throttled StateReplicate only send empty MsgApp, to ensure progress.
```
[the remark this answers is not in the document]
tbg: Also, could you comment on the significance of `sendIfEmpty`? If that is `false`, then if the follower's inflight is full we are definitely returning on line 451. Is that what we want? Feels like we'd really want to get the entries unconditionally, and then, if there are entries, set `sendIfEmpty` to false, and if the follower is paused send an empty append anyway (instead of a "full" append).

Right now I'm left wondering why we have a caller that passes `sendIfEmpty==true` in when we need it. I know now, after looking at the handling of `MsgHeartbeatResp`, that we will in fact pass `sendIfEmpty==true`.

Having looked into all of this, the code here checks out - we want to send the empty append resulting from being full only if the caller requested empty appends to be sent anyway. This is the case on heartbeats (and in some other cases, for example whenever the commit index changes). But try to package all of this lifecycle into a comment somewhere.

raft/tracker/progress.go
tbg: +1 on keeping only one field, but I found `MsgAppSent` confusing. Raft sends MsgApp all the time; if this field is specific to throttling let's give it a more fitting name. I think ProbeSent actually made more sense: in the steady state, are not necessary because the follower is acking writes before inflight buffer fills up. We send probes periodically; the field is reset on each heartbeat or when a probe is successful (which means we don't need to probe any more). Probes can occur in two states: either in StateProbe, where we are trying to find out at what index the follower can accept appends (and here I think we're sending full log entries), or when inflights is full (in which case we send an empty append, just to make sure the follower can give us an ack or rej back in case previous msgs got dropped).

tbg: In particular, `IsPaused()` just becomes a read on `ProbeSent` and we don't need the awkward conjunction with `Inflights.Full()`. Instead, when we see that inflights are full, and we send a probe, we set ProbeSent if it isn't already set. That seems simpler because now the field directly controls what is going on.

tbg (approved):
Looks good! I'd mainly try to rearrange code and add more comments so that the intricacies of this dance are bit easier to grasp for future readers. See inline comments.

I also think it is worthwhile to add a `TestInteraction` test for this.

raft/raft.go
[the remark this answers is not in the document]
pavelkalinnikov: > Could add here:

Done.

> Also, could you check that if Inflights is full, and the Log{Term,Index} in the empty MsgApp that reaches the follower actually matches the end of their log, the follower still responds with an MsgAppResp? If it didn't then we'd get stuck anyway.

Verified, seems ok. See the data-driven test.

> I'd also appreciate a TestInteraction test for a follower that gets these empty appends, because it essentially serves as an end-to-end test.

Done. See the first commit introducing the data-driven test. Note the TODO in it which flags the issue that this PR solves. Note the change in the data-driven test after the fix.

> Also, isn't there a (pre-existing) bug here?

Is this addressable independently? Let's perhaps move it to a separate issue/PR to keep this one self-contained.

> But try to package all of this lifecycle into a comment somewhere.

Not yet done, will do tomorrow.

raft/tracker/progress.go
pavelkalinnikov: Removed the rename commit, it's `ProbeSent` again. Will try your proposal tomorrow.

pavelkalinnikov (commented):
@tbg Addressed most comments, notably added the data-driven test.

raft/tracker/progress.go
pavelkalinnikov: Trying this proposal. The thing is, when we remove the awkward conjunction in one place, it reappears in other places (we should check both `Full()` and "recently sent" facts anyway), so the net awkwardness reduction is 0. I think we should just push a bit more stuff to `Progress` type, as noted in a TODO.

pavelkalinnikov: Added another commit with a better encapsulated logic in `Progress`.

pavelkalinnikov:
@ahrtr Please review or redirect to someone in etcd/raft folks. Commit-by-commit review is recommended. Only the 5-th commit `raft: send empty appends when replication is paused` changes the behaviour.

Overall commits structure:

- First 2 commits only add and strengthen tests. In particular, the first commit captures the problematic behaviour that this PR solves.
- Commits 3 and 4 contain drive-by small refactorings that don't change behaviour, but pave the way towards the fix.
- Commit 5 is the fix. Notice how it modifies the behaviour, which is reflected in tests.
- The last commit removes a no longer used method.

raft/testdata/replicate_pause.txt
```diff
@@ -0,0 +1,190 @@
+# This test ensures that MsgApp stream to a follower is paused when the
```
tbg, line 1: Nice!

raft/tracker/progress.go
```diff
@@ -55,9 +55,12 @@ type Progress struct {
...
 
-	// ProbeSent is used while this follower is in StateProbe. When ProbeSent is
-	// true, raft should pause sending replication message to this peer until
-	// ProbeSent is reset. See ProbeAcked() and IsPaused().
+	// ProbeSent is true when a "probe" MsgApp was sent to this follower recently,
+	// and we haven't heard from it back yet. Used when the MsgApp flow is
+	// throttled, i.e. when State is StateProbe, or StateReplicate with saturated
+	// Inflights. In both cases, we need to continue sending MsgApp once in a
+	// while to guarantee progress, but we do so only when ProbeSent is false, to
+	// not overflow the receiver. See IsPaused().
```
tbg, line 63: Might as well add here that there is periodicity here, i.e.

> but we only do so when ProbeSent is false (it is reset when receiving a heartbeat response) [...]

tbg (approved):

tbg:
Looks good to me.

@ahrtr mind taking a look? There is an E2E failure[^1]; I find these impossible to decode and they often seem to be flaky, so not sure if it's something that we should just retry.

[^1]: https://github.com/etcd-io/etcd/actions/runs/3373310602/jobs/5598102673#step:5:1112

ahrtr:
@tbg @pavelkalinnikov Sorry for the late response. I was busy with something else. Will have a deep dive into this PR sometime next week.

raft/tracker/progress.go
pavelkalinnikov: Done.

raft/raft.go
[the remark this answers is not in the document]
pavelkalinnikov: All done, except the `sendIfEmpty` commenting. I think that `maybeSendAppend()` method comment speaks about it quite ok, wasn't sure what your concern was. Do you have a specific place/comment in mind?

pavelkalinnikov:
> There is an E2E failure

Probably flaky. The last run is good.

raft/tracker/progress.go
```diff
@@ -55,9 +55,13 @@ type Progress struct {
...
-	// true, raft should pause sending replication message to this peer until
-	// ProbeSent is reset. See ProbeAcked() and IsPaused().
+	// ProbeSent is true when a "probe" MsgApp was sent to this follower recently,
+	// and we haven't heard from it back yet. Used when the MsgApp flow is
+	// throttled, i.e. when State is StateProbe, or StateReplicate with saturated
+	// Inflights. In both cases, we need to continue sending MsgApp once in a
+	// while to guarantee progress, but we only do so when ProbeSent is false (it
+	// is reset on receiving a heartbeat response), to not overflow the receiver.
+	// See IsPaused().
 	ProbeSent bool
```
ahrtr, line 65: `ProbeSent` is supposed to be used only for the `StateProbe`.

Based on the comment above, the `ProbeSent` doesn't mean exactly what its name indicates; it's also used for the `tracker.StateReplicate` (set to `true` when the Inflight is full). It's an anti-pattern to me. I would suggest to dedicate `ProbeSent` to what's its name indicates (preferred), otherwise we need to change the name (not preferred).

raft/tracker/progress.go
```diff
@@ -137,6 +134,32 @@ func (pr *Progress) BecomeSnapshot(snapshoti uint64) {
...
+	switch pr.State {
+	case StateReplicate:
+		if entries > 0 {
+			last := nextIndex + uint64(entries) - 1
+			pr.OptimisticUpdate(last)
+			pr.Inflights.Add(last)
+		}
+		// If this message overflows the in-flights tracker, or it was already full,
+		// consider this message being a probe, so that the flow is paused.
+		pr.ProbeSent = pr.Inflights.Full()
```
ahrtr, line 149: See my comment  https://github.com/etcd-io/etcd/pull/14633#discussion_r1015986504

raft/tracker/progress.go
```diff
@@ -202,7 +225,7 @@ func (pr *Progress) IsPaused() bool {
 	case StateProbe:
 		return pr.ProbeSent
 	case StateReplicate:
-		return pr.Inflights.Full()
+		return pr.ProbeSent
```
ahrtr, line 228: Based on my previous comments, I would suggest to revert this change.

raft/raft.go
```diff
@@ -437,11 +437,20 @@ func (r *raft) maybeSendAppend(to uint64, sendIfEmpty bool) bool {
...
+	// Otherwise, if we had a full Inflights and all inflight messages were in
+	// fact dropped, replication to that follower would stall. Instead, an empty
+	// MsgApp will eventually reach the follower (heartbeats responses prompt the
+	// leader to send an append), allowing it to be acked or rejected, both of
+	// which will clear out Inflights.
+	if pr.State != tracker.StateReplicate || !pr.Inflights.Full() {
```
ahrtr, line 450:
```suggestion
	if !pr.Inflights.Full() {
```

Based on the condition, `r.raftLog.entries(...)` will not be called only when `pr.State == tracker.StateReplicate && pr.Inflights.Full() `.  But shouldn't it be an issue if `pr.Inflights.Full() &&  pr.State != tracker.StateReplicate` is `true`?

pavelkalinnikov (commented):
Addressed comments form @ahrtr

tbg (approved):

raft/tracker/progress.go
ahrtr: Please always keep readability in mind, otherwise any new contributors might get confused. 

Good to see the new name, which is clearer.

raft/tracker/progress.go
ahrtr: I am good after you changed the name (ProbeSent --> MsgAppFlowPaused).

ahrtr (approved):
Overall looks good to me. 

Thank you @pavelkalinnikov

ahrtr:
We recently added some mix-versions test, and also removed some dependencies, so please rebase this PR. Once all pipelines are green, then it's OK to merge this PR.

pavelkalinnikov:
> We recently added some mix-versions test, and also removed some dependencies, so please rebase this PR. Once all pipelines are green, then it's OK to merge this PR.

@ahrtr Rebased. Could you trigger CI once again and/or merge?

ahrtr:
> > We recently added some mix-versions test, and also removed some dependencies, so please rebase this PR. Once all pipelines are green, then it's OK to merge this PR.
> 
> @ahrtr Rebased. Could you trigger CI once again and/or merge?

Done.

PR #14637, opened 2022-10-27
[e2e] don't pass --endpoints flag in test HashKV implementation

veshij:
Flag is already set in https://github.com/etcd-io/etcd/blob/9bc4a63a419cd316f6920e37ec551b87d1c7c3a7/tests/framework/e2e/etcdctl.go#L317.
Duplicate flag leads to duplicate entries in hashkv output.


ahrtr (approved):
LGTM

Thank you @veshij

PR #14652, opened 2022-10-29
etcdclient: check mutex state in Unlock method of concurrency.Mutex.

chenyahui:
Check the values of myKey and myRev first in Unlock method to prevent calling Unlock without Lock. Because this may cause the value of pfx to be deleted by mistake.

### DIFF
--- a/client/v3/concurrency/mutex.go
+++ b/client/v3/concurrency/mutex.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
@@ -27,6 +28,7 @@ import (
 // ErrLocked is returned by TryLock when Mutex is already locked by another session.
 var ErrLocked = errors.New("mutex: Locked by another session")
 var ErrSessionExpired = errors.New("mutex: session is expired")
+var ErrLockReleased = errors.New("the lock has already been released")
 
 // Mutex implements the sync Locker interface with etcd
 type Mutex struct {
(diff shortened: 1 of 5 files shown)


client/v3/concurrency/mutex.go
```diff
@@ -128,6 +128,10 @@ func (m *Mutex) tryAcquire(ctx context.Context) (*v3.TxnResponse, error) {
 }
 
 func (m *Mutex) Unlock(ctx context.Context) error {
+	if m.myKey == "" || m.myRev < 0 {
+		panic("inconsistent mutex state")
+	}
```
ahrtr, line 133: Panicking might be a breaking change. Users might call `Unlock` multiple times, we should just ignore any following calling or return an error something "the lock has already been released" if it has already unlocked successfully instead of panicking.
 
```suggestion
	if m.myKey == "" || m.myKey == "\x00" || m.myRev <= 0 {
		return errors.New("the lock has already been released") // or return nil
	}
	if !strings.HasPrefix(s.myKey, s.pfx) {
		return fmt.Errorf("invalid key %q, it should have prefix %q", m.myKey, m.pfx)
	}	
```

Please also consider to add integration test cases.

ahrtr:
Please also signoff the commit: `git rebase HEAD~1 --signoff`

client/v3/concurrency/mutex.go
chenyahui: I will provide a new pr later~

chenyahui:
done

chenyahui:
The signature of some commits in this pr was accidentally polluted by me, I resubmitted a new pr. https://github.com/etcd-io/etcd/pull/14704

PR #14656, opened 2022-10-30
test: added e2e test case for issue 14571: etcd doesn't load auth info when recovering from a snapshot

ahrtr:

cc @mitake @biosvs @veshij

### DIFF
--- a/tests/e2e/ctl_v3_auth_test.go
+++ b/tests/e2e/ctl_v3_auth_test.go
@@ -17,10 +17,14 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"syscall"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"go.etcd.io/etcd/client/v3"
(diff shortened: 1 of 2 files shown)


ahrtr:
Added e2e test case for 3.5 for https://github.com/etcd-io/etcd/issues/14571

mitake:
@ahrtr LGTM, thanks a lot for creating the e2e test case! I tested that not refreshing range perm cache can result in failure of the test on my env. I think it's also valuable to add the test case to the main branch?

ahrtr:
> @ahrtr LGTM, thanks a lot for creating the e2e test case! I tested that not refreshing range perm cache can result in failure of the test on my env. I think it's also valuable to add the test case to the main branch?

@veshij already added e2e test on `main` branch in https://github.com/etcd-io/etcd/pull/14574. ` release-3.5` has huge difference from `main` branch, and there is no easy way to backport the e2e test to `release-3.5`. So I have to add a new e2e test for 3.5.

mitake:
Sorry I compared the e2e tests and checked that they are essentially testing the exact same scenario, I misunderstood.

PR #14657, opened 2022-10-30
test: deflake TestDowngradeUpgradeClusterOf3 timeout

fuweid:
In the TestDowngradeUpgradeCluster case, the brand-new cluster is using simple-config-changer, which means that entries has been committed before leader election and these entries will be applied when etcdserver starts to receive apply-requests. The simple-config-changer will mark the `confState` dirty and the storage backend precommit hook will update the `confState`.

For the new cluster, the storage version is nil at the beginning. And it will be v3.5 if the `confState` record has been committed. And it will be >v3.5 if the `storageVersion` record has been committed.

When the new cluster is ready, the leader will set init cluster version with v3.6.x. And then it will trigger the `monitorStorageVersion` to update the `storageVersion` to v3.6.x. If the `confState` record has been updated before cluster version update, we will get storageVersion record.

If the storage backend doesn't commit in time, the `monitorStorageVersion` won't update the version because of `cannot detect storage schema version: missing confstate information`.

And then we file the downgrade request before next round of `monitorStorageVersion`(per 4 second), the cluster version will be v3.5.0 which is equal to the `UnsafeDetectSchemaVersion`'s result. And we won't see that `The server is ready to downgrade`.

It is easy to reproduce the issue if you use cpuset or taskset to limit in two cpus.

So, we should wait for the new cluster's storage ready before downgrade request.

Fixes: #14540

### DIFF
--- a/tests/e2e/cluster_downgrade_test.go
+++ b/tests/e2e/cluster_downgrade_test.go
@@ -16,7 +16,9 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -27,6 +29,7 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
+	"go.uber.org/zap"
 )
 
 func TestDowngradeUpgradeClusterOf1(t *testing.T) {
@@ -43,17 +46,24 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 	if !fileutil.Exist(lastReleaseBinary) {
 		t.Skipf("%q does not exist", lastReleaseBinary)
 	}
+
 	currentVersion := semver.New(version.Version)
 	lastVersion := semver.Version{Major: currentVersion.Major, Minor: currentVersion.Minor - 1}
-	currentVersionStr := fmt.Sprintf("%d.%d", currentVersion.Major, currentVersion.Minor)
-	lastVersionStr := fmt.Sprintf("%d.%d", lastVersion.Major, lastVersion.Minor)
+
+	currentVersion.PreRelease = ""
+	currentVersionStr := currentVersion.String()
+	lastVersionStr := lastVersion.String()
 
 	e2e.BeforeTest(t)
 
 	t.Logf("Create cluster with version %s", currentVersionStr)
 	epc := newCluster(t, currentEtcdBinary, clusterSize)
 	for i := 0; i < len(epc.Procs); i++ {
-		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: currentVersionStr, Server: currentVersionStr})
+		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{
+			Cluster: currentVersionStr,
+			Server:  version.Version,
+			Storage: currentVersionStr,
+		})
 	}
 	t.Logf("Cluster created")
 
@@ -62,36 +72,49 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 
 	t.Log("Downgrade enabled, validating if cluster is ready for downgrade")
 	for i := 0; i < len(epc.Procs); i++ {
+		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{
+			Cluster: lastVersionStr,
+			Server:  version.Version,
+			Storage: lastVersionStr,
+		})
 		e2e.AssertProcessLogs(t, epc.Procs[i], "The server is ready to downgrade")
-		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: lastVersionStr, Server: currentVersionStr})
 	}
-	t.Log("Cluster is ready for downgrade")
 
+	t.Log("Cluster is ready for downgrade")
 	t.Logf("Starting downgrade process to %q", lastVersionStr)
 	for i := 0; i < len(epc.Procs); i++ {
 		t.Logf("Downgrading member %d by running %s binary", i, lastReleaseBinary)
 		stopEtcd(t, epc.Procs[i])
 		startEtcd(t, epc.Procs[i], lastReleaseBinary)
 	}
+
 	t.Log("All members downgraded, validating downgrade")
 	e2e.AssertProcessLogs(t, leader(t, epc), "the cluster has been downgraded")
 	for i := 0; i < len(epc.Procs); i++ {
-		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: lastVersionStr, Server: lastVersionStr})
+		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{
+			Cluster: lastVersionStr,
+			Server:  lastVersionStr,
+		})
 	}
-	t.Log("Downgrade complete")
 
+	t.Log("Downgrade complete")
 	t.Logf("Starting upgrade process to %q", currentVersionStr)
 	for i := 0; i < len(epc.Procs); i++ {
 		t.Logf("Upgrading member %d", i)
 		stopEtcd(t, epc.Procs[i])
 		startEtcd(t, epc.Procs[i], currentEtcdBinary)
-		if i+1 < len(epc.Procs) {
-			validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: lastVersionStr, Server: currentVersionStr})
-		}
+		// NOTE: The leader has monitor to the cluster version, which will
+		// update cluster version. We don't need to check the transient
+		// version just in case that it might be flaky.
 	}
+
 	t.Log("All members upgraded, validating upgrade")
 	for i := 0; i < len(epc.Procs); i++ {
-		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: currentVersionStr, Server: currentVersionStr})
+		validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{
+			Cluster: currentVersionStr,
+			Server:  version.Version,
+			Storage: currentVersionStr,
+		})
 	}
 	t.Log("Upgrade complete")
 }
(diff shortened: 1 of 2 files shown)


fuweid:
I think we can add validation about storage version when handle downgrade.
If the storage version is not aligned with cluster version, we should reject the request.

serathius:
> And then we file the downgrade request before next round of monitorStorageVersion(per 4 second), the cluster version will be v3.5.0 which is equal to the UnsafeDetectSchemaVersion's result. And we won't see that The server is ready to downgrade.

Not sure what you mean. I would like to leave checking `The server is ready to downgrade` log as this is the current way user is informed that they can start downgrading members.

tests/e2e/cluster_downgrade_test.go
```diff
@@ -45,15 +48,19 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 	}
 	currentVersion := semver.New(version.Version)
 	lastVersion := semver.Version{Major: currentVersion.Major, Minor: currentVersion.Minor - 1}
-	currentVersionStr := fmt.Sprintf("%d.%d", currentVersion.Major, currentVersion.Minor)
-	lastVersionStr := fmt.Sprintf("%d.%d", lastVersion.Major, lastVersion.Minor)
+	currentVersionStr := fmt.Sprintf("%d.%d.%d", currentVersion.Major, currentVersion.Minor, currentVersion.Patch)
```
serathius, line 51: Isn't that just a currentVersion.String() now?

tests/e2e/cluster_downgrade_test.go
```diff
@@ -45,15 +48,19 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
...
+	lastVersionStr := fmt.Sprintf("%d.%d.0", lastVersion.Major, lastVersion.Minor)
```
serathius, line 52: Isn't that just lastVersion.String() now?

tests/e2e/cluster_downgrade_test.go
```diff
@@ -62,36 +69,50 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 
 	t.Log("Downgrade enabled, validating if cluster is ready for downgrade")
 	for i := 0; i < len(epc.Procs); i++ {
-		e2e.AssertProcessLogs(t, epc.Procs[i], "The server is ready to downgrade")
```
serathius, line 65: Would like to leave checking this log. Can you clarify why you removed it?

tests/e2e/cluster_downgrade_test.go
```diff
@@ -62,36 +69,50 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
...
+	t.Log("Downgrade complete")
 	t.Logf("Starting upgrade process to %q", currentVersionStr)
 	for i := 0; i < len(epc.Procs); i++ {
 		t.Logf("Upgrading member %d", i)
 		stopEtcd(t, epc.Procs[i])
 		startEtcd(t, epc.Procs[i], currentEtcdBinary)
-		if i+1 < len(epc.Procs) {
-			validateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{Cluster: lastVersionStr, Server: currentVersionStr})
-		}
+		// NOTE: The leader has monitor to the cluster version, which will
```
serathius, line 104: SG, feel free to remove the log.

tests/framework/e2e/util.go
```diff
@@ -81,6 +81,25 @@ func SpawnWithExpectLines(ctx context.Context, args []string, envVars map[string
 	return lines, perr
 }
 
+// SpawnWithExpectResult is used to verify the output from short-running commands.
+func SpawnWithExpectResult(args []string, envVars map[string]string, f func([]string) error) error {
```
serathius, line 85: Don't think we need a `f func([]string) error` argument, we can just return string from function.

fuweid:
> > And then we file the downgrade request before next round of monitorStorageVersion(per 4 second), the cluster version will be v3.5.0 which is equal to the UnsafeDetectSchemaVersion's result. And we won't see that The server is ready to downgrade.
> 
> Not sure what you mean. I would like to leave checking `The server is ready to downgrade` log as this is the current way user is informed that they can start downgrading members.

Hi @serathius, thanks for the review! Let me explain it.

when the new cluster starts and before lead election, the client can only get the `not_decided` cluster version and `unknown` storage version [codelink][1].
The [monitorClusterVersions][2] will initialize the cluster version with v3.6.0, which will trigger the [monitorStorageVersion][3] to update the storage version.

The `monitorStorageVersion` updates storage version only if the storage version is nil or it is not aligned with cluster version.
If the storage backend doesn't commit `confState` in time, the `monitorStorageVersion` won't update storage version and it is still be `unknown` via `/version` http api.
Since the `monitorStorageVersion` does update periodically, the storage version will be correct eventually.

So, the client might see three values before storage version has been finilized: `unknown`, `v3.5.0` and `v3.6.0`.
For the value `v3.5.0`, it happens after `confState` committed and before `monitorStorageVersion` updates [codelink][4].
It is intermediate state.

https://github.com/etcd-io/etcd/blob/a1018dbddf22c8aebfb5ae348a55440b266aa6e8/server/storage/schema/schema.go#L93-L107

The storage version becomes `v3.6.0` only if the cluster version is `v3.6.0`.
If the cluster version has been updated to be `v3.5.0` by Downgrade request, and the storage version is still being `v3.5.0`, the `monitorStorageVersion` won't update the storage version,
because the storage version is equal to the cluster version. So we won't see the log `The server is ready to downgrade`.

https://github.com/etcd-io/etcd/blob/a1018dbddf22c8aebfb5ae348a55440b266aa6e8/server/etcdserver/version/monitor.go#L106-L131


The original test case doesn't wait for the storage version ready. So I check the storage version to make sure that everything is ready.
I was thinking that it is hard to check server log to make sure that the Downgrade is ready. Use `/version` api is easy for the operator.
Does it make senses?

[1]: <https://github.com/etcd-io/etcd/blob/a1018dbddf22c8aebfb5ae348a55440b266aa6e8/server/etcdserver/api/etcdhttp/version.go#L38>
[2]: <https://github.com/etcd-io/etcd/blob/a1018dbddf22c8aebfb5ae348a55440b266aa6e8/server/etcdserver/server.go#L2160>
[3]: <https://github.com/etcd-io/etcd/blob/a1018dbddf22c8aebfb5ae348a55440b266aa6e8/server/etcdserver/server.go#L2182>
[4]: <https://github.com/etcd-io/etcd/blob/a1018dbddf22c8aebfb5ae348a55440b266aa6e8/server/etcdserver/server.go#L2147>

tests/e2e/cluster_downgrade_test.go
fuweid: Please check this reply https://github.com/etcd-io/etcd/pull/14657#issuecomment-1296493139 Thanks!

tests/e2e/cluster_downgrade_test.go
fuweid: The cluster version is like `v3.6.0` instead of etcd server version `v3.6.0-alpha0`.
So I go in this way. 😂

serathius:
Nether this test nor user should start downgrade before cluster version was figured out during cluster bootstrap. We should update the test and wait for cluster version to be equal v3.6.

> I was thinking that it is hard to check server log to make sure that the Downgrade is ready. Use /version api is easy for the operator.
I can agree with that, however the process should also be visible for log so administrator that is sshed to know when they can start replacing nodes.

fuweid:
> Nether this test nor user should start downgrade before cluster version was figured out during cluster bootstrap. We should update the test and wait for cluster version to be equal v3.6.
> 
> 
Yes. And we should also wait for storage version to be equal to v3.6. This commit is used to deflake the case.

tests/framework/e2e/util.go
fuweid: It is hard to verify the json in one query. If follow current pattern, we need to send three requests and verify it separately. It can't guarantee that cluster version won't change when we verify the storage version.

tjungblu:
> the brand-new cluster is using simple-config-changer, which means that entries has been committed before leader election and these entries will be applied when etcdserver starts to receive apply-requests. 

@fuweid good catch, I've thought there's a race condition somewhere. Thanks for digging this up.

tests/e2e/cluster_downgrade_test.go
serathius: could we then remove internal version `-alpha.0` from `currentVersion` instead?

fuweid: Updated. PTAL

tests/e2e/cluster_downgrade_test.go
fuweid: Added it back. PTAL

tests/e2e/cluster_downgrade_test.go
fuweid: Done.

tests/framework/e2e/util.go
fuweid: Sorry for misundestanding. The suggestions looks good to me. Updated. PTAL.

fuweid:
@serathius Sorry for the late reply. The code has been updated. PTAL.

tests/e2e/cluster_downgrade_test.go
```diff
@@ -140,30 +163,36 @@ func stopEtcd(t *testing.T, ep e2e.EtcdProcess) {
...
 
 func validateVersion(t *testing.T, cfg *e2e.EtcdProcessClusterConfig, member e2e.EtcdProcess, expect version.Versions) {
-	// Two separate calls to expect as it doesn't support multiple matches on the same line
-	var err error
-	testutils.ExecuteWithTimeout(t, 20*time.Second, func() {
+	testutils.ExecuteWithTimeout(t, 30*time.Second, func() {
 		for {
-			if expect.Server != "" {
-				err = e2e.SpawnWithExpects(e2e.CURLPrefixArgs(cfg, member, "GET", e2e.CURLReq{Endpoint: "/version"}), nil, `"etcdserver":"`+expect.Server)
+			err := func() error {
```
serathius, line 168: Please remove creating anonymous function. There is no need for it if we immidietly call it.

serathius (approved):
LGTM

tests/e2e/cluster_downgrade_test.go
fuweid: The `func` is used to unify the `time.Sleep` statement. If not, there are several `time.Sleep` if err != nil, like

```
if expect.Cluster != result.Cluster {
    time.Sleep(time.Second)
    continue
}
if expect.Server != result.Server {
    time.Sleep(time.Second)
    continue
} 
...
```

Just wonder that there is any reason to prevent this usage? :joy: Thanks!

serathius: I would say it obfuscates code. It's better to be explicit than implicit. There is nothing wrong with repetition.

serathius: Better alternative would be to change anonymous function to normal function that takes `expect` and `result` versions.

fuweid:
reopen to trigger the CI

Flake case: https://github.com/etcd-io/etcd/actions/runs/3378570366/jobs/5608948355

serathius:
Thanks for looking into this. Downgrades are important feature for v3.6 so it's great to see fixes here.

PR #14660, opened 2022-10-31
tests: TestRunner.NewCluster accepts options

clarkfw:
As mentioned in https://github.com/etcd-io/etcd/pull/14632#pullrequestreview-1158400017 @serathius, `TestRunner.NewCluster` method now accepts options.

This implementation avoids adding a new method to `TestRunner` interface. The downside is that
```
clus := testRunner.NewCluster(ctx, t, tc.config)
```
becomes more verbose as shown below:
```
clus := testRunner.NewCluster(ctx, t, config.WithClusterConfig(tc.config))
```


serathius (approved):

ahrtr (approved):
LGTM

Thank you @clarkfw

PR #14661, opened 2022-10-31
[3.4] fix: close maintenance conn

falser101:
Signed-off-by: jianfei.zhang <jianfei.zhang@daocloud.io>

### DIFF
--- a/clientv3/maintenance.go
+++ b/clientv3/maintenance.go
@@ -81,6 +81,7 @@ func NewMaintenance(c *Client) Maintenance {
 		dial: func(endpoint string) (pb.MaintenanceClient, func(), error) {
 			conn, err := c.Dial(endpoint)
 			if err != nil {
+				conn.Close()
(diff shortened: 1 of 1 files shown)


falser101:
@ahrtr

clientv3/maintenance.go
```diff
@@ -81,6 +81,7 @@ func NewMaintenance(c *Client) Maintenance {
 		dial: func(endpoint string) (pb.MaintenanceClient, func(), error) {
 			conn, err := c.Dial(endpoint)
 			if err != nil {
+				conn.Close()
```
ahrtr, line 84: When `err` isn't nil, then `conn` should be nil. So it will run into nil pointer panic. 

It seems that 3.4 has different implementation, and we don't need to backport the PR to 3.4.

ahrtr:
It seems that we don't need to fix it for 3.4.  Thanks anyway.

PR #14663, opened 2022-10-31
auth: check permission when performing maintainenence operations if auth is enabled

ahrtr:
When auth is enabled, etcd doesn't check permission for some important operations, such as downgrade, move-leader... at all.  The existing behavior isn't consistent, we check admin permission for some operations (e.g. HashKV, snapshot, defragment), but not for others. I think only admin (users with `root` role) can perform such operations. 

I think the following operations require admin permission,
1. all maintenance operations;
2. all lease operations;

The only exception is `memberList` operation, there is no check on this API at all. Should we require admin permission to perform `MemberList` operation? 

cc @mitake @ptabor  @spzala @serathius 

### DIFF
--- a/server/etcdserver/api/v3rpc/auth.go
+++ b/server/etcdserver/api/v3rpc/auth.go
@@ -18,6 +18,7 @@ import (
 	"context"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/server/v3/auth"
 	"go.etcd.io/etcd/server/v3/etcdserver"
 )
 
@@ -163,4 +164,24 @@ func (as *AuthServer) UserChangePassword(ctx context.Context, r *pb.AuthUserChan
 		return nil, togRPCError(err)
 	}
 	return resp, nil
+}
+
+type AuthGetter interface {
+	AuthInfoFromCtx(ctx context.Context) (*auth.AuthInfo, error)
+	AuthStore() auth.AuthStore
+}
+
+type AuthAdmin struct {
+	ag AuthGetter
+}
+
+// isPermitted verifies the user has admin privilege.
+// Only users with "root" role are permitted.
+func (aa *AuthAdmin) isPermitted(ctx context.Context) error {
+	authInfo, err := aa.ag.AuthInfoFromCtx(ctx)
+	if err != nil {
+		return err
+	}
+
+	return aa.ag.AuthStore().IsAdminPermitted(authInfo)
 }
--- a/client/v3/maintenance.go
+++ b/client/v3/maintenance.go
@@ -119,19 +119,7 @@ func NewMaintenance(c *Client) Maintenance {
 				return nil, nil, fmt.Errorf("failed to dial endpoint %s with maintenance client: %v", endpoint, err)
 			}
 
-			//get token with established connection
-			dctx := c.ctx
-			cancel := func() {}
-			if c.cfg.DialTimeout > 0 {
-				dctx, cancel = context.WithTimeout(c.ctx, c.cfg.DialTimeout)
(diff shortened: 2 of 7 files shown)


ahrtr:
It might make more sense to check the permission during applying stage for `lease` operation, because we need to make sure the local permission data is in sync.

@mitake  Do you have a overall design & thought on this?

mitake:
Thanks @ahrtr , I think the maintenance operations should be protected.
In the case of lease, the apply layer perform put permission like this: https://github.com/etcd-io/etcd/blob/main/server/etcdserver/apply/apply_auth.go#L125
For revoking, the revoke logic uses a temporal permission of root like this: https://github.com/etcd-io/etcd/blob/main/server/etcdserver/server.go#L877 This is for making the design of lease simple and we can have an assumption that lease revoking can happen if the above apply allows it.

server/etcdserver/api/v3rpc/auth.go
```diff
@@ -164,3 +165,21 @@ func (as *AuthServer) UserChangePassword(ctx context.Context, r *pb.AuthUserChan
...
+type AuthGetter interface {
+	AuthInfoFromCtx(ctx context.Context) (*auth.AuthInfo, error)
+	AuthStore() auth.AuthStore
+}
+
+type AuthChecker struct {
+	ag AuthGetter
+}
+
+func (achk *AuthChecker) isAuthenticated(ctx context.Context) error {
```
serathius, line 178: `Authentication` is not `Authorization`. 
```suggestion
func (achk *AuthChecker) isAuthorized(ctx context.Context) error {
```

server/etcdserver/api/v3rpc/auth.go
```diff
@@ -164,3 +165,21 @@ func (as *AuthServer) UserChangePassword(ctx context.Context, r *pb.AuthUserChan
 	}
 	return resp, nil
 }
+
+type AuthGetter interface {
+	AuthInfoFromCtx(ctx context.Context) (*auth.AuthInfo, error)
+	AuthStore() auth.AuthStore
+}
+
+type AuthChecker struct {
```
serathius, line 174:
```suggestion
type RequireAdmin struct {
```
This auth checker ony checks admin permission.

ahrtr:
Thanks @mitake for the feedback. Let's do it step by step.  In the first step, this PR protects all [maintenance API](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/maintenance.go). This PR is ready to review.  Please take a look. I suggest to review it commit by commit. Thanks.

serathius:
I think that requiring admin for maintenance makes sense as they are not expected to ever be executed by application using etcd API. Leases however are used for leader election and locking so we cannot require admin for it. 

I think that we should introduce authorization for leases, however we need a proper design. As this was never brought up by etcd users I don't think it should be prioritized.

ahrtr:
> Leases however are used for leader election and locking so we cannot require admin for it.
> 
> I think that we should introduce authorization for leases, however we need a proper design. As this was never brought up by etcd users I don't think it should be prioritized.

Thanks for the feedback. We can discuss lease in separate thread. Since K8s heavily depends on lease, so we should be very cautious about it.

mitake:
@ahrtr I think overall changes seem to be good, but let me check the client side change again later. I want to make sure that retry interceptors are functioning as expected.

ahrtr:
Thanks @mitake for the response. I just rebased this PR, and added one more commit (the last one) https://github.com/etcd-io/etcd/pull/14663/commits/74085136b38fc90d71b1e5843fd6c5668393e1e5

mitake:
Thanks a lot @ahrtr, LGTM

PR #14664, opened 2022-10-31
ci: ensure the generated code is up-to-date

spacewander:
See https://github.com/etcd-io/etcd/pull/14612#issue-1419792069


scripts/verify_genproto.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# This scripts is automatically run by CI to prevent pull requests missing running genproto.sh
+# after changing *.proto file.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+./scripts/genproto.sh
+diff=$(git diff --name-only | grep -c ".pb.")
```
serathius, line 10: This implementation checks if there is any uncommitted proto files, this should work for CI, but not for local development. 
I think this should be ok, but maybe you can figure out if we could output `./scripts/genproto.sh` to tmp directory and compare it's contents to local files.

scripts/verify_genproto.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ "$diff" -eq 0 ]; then
+  echo "PASSED genproto-verification!"
+  exit 0
+fi
+echo "Failed genproto-verification!" >&2
+echo "* Found changed files $(git diff --name-only | grep '.pb.')" >&2
+echo "* Please rerun genproto.sh after changing *.proto file" >&2
+echo "* Run ./scripts/genproto.sh" >&2
```
serathius, line 18: This recommendation only makes sense when command is run in CI. 
When running locally user needs to just commit the changes.

I think we should explain that user should run `./scripts/genproto.sh` only if the commend was run in CI and then commit the changes.

serathius:
Thanks for looking into this!

scripts/verify_genproto.sh
```diff
@@ -0,0 +1,26 @@
...
+# after changing *.proto file.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+tmpWorkDir=$(mktemp -d -t 'twd.XXXXXX')
+mkdir "$tmpWorkDir/etcd"
+tmpWorkDir="$tmpWorkDir/etcd"
+cp -r . "$tmpWorkDir"
```
serathius, line 12: Coping whole repo seems like a lot of work. Could we maybe just copy `.proto` files?

scripts/verify_genproto.sh
```diff
@@ -0,0 +1,26 @@
...
+pushd "$tmpWorkDir"
+git add -A && git commit -m init
+./scripts/genproto.sh
+diff=$(git diff --numstat | awk '{print $3}')
```
serathius, line 16: This script will still compare generated proto versus proto committed, so there is no difference from older script.

scripts/verify_genproto.sh
spacewander: I tried it, but the genproto.sh requires too many files (not just `*.proto`).

scripts/verify_genproto.sh
spacewander: @serathius 
Sorry for affecting your review. I just force-push a new commit that handles a case that `git commit` will fail.

scripts/verify_genproto.sh
serathius: Could we then simplify genproto.sh so it can be run on non local directory?

scripts/verify_genproto.sh
serathius: I'm little worried about `git` shenanigans (`add` `commit`), it makes the code very fragile. 

Have you considered avoiding them? I will not push too much as overall change should be beneficial, just worried about over-complicating `verify_genproto.sh` if we could maybe simplify `genproto.sh`

scripts/verify_genproto.sh
spacewander: It is not so easy, as the genproto.sh requires some Go files, extra shell scripts in the scripts directory, the Documentation directory, and so on.

scripts/verify_genproto.sh
spacewander: I have tried using regular diff. But the genproto.sh will generate some files which are ignored by .gitignore. And as regular diff doesn't support .gitignore, we have to maintain an additional ignore list...

scripts/verify_genproto.sh
serathius: Ok, let's leave for future.

scripts/verify_genproto.sh
serathius: Ok, I think we are good enough. We can iterate on the code.

serathius (approved):
Thanks for looking into this. cc @ahrtr @spzala @ptabor

scripts/verify_proto_annotations.sh
```diff
@@ -15,6 +15,6 @@ fi
 echo "Failed proto-annotations-verification!" >&2
 echo "If you are adding new proto fields/messages that will be included in raft log:" >&2
 echo "* Please add etcd_version annotation in *.proto file with next etcd version" >&2
-echo "* Run ./scripts/getproto.sh" >&2
+echo "* Run ./scripts/genproto.sh" >&2
```
ahrtr, line 18: nice, good catch!

ahrtr:
Overall looks good to me. Just as I mentioned in https://github.com/etcd-io/etcd/pull/14612#issuecomment-1289468108, the reason is we do not fix the versions of some binaries. 

We might need to revisit this after https://github.com/etcd-io/etcd/issues/14533 if resolved.

ahrtr (approved):
Looks good for now. 

Thank yo @spacewander

PR #14669, opened 2022-11-01
.github: make govuln-check generic

vivekpatani:
Signed-off-by: vivekpatani <9080894+vivekpatani@users.noreply.github.com>

### DIFF
--- a/.github/workflows/govuln.yaml
+++ b/.github/workflows/govuln.yaml
@@ -3,24 +3,10 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        target:
-          - linux-amd64-govuln-check
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
           go-version: "1.19.3"
       - run: date
-      - env:
-          TARGET: ${{ matrix.target }}
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/govuln.yaml
```diff
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         target:
-          - linux-amd64-govuln-check
+          - govuln-check
```
ahrtr, line 10: If you look at some other examples, such as [static-analysis.yaml](https://github.com/etcd-io/etcd/blob/main/.github/workflows/static-analysis.yaml), there is no target at all. 

It isn't a big deal to to me. Keeping it as it is or fixing it, either way works for me.

serathius: I would prefer to go with simplicity.

ahrtr: @vivekpatani  Please remove the target so as to make it as simple as possible. So that we can merge this PR.

ahrtr: @vivekpatani please resolve this comment so that we can merge this PR. It should be a simple change.

vivekpatani: @serathius @ahrtr removed everything but `run: date`, can remove that too. Let me know your thoughts. Sorry for the delay.

ahrtr (approved):
LGTM

PR #5056, opened 2016-04-13
pkg/expect, e2e: support sending Signals to expect process, test etcdctl lock

heyitsanthony:
Trying to figure out what is going on with pids and signals on CI.

PR #14672, opened 2022-11-01
Expect exit code enhancement

tjungblu:
ExpectProcess and ExpectFunc now take the exit code of the process into account, not just the matching of the tty output.

I'm still fixing some of the tests that are now failing due to this.

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -19,6 +19,7 @@ package expect
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -33,6 +34,10 @@ import (
 
 const DEBUG_LINES_TAIL = 40
 
+var (
+	ErrProcessRunning = fmt.Errorf("process is still running")
+)
+
 type ExpectProcess struct {
 	cfg expectConfig
 
@@ -40,11 +45,12 @@ type ExpectProcess struct {
 	fpty *os.File
 	wg   sync.WaitGroup
 
-	mu    sync.Mutex // protects lines and err
-	lines []string
-	count int // increment whenever new line gets added
-	cur   int // current read position
-	err   error
+	mu       sync.Mutex // protects lines, count, cur, exitErr and exitCode
+	lines    []string
+	count    int   // increment whenever new line gets added
+	cur      int   // current read position
+	exitErr  error // process exit error
+	exitCode int
 }
 
 // NewExpect creates a new process for expect testing.
@@ -69,8 +75,9 @@ func NewExpectWithEnv(name string, args []string, env []string, serverProcessCon
 		return nil, err
 	}
 
-	ep.wg.Add(1)
+	ep.wg.Add(2)
 	go ep.read()
+	go ep.waitSaveExitErr()
 	return ep, nil
 }
 
@@ -95,46 +102,83 @@ func (ep *ExpectProcess) Pid() int {
 
 func (ep *ExpectProcess) read() {
 	defer ep.wg.Done()
-	printDebugLines := os.Getenv("EXPECT_DEBUG") != ""
+	defer func(fpty *os.File) {
+		err := fpty.Close()
+		if err != nil {
+			// we deliberately only log the error here, closing the PTY should mostly be (expected) broken pipes
+			fmt.Printf("error while closing fpty: %v", err)
+		}
+	}(ep.fpty)
+
 	r := bufio.NewReader(ep.fpty)
 	for {
-		l, err := r.ReadString('\n')
-		ep.mu.Lock()
-		if l != "" {
-			if printDebugLines {
-				fmt.Printf("%s (%s) (%d): %s", ep.cmd.Path, ep.cfg.name, ep.cmd.Process.Pid, l)
-			}
-			ep.lines = append(ep.lines, l)
-			ep.count++
-		}
+		err := ep.tryReadNextLine(r)
 		if err != nil {
-			ep.err = err
-			ep.mu.Unlock()
 			break
 		}
-		ep.mu.Unlock()
+	}
+}
+
+func (ep *ExpectProcess) tryReadNextLine(r *bufio.Reader) error {
+	printDebugLines := os.Getenv("EXPECT_DEBUG") != ""
+	l, err := r.ReadString('\n')
+
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	if l != "" {
+		if printDebugLines {
+			fmt.Printf("%s (%s) (%d): %s", ep.cmd.Path, ep.cfg.name, ep.cmd.Process.Pid, l)
+		}
+		ep.lines = append(ep.lines, l)
+		ep.count++
+	}
+
+	// we're checking the error here at the bottom to ensure any leftover reads are still taken into account
+	return err
+}
+
+func (ep *ExpectProcess) waitSaveExitErr() {
+	defer ep.wg.Done()
+	err := ep.waitProcess()
+
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+	if err != nil {
+		ep.exitErr = err
 	}
 }
 
 // ExpectFunc returns the first line satisfying the function f.
 func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (string, error) {
 	i := 0
-
 	for {
-		ep.mu.Lock()
-		for i < len(ep.lines) {
-			line := ep.lines[i]
-			i++
-			if f(line) {
-				ep.mu.Unlock()
-				return line, nil
+		line, errsFound := func() (string, bool) {
+			ep.mu.Lock()
+			defer ep.mu.Unlock()
+
+			// check if this expect has been already closed
+			if ep.cmd == nil {
+				return "", true
 			}
+
+			for i < len(ep.lines) {
+				line := ep.lines[i]
+				i++
+				if f(line) {
+					return line, false
+				}
+			}
+			return "", ep.exitErr != nil
+		}()
+
+		if line != "" {
+			return line, nil
 		}
-		if ep.err != nil {
-			ep.mu.Unlock()
+
+		if errsFound {
 			break
 		}
-		ep.mu.Unlock()
 
 		select {
 		case <-ctx.Done():
@@ -143,16 +187,18 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 			// continue loop
 		}
 	}
+
 	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
 	lastLinesIndex := len(ep.lines) - DEBUG_LINES_TAIL
 	if lastLinesIndex < 0 {
 		lastLinesIndex = 0
 	}
 	lastLines := strings.Join(ep.lines[lastLinesIndex:], "")
-	ep.mu.Unlock()
-	return "", fmt.Errorf("match not found."+
-		" Set EXPECT_DEBUG for more info Err: %v, last lines:\n%s",
-		ep.err, lastLines)
+	return "", fmt.Errorf("match not found. "+
+		" Set EXPECT_DEBUG for more info Errs: [%v], last lines:\n%s",
+		ep.exitErr, lastLines)
 }
 
 // ExpectWithContext returns the first line containing the given string.
@@ -174,61 +220,90 @@ func (ep *ExpectProcess) LineCount() int {
 	return ep.count
 }
 
-// Stop kills the expect process and waits for it to exit.
-func (ep *ExpectProcess) Stop() error { return ep.close(true) }
+// ExitCode returns the exit code of this process.
+// If the process is still running, it returns exit code 0 and ErrProcessRunning.
+func (ep *ExpectProcess) ExitCode() (int, error) {
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	if ep.cmd == nil {
+		return ep.exitCode, nil
+	}
+
+	return 0, ErrProcessRunning
+}
+
+// ExitError returns the exit error of this process (if any).
+// If the process is still running, it returns ErrProcessRunning instead.
+func (ep *ExpectProcess) ExitError() error {
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	if ep.cmd == nil {
+		return ep.exitErr
+	}
+
+	return ErrProcessRunning
+}
+
+// Stop signals the process to terminate via SIGTERM
+func (ep *ExpectProcess) Stop() error {
+	err := ep.Signal(syscall.SIGTERM)
+	if err != nil && strings.Contains(err.Error(), "os: process already finished") {
+		return nil
+	}
+	return err
+}
 
 // Signal sends a signal to the expect process
 func (ep *ExpectProcess) Signal(sig os.Signal) error {
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	if ep.cmd == nil {
+		return errors.New("expect process already closed")
+	}
+
 	return ep.cmd.Process.Signal(sig)
 }
 
-func (ep *ExpectProcess) Wait() error {
-	_, err := ep.cmd.Process.Wait()
-	return err
+func (ep *ExpectProcess) waitProcess() error {
+	state, err := ep.cmd.Process.Wait()
+	if err != nil {
+		return err
+	}
+
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+	ep.exitCode = state.ExitCode()
+
+	if !state.Success() {
+		return fmt.Errorf("unexpected exit code [%d] after running [%s]", ep.exitCode, ep.cmd.String())
+	}
+
+	return nil
 }
 
-// Close waits for the expect process to exit.
-// Close currently does not return error if process exited with !=0 status.
-// TODO: Close should expose underlying process failure by default.
-func (ep *ExpectProcess) Close() error { return ep.close(false) }
+// Wait waits for the process to finish.
+func (ep *ExpectProcess) Wait() {
+	ep.wg.Wait()
+}
 
-func (ep *ExpectProcess) close(kill bool) error {
-	if ep.cmd == nil {
-		return ep.err
-	}
-	if kill {
-		ep.Signal(syscall.SIGTERM)
-	}
-
-	err := ep.cmd.Wait()
-	ep.fpty.Close()
+// Close waits for the expect process to exit and return its error.
+func (ep *ExpectProcess) Close() error {
 	ep.wg.Wait()
 
-	if err != nil {
-		if !kill && strings.Contains(err.Error(), "exit status") {
-			// non-zero exit code
-			err = nil
-		} else if kill && strings.Contains(err.Error(), "signal:") {
-			err = nil
-		}
-	}
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
 
+	// this signals to other funcs that the process has finished
 	ep.cmd = nil
-	return err
+	return ep.exitErr
 }
 
 func (ep *ExpectProcess) Send(command string) error {
 	_, err := io.WriteString(ep.fpty, command)
 	return err
-}
-
-func (ep *ExpectProcess) ProcessError() error {
-	if strings.Contains(ep.err.Error(), "input/output error") {
-		// TODO: The expect library should not return
-		// `/dev/ptmx: input/output error` when process just exits.
-		return nil
-	}
-	return ep.err
 }
 
 func (ep *ExpectProcess) Lines() []string {
--- a/tests/e2e/ctl_v3_auth_test.go
+++ b/tests/e2e/ctl_v3_auth_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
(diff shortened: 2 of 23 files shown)


serathius:
Nice, one additional option to consider is returning process status so the caller can decide based on exit code. Example linearizability tests expect process to fail when tailpoint is injected.

tjungblu:
added the status code as well, I think this is going to become a bigger refactor since we there was no exit code assumption beforehand. Some tests still don't run, I also think there's some locking issue. Gotta continue tomorrow.

tests/e2e/ctl_v3_auth_test.go
```diff
@@ -679,11 +680,10 @@ func authTestCertCN(cx ctlCtx) {
 		cx.t.Error(err)
 	}
 
-	// try a non granted key
+	// try a non -ranted key
```
tjungblu, line 683: sorry this was some autocorrect (wtf)

ahrtr:
Thanks for working on this. 

It seems the PR caused lots of workflow failures. Usually 1~2 failures might be caused by flaky test. But 6 failures most likely mean that it's caused by the PR.  Please take a look at the failures.

pkg/expect/expect.go
```diff
@@ -174,63 +213,89 @@ func (ep *ExpectProcess) LineCount() int {
...
+		return errors.New("expect process already closed")
+	}
+
 	return ep.cmd.Process.Signal(sig)
 }
 
 func (ep *ExpectProcess) Wait() error {
-	_, err := ep.cmd.Process.Wait()
-	return err
+	panic("who uses this")
```
tjungblu, line 245: this is just temporary, a compile error regarding Wait() came out of nowhere, just interested what was trying to consume this. There's no interface around it...

tjungblu: Turned out to be a change in main, I didn't know the action would apply this commit on top of main and run the tests on top of it. I've rebased and could fix it.

tjungblu:
@ahrtr @serathius finally managed to get all to green. There are a couple of weird choices I've made around the error assertions, so let me know what you think about it.

pkg/expect/expect.go
```diff
@@ -98,43 +102,76 @@ func (ep *ExpectProcess) read() {
 	printDebugLines := os.Getenv("EXPECT_DEBUG") != ""
 	r := bufio.NewReader(ep.fpty)
 	for {
-		l, err := r.ReadString('\n')
-		ep.mu.Lock()
-		if l != "" {
-			if printDebugLines {
-				fmt.Printf("%s (%s) (%d): %s", ep.cmd.Path, ep.cfg.name, ep.cmd.Process.Pid, l)
+		err := func() error {
```
serathius, line 105: no need for anonymous function

tjungblu: disagree, this makes the locking more obvious in 

```
ep.mu.Lock()
defer ep.mu.Unlock()
```

tests/e2e/ctl_v3_auth_test.go
```diff
@@ -679,11 +675,10 @@ func authTestCertCN(cx ctlCtx) {
...
 	}
 
-	// try a non granted key
+	// try a non-granted key
 	cx.user, cx.pass = "", ""
-	if err := ctlV3PutFailPerm(cx, "baz", "bar"); err != nil {
-		cx.t.Error(err)
-	}
+	err := ctlV3PutFailPerm(cx, "baz", "bar")
+	assert.ErrorContains(cx.t, err, "unexpected exit code [1]")
```
tjungblu, line 681: this should also be a require, I dunno why I picked assert here instead

pkg/expect/expect.go
```diff
@@ -174,63 +213,90 @@ func (ep *ExpectProcess) LineCount() int {
...
+// This will return the correct exit code after Close/Stop has been called, otherwise it returns math.MinInt
+func (ep *ExpectProcess) ExitCode() int {
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	if ep.cmd == nil {
+		return ep.exitCode
+	}
+
+	return math.MinInt
```
serathius, line 226: this is weird decision. would prefer to return an error or some known error code like 255

tjungblu: agreed, will change this

tests/common/defrag_test.go
```diff
@@ -44,8 +45,9 @@ func TestDefragOnline(t *testing.T) {
 			t.Fatalf("defrag_test: compact with revision error (%v)", err)
 		}
 
-		if err = cc.Defragment(ctx, options); err != nil {
-			t.Fatalf("defrag_test: defrag error (%v)", err)
+		err = cc.Defragment(ctx, options)
+		if err != nil {
+			require.ErrorContains(t, err, "Finished defragmenting etcd")
```
serathius, line 50: Why defragment command fails here?

tests/e2e/cluster_downgrade_test.go
```diff
@@ -47,13 +48,19 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 		t.Skipf("%q does not exist", lastReleaseBinary)
 	}
 
-	currentVersion := semver.New(version.Version)
-	lastVersion := semver.Version{Major: currentVersion.Major, Minor: currentVersion.Minor - 1}
+	currentVersion, err := getVersionFromBinary(currentEtcdBinary)
```
serathius, line 51: This looks like unrelated changes from other PR. Please rebase.

tests/e2e/ctl_v3_auth_test.go
```diff
@@ -118,9 +120,8 @@ func authDisableTest(cx ctlCtx) {
 
 	// test-user doesn't have the permission, it must fail
 	cx.user, cx.pass = "test-user", "pass"
-	if err := ctlV3PutFailPerm(cx, "hoo", "bar"); err != nil {
-		cx.t.Fatal(err)
-	}
+	err := ctlV3PutFailPerm(cx, "hoo", "bar")
+	require.ErrorContains(cx.t, err, "permission denied")
```
serathius, line 124: Nice!

tests/common/defrag_test.go
tjungblu: odd, works again. Reverted that.

tests/e2e/cluster_downgrade_test.go
tjungblu: I can split this into another commit, but it was never passing for me locally. Not sure if that works on your box.

tests/e2e/ctl_v3_elect_test.go
```diff
@@ -98,7 +100,7 @@ func testElect(cx ctlCtx) {
 }
 
 // ctlV3Elect creates a elect process with a channel listening for when it wins the election.
-func ctlV3Elect(cx ctlCtx, name, proposal string) (*expect.ExpectProcess, <-chan string, error) {
+func ctlV3Elect(cx ctlCtx, name, proposal string, expectFailure bool) (*expect.ExpectProcess, <-chan string, error) {
```
serathius, line 103: I would prefer to handle failure by caller.

tests/e2e/ctl_v3_move_leader_test.go
```diff
@@ -110,27 +111,34 @@ func testCtlV3MoveLeader(t *testing.T, cfg e2e.EtcdProcessClusterConfig, envVars
 	}
 
 	tests := []struct {
-		eps    []string
-		expect string
+		eps         []string
+		expect      string
+		expectAsErr bool
```
serathius, line 116:
```suggestion
		expectError bool
```

tests/e2e/ctl_v3_elect_test.go
tjungblu: me too, do you have an idea how I can do this from the go routine down there? Otherwise I'd have to add another error channel

tests/e2e/ctl_v3_move_leader_test.go
tjungblu: updated

tests/framework/e2e/etcd_process.go
```diff
@@ -183,7 +180,7 @@ func (ep *EtcdServerProcess) Kill() error {
 }
 
 func (ep *EtcdServerProcess) Wait() error {
-	err := ep.proc.Wait()
+	err := ep.proc.Close()
```
serathius, line 183: This totally changes the result. Don't know any case we would want to change it like this.

tjungblu: good call, it's only used in the failpoints. Seems to work though,

serathius:
For commands that fail authorization it makes sense to introduce `require.ErrorContains(cx.t, err, "permission denied")`, however I don't understand why we change `t.Fatal(err)` to require.ErrorContains(cx.t, err, "unexpected exit code").

tjungblu:
Let me go through all of them again and replace them with something more meaningful. There were a bunch of cases where the process was killed and the return was non-zero with no meaningful output.

tests/e2e/ctl_v3_auth_test.go
```diff
@@ -984,14 +975,12 @@ func authTestWatch(cx ctlCtx) {
...
-			err = ctlV3WatchFailPerm(cx, tt.args)
-		}
-
-		if err != nil {
-			if cx.dialTimeout > 0 && !isGRPCTimedout(err) {
+			if err != nil && cx.dialTimeout > 0 && !isGRPCTimedout(err) {
 				cx.t.Errorf("watchTest #%d: ctlV3Watch error (%v)", i, err)
 			}
+		} else {
+			err = ctlV3WatchFailPerm(cx, tt.args)
```
tjungblu, line 982: @serathius here is one of such cases where no meaningful error output comes due to the watch being cancelled:

>     ctl_v3_auth_test.go:982: [unexpected exit code [5] after running [/home/tjungblu/git/etcd/bin/etcdctl --endpoints=http://localhost:20000 --dial-timeout=7s --user=test-user:pass watch key5 --rev 1]]

serathius:
>  There were a bunch of cases where the process was killed and the return was non-zero with no meaningful output.

It should be clear on caller side that the process was killed and is expected to exit with specific exit code. Processes killed by signal should return specific exit codes.

tests/e2e/ctl_v3_elect_test.go
```diff
@@ -74,8 +75,9 @@ func testElect(cx ctlCtx) {
 	if err = blocked.Signal(os.Interrupt); err != nil {
 		cx.t.Fatal(err)
 	}
-	if err = e2e.CloseWithTimeout(blocked, time.Second); err != nil {
-		cx.t.Fatal(err)
+	err = e2e.CloseWithTimeout(blocked, time.Second)
+	if err != nil {
+		require.ErrorContains(cx.t, err, "unexpected exit code")
```
tjungblu, line 80: @serathius that's another candidate:
```
[unexpected exit code [1] after running [/home/tjungblu/git/etcd/bin/etcdctl --endpoints=http://localhost:20000 --dial-timeout=7s elect a p2]]
```

tjungblu: actually `CloseWithTimeout` also looks a little more difficult than it needs to be.

tjungblu:
@serathius there are two odd cases where it could either be canceled, killed or successful using `CloseWithTimeout` on interrupted processes (`testElect` + `testLock`):

https://github.com/etcd-io/etcd/blob/7ed4eda4c19ca2f1eeb461d7abb7aee6a244f7a7/tests/framework/e2e/util.go#L112-L124

I've specifically commented those cases, but let me know if you think there's a better way to handle this. The rest should be cleaned up properly now.

TBH I'm also not entirely sure what these tests are supposed to test?
Introduced in https://github.com/etcd-io/etcd/pull/5059, kinda duped to https://github.com/etcd-io/etcd/pull/5056

pkg/expect/expect.go
serathius: Sounds good.

tests/e2e/ctl_v3_auth_test.go
serathius: Makes sense.

tests/e2e/ctl_v3_elect_test.go
serathius: Seems ok for now.

tests/framework/e2e/etcd_process.go
```diff
@@ -101,9 +101,6 @@ func (ep *EtcdServerProcess) EndpointsMetrics() []string { return []string{ep.cf
 
 func (ep *EtcdServerProcess) Start(ctx context.Context) error {
 	ep.donec = make(chan struct{})
-	if ep.proc != nil {
```
serathius, line 104: Why remove this?

tests/framework/e2e/etcd_process.go
serathius: It might work, but it causes a race between previously sent SIGKILL signal and process close. Please revert.

tests/framework/e2e/util.go
```diff
@@ -50,7 +50,10 @@ func SpawnWithExpectWithEnv(args []string, envVars map[string]string, expected s
 }
 
 func SpawnWithExpects(args []string, envVars map[string]string, xs ...string) error {
-	_, err := SpawnWithExpectLines(context.TODO(), args, envVars, xs...)
+	ctx, cancel := context.WithTimeout(context.TODO(), 2*time.Second)
```
serathius, line 53: There is no expectation that this commands should only take 2 seconds. I expect because this function is mostly used for running `etcdctl` it work, but the the name of the function do not suggest limited run time. 

If limiting run time is required for tests to pass, please move it up the call stack.

tests/e2e/cluster_downgrade_test.go
serathius: I expect that after rebase those changes will not be needed/present.

pkg/expect/expect.go
```diff
@@ -69,7 +75,8 @@ func NewExpectWithEnv(name string, args []string, env []string, serverProcessCon
 		return nil, err
 	}
 
-	ep.wg.Add(1)
+	ep.wg.Add(2)
```
serathius, line 78: Waiting for `waitSaveExitErr` changes the behavior of close, now it will always wait till process exit. I think we should keep `Wait` and `Close` separate.
* `Close` should be used to ensure cleanup, even if process doesn't close by itself. To ensure process exists by itself it should send SIGTERM and wait couple of seconds for process to exit. Process should be expected to exist with success.
* `Wait` should expect the process to exit by itself and just wait till that moment.

tests/framework/e2e/etcd_process.go
serathius: Reason is that you changed implementation of `Close` to always wait till the process finishes. This however not correct behavior.

tests/framework/e2e/util.go
```diff
@@ -74,26 +77,28 @@ func SpawnWithExpectLines(ctx context.Context, args []string, envVars map[string
...
 
 func RunUtilCompletion(args []string, envVars map[string]string) ([]string, error) {
 	proc, err := SpawnCmd(args, envVars)
 	if err != nil {
-		return nil, fmt.Errorf("failed to spawn command: %w", err)
+		return nil, fmt.Errorf("failed to spawn command %v with error: %w", args, err)
 	}
-	defer proc.Stop()
 
-	perr := proc.Wait()
```
serathius, line 91: We should wait for process to exit by itself.

serathius:
Great work overall. I expect this will be crucial to reduce flakiness of e2e tests. One think I would like to fix before we merge is separation between `Wait` and `Close`. They should behave like their process level counterparts.

tests/e2e/cluster_downgrade_test.go
tjungblu: I don't understand, the version inconsistency I run into still exists in main:
https://github.com/etcd-io/etcd/blob/main/tests/e2e/cluster_downgrade_test.go#L50-L55

The issue stems from the fact that the logic takes what is assumed to be in the binary, not what is actually reported by it. If you check the diff, it's already against the latest revision of the test.

pkg/expect/expect.go
tjungblu: So you always have to run `Close` after `Wait`? You can run `close` without `wait` however.

I'll check this out, but I'm sure we have to rewrite a few more tests for this.

tests/framework/e2e/util.go
tjungblu: removed again

tests/framework/e2e/etcd_process.go
tjungblu: added it back

tests/e2e/cluster_downgrade_test.go
serathius: Can you create a separate PR to fix the downgrade test?

pkg/expect/expect.go
serathius: > So you always have to run Close after Wait?

Don't think so, process exits if called by any of them. The main difference is that Wait wait for process to exit normally and Close should always guarantee the exit by sending termination signal.

tjungblu: hmm. What do you think about the following semantics:
* `Wait` just waits for the process to exit
* `Close` waits for the process to exit + cleanup (required call)
* `Stop` signals SIGTERM to the process

> Close should always guarantee the exit by sending termination signal.

So `Close` should always send SIGTERM?

pkg/expect/expect.go
```diff
@@ -174,63 +216,90 @@ func (ep *ExpectProcess) LineCount() int {
...
-			// non-zero exit code
-			err = nil
-		} else if kill && strings.Contains(err.Error(), "signal:") {
-			err = nil
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	if err := ep.fpty.Close(); err != nil {
+		if !strings.Contains(err.Error(), "file already closed") {
+			ep.errs = append(ep.errs, err)
```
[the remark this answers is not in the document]
tjungblu: I was a bit on the fence when I moved to a slice of errors here, the reason is that we shouldn't really be hiding any errors as this makes CI debugging a lot more difficult. 
As you can see I ended up filtering all relevant io error messages anyway. Next commit will go back to a single error.

tests/e2e/cluster_downgrade_test.go
tjungblu: sure, I'll pull this out of that PR.

tjungblu: created it: https://github.com/etcd-io/etcd/pull/14710

pkg/expect/expect.go
serathius: I think `Close` should work like standard linux process close procedure. You first ask nicely for process to exit by sending `SIGTERM` and then wait until process exits. If it doesn't exit within allotted time (e.g. 10s), you kill it using `SIGKILL`.

pkg/expect/expect.go
```diff
@@ -174,63 +216,94 @@ func (ep *ExpectProcess) LineCount() int {
...
+}
+
+// Close waits for the expect process to exit and return its error.
+func (ep *ExpectProcess) Close() error {
+	ep.wg.Wait()
+
+	ep.mu.Lock()
+	defer ep.mu.Unlock()
+
+	_ = ep.fpty.Close()
```
serathius, line 295: I think it might be useful to at least log the error.

pkg/expect/expect.go
serathius: Feel free to disregard by comment. Looking at current implementation of `etcd_process.go` I think what you proposed is ok.

serathius (approved):
LGTM, this is a big change so lets get also feedback from other maintainers.
cc @ahrtr @spzala @ptabor

tjungblu:
Thanks Marek, I'm still fixing the remainder of the tests that are failing due to the changes

pkg/expect/expect.go
tjungblu: added

pkg/expect/expect.go
tjungblu: okay cool, I think we should still add what you proposed to `close` as there is a risk it can hang waiting. It could be a good follow-up PR.

serathius: sg

serathius:
> oh wow, almost all green. 

Great job!

> The last check seems to be an unrelated flake:

Rerun the failed scenario

> I'm just testing whether the e2e test got a lot longer than before, otherwise this is good to go from my side.

Correctness here is more important then time. Let's revisit it when time becomes a problem.

tjungblu:
Thanks for the rerun. It's almost the same runtime on e2e on my local box. I'll keep an eye on the execution times on the workflow over the next couple days.

pkg/expect/expect.go
```diff
@@ -95,46 +102,87 @@ func (ep *ExpectProcess) Pid() int {
...
+		ep.lines = append(ep.lines, l)
+		ep.count++
+	}
+
+	// we're checking the error here at the bottom to ensure any leftover reads are still taken into account
+	if err != nil {
+		return err
+	}
+
+	return nil
```
ahrtr, line 142:
```suggestion
	return err
```

tests/framework/e2e/cluster.go
```diff
@@ -639,18 +638,23 @@ func (epc *EtcdProcessCluster) CloseProc(ctx context.Context, finder func(EtcdPr
...
 		}
-		break
+
+		time.Sleep(500 * time.Millisecond)
 	}
-	if err != nil {
-		return fmt.Errorf("failed to remove member: %w", err)
+
+	if !memberRemoved {
+		return fmt.Errorf("failed to remove member after 10 tries")
```
ahrtr, line 653: Minor comment: if there is no parameter, then you'd better use `errors.New` instead.

tjungblu: fixed

pkg/expect/expect.go
tjungblu: fixed

ahrtr:
This PR is based on an old commit https://github.com/tjungblu/etcd/commit/0d4a516d11d933dc15388fe64b760820ab3d77fc, please rebase this PR, then I will take another look.

tjungblu:
@ahrtr just rebased, PTAL.

ahrtr (approved):
Overall looks good to me. 

Great work. Thank you @tjungblu

tjungblu:
Thanks @ahrtr. Please ping me on any flakes and perf regressions, I'll monitor this throughout the week.

Issue #14513, opened 2022-09-23
etcd/raft go.mod polluted with test only dependencies


dims:
see graph below on how things like `cockroachdb/datadriven` etc end up in kubernetes via etcd dependencies. Even though we have a go.mod file under `raft` directory, since there is a intermix of code and tests, the go.mod file ends up with things it should not. Is it possible to move test only code to a separate directory and drop a new go.mod file there?

![image](https://user-images.githubusercontent.com/23304/192025017-6e2ae544-a0b4-437b-87af-980122f17911.png)

serathius:
What kind of test dependencies do you care the most? I don't know if it's possible to just move all unit tests to separate package or remove dependencies from all unit tests. In the graph dependencies also go etcd/server package. Is it problematic for you too?

Overall note. This doesn't seem like etcd only issue, but a language issue. Test dependencies not being separated from the module. I don't know if issue can be fixes for every K8s dependency.

dims:
@serathius for this specific ask, if we can introduce a go.mod in https://github.com/etcd-io/etcd/tree/main/raft/rafttest that will cut out a whole lot of stuff.

NOTE: if someone wants to use raft, they don't really want cockroachdb to go with it. yes, i agree it is a language issue, i am asking for the integration test dependencies not present in the library's go.mod file here. 

thanks!

serathius:
@tbg thoughts?

tbg:
`datadriven` is a small testing dependency (like `require`, for example). It's not ideal that it is in the import graph, as @serathius points out this is because of the `rafttest` package. I think the solution would be more simply to move the contents of `rafttest` (or at least `InteractionEnv`) into `_test.go` files in the raft package itself, as it's only used from within the raft tests. The original idea was to make it available to implementers as well, to check their implementations for correctness, but we can revisit that if that work actually happens.

By the way, I also have a pet peeve about dependencies, there is a dependency on `etcd/api/v3/versionpb`[^1] which has caused problems using the library from CockroachDB[^2]; effectively we're vendoring with a custom patch[^3]. I suspect this causes problems for other folks as well, so hopefully we can fix that too. Can etcd folks find a way to do their versioning in a way that doesn't require these field annotations and associated import?

[^2]: https://github.com/cockroachdb/cockroach/pull/88985#issuecomment-1261868315
[...]

PR #14674, opened 2022-11-01
raft: add fuzzer

AdamKorcz:
Adds a fuzzer from cncf fuzzing: https://github.com/cncf/cncf-fuzzing/blob/main/projects/etcd/raft_fuzzer.go

Signed-off-by: AdamKorcz <adam@adalogics.com>


raft/fuzz_test.go
```diff
@@ -0,0 +1,215 @@
...
+	"runtime"
+	"strings"
+	"sync"
+	"testing"
+
+	fuzz "github.com/AdaLogics/go-fuzz-headers"
+	pb "go.etcd.io/etcd/raft/v3/raftpb"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
```
ahrtr, line 28: Minor comment: let's try to get third-party dependencies included in one group, and etcd dependencies in another group.
```suggestion
	fuzz "github.com/AdaLogics/go-fuzz-headers"
	"go.uber.org/zap"
	"go.uber.org/zap/zapcore"
	
	pb "go.etcd.io/etcd/raft/v3/raftpb"
```

raft/fuzz_test.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	core := zapcore.NewCore(zapcore.NewJSONEncoder(encoderCfg), os.Stdout, zap.FatalLevel)
+	return zap.New(core).WithOptions(options...)
+}
+
+func FuzzStep(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		initter.Do(initFunc)
+		defer catchPanics()
+		f := fuzz.NewConsumer(data)
```
ahrtr, line 130: I do not get time to dig into [NewConsumer](github.com/AdaLogics/go-fuzz-headers). Is it similar to the [seed corpus](https://go.dev/security/fuzz/) supported starting from golang 1.18? What's the benefit to use [go-fuzz-headers](github.com/AdaLogics/go-fuzz-headers)? 

Have you considered to use seed corpus directly?  Please see example https://github.com/etcd-io/etcd/blob/2e790d2c4f57eebdb266537ffe9eb7801ebc7cc3/server/etcdserver/api/v3rpc/validationfuzz_test.go#L27-L31

raft/fuzz_test.go
```diff
@@ -0,0 +1,215 @@
...
+	if strings.Contains(err, "GOT A FUZZ ERROR") {
+		return false
+	}
+
+	return true
+}
+
+func catchPanics() {
+	if r := recover(); r != nil {
+		var err string
```
ahrtr, line 98: Usually `err` is error type. 
```suggestion
		var errMsg string
```

ahrtr:
Please fix the static analysis issue, and try to integrate the fuzz test into [fuzzing.sh](https://github.com/etcd-io/etcd/blob/main/scripts/fuzzing.sh). FYI. https://github.com/etcd-io/etcd/pull/14561

raft/fuzz_test.go
AdamKorcz: `NewConsumer` is not related to the seed corpus feature of golang 1.18 fuzzing. `NewConsumer` is a helper utility to create different types of structured data formats which in this fuzzer is a struct. You can also easily use it to creates slice and/or tar archives.

raft/go.mod
```diff
@@ -12,17 +12,23 @@ require (
 	go.etcd.io/etcd/client/pkg/v3 v3.6.0-alpha.0
 )
 
+require (
```
serathius, line 15: I think we should not put raft fuzz tests in the same package. There have been already requests to stop polluting package with test dependencies https://github.com/etcd-io/etcd/issues/14513. This will do it even more. 

I would propose to create rafttest package in this PR to avoid adding more dependencies to raft.
Would be good to get opinion from @tbg and other people working on raft @nvanbenschoten @pavelkalinnikov

tbg: Maybe there are some guidelines I'm not aware of, but the "go way" is to write unit tests in the same package. Unit tests use helpers such as datadriven, require, fuzzing, etc and so these would be in the same package's tests. I would be very hesitant to bend over backwards and test in unusual ways just because we're worried about `go.mod` hygiene.

Testing dependencies will (should) be pruned by anyone importing the package[^1], i.e. won't actually be pulled.
For example, we use `datadriven` in tests (and it's even used in the main `rafttest` package), but because `raft/v3` doesn't reference `rafttest`, it's not a dependency:

```
go list -deps | grep datadriven
```

The same will be the case here, so I don't see an issue.

[^1]: https://go.dev/ref/mod#graph-pruning

pavelkalinnikov: You probably mean creating a separate module? I've seen a few projects using this technique: https://dev.to/wallyqs/introducing-gotest-mod-18a1. Although I think it might be unnecessary with Go [module graph pruning](https://go.dev/ref/mod#graph-pruning): https://stackoverflow.com/questions/64071364/best-way-to-use-test-dependencies-in-go-but-prevent-export-them.

pavelkalinnikov: Sorry, didn't see Tobi's comment. But we essentially agree.

liggitt: +1 on not adding a lot of transitive deps for fuzzing. with fuzz support being added directly to go, I would expect to see fuzzing-related dependencies decreasing, not increasing

ahrtr: Indeed the `datadriven` will not be included in the the module graph of a module which depends on `raft` because it's pruned as long as the modules specifies go 1.17 or higher. The reason of https://github.com/etcd-io/etcd/issues/14513 is the etcd 3.5.x specifies go 1.16.  So adding the dependency "github.com/AdaLogics/go-fuzz-header" will not pollute the module graph of Kubernetes any more, because both etcd `main` (3.6) and Kubernetes `main` already specify go 1.19.

So the focus for now is should etcd accept adding the new dependency into `raft`? Personally I think it might be OK, as long as the dependency is well maintained.

raft/fuzz_test.go
```diff
@@ -0,0 +1,211 @@
...
+		EncodeTime:     zapcore.ISO8601TimeEncoder,
+		EncodeDuration: zapcore.StringDurationEncoder,
+	}
+	core := zapcore.NewCore(zapcore.NewJSONEncoder(encoderCfg), os.Stdout, zap.FatalLevel)
+	return zap.New(core).WithOptions(options...)
+}
+
+func FuzzStep(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		initter.Do(initFunc)
```
tbg, line 128: This changes global state for anything else that runs after `FuzzStep`, can we instead set the logger and reset it again at the end?

```go
defer SetLogger(getLogger())
SetLogger(discardLogger)
```

tbg:
Haven't reviewed in detail, but at a high level, could you add more explanation of how this mechanism is invoked, if at all? Are we adding the *ability* to fuzz without actually invoking that ability anywhere (and if so, what are next step), or does CNCF fuzzing invoke this automagically? A link to that mechanism might be helpful for interested readers such as myself.
I suspect the answer is that we should integrate with OSS-fuzz[^1].

[^1]: https://github.com/cncf/cncf-fuzzing#2-integrate-continuous-fuzzing-with-oss-fuzz

tbg:
Great, thanks for the clarification, Adam.

raft/go.mod
liggitt: I still think the project should avoid taking on additional dependencies like this, but if they don't show up in `go mod graph`, I object less strenuously.

raft/fuzz_test.go
ahrtr: +1

I recall I added a comment to remove the global `initter`, but somehow the comment disappeared.

scripts/fuzzing.sh
```diff
@@ -3,16 +3,32 @@ set -e
...
+    run_fuzzer "${target_path}" "${target}"
 done
 
+# raft fuzzers
+target_path=${TARGET_PATH:-"./raft"}
+TARGETS="FuzzStep"
+
+for target in ${TARGETS}; do
+    run_fuzzer "${target_path}" "${target}"
+done
```
ahrtr, line 34: If users specify the `TARGET_PATH`, then this script will run into issue? 

I would suggest some script/code something like below so as to make it more extensible,
```
for target_path in ${TARGET_PATHS}; do 
    for target in ${TARGETS}; do
        run_fuzzer "${target_path}" "${target}"
    done
done
```

scripts/fuzzing.sh
```diff
@@ -3,16 +3,32 @@ set -e
 source ./scripts/test_lib.sh
 
 GO_CMD="go"
-fuzz_time=${FUZZ_TIME:-"300s"}
-target_path=${TARGET_PATH:-"./server/etcdserver/api/v3rpc"}
-TARGETS="FuzzTxnRangeRequest  FuzzTxnPutRequest  FuzzTxnDeleteRangeRequest"
+fuzz_time=${FUZZ_TIME:-"3s"}
```
ahrtr, line 6: Why do you change the default value (300s)? I am not sure what's the best practice, but "3s" seems too small.

AdamKorcz: That should not have been changed. Thanks for catching that!

scripts/fuzzing.sh
AdamKorcz: I wonder if that is an issue. This comes from the current version of `scripts/fuzzing.sh`:
https://github.com/etcd-io/etcd/blob/2f558ca0dbf1217c1cc5b82a1e2aec428bf6a04f/scripts/fuzzing.sh#L1-L8

Is `TARGET_PATH` meant to be specified by users?

ahrtr: When we want to manually run part of the fuzzing test, then we might set the `TARGET_PATH`.

AdamKorcz: Given that each target path will have differently named fuzz tests, I think it would look something like this. Is that what you had in mind? 
```bash
#!/usr/bin/env bash
set -e
source ./scripts/test_lib.sh

GO_CMD="go"
fuzz_time=${FUZZ_TIME:-"3s"}

function run_fuzzer() {
    target_path=$1
    target=$2

    log_callout -e "\\nExecuting fuzzing with target ${target} in $target_path with a timeout of $fuzz_time\\n"
    run pushd "${target_path}"
        $GO_CMD test -fuzz "${target}" -fuzztime "${fuzz_time}"
    run popd
    log_success -e "\\COMPLETED: fuzzing with target $target in $target_path \\n"
}

if [ ! -z "${TARGET_PATHS}" ] # If TARGET_PATHS is set
then
        for target_path in "${TARGET_PATHS}"; do
                if [ $target_path == "./server/etcdserver/api/v3rpc" ]; then
                        TARGETS="FuzzTxnRangeRequest  FuzzTxnPutRequest  FuzzTxnDeleteRangeRequest"
                        for target in ${TARGETS}; do
                                run_fuzzer "${target_path}" "${target}"
                        done
                elif [ $target_path == "./raft" ]; then
                        TARGETS="FuzzStep"
                        for target in ${TARGETS}; do
                                run_fuzzer "${target_path}" "${target}"
                        done
                fi
        done
else
        echo "TARGET_PATHS not set. Running all fuzzers"
        # api fuzzers
        target_path=${TARGET_PATH:-"./server/etcdserver/api/v3rpc"}
        TARGETS="FuzzTxnRangeRequest  FuzzTxnPutRequest  FuzzTxnDeleteRangeRequest"

        for target in ${TARGETS}; do
                run_fuzzer "${target_path}" "${target}"
        done

        # raft fuzzers
        target_path=${TARGET_PATH:-"./raft"}
        TARGETS="FuzzStep"

        for target in ${TARGETS}; do
                run_fuzzer "${target_path}" "${target}"
        done
fi
```

ahrtr: Yes, when $TARGET_PATHS isn't set. Can we use a two level loops just I mentioned in https://github.com/etcd-io/etcd/pull/14674#discussion_r1015821330? Otherwise, when any new target is added, then the new contributor needs to copy-paste the duplicated code.

ahrtr:
A couple of comments:
1. One PR is supposed to do one thing. But I see you bumped some dependencies as well. Could you do that in separate PRs?
2. Please signoff the commit: git rebase HEAD~1 --signoff
3. You introduced new dependency `github.com/AdaLogics/go-fuzz-headers` in this PR. I am neutral about this, but I see some concerns on this. I am not sure whether you can remove the dependency, and use golang fuzz corpus?

PR #14682, opened 2022-11-03
Reproduce issue 13766 in linearizability tests

serathius:
Decided to use the original approach of reproduction of the issue. I was not able to reliable reproduction via SIGKILL on lower QPS or clients. This approach has downside that validating linerizability for so many clients is too costly.  Memory usage for pourcupine just exploded and would definitely not fit in github action worker. 

Decided to use the initial data inconsistency validation for reproduction.

### DIFF
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -36,12 +36,23 @@ import (
 )
 
 const (
-	// minimalQPS is used to validate if enough traffic is send to make tests accurate.
-	minimalQPS = 100.0
-	// maximalQPS limits number of requests send to etcd to avoid linearizability analysis taking too long.
-	maximalQPS = 200.0
 	// waitBetweenFailpointTriggers
 	waitBetweenFailpointTriggers = time.Second
+)
+
+var (
+	DefaultTrafficConfig = trafficConfig{
+		minimalQPS:  100,
+		maximalQPS:  200,
+		clientCount: 8,
+		traffic:     DefaultTraffic,
+	}
+	HighTrafficConfig = trafficConfig{
+		minimalQPS:  200,
+		maximalQPS:  1000,
+		clientCount: 12,
+		traffic:     DefaultTraffic,
+	}
 )
 
 func TestLinearizability(t *testing.T) {
@@ -50,12 +61,15 @@ func TestLinearizability(t *testing.T) {
 		name      string
 		failpoint Failpoint
 		config    e2e.EtcdProcessClusterConfig
+		traffic   *trafficConfig
 	}{
 		{
 			name:      "ClusterOfSize1",
 			failpoint: RandomFailpoint,
+			traffic:   &HighTrafficConfig,
 			config: *e2e.NewConfig(
 				e2e.WithClusterSize(1),
+				e2e.WithSnapshotCount(100),
 				e2e.WithPeerProxy(true),
 				e2e.WithGoFailEnabled(true),
 				e2e.WithCompactionBatchLimit(100), // required for compactBeforeCommitBatch and compactAfterCommitBatch failpoints
@@ -64,7 +78,9 @@ func TestLinearizability(t *testing.T) {
 		{
 			name:      "ClusterOfSize3",
 			failpoint: RandomFailpoint,
+			traffic:   &HighTrafficConfig,
 			config: *e2e.NewConfig(
+				e2e.WithSnapshotCount(100),
 				e2e.WithPeerProxy(true),
 				e2e.WithGoFailEnabled(true),
 				e2e.WithCompactionBatchLimit(100), // required for compactBeforeCommitBatch and compactAfterCommitBatch failpoints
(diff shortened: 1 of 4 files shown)


serathius:
cc @ahrtr @spzala @chaochn47 @ptabor

tests/linearizability/linearizability_test.go
```diff
@@ -67,6 +76,27 @@ func TestLinearizability(t *testing.T) {
...
+			},
+		},
+		{
+			name:      "Issue13766",
+			failpoint: KillFailpoint,
+			config: e2e.EtcdProcessClusterConfig{
+				ClusterSize:         3,
+				InitialCorruptCheck: true,
+			},
+			skipValidation: true,
```
ahrtr, line 93: If you skip the validation, then it isn't linearizablity test anymore. So it doesn't make sense to get the test included in the linearzability test suite. 

And also shouldn't we check HashKV after the test? The existing functional test already covers the case. Linearizablity test is good, but please do not recreate the wheel. Instead, I would suggest to enhance both the linearizablity and functional tests.

chaochn47:
> I was not able to reliable reproduction via SIGKILL on lower QPS or clients. This approach has downside that validating linerizability for so many clients is too costly. Memory usage for pourcupine just exploded and would definitely not fit in github action worker.

What's the machine requirement to reliably run this test to reproduce the issue #14682 in case someone wants to locally try it out.

tests/linearizability/linearizability_test.go
serathius: > Linearizablity test is good, but please do not recreate the wheel. Instead, I would suggest to enhance both the linearizablity and functional tests.

As long functional tests are flaky they are useless. Using  initial hash check in linerazibility tests I see as a shortcut that I hope to remove in future, however I don't see same possibility in functional tests. 

We are already not benefiting from running functional tests, so when we reproduce all standing inconstancy issues and add network blackholing in lineariziability tests I will propose to remove them.

serathius:
> What's the machine requirement to reliably run this test to reproduce the issue https://github.com/etcd-io/etcd/pull/14682? Just in case someone wants to locally try it out.

Pretty low as github worker can reach 2k qps that is required for reproduction. Hard to give estimates.

serathius:
Working on making the test not dependent on data inconsistency check.

serathius:
> What's the machine requirement to reliably run this test to reproduce the issue https://github.com/etcd-io/etcd/pull/14682? Just in case someone wants to locally try it out.

Problem is that linearization of history is an NP-Hard problem. When we increase number of clients and qps the compute/space required explodes. I was not able to get results from linearization as I usually stop it after 5 minutes when it allocates over 100GB of RAM.

serathius:
Marking as draft as I'm still working on making tests execute within reasonable time.

serathius:
Got first success! cc @ahrtr @chaochn47 

I managed to get over issue of too many clients by looking up persisted requests via watch. Fixing this requires a lot of changes so I will do it one by one, starting from https://github.com/etcd-io/etcd/pull/15044. 

Reproducing #13766 requires pretty high QPS load (>1000 qps). To achieve that we need increase number of clients from 8 to 20. This caused issue for linearizability verification as it's complexity is exponential to number of clients. This is because when we crash etcd, ongoing request from each client is interrupted, meaning for 8 clients we lost 8 requests, for 20 clients we loos 20 requests. 

Client doesn't know if a lost request is persisted on not. If requests was lost on invocation (before it gets to etcd), it will not be persisted. However if request was lost on return (after etcd processed it), it will be persisted without notifying the client. Consecutive lost requests cause complexity explosion, because current etcd model will try to find whether they were persisted and their exact order of their execution, so for each n lost requests there are `n!` orderings of requests and `2^n` possible final states for each path. Pretty big!

Solution was to minimize number of lost requests by checking if they were really persisted. There is a easy source of this information in etcd, watch. By collecting all the events in separate running watch I can verify if particular request was persisted (all Put requests have unique id) and also get revision it was persisted. This eliminates both problems of request ordering and checking persistence. 

I have a draft code, just need to split it into PRs for review. First two https://github.com/etcd-io/etcd/pull/15045 https://github.com/etcd-io/etcd/pull/15044

serathius:
Sending code for review! Possibly will also need https://github.com/etcd-io/etcd/pull/15101 due to increase qps this issue happens more frequently. 
cc @ahrtr @ptabor

Makefile
```diff
@@ -129,6 +129,60 @@ gofail-disable: install-gofail
...
+	  git clone --depth 1 --branch main https://github.com/etcd-io/etcd.git .; \
+	  make gofail-enable; \
+	  make build;
+	mkdir -p ./bin
+	cp /tmp/etcd-release-v3.6.0/bin/etcd ./bin/etcd-v3.6.0-failpoints
+
+./bin/etcd-v3.5.2-failpoints:
+./bin/etcd-v3.5.4-failpoints:
+./bin/etcd-v3.5.6-failpoints:
+./bin/etcd-v3.5.%-failpoints:
```
ptabor, line 162: Do we need the previous L159-161 lines in context of '162' that contains '%'

Makefile
```diff
@@ -129,6 +129,60 @@ gofail-disable: install-gofail
 install-gofail:
 	cd tools/mod; go install go.etcd.io/gofail@${GOFAIL_VERSION}
 
+# Reproduce historical issues
```
ptabor, line 132: How about moving this to a separate Makefile within './tests/repros' directory. 

I think it's too detailed for a top-level Makefile (and have big potential to growth over time).

serathius: Ups, didn't want this included with PR.

ptabor (approved):

PR #14683, opened 2022-11-04
test: refactor the framework structure to make it clearer

ahrtr:


ahrtr:
Previously we added both [e2e.go](https://github.com/etcd-io/etcd/blob/main/tests/framework/e2e.go) and and [integration.go ](https://github.com/etcd-io/etcd/blob/main/tests/framework/integration.go) in the same package [framework](https://github.com/etcd-io/etcd/tree/main/tests/framework).  It can work, but it isn't elegant, because anything specific to `e2e` should be invisible to `integration`, and anything specific to `integration` should be invisible to `e2e`. 

If there are something which are common to both `e2e` and `integration`, then let's get them included in [framework/config](https://github.com/etcd-io/etcd/tree/main/tests/framework/config), [framework/testutils](https://github.com/etcd-io/etcd/tree/main/tests/framework/testutils), [tests/common](https://github.com/etcd-io/etcd/tree/main/tests/common) or [framework/intf](https://github.com/ahrtr/etcd/tree/test_framework_20221104/tests/framework/intf).

Please let me know if anyone has any comments or concerns.

cc @mitake @ptabor @serathius @spzala @chaochn47 @biosvs

chaochn47 (approved):

tests/framework/intf/interface.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package framework
+package intf
```
serathius, line 15: Don't understand where `intf` name comes from. When importing package, it should be clear what it contains by it's name.

ahrtr: `intf` stands for `interface`. I vaguely recall that it's common used when I was coding with C++ more than 10 years ago. 

Also FYI. https://www.acronymfinder.com/Interface-(INTF).html

serathius: Never seen it in Go though. Please let's not isolate new contributors that haven't used C++ to just save 5 characters in name. 
Balance in naming is definitely important, for example it doesn't make sense to use full name for local variable, however package name should be clear and informative over short and cryptic.

ahrtr: Note that `interface` is a key word in golang, I intentionally used "intf" so as not to conflict with the keyword.

ahrtr: If you don't like "intf", do you have any better proposal?

ahrtr: Changed to "`interfaces`". 

Please let me know if you still have any comment.

serathius (approved):

PR #14685, opened 2022-11-04
Revision inconsistency caused by panic during defrag

serathius:

### DIFF
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -101,12 +101,11 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
 		if state.Value == response.getData {
 			return true, state
 		}
-		for write := range state.FailedWrites {
-			if write == response.getData {
-				state.Value = response.getData
-				delete(state.FailedWrites, write)
-				return true, state
-			}
+		_, ok := state.FailedWrites[response.getData]
+		if ok {
+			state.Value = response.getData
+			delete(state.FailedWrites, response.getData)
+			return true, state
 		}
 	case Put:
 		if response.err == nil {
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -33,13 +33,15 @@ type etcdRequest struct {
 }
 
 type etcdResponse struct {
-	getData string
-	err     error
+	getData  string
+	revision int64
+	err      error
 }
 
 type EtcdState struct {
 	Key          string
 	Value        string
+	LastRevision int64
 	FailedWrites map[string]struct{}
 }
 
@@ -51,9 +53,6 @@ var etcdModel = porcupine.Model{
 		if err != nil {
 			panic(err)
 		}
-		if state.FailedWrites == nil {
-			state.FailedWrites = map[string]struct{}{}
-		}
 		ok, state := step(state, in.(etcdRequest), out.(etcdResponse))
 		data, err := json.Marshal(state)
 		if err != nil {
@@ -64,22 +63,19 @@ var etcdModel = porcupine.Model{
 	DescribeOperation: func(in, out interface{}) string {
 		request := in.(etcdRequest)
 		response := out.(etcdResponse)
-		var resp string
 		switch request.op {
 		case Get:
 			if response.err != nil {
-				resp = response.err.Error()
+				return fmt.Sprintf("get(%q) -> %q", request.key, response.err)
 			} else {
-				resp = response.getData
+				return fmt.Sprintf("get(%q) -> %q, rev: %d", request.key, response.getData, response.revision)
 			}
-			return fmt.Sprintf("get(%q) -> %q", request.key, resp)
 		case Put:
 			if response.err != nil {
-				resp = response.err.Error()
+				return fmt.Sprintf("put(%q, %q) -> %s", request.key, request.putData, response.err)
 			} else {
-				resp = "ok"
+				return fmt.Sprintf("put(%q, %q) -> ok, rev: %d", request.key, request.putData, response.revision)
 			}
-			return fmt.Sprintf("put(%q, %q) -> %s", request.key, request.putData, resp)
 		default:
 			return "<invalid>"
 		}
@@ -88,32 +84,68 @@ var etcdModel = porcupine.Model{
 
 func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
 	if request.key == "" {
-		panic("Invalid request")
+		panic("invalid request")
 	}
 	if state.Key == "" {
-		state.Key = request.key
+		return true, initState(request, response)
 	}
 	if state.Key != request.key {
 		panic("Multiple keys not supported")
 	}
 	switch request.op {
 	case Get:
-		if state.Value == response.getData {
-			return true, state
-		}
-		_, ok := state.FailedWrites[response.getData]
-		if ok {
-			state.Value = response.getData
-			delete(state.FailedWrites, response.getData)
-			return true, state
-		}
+		return stepGet(state, request, response)
+	case Put:
+		return stepPut(state, request, response)
+	default:
+		panic("Unknown operation")
+	}
+}
+
+func initState(request etcdRequest, response etcdResponse) EtcdState {
+	state := EtcdState{
+		Key:          request.key,
+		LastRevision: response.revision,
+		FailedWrites: map[string]struct{}{},
+	}
+	switch request.op {
+	case Get:
+		state.Value = response.getData
 	case Put:
 		if response.err == nil {
 			state.Value = request.putData
 		} else {
 			state.FailedWrites[request.putData] = struct{}{}
 		}
+	default:
+		panic("Unknown operation")
+	}
+	return state
+}
+
+func stepGet(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
+	if state.Value == response.getData && state.LastRevision <= response.revision {
+		return true, state
+	}
+	_, ok := state.FailedWrites[response.getData]
+	if ok && state.LastRevision < response.revision {
+		state.Value = response.getData
+		state.LastRevision = response.revision
+		delete(state.FailedWrites, response.getData)
 		return true, state
 	}
 	return false, state
+}
+
+func stepPut(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
+	if response.err != nil {
+		state.FailedWrites[request.putData] = struct{}{}
+		return true, state
+	}
+	if state.LastRevision >= response.revision {
+		return false, state
+	}
+	state.Value = request.putData
+	state.LastRevision = response.revision
+	return true, state
 }
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -66,7 +66,7 @@ func (c *recordingClient) Get(ctx context.Context, key string) error {
 		ClientId: c.id,
 		Input:    etcdRequest{op: Get, key: key},
 		Call:     callTime.UnixNano(),
-		Output:   etcdResponse{getData: readData},
+		Output:   etcdResponse{getData: readData, revision: resp.Header.Revision},
 		Return:   returnTime.UnixNano(),
 	})
 	return nil
@@ -74,13 +74,17 @@ func (c *recordingClient) Get(ctx context.Context, key string) error {
 
 func (c *recordingClient) Put(ctx context.Context, key, value string) error {
 	callTime := time.Now()
-	_, err := c.client.Put(ctx, key, value)
+	resp, err := c.client.Put(ctx, key, value)
 	returnTime := time.Now()
+	var revision int64
+	if resp != nil && resp.Header != nil {
+		revision = resp.Header.Revision
+	}
 	c.operations = append(c.operations, porcupine.Operation{
 		ClientId: c.id,
 		Input:    etcdRequest{op: Put, key: key, putData: value},
 		Call:     callTime.UnixNano(),
-		Output:   etcdResponse{err: err},
+		Output:   etcdResponse{err: err, revision: revision},
 		Return:   returnTime.UnixNano(),
 	})
 	return nil
--- a/tests/linearizability/model_test.go
+++ b/tests/linearizability/model_test.go
@@ -16,68 +16,112 @@ package linearizability
 
 import (
 	"errors"
-	"github.com/anishathalye/porcupine"
 	"testing"
 )
 
 func TestModel(t *testing.T) {
 	tcs := []struct {
-		name          string
-		okOperations  []porcupine.Operation
-		failOperation *porcupine.Operation
+		name       string
+		operations []testOperation
 	}{
 		{
-			name: "Etcd must return what was written",
-			okOperations: []porcupine.Operation{
-				{Input: etcdRequest{op: Put, key: "key", putData: "1"}, Output: etcdResponse{}},
-				{Input: etcdRequest{op: Get, key: "key"}, Output: etcdResponse{getData: "1"}},
-			},
-			failOperation: &porcupine.Operation{Input: etcdRequest{op: Get, key: "key"}, Output: etcdResponse{getData: "2"}},
-		},
-		{
-			name: "Etcd can crash after storing result but before returning success to client",
-			okOperations: []porcupine.Operation{
-				{Input: etcdRequest{op: Put, key: "key", putData: "1"}, Output: etcdResponse{err: errors.New("failed")}},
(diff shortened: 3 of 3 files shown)


serathius:
cc @ahrtr @spzala @ptabor @chaochn47

tests/linearizability/model.go
```diff
@@ -96,21 +95,25 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
+		if ok {
+			state.Value = response.getData
+			state.LastRevision = response.revision
+			delete(state.FailedWrites, response.getData)
+			return true, state
 		}
 	case Put:
 		if response.err == nil {
 			state.Value = request.putData
+			state.LastRevision = response.revision
```
ahrtr, line 116: We might need to update `state.LastRevision` as long as `response.revision != 0`

tests/linearizability/model.go
```diff
@@ -96,21 +95,25 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
 	if state.Key != request.key {
 		panic("Multiple keys not supported")
 	}
+	if response.revision != 0 && response.revision < state.LastRevision {
+		return false, state
+	}
```
chaochn47, line 100: Does it deserve a unit test to explain and validate the expected behavior?

tests/linearizability/model.go
serathius: Fixed by requiring revision on Put to always increase.

tests/linearizability/model.go
serathius: I have updated the unit tests.

serathius:
ping @ahrtr

tests/linearizability/model.go
```diff
@@ -99,22 +98,32 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
-		} else {
+		if response.err != nil {
 			state.FailedWrites[request.putData] = struct{}{}
+			return true, state
+		}
+		if state.LastRevision < response.revision {
+			state.Value = request.putData
+			state.LastRevision = response.revision
+			return true, state
 		}
```
ahrtr, line 126: My previous comment https://github.com/etcd-io/etcd/pull/14685/files#r1014471776 might still valid.

Since `response.err` is `nill`, so we should save the `request.putData` and `response.revision` anyway. 

```suggestion
		lastRev := state.LastRevision
		state.Value = request.putData
		state.LastRevision = response.revision
		if lastRev < response.revision {
			return true, state
		}
```

tests/linearizability/model.go
```diff
@@ -99,22 +98,32 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
+		if state.LastRevision < response.revision {
+			_, ok := state.FailedWrites[response.getData]
+			if ok {
 				state.Value = response.getData
-				delete(state.FailedWrites, write)
+				state.LastRevision = response.revision
+				delete(state.FailedWrites, response.getData)
 				return true, state
 			}
 		}
```
ahrtr, line 116: It's a little hard to understand. I would suggest to rewrite the case Get branch something like below,

```
		if state.LastRevision == 0 {
			// run into this branch if there is no any previous successful
			// put operations.
			state.Value = response.getData
			state.LastRevision = response.revision
			return true, state
		}
		if state.Value == response.getData && state.LastRevision == response.revision {
			return true, state
		} else {
			_, ok := state.FailedWrites[response.getData]
			if ok {
				state.Value = response.getData
				state.LastRevision = response.revision
				delete(state.FailedWrites, response.getData)
				return true, state
			}
		}
```

tests/linearizability/model.go
serathius: Commit suggestion you proposed doesn't change the logic at all, you just save data before you validate the state transition is correct. 

> We might need to update state.LastRevision as long as response.revision != 0

Please remember that revision is a positive number, so checking that revision is increasing like in `state.LastRevision < response.revision` already ensures that `response.revision != 0`

ahrtr: As long as `response.err` is nil, then we need to save `request.putData` and `response.revision` to state. While your code only save them when `state.LastRevision < response.revision` and of course `response.err `is nil,.

serathius:
The current model found 2 linearizability issues found in CI run https://github.com/etcd-io/etcd/actions/runs/3418916665. Possibly model is too strict but would like to discuss it openly to make sure I didn't miss anything. I'm working on reproducing them locally.

First case in single cluster test we have a case of etcd skipping 9 revisions even though there was no other writes. This breaks assumption that revision increases only on successfull write. image
Would really love to see the db file, so I think the follow up to that issue is to implement uploading etcd data directory on failed test.

Second case in 3 node cluster is even stranger, etcd reported different data for the same revision. There was a long running get transaction that returned old data with much newer revision.
image

Conclusion, there seems to be an issue with revision header returned in Get header. cc @ahrtr

serathius:
Now running test with relaxed model that doesn't require revision to be continuous.

tests/linearizability/model.go
serathius: > As long as response.err is nil, then we need to save request.putData and response.revision to state

That's not true. Model should only allow operations that would be correct for etcd. Etcd revision should never decrease. That's why I put `state.LastRevision < response.revision` check.

tests/linearizability/model.go
```diff
@@ -64,57 +66,85 @@ var etcdModel = porcupine.Model{
...
+	}
+}
+
+func initState(request etcdRequest, response etcdResponse) EtcdState {
+	var value string
+	switch request.op {
+	case Get:
+		value = response.getData
+	case Put:
+		value = request.putData
```
ahrtr, line 111: Do you need to check the `response.err`?

tests/linearizability/model.go
```diff
@@ -64,57 +66,85 @@ var etcdModel = porcupine.Model{
...
+	}
+}
+
+func stepGet(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
+	if state.Value == response.getData && state.LastRevision <= response.revision {
+		return true, state
+	}
+	if state.LastRevision >= response.revision {
+		return false, state
+	}
```
ahrtr, line 128: It seems not correct, because both `if` statements cover the condition `state.LastRevision == response.revision`.

ahrtr:
> Conclusion, there seems to be an issue with revision header returned in Get header. cc @ahrtr

Please note that the revision in the `RangeResponse.ResponseHeader` is coming from [API](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/server/storage/mvcc/kvstore_txn.go#L63) instead of raft. We should ignore revision in `RangeResponse`; instead we should only check revision in `PutResponse`.

tests/linearizability/model.go
serathius: that's a good point. First request can be an error. There is a low probability of this happening in linearization tests as we wait 1 second before disrupting cluster, but model should be correct by itself.

tests/linearizability/model.go
serathius: This should be correct, let me explain.

First if statement validates that Get read result of last successful put. Value that is stored in state needs to equal the result in put. On the other hand second and third statement handle a put that returned an error to client but was persisted. To "restore" a lost put, there are two requirements:
* Revision must have increased (second if statement)
* There must be a request that failed that stored the data (third if statement).

Let me rewrite the code for this to be more clear.

serathius:
> Please note that the revision in the RangeResponse.ResponseHeader is coming from [API](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/server/storage/mvcc/kvstore_txn.go#L63) instead of raft.

You are right, but in both cases revision should come from bbolt database. It is expected to be the revision for the latest write. For put requests it means that it is revision of the write itself. For get request it's revision of the request, but revision of latest write that happened before. I still think it should be monotonically increasing and this is what the model tests.

ahrtr:
I think learnability test only guarantees that read request never reads stale data (of course, the revision in RangeResponse Header will >= the revision in previous write request). What you have done so far (not including this PR) is good. 

But the revision in RangeResponse header is volatile (although it's always increasing), even there is no any write request on the key you are reading. As long as there are write requests on other keys, revision in RangeResponse header will change. . It means that when two read requests read the same key, one request might get new data but old revision. Please note that in a range request, reading the revision and creating a "snapshot" of the data isn't an atomic operation. See [kvstore_txn.go#L46-L54](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/server/storage/mvcc/kvstore_txn.go#L46-L54)

Probably you can use [CreateRevision](https://github.com/etcd-io/etcd/blob/49ecea5dae105c63cc19642af9e13c4be1f9f992/api/mvccpb/kv.pb.go#L56) instead.  But per my immediate feeling, it might not be a good idea to test revision in learnability test. Instead, we should verify the revision at the end of the test, as long as all members can converge the same revision, then it's OK. Note: I was planning to add such checker for functional test.

serathius:
Implemented https://github.com/etcd-io/etcd/pull/14714 to dump data directory after the tests. I got interesting results, as mentioned before revisions in each instance does not match. Same PUT request `put("key", "1093") in different etcd instances has different revision. 8662 in first member, 8669 in second member, 8656 in third member.

I got the result by downloading artifact from https://github.com/etcd-io/etcd/actions/runs/3426561978 and analysing files using `etcd-dump-db iterate-bucket ./TestLinearizability_ClusterOfSize3/TestLinearizabilityClusterOfSize3-test-0/member/snap/db key --decode | head`. Results

Member 0:
```
rev={main:8662 sub:0}, value=[key "key" | val "1093" | created 2 | mod 8662 | ver 8661]
rev={main:8661 sub:0}, value=[key "key" | val "6001081" | created 2 | mod 8661 | ver 8660]
rev={main:8660 sub:0}, value=[key "key" | val "7001068" | created 2 | mod 8660 | ver 8659]
rev={main:8659 sub:0}, value=[key "key" | val "5001144" | created 2 | mod 8659 | ver 8658]
rev={main:8658 sub:0}, value=[key "key" | val "3001079" | created 2 | mod 8658 | ver 8657]
rev={main:8657 sub:0}, value=[key "key" | val "2001142" | created 2 | mod 8657 | ver 8656]
rev={main:8656 sub:0}, value=[key "key" | val "1001060" | created 2 | mod 8656 | ver 8655]
rev={main:8655 sub:0}, value=[key "key" | val "4001071" | created 2 | mod 8655 | ver 8654]
rev={main:8654 sub:0}, value=[key "key" | val "1092" | created 2 | mod 8654 | ver 8653]
rev={main:8653 sub:0}, value=[key "key" | val "6001080" | created 2 | mod 8653 | ver 8652]
```
Member 1 
```
rev={main:8669 sub:0}, value=[key "key" | val "1093" | created 2 | mod 8669 | ver 8668]
rev={main:8668 sub:0}, value=[key "key" | val "6001081" | created 2 | mod 8668 | ver 8667]
rev={main:8667 sub:0}, value=[key "key" | val "7001068" | created 2 | mod 8667 | ver 8666]
rev={main:8666 sub:0}, value=[key "key" | val "5001144" | created 2 | mod 8666 | ver 8665]
rev={main:8665 sub:0}, value=[key "key" | val "3001079" | created 2 | mod 8665 | ver 8664]
rev={main:8664 sub:0}, value=[key "key" | val "2001142" | created 2 | mod 8664 | ver 8663]
rev={main:8663 sub:0}, value=[key "key" | val "1001060" | created 2 | mod 8663 | ver 8662]
rev={main:8662 sub:0}, value=[key "key" | val "4001071" | created 2 | mod 8662 | ver 8661]
rev={main:8661 sub:0}, value=[key "key" | val "1092" | created 2 | mod 8661 | ver 8660]
rev={main:8660 sub:0}, value=[key "key" | val "6001080" | created 2 | mod 8660 | ver 8659]
```
Member 2:
```
rev={main:8656 sub:0}, value=[key "key" | val "1093" | created 2 | mod 8656 | ver 8655]
rev={main:8655 sub:0}, value=[key "key" | val "6001081" | created 2 | mod 8655 | ver 8654]
rev={main:8654 sub:0}, value=[key "key" | val "7001068" | created 2 | mod 8654 | ver 8653]
rev={main:8653 sub:0}, value=[key "key" | val "5001144" | created 2 | mod 8653 | ver 8652]
rev={main:8652 sub:0}, value=[key "key" | val "3001079" | created 2 | mod 8652 | ver 8651]
rev={main:8651 sub:0}, value=[key "key" | val "2001142" | created 2 | mod 8651 | ver 8650]
rev={main:8650 sub:0}, value=[key "key" | val "1001060" | created 2 | mod 8650 | ver 8649]
rev={main:8649 sub:0}, value=[key "key" | val "4001071" | created 2 | mod 8649 | ver 8648]
rev={main:8648 sub:0}, value=[key "key" | val "1092" | created 2 | mod 8648 | ver 8647]
rev={main:8647 sub:0}, value=[key "key" | val "6001080" | created 2 | mod 8647 | ver 8646]
```

ahrtr:
I reproduced this issue locally, it seems like a real issue. Let me have a deep dive on this. Note: please ignore my previous comment, part (the "two read requests" part) of which isn't correct.

serathius:
@ahrtr Thanks for looking into this, did you do anything specific to reproduce issue locally?

ahrtr:
> @ahrtr Thanks for looking into this, did you do anything specific to reproduce issue locally? With local reproduction can you validate if v3.4 and v3.5 is affected. If so we should delay v3.5.6 release.

FYI. https://github.com/etcd-io/etcd/pull/14724

serathius:
Ok, I got a local repro. I run different etcd version that is not affected.

serathius:
Confirmed that issue is reproducable on main branch and on v3.5.5 release. Looks like v3.4.22 is not affected.

ahrtr:
> Looks like the cause is panic during etcd defrag. Issue was found thanks to `defragBeforeRename` and `defragBeforeCopy`

Confirmed. I have the same finding.

serathius:
Git history bisection has pointed out https://github.com/etcd-io/etcd/commit/2dbecea5b26282b5ded675b5ba5922976340784f as the cause.

ahrtr:
The reason is that etcd reapplied duplicated entries,
duplicated_items

Right now I can exclude raft from the criminal list. Most likely it's still caused by consistent index not updated in time
wal_entries

ahrtr:
Eventually I figured out the root cause, Please refer to https://github.com/etcd-io/etcd/pull/14730

ahrtr:
FYI. https://github.com/ahrtr/etcd-issues/tree/master/issues/revision_inconsistency#how-to-workaround-this-issue

serathius:
With fix merged with can also merge this change. ping @ahrtr

tests/linearizability/model.go
```diff
@@ -64,57 +66,85 @@ var etcdModel = porcupine.Model{
...
+		Value:        value,
+		LastRevision: response.revision,
+	}
+}
+
+func stepGet(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
+	if state.Value == response.getData && state.LastRevision <= response.revision {
+		return true, state
+	}
+	if state.LastRevision >= response.revision {
```
ahrtr, line 126: Two successive read requests may have the same revision. 
```suggestion
	if state.LastRevision > response.revision {
```

tests/linearizability/model.go
ahrtr: It seems this comment isn't resolved.

tests/linearizability/model.go
serathius: I explained it above, this if statements doesn't cover successive reads. It covers cases where a PUT request has failed before the read.

ahrtr (approved):
Overall looks good to me.

I'll leave you to decide whether/how to resolve https://github.com/etcd-io/etcd/pull/14685#discussion_r1017300018

tests/linearizability/model.go
serathius: Done

serathius:
> I'll leave you to decide whether/how to resolve https://github.com/etcd-io/etcd/pull/14685#discussion_r1017300018

Resolved the comment and added tests for it.

spzala (commented):
Thanks @serathius lgtm

ahrtr:
All the existing etcd 3.5 releases are affected, including 3.5.0 ~ 3.5.5.

**Note**: etcd 3.4 doesn't have this issue. Note that there is no impact on etcd 3.5 either if performing the defragmentation operation offline using etcdutl.

FYI. https://github.com/ahrtr/etcd-issues/tree/master/issues/revision_inconsistency#how-to-workaround-this-issue 

cc @dims @liggitt @deads2k @lavalamp

PR #14695, opened 2022-11-06
changelog: add item to cover the fix on adding protection on maintenance requests when auth is enabled

ahrtr:
Related PR: https://github.com/etcd-io/etcd/pull/14663

cc @mitake


mitake:
LGTM, thanks @ahrtr

PR #14697, opened 2022-11-07
test: support mix versions testing 

ahrtr:

### DIFF
--- a/tests/common/e2e_test.go
+++ b/tests/common/e2e_test.go
@@ -18,6 +18,7 @@
 package common
 
 import (
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
 	"go.etcd.io/etcd/tests/v3/framework"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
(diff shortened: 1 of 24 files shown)


ahrtr:
This PR is on top of https://github.com/etcd-io/etcd/pull/14610. 

The first commit is coming from the base PR directly, so please pay attention to the other commits in this PR. It should be much more elegant than the original PR. 

serathius:
> The first commit is coming from the base PR directly, so please pay attention to the other commits in this PR. It should be much more elegant than the original PR.

Definitely would not call using `any` as elegant as is can be very deceptive when checking for `nil`. I would have been happy to implement the change myself, if that was directly pointed out during code review. Please be more explicit in comments. 

Happy to see this unblocked.

serathius (approved):

ahrtr:
> Please be more explicit in comments.

I would have already done that if knew the answer:). Actually I don't have a clear thought before submitting this PR. What I was only sure about is that it's definitely an anti-pattern for integration test to check something it shouldn't care about at all.

ahrtr:
Usually when I am confused about something, then I will try to do some exercise/experiment myself, eventually the thought becomes clearer. Eventually the PR is almost ready when I get everything sorted out and my thought gets clearer, so I just delivered the PR directly.

PR #14707, opened 2022-11-08
test: support mix versions sending snapshot test

ahrtr:
cc  @mitake @ptabor @serathius @spzala

### DIFF
--- a/tests/e2e/etcd_mix_versions_test.go
+++ b/tests/e2e/etcd_mix_versions_test.go
@@ -1,0 +1,127 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+// TODO(ahrtr): add network partition scenario to trigger snapshots.
+func TestMixVersionsSendSnapshot(t *testing.T) {
+	cases := []struct {
+		name              string
+		clusterVersion    e2e.ClusterVersion
+		newInstaceVersion e2e.ClusterVersion
+	}{
+		// etcd doesn't support adding a new member of old version into
+		// a cluster with higher version. For example, etcd cluster
+		// version is 3.6.x, then a new member of 3.5.x can't join the
+		// cluster. Please refer to link below,
+		// https://github.com/etcd-io/etcd/blob/3e903d0b12e399519a4013c52d4635ec8bdd6863/server/etcdserver/cluster_util.go#L222-L230
+		/*{
+			name:              "etcd instance with last version receives snapshot from the leader with current version",
+			clusterVersion:    e2e.CurrentVersion,
+			newInstaceVersion: e2e.LastVersion,
+		},*/
+		{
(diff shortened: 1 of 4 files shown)


ahrtr:
cc @tbg and @nvanbenschoten as well.

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,129 @@
...
+		name              string
+		clusterVersion    e2e.ClusterVersion
+		newInstaceVersion e2e.ClusterVersion
+	}{
+		// etcd doesn't support this scenario. For example, etcd cluster
+		// version is 3.6.x, then a new member of 3.5.x can't join the
+		// cluster. Please refer to link below,
+		// https://github.com/etcd-io/etcd/blob/3e903d0b12e399519a4013c52d4635ec8bdd6863/server/etcdserver/cluster_util.go#L222-L230
+		//
+		// It might be OK. Usually it's only possible to run into this
```
serathius, line 41: I think this can happen not only during downgrades but also during upgrades. Etcd supports hybrid version clusters during upgrade to ensure zero downtime upgrade. Snapshot can also be downloaded from leader when member is restarted after long enough time that leader has already pruned it's wal. 

So imagine the situation you are upgrading cluster, upgraded one member to new version. Upgraded member becomes a leader. If one of other members gets disconnected for long time it will download snapshot from upgraded member.

ahrtr: Actually I thought about intentionally creating network partition scenario to trigger at least a snapshot. But it needs a long time to write [5K](https://github.com/etcd-io/etcd/blob/f482f7b4a417de4920b213975822e8c83fca98e6/server/etcdserver/server.go#L83)+ entries. In my local env (32G memory, 8 Core 2.4GHz CPU), I created 10 goroutines, and it took more than 4 minutes to finish it (I did not get time to verify it in github workflow yet). Let's address this scenario in a separate PR, but probably it isn't a high priority task, please see my comment below.

> Upgraded member becomes a leader. If one of other members gets disconnected for long time it will download snapshot from upgraded member.

Yes, it's possible to run into the issue during upgrade in theory, but not in practice. It's more likely for upgraded member to run into issue instead of the members which are not upgraded yet. For example, for a cluster with 3 members, when upgrading the first member in a rolling upgrade, it's more likely for the first member to run into issue instead of other two members. And the first upgraded member is unlikely elected as leader if the cluster is under high load during upgrade, because its log isn't up-to-date as other members. 

Actually I think usually it's unlikely to trigger a snapshot during cluster upgrade, because (1) usually there is a maintenance window for cluster upgrade. It means there is no requesting coming from client. At least there shouldn't too many requests, because users will not perform the upgrade during busy hours.

serathius: > Actually I thought about intentionally creating network partition scenario to trigger at least a snapshot. But it needs a long time to write [5K](https://github.com/etcd-io/etcd/blob/f482f7b4a417de4920b213975822e8c83fca98e6/server/etcdserver/server.go#L83)+ entries.

Our functional tests make 2k qps. 10 goroutines writing is definetly not enough. Let's not make correctness dependent on qps

> Actually I think usually it's unlikely to trigger a snapshot during cluster upgrade, because (1) usually there is a maintenance window for cluster upgrade. It means there is no requesting coming from client. At least there shouldn't too many requests, because users will not perform the upgrade during busy hours.

I totally not agree. No modern distributed system assumes a maintenance window. There is no maintenance window in Kubernetes nor any database/key value store I know.

ahrtr: I agree with you in principle. 

Let me address the network partition scenario in a separate PR. Do you have any other comment on this PR.

ahrtr: Note: I removed the comment which you raised comment.

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestMixVersionsSendSnapshot(t *testing.T) {
+	cases := []struct {
+		name              string
+		clusterVersion    e2e.ClusterVersion
+		newInstaceVersion e2e.ClusterVersion
+	}{
+		// etcd doesn't support this scenario. For example, etcd cluster
```
serathius, line 36: This is still incorrect as I mentioned in comment above. I would be ok if you marked it as TODO.

ahrtr: Updated comment and also added TODO comment.

ahrtr: PTAL

serathius (approved):

PR #14712, opened 2022-11-09
keyIndex: print key and modified for String()

qsyqian:
Signed-off-by: qsyqian <qsyqian@gmail.com>

### DIFF
--- a/server/storage/mvcc/key_index.go
+++ b/server/storage/mvcc/key_index.go
@@ -329,6 +329,12 @@ func (ki *keyIndex) equal(b *keyIndex) bool {
 
 func (ki *keyIndex) String() string {
 	var s string
+	if len(ki.key) > 128 {
+		s += string(ki.key[:128]) + "...\n"
+	} else {
+		s += string(ki.key[:]) + "\n"
+	}
+	s += fmt.Sprintf("modified[%d]\n", ki.modified)
 	for _, g := range ki.generations {
 		s += g.String()
 	}


ahrtr (approved):
LGTM

Thank you @qsyqian

server/storage/mvcc/key_index.go
```diff
@@ -329,6 +329,8 @@ func (ki *keyIndex) equal(b *keyIndex) bool {
 
 func (ki *keyIndex) String() string {
 	var s string
+	s += string(ki.key[:]) + "\n"
```
ahrtr, line 332: After second thought, the len(ki.key) might be very large, so it may pollute the output. I would suggest to truncate it and add "..." if it exceeds a threshold, e.g. 128 bytes.

qsyqian: update, PTAL

server/storage/mvcc/key_index.go
```diff
@@ -329,6 +329,12 @@ func (ki *keyIndex) equal(b *keyIndex) bool {
 
 func (ki *keyIndex) String() string {
 	var s string
+	if len(ki.key) > 128 {
+		s += string(ki.key[:128]) + "...\n"
+	} else {
+		s += string(ki.key[:]) + "\n"
+	}
+	s += fmt.Sprintf("modified[%d]\n", ki.modified)
```
serathius, line 337: Not sure what's the reason to change `String` method, however the code:
* Looks more and more inefficient (strings a immutable in Go, so each `+=` does a full copy)
* will not work with zap correctly as newline characters will be escaped.

serathius (changes requested):
Don't see this as any enhancement and the increased performance cost just for `String` makes me worried about about impact of reliability. 

If you want to improve some log line then please make a direct change there, but not change `String` method that can be used for other purposes.

qsyqian:
> Don't see this as any enhancement and the increased performance cost just for `String` makes me worried about about impact of reliability.
> 
> If you want to improve some log line then please make a direct change there, but not change `String` method that can be used for other purposes.

i think you are right, just close this pr.

PR #14708, opened 2022-11-08
raft/tracker: use testify packages in tests

pavelkalinnikov:
This commit removes the verbose comparisons in tests, in favor of using assert/require helpers from the testify packages.

Part of #14709

Signed-off-by: Pavel Kalinnikov <pavel@cockroachlabs.com>

PR #14717, opened 2022-11-09
feat: use testify packages in tests

falser101:
Signed-off-by: jianfei.zhang <jianfei.zhang@daocloud.io>

### DIFF
--- a/raft/storage_test.go
+++ b/raft/storage_test.go
@@ -16,9 +16,9 @@ package raft
 
 import (
 	"math"
-	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	pb "go.etcd.io/etcd/raft/v3/raftpb"
 )
 
@@ -51,12 +51,8 @@ func TestStorageTerm(t *testing.T) {
 			}()
 
 			term, err := s.Term(tt.i)
-			if err != tt.werr {
-				t.Errorf("#%d: err = %v, want %v", i, err, tt.werr)
-			}
-			if term != tt.wterm {
-				t.Errorf("#%d: term = %d, want %d", i, term, tt.wterm)
-			}
+			require.Equal(t, tt.werr, err, "#%d", i)
+			require.Equal(t, tt.wterm, term, "#%d", i)
 		}()
 	}
 }
@@ -85,15 +81,11 @@ func TestStorageEntries(t *testing.T) {
 		{4, 7, uint64(ents[1].Size() + ents[2].Size() + ents[3].Size()), nil, []pb.Entry{{Index: 4, Term: 4}, {Index: 5, Term: 5}, {Index: 6, Term: 6}}},
 	}
 
-	for i, tt := range tests {
+	for _, tt := range tests {
 		s := &MemoryStorage{ents: ents}
 		entries, err := s.Entries(tt.lo, tt.hi, tt.maxsize)
-		if err != tt.werr {
-			t.Errorf("#%d: err = %v, want %v", i, err, tt.werr)
-		}
-		if !reflect.DeepEqual(entries, tt.wentries) {
-			t.Errorf("#%d: entries = %v, want %v", i, entries, tt.wentries)
-		}
+		require.Equal(t, tt.werr, err)
+		require.Equal(t, tt.wentries, entries)
 	}
 }
 
@@ -102,21 +94,13 @@ func TestStorageLastIndex(t *testing.T) {
 	s := &MemoryStorage{ents: ents}
 
 	last, err := s.LastIndex()
-	if err != nil {
-		t.Errorf("err = %v, want nil", err)
-	}
-	if last != 5 {
-		t.Errorf("last = %d, want %d", last, 5)
-	}
+	require.NoError(t, err)
+	require.Equal(t, uint64(5), last)
 
 	s.Append([]pb.Entry{{Index: 6, Term: 5}})
 	last, err = s.LastIndex()
-	if err != nil {
-		t.Errorf("err = %v, want nil", err)
-	}
-	if last != 6 {
-		t.Errorf("last = %d, want %d", last, 6)
-	}
+	require.NoError(t, err)
+	require.Equal(t, uint64(6), last)
 }
 
 func TestStorageFirstIndex(t *testing.T) {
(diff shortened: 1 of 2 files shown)


serathius:
This duplicates https://github.com/etcd-io/etcd/pull/14708

tbg:
https://github.com/etcd-io/etcd/pull/14708 is only for `raft/tracker`, this PR is for `raft`.

@falser101 could you rebase? Thanks!

falser101:
> #14708 is only for `raft/tracker`, this PR is for `raft`.
> 
> @falser101 could you rebase? Thanks!

rebased!

falser101:
@tbg

raft/storage_test.go
```diff
@@ -102,43 +94,27 @@ func TestStorageLastIndex(t *testing.T) {
...
 	s.Append([]pb.Entry{{Index: 6, Term: 5}})
 	last, err = s.LastIndex()
-	if err != nil {
-		t.Errorf("err = %v, want nil", err)
-	}
-	if last != 6 {
-		t.Errorf("last = %d, want %d", last, 6)
-	}
+	require.Equal(t, nil, err)
+	require.EqualValues(t, 6, last)
```
ahrtr, line 103:
```suggestion
	require.Equal(t, 6, last)
```

raft/storage_test.go
```diff
@@ -102,43 +94,27 @@ func TestStorageLastIndex(t *testing.T) {
...
 
 	last, err := s.LastIndex()
-	if err != nil {
-		t.Errorf("err = %v, want nil", err)
-	}
-	if last != 5 {
-		t.Errorf("last = %d, want %d", last, 5)
-	}
+	require.Equal(t, nil, err)
+	require.EqualValues(t, 5, last)
```
ahrtr, line 98:
```suggestion
	require.Equal(t, 5, last)
```

falser101:
down!  @ahrtr @tbg

ahrtr:
Please resolve the unit test failures.

falser101:
> Please resolve the unit test failures.

i don't know why this is failed
image
@ahrtr

raft/util_test.go
```diff
@@ -64,10 +56,8 @@ func TestLimitSize(t *testing.T) {
 		{uint64(ents[0].Size() + ents[1].Size() + ents[2].Size()), []pb.Entry{{Index: 4, Term: 4}, {Index: 5, Term: 5}, {Index: 6, Term: 6}}},
 	}
 
-	for i, tt := range tests {
-		if !reflect.DeepEqual(limitSize(ents, tt.maxsize), tt.wentries) {
-			t.Errorf("#%d: entries = %v, want %v", i, limitSize(ents, tt.maxsize), tt.wentries)
-		}
+	for _, tt := range tests {
+		require.Equal(t, limitSize(ents, tt.maxsize), tt.wentries)
```
ahrtr, line 60: Minor comment: Usually the first value is expected value, and the second one is actual one. 
```suggestion
		require.Equal(t, tt.wentries, limitSize(ents, tt.maxsize))
```

raft/util_test.go
```diff
@@ -97,10 +87,7 @@ func TestIsLocalMsg(t *testing.T) {
 		{pb.MsgPreVoteResp, false},
 	}
 
-	for i, tt := range tests {
-		got := IsLocalMsg(tt.msgt)
-		if got != tt.isLocal {
-			t.Errorf("#%d: got %v, want %v", i, got, tt.isLocal)
-		}
+	for _, tt := range tests {
+		require.Equal(t, IsLocalMsg(tt.msgt), tt.isLocal)
```
ahrtr, line 91: ditto
```suggestion
		require.Equal(t, tt.isLocal, IsLocalMsg(tt.msgt))
```

ahrtr:
Overall looks good to me. Once you resolve the two minor comments, then this  PR is ready to merge.

falser101:
> Overall looks good to me. Once you resolve the two minor comments, then this PR is ready to merge.

done！

ahrtr (approved):
LGTM

Thank you @falser101

falser101:
@tbg help review

raft/storage_test.go
```diff
@@ -102,43 +94,27 @@ func TestStorageLastIndex(t *testing.T) {
 	s := &MemoryStorage{ents: ents}
 
 	last, err := s.LastIndex()
-	if err != nil {
-		t.Errorf("err = %v, want nil", err)
-	}
-	if last != 5 {
-		t.Errorf("last = %d, want %d", last, 5)
-	}
+	require.Equal(t, nil, err)
```
fuweid, line 97: suggest to use `require.NoError(t, err)` here.

raft/storage_test.go
```diff
@@ -51,12 +51,8 @@ func TestStorageTerm(t *testing.T) {
 			}()
 
 			term, err := s.Term(tt.i)
-			if err != tt.werr {
-				t.Errorf("#%d: err = %v, want %v", i, err, tt.werr)
-			}
-			if term != tt.wterm {
-				t.Errorf("#%d: term = %d, want %d", i, term, tt.wterm)
-			}
+			require.Equal(t, tt.werr, err)
```
fuweid, line 54: maybe we should keep the index so that it is easier to find out which case failed, like

`require.Equal(t, tt.werr, err, "#%d", i)`

falser101:
@fuweid done!

raft/storage_test.go
```diff
@@ -102,43 +94,27 @@ func TestStorageLastIndex(t *testing.T) {
...
 
 	s.Append([]pb.Entry{{Index: 6, Term: 5}})
 	last, err = s.LastIndex()
-	if err != nil {
-		t.Errorf("err = %v, want nil", err)
-	}
-	if last != 6 {
-		t.Errorf("last = %d, want %d", last, 6)
-	}
+	require.Equal(t, nil, err)
```
fuweid, line 102: Sorry, I don't point it out in my last comment. I think it should be changed, and including the following lines.

falser101: already changed!

falser101:
done! @fuweid

tbg:
Thanks! I added another commit worth of polish, no need to rebase/amend, I'll merge when it's green (or maybe one of the other maint'ners can do it since I might be off for a few days).

ahrtr:
Thanks @fuweid and @tbg . 

Points:
1. We should always use [sub test](https://go.dev/blog/subtests) as much as we can. In this case, no need to add index as @fuweid pointed out above.
2. One PR is supposed to do only one thing. This PR is just to introduce testify in tests. I was think we should change the test to use subtest in a separate PR.

tbg:
> One PR is supposed to do only one thing. This PR is just to introduce testify in tests. I was think we should change the test to use subtest in a separate PR.

Don't fully agree - a *commit* should do one thing, as this eases reviewing and makes it more likely that problematic code can be backed out easily. I think it can be helpful to be able to group multiple commits into a PR. A PR has a fixed cost in terms of getting it through CI, review cycles, etc; splitting for the sake of splitting should not be a goal. I don't see what benefit it would have provided to open a separate PR just to introduce subtests; that PR would necessarily have conflicted with this one, which would've caused extra work due to the necessary conflict resolution.

ahrtr:
> a _commit_ should do one thing, as this eases reviewing and makes it more likely that problematic code can be backed out easily

Yes, sometimes it's also a valid option. There isn't a clear boundary between `One PR doing one thing` vs `one commit doing one thing`. Usually I prefer to the former if the PR is a big or complex one, and it would be better to breakdown the PR into small ones. If the PR is a small one (e.g. this PR), then it isn't a big deal. 

An observation: I reviewed lots of PR, and not all contributors can organize or group commits nicely, and the history isn't always unchanged. So usually I have to review the whole PR again and again to make sure I do not miss anything.

PR #14718, opened 2022-11-10
test: Create a subdirectory for each member when DataDirPath is specified

ahrtr:
When e2e test cases specify the DataDirPath, we need to create a subdirectory for each member, otherwise all members share the same directory and accordingly lead to conflict.

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/url"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


serathius:
Would it make more sense rather to remove the data dir argument from cluster? Setting only makese sense for single member.

tests/framework/e2e/cluster.go
```diff
@@ -423,9 +424,15 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
...
 	name := fmt.Sprintf("%s-test-%d", testNameCleanRegex.ReplaceAllString(tb.Name(), ""), i)
+
 	dataDirPath := cfg.DataDirPath
 	if cfg.DataDirPath == "" {
 		dataDirPath = tb.TempDir()
+	} else if cfg.ClusterSize > 1 {
+		// When test cases specify the DataDirPath and there are more than
+		// one member in the cluster, we create a subdirectory for each member
+		// to avoid conflict.
+		dataDirPath = filepath.Join(cfg.DataDirPath, name)
```
serathius, line 435: The end path that cluster member uses for data-dir should not be dependent on cluster size. Let's always add a subdirectory.

ahrtr: Done.

ahrtr:
> Would it make more sense rather to remove the data dir argument from cluster? Setting only makes sense for single member.

I think it makes sense to keep the flexibility. When I debug / execute E2E locally, I'd like to specify a data dir instead of letting e2e framework to create a temporary directory.

serathius (approved):

PR #14719, opened 2022-11-10
raft: rename raftLog.nextEnts to raftLog.nextCommittedEnts

nvanbenschoten:
Also rename `hasNextEnts` to `hasNextCommittedEnts`.
Also rename `maxNextEntsSize` to `maxNextCommittedEntsSize`.

Extracted from #14627. cc. @tbg 

Pure refactor.

Signed-off-by: Nathan VanBenschoten <nvanbenschoten@gmail.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #14721, opened 2022-11-10
raft: don't apply entries when applying snapshot

nvanbenschoten:
This commit removes the ability to apply log entries at the same time as applying a snapshot. Doing so it possible, but it leads to complex code and raises questions about what should be applied first. It also raises additional complexity when we start allowing concurrent, asynchronous log appends and log application. It's easiest to just disallow this.

Extracted from https://github.com/etcd-io/etcd/pull/14627.

cc. @tbg

### DIFF
--- a/raft/log.go
+++ b/raft/log.go
@@ -201,7 +201,7 @@ func (l *raftLog) hasNextCommittedEnts() bool {
 
 // hasPendingSnapshot returns if there is pending snapshot waiting for applying.
 func (l *raftLog) hasPendingSnapshot() bool {
-	return l.unstable.snapshot != nil && !IsEmptySnap(*l.unstable.snapshot)
+	return l.unstable.snapshot != nil
 }
 
 func (l *raftLog) snapshot() (pb.Snapshot, error) {
--- a/raft/log.go
+++ b/raft/log.go
@@ -181,9 +181,13 @@ func (l *raftLog) unstableEntries() []pb.Entry {
 // If applied is smaller than the index of snapshot, it returns all committed
 // entries after the index of snapshot.
 func (l *raftLog) nextCommittedEnts() (ents []pb.Entry) {
-	off := max(l.applied+1, l.firstIndex())
-	if l.committed+1 > off {
-		ents, err := l.slice(off, l.committed+1, l.maxNextCommittedEntsSize)
+	if l.hasPendingSnapshot() {
+		// See comment in hasNextCommittedEnts.
+		return nil
+	}
+	if l.committed > l.applied {
+		lo, hi := l.applied+1, l.committed+1 // [lo, hi)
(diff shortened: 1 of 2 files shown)


tbg:
Give me a ping when it's rebased, I'd like to glance at the diff again before merge. Thanks!

nvanbenschoten:
Rebased on main now that https://github.com/etcd-io/etcd/pull/14719 is merged. PTAL!

raft/log.go
```diff
@@ -181,9 +181,13 @@ func (l *raftLog) unstableEntries() []pb.Entry {
 // If applied is smaller than the index of snapshot, it returns all committed
```
tbg, line 181: stale comment.

raft/log.go
```diff
@@ -181,9 +181,13 @@ func (l *raftLog) unstableEntries() []pb.Entry {
...
 // entries after the index of snapshot.
 func (l *raftLog) nextCommittedEnts() (ents []pb.Entry) {
-	off := max(l.applied+1, l.firstIndex())
-	if l.committed+1 > off {
-		ents, err := l.slice(off, l.committed+1, l.maxNextCommittedEntsSize)
+	if l.hasPendingSnapshot() {
+		// See comment in hasNextCommittedEnts.
```
tbg, line 185: You could even structure this method as

1. call hasNextCommittedEnts; if false, return early.
2. otherwise, do the slice.

That way there's no duplication in logic between the two and one need not worry about them diverging via a future change.

tbg (approved):

raft/log.go
ahrtr: +1

raft/log.go
```diff
@@ -195,13 +199,18 @@ func (l *raftLog) nextCommittedEnts() (ents []pb.Entry) {
...
 func (l *raftLog) hasNextCommittedEnts() bool {
-	off := max(l.applied+1, l.firstIndex())
-	return l.committed+1 > off
+	if l.hasPendingSnapshot() {
+		// If we have a snapshot to apply, don't also return any committed
+		// entries. Doing so raises questions about what should be applied
+		// first.
+		return false
+	}
+	return l.committed > l.applied
```
ahrtr, line 208: It seems that `l.applied` will never smaller than `l.firstIndex()`, so the change makes sense to me. But I am curious why previous `max(l.applied+1, l.firstIndex()) `was added.

tbg: I think the first index only had an effect when there was an unstable snapshot, as this could bump the first index above the applied index. I was happy to see the simplification that avoids this special case.

ahrtr: Thanks for the clarification, which makes sense to me.

ahrtr (approved):
LGTM

Thank you @nvanbenschoten

PR #14723, opened 2022-11-10
raft: clean up IsLocalMsg and IsResponseMsg logic

nvanbenschoten:
Use array indexing to clean up the code and make it constant time.

Also, add a test for `IsResponseMsg`.

cc. @tbg 

### DIFF
--- a/raft/raftpb/raft.proto
+++ b/raft/raftpb/raft.proto
@@ -60,6 +60,9 @@ enum MessageType {
 	MsgReadIndexResp   = 16;
 	MsgPreVote         = 17;
 	MsgPreVoteResp     = 18;
+	// NOTE: when adding new message types, remember to update the isLocalMsg and
+	// isResponseMsg arrays in raft/util.go and update the corresponding tests in
+	// raft/util_test.go.
 }
 
 message Message {
--- a/raft/util.go
+++ b/raft/util.go
@@ -40,13 +40,34 @@ func max(a, b uint64) uint64 {
 	return b
 }
 
+var isLocalMsg = [...]bool{
+	pb.MsgHup:         true,
+	pb.MsgBeat:        true,
+	pb.MsgUnreachable: true,
+	pb.MsgSnapStatus:  true,
+	pb.MsgCheckQuorum: true,
+}
+
+var isResponseMsg = [...]bool{
+	pb.MsgAppResp:       true,
+	pb.MsgVoteResp:      true,
+	pb.MsgHeartbeatResp: true,
+	pb.MsgUnreachable:   true,
+	pb.MsgReadIndexResp: true,
+	pb.MsgPreVoteResp:   true,
+}
+
+func isMsgInArray(msgt pb.MessageType, arr []bool) bool {
+	i := int(msgt)
+	return i < len(arr) && arr[i]
+}
+
 func IsLocalMsg(msgt pb.MessageType) bool {
-	return msgt == pb.MsgHup || msgt == pb.MsgBeat || msgt == pb.MsgUnreachable ||
-		msgt == pb.MsgSnapStatus || msgt == pb.MsgCheckQuorum
+	return isMsgInArray(msgt, isLocalMsg[:])
 }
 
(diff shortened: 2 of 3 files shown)


tbg (approved):

ahrtr:
I would suggest code something like below. It has two slightly benefits over  @nvanbenschoten 's code:
1. It consumes smaller memory;
2. It doesn't compromise the data type. Specifically it doesn't cast pb.MessageType to `int`

```
var (
	localMsg = map[pb.MessageType]struct{}{
		pb.MsgHup:         struct{}{},
		pb.MsgBeat:        struct{}{},
		pb.MsgUnreachable: struct{}{},
		pb.MsgSnapStatus:  struct{}{},
		pb.MsgCheckQuorum: struct{}{},
	}

	responseMsg = map[pb.MessageType]struct{}{
		pb.MsgAppResp:       struct{}{},
		pb.MsgVoteResp:      struct{}{},
		pb.MsgHeartbeatResp: struct{}{},
		pb.MsgUnreachable:   struct{}{},
		pb.MsgReadIndexResp: struct{}{},
		pb.MsgPreVoteResp:   struct{}{},
	}
)

func isMsgInMap(msgt pb.MessageType, msgs map[pb.MessageType]struct{}) bool {
	_, ok := msgs[msgt]
	return ok
}

func IsLocalMsg(msgt pb.MessageType) bool {
	return isMsgInMap(msgt, localMsg)
}

func IsResponseMsg(msgt pb.MessageType) bool {
	return isMsgInMap(msgt, responseMsg)
}
```

tbg:
@ahrtr have to politely disagree, while access to a map and a slice is both amortized O(1) the constant for the slice access is *much* smaller and that's what counts here. I'm not even sure the map will be smaller, btw. At any rate, it seems that the solution already implemented is optimal on the dimension that matters so I would prefer we keep it.

ahrtr:
Thanks @tbg for the feedback, which partially makes sense to me. It's tradeoff between performance and readability. 

The key point is it (the code I proposed above) has a little better readability (Sorry for not to explicitly explain this in my previous comment). But it might be nitpicking, so **my comment isn't a blocker for merging this PR**. Our purpose is to check whether a message is a local message or a response message. We just need to maintain two sets of message types, one for the local messages, and the other for the response messages; and check whether a message belongs to one of the sets. My suggested code is more direct and clear. 

@nvanbenschoten 's code can definitely work, but it's a little indirect, and might cause confusion at first look. The message type is actually the index of the arrays(Note: they are arrays instead of slices), but it looks like a key in map at first look, this is the first point which might cause confusion, and according hurt the readability. Secondly, currently all the specified values are `true`, but it can be mistakenly specified as `false`, this is the second minor point which hurt readability & maintainability. Thirdly, we should focus on the message types themselves, instead of what are the values (`true` or `false`), and it's exactly the reason why I use struct{}{} as the value of the map in my proposed code, it means we don't care the values. Actually a [Set data type ](https://github.com/ahrtr/gocontainer/blob/06553eb6a662ce989f29c8338894dde5717b0639/set/set.go#L23-L36)makes more sense here, but unfortunately, golang doesn't have built-in set. 

Please also see my responses below,

 

> while access to a map and a slice is both amortized O(1) the constant for the slice access is _much_ smaller

Yes, it's a valid point to me. But I doubt that the overall performance impact might be negligible. 



> I'm not even sure the map will be smaller, btw

There are only 6 response message types, but this PR allocates 19 elements because the max index is 18 (`pb.MsgPreVoteResp`). On the other hand, a bool value consumes only 1 byte (you can confirm this using [unsafe#Sizeof](https://pkg.go.dev/unsafe#Sizeof) ), but it isn't clearly stated in [Golang spec](https://go.dev/ref/spec#Boolean_types), so it might change in future. But again, it's a little nitpicking.

tbg:
@ahrtr yeah, a lot of it is just matter of taste. Both solutions work here and even if one is superior in a microbenchmark to another, this would be meaningless in a real system. Thanks for your thoughts, with your permission I will merge this in as-is.

PR #14724, opened 2022-11-10
Reproduce revision inconsistency issue

ahrtr:

### DIFF
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -16,7 +16,6 @@ package linearizability
 
 import (
 	"context"
-	"fmt"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -52,6 +51,7 @@ func TestLinearizability(t *testing.T) {
 			config: e2e.EtcdProcessClusterConfig{
 				ClusterSize:   3,
 				GoFailEnabled: true,
+				DataDirPath:   "/Users/wachao/tmp/etcd/test/etcd",
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


tests/linearizability/traffic.go
```diff
@@ -33,7 +33,7 @@ type Traffic interface {
 type putGetTraffic struct{}
 
 func (t putGetTraffic) Run(ctx context.Context, c *recordingClient, limiter *rate.Limiter) {
-	maxOperationsPerClient := 1000000
+	maxOperationsPerClient := 10
```
serathius, line 36: Any reason to set this? This number is used to ensure each client write has unique value. This is done by shifting starting id value for each client and limiting number of writes. For example first client writes values `0...1000000`, second client writes `1000000...2000000` and so forth. 

Lowering this number just results in clients making 10 requests and finishing.

tests/linearizability/linearizability_test.go
```diff
@@ -94,30 +78,31 @@ func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProce
...
 	go func() {
-		defer cancel()
 		err := triggerFailpoints(ctx, t, clus, failpoint)
 		if err != nil {
 			t.Error(err)
 		}
 	}()
 	operations := simulateTraffic(ctx, t, clus, traffic)
-	clus.Close()
+	time.Sleep(10 * time.Second)
```
serathius, line 92: I would not recommend removing cluster.Stop here as this is needed to saftly copy out the data dir.

tests/linearizability/linearizability_test.go
```diff
@@ -94,30 +78,31 @@ func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProce
...
 
 	linearizable, info := porcupine.CheckOperationsVerbose(etcdModel, operations, 0)
 	if linearizable != porcupine.Ok {
-		t.Error("Model is not linearizable")
+		t.Error("###### Model is not linearizable")
 	}
 
-	path, err := filepath.Abs(filepath.Join(resultsDirectory, strings.Replace(t.Name(), "/", "_", -1)+".html"))
+	path, err := filepath.Abs(filepath.Join("/Users/wachao/tmp/etcd/test/", strings.Replace(t.Name(), "/", "_", -1)+".html"))
```
serathius, line 99: You can just use `RESULTS_DIR` environment variable

serathius:
Strange that you are able to reproduce issue locally as I don't see any changes that influences the test itself.

ahrtr:
This PR isn't targeting for merge. It's just show you how I reproduced the issue with simplified test case.

PR #14725, opened 2022-11-10
tests: refactor `EtcdProcessClusterConfig` using Functional Options Pattern

clarkfw:
Context #14711

This PR focuses on 2nd step: replaces 1/3(the rest is on the way) initialisations with `DefaultConfig`, `NewConfig` and `EPClusterOption`.

### DIFF
--- a/tests/e2e/cluster_downgrade_test.go
+++ b/tests/e2e/cluster_downgrade_test.go
@@ -130,11 +130,10 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 }
 
 func newCluster(t *testing.T, clusterSize int) *e2e.EtcdProcessCluster {
-	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t, &e2e.EtcdProcessClusterConfig{
-		ClusterSize:  clusterSize,
-		InitialToken: "new",
-		KeepDataDir:  true,
-	})
+	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t, nil,
+		e2e.WithClusterSize(clusterSize),
+		e2e.WithKeepDataDir(true),
+	)
 	if err != nil {
 		t.Fatalf("could not start etcd process cluster (%v)", err)
 	}
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -197,6 +197,10 @@ func WithConfig(cfg *EtcdProcessClusterConfig) EPClusterOption {
 	return func(c *EtcdProcessClusterConfig) { *c = *cfg }
 }
 
+func WithVersion(version ClusterVersion) EPClusterOption {
+	return func(c *EtcdProcessClusterConfig) { c.Version = version }
+}
+
 func WithDataDirPath(path string) EPClusterOption {
 	return func(c *EtcdProcessClusterConfig) { c.DataDirPath = path }
 }
@@ -303,7 +307,14 @@ func WithGoFailEnabled(enabled bool) EPClusterOption {
 
 // NewEtcdProcessCluster launches a new cluster from etcd processes, returning
 // a new EtcdProcessCluster once all nodes are ready to accept client requests.
-func NewEtcdProcessCluster(ctx context.Context, t testing.TB, cfg *EtcdProcessClusterConfig) (*EtcdProcessCluster, error) {
+func NewEtcdProcessCluster(ctx context.Context, t testing.TB, cfg *EtcdProcessClusterConfig, opts ...EPClusterOption) (*EtcdProcessCluster, error) {
+	if cfg == nil {
+		cfg = NewConfig(opts...)
+	} else {
+		for _, opt := range opts {
+			opt(cfg)
+		}
+	}
 	epc, err := InitEtcdProcessCluster(t, cfg)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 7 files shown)


tests/e2e/cluster_downgrade_test.go
```diff
@@ -130,11 +130,10 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 }
 
 func newCluster(t *testing.T, clusterSize int) *e2e.EtcdProcessCluster {
-	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t, &e2e.EtcdProcessClusterConfig{
-		ClusterSize:  clusterSize,
-		InitialToken: "new",
```
clarkfw, line 135: `InitialToken: "new"` is now redundant.
https://github.com/etcd-io/etcd/blob/2f558ca0dbf1217c1cc5b82a1e2aec428bf6a04f/tests/framework/e2e/cluster.go#L173-L178

tests/framework/e2e/cluster.go
```diff
@@ -306,6 +310,16 @@ func NewEtcdProcessCluster(ctx context.Context, t testing.TB, cfg *EtcdProcessCl
...
+func NewEtcdProcessClusterWithOptions(ctx context.Context, t testing.TB, opts ...EPClusterOption) (*EtcdProcessCluster, error) {
+	cfg := NewConfig(opts...)
+	epc, err := InitEtcdProcessCluster(t, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return StartEtcdProcessCluster(ctx, epc, cfg)
+}
+
```
clarkfw, line 322: For smooth transition, `NewEtcdProcessClusterWithOptions` was created.
After all initialisations are replaced, `NewEtcdProcessClusterWithOptions` will be renamed to `NewEtcdProcessCluster`, and the original `NewEtcdProcessCluster` will be deleted.

serathius (approved):

tests/framework/e2e/cluster.go
serathius: sg

clarkfw:
Just force-pushed to stay sync with main branch

tests/framework/e2e/cluster.go
ahrtr: I'd suggest to enhance the existing function `NewEtcdProcessCluster` instead of adding a new one. See code example below. Note that it has no any impact on existing test cases. 

```
func NewEtcdProcessCluster(ctx context.Context, t testing.TB, cfg *EtcdProcessClusterConfig, opts ...EPClusterOption) (*EtcdProcessCluster, error) {
	if cfg == nil {
		cfg = NewConfig(opts...)
	} else {
		for _, opt := range opts {
			opt(cfg)
		}
	}
	epc, err := InitEtcdProcessCluster(t, cfg)
	if err != nil {
		return nil, err
	}

	return StartEtcdProcessCluster(ctx, epc, cfg)
}
```

clarkfw: It looks much better👍 I'm gonna change code to this.

After the transition period, I envision it eventually going to be this:
```go
func NewEtcdProcessCluster(ctx context.Context, t testing.TB, opts ...EPClusterOption) (*EtcdProcessCluster, error) {
 	cfg := NewConfig(opts...)
 	epc, err := InitEtcdProcessCluster(t, cfg)
 	if err != nil {
 		return nil, err
 	}
 	return StartEtcdProcessCluster(ctx, epc, cfg)
 }
```

And there will be 2 ways to call `NewEtcdProcessCluster`:
```go
epc, err := NewEtcdProcessCluster(ctx, t,
 	e2e.WithKeepDataDir(true),
 	e2e.WithCorruptCheckTime(time.Second),
)
```
```go
epc, err := NewEtcdProcessCluster(ctx, t, e2e.WithConfig(cfg))
```
What do you think?

ahrtr: Basically makes sense to me. 



> And there will be 2 ways to call `NewEtcdProcessCluster`

There is only one way. 

```
epc, err := NewEtcdProcessCluster(ctx, t, e2e.WithConfig(cfg), e2e.WithKeepDataDir(true))
```

Ideally, if we follow this approach, then all fields in `EtcdProcessClusterConfig` shouldn't be exported, but of course it can be addressed in a separate PR.

ahrtr: > Ideally, if we follow this approach, then all fields in `EtcdProcessClusterConfig` shouldn't be exported, but of course it can be addressed in a separate PR.

This comment might not be correct. As long as we keep the `e2e.WithConfig`, then the fields are still needed to be exported.

ahrtr (approved):
LGTM

Thank you @clarkfw 

Please feel free to finish the proposal as your comment https://github.com/etcd-io/etcd/pull/14725#discussion_r1020724228 in a separate PR.

tests/framework/e2e/cluster.go
clarkfw: > This comment might not be correct. As long as we keep the `e2e.WithConfig`, then the fields are still needed to be exported.

@ahrtr  From my perspective, keeping `e2e.WithConfig` doesn't require fields in `EtcdProcessClusterConfig` to be exported. Did I overlook something?

```
cfg :=NewConfig(
	WithClientTLS(ClientTLS),
	WithIsPeerTLS(true),
)

cluster := NewEtcdProcessCluster(ctx, t, WithConfig(cfg))
```

clarkfw:
> LGTM
> 
> Thank you @clarkfw
> 
> Please feel free to finish the proposal as your comment [#14725 (comment)](https://github.com/etcd-io/etcd/pull/14725#discussion_r1020724228) in a separate PR.

Thanks! I'm still working on it.

tests/framework/e2e/cluster.go
ahrtr: Keeping the fields exported means keeping the flexibility to specify the values for fields directly, and then pass it to `NewEtcdProcessCluster` using `WithConfig(cfg)`

PR #14730, opened 2022-11-11
etcdserver: call the OnPreCommitUnsafe in unsafeCommit

ahrtr:
Fix the issue discovered in https://github.com/etcd-io/etcd/pull/14685

`unsafeCommit` is called by both `(*batchTxBuffered) commit` and `(*backend) defrag`. When users perform the defragmentation operation, etcd doesn't update the consistent index, because `unsafeCommit` doesn't call the `OnPreCommitUnsafe`. If etcd crashes(e.g. panicking) in the process for whatever reason, when etcd starts again it replays the WAL entries starting from the latest snapshot, accordingly it may re-apply entries which might have already been applied, eventually the revision isn't consistent with other members.

Refer to discussion in [pull/14685](https://github.com/etcd-io/etcd/pull/14685), especially the comment https://github.com/etcd-io/etcd/pull/14685#issuecomment-1310999115.

Signed-off-by: Benjamin Wang <wachao@vmware.com>

cc @mitake @ptabor @serathius @spzala

PR #13885, opened 2022-04-05
server: Add verification of whether lock was called within out outside of apply

serathius:
cc @ptabor @ahrtr

PR #14733, opened 2022-11-11
[3.5] etcdserver: call the OnPreCommitUnsafe in unsafeCommit

ahrtr:
Backport https://github.com/etcd-io/etcd/pull/14730 to 3.5.

`unsafeCommit` is called by both `(*batchTxBuffered) commit` and `(*backend) defrag`. When users perform the defragmentation operation, etcd doesn't update the consistent index. If etcd crashes(e.g. panicking) in the process for whatever reason, then etcd replays the WAL entries starting from the latest snapshot, accordingly it may re-apply entries which might have already been applied, eventually the revision isn't consistent with other members.

Refer to discussion in https://github.com/etcd-io/etcd/pull/14685

### DIFF
--- a/server/mvcc/backend/batch_tx.go
+++ b/server/mvcc/backend/batch_tx.go
@@ -329,10 +329,6 @@ func (t *batchTxBuffered) CommitAndStop() {
 }
 
 func (t *batchTxBuffered) commit(stop bool) {
-	if t.backend.hooks != nil {
-		t.backend.hooks.OnPreCommitUnsafe(t)
-	}
-
 	// all read txs must be closed to acquire boltdb commit rwlock
 	t.backend.readTx.Lock()
 	t.unsafeCommit(stop)
@@ -340,6 +336,9 @@ func (t *batchTxBuffered) commit(stop bool) {
 }
 
 func (t *batchTxBuffered) unsafeCommit(stop bool) {
+	if t.backend.hooks != nil {
+		t.backend.hooks.OnPreCommitUnsafe(t)
+	}
 	if t.backend.readTx.tx != nil {
 		// wait all store read transactions using the current boltdb tx to finish,
 		// then close the boltdb tx


serathius:
This is pretty low level change, how sure we are that no other things are affected by this change? I think we should validate it before we merge the PR. 

My suggestion, we should add more gofailpoints into the code and use linearizability tests to validate the fix. 

@ahrtr Can you propose more failpoint location that would thoroughly test backend transation commit?

ahrtr:
> @ahrtr Can you propose more failpoint location that would thoroughly test backend transation commit?

Sure. Will think about it in separate discussion session.

ahrtr:
> This is pretty low level change, how sure we are that no other things are affected by this change? I

This is a simple & safe change to me. I do not see any risk. Which part you do not have confidence?  There might be other issues which could be discovered by other failpoints, but this fix is straightforward and safe to me. 

We already two important fixes, this one and the auth one. I think we need to release 3.5.6 soon. Of course, if you want to do more test, It's OK to me. But we shouldn't wait until all failpoints test are done.

serathius:
> This is a simple & safe change to me.

This is what worrying to me. It means that our rudiments were wrong, the fundamentals that we built a large part of code. This means that it might have far reaching impact as the change influences a large part of the codebase. 

I just want us to be careful and consider what this change could be also impacting. Thus the idea to add more failpoints.

ahrtr:
> It means that our rudiments were wrong, the fundamentals that we built a large part of code. This means that it might have far reaching impact as the change influences a large part of the codebase.

I do not get your point. What's rudiments were wrong? This is just an corner case we missed previously.

serathius:
I mean that we should take similar steps as we did in https://github.com/etcd-io/etcd/pull/13885. That time we introduced validation by checking stacktrace during runtime. I would like to double check if we can roll out similar validation. For this issue I would like to propose adding more gofailpoints in processes touching backend like defrag.

ahrtr:
> For this issue I would like to propose adding more gofailpoints in processes touching backend like defrag.

I agree that we can think about adding more failpoints, but it should be discussed and tracked in separate session (e.g. https://github.com/etcd-io/etcd/issues/14735 ). So all discussion in this PR should be focusing on what's the impact this PR might cause. 



> This is pretty low level change, how sure we are that no other things are affected by this change?

The [unsafeCommit](https://github.com/etcd-io/etcd/blob/3cf2e79e16a7a0e08de0ac168c829b4df20ae644/server/storage/backend/batch_tx.go#L338) is only called in two places (see below). I just move the code of calling preCommitHood from `commit` to `unsafeCommit`, I don't see any risk here. Please let me know if you see a risk. 
  - In [commit](https://github.com/etcd-io/etcd/blob/3cf2e79e16a7a0e08de0ac168c829b4df20ae644/server/storage/backend/batch_tx.go#L334);
  - In [defrag](https://github.com/etcd-io/etcd/blob/3cf2e79e16a7a0e08de0ac168c829b4df20ae644/server/storage/backend/backend.go#L466).



> I would like to double check if we can roll out similar validation.

I am afraid there is no similar validation this time. The [defragment failpoints](https://github.com/etcd-io/etcd/blob/3cf2e79e16a7a0e08de0ac168c829b4df20ae644/tests/linearizability/failpoints.go#L35-L36) successfully discovered this issue, and it's good. But anyway, I added two other related failpoints in https://github.com/etcd-io/etcd/pull/14746.

serathius (approved):

Issue #13919, opened 2022-04-10
Cover learners with e2e tests


serathius:
Based on https://github.com/etcd-io/etcd/pull/13837 I expect we are missing some e2e testing.

TinySong:
@serathius  would be ok if I work on this?

serathius:
Sure, to give you some context, most of etcd e2e tests either run 1 or 3 member cluster. Most test scenarios would benefit if we add a case where there is an additional learner in cluster. For examples of test code I would look into code in `tests/common`.

TinySong:
Working on it

serathius:
@TinySong any progress?

halegreen:
want to work on this issue

ishan16696:
Hi @halegreen 
are you working on this ?
@serathius this PR: https://github.com/etcd-io/etcd/pull/14739 got closed
I can work on this, feel free to assign me

pacoxu:
kubeadm using etcd learner mode to join the control plane node.
- we add some e2e test in https://github.com/kubernetes/test-infra/pull/28574

Elbehery:
@jmhbnz can u assign me please ?

jmhbnz:
Absolutely! Thanks for volunteering 🙏🏻 

/assign @Elbehery

Elbehery:
Hello @serathius  

I need more context to resolve this, below are initial ideas

- add `IsLearner` field to https://github.com/etcd-io/etcd/blob/f37a4365ee824825984fd0660f970eb596cb02d5/tests/framework/e2e/etcd_process.go#L80 
- add `AddMemberAsLearner` to e2e interface

i am quiet lost, where to start .. also could you inline some specific cases to cover ? 

thanks :) 

cc @ahrtr @jmhbnz

serathius:
@Elbehery we would want to add learners to basic e2e tests scenarios that should not care that there is a learner in the cluster. Take a look into https://github.com/etcd-io/etcd/blob/main/tests/common/e2e_test.go on how we automatically generate subtests for different cluster configurations.

Elbehery:
Thanks @serathius  for your response .. 

I reviewed the tests yesterday .. So for example, for https://github.com/etcd-io/etcd/blob/main/tests/common/member_test.go, we should add a test that adds a `learner` member ? 

also to create an etcd learner member server process, do i follow the same `EtcdServerProcessConfig` while adding a boolean for `IsLearner` ? 

wdyt ?

[...]

PR #14739, opened 2022-11-12
tests:add learner e2e tests

halegreen:
add learner e2e tests, fix issue #13919

### DIFF
--- a/tests/common/e2e_test.go
+++ b/tests/common/e2e_test.go
@@ -51,6 +51,10 @@ func e2eClusterTestCases() []testCase {
 			name:   "ClientAutoTLS",
 			config: config.ClusterConfig{ClusterSize: 1, ClientTLS: config.AutoTLS},
 		},
+		{
+			name:   "ClusterWithLearner",
+			config: config.ClusterConfig{ClusterSize: 3, ClientTLS: config.AutoTLS, WithLearner: true},
+		},
 	}
 
 	if fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
--- a/tests/framework/config/cluster.go
+++ b/tests/framework/config/cluster.go
@@ -36,6 +36,7 @@ type ClusterConfig struct {
 	StrictReconfigCheck bool
 	AuthToken           string
 	SnapshotCount       int
+	WithLearner         bool
 
 	// ClusterContext is used by "e2e" or "integration" to extend the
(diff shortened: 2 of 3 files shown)


tests/framework/e2e/e2e.go
```diff
@@ -91,7 +91,24 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, opts ...config.
 	if err != nil {
 		t.Fatalf("could not start etcd integrationCluster: %s", err)
 	}
-	return &e2eCluster{t, *epc}
+	clus := &e2eCluster{t, *epc}
+	if cfg.WithLearner {
```
serathius, line 95: This is not a proper implementation of learner e2e. We should have a dedicated server that is in learner mode. We might need to modify tests so that they also do some validation on learner, for example check if it's up to date with main cluster.

tests/framework/config/cluster.go
```diff
@@ -36,6 +36,7 @@ type ClusterConfig struct {
 	StrictReconfigCheck bool
 	AuthToken           string
 	SnapshotCount       int
+	WithLearner         bool
```
serathius, line 39: Learner should be a running server instance as one of the processes. We should make it explicit if by renaming it to  `ClusterSize` -> `MemberCount` and `WIthLearner` -> `LearnerCount`.

tests/common/e2e_test.go
```diff
@@ -51,6 +51,10 @@ func e2eClusterTestCases() []testCase {
 			name:   "ClientAutoTLS",
 			config: config.ClusterConfig{ClusterSize: 1, ClientTLS: config.AutoTLS},
 		},
+		{
+			name:   "ClusterWithLearner",
+			config: config.ClusterConfig{ClusterSize: 3, ClientTLS: config.AutoTLS, WithLearner: true},
```
serathius, line 56:
```suggestion
			name:   "Learner",
			config: config.ClusterConfig{ClusterSize: 1, WithLearner: true},
```

No need to have 3 member cluster to test learner nor configure client TLS.

serathius:
Apart of expanding cluster configuration to cover learners, we should have an explicit e2e tests that takes grows cluster size via learners. For example start from cluster size 1 and while making writes add learner, promote learner, add another learner, promote them and validate that all writes are available on learners.

PR #14741, opened 2022-11-13
README: remove Tobias Grieger from the Emeritus Maintainers list

ahrtr:
Tobias Grieger is actively working on etcd/raft (thanks @tbg ), and has already been added back as a raft maintainer, so no reason to keep him in the Emeritus Maintainers list.

cc @tbg


tbg:
Thanks! I do wonder why we are even keeping that Emeritus list.

spzala:
Emeritus list gives good idea about the maintainers history and also recognizes their work. We should keep it.

PR #14742, opened 2022-11-13
tests/e2e: disable testShellCompletion if enable cov flag

fuweid:
The etcdctl and etcdutl built with `-tags cov` mode will show go-test result
after each execution, like

 ```
 ...

PASS
coverage: 0.0% of statements in ./...
```

Since the PASS is not real command, the `source completion` command will
fail with command-not-found error. And there is no easy way to disable
the (*testing.M).Run's output. Therefore, this patch uses build tag !cov
to disable cases when enable coverage.

Fixes: #14694

### DIFF
--- a/tests/e2e/ctl_v3_completion_test.go
+++ b/tests/e2e/ctl_v3_completion_test.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !cov
+// +build !cov
+
 package e2e
 
 import (
@@ -33,6 +36,14 @@ func TestUtlV3CompletionBash(t *testing.T) {
 	testShellCompletion(t, e2e.BinPath.Etcdutl, "bash")
 }
 
+// testShellCompletion can only run in non-coverage mode. The etcdctl and etcdutl
+// built with `-tags cov` mode will show go-test result after each execution, like
+//
+//	PASS
+//	coverage: 0.0% of statements in ./...
+//
+// Since the PASS is not real command, the `source completion" fails with
+// command-not-found error.
 func testShellCompletion(t *testing.T, binPath, shellName string) {
 	e2e.BeforeTest(t)
 

### CODE REVIEW

**Comment by serathius** [comment]
If `TestCtlV3CompletionBash` test is problematic when running coverage tests, maybe we could just disable the test during coverage instead? I don't think it makes sense to change `etcdctl` code just for sake of coverage testing.

**Comment by fuweid** [comment]
>  I don't think it makes sense to change `etcdctl` code just for sake of coverage testing.

```
➜  etcdctl git:(test-downgrade-with-latest-point-release) go test -tags cov -c -o ../bin/etcdctl_test
➜  etcdctl git:(test-downgrade-with-latest-point-release) ../bin/etcdctl_test | tail -n 1
PASS
```

I was thinking that it is fine to add to main_test.go. 

> If `TestCtlV3CompletionBash` test is problematic when running coverage tests, maybe we could just disable the test during coverage instead?

So is it ok to use build tag `!cov` to skip the case?

**Comment by serathius** [comment]
I don't see any benefits to testing coverage of shell completion. Complicating  test code just seems needless, if we can avoid it.

**Comment by ahrtr** [comment]
> I don't see any benefits to testing coverage of shell completion. Complicating test code just seems needless, if we can avoid it.

Makes sense to me.

**Comment by fuweid** [comment]
@serathius @ahrtr 

I updated it with build tag. Please take a look.

```bash
➜  etcd git:(main) cd ./tests/e2e
➜  e2e git:(main) go test -v -tags cov -run TestCtlV3CompletionBash
=== RUN   TestCtlV3CompletionBash
    before.go:35: Changing working directory to: /tmp/TestCtlV3CompletionBash3256034565/001
    ctl_v3_completion_test.go:49:
                Error Trace:    ctl_v3_completion_test.go:49
                                                        ctl_v3_completion_test.go:29
                Error:          Received unexpected error:
                                exit status 127
                Test:           TestCtlV3CompletionBash
--- FAIL: TestCtlV3CompletionBash (0.02s)
FAIL
exit status 1
FAIL    go.etcd.io/etcd/tests/v3/e2e    0.035s
➜  e2e git:(main) git checkout deflake-issue-14694
Switched to branch 'deflake-issue-14694'
➜  e2e git:(deflake-issue-14694) go test -v -tags cov -run TestCtlV3CompletionBash
testing: warning: no tests to run
PASS
ok      go.etcd.io/etcd/tests/v3/e2e    0.011s
```

serathius (approved):

tests/e2e/ctl_v3_completion_test.go
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !cov
+// +build !cov
```
ahrtr, line 16: No need to keep ` +build !cov`, it's just for backward compatibility.
```suggestion
```

ahrtr: It isn't a big deal. Please feel free to resolve it in a  separate PR if you want.

ahrtr (approved):
LGTM

Thank you @fuweid

PR #14744, opened 2022-11-13
tests: refactor `EtcdProcessClusterConfig` using Functional Options Pattern

clarkfw:
Context https://github.com/etcd-io/etcd/pull/14711

This PR continues to work on 2nd step: replaces some initialisations of `EtcdProcessClusterConfig` with DefaultConfig, NewConfig and EPClusterOption.

Eventually, `EtcdProcessClusterConfig` should look like:
```go
func NewEtcdProcessCluster(ctx context.Context, t testing.TB, opts ...EPClusterOption) (*EtcdProcessCluster, error) {
 	cfg := NewConfig(opts...)
 	epc, err := InitEtcdProcessCluster(t, cfg)
 	if err != nil {
 		return nil, err
 	}
 	return StartEtcdProcessCluster(ctx, epc, cfg)
 }
```
see https://github.com/etcd-io/etcd/pull/14725#discussion_r1020724228

### DIFF
--- a/tests/e2e/etcd_grpcproxy_test.go
+++ b/tests/e2e/etcd_grpcproxy_test.go
@@ -35,9 +35,7 @@ func TestGrpcProxyAutoSync(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	epc, err := e2e.NewEtcdProcessCluster(ctx, t, &e2e.EtcdProcessClusterConfig{
-		ClusterSize: 1,
-	})
+	epc, err := e2e.NewEtcdProcessCluster(ctx, t, nil, e2e.WithClusterSize(1))
 	require.NoError(t, err)
 	defer func() {
 		assert.NoError(t, epc.Close())
@@ -59,7 +57,7 @@ func TestGrpcProxyAutoSync(t *testing.T) {
 		assert.NoError(t, proxyProc.Stop())
 	}()
 
-	proxyCtl, err := e2e.NewEtcdctl(&e2e.EtcdProcessClusterConfig{}, []string{proxyClientURL})
+	proxyCtl, err := e2e.NewEtcdctl(e2e.DefaultConfig(), []string{proxyClientURL})
 	require.NoError(t, err)
 	err = proxyCtl.Put(ctx, "k1", "v1", config.PutOptions{})
 	require.NoError(t, err)
--- a/tests/framework/e2e/e2e.go
+++ b/tests/framework/e2e/e2e.go
@@ -48,14 +48,13 @@ func (e e2eRunner) BeforeTest(t testing.TB) {
 
 func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, opts ...config.ClusterOption) intf.Cluster {
 	cfg := config.NewClusterConfig(opts...)
-	e2eConfig := EtcdProcessClusterConfig{
-		InitialToken:               "new",
-		ClusterSize:                cfg.ClusterSize,
-		QuotaBackendBytes:          cfg.QuotaBackendBytes,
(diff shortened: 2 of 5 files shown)


tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -59,7 +57,7 @@ func TestGrpcProxyAutoSync(t *testing.T) {
 		assert.NoError(t, proxyProc.Stop())
 	}()
 
-	proxyCtl, err := e2e.NewEtcdctl(&e2e.EtcdProcessClusterConfig{}, []string{proxyClientURL})
```
clarkfw, line 62: Among all fields of `EtcdProcessClusterConfig`, only `ClientTLS`, `IsClientAutoTLS` and `IsClientCRL` are visited in the life cycle of `proxyCtl`, and these 3 fields are not modified by `e2e.DefaultConfig()`. It should be safe to replace `e2e.EtcdProcessClusterConfig{}` with `e2e.DefaultConfig()`,

tests/framework/e2e/e2e.go
```diff
@@ -87,7 +86,7 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, opts ...config.
 	default:
 		t.Fatalf("PeerTLS config %q not supported", cfg.PeerTLS)
 	}
-	epc, err := NewEtcdProcessCluster(ctx, t, &e2eConfig)
+	epc, err := NewEtcdProcessCluster(ctx, t, e2eConfig)
```
ahrtr, line 89: Can we update it for now? Eventually we will get rid of the third parameter `nil`.
```suggestion
	epc, err := NewEtcdProcessCluster(ctx, t, nil, WithConfig(&e2eConfig))
```

tests/framework/e2e/e2e.go
```diff
@@ -48,14 +48,13 @@ func (e e2eRunner) BeforeTest(t testing.TB) {
...
-		AuthTokenOpts:              cfg.AuthToken,
-		SnapshotCount:              cfg.SnapshotCount,
-	}
+	e2eConfig := NewConfig(
+		WithClusterSize(cfg.ClusterSize),
+		WithQuotaBackendBytes(cfg.QuotaBackendBytes),
+		WithDisableStrictReconfigCheck(!cfg.StrictReconfigCheck),
+		WithAuthTokenOpts(cfg.AuthToken),
+		WithSnapshotCount(cfg.SnapshotCount),
+	)
```
ahrtr, line 57: Either way can work, it seems that I can we can just keep it as it's for now.

ahrtr (approved):
Overall looks good to me. 

Since it's still in progress of evolution, so it's OK for now.

ahrtr:
cc @serathius

tests/framework/e2e/e2e.go
clarkfw: The next PR will replaces ALL initialisations of `EtcdProcessClusterConfig` and remove the `nil` parameter.

tests/framework/e2e/e2e.go
clarkfw: Removing anti-pattern-naming fields(`DisableStrictReconfigCheck` and `NoCN`) is part of the motivation. To achieve that, I think all initialisations of `EtcdProcessClusterConfig` should be replaced by `DefaultConfig()` or `NewConfig `.

```go
func DefaultConfig() *EtcdProcessClusterConfig {
	return &EtcdProcessClusterConfig{
		ClusterSize:         3,
		InitialToken:        "new",
		CN:                  true,
		StrictReconfigCheck: true,
	}
}

func NewConfig(opts ...EPClusterOption) *EtcdProcessClusterConfig {
	c := DefaultConfig()
	for _, opt := range opts {
		opt(c)
	}
	return c
}
```

PR #14754, opened 2022-11-14
client: Use func without format

sashamelentyev:
Signed-off-by: Sasha Melentyev <sasha@melentyev.io>


ahrtr (approved):
LGTM

Thank you @sashamelentyev

PR #14755, opened 2022-11-14
tests: Make linearizability traffic extendable

serathius:
cc @ahrtr @geetasg

### DIFF
--- a/tests/linearizability/traffic.go
+++ b/tests/linearizability/traffic.go
@@ -17,48 +17,91 @@ package linearizability
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"time"
 
 	"golang.org/x/time/rate"
 )
 
 var (
-	PutGetTraffic Traffic = putGetTraffic{}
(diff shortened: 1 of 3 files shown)


tests/linearizability/traffic.go
```diff
@@ -17,48 +17,91 @@ package linearizability
...
 	"fmt"
+	"math/rand"
 	"time"
 
 	"golang.org/x/time/rate"
 )
 
 var (
-	PutGetTraffic Traffic = putGetTraffic{}
+	DefaultTraffic Traffic = readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 100}}}
```
ahrtr, line 27: The only comment is not sure how will you extend the operation types (e.g, delete, txn). It may be difficult to extend the lineariziablity mode.

ahrtr (approved):
Overall looks good to me with comment, which can be discussed separately. 

Note that it isn't good to add too many code with a high level design diagram, otherwise it will be difficult for new contributors to warm up.

PR #14766, opened 2022-11-15
Avoid getting gofail/runtime when failpoints are disabled

serathius:
cc @ahrtr


scripts/build_lib.sh
```diff
@@ -19,22 +19,24 @@ toggle_failpoints() {
...
-    ) || exit 2
-    (
-      cd ./etcdctl
-      run go get go.etcd.io/gofail/runtime
-    ) || exit 2
-    (
-      cd ./tests
-      run go get go.etcd.io/gofail/runtime
-    ) || exit 2
+    if [[ "$mode" == "enable" ]]; then
```
ahrtr, line 22: Please move `run gofail "$mode" server/etcdserver/ server/storage/backend/` (line 21) inside the `if` statement as well.

ahrtr:
I might be a little better to rewrite the logic `toggle_failpoints` something like below,
```
    if [[ "$mode" == "enable" ]]; then
        if command -v gofail >/dev/null 2>&1; then
              run gofail "$mode" server/etcdserver/ server/storage/backend/
              (
                cd ./server
                run go get go.etcd.io/gofail/runtime
              ) || exit 2
              (
                cd ./etcdutl
                run go get go.etcd.io/gofail/runtime
              ) || exit 2
              (
                cd ./etcdctl
                run go get go.etcd.io/gofail/runtime
              ) || exit 2
              (
                cd ./tests
                run go get go.etcd.io/gofail/runtime
              ) || exit 2
        else
            log_error "FAILPOINTS set but gofail not found"
        fi
    fi
```

scripts/build_lib.sh
serathius: `run gofail "$mode" server/etcdserver/ server/storage/backend/` supports both `mode=enable` and `mode=disable`. Then mode is disable gofail will revert the failpoint code back to comments. I think this is still useful.

ahrtr: Makes sense.

ahrtr (approved):
LGTM

Issue #14764, opened 2022-11-15
Manipulating Keys and Manipulating Users Concurrency Causes Segmentation Violation in Client


Magnitus-:
### What happened?

I created a bunch of keys concurrently with goroutines. I created a bunch of users concurrency with goroutines in a separate process.

The process generating the keys aborted with this:
```
{"level":"warn","ts":"2022-11-15T08:33:12.382-0500","logger":"etcd-client","caller":"v3@v3.5.5/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc000215340/127.0.0.1:32379","attempt":0,"error":"rpc error: code = InvalidArgument desc = etcdserver: revision of auth store is old"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x95a390]

goroutine 29 [running]:
go.etcd.io/etcd/client/v3.(*Client).unaryClientInterceptor.func1({0xb7ee68?, 0xc000122000?}, {0xaa0935, 0x14}, {0xa79540, 0xc0002ff1a0}, {0xa6f500, 0xc000376e70}, 0xc000266700, 0xadf000, ...)
	/home/eric/go/pkg/mod/go.etcd.io/etcd/client/v3@v3.5.5/retry_interceptor.go:81 +0x7f0
google.golang.org/grpc.(*ClientConn).Invoke(0xc000078a98?, {0xb7ee68?, 0xc000122000?}, {0xaa0935?, 0x8?}, {0xa79540?, 0xc0002ff1a0?}, {0xa6f500?, 0xc000376e70?}, {0xf69120, ...})
	/home/eric/go/pkg/mod/google.golang.org/grpc@v1.41.0/call.go:35 +0x21f
go.etcd.io/etcd/api/v3/etcdserverpb.(*kVClient).Put(0xc000139230, {0xb7ee68, 0xc000122000}, 0x40f107?, {0xf69120, 0x3, 0x3})
	/home/eric/go/pkg/mod/go.etcd.io/etcd/api/v3@v3.5.5/etcdserverpb/rpc.pb.go:6469 +0xc9
go.etcd.io/etcd/client/v3.(*retryKVClient).Put(0xc000123bf0?, {0xb7ee68?, 0xc000122000?}, 0xc000078b38?, {0xf69120?, 0x8?, 0x0?})
[...]

PR #14770, opened 2022-11-16
clientv3: do not refresh token when users use CommonName based authentication

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/14764


ahrtr:
When users use CommonName based authentication, then no need to fresh token at all. And actually the [authTokenBundle](https://github.com/etcd-io/etcd/blob/8438ddae3e738c1ed0c92f67b511741a08ce1edd/client/v3/client.go#L70) is always nil, so the client will crash when refreshing the token. 

We do not have any e2e test cases to cover the Common Name based authentication so far. So I added the first one in this PR.

@mitake  Please take a look at this PR. thx. cc @ptabor @serathius @spzala

ahrtr:
I just noticed that we already have 3 cases to cover the CommonName based authentication in [ctl_v3_auth_no_proxy_test.go](https://github.com/etcd-io/etcd/blob/main/tests/e2e/ctl_v3_auth_no_proxy_test.go), it's good news.

tests/e2e/ctl_v3_auth_no_proxy_test.go
```diff
@@ -34,3 +39,103 @@ func TestCtlV3AuthCertCNAndUsername(t *testing.T) {
...
+func TestCtlV3AuthCertCNWithWithConcurrentOperation(t *testing.T) {
+	e2e.BeforeTest(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// apply the certificate which has `root` CommonName,
+	// and reset the setting when the test case finishes.
+	t.Log("Apply certificate with root CommonName")
+	resetCert := applyTLSWithRootCommonName()
+	defer resetCert()
```
serathius, line 52: There should be a separate cert bundle and an option in e2e tests pick certs with CommonName.

ahrtr: I thought about enhancing the e2e framework, but it needs more time. I will think about it in a separate PR. Let's get 3.5.6 out firstly.

ahrtr: Just rebased this PR, and added a TODO comment. PTAL.

serathius: Sounds good.

serathius (approved):

mitake:
@ahrtr Thanks a lot for fixing the issue, the change LGTM

ahrtr:
> @ahrtr Thanks a lot for fixing the issue, the change LGTM

Thanks @mitake for the confirmation.

PR #14788, opened 2022-11-16
all: goimports -w .

sashamelentyev:
Signed-off-by: Sasha Melentyev <sasha@melentyev.io>

### DIFF
--- a/tests/integration/clientv3/connectivity/dial_test.go
+++ b/tests/integration/clientv3/connectivity/dial_test.go
@@ -23,7 +23,7 @@ import (
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/transport"
-	"go.etcd.io/etcd/client/v3"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
 	clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"
--- a/client/v3/clientv3util/example_key_test.go
+++ b/client/v3/clientv3util/example_key_test.go
@@ -18,7 +18,7 @@ import (
 	"context"
 	"log"
 
-	"go.etcd.io/etcd/client/v3"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/clientv3util"
 )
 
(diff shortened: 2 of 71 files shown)


tests/integration/clientv3/connectivity/dial_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	"go.etcd.io/etcd/client/v3"
 	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
-	clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"
+	"go.etcd.io/etcd/tests/v3/integration/clientv3"
```
ahrtr, line 29: It's updated by `goimports -w` automatically? Why did previous file add `clientv3test`, while this file remove `clientv3test`?

sashamelentyev: goimports -w add this:
```go
	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
	"go.etcd.io/etcd/tests/v3/framework/testutils"
	"go.etcd.io/etcd/tests/v3/integration/clientv3"
	clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"
	"google.golang.org/grpc"
```

Mb I remove changes in tests?

ptabor: We want to keep 

```
clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"
```

And change in this file goes in the other direction. So it's why we are surprised.

sashamelentyev: Fixed.

I add `clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"`myself and run goimports and all ok. Rerun goimports have no changes

ptabor:
Please add it to: 
https://github.com/etcd-io/etcd/blob/main/scripts/fix.sh

as this change makes sense if we want to keep it consistent going forward. 

And the benefit of this change is that there is less 'surprise' that the `/client/v3` package is not addressable by `v3`.

sashamelentyev:
> Please add it to: https://github.com/etcd-io/etcd/blob/main/scripts/fix.sh
> 
> as this change makes sense if we want to keep it consistent going forward.
> 
> And the benefit of this change is that there is less 'surprise' that the `/client/v3` package is not addressable by `v3`.

After `goimports -w .` have changes in *.pb.go files...

ahrtr (approved):
LGTM

Thank you @sashamelentyev 

Please consider to resolve https://github.com/etcd-io/etcd/pull/14788#issuecomment-1318883581 suggested by @ptabor , of course in a separate PR.

PR #14798, opened 2022-11-17
Fix TestCtlV3Lock/Elect flakes

tjungblu:
waiting for a proper exit after a signal should ensure the goroutines don't leak.

### DIFF
--- a/tests/e2e/ctl_v3_lock_test.go
+++ b/tests/e2e/ctl_v3_lock_test.go
@@ -69,7 +69,12 @@ func testLock(cx ctlCtx) {
 	if err != nil {
 		cx.t.Fatal(err)
 	}
-	defer blockAcquire.Stop()
+	defer func(blockAcquire *expect.ExpectProcess) {
+		err = blockAcquire.Stop()
+		require.NoError(cx.t, err)
+		blockAcquire.Wait()
+	}(blockAcquire)
+
 	select {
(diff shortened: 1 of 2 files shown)


tjungblu:
the grpc proxy one seems unrelated, referring to the watch cache. Could you please re-run the e2e and the grpc tests a couple times to ensure it doesn't flake? I'm unable to trigger :/

serathius:
First pass e2e tests passed, rerunning to see if they pass the second time.

tests/e2e/ctl_v3_lock_test.go
```diff
@@ -69,7 +69,12 @@ func testLock(cx ctlCtx) {
 	if err != nil {
 		cx.t.Fatal(err)
 	}
-	defer blockAcquire.Stop()
+	defer func(blockAcquire *expect.ExpectProcess) {
```
serathius, line 72: Should this be in defer? Would putting it after select that reads from `ch` enough?

tjungblu: true, it makes more sense to not defer stop here

tjungblu: hmm, seems like there was some reason after all:
> 2022-11-17T13:44:28.5305567Z     ctl_v3_lock_test.go:106: expected different lock name, got l1="a/186d8485d2349505\r\n", l2="Error: context canceled\r\n"

PR #14802, opened 2022-11-17
linearizability tests - Add support for delete api

geetasg:

### DIFF
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -94,9 +94,6 @@ func (c *recordingClient) Delete(ctx context.Context, key string) error {
 	callTime := time.Now()
 	resp, err := c.client.Delete(ctx, key)
 	returnTime := time.Now()
-	if err != nil {
-		return err
-	}
 	var revision int64
 	var deleted int64
 	if resp != nil && resp.Header != nil {
@@ -107,7 +104,7 @@ func (c *recordingClient) Delete(ctx context.Context, key string) error {
 		ClientId: c.id,
 		Input:    etcdRequest{op: Delete, key: key},
 		Call:     callTime.UnixNano(),
-		Output:   etcdResponse{revision: revision, deleted: deleted},
+		Output:   etcdResponse{revision: revision, deleted: deleted, err: err},
 		Return:   returnTime.UnixNano(),
 	})
 	return nil
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -89,3 +89,26 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
 	})
 	return nil
 }
+
+func (c *recordingClient) Delete(ctx context.Context, key string) error {
+	callTime := time.Now()
+	resp, err := c.client.Delete(ctx, key)
+	returnTime := time.Now()
+	if err != nil {
+		return err
+	}
+	var revision int64
+	var deleted int64
+	if resp != nil && resp.Header != nil {
+		revision = resp.Header.Revision
+		deleted = resp.Deleted
+	}
+	c.operations = append(c.operations, porcupine.Operation{
+		ClientId: c.id,
+		Input:    etcdRequest{op: Delete, key: key},
+		Call:     callTime.UnixNano(),
+		Output:   etcdResponse{revision: revision, deleted: deleted},
+		Return:   returnTime.UnixNano(),
+	})
+	return nil
+}
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -133,7 +133,6 @@ func initState(request etcdRequest, response etcdResponse) EtcdState {
 		if response.err != nil {
 			state.FailedWrites[""] = struct{}{}
 		}
-		//TODO preserve information about failed deletes
 	default:
 		panic("Unknown operation")
 	}
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -130,8 +130,10 @@ func initState(request etcdRequest, response etcdResponse) EtcdState {
 			state.FailedWrites[request.putData] = struct{}{}
 		}
 	case Delete:
-		//TODO should state have 'deleted' field?
-		state.Value = ""
+		if response.err != nil {
+			state.FailedWrites[""] = struct{}{}
+		}
+		//TODO preserve information about failed deletes
 	default:
 		panic("Unknown operation")
 	}
@@ -167,12 +169,26 @@ func stepPut(state EtcdState, request etcdRequest, response etcdResponse) (bool,
 
 func stepDelete(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
 	if response.err != nil {
-		state.FailedWrites[request.putData] = struct{}{}
+		state.FailedWrites[""] = struct{}{}
 		return true, state
 	}
-	if response.deleted == 1 && state.LastRevision >= response.revision {
+	deleteSucceeded := response.deleted != 0
+	keySet := state.Value != ""
+
+	//non-existent key cannot be deleted.
+	if deleteSucceeded != keySet {
 		return false, state
 	}
+	//if key was deleted, response revision should go up
+	if deleteSucceeded && state.LastRevision >= response.revision {
+		return false, state
+	}
+	//if key was not deleted, response revision should not change
+	if !deleteSucceeded && state.LastRevision != response.revision {
+		return false, state
+	}
+
+	state.Value = ""
 	state.LastRevision = response.revision
 	return true, state
 }
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -24,8 +24,9 @@ import (
 type Operation string
 
 const (
-	Get Operation = "get"
-	Put Operation = "put"
+	Get    Operation = "get"
+	Put    Operation = "put"
+	Delete Operation = "delete"
 )
 
 type etcdRequest struct {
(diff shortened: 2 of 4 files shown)


ahrtr:
Please mark it `ready for review` when it's ready.

tests/linearizability/model.go
```diff
@@ -119,6 +129,9 @@ func initState(request etcdRequest, response etcdResponse) EtcdState {
 		} else {
 			state.FailedWrites[request.putData] = struct{}{}
 		}
+	case Delete:
+		//TODO should state have 'deleted' field?
```
serathius, line 133: Not needed.

tests/linearizability/model.go
```diff
@@ -119,6 +129,9 @@ func initState(request etcdRequest, response etcdResponse) EtcdState {
...
+		state.Value = ""
```
serathius, line 134:
```suggestion
		if response.err != nil {
		  state.FailedWrites[""] = struct{}{}
		}
```

Delete can fail, so we should preserve information that.

tests/linearizability/model.go
```diff
@@ -151,3 +164,15 @@ func stepPut(state EtcdState, request etcdRequest, response etcdResponse) (bool,
 	state.LastRevision = response.revision
 	return true, state
 }
+
+func stepDelete(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
+	if response.err != nil {
+		state.FailedWrites[request.putData] = struct{}{}
```
serathius, line 170:
```suggestion
		state.FailedWrites[""] = struct{}{}
```
It's true that `request.putData` for delete request will be empty, but let's not use fields meant only for put requests.

tests/linearizability/model.go
```diff
@@ -151,3 +164,15 @@ func stepPut(state EtcdState, request etcdRequest, response etcdResponse) (bool,
...
+		return true, state
+	}
+	if response.deleted == 1 && state.LastRevision >= response.revision {
```
serathius, line 173:
```suggestion
    deleteSucceded := response.deleted != 0
    keySet := state.Value != ""
	if (deleteSucceded != keySet) || state.LastRevision >= response.revision {
```

Delete request is invalid if either:
* response says there was a key deleted (`reponse.deleted != 0`) but the key was not present (`state.Value == ""`)
* response says there was a key was not deleted (`reponse.deleted == 0`) but the key was present (`state.Value != ""`)
* revision has not increased  `state.LastRevision >= response.revision`

all cases:

|state.Value | reponse.deleted | correct? |  explanation  |
|-----------------|-------------------------|--------------|-------------------|
| "a"                | 1                           | true         | if key was set, delete should return that it was deleted.
| "a"                | 0                           | false        | if key was set, delete cannot return that it didn't delete any keys
| ""                  | 1                           | false        | if key was not set, delete cannot return that it deleted it.
| ""                  | 0                           | true         | if key was not set, delete should report that it didn't delete anything

tests/linearizability/model_test.go
```diff
@@ -104,6 +104,18 @@ func TestModel(t *testing.T) {
...
+			operations: []testOperation{
+				{req: etcdRequest{op: Delete, key: "NotThere"}, resp: etcdResponse{deleted: 0, revision: 4}},
+			},
+		},
+		{
+			name: "Deleting  existent key bumps up revision",
+			operations: []testOperation{
+				{req: etcdRequest{op: Delete, key: "key"}, resp: etcdResponse{deleted: 1, revision: 5}},
+			},
+		},
```
serathius, line 118:
```suggestion
		{
			name: "Delete only increases revision on success",
			operations: []testOperation{
				{req: etcdRequest{op: Put, key: "key", putData: "1"}, resp: etcdResponse{revision: 1}},
				{req: etcdRequest{op: Delete, key: "key"}, resp: etcdResponse{deleted: 1, revision: 1}, failure: true},
				{req: etcdRequest{op: Delete, key: "key"}, resp: etcdResponse{deleted: 1, revision: 2}},
				{req: etcdRequest{op: Delete, key: "key"}, resp: etcdResponse{deleted: 0, revision: 3}, failure: true},
				{req: etcdRequest{op: Delete, key: "key"}, resp: etcdResponse{deleted: 0, revision: 2}},
			},
		},

```
Let's combine the cases

serathius:
You still need to integrate tests into `./tests/linearizability/traffic.go`. 
You can just change 
```
DefaultTraffic Traffic = readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 100}}}
```
to
```
DefaultTraffic Traffic = readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 90}, {operation: Delete, chance: 10}}}
```

And add a case `Delete` to switch within  (t readWriteSingleKey) Write` function to that will call `client.Delete`.

serathius:
We are also missing tests for failed `Delete` request, but those are hard to write due to how model works. I can fill them later.

tests/linearizability/client.go
```diff
@@ -89,3 +89,26 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
 	})
 	return nil
 }
+
+func (c *recordingClient) Delete(ctx context.Context, key string) error {
+	callTime := time.Now()
+	resp, err := c.client.Delete(ctx, key)
+	returnTime := time.Now()
+	if err != nil {
+		return err
```
serathius, line 98: Even on error you should record an Delete operation. Reason is if the error was connection issue the delete may still be persisted in etcd.

tests/linearizability/model.go
```diff
@@ -119,6 +129,11 @@ func initState(request etcdRequest, response etcdResponse) EtcdState {
 		} else {
 			state.FailedWrites[request.putData] = struct{}{}
 		}
+	case Delete:
+		if response.err != nil {
+			state.FailedWrites[""] = struct{}{}
+		}
+		//TODO preserve information about failed deletes
```
serathius, line 136:
```suggestion
```
Not needed the lines above should handle it

serathius:
Overall it looks good, just need to handle delete errors. Great job!

serathius:
Checked the linearizability failure. As expected it's an issue with handling Delete error.

image

tests/linearizability/client.go
```diff
@@ -89,3 +89,26 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
...
+		revision = resp.Header.Revision
+		deleted = resp.Deleted
+	}
+	c.operations = append(c.operations, porcupine.Operation{
+		ClientId: c.id,
+		Input:    etcdRequest{op: Delete, key: key},
+		Call:     callTime.UnixNano(),
+		Output:   etcdResponse{revision: revision, deleted: deleted},
+		Return:   returnTime.UnixNano(),
+	})
```
serathius, line 112:
```suggestion
	var revision int64
	var deleted int64
	if resp != nil && resp.Header != nil {
		revision = resp.Header.Revision
		deleted = resp.Deleted
	}
	c.operations = append(c.operations, porcupine.Operation{
		ClientId: c.id,
		Input:    etcdRequest{op: Delete, key: key},
		Call:     callTime.UnixNano(),
		Output:   etcdResponse{revision: revision, deleted: deleted, err: err},
		Return:   returnTime.UnixNano(),
	})
```

serathius:
Can you fix DCO check? you just need to do `git fetch upstream && git checkout origin/delete_api && git rebase origin/main --signoff && git push origin HEAD:delete_api -f`

serathius:
I started testing the model and found issues related to put request failed but was persisted and following delete checks if object was deleted.

geetasg:
notes -
test using 

       pushd tools/mod; go install go.etcd.io/gofail; popd
        mkdir -p /tmp/linearizability
        FAILPOINTS=true make build
        cat server/etcdserver/raft.fail.go
        EXPECT_DEBUG=true GO_TEST_FLAGS=-v RESULTS_DIR=/tmp/linearizability make test-linearizability

or - 
make gofail-enable
make build
GO_TEST_FLAGS=-v make test-linearizability

geetasg:
Marking ready for review based on checked - https://github.com/etcd-io/etcd/pull/14802/checks 
Please take a look @serathius @ahrtr

tests/linearizability/model.go
```diff
@@ -151,3 +165,29 @@ func stepPut(state EtcdState, request etcdRequest, response etcdResponse) (bool,
 	state.LastRevision = response.revision
 	return true, state
 }
+
+func stepDelete(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
+	if response.err != nil {
+		state.FailedWrites[""] = struct{}{}
```
ahrtr, line 171:
```suggestion
		state.FailedWrites[state.Value] = struct{}{}
```

ahrtr: I think we need to record whatever value in state in `FailedWrites`.

ahrtr (approved):
Overall looks good to me with a minor comment.

Thank you @geetasg 

This PR has some conflict with https://github.com/etcd-io/etcd/pull/14880. I tend to let this PR in firstly, and then update 14880 afterwards.

serathius:
FYI, this PR doesn't implement correct  handling of failed delete requests. I looked into fixing this, however code got so complicated that I decided to change our approach of handling errors. This PR without https://github.com/etcd-io/etcd/pull/14880 will introduce flakiness.

I would be ok to merge this PR first assuming that we will merge https://github.com/etcd-io/etcd/pull/14880 right after.

serathius:
All PRs approved, let's start merging.

Issue #14809, opened 2022-11-19
Support paginating by key-value size


linxiulei:
### What would you like to be added?

Allow RangeRequest to specify a size limit to paginate the result if the size of result exceeds the limit.

### Why is this needed?

This would allow clients to constrain the response size to avoid overwhelming etcd server with expensive requests.

linxiulei:
cc @serathius

lavacat:
Is this for Kubernetes use case or general case?

FYI, Kubernetes recently [added](https://github.com/kubernetes/apiserver/commit/f4edb394b0e83de067176cccc2fbefea217632a8) `maxLimit = 10000`

fuweid:
> Is this for Kubernetes use case or general case?
> 
> FYI, Kubernetes recently [added](https://github.com/kubernetes/apiserver/commit/f4edb394b0e83de067176cccc2fbefea217632a8) `maxLimit = 10000`

Thanks for the link! Be default the object size is limited in 1.5MiB. The pageSize can be used to reduce memory spike.
In kubernetes, it seems it can resolve this issue.

linxiulei:
> Is this for Kubernetes use case or general case?
> 

I am thinking of Kubernetes use case, but it definitely benefits general cases as well

> FYI, Kubernetes recently [added](https://github.com/kubernetes/apiserver/commit/f4edb394b0e83de067176cccc2fbefea217632a8) `maxLimit = 10000`

It seems somehow to mitigate the similar problem. AIUI, the intention of it is actually increase the response size to reduce the calls to etcd.

linxiulei:
@ahrtr can you please also try re-opening this?

serathius:
Note, not sure if large objects are as much of a concern when compared to large amounts of small objects. Response size can be controlled on user side as long as they limit maximal size of the object and number of objects per response. 

Context https://github.com/kubernetes/enhancements/issues/2340#issuecomment-1640185697

Small objects are much more disruptful for clients. See the difference in CPU usage between "Disabled Large Objects" and "Disabled small objects" for apiserver. Difference is between 3 CPUs and 72 CPUs.

Overall I still think that long term better solution is to introduce watch cache to etcd client to get the same awesome benefits (2-10 times CPU reduction, 20-50 times latency reduction) instead of creating short term workaround.

linxiulei:
[...]

PR #14795, opened 2022-11-17
feat: raft/log_unstable_test use testify packages in tests

falser101:
Signed-off-by: jianfei.zhang <jianfei.zhang@daocloud.io>
Fixed #14709

PR #14810, opened 2022-11-19
Support specifying `MaxBytes` in range request

linxiulei:
fixes: #14809


server/storage/mvcc/kvstore_txn.go
```diff
@@ -120,9 +126,20 @@ func (tr *storeTxnRead) rangeKeys(ctx context.Context, key, end []byte, curRev i
 				zap.Error(err),
 			)
 		}
+		totalBytes += int64(len(vs[0]))
+		count += 1
+		if ro.MaxBytes > 0 && totalBytes >= ro.MaxBytes {
+			break
+		}
```
ahrtr, line 133: Do this check before `Unmarshal` (line 123)

server/storage/mvcc/kvstore_txn.go
```diff
@@ -85,13 +85,19 @@ func (tr *storeTxnRead) rangeKeys(ctx context.Context, key, end []byte, curRev i
 		return &RangeResult{KVs: nil, Count: total, Rev: curRev}, nil
 	}
 
-	limit := int(ro.Limit)
+	limit, more := int(ro.Limit), false
 	if limit <= 0 || limit > len(revpairs) {
```
ahrtr, line 89: Note you removed the `limit = limit + 1` in `txn.go`, so we should include the case `limit == len(revpairs)` as  well
```suggestion
	if limit <= 0 || limit >= len(revpairs) {
```

server/storage/mvcc/kvstore_txn.go
```diff
@@ -120,9 +126,20 @@ func (tr *storeTxnRead) rangeKeys(ctx context.Context, key, end []byte, curRev i
...
 		}
+		totalBytes += int64(len(vs[0]))
+		count += 1
+		if ro.MaxBytes > 0 && totalBytes >= ro.MaxBytes {
+			break
+		}
+	}
+	if count != len(kvs) {
+		newkvs := make([]mvccpb.KeyValue, count)
+		copy(newkvs, kvs)
```
ahrtr, line 137: Why do yo copy the kvs?

ahrtr:
The new feature makes sense to me,  only limiting the max number of entries isn't enough if each entry is too big.

When client specifies both limit and maxBytes, then server side truncates the result when whichever is satisfied first. 

It should have no impact on Kubernetes, because the maxByte defaults to 0, which means no limit.

ahrtr:
Please also add some [common](https://github.com/etcd-io/etcd/tree/main/tests/common) e2e/integration test. Since this PR is already too big, so I'm OK to add it in a separate PR.

ahrtr:
When this PR gets merged, probably someone could enhance K8s side to support `maxBytes` as well. cc @dims @liggitt @deads2k @lavalamp

FYI. https://github.com/etcd-io/etcd/pull/14810#issuecomment-1320984608

server/storage/mvcc/kvstore_txn.go
linxiulei: We need to shrink the slice to the actual length. The RangeResult returned will be used in caller. For example, the length will be used to determine if there is 'More'

server/storage/mvcc/kvstore_txn.go
linxiulei: Done

server/storage/mvcc/kv_test.go
```diff
@@ -252,6 +253,63 @@ func testKVRangeLimit(t *testing.T, f rangeFunc) {
...
+		}
+		if r.Rev != wrev {
+			t.Errorf("#%d: rev = %d, want %d", i, r.Rev, wrev)
+		}
+		if r.Count != len(kvs) {
+			t.Errorf("#%d: count = %d, want %d", i, r.Count, len(kvs))
+		}
+		if r.More != tt.wmore {
+			t.Errorf("#%d: more = %v, want %v", i, r.More, tt.wmore)
+		}
```
ahrtr, line 312: Please use golang subtest, there are lots of examples in the repo.  FYI. https://github.com/etcd-io/etcd/pull/14795

server/storage/mvcc/kvstore_txn.go
ahrtr: No need to copy the slice, you just need to generate a new slice based something like below,
```
kvs = kvs[:count]
```

linxiulei: Done. Thanks!

server/storage/mvcc/kv_test.go
linxiulei: Done. Thanks!

ahrtr (approved):
LGTM

Thank yo @linxiulei 

cc @ptabor @serathius @spzala

ahrtr:
@linxiulei Please also add a changelog in https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md, of course, can be in a separate PR. 

TODO list:
1. add e2e/integration test. Please see https://github.com/etcd-io/etcd/pull/14810#issuecomment-1320985021;
2. add changelog item.

linxiulei:
Will do. Thanks @ahrtr !

server/etcdserver/txn/txn.go
```diff
@@ -200,10 +198,12 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, txnRead mvcc.TxnRead
 		}
 	}
 
+	resp.More = rr.More
```
serathius, line 201: What's the reason to add this line?

server/etcdserver/txn/txn.go
```diff
@@ -200,10 +198,12 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, txnRead mvcc.TxnRead
...
 	if r.Limit > 0 && len(rr.KVs) > int(r.Limit) {
 		rr.KVs = rr.KVs[:r.Limit]
 		resp.More = true
 	}
```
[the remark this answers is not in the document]
linxiulei: I intentionally didn't implement this for maxBytes. It seems not a big change. Let me see how to it.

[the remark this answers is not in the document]
ahrtr: Actually I think we can even the following code, because `limit = limit + 1` was removed (line 136), and `len(rr.KVs) > int(r.Limit)` will never be true. But it isn't a big deal, so I did not raise the comment. 

```
if r.Limit > 0 && len(rr.KVs) > int(r.Limit) {
		rr.KVs = rr.KVs[:r.Limit]
		resp.More = true
	}
```

server/storage/mvcc/kvstore_txn.go
```diff
@@ -114,15 +118,23 @@ func (tr *storeTxnRead) rangeKeys(ctx context.Context, key, end []byte, curRev i
...
+		if ro.MaxBytes > 0 && totalBytes > ro.MaxBytes {
+			break
+		}
 		if err := kvs[i].Unmarshal(vs[0]); err != nil {
 			tr.s.lg.Fatal(
 				"failed to unmarshal mvccpb.KeyValue",
 				zap.Error(err),
 			)
 		}
+		count += 1
```
serathius, line 131: `count == i` Please remove redundancy

server/storage/mvcc/kvstore_txn.go
```diff
@@ -114,15 +118,23 @@ func (tr *storeTxnRead) rangeKeys(ctx context.Context, key, end []byte, curRev i
...
+		if ro.MaxBytes > 0 && totalBytes == ro.MaxBytes {
```
serathius, line 132: This is premature optimization. I expect that cost of calculating this on every loop is larger than saving we would get. 

Don't know what values of maxBytes you expect users to select, but let's say user sets 1MB. What's the probability that byte sizes align exactly to 1 MB?

server/etcdserver/txn/txn.go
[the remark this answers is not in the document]
serathius: Don't agree, please read 
https://github.com/etcd-io/etcd/blob/ddd65e7fe2e2d6fbb9a89a07ddbef79412885659/server/etcdserver/txn/txn.go#L127-L133

Condition `len(rr.KVs) > int(r.Limit)` can definitely be broken when sorting or filtering is enabled.

server/storage/mvcc/kvstore_txn.go
linxiulei: Not sure I understand this. We still need `count` outside of the loop to resize the slice at the end.

server/storage/mvcc/kvstore_txn.go
linxiulei: Do you want to combine this to the condition at line 122 (ie if `ro.MaxBytes > 0 && totalBytes >= ro.MaxBytes`)

linxiulei: This is not just optimization but also the correctness. We still need this KV pair if the total size equals to `maxBytes`

server/etcdserver/txn/txn.go
```diff
@@ -200,10 +198,30 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, txnRead mvcc.TxnRead
...
+				break
+			}
+			if totalBytes == r.MaxBytes {
+				rr.KVs = rr.KVs[:i+1]
+				break
+			}
+		}
+		if l > len(rr.KVs) {
+			resp.More = true
+		}
```
serathius, line 222:
```suggestion
		var totalBytes int64
		for i, kv := range rr.KVs {
			size := int64(kv.Size())
			if totalBytes + size > r.MaxBytes {
			    resp.More = true
				break
			}
            totalBytes += size
		}
```

server/storage/mvcc/kvstore_txn.go
serathius: Please see comment above about how similar for loop can be made simpler.

server/storage/mvcc/kvstore_txn.go
serathius: Please see comment above about how similar for loop can be made simpler.

serathius:
I would like to see more tests before we merge this PR. This code is on a pretty important code path and there were already multiple edge cases not handled properly in this PR (for example maxBytes not working when sorting is enabled).

server/etcdserver/txn/txn.go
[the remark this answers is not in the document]
linxiulei: Done

server/storage/mvcc/kvstore_txn.go
linxiulei: Done

serathius (dismissed):

server/storage/mvcc/kvstore_txn.go
```diff
@@ -122,7 +128,7 @@ func (tr *storeTxnRead) rangeKeys(ctx context.Context, key, end []byte, curRev i
 		}
 	}
 	tr.trace.Step("range keys from bolt db")
-	return &RangeResult{KVs: kvs, Count: total, Rev: curRev}, nil
+	return &RangeResult{KVs: kvs, Count: total, Rev: curRev, More: total > len(kvs)}, nil
```
fuweid, line 131: I think it should be `More: limit > len(kvs)`?

fuweid (approved):
LGTM(non-binding)

It seems that it reduces that grpc unary API memory usage. The grpc http2 transport only releases the memory after last byte has been sent. I was thinking that if the Range can be streaming type, the List-All from kube-apiserver won't cause memory spike in ETCD side because the pending bytes in memory is limit by http2 stream/connection windows size.  However, it might be hard to integrate.

This patch can chunk the data. Thanks! @linxiulei Help a lot.

serathius:
Would also want to get some feedback from K8s scalability people who have some experience with issue that this feature tries to fix. cc @mborsz

mborsz:
I think it would be worth discussing this with authors/reviewers of https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/3157-watch-list which currently tries to solve similar (the same?) problem.


Probably it makes sense to get buy-in from sig-apimachinery that we will be able to use this in kube-apiserver before we make changes to etcd.

linxiulei:
I agree to get more buy-ins from k8s side to make this feature more appealing. But I do not think it should be a blocker for etcd as this feature can benefit other use cases other than k8s. Also getting buy-ins from k8s side involves more discussions and takes time. 

Nonetheless, I will try to initiate some discussion within k8s.

server/etcdserver/txn/txn.go
```diff
@@ -200,10 +198,23 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, txnRead mvcc.TxnRead
...
 		rr.KVs = rr.KVs[:r.Limit]
 		resp.More = true
 	}
+	if r.MaxBytes > 0 && maxBytes == 0 {
+		var totalBytes int64
+		for i, kv := range rr.KVs {
+			totalBytes += int64(kv.Size())
+			if totalBytes > r.MaxBytes {
+				resp.More = true
+				rr.KVs = rr.KVs[:i]
```
ptabor, line 212: If I understand the code correctly, it means we might be sending '0' items, it is, we might stuck completely, 
giving empty responses and retries. 

I think we should prioritize giving at least 1 item in such situation. And be very explicit that in this situation the maxBytes contract can be exceeded.
But this is the exact semantic discussion we should have with k8s api machinery team @mborsz @lavalamp

We also need to have tests for this.

linxiulei: Yes, you understood correctly. The initial version was to return the result with size just exceeding the `maxBytes` so at least one item is returned. But this made the `maxBytes` a bit ambiguous. 

Given the max object size is 1.5MB, we can warm users the maxBytes should be bigger than 1.5MB to avoid returning 0 items. WDYT?

lavalamp:
> When this PR gets merged, probably someone could enhance K8s side to support maxBytes as well

k8s would not directly expose this to end users, to be clear. (but thanks for the heads up!)

linxiulei:
> Please also add some [common](https://github.com/etcd-io/etcd/tree/main/tests/common) e2e/integration test. Since this PR is already too big, so I'm OK to add it in a separate PR.

I added a commit of e2e tests in the previous push

linxiulei:
How can I re-open it? I am still active on this PR if maintainers are able to evaluate it

ahrtr:
I am not able to reopen either. Please raise a new PR, thx

fuweid:
@linxiulei maybe it will be active after you repush?

fuweid:
reopen...

linxiulei:
repush didn't work either. Let me raise a new PR.

PR #14818, opened 2022-11-21
client: add log to notify default TTL value usage

### LINKED ISSUE #14811: error/info log when TTL is <= 0

This is an enhancement rather than a feature.

Currently, if I try to create a session as such: `concurrency.NewSession(etcdClient, concurrency.WithTTL(0))` it automatically defaults to 60 without any notice.

It would be nice to have a log getting printed. Example: `"TTL should be > 0, defaulting to 60"`

### Why is this needed?

To improve usability.

shreemaan-abhishek:
Closes #14811 

Signed-off-by: Abhishek Choudhary <shreemaan.abhishek@gmail.com>

### DIFF
--- a/client/v3/concurrency/session.go
+++ b/client/v3/concurrency/session.go
@@ -24,8 +24,6 @@ import (
 
 const defaultSessionTTL = 60
 
-var logger *zap.Logger
-
 // Session represents a lease kept alive for the lifetime of a client.
 // Fault-tolerant applications may use sessions to reason about liveness.
 type Session struct {
@@ -39,10 +37,10 @@ type Session struct {
 
 // NewSession gets the leased session for a client.
 func NewSession(client *v3.Client, opts ...SessionOption) (*Session, error) {
-	logger = client.GetLogger()
+	logger := client.GetLogger()
 	ops := &sessionOptions{ttl: defaultSessionTTL, ctx: client.Ctx()}
 	for _, opt := range opts {
-		opt(ops)
+		opt(ops, logger)
 	}
 
 	id := ops.leaseID
@@ -112,12 +110,12 @@ type sessionOptions struct {
 }
 
 // SessionOption configures Session.
-type SessionOption func(*sessionOptions)
+type SessionOption func(*sessionOptions, *zap.Logger)
 
 // WithTTL configures the session's TTL in seconds.
 // If TTL is <= 0, the default 60 seconds TTL will be used.
 func WithTTL(ttl int) SessionOption {
-	return func(so *sessionOptions) {
+	return func(so *sessionOptions, logger *zap.Logger) {
 		if ttl > 0 {
 			so.ttl = ttl
 		} else {
@@ -130,7 +128,7 @@ func WithTTL(ttl int) SessionOption {
 // This is useful in process restart scenario, for example, to reclaim
 // leadership from an election prior to restart.
 func WithLease(leaseID v3.LeaseID) SessionOption {
-	return func(so *sessionOptions) {
+	return func(so *sessionOptions, logger *zap.Logger) {
 		so.leaseID = leaseID
 	}
 }
@@ -141,7 +139,7 @@ func WithLease(leaseID v3.LeaseID) SessionOption {
 // context is canceled before Close() completes, the session's lease will be
 // abandoned and left to expire instead of being revoked.
 func WithContext(ctx context.Context) SessionOption {
-	return func(so *sessionOptions) {
+	return func(so *sessionOptions, logger *zap.Logger) {
 		so.ctx = ctx
 	}
 }
--- a/client/v3/concurrency/session.go
+++ b/client/v3/concurrency/session.go
@@ -19,9 +19,12 @@ import (
 	"time"
 
 	v3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
 )
 
 const defaultSessionTTL = 60
+
+var logger *zap.Logger
 
 // Session represents a lease kept alive for the lifetime of a client.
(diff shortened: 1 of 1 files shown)


shreemaan-abhishek:
Sorry for the frequent force pushes.

I was trying to amend the existing commit and mistakenly amended it without staging the changes. 🤦🏻

client/v3/concurrency/session.go
```diff
@@ -19,10 +19,13 @@ import (
 	"time"
 
 	v3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
 )
 
 const defaultSessionTTL = 60
 
+var logger *zap.Logger
```
serathius, line 27: Please no global logger

client/v3/concurrency/session.go
```diff
@@ -116,6 +120,8 @@ func WithTTL(ttl int) SessionOption {
 	return func(so *sessionOptions) {
 		if ttl > 0 {
 			so.ttl = ttl
+		} else {
+			logger.Info("TTL should be > 0, defaulting to 60")
```
serathius, line 124: Please move this check to `NewSesson` function after options are applied.

lavacat: @serathius alternative is to add logger as SessionOption param
`type SessionOption func(*sessionOptions, *zap.Logger)`
and then `opt(ops, client.GetLogger())`

lavacat, line 124: Should log `defaultSessionTTL` constant not `60`. For example `logger.Info("TTL should be > 0, using default", zap.Int64("default-session-ttl", defaultSessionTTL))`

client/v3/concurrency/session.go
ahrtr: +1

client/v3/concurrency/session.go
shreemaan-abhishek: Great catch! Thank you!!

shreemaan-abhishek: This sounds more intuitive lemme give this a try.

client/v3/concurrency/session.go
```diff
@@ -135,7 +139,7 @@ func WithLease(leaseID v3.LeaseID) SessionOption {
 // context is canceled before Close() completes, the session's lease will be
 // abandoned and left to expire instead of being revoked.
 func WithContext(ctx context.Context) SessionOption {
-	return func(so *sessionOptions) {
+	return func(so *sessionOptions, logger *zap.Logger) {
```
shreemaan-abhishek, line 142: Should I replace `logger` with `_` since this parameter is unused?

client/v3/concurrency/session.go
```diff
@@ -124,7 +128,7 @@ func WithTTL(ttl int) SessionOption {
 // This is useful in process restart scenario, for example, to reclaim
 // leadership from an election prior to restart.
 func WithLease(leaseID v3.LeaseID) SessionOption {
-	return func(so *sessionOptions) {
+	return func(so *sessionOptions, logger *zap.Logger) {
```
shreemaan-abhishek, line 131: Should I replace `logger` with `_` since this parameter is unused?

client/v3/concurrency/session.go
lavacat: yes

lavacat:
Please squash your 3 commits into 1.

client/v3/concurrency/session.go
lavacat: yes

client/v3/concurrency/session.go
shreemaan-abhishek: @serathius @ahrtr, if you have the time and patience. Would you mind explaining the reason behind this? I am new to coding in golang. Or please drop some guides/blogs for my self-study.

Thanks.

shreemaan-abhishek:
Lemme know if this looks good. Then I will squash these commits. Thanks :)

lavacat:
> Lemme know if this looks good. Then I will squash these commits. Thanks :)

Also add unit test. I think you'll need to create session_test.go, since there are no unit tests for session.go

client/v3/concurrency/session.go
```diff
@@ -108,14 +110,16 @@ type sessionOptions struct {
...
 
 // WithTTL configures the session's TTL in seconds.
 // If TTL is <= 0, the default 60 seconds TTL will be used.
 func WithTTL(ttl int) SessionOption {
-	return func(so *sessionOptions) {
+	return func(so *sessionOptions, lg *zap.Logger) {
 		if ttl > 0 {
 			so.ttl = ttl
+		} else {
+			lg.Info("TTL should be > 0, using default", zap.Int64("default-session-ttl", defaultSessionTTL))
```
ptabor, line 122: I would even make it "Warn", as this option to reset to default does not have much sense.

ptabor, line 122: 'using default' might be misleading. 

See what happens in case of akward setup: 

 `..., WithTTL(30), WithTTL(0),`
 
 We would land with 30, but the message would say: `using default`. 

How about: 
  lg.Info("WithTTL(): TTL should be > 0, preserving current ttl", zap.Int64("current-ttl", so.ttl)) ?

ptabor (changes requested):
Thank you for the fixing the usablity traps.
+1 to @lavacat requests to add a test.

lavacat:
@shreemaan-abhishek are you still working on this? If you need more time, that's fine, just post here. If I don't hear from you  in next couple weeks, I'll take over this PR

shreemaan-abhishek:
#14993 already fixed this issue. Closing this.

Issue #14726, opened 2022-11-10
Implement remaining gofailpoints in linearizability tests


serathius:
### What would you like to be added?

This is followup to https://github.com/etcd-io/etcd/pull/14687 that introduced triggering random failpoint.
Failpoints are injected by sending http request to modified etcd server that will panic on when it reaches specific part of code. 
Problem is that some failpoints will not be naturally triggered and require additional action to server to reach the code. For example  `defragBeforeCopy` and `defragBeforeRename` require starting a Defrag. 

There are still failpoints that require implementing their triggers. Goal of this issue is to implement triggers for failpoints and adding them to `RandomFailpoint` which is used by automated testing. Code: https://github.com/etcd-io/etcd/blob/main/tests/linearizability/failpoints.go#L33-L53

Failpoints to be implemented:
* [x] `etcdserver/raftBeforeLeaderSend` (@serathius, https://github.com/etcd-io/etcd/pull/14752)
* [x] `etcdserver/raftBeforeApplySnap`
* [x] `etcdserver/raftAfterApplySnap`
* [x] `etcdserver/raftAfterWALRelease`
* [x] `etcdserver/raftBeforeFollowerSend`
* [x] `etcdserver/raftBeforeSaveSnap`
* [x] `etcdserver/raftAfterSaveSnap`

### Why is this needed?

Increases the coverage of linearizability tests.

lavacat:
@serathius please assign this to me. I'm already working on all `Snap` endpoints. The way to reproduce is outlined here https://github.com/etcd-io/etcd/issues/10219.
I got it working within current framework using trigger hook, but sometimes it's flaky. Will submit draft pr after it's passing consistently.

jmhbnz:
Hey @serathius, @lavacat - Can this one be closed now?

I'm looking at the list of tasks on the issue and to me it looks like #15104 and #15268 completed all these tasks?

serathius:
Yep, looks like we are done!

Issue #10219, opened 2018-10-26
Corrupted WAL and snapshot restoring process


brk0v:
## Issue 

Node that was offline more than `max(SnapshotCount, DefaultSnapshotCatchUpEntries)` corrupts its WAL log with bad  `HardState.Commit`number if it's killed right after `HardState` was saved to non-volatile storage (`failpoint: raftBeforeSaveSnap`). 

## Specific
Version: master
Environment: any (tested on Linux, MacOS X)

## Steps to reproduce

1. Patch snapshots defaults to trigger the issues easier:

```go
+++ b/etcdserver/server.go
@@ -65,14 +65,14 @@ import (
 )

 const (
-       DefaultSnapshotCount = 100000
+       DefaultSnapshotCount = 10

        // DefaultSnapshotCatchUpEntries is the number of entries for a slow follower
        // to catch-up after compacting the raft storage entries.
        // We expect the follower has a millisecond level latency with the leader.
        // The max throughput is around 10K. Keep a 5K entries is enough for helping
        // follower to catch up.
-       DefaultSnapshotCatchUpEntries uint64 = 5000
+       DefaultSnapshotCatchUpEntries uint64 = 10
```

2. Compile with failpoints to emulate power failure during the process of snapshot restoration:

``` 
FAILPOINTS="true" make build
```

3. Create a `Procfile` with a failpoint `raftBeforeSaveSnap` for etcd2 node :

```
etcd1: bin/etcd --name infra1 --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 --listen-peer-urls http://127.0.0.1:12380 --initial-advertise-peer-urls http://127.0.0.1:12380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new --enable-pprof --logger=zap --log-outputs=stderr --debug 
[...]

PR #14819, opened 2022-11-22
tests linearizability: reproduce and prevent 14571

chaochn47:
Related to https://github.com/etcd-io/etcd/issues/14045

Signed-off-by: Chao Chen <chaochn@amazon.com>

### DIFF
--- a/tests/linearizability/failpoints.go
+++ b/tests/linearizability/failpoints.go
@@ -22,7 +22,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
 
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
@@ -33,7 +36,8 @@ const (
 )
 
 var (
-	KillFailpoint                            Failpoint = killFailpoint{}
+	KillFailpoint                            Failpoint = killFailpoint{target: AnyMember}
+	EnableAuthKillFailpoint                  Failpoint = killFailpoint{enableAuth: true, target: Follower}
 	DefragBeforeCopyPanic                    Failpoint = goPanicFailpoint{"defragBeforeCopy", triggerDefrag, AnyMember}
 	DefragBeforeRenamePanic                  Failpoint = goPanicFailpoint{"defragBeforeRename", triggerDefrag, AnyMember}
 	BeforeCommitPanic                        Failpoint = goPanicFailpoint{"beforeCommit", nil, AnyMember}
@@ -78,15 +82,22 @@ var (
 )
 
 type Failpoint interface {
-	Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error
+	Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error
 	Name() string
 	Available(e2e.EtcdProcess) bool
 }
 
-type killFailpoint struct{}
+type killFailpoint struct {
+	enableAuth bool
+	target     failpointTarget
+}
 
-func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
-	member := clus.Procs[rand.Int()%len(clus.Procs)]
+func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error {
+	opt := func(any) {}
+	if f.enableAuth {
+		opt = e2e.WithAuth(rootUserName, rootUserPassword)
+	}
+	member := pickMember(f.target, t, clus, opt)
 
 	killCtx, cancel := context.WithTimeout(ctx, triggerTimeout)
 	defer cancel()
@@ -101,11 +112,19 @@ func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Etcd
 		}
 	}
 
-	err := member.Start(ctx)
-	if err != nil {
-		return err
+	// get endpoints excluding the killed member client URLs
+	endpoints := make([]string, 0, len(clus.EndpointsV3()))
+	for _, ed := range clus.EndpointsV3() {
+		if ed != member.EndpointsV3()[0] {
+			endpoints = append(endpoints, ed)
+		}
 	}
-	return nil
+
+	if f.enableAuth {
+		require.NoError(t, addTestUserAuth(ctx, endpoints))
+	}
+
+	return member.Start(ctx)
 }
 
 func (f killFailpoint) Name() string {
@@ -114,6 +133,38 @@ func (f killFailpoint) Name() string {
 
 func (f killFailpoint) Available(e2e.EtcdProcess) bool {
 	return true
+}
+
+func addTestUserAuth(ctx context.Context, endpoints []string) (err error) {
+	cc, err := clientv3.New(clientv3.Config{
+		Endpoints:            endpoints,
+		Logger:               zap.NewNop(),
+		DialKeepAliveTime:    1 * time.Millisecond,
+		DialKeepAliveTimeout: 5 * time.Millisecond,
+		Username:             rootUserName,
+		Password:             rootUserPassword,
+	})
+	defer func() {
+		if cc != nil {
+			cc.Close()
+		}
+	}()
+	if err != nil {
+		return err
+	}
+	if _, err := cc.UserAdd(ctx, testUserName, testUserPassword); err != nil {
+		return err
+	}
+	if _, err := cc.RoleAdd(ctx, testRoleName); err != nil {
+		return err
+	}
+	if _, err := cc.UserGrantRole(ctx, testUserName, testRoleName); err != nil {
+		return err
+	}
+	if _, err := cc.RoleGrantPermission(ctx, testRoleName, startKey, endKey, clientv3.PermissionType(clientv3.PermReadWrite)); err != nil {
+		return err
+	}
+	return nil
 }
 
 type goPanicFailpoint struct {
@@ -127,10 +178,11 @@ type failpointTarget string
 const (
 	AnyMember failpointTarget = "AnyMember"
 	Leader    failpointTarget = "Leader"
+	Follower  failpointTarget = "Follower"
 )
 
-func (f goPanicFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
-	member := f.pickMember(t, clus)
+func (f goPanicFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error {
+	member := pickMember(f.target, t, clus)
 
 	triggerCtx, cancel := context.WithTimeout(ctx, triggerTimeout)
 	defer cancel()
@@ -159,12 +211,17 @@ func (f goPanicFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.E
 	return nil
 }
 
-func (f goPanicFailpoint) pickMember(t *testing.T, clus *e2e.EtcdProcessCluster) e2e.EtcdProcess {
-	switch f.target {
+func pickMember(target failpointTarget, t *testing.T, clus *e2e.EtcdProcessCluster, opts ...config.ClientOption) e2e.EtcdProcess {
+	switch target {
 	case AnyMember:
 		return clus.Procs[rand.Int()%len(clus.Procs)]
 	case Leader:
-		return clus.Procs[clus.WaitLeader(t)]
+		return clus.Procs[clus.WaitLeader(t, opts...)]
+	case Follower:
+		if len(clus.Procs) == 1 {
+			panic("single node cluster does not have follower")
+		}
+		return clus.Procs[(clus.WaitLeader(t, opts...)+1)%len(clus.Procs)]
 	default:
 		panic("unknown target")
 	}
@@ -228,7 +285,7 @@ type randomFailpoint struct {
 	failpoints []Failpoint
 }
 
-func (f randomFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
+func (f randomFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error {
 	availableFailpoints := make([]Failpoint, 0, len(f.failpoints))
 	for _, failpoint := range f.failpoints {
 		count := 0
@@ -243,7 +300,7 @@ func (f randomFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Et
 	}
 	failpoint := availableFailpoints[rand.Int()%len(availableFailpoints)]
 	t.Logf("Triggering %v failpoint\n", failpoint.Name())
-	return failpoint.Trigger(t, ctx, clus)
+	return failpoint.Trigger(t, ctx, clus, lg)
 }
 
 func (f randomFailpoint) Name() string {
@@ -258,7 +315,7 @@ type blackholePeerNetworkFailpoint struct {
 	duration time.Duration
 }
 
-func (f blackholePeerNetworkFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
+func (f blackholePeerNetworkFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error {
 	member := clus.Procs[rand.Int()%len(clus.Procs)]
 	proxy := member.PeerProxy()
 
@@ -286,7 +343,7 @@ type delayPeerNetworkFailpoint struct {
 	randomizedLatency time.Duration
 }
 
-func (f delayPeerNetworkFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
+func (f delayPeerNetworkFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error {
 	member := clus.Procs[rand.Int()%len(clus.Procs)]
 	proxy := member.PeerProxy()
 
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -28,10 +28,14 @@ import (
 	"github.com/anishathalye/porcupine"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/time/rate"
 
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/integration"
 	"go.etcd.io/etcd/tests/v3/linearizability/identity"
 	"go.etcd.io/etcd/tests/v3/linearizability/model"
 )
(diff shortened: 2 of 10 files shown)


ptabor (approved):

ptabor (changes requested):
(I'm sorry - prematurly clicked approve)

tests/linearizability/traffic.go
```diff
@@ -105,3 +109,37 @@ func (t readWriteSingleKey) pickWriteOperation() Operation {
 	}
 	panic("unexpected")
 }
+
+func (t readWriteSingleKey) AuthEnabled() bool { return false }
+
+type authTraffic struct {
```
serathius, line 115: Can auth be just a bool flag on `readWriteSingleKey`?

tests/linearizability/linearizability_test.go
```diff
@@ -43,77 +45,112 @@ const (
...
 }
 
-func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProcessClusterConfig, failpoint FailpointConfig, traffic trafficConfig) {
+func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProcessClusterConfig, failpoint FailpointConfig, traffic trafficConfig, lg *zap.Logger) {
 	clus, err := e2e.NewEtcdProcessCluster(ctx, t, e2e.WithConfig(&config))
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer clus.Close()
+	if traffic.traffic.AuthEnabled() {
```
serathius, line 131: Setting up authority for traffic should be done by traffic.

tests/linearizability/linearizability_test.go
```diff
@@ -43,77 +45,112 @@ const (
...
+			name: "Issue14571",
+			failpoint: FailpointConfig{
+				failpoint: MemberAddFailpoint,
+				count:     1,
+			},
+			config: e2e.EtcdProcessClusterConfig{
+				ClusterSize:   1,
+				SnapshotCount: 2,
+			},
+			clientCount: 3,
```
serathius, line 101: Why this requires 3 clients?

tests/linearizability/auth.go
```diff
@@ -0,0 +1,60 @@
...
+const (
+	rootUserName     = "root"
+	rootRoleName     = "root"
+	rootUserPassword = "123"
+	testUserName     = "test-user"
+	testRoleName     = "test-role"
+	testUserPassword = "abc"
+)
+
+func setupAuth(ctx context.Context, t *testing.T, c interfaces.Client) {
```
serathius, line 35: Please move this to traffic

tests/linearizability/traffic.go
```diff
@@ -18,17 +18,21 @@ import (
...
 
 var (
 	DefaultTraffic Traffic = readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 100}}}
+	AuthTraffic    Traffic = authTraffic{readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 100}}}}
 )
 
 type Traffic interface {
-	Run(ctx context.Context, c *recordingClient, limiter *rate.Limiter)
+	Run(ctx context.Context, c *recordingClient, limiter *rate.Limiter, lg *zap.Logger)
+	AuthEnabled() bool
```
serathius, line 35: I think traffic can handle it's own setup. Maybe instead of exposing `AuthEnabled() bool` we have `Setup(ctx context.Context, c interfaces.Client, lg *zap.Logger)`

tests/linearizability/failpoints.go
```diff
@@ -228,3 +230,23 @@ func (f randomFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Et
...
+
+func (f memberAddFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
+	// write more than SnapshotCount keys to single leader to make sure snapshot is created
+	testUserClientOpts := e2e.WithAuth("test-user", "abc")
+	for i := 0; i <= 10; i++ {
+		if err := clus.Client(testUserClientOpts).Put(ctx, "key", "test", config.PutOptions{}); err != nil {
+			panic(err)
+		}
+	}
+	if err := clus.StartNewProc(ctx, nil, t, e2e.WithAuth("root", "123")); err != nil {
```
serathius, line 244: Not sure how running a new member is a trigger. Could we instead setup 3 member cluster, terminate one of the members (could be gradefully) and wait until there are enough transaction to trigger snapshot before starting it again?

tests/linearizability/linearizability_test.go
chaochn47: The initial authority is set up once ideally. 

If moving to `simulateTraffic`, it will be ran by multiple clients.

tests/linearizability/linearizability_test.go
chaochn47: Only 3 clients are needed to reproduce on v3.5.5.

tests/linearizability/auth.go
chaochn47: Looks like the same comment about moving the auth set up into traffic.

tests/linearizability/failpoints.go
chaochn47: Yeah, I can try this out.

tests/linearizability/traffic.go
chaochn47: Okay. I can add one auth related configuration to handle the set up.

chaochn47:
@serathius

> Not sure how running a new member is a trigger. Could we instead setup 3 member cluster, terminate one of the members (could be gradefully) and wait until there are enough transaction to trigger snapshot before starting it again?

I am not able to reproduce the issue with this set up. 

Would you mind shed some lights on if I set up the framework incorrectly? Here is the diff in my fork. https://github.com/chaochn47/etcd/pull/7

image

chaochn47:
I think I get it reproduced with `100` compared to original `5000` `DefaultSnapshotCatchUpEntries`. With the original value, porcupine checking all the operations takes too long and test is killed after 10mins. 

Will try to not to include those recorded operations before the "real" failure is triggered. 

Does it make sense before I proceed with this approach?

tests/linearizability/failpoints.go
```diff
@@ -99,13 +103,92 @@ func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Etcd
...
+	initialRevision, err := getRevision(ctx, cc)
 	if err != nil {
 		return err
 	}
+	lg.Info("got revision",
+		zap.Int64("initial-revision", initialRevision),
+		zap.Int64("snapshot-catchup-count", snapshotCount),
+	)
+	for {
+		time.Sleep(time.Second)
```
lavacat, line 163: nit: this doesn't abide 1s triggerTimeout.

tests/linearizability/failpoints.go
```diff
@@ -99,13 +103,92 @@ func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Etcd
...
+			DialKeepAliveTimeout: 5 * time.Millisecond,
+			Username:             rootUserName,
+			Password:             rootUserPassword,
+		})
+		require.NoError(t, err)
+		require.NoError(t, addTestUserAuth(ctx, cc))
+		// 5000 is the default number of snapshot count catch up entries
+		// DefaultSnapshotCatchUpEntries = 5000
+		// Try manipulating e2e etcd test binary to 100 locally and snapshot-count here to speed up test
+		require.NoError(t, waitForSnapshot(ctx, cc, 5000, lg))
```
lavacat, line 130: are you trying to hit `s.authStore.Recover` etcdserver/server.go:1042? If yes, I think you need to bring back killed member, so it will actually apply snapshot.  Also you need to set snapshot-count to lower number. See https://github.com/etcd-io/etcd/issues/10219 for an example.

FYI, seams we are both working on reproducing `apply snapshot` path. My logic for https://github.com/etcd-io/etcd/issues/14726 looks similar, except I use goPanicFailpoint.

tests/linearizability/failpoints.go
```diff
@@ -77,13 +79,15 @@ var (
...
 }
 
-type killFailpoint struct{}
+type killFailpoint struct {
+	waitForSnapshot bool
+}
 
-func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
+func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster, lg *zap.Logger) error {
 	member := clus.Procs[rand.Int()%len(clus.Procs)]
```
lavacat, line 91: how about using `member := clus.Procs[(clus.WaitLeader(t)+1)%len(clus.Procs)]` to skip leader election and save time? Or do you want to test killing leader?

chaochn47: The kill failpoint should not only applicable to followers. 2s wait time is fine.

tests/linearizability/failpoints.go
```diff
@@ -99,13 +103,92 @@ func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Etcd
...
+			Logger:               zap.NewNop(),
+			DialKeepAliveTime:    1 * time.Millisecond,
+			DialKeepAliveTimeout: 5 * time.Millisecond,
+			Username:             rootUserName,
+			Password:             rootUserPassword,
+		})
+		require.NoError(t, err)
+		require.NoError(t, addTestUserAuth(ctx, cc))
+		// 5000 is the default number of snapshot count catch up entries
+		// DefaultSnapshotCatchUpEntries = 5000
```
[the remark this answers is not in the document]
chaochn47: Do you have a production use case? 

I am supportive of making these available to e2e test though.

tests/linearizability/failpoints.go
```diff
@@ -99,13 +103,92 @@ func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Etcd
...
+	endpoints := make([]string, 0, len(clus.EndpointsV3()))
+	for _, ed := range clus.EndpointsV3() {
+		if ed != member.EndpointsV3()[0] {
+			endpoints = append(endpoints, ed)
+		}
+	}
+
+	if f.waitForSnapshot {
+		// wait for potential leader election
+		time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
chaochn47: Thanks for the call out.

tests/linearizability/failpoints.go
chaochn47: Yeah, line 134 is restart the member.

tests/linearizability/failpoints.go
chaochn47: ack

chaochn47:
Looks like in the CI box particularlly, 

```
logger.go:130: 2022-12-20T01:44:47.062Z	INFO	Issue14571	linearizability/failpoints.go:169	got revision	{"initial-revision": 379, "current-revision": 379, "snapshot-catchup-count": 5000}
```

The write does not go though. Need to investigate more before publishing it to review.

tests/linearizability/failpoints.go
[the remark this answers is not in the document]
lavacat: I don't have prod case and the comment in config.go says
>// WARNING: only change this for tests. Always use "DefaultSnapshotCatchUpEntries"
On one hand, it's not a good idea to expose flag just because of testing, but doing 5k ops to trigger snapshot apply is also not great.

Another e2e test use case:
https://github.com/etcd-io/etcd/pull/15013#discussion_r1051683699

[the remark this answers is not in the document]
ahrtr: I agree to make `DefaultSnapshotCatchUpEntries` configuable via a user-facing flag, two reasons:
1. It can benefit edge case, which usually is one-node cluster, and with limited resource (e.g memory). Setting a small value for `DefaultSnapshotCatchUpEntries` can reduce the memory usage. 
2. It can benefit the E2E test case.

Please let me know if any of you can do this, otherwise let me take care of it.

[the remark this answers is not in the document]
chaochn47: Thanks @ahrtr I can take that as it looks like blocking my current PR. 

Because I just realize that linearizability test will repeat 60 times.. The current PR will add average 1min execution time per run. So It's not acceptable.  

https://github.com/etcd-io/etcd/blob/a59276c171d1647653b75679a12f85068d3b192b/.github/workflows/linearizability.yaml#L17

tests/linearizability/failpoints.go
chaochn47: Second thought, we can definitely optimize that with your suggestion after stabilizing this test case..

tests/linearizability/failpoints.go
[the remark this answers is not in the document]
chaochn47: ack. Will do on PST daytime and cc you and other stakeholders.

chaochn47:
Tested with v3.5.5 binary with experimental-snapshot-catchup-entry commit. `linearizability_test.go:277: Model is not linearizable`

ahrtr:
> Tested 60 times with v3.5.5 binary with experimental-snapshot-catchup-entry commit. `linearizability_test.go:277: Model is not linearizable` while on mainline, it is.

Thanks @chaochn47, please provide a detail steps. I may take a look sometime on weekend or next week.

chaochn47:
Hi @ahrtr, it is not a new issue. The test is added to reproduce and prevent #14571. It has been fixed in v3.5.6.

chaochn47:
The PR is ready for review.

@ptabor @serathius Could you please take a second look, thanks!

serathius:
Please  resolve conficts.

chaochn47:
```
TestLinearizability_ClusterOfSize3

linearizability_test.go:428: Linearization timed out
```

chaochn47:
Conflicts have been resolved. Would you mind take a second look? @serathius Thanks!!

tests/linearizability/failpoints.go
```diff
@@ -33,7 +36,8 @@ const (
 )
 
 var (
-	KillFailpoint                            Failpoint = killFailpoint{}
+	KillFailpoint                            Failpoint = killFailpoint{target: AnyMember}
+	EnableAuthKillFailpoint                  Failpoint = killFailpoint{enableAuth: true, target: Follower}
```
serathius, line 40: Do you need to target follower here? I understand that 1471 happens when follower is killed, however we don't need to hardcode it.

serathius, line 40: I don't understand why failpoint needs to be aware of auth. It's true that at this moment it creates the client, but maybe we can move client creation to different place. Either do dependency injection and provide failpoint with client that is already authorized. Alternative would be to add method to e2e.EtcdProcessCluster that provides the client.

tests/linearizability/failpoints.go
```diff
@@ -101,11 +112,19 @@ func (f killFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.Etcd
...
+	endpoints := make([]string, 0, len(clus.EndpointsV3()))
+	for _, ed := range clus.EndpointsV3() {
+		if ed != member.EndpointsV3()[0] {
+			endpoints = append(endpoints, ed)
+		}
 	}
-	return nil
+
+	if f.enableAuth {
+		require.NoError(t, addTestUserAuth(ctx, endpoints))
```
serathius, line 124: Don't like that auth setup is part of failpoint injection. Those are totally separate things. Please move auth setup to cluster setup.

tests/linearizability/linearizability_test.go
```diff
@@ -54,15 +58,22 @@ var (
...
 	tcs := []struct {
-		name      string
-		failpoint Failpoint
-		config    e2e.EtcdProcessClusterConfig
-		traffic   *trafficConfig
+		name        string
+		failpoint   Failpoint
+		config      e2e.EtcdProcessClusterConfig
+		traffic     *trafficConfig
+		clientCount int
```
serathius, line 76: This client count doesn't seem to be used. Please remove it and use `TrafficConfig.clientCount`

tests/linearizability/linearizability_test.go
```diff
@@ -124,39 +149,40 @@ func TestLinearizability(t *testing.T) {
 				t.Fatal(err)
 			}
 			defer clus.Close()
+			lg := zaptest.NewLogger(t, zaptest.WrapOptions(zap.AddCaller())).Named(tc.name)
```
serathius, line 152: This is nice, but please consider moving it to separate PR.

tests/linearizability/linearizability_test.go
```diff
@@ -277,21 +303,26 @@ type FailpointConfig struct {
...
 	lm := identity.NewLeaseIdStorage()
-	h := model.History{}
+	h := &model.History{}
 	limiter := rate.NewLimiter(rate.Limit(config.maximalQPS), 200)
 
 	startTime := time.Now()
 	wg := sync.WaitGroup{}
+
 	for i := 0; i < config.clientCount; i++ {
+		i := i
```
serathius, line 321: This should not be needed as we pass `i` as argument to goroutine funtion.

tests/linearizability/linearizability_test.go
```diff
@@ -301,23 +332,62 @@ func simulateTraffic(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessClu
...
 	qps := float64(len(operations)) / float64(endTime.Sub(startTime)) * float64(time.Second)
-	t.Logf("Average traffic: %f qps", qps)
+	lg.Info("Average traffic", zap.Float64("qps", qps))
 	if qps < config.minimalQPS {
 		t.Errorf("Requiring minimal %f qps for test results to be reliable, got %f qps", config.minimalQPS, qps)
 	}
 	return operations
 }
 
+func simulatePostFailpointTraffic(ctx context.Context, wg *sync.WaitGroup, endpoints []string, clientId int, ids identity.Provider, h *model.History, mux *sync.Mutex, config trafficConfig, limiter *rate.Limiter, lm identity.LeaseIdStorage) {
```
serathius, line 355: Don't understand why you cannot incorporate this into normal traffic

tests/linearizability/traffic.go
```diff
@@ -17,39 +17,57 @@ package linearizability
...
 type opChance struct {
 	operation model.OperationType
 	chance    int
 }
 
-func (t readWriteSingleKey) Run(ctx context.Context, clientId int, c *recordingClient, limiter *rate.Limiter, ids identity.Provider, lm identity.LeaseIdStorage) {
+func (t readWriteSingleKey) PreRun(ctx context.Context, c interfaces.Client, lg *zap.Logger) error {
+	if t.AuthEnabled() {
+		lg.Info("set up auth")
+		return setupAuth(ctx, c)
```
serathius, line 65: Authorization setup should be done at cluster setup, not at this point.

tests/linearizability/linearizability_test.go
chaochn47: To incorporate this into normal traffic, a client with test user authorization has to be set up upfront. However, before the user is added to the cluster, client creation will fail. 

After a couple of failed attempts, I adopted this workaround. It's more deterministic to reliably reproduce on the impacted 3.5 versions.

tests/linearizability/failpoints.go
chaochn47: Good suggestion! I can explore each option and see what's the best fit.

tests/linearizability/traffic.go
chaochn47: Make sense.

tests/linearizability/failpoints.go
serathius: I think that enabling auth is orthogonal to failure injection.

tests/linearizability/linearizability_test.go
serathius: Please setup authorization in cluster setup.

chaochn47:
Revisit this task again, I think the linearizability test does not have to reproduce the exact same scenario how #14571 happened. 

#14571 uncovers the issue that auth recover from snapshot failed to update `rangePermCache`. It's a in-memory map, for each user, maintains a interval tree that checks requested key start to key end is the subnet of the interval tree. 

The plan will be

At cluster setup stage: 
1. create root user, root role has access to all operations. 
2. create test user, test role, grant test role RW permissions from key `foo` to `zoo`
Traffic generator 
1. half of the clients using root user permissions
2. half of the clients using test user permissions
3. root user client will periodically grant / revoke test role permissions

Fault injector
1. kill a random member

Issue #14823, opened 2022-11-22
[tests/integration] flake test case TestCompactHashCheckDetectCorruption 


fuweid:
### Which github workflows are flaking?

Tests

### Which tests are flaking?

TestCompactHashCheckDetectCorruption

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/3356040544/jobs/5560786679

### Reason for failure (if possible)

If the corrupted member has been elected as leader, the `memberID` in alert response won't be the corrupted one.
It will be a smaller follower ID since the [raftCluster.Members][1] always sorts by ID.

```
 logger.go:130: 2022-10-30T15:35:17.268Z	INFO	m0.raft	2a4384d1c9ba4399 received MsgVoteResp from 27f6db18bed0afb9 at term 3	{"member": "m0"}
    logger.go:130: 2022-10-30T15:35:17.268Z	INFO	m0.raft	2a4384d1c9ba4399 has received 2 MsgVoteResp votes and 0 vote rejections	{"member": "m0"}
    logger.go:130: 2022-10-30T15:35:17.268Z	INFO	m0.raft	2a4384d1c9ba4399 became leader at term 3	{"member": "m0"}
    logger.go:130: 2022-10-30T15:35:17.268Z	INFO	m0.raft	raft.node: 2a4384d1c9ba4399 elected leader 2a4384d1c9ba4399 at term 3	{"member": "m0"}
 logger.go:130: 2022-10-30T15:35:17.269Z	INFO	m2.raft	27f6db18bed0afb9 no leader at term 3; dropping index reading msg	{"member": "m2"}
    logger.go:130: 2022-10-30T15:35:17.269Z	INFO	m2.raft	raft.node: 27f6db18bed0afb9 elected leader 2a4384d1c9ba4399 at term 3	{"member": "m2"}
    logger.go:130: 2022-10-30T15:35:17.269Z	INFO	m1.raft	ea2efb1261ea4b7e [term: 2] received a MsgApp message with higher term from 2a4384d1c9ba4399 [term: 3]	{"member": "m1"}
    logger.go:130: 2022-10-30T15:35:17.269Z	INFO	m1.raft	ea2efb1261ea4b7e became follower at term 3	{"member": "m1"}
[...]

PR #14824, opened 2022-11-22
tests/integration: deflake Corruption cases

fuweid:
If the corrupted member has been elected as leader, the memberID in alert response won't be the corrupted one. It will be a smaller follower ID since the raftCluster.Members always sorts by ID. We should check the leader ID and decide to use which memberID.

Fixes: #14823

### DIFF
--- a/tests/integration/corrupt_test.go
+++ b/tests/integration/corrupt_test.go
@@ -92,12 +92,31 @@ func TestPeriodicCheckDetectsCorruption(t *testing.T) {
 	assert.NoError(t, err)
 	time.Sleep(50 * time.Millisecond)
 	leader := clus.WaitLeader(t)
+
+	// Get sorted member IDs
+	members, err := cc.MemberList(ctx)
+	assert.NoError(t, err, "error on member list %v")
+
+	// NOTE: If the corrupted member has been elected as leader, the
+	// alarm will show the smaller member.
+	var expectedID = uint64(clus.Members[0].ID())
+	if leader == 0 {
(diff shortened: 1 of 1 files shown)


serathius (approved):
Great find!

serathius:
This flake shows that there is a issue with meaning of `MemberID` field. For corruption alarm it depend on which member is corrupted:
* If follower is corrupted, leader will set `MemberID` to corrupted follower.
* If leader is corrupted, leader will set `MemberID` to follower with lower ID (assuming that both followers are up during check).


Might be worth creating an issue to fix it, however I don't think this can be easily fixed. 
cc @ahrtr @ptabor 

At least I don't a solution now. First thoughts: validating that which member is corrupted requires  separate quorum of members. In 3 node clusters you need all 3 members up (2 members for quorum + 1 corrupted member), in 5 node clusters you need (3 members for quorum and 1 corrupted member). 

One possible fix would be reporting `MemberId` equal 0 and delaying setting true `MemberId` until enough members are up.

ahrtr:
Thanks @fuweid for fixing the flaky test case,  but it should be a product/etcd issue instead of a test case issue, so we should fix product code. At least, we should leave a TODO comment in the modified test case.

Actually I raised this comment when I was reviewing the code more than 4 months ago, https://github.com/etcd-io/etcd/pull/14120#discussion_r915542551. We should depend on quorum to identify the corrupted member, let me deliver a PR for this.

fuweid:
@ahrtr Looking forward to your pr!

PR #14825, opened 2022-11-22
tests: fix anti-pattern name in `NewEtcdProcessCluster`

clarkfw:
Context #14711 

Rename anti-pattern-naming fields(DisableStrictReconfigCheck and NoCN) and move them to DefaultConfig.

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -103,7 +103,7 @@ func NewConfigClientTLSCertAuthWithNoCN() *EtcdProcessClusterConfig {
 		WithClusterSize(1),
 		WithClientConnType(ClientTLS),
 		WithClientCertAuthority(true),
-		WithNoCN(true),
+		WithCN(false),
 	)
 }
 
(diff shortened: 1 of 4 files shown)


tests/framework/e2e/cluster.go
```diff
@@ -152,18 +152,18 @@ type EtcdProcessClusterConfig struct {
...
 	IsPeerAutoTLS bool
-	NoCN          bool
+	CN            bool
 
 	CipherSuites []string
 
-	ForceNewCluster            bool
-	InitialToken               string
-	QuotaBackendBytes          int64
-	DisableStrictReconfigCheck bool
```
fuweid, line 162: No sure we should change it.

Both Disable or No seems to enable it by default. If we switch to `CN`,  I think we should change all the places to `DefaultXXXConfig`, like [DefaultConfig][1]. What do you think?

[1]: <https://github.com/etcd-io/etcd/blob/ddd65e7fe2e2d6fbb9a89a07ddbef79412885659/tests/framework/e2e/cluster.go#L187>

serathius: Looks like DefaultConfig was updated. As this is just an internal test framework I don't see any negative impacts of changing the defaults.

fuweid: My bad. The `WithConfigXYZ` always call `DefaultConfig`. For sure, it looks good.

fuweid (approved):
LGTM(non-binding)

serathius (approved):

tests/framework/e2e/cluster.go
clarkfw: Thanks for taking the time to review, I should have explained it earlier😬

PR #14272, opened 2022-07-25
memberID equals zero in corruption alarm

serathius:
cc @ahrtr

Issue #13839, opened 2022-03-25
Proposals should include a merkle root


lavalamp:
In response to #13766 and past issues that get into the same condition:

It is possible for etcd to get into a condition where the databases on different replicas have different contents, **and then proceed committing changes**. There have been other routes to this in the past. I don't think it's possible to prevent all db corruption issues, but it is certainly possible to make etcd stop and recover when the DB doesn't match.

The reason etcd can proceed is because the protocol has replicas agree on the *contents of changes* but not on the *state of the db after applying the change*. The easiest fix for this is to adjust things to compute a hash covering the entire db. This can be done efficiently (log N hashing operations per db change) by using a merkle tree. If proposed changes also included a proposed new merkle root hash, a replica with a differing db would not be able to accept a change, and this condition would be caught the instant it happened.

(And it's recoverable at that point by reading the correct state from the other replicas. Moreover, depending on how you implement it, the merkle tree could be examined a layer at a time to find *what* is corrupted instead of needing to copy the entire state, which could be large.)

The merkle tree technique is very common in cryptocurrencies. etcd is basically a cryptocurrency with no token and a non-byzantine-resistant coordination mechanism. These are both correct tradeoffs given what etcd does, but continuing to accept changes to a corrupted state is extremely bad -- any cryptocurrency with a corresponding bug would totally go to zero.

lavalamp:
cc @jpbetz, whom I've tried to sell on this in the past :)

serathius:
Yep, I have pointed to problem of this issue going stale in https://github.com/etcd-io/etcd/issues/13775 and wanted to propose implementing https://github.com/etcd-io/etcd/issues/10893 as part of graduation of corruption check in https://github.com/etcd-io/etcd/issues/9190

Let's abovid tracking same issue in multiple places. @lavalamp do you want to propose implementing merkele trees as graduation criteria in https://github.com/etcd-io/etcd/issues/9190 and close this one?

lavalamp:
That sounds great to me! How do I do that?

ptabor:
+1.  Piece for consideration is whether:
  a) merkle tree should be integrated into bbolt, i.e. we can reliably ask bbolt at each transactional state about it's checksum.  This would guarantee bbolt level consistency and is probably the strongest option of physical consistency we can get... but requires impacting the data-storage format and might have biggest performance impact.
  
[...]

PR #14828, opened 2022-11-23
Identify corrupted member depending on quorum

ahrtr:
Currently when the compact hash checker detects hash mismatch, it assumes that the corrupted member is always one of the followers. This isn't correct, because it's also possible that it's the leader's data corrupted. It's also possible that there are multiple members corrupted, for example 2 members out of a 5 member cluster.

The solution is to depend on quorum to identify the corrupted member. For example, for a 3 member cluster, if 2 members have the same compactRevision and hash, then the left one member is the corrupted one. For a 5 member cluster, if at least 3 members have the same CompactRevision and hash, then the left members are the corrupted ones.

If there isn't a quorum, then the least minority are regarded as the corrupted member. For example, for a 5 member cluster, m1 and m2 have the same CompactRevision and hash, m3 and 4 have the same CompactRevision and hash, the m5 is the corrupted member.

### DIFF
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
 	"sort"
 	"strings"
@@ -343,36 +342,16 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 	}
 
 	if !quorumExist {
-		// If quorumExist doesn't exist, then only raise alarm for the least minority.
-		// The first step is to find out the members which are the least minority.
-		var (
-			minCnt    int = math.MaxInt
-			hashVal   uint32
-			memberIDs types.IDSlice
-		)
-
-		for k, v := range hash2members {
-			if v.Len() < minCnt {
-				minCnt = v.Len()
-				hashVal = k
-				memberIDs = v
-			}
-		}
-
-		// raise alarms
-		for _, pid := range memberIDs {
-			cm.hasher.TriggerCorruptAlarm(pid)
-		}
-		delete(hash2members, hashVal)
-
-		cm.lg.Error("Detected compaction hash mismatch but can't identify the corrupted members, so only raise alarm for the least minority",
+		// If quorum doesn't exist, we don't know which members data are
+		// corrupted. In such situation, we intentionally set the memberID
+		// as 0, it means it affects the whole cluster.
+		cm.lg.Error("Detected compaction hash mismatch but cannot identify the corrupted members, so intentionally set the memberID as 0",
 			zap.String("leader-id", leaderId.String()),
 			zap.Int64("leader-revision", leaderHash.Revision),
 			zap.Int64("leader-compact-revision", leaderHash.CompactRevision),
 			zap.Uint32("leader-hash", leaderHash.Hash),
-			zap.Uint32("peer-hash", hashVal),
-			zap.String("peer-ids", memberIDs.String()),
 		)
+		cm.hasher.TriggerCorruptAlarm(0)
 	}
 
 	// Raise alarm for the left members if the quorum is present.
@@ -391,7 +370,7 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 			zap.Uint32("leader-hash", leaderHash.Hash),
 			zap.Uint32("peer-hash", k),
 			zap.String("peer-ids", v.String()),
-			zap.Bool("alarm-raised", quorumExist),
+			zap.Bool("quorum-exist", quorumExist),
 		)
 	}
 
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -279,7 +279,7 @@ func (cm *corruptionChecker) CompactHashCheck() {
 //	false: skipped some members, so need to check next hash
 func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers []*peerHashKVResp) bool {
 	leaderId := cm.hasher.MemberId()
-	hashMap := map[uint32]types.IDSlice{leaderHash.Hash: {leaderId}}
+	hash2members := map[uint32]types.IDSlice{leaderHash.Hash: {leaderId}}
 
 	peersChecked := 0
 	// group all peers by hash
@@ -304,16 +304,16 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 		}
 
 		peersChecked++
-		if ids, ok := hashMap[peer.resp.Hash]; !ok {
-			hashMap[peer.resp.Hash] = []types.ID{peer.id}
+		if ids, ok := hash2members[peer.resp.Hash]; !ok {
+			hash2members[peer.resp.Hash] = []types.ID{peer.id}
 		} else {
 			ids = append(ids, peer.id)
-			hashMap[peer.resp.Hash] = ids
+			hash2members[peer.resp.Hash] = ids
 		}
 	}
 
 	// All members have the same CompactRevision and Hash.
-	if len(hashMap) == 1 {
+	if len(hash2members) == 1 {
 		if peersChecked == len(peers) {
 			cm.lg.Info("successfully checked hash on whole cluster",
 				zap.Int("number-of-peers-checked", peersChecked),
@@ -342,32 +342,25 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 	memberCnt := len(peers) + 1
 	quorum := memberCnt/2 + 1
 	quorumExist := false
-	for k, v := range hashMap {
+	for k, v := range hash2members {
 		if len(v) >= quorum {
 			quorumExist = true
 			// remove the majority, and we might raise alarms for the left members.
-			delete(hashMap, k)
+			delete(hash2members, k)
 			break
 		}
 	}
 
 	if !quorumExist {
-		// If quorumExist doesn't exist, then only raise alarm for the least minority
-		cm.lg.Error("Detected compaction hash mismatch but can't identify the corrupted members, so only raise alarm for the least minority",
-			zap.String("leader-id", leaderId.String()),
-			zap.Int64("leader-revision", leaderHash.Revision),
-			zap.Int64("leader-compact-revision", leaderHash.CompactRevision),
-			zap.Uint32("leader-hash", leaderHash.Hash),
-		)
-
-		// find out the members which are the least minority
+		// If quorumExist doesn't exist, then only raise alarm for the least minority.
+		// The first step is to find out the members which are the least minority.
 		var (
 			minCnt    int = math.MaxInt
 			hashVal   uint32
 			memberIDs types.IDSlice
 		)
 
-		for k, v := range hashMap {
+		for k, v := range hash2members {
 			if v.Len() < minCnt {
 				minCnt = v.Len()
 				hashVal = k
@@ -375,10 +368,11 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 			}
 		}
 
+		// raise alarms
 		for _, pid := range memberIDs {
 			cm.hasher.TriggerCorruptAlarm(pid)
 		}
-		delete(hashMap, hashVal)
+		delete(hash2members, hashVal)
 
 		cm.lg.Error("Detected compaction hash mismatch but can't identify the corrupted members, so only raise alarm for the least minority",
 			zap.String("leader-id", leaderId.String()),
@@ -392,12 +386,13 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 
 	// Raise alarm for the left members if the quorum is present.
 	// But we should always generate error log for debugging.
-	for k, v := range hashMap {
-		for _, pid := range v {
-			if quorumExist {
+	for k, v := range hash2members {
+		if quorumExist {
+			for _, pid := range v {
 				cm.hasher.TriggerCorruptAlarm(pid)
 			}
 		}
+
 		cm.lg.Error("Detected compaction hash mismatch",
 			zap.String("leader-id", leaderId.String()),
 			zap.Int64("leader-revision", leaderHash.Revision),
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -252,6 +252,17 @@ func (cm *corruptionChecker) PeriodicCheck() error {
 	return nil
 }
 
+// CompactHashCheck is based on the fact that 'compactions' are coordinated
+// between raft members and performed at the same revision. For each compacted
+// revision there is KV store hash computed and saved for some time.
+//
+// This method communicates with peers to find a recent common revision across
+// members, and raises alarm if 2 or more members at the same compact revision
+// have different hashes.
+//
+// We might miss opportunity to perform the check if the compaction is still
+// ongoing on one of the members or it was unresponsive. In such situation the
+// method still passes without raising alarm.
 func (cm *corruptionChecker) CompactHashCheck() {
 	cm.lg.Info("starting compact hash check",
 		zap.String("local-member-id", cm.hasher.MemberId().String()),
@@ -314,27 +325,7 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 
 	// All members have the same CompactRevision and Hash.
 	if len(hash2members) == 1 {
-		if peersChecked == len(peers) {
-			cm.lg.Info("successfully checked hash on whole cluster",
-				zap.Int("number-of-peers-checked", peersChecked),
-				zap.Int64("revision", leaderHash.Revision),
-				zap.Int64("compactRevision", leaderHash.CompactRevision),
-			)
-			cm.mux.Lock()
-			if leaderHash.Revision > cm.latestRevisionChecked {
-				cm.latestRevisionChecked = leaderHash.Revision
-			}
-			cm.mux.Unlock()
-			return true
-		}
-		cm.lg.Warn("skipped revision in compaction hash check; was not able to check all peers",
-			zap.Int("number-of-peers-checked", peersChecked),
-			zap.Int("number-of-peers", len(peers)),
-			zap.Int64("revision", leaderHash.Revision),
-			zap.Int64("compactRevision", leaderHash.CompactRevision),
-		)
-		// The only case which needs to check next hash
-		return false
+		return cm.handleConsistentHash(leaderHash, peersChecked, len(peers))
 	}
 
 	// Detected hashes mismatch
@@ -405,6 +396,30 @@ func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers
 	}
 
 	return true
+}
+
+func (cm *corruptionChecker) handleConsistentHash(hash mvcc.KeyValueHash, peersChecked, peerCnt int) bool {
+	if peersChecked == peerCnt {
+		cm.lg.Info("successfully checked hash on whole cluster",
+			zap.Int("number-of-peers-checked", peersChecked),
+			zap.Int64("revision", hash.Revision),
+			zap.Int64("compactRevision", hash.CompactRevision),
+		)
+		cm.mux.Lock()
+		if hash.Revision > cm.latestRevisionChecked {
+			cm.latestRevisionChecked = hash.Revision
+		}
+		cm.mux.Unlock()
+		return true
+	}
+	cm.lg.Warn("skipped revision in compaction hash check; was not able to check all peers",
+		zap.Int("number-of-peers-checked", peersChecked),
+		zap.Int("number-of-peers", peerCnt),
+		zap.Int64("revision", hash.Revision),
+		zap.Int64("compactRevision", hash.CompactRevision),
+	)
+	// The only case which needs to check next hash
+	return false
 }
 
 func (cm *corruptionChecker) uncheckedRevisions() []mvcc.KeyValueHash {
--- a/server/etcdserver/corrupt.go
+++ b/server/etcdserver/corrupt.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


ahrtr:
With the change in this PR, I reverted the change in https://github.com/etcd-io/etcd/pull/14824 

@fuweid @ptabor @serathius @spzala

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,157 @@ func (cm *corruptionChecker) CompactHashCheck() {
...
+			zap.Uint32("peer-hash", hashVal),
+			zap.String("peer-ids", memberIDs.String()),
+		)
+	}
+
+	// Raise alarm for the left members if the quorum is present.
+	// But we should always generate error log for debugging.
+	for k, v := range hashMap {
+		for _, pid := range v {
+			if quorumExist {
```
fuweid, line 397: what about this?

```
if quorumExist {
   for _, pid := range v {
       cm.hasher.TriggerCorruptAlarm(pid)
   }
}
```

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,157 @@ func (cm *corruptionChecker) CompactHashCheck() {
...
+				memberIDs = v
+			}
+		}
+
+		for _, pid := range memberIDs {
+			cm.hasher.TriggerCorruptAlarm(pid)
+		}
+		delete(hashMap, hashVal)
+
+		cm.lg.Error("Detected compaction hash mismatch but can't identify the corrupted members, so only raise alarm for the least minority",
```
fuweid, line 383: just wonder that we can merge this log with L356

```
cm.lg.Error("Detected compaction hash mismatch but can't identify the corrupted members, so only raise alarm for the least minority",
			zap.String("leader-id", leaderId.String()),
			zap.Int64("leader-revision", leaderHash.Revision),
			zap.Int64("leader-compact-revision", leaderHash.CompactRevision),
			zap.Uint32("leader-hash", leaderHash.Hash),
		)
```

It seems that L365 is duplicate.

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,42 @@ func TestCompactHashCheck(t *testing.T) {
...
+			name: "Peer returned same compaction revision but different hash triggers alarm for the least minority",
+			hasher: fakeHasher{
+				hashes: []mvcc.KeyValueHash{{Revision: 1, CompactRevision: 1, Hash: 1}, {Revision: 2, CompactRevision: 1, Hash: 2}},
+				peerHashes: []*peerHashKVResp{
+					{peerInfo: peerInfo{id: 42}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 44}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 45}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 46}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 4}},
+					{peerInfo: peerInfo{id: 47}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
```
fuweid, line 286: Would you add a case that all the members are running with different hash ?

server/etcdserver/corrupt.go
ahrtr: I forgot to remove the log at line 356. Removed.

server/etcdserver/corrupt.go
ahrtr: sg

server/etcdserver/corrupt_test.go
ahrtr: Added.

ahrtr:
Thanks @fuweid for the comments, all look good to me.

fuweid (approved):
LGTM(non-binding)

serathius:
Note that this is not something that was introduced for compact hash verification, the issue was already present in `PeriodicCheck`.

My suggestion: 
* Let's send a fix that removes setting `MemberID` in alarm check for both `PeriodicCheck` and `CompactHashCheck`. This can be backported to v3.5 and v3.4
* This PR that implements proper detection of which member is corrupted should be treated as a v3.6 feature.

What do you think?

ahrtr:
> Note that this is not something that was introduced for compact hash verification, the issue was already present in `PeriodicCheck`.

Yes, I was aware of it that `PeriodicCheck` has similar "issue". The reason why I did not update `PeriodicCheck` is that it has a little different & incorrect logic from the `CompactCheck`, please see my comment https://github.com/etcd-io/etcd/pull/14536#discussion_r983267367. So let's do this step by step. We can update & enhance `PeriodicCheck` in a separate PR. 



> * Let's send a fix that removes setting `MemberID` in alarm check for both `PeriodicCheck` and `CompactHashCheck`. This can be backported to v3.5 and v3.4

Seems like a good suggestion, but the only concern is that it may break the existing user experience. @ptabor @spzala what's your thought?



> * This PR that implements proper detection of which member is corrupted should be treated as a v3.6 feature.

Agreed. Let's do similar change for `PeriodicCheck` in a separate PR.

serathius:
> Seems like a good suggestion, but the only concern is that it may break the existing user experience. @ptabor @spzala what's your thought?

This was broken for long time until we fixed only recently https://github.com/etcd-io/etcd/pull/14272. A bug turned out to be a feature :P. I don't see an issue with backporting this.

Corruption checks previously set `MemberID` value based on member id field in  response for Hash request. When I investigated hash verification I found that member never sets memberid in the Hash response.

ahrtr:
> This was broken for long time until we fixed only recently #14272. A bug turned out to be a feature :P. I don't see an issue with backporting this.

OK. Let's discuss this in a separate session. Just raised a ticket to track it. https://github.com/etcd-io/etcd/issues/14849

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,157 @@ func (cm *corruptionChecker) CompactHashCheck() {
...
+
+		peersChecked++
+		if ids, ok := hashMap[peer.resp.Hash]; !ok {
+			hashMap[peer.resp.Hash] = []types.ID{peer.id}
+		} else {
+			ids = append(ids, peer.id)
+			hashMap[peer.resp.Hash] = ids
 		}
-		if len(peers) == peersChecked {
+	}
```
serathius, line 313: Could you just move it to separate function?

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,152 @@ func (cm *corruptionChecker) CompactHashCheck() {
 	)
 	hashes := cm.uncheckedRevisions()
 	// Assume that revisions are ordered from largest to smallest
-	for i, hash := range hashes {
+	for _, hash := range hashes {
 		peers := cm.hasher.PeerHashByRev(hash.Revision)
```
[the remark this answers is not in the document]
ptabor: Improvements for the backlog:
1. getPeerHashKVs() could fetch the hashes in paralle.
2. `ServeHTTP` could populate list of hashes we have (for consumption in v3.7, 2028)

ptabor (changes requested):

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,157 @@ func (cm *corruptionChecker) CompactHashCheck() {
...
+		for k, v := range hashMap {
+			if v.Len() < minCnt {
+				minCnt = v.Len()
+				hashVal = k
+				memberIDs = v
+			}
+		}
+
+		for _, pid := range memberIDs {
+			cm.hasher.TriggerCorruptAlarm(pid)
```
serathius, line 373: Don't agree with using `least minority` to alarm when we cannot confirm source of corruption. I would prefer to just report `0` which for alarms is meant that if affects whole cluster.

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,157 @@ func (cm *corruptionChecker) CompactHashCheck() {
...
+			}
+		}
+		cm.lg.Error("Detected compaction hash mismatch",
+			zap.String("leader-id", leaderId.String()),
+			zap.Int64("leader-revision", leaderHash.Revision),
+			zap.Int64("leader-compact-revision", leaderHash.CompactRevision),
+			zap.Uint32("leader-hash", leaderHash.Hash),
+			zap.Uint32("peer-hash", k),
+			zap.String("peer-ids", v.String()),
+			zap.Bool("alarm-raised", quorumExist),
```
serathius, line 403: This doesn't make sense, we always raise at this point in code even if `quorumExist = false`

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,42 @@ func TestCompactHashCheck(t *testing.T) {
...
 			name: "Peer returned different compaction revision is skipped",
 			hasher: fakeHasher{
 				hashes:     []mvcc.KeyValueHash{{Revision: 1, CompactRevision: 1}, {Revision: 2, CompactRevision: 2}},
 				peerHashes: []*peerHashKVResp{{resp: &pb.HashKVResponse{CompactRevision: 3}}},
 			},
-			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "PeerHashByRev(1)"},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "PeerHashByRev(1)", "MemberId()"},
 		},
 		{
 			name: "Peer returned same compaction revision but different hash triggers alarm",
```
serathius, line 265: Please add more test cases like:
* Leader is corrupted.
* 2/5 members are corrupted.

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,60 @@ func TestCompactHashCheck(t *testing.T) {
...
 			hasher: fakeHasher{
-				hashes:     []mvcc.KeyValueHash{{Revision: 1, CompactRevision: 1, Hash: 1}, {Revision: 2, CompactRevision: 1, Hash: 2}},
-				peerHashes: []*peerHashKVResp{{peerInfo: peerInfo{id: 42}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}}},
+				hashes: []mvcc.KeyValueHash{{Revision: 1, CompactRevision: 1, Hash: 1}, {Revision: 2, CompactRevision: 1, Hash: 2}},
```
serathius, line 267:
```suggestion
				hashes: []mvcc.KeyValueHash{{Revision: 2, CompactRevision: 1, Hash: 2}},
```

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,152 @@ func (cm *corruptionChecker) CompactHashCheck() {
 	)
```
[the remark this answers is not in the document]
serathius: Would be great if points listed by @ptabor are better conveyed in code too.

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
serathius: For v3.6 release we should consider hash being negotiated via raft.

server/etcdserver/corrupt.go
```diff
@@ -258,57 +259,152 @@ func (cm *corruptionChecker) CompactHashCheck() {
...
-				zap.String("peer-id", p.id.String()),
-			)
+// check peers hash and raise alarms if detected corruption.
+// return a bool indicate whether to check next hash.
+//
+//	true: successfully checked hash on whole cluster or raised alarms, so no need to check next hash
+//	false: skipped some members, so need to check next hash
+func (cm *corruptionChecker) checkPeerHashes(leaderHash mvcc.KeyValueHash, peers []*peerHashKVResp) bool {
+	leaderId := cm.hasher.MemberId()
+	hashMap := map[uint32]types.IDSlice{leaderHash.Hash: {leaderId}}
```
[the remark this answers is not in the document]
ahrtr: good call & updated.

server/etcdserver/corrupt.go
ahrtr: Makes sense & updated. It's clearer.

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
ahrtr: Really good summary. We should add it in the first place. Basically it's exactly how this PR is implemented (of course I need to update the least minority mechanism) . 

With some minor update to the summary, it's now:
```

// CompactHashCheck is based on the fact that 'compactions' are coordinated
// between raft members and performed at the same revision. For each compacted
// revision there is KV store hash computed and saved for some time.
//
// This method communicates with peers to find a recent common revision across
// members, and raises alarm if 2 or more members at the same compact revision
// have different hashes.
//
// We might miss opportunity to perform the check if the compaction is still
// ongoing on one of the members or it was unresponsive. In such situation the
// method still passes without raising alarm. 
```

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
ahrtr: Let's discuss & address this in a separate session/PR.  Renaming and fetching the hashes in parallel make sense to me. @ptabor  where & how is the `2028` coming from? :) 

For @serathius 's comment "`consider hash being negotiated via raft`", I did not get the point. My immediate feeling there is no need, because the compaction is already coordinated by raft.

server/etcdserver/corrupt.go
ahrtr: Agreed. It's align with what we did for 3.4 and 3.5

ahrtr: updated.

server/etcdserver/corrupt.go
ahrtr: updated.

server/etcdserver/corrupt_test.go
ahrtr: updated.

server/etcdserver/corrupt_test.go
ahrtr: Added all your suggested and some additional test cases.

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
fuweid: > we should consider hash being negotiated via raft.

Is it going to persist the hash result?

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+				peerHashes: []*peerHashKVResp{
+					{peerInfo: peerInfo{id: 42}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(42)"},
+			expectCorrupt: true,
+		},
+		{
+			name: "Peer returned different hash but no quorum triggers alarm with memberID 0",
```
serathius, line 277: This test seems redundant to other tests

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+					{peerInfo: peerInfo{id: 45}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 46}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 4}},
+					{peerInfo: peerInfo{id: 47}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(0)"},
+			expectCorrupt: true,
+		},
+		{
+			name: "Peer returned same compaction revision but all members have different hashes",
```
serathius, line 293: This test seems redundant to other tests

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+					{peerInfo: peerInfo{id: 45}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 6}},
+					{peerInfo: peerInfo{id: 46}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 7}},
+					{peerInfo: peerInfo{id: 47}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 8}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(0)"},
+			expectCorrupt: true,
+		},
+		{
+			name: "1 in 3 peer unresponsive checks next hash",
```
serathius, line 309: This test is redundant to `Peer error` test. Name is better thou.

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+				peerHashes: []*peerHashKVResp{
+					{err: fmt.Errorf("failed getting hash")},
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "PeerHashByRev(1)", "MemberId()"},
+			expectCorrupt: false,
+		},
+		{
+			name: "1 in 5 peer unresponsive checks next hash",
```
serathius, line 321: Testing 1/5 vs 1/3 for unresponsive check doesn't seem too different.

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+					{peerInfo: peerInfo{id: 44}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+					{peerInfo: peerInfo{id: 45}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "PeerHashByRev(1)", "MemberId()"},
+			expectCorrupt: false,
+		},
+		{
+			name: "1 in 3 peer corrupted triggers alarm",
```
serathius, line 335:
```suggestion
			name: "Etcd can identify single corrupted member",
```

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+				peerHashes: []*peerHashKVResp{
+					{peerInfo: peerInfo{id: 42}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(43)"},
+			expectCorrupt: true,
+		},
+		{
+			name: "1 in 5 peer corrupted triggers alarm",
```
serathius, line 347:
```suggestion
			name: "Etcd can identify single corrupted member in 5 member cluster",
```

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+				peerHashes: []*peerHashKVResp{
+					{peerInfo: peerInfo{id: 42}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+					{peerInfo: peerInfo{id: 44}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 45}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(44)"},
+			expectCorrupt: true,
+		},
```
serathius, line 359: Can we add a test "Etcd can indentify two corrupted members in 5 member cluster"?

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+		{
+			name: "1 in 3 peer unresponsive and 1 in 3 peer corrupted triggers alarm on memberID 0",
```
serathius, line 361:
```suggestion
			name: "Etcd triggers corrupted alarm on whole cluster if in 3 member cluster one member is down",
```

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+				peerHashes: []*peerHashKVResp{
+					{err: fmt.Errorf("failed getting hash")},
+					{peerInfo: peerInfo{id: 43}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(0)"},
+			expectCorrupt: true,
+		},
+		{
+			name: "1 in 5 peer unresponsive and 1 in 5 peer corrupted triggers alarm",
```
serathius, line 373:
```suggestion
			name: "Etcd can identify corrupted member in 5 member cluster even if one member is down",
```

server/etcdserver/corrupt_test.go
ahrtr: updated.

server/etcdserver/corrupt_test.go
```diff
@@ -251,23 +251,162 @@ func TestCompactHashCheck(t *testing.T) {
...
+					{err: fmt.Errorf("failed getting hash")},
+					{peerInfo: peerInfo{id: 44}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 3}},
+					{peerInfo: peerInfo{id: 45}, resp: &pb.HashKVResponse{CompactRevision: 1, Hash: 2}},
+				},
+			},
+			expectActions: []string{"MemberId()", "ReqTimeout()", "Hashes()", "PeerHashByRev(2)", "MemberId()", "TriggerCorruptAlarm(44)"},
+			expectCorrupt: true,
+		},
+		{
+			name: "leader corrupted in 3 member cluster triggers alarm",
```
serathius, line 387:
```suggestion
			name: "Etcd can identify that leader is corrupted",
```

ahrtr: updated.

serathius:
Proposed to rename some tests to make it easier to identify if we are missing any tests, but feel free to skip suggestions if you don't agree with them. Would love to see consistent naming for those scenarios, but maybe in next PR.

server/etcdserver/corrupt_test.go
ahrtr: done.

server/etcdserver/corrupt_test.go
ahrtr: remove one case.

ahrtr:
> Proposed to rename some tests to make it easier to identify if we are missing any tests, but feel free to skip suggestions if you don't agree with them. Would love to see consistent naming for those scenarios, but maybe in next PR.

Resolved all the comments. Renaming isn't a big deal. But you are the original author the unit test case, so I followed all your suggestion. 

PTAL, thx.

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
ptabor: > how is the 2028 coming from ?  

Just extrapolation or release frequency

> Rafting hashes ?

I imagine this could work that way that way: 
  1)  there is new RAFT message 'start-checksum' that triggers all members to compute checksum at the exact revision. So it's 'simpler' compaction.    Compaction stays as doing this implicitly. 
  2) Whenever member finishes the computation it sends to leader their result (pair: rev, hash). 
  3) Leader broadcasts ? the received results through RAFT
  4) Every-member can react on discrepancy. 

Benefit: Does not require custom service and best-effort attempts to check whether we have checks-in-sync. 

But I would consider evaluating (on-line) merkle root ( https://github.com/etcd-io/etcd/issues/13839 ) sums design first and thinking what raft changes would be needed for this.

ptabor (approved):

server/etcdserver/corrupt.go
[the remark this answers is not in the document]
ahrtr: Thanks @ptabor for the comment.  It's a big topic, let's discuss it separately.

thalinda-oss (approved):

Issue #14832, opened 2022-11-23
limit the size of the etcd wal file


new-dream:
### What would you like to be added?

add a startup flag to limit the size of the WAL file.

### Why is this needed?

The size of disk is strictly limited in our product so the disk storage space for etcd cannot be sufficient. Therefore, we want to limit the size of the etcd wal file.

ptabor:
Currently the wal needs to be at least 128MB + 64MB for 'preallocated page' for next WAL segment. 
How low do you need to go ?

PR #14833, opened 2022-11-23
wal: add wal-segment-size-bytes flags to limit size of each wal segment file

new-dream:
fix [14832](https://github.com/etcd-io/etcd/issues/14832)


lavacat:
You probably want to make this flag `--experimental`

See https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/features.md#adding-a-new-feature
> Any configuration flags related to the implementation of the feature must be prefixed with experimental e.g. --experimental-feature-name.

ptabor:
Added additional questions about the usecase in the original bug: https://github.com/etcd-io/etcd/issues/14832

PR #14835, opened 2022-11-23
server: optimizing memory overhead of copy operation in ConcurrentReadTxn

new-dream:


ptabor:
Could you add a test that presents where it matters in practice (and also avoid regression in the future). 

Trimming a buffer as part of copy makes sense (but we need to be more explicit in naming, e.g. `CopyUsed` or `TrimmedCopy`) to avoid miss-assumption about copy being equivalent of source.

ahrtr:
The change looks good to me. 

Please add at least unit test cases to verify the change.  Another point is the value of `used`  should never be bigger than len(buf), probably we can add a verification. See [verify.go#L76](https://github.com/etcd-io/etcd/blob/b6a35947100f98920e4adbb80f475f08a44cdff7/client/pkg/verify/verify.go#L76)

ptabor (changes requested):
(formally requesting for tests - to remove PR from reviews that need attention)

Issue #14820, opened 2022-11-22
Unified test framework


ahrtr:
### What would you like to be added?

We reinvented some wheels on test frameworks. Currently there are functional test, e2e test and linearizability test (based on e2e). The Jepsen test is also an option, but it's written in Clojure, and out of the existing maintainers' control, so it isn't in the scope for now.

Each test framework has pros and cons. 

## Functional test
**Pros**
1. The functional test has a well-designed framework, and it's flexible & extensible. Currently it only checks hash (and also has lease related check), but we can easily add more checker. 
2. It supports flexible failure injections via [gofail](https://github.com/etcd-io/gofail). It lists all the available failpoints automatically, and users just need to configure the commands/terms in [failpoint-commands](https://github.com/etcd-io/etcd/blob/9cc2f645cbdcca49cf3779a07028088ef935321b/tests/functional/functional.yaml#L215-L218), then it automatically creates all the possible combinations. For example, assuming there are 100 failpoints, and we define 3 commands/terms, then there will be 100*3 cases of failure injection.
3. It also supports other kinds of failure injection, such as delay, network blackhole etc.

**Cons**
1. It reinvented an [agent](https://github.com/etcd-io/etcd/tree/main/tests/functional/agent) to start/stop etcd clusters. Ideally it should use the same utilities as the e2e test.
2. It's out of maintenance, and accordingly flaky. This shouldn't be a big deal, as long as we decided to spend more time on it.

## Linearizability test
**Pros**
1. Linearizability is really important for distributed system. It's good to have a test framework to focus on verifying the Linearizability of etcd.

**Crons**
 1. It's based on a personal repo [porcupine](https://github.com/anishathalye/porcupine), which isn't actively maintained. When a test case reports failure, I don't think we even have enough confidence that the failure isn't caused by the test framework/utility. 
 2. It isn't well-designed, and of course isn't flexible & easily extensible. All the concepts (e.g. traffic player, failure injection, result checker etc.) are basically tightly coupled. 
 3. It's doing things that are beyond what the concept of Linearizability can hold. Conceptually, Linearizability test is just part of the functional test, but it's trying to do the same thing as the functional test. 

## Unified test framework
So we really need to create a unified test framework, otherwise future maintainers may reinvent a new test framework. It also isn't good to add too many code to the existing Linearizability test without having a well-designed framework beforehand. 

[...]

PR #14838, opened 2022-11-23
tests: Document robustness tests

serathius:

cc @ahrtr @ptabor @spzala 
Do you have any suggestion on topics to cover in the README?


tests/linearizability/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+As proposed in [v3.5 data inconsistency postmortem], etcd needed a new framework
+for release qualification. A framework that can not only be able to prevent
+historical issues, but also discover new issues. At the time etcd had a big
+problem of test flakiness, so it was also important to easily distinguish between
+etcd problem and a bug in testing.
+
+In the end we were inspired by [Jepsen] and implemented a similar set of tests
+that would validate etcd [linearizability]. The framework is able to reproduce
+historical inconsistency issues by inject specific failpoints into etcd server.
```
ptabor, line 21: nit: inject-> "injecting" or "injection of"

tests/linearizability/README.md
```diff
@@ -0,0 +1,102 @@
...
+e2e binaries to create a cluster and validate etcd behavior by sending requests
+to API. There are however to major differences. In addition to sending requests
+they also inject different failures into cluster (for example kill etcd process).
+Instead of running any specific test scenario (for example 1. put value 2. get
+value 3. compare if read equals put) they record all the history of requests
+and try to replay them against a simplified model of etcd.
+
+Steps in linearizability tests
+
+1. Setup a etcd cluster
```
ptabor, line 42: "an etcd" seems to be more popular (at least when grepping our repo or https://etcd.io/docs/v3.5/faq/)

tests/linearizability/README.md
```diff
@@ -0,0 +1,102 @@
...
+2. Send traffic to each member and record all requests
+3. Inject failures into the etcd clusters
+4. Validate request history against a etcd model
+
```
ptabor, line 46: It would be good to mentioned the relationship with to the 'functional' tests. I assumes the differences are following: 

 - functional tests are using multiple processes both: for servers and clients, while linearizable are combining (at least clients). 
 - functional tests were focusing whether after failure the cluster will recover, i.e. will be able to continue serving, not verifying the data correctness against requests. 
 - functional were more 'randomized' while linearizable are aiming at more determinsm. 

Are there other fundamental differences ?

ahrtr: Based on the discussion in https://github.com/etcd-io/etcd/issues/14820, we need to unify the linearizablity and function test. Eventually we will only maintain one unified test.

geetasg:
It would be nice to have a contributor section that guides on how to extend the model.

tests/linearizability/README.md
geetasg: From discussion above, seems like contributors should not be adding any new functional tests but add new tests here. It would help if this Readme can clarify.

tests/linearizability/README.md
```diff
@@ -0,0 +1,102 @@
...
+    FAILPOINTS=true make build
+    ```
+2. Run linearizability tests
+
+    ```bash
+    make test-linearizability
+    ```
+
+### Reading the test results
+
```
geetasg, line 97: we should clarify the info that user should submit if they decide to file any issue regarding linearizability test failure that they need help debugging.

serathius:
Finally got to writing down the docs. 
cc @ahrtr @ptabor @logicalhan @chaochn47 @jmhbnz

logicalhan (approved):
/lgtm
/approve

tests/robustness/README.md
```diff
@@ -0,0 +1,71 @@
...
+[API guarantees] and [watch guarantees] under any condition or failure.
+
+Robustness tests achieve that comparing etcd cluster behavior against a simplified model.
+Multiple test encompass different etcd cluster setups, client traffic types and failures experienced by cluster.
+During a single test we create a cluster and inject failures while sending and recording client traffic.
+Correctness is validated by running collected history of client operations against the etcd model and a set of validators.
+Upon failure tests generate a report that can be used to attribute whether failure was caused by bug in etcd or test framework. 
+
+[API guarantees]: https://etcd.io/docs/v3.6/learning/api_guarantees/
+[watch guarantees]: https://etcd.io/docs/v3.6/learning/api/#watch-streams
```
ahrtr, line 13: cc @spzala to double confirm the link. Can we count on "`latest`"? 

```suggestion
[API guarantees]: https://etcd.io/docs/latest/learning/api_guarantees/
[watch guarantees]: https://etcd.io/docs/latest/learning/api/#watch-streams
```

Makefile
```diff
@@ -137,7 +137,52 @@ gofail-disable: install-gofail
...
+
+.PHONY: reproduce-issue13766
+reproduce-issue13766: bin/etcd-v3.5.2-failpoints
+	cp bin/etcd-v3.5.2-failpoints ./bin/etcd
+	GO_TEST_FLAGS='-v --run=TestRobustness/Issue13766 --count 100 --failfast' make test-robustness
+
+.PHONY: reproduce-issue14685
+reproduce-issue14685: bin/etcd-v3.5.5-failpoints
+	cp bin/etcd-v3.5.5-failpoints ./bin/etcd
+	GO_TEST_FLAGS='-v --run=TestRobustness/Issue14685 --count 100 --failfast' make test-robustness
```
ahrtr, line 155: I recall I raised a comment somewhere that the top level `Makefile` isn't a good place to just demonstrate the ability to the robustness test, and @ptabor suggested to move them into a script or doc inside a sub-directory.  

I still have the same comment. Let's keep the top level Makefile as simple & clean as possible.

serathius: Don't agree with this. Robustness test targets touch files whole repo making it impossible/bothersome to put it in subdirectory. For example moving proposed targets to `tests/robustness/Makefile` would result in all paths needing to be prefixed `cd ../.. &&`.

> Let's keep the top level Makefile as simple & clean as possible. 

Don't agree, each Makefile target should be simple and clean, however it doesn't mean that root Makefile should have limited list of targets. I would argue that makefile scales pretty well with number of targets, assuming that we maintain a proper structure. For example https://github.com/kubernetes-sigs/metrics-server/blob/master/Makefile

serathius: For me the problem is that we want to avoid adding more targets to root Makefile for it to remain `simple & clean`, however I expect that it will cause reverse effect. Make code more complicated by introducing hard to follow dependencies. Proposed targets cannot be simply moved to subdir as they depend on targets (like `GOFAIL_VERSION`, `gofail-enable`, `gofail-disable`, `test-robustness`) and directories (like `./bin`) from root directory.

Simple and clean solution here is just put those targets in root directory. Distributing makefile targets between subdirectories is the more complicated option. Makefile was just not designed for hierarchy.

ahrtr: I didn't propose to add another Makefile in a sub-directory.

serathius: Oh, I misread.

> @ptabor suggested to move them into a script or doc inside a sub-directory.

ptabor: I still would prefer to split the files. Even for sake of avoiding long content in the main directory. 
My approach would be following: 

1.  Decide how we want to run robustness tests against historical versions long-term:
2. Address in a similar way the historical cases the requires very specific etcd revision for repro. 

----

1.  I think that the build code should be as much as possible versioned together with the System-Under-Test. In this case for 3.4.x, 3.5.x, 3.6.0+ I would cherry pick to the main Makefile a rule that allows to build:  

```
make build-with-failpoints FAILPOINTS=${GO_FAIL_VERSION}
```
that as an outcome would produce `./bin/etcd-failpoints` (and potentially the other binaries)

This would allow us to have in the main file a generic rule:
```
./tmp/etcd-${}/bin/etcd-failpoints:
  (clones the repo and calls the nested build)
  
robustness-test-{...}: ./tmp/etcd-{...}/bin/etcd-failpoints 
    ETCD_BIN=./tmp/etcd-{...}/bin/etcd-failpoints  make test-robustness
```

2. For the specific repro, I would consider using the Makefile's include (https://www.gnu.org/software/make/manual/html_node/Include.html)

So having the code in `./tests/robustness/makefile.mk` and including it from the main MakeFile. 

I think that such inclusion does not causes the `cd ../..` problems still allowing to encapsulate the robustness related code with the robustness tests.

tests/robustness/README.md
serathius: Done.

tests/robustness/README.md
```diff
@@ -0,0 +1,71 @@
...
+### Example analysis of linearization issue
+
+Let's analyse issue [#14370].
+To reproduce the issue by yourself run `make reproduce-issue14370`.
+After a couple of tries robustness tests should report `Model is not linearizable` and save report locally.
+Lineralization issues are easiest to analyse via history visualization. 
+Open `/tmp/TestRobustness_Issue14370/history.html` file in your browser.
+Jump to the error in linearization by clicking `[ jump to first error ]` on the top of the page.
+
+You should something like image below.
```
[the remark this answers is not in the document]
serathius: Fixed

tests/robustness/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+    ```bash
+    make test-robustness
+    ```
+   
+    Optionally you can pass environment variables:
+    * `GO_TEST_FLAGS` - to pass additional arguments to `go test`. 
+      It is recommended to run tests multiple times with failfast enabled. this can be done by setting `GO_TEST_FLAGS='--count=100 --failfast'`.
+    * `EXPECT_DEBUG=true` - to get logs from the cluster.
+    * `RESULTS_DIR` - to change location where results report will be saved.
```
ptabor, line 33: What's the default for RESULTS_DIR ? within the test's tmp directory that is cleaned after success ?

ptabor (approved):

tests/robustness/README.md
serathius: No, robustness tests specifically create its own directory in tmp that is not created via `t.TmpDir()`

scripts/test_lib.sh
```diff
@@ -315,7 +315,7 @@ function go_test {
     additional_flags=$(${flags_for_package_func} ${pkg})
 
     # shellcheck disable=SC2206
-    local cmd=( go test ${goTestFlags} ${additional_flags} "$@" ${pkg} )
+    local cmd=( go test ${goTestFlags} ${additional_flags} ${pkg} "$@" )
```
serathius, line 318: Flags defined by tests like `--bin-dir` need to be passed after package.

PR #14839, opened 2022-11-23
doc: add setup steps for client dev environment

shreemaan-abhishek:
I think this is necessary to include for clarity. Especially for newbies. To test/use the etcd client locally, I personally had to spend a lot of time making my test project depend on the local etcd client instead of the downloaded version as I did not have a proper idea of how to do it. [This](https://github.com/etcd-io/etcd/discussions/14737) GitHub discussion solved my issue.

Having this section in the contributing guide would bridge the gap greatly.

Please let me know if you would like to see any edits/clarifications.

Signed-off-by: Abhishek Choudhary <shreemaan.abhishek@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,6 +57,21 @@ Setup environment:
   - For ubuntu and debian run `sudo apt-get install build-essentials`
 - Verify that everything is installed by running `make build`
 
+Development environment for etcd client:
+
+If you want to quickly test the sanity of your code changes to the etcd client. You can create a project that depends on the local etcd client as such:
+
+- Create a go project
+- Use [go mod replace](https://go.dev/ref/mod#go-mod-file-replace) to use depend on local code instead of the published one.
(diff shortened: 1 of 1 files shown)


shreemaan-abhishek:
How do y'all perform sanity tests for etcd? I think your answer might be unit tests.

But if someone wants to play around and tinker with etcd this should be the way to go 😁

ahrtr:
> But if someone wants to play around and tinker with etcd this should be the way to go 😁

This should be a generic problem, not just specific to etcd. When you locally play with whatever client sdk, you need use `replace` directive to replace with local directory and replace all dependencies (e.g. api/v3, client/pkg/v3) which the client sdk depends on.

shreemaan-abhishek:
> I would suggest to update [client/v3/README.md](https://github.com/etcd-io/etcd/blob/main/client/v3/README.md) to explicitly mention that it depends on api/v3 and client/pkg/v3.

I think someone planning to tinker with the project would likely check out the CONTRIBUTING.md file instead of the [client/v3/README.md](https://github.com/etcd-io/etcd/blob/main/client/v3/README.md). This is why I think the changes should go to `CONTRIBUTING.md`. (Or probably to both the places)

Issue #14837, opened 2022-11-23
add new function in etcdutl to check if wal and db files is fine.


new-dream:
### What would you like to be added?

add new function in etcdutl to check if wal and db files is fine.
such as `etcdutl check broken`

### Why is this needed?

The function will detect wrong data files in one of etcd instance and then delete this file and recover to join the cluster.

lavacat:
Can you describe what type of `broken`? Is this to avoid `data-dir used by this member must be removed` message during etcd joining cluster?

If yes, I think the idea is to leave data intact so operator can inspect it and make decision.

new-dream:
I added a pr [14844](https://github.com/etcd-io/etcd/pull/14844) to do this.

This command checks whether the wal file is normal .
Second, check whether the db file can be opened normally.
Finally, try to read the consistent index from the db file.
If there are no errors and panics, the command returns 0.

How to use this command:
1. This command is invoked when the etcd process exits with error codes ranging from 1 to 127.
2. If a non-zero value is returned, check whether all other etcd processes in the cluster are normal.
3. If yes, delete the etcd data on the node, delete the member from the cluster, and add the member to the cluster again.

new-dream:
we have encountered the following panic:
```panic: tocommit(1808562) is out of range [1808561]. Was the raft log corrupted, truncated, or lost?```
```panic: freepages: failed to get all reachable pages (page 633: multiple references)```
etc.

Issue #10166, opened 2018-10-09
raft: Don't panic on commit index regression


bdarnell:
We occasionally see issues in production in which writes to our raft log are not fully persisted before MsgAppResps are sent (Sometimes this is deliberate, as when users disable fsync for performance. Sometimes it's a bug in some layer or another). The raft package does not handle this gracefully: if a follower has sent an MsgAppResp for a given index, the leader will send it that index as committed in future MsgApps. If the raft log was not written successfully, this will cause the follower to crash with `panic: tocommit(265625) is out of range [lastIndex(265624)]. Was the raft log corrupted, truncated, or lost?`. 

The message is correct because the log *was* truncated, and this behavior could result in a loss of committed writes if it happened on multiple nodes. However, this is a very severe failure mode. The follower panics, and will repeatedly panic until its disk is wiped and it is started from scratch. In most cases a less extreme recovery path is possible. If the leader still has the log entries in question, it can simply re-send them, and if it does not, it can send a snapshot. I'd like to at least have the option to disable this panic and recover a node in-place when it reaches this state.

I haven't thought this all the way through, but I think the solution would involve synthesizing an MsgAppResp with Reject: true and an appropriate IndexHint when the panic would be triggered.

xiang90:
>  I'd like to at least have the option to disable this panic and recover a node in-place when it reaches this state.

how about adding a corrupted hint in ready or something like that? so the upper layer can do something when raft thinks itself is in a bad state?

bdarnell:
Maybe (that would be an alternative to the panics we have all over the place now), but I think it might be difficult to use correctly in this case. Here we know we're missing information but we can recover if we get the right log entries. "Corruption" in the general case is difficult (if not impossible) to handle without just crashing and giving up.

chaochn47:
Another use case not to panic write away:
- add a new member with etcd db snapshot to speed up time ready to accept client traffic. 

To support this, a few more things may need to be changed. etcd db snapshot as backup right now is only for creating a fresh new cluster with that data as disaster recovery mechanism.

lavacat:
For ref, here is original commit that added `commitTo` with `panic`
https://github.com/etcd-io/etcd/pull/1740

ahrtr:
@tbg Is this issue still valid? Please feel free to raise an issue in the new raft repo if needed.

tbg:
I raised it here again: https://github.com/etcd-io/raft/issues/18
Thanks for the heads up.

PR #14844, opened 2022-11-24
etcdutl: add check broken command to check whether the data file is broked

new-dream:
fix [14837](https://github.com/etcd-io/etcd/issues/14837)

### DIFF
--- a/etcdutl/etcdutl/check_command.go
+++ b/etcdutl/etcdutl/check_command.go
@@ -1,0 +1,120 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package etcdutl
+
+import (
+	"encoding/binary"
+	"fmt"
+	"os"
+
+	"go.etcd.io/etcd/pkg/v3/cobrautl"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
+	"go.etcd.io/etcd/server/v3/storage/backend"
+	"go.etcd.io/etcd/server/v3/storage/datadir"
+	"go.etcd.io/etcd/server/v3/storage/schema"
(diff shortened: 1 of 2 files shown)


etcdutl/etcdutl/check_command.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	cmd.Flags().StringVar(&checkDataDir, "data-dir", "", "Path to the output data directory")
+	cmd.MarkFlagRequired("data-dir")
+
+	return cmd
+}
+
+func newCheckBrokenCommand(cmd *cobra.Command, args []string) {
+	walSnaps, err := wal.ValidSnapshotEntries(GetLogger(), datadir.ToWalDir(checkDataDir))
```
lavacat, line 65: nit: I don't think this will help with `panic: tocommit(1808562) is out of range [1808561]` error that you've mentioned in the issue [comment](https://github.com/etcd-io/etcd/issues/14837#issuecomment-1325906170) Here is longer [discussion](https://github.com/etcd-io/etcd/issues/10166) why that error happens.

etcdutl/etcdutl/check_command.go
```diff
@@ -0,0 +1,120 @@
...
+		Run:   newCheckBrokenCommand,
+	}
+
+	cmd.Flags().StringVar(&checkDataDir, "data-dir", "", "Path to the output data directory")
+	cmd.MarkFlagRequired("data-dir")
+
+	return cmd
+}
+
+func newCheckBrokenCommand(cmd *cobra.Command, args []string) {
```
lavacat, line 64: Have you seen [Verify](https://github.com/etcd-io/etcd/blob/c1da9dd0f0411293d1c3555494caedabd90926e9/server/verify/verify.go#L49) function? Logic seams to be similar. Maybe it can be reused and command can be renamed to `verify`

lavacat (commented):
One of the maintainers should comment if this command is needed.

etcdutl/etcdutl/check_command.go
ptabor: Thank you Bogdan, 
I'm all for driving this around 'Verify' that is also used in tests to check whether they are living etcd storage in the consistence state and should be expanded with additional checks.

ptabor (changes requested):
1. The command needs a test. 
2. Naming:  'check', 'verify' or 'fsck'  
 - Reason for `check` is that bbolt has check as well: https://github.com/etcd-io/bbolt/blob/eedea6cb26ef7ce2d093402416fe00d3a81fce64/cmd/bbolt/main.go#L181
   (we might call it as part of this command as well)
 - `verify` is most consistent with our internal naming 
 - `fsck` most familiar to linux users. Opens possibility to add --repair flag.
I think that `check` for surface consistency with bolt makes sense, but no strong opinion.

Issue #14846, opened 2022-11-24
TestSignal unit test failes


mkumatag:
### What happened?

```shell
=== Failed
=== FAIL: expect TestSignal (0.01s)
    expect_test.go:192: 
        	Error Trace:	expect_test.go:192
        	            				asm_ppc64x.s:899
        	Error:      	Error "unexpected exit code [-1] after running [/bin/sleep 100]" does not contain "unexpected exit code [-1] after running [/usr/bin/sleep 100]"
        	Test:       	TestSignal
```
### What did you expect to happen?

Test should pass

### How can we reproduce it (as minimally and precisely as possible)?

$ make test

### Anything else we need to know?

In the test the error message comparison fails because sleep is in a different path in the machine

```shell
root@85b22abf26c3:/workspace# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
root@85b22abf26c3:/workspace# which sleep
/bin/sleep
root@85b22abf26c3:/workspace#
```

test case error check needs to be more generic to handle such situations.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

mkumatag:
Introduced in https://github.com/etcd-io/etcd/pull/14672

PR #14847, opened 2022-11-24
fix TestSignal test

mkumatag:

Fixes: #14846

### DIFF
--- a/pkg/expect/expect_test.go
+++ b/pkg/expect/expect_test.go
@@ -189,7 +189,8 @@ func TestSignal(t *testing.T) {
 	go func() {
 		defer close(donec)
 		err = ep.Close()
-		assert.ErrorContains(t, err, "unexpected exit code [-1] after running [/usr/bin/sleep 100]")
+		assert.ErrorContains(t, err, "unexpected exit code [-1]")
+		assert.ErrorContains(t, err, "sleep 100")
 	}()
 	select {
 	case <-time.After(5 * time.Second):


Rajalakshmi-Girish:
\lgtm

pkg/expect/expect_test.go
```diff
@@ -189,7 +190,7 @@ func TestSignal(t *testing.T) {
 	go func() {
 		defer close(donec)
 		err = ep.Close()
-		assert.ErrorContains(t, err, "unexpected exit code [-1] after running [/usr/bin/sleep 100]")
+		assert.Regexp(t, regexp.MustCompile(`unexpected exit code \[-1\] after running \[.*sleep 100\]`), err)
```
serathius, line 193:
```suggestion
		assert.ErrorContains(t, err, "unexpected exit code [-1]")
```
@tjungblu to confirm

mkumatag: I don't mind accepting the suggestion and ignore the rest of the message :) lets see what @tjungblu says.!

ptabor: We don't care about the path... but I think that it's convenient that expect writes which command failed,
so I'm on the side of keeping check whether `sleep 100` is mentioned. 

The original change or: 
```
assert.ErrorContains(t, err, "unexpected exit code [-1]")
assert.ErrorContains(t, err, "sleep 100")
```
both works for me.

mkumatag: I feel multiple asserts are better and easy to read as well, lemme adopt them instead of regex.

ptabor (approved):

ptabor:
I see @serathius thumb-up on my comment, so I think its ok to merge. Thank you @mkumatag .

PR #14858, opened 2022-11-26
fix release pipeline failure

ahrtr:
I forgot to execute ./script/fix.sh in previous PR. :(

cc @serathius


ahrtr:
The Release pipeline is broken,

https://github.com/etcd-io/etcd/actions/runs/3553267055/jobs/5968689132

ahrtr:
cc @spzala and @ptabor as well.

serathius (approved):

ahrtr:
The Release failure is because this PR isn't merged yet. Once this PR is merged, it can be resolved automatically.

serathius:
@ahrtr That's annoying, have you looked if this can be fixed?

ahrtr:
Yes, it's possible, and it should a minor fix. But it's a tradeoff. 

Currently the release.sh always pull the repo and checkout the `main` branch. So any change in curernt PR isn't included. It's safer because we can kick off the release anywhere and in any branch. 

The other option is to always kick off release process in current branch. So it requires us to manually checkout main branch before triggering the release process. 

It seems that we should follow the second approach?

ahrtr:
The better choice seems to be adding a flag to control it. We should enable the flag to trigger release in current branch in github workflow. 

When we manually trigger release, follow the existing logic.

PR #14860, opened 2022-11-26
Trigger release in current branch for github workflow case

ahrtr:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,4 +22,4 @@ jobs:
         Name-Email: github-action@etcd.io
         Expire-Date: 0
         EOF
-        DRY_RUN=true BRANCH=main ./scripts/release.sh --no-upload --no-docker-push 3.6.99
+        DRY_RUN=true ./scripts/release.sh --no-upload --no-docker-push --in-place 3.6.99
--- a/scripts/build-binary.sh
+++ b/scripts/build-binary.sh
@@ -7,7 +7,6 @@ source ./scripts/test_lib.sh
 VER=$1
 REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd.git}"
 
-
 if [ -z "$1" ]; then
   echo "Usage: ${0} VERSION" >> /dev/stderr
   exit 255
@@ -25,9 +24,7 @@ function setup_env {
 
   pushd "${proj}" >/dev/null
     run git fetch --all
-    git_assert_branch_in_sync || exit 2
     run git checkout "${ver}"
-    git_assert_branch_in_sync || exit 2
   popd >/dev/null
 }
 
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -41,6 +41,7 @@ help() {
   echo "  flags:"
   echo "    --no-upload: skip gs://etcd binary artifact uploads."
   echo "    --no-docker-push: skip docker image pushes."
+  echo "    --in-place: build binaries using current branch."
   echo ""
   echo "One can perform a (dry-run) test release from any (uncommitted) branch using:"
   echo "  DRY_RUN=true REPOSITORY=\`pwd\` BRANCH='local-branch-name' ./scripts/release 3.5.0-foobar.2"
@@ -54,8 +55,15 @@ main() {
   fi
   RELEASE_VERSION="v${VERSION}"
   MINOR_VERSION=$(echo "${VERSION}" | cut -d. -f 1-2)
-  BRANCH=${BRANCH:-"release-${MINOR_VERSION}"}
-  REPOSITORY=${REPOSITORY:-"https://github.com/etcd-io/etcd.git"}
+
+  if [ "${IN_PLACE}" == 1 ]; then
+      # Trigger release in current branch
+      REPOSITORY=$(pwd)
+      BRANCH=$(git rev-parse --abbrev-ref HEAD)
+  else
+      REPOSITORY=${REPOSITORY:-"https://github.com/etcd-io/etcd.git"}
+      BRANCH=${BRANCH:-"release-${MINOR_VERSION}"}
+  fi
 
   log_warning "DRY_RUN=${DRY_RUN}"
   log_callout "RELEASE_VERSION=${RELEASE_VERSION}"
@@ -78,18 +86,19 @@ main() {
   # Set up release directory.
   local reldir="/tmp/etcd-release-${VERSION}"
   log_callout "Preparing temporary directory: ${reldir}"
-  if [ ! -d "${reldir}/etcd" ]; then
+  if [ ! -d "${reldir}/etcd" ] && [ "${IN_PLACE}" == 0 ]; then
     mkdir -p "${reldir}"
     cd "${reldir}"
     run git clone "${REPOSITORY}" --branch "${BRANCH}"
+    run cd "${reldir}/etcd" || exit 2
+    run git checkout "${BRANCH}" || exit 2
+    run git pull origin
+
+    git_assert_branch_in_sync || exit 2
   fi
-  run cd "${reldir}/etcd" || exit 2
+
   # mark local directory as root for test_lib scripts executions
   set_root_dir
-
-  run git checkout "${BRANCH}" || exit 2
-  run git pull origin
-  git_assert_branch_in_sync || exit 2
 
   # If a release version tag already exists, use it.
   local remote_tag_exists
@@ -101,6 +110,7 @@ main() {
   fi
 
   # Check go version.
+  log_callout "Check go version"
   local go_version current_go_version
   go_version="go$(grep go-version .github/workflows/build.yaml | awk '{print $2}' | tr -d '"')"
   current_go_version=$(go version | awk '{ print $3 }')
@@ -110,6 +120,7 @@ main() {
   fi
 
   # If the release tag does not already exist remotely, create it.
+  log_callout "Create tag if not present"
   if [ "${remote_tag_exists}" -eq 0 ]; then
     # Bump version/version.go to release version.
     local source_version
@@ -148,7 +159,7 @@ main() {
     fi
 
     # Push the version change if it's not already been pushed.
-    if [ "$DRY_RUN" != "true" ] && [ "$(git rev-list --count "origin/${BRANCH}..${BRANCH}")" -gt 0 ]; then
+    if [ "${DRY_RUN}" != "true" ] && [ "$(git rev-list --count "origin/${BRANCH}..${BRANCH}")" -gt 0 ]; then
       read -p "Push version bump up to ${VERSION} to '$(git remote get-url origin)' [y/N]? " -r confirm
       [[ "${confirm,,}" == "y" ]] || exit 1
       maybe_run git push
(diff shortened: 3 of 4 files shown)


serathius:
I think you also need to override the repository name

ahrtr:
cc @serathius

scripts/build-release.sh
```diff
@@ -8,6 +8,7 @@ set -e
 source ./scripts/test_lib.sh
 
 VERSION=$1
+CURRENT_BRANCH=${CURRENT_BRANCH:-0}
```
ptabor, line 11: Please add a comment what values does the parameter takes. 
It's not obvious that it takes 0 & 1 and what 1 means.

scripts/release.sh
```diff
@@ -54,8 +54,16 @@ main() {
...
-  REPOSITORY=${REPOSITORY:-"https://github.com/etcd-io/etcd.git"}
+
+
+  if [ "${CURRENT_BRANCH}" == 1 ]; then
+      # Trigger release in current branch, used in github workflow
+      REPOSITORY=$(pwd)
+      BRANCH=$(git branch)
+  else
+      REPOSITORY=${REPOSITORY:-"https://github.com/etcd-io/etcd.git"}
+      BRANCH=${BRANCH:-"release-${MINOR_VERSION}"}
```
ptabor, line 65: Maybe instead of introduction of new variable 'CURRENT_BRANCH' we can add support for 
`BRANCH="#CURRENT"` OR `BRANCH="."` magic value. This would be IMHO cleaner about interactions with other variables.

scripts/release.sh
```diff
@@ -193,7 +213,7 @@ main() {
     log_warning "Skipping release build step. /release directory already exists."
   else
     log_callout "Building release..."
-    REPOSITORY=$(pwd) ./scripts/build-release.sh "${RELEASE_VERSION}"
+    REPOSITORY=$(pwd) CURRENT_BRANCH=${CURRENT_BRANCH} ./scripts/build-release.sh "${RELEASE_VERSION}"
```
ptabor, line 216: We just tagged locally release and we want to build it. I don't think passing here CURRENT_BRANCH=1 would do guarantee building the just tagged branch. Please double check.

ahrtr: When CURRENT_BRANCH is 1, build_binary.sh will not clone the repo, and checkout the branch. Instead, it just builds the binaries from local current directory.

scripts/build-release.sh
ahrtr: Since build-release.sh just transparently passes the parameter to build-binary.sh, so I added comment in build-binary.sh like below,
```

# When ${CURRENT_BRANCH} is 0, the scripts doesn't clone the remote
# repository at all, nor checkout the branch; instead, it just builds
# the binaries in current branch in local existing directory.
# Otherwise if ${CURRENT_BRANCH} is 1, it clones the remote repository
# firstly, then checkout the target tagged version, and finally build
# binaries.
```

scripts/release.sh
ahrtr: `CURRENT_BRANCH` means current branch, so no `clone`, no `checkout`.

scripts/release.sh
ahrtr: It seems that the the approach `BRANCH="#CURRENT"` complicates the implementation, because we need to evaluate `[ "${BRANCH}" != "#CURRENT" ] ` in all places where `BRANCH` is used. Probably I do not get your point.  So I'd like to keep it as it's.

ahrtr: So introducing a new variable `CURRENT_BRANCH` instead of updating the existing `BRANCH` is simpler.

scripts/release.sh
```diff
@@ -54,8 +54,16 @@ main() {
   fi
   RELEASE_VERSION="v${VERSION}"
   MINOR_VERSION=$(echo "${VERSION}" | cut -d. -f 1-2)
-  BRANCH=${BRANCH:-"release-${MINOR_VERSION}"}
-  REPOSITORY=${REPOSITORY:-"https://github.com/etcd-io/etcd.git"}
+
```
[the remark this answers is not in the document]
ahrtr: Added document for the new flag `--current-branch`.

scripts/build-binary.sh
```diff
@@ -6,7 +6,13 @@ source ./scripts/test_lib.sh
 
 VER=$1
 REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd.git}"
-
+# When ${CURRENT_BRANCH} is 0, the scripts doesn't clone the remote
```
fuweid, line 9: it should be `1`? if the `CURRENT_BRANCH` has been set, should we log the `VER` will be ignored?

ahrtr: Good catch. updated.

ahrtr:
ping @serathius @ptabor

.github/workflows/release.yaml
```diff
@@ -21,4 +21,4 @@ jobs:
         Name-Email: github-action@etcd.io
         Expire-Date: 0
         EOF
-        DRY_RUN=true BRANCH=main ./scripts/release.sh --no-upload --no-docker-push 3.6.99
+        DRY_RUN=true ./scripts/release.sh --no-upload --no-docker-push --current-branch 3.6.99
```
serathius, line 24: This sounds more like a `--local` flag.

scripts/build-binary.sh
```diff
@@ -6,7 +6,14 @@ source ./scripts/test_lib.sh
...
 VER=$1
 REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd.git}"
-
+# When ${CURRENT_BRANCH} is 1, the scripts doesn't clone the remote
+# repository at all, nor checkout the branch; instead, it just builds
+# the binaries in current branch in local existing directory. The ${VER}
+# will be ignored as well in such case.
+# Otherwise if ${CURRENT_BRANCH} is 0, it clones the remote repository
+# firstly, then checkout the target tagged version, and finally build
+# binaries.
```
serathius, line 15: Explaining the flag is just a consequence of a bad name of variable. Please prefer more correct naming scheme. Like in above comment this feature sounds more like `--local`

scripts/release.sh
```diff
@@ -54,8 +55,16 @@ main() {
...
+
+  if [ "${CURRENT_BRANCH}" == 1 ]; then
+      # Trigger release in current branch, used in github workflow
+      REPOSITORY=$(pwd)
+      BRANCH=$(git branch)
```
serathius, line 63: What do you expect `$(git branch)` to return? local git repository doesn't need to be on any branch.

scripts/release.sh
```diff
@@ -316,6 +338,10 @@ while test $# -gt 0; do
             help
             exit 0
             ;;
+          --current-branch)
```
serathius, line 341: Should we have a validation that prevents using incompatible flags/environment variables?

For example 
* `--current-branch` should only be called with `DRY_RUN=true`.
* `BRANCH` should be forbidden when `--current-branch` is set
* `REPO` should be forbidden when `--current-branch` is set

scripts/release.sh
ahrtr: Note that we need to print the branch on which the release.sh run, 

```
log_callout "BRANCH=${BRANCH}"
```

ahrtr: That's all for --current-branch mode

scripts/release.sh
ahrtr: > --current-branch should only be called with DRY_RUN=true

They are two different concepts. `DRY_RUN` means no push commit & image and no upload. --current-branch means no clone and no checkout. 

> BRANCH should be forbidden when --current-branch is set

Already explained above.

> REPO should be forbidden when --current-branch is set

The same as BRANCH, we need to output the repo in log.

.github/workflows/release.yaml
ahrtr: This seems not a big deal, I'd like to keep it as it's. `--current-branch ` seems a little better.  A current branch isn't necessary a local branch. github runner already cloned the repo and checked out the branch, we just kick off the release process in current branch which the github runner already checked out.

scripts/build-binary.sh
ahrtr: see above.

scripts/release.sh
ahrtr: FYI. in github workflow, the `$(git branch)` returns something like 

```
* (HEAD detached at pull/14860/merge)
```

.github/workflows/release.yaml
serathius: I'm against writing scripts that work only for github runner. As a maintainer I want to be able to run the same script locally to be able to debug the release script. First phase of debugging release script should not be figuring out differences between local and github runner setup. 

Please don't put any assumptions about running in github runner.

scripts/release.sh
ahrtr: > You should not be allowed to upload artifacts from local changes.

This is a valid point. thx & updated.

scripts/release.sh
serathius: `git branch` will return list of all branches. On github workflow I expect it works because there is only one branch available. This code will break or even be misleading (may pick first branch from alphabetically sorted names) when running locally.

ptabor: Good catch: 
`git symbolic-ref --short HEAD` should work.

ahrtr: Thanks for pointing out this. I need more sleep again :)

.github/workflows/release.yaml
ptabor: To understand both options: 
  1. `--local` would mean: do not clone, keep using 'release-3.6.46'  branch.  
  2. `--current_branch` overrides both BRANCH name and forces usage of local repository. 

So the question is whether there is a valid use-case when someone has a git-cloned repository and want's to build an already tagged release. And to me it makes sense - for a case someone does not trust etcd-team provided binaries. 

So the less interactions between flags the better. 

To accommodate the github usecase the invocation would look like: 

```
BRANCH=$(git symbolic-ref --short HEAD) DRY_RUN=true ./scripts/release.sh --no-upload --no-docker-push --local 3.6.99
```

I would also consider `--in-place` flag, as `--local` might suggests there is an option to perform the build on a remote machine.

scripts/release.sh
ahrtr: Updated.

ahrtr:
The reason why I submitted this PR is there is flaw in current release script. Each time triggering a release process, the `scripts/release.sh` in current PR/branch is executed, but all other scripts and go src files are coming from `main` or a stable release branch (e.g. `release-3.5`). Obviously this isn't correct. Each time when we fix anything in the release scripts other than the `release.sh` itself, the changes can never be covered by current PR's pipeline. 

I split the release workflow into two, 
- One is triggered on pull request. We should always get everything from current branch in this case. 
- The other is triggered on push to `main`, we keep following the existing behavior.

Please let me know if you have any comments or concerns. @ptabor @serathius @spzala

.github/workflows/release.yaml
ahrtr: Updated.

ahrtr: thx

ahrtr:
Marked the PR ready for review.

ahrtr:
Just rebased an squshed the PR.

ahrtr:
cc @ptabor @serathius @spzala PTAL, thx

serathius (approved):

PR #14865, opened 2022-11-28
test: fix unit test Instability

ahrtr:
When two members in a 5 member cluster are corrupted, and they have different hashes, etcd will raise alarm for both members, but the order isn't guaranteed. But if the two corrupted members have the same hash, then the order is guaranteed. The leader always raise alarm in the same order as the member list.

Refer to: https://github.com/etcd-io/etcd/actions/runs/3569072873/jobs/5998638380 

cc @ptabor @serathius @spzala


fuweid (approved):
LGTM(non-binding)

serathius (approved):

PR #14872, opened 2022-11-30
Make sure shellcheck exist

### LINKED ISSUE #14869: `make verify-shellcheck` only warns, doesn't fail when tool doesn't exist
### What happened?

Make target `verify-shellcheck` only warns and doesn't fail in absence of shellcheck binary.

When `shellcheck` is missing in local, `verify-shellcheck` should fail.

1. Uninstall (or unset path to) `shellcheck`.
2. Run `make verify-shellcheck`.
    It gives warning:
    ```console
    Tool: 'shellcheck' not found on PATH. https://github.com/koalaman/shellcheck#installing
    ```
    But doesn't fail, due to which the [test.sh][1] marks it as `SUCCESS`.

[1]: https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L700

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.6.0-alpha.0
Git SHA: 61e9ec983
Go Version: go1.19.3
Go OS/Arch: linux/amd64

$ etcdctl version
API version: 3.6
```

</details>

_No response_

_No response_

### Relevant log output

```shell
$ which shellcheck
shellcheck not found

$ make verify-shellcheck             
PASSES="shellcheck" ./scripts/test.sh
Running with --race
Starting at: Tuesday 29 November 2022 11:01:36 PM IST

'shellcheck' started at Tuesday 29 November 2022 11:01:36 PM IST
Tool: 'shellcheck' not found on PATH. https://github.com/koalaman/shellcheck#installing
'shellcheck' completed at Tuesday 29 November 2022 11:01:36 PM IST
SUCCESS
```

HeavenTonight:
Signed-off-by: guiyong.ou <guiyong.ou@daocloud.io>

Fixed #14869 


Bhargav-InfraCloud:
I don't think `staticcheck` will cover for `shellcheck`. Need to check for shellcheck installation or an alternative.

ahrtr:
Thanks @Bhargav-InfraCloud for pointing this out. I mis-read the `shellcheck` as `staticcheck`. It seems that the `shellcheck`  has never worked in the github workflow? @serathius @ptabor

ahrtr:
Based on the guide https://github.com/koalaman/shellcheck#installing, it seems that we should install the tool using command `sudo apt install shellcheck`, but we should install a specified version so as to make the workflow stable. WDYT? @serathius @ptabor

HeavenTonight:
`.PHONY: install-shellcheck
install-shellcheck:
	if ! command -v shellcheck > /dev/null 2>&1; then
	    sudo apt install shellcheck
	fi`

like this,check first,is that ok? @ahrtr

ahrtr:
We need to install a specific version of `shellcheck`.  `sudo apt install shellcheck` somehow can only specify version `0.3.7-5`.  Probably we can download the tool directly something like below,

```
VERSION=v0.8.0

wget https://github.com/koalaman/shellcheck/releases/download/${VERSION}/shellcheck-v0.8.0.linux.x86_64.tar.xz
```

Makefile
```diff
@@ -70,9 +71,16 @@ fix-lint:
...
 .PHONY: verify-shellcheck
-verify-shellcheck:
+verify-shellcheck: install-shellcheck
 	PASSES="shellcheck" ./scripts/test.sh
 
+.PHONY: install-shellcheck
+install-shellcheck:
+	if ! command -v shellcheck > /dev/null 2>&1; then \
+        wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar -xJv && \
+        cp "shellcheck-${SHELLCHECK_VERSION}/shellcheck" /usr/bin/; \
```
serathius, line 81: copying to `/usr/bin/` is not a great idea as it requires root permissions.

HeavenTonight: /usr/local/bin/ ? What do you think?  Maybe have better way @serathius

ahrtr: Somehow it passed. No any error in the workflow. 

```
if ! command -v shellcheck > /dev/null 2>&1; then \
        wget -qO- "https://github.com/koalaman/shellcheck/releases/download/v0.8.0/shellcheck-v0.8.0.linux.x86_64.tar.xz" | tar -xJv && \
SUCCESS
        cp "shellcheck-v0.8.0/shellcheck" /usr/bin/; \
fi
PASSES="shellcheck" ./scripts/test.sh
Running with --race
Starting at: Wed Nov 30 21:41:14 UTC 2022

'shellcheck' started at Wed Nov 30 21:41:14 UTC 2022
% 'shellcheck' '-fgcc' 'scripts/build-binary.sh' 'scripts/build-docker.sh' 'scripts/build-release.sh' 'scripts/build.sh' 'scripts/build_lib.sh' 'scripts/build_tools.sh' 'scripts/codecov_upload.sh' 'scripts/fix.sh' 'scripts/fuzzing.sh' 'scripts/genproto.sh' 'scripts/install-marker.sh' 'scripts/measure-test-flakiness.sh' 'scripts/release.sh' 'scripts/release_mod.sh' 'scripts/test.sh' 'scripts/test_lib.sh' 'scripts/update_dep.sh' 'scripts/update_proto_annotations.sh' 'scripts/updatebom.sh' 'scripts/verify_genproto.sh' 'scripts/verify_proto_annotations.sh'
'shellcheck' completed at Wed Nov 30 21:41:19 UTC 2022
SUCCESS

```

But the better way seems to be something like below. Could you try this out? 
```
mkdir -p $HOME/.local/bin
cp "shellcheck-${SHELLCHECK_VERSION}/shellcheck" $HOME/.local/bin


```

Makefile
```diff
@@ -70,9 +71,17 @@ fix-lint:
...
 
 .PHONY: verify-shellcheck
-verify-shellcheck:
+verify-shellcheck: install-shellcheck
 	PASSES="shellcheck" ./scripts/test.sh
 
+.PHONY: install-shellcheck
+install-shellcheck:
+	if ! command -v shellcheck > /dev/null 2>&1; then \
+        wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar -xJv; \
```
lowryxiao, line 80: Just out of curious, do we need to conside other os distro and arch? like darwin and x86?

Bhargav-InfraCloud: Currently, etcd dev supports only **linux-amd64**. Check [Setup development environment][1].

[1]: https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#setup-development-environment

lowryxiao:
My cents here is if it's better to add check and download shellcheck logic here if it doesn't exist? that we needn't check twice.
https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L393
Since the original error comes from the function tool_exists:
https://github.com/etcd-io/etcd/blob/main/scripts/test_lib.sh#L351

As what I can image is like this:
```
function shellcheck_pass {
  # make sure shellcheck is there, if not, install it first
  if ! tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
    # if shellcheck is not there, we need to install it first
  fi
  # run shellcheck
  generic_checker run shellcheck -fgcc scripts/*.sh
  # return $? of above command
  return $?
}
```
It's better to return corresponding error code of shellcheck to the caller then the caller will take care if there error happens.

Makefile
serathius: Two notes, it's bad to touch user home directory and `$HOME/.local/bin` is not guaranteed to be in `$PATH` environment variable especially when it doesn't exist. 

My recommendation is to put shellcheck in `./bin/tools` and don't assume that it is available in `PATH`

ahrtr: FYI. https://github.com/actions/runner-images/pull/2491 

Another option is https://github.com/etcd-io/etcd/pull/14872#issuecomment-1333312588

serathius: Runner != Developer

scripts/test.sh
```diff
@@ -390,9 +391,13 @@ function cov_pass {
 ######### Code formatting checkers #############################################
 
 function shellcheck_pass {
-  if tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
-    generic_checker run shellcheck -fgcc scripts/*.sh
+  if ! tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
+    wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar -xJv
+    mkdir -p ${HOME}/.local/bin
```
ptabor, line 396: I think it's a bad pattern when some innocent test script is playing with operation-system wide or user-wide settings / layout and can substitute scripts on the PATH with vulnerable code. 

As a risk we could imagine here another script that expect different version of shellcheck and them racing about the content of ${HOME}/.local/bin. 

I think everything that ./test.sh script is doing should be limited to the: 
  - `./etcd` directory
  - unique `/tmp` subdirectory that should be cleaned afterwards
  - golang-level module caches that golang takes responsibility for versioning and race-protection. 

Although this code would work in github workflows, I think it's not recommended for developer's environment. 

I would go in the direction of @serathius proposal. 
Maybe even `./etcd/.tmp/bin/shellcheck` and making sure .tmp is on the `.gitignore`.

ahrtr:
I agree that we shouldn't change anything of system-wide or user-wide settings. The proposal https://github.com/etcd-io/etcd/pull/14872#issuecomment-1333312588 seems like a valid solution to me. 

If the command `shellcheck` already exists, then use it directly, and set `SHELLCHECK="shellcheck"`. Otherwise, download and install it in `./bin`, and set `SHELLCHECK="./bin/shellcheck"`. Eventually execute command ${SHELLCHECK} something like below,

```
generic_checker run ${SHELLCHECK}  -fgcc scripts/*.sh
```

scripts/test.sh
```diff
@@ -388,9 +389,13 @@ function cov_pass {
...
 function shellcheck_pass {
-  if tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
-    generic_checker run shellcheck -fgcc scripts/*.sh
+  SHELLCHECK=shellcheck
+  if ! tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
+    wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar -xJv
+    SHELLCHECK=$(pwd)/bin/shellcheck
   fi
+  generic_checker run ${SHELLCHECK} -fgcc scripts/*.sh
+  return $?
```
ahrtr, line 398:
```suggestion
  if ! tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
    log_callout "Installing shellcheck $SHELLCHECK_VERSION"
    wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar -xJv -C /tmp/ --strip-components=1
    mkdir -p ./bin
    mv /tmp/shellcheck ./bin/
    SHELLCHECK=./bin/shellcheck
  fi
  generic_checker run ${SHELLCHECK} -fgcc scripts/*.sh
```

ahrtr:
Please also add an item `rm -rf ./bin/shellcheck*` into the target `clean` in `Makefile`.

ahrtr (approved):
LGTM

Thank you @HeavenTonight

serathius (approved):
Looks reasonable. Personally I would prefer we simplify dev scripting enough to migrate most commands to Makefile.

PR #14875, opened 2022-11-30
lease: fix potential goroutine leak in lessor_test.go

BarrensZeppelin:
This PR fixes a potential goroutine leak in `TestLessorRenew` and `TestLessorRenewWithCheckpointer`.

In the unlikely event that the test lease expires before `Renew` is called, the call to `Renew` enters the select statement at [server/lease/lessor.go#L412-L424](https://github.com/etcd-io/etcd/blob/main/server/lease/lessor.go#L412-L424) and remains blocked indefinitely. This is reproducible by inserting an appropriate call to `time.Sleep` before the call to `Renew`.

The fix is to move the potentially blocking call to a separate goroutine. If waiting for the call times out, the test fails and the deferred `(*lessor).Stop` method is called which will unblock the call to `Renew`.

### DIFF
--- a/server/lease/lessor_test.go
+++ b/server/lease/lessor_test.go
@@ -207,6 +207,29 @@ func TestLessorRevoke(t *testing.T) {
 	}
 }
 
+func renew(t *testing.T, le *lessor, id LeaseID) int64 {
+	ch := make(chan int64, 1)
+	errch := make(chan error, 1)
+	go func() {
+		ttl, err := le.Renew(id)
+		if err != nil {
+			errch <- err
+		} else {
+			ch <- ttl
+		}
+	}()
+
+	select {
+	case ttl := <-ch:
+		return ttl
+	case err := <-errch:
+		t.Fatalf("failed to renew lease (%v)", err)
+	case <-time.After(2 * time.Second):
+		t.Fatal("timed out while renewing lease")
+	}
+	panic("unreachable")
+}
+
 // TestLessorRenew ensures Lessor can renew an existing lease.
 func TestLessorRenew(t *testing.T) {
 	lg := zap.NewNop()
@@ -227,10 +250,7 @@ func TestLessorRenew(t *testing.T) {
 	le.mu.Lock()
 	l.ttl = 10
 	le.mu.Unlock()
-	ttl, err := le.Renew(l.ID)
-	if err != nil {
-		t.Fatalf("failed to renew lease (%v)", err)
-	}
+	ttl := renew(t, le, l.ID)
 	if ttl != l.ttl {
 		t.Errorf("ttl = %d, want %d", ttl, l.ttl)
 	}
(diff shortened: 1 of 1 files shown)


server/lease/lessor_test.go
```diff
@@ -207,6 +207,29 @@ func TestLessorRevoke(t *testing.T) {
...
+	select {
+	case ttl := <-ch:
+		return ttl
+	case err := <-errch:
+		t.Fatalf("failed to renew lease (%v)", err)
+	case <-time.After(2 * time.Second):
+		t.Fatal("timed out while renewing lease")
+	}
+	panic("unreachable")
+}
```
ahrtr, line 231: I am thinking [testify](https://github.com/stretchr/testify) should support this, but unfortunately I do not see such function.  The similar one is [Eventually](https://github.com/stretchr/testify/blob/1333b5d3bda8cf5aedcf3e1aaa95cac28aaab892/require/require.go#L357), but it can't satisfy this case. Probably you can deliver a PR for [testify](https://github.com/stretchr/testify) to support it, and enhance the test case in future.

ahrtr (approved):
LGTM

Thank you @BarrensZeppelin

server/lease/lessor_test.go
```diff
@@ -207,6 +207,29 @@ func TestLessorRevoke(t *testing.T) {
...
+			ch <- ttl
+		}
+	}()
+
+	select {
+	case ttl := <-ch:
+		return ttl
+	case err := <-errch:
+		t.Fatalf("failed to renew lease (%v)", err)
+	case <-time.After(2 * time.Second):
```
fuweid, line 227: Would you mind to increase the number into 30s or 1m? since the CI instance has limitted resource, it might be flaky case. The large number won't hurt the case if it is supposed to be fast. WDYT?

ahrtr: It's just a unit test case, and there isn't too many lease in both cases, 30s timeout seems to be overkill. 2s should be OK. But it isn't a big deal, if you want, any value between 2 ~ 10s is OK to me.

BarrensZeppelin: I increase the timeout to 10s. :slightly_smiling_face:

ptabor:
> In the unlikely event that the test lease expires before `Renew` is called, the call to `Renew` enters the select statement at [server/lease/lessor.go#L412-L424](https://github.com/etcd-io/etcd/blob/main/server/lease/lessor.go#L412-L424) and remains blocked indefinitely. 

Thank you for finding this. This sounds like a clear violation of the method's contract:

https://github.com/etcd-io/etcd/blob/40cc73c22eff232534602031f0db6fa9bb918a75/server/lease/lessor.go#L390-L391

Could you, please report an issue on this problem alone... , add a deterministic repro, and point in the error message on the newly reported issue.

ahrtr:
> Thank you for finding this. This sounds like a clear violation of the method's contract:
> 
> https://github.com/etcd-io/etcd/blob/40cc73c22eff232534602031f0db6fa9bb918a75/server/lease/lessor.go#L390-L391
> 
> Could you, please report an issue on this problem alone... , add a deterministic repro, and point in the error message on the newly reported issue.
> 

It isn't a problem in production environment, because expired lease will be revoked. But this is a unit test case, there is no goroutine to revoke the lease.

ahrtr:
@ptabor Do you have any other comment on this PR?

ptabor (approved):

PR #14880, opened 2022-12-01
Improve support for failed requests in linearizability tests

serathius:
Based on recommendations in porcupine issues, failed requests should be marked as they failed at the end of test. This allows to stop aggregating errors within model making it:
* Much simpler to implement. Handling all the edge cases turned out to be very hard as we added new methods like delete.
* Much more accurate. Now we can validate revision being continuous as failed request cannot just pop up.

Only downside is history visualization is much less readable.

### DIFF
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -25,9 +25,13 @@ import (
 
 type recordingClient struct {
 	client clientv3.Client
-	id     int
+
+	// id of the next write operation. If needed a new id might be requested from idProvider.
+	id         int
+	idProvider idProvider
 
 	operations []porcupine.Operation
+	failed     []porcupine.Operation
 }
 
 func NewClient(endpoints []string, ids idProvider) (*recordingClient, error) {
@@ -43,7 +47,9 @@ func NewClient(endpoints []string, ids idProvider) (*recordingClient, error) {
 	return &recordingClient{
 		client:     *cc,
 		id:         ids.ClientId(),
+		idProvider: ids,
 		operations: []porcupine.Operation{},
+		failed:     []porcupine.Operation{},
 	}, nil
 }
 
@@ -76,6 +82,19 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
 	callTime := time.Now()
 	resp, err := c.client.Put(ctx, key, value)
 	returnTime := time.Now()
+	if err != nil {
+		c.failed = append(c.failed, porcupine.Operation{
+			ClientId: c.id,
+			Input:    etcdRequest{op: Put, key: key, putData: value},
+			Call:     callTime.UnixNano(),
+			Output:   etcdResponse{err: err},
+			Return:   0, // For failed writes we don't know when request has really finished.
+		})
+		// Operations of single client needs to be sequential.
+		// As we don't know return time of failed operations, all new writes need to be done with new client id.
+		c.id = c.idProvider.ClientId()
+		return err
+	}
 	var revision int64
 	if resp != nil && resp.Header != nil {
 		revision = resp.Header.Revision
@@ -94,6 +113,19 @@ func (c *recordingClient) Delete(ctx context.Context, key string) error {
 	callTime := time.Now()
 	resp, err := c.client.Delete(ctx, key)
 	returnTime := time.Now()
+	if err != nil {
+		c.failed = append(c.failed, porcupine.Operation{
+			ClientId: c.id,
+			Input:    etcdRequest{op: Delete, key: key},
+			Call:     callTime.UnixNano(),
+			Output:   etcdResponse{err: err},
+			Return:   0, // For failed writes we don't know when request has really finished.
+		})
+		// Operations of single client needs to be sequential.
+		// As we don't know return time of failed operations, all new writes need to be done with new client id.
+		c.id = c.idProvider.ClientId()
+		return err
+	}
 	var revision int64
 	var deleted int64
 	if resp != nil && resp.Header != nil {
@@ -108,4 +140,23 @@ func (c *recordingClient) Delete(ctx context.Context, key string) error {
 		Return:   returnTime.UnixNano(),
 	})
 	return nil
+}
+
+func (c *recordingClient) Operations() []porcupine.Operation {
+	operations := make([]porcupine.Operation, 0, len(c.operations)+len(c.failed))
+	var maxTime int64
+	for _, op := range c.operations {
+		operations = append(operations, op)
+		if op.Return > maxTime {
+			maxTime = op.Return
+		}
+	}
+	for _, op := range c.failed {
+		if op.Call > maxTime {
+			continue
+		}
+		op.Return = maxTime + 1
+		operations = append(operations, op)
+	}
+	return operations
 }
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -70,9 +70,9 @@ func (c *recordingClient) Get(ctx context.Context, key string) error {
 	}
 	c.operations = append(c.operations, porcupine.Operation{
 		ClientId: c.id,
-		Input:    etcdRequest{op: Get, key: key},
+		Input:    EtcdRequest{Op: Get, Key: key},
 		Call:     callTime.UnixNano(),
-		Output:   etcdResponse{getData: readData, revision: resp.Header.Revision},
+		Output:   EtcdResponse{GetData: readData, Revision: resp.Header.Revision},
 		Return:   returnTime.UnixNano(),
 	})
 	return nil
@@ -85,9 +85,9 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
 	if err != nil {
 		c.failed = append(c.failed, porcupine.Operation{
 			ClientId: c.id,
-			Input:    etcdRequest{op: Put, key: key, putData: value},
+			Input:    EtcdRequest{Op: Put, Key: key, PutData: value},
 			Call:     callTime.UnixNano(),
-			Output:   etcdResponse{err: err},
+			Output:   EtcdResponse{Err: err},
 			Return:   0, // For failed writes we don't know when request has really finished.
 		})
 		// Operations of single client needs to be sequential.
@@ -101,9 +101,9 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
 	}
 	c.operations = append(c.operations, porcupine.Operation{
 		ClientId: c.id,
-		Input:    etcdRequest{op: Put, key: key, putData: value},
+		Input:    EtcdRequest{Op: Put, Key: key, PutData: value},
 		Call:     callTime.UnixNano(),
-		Output:   etcdResponse{err: err, revision: revision},
+		Output:   EtcdResponse{Err: err, Revision: revision},
 		Return:   returnTime.UnixNano(),
 	})
 	return nil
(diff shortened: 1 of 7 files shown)


serathius:
cc @ahrtr @geetasg @ptabor

serathius:
Did some testing and found out that new model is faster on linearizable history, but can be much much slower on non-linearizable history. I expect this is due fact that we have multiple failure injections in one history. This results in a lot of timeout requests which gives exponential cost. Before merging this PR we would need to rewrite testing to trigger only one failure per history.

tjungblu:
when you want to implement multiple keys at some point, don't forget to implement the partition function:
https://github.com/anishathalye/porcupine/blob/master/porcupine_test.go#L1142-L1158

porcupine can parallelize over each partition, making it even quicker.

serathius:
> when you want to implement multiple keys at some point, don't forget to implement the partition function:

This is a good suggestion. Based on different usage patters I was considering to add separate key when testing leases. cc @geetasg

serathius:
Ok, fixed issue with linearizability verification going into exponential time. Solution was to shorten test runs to only one failpoint injections. We still do 60 iterations, however each of them recreates the cluster. 

This is a little costly for 3 node cluster, but it was a quickest way to shorten request history and guarantee correctness. We might consider reusing cluster in future.

serathius:
re-running linearizability tests to validate for flakes.

serathius:
Recreating cluster on every run makes tests unstable. It is required for this change to work, so I decided to separate it to https://github.com/etcd-io/etcd/pull/14885 to be able to iterate on it quicker.

tests/linearizability/client.go
```diff
@@ -76,16 +84,45 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
...
 	resp, err := c.client.Put(ctx, key, value)
 	returnTime := time.Now()
+	if err != nil {
+		c.failed = append(c.failed, porcupine.Operation{
+			ClientId: int(c.id),
+			Input:    EtcdRequest{Op: Put, Key: key, PutData: value},
+			Call:     callTime.UnixNano(),
+			Output:   EtcdResponse{Err: err},
+		})
+		c.id = clientId.Add(1) - 1
```
tjungblu, line 94: I assume this is to "create" a new client? a comment on why this is necessary would be quite useful for others to understand

tests/linearizability/client.go
```diff
@@ -76,16 +84,45 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
...
+	operations := make([]porcupine.Operation, 0, len(c.operations)+len(c.failed))
+	var maxTime int64
+	for _, op := range c.operations {
+		operations = append(operations, op)
+		if op.Return > maxTime {
+			maxTime = op.Return
+		}
+	}
+	for _, op := range c.failed {
+		if op.Call > maxTime {
```
tjungblu, line 121: can this really happen? not sure if it would be better to panic here when the clocks tick the other way and we have Returns finishing before Calls.

tests/linearizability/client.go
```diff
@@ -76,16 +84,45 @@ func (c *recordingClient) Put(ctx context.Context, key, value string) error {
...
+			continue
+		}
+		op.Return = maxTime + 1
```
tjungblu, line 124: this is a really nice trick :) please also leave a comment on why this is necessary and what it effectively does for others. I reckon this really makes the visualization unreadable?

tests/linearizability/client.go
tjungblu: also don't know if this is confusing when you configure `config.clientCount` in a test and suddenly you have 5 more due to a failpoint.

tests/linearizability/model.go
```diff
@@ -104,50 +105,70 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
-	if ok && state.LastRevision < response.revision {
-		state.Value = response.getData
-		state.LastRevision = response.revision
-		delete(state.FailedWrites, response.getData)
-		return true, state
+	if state.FailedWrite != nil && state.LastRevision < response.Revision && state.LastRevision+state.FailedCount >= response.Revision {
+		var ok bool
+		switch state.FailedWrite.Op {
+		case Get:
+			ok = false
```
tjungblu, line 140: just reading this for the first time and it totally confused me, maybe `FailedRequest` would be a better name if there can be Gets?

tests/linearizability/model.go
```diff
@@ -104,50 +105,70 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
+		state.FailedCount = 0
 		return true, state
 	}
-	_, ok := state.FailedWrites[response.getData]
-	if ok && state.LastRevision < response.revision {
-		state.Value = response.getData
-		state.LastRevision = response.revision
-		delete(state.FailedWrites, response.getData)
-		return true, state
+	if state.FailedWrite != nil && state.LastRevision < response.Revision && state.LastRevision+state.FailedCount >= response.Revision {
```
tjungblu, line 136: I'm wondering whether you can just keep an "expectedRevision" around instead of doing the mental math around failed requests and response revisions. Then this check and the one below would be fairly straight forward != checks.

tests/linearizability/client.go
serathius: Adding a comment that recordingClient id might change.

tests/linearizability/client.go
serathius: I found this recommendation in porcupine issues. Previous approach was to handle errors in model, but that made model super complex as it needed to handle possibility that a failed request might pop up much later in history.

tests/linearizability/model.go
serathius: Changed to panic.

tests/linearizability/model.go
serathius: Decided that this is overkill for now. I will create a separate PR that will make this more readable.

serathius:
Rewrite the PR based on comments from @tjungblu. Thanks for review!
Please take another look @ahrtr @tjungblu

tests/linearizability/model.go
```diff
@@ -104,50 +104,65 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
+		state.FailedWrite = nil
 		return true, state
 	}
-	_, ok := state.FailedWrites[response.getData]
-	if ok && state.LastRevision < response.revision {
-		state.Value = response.getData
-		state.LastRevision = response.revision
-		delete(state.FailedWrites, response.getData)
-		return true, state
+	if state.FailedWrite != nil && state.LastRevision < response.Revision {
```
ahrtr, line 133:
```suggestion
	if state.FailedWrite != nil && state.LastRevision <= response.Revision {
```

tests/linearizability/client.go
```diff
@@ -58,34 +54,14 @@ func (c *recordingClient) Get(ctx context.Context, key string) error {
...
-		readData = string(resp.Kvs[0].Value)
-	}
-	c.operations = append(c.operations, porcupine.Operation{
-		ClientId: c.id,
-		Input:    etcdRequest{op: Get, key: key},
-		Call:     callTime.UnixNano(),
-		Output:   etcdResponse{getData: readData, revision: resp.Header.Revision},
-		Return:   returnTime.UnixNano(),
-	})
+	c.history.AppendGet(key, callTime, returnTime, resp)
```
ahrtr, line 57: Nice encapsulation!

tests/linearizability/id.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (id *atomicProvider) ClientId() int {
+	// Substract one as ClientId should start from zero.
+	return int(id.clientId.Add(1) - 1)
+}
+
+func (id *atomicProvider) RequestId() int {
+	return int(id.requestId.Add(1))
+}
```
ahrtr, line 40: Suggest to rewrite the idProvider something like below. It shouldn't care about the business logic (e.g. `clientId`, `requestId`). 

```
type idProvider interface {
	NextId() int64
}

func newIdProvider(startFromZero bool) idProvider {
	idp := &atomicProvider{}
	if startFromZero {
		idp.nextId.Add(-1)
	}
	return idp
}

type atomicProvider struct {
	startFromZero bool
	nextId        atomic.Int64
}

func (idp *atomicProvider) NextId() int64 {
	return idp.nextId.Add(1)
}
```

tests/linearizability/history.go
```diff
@@ -0,0 +1,119 @@
...
+			maxTime = op.Return
+		}
+	}
+	for _, op := range h.failed {
+		if op.Call > maxTime {
+			continue
+		}
+		op.Return = maxTime + 1
+		operations = append(operations, op)
+	}
```
ahrtr, line 117: If I understand correctly, you followed the suggestion in https://github.com/anishathalye/porcupine/issues/10? If yes, please add a comment here, otherwise it's hard for other contributors to understand.

tests/linearizability/model.go
```diff
@@ -104,50 +104,65 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
 		panic("Unknown operation")
 	}
 	return state
 }
 
-func stepGet(state EtcdState, request etcdRequest, response etcdResponse) (bool, EtcdState) {
-	if state.Value == response.getData && state.LastRevision <= response.revision {
+func stepGet(state EtcdState, request EtcdRequest, response EtcdResponse) (bool, EtcdState) {
+	if state.Value == response.GetData && state.LastRevision == response.Revision {
+		state.LastRevision = response.Revision
```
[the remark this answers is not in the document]
serathius: This is needed. I missed it in original implementation.

tests/linearizability/model.go
```diff
@@ -104,50 +104,65 @@ func step(state EtcdState, request etcdRequest, response etcdResponse) (bool, Et
...
+			ok = response.GetData == state.FailedWrite.PutData
+		default:
+			panic("Unknown operation")
+		}
+		if ok {
+			state.Value = response.GetData
+			state.LastRevision = response.Revision
+			state.FailedWrite = nil
+			return true, state
+		}
```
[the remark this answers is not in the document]
serathius: We are just adding Delete. Simplification here doesn't make sense as we know that we will add more operations. I think it's good thing to leave a switch here as it makes it obvious that new operations need to handle it.

tests/linearizability/model.go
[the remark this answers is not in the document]
ahrtr: When the `if` statement is true, then `state.LastRevision` must be equal to `response.Revision`, so no need to set `state.LastRevision` again.

tests/linearizability/id.go
ahrtr: It's OK to refactor this in a separate PR.

tests/linearizability/model.go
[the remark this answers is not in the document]
ahrtr: It's OK to update this in a separate PR if you want.

tests/linearizability/id.go
serathius: Proposed refactor doesn't benefit current code. It could help with code reuse, but we are not planning any reuse here. 

My main motivation to refactor this code to idProvider was to avoid writing to many comments explaining why ids behave as they do. By encapsulating logic we can make it easier to understand.

tests/linearizability/model.go
ahrtr: > writes always increase revision.

With the addition of delete operation, it doesn't stand anymore. Deleting an non-exist will not increase the revision.

ahrtr: https://github.com/etcd-io/etcd/pull/14802/files#diff-f6cd9cfbf4b5310cea9742f346927212c88f39b5cab928a434fe3ffb44d50cccR182-R187

tests/linearizability/history.go
serathius: Added the comment

tests/linearizability/model.go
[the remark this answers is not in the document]
serathius: Ups, you're right. Forgot that if statement changed in the mean time.

tests/linearizability/model.go
serathius: Yes, but that should be changed by PR that introduced delete.

serathius:
Rebased on delete changes.

Issue #14713, opened 2022-11-09
raft: decouple raft release from etcd release


ahrtr:
### What would you like to be added?

Currently the [raft](https://github.com/etcd-io/etcd/tree/main/raft) is released & tagged together with etcd. For example, when we release etcd 3.5.5, we will create tag `v3.5.5` for etcd and `raft/v3.5.5` for raft at the same time respectively. 

raft is an important & stable module, it may have different evolution pace from etcd. I think it would be better to qualify and release raft separately, just like what we did for [BoltDB](https://github.com/etcd-io/bbolt). 

We probably need to change the module name from `go.etcd.io/etcd/raft/v3` to `go.etcd.io/raft`. Since it's a new beginning, we can release raft starting from `v1`, which isn't needed in the module name. We need to add the version into the module name starting from `v2`, namely `go.etcd.io/raft/v2`.  FYI. [Module version numbering](https://go.dev/doc/modules/version-numbers)

We can even create a separate repository for raft, such as `github.com/etcd-io/raft`. But it isn't a blocker for this request, either way we will release raft separately. 

We need to agree on the release criteria of raft if we follow this approach. For the first step, I think we can keep doing what we already have,
1. The [datadriven test](https://github.com/etcd-io/etcd/tree/main/raft/rafttest);
2. etcd [workflows](https://github.com/etcd-io/etcd/tree/main/.github/workflows);
3. Kubernetes SIG Scalability test. (Note: @serathius maintains etcd version in Kubernetes)

Since [release-3.5](https://github.com/etcd-io/etcd/tree/release-3.5) is a stable release, so we will not change anything on 3.5 and [raft in 3.5](https://github.com/etcd-io/etcd/tree/release-3.5/raft). This change is only for the `main` branch. 

## So in summary, the request has three points:
1. release raft separately;
2. rename raft module name from `go.etcd.io/etcd/raft/v3` to `go.etcd.io/raft`.
3. (optional) create a separate repository (e.g github.com/etcd-io/raft) for raft. 

@hexfusion @mitake @ptabor @serathius @spzala @tbg @nvanbenschoten @pavelkalinnikov Please share your thoughts. thx. 

### Why is this needed?

see above.

serathius:
Moving raft to separate repo was already discussed and got early support in etcd maintainer mailing list. Overall it make sense for me as it will allow raft maintainers to implement their changes without being constrained by etcd. 

[...]

PR #14881, opened 2022-12-02
Update etcd to use the new raft module go.etcd.io/raft/v3

ahrtr:
Update etcd to use the new raft module `go.etcd.io/raft/v3`. Currently we depend on the latest commit of [raft](https://github.com/etcd-io/raft). When we  release new tagged raft version, we can bump etcd by then.

The only left work is to update [modules.svg](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/modules.svg).  Of course, we also need to update the changelog. Both tasks will be done in separate PR(s).

Please note that the [etcd/raft](https://github.com/etcd-io/etcd/tree/main/raft) isn't removed, instead it is just marked as Deprecated for now. We can remove it in 3.7.


ahrtr:
cc @mitake @ptabor @serathius @spzala @tbg PTAL.

ptabor (approved):

serathius:
Not sure about deprecation and removal in v3.7. I'm worried that wait doesn't have any benefits but could cause a lot of confusion for etcd contributors. Can be discussed in separate PR.

serathius (approved):

ahrtr:
Thanks both @ptabor and @serathius for the quick review.

> I'm worried that wait doesn't have any benefits but could cause a lot of confusion for etcd contributors. Can be discussed in separate PR.

The only concern is that some applications might already depend on the raft on main, such as cockroachdb db (@tbg ). So we can't directly remove it. It's safer to remove it in 3.7.  

Of course, we can discuss it separately.

ahrtr:
> Based on how much technical depth is in etcd repo, I would prefer to pay it off earlier than later.

Personally I'd like to remove it soon. But at least, let's send formal notification and wait some time, e.g. 1 month?

serathius:
> The only concern is that some applications might already depend on the raft on main, such as cockroachdb db (@tbg ). So we can't directly remove it. It's safer to remove it in 3.7.

You cannot depend on branches, you depend on commit hashes. Don't think  cockroachdb makes a difference here.

ahrtr:
> You cannot depend on branches, you depend on commit hashes.

Right. If there is no any objection, Let me deliver a PR to remove it.

ahrtr:
https://github.com/etcd-io/etcd/pull/14882

ahrtr:
Linked to https://github.com/etcd-io/etcd/issues/14713

PR #14882, opened 2022-12-02
Remove raft from etcd

ahrtr:
Please use go.etcd.io/raft/v3 instead. Linked to https://github.com/etcd-io/etcd/issues/14713


ahrtr:
Per the discussion in https://github.com/etcd-io/etcd/pull/14881.

cc @mitake @ptabor @serathius @spzala @tbg

ptabor (approved):
LGTM. But please wait before merging for @tbg approval.

tbg:
LGTM

CRDB pins a SHA, so no problem there.

tbg (approved):

ahrtr:
Thanks both @ptabor and @tbg

ahrtr:
> If we are on topic of project organization/code cleanup, maybe we could start discussion about splitting etcd gateway and etcd proxy into separate projects. @ptabor @ahrtr What do you think?

Actually I have been thinking this topic since the first day I was working on etcd. There are multiple modules in the same repo, and there are dependencies between them.  We use `replace` directive in go.mod files to always pick the latest code, but it's a pain-point for client applications, whenever they want to try the latest code of one of the module. 

Anyway, it's a big topic and let's do & evaluate this step by step.

Issue #14891, opened 2022-12-05
Segfault in `warnOfExpensiveReadOnlyTxnRequest` due to process pause


aphyr:
### What happened?

In Jepsen tests of etcd 3.5.6, process pauses can occasionally [induce etcd nodes to segfault](http://jepsen.io.s3.amazonaws.com/analyses/etcd-3.5.6/20221205T001403.482-0500.zip), logging

```
[...]

PR #12459, opened 2020-11-06
server: proper cancellation for range request

jingyih:
If a range request is timed out or canceled by client, server could stop serving this request in order to save resource. Note we cannot do similar things for mutating requests - once they passed raft and were appended to the WAL log, they must be finished applying to the database.

This PR wires the context to range function in apply. Ranging from database is stopped once context is canceled or timed out.

fix: #12458

Using the same example provided in the original issue. Client issues an expensive range request which timed out after 2 sec.
```
$ time etcdctl --command-timeout=2s get "" --prefix
{"level":"warn","ts":"2020-11-06T01:19:45.943+0800","caller":"v3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"endpoint://client-3e579972-3aa2-4942-a9f7-3d7fbd36e380/127.0.0.1:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
Error: context deadline exceeded

real    0m2.040s
user    0m0.023s
sys     0m0.028s
```

Server stopped serving the range request upon context time out.
```
{"level":"warn","ts":"2020-11-06T01:19:45.944+0800","caller":"etcdserver/util.go:158","msg":"apply request took too long","took":"1.998370669s","expected-duration":"100ms","prefix":"read-only range ","request":"key:\"\\000\" range_end:\"\\000\" ","response":"","error":"context canceled"}
{"level":"info","ts":"2020-11-06T01:19:45.944+0800","caller":"traceutil/trace.go:171","msg":"trace[1039134524] range","detail":"{range_begin:\u0000; range_end:\u0000; }","duration":"1.998720987s","start":"2020-11-06T01:19:43.945+0800","end":"2020-11-06T01:19:45.944+0800","steps":["trace[1039134524] 'range keys from in-memory index tree'  (duration: 170.321926ms)"],"step_count":1}
```

PR #14895, opened 2022-12-05
etcdserver: fix nil pointer panic for readonly txn

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/14891 

Please see https://github.com/etcd-io/etcd/issues/14891#issuecomment-1337191993

I need to add unit test to verify it.

cc @ptabor @serathius

### DIFF
--- a/server/etcdserver/txn/util.go
+++ b/server/etcdserver/txn/util.go
@@ -63,7 +63,11 @@ func WarnOfExpensiveReadOnlyTxnRequest(lg *zap.Logger, warningApplyDuration time
 		for _, r := range txnResponse.Responses {
 			switch op := r.Response.(type) {
 			case *pb.ResponseOp_ResponseRange:
-				resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.ResponseRange.Kvs)))
+				if op.ResponseRange != nil {
+					resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.ResponseRange.Kvs)))
+				} else {
+					resps = append(resps, "range_response:nil")
+				}
 			default:
 				// only range responses should be in a read only txn request
 			}
--- a/server/etcdserver/txn/util_test.go
+++ b/server/etcdserver/txn/util_test.go
@@ -1,0 +1,125 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package txn
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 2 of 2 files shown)


server/etcdserver/txn/util.go
```diff
@@ -63,7 +63,11 @@ func WarnOfExpensiveReadOnlyTxnRequest(lg *zap.Logger, warningApplyDuration time
 		for _, r := range txnResponse.Responses {
 			switch op := r.Response.(type) {
 			case *pb.ResponseOp_ResponseRange:
-				resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.ResponseRange.Kvs)))
+				if op.ResponseRange != nil {
+					resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.ResponseRange.Kvs)))
+				} else {
+					resps = append(resps, "range_response:nil_response")
```
serathius, line 69:
```suggestion
					resps = append(resps, "range_response:nil")
```
It's redundant to mention "response" second time.

ahrtr:
The old behavior of etcd was it always panic when the application process fails. But we changed the behavior for readonly txn in https://github.com/etcd-io/etcd/pull/14149. Readonly txn doesn't panic any more, but it causes this issue. 

I am thinking probably we should rollback the change in https://github.com/etcd-io/etcd/pull/14149, so as to keep the original behavior. In short, we should intentionally panic/crash etcd if the application fails, no matter it's readonly or not. More simpler, more stable. 

WDYT? @ptabor @serathius @ptabor

server/etcdserver/txn/util.go
ahrtr: updated

lavacat:
For ref, `panic` was originally added [here](https://github.com/etcd-io/etcd/pull/3534)
> we should check potential errors before executing the transaction. we cannot abort executed commands.
This comment wasn't fully implemented. All potential errors in txn are from kv layer. For example ErrFutureRev, ErrCompacted. None of `backend` interface methods return errors.

I think https://github.com/etcd-io/etcd/pull/14149 fixes legit use-case (see PR description). Context checking in range requests was [added](https://github.com/etcd-io/etcd/pull/12459) long after original `panic` logic. 
Potential improvement is to use `RangeResult.More` from https://github.com/etcd-io/etcd/pull/14810/files#diff-3b5c2c015ac7291d97cfb1daa9254c763300fc1cc0dfe668b3deed0080dffafbR37 instead of error when context is done.

ahrtr:
> Potential improvement is to use `RangeResult.More` from https://github.com/etcd-io/etcd/pull/14810/files#diff-3b5c2c015ac7291d97cfb1daa9254c763300fc1cc0dfe668b3deed0080dffafbR37 instead of error when context is done.

This isn't correct. `RangeResult.More` means read successfully, but only partial data is returned.  In this case, an error will explicitly returned to client in this case. 

Added an unit test case.  PTAL @ptabor @serathius @spzala

lavacat:
> This isn't correct. RangeResult.More means read successfully, but only partial data is returned. In this case, an error will explicitly returned to client in this case.

I didn't explain well. The idea is to return partial result and not fail on cancel/timeout. An example of similar behavior is Solr [timeAllowed](https://solr.apache.org/guide/solr/latest/query-guide/common-query-parameters.html#timeallowed-parameter) param. It's not the best solution, but I think better than `panic` inside txn range.

serathius (approved):
LGTM for fixing the panic. Don't have enough context about reverting the PR.

Issue #14894, opened 2022-12-05
auto-compaction-mode don't have default value when server configed with --config-file


monsterxx03:
### What happened?

config.yaml:

```
auto-compaction-retention: "2"
```

start etcd:

```
./etcd --config-file  config.yaml
```

check log:

"starting an etcd server","etcd-version":"3.6.0-alpha.0","git-sha":"9cc2f645c","go-version":"go1.19.1","go-os":"darwin","go-arch":"arm64","max-cpu-set":8,"max-cpu-available":8,"member-initialized":true,"name":"default","data-dir":"default.etcd","wal-dir":"","wal-dir-dedicated":"","member-dir":"default.etcd/member","force-new-cluster":false,"heartbeat-interval":"100ms","election-timeout":"1s","wait-cluster-ready-timeout":"5s","initial-election-tick-advance":true,"snapshot-count":100000,"max-wals":5,"max-snapshots":5,"snapshot-catchup-entries":5000,"initial-advertise-peer-urls":["http://localhost:2380"],"listen-peer-urls":["http://localhost:2380"],"advertise-client-urls":["http://localhost:2379"],"listen-client-urls":["http://localhost:2379"],"listen-metrics-urls":[],"cors":["*"],"host-whitelist":["*"],"initial-cluster":"","initial-cluster-state":"new","initial-cluster-token":"","quota-backend-bytes":2147483648,"max-request-bytes":1572864,"max-concurrent-streams":4294967295,"pre-vote":true,"initial-corrupt-check":false,"corrupt-check-time-interval":"0s","compact-check-time-enabled":false,"compact-check-time-interval":"1m0s",**"auto-compaction-mode":"","auto-compaction-retention":"0s","auto-compaction-interval":"0s"**,"discovery-url":"","discovery-proxy":"","discovery-token":"","discovery-endpoints":"","discovery-dial-timeout":"2s","discovery-request-timeout":"5s","discovery-keepalive-time":"2s","discovery-keepalive-timeout":"6s","discovery-insecure-transport":false,"discovery-insecure-skip-tls-verify":false,"discovery-cert":"","discovery-key":"","discovery-cacert":"","discovery-user":"","downgrade-check-interval":"5s","max-learners":1}

etcd is expected to do auto compaction every 2 hour, but compaction didn't happen.

if started with cli flag:  ./etcd --auto-compaction-retention 2, compaction mode and interval will be set correctly:

```
[...]

PR #14896, opened 2022-12-05
server: require set auto-compaction-mode explictly in config file

monsterxx03:
Fix https://github.com/etcd-io/etcd/issues/14894

Signed-off-by: monsterxx03 <xyj.asmy@gmail.com>

### DIFF
--- a/CHANGELOG/CHANGELOG-3.6.md
+++ b/CHANGELOG/CHANGELOG-3.6.md
@@ -67,6 +67,7 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 - Add [Protection on maintenance request when auth is enabled](https://github.com/etcd-io/etcd/pull/14663).
 - Graduated [`--experimental-warning-unary-request-duration` to `--warning-unary-request-duration`](https://github.com/etcd-io/etcd/pull/14414). Note the experimental flag is deprecated and will be decommissioned in v3.7.
 - Add [field `hash_revision` into `HashKVResponse`](https://github.com/etcd-io/etcd/pull/14537).
+- Require [Set auto-compaction-mode explicitly in config file](https://github.com/etcd-io/etcd/pull/14896).
 
 ### etcd grpc-proxy
 
--- a/server/embed/config_test.go
+++ b/server/embed/config_test.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+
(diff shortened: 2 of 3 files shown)


server/embed/config.go
```diff
@@ -628,6 +628,9 @@ func (cfg *configYAML) configFromFile(path string) error {
 	if cfg.ClusterState == "" {
 		cfg.ClusterState = ClusterStateFlagNew
 	}
+	if cfg.AutoCompactionMode == "" {
```
fuweid, line 631: It seems it is trying to align with commandline. https://github.com/etcd-io/etcd/blob/5baf8377bbb72d1c7f2b370fdb921091a069ab9d/server/etcdmain/config.go#L238

But etcd ignores any config from commandline or env if you use `config-file`.

https://github.com/etcd-io/etcd/blob/5baf8377bbb72d1c7f2b370fdb921091a069ab9d/server/etcdmain/help.go#L40)

IMO, I think the user should set the auto compaction mode explicitly in config-file since it is easy to add and maintain options in yaml than commandline. And it doesn't depend on the default behaviour. 

And I also think we can add the validation in `parseCompactionRetention`. https://github.com/etcd-io/etcd/blob/ddd65e7fe2e2d6fbb9a89a07ddbef79412885659/server/embed/etcd.go#L794-L802

If the `retention` is just number and `mode` is empty, we should return error.

monsterxx03: Thanks for your review. 

Make sense. Currently if auto-compaction-mode is set as '1m' in config-file, etcd will report "unsupported compaction mode", but set number "1", will disable auto compaction silently, it's confusion.

fuweid (approved):
LGTM(non-binding)

ping @ahrtr @ptabor @serathius

ahrtr:
This PR changes the existing behavior. Once an `auto-compaction-retention` is set, it requires that users must set a correct `auto-compaction-mode`. 

The etcd CLI sets a default value "`periodic`" for compact-mode, but there is no default value for embedded use case. So this PR is a breaking change for the embedded use case. Of course, it also affects the CLI + config-file use cases. cc @ptabor @serathius @spzala for options.

If we agree on this, then I would suggest to add validation in `(*Config) Validate`. We can enhance the existing [validation](https://github.com/etcd-io/etcd/blob/142fa76a11ba3cbb0e4303d353a1ad4b75a4aafa/server/embed/config.go#L742-L747).

ahrtr:
Please also rebase this PR. It should can resolve the workflow failures.

PR #14897, opened 2022-12-05
Allow failpoint requests to fail assuming that process exists within 1 second

serathius:
cc @ahrtr @tjungblu @ptabor

### DIFF
--- a/tests/framework/e2e/etcd_process.go
+++ b/tests/framework/e2e/etcd_process.go
@@ -42,7 +42,8 @@ type EtcdProcess interface {
 	EndpointsMetrics() []string
 	Client(opts ...config.ClientOption) *EtcdctlV3
 
-	Wait() error
+	IsRunning() bool
+	Wait(ctx context.Context) error
 	Start(ctx context.Context) error
 	Restart(ctx context.Context) error
 	Stop() error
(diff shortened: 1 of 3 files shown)


fuweid (approved):
LGTM(non-binding)

tests/linearizability/failpoints.go
```diff
@@ -115,44 +125,55 @@ const (
...
+	member := f.pickMember(t, clus)
 	address := fmt.Sprintf("127.0.0.1:%d", member.Config().GoFailPort)
-	err := setupGoFailpoint(address, f.failpoint, f.payload)
-	if err != nil {
-		return fmt.Errorf("gofailpoint setup failed: %w", err)
-	}
-	if f.trigger != nil {
-		err = f.trigger(ctx, member)
+
+	triggerCtx, cancel := context.WithTimeout(ctx, time.Second)
```
ahrtr, line 131: Reuse `triggerTimeout`?

tests/linearizability/failpoints.go
```diff
@@ -115,44 +125,55 @@ const (
...
+			err = f.trigger(triggerCtx, member)
+			if err != nil {
+				t.Logf("triggering gofailpoint failed: %v", err)
+			}
+		}
+		err = member.Wait(triggerCtx)
+		if err != nil && !strings.Contains(err.Error(), "unexpected exit code") {
+			return fmt.Errorf("failed to trigger a process panic within %s, err: %w", triggerTimeout, err)
 		}
 	}
```
ahrtr, line 150: This isn't correct. A gofailpoint isn't necessarily causing etcd crash. It works for now because all current failpoints are `panic`, but we may add other failpoints, e.g. `sleep`.

serathius:
Rerun tests in this PR 6 times. Removing flakiness was fixed.

tests/linearizability/failpoints.go
serathius: Fixed

tests/linearizability/failpoints.go
serathius: Yes but we currently only support crash. I have added a refactor PR that makes it explicit.

ahrtr (approved):
LGTM

It's OK to fix the comment in separate PR.

tests/linearizability/failpoints.go
```diff
@@ -114,45 +123,55 @@ const (
...
 		return err
 	}
 	return nil
 }
 
-func setupGoFailpoint(host, failpoint, payload string) error {
+func (f goPanicFailpoint) pickMember(t *testing.T, clus *e2e.EtcdProcessCluster) e2e.EtcdProcess {
+	switch f.target {
+	case AnyMember:
+		return clus.Procs[rand.Int()%len(clus.Procs)]
```
ptabor, line 160: [off topic] Shell we have 3 targets: AnyMember, Leader, NotLeader ? 
Randomly killing either leader or not leader might be a source of flakiness (on the other hand it expands exploratory test coverage).

tests/framework/e2e/etcd_process.go
```diff
@@ -201,11 +202,35 @@ func (ep *EtcdServerProcess) Kill() error {
...
+			ep.proc.Wait()
+			ep.cfg.lg.Info("server exited", zap.String("name", ep.cfg.Name))
+		}
+	}()
+	select {
+	case <-ch:
+		ep.proc = nil
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
```
ptabor, line 219: This logic is leaking go-routine if I understands correctly... 
We will exit with error, but we will keep waiting on the ep.proc.Wait(). 

Eventually the process we will finish and the go-routine will get cleaned up. Just be aware that it can lead to goroutine-leak detection (as long as we don't pass ep.proc.Wait(ctx) that enables Wait cancelation when context is done.

PR #14902, opened 2022-12-06
Linearizability issue14685

serathius:
cc @ahrtr @ptabor @geetasg @tjungblu


tests/linearizability/model.go
```diff
@@ -187,24 +197,32 @@ func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bo
...
+		if !deleteSucceeded && state.LastRevision != response.Revision {
+			return false, state
+		}
+	} else {
+		// if key was set, revision must increase as either key was deleted or failed request was persisted.
+		if keySet && response.Revision < state.LastRevision+1 {
+			return false, state
+		}
+
+		// if delete succeeded (+1 rev) without any key, means that fail request was persisted (+1 rev).
```
ptabor, line 222: Could you, please elaborate here more. What do you mean by persisting failRequest (in MVCC) ?

ptabor (approved):

tests/linearizability/model.go
serathius: From client perspective etcd is a black box. When clients get an error, etcd might never received the request as well already committed it to storage.

tjungblu: I also didn't catch that part, how can the delete revision jump by up to two here?

tests/linearizability/model.go
```diff
@@ -170,11 +176,15 @@ func stepPut(state EtcdState, request EtcdRequest, response EtcdResponse) (bool,
...
 		return true, state
 	}
-	if response.Revision <= state.LastRevision {
+	// Put should always increase revision
+	if response.Revision == state.LastRevision {
 		return false, state
 	}
-	if response.Revision != state.LastRevision+1 && state.FailedWrite == nil {
-		return false, state
+	if state.FailedWrite == nil {
```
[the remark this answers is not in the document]
serathius: > for a PUT this would be clearer?

Rewrote this code to be consistent with Delete. Goal is to make `state.FailedWrite == nil` to top of checks because it impacts logic a lot. Put logic is much simpler, but I would want to keep the same code template.

> What the guarantee for a failed write? revision can be +1/+0?

Revision should increase by at least 1, need to fix this

tests/linearizability/model.go
ptabor: Yes. That's the exact concern. The requests either failed or succeeded... but not both (Schrödinger's cat), so if `+2` is justified, please provide more in-detail explanation.

tests/linearizability/model.go
```diff
@@ -187,24 +198,32 @@ func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bo
...
+			return false, state
+		}
+
+		// if delete succeeded (+1 rev) without any key, means that fail request was persisted (+1 rev).
+		if deleteSucceeded && !keySet && response.Revision < state.LastRevision+2 {
+			return false, state
+		}
 	}
 
 	state.Value = ""
```
ahrtr, line 229: We need to cleanup `FailedWrite` on successful delete operation.
```
state.FailedWrite = nil
```

tests/linearizability/model.go
```diff
@@ -187,24 +198,32 @@ func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bo
...
+		}
+		// if key was not deleted, response revision should not change
+		if !deleteSucceeded && state.LastRevision != response.Revision {
+			return false, state
+		}
+	} else {
+		// if key was set, revision must increase as either key was deleted or failed request was persisted.
+		if keySet && response.Revision < state.LastRevision+1 {
+			return false, state
+		}
```
ahrtr, line 221: This may not be correct. Let's work with an example, assuming there are three successive requests:
1. read: successful, so the state.Value is set as the `response.GetData`
2. delete:  failed on client, but the etcdserver actually successfully deleted the k/v. So the state.Value keeps unchanged, but `state.FailedWrite` is updated 
3. delete: Since the k/v has already been deleted, so the revision will not increase.

ahrtr:
Suggest to rewrite the `stepDelete` below, 

```
func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bool, EtcdState) {
	if response.Err != nil {
		state.FailedWrite = &request
		return true, state
	}
	deleteSucceeded := response.Deleted != 0
	keySet := state.Value != ""

	if state.FailedWrite == nil {
		// non-existent key cannot be deleted.
		if deleteSucceeded != keySet {
			return false, state
		}
		// if key was deleted, response revision should increase by one
		if deleteSucceeded && response.Revision != state.LastRevision+1 {
			return false, state
		}
		// if key was not deleted, response revision should not change
		if !deleteSucceeded && state.LastRevision != response.Revision {
			return false, state
		}
	} else {
		// When the client side got an error response on previous write
		// operation, the etcdserver might actually have succeeded.
		// In this case the response revision should increase by two.
		// So the response revision should increase at least by one.
		if deleteSucceeded && response.Revision < state.LastRevision+1 {
			return false, state
		}
	}

	state.Value = ""
	state.LastRevision = response.Revision
	state.FailedWrite = nil
	return true, state
}
```

tests/linearizability/model.go
ahrtr: Please see https://github.com/etcd-io/etcd/pull/14902#issuecomment-1340041130, pls pay attention to the comment I added.

serathius:
As most comments for this PR are about refactor I will split out the reproduction commit so we can iterate on refactor more.

tests/linearizability/model.go
serathius: You are right.

tests/linearizability/model.go
serathius: It will increase due to delete in point 2. 

Assuming that we know that at least one write failed `state.FailedWrite != nil`, table of cases is:

keySet | deleteSucceeded | revision change | comment
-----------|--------------------------|-------------------------|-----------------
 true    |  true                        | +1                         | Delete just worked on existing key. Failed request were lost
 false    |  true                        | +2                         | Delete succeeded on non-existing key. Means that there was at least one Put request that failed. Meaning since last state update there were two transactions: Put, Delete(success).
 true    |  false                        | +1                         | Delete failed on existing key. Means that there was at least one Put request that failed. Meaning since last state update there were two transactions: Put, Delete(failed).
 false    |  false                        | 0                         | Delete just failed on non-existing key. Failed request were lost.

tests/linearizability/model.go
serathius: I think that `state.FailedWrite = nil` should be set even if delete failed. No matter if delete failed or not, we know final state. Value will be empty no matter what.

ahrtr: > we know final state. Value will be empty no matter what

Do not understand this. If the client gets an error response, how can you know the real server side state?

tests/linearizability/model.go
ahrtr: Anyway, please see https://github.com/etcd-io/etcd/pull/14902#issuecomment-1340041130

serathius:
Rewrote the model to be much simpler. Preferred correctness and simplicity but sacrificed performance. I think we can look into performance as a separate goal.

tests/linearizability/model.go
```diff
@@ -43,10 +43,12 @@ type EtcdResponse struct {
...
-	Value        string
-	LastRevision int64
-	FailedWrite  *EtcdRequest
+	Key            string
+	PossibleValues []valueRevision
+}
+
+func (resp EtcdResponse) Equals(other EtcdResponse) bool {
+	return resp.GetData == other.GetData && resp.Revision == other.Revision && resp.Deleted == other.Deleted && resp.Err == other.Err
 }
```
ahrtr, line 52: Do not need at all. You can just use `etcdResp1 == etcdResp2` directly. 
```suggestion
```

serathius: Done, I was worried about cases where we add pointers, but let's not worry about this now.

serathius:
In the PR I proposed I have separated handling of errors, so code for handling correct path `handle` is trivial.
This makes adding new operations also trivial, as contributor doesn't need to handle failed states. They just code the correct path.

This is examplified by Txn implementation https://github.com/etcd-io/etcd/pull/14903/commits/ad5840c6147372ade5b499bc127d925f66cc2038

Changes to model are just 14 lines

tests/linearizability/model.go
```diff
@@ -93,121 +91,83 @@ var etcdModel = porcupine.Model{
...
+	switch request.Op {
+	case Get:
+		response.GetData = v.Value
+	case Put:
+		v.Value = request.PutData
+		v.Revision += 1
+	case Delete:
+		if v.Value != "" {
+			v.Value = ""
+			v.Revision += 1
```
ahrtr, line 165: It might not be correct to in-place update the history?  Otherwise, you need to differentiate successful and failed request? If I understand correctly, this method is just to generate an expected response based on current state.

tests/linearizability/model.go
```diff
@@ -93,121 +91,83 @@ var etcdModel = porcupine.Model{
...
-		panic("invalid request")
-	}
-	if state.Key == "" {
-		return true, initState(request, response)
+	if len(state.PossibleValues) == 0 {
+		state.Key = request.Key
+		ok, v := initValueRevision(request, response)
+		if ok {
+			state.PossibleValues = append(state.PossibleValues, v)
+		}
```
ahrtr, line 99:
```suggestion
		if ok, v := initValueRevision(request, response); ok {
			state.PossibleValues = append(state.PossibleValues, v)
		}
```

tests/linearizability/model.go
serathius: I wrote the code in a way that it is. When iterating through possible values I always copy.
This is tested by all pre-existing tests.

serathius: If you prefer that we don't update `ValueRevision` in place for readability please let me know. I compared both versions and picked this one.

ahrtr: This is not just readability, it also is wrong.  The receiver is pointer, any in-place updating will affect the caller. The reason why you do not run into issue is because the golang for--range loop made a copy for you. It's a potential risk that new contributors may misunderstand and generate buggy code. 

Please do not modify the value in-place.

tests/linearizability/model.go
```diff
@@ -93,121 +91,82 @@ var etcdModel = porcupine.Model{
...
+				state.PossibleValues[i] = kv
+				i++
+			}
+		}
+		state.PossibleValues = state.PossibleValues[:i]
 	}
+	return len(state.PossibleValues) > 0, state
+}
+
+type valueRevision struct {
```
ahrtr, line 123: Please put `valueRevision` at the place either close to its method `(*valueRevision) handle` or where it is used `PossibleValues []valueRevision`

tests/linearizability/model.go
```diff
@@ -93,121 +91,82 @@ var etcdModel = porcupine.Model{
...
-		default:
-			panic("Unknown operation")
-		}
-		if ok {
-			state.Value = response.GetData
-			state.LastRevision = response.Revision
-			state.FailedWrite = nil
-			return true, state
+func (v *valueRevision) handle(request EtcdRequest) EtcdResponse {
+	response := EtcdResponse{}
```
ahrtr, line 154:
```suggestion
	response := EtcdResponse{Revision: v.Revision}
```

tests/linearizability/model.go
```diff
@@ -93,121 +91,82 @@ var etcdModel = porcupine.Model{
...
-	}
-	//if key was not deleted, response revision should not change
-	if !deleteSucceeded && state.LastRevision != response.Revision && state.FailedWrite == nil {
-		return false, state
-	}
-
-	state.Value = ""
-	state.LastRevision = response.Revision
-	return true, state
+	response.Revision = v.Revision
```
ahrtr, line 170:
```suggestion
```

tests/linearizability/model.go
```diff
@@ -93,121 +91,82 @@ var etcdModel = porcupine.Model{
...
-			ok = response.GetData == ""
-		default:
-			panic("Unknown operation")
-		}
-		if ok {
-			state.Value = response.GetData
-			state.LastRevision = response.Revision
-			state.FailedWrite = nil
-			return true, state
+func (v *valueRevision) handle(request EtcdRequest) EtcdResponse {
```
ahrtr, line 153:
```suggestion
func (v valueRevision) handle(request EtcdRequest) EtcdResponse {
```

tests/linearizability/model.go
```diff
@@ -93,121 +91,82 @@ var etcdModel = porcupine.Model{
...
+			state.PossibleValues = append(state.PossibleValues, kv)
+		}
+	} else {
+		var i = 0
+		for _, kv := range state.PossibleValues {
+			expectedResponse := kv.handle(request)
+			if expectedResponse == response {
+				state.PossibleValues[i] = kv
+				i++
+			}
```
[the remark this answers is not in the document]
serathius: We do.

`state.PossibleValues = state.PossibleValues[:i]` removes all the history that is not consistent with current state.

tests/linearizability/model.go
serathius: > The reason why you do not run into issue is because the golang for--range loop made a copy for you.

Yes, I know, That was intentional so the code was correct. Rewrote to copy for clarity.

tests/linearizability/model.go
```diff
@@ -93,121 +96,77 @@ var etcdModel = porcupine.Model{
...
+			newV, _ := stepValue(v, request)
+			state.PossibleValues = append(state.PossibleValues, newV)
+		}
+	} else {
+		var i = 0
+		for _, v := range state.PossibleValues {
+			newV, expectedResponse := stepValue(v, request)
+			if expectedResponse == response {
+				state.PossibleValues[i] = newV
+				i++
```
ahrtr, line 120: Why not just like this? 

```suggestion
				state.PossibleValues = []valueRevision{newV}
				return  true, state
```

tests/linearizability/model.go
```diff
@@ -93,121 +96,77 @@ var etcdModel = porcupine.Model{
...
-		return stepPut(state, request, response)
-	case Delete:
-		return stepDelete(state, request, response)
-	default:
-		panic("Unknown operation")
+	if response.Err != nil {
+		for _, v := range state.PossibleValues {
+			newV, _ := stepValue(v, request)
+			state.PossibleValues = append(state.PossibleValues, newV)
+		}
```
ahrtr, line 113: add 

```
return true, state
```

ahrtr:
I would also suggest to add the following check:
1. The returned read data must be one of the historical data. FYI. https://github.com/etcd-io/etcd/pull/14914/files#diff-f6cd9cfbf4b5310cea9742f346927212c88f39b5cab928a434fe3ffb44d50cccR139-R158 
2. If the request succeeds, then the revision should be bigger than any historical revision + 1 for put and delete operation if deleted != 0.

ahrtr:
https://github.com/etcd-io/etcd/pull/14914/files#diff-f6cd9cfbf4b5310cea9742f346927212c88f39b5cab928a434fe3ffb44d50cccR175-R181

https://github.com/etcd-io/etcd/pull/14914/files#diff-f6cd9cfbf4b5310cea9742f346927212c88f39b5cab928a434fe3ffb44d50cccR200-R211

serathius:
> I would also suggest to add the following check:
> 
> 1. The returned read data must be one of the historical data. FYI. https://github.com/etcd-io/etcd/pull/14914/files#diff-f6cd9cfbf4b5310cea9742f346927212c88f39b5cab928a434fe3ffb44d50cccR139-R158
> 2. If the request succeeds, then the revision should be bigger than any historical revision + 1 for put and delete operation if deleted != 0.

Those are given for free.

tests/linearizability/model.go
serathius: Not needed, by comparing response vs expected response we will not allow revision to decrease.

tests/linearizability/model.go
```diff
@@ -93,121 +96,77 @@ var etcdModel = porcupine.Model{
...
+	} else {
+		var i = 0
+		for _, v := range state.PossibleValues {
+			newV, expectedResponse := stepValue(v, request)
+			if expectedResponse == response {
+				state.PossibleValues[i] = newV
+				i++
+			}
+		}
+		state.PossibleValues = state.PossibleValues[:i]
```
ahrtr, line 123: https://github.com/etcd-io/etcd/pull/14902/files#r1043092652 

Why do you still keep the entries [0, i)? I think we only need to keep the latest one.

tests/linearizability/model.go
ahrtr: This might not be correct. It will return `true` when any expected response matches the actual response. How can you guarantee that all the historical entries' revision <= response.Revision?

ahrtr: Let's work with an example, 

Assuming initial value: revision is `10: 1`, and there are 3 successive failed PUT operation, so the highest revision cached will be 4, thanks to code at line 101 ~ 102. But actually the real revision on server side is still 1.  If the 4th PUT operation succeeded, then the returned revision will be 2, obviously it breaks the rule that revision should never decrease.

tests/linearizability/model.go
serathius: Because depending on request type, there might be multiple possible outputs. Imagine a situation:
* Init state: `[{"rev": 1, value:"1"}]`
* failed PUT("2") request. State: `[{"rev": 1, value:"1"}, {"rev": 2, value:"2"}]`
* TXN(if(value="3").then(value="4") that failed. State `[{"rev": 1, value:"1"}, {"rev": 2, value:"2"}]`
* Get. 

Both values `1` and `2` are equally probable. Txn doesn't cause state to collapse.

tests/linearizability/model.go
serathius: What you described is a correct history. Failed requests are those that client didn't response from. Maybe they succeeded maybe not, client didn't see the results. So in case you described, failed requests were just lost.

Note, goal of the model is to confirm that history up to some point is correct. It porcupine library which tries to find a order that model accepts as correct.    

> How can you guarantee that all the historical entries' revision [+1] <= response.Revision?

This is because revision in state never decreases. Please read `stepValue` function it will never result in lowering the revision. So when comparing response, response that lowered revision will never match. If no responses match, we will remove all possible values and return invalid state.

tests/linearizability/model.go
ahrtr: > TXN(if(value="3").then(value="4") that failed

What we are talking about is the `else` branch (`response.Err == nil`), it means the request succeeds. So I assume "that failed" means the TXN's Compare condition is evaluated to `false`.  In this case, we might still need to cache the revision in the response header, because it's a real current state returned from the server side. 

The `stepValue` function doesn't support TXN yet, so I suggest not to enhance the implementation for a case we do not support yet. We can continue to enhance once we support TXN.  The TXN complicates the conversation and implementation. Can we keep it as simple as possible for now?

serathius: No, there might be cases where multiple states are correct after transaction. For example:
* `get a -> 1, rev 1`. Possible states: [{"value": "1", "revision": 1}]
* `put a 2 -> error`. Possible states: [{"value": "1", "revision": 1}, {"value": "2", "revision": 2}]
* `put a 3 -> error`. Possible states: [{"value": "1", "revision": 1}, {"value": "2", "revision": 2}, {"value": "3", "revision": 2}, {"value": "3", "revision": 3}]
* `txn if(a=4).then(put a 5) -> false, rev 2`. Possible states: [{"value": "2", "revision": 2}, {"value": "3", "revision": 2}]

Because failed TXN doesn't know read the key, we don't know the end state. There are 4 possible histories here:
1. `get`, `txn`
2. `get`, `put 2`, `txn` 
3. `get`, `put 3`, `txn` 
4. `get`, `put 2`, `put 3`, `txn`

We can eliminate 1 and 4, because they would not match returned revision 2. But both 2 and 3 are possible as they both end with revision equal 2. So we don't know whether `put 2` or `put 3` was peristed.

ahrtr: > * `txn if(=4).then(put 5) -> false, rev 2`. Possible states: [{"value": "2", "revision": 2}, {"value": "3", "revision": 2}]

This is a good point. We successfully removed one impossible entry from the `PossibleValues`.  Just as I mentioned in https://github.com/etcd-io/etcd/pull/14902#discussion_r1043849481, can we not consider TXN for now? Instead, let's consider supporting it after merging this PR?

ahrtr: `state.PossibleValues = state.PossibleValues[:i]`

When there isn't any match, `then i == 0`, accordingly all values are cleaned up. This might not be correct.  As long as the request succeeds, we need to save the latest response (data and revision). WDYT?

ahrtr:
After a long back and forth conversation (Please provide as much details/description as possible beforehand, especially when you completely changed the design/direction),  what this PR is trying to do gets clearer. Overall looks good to me, and the performance might not be a problem as long as there isn't too many failed requests.

The only big comment is that I suggest to remove supporting TXN in this PR, and keep this PR as simple & reliable as possible. We can consider supporting TXN in next step. TXN really complicates the conversation and implementation, it definitely deserve another round of discussion & review session.

tests/linearizability/model.go
serathius: > What we are talking about is the else branch (response.Err == nil), it means the request succeeds. So I assume "that failed" means the TXN's

Yes, sorry for not clarifying it. 

> In this case, we might still need to cache the revision in the response header, because it's a real current state returned from the server side.

We don't need to cache. Major change from old code to the new code is that we get from partial to full traceability.
In previous model we just noted last failed request. If there was multiple failed requests in row, we stopped tracking the state until we get a successful request. Only think we validated was that revision increased. 

Proposed model does full tracing of requests. Which means for every failed request, we fork internal state to represent both possibilities of it being persisted and being lost. It has  exponential complexity, but guarantees correctness. 

We don't do any caching of revision as one of possible states should already have the revision. This is what filtering on `response == expectedResponse` does. It only leaves values that would end up the revision that we got from the etcd server.

 > The TXN complicates the conversation and implementation. Can we keep it as simple as possible for now?
 
 Don't agree, picking first match is not correct behavior. It's a difference of filtering vs picking random. Txn doesn't change that. If there are multiple states that would return correct response, we cannot just pick randomly from them. We need to allow both possibilities and assuming that on next `step` we will collapse the state.

serathius:
Merging based on lgtm from @ahrtr and approve from @ptabor

ahrtr:
Two minor comments: 
1. add a break below, 
```
diff --git a/tests/linearizability/model.go b/tests/linearizability/model.go
index 00fd2e3d3..a98019d80 100644
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -118,6 +118,7 @@ func step(state EtcdState, request EtcdRequest, response EtcdResponse) (bool, Et
                        if expectedResponse == response {
                                state.PossibleValues[i] = newV
                                i++
+                               break
                        }
                }
                state.PossibleValues = state.PossibleValues[:i]
```
2. When `i == 0`, all entries are cleanedup. I think we still need save the latest entry (which should be calculated based on the req and resp).

PR #14903, opened 2022-12-06
tests: Add Txn operation to linearizability tests

serathius:
cc @ahrtr @ptabor @geetasg @tjungblu

### DIFF
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -27,19 +27,23 @@ const (
 	Get    Operation = "get"
 	Put    Operation = "put"
 	Delete Operation = "delete"
+	Txn    Operation = "txn"
 )
 
 type EtcdRequest struct {
-	Op      Operation
-	Key     string
-	PutData string
+	Op            Operation
+	Key           string
+	PutData       string
+	TxnExpectData string
+	TxnNewData    string
 }
 
 type EtcdResponse struct {
-	GetData  string
-	Revision int64
-	Deleted  int64
-	Err      error
+	GetData      string
+	Revision     int64
+	Deleted      int64
+	TxnSucceeded bool
+	Err          error
 }
 
 type EtcdState struct {
@@ -88,6 +92,12 @@ var etcdModel = porcupine.Model{
 				return fmt.Sprintf("delete(%q) -> %s", request.Key, response.Err)
 			} else {
 				return fmt.Sprintf("delete(%q) -> ok, rev: %d deleted:%d", request.Key, response.Revision, response.Deleted)
+			}
+		case Txn:
+			if response.Err != nil {
+				return fmt.Sprintf("txn(if(value(%q)=%q).then(put(%q, %q)) -> %s", request.Key, request.TxnExpectData, request.Key, request.TxnNewData, response.Err)
+			} else {
+				return fmt.Sprintf("txn(if(value(%q)=%q).then(put(%q, %q)) -> %v, rev: %d", request.Key, request.TxnExpectData, request.Key, request.TxnNewData, response.TxnSucceeded, response.Revision)
 			}
 		default:
 			return "<invalid>"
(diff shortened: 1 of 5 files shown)


tests/linearizability/traffic.go
```diff
@@ -50,34 +50,42 @@ func (t readWriteSingleKey) Run(ctx context.Context, c *recordingClient, limiter
...
 	case Put:
 		err = c.Put(putCtx, t.key, fmt.Sprintf("%d", id))
 	case Delete:
 		err = c.Delete(putCtx, t.key)
+	case Txn:
+		var value string
+		if readResponse == "" {
+			value = fmt.Sprintf("%d", id)
+		} else {
+			value = fmt.Sprintf("%s,%d", readResponse, id)
```
ahrtr, line 86: Have you checked the visualized diagram? I guess it may be hard to read, because the value may be long long string.

ahrtr: It would be better to replace the suffix ID instead of always appending an ID.

tests/linearizability/model.go
```diff
@@ -209,5 +231,51 @@ func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bo
...
+
+func stepTxn(state EtcdState, request EtcdRequest, response EtcdResponse) (bool, EtcdState) {
+	if response.Err != nil {
+		state.FailedWrite = &request
+		return true, state
+	}
+	// revision should never decrease
+	if response.Revision < state.LastRevision {
+		return false, state
+	}
```
ahrtr, line 246: I recall that you added this check in the `step` function in another PR. So pls remove this now or after you merge the other PR.

tests/linearizability/model.go
```diff
@@ -209,5 +231,51 @@ func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bo
...
+	valueMatch := state.Value == request.TxnExpectData
+
+	if state.FailedWrite == nil {
+		// transaction should succeed if value matches
+		if response.TxnSucceeded != valueMatch {
+			return false, state
+		}
```
ahrtr, line 253: This isn't correct.  Note that in the `(*readWriteSingleKey) Run`, the `Read` and `Write` operation are not an atomic operation.  Pls see example below, obviously `valueMatch` will be `false`, but it's expected behavior. 

1. client 1 read:  `request.TxnExpectData = readResonse`
2. client 2 write (PUT):  The value changed.
3. client 1 write (TXN): The `txn.If `will be evaluated to `false`

tests/linearizability/model.go
```diff
@@ -209,5 +231,51 @@ func stepDelete(state EtcdState, request EtcdRequest, response EtcdResponse) (bo
...
+	} else {
+		// if value matches, revision must increase as either transaction succeeded or failed request was persisted.
+		if valueMatch && response.Revision < state.LastRevision+1 {
+			return false, state
+		}
+
+		// if txn succeeded (+1 rev) without matching key, means that fail request was persisted (+1 rev).
+		if response.TxnSucceeded && !valueMatch && response.Revision < state.LastRevision+2 {
+			return false, state
+		}
```
ahrtr, line 272: Note that we shouldn't depend on `valueMatch` at all.
```suggestion
		// See https://github.com/etcd-io/etcd/pull/14902#issuecomment-1340041130
		if response.TxnSucceeded && response.Revision < state.LastRevision+1 {
			return false, state
		}
```

tests/linearizability/traffic.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	DefaultTraffic Traffic = readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 90}, {operation: Delete, chance: 10}}}
+	DefaultTraffic Traffic = readWriteSingleKey{key: "key", writes: []opChance{{operation: Put, chance: 50}, {operation: Delete, chance: 5}, {operation: Txn, chance: 5}}}
```
ahrtr, line 27: Shouldn't the three values be summed up to 100?

serathius: Fixed

tests/linearizability/traffic.go
serathius: It's manageable as values are overriden/reset by Put and Delete requests.

serathius:
Rebased the change based on https://github.com/etcd-io/etcd/pull/14902, much much cleaner.

ahrtr:
FYI. The Go Vulnerability failure is resolved in https://github.com/etcd-io/etcd/pull/14916

ahrtr (approved):
LGTM

Issue #14890, opened 2022-12-04
Lost update and cyclic information flow on process crash


aphyr:
### What happened?

In Jepsen tests of etcd 3.5.5 with process crashes, running in five node clusters on Debian Bullseye, the wr workload (which performs transactions which read and write keys without any guards) appears to exhibit both lost update and cyclic information flow.

## Lost Update

In [this test run](http://jepsen.io.s3.amazonaws.com/analyses/etcd-3.5.6/20221016T054319.456-0400.zip), we executed the following pair of transactions:

```clj
            :anomalies {:lost-update [{:key 277,
                                       :value 1,
                                       :txns [{:type :ok,
                                               :f :txn,
                                               :value [[:r 277 1]
                                                       [:w 277 4]],
                                               :time 26267084609,
                                               :process 50,
                                               :index 14140,
                                               :debug {:txn-res {:succeeded? true,
                                                                 :txns [],
                                                                 :header {:member-id -5790610604219245522,
                                                                          :revision 3927,
                                                                          :raft-term 5},
                                                                 :results [{:count 1,
                                                                            :kvs {277 {:value 1,
                                                                                       :version 1,
                                                                                       :create-revision 3925,
                                                                                       :mod-revision 3925}},
                                                                            :more? false,
                                                                            :header {:member-id 0,
                                                                                     :revision 3926,
                                                                                     :raft-term 0}}
                                                                           {:prev-kv [277
                                                                                      {:value 1,
                                                                                       :version 1,
                                                                                       :create-revision 3925,
                                                                                       :mod-revision 3925}],
                                                                            :prev-kv? true,
                                                                            :header {:member-id 0,
                                                                                     :revision 3927,
                                                                                     :raft-term 0}}]}}}
                                              {:type :ok,
                                               :f :txn,
                                               :value [[:r 277 1]
                                                       [:w 277 11]],
                                               :time 26511349152,
                                               :process 99,
                                               :index 14569,
                                               :debug {:txn-res {:succeeded? true,
                                                                 :txns [],
                                                                 :header {:member-id 5198333996179334402,
                                                                          :revision 4045,
                                                                          :raft-term 5},
                                                                 :results [{:count 1,
                                                                            :kvs {277 {:value 1,
                                                                                       :version 17,
                                                                                       :create-revision 3925,
                                                                                       :mod-revision 3985}},
                                                                            :more? false,
                                                                            :header {:member-id 0,
                                                                                     :revision 4044,
[...]

PR #14904, opened 2022-12-06
Linearizability issue14890

serathius:
cc @ahrtr @ptabor @geetasg @tjungblu @aphyr

Failed attempt to try reproducing https://github.com/etcd-io/etcd/issues/14890

I switched linearization model to append one resembling Jepsen https://jepsen.io/analyses/etcd-3.4.3#append

For now use single key and validates exact contents.
In loop each client:
* Read the value of key
* Create Txn if value didn't change, put original value appended with unique identifier.

This approach has one advantage that failed requests can be ignored. Model doesn't check exact requests, just ensures that prefix matches. 

@aphyr From what I understood (don't know closure at all), one difference with append flow is that Jepsen compares modRevision in Txn instead value. Do you see any other differences or suggestions?

### DIFF
--- a/.github/workflows/linearizability.yaml
+++ b/.github/workflows/linearizability.yaml
@@ -13,7 +13,7 @@ jobs:
         make build
         mkdir -p /tmp/linearizability
         cat server/etcdserver/raft.fail.go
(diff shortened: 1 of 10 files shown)


serathius:
No longer needed with TXN tests present.

PR #14905, opened 2022-12-06
Set Github workflows default permissions to read-only

### LINKED ISSUE #14878: Set Github workflows default permissions to read-only

It is a [scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) recommendation to set the [github workflow permissions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to read only by default.  Thus, any action that would need a greater permission should be specified to have the permission, otherwise it will be read-only.

Can I suggest a PR with this changes?

### Why is this needed?

When the default permission of a workflow is not set to read-only, it is consider "write", by default. This allows any action listed in the workflow the write permission in the repository, even when they are not supposed to be able to write. 

The [Token Permission Check Documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) list some security concerns that allowing write permission to actions raises.

joycebrum:
Closes #14878 

I've configured and tested all workflows using default permissions as read-only, only the release.yaml and converage.yaml I wasn't able to test. I've assumed, looking to the scripts, that the release.yaml would need a write permission on it's job, the coverage.yaml didn't seem to need any write permission.

Let me know if I've misjudged it.

Thanks!

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,8 +1,11 @@
 name: E2E
 on: [push, pull_request]
+permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     strategy:
       fail-fast: true
       matrix:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -4,8 +4,6 @@ permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
     strategy:
       fail-fast: true
       matrix:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,8 +3,6 @@ on: [push, pull_request]
 permissions: read-all
 jobs:
   main:
-    permissions:
-      contents: write
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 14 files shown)


.github/workflows/e2e.yaml
```diff
@@ -1,8 +1,11 @@
 name: E2E
 on: [push, pull_request]
+permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
ahrtr, line 8: e2e doesn't need write permission.

.github/workflows/release.yaml
```diff
@@ -1,7 +1,10 @@
 name: Release
 on: [push, pull_request]
+permissions: read-all
 jobs:
   main:
+    permissions:
+      contents: write
```
ahrtr, line 7: I don't think the release workflow need write permission either

ahrtr:
Thanks @joycebrum for delivering this PR. 

All workflows only need read-only permission. cc @serathius to double confirm.

serathius:
Don't think any workflows needs write permission none of them do any writes to repository or artifacts on Github. 
As pointed by @ahrtr, release workflow runs in dry mode and e2e also doesn't need write access.

joycebrum:
Good to know, I'll revoke the write permissions then. Thanks!

PR #14907, opened 2022-12-07
chore: use Getter in WarnOfExpensiveReadOnlyTxnRequest 

fuweid:
The pb provides an accessor method to get field and it will not panic if
the owner is nil. And add non-empty RangeRespone into the test case.

ping @ahrtr

### DIFF
--- a/server/etcdserver/txn/util.go
+++ b/server/etcdserver/txn/util.go
@@ -61,10 +61,10 @@ func WarnOfExpensiveReadOnlyTxnRequest(lg *zap.Logger, warningApplyDuration time
 	if !isNil(txnResponse) {
 		var resps []string
 		for _, r := range txnResponse.Responses {
-			switch op := r.Response.(type) {
+			switch r.Response.(type) {
 			case *pb.ResponseOp_ResponseRange:
-				if op.ResponseRange != nil {
-					resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.ResponseRange.Kvs)))
+				if op := r.GetResponseRange(); op != nil {
+					resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.GetKvs())))
 				} else {
 					resps = append(resps, "range_response:nil")
 				}
--- a/server/etcdserver/txn/util_test.go
+++ b/server/etcdserver/txn/util_test.go
@@ -19,12 +19,19 @@ import (
 	"time"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/mvccpb"
+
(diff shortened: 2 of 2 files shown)


server/etcdserver/txn/util.go
```diff
@@ -61,13 +61,9 @@ func WarnOfExpensiveReadOnlyTxnRequest(lg *zap.Logger, warningApplyDuration time
...
 		for _, r := range txnResponse.Responses {
-			switch op := r.Response.(type) {
+			switch r.Response.(type) {
 			case *pb.ResponseOp_ResponseRange:
-				if op.ResponseRange != nil {
-					resps = append(resps, fmt.Sprintf("range_response_count:%d", len(op.ResponseRange.Kvs)))
-				} else {
-					resps = append(resps, "range_response:nil")
-				}
+				resps = append(resps, fmt.Sprintf("range_response_count:%d", len(r.GetResponseRange().GetKvs())))
```
ahrtr, line 66: It seems `range_response:nil` is a little better than `range_response:0` in this case. We can easily tell the TXN terminates/fails at which request. 

For example, there are 5 read requests in a TXN, it's easy to tell the TXN terminates/fails at the 3rd request, 
```
request 1:  "range_response_count: 10"
request 2:  "range_response_count: 0" (Note: no any error yet)
request 3:  "range_response_count: nil", some error occurs
request 4:  "range_response_count: nil" (will not be executed at all)
request 5:  "range_response_count: nil" (will not be executed at all)
```

But it's hard to tell which request the TXN terminates/fails at in this PR, 
```
request 1:  "range_response_count: 10"
request 2:  "range_response_count: 0" (Note: no any error yet)
request 3:  "range_response_count: 0", some error occurs
request 4:  "range_response_count: 0" (will not be executed at all)
request 5:  "range_response_count: 0" (will not be executed at all)
```

ptabor: No strong opinion: 
 1.  Usage of `Get` and `Has` is definitely more aligned with the long-term plans of evolution of protobuf library
     (As getter allow for more eficienient representation than the 'unpacked' properties). So I'm all for preferring this stale of accessing protos. 
     
 2. Agree that the old format caries more information. If the goal is readability, we should pass the underling error as well (but it's a bigger change). 


How about still having the If - as there is no ternary operation in Golang, but using the getters ?

fuweid: Thanks for the suggestion. Agree with @ahrtr. I updated it with Getter and keep `range_response:nil`. And also updated the test case with non-empty RangeResponse. PTAL. Thanks!

ptabor (approved):

ahrtr (approved):
LGTM

Thank you @fuweid

PR #14911, opened 2022-12-07
Create a nightly job for linearizability tests

tjungblu:
Start with a simple job against main that runs for 3h by repeating it more often than the PR job.

---

fix for #14045 
- [x]  Add a periodic (once a day?) long running (1h?) github action that runs linearizability tests

I didn't find a much better way to make it run for longer, I basically just incremented the --count by two orders of manitude. Let me know if there's a nicer way, happy to implement that.

The caveat with the schedule is that this is only for the main branch, we have to maybe split the jobs up to checkout the respective release branches in parallel and run the tests.

### DIFF
--- a/.github/workflows/linearizability-nightly.yaml
+++ b/.github/workflows/linearizability-nightly.yaml
@@ -1,0 +1,25 @@
+name: Linearizability Nightly
+on:
+  # schedules always run against the main branch
+  schedule:
+    - cron: '25 9 * * *'
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    # GHA has a maximum amount of 6h execution time, we try to get done within 3h
+    timeout-minutes: 180
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: "1.19.3"
+    - run: |
+        make gofail-enable
+        make build
+        mkdir -p /tmp/linearizability
+        cat server/etcdserver/raft.fail.go
+        EXPECT_DEBUG=true GO_TEST_FLAGS='-v --count 500 --failfast --run TestLinearizability' RESULTS_DIR=/tmp/linearizability make test-linearizability
+    - uses: actions/upload-artifact@v2
+      if: always()
+      with:
+        path: /tmp/linearizability/*


.github/workflows/linearizability-nightly.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - uses: actions/checkout@v2
+    - uses: actions/setup-go@v2
+      with:
+        go-version: "1.19.3"
+    - run: |
+        make gofail-enable
+        make build
+        mkdir -p /tmp/linearizability
+        cat server/etcdserver/raft.fail.go
+        EXPECT_DEBUG=true GO_TEST_FLAGS='-v --count 6000 --failfast --run TestLinearizability' RESULTS_DIR=/tmp/linearizability make test-linearizability
```
serathius, line 21: 6000 might be an overkill :P. 
60 runs take up to 20 minutes, so 500 should be enough.

.github/workflows/linearizability-nightly.yaml
```diff
@@ -0,0 +1,25 @@
+name: Linearizability Nightly
+on:
+  # schedules always run against the main branch
+  schedule:
+    - cron: '25 9 * * *'
```
serathius, line 5: Can you add a comment explaining the cron.

serathius (approved):
cc @ahrtr @ptabor

ahrtr:
Thanks @tjungblu .

The only concern is the Linearizability is still under heavily development, so it may be flaky.  I would suggest holding on enabling it or run it weekly instead of daily. WDYT? cc  @ptabor @serathius @spzala

serathius:
> The only concern is the Linearizability is still under heavily development, so it may be flaky.

That's why enabling periodics is super useful, it allow us to qualify both etcd and linearizability tests while not blocking any PRs.

> I would suggest holding on enabling it or run it weekly instead of daily. WDYT?

Don't see much difference, if we are worried about occupying github action workers we might decide to move the job to CNCF provided larger workers.

ahrtr (approved):
The main concern is maintainers' and some contributors' bandwidth. Hopefully it will **NOT** add too much additional overhead.

.github/workflows/linearizability-nightly.yaml
tjungblu: will add on the next PR for testing the release branches

PR #14917, opened 2022-12-09
tests: Increase test timeout for nighly runs to match job timeout min…

serathius:
cc @ahrtr @tjungblu


ahrtr (approved):
LGTM

Why not 180m (3h)?

serathius:
2 reasons:
* We need time to build the binaries.
* Prefer that tests timeout before the worker. This should give us full report of stacktraces so we could find where code got stuck (if it did).

PR #14922, opened 2022-12-10
resolver: cleanup of deprecated grpc resolver target.Endpoint field

ramil600:
`grpc.resolver.Target.Endpoint` and some other fields are deprecated, URL field is suggested to use instead.
`URL.Path` is required to be stripped of "/" prefix for naming/resolver to work properly

### DIFF
--- a/client/v3/naming/resolver/resolver.go
+++ b/client/v3/naming/resolver/resolver.go
@@ -16,6 +16,7 @@ package resolver
 
 import (
 	"context"
+	"strings"
 	"sync"
 
 	clientv3 "go.etcd.io/etcd/client/v3"
@@ -31,9 +32,15 @@ type builder struct {
 }
 
 func (b builder) Build(target gresolver.Target, cc gresolver.ClientConn, opts gresolver.BuildOptions) (gresolver.Resolver, error) {
+	// Refer to https://github.com/grpc/grpc-go/blob/16d3df80f029f57cff5458f1d6da6aedbc23545d/clientconn.go#L1587-L1611
+	endpoint := target.URL.Path
+	if endpoint == "" {
+		endpoint = target.URL.Opaque
+	}
+	endpoint = strings.TrimPrefix(endpoint, "/")
 	r := &resolver{
 		c:      b.c,
-		target: target.Endpoint,
+		target: endpoint,
 		cc:     cc,
 	}
 	r.ctx, r.cancel = context.WithCancel(context.Background())


client/v3/naming/resolver/resolver.go
```diff
@@ -33,7 +34,7 @@ type builder struct {
 func (b builder) Build(target gresolver.Target, cc gresolver.ClientConn, opts gresolver.BuildOptions) (gresolver.Resolver, error) {
 	r := &resolver{
 		c:      b.c,
-		target: target.Endpoint,
+		target: strings.TrimPrefix(target.URL.Path, "/"),
```
fuweid, line 37: side note: https://github.com/grpc/grpc-go/blob/16d3df80f029f57cff5458f1d6da6aedbc23545d/clientconn.go#L1587-L1611

fuweid (approved):
LGTM(non-binding)

>  TestEtcdGrpcResolver already test it.

client/v3/naming/resolver/resolver.go
ahrtr: The comment says `Deprecated: use URL.Path or URL.Opaque instead.`. Should we consider the `URL.Opaque` as well? Shouldn't the logic something like below? 

```
	target := target.URL.Path
	if target == "" {
		target = target.URL.Opaque
	}
	target = strings.TrimPrefix(target, "/")
```

ramil600: Hi @ahrtr,
Done.
 I looked into your query: indeed if we declare URL with no '/' after the scheme, endpoint can be written to URL.Opaque.

client/v3/naming/resolver/resolver.go
```diff
@@ -31,9 +32,14 @@ type builder struct {
 }
 
 func (b builder) Build(target gresolver.Target, cc gresolver.ClientConn, opts gresolver.BuildOptions) (gresolver.Resolver, error) {
+	endpoint := target.URL.Path
```
ahrtr, line 35: Could you add a comment above line 35 something like below?

```
Refer to https://github.com/grpc/grpc-go/blob/16d3df80f029f57cff5458f1d6da6aedbc23545d/clientconn.go#L1587-L1611
```

ramil600: Done, thank you for the review.

ahrtr (approved):
LGTM

Thank you @ramil600

fuweid (approved):
LGTM（non-binding)

PR #14935, opened 2022-12-12
client: enhance the function shouldRetryWatch and added unit test

ahrtr:
Resolve the comment https://github.com/kubernetes/kubernetes/pull/114403#discussion_r1045906333 

cc @mitake @spzala

### DIFF
--- a/client/v3/watch.go
+++ b/client/v3/watch.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
@@ -588,8 +587,7 @@ func (w *watchGrpcStream) run() {
 
 			switch {
 			case pbresp.Created:
-				cancelReasonError := v3rpc.Error(errors.New(pbresp.CancelReason))
-				if shouldRetryWatch(cancelReasonError) {
+				if shouldRetryWatch(pbresp.CancelReason) {
 					var newErr error
 					if wc, newErr = w.newWatchClient(); newErr != nil {
 						w.lg.Error("failed to create a new watch client", zap.Error(newErr))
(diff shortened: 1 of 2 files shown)


client/v3/watch.go
```diff
@@ -717,9 +716,9 @@ func (w *watchGrpcStream) run() {
 	}
 }
 
-func shouldRetryWatch(cancelReasonError error) bool {
-	return (strings.Compare(cancelReasonError.Error(), v3rpc.ErrGRPCInvalidAuthToken.Error()) == 0) ||
-		(strings.Compare(cancelReasonError.Error(), v3rpc.ErrGRPCAuthOldRevision.Error()) == 0)
+func shouldRetryWatch(cancelReason string) bool {
```
liggitt, line 719: if cancelReason is empty, should this return false early?

client/v3/watch.go
```diff
@@ -717,9 +716,9 @@ func (w *watchGrpcStream) run() {
...
+	return (strings.Compare(cancelReason, v3rpc.ErrGRPCInvalidAuthToken.Error()) == 0) ||
```
liggitt, line 720: is `v3rpc.ErrGRPCInvalidAuthToken.Error()` constant? can we compute this once as a local package var? same question for `v3rpc.ErrGRPCAuthOldRevision.Error()`

liggitt, line 720: can we use `==` instead of strings.Compare since we only care about equality and not ordering

client/v3/watch.go
ahrtr: Makes sense, it should have very little improvement on the performance

client/v3/watch.go
ahrtr: Resolved & thx

ahrtr: Resolved & thx

mitake (approved):
LGTM, thanks for enhancing it @ahrtr

client/v3/watch_test.go
```diff
@@ -53,3 +56,39 @@ func TestEvent(t *testing.T) {
...
+		{
+			name:          "valid grpc error but not equal to ErrGRPCInvalidAuthToken or ErrGRPCAuthOldRevision",
+			msg:           rpctypes.ErrGRPCUserEmpty.Error(),
+			expectedRetry: false,
+		},
+		{
+			name:          "invalid grpc error and not equal to ErrGRPCInvalidAuthToken or ErrGRPCAuthOldRevision",
+			msg:           "whatever error message",
+			expectedRetry: false,
+		},
```
ptabor, line 85: Now msg: "" is a special codepath - so probably worth having a testcase.

ahrtr: Eventually we need to revert the change and need to add a field something like `ReasonCode` in the `WatchResponse`. Or just rollback the change, do not fix it at all. What's your thought? 

https://github.com/etcd-io/etcd/discussions/14992

PR #14993, opened 2022-12-14
clientv3/concurrency: add logger to session, add unit test

ramil600:

### DIFF
--- a/tests/integration/clientv3/concurrency/session_test.go
+++ b/tests/integration/clientv3/concurrency/session_test.go
@@ -1,0 +1,84 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -0,0 +1,84 @@
...
+		t.Log(err)
+	}
+	if resp.TTL == -1 {
+		t.Errorf("client lease should not be expired: %d", resp.TTL)
+
+	}
+	if resp.TTL == 60 {
+		t.Errorf("default TTL value is used in the session, instead of set TTL: %d", setTTL)
+	}
+	if resp.TTL >= int64(setTTL) || resp.TTL < int64(setTTL)-5 {
```
ptabor, line 80: When many tests are running concurrently, the responsivness is really low... and it manifests as test flakiness. 

I assume we don't loose any coverage if we change it to -20.

ptabor (approved):

tests/integration/clientv3/concurrency/session_test.go
ramil600: @ptabor, thank you for reviewing. Done, changed the delay to 20sec and TTL to 90sec

PR #14547, opened 2022-10-03
Backport PR 14322 to release-3.5

mitake:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

This PR backports https://github.com/etcd-io/etcd/pull/14322 to `release-3.5`. cc @ahrtr @kafuu-chino

PR #14995, opened 2022-12-14
clientv3: revert the client side change in 14547

ahrtr:
In order to fix https://github.com/etcd-io/etcd/issues/12385, PR https://github.com/etcd-io/etcd/pull/14322 introduced a change in which the client side may retry based on the error message returned from server side.

This is not good, as it's too fragile and it's also changed the protocol between client and server. Please see the discussion in https://github.com/kubernetes/kubernetes/pull/114403

Note: The issue https://github.com/etcd-io/etcd/issues/12385 only happens when auth is enabled, and client side reuse the same client to watch.

So we decided to rollback the change on 3.5, reasons: 
1. K8s doesn't enable auth at all. It has no any impact on K8s. 
2. It's very easy for client application to workaround the issue.
  The client just needs to create a new client each time before watching.

cc  @mitake @ptabor @serathius @spzala @liggitt @dims @lavalamp @deads2k 

### DIFF
--- a/client/v3/watch.go
+++ b/client/v3/watch.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ahrtr:
Reference: https://github.com/etcd-io/etcd/pull/14547

dims:
yes please. thanks!

mitake (approved):
LGTM, sorry for the issue...

ahrtr:
Thanks @mitake for the quick review.

jeansimonbarry:
@ahrtr Sorry but I just want to confirm, as we are seeing the same issue in our cluster where we enabled auth. Is the correct way to fix this by instead creating a new client for every Watch call made?

ahrtr:
> Is the correct way to fix this by instead creating a new client for every Watch call made?

Yes, it should can workaround this issue. 



> Or I guess we can upgrade to 3.5.6 which does have the fix?

3.5.6 has this fix, but we reverted the change in 3.5.7( to be  released soon). 

We are working on a more formal solution in https://github.com/etcd-io/etcd/issues/15058. I am not sure whether we will backport the fix to 3.5 eventually for now. If lots of people request backporting, we can do it.

jeansimonbarry:
Awesome thanks so much 👍

PR #15018, opened 2022-12-19
[3.5] Security: address HIGH Vulnerabilities

ahrtr:
Firstly, bumped `golang.org/x/net `to v0.4.0 to address some HIGH CVEs. 

Secondly, `golang.org/x/sys` and `golang.org/x/text` are also automatically bumped when bumping `golang.org/x/net`. They requires Go 1.17+. So let's bump go version to 1.17.13.


ahrtr:
Please ignore the `Release` failure for now. Once this PR gets merged, it should can be re solved automatically.

.github/workflows/e2e.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
       with:
-        go-version: "1.16.15"
```
fuweid, line 16: Right now, the 1.17.x is unsupported. 
Just wonder that it will be great to use current stable version like 1.18.x or 1.19.x.

ptabor:
Thinking what consequences for k8s would have bumping the go version: 

- k8s [1.22](https://github.com/kubernetes/kubernetes/blob/release-1.22/go.mod),[1.23](https://github.com/kubernetes/kubernetes/blob/release-1.23/go.mod)   uses golang 1.16 and etcd-3.5.0 
(so didn't bother to patch with 3.5.[5-6])
- k8s [1.24](https://github.com/kubernetes/kubernetes/blob/release-1.24/go.mod), go 1.18, etcd-3.5.1
- k8s [1.25](https://github.com/kubernetes/kubernetes/blob/release-1.25/go.mod), go 1.19, etcd-3.5.4
- k8s [1.26](https://github.com/kubernetes/kubernetes/blob/release-1.26/go.mod), go 1.19, etcd-3.5.5

As etcd-3.5 will be live for a long time, I think we bump the version and let customer's (like k8s) decide whether they want to cary the CVE risk or not. Etcd version for k8s is not critical as the all http traffic is shielded through API server (assuming proper configuration of firewalls).

.github/workflows/e2e.yaml
ptabor: I would minimize the changes (so keep 1.17), as we are already breaking the usual promise of the patch versions (due to etcd long lifecycle).

ptabor (approved):
Please add a changelog entry in the 'breaking' section.

.github/workflows/e2e.yaml
ahrtr: > minimize the changes

This was exactly what I was thinking. In the beginning, I bumped to go1.19.4 directly, but after second thought, I downgrade it to 1.17.13. 

If all maintainers agree to bump 1.19.x, we can do it in a separate PR.

PR #15021, opened 2022-12-19
clientv3/naming/endpoints: fix endpoints prefix bug

ramil600:
fixes bug with multiple endpoints with same prefix, 
any endpoint manager with the same prefix will try to list endpoints,
 get updates from other one unintentionally.


ptabor:
Thank you Ramil, 
It's a valid use-case. But I'm also afraid that it might be a breaking change for all that depend on the old semantic (just prefix matching).

Do you know whether it was introduced by the rewrite to new grpc API in etcd v-3.5 or this behavior existed forever ?

tests/integration/clientv3/naming/resolver_test.go
```diff
@@ -111,3 +111,55 @@ func TestEtcdGrpcResolver(t *testing.T) {
 		break
 	}
 }
+
+func TestEtcdEndpointManager(t *testing.T) {
```
ahrtr, line 115: Please use `github.com/stretchr/testify` to simplify the test case.

ahrtr:
The PR looks good to me. Each key must be prefixed with `m.target+"/"` per current implementation.

https://github.com/etcd-io/etcd/blob/054b24b42536bdacb9d33acf196637c586b8c2d2/client/v3/naming/endpoints/endpoints_impl.go#L59

ramil600:
> Thank you Ramil, It's a valid use-case. But I'm also afraid that it might be a breaking change for all that depend on the old semantic (just prefix matching).
> 
> Do you know whether it was introduced by the rewrite to new grpc API in etcd v-3.5 or this behavior existed forever ?

Hello Piotr,

currently resolver is  being used for grpc proxy services discovery, where you can start a new cluster and optionally provide a prefix to register/publish peer grpc-proxy servers for high availability:
[https://github.com/etcd-io/etcd/blob/6d8139dcea972ca0cb3ab1fd9cc22c959bcf930a/server/proxy/grpcproxy/cluster.go#L50-L64](url)

Then you have the option to provide prefix to register new grpc proxy server in the cluster  with given resolver prefix. The way that it adds itself with {prefix/addr} line it will not break the logic you are enquiring about:
[https://github.com/etcd-io/etcd/blob/6d8139dcea972ca0cb3ab1fd9cc22c959bcf930a/server/proxy/grpcproxy/register.go#L74-L76
](url)

Here is also the link to the usage guide:
[https://etcd.io/docs/v3.6/op-guide/grpc_proxy/#client-endpoint-synchronization-and-name-resolution](url)

I will also look into previous implementation later.

tests/integration/clientv3/naming/resolver_test.go
ramil600: Hello Benjamin,

thank you for reviewing, added asserts for the test.

ptabor (approved):

ptabor:
@ramil600 Thank you for checking.

ahrtr (approved):
LGTM

Thank you @ramil600

PR #12165, opened 2020-07-24
clientv3:get AuthToken gracefully without extra connection.

cfc4n:
clientv3: get AuthToken gracefully without extra connection.

ignore AuthToken check when InternalRequest was `AuthenticateRequest` also `etcdserverpb.Auth/Authenticate`.

PR #15032, opened 2022-12-20
refactor(client): remove duplicate processing

wafuwafu13:
Signed-off-by: wafuwafu13 <mariobaske@i.softbank.jp>

- as [this comment](https://github.com/etcd-io/etcd/blob/dc680e3580509b9a90d786c658030dde67f7d39f/client/v3/retry_interceptor.go#L114) say, [`c.authTokenBundle != nil` is equal to `c.Username != "" && c.Password != ""`](https://github.com/etcd-io/etcd/blob/dc680e3580509b9a90d786c658030dde67f7d39f/client/v3/client.go#L387-L391), and `c.getToken` just return `nil` [if either `c.Username` or  `c.Password` is `""`](https://github.com/etcd-io/etcd/blob/dc680e3580509b9a90d786c658030dde67f7d39f/client/v3/client.go#L264-L266). so `if c.authTokenBundle != nil` is unnecessary.
- `getToken` returns `nil` [if `err == rpctypes.ErrAuthNotEnabled`](https://github.com/etcd-io/etcd/blob/dc680e3580509b9a90d786c658030dde67f7d39f/client/v3/client.go#L270-L273), and [rpctypes.Error(err) != rpctypes.ErrAuthNotEnabled](https://github.com/etcd-io/etcd/blob/dc680e3580509b9a90d786c658030dde67f7d39f/client/v3/retry_interceptor.go#L116) is always `true`, so unnecessary.

ref: `// TODO(cfc4n): keep this code block, remove codes about getToken in client.go after pr #12165 merged.` https://github.com/etcd-io/etcd/pull/12165/files


ahrtr (approved):
LGTM

THank you @wafuwafu13

client/v3/retry_interceptor.go
```diff
@@ -109,14 +109,10 @@ func (c *Client) streamClientInterceptor(optFuncs ...retryOption) grpc.StreamCli
...
 		// getToken automatically
-		// TODO(cfc4n): keep this code block, remove codes about getToken in client.go after pr #12165 merged.
-		if c.authTokenBundle != nil {
-			// equal to c.Username != "" && c.Password != ""
-			err := c.getToken(ctx)
-			if err != nil && rpctypes.Error(err) != rpctypes.ErrAuthNotEnabled {
-				c.GetLogger().Error("clientv3/retry_interceptor: getToken failed", zap.Error(err))
-				return nil, err
-			}
+		err := c.getToken(ctx)
```
ptabor, line 112: I'm a little confused. Are we getting a new token for each creation of a stream... and getting the token requires a separate call to etcd ?  I thought we use token stored in the authTokenBundle... as long as we don't get error suggesting `shouldRefreshToken`... 

I know it's not change in this PR, but still if you know, please let me know.

wafuwafu13: I didn't understand why `getToken` needed to be called, but since it was [mentioned in the comment](https://github.com/etcd-io/etcd/blob/7d5151532cf7533198f3f3067152d0fde8efb8f4/client/v3/retry_interceptor.go#L111), I didn't remove it.

But I think that `getToken` is not required on the first run of Interceptor because [getToken is called at client initialization](https://github.com/etcd-io/etcd/blob/7d5151532cf7533198f3f3067152d0fde8efb8f4/client/v3/client.go#L441) and token is stored in the authTokenBundle.

ahrtr: Per my understanding `streamClientInterceptor` is just a client side stream Interceptor, but etcd only uses server side streaming (watch stream and lease keep alive), so `streamClientInterceptor` is only executed once on the very beginning of a stream.  Note that for server-side streaming, the client side only calls `SendMsg` once.

So actually the token will never be refreshed during a watch stream, and it's exactly the reason of https://github.com/etcd-io/etcd/issues/12385, which was resolved in https://github.com/etcd-io/etcd/pull/14322. But unfortunately, we rollbacked the fix due to some concern on K8s side.  FYI. https://github.com/etcd-io/etcd/discussions/14992

ahrtr: Update: `lease keep alive` should be a bi-directional stream. The client side keeps sending `LeaseKeepAliveRequest`, and server side responds with a `LeaseKeepAliveResponse` each time.

wafuwafu13: Thank you for letting me know. I added comment.

wafuwafu13: @ahrtr @ptabor CI passed and I have nothing to change in this PR :)

ahrtr: @wafuwafu13 Please squash the commits and rebase this PR.

ahrtr:
The second commit https://github.com/etcd-io/etcd/pull/15032/commits/b606721a49a734a40f2c1b056caa30a4a192fd5a doesn't make any sense. Please rebase this PR instead of merging `main`

wafuwafu13:
I made mistakes and will remake it.

PR #15013, opened 2022-12-17
tests: add mix version for snapshot e2e case

halegreen:
fix issue #14807
Signed-off-by: WangXiaoxiao <1141195807@qq.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Issue #14807, opened 2022-11-18
Add mix-version testing for sending snaspshot cases


ahrtr:
### What would you like to be added?

We already have mix-version (e.g, A cluster has both 3.5 node and 3.6 node) test, but do not cover the cases of sending snapshot. Specifically we need to cover the following two cases,
1. 3.5 node sends a snapshot to 3.6 node;
2. 3.6 node sends a snapshot to 3.5 node;

Note that etcd doesn't support adding a new member of old version into a cluster with higher version. For example, etcd cluster version is 3.6.x, then a new member of 3.5.x can't join the cluster. Please refer to [cluster_util.go#L222-L230](https://github.com/etcd-io/etcd/blob/3e903d0b12e399519a4013c52d4635ec8bdd6863/server/etcdserver/cluster_util.go#L222-L230).  So we can't implement case 2 by dynamically adding a node with older version, please refer to https://github.com/etcd-io/etcd/pull/14707.  Instead, we can use network partition scenario to trigger snapshots.

### Why is this needed?

We need to make sure there is no any issue when upgrading multi-node etcd cluster. Also refer to https://github.com/etcd-io/etcd/pull/14592

ahrtr:
Reference: [cluster.go#L1343](https://github.com/etcd-io/etcd/blob/07a3cfe22eaa96c34898922ddd27a2354bf9459a/tests/framework/integration/cluster.go#L1343)

halegreen:
@ahrtr Hi , I want to take this. While I've not seen the `injectPartition` method for e2e test cases., only in integration test case. So I plan to modify iptables,  dorp peer port transporting , in order to mock network partition for e2e cluster , should it be OK ?

ahrtr:
Thanks @halegreen for working on this. `Programmatically modifying iptables` seems not a good solution, reasons:
1. It might have impact on other test case;
2. It isn't good to update any system-wide setting even you programmatically rollback the change afterwards. Note we may manually run any e2e test case locally, so it may affect developers' local environment.

A simpler & safer solution is to stop a member, play traffic and start it again. @ptabor @serathius any thoughts?

halegreen:
@ahrtr "stop a member, play traffic and start it again", I was wonder will this action always leads to a network partition ?

ahrtr:
> hi, @ahrtr "stop a member, play traffic and start it again", I was wonder will this action always leads to a network partition ?

[...]

PR #2403, opened 2015-03-01
etcdserver: keep a min number of entries in memory

xiang90:
Do not aggressively compact the raft log entries. After a snapshot,
etcd server can compact the raft log upto snapshot index. etcd server
compacts to an index smaller index to keep some entries in memory.
The leader can still read out the in memory entries to send to a slightly
slow follower. If all the entries are compacted, the leader will send the
whole snapshot or read entries from disk if possible.

PR #15033, opened 2022-12-20
externalize snapshot catchup entries to etcd flag

chaochn47:

Motivation
1. It can benefit one-node cluster with limited resource (e.g memory) without any performance impact. 
2. For multi-node cluster, setting a small value for DefaultSnapshotCatchUpEntries can reduce the memory usage while less tolerant of network events that will cause snapshot transfer (reduced availability).
3. It can benefit the E2E test cases to speed up test time and improve development velocity.

Questions
1. open to discussion if we want to mark this flag as `experimental`.
2. should we set a lower bound limit on the snapshot catchup entries count? 

Related https://github.com/etcd-io/etcd/pull/14819 and https://github.com/etcd-io/etcd/pull/15013#discussion_r1051683699

### DIFF
--- a/server/etcdmain/config.go
+++ b/server/etcdmain/config.go
@@ -279,6 +279,7 @@ func newConfig() *config {
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
 	fs.IntVar(&cfg.ec.ExperimentalMaxLearners, "experimental-max-learners", membership.DefaultMaxLearners, "Sets the maximum number of learners that can be available in the cluster membership.")
 	fs.DurationVar(&cfg.ec.ExperimentalWaitClusterReadyTimeout, "experimental-wait-cluster-ready-timeout", cfg.ec.ExperimentalWaitClusterReadyTimeout, "Maximum duration to wait for the cluster to be ready.")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "Number of entries for a slow follower to catch up after compacting the the raft storage entries.")
 
 	// unsafe
 	fs.BoolVar(&cfg.ec.UnsafeNoFsync, "unsafe-no-fsync", false, "Disables fsync, unsafe, will cause data loss.")
--- a/server/etcdmain/help.go
+++ b/server/etcdmain/help.go
@@ -269,6 +269,8 @@ Experimental feature:
     Set the max number of learner members allowed in the cluster membership.
   --experimental-wait-cluster-ready-timeout '5s'
     Set the maximum time duration to wait for the cluster to be ready.
+  --experimental-snapshot-catch-up-entries '5000'
+    Number of entries for a slow follower to catch up after compacting the the raft storage entries.
 
 Unsafe feature:
   --force-new-cluster 'false'
--- a/tests/e2e/etcd_config_test.go
+++ b/tests/e2e/etcd_config_test.go
@@ -20,10 +20,14 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
+	"golang.org/x/sync/errgroup"
+
 	"go.etcd.io/etcd/pkg/v3/expect"
+	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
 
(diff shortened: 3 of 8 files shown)


chaochn47:
@ahrtr @lavacat Could you please take a look and provide guidance? I am not sure about the answer of the above two questions. 

/cc @serathius

lavacat (changes requested):
LGTM, but please update changelog.

lavacat:
I think this setting can be qualified as stable [feature](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/features.md). So, it's ok to drop experimental. When it was added, there was a comment about making it tunable https://github.com/etcd-io/etcd/pull/2403#issuecomment-76610454

ahrtr:
> 1.open to discussion if we want to mark this flag as experimental.

YES, a new flag usually is experimental. See [features.md#experimental](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/features.md#experimental). It isn't super clear what's the exact impact for now. So keeping it experimental is a better choice.

> 2.should we set a lower bound limit on the snapshot catchup entries count?

I would expect NO for now. For single node cluster for edge case, I even want to set a very small value, e.g. 10.  But we need to provide general guidance on this, usually it should be `expected latency` * `max throughput` for multi-node cluster.

cc  @ptabor & @spzala for opinion as well.

server/etcdmain/config.go
```diff
@@ -125,6 +125,7 @@ func newConfig() *config {
 	fs.UintVar(&cfg.ec.MaxWalFiles, "max-wals", cfg.ec.MaxWalFiles, "Maximum number of wal files to retain (0 is unlimited).")
 	fs.StringVar(&cfg.ec.Name, "name", cfg.ec.Name, "Human-readable name for this member.")
 	fs.Uint64Var(&cfg.ec.SnapshotCount, "snapshot-count", cfg.ec.SnapshotCount, "Number of committed transactions to trigger a snapshot to disk.")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "Number of entries for a slow follower to catch up after compacting the the raft storage entries.")
```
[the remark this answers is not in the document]
chaochn47: Agree it also means how many raft log entries to retain in memory for each raft log compaction.

server/etcdmain/help.go
```diff
@@ -55,6 +55,8 @@ Member:
     Path to the dedicated wal directory.
   --snapshot-count '100000'
     Number of committed transactions to trigger a snapshot to disk.
+  --snapshot-catch-up-entries '5000'
```
ahrtr, line 58: And please put it together with all other experimental flags,

```suggestion
  --experimental-snapshot-catch-up-entries '5000'
```

tests/e2e/etcd_config_test.go
```diff
@@ -358,4 +359,23 @@ func TestBootstrapDefragFlag(t *testing.T) {
...
+	proc.Close()
+}
+
+func TestSnapshotCatchupEntriesFlag(t *testing.T) {
+	e2e.SkipInShortMode(t)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	proc, err := e2e.SpawnCmd([]string{e2e.BinPath.Etcd, "--snapshot-catchup-entries", "1000"}, nil)
```
ahrtr, line 373:
```suggestion
	proc, err := e2e.SpawnCmd([]string{e2e.BinPath.Etcd, "--experimental-snapshot-catchup-entries", "1000"}, nil)
```

ahrtr (approved):
LGTM

Thank you @chaochn47

lavacat (approved):

tests/e2e/etcd_config_test.go
```diff
@@ -358,4 +359,23 @@ func TestBootstrapDefragFlag(t *testing.T) {
...
+	e2e.SkipInShortMode(t)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	proc, err := e2e.SpawnCmd([]string{e2e.BinPath.Etcd, "--experimental-snapshot-catchup-entries", "1000"}, nil)
+	require.NoError(t, err)
+	require.NoError(t, e2e.WaitReadyExpectProc(ctx, proc, []string{"\"snapshot-catchup-entries\":1000"}))
+	require.NoError(t, e2e.WaitReadyExpectProc(ctx, proc, []string{"serving client traffic"}))
+	require.NoError(t, proc.Stop())
```
fuweid, line 377: (just side note) currently, the proc.Stop just sends SIGTERM signal to process. basically, it won't return error.
The `proc.Close` might return error if there is somewrong in server.

ptabor (approved):

ptabor (commented):
- Shell we have a test that checks whether etcd instance works in stable way when the setting is unreasonably small (e.g. 1) ? 

- Do we disallow setting this to 0 or negative ?

ahrtr:
> * Do we disallow setting this to 0 or negative ?

Negative value is definitely NOT possible, because its type is `uint64`. `0` is the default value, and etcd will use the default value `5000` if it's 0.



> * Shell we have a test that checks whether etcd instance works in stable way when the setting is unreasonably small (e.g. 1) ?

Agreed. @chaochn47 please add test cases to intentionally to set a small value (e.g 1, 10). Please also resolve comment https://github.com/etcd-io/etcd/pull/15033#discussion_r1054134913, and also rebase this PR.

tests/e2e/etcd_config_test.go
chaochn47: Yeah, sending SIGTERM to the process will lead to the exit code to be `-1`

https://github.com/etcd-io/etcd/blob/5b163aa50752c08705a0e8d8b6a6b36beb054b16/pkg/expect/expect.go#L280-L282

```
ExitCode returns the exit code of the exited process, or -1 if the process hasn't exited or was terminated by a signal. 
```

`proc.Stop` can be potentially improved to include cleanly wait until process exited.

chaochn47:
Added a test to set snapshot catchup entry to 1 and validate 3 node etcd cluster is healthy under put traffic. 

Is it sufficient as an experimental flag?

ahrtr:
LGTM, thank you. Merging...

halegreen:
@chaochn47 hi, will this change be picked to v3.5 too?

chaochn47:
Hi @halegreen

I am afraid that this experimental flag will not be back-ported to 3.5 which is meant to be stable release. 

On the other way, 3.5 node cannot send snapshot to 3.6 node immediately without current PR. As an alternative pointed out in #14807, you may want to add a 3.6 node to a 3.5 cluster to unblock your test. 

/cc @ahrtr @serathius for thoughts

ahrtr:
No plan to backport this PR to 3.5. Please see https://github.com/etcd-io/etcd/pull/15013#discussion_r1063944346

PR #15044, opened 2022-12-24
Watch events during linearizability test and compare history

serathius:
cc @ahrtr @spzala @ptabor

### DIFF
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -16,7 +16,6 @@ package linearizability
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -25,6 +24,7 @@ import (
 	"time"
 
 	"github.com/anishathalye/porcupine"
+	"github.com/google/go-cmp/cmp"
 	"golang.org/x/time/rate"
 
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
@@ -107,29 +107,38 @@ func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProce
 		t.Fatal(err)
 	}
 	defer clus.Close()
-	ctx, cancel := context.WithCancel(ctx)
+	trafficCtx, trafficCancel := context.WithCancel(ctx)
 	go func() {
-		defer cancel()
-		err := triggerFailpoints(ctx, t, clus, failpoint)
-		if err != nil {
-			t.Error(err)
-		}
+		defer trafficCancel()
+		triggerFailpoints(ctx, t, clus, failpoint)
+		// Wait second to collect traffic after triggering last failpoint.
+		time.Sleep(time.Second)
 	}()
-	operations := simulateTraffic(ctx, t, clus, traffic)
+	watchCtx, watchCancel := context.WithCancel(ctx)
+	var operations []porcupine.Operation
+	go func() {
+		defer watchCancel()
+		operations = simulateTraffic(trafficCtx, t, clus, traffic)
+		// Wait second to collect watch events after all traffic was sent.
+		time.Sleep(time.Second)
+	}()
+	events := collectClusterWatchEvents(watchCtx, t, clus)
 	err = clus.Stop()
 	if err != nil {
 		t.Error(err)
 	}
+	validateEventsMatch(t, events)
 	checkOperationsAndPersistResults(t, operations, clus)
 }
 
-func triggerFailpoints(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, config FailpointConfig) error {
+func triggerFailpoints(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, config FailpointConfig) {
 	var err error
 	successes := 0
 	failures := 0
 	for _, proc := range clus.Procs {
 		if !config.failpoint.Available(proc) {
-			return fmt.Errorf("failpoint %q not available on %s", config.failpoint.Name(), proc.Config().Name)
+			t.Errorf("Failpoint %q not available on %s", config.failpoint.Name(), proc.Config().Name)
+			return
 		}
 	}
 	for successes < config.count && failures < config.retries {
@@ -143,10 +152,8 @@ func triggerFailpoints(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessC
 		successes++
 	}
 	if successes < config.count || failures >= config.retries {
-		return fmt.Errorf("failed to trigger failpoints enough times, err: %v", err)
+		t.Errorf("failed to trigger failpoints enough times, err: %v", err)
 	}
-	time.Sleep(config.waitBetweenTriggers)
-	return nil
 }
 
 type FailpointConfig struct {
@@ -201,6 +208,25 @@ type trafficConfig struct {
 	maximalQPS  float64
 	clientCount int
 	traffic     Traffic
+}
+
+func validateEventsMatch(t *testing.T, ops [][]watchEvent) {
+	// Move longest history to ops[0]
+	maxLength := len(ops[0])
+	for i := 1; i < len(ops); i++ {
+		if len(ops[i]) > maxLength {
+			maxLength = len(ops[i])
+			ops[0], ops[i] = ops[i], ops[0]
+		}
+	}
+
+	for i := 1; i < len(ops); i++ {
+		length := len(ops[i])
+		// We compare prefix of watch events, as we are not guaranteed to collect all events from each node.
+		if diff := cmp.Diff(ops[0][:length], ops[i][:length]); diff != "" {
+			t.Errorf("Events in watches do not match, %s", diff)
+		}
+	}
 }
 
 func checkOperationsAndPersistResults(t *testing.T, operations []porcupine.Operation, clus *e2e.EtcdProcessCluster) {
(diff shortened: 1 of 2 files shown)


serathius:
PTAL @ahrtr

tests/linearizability/linearizability_test.go
```diff
@@ -96,23 +97,31 @@ func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProce
...
-	operations := simulateTraffic(ctx, t, clus, traffic)
+	watchCtx, watchCancel := context.WithCancel(ctx)
+	var operations []porcupine.Operation
+	go func() {
+		defer watchCancel()
+		operations = simulateTraffic(trafficCtx, t, clus, traffic)
+		// Wait second to collect watch events after all traffic was sent.
+		time.Sleep(time.Second)
+	}()
+	events := watchClusterChanges(watchCtx, t, clus)
```
ahrtr, line 115: You'd better add comment to explain the relationship between three kinds of goroutines and how they are related by the cancel functions,
- The triggering failpoint goroutine
- The simulating traffic goroutine
- The watching event goroutines.

tests/linearizability/linearizability_test.go
```diff
@@ -96,23 +97,31 @@ func testLinearizability(ctx context.Context, t *testing.T, config e2e.EtcdProce
...
 	err = clus.Stop()
 	if err != nil {
 		t.Error(err)
 	}
+	validateEventsMatch(t, events)
```
ahrtr, line 120: Since all the watch events are cached, so I guess you will almost double the memory usage? Do you have any data related to the memory usage diff to share?

tests/linearizability/linearizability_test.go
serathius: Comments should explain "why?" and not "what?".  Isn't it clear in:
```
go func() {
		defer trafficCancel()
		triggerFailpoints(ctx, t, clus, failpoint)
		// Wait second to collect traffic after triggering last failpoint.
		time.Sleep(time.Second)
	}()
```
The failpoints gorouting cancels traffic?

tests/linearizability/linearizability_test.go
serathius: Don't understand the relevance. I didn't collect any data. What's the reason to collect memory usage here?

tests/linearizability/linearizability_test.go
ahrtr:
```
Isn't it clear
```

It's clear to you because you are the author. It doesn't mean it's clear to other contributors.  If I understood it correctly, the The **triggering failpoint goroutine** stops the **simulating traffic goroutine** by executing `defer trafficCancel()`, and the **simulating traffic goroutine** stops the **watching event goroutines** by **defer watchCancel()**. 

Note that it isn't Intuitive at first look. I was even thinking it might be a copy-paste typo, because I was confused why you use `ctx`, but defer cancel the `trafficCancel`. 

It isn't the first time I raise similar comment. Please always keep in mind to add as much self-contained comment as possible.

tests/linearizability/linearizability_test.go
ahrtr: See line 33 in watch.go, all watch events of all members are cached. It definitely causes more memory usage, and it's the reason why I raised this comment.  
```
memberEvents := make([][]watchEvent, len(clus.Procs))
```

serathius: Are you asking why we keep events from all members? It's to compare if the history matches.

tests/linearizability/linearizability_test.go
serathius: > It's clear to you because you are the author. It doesn't mean it's clear to other contributors.

This is a very vague generic statement. Code is clean or it's not. If concept of using context to propagate stop signal is not familiar or standard then let's rewrite it. It's not about throwing as many comments as possible as it doesn't solve the problem with code.

Moved the code to separate function added comment explaining idea to use context to propagate cancellation.

tests/linearizability/linearizability_test.go
ahrtr: No, I was concerning about the memory usage. Usually github runner has limited resource, I am not sure whether it will cause more flaky test.

If you do not see any issue after multiple runs, then I am OK.

serathius: I'm not worried as we can leverage bigger nodes provided by CNCF if needed.

serathius:
PTAL @ahrtr @ptabor

tests/linearizability/watch.go
```diff
@@ -0,0 +1,96 @@
...
+			mux.Lock()
+			memberEvents[i] = events
+			mux.Unlock()
+		}(i, c)
+	}
+	wg.Wait()
+	return memberEvents
+}
+
+func watchMemberEvents(ctx context.Context, c *clientv3.Client) []watchEvent {
```
ptabor, line 59: nit: maybe `collectAllWatchMemberEvents(...)` <- otherwise I assume there is asynchronous channel for the events.

tests/linearizability/watch.go
```diff
@@ -0,0 +1,96 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return events
+		default:
+		}
+		for resp := range c.Watch(ctx, "", clientv3.WithPrefix(), clientv3.WithRev(lastRevision)) {
+			lastRevision = resp.Header.Revision
+			for _, event := range resp.Events {
+				switch event.Type {
```
ptabor, line 71: Can You, please check/log or even panic on: 

```
	closeErr error

	// cancelReason is a reason of canceling watch
	cancelReason string
```
It should make debugging easier if the channel is closed in not the natural way.
?

tests/linearizability/linearizability_test.go
```diff
@@ -196,6 +204,18 @@ type trafficConfig struct {
 	traffic     Traffic
 }
 
+func validateEventsMatch(t *testing.T, ops [][]watchEvent) {
+	for i := 1; i < len(ops); i++ {
```
ptabor, line 208: Please add a comment describing when we expect different lengths... (one node off/partitioned during the test).

tests/linearizability/linearizability_test.go
```diff
@@ -196,6 +204,18 @@ type trafficConfig struct {
...
+		shorterLength := len(ops[0])
+		if len(ops[i]) < shorterLength {
```
ptabor, line 210: How about sorting ops by length DESC ? This way we could compare 2 longest streams .

Assuming we have watch streams of length: [0, 5, 7] we will not validate that 5 & 7 are matching in the current implementation.

tests/linearizability/linearizability_test.go
```diff
@@ -82,46 +82,56 @@ func TestLinearizability(t *testing.T) {
...
-	clus, err := e2e.NewEtcdProcessCluster(ctx, t, e2e.WithConfig(&config))
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer clus.Close()
-	ctx, cancel := context.WithCancel(ctx)
+func testLinearizability(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, failpoint FailpointConfig, traffic trafficConfig) (operations []porcupine.Operation, events [][]watchEvent) {
+	// Run multiple test components (traffic, failpoints, etc) in parallel and use canceling context to propagate stop signal.
+	wg := sync.WaitGroup{}
+	wg.Add(3)
```
ptabor, line 111: Wouldn't `errgroup` (https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go) 
be a little cleaner here: 
  - no need to remember about `wg.Done`
  - no need to proactively declare how many routines we will start...

tests/linearizability/watch.go
serathius: Done

tests/linearizability/watch.go
serathius: Done

tests/linearizability/linearizability_test.go
serathius: Done

tests/linearizability/linearizability_test.go
serathius: Good point. For simplicity I added comparing all streams.

serathius: Nice catch, forgot to fix change this. Moved longest history to 0 index.

tests/linearizability/linearizability_test.go
serathius: sg

tests/linearizability/watch.go
```diff
@@ -0,0 +1,97 @@
...
+	"go.uber.org/zap"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func collectClusterWatchEvents(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) [][]watchEvent {
+	mux := sync.Mutex{}
+	var wg sync.WaitGroup
```
ptabor, line 32: consider errgroup as well

ptabor (approved):

tests/linearizability/watch.go
serathius: Will follow this up

PR #15046, opened 2022-12-26
tests: Trigger only failpoints available in binary

serathius:
cc @ahrtr @spzala

### DIFF
--- a/tests/linearizability/failpoints.go
+++ b/tests/linearizability/failpoints.go
@@ -15,12 +15,9 @@
 package linearizability
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"math/rand"
-	"net/http"
-	"net/url"
 	"strings"
 	"testing"
 	"time"
@@ -83,6 +80,7 @@ var (
 type Failpoint interface {
 	Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error
 	Name() string
+	Available(e2e.EtcdProcess) bool
 }
 
 type killFailpoint struct{}
@@ -114,6 +112,10 @@ func (f killFailpoint) Name() string {
 	return "Kill"
 }
 
+func (f killFailpoint) Available(e2e.EtcdProcess) bool {
+	return true
+}
+
 type goPanicFailpoint struct {
 	failpoint string
 	trigger   func(ctx context.Context, member e2e.EtcdProcess) error
@@ -129,13 +131,12 @@ const (
 
 func (f goPanicFailpoint) Trigger(t *testing.T, ctx context.Context, clus *e2e.EtcdProcessCluster) error {
 	member := f.pickMember(t, clus)
-	address := fmt.Sprintf("127.0.0.1:%d", member.Config().GoFailPort)
 
 	triggerCtx, cancel := context.WithTimeout(ctx, triggerTimeout)
 	defer cancel()
 
 	for member.IsRunning() {
-		err := setupGoFailpoint(triggerCtx, address, f.failpoint, "panic")
+		err := member.Failpoints().Setup(triggerCtx, f.failpoint, "panic")
 		if err != nil {
 			t.Logf("gofailpoint setup failed: %v", err)
 		}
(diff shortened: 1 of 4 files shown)


tests/linearizability/failpoints.go
```diff
@@ -165,6 +171,42 @@ func (f goPanicFailpoint) pickMember(t *testing.T, clus *e2e.EtcdProcessCluster)
...
+func (f goPanicFailpoint) available(fs []string) bool {
+	for _, failpoint := range fs {
+		if f.failpoint == failpoint {
+			return true
+		}
+	}
+	return false
+}
+
+func listFailpoints(member e2e.EtcdProcess) ([]string, error) {
```
ahrtr, line 191: Please consider to get all available failpoints only once in `TestMain`. Obviously it doesn't make sense to trigger a http request on each test case.

ahrtr: We can only get the available failpoints after starting the cluster, so the other option is to cache the available failpoints so that they can be reused.

tests/linearizability/failpoints.go
```diff
@@ -239,11 +281,31 @@ type randomFailpoint struct {
...
+	if err != nil {
+		return err
+	}
+	availableFailpoints := make([]Failpoint, 0, len(f.failpoints))
+	for _, failpoint := range f.failpoints {
+		var available bool
+		if gofailpoint, ok := failpoint.(goPanicFailpoint); ok {
+			available = gofailpoint.available(gofailpointsAvailable)
+		} else {
+			available = failpoint.Available(clus.Procs[0])
```
ahrtr, line 294: get all the available failpoints (issue a http request) again in this case?

tests/linearizability/failpoints.go
serathius: its not a problem as we currently trigger one failpoint per cluster.

tests/linearizability/failpoints.go
serathius: Don't understand the question.

ahrtr: You already get all the available failpoints at line 284, and `(goPanicFailpoint) Available` gets all the failpoints again (line 175). 

Do not get the point why do you process `goPanicFailpoint` separately. Please add comment to explain the reason if you have a good one.

tests/linearizability/failpoints.go
ahrtr: I was talking about reuse the failpoints across test cases. But since it is possible that different test case might run test against different etcd versions, so it might not be correct.

Have you considered adding retry logic when getting an HTTP status code >= 500?

tests/linearizability/failpoints.go
serathius: Because each call to function `Available` makes http call to get list of failpoints. Instead of calling it for each go failpoint, I just get available failpoints once and call internal function `available` that takes a list of strings.

tests/linearizability/failpoints.go
serathius: For original question, each test should be independent so I don't want to reuse anything. 

As for checking 500, I prefer to check availability early and assume all errors are errors. I think this is less flake prone approach than trying to handle errors on request side.

tests/linearizability/failpoints.go
ahrtr: I still do not get the answer to my original question. Why do you get the list of failpoints again when it isn't `goPanicFailpoint` (line 294)?

tests/linearizability/failpoints.go
ahrtr: > I think this is less flake prone approach than trying to handle errors on request side.

Do not agree this. Since you already handle the retry mechanism at [linearizability_test.go#L88](https://github.com/etcd-io/etcd/blob/e73f55d4e94666c99558baa2fd4e365aeaca4dc4/tests/linearizability/linearizability_test.go#L88). So there is no need to retry in `listFailpoints`.

ptabor, line 191: shell it return map[string]struct{} -> so effectively a set for easier presence checking ?

ptabor, line 191: How about calling it `fetchFailpoints` to stress that it's a remote call ?

tests/linearizability/failpoints.go
ptabor: I think the question is syndrome of a wider issue: 
   - gofailpoints are subset of all 'failpoints'. There are not go failpoints (like network latency failpoints). 
   - the introduced mechanism takes care only of gofailpoints. 

1.  A binary is awere of all gofailpoints (panic / sleep) it exposes - i.e. lines where it can be triggered. 
2. A binary might expose some fail-points not related to specific line of code (e.g. FS access problems assuming file interface is proxied within the binary itself). 
3. There are failpoints provided externally (e.g. by proxy) that the binary is not aware of. 

Abstractly thinking:
  1.  Binary can expose `failpointManifests` that describes all failpoints it supports (internally). 
  2. ALL the endpoints should be able to take the failpoints manifests (some will ignore it) and say whether they are supported on the particular binary. 
  3. Failpoint manifest could get cached within https://pkg.go.dev/go.etcd.io/etcd/tests/v3/framework/e2e#EtcdProcess. 
such that public `Failpoint.Available(etcdProcess)` method could has as simple API as it currently has.

tests/linearizability/failpoints.go
serathius: Done

serathius:
PTAL @ahrtr @ptabor

ahrtr (approved):

PR #15055, opened 2022-12-29
etcd-dump-logs: Make the tests work again.

ptabor:
The tests are subtle as they skip if the binary is not generated in the local directory.

$ go build . && go test
PASS
ok      go.etcd.io/etcd/v3/tools/etcd-dump-logs 0.769s

Signed-off-by: Piotr Tabor <ptab@google.com>


ahrtr (approved):
LGTM

PR #15057, opened 2022-12-30
Add support for lease api to linearizability tests

geetasg:

### DIFF
--- a/tests/linearizability/lease_ids.go
+++ b/tests/linearizability/lease_ids.go
@@ -18,35 +18,36 @@ import (
 	"sync"
 )
 
-type leaseIdProvider interface {
+type clientId2LeaseIdMapper interface {
 	LeaseId(int) int64
 	AddLeaseId(int, int64)
-	RemoveLeaseId(int, int64)
+	RemoveLeaseId(int)
 }
 
-func newLeaseIdProvider() leaseIdProvider {
-	return &atomicLeaseIdProvider{m: map[int]int64{}}
+func newClientId2LeaseIdMapper() clientId2LeaseIdMapper {
+	return &atomicClientId2LeaseIdMapper{m: map[int]int64{}}
 }
 
-type atomicLeaseIdProvider struct {
+type atomicClientId2LeaseIdMapper struct {
 	sync.RWMutex
+	// m is used to store clientId to leaseId mapping.
 	m map[int]int64
 }
 
-func (lp *atomicLeaseIdProvider) LeaseId(clientId int) int64 {
-	lp.RLock()
-	defer lp.RUnlock()
-	return lp.m[clientId]
+func (lm *atomicClientId2LeaseIdMapper) LeaseId(clientId int) int64 {
+	lm.RLock()
+	defer lm.RUnlock()
+	return lm.m[clientId]
 }
 
-func (lp *atomicLeaseIdProvider) AddLeaseId(clientId int, leaseId int64) {
-	lp.Lock()
-	defer lp.Unlock()
-	lp.m[clientId] = leaseId
+func (lm *atomicClientId2LeaseIdMapper) AddLeaseId(clientId int, leaseId int64) {
+	lm.Lock()
+	defer lm.Unlock()
+	lm.m[clientId] = leaseId
 }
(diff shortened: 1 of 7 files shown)


geetasg:
If https://github.com/etcd-io/etcd/pull/15059/files is merged before this one, I will update this PR. 
 /cc @serathius @ahrtr

tests/linearizability/lease_ids.go
```diff
@@ -0,0 +1,52 @@
...
+	RemoveLeaseId(int, int64)
+}
+
+func newLeaseIdProvider() leaseIdProvider {
+	return &atomicLeaseIdProvider{m: map[int]int64{}}
+}
+
+type atomicLeaseIdProvider struct {
+	sync.RWMutex
+	m map[int]int64
```
ahrtr, line 33: Please add a comment something like: `clientId to LeaseId mapping`

tests/linearizability/lease_ids.go
```diff
@@ -0,0 +1,52 @@
...
+	return lp.m[clientId]
+}
+
+func (lp *atomicLeaseIdProvider) AddLeaseId(clientId int, leaseId int64) {
+	lp.Lock()
+	defer lp.Unlock()
+	lp.m[clientId] = leaseId
+}
+
+func (lp *atomicLeaseIdProvider) RemoveLeaseId(clientId int, leaseId int64) {
```
ahrtr, line 48: It seems that the second parameter `leaseId` isn't needed at all?

tests/linearizability/lease_ids.go
```diff
@@ -0,0 +1,52 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package linearizability
+
+import (
+	"sync"
+)
+
+type leaseIdProvider interface {
```
ahrtr, line 21: The name `*Provider` is a little misleading. Usually a Provider means it can provide or generate something. In this case, it looks like it's just used to cache the clientId to LeaseId mapping.  So how about renaming it to something like `clientId2LeaseIdMap`?

tests/linearizability/model.go
```diff
@@ -46,11 +50,17 @@ type EtcdResponse struct {
 	Err          error
 }
 
+type EtcdLease struct {
+	LeaseID int64
+	Keys    map[string]bool
```
ahrtr, line 55: What does the value (of type `bool`) mean?  If we don't care about the value at all, then I'd suggest to change it to  below,
```suggestion
	Keys    map[string]struct{}
```

tests/linearizability/traffic.go
```diff
@@ -87,6 +89,23 @@ func (t readWriteSingleKey) Write(ctx context.Context, c *recordingClient, limit
...
+		if leaseId != 0 {
+			err = c.PutWithLease(putCtx, key, newValue, leaseId)
+		}
+	case LeaseRevoke:
+		leaseId := lp.LeaseId(cid)
+		if leaseId != 0 {
+			err = c.LeaseRevoke(putCtx, leaseId)
+			if err == nil {
+				lp.RemoveLeaseId(cid, leaseId)
+			}
```
ahrtr, line 107: What if err isn't `nil`? Do you intentionally ignore the case (`err != nil`)?

tests/linearizability/lease_ids.go
geetasg: updated.

tests/linearizability/lease_ids.go
geetasg: updated

tests/linearizability/lease_ids.go
geetasg: in later revisions we may consider adding multiple leases per client. not needed as of now. updated.

tests/linearizability/model.go
geetasg: updated

geetasg:
Replaced by https://github.com/etcd-io/etcd/pull/15080. Review feedback from here is incorporated into the new PR along with refactoring from https://github.com/etcd-io/etcd/pull/15059/files

PR #15059, opened 2023-01-01
tests: Rewrite etcd requests to use operations

serathius:
cc @ahrtr @ptabor


tests/linearizability/model.go
```diff
@@ -17,33 +17,64 @@ package linearizability
...
+	Get    OperationType = "get"
+	Put    OperationType = "put"
+	Delete OperationType = "delete"
+	Txn    OperationType = "txn"
 )
 
 type EtcdRequest struct {
-	Op            Operation
+	Conds []EtcdCondition
+	Ops   []EtcdOperation
```
geetasg, line 36: can you please comment on when to expect multiple operations in a request? what is the main use case? thanks

serathius: Multi operational transactions. KV requests like put, get can be also modeled as transactions with one operation.

tests/linearizability/model_test.go
```diff
@@ -424,3 +426,60 @@ type testOperation struct {
...
+		},
+		{
+			req:            deleteRequest("key6"),
+			resp:           failedResponse(errors.New("failed")),
+			expectDescribe: `delete("key6") -> err: "failed"`,
+		},
+		{
+			req:            txnRequest("key7", "7", "77"),
+			resp:           txnResponse(false, 7),
+			expectDescribe: `if(key7=="7").then(put("key7", "77")) -> txn failed, rev: 7`,
```
ptabor, line 469: That's nice !!!

tests/linearizability/model.go
```diff
@@ -71,39 +102,79 @@ var etcdModel = porcupine.Model{
...
+
+	return fmt.Sprintf("%s -> %s", prefix, describeEtcdResponse(request.Ops, response))
+}
+
+func describeEtcdConditions(conds []EtcdCondition) string {
+	opsDescription := make([]string, len(conds))
+	for i := range conds {
+		opsDescription[i] = fmt.Sprintf("%s==%q", conds[i].Key, conds[i].ExpectedValue)
+	}
+	return strings.Join(opsDescription, ", ")
```
ptabor, line 123: If you go for golang/c++ like syntax, maybe && to express logical end is more natural

tests/linearizability/model.go
```diff
@@ -71,39 +102,79 @@ var etcdModel = porcupine.Model{
...
+		if resp.Value == "" {
+			return "nil"
+		}
+		return fmt.Sprintf("%q", resp.Value)
+	case Put:
+		return fmt.Sprintf("ok")
+	case Delete:
+		return fmt.Sprintf("deleted: %d", resp.Deleted)
+	default:
+		return "<invalid>"
```
ptabor, line 174: It might be misinterpreted as expected thing. 

`<! unknown op: %d !>`

tests/linearizability/model.go
ptabor: [note: k8s uses only one operation/put per transaction, but etcd supports multiple]

tests/linearizability/model.go
```diff
@@ -17,33 +17,64 @@ package linearizability
...
+	TxnFailure bool
+	Result     []EtcdOperationResult
+}
+
+type EtcdOperationResult struct {
+	Value   string
+	Deleted int64
+}
+
+func Match(r1, r2 EtcdResponse) bool {
```
ptabor, line 62: Why `Match` and not `Equals` ?

tests/linearizability/model.go
```diff
@@ -71,39 +102,79 @@ var etcdModel = porcupine.Model{
...
+
+func describeEtcdOperation(op EtcdOperation) string {
+	switch op.Type {
+	case Get:
+		return fmt.Sprintf("get(%q)", op.Key)
+	case Put:
+		return fmt.Sprintf("put(%q, %q)", op.Key, op.Value)
+	case Delete:
+		return fmt.Sprintf("delete(%q)", op.Key)
+	default:
```
ptabor, line 157: Maybe we need an explicit case for Txn (even to say that we explicitly don't support nested transactions in the framework).

tests/linearizability/model.go
```diff
@@ -167,33 +240,41 @@ func applyRequest(states PossibleStates, request EtcdRequest, response EtcdRespo
...
+		case Put:
+			s.KeyValues[op.Key] = op.Value
+			increaseRevision = true
+		case Delete:
+			if _, ok := s.KeyValues[op.Key]; ok {
+				delete(s.KeyValues, op.Key)
+				increaseRevision = true
+				opResp[i].Deleted = 1
+			}
+		default:
```
ptabor, line 272: ditto about Txn case (i.e. explicit panic with explaination)

ptabor (approved):

tests/linearizability/model.go
ahrtr: Can the function be as simple as 
```
return reflect.DeepEqual(r1, r2)
```

tests/linearizability/model.go
```diff
@@ -167,33 +240,41 @@ func applyRequest(states PossibleStates, request EtcdRequest, response EtcdRespo
 
 // applyRequestToSingleState handles a successful request, returning updated state and response it would generate.
 func applyRequestToSingleState(s EtcdState, request EtcdRequest) (EtcdState, EtcdResponse) {
+	success := true
+	for _, cond := range request.Conds {
+		if val := s.KeyValues[cond.Key]; val != cond.ExpectedValue {
+			success = false
```
ahrtr, line 246:
```suggestion
			success = false
			break
```

tests/linearizability/model.go
```diff
@@ -167,33 +240,41 @@ func applyRequest(states PossibleStates, request EtcdRequest, response EtcdRespo
...
-			resp.TxnSucceeded = true
+	opResp := make([]EtcdOperationResult, len(request.Ops))
+	increaseRevision := false
+	for i, op := range request.Ops {
+		switch op.Type {
+		case Get:
+			opResp[i].Value = s.KeyValues[op.Key]
+		case Put:
+			s.KeyValues[op.Key] = op.Value
+			increaseRevision = true
```
ahrtr, line 265: The slice `opResp` isn't updated at all in this case (`Put`), it might not be correct. Let's work with an example, assuming there are three operations in one TXN,
- read
- put
- delete

The length of `opResp` will be 3, but only the first and probably third elements are populated, the second element is still `nil`. Accordingly it will not match the actual response, see https://github.com/etcd-io/etcd/blob/3b623afddd27c419f00fb312128af466031b5549/tests/linearizability/model.go#L234

The reason why the workflow is green is there is only one operation in each request for now. When you start to support multiple operations, it might be failed.

ahrtr (approved):
Overall looks good to me with a couple of minor comments, which can be resolved in separate PRs.

tests/linearizability/model.go
serathius: Done

tests/linearizability/model.go
serathius: Done

tests/linearizability/model.go
serathius: Done. I plan to add unknown result in future but let's keep it simple for now.

tests/linearizability/model.go
serathius: Done

tests/linearizability/model.go
serathius: Done

tests/linearizability/model.go
serathius: Done.

tests/linearizability/model.go
serathius: It should be correct as Txn results are positional, elements are left empty intentionally.

PR #15060, opened 2023-01-02
Reimplement the tool `rw-heatmaps` using golang and rename it to `rw-benchmark`

ahrtr:
Points:
- The original [plot_data.py](https://github.com/etcd-io/etcd/blob/main/tools/rw-heatmaps/plot_data.py) is out of maintainance.
 - Remove the dependency on `python` and related python modules.
 - It isn't a best practice to compare benchmark results of two difference branches (e.g. main vs dev) in two charts. Instead, it's better to display the benchmarks to be compared in one chart.

Please see the example HTML page at [rw_benchmark.html](https://github.com/ahrtr/etcd/blob/go_chart_20230101/tools/rw-benchmark/example/rw_benchmark.html). One of the line charts is below,

etcd rw benchmark (RW Ratio 0 500000, Value Size 16)

The read QPS is displayed as <span style="color:blue">blue</span>, and write QPS is displayed as <span style="color:red">red</span>. The data in the second CSV file is rendered as dashed line if present.  Note each line in the line chart can be hidden or displayed by clicking on the related legend.

**Note**: although there are 2.5K+ lines of change, actually only `tools/rw-benchmark/plot_data.go`  and the `tools/rw-benchmark/README.md` need to be carefully reviewed.


tools/rw-benchmark/linkedmap.go
```diff
@@ -0,0 +1,310 @@
+// Copyright 2022 The etcd Authors
```
ptabor, line 1: Is there externally maintained representation that could be a replacement...

It's significant piece of generic code that would be good if we delegated elsewhere... 
Something from https://github.com/zyedidia/generic ?

If not - we should put it into ./pkg/linkedmap ...

tjungblu: I'm only briefly taking a look between meetings right now, but it seems to me that the LinkedMap is only an intermediary format between CSVs and float slices (`convertBenchmarkData`). Can't this be directly read into the slice format?

ahrtr: > Is there externally maintained representation that could be a replacement...
> 
> It's significant piece of generic code that would be good if we delegated elsewhere... Something from https://github.com/zyedidia/generic ?

Ah:), it's coming from my personal repo [linkedmap](https://github.com/ahrtr/gocontainer/tree/master/map/linkedmap). I also added [a comment](https://github.com/ahrtr/etcd/blob/c41fa9a04f4178d3da81a595ddc7208de7c60e40/tools/rw-benchmark/linkedmap.go#L38-L39) in this PR. In the beginning, I did introduce a dependency on [ahrtr/gocontainer/linkedmap]([linkedmap](https://github.com/ahrtr/gocontainer/tree/master/map/linkedmap); but on second thought, I don't think it's good to depend a personal repo, so I copied the file into etcd directly. 

If there is no objection, I can remove linkedmap.go (and linkedmap_test.go) and introduce dependency on [ahrtr/gocontainer/linkedmap]([linkedmap](https://github.com/ahrtr/gocontainer/tree/master/map/linkedmap). 



> Can't this be directly read into the slice format?

I intentionally did it in order to decouple the data parser and data render. It's possible that we may change the CSV data format, and in that case we do not need to change the data render part.

ahrtr: Anyway, I just removed `linkedmap.go` and `linkedmap_test.go`, and added dependency on [gocontainer#linkedMap](https://github.com/ahrtr/gocontainer#linkedMap). 

Note it will NOT populate the go.mod/go.sum files of any etcd users (e.g. clientv3 users, or embedded etcdserver users).

ahrtr: Please only pay attention to the latest two new added commits if you have already reviewed this PR. thx

ahrtr: added two more minor commits today.

ahrtr:
Rebased this PR.

ptabor:
I'm definitely for having this uniform in go. But I'm not convinced the new presentation form has sufficient information density. 

Previous representation has 3 advantages: 
  1. as a single generated picture could be easily pasted into PR discussion. Linking to HTML is difficult as you need to host them somewhere. I think it's very important from usability perspective.  
[the rest is aesthetics]
  2. it showed percentage advantage of one version of the other vs. the absolute values  [at least the right chart I was looking at usually]
  3. it was more clear that we compare the 'discrete' dimensions. The current lines chart creates impression we have continuous observations in the range of 32->4096 connections.  To some degree scatter plot (or at least marking the measured points would address that). But this also creates opportunity to make the charts more narrow. 
  4. [here I'm less sure] Having the charts more narrow (20 pixel * 8=160pixel) would allow to place multiple side charts in a single row (Value size is currently 7, but even if it grow to 15, it would be manageable 4k pixels).  Than we would encode 3 dimensions on something that can fit on a large screen instead of 'keep scrolling'. 

---
The PR seems to be very huge due to 'incremental' commits. I think there were not many comments so squashing it should not be net negative and more convencing to read it e2e.

ahrtr:
Thanks @ptabor for the feedback. Please see my response below,


> The PR seems to be very huge due to 'incremental' commits. I think there were not many comments so squashing it should not be net negative and more convencing to read it e2e.

I squashed the commits into one.  Although the PR looks huge (2K+ line of change), actually only `tools/rw-benchmark/plot_data.go` and `tools/rw-benchmark/README.md ` need to be carefully reviewed. The three files under directory `tools/rw-benchmark/example` are just example CSV files and a generated HTML page.

> 4\. [here I'm less sure] Having the charts more narrow (20 pixel * 8=160pixel) would allow to place multiple side charts in a single row (Value size is currently 7, but even if it grow to 15, it would be manageable 4k pixels).  Than we would encode 3 dimensions on something that can fit on a large screen instead of 'keep scrolling'.

I added more flags so that users can customize the charts, see usage below,
```
$ ./rw-benchmark  -h
rw-benchmark is a tool for visualize etcd read-write performance result.

Usage:
    rw-benchmark [options] result-file1.csv [result-file2.csv]

Additional options:
    -legend: Comma separated names of legends, such as "main,pr", defaults to "1" or "1,2" depending on the number of CSV files provided.
    -layout: The layout of the page, valid values: none, center and flex, defaults to "flex".
    -width: The width(pixel) of the each line chart, defaults to 600.
    -height: The height(pixel) of the each line chart, defaults to 300.
    -o: The HTML file name in which the benchmark data will be rendered, defaults to "rw_benchmark.html".
    -h: Print usage.
```

The example file `tools/rw-benchmark/example/rw_benchmark.html` was generated using command `go run plot_data.go ./example/main.csv  ./example/dev.csv`, see screen shot below,

Screen Shot 2023-01-12 at 15 16 20


> 
> * it showed percentage advantage of one version of the other vs. the absolute values  [at least the right chart I was looking at usually]
> 
> * it was more clear that we compare the 'discrete' dimensions. The current lines chart creates impression we have continuous observations in the range of 32->4096 connections.  To some degree scatter plot (or at least marking the measured points would address that). But this also creates opportunity to make the charts more narrow.

Actually I think the new version (delivered by this PR) is clearer and accordingly better to me. It gets data from both CSV files rendered in one chart, previously it was rendered in two separate charts. Previously it had to be `discrete` because it had three dimension in one chart (e.g. X axis is connections, Y (left) is value size, Y(right) is QPS), so there is no way to display a line for each data source. In the new version, we can render both line charts and display data as table directly. Please open and explore the example HTML file `tools/rw-benchmark/example/rw_benchmark.html`, you will love it.



> 1. as a single generated picture could be easily pasted into PR discussion. Linking to HTML is difficult as you need to host them somewhere. I think it's very important from usability perspective.
>        [the rest is aesthetics]

Yes, this is partially true to me. Indeed previously only two images were generated, but the data density is too big and I don't think it's clearer. I think many people only care about the rightest chart to get the overall comparison result. 

The new version (delivered in this PR) can also export images, but one image for one line chart. We can zip the HTML file and share it to github, and export a couple of images for better discussion in github. I agree this might be a little inconvinence, because users have to download the zip file and open the HTML file locally. 



> I'm definitely for having this uniform in go.

Yes, this is the main reason why I deliver this PR.

ahrtr:
R_W benchmark (RW Ratio 2, Value Size 2048)

The above example is exported from @tjungblu 's benchmark result in https://github.com/etcd-io/etcd/pull/14877. The blue lines are read QPS, while the red lines are write QPS. The solid lines are results tested against main branch, and the dash lines are tested against his PR. 

Note:
1. It's easy to hide or show each line by clicking right side legend;
2. It's easy to export an image for each line chart from the generated HTML page;
3. Users can also display the original data in a table by click the button on top right side of each line chart. 

We can also improve the way to plot the data in separate PR.

ahrtr:
I just removed the example (to be added in a separate PR) to make this PR smaller and therefore less scary.

ahrtr:
Is there any interest on this new tool? cc @mitake @ptabor @serathius @spzala

tools/rw-benchmark/plot_data.go
```diff
@@ -0,0 +1,498 @@
+// Copyright 2022 The etcd Authors
```
Elbehery, line 1:
```suggestion
// Copyright 2025 The etcd Authors
```

hwdef:
hey, do we still need this PR?

Issue #15062, opened 2023-01-05
`urlsEqual` might  wrongly skip resolving DNS


ahrtr:
### What happened?

See https://github.com/etcd-io/etcd/pull/13224#discussion_r1062216483

The correct logic should be:
1. If the URL doesn't contains wildcard, keep current behavior; otherwise follow https://github.com/etcd-io/etcd/pull/13224#discussion_r674022785

tjungblu:
Thanks Ben, I think we need to discuss what behavior we want to fix. Clearly this is a regression in 3.5 (and 3.4?), but implicitly blocking the bootstrap config on resolving a DNS entry in an equality check is also not a great idea. 

I would rather have the config.go explicitly lookup the important URLs and fail with a meaningful error message when they can't be resolved (akin to the init container the hypershift team added in https://github.com/openshift/hypershift/pull/1985). Then validate whether the URLs and their IP configs are consistent and equal. Potentially with a cmd flag to turn off that check.

#7798 does have some merit after all, there are implications on cluster membership as well.

ahrtr:
> we've seen troubles with wildcard dns names matching the server name early

In case I misunderstood anything, could you clarify this? How did `wildcard dns names matching the server name early` happen?  Do you have detailed steps to reproduce this issue?

tjungblu:
I only have fragments of the whole issue myself, I'll come back with a reproducer in a test case once I receive more information later today.

ahrtr:
It seems not related to https://github.com/etcd-io/etcd/pull/13224. The function `urlsEqual` just compares two slices of URL, and skip resolving DNS if they are deep equal. The logic seems simple and correct to me. 

Anyway, please clarify the issue and provide more detailed info firstly. FYI. https://github.com/ahrtr/etcd-issues/blob/master/docs/troubleshooting/sanity_check_and_collect_basic_info.md

tjungblu:
> The logic seems simple and correct to me.

The logic is not the issue, it's the ordering that breaks an implicit assumption users had about the startup of etcd.

tjungblu:
Here's how the team runs etcd inside k8s:

```
 /usr/bin/etcd
        --data-dir=/var/lib/data
        --name=etcd-0
        --initial-advertise-peer-urls=https://etcd-0.etcd-discovery.clusters-cewong-guest.svc:2380
        --listen-peer-urls=https://10.137.0.28:2380
[...]

PR #13224, opened 2021-07-18
netutil: add url comparison without resolver to URLStringsEqual

sakateka:
If one of the nodes in the cluster has lost a dns record,
restarting the second node will break it.
This PR makes an attempt to add a comparison without using a resolver,
which allows to protect cluster from dns errors and does not break
the current logic of comparing urls in the URLStringsEqual function.
You can read more in the issue #7798

Fixes #7798

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #15064, opened 2023-01-05
Ensure DNS is looked up on boostrap

tjungblu:
PR #13224 introduced an early exit when the names match. This change restores the previous behaviour by doing an explicit DNS resolution before checking the URLs for equality.

This is fixing https://github.com/etcd-io/etcd/issues/15062


ahrtr:
Mark this PR as draft until we clearly clarify the issue in https://github.com/etcd-io/etcd/issues/15062

Issue #14114, opened 2022-06-15
Restriction on MaxRequestSize and maxWALEntrySize


ahrtr:
# etcd build-in gRPC services
Users can configure the MaxRequestSize using flag [--max-request-bytes](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/etcdmain/config.go#L134), and its default value is [1.5MB](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/embed/config.go#L61).  When users configure a value greater than [10MB](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/etcdserver/server.go#L103), then etcd generates a [warning message](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/etcdserver/bootstrap.go#L55).  The existing logic on `--max-request-bytes` looks good to me, and we shouldn't set any hard limit on it. We just need to keep it as it's for now.

But there is a hard limit (10MB) on the [maxWALEntrySize](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/storage/wal/decoder.go#L65). If users configure a value > 10MB for the `--max-request-bytes`, then etcd can receive and process request > 10MB. But when etcd restarts later, it will fail to decode the WAL entry, see [decoder.go#L88](https://github.com/etcd-io/etcd/blob/71bba3c761b0078c81c2b39781ec74853c458303/server/storage/wal/decoder.go#L88). This is definitely a bug, see also [issues/14025](https://github.com/etcd-io/etcd/issues/14025).  We can't just remove the max size limitation on the WAL entry, because etcd may run out of memory in case the WAL entry is somehow corrupted, and its size is a huge value. The proposed solution for this issue is to replace the hard code limitation 10MB with a dynamic value, which is the remaining size of the current WAL file.  

Let's work with an example. Assuming the reading offset is  800 bytes, and the current WAL size is 64MB, then the max size of the current WAL entry should be `(64MB - 800)`.

[...]

PR #15068, opened 2023-01-07
etcdserver: process the case of the last WAL record being partially synced to disk [backport 3.5]

ahrtr:
There is a regression in https://github.com/etcd-io/etcd/pull/14122. When the last record is only partially synced to disk, we should return `io.ErrUnexpectedEOF` in the error chain, otherwise etcdserver will not repair it and etcdserver will also fail to get started.

See also https://github.com/etcd-io/etcd/issues/14114#issuecomment-1373620727


ptabor (approved):

ptabor:
I'm sorry I though I already approved this.

PR #15078, opened 2023-01-10
tests: Use watch events to patch history to speed up linearization

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"testing"
@@ -25,6 +26,7 @@ import (
 
 	"github.com/anishathalye/porcupine"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/time/rate"
 
@@ -102,7 +104,9 @@ func TestLinearizability(t *testing.T) {
 				clientCount: 8,
 				traffic:     DefaultTraffic,
 			})
-			validateEventsMatch(t, events)
+			longestHistory, remainingEvents := pickLongestHistory(events)
+			validateEventsMatch(t, longestHistory, remainingEvents)
+			operations = patchOperationBasedOnWatchEvents(operations, longestHistory)
 			checkOperationsAndPersistResults(t, operations, clus)
 		})
 	}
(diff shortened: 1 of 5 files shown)


tests/linearizability/linearizability_test.go
```diff
@@ -133,6 +136,62 @@ func testLinearizability(ctx context.Context, t *testing.T, clus *e2e.EtcdProces
 	return operations, events
 }
 
+func patchOperationHistory(operations []porcupine.Operation, watchOps []watchEvent) []porcupine.Operation {
+	newOperations := make([]porcupine.Operation, 0, len(operations))
+	persisted := map[EtcdOperation]watchEvent{}
```
ptabor, line 141: operationId2watch ?

tests/linearizability/linearizability_test.go
```diff
@@ -133,6 +136,62 @@ func testLinearizability(ctx context.Context, t *testing.T, clus *e2e.EtcdProces
...
+		hasPut := false
+		wasPersisted := false
+		hasDelete := false
+		for _, op1 := range request.Ops {
+			switch op1.Type {
+			case Delete:
+				hasDelete = true
+			case Put:
+				hasPut = true
+				op2, ok := persisted[op1]
```
ptabor, line 167: watchedOp, := ...

tests/linearizability/linearizability_test.go
```diff
@@ -133,6 +136,62 @@ func testLinearizability(ctx context.Context, t *testing.T, clus *e2e.EtcdProces
...
+			newOperations = append(newOperations, op)
+			continue
+		}
+		hasPut := false
+		wasPersisted := false
+		hasDelete := false
+		for _, op1 := range request.Ops {
+			switch op1.Type {
+			case Delete:
+				hasDelete = true
```
ptabor, line 164: I would comment here why we cannot do the same for deletes. (I see the comment below, about deletes being not unique... but here might be more natural place)

tests/linearizability/linearizability_test.go
```diff
@@ -212,20 +271,22 @@ type trafficConfig struct {
 	traffic     Traffic
 }
 
-func validateEventsMatch(t *testing.T, ops [][]watchEvent) {
-	// Move longest history to ops[0]
+func pickLongestHistory(ops [][]watchEvent) (longest []watchEvent, rest [][]watchEvent) {
```
ptabor, line 274: Just having 'sortDescByLenght' would have simpler contract and shorter implementation. 
Cost of sorting of O(3) pointers to lists is negligible. 

Now we have function that sounds like getter... but is mutating 'ops'. 

```
sort.Slice(ops, func(i, j []watchEvent) bool {
  return len(i)>len();
})
```

ptabor (approved):
Very good idea ! Just a few nits.

tests/linearizability/linearizability_test.go
serathius: Not very convinced. This function takes operations and patches them based on watch events.

serathius: Renamed to `patchOperationBasedOnWatchEvents`

tests/linearizability/linearizability_test.go
serathius: Thanks for pointing this out. Forgot about this :P

tests/linearizability/linearizability_test.go
serathius: Done

serathius:
Do not merge, still working on some weird interactions with leases that were just added.

PR #15080, opened 2023-01-11
Add support for lease api to linearizability tests

geetasg:

### DIFF
--- a/tests/linearizability/lease_ids.go
+++ b/tests/linearizability/lease_ids.go
@@ -1,0 +1,53 @@
+// Copyright 2022 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package linearizability
+
+import (
+	"sync"
+)
+
+type clientId2LeaseIdMapper interface {
+	LeaseId(int) int64
+	AddLeaseId(int, int64)
+	RemoveLeaseId(int)
+}
+
+func newClientId2LeaseIdMapper() clientId2LeaseIdMapper {
+	return &atomicClientId2LeaseIdMapper{m: map[int]int64{}}
+}
+
+type atomicClientId2LeaseIdMapper struct {
+	sync.RWMutex
+	// m is used to store clientId to leaseId mapping.
+	m map[int]int64
+}
+
+func (lm *atomicClientId2LeaseIdMapper) LeaseId(clientId int) int64 {
+	lm.RLock()
+	defer lm.RUnlock()
+	return lm.m[clientId]
+}
+
+func (lm *atomicClientId2LeaseIdMapper) AddLeaseId(clientId int, leaseId int64) {
+	lm.Lock()
+	defer lm.Unlock()
+	lm.m[clientId] = leaseId
+}
+
+func (lm *atomicClientId2LeaseIdMapper) RemoveLeaseId(clientId int) {
+	lm.Lock()
+	defer lm.Unlock()
+	delete(lm.m, clientId)
+}
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -26,10 +26,13 @@ import (
 type OperationType string
 
 const (
-	Get    OperationType = "get"
-	Put    OperationType = "put"
-	Delete OperationType = "delete"
-	Txn    OperationType = "txn"
+	Get          OperationType = "get"
+	Put          OperationType = "put"
+	Delete       OperationType = "delete"
+	Txn          OperationType = "txn"
+	PutWithLease OperationType = "putWithLease"
+	LeaseGrant   OperationType = "leaseGrant"
+	LeaseRevoke  OperationType = "leaseRevoke"
 )
 
 type EtcdRequest struct {
@@ -43,9 +46,10 @@ type EtcdCondition struct {
 }
 
 type EtcdOperation struct {
-	Type  OperationType
-	Key   string
-	Value string
+	Type    OperationType
+	Key     string
+	Value   string
+	LeaseID int64
 }
 
 type EtcdResponse struct {
@@ -60,11 +64,20 @@ type EtcdOperationResult struct {
 	Deleted int64
 }
 
+var leased = struct{}{}
+
+type EtcdLease struct {
+	LeaseID int64
+	Keys    map[string]struct{}
+}
+
 type PossibleStates []EtcdState
 
 type EtcdState struct {
 	Revision  int64
 	KeyValues map[string]string
+	KeyLeases map[string]int64
+	Leases    map[int64]EtcdLease
 }
 
 var etcdModel = porcupine.Model{
@@ -139,6 +152,12 @@ func describeEtcdOperation(op EtcdOperation) string {
 		return fmt.Sprintf("delete(%q)", op.Key)
 	case Txn:
 		return "<! unsupported: nested transaction !>"
+	case LeaseGrant:
+		return fmt.Sprintf("leaseGrant(%d)", op.LeaseID)
+	case LeaseRevoke:
+		return fmt.Sprintf("leaseRevoke(%d)", op.LeaseID)
+	case PutWithLease:
+		return fmt.Sprintf("putWithLease(%q, %q, %d)", op.Key, op.Value, op.LeaseID)
 	default:
 		return fmt.Sprintf("<! unknown op: %q !>", op.Type)
 	}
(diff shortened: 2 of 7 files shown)


geetasg:
Updated with review feedback from https://github.com/etcd-io/etcd/pull/15057. /cc @ahrtr @serathius

tests/linearizability/traffic.go
```diff
@@ -87,6 +89,24 @@ func (t readWriteSingleKey) Write(ctx context.Context, c *recordingClient, limit
...
+		}
+	case LeaseRevoke:
+		leaseId := lm.LeaseId(cid)
+		if leaseId != 0 {
+			err = c.LeaseRevoke(putCtx, leaseId)
+			//if LeaseRevoke has failed, do not remove the mapping.
+			if err == nil {
+				lm.RemoveLeaseId(cid)
+			}
+		}
```
ahrtr, line 109: Should you execute `c.LeaseGrant` as well if there is no a leaseId related to the clientId in this case (`LeaseRevoke`), just in the same way as you process `PutWithLease`? Otherwise the `(readWriteSingleKey) Write` will do nothing in this case.

The [Delete operation](https://github.com/etcd-io/etcd/blob/ff898640a5c9bad0bb99a74d0799f810d54b3586/tests/linearizability/traffic.go#L82) might delete nothing if it happens before the related `PUT` operation. But there is always a communication with the etcdserver. 

I agree that the existing implementation is simpler. 

Anyway, I am  OK to discuss this in a separate PR.

ahrtr (approved):
Overall looks good to me.

Thank you @geetasg

tests/linearizability/model.go
```diff
@@ -197,6 +224,24 @@ func initState(request EtcdRequest, response EtcdResponse) EtcdState {
 		case Put:
 			state.KeyValues[op.Key] = op.Value
 		case Delete:
+		case PutWithLease:
+			if _, ok := state.Leases[op.LeaseID]; ok {
```
serathius, line 228: This should never happen as `state.Leases` should be empty.

tests/linearizability/model.go
```diff
@@ -197,6 +224,24 @@ func initState(request EtcdRequest, response EtcdResponse) EtcdState {
...
+				state.KeyValues[op.Key] = op.Value
+				//detach from old lease id but we dont expect that at init
+				if _, ok := state.KeyLeases[op.Key]; ok {
+					panic("old lease id found at init")
+				}
+				//attach to new lease id
+				state.KeyLeases[op.Key] = op.LeaseID
```
ptabor, line 235: How about calling `attachToNewLease` here ?

tests/linearizability/model.go
```diff
@@ -197,6 +224,24 @@ func initState(request EtcdRequest, response EtcdResponse) EtcdState {
 		case Put:
 			state.KeyValues[op.Key] = op.Value
 		case Delete:
+		case PutWithLease:
+			if _, ok := state.Leases[op.LeaseID]; ok {
+				state.KeyValues[op.Key] = op.Value
+				//detach from old lease id but we dont expect that at init
+				if _, ok := state.KeyLeases[op.Key]; ok {
```
serathius, line 231: Just skip it

tests/linearizability/model.go
```diff
@@ -244,25 +289,76 @@ func applyRequestToSingleState(s EtcdState, request EtcdRequest) (EtcdState, Etc
...
 	}
+
 	if increaseRevision {
 		s.Revision += 1
 	}
+
 	return s, EtcdResponse{Result: opResp, Revision: s.Revision}
 }
+
+func detachFromOldLease(s EtcdState, op EtcdOperation) EtcdState {
```
serathius, line 352:
```suggestion
func detachFromOldLease(s EtcdState, key string) EtcdState {
```

ptabor (approved):
Looks good to me, but I think @serathius opinion is crucial here.

tests/linearizability/model.go
```diff
@@ -244,25 +289,76 @@ func applyRequestToSingleState(s EtcdState, request EtcdRequest) (EtcdState, Etc
...
+			s = detachFromOldLease(s, op)
 		case Delete:
 			if _, ok := s.KeyValues[op.Key]; ok {
 				delete(s.KeyValues, op.Key)
 				increaseRevision = true
+				s = detachFromOldLease(s, op)
 				opResp[i].Deleted = 1
 			}
+		case PutWithLease:
+			if _, ok := s.Leases[op.LeaseID]; ok {
```
serathius, line 309: Question, does etcd allows to make a Put with lease that doesn't exist? I expect that it returns an client error, but maybe I remember wrong.

tests/linearizability/model.go
```diff
@@ -244,25 +289,76 @@ func applyRequestToSingleState(s EtcdState, request EtcdRequest) (EtcdState, Etc
...
+
+func detachFromOldLease(s EtcdState, op EtcdOperation) EtcdState {
+	if oldLeaseId, ok := s.KeyLeases[op.Key]; ok {
+		delete(s.Leases[oldLeaseId].Keys, op.Key)
+		delete(s.KeyLeases, op.Key)
+	}
+	return s
+}
+
+func attachToNewLease(s EtcdState, op EtcdOperation) EtcdState {
```
serathius, line 360:
```suggestion
func attachKeyToLease(s EtcdState, leaseID int64, key string) EtcdState {
```

tests/linearizability/model_test.go
```diff
@@ -402,6 +402,122 @@ func TestModelStep(t *testing.T) {
...
+			name: "Update key with same lease",
+			operations: []testOperation{
+				{req: leaseGrantRequest(1), resp: leaseGrantResponse(1)},
+				{req: putWithLeaseRequest("key", "2", 1), resp: putResponse(2)},
+				{req: putWithLeaseRequest("key", "3", 1), resp: putResponse(3)},
+				{req: getRequest("key"), resp: getResponse("3", 3)},
+			},
+		},
+		{
+			name: "Deleting a leased key - revoke should not increment revision",
```
serathius, line 467: One additional scenario would be useful, "Overriding a leased key should remove lease"

tests/linearizability/lease_ids.go
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package linearizability
+
+import (
+	"sync"
+)
+
+type clientId2LeaseIdMapper interface {
```
serathius, line 21: What's the reason to map leases to clients? Would it be more interesting if we allowed every client to touch every lease? Changes needed:
* We just store available lease ids. Possibly with some limit to avoid explosion on to many LeaseGrant.
* When client needs to use lease it will take it out from storage to avoid racing with other clients. When client is done with a lease it can return it to storage, assuming it didn't revoke it.

serathius (approved):

serathius:
Let's merge it to avoid conflicts with https://github.com/etcd-io/etcd/pull/15078
@geetasg can you address the comments in future PR?

rchoulak:
require_head: true #

rchoulak:
merci mon dieu

geetasg:
Thanks much for the review @serathius @ahrtr and @ptabor . I will create a new PR to address the review comments in this one.

tests/linearizability/model.go
geetasg: updated in https://github.com/etcd-io/etcd/pull/15093

tests/linearizability/model.go
geetasg: removed the code in init - now putwithlease at init is nop since the lease wont be there. updated in https://github.com/etcd-io/etcd/pull/15093

tests/linearizability/model.go
geetasg: updated in https://github.com/etcd-io/etcd/pull/15093

tests/linearizability/model.go
geetasg: updated in https://github.com/etcd-io/etcd/pull/15093

tests/linearizability/model.go
geetasg: updated in https://github.com/etcd-io/etcd/pull/15093

tests/linearizability/model_test.go
geetasg: there is a scenario for Put following PutWithLease. https://github.com/etcd-io/etcd/blob/main/tests/linearizability/model_test.go#L435. Do you mean something else? Can you please clarify what you mean by override? thanks

tests/linearizability/model.go
geetasg: Adding test result for quick ref -
Put with non existent lease -

{"level":"warn","ts":"2023-01-12T15:39:35.877-0800","caller":"txn/util.go:47","msg":"failed to apply request","took":"18.335µs","request":"header:<ID:7587867905305039876 > put:<key:\"x\" value_size:1 lease:4 >","response":"","error":"lease not found"}

client (etcdctl in my case) -
macbook % ./etcdctl put x y --lease=4
{"level":"warn","ts":"2023-01-12T15:39:35.878-0800","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc00017e000/127.0.0.1:2379","method":"/etcdserverpb.KV/Put","attempt":0,"error":"rpc error: code = NotFound desc = etcdserver: requested lease not found"}
Error: etcdserver: requested lease not found

PR #15084, opened 2023-01-12
tests: Fix putWithLease traffic timeout/limit and double overall timeout to reduce number of failed requests

serathius:
Maybe due addition of leases, but I noticed increases number of request timeout even without any failure injections. 
cc @ahrtr @ptabor @geetasg

### DIFF
--- a/tests/linearizability/traffic.go
+++ b/tests/linearizability/traffic.go
@@ -27,6 +27,7 @@ import (
 
 var (
 	DefaultLeaseTTL int64   = 7200
+	RequestTimeout          = 40 * time.Millisecond
 	DefaultTraffic  Traffic = readWriteSingleKey{keyCount: 4, leaseTTL: DefaultLeaseTTL, writes: []opChance{{operation: Put, chance: 50}, {operation: Delete, chance: 10}, {operation: PutWithLease, chance: 10}, {operation: LeaseRevoke, chance: 10}, {operation: Txn, chance: 20}}}
 )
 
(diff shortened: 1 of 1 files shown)


serathius:
cc @geetasg

serathius:
Before
image

serathius:
After 
image

tests/linearizability/traffic.go
```diff
@@ -75,33 +76,38 @@ func (t readWriteSingleKey) Read(ctx context.Context, c *recordingClient, limite
...
 	case PutWithLease:
 		leaseId := lm.LeaseId(cid)
 		if leaseId == 0 {
-			leaseId, err = c.LeaseGrant(ctx, t.leaseTTL)
-			lm.AddLeaseId(cid, leaseId)
+			leaseId, err = c.LeaseGrant(writeCtx, t.leaseTTL)
+			if err != nil {
+				lm.AddLeaseId(cid, leaseId)
+				limiter.Wait(ctx)
+			}
```
ahrtr, line 100: Do you intentionally ignore the lease when `err == nil`? 

```suggestion
			lm.AddLeaseId(cid, leaseId)
			if err != nil {
				limiter.Wait(ctx)
			}
```

ahrtr:
It would be better to support limiting the number of lease in test case, and print more stats info (e.g. number of lease) for lease.

tests/linearizability/traffic.go
serathius: Ups, I had wrong condition.

serathius:
> It would be better to support limiting the number of lease in test case, and print more stats info (e.g. number of lease) for lease.

Don't understand. Leases are limited to 1 per client.

ptabor (approved):

geetasg:
Thanks Marek.

PR #15087, opened 2023-01-12
fix(clientv3):concurrent put or get permission denied when refreshToken

liuyi0618:
Signed-off-by: 刘易 <liuyi08@bilibili.com>


ptabor:
Could you, please add a repro test ?

Maybe we should get token in the background and switch it more atomically.

liuyi0618:
> Could you, please add a repro test ?
> 
> Maybe we should get token in the background and switch it more atomically.

Unfortunately, the client authTokenBundle can not refresh in the background. Maybe we can use locks to solve this problem, but this means that the same client does not support concurrent operations when the token has expired.

PR #15091, opened 2023-01-12
tests: Move stopping proxy to after process stop to speed up cluster …

serathius:
cc @ahrtr @ptabor


serathius:
On first run time went from 38m to 26m.

serathius:
Second run 25m

ahrtr (approved):

PR #15093, opened 2023-01-12
Lease support for linearizability tests :  address review feedback

geetasg:
Lease support for linearizability tests :  address review feedback

### DIFF
--- a/tests/linearizability/model.go
+++ b/tests/linearizability/model.go
@@ -240,16 +240,7 @@ func initState(request EtcdRequest, response EtcdResponse) EtcdState {
 			state.KeyValues[op.Key] = op.Value
 		case Delete:
 		case PutWithLease:
-			if _, ok := state.Leases[op.LeaseID]; ok {
-				state.KeyValues[op.Key] = op.Value
-				//detach from old lease id but we dont expect that at init
-				if _, ok := state.KeyLeases[op.Key]; ok {
-					panic("old lease id found at init")
-				}
-				//attach to new lease id
-				state.KeyLeases[op.Key] = op.LeaseID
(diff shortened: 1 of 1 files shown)


geetasg:
This PR addresses some of the feedback from https://github.com/etcd-io/etcd/pull/15080.
Following review comments are not processed here. I will submit separate PR for these ones -
* decouple leases from client ids - https://github.com/etcd-io/etcd/pull/15080#discussion_r1066717323
* should we call lease grant when lease revoke is invoked without valid lease - https://github.com/etcd-io/etcd/pull/15080#discussion_r1066659436

tests/linearizability/model.go
```diff
@@ -240,16 +240,7 @@ func initState(request EtcdRequest, response EtcdResponse) EtcdState {
...
-				state.KeyValues[op.Key] = op.Value
-				//detach from old lease id but we dont expect that at init
-				if _, ok := state.KeyLeases[op.Key]; ok {
-					panic("old lease id found at init")
-				}
-				//attach to new lease id
-				state.KeyLeases[op.Key] = op.LeaseID
-				state.Leases[op.LeaseID].Keys[op.Key] = leased
-			}
+			//nop here since lease wont be there
```
ahrtr, line 243: We already support multiple operations, so in theory it's possible that `LeaseGrant` happens before `PutWithLease`.  But current implementation doesn't support standalone `LeaseGrant`, which is executed along each `PutWithLease`, so it's OK for now.

ahrtr (approved):
LGTM

Thanks @geetasg

serathius (approved):

PR #15095, opened 2023-01-13
etcdserver: return `membership.ErrIDNotFound` when the memberID not found

ahrtr:
This PR should fix the failure in https://github.com/etcd-io/etcd/actions/runs/3908224614/jobs/6678181444 
```
cluster_test.go:259: expecting promote not ready learner to fail, got no error
```

The existing test cases `TestMemberPromote*` in [cluster_test.go](https://github.com/etcd-io/etcd/blob/main/tests/integration/clientv3/cluster_test.go) already covers the code path updated in this PR.

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1484,6 +1484,10 @@ func (s *EtcdServer) mayPromoteMember(id types.ID) error {
 // Note: it will return nil if member is not found in cluster or if member is not learner.
 // These two conditions will be checked before toApply phase later.
 func (s *EtcdServer) isLearnerReady(id uint64) error {
+	if err := s.waitAppliedIndex(); err != nil {
+		return err
+	}
+
 	rs := s.raftStatus()
 
 	// leader's raftStatus.Progress is not nil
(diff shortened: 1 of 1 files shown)


ptabor (commented):
Thank you. The logic seems better than it used to. 

Still I'm not confident the code will work well if e.g. during waitAppliedIndex the leader changes]... It's likely rs.Progress will report NULL in such a situation. 

Shell we fetch 

```
waitAppliedIndex(...)
progress:=rs.Progress
if (progress == nil) ... 
...
fir memberId, progress := range progress { ...
```

ahrtr:
> during waitAppliedIndex the leader changes]...

It seems not a problem. Even the leader changes during `waitAppliedIndex`, it might be also OK to  promote the learner member as long as it already catches up to the (old) leader's log.

But actually it doesn't make sense to execute `waitAppliedIndex` after `rs := s.raftStatus()`, because the `rs` is a copy and will not change anymore. So Moved the `waitAppliedIndex` to the beginning of the method. PTAL.

ptabor (approved):

ahrtr:
thx @ptabor for the quick review.

PR #15072, opened 2023-01-09
[3.4] backports #11990 #12933

fuweid:
It is trying to backport two improvements into release 3.4.

* Reduce memory when Range for total or with limit count
    * #11771
    * #11990
    * #13060
* Reduce expensive deepCopy when create ConcurrentReadTx
    * #12933
    * Fixes #11783: When there is only read requests, the throughput will be limited by go-runtime gc.

<details>
<summary>Click to check rw-heatmap benchmark result</summary>

Use https://github.com/fuweid/etcd/tree/tools-use-imshow-instead-of-tripcolor to render heatmap

backports_read
backports_write

[backports-based-on-v3.4.23.csv](https://github.com/etcd-io/etcd/files/10372111/backports-based-on-v3.4.23.csv)
[v3.4.23.csv](https://github.com/etcd-io/etcd/files/10372122/v3.4.23.csv)

</details>

<details>
<summary>Click to check range not-found key benchmark result</summary>

Based on steps provided by #11783, the script is like

```bash
   cd /data
    export CLIENT_PORT="23790"
    export BACKEND_SIZE="$((20 * 1024 * 1024 * 1024))"
    /root/etcd/bin/etcd --quota-backend-bytes=${BACKEND_SIZE} \
      --log-level 'error' \
      --listen-client-urls http://0.0.0.0:${CLIENT_PORT} \
      --advertise-client-urls http://127.0.0.1:${CLIENT_PORT} \
      &>/dev/null &
    sleep 5

    /root/etcd/bin/tools/benchmark --endpoints http://127.0.0.1:${CLIENT_PORT} \
       --clients 1000 --conns 100 \
       put --key-size 8 --val-size 256 --total 100000

    sleep 2

    /root/etcd/bin/tools/benchmark --endpoints http://127.0.0.1:${CLIENT_PORT} \
       --clients 1000 --conns 100 \
       range 12345678123456781234567812345678 --consistency=l --total 100000
```

* v3.4.23: Requests/sec: 26169.3610
* Backport: Requests/sec: 127041.7200
</details>

Hope it is acceptable!

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #11771, opened 2020-04-08
mvcc: reduce count-only range overhead

tangcong:
image
if there are too much keys(> 1 millions),count-only range is very slow(slice expansion) and cost too much memory. it is not necessary to return revpairs,so we can do a small optimization to reduce overhead.

PR #11743, opened 2020-04-02
etcdctl: support query count only of specified prefix

mlmhl:
Sometimes we only want to know how many objects under a specified prefix, instead of extracting the whole keys or values.

PR #12933, opened 2021-05-09
server: set multiple concurrentReadTx instances share one txReadBuffer

wilsonwang371:
Currently, the concurrentReadTx does a copy of the txReadBuffer inside backend readTx. However, this buffer is never modified by concurrentReadTx. Therefore, we can just use a single copy of the txReadBuffer as long as there are no new changes committed to backend DB.

Therefore, we use a cached txReadBuffer to avoid excessive deep copy operations for the new read transaction.

More performance testing results will be uploaded later.

@gyuho @ptabor 

Guys, please take a look at this and let me know if there is any concern or issue.

PR #15099, opened 2023-01-13
[3.4] mvcc: reduce count-only range overhead

fuweid:
Backports

* https://github.com/etcd-io/etcd/pull/11771 (Only including the code change and test)
* https://github.com/etcd-io/etcd/pull/11743

The #11771 's test case needs #11743 change. So I backport two prs in one pr.

-----

Related to https://github.com/etcd-io/etcd/pull/15072

### DIFF
--- a/etcdctl/ctlv3/command/get_command.go
+++ b/etcdctl/ctlv3/command/get_command.go
@@ -31,6 +31,7 @@ var (
 	getFromKey     bool
 	getRev         int64
 	getKeysOnly    bool
+	getCountOnly   bool
 	printValueOnly bool
 )
 
@@ -50,6 +51,7 @@ func NewGetCommand() *cobra.Command {
 	cmd.Flags().BoolVar(&getFromKey, "from-key", false, "Get keys that are greater than or equal to the given key using byte compare")
 	cmd.Flags().Int64Var(&getRev, "rev", 0, "Specify the kv revision")
 	cmd.Flags().BoolVar(&getKeysOnly, "keys-only", false, "Get only the keys")
+	cmd.Flags().BoolVar(&getCountOnly, "count-only", false, "Get only the count")
 	cmd.Flags().BoolVar(&printValueOnly, "print-value-only", false, `Only write values when using the "simple" output format`)
 	return cmd
 }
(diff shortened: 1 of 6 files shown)


serathius:
We don't backport features.

fuweid:
@serathius Thanks for the comment.

Before I file https://github.com/etcd-io/etcd/pull/15072 pr, I checked the contributor document about the backport policy.

> The commits should be restricted to bug fixes and security patches.

Except that it introduces new feature `--count-only` into the `etcdctl`, the `CountRevisions` is new internal interface to prevent  unnecessary memory allocation. It won't change the existing interface to end-user. So from the point of view of interface, could the enhancement be considered to be bugfix? :)  Thanks

And  I'm also trying to backport #12933 and wondering how to handle this case.

We are running etcd-3.4 (upgraded from 3.3) and it provides concurrent read transaction feature.
But with the high read-tx ratio, the etcd-3.4 needs more memory than etcd-3.3. It is easy to run into memory spike.
It seems that it is kind of memory regression.  If the following release has a patch to fix the memory issue and the patch is very suitable to older release, is it possible to consider the patch as bugfix?

Look forward to the release managers' comments. Thanks!

ahrtr:
In principle, we do not backport features. But we should always make judgement based on benefit and risk.

The `etcdctl` CLI flag `--count-only` is new to 3.4, but `clientv3.WithCountOnly()` is already in release-3.4 there for a long time. It's an issue to me that previous implementation of this feature (CountOnly) in 3.4 isn't complete. So I am not against to make it complete as long as it's a simple & minor change.

The simple improvement in `(ti *treeIndex) CountRevisions` can definitely reduce the memory usage when users perform only Count-Only operation, because we don't need to cache all K/V pairs any more. Again, it's a simple & minor change. So I am not against adding it in release-3.4. But @fuweid please backport the related unit test case (e.g. `TestIndexRevision`) or add a new unit test as well.

fuweid:
Thanks @ahrtr for the comment! Updated the `TestIndexRevision` ut and it is ready to review. PTAL.

ahrtr (approved):
LGTM

Thank you @fuweid 

Please also add a changelog item for 3.4.24.

fuweid:
> Please also add a changelog item for 3.4.24.

Will do. Thanks!

fuweid:
ping @serathius @ptabor ~ is it acceptable?

PR #15105, opened 2023-01-14
tests: Reduce number of runs in nightly action

serathius:
Single run takes up to 30s. Let's reduce number of runs to reduce chance of timeout.

cc @ahrtr @tjungblu @ptabor


ahrtr (approved):
LGTM

PR #15129, opened 2023-01-16
Bump some dependencies

ahrtr:
dependabot automaticaly bumped the dependencies, but unfortunately it doesn't execute "./scripts/fix.sh", and it also created a PR for each module.


ptabor (approved):
[When test passes]

ahrtr:
> [When test passes]

Sure. Will  try to bump gRPC in a separate PR.

PR #15130, opened 2023-01-17
etcdserver: use time.Ticker instead of time.After

guozhao-coder:
Using time.After will create a new Timer in each cycle, In these cases, it is better to use time.Ticker.

Signed-off-by: Zhao Guo <guozhao-coder@foxmail.com>

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -2206,6 +2206,8 @@ func (s *EtcdServer) monitorKVHash() {
 	if t == 0 {
 		return
 	}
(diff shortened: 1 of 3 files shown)


ptabor (approved):
LGTM -assuming tests are green.

guozhao-coder:
thanks for review, there are something wrong in this case, the param can't be '0' in timeNewTicker(), I will fix it

guozhao-coder:
ready for review

ahrtr (approved):
LGTM

Thank you

PR #13060, opened 2021-05-31
etcdserver:  Fix invalid count returned on Range with Limit

serathius:
Fixes https://github.com/etcd-io/etcd/issues/13056

PR https://github.com/etcd-io/etcd/pull/11990 introduced a breaking change where Count would no longer count all keys under range, but only up to the Limit. This change is a breaking change in behavior and is blocking Kubernetes from using 3.5 release as this breaks `RemainingItemCount` in K8s API.

PR #11990, opened 2020-06-08
 mvcc: push down RangeOptions.limit argv into index tree to reduce memory overhead

tangcong:
if there are too many keys(key,rangeEnd), even if we set RangeOptions.limit to 1,range request is very expensive and it will result in oom. this pr will push down RangeOptions.limit argv into index tree to reduce memory overhead. @gyuho 
image

PR #15137, opened 2023-01-18
[3.4] mvcc: push down RangeOptions.limit argv into index tree to reduce memory overhead

fuweid:
Backports:
* #11990 
* #13060
* And update the UT from #14124

----

Related to #15072

### DIFF
--- a/integration/v3_grpc_test.go
+++ b/integration/v3_grpc_test.go
@@ -1285,8 +1285,9 @@ func TestV3RangeRequest(t *testing.T) {
 		putKeys []string
 		reqs    []pb.RangeRequest
 
-		wresps [][]string
-		wmores []bool
+		wresps  [][]string
+		wmores  []bool
+		wcounts []int64
 	}{
 		// single key
 		{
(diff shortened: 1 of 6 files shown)


ahrtr:
This PR changed the protocol between etcdserver and client side, it may have some impact on all clients, e.g., K8s. So I tend not to backport this one.

fuweid:
Hi @ahrtr , would you please to point out the protocol change in the pr? I didn't follow you. Thanks!

ahrtr:
> Hi @ahrtr , would you please to point out the protocol change in the pr? I didn't follow you. Thanks!

Previously I was thinking the `RangeResult.Count` value might change with this PR,  but it seems that it doesn't. So I think this PR is acceptable.

ahrtr (approved):
LGTM

Thank you!

fuweid:
Thanks!

ahrtr:
Please also add a changelog item.

fuweid:
> Please also add a changelog item.

Sure!

It might be late because of holiday.
I will send the pr next Tuesday~

PR #11110, opened 2019-09-03
*: update test Go version / release version

gyuho:
Test images have been uploaded already.

PR #15156, opened 2023-01-20
Add TLSv1.3 support.

### LINKED ISSUE #13506: TLS1.3 support
version:
image
config:
image
image

Port detection status：
image
image

but, old version(3.4.10) is ok:
version:
image
config:
image
image
Port detection status：
image
image

tsaarni:
This PR adds optional TLS min/max protocol versions fields to `TLSInfo` struct, which is used in both server and client configs. Instead of hardcoding max version to TLSv1.2 it lets Go select the max version. This effectively enables TLSv1.3 support for etcd. The min/max values are also exposed via optional server flags and config file fields so that users can explicitly set the minimum and/or maximum TLS versions, if they want.

It seems that TLSv1.3 was previously supported (prior etcd 3.5) but it got disabled as a workaround to fix a test suite failure which appeared when Go 1.13 release enabled TLSv1.3 by default, see discussion https://github.com/etcd-io/etcd/pull/11110#issuecomment-528617457. In this PR, the failing test case is changed to set the max version to TLSv1.2. This approach allows testing ciphers with <TLSv1.3 without disabling TLSv1.3 altogether.

Related documentation change https://github.com/etcd-io/website/pull/641.

Fixes #13506

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

### DIFF
--- a/client/pkg/tlsutil/versions.go
+++ b/client/pkg/tlsutil/versions.go
@@ -1,0 +1,47 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tlsutil
+
+import (
+	"crypto/tls"
+	"fmt"
+)
+
+type TLSVersion string
+
+// Constants for TLS versions.
+const (
+	TLSVersionDefault TLSVersion = ""
+	TLSVersion12      TLSVersion = "TLS1.2"
+	TLSVersion13      TLSVersion = "TLS1.3"
+)
+
+// GetTLSVersion returns the corresponding tls.Version or error.
+func GetTLSVersion(version string) (uint16, error) {
+	var v uint16
+
+	switch version {
+	case string(TLSVersionDefault):
+		v = 0 // 0 means let Go decide.
+	case string(TLSVersion12):
+		v = tls.VersionTLS12
+	case string(TLSVersion13):
+		v = tls.VersionTLS13
+	default:
+		return 0, fmt.Errorf("unexpected TLS version %q (must be one of: TLS1.2, TLS1.3)", version)
+	}
+
+	return v, nil
+}
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -15,6 +15,7 @@
 package embed
 
 import (
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"math"
@@ -225,6 +226,11 @@ type Config struct {
 	// client/server and peers. If empty, Go auto-populates the list.
 	// Note that cipher suites are prioritized in the given order.
 	CipherSuites []string `json:"cipher-suites"`
+
+	// TlsMinVersion is the minimum accepted TLS version between client/server and peers.
+	TlsMinVersion string `json:"tls-min-version"`
+	// TlsMaxVersion is the maximum accepted TLS version between client/server and peers.
+	TlsMaxVersion string `json:"tls-max-version"`
 
 	ClusterState          string `json:"initial-cluster-state"`
 	DNSCluster            string `json:"discovery-srv"`
@@ -660,6 +666,17 @@ func updateCipherSuites(tls *transport.TLSInfo, ss []string) error {
 	return nil
 }
 
+func updateMinMaxVersions(info *transport.TLSInfo, min, max string) {
+	// Validate() has been called to check the user input, so it should never fail.
+	var err error
+	if info.MinVersion, err = tlsutil.GetTLSVersion(min); err != nil {
+		panic(err)
+	}
+	if info.MaxVersion, err = tlsutil.GetTLSVersion(max); err != nil {
+		panic(err)
+	}
+}
+
 // Validate ensures that '*embed.Config' fields are properly configured.
 func (cfg *Config) Validate() error {
 	if err := cfg.setupLogging(); err != nil {
@@ -774,6 +791,25 @@ func (cfg *Config) Validate() error {
 			"it isn't recommended to use default name, please set a value for --name. "+
 				"Note that etcd might run into issue when multiple members have the same default name",
 			zap.String("name", cfg.Name))
+	}
+
+	minVersion, err := tlsutil.GetTLSVersion(cfg.TlsMinVersion)
+	if err != nil {
+		return err
+	}
+	maxVersion, err := tlsutil.GetTLSVersion(cfg.TlsMaxVersion)
+	if err != nil {
+		return err
+	}
+
+	// maxVersion == 0 means that Go selects the highest available version.
+	if maxVersion != 0 && minVersion > maxVersion {
+		return fmt.Errorf("min version (%s) is greater than max version (%s)", cfg.TlsMinVersion, cfg.TlsMaxVersion)
+	}
+
+	// Check if user attempted to configure ciphers for TLS1.3 only: Go does not support that currently.
+	if minVersion == tls.VersionTLS13 && len(cfg.CipherSuites) > 0 {
+		return fmt.Errorf("cipher suites cannot be configured when only TLS1.3 is enabled")
 	}
 
 	return nil
--- a/server/embed/config_test.go
+++ b/server/embed/config_test.go
@@ -15,6 +15,7 @@
 package embed
 
 import (
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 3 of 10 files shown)


tsaarni:
Hi @serathius, please could you check this PR. Thank You!

serathius (approved):
LGTM, is the min/max tls config exposed in etcd server binary flags?
cc @ahrtr @ptabor for second opinion

tsaarni:
> is the min/max tls config exposed in etcd server binary flags

Currently not, I targeted only to the minimum change. I can add server flags & config file options as well if requested.

serathius:
> > is the min/max tls config exposed in etcd server binary flags
> 
> Currently not, I targeted only to the minimum change. I can add server flags & config file options as well if requested.

Sounds good

tsaarni:
I've now added config options for etcd server binary and new tests. I see the docs are in website repo so a follow up PR is needed for the docs after confirming the proposed flags are OK.

server/embed/config.go
```diff
@@ -226,6 +226,11 @@ type Config struct {
 	// Note that cipher suites are prioritized in the given order.
 	CipherSuites []string `json:"cipher-suites"`
 
+	// MinVersion is the minimum accepted TLS version between client/server and peers.
+	MinVersion string `json:"min-tls-version"`
```
tuminoid, line 230: Can we have this as `--tls-min-version` and `--tls-max-version` ?

client/pkg/tlsutil/versions.go
```diff
@@ -0,0 +1,38 @@
...
+
+package tlsutil
+
+import (
+	"crypto/tls"
+	"fmt"
+)
+
+// tlsVersions is a map of TLS version string to the value of tls.Config.Min/MaxVersion.
+var tlsVersions = map[string]uint16{
```
tuminoid, line 23: Since etcd only supports TLS 1.2 pre-PR, does it make sense to add support for obsolete, insecure and disabled-by-default TLS 1.0 and TLS 1.1? Enabling TLS 1.3 would be sufficient in my opinion, and configuration would then allow a meaningful choice between forcing TLS 1.2 only or TLS 1.3 only in addition to default behavior of enabling both.

tuminoid (commented):
Thanks @tsaarni for the PR! I added two comments, let me know what you think @serathius.

tsaarni:
Thanks @tuminoid I agree with your review comments, so I went ahead and made those changes. Eagerly waiting for more reviews :-)

serathius:
cc @ptabor

tsaarni:
@serathius One test failed but could it be a flake https://github.com/etcd-io/etcd/actions/runs/3994448937/jobs/6852174459?  Would it be possible to trigger a re-run?

server/embed/config.go
```diff
@@ -226,6 +226,11 @@ type Config struct {
 	// Note that cipher suites are prioritized in the given order.
 	CipherSuites []string `json:"cipher-suites"`
 
+	// MinVersion is the minimum accepted TLS version between client/server and peers.
+	MinVersion string `json:"tls-min-version"`
+	// MaxVersion is the maximum accepted TLS version between client/server and peers.
+	MaxVersion string `json:"tls-max-version"`
```
ahrtr, line 232:
```suggestion
	TlsMinVersion string `json:"tls-min-version"`
	// MaxVersion is the maximum accepted TLS version between client/server and peers.
	TlsMaxVersion string `json:"tls-max-version"`
```

server/etcdmain/config.go
```diff
@@ -215,6 +215,8 @@ func newConfig() *config {
 	fs.StringVar(&cfg.ec.PeerTLSInfo.AllowedHostname, "peer-cert-allowed-hostname", "", "Allowed TLS hostname for inter peer authentication.")
 	fs.Var(flags.NewStringsValue(""), "cipher-suites", "Comma-separated list of supported TLS cipher suites between client/server and peers (empty will be auto-populated by Go).")
 	fs.BoolVar(&cfg.ec.PeerTLSInfo.SkipClientSANVerify, "experimental-peer-skip-client-san-verification", false, "Skip verification of SAN field in client certificate for peer connections.")
+	fs.StringVar(&cfg.ec.MinVersion, "tls-min-version", "TLS12", "Minimum TLS version supported for peer and client connections. Possible values: TLS12, TLS13.")
+	fs.StringVar(&cfg.ec.MaxVersion, "tls-max-version", "", "Maximum TLS version supported for peer and client connections. Possible values: TLS12, TLS13 (empty will be auto-populated by Go).")
```
ahrtr, line 219: Please keep the description consistent with the messages in `help.go`.  The flags are only used by etcdserver. Suggest to simplify it to something like "`Minimum TLS version supported by etcd. Possible values: TLS12, TLS13.`"

tests/integration/v3_tls_test.go
```diff
@@ -72,3 +78,57 @@ func testTLSCipherSuites(t *testing.T, valid bool) {
 		t.Fatalf("expected TLS handshake success, got %v", cerr)
 	}
 }
+
+func TestTLSMinMaxVersion(t *testing.T) {
```
ahrtr, line 82: minor comment, it would be better to refactor this test case using table driven + golang subtest.

server/embed/config_test.go
```diff
@@ -429,3 +430,80 @@ func TestLogRotation(t *testing.T) {
...
+				MaxVersion: "invalid version",
+			},
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := updateMinMaxVersions(&tt.config.PeerTLSInfo, tt.config.MinVersion, tt.config.MaxVersion)
+			if err != nil && !tt.expectError {
```
ahrtr, line 494: Note that it doesn't cover the case `err == nil && tt.expectError`.

```suggestion
			if (err != nil) != tt.expectError {
```

ahrtr:
Thank you @tsaarni 

Overall looks good to me, with a couple of minor comments.

server/etcdmain/help.go
```diff
@@ -199,6 +199,10 @@ Security:
     Comma-separated whitelist of origins for CORS, or cross-origin resource sharing, (empty or * means allow all).
   --host-whitelist '*'
     Acceptable hostnames from HTTP client requests, if server is not secure (empty or * means allow all).
+  --tls-min-version 'TLS12'
+    Minimum TLS version supported for client and peer connections. Possible values: TLS12, TLS13.
+  --tls-max-version ''
+    Maximum TLS version supported for client ane peer connections. Possible values: TLS12, TLS13 (empty will be auto-populated by Go).
```
ahrtr, line 205: Please update the messages as well, to keep consistent with the message in config.go

ahrtr (approved):
LGTM

Please consider to use `github.com/stretchr/testify` to simplify the test cases if you want.

tsaarni:
Thank you @ahrtr, great feedback! 

If it's OK, I will leave the tests as they are, but I also like `github.com/stretchr/testify` asserts better. When looking around the test files I touched, I thought standard `testing` package was preferred, so I used that mostly.

tsaarni:
I've submitted also a proposal for the documentation change https://github.com/etcd-io/website/pull/641.

client/pkg/tlsutil/versions.go
```diff
@@ -0,0 +1,36 @@
...
```
serathius, line 23: Please create a const for "TLS12" and "TLS13", best if we had type alias for it.

client/pkg/tlsutil/versions.go
```diff
@@ -0,0 +1,36 @@
...
+	"":      0, // If version was not given use 0 (uninitialized version) to let Go decide.
+	"TLS12": tls.VersionTLS12,
+	"TLS13": tls.VersionTLS13,
```
serathius, line 26:
```suggestion
	"TLS1.2": tls.VersionTLS12,
	"TLS1.3": tls.VersionTLS13,
```

server/embed/config.go
```diff
@@ -660,6 +665,27 @@ func updateCipherSuites(tls *transport.TLSInfo, ss []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	maxVersion, err := tlsutil.GetTLSVersion(max)
+	if err != nil {
+		return err
+	}
+
+	// Check if both min and max were provided and min is greater than max
```
serathius, line 679: This should be done during configuration validation stage, not during apply.

server/embed/etcd.go
```diff
@@ -611,6 +614,9 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
 	if err = cfg.ClientSelfCert(); err != nil {
 		cfg.logger.Fatal("failed to get client self-signed certs", zap.Error(err))
 	}
+	if err = updateMinMaxVersions(&cfg.ClientTLSInfo, cfg.TlsMinVersion, cfg.TlsMaxVersion); err != nil {
+		return nil, err
+	}
```
serathius, line 619: This side effect approach of generating TLSInfo is very error prone. Are we sure we called `updateMinMaxVersions` in all places that TLSInfo is changed?


Would really like to have this code cleaned up.

server/embed/config.go
```diff
@@ -660,6 +665,27 @@ func updateCipherSuites(tls *transport.TLSInfo, ss []string) error {
...
+		return fmt.Errorf("min version (%s) is greater than max version (%s)", min, max)
+	}
+
+	info.MinVersion = minVersion
+	info.MaxVersion = maxVersion
+	return nil
+}
+
 // Validate ensures that '*embed.Config' fields are properly configured.
 func (cfg *Config) Validate() error {
```
serathius, line 690: Can you add validation of min-tls and max-tls flags here?

serathius:
Please add e2e tests for new etcdserver flags.

server/embed/config.go
```diff
@@ -660,6 +665,27 @@ func updateCipherSuites(tls *transport.TLSInfo, ss []string) error {
...
+		return err
+	}
+
+	maxVersion, err := tlsutil.GetTLSVersion(max)
+	if err != nil {
+		return err
+	}
+
+	// Check if both min and max were provided and min is greater than max
+	if minVersion != 0 && maxVersion != 0 && minVersion > maxVersion {
```
serathius, line 680: This validation doesn't check if cipher suites are configured. As cipher suits only work with TLS 1.2, we should not allow users to set them if maxTLS is not set to TLS1.2

server/embed/etcd.go
tsaarni: I copied the approach from how the ciphers seemed to be generated to `TLSInfo` by similar `updateCipherSuites()` function (called [here](https://github.com/etcd-io/etcd/blob/main/server/embed/config.go#L932) and [here]( https://github.com/etcd-io/etcd/blob/3b612ce345753f97156833e20e93bf78b56d5267/server/embed/config.go#L951)). I'm little doubtful if I'm able to do bigger refactoring for `TLSInfo` generation. Maybe there is an approach to clean up but with minimal impact to existing code?

client/pkg/tlsutil/versions.go
serathius: Let's use correct semantics names, we don't want to run into Windows 9 problems.

server/embed/config.go
tuminoid: > This validation doesn't check if cipher suites are configured. As cipher suits only work with TLS 1.2, we should not allow users to set them if maxTLS is not set to TLS1.2

You must be able to set cipher suites for TLS 1.2 in case you have both TLS 1.2 and TLS 1.3 enabled (the default), so what you said does not make sense. What you probably wanted is failing validation if min-version and max-version is TLS 1.3.

tsaarni: I added a check when TLSv1.3 is the only enabled version. 

It is not out of the question that there will be configurability in the future versions of Go and the check might need to be removed later. At this time, it is of course unknown how the possible configurability would look like.

server/embed/config.go
```diff
@@ -660,6 +666,12 @@ func updateCipherSuites(tls *transport.TLSInfo, ss []string) error {
 	return nil
 }
 
+func updateMinMaxVersions(info *transport.TLSInfo, min, max string) {
+	// Validate() has been called to check the user input, so it is safe to ignore errors here.
```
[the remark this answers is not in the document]
serathius: panic or `cfg.logger.Fatal`.

server/etcdmain/config.go
```diff
@@ -215,6 +216,8 @@ func newConfig() *config {
 	fs.StringVar(&cfg.ec.PeerTLSInfo.AllowedHostname, "peer-cert-allowed-hostname", "", "Allowed TLS hostname for inter peer authentication.")
 	fs.Var(flags.NewStringsValue(""), "cipher-suites", "Comma-separated list of supported TLS cipher suites between client/server and peers (empty will be auto-populated by Go).")
 	fs.BoolVar(&cfg.ec.PeerTLSInfo.SkipClientSANVerify, "experimental-peer-skip-client-san-verification", false, "Skip verification of SAN field in client certificate for peer connections.")
+	fs.StringVar(&cfg.ec.TlsMinVersion, "tls-min-version", string(tlsutil.TLSVersion12), "Minimum TLS version supported by etcd. Possible values: TLS1.2, TLS1.3.")
+	fs.StringVar(&cfg.ec.TlsMaxVersion, "tls-max-version", "", "Maximum TLS version supported by etcd. Possible values: TLS1.2, TLS1.3 (empty will be auto-populated by Go).")
```
serathius, line 220:
```suggestion
	fs.StringVar(&cfg.ec.TlsMaxVersion, "tls-max-version", string(tlsutil.TLSVersionDefault), "Maximum TLS version supported by etcd. Possible values: TLS1.2, TLS1.3 (empty will be auto-populated by Go).")
```

serathius, line 220:
```suggestion
	fs.StringVar(&cfg.ec.TlsMaxVersion, "tls-max-version", "", "Maximum TLS version supported by etcd. Possible values: TLS1.2, TLS1.3 (empty defers to Go version).")
```

server/embed/etcd.go
serathius: Sounds good.

serathius (approved):

tsaarni:
I have made the requested changes but tests failed, possibly flukes again?

Issue #11954, opened 2020-05-29
auth token invalid after watch reconnects


jschwinger23:
# Reproduce Procedure

1. prepare etcd cluster with 3 nodes and respective etcdctl

```
/go/src/github.com/coreos/etcd # etcd --version
etcd Version: 3.4.9
Git SHA: Not provided (use ./build instead of go build)
Go Version: go1.14.2
Go OS/Arch: linux/amd64
```

though I don't think configure matters under this issue, let me present one of them

```
name: etcd2
data-dir: /data/zz_6129484611666145821
listen-peer-urls: http://0.0.0.0:23800
listen-client-urls: http://0.0.0.0:23790
initial-advertise-peer-urls: http://10.213.20.39:23800
advertise-client-urls: http://10.213.20.39:23790
initial-cluster: etcd0=http://10.213.20.40:23800,etcd1=http://10.213.20.38:23800,etcd2=http://10.213.20.39:23800
initial-cluster-token: zz
initial-cluster-state: new
auto-compaction-retention: "1"
quota-backend-bytes: -1
```

and the etcd processes are running simply by 

```
etcd --config-file /etc/etcd.conf
```

no envirnments, no command line arguments

2. enable auth

```
etcdctl user add root
# then type root password, mine is 'root'
etcdctl --user root:root auth enable
```

3. watch

```
etcdctl --user root:root watch / --prefix
```

we can put a key to ensure the watch is working right now

```
etcdctl put /a b
```

4. wait for token deleting

wait for 5 min until the token is deleted by `simpleTokenKeeper`, then kill the etcd processes one by one and restart them immediately after kill

be noted do NOT kill the process until the cluster recovers healthy

5. watch fails

then you'll realize the watch is down with the output `permission deny`

# Analysis

the issue is cause by `simpleTokenKeeper`, here is the timeline

1. etcdctl dials grpc and fetches an auth token, let's say `TOKEN-A`
2. etcdctl dials grpc with `TOKEN-A`, and watch / --prefix as expected
3. after 5 min, `simpleTokenKeeper` delete `TOKEN-A`
4. watch continues working even if `TOKEN-A` has been deleted, because token is only checked upon grpc invocation
5. killing etcd process terminates connection, and etcdv3 client will re-invoke grpc Watch with the same token `TOKEN-A`
6. `authStore.AuthInfoFromCtx` will return `ErrInvalidAuthToken` due to `TOKEN-A` no longer exists

# Impact

[...]

PR #12549, opened 2020-12-14
client/v3: clear auth token when encounter ErrInvalidAuthToken

bbiao:
Old etcdserver which have not apply pr of #12165 will check auth token
even if the request is an Authenticate request.

If the client has a invalid auth token, it will not able to update it's
token, since the Authenticate has a invalid auth token.
This fix clear the auth token when encounter an ErrInvalidAuthToken to
talk with old version etcd servers.
With pr #12165 and #12264 this will fix the problem in #12385 

Fix #12385

PR #15164, opened 2023-01-23
fix(client):getToken automatically when the token has expired

upuplouis:
Signed-off-by: upuplouis <45809347+upuplouis@users.noreply.github.com>

getToken automatically. Otherwise, auth token may be invalid when the token has expired
(see https://github.com/etcd-io/etcd/issues/11954 for more).

### DIFF
--- a/client/v3/retry_interceptor.go
+++ b/client/v3/retry_interceptor.go
@@ -78,14 +78,14 @@ func (c *Client) unaryClientInterceptor(optFuncs ...retryOption) grpc.UnaryClien
 				continue
 			}
 			if c.shouldRefreshToken(lastErr, callOpts) {
-				gterr := c.refreshToken(ctx)
-				if gterr != nil {
+				gtErr := c.refreshToken(ctx)
+				if gtErr != nil {
 					c.GetLogger().Warn(
 						"retrying of unary invoker failed to fetch new auth token",
 						zap.String("target", cc.Target()),
-						zap.Error(gterr),
+						zap.Error(gtErr),
 					)
-					return gterr // lastErr must be invalid auth token
+					return gtErr // lastErr must be invalid auth token
 				}
 				continue
 			}
@@ -169,8 +169,7 @@ func (c *Client) refreshToken(ctx context.Context) error {
 		// clients just need to retry the operations (e.g. Put, Delete etc).
 		return nil
 	}
-	// clear auth token before refreshing it.
-	c.authTokenBundle.UpdateAuthToken("")
+
 	return c.getToken(ctx)
 }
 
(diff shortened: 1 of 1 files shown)


ahrtr:
It seems that we don't need `refreshToken` at all. 

cc @mitake to double confirm

upuplouis:
UpdateAuthToken("") make it non-atomic.

mitake:
@upuplouis Is the motivation described in https://github.com/etcd-io/etcd/pull/15087 ?Do you have a go client which has multiple goroutines and they share a single etcd client instance?

ahrtr:
If we don't need `refreshToken`, can we remove it? @mitake 

@upuplouis Please squash the commits.

upuplouis:
> If we don't need `refreshToken`, can we remove it? @mitake
> 
> @upuplouis Please squash the commits.

done

upuplouis:
> @upuplouis Is the motivation described in #15087 ?Do you have a go client which has multiple goroutines and they share a single etcd client instance? Then I think it's understandable.
> 
> Probably the PR name should be `fix(client):getToken atomically when the token has expired`?

Thanks.

ahrtr (approved):
LGTM

mitake:
@ahrtr 
> If we don't need refreshToken, can we remove it? @mitake

Let me double check it this weekend, probably it’s safe to remove though.

mitake:
I think it's better to keep `refreshToken()` for `if c.authTokenBundle == nil` checking (or move this check to `getToken()`). @upuplouis could you remove `c.authTokenBundle.UpdateAuthToken("")` from `refreshToken()`? I think that would be a better fix.

[This token update with ""](https://github.com/upuplouis/etcd/blob/f2263f3b5d97171651838bc0ab22ac733e9f21d8/client/v3/retry_interceptor.go#L174) was introduced in https://github.com/etcd-io/etcd/pull/12549 and it's for old etcd servers which doesn't have https://github.com/etcd-io/etcd/pull/12165 (like release-3.4). So the above change can be backported to release-3.5, but it's better not to backport to release-3.4. I'll make a custom change for release-3.4 for avoiding the race condition.

ahrtr:
Thanks @mitake for the clarification.

> I think it's better to keep `refreshToken()` for `if c.authTokenBundle == nil` checking (or move this check to `getToken()`).

Essentially `if c.authTokenBundle == nil` is the same thing as `if c.Username == "" || c.Password == ""` in `getToken`, so it isn't necessary. But it isn't a big deal, and the immediate feeling is `if c.authTokenBundle == nil` is safer. It's up to you.

upuplouis:
> Thanks @mitake for the clarification.
> 
> > I think it's better to keep `refreshToken()` for `if c.authTokenBundle == nil` checking (or move this check to `getToken()`).
> 
> Essentially `if c.authTokenBundle == nil` is the same thing as `if c.Username == "" || c.Password == ""` in `getToken`, so it isn't necessary. But it isn't a big deal, and the immediate feeling is `if c.authTokenBundle == nil` is safer. It's up to you.

done
True enough.

ahrtr:
Please fix the format error.

```
$ ./script/fix.sh
```

upuplouis:
> Please fix the format error.
> 
> ```
> $ ./script/fix.sh
> ```

done

ahrtr:
There is still format error. Please make sure your go version is 1.19.5.

upuplouis:
> There is still format error. Please make sure your go version is 1.19.5.

done

upuplouis:
> There is still format error. Please make sure your go version is 1.19.5.

done

PR #15178, opened 2023-01-24
tests: Add defragment request to linearizability tests

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/tests/linearizability/model/model.go
+++ b/tests/linearizability/model/model.go
@@ -58,6 +58,7 @@ const (
 	Txn         RequestType = "txn"
 	LeaseGrant  RequestType = "leaseGrant"
 	LeaseRevoke RequestType = "leaseRevoke"
+	Defragment  RequestType = "defragment"
 )
 
 type EtcdRequest struct {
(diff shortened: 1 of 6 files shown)


tests/linearizability/model/model.go
```diff
@@ -288,7 +298,9 @@ func initState(request EtcdRequest, response EtcdResponse) EtcdState {
 func applyFailedRequest(states PossibleStates, request EtcdRequest) PossibleStates {
 	for _, s := range states {
 		newState, _ := applyRequestToSingleState(s, request)
-		states = append(states, newState)
+		if !reflect.DeepEqual(newState, s) {
+			states = append(states, newState)
+		}
```
ahrtr, line 303: Could you clarify this change? Shouldn't we always cache the state for failed request?

serathius: I'm adding an exception because of the new Defragment request. As defragment should not impact etcd results, the etcd state does not change. If etcd state does not change, there is no need to "cache" the state on failed request.

ahrtr (approved):

PR #15181, opened 2023-01-25
linearizability tests : allow clients to revoke random lease

geetasg:

### DIFF
--- a/tests/linearizability/identity/lease_ids.go
+++ b/tests/linearizability/identity/lease_ids.go
@@ -1,5 +1,4 @@
 // Copyright 2022 The etcd Authors
-//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


geetasg:
This addresses review feedback from previous PR - the lease revoke will now not be a nop as long as atleast one lease is alive. Secondly the clients will revoke lease for other clients.

ref:
https://github.com/etcd-io/etcd/pull/15093#issuecomment-1381107231

tests/linearizability/traffic.go
```diff
@@ -122,12 +122,12 @@ func (t traffic) Write(ctx context.Context, c *recordingClient, limiter *rate.Li
...
 		}
 	case LeaseRevoke:
-		leaseId := lm.LeaseId(cid)
+		clientId, leaseId := lm.GetRandomClientWithLease()
 		if leaseId != 0 {
 			err = c.LeaseRevoke(writeCtx, leaseId)
 			//if LeaseRevoke has failed, do not remove the mapping.
 			if err == nil {
-				lm.RemoveLeaseId(cid)
+				lm.RemoveLeaseId(clientId)
```
serathius, line 130: Don't understand why leases are connected to clientId.

tests/linearizability/identity/lease_ids.go
```diff
@@ -15,39 +15,84 @@
...
+	"math/rand"
 	"sync"
 )
 
 type LeaseIdStorage interface {
 	LeaseId(int) int64
 	AddLeaseId(int, int64)
 	RemoveLeaseId(int)
+	GetRandomClientWithLease() (int, int64)
+}
```
serathius, line 27: How I imagine the interface for leases.
```suggestion
type LeaseIdStorage interface {
    // Add adds lease to storage. Should be called after LeaseGrant. Increases number leases.
	Add(leaseId int64)
	// Remove permanently removes lease from storage. Should be called after LeaseRevoke. Decreases number of leases.
    Remove(leaseId int64)
    // Take returns a no in use lease from storage and marks it as in use. Should be called before Put with Lease. Doesn't change number of leases. Can return 0 if there are no leases, or all leases are in use.
    Take() leaseId
    // Return returns in use lease to storage so it can be used later. Should be called after Put with Lease. Doesn't change number of leases. 
    Return(leaseId)
}
```

geetasg: @serathius can you please comment on following questions
1.Any suggestion on how many total leases should be kept?
2.The 'Take' interface suggested above is marking a lease in-use. Does that mean multiple putwithlease requests should not use the same lease? How about we keep total pool of say 8 leases at a time and putwithlease will pick one randomly in-use or not?

serathius: 4 or 5, we can adjust them whenever.

geetasg:
requesting review. /cc @serathius @ahrtr 
open question - should the test panic if outstanding leases are more than maxLeases count? currently it is being treated as an error.

ahrtr:
> should the test panic if outstanding leases are more than maxLeases count

Generally speaking, It depends on whether you expect the test continue to execute. If yes, then do no panic; otherwise panic. 

Back to this case, I don't think it should panic. You control the number of lease, but should also accept it occasionally exceeds the threshold (e.g. due to race condition)?

tests/linearizability/identity/lease_ids.go
```diff
@@ -15,39 +14,151 @@
...
+	Remove(int64)
+
+	// Take returns an unused lease from storage
+	Take() int64
+
+	// Return returns a used lease to storage so it can be used later.
+	Return(int64)
+
+	// GetRandom returns a random lease (could be in use or not)
+	GetRandom() int64
```
serathius, line 35: GetRandom only complicates the interface without real benefit. It just a combination of `Remove(Take())`

tests/linearizability/identity/lease_ids.go
```diff
@@ -15,39 +14,151 @@
...
+
+	// Count returns count of leases (could be in use or not)
+	Count() int
```
serathius, line 38: Count it not needed and is not used

tests/linearizability/identity/lease_ids.go
```diff
@@ -15,39 +14,151 @@
...
-	return &atomicClientId2LeaseIdMapper{m: map[int]int64{}}
+	return &atomicLeaseKeeper{ma: map[int64]int{}, mf: map[int64]int{}}
 }
 
-type atomicClientId2LeaseIdMapper struct {
+const noLease int64 = 0
+
+// atomicLeaseKeeper stores the leases in two lists - one for free leases and one for all leases.
+// two auxiliary maps are kept to allow fast deletion from the lists
+type atomicLeaseKeeper struct {
```
serathius, line 49: This is very overdesigned. You just need a list with a mutex. No need to track in use leases, no need to do any mapping.

```
type LeaseStorage struct {
  mux sync.RWMutex
  free []int64
  inUse int
}

func (s *LeaseStorage) Add(leaseId int64) {
	s.mux.Lock()
	defer s.mux.Unlock()

	s.free = append(s.free, leaseId)
}

func (s *LeaseStorage) Remove(leaseId int64) {
	s.mux.Lock()
	defer s.mux.Unlock()
    s.remove(leaseId)
}

func (s *LeaseStorage) Take(toRemove int64) {
	s.mux.Lock()
	defer s.mux.Unlock()
	if s.remove(toRemove) {
      inUse++
    }
}

func (s *LeaseStorage) remove(toRemove int64) bool {
   i := 0 
   for _, lease := range s.free {
     if lease == toRemove {
       continue
     }
     s.free[i] = lease
     i++
   }
   return i != len(s.free)
}

func (s *LeaseStorage) Return(leaseId int64) {
	s.mux.Lock()
	defer s.mux.Unlock()
    inUse--
    s.free = append(s.free, leaseId)
}

func (s *LeaseStorage) Count() int {
	s.mux.RLock()
	defer s.mux.RUnlock()
    return len(s.free) + s.inUse
}
```

tests/linearizability/traffic.go
```diff
@@ -108,26 +111,32 @@ func (t traffic) Write(ctx context.Context, c *recordingClient, limiter *rate.Li
 		}
 		err = c.CompareAndSet(writeCtx, key, expectValue, newValue)
 	case PutWithLease:
-		leaseId := lm.LeaseId(cid)
+		leaseId := lm.Take()
 		if leaseId == 0 {
+			if lm.Count() >= t.maxLeases {
+				err = errors.New("too many leases")
```
serathius, line 117: We shouldn't error here, we should pick another request.

tests/linearizability/identity/lease_ids.go
```diff
@@ -15,39 +14,151 @@
...
-	// m is used to store clientId to leaseId mapping.
-	m map[int]int64
+	// la is used to keep all leases
+	la []int64
+	// ma is used to store mapping of leaseId to index in the slice la.
+	ma map[int64]int
+	// lf stores leases that are free(not in use)
+	lf []int64
+	// mf is used to store mapping of leaseId to index in the slice lf.
+	mf map[int64]int
```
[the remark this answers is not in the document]
serathius: No need for linked list, slice is good enough. We are talking about collection with 4-5 elements, performance is not a concern here. 
Tracking just free elements is enough.

PR #15183, opened 2023-01-25
fix(client):observe new watcher always permission denied  when the token has expired 

upuplouis:
see #12385 for more.
similar problems still exist When observe and watch use the same client and ctx.


upuplouis:
@ahrtr Can you help review this PR?

lavacat:
I've commented in https://github.com/etcd-io/etcd/issues/12385#issuecomment-1417072469 about latest changes and an upcoming fix.
Is this PR trying to fix that issue?

Issue #15189, opened 2023-01-26
Linearizability Nightly times out on defragBeforeCopy


serathius:
### Which github workflows are flaking?

Linearizability Nightly

### Which tests are flaking?

TestLinearizability/Issue14685

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4004517262/jobs/6873735367

### Reason for failure (if possible)

Looks like etcd paniced but haven't exited in time. We could consider:
* Increasing the failpoint timeout to 2 seconds
* Retrying failpoint trigger.

serathius:
Duplicate https://github.com/etcd-io/etcd/actions/runs/4013851380/jobs/6893668540

serathius:
Addressed some time ago

PR #15191, opened 2023-01-26
tests: Increase failpoint timeout and add more logs

serathius:
Ref https://github.com/etcd-io/etcd/issues/15189


ahrtr (approved):

serathius:
Ups, it doesn't help. I think that recent change that added Defrag to traffic is interrupting with failpoint on defrag.

PR #15202, opened 2023-01-30
tests: Propagage logger through linearizability tests

serathius:
Main benefit is consistent log header that includes date making it easier to derive when event happend and how long it took.
cc @ahrtr @ptabor


ahrtr (approved):
lgtm

ptabor (approved):

PR #15172, opened 2023-01-23
Add functional tests on arm64

dims:
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Issue #14748, opened 2022-11-14
T1 support for Arm64 


kevinzs2048:
### What would you like to be added?

Support Arm64 as the stable release.

### Why is this needed?

According to upstream's website (https://etcd.io/docs/v3.3/op-guide/supported-platform/), the Arm64 is not classed as a stable architecture.

There are many the Kubernetes cluster running on Arm64 and already use Arm64 etcd for quite a lot time, but some users are a little concerned, since the ETCD on Arm64 is not sanctioned by upstream. E.g, Etcd works quite well on Arm64 and there are still some blog recording this, like this [ETCD optimization doc on Arm64](https://community.arm.com/arm-community-blogs/b/infrastructure-solutions-blog/posts/improve-etcd-performance-by-18-percent-by-deploying-on-aws-graviton2).

We are from [Linaro](https://www.linaro.org/) and our targeting is to enrich the Arm64 open source ecosystem. Actually from the Arm ecosystem, we do have several companies raise this topic and we do want to see if there is some gaps/TODO we can do for ETCD to support Arm64 as the stable release.

serathius:
Please read https://etcd.io/docs/v3.6/op-guide/supported-platform/#supporting-a-new-platform

kevinzs2048:
Hi @serathius, thanks for that info, it's quite clear and I find the main gaps for Arm64 to be a Tier 1 platform. I will discuss internally and adding more CI on linux-arm64 for ETCD to cover the test scenario.

dims:
/assign @dims 

xref: https://github.com/cncf/cluster/issues/226

dims:
We have 2 machines, machine spec is as follows:
![image](https://user-images.githubusercontent.com/23304/214130923-7cd8ae70-444e-4635-ab73-d3265bf3ae29.png)

dims:
To get access to the Equinix console, please reach out to me or folks in https://github.com/cncf/cluster/issues/226 and ask for access to the `etcd` project under `CNCF` org. Direct url is https://console.equinix.com/projects/1b8c1eb7-983c-4b40-97e0-e317406e232e

the two machines are:
- etcd-c3-large-arm64-runner-01 / 86.109.7.233
- etcd-c3-large-arm64-runner-02 / 147.28.151.226

dims:
testing a CI job in https://github.com/etcd-io/etcd/pull/15172

dims:
We are green!

[...]

PR #15230, opened 2023-02-01
Add E2E tests on arm64

kevinzs2048:
Signed-off-by: Kevin Zhao <kevin.zhao@linaro.org>

### DIFF
--- a/.github/workflows/e2e-arm64.yaml
+++ b/.github/workflows/e2e-arm64.yaml
@@ -1,0 +1,34 @@
+name: E2E-arm64
+on:
+  schedule:
+    - cron: '0 1 * * *' # runs daily at 1am.
+permissions: read-all
+jobs:
+  test:
+    runs-on: [Linux, ARM64]
+    strategy:
+      fail-fast: true
+      matrix:
+        target:
+        - linux-arm64-e2e
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+    - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+      with:
+        ref: main
+        go-version: "1.19.5"
+    - run: date
+    - env:
+        TARGET: ${{ matrix.target }}
+      run: |
+        echo "${TARGET}"
+        case "${TARGET}" in
+          linux-arm64-e2e)
+            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh 2>&1 | tee test.log
+            ! grep -E "(--- FAIL:|FAIL:|DATA RACE|panic: test timed out|appears to have leaked)" -B50 -A10 test.log
+            ;;
+          *)
+            echo "Failed to find target"
+            exit 1
+            ;;
+        esac


kevinzs2048:
This PR is regarded with [#14748](https://github.com/etcd-io/etcd/issues/14748)

Just a PR to add e2e test for Arm64 to check if it is green.
If the job is OK, will change it to a periodical job.

kevinzs2048:
@dims @geetasg @chaochn47 first time commit here, so needs maintainers to approve first. Thanks!

ahrtr:
I think we need to add similar tests (see below) for ARM64,
- https://github.com/etcd-io/etcd/blob/main/.github/workflows/tests.yaml
- https://github.com/etcd-io/etcd/blob/main/.github/workflows/e2e.yaml

And remove https://github.com/etcd-io/etcd/blob/main/.github/workflows/functional-arm64.yaml

kevinzs2048:
Hi @ahrtr, thanks!
This one is to add e2e test for Arm64 to verify the function.
After it work, I will add the integration test in the tests.yaml and remove the functional test since it is going to be deprecated.

The reason why it is located to a separate file is that it needs to run on a special Arm64 node and also this will be changed to periodical job.

.github/workflows/e2e-arm64.yaml
```diff
@@ -0,0 +1,31 @@
+name: E2E-arm64
```
serathius, line 1: Instead of adding a separate workflow, can you add arm64 as a job to existing "E2E" workflow?

ahrtr: It doesn't work because the ARM machines have different labels. FYI. https://github.com/etcd-io/etcd/pull/15172#discussion_r1084600047

Probably we can add a template to be shared by both the existing E2E workflow and ARM64. But I don't think it's a big deal, It's OK to leave it as it's for now.

ahrtr:
> this will be changed to periodical job.

Why?

dims:
> Why?

@ahrtr the aim is not to be a bottleneck to the current set of CI jobs that are run before we land a PR. The problem is we have only 2 github runners for arm64 and if we have a lot of PRs in flight the time that folks have to wait to land something will be a lot.

ahrtr:
> The problem is we have only 2 github runners for arm64 and if we have a lot of PRs in flight the time that folks have to wait to land something will be a lot.

Makes sense. Thanks @dims for the clarification.

ahrtr (approved):
LGTM 

Thank you @kevinzs2048 

You might want to change it to periodical job before merging.

kevinzs2048:
@ahrtr Thanks! Modify to periodical job now.

The perviously submit for the e2e-arm64 job is green.

.github/workflows/e2e-arm64.yaml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  schedule:
+    - cron: '0 0/4 * * *' # runs every 4 hours
```
ahrtr, line 4: 4 hours is too frequent to me. I would suggest to run it daily.

kevinzs2048: sure, change to daily, nightly test at 1 am.

ahrtr:
Please squash the commits.

kevinzs2048:
> Please squash the commits.

Thanks! Done

ahrtr (approved):
LGTM

Thanks @kevinzs2048

.github/workflows/e2e-arm64.yaml
serathius: Labels are per job not per forkflow.
```
jobs:
  test:
    runs-on: [Linux, ARM64]
```
You can change it to:

```
jobs:
  amd64:
    runs-on: ubuntu-latest
    ...
  arm64:
    runs-on: [Linux, ARM64]
```

serathius (approved):

PR #15235, opened 2023-02-02
tests: Reproduce #15220 via manual progress notify

serathius:
cc @ahrtr @wojtek-t @scpmw

Issue #9855, opened 2018-06-15
Add progress operation to watch API


jpbetz:
There is currently no general way to check how far a watch stream has progressed relative to a specific etcd `revision`.  The `revisions` in watch events are only enough to establish a lower bound for the `revision` the watch stream has progressed past. It would be useful for the client to be able to query "what's the progress of this watch stream?" and get a "this watch stream has received all events it is subscribed for up to revision N", where N is the highest revision the server can guarantee the response is accurate for.

E.g. if the latest event revision received on a stream is `r_event`, and the latest applied revision observed by the etcd server is `r_applied`, then we currently can only know that the watch progress `r_watch_progress` is . `r_event` <= `r_watch_progress` <= `r_applied`, and that range can be arbitrarily large because events are only received for changes to the watch range.

https://github.com/kubernetes/kubernetes/issues/59848 describes a use case where this would be valuable-- Determining if a client side cache that updated via a watch, is up-to-date to a particular `revision`. @xiang90's proposal in this case was to consider adding a progress operation to the watch API.

Since the Watch API s a bi-directional gRPC stream, the progress operation might look something like:
- Add a `WatchProgressRequest` union member to [`WatchRequest`](https://github.com/coreos/etcd/blob/a265872859dddb08649d8fa94a07305fb3163cb8/etcdserver/etcdserverpb/rpc.proto#L653)
- Add fields (or a union member?) to [`WatchResponse`](https://github.com/coreos/etcd/blob/a265872859dddb08649d8fa94a07305fb3163cb8/etcdserver/etcdserverpb/rpc.proto#L711) to identify progress responses and provide the max etcd storage `revision` that the server has no pending watch events to send to this watch stream for.

Specifically, the progress revision should be
- `>=` the revision of the most recent WatchResponse event sent in the watch stream to the client
- `<` the revision of WatchResponse events yet to be sent in the watch stream to the client
- `==` to the latest etcd storage revision the etcd member has applied if there are no pending events to to be send in the watch stream to the client

cc @xiang90 @gyuho @wenjiaswe

gyuho:
> Since the Watch API s a bi-directional gRPC stream, the progress operation might look something like:
> Add a WatchProgressRequest union member to WatchRequest

Q. How `WatchProgressRequest` would be different than  `WatchCreateRequest` (that returns all events since the specified revision)? Would it duplicate all fields from `WatchCreateRequest`?

jpbetz:
[...]

PR #15237, opened 2023-02-02
etcdserver: Send requested progress notifications through watchStream (fix #15220)

### LINKED ISSUE #15220: Requested watcher progress notifications are not synchronised with stream
### What happened?

When you request a progress notification from a watch stream, the header of the generated progress notification can indicate a revision for which not all watch responses have been generated yet.

This means that there is a (rare) race here - if the client receives the progress notification, and the connection drops immediately, it will presumably assume that it has received all updates in the database up to the revision indicated in the progress notification. It might therefore create a new watcher starting from this revision, which will cause a watch notification to be dropped.

A more easily reproduced variant of this problem actually happens when you create a new watcher with a start revision and immediately request a progress notification: The progress notification will typically arrive **before** any watch responses between the start revision and the current revision.

It looks like this might be because the implementation of the response code in https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/v3rpc/watch.go#L363 immediately sends the response through `sws.ctrlStream`, which can overtake `sws.watchStream` for the purpose of the `sendLoop`.

Note that this only happens for manually requested progress notifications. Timed progress notifications seem to go to great pain to ensure that the watchers are synced: The `progressTicker` in `sendLoop` uses `sws.watchStream.RequestProgress(id)`, which makes sure the watcher is synched and that the progress notification is ordered correctly relative to other watch responses.

In fact, it seems like the immediate problem can be fixed by changing how `pb.WatchRequest_ProgressRequest` is handled?
```go
			sws.mu.Lock()
			for id, _ := range sws.watchers {
                                sws.watchStream.RequestProgress(id)
 			}
			sws.mu.Unlock()
```

Where `sws.watchers` is a map introduced just to keep track of *all* active watchers. Not sure this is actually the right fix - at minimum it changes the semantics of progress notifications in some more-or-less subtle ways:
* Now we get separate progress notifications for every watcher (ID)
* Because `progress` in `watchable_store.go` just skips the progress notification for unsynced clients, it is no longer guaranteed that every progress notification request actually generates a progress notification

Not sure whether this would be a problem?

Here is a Python script that reproduces the issue:
```python
import queue
import rpc_pb2_grpc
import rpc_pb2
import grpc
# Open channel, create watch
with grpc.insecure_channel('localhost:12346') as channel:
    stub = rpc_pb2_grpc.WatchStub(channel)
    kv = rpc_pb2_grpc.KVStub(channel)
    test_key = b"test"
    out_queue = queue.Queue()
    out_queue.put(
        rpc_pb2.WatchRequest(create_request=rpc_pb2.WatchCreateRequest(
            key=test_key,
            range_end=test_key + b"\ff"
    )))
    # Receive watch events
    last_rev = None
    for i, resp in enumerate(stub.Watch(iter(out_queue.get, None))):
        # Log message
        print(resp)
        # Check for out-of-order
        if last_rev is not None and resp.header.revision < last_rev and resp.events:
            print(" ^^^ Out-of-order watch response found")
            exit(1)
        # Change key & request progress notificiation
        if last_rev is None or last_rev != resp.header.revision:
            out_queue.put(
                rpc_pb2.WatchRequest(progress_request=rpc_pb2.WatchProgressRequest()))
            kv.Put(rpc_pb2.PutRequest(key=test_key,value=str(i).encode()))
            last_rev = resp.header.revision

```

You might need to run multiple instances for a while to put enough stress on the server, but eventually you will get something like this:
```
header {
  cluster_id: 14841639068965178418
  member_id: 10276657743932975437
  revision: 373837
  raft_term: 2
}
watch_id: -1

header {
  cluster_id: 14841639068965178418
  member_id: 10276657743932975437
  revision: 373835
  raft_term: 2
}
events {
  kv {
    key: "test"
    create_revision: 2
    mod_revision: 373835
    version: 373834
    value: "46952"
  }
}

 ^^^ Out-of-order watch response found
```
Which demonstrates that the problem can happen: We have a progress notification for revision 373837 *before* a genuine watch notification for revision 373835.

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.5.0-pre
Git SHA: 772dfbfe3
Go Version: go1.18.1
Go OS/Arch: linux/amd64

$ etcdctl version
etcdctl version: 3.4.5
API version: 3.4
```
</details>

_No response_

_No response_

### Relevant log output

_No response_

scpmw:
Progress notifications requested using `ProgressRequest` were sent directly using the `ctrlStream`, which means that they could race against watch responses in the `watchStream`.

This would especially happen when the stream was not synced - e.g. if you requested a progress notification on a freshly created unsynced watcher, the notification would typically arrive indicating a revision for which not all watch responses had been sent.

This changes the behaviour so that v3rpc always goes through the watch stream, using a new RequestProgressAll function that closely matches the behaviour of the v3rpc code - i.e.

1. Generate a message with WatchId -1, indicating the revision for *all* watchers in the stream

2. Guarantee that a response is (eventually) sent

The latter might require us to defer the response until all watchers are synced, which is likely as it should be. Note that we do *not* guarantee that the number of progress notifications matches the number of requests, only that eventually at least one gets sent.

### DIFF
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -145,6 +145,10 @@ type serverWatchStream struct {
 	// records fragmented watch IDs
 	fragment map[mvcc.WatchID]bool
 
+	// indicates whether we have an outstanding global progress
+	// notification to send
+	deferredProgress bool
+
 	// closec indicates the stream is closed.
 	closec chan struct{}
 
@@ -173,6 +177,8 @@ func (ws *watchServer) Watch(stream pb.Watch_WatchServer) (err error) {
 		progress: make(map[mvcc.WatchID]bool),
 		prevKV:   make(map[mvcc.WatchID]bool),
 		fragment: make(map[mvcc.WatchID]bool),
+
+		deferredProgress: false,
 
 		closec: make(chan struct{}),
 	}
@@ -360,10 +366,16 @@ func (sws *serverWatchStream) recvLoop() error {
 			}
 		case *pb.WatchRequest_ProgressRequest:
 			if uv.ProgressRequest != nil {
-				sws.ctrlStream <- &pb.WatchResponse{
-					Header:  sws.newResponseHeader(sws.watchStream.Rev()),
-					WatchId: clientv3.InvalidWatchID, // response is not associated with any WatchId and will be broadcast to all watch channels
+				sws.mu.Lock()
+				// Ignore if deferred progress notification is already in progress
+				if !sws.deferredProgress {
+					// Request progress for all watchers,
+					// force generation of a response
+					if !sws.watchStream.RequestProgressAll() {
+						sws.deferredProgress = true
+					}
 				}
+				sws.mu.Unlock()
 			}
 		default:
 			// we probably should not shutdown the entire stream when
@@ -432,11 +444,15 @@ func (sws *serverWatchStream) sendLoop() {
 				Canceled:        canceled,
 			}
 
-			if _, okID := ids[wresp.WatchID]; !okID {
-				// buffer if id not yet announced
-				wrs := append(pending[wresp.WatchID], wr)
-				pending[wresp.WatchID] = wrs
-				continue
+			// Progress notifications can have WatchID -1
+			// if they announce on behalf of multiple watchers
+			if wresp.WatchID != clientv3.InvalidWatchID {
+				if _, okID := ids[wresp.WatchID]; !okID {
+					// buffer if id not yet announced
+					wrs := append(pending[wresp.WatchID], wr)
+					pending[wresp.WatchID] = wrs
+					continue
+				}
 			}
 
 			mvcc.ReportEventReceived(len(evs))
@@ -466,6 +482,11 @@ func (sws *serverWatchStream) sendLoop() {
 			if len(evs) > 0 && sws.progress[wresp.WatchID] {
 				// elide next progress update if sent a key update
 				sws.progress[wresp.WatchID] = false
+			}
+			if sws.deferredProgress {
+				if sws.watchStream.RequestProgressAll() {
+					sws.deferredProgress = false
+				}
 			}
 			sws.mu.Unlock()
 
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/pkg/v3/traceutil"
 	"go.etcd.io/etcd/server/v3/lease"
 	"go.etcd.io/etcd/server/v3/storage/backend"
@@ -41,6 +42,7 @@ var (
 type watchable interface {
 	watch(key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc)
 	progress(w *watcher)
+	progressAll(watchers map[WatchID]*watcher) bool
 	rev() int64
 }
 
@@ -475,14 +477,34 @@ func (s *watchableStore) addVictim(victim watcherBatch) {
 func (s *watchableStore) rev() int64 { return s.store.Rev() }
 
 func (s *watchableStore) progress(w *watcher) {
+	s.progressIfSync(map[WatchID]*watcher{w.id: w}, w.id)
+}
+
+func (s *watchableStore) progressAll(watchers map[WatchID]*watcher) bool {
+	return s.progressIfSync(watchers, clientv3.InvalidWatchID)
+}
+
+func (s *watchableStore) progressIfSync(watchers map[WatchID]*watcher, responseWatchID WatchID) bool {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	if _, ok := s.synced.watchers[w]; ok {
-		w.send(WatchResponse{WatchID: w.id, Revision: s.rev()})
-		// If the ch is full, this watcher is receiving events.
-		// We do not need to send progress at all.
+	// Any watcher unsynced?
+	for _, w := range watchers {
+		if _, ok := s.synced.watchers[w]; !ok {
+			return false
+		}
 	}
+
+	// If all watchers are synchronised, send out progress
+	// notification on first watcher. Note that all watchers
+	// should have the same underlying stream, and the progress
+	// notification will be broadcasted client-side if required
+	// (see dispatchEvent in client/v3/watch.go)
+	for _, w := range watchers {
+		w.send(WatchResponse{WatchID: responseWatchID, Revision: s.rev()})
+		return true
+	}
+	return true
 }
 
 type watcher struct {
--- a/server/storage/mvcc/watcher.go
+++ b/server/storage/mvcc/watcher.go
@@ -58,6 +58,13 @@ type WatchStream interface {
 	// of the watchers since the watcher is currently synced.
 	RequestProgress(id WatchID)
 
+	// RequestProgressAll requests a progress notification for all
+	// watchers sharing the stream.  If all watchers are synced, a
+	// progress notification with watch ID -1 will be sent to an
+	// arbitrary watcher of this stream, and the function returns
+	// true.
+	RequestProgressAll() bool
+
 	// Cancel cancels a watcher by giving its ID. If watcher does not exist, an error will be
 	// returned.
 	Cancel(id WatchID) error
@@ -187,4 +194,10 @@ func (ws *watchStream) RequestProgress(id WatchID) {
 		return
 	}
 	ws.watchable.progress(w)
+}
+
+func (ws *watchStream) RequestProgressAll() bool {
+	ws.mu.Lock()
+	defer ws.mu.Unlock()
+	return ws.watchable.progressAll(ws.watchers)
 }
--- a/tests/integration/v3_watch_test.go
+++ b/tests/integration/v3_watch_test.go
@@ -1397,3 +1397,71 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
 		t.Fatalf("expected %s watch, got %s", expected, minWatches)
 	}
 }
+
+// TestV3WatchProgressWaitsForSync checks that progress notifications
+// don't get sent until the watcher is synchronised
+func TestV3WatchProgressWaitsForSync(t *testing.T) {
+
+	// Disable for gRPC proxy, as it does not support requesting
+	// progress notifications
+	if integration.ThroughProxy {
+		t.Skip("grpc proxy currently does not support requesting progress notifications")
+	}
+
+	integration.BeforeTest(t)
+
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
+
+	client := clus.RandClient()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Write a couple values into key to make sure there's a
+	// non-trivial amount of history.
+	count := 1001
+	t.Logf("Writing key 'foo' %d times", count)
+	for i := 0; i < count; i++ {
+		_, err := client.Put(ctx, "foo", fmt.Sprintf("bar%d", i))
+		require.NoError(t, err)
+	}
+
+	// Create watch channel starting at revision 1 (i.e. it starts
+	// unsynced because of the update above)
+	wch := client.Watch(ctx, "foo", clientv3.WithRev(1))
+
+	// Immediately request a progress notification. As the client
+	// is unsynchronised, the server will have to defer the
+	// notification internally.
+	err := client.RequestProgress(ctx)
+	require.NoError(t, err)
+
+	// Verify that we get the watch responses first. Note that
+	// events might be spread across multiple packets.
+	var event_count = 0
+	for event_count < count {
+		wr := <-wch
+		if wr.Err() != nil {
+			t.Fatal(fmt.Errorf("watch error: %w", wr.Err()))
+		}
+		if wr.IsProgressNotify() {
+			t.Fatal("Progress notification from unsynced client!")
+		}
+		if wr.Header.Revision != int64(count+1) {
+			t.Fatal("Incomplete watch response!")
+		}
+		event_count += len(wr.Events)
+	}
+
+	// ... followed by the requested progress notification
+	wr2 := <-wch
+	if wr2.Err() != nil {
+		t.Fatal(fmt.Errorf("watch error: %w", wr2.Err()))
+	}
+	if !wr2.IsProgressNotify() {
+		t.Fatal("Did not receive progress notification!")
+	}
+	if wr2.Header.Revision != int64(count+1) {
+		t.Fatal("Wrong revision in progress notification!")
+	}
+}
--- a/tests/robustness/linearizability_test.go
+++ b/tests/robustness/linearizability_test.go
@@ -36,10 +36,11 @@ const (
 
 var (
 	LowTraffic = trafficConfig{
-		name:        "LowTraffic",
-		minimalQPS:  100,
-		maximalQPS:  200,
-		clientCount: 8,
+		name:            "LowTraffic",
+		minimalQPS:      100,
+		maximalQPS:      200,
+		clientCount:     8,
+		requestProgress: false,
 		traffic: traffic{
 			keyCount:     10,
 			leaseTTL:     DefaultLeaseTTL,
@@ -56,10 +57,11 @@ var (
 		},
 	}
 	HighTraffic = trafficConfig{
-		name:        "HighTraffic",
-		minimalQPS:  200,
-		maximalQPS:  1000,
-		clientCount: 12,
+		name:            "HighTraffic",
+		minimalQPS:      200,
+		maximalQPS:      1000,
+		clientCount:     12,
+		requestProgress: false,
 		traffic: traffic{
 			keyCount:     10,
 			largePutSize: 32769,
@@ -67,6 +69,22 @@ var (
 			writes: []requestChance{
 				{operation: Put, chance: 85},
 				{operation: MultiOpTxn, chance: 10},
+				{operation: LargePut, chance: 5},
+			},
+		},
+	}
+	ReqProgTraffic = trafficConfig{
+		name:            "RequestProgressTraffic",
+		minimalQPS:      200,
+		maximalQPS:      1000,
+		clientCount:     12,
+		requestProgress: true,
+		traffic: traffic{
+			keyCount:     10,
+			largePutSize: 8196,
+			leaseTTL:     DefaultLeaseTTL,
+			writes: []requestChance{
+				{operation: Put, chance: 95},
 				{operation: LargePut, chance: 5},
 			},
 		},
@@ -141,6 +159,14 @@ func TestRobustness(t *testing.T) {
 			e2e.WithSnapshotCount(100),
 		),
 	})
+	scenarios = append(scenarios, scenario{
+		name:      "Issue15220",
+		failpoint: RandomOneNodeClusterFailpoint,
+		traffic:   &ReqProgTraffic,
+		config: *e2e.NewConfig(
+			e2e.WithClusterSize(1),
+		),
+	})
 	snapshotOptions := []e2e.EPClusterOption{
 		e2e.WithGoFailEnabled(true),
 		e2e.WithSnapshotCount(100),
@@ -191,7 +217,7 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, config e2
 	forcestopCluster(r.clus)
 
 	watchProgressNotifyEnabled := r.clus.Cfg.WatchProcessNotifyInterval != 0
-	validateWatchResponses(t, r.responses, watchProgressNotifyEnabled)
+	validateWatchResponses(t, r.responses, traffic.requestProgress || watchProgressNotifyEnabled)
 
 	r.events = watchEvents(r.responses)
 	validateEventsMatch(t, r.events)
@@ -218,7 +244,7 @@ func runScenario(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.Et
 		return nil
 	})
 	g.Go(func() error {
-		responses = collectClusterWatchEvents(ctx, t, clus, maxRevisionChan)
+		responses = collectClusterWatchEvents(ctx, t, clus, maxRevisionChan, traffic.requestProgress)
 		return nil
 	})
 	g.Wait()
--- a/server/storage/mvcc/watcher_test.go
+++ b/server/storage/mvcc/watcher_test.go
@@ -25,6 +25,7 @@ import (
 	"go.uber.org/zap/zaptest"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/lease"
 	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
 )
@@ -332,6 +333,55 @@ func TestWatcherRequestProgress(t *testing.T) {
 
 	w.RequestProgress(id)
 	wrs := WatchResponse{WatchID: id, Revision: 2}
+	select {
+	case resp := <-w.Chan():
+		if !reflect.DeepEqual(resp, wrs) {
+			t.Fatalf("got %+v, expect %+v", resp, wrs)
+		}
+	case <-time.After(time.Second):
+		t.Fatal("failed to receive progress")
+	}
+}
+
+func TestWatcherRequestProgressAll(t *testing.T) {
+	b, _ := betesting.NewDefaultTmpBackend(t)
+
+	// manually create watchableStore instead of newWatchableStore
+	// because newWatchableStore automatically calls syncWatchers
+	// method to sync watchers in unsynced map. We want to keep watchers
+	// in unsynced to test if syncWatchers works as expected.
+	s := &watchableStore{
+		store:    NewStore(zaptest.NewLogger(t), b, &lease.FakeLessor{}, StoreConfig{}),
+		unsynced: newWatcherGroup(),
+		synced:   newWatcherGroup(),
+		stopc:    make(chan struct{}),
+	}
+
+	defer cleanup(s, b)
+
+	testKey := []byte("foo")
+	notTestKey := []byte("bad")
+	testValue := []byte("bar")
+	s.Put(testKey, testValue, lease.NoLease)
+
+	// Create watch stream with watcher. We will not actually get
+	// any notifications on it specifically, but there needs to be
+	// at least one Watch for progress notifications to get
+	// generated.
+	w := s.NewWatchStream()
+	w.Watch(0, notTestKey, nil, 1)
+
+	w.RequestProgressAll()
+	select {
+	case resp := <-w.Chan():
+		t.Fatalf("unexpected %+v", resp)
+	default:
+	}
+
+	s.syncWatchers()
+
+	w.RequestProgressAll()
+	wrs := WatchResponse{WatchID: clientv3.InvalidWatchID, Revision: 2}
 	select {
 	case resp := <-w.Chan():
 		if !reflect.DeepEqual(resp, wrs) {
--- a/tests/robustness/traffic.go
+++ b/tests/robustness/traffic.go
@@ -109,11 +109,12 @@ func simulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
 }
 
 type trafficConfig struct {
-	name        string
-	minimalQPS  float64
-	maximalQPS  float64
-	clientCount int
-	traffic     Traffic
+	name            string
+	minimalQPS      float64
+	maximalQPS      float64
+	clientCount     int
+	traffic         Traffic
+	requestProgress bool // Request progress notifications while watching this traffic
 }
 
 type Traffic interface {
--- a/tests/robustness/watch.go
+++ b/tests/robustness/watch.go
@@ -31,7 +31,7 @@ import (
 	"go.etcd.io/etcd/tests/v3/robustness/model"
 )
 
-func collectClusterWatchEvents(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, maxRevisionChan <-chan int64) [][]watchResponse {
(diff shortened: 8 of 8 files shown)


scpmw:
Previous discussion on this commit: https://github.com/scpmw/etcd/commit/8de21779ba755c251f2cd5a4383ca99d31aeb7c0#r98961857

ahrtr:
Please signoff the commit and resolve the pipeline errors.

cc @mitake

ahrtr:
```
$ git rebase HEAD~1 --signoff
$ git push origin guarantee_progress_notify_order -f
```

scpmw:
Done, I think.

serathius (changes requested):
Didn't dig into code, however I run https://github.com/etcd-io/etcd/pull/15235 on this and found an issue with implementation. 

Looks like same watch event is emitted in twice. From the local run:
```
panic: duplicate operation in two responses on revision 332
---
first:
{
    "Header": {
        "cluster_id": 4512387775551231959,
        "member_id": 11173105686495536258,
        "revision": 332,
        "raft_term": 2
    },
    "Events": [
        {
            "kv": {
                "key": "MA==",
                "create_revision": 2,
                "mod_revision": 332,
                "version": 78,
                "value": "MzI4"
            }
        }
    ],
    "CompactRevision": 0,
    "Canceled": false,
    "Created": false
}
---
second:
{
    "Header": {
        "cluster_id": 4512387775551231959,
        "member_id": 11173105686495536258,
        "revision": 333,
        "raft_term": 3
    },
    "Events": [
        {
            "kv": {
                "key": "MA==",
                "create_revision": 2,
                "mod_revision": 332,
                "version": 78,
                "value": "MzI4"
            }
        },
        {
            "kv": {
                "key": "Mw==",
                "create_revision": 3,
                "mod_revision": 333,
                "version": 92,
                "value": "MzMw"
            }
        }
    ],
    "CompactRevision": 0,
    "Canceled": false,
    "Created": false
}

```
This shows two consecutive watch responses that include same event. First watch returned events for revisions 332-332 and second 332-333. 
This situation doesn't happen on main branch nor when manual progress notifications are disabled.

scpmw:
Uh, that's concerning. This doesn't happen for my test script - how do I reproduce this? I tried `make test-linearizability` for your branch, but that seems to yield me a "connection refused" error from a goPanicFailpoint. I hope I'm not overlooking something obvious here?

serathius:
Sorry I still need to document [linearizability tests](https://github.com/etcd-io/etcd/blob/main/.github/workflows/linearizability-template.yaml) properly. Linearizability tests require a etcd server binary compiled with gofail enabled. To do that please run
```
make gofail-enable
make build
make gofail-diable
```

scpmw:
Ah, thanks, that made it work. Yes, I am getting the same error, but this might be a separate issue... I disabled the progress notification entirely server-side (i.e. commented out the code answering `ProgressRequest`) for both my branch and the version from your branch, and after disabling the `Expected at least one notify` check it still reproduced your error. So fairly confident that my fix is not what made this appear (that would have been *really* puzzling).

I'll see whether I can dig some more to find out why it happened in the first place. Using my test script I was unable to reproduce responses getting sent twice, though interestingly enough I found other instances where the revision went backwards. This time it was on actual watch responses, not progress notifications though. Could even be a connection there - if your script does reconnects, a wrong revision in the header might lead to duplicates?

server/storage/mvcc/watchable_store.go
```diff
@@ -482,6 +496,27 @@ func (s *watchableStore) progress(w *watcher) {
 	}
 }
 
+func (s *watchableStore) progress_all(force bool) {
+	s.mu.RLock()
```
mitake, line 500: `progress_all()` updates `s.progressOnSync`, so it's better to use `s.mu.Lock()`?
Note that I made sure that this line isn't related to the duplicated events issue. I'm still trying to understand why it happens.

server/storage/mvcc/watchable_store.go
```diff
@@ -482,6 +496,27 @@ func (s *watchableStore) progress(w *watcher) {
 	}
 }
 
+func (s *watchableStore) progress_all(force bool) {
```
mitake, line 499: `force` seems to be unnecessary because it never becomes false.

scpmw: True. It is a non-obvious semantic difference between `progress` and `progress_all` though, so I thought it would be better to make it a bit more obvious.

ahrtr:
The PR https://github.com/etcd-io/etcd/pull/15248 should can fix the duplicate event issue. No matter how we fix the `ProgressRequest` issue, we should guarantee that the `watchRequest.rev` never decrease. cc @mitake

ahrtr:
Mark this PR as draft until it's confirmed that it indeed fixes the `RequestProgress` issue.

ahrtr:
FYI. The root cause of the duplicated event issue:  https://github.com/etcd-io/etcd/pull/15248#issuecomment-1420125977

ahrtr:
Confirmed that this PR indeed can fix this issue https://github.com/etcd-io/etcd/issues/15220. But the solution has major issue, because it requires all watchers (from all clients) to be synced, obviously it isn't good. Instead, we should only require all [the watchers which share the same gRPC stream](https://github.com/etcd-io/etcd/blob/9bae01099413b40779000168dcef4d71501d1ac4/server/storage/mvcc/watcher.go#L101) to be synced. 

We can manage to resolve it, but I don't see a good reason to continue to support it as we already have timed progress notification. It may complicate the overall implementation, and is also error prone. The source of the issue is coming from the addition of [RequestProgress](https://github.com/etcd-io/etcd/blob/9bae01099413b40779000168dcef4d71501d1ac4/client/v3/watch.go#L90). Please read https://github.com/etcd-io/etcd/issues/9855 and https://github.com/kubernetes/kubernetes/issues/59848 to get more background. The motivation of adding the `RequestProgress` is to get the answer to question "what's my progress", and accordingly know whether the local cache is up to date. Read https://github.com/kubernetes/kubernetes/issues/59848#issuecomment-397690347 for a quick reference. But the existing timed progress notification should be enough.

I am thinking probably we should deprecate the manual `RequestProgress`. Actually K8s doesn't use the manual `RequestProgress` at all.  See https://github.com/etcd-io/etcd/pull/15260.  @mitake @ptabor @serathius @spzala 
@wojtek-t @liggitt @dims any objections or comments?

liggitt:
@wojtek-t or @jpbetz might know more about the plans for requested progress notification than I do... I thought there was still a desire to use that in Kubernetes and it just hadn't happened yet

scpmw:
@ahrtr - ah, good point. I had overlooked that the `watchable` is shared between multiple `watchStream`s. That obviously meant that the fix was wrong. Should be fixed now, also addressing @mitake's comments.

Deprecating manual progress notifications would send me back to the drawing board on how to determine the database state of watched keys from a watch stream with multiple watchers - this was the only feature in the API that looked like it would cross-synchronise watchers (after all, the documentation effectively claimed it would!). So not a fan, but obviously not up to me in the end.

server/storage/mvcc/watchable_store.go
```diff
@@ -482,6 +483,26 @@ func (s *watchableStore) progress(w *watcher) {
...
+	for w, _ := range watchers {
+		if _, ok := s.synced.watchers[watchers[w]]; !ok {
+			return false
+		}
+	}
+
+	// If all watchers are synchronised, send out progress notification
+	// on first watcher
+	for w, _ := range watchers {
+		watchers[w].send(WatchResponse{WatchID: -1, Revision: s.rev()})
```
ahrtr, line 500: Please use the constant `InvalidWatchID`

server/storage/mvcc/watcher.go
```diff
@@ -58,6 +58,13 @@ type WatchStream interface {
 	// of the watchers since the watcher is currently synced.
 	RequestProgress(id WatchID)
 
+	// RequestProgressAll requests a progress notification for the
+	// entire watcher group.  The responses will be sent through
```
ahrtr, line 62: Please change the comment to something like below, (not a native English speaker, please feel free to adjust the wording.)
```
RequestProgressAll requests a progress notification for the
all watchers (substreams) which share the same gRPC stream. 
......
```

server/storage/mvcc/watchable_store.go
```diff
@@ -41,6 +41,7 @@ var (
 type watchable interface {
 	watch(key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc)
 	progress(w *watcher)
+	progress_all(watchers map[WatchID]*watcher) bool
```
ahrtr, line 44:
```suggestion
	progressAll(watchers map[WatchID]*watcher) bool
```

ahrtr:
> this was the only feature in the API that looked like it would cross-synchronise watchers

Yes, this is still true. The overall complexity seems acceptable, and that's good news, so I am not insist on deprecating it. But the manual progress notification needs to require the write lock, so I think it may cause reduce of performance if users use it too much. 

So the timed progress notification is still recommended.

server/storage/mvcc/watchable_store.go
```diff
@@ -482,6 +483,26 @@ func (s *watchableStore) progress(w *watcher) {
 	}
 }
 
+func (s *watchableStore) progress_all(watchers map[WatchID]*watcher) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
ahrtr, line 488: I think we only need to require read lock here, in the same way as the existing `(*watchableStore) progress`

```suggestion
	s.mu.RLock()
	defer s.mu.RUnlock()
```

serathius:
@scpmw Can you add an integration test like in https://github.com/etcd-io/etcd/pull/15248/files#diff-78cf23d7d421b7d0e6c98ba46b62ceaa45280f11f2a7227f47cc30fb735f16e7R879 ?

scpmw:
It is somewhat more expensive than the old manual progress notification for sure. However for my purposes the fallback would be to have a complete transaction verifying (essentially) all watched modrevisions - which would be a lot worse.

I will work on an integration test, yes. For the moment still puzzling over why this MR plus https://github.com/etcd-io/etcd/pull/15248 still doesn't pass https://github.com/etcd-io/etcd/pull/15235 - for some reason there's *still* duplicated events?

ahrtr:
> I will work on an integration test, yes. For the moment still puzzling over why this MR plus #15248 still doesn't pass #15235 - for some reason there's _still_ duplicated events?

@scpmw Note the test case in https://github.com/etcd-io/etcd/pull/15235 has issue. It may regard different DELETE operation as duplicated operation (For details, please see the third commit in https://github.com/etcd-io/etcd/pull/15262, at line 119 ~ 121 in test/linearizablity/watch.go).

server/storage/mvcc/watcher.go
scpmw: Re-worded it. Felt like it might be the wrong level of abstraction to talk about gRPC (strictly speaking we don't know where the WatchResponses go) and substreams (that's a client-side idea?), so I left it just at "share the same stream".

scpmw:
Addressed comments, added a simple integration test using the easy-to-reproduce case mentioned in #15220. It's technically a race, but I haven't seen `main` pass yet.

ahrtr:
A quick comment, please signoff the commits. Please read https://github.com/etcd-io/etcd/pull/15237/checks?check_run_id=11326356800. 

Will take a look at this PR later.

server/proxy/grpcproxy/watch.go
```diff
@@ -271,6 +272,14 @@ func (wps *watchProxyStream) recvLoop() error {
...
 			wps.delete(uv.CancelRequest.WatchId)
 			wps.lg.Debug("cancel watcher", zap.Int64("watcherId", uv.CancelRequest.WatchId))
+		case *pb.WatchRequest_ProgressRequest:
+			wps.mu.Lock()
+			for _, wps := range wps.watchers {
+				wps.requestedProgress = true
+			}
+			wps.mu.Unlock()
+			wps.ranges.wp.cw.RequestProgress(wps.ranges.wp.ctx)
+			wps.lg.Debug("request watcher progress")
```
ahrtr, line 282: This seems like a new feature. I would suggest to address in a separate discussion session.

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,65 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+	// is unsynchronised, the server will have to defer the
+	// notification internally.
+	err := client.RequestProgress(ctx)
+	require.NoError(t, err)
+
+	// Verify that we get the watch response first
+	wr := <-wch
+	if wr.Err() != nil {
+		t.Fatal(fmt.Errorf("watch error: %w", wr.Err()))
+		return
```
ahrtr, line 1436: Not needed
```suggestion
```

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,65 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+	}
+	if !wr2.IsProgressNotify() {
+		t.Fatal("Did not receive progress notification!")
+		return
+	}
+	if wr2.Header.Revision != int64(count+1) {
+		t.Fatal("Wrong revision in progress notification!")
+		return
+	}
+}
```
ahrtr, line 1461: The test case isn't clear to me. I suggest to create two goroutines, one keeps calling RequestProgress and writing data, and the other keeps checking the revision. 

The first goroutine:
```
go func() {
    defer func() {
        time.Sleep(10*time.Millisecond)
        cancel()
    }()
    for i:=0; i<100; i++ {
        // RequestProgress before writing data
        err := client.RequestProgress(ctx)
        require.NoError(t, err)

        _, err = client.Put(ctx, "foo", fmt.Sprintf("bar%d", i))
        require.NoError(t, err)

       // RequestProgress after writing data
        err := client.RequestProgress(ctx)
        require.NoError(t, err)

        time.Sleep(10*time.Millisecond)
    }
}()
```

The second goroutine:
```
go func() {
    wch := client.Watch(ctx, "foo")
    for resp := range wch {
        // Follow https://github.com/etcd-io/etcd/blob/c24ba4d3e85ef1147d9245f544c125e251ff03e8/tests/linearizability/watch.go#L101-L139
    }
}()

```

ahrtr: Please add this test case as the first commit, and confirm that it can easily reproduce the original issue. Then apply the fix as the second commit, and confirm that you can't reproduce the issue any more.

scpmw: Your proposal is actually a very different test case, which would especially have a much lower likelihood of reproducing the problem. The point here is that we
1. Create a database state where a key has 100+ revisions of history (no watcher active at this point!)
2. Create a new watcher that starts at revision 1 (and therefore is 100 revisions "behind" at creation time)
3. Immediately ask for a progress notification. This is highly likely to catch the watcher in a state where it has not "caught up" yet.

Before the patch, this typically gives you:
1. Confirmation that worker was created
2. Progress notification (revision 101 in header)
3. Watcher notification for values 1-100

Because it's basically a race between the watch and the control stream, and given the complexity in generating the response for all 100 revisions, the control stream basically always wins. From the client's perspective, this leaves a very obvious window where a connection drop would cause 100 revisions of history to be lost.

The fix basically switches the order between answers 2 and 3 (by deferring the progress notification). I hope that makes sense? Any way I could document it better?

server/proxy/grpcproxy/watch.go
scpmw: Yes, I see your point. I'll create a new PR.

ahrtr:
Please also rebase this PR. It should can resolve the `Go Vulnerability Checker` failure.

server/storage/mvcc/watchable_store.go
mitake: Sorry for my late response. The latest PR doesn't have the boolean flag and the parameter is a map of watchers, so it's clearer for me.
I think the semantics of plain `progress` is an independent topic. Focusing only one change in one PR might be good. Anyway, thanks for addressing my comment.

server/storage/mvcc/watchable_store.go
```diff
@@ -482,6 +484,26 @@ func (s *watchableStore) progress(w *watcher) {
...
+
+	// Any watcher unsynced?
+	for w, _ := range watchers {
+		if _, ok := s.synced.watchers[watchers[w]]; !ok {
+			return false
+		}
+	}
+
+	// If all watchers are synchronised, send out progress notification
+	// on first watcher
```
mitake, line 499: nit and optional: describing that the broadcasting mechanism is implemented in [client side](https://github.com/etcd-io/etcd/blob/main/client/v3/watch.go#L762) might be a little bit friendly? No strong preference at all.

server/proxy/grpcproxy/watch.go
scpmw: Done: https://github.com/etcd-io/etcd/pull/15339

tests/integration/v3_watch_test.go
scpmw: Well, it's a race, so there's never 100%. According to a series of runs, with 100 revisions it reproduces roughly 99.5% for me. With 1000 revisions it's solidly > 99.99% now according to my test, so let's stay with that.

scpmw:
Okay, should all be addressed now. I think we're getting closer (thanks everybody!), so I've cleared the draft status for now.

server/proxy/grpcproxy/watch.go
ahrtr: Please remove all grpcproxy related change from this PR.

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,100 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+	defer clus.Terminate(t)
+
+	client := clus.RandClient()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Write a couple values into key to make sure there's a
+	// non-trivial amount of history.
+	count := 1001
+	t.Log(fmt.Sprintf("Writing key 'foo' %d times", count))
```
ahrtr, line 1446: nit: 
```suggestion
	t.Logf("Writing key 'foo' %d times", count)
```

server/proxy/grpcproxy/watch.go
scpmw: The test case won't pass then - we either need to merge https://github.com/etcd-io/etcd/pull/15339 first or disable the test case for grpcproxy here. Not sure what'd be preferred?

ahrtr: Definitely the latter for now: `disable the test case for grpcproxy here.`. We need to backport this PR to 3.5 and 3.4, and usually we don't backport new features.

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,77 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Verify that we get the watch responses first. Note that
+	// events might be spread across multiple packets.
+	var event_count = 0
+	for event_count < count {
+		wr := <-wch
+		if wr.Err() != nil {
+			t.Fatal(fmt.Errorf("watch error: %w", wr.Err()))
+			return
```
ahrtr, line 1446: Note: t.Fatal & t.Fatalf will stop the execution of the test case, so the `return` isn't needed.
```suggestion
			t.Fatalf("watch error: %v", wr.Err())
```

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,77 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+		}
+		if wr.IsProgressNotify() {
+			t.Fatal("Progress notification from unsynced client!")
+			return
+		}
+		if wr.Header.Revision != int64(count+1) {
+			t.Fatal("Incomplete watch response!")
+			return
```
ahrtr, line 1454: ditto
```suggestion
			t.Fatal("Incomplete watch response!")
```

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,77 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+
+	// ... followed by the requested progress notification
+	wr2 := <-wch
+	if wr2.Err() != nil {
+		t.Fatal(fmt.Errorf("watch error: %w", wr2.Err()))
+		return
+	}
+	if !wr2.IsProgressNotify() {
+		t.Fatal("Did not receive progress notification!")
+		return
```
ahrtr, line 1467:
```suggestion
		t.Fatal("Did not receive progress notification!")
```

tests/integration/v3_watch_test.go
```diff
@@ -1397,3 +1397,77 @@ func TestV3WatchCloseCancelRace(t *testing.T) {
...
+	}
+	if wr2.Header.Revision != int64(count+1) {
+		t.Fatal("Wrong revision in progress notification!")
+		return
```
ahrtr, line 1471:
```suggestion
		t.Fatal("Wrong revision in progress notification!")
```

ahrtr (approved):
Overall looks good to me. 

We can also simplify the test case using `github.com/stretchr/testify`, but it isn't a big deal so I won't insist on it.

cc @mitake to double check.

ahrtr:
Thanks @scpmw for the great work!

serathius:
Will take a look at it today. @scpmw Could you maybe enable manual progress notification requesting in robustness tests?
The recently renamed linearizability tests similar to https://github.com/etcd-io/etcd/pull/15235/files

mitake:
LGTM from my side, thanks a lot @scpmw ! will wait comments from @serathius

scpmw:
> In the meantime, @scpmw could you maybe enable manual progress notification requesting in robustness tests?

Right, that seems appropriate. Will likely need some adjustments though, as requesting progress notifications on every answer should probably be restricted to particular robustness test scenarios? Some flag in the `trafficConfig`, maybe, moving generation of the requests to `simulateTraffic`?

Unfortunately I'll be traveling for at least another week, but will look into it once I'm back.

serathius:
any progress?

scpmw:
Sorry, took me a bit to get back to this. Solution is slightly awkward because now `trafficConfig` has impact on `watchMember`, but I couldn't think of a more elegant solution. At least it seems to reproduce the problem reliably, and the new watch response validation catches it nicely.

tests/robustness/linearizability_test.go
```diff
@@ -70,6 +72,22 @@ var (
 			},
 		},
 	}
+	ReqProgTraffic = trafficConfig{
```
serathius, line 75: No need for a separate traffic, I think we can enable requestProgress on default traffic.

scpmw: Not sure... Requesting progress notifications on every event is rather untypical behaviour, so I felt having it enabled more broadly might end up shadowing other bugs, or at least make debugging them messier. So it felt worth it to put in the extra effort to construct a specific traffic use case.

If you insist I can revert to a simpler approach. Would you suggest I also remove the traffic flag then, and just hard-code it into `watchMember`? That seemed like non-obvious behaviour to me...

serathius: That's a valid point, let's leave it as it is and revisit it when we introduce dynamic scenarios..

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -432,11 +445,15 @@ func (sws *serverWatchStream) sendLoop() {
...
 			}
 
-			if _, okID := ids[wresp.WatchID]; !okID {
-				// buffer if id not yet announced
-				wrs := append(pending[wresp.WatchID], wr)
-				pending[wresp.WatchID] = wrs
-				continue
+			// Progress notifications can have WatchID -1
+			// if they announce on behalf of multiple watchers
+			if !progressNotify || wresp.WatchID != clientv3.InvalidWatchID {
```
serathius, line 450: Why do we need to check both `!progressNotify` and `wresp.WatchID != clientv3.InvalidWatchID`? 
Based on the comment above a would conclude that `progressNotify => (wresp.WatchID == clientv3.InvalidWatchID)`.

server/storage/mvcc/watchable_store.go
```diff
@@ -485,6 +487,29 @@ func (s *watchableStore) progress(w *watcher) {
...
+		if _, ok := s.synced.watchers[watchers[w]]; !ok {
+			return false
+		}
+	}
+
+	// If all watchers are synchronised, send out progress
+	// notification on first watcher. Note that all watchers
+	// should have the same underlying stream, and the progress
+	// notification will be broadcasted client-side if required
+	// (see dispatchEvent in client/v3/watch.go)
```
serathius, line 505: This is not obvious when calling function. For me this function is `progressIfSynced(w *watcher, requiredSynced []*watcher)`. I would even go further and merge it with current `progress` function. In both cases we send progress when set of watchers are synced. Function `progress` just has `requiredSynced = []*watcher{w}`. 

Something like
```
func (s *watchableStore) progress(w *watcher) {
    s.progressIfSynced(w, []*watcher{w})
}

// progressIfSynced sends a progress to a watcher if all required watchers are synced. 
// Useful when wanting to send progress to a whole stream.
func (s *watchableStore) progressIfSynced(w *watcher, requiredSynched []*watcher) bool {
	s.mu.RLock()
	defer s.mu.RUnlock()

	for _, w := range requiredSynched {
		if _, ok := s.synced.watchers[w]; !ok {
    			return false
		}
	}
	for w, _ := range watchers {
		watchers[w].send(WatchResponse{WatchID: clientv3.InvalidWatchID, Revision: s.rev()})
		return true
	}
	return true
}
```

server/etcdserver/api/v3rpc/watch.go
scpmw: Progress notifications *only* have WatchID -1 *if* they are announced on behalf of multiple workers. The counter-example is timed progress notifications, which are specific to a worker, and could therefore be buffered here.

The other edge case here would be a non-notification (i.e. actual payload) with a watch ID of -1, which I am reasonably (?) sure never happen. After all, this would lead to them getting "buffered" forever. So I guess I agree, I'll fix that.

serathius:
Please rebase and let me know when it's ready for review.

scpmw:
Done - assuming we leave the `progress`/`progressAll` names.

ahrtr:
@scpmw Please resolve the pipeline failures, thx.

ahrtr:
@scpmw This PR is blocking 3.5.8. If this can't be resolved this week, then we have to take over it or remove it from 3.5.8 list.

serathius:
Sorry for conflicts, I recommend moving robustness tests changes to separate PR as there is a lot of development happening there.

scpmw:
Resolved conflicts, fixed the test case failure related to progres notifications - apologies, had timed progress notifications also use watchID -1 there for a bit.

ahrtr (approved):
LGTM

Thanks @scpmw 

cc @mitake @serathius PTAL

serathius (approved):

ahrtr:
@scpmw could you please backport this PR to 3.5 and 3.4?

scpmw:
I can give it a try - separate PRs, I assume? And I'd assume I can skip the linearisability / robustness tests, because that would basically require also backporting the entire infrastructure?

ahrtr:
> I can give it a try - separate PRs, I assume? And I'd assume I can skip the linearisability / robustness tests, because that would basically require also backporting the entire infrastructure?

Thanks @scpmw . Yes, please submit two separate PRs,  one for release-3.5, and the other for release-3.4 branch.  Regarding to robustness test, It's OK to skip it to me, but defer to @serathius to comment.

ahrtr:
@scpmw please backport this PR to release-3.5 firstly, because we need to release `3.5.8` firstly. Thank you!

ahrtr:
@scpmw Please let me know if you need my assistance on this. 

My goal is to get it done by the end of next Monday, so that we can get 3.5.8 out before KubeCon. @serathius Do you have concern on this?

serathius:
If the original PR is done, I don't see anything wrong with others finishing work by backporting them.

ahrtr:
> If the original PR is done, I don't see anything wrong with others finishing work by backporting them

Agreed, but @scpmw already said he can give it a try. I will take over if no any response till next Monday.

scpmw:
You guys move fast :)

Haven't got to it yet, and likely won't until the end of the week. I'm not confident enough in my general Go skills to squeeze something like this into the margins around Easter... So if you need this done more quickly, please don't wait for me.

serathius:
This PR introduced another bug https://github.com/etcd-io/etcd/pull/17557

@scpmw @ahrtr do you know why we decided to defer progress notification until an event? It makes sense that we don't send a progress notification if any watch is unsynced, but why we defer notification? There is no guarantee that event comes.

> The latter might require us to defer the response until all watchers are synced, which is likely as it should be. Note that we do not guarantee that the number of progress notifications matches the number of requests, only that eventually at least one gets sent.

Eventually is good, but we currently get never.

scpmw:
Right. Mea culpa - I actually found this issue myself. The only situation this really happens is indeed on creation, because typically the reason we aren't synchronised is because there are messages left to send. So I tested that it can be fixed by just sending an "empty" message downstream in that particular case (i.e. at https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/watchable_store.go#L378). But this struck me as a fairly ugly way to approach this and I subsequently got a bit stuck trying to think of a better way.

Your proposed solution is to just not send a response to a progress request if the stream isn't fully synchronised? I think the key comment is in your patch:

>   As the client is unsynchronised, the server will not sent any notification, as client can infer progress from events.

With your patch that is not quite true for two reasons. Firstly - ironically - in the problematic scenario (create a new watcher on a "quiet" key, immediately send progress notification request) the client gets nothing back at all, so the client can in fact not infer progress.

Secondly - and more subtly, so apologies in advance if I get this slightly wrong - if you see "progress" as defined in https://github.com/etcd-io/etcd/blob/main/api/etcdserverpb/rpc.proto:

>  // For watch progress responses, the header.revision indicates progress. All future events
>  // received in this stream are guaranteed to have a higher revision number than the
>  // header.revision number.

Then that is something you cannot tell from standard events, as in a situation where we are watching many keys it is always possible that you will more events for a certain revision. Therefore there are situations where you really want to force a progress notification to get that particular guarantee, not because you are particularly worried that you might be falling that compaction might have caught up with you.

Now admittedely this could be characterised as somewhat misusing progress notifications, and your change isn't even catastrophic (you'd just have to poll a bit until you get a notification back). There might even be better conceptually better ways to do this - e.g. have some sort of flag in the response of whether there are more events at this revision? However, that introduces a protocol change and wouldn't even be that much easier to implement. So that's where I got stuck on this question...

serathius:
> Then that is something you cannot tell from standard events, as in a situation where we are watching many keys it is always possible that you will more events for a certain revision. Therefore there are situations where you really want to force a progress notification to get that particular guarantee, not because you are particularly worried that you might be falling behind enough that compaction might have caught up with you.

In standard events you just use `event.Kv.ModRevision` to get progress.

PR #15238, opened 2023-02-03
tests: Linearizability tests etcd logs

anushkamittal20:
We want linearizability tests etcd logs to go to artifacts and not spam output to handle their overwhelming volume.

Signed-off-by: Anushka Mittal <anumittal4641@gmail.com>


lavacat:
Can you describe the use case a bit more? It seams this is only for `EXPECT_DEBUG` case. I think users expect test logs to go to stdout. Users can redirect to file if they want.

pkg/expect/expect.go
```diff
@@ -122,22 +124,82 @@ func (ep *ExpectProcess) read() {
...
+			}
+			//fmt.Printf("%s (%s) (%d): %s", ep.cmd.Path, ep.cfg.name, ep.cmd.Process.Pid, l)
 		}
 		ep.lines = append(ep.lines, l)
 		ep.count++
 	}
-
-	// we're checking the error here at the bottom to ensure any leftover reads are still taken into account
+	// Create output file
+	out, err3 := os.Create("logs.tar.gz")
```
serathius, line 148: No need to gzip. Let's start with simple writing to file.

pkg/expect/expect.go
```diff
@@ -122,22 +124,82 @@ func (ep *ExpectProcess) read() {
 func (ep *ExpectProcess) tryReadNextLine(r *bufio.Reader) error {
 	printDebugLines := os.Getenv("EXPECT_DEBUG") != ""
 	l, err := r.ReadString('\n')
-
+	f, err1 := os.Create("logs.txt")
```
serathius, line 127: What happens if file exits? Can you rewrite it so it:
* Doesn't write to current directory but to test tmp directory like `t.TempDir`.
* Each etcd cluster member writes to separate file that can be recognised by name e.g. `member-1.log`
*  As member process can be restarted, we should not override the file but append to it's end.

pkg/expect/expect.go
```diff
@@ -122,22 +124,82 @@ func (ep *ExpectProcess) read() {
...
+		fmt.Printf("error: %v", err1)
+	}
+	defer f.Close()
 	ep.mu.Lock()
 	defer ep.mu.Unlock()
 
 	if l != "" {
 		if printDebugLines {
-			fmt.Printf("%s (%s) (%d): %s", ep.cmd.Path, ep.cfg.name, ep.cmd.Process.Pid, l)
+			logVal := fmt.Sprintf("%s (%s) (%d): %s", ep.cmd.Path, ep.cfg.name, ep.cmd.Process.Pid, l)
```
serathius, line 137: Please provide an option to configure if logs are written to file. Depending when and where we run the test we might prefer to get logs to stdout

PR #15246, opened 2023-02-06
etcdserver: add verification on the watch events

ahrtr:


ahrtr:
This verification may not be valid because the clients may watch starting from an old rev.

PR #15248, opened 2023-02-06
clientv3: correct `nextRev` on receiving progress notification response

ahrtr:
The manual ProgressRequest has issue, it may send inconsistent rev back to client. We should guard the `watchRequest.rev` from decreasing, otherwise, the client may receive duplicate events.

Confirmed that this PR can fix the duplicated event issue.

### DIFF
--- a/client/v3/watch.go
+++ b/client/v3/watch.go
@@ -880,12 +880,13 @@ func (w *watchGrpcStream) serveSubstream(ws *watcherStream, resumec chan struct{
 				}
 			} else {
 				// current progress of watch; <= store revision
-				nextRev = wr.Header.Revision
+				nextRev = wr.Header.Revision + 1
 			}
 
 			if len(wr.Events) > 0 {
 				nextRev = wr.Events[len(wr.Events)-1].Kv.ModRevision + 1
 			}
+
 			ws.initReq.rev = nextRev
 
 			// created event is already sent above,
--- a/tests/integration/v3_watch_test.go
+++ b/tests/integration/v3_watch_test.go
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
@@ -868,6 +870,99 @@ func TestV3WatchMultipleEventsPutUnsynced(t *testing.T) {
 	rok, nr := waitResponse(wStream, 1*time.Second)
 	if !rok {
 		t.Errorf("unexpected pb.WatchResponse is received %+v", nr)
+	}
+}
+
+// TestV3WatchProgressOnMemberRestart verifies the client side doesn't
+// receive duplicated events.
+// Refer to https://github.com/etcd-io/etcd/pull/15248#issuecomment-1423225742.
+func TestV3WatchProgressOnMemberRestart(t *testing.T) {
+	integration.BeforeTest(t)
+
+	clus := integration.NewCluster(t, &integration.ClusterConfig{
+		Size:                        1,
+		WatchProgressNotifyInterval: time.Second,
+	})
+	defer clus.Terminate(t)
+
+	client := clus.RandClient()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	errC := make(chan error, 1)
+	doneC := make(chan struct{}, 1)
+	progressNotifyC := make(chan struct{}, 1)
+	go func() {
+		defer close(doneC)
+
+		var (
+			lastWatchedModRevision  int64
+			gotProgressNotification bool
+		)
+
+		wch := client.Watch(ctx, "foo", clientv3.WithProgressNotify())
+		for wr := range wch {
+			if wr.Err() != nil {
+				errC <- fmt.Errorf("watch error: %w", wr.Err())
+				return
+			}
+
+			if wr.IsProgressNotify() {
+				// We need to make sure at least one progress notification
+				// is received after receiving the normal watch response
+				// and before restarting the member.
+				if lastWatchedModRevision > 0 {
+					gotProgressNotification = true
+					progressNotifyC <- struct{}{}
+				}
+				continue
+			}
+
+			for _, event := range wr.Events {
+				if event.Kv.ModRevision <= lastWatchedModRevision {
+					errC <- fmt.Errorf("got an unexpected revision: %d, lastWatchedModRevision: %d",
+						event.Kv.ModRevision,
+						lastWatchedModRevision)
+					return
+				}
+				lastWatchedModRevision = event.Kv.ModRevision
+			}
+
+			if gotProgressNotification {
+				return
+			}
+		}
+	}()
+
+	// write the key before the member restarts
+	t.Log("Writing key 'foo'")
+	_, err := client.Put(ctx, "foo", "bar1")
+	require.NoError(t, err)
+
+	// make sure at least one progress notification is received
+	// before restarting the member
+	t.Log("Waiting for the progress notification")
+	<-progressNotifyC
+
+	// restart the member
+	t.Log("Restarting the member")
+	clus.Members[0].Stop(t)
+	clus.Members[0].Restart(t)
+	clus.WaitMembersForLeader(t, clus.Members)
+
+	// write the same key again after the member restarted
+	t.Log("Writing the same key 'foo' again after restarting the member")
+	_, err = client.Put(ctx, "foo", "bar2")
+	require.NoError(t, err)
+
+	t.Log("Waiting for result")
+	select {
+	case err := <-errC:
+		t.Fatal(err)
+	case <-doneC:
+		t.Log("Done")
+	case <-time.After(10 * time.Second):
+		t.Fatal("Timed out waiting for the response")
 	}
 }
 


ahrtr:
cc @mitake

serathius:
Not sure what's the purpose of this change. Bumping revision to handle manual progress request might have an adverse effect of confusing the client. 
Server reporting lower revision is not expected and a strong indicated of something being really wrong. Client should not try to patch it as 

I it would be better if client just hard failed and refuse to handle invalid response from server (panic, return error or reset watch). 
cc @liggitt @wojtek-t

serathius:
@ptabor

ahrtr:
Note I am still in progress of debugging this issue, just marked the PR as draft.

ahrtr:
It turns out to be a clientv3's minor issue. 

When the clientv3 receives a progress notification response, it sets `nextRev` to the current revision (`wr.Header.Revision`), it isn't correct. It should set `nextRev` to `wr.Header.Revision + 1`. Otherwise, the client side may receive duplicated events when etcdserver restarted for whatever reason. 

Let's work with an example,
1. Assuming the latest revision is `79`, and the client side already received the event with rev `79`;
2. etcdserver keeps sending progress notification with `79`, and the client side update `nextRev` to 79 (_It's the root cause of this issue_);
3. If the etcdserver crashes for whatever reason, the client side re-sends the `watchRequest` with rev `79`;
4. When the etcdserver starts again, it sends watchResponse starting from rev `79`. Accordingly the client side receives the event with rev `79` twice.

Note that it's pretty easy to reproduce this issue using https://github.com/etcd-io/etcd/pull/15235. After applying this PR, it can't be reproduced anymore.

We need to backport this PR to both 3.4 and 3.5. @mitake PTAL.

ahrtr:
@wojtek-t does this issue have any impact on K8s?

serathius:
@ahrtr As we have established, K8s doesn't use manual progress notify so it's not affected.

ahrtr:
> @ahrtr As we have established, K8s doesn't use manual progress notify so it's not affected.

This isn't correct. Both manual progress notification and timed progress notification have this issue. Note that it can only happen when etcdserver restarts (`crashes and starts again`) and the last watch response received by the client side is a progress notification. I already reproduced it using https://github.com/etcd-io/etcd/pull/15235 (with some minor modification). Please see the example workflow in https://github.com/etcd-io/etcd/pull/15248#issuecomment-1420125977

Could you evaluate the impact on K8s? cc @wojtek-t @dims @liggitt @nikhita @neolit123

serathius:
I don't understand. https://github.com/etcd-io/etcd/pull/15235  just reproduces manual progress notification issue. If you did any modifications to it please publish it because again we are playing "she said he said". 

I have already added verification for duplicated revisions in https://github.com/etcd-io/etcd/pull/15250 and validated timed progress notification https://github.com/etcd-io/etcd/pull/15257.

client/v3/watch.go
```diff
@@ -880,12 +880,13 @@ func (w *watchGrpcStream) serveSubstream(ws *watcherStream, resumec chan struct{
 				}
 			} else {
 				// current progress of watch; <= store revision
-				nextRev = wr.Header.Revision
+				nextRev = wr.Header.Revision + 1
```
ptabor, line 883: I haven't read this code deeply... but please make sure the 'fragment'  field is properly interpreted:

https://github.com/etcd-io/etcd/blob/586eaccc4d28d4c2af507eaef0051d21a7e4b04c/api/etcdserverpb/rpc.proto#L835

I assume that in case of fragmented response, there might be multiple responses with the same revision.

ahrtr:
> I don't understand. #15235 just reproduces manual progress notification issue. If you did any modifications to it please publish it because again we are playing "she said he said".

First of all, please make sure if you need any clarification to the example workflow I mentioned in https://github.com/etcd-io/etcd/pull/15248#issuecomment-1420125977. If it's clear to you, then I suppose it's pretty easy to intentionally create the scenario.  

To avoid any confusion, please see https://github.com/etcd-io/etcd/pull/15262, please only read the third commit. 
- The first commit is directly coming from your PR (15235, rebased again main this morning my time);
- the second commit is coming from https://github.com/etcd-io/etcd/pull/15237, actually it isn't necessary.  
- I made some change in the third commit to reproduce the duplicated event issue.

Please see my local reproduction.
```
$ go test -run TestLinearizability
panic: duplicate operation in two responses on revision 40
---
first:
{
    "Header": {
        "cluster_id": 238453183653593855,
        "member_id": 14578408409545168728,
        "revision": 40,
        "raft_term": 3
    },
    "Events": [
        {
            "kv": {
                "key": "MQ==",
                "create_revision": 25,
                "mod_revision": 40,
                "version": 4,
                "value": "NDg="
            }
        }
    ],
    "CompactRevision": 0,
    "Canceled": false,
    "Created": false
}
---
second:
{
    "Header": {
        "cluster_id": 238453183653593855,
        "member_id": 14578408409545168728,
        "revision": 40,
        "raft_term": 4
    },
    "Events": [
        {
            "kv": {
                "key": "MQ==",
                "create_revision": 25,
                "mod_revision": 40,
                "version": 4,
                "value": "NDg="
            }
        }
    ],
    "CompactRevision": 0,
    "Canceled": false,
    "Created": false
}

goroutine 77 [running]:
go.etcd.io/etcd/tests/v3/linearizability.collectMemberWatchEvents({0x1e61d10, 0xc0003ba600}, 0xc0001a23f0?, 0xc0000421e0)
	/Users/wachao/go/src/github.com/ahrtr/etcd/tests/linearizability/watch.go:120 +0xb4a
go.etcd.io/etcd/tests/v3/linearizability.collectClusterWatchEvents.func1(0x0, 0x0?)
	/Users/wachao/go/src/github.com/ahrtr/etcd/tests/linearizability/watch.go:52 +0xc5
created by go.etcd.io/etcd/tests/v3/linearizability.collectClusterWatchEvents
	/Users/wachao/go/src/github.com/ahrtr/etcd/tests/linearizability/watch.go:49 +0x4cd
exit status 2
FAIL	go.etcd.io/etcd/tests/v3/linearizability	5.483s
```

client/v3/watch.go
ahrtr: It's a good question. The fragments have already been merged before the merged response being dispatched to the `serveSubstream`. The rough workflow is as below,
1. Firstly the client side merges all fragments if needed. See [watch.go#L644-L648](https://github.com/etcd-io/etcd/blob/586eaccc4d28d4c2af507eaef0051d21a7e4b04c/client/v3/watch.go#L644-L648);
2. dispatch the response. See [watch.go#L651](https://github.com/etcd-io/etcd/blob/586eaccc4d28d4c2af507eaef0051d21a7e4b04c/client/v3/watch.go#L651);
3. The `serveSubstream` forwards the response to the subscriber. See [watch.go#L845](https://github.com/etcd-io/etcd/blob/586eaccc4d28d4c2af507eaef0051d21a7e4b04c/client/v3/watch.go#L845);

logicalhan: Er... this is going to break Kubernetes, which manually does this already.

liggitt: agreed, please don't change the meaning of the data returned to offset by one

lavalamp: Actually this may be correct -- look at line 887, which is the normal case. It adds one.

ahrtr: I agree with @lavalamp . 

@logicalhan @liggitt Could you share the thought why K8s manually do this? I guess it's just because etcd did not fix it?

liggitt: is `wr.Header.Revision == wr.Events[len(wr.Events)-1].Kv.ModRevision` in the normal case?

ahrtr: here
https://github.com/etcd-io/etcd/blob/586eaccc4d28d4c2af507eaef0051d21a7e4b04c/client/v3/watch.go#L886-L888

liggitt: @logicalhan was this the spot you were thinking of where Kubernetes adds one to the revision?

Saving `resp.Header.Revision`:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L209

Taking a revision to watch from: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L105-L109

Starting a watch one revision higher than the requested or last returned revision:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L236

ahrtr: Note this PR is fixing the etcd clientv3. My understanding is that it doesn't impact K8s at all. When etcdserver crashes, the etcd clientv3 will automatically re-send the watch request with the `nextRev`; this process is transparent to K8s. 

From client perspective, K8s remember the last rev, and manually adding 1 is totally client's behavior, and obviously it's also correct.

logicalhan: This is the type of change that seems a bit crazy to me without unit tests that demonstrate what is currently broken and what this fixes.

logicalhan: > @logicalhan was this the spot you were thinking of where Kubernetes adds one to the revision?

Yeah.

ahrtr: > This is the type of change that seems a bit crazy to me without unit tests that demonstrate what is currently broken and what this fixes.

We have test cases in separate PRs.

liggitt: ok, if this doesn't change the revisions returned to clients, and aligns the zero-event progress-notification handling calculation of `nextRev`, I think that makes sense

ahrtr: > ok, if this doesn't change the revisions returned to clients

Right. It doesn't anything being returned to client (`again, etcd clientv3's user.`)

ahrtr:
To give all people more confidence, I also manually reproduced this issue against `main` branch using a simple program below.  

```
package main

import (
	"context"
	"encoding/json"
	"fmt"

	"go.etcd.io/etcd/client/v3"
)

func main() {
	cfg := clientv3.Config{
		Endpoints: []string{"http://127.0.0.1:2379"},
		LogConfig: nil,
	}

	cli, err := clientv3.New(cfg)
	if err != nil {
		panic(err)
	}

	for wr := range cli.Watch(context.Background(), "k1", clientv3.WithProgressNotify()) {
		if wr.Err() != nil {
			fmt.Printf("Error: %v\n", wr.Err())
		} else {
			fmt.Println("------")
			respMsg, err := json.MarshalIndent(wr, "", " ")
			if err != nil {
				fmt.Printf("Failed to marshal response, error: %v\n", err)
			} else {
				fmt.Println(fmt.Sprintf("%s", respMsg))
			}
		}
	}
}
```

The steps are:
1. Start etcd: ` ./etcd --experimental-watch-progress-notify-interval="5s"`;
2. Run the program above: `go run main.go `; You will see periodically progress notification every 5 seconds;
3. Execute `./etcdctl  put k1 v1`. The program started at step 2 received the response.
```
------
{
 "Header": {
  "cluster_id": 14841639068965178418,
  "member_id": 10276657743932975437,
  "revision": 2,
  "raft_term": 3
 },
 "Events": [
  {
   "kv": {
    "key": "azE=",
    "create_revision": 2,
    "mod_revision": 2,
    "version": 1,
    "value": "djE="
   }
  }
 ],
 "CompactRevision": 0,
 "Canceled": false,
 "Created": false
}
```
4. Wait 6 seconds ( please make sure > 5s to receive at least one progress notification), then stop etcd.
5. Start etcd again, then the client ( the program started at step 2) will receive exactly the same response as step 3.

ahrtr:
It's also reproduced on etcd 3.5.7. Is this a big problem to K8s? @liggitt

client/v3/watch.go
lavalamp: I would expect test changes to come in the same PR as the fix -- like did this not break any unit tests? That would be very surprising. (but I'm not familiar with etcd practice)

ahrtr: > I would expect test changes to come in the same PR as the fix -- like did this not break any unit tests? That would be very surprising. (but I'm not familiar with etcd practice)

Right. I am planning to add a test case today.

ahrtr: I was thinking to depend on the [generic linearizablity test case](https://github.com/etcd-io/etcd/pull/15257), but it seems it makes more sense to add a specific test case to reproduce this issue.

lavalamp: That there isn't a unit test exercising this line is probably most of the problem...

ahrtr:
Added an integration test case. Without the fix in this PR, the test case always fails. After applying the fix, the test case always succeeds. 

It's pretty sure it fixes the duplicated event issue .

ahrtr:
cc @mitake @ptabor @spzala as well, PTAL.

serathius (approved):
Looks good! 
With the integration test it's clear that there was a bug here. Thanks for going additional mile. 

Any inclination what is missing in linearizability test that prevented reproducing this? I expect that we didn't have as it's etcd failing scenario that is not as well covered by integration/e2e tests. 

Failure injection is domain of functional/linearizability tests, so I would like to identify why we didn't reproduce it that way.

liggitt:
> It's also reproduced on etcd 3.5.7. Is this a big problem to K8s? @liggitt

umm... I'm trying to think how bad repeating the latest event in a watch stream _could_ be... it's definitely not as bad as replaying non-latest historical events. It could definitely confuse a watcher that was taking action on observing transitions and didn't expect to see the same transition occur two times, but a quick sweep didn't turn up any obvious issues with in-tree watchers in kubernetes (we almost universally work on the latest observed object from a locally maintained cache, and replaying the latest event doesn't break that).

for this scenario to occur, would all of these would have to happen:
1. the API server opts into progress notifications (default behavior)
2. no writes to etcd have occurred since the last object create/update/delete write received by the etcd watcher (so the latest revision is the same as the revision on the last CRUD event received)
3. a progress notification is the last event received by the etcd watcher
4. something would have to trigger the etcd watcher reestablishing the watch (connection loss, etcd server restart, etc)

ahrtr:
> Any inclination what is missing in linearizability test that prevented reproducing this?

The linearizability test is sort of generic test, I don't surprise that it can't reliably reproduce this issue. It's exactly the reason why I implemented a dedicated integration test for this issue. For more details, please refer to https://github.com/etcd-io/etcd/pull/15248#issuecomment-1423225742 and the third commit in https://github.com/etcd-io/etcd/pull/15262

> umm... I'm trying to think how bad repeating the latest event in a watch stream _could_ be... it's definitely not as bad as replaying non-latest historical events

Yes, it only replay the **latest** event.

> for this scenario to occur, would all of these would have to happen:
> 
>     1. the API server opts into progress notifications (default behavior)
> 
>     2. no writes to etcd have occurred since the last object create/update/delete write received by the etcd watcher (so the latest revision is the same as the revision on the last CRUD event received)
> 
>     3. a progress notification is the last event received by the etcd watcher
> 
>     4. something would have to trigger the etcd watcher reestablishing the watch (connection loss, etcd server restart, etc)

Exactly!

ahrtr:
> umm... I'm trying to think how bad repeating the latest event in a watch stream _could_ be... it's definitely not as bad as replaying non-latest historical events. It could definitely confuse a watcher that was taking action on observing transitions and didn't expect to see the same transition occur two times, but a quick sweep didn't turn up any obvious issues with in-tree watchers in kubernetes (we almost universally work on the latest observed object from a locally maintained cache, and replaying the latest event doesn't break that).

Thanks @liggitt for the feedback.

serathius:
FYI Linearizability tests have reproduced the even after the fix was merged!

Run https://github.com/etcd-io/etcd/actions/runs/4142560023/jobs/7163370423

From run actions/checkout
```
/usr/bin/git log -1 --format='%H'
'36fc3cae65cac1b844d2561471dc95b44753618d'
```

From logs
```
2023-02-10T09:53:31.8967095Z     logger.go:130: 2023-02-10T09:53:30.726Z	INFO	Recorded operations	{"count": 690}
2023-02-10T09:53:31.8969114Z     logger.go:130: 2023-02-10T09:53:30.726Z	INFO	Average traffic	{"qps": 278.4656727398674}
2023-02-10T09:53:31.8970286Z     logger.go:130: 2023-02-10T09:53:31.726Z	INFO	killing server...	{"name": "TestLinearizabilityIssue14370-test-0"}
2023-02-10T09:53:31.8971440Z     logger.go:130: 2023-02-10T09:53:31.727Z	INFO	stopping server...	{"name": "TestLinearizabilityIssue14370-test-0"}
2023-02-10T09:53:31.8972771Z     logger.go:130: 2023-02-10T09:53:31.730Z	INFO	stopped server.	{"name": "TestLinearizabilityIssue14370-test-0"}
2023-02-10T09:53:31.8973344Z     watch.go:99: Expect revision to grow by 1, last: 166, mod: 166
2023-02-10T09:53:31.8974457Z     logger.go:130: 2023-02-10T09:53:31.863Z	INFO	Saving visualization	{"path": "/tmp/linearizability/TestLinearizability_Issue14370/history.html"}
2023-02-10T09:53:31.8975158Z     logger.go:130: 2023-02-10T09:53:31.889Z	INFO	closing test cluster...
2023-02-10T09:53:31.8975976Z     logger.go:130: 2023-02-10T09:53:31.889Z	INFO	stopping server...	{"name": "TestLinearizabilityIssue14370-test-0"}
2023-02-10T09:53:31.8976770Z     logger.go:130: 2023-02-10T09:53:31.889Z	INFO	closing server...	{"name": "TestLinearizabilityIssue14370-test-0"}
2023-02-10T09:53:31.8977579Z     logger.go:130: 2023-02-10T09:53:31.889Z	INFO	stopping server...	{"name": "TestLinearizabilityIssue14370-test-0"}
2023-02-10T09:53:31.8978441Z     logger.go:130: 2023-02-10T09:53:31.889Z	INFO	removing directory	{"data-dir": "/tmp/TestLinearizabilityIssue143702810931773/001"}
2023-02-10T09:53:31.8979302Z     logger.go:130: 2023-02-10T09:53:31.892Z	INFO	closed test cluster.
2023-02-10T09:53:31.8979814Z --- FAIL: TestLinearizability (25.73s)
2023-02-10T09:53:31.9976172Z [1m[0;31mFAIL: (code:1):
2023-02-10T09:53:31.9977152Z     --- PASS: TestLinearizability/ClusterOfSize1/LowTraffic (3.95s)
2023-02-10T09:53:31.9978328Z     --- PASS: TestLinearizability/ClusterOfSize3/LowTraffic (6.70s)
2023-02-10T09:53:31.9979892Z     --- PASS: TestLinearizability/ClusterOfSize1/HighTraffic (4.49s)
2023-02-10T09:53:31.9980995Z     --- PASS: TestLinearizability/ClusterOfSize3/HighTraffic (6.29s)
2023-02-10T09:53:31.9981901Z     --- FAIL: TestLinearizability/Issue14370 (4.29s)
```

ptabor:
@serathius Do you mean they reproduced the bigger issue of https://github.com/etcd-io/etcd/issues/15271  or something else ?

ahrtr:
> 2023-02-10T09:53:31.8973344Z     watch.go:99: Expect revision to grow by 1, last: 166, mod: 166

I saw similar error message with the help of https://github.com/etcd-io/etcd/pull/15272. Do not get time to dig into it so far. It isn't confirmed that whether it's a test issue or production issue for now. Will take a deep dive later.

ahrtr:
It's my top priority next week.

serathius:
> @serathius Do you mean they reproduced the bigger issue of https://github.com/etcd-io/etcd/issues/15271 or something else ?

It's totally a different beast as they have different failpoint triggers.

ahrtr:
Please all people in this loop do not panic, it's confirmed that it's just a test issue.  FYI. https://github.com/etcd-io/etcd/pull/15288

PR #15250, opened 2023-02-06
tests: Validate watch responses in linearizability tests

serathius:
cc @ahrtr @ptabor


tests/linearizability/watch.go
```diff
@@ -47,53 +47,90 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+	for _, memberResponses := range responses {
+		validateMemberWatchResponses(t, memberResponses)
+	}
+}
+
+func validateMemberWatchResponses(t *testing.T, responses []watchResponse) {
+	var lastRevision int64 = 1
+	for _, resp := range responses {
+		if resp.Header.Revision < lastRevision {
+			t.Errorf("Revision should never decrease")
```
ahrtr, line 87: This minor comment applies to the following couple of places (`t.Errorf`)
```suggestion
			t.Error("Revision should never decrease")
```

ahrtr (approved):
LGTM with a minor comment.

I assume you will add the duplicated event detection in a separate PR.

serathius:
> I assume you will add the duplicated event detection in a separate PR.

I expect that it should already detect duplicate events as I added validation for watch event mod revision being incremental.

tests/linearizability/watch.go
```diff
@@ -47,53 +47,90 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+		}
+		if resp.Header.Revision == lastRevision && len(resp.Events) != 0 {
+			t.Errorf("Got two non-empty responses about same revision")
+		}
+		for _, event := range resp.Events {
+			if event.Kv.ModRevision != lastRevision+1 {
```
ahrtr, line 93: This may not be correct, because all events in one TXN have the same revision. The reason why we do not see issue is we do not support multi-operations TXN yet.

serathius: Thanks for pointing this out.

Issue #15058, opened 2022-12-31
Add a Code based member for cancel reason in WatchResponse


mitake:
### What would you like to be added?

Currently `WatchResponse` message has a string type member [CancelReason](https://github.com/etcd-io/etcd/blob/main/api/etcdserverpb/rpc.pb.go#L2214). The member is used for storing an error which happened during watch.
 It's better to add a new [Code](https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/codes/codes.go#L29) based member for representing an error in the watch API.

### Why is this needed?

Using the existing member for error handling code isn't good because of fragility and hard to track behavior of the etcd wire protocol.

See also:
- https://github.com/kubernetes/kubernetes/pull/114403
- https://github.com/etcd-io/etcd/pull/14995
- https://github.com/etcd-io/etcd/discussions/14992

lavacat:
@mitake seams like you've been working on this code before, are you planning to contribute this change if approved?

How are you planning to handle different [values](https://github.com/etcd-io/etcd/blob/9cb57648593cdd34a4f5789a31da7dd390d46251/server/etcdserver/api/v3rpc/watch.go#L276) of `CancelReason`? I don't think they map to grpc codes directly.

ahrtr:
The direction of adding a [Code](https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/codes/codes.go#L29) (e.g. `ReasonCode`) looks good to me. Please evaluate the impact and provide a design firstly. 

Since it changes the protocol between client and server, so we need K8s folks to review the changes as well. cc @ptabor @serathius @liggitt @dims @lavalamp @deads2k as well.

liggitt:
a new optional (since old servers won't send it) code field the client reacts to if present seems like a minimal and safe approach

ahrtr:
> a new optional (since old servers won't send it) code field the client reacts to if present seems like a minimal and safe approach

Agree. It would be something like below,

```
$ git diff
diff --git a/api/etcdserverpb/rpc.proto b/api/etcdserverpb/rpc.proto
index 9cdc0b37f..7b37b1556 100644
--- a/api/etcdserverpb/rpc.proto
+++ b/api/etcdserverpb/rpc.proto
@@ -834,6 +834,8 @@ message WatchResponse {
   // framgment is true if large watch response was split over multiple responses.
[...]

PR #15253, opened 2023-02-07
etcdserverpb: add "retry" flag to WatchResponse

lavacat:
Problem: client relies on string value of CancelReason to determine if watcher should be retried. This is too fragile. See https://github.com/etcd-io/etcd/discussions/14992

Solution: add explicit `retry` flag

fixes: https://github.com/etcd-io/etcd/issues/15058


api/etcdserverpb/rpc.proto
```diff
@@ -834,6 +834,9 @@ message WatchResponse {
   // framgment is true if large watch response was split over multiple responses.
   bool fragment = 7 [(versionpb.etcd_version_field)="3.4"];
 
+  // retry is true if watcher should be retried on cancel.
+  bool retry = 8 [(versionpb.etcd_version_field)="3.5"];
```
lavacat, line 838: I'll backport to 3.5, so setting version 3.5 here

lavacat:
This code is covered by TestV3AuthWatchAndTokenExpire integration test.
Looked at adding unit test to v3rpc/watch_test.go, but it requires setting up serverWatchStream.recvLoop.

api/etcdserverpb/rpc.proto
serathius: No backports of fields!

serathius:
This is a small change but with a big impact on important API semantics. Please start with creating issue discussing problem you want to solve and why solution you propose is worth adopting.

serathius (changes requested):
Would like to first discuss different options before picking a solution.

lavacat:
@serathius do you mind continuing discussion in https://github.com/etcd-io/etcd/issues/15058? Here is my [comment](https://github.com/etcd-io/etcd/issues/15058#issuecomment-1399342701) explainig `retry`. Other folks commented about this, so don't want to loose context.

mitake (commented):
After having consensus about the approach, I think we should update the changelog for describing the new field (or changes for other approach): https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md

Issue #14665, opened 2022-10-31
Proxy: Add watchers for keys in cache


biosvs:
### What would you like to be added?

Run watcher inside proxy for every key that is added to the cache.

(Maybe you have some better suggestions, this issue is created for architecture discussion.)

### Why is this needed?

I have noticed, that proxy cache doesn't have any TTL for keys or any other options to update values other than by making direct request to the proxy node. Either to get key (with "linearizable" consistency level), or to delete, or to compact.

If key was changed by request through another proxy node or with direct request to the cluster node, stale value would stuck forever.

That is mean that if more than one proxy node in use (or direct requests to cluster nodes are possible), cache could not be used at all (all read requests should be linearizable). The only usecase for proxy in this case is watchers coalescing.

tjungblu:
> If key was changed by request through another proxy node or with direct request to the cluster node, stale value would stuck forever.

Just to clarify. Your setup looks like this:

```mermaid
graph TD;
    YourCode-->|grpc|gRPCProxy;
    gRPCProxy-->|http|etcd;

```

then "YourCode" writes some key/value, let's say `a=1` and another actor removes that key from etcd directly?

biosvs:
I have client and several proxy nodes:

```mermaid
graph TD;
    YourCode-->|grpc|gRPCProxy1;
    YourCode-->|grpc|gRPCProxy2;
    gRPCProxy1-->|http|etcd;
    gRPCProxy2-->|http|etcd;
```

Let's imagine I have key `k1` with value `v1` in etcd.
I read `k1` from Proxy1, then read `k1` from Proxy2. Both proxies have "k1 -> v1" in cache now.
Then I overwrite `k1` with value `v2` through Proxy1.
Now Proxy1 has actual "k1 -> v2", but Proxy2 has stale "k1 -> v1".

tjungblu:
Thanks for the clarification, is the issue that you read stale values from `Proxy2`? Or the memory usage from the stale entries?

As far as I understand the proxy code it will only return serializable results from the cache and save the last response as such (basically a write-through caching):
https://github.com/etcd-io/etcd/blob/main/server/proxy/grpcproxy/kv.go#L40-L68

biosvs:
So, yes, I think that stale reads from Proxy2 is the issue.

You're right, stale reads are possible only with "serializable" mode. And it's true for direct connection toward etcd cluster.
[...]

PR #15254, opened 2023-02-07
etcdproxy: Added flag for enabling/disabling cache.

biosvs:
Originally discussed in [issue 14665](https://github.com/etcd-io/etcd/issues/14665#issuecomment-1419836385)

Currently cache may have stale data, and the only option for clients to resolve it is to use linearizable reads, which can be not easy for big codebase and make it impossible to transparently replace etcd cluster with grpc proxy.
New flag adds second option to solve stale reads problem (serializable reads still has less garantuess, but at least they are eventually consistent).

cc @ahrtr @Hexta

### DIFF
--- a/server/etcdmain/grpc_proxy.go
+++ b/server/etcdmain/grpc_proxy.go
@@ -66,6 +66,7 @@ var (
 	grpcProxyDNSClusterServiceName     string
 	grpcProxyInsecureDiscovery         bool
 	grpcProxyDataDir                   string
+	grpcProxyEnableCache               bool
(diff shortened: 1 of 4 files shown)


biosvs:
Regarding failing check (https://github.com/etcd-io/etcd/actions/runs/4112900236/jobs/7099337814).

Maybe I don't understand something about how integration tests are run. But locally those two particular tests are passed. Whole list of tests can't passed due to timeout.

ahrtr:
cc @ptabor @serathius @spzala

PR #15257, opened 2023-02-07
tests: Enable progress notify in linearizability tests

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -103,6 +103,7 @@ func TestLinearizability(t *testing.T) {
 				e2e.WithSnapshotCount(100),
 				e2e.WithGoFailEnabled(true),
 				e2e.WithCompactionBatchLimit(100), // required for compactBeforeCommitBatch and compactAfterCommitBatch failpoints
+				e2e.WithWatchProcessNotifyInterval(100*time.Millisecond),
 			),
 		})
 		scenarios = append(scenarios, scenario{
@@ -114,6 +115,7 @@ func TestLinearizability(t *testing.T) {
 				e2e.WithPeerProxy(true),
 				e2e.WithGoFailEnabled(true),
 				e2e.WithCompactionBatchLimit(100), // required for compactBeforeCommitBatch and compactAfterCommitBatch failpoints
+				e2e.WithWatchProcessNotifyInterval(100*time.Millisecond),
 			),
 		})
 	}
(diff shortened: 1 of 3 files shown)


tests/linearizability/linearizability_test.go
```diff
@@ -177,7 +179,7 @@ func TestLinearizability(t *testing.T) {
 				waitBetweenTriggers: waitBetweenFailpointTriggers,
 			}, *scenario.traffic)
 			forcestopCluster(clus)
-			validateWatchResponses(t, watchResponses)
+			validateWatchResponses(t, watchResponses, clus.Cfg.WatchProcessNotifyInterval != 0)
```
fuweid, line 182: Not sure I understand it correctly.
.
By default, the interval is 10 minutes. Currently, the test case won't hit the limit.
So `x != 0` can be used as disabled progress-notify? is that correct?
If so, is it possible to add some comment for this?

serathius: > So x != 0 can be used as disabled progress-notify?

No, in e2e cluster config 0 is empty value which means, don't set. In that case etcd will use the default. As you mentioned the default interval is 10 minutes which is too much for test takes around 5 seconds.

fuweid: Thanks for the comment!

fuweid (approved):
LGTM

serathius:
As per @logicalhan suggestion I have added validation that there at least one progress notify that returns same revision as previous watch event.

tests/linearizability/linearizability_test.go
```diff
@@ -177,7 +179,8 @@ func TestLinearizability(t *testing.T) {
 				waitBetweenTriggers: waitBetweenFailpointTriggers,
 			}, *scenario.traffic)
 			forcestopCluster(clus)
-			validateWatchResponses(t, watchResponses)
+			watchProgressNotifyEnabled := clus.Cfg.WatchProcessNotifyInterval != 0
```
ahrtr, line 182: Timed progress notification is always enabled. Users just can set a different interval. So the name `watchProgressNotifyEnabled` is a little misleading.

ahrtr (approved):
Overall looks good to me with a minor comment, which can be resolved in a separate PR. 

Merging this PR for now, and I will add a new test case on top of this one.

tests/linearizability/linearizability_test.go
logicalhan: It's not though, he's comparing the interval to a zero value, and if it is a zero value, timed progress notifications are effectively disabled.

ahrtr: I was saying etcdserver always enables the timed progress notification. Are you talking about the same thing?

ahrtr: Note: etcdserver always triggers the periodical progress notification for each established watch stream, and it sends progress notification for each watch substream if [ProgressNotify](https://github.com/etcd-io/etcd/blob/39d98522d6b7e3ac4463857d9e7db78eb76131b3/api/etcdserverpb/rpc.pb.go#L1995) is `true`.

serathius: @ahrtr is right, will rephrase it.

PR #15260, opened 2023-02-08
clientv3: deprecate RequestProgress in Watcher interface

ahrtr:
See discussion in https://github.com/etcd-io/etcd/pull/15237

### DIFF
--- a/client/v3/watch.go
+++ b/client/v3/watch.go
@@ -86,7 +86,19 @@ type Watcher interface {
 	// (see https://github.com/etcd-io/etcd/issues/8980)
 	Watch(ctx context.Context, key string, opts ...OpOption) WatchChan
 
-	// RequestProgress requests a progress notify response be sent in all watch channels.
+	// RequestProgress requests a progress notify response be sent in all
+	// watch channels.
+	//
+	// Deprecated: There is a known issue on this method. It may cause the
+	// client side to receive a watch progress notification with a revision
+	// equal to or higher than the following watch responses. Please see
+	// the discussion in https://github.com/etcd-io/etcd/issues/15220.
+	//
+	// It's planned to be decommissioned in 3.7. Please use the periodic
+	// progress updates. For example,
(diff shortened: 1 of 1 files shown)


liggitt:
cc @wojtek-t @jpbetz for the intent behind using this in Kubernetes

ahrtr:
See https://github.com/etcd-io/etcd/pull/15237#issuecomment-1423298262

jpbetz:
Why would we deprecated just this feature if the underlying problem also impacts `[clientv3.WithProgressNotify()](https://github.com/etcd-io/etcd/blob/b10ebff3bb0c337b3b153a39474e33fa9698790a/client/v3/op.go#L477)` (which is in production use in k8s)?  If we fix the underlying problem, would there be any reason to deprecate this API?

jpbetz:
Motivation for this feature is https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md, which we never got around to implementing, but which I still believe may be a valuable optimization. This became less urgent when https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/956-watch-bookmark/README.md was introduced, because "bookmarks" resolved the most pressing problems in keeping caches from becoming too stale, but  "consistent reads from cache" still has scale properties that are untapped.

ahrtr:
Thanks @jpbetz for the feedback. Will do more archaeological work when I get free cycle.  Two quick comments as below for now, 

1. We are heading to the direction of fixing the issue instead of just deprecating the [manual progress notification](https://github.com/etcd-io/etcd/blob/9bae01099413b40779000168dcef4d71501d1ac4/client/v3/watch.go#L90). We will keep discussing under the PR https://github.com/etcd-io/etcd/pull/15237
2. The [clientv3.WithProgressNotify()](https://github.com/etcd-io/etcd/blob/b10ebff3bb0c337b3b153a39474e33fa9698790a/client/v3/op.go#L477) is used for the timed progress notification, it isn't impacted at all.

Issue #15243, opened 2023-02-04
`MemberList` doesn't work after adding a new member to one-node cluster


justinsb:
### What happened?

etcdadm automates bringing up a cluster, and has a test where it brings up a cluster node-by-node.  That test regressed when we updated the client from 3.4 to 3.5, seemingly a reintroduction of #9949

### What did you expect to happen?

No regression

### How can we reproduce it (as minimally and precisely as possible)?

https://github.com/kubernetes-sigs/etcdadm/pull/364 demonstrates the problem.

The failing test is relatively simple: https://github.com/kubernetes-sigs/etcdadm/blob/master/test/e2e/cluster_phases.sh

A failing run can be seen [here](https://github.com/kubernetes-sigs/etcdadm/actions/runs/4092492086/jobs/7057270193)

```
+ docker exec etcdadm-1 /etcdadm/etcdadm join phase membership https://172.17.0.2:2379/ --name etcdadm-1
time="2023-02-04T16:42:58Z" level=info msg="[membership] Checking if this member was added"
time="2023-02-04T16:42:58Z" level=info msg="[membership] Member was not added"
time="2023-02-04T16:42:58Z" level=info msg="Removing existing data dir \"/var/lib/etcd\""
time="2023-02-04T16:42:58Z" level=info msg="[membership] Adding member"
time="2023-02-04T16:42:58Z" level=info msg="[membership] Checking if member was started"
time="2023-02-04T16:42:58Z" level=info msg="[membership] Member was not started"
time="2023-02-04T16:42:58Z" level=info msg="[membership] Removing existing data dir \"/var/lib/etcd\""
+ docker exec etcdadm-1 /etcdadm/etcdadm join phase install https://172.17.0.2:2379/ --name etcdadm-1
2023/02/04 16:42:58 [install] Artifact not found in cache. Trying to fetch from upstream: https://github.com/coreos/etcd/releases/download
time="2023-02-04T16:42:58Z" level=info msg="[install] Downloading & installing etcd https://github.com/coreos/etcd/releases/download from 3.5.7 to /var/cache/etcdadm/etcd/v3.5.7\n"
time="2023-02-04T16:42:58Z" level=info msg="[install] downloading etcd from https://github.com/coreos/etcd/releases/download/v3.5.7/etcd-v3.5.7-linux-amd64.tar.gz to /var/cache/etcdadm/etcd/v3.5.7/etcd-v3.5.7-linux-amd64.tar.gz\n"
...
[...]

PR #15261, opened 2023-02-08
clientv3: support serializable `MemberList` operation

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/15243

### DIFF
--- a/CHANGELOG/CHANGELOG-3.6.md
+++ b/CHANGELOG/CHANGELOG-3.6.md
@@ -32,12 +32,17 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 - [Always print the raft_term in decimal](https://github.com/etcd-io/etcd/pull/13711) when displaying member list in json.
 - [Add one more field `storageVersion`](https://github.com/etcd-io/etcd/pull/13773) into the response of command `etcdctl endpoint status`.
 - Add [`--max-txn-ops`](https://github.com/etcd-io/etcd/pull/14340) flag to make-mirror command.
+- Add [`--consistency`](https://github.com/etcd-io/etcd/pull/15261) flag to member list command.
 - Display [field `hash_revision`](https://github.com/etcd-io/etcd/pull/14812) for `etcdctl endpoint hash` command.
 
 ### etcdutl v3
 
 - Add command to generate [shell completion](https://github.com/etcd-io/etcd/pull/13142).
 - Add `migrate` command for downgrading/upgrading etcd data dir files.
+
+### Package `clientv3`
+
+- [Support serializable `MemberList` operation](https://github.com/etcd-io/etcd/pull/15261).
 
 ### Package `server`
 
--- a/client/v3/client_test.go
+++ b/client/v3/client_test.go
@@ -426,7 +426,7 @@ type mockCluster struct {
 	members []*etcdserverpb.Member
 }
 
-func (mc *mockCluster) MemberList(ctx context.Context) (*MemberListResponse, error) {
+func (mc *mockCluster) MemberList(ctx context.Context, opts ...OpOption) (*MemberListResponse, error) {
 	return &MemberListResponse{Members: mc.members}, nil
 }
 
--- a/client/v3/cluster.go
+++ b/client/v3/cluster.go
@@ -34,7 +34,7 @@ type (
 
 type Cluster interface {
 	// MemberList lists the current cluster membership.
-	MemberList(ctx context.Context) (*MemberListResponse, error)
+	MemberList(ctx context.Context, opts ...OpOption) (*MemberListResponse, error)
 
 	// MemberAdd adds a new member into the cluster.
 	MemberAdd(ctx context.Context, peerAddrs []string) (*MemberAddResponse, error)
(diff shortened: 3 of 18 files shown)


serathius:
Let's discuss other options to solve https://github.com/etcd-io/etcd/issues/15243 before we start implementation.

ahrtr:
Please see my comment in https://github.com/etcd-io/etcd/issues/15243#issuecomment-1423257824.

ahrtr:
This PR should be a simple & straightforward change, and @serathius seems be also OK with this change per his comment https://github.com/etcd-io/etcd/issues/15243#issuecomment-1423788295. 

Do you have any other comments or objections? @mitake @ptabor @spzala

ahrtr:
Just added changelog items and updated etcdctl's README.

mitake:
@ahrtr I think it's fine to add the flag to etcdctl. This PR exposes the flag of the API as from the CLI, it should be safe. I have no objections.
Membership change especially for one node cluster is tricky, let me follow https://github.com/etcd-io/etcd/issues/15243 for understanding essential problems.

ahrtr:
> @ahrtr I think it's fine to add the flag to etcdctl. This PR exposes the flag of the API as from the CLI, it should be safe. I have no objections. Membership change especially for one node cluster is tricky, let me follow #15243 for understanding essential problems.

Thanks @mitake for the feedback. Please read https://github.com/etcd-io/etcd/issues/15243#issuecomment-1418305327 for quick reference.

spzala:
@ahrtr I also think this approach is fine. In the past, I think, before we changed to linearizable to always true user could get stale data mistakenly but now with providing options for l or s, users should be aware of the output somthing similar to get operation behavior. One thought here is to add a brief doc on when the user may want to use serializable for cases like the one described in the issue related to this PR. Thanks!

ahrtr:
> One thought here is to add a brief doc on when the user may want to use serializable for cases like the one described in the issue related to this PR

Makes sense to me. thx @spzala 

Just updated the comment of `clientv3.WithSerializable` to clarify the warning (of reading stale data) and use cases. Please see the second commit: https://github.com/etcd-io/etcd/pull/15261/commits/9aced4c996a5422ff9b70b7dd7b073639c241dd8

spzala (commented):
Thanks so much for the quick doc update @ahrtr !! While we wait on other reviewer feedback, I think also a good idea to update some of this specially the second paragraph content in the etcdctl doc updates for the memberlist?

ahrtr:
> I think also a good idea to update some of this specially the second paragraph content in the etcdctl doc updates for the memberlist?

Thanks @spzala for the feedback. Do you mean to update the [memberlist doc in etcdctl/README.md](https://github.com/ahrtr/etcd/blob/memberlist_20230208/etcdctl/README.md#member-list)? I intentionally keep it consistent with [the doc for Get](https://github.com/ahrtr/etcd/blob/memberlist_20230208/etcdctl/README.md#get-options-key-range_end).

The new added flag `--consistency` for [MemberList](https://github.com/etcd-io/etcd/blob/822eeaf9f2d9e327df5f4cab8ff075c9b5727320/client/v3/cluster.go#L37) has exactly the same behavior as the existing [Get](https://github.com/etcd-io/etcd/blob/822eeaf9f2d9e327df5f4cab8ff075c9b5727320/client/v3/kv.go#L49) operation. I think it makes more sense to update (already done) the [clientv3.WithSerializable()](https://github.com/ahrtr/etcd/blob/9aced4c996a5422ff9b70b7dd7b073639c241dd8/client/v3/op.go#L421-L431) which is shared by both `MemberList` and `Get`. 

If we update the doc for memberlist, then we need to update it for the Get as well. So we either keep it as it's or update both. WDYT?

spzala:
> > I think also a good idea to update some of this specially the second paragraph content in the etcdctl doc updates for the memberlist?
> 
> Thanks @spzala for the feedback. Do you mean to update the [memberlist doc in etcdctl/README.md](https://github.com/ahrtr/etcd/blob/memberlist_20230208/etcdctl/README.md#member-list)? I intentionally keep it consistent with [the doc for Get](https://github.com/ahrtr/etcd/blob/memberlist_20230208/etcdctl/README.md#get-options-key-range_end).
> 
> The new added flag `--consistency` for [MemberList](https://github.com/etcd-io/etcd/blob/822eeaf9f2d9e327df5f4cab8ff075c9b5727320/client/v3/cluster.go#L37) has exactly the same behavior as the existing [Get](https://github.com/etcd-io/etcd/blob/822eeaf9f2d9e327df5f4cab8ff075c9b5727320/client/v3/kv.go#L49) operation. I think it makes more sense to update (already done) the [clientv3.WithSerializable()](https://github.com/ahrtr/etcd/blob/9aced4c996a5422ff9b70b7dd7b073639c241dd8/client/v3/op.go#L421-L431) which is shared by both `MemberList` and `Get`.
> 
> If we update the doc for memberlist, then we need to update it for the Get as well. So we either keep it as it's or update both. WDYT?

@ahrtr thank you and yes, what you did to keep consistent with Get in the README is good. But, yes, I was referring to the README. The reason I was thinking to update README (and as you pointed out for MemberList and we should add similar doc for Get as well) is because end users may not refer to the Go code ([clientv3.WithSerializable()](https://github.com/ahrtr/etcd/blob/9aced4c996a5422ff9b70b7dd7b073639c241dd8/client/v3/op.go#L421-L431)) always and providing them heads up in README can be handy. WDYT :) Thanks!

ahrtr:
Done & thx @spzala

spzala (approved):
@ahrtr lgtm, I think it's a good quick fix. Thanks for working on it.

serathius (approved):

Issue #15185, opened 2023-01-26
Move installation scripts out of release info


serathius:
### What would you like to be added?

Installation script should be moved into either etcd.io or etcd repo. Options:
* Update https://etcd.io/docs/v3.5/install/ and link to it from release
* Add script to repo that can be executed directly from . Execution by calling 
    ```
    wget https://raw.githubusercontent.com/etcd-io/etcd/v3.5.7/projects/scripts/install.sh
    chmod +x ./install.sh
    ./install.sh
    ```
    or more extreme
    ```
    bash < <(curl -s -S -L https://raw.githubusercontent.com/etcd-io/etcd/v3.5.7/projects/scripts/install.sh)
    ```

Prefer first option as second doesn't allow user good chance to read what script is doing.

### Why is this needed?

From time to time we need to make a fix in release script and as it's separate for each release it's very toilsome. During release maintainer copies it from last release and needs to manually update version which is error prone as there are a lot of references to version. If change is script needs to be done it needs to be done by maintainers as release info is not stored in repository. 

All those things result in install script being not very maintained and low quality.

pchan:
Created a PR (https://github.com/etcd-io/etcd/pull/15267) that adds the installation script. Please review, I will update the website  [v3.6 docs](https://github.com/etcd-io/website/tree/main/content/en/docs/v3.6)  ? post review.

serathius:
Would be good to check what other project do in this regard.

pchan:
> Would be good to check what other project do in this regard.

Analyzed 2 examples 

1. Build from source, as in, use "make" to build and then install ? (like ([runc](https://github.com/containerd/containerd/blob/main/script/setup/install-runc). Currently we install to a temp directory.
2.  Other option is to install with popular package managers (e.g. [kubeadm binaries](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/))

Issue #15030, opened 2022-12-20
etcd 3.5 releases should call out etcdutl binary is released in installation script


chaochn47:
### What happened?

I cannot find `etcdutl` from https://github.com/etcd-io/etcd/releases/tag/v3.5.6 in the installation script. 

Compared to https://github.com/etcd-io/etcd/releases/tag/v3.4.22. 

`etcdctl` snapshot restore was moved to `etcdutl` in 3.5. 

### What did you expect to happen?

I expect `etcdutl` is released in 3.5. 

### How can we reproduce it (as minimally and precisely as possible)?

It is packaged in `etcd-v3.5.6-linux-amd64.tar.gz` but not called out in the installation script. 

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands blow, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

chaochn47:
Closing as `etcdctl snapshot restore` only has a deprecation notice. It's still usable.

serathius:
Still this is a issue with v3.5 release.

ahrtr:
Probably I am missing something. The `etcdutl` binary has already been included in the released package, e.g. 
`etcd-v3.5.6-linux-amd64.tar.gz `. @chaochn47 please clarify this issue.

```
$ ls -lrt
total 112552
-rwxr-xr-x@ 1 wachao  staff  15970304 Nov 21 22:55 etcdutl
-rwxr-xr-x@ 1 wachao  staff  17891328 Nov 21 22:55 etcdctl
-rwxr-xr-x@ 1 wachao  staff  23691264 Nov 21 22:55 etcd
-rw-r--r--@ 1 wachao  staff      7896 Nov 21 22:55 READMEv2-etcdctl.md
-rw-r--r--@ 1 wachao  staff      9394 Nov 21 22:55 README.md
-rw-r--r--@ 1 wachao  staff      7359 Nov 21 22:55 README-etcdutl.md
-rw-r--r--@ 1 wachao  staff     42066 Nov 21 22:55 README-etcdctl.md
[...]

PR #15267, opened 2023-02-09
scripts: Add an install script that can be referred to during release.

pchan:
Currently installation instructions are part of release notes. This PR attempts to separate them. 
Fixes: https://github.com/etcd-io/etcd/issues/15185
Also addresses the issue of etcdutl not being called out during installation (see https://github.com/etcd-io/etcd/issues/15030)

cc: @serathius @ahrtr


serathius:
We haven't still agreed on which option to choose in https://github.com/etcd-io/etcd/issues/15185. 
Would like to continue the discussion on the issue so pick the best solution.

ahrtr:
FYI. https://github.com/etcd-io/etcd/issues/15382#issuecomment-1455372923

PR #15268, opened 2023-02-09
tests: trigger raftBeforeFollowerSend failpoint in linearizability test

lavacat:
raftBeforeFollowerSend can only be triggered on Follower and won't work on 1 node cluster. Had to split RandomFailpoint into RandomOneNodeClusterFailpoint and RandomMultiNodeClusterFailpoint


lavacat:
This is the last failpoint from https://github.com/etcd-io/etcd/issues/14726 cc @serathius

serathius:
This is good enough as temporary workaround, however would like to avoid having to manually maintain per cluster config list of  failpoints. Rather `failpoint.Available` should be made aware of cluster config so it will return false when it cannot be used on a specific cluster config.

PR #15281, opened 2023-02-10
Show darwin as supported in fileutil.Preallocate docstring

MarkintoshZ:
`fileutil.Preallocate` function is supported on Darwin yet it is not documented. This fix the issue by updating the docstring.

### DIFF
--- a/client/pkg/fileutil/preallocate.go
+++ b/client/pkg/fileutil/preallocate.go
@@ -19,9 +19,9 @@ import (
 	"os"
 )
 
-// Preallocate tries to allocate the space for given
-// file. This operation is only supported on linux by a
-// few filesystems (btrfs, ext4, etc.).
+// Preallocate tries to allocate the space for given file. This 
+// operation is only supported on darwin and linux by a few 
+// filesystems (APFS, btrfs, ext4, etc.).
 // If the operation is unsupported, no error will be returned.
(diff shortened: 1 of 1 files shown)


jmhbnz (commented):
Hey @MarkintoshZ thanks for raising this.  At face value it looks ok but I'm not too familiar with this function. To help us get this merged can you please give a bit more context on how you tested this etc to verify it is working correctly under darwin?

MarkintoshZ:
Hi @jmhbnz, I only change the docstring of that function because it is not up to date. There is no change in the actually code of the function so the behavior should be exactly the same. As a result, I believe testing is not necessary.

The function `Preallocate` calls different versions of `preallocateFixed` depending on the system (Darwin, Unix, or Unsupported). Since there is a Darwin specific implementation of `preallocateFixed` in `preallocate_darwin.go`, the docstring of `Preallocate` is incorrect by saying "This operation is only supported on linux..."

Hope this helps!

chaochn47 (approved):

jmhbnz (approved):
Thanks for the clarification @MarkintoshZ, that makes sense. We have the [preallocate_darwin.go](https://github.com/etcd-io/etcd/blob/main/client/pkg/fileutil/preallocate_darwin.go) implementation already so this looks ok.

cc @ahrtr , @ptabor

PR #15282, opened 2023-02-11
test: Report watch histories

serathius:
cc @ahrtr @ptabor


ahrtr:
One PR is supposed to do only one thing, the secondary choice is one commit doing one thing. Please consider to breakdown the refactoring and persisting watch history into two separate PR (**preferred**) or two commits, so as to make it easier to review.

tests/linearizability/watch.go
```diff
@@ -156,3 +158,37 @@ type watchEvent struct {
...
+	}
+	defer file.Close()
+	encoder := json.NewEncoder(file)
+	for _, event := range events {
+		err := encoder.Encode(event)
+		if err != nil {
+			t.Errorf("Failed to encode response: %v", err)
+		}
+	}
+}
```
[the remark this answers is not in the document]
serathius: Yes, but would not want to rewrite this to generics.

serathius:
Split the PR into multiple commits

serathius:
cc @ptabor

serathius:
cc @ptabor

tests/linearizability/linearizability_test.go
```diff
@@ -449,31 +494,28 @@ func persistOperationHistory(t *testing.T, lg *zap.Logger, path string, operatio
...
+	lg.Info("Saving member data dir", zap.String("member", member.Config().Name), zap.String("path", path))
+	err := os.Rename(member.Config().DataDirPath, path)
+	if err != nil {
+		t.Fatal(err)
 	}
 }
 
-func testResultsDirectory(t *testing.T) (string, error) {
+func testResultsDirectory(t *testing.T) string {
 	path, err := filepath.Abs(filepath.Join(resultsDirectory, strings.ReplaceAll(t.Name(), "/", "_")))
```
ptabor, line 506: FYI: 
There was discussion of technique to preserve the t.TempDir file in case of 'catastrophic' test failure:
https://github.com/golang/go/issues/35998#issuecomment-562413587

But I assume you want to preserve the results in case of successful test results as well. 
On flip side - implicit doing it, might lead to getting out of disk space.

tests/linearizability/linearizability_test.go
```diff
@@ -383,58 +430,56 @@ type trafficConfig struct {
...
-		length := len(other[i])
+func validateEventsMatch(t *testing.T, histories [][]watchEvent) {
+	longestHistory := longestHistory(histories)
+	for i := 0; i < len(histories); i++ {
+		length := len(histories[i])
 		// We compare prefix of watch events, as we are not guaranteed to collect all events from each node.
-		if diff := cmp.Diff(longestHistory[:length], other[i][:length], cmpopts.IgnoreFields(watchEvent{}, "Time")); diff != "" {
-			t.Errorf("Events in watches do not match, %s", diff)
+		if diff := cmp.Diff(longestHistory[:length], histories[i][:length], cmpopts.IgnoreFields(watchEvent{}, "Time")); diff != "" {
+			t.Error("Events in watches do not match")
```
ptabor, line 447: Why did you get rid of prininting the diff ?

ptabor (approved):

tests/linearizability/linearizability_test.go
serathius: Problem I found is that this can spam whole console with thousands and thousands of lines of logs. This can obscure some other errors logs.
With reports written to file, it's now possible to do the diff yourself after running the test. 

Will add a recommendation to diff the files manually.

tests/linearizability/linearizability_test.go
serathius: True, however interrupting tests (ctrl+c) will give the same effect. Difference here that those files are shared between multiple runs so they should not grow in size over time.

Issue #15271, opened 2023-02-09
Watch response traveling back in time when reconnecting member downloads snapshot from the leader


serathius:
### What happened?

Recently added failpoints started showing issues rated to etcd recovery from leaders snapshot. https://github.com/etcd-io/etcd/pull/15104#issuecomment-1416371288

In report uploaded by @lavacat we see revision jumping by 10 which resembles previous data inconsistencies where during restore, member applied same entries multiple times.
![image](https://user-images.githubusercontent.com/5873433/217946603-61ab0290-1403-4692-a323-05f76b6842f1.png)

I was not able to reproduce issue with linearizability model, however In my case recently introduced watch verification is triggered.
```
    watch.go:104: Expect response revision equal last event mod revision
    watch.go:99: Expect revision to grow by 1, last: 3924, mod: 2171
    watch.go:104: Expect response revision equal last event mod revision
    watch.go:99: Expect revision to grow by 1, last: 3979, mod: 3171
```

### What did you expect to happen?

As https://github.com/etcd-io/etcd/pull/15104#issuecomment-1416371288 only introduced a failpoint and didn't make any changes to traffic I would not Snapshot failpoints to cause problems. This a strong sign of etcd issue.

### How can we reproduce it (as minimally and precisely as possible)?

1. Clone latest main branch
2. Run `make gofail-enable && make && make gofail-disable`
3. Uncomment Snapshot scenario from tests/linearizability/linearizability_test.go
4. Run `GO_TEST_FLAGS='-v --count=100 --failfast --run TestLinearizability/Snapshot ' make test-linearizability`

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
etcd Version: 3.6.0-alpha.0
Git SHA: 3c64ae4b7
Go Version: go1.19
Go OS/Arch: linux/amd64
```
</details>

### Etcd configuration (command line flags or environment variables)

<details>
One from Snapshot scenario

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>
Not relevant

</details>

serathius:
cc @ahrtr @ptabor

serathius:
cc @logicalhan

serathius:
Looks like it reproduces https://github.com/etcd-io/etcd/pull/15273

Problem is not caused by specific golang failpoint injection, but just blackholing traffic for enough time to Snapshot be sent. Managed to reproduce it on v3.5. 

[...]

PR #15273, opened 2023-02-09
Reproduce #15271

serathius:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #15288, opened 2023-02-12
test: correct the logic of comparing revision in linearizablity/watch.go

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/15271

Note that there are two revisions:
1. The revision in the response header;
2. The ModRevision in each event.

Note that in each watchResponse, the highest event.ModRevision may be less than the header.Revision. 

Let's work with an example. Assuming the latest header.Revision and (highest) event.ModRevision are both 1000, but the etcdserver is somehow partitioned status, and when it comes back to normal, it syncs the latest logs (e.g. sync a snapshot) from the leader, and it's latest revision may jump to 4000 directly. But the client side still need to receive events starting from 1001. In this case, what the client will receive in the first watchResponse will be something like below,
- watchResponse.Header.Revision: 4000
- First event.ModRevision: 1001
- Last event.ModRevision: 2000 (1000 + len(response.Events) ); Note that this value may be less than resp.Header.Revision (4000 in this example). Note the maximum revisions sent at a time is [1000](https://github.com/etcd-io/etcd/blob/87e271701b3449c35d9ef41f855c7792488773ce/server/storage/mvcc/watcher_group.go#L29).

### DIFF
--- a/tests/linearizability/watch.go
+++ b/tests/linearizability/watch.go
@@ -82,28 +82,37 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
 }
 
 func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expectProgressNotify bool) {
-	var gotProgressNotify = false
-	var lastRevision int64 = 1
+	var (
+		gotProgressNotify          = false
+		lastEventModRevision int64 = 1 // The event.Kv.ModRevision in the latest event.
+		lastHeadRevision     int64 = 1 // The resp.Header.Revision in last watch response.
+	)
 	for _, resp := range responses {
-		if resp.Header.Revision < lastRevision {
-			t.Errorf("Revision should never decrease")
+		if resp.Header.Revision < lastHeadRevision {
+			t.Errorf("Server revision should never decrease, lastHeadRevision: %d, resp.Header.Revision: %d",
+				lastHeadRevision, resp.Header.Revision)
 		}
-		if resp.IsProgressNotify() && resp.Header.Revision == lastRevision {
+
+		if resp.IsProgressNotify() && resp.Header.Revision == lastHeadRevision {
 			gotProgressNotify = true
 		}
-		if resp.Header.Revision == lastRevision && len(resp.Events) != 0 {
+		if resp.Header.Revision == lastHeadRevision && len(resp.Events) != 0 {
 			t.Errorf("Got two non-empty responses about same revision")
 		}
+
 		for _, event := range resp.Events {
-			if event.Kv.ModRevision != lastRevision+1 {
-				t.Errorf("Expect revision to grow by 1, last: %d, mod: %d", lastRevision, event.Kv.ModRevision)
+			if event.Kv.ModRevision != lastEventModRevision+1 {
+				t.Errorf("Expect event revision to grow by 1, last: %d, mod: %d", lastEventModRevision, event.Kv.ModRevision)
 			}
-			lastRevision = event.Kv.ModRevision
+			lastEventModRevision = event.Kv.ModRevision
 		}
-		if resp.Header.Revision != lastRevision {
-			t.Errorf("Expect response revision equal last event mod revision")
+
+		if resp.Header.Revision < lastEventModRevision {
+			t.Errorf("Event revision should never exceed the server's revision, lastEventRevision: %d, resp.Header.Revision: %d",
+				lastEventModRevision, resp.Header.Revision)
 		}
-		lastRevision = resp.Header.Revision
+
+		lastHeadRevision = resp.Header.Revision
 	}
 	if gotProgressNotify != expectProgressNotify {
 		t.Errorf("Expected progress notify: %v, got: %v", expectProgressNotify, gotProgressNotify)
--- a/tests/linearizability/linearizability_test.go
+++ b/tests/linearizability/linearizability_test.go
@@ -144,9 +144,7 @@ func TestLinearizability(t *testing.T) {
 				e2e.WithSnapshotCount(100),
 			),
 		},
-		// TODO: investigate periodic `Model is not linearizable` failures
-		// see https://github.com/etcd-io/etcd/pull/15104#issuecomment-1416371288
-		/*{
+		{
 			name:      "Snapshot",
 			failpoint: RandomSnapshotFailpoint,
 			traffic:   &HighTraffic,
@@ -156,7 +154,7 @@ func TestLinearizability(t *testing.T) {
 				e2e.WithSnapshotCatchUpEntries(100),
 				e2e.WithPeerProxy(true),
 			),
-		},*/
+		},
 	}...)
 	for _, scenario := range scenarios {
 		if scenario.traffic == nil {


ahrtr:
Fix https://github.com/etcd-io/etcd/issues/15271

tests/linearizability/watch.go
```diff
@@ -82,28 +82,37 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
-	var lastRevision int64 = 1
+	var (
+		gotProgressNotify       = false
+		lastEventRevision int64 = 1 // The event.Kv.ModRevision in the latest event.
+		lastHeadRevision  int64 = 1 // The resp.Header.Revision in last watch response.
+	)
 	for _, resp := range responses {
-		if resp.Header.Revision < lastRevision {
-			t.Errorf("Revision should never decrease")
+		if resp.Header.Revision < lastEventRevision || resp.Header.Revision < lastHeadRevision {
```
ptabor, line 91: Could you please split it into 2 ifs ? 
 - failure that event revisions are decreasing should be different from the response revisions.

tests/linearizability/watch.go
```diff
@@ -82,28 +82,37 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
+			if event.Kv.ModRevision != lastEventRevision+1 {
+				t.Errorf("Expect revision to grow by 1, last: %d, mod: %d", lastEventRevision, event.Kv.ModRevision)
 			}
-			lastRevision = event.Kv.ModRevision
+			lastEventRevision = event.Kv.ModRevision
 		}
-		if resp.Header.Revision != lastRevision {
-			t.Errorf("Expect response revision equal last event mod revision")
+
+		if resp.Header.Revision < lastEventRevision || resp.Header.Revision < lastHeadRevision {
```
ptabor, line 110: 'resp.Header.Revision < lastHeadRevision' | could this be successful in line 91 and fail here ?

tests/linearizability/watch.go
```diff
@@ -82,28 +82,37 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
+			t.Errorf("Revision should never decrease[2], lastEventRevision: %d, lastHeadRevision: %d, resp.Header.Revision: %d",
```
ptabor, line 111: I would say something like: `Event revision should never exceed the server's revision`

tests/linearizability/watch.go
```diff
@@ -82,28 +82,37 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
 		}
-		if resp.Header.Revision == lastRevision && len(resp.Events) != 0 {
+		if resp.Header.Revision == lastHeadRevision && len(resp.Events) != 0 {
 			t.Errorf("Got two non-empty responses about same revision")
 		}
+
 		for _, event := range resp.Events {
-			if event.Kv.ModRevision != lastRevision+1 {
-				t.Errorf("Expect revision to grow by 1, last: %d, mod: %d", lastRevision, event.Kv.ModRevision)
+			if event.Kv.ModRevision != lastEventRevision+1 {
```
ptabor, line 104: What if multiple KVs are changed within a single transaction ? 

I assume its possible to emit multiple events at the same events.Kv.ModRevision.

tests/linearizability/watch.go
ahrtr: Since we print all events (lastHeader.Revision, lastEvent.ModRevision, current header.Revision), so we have enough info once running into this issue.

Updated anyway & thx.

tests/linearizability/watch.go
ahrtr: Updated & thx

tests/linearizability/watch.go
ahrtr: NO. I am aware of that it, since it isn't a big deal, so I did not remove the duplicated if condition. 

But updated anyway & thx

tests/linearizability/watch.go
ahrtr: Exactly, I raised [similar comment ](https://github.com/etcd-io/etcd/pull/15250#discussion_r1099327154) in another PR. It's fine since we do not support multi-operations TXN so far. Let's fix this in the PR which supports multi-operation TXN.

ahrtr: Usually I tend to resolve issue step by step. I am fixing the issue what we have so far. Please let me know if you have big concern on this. thx

tests/linearizability/watch.go
```diff
@@ -82,28 +82,37 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
 }
 
 func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expectProgressNotify bool) {
-	var gotProgressNotify = false
-	var lastRevision int64 = 1
+	var (
+		gotProgressNotify       = false
+		lastEventRevision int64 = 1 // The event.Kv.ModRevision in the latest event.
+		lastHeadRevision  int64 = 1 // The resp.Header.Revision in last watch response.
```
serathius, line 88:
```suggestion
		lastEventModRevision int64 = 1
		lastResponseHeaderRevision  int64 = 1
```
Don't use comments for things easily expressed in name

ahrtr: Why can't we use comment in code?

Changed `lastEventRevision` to `lastEventModRevision`.

serathius: Adding comments should not replace good naming.

ahrtr: The purpose of adding comment is **NOT** to replace good naming. 

Even experienced contributors may be confused about the two concepts (Header.Revision and event.ModRevision) and raise false alarm, you never know how other contributors (especially new contributors) read/understand your code, and what's the difference between them. It makes sense to me to add comment to clearly clarify what each of the two variables exactly means.

Note: I agree that we should give each variable/function/method a good name as possible as we can. Each event includes both `CreateRevision` and `ModRevision`, so changed `lastEventRevision` to `lastEventModRevision` to make it clearer.

tests/linearizability/watch.go
serathius: Have a ready PR that will add multi op TXN https://github.com/etcd-io/etcd/pull/15259
Let's address it there.

serathius:
Change makes sense, great catch!
Still I don't think this issue was responsible for https://github.com/etcd-io/etcd/issues/15271

ahrtr:
> Still I don't think this issue was responsible for #15271

I do not get the point. Do you mean you can still reproduce https://github.com/etcd-io/etcd/issues/15271 after applying this PR?

serathius:
> > Still I don't think this issue was responsible for #15271
> 
> I do not get the point. Do you mean you can still reproduce #15271 after applying this PR?

Yes

serathius:
Let's merge as to remove flakiness in lineariability tests. I have confirmed that with this we can close timed watch progress notification as fixed.

ahrtr:
> Yes

Based on your following comment https://github.com/etcd-io/etcd/pull/15288#issuecomment-1427774366, so this comment isn't valid? Just to double check to make sure I do not misunderstand anything.

serathius:
Please check PR with repro https://github.com/etcd-io/etcd/pull/15273

serathius:
Oh, I think I know what it can be. Lease delete might delete multiple keys at once.

PR #15291, opened 2023-02-13
etcd-mixin: Limit etcdDatabaseHighFragmentationRatio to 100MB+

ggustafsson:
Prometheus rule `etcdDatabaseHighFragmentationRatio` is excessively trigger happy in its current form when the etcd database is small. I think it would be better if the metrics are ignored unless total database size is 100MB+ (or something similar). When the size is small the percentage is large and there isn't a real problem with etcd so this is just creates noise.

Signed-off-by: Göran Gustafsson <gustafsson.g@gmail.com>

### DIFF
--- a/contrib/mixin/mixin.libsonnet
+++ b/contrib/mixin/mixin.libsonnet
@@ -241,7 +241,7 @@
           {
             alert: 'etcdDatabaseHighFragmentationRatio',
             expr: |||
-              (last_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / last_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5
+              (last_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / last_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5 and etcd_mvcc_db_total_size_in_use_in_bytes > 104857600
             ||| % $._config,
             'for': '10m',
             labels: {
--- a/contrib/mixin/test.yaml
+++ b/contrib/mixin/test.yaml
@@ -143,13 +143,13 @@ tests:
   - interval: 1m
     input_series:
       - series: 'etcd_mvcc_db_total_size_in_use_in_bytes{job="etcd",instance="10.10.10.0"}'
-        values: '30000+0x10'
+        values: '300000000+0x10'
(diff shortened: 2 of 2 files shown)


jmhbnz (commented):
Hey @ggustafsson  Thanks for raising this, this change itself looks reasonable, can you please ensure your commit is signed to ensure this contribution can be merged.

```bash
$ git rebase HEAD~1 --signoff
$ git push --force-with-lease
```

ggustafsson:
@jmhbnz Done!

ahrtr:
Thanks for the PR, and it looks good to me. Please fix the test failure.

ggustafsson:
I've fixed the test. Running `promtool test rules test.yml` locally works fine for me.

contrib/mixin/mixin.libsonnet
```diff
@@ -241,7 +241,7 @@
           {
             alert: 'etcdDatabaseHighFragmentationRatio',
             expr: |||
-              (last_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / last_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5
+              (last_over_time(etcd_mvcc_db_total_size_in_use_in_bytes[5m]) / last_over_time(etcd_mvcc_db_total_size_in_bytes[5m])) < 0.5 and etcd_mvcc_db_total_size_in_use_in_bytes > 104857600
```
ahrtr, line 244: I am not a prometheus expert. Based on the official guide: https://prometheus.io/docs/prometheus/latest/querying/operators/#logical-set-binary-operators,  the operator `and` is an intersection operation on two vectors. But in this case, both side are scalars that is either 0 (false) or 1 (true).

It will be great if any prometheus expert can double confirm this.

ggustafsson: Alert is triggered if both are true. No gotchas.

Previous test failed because etcd_mvcc_db_total_size_in_use_in_bytes was below 100MB.
Current test succeeds because etcd_mvcc_db_total_size_in_use_in_bytes is above 100MB.

serathius (approved):

ggustafsson:
@ahrtr @jmhbnz Is this good to go?

jmhbnz (approved):

ahrtr (approved):

PR #15293, opened 2023-02-13
 Codify kubernetes using response header revision as bookmark

serathius:
Kubernetes keeps progress on last revision it has seen so if broken, watches can be re-established from last seen revision. It uses:
* [event ModRevision for normal watch response](https://github.com/kubernetes/kubernetes/blob/bf7906674952a08f26e3a7cce7aaf35b1d56bd1e/staging/src/k8s.io/apiserver/pkg/storage/etcd3/event.go#L56)
* [response header Revision for progress notify](https://github.com/kubernetes/kubernetes/blob/bf7906674952a08f26e3a7cce7aaf35b1d56bd1e/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L253)

For this bookmarking to work we need to guarantee that there are no events with revision prior. For ModRevision this is trivial as it's already assumed to be incremental. However for progress notify revision we need to add a specific validation. 

cc @ahrtr @ptabor

### DIFF
--- a/tests/linearizability/watch.go
+++ b/tests/linearizability/watch.go
@@ -83,9 +83,10 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
 
 func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expectProgressNotify bool) {
 	var (
-		gotProgressNotify          = false
-		lastEventModRevision int64 = 1 // The event.Kv.ModRevision in the latest event.
-		lastHeadRevision     int64 = 1 // The resp.Header.Revision in last watch response.
+		gotProgressNotify                = false
+		lastEventModRevision       int64 = 1 // The event.Kv.ModRevision in the latest event.
+		lastHeadRevision           int64 = 1 // The resp.Header.Revision in last watch response.
+		lastProgressNotifyRevision int64 = 0 // The resp.Header.Revision in the last progress notify watch response.
 	)
 	for _, resp := range responses {
 		if resp.Header.Revision < lastHeadRevision {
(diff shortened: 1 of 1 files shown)


ahrtr:
It's a little over-engineering.  The function `validateMemberWatchResponses` is important for watch responses validation, and I can imagine that it will be frequently read/checked by contributors, so let's keep it as simple & easy-to-ready as possible. The current function looks better than what this PR changes.

What this PR changes is even worse than https://github.com/etcd-io/etcd/pull/15312 in my opinion. 

At least, let's hold-off the code change until it's clear how we will support multi-operation TXN, because we will definitely change this function by then.

serathius:
I think we are hitting the same topic again, comments. I have a strong opinion that you should never write a comment explaining what your code does. Explaining "WHAT" is responsibility of your code, if it's not clear from reading your code, your code needs to be cleaned up. When to comment? To explain "WHY". Code can only explain what it does, but it's not always clear why you do what you do. 

I strongly recommend reading https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882 and it's section on comments to better understand where I'm comming from.

I don't want continue the discussion more. I wrote code in iterative manner, so let's leave it like this. I will remove the refactor from the code.

ahrtr:
Note actually I love Robert C. Martin's books, and read a couple of his books (e.g. `Agile Software Development,. Principles, Patterns, and Practices`) more than 10 years ago, and partially read his <Clean Code> about 5 years ago.  But it doesn't mean I agree all his points.

ahrtr:
It's normal that different people have different opinions, and usually I don't insist on something which isn't a big deal. 

But just as I mentioned in https://github.com/etcd-io/etcd/pull/15293#issuecomment-1428963975, I can imagine `validateMemberWatchResponses` will be frequently read by contributors, at least I read it a lot of times in the past couple of days. I really don't want to jump to different functions back and forth. Instead read all small & simple validations in one function makes more sense to me.

tests/linearizability/watch.go
```diff
@@ -104,6 +105,9 @@ func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expec
 		}
 
 		for _, event := range resp.Events {
+			if event.Kv.ModRevision <= lastProgressNotifyRevision {
+				t.Errorf("Got event with revision lower than last progress notify event, eventRevision: %d, progressNotifyRevision: %d", event.Kv.ModRevision, lastProgressNotifyRevision)
+			}
```
ahrtr, line 110: This checks looks good to me. But we'd better add comment something like below to explain why we have such check. 

```
etcd will not send progress notification to a watcher until it has synced all events.
```

serathius: Sure, let me put it in the code that is easier to access when debugging a test failure.

ahrtr (approved):

PR #15294, opened 2023-02-13
server/auth: disallow creating empty permission ranges

mitake:
Current `RoleGrantPermission` allows to create permission for ranges like (a, a) or (b, a). These are treated as empty ranges so not harmful. But they are quite confusing and etcd should disallow it.

cc @ahrtr @serathius @spzala @ptabor

### DIFF
--- a/server/auth/range_perm_cache.go
+++ b/server/auth/range_perm_cache.go
@@ -75,8 +75,10 @@ func checkKeyInterval(
 	cachedPerms *unifiedRangePermissions,
 	key, rangeEnd []byte,
 	permtyp authpb.Permission_Type) bool {
-	if len(rangeEnd) == 1 && rangeEnd[0] == 0 {
+	if isOpenEnded(rangeEnd) {
 		rangeEnd = nil
+		// nil rangeEnd will be converetd to []byte{}, the largest element of BytesAffineComparable,
+		// in NewBytesAffineInterval().
 	}
 
 	ivl := adt.NewBytesAffineInterval(key, rangeEnd)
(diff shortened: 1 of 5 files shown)


mitake:
Sorry for missing the issue. As a retrospective, I'm considering to improve the interval tree library for denying inputs like begin >= end in New<Type>Interval() functions e.g. [NewBytesAffineInterval ](https://github.com/etcd-io/etcd/blob/main/pkg/adt/interval_tree.go#L938). I also think updating other functions like [Insert()](https://github.com/etcd-io/etcd/blob/main/pkg/adt/interval_tree.go#L469) might be possible. But they will update signature types for having a new error. I'd like to have other people's opinions.

tests/integration/v3_auth_test.go
```diff
@@ -406,7 +406,7 @@ func TestV3AuthOldRevConcurrent(t *testing.T) {
 		role, user := fmt.Sprintf("test-role-%d", i), fmt.Sprintf("test-user-%d", i)
 		_, err := c.RoleAdd(context.TODO(), role)
 		testutil.AssertNil(t, err)
-		_, err = c.RoleGrantPermission(context.TODO(), role, "", clientv3.GetPrefixRangeEnd(""), clientv3.PermissionType(clientv3.PermReadWrite))
+		_, err = c.RoleGrantPermission(context.TODO(), role, "\x00", clientv3.GetPrefixRangeEnd(""), clientv3.PermissionType(clientv3.PermReadWrite))
```
mitake, line 409: This role isn't used in the test scenario actually, so it hadn't been harmful.

server/auth/range_perm_cache.go
```diff
@@ -155,3 +157,49 @@ type unifiedRangePermissions struct {
...
+		return true
+	}
+
+	begin := adt.BytesAffineComparable(key)
+	end := adt.BytesAffineComparable(rangeEnd)
+	if begin.Compare(end) == -1 { // rule b2
+		return true
+	}
+
+	if len(rangeEnd) == 1 && rangeEnd[0] == 0x00 { // rule b3
```
ahrtr, line 200: I think it makes sense to create a function something like below to clearly define such open ended case,

```
func isOpenEnded(rangeEnd []byte) bool {
    return len(rangeEnd) == 1 && rangeEnd[0] == 0
}
```

[line 78](https://github.com/etcd-io/etcd/blob/8e56e2f5ce4c43e81877b16239c48f8ff3a0dc14/server/auth/range_perm_cache.go#L78) should be updated to use this function as well.

server/auth/store_test.go
```diff
@@ -525,6 +526,129 @@ func TestRoleGrantPermission(t *testing.T) {
...
+func TestRoleGrantInvalidPermission(t *testing.T) {
+	as, tearDown := setupAuthStore(t)
+	defer tearDown(t)
+
+	_, err := as.RoleAdd(&pb.AuthRoleAddRequest{Name: "role-test-1"})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tests := []struct {
```
ahrtr, line 538: Can we add a field `name` in the struct to describe each case, and use golang sub-test?

mitake:
Thanks for checking @ahrtr , I updated this PR based on your comments. Could you take a look?

server/auth/range_perm_cache_test.go
```diff
@@ -100,3 +130,85 @@ func TestKeyPermission(t *testing.T) {
...
+		},
+		{
+			// valid key range, which includes all keys
+			[]byte("\x00"),
+			[]byte("\x00"),
+			true,
+		},
+	}
+
+	for i, tt := range tests {
```
ahrtr, line 208: The comment https://github.com/etcd-io/etcd/pull/15294#discussion_r1111372666 applies here as well.

server/auth/range_perm_cache.go
```diff
@@ -155,3 +157,54 @@ type unifiedRangePermissions struct {
...
+	return len(rangeEnd) == 1 && rangeEnd[0] == 0
+}
+
+func isValidPermissionRange(key, rangeEnd []byte) bool {
+	if key == nil { // ensure rule a1
+		return false
+	}
+	if len(key) == 0 { // ensure rule a2
+		return false
+	}
```
ahrtr, line 194: nit: len(nil) == 0

```suggestion
	if len(key) == 0 {
		return false
	}
```

ahrtr (approved):
LGTM with a couple of minor comments.

Thanks @mitake

server/auth/range_perm_cache_test.go
mitake: ah sorry, let me update here too

ahrtr (approved):

ahrtr:
cc @ptabor @serathius @spzala the PR has more than 300 lines of code change, but actually it's just a minor change. It has only around 30 lines of production code change, all others are unit test cases.  PTAL.

mitake:
@ptabor @serathius @spzala it’s great if you can cross check

spzala (approved):
Thank you @mitake @ahrtr

server/auth/range_perm_cache.go
```diff
@@ -155,3 +157,51 @@ type unifiedRangePermissions struct {
 	readPerms  adt.IntervalTree
 	writePerms adt.IntervalTree
 }
+
+// Constraints related to key range
```
serathius, line 161: Does same rules apply for making a range request?

mitake: Yes, the same rules are used by range, watch and delete.

serathius: Could they use the same function to validate the range?

serathius: I'm more worried about consistency, feel free to create an issue with the suggestion.

PR #15312, opened 2023-02-14
test: add more comment for linearizalibity watch validation

ahrtr:

### DIFF
--- a/tests/linearizability/watch.go
+++ b/tests/linearizability/watch.go
@@ -88,6 +88,7 @@ func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expec
 		lastHeadRevision     int64 = 1 // The resp.Header.Revision in last watch response.
 	)
 	for _, resp := range responses {
+		// Validation 1: server revision should never decrease.
 		if resp.Header.Revision < lastHeadRevision {
 			t.Errorf("Server revision should never decrease, lastHeadRevision: %d, resp.Header.Revision: %d",
 				lastHeadRevision, resp.Header.Revision)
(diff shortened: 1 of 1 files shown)


ptabor:
I understand that difference vs. #15319 is that this CL is using a single loop why @serathius tries to assert fully different properties independently. 

This PR: 
  + might be slightly faster (but I would assume the computational costs is not significant vs. disk operations)
  + is  commented about individual checks. 
    
#15319: 
  + makes each check more self-contained: all the 'state' is close to the actual check, so is easier to reason about.
  ~ might require expanding comments (but also due property above, the comments will be more like repeating the code)
  + I understand that @serathius intention is to 'name' each individual invariant (and method name is here the invariant name), such that this names can be used in documentation / error messages. 
  
Is there a discussion somewhere with additional  arguments, I should take a look at ?

ahrtr:
Thanks @ptabor for the feedback, which basically makes sense to me. I always believe and agree that **source code is (and should be) the best document.** 

The key comment on https://github.com/etcd-io/etcd/pull/15319 (extracted from previous https://github.com/etcd-io/etcd/pull/15293) is over-engineering. When I was raising the comment (https://github.com/etcd-io/etcd/pull/15293#issuecomment-1428963975 and https://github.com/etcd-io/etcd/pull/15293#issuecomment-1429446382), the watch validation had only 25 lines of code; and each validation is simple & straightforward, it didn't make any sense to split the each small validation to a separate small function. Reading all the small validations in one function (which was already small enough) definitely make more sense to me. Of course, it's a little bigger and more complex ( [33 lines of code](https://github.com/etcd-io/etcd/blob/116a3150c0a52389f7ccdec3f25085dabfc8ce6f/tests/linearizability/watch.go#L100-L133)  ) now, I still think it's a little over-engineering, but less diligent.

https://github.com/etcd-io/etcd/blob/b4f9bb45fcb68b568e261850763df9e99c93fd16/tests/linearizability/watch.go#L91-L116

Note that I didn't mean to merge this PR, I just wanted to show that https://github.com/etcd-io/etcd/pull/15319 (extracted from previous https://github.com/etcd-io/etcd/pull/15293) was even worse than this PR.

Since majority maintainers agree https://github.com/etcd-io/etcd/pull/15319 is better, I will not insist on it; and accordingly will not spend more time on this discussion.

PR #15316, opened 2023-02-14
Do not run arm64 jobs on forks

dims:
Using the tip from [1]. 

[1] https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#example-only-run-job-for-specific-repository

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


ptabor (approved):
Thank you.

Issue #15317, opened 2023-02-14
Linearizability test duplicate event mod revision


serathius:
### Which github workflows are flaking?

Linearizability & Linearizability Nightly

### Which tests are flaking?

TestLinearizability/Issue14370 x2 https://github.com/etcd-io/etcd/actions/runs/4173051956/jobs/7224915747 https://github.com/etcd-io/etcd/actions/runs/4172783085/jobs/7224313444
TestLinearizability/Issue14685 x3 https://github.com/etcd-io/etcd/actions/runs/4172779926/jobs/7224306295 https://github.com/etcd-io/etcd/actions/runs/4167747802/jobs/7213715136 https://github.com/etcd-io/etcd/actions/runs/4172334944/jobs/7223305362
 
 

### Github Action link

Above

### Reason for failure (if possible)

```
watch.go:105: Expect event revision to grow by 1, last: 198, mod: 198
```

I think the reason is that watch validation assumes one operation per revision. This is almost true, however exception is LeaseRevoke which might delete multiple keys bound to single lease.

### Anything else we need to know?

Will implement the fix.
cc @ahrtr @ptabor

serathius:
On second thought I don't think the tests that flaked here could be impacted by leases. Those two specific scenarios do not use traffic that uses leases. I will fix the duplicate mod revision and wait for more data.

ptabor:
Maybe it makes sense to print the type/key of the problematic Watch event to have more clue in case of such problems.

serathius:
Please see https://github.com/etcd-io/etcd/pull/15282
It add dumping whole watch response history into json file.

ahrtr:
Can we close this issue since https://github.com/etcd-io/etcd/pull/15318 is already merged?

serathius:
> Can we close this issue since #15318 is already merged?

It's not clear if duplicate operations could be caused by lease as they were not present in the scenario that flaked. We relaxed the revision validation in https://github.com/etcd-io/etcd/pull/15318, however I think there might still be incorrect behavior. I'm considering adding validation for duplicate watch events.

serathius:
Added https://github.com/etcd-io/etcd/pull/15324
Let's wait for couple of runs of linearizability tests. Maybe one nightly run too.

serathius:
No issues with duplicated event mod revision since yesterday

PR #15318, opened 2023-02-14
Allow multiple operations with same mod revisions, but forbid …

serathius:
…those operations to be spread between multiple responses

ref https://github.com/etcd-io/etcd/issues/15317
cc @ahrtr @ptabor

### DIFF
--- a/tests/linearizability/watch.go
+++ b/tests/linearizability/watch.go
@@ -99,10 +99,13 @@ func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expec
 		if resp.Header.Revision == lastHeadRevision && len(resp.Events) != 0 {
 			t.Errorf("Got two non-empty responses about same revision")
 		}
+		if len(resp.Events) > 0 && resp.Events[0].Kv.ModRevision <= lastEventModRevision {
+			t.Errorf("Events with same revision should not be split between multiple responses, lastResponseModRevision: %d, firstEventModRevision: %d", lastEventModRevision, resp.Events[0].Kv.ModRevision)
+		}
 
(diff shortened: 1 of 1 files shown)


tests/linearizability/watch.go
```diff
@@ -99,10 +99,13 @@ func validateMemberWatchResponses(t *testing.T, responses []watchResponse, expec
 		if resp.Header.Revision == lastHeadRevision && len(resp.Events) != 0 {
 			t.Errorf("Got two non-empty responses about same revision")
 		}
+		if len(resp.Events) > 0 && resp.Events[0].Kv.ModRevision <= lastEventModRevision {
```
ptabor, line 102: I thought that this fields: 

https://github.com/etcd-io/etcd/blob/6c439aa9e2d4a53c63e8a28a227fc6e2f3a36ac3/api/etcdserverpb/rpc.proto#L835

determines whether we can expect continuation of the same modRevision in next answer or not.

serathius: Forgot about this, however it doesn't seem it's exposed to client https://github.com/etcd-io/etcd/blob/6c439aa9e2d4a53c63e8a28a227fc6e2f3a36ac3/client/v3/watch.go#L96-L115

serathius: Fragment handling is done on client library side. Client will aggregate fragmented responses until `fragment==false`. https://github.com/etcd-io/etcd/blob/6c439aa9e2d4a53c63e8a28a227fc6e2f3a36ac3/client/v3/watch.go#L644 

This means proposed change is correct.

ptabor (approved):

PR #15319, opened 2023-02-14
tests: Refactor watch validation

serathius:
Spin off from https://github.com/etcd-io/etcd/pull/15293.

This refactor competes with https://github.com/etcd-io/etcd/pull/15312

End goal: Create a list of semantic behaviors in etcd Watch API that can tested and crossed checked with how K8s uses etcd.
cc @ahrtr


tests/linearizability/watch.go
```diff
@@ -82,40 +82,73 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
+	validateResponseHeaderRevisionNeverDecreases(t, responses)
+	validateModRevisionIsStrictlyIncremental(t, responses)
+	validateHeaderRevisionIsLargerThanModRevision(t, responses)
+}
+
+func validateDuplicateHeaderRevisionMeansNoEvents(t *testing.T, responses []watchResponse) {
+	var lastHeadRevision int64 = 1
+	for _, resp := range responses {
+		if resp.Header.Revision == lastHeadRevision && len(resp.Events) != 0 {
+			t.Errorf("Got two non-empty responses about same revision")
```
ptabor, line 99: Don't we want to print the problematic revision. 

In theory it might be useful to correlate it later with e.g. watch-log ?

tests/linearizability/watch.go
```diff
@@ -82,40 +82,73 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
 		if resp.IsProgressNotify() && resp.Header.Revision == lastHeadRevision {
 			gotProgressNotify = true
 		}
-		if resp.Header.Revision == lastHeadRevision && len(resp.Events) != 0 {
-			t.Errorf("Got two non-empty responses about same revision")
-		}
+		lastHeadRevision = resp.Header.Revision
+	}
+	if gotProgressNotify != expectProgressNotify {
+		t.Errorf("Expected progress notify: %v, got: %v", expectProgressNotify, gotProgressNotify)
```
ptabor, line 126: Why do we expect progress notify in each response. 
Would be good to give more hints in the error message, or at least a comment.

tests/linearizability/watch.go
```diff
@@ -82,40 +82,73 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
 			}
 			lastEventModRevision = event.Kv.ModRevision
 		}
+	}
+}
 
-		if resp.Header.Revision < lastEventModRevision {
-			t.Errorf("Event revision should never exceed the server's revision, lastEventRevision: %d, resp.Header.Revision: %d",
-				lastEventModRevision, resp.Header.Revision)
+func validateHeaderRevisionIsLargerThanModRevision(t *testing.T, responses []watchResponse) {
```
ptabor, line 144: IsNotLower... as equals is OK.

ptabor (approved):

tests/linearizability/watch.go
serathius: Done

tests/linearizability/watch.go
serathius: Done

serathius:
cc @ahrtr @ptabor 
Ready to review

tests/linearizability/watch.go
```diff
@@ -85,61 +85,94 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
+	validateHeaderRevisionGreaterOrEqualToEventRevision(t, responses)
+	validateEventsSharingRevisionWithinSameResponse(t, responses)
+}
 
+func validateGotAtLeastOneProgressNotify(t *testing.T, responses []watchResponse, expectProgressNotify bool) {
+	var gotProgressNotify = false
+	var lastHeadRevision int64 = 1
+	for _, resp := range responses {
 		if resp.IsProgressNotify() && resp.Header.Revision == lastHeadRevision {
 			gotProgressNotify = true
```
ahrtr, line 104:
```suggestion
			gotProgressNotify = true
			break
```

tests/linearizability/watch.go
```diff
@@ -85,61 +85,94 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
 		}
+	}
+}
 
-		if resp.IsProgressNotify() {
-			lastProgressNotifyRevision = resp.Header.Revision
+func validateHeaderRevisionGreaterOrEqualToEventRevision(t *testing.T, responses []watchResponse) {
+	for _, resp := range responses {
+		for _, event := range resp.Events {
+			if resp.Header.Revision < event.Kv.ModRevision {
```
ahrtr, line 159: We have already verified that the event revision is strictly incremental in `validateEventRevisionIsContinuousAndNonDecreasing`, please just compare headerRevision with the last event revision if present.

tests/linearizability/watch.go
```diff
@@ -85,61 +85,94 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
+				t.Errorf("BROKE: Events with revision lower than progress notify were delivered, eventRevision: %d, progressNotifyRevision: %d", event.Kv.ModRevision, lastProgressNotifyRevision)
 			}
+		}
+		if resp.IsProgressNotify() {
+			lastProgressNotifyRevision = resp.Header.Revision
+		}
+	}
+}
+
+func validateEventRevisionIsContinuousAndNonDecreasing(t *testing.T, responses []watchResponse) {
```
ahrtr, line 127: I raised the comment `Events in the same watch response must have the same revision, otherwise it must be strictly incremental across responses`,  but eventually removed the comment after seeing function `validateEventsSharingRevisionWithinSameResponse`.

Please merge it and `validateEventsSharingRevisionWithinSameResponse` into just one function.  It's clearer to get them included in one function instead of two.

tests/linearizability/watch.go
serathius: Want to keep the validators independent.

tests/linearizability/watch.go
ahrtr: I'd like to repeat the removed comment: `Events in the same watch response must have the same revision, otherwise it must be strictly incremental across responses`.  The two functions are actually validating this rule. Please consider to define just one function.

tests/linearizability/watch.go
ahrtr: OK, this isn't a big deal. Feel free to keep it as it's.

serathius:
Updated the validators based on pubclicly documented guarantees https://etcd.io/docs/v3.6/learning/api/#watch-streams plus one guarantee introduced for Kubernetes. Possibly we should document it too.

ahrtr (changes requested):
Why do you remove some verification on the response header revision? 

Can this PR verify the case "`Events in the same watch response must have the same revision`" just as I mentioned in https://github.com/etcd-io/etcd/pull/15319#discussion_r1119447841

serathius:
> Why do you remove some verification on the response header revision?
> 
> Can this PR verify the case "`Events in the same watch response must have the same revision`" just as I mentioned in [#15319 (comment)](https://github.com/etcd-io/etcd/pull/15319#discussion_r1119447841)

This is called atomic in https://etcd.io/docs/v3.6/learning/api/#watch-streams

serathius:
> > single response can have multiple revisions
> 
> No, it should never happen, otherwise, it's a bug. All events included in one watchresponse must have the same ModRevision. It's exactly the reason why I repeat [#15319 (comment)](https://github.com/etcd-io/etcd/pull/15319#discussion_r1119447841).

Make the change to your suggestion, but the validator immediately reports error. Please double check and provide code line that makes you think this is an error. Please remember to distinguish server level watchResponse a a clientv3 client watchResponse. Difference is that client does some response merging to hide chunking.

> Note that WatchResponse only includes DELETE and PUT operations. It means each event's ModRevision must have already increased by one as compared to previous TXN. For a TXN, it's highest ModRevision can only increase by one at most.

What I'm seeing is events from multiple TXN in same response.

serathius:
Please remember this is an refactor, what you are proposing is a new validation.

ahrtr:
> What I'm seeing is events from multiple TXN in same response.

There are two cases. For real time watch notification in `(*watchableStoreTxnWrite) End()`, each watchResponse has events from the same TXN. But in the `(*watchableStore) syncWatchers()`,  up to 1000 distinct revisions may be sent to a unsynced watcher in a watchResponse. Sorry for the confusion. 




> Why do you remove some verification on the response header revision?

It seems you do not resolve this comment yet.

tests/robustness/watch.go
```diff
@@ -85,61 +85,97 @@ func validateWatchResponses(t *testing.T, responses [][]watchResponse, expectPro
...
+	for _, resp := range responses {
+		for _, event := range resp.Events {
+			if event.Kv.ModRevision < lastEventRevision {
+				t.Errorf("BROKE: Ordered - events are ordered by revision; an event will never appear on a watch if it precedes an event in time that has already been posted, lastRevision: %d, currentRevision: %d", lastEventRevision, event.Kv.ModRevision)
 			}
-			lastEventModRevision = event.Kv.ModRevision
+			lastEventRevision = event.Kv.ModRevision
 		}
+	}
+}
```
ahrtr, line 139: `validateOrdered` seems not needed at all, it's already covered by `validateReliable`.  If we ensure each `event.Kv.ModRevision` is either `lastEventRevision` or `lastEventRevision + 1`, then it means that the events must be ordered by the ModRevision.

We can consider to rename `validateReliable` to `validateReliableAndOrdered`. WDYT?

serathius: Should be good for now, however in future I would like to separate ordered. Ordered should mean that watch events are ordered in reference to their execution on server.

serathius:
Please take another look @ahrtr

ahrtr (approved):

PR #15104, opened 2023-01-13
tests linearizability: trigger snapshot related failpoints

lavacat:
fixes https://github.com/etcd-io/etcd/issues/14726, except `raftBeforeFollowerSend`

Signed-off-by: Bogdan Kanivets <bkanivets@apple.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #15323, opened 2023-02-15
tests: use monotonic clock for linearizability test history

lavacat:
Problem: time.Now() uses wall clock reading. See https://github.com/golang/go/blob/master/src/time/time.go#L17 "later time-telling operations use the wall clock reading, but later time-measuring operations, specifically comparisons and subtractions, use the monotonic clock reading." This can cause 'Return' to be before 'Call' and wrong order of operations from different clients.

Solution: use same base time for all clients and only use 'time-measuring' operations to record timestamps for history.

### DIFF
--- a/tests/linearizability/client.go
+++ b/tests/linearizability/client.go
@@ -27,11 +27,12 @@ import (
 )
 
 type recordingClient struct {
-	client  clientv3.Client
-	history *model.AppendableHistory
+	client   clientv3.Client
+	history  *model.AppendableHistory
+	baseTime time.Time
 }
 
-func NewClient(endpoints []string, ids identity.Provider) (*recordingClient, error) {
+func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time) (*recordingClient, error) {
 	cc, err := clientv3.New(clientv3.Config{
 		Endpoints:            endpoints,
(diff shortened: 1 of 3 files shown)


lavacat:
cc @serathius this fixes for me https://github.com/etcd-io/etcd/pull/15104#issuecomment-1416371288

it might be possible to get the same result without `baseTime` var. See https://stackoverflow.com/questions/71689285/how-to-get-monotonic-part-of-time-time-in-go

tests/linearizability/client.go
```diff
@@ -27,11 +27,12 @@ import (
 )
 
 type recordingClient struct {
-	client  clientv3.Client
-	history *model.AppendableHistory
+	client   clientv3.Client
+	history  *model.AppendableHistory
+	baseTime time.Time
```
serathius, line 32: Can you add a comment explaining why we have baseTime and why time measurements should  relative to baseTime?

serathius (approved):
Awesome find!

serathius:
Would be also good to use relative time in watches to make a single timeline to all types of observations we make in the test. Would you be able to contribute it in followup?

serathius:
Let's merge it to reduce flakes in linearizability tests.

PR #15325, opened 2023-02-15
tests: Refactor file structure

serathius:
cc @ahrtr @ptabor


ahrtr (approved):
Overall looks good to me. 

Having more than 300 lines of change in one commit looks a little scary. If I did this PR, I would split it into a couple of small commits something like below at least so as to make it easier to review,
- Move failpoint related code (e.g. `FailpointConfig` and `triggerFailpoints`) into `failpoints.go`;
- Move traffic related code (e.g. trafficConfig and `simulateTraffic`) into `traffic.go`;
- Move Watch related code into watch.go;
- ....

PR #15326, opened 2023-02-15
tests/linearizability: add client option and refactor history Merge

chaochn47:
An incremental change to add auth support in linearizable test. 
1. make root user and test user accessible from outside of the `test/common` package
2. add client option
3. history merge can be simplified

### DIFF
--- a/tests/common/auth_util.go
+++ b/tests/common/auth_util.go
@@ -24,48 +24,48 @@ import (
 )
 
 const (
-	rootUserName = "root"
-	rootRoleName = "root"
-	rootPassword = "rootPassword"
-	testUserName = "test-user"
-	testRoleName = "test-role"
-	testPassword = "pass"
+	RootUserName = "root"
+	RootRoleName = "root"
+	RootPassword = "rootPassword"
+	TestUserName = "test-user"
+	TestRoleName = "test-role"
+	TestPassword = "pass"
 )
 
 var (
-	rootUser = authUser{user: rootUserName, pass: rootPassword, role: rootRoleName}
-	testUser = authUser{user: testUserName, pass: testPassword, role: testRoleName}
+	RootUser = AuthUser{User: RootUserName, Pass: RootPassword, Role: RootRoleName}
+	TestUser = AuthUser{User: TestUserName, Pass: TestPassword, Role: TestRoleName}
 
-	testRole = authRole{
-		role:       testRoleName,
-		permission: clientv3.PermissionType(clientv3.PermReadWrite),
-		key:        "foo",
-		keyEnd:     "",
+	TestRole = AuthRole{
+		Role:       TestRoleName,
+		Permission: clientv3.PermissionType(clientv3.PermReadWrite),
+		Key:        "foo",
+		KeyEnd:     "",
 	}
 )
 
-type authRole struct {
-	role       string
-	permission clientv3.PermissionType
-	key        string
-	keyEnd     string
+type AuthRole struct {
+	Role       string
+	Permission clientv3.PermissionType
+	Key        string
+	KeyEnd     string
 }
 
-type authUser struct {
-	user string
-	pass string
-	role string
+type AuthUser struct {
+	User string
+	Pass string
+	Role string
 }
 
-func createRoles(c interfaces.Client, roles []authRole) error {
+func createRoles(c interfaces.Client, roles []AuthRole) error {
 	for _, r := range roles {
 		// add role
-		if _, err := c.RoleAdd(context.TODO(), r.role); err != nil {
+		if _, err := c.RoleAdd(context.TODO(), r.Role); err != nil {
 			return fmt.Errorf("RoleAdd failed: %w", err)
 		}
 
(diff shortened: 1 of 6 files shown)


tests/linearizability/model/history.go
```diff
@@ -382,19 +382,12 @@ type History struct {
...
-	result := History{
-		successful: make([]porcupine.Operation, 0, len(h.successful)+len(h2.successful)),
-		failed:     make([]porcupine.Operation, 0, len(h.failed)+len(h2.failed)),
-	}
-	result.successful = append(result.successful, h.successful...)
-	result.successful = append(result.successful, h2.successful...)
-	result.failed = append(result.failed, h.failed...)
-	result.failed = append(result.failed, h2.failed...)
-	return result
+func (h *History) Merge(h2 History) {
```
lavacat, line 385: nit: LGTM, but I wonder what was the reason for original [implementation](https://github.com/etcd-io/etcd/commit/d8da98c49fbb6b61faf805419c5356a05f43793f#diff-6c2dd0f0e7caef34835e61886c9988d902e3184c2433b30dc51686d30e7538c1) using value receiver? cc @serathius

lavacat (approved):

tests/common/auth_util.go
```diff
@@ -24,72 +24,72 @@ import (
...
-type authRole struct {
-	role       string
-	permission clientv3.PermissionType
-	key        string
-	keyEnd     string
+type AuthRole struct {
+	Role       string
+	Permission clientv3.PermissionType
+	Key        string
+	KeyEnd     string
```
ahrtr, line 51: Exporting each field isn't a best practice to me. I would suggest to add a method something like below for each field,

```
func (ar AuthRole) Role() string {
    return ar.role
}
```

Since currently all fields will not written from outside the package, so no need to add method something like `SetRole(string)` for now.

tests/linearizability/model/history.go
serathius: Value receiver implies that value is not mutated.

serathius, line 385: Don't understand why making this method mutable helps anything.

tests/linearizability/model/history.go
```diff
@@ -382,19 +382,12 @@ type History struct {
...
+	h.successful = append(h.successful, h2.successful...)
+	h.failed = append(h.failed, h2.failed...)
 }
 
-func (h History) Operations() []porcupine.Operation {
+func (h *History) Operations() []porcupine.Operation {
```
serathius, line 390: Why?

tests/linearizability/client.go
```diff
@@ -32,13 +33,17 @@ type recordingClient struct {
 	baseTime time.Time
 }
 
-func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time) (*recordingClient, error) {
-	cc, err := clientv3.New(clientv3.Config{
+func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time, opts ...config.ClientOption) (*recordingClient, error) {
```
serathius, line 36: Why you add this, I don't see this used anywhere?

tests/common/auth_util.go
serathius: I don't have anything against this. Getters and setters are overkill here.

tests/common/auth_util.go
```diff
@@ -24,72 +24,72 @@ import (
 )
 
 const (
-	rootUserName = "root"
-	rootRoleName = "root"
-	rootPassword = "rootPassword"
-	testUserName = "test-user"
-	testRoleName = "test-role"
-	testPassword = "pass"
+	RootUserName = "root"
```
serathius, line 27: Don't understand reason for making this public. If you plan to use it in linearizability tests, please don't.
Fact that we also need to setup authorization in multiple tests doesn't mean that we need to reuse make configuration global. I would prefer to avoid unnecessary dependendencies. 

If you want to create similar roles in linearizability tests, it would be better to just copy it. Copying is not bad as it implies that don't things are not related.

tests/linearizability/model/history.go
chaochn47: Wanted to pass the pointer of History to another function instead of copying it around. 

Is the concern of making this pointer receiver that multiple Merges mutates the history? 

I can add a lock to prevent in this case then.

tests/linearizability/model/history.go
chaochn47: The same reason as above.

tests/linearizability/client.go
chaochn47: It will be used to add user and password options.

tests/common/auth_util.go
chaochn47: Okay. Makes sense if isolation between linearizability tests and integration/e2e tests is more important. 

I will backout those changes then.

chaochn47:
Given the feedback, I will close this PR and continue on https://github.com/etcd-io/etcd/pull/14819, thanks!

Issue #15332, opened 2023-02-17
Bump golang version to 1.19.6 for both `release-3.5` and `release-3.4` and bump `golang.org/x/net` to v0.7.0


ahrtr:
### What would you like to be added?

For both `release-3.5` and `release-3.4`
1. Bump golang version to 1.19.6;
2. Bump golang.org/x/net to v0.7.0 to address CVE GO-2023-1571 . See also https://github.com/etcd-io/etcd/pull/15331

### Why is this needed?

1. golang 1.17 is out of support; and all active K8s branches (1.23 ~ 1.26) have already bumped to 1.19
2. Refer to https://pkg.go.dev/vuln/GO-2023-1571

ahrtr:
cc @mitake @ptabor @serathius @spzala

fuweid:
@ahrtr it will be included in 3.4.24 release?

ahrtr:
> @ahrtr it will be included in 3.4.24 release?

3.4.24 is already released, so it will only be included in 3.4.25.

fuweid:
Thanks! I missed the release notification.

spzala:
cc @tjungblu

jmhbnz:
@ahrtr I'm keen to give this a go if you don't mind?   I Just want to double check before raising pull request as I've done a couple of the `good-first-issue` contributions recently and I don't mind leaving this if it would be better to let someone else.

ahrtr:
Thanks @jmhbnz .  Just assigned to you. 

We are still pending on other maintainers' comment. It's OK to have more discussion under your PR.

tjungblu:
Thanks, that helps a lot. Do you guys foresee any challenges with going directly to 1.19?

serathius:
cc @liggitt

liggitt:
> Thanks, that helps a lot. Do you guys foresee any challenges with going directly to 1.19?

Not really... there were only three changes Kubernetes encountered going from go1.17 → go1.19, and they don't seem relevant to etcd:
* sha-1 support disabled by default (not an issue unless someone is using etcd with insecure certificates, and they can re-enable with a GODEBUG envvar)
* LookPath returns an error when resolving binaries in the current directory (I don't see etcd using this)
* GC tuning differences that improved latency but could increase peak memory use in kube-apiserver (we didn't see this in all components or clusters, only specific kubernetes configurations, and this was tunable to get the go1.17 balance back by setting a GOGC envvar)

[...]

PR #15337, opened 2023-02-19
[3.5] Backport bump to go 1.19.6 and golang.org/x/net to v0.7.0

jmhbnz:
Golang 1.17 is out of support; and all active K8s branches (1.23 ~ 1.26) have already bumped to 1.19.

Additionally the x/net package has a CVE we need to address: https://pkg.go.dev/vuln/GO-2023-1571

Part of https://github.com/etcd-io/etcd/issues/15332


client/v3/doc.go
```diff
@@ -61,7 +61,7 @@
 //
 //  1. context error: canceled or deadline exceeded.
 //  2. gRPC error: e.g. when clock drifts in server-side before client's context deadline exceeded.
-//  See https://github.com/etcd-io/etcd/blob/main/api/v3rpc/rpctypes/error.go
+//     See https://github.com/etcd-io/etcd/blob/main/api/v3rpc/rpctypes/error.go
```
anqurvanillapy, line 64: This sentence might be started at a new line with a blank one above. Might be mis-formatted by the new Go tools.

etcdctl/main.go
```diff
@@ -27,7 +27,8 @@ const (
 	apiEnv = "ETCDCTL_API"
 )
 
-/**
+/*
+*
```
anqurvanillapy, line 31: Seems like the second asterisk `*` should be deleted? Looks like a mandatory rule of the new Go tools.

raft/log_test.go
```diff
@@ -146,12 +146,14 @@ func TestAppend(t *testing.T) {
 
 // TestLogMaybeAppend ensures:
 // If the given (index, term) matches with the existing log:
-// 	1. If an existing entry conflicts with a new one (same index
-// 	but different terms), delete the existing entry and all that
-// 	follow it
-// 	2.Append any new entries not already in the log
+//  1. If an existing entry conflicts with a new one (same index
```
anqurvanillapy, line 149: This list should be rearranged I guess.

raft/raft_test.go
```diff
@@ -2818,7 +2818,7 @@ func TestRestoreWithLearner(t *testing.T) {
 	}
 }
 
-/// Tests if outgoing voter can receive and apply snapshot correctly.
+// / Tests if outgoing voter can receive and apply snapshot correctly.
```
anqurvanillapy, line 2821: Unnecessary third slash here?

raft/log_test.go
jmhbnz: Good spotting, the formatting tools were a bit annoying here but I think it's looking a bit better now.

ahrtr:
Note this PR can resolve the following CVEs (score 7.5 high) beside the https://pkg.go.dev/vuln/GO-2023-1571:
- https://nvd.nist.gov/vuln/detail/CVE-2022-27664
- https://nvd.nist.gov/vuln/detail/CVE-2022-41715
- https://nvd.nist.gov/vuln/detail/CVE-2022-2880
- https://nvd.nist.gov/vuln/detail/CVE-2022-2879
- https://nvd.nist.gov/vuln/detail/CVE-2022-41716
- https://nvd.nist.gov/vuln/detail/CVE-2022-41720

@jmhbnz could you resolve the workflow (linux-amd64-fmt) failure?

ahrtr:

jmhbnz:
> @jmhbnz could you resolve the workflow (linux-amd64-fmt) failure?

Looks like a bit more than a standard `go fmt` error, the receiver_name test from `./tests.sh` failed with:

```bash
 'receiver_name' started at Mon Feb 20 08:36:39 UTC 2023
Mismatched receiver for Copy(w...
func (d *dispatcherPool) Copy(w io.Writer, f fetchFunc) error {
func (d *dispatcherImmediate) Copy(w io.Writer, f fetchFunc) error {
FAIL: 'receiver_name' failed at Mon Feb 20 08:36:41 UTC 2023
```

I'm not sure why but when I run the same `GOARCH=amd64 PASSES='fmt bom dep' ./test.sh` that the actions workflow is running it does get to the receiver_name section then seems to hang.

While the actions output doesn't print a filename I think the code it's worried about is in `tools/local-tester/bridge/dispatch.go`.

@ahrtr are you able to give me any guidance or background on resolving a receiver mismatch?

ahrtr:
It seems we need to update the [go_srcs_in_module](https://github.com/etcd-io/etcd/blob/822eeaf9f2d9e327df5f4cab8ff075c9b5727320/scripts/test_lib.sh#L110) per the main branch. 

Please try it out.

jmhbnz:
> It seems we need to update the [go_srcs_in_module](https://github.com/etcd-io/etcd/blob/822eeaf9f2d9e327df5f4cab8ff075c9b5727320/scripts/test_lib.sh#L110) per the main branch.

Pushed a new commit that brings the changes to that function back to 3.5. It did open a can of worms with `goword` so I worked my way through fixing all those, hopefully that was the right approach 🙏🏻

ahrtr (approved):
LGTM

Thanks @jmhbnz 

We can ignore the workflow failure for Release, because it always runs test against upstream `release-3.5` instead of the dev branch.

cc @mitake @ptabor @serathius @spzala  PTAL. I think this PR is required for etcd 3.5.8, because it resolves some high CVEs.

serathius (approved):

ptabor (approved):

ahrtr:
@jmhbnz Could you please update both 3.5 and 3.4 changelog?

ahrtr:
This PR also resolved [CVE-2022-41723](https://www.cve.org/CVERecord?id=CVE-2022-41723).

We recently resolved a couple of high CVEs, I suggest to release 3.5.8 and 3.4.25 soon. All other bug fixes can be included in next patches. cc @mitake  @ptabor @serathius @spzala 

I think we need also clearly document that we should release patches asap to address any CVE with a score >= 7.5, unless there is any critical issue which must be included in next patch, otherwise no reason to wait.  WDYT

Issue #15139, opened 2023-01-18
Test etcd container in K8s cluster


ahrtr:
### What would you like to be added?

Currently all the existing automatic test cases are running etcd as process instead of container. But usually etcd is running as POD/container (e.g. static POD) in K8s. So It would be better to add automatic test case to running etcd container in a K8s cluster. 

At least, we should do some sanity test for such scenario. For example, all members should can communicate with each other via the K8s service name (e.g `etcd-1.etcd-headless-svc.svc.cluster.local`). I am not sure whether we can use [Kind](https://github.com/kubernetes-sigs/kind) to do this.

I am thinking the other alternative is to leverage the existing K8s pipeline to do such testing, but in that case we should clearly document the guide/steps, so that everyone can do it.

### Why is this needed?

It can improve etcd release's quality.

serathius:
Not sure about it, have we got any complaints about problems with running etcd in K8s? There are many different way to deploy etcd in K8s and I don't see any benefits on just testing one of them. I think it's up to projects packaging etcd on K8s to test in properly.  

On the other hand, etcd project distributes container image that those project use. This is the artifact passed between etcd project and K8s distributions. We might want to test the image itself to eliminate chances we break K8s users.

To test etcd image we don't need to run K8s, just can start etcd with docker, which should be much simpler.

serathius:
Idea to add test is ok, however I would want to avoid adding new testing framework as we still didn't addressed issues of test flakiness and deduplicating integration and e2e tests.

ahrtr:
Testing the etcd image itself is **must-to-have**. We can do some sanity test with command something like `docker run `.

Testing etcd POD/container in K8s is **nice-to-have** to me. Note we are not going to test K8s, instead we are testing the etcd's integration with K8s. K8s is the most important user of etcd, I think it deserves the effort. At least we should run some sanity test in K8s before publishing each release. 

Some issues which aren't discovered in etcd's existing workflow/pipeline might be relatively easier to be reproduced in K8s environment, such as https://github.com/etcd-io/etcd/pull/14420. We also received issue that etcd's DNS name (e.g. `etcd-0.etcd-discovery.e2e-clusters-8f75l-example-54cx2.svc`) can't be correctly resolved or reversely lookup (https://github.com/etcd-io/etcd/issues/15062) in K8s environment. Although it looks not like an etcd issue (I mean 15062), but it's complained that it's related to a recent change https://github.com/etcd-io/etcd/pull/14573. 

[...]

PR #15270, opened 2023-02-09
Fixes: #15266 All docker images of Architecture show amd64

zhangguanzhang:
Fixes: https://github.com/etcd-io/etcd/issues/15266
Signed-off-by: zhangguanzhang <zhangguanzhang@qq.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #15350, opened 2023-02-23
scripts: Add testing of etcd in local image as part of the release wo…

pchan:
…rkflow. 

cc: @ahrtr 

Fixes: #15139


ahrtr:
Thank you for the PR.

Just as I mentioned in https://github.com/etcd-io/etcd/issues/15139#issuecomment-1419838915, I would suggest to add a step in the existing [release.yaml](https://github.com/etcd-io/etcd/blob/main/.github/workflows/release.yaml), the structure will be something like below,
```
steps:
......
- name: release
  run: |
      ......
      DRY_RUN=true ./scripts/release.sh --no-upload --no-docker-push --in-place 3.6.99
- name: test-image
  run: |
      make test-images 
```

You can add a  new target test-images in Makefile, and add a new script file such as `scripts/test_images.sh`. Usually we add new test in the existing `scripts/test.sh`, but it's too big, so let's add a separate script file. Please keep the build-docker.sh untouched.

.github/workflows/release.yaml
```diff
@@ -9,7 +9,12 @@ jobs:
     - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
       with:
         go-version: "1.19.6"
-    - run: |
+    - name: test-image
+      run: |
+        make test-images
+        make clean
+    - name: release
```
ahrtr, line 16: The step `test-image` should be put after `release`, and uses the images generated by `release` directly.

serathius: True, however we should test image before we release it. So flow should be
```
make release
make test-release
make publish-release
```

ahrtr: Makes sense to me. 

Please try to breakdown the task, either into two separate PRs (preferred) or a couple of small commits in this PR. @pchan 
- breakdown the existing `release` into two targets such as `create-release` and `publish-release`;
- add a new target `test-release` in between to verify the image.

ahrtr: @pchan Please let me know if you need help on the first task (breakdown the existing release).

pchan: That implies we need to refactor release.sh, maybe modularize it or use separate scripts for `create-release` and `publish-release`. Can we use this PR to just track the addition of tests, I will create a separate issue and PRs to address those ? 

One stop-gap fix, till we fix the above, is to add a release-dir to the release script and test images so that we can re-use it. I have attempted it in the latest commit.

ahrtr: > Can we use this PR to just track the addition of tests

It's OK to me. It's always recommended to do things step by step.

pchan: I have created issue https://github.com/etcd-io/etcd/issues/15396 to track the refactoring. As an interim fix, I have changed the logic to re-use the artefacts of the release workflow.

pchan:
@ahrtr  Please review, I have added the testing images as part of the release workflow.

scripts/test_images.sh
```diff
@@ -0,0 +1,75 @@
...
+# Build if binaries are not present
+if [ -z "$BINARYDIR" ]; then
+    run ./scripts/build.sh
+fi
+
+# Build a local image from bin directory
+if ! docker build -t "${TAG}" -f "${DOCKERFILE}" "${BINARYDIR}"; then
+    echo "Docker build unsuccessful. Exit code $?"
+    exit 1
+fi
```
ahrtr, line 47: Why do you build the image? Can you use the images built by previous step directly? 

We can assume that the images are already available, how about passing in the image tag, i.e. v3.5.99?

ahrtr:
I also expect scripts/test_images.sh something like below,
```
function startContainer {
...
}

function stopContainer() {
......
}

function main {
    startContainer

    sanityKVTest
    // Others can add new test cases later, such as verify arch: https://github.com/etcd-io/etcd/pull/15270

    stopContainer

}

```

ahrtr:
@pchan any progress? Note It's blocking a couple of PRs, 
- https://github.com/etcd-io/etcd/pull/15270
- https://github.com/etcd-io/etcd/pull/15424

.github/workflows/release.yaml
```diff
@@ -23,3 +24,6 @@ jobs:
         Expire-Date: 0
         EOF
         DRY_RUN=true ./scripts/release.sh --no-upload --no-docker-push --in-place 3.6.99
+    - name: test-image
+      run: |
+        make test-images
```
ahrtr, line 29: It seems the target test-image isn't necessary. 
```suggestion
        make test-images
        VERSION=3.6.99 ./scripts/test_image.sh
```

scripts/test_images.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# Docker build
+ARCH=$(go env GOARCH)
+DOCKERFILE="Dockerfile-release.${ARCH}"
+BINARYDIR=${BINARYDIR:-"bin"}
+# Pick defaults based on release workflow
+IMAGE=${IMAGE:-"gcr.io/etcd-development/etcd"}
+VERSION=${VERSION:-"3.6.99-amd64"}
+TAG="${IMAGE}:${VERSION}"
```
ahrtr, line 45: User must explicitly provide a version, otherwise fail the test. 

Suggest to follow the name convention. 

```suggestion
REPOSITORY=${IMAGE:-"gcr.io/etcd-development/etcd"}
TAG=${VERSION}-${ARCH}
IMAGE=${REPOSITORY}:${TAG}
```
```

File: scripts/test_images.sh [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+    exit 1
+fi
+
+GET=$(docker exec "${RUN_NAME}" /usr/local/bin/etcdctl get "$KEY" --print-value-only)
+if [ "${GET}" != "${VALUE}" ]; then
+    echo "Problem with getting foo bar in etcd. Got ${GET}"
+    stopContainer
+    exit 1
+fi
+
```
ahrtr, line 83: Please add test cases to verify the version of etcd, etcdctl and etcdutl.

scripts/test_images.sh
```diff
@@ -0,0 +1,93 @@
...
+# Build only if image is not present
+if [[ "$(docker images -q "${IMAGE}" 2> /dev/null)" == "" ]]; then
+    echo "${TAG} not present locally, building it..."
+    # Build a local image from bin directory
+    if ! docker build -t "${IMAGE}" -f "${DOCKERFILE}" "${BINARYDIR}"; then
+        echo "Docker build unsuccessful. Exit code $?"
+        exit 1
+    fi
+fi
```
ahrtr, line 71: It seems that the $BINARYDIR doesn't exist even the image is available.

```suggestion
# Build only if image is not present
if [[ "$(docker images -q "${IMAGE}" 2> /dev/null)" == "" ]]; then
    echo "${TAG} not present locally, building it..."
    
    # Build if binaries are not present
    if [ -z "$BINARYDIR" ]; then
        run ./scripts/build.sh
    fi
    
    # Build a local image from bin directory
    if ! docker build -t "${IMAGE}" -f "${DOCKERFILE}" "${BINARYDIR}"; then
        echo "Docker build unsuccessful. Exit code $?"
        exit 1
    fi
fi
```

I am thinking to totally remove the above build script for simplicity, we can assume the image is already available (fail the test if not present).

Makefile
```diff
@@ -37,6 +37,10 @@ test-e2e-release: build
 test-robustness:
 	PASSES="robustness" ./scripts/test.sh $(GO_TEST_FLAGS)
 
+.PHONY: test-images
+test-images:
+	./scripts/test_images.sh
+
```
ahrtr, line 43:
```suggestion
```

scripts/test_images.sh
pchan: I initially created the script so that it can work in a stand-alone manner, this is what is followed in other scripts like build-docker.sh [1], so my vote is to keep the above changes. 


[1]
scripts/build-docker.sh

> \#If binarydirectory is not specified download tarballs and extract...
> if [ -z "${BINARYDIR}" ]; then
>   RELEASE="etcd-${1}"-$(go env GOOS)-$(go env GOARCH)
>   BINARYDIR="${RELEASE}"
>   TARFILE="${RELEASE}.tar.gz"
>   TARURL="https://github.com/etcd-io/etcd/releases/download/${1}/${TARFILE}"
>   if ! curl -f -L -o "${TARFILE}" "${TARURL}" ; then
>     echo "Failed to download ${TARURL}."
>     exit 1
>   fi
>   tar -zvxf "${TARFILE}"
> fi

chaochn47:
Please rebase and squash commits into one and sign off the commit 

`git rebase HEAD~6 --signoff`

scripts/test_images.sh
```diff
@@ -0,0 +1,94 @@
...
+    # run docker in the background
+    docker run -d --rm --name "${RUN_NAME}" "${IMAGE}" &
+
+    # wait for etcd daemon to bootstrap
+    sleep 5
+}
+
+function stopContainer {
+    docker stop "${RUN_NAME}"
+    docker image rm "${IMAGE}"
```
ahrtr, line 23: Why remove the image? 
```suggestion
```

ahrtr:
> Please rebase and squash commits into one and sign off the commit
> 
> `git rebase HEAD~6 --signoff`

Please also squash the commits as well.

pchan:
I messed up the git history on this one. Continuing the PR here https://github.com/etcd-io/etcd/pull/15505

PR #15356, opened 2023-02-24
Ensure we are using an up to date govulncheck

### LINKED ISSUE #15355: Should we version pin or use latest for govulncheck action?

We have a github action to check for known vulnerabilities that affect Go code `.github/workflows/govuln.yaml`.

Currently this is version pinned to a specific release in late last year as it looks like at that time there was no `@latest`. Refer: https://github.com/etcd-io/etcd/commit/808099dc2483803ea1cf881b6054a26315431f41

There have been some recent releases for `govulncheck` and it's not something that we have dependabot etc picking up. Given this is security related I think we should try and stay up to date so should we revert to using `latest` or should I just raise a PR to maintain the version pin?

Alternatively if someone knows of an easy way to get dependabot or something similar to pick it up please outline below.

### Why is this needed?

Keeping up to date with static security analysis for project go code.

jmhbnz:
Historically we version pinned `govulncheck` because `@latest` wasn't working, refer https://github.com/etcd-io/etcd/commit/808099dc2483803ea1cf881b6054a26315431f41.  This appears to now be resolved as `go install golang.org/x/vuln/cmd/govulncheck@latest` works for me locally. 

Suggest we shift back to using `@latest` or at least update version pin.

Fixes #15355


ahrtr (approved):
LGTM

Thanks @jmhbnz

PR #15380, opened 2023-03-01
The node fails to restart after being removed

wind-c:
If a node is removed via confChangeC, the removal will be replayed when recovering from the log on restart, resulting in a restart failure.

```
 cc := raftpb.ConfChange{ 
     Type:   raftpb.ConfChangeRemoveNode, 
     NodeID: nodeId, 
 } 
 r.confChangeC <- cc
```

```
case raftpb.ConfChangeRemoveNode:
    if cc.NodeID == r.id {
        logger.Info().Uint64("node", r.id).Msg("[raft] node is removed from the cluster. shutting down.")
	return nil, false
    }
    r.transport.RemovePeer(types.ID(cc.NodeID))
}
```

### DIFF
--- a/contrib/raftexample/raft.go
+++ b/contrib/raftexample/raft.go
@@ -174,10 +174,6 @@ func (rc *raftNode) publishEntries(ents []raftpb.Entry) (<-chan struct{}, bool) 
 					rc.transport.AddPeer(types.ID(cc.NodeID), []string{string(cc.Context)})
 				}
 			case raftpb.ConfChangeRemoveNode:
-				if cc.NodeID == uint64(rc.id) {
-					log.Println("I've been removed from the cluster! Shutting down.")
-					return nil, false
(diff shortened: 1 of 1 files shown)


jmhbnz:
Hey @wind-c - Thanks for raising this, can you please sign your commit to ensure the DCO check passes and ensure your contribution will be able to be accepted. 

If you click `Details` on the right hand side of the DCO workflow line you will see instructions for how to do this.

ahrtr:
It's expected behavior. Removed member can't join the cluster with previous data. If you want to add the removed member back, you need to remove the data directory and then follow https://etcd.io/docs/v3.5/op-guide/runtime-configuration/#add-a-new-member

ws-weisb:
Deleting the data on the deleted machine is useless. All machine configurations need to be removed. Is this expected?

Issue #15402, opened 2023-03-03
Etcd watch stream starvation under high read response load when sharing same connection and TLS is enabled


serathius:
### What happened?

When etcd client is generating high read response load, it can result in watch response stream in the same connection being starved. For example a client with an open watch and running 10 concurrent Range requests each returning around 10MB. The watch might get starved and not get any response for tens of seconds.

Problem does not occur when TLS is not enabled nor when watch is created on separate client/connection.

This affects also K8s (any version) as Kubernetes has one client per resource. Problem will trigger when single K8s resource has a lot of data and there are 10+ concurrent LIST requests for the same resource send to apiserver. For example 10k pods. This can cause serious correctness issues in K8s, like controllers not doing any job as they depend on watch to get updates. For example scheduler not scheduling any pods.

We tested and confirmed that all v3.4+ versions are affected. 

Issue affects any watch response type:
* Event
* Periodic watch progress notification
* Manual watch progress notification 

### What did you expect to happen?

Watch stream should never get blocked no matter the load.

### How can we reproduce it (as minimally and precisely as possible)?

Repro for progress notify:

Start etcd with TLS and progress notify, for example:
```
./bin/etcd   --cert-file ./tests/fixtures/server2.crt --key-file ./tests/fixtures/server2.key.insecure --listen-client-urls https://127.0.0.1:2379 --advertise-client-urls https://127.0.0.1:2379 --experimental-watch-progress-notify-interval=100ms
```
insert a lot of data into etcd, for example run command below couple of times.
```
go run ./tools/benchmark/main.go  put --total 10000 --val-size=10000 --key-space-size 10000 --key ./tests/fixtures/server.key.insecure --cacert ./tests/fixtures/ca.crt --cert ./tests/fixtures/server.crt
```

Run program below
```
package main

import (
	"context"
	"crypto/tls"
	"fmt"
	"time"

	clientv3 "go.etcd.io/etcd/client/v3"
)

func main() {
	c, err := clientv3.New(clientv3.Config{
		Endpoints: []string{"127.0.0.1:2379"},
		TLS:       &tls.Config{InsecureSkipVerify: true},
	})
	if err != nil {
		panic(err)
	}
	defer c.Close()
	for i := 0; i < 10; i++ {
		go func() {
			for {
				start := time.Now()
				resp, err := c.Get(context.Background(), "", clientv3.WithPrefix())
				if err != nil {
					fmt.Printf("Error: %v\n", err)
				}
[...]

PR #12370, opened 2020-10-07
embed: Add `insecure-health-endpoint` flag to enable Kubernetes HTTP probes

ironcladlou:
Before this patch, Kubernetes users deploying etcd who configure mTLS
authentication for client and metrics endpoints are unable to leverage etcd's
`/health` endpoint for liveness or readiness probes because Kubernetes HTTP
probes don't support mTLS. Users must work around this limitation with
strategies like weak health checks using custom `exec` probes. Bypassing the
`/health` endpoint for this purpose can lead to false positive probe results and
quorum loss if a member is falsely determined to be ready during a rollout.

This patch introduces an `--insecure-health-endpoint` flag which, if specified,
enables an insecure HTTP `/health` endpoint which is functionally equivalent to
other `/health` endpoints. This enables etcd pods to specify reliable HTTP GET
probes for readiness and liveness checking.

This approach stands as an alternative to introducing a parallel TLS
configuration and listener just for the `/health` endpoint. The typical use case
is probably a non-exposed port used only by a container orchestration component
(e.g. the Kubelet).

Fixes https://github.com/etcd-io/etcd/issues/11993.

Issue #11993, opened 2020-06-09
The health endpoint should have its own listener TLS configuration


ironcladlou:
# Problem

It seems there is no way to simultaneously support mTLS authenticated /metrics and HTTP/S (not mTLS) /health endpoints.

The /health endpoint [shares a listener](https://github.com/etcd-io/etcd/blob/master/embed/etcd.go#L668) with the /metrics endpoint. Using the `-–client-cert-auth` option thus subjects both endpoints to mTLS authentication. However, the use cases for metrics and health probing could be quite different. Metrics might tend to be secured, while health endpoints may be subject to less strict authentication requirements.

The [health check docs](https://etcd.io/docs/v3.4.0/op-guide/monitoring/#health-check) seem to acknowledge this problem area, but in a way that's slightly confusing to me:
>Since v3.3.0, in addition to responding to the /metrics endpoint, any locations specified by --listen-metrics-urls will also respond to the /health endpoint. This can be useful if the standard endpoint is configured with mutual (client) TLS authentication, but a load balancer or monitoring service still needs access to the health check.

The health endpoint seems like the practical choice for a Kubernetes HTTP health probe, but the probe APIs don't support mTLS.

This leaves users left implementing health checks through something like `exec` probes, which can add unnecessary complexity and [inefficiency](https://github.com/kubernetes/kubernetes/issues/82440).

# Solutions

One solution might be to decouple the health and metrics endpoint listeners. Then health authentication must be addressed in some way to provide the needed flexibility. For example:

* For the health listener, honor global the global TLS configuration by default (for compatibility with today's behavior), but provide a new flag to allow control over health endpoint client certificate auth (e.g. `--health-client-cert-auth=false`).
* For the health listener, introduce a complete parallel set of TLS configurations to allow granular control of the health endpoint independent of metrics.

# Notes

Of course, if I've missed something and my core claim is invalid:
>It seems there is no way to simultaneously support mTLS authenticated /metrics and HTTP/S (not mTLS) /health endpoints.

I'd love to understand how to implement this without any changes.

wenjiaswe:
> It seems there is no way to simultaneously support mTLS authenticated /metrics and HTTP/S (not mTLS) /health endpoints.

I think you are right, like you said "The /health endpoint shares a listener with the /metrics endpoint."

> One solution might be to decouple the health and metrics endpoint listeners. 

I think that makes sense. But I would like to defer to @gyuho @xiang90 @jingyih  for opinion.

[...]

PR #15403, opened 2023-03-03
Fix watch stream starvation

serathius:
Ref https://github.com/etcd-io/etcd/issues/15402

Cleaning up serving code and moving grpc server from under http server. Now multiplexing between grpc API and http API will be the same for both TLS and non-TLS config. It will be done by checking HTTP1 (http) vs HTTP2 (grpc).

cc @ahrtr @ptabor

### DIFF
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -28,7 +28,7 @@ import (
 
 	etcdservergw "go.etcd.io/etcd/api/v3/etcdserverpb/gw"
 	"go.etcd.io/etcd/client/pkg/v3/transport"
-	"go.etcd.io/etcd/client/v3/credentials"
+	clientcreds "go.etcd.io/etcd/client/v3/credentials"
 	"go.etcd.io/etcd/pkg/v3/debugutil"
 	"go.etcd.io/etcd/pkg/v3/httputil"
 	"go.etcd.io/etcd/server/v3/config"
@@ -49,6 +49,7 @@ import (
 	"golang.org/x/net/http2"
 	"golang.org/x/net/trace"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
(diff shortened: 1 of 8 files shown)


serathius:
This change breaks curl test :( 
```
curl: (92) HTTP/2 stream 1 was not closed cleanly before end of the underlying stream
```
Marking as draft for debug

tests/common/grpc_test.go
```diff
@@ -0,0 +1,236 @@
...
+	g.Go(func() error {
+		lastSize := size
+		for range time.Tick(time.Second) {
+			select {
+			case <-ctx.Done():
+				return nil
+			default:
+			}
+			mux.RLock()
+			t.Logf("Generating read load around %.1f MB/s", float64(size-lastSize)/1000/1000)
```
ptabor, line 215: probably in this statistics would be nice to write how many calls succeeded/deadlined/failed.

tests/common/grpc_test.go
```diff
@@ -0,0 +1,236 @@
...
+	if sinceLast > maxDelay && sinceLast > watchResponsePeriod+maxWatchDelay {
+		t.Errorf("Unexpected watch response delayed over allowed threshold %s, delay: unknown", maxWatchDelay)
+		t.Errorf("Test finished while in middle of delayed response, measured delay: %s", sinceLast-watchResponsePeriod)
+		t.Logf("Please increase the test duration to measure delay")
+	} else {
+		t.Logf("Max delay: %s", maxDelay-watchResponsePeriod)
+	}
+}
+
+func fillEtcdWithData(ctx context.Context, c *clientv3.Client, keyCount int, valueSize uint) error {
```
ptabor, line 166: This sounds like a shared method in the testing framework.

tests/common/grpc_test.go
```diff
@@ -0,0 +1,236 @@
...
+							return err
+						}
+					}
+					time.Sleep(watchResponsePeriod)
+				}
+			})
+			continuouslyExecuteGetAll(ctx, t, &g, c)
+			validateWatchDelay(t, c.Watch(ctx, "key"))
+			err := g.Wait()
+			if err != nil {
```
ptabor, line 134: require.NoError(g.Wait()) ?

server/embed/serve.go
```diff
@@ -109,113 +111,78 @@ func (sctx *serveCtx) serve(
...
-		srv := &http.Server{
-			Handler:   createAccessController(sctx.lg, s, httpmux),
-			TLSConfig: tlscfg,
-			ErrorLog:  logger, // do not log user error
-		}
-		if err := configureHttpServer(srv, s.Cfg); err != nil {
-			sctx.lg.Error("Configure https server failed", zap.Error(err))
+	listener := sctx.l
+	if sctx.secure {
+		listener, err = transport.NewTLSListener(listener, tlsinfo)
```
ptabor, line 171: Are we covering here /healthz and /metrics endpoints with mTLS that used to be unprotected previously ?

https://github.com/etcd-io/etcd/issues/11993
https://github.com/etcd-io/etcd/pull/12370#issuecomment-714676715

tests/common/grpc_test.go
serathius: All calls succeed, any failure would mark test as failed.

tests/framework/e2e/e2e.go
```diff
@@ -51,6 +51,7 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, opts ...config.
 		WithStrictReconfigCheck(cfg.StrictReconfigCheck),
 		WithAuthTokenOpts(cfg.AuthToken),
 		WithSnapshotCount(cfg.SnapshotCount),
+		WithWatchProcessNotifyInterval(cfg.WatchProgressNotifyInterval),
```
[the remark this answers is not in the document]
serathius: Yes, this change comes with 3 dedicated tests for all types of watch response.

server/embed/serve.go
```diff
@@ -109,113 +111,78 @@ func (sctx *serveCtx) serve(
...
+		if err != nil {
 			return err
 		}
-		go func() { errHandler(srv.Serve(tlsl)) }()
+	}
+	m := cmux.New(listener)
+	go func() { errHandler(srv.Serve(m.Match(cmux.HTTP1()))) }()
+	go func() { errHandler(gs.Serve(m.Match(cmux.HTTP2()))) }()
```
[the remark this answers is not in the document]
serathius: I think what you suggest will be better, however we need to validate if it breaks non-TLS in any way.

tests/common/grpc_test.go
serathius: Yep, however let's keep refactor only when needed.

tests/common/grpc_test.go
serathius: Done.

server/embed/serve.go
serathius: Don't think so. There is no healthz or metrics handler that I'm touching there. The only existing handler in serve was under TLS and will be under TLS.

serathius:
Closing for https://github.com/etcd-io/etcd/pull/15431

Issue #15352, opened 2023-02-24
Flacky test TestLeasingDeleteRangeContendTxn


chaochn47:
### Which github workflows are flaking?

Failed in the forked repo github workflow https://github.com/chaochn47/etcd/actions/runs/4258007188/jobs/7408759584

The flacky test has been mentioned in the PR comment https://github.com/etcd-io/etcd/pull/14918#issuecomment-1344137224 and it still can be reproduced now. 

### Which tests are flaking?

TestLeasingDeleteRangeContendTxn

### Anything else we need to know?

Reproduced in the local box

```
cd tests

taskset -c 1 go test -v -failfast -count=100 -run TestLeasingDeleteRangeContendTxn  ./integration/clientv3/lease
```

serathius:
Thanks for report and repro!

tjungblu:
> leasing_test.go:1313: #0: expected [key:"key/0" create_revision:1120 mod_revision:1121 version:2 value:"123" ], got [key:"key/0" create_revision:1120 mod_revision:1120 version:1 value:"123" ]

just adding another error message here, in the test linked above the result was empty [].

PR #15425, opened 2023-03-07
tests/integration: Update TestLeasingDeleteRangeContendTxn

fuweid:
The TestLeasingDeleteRangeContendTxn is trying to test for RangeDelete when the target resources are being updated.  When the `txnLeasing` wants a server-side transaction, it needs to ensure all the keys mod revision should be leass than what it saw. If the compare fails, it will repeat to apply the server-side transaction until it is sucessful. I believe the test-case is trying to verify how the `txnLeasing` handles the race issue.

Before the patch #15401, the resource-updating goroutine keeps updating until the RangeDelete finishes. The testcase is flaky because two goroutines are sharing one `ctx` and grpc-go client won't wait for the response if `ctx` has been canceled.

For example,

| DelLease Goroutine   | PutLease Goroutine         | ETCD Server                    | Key/0 Status |
| --                   | ---                        | --                             | --           |
| deleted              |                            |                                | version = 0  |
|                      | send update(key/0=123) req | received update(key/0=123) req | version = 0  |
| cancel               |                            |                                | version = 0  |
|                      | exit because of cancel     |                                | version = 0  |
| get key/0 by putkv   |                            |                                | version = 0  |
|                      |                            | applied update(key/0=123)      | version = 1  |
| get key/0 by raw-cli |                            |                                | version = 1  |

So `raw-cli` gets `[key/0=123]` while the `putkv` gets `[]`. If `putkv` applies two update reqs to ETCD server and the last one is canceled before apply, the error will be like:

```
expected [key:"key/0" version:2 value:"123" ], got [key:"key/0" version:1 value:"123" ]
```

The resource-updating goroutine should not share the ctx with RangeDelete here. And I also revert current main branch because the resource-update goroutine only updates 8 times and might exit before `RangeDelete`. In this case, the `txnLeasing` is not handling the race issue.

Fixes: #15352

-----

cc @ahrtr @chaochn47 @serathius @tjungblu

### DIFF
--- a/tests/integration/clientv3/lease/leasing_test.go
+++ b/tests/integration/clientv3/lease/leasing_test.go
@@ -16,7 +16,6 @@ package lease_test
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math/rand"
 	"reflect"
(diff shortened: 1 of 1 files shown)


tjungblu:
ahh it's the context, not the counter. Great find @fuweid!

ahrtr (approved):
Good catch, thank you @fuweid 

I really like such fix for flaky test.

PR #9817, opened 2018-06-07
*: support creating a user without password

mitake:
This commit adds a feature for creating a user without password. The
purpose of the feature is reducing attack surface by configuring bad
passwords (CN based auth will be allowed for the user).

The feature can be used with `--no-password` of `etcdctl user add`
command.

Fix https://github.com/coreos/etcd/issues/9590

PR #15432, opened 2023-03-09
server/auth: fix auth panic bug when user changes password

tangcong:
## What happened?

When the etcd cluster is migrated from the old version before 3.5 to the version 3.5+, if user changes the password will trigger a panic bug.

image

## Reason

Old etcd version user's option is nil.  I checked the code and found that user.Options.NoPassword is always false. It seems that there is no need to check here, and it is not allowed to set an empty password.

### DIFF
--- a/server/auth/store.go
+++ b/server/auth/store.go
@@ -493,7 +493,8 @@ func (as *authStore) UserChangePassword(r *pb.AuthUserChangePasswordRequest) (*p
 	var password []byte
 	var err error
 
-	if !user.Options.NoPassword {
+	// Backward compatible with old versions of etcd, user options is nil
+	if user.Options == nil || !user.Options.NoPassword {
 		password, err = as.selectPassword(r.Password, r.HashedPassword)
 		if err != nil {
 			return nil, ErrNoPasswordUser
--- a/server/auth/store_test.go
+++ b/server/auth/store_test.go
@@ -114,10 +114,26 @@ func setupAuthStore(t *testing.T) (store *authStore, teardownfunc func(t *testin
 		t.Fatal(err)
 	}
 
+	// The UserAdd function cannot generate old etcd version user data (user's option is nil)
+	// add special users through the underlying interface
+	addUserWithNoOption(as)
+
 	tearDown := func(_ *testing.T) {
 		as.Close()
 	}
 	return as, tearDown
+}
+
+func addUserWithNoOption(as *authStore) {
+	tx := as.be.BatchTx()
+	tx.Lock()
+	defer tx.Unlock()
+	tx.UnsafePutUser(&authpb.User{
+		Name:     []byte("foo-no-user-options"),
+		Password: []byte("bar"),
+	})
+	as.commitRevision(tx)
+	as.refreshRangePermCache(tx)
 }
 
 func enableAuthAndCreateRoot(as *authStore) error {
@@ -191,8 +207,8 @@ func TestRecoverWithEmptyRangePermCache(t *testing.T) {
 		t.Fatalf("expected auth enabled got disabled")
 	}
 
-	if len(as.rangePermCache) != 2 {
-		t.Fatalf("rangePermCache should have permission information for 2 users (\"root\" and \"foo\"), but has %d information", len(as.rangePermCache))
+	if len(as.rangePermCache) != 3 {
+		t.Fatalf("rangePermCache should have permission information for 3 users (\"root\" and \"foo\",\"foo-no-user-options\"), but has %d information", len(as.rangePermCache))
 	}
 	if _, ok := as.rangePermCache["root"]; !ok {
 		t.Fatal("user \"root\" should be created by setupAuthStore() but doesn't exist in rangePermCache")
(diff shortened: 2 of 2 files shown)


server/auth/store.go
```diff
@@ -493,11 +493,9 @@ func (as *authStore) UserChangePassword(r *pb.AuthUserChangePasswordRequest) (*p
 	var password []byte
 	var err error
 
-	if !user.Options.NoPassword {
```
ahrtr, line 496: why not `user.Options != nil && !user.Options.NoPassword`?

tangcong: yes, at the beginning, I fixed it like this. After checking the code carefully, I found that this if check is unnecessary.

tangcong: > why not `user.Options != nil && !user.Options.NoPassword`?

In this way, changing the password will not succeed, and the final passwd is an empty password

ahrtr: After second thought, I think the correct way should be `user.Options == nil || !user.Options.NoPassword`.  If `user.Options` is `nil`, then it means the etcd is at a version older than https://github.com/etcd-io/etcd/pull/9817; It means each user must have a password.

tangcong: good catch. however, in the UserChangePassword function, it seems that there will never be an empty password.  I chose the current optimization solution based on the simplicity of the code. can you give a question about the current solution?  If --no-passwd is specified when creating user, is it not allowed to change the password?

ahrtr:
Please also add unit tests

cc @mitake

server/auth/store.go
ahrtr: > If --no-passwd is specified when creating user, is it not allowed to change the password?

In such case, the users should use CN based authentication. There should be no way to change password for such user. cc @mitake to double confirm.

tangcong: There is also a question. At present, there is no error to change the password for no password user, it seems succ.  In fact, it is failed. do we need to add an error code here?
```
% etcdctl user add test --no-password
User test created
% etcdctl user passwd test
Password of test:
Type password of test again for confirmation:
Password updated
```

ahrtr: > do we need to add an error code here?

Agreed. Defer to @mitake

mitake: Thanks a lot for fixing the issue @tangcong !

> If --no-passwd is specified when creating user, is it not allowed to change the password?

Yes, that should be the assumed design. Once a user is created, its `authpb.User.Option` doesn't have a way to be changed.

> do we need to add an error code here?

That's also true, the new password will be simply discarded. Showing an error is much better.

tangcong: ok, I will do it in another PR, this PR will solve the panic problem first, and we need to cherry-pick it to 3.5.  PR has updated and added a unit test case. without this PR, the new unit test cases will not pass.
```

--- FAIL: TestUserChangePassword (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x166f6c5]
```

tangcong:
> Please also add unit tests
> 
> cc @mitake

done.

ahrtr (approved):
LGTM

Thank you @tangcong 

cc @mitake

server/auth/store.go
mitake: Understood, yeah it's great if you can open in another PR.

server/auth/store_test.go
```diff
@@ -114,12 +114,28 @@ func setupAuthStore(t *testing.T) (store *authStore, teardownfunc func(t *testin
 		t.Fatal(err)
 	}
 
+	// The UserAdd function cannot generate old etcd version user data (user's option is nil)
+	// add special users through the underlying interface
+	addUserWithNoOption(as)
```
mitake, line 119: Nit: do we need to add a user with no `Options` for all test cases? I feel just adding it in `TestUserChangePassword` might be fine.

server/auth/store.go
```diff
@@ -493,7 +493,7 @@ func (as *authStore) UserChangePassword(r *pb.AuthUserChangePasswordRequest) (*p
...
+	if user.Options == nil || !user.Options.NoPassword {
```
mitake, line 496: Nit: It's great if you can add a comment which describes the reason of backward compatibility.

mitake:
Thanks a lot for finding and fixing the issue @tangcong! I commented for 2 minor points, could you take a look? The overall change LGTM.

server/auth/store_test.go
tangcong: I intend to add to the global test case, generate an old version of data, to avoid similar problems in the future, and at the same time all the current test cases are passed

server/auth/store.go
tangcong: done

serathius (approved):
This shows lack of etcd upgrade testing.

server/auth/store_test.go
mitake: Understood, then it makes sense.

mitake (approved):

mitake:
LGTM, I'm merging. Thanks a lot @tangcong !
@serathius I agree, it's motivating upgrade testing.

ahrtr:
@mitake The PR https://github.com/etcd-io/etcd/pull/9817 is included in release-3.4, but `(*authStore) UserChangePassword() in 3.4` doesn't check `user.Options.NoPassword` at all. I guess it's missed when implementing https://github.com/etcd-io/etcd/pull/9817? 

see https://github.com/etcd-io/etcd/blob/4b91b6d800f197da9b4ebf8390d45a448a7cd2c8/auth/store.go#L518-L569

ahrtr:
Just as we discussed in slack, 3.4 doesn't have this issue because it doesn't check `user.Options.NoPassword` at all, so it will not panic. Let's leave 3.4 as it's.

server/auth/store.go
mitake: @tangcong if you can open additional PR, that's really great. But feel free to assign it to me if you are busy.

PR #8427, opened 2017-08-21
mvcc: sending events after restore

gyuho:
Fixes: #8411

Replace https://github.com/coreos/etcd/pull/8412.

Cherry-pick for today's release.

PR #5464, opened 2016-05-27
mvcc: tighten up watcher cancelation and revision handling

heyitsanthony:
Makes w.cur into w.minrev, the minimum revision for the next update, and
retries cancelation if the watcher isn't found (because it's being processed
by moveVictims).

Fixes: #5459

PR #15439, opened 2023-03-09
mvcc: update minRev when watcher stays synced

lavacat:
Problem: during restore in watchableStore.Restore, synced watchers are moved to unsynced. minRev will be behind since it's not updated when watcher stays synced.

Solution: update minRev

fixes: https://github.com/etcd-io/etcd/issues/15271

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -448,12 +448,15 @@ func (s *watchableStore) notify(rev int64, evs []mvccpb.Event) {
 			pendingEventsGauge.Add(float64(len(eb.evs)))
 		} else {
 			// move slow watcher to victims
-			w.minRev = rev + 1
 			w.victim = true
 			victim[w] = eb
 			s.synced.delete(w)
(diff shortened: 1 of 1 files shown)


lavacat:
@serathius I've tested that this fixes https://github.com/etcd-io/etcd/pull/15273/files
Without the fix, test fails, with the fix it succeeds consistently.

I've marked this PR as draft because I'm not yet fluent with all logic for syncWatchers. I'd like to add an integration test.

For ref:
minRev was added [here](https://github.com/etcd-io/etcd/pull/5464) 

Restore logic was added [here](https://github.com/etcd-io/etcd/pull/8427)

xiang90:
We should probably add a test case to ensure the coverage.

serathius:
cc @logicalhan

serathius:
Any progress?

serathius:
Confirmed that this fixes https://github.com/etcd-io/etcd/issues/15271

serathius:
Created https://github.com/etcd-io/etcd/pull/15492 based on this

Issue #15145, opened 2023-01-18
Remove the dependency on grpc-go's experimental API (e.g. resolver & balancer)


ahrtr:
### What would you like to be added?

etcd is depending on some experimental APIs of grpc-go, such as resolver. Once these APIs change in future, it might break etcd. So we should try to remove the dependency on any experimental APIs; instead, let's try to use stable APIs.

References:
- https://github.com/grpc/grpc-go/pull/5748#issuecomment-1385964654
- https://github.com/grpc/grpc-go/pull/5748#issuecomment-1387426476
- https://github.com/etcd-io/etcd/issues/19700#issuecomment-2779240534

We should have
- a summary on existing design/implementation first
- proposal(s) to replace them, and with pros & cons. invite grpc's experts to review.
- evaluate the impact on etcd itself and users of etcd client sdk.

Mitigations:
- https://github.com/etcd-io/etcd/pull/19780
- https://github.com/etcd-io/etcd/issues/19706
  - https://github.com/etcd-io/etcd/pull/19785
  - https://github.com/etcd-io/etcd/pull/19825
- mark the [gRPC naming and discovery](https://etcd.io/docs/v3.5/dev-guide/grpc_naming/) feature as experimental, see also https://github.com/etcd-io/etcd/issues/19700#issuecomment-2821703505
  - https://github.com/etcd-io/website/issues/990

Long-term solution?
- use the GCP gRPC connection pool ?
  - https://github.com/etcd-io/etcd/issues/19700#issuecomment-2815923497

Related followup:
- An endpoint may have multiple addresses: i.e. IPv4 and IPv6. Currently we only assume each endpoint only has one address
  - https://github.com/etcd-io/etcd/issues/19700#issuecomment-2819173767
  - https://github.com/etcd-io/etcd/pull/19780

- https://github.com/etcd-io/etcd/issues/19772

Summary:
- https://docs.google.com/document/d/14pPPyQkJ8NvPZSM6Pe1Wr0BDHAsw_Xcp0mepoSVTQHs/edit?tab=t.0
### Why is this needed?

Improve stability

dfawley:
grpc-go TL here.  I'm happy to work with you to find / provide a stable solution, but I would first need to know more about your use cases & requirements.

tjungblu:
[...]

PR #8221, opened 2017-07-06
embed: connect json gateway with user-provided listen address

heyitsanthony:
net.Listener says its address is [::] when given 0.0.0.0, breaking
hosts that have ipv6 disabled.

Fixes #8151
Fixes #7961

PR #15446, opened 2023-03-10
Allow user to separate http and grpc server

serathius:
Phase 2 of https://github.com/etcd-io/etcd/issues/15402#issuecomment-1460061395
Adding a flag to separate http server. No changes in default so the PR can be backported.

    Difference in load configuration for watch delay tests show how huge the
    impact is. Even with random write scheduler grpc under http
    server can only handle 500 KB with 2 seconds delay. On the other hand,
    separate grpc server easily hits 10, 100 or even 1000 MB within 100 miliseconds.
    
    Priority write scheduler that was used in most previous releases
    is far worse than random one.
    
    Tests configured to only 10 MB to avoid them taking too long to fill
    etcd.

cc @ahrtr @ptabor @spzala @fuweid @aojea


server/embed/serve.go
```diff
@@ -127,97 +136,137 @@ func (sctx *serveCtx) serve(
...
+			//	if err != nil {
+			//		sctx.lg.Error("registerGateway failed", zap.Error(err))
+			//		return err
+			//	}
+			//}
+		}
+		var srvhttp *http.Server
+		if onlyGRPC {
+			server = func() error {
+				return gs.Serve(sctx.l)
```
fuweid, line 158: Nice to see this!

ahrtr:
I believe adding a new user facing flag "`--listen-client-http-urls`" can resolve https://github.com/etcd-io/etcd/issues/15402, and I think it's a safe. But it definitely changes user behavior. All users have adapted the behavior of 2379 for all user connection/request, and 2380 for peer requests (and probably 2381 for metrics). Now users have to add add a new port (e.g. 2382) for all http client requests (e.g. /health, /metrics, etc.). **I think it's a non-trivial change, and we should get more feedback before we start to make change**.

From users perspective, I think it still makes sense to multiplex both gRPC and HTTP (including 1 and 2) using one port (`2379`), and personally I think it should still the long-term goal. 

gRPC-go has always been a pain point in etcd, which is still depending on some experimental APIs right now (see https://github.com/etcd-io/etcd/issues/15145), one of which is [grpcServer.ServeHTTP](https://github.com/etcd-io/etcd/blob/e2a5df534cf6b0a12859e61bfc2758d56a1e1971/server/embed/serve.go#L237).

I think it would be great if this issue can be resolved by network layer instead of etcd? Or we follow this PR and eventually revert to the behavior of multiplexing gRPC and HTTP in one port in the future? We (most etcd contributors) are not network (e.g. gRPC, golang.org/x/net, etc) experts, so it would be great if we can get more feedback from network experts. cc @dfawley @aojea

ahrtr:
cc K8s folks @liggitt @dims @nikhita @neolit123 @wojtek-t

serathius:
Please let's move discussion to the original issue https://github.com/etcd-io/etcd/issues/15402 where this design was approved by @ptab, @mborsz and @aojea. 

The change is non-trivial, but everyone I talked to agreed that this is the correct long-term direction. GRPC and http should not be mixed, @aojea even called it "an abomination".

aojea:
> GRPC and http should not be mixed, @aojea even called it "an abomination".

well, I just was trying to add some drama 😄 , but you can see similar comments from its own author https://github.com/grpc/grpc-go/issues/586#issuecomment-286261423 

> And I don't officially work on grpc-go. It's not surprising that ServeHTTP is slower considering it's shoehorned into the existing ServerTransport interface somewhat awkwardly.


and on the documentation it is still considered Experimental 

https://pkg.go.dev/google.golang.org/grpc#Server.ServeHTTP

ptabor:
@ahrtr - I agree that all changes requiring action from customers are difficult and should go through a deep consideration. 

I think that this PR is not yet controversial and not introducing the breaking change. It's just adding new
(safe) mode of operation. 
The breaking change (changing default) will be a follow up PR... that will not be backported.  Users will have both v3.4 and v3.5 to gracefully migrated to the new safer semantic and only in v3.6 (or later) we will change the default. 
Etcd minor releases are so in frequent that we do use it as a train to ship breaking changes when they are justified by project's robustness or security.

server/embed/config.go
```diff
@@ -211,12 +211,12 @@ type Config struct {
 	// streams that each client can open at a time.
 	MaxConcurrentStreams uint32 `json:"max-concurrent-streams"`
 
-	ListenPeerUrls, ListenClientUrls       []url.URL
-	AdvertisePeerUrls, AdvertiseClientUrls []url.URL
-	ClientTLSInfo                          transport.TLSInfo
-	ClientAutoTLS                          bool
-	PeerTLSInfo                            transport.TLSInfo
-	PeerAutoTLS                            bool
+	ListenPeerUrls, ListenClientUrls, ListenClientHttpUrls []url.URL
```
ptabor, line 214: As mentioned during today's discussions. 
Maybe to avoid changing fields semantic, we should rather be very explicit here, i.e. introduce
```
  --listen_client_http_urls  -> that servers only http handlers
  --listen_client_grpc_urls -> that servers only grpc 
```
And `listen_client_urls` stays as the cmux'ed variant of both that is discouraged...

In v3.6 we will make a call based on cmux and golang http server health whether we 
swap the defaults and default `listen_client_urls` to '', while `listen_client_grpc_urls` gets the original endpoint. 

But thanks to introduction of 2 flags: 
 a) We still have an option to be fully backward compatible in v3.6
 b) We minimize the backported change in v3.4 and v3.5.

serathius: I have worried about dedicated `--listen-client-grpc-urls` flag. Reason is that making it work like `--listen-client-urls` is much more complicated then what is done here. Please remember that `--listen-client-urls` is used in cluster setup, like in `advertise-client-urls`. I think that would be too big change to backport. 

Just adding `--listen-client-http-urls` is much simpler as it doesn't make any changes in etcd API setup (+/- grpcgateway).

serathius:
Hmm cmux test failure ?!
```
2023-03-16T16:56:35.0274754Z --- FAIL: TestConnectionMultiplexing (13.68s)
2023-03-16T16:56:35.0275287Z     before.go:36: Changing working directory to: /tmp/TestConnectionMultiplexing1002003079/001
2023-03-16T16:56:35.0275898Z     --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS (11.68s)
2023-03-16T16:56:35.0276887Z         logger.go:130: 2023-03-16T16:46:10.486Z	INFO	starting server...	{"name": "TestConnectionMultiplexingServerTLSAndNonTLS-test-0"}
2023-03-16T16:56:35.0284372Z         logger.go:130: 2023-03-16T16:46:10.487Z	INFO	spawning process	{"args": ["/home/runner/work/etcd/etcd/bin/etcd", "--name", "TestConnectionMultiplexingServerTLSAndNonTLS-test-0", "--listen-client-urls", "http://localhost:20000,https://localhost:20000", "--advertise-client-urls", "http://localhost:20000,https://localhost:20000", "--listen-peer-urls", "http://localhost:20001", "--initial-advertise-peer-urls", "http://localhost:20001", "--initial-cluster-token", "", "--data-dir", "/tmp/TestConnectionMultiplexingServerTLSAndNonTLS4257936873/001", "--snapshot-count", "10000", "--strict-reconfig-check=false", "--cert-file", "/home/runner/work/etcd/etcd/tests/fixtures/server.crt", "--key-file", "/home/runner/work/etcd/etcd/tests/fixtures/server.key.insecure", "--trusted-ca-file", "/home/runner/work/etcd/etcd/tests/fixtures/ca.crt", "--initial-cluster", "TestConnectionMultiplexingServerTLSAndNonTLS-test-0=http://localhost:20001", "--initial-cluster-state", "new"], "working-dir": "/tmp/TestConnectionMultiplexing1002003079/001", "name": "TestConnectionMultiplexingServerTLSAndNonTLS-test-0", "environment-variables": ["ETCD_VERIFY=all", "EXPECT_DEBUG=true", "PATH=/opt/hostedtoolcache/go/1.19.7/x64/bin:/home/runner/go/bin:/opt/hostedtoolcache/go/1.19.7/x64/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin", "ETCD_UNSUPPORTED_ARCH=386"]}
2023-03-16T16:56:35.0287532Z         logger.go:130: 2023-03-16T16:46:11.214Z	INFO	started server.	{"name": "TestConnectionMultiplexingServerTLSAndNonTLS-test-0", "pid": 10906}
2023-03-16T16:56:35.0288314Z         --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS (10.24s)
2023-03-16T16:56:35.0289249Z             --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS/etcdctl (5.02s)
2023-03-16T16:56:35.0289801Z                 cmux_test.go:101: 
2023-03-16T16:56:35.0290604Z                     	Error Trace:	/home/runner/work/etcd/etcd/tests/e2e/cmux_test.go:101
2023-03-16T16:56:35.0291609Z                     	Error:      	Received unexpected error:
2023-03-16T16:56:35.0294285Z                     	            	match not found.  Set EXPECT_DEBUG for more info Errs: [unexpected exit code [1] after running [/home/runner/work/etcd/etcd/bin/etcdctl --endpoints=http://localhost:20000 get a -w json -w json -w json]], last lines:
2023-03-16T16:56:35.0300027Z                     	            	{"level":"warn","ts":"2023-03-16T16:46:16.930432Z","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xaa00700/localhost:20000","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"error reading server preface: read tcp 127.0.0.1:39924->127.0.0.1:20000: read: connection reset by peer\""}
2023-03-16T16:56:35.0301247Z                     	            	Error: context deadline exceeded
2023-03-16T16:56:35.0302107Z                     	Test:       	TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS/etcdctl
2023-03-16T16:56:35.0302942Z             --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS/clientv3 (5.00s)
2023-03-16T16:56:35.0303569Z                 utils.go:53: context deadline exceeded
```

fuweid:
> Hmm cmux test failure ?!
> 
> ```
> 
> 2023-03-16T16:56:35.0274754Z --- FAIL: TestConnectionMultiplexing (13.68s)
> 
> 2023-03-16T16:56:35.0275287Z     before.go:36: Changing working directory to: /tmp/TestConnectionMultiplexing1002003079/001
> 
> 2023-03-16T16:56:35.0275898Z     --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS (11.68s)
> 
> 2023-03-16T16:56:35.0276887Z         logger.go:130: 2023-03-16T16:46:10.486Z	INFO	starting server...	{"name": "TestConnectionMultiplexingServerTLSAndNonTLS-test-0"}
> 
> 2023-03-16T16:56:35.0284372Z         logger.go:130: 2023-03-16T16:46:10.487Z	INFO	spawning process	{"args": ["/home/runner/work/etcd/etcd/bin/etcd", "--name", "TestConnectionMultiplexingServerTLSAndNonTLS-test-0", "--listen-client-urls", "http://localhost:20000,https://localhost:20000", "--advertise-client-urls", "http://localhost:20000,https://localhost:20000", "--listen-peer-urls", "http://localhost:20001", "--initial-advertise-peer-urls", "http://localhost:20001", "--initial-cluster-token", "", "--data-dir", "/tmp/TestConnectionMultiplexingServerTLSAndNonTLS4257936873/001", "--snapshot-count", "10000", "--strict-reconfig-check=false", "--cert-file", "/home/runner/work/etcd/etcd/tests/fixtures/server.crt", "--key-file", "/home/runner/work/etcd/etcd/tests/fixtures/server.key.insecure", "--trusted-ca-file", "/home/runner/work/etcd/etcd/tests/fixtures/ca.crt", "--initial-cluster", "TestConnectionMultiplexingServerTLSAndNonTLS-test-0=http://localhost:20001", "--initial-cluster-state", "new"], "working-dir": "/tmp/TestConnectionMultiplexing1002003079/001", "name": "TestConnectionMultiplexingServerTLSAndNonTLS-test-0", "environment-variables": ["ETCD_VERIFY=all", "EXPECT_DEBUG=true", "PATH=/opt/hostedtoolcache/go/1.19.7/x64/bin:/home/runner/go/bin:/opt/hostedtoolcache/go/1.19.7/x64/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin", "ETCD_UNSUPPORTED_ARCH=386"]}
> 
> 2023-03-16T16:56:35.0287532Z         logger.go:130: 2023-03-16T16:46:11.214Z	INFO	started server.	{"name": "TestConnectionMultiplexingServerTLSAndNonTLS-test-0", "pid": 10906}
> 
> 2023-03-16T16:56:35.0288314Z         --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS (10.24s)
> 
> 2023-03-16T16:56:35.0289249Z             --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS/etcdctl (5.02s)
> 
> 2023-03-16T16:56:35.0289801Z                 cmux_test.go:101: 
> 
> 2023-03-16T16:56:35.0290604Z                     	Error Trace:	/home/runner/work/etcd/etcd/tests/e2e/cmux_test.go:101
> 
> 2023-03-16T16:56:35.0291609Z                     	Error:      	Received unexpected error:
> 
> 2023-03-16T16:56:35.0294285Z                     	            	match not found.  Set EXPECT_DEBUG for more info Errs: [unexpected exit code [1] after running [/home/runner/work/etcd/etcd/bin/etcdctl --endpoints=http://localhost:20000 get a -w json -w json -w json]], last lines:
> 
> 2023-03-16T16:56:35.0300027Z                     	            	{"level":"warn","ts":"2023-03-16T16:46:16.930432Z","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xaa00700/localhost:20000","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = \"error reading server preface: read tcp 127.0.0.1:39924->127.0.0.1:20000: read: connection reset by peer\""}
> 
> 2023-03-16T16:56:35.0301247Z                     	            	Error: context deadline exceeded
> 
> 2023-03-16T16:56:35.0302107Z                     	Test:       	TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS/etcdctl
> 
> 2023-03-16T16:56:35.0302942Z             --- FAIL: TestConnectionMultiplexing/ServerTLSAndNonTLS/ClientNonTLS/clientv3 (5.00s)
> 
> 2023-03-16T16:56:35.0303569Z                 utils.go:53: context deadline exceeded
> 
> ```
> 
> Can't repro locally.

I think I run into this as well. https://github.com/etcd-io/etcd/actions/runs/4439049893/jobs/7790956875.

serathius:
@fuweid Thanks for confirming. Looks like a flake :( 
Will file an issue to investigate it.

fuweid:
> @fuweid Thanks for confirming. Looks like a flake :( 
> 
> Will file an issue to investigate it.

If no one takes it, please assign it to me. Thanks~ 🤓

serathius:
> > @fuweid Thanks for confirming. Looks like a flake :(
> > Will file an issue to investigate it.
> 
> If no one takes it, please assign it to me. Thanks~ nerd_face

I just added the test :(, so let me take responsibility.

serathius:
Added cmux test for separating http port.

Opening for review cc @ahrtr @ptabor

tests/e2e/utils.go
```diff
@@ -77,15 +77,17 @@ func tlsInfo(t testing.TB, cfg e2e.ClientConfig) (*transport.TLSInfo, error) {
 	}
 }
 
-func fillEtcdWithData(ctx context.Context, c *clientv3.Client, keyCount int, valueSize uint) error {
+func fillEtcdWithData(ctx context.Context, c *clientv3.Client, dbSize int) error {
```
ptabor, line 80: Why this change ? 

For different tests the number of keys and size of values might be more important than overall size. 
If you need a simpler api, I would propose a wrapping function: 

`fillEtcdWithDataTotal(...)`

or 
`fillEtcdWithData(..., dbSize)` calls
`fillEtcdWithKeyValues(..., keyCount, keySize, valueSize)`

tests/e2e/watch_delay_test.go
```diff
@@ -35,29 +35,48 @@ import (
...
+	},
+	{
+		name:          "TLS",
+		config:        e2e.EtcdProcessClusterConfig{ClusterSize: 1, Client: e2e.ClientConfig{ConnectionType: e2e.ClientTLS}},
+		maxWatchDelay: 2 * time.Second,
+		dbSizeBytes:   500 * Kilo,
+	},
+	{
+		name:          "SeparateHttpNoTLS",
+		config:        e2e.EtcdProcessClusterConfig{ClusterSize: 1, Client: e2e.ClientConfig{ConnectionType: e2e.ClientTLS}, ClientHttpSeparate: true},
```
ptabor, line 71: Looks like TLS (rather than noTLS)

server/embed/serve.go
```diff
@@ -95,6 +95,8 @@ func (sctx *serveCtx) serve(
 	tlsinfo *transport.TLSInfo,
 	handler http.Handler,
 	errHandler func(error),
+	grpcDial func(ctx context.Context) (*grpc.ClientConn, error),
```
ptabor, line 98: Please document why we do need grpc function. 
It's very counter intuitive when I looked at the call... and signature...

ptabor: Maybe call it `grpcDialForRestGatewayBackends`

server/etcdmain/config.go
```diff
@@ -115,7 +115,11 @@ func newConfig() *config {
 	)
 	fs.Var(
 		flags.NewUniqueURLsWithExceptions(embed.DefaultListenClientURLs, ""), "listen-client-urls",
-		"List of URLs to listen on for client traffic.",
+		"List of URLs to listen on for client grpc traffic. Http services are available as long --listen-client-http-urls is not specified.",
```
ahrtr, line 118:
```suggestion
		"List of URLs to listen on for client grpc traffic. Http services are available as long as --listen-client-http-urls is not specified.",
```

server/embed/serve.go
```diff
@@ -109,63 +111,87 @@ func (sctx *serveCtx) serve(
 
 	sctx.lg.Info("ready to serve client requests")
 
-	m := cmux.New(sctx.l)
+	var m cmux.CMux
+	var server func() error
+	onlyGRPC := splitHttp && !sctx.httpOnly
+	grpcEnabled := !splitHttp || !sctx.httpOnly
```
ahrtr, line 117: It's a little hard to understand, suggest to change the logic to 
```
onlyHttp := sctx.httpOnly
onlyGRPC := splitHttp && !sctx.httpOnly

if onlyHttp {

} else if onlyGRPC {

} else {
   // support legacy behavior: both http and gRPC share the same conection
}
```

ahrtr:
This PR mixes the code refactoring (e.g. configureClientListener, and grpcGatewayDial) and new feature, can you consider to split the code refactoring into separate PR?

server/etcdmain/config.go
```diff
@@ -115,7 +115,11 @@ func newConfig() *config {
...
+	)
+	fs.Var(
+		flags.NewUniqueURLsWithExceptions("", ""), "listen-client-http-urls",
```
ahrtr, line 121: Do we need to add the corresponding `--advertise-client-http-urls`?

server/embed/config.go
```diff
@@ -688,6 +698,12 @@ func (cfg *Config) Validate() error {
 	if err := checkBindURLs(cfg.ListenClientUrls); err != nil {
 		return err
 	}
+	if err := checkBindURLs(cfg.ListenClientHttpUrls); err != nil {
+		return err
+	}
+	if len(cfg.ListenClientHttpUrls) == 0 {
+		cfg.logger.Warn("TODO")
```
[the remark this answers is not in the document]
serathius: Done

server/embed/etcd.go
```diff
@@ -631,32 +637,68 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
...
 		if u.Scheme == "unix" || u.Scheme == "unixs" {
-			network = "unix"
 			addr = u.Host + u.Path
 		}
+		secure := u.Scheme == "https" || u.Scheme == "unixs"
+		insecure := !secure
+		network := "tcp"
+		if u.Scheme == "unix" || u.Scheme == "unixs" {
+			network = "unix"
+		}
```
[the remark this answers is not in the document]
serathius: Done

server/embed/etcd.go
```diff
@@ -631,32 +637,68 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
...
+
+		sctx := sctxs[addr]
+		if sctx == nil {
```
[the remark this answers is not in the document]
serathius: Yes, we expect that user can list same host twice with different scheme. One for `http`, second for `https`

server/embed/etcd.go
```diff
@@ -631,32 +637,68 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
 		if (u.Scheme == "https" || u.Scheme == "unixs") && cfg.ClientTLSInfo.Empty() {
 			return nil, fmt.Errorf("TLS key/cert (--cert-file, --key-file) must be provided for client url %s with HTTPS scheme", u.String())
 		}
+	}
 
-		network := "tcp"
+	sctxs = make(map[string]*serveCtx)
```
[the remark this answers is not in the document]
serathius: This is the previous name from line 620.

tests/e2e/utils.go
serathius: I have tested it, keys have some impact but not enough to expose it now. I think exposing etcd gives better understanding of intention.

tests/e2e/watch_delay_test.go
serathius: Fixed

server/embed/serve.go
```diff
@@ -109,63 +111,87 @@ func (sctx *serveCtx) serve(
...
+
 	v3c := v3client.New(s)
 	servElection := v3election.NewElectionServer(v3c)
 	servLock := v3lock.NewLockServer(v3c)
 
 	// Make sure serversC is closed even if we prematurely exit the function.
 	defer close(sctx.serversC)
+	var gwmux *gw.ServeMux
+	if s.Cfg.EnableGRPCGateway {
+		gwmux, err = sctx.registerGateway(grpcDial)
```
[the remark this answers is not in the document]
serathius: As discussed offline, left a comment explaining how grpc gateway connects to via grpc dial.

server/etcdmain/config.go
serathius: No, as etcd client and members only communicate via grpc to client port.

serathius:
> This PR mixes the code refactoring (e.g. configureClientListener, and grpcGatewayDial) and new feature, can you consider to split the code refactoring into separate PR?

Normally yes, however I would argue that for backports they should be kept together. I extracted refactors to make review easier, however without them it is impossible to implement the fix.

tests/e2e/utils.go
ptabor: [I assume this file is a generic testing infrastructure and different tests might have a different needs]

server/embed/serve.go
serathius: Will try it, we will see if that's more readable.

tests/e2e/utils.go
serathius: Sure, however let's not design something for future that can be easily changed and might never be useful.

server/embed/etcd.go
```diff
@@ -456,11 +459,16 @@ func (e *Etcd) Close() {
...
+		ss.http.Shutdown(ctx)
+	}
+	if ss.grpc == nil {
+		return
+	}
+	// do not grpc.Server.GracefulStop when grpc runs under http server
 	// See https://github.com/grpc/grpc-go/issues/1384#issuecomment-317124531
 	// and https://github.com/etcd-io/etcd/issues/8916
-	if ss.secure {
+	if ss.secure && ss.http != nil {
```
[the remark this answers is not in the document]
serathius: As above comment states ` do not grpc.Server.GracefulStop when grpc runs under http server`. With this PR, grpc server no longer runs under http server when in httpOnly mode. So if there is no http server, grpc server runs standalone mode.

server/etcdmain/config.go
serathius: Done

serathius:
Addressed all feedback, please take another look @ahrtr @ptabor

tests/e2e/cmux_test.go
```diff
@@ -39,8 +39,9 @@ import (
 func TestConnectionMultiplexing(t *testing.T) {
 	e2e.BeforeTest(t)
 	for _, tc := range []struct {
-		name      string
-		serverTLS e2e.ClientConnType
+		name             string
+		serverTLS        e2e.ClientConnType
+		separateHttpPort bool
```
chaochn47, line 44: Could you point to where is `separateHttpPort` used?

tests/framework/e2e/cluster.go
```diff
@@ -457,22 +458,24 @@ func (cfg *EtcdProcessClusterConfig) SetInitialOrDiscovery(serverCfg *EtcdServer
...
+		curls = []string{curl, clientURL(clientPort, ClientTLS)}
+	} else {
+		curl = clientURL(clientPort, cfg.Client.ConnectionType)
 		curls = []string{curl}
-	case ClientTLSAndNonTLS:
-		curl = (&url.URL{Scheme: "http", Host: curlHost}).String()
-		curltls = (&url.URL{Scheme: "https", Host: curlHost}).String()
-		curls = []string{curl, curltls}
+	}
+	clientHttpUrl := curl
```
chaochn47, line 476: `var clientHttpUrl string`. I think the initial value should be empty.

server/embed/config.go
```diff
@@ -688,6 +698,12 @@ func (cfg *Config) Validate() error {
 	if err := checkBindURLs(cfg.ListenClientUrls); err != nil {
 		return err
 	}
+	if err := checkBindURLs(cfg.ListenClientHttpUrls); err != nil {
+		return err
+	}
+	if len(cfg.ListenClientHttpUrls) == 0 {
+		cfg.logger.Warn("Running http and grpc server on single port. This is not recommended for production.")
```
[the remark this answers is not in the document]
ahrtr: >  for http without tls turned on

Definitely not recommended in production. Also turning on/off TLS should be irrelevant to this PR.

ahrtr:
> I think that this PR is not yet controversial

@ptabor 
That's the concern, because of no counter-question. I still think it makes sense to expose only one port (e.g. 2379) to clients from users perspective. Users shouldn't care about gRPC or HTTP. I am not a network expert, and no extra bandwidth to have a deep dive either. But happy to see @aojea keeps trying https://github.com/etcd-io/etcd/pull/15510, I may spend some time on it in future. 

Anyway, I agree this PR is the best choice for 3.4 and 3.5, and overall looks good. Great work! @serathius . (Please resolve the couple of minor comments from @chaochn47 )

server/etcdmain/config.go
```diff
@@ -115,7 +115,11 @@ func newConfig() *config {
 	)
 	fs.Var(
 		flags.NewUniqueURLsWithExceptions(embed.DefaultListenClientURLs, ""), "listen-client-urls",
-		"List of URLs to listen on for client traffic.",
+		"List of URLs to listen on for client grpc traffic. Http services are available as long as --listen-client-http-urls is not specified.",
```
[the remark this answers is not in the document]
serathius: Done

server/embed/config.go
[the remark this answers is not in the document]
serathius: Agree with @ahrtr

server/embed/etcd.go
```diff
@@ -631,32 +638,48 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
...
-		if sctx.l, err = transport.NewListenerWithOpts(addr, u.Scheme,
+	for _, sctx := range sctxs {
+		if sctx.l, err = transport.NewListenerWithOpts(sctx.addr, sctx.scheme,
 			transport.WithSocketOpts(&cfg.SocketOpts),
 			transport.WithSkipTLSInfoCheck(true),
 		); err != nil {
 			return nil, err
 		}
 		// net.Listener will rewrite ipv4 0.0.0.0 to ipv6 [::], breaking
 		// hosts that disable ipv6. So, use the address given by the user.
```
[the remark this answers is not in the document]
serathius: Not sure what this comment refers to. I didn't move it.

tests/framework/e2e/cluster.go
serathius: Done

server/embed/etcd.go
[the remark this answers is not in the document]
chaochn47: It was introduced by https://github.com/etcd-io/etcd/pull/8221 to fix the grpc gateway creating grpc connections using [::] caused by `sctx.listener.addr()` when ipv6 is disabled but user configures 0.0.0.0 as listen client urls. 

I think we can leave it as it is right now and improve it later.

server/embed/etcd.go
```diff
@@ -690,11 +713,23 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
...
+func resolveUrl(u url.URL) (addr string, secure bool, network string) {
+	addr = u.Host
+	if u.Scheme == "unix" || u.Scheme == "unixs" {
+		addr = u.Host + u.Path
+	}
+	secure = u.Scheme == "https" || u.Scheme == "unixs"
+	network = "tcp"
+	if u.Scheme == "unix" || u.Scheme == "unixs" {
+		network = "unix"
+	}
```
ahrtr, line 729:
```suggestion
	addr = u.Host
	network = "tcp"
	if u.Scheme == "unix" || u.Scheme == "unixs" {
		addr = u.Host + u.Path
		network = "unix"
	}
	secure = u.Scheme == "https" || u.Scheme == "unixs"
```

ahrtr (approved):
LGTM with a minor comment. 

cc @ptabor @chaochn47 @fuweid to double check.

chaochn47:
Just a reminder, to avoid breaking the client that enables syncing, we should suggest users to include the splitted listen client http urls in the advertised client urls. 

https://github.com/etcd-io/etcd/blob/275e10bcf72c03ca0884402db927c9168fc6dc3f/server/etcdmain/help.go#L108-L110

chaochn47 (approved):

server/etcdmain/config.go
ptabor: I assume a scenario when someone lists members, takes 'advertise-client-url` from the output and uses it for metrics gathering / http communication. That is fair to say that the advertised service in case of split, points on the grpc service.

ptabor (approved):
LGTM. Thank you.

server/embed/etcd.go
```diff
@@ -669,17 +692,17 @@ func configureClientListeners(cfg *Config) (sctxs map[string]*serveCtx, err erro
 			sctx.l = transport.LimitListener(sctx.l, int(fdLimit-reservedInternalFDNum))
 		}
 
-		defer func(u url.URL) {
-			if err == nil {
+		defer func(addr string) {
+			if err == nil || sctx.l == nil {
```
fuweid, line 696: I think we should pass *sctx* in the defer function instead of address.

https://go.dev/play/p/ZzmQADcBkJl

fuweid (commented):
Thanks for the patch!

Left one comment for the defer handle, IIUC.

PR #7706, opened 2017-04-11
etcdserver: wait apply on conf change Raft entry

gyuho:
When apply-layer sees configuration change entry in
raft.Ready.CommittedEntries, the server should not proceed
until that entry is applied. Otherwise, follower's raft
layer advances, possibly election-timeouts, and becomes
the leader in single-node cluster, before add-node conf
change of other nodes is applied.

Fix https://github.com/coreos/etcd/issues/7595.

@xiang90 @heyitsanthony 

Could you check if this is the correct approach?
I think we can also remove the `isCandidate` check,
if we had that only for config changes?

Thanks.

Issue #7595, opened 2017-03-24
TestRestartMember: context deadline exceeded


mitake:
via https://jenkins-etcd-public.prod.coreos.systems/job/etcd-proxy/1084/consoleFull in the PR https://github.com/coreos/etcd/pull/7569

```
--- FAIL: TestRestartMember (163.14s)
	cluster_test.go:521: #2: watch on unix://127.0.0.1:2193617511 error: context deadline exceeded
```

Possibly related to https://github.com/coreos/etcd/issues/7258 and https://github.com/coreos/etcd/pull/7379 ?

heyitsanthony:
Might be a raft bug, the logs are getting spammed with leadership change messages:
```
2017-03-24 02:14:23.820073 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3626c71ed93b52c4 to 3554539849c8e8e at term 8
2017-03-24 02:14:23.823444 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3554539849c8e8e to 3626c71ed93b52c4 at term 8
2017-03-24 02:14:23.830189 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3626c71ed93b52c4 to 3554539849c8e8e at term 8
2017-03-24 02:14:23.832424 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3554539849c8e8e to 3626c71ed93b52c4 at term 8
2017-03-24 02:14:23.840226 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3626c71ed93b52c4 to 3554539849c8e8e at term 8
2017-03-24 02:14:23.842899 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3554539849c8e8e to 3626c71ed93b52c4 at term 8
2017-03-24 02:14:23.850610 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3626c71ed93b52c4 to 3554539849c8e8e at term 8
2017-03-24 02:14:23.852937 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3554539849c8e8e to 3626c71ed93b52c4 at term 8
2017-03-24 02:14:23.862708 I | raft: raft.node: b4501d80aad6ac80 changed leader from 3626c71ed93b52c4 to 3554539849c8e8e at term 8
[...]

PR #15450, opened 2023-03-13
raftexample: apply conf change entries before send message

KumKeeHyun:
Update raftexample to apply conf changes before send raft.Ready.Messages. Candidate or follower needs to wait for all pending conf changes. Otherwise there might be incorrectly count votes.
This PR introduces the fix from #7595 and pr num [7706](https://github.com/etcd-io/etcd/pull/7706).

Same logic as [etcdserver/raft.go at main](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/raft.go#L278)
Since conf changes are applied directly in publishEntries, it doesn't need to wait for the applyDoneC.

### DIFF
--- a/contrib/raftexample/raft.go
+++ b/contrib/raftexample/raft.go
@@ -460,12 +460,12 @@ func (rc *raftNode) serveChannels() {
 				rc.publishSnapshot(rd.Snapshot)
 			}
 			rc.raftStorage.Append(rd.Entries)
-			rc.transport.Send(rc.processMessages(rd.Messages))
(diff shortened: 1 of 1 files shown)


serathius:
cc @tbg

ahrtr:
FYI. https://github.com/etcd-io/etcd/pull/15471#issuecomment-1467254962

ahrtr:
Thanks @KumKeeHyun anyway.

KumKeeHyun:
Thanks for reply :)

PR #15454, opened 2023-03-13
Run go mod tidy

serathius:
cc @ahrtr @ptabor


ahrtr:
The PR https://github.com/etcd-io/etcd/pull/15337 forgot to update the go version in `raft/go.mod`, `client/v2/go.mod`, `client/v3/go.mod` and `pkg/go.mod`. The release workflow failed in that PR, because the release always runs against release-3.5, so we ignored the error.

But what's confusing me is why the workflow succeeded after @ptabor merged that PR in https://github.com/etcd-io/etcd/commit/c6029422465d17e257394c99ffe3f35737ced798. There may be something wrong. 

Anyway, thanks for the fix. Let's merge it to unblock any following PRs against release-3.5.

Issue #15469, opened 2023-03-13
add `verify-goimport` to verify all import items are well sorted.


ahrtr:
### What would you like to be added?

We have [go_imports_fix](https://github.com/etcd-io/etcd/blob/bea9c95b7732950e2d04a33dc21e82acf5bf76c2/scripts/fix.sh#L47), but we don't have related verification. 

Let's add a `verify-goimport` in [Makefile#L46](https://github.com/etcd-io/etcd/blob/bea9c95b7732950e2d04a33dc21e82acf5bf76c2/Makefile#L46) to prevent https://github.com/etcd-io/etcd/pull/15467 from happening again in future.

### Why is this needed?

Improve test workflow

HanningLin:
I'd like to work on it and contribute to etcd!

ahrtr:
Thanks. Just assigned to you. @HanningLin

HanningLin:
Can we close this issue?

ahrtr:
Resolved in https://github.com/etcd-io/etcd/pull/15485

PR #15485, opened 2023-03-16
Add verify-goimport to verify all import items are well sorted

HanningLin:

PR fixing this issue https://github.com/etcd-io/etcd/issues/15469

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -632,6 +632,21 @@ function genproto_pass {
   "${ETCD_ROOT_DIR}/scripts/verify_genproto.sh"
 }
 
+function goimport_for_module {
+  GOFILES=$(run go list  --f "{{with \$d:=.}}{{range .GoFiles}}{{\$d.Dir}}/{{.}}{{\"\n\"}}{{end}}{{end}}" ./...)
+  TESTGOFILES=$(run go list  --f "{{with \$d:=.}}{{range .TestGoFiles}}{{\$d.Dir}}/{{.}}{{\"\n\"}}{{end}}{{end}}" ./...)
+  cd "${ETCD_ROOT_DIR}/tools/mod"
+  FILESNEEDSFIX=$(echo "${GOFILES}" "${TESTGOFILES}" | grep -v '.gw.go' | grep -v '.pb.go' | xargs -n 100 go run golang.org/x/tools/cmd/goimports -l -local go.etcd.io)
+  if [ -n "$FILESNEEDSFIX" ]; then
+    log_error -e "the following files are not sync with 'goimports'. run 'make fix'\\n$FILESNEEDSFIX"
+    return 255
+  fi
+}
+
+function goimport_pass {
+  run_for_modules goimport_for_module
+}
+
 ########### MAIN ###############################################################
 
 function run_pass {
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ fuzz:
 
 verify: verify-gofmt verify-bom verify-lint verify-dep verify-shellcheck verify-goword \
 	verify-govet verify-license-header verify-receiver-name verify-mod-tidy verify-shellcheck \
-	verify-shellws verify-proto-annotations verify-genproto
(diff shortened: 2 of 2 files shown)


scripts/test.sh
```diff
@@ -632,6 +632,19 @@ function genproto_pass {
   "${ETCD_ROOT_DIR}/scripts/verify_genproto.sh"
 }
 
+function goimport_for_module {
+  GOFILES=$(run go list  --f "{{with \$d:=.}}{{range .GoFiles}}{{\$d.Dir}}/{{.}}{{\"\n\"}}{{end}}{{end}}" ./...)
+  TESTGOFILES=$(run go list  --f "{{with \$d:=.}}{{range .TestGoFiles}}{{\$d.Dir}}/{{.}}{{\"\n\"}}{{end}}{{end}}" ./...)
+  cd "${ETCD_ROOT_DIR}/tools/mod"
+  FILESNEEDSFIX=$(echo "${GOFILES}" "${TESTGOFILES}" | grep -v '.gw.go' | grep -v '.pb.go' | xargs -n 100 go run golang.org/x/tools/cmd/goimports -l -local go.etcd.io)
```
jmhbnz, line 639: Thanks for raising this @HanningLin!

I'm curious about the `-n 100` for xargs, wouldn't that mean the the full output of files needing to be fixed wouldn't be printed if it was more than 100 files?  If so can we just leave that flag off?

ahrtr: It should be fine. It means at most 100 files per command line.


```
       -n max-args, --max-args=max-args
              Use at most max-args arguments per command line.  Fewer
              than max-args arguments will be used if the size (see the
              -s option) is exceeded, unless the -x option is given, in
              which case xargs will exit.
```

The script is copied from [fix.sh#L34-L39](https://github.com/etcd-io/etcd/blob/c6d8b65ea7a77556f44a826eecf08c83cc966bca/scripts/fix.sh#L34-L39). The only minor comment is duplication.

scripts/test.sh
```diff
@@ -632,6 +632,19 @@ function genproto_pass {
...
+  if [ -n "$FILESNEEDSFIX" ]; then
+    log_error -e "the following files are not sync with 'goimports'. run 'make fix'\\n$FILESNEEDSFIX"
+    return 255
+  fi
+}
+function goimport_pass {
```
ahrtr, line 645: nit: please add a blank line before and after this function.

scripts/test.sh
jmhbnz: Ahh right, xargs will split and sequentially send to  `go run` ok in that case this all looks good to me.

scripts/test.sh
HanningLin: a new line was added

scripts/test.sh
HanningLin: Yes, this is mostly copied from fix.sh but with a minor diff using `-l` to list the files

scripts/test.sh
```diff
@@ -632,6 +632,20 @@ function genproto_pass {
...
+  FILESNEEDSFIX=$(echo "${GOFILES}" "${TESTGOFILES}" | grep -v '.gw.go' | grep -v '.pb.go' | xargs -n 100 go run golang.org/x/tools/cmd/goimports -l -local go.etcd.io)
+  if [ -n "$FILESNEEDSFIX" ]; then
+    log_error -e "the following files are not sync with 'goimports'. run 'make fix'\\n$FILESNEEDSFIX"
+    return 255
+  fi
+}
+
+function goimport_pass {
+  run_for_modules goimport_for_module
+}
```
ahrtr, line 648: nit: please add a new line after the function as well. Please also squash the commits into just one commit.

HanningLin: added and squashed

ahrtr (approved):
LGTM

Thank you @HanningLin

HanningLin:
> LGTM
> 
> Thank you @HanningLin

Thanks for the review!

jmhbnz (approved):
/lgtm

serathius (approved):

Issue #15437, opened 2023-03-09
Provide TLS 1.3 support for 3.4.


ramses:
### What would you like to be added?

TLS 1.3 support was (re-)added to the main branch in https://github.com/etcd-io/etcd/pull/15156

... thus resolving https://github.com/etcd-io/etcd/issues/13506

However, organizations running etcd 3.4 and 3.5 are starting to require TLS 1.3 for security compliance.

---

The "3.5 version" analogue issue: https://github.com/etcd-io/etcd/issues/15436

### Why is this needed?

As stated above, organizations running etcd 3.4 and 3.5 are starting to require TLS 1.3 for security compliance. Also, waiting for etcd 3.6 is not compatible with their timelines.

ahrtr:
Makes sense to me. Please feel free to deliver a PR, thx

jmhbnz:
Initial backport pr open #15486 and docs pr open here https://github.com/etcd-io/website/pull/650. I will raise pr for changelog shortly.

/assign
/remove help

PR #15486, opened 2023-03-16
[3.4] Backport tls 1.3 support

jmhbnz:
TLS 1.3 support was (re-)added to the main branch in https://github.com/etcd-io/etcd/pull/15156

Organizations running etcd 3.4 are starting to require TLS 1.3 for security compliance.

Fixes https://github.com/etcd-io/etcd/issues/15437

### DIFF
--- a/tests/e2e/etcd_config_test.go
+++ b/tests/e2e/etcd_config_test.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"go.etcd.io/etcd/pkg/expect"
 )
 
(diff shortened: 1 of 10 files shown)


tests/e2e/etcd_config_test.go
```diff
@@ -336,3 +337,32 @@ func TestGrpcproxyAndListenCipherSuite(t *testing.T) {
 		})
 	}
 }
+
+func TestEtcdTLSVersion(t *testing.T) {
+	skipInShortMode(t)
```
ahrtr, line 342: `skipInShortMode` doesn't exist in release-3.4, how does the test passed?

tests/e2e/etcd_config_test.go
```diff
@@ -336,3 +337,32 @@ func TestGrpcproxyAndListenCipherSuite(t *testing.T) {
...
+
+	d := t.TempDir()
+	proc, err := spawnCmd(
```
ahrtr, line 345: `spawnCmd` only accepts one parameter, curious why the workflow passed?

tests/e2e/etcd_config_test.go
jmhbnz: Strange yeah, when I run `TEST_OPTS="PASSES='build unit release integration_e2e functional'" make test` locally I can actually see the failure in the logs below.  I was relying on the github actions workflow though, there must be a bug with it which is worrying!

```bash
Starting 'integration_e2e' pass at Thu 16 Mar 2023 18:40:14 NZDT
Running integration and e2e tests...
# go.etcd.io/etcd/tests/e2e [go.etcd.io/etcd/tests/e2e.test]
tests/e2e/etcd_config_test.go:342:2: undefined: skipInShortMode
tests/e2e/etcd_config_test.go:362:6: too many arguments in call to spawnCmd
        have ([]string, nil)
        want ([]string)
FAIL    go.etcd.io/etcd/tests/e2e [build failed]
FAIL
```

Putting that aside - I will get these two issues fixed shortly.

ahrtr: Could you raise an issue to track these two issues? thx

ahrtr (approved):
LGTM

Thanks @jmhbnz

ahrtr:
@jmhbnz  could you please update 3.4 and 3.5 changelog?

Issue #15034, opened 2022-12-21
etcd images: aspect of licensing/busybox


ptabor:
### What would you like to be added?

Follow up to https://github.com/etcd-io/etcd/pull/13556

Currently we generate images based on: 
```
FROM --platform=linux/s390x busybox:1.34.1 as source
FROM --platform=linux/s390x gcr.io/distroless/base-debian11
...
RUN mkdir -p /var/etcd/
RUN mkdir -p /var/lib/etcd/
...
RUN echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf

CMD ["/usr/local/bin/etcd"]
```

Seems that: 
 1. We can use https://docs.docker.com/engine/reference/builder/#workdir to get rid of MKDIR
 2. We can copy the `/etc/nsswitch.conf` to avoid need for `echo` and bash redirect syntax (`>>`)
 3. I don't see other reasons to keep `/bin/sh`
 4. As etcd is statically compiled, we might not need `base-debian11` image at all

### Why is this needed?

To isolate from security and licensing problems.

ahrtr:
Done.

PR #15490, opened 2023-03-16
feat: add riscv64 support

ernado:
Should be marked as [Tier 3](https://etcd.io/docs/v3.5/op-guide/supported-platform/#support-tiers).

### DIFF
--- a/Dockerfile-release.riscv64
+++ b/Dockerfile-release.riscv64
@@ -1,0 +1,16 @@
+FROM --platform=linux/riscv64 scratch
+
+# TODO: use distroless once it supports riscv64
+# https://github.com/GoogleContainerTools/distroless/issues/1269
+
+ADD etcd /usr/local/bin/
+ADD etcdctl /usr/local/bin/
+ADD etcdutl /usr/local/bin/
+
+WORKDIR /var/etcd/
+WORKDIR /var/lib/etcd/
+
+EXPOSE 2379 2380
+
+# Define default command.
+CMD ["/usr/local/bin/etcd"]
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,6 +18,7 @@ jobs:
           - linux-arm64
           - linux-ppc64le
           - linux-s390x
+          - linux-riscv64
     steps:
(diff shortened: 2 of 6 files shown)


Dockerfile-release.riscv64
```diff
@@ -0,0 +1,13 @@
+FROM --platform=linux/ppc64le riscv64/debian:sid-slim
```
serathius, line 1: New releases were migrated to gcr.io/distroless/static-debian11

Don't think we want to have use different images.

ernado:
Looks like it would be currently hard to add riscv64 to distroless, please see https://github.com/GoogleContainerTools/distroless/pull/1268.

Can we use `riscv64/debian:sid-slim` for now?

Dockerfile-release.riscv64
ahrtr, line 1: copy paste typo?  `--platform=linux/riscv64`

ahrtr: Previously I saws lots of CVEs, so eventually we migrated to distroless base image.

ernado: y, fixed

serathius: We don't want to use busybox due to licensing issues https://github.com/etcd-io/etcd/issues/15034

serathius:
Need to agree with distroless maintainer arches to the build is quite expensive and makes it harder to maintain. Especially if we would need to use separate base image for riscv64. For now I'm against adding support riscv64 until https://www.debian.org/ports/ and distroless supports it

Dockerfile-release.riscv64
ernado: It looks like `FROM scratch` just works.

ernado:
Closing this until distroless dependency is resolved.

Will be maintaining forks and binaries under https://github.com/go-riscv.

Also, there is signed distroless image from go-riscv:
https://github.com/go-riscv/distroless

```bash
docker pull ghcr.io/go-riscv/distroless/static-unstable:latest
```

```dockerfile
FROM ghcr.io/go-riscv/distroless/static-unstable:latest
```

PR #15492, opened 2023-03-16
Fix 15271

serathius:
Fixes https://github.com/etcd-io/etcd/issues/15271
cc @ahrtr @ptabor

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -448,12 +448,15 @@ func (s *watchableStore) notify(rev int64, evs []mvccpb.Event) {
 			pendingEventsGauge.Add(float64(len(eb.evs)))
 		} else {
 			// move slow watcher to victims
-			w.minRev = rev + 1
 			w.victim = true
 			victim[w] = eb
 			s.synced.delete(w)
 			slowWatcherGauge.Inc()
 		}
+		// always update minRev
+		// in case 'send' returns true and watcher stays synced, this is needed for Restore when all watchers become unsynced
+		// in case 'send' returns false, this is needed for syncWatchers
+		w.minRev = rev + 1
 	}
 	s.addVictim(victim)
 }
--- a/server/storage/mvcc/watchable_store_test.go
+++ b/server/storage/mvcc/watchable_store_test.go
@@ -296,39 +296,38 @@ func TestWatchRestore(t *testing.T) {
 
 			testKey := []byte("foo")
 			testValue := []byte("bar")
-			rev := s.Put(testKey, testValue, lease.NoLease)
-
-			newBackend, _ := betesting.NewDefaultTmpBackend(t)
-			newStore := newWatchableStore(zaptest.NewLogger(t), newBackend, &lease.FakeLessor{}, StoreConfig{})
-			defer cleanup(newStore, newBackend)
-
-			w := newStore.NewWatchStream()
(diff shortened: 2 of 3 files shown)


serathius:
cc @lavacat

serathius:
Proposed followup: ensure that watchableStorage tests don't access storage method.

ahrtr:
> Found problem with `watchableStore` tests. They don't call their own methods, but ones from underlying `storage`. As result they don't trigger same watch notification code as ones done by apply flow.

The second commit makes no sense to me. Note the `watchableStore` just adds a watchable layer on top of the underlying `store`, why do we encapsulate the KV API again? If you see any issue in unit test, shouldn't we fix test, not the production code?

serathius:
> > Found problem with `watchableStore` tests. They don't call their own methods, but ones from underlying `storage`. As result they don't trigger same watch notification code as ones done by apply flow.
> 
> The second commit makes no sense to me. Note the `watchableStore` just adds a watchable layer on top of the underlying `store`, why do we encapsulate the KV API again? If you see any issue in unit test, shouldn't we fix test, not the production code?

Problem is that `watchableStore` exposes two ways to do writes via `s.Put(...)` and via `s.Write(...).Put(...)` (symmetrical for delete). Only the second case goes through proper `watchableStore` code. First case goes directly to `storage` instance underneath so it doesn't trigger proper watch notification.

Finding the issue required two things:
* Watching for all events within 1 second, instead of first event. Thus we can find duplication in event.
* Changing the `Put` call to go to `watchableStore` and not `store` underneath. Reason was that `notify` function that had the bug, was not called in tests at all because all `Put` requests went to `store`.

I would argue that this is not a unit test as it tests multiple layers (`watchableStore`, `store` and `backend`) working together. I consider it a bug that testing of `watchableStore` called `Put` method not from `watchableStore`, but from `store`. 

Aside of the issue with test, I find it confusing and weird that calling `watchableStore.Put(...)` would not notify watchers.
 cc @mitake

ahrtr:
Note neither `watchableStore` nor the underlying `store` expose the `DeleteRange` & `Put` method. Instead, only interface `WriteView` exposes them. 



> * Watching for all events within 1 second, instead of first event. Thus we can find duplication in event.

This is a valid point.

ahrtr (changes requested):
https://github.com/etcd-io/etcd/pull/15492#issuecomment-1475456277

lavacat:
Thanks for adding tests.

I think @ahrtr has a point. When I tried running your updated test without new watchableStore.Put, it passed. When stepping through in debug mode, it goes to writeView.Put and `notify` is triggered in watchableStoreTxnWrite.End()

serathius:
Problem is that `writeView` is a sub-struct of `store` and `store` is sub-struct `watchableStore`. This means that `watchableStore` can be working even though it was not incorrectly initialized. For correct functioning `writeView` needs to be [reconfigured to point to `watchableStore`](https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/watchable_store.go#L88-L89). 

This happens in tests where we create `watchableStore` struct without redirecting `writeView` to it. This is still very error prone because `mvcc` package does a lot of usage of private fields/methods. I agree that proposed change in this PR is not optimal, so I will create separate PR to clean up the golang api.

ahrtr (approved):
LGTM

Thanks

Issue #15247, opened 2023-02-06
All leases are revoked when the etcd leader is stuck in handling raft Ready due to slow fdatasync or high CPU.


aaronjzhang:
### What happened?

We have 3 etcd nodes which are running on k8s, the etcd pod use portworx storage for db saving, we've found when rebooting the portworx node, it will cause etcd to be stuck in fdatasync, if this etcd node is leader, then it will cause below 2 lease keep alive issues.

1. the old leader still revokes lease after it steped to follower, this is the same issue as [#12528 ](https://github.com/etcd-io/etcd/issues/12528)

some clients connected to other etcd nodes and send keep alive to them, after the old leader steped to follower, the new leader will handle the lease keep alive, but the old leader's lessor still considered itself is the primary and revoked all leases after their ttl expired in its view, after the old leader returned from fdatasync, these revoke request will be still sent out when handling the next Ready.

2. the old leader still handle lease keep alive request from the connected client.

some clients connected to the old leader etcd node, when the old leader was stuck in fdatasync, the raft layer steped to follower, but the lessor was not demoted, the connected clients can still send keep alive request to the old leader, at the same time the other two etcd nodes elected a new leader and didn't receive the lease keep alive message which should be sent to them but they were still sent to old leader, so these leases were revoked after the ttl expired.

Essentially the above 2 issues are because of the lessor primary is not synced up with the raft leader, there is big gap between lessor primary and raft leader when etcd server is stuck in processing raft Ready.  

### What did you expect to happen?

1. The lease which is kept alive on old leader should return an error when its raft stepped to follower so that the clients can connect to other etcd nodes.
2. The old leader should not revoke the other leases.

### How can we reproduce it (as minimally and precisely as possible)?

1. Add a piece of codes in func (w *WAL) sync() function to simulate fdatasync stuck
```
func (w *WAL) sync() error {
	if w.encoder != nil {
		if err := w.encoder.flush(); err != nil {
			return err
		}
	}
	start := time.Now()
	err := fileutil.Fdatasync(w.tail().File)

    //Simulate fdatasync is stuck
	if fileutil.Exist("/run/etcd/fdatasync_block") {
		f, err := ioutil.ReadFile("/run/etcd/fdatasync_block")
		if err == nil {
			str_num := strings.Trim(string(f), "\n")
			block_duration, err := strconv.Atoi(str_num)
			if err == nil {
				time.Sleep(time.Duration(block_duration) * time.Second)
			} else {
[...]

Issue #15498, opened 2023-03-17
Invalid db page layout


serathius:
### What happened?

Etcd started crashlooping with 
```
From etcd logs:

panic: runtime error: slice bounds out of range [:8] with capacity 5

goroutine 111 [running]:
go.etcd.io/etcd/mvcc.bytesToRev(0x7f760418e0b0, 0x5, 0x5, 0x0, 0x17f7780)
        /go/src/go.etcd.io/etcd/mvcc/revision.go:58 +0x85
go.etcd.io/etcd/mvcc.restoreIntoIndex.func1(0xc000070000, 0xc000074720, 0x11edc08, 0xc00017d620, 0xc00024a060)
        /go/src/go.etcd.io/etcd/mvcc/kvstore.go:515 +0x287
created by go.etcd.io/etcd/mvcc.restoreIntoIndex
        /go/src/go.etcd.io/etcd/mvcc/kvstore.go:490 +0xaf
```

When  analysing the db file I found invalid etcd db file layout.
Under bucket branch page, in `keys` bucket there was a branch page linking to another bucket branch page.
This resulted in bbolt returning key `alarm` when reading whole `keys` bucket. This is correct layout for bbolt, but not for etcd.

From etcd point of view this is invalid as it assumes that all keys in `keys` bucket are revision numbers.
Panic from above comes from `bytesToRev` function that parses revision. It failed as main rev has 8 bytes, while key "alarm" has only 5 bytes.

This means that at some point bbolt either:
* overridden page that was in use with buckets page
* incorrectly pointed branch page into buckets page

We can't exclude hardware issue that resulted in memory corruption. 

Providing the [dump.txt](https://github.com/etcd-io/etcd/files/11002136/dump.txt) for further investigation 

### What did you expect to happen?

Want to report the issue to start the discussion of etcd handling potential memory corruptions.

Assuming that this was indeed a memory corruption, I expect that should avoid writing corrupted page to disk.
Running mmapped memory comes with risk with memory stamping, etcd should have mechanism that prevent corruption from being persisted.

Was discussing with @ptabor idea of protected mode for bbolt where it would verifying every write to ensure corruptions are not persisted.

### How can we reproduce it (as minimally and precisely as possible)?

Don't think so. 

### Etcd version (please run commands below)

v3.4.21

### Etcd configuration (command line flags or environment variables)

Nothing unusual

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

N/A

[...]

PR #15131, opened 2023-01-17
dependency: bump google.golang.org/grpc from 1.51.0 to 1.52.0

ahrtr:
dependabot automaticaly bumped gRPC, but unfortunately it doesn't execute "./scripts/fix.sh", and it also created a PR for each module.

PR #15499, opened 2023-03-17
Propose project roadmap

serathius:
Format based on https://github.com/etcd-io/etcd/blob/release-3.5/ROADMAP.md
cc @ahrtr @ptabor

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,24 @@
+# etcd roadmap
+
+This document defines high level goals for project.
+
+## Milestones
+
+* [P0] Etcd releases are qualified by rigorous robustness testing
+* [P0] Etcd can reliably detect data corruption
+* [P1] Experimental features are graduated or removed
+* [P1] Etcd testing is high quality, easy to maintain and expand
+* [P1] Etcd v2 API and storage is removed and code cleaned up
+* [P1] Etcd supports zero-downtime downgrade
+* [P2] Etcd can automatically recover from data corruption
+
+Each listed milestone should have a corresponding 
+[issue](https://github.com/etcd-io/etcd/issues) or
+[milestone](https://github.com/etcd-io/etcd/milestones) on GitHub. 
+If it doesn't please [let us know](https://github.com/etcd-io/etcd#contact).
+
+### Priorities
+
+* P0 - Critical for reliability of the v3.5 and v3.4 releases. Should be prioritized this over all other work and back-ported.
+* P1 - Important for long term success of the project. Blocks v3.6 release.
+* P2 - Stretch goals that would be nice to have for v3.6, however should not be blocking.
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Now it's time to dig into the full etcd API and other guides.
 
 - Email: [etcd-dev](https://groups.google.com/forum/?hl=en#!forum/etcd-dev)
 - Slack: [#etcd](https://kubernetes.slack.com/messages/C3HD8ARJ5/details/) channel on Kubernetes ([get an invite](http://slack.kubernetes.io/))
+- [Roadmap](./ROADMAP.md)
 - [Community meetings](#Community-meetings)
 
 ### Community meetings


serathius:
cc @spzala @lavacat @jmhbnz @chaochn47 @tjungblu @mitake

jmhbnz (approved):
Looks good to me. 

Two things I think would be nice to add:
  - Section to mention feature tracking, with reference to feature tracking board if we are still going to use it for future releases: https://github.com/etcd-io/etcd/projects/1
  - Once known, we should add links to each item so it links through to the issue or milestone directly.

ROADMAP.md
```diff
@@ -0,0 +1,24 @@
+# etcd roadmap
+
+This document defines high level goals for project.
+
+## Milestones
+
+* [P0] Etcd releases are qualified by rigorous robustness testing
+* [P0] Etcd can reliably detect data corruption
```
chaochn47, line 8: How does the above `P1` 3 action items fit into the Milestones section here? 

> etcd apply code should be easy to understand and validate correctness
> etcd can reliably detect data corruption (hash is linearizable)
(This P1 action item is mentioned as P0)

> etcd recovery from data inconsistency procedures are documented and tested

How does the stalled write due to slow disk fit into the Milestone? @ahrtr 

What about lease redesign? https://github.com/etcd-io/etcd/pull/14094

serathius: > etcd apply code should be easy to understand and validate correctness

I would want to avoid touching random parts of apply code without better testing and clear goal. Removal of v2 API and following cleanup should already improve the situation and takes priority.

 > etcd can reliably detect data corruption (hash is linearizable)
 
 This should be covered by ` Etcd can reliably detect data corruption`, maybe requires rewrite/clarification. I might downgrade it back to P1 as it doesn't help with v3.4 release.
 
 > etcd recovery from data inconsistency procedures are documented and tested.
 
 Documentation is done, we just need to add test. I think we should file an issue as important but goal milestone itself should be tracked as part of improvements to testing.
 
 > How does the stalled write due to slow disk fit into the Milestone? @ahrtr

This is somewhat new effort that is still not well defined. For me it comes under reliability, which is important, but as it relates to hardware failures it's not something etcd tackled yet. However with recently reported https://github.com/etcd-io/etcd/issues/15498 I would want to propose "etcd is resilient to hardware failures" soon.

> What about lease redesign? 

Correctness should be our top priority, however leases have been broken for long time and no-one cared (K8s also doesn't). As so I would treat it second priority to KV API.
At some point we could consider an larger effort "etcd APIs is high quality and has consistent behavior" that would encompass leases.

tjungblu: Is the scope for this item clear? IIRC there was a discussion on corruption detection per key/value, then there were some discussions around merkle trees and partitioning the keyspace.

serathius: > Is the scope for this item clear? IIRC there was a discussion on corruption detection per key/value, then there were some discussions around merkle trees and partitioning the keyspace.

If you mean corruption detection scope, then not. I didn't have time to define it. It's pretty large issue to tackle and are multiple ways to approach it. Main challenge is balancing breaking changes and short term vs long term improvements. I have couple of ideas that I discussed with @ptabor, but didn't have time to write them down as I want to focus on finishing robustness tests first (not too long). 

Happy to make the scope clearer if someone is interested in working on it. Still would like to encourage people to work on robustness tests first, as they also help v3.4 and v3.5 releases.

serathius:
PING @ahrtr @ptabor for feedback

ROADMAP.md
```diff
@@ -0,0 +1,24 @@
+# etcd roadmap
+
+This document defines high level goals for project.
+
+## Milestones
```
ptabor, line 5: We should be careful about redefining meaning of the words. 

So far in etcd we were using milestones = future minor release:   v3.5, v3.6.
Here we naming as milestone a focus area we want to invest. 

Let's use consistent terms.  My proposal it to think about this as 3 level hierarchy:  

  1. Milestones -> are milestones as defined in https://github.com/etcd-io/etcd/milestones. Let's keep them as publicly visible releases (might be patch). 
  
  2. Efforts
       - are bugs that track progress on multiple issues that need to be addressed with a common objective
       - [or] (for bigger efforts) projects: https://github.com/etcd-io/etcd/projects. 
         Still I would represent project as umbrella issue -> as it seems that project cannot be assigned to a milestone. 
   3. Issues - for individual work items.
   
 Now the question remains: 
 If we have a tool to dynamically track the milestones with attached efforts / items, do we need to redundantly track it in a markdown doc ? 
 
And I would say - we don't.  I assume that the purpose of the doc is different. It's a statement of intent what we want to focus in following releases. And thanks to being submitted by maintainers and reviewed, it forces them to be on the same page (as opposed to an individual maintainer assigning an issue to a milestone). But If that's the goal, let's call it explicitly in the preamble to this doc. 

Then let's have: 

## Milestones:
 
### [release-v3.6](https://github.com/etcd-io/etcd/milestone/38)

The main focus of the v3.6 is the reduction of technical debts. The explicit goal is to avoid new features. 
The focus will be on: 
  1.  deprecate/decomission experimental / legacy cody:
       - decommission storage v2 (link to a tracking bug)
       - experimental features are graduated or removed  (link to a tracking bug)
  3. ...

### [release-v3.5.x](https://github.com/etcd-io/etcd/milestone/38)

The same as release-v3.4.x.

### [release-v3.4.x](https://github.com/etcd-io/etcd/milestone/38)

The release focuses on stability. Etcd maintainers are going to backport: 
   - critical/important vulnerabilities
   - backportable test coverage (including robustness tests)
   - critical correctness or robustness issue fixes
   - not invasive flags/warnings that enable easier transition to next release.

ahrtr:
I think this PR mixed up three concepts/things:
1. etcd's core value as a K/V datastore;
2. priorities for now and in the near future.
    I agree the data inconsistent or corruption related issues should be our top priority. Currently I want to focus on https://github.com/etcd-io/bbolt/issues/402 and https://github.com/etcd-io/etcd/issues/15498
3. roadmap
    roadmap gives people what will be included in future releases, e.g. 3.6, 3.7, etc. I don't think we need a roadmap for stable release (3.4 and 3.5); because we don't add new features for stable release, we just fix issues, including bugs and CVEs. Note that the items included in roadmap aren't necessary to be priorities for now, they are just direction of the project. Actually there is no rush to add a roadmap for now, because we all know we are still focusing on qualify our stable releases.

What should be included in future releases? This should be open to discussion. See my proposal (on top of @ptabor 's ). Note that adding whatever testing is a continuous effort, and is always welcome.

## 3.6.0
1. remove the dependency on experimental gRPC API: https://github.com/etcd-io/etcd/issues/15145 . Note that currently we also also stuck on bumping gRPC due to dependency on experimental API, see https://github.com/etcd-io/etcd/pull/15131
2. Partially resolve https://github.com/etcd-io/etcd/issues/15247. e.g. https://github.com/etcd-io/etcd/pull/15440. I tend to refactoring lease in 3.7 or 4.0, after we have more confidence on the project (etcd).
3. Watch stream starvation. Note although it's an issue, it may introduce huge change in https://github.com/etcd-io/etcd/pull/15510.

## 3.7.0 (or 4.0)
1. Refactoring leases to completely resolve issue like https://github.com/etcd-io/etcd/issues/15247. It's big & challenge task. Raft is just like a network transportation channel instead of a consensus mechanism, see https://github.com/etcd-io/etcd/pull/14094
2. Cleanup legacy protobuf libraries, see https://github.com/etcd-io/etcd/issues/14533. It's an important but not urgent task.
3. Evaluate & Integrate the new raft feature (https://github.com/etcd-io/raft/pull/8) into etcdserver;
4. ...

serathius:
For now I separated postmortem update https://github.com/etcd-io/etcd/pull/15552

ROADMAP.md
geetasg: is there any tracking issue for changes to catch corruption per key/value?

serathius:
Closing for https://docs.google.com/document/d/1KFrLeKyHvNDv4b0GswYU3_zbeSATgCanT5dzCID4oJY/edit?usp=sharing

PR #15500, opened 2023-03-17
document measure-test-flakiness.sh

chaochn47:

### DIFF
--- a/scripts/measure-test-flakiness.sh
+++ b/scripts/measure-test-flakiness.sh
@@ -15,6 +15,33 @@ trap '{ rm -rf -- "${temp_dir}"; }' EXIT
 
 json_file="${temp_dir}/commit-and-check-data.json"
 
+# This GraphQL query retrieves information about the most recent commits to the default branch of the "etcd" repository owned by "etcd-io".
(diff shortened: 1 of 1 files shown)


chaochn47:
The failed test has been addressed by https://github.com/etcd-io/etcd/pull/15502

jmhbnz (changes requested):
Good idea to make this more readable.  I'm wondering if there is an opportunity here to eliminate duplication altogether and just set a bash variable that is readable and refer to it in the curl, something like:

```bash

graphql_query='
query {
  repository(owner: "etcd-io", name: "etcd") {
    defaultBranchRef {
      target {
        ... on Commit {
          history(first: 100) {
            edges {
              node {
                ... on Commit {
                  commitUrl
                  statusCheckRollup {
                    state
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
'
```

Then we could use it in the `curl` command something like (haven't tested):

```bash
curl --fail --show-error --silent -H "Authorization: token ${GITHUB_TOKEN}" \
    -X POST \
    -d "{
        \"query\": \"$graphql_query\"
    }" \
    https://api.github.com/graphql | jq . > "${json_file}"
```

If something like that could work it would remove the need to maintain the query in two places.

serathius (approved):

Issue #15487, opened 2023-03-16
Actions workflow for 3.4 passed despite failures within tests


jmhbnz:
### Which github workflows are flaking?

This workflow run succeeded despite me having errors with one of the tests defined: https://github.com/etcd-io/etcd/actions/runs/4433391180 

Issue was with file `tests/e2e/etcd_config_test.go`

### Which tests are flaking?

Specifically this job: https://github.com/etcd-io/etcd/actions/runs/4433391180/jobs/7778376779

Which finished with the following job output:
```bash
# go.etcd.io/etcd/tests/e2e [go.etcd.io/etcd/tests/e2e.test]
Error: tests/e2e/etcd_config_test.go:342:2: undefined: skipInShortMode
Error: tests/e2e/etcd_config_test.go:362:6: too many arguments in call to spawnCmd
	have ([]string, nil)
	want ([]string)
FAIL	go.etcd.io/etcd/tests/e2e [build failed]
FAIL
```

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4433391180/jobs/7778376779

### Anything else we need to know?

Expectation is this kind of failure would cause the job to fail.

Relates to https://github.com/etcd-io/etcd/pull/15486

fuweid:
/assign 

I'm debugging some flaky testcases in pipeline. Let me handle this. Thanks

fuweid:
Closed by #15504

PR #8655, opened 2017-10-06
Makefile: suffix test log files

gyuho:
In preparation of running all tests inside container.
Currently, we run Jenkins in shared environment.
This is not good. Need manual Go runtime updates,
cannot run two different branches, port conflicts,
out of disk on /tmp, etc.

PR #15504, opened 2023-03-18
fix: use pipefail to show the error

fuweid:
fixes: #15487


.github/workflows/tests.yaml
```diff
@@ -58,8 +58,12 @@ jobs:
             GO_BUILD_FLAGS='-v' GOARCH=s390x ./build
             ;;
           linux-amd64-grpcproxy)
+            set -o pipefail; set +e;
```
fuweid, line 61: if it looks good, I will file a fix pr for main branch.

jmhbnz (approved):
LGTM

To test I checked out this pr, pushed as a new branch with a bogus edit to a test file.
Opening a PR against 3.4 the workflow now correctly failed, refer: https://github.com/etcd-io/etcd/actions/runs/4454519493/jobs/7823765743?pr=15506 

Thanks @fuweid 🙏🏻

ahrtr (approved):
Good catch.

Thank you @fuweid 

Usually we fix the main branch firstly, then backport to 3.5 & 3.4. But I won't insist on it for such a small change.

.github/workflows/tests.yaml
```diff
@@ -58,8 +58,12 @@ jobs:
...
             PASSES='build grpcproxy' CPU='4' RACE='true' ./test 2>&1 | tee test.log
-            ! egrep "(--- FAIL:|DATA RACE|panic: test timed out|appears to have leaked)" -B50 -A10 test.log
+            ecode=$?
+            egrep "(--- FAIL:|DATA RACE|panic: test timed out|appears to have leaked)" -B50 -A10 test.log
```
ahrtr, line 64: Why do you remove the `!` ? We may need to keep it. If we find any `(--- FAIL:|DATA RACE|panic: test timed out|appears to have leaked)`, we may still consider the test failed.

fuweid: Sorry for the late reply.

 I was thinking if we can delete the `egrep` command in this patch. 

I didn't see any comment about egrep introduced in https://github.com/etcd-io/etcd/pull/8655. 
It seems that the egrep is used to fix the unexpected exit code, since `| tee` won't return expected exit code because of `pipefail`.

As the `tee` already writes the log in the stdout, we won't lost any details and the command will return expected exit code. Maybe we can go with the following script:

```bash
set -o pipefail
PASSES='build grpcproxy' CPU='4' RACE='true' ./test 2>&1 | tee test.log
```

Does it make senses?

ahrtr: It should be OK, because the following case will never happen:
The test script/cases exits with 0, but somehow there are some `(--- FAIL:|DATA RACE|panic: test timed out|appears to have leaked)` errors.

fuweid: Yes. So I believe that egrep is to fix the `pipefail` issue. I will update the pr later and send new pr to main branch as well.

serathius: One of the things that first caught my eye in etcd scripting is fact that it doesn't run in strict mode and it even depends on some parts failing for correctness. By strict mode I mean `set -euo pipefail` as defined in http://redsymbol.net/articles/unofficial-bash-strict-mode/. 

To my understanding `!grep (--- FAIL:|...` was a workaround for not properly written scripts. I think long term it would be better to fix all scripts and switch on strict mode. What do you think? @ahrtr @fuweid

ahrtr: > By strict mode I mean `set -euo pipefail` as defined in

Makes sense to me.

serathius (changes requested):
It came to me while reviewing https://github.com/etcd-io/etcd/pull/15519.
We cannot just set `set -o pipefail; set +e;` in `run` as bash doesn't pass the flags we set to scripts it calls. This will only affect `run` and not `./test`.

So this change doesn't work if `./test` runs `go test` that writes `--- FAIL:` to stdout and doesn't exit with non-zero status code.

test
```diff
@@ -588,6 +596,8 @@ function receiver_name_pass {
 }
 
 function commit_title_pass {
+	# TODO: git rev-parse --abbrev-ref doesn't work in pull-request and it
```
fuweid, line 599: image

https://github.com/etcd-io/etcd/actions/runs/4488556590/jobs/7893308908?pr=15504#step:5:135

fuweid:
ping @serathius @ahrtr @jmhbnz please take a look. Thanks

serathius (approved):

serathius:
Can you prepare similar PR for v3.5?

fuweid:
> Can you prepare similar PR for v3.5?

yeah. Still working on it. Will file it when it is ready.

Issue #15495, opened 2023-03-17
Flaking TestConnectionMultiplexing


serathius:
### Which github workflows are flaking?

e2e tests

### Which tests are flaking?

TestConnectionMultiplexing

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4439049893/jobs/7790956875

serathius:
Blocks https://github.com/etcd-io/etcd/pull/15446

serathius:
cc @fuweid

serathius:
Problem I encountered is lack of proper logs from etcdctl spawn. Because we don't pass logger to `SpawnCmd` we don't get logs about the process execution. Will send PR to fix that.

chaochn47:
All the unrelated successful sub tests are commented out. Only `ServerTLSAndNonTLS` and `clientScenarios = []e2e.ClientConnType{e2e.ClientNonTLS}` is left. 

```
cd tests
ETCD_CLIENT_DEBUG=true EXPECT_DEBUG=true taskset -c 1 go test -v -failfast -count=1 \
  -run TestConnectionMultiplexing ./e2e/
```

```
/home/chaochn/bin/etcd (TestConnectionMultiplexingServerTLSAndNonTLS-test-0) (1121874): 
{
    "level": "info",
    "ts": "2023-03-18T07:44:35.806311Z",
    "caller": "zapgrpc/zapgrpc.go:178",
    "msg": "[core] [Server #4] grpc: Server.Serve failed to create ServerTransport: 
        connection error: desc = \"ServerHandshake(\\\"127.0.0.1:34630\\\") failed: 
        tls: first record does not look like a TLS handshake\""
}
```

Not sure why client is configured to use NonTLS but server requires TLS handshake.. only when CPU core is low.  From the tcpdump, without setting CPU core, TLS traffic can go through.. I hope the information is useful

serathius:
Great find @chaochn47.

Running `ETCD_CLIENT_DEBUG=true EXPECT_DEBUG=true taskset -c 1 go test -v -failfast -count=1  -run TestConnectionMultiplexing ./e2e/`
```
--- FAIL: TestConnectionMultiplexing (11.75s)
    --- PASS: TestConnectionMultiplexing/ServerTLS (2.24s)
        --- PASS: TestConnectionMultiplexing/ServerTLS/ClientTLS (1.89s)
[...]

PR #15509, opened 2023-03-18
server/embed: fix data race when start insecure grpc

fuweid:

fix: #15495

### DIFF
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -114,27 +114,29 @@ func (sctx *serveCtx) serve(
 	servElection := v3election.NewElectionServer(v3c)
 	servLock := v3lock.NewLockServer(v3c)
 
-	var gs *grpc.Server
-	defer func() {
-		if err != nil && gs != nil {
-			sctx.lg.Warn("stopping grpc server due to error", zap.Error(err))
-			gs.Stop()
-			sctx.lg.Warn("stopped grpc server due to error", zap.Error(err))
-		}
-	}()
-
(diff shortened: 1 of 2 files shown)


serathius (approved):

server/embed/serve.go
```diff
@@ -170,12 +175,22 @@ func (sctx *serveCtx) serve(
...
 			sctx.serviceRegister(gs)
 		}
+
+		defer func(gs *grpc.Server) {
+			if err != nil {
+				sctx.lg.Warn("stopping secure grpc server due to error", zap.Error(err))
+				gs.Stop()
+				sctx.lg.Warn("stopped secure grpc server due to error", zap.Error(err))
+			}
+		}(gs)
```
ahrtr, line 192: Minor comment: the code block is almost identical to the code block for insecure. I am thinking to define a separate function something like below, then call it in both secure and insecure code block. 

```
func closegRPCServer(gs *grpc.Server) {
  ...
}

```

ahrtr (approved):
Great work! Thank you @fuweid

server/embed/serve.go
fuweid: Since it needs `return error` to the log, I'm not sure how to handle this with your idea. In here, I follow the defer pattern~

serathius:
@fuweid Could you backport it to release-3.5 and release-3.4?

fuweid:
> @fuweid Could you backport it to release-3.5 and release-3.4?

Sure! Will send prs later.

PR #15517, opened 2023-03-20
[3.5] server/embed: fix data race when start insecure grpc

fuweid:

fix: #15495

(cherry picked from commit a9988e2625eede1af81d189b5f2ecf7d4af3edf1)

cherry-pick: #15509


serathius (approved):

ahrtr (approved):
LGTM

Thank you @fuweid

PR #15518, opened 2023-03-20
[3.4] server/embed: fix data race when start insecure grpc

fuweid:

fix: #15495

(cherry picked from commit a9988e2625eede1af81d189b5f2ecf7d4af3edf1)

cherry-pick: #15509

### DIFF
--- a/embed/serve.go
+++ b/embed/serve.go
@@ -17,7 +17,6 @@ package embed
 import (
 	"context"
 	"fmt"
-	"golang.org/x/net/http2"
 	"io/ioutil"
 	defaultLog "log"
 	"math"
@@ -44,6 +43,7 @@ import (
 	"github.com/soheilhy/cmux"
 	"github.com/tmc/grpc-websocket-proxy/wsproxy"
 	"go.uber.org/zap"
+	"golang.org/x/net/http2"
 	"golang.org/x/net/trace"
 	"google.golang.org/grpc"
 )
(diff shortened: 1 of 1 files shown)


embed/serve.go
```diff
@@ -102,22 +102,38 @@ func (sctx *serveCtx) serve(
...
 		v3electionpb.RegisterElectionServer(gs, servElection)
 		v3lockpb.RegisterLockServer(gs, servLock)
 		if sctx.serviceRegister != nil {
 			sctx.serviceRegister(gs)
 		}
+
+		defer func(gs *grpc.Server) {
+			if err == nil {
+				return
+			}
```
ahrtr, line 116: The gs will not be stopped in such case?

chaochn47: `gs` is serving in the background. So I guess on etcd stop, gs will be gracefully stopped.

chaochn47 (approved):

embed/serve.go
fuweid: > The gs will not be stopped in such case?

Currently, it won't happen. The `cMux.Serve` always return error. I was thinking that we can remove that error check and just stop it. But this pr is used to fix data-race issue. It can be updated in the follow-up.

fuweid:
@serathius @chaochn47 @ahrtr thanks for the review. I would like to mark it ready after we introduces that strict mode script.

embed/serve.go
ahrtr: OK, thx for the clarification.

ahrtr:
> @serathius @chaochn47 @ahrtr thanks for the review. I would like to mark it ready after we introduce the strict mode script.

The two PRs (this one and 15519) seem unrelated.  Feel free to mark it ready for review when you feel conformable.

fuweid:
@ahrtr It is ready. Thanks

ahrtr (approved):
LGTM

Thank you @fuweid

ahrtr:
Could you please update both 3.4 and 3.5 changelog as well?

fuweid:
@ahrtr Sure. Will file pr later. Thanks for the reminder.

Issue #15514, opened 2023-03-20
Rewrite etcd scripts in strict mode


serathius:
### What would you like to be added?

Follow up from https://github.com/etcd-io/etcd/pull/15504#discussion_r1141795658

Strict mode as defined in http://redsymbol.net/articles/unofficial-bash-strict-mode/

List of scripts:
```
$ find . -name *.sh
./tools/mod/install_all.sh
./tools/local-tester/faults.sh
./tools/local-tester/bridge.sh
./tools/rw-heatmaps/rw-benchmark.sh
./tools/etcd-dump-logs/testdecoder/decoder_wrongoutputformat.sh
./tools/etcd-dump-logs/testdecoder/decoder_correctoutputformat.sh
./tests/integration/fixtures-expired/gencerts.sh
./tests/fixtures/gencerts.sh
./tests/manual/docker-dns/insecure/run.sh
./tests/manual/docker-dns/certs-gateway/run.sh
./tests/manual/docker-dns/certs-gateway/gencerts.sh
./tests/manual/docker-dns/certs-wildcard/run.sh
./tests/manual/docker-dns/certs-wildcard/gencerts.sh
./tests/manual/docker-dns/certs/run.sh
./tests/manual/docker-dns/certs/gencerts.sh
./tests/manual/docker-dns/certs-common-name-auth/run.sh
./tests/manual/docker-dns/certs-common-name-auth/gencerts.sh
./tests/manual/docker-dns/certs-common-name-multi/run.sh
./tests/manual/docker-dns/certs-common-name-multi/gencerts.sh
./tests/manual/docker-dns/certs-san-dns/run.sh
./tests/manual/docker-dns/certs-san-dns/gencerts.sh
./tests/manual/docker-dns-srv/certs-gateway/run.sh
./tests/manual/docker-dns-srv/certs-gateway/gencerts.sh
./tests/manual/docker-dns-srv/certs-wildcard/run.sh
./tests/manual/docker-dns-srv/certs-wildcard/gencerts.sh
./tests/manual/docker-dns-srv/certs/run.sh
./tests/manual/docker-dns-srv/certs/gencerts.sh
./tests/manual/docker-static-ip/certs-metrics-proxy/run.sh
./tests/manual/docker-static-ip/certs-metrics-proxy/gencerts.sh
./tests/manual/docker-static-ip/certs/run.sh
./tests/manual/docker-static-ip/certs/gencerts.sh
./hack/patch/cherrypick.sh
[...]

PR #15519, opened 2023-03-20
chore: introduce strict bash mode for scripts,.github

fuweid:
~~According to Github Actions [doc][1], linux platform will use `bash -e {0}` mode to run script. But we also need `pipefail` and `nounset` in CI. This patch is aimed to add strict mode setting before any run.~~

~~And remove the `| tee` and egrep commands since we don't need this workaround to show error.~~

~~[1]: <https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell>~~

REF: #15514


.github/workflows/e2e-arm64.yaml
```diff
@@ -23,11 +23,12 @@ jobs:
...
         TARGET: ${{ matrix.target }}
       run: |
+        set -euo pipefail
+
         echo "${TARGET}"
         case "${TARGET}" in
           linux-arm64-e2e)
-            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh 2>&1 | tee test.log
-            ! grep -E "(--- FAIL:|FAIL:|DATA RACE|panic: test timed out|appears to have leaked)" -B50 -A10 test.log
+            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
```
serathius, line 31: Does this affect child bash that is created when running scripts?

fuweid: Good point! Unfortunately, child bash doesn't inherit it. 

```bash
➜  ~ bash
fuwei@wenchang:~$ echo $SHELLOPTS
braceexpand:emacs:hashall:histexpand:history:interactive-comments:monitor
fuwei@wenchang:~$
fuwei@wenchang:~$ # no pipefail
fuwei@wenchang:~$ (echo 123; exit 2) | tee
123
fuwei@wenchang:~$ echo $?
0
fuwei@wenchang:~$
fuwei@wenchang:~$ set -o pipefail
fuwei@wenchang:~$ echo $SHELLOPTS
braceexpand:emacs:hashall:histexpand:history:interactive-comments:monitor:pipefail
fuwei@wenchang:~$
fuwei@wenchang:~$ (echo 123; exit 2) | tee
123
fuwei@wenchang:~$ echo $?
2
fuwei@wenchang:~$
fuwei@wenchang:~$ # run a child bash to check inherit
fuwei@wenchang:~$ cat /tmp/testing.sh
#!/usr/bin/env bash

echo $SHELLOPTS

(echo 12345; exit 2) | tee
fuwei@wenchang:~$ /tmp/testing.sh
braceexpand:hashall:interactive-comments
12345
fuwei@wenchang:~$ echo $?
0
fuwei@wenchang:~$
fuwei@wenchang:~$ export SHELLOPTS
fuwei@wenchang:~$ # run again
fuwei@wenchang:~$ /tmp/testing.sh
braceexpand:emacs:hashall:histexpand:history:interactive-comments:monitor:pipefail
12345
fuwei@wenchang:~$ echo $?
2
fuwei@wenchang:~$
```


The previous test https://github.com/etcd-io/etcd/actions/runs/4453530207/jobs/7822105485 pass because the `test.sh` sets `set -e` and `pipefail` works in parent side. I think I should export `$SHELLOPTS` to ensure all the bash run with the setting. 

> SHELLOPTS
A colon-separated list of enabled shell options. Each word in the list is a valid argument for the -o option to the set builtin command (see [The Set Builtin](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)). The options appearing in SHELLOPTS are those reported as ‘on’ by ‘set -o’. If this variable is in the environment when Bash starts up, each shell option in the list will be enabled before reading any startup files. This variable is readonly.

Does it make senses?

REF: https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-SHELLOPTS

serathius: Adding `SHELLOPTS` will not work for local development. Could we start from adding `set -euo pipefail` to `./scripts/test.sh`?

fuweid: OK. Will update it.

scripts/release_mod.sh
```diff
@@ -10,7 +10,9 @@
 #
 # % DRY_RUN=false REMOTE_REPO="origin" ./scripts/release_mod.sh push_mod_tags
 
-set -e
+set -o pipefail
```
serathius, line 13: Any reason to do it differently?

serathius:
That's a lot of scripts changed, most of them are not tested on CI. Have you tested all of them manually? You don't need to migrate all scripts at once. We could split PR based on script type. Most important is `./scripts/tests.sh`

fuweid:
> Have you tested all of them manually? You don't need to migrate all scripts at once. We could split PR based on script type. Most important is ./scripts/tests.sh

Yes. I'm still testing it in my local. Will update it later.~

scripts/test.sh
```diff
@@ -170,8 +173,8 @@ function grpcproxy_e2e_pass {
 
 # Builds artifacts used by tests/e2e in coverage mode.
 function build_cov_pass {
-  run_for_module "server" run go test -tags cov -c -covermode=set -coverpkg="./..." -o "../bin/etcd_test"
-  run_for_module "etcdctl" run go test -tags cov -c -covermode=set -coverpkg="./..." -o "../bin/etcdctl_test"
+  run_for_module "server" run go test -tags cov -c -covermode=set -coverpkg="./..." -o "../bin/etcd_test" || return $?
```
fuweid, line 176: side note: fast fail return if there is any error during build

scripts/test_lib.sh
```diff
@@ -105,7 +105,7 @@ function relativePath {
 
 ####   Discovery of files/packages within a go module #####
 
-# go_srcs_in_module [package]
```
fuweid, line 108: side note:  **go_srcs_in_module** doesn't receive any input.

fuweid:
ping @ahrtr @serathius please take a look. The fail one is tracked by #15545

serathius (approved):
LGTM, what do you plan about TODOs left? Address in next PR? It's usually better to set TODOs on issues not specific people. This way they can be tracked.

scripts/test.sh
```diff
@@ -120,7 +123,7 @@ function integration_pass {
 
 function e2e_pass {
   # e2e tests are running pre-build binary. Settings like --race,-cover,-cpu does not have any impact.
-  run_for_module "tests" go_test "./e2e/..." "keep_going" : -timeout="${TIMEOUT:-30m}" "${RUN_ARG[@]}" "$@"
+  run_for_module "tests" go_test "./e2e/..." "keep_going" : -timeout="${TIMEOUT:-30m}" "${RUN_ARG[@]}" "$@" || return $?
```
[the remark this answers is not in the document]
ahrtr: > If not, test.sh just reports it fails but no error exit code.

Could you explain this? Isn't it a bug? The script should exit on error, because of `set -e`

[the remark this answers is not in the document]
ahrtr: Thanks for the clarification.

The `exit 255 ` will still exit the script?

serathius:
cc @ptabor @ahrtr

scripts/test.sh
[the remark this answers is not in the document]
ahrtr: Really good catch!! It works after updating your example like below. 

```
#!/usr/bin/env bash

set -euo pipefail

main() {
  command -v xx || return $? # should fail
  command -v ls
}

#echo $SHELLOPTS
if ! main; then
  echo oops
fi
echo $?

main
```

fuweid:
Thanks for the review. The TODO items will be handled in the follow-up.

PR #15313, opened 2023-02-14
workflows: Remove non-root gomod dependabot rules

serathius:
cc @ahrtr @ptabor @spzala 

Current configuration generates a PR per each package, even though all dependencies are also available in root go.mod.
By having dependabot only update root go.mod we will avoid a lot of spam and still be able to get all relevant dependency upgrade suggestions.

PR #15538, opened 2023-03-21
Bump some dependencies

ahrtr:
- Bump github.com/mgechev/revive to 1.3.1
- Bump honnef.co/go/tools to 0.4.3
- Bump golang.org/x/ to 0.7.0

cc  @serathius @spzala


chaochn47:
Maybe a numb question, what's the difference between this PR and https://github.com/etcd-io/etcd/blob/main/.github/dependabot.yml?

ahrtr:
> Maybe a dumb question, what's the difference between this PR and https://github.com/etcd-io/etcd/blob/main/.github/dependabot.yml?

Please read https://github.com/dependabot/dependabot-core/issues/6678

chaochn47:
Thanks @ahrtr for the pointer. I guess `go.mod` and `tools/mod/go.mod` will be updated weekly but the other module won't be automatically updated, so this PR is needed.

chaochn47 (approved):

ahrtr:
> but the other module won't be automatically updated

Other modules will also be automatically updated. Please read https://github.com/etcd-io/etcd/pull/15313

chaochn47:
I just get the validation of my understanding from #15531. Right now #15531 only updates root `go.mod` and is used as a notification to bump dependency version. 

So current PR #15538 (manual process) is needed. The underlying issue is mentioned in https://github.com/dependabot/dependabot-core/issues/6678.

Issue #15539, opened 2023-03-21
EtcdError is missing GRPCStatus that causes grpc.status.FromError to always return codes.Unknown


lavacat:
### What happened?

I've discovered this issue when debugging watch not retrying in [openWatchClient](https://github.com/etcd-io/etcd/blob/36fc3cae65cac1b844d2561471dc95b44753618d/client/v3/watch.go#L1040) 

isHaltErr was returning true on ErrLeaderChanged

This is only applicable when auth is enabled. 
[w.remote.Watch(w.ctx, w.callOpts...)](https://github.com/etcd-io/etcd/blob/36fc3cae65cac1b844d2561471dc95b44753618d/client/v3/watch.go#L1037) -> `streamClientInterceptor`-> [getToken](https://github.com/etcd-io/etcd/blob/b10ebff3bb0c337b3b153a39474e33fa9698790a/client/v3/retry_interceptor.go#L114) -> `Auth.Authenticate` -> `toErr`
`toErr` will convert ErrGRPCLeaderChanged to ErrLeaderChanged, `isHaltErr(ErrLeaderChanged)` returns `true`

When auth is disabled, `Auth.Authenticate` isn't called and `toErr` conversion doesn't happen.

### What did you expect to happen?

expected watch to retry

### How can we reproduce it (as minimally and precisely as possible)?

See unit and integration test in the [PR](https://github.com/etcd-io/etcd/pull/15540).

Note, test is using ErrGRPCNoLeader, but in production we've observed ErrLeaderChanged. 

### Anything else we need to know?

Another side effect of this issue is that prometheus interceptor will only increment Unknown grpc code in metrics on error.

I suspect this might also affect behavior of retry_interceptor and client/v3/leasing that retries on transient error.

### Etcd version (please run commands below)
etcd 3.4, 3.5 and main
<details>

```console
$ etcd --version
etcd Version: 3.6.0-alpha.0
Git SHA: f43dccb2f
Go Version: go1.19.2
Go OS/Arch: darwin/amd64

$ etcdctl version
etcdctl version: 3.6.0-alpha.0
API version: 3.6
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

[...]

PR #15540, opened 2023-03-21
v3rpc/rpctypes: added GRPCStatus for EtcdError

lavacat:
Problem: grpc status.FromError can't extract code and defaults to Unknown. This causes isHaltErr to always return true and prevent watch retry on Unavailable errors.

Solution: implement GRPCStatus() *Status for EtcdError

fixes: https://github.com/etcd-io/etcd/issues/15539

### DIFF
--- a/tests/common/watch_test.go
+++ b/tests/common/watch_test.go
@@ -19,6 +19,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/stretchr/testify/assert"
 
 	"go.etcd.io/etcd/tests/v3/framework/config"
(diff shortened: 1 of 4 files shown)


serathius:
cc @ptabor

serathius:
LGTM, but would like to double check all the code paths you mention in the issue before approving.

tests/common/watch_test.go
```diff
@@ -95,3 +97,42 @@ func TestWatch(t *testing.T) {
 		})
 	}
 }
+
+func TestWatchRetryNoLeader(t *testing.T) {
```
mitake, line 101: It might be valuable to have non auth version of this test for making sure that the non auth behavior is preserved. What do you think @lavacat ?

lavacat:
@serathius although `EtcdError` and `toErr()` function are used in lot of places in client, I think the areas we care about are those that compare `code` of the error. (see Sourcegraph search [results](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/etcd-io/etcd%24+file:%5Eclient/v3*+%3D+codes&patternType=standard&sm=0&groupBy=path)) 

I'll go through each search result and will report back. 
Right now I think [isSafeRetryImmutableRPC](https://sourcegraph.com/github.com/etcd-io/etcd/-/blob/client/v3/retry.go?L55) and [leasingKV.acquire ](https://sourcegraph.com/github.com/etcd-io/etcd/-/blob/client/v3/leasing/kv.go?L288) will change behavior based on this PR. Need to check why there were no test failures, possibly missing test case.

I'll also add non auth version of TestWatchRetryNoLeader per @mitake comment.

PR #14548, opened 2022-10-04
Backport PR 14322 to release-3.4

mitake:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

This PR backports https://github.com/etcd-io/etcd/pull/14322 to `release-3.4`. cc @ahrtr @spzala @kafuu-chino

PR #15542, opened 2023-03-21
[3.4] Revert 14548

mitake:

https://github.com/etcd-io/etcd/pull/14322 was backported to release-3.4 in https://github.com/etcd-io/etcd/pull/14548, but it should be reverted too.

cc @ahrtr @serathius

### DIFF
--- a/etcdserver/api/v3rpc/watch.go
+++ b/etcdserver/api/v3rpc/watch.go
@@ -234,16 +234,16 @@ func (ws *watchServer) Watch(stream pb.Watch_WatchServer) (err error) {
 	return err
 }
 
-func (sws *serverWatchStream) isWatchPermitted(wcr *pb.WatchCreateRequest) error {
+func (sws *serverWatchStream) isWatchPermitted(wcr *pb.WatchCreateRequest) bool {
 	authInfo, err := sws.ag.AuthInfoFromCtx(sws.gRPCStream.Context())
 	if err != nil {
-		return err
+		return false
 	}
 	if authInfo == nil {
 		// if auth is enabled, IsRangePermitted() can cause an error
 		authInfo = &auth.AuthInfo{}
 	}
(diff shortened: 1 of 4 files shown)


ahrtr:
@mitake  could you please also update the 3.4 changelog ?

ahrtr (approved):
LGTM

Thank you @mitake

mitake:
Thanks @ahrtr @lavacat . I'll open a PR for updating the changelog tomorrow.

etcdserver/api/v3rpc/watch.go
```diff
@@ -277,29 +277,13 @@ func (sws *serverWatchStream) recvLoop() error {
...
-				case auth.ErrUserEmpty:
-					cancelReason = rpctypes.ErrGRPCUserEmpty.Error()
-				default:
-					if err != auth.ErrPermissionDenied {
-						sws.lg.Error("unexpected error code", zap.Error(err))
-					}
-					cancelReason = rpctypes.ErrGRPCPermissionDenied.Error()
-				}
-
+			if !sws.isWatchPermitted(creq) {
```
ahrtr, line 280: I think we can keep the server side change?

mitake: ah sorry, let me open a follow up PR

mitake: On the second thought, I guess just reverting the entire change might be safer? What do you think?

ahrtr: Actually we kept the server side change when reverting the PR on 3.5.  Keeping the server side change can provide more useful error message (`cancelReason`).

But it isn't a big deal, it's up to you. thx

mitake: got it, opened https://github.com/etcd-io/etcd/pull/15584

PR #15561, opened 2023-03-27
Use makefile as entrypoint in CI

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/.github/workflows/e2e-arm64.yaml
+++ b/.github/workflows/e2e-arm64.yaml
@@ -28,7 +28,7 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-arm64-e2e)
-            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOOS=linux GOARCH=arm64 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e-release
             ;;
           *)
             echo "Failed to find target"
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -24,10 +24,10 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-amd64-e2e)
-            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOOS=linux GOARCH=amd64 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e-release
             ;;
           linux-386-e2e)
-            GOARCH=386 PASSES='build e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOOS=linux GOARCH=386 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e
             ;;
           *)
             echo "Failed to find target"
--- a/.github/workflows/tests-arm64.yaml
+++ b/.github/workflows/tests-arm64.yaml
@@ -31,16 +31,16 @@ jobs:
         export JUNIT_REPORT_DIR=$(realpath ${TARGET})
         case "${TARGET}" in
           linux-arm64-integration-1-cpu)
-            GOARCH=arm64 CPU=1 PASSES='integration' RACE='false' ./scripts/test.sh
+            GOOS=linux GOARCH=arm64 CPU=1 make test-integration
             ;;
           linux-arm64-integration-2-cpu)
-            GOARCH=arm64 CPU=2 PASSES='integration' RACE='false' ./scripts/test.sh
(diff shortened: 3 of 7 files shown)


.github/workflows/e2e.yaml
```diff
@@ -24,10 +24,10 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-amd64-e2e)
-            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOOS=linux GOARCH=arm64 EXPECT_DEBUG=true make test-e2e-release
```
jmhbnz, line 27: I might be missing something but it looks like this needs to be `GOARCH=amd64`?

jmhbnz (approved):
LGTM

.github/workflows/e2e-arm64.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-arm64-e2e)
-            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOARCH=arm64 EXPECT_DEBUG=true make test-e2e-release
```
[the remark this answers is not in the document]
serathius: Let's leave those variables. Thanks for bringing it up!

.github/workflows/tests.yaml
```diff
@@ -28,19 +28,19 @@ jobs:
...
-            GOARCH=amd64 CPU=2 PASSES='integration' RACE='false' ./scripts/test.sh
+            GOOS=linux GOARCH=amd64 CPU=2 make test-integration
             ;;
           linux-amd64-integration-4-cpu)
-            GOARCH=amd64 CPU=4 PASSES='integration' RACE='false' ./scripts/test.sh
+            GOOS=linux GOARCH=amd64 CPU=4 make test-integration
             ;;
           linux-amd64-unit-4-cpu-race)
-            GOARCH=amd64 PASSES='unit' RACE='true' CPU='4' ./scripts/test.sh -p=2
+            GOOS=linux GOARCH=amd64 CPU=4 RACE=true make test-unit
```
chaochn47, line 40:
```suggestion
            GO_TEST_FLAGS='-p=2' GOOS=linux GOARCH=amd64 CPU=4 RACE=true make test-unit
```

.github/workflows/tests-arm64.yaml
```diff
@@ -31,16 +31,16 @@ jobs:
...
-            GOARCH=arm64 CPU=2 PASSES='integration' RACE='false' ./scripts/test.sh
+            GOOS=linux GOARCH=arm64 CPU=2 make test-integration
             ;;
           linux-arm64-integration-4-cpu)
-            GOARCH=arm64 CPU=4 PASSES='integration' RACE='false' ./scripts/test.sh
+            GOOS=linux GOARCH=arm64 CPU=4 make test-integration
             ;;
           linux-arm64-unit-4-cpu-race)
-            GOARCH=arm64 PASSES='unit' RACE='true' CPU='4' ./scripts/test.sh -p=2
+            GOOS=linux GOARCH=arm64 CPU=4 RACE=true make test-unit
```
chaochn47, line 43:
```suggestion
            GO_TEST_FLAGS='-p=2' GOOS=linux GOARCH=arm64 CPU=4 RACE=true make test-unit
```

chaochn47 (approved):

serathius:
PTAL @ahrtr @ptabor

.github/workflows/e2e.yaml
```diff
@@ -24,10 +24,10 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-amd64-e2e)
-            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOOS=linux GOARCH=amd64 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e-release
             ;;
           linux-386-e2e)
-            GOARCH=386 PASSES='build e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./scripts/test.sh
+            GOOS=linux GOARCH=386 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e
```
ahrtr, line 30: Why the target for `linux-386-e2e` is `test-e2e`, but it's `test-e2e-release` for `linux-amd64-e2e`? Shouldn't they be both `test-e2e`?

serathius: Because on `linux-amd64-e2e` we also test upgrades by `PASSES='build release e2e'`. We should test upgrades on at least one target, however  maybe we can update the name to reflect that.

ahrtr (approved):

PR #15562, opened 2023-03-27
[3.4] tests: Avoid testing package root tests in e2e 

serathius:
Changes invocation from `go test -timeout 30m -v -cpu 1,2,4 '' -v --count 1 go.etcd.io/etcd/tests/e2e` to `go test -timeout 30m -v -cpu 1,2,4 -v --count 1 go.etcd.io/etcd/tests/e2e` (removes ''). Those braces caused tests run in root package instead of e2e.

Logs from a failed run https://github.com/etcd-io/etcd/actions/runs/4533349980/jobs/7986037591:
```
=== RUN   TestMain
    main_test.go:32: skip launching etcd server when invoked via go test
--- SKIP: TestMain (0.00s)
=== RUN   TestMain
    main_test.go:32: skip launching etcd server when invoked via go test
--- SKIP: TestMain (0.00s)
=== RUN   TestMain
    main_test.go:32: skip launching etcd server when invoked via go test
--- SKIP: TestMain (0.00s)
PASS
ok  	go.etcd.io/etcd	0.020s
Error: Process completed with exit code 143.
```
cc @ahrtr @ptabor

### DIFF
--- a/test
+++ b/test
@@ -340,7 +340,7 @@ function e2e_pass {
 		USERTIMEOUT="${TIMEOUT}"
 	fi
 
-	go test -timeout "${USERTIMEOUT}" -v -cpu "${TEST_CPUS}" "${RUN_ARG}"  "$@" "${REPO_PATH}/tests/e2e"
+	go test -timeout "${USERTIMEOUT}" -v -cpu "${TEST_CPUS}" ${RUN_ARG}  "$@" "${REPO_PATH}/tests/e2e"
 }
 
 function integration_e2e_pass {


fuweid:
I didn't reproduce it in my local. 143 - 128 = 15 sigterm maybe from github action.
I run into this issue in https://github.com/etcd-io/bbolt/pull/373#discussion_r1060641823. It is caused by high load and killed by github action node-agent.

REF: https://github.com/actions/runner-images/issues/6709

ahrtr:
curious why we do not see such error on [the head of release-3.4](https://github.com/etcd-io/etcd/actions/runs/4531983861/jobs/7982796629)?

serathius:
Ok, I think I got it. There are two problems:
* running e2e tests against root path, which results in confusing results like above. Causes log `ok  	go.etcd.io/etcd	0.020s`
* running e2e tests without `-v` flag which results in no logs when test is killed by runner. Causes log `Error: Process completed with exit code 143.`

I think we should solve both.

serathius:
I think the problem is that running tests in root package results in test logs not being flushed properly. They are only flushed when tests finishes. So if the test is killed by GitHub as @fuweid mentioned, it causes logs to be missing.

On my local machine change from this PR results in logs being properly flushed, instead of flushed on the end.
@fuweid please confirm if it's the same for you.

serathius:
I still think it's worth merging this PR as without it, it's impossible to debug runs like https://github.com/etcd-io/etcd/actions/runs/4533349980/jobs/7986037591

ahrtr (approved):

fuweid (approved):
LGTM(non-binding)

Would you mind to update the pr title with prefix `[3.4]`? It's easy to review during release. Thanks!

fuweid:
> please confirm if it's the same for you.

Yes. Thanks for the patch!

Issue #15330, opened 2023-02-16
Linearizability test timesout due to large number failed requests when test finishes


serathius:
### Which github workflows are flaking?

Linearizability Nightly

### Which tests are flaking?

Linearizability/ClusterOfSize1/LowTraffic

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4192533357

### Reason for failure (if possible)

Last 10 request in the traffic fail due to context timeout. Possibly non of them are persisted so we don't get watch event and cannot confirm if we can just ignore them. 

![image](https://user-images.githubusercontent.com/5873433/219350884-0f902e58-ca1e-4660-983d-5e67d88d4e93.png)

Failed requests don't have specific order, so running linearization on them results in exponential complexity. As result linearization times out. 

### Anything else we need to know?

cc @ahrtr @ptabor @lavacat

I think we can try just pruning failed request from the end of history ordered by request start time.

serathius:
Addressed this

PR #15575, opened 2023-03-27
tests: Ensure watch catches all events generated in traffic

serathius:
Ref https://github.com/etcd-io/etcd/issues/15330
cc @ahrtr @ptabor 

Might need to cleanup code


tests/robustness/watch.go
```diff
@@ -45,35 +46,59 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+	defer cancel()
+	watch := c.Watch(ctx, "", clientv3.WithPrefix(), clientv3.WithRev(lastRevision+1), clientv3.WithProgressNotify())
 	for {
 		select {
 		case <-ctx.Done():
 			return resps
-		default:
-		}
-		for resp := range c.Watch(ctx, "", clientv3.WithPrefix(), clientv3.WithRev(lastRevision+1), clientv3.WithProgressNotify()) {
+		case maxRevision = <-maxRevisionChan:
```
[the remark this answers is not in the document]
ptabor: After reread it seems we support only 1 value in the maxRevisions. I would assert it explicitly here then.

tests/robustness/watch.go
```diff
@@ -45,35 +46,59 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+			if lastRevision >= maxRevision {
+				cancel()
+			}
+		case resp := <-watch:
 			resps = append(resps, watchResponse{resp, time.Now()})
```
ptabor, line 88: Shell we have a case if resp is nil (it is watch stream got terminated on the server or watch stream ctx got cancelled ) ?

tests/robustness/linearizability_test.go
```diff
@@ -202,15 +202,22 @@ func runScenario(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.Et
...
+			if maxRevision > revision {
+				maxRevision = revision
+			}
+		}
+		maxRevisionChan <- maxRevision
 		return nil
 	})
 	g.Go(func() error {
-		responses = collectClusterWatchEvents(watchCtx, t, lg, clus)
+		responses = collectClusterWatchEvents(t, lg, clus, maxRevisionChan)
```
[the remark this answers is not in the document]
serathius: Done

tests/robustness/watch.go
```diff
@@ -45,35 +46,59 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+		for _, memberChan := range memberMaxRevisionChans {
+			memberChan <- maxRevision
+		}
+	}()
 	wg.Wait()
 	return memberResponses
 }
 
-func watchMember(ctx context.Context, lg *zap.Logger, c *clientv3.Client) (resps []watchResponse) {
+func watchMember(lg *zap.Logger, c *clientv3.Client, maxRevisionChan <-chan int64) (resps []watchResponse) {
```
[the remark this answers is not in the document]
serathius: Done

tests/robustness/watch.go
serathius: resp is not a pointer so we will never get an nil. In cases you mentioned I expect resp will hold an error. In such case we report it in logs. Still I don't see reason why not append it. This function should collect all responses so during validation step we can check it.

tests/robustness/watch.go
```diff
@@ -45,35 +46,59 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
-			lastRevision = resp.Header.Revision
+			if resp.IsProgressNotify() {
+				lastRevision = resp.Header.Revision
+			} else {
+				lastRevision = resp.Events[len(resp.Events)-1].Kv.ModRevision
+			}
 			if resp.Err() != nil {
 				lg.Info("Watch error", zap.Error(resp.Err()))
 			}
+			if maxRevision != 0 {
```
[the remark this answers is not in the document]
serathius: Done

tests/robustness/watch.go
```diff
@@ -45,35 +46,59 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+	}()
 	wg.Wait()
 	return memberResponses
 }
 
-func watchMember(ctx context.Context, lg *zap.Logger, c *clientv3.Client) (resps []watchResponse) {
+func watchMember(lg *zap.Logger, c *clientv3.Client, maxRevisionChan <-chan int64) (resps []watchResponse) {
 	var lastRevision int64 = 0
+	var maxRevision int64 = 0
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
serathius: Done

serathius:
PTAL @ptabor

tests/robustness/linearizability_test.go
```diff
@@ -202,21 +202,32 @@ func runScenario(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.Et
...
 	})
 	g.Wait()
 	return operations, responses
 }
 
+func operationsMaxRevision(operations []porcupine.Operation) int64 {
+	var maxRevision int64
+	for _, op := range operations {
+		revision := op.Output.(model.EtcdResponse).Revision
+		if maxRevision > revision {
```
ahrtr, line 224: typo? 
```suggestion
		if maxRevision < revision {
```

tests/robustness/watch.go
```diff
@@ -45,37 +46,79 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+		}
+	}()
 	wg.Wait()
 	return memberResponses
 }
 
-func watchMember(ctx context.Context, lg *zap.Logger, c *clientv3.Client) (resps []watchResponse) {
+// watchMember collects all responses until context is cancelled or has observed revision provided via maxRevisionChan.
+func watchMember(ctx context.Context, t *testing.T, c *clientv3.Client, maxRevisionChan <-chan int64) (resps []watchResponse) {
+	var maxRevision int64 = 0
```
chaochn47, line 75: I think rename `maxRevision` to `maxRevisionFromTrafficRecording` or something else will help improve code readability.

tests/robustness/watch.go
```diff
@@ -45,37 +46,79 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+		case maxRevision = <-maxRevisionChan:
+			if lastRevision >= maxRevision {
+				cancel()
+			}
+		case resp := <-watch:
+			if resp.Err() == nil {
+				resps = append(resps, watchResponse{resp, time.Now()})
+			} else if !resp.Canceled {
+				t.Errorf("Watch stream received error, err %v", resp.Err())
+			}
```
chaochn47, line 100: Theoretically, after `cancel()`, will `resp := <-watch` wins against `case <-ctx.Done()`  and error out due to `resp.Canceled = true`?

tests/robustness/watch.go
```diff
@@ -45,37 +46,79 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
+	watch := c.Watch(ctx, "", clientv3.WithPrefix(), clientv3.WithRev(1), clientv3.WithProgressNotify())
 	for {
 		select {
 		case <-ctx.Done():
+			revision := watchResponsesMaxRevision(resps)
+			if maxRevision == 0 {
+				t.Errorf("Client didn't collect all events, max revision not set")
+			}
+			if revision < maxRevision {
+				t.Errorf("Client didn't collect all events, revision got %d, expected: %d", revision, maxRevision)
```
chaochn47, line 88:
```suggestion
				t.Errorf("Watch client didn't collect all events, revision got %d, expected: %d", revision, maxRevision)
```

tests/robustness/watch.go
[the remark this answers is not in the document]
chaochn47: I feel like it is more natural to put `if maxRevision == 0 and then error out` logic in the receiver side.

chaochn47:
LGTM. Thanks @serathius 

I put some minor comments. From deflake test perspective, could you please take a look at this [comment](https://github.com/etcd-io/etcd/pull/15575#discussion_r1152264765) especially?

tests/robustness/watch.go
serathius: This is an independent function and it should not be aware who writes to maxRevisionChan. As defined in function doc, it collects all events until maxRevision is provided and observed on watch. No need to push name `traffic` everywhere.

tests/robustness/watch.go
```diff
@@ -45,37 +46,79 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, lg *zap.Logger
...
 	var lastRevision int64 = 0
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	watch := c.Watch(ctx, "", clientv3.WithPrefix(), clientv3.WithRev(1), clientv3.WithProgressNotify())
 	for {
 		select {
 		case <-ctx.Done():
+			revision := watchResponsesMaxRevision(resps)
+			if maxRevision == 0 {
+				t.Errorf("Client didn't collect all events, max revision not set")
```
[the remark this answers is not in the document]
serathius: This error doesn't describe traffic collection issue, but situation when context provided as function argument is cancelled. This function is not aware who owns or can cancel the ctx, so note about traffic is incorrect. Client mentioned in message here is the local client, not the traffic recorder.

tests/robustness/watch.go
serathius: `client` is the `watch client`, also `c *clientv3.Client`  is an function argument which this refers to, so naming it `watch client` will break this connection.

tests/robustness/watch.go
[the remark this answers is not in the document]
serathius: `0` is not correct revision so I never expect it will be written to maxRevisionChan. If it is, function will exit via `context.Done()`. Having one return point via context is clearer.

tests/robustness/watch.go
serathius: On `cancel()` I expect that watch will get a response with `closeErr` equal `cancelled` and `Cancelled` equal `true`.
 So I expect this if statement to not error as we only error on `!resp.Canceled` so cancel will never cause error.

tests/robustness/watch.go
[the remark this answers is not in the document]
chaochn47: Make sense. Thanks for explanations!!

tests/robustness/watch.go
chaochn47: Thanks! I must misread the `if !resp.Canceled` condition check in the opposite way.

chaochn47 (approved):

chaochn47:
LGTM. Thanks @serathius!!

ptabor (approved):

PR #15577, opened 2023-03-27
tests: Add new test for round robin resolver

jmhbnz:
The [default grpc-go load balancing strategy is pick first](https://github.com/grpc/grpc-go/issues/5009) which may not be expected, particularly for production scenarios. We should consider adding a test for round robin as an alternative.

Refer: https://github.com/etcd-io/website/pull/633

### DIFF
--- a/pkg/grpc_testing/stub_server.go
+++ b/pkg/grpc_testing/stub_server.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
+	"sync/atomic"
 
 	"google.golang.org/grpc"
 	testpb "google.golang.org/grpc/test/grpc_testing"
@@ -93,14 +95,16 @@ func (ss *StubServer) Addr() string {
 
 type dummyStubServer struct {
 	testpb.UnimplementedTestServiceServer
-	body []byte
+	counter uint64
 }
 
-func (d dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+func (d *dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+	newCount := atomic.AddUint64(&d.counter, 1)
+
 	return &testpb.SimpleResponse{
 		Payload: &testpb.Payload{
 			Type: testpb.PayloadType_COMPRESSABLE,
-			Body: d.body,
+			Body: []byte(strconv.FormatUint(newCount, 10)),
 		},
 	}, nil
 }
@@ -108,5 +112,5 @@ func (d dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*tes
 // NewDummyStubServer creates a simple test server that serves Unary calls with
 // responses with the given payload.
 func NewDummyStubServer(body []byte) *StubServer {
-	return New(dummyStubServer{body: body})
+	return New(&dummyStubServer{})
 }
--- a/tests/integration/clientv3/naming/resolver_test.go
+++ b/tests/integration/clientv3/naming/resolver_test.go
@@ -17,8 +17,9 @@ package naming_test
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"strconv"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"go.etcd.io/etcd/client/v3/naming/endpoints"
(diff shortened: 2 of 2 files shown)


jmhbnz:
Useful context on grpc-go resolver within etcd: https://github.com/etcd-io/etcd/issues/15145 

This pr may not be required based on the above issue, or alternatively be reverted at a later date once a new approach is merged.

tests/integration/clientv3/naming/resolver_test.go
```diff
@@ -113,6 +113,91 @@ func TestEtcdGrpcResolver(t *testing.T) {
...
+		}
+		if resp.GetPayload() == nil || !bytes.Equal(resp.GetPayload().GetBody(), s2PayloadBody) {
+			if retries < maxRetries {
+				continue
+			}
+			t.Fatalf("unexpected response from foo (e2): %s", resp.GetPayload().GetBody())
+		}
+		break
+	}
+}
```
ptabor, line 199: Could you please verify here that both services got >0 number of requests. 
Probably this requires expanding dummyStubServer to have a counter of number of invocations. 
I think that even each request can return the state of atomic counter in the response.

ptabor (commented):
Thank you. Looks good. But I think that the number-of-calls assertion is crucial to verify it's really RR. 

Probably most of the code could be shared with the previous test... just make the 'policy' options a parameter and post-run verification.

tests/integration/clientv3/naming/resolver_test.go
jmhbnz: Great suggestion @ptabor.  I've made an attempt at this as well as trying to re-use one test just with parameter for load balancing policy.  Can you please take another look and let me know what you think?

pkg/grpc_testing/stub_server.go
```diff
@@ -93,20 +94,27 @@ func (ss *StubServer) Addr() string {
...
 type dummyStubServer struct {
 	testpb.UnimplementedTestServiceServer
-	body []byte
+	body    []byte
+	counter int
 }
 
-func (d dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+func (d *dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+	d.counter++
```
tjungblu, line 102: don't know if we do parallel requests in our testing, but maybe do an atomic increment? that way people don't have to think about this going forward

tests/integration/clientv3/naming/resolver_test.go
```diff
@@ -64,55 +63,93 @@ func TestEtcdGrpcResolver(t *testing.T) {
...
 		}
-		if resp.GetPayload() == nil || !bytes.Equal(resp.GetPayload().GetBody(), s2PayloadBody) {
-			if retries < maxRetries {
-				continue
-			}
-			t.Fatalf("unexpected response from foo (e2): %s", resp.GetPayload().GetBody())
+
+		// Allow 10% tolerance a round robin is not perfect and we don't want the test to flake
+		threshold := float64(totalRequests) * 0.5 * 0.9
+		if responses < int(threshold) {
```
tjungblu, line 129: stretchr has a nice function for deltas:

> assert.InDelta(t, math.Pi, 22/7.0, 0.01)

in case you want to make use of the upper boundary too.

jmhbnz: Great suggestion, added upper bound via stretchr, thanks! 🙏🏻

pkg/grpc_testing/stub_server.go
jmhbnz: Makes sense, updated :)

pkg/grpc_testing/stub_server.go
```diff
@@ -93,20 +95,23 @@ func (ss *StubServer) Addr() string {
...
 
-func (d dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+func (d *dummyStubServer) UnaryCall(context.Context, *testpb.SimpleRequest) (*testpb.SimpleResponse, error) {
+	newCount := atomic.AddUint64(&d.counter, 1)
+
 	return &testpb.SimpleResponse{
 		Payload: &testpb.Payload{
 			Type: testpb.PayloadType_COMPRESSABLE,
-			Body: d.body,
+			Body: []byte(strconv.FormatUint(newCount, 10)),
```
ahrtr, line 108: It seems that the field `dummyStubServer.body` isn't used at all? If yes, then I suggest to remove it.

tests/integration/clientv3/naming/resolver_test.go
```diff
@@ -64,55 +63,91 @@ func TestEtcdGrpcResolver(t *testing.T) {
...
-			t.Fatal("failed to invoke rpc to foo (e2)", err)
+		if resp.GetPayload() == nil {
+			t.Fatalf("unexpected response from foo: %s", resp.GetPayload().GetBody())
+		}
+		lastResponse = resp.GetPayload().GetBody()
+	}
+
+	// If the load balancing policy is pick first then return payload should equal number of requests
+	if lbPolicy == "pick_first" {
+		t.Logf("Last response: %v", string(lastResponse))
```
ahrtr, line 113: minor comment: there are two identical logs (this line and line 121, suggest to move it above the `if statement` (e.g line 111.)

ahrtr:
Please also rebase this PR.

tests/integration/clientv3/naming/resolver_test.go
jmhbnz: Makes sense, done!

pkg/grpc_testing/stub_server.go
jmhbnz: Good catch, updated.  It was a little while back I worked on this but I think I was nervous about changing as I wasn't sure if there was going to be other reuse that depended on it.  After looking closer it seems ok to remove.

ahrtr (approved):
LGTM

Thanks @jmhbnz

PR #15578, opened 2023-03-28
migrate auth tests to common #7

chaochn47:
Related to the ask from https://github.com/etcd-io/etcd/pull/14041 to split tests into smaller pieces.

So code review will be reviewer friendly and easier to reason about.

### DIFF
--- a/tests/common/auth_test.go
+++ b/tests/common/auth_test.go
@@ -629,6 +629,99 @@ func TestAuthTestInvalidMgmt(t *testing.T) {
 	})
 }
 
+func TestAuthLeaseRevoke(t *testing.T) {
+	testRunner.BeforeTest(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	clus := testRunner.NewCluster(ctx, t, config.WithClusterConfig(config.ClusterConfig{ClusterSize: 1}))
+	defer clus.Close()
+	cc := testutils.MustClient(clus.Client())
+	testutils.ExecuteUntil(ctx, t, func() {
+		require.NoErrorf(t, setupAuth(cc, []authRole{testRole}, []authUser{rootUser, testUser}), "failed to enable auth")
+		rootAuthClient := testutils.MustClient(clus.Client(WithAuth(rootUserName, rootPassword)))
+
+		lresp, err := rootAuthClient.Grant(ctx, 10)
+		require.NoError(t, err)
+		err = rootAuthClient.Put(ctx, "key", "value", config.PutOptions{LeaseID: lresp.ID})
+		require.NoError(t, err)
+
+		_, err = rootAuthClient.Revoke(ctx, lresp.ID)
+		require.NoError(t, err)
+
+		_, err = rootAuthClient.Get(ctx, "key", config.GetOptions{})
+		require.NoError(t, err)
+	})
+}
+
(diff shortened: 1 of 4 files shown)


tests/common/auth_test.go
```diff
@@ -629,6 +629,99 @@ func TestAuthTestInvalidMgmt(t *testing.T) {
...
+	testutils.ExecuteUntil(ctx, t, func() {
+		require.NoErrorf(t, setupAuth(cc, []authRole{testRole}, []authUser{rootUser, testUser}), "failed to enable auth")
+		rootAuthClient := testutils.MustClient(clus.Client(WithAuth(rootUserName, rootPassword)))
+
+		lresp, err := rootAuthClient.Grant(ctx, 10)
+		require.NoError(t, err)
+		err = rootAuthClient.Put(ctx, "key", "value", config.PutOptions{LeaseID: lresp.ID})
+		require.NoError(t, err)
+
+		_, err = rootAuthClient.Revoke(ctx, lresp.ID)
```
ptabor, line 648: If we have a test targeting this method specifically, it would be good to assert what's returning actually. 

[I know that it was not tested previously... ]

tests/common/auth_test.go
```diff
@@ -629,6 +629,99 @@ func TestAuthTestInvalidMgmt(t *testing.T) {
...
+func TestAuthUserGet(t *testing.T) {
+	testRunner.BeforeTest(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	clus := testRunner.NewCluster(ctx, t, config.WithClusterConfig(config.ClusterConfig{ClusterSize: 1}))
+	defer clus.Close()
+	cc := testutils.MustClient(clus.Client())
+	testutils.ExecuteUntil(ctx, t, func() {
+		require.NoErrorf(t, setupAuth(cc, []authRole{testRole}, []authUser{rootUser, testUser}), "failed to enable auth")
+		rootAuthClient := testutils.MustClient(clus.Client(WithAuth(rootUserName, rootPassword)))
```
ptabor, line 691: Logically this are 2 subtests, so ideally we should run them in tested t.run(...).

ptabor (approved):
Thank you. 2 comments, but not blocking as already the code is a big improvement.

tests/common/auth_test.go
chaochn47: clientv3 Lease revoke response is the type alias of the proto buf generated one. It only contains `ResponseHeader`. 

```
type LeaseRevokeResponse struct {
	Header               *ResponseHeader `protobuf:"bytes,1,opt,name=header,proto3" json:"header,omitempty"`
	XXX_NoUnkeyedLiteral struct{}        `json:"-"`
	XXX_unrecognized     []byte          `json:"-"`
	XXX_sizecache        int32           `json:"-"`
}
```

I was blindly migrate the old logic to this one. How about assert the revoked lease is no longer in LeaseList response and get response is empty?

tests/common/auth_test.go
chaochn47: Could you please point to which 2 tests? I guess root and test user, respectively?

ahrtr (approved):
LGTM

Thank you @chaochn47

tests/common/auth_test.go
ahrtr: It seems not a big deal to me.

PR #15603, opened 2023-03-31
tests: Ensure that operation history finishes with successful request

serathius:
Should help with https://github.com/etcd-io/etcd/issues/15330
cc @ahrtr @ptabor

### DIFF
--- a/tests/robustness/traffic.go
+++ b/tests/robustness/traffic.go
@@ -54,7 +54,7 @@ const (
 	Defragment    TrafficRequestType = "defragment"
 )
 
-func simulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, config trafficConfig) []porcupine.Operation {
+func simulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, config trafficConfig, finish <-chan struct{}) []porcupine.Operation {
 	mux := sync.Mutex{}
 	endpoints := clus.EndpointsV3()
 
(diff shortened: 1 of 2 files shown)


tests/robustness/traffic.go
```diff
@@ -64,26 +64,39 @@ func simulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
...
 			mux.Lock()
 			h = h.Merge(c.history.History)
 			mux.Unlock()
 		}(c, i)
 	}
 	wg.Wait()
 	endTime := time.Now()
+
+	// Ensure that last operation is succeeds
+	time.Sleep(time.Second)
```
fuweid, line 93: We already have `wg.Wait()` to ensure traffic.run to finish. Is there any reason to wait 1 second here?

fuweid (approved):
LGTM(non-binding)

tests/robustness/traffic.go
serathius: Same us before, we require some time without traffic to get periodic watch notification.

serathius:
This is required to avoid flakes in periodic tests.

tests/robustness/traffic.go
```diff
@@ -64,26 +64,39 @@ func simulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
 	limiter := rate.NewLimiter(rate.Limit(config.maximalQPS), 200)
 
 	startTime := time.Now()
+	c, err := NewClient(endpoints, ids, startTime)
```
ahrtr, line 67: Minor comment: It isn't a problem to define the same variable name (line 67 and line 75), but please give it another name, just in case the following contributors accidentally change the `:=` to `=` at line 75.
```suggestion
	cc, err := NewClient(endpoints, ids, startTime)
```

ahrtr (approved):
LGTM with a minor comment. thx

chaochn47 (approved):

tests/robustness/traffic.go
fuweid: Sounds good to me. When the last request is timeout and the server already received the request, it might need time to handle it. Thanks for the comment.

tests/robustness/traffic.go
serathius: Done

Issue #15596, opened 2023-03-30
Ensure healthy cluster after robustness failpoint


serathius:
### What would you like to be added?

Recent change that wait for all watch events caused nightly robustness tests runs  to wait infinitely https://github.com/etcd-io/etcd/actions/runs/4562802681

Bug in raft https://github.com/etcd-io/etcd/issues/15595 causes etcd members to crash. With only 1 out of three members cluster is not able to proceed.

In this case instead of waiting infinity we should just detect that cluster is unhealthy and abort collecting watch.
Bonus points for implementing timeout for watching on events

### Why is this needed?

Robustness tests should not timeout

jmhbnz:
Hey @serathius - Had a quick look at this to try and understand how these tests work, a lot to unpick! Am I on the right track that this function is responsible for the infinite wait? https://github.com/etcd-io/etcd/blob/main/tests/robustness/watch.go#L34

If I'm on the right track feel free to assign this one to me and I will have a go at raising the fix 😅

serathius:
Yes, this is the correct function. It is infinite due to recent change in https://github.com/etcd-io/etcd/pull/15575.

However the function by itself is not the problem. I think we should have an external mechanism that checks cluster is totally down and cancels the context passed to `collectClusterWatchEvents` and `watchMember`.

My first guess would be that [triggerFailpoints](https://github.com/etcd-io/etcd/blob/e11a32366e05cc60e95a4e818f6745aa3113aaaf/tests/robustness/failpoints.go#L86-L110) function should validate that cluster is healthy between and after injecting failpoint. If it's not it should propagate the signal up as an error. With error the [runScenario](https://github.com/etcd-io/etcd/blob/e11a32366e05cc60e95a4e818f6745aa3113aaaf/tests/robustness/linearizability_test.go#L195-L218) function can cancel the watch.

Issue #15595, opened 2023-03-30
Etcd panic in robustness tests blackhole failpoint


serathius:
### What happened?

Robustness Nightly are failing due to etcd panic.
https://github.com/etcd-io/etcd/actions/runs/4562802681

### What did you expect to happen?

Etcd should not panic due to network issues

### How can we reproduce it (as minimally and precisely as possible)?

Run robustness tests blackhole tests enough times.
See https://github.com/etcd-io/etcd/actions/workflows/robustness-nightly.yaml

### Etcd version (please run commands below)

N/A

### Etcd configuration (command line flags or environment variables)

N/A

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

N/A

### Relevant log output

```shell
2023-03-30T10:25:27.3144193Z     logger.go:130: 2023-03-30T10:25:26.312Z	INFO	Triggering failpoint
2023-03-30T10:25:27.3144599Z         	{"failpoint": "blackhole"}
2023-03-30T10:25:27.3145371Z     failpoints.go:328: Blackholing traffic from and to member "TestRobustnessClusterOfSize3HighTraffic-test-1"
2023-03-30T10:25:27.3146769Z /home/runner/work/etcd/etcd/bin/etcd (TestRobustnessClusterOfSize3HighTraffic-test-2) (28025): {"level":"warn","ts":"2023-03-30T10:25:26.811397Z","caller":"etcdserver/v3_server.go:831","msg":"waiting for ReadIndex response took too long, retrying","sent-request-id":11854185818228622438,"retry-timeout":"500ms"}
2023-03-30T10:25:27.3148279Z /home/runner/work/etcd/etcd/bin/etcd (TestRobustnessClusterOfSize3HighTraffic-test-0) (28026): {"level":"warn","ts":"2023-03-30T10:25:26.812944Z","caller":"etcdserver/v3_server.go:831","msg":"waiting for ReadIndex response took too long, retrying","sent-request-id":14186768950229826371,"retry-timeout":"500ms"}
2023-03-30T10:25:27.3149756Z /home/runner/work/etcd/etcd/bin/etcd (TestRobustnessClusterOfSize3HighTraffic-test-1) (28035): {"level":"warn","ts":"2023-03-30T10:25:26.813508Z","caller":"etcdserver/v3_server.go:831","msg":"waiting for ReadIndex response took too long, retrying","sent-request-id":6108155643657295548,"retry-timeout":"500ms"}
[...]

PR #15604, opened 2023-03-31
Ensure healthy cluster after robustness failpoint

jmhbnz:
We need a way to verify if the cluster is healthy before and after injecting failpoints in robustness tests so we can surface these errors and ensure watch does not wait indefinitely causing the robustness suite to fail.


tests/robustness/failpoints.go
```diff
@@ -85,30 +85,59 @@ var (
...
 )
 
-func triggerFailpoints(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, config FailpointConfig) {
+func verifyClusterHealth(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, failpointName string) error {
+	for i := 0; i < len(clus.Procs); i++ {
+		resp, err := clus.Client().Status(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to retrieve cluster health, err: %v", err)
+		}
+		for _, memberError := range resp[i].Errors {
```
jmhbnz, line 94: Hey @serathius - This is the part of what I have proposed that I am the most uncertain about, can you please provide some guidance on if this is even a reasonable way to check for cluster health?  I still have much to learn here 😕

tests/robustness/linearizability_test.go
```diff
@@ -197,8 +197,12 @@ func runScenario(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.Et
 	// Run multiple test components (traffic, failpoints, etc) in parallel and use canceling context to propagate stop signal.
 	g := errgroup.Group{}
 	trafficCtx, trafficCancel := context.WithCancel(ctx)
+	failpointCtx, failpointCancel := context.WithCancel(ctx)
```
serathius, line 200:
```suggestion
	watchCtx, watchCancel := context.WithCancel(ctx)
```
Even though failpoints will cancel this, the context and cancel impacts watch.

tests/robustness/failpoints.go
```diff
@@ -85,30 +85,59 @@ var (
 	}}
 )
 
-func triggerFailpoints(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, config FailpointConfig) {
+func verifyClusterHealth(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, failpointName string) error {
+	for i := 0; i < len(clus.Procs); i++ {
+		resp, err := clus.Client().Status(ctx)
```
serathius, line 90: Please use client library, and not cluster client which runs a separate process.
```
		Endpoints:            member.EndpointsV3(),
		DialKeepAliveTime:    1 * time.Millisecond,
		DialKeepAliveTimeout: 5 * time.Millisecond,
	})
```

tests/robustness/failpoints.go
serathius: One way to do that is check health of each cluster member by sending health request to each member.

tests/robustness/linearizability_test.go
jmhbnz: Good point, thanks!

tests/robustness/failpoints.go
```diff
@@ -85,30 +88,87 @@ var (
...
+	for i := 0; i < len(clus.Procs); i++ {
+		clusterClient, err := clientv3.New(clientv3.Config{
+			Endpoints:            clus.Procs[i].EndpointsGRPC(),
+			Logger:               zap.NewNop(),
+			DialKeepAliveTime:    1 * time.Millisecond,
+			DialKeepAliveTimeout: 5 * time.Millisecond,
+		})
+		defer clusterClient.Close()
+
+		// Attempt to check status for the member
```
serathius, line 101: No need to check both status and health. Just health is enough.

tests/robustness/failpoints.go
```diff
@@ -85,30 +88,87 @@ var (
...
+		if err != nil {
+			return fmt.Errorf("failed to get cluster member status, err: %v", err)
+		}
+		for _, statusError := range statusResp.Errors {
+			return fmt.Errorf("cluster member status contains error, err: %v", statusError)
+		}
+
+		// Also check member /health endpoint
+		httpClient := &http.Client{}
+		healthResp, err := httpClient.Get(fmt.Sprintf("%s/health", clusterClient.Endpoints()[0]))
```
serathius, line 112: For checking health I meant it via grpc endpoint like https://github.com/etcd-io/etcd/blob/da4bf0f76fb708e0b57763edb46ba523447b9510/tests/framework/integration/integration.go#L235-L245

jmhbnz: My misunderstanding, fixed 🤦🏻‍♂️

tests/robustness/failpoints.go
```diff
@@ -85,30 +87,68 @@ var (
 	}}
 )
 
-func triggerFailpoints(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, config FailpointConfig) {
+func verifyClusterHealth(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, failpointName string) error {
```
serathius, line 90: failpointName is not used

serathius, line 90: Can you move the function below `triggerFailpoints`, they should be ordered top-down.

tests/robustness/failpoints.go
```diff
@@ -85,30 +87,68 @@ var (
...
+			Logger:               zap.NewNop(),
+			DialKeepAliveTime:    1 * time.Millisecond,
+			DialKeepAliveTimeout: 5 * time.Millisecond,
+		})
+		defer clusterClient.Close()
+
+		cli := healthpb.NewHealthClient(clusterClient.ActiveConnection())
+		resp, err := cli.Check(ctx, &healthpb.HealthCheckRequest{})
+		if err != nil {
+			return fmt.Errorf("Error checking member health: %v", err)
```
serathius, line 103: Can you provide name of member that is failing health?  `clus.Procs[i].Config().Name`

tests/robustness/failpoints.go
```diff
@@ -85,30 +87,68 @@ var (
...
+		}
+		if resp.Status != healthpb.HealthCheckResponse_SERVING {
+			return fmt.Errorf("Member health status expected %s, got %s", healthpb.HealthCheckResponse_SERVING, resp.Status)
```
serathius, line 106: ditto name of member

tests/robustness/failpoints.go
```diff
@@ -85,30 +87,68 @@ var (
...
 		if !config.failpoint.Available(proc) {
 			t.Errorf("Failpoint %q not available on %s", config.failpoint.Name(), proc.Config().Name)
-			return
+			return nil
 		}
 	}
 	for successes < config.count && failures < config.retries {
 		time.Sleep(config.waitBetweenTriggers)
+
+		lg.Info("Verifying cluster health before failpoint\n", zap.String("failpoint", config.failpoint.Name()))
```
serathius, line 125: No need for a newline character '\n', you can remove one from log below too.

tests/robustness/failpoints.go
jmhbnz: Good idea, done 👍🏻

tests/robustness/failpoints.go
jmhbnz: Good catch, leftover from debug!

tests/robustness/failpoints.go
```diff
@@ -85,30 +87,71 @@ var (
...
+		clusterClient, err := clientv3.New(clientv3.Config{
+			Endpoints:            clus.Procs[i].EndpointsGRPC(),
+			Logger:               zap.NewNop(),
+			DialKeepAliveTime:    1 * time.Millisecond,
+			DialKeepAliveTimeout: 5 * time.Millisecond,
+		})
+		defer clusterClient.Close()
+
+		cli := healthpb.NewHealthClient(clusterClient.ActiveConnection())
+		resp, err := cli.Check(ctx, &healthpb.HealthCheckRequest{})
```
chaochn47, line 143: Just curious, right now the server health status is always serving as long as the etcd process is live. 

https://github.com/etcd-io/etcd/blob/622017468725647028e75aea2c971966ab70b3b5/server/etcdserver/api/v3rpc/grpc.go#L84-L86

So the health check here is meant to verify each process liveness, correct?

serathius: That's good enough for our use case

serathius (approved):

tests/robustness/failpoints.go
```diff
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
```
fuweid, line 25: minor issue: please remove the following empty line.

tests/robustness/failpoints.go
```diff
@@ -85,30 +87,71 @@ var (
...
 	if successes < config.count || failures >= config.retries {
 		t.Errorf("failed to trigger failpoints enough times, err: %v", err)
 	}
+
+	return nil
+}
+
+func verifyClusterHealth(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) error {
+	for i := 0; i < len(clus.Procs); i++ {
+		clusterClient, err := clientv3.New(clientv3.Config{
```
fuweid, line 134: I think we should check err before go defer, right?

jmhbnz: Good call, done.

serathius:
ping @ahrtr @ptabor

ahrtr:
I don't understand why we need this (Thanks @jmhbnz 's effort anyway).  
- When we first start a cluster, all member must be healthy, because the e2e test framework waits for the message "`ready to serve client requests`".
- When the robustness test  triggers any failpoint, it also needs to guarantee the member to come back to healthy status afterwards. If any failpoint doesn't guarantee this, we should fix the failpoint's implementation.

serathius:
Please read https://github.com/etcd-io/etcd/issues/15595, we injected the failpoint on one member, but other members crashed. This is unexpected and should be detected by failpoint code as we cannot say that failpoint injection succeeded if cluster was unhealthy before or after.

ahrtr:
> Please read #15595, we injected the failpoint on one member, but other members crashed.

Based on the discussion in  https://github.com/etcd-io/etcd/issues/15595, it's because the proxy layer has issue. Shouldn't the proxy layer be fixed?  No matter it's production or test environments, if a member crashes unexpectedly, it should be an critical or major issue, we should fix it. Adding more protection may not be good, because we may regard it as a flaky case and just retry, and accordingly hiding the real issue.

serathius:
> > Please read #15595, we injected the failpoint on one member, but other members crashed.
> 
> Based on the discussion in #15595, it's because the proxy layer has issue. Shouldn't the proxy layer be fixed? No matter it's production or test environments, if a member crashes unexpectedly, it should be an critical or major issue, we should fix it. Adding more protection may not be good, because we may regard it as a flaky case and just retry, and accordingly hiding the real issue.

No, the trigger was the proxy blackholing, but for robustness tests problem was that etcd followers crashed. This was unexpected. Because tests do not expect this they:
* continued to run as normal
* hid the follower panic message
* reported incorrect error "not enough qps" 

This is an unexpected error, thus tests should not retry it but exit immediately. And this is what @jmhbnz implemented. We mark the test as fail with `t.Error` and cancel all the concurrent processes. I prefer a graceful shutdown here vs a `t.Fatal` as it avoids obscuring etcd panic and gives us report of operations and db files.

Please ask about the code, instead of making an incorrect assumption. The design was also discussed https://github.com/etcd-io/etcd/issues/15596#issuecomment-1491536047

tests/robustness/failpoints.go
```diff
@@ -83,30 +84,74 @@ var (
...
+			DialKeepAliveTime:    1 * time.Millisecond,
+			DialKeepAliveTimeout: 5 * time.Millisecond,
+		})
+		if err != nil {
+			return fmt.Errorf("Error creating client for cluster %s: %v", clus.Procs[i].Config().Name, err)
+		}
+		defer clusterClient.Close()
+
+		cli := healthpb.NewHealthClient(clusterClient.ActiveConnection())
+		resp, err := cli.Check(ctx, &healthpb.HealthCheckRequest{})
```
ptabor, line 143: Potentially we should have a 'helper' retrier (3 times) in case of connection flakiness around such semi-grpc code. 
We might monitor it for flakes... but intuitively there will be (even though it's a 'localhost' communication).

ptabor (approved):
LGTM. Thank you. 

I think it's better to have even redundant sources of signal and fail the tests early if anything is not going as expected.

tests/robustness/failpoints.go
serathius: If there are flakes we are sure to discover them in nightly tests. We can consider it as a followup.

serathius:
Ups, wanted to fix the conflict myself, but GitHub editor is terrible. Please rebase the PR and sorry for the mess.

Issue #15594, opened 2023-03-30
Simplify etcd image release process


jmhbnz:
### What would you like to be added?

Currently Etcd publishes multi-arch along with arch specific images, i.e:

![image](https://user-images.githubusercontent.com/39425256/228929623-9a09515a-d62f-4b2e-ad00-ca711b9cbeca.png)

Now that multi-arch image manifests are mature and widely adopted should we stop creating and pushing single arch images?

We would need to consider how much lead time to give for this change and how we communicate this to folks that might be consuming the single arch image tags currently.

### Why is this needed?

This would simplify `scripts/release.sh` and `scripts/test_images.sh`.

jmhbnz:
From the discussion in #15270 it sounds like there might be agreement to go ahead with this feature. I've created #15605 but left in draft because I think this still needs wider discussion, particularly on when it would be ok to make this change and what steps we should take to communicate it.

PR #15605, opened 2023-03-31
Simplify etcd image release process

jmhbnz:
Only test and publish the single multi arch image to simplify etcd release process and the image landscape for downstream users to consume.

Leaving this in draft as it needs to be discussed when it would be appropriate to make this change and how it would be communicated.

Fixes: #15594

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -258,13 +258,6 @@ main() {
       log_warning "login failed, retrying"
     done
 
-    for TARGET_ARCH in "amd64" "arm64" "ppc64le" "s390x"; do
-      log_callout "Pushing container images to quay.io ${RELEASE_VERSION}-${TARGET_ARCH}"
-      maybe_run docker push "quay.io/coreos/etcd:${RELEASE_VERSION}-${TARGET_ARCH}"
-      log_callout "Pushing container images to gcr.io ${RELEASE_VERSION}-${TARGET_ARCH}"
-      maybe_run docker push "gcr.io/etcd-development/etcd:${RELEASE_VERSION}-${TARGET_ARCH}"
-    done
-
     log_callout "Creating manifest-list (multi-image)..."
 
     for TARGET_ARCH in "amd64" "arm64" "ppc64le" "s390x"; do
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -5,6 +5,7 @@ set -o nounset
 set -o pipefail
 
 source ./scripts/test_lib.sh
+source ./scripts/docker_lib.sh
 source ./scripts/release_mod.sh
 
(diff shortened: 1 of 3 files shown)


scripts/test_images.sh
```diff
@@ -82,13 +82,3 @@ if [ "${GET}" != "${VALUE}" ]; then
 fi
 
 echo "Succesfully tested etcd local image ${TAG}"
-
-for TARGET_ARCH in "amd64" "arm64" "ppc64le" "s390x"; do
```
serathius, line 86: Can you update the testing of ARCH, so we validate that multi-arch image correctly links to subimages for all supported architectures?

scripts/release.sh
```diff
@@ -258,13 +258,6 @@ main() {
       log_warning "login failed, retrying"
     done
 
-    for TARGET_ARCH in "amd64" "arm64" "ppc64le" "s390x"; do
```
serathius, line 261: Would be good to test if you can push multi-arch image without pushing each arch. Workaround would be to just push images but avoid tagging them.

scripts/test_images.sh
jmhbnz: Sure thing - I noticed while I was working on it there is duplication of key variables and functions between `test_images.sh` and `release.sh` so I have refactored into a new `docker_lib.sh`, added a lot more error handling and improved consistency with how errors are reported. Let me know what you think.

scripts/release.sh
jmhbnz: Ahh bad news, yeah `docker manifest` will only ever work with the registry images and can't do any magic that would allow us not to push the multi arch only.

Unfortunately even `docker buildx` works pretty much the same way, just with simpler ux.

So yeah we could push and not tag but actually reviewing kubernetes docs they are still publishing individual arch images:

> https://kubernetes.io/releases/download/
> All container images are available for multiple architectures, whereas the container runtime should choose the correct one based on the underlying platform. It is also possible to pull a dedicated architecture by suffixing the container image name, for example registry.k8s.io/kube-apiserver-arm64:v1.26.0. All those derivations are signed in the same way as the multi-architecture manifest lists.

So actually I don't think we can simplify the image pushing at all?  Your call if we just close this or continue with just the refactoring I mentioned above.

scripts/test_images.sh
jmhbnz: Actually there is a hitch here as well, when in dry run the images don't actually exist in the remote registry. With `docker manifest` only working on an actual registry rather than local, refer https://github.com/docker/cli/issues/3350

Reading through there are some potential hacks to work around, or we could spin up a temporary registry as part of ci to test this more naturally, not sure what the appetite for either is though.

PR #15610, opened 2023-04-01
Robustness fix blackhole

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -478,6 +478,9 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 	peerAdvertiseUrl := url.URL{Scheme: cfg.PeerScheme(), Host: fmt.Sprintf("localhost:%d", peerPort)}
 	var proxyCfg *proxy.ServerConfig
 	if cfg.PeerProxy {
+		if !cfg.IsPeerTLS {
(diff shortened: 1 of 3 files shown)


serathius:
Enabling TLS fixes the issue  with blackholing https://github.com/etcd-io/etcd/actions/runs/4583625819

ahrtr:
do not get time to follow up https://github.com/etcd-io/etcd/issues/15595, please clarify why the issue did not affect TLS.

serathius:
Proxy we are using is doing byte manipulation on arbitrary byte, with no regards for packets. It just reads the stream of bytes and writes stream of bytes. This means that it doesn't really simulate packet loss line normal network will experience, as it will repackage remaining bytes into new packets. So for receiver behind the proxy they just get arbitrary bytes from stream, without any signs of manipulation (repackaging by proxy provides new control sums). I expect that for TLS this doesn't work as byte manipulation will make data invalid after decryption.

So without TLS blackholing is not really packet loss, but a removal of arbitrary bytes from connection stream. This is works more like arbitrary packet manipulation than what can happen in real network. Enabling TLS helps as receiver can detect corrupted byte stream and avoid using corrupted byte stream.

ahrtr:
> I expect that for TLS this doesn't work as byte manipulation will make data invalid after decryption.

Thanks for the clarification.

ahrtr (approved):

jmhbnz (approved):
LGTM

ptabor (approved):

PR #15633, opened 2023-04-03
tests: Enfoce timeout on failpoints

serathius:
cc @ahrtr @ptabor


ahrtr (approved):
LGTM

tests/robustness/linearizability_test.go
```diff
@@ -203,18 +203,21 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, config e2
...
+		defer trafficCancel()
 		triggerFailpoints(ctx, t, lg, clus, failpoint)
 		time.Sleep(time.Second)
-		trafficCancel()
 		return nil
 	})
 	maxRevisionChan := make(chan int64, 1)
 	g.Go(func() error {
+		defer func() {
+			maxRevisionChan <- operationsMaxRevision(operations)
```
fuweid, line 217: Just curiosity, the defer is used to recover just in case that `simulateTraffic` panic?

serathius: That was the intention, however I don't think it's correct. I will revert this line

serathius: Yes, it would not work without a change I just made. I have updated `operationsMaxRevision` to always correct revision which is at least 1.

PR #15656, opened 2023-04-06
protect LeaseTimeToLive with RBAC

mitake:

Currently, `LeaseTimeToLive()` API or `etcdctl lease timetolive` command don't require RBAC permission. However, their responses can have key names attached to the target lease. Value of the keys cannot be accessed through the API, but protecting the key name is better. So this PR let the API require read permission of keys attached to a lease (if a lease isn't attached to specific keys, behavior won't be changed).

cc @ahrtr @serathius @spzala @ptabor

### DIFF
--- a/server/etcdserver/v3_server.go
+++ b/server/etcdserver/v3_server.go
@@ -317,7 +317,32 @@ func (s *EtcdServer) LeaseRenew(ctx context.Context, id lease.LeaseID) (int64, e
 	return -1, errors.ErrCanceled
 }
 
-func (s *EtcdServer) LeaseTimeToLive(ctx context.Context, r *pb.LeaseTimeToLiveRequest) (*pb.LeaseTimeToLiveResponse, error) {
+func (s *EtcdServer) checkLeaseTimeToLive(ctx context.Context, leaseID lease.LeaseID) (error, uint64) {
+	rev := s.AuthStore().Revision()
+	if !s.AuthStore().IsAuthEnabled() {
+		return nil, rev
+	}
+	authInfo, err := s.AuthInfoFromCtx(ctx)
+	if err != nil {
+		return err, rev
+	}
+	if authInfo == nil {
+		return auth.ErrUserEmpty, rev
+	}
+
+	l := s.lessor.Lookup(leaseID)
+	if l != nil {
+		for _, key := range l.Keys() {
+			if err := s.AuthStore().IsRangePermitted(authInfo, []byte(key), []byte{}); err != nil {
+				return err, 0
+			}
+		}
+	}
+
+	return nil, rev
+}
+
+func (s *EtcdServer) leaseTimeToLive(ctx context.Context, r *pb.LeaseTimeToLiveRequest) (*pb.LeaseTimeToLiveResponse, error) {
 	if s.isLeader() {
 		if err := s.waitAppliedIndex(); err != nil {
 			return nil, err
@@ -365,6 +390,30 @@ func (s *EtcdServer) LeaseTimeToLive(ctx context.Context, r *pb.LeaseTimeToLiveR
 		return nil, errors.ErrTimeout
 	}
 	return nil, errors.ErrCanceled
+}
+
+func (s *EtcdServer) LeaseTimeToLive(ctx context.Context, r *pb.LeaseTimeToLiveRequest) (*pb.LeaseTimeToLiveResponse, error) {
+	var rev uint64
+	var err error
+	if r.Keys {
+		// check RBAC permission only if Keys is true
+		err, rev = s.checkLeaseTimeToLive(ctx, lease.LeaseID(r.ID))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	resp, err := s.leaseTimeToLive(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.Keys {
+		if s.AuthStore().IsAuthEnabled() && rev != s.AuthStore().Revision() {
+			return nil, auth.ErrAuthOldRevision
+		}
+	}
+	return resp, nil
 }
 
 func (s *EtcdServer) newHeader() *pb.ResponseHeader {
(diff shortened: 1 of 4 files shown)


server/etcdserver/v3_server.go
```diff
@@ -367,6 +392,23 @@ func (s *EtcdServer) LeaseTimeToLive(ctx context.Context, r *pb.LeaseTimeToLiveR
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.leaseTimeToLive(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+
+	if rev != s.AuthStore().Revision() {
```
ahrtr, line 406: I am thinking probably we should just skip the check when auth isn't enabled. For example, when auth isn't enabled, users may also update RBAC to cause the autuRev to increase; obviously we shouldn't return `auth.ErrAuthOldRevision` in that case.
```suggestion
	if s.AuthStore().IsAuthEnabled() && rev != s.AuthStore().Revision() {
```

tests/e2e/ctl_v3_auth_test.go
```diff
@@ -634,3 +635,45 @@ func authTestCacheReload(cx ctlCtx) {
...
+	if err != nil {
+		cx.t.Fatal(err)
+	}
+
+	cx.user = "test-user"
+	cx.pass = "pass"
+	// the lease is attached to bar, which test-user cannot access
+	err = ctlV3LeaseTimeToLive(cx, leaseID)
+	if err == nil {
+		cx.t.Fatalf("test-user must not be able to access to the lease, because it's attached to the key bar")
```
ahrtr, line 677: nit:
```suggestion
		cx.t.Fatal("test-user must not be able to access to the lease, because it's attached to the key bar")
```

server/etcdserver/v3_server.go
mitake: Thanks, it makes sense! Let me update this PR.

ahrtr:
@mitake Overall looks good to me, and sorry for the late response. Could you rebase this PR although there is no conflict?

mitake:
@ahrtr thanks, rebased on the latest main, could you check?

serathius:
Can we measure the performance impact when someone has enabled auth?

 I think that all other etcd methods authorization check can be done in constant time. Here we have linear to number of keys under lease. Would be good to find impact for higher number of leases. For example K8s (doesn't use auth, but just an example) has up to 1000 keys per lease.

server/etcdserver/v3_server.go
```diff
@@ -317,7 +317,32 @@ func (s *EtcdServer) LeaseRenew(ctx context.Context, id lease.LeaseID) (int64, e
...
+	}
+
+	l := s.lessor.Lookup(leaseID)
+	if l != nil {
+		for _, key := range l.Keys() {
+			if err := s.AuthStore().IsRangePermitted(authInfo, []byte(key), []byte{}); err != nil {
+				return err, 0
+			}
+		}
+	}
```
ahrtr, line 340: Before iterating over all the keys, I think we also need to check [LeaseTimeToLiveRequest.Keys](https://github.com/etcd-io/etcd/blob/8c5e9ad455e05a9435dad4a1a1e52733b538e3ba/api/etcdserverpb/rpc.pb.go#L2805); if it's `false`, it means users don't want to query al the keys at all, then no need to check the permission either on the server side.

We also need to add test to cover such case, such as the client has no range permission on a key, but it intentionally set `LeaseTimeToLiveRequest.Keys` to `false`, then it should not get an error response.

mitake:
@serathius I measured overhead with [this simple benchmark](https://gist.github.com/mitake/1ec92b27655f703012ffd636b890e220) which emulates 1000 keys/lease on my local laptop (just for comparing the difference w/ and w/o the checking overhead). Result was like below (unit is microsecond):

auth enabled:
- 1st run: min: 50.000000, median: 116.000000, 99 percentile: 539.000000
- 2nd run: min: 52.000000, median: 117.000000, 99 percentile: 617.000000
- 3rd run: min: 48.000000, median: 115.000000, 99 percentile: 686.000000

auth disabled:
- 1st run: min: 49.000000, median: 111.500000, 99 percentile: 700.000000
- 2nd run: min: 47.000000, median: 112.000000, 99 percentile: 642.000000
- 3rd run: min: 48.000000, median: 111.000000, 99 percentile: 666.000000

It's a noisy environment so 99 percentile wouldn't be so informative. The workload is quite artificial (e.g. cache hit rate should be quite high) but I think overhead is low enough. What do you think?

server/etcdserver/v3_server.go
mitake: Thanks for your comment, I'll update tomorrow.

serathius:
> auth enabled:
> 
> * 1st run: min: 50.000000, median: 116.000000, 99 percentile: 539.000000
> * 2nd run: min: 52.000000, median: 117.000000, 99 percentile: 617.000000
> * 3rd run: min: 48.000000, median: 115.000000, 99 percentile: 686.000000
> 
> auth disabled:
> 
> * 1st run: min: 49.000000, median: 111.500000, 99 percentile: 700.000000
> * 2nd run: min: 47.000000, median: 112.000000, 99 percentile: 642.000000
> * 3rd run: min: 48.000000, median: 111.000000, 99 percentile: 666.000000

One observation, shouldn't we expect auth enabled take more as it one that does checks of keys?

mitake:
@serathius Sorry I noticed that the benchmark isn't attaching lease correctly :( I did it again and could observe overhead correctly:

auth enabled:
- 1st run: min: 583.000000, median: 705.000000, 99 percentile: 3411.000000
- 2nd run: min: 592.000000, median: 703.000000, 99 percentile: 3535.000000
- 3rd run: min: 594.000000, median: 707.000000, 99 percentile: 3659.000000

auth disabled:
- 1st run: min: 53.000000, median: 118.000000, 99 percentile: 708.000000
- 2nd run: min: 53.000000, median: 119.000000, 99 percentile: 693.000000
- 3rd run: min: 52.000000, median: 118.000000, 99 percentile: 631.000000

(checking 1000 permission in a few microseconds is clearly impossible even for high cache hit condition...)

What do you think?

> Looks good, for the next time I recommend looking into benchmarking https://about.sourcegraph.com/blog/go/gophercon-2019-optimizing-go-code-without-a-blindfold and using https://pkg.go.dev/golang.org/x/perf/cmd/benchstat to eliminate noisy environment and confirming that results are statistically significant.

Thanks for sharing, I'll check these materials.

server/etcdserver/v3_server.go
mitake: Updated, could you check @ahrtr ?

ahrtr (approved):
LGTM

Thanks @mitake

spzala (approved):
Thanks @mitake

Issue #15545, opened 2023-03-22
TestV3WatchRestoreSnapshotUnsync fails


fuweid:
### Which github workflows are flaking?

[test (linux-amd64-integration-2-cpu)](https://github.com/etcd-io/etcd/runs/12180892595?check_suite_focus=true)

### Which tests are flaking?

TestV3WatchRestoreSnapshotUnsync

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4486206847/jobs/7888474622?pr=15519

### Reason for failure (if possible)

Unknown

chaochn47:
https://github.com/etcd-io/etcd/actions/runs/4548835671/jobs/8020273579?pr=15580

fuweid:
> Did you get a repro locally?

@chaochn47 Yes. It's easy to reproduce it. Still thinking how to fix it.

PR #15667, opened 2023-04-07
tests: deflake TestV3WatchRestoreSnapshotUnsync

fuweid:
The TestV3WatchRestoreSnapshotUnsync setups three members' cluster. Before serving any update requests from client, after leader elected, each member will have index 8 log: 3 x ConfChange + 3 x ClusterMemberAttrSet + 1 x ClusterVersionSet.

Based on the config (SnapshotCount: 10, CatchUpCount: 5), we need to file update requests to trigger snapshot at least twice.

T1: L(snapshot-index: 11, compacted-index:  6) F_m0(index: 8)
T2: L(snapshot-index: 22, compacted-index: 17) F_m0(index: 8, out of date)

After member0 recovers from network partition, it will reject leader's request and return hint (index:8, term:x). If it happens after second snapshot, leader will find out the index:8 is out of date and force to transfer snapshot.

However, the client only files 15 update requests and leader doesn't finish the process of snapshot in time. Since the last of compacted-index is 6, leader can still replicate index:9 to member0 instead of snapshot.

```bash
cd tests/integration
CLUSTER_DEBUG=true go test -v -count=1 -run TestV3WatchRestoreSnapshotUnsync ./
...

INFO    m2.raft 3da8ba707f1a21a4 became leader at term 2        {"member": "m2"}
...
INFO    m2      triggering snapshot     {"member": "m2", "local-member-id": "3da8ba707f1a21a4", "local-member-applied-index": 22, "local-member-snapshot-index": 11, "local-member-snapshot-count": 10, "snapshot-forced": false}
...

cluster.go:1359: network partition between: 99626fe5001fde8b <-> 1c964119da6db036
cluster.go:1359: network partition between: 99626fe5001fde8b <-> 3da8ba707f1a21a4
cluster.go:416: WaitMembersForLeader

INFO    m0.raft 99626fe5001fde8b became follower at term 2      {"member": "m0"}
INFO    m0.raft raft.node: 99626fe5001fde8b elected leader 3da8ba707f1a21a4 at term 2   {"member": "m0"}
DEBUG   m2.raft 3da8ba707f1a21a4 received MsgAppResp(rejected, hint: (index 8, term 2)) from 99626fe5001fde8b for index 23      {"member": "m2"}
DEBUG   m2.raft 3da8ba707f1a21a4 decreased progress of 99626fe5001fde8b to [StateReplicate match=8 next=9 inflight=15]  {"member": "m2"}

DEBUG   m0      Applying entries        {"member": "m0", "num-entries": 15}
DEBUG   m0      Applying entry  {"member": "m0", "index": 9, "term": 2, "type": "EntryNormal"}

....

INFO    m2      saved snapshot  {"member": "m2", "snapshot-index": 22}
INFO    m2      compacted Raft logs     {"member": "m2", "compact-index": 17}
```

To fix this issue, we need to file more requests to force leader to trigger snapshot three times. And we also need to scan snapshot dir to ensure that last one snapshot is generated.

Fixes: #15545


tests/integration/v3_watch_restore_test.go
```diff
@@ -69,7 +73,7 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
 		t.Fatal(errW)
 	}
 	if err := wStream.Send(&pb.WatchRequest{RequestUnion: &pb.WatchRequest_CreateRequest{
-		CreateRequest: &pb.WatchCreateRequest{Key: []byte("foo"), StartRevision: 5}}}); err != nil {
+		CreateRequest: &pb.WatchCreateRequest{Key: []byte("foo"), StartRevision: 25}}}); err != nil {
```
chaochn47, line 76: Just curious, is watching from 25 revision counted as "old revision that were created in synced watcher group in the first place"?

chaochn47:
Great find!

tests/integration/v3_watch_restore_test.go
```diff
@@ -90,19 +94,38 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+	// there are 3xConfChange + 3xMemberAttrSet + 1xClusterVersionSet logs.
+	// So member 0 has index 8 in raft log before network partition. We
+	// need to trigger snapshot at least twice.
+	//
+	// SnapshotCount: 10, SnapshotCatchUpEntries: 5
+	// T1: L(snapshot-index: 11, compacted-index:  6), F_m0(index:8)
+	// T2: L(snapshot-index: 22, compacted-index: 17), F_m0(index:8, out of date)
+	// T3: L(snapshot-index: 33, compacted-index: 28), F_mo(index:8, out of date)
+	//
+	// Since the snapshot is handled in GoAttach, we need to trigger
```
lavacat, line 108: The key part for leader to send snapshot is to to get leader to [compact](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/server.go#L2114) index 8. If you change SnapshotCatchUpEntries to 1, this can be achieved in 1 call of EtcdServer.snapshot.

tests/integration/v3_watch_restore_test.go
```diff
@@ -90,19 +94,38 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+	// snapshot 3 times and then leader starts to send snapshot after
+	// the leader realizes the hint index from member 0 is out of date.
+	for i := 0; i < 35; i++ {
 		_, err := kvc.Put(context.TODO(), &pb.PutRequest{Key: []byte("foo"), Value: []byte("bar")})
 		if err != nil {
 			t.Errorf("#%d: couldn't put key (%v)", i, err)
 		}
 	}
+	// The 33 is latest snapshot index.
+	ensureTriggerSnapshot(t, clus.Members[initialLead], 33, 5*time.Second)
```
lavacat, line 118: ideally this should wait for leader to compact index 8, but there is no way to check for that.
Writing snapshot files isn't enough to trigger snapshot send. For example if you set SnapshotCatchUpEntries to smth high, new version of the test will still fail. Maybe you can add a comment that implicitly it's also waiting for compaction.

tests/integration/v3_watch_restore_test.go
fuweid: Thanks for the link! Not sure I understand it correctly: we don't defrag it so the revision is still available in kv store. Is it correct?

chaochn47: Yeah, from my understanding, as long as the requested revision is not compacted, the key value is still available. 

image

Here is a simple diagram. /cc @ahrtr @serathius in case the understanding or direction is wrong ==

tests/integration/v3_watch_restore_test.go
fuweid: Correct. It was my first patch. But I run into other error.

```diff
diff --git a/tests/integration/v3_watch_restore_test.go b/tests/integration/v3_watch_restore_test.go
index bdebeacfc..552f4aecf 100644
--- a/tests/integration/v3_watch_restore_test.go
+++ b/tests/integration/v3_watch_restore_test.go
@@ -57,7 +57,7 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
        clus := integration.NewCluster(t, &integration.ClusterConfig{
                Size:                   3,
                SnapshotCount:          10,
-               SnapshotCatchUpEntries: 5,
+               SnapshotCatchUpEntries: 1,
        })
        defer clus.Terminate(t)
```

```bash
➜  integration git:(main) ✗ taskset -c 0,1 go test -v -count=1000 -timeout=700m -failfast -run TestV3WatchRestoreSnapshotUnsync ./

... 
    logger.go:130: 2023-04-08T09:24:52.907+0800 INFO    m1      saved snapshot  {"member": "m1", "snapshot-index": 22}
    logger.go:130: 2023-04-08T09:24:52.907+0800 INFO    m1      compacted Raft logs     {"member": "m1", "compact-index": 21}
    logger.go:130: 2023-04-08T09:24:52.907+0800 INFO    m0.raft d6a69b975d0a7658 became follower at term 2      {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.907+0800 INFO    m0.raft raft.node: d6a69b975d0a7658 elected leader c0c978b536d663c2 at term 2   {"member": "m0"}
{"level":"warn","ts":"2023-04-08T09:24:52.90773+0800","logger":"etcd-client","caller":"v3/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc00233d860/localhost:m0","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = Unavailable desc = etcdserver: leader changed"}
    logger.go:130: 2023-04-08T09:24:52.911+0800 INFO    m2      saved snapshot  {"member": "m2", "snapshot-index": 22}
    logger.go:130: 2023-04-08T09:24:52.911+0800 INFO    m2      skip compaction since there is an inflight snapshot     {"member": "m2"}
    logger.go:130: 2023-04-08T09:24:52.911+0800 INFO    m2      sending database snapshot       {"member": "m2", "snapshot-index": 22, "remote-peer-id": "d6a69b975d0a7658", "bytes": 33666, "size": "34 kB"}
    logger.go:130: 2023-04-08T09:24:52.911+0800 INFO    m2      sending merged snapshot {"member": "m2", "from": "c0c978b536d663c2", "to": "d6a69b975d0a7658", "bytes": 33666, "size": "34 kB"}
    logger.go:130: 2023-04-08T09:24:52.911+0800 INFO    m2      sent database snapshot to writer        {"member": "m2", "bytes": 24576, "size": "25 kB"}
    logger.go:130: 2023-04-08T09:24:52.911+0800 INFO    m0      receiving database snapshot     {"member": "m0", "local-member-id": "d6a69b975d0a7658", "remote-snapshot-sender-id": "c0c978b536d663c2", "incoming-snapshot-index": 22, "incoming-snapshot-message-size-bytes": 9090, "incoming-snapshot-message-size": "9.1 kB"}

...

    logger.go:130: 2023-04-08T09:24:52.915+0800 INFO    m0      restoring lease store   {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.917+0800 INFO    m0      restored lease store    {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.917+0800 INFO    m0      restoring mvcc store    {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.917+0800 INFO    m0      kvstore restored        {"member": "m0", "current-rev": 15}
    logger.go:130: 2023-04-08T09:24:52.917+0800 INFO    m0      restored mvcc store     {"member": "m0", "consistent-index": 22}
    logger.go:130: 2023-04-08T09:24:52.917+0800 INFO    m0      restoring alarm store   {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.917+0800 INFO    m0      closing old backend file        {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.918+0800 INFO    m0      restored alarm store    {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.918+0800 INFO    m0      restoring auth store    {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.918+0800 INFO    m0      restored auth store     {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.918+0800 INFO    m0      restoring v2 store      {"member": "m0"}
    logger.go:130: 2023-04-08T09:24:52.919+0800 INFO    m0      restored v2 store       {"member": "m0"}

...

    v3_watch_restore_test.go:128: sleeping for 2 seconds DONE
    v3_watch_restore_test.go:153: wStream.Recv error: expected 12 events, got [kv:<key:"foo" create_revision:2 mod_revision:5 version:4 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:6 version:5 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:7 version:6 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:8 version:7 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:9 version:8 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:10 version:9 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:11 version:10 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:12 version:11 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:13 version:12 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:14 version:13 value:"bar" >  kv:<key:"foo" create_revision:2 mod_revision:15 version:14 value:"bar" > ]
```

m0 should recover to have (consistent-index:23, current-rev:16). But it got (consisten-index:22, curent-rev:15). 
It is timing issue. 

https://github.com/etcd-io/etcd/blob/7153a8f2f48bc36dec1caafb97001634f0366fc4/server/etcdserver/server.go#L938

The server uses old index which doesn't match the snapshot content. Not sure that it is bug or fault-tolerant. Finally, it can catch-up. However, the watch event is not expected. cc @ahrtr 

So I try to make this patch to let it has time to handle memory compact. 
I am still testing it with limit core by `taskset -c`. Will update it later.
(If it doesn't work, I will switch to watch the log...)

tests/integration/v3_watch_restore_test.go
fuweid: I switch to a solution which monitors testing log to ensure cluster has compacted the raft log. please take a look. Thanks

tests/integration/v3_watch_restore_test.go
```diff
@@ -81,14 +87,30 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
 	t.Logf("sleeping for 2 seconds")
 	time.Sleep(2 * time.Second)
 	t.Logf("sleeping for 2 seconds DONE")
 
 	kvc := integration.ToGRPC(clus.Client(1)).KV
 
+	// NOTE: In 3 members cluster, after initial lead has been elected,
+	// there are 3xConfChange + 3xMemberAttrSet + 1xClusterVersionSet logs.
+	// So member 0 has index 8 in raft log before network partition. We
+	// need to trigger snapshot at least twice.
```
lavacat, line 101: nit: `need to trigger EtcdServer.snapshot()` 
just a clarification for anyone new to this.

tests/integration/v3_watch_restore_test.go
```diff
@@ -97,12 +119,22 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
 	}
 
+	// ensure two members has compacted the log twice.
+	if err := logSub.wait(5 * time.Second); err != nil {
+		t.Fatal("Failed to ensure that members compacted Raft log in 5 seconds")
+	}
+	logMonitor.delSubscriber(logTopic)
+	t.Logf("two members have compacted raft logs")
+
 	// trigger snapshot send from leader to this slow follower
```
lavacat, line 129: nit: replace with `After RecoverPartition, leader L will send snapshot to slow F_m0 follower, because  F_m0(index:8) is 'out of date' compared to L(compacted-index:17)`

lavacat (approved):
LGTM, some nitpicks

tests/integration/v3_watch_restore_test.go
```diff
@@ -154,3 +186,89 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+type lineCountExpecter struct {
+	doneOnce sync.Once
+	doneCh   chan struct{}
+
+	content string
+	count   int64
+	seen    int64
+}
+
+func newlineCountExpecter(interestingContent string, expectedCount int64) *lineCountExpecter {
```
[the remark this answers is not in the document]
lavacat: Also, `newlineCountExpecter ` -> `newLineCountExpecter`

tests/integration/v3_watch_restore_test.go
```diff
@@ -154,3 +186,89 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+}
+
+func newTestingLogfMonitor(tb testutil.TB) *testingLogfMonitor {
+	return &testingLogfMonitor{
+		TB:          tb,
+		subscribers: make(map[string]testingLogfSubscriber),
+	}
+}
+
+func (m *testingLogfMonitor) addSubscriber(topic string, sub testingLogfSubscriber) {
```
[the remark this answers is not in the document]
fuweid: updated!

tests/integration/v3_watch_restore_test.go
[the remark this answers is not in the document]
fuweid: updated!

tests/integration/v3_watch_restore_test.go
fuweid: Updated!

tests/integration/v3_watch_restore_test.go
fuweid: Nice comment. Updated!

tests/integration/v3_watch_restore_test.go
```diff
@@ -154,3 +186,89 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+		return ctx.Err()
+	}
+}
+
+type testingLogfSubscriber interface {
+	Notify(log string)
+}
+
+// testingLogfMonitor is to monitor t.Logf output.
+type testingLogfMonitor struct {
```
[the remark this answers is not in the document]
fuweid: https://github.com/etcd-io/etcd/blob/f7af6b64bafcc32071b9867ffc7447e3c66111ae/tests/framework/integration/cluster.go#L746

I think we can extend the zaptest.Core with tee function. It can be aligned with AssertProcessLogs for process expecter~

tests/integration/v3_watch_restore_test.go
```diff
@@ -81,14 +87,30 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+	initialLead := clus.WaitMembersForLeader(t, clus.Members[1:]) + 1
 	t.Logf("elected lead: %v", clus.Members[initialLead].Server.MemberId())
 	t.Logf("sleeping for 2 seconds")
 	time.Sleep(2 * time.Second)
 	t.Logf("sleeping for 2 seconds DONE")
 
 	kvc := integration.ToGRPC(clus.Client(1)).KV
 
+	// NOTE: In 3 members cluster, after initial lead has been elected,
+	// there are 3xConfChange + 3xMemberAttrSet + 1xClusterVersionSet logs.
```
ahrtr, line 99: This isn't correct. Note that when starting a new cluster with 3 members, each member will apply 3 ConfChange directly at the beginning before a leader is elected. 

Please read https://github.com/etcd-io/raft/blob/918455b897764d8e4d12af70657280365ffaaa04/bootstrap.go#L76-L78

tests/integration/v3_watch_restore_test.go
```diff
@@ -97,12 +119,23 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+	// L(compacted-index:17).
 	clus.Members[0].RecoverPartition(t, clus.Members[1:]...)
-	// We don't expect leadership change here, just recompute the leader'Server index
-	// within clus.Members list.
+	// We don't expect leadership change here, recompute the leader'Server
+	// index within clus.Members list.
 	lead := clus.WaitLeader(t)
+	if lead != initialLead {
+		t.Fatalf("expected leader index (%v), but got (%v)", initialLead, lead)
+	}
```
ahrtr, line 138: In theory, the leader might change, although usually it will not.  So suggest to remove it.

ahrtr:
Thanks @fuweid 

The `lineCountExpecter` and `testingLogfMonitor` seem like a generic solution, should be moved into test framework, so that it can be reused. 

But my immediate feeling is it's a little over complicated,  can we just increase the [timeout](https://github.com/etcd-io/etcd/blob/3393d135679ecdb60d3d19f86b939e276c6ed096/tests/integration/v3_watch_restore_test.go#L149) (e.g. to 20s)?

tests/integration/v3_watch_restore_test.go
fuweid: Thanks. I will update it later.

fuweid:
> The lineCountExpecter and testingLogfMonitor seem like a generic solution, should be moved into test framework, so that it can be reused.

Hi @ahrtr, I can do it in the follow-up. This is to fix the flaky test. I want to keep it small. And I also want to add function helper for in-process member so that we can capture the log for a given member, instead of test case level. Does it make senses to you?

> But my immediate feeling is it's a little over complicated, can we just increase the [timeout](https://github.com/etcd-io/etcd/blob/3393d135679ecdb60d3d19f86b939e276c6ed096/tests/integration/v3_watch_restore_test.go#L149) (e.g. to 20s)? And also wait more time when receiving events: (1) add a timeout, e.g. 15s, and (2) cache the partial event.

Basically, the problem is that after recover network partition, the leader hasn't compacted the raft log yet. The member 0 receives the heartbeat but it will refuse the appendEntries and return hint (Term 2, Index: 8). And then leader still can find the (index:8) so that leader applies the index:(9,10,...) to member 0 instead of snapshot.

We can sleep few seconds before recover network partition. But it might be flaky somehow and the case takes more time.
And there is no way to check the server has compacted raft log without watching log. So based on my concern, I think we should use expect-style to verify log, just like what we do in e2e. WDYT?

ahrtr:
> > The lineCountExpecter and testingLogfMonitor seem like a generic solution, should be moved into test framework, so that it can be reused.
> 
> Hi @ahrtr, I can do it in the follow-up.

OK.



> Basically, the problem is that after recover network partition, the leader hasn't compacted the raft log yet.

Got it. The reason is etcdserver creates snapshot & compact raft log asynchronously.

ahrtr:
Overall looks good to me. Great finding. @fuweid Please resolve the minor comments.

tests/integration/v3_watch_restore_test.go
fuweid: Updated!

tests/integration/v3_watch_restore_test.go
fuweid: Updated! Please take a look. Thanks!

chaochn47:
LGTM.

FYI: Robustness test triggering snapshot can also be merged into raft log compaction log assertion style if the 100 milliseconds wait turns out to be flaky. 
https://github.com/etcd-io/etcd/blob/1227754284a0f452e7ea006288797a93da0c7afa/tests/robustness/failpoints.go#L422-L438

chaochn47 (approved):

tests/integration/v3_watch_restore_test.go
ahrtr: > However, the watch event is not expected. cc @ahrtr

Sorry, I do not get the point, and also did not get time dig into it for now. Please file a separate issue if you think it's an issue or something you can't explain or understand.

tests/integration/v3_watch_restore_test.go
```diff
@@ -154,3 +185,89 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+}
+
+func newTestingLogfMonitor(tb testutil.TB) *testingLogfMonitor {
+	return &testingLogfMonitor{
+		TB:          tb,
+		subscribers: make(map[string]testingLogfSubscriber),
+	}
+}
+
+func (m *testingLogfMonitor) addSubscriber(id string, sub testingLogfSubscriber) {
```
ahrtr, line 249: minor comment: change `addSubscriber` to `register`, and `delSubscriber` to `deregister`

tests/integration/v3_watch_restore_test.go
```diff
@@ -154,3 +185,89 @@ func TestV3WatchRestoreSnapshotUnsync(t *testing.T) {
...
+	m.subscribers[id] = sub
+}
+
+func (m *testingLogfMonitor) delSubscriber(id string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	delete(m.subscribers, id)
+}
+
+func (m *testingLogfMonitor) Logf(format string, args ...interface{}) {
```
ahrtr, line 261: As a generic solution, probably we should implementa `Log` as well.

ahrtr (approved):
LGTM with a couple of minor comments, which can be resolved in a followup PR.

fuweid:
Thanks @ahrtr @chaochn47 @lavacat for the review. I will file a pr for the followup.

PR #15672, opened 2023-04-09
chore: cleanup #15514 (Rewrite etcd scripts in strict mode)

fuweid:
* [chore: use tools/mod to lock the cfssl cmd version](https://github.com/etcd-io/etcd/pull/15672/commits/9f034fbaa8311e78e7e34fb80a2e0131886238b8)
* [chore: add strict mode for pkg/proxy/fixtures/gencerts.sh](https://github.com/etcd-io/etcd/commit/b3cb4ec771648ec57fafe47d15812fa6f2f6411d)
* [chore: add strict mode to scripts/test_lib.sh](https://github.com/etcd-io/etcd/commit/85fa2b1c6b0a8e092ed6e0750d112d3e9aad5532)
* [chore: use strict mode for tests/*/*.sh](https://github.com/etcd-io/etcd/commit/68fe3f3a1d2a11fd525d854a09edb2d7519ba9cf)
* [chore: deprecate tests/manual folder](https://github.com/etcd-io/etcd/commit/9717251934a0fa6e7811c7f5d7108dd6086991b5)


serathius (approved):
LGTM. Great cleanup! 

Some background, those scripts were already broken when I joined the project. We left them so someone maybe will want to fix them, however no-one wanted to use manual tests and robustness tests automated majority of the tests done here.

fuweid:
Rebased! Mark it ready to review.

fuweid:
ping @serathius @ahrtr ~

chaochn47:
> Some background, those scripts were already broken when I joined the project. We left them so someone maybe will want to fix them, however no-one wanted to use manual tests

I don't see the commit was justified with any test results or associated with a PR with discussion https://github.com/etcd-io/etcd/commit/a2256a6f24115237f0046435d67e5548fe12744a. 

I agree we should delete not maintained manual test scripts and in favor of automated and verified with each CI.

chaochn47 (approved):

pkg/proxy/fixtures/gencerts.sh
```diff
@@ -1,12 +1,21 @@
...
+
 if ! [[ "$0" =~ "./gencerts.sh" ]]; then
   echo "must be run from 'fixtures'"
   exit 255
 fi
 
-if ! which cfssl; then
+if ! command -v cfssl; then
   echo "cfssl is not installed"
+  echo "use: go install -mod mod github.com/cloudflare/cfssl/cmd/cfssl"
```
ahrtr, line 12: FYI. `-mod=mod tells the go command to ignore the vendor directory and to [automatically update](https://go.dev/ref/mod#go-mod-file-updates) go.mod, for example, when an imported package is not provided by any known module.`. See https://go.dev/ref/mod#build-commands.

This comment applies to a couple of places below.

```suggestion
  echo "use: go install github.com/cloudflare/cfssl/cmd/cfssl"
```

fuweid: You're right. It will dirty the go.mod if the user runs the command in the repo. I updated with your suggestion. Please take a look. Thanks!

fuweid:
>  cluster_test.go:326: expect error to contain can only promote a learner member, got etcdserver: request timed out

TestMemberPromoteMemberNotLearner is flaky as known issue. Need a maintainer to retry~ Thanks

pkg/proxy/fixtures/gencerts.sh
ahrtr: Please consider to install a specific version of tool (e.g. cfssl in this case). 

Usually we specify the version in tools/mod:
1. import the package in tools/mod/tools.go;
2. specify the version in tools/mod/go.mod;
3. Execute `cd tools/mod; go install github.com/cloudflare/cfssl/cmd/cfssl`

This can be resolved in a separate PR.

ahrtr (approved):
LGTM with a minor comment, which can be resolved separately.

Thank you @fuweid

pkg/proxy/fixtures/gencerts.sh
fuweid: Sure! That makes sense. Will file other pr later.

ahrtr, line 12: It still has "`-mod mod`"
```suggestion
  echo "use: go install github.com/cloudflare/cfssl/cmd/cfssl"
```

pkg/proxy/fixtures/gencerts.sh
```diff
@@ -1,12 +1,21 @@
...
+
 if ! [[ "$0" =~ "./gencerts.sh" ]]; then
   echo "must be run from 'fixtures'"
   exit 255
 fi
 
-if ! which cfssl; then
+if ! command -v cfssl; then
   echo "cfssl is not installed"
+  echo 'use: bash -c "cd ../../../tools/mod; go install github.com/cloudflare/cfssl/cmd/cfssl"'
```
ahrtr, line 12: This comment applies to some places below as well. 

Note that this script may be executed at any directory, so set the absolute directory is better.
```suggestion
  echo 'use: bash -c "cd ${ETCD_ROOT_DIR}/tools/mod; go install github.com/cloudflare/cfssl/cmd/cfssl"'
```

fuweid: This script is only allowed to run in current dir.

```bash
if ! [[ "$0" =~ "./gencerts.sh" ]]; then
  echo "must be run from 'fixtures'"
  exit 255
fi
```

So I was thinking that the user can copy-parse the echo string to install.
If we use ETCD_ROOT_DIR, the user needs to get the root dir path first.

I can add one line to resolve ETCD_ROOT_DIR first.

ahrtr: OK, then let's keep it as it is for now.

Issue #15666, opened 2023-04-07
TestTxnPanics need to cancel testcase from sending error to error channel


Mskxn:
### What happened?

This `test` has 6 testcases and run them one by one, with a select stmt to handle the timeout, but timeout can occurs no because the testcase is not panic, but is blocking for some reason.When this occurs, the errors in `errc` is not in the order as expected. Can we cancel the `errc <- s.(string)` in `df` when timeout?
https://github.com/etcd-io/etcd/blob/0bdc660ec2a8c6c3de03d78dfc96343d9c2068b3/client/v3/txn_test.go#L95-L105

### What did you expect to happen?

If the testcase is failed because of timeout, do not send error to `errc`, so the remaining testcases can work as expected.

### How can we reproduce it (as minimally and precisely as possible)?

Add some delay to any of these testcases to make the timeout error occurs.

### Etcd version (please run commands below)

<details>

```console

```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
```

</details>

### Relevant log output

```shell
bug report:

txn_test.go:103: #0: did not panic, wanted panic cannot call If twice!
txn_test.go:103: #1: did not panic, wanted panic cannot call If after Then!
txn_test.go:103: #2: did not panic, wanted panic cannot call If after Else!
txn_test.go:103: #3: did not panic, wanted panic cannot call Then twice!
txn_test.go:103: #4: did not panic, wanted panic cannot call Then after Else!
txn_test.go:100: #5: got cannot call If twice!, wanted cannot call Else twice!
```
```

lavacat:
Tried adding a delay and got failure as expected. @Mskxn am I reproducing not as you've intended?

>make test-unit
...
?   	go.etcd.io/etcd/client/pkg/v3/verify	[no test files]
% (cd client/internal/v2 && 'env' 'ETCD_VERIFY=all' 'go' 'test' './...' '-short' '-timeout=3m' '--race')
ok  	go.etcd.io/etcd/client/v2	0.115s
% (cd client/v3 && 'env' 'ETCD_VERIFY=all' 'go' 'test' './...' '-short' '-timeout=3m' '--race')
--- FAIL: TestTxnPanics (1.00s)
    txn_test.go:104: #2: did not panic, wanted panic cannot call If after Else!
FAIL
[...]

PR #15675, opened 2023-04-09
fix: different errc to avoid confused when timeout

Mskxn:
This pr is to fix issue #15666, using different errc for each testcase to avoid being confused when some of the testcases get a timeout.

### DIFF
--- a/client/v3/txn_test.go
+++ b/client/v3/txn_test.go
@@ -27,8 +27,7 @@ func TestTxnPanics(t *testing.T) {
 
 	kv := &kv{}
 
-	errc := make(chan string, 6)
-	df := func() {
+	df := func(errc chan string) {
 		if s := recover(); s != nil {
 			errc <- s.(string)
 		}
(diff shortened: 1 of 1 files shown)


lavacat:
Looks good, but one nitpick - can you change commit msg?
Part before `:` is usually a component name. My suggestion:
`test: use separate errc for each case in TestTxnPanics`

jmhbnz (commented):
Looks good, nice work @Mskxn.

Please address above feedback regarding commit message, additionally please ensure your updated commit is signed so the DCO check passes, i.e:

```bash
$ git reset --soft HEAD~1
$ git commit --signoff -m "tests: use separate errc for each case in TestTxnPanics"
$ git push --force-with-lease
```

ahrtr:
Thanks @Mskxn for the fix, which makes sense to me. Please resolve the above comments, then it's ready to merge after passing the workflow tests.

BTW, the simpler way to signoff the commit: 
```
$ git rebase HEAD~1 --signoff
$ git push origin main -f
```

ahrtr (approved):
LGTM

Thanks @Mskxn

Mskxn:
Thanks for your suggestion, I will pay more attention to these details in my subsequent pr.

Issue #15528, opened 2023-03-20
Flaky Test TestMemberPromoteMemberNotLearner


chaochn47:
### Which github workflows are flaking?

Integration Test

### Which tests are flaking?

TestMemberPromoteMemberNotLearner

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4469937666/jobs/7852763845

second time: https://github.com/etcd-io/etcd/actions/runs/4548835671/jobs/8020273443?pr=15580

### Reason for failure (if possible)

```
     logger.go:130: 2023-03-20T15:29:08.541Z	WARN	m2.client	retrying of unary invoker failed	{"member": "m2", "target": "etcd-endpoints://0xc00044b0e0/localhost:m2", "method": "/etcdserverpb.Cluster/MemberPromote", "attempt": 0, "error": "rpc error: code = FailedPrecondition desc = etcdserver: can only promote a learner member"}

     logger.go:130: 2023-03-20T15:29:08.542Z	WARN	m2.client	retrying of unary invoker failed	{"member": "m2", "target": "etcd-endpoints://0xc00044b0e0/localhost:m2", "method": "/etcdserverpb.Cluster/MemberPromote", "attempt": 0, "error": "rpc error: code = FailedPrecondition desc = etcdserver: can only promote a learner member"}

     logger.go:130: 2023-03-20T15:29:13.743Z	WARN	m2.client	retrying of unary invoker failed	{"member": "m2", "target": "etcd-endpoints://0xc00044b0e0/localhost:m2", "method": "/etcdserverpb.Cluster/MemberPromote", "attempt": 0, "error": "rpc error: code = Unavailable desc = etcdserver: request timed out"}

     cluster_test.go:326: expect error to contain can only promote a learner member, got etcdserver: request timed out
```

The test just created 3 member cluster and expect member promote 

### Anything else we need to know?

Also has been caught in https://github.com/etcd-io/etcd/issues/12372

https://github.com/etcd-io/etcd/blob/22bdc913025de198ef2759e9f534ed8dd555de66/server/etcdserver/server.go#L1487 is unnecessary if the local node is not leader. 

It could be a culprit but test cluster does not have much traffic.

chaochn47:
### Happy case

1. PromoteMember forward the request to leader node via HTTP https://github.com/etcd-io/etcd/blob/caed563e088adb6a381b6ff12fc190bf961ad2ea/server/etcdserver/server.go#L1406
[...]

PR #15696, opened 2023-04-11
tests/integration/clientv3: fix flaky TestMemberPromoteMemberNotLearner

chaochn47:
Fix https://github.com/etcd-io/etcd/issues/15528

Analysis is summarized in https://github.com/etcd-io/etcd/issues/15528#issuecomment-1502077331

> etcd server apply and raftLog.applied increment is asynchronous. Because the test tries member promote back to back 3 consecutive times immediately, the configuration change apply failed as expected but raftLog.applied may not yet advanced so raft rejects the next configuration change until the current pending one finishes. 

So Fix is to ensure raftLog.applied advancement

### DIFF
--- a/tests/integration/clientv3/util.go
+++ b/tests/integration/clientv3/util.go
@@ -16,14 +16,20 @@ package clientv3test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"go.etcd.io/etcd/client/pkg/v3/transport"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
 )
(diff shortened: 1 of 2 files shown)


tests/integration/clientv3/util.go
```diff
@@ -119,3 +125,62 @@ func populateDataIntoCluster(t *testing.T, cluster *integration2.Cluster, numKey
...
+				t.Logf("marshalled raft status is %s", string(rs))
+				require.NoError(t, json.Unmarshal(rs, &raftStatus))
+				break
+			}
+		}
+		return raftStatus.Applied == appliedIndex
+	}, 5*time.Second, 20*time.Millisecond)
+}
+
+// raftStatus should be in sync with raft.Status
```
lavacat, line 172: is it possible to reuse raft.Status without defining raftStatus?

tests/integration/clientv3/util.go
```diff
@@ -119,3 +125,62 @@ func populateDataIntoCluster(t *testing.T, cluster *integration2.Cluster, numKey
...
+		DialTimeout: 5 * time.Second,
+	})
+	require.NoError(t, err)
+	defer cc.Close()
+	cresp, err := cc.Status(context.TODO(), gRPCURL)
+	require.NoError(t, err)
+	appliedIndex := cresp.RaftAppliedIndex
+
+	require.Eventually(t, func() bool {
+		tr, err := transport.NewTransport(transport.TLSInfo{}, 5*time.Second)
```
lavacat, line 144: nit: I think getting raftStatus should be it's own helper method

lavacat (commented):
LGTM, minor comments

tests/integration/clientv3/util.go
chaochn47: Unfortunately not. The raft.Status does not provide an Unmarshal method. I think it can be raised as a follow up PR in etcd-it/raft.

tests/integration/clientv3/util.go
chaochn47: Yeah, I thought guaranteeRaftAppliedIndexAdvance can be reused to solve the similar issue. I will create a separate function for get raft status reuse. Thanks for the suggestion.

chaochn47: Updated in the new revision.

tests/integration/clientv3/util.go
ahrtr: I do not get this, why can't reuse `raft.Status`? Your `raftStatus` also doesn't have an `Unmarshal` method.

ahrtr:
I am thinking probably it would be better to expose & reuse [(*EtcdServer) RaftStatus()](https://github.com/etcd-io/etcd/blob/817aba57c797b2984fbdd785f736e05d98438833/server/etcdserver/server.go#L2444), the benefits are:
1. Greatly simplify the implementation;
2. Remove the dependency on endpoint `/debug/vars`.

tests/integration/clientv3/util.go
```diff
@@ -119,3 +125,67 @@ func populateDataIntoCluster(t *testing.T, cluster *integration2.Cluster, numKey
...
+	})
+
+	body, err := io.ReadAll(response.Body)
+	require.NoError(t, err)
+	var resp map[string]interface{}
+	require.NoError(t, json.Unmarshal(body, &resp))
+	var raftStatus raftStatus
+
+	for key, value := range resp {
+		if key == "raft.status" {
```
fuweid, line 166: what about `value, ok := resp["raft.status"]` here?

ahrtr:
Once the PR gets merged, this PR isn't needed any more.

chaochn47:
Close to PR since the work will be continued on https://github.com/etcd-io/etcd/pull/15708

PR #15703, opened 2023-04-12
etcdserver: wait for raft being notified on confChange before responding to client

ahrtr:
See discussion in https://github.com/etcd-io/etcd/issues/15528

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #15708, opened 2023-04-13
raft node notifies configure when confChanged

### LINKED ISSUE #15528: Flaky Test TestMemberPromoteMemberNotLearner

chaochn47:
Fix #15528

The PR is built on top of https://github.com/etcd-io/etcd/pull/15703

PTAL @ahrtr @serathius, thanks!

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1133,7 +1133,7 @@ func (s *EtcdServer) applyEntries(ep *etcdProgress, apply *toApply) {
 		return
 	}
 	var shouldstop bool
-	if ep.appliedt, ep.appliedi, shouldstop = s.apply(ents, &ep.confState, apply.confChangeCh); shouldstop {
+	if ep.appliedt, ep.appliedi, shouldstop = s.apply(ents, &ep.confState, apply.raftAdvancedC); shouldstop {
 		go s.stopWithDelay(10*100*time.Millisecond, fmt.Errorf("the member has been permanently removed from the cluster"))
 	}
 }
@@ -1649,8 +1649,9 @@ func (s *EtcdServer) AppliedIndex() uint64 { return s.getAppliedIndex() }
 func (s *EtcdServer) Term() uint64 { return s.getTerm() }
 
 type confChangeResponse struct {
-	membs []*membership.Member
-	err   error
+	membs        []*membership.Member
+	raftAdvanceC <-chan struct{}
+	err          error
 }
 
 // configure sends a configuration change through consensus and
@@ -1673,6 +1674,11 @@ func (s *EtcdServer) configure(ctx context.Context, cc raftpb.ConfChange) ([]*me
 			lg.Panic("failed to configure")
 		}
 		resp := x.(*confChangeResponse)
+		// etcdserver need to ensure the raft has already been notified
+		// or advanced before it responds to the client. Otherwise, the
+		// following config change request may be rejected.
+		// See https://github.com/etcd-io/etcd/issues/15528.
+		<-resp.raftAdvanceC
 		lg.Info(
 			"applied a configuration change through raft",
 			zap.String("local-member-id", s.MemberId().String()),
@@ -1810,7 +1816,7 @@ func (s *EtcdServer) sendMergedSnap(merged snap.Message) {
 func (s *EtcdServer) apply(
 	es []raftpb.Entry,
 	confState *raftpb.ConfState,
-	confChangeCh chan struct{},
+	raftAdvancedC <-chan struct{},
 ) (appliedt uint64, appliedi uint64, shouldStop bool) {
 	s.lg.Debug("Applying entries", zap.Int("num-entries", len(es)))
 	for i := range es {
@@ -1842,19 +1848,7 @@ func (s *EtcdServer) apply(
 			s.setAppliedIndex(e.Index)
 			s.setTerm(e.Term)
 			shouldStop = shouldStop || removedSelf
-
-			// etcdserver need to ensure the raft has already been notified
-			// or advanced before it responds to the client. Otherwise, the
-			// following config change request may be rejected.
-			// See https://github.com/etcd-io/etcd/issues/15528.
-			select {
-			case <-time.After(500 * time.Millisecond):
-				lg := s.Logger()
-				lg.Warn("timed out waiting for configChange notification")
-			case <-confChangeCh:
-			}
-
-			s.w.Trigger(cc.ID, &confChangeResponse{s.cluster.Members(), err})
+			s.w.Trigger(cc.ID, &confChangeResponse{s.cluster.Members(), raftAdvancedC, err})
 
 		default:
 			lg := s.Logger()
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1133,7 +1133,7 @@ func (s *EtcdServer) applyEntries(ep *etcdProgress, apply *toApply) {
 		return
 	}
 	var shouldstop bool
-	if ep.appliedt, ep.appliedi, shouldstop = s.apply(ents, &ep.confState); shouldstop {
+	if ep.appliedt, ep.appliedi, shouldstop = s.apply(ents, &ep.confState, apply.confChangeCh); shouldstop {
 		go s.stopWithDelay(10*100*time.Millisecond, fmt.Errorf("the member has been permanently removed from the cluster"))
 	}
 }
@@ -1810,6 +1810,7 @@ func (s *EtcdServer) sendMergedSnap(merged snap.Message) {
 func (s *EtcdServer) apply(
 	es []raftpb.Entry,
 	confState *raftpb.ConfState,
+	confChangeCh chan struct{},
 ) (appliedt uint64, appliedi uint64, shouldStop bool) {
 	s.lg.Debug("Applying entries", zap.Int("num-entries", len(es)))
 	for i := range es {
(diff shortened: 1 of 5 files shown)


ahrtr:
Can https://github.com/etcd-io/etcd/issues/15528 be easily reproduced? If not, then we need to add an integration test; we can intentionally inject a sleep failpoint before [r.Advanced()](https://github.com/etcd-io/etcd/blob/22f3e50adafd9d4cf9dd29dd5837483a6417238c/server/etcdserver/raft.go#L310)

chaochn47:
> Can #15528 be easily reproduced? If not, then we need to add an integration test; we can intentionally inject a sleep failpoint before [r.Advanced()](https://github.com/etcd-io/etcd/blob/22f3e50adafd9d4cf9dd29dd5837483a6417238c/server/etcdserver/raft.go#L310)

Not easily. Inject a sleep should do the trick.

server/etcdserver/server.go
```diff
@@ -107,6 +107,8 @@ const (
 	readyPercent = 0.9
 
 	DowngradeEnabledPath = "/downgrade/enabled"
+
+	maxRaftAdvanceDelay = 500 * time.Millisecond
```
serathius, line 111: Where this variable comes from?

fuweid:
It's hard to reproduce in few requests. It needs to send request one by one without pause. 
FYI: You can use the following test code to reproduce it.  

```go
package verify

import (
        "context"
        "strings"
        "testing"
        "time"

        integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
        "go.etcd.io/etcd/tests/v3/integration"
)

var lazyCluster = integration.NewLazyClusterWithConfig(
        integration2.ClusterConfig{
                Size:                        3,
                WatchProgressNotifyInterval: 200 * time.Millisecond,
                DisableStrictReconfigCheck:  true})

func TestVerify(t *testing.T) {
        defer lazyCluster.Terminate()

        clus := lazyCluster.Cluster()
        leaderIdx := clus.WaitLeader(t)
        members := clus.Members

        followerIdx := (leaderIdx + 1) % 3
        cli := clus.Client(leaderIdx)

        // promoting any of the voting members in cluster should fail
        expectedErrKeywords := "can only promote a learner member"
        var err error
        for i := 0; i < 10000; i++ {
                ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
                _, err = cli.MemberPromote(ctx, uint64(members[followerIdx].ID()))
                cancel()
                if err == nil {
                        t.Fatalf("expect promoting voting member to fail, got no error")
                }

                if strings.Contains(err.Error(), expectedErrKeywords) {
                        continue
                }

                t.Fatalf("unexpected error: %v", err)
        }
}

```

server/etcdserver/server.go
```diff
@@ -1833,6 +1836,18 @@ func (s *EtcdServer) apply(
 			s.setAppliedIndex(e.Index)
 			s.setTerm(e.Term)
 			shouldStop = shouldStop || removedSelf
+
+			// etcdserver need to ensure the raft has already been notified
+			// or advanced before it responds to the client. Otherwise, the
+			// following config change request may be rejected.
+			// See https://github.com/etcd-io/etcd/issues/15528.
+			select {
+			case <-time.After(maxRaftAdvanceDelay):
```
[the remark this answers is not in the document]
chaochn47: Second thought, from linearizability perspective, we should not. It was carried over from the original commit and I was thinking member promote response can be still be sent out even with slow disk with fsync latency because it has already applied even with failure.

server/etcdserver/server.go
chaochn47: Will remove this hardcoded timeout value

serathius:
This CL not only changes MemberPromote, but introduces a big change for all membership changes. We should double check that all configuration changes are properly tested. If not we should add a large set of e2e tests that:
* Grow cluster from 1  to 3 members (without learners)
* Grow cluster from 1 to 3 members (with learners) in both 1 at the time and two at one time scenario
* Reduce cluster from 3 to 1 member 
etc

chaochn47:
> Can #15528 be easily reproduced? If not, then we need to add an integration test; we can intentionally inject a sleep failpoint before [r.Advanced()](https://github.com/etcd-io/etcd/blob/22f3e50adafd9d4cf9dd29dd5837483a6417238c/server/etcdserver/raft.go#L310)

Hi @ahrtr could you please help clarify how would integration test inject a sleep failpoint? I know with etcd binary built with gofail enabled it is achievable. https://github.com/etcd-io/etcd/blob/b75499437b8eb5d0e027b3abfb19932359aebf9b/tests/framework/e2e/etcd_process.go#L298-L318 Thanks to @serathius as part of the robustness test work.

ahrtr:
> Hi @ahrtr could you please help clarify how would integration test inject a sleep failpoint? I know with etcd binary built with gofail enabled it is achievable.

Please refer to the followings,
1. the section "Test" in https://docs.google.com/document/d/1U9hAcZQp3Y36q_JFiw2VBJXVAo2dK2a-8Rsbqv3GgDo/edit#heading=h.wwktogcu1ekk. 
2. https://github.com/etcd-io/bbolt/blob/97d2cc34ac943a03ca596d15e0629b163468f59f/Makefile#L77

chaochn47:
Ping @serathius @ahrtr @fuweid @tjungblu @jmhbnz for review ~

jmhbnz (commented):
A comment to respond to review ping. I've read through this pr a few times and at a code level it looks good, however I'm abstaining from hitting approve as I don't understand overall system implications well enough.

serathius (approved):

tjungblu (approved):
/lgtm (non-binding)

Thanks for finally fixing this!

server/etcdserver/server.go
```diff
@@ -1673,6 +1674,11 @@ func (s *EtcdServer) configure(ctx context.Context, cc raftpb.ConfChange) ([]*me
 			lg.Panic("failed to configure")
 		}
 		resp := x.(*confChangeResponse)
+		// etcdserver need to ensure the raft has already been notified
+		// or advanced before it responds to the client. Otherwise, the
+		// following config change request may be rejected.
+		// See https://github.com/etcd-io/etcd/issues/15528.
+		<-resp.raftAdvanceC
```
ahrtr, line 1681: This can work, but I'd suggest not to change the API layer. Let's get all the details encapsulated on the raft & etcd application layer, and the API layer should be agnostic to the raft & application side details. 

Please try to follow my original PR https://github.com/etcd-io/etcd/pull/15703/files

chaochn47: The original PR waits for the added `raftAdvanceC`, however `raftAdvanceC` will not be sent because it is blocked until 'applyAll' calls 'applyWait.Trigger'. It's a deadlock and will be stuck forever. 

https://github.com/etcd-io/etcd/blob/6cdc9ae4fe1258247de1f7de372b3e8f0f0ea555/server/etcdserver/raft.go#L293-L310

It can be easily reproduced in [TestEnableAuth](https://github.com/etcd-io/etcd/blob/e9edc5432cffce26d53906528ecd125dfae3a699/server/embed/auth_test.go#L24) with https://github.com/etcd-io/etcd/commit/2d4c8eca3b3e51e2b3c43beefa35279955ea7c3d. In a single-node cluster, its first entry and committed entry type is `confChangeAdd` will be blocked in apply forever.

chaochn47: I am open to considering other solutions, but I believe that it should be fixed just like this PR.

fuweid (approved):
LGTM

Thanks!

server/etcdserver/server.go
ahrtr: Got it, thx for the clarification. 

But I think the implementation is slowly rotting. It's getting harder to understand and maintain. We should consider simplifying the implementation per my above comment https://github.com/etcd-io/etcd/pull/15708#discussion_r1243435646.

ahrtr (approved):

server/etcdserver/server.go
chaochn47: Agreed. I feel it's inevitable because of the `r.Advance()` is invoked after `r.transport.Send(msgs)`. However, I doubt the reason of this ordering and pending for investigation. 

Changing the order might be risky so I lean towards to this fix (just for now).

PR #15713, opened 2023-04-13
Add issue template for new membership nominations.

jmhbnz:
This pull request proposes adding a new github issue template for nominating new etcd org members. Currently I have only set this up to cater for the base "Member" role as this would be the one happening on the most frequent basis versus adding reviewers or maintainers.

The issue template is partially inspired by the kubernetes github organisation membership template but adjusted to be from the angle of a reviewer or maintainer raising the nomination instead of the member.

The intent of the template is to save a small amount of the creators time and ensure issues automatically have the right labels and follow a consistent approach.

Please let me know what you think of the suggested layout, happy to tweak it!

### DIFF
--- a/.github/ISSUE_TEMPLATE/membership-request.yml
+++ b/.github/ISSUE_TEMPLATE/membership-request.yml
@@ -1,0 +1,30 @@
+name: Membership nomination
+description: Nominate new etcd members
+labels:
+- area/community
+body:
+- type: textarea
+  id: feature
+  attributes:
+    label: Who would you like to nominate?
+  validations:
+    required: true
+
+- id: requirements
+  type: checkboxes
+  attributes:
+    label: Requirements
+    options:
+    - label: I have reviewed the [community membership guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/community-membership.md)
+      required: true
+    - label: The members are actively contributing to 1 or more etcd subprojects
+      required: true
+    - label: The members are being sponsored by two current reviewers or a current maintainer.
+      required: true
+
+- type: textarea
+  id: rationale
+  attributes:
+    label: How do the new members meet the regular active contribution requirements?
+  validations:
+    required: true
--- a/Documentation/contributor-guide/community-membership.md
+++ b/Documentation/contributor-guide/community-membership.md
@@ -28,7 +28,7 @@ Members are continuously active contributors in the community.  They can have
 issues and PRs assigned to them. Members are expected to remain active 
 contributors to the community.
 
-**Defined by:** Member of the etcd GitHub organization
+**Defined by:** Member of the etcd GitHub organization.
 
 ### Requirements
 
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/membership-request.yml
```diff
@@ -0,0 +1,35 @@
+name: Membership nomination
+description: Nominate new etcd members
+labels:
+- type/feature
```
serathius, line 4:
```suggestion
```

ptabor (approved):
Thank you.

.github/ISSUE_TEMPLATE/membership-request.yml
```diff
@@ -0,0 +1,35 @@
...
+    required: true
+
+- id: requirements
+  type: checkboxes
+  attributes:
+    label: Requirements
+    options:
+    - label: I have reviewed the [community membership guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/community-membership.md)
+      required: true
+    - label: The members I am nominating have [enabled 2FA on their GitHub account](https://github.com/settings/security)
```
ptabor, line 20: I'm only not sure whether we should require from the sponsor, confirming whether formal requirements are full-filed by the nominee.  I would expect this to be self-declaration of nominee that happens after decision but probably before adding to the membership group.

serathius:
Similar to K8s, should we document using this template in the https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/community-membership.md

ptabor (approved):

.github/ISSUE_TEMPLATE/membership-request.yml
jmhbnz: Fair point, I've taken the 2fa and dev mailing list tickboxes off for now as expectation is on member to self manage these and not put burden on sponsors to verify.

We can revisit this down the track if we shift from reviewer/maintainer nomination to self nomination.

Documentation/contributor-guide/community-membership.md
```diff
@@ -28,7 +28,7 @@ Members are continuously active contributors in the community.  They can have
 issues and PRs assigned to them. Members are expected to remain active 
 contributors to the community.
 
-**Defined by:** Member of the etcd GitHub organization
+**Defined by:** Member of the etcd GitHub organization, added via [membership nomination] issue template.
```
serathius, line 31: It's not defined by it, please add it to requirements like in https://github.com/kubernetes/community/blob/master/community-membership.md

ahrtr (approved):
Thank you @jmhbnz

Documentation/contributor-guide/community-membership.md
jmhbnz: Sure thing - have adopted k/org style approach in requirements 👍🏻

serathius (approved):

PR #15715, opened 2023-04-14
[3.4] fix release.sh: git_assert_branch_in_sync not exist in 3.4

ahrtr:
The `git_assert_branch_in_sync` doesn't exist in 3.4, so let's just remove it.

The issue was introduced in https://github.com/etcd-io/etcd/commit/49d05f88c31c71a48368bf20927cd8ba61a3bf1a. In our workflow, it always run in `in-place` mode, so it never triggers the branch.

### DIFF
--- a/scripts/build-binary.sh
+++ b/scripts/build-binary.sh
@@ -6,7 +6,7 @@ source ./scripts/test_lib.sh
 
 VER=$1
 PROJ="etcd"
-REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd.git}"
+REPOSITORY="${REPOSITORY:-https://github.com/etcd-io/etcd.git}"
 
 if [ -z "$1" ]; then
 	echo "Usage: ${0} VERSION" >> /dev/stderr
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -69,8 +69,6 @@ main() {
     cd "${reldir}/etcd" || exit 2
     git checkout "${BRANCH}" || exit 2
     git pull origin
-
-    git_assert_branch_in_sync || exit 2
   fi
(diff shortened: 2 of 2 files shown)


ahrtr:
cc @serathius PTAL

serathius:
Should we backport it then?

scripts/build-binary.sh
```diff
@@ -6,7 +6,7 @@ source ./scripts/test_lib.sh
 
 VER=$1
 PROJ="etcd"
-REPOSITORY="${REPOSITORY:-git@github.com:etcd-io/etcd.git}"
+REPOSITORY="${REPOSITORY:-https://github.com/etcd-io/etcd.git}"
```
serathius, line 9: Note that also applies on other branches. Not as important for this PR.

I prefer git@github.com:etcd-io/etcd.git as during push I can depend on SSH security instead of typing in credentials. It's much safer. Best would be to take address of `origin` remote from original repo.

ahrtr: I agree. But in 3.4, actually I never depend on the automatically `git push`. Instead, each time, I manually bump the version and create/push the tag. 

Let's just keep it as it was. It was working well previous. After the https://github.com/etcd-io/etcd/commit/49d05f88c31c71a48368bf20927cd8ba61a3bf1a, it broke the 3.4 release, so I tend to partially revert the change. 


We can consider to enhance it later; but I don't expect to make much big change on 3.4.

serathius (approved):

ahrtr:
thx @serathius 

I will not allow any release script change anymore in future, so as not to break it again. Because usually our workflow/pipeline will never test the real release execution path.

PR #15722, opened 2023-04-17
test: rename `tests/framework/testutils` to `tests/framework/utils`

ahrtr:
Please read https://go.dev/doc/effective_go#package-names

### DIFF
--- a/tests/common/alarm_test.go
+++ b/tests/common/alarm_test.go
@@ -23,7 +23,7 @@ import (
 
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/config"
-	"go.etcd.io/etcd/tests/v3/framework/testutils"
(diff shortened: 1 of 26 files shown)


ahrtr:

jmhbnz (approved):
Based on the go guidelines this change makes sense to me.

One question though, are we aware of anything external to etcd that might rely on this? I would think unlikely given this is just a project testing component?

ahrtr:
> One question though, are we aware of anything external to etcd that might rely on this?

I don't think we guarantee API compatibility for test package.  Also note that [test framework](https://github.com/etcd-io/etcd/tree/main/tests/framework) is only present in `main` branch.

ptabor:
The motivation for 'testutils' is to have ability to import the packages to a lot of tests without need to alias. 
'utils' is a very ubiquitous names so requires picking an alias name that might be difference between different tests.

ahrtr:
thx  @ptabor

Issue #15674, opened 2023-04-09
Reduce maintenance burden for go patch releases


jmhbnz:
### What would you like to be added?

Currently we specify which go patch version we build etcd with in 13 different files:

```bash
.github/workflows/tests-arm64.yaml:        go-version: "1.19.8"
.github/workflows/tests.yaml:        go-version: "1.19.8"
.github/workflows/govuln.yaml:          go-version: "1.19.8"
.github/workflows/grpcproxy.yaml:        go-version: "1.19.8"
.github/workflows/e2e.yaml:        go-version: "1.19.8"
.github/workflows/static-analysis.yaml:          go-version: "1.19.8"
.github/workflows/robustness-template.yaml:        go-version: '1.19.8'
.github/workflows/build.yaml:          go-version: "1.19.8"
.github/workflows/release.yaml:        go-version: "1.19.8"
.github/workflows/e2e-arm64.yaml:        go-version: "1.19.8"
.github/workflows/coverage.yaml:        go-version: "1.19.8"
.github/workflows/fuzzing.yaml:        go-version: "1.19.8"
.github/workflows/contrib.yaml:        go-version: "1.19.8"
````

Whenever a new go patch release comes out we have updated rapidly via pull request to ensure we are addressing CVE's, for example: #15329 #15427 #15649

I would like to propose that we manage this more autonomously by updating our workflows as follows:

```yaml
steps:
  - uses: actions/checkout@<hash>
  - uses: actions/setup-go@<hash>
    with:
      go-version-file: 'go.mod'
```

As outlined in the [setup-go action docs](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file) the `go-version-file` input accepts a path to a `go.mod` file or a `go.work` file that contains the version of Go to be used by a project. As the `go.mod` file contains only major and minor (e.g. 1.19) tags, the action will search for the latest available patch version sequentially in the runner's directory with the cached tools, in the [versions-manifest.json](https://github.com/actions/go-versions/blob/main/versions-manifest.json) file or at the go servers.

### Why is this needed?

This would reduce maintenance burden for etcd project, and also increase speed of adoption for new patch releases.

This also means we retain full decision making via pull request for the more important decision of when adopt new go major or minor releases and simplifies this process as we would just update `go.mod` files.

[...]

PR #15748, opened 2023-04-20
Adds a reusable workflow to setup the GoVersion

judavi:
Adds a reusable workflow to setup the GoVersion. 
The output of this workflow will be used in the other workflows to reduce maintenance burden for the project  and also increase speed of adoption for new patch releases.

Related to https://github.com/etcd-io/etcd/issues/15674

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,8 +3,11 @@ on: [push, pull_request]
 permissions: read-all
 
 jobs:
+  goversion:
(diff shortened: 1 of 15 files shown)


jmhbnz (commented):
Thanks for solving this @judavi! This will make future go patch releases easier to handle.

Overall I think this change looks great - @serathius can you please enable workflows for this so we can double check?

serathius (approved):

serathius:
Thanks for contributing!

Issue #15758, opened 2023-04-22
tests/common: TestAuthMemberRemove


fuweid:
### Which github workflows are flaking?

Tests

### Which tests are flaking?

TestAuthMemberRemove

<details>
<summary>log</summary>
=== FAIL: common TestAuthMemberRemove (0.30s)
    ...

    logger.go:130: 2023-04-15T15:17:43.162Z	INFO	grpc	[[core] [Channel #355] parsed dial target is: {Scheme:etcd-endpoints Authority:0xc00039b680 Endpoint:localhost:m0 URL:{Scheme:etcd-endpoints Opaque: User: Host:0xc00039b680 Path:/localhost:m0 RawPath: OmitHost:false ForceQuery:false RawQuery: Fragment: RawFragment:}}]
    logger.go:130: 2023-04-15T15:17:43.162Z	INFO	grpc	[[core] [Channel #355] Channel authority set to "localhost:m0"]
    logger.go:130: 2023-04-15T15:17:43.163Z	INFO	grpc	[[core] [Channel #355] Resolver state updated: {
          "Addresses": [
            {
              "Addr": "unix:localhost:m0",
              "ServerName": "localhost",
              "Attributes": null,
              "BalancerAttributes": null,
              "Type": 0,
              "Metadata": null
            },
            {
              "Addr": "unix:localhost:m1",
              "ServerName": "localhost",
              "Attributes": null,
              "BalancerAttributes": null,
              "Type": 0,
              "Metadata": null
            },
            {
              "Addr": "unix:localhost:m2",
              "ServerName": "localhost",
              "Attributes": null,
              "BalancerAttributes": null,
              "Type": 0,
              "Metadata": null
            }
          ],
          "ServiceConfig": {
            "Config": {
              "Config": null,
              "LB": "round_robin",
              "Methods": {}
            },
            "Err": null
          },
          "Attributes": null
        } (service config updated; resolver returned new addresses)]
    logger.go:130: 2023-04-15T15:17:43.163Z	INFO	grpc	[[core] [Channel #355] Channel switches to new LB policy "round_robin"]
    logger.go:130: 2023-04-15T15:17:43.163Z	INFO	grpc	[[balancer] base.baseBalancer: got new ClientConn state:  {{[{
          "Addr": "unix:localhost:m0",
          "ServerName": "localhost",
          "Attributes": null,
          "BalancerAttributes": null,
          "Type": 0,
          "Metadata": null
        } {
[...]

PR #15759, opened 2023-04-22
server/etcdserver: togRPCError for maintenance API

fuweid:
It's to deflake TestAuthMemberRemove.

When the client has multiple endpoints, the client might send a request with valid token to the follower member which hasn't received token replicated log yet. The member will reject the request.

For instance, the maintenance.Status API will return "auth: invalid auth token". But the client doesn't identify the error. The client won't retry to refresh auth token. The maintenance.Status should togRPCError before return so that the client can reflesh token. It's align with existing API.

Since the maintenance client always creates one connection to target member, the member will have the token after refresh auth.

Maybe we can introduce a sync to wait for member is ready with token, instead of refreshing.

Fixes: #15758

### DIFF
--- a/server/etcdserver/api/v3rpc/maintenance.go
+++ b/server/etcdserver/api/v3rpc/maintenance.go
@@ -90,7 +90,7 @@ func (ms *maintenanceServer) Defragment(ctx context.Context, sr *pb.DefragmentRe
 	err := ms.bg.Backend().Defrag()
 	if err != nil {
 		ms.lg.Warn("failed to defragment", zap.Error(err))
-		return nil, err
+		return nil, togRPCError(err)
 	}
 	ms.lg.Info("finished defragment")
 	return &pb.DefragmentResponse{}, nil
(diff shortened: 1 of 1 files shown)


lavacat (approved):
Thanks for finding this. I've tested locally and it fixed flakiness for me.

I think all grpc methods should convert GRPCError type. Wonder if there are any more places that missed this conversion. I'll open a separate issue to check

ahrtr (approved):
LGTM

ahrtr:
Thanks @fuweid for the fix. 

> Maybe we can introduce a sync to wait for member is ready with token, instead of refreshing.

For keys space reading and writing, etcd ensures linearizability. For maintenance APIs, most operations (e.g. defragmentation, Snapshot. etc) are just on the current endpoint instead of a cluster-wide operation, so it might be fine not to be linearizable. But I agree that it might run into permission denied issue just as you mentioned, I tend to keep it as it's because the client side can retry and usually maintenance operations are supposed to be performed by human (e.g. administrators).

serathius:
As long I think about the issue, more I think that whole etcd API should be linearizable including maintenance. Reason is that it just is not natural and makes it much harder to automate any etcd operations. I expect this has gone unnoticed as most cloud Kubernetes providers run etcd on VMs that already have a large operation overhead, however for someone using/building etcd operator this is problematic. This is unbecoming of cloud native project.

I would want to propose that as part of goal of making etcd easier to operate, we start an effort to make maintenance API linearizable. As mention on other comment, I would like to start by listing whole etcd API surface and identifying whether it's linearizable or not.

serathius:
cc @wojtek-t @logicalhan

PR #15762, opened 2023-04-24
refactor(util): remove duplicate lg check

aimuz:
lg always has a value

Signed-off-by: aimuz <mr.imuz@gmail.com>

### DIFF
--- a/server/etcdmain/util.go
+++ b/server/etcdmain/util.go
@@ -35,13 +35,11 @@ func discoverEndpoints(lg *zap.Logger, dns string, ca string, insecure bool, ser
 	}
 	endpoints := srvs.Endpoints
 
-	if lg != nil {
-		lg.Info(
-			"discovered cluster from SRV",
-			zap.String("srv-server", dns),
-			zap.Strings("endpoints", endpoints),
-		)
-	}
+	lg.Info(
+		"discovered cluster from SRV",
+		zap.String("srv-server", dns),
+		zap.Strings("endpoints", endpoints),
+	)
 
 	if insecure {
 		return *srvs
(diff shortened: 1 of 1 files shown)


serathius (changes requested):
We don't plan or we never discussed supporting other logging implementations. Proposed interface not part of any standard, just a workaround.

aimuz:
The implementation here is only to avoid duplicate judgments 

@serathius

aimuz:
Sorry for the misunderstanding of the wrong commit

server/etcdmain/util.go
```diff
@@ -35,14 +46,16 @@ func discoverEndpoints(lg *zap.Logger, dns string, ca string, insecure bool, ser
...
 	endpoints := srvs.Endpoints
 
-	if lg != nil {
-		lg.Info(
-			"discovered cluster from SRV",
-			zap.String("srv-server", dns),
-			zap.Strings("endpoints", endpoints),
-		)
+	if lg == nil {
+		lg = notLogger{}
```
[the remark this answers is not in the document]
aimuz: Thank you for your reminder. I didn't notice the presence of zap.NewNop in zap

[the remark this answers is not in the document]
aimuz: I have made modifications

ptabor (commented):
I think it's a step in the right direction. 

But really I would prefer to be at: 

```verify.Assert(lg != nil, "nil log isn't allowed")```

The motivation is to go to always 'verification' is that usually the code
that does not pass the loggers is unit test's code. 
And lack of logs in tests used to led to more difficult to diagnose flakes than they should be.

**Comment by ahrtr** [comment]
I would suggest to remove the` if lg == nil` directly, and only add assertion something like `verify.Assert(lg != nil, "nil log isn't allowed")` in user-facing (exported) methods/functions.

**Comment by aimuz** [comment]
Very good suggestion, I also see this usage in the code

File: server/etcdmain/util.go [hunk 2]
```diff
@@ -20,11 +20,14 @@ import (
 
 	"go.etcd.io/etcd/client/pkg/v3/srv"
 	"go.etcd.io/etcd/client/pkg/v3/transport"
+	"go.etcd.io/etcd/client/pkg/v3/verify"
 
 	"go.uber.org/zap"
 )
 
 func discoverEndpoints(lg *zap.Logger, dns string, ca string, insecure bool, serviceName string) (s srv.SRVClients) {
+	verify.Assert(lg != nil, "nil log isn't allowed")
```
ahrtr, line 29: Just as I mentioned, we should only add assertion like this in user-facing (exported) methods/functions. For internal functions & methods, we have 100% control over the usage, why should we add such assertion? 

```suggestion
```

aimuz: Indeed, I've investigated and no non-empty lg is passed in, just removing the validation is enough

aimuz: thank you

ahrtr (approved):
LGTM

Thanks @aimuz

serathius (approved):

PR #15783, opened 2023-04-25
Consolidate etcd dockerfiles

### LINKED ISSUE #15777: Merge per platform dockerfiles

We don't need per platform dockerfile, they are identical aside of  --platform flag which can be templated using `docker build --args`.

### Why is this needed?

Simpler maintenance.

jmhbnz:
We can consolidate by using a simple utility like `envsubst` with our existing `ARCH` variable to create the individual release files. This reduces some clutter from our etcd root directory. 

Fixes #15777


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 gcr.io/distroless/static-debian11
+FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian11
```
ahrtr, line 1:
```suggestion
ARG ARCH=amd64
FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian11
```

Use command something like below when building image,
```
docker build --build-arg ARCH=${ARCH} ...
```

ahrtr:
Thanks @jmhbnz for the PR, which I agree is the correct thing to do. Please resolve the comment and also fix the workflow failure of `Release`

serathius (approved):

PR #15790, opened 2023-04-27
Add -failfast flag when the mode is fail_fast

Rajalakshmi-Girish:
Though there is mention of mode `fail_fast`, at https://github.com/etcd-io/etcd/blob/main/scripts/test_lib.sh#L252, and usages, there is no actual pass of `-failfast` flag.

_Usages locations:_
https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L163
https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L168

### DIFF
--- a/scripts/test_lib.sh
+++ b/scripts/test_lib.sh
@@ -293,7 +293,7 @@ function go_test {
   junit_filename_prefix=$(junitFilenamePrefix)
 
   if [ "${VERBOSE:-}" == "1" ]; then
-    goTestFlags="-v"
+    goTestFlags="-v "
   fi
 
   # Expanding patterns (like ./...) into list of packages
@@ -306,6 +306,10 @@ function go_test {
       log_error "Cannot resolve packages: ${packages}"
       return 255
     fi
+  fi
+
+  if [ "${mode}" == "fail_fast" ]; then
+    goTestFlags+="-failfast "
   fi
 
   local failures=""


serathius (approved):

scripts/test_lib.sh
```diff
@@ -308,6 +308,10 @@ function go_test {
     fi
   fi
 
+  if [ "${mode}" == "fail_fast" ]; then
+    goTestFlags+="-failfast "
```
jmhbnz, line 312: I think this change introduces a subtle bug, but it is an easy fix.

On line 296 there is currently a final mention of `goTestFlags="-v"` which does not add any trailing space padding.  So if flags are both `verbose` and `fail_fast` I believe the resulting string will look like `-v-failfast` without any spaces inbetween.

Can you please double check this and adjust spacing to resolve if my theory is right?

jmhbnz (approved):

ptabor (approved):
I think that the role of this flag was to drive the behavior across many packages rather than pass it down to a specific package to not continue executing other tests in the package.

If I understand the tradeoff with this change is mostly about our pre-merge test: 
 - tests (if failing) will fail faster
 - the results will be less complete (as they will contain only a single failed test, rather than all of them in a single package).

As we expect developer to rerun the whole target locally after fixing a test it sounds like a fair tradeoff.

Rajalakshmi-Girish:
Will this PR need an LGTM comment to merge? @jmhbnz Can you please help me understand what it needs to merge this one.

Rajalakshmi-Girish:
> Will this PR need an LGTM comment to merge? @jmhbnz Can you please help me understand what it needs to merge this one.

or may be should all checks be successful?

serathius:
Two maintainer approvals are enough.

serathius:
Let's wait for test to pass.

Rajalakshmi-Girish:
> Let's wait for test to pass.
Will it auto merge if all tests pass and there are two approvals?

Issue #15763, opened 2023-04-24
Recent flakes in Robustness Nightly


serathius:
### Which github workflows are flaking?

https://github.com/etcd-io/etcd/actions/workflows/robustness-nightly.yaml

Would be great if someone can take a look in last weeks flakes.

### Which tests are flaking?

N/A

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/4772033408

serathius:
Looks like flakes are gone. Awesome job @fuweid !
![image](https://user-images.githubusercontent.com/5873433/236449706-20f6ed80-e6f7-426c-9e76-0653c749c578.png)

PR #15792, opened 2023-04-28
tests/robustness: tune timeout policy

fuweid:
In a [scheduled test][1], the error shows

```
2023-04-19T11:16:15.8166316Z     traffic.go:96: rpc error: code = Unavailable desc = keepalive ping failed to receive ACK within timeout
```

According to [grpc-keepalive@v1.51.0][2], each frame from server will fresh the `lastRead` and it won't file `Ping` frame to server. But the client used by [`tombstone` request][3] might hit the race. Since we use 5ms as timeout, the client might not receive the result of `Ping` from server in time. The keepalive will mark it timeout and close the connection.

I didn't reproduce it in my local. If we add the sleep before update `lastRead`, it can reproduce it sometimes. Still investigating this part.

```diff
diff --git a/internal/transport/http2_client.go b/internal/transport/http2_client.go
index d518b07e..bee9c00a 100644
--- a/internal/transport/http2_client.go
+++ b/internal/transport/http2_client.go
@@ -1560,6 +1560,7 @@ func (t *http2Client) reader(errCh chan<- error) {
                t.controlBuf.throttle()
                frame, err := t.framer.fr.ReadFrame()
                if t.keepaliveEnabled {
+                       time.Sleep(2 * time.Millisecond)
                        atomic.StoreInt64(&t.lastRead, time.Now().UnixNano())
                }
                if err != nil {
```

`DialKeepAliveTime` is always >= [10s][4]. I think we should increase the timeout to avoid flaky caused by unstable env.

And in a [scheduled test][5], the error shows

```
logger.go:130: 2023-04-22T10:45:52.646Z	INFO	Failed to trigger failpoint	{"failpoint": "blackhole", "error": "context deadline exceeded"}
```

Before sending `Status` to member, the client doesn't [pick][6] the connection in time (100ms) and returns the error.

The `waitTillSnapshot` is used to ensure that it is good enough to trigger snapshot transfer. And we have 1min timeout for injectFailpoints, so I think we can remove the 100ms timeout to reduce unnecessary stop.

```
injectFailpoints(1min timeout)
  failpoint.Inject
    triggerBlockhole.Trigger
      blackhole
        waitTillSnapshot
```

> NOTE: I didn't reproduce it either. :(

Reference:

[1]: <https://github.com/etcd-io/etcd/actions/runs/4741737098/jobs/8419176899>
[2]: <https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/internal/transport/http2_client.go#L1647>
[3]: <https://github.com/etcd-io/etcd/blob/7450cd886dd34458e28033588859c662822cb010/tests/robustness/traffic.go#L94>
[4]: <https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/dialoptions.go#L445>
[5]: <https://github.com/etcd-io/etcd/actions/runs/4772033408/jobs/8484334015>
[6]: <https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/clientconn.go#L932>

REF: #15763

### DIFF
--- a/tests/robustness/client.go
+++ b/tests/robustness/client.go
@@ -36,8 +36,8 @@ func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time) (*
 	cc, err := clientv3.New(clientv3.Config{
 		Endpoints:            endpoints,
 		Logger:               zap.NewNop(),
-		DialKeepAliveTime:    1 * time.Millisecond,
-		DialKeepAliveTimeout: 5 * time.Millisecond,
+		DialKeepAliveTime:    10 * time.Second,
+		DialKeepAliveTimeout: 100 * time.Millisecond,
 	})
 	if err != nil {
 		return nil, err
--- a/tests/robustness/failpoints.go
+++ b/tests/robustness/failpoints.go
@@ -132,8 +132,8 @@ func verifyClusterHealth(ctx context.Context, t *testing.T, clus *e2e.EtcdProces
 		clusterClient, err := clientv3.New(clientv3.Config{
 			Endpoints:            clus.Procs[i].EndpointsGRPC(),
 			Logger:               zap.NewNop(),
-			DialKeepAliveTime:    1 * time.Millisecond,
-			DialKeepAliveTimeout: 5 * time.Millisecond,
+			DialKeepAliveTime:    10 * time.Second,
+			DialKeepAliveTimeout: 100 * time.Millisecond,
 		})
 		if err != nil {
 			return fmt.Errorf("Error creating client for cluster %s: %v", clus.Procs[i].Config().Name, err)
@@ -298,8 +298,8 @@ func (t triggerDefrag) Trigger(_ *testing.T, ctx context.Context, member e2e.Etc
 	cc, err := clientv3.New(clientv3.Config{
 		Endpoints:            member.EndpointsGRPC(),
 		Logger:               zap.NewNop(),
-		DialKeepAliveTime:    1 * time.Millisecond,
-		DialKeepAliveTimeout: 5 * time.Millisecond,
+		DialKeepAliveTime:    10 * time.Second,
+		DialKeepAliveTimeout: 100 * time.Millisecond,
 	})
 	if err != nil {
 		return fmt.Errorf("failed creating client: %w", err)
@@ -322,8 +322,8 @@ func (t triggerCompact) Trigger(_ *testing.T, ctx context.Context, member e2e.Et
 	cc, err := clientv3.New(clientv3.Config{
 		Endpoints:            member.EndpointsGRPC(),
 		Logger:               zap.NewNop(),
-		DialKeepAliveTime:    1 * time.Millisecond,
-		DialKeepAliveTimeout: 5 * time.Millisecond,
+		DialKeepAliveTime:    10 * time.Second,
+		DialKeepAliveTimeout: 100 * time.Millisecond,
 	})
 	if err != nil {
 		return fmt.Errorf("failed creating client: %w", err)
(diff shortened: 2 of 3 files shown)


tests/robustness/client.go
```diff
@@ -36,8 +36,8 @@ func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time) (*
 	cc, err := clientv3.New(clientv3.Config{
 		Endpoints:            endpoints,
 		Logger:               zap.NewNop(),
-		DialKeepAliveTime:    1 * time.Millisecond,
-		DialKeepAliveTimeout: 5 * time.Millisecond,
+		DialKeepAliveTime:    10 * time.Second,
+		DialKeepAliveTimeout: 5 * time.Second,
```
serathius, line 40: The reason of having so long keep alive timeouts was to make sure that client is quick to switch server after failpoint is injected. Traffic generation during robustness tests duration can be as low as 2s. During the test we inject failpoint that usually kills one of the servers. Clients connected to this server will need to quickly react so notice issue before 2s time out.

I think the current keep alive time is too low, but I would like it to be below 0.5s. What do you think?

fuweid: Sounds good. Will update it to 100ms.

fuweid:
Run 5 rounds by myself. https://github.com/fuweid/etcd/actions/workflows/robustness.yaml?query=branch%3Alocal-deflake-robustness-cases

One fail https://github.com/fuweid/etcd/actions/runs/4829225472/jobs/8604025015. It looks like the flaky is caused by other.

tests/robustness/client.go
```diff
@@ -36,8 +36,8 @@ func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time) (*
 	cc, err := clientv3.New(clientv3.Config{
 		Endpoints:            endpoints,
 		Logger:               zap.NewNop(),
-		DialKeepAliveTime:    1 * time.Millisecond,
-		DialKeepAliveTimeout: 5 * time.Millisecond,
+		DialKeepAliveTime:    10 * time.Second,
```
serathius, line 39:
```
	// DialKeepAliveTime is the time after which client pings the server to see if
	// transport is alive.
```
This is pretty big value, as it's greater then 2 seconds, isn't it just saying don't do any keep alives?

fuweid: The gRPC client will update it to 10s if it is smaller than 10s. 

https://github.com/grpc/grpc-go/blob/eeb9afa1f6b6388152955eeca8926e36ca94c768/dialoptions.go#L445

fuweid:
> === FAIL: integration/clientv3/naming TestEtcdGrpcResolverRoundRobin (0.28s)

It looks like it is new flaky case.

serathius:
> Run 5 rounds by myself. https://github.com/fuweid/etcd/actions/workflows/robustness.yaml?query=branch%3Alocal-deflake-robustness-cases

Failure of 1 per 5 already looks good. Thanks for confirming it!

tests/robustness/failpoints.go
```diff
@@ -442,21 +442,17 @@ func waitTillSnapshot(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCl
 		endpoints = append(endpoints, ep)
 	}
 	clusterClient, err := clientv3.New(clientv3.Config{
-		Endpoints:            endpoints,
-		Logger:               zap.NewNop(),
-		DialKeepAliveTime:    1 * time.Millisecond,
-		DialKeepAliveTimeout: 5 * time.Millisecond,
+		Endpoints: endpoints,
```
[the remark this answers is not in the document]
fuweid: Ok. Let me update it with 10s and 100ms

tests/robustness/client.go
serathius: ok, so maybe we can just skip it as it's not really useful config

fuweid: The etcd client pkg only uses keepalive grpc opt when the time is large than 0. So we still need this setting.

serathius: Oh, thanks for pointing this out.

serathius (approved):

PR #15813, opened 2023-05-02
keep going with other test suite runs when failure with one

Rajalakshmi-Girish:
The current logic in [test.sh](https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L644:#L659) exits the run when there is first failure in the suite names passed to PASSES variable.

This is leading to incomplete results when the user intends to run multiple test suites.
This change continues with the for loop when there is failure in any suite and exits after the completion of all test suite runs.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -55,6 +55,7 @@ if [ -n "${OUTPUT_FILE}" ]; then
 fi
 
 PASSES=${PASSES:-"gofmt bom dep build unit"}
+PASSES_CONTINUE=${PASSES_CONTINUE:-false}
 PKG=${PKG:-}
 SHELLCHECK_VERSION=${SHELLCHECK_VERSION:-"v0.8.0"}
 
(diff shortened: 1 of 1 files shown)


Rajalakshmi-Girish:
@serathius @jmhbnz Can you please take a look at this PR?

jmhbnz (commented):
Is continuing with other test suites actually what we want though?  

I think it might depend on the situation as when I am developing locally and I run `make test` which runs a bunch of suites if there is a failure with unit tests my preference is to have it stop immediately rather than continue on trying to run other tests which will waste time.

If I'm misunderstanding something about this change please let me know, also curious what other folks think here.

scripts/test.sh
```diff
@@ -646,16 +646,26 @@ function run_pass {
...
   log_callout -e "\\n'${pass}' started at $(date)"
   if "${pass}_pass" "$@" ; then
-    log_success "'${pass}' completed at $(date)"
+    log_success "'${pass}' PASSED and completed at $(date)"
+    return 0
   else
-    log_error "FAIL: '${pass}' failed at $(date)"
-    exit 255
+    log_error "FAIL: '${pass}' FAILED and completed at $(date)"
+    return 2
```
chaochn47, line 653: Is it possible to make this gated by environment variables so you can turn it on in your environment?

jmhbnz: Great suggestion, this approach sounds good to me.

Rajalakshmi-Girish: I have added the gate by the variable `PASSES_CONTINUE`.
Can you please check now?

jmhbnz (approved):
LGTM

chaochn47 (approved):

ahrtr (approved):
LGTM

Thanks @Rajalakshmi-Girish

Issue #15598, opened 2023-03-31
Use raft log  to patch operation history in Robustness tests


serathius:
### What would you like to be added?

Robustness tests execute traffic against a cluster and record their history so they can be validated for linearizability. Unfortunately linearizing history is an NP-Hard problem, meaning it has exponential complexity by default. We maneuver around the problem by special algorithms (https://github.com/anishathalye/porcupine#porcupine-- uses https://arxiv.org/pdf/1504.00204.pdf) and history patching. 

History patching is needed to remove operations that cause linearization to require exponential complexity. Those operations are client requests that returned error for any reason, connection timeout, request timeout etc. In those cases we don't know if the request was persisted or not, and if it was persisted what revision it got. Without knowing this we need to consider each possibility and any order of requests. For sequence of requests of length n, we already get 2^n states (persisted or not) that is also multiplied by n! (order of operations). 

History patching uses independent history of events collected via watches to either remove failed requests from the operation history or provides them with their revision. This can be done for PUT and TXN as they are unique because robustness tests write unique value in each request.

## Problem
Watch events only include transactions that succeeded. Delete request for a non-present key will not be present in watch history. Same for TXN transaction that failed it's condition and had empty onFail operations. Those operations where still executed by etcd and their condition checked, however current patching marks them as never existing. 

Using watch to patch operation history means that etcd could still have an linearization issue on condition checking and we would never know about this. We need a way to not only collect operations that succeeded by operations that were `applied` to backend.

## Proposed solution
* Validate best way to access raft log in cluster, either by WAL files or connecting to raft as a learner
* Use raft log to decide whether operation was executed or not.
* Still use watch history, but only to provide revision for operations peristed.

### Why is this needed?

Patching operation history is critical to make robustness tests performant, however in it's current state limits our ability to validate correctness. We should be aware of this flaw and when needed we should fix it.

Not a priority for now, but an important thing to have in mind.

serathius:
I should consider prioritizing this to implement Compaction for Kubernetes. With more aggressive compaction watches will be interrupted, which would lead to gaps in our event history. With gaps we could not patch operation history, which could lead to linearization timeouts.

serathius:
> Overall it seems not a good idea to verify etcd's correctness depending on WAL data.

[...]

PR #15819, opened 2023-05-03
tests/robustness: Split model code into deterministic and non-deterministic

serathius:
Big refactor to separate deterministic etcd model (assumes all request succeed) and non-deterministic (considers failed request both persisted and not). This way a non-deterministic model just layer above deterministic one. 

cc @ahrtr @ptabor


serathius:

serathius:
PTAL @ahrtr 
This refactor makes code much clearer, however it's nightmare to rebase against.

tests/robustness/model/model.go
```diff
@@ -33,469 +30,92 @@ const (
...
-	}
-	return state
+func initNonDeterministicState(request EtcdRequest, response EtcdNonDeterministicResponse) nonDeterministicState {
+	return nonDeterministicState{initState(request, response.EtcdResponse)}
 }
 
-// applyFailedRequest handles a failed requests, one that it's not known if it was persisted or not.
-func applyFailedRequest(states PossibleStates, request EtcdRequest) PossibleStates {
+// stepFailedRequest duplicates number of states by considering request persisted and lost.
+func (states nonDeterministicState) stepFailedRequest(request EtcdRequest) nonDeterministicState {
```
ahrtr, line 95: You want change the receiver's state, but it isn't a pointer receiver. So a function makes more sense than a method here. This comment applies the `stepSuccessfulRequest` as well.

ahrtr:
This's a huge PR, and it's hard to carefully review each line of code change. Please try to breakdown such huge PR as possible as you can next time.

I will approve this PR for now given:
1. It's just test code;
2. You have deep understanding on the robustness test;
3. This PR will definitely conflict with any other PRs which update robustness test, and you or others will have to do the boring rebasing.

ahrtr:
Before I approve this PR, please other members also take a look. cc @chaochn47 @fuweid @jmhbnz @ptabor if possible.

tests/robustness/model/model.go
```diff
@@ -33,469 +30,92 @@ const (
...
+	newStates := make(nonDeterministicState, 0, len(states)*2)
```
lavacat, line 96: you are doubling number of states, is there a scenario that the number will grow very quickly and checker won't be able to handle it?

tests/robustness/model/model.go
```diff
@@ -33,469 +30,92 @@ const (
...
 	for _, s := range states {
-		newState, _ := applyRequestToSingleState(s, request)
+		newStates = append(newStates, s)
+		newState, _ := s.step(request)
 		if !reflect.DeepEqual(newState, s) {
-			states = append(states, newState)
```
lavacat, line 364: nit: I see that the check was added in different PR, but can you add a comment why we need `!reflect.DeepEqual(newState, s)`

tests/robustness/model/history.go
```diff
@@ -295,11 +295,15 @@ func (h *AppendableHistory) AppendDefragment(start, end time.Duration, resp *cli
 		h.appendFailed(request, start, err)
 		return
 	}
+	var revision int64
```
lavacat, line 298: adding revision to defragmentResponse is nice to have, not critical for this PR. I that correct? Just trying to understand

tests/robustness/model/model.go
chaochn47: I guess it is just used to avoid same states. Please confirm.

tests/robustness/model/model.go
chaochn47: I guess `stepSuccessfulRequest` would remove possible states. One successful response should clean up all previous possible states generated back to back.

chaochn47:
LGTM, It's a clean refactor from my point of view. The changed lines look scary though. No regression spotted from my eyes.

tests/robustness/model/base.go
```diff
@@ -0,0 +1,341 @@
...
+	"hash/fnv"
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/anishathalye/porcupine"
+)
+
+// Base model implements a deterministic etcd model that assumes that all request succeed.
+var Base = porcupine.Model{
```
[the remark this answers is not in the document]
serathius: Sounds good.

tests/robustness/model/base.go
```diff
@@ -0,0 +1,341 @@
...
+	return state
+}
+
+// applyRequestToSingleState handles a successful request, returning updated state and response it would generate.
+func (s etcdState) step(request EtcdRequest) (etcdState, EtcdResponse) {
+	newKVs := map[string]ValueRevision{}
+	for k, v := range s.KeyValues {
+		newKVs[k] = v
+	}
+	s.KeyValues = newKVs
```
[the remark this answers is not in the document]
serathius: Avoid modifying the `s.KeyValues`, I need to copy whole map. Would love to get snapshotable map to clean this up.

tests/robustness/model/base.go
```diff
@@ -0,0 +1,341 @@
...
+		lease := EtcdLease{
+			LeaseID: request.LeaseGrant.LeaseID,
+			Keys:    map[string]struct{}{},
+		}
+		s.Leases[request.LeaseGrant.LeaseID] = lease
+		return s, EtcdResponse{Revision: s.Revision, LeaseGrant: &LeaseGrantReponse{}}
+	case LeaseRevoke:
+		//Delete the keys attached to the lease
+		keyDeleted := false
+		for key := range s.Leases[request.LeaseRevoke.LeaseID].Keys {
```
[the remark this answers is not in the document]
serathius: Yes for now we don't test lease expiration.

tests/robustness/model/base.go
```diff
@@ -0,0 +1,341 @@
...
+type EtcdResponse struct {
+	Revision    int64
+	Txn         *TxnResponse
+	LeaseGrant  *LeaseGrantReponse
+	LeaseRevoke *LeaseRevokeResponse
+	Defragment  *DefragmentResponse
+}
+
+type TxnResponse struct {
+	TxnResult bool
```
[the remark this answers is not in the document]
serathius: Will fix it, sorry for this abomination in naming.

tests/robustness/model/history.go
serathius: It is based on other code changes due to some shortcuts I took. By separating models I separate `EtcdResponse` and `EtcdNonDeterministicResponse` which means that now `EtcdResponse` needs to have accurate `Revision`, which means that I need to make sure that defragmentResponse returns response with accurate revision.

tests/robustness/model/model.go
```diff
@@ -33,469 +30,92 @@ const (
 	Delete OperationType = "delete"
 )
 
-var Etcd = porcupine.Model{
+// NonDeterministic model implements etcd model that handles responses with error.
+// Error response don't inform whether request was persisted or not, so model needs to
+// consider both cases. This is done by state being represented as multiple equally
+// possible deterministic states. Failed requests fork the possible states, while
+// successful requests filter them.
+var NonDeterministic = porcupine.Model{
```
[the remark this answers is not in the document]
serathius: Done

tests/robustness/model/model.go
serathius: I think it's useful for clarity to have a both models implement `Step` method. It makes it more natural to understand when compared to have `stepDeterministicModel` and `stepNonDeterministicModel`.

Fact that I don't use pointer receiver and return new state is just a design decision that state is non mutable. I think it's ok that methods to be non mutable.

tests/robustness/model/model.go
serathius: You are both right. NonDeterministic model can grow exponentially with number of operations. This however mostly happens on failed/error (wanna guess when it can happen on request success?). As @chaochn47 mentioned successful requests will cleanup previous possible states. 

So the problem is that we should avoid long sequences of failed requests as they cause complexity to explode. I have mitigated the problem by introducing history patching https://github.com/etcd-io/etcd/pull/15078 https://github.com/etcd-io/etcd/issues/15598

tests/robustness/model/model.go
serathius: If the request doesn't impact the state then there is no reason to fork it. Example is failed defrag request.

serathius:
Sorry for large PR, I didn't have time to split it and was worried that it would only make review take longer. As code structure clarifies it should be easier to make small independent changes.

FYI I'm pretty sure that this change is correct as etcd model have absurd number of tests and any error would result would immediately result in robustness tests failing. 

Main motivation for this was exponential growth of tests for non-deterministic model. If you look through scenarios I test every combination of requests both their success and failure. Separating deterministic model tests allows me to just implement tests for request success.

ahrtr:
@serathius Please ping me if you resolve all comments, I will take another look so that this PR can merged today.

serathius:
I resolved all comments, however I didn't mark all of them as "done", as some of them included questions and I want to give people chance to read answers.

Issue #13896, opened 2022-04-06
Improve contributor experience with CI


serathius:
* In CONTRIBUTING.md, list scripts in scripts/ that a first-time contributor would need (and what they're for). Or - have a short explanation at the top of each script which briefly says what it's for. (Doesn't have to be a full usage guide, just "Updates license information in modules and submodules. CI checks that running this causes no changes." would go a long way)
* For CI failures, suggest the appropriate script to fix the corresponding failure. For example, if the bom test fails, most likely the contributor just needs to run ./scripts/fix.sh

ptabor:
`make test-smoke` and `make test-full` are short cut's for O(1min) feedback and full feedback comparable with github presubmits respectively.

vimalk78:
@willbeason the PR https://github.com/etcd-io/etcd/pull/13956 should fix the issue you are facing

serathius:
Not all issues listed were addressed

serathius:
With introduction of `make test-*` `make verify-*` and `make fix-*` I would propose to split `./scripts/test.sh` to separate targets and document them all in makefile

sharathsivakumar:
Hey @serathius @jmhbnz Can you please assign me to this? I am keen to work on it.

With the example of `make test-e2e`, this currently calls `scripts/test.sh` as follows. 
`PASSES="e2e" ./scripts/test.sh $(GO_TEST_FLAGS)`

Are we proposing to split this individual target to a separate file called `scripts/test-e2e.sh`

serathius:
Sounds good @sharathsivakumar, thanks for looking into this!

sharathsivakumar:
@serathius and @jmhbnz I have been  away due to personal issues and could not complete this. However I am back now and I will get this done soon. I am looking to turn in the first PR by early next week. Hope that's fine.

serathius:
Not a problem, great to hear you are still interested. Awaiting your PR.

PR #15824, opened 2023-05-04
Add an option to keep_going with run for modules on failure

Rajalakshmi-Girish:
The existing logic in the function `run_for_modules` exits the loop through modules when there is a failure.

This change adds an environment variable ` KEEP_GOING_MODULE` to have the option to continue running tests in other modules even on failure with one of them.


serathius:
FYI In https://github.com/etcd-io/etcd/issues/13896#issuecomment-1502846512 we proposed to split/remove `./scripts/test.sh` which is somewhat separate direction from this PR that tries to fix the script.

jmhbnz (approved):
As mentioned above we have a plan to restructure how the test scripts work in future. Please feel welcome to continue with that after this merges if you are interested.

Putting the future changes aside, in the interim this change itself looks OK to me, thanks @Rajalakshmi-Girish.

ahrtr:
Previously you added `PASSES_CONTINUE` in https://github.com/etcd-io/etcd/pull/15813; now you add `KEEP_GOING_MODULE` in this PR. Can we have consistent naming?
- `KEEP_GOING_TESTSUITE` or `KEEP_GOING_SUITE`
- `KEEP_GOING_MODULE`

Rajalakshmi-Girish:
Messed up while rebasing and resolving conflicts :(

Issue #15820, opened 2023-05-03
Codify the the kubernetes-etcd contract into etcd robustness tests


serathius:
### What would you like to be added?

Etcd robustness tests are pretty great way to validate generic contracts, however it's pretty huge problem space and it's hard to know what should be done next.

To drive robustness testing direction and priorities I would like to propose to first to focus on Kubernetes-Etcd contract. Work would include clarifying, documenting and then codifying it into etcd robustness tests. 

The end result would be an etcd model that work as a fake that Kubernetes project can use to cross test the contract. A fake is a lightweight implementation that behaves like a real system and can be used for testing.

## Plan
* [x] [Collect the collective knowledge of the implicit Kubernetes-etcd contract in a document and collect feedback](https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit?usp=sharing) requires joining etcd-dev@googlegroups.com or dev@kubernetes.io to comment.
* [x] https://github.com/etcd-io/website/pull/676
* [x] Validate K8s-etcd contract in etcd robustness tests
  * [x] https://github.com/etcd-io/etcd/pull/15817
  * [x] https://github.com/etcd-io/etcd/pull/15827
  * [x] https://github.com/etcd-io/etcd/pull/15840
  * [x] https://github.com/etcd-io/etcd/pull/16072
  * [x] Implement list-watch protocol
  * [x] Implement pagination https://github.com/etcd-io/etcd/pull/16111
  * [x] Update watch validation to match contract watch validation https://github.com/etcd-io/etcd/pull/16136
  * [x] PrevKey https://github.com/etcd-io/etcd/issues/16838
  * [x] Use WithRequireLeader
  * [x]  https://github.com/etcd-io/etcd/pull/17833
  * [ ] Implement non-recursive watch
  * [x] Implement Kubernetes like compaction

### Why is this needed?

Kubernetes is currently the main user of etcd, still there are a number of untested assumptions in the K8s-etcd contract. Changes to these underlying features may cause Kubernetes to behave incorrectly, or even not work at all.

chaochn47:
> RemainingItemCount: Allow the API servers to show a count of remaining items in the response to a [chunking list request](https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks)

[...]

PR #15827, opened 2023-05-04
tests/robustness: Implement delete request for kubernetes scenario

serathius:

cc @ahrtr @ptabor


tests/robustness/traffic.go
```diff
@@ -147,27 +147,47 @@ func (t kubernetesTraffic) Run(ctx context.Context, clientId int, c *recordingCl
...
+		} else {
+			switch t.writes.PickRandom() {
+			case Delete:
+				err = t.Delete(writeCtx, c, string(randomPod.Key), randomPod.ModRevision)
+			case Update:
+				err = t.Update(writeCtx, c, string(randomPod.Key), fmt.Sprintf("%d", ids.RequestId()), randomPod.ModRevision)
+			case Create:
+				err = t.Create(writeCtx, c, t.generateKey(), fmt.Sprintf("%d", ids.RequestId()))
+			default:
+				panic("invalid operation")
```
ahrtr, line 180: nit
```suggestion
				panic(fmt.Sprintf("invalid operation: %q", t.writes.PickRandom())
```

tests/robustness/traffic.go
```diff
@@ -53,6 +53,8 @@ const (
 	LeaseRevoke   TrafficRequestType = "leaseRevoke"
 	CompareAndSet TrafficRequestType = "compareAndSet"
 	Defragment    TrafficRequestType = "defragment"
+	Update        TrafficRequestType = "update"
+	Create        TrafficRequestType = "create"
```
ahrtr, line 57: IT seems that the relationship between all the `TrafficRequestType` (Operation type?) is overlapped, and accordingly a little confusing. For example, 
- The Delete is reused by `traffic` and `kubernetesTraffic`, but has different meaning in different context;
- The `Update`, `Create` and `Delete` (used in KuberneteTraffic) are actualy TXN; Have you consider reusing `MultiOpTxn`? 
- `CompareAndSet` is also a TXN.

ahrtr (approved):
Overall looks good to me with minor comments. 

Thanks.

tests/robustness/traffic.go
serathius: Done

tests/robustness/traffic.go
serathius: Yea, `TrafficRequestType` is more of an identificator for `requestRandomizer` and can have different meaning for each traffic. It would be nicer if each traffic generator had specific type alias and have requestRandomizer work on generic type.

serathius:
I have implemented a weighted choice for picking operations, for now it requires string type as label, however with Go generics we can make it work on any type and make it more readable.

PR #15505, opened 2023-03-18
scripts: Add testing of etcd in local image in release workflow.

pchan:
Please see the previous PR #15350 for details.
Fixes: https://github.com/etcd-io/etcd/issues/15139

@ahrtr @chaochn47

PR #15833, opened 2023-05-05
Test Arm64 etcd container in K8s cluster

kevinzs2048:
Since the multiple architecture image issue #15505 has been sorted out, continue to introduce the test for Arm64 for local image test as it does on #15139

Again, on PR trigger first to see the test result and will change back to periodically later.


serathius (changes requested):
Release workflow goal is not to test images, but to test the release process. There is no benefit of running them on arm.

kevinzs2048:
Aha, OK, thanks for this info, I will close this PR.

PR #15835, opened 2023-05-06
grpcproxy: fix memberlist results not update when proxy node down

yellowzf:
If start grpc proxy with --resolver-prefix, memberlist will return all alive proxy nodes, when one grpc proxy node is down, it is expected to not return the down node, but it is still return

### DIFF
--- a/server/proxy/grpcproxy/cluster.go
+++ b/server/proxy/grpcproxy/cluster.go
@@ -113,9 +113,9 @@ func (cp *clusterProxy) monitor(wa endpoints.WatchChannel) {
 			for _, up := range updates {
 				switch up.Op {
 				case endpoints.Add:
-					cp.umap[up.Endpoint.Addr] = up.Endpoint
+					cp.umap[up.Key] = up.Endpoint
 				case endpoints.Delete:
-					delete(cp.umap, up.Endpoint.Addr)
+					delete(cp.umap, up.Key)
 				}
 			}
 			cp.umu.Unlock()
@@ -139,12 +139,12 @@ func (cp *clusterProxy) membersFromUpdates() ([]*pb.Member, error) {
 	cp.umu.RLock()
 	defer cp.umu.RUnlock()
 	mbs := make([]*pb.Member, 0, len(cp.umap))
-	for addr, upt := range cp.umap {
+	for _, upt := range cp.umap {
 		m, err := decodeMeta(fmt.Sprint(upt.Metadata))
 		if err != nil {
 			return nil, err
 		}
-		mbs = append(mbs, &pb.Member{Name: m.Name, ClientURLs: []string{addr}})
+		mbs = append(mbs, &pb.Member{Name: m.Name, ClientURLs: []string{upt.Addr}})
 	}
 	return mbs, nil
 }
--- a/tests/integration/proxy/grpcproxy/cluster_test.go
+++ b/tests/integration/proxy/grpcproxy/cluster_test.go
@@ -17,15 +17,18 @@ package grpcproxy
 import (
 	"context"
 	"net"
+	"os"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 	"google.golang.org/grpc"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/client/v3/naming/endpoints"
 	"go.etcd.io/etcd/server/v3/proxy/grpcproxy"
 	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
 )
@@ -36,7 +39,11 @@ func TestClusterProxyMemberList(t *testing.T) {
 	clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 1})
 	defer clus.Terminate(t)
 
-	cts := newClusterProxyServer(zaptest.NewLogger(t), []string{clus.Members[0].GRPCURL()}, t)
+	lg := zaptest.NewLogger(t)
+	serverEps := []string{clus.Members[0].GRPCURL()}
+	prefix := "test-prefix"
+	hostname, _ := os.Hostname()
+	cts := newClusterProxyServer(lg, serverEps, prefix, t)
 	defer cts.close(t)
 
 	cfg := clientv3.Config{
@@ -50,7 +57,7 @@ func TestClusterProxyMemberList(t *testing.T) {
 	defer client.Close()
 
 	// wait some time for register-loop to write keys
-	time.Sleep(time.Second)
+	time.Sleep(200 * time.Millisecond)
 
 	var mresp *clientv3.MemberListResponse
 	mresp, err = client.Cluster.MemberList(context.Background())
@@ -64,9 +71,38 @@ func TestClusterProxyMemberList(t *testing.T) {
 	if len(mresp.Members[0].ClientURLs) != 1 {
 		t.Fatalf("len(mresp.Members[0].ClientURLs) expected 1, got %d (%+v)", len(mresp.Members[0].ClientURLs), mresp.Members[0].ClientURLs[0])
 	}
-	if mresp.Members[0].ClientURLs[0] != cts.caddr {
-		t.Fatalf("mresp.Members[0].ClientURLs[0] expected %q, got %q", cts.caddr, mresp.Members[0].ClientURLs[0])
+	assert.Contains(t, mresp.Members, &pb.Member{Name: hostname, ClientURLs: []string{cts.caddr}})
+
+	//test proxy member add
+	newMemberAddr := "127.0.0.2:6789"
+	grpcproxy.Register(lg, cts.c, prefix, newMemberAddr, 7)
+	// wait some time for proxy update members
+	time.Sleep(200 * time.Millisecond)
+
+	//check add member succ
+	mresp, err = client.Cluster.MemberList(context.Background())
+	if err != nil {
+		t.Fatalf("err %v, want nil", err)
 	}
+	if len(mresp.Members) != 2 {
+		t.Fatalf("len(mresp.Members) expected 2, got %d (%+v)", len(mresp.Members), mresp.Members)
+	}
+	assert.Contains(t, mresp.Members, &pb.Member{Name: hostname, ClientURLs: []string{newMemberAddr}})
+
+	//test proxy member delete
+	deregisterMember(cts.c, prefix, newMemberAddr, t)
+	// wait some time for proxy update members
+	time.Sleep(200 * time.Millisecond)
+
+	//check delete member succ
+	mresp, err = client.Cluster.MemberList(context.Background())
+	if err != nil {
+		t.Fatalf("err %v, want nil", err)
+	}
+	if len(mresp.Members) != 1 {
+		t.Fatalf("len(mresp.Members) expected 1, got %d (%+v)", len(mresp.Members), mresp.Members)
+	}
+	assert.Contains(t, mresp.Members, &pb.Member{Name: hostname, ClientURLs: []string{cts.caddr}})
 }
 
 type clusterproxyTestServer struct {
@@ -90,7 +126,7 @@ func (cts *clusterproxyTestServer) close(t *testing.T) {
 	}
 }
 
-func newClusterProxyServer(lg *zap.Logger, endpoints []string, t *testing.T) *clusterproxyTestServer {
+func newClusterProxyServer(lg *zap.Logger, endpoints []string, prefix string, t *testing.T) *clusterproxyTestServer {
 	cfg := clientv3.Config{
 		Endpoints:   endpoints,
 		DialTimeout: 5 * time.Second,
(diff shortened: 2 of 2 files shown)


server/proxy/grpcproxy/cluster.go
```diff
@@ -113,9 +113,9 @@ func (cp *clusterProxy) monitor(wa endpoints.WatchChannel) {
 			for _, up := range updates {
 				switch up.Op {
 				case endpoints.Add:
-					cp.umap[up.Endpoint.Addr] = up.Endpoint
+					cp.umap[up.Key] = up.Endpoint
 				case endpoints.Delete:
-					delete(cp.umap, up.Endpoint.Addr)
```
yellowzf, line 118: when op delete, up.Endpoint.Addr is not specified, can not delete the endpoint from map, use up.Key instead

lavacat: looks good, I think using Key is ok because it's already used in resolver.watch https://github.com/etcd-io/etcd/blob/249c0d71d4e106e0d2f5576b9638f2e8bd75ca47/client/v3/naming/resolver/resolver.go#LL97C6-L97C12
Not sure why Endpoint.Addr was used initially

lavacat (changes requested):
LGTM, but needs a test. 
`case endpoints.Add:` is covered in TestClusterProxyMemberList. But `case endpoints.Delete` isn't [covered](https://app.codecov.io/gh/etcd-io/etcd/blob/main/server/proxy/grpcproxy/cluster.go#L117) at all.

yellowzf:
> LGTM, but needs a test. `case endpoints.Add:` is covered in TestClusterProxyMemberList. But `case endpoints.Delete` isn't [covered](https://app.codecov.io/gh/etcd-io/etcd/blob/main/server/proxy/grpcproxy/cluster.go#L117) at all.

yellowzf:
> LGTM, but needs a test. `case endpoints.Add:` is covered in TestClusterProxyMemberList. But `case endpoints.Delete` isn't [covered](https://app.codecov.io/gh/etcd-io/etcd/blob/main/server/proxy/grpcproxy/cluster.go#L117) at all.

tests added

yellowzf:
PTAL, it's ready for review @lavacat

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -90,6 +129,27 @@ func (cts *clusterproxyTestServer) close(t *testing.T) {
...
+	client, err := integration2.NewClient(t, cfg)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	grpcproxy.Register(lg, client, "test-prefix", addr, ttl)
+	time.Sleep(time.Second)
+	donec := make(chan struct{})
+	go func() {
+		time.Sleep(delay)
```
lavacat, line 146: what is the point this goroutine if it's just waiting? I think this can be removed

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +69,43 @@ func TestClusterProxyMemberList(t *testing.T) {
...
+	}
+	if len(mresp.Members) != 2 {
+		t.Fatalf("len(mresp.Members) expected 2, got %d (%+v)", len(mresp.Members), mresp.Members)
+	}
+	if mresp.Members[1].ClientURLs[0] != newMemberAddr {
+		t.Fatalf("mresp.Members[1].ClientURLs[0] expected %q, got %q", newMemberAddr, mresp.Members[1].ClientURLs[0])
+	}
+
+	//check del member succ
+	time.Sleep(wait)
```
lavacat, line 93: `wait` > `delay`. What is the point if `delay`?

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -90,6 +129,27 @@ func (cts *clusterproxyTestServer) close(t *testing.T) {
 	}
 }
 
+func registDelayDeleteMember(lg *zap.Logger, endpoints []string, addr string, delay time.Duration, ttl int, t *testing.T) chan struct{} {
+	cfg := clientv3.Config{
+		Endpoints:   endpoints,
+		DialTimeout: 5 * time.Second,
+	}
+	client, err := integration2.NewClient(t, cfg)
```
lavacat, line 137: Is it possible to pass client from the test without creating a new one?

lavacat (changes requested):
Please cleanup the logic and add comments to any non-obvious test setup and checks. 

There is also TestRegister test that uses NewWatchChannel to listen to updates, it might be easier to update. Up to you.

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -90,6 +129,27 @@ func (cts *clusterproxyTestServer) close(t *testing.T) {
 	}
 }
 
+func registDelayDeleteMember(lg *zap.Logger, endpoints []string, addr string, delay time.Duration, ttl int, t *testing.T) chan struct{} {
```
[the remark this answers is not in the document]
yellowzf: removed this func in new code, use DeleteEndpoint is better

tests/integration/proxy/grpcproxy/cluster_test.go
yellowzf: this goroutine used to delay close the etcd client, if client is closed, lease will expire soon, then the key which lease attached will delete, it is complex, use DeleteEndpoint is more convenient

tests/integration/proxy/grpcproxy/cluster_test.go
yellowzf: use delay then test case can check proxy member has successfully updayed

tests/integration/proxy/grpcproxy/cluster_test.go
yellowzf: yes, it is

yellowzf:
> Please cleanup the logic and add comments to any non-obvious test setup and checks.
> 
> There is also TestRegister test that uses NewWatchChannel to listen to updates, it might be easier to update. Up to you.

TestRegister only cover watch channel logic, TestClusterProxyMemberList cover member list api and watch channel.
in new code, use DeleteEndpoint to delete key of member, seems more clear.
it's ready for review, @lavacat , PTAL

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +71,40 @@ func TestClusterProxyMemberList(t *testing.T) {
...
+
+	//check add member succ
+	mresp, err = client.Cluster.MemberList(context.Background())
+	if err != nil {
+		t.Fatalf("err %v, want nil", err)
+	}
+	if len(mresp.Members) != 2 {
+		t.Fatalf("len(mresp.Members) expected 2, got %d (%+v)", len(mresp.Members), mresp.Members)
+	}
+	if mresp.Members[1].ClientURLs[0] != newMemberAddr {
```
lavacat, line 89: nit: I think `Members` is populated from a [map](https://github.com/etcd-io/etcd/blob/9a922091ed07c56c0c09575c2328e9576c76d942/server/proxy/grpcproxy/cluster.go#L142), so order isn't [guarantied](https://go.dev/blog/maps). Might cause test flakiness. 

Can we add a check for Members[0] and add a comment about order?

lavacat:
I wonder if anything can be done with those `time.Sleep(time.Second)` in the test? I understand that you are reusing prior code, but if you see some other solution would be nice to update.

yellowzf:
> I wonder if anything can be done with those `time.Sleep(time.Second)` in the test? I understand that you are reusing prior code, but if you see some other solution would be nice to update.

proxy watch specific prefix for members update, when register or delete members by put/delete key, it updates members by process watch resp, it is asynchronous, so use `time.Sleep(time.Second)` to wait for update, i think prior code is the same meaning

tests/integration/proxy/grpcproxy/cluster_test.go
yellowzf: map can not guarantee order, we should check if the new member is in `mresp.Member`, codes have updated, PTAL

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +71,47 @@ func TestClusterProxyMemberList(t *testing.T) {
 	if mresp.Members[0].ClientURLs[0] != cts.caddr {
 		t.Fatalf("mresp.Members[0].ClientURLs[0] expected %q, got %q", cts.caddr, mresp.Members[0].ClientURLs[0])
 	}
+
+	//test proxy member up/down
```
lavacat, line 75: nit: `test proxy member add`

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +71,47 @@ func TestClusterProxyMemberList(t *testing.T) {
...
+	mresp, err = client.Cluster.MemberList(context.Background())
+	if err != nil {
+		t.Fatalf("err %v, want nil", err)
+	}
+	if len(mresp.Members) != 2 {
+		t.Fatalf("len(mresp.Members) expected 2, got %d (%+v)", len(mresp.Members), mresp.Members)
+	}
+
+	succ := false
+	for _, member := range mresp.Members {
```
lavacat, line 91: nit: replace with 
```
	hostname, _ := os.Hostname()
	...
	assert.Contains(t, mresp.Members, &pb.Member{Name: hostname, ClientURLs: []string{newMemberAddr}})
```

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +71,47 @@ func TestClusterProxyMemberList(t *testing.T) {
 	if mresp.Members[0].ClientURLs[0] != cts.caddr {
```
lavacat, line 71: nit: `assert.Contains(t, mresp.Members, &pb.Member{Name: hostname, ClientURLs: []string{cts.caddr}})`

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +71,47 @@ func TestClusterProxyMemberList(t *testing.T) {
...
+
+	//check delete member succ
+	mresp, err = client.Cluster.MemberList(context.Background())
+	if err != nil {
+		t.Fatalf("err %v, want nil", err)
+	}
+	if len(mresp.Members) != 1 {
+		t.Fatalf("len(mresp.Members) expected 1, got %d (%+v)", len(mresp.Members), mresp.Members)
+	}
+	if mresp.Members[0].ClientURLs[0] != cts.caddr {
```
lavacat, line 112: nit: `assert.Contains(t, mresp.Members, &pb.Member{Name: hostname, ClientURLs: []string{cts.caddr}})`

lavacat (approved):

lavacat:
LGTM, please use `assert.Contains` as per comments. cc @ahrtr for final review

tests/integration/proxy/grpcproxy/cluster_test.go
```diff
@@ -67,6 +71,47 @@ func TestClusterProxyMemberList(t *testing.T) {
 	if mresp.Members[0].ClientURLs[0] != cts.caddr {
 		t.Fatalf("mresp.Members[0].ClientURLs[0] expected %q, got %q", cts.caddr, mresp.Members[0].ClientURLs[0])
 	}
+
+	//test proxy member up/down
+	newMemberAddr := "127.0.0.2:6789"
+	grpcproxy.Register(lg, cts.c, prefix, newMemberAddr, 7)
+	// wait some time for proxy update members
+	time.Sleep(time.Second)
```
lavacat, line 79: nit: I think we can switch to `time.Sleep(200 * time.Millisecond)` for all `waits` in this test. 1s seams too long.

ahrtr:
> If start grpc proxy with --resolver-prefix, memberlist will return all alive proxy nodes, when one grpc proxy node is down, it is expected to not return the down node, but it is still return

I do not get the real issue. Please provide the detailed steps to reproduce the issue.

yellowzf:
> > If start grpc proxy with --resolver-prefix, memberlist will return all alive proxy nodes, when one grpc proxy node is down, it is expected to not return the down node, but it is still return
> 
> I do not get the real issue. Please provide the detailed steps to reproduce the issue.

step1: start a grpc-proxy cluster which contains 3 member, member1/member2/member3, use --resolver-prefix flag when member start.
step2: use etcd client connect to member1, call `MemberList`, get a response which contains member1/member2/member3.
step3: stop member3, then lease of member3 in etcd server will expire and key will be deleted.
step4:  use etcd client connect to member1, call `MemberList`, still get a response which contains member1/member2/member3, it is expect to only contains member1/member2 because member3 is stoped. @ahrtr

yellowzf:
> LGTM, please use `assert.Contains` as per comments. cc @ahrtr for final review

all codes memtioned on comments has updated, @lavacat  PTAL

ahrtr (approved):
LGTM

Thanks @yellowzf for the fix.

Could you backport this PR to 3.4 and 3.5?

yellowzf:
> LGTM
> 
> Thanks @yellowzf for the fix.
> 
> Could you backport this PR to 3.4 and 3.5?

3.4 use preKv to assign addr, no need to backport, https://github.com/etcd-io/etcd/blob/release-3.4/clientv3/naming/grpc.go#L91-L100

3.5 backport at https://github.com/etcd-io/etcd/pull/15907

PR #15840, opened 2023-05-07
tests/robustness: Implement Range limit and count

serathius:
cc @ahrtr @ptabor 

### DIFF
--- a/tests/robustness/model/deterministic.go
+++ b/tests/robustness/model/deterministic.go
@@ -146,14 +146,20 @@ func (s etcdState) step(request EtcdRequest) (etcdState, EtcdResponse) {
 					KVs: []KeyValue{},
 				}
 				if op.WithPrefix {
+					var count int64
 					for k, v := range s.KeyValues {
 						if strings.HasPrefix(k, op.Key) {
 							opResp[i].KVs = append(opResp[i].KVs, KeyValue{Key: k, ValueRevision: v})
+							count += 1
 						}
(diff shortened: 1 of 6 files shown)


tests/robustness/model/deterministic.go
```diff
@@ -146,21 +146,28 @@ func (s etcdState) step(request EtcdRequest) (etcdState, EtcdResponse) {
...
 						if strings.HasPrefix(k, op.Key) {
 							opResp[i].KVs = append(opResp[i].KVs, KeyValue{Key: k, ValueRevision: v})
+							count += 1
 						}
 					}
 					sort.Slice(opResp[i].KVs, func(j, k int) bool {
 						return opResp[i].KVs[j].Key < opResp[i].KVs[k].Key
 					})
+					if op.Limit != 0 && count > op.Limit {
+						opResp[i].KVs = opResp[i].KVs[:op.Limit]
```
ahrtr, line 160: Should we update the `count` in such case? 
```suggestion
						opResp[i].KVs = opResp[i].KVs[:op.Limit]
						count = op.Limit
```

ahrtr (approved):
LGTM with a minor comment.

tests/robustness/model/deterministic.go
serathius: No, count is the total count of entries, it intentionally can be higher then response size to inform client that they should paginate.

PR #15823, opened 2023-05-04
[3.4] Backport updating go to latest patch release 1.19.9

jmhbnz:
Backport the following to `release-3.4`:
- https://github.com/etcd-io/etcd/pull/15748
- https://github.com/etcd-io/etcd/pull/15821

PR #15841, opened 2023-05-07
Update 3.4 and 3.5 changelogs for compile with 1.19.9.

jmhbnz:
We recently backported:
 - https://github.com/etcd-io/etcd/pull/15823
 - https://github.com/etcd-io/etcd/pull/15822

Let's prep the changelog to show which go release binaries would compile with.

There could be more go patch releases between now and the next releases but we can always tweak it later to take those into account.


ahrtr (approved):
LGTM

Thanks @jmhbnz

PR #15842, opened 2023-05-08
log time when open backend db took too long

qsyqian:


jmhbnz (approved):
LGTM, thanks @qsyqian.

ahrtr:
> Just one question, given this is using `defer`, is there a situation where the open could fail but the log would still say `opened backend db`?

It etcd somehow fails to open the bboltDB, then etcd will panic. There is no chance to see the log `opened backend db`

PR #15864, opened 2023-05-09
chore: add newline as last char for .go-version

fuweid:


ahrtr:
It isn't a problem to not have a newline. There is no newline in K8s either,  I think it might be intentional.

fuweid:
@ahrtr Most of files are end of newline. This pr is to align with existing files. If it is not acceptable, I can close it.

ahrtr:
> @ahrtr Most of files are end of newline.

Yes, I am aware of it. But we get the go version using command `cat .go-version`, so I think we intentionally do not add the newline character. 

>  If it is not acceptable, I can close it.

Let's wait @serathius 's option, since he added the .go-version file.

serathius:
Don't have strong opinion. As long as the goversion and release workflows work I'm happy.

ahrtr:
Thanks for the feedback. 
 
I suggest to keep it consistent with K8s [.go-version](https://github.com/kubernetes/kubernetes/blob/master/.go-version), which has not a newline character at the end.

fuweid:
Thanks for the comment. Closing~

Issue #14040, opened 2022-05-13
Flaked integration/clientv3/examples ExampleCluster_memberAddAsLearner 


serathius:
https://github.com/etcd-io/etcd/runs/6423405520

lavacat:
I'll take a look at this

lavacat:
Quick update on this issue:

Found another run of this test that failed https://github.com/etcd-io/etcd/runs/6441619707?check_suite_focus=true

In both cases, we see the following logs:

`2022/05/13 13:11:47 2022-05-13T13:11:47.769Z	INFO	m0	applied a configuration change through raft	{"member": "m0", "local-member-id": "b81308bc4e5da240", "raft-conf-change": "ConfChangeAddLearnerNode", "raft-conf-change-node-id": "517ae5b12ac546a2"}
2022/05/13 13:11:47 2022-05-13T13:11:47.770Z	INFO	m0.raft	b81308bc4e5da240 ignoring conf change {ConfChangeRemoveNode 5871257613320406690 [] 11691486184239052569} at config voters=(12662322919470357889 13263954932397875776 18388213170794053236) learners=(5871257613320406690): possible unapplied conf change at index 25 (applied to 24)	{"member": "m0"}`

The problem is that raftLog.applied can be advanced after request is done. It's not synchronized with EtcdServer.appliedIndex increment in EtcdServer.apply

Looking for a good fix

serathius:
Nice find!

halegreen:
want to working on this while the run log is expired, can I see the log any where else? 

https://github.com/etcd-io/etcd/runs/6423405520 

<img width="656" alt="Screen Shot 2022-11-07 at 11 29 47 PM" src="https://user-images.githubusercontent.com/13792102/200349449-4789a1e9-7646-48e3-bdb6-7394509b9ee6.png">

lavacat:
@halegreen I don't have log from CI, but imo the root cause was `ignoring conf change ... possible unapplied conf change at index` that causes second ProposeConfChange to timeout. I've used TestMemberAddRemoveFail (see below) to reproduce.

My theory is that flakiness was eliminated by https://github.com/etcd-io/etcd/commit/73029fe89fece14af3a205490803b5b337c688a7 because it reduced time between [applying](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/raft.go#L213) and calling [r.Advance()](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/raft.go#L309).

[...]

Issue #13212, opened 2021-07-13
Issue with Raft Add or Remove node with `proposeConfChange`


Emon46:
I am using a 3 node cluster with raft leader election. Now I need to make it 5 nodes. Or sometimes need to remove a node. I have tried to add or remove nodes one by one with `proposeConfChange` where I have created a request like this
```   
  cc := raftpb.ConfChange{
        Type:    raftpb.ConfChangeRemoveNode,
        NodeID:  uint64(*node.NodeId),
      }
```
Followed the raft example API.
Then I have passed this cc in `ProposeConfChange(ctx, cc)` I have done this in a continuous loop until the node is added or removed into the cluster.
But After trying to remove node 4, I am facing the below error.
```
raft2021/07/12 21:14:25 INFO: 1 ignoring conf change {ConfChangeRemoveNode 4 [] 747} at config voters=(1 4) learners=(2 3): possible unapplied conf change at index 44 (applied to 39)
raft2021/07/12 21:14:26 INFO: 1 ignoring conf change {ConfChangeRemoveNode 4 [] 748} at config voters=(1 4) learners=(2 3): possible unapplied conf change at index 44 (applied to 39)
raft2021/07/12 21:14:29 INFO: 1 ignoring conf change {ConfChangeRemoveNode 4 [] 749} at config voters=(1 4) learners=(2 3): possible unapplied conf change at index 44 (applied to 39)
```
Is it a network latency problem? or I am missing something?

alexanderstephan:
I am running into the same issue. Did you manage to fix this?

Issue #12983, opened 2021-05-17
integration/clientv3/examples test flakes frequently


ptabor:
ExampleCluster_memberAddAsLearner

I managed to repro this with:

```
for i in `seq 1 100`; do (cd tests && 'env' 'go' 'test' '-timeout=15m' '--race=false' '--cpu=4' './integration/clientv3/examples' --count=1 -v -run ExampleCluster_memberAddAsLearner| tee log.log); done
```
but it does not always flake. 

Flake on Actions: 
https://github.com/etcd-io/etcd/pull/12981/checks?check_run_id=2595306923

Uploaded log: [logs_1206.zip](https://github.com/etcd-io/etcd/files/6491927/logs_1206.zip)

```
2021/05/16 18:49:34 2021-05-16T18:49:34.400Z	INFO	m2	started stream writer with remote peer	{"member": "m2", "local-member-id": "685c5c71ce9a9328", "remote-peer-id": "3b13bbeaeef551eb"}
2021/05/16 18:49:34 2021-05-16T18:49:34.400Z	INFO	m2	started stream writer with remote peer	{"member": "m2", "local-member-id": "685c5c71ce9a9328", "remote-peer-id": "3b13bbeaeef551eb"}
2021/05/16 18:49:34 2021-05-16T18:49:34.400Z	INFO	m2	started stream reader with remote peer	{"member": "m2", "stream-reader-type": "stream MsgApp v2", "local-member-id": "685c5c71ce9a9328", "remote-peer-id": "3b13bbeaeef551eb"}
2021/05/16 18:49:34 2021-05-16T18:49:34.400Z	INFO	m0	started remote peer	{"member": "m0", "remote-peer-id": "3b13bbeaeef551eb"}
2021/05/16 18:49:34 2021-05-16T18:49:34.400Z	INFO	m0	added remote peer	{"member": "m0", "local-member-id": "c5c5a20ca4073d6a", "remote-peer-id": "3b13bbeaeef551eb", "remote-peer-urls": ["http://localhost:32381"]}
2021/05/16 18:49:34 2021-05-16T18:49:34.400Z	INFO	m0	started stream writer with remote peer	{"member": "m0", "local-member-id": "c5c5a20ca4073d6a", "remote-peer-id": "3b13bbeaeef551eb"}
[...]

PR #15865, opened 2023-05-09
add failpoint raftBeforeAdvance to reproduce TestMemberPromoteMemberNotLearner reliably 

chaochn47:
REF: https://github.com/etcd-io/etcd/pull/15708#issuecomment-1540625591 and https://github.com/etcd-io/etcd/pull/15708#issuecomment-1535549919

> Since gofailpoint isn’t enabled by default for the existing [e2e test cases](https://github.com/etcd-io/etcd/tree/main/tests/e2e), so we will create a separate test suite under a new package “go.etcd.io/etcd/tests/v3/negative”(preferred) or “go.etcd.io/etcd/tests/v3/failpoint”

REF: [Terminate etcd on stalled storage writes Google doc Test section](https://docs.google.com/document/d/1U9hAcZQp3Y36q_JFiw2VBJXVAo2dK2a-8Rsbqv3GgDo/edit#heading=h.wwktogcu1ekk)

To reproduce
```
make gofail-enable 
pushd tests/failpoint 
go test -v -run TestMemberPromoteMemberNotLearnerWithFailpoint
popd
```

### DIFF
--- a/tests/failpoint/cluster_test_common.go
+++ b/tests/failpoint/cluster_test_common.go
@@ -1,0 +1,64 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package failpoint
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
+)
+
+func MemberPromoteMemberNotLearnerTest(t *testing.T, f Failpoint) {
+	integration2.BeforeTest(t)
+
+	require.NoError(t, f.Enable())
+	defer func() {
+		require.NoError(t, f.Disable())
+	}()
+
+	clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 3})
+	defer clus.Terminate(t)
+	// member promote request can be sent to any server in cluster,
+	// the request will be auto-forwarded to leader on server-side.
+	// This test explicitly includes the server-side forwarding by
+	// sending the request to follower.
+	leaderIdx := clus.WaitLeader(t)
+	followerIdx := (leaderIdx + 1) % 3
+	cli := clus.Client(followerIdx)
+
+	resp, err := cli.MemberList(context.Background())
+	if err != nil {
+		t.Fatalf("failed to list member %v", err)
+	}
+	if len(resp.Members) != 3 {
+		t.Fatalf("number of members = %d, want %d", len(resp.Members), 3)
+	}
+
+	// promoting any of the voting members in cluster should fail
+	expectedErrKeywords := "can only promote a learner member"
+	for _, m := range resp.Members {
+		_, err = cli.MemberPromote(context.Background(), m.ID)
+		if err == nil {
+			t.Fatalf("expect promoting voting member to fail, got no error")
+		}
+		if !strings.Contains(err.Error(), expectedErrKeywords) {
+			t.Fatalf("expect error to contain %s, got %s", expectedErrKeywords, err.Error())
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


tests/failpoint/failpoint.go
```diff
@@ -0,0 +1,41 @@
...
+	Disable() error
+}
+
+type failpointFunc struct {
+	name    string
+	payload string
+}
+
+func (f *failpointFunc) Enable() error {
+	return gofail.Enable(f.name, f.payload)
```
ahrtr, line 32: It isn't enough, you still need to run command `gofail enable <package> ` to transform the source in advance.

Reference:
- https://github.com/etcd-io/bbolt/blob/8b1ee10512ccb01d9d8744a620afc12939d26fb2/Makefile#L65-L66
- https://github.com/etcd-io/etcd/blob/b404d25d842cff4a281d1debdff764b2b126e74e/tests/robustness/makefile.mk#L38

chaochn47: Yeah, I intentionally leave it in line 22 of `tests/failpoint/cluster_test.go` as a comment. I wanted to enable this test verification in the CI one the fix is merged. So this test can only be run manually in local box.

ahrtr, line 32:
```suggestion
	if f == nil || len(f.name) == 0 {
	    return nil
	}
	return gofail.Enable(f.name, f.payload)
```

tests/failpoint/failpoint.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (f *failpointFunc) Disable() error {
+	return gofail.Disable(f.name)
```
ahrtr, line 36:
```suggestion
	if f == nil || len(f.name) == 0 {
	    return nil
	}
	return gofail.Disable(f.name)
```

tests/failpoint/cluster_test_common.go
```diff
@@ -0,0 +1,69 @@
...
+	integration2.BeforeTest(t)
+
+	if f != nil {
+		require.NoError(t, f.Enable())
+	}
+	defer func() {
+		if f != nil {
+			require.NoError(t, f.Disable())
+		}
+	}()
```
ahrtr, line 37:
```suggestion
	require.NoError(t, f.Enable())
	defer func() {
		require.NoError(t, f.Disable())
	}()
```

ahrtr, line 37:
```suggestion
	require.NoError(t, f.Enable())
	defer func() {
		require.NoError(t, f.Disable())
	}()
```

tests/failpoint/failpoint.go
```diff
@@ -0,0 +1,41 @@
...
+package failpoint
+
+import (
+	gofail "go.etcd.io/gofail/runtime"
+)
+
+type Failpoint interface {
+	Enable() error
+	Disable() error
+}
```
ahrtr, line 24: It doesn't make much sense to define an interface, because `failpointFunc` is only used in the same package. It makes sense if you move the interface into the test framework. 

We can keep it as it's for now, and discuss it separately.

chaochn47:
>  Update - Actions customers may see an increase in delayed runs, and some customers may see issues creating PRs and starting or resuming Codespaces. We continue to investigate and drive towards a mitigation.
> May 10, 2023 - 16:17 UTC 

https://www.githubstatus.com/

ahrtr (approved):
Overall looks good to me. 

Thanks @chaochn47 

To be clearer, currently only the robustness test cases are executed with failpoint enabled by default. All test cases under `tests/failpoint` will also be executed with failpoint enabled by default, and a followup PR is required to update the Makefile and workflow to enable gofailpoint for cases under `tests/failpoint`.

serathius:
I don't see the reason for creating another test type. For me this is an integration test that should skipped ('t.Skip`) if we didn't compile with gofail. For CI integration we should just enable gofail before running all tests.

chaochn47:
> I don't see the reason for creating another test type. For me this is an integration test that should skipped ('t.Skip`) if we didn't compile with gofail. For CI integration we should just enable gofail before running all tests.

Thanks @serathius! The conversation is moved to https://github.com/etcd-io/etcd/issues/15879. Hopefully there is an agreement on how to approach this.

chaochn47:
Hi @serathius, seems like from https://github.com/etcd-io/etcd/issues/15879 majority of commenters agree with adding the limited number of failpoint type of tests just like https://github.com/etcd-io/bbolt/tree/master/tests/failpoint. Do you have strong objections of this direction? 

Otherwise, can we let the PR merge and continue fixing https://github.com/etcd-io/etcd/issues/15528, https://github.com/etcd-io/etcd/issues/14040, https://github.com/etcd-io/etcd/issues/13212, https://github.com/etcd-io/etcd/issues/12983?

chaochn47:
okay. That should also work given just compiling etcd integration test with go fail should not produce any new issues. I can start from that and circle back on this topic if necessary.

PR #15873, opened 2023-05-10
tests/robustness: Add safeguards to client and history

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/tests/robustness/traffic/client.go
+++ b/tests/robustness/traffic/client.go
@@ -17,6 +17,7 @@ package traffic
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -27,10 +28,15 @@ import (
 	"go.etcd.io/etcd/tests/v3/robustness/model"
 )
 
+// RecordingClient provides a semi etcd client (different interface than
+// clientv3.Client) that records all the requests and responses made. Doesn't
+// allow for concurrent requests to ensure correct appending to history.
 type RecordingClient struct {
 	client   clientv3.Client
-	history  *model.AppendableHistory
 	baseTime time.Time
+	// mux ensures order of request appending.
+	mux     sync.Mutex
+	history *model.AppendableHistory
 }
 
 func NewClient(endpoints []string, ids identity.Provider, baseTime time.Time) (*RecordingClient, error) {
@@ -54,6 +60,10 @@ func (c *RecordingClient) Close() error {
 	return c.client.Close()
 }
 
+func (c *RecordingClient) Operations() model.History {
+	return c.history.History
+}
+
 func (c *RecordingClient) Get(ctx context.Context, key string) (*mvccpb.KeyValue, error) {
 	resp, err := c.Range(ctx, key, false)
 	if err != nil || len(resp) == 0 {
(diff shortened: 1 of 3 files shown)


chaochn47 (approved):
LGTM. 

The CI test was not kicked off due to github outage.

tests/robustness/traffic/client.go
```diff
@@ -27,10 +28,15 @@ import (
 	"go.etcd.io/etcd/tests/v3/robustness/model"
 )
 
+// RecordingClient provides a semi etcd client (different interface than
+// clientv3.Client) that records all the requests and responses made. Doesn't
+// allow for concurrent requests to ensure correct appending to history.
```
jmhbnz, line 33: Forgive me if this is a silly question, still learning... Does using a non standard client for robustness create any risk that the robustness tests will not behave as real world etcd will behave and therefore miss potential issues?

chaochn47: Good question: from my understanding: current k8s/etcd traffic [run](https://github.com/etcd-io/etcd/blob/ad20230e07241f7c8af1750eefd4d3b09a58cc54/tests/robustness/traffic/kubernetes.go#L63) in robustness test does not support concurrent read/write for each recording client. 

The nuance between robustness test and k8s usage is 
1. k8s re-uses the same etcd client per resource type to achieve concurrent r/w. 
2. robustness test uses multiple etcd clients to achieve concurrent r/w. 

To conform robustness test to k8s usage requires some re-factoring on appending to history sequentially that might complicates the code base. Could you please comment if there is a plan to do that? @serathius

tests/robustness/traffic/client.go
```diff
@@ -66,49 +76,62 @@ func (c *RecordingClient) Get(ctx context.Context, key string) (*mvccpb.KeyValue
 }
 
 func (c *RecordingClient) Range(ctx context.Context, key string, withPrefix bool) ([]*mvccpb.KeyValue, error) {
-	callTime := time.Since(c.baseTime)
 	ops := []clientv3.OpOption{}
 	if withPrefix {
 		ops = append(ops, clientv3.WithPrefix())
 	}
+	c.mux.Lock()
```
ahrtr, line 83: and remove line 92
```suggestion
	c.mux.Lock()
	defer c.mux.Unlock()
```

tests/robustness/traffic/client.go
```diff
@@ -66,49 +76,62 @@ func (c *RecordingClient) Get(ctx context.Context, key string) (*mvccpb.KeyValue
...
 		return nil, err
 	}
+	returnTime := time.Since(c.baseTime)
 	c.history.AppendRange(key, withPrefix, callTime, returnTime, resp)
+	c.mux.Unlock()
 	return resp.Kvs, nil
 }
 
 func (c *RecordingClient) Put(ctx context.Context, key, value string) error {
+	c.mux.Lock()
```
ahrtr, line 97: minor comment: defer unlock?  This comment applies the following couple of places

ahrtr (approved):
LGTM with a couple of minor comments.

tests/robustness/traffic/client.go
serathius: No, this is not a limitation of robustness test, nor k8s client. What I implemented here are safeguards over a naive  workaround around limitation of linearizability checking. "There can be only one concurrent request in one request stream". For now I assign to client only one request stream at a time. I could assign each request a unique stream, but that would make visualizations not readable.

chaochn47: @serathius Please bear me with too many question~

1. Why you name it as `we rotate stream id`, looks like `h.idProvider.ClientId()` always adds up by 1. 
2. What if we keep adding new records into the same stream and what will break?

chaochn47: Just to correlate between statement and code for other readers. 

`h.id` is updated with the next stream ID. (I guess porcupine calls it client ID but it is actually a stream from visualization). 

> For porcupine it's more like a identifier for sequential stream of operations. For example on failed requests we rotate stream id for client, to continue executing requests

https://github.com/etcd-io/etcd/blob/7c68be4cf3083c0f5e11808fa3816dc9eab23cf7/tests/robustness/model/history.go#L335-L346

serathius: > @serathius Please bear me with too many question~
> 
> 1. Why you name it as `we rotate stream id`, looks like `h.idProvider.ClientId()` always adds up by 1.

Yep, by rotate I mean pick another unique value. Porcupine requires it to be a counter from 0 to render nicely. 

> 2. What if we keep adding new records into the same stream and what will break?

I think you might get incorrect result from porcupine linearizer and porcupine visualization will render two blocks overlapping making it not readable.

serathius: Tried to document answers in https://github.com/etcd-io/etcd/pull/15880

PR #15874, opened 2023-05-10
tests/robustness: Fix pointer causing all cluster tests using kuberne…

serathius:
…tes traffic
cc @ahrtr @ptabor

### DIFF
--- a/tests/robustness/linearizability_test.go
+++ b/tests/robustness/linearizability_test.go
@@ -41,7 +41,7 @@ func TestRobustness(t *testing.T) {
 		scenarios = append(scenarios, testScenario{
 			name:      "ClusterOfSize1/" + traffic.Name,
 			failpoint: RandomFailpoint,
-			traffic:   &traffic,
+			traffic:   traffic,
 			cluster: *e2e.NewConfig(
 				e2e.WithClusterSize(1),
 				e2e.WithSnapshotCount(100),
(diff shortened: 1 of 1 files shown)


tests/robustness/linearizability_test.go
```diff
@@ -41,7 +41,7 @@ func TestRobustness(t *testing.T) {
 		scenarios = append(scenarios, testScenario{
 			name:      "ClusterOfSize1/" + traffic.Name,
 			failpoint: RandomFailpoint,
-			traffic:   &traffic,
+			traffic:   traffic,
```
chaochn47, line 44: Alternative way that does not depend on loop variable.

```golang
traffics := []traffic.Config{traffic.LowTraffic, traffic.HighTraffic, traffic.KubernetesTraffic}
for i, _ := range traffics {
   		scenarios = append(scenarios, testScenario{
			name:      "ClusterOfSize1/" + traffic.Name,
			failpoint: RandomFailpoint,
			traffic:   &traffics[i],
                }
}
```

ahrtr (approved):

tests/robustness/linearizability_test.go
serathius: Yea, having interface in config is not great. In ideal place you want configuration to be serializable and independent of implementation.

Issue #15647, opened 2023-04-06
ARM E2E test consistently failed


chaochn47:
### What happened?

All the E2E test arm64 workflows failed within 2 mins based on https://github.com/etcd-io/etcd/actions/workflows/e2e-arm64.yaml

### What did you expect to happen?

As part of the efforts of [T1 support for Arm64](https://github.com/etcd-io/etcd/issues/14748), the periodic test job should continuously pass. 

### Anything else we need to know?

It has been failing for 2 months.

chaochn47:
@kevinzs2048 @dims @geetasg for awareness.

pchan:
Is there anyone looking at this problem ? These tests have not been passing for quite sometime. The latest errors happen when trying to start etcd and not finding the db file [1]. 

To debug further we need access to equinix machines ([link](https://github.com/etcd-io/etcd/issues/14748#issuecomment-1400865267)). I can ask for access if no one is looking at it.


[1]
[...]

PR #15890, opened 2023-05-14
Ensure arm64 runners start with a clean checkout

jmhbnz:
As identified in https://github.com/etcd-io/etcd/issues/15647 we need to ensure that our arm64 runners are going to start with a fresh checkout each time they start a job.

An idea was mentioned in [stackoverflow](https://stackoverflow.com/questions/70483902/how-to-actually-clean-up-the-repository-on-self-hosted-runner-after-github-actio)  that I would like to propose. That is to have a pre-checkout step for all our `arm64` jobs that will clear the directory and ensure a fresh checkout.

cc @chaochn47, @serathius, @ahrtr

### DIFF
--- a/.github/workflows/e2e-arm64.yaml
+++ b/.github/workflows/e2e-arm64.yaml
@@ -17,6 +17,12 @@ jobs:
         target:
         - linux-arm64-e2e
     steps:
+    - name: 'Cleanup build folder'
+      run: |
+        ls -la ./
+        rm -rf ./* || true
+        rm -rf ./.??* || true
+        ls -la ./
     - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
     - uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
       with:
--- a/.github/workflows/robustness-nightly.yaml
+++ b/.github/workflows/robustness-nightly.yaml
@@ -22,6 +22,7 @@ jobs:
       testTimeout: 200m
       artifactName: main-arm64
       runs-on: "['self-hosted', 'Linux', 'ARM64']"
+      clean: 'true'
   release-35:
     uses: ./.github/workflows/robustness-template.yaml
     with:
--- a/.github/workflows/robustness-template.yaml
+++ b/.github/workflows/robustness-template.yaml
@@ -19,6 +19,10 @@ on:
         required: false
         type: string
         default: "['ubuntu-latest']"
+      clean:
+        required: false
+        type: string
+        default: 'false'
 permissions: read-all
 jobs:
   goversion:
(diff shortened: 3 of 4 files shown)


jmhbnz (commented):
One potential concern is that there could be a leaked process still running despite files being deleted, or the leaked process from previous test could interfere with `rm`?

If so we could add a process kill into the cleanup before the `rm` but I don't have access to the boxes so I'm not certain how the processes look when a test is running.

jmhbnz:
Keen for thoughts on this approach before I do a test via temporarily dropping schedule in favor of on pr.

ahrtr:
I do not get time to read through [customizing-the-containers-used-by-jobs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/customizing-the-containers-used-by-jobs). My immediate feeling is that it makes more sense to delegate such task ( cleanup ) to other tool (e.g. container) or service (not sure whether or not github has any such tool or service).

At least, I suggest to raise a ticket in https://github.com/actions/runner to get feedback or suggestion firstly.

jmhbnz:
> My immediate feeling is that it makes more sense to delegate such task ( cleanup ) to other tool (e.g. container) or service (not sure whether or not github has any such tool or service).

An alternative approach we can consider is https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job

ahrtr:
> An alternative approach we can consider is https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job

There is no essential difference from your solution in this PR.

jmhbnz:
> There is no essential difference from your solution in this PR.

Just a different implementation for us to consider, but still based on bash scripts yeah :)

> At least, I suggest to raise a ticket in https://github.com/actions/runner to get feedback or suggestion firstly.

Like us github actions project requests questions be raised via discussions so I went through to find any existing on this topic and there already several:
- https://github.com/orgs/community/discussions/45823
- https://github.com/orgs/community/discussions/46030
- https://github.com/orgs/community/discussions/51329

The only one that has a response suggested a similar approach of using a script to `rm` the workspace between runs and force a clean checkout.

chaochn47:
@dims can help add you to the box. 

We need to ensure multiple workflow runs do not have conflicts due to the pre-checkout step or running scripts after a job.

dims:
@jmhbnz Done try the following:

- ssh `root@86.109.7.233`
- ssh `root@147.28.151.226`

jmhbnz:
> * root@86.109.7.233

Working - many thanks @dims! 🙏🏻

jmhbnz:
Closing this in favor of work I have started for running jobs within containers in https://github.com/etcd-io/etcd/pull/15927

Issue #12913, opened 2021-04-30
RFC: StoreV2 deprecation plan


ptabor:
[The content of this post is being edited, and is not yet approved plan of record ]

# Background

In 3.4 storeV2 is still extensively used: 
  - User can opt in `--enable-v2` to unable V2 API that writes solely data to storeV2 files. 
  - Membership information is both stored in V2 & V3 (backend) stores  
  - Membership information is being read (recovered) from storeV2
  - Publishing membership happens through StoreV2 raft operation.
     - The new ClusterMemberAttrSet applier was implemented in etcd 3.5 so cannot be used by default in 3.5: to allow 3.4->3.5 rollback.   
  - During startup etcd assumes each WAL log snapshot is accompanied by storev2 snapshot. 

- Note: `etcdctl snapshot restore ` is not restoring V2 content   (producing fake storeV2 with membership information)

# Plan: 

## 3.5 release (updated by ahrtr )

Both V2 and V3 state are supported, but V3 becomes the source of truth. TODO:
 - [x] --experimental-enable-v2v3= promoted to `--enable-v2v3=...` or deprecated. [see https://github.com/etcd-io/etcd/issues/12905]
 - [x] --enable-v2v3=...  should allow mounting as root ?
 - [x] --enable-v2 prints deprecation warning [optional: when not combined with `--enable-v2v3=...` ]
 - [x] ClusterMemberAttrSet is implemented. 
 - [x] Membership is only read from backend v2  (https://github.com/etcd-io/etcd/pull/12914)
 - [x] --experimental-no-v2=PHASE_1_WRITE_ONLY mode that: 
     - cannot be used when `--enable-v2` is ON
     - if the storev2 snapshot is found accompanying the (used to be used) WAL log snapshot, 
       its validated whether it has no used-data (apart of membership & version)
     - storev2 is content created on demand using 'etcdctl snapshot restore' logic based on the last WAL snapshot 
 - [x] https://github.com/etcd-io/etcd/issues/18993  
 - [x] https://github.com/etcd-io/etcd/pull/18999
## 3.6 release (updated by serathius@, ahrtr )

V3 is the only meaningful state. V2 state is generated from V3 state to maintained backward compability.  TODO:
- [x] `--enable-v2` is docomissioned
- [x] PublishV3 is used solely
- [x] Remove v2 commands from etcdctl (https://github.com/etcd-io/etcd/pull/13643)
[...]

PR #15905, opened 2023-05-16
Update ValidateConfigurationChange to nil check for v2store

geetasg:
Part of v2 deprecation changes


server/config/v2_deprecation.go
```diff
@@ -24,10 +24,10 @@ const (
 	V2_DEPR_1_WRITE_ONLY = V2DeprecationEnum("write-only")
 	// V2store is WIPED if found !!!
 	V2_DEPR_1_WRITE_ONLY_DROP = V2DeprecationEnum("write-only-drop-data")
-	// V2store is neither written nor read. Usage of this configuration is blocking
-	// ability to rollback to etcd v3.5.
+	// V2store is neither written nor read for 3.6. v2snapshot is published only for backward compatibility
 	V2_DEPR_2_GONE = V2DeprecationEnum("gone")
 
+	//TODO geetasg does thie default need to change
```
serathius, line 30: By implementing v2snapshots from v3 membership data, you basically implement the `write-only-drop-data`. 
I would be for totally implementing this control flags as it gives false information on what is really supported. 

Maybe we should improve documentation and explain:
* etcd v3.4.X - supports both v2, and v3 API.
* etcd v3.5.X - supports v3 api by default, v2 API needs to be separably enabled.
* etcd v3.6.X - supports only v3 api, v2 API is not enabled and any v2 data will be purged. 

Note; upgrading from v3.5 to v3.6 and then downgrading back to v3.5 will be supported, but v2 data will be lost.  cc @ptabor @ahrtr

server/etcdserver/api/membership/cluster.go
```diff
@@ -303,9 +295,16 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
...
-func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
-	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
+
+	var membersMap map[types.ID]*Member
+	var removedMap map[types.ID]bool
+	membersMap, removedMap = c.be.MustReadMembersFromBackend()
+	if !shouldApplyV3 {
+		//TODO geetasg - RaftCluster purely based on backend cannot validate an entry older than ci
```
serathius, line 304: What do you mean by validate?

server/etcdserver/api/membership/cluster_test.go
```diff
@@ -975,3 +978,48 @@ func TestIsReadyToPromoteMember(t *testing.T) {
 		}
 	}
 }
+
+/*
+covered by TestV2DeprecationSnapshotMatches
+// TestMembershipStore tests code path used by snapshot
```
serathius, line 984: What's the plan with this test? Having both e2e and unit test is ok.

server/etcdserver/server.go
```diff
@@ -1104,6 +1097,14 @@ func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
 	})
 }
 
+func verifyConsistentIndexIsLatest(snapshot raftpb.Snapshot, cindex uint64) {
+	verify.Verify(func() {
```
serathius, line 1101: Nice

server/config/v2_deprecation.go
geetasg: sg. I will change the default to write-only-drop-data and update the docs when ready.

server/etcdserver/api/membership/cluster.go
geetasg: I had a test failure on validation which looked like was due to trying to validate using 'future' info. Ref:https://github.com/etcd-io/etcd/pull/15905/files#diff-164b95d19232df0c51cd511fe4ca61de738660fc11db6ddcdafa0ec5f5c80a35R1978

I will revisit to add a test for that code path once it is better understood.

server/etcdserver/api/membership/cluster_test.go
geetasg: I had added this test to simulate 3.5 style recovery. But now it is getting in the way of removing SetStore (and v2store from RaftCluster). I am investigating modification to the test or will remove it.

server/etcdserver/api/membership/storev2.go
```diff
@@ -95,11 +96,13 @@ func mustSaveMemberToStore(lg *zap.Logger, s v2store.Store, m *Member) {
 
 func mustDeleteMemberFromStore(lg *zap.Logger, s v2store.Store, id types.ID) {
 	if _, err := s.Delete(MemberStoreKey(id), true, true); err != nil {
-		lg.Panic(
-			"failed to delete member from store",
-			zap.String("path", MemberStoreKey(id)),
-			zap.Error(err),
-		)
+		if !strings.Contains(err.Error(), "Key not found") {
```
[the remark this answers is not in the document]
geetasg: so far only found this one ("key not found") that needs to be absorbed - because we start with empty store now. I will keep an eye out for any other failures that can be expected.

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/12913

geetasg:
Notes - removing v2store from etcdserver is causing following test to fail. Debugging in progress.

Test from - 
etcd/tests/common

To repro failure - 
go test -v --tags e2e -run TestMemberList --race --cpu=4

Currently backed out the checkin to not use v2store and v2applier. It will be reintroduced once above failure is fixed. If anyone has any insight on why this test would fail due to v2store removal, please drop a note here.

PR #15906, opened 2023-05-16
clientv3: refactor snapshot SaveWithVersion

qsyqian:

### DIFF
--- a/client/v3/snapshot/v3_snapshot.go
+++ b/client/v3/snapshot/v3_snapshot.go
@@ -45,7 +45,7 @@ func hasChecksum(n int64) bool {
 // selected node, and saved snapshot is the point-in-time state of
 // the selected node.
 // Etcd <v3.6 will return "" as version.
-func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, dbPath string) (version string, err error) {
+func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, dbPath string) (string, error) {
 	cfg.Logger = lg.Named("client")
 	if len(cfg.Endpoints) != 1 {
 		return "", fmt.Errorf("snapshot must be requested to one selected node, not multiple %v", cfg.Endpoints)
@@ -91,7 +91,7 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 		zap.String("endpoint", cfg.Endpoints[0]),
 		zap.String("size", humanize.Bytes(uint64(size))),
 		zap.Duration("took", time.Since(start)),
-		zap.String("etcd-version", version),
+		zap.String("etcd-version", resp.Version),
 	)
 
 	if err = os.Rename(partpath, dbPath); err != nil {


jmhbnz (commented):
Hey @qsyqian - Thanks for raising this suggestion. Can you please provide an explanation on what problems you believe this refactor solves? After a very quick read I'm not sure if the refactor is an improvement but I'm relatively new to this project so I might be missing something.

qsyqian:
> Hey @qsyqian - Thanks for raising this suggestion. Can you please provide an explanation on what problems you believe this refactor solves? After a very quick read I'm not sure if the refactor is an improvement but I'm relatively new to this project so I might be missing something.

Thx for review. Actually i want to fix the value named `version` always is empty in https://github.com/etcd-io/etcd/blob/main/client/v3/snapshot/v3_snapshot.go#L94 😄

tjungblu:
good catch, totally makes sense to rather use the declared return variable instead.

/lgtm (non-binding)

jmhbnz (approved):
> Actually i want to fix the value named version always is empty in https://github.com/etcd-io/etcd/blob/main/client/v3/snapshot/v3_snapshot.go#L94 smile

Ahhh I see now - by default github has that section of the diff collapsed with no change. Agree makes sense now, thanks @qsyqian!

client/v3/snapshot/v3_snapshot.go
```diff
@@ -95,8 +96,8 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 	)
```
fuweid, line 96: maybe we use `(_ string, err error)` for return value and update L94 with `resp.Version`.
In go, it's unlikely to use `resp.Version` if the error is not nil. So, I think we can update the L79/L82/L88/L98 with `return "", error`. What do you think?

ahrtr:
`Don't name result parameters just to avoid declaring a var inside the function; that trades off a minor implementation brevity at the cost of unnecessary API verbosity.`

FYI. https://github.com/golang/go/wiki/CodeReviewComments/#named-result-parameters

ahrtr:
Overall it isn't a big deal, but let's try to follow the best practice. Please
- change [version](https://github.com/etcd-io/etcd/blob/9b10f2000443df0ea9cfc7aa2c5587b8af2d8544/client/v3/snapshot/v3_snapshot.go#L94) to `resp.Version`;
- Remove the named result parameters.

fuweid (approved):
LGTM

chaochn47 (approved):

client/v3/snapshot/v3_snapshot.go
```diff
@@ -76,26 +76,26 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 	var size int64
 	size, err = io.Copy(f, resp.Snapshot)
 	if err != nil {
-		return resp.Version, err
+		return "", err
```
ahrtr, line 79: Why change this? I would suggest to return the resp.Version, because it's the real value returned from the server. It's up the caller how to use it. We should only return "" when we get [an error response](https://github.com/etcd-io/etcd/blob/f3c9db9c460bba556f4922a5acc3be496eccda40/client/v3/snapshot/v3_snapshot.go#L72) from etcdserver.

This comment applies to the following couple of places.

fuweid: if the err is not nil, I think that in go practice, the caller won't use `version`.

ahrtr: Overall it isn't a big deal. 

Note that `SaveWithVersion` is a public API. Previously it returns a valid version as long as the first communication with etcdserver is successful. Now the behavior changed in this PR although it isn't a big problem.

You never know how users will use a API. If I use the API, I might get the version included in an error log message anyway; and I will be a little surprised that why previously I can see a valid version but not now.

So let's only fix bugs, and try not to change any behavior unless it's necessary.

serathius:
Maybe I don't understand something, why we are spending so much time on a refactor that doesn't help overall code quality? I'm not saying that such refactors are not useful, but the decision to accept it should be swift. It either makes local code better and we merge it or it doesn't help and we reject it. Please don't try to add more suggestions over a simple refactor, merge quick and then propose a follow up PR.

qsyqian:
sorry for take so long time on this litte change, just update as your suggestion. cc @ahrtr

ahrtr:
> It either makes local code better and we merge it or it doesn't help and we reject it. Please don't try to add more suggestions over a simple refactor, merge quick and then propose a follow up PR.

We just accept the bug fix, but reject the unnecessary behavior change.

ahrtr (approved):
LGTM

Thanks.

ahrtr:
@qsyqian could you please double check release-3.5 and release-3.4 and backport the fix if needed? thx

qsyqian:
> @qsyqian could you please double check release-3.5 and release-3.4 and backport the fix if needed? thx

It's no need to backport after check release-3.5 and release-3.4. Thanks for reminding

PR #15909, opened 2023-05-16
tests/robustness: Expect revions to be unique for Kubernetes Traffic

serathius:


tests/robustness/watch.go
```diff
@@ -202,19 +203,31 @@ func validateOrdered(t *testing.T, report traffic.ClientReport) {
...
-		revision int64
-		key      string
-	}
-	uniqueOperations := map[revisionKey]struct{}{}
-	for _, resp := range report.Watch {
-		for _, event := range resp.Events {
-			rk := revisionKey{key: event.Op.Key, revision: event.Revision}
-			if _, found := uniqueOperations[rk]; found {
-				t.Errorf("Broke watch guarantee: Unique - an event will never appear on a watch twice, key: %q, revision: %d, client: %d", rk.key, rk.revision, report.ClientId)
+func validateUnique(t *testing.T, expectUniqueRevision bool, report traffic.ClientReport) {
```
jmhbnz, line 206: There is a bit of duplication in this function that we could eliminate with something like below:

```go
func validateUnique(t *testing.T, expectUniqueRevision bool, report traffic.ClientReport) {
	uniqueOperations := map[interface{}]struct{}{}

	for _, resp := range report.Watch {
			var key interface{}
			if expectUniqueRevision {
				key = event.Revision
			} else {
				key = struct {
					revision int64
					key      string
				}{event.Revision, event.Op.Key}
			}

			if _, found := uniqueOperations[key]; found {
				t.Errorf("Broke watch guarantee: Unique - an event will never appear on a watch twice, key: %q, revision: %d, client: %d", event.Op.Key, event.Revision, report.ClientId)
			}
			uniqueOperations[key] = struct{}{}
		}
	}
}
```

Although there are some potential downsides for the interface type approach like reduced type safety and probably a little less performant.

This comment is just an idea, I'm happy for it to be discarded if it doesn't make sense in the context of etcd code priorities.

tests/robustness/traffic/etcd.go
```diff
@@ -73,6 +73,10 @@ type etcdTraffic struct {
 	largePutSize int
 }
 
+func (t etcdTraffic) ExpectUniqueRevision() bool {
```
jmhbnz, line 76: I'm curious about the reasons for putting these bools into functions versus just having it as part of traffic config? Are we expecting to have complex criteria in future for determining if a given traffic type will expect unique revisions? It feels like something that is more static to me but I might be missing something.

tests/robustness/watch.go
```diff
@@ -202,19 +203,31 @@ func validateOrdered(t *testing.T, report traffic.ClientReport) {
...
+	if expectUniqueRevision {
+		uniqueOperations := map[int64]struct{}{}
+		for _, resp := range report.Watch {
+			for _, event := range resp.Events {
+				if _, found := uniqueOperations[event.Revision]; found {
+					t.Errorf("Broke watch guarantee: Unique - an event will never appear on a watch twice, key: %q, revision: %d, client: %d", event.Op.Key, event.Revision, report.ClientId)
```
chaochn47, line 212: This might be broken if we support `DeleteCollections` in robustness test. For example, `Remove all pods under a namespace`.

chaochn47 (approved):
LGTM (non-binding).

tests/robustness/watch.go
ahrtr, line 206: The revision must be unique: either global or on each key.
```suggestion
func validateUnique(t *testing.T, expectUniqueRevisionOnEachKey bool, report traffic.ClientReport) {
```

Does it mean that Kubernetes traffic will never have duplicated key?

serathius: Note: duplication is not bad and even good if it makes code simpler. Using `interface{}` could lead to code bloat due to casting and make it less readable. 

Still I like your example because usage `interface{}` is not bloating code too much.

tests/robustness/traffic/etcd.go
serathius: Because it's property of traffic code, it could be inferred from write types used, however it's also not clear for them. For example there are three etcd requests types currently that generate multiple operation per revision.

tests/robustness/watch.go
serathius: Yes, K8s watch correctness depends on that.

serathius:
Addressed comments, PTAL @ahrtr @jmhbnz

ahrtr (approved):

jmhbnz (approved):

PR #15915, opened 2023-05-17
server/storage/schema: prefer equal to compare for equality comparisons

hexfusion:
Benchmarks show `bytes.Equal` is much faster than `bytes.Compare`[1] for equality comparisons and preferred per the example in docs[2].

[1] https://www.sobyte.net/post/2022-01/go-bytes-compare-equal/
[2] https://pkg.go.dev/bytes#Compare

### DIFF
--- a/server/storage/schema/bucket.go
+++ b/server/storage/schema/bucket.go
@@ -89,8 +89,8 @@ func DefaultIgnores(bucket, key []byte) bool {
 	// consistent index & term might be changed due to v2 internal sync, which
 	// is not controllable by the user.
 	// storage version might change after wal snapshot and is not controller by user.
-	return bytes.Compare(bucket, Meta.Name()) == 0 &&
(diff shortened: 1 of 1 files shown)


ahrtr (approved):
LGTM

Thanks @hexfusion

jmhbnz (approved):
LGTM - Though is there any context on how much of a difference this will make for etcd?   The linked benchmarks are in nanoseconds per operation, I'm wondering if there is expected to be a more pronounced impact for etcd? 

If not it would probably not be a priority to review any other code using `bytes.Compare` though at first glance this looks like it might be the only instance where we do a straight `==`.

fuweid (approved):
LGTM

ahrtr:
> where we do a straight `==`.

FYI. Slice, map, and function types are not comparable in golang.

PR #15927, opened 2023-05-19
Run arm64 tests in containers on self hosted runners

jmhbnz:
This pull request updates our suite of nightly `arm64` tests to run within docker containers on our self hosted `arm64` runners in order to fix isolation issues with workflow runs on self hosted runners.

Below is a summary of what was required to get this working:
- Update both runners to ensure `docker-io` was installed and running (was missing on one).
- Ensure the `runner` user on both machines was added to the `docker` group.
- Restart runner to propagate user permissions update.
- Update workflow files to add new `container: golang:1.19-bullseye` as the workflow image (this matches our [devcontainer](https://github.com/etcd-io/etcd/blob/main/.devcontainer/devcontainer.json).
- Set default shell for containers to bash to work around https://github.com/actions/checkout/issues/1169
- Create separate robustness template for `arm64`.
- Verified workflows running successfully within containers by temporarily changing to `on: pull_request`

I've verified that the updated workflows are running successfully, refer below:

| Workflow         | Link                                                                    |
|------------------|-------------------------------------------------------------------------|
| Unit/Integration | https://github.com/etcd-io/etcd/actions/runs/5043837640                 |
| End to end       | https://github.com/etcd-io/etcd/actions/runs/5043837630/jobs/9046140490 |
| Robustness       | https://github.com/etcd-io/etcd/actions/runs/5043837736/jobs/9046143973 |


jmhbnz:
Note for reviewers, commit https://github.com/etcd-io/etcd/pull/15927/commits/a10ae70e6e693d869628db32652a9589f81eba5d is specifically addressing a variety of lint issues with our workflows files reported by `yamllint` to bring the files closer to yaml best practices.

Perhaps review commits one by one will be easier than all at same time.

.github/workflows/robustness-template-arm64.yaml
```diff
@@ -0,0 +1,73 @@
+---
```
jmhbnz, line 1: To make this work we are faced with the two choices:
  1. Update `robustness-template.yaml` to run all robustness jobs (not just `arm64`) in containers.
  2. Add a robustness template specifically for `arm64`.

As you will see in this pull request I've gone with option 2 as it is a more gradual change. We could revisit later on and adopt running jobs in containers more widely to consolidate once we know it's operating nicely.

jmhbnz:
Another note for reviewers - as a result of this pull request, in future we will have the ability to pivot to running `arm64` tests on each pull request.

Our machines have capacity to do this, and with the isolation introduced in this pull request we will be free to add more runners and scale up so multiple jobs can be running at the same time. 

I will monitor how this initial change goes and in a month or two can create the issue to propose how we can scale up the `arm64` runners and start running at least basic unit and integration on `arm64` for each pull request.

ahrtr:
I see that there are lots of format (e.g. indent) change. Could you raise a separate PR for the format only change?

serathius:
Four things.As @ahrtr requested, please make format change a separate file. 

Second, please remember that running in containers not always mean it's leaving clean slate. Containers and images will also leak from time to time. I don't think we need to implement a mitigation yet, but I would suggest we run a cron that cleans docker images and checks if there are containers running for longer than 3 hour (run timeout time) and kills them.

Third, looks like we are adding more and more changes to arm64 workers and we don't have a unified way to document it. I think we should create a bootstrap script for runners so we will not lose the knowledge on setting up docker for running tests in containers. 

Fourth, we should discuss the tradeoff of having arm64 running different setup versus moving all workflows to containers. I personally prefer to avoid special cases so it's easier to use and reason about workers. I don't think there are any blockers for running all workflows on container and there are benefits like making the test environment more consistent with development configuration provided as devcontainer.

jmhbnz:
> Four things.As @ahrtr requested, please make format change a separate file.

Done - Thanks for the feedback, I will raise a follow-up to do the lint fixing to keep this one focused.

 
> Second, please remember that running in containers not always mean it's leaving clean slate. Containers and images will also leak from time to time. I don't think we need to implement a mitigation yet, but I would suggest we run a cron that cleans docker images and checks if there are containers running for longer than 3 hour (run timeout time) and kills them.

Agree - I will raise a follow-up issue and get this done and documented.


> Third, looks like we are adding more and more changes to arm64 workers and we don't have a unified way to document it. I think we should create a bootstrap script for runners so we will not lose the knowledge on setting up docker for running tests in containers.

Completely agree -  I had to figure some things out the hard way to get this pr done. I think at bare minimum we should have some procedures documented for restarting or updating the runners. As above I will raise this as a second follow-up task and sort this out.

 
> Fourth, we should discuss the tradeoff of having arm64 running different setup versus moving all workflows to containers. I personally prefer to avoid special cases so it's easier to use and reason about workers. I don't think there are any blockers for running all workflows on container and there are benefits like making the test environment more consistent with development configuration provided as devcontainer.

Also agree - In my view we should move towards all workflows running in container once we prove from this pull request that it works nicely. It will allow us to consolidate the robustness template file once more and means we aren't maintaining two different approaches.

.github/workflows/e2e-arm64.yaml
```diff
@@ -11,13 +11,19 @@ jobs:
...
       matrix:
         target:
         - linux-arm64-e2e
     steps:
     - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+    # https://github.com/actions/checkout/issues/1169
+    - run: git config --system --add safe.directory '*'
     - uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
       with:
         go-version: ${{ needs.goversion.outputs.goversion }}
```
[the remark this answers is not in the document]
jmhbnz: I believe there is zero impact - the `setup-go` action is correctly setting go path. I created a quick repo to test: https://github.com/jmhbnz/actions-go-test/actions/runs/5063595694/jobs/9090374749 

After initial checkout inside the `golang:1.19-bullseye` we can see that the go path is the standard container one.  After `setup-go` we can see go path change so our explicitly defined go version will continue to take priority:

Logs of `setup-go`
```bash
Setup go version spec 1.19.9
Found in cache @ /__t/go/1.19.9/x64
Added go to the path
Successfully set up Go version 1.19.9
```

Go path before and after:

image

.github/workflows/robustness-template-arm64.yaml
```diff
@@ -0,0 +1,72 @@
...
+  goversion:
+    uses: ./.github/workflows/go-version.yaml
+  test:
+    timeout-minutes: 210
+    runs-on: ${{ fromJson(inputs.runs-on) }}
+    needs: goversion
+    container: golang:1.19-bullseye
+    defaults:
+      run:
+        shell: bash
```
ahrtr, line 34: You do not configure the volumes, so I think the artifact may not be correctly uploaded when robustness runs into any issues?

ahrtr:
> Migrating all workflows to running containers


I am neutral to this. The github standard runners do not have the cleanup issue as our self hosted ARM runners. Technically speaking it isn't necessary to migrate all workflows into containers. But the good side to migrate all workflows into containers is that we have unified way to run all workflows, and accordingly less maintenance burden. 

If we migrate all workflow,  then eventually `robustness-template-arm64.yaml` and `robustness-template.yaml` should be combined into one YAML file.

Either way, let's watch how it's going on the self-hosted runner firstly.

Great work, thanks @jmhbnz

.github/workflows/robustness-template-arm64.yaml
ahrtr: Based on [running-jobs-in-a-container#overview](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container#overview), the test script and `actions/upload-artifact` will run in separate containers, and accordingly they will have different local filesystems & volumes if you do not mounts the host paths into the containers. 

I think you can verify this by intentionally fail the test.

```
If you have steps that use both script and container actions, the container actions 
will run as sibling containers on the same network with the same volume mounts.
```

jmhbnz: Have intentionally failed test, can you please check https://github.com/etcd-io/etcd/actions/runs/5066334564/jobs/9096012829?pr=15927 ?

ahrtr: Confirmed that the artifact was successfully uploaded. It seems that I misunderstood the meaning of `container actions`, and the `actions/upload-artifact` isn't a container action, so all steps are running in the same container.

Thanks for the confirmation.

jmhbnz: Was a good question and worth double checking, thanks! 🙏🏻

ahrtr (approved):
LGTM

Thanks @jmhbnz

serathius (approved):

ahrtr:
We already have the overall plan https://github.com/etcd-io/etcd/issues/15951, so let me merge this PR and let's watch how it's going on the self-hosted runner for 1 ~ 2 weeks.

jmhbnz:
Happy to report after the first four days things are looking green for the new container based `arm64` job runs! 😌 
- https://github.com/etcd-io/etcd/actions/workflows/e2e-arm64.yaml 
- https://github.com/etcd-io/etcd/actions/workflows/robustness-nightly.yaml
- https://github.com/etcd-io/etcd/actions/workflows/tests-arm64.yaml

Will keep monitoring for another week or so.

dims:
@jmhbnz very cool!

chaochn47:
In case it's overlooked, it looks like unexpected that the common e2e test result will be cached. I guess the container volume needs to be cleaned up each workflow run. 

```
% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' 'go.etcd.io/etcd/tests/v3/common' '--tags=e2e' '-timeout=30m')
ok  	go.etcd.io/etcd/tests/v3/common	(cached)
```

jmhbnz:
> In case it's overlooked, it looks like unexpected that the common e2e test result will be cached. I guess the container volume needs to be cleaned up each workflow run.

Great spotting @chaochn47 that definitely needs to be looked at!  I've raised https://github.com/etcd-io/etcd/issues/15986 to investigate and will raise any tweak required.

PR #15942, opened 2023-05-23
benchmarks: pin dependencies of rw-heatmaps

cenkalti:
Signed-off-by: Cenk Alti <cenkalti@gmail.com>

### DIFF
--- a/tools/rw-heatmaps/plot_data.py
+++ b/tools/rw-heatmaps/plot_data.py
@@ -79,106 +79,6 @@ def load_data_files(*args):
     return res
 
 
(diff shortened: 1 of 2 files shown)


serathius (approved):

jmhbnz (approved):
LGTM

ahrtr (approved):
LGTM

I tried to replace the python modules with golang package in https://github.com/etcd-io/etcd/pull/15060, but it has concern from other maintainers. Anyway, It would be better we can completely remove the dependency on python and its modules eventually.

PR #15946, opened 2023-05-24
Add the rotation worksheet into the dependency management document

ahrtr:

Please feel free to suggest a better way to track this. I am OK as long as majority are happy.

cc @mitake @ptabor @serathius @jmhbnz @chaochn47 @fuweid @lavacat @pchan @ArkaSaha30 @sharathsivakumar

### DIFF
--- a/Documentation/contributor-guide/dependency_management.md
+++ b/Documentation/contributor-guide/dependency_management.md
@@ -8,6 +8,7 @@ Dependency management
     - [Steps to bump a dependency](#steps-to-bump-a-dependency)
     - [Indirect dependencies](#indirect-dependencies)
     - [About gRPC](#about-grpc)
+    - [Rotation worksheet](#rotation-worksheet)
 - **[Stable branches](#stable-branches)**
 
 # Main branch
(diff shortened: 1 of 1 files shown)


fuweid (commented):
LGTM

Maybe we should add the final state is that the dep can be handled by bot, something like that.

chaochn47 (approved):

jmhbnz (approved):
LGTM - So far the roster is working ok I think.  In future if we have any gaps we can always do a quick message on etcd slack or make a `good-first-issue` to get some more people involved.

Documentation/contributor-guide/dependency_management.md
```diff
@@ -83,6 +84,11 @@ The plan is to remove the dependency on some grpc-go's experimental API firstly,
 
 `go.opentelemetry.io/otel` version update is indirectly blocked due to this gRPC issue. Please get more details in [pull/15810](https://github.com/etcd-io/etcd/pull/15810).
 
+## Rotation worksheet
+The dependabot scheduling interval is weekly; it means dependabot will automatically raise a bunch of PRs per week.
```
pchan, line 88: Does it make sense to make it every other week ? There are tradeoffs either way, but it in my observation the number of PRs is uneven weekly.

ahrtr: I do not see a obvious concern to set schedule to weekly.  Also it seems that there are [only three interval types](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval):
- daily
- weekly
- monthly

ahrtr:
ping @mitake @ptabor @serathius @spzala

serathius (approved):

Issue #15955, opened 2023-05-25
Introduce linting for actions workflow files


jmhbnz:
### What would you like to be added?

Our etcd github actions workflows currently fail a number of very basic yamllint tests, examples that are worth fixing are below:

```bash
./.github/workflows/contrib.yaml
  1:1       warning  missing document start "---"  (document-start)
  11:5      error    wrong indentation: expected 6 but found 4  
```

```bash
./.github/workflows/scorecards.yml
  1:1       warning  missing document start "---"  (document-start)
  8:16      error    too many spaces inside brackets  (brackets)
  8:23      error    too many spaces inside brackets  (brackets)
  49:1      error    trailing spaces  (trailing-spaces)
```

We should ignore the follow warnings:

Warning | Reason
-- | --
Line too long | Usually problematic to try and address.
truthy value should be one of [false, true] | Yamllint doesn’t understand the concept of `on: ` for github actions files.
too few spaces before comment | This generally refers to the comments added automatically by dependabot for ci actions versions so we should leave these.

While the existing errors should be fixed we should also consider introducing a minimal yamllint configuration and potentially `make` target to ensure new errors don't creep in over time and we maintain best practices.

### Why is this needed?

Enforce best practices for yaml files.

tao12345666333:
Please assign to me, thanks

jmhbnz:
All yours @tao12345666333, thanks for offering to help!  

A couple quick notes on your approach for this:
- Consider doing an initial pull request or multiple smaller pull requests just to fix existing lint issues, then doing a follow-up pull request for the implementation of automated lint checks as part of the static analysis github actions workflow.
- When introducing the automated checking please do so via new `Makefile` `make verify ...` and `make fix ...` targets.

Feel free to reply here or message me on the Kubernetes slack if any questions pop up 🙂

jmhbnz:
Closing as both sub tasks have been completed, thanks @tao12345666333 🚀

PR #15960, opened 2023-05-26
ci: Introduce yamllint for actions workflow files

tao12345666333:

Initial resolution of https://github.com/etcd-io/etcd/issues/15955
Automated lint checks will be implemented in subsequent PRs.


Makefile
```diff
@@ -132,6 +132,10 @@ verify-genproto:
 verify-goimport:
 	PASSES="goimport" ./scripts/test.sh
 
+.PHONY: verify-yamllint
+verify-yamllint:
```
serathius, line 136: Can you add a matching fix-yamllint rule?

Makefile
```diff
@@ -132,6 +132,10 @@ verify-genproto:
...
+	yamllint .github/
```
serathius, line 137: We also have file .golangci.yaml in root directory. Can you run yamllint on whole project?

Makefile
jmhbnz: Note for context `yamllint` is only a linter, it doesn't have automatic formatting.  We would need to add another utility like `yamlfmt` for auto formatting https://github.com/google/yamlfmt.

serathius: If they are compatible then we should. As we add linters it can get more and more annoying for contributors to fix any small thing. Here comes the `make fix` that should make addressing majority of errors reported by linters.

tao12345666333: Yes, I plan to add rules for `fix-yamlint`. There are many tools that can automatically format YAML, but not all of them work well.  https://github.com/adrienverge/yamllint/issues/62

I want to add it in the follow-up PR after comparing these tools. I will also try [the tool ](https://github.com/google/yamlfmt) mentioned by @jmhbjz.

Makefile
serathius: Bump

tao12345666333: Addressed, thanks!

serathius:
Please prioritize adding `make fix-yamllint` as I expect not having it will annoy contributors changing yaml file. At least me as I write yaml in totally different format then yamllint.

jmhbnz (approved):
LGTM - Thanks for getting this initial tidy up completed @tao12345666333!  

As mentioned by @serathius, please explore the options for a automated fix utility 🙏🏻

ahrtr (approved):
Thanks @tao12345666333

tao12345666333:
Thanks for review. I will submit a follow-up PR soon

Issue #15919, opened 2023-05-18
Incorrect hash when resuming scheduled compaction after etcd restarts


kennydo:
### What happened?

I tried out using the `--experimental-compact-hash-check-enabled` flag on my etcd cluster today, and I think I might have encountered a race condition when a server starts up/resumes a scheduled compaction that causes the hash value to be incorrect.

About my setup: I have a 5 member cluster, running as k8s pods. They already have the `--experimental-initial-corrupt-check` flag enabled.

I was adding the `--experimental-compact-hash-check-enabled` flag (and also upgrading from 3.5.6 to 3.5.9) on a 5-member node. For 3 of the 4 etcd members, I had already stopped the etcd pod, updated their container URLs to 3.5.9, added the new hash check flag, and then restarted the etcd pod. I then stopped the 4th follower (which I'll call FOO), edited its pod config, and brought it back up. I then stopped the leader, and did the same update/restart. This triggered a leader election, and one of the former followers (which I'll call BAR) became the new leader.

The new leader then performed the periodic compact hash check, as configured. But the strange thing is that, even though all of my pods were up and running (and had passed the initial corruption check), the very first periodic check failed!

Due to the liveness probe problem mentioned in https://github.com/etcd-io/etcd/issues/13340, I had to deal with my pods getting restarted by kube due to the alarm... But when I disarmed the CORRUPT alarm so my etcd pods could stay up and running, they all again passed the initial corrupt data check, and logged the same hashes for the following "storing new hash" logs. So that's why I suspect there was very unlucky timing and a race condition that caused FOO to calculate the wrong hash value and use that when BAR asked for its hash during that periodic compaction hash check.

### What did you expect to happen?

I expected the periodic compaction hash check to succeed and for my cluster be healthy because:
1) the initial data check passed on all members of the cluster
2) the hash/compact-revision that FOO logged with the "storing new hash" line both before and after the problematic hash matched the hashes in the rest of the cluster

### How can we reproduce it (as minimally and precisely as possible)?

I'm not sure how to reproduce it. Maybe create a multi-member cluster with enough data so that compaction takes a non-trivial amount of time, trigger a compaction, and then stop one of the members right in the middle of the compaction so that compaction fails and needs to resume compacting when it comes back up?

### Anything else we need to know?

I saw https://github.com/etcd-io/etcd/issues/15548, but that mentions the cluster ID changing, and in this scenario where just I'm restarting etcd with different flags, I don't think the cluster ID would change.

[...]

PR #15965, opened 2023-05-26
cli: Add etcdutl snapshot hashkv command

### LINKED ISSUE #15061: Enhance `etcdutl` to calculate hash of the data up to a given rev

Currently command `etcdctl endpoint hashkv` can return the hash up to a given rev of endpoints. It requires the etcdserver is still running.   But in some cases, when users raise inconsistency issue, the etcd cluster might not be running any more. So we need to support calculating the hash offline, in other words, we need to support calculating hash using `etcdutl` as well.

### Why is this needed?

It can improve the diagnosability / supportability

cenkalti:
Fixes #15061

```json
% bin/etcdctl endpoint hashkv -w json | jq
[
  {
    "Endpoint": "127.0.0.1:2379",
    "HashKV": {
      "header": {
        "revision": 214,
        "raft_term": 17
      },
      "hash": 902327963,
      "compact_revision": 150,
      "hash_revision": 214
    }
  }
]
% bin/etcdutl snapshot hashkv default.etcd/member/snap/db -w json | jq
{
  "hash": 902327963,
  "hashRevision": 214,
  "compactRevision": 150
}
```

### DIFF
--- a/etcdutl/etcdutl/hashkv_command.go
+++ b/etcdutl/etcdutl/hashkv_command.go
@@ -1,0 +1,74 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package etcdutl
+
+import (
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+
+	"go.etcd.io/etcd/pkg/v3/cobrautl"
+	"go.etcd.io/etcd/server/v3/storage/backend"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
+)
+
+var (
+	hashKVRevision int64
+)
+
+// NewHashKVCommand returns the cobra command for "hashkv".
+func NewHashKVCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "hashkv <filename>",
+		Short: "Prints the KV history hash of a given file",
+		Args:  cobra.ExactArgs(1),
+		Run:   hashKVCommandFunc,
+	}
+	cmd.Flags().Int64Var(&hashKVRevision, "rev", 0, "maximum revision to hash (default: latest revision)")
+	return cmd
+}
+
+func hashKVCommandFunc(cmd *cobra.Command, args []string) {
+	printer := initPrinterFromCmd(cmd)
+
+	ds, err := calculateHashKV(args[0], hashKVRevision)
+	if err != nil {
+		cobrautl.ExitWithError(cobrautl.ExitError, err)
+	}
+	printer.DBHashKV(ds)
+}
+
+type HashKV struct {
+	Hash            uint32 `json:"hash"`
+	HashRevision    int64  `json:"hashRevision"`
+	CompactRevision int64  `json:"compactRevision"`
+}
+
+func calculateHashKV(dbPath string, rev int64) (HashKV, error) {
+	cfg := backend.DefaultBackendConfig(zap.NewNop())
+	cfg.Path = dbPath
+	b := backend.New(cfg)
+	st := mvcc.NewStore(zap.NewNop(), b, nil, mvcc.StoreConfig{})
+	hst := mvcc.NewHashStorage(zap.NewNop(), st)
+
+	h, _, err := hst.HashByRev(rev)
+	if err != nil {
+		return HashKV{}, err
+	}
+	return HashKV{
+		Hash:            h.Hash,
+		HashRevision:    h.Revision,
+		CompactRevision: h.CompactRevision,
+	}, nil
+}
--- a/etcdutl/etcdutl/printer.go
+++ b/etcdutl/etcdutl/printer.go
@@ -31,6 +31,7 @@ var (
 
 type printer interface {
 	DBStatus(snapshot.Status)
+	DBHashKV(HashKV)
 }
 
 func NewPrinter(printerType string) printer {
@@ -64,6 +65,7 @@ func newPrinterUnsupported(n string) printer {
 }
 
 func (p *printerUnsupported) DBStatus(snapshot.Status) { p.p(nil) }
+func (p *printerUnsupported) DBHashKV(HashKV)          { p.p(nil) }
 
 func makeDBStatusTable(ds snapshot.Status) (hdr []string, rows [][]string) {
 	hdr = []string{"hash", "revision", "total keys", "total size", "version"}
@@ -73,6 +75,16 @@ func makeDBStatusTable(ds snapshot.Status) (hdr []string, rows [][]string) {
 		fmt.Sprint(ds.TotalKey),
 		humanize.Bytes(uint64(ds.TotalSize)),
 		ds.Version,
+	})
+	return hdr, rows
+}
+
+func makeDBHashKVTable(ds HashKV) (hdr []string, rows [][]string) {
+	hdr = []string{"hash", "hash revision", "compact revision"}
+	rows = append(rows, []string{
+		fmt.Sprint(ds.Hash),
+		fmt.Sprint(ds.HashRevision),
+		fmt.Sprint(ds.CompactRevision),
 	})
 	return hdr, rows
 }
--- a/server/storage/mvcc/hash.go
+++ b/server/storage/mvcc/hash.go
@@ -111,7 +111,7 @@ type hashStorage struct {
 	lg     *zap.Logger
 }
 
-func newHashStorage(lg *zap.Logger, s *store) *hashStorage {
+func NewHashStorage(lg *zap.Logger, s *store) HashStorage {
 	return &hashStorage{
 		store: s,
 		lg:    lg,
--- a/etcdutl/README.md
+++ b/etcdutl/README.md
@@ -119,6 +119,44 @@ Prints a line of JSON encoding the database hash, revision, total keys, and size
 +----------+----------+------------+------------+
 ```
 
+### HASHKV [options] \<filename\>
+
+HASHKV prints hash of keys and values up to given revision.
+
+#### Options
+
+- rev -- Revision number. Default is 0 which means the latest revision.
+
+#### Output
+
+##### Simple format
+
+Prints a humanized table of the KV hash, hash revision and compact revision.
+
+##### JSON format
+
+Prints a line of JSON encoding the KV hash, hash revision and compact revision.
+
+#### Examples
+```bash
+./etcdutl hashkv file.db
+# 35c86e9b, 214, 150
+```
+
+```bash
+./etcdutl --write-out=json hashkv file.db
+# {"hash":902327963,"hashRevision":214,"compactRevision":150}
+```
+
+```bash
+./etcdutl --write-out=table hashkv file.db
++----------+---------------+------------------+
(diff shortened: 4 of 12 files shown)


jmhbnz (commented):
Nice work @cenkalti.

Should we also add a quick segment of documentation to the `README.md` at https://github.com/etcd-io/etcd/tree/main/etcdutl?

cenkalti:
@jmhbnz Added.

jmhbnz (approved):
LGTM - Thanks @cenkalti.

Note at some point I think we need to take some of the `README.md` content and put together some overall human friendly documentation for `etcdutl` on the website. Something similar to https://etcd.io/docs/v3.5/dev-guide/interacting_v3. Perhaps named `Interacting with etcd data files`. A separate issue than this pr though.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -110,6 +114,30 @@ type Status struct {
 	Version string `json:"version"`
 }
 
+type HashKV struct {
+	Hash            uint32 `json:"hash"`
+	HashRevision    int64  `json:"hashRevision"`
+	CompactRevision int64  `json:"compactRevision"`
```
chaochn47, line 120: I think `CompactRevision` in the response is noise for end users. I am inclined to remove it just like `etcdctl endpoints hashkv` https://github.com/etcd-io/etcd/blob/main/etcdctl/README.md#endpoint-hashkv

serathius (approved):

etcdutl/snapshot/v3_snapshot.go
cenkalti: @chaochn47 are you suggesting to remove `CompactRevison` from simple and table outputs but keep in JSON output?

chaochn47: Yeah, that matches with my expectation.

serathius: I'm against removing `CompactRevision` it is important part of confirming data corruption and hiding this value would lead to false positives.

etcdutl/etcdutl/printer.go
```diff
@@ -78,6 +80,16 @@ func makeDBStatusTable(ds snapshot.Status) (hdr []string, rows [][]string) {
 	return hdr, rows
 }
 
+func makeDBHashKVTable(ds snapshot.HashKV) (hdr []string, rows [][]string) {
+	hdr = []string{"hash", "hash revision", "compact revision"}
+	rows = append(rows, []string{
+		fmt.Sprintf("%x", ds.Hash),
```
ahrtr, line 86: Let's keep it consistent with the output of `etcdctl endpoint hashkv`
```suggestion
		fmt.Sprint(ds.Hash),
```

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -62,6 +64,19 @@ The items in the lists are hash, revision, total keys, total size.
 	}
 }
 
+func newSnapshotHashKVCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "hashkv <filename>",
+		Short: "Prints the KV history hash of a given file",
+		Long: `When --write-out is set to simple, this command prints out comma-separated status lists for each endpoint.
+The items in the lists are hash, hash revision, compact revision.
```
ahrtr, line 72: etcdutl is executed on a file, and there is no endpoint.
```suggestion
```

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -62,6 +64,19 @@ The items in the lists are hash, revision, total keys, total size.
...
+`,
+		Run: SnapshotHashKVCommandFunc,
+	}
+	cmd.Flags().Int64Var(&hashKVRevision, "rev", 0, "maximum revision to hash (default: all revisions)")
```
ahrtr, line 76:
```suggestion
	cmd.Flags().Int64Var(&hashKVRevision, "rev", 0, "maximum revision to hash (default: latest revision)")
```

FYI. https://github.com/etcd-io/etcd/pull/15997

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -62,6 +64,19 @@ The items in the lists are hash, revision, total keys, total size.
...
 }
 
+func newSnapshotHashKVCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "hashkv <filename>",
+		Short: "Prints the KV history hash of a given file",
+		Long: `When --write-out is set to simple, this command prints out comma-separated status lists for each endpoint.
+The items in the lists are hash, hash revision, compact revision.
+`,
+		Run: SnapshotHashKVCommandFunc,
```
ahrtr, line 74:
```suggestion
		Run: snapshotHashKVCommandFunc,
```

etcdutl/snapshot/v3_snapshot.go
ahrtr: > I'm against removing CompactRevision it is important part of confirming data corruption and hiding this value would lead to false positives.


Agreed. I am even thinking about display both `ScheduledCompactKeyName` and `FinishedCompactKeyName`.  If the latest compaction somehow did not complete, we may get different hashes for different files even they have the same `CompactRevision`  value.  FYI. https://github.com/etcd-io/etcd/issues/15919

ahrtr (changes requested):
Please resolve the review comments.

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -48,6 +49,7 @@ func NewSnapshotCommand() *cobra.Command {
 	}
 	cmd.AddCommand(NewSnapshotRestoreCommand())
 	cmd.AddCommand(newSnapshotStatusCommand())
+	cmd.AddCommand(newSnapshotHashKVCommand())
```
ahrtr, line 52: Probably it makes more sense to add a separate command something like `etcdutl hashkv <filename> [flags]`, because it isn't necessary related to a snapshot.

etcdutl/snapshot/v3_snapshot.go
chaochn47: > I'm against removing CompactRevision it is important part of confirming data corruption and hiding this value would lead to false positives.

It's fair point. But as Benjamin mentioned, compact is an asynchronous operation which does not guarantee `CompactRevision` indicating any revision before it is deleted. 

How about wait `FinishedCompactKeyName` value is consistent with `ScheduledCompactKeyName` after `mvcc.NewStore()`?

ahrtr: > How about wait FinishedCompactKeyName value is consistent with ScheduledCompactKeyName value to guarantee the previous scheduledCompaction is completed after mvcc.NewStore()?


The most useful case to run `etcdutl hashkv` is that etcd instance isn't running anymore. `wait` makes no sense.

chaochn47: > The most useful case to run etcdutl hashkv is that etcd instance isn't running anymore.

Isn't `mvcc.NewStore()` called in `etcdutl hashkv` will resume the scheduled compaction if any?  https://github.com/etcd-io/etcd/blob/1200b1006da69f555ac1959e70bdddf2f1cc9226/server/mvcc/kvstore.go#L396

If we decide to expose compactRevision, sounds like this compaction must be finished.

ahrtr: It isn't good for an offline tool (`etcdutl` in this case) to silently modify the db file. But since it's doing the right thing (compaction), so I am less diligent to against it.

There is race condition in current implementation, the result is uncertain if previous compaction did not finish. If the compacted keys is bigger than batchNum, it will commit the transaction multiple times, then the hashkv may get an intermediate result. https://github.com/etcd-io/etcd/blob/8da2a5bf46cf963373d2841ed90e33560865884c/server/storage/mvcc/kvstore_compaction.go#L83

So proposal:
- Do not silently modify the data. Instead, if `etcdutl` detects any unfinished compaction, then stop the operation, and notify user, and let users to explicitly do it. 
- Add two commands, one for finishing previous unfinished compaction; the other is for calculating the hashkv of the db file (it's what this PR is doing).

chaochn47: > It isn't good for an offline tool (etcdutl in this case) to silently modify the db file.
Oh you are right but anyway it is silently modify the db file in `mvcc.NewStore()` ..

Fail the command explicitly option looks good to me.

jmhbnz:
Hey @cenkalti - We highlighted this older pr during the sig-etcd triage meeting today. It would be great to get this tidied up and merged given we already had partial maintainer approval.

Can you please rebase and address the remaining comments?  Thanks! 🙏🏻

cenkalti:
I'll do it this weekend.

cenkalti:
Rebased. Ready to merge.

etcdutl/etcdutl/hashkv_command.go
```diff
@@ -0,0 +1,79 @@
...
+		Run:   hashKVCommandFunc,
+	}
+	cmd.Flags().Int64Var(&hashKVRevision, "rev", 0, "maximum revision to hash (default: latest revision)")
+	return cmd
+}
+
+func hashKVCommandFunc(cmd *cobra.Command, args []string) {
+	if len(args) != 1 {
+		err := fmt.Errorf("hashkv requires exactly one argument")
+		cobrautl.ExitWithError(cobrautl.ExitBadArgs, err)
```
ahrtr, line 46: Suggest to use `cobra.ExactArgs(1)`, but not a blocker to merge this PR.

Refer to https://github.com/etcd-io/bbolt/blob/f663d4e6043e37fa77c3b554f2692d4d02cc8c95/cmd/bbolt/command_surgery_freelist.go#L31

ahrtr:
We also need to add a changelog item, but it's OK to add it in a separate PR. 

cc @jmhbnz @ivanvc @ArkaSaha30 to take a second look.

server/storage/mvcc/hash.go
```diff
@@ -111,7 +111,7 @@ type hashStorage struct {
 	lg     *zap.Logger
 }
 
-func newHashStorage(lg *zap.Logger, s *store) *hashStorage {
+func NewHashStorage(lg *zap.Logger, s *store) *hashStorage {
```
ivanvc, line 114: This (now) exported function returns an unexported struct. According to golangci-lint (and a linter rule we still need to enable), this is not a good practice. Refer to golang/lint#210.

However, `HashStorage` is already defined as the interface. So, addressing this may require renaming either the interface or the struct :thinking:. This may open a discussion rather than an immediate action item.

etcdutl/etcdutl/hashkv_command.go
```diff
@@ -0,0 +1,79 @@
...
+	printer.DBHashKV(ds)
+}
+
+type HashKV struct {
+	Hash            uint32 `json:"hash"`
+	HashRevision    int64  `json:"hashRevision"`
+	CompactRevision int64  `json:"compactRevision"`
+}
+
+func calculateHashKV(dbPath string, rev int64) (ds HashKV, err error) {
```
ivanvc, line 63: Nit: I don't think this function benefits from defining named return values. `ds` is not used, and `err` is defined in line 70.

ivanvc (commented):
This PR is looking good. I left two minor comments.

etcdutl/etcdutl/hashkv_command.go
cenkalti: Updated.

server/storage/mvcc/hash.go
cenkalti: Updated to return the interface.

etcdutl/etcdutl/hashkv_command.go
cenkalti: Removed return var names.

ahrtr (approved):
LGTM

Thanks.  Please also add a changelog item for 3.6 in a separate PR.

ivanvc (approved):
LGTM. Thanks, @cenkalti.

Issue #15778, opened 2023-04-25
Merge/cleanup Procfiles


serathius:
### What would you like to be added?

Procfiles are nice and all, but why do we need to have 3 in root directory `Procfile`, `Procfile.learner`, `Procfile.v2`.

I propose one of three things:
* Merge/remove `Procfile.learner` and `Procfile.v2` so only `Procfile` is left.
* Move procfiles to `Documentation/contributor-guide` or similar place for contributors.
* Remove after confirming they are not used by any contributor.

### Why is this needed?

Root directory should be minimal to improve readability and avoid overwhelming new contributors.
It's not clear which one of them they should use, nor where it is documented.

sharathsivakumar:
@serathius and @jmhbnz I would like to work on this.

serathius:
Let's discuss which option we should do. I don't use procfile as much so I don't want to pick myself.

sharathsivakumar:
Cool. I am doing a bit of digging around. I will get back to you soon and we can go ahead. In the meantime, do you think it might be a good idea to ask around in the etcd channel if other contributors have any opinions on this as well?

serathius:
Sure, cc @ahrtr @ptabor @jmhbnz @chaochn47 @lavacat

ahrtr:
`Profile` is super useful to launch a local etcd cluster in seconds, let's keep it as it's. 
- Remove `Procfile.v2`, since we only support `zap` for the structure log starting from 3.5. And 3.4 also support zap.
- Append the content of `Profile.learner` to `Profile` as comment.

jmhbnz:
Suggested approach from @ahrtr sounds good to me.  That would cut two files out of root directory which is a great win.

sharathsivakumar:
Based on the inputs provided by everyone, I will make the changes and create a PR.

shivanshu1333:
@jmhbnz this is a small one, I've opened up a [PR](https://github.com/etcd-io/etcd/pull/16040), if you've time, kindly review.
Assign this issue to me if needed.
Thanks!

jmhbnz:
Thanks for your pull request @shivanshu1333, I've reassigned the issue to you.

This issue was originally assigned to another contributor, so in future from a contributor experience perspective my preference would be for you to check in with the assigned contributor first and ask for them to release the issue before raising a pr.

shivanshu1333:
Thanks @jmhbnz and apologies, suggestion noted!

PR #15966, opened 2023-05-26
ci: add fix-yamllint rule in Makefile

tao12345666333:

follow-up #15955 

Two rules have been specified for `yamlfmt` here:

* Automatically add the header `---` to YAML files.
* Preserve line breaks.
    
ref:     https://github.com/google/yamlfmt/blob/main/docs/config-file.md#configuration-1


Makefile
```diff
@@ -132,6 +132,13 @@ verify-genproto:
 verify-goimport:
 	PASSES="goimport" ./scripts/test.sh
 
+.PHONY: update-yamllint
```
serathius, line 135:
```suggestion
.PHONY: fix-yamllint
```

tao12345666333: Thanks! Addressed.

.yamlfmt
```diff
@@ -0,0 +1,4 @@
+formatter:
```
jmhbnz, line 1: Should we move this configuration file out of the root directory to reduce initial clutter?

It's possible to specify the file location via https://github.com/google/yamlfmt/blob/main/docs/command-usage.md#configuration-flags 

It could sit under `/tools` or `/tests` instead perhaps.  Just an idea given we have been [discussing tidying root directory](https://github.com/etcd-io/etcd/issues/15778).

In hindsight we could actually do the same for `yamllint` https://yamllint.readthedocs.io/en/stable/configuration.html.

tao12345666333: Thank you for your suggestion. Using parameters can indeed reduce the content of the root directory.

There are several reasons why I use configuration files:

* These configuration files are hidden by default, and most people will not see them directly;
* The names of these configuration files are very intuitive, and their functions can be understood through their names.


The difference from https://github.com/etcd-io/etcd/issues/15778 is that we expect `yamllint` and `yamlfmt` to be triggered through `make`, so there is no need to touch these configuration files.
But the Procfile file is included in the documentation, and people need to understand its function and significance.



If modifications are needed, I can make adjustments. `yamlfmt` passes parameters through flags, and the configuration file for `yamllint` is located in either the `tools` or `scripts` directory.

Makefile
```diff
@@ -136,6 +136,13 @@ verify-goimport:
 verify-yamllint:
 	yamllint .
 
+.PHONY: fix-yamllint
+fix-yamllint:
+ifeq (, $(shell which yamlfmt))
+	$(shell go install github.com/google/yamlfmt/cmd/yamlfmt@latest)
```
fuweid, line 142: maybe we should pin the version in the tools/mod.

tao12345666333: good catch! thanks

.yamlfmt
jmhbnz: Thanks @tao12345666333, my preference would be to:

1. Move `.yamllint` and `.yamlfmt` to `tools/...`
2. Update `Makefile` as follows:

```diff
verify-yamllint:
-	yamllint .
+      yamllint --config-file tools/.yamllint .
```

```diff
fix-yamllint:
ifeq (, $(shell which yamlfmt))
	$(shell go install github.com/google/yamlfmt/cmd/yamlfmt@latest)
endif
-	yamlfmt .
+      yamlfmt -conf tools/.yamllint .

```

tao12345666333: Addressed!

Makefile
tao12345666333: Addressed

.yamlfmt
serathius: Agree with the proposed direction. @jmhbnz Can you file an issue to consider moving other dotfiles?

serathius: FYI there is also `hack` directory borrowed from kubernetes that serves similar purpose.

Makefile
```diff
@@ -134,7 +134,14 @@ verify-goimport:
 
 .PHONY: verify-yamllint
 verify-yamllint:
-	yamllint .
+	yamllint --config-file tools/.yamllint .
+
+.PHONY: fix-yamllint
+fix-yamllint:
+ifeq (, $(shell which yamlfmt))
+	$(shell go install github.com/google/yamlfmt/cmd/yamlfmt@v0.9.0)
```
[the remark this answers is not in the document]
tao12345666333: Good point, thanks!

fuweid (approved):
LGTM

.yamlfmt
ahrtr: Don't like the idea of putting the config files under `hack`.  

https://github.com/etcd-io/etcd/issues/15987#issuecomment-1571393182

ahrtr:
Overall looks good to me with a minor comment https://github.com/etcd-io/etcd/issues/15987#issuecomment-1571393182.

Thanks @tao12345666333 and @jmhbnz

serathius (approved):

jmhbnz:
Hey @ahrtr - Are we ok to merge this now?  

Noting we have a decision to make under #15987 about where the config files are going to live, however it would be nice to get this makefile fix functionality merged and close down #15955.

ahrtr:
Just as I mentioned in https://github.com/etcd-io/etcd/issues/15987#issuecomment-1571393182, I don't think it's good to put the config files under hack. The utilities & scripts under `hack` are various hacks which are used by developers; while all tools under `tools` are used by etcd itself.  Removing anything under tools will break the etcd (e.g. build, workflow check etc), but removing anything under `hack` will not impact etcd itself at all.

jmhbnz:
> Just as I mentioned in https://github.com/etcd-io/etcd/issues/15987#issuecomment-1571393182, I don't think it's good to put the config files under hack.

Thanks for sharing your perspective re `hack`, note this pr does not propose anything under that directory.

tao12345666333:
So can we merge this PR by moving the config files currently in the tools root directory to the tools/config directory?

I personally don't have a strong preference for these two positions.

ahrtr:
I am OK as long as the config files are under `tools` instead of `hack`, thanks.

ahrtr (approved):

PR #15985, opened 2023-05-31
Check ScheduledCompactKeyName and FinishedCompactKeyName before writing hash

### LINKED ISSUE #15919: Incorrect hash when resuming scheduled compaction after etcd restarts

CaojiamingAlan:
Fix #15919 
Commit 1: add an e2e test to reproduce the problem. This test won't pass without the fix.
Commit 2: check ScheduledCompactKeyName and FinishedCompactKeyName
before writing hash to hashstore. If they do not match, then it means this compaction has once been interrupted and its hash value is invalid. In such cases, we won't write the hash values to the hashstore, and avoids the incorrect corruption alarm.

### DIFF
--- a/server/storage/mvcc/kvstore.go
+++ b/server/storage/mvcc/kvstore.go
@@ -225,7 +225,17 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
 	return nil, compactMainRev, nil
 }
 
-func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-chan struct{}, error) {
+// checkPrevCompactionCompleted checks whether the previous scheduled compaction is completed.
+func (s *store) checkPrevCompactionCompleted() bool {
+	tx := s.b.ReadTx()
+	tx.Lock()
+	defer tx.Unlock()
+	scheduledCompact, scheduledCompactFound := UnsafeReadScheduledCompact(tx)
+	finishedCompact, finishedCompactFound := UnsafeReadFinishedCompact(tx)
+	return scheduledCompact == finishedCompact && scheduledCompactFound == finishedCompactFound
+}
+
+func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64, prevCompactionCompleted bool) (<-chan struct{}, error) {
 	ch := make(chan struct{})
 	j := schedule.NewJob("kvstore_compact", func(ctx context.Context) {
 		if ctx.Err() != nil {
@@ -238,7 +248,13 @@ func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-ch
 			s.compactBarrier(context.TODO(), ch)
 			return
 		}
-		s.hashes.Store(hash)
+		// Only store the hash value if the previous hash is completed, i.e. this compaction
+		// hashes every revision from last compaction. For more details, see #15919.
+		if prevCompactionCompleted {
+			s.hashes.Store(hash)
+		} else {
+			s.lg.Info("previous compaction was interrupted, skip storing compaction hash value")
+		}
 		close(ch)
 	})
 
@@ -248,17 +264,18 @@ func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-ch
 }
 
 func (s *store) compactLockfree(rev int64) (<-chan struct{}, error) {
+	prevCompactionCompleted := s.checkPrevCompactionCompleted()
 	ch, prevCompactRev, err := s.updateCompactRev(rev)
 	if err != nil {
 		return ch, err
 	}
 
-	return s.compact(traceutil.TODO(), rev, prevCompactRev)
+	return s.compact(traceutil.TODO(), rev, prevCompactRev, prevCompactionCompleted)
 }
 
 func (s *store) Compact(trace *traceutil.Trace, rev int64) (<-chan struct{}, error) {
 	s.mu.Lock()
-
+	prevCompactionCompleted := s.checkPrevCompactionCompleted()
 	ch, prevCompactRev, err := s.updateCompactRev(rev)
 	trace.Step("check and update compact revision")
 	if err != nil {
@@ -267,7 +284,7 @@ func (s *store) Compact(trace *traceutil.Trace, rev int64) (<-chan struct{}, err
 	}
 	s.mu.Unlock()
 
-	return s.compact(trace, rev, prevCompactRev)
+	return s.compact(trace, rev, prevCompactRev, prevCompactionCompleted)
 }
 
 func (s *store) Commit() {
--- a/tests/e2e/corrupt_test.go
+++ b/tests/e2e/corrupt_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
@@ -192,4 +193,82 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
 	alarmResponse, err := cc.AlarmList(ctx)
 	assert.NoError(t, err, "error on alarm list")
 	assert.Equal(t, []*etcdserverpb.AlarmMember{{Alarm: etcdserverpb.AlarmType_CORRUPT, MemberID: memberID}}, alarmResponse.Alarms)
+}
+
+func TestCompactHashCheckDetectCorruptionInterrupt(t *testing.T) {
+	checkTime := time.Second
+	e2e.BeforeTest(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	slowCompactionNodeIndex := 1
+
+	// Start a new cluster, with compact hash check enabled.
+	t.Log("creating a new cluster with 3 nodes...")
+
+	dataDirPath := t.TempDir()
+	cfg := e2e.NewConfig(
+		e2e.WithKeepDataDir(true),
+		e2e.WithCompactHashCheckEnabled(true),
+		e2e.WithCompactHashCheckTime(checkTime),
+		e2e.WithClusterSize(3),
+		e2e.WithDataDirPath(dataDirPath),
+		e2e.WithLogLevel("info"),
+	)
+	epc, err := e2e.InitEtcdProcessCluster(t, cfg)
+	require.NoError(t, err)
+
+	// Assign a node a very slow compaction speed, so that its compaction can be interrupted.
+	err = epc.UpdateProcOptions(slowCompactionNodeIndex, t,
+		e2e.WithCompactionBatchLimit(1),
+		e2e.WithCompactionSleepInterval(1*time.Hour),
+	)
+	require.NoError(t, err)
+
+	epc, err = e2e.StartEtcdProcessCluster(ctx, epc, cfg)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	})
+
+	// Put 10 identical keys to the cluster, so that the compaction will drop some stale values.
+	t.Log("putting 10 values to the identical key...")
+	cc, err := e2e.NewEtcdctl(epc.Cfg.Client, epc.EndpointsGRPC())
+	require.NoError(t, err)
+	for i := 0; i < 10; i++ {
+		err := cc.Put(ctx, "key", fmt.Sprint(i), config.PutOptions{})
+		require.NoError(t, err, "error on put")
+	}
+
+	t.Log("compaction started...")
+	_, err = cc.Compact(ctx, 5, config.CompactOption{})
+
+	err = epc.Procs[slowCompactionNodeIndex].Close()
+	require.NoError(t, err)
+
+	err = epc.UpdateProcOptions(slowCompactionNodeIndex, t)
+	require.NoError(t, err)
+
+	t.Logf("restart proc %d to interrupt its compaction...", slowCompactionNodeIndex)
+	err = epc.Procs[slowCompactionNodeIndex].Restart(ctx)
+
+	// Wait until the node finished compaction and the leader finished compaction hash check
+	_, err = epc.Procs[slowCompactionNodeIndex].Logs().ExpectWithContext(ctx, "finished scheduled compaction")
+	require.NoError(t, err, "can't get log indicating finished scheduled compaction")
+
+	leaderIndex := epc.WaitLeader(t)
+	_, err = epc.Procs[leaderIndex].Logs().ExpectWithContext(ctx, "finished compaction hash check")
+	require.NoError(t, err, "can't get log indicating finished compaction hash check")
+
+	alarmResponse, err := cc.AlarmList(ctx)
+	require.NoError(t, err, "error on alarm list")
+	for _, alarm := range alarmResponse.Alarms {
+		if alarm.Alarm == etcdserverpb.AlarmType_CORRUPT {
+			t.Fatal("there should be no corruption after resuming the compaction, but corruption detected")
+		}
+	}
+	t.Log("no corruption detected.")
 }
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -182,6 +182,7 @@ type EtcdProcessClusterConfig struct {
 	CompactHashCheckTime    time.Duration
 	GoFailEnabled           bool
 	CompactionBatchLimit    int
+	CompactionSleepInterval time.Duration
 
 	WarningUnaryRequestDuration             time.Duration
 	ExperimentalWarningUnaryRequestDuration time.Duration
@@ -339,6 +340,10 @@ func WithExperimentalWarningUnaryRequestDuration(time time.Duration) EPClusterOp
 
 func WithCompactionBatchLimit(limit int) EPClusterOption {
 	return func(c *EtcdProcessClusterConfig) { c.CompactionBatchLimit = limit }
+}
+
+func WithCompactionSleepInterval(time time.Duration) EPClusterOption {
+	return func(c *EtcdProcessClusterConfig) { c.CompactionSleepInterval = time }
 }
 
 func WithWatchProcessNotifyInterval(interval time.Duration) EPClusterOption {
@@ -582,6 +587,9 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 	if cfg.CompactionBatchLimit != 0 {
 		args = append(args, "--experimental-compaction-batch-limit", fmt.Sprintf("%d", cfg.CompactionBatchLimit))
 	}
+	if cfg.CompactionSleepInterval != 0 {
+		args = append(args, "--experimental-compaction-sleep-interval", cfg.CompactionSleepInterval.String())
+	}
 	if cfg.WarningUnaryRequestDuration != 0 {
 		args = append(args, "--warning-unary-request-duration", cfg.WarningUnaryRequestDuration.String())
 	}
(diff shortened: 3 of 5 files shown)


server/storage/mvcc/kvstore.go
```diff
@@ -238,7 +247,11 @@ func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-ch
 			s.compactBarrier(context.TODO(), ch)
 			return
 		}
-		s.hashes.Store(hash)
+		// Only store the hash value is the previous hash is completed, i.e. this compaction
```
jmhbnz, line 250: nit: looks like a small typo?

```suggestion
		// Only store the hash value if the previous hash is completed, i.e. this compaction
```

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,71 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+	)
+
+	t.Cleanup(func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	})
+
+	// Add a member to the cluster, assign it a very slow compaction speed, so that its compaction can be interrupted.
+	//
```
jmhbnz, line 219: Should this empty comment line be removed?

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,71 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("could not start etcd process cluster (%v)", err)
+	}
+
+	// Put 10 identical keys to the cluster, so that the compaction will drop some stale values.
+	t.Log("putting 10 values to the identical key...")
+	cc, err := e2e.NewEtcdctl(epc.Cfg.Client, epc.EndpointsGRPC())
+	assert.NoError(t, err)
+	for i := 0; i < 10; i++ {
+		//use the same key
```
jmhbnz, line 234: nit: spacing

```suggestion
		// Use the same key
```

jmhbnz (commented):
Thanks for your work on this @CaojiamingAlan!

Still getting my head around the changes and linked issue, a few tiny nitpicks to consider after an initial read through.

CaojiamingAlan:
Still need to make some changes to the tests using a fake store.

server/storage/mvcc/kvstore.go
```diff
@@ -225,7 +225,16 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
 	return nil, compactMainRev, nil
 }
 
-func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-chan struct{}, error) {
+// checkPrevCompactionCompleted checks whether the previous scheduled compaction is completed.
+func (s *store) checkPrevCompactionCompleted() bool {
+	s.b.ReadTx().Lock()
+	defer s.b.ReadTx().Unlock()
+	scheduledCompact, scheduledCompactFound := UnsafeReadScheduledCompact(s.b.ReadTx())
+	finishedCompact, finishedCompactFound := UnsafeReadFinishedCompact(s.b.ReadTx())
```
serathius, line 233:
```suggestion
	tx := s.b.ReadTx()
	defer tx.Unlock()
	scheduledCompact, scheduledCompactFound := UnsafeReadScheduledCompact(tx)
	finishedCompact, finishedCompactFound := UnsafeReadFinishedCompact(tx)
```

server/storage/mvcc/store.go
```diff
@@ -19,15 +19,7 @@ import (
...
-func UnsafeReadFinishedCompact(tx backend.ReadTx) (finishedComact int64, found bool) {
-	_, finishedCompactBytes := tx.UnsafeRange(schema.Meta, schema.FinishedCompactKeyName, nil, 0)
-	if len(finishedCompactBytes) != 0 {
-		return bytesToRev(finishedCompactBytes[0]).main, true
-	}
-	return 0, false
-}
-
-func UnsafeReadScheduledCompact(tx backend.ReadTx) (scheduledComact int64, found bool) {
+func UnsafeReadScheduledCompact(tx backend.ReadTx) (scheduledCompact int64, found bool) {
```
serathius, line 22: Please don't reorder the functions, it mudies the git history.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,69 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+		e2e.WithClusterSize(1),
+	)
+
+	t.Cleanup(func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	})
+
+	// Add a member to the cluster, assign it a very slow compaction speed, so that its compaction can be interrupted.
```
serathius, line 218: Why not create 3 node cluster? I would prefer to avoid reconfiguring member flags if possible.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,69 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+		assert.NoError(t, err, "error on put")
+	}
+
+	// HACK: manually change the --experimental-compaction-sleep-interval to a small value,
+	// so that the compaction can finish.
+	// todo: add disable method to robustness/failpoint interface
+	t.Log("compaction started...")
+	_, err = cc.Compact(ctx, 5, config.CompactOption{})
+	for i, arg := range epc.Procs[1].Config().Args {
+		if arg == "--experimental-compaction-sleep-interval" {
```
serathius, line 243: Please avoid using args modification. It's not readable.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,69 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+			epc.Procs[1].Config().Args[i+1] = "1ms"
+		}
+	}
+
+	t.Log("restart proc 1 to interrupt its compaction...")
+	err = epc.Procs[1].Restart(ctx)
+
+	t.Log("sleeping 5 sec to let nodes do compact hash check...")
+	time.Sleep(5 * time.Second)
```
serathius, line 252: Can we wait for etcd to write a log that hash check completed instead of waiting arbitrary 5s?

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +194,106 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("can't get log indicating finished scheduled compaction")
+	}
+
+	leaderIndex := epc.WaitLeader(t)
+	_, err = epc.Procs[leaderIndex].Logs().ExpectFunc(ctx, func(s string) bool {
+		// NB: there are several possible branches in CompactHashCheck(),
+		// need to check each of them to confirm the check is finished,
+		// otherwise this ExpectFunc would be blocked.
+		// todo log "finished compaction hash check" on every possible branch
```
jmhbnz, line 274: nit: Write as `// TODO:` to align with etcd standard way of writing todo comments in case someone is searching codebase in future to identify.

```diff
-// todo log "finished compaction hash check" on every possible branch
+// TODO: log "finished compaction hash check" on every possible branch
```

serathius:
Please mark conversations as resolved if you have addressed the comment.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +194,106 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("can't get log indicating finished scheduled compaction")
+	}
+
+	leaderIndex := epc.WaitLeader(t)
+	_, err = epc.Procs[leaderIndex].Logs().ExpectFunc(ctx, func(s string) bool {
+		// NB: there are several possible branches in CompactHashCheck(),
+		// need to check each of them to confirm the check is finished,
+		// otherwise this ExpectFunc would be blocked.
+		// TODO: log "finished compaction hash check" on every possible branch
```
serathius, line 274: Could you implement the TODO?

tests/framework/e2e/cluster.go
```diff
@@ -813,6 +821,33 @@ func (epc *EtcdProcessCluster) StartNewProc(ctx context.Context, cfg *EtcdProces
...
+	epc.Cfg.SetInitialOrDiscovery(serverCfg, initialCluster, "new")
+
+	proc, err := NewEtcdProcess(serverCfg)
+	if err != nil {
+		return err
+	}
+	err = epc.Procs[i].Close()
+	if err != nil {
+		return err
+	}
```
serathius, line 846: UpdateProcOptions function stopping a process is a little unexpected for me. I know that to override the `epc.Procs[i] = proc` you need to stop old process, but the approach here is not very natural. 

What about `UpdateProcOptions` expecting that `epc.Procs[i]` is already stopped and assert that? Before calling `UpdateProcOptions` test should close the process.

jmhbnz (approved):
LGTM - Thanks @CaojiamingAlan 👍🏻

serathius:
cc @ahrtr

tests/framework/e2e/etcd_process.go
```diff
@@ -62,6 +62,7 @@ type EtcdProcess interface {
 
 type LogsExpect interface {
 	ExpectWithContext(context.Context, string) (string, error)
+	ExpectFunc(ctx context.Context, f func(string) bool) (string, error)
```
ahrtr, line 65: Why add `ExpectFunc`?

CaojiamingAlan: Because once I expected multiple strings(different branches may have different logs). But after adding the logs it is not needed. However I think it provides more flexibility and is something good to have so I didn't delete it.

ahrtr: It makes no sense to add something which is not used at all.

Please remove it in this PR, and If you really think it's useful, let's discuss it separately.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,95 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+	_, err = epc.Procs[slowCompactionNodeIndex].Logs().ExpectWithContext(ctx, "finished scheduled compaction")
+	if err != nil {
+		t.Fatalf("can't get log indicating finished scheduled compaction")
+	}
+
+	leaderIndex := epc.WaitLeader(t)
+	_, err = epc.Procs[leaderIndex].Logs().ExpectWithContext(ctx, "finished compaction hash check")
+	if err != nil {
+		t.Fatalf("can't get log indicating finished compaction hash check")
+	}
```
ahrtr, line 277: same as above

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,95 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+
+	alarmResponse, err := cc.AlarmList(ctx)
+	assert.NoError(t, err, "error on alarm list")
+	for _, alarm := range alarmResponse.Alarms {
+		if alarm.Alarm == etcdserverpb.AlarmType_CORRUPT {
+			t.Fatalf("there should be no corruption after resuming the compaction, but corruption detected")
```
ahrtr, line 283: same as above

ahrtr (commented):
Overall look good to me, with a couple of minor comments.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +193,95 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+		e2e.WithCompactHashCheckEnabled(true),
+		e2e.WithCompactHashCheckTime(checkTime),
+		e2e.WithClusterSize(3),
+		e2e.WithDataDirPath(dataDirPath),
+		e2e.WithLogLevel("info"),
+	)
+	epc, err := e2e.InitEtcdProcessCluster(t, cfg)
+	if err != nil {
+		t.Fatalf("could not init etcd process cluster (%v)", err)
+	}
```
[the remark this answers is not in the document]
serathius: Could we make sure this suggestion is obvious to reviewers next time. Maybe we should have a test writing guide for contributors? Do you know what percent of files are using `require.NoError` vs `if err != nil { t.Fatal(...)}`?

cc @jmhbnz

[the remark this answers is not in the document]
jmhbnz: For a rough estimation (inside `tests/`):

```bash
find | grep ".go" | xargs grep -il "require.noerror" | wc -l
24

find | grep ".go" | xargs grep -Pzol 'if err != nil.*(\n|.)*t.Fatal' | wc -l
107
```

Obviously just a quick ball park but looks like only about 25% of test files are using `require.NoError` currently.

Agree we could highlight this as something important for reviewers to call out. If folks are on board with it I'm happy to make an issue so we don't forget.

tests/e2e/corrupt_test.go
```diff
@@ -193,3 +195,81 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+func TestCompactHashCheckDetectCorruptionInterrupt(t *testing.T) {
+	checkTime := time.Second
+	e2e.BeforeTest(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	slowCompactionNodeIndex := 1
+
+	// Start a new cluster, with compact hash check enabled.
+	t.Log("creating a new cluster with 1 node...")
```
chaochn47, line 208: 3 nodes.

chaochn47 (approved):
LGTM (non-binding)

ahrtr (approved):
LGTM

Thanks @CaojiamingAlan

tests/e2e/corrupt_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
```
ahrtr, line 24: remove the blank line?

ahrtr:
@CaojiamingAlan could you backport this PR to 3.5?

CaojiamingAlan:
@ahrtr Sure, does it mean proposing a PR to merge this change to the release-3.5 branch?

ahrtr:
> @ahrtr Sure, does it mean proposing a PR to merge this change to the release-3.5 branch?

Yes. Please note that `release-3.5` has big difference with `main` branch, so please try not to cherry-pick the commit.

Issue #15986, opened 2023-05-31
Etcd nightly arm64 e2e common tests state they have been cached


jmhbnz:
### What happened?

Sub issue under https://github.com/etcd-io/etcd/issues/15951.

We recently moved to container based nightly `arm64` workflow runs which appear to be completing successfully however the e2e `common` suite is reporting unexpected caching, refer: https://github.com/etcd-io/etcd/actions/runs/5116632303/jobs/9198983536 

```bash
'e2e' started at Tue May 30 01:34:05 UTC 2023
% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' 'go.etcd.io/etcd/tests/v3/e2e' '-timeout=30m')
ok  	go.etcd.io/etcd/tests/v3/e2e	529.611s
% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' 'go.etcd.io/etcd/tests/v3/common' '--tags=e2e' '-timeout=30m')
ok  	go.etcd.io/etcd/tests/v3/common	(cached)
'e2e' PASSED and completed at Tue May 30 01:42:58 UTC 2023
SUCCESS
```

### What did you expect to happen?

Caching between test runs is not expected. We need to investigate if this will lead to issues and raise a fix if required.

### How can we reproduce it (as minimally and precisely as possible)?

1. Run an e2e arm64 workflow job either on demand or via standard nightly schedule.
2. Open job logs for the `Run set -euo pipefail` step.
3. Review e2e test result summary showing cache indicator at end of test ok line.

jmhbnz:
This is an example of the docker command being run when a job starts:

```bash
[...]

PR #15989, opened 2023-05-31
Clean golang cache for arm64 ci workflows

jmhbnz:
We recently moved to container based nightly arm64 workflow runs which appear to be completing successfully however the e2e common suite is reporting unexpected caching, refer: https://github.com/etcd-io/etcd/actions/runs/5116632303/jobs/9198983536

We have a hypothesis that we can prevent this behavior with `cache: false` on the `setup-go` action, refer discussion in #15986.


ahrtr:
After second thought, I think the better choice is to only disable cache for golang test by running `go clean -testcache` before each test suite. There is no reason not to reuse the setup-go cache, as long as there is no code change.

jmhbnz:
> After second thought, I think the better choice is to only disable cache for golang test by running `go clean -testcache` before each test suite. There is no reason not to reuse the setup-go cache, as long as there is no code change.

This also crossed my mind.  I'm going to hit pause on any more work here just to see if @serathius or anyone else has a preference on which cache prevention option to use (for the record I'm happy with either approach as both address the issue and I think overall performance of one versus the other will be minimal).

fuweid:
No cache looks good to me (8s is ok). or we can force to run test case with ` GO_TEST_FLAGS="-count=1"` with cache.

serathius:
Caching go mod 👍
Caching go build 👍
Caching go test 👎

jmhbnz:
Thanks for the feedback everyone - I've switched to the more targeted approach of just doing the `go clean -testcache` before each suite runs.

I've run multiple workflow runs to verify this is working as expected and we no longer see any cache for the common e2e tests as we were historically. Refer latest run https://github.com/etcd-io/etcd/actions/runs/5143019041/jobs/9257510705?pr=15989.

fuweid (approved):
LGTM

ahrtr (approved):
LGTM

Thanks @jmhbnz 

Please consider to disable golang test cache for all workflows as a followup PR

serathius (approved):

chaochn47 (approved):
LGTM

jmhbnz:
> Please consider to disable golang test cache for all workflows as a followup PR

Defined follow-up task here as a good first issue https://github.com/etcd-io/etcd/issues/15995

Issue #15878, opened 2023-05-11
Downgrade support from 3.5 to 3.4


logicalhan:
### What would you like to be added?

I would like to be able to safely downgrade from 3.5 to 3.4, and then safely reupgrade back to 3.5.

### Why is this needed?

Given the vast number of data correctness issues we've unearthed in etcd 3.5 (many of them fixed by @ahrtr and @serathius), I have personal reservations about upgrading my k8s clusters to use 3.5. If there was a working rollback strategy (tested of course, as well), then I would be much more inclined to update my etcds to a more recent version.

serathius:
I think it could be easily added to `etcdutl migrate` command allowing for safe offline downgrade and upgrade operations.
Code: https://github.com/etcd-io/etcd/blob/main/etcdutl/etcdutl/migrate_command.go 

This would also help with https://github.com/kubernetes/kubernetes/issues/117906 and cleanup of [kubernetes migrate script for etcd](https://github.com/kubernetes/kubernetes/tree/master/cluster/images/etcd).

lavacat:
Please assign this to me, we already have a minimal internal patch to address this. In current form - it's a 3.4 patch that allows 3.4 to be deployed within 3.5 cluster to avoid downtime and perform a rolling downgrade.
It's done by hacking version and removing `confState` and `term` keys. 
But it would be great to make it part of `migrate` and add more testing around it.

serathius:
Just a note, we support for rolling update is out of scope for now. Let's start with migrate script

lavacat:
Quick update - trying to get POC to work. The idea is to run `etcdutl migrate --data-dir data-3.5 --target-version 3.4` and get a data dir that etcd 3.4 can be started with.
My understanding is that currently `migrate` only updates `MetaStorageVersionName` key that was added since 3.6. But it won't update `ClusterClusterVersionKeyName` and `version` in v2store. 

At the moment, running into 
```
etcdserver/membership: cluster cannot be downgraded (current version: 3.4.26 is lower than determined cluster version: 3.5).
```
because of v2store `version`.

lavacat:
[...]

PR #15994, opened 2023-06-01
server: add fields to migrate from 3.5 to 3.4

lavacat:
Tested migrate from 3.5 to 3.4 manually. 
```
$ ./bin/etcdutl migrate --data-dir <data-dir> --target-version 3.4
2023-06-01T10:47:51-07:00	info	schema/migration.go:65	updated storage version	{"new-storage-version": "3.4.0"}

$ bbolt get <data-dir>/member/snap/db meta confState
key not found
$ bbolt get <data-dir>/member/snap/db meta term
key not found
```

fixes: https://github.com/etcd-io/etcd/issues/15878


server/storage/schema/schema.go
```diff
@@ -131,8 +131,19 @@ var (
 		version.V3_6: {
 			addNewField(Meta, MetaStorageVersionName, emptyStorageVersion),
 		},
+		version.V3_5: {
+			// 'migrate' command from v3.4 to v3.5 isn't supported and will fail with `missing confstate information`
```
serathius, line 135: Not sure I understand this comment. Can you elaborate?

server/storage/schema/schema_test.go
```diff
@@ -192,12 +192,10 @@ func TestMigrate(t *testing.T) {
...
-			name:           "Downgrading v3.5 to v3.4 is not supported as schema was introduced in v3.6",
-			version:        version.V3_5,
-			targetVersion:  version.V3_4,
-			expectVersion:  nil,
-			expectError:    true,
-			expectErrorMsg: `cannot create migration plan: version "3.5.0" is not supported`,
+			name:          "Downgrading v3.5 to v3.4 is supported",
+			version:       version.V3_5,
+			targetVersion: version.V3_4,
+			expectVersion: nil,
```
serathius, line 198: Hmm this test was not written with assumption storage version is empty. Maybe we could separate validation of storage version field (one read from UnsafeReadStorageVersion) and storage version it's compatible with (one inferred by fields present like `UnsafeDetectSchemaVersion`) 

Can you rename `expectVersion` to `expectedStorageVersion` and comment that v3.4 didn't have storage version. And update `UnsafeDetectSchemaVersion` to handle v3.5 use it to test for compatibility?

serathius:
The proposed approach looks good, please add e2e upgrade downgrade tests. We need v3.5 -> v3.4 -> v3.5 if we want to officially support this in migrate. 

Note; current e2e framework supports downloading only the last etcd minor version. You might need to rewrite some e2e code to support downloading multiple old releases.

cc @ahrtr @ptabor to get feedback about adding downgrade support.

server/storage/schema/schema.go
```diff
@@ -131,8 +131,19 @@ var (
...
+			// these fields are added for v3.5 -> v3.4 downgrade only
+			//
+			// note about emptyValue:
+			// even if these fields are used for upgrade, it's safe to set emptyValue
+			// UnsafeReadConsistentIndex and UnsafeConfStateFromBackend checks for zero-length values
+			addNewField(Meta, MetaTermKeyName, emptyValue),
+			addNewField(Meta, MetaConfStateName, emptyValue),
```
ahrtr, line 142: Also add `ClusterClusterVersionKeyName` and `ClusterDowngradeKeyName`?  https://github.com/etcd-io/etcd/blob/8cb3fab8d9ec6d4a395d0f46d85049b25cb90938/server/storage/schema/bucket.go#L78-L81

ahrtr:
> We need v3.5 -> v3.4 -> v3.5

Good point. We need e2e test for v3.6 -> v3.5 -> v3.6 as well.

serathius:
> Good point. We need e2e test for v3.6 -> v3.5 -> v3.6 as well.

I think the current downgrade test already does this. It tests current main vs last release which now is v3.5. I would like to extend the current e2e release tests to test last 2 releases as that's our support window.

server/storage/schema/bucket.go
```diff
@@ -69,16 +69,16 @@ func (b bucket) IsSafeRangeBucket() bool { return b.safeRangeBucket }
...
 	// Pre v3.5
-	ScheduledCompactKeyName    = []byte("scheduledCompactRev")
-	FinishedCompactKeyName     = []byte("finishedCompactRev")
-	MetaConsistentIndexKeyName = []byte("consistent_index")
-	AuthEnabledKeyName         = []byte("authEnabled")
-	AuthRevisionKeyName        = []byte("authRevision")
-	// Since v3.5
-	MetaTermKeyName              = []byte("term")
-	MetaConfStateName            = []byte("confState")
+	ScheduledCompactKeyName      = []byte("scheduledCompactRev")
```
serathius, line 72: Can you create section `Since v3.4` to match the schema?

server/storage/schema/confstate.go
```diff
@@ -49,6 +49,10 @@ func UnsafeConfStateFromBackend(lg *zap.Logger, tx backend.ReadTx) *raftpb.ConfS
 			zap.Int("number-of-key", len(keys)),
 		)
 	}
+	if len(vals[0]) == 0 {
```
serathius, line 52: That's not a correct change, we cannot depend on changing server logic for implementing downgrades.

server/storage/schema/schema.go
serathius: Those fields were added in v3.4 not v3.5

server/storage/schema/schema.go
```diff
@@ -131,8 +134,16 @@ var (
...
+			// UnsafeReadConsistentIndex, UnsafeConfStateFromBackend, DowngradeInfoFromBackend checks for zero-length values
+			addNewField(Meta, MetaTermKeyName, emptyValue),
+			addNewField(Meta, MetaConfStateName, emptyValue),
+			addNewField(Cluster, ClusterDowngradeKeyName, emptyValue),
+		},
 	}
 	// emptyStorageVersion is used for v3.6 Step for the first time, in all other version StoragetVersion should be set by migrator.
 	// Adding a addNewField for StorageVersion we can reuse logic to remove it when downgrading to v3.5
 	emptyStorageVersion = []byte("")
+	emptyValue          = []byte("")
```
serathius, line 148: Think that `emptyValue` for v3.4 fields should behave the same way as `emptyStorageVersion` for v3.5

server/storage/schema/schema_test.go
```diff
@@ -187,17 +189,23 @@ func TestMigrate(t *testing.T) {
...
-			name:           "Downgrading v3.5 to v3.4 is not supported as schema was introduced in v3.6",
-			version:        version.V3_5,
-			targetVersion:  version.V3_4,
-			expectVersion:  nil,
-			expectError:    true,
-			expectErrorMsg: `cannot create migration plan: version "3.5.0" is not supported`,
+			name:          "Downgrading v3.5 to v3.4 is supported",
+			version:       version.V3_5,
+			targetVersion: version.V3_4,
+			// note: 3.4 doesn't have storageVersion, this field was added in 3.6
```
serathius, line 200: That's right, could we however also test that `DetectSchemaVersion` will properly recognize v3.4?

server/storage/schema/schema_test.go
```diff
@@ -187,17 +189,23 @@ func TestMigrate(t *testing.T) {
...
+			expectedStorageVersion: nil,
+		},
+		{
+			name:          `Upgrading v3.4 to v3.5 should succeed`,
+			version:       version.V3_4,
+			targetVersion: version.V3_5,
+			// note: 3.5 doesn't have storageVersion, this field was added in 3.6
```
serathius, line 207: Ditto

server/storage/schema/membership_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	mbe := NewMembershipBackend(lg, be)
+
+	mbe.MustCreateBackendBuckets()
+	mbe.be.ForceCommit()
+	assert.Nil(t, mbe.DowngradeInfoFromBackend())
+
+	mbe.be.BatchTx().UnsafePut(Cluster, ClusterDowngradeKeyName, []byte(""))
+	mbe.be.ForceCommit()
+	assert.Nil(t, mbe.DowngradeInfoFromBackend())
```
serathius, line 39: Nice

server/storage/schema/schema_test.go
lavacat: I've added this, but had to add another case to `UnsafeDetectSchemaVersion`, otherwise this won't work for `Upgrading v3.4 to v3.5 should succeed`

server/storage/schema/bucket.go
lavacat: I've searched release-3.3 branch and compared to 3.4 and I don't think any new fields were added

lavacat:
@serathius main change since last review is that I had to add better defaults for fields to make sure they can be read on upgrade.

server/storage/schema/bucket.go
serathius: Thanks.

server/storage/schema/membership_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	mbe := NewMembershipBackend(lg, be)
+
+	mbe.MustCreateBackendBuckets()
+	mbe.be.ForceCommit()
+	assert.Nil(t, mbe.DowngradeInfoFromBackend())
+
+	dinfo := &serverversion.DowngradeInfo{Enabled: true, TargetVersion: version.V3_5.String()}
+	mbe.MustSaveDowngradeToBackend(dinfo)
+	mbe.be.ForceCommit()
```
serathius, line 39: ForceCommit is only useful if you want to dump data on disk and read it again. With only one `NewDefaultTmpBackend` there is no need for it.

server/storage/schema/schema.go
```diff
@@ -96,12 +98,18 @@ func UnsafeDetectSchemaVersion(lg *zap.Logger, tx backend.ReadTx) (v semver.Vers
...
 	_, term := UnsafeReadConsistentIndex(tx)
-	if term == 0 {
+	if confstate == nil && term == 0 {
+		// if both confstate and term are missing, assume it's v3.4
+		return version.V3_4, nil
+	} else if confstate == nil {
+		return v, fmt.Errorf("missing confstate information")
+	} else if len(confstate.Voters) == 0 && term == 0 {
+		// if confstate is empty and term is missing, assume it's v3.5 that was migrated from v3.4 and never started
+		return version.V3_5, nil
```
serathius, line 111: Don't understand this case. I would prefer to avoid supporting `etcdutl migrate` from v3.4 -> v3.5.

server/storage/schema/schema.go
```diff
@@ -129,10 +137,21 @@ var (
...
 		version.V3_6: {
-			addNewField(Meta, MetaStorageVersionName, emptyStorageVersion),
+			// emptyValue is used for v3.6 Step for the first time, in all other version StoragetVersion should be set by migrator.
+			addNewField(Meta, MetaStorageVersionName, emptyValue),
+		},
+		version.V3_5: {
+			// UnsafeReadConsistentIndex checks for zero-length values
+			addNewField(Meta, MetaTermKeyName, emptyValue),
+			// UnsafeConfStateFromBackend will fail on []byte(""), use empty struct as default
+			// note: UnsafeConfStateFromBackend is never used in 3.5, in 3.6 it's only used in `migrate`
```
serathius, line 147: Don't understand this note. Can you elaborate?

server/storage/schema/schema_test.go
```diff
@@ -221,8 +228,18 @@ func TestMigrate(t *testing.T) {
...
+
+			storeV := UnsafeReadStorageVersion(b.BatchTx())
+			assert.Equal(t, tc.expectedStorageVersion, storeV)
+
+			if !tc.expectError {
+				v, err := DetectSchemaVersion(lg, b.ReadTx())
+				if err != nil {
+					t.Errorf("Migrate(lg, tx, %q) failed DetectSchemaVersion: %q", tc.targetVersion, err)
+				}
+				assert.Equal(t, tc.targetVersion, v)
```
serathius, line 241: Can you add a separate field `expectSchemaVersion`?

server/storage/schema/schema_test.go
serathius: I'm little worried about implementing `etcdutl migrate` from v3.4 -> v3.5.

serathius:
Looks almost done. Would like to also have a offline downgrade test. Basically start etcd v3.5, do some requests, use etcdutl do downgrade it, do couple of requests, upgrade it using etcdutl, do couple of requests. Do you think you can add it?

lavacat:
> Would like to also have a offline downgrade test. Basically start etcd v3.5, do some requests, use etcdutl do downgrade it, do couple of requests, upgrade it using etcdutl, do couple of requests. Do you think you can add it?

Yes, I was thinking about adding `start` after migrate to TestEtctlutlMigrate test. I'll give it a go, if it makes TestEtctlutlMigrate unreadable, I'll move it to a separate test.

server/storage/schema/schema.go
lavacat: I was wondering if setting empty struct as default for `MetaConfStateName` will cause any issues when running 3.5 after upgrade from 3.4. So, I've checked for [usage](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/etcd-io/etcd%24+rev:release-3.5+UnsafeConfStateFromBackend&patternType=standard&sm=0&groupBy=path) of `UnsafeConfStateFromBackend` in release-3.5 branch and didn't find any.

tests/e2e/utl_migrate_test.go
```diff
@@ -164,12 +186,104 @@ func TestEtctlutlMigrate(t *testing.T) {
...
+
+			for _, bk := range tc.expectNonFoundKeys {
+				_, vs := be.ReadTx().UnsafeRange(bk.bucket, bk.key, nil, 0)
+				assert.Zero(t, len(vs))
+			}
+
+			be.Close()
+
+			if tc.delWal {
+				// clear out v2store to avoid "etcdserver/membership: cluster cannot be downgraded (current version: X is lower than determined cluster version: X)"
```
serathius, line 203: Don't think this is correct. Reason `etcdserver/membership: cluster cannot be downgraded` is written is because etcd server will still replay wal entries from since last snapshot. Proper solution would be to ensure that there is no `SetClusterVersion` wal entries since last snapshot.

serathius:
ping @lavacat

tests/e2e/utl_migrate_test.go
lavacat: quick update: working on adding snapshot save/restore for the case when `/0/version` entry is in wal log. I think that's the best way to go about it, since that's how users will need to handle this.

serathius: > quick update: working on adding snapshot save/restore for the case when /0/version entry is in wal log

Don't understand. You mean to use snapshot save/restore to workaround the issue. I don't think it's correct, the only reason it works is fact that it throws out the WAL completely. Don't think this is correct in 3 node clusters as there is no guarantee that snapshot will save the same data on different nodes.

Proper solution would be to modify/remove the SetClusterVersion WAL entry. Generally it can be risky, as this entry protects etcd from writing incompatible WAL entries that would be interpreted wrongly. There were some new fields introduced into etcd v3.5 WAL. List is available in https://github.com/etcd-io/etcd/blob/main/scripts/etcd_version_annotations.txt. 

For v3.6 downgrades I implemented a mechanism to detect WAL version based on proto annotations https://github.com/etcd-io/etcd/blob/main/server/storage/wal/version.go. However maybe we could check what proto versions were between v3.4 and v3.4, maybe the differences are not critical so we could yolo it.

lavacat: @serathius you are correct, one has to use same snapshot. I was thinking of something like [this](https://github.com/etcd-io/etcd/issues/15878#issuecomment-1571620392) for 3 node cluster. There will be cluster downtime if using this approach. There is probably a chance of data loss for entries that are in wal but not applied to backend (maybe there is a way to mitigate this).

Updating WAL is probably ideal. [Here](https://www.diffnow.com/report/2fnge) is the diff of raft_internal. From the first glance all new requests can be removed during migrate.`ClusterVersionSetRequest`, `ClusterMemberAttrSetRequest`, `DowngradeInfoSetRequest` and `AuthStatusRequest`. I still need to check every existing request for possible new non-optional fields.

I'll give it a shot and try to add wal update to `migrate`.

serathius: Please note that each member could be on different stage of applying WAL and storing db. We need to make sure that applying those entries works the same on both v3.5/v3.4.

scripts/test.sh
```diff
@@ -559,22 +559,42 @@ function dep_pass {
...
+    UPGRADE_VER_BEFORE_LAST=$MANUAL_VER_BEFORE_LAST
+  fi
+  if [[ -z ${UPGRADE_VER_BEFORE_LAST} ]]; then
+    UPGRADE_VER_BEFORE_LAST="v3.4.0"
+    log_warning "fallback to" ${UPGRADE_VER_BEFORE_LAST}
+  fi
+
+  download_etcd_ver_to_tem ${UPGRADE_VER_BEFORE_LAST}
+  mv /tmp/etcd ./bin/etcd-before-last-release
+}
```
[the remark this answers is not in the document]
lavacat: @ahrtr trying to rebase from main. New logic uses minor version to determine prior release. But current [test](https://github.com/etcd-io/etcd/blob/f3cc759afc679983a609e462c2808265d989e2dd/tests/e2e/utl_migrate_test.go#L80) (even without my changes) assumes major version. It works right now because version in `main` is `3.6.0-alpha.0`, so it will get 3.5.X.
Shall I change the logic of the script to use major version?

jmhbnz:
Hey @lavacat - When you have a moment can you please resolve conflicts on this?  Marek, Benjamin, Wenjia and I discussed this pr today as still being key for upcoming release.

lavacat:
related doc [Downgrade Support 3.5 to 3.4 ](https://docs.google.com/document/d/16XEGyPBisZvmmoIHSZzv__LoyOeluC5a4x353CX0SIM/edit)

Issue #15993, opened 2023-06-01
Put operation with lease takes linear apply time depending on number of keys already attached to lease 


marseel:
### What happened?

Put request with lease takes linear apply time depending on the number of keys already attached to the lease. With the increasing number of keys and constant QPS, it's increasing to the point, where etcd spend almost 100% of time applying put requests making lease renewal fail. 

### What did you expect to happen?

I would expect put with lease not to depend linearly on number of already attached keys to lease.

### How can we reproduce it (as minimally and precisely as possible)?

Run Etcd with:
```
docker run   -p 2379:2379   -p 2380:2380   --mount type=bind,source=/tmp/etcd-data.tmp,destination=/etcd-data   --name etcd-gcr-v3.5.9   gcr.io/etcd-development/etcd:v3.5.9   /usr/local/bin/etcd   --name s1   --data-dir /etcd-data   --listen-client-urls http://0.0.0.0:2379   --advertise-client-urls http://0.0.0.0:2379   --listen-peer-urls http://0.0.0.0:2380   --initial-advertise-peer-urls http://0.0.0.0:2380   --initial-cluster s1=http://0.0.0.0:2380   --initial-cluster-token tkn   --initial-cluster-state new  --auto-compaction-retention=1 --log-level info   --logger zap   --log-outputs stderr
```
Test client:
```golang
package main

import (
	"context"
	"fmt"
	"strconv"
	"sync"
	"time"

	client "go.etcd.io/etcd/client/v3"
	"go.etcd.io/etcd/client/v3/concurrency"
	"golang.org/x/time/rate"
)

func main() {
	c, err := client.New(client.Config{
		Endpoints:   []string{"localhost:2379"},
		DialTimeout: 5 * time.Second})
	fmt.Println("Opening session")
	session, err := concurrency.NewSession(c, concurrency.WithTTL(int(30)))
	if err != nil {
		return
	}

	limiter := rate.NewLimiter(1000, 20)
	maxDuration := time.Duration(0)
	mu := sync.Mutex{}

	fmt.Println("Starting writes")
	for i := 0; i <= 10000000; i++ {
		i := i
		limiter.Wait(context.Background())
		key := "asdasdasdasdasd" + strconv.Itoa(i)
		go func() {
			leaseID := session.Lease()
			start := time.Now()
			c.Put(context.Background(), key, key, client.WithLease(leaseID))
			duration := time.Since(start)

			mu.Lock()
			defer mu.Unlock()
[...]

PR #16005, opened 2023-06-05
Early exit auth check on lease puts

tjungblu:
Mitigates etcd-io#15993 by not checking each key individually for permission when auth is entirely disabled or admin user is calling the method.

### DIFF
--- a/server/etcdserver/apply/apply_auth.go
+++ b/server/etcdserver/apply/apply_auth.go
@@ -125,19 +125,24 @@ func (aa *authApplierV3) LeaseRevoke(lc *pb.LeaseRevokeRequest) (*pb.LeaseRevoke
 func (aa *authApplierV3) checkLeasePuts(leaseID lease.LeaseID) error {
 	l := aa.lessor.Lookup(leaseID)
 	if l != nil {
-		// early return for most-common scenario of either disabled auth or admin user.
-		// IsAdminPermitted also checks whether auth is enabled
-		if err := aa.as.IsAdminPermitted(&aa.authInfo); err == nil {
-			return nil
-		}
-
-		for _, key := range l.Keys() {
-			if err := aa.as.IsPutPermitted(&aa.authInfo, []byte(key)); err != nil {
-				return err
-			}
-		}
+		return aa.checkLeasePutsKeys(l)
 	}
 
+	return nil
+}
+
+func (aa *authApplierV3) checkLeasePutsKeys(l *lease.Lease) error {
+	// early return for most-common scenario of either disabled auth or admin user.
+	// IsAdminPermitted also checks whether auth is enabled
+	if err := aa.as.IsAdminPermitted(&aa.authInfo); err == nil {
+		return nil
+	}
+
+	for _, key := range l.Keys() {
+		if err := aa.as.IsPutPermitted(&aa.authInfo, []byte(key)); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
--- a/server/etcdserver/apply/apply_auth.go
+++ b/server/etcdserver/apply/apply_auth.go
@@ -125,6 +125,12 @@ func (aa *authApplierV3) LeaseRevoke(lc *pb.LeaseRevokeRequest) (*pb.LeaseRevoke
 func (aa *authApplierV3) checkLeasePuts(leaseID lease.LeaseID) error {
 	l := aa.lessor.Lookup(leaseID)
 	if l != nil {
+		// early return for most-common scenario of either disabled auth or admin user.
+		// IsAdminPermitted also checks whether auth is enabled
+		if err := aa.as.IsAdminPermitted(&aa.authInfo); err == nil {
+			return nil
+		}
+
 		for _, key := range l.Keys() {
 			if err := aa.as.IsPutPermitted(&aa.authInfo, []byte(key)); err != nil {
 				return err
--- a/server/etcdserver/apply/apply_auth_test.go
+++ b/server/etcdserver/apply/apply_auth_test.go
@@ -1,0 +1,124 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apply
+
+import (
+	"encoding/base64"
+	"testing"
+	"time"
+
+	"golang.org/x/crypto/bcrypt"
+
+	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap/zaptest"
+
+	"go.etcd.io/etcd/server/v3/storage/schema"
+
+	"go.etcd.io/etcd/api/v3/authpb"
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/server/v3/auth"
+	"go.etcd.io/etcd/server/v3/lease"
+)
+
+func TestCheckLeasePutsKeys(t *testing.T) {
+	lg := zaptest.NewLogger(t)
+
+	b, _ := betesting.NewDefaultTmpBackend(t)
+	defer betesting.Close(t, b)
+
+	simpleTokenTTLDefault := 300 * time.Second
+	tokenTypeSimple := "simple"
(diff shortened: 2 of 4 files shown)


jmhbnz (commented):
Hey @tjungblu nice work looking into this!

Any chance you could replicate the metrics in the linked issue and post here so we could compare before and after?

server/etcdserver/apply/apply_auth.go
```diff
@@ -125,6 +125,11 @@ func (aa *authApplierV3) LeaseRevoke(lc *pb.LeaseRevokeRequest) (*pb.LeaseRevoke
 func (aa *authApplierV3) checkLeasePuts(leaseID lease.LeaseID) error {
 	l := aa.lessor.Lookup(leaseID)
 	if l != nil {
+		// early return for most-common scenario of either disabled auth or admin user
+		if aa.as.IsAdminPermitted(&aa.authInfo) == nil {
+			return nil
```
serathius, line 130: Do we have a test covering this case?

server/etcdserver/apply/apply_auth.go
```diff
@@ -125,6 +125,11 @@ func (aa *authApplierV3) LeaseRevoke(lc *pb.LeaseRevokeRequest) (*pb.LeaseRevoke
 func (aa *authApplierV3) checkLeasePuts(leaseID lease.LeaseID) error {
 	l := aa.lessor.Lookup(leaseID)
 	if l != nil {
+		// early return for most-common scenario of either disabled auth or admin user
+		if aa.as.IsAdminPermitted(&aa.authInfo) == nil {
```
serathius, line 129: It's not obvious from proposed check that it covers case that auth is disabled. From code design it seems strange  for me that `authApplierV3` will be used at all if auth is disabled.

tjungblu: I was surprised too, it even goes through an rw mutex everytime it checks...

https://github.com/etcd-io/etcd/issues/15993#issuecomment-1573550357

I reckon we should discuss a bigger refactoring for the auth system, potentially?

server/etcdserver/apply/apply_auth.go
tjungblu: Neither apply.go, nor apply_auth.go have unit test coverage-  there are no unit tests as far as I can tell.

server/etcdserver/apply/apply_auth.go
serathius: Maybe, depends on whether we want to backport this fix. This seems like a trivial but important change. Appliers changed a lot between v3.5 and main.

ahrtr: > From code design it seems strange for me that authApplierV3 will be used at all if auth is disabled.

> I reckon we should discuss a bigger refactoring for the auth system, potentially?

Because the existing chain of applier is static. If we want to get rid of authApplier completely when auth isn't enabled, then we need to make the chain dynamic. We don't have to do it, please feel free to evaluate the effort and impact separately if anyone is interested.

> It's not obvious from proposed check that it covers case that auth is disabled.

It's a common "issue" in all existing `(*authStore) IsXXXPermitted(...) error` methods, when auth isn't enabled, they return nil.

> depends on whether we want to backport this fix

I think we need to backport the fix.

ahrtr:
The PR looks good to me. @tjungblu did you compare the performance before and after applying this PR?

tjungblu:
running this entirely local with the test supplied from the issue, I can see the p99 latency to be consistently good:

image

here's the same result without the patch right afterwards in comparison:

image

The p99 looks very low in absolute by prometheus because of the scraping interval and one minute resolution, it actually degrades almost immediately into 100ms+ latency after 10k entries:

> MAX latency: 103.208013ms, entries: 11405 

which is basically a few seconds into the unit test. 


--- 

Just for completeness with a comparison of the same metric used in the issue report also stays flat over the whole execution time:
image

Even though I'm not entirely sure whether it makes sense to sum over the rate duration sum here.

server/etcdserver/apply/apply_auth.go
tjungblu: > we need to make the chain dynamic.  

I haven't researched too deeply yet, but the whole system is already dynamic because the whole auth goes through raft (even enable/disable). I would've expected the enable/disable methods to be a configuration for startup, not a dynamic apply request. Hence the static implementation, I guess.

I'll open a discussion issue up as a feature request, will do some benchmarks in the meantime around the impact of either approach.

ahrtr: The link you provide just replaces the whole applier chain with `applierV3Capped` or `applierV3Corrupt`. But you can't change (swap out) part of the chain: `authApplier -> quotaApplier -> backendApplier`. 

It's similar to `string` in golang, it's immutable. You can't modify part of a string; instead, you can only replace the whole string.

Just as I mentioned previously, we don't have to make the applier chain dynamic. But of course, it's open to discussion in a separate session.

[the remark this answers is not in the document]
tjungblu: updated

tjungblu: No doubt on the immutability of the chain, but the pointer to the applier struct needs to be guarded by a mutex when you swap it out from another goroutine?

https://github.com/etcd-io/etcd/blob/release-3.5/server/etcdserver/server.go#L252

let me try to tease the race detector on this case more specifically. Not that it matters much, if you're corrupted/OOS it's game over anyway, it just tickled my inner race detector.

ahrtr (approved):
LGTM

Thanks @tjungblu 

It's a straightforward change, and also a low-hang fruit. 

Could you please double check also backport this PR to 3.5 and 3.4 if needed? @tjungblu  Of course, waiting for other maintainers approve and merge this PR.

tjungblu:
@marseel I assume you would need this at the very least in the next 3.5.x release?

marseel:
3.5.x release would be great.

tjungblu:
cool, then I'll get to write a backport. Thanks everyone!

jmhbnz (approved):

serathius:
One think before merging. @tjungblu I know we don't have tests, however I would feel much safer for backports if we could add some even trivial test.

tjungblu:
Totally makes sense, let me try to add one right here as a separate commit. I'll cherry-pick it down for the backport PRs then.

serathius (changes requested):
Can we add a trivial test to make backports safer?

server/etcdserver/apply/apply_auth.go
serathius: My suggestion would be to add `ArePutsPermitted` method that will for the range and add tests for it.

tjungblu:
@serathius can we move this forward? Is the testcase sufficient for you or too much refactoring already?

server/auth/store_mock.go
```diff
@@ -14,7 +14,39 @@
```
serathius, line 14: Don't move test file to normal file. Don't share mocks outside of package.

server/auth/nop.go
```diff
@@ -18,18 +18,18 @@ import (
 	"context"
 )
 
-type tokenNop struct{}
+type TokenNop struct{}
```
serathius, line 21: Don't expose internal testing code from package! Use proper TokenProvider in external tests.

tjungblu: reverted

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -0,0 +1,122 @@
...
+			ch <- struct{}{}
+		}()
+		return ch
+	}
+
+	tp, _ := auth.NewTokenProvider(zaptest.NewLogger(t), tokenTypeSimple, dummyIndexWaiter, simpleTokenTTLDefault)
+	as := auth.NewAuthStore(lg, schema.NewAuthBackend(lg, b), tp, 4)
+
+	aa := authApplierV3{as: as}
+	assert.Nil(t, aa.checkLeasePutsKeys(lease.NewLease(lease.LeaseID(1), 3600)), "auth is disabled, should allow puts")
```
serathius, line 55:
```suggestion
	assert.NoError(t, aa.checkLeasePutsKeys(lease.NewLease(lease.LeaseID(1), 3600)), "auth is disabled, should allow puts")
```

serathius:
Overall looks very good! Tests you added are awesome! Left some small comments.

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -0,0 +1,122 @@
...
+	dummyIndexWaiter := func(index uint64) <-chan struct{} {
+		ch := make(chan struct{}, 1)
+		go func() {
+			ch <- struct{}{}
+		}()
+		return ch
+	}
+
+	tp, _ := auth.NewTokenProvider(zaptest.NewLogger(t), tokenTypeSimple, dummyIndexWaiter, simpleTokenTTLDefault)
+	as := auth.NewAuthStore(lg, schema.NewAuthBackend(lg, b), tp, 4)
```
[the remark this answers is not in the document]
tjungblu: replaced with bcrypt.MinCost

tjungblu:
fixed the goimport manually, I reckon `make fix` doesn't actually do anything?

serathius (approved):

Issue #16007, opened 2023-06-05
Livez/Readyz


logicalhan:
### What would you like to be added?

We currently have a single health endpoint for etcd `/health` which is used in Kubernetes distros as both liveness and readiness checking. In order to be fully api-compliant, we should have both a liveness check (i.e. `/livez`) which checks that this individual etcd member is "alive" and does not need to be restarted and a readiness check (i.e. `/readyz`) which signals that the etcd member is ready to accept traffic. 

### Why is this needed?

There is a difference between "please restart me I'm that unhealthy" vs "please send me all sorts of traffic, I'm ready for it".

chaochn47:
Please check this https://github.com/etcd-io/etcd/issues/13340#issuecomment-963524047

logicalhan:
> Please check this [#13340 (comment)](https://github.com/etcd-io/etcd/issues/13340#issuecomment-963524047)

Yeah I don't buy it. No one is going to dig up an obscure github issue in order to properly configure their etcd configurations for Kubernetes.

chaochn47:
Yeah that makes sense. We should rethink and document it properly, for example it applies only which etcd version, etc.

ahrtr:
Thanks @logicalhan for raising this request. I am supportive on it.  `health/serializable=<true|false>` isn't an an explicit API, and it also requires people to understand what's serializable.

`livez` and `readyz` are more explicit and easy to understand & use.

`/livez` is similar to (or a syntax sugar ) `health/serializable=true`; while `readyz` should require quorum, and actually check the health of the cluster, so it's similar to (or a syntax sugar)` `health/serializable=false` or `health`.

ahrtr:
cc @neolit123

neolit123:
+1

serathius:
Don't want to rush into adding livez/readyz probe. Main problem with existing health probe we just added it to have it without proper consideration. 

I want livez to properly reflect fact that etcd needs restart, for example etcd is stuck on stalled storage https://docs.google.com/document/d/1U9hAcZQp3Y36q_JFiw2VBJXVAo2dK2a-8Rsbqv3GgDo/edit?usp=sharing. 

Readyz should properly reflect fact that etcd is ready to serve traffic. Don't think alarms matter here. It's a degradation, however it doesn't mean we shouldn't serve reads.

[...]

PR #16008, opened 2023-06-05
[concept] add livez/readyz for etcd

logicalhan:
This is a prototype for adding livez/readyz support to etcd. Currently I've configured the `NO_SPACE` alarm to only count towards `\readyz` since it means etcd is degraded. Whether quorum should be included in liveness is an open question.

### DIFF
--- a/server/etcdserver/api/etcdhttp/health.go
+++ b/server/etcdserver/api/etcdhttp/health.go
@@ -23,16 +23,19 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
+	"go.etcd.io/raft/v3"
+
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/auth"
 	"go.etcd.io/etcd/server/v3/config"
-	"go.etcd.io/raft/v3"
 )
 
 const (
 	PathHealth      = "/health"
+	PathLivez       = "/livez"
+	PathReadyz      = "/readyz"
 	PathProxyHealth = "/proxy/health"
 )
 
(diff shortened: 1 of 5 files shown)


server/etcdserver/api/etcdhttp/health.go
```diff
@@ -46,33 +49,59 @@ type ServerHealth interface {
...
+
+// HandleLivez registers metrics and health handlers. it checks health by using v3 range request
+// and its corresponding timeout.
+func HandleLivez(lg *zap.Logger, mux *http.ServeMux, srv ServerHealth) {
+	mux.Handle(PathLivez, NewHealthHandler(lg, func(excludedAlarms AlarmSet, serializable bool, endpoint string) Health {
+		if h := checkAlarms(lg, srv, excludedAlarms, endpoint); h.Health != "true" {
+			return h
+		}
+		// TODO(logicalhan) should we require quorum for livez?
+		return checkAPI(lg, srv, serializable)
```
ahrtr, line 71:
```suggestion
		return checkAPI(lg, srv, true)
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -46,33 +49,59 @@ type ServerHealth interface {
...
+}
+
+// HandleReadyz registers metrics and health handlers. it checks health by using v3 range request
+// and its corresponding timeout.
+func HandleReadyz(lg *zap.Logger, mux *http.ServeMux, srv ServerHealth) {
+	mux.Handle(PathReadyz, NewHealthHandler(lg, func(excludedAlarms AlarmSet, serializable bool, endpoint string) Health {
+		if h := checkAlarms(lg, srv, excludedAlarms, endpoint); h.Health != "true" {
+			return h
+		}
+		return checkAPI(lg, srv, serializable)
```
ahrtr, line 82:
```suggestion
		return checkAPI(lg, srv, false)
```

server/etcdserver/api/etcdhttp/health.go
ahrtr: `livez` doesn't require quorum

logicalhan: You mean it shouldn't?

logicalhan: I'm confused, do you mean 'false' here and 'true' on line 82?

logicalhan: Also I have a todo here for a specific reason. I'm not even sure that this code is doing the right thing, since it's going through the etcdclient and therefore can hit any particular etcd member. So even if serializable is set to false and we're getting stale reads from a particular etcd member, we're not actually guaranteed to hit our own etcd member, which is the thing we need to be testing for.

ahrtr: Right. `livez` just check local instance's health. We shouldn't restart the local instance if the cluster is NOT health, e.g. quorum isn't satisfied.

logicalhan: Yes, but we shouldn't even be going through the etcd client, is my point, since it load balances.

logicalhan: Ah, it's not going through the client if serializable is false. I see.

serathius (changes requested):
Don't want to rash into adding livez/readyz probe. Main problem with existing health probe we just added it to have it without proper consideration. 

I want livez to properly reflect fact that etcd needs restart, for example etcd is stuck on stalled storage https://docs.google.com/document/d/1U9hAcZQp3Y36q_JFiw2VBJXVAo2dK2a-8Rsbqv3GgDo/edit?usp=sharing. 

Readyz should properly reflect fact that etcd is ready to serve traffic. Don't think alarms matter here. It's a degradation, however it doesn't mean we shouldn't serve reads.

TLDR; I would like to have a design written that will do a proper analysis etcd failure modes and propose matching probes to detect them. Example https://github.com/kubernetes-sigs/metrics-server/issues/542

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -141,7 +169,7 @@ func getSerializableFlag(r *http.Request) bool {
 
 // TODO: etcdserver.ErrNoLeader in health API
 
-func checkAlarms(lg *zap.Logger, srv ServerHealth, excludedAlarms AlarmSet) Health {
+func checkAlarms(lg *zap.Logger, srv ServerHealth, excludedAlarms AlarmSet, healthType string) Health {
```
ahrtr, line 172: The `healthType string` isn't used at all, can we remove it? 
```suggestion
func checkAlarms(lg *zap.Logger, srv ServerHealth, excludedAlarms AlarmSet) Health {
```

jmhbnz:
Discussed during sig-etcd triage meeting. This original concept has now been superceeded by the work from @siyuanfoundation 🎉

siyuanfoundation:
the work is tracked in https://github.com/etcd-io/etcd/issues/16007

Issue #16028, opened 2023-06-07
Restoring etcd without breaking API guarantees, aka etcd restore for Kubernetes


serathius:
### What would you like to be added?

This is etcd side bug for  https://github.com/kubernetes/kubernetes/issues/118501

tldr; Etcd restore operation breaks existing etcd guarantees about revision never decreasing.

## Proposal
* Etcd restore operation should support bumping revision of the next etcd operations by an arbitrary number.
* Etcd restore operation should support marking old revisions as compacted.
* Etcd restore operation should support bumping revision of the latest key versions.

### Tracking work
* [x] Implement optional `--bump-revision` flag to `etcdutl snapshot restore` operation.
* [x] Implement optional `--mark-compacted` flag to `etcdutl snapshot restore` operation.
* [x] Confirm that those changes are backport compatible for v3.5 and v3.4
* [x] Add etcd restore with e2e tests that confirm that:
    * Revision doesn't decrease if value provided via `--bump-revision` is high enough
    * Client is notified about restore by watch being broken
    * Client cannot access revisions from between backup and restore operation
* [x] Backport the changes and tests to v3.5 and v3.4 etcd.

 

### Why is this needed?

Without zero official guidance each vendor/administrator had a different plan for etcd restore many of which were broken and incorrect. Having an official guided restore operation allow whole community to work together and share their experience leading to improvements in Kubernetes disaster recovery handling.

serathius:
cc @ptabor @ahrtr

ahrtr:
Offline compaction makes sense to me. We may need to handle it from boht etcd and bbolt perspective. Let me take care of it. But it may take some time, because I will be on some personal stuff starting from next Tuesday and until 1 ~ 2 weeks later.

I am not against offline bumping revision, but please see my concern/comment in https://github.com/etcd-io/etcd/discussions/15381#discussioncomment-6117015. Would like to get more feedback before we move on.

tjungblu:
I think we already have `etcdctl compaction <rev>` that does that, albeit online.

Maybe before we go into the debate on _what_ we need to implement for compaction, do we even need to compact?
I think what Marek meant, is that we return the current revision as "compacted" to force reset the informer caches, not do an actual compaction.

tjungblu:
(moving your arguments over from https://github.com/etcd-io/etcd/discussions/15381#discussioncomment-6117015 so we don't split the discussion into too many places)

> Usually it is risky for an offline tool to modify the db data

[...]

PR #16029, opened 2023-06-07
Adding optional revision bump to snapshot restore

dusk125:
This PR adds the flags `--bump-revision` and `--mark-compacted` to etcdutl snapshot restore for #16028.

I chose to handle this after the database is loaded from the snapshot, so as to not modify the snapshot, I didn't want to risk corrupting the snapshot as a part of this optional step. If for some reason the bump fails, a user would still be able to restore the snapshot again without the bump and restart the watchers manually.

### DIFF
--- a/etcdutl/etcdutl/snapshot_command.go
+++ b/etcdutl/etcdutl/snapshot_command.go
@@ -38,6 +38,8 @@ var (
 	restorePeerURLs     string
 	restoreName         string
 	skipHashCheck       bool
+	markCompacted       bool
+	revisionBump        uint64
 )
 
 // NewSnapshotCommand returns the cobra command for "snapshot".
@@ -75,6 +77,8 @@ func NewSnapshotRestoreCommand() *cobra.Command {
 	cmd.Flags().StringVar(&restorePeerURLs, "initial-advertise-peer-urls", defaultInitialAdvertisePeerURLs, "List of this member's peer URLs to advertise to the rest of the cluster")
 	cmd.Flags().StringVar(&restoreName, "name", defaultName, "Human-readable name for this member")
 	cmd.Flags().BoolVar(&skipHashCheck, "skip-hash-check", false, "Ignore snapshot integrity hash value (required if copied from data directory)")
+	cmd.Flags().Uint64Var(&revisionBump, "bump-revision", 0, "How much to increase the latest revision after restore (required if --mark-compacted)")
+	cmd.Flags().BoolVar(&markCompacted, "mark-compacted", false, "Mark the latest revision after restore as the point of scheduled compaction (required if --bump-revision > 0)")
 
 	cmd.MarkFlagDirname("data-dir")
 	cmd.MarkFlagDirname("wal-dir")
@@ -100,7 +104,7 @@ func SnapshotStatusCommandFunc(cmd *cobra.Command, args []string) {
 
 func snapshotRestoreCommandFunc(_ *cobra.Command, args []string) {
 	SnapshotRestoreCommandFunc(restoreCluster, restoreClusterToken, restoreDataDir, restoreWalDir,
-		restorePeerURLs, restoreName, skipHashCheck, args)
+		restorePeerURLs, restoreName, skipHashCheck, revisionBump, markCompacted, args)
 }
 
 func SnapshotRestoreCommandFunc(restoreCluster string,
@@ -110,9 +114,16 @@ func SnapshotRestoreCommandFunc(restoreCluster string,
 	restorePeerURLs string,
 	restoreName string,
 	skipHashCheck bool,
+	revisionBump uint64,
+	markCompacted bool,
 	args []string) {
 	if len(args) != 1 {
 		err := fmt.Errorf("snapshot restore requires exactly one argument")
+		cobrautl.ExitWithError(cobrautl.ExitBadArgs, err)
+	}
+
+	if (revisionBump == 0 && markCompacted) || (revisionBump > 0 && !markCompacted) {
+		err := fmt.Errorf("--mark-compacted required if --revision-bump > 0")
 		cobrautl.ExitWithError(cobrautl.ExitBadArgs, err)
 	}
 
@@ -138,6 +149,8 @@ func SnapshotRestoreCommandFunc(restoreCluster string,
 		InitialCluster:      restoreCluster,
 		InitialClusterToken: restoreClusterToken,
 		SkipHashCheck:       skipHashCheck,
+		RevisionBump:        revisionBump,
+		MarkCompacted:       markCompacted,
 	}); err != nil {
 		cobrautl.ExitWithError(cobrautl.ExitError, err)
 	}
--- a/etcdutl/snapshot/util.go
+++ b/etcdutl/snapshot/util.go
@@ -23,9 +23,31 @@ type revision struct {
 	sub  int64
 }
 
+// GreaterThan should be synced with function in server
+// https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/revision.go
+func (a revision) GreaterThan(b revision) bool {
+	if a.main > b.main {
+		return true
+	}
+	if a.main < b.main {
+		return false
+	}
+	return a.sub > b.sub
+}
+
+// bytesToRev should be synced with function in server
+// https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/revision.go
 func bytesToRev(bytes []byte) revision {
 	return revision{
 		main: int64(binary.BigEndian.Uint64(bytes[0:8])),
 		sub:  int64(binary.BigEndian.Uint64(bytes[9:])),
 	}
+}
+
+// revToBytes should be synced with function in server
+// https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/revision.go
+func revToBytes(bytes []byte, rev revision) {
+	binary.BigEndian.PutUint64(bytes[0:8], uint64(rev.main))
+	bytes[8] = '_'
+	binary.BigEndian.PutUint64(bytes[9:], uint64(rev.sub))
 }
--- a/etcdutl/snapshot/v3_snapshot.go
+++ b/etcdutl/snapshot/v3_snapshot.go
@@ -41,6 +41,7 @@ import (
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v2store"
 	"go.etcd.io/etcd/server/v3/etcdserver/cindex"
 	"go.etcd.io/etcd/server/v3/storage/backend"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
 	"go.etcd.io/etcd/server/v3/storage/schema"
 	"go.etcd.io/etcd/server/v3/storage/wal"
 	"go.etcd.io/etcd/server/v3/storage/wal/walpb"
@@ -203,6 +204,16 @@ type RestoreConfig struct {
 	// SkipHashCheck is "true" to ignore snapshot integrity hash value
 	// (required if copied from data directory).
 	SkipHashCheck bool
+
+	// RevisionBump is the amount to increase the latest revision after restore,
+	// to allow administrators to trick clients into thinking that revision never decreased.
+	// If 0, revision bumping is skipped.
+	// (required if MarkCompacted == true)
+	RevisionBump uint64
+
+	// MarkCompacted is "true" to mark the latest revision as compacted.
+	// (required if RevisionBump > 0)
+	MarkCompacted bool
 }
 
 // Restore restores a new etcd data directory from given snapshot file.
@@ -265,6 +276,13 @@ func (s *v3Manager) Restore(cfg RestoreConfig) error {
 	if err = s.saveDB(); err != nil {
 		return err
 	}
+
+	if cfg.MarkCompacted && cfg.RevisionBump > 0 {
+		if err = s.modifyLatestRevision(cfg.RevisionBump); err != nil {
+			return err
+		}
+	}
+
 	hardstate, err := s.saveWALAndSnap()
 	if err != nil {
 		return err
@@ -309,6 +327,70 @@ func (s *v3Manager) saveDB() error {
 	}
 
 	return nil
+}
+
+// modifyLatestRevision can increase the latest revision by the given amount and sets the scheduled compaction
+// to that revision so that the server will consider this revision compacted.
+func (s *v3Manager) modifyLatestRevision(bumpAmount uint64) error {
+	be := backend.NewDefaultBackend(s.lg, s.outDbPath())
+	defer func() {
+		be.ForceCommit()
+		be.Close()
+	}()
+
+	tx := be.BatchTx()
+	tx.LockOutsideApply()
+	defer tx.Unlock()
+
+	latest, err := s.unsafeGetLatestRevision(tx)
+	if err != nil {
+		return err
+	}
+
+	latest = s.unsafeBumpRevision(tx, latest, int64(bumpAmount))
+	s.unsafeMarkRevisionCompacted(tx, latest)
+
+	return nil
+}
+
+func (s *v3Manager) unsafeBumpRevision(tx backend.BatchTx, latest revision, amount int64) revision {
+	s.lg.Info(
+		"bumping latest revision",
+		zap.Int64("latest-revision", latest.main),
+		zap.Int64("bump-amount", amount),
+		zap.Int64("new-latest-revision", latest.main+amount),
+	)
+
+	latest.main += amount
+	latest.sub = 0
+	k := make([]byte, 17)
+	revToBytes(k, latest)
+	tx.UnsafePut(schema.Key, k, []byte{})
+
+	return latest
+}
+
+func (s *v3Manager) unsafeMarkRevisionCompacted(tx backend.BatchTx, latest revision) {
+	s.lg.Info(
+		"marking revision compacted",
+		zap.Int64("revision", latest.main),
+	)
+
+	mvcc.UnsafeSetScheduledCompact(tx, latest.main)
+}
+
+func (s *v3Manager) unsafeGetLatestRevision(tx backend.BatchTx) (revision, error) {
+	var latest revision
+	err := tx.UnsafeForEach(schema.Key, func(k, _ []byte) (err error) {
+		rev := bytesToRev(k)
+
+		if rev.GreaterThan(latest) {
+			latest = rev
+		}
+
+		return nil
+	})
+	return latest, err
 }
 
 func (s *v3Manager) copyAndVerifyDB() error {
--- a/tests/e2e/ctl_v3_snapshot_test.go
+++ b/tests/e2e/ctl_v3_snapshot_test.go
@@ -27,9 +27,12 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	v3rpc "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdutl/v3/snapshot"
 	"go.etcd.io/etcd/pkg/v3/expect"
+	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
 )
 
 func TestCtlV3Snapshot(t *testing.T) { testCtl(t, snapshotTest) }
@@ -303,5 +306,132 @@ func snapshotVersionTest(cx ctlCtx) {
 	}
 	if st.Version != "3.6.0" {
 		cx.t.Fatalf("expected %q, got %q", "3.6.0", st.Version)
+	}
+}
+
+func TestRestoreCompactionRevBump(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t,
+		e2e.WithClusterSize(1),
+		e2e.WithKeepDataDir(true),
+	)
+	if err != nil {
+		t.Fatalf("could not start etcd process cluster (%v)", err)
+	}
+	defer func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	}()
+
+	ctl := epc.Etcdctl()
+
+	watchCh := ctl.Watch(context.Background(), "foo", config.WatchOptions{Prefix: true})
+	// flake-fix: the watch can sometimes miss the first put below causing test failure
+	time.Sleep(100 * time.Millisecond)
+
+	kvs := []testutils.KV{{Key: "foo1", Val: "val1"}, {Key: "foo2", Val: "val2"}, {Key: "foo3", Val: "val3"}}
+	for i := range kvs {
+		require.NoError(t, ctl.Put(context.Background(), kvs[i].Key, kvs[i].Val, config.PutOptions{}))
+	}
+
+	watchTimeout := 1 * time.Second
+	watchRes, err := testutils.KeyValuesFromWatchChan(watchCh, len(kvs), watchTimeout)
+	require.NoErrorf(t, err, "failed to get key-values from watch channel %s", err)
+	require.Equal(t, kvs, watchRes)
+
+	// ensure we get the right revision back for each of the keys
+	currentRev := 4
+	baseRev := 2
+	hasKVs(t, ctl, kvs, currentRev, baseRev)
+
+	fpath := filepath.Join(t.TempDir(), "test.snapshot")
+
+	t.Log("etcdctl saving snapshot...")
+	cmdPrefix := []string{e2e.BinPath.Etcdctl, "--endpoints", strings.Join(epc.EndpointsGRPC(), ",")}
+	require.NoError(t, e2e.SpawnWithExpects(append(cmdPrefix, "snapshot", "save", fpath), nil, fmt.Sprintf("Snapshot saved at %s", fpath)))
+
+	// add some more kvs that are not in the snapshot that will be lost after restore
+	unsnappedKVs := []testutils.KV{{Key: "unsnapped1", Val: "one"}, {Key: "unsnapped2", Val: "two"}, {Key: "unsnapped3", Val: "three"}}
+	for i := range unsnappedKVs {
+		require.NoError(t, ctl.Put(context.Background(), unsnappedKVs[i].Key, unsnappedKVs[i].Val, config.PutOptions{}))
+	}
+
+	t.Log("Stopping the original server...")
+	require.NoError(t, epc.Stop())
+
+	newDataDir := filepath.Join(t.TempDir(), "test.data")
+	t.Log("etcdctl restoring the snapshot...")
+	bumpAmount := 10000
+	err = e2e.SpawnWithExpect([]string{
+		e2e.BinPath.Etcdutl,
+		"snapshot",
+		"restore", fpath,
+		"--name", epc.Procs[0].Config().Name,
+		"--initial-cluster", epc.Procs[0].Config().InitialCluster,
+		"--initial-cluster-token", epc.Procs[0].Config().InitialToken,
+		"--initial-advertise-peer-urls", epc.Procs[0].Config().PeerURL.String(),
+		"--bump-revision", fmt.Sprintf("%d", bumpAmount),
+		"--mark-compacted",
+		"--data-dir", newDataDir,
+	}, "added member")
+	require.NoError(t, err)
+
+	t.Log("(Re)starting the etcd member using the restored snapshot...")
+	epc.Procs[0].Config().DataDirPath = newDataDir
+	for i := range epc.Procs[0].Config().Args {
+		if epc.Procs[0].Config().Args[i] == "--data-dir" {
+			epc.Procs[0].Config().Args[i+1] = newDataDir
+		}
+	}
+
+	require.NoError(t, epc.Restart(context.Background()))
+
+	t.Log("Ensuring the restored member has the correct data...")
+	hasKVs(t, ctl, kvs, currentRev, baseRev)
+	for i := range unsnappedKVs {
+		v, err := ctl.Get(context.Background(), unsnappedKVs[i].Key, config.GetOptions{})
+		require.NoError(t, err)
+		require.Equal(t, int64(0), v.Count)
+	}
+
+	cancelResult, ok := <-watchCh
+	require.True(t, ok, "watchChannel should be open")
+	require.Equal(t, v3rpc.ErrCompacted, cancelResult.Err())
+	require.Truef(t, cancelResult.Canceled, "expected ongoing watch to be cancelled after restoring with --mark-compacted")
+	require.Equal(t, int64(bumpAmount+currentRev), cancelResult.CompactRevision)
+	_, ok = <-watchCh
+	require.False(t, ok, "watchChannel should be closed after restoring with --mark-compacted")
+
+	// clients might restart the watch at the old base revision, that should not yield any new data
+	// everything up until bumpAmount+currentRev should return "already compacted"
+	for i := bumpAmount - 2; i < bumpAmount+currentRev; i++ {
+		watchCh = ctl.Watch(context.Background(), "foo", config.WatchOptions{Prefix: true, Revision: int64(i)})
+		cancelResult := <-watchCh
+		require.Equal(t, v3rpc.ErrCompacted, cancelResult.Err())
+		require.Truef(t, cancelResult.Canceled, "expected ongoing watch to be cancelled after restoring with --mark-compacted")
+		require.Equal(t, int64(bumpAmount+currentRev), cancelResult.CompactRevision)
+	}
+
+	// a watch after that revision should yield successful results when a new put arrives
+	ctx, cancel := context.WithTimeout(context.Background(), watchTimeout*5)
+	defer cancel()
+	watchCh = ctl.Watch(ctx, "foo", config.WatchOptions{Prefix: true, Revision: int64(bumpAmount + currentRev + 1)})
+	require.NoError(t, ctl.Put(context.Background(), "foo4", "val4", config.PutOptions{}))
+	watchRes, err = testutils.KeyValuesFromWatchChan(watchCh, 1, watchTimeout)
+	require.NoErrorf(t, err, "failed to get key-values from watch channel %s", err)
+	require.Equal(t, []testutils.KV{{Key: "foo4", Val: "val4"}}, watchRes)
+}
+
+func hasKVs(t *testing.T, ctl *e2e.EtcdctlV3, kvs []testutils.KV, currentRev int, baseRev int) {
+	for i := range kvs {
+		v, err := ctl.Get(context.Background(), kvs[i].Key, config.GetOptions{})
+		require.NoError(t, err)
+		require.Equal(t, int64(1), v.Count)
+		require.Equal(t, kvs[i].Val, string(v.Kvs[0].Value))
+		require.Equal(t, int64(baseRev+i), v.Kvs[0].CreateRevision)
+		require.Equal(t, int64(baseRev+i), v.Kvs[0].ModRevision)
+		require.Equal(t, int64(1), v.Kvs[0].Version)
 	}
 }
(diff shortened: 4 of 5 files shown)


serathius:
To clarify by `bumping revision` I didn't mean take transform whole change history by shifting revisions of all modifications, but to just bump revision of the next transaction.

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -75,6 +76,7 @@ func NewSnapshotRestoreCommand() *cobra.Command {
 	cmd.Flags().StringVar(&restorePeerURLs, "initial-advertise-peer-urls", defaultInitialAdvertisePeerURLs, "List of this member's peer URLs to advertise to the rest of the cluster")
 	cmd.Flags().StringVar(&restoreName, "name", defaultName, "Human-readable name for this member")
 	cmd.Flags().BoolVar(&skipHashCheck, "skip-hash-check", false, "Ignore snapshot integrity hash value (required if copied from data directory)")
+	cmd.Flags().Int64Var(&revisionBump, "bump-revision", 0, "How much to increase the revision of each key, after restore, to force a resync of watchers")
```
serathius, line 79:
```suggestion
	cmd.Flags().Int64Var(&revisionBump, "bump-revision", 0, "How much to increase the revision of the next etcd transaction ")
```

dusk125:
With this latest update, I've changed the revision-bump flag to only increase the latest revision by the amount so that clients shouldn't see decreasing revisions.
Also, I added the other flag, mark-compacted, to set the latest revision as a scheduled "in progress" compaction so that watchers will probably fail and resync.

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,23 @@ import (
 
 type revision struct {
 	main int64
+	sep  byte
 	sub  int64
 }
 
+// The binary representation of the revision is 17 bytes:
```
serathius, line 27: Dob't think that's true. I remember that there might be 18byte to mark a tombstone. I will need to confirm, but we should not copy the code to parsing/dumping revisions lightly.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
...
 
+// revBump increases the revision, modrevision, and version of every key to force watchers to resync.
+//
+// In order to ensure that watchers don't have a revision that's higher than the restored revision,
+// 'amount' should be a very large number, > 1_000_000.
+//
+// Compaction should be run after this to further help invalidate watcher caches.
+func (s *v3Manager) revBump(amount int64) error {
+	// Don't do anything if there's nothing to do
+	if amount == 0 {
```
serathius, line 344: Such checks should be outside of function or function should be called `maybeRevBump` as it might not do the bump.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
 	return nil
 }
 
+// revBump increases the revision, modrevision, and version of every key to force watchers to resync.
+//
+// In order to ensure that watchers don't have a revision that's higher than the restored revision,
+// 'amount' should be a very large number, > 1_000_000.
+//
+// Compaction should be run after this to further help invalidate watcher caches.
+func (s *v3Manager) revBump(amount int64) error {
```
serathius, line 342:
```suggestion
func (s *v3Manager) bumpRevision(amount int64) error {
```

etcdutl/snapshot/util.go
serathius: https://github.com/etcd-io/etcd/blob/7444985cab75e3d8ef4ef5ffe71634547556aa68/server/storage/mvcc/kvstore.go#L523-L532

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,23 @@ import (
...
+//
+// [8 bytes: main] [1 byte: sep] [8 bytes: sub]
 func bytesToRev(bytes []byte) revision {
```
serathius, line 30: Don't like the idea of copying this code. It's too fragile. Still making this public might not be best solution. Can you  comment that `bytesToRev` and `revToBytes` functions should be synced with function server.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
...
+	if amount == 0 {
+		s.lg.Info("skipping revision bump")
+		return nil
+	}
+
+	be := backend.NewDefaultBackend(s.lg, s.outDbPath())
+	defer be.Close()
+
+	tx := be.BatchTx()
+	tx.LockInsideApply()
```
serathius, line 353: This is not apply function so `LockOutsideApply`

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
...
+
+		if latest.main < rev.main {
+			latest = rev
+			value = v
+		}
+
+		return nil
+	}); err != nil {
+		return err
+	}
```
serathius, line 369: Move to separate function.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
...
+		rev := bytesToRev(k)
+
+		if latest < rev.main {
+			latest = rev.main
+		}
+
+		return nil
+	}); err != nil {
+		return err
+	}
```
serathius, line 400: This code repeats, please move to separate function.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
 	return nil
 }
 
+// revBump increases the revision, modrevision, and version of every key to force watchers to resync.
+//
+// In order to ensure that watchers don't have a revision that's higher than the restored revision,
+// 'amount' should be a very large number, > 1_000_000.
```
serathius, line 339: 1`000`000 is not a large number for revision. Restored backup could have couple of days. 1 QPS for 11 days is enough to hit 1000000. Please suggest at least 1`000`000`000, even with such large bump it's enough to make 9`223`372`036 backups.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +333,79 @@ func (s *v3Manager) saveDB() error {
...
+
+	return nil
+}
+
+func (s *v3Manager) markLatestRevCompacted() error {
+	be := backend.NewDefaultBackend(s.lg, s.outDbPath())
+	defer be.Close()
+
+	tx := be.BatchTx()
+	tx.LockInsideApply()
```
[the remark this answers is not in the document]
dusk125: Looking at [mvcc.SetScheduledCompact](https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/store.go#L38), it does a LockInsideApply then calls `UnsafeSetScheduledCompact`. So for getting the latest revision I don't need LockInsideApply, but for setting the schedule, it looks like I should. 
Should I have two different transactions or just stick with InsideApply?

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,27 @@ import (
...
 type revision struct {
 	main int64
+	sep  byte
 	sub  int64
 }
 
+// The binary representation of the revision is 17 bytes
+// (sometimes 18 bytes see mvcc.isTombstone)
+//
+// [8 bytes: main] [1 byte: sep] [8 bytes: sub]
```
serathius, line 30: Please don't document this, especially here. Don't want to spent too much time on getting this right when we just want to fix restore function. If you want to improve function documentation please create a separate PR.

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,27 @@ import (
...
+//
+// TODO bytesToRev should be synced with function server
```
serathius, line 32: Please provide a proper link to the bytesToRev function in server

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,27 @@ import (
...
 func bytesToRev(bytes []byte) revision {
 	return revision{
 		main: int64(binary.BigEndian.Uint64(bytes[0:8])),
+		sep:  bytes[8],
 		sub:  int64(binary.BigEndian.Uint64(bytes[9:])),
 	}
 }
+
+// TODO revToBytes should be synced with function server
```
serathius, line 41: ditto

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -203,6 +204,16 @@ type RestoreConfig struct {
 	// SkipHashCheck is "true" to ignore snapshot integrity hash value
 	// (required if copied from data directory).
 	SkipHashCheck bool
+
+	// RevisionBump is the amount to increase the latest revision after restore,
+	// to ensure that the revisions never decrease from the clients perspective
```
serathius, line 209:
```suggestion
	// to allow administrators to trick clients into thinking that revision never decreased.
```

serathius: bump revision doesn't ensure anything, it's up to administrators to provide RevisionBump large enough to trick users.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -203,6 +204,16 @@ type RestoreConfig struct {
...
+	// If 0, revision bumping is skipped.
+	RevisionBump int64
+
+	// MarkCompacted is "true" to mark the latest revision as scheduled for compaction.
```
serathius, line 213:
```suggestion
	// MarkCompacted is "true" to mark the latest revision as compacted.
```

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -203,6 +204,16 @@ type RestoreConfig struct {
...
+	// This will cause the server to compact the restored revisions (at some point after startup),
```
serathius, line 214:
```suggestion
	// This will cause the server to compact the restored revisions (at some point after startup),
```
This is implementation detail.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -203,6 +204,16 @@ type RestoreConfig struct {
...
+	// and correctly cancel watchers so they can invalidate their caches.
```
serathius, line 215: Comments are nice, but it would be better to just provide all this information as part of user visible information in flag help section.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
 	return nil
 }
 
+// unsafeGetLatestRevision returns the latest revision without locking the given transaction.
+func (s *v3Manager) unsafeGetLatestRevision(tx backend.BatchTx) (latest revision, err error) {
```
serathius, line 339: Code is read from top to bottom, please move the function to after `markLatestRevCompacted` as it's the least important one.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
 	return nil
 }
 
+// unsafeGetLatestRevision returns the latest revision without locking the given transaction.
```
serathius, line 338:
```suggestion
```

serathius: Whole comment is nicely reflected in the function name. Comments get outdated.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+
+		return nil
+	})
+
+	return
+}
+
+// bumpRevision increases the revision so that, from the client's perspective, the revision never decreases.
+//
+//	If we don't bump the revision, a client looking for a key at one revision finds that the key is actually at a
```
serathius, line 355: Don't put feature documentation within a small function. This documentation should be user facing not be a comment that no user will see.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+//	smaller revision which breaks the promise that revisions never decrease. The client has no way of knowing that
+//	etcd was restored, it only sees the decrease in revision.
+//
+// To ensure that clients don't have a revision that's higher than the restored revision,
```
serathius, line 359: ditto.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+// 'amount' should be a very large number, >= 1_000_000_000.
+//
+// Compaction should be run after this to further help invalidate watcher caches.
```
serathius, line 362: This is incorrect behavior as those flags are independent. Please implement bumpRevision in a way that doesn't require `markLatestRevCompacted` or add a validation that forbids using `bumpRevision` without using `markLatestRevCompacted`.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+// (see markLatestRevCompacted)
+func (s *v3Manager) bumpRevision(amount int64) error {
+	be := backend.NewDefaultBackend(s.lg, s.outDbPath())
```
serathius, line 365: Please open the backed only once.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+	defer be.Close()
```
serathius, line 366: Please call `ForceCommit` before close.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+	revToBytes(k[:], latest)
+	tx.UnsafePut(schema.Key, k[:], []byte{})
+
+	return nil
+}
+
+// markLatestRevCompacted sets the scheduled compaction to the latest revision in order to tell the server
+// that it should compact the restored database up to the latest revision.
+//
+// This will cause any watcher to be cancelled as the revision they're watching will be compacted,
```
serathius, line 395: This is restore feature documentation, not a good much for a low level function.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -265,6 +276,19 @@ func (s *v3Manager) Restore(cfg RestoreConfig) error {
 	if err = s.saveDB(); err != nil {
 		return err
 	}
+
+	if cfg.RevisionBump > 0 {
```
serathius, line 280: Both bumpRevision and markLatestRevCompacted should be done in one transaction so it can be reverted on error.

serathius:
Please add e2e test.

etcdutl/snapshot/v3_snapshot.go
chaochn47: +1. Sounds like `bumpRevision` + `markLatestRevCompacted` should be used together to mark user last observed revision is compacted.

etcdutl/snapshot/v3_snapshot.go
tjungblu: as commented earlier, I think this is too focused on K8s as a consumer. In both places, I would leave that information out.

etcdutl/snapshot/v3_snapshot.go
tjungblu, line 280: I think `!= 0` would be more idiomatic. I'm not sure if we even want to allow to pass negative numbers in the first place?

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+
+	s.lg.Info(
+		"bumping latest revision to ensure revisions never decrease",
+		zap.Int64("latest-revision", latest.main),
+		zap.Int64("bump-amount", amount),
+		zap.Int64("new-latest-revision", latest.main+amount),
+	)
+
+	latest.main += amount
+	k := [17]byte{}
```
tjungblu, line 385: also a good place to rather make the revision.go public and reuse `revBytesLen`

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +335,88 @@ func (s *v3Manager) saveDB() error {
...
+	tx.LockOutsideApply()
+	defer tx.Unlock()
+
+	latest, err := s.unsafeGetLatestRevision(tx)
+	if err != nil {
+		return err
+	}
+
+	s.lg.Info(
+		"bumping latest revision to ensure revisions never decrease",
```
tjungblu, line 378: I think a short statement of `bumping latest revision` is already enough, you don't need to justify everything  :)

etcdutl/snapshot/v3_snapshot.go
dusk125: Agreed that if we bumpRevision, we want to compact. Can either of you foresee wanting to compact without bumping: should that be an option?

etcdutl/snapshot/v3_snapshot.go
dusk125: Yeah, I can't think of a reason that we would want to allow someone to decrease the revision, especially since the bump is there to help make sure that doesn't happen.

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,27 @@ import (
...
+func revToBytes(bytes []byte, rev revision) {
```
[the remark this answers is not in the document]
serathius: It's out of scope.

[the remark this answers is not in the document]
dusk125: I can handle that in another PR so the refactor doesn't muddy this PR, it would be nice to remove the duplication.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +327,73 @@ func (s *v3Manager) saveDB() error {
 	return nil
 }
 
+// maybeModifyLatestRevision can increase the latest revision by the given amount and sets the scheduled compaction
+// to that revision so that the server will compact the database up to that revision.
```
serathius, line 331:
```suggestion
// to that revision so that the server will consider this revision compacted.
```

etcdutl/snapshot/util.go
```diff
@@ -20,12 +20,24 @@ import (
 
 type revision struct {
 	main int64
+	sep  byte
```
serathius, line 23: Why add `sep`?  Not needed

serathius: This struct should be kept synched with revision from https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/revision.go

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +327,73 @@ func (s *v3Manager) saveDB() error {
...
+
+	mvcc.UnsafeSetScheduledCompact(tx, latest.main)
+}
+
+func (s *v3Manager) unsafeGetLatestRevision(tx backend.BatchTx) (revision, error) {
+	var latest revision
+	err := tx.UnsafeForEach(schema.Key, func(k, _ []byte) (err error) {
+		rev := bytesToRev(k)
+
+		if latest.main < rev.main {
```
serathius, line 388: This is not the latest revision as you are not considering subrevision. Either expose main revision as int64 or implement a proper `unsafeGetLatestRevision`

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +327,73 @@ func (s *v3Manager) saveDB() error {
...
+	return latest
+}
+
+func (s *v3Manager) unsafeMarkRevisionCompacted(tx backend.BatchTx, latest revision) {
+	s.lg.Info(
+		"marking revision compacted",
+		zap.Int64("revision", latest.main),
+	)
+
+	mvcc.UnsafeSetScheduledCompact(tx, latest.main)
```
serathius, line 380: Please make sure to get semantics of subrevision right. I think to mark revision as compacted (X,Y) we put (X+1, 0) as compacted.

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +327,73 @@ func (s *v3Manager) saveDB() error {
...
+
+func (s *v3Manager) unsafeBumpRevision(tx backend.BatchTx, latest revision, amount int64) revision {
+	s.lg.Info(
+		"bumping latest revision",
+		zap.Int64("latest-revision", latest.main),
+		zap.Int64("bump-amount", amount),
+		zap.Int64("new-latest-revision", latest.main+amount),
+	)
+
+	latest.main += amount
```
serathius, line 366: Please make sure to get semantics of subrevision right. I think to bumping revision (X,Y) by N should result in revision (X+N, 0).

etcdutl/snapshot/v3_snapshot.go
dusk125: Do you mean that you would put an additional revision at X+1 then pass that into SetScheduledCompact or pass in X+1 to SetScheduledCompact? The latter at least doesn't work and causes a future revision error on startup.

dusk125:
That last test failure was a flake, sometimes the watch would take a tad too long to start and miss the first put. I've fixed this in the latest iteration with a short sleep after starting the watch: I ran the current iteration a couple thousand times with no more flakes (beforehand it was at least 1 in 100 flaky).

tests/e2e/ctl_v3_snapshot_test.go
```diff
@@ -305,3 +308,126 @@ func snapshotVersionTest(cx ctlCtx) {
...
+		require.Equal(t, int64(currentRev), v.Header.Revision)
+	}
+
+	fpath := filepath.Join(t.TempDir(), "test.snapshot")
+
+	t.Log("etcdctl saving snapshot...")
+	cmdPrefix := []string{e2e.BinPath.Etcdctl, "--endpoints", strings.Join(epc.EndpointsGRPC(), ",")}
+	require.NoError(t, e2e.SpawnWithExpects(append(cmdPrefix, "snapshot", "save", fpath), nil, fmt.Sprintf("Snapshot saved at %s", fpath)))
+
+	t.Log("Stopping the original server...")
```
serathius, line 364: Please add some writes after snapshot but before the stopping of the server. We should test that writes after snapshot are lost and watch correctly behaved thanks to revision bump.

tests/e2e/ctl_v3_snapshot_test.go
```diff
@@ -305,3 +308,126 @@ func snapshotVersionTest(cx ctlCtx) {
...
+	}, "added member")
+	require.NoError(t, err)
+
+	t.Log("(Re)starting the etcd member using the restored snapshot...")
+	epc.Procs[0].Config().DataDirPath = newDataDir
+	for i := range epc.Procs[0].Config().Args {
+		if epc.Procs[0].Config().Args[i] == "--data-dir" {
+			epc.Procs[0].Config().Args[i+1] = newDataDir
+		}
+	}
```
serathius, line 390: Why not delete old data directory and restore into it?

tests/e2e/ctl_v3_snapshot_test.go
```diff
@@ -305,3 +308,126 @@ func snapshotVersionTest(cx ctlCtx) {
...
+		require.Equal(t, kvs[i].Val, string(v.Kvs[0].Value))
+		// the currentRev we had before the snapshot should be bumped by now
+		require.Equal(t, int64(bumpAmount+currentRev), v.Header.Revision)
+		// the existing revisions and versions shouldn't be touched by the restore process
+		require.Equal(t, int64(baseRev+i), v.Kvs[0].CreateRevision)
+		require.Equal(t, int64(baseRev+i), v.Kvs[0].ModRevision)
+		require.Equal(t, int64(1), v.Kvs[0].Version)
+	}
+
+	cancelResult := <-watchCh
```
serathius, line 408: Can you also test that `watchCh` is closed?

tests/e2e/ctl_v3_snapshot_test.go
```diff
@@ -305,3 +308,126 @@ func snapshotVersionTest(cx ctlCtx) {
...
+	require.NoError(t, epc.Restart(context.Background()))
+
+	t.Log("Ensuring the restored member has the correct data...")
+	for i := range kvs {
+		v, err := ctl.Get(context.Background(), kvs[i].Key, config.GetOptions{})
+		require.NoError(t, err)
+		require.Equal(t, int64(1), v.Count)
+		require.Equal(t, kvs[i].Val, string(v.Kvs[0].Value))
+		// the currentRev we had before the snapshot should be bumped by now
+		require.Equal(t, int64(bumpAmount+currentRev), v.Header.Revision)
```
serathius, line 401: Testing header revision is not important here. Aside of that data should match one from before, so we could refactor it into separate function.

tests/e2e/ctl_v3_snapshot_test.go
```diff
@@ -305,3 +308,126 @@ func snapshotVersionTest(cx ctlCtx) {
...
+	epc.Procs[0].Config().DataDirPath = newDataDir
+	for i := range epc.Procs[0].Config().Args {
+		if epc.Procs[0].Config().Args[i] == "--data-dir" {
+			epc.Procs[0].Config().Args[i+1] = newDataDir
+		}
+	}
+
+	require.NoError(t, epc.Restart(context.Background()))
+
+	t.Log("Ensuring the restored member has the correct data...")
```
serathius, line 394: Please add validation that user can no longer request old revisions from before restore. Like revisions that were lost.

tests/e2e/ctl_v3_snapshot_test.go
tjungblu: to be fair, I've copy&pasted from the testcase above this one yesterday. I assume it was easier to debug the older data if anything goes wrong, but can't tell what the original authors intention here was.

tests/e2e/ctl_v3_snapshot_test.go
tjungblu: it's not important, but I don't really see how those assertions are hurting. Good point on the function though, @dusk125 do you want to take this on from here? I need to start working on the other stuff today.

tests/e2e/ctl_v3_snapshot_test.go
serathius: Ok, maybe let's have this as a followup to avoid inconsistent code.

tjungblu (approved):
/lgtm (nonbinding)

serathius (approved):

serathius:
cc @ahrtr @ptabor @spzala @mitake

etcdutl/snapshot/v3_snapshot.go
ptabor: Wasn't this about watch ?

tests/framework/e2e/etcdctl.go
```diff
@@ -698,6 +698,9 @@ func (ctl *EtcdctlV3) Watch(ctx context.Context, key string, opts config.WatchOp
 					var resp clientv3.WatchResponse
 					json.Unmarshal([]byte(line), &resp)
 					if resp.Canceled {
+						if opts.IncludeCancellationResponse {
+							ch <- resp
```
ptabor, line 702: Why this change. Please add a comment .

ptabor (approved):

tests/framework/e2e/etcdctl.go
tjungblu: the cancellation response was swallowed before and existing tests are not expecting this result. We need to assert that it is returned in: 
https://github.com/etcd-io/etcd/pull/16029/files#diff-d99fed41bb3f2962001bb422375c286f00cf1e1e9f57d8bd596f11bde187dea7R399-R405

serathius: I think this is because etcdctl doesn't print cancel response, just exists.

serathius: Oh, forgot to comment on this. Not passing cancel response is just oversight when implementing etcdctl struct. This should happen by default. 

@tjungblu Have you tested just passing it by default? Does it cause other test to fail?

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -311,6 +327,75 @@ func (s *v3Manager) saveDB() error {
...
+	latest.sub = 0
+	k := [17]byte{}
+	revToBytes(k[:], latest)
+	tx.UnsafePut(schema.Key, k[:], []byte{})
```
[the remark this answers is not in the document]
dusk125: Because k is a fixed-sized array and both functions take in a slice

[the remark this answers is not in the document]
ahrtr: Got it, then I would suggest to change line 369 to `k := make([]byte, 17)`.  Your current code works, but it's less readable.

tests/framework/e2e/etcdctl.go
dusk125: Doesn't look like making that the default broke anything, I'll make this change

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -265,6 +274,13 @@ func (s *v3Manager) Restore(cfg RestoreConfig) error {
 	if err = s.saveDB(); err != nil {
 		return err
 	}
+
+	if cfg.MarkCompacted || cfg.RevisionBump > 0 {
```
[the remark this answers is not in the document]
dusk125: I'm not sure that's necessarily the case; we need to compact if we're bumping the revision, but if we're not bumping the revision, the user may still want to mark everything in the restore as compacted. That was at least my understanding as to why to have them be separate options. @serathius thoughts?

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -38,6 +38,8 @@ var (
 	restorePeerURLs     string
 	restoreName         string
 	skipHashCheck       bool
+	markCompacted       bool
+	revisionBump        int64
```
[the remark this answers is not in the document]
ahrtr: You do not make the suggested change, nor add any comment, why do you resolve the comment directly?

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -75,6 +77,8 @@ func NewSnapshotRestoreCommand() *cobra.Command {
 	cmd.Flags().StringVar(&restorePeerURLs, "initial-advertise-peer-urls", defaultInitialAdvertisePeerURLs, "List of this member's peer URLs to advertise to the rest of the cluster")
 	cmd.Flags().StringVar(&restoreName, "name", defaultName, "Human-readable name for this member")
 	cmd.Flags().BoolVar(&skipHashCheck, "skip-hash-check", false, "Ignore snapshot integrity hash value (required if copied from data directory)")
+	cmd.Flags().Int64Var(&revisionBump, "bump-revision", 0, "How much to increase the latest revision after restore")
```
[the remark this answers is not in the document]
ahrtr: ditto

etcdutl/snapshot/v3_snapshot.go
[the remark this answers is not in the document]
ahrtr: > but if we're not bumping the revision, the user may still want to mark everything in the restore as compacted

Why? What this PR does is already a little hacky, why do we need to provide more flexibility? I do not see any real use case, but it provides flexibility for users to make mistake.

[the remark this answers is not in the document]
serathius: Separate flags for clarity, but I agree with @ahrtr that we don't need to make them independent. It would only require more work to test all edge cases.

etcdutl/etcdutl/snapshot_command.go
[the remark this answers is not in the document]
dusk125: I have been marking simple changes like this resolved as I handle them locally, so I can keep track of what I've fixed. This, and the below, changes will be in my next update.

etcdutl/etcdutl/snapshot_command.go
```diff
@@ -110,12 +114,19 @@ func SnapshotRestoreCommandFunc(restoreCluster string,
...
 	skipHashCheck bool,
+	revisionBump uint64,
+	markCompacted bool,
 	args []string) {
 	if len(args) != 1 {
 		err := fmt.Errorf("snapshot restore requires exactly one argument")
 		cobrautl.ExitWithError(cobrautl.ExitBadArgs, err)
 	}
 
+	if (revisionBump == 0 && markCompacted) || (revisionBump > 0 && !markCompacted) {
```
ahrtr, line 125: Do we support only bumping revision, and not marking compacted?

dusk125: No, we only support bumping revision and marking compacted or not bumping revision and not marking compacted. As Marek said above, the flags are separate for clarity, but they need to be used together.

dusk125:
I think that test failure may be a flake, I'm retesting locally to see if it'll reproduce

etcdutl/etcdutl/snapshot_command.go
dusk125: @serathius thoughts?

serathius: Don't understand the argument presented by @ahrtr, I think we are getting into non important details of personal preference. I would recommend to keep the flags separate as it's more clear to user and it's easier later to remove a flag then split the flag if we would decide to support bumping revision without marking it as compacted.

ahrtr: Don't think any user-facing change is non-important. Marking compact should never be a must. I am evening think why do we need it on second thought. Please make sure you understand my comment above and let me know if you have confusion.

serathius: Please read and comment on the original proposal https://github.com/kubernetes/kubernetes/issues/118501, not spread the discussion and design options in just before merging the code just because of "second thought". If we need to change the design, then let's discuss changing the design.

ahrtr: There is no change on the design per my previous comment https://github.com/etcd-io/etcd/pull/16029#discussion_r1243910953.  The reason for raising the comment is to make sure we have thought about other possibilities.

This PR fall into the case 2.2 mentioned above, which I don't strongly disagree. If we decide to follow this way, then we may consider removing "`--mark-compact`", and do it (mark compaction) automatically when bumping the revision. Obviously it's clearer & easier to users.

serathius: The design https://github.com/kubernetes/kubernetes/issues/118501 proposed your case 2.2, two points `Etcd restore operation should support bumping revision by an arbitrary number` and `
Etcd restore operation should support marking old revisions as compacted` are separated and are not a single point. Meaning those are independent.

ahrtr: Unfortunately we don't have enough active maintainers and/or reviewers. Each user facing change or etcd core change should have enough reviews and feedback beforing rejecting or merging.

For this case, let's go for 2.2, and let's be aware that  `The only purpose is to make the users be explicitly aware of the operation` . Otherwise, when we are challenged by users, at least we know how to respond.

ahrtr (approved):
LGTM

Thanks @dusk125

etcdutl/etcdutl/snapshot_command.go
serathius: Ok, could we clarify what the action for @dusk125 ? Is this comment addressed, or they should change something?

ahrtr:
Two left actions:
- Please raise a followup ticket for moving some revision related common functions to package "`pkg`", see https://github.com/etcd-io/etcd/pull/16029#discussion_r1242315620
- Please raise a followup ticket for documentation for the flag "--mark-compact". 
  - It isn't allowed to mark compact when users don't bump revision;
  - If users bump revision, then users must specify the flag. The purpose is to make sure users are explicitly aware of the operation (mark compaction). (Note that if we receive too many negative feedback from users, we may consider to remove the flag, and go for 2.1 in future).

dusk125:
> Two left actions:
> 
> * Please raise a followup ticket for moving some revision related common functions to package "`pkg`", see [Adding optional revision bump to snapshot restore #16029 (comment)](https://github.com/etcd-io/etcd/pull/16029#discussion_r1242315620)
> * Please raise a followup ticket for documentation for the flag "--mark-compact".
>   
>   * It isn't allowed to mark compact when users don't bump revision;
>   * If users bump revision, then users must specify the flag. The purpose is to make sure users are explicitly aware of the operation (mark compaction). (Note that if we receive too many negative feedback from users, we may consider to remove the flag, and go for 2.1 in future).

Tickets have been raised

tjungblu:
While @dusk125 is out, I can take over the 3.5 backport. Can anyone hit the merge button? Thanks.

PR #16035, opened 2023-06-08
lessor: fix le.itemMap leak after lease id revoked

qsyqian:
When lease is revoked, we should delete item in le.itemMap.

### DIFF
--- a/tests/robustness/model/replay.go
+++ b/tests/robustness/model/replay.go
@@ -36,16 +36,14 @@ func NewReplay(persistedRequests []EtcdRequest) *EtcdReplay {
 		state = newState
 	}
 	return &EtcdReplay{
-		Requests:            persistedRequests,
 		revisionToEtcdState: revisionToEtcdState,
-		events:              events,
+		Events:              events,
 	}
 }
 
 type EtcdReplay struct {
-	Requests            []EtcdRequest
 	revisionToEtcdState []EtcdState
-	events              []PersistedEvent
+	Events              []PersistedEvent
 }
 
 func (r *EtcdReplay) StateForRevision(revision int64) (EtcdState, error) {
@@ -56,7 +54,7 @@ func (r *EtcdReplay) StateForRevision(revision int64) (EtcdState, error) {
 }
 
 func (r *EtcdReplay) EventsForWatch(watch WatchRequest) (events []PersistedEvent) {
-	for _, e := range r.events {
+	for _, e := range r.Events {
 		if e.Revision < watch.Revision || !e.Match(watch) {
 			continue
 		}
--- a/tests/robustness/validate/validate_test.go
+++ b/tests/robustness/validate/validate_test.go
@@ -1332,34 +1332,6 @@ func TestValidateWatch(t *testing.T) {
 			},
 		},
 		{
-			name: "Reliable - issue #18089 - pass",
-			reports: []report.ClientReport{
-				{
-					Watch: []model.WatchOperation{
-						{
-							Request: model.WatchRequest{
-								WithPrefix: true,
-								Revision:   3,
-							},
-							Responses: []model.WatchResponse{
-								{
-									Events: []model.WatchEvent{
-										putWatchEvent("b", "2", 4, true),
-									},
-								},
-							},
-						},
-					},
-				},
-			},
-			persistedRequests: []model.EtcdRequest{
-				putRequest("a", "1"),
-				deleteRequest("a"),
-				putRequest("b", "2"),
-				compactRequest(3),
-			},
-		},
-		{
 			name: "Resumable - watch revision from middle event - pass",
 			reports: []report.ClientReport{
 				{
@@ -1985,14 +1957,5 @@ func deleteRequest(key string) model.EtcdRequest {
 			OperationsOnFailure: nil,
 		},
 		Defragment: nil,
-	}
-}
-
-func compactRequest(revision int64) model.EtcdRequest {
-	return model.EtcdRequest{
-		Type: model.Compact,
-		Compact: &model.CompactRequest{
-			Revision: revision,
-		},
 	}
 }
--- a/tests/robustness/validate/watch.go
+++ b/tests/robustness/validate/watch.go
@@ -205,11 +205,6 @@ func validateReliable(lg *zap.Logger, replay *model.EtcdReplay, report report.Cl
 				gotEvents = append(gotEvents, event.PersistedEvent)
 			}
 		}
-		// TODO(https://github.com/etcd-io/etcd/issues/18089): Remove when bug is fixed
-		detected, newWantEvents := checkIssue18089(lg, report.ClientID, wantEvents, gotEvents, replay)
-		if detected {
-			wantEvents = newWantEvents
-		}
 		if diff := cmp.Diff(wantEvents, gotEvents, cmpopts.IgnoreFields(model.PersistedEvent{}, "IsCreate")); diff != "" {
 			lg.Error("Broke watch guarantee", zap.String("guarantee", "reliable"), zap.Int("client", report.ClientID), zap.String("diff", diff))
 			err = errBrokeReliable
@@ -223,22 +218,18 @@ func validateResumable(lg *zap.Logger, replay *model.EtcdReplay, report report.C
 		if watch.Request.Revision == 0 {
 			continue
 		}
-		wantEvents := replay.EventsForWatch(watch.Request)
+		events := replay.EventsForWatch(watch.Request)
 		index := 0
-		for index < len(wantEvents) && (wantEvents[index].Revision < watch.Request.Revision || !wantEvents[index].Match(watch.Request)) {
+		for index < len(events) && (events[index].Revision < watch.Request.Revision || !events[index].Match(watch.Request)) {
 			index++
 		}
-		if index == len(wantEvents) {
+		if index == len(events) {
 			continue
 		}
-		gotFirstEvent := firstWatchEvent(watch)
+		firstEvent := firstWatchEvent(watch)
 		// If watch is resumable, first event it gets should the first event that happened after the requested revision.
-		if gotFirstEvent != nil && wantEvents[index] != gotFirstEvent.PersistedEvent {
-			// TODO(https://github.com/etcd-io/etcd/issues/18089): Remove when bug is fixed
-			if detected, _ := checkIssue18089(lg, report.ClientID, wantEvents, []model.PersistedEvent{gotFirstEvent.PersistedEvent}, replay); detected {
-				continue
-			}
-			lg.Error("Broke watch guarantee", zap.String("guarantee", "resumable"), zap.Int("client", report.ClientID), zap.Any("request", watch.Request), zap.Any("got-event", *gotFirstEvent), zap.Any("want-event", wantEvents[index]))
+		if firstEvent != nil && events[index] != firstEvent.PersistedEvent {
+			lg.Error("Broke watch guarantee", zap.String("guarantee", "resumable"), zap.Int("client", report.ClientID), zap.Any("request", watch.Request), zap.Any("got-event", *firstEvent), zap.Any("want-event", events[index]))
 			err = errBrokeResumable
 		}
 	}
@@ -340,42 +331,4 @@ func firstWatchEvent(op model.WatchOperation) *model.WatchEvent {
 		}
 	}
 	return nil
-}
-
-func checkIssue18089(lg *zap.Logger, clientID int, want, got []model.PersistedEvent, replay *model.EtcdReplay) (bool, []model.PersistedEvent) {
-	type keyRevision struct {
-		Key      string
-		Revision int64
-	}
-	gotKeyRevision := map[keyRevision]struct{}{}
-	for _, event := range got {
-		gotKeyRevision[keyRevision{
-			Key:      event.Key,
-			Revision: event.Revision,
-		}] = struct{}{}
-	}
-	newWant := []model.PersistedEvent{}
-	issueDetected := false
-	for _, event := range want {
-		_, found := gotKeyRevision[keyRevision{
-			Key:      event.Key,
-			Revision: event.Revision,
-		}]
-		if !found && event.Type == model.DeleteOperation && matchingCompaction(event.Revision, replay) {
-			issueDetected = true
-			lg.Info("Detected issue 18089 still present, missing delete watch event for a compacted revision", zap.Int("client", clientID), zap.Any("missing-event", event))
-			continue
-		}
-		newWant = append(newWant, event)
-	}
-	return issueDetected, newWant
-}
-
-func matchingCompaction(revision int64, replay *model.EtcdReplay) bool {
-	for _, req := range replay.Requests {
-		if req.Type == model.Compact && req.Compact.Revision == revision {
-			return true
-		}
-	}
-	return false
 }


server/lease/lessor.go
```diff
@@ -572,6 +572,9 @@ func (le *lessor) Detach(id LeaseID, items []LeaseItem) error {
 
 	l := le.leaseMap[id]
 	if l == nil {
+		for _, it := range items {
+			delete(le.itemMap, it)
+		}
```
tjungblu, line 577: it does that, exactly 5 lines below

qsyqian: thx for review. when lease is revoked, in le.Revoke it will delete le.leaseMap , and then delete key in . when delete key, it will find one lease use le.GetLease use leaseItem. and call le.Detach. for now, the leaseMap does not contain the lease,  there is no chance to delete le.itemMap.

tjungblu: Ah, there's a return before. Apologies, that makes sense.

tjungblu:
I've been trying to repro this a bit further with a testcase. 

It seems when you do a grant -> attach -> revoke, the keys attached are not freed from the itemMap. Is this the same leakage you've observed?


just for reference on further discussion:

```go
func TestLessorGrantAttachRevokeDetach(t *testing.T) {
	lg := zap.NewNop()
	dir, be := NewTestBackend(t)
	defer os.RemoveAll(dir)
	defer be.Close()

	le := newLessor(lg, be, clusterLatest(), LessorConfig{MinLeaseTTL: minLeaseTTL})
	defer le.Stop()
	le.SetRangeDeleter(func() TxnDelete { return newFakeDeleter(be) })

	// grant a lease with long term (100 seconds) to
	// avoid early termination during the test.
	l, err := le.Grant(1, 100)
	if err != nil {
		t.Fatalf("could not grant lease for 100s ttl (%v)", err)
	}

	items := []LeaseItem{
		{"foo"},
		{"bar"},
	}

	if err := le.Attach(l.ID, items); err != nil {
		t.Fatalf("failed to attach items to the lease: %v", err)
	}

	if err := le.Revoke(l.ID); err != nil {
		t.Fatalf("failed to revoke lease: %v", err)
	}

	err = le.Detach(l.ID, items[0:1])
	assert.Equal(t, ErrLeaseNotFound, err)
	assert.Equal(t, 0, len(le.leaseMap))
	assert.Equal(t, 2, len(le.itemMap)) <-- leak
}

```

qsyqian:
@tjungblu almost yes, when lease is revoked, the keys attached with it will delete forever, and the lessor.ItemMap will never delete them.
But i have little confused because i am not similar to the ut code. The ut you provide has something different with real env.
In real env, the RangeDeleter.DeleteRange will call le.Detach. so when revoked, all the keyItem will be called with  le.Detach. No need to call `err = le.Detach(l.ID, items[0:1])`. 
In ut, it seems that should call le.Detach after le.Revoke.

Use my code, it will PASS just like:
        err = le.Detach(l.ID, items)
	assert.Equal(t, ErrLeaseNotFound, err)
	assert.Equal(t, 0, len(le.leaseMap))
	assert.Equal(t, 0, len(le.itemMap))

tjungblu:
I assume the rangeDeleter uses this code below?

https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/kvstore_txn.go#L304-L315

Looking from the other angle of the revoke, the lease in the rangeDeleter won't find the lease anymore because it was already deleted earlier?

https://github.com/etcd-io/etcd/blob/main/server/lease/lessor.go#L330-L344

qsyqian:
> Looking from the other angle of the revoke, the lease in the rangeDeleter won't find the lease anymore because it was already deleted earlier?

it seems not. in Revoke it delete leaseMap, but in the rangeDeleter, it call GetLease with LeaseItem. In GetLease, it find lease from itemMap.

qsyqian:
@ahrtr thx for review. @tjungblu i will add the ut code you provide above if you don't mind. :)

qsyqian:
@ahrtr @tjungblu updated, PTAL.

server/lease/lessor_test.go
```diff
@@ -208,6 +209,43 @@ func TestLessorRevoke(t *testing.T) {
 	}
 }
 
+func TestLessorGrantAttachRevokeDetach(t *testing.T) {
+	lg := zap.NewNop()
+	dir, be := NewTestBackend(t)
+	defer os.RemoveAll(dir)
```
ahrtr, line 215: nit: this is unnecessary, because the data directory is created via `t.TempDir()`, and it will be cleanuped automatically when the test finishes.

ahrtr:
This fix needs to be backported to 3.4 and 3.5.

Could you raise an issue to track this bug fix? We keep the issue open until all  backport and changelog are done. We see many times that bug fixes were forgotten to be backported.

ahrtr (approved):
LGTM

Thank you @qsyqian

server/lease/lessor.go
```diff
@@ -572,6 +572,9 @@ func (le *lessor) Detach(id LeaseID, items []LeaseItem) error {
 
 	l := le.leaseMap[id]
 	if l == nil {
+		for _, it := range items {
+			delete(le.itemMap, it)
```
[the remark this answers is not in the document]
serathius: OMG, this code.... aaaa 😵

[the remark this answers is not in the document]
serathius: Can we add a unit test that will verify integration between lessor and KV store? Because of tight coupling of those two it doesn't make sense to unit test lessor by itself.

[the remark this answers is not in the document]
serathius: We have a circular dependency because lack of hierarchy. Looks like that lessor is part of KVstore, meaning it should be higher in hierarchy. Can you please either:
* Replace `fakeLessor` in KVstore to be real lessor. 
* Add new tests to KVstore that use real lessor.

What we should is testing an invariant that call flows:
* `Grant`, `Attach`, `Revoke` and
* `Grant`, `Attach`, `Detach`

Don't leak any objects, aka `itemSet` is empty.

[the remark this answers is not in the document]
serathius: @jmhbnz @lavacat Looks like a place for a nice refactor, move lessor under KVstore to enforce hierarchy. Also I remember old proposal to split lessor API based on whether is called from API or from apply flow.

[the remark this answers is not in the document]
qsyqian: > We have a circular dependency because lack of hierarchy. Looks like that lessor is part of KVstore, meaning it should be higher in hierarchy. Can you please either:
> 
> * Replace `fakeLessor` in KVstore to be real lessor.
> * Add new tests to KVstore that use real lessor.
> 
> What we should is testing an invariant that call flows:
> 
> * `Grant`, `Attach`, `Revoke` and
> * `Grant`, `Attach`, `Detach`
> 
> Don't leak any objects, aka `itemSet` is empty.

thx, i will try it later.

chaochn47:
FWIW, if you run the e2e test, it should reproduce the issue with the following error message. It does not fail the test and we could possibly use log assertion to guarantee detach in RangeDeleter succeeds with no error. 

`cd tests/common && EXPECT_DEBUG=true go test -v -run "^TestLeaseGrantTimeToLiveExpired/NoTLS$" --tags e2e`

```
"msg":"failed to detach old lease from a key","error":"lease not found"
```

> Support for assigning time to live to a key during a write - used for [master leases](https://github.com/kubernetes/kubernetes/blob/403a8fdf1e81737f97196928801f45cf16ed01cb/pkg/controlplane/instance.go#L273) and limiting growth of Kubernetes Events

Trying to understand deeply about this issue in k8s etcd usage, why I don't see those lease not found error messages in production etcd clusters log scanning by past 1 day? == (We are using etcd 3.5.7). I expect event TTL expired will trigger this issue. 

Okay, [release-3.5 lessor](https://github.com/etcd-io/etcd/blob/release-3.5/server/lease/lessor.go#L343-L351) and [release-3.4](https://github.com/etcd-io/etcd/blob/release-3.4/lease/lessor.go#L342-L352) is confirmed not impacted. Only main branch is impacted introduced by https://github.com/etcd-io/etcd/pull/14080.

qsyqian:
> FWIW, if you run the e2e test, it should reproduce the issue with the following error message. It does not fail the test and we could possibly use log assertion to guarantee detach in RangeDeleter succeeds with no error.
> 
> `cd tests/common && EXPECT_DEBUG=true go test -v -run "^TestLeaseGrantTimeToLiveExpired/NoTLS$" --tags e2e`
> 
> ```
> "msg":"failed to detach old lease from a key","error":"lease not found"
> ```
> 
> > Support for assigning time to live to a key during a write - used for [master leases](https://github.com/kubernetes/kubernetes/blob/403a8fdf1e81737f97196928801f45cf16ed01cb/pkg/controlplane/instance.go#L273) and limiting growth of Kubernetes Events
> 
> Trying to understand deeply about this issue in k8s etcd usage, why I don't see those lease not found error messages in production etcd clusters log scanning by past 1 day? == (We are using etcd 3.5.7). I expect event TTL expired will trigger this issue.
> 
> Okay, [release-3.5 lessor](https://github.com/etcd-io/etcd/blob/release-3.5/server/lease/lessor.go#L343-L351) and [release-3.4](https://github.com/etcd-io/etcd/blob/release-3.4/lease/lessor.go#L342-L352) is confirmed not impacted. Only main branch is impacted introduced by #14080.
> 
> I can remove the backport tag if we are on the same page :p

yes, I think you are right. In `Revoke()`, if `delete(le.leaseMap, l.ID)`  is behind `txn.DeleteRange([]byte(key), nil)`, i will not has memory leak. I think we can remove the backport tag.

qsyqian:
@serathius added two ut for kv store. Also add one `Items()` function for `Lessor` interface. because it seems there is no way to get `ltemMap` out of package lease. PTAL

server/storage/mvcc/kvstore_test.go
```diff
@@ -178,6 +178,186 @@ func TestStorePut(t *testing.T) {
...
+		itemSet = leaseAfterDelete.Keys()
+
+		if !reflect.DeepEqual(itemSet, wantItemSet) {
+			t.Errorf("#%d: leaseAfterDelete.Keys() = %+v, want %+v", i, itemSet, wantItemSet)
+		}
+
+		itemMap = s.le.Items()
+		wantItemMap = []string{}
+		if !reflect.DeepEqual(itemMap, wantItemMap) {
+			t.Errorf("#%d: lessor.Items() = %+v, want %+v", i, itemMap, wantItemMap)
```
tjungblu, line 263: maybe you can get away easier by just checking with `GetLease`? 

> assert.Equal(t, lease.NoLease, s.le.GetLease(lease.LeaseItem{Key: "foo"}))

That way we don't have to create another func on the lease api surface. Just to make everybody elses lives easier when we want to refactor it.

server/storage/mvcc/kvstore_test.go
```diff
@@ -178,6 +178,186 @@ func TestStorePut(t *testing.T) {
...
+		leaseAfterPut := s.le.Lookup(l.ID)
+		wantItemSet := []string{"foo"}
+		itemSet := leaseAfterPut.Keys()
+		if !reflect.DeepEqual(itemSet, wantItemSet) {
+			t.Errorf("#%d: leaseAfterPut.Keys() = %+v, want %+v", i, itemSet, wantItemSet)
+		}
+
+		itemMap := s.le.Items()
+		wantItemMap := []string{"foo"}
+		if !reflect.DeepEqual(itemMap, wantItemMap) {
```
tjungblu, line 241: Similar to the comment below, this would become:

> assert.Equal(t, l.ID, s.le.GetLease(lease.LeaseItem{Key: "foo"}))

server/lease/lessor_test.go
```diff
@@ -208,6 +209,42 @@ func TestLessorRevoke(t *testing.T) {
...
+
+	items := []LeaseItem{
+		{"foo"},
+		{"bar"},
+	}
+
+	if err := le.Attach(l.ID, items); err != nil {
+		t.Fatalf("failed to attach items to the lease: %v", err)
+	}
+
```
tjungblu, line 236: Thanks for the co-authorship, maybe you can add another assertion here that it was indeed put into the itemMap:

> 	assert.Equal(t, 1, len(le.leaseMap))
>	assert.Equal(t, 2, len(le.itemMap))

so that way we can say it was actually deleted, otherwise it might have not been there in the first place :safety_pin:

server/storage/mvcc/kvstore_test.go
qsyqian: updated

server/storage/mvcc/kvstore_test.go
qsyqian: updated

server/lease/lessor_test.go
qsyqian: updated

server/storage/mvcc/kvstore_test.go
```diff
@@ -178,6 +178,149 @@ func TestStorePut(t *testing.T) {
...
+		ModRevision:    2,
+		Version:        1,
+		Lease:          1,
+	}
+	kvb, err := kv.Marshal()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tests := []struct {
```
serathius, line 197: Why table test when there is only 1 test case?

server/storage/mvcc/kvstore_test.go
```diff
@@ -178,6 +178,149 @@ func TestStorePut(t *testing.T) {
...
+			t.Errorf("failed to revoke lease ID: %d, err: %v", l.ID, err)
+		}
+
+		if s.le.Lookup(l.ID) != nil {
+			fmt.Println("not nil")
+		} else {
+			fmt.Println("nil")
+		}
+
+		assert.Equal(t, (*lease.Lease)(nil), s.le.Lookup(l.ID))
```
serathius, line 316:
```suggestion
		assert.Nil(t, s.le.Lookup(l.ID))
```

server/storage/mvcc/kvstore_test.go
```diff
@@ -178,6 +178,149 @@ func TestStorePut(t *testing.T) {
...
+		}
+		if err := s.le.Revoke(l.ID); err != nil {
+			t.Errorf("failed to revoke lease ID: %d, err: %v", l.ID, err)
+		}
+
+		if s.le.Lookup(l.ID) != nil {
+			fmt.Println("not nil")
+		} else {
+			fmt.Println("nil")
+		}
```
serathius, line 314: What this is for?

server/lease/lessor.go
```diff
@@ -572,6 +572,9 @@ func (le *lessor) Detach(id LeaseID, items []LeaseItem) error {
...
+		}
 		return ErrLeaseNotFound
```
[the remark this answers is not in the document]
qsyqian: thx for review. I have a litte worry.

serathius:
ping @qsyqian

qsyqian:
@serathius @chaochn47 fix it in another way. add step to delete le.itemMap in le.Revoke() after txn.End().

server/lease/lessor.go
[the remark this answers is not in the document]
ahrtr: Thanks @chaochn47 .  Indeed https://github.com/etcd-io/etcd/pull/14080 introduced the leak, will send a PR to rollback the change. 

Actually the deadlock fixed in https://github.com/etcd-io/etcd/pull/14080 doesn't likely happen because all lease operations (grant, revoke, checkpoint) are executed sequentially in the applying workflow. If anyone still thinks it's possible to run into a deadlock, please create a test to reproduce it firstlhy.

ahrtr:
@qsyqian The` le.itemMap leak` issue has been resolved in https://github.com/etcd-io/etcd/pull/17058.  Please rebase this PR, and remove the change in lessor.go. The test cases should still pass. Thanks.

qsyqian:
@ahrtr updated, remove changes in lessor.go and also change the commit message.

server/lease/lessor_test.go
```diff
@@ -208,6 +208,47 @@ func TestLessorRevoke(t *testing.T) {
...
+		t.Fatalf("failed to detach items to the lease: %v", err)
+	}
+	if err := le.Revoke(l.ID); err != nil {
+		t.Fatalf("failed to revoke lease: %v", err)
+	}
+
+	if len(le.leaseMap) != 0 {
+		t.Errorf("le.leaseMap's length is %d, expected: 0", len(le.leaseMap))
+	}
+	if len(le.itemMap) != 0 {
```
serathius, line 247: Duplicate with lines 244-246?

PR #16041, opened 2023-06-09
Adding test to verify membership after old version recovery

geetasg:
Related to task - "V2 snapshots are generated from V3 data" - https://github.com/etcd-io/etcd/issues/12913


tests/e2e/v2store_deprecation_test.go
```diff
@@ -182,6 +184,74 @@ func TestV2DeprecationSnapshotRecover(t *testing.T) {
...
+	var snapshotCount uint64 = 10
+	epc := runEtcdAndCreateSnapshot(t, e2e.CurrentVersion, dataDir, snapshotCount)
+
+	lastVersion, err := e2e.GetVersionFromBinary(e2e.BinPath.EtcdLastRelease)
+	lastVersionStr := lastVersion.String()
+	lastClusterVersion := semver.New(lastVersionStr)
+	lastClusterVersion.Patch = 0
+	lastClusterVersionStr := lastClusterVersion.String()
+	t.Logf("etcdctl downgrade enable %s", lastVersionStr)
+	downgradeEnable(t, epc, lastVersion)
```
serathius, line 205: Can you refactor this code to separate function?

tests/e2e/v2store_deprecation_test.go
```diff
@@ -182,6 +184,74 @@ func TestV2DeprecationSnapshotRecover(t *testing.T) {
 	assert.NoError(t, epc.Close())
 }
 
+func TestV2DeprecationSnapshotRecoverOldVersion(t *testing.T) {
```
serathius, line 187: This looks like a standard downgrade test. Have you seen the test in https://github.com/etcd-io/etcd/blob/a708bed336bb95446f1a00e3cf1e0e8858af0cb4/tests/e2e/cluster_downgrade_test.go#L22 ? Why do we need a separate test for V2 backward compatibility?

tests/e2e/v2store_deprecation_test.go
geetasg: updating.

tests/e2e/v2store_deprecation_test.go
geetasg: https://github.com/etcd-io/etcd/blob/a708bed336bb95446f1a00e3cf1e0e8858af0cb4/tests/e2e/cluster_downgrade_test.go#L22 test does not seem to create a snapshot and kvs. 
In next changes when we change how the .snap is published by 3.6, we need a test that recovers from .snap and verifies membership match.

geetasg: please let me know if I should add the test to cluster_downgrade_test.go instead of this file.

tests/e2e/cluster_downgrade_test.go
```diff
@@ -243,3 +253,34 @@ func getMemberVersionByCurl(cfg *e2e.EtcdProcessClusterConfig, member e2e.EtcdPr
 	}
 	return result, nil
 }
+
+func generateSnapshot(t *testing.T, snapshotCount uint64, epc *e2e.EtcdProcessCluster) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	t.Log("Adding and removing keys")
```
ahrtr, line 261: nit: I only see adding keys, and do not se removing keys:)

tests/e2e/cluster_downgrade_test.go
```diff
@@ -67,7 +69,7 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 	e2e.BeforeTest(t)
 
 	t.Logf("Create cluster with version %s", currentVersionStr)
-	epc := newCluster(t, clusterSize)
+	epc := newCluster(t, clusterSize, 10, t.TempDir())
```
ahrtr, line 72: The dataDirPath is t.TempDir by default, FYI. https://github.com/etcd-io/etcd/blob/61736c329d19cadfd0bef3c3ee6af63ee010d9ba/tests/framework/e2e/cluster.go#L504
```suggestion
	epc := newCluster(t, clusterSize, 10)
```

ahrtr: And also the hard coded value 10 is used multiple times, can you define a constant?

tests/e2e/cluster_downgrade_test.go
```diff
@@ -76,6 +78,8 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int) {
 		})
 	}
 	t.Logf("Cluster created")
+	generateSnapshot(t, 10, epc)
```
serathius, line 81: Note, it might be useful to have both downgrade tests with and without snapshot. 
`TestDowngradeUpgradeClusterOf1` and `TestDowngradeUpgradeClusterOf1WithSnapshot`

serathius, line 81: Please use names const variables instead of magical numbers `10`

tests/e2e/cluster_downgrade_test.go
geetasg: Thanks, I will update the comment.

tests/e2e/cluster_downgrade_test.go
geetasg: updating.

tests/e2e/cluster_downgrade_test.go
geetasg: sg. updating.

geetasg: updating.

serathius (approved):

ahrtr (approved):
LGTM

Thanks @geetasg

tests/e2e/cluster_downgrade_test.go
```diff
@@ -243,3 +264,34 @@ func getMemberVersionByCurl(cfg *e2e.EtcdProcessClusterConfig, member e2e.EtcdPr
...
+
+	cc, err := e2e.NewEtcdctl(epc.Cfg.Client, epc.EndpointsGRPC())
+	assert.NoError(t, err)
+
+	var i uint64
+	t.Logf("Adding keys")
+	for i = 0; i < snapshotCount*3; i++ {
+		err := cc.Put(ctx, fmt.Sprintf("%d", i), "1", config.PutOptions{})
+		assert.NoError(t, err)
+	}
```
serathius, line 280: Followup @geetasg. Could this function confirm that there is a snapshot file on disk present?

geetasg: updating.

serathius: Awesome!

chaochn47:
Thanks @geetasg for sharing the doc [v2deprecation-3.6](https://docs.google.com/document/d/19_7xBwZBNrM-75h5rr2PUWFM_eNjSXRVPZZpVaTS2qw/edit) !!

For context to anyone else that has interest in this PR, the added test case is to ensure the following attempt won't regress.. 

> With the v2 removal work for 3.6, the snapshot will continue to look the same but 3.6 will NOT keep a running copy of v2store. It will keep membership info only in the backend and at the time of the snapshot it will export a v2store json to put into the .snap file.

> For etcd 3.6, we still want to publish membership information in v2 format in the .snap files so that etcd 3.5 can recover membership from those in case of downgrade.

PR #16042, opened 2023-06-09
tests/e2e: Use epc.Etcdctl

serathius:
cc @ahrtr @ptabor

### DIFF
--- a/tests/e2e/etcd_config_test.go
+++ b/tests/e2e/etcd_config_test.go
@@ -411,12 +411,9 @@ func TestEtcdHealthyWithTinySnapshotCatchupEntries(t *testing.T) {
 	for i := 0; i < 10; i++ {
 		clientId := i
 		g.Go(func() error {
-			cc, err := e2e.NewEtcdctl(epc.Cfg.Client, epc.EndpointsGRPC())
-			if err != nil {
-				return err
-			}
+			cc := epc.Etcdctl()
 			for j := 0; j < 100; j++ {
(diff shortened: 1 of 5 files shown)


tests/e2e/etcd_config_test.go
```diff
@@ -411,10 +411,7 @@ func TestEtcdHealthyWithTinySnapshotCatchupEntries(t *testing.T) {
 	for i := 0; i < 10; i++ {
 		clientId := i
 		g.Go(func() error {
-			cc, err := e2e.NewEtcdctl(epc.Cfg.Client, epc.EndpointsGRPC())
```
jmhbnz, line 414: Grpc e2e failing with data race after this removal. I'm not sure we can remove this instance of the old approach as the test is intended to simulate 10 different clients?

serathius: `Etcdctl` is not a client, it a wrapper over a etcdctl binary. Not only each `Etcdctl` call is unique, but also method call is unique execution of binary

jmhbnz: Ahh right I see the spawns in `etcdctl.go`, apologies.  Any ideas why this change would be creating the data race with that in mind?

serathius: Simple, I removed  scoped `err` variable.

jmhbnz: Nice and checks now green. Thanks for indulging the questions.

jmhbnz (approved):
LGTM - Nice to cut down on all the duplication.

ahrtr (approved):
LGTM

Thanks

PR #15886, opened 2023-05-12
Run robustness tests on arm64

jmhbnz:
In order to progress tier one support for `arm64` we need to get nightly on per pull request actions workflows running our robustness test suite on the `arm64` platform.

This pull request completes the work started in https://github.com/etcd-io/etcd/pull/15310:
- Robustness template updated to support dynamic `runs-on`.
- New `main-arm64` jobs added to robustness nightly workflow.

Additionally this pull request adds the [recommended `self-hosted` label](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-self-hosted-runners) to existing `arm64` tests and e2e workflows.

Relates to https://github.com/etcd-io/etcd/issues/14748

cc @serathius, @ahrtr, @chaochn47

PR #16043, opened 2023-06-09
.github/workflows: Read .go-version as a step and not separate workflow

serathius:
cc @jmhbnz @ahrtr @ptabor


jmhbnz (approved):
LGTM - As nice as the idea of the re-usable workflow file is, for something like setting go version the ux of github is too problematic with spam output as we discuss here https://github.com/etcd-io/etcd/pull/15886#issuecomment-1546549231

This pr is best of both worlds, only one file to maintain for version, and no spam from github from the re-usable workflow.

@serathius if others agree and we merge this I'll sort out the backport either myself or with an issue for a newer contributor.

PR #16044, opened 2023-06-09
tests/robustness: Assume starting from empty etcd instead of throwing out first failed request

serathius:
cc @ahrtr @jmhbnz

### DIFF
--- a/tests/robustness/model/non_deterministic.go
+++ b/tests/robustness/model/non_deterministic.go
@@ -71,11 +71,10 @@ type EtcdNonDeterministicResponse struct {
 
 func (states nonDeterministicState) Step(request EtcdRequest, response EtcdNonDeterministicResponse) (bool, nonDeterministicState) {
 	if len(states) == 0 {
-		// states were not initialized
-		if response.Err != nil || response.ResultUnknown || response.Revision == 0 {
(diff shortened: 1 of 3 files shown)


tests/robustness/model/non_deterministic.go
```diff
@@ -71,11 +71,10 @@ type EtcdNonDeterministicResponse struct {
 
 func (states nonDeterministicState) Step(request EtcdRequest, response EtcdNonDeterministicResponse) (bool, nonDeterministicState) {
 	if len(states) == 0 {
-		// states were not initialized
-		if response.Err != nil || response.ResultUnknown || response.Revision == 0 {
```
ahrtr, line 75: Print a log if `response.Err != nil || response.ResultUnknown`?

ahrtr (approved):

ahrtr (approved):

jmhbnz (approved):
LGTM

tests/robustness/model/non_deterministic.go
serathius: This is model code which is used for linearization. This means that this is an expected path that could be potentially called hundreds of times. Don't think we should have any logging in model.

PR #16068, opened 2023-06-12
[3.5] etcdserver: backport check scheduledCompactKeyName and finishedCompac…

### LINKED ISSUE #15919: Incorrect hash when resuming scheduled compaction after etcd restarts

CaojiamingAlan:
…tKeyName before writing hash to release-3.5.

Follow up of #15985

Fix #15919.
Check ScheduledCompactKeyName and FinishedCompactKeyName before writing hash to hashstore.
If they do not match, then it means this compaction has once been interrupted and its hash value is invalid. In such cases, we won't write the hash values to the hashstore, and avoids the incorrect corruption alarm.

### DIFF
--- a/server/mvcc/kvstore.go
+++ b/server/mvcc/kvstore.go
@@ -227,7 +227,27 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
 	return nil, compactMainRev, nil
 }
 
-func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-chan struct{}, error) {
+// checkPrevCompactionCompleted checks whether the previous scheduled compaction is completed.
+func (s *store) checkPrevCompactionCompleted() bool {
+	tx := s.b.ReadTx()
+	tx.Lock()
+	defer tx.Unlock()
+	_, scheduledCompactBytes := tx.UnsafeRange(buckets.Meta, scheduledCompactKeyName, nil, 0)
+	scheduledCompact := int64(0)
+	if len(scheduledCompactBytes) != 0 {
+		scheduledCompact = bytesToRev(scheduledCompactBytes[0]).main
+	}
+
+	_, finishedCompactBytes := tx.UnsafeRange(buckets.Meta, finishedCompactKeyName, nil, 0)
+	finishedCompact := int64(0)
+	if len(finishedCompactBytes) != 0 {
+		finishedCompact = bytesToRev(finishedCompactBytes[0]).main
+
+	}
+	return scheduledCompact == finishedCompact
+}
+
+func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64, prevCompactionCompleted bool) (<-chan struct{}, error) {
 	ch := make(chan struct{})
 	var j = func(ctx context.Context) {
 		if ctx.Err() != nil {
@@ -240,7 +260,13 @@ func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-ch
 			s.compactBarrier(context.TODO(), ch)
 			return
 		}
-		s.hashes.Store(hash)
+		// Only store the hash value if the previous hash is completed, i.e. this compaction
+		// hashes every revision from last compaction. For more details, see #15919.
+		if prevCompactionCompleted {
+			s.hashes.Store(hash)
+		} else {
+			s.lg.Info("previous compaction was interrupted, skip storing compaction hash value")
+		}
 		close(ch)
 	}
 
@@ -250,17 +276,19 @@ func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-ch
 }
 
 func (s *store) compactLockfree(rev int64) (<-chan struct{}, error) {
+	prevCompactionCompleted := s.checkPrevCompactionCompleted()
 	ch, prevCompactRev, err := s.updateCompactRev(rev)
 	if err != nil {
 		return ch, err
 	}
 
-	return s.compact(traceutil.TODO(), rev, prevCompactRev)
+	return s.compact(traceutil.TODO(), rev, prevCompactRev, prevCompactionCompleted)
 }
 
 func (s *store) Compact(trace *traceutil.Trace, rev int64) (<-chan struct{}, error) {
 	s.mu.Lock()
 
+	prevCompactionCompleted := s.checkPrevCompactionCompleted()
 	ch, prevCompactRev, err := s.updateCompactRev(rev)
 	trace.Step("check and update compact revision")
 	if err != nil {
@@ -269,7 +297,7 @@ func (s *store) Compact(trace *traceutil.Trace, rev int64) (<-chan struct{}, err
 	}
 	s.mu.Unlock()
 
-	return s.compact(trace, rev, prevCompactRev)
+	return s.compact(trace, rev, prevCompactRev, prevCompactionCompleted)
 }
 
 func (s *store) Commit() {
--- a/tests/e2e/corrupt_test.go
+++ b/tests/e2e/corrupt_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/datadir"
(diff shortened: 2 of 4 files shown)


serathius (approved):

server/mvcc/kvstore.go
```diff
@@ -227,7 +227,27 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
 	return nil, compactMainRev, nil
 }
 
-func (s *store) compact(trace *traceutil.Trace, rev, prevCompactRev int64) (<-chan struct{}, error) {
+// checkPrevCompactionCompleted checks whether the previous scheduled compaction is completed.
+func (s *store) checkPrevCompactionCompleted() bool {
+	tx := s.b.ReadTx()
+	tx.Lock()
```
chaochn47, line 233: Just curious why not use `tx.RLock()` but understand the PR is a backport of https://github.com/etcd-io/etcd/pull/15985

chaochn47 (approved):

server/mvcc/kvstore.go
serathius: `Lock()` for `ReadTx()` should be by default a `Rlock()`.

chaochn47: For my understanding, do you mean the following RWMutex `Lock` is equal to `RLock` to protect the read buffer? 

https://github.com/etcd-io/etcd/blob/5e40a8b00cc1be3ce0dd3d7e6947d417253fb03f/server/storage/backend/read_tx.go#L127-L130

CaojiamingAlan: After investigating a little bit, I think you are right, we should use RLock() instead. Lock() should only be used when the read buffer is updated by the writes committed by the BatchTxs. However, there are several positions where use read-only operations but Lock() is used. I think it is good to open a separate PR to correct them all.

chaochn47: Yep. That aligns with my understanding. 

I also don't think it blocks this PR to be merged. This can be corrected separately.

tests/e2e/corrupt_test.go
```diff
@@ -180,3 +181,61 @@ func TestCompactHashCheckDetectCorruption(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	})
+
+	// Put 200 identical keys to the cluster, so that the compaction will drop some stale values.
+	// We need a relatively big number here to make the compaction takes a non-trivial time, and we can interrupt it.
```
jmhbnz, line 210: Double checking my understanding, this backport uses a much larger value here compared to original pr because in 3.5 we don't have the luxury of setting very slow compaction speed is that right?

server/mvcc/kvstore.go
jmhbnz: Nice catch @chaochn47 - suggest we raise an issue for this so we don't forget to come back to it.

tests/e2e/corrupt_test.go
CaojiamingAlan: Yes, it does not have a batch-sleeping-time flag.

jmhbnz (approved):
LGTM

server/mvcc/kvstore.go
ahrtr: The `Lock()/Unlock()` should only be used in the following three cases,
- When [committing a transaction](https://github.com/etcd-io/etcd/blob/dee90e19f1c6fc2aceed38cd2855010df0b6c6a7/server/storage/backend/batch_tx.go#L336). Because it will reset the read tx, including the buffer;
- When [writing the pending data (not committed to bbolt yet) back to read buf](https://github.com/etcd-io/etcd/blob/dee90e19f1c6fc2aceed38cd2855010df0b6c6a7/server/storage/backend/batch_tx.go#L310), because obviously no reading is allowed when writing the read buffer;
- When [performing defragmentation](https://github.com/etcd-io/etcd/blob/dee90e19f1c6fc2aceed38cd2855010df0b6c6a7/server/storage/backend/backend.go#L464),  because again it will reset the read tx, including the read buffer.

`RLock()/RUnlock()` should be used for all other cases.

Please raise a separate ticket to track the change. It should also be backported to 3.5 and 3.4.

server/mvcc/kvstore.go
```diff
@@ -227,7 +227,27 @@ func (s *store) updateCompactRev(rev int64) (<-chan struct{}, int64, error) {
...
+	if len(scheduledCompactBytes) != 0 {
+		scheduledCompact = bytesToRev(scheduledCompactBytes[0]).main
+	}
+
+	_, finishedCompactBytes := tx.UnsafeRange(buckets.Meta, finishedCompactKeyName, nil, 0)
+	finishedCompact := int64(0)
+	if len(finishedCompactBytes) != 0 {
+		finishedCompact = bytesToRev(finishedCompactBytes[0]).main
+
+	}
```
ahrtr, line 246: Please add a `UnsafeReadScheduledCompact` and `UnsafeReadFinishedCompact` in server/mvcc/util.go for 3.5.

The two functions should be reused in multiple places, (Please search `scheduledCompactKeyName` and `finishedCompactKeyName` in the code base)

It's accepted to fix it in a separate PR. Please also raise a followup ticket.

ahrtr:
Please raise two followup ticket as mentioned in https://github.com/etcd-io/etcd/pull/16068#discussion_r1263664700 and https://github.com/etcd-io/etcd/pull/16068#discussion_r1263667496. 

Pls also rebase this PR.

CaojiamingAlan:
Just raised #16247. I will solve https://github.com/etcd-io/etcd/pull/16068#discussion_r1263667496 real quick, so no need for another PR.

ahrtr (approved):
LGTM

thx @CaojiamingAlan

server/mvcc/kvstore.go
serathius: @ahrtr thanks for figuring out proper usage of transaction lock. However, I don't_ see reason why we should backport it. Old implementation uses stricter lock, which is sub-optimal, but is not a bug. 

To ensure release stability, we need to assume that the list you proposed can have a mistake. I would prefer that we don't verify correctness of performance improvements on stable releases.

ahrtr: Note I prefer to backporting it, but I am not insist on backporting it if there is strong objection from other maintainers.

serathius: > stricter lock is true, it's also wrong lock. It's a bug to me.

Without a change system behaves correctly, after the change only thing you expect is performance improvement. How this is a bug?

ahrtr: I would say it's harmless (on functionality) bug, obviously the lock isn't correctly be used. Let alone it has impact on performance.

serathius: Bugs is when locks don't protect the critical section, relaxing locks is performance improvement. Nothing is harmless until proven, let alone concurrency change.

Issue #16069, opened 2023-06-13
Return snapshot size when snapshot.


qsyqian:
### What would you like to be added?

Return snapshot size when do snapshot for one etcd cluster.

### Why is this needed?

when snapshot one etcd cluster it wll return one io.ReadCloser. the client get the snapshot size only when finish copy all bytes from io.ReadCloser. If we can return the snapshot size, it will be convenient for client to do something.
For example, some object storage on the cloud has size limit when upload object from one io.reader.

serathius:
Don't understand the motivation. Not sure what client can do with size if they already downloading the snapshot.  If client wants a size of data to avoid downloading to much data they should check before requesting snapshot by checking etcd metric for db size.

jmhbnz:
Agreed - Reviewing the linked pull request I think a better way to tackle something like this is upstream in the client itself before any snapshot download commences.

I can understand the motivation from a client perspective, trying to prevent a user from shooting themselves in the foot by attempting to download and save a snapshot they don't have space on disk for.  Maybe something we could build into https://github.com/etcd-io/etcd/blob/main/etcdctl/ctlv3/command/snapshot_command.go at some stage but low priority right now I think.

tjungblu:
I think this has some merit to include in the response. It happened in OpenShift where a customer had too many backup snapshots on the same partition as etcd, which at some point filled it up and caused other trouble. 
It was our fault for the lack of a retention mechanism, but we could've degraded our operator if we detect the snapshot would be bigger than the available space on disk.

We have the total size readily available:
https://github.com/etcd-io/etcd/blob/fd3e338d888216b7924f45136ace132ab45fad4e/server/etcdserver/api/v3rpc/maintenance.go#L125-L134

The storage version is also constant throughout the whole streaming, so adding another integer to denote the total size probably isn't going to blow it up too much.

zaitcev:
Discussion ended in June. Should this Issue be rejected and closed now, or is there a possibility of a merit?

PR #16071, opened 2023-06-13
clientv3: add snapshot size in SnapshotResponse

qsyqian:
For #16069

 @ahrtr @serathius PTAL, feel free to close it if this PR is unreasonable.


jmhbnz (commented):
While it's discussed in the linked issue, just to reiterate here my preference would be not to handle the problem this way, though many thanks to @qsyqian for putting the proposal forward.

My view is that any snapshot size vs disk space considerations should be handled earlier in the sequence of events, before a snapshot starts streaming.

I suggest this pr is closed but happy to be out voted if others disagree.

PR #16018, opened 2023-06-06
Add failpoint for nospace on puts

tjungblu:
This CR introduces a new failput that will trigger a member to report no space.

---

This might need some more legwork, it's the first time I'm adding a failpoint here.

Issue #15079, opened 2023-01-10
Robustness tests etcd logs should go to artifacts and not spam output 


serathius:
### What would you like to be added?

In Linearizability action we use EXPECT_DEBUG=true to get logs to stdout. Because of continious failpoint injection their volume is overwhelming. Better approach would be to redirect logs to file and upload them to artifacts on failed test.

### Why is this needed?

Should improve clarity and performance of "Linearizability" test action.

anushkamittal20:
Hey @serathius, I would like to start contributing to etcd with this issue. 
I see that we have set EXPECT_DEBUG=true in linearizability-template.yaml, to redirect them to a separate file we could use the braces '{ ... } 1> logs.txt' to provide a unit of I/O redirection. But we want to eliminate an stdout completely and redirect to a separate file. For the same, I was trying to look to read up on the EXPECT_DEBUG ( parameter but couldn't find much on it. Could you please direct me to some documentation? ) I was thinking that in this part of the [code](https://github.com/etcd-io/etcd/blob/ee566c492bb2e0962068a531666c68e1c39d3723/pkg/expect/expect.go#L131) we could change the code to send to a file and tar it instead of printing it directly using fmt.
It would be great to hear your feedback on this!
Thank you very much!

serathius:
This issue is about changing https://github.com/etcd-io/etcd/blob/ee566c492bb2e0962068a531666c68e1c39d3723/pkg/expect/expect.go package to support writing to files.

pchan:
Is anyone looking at this ? If not, I can attempt to add the capability.

jmhbnz:
Hi @pchan there are no work in progress pull requests linked so feel free to work on it, I have assigned the issue to you, thanks! :)

chaochn47:
Hi @pchan, is there any progress on this? 

Feel free to flush the local changes as a drafted PR even if it is not yet fully ready for review. 

> Handle inactive PRs
> Poke PR owner if review comments are not addressed in 15 days. If PR owner does not reply in 90 days, update the PR with a new commit if possible. If not, inactive PR should be closed after 180 days. 

https://etcd.io/docs/v3.5/triage/prs/#handle-inactive-prs

serathius:
Note, with migration to Prow this might not be needed. Prow is much much better in handling logs.

jmhbnz:
> Note, with migration to Prow this might not be needed. Prow is much much better in handling logs. #18136

[...]

PR #16077, opened 2023-06-14
Retain process logs on robustness tests

tjungblu:
Adding a new flag to retain e2e etcd process logs after stop and saving next to the visualized model.

Spun out of #16018 where I used it for easier local debugging on model violations.


serathius:
Etcd logs are available the same way as e2e logs. By passing `EXPECT_DEBUG=true` flag to the test run.

tjungblu:
> Etcd logs are available the same way as e2e logs. By passing EXPECT_DEBUG=true flag to the test run.

If you feel like browsing the logs is convenient enough already I'm happy to close. For me it makes more sense to have the member debug logs next to the history.html in the artifacts zip.

tests/framework/e2e/etcd_process.go
```diff
@@ -57,6 +57,7 @@ type EtcdProcess interface {
 	PeerProxy() proxy.Server
 	Failpoints() *BinaryFailpoints
 	Logs() LogsExpect
+	LogLines() []string
```
serathius, line 60: Please add Lines() to `LogsExpect` interface

serathius:
> If you feel like browsing the logs is convenient enough already I'm happy to close. For me it makes more sense to have the member debug logs next to the history.html in the artifacts zip.

It's not https://github.com/etcd-io/etcd/issues/15079

tests/framework/e2e/etcd_process.go
```diff
@@ -72,6 +73,8 @@ type EtcdServerProcess struct {
 	proxy      proxy.Server
 	failpoints *BinaryFailpoints
 	donec      chan struct{} // closed when Interact() terminates
+
+	keptLogLines []string // only filled when KeepLogs=true
```
serathius, line 77: Not sure why this is needed. ExpectLogs already stores all logs in memory. Don't see a problem to always keep logs.

tjungblu: it's because of the ep.proc lifecycle below:

```
		if ep.cfg.KeepLogs {
			ep.keptLogLines = ep.proc.Lines()
		}
***		ep.proc = nil ***
```

it's also used to flag whether a process was stopped, which coincidentally wipes the logs with it.

serathius: Then I would suggest storing `ExpectLog` interface instead of `[]string{}`

tjungblu: sounds good

serathius:
FYI main challenge here is to merge logs from between restarts.

tjungblu:
@serathius created an aggregation of LogExpects to merge the logs. Let me know if the implemented behavior works for you, too (mostly on the expect side of things).

tests/framework/e2e/etcd_process.go
```diff
@@ -230,7 +233,7 @@ func (ep *EtcdServerProcess) Config() *EtcdServerProcessConfig { return ep.cfg }
 
 func (ep *EtcdServerProcess) Logs() LogsExpect {
```
serathius, line 234: Code looks nice, but changing `func (ep *EtcdServerProcess) Logs() LogsExpect` semantic behavior is risky. When I expect most people will assume `member.Logs().ExpectWithContext()` will analyse logs emitted from the current run and might stumble on problem that log they are expecting was already emitted before restart. 

Proposal:
* Have `func (ep *EtcdServerProcess) Logs() LogsExpect` return logsExpect from the latest run `ep.logs.procLogs[len(ep.logs.procLogs)-1]`
* Have a separate method like `allLogLines` or `aggregatedLogLines` etcd that will return appended results of all runs.

What do you think?

chaochn47: Agreed with the incremental changes to be safe. 

`func (ep *EtcdServerProcess) Logs() LogsExpect` was using `ep.proc` in happy case and most of the cases. 

The PR is intended to retain all the previous runs logs after all the runs of the process are stopped. 

Seems only `func (m *multiLogsExpect) Lines()` is needed.

tjungblu: > Have func (ep *EtcdServerProcess) Logs() LogsExpect return logsExpect from the latest run ep.logs.procLogs[len(ep.logs.procLogs)-1]

Sounds like a good proposal to me. Is that also acceptable for you @chaochn47?

> Seems only func (m *multiLogsExpect) Lines() is needed.

Currently yes, I'm surprised we don't have expect tests across stop/restart scenarios.

serathius: > Currently yes, I'm surprised we don't have expect tests across stop/restart scenarios.

We want to explicitly avoid that because it not natural. Imagine that you want to check that etcd member has started using one of the starting logs. It's natural to call `member.Logs().Expect("etcd started")`, but it needs to only consider the logs from the last restart to make sense. Otherwise it might always return true when member was already restarted.

tldr. Don't want to change the semantics of `func (ep *EtcdServerProcess) Logs() LogsExpect`.

tjungblu: Cool, pushed another by just moving up the Lines and Count via interface, so the current LogsExpect stays entirely untouched.

tests/framework/e2e/etcd_process.go
```diff
@@ -235,6 +242,14 @@ func (ep *EtcdServerProcess) Logs() LogsExpect {
 	return ep.proc
 }
 
+func (ep *EtcdServerProcess) Lines() []string {
+	return ep.logs.Lines()
+}
+
+func (ep *EtcdServerProcess) LineCount() int {
```
serathius, line 249: Doesn't looks like it's used by anyone. Remove?

tests/framework/e2e/etcd_process.go
```diff
@@ -72,6 +77,7 @@ type EtcdServerProcess struct {
 	proxy      proxy.Server
 	failpoints *BinaryFailpoints
 	donec      chan struct{} // closed when Interact() terminates
+	logs       *procLogs
```
serathius, line 80:
```suggestion
	logs       procLogs
```
It doesn't need to be a pointer you know?

tests/framework/e2e/etcd_process.go
serathius: I mean it's not needed for EtcdServerProcess.

serathius:
Also, please remove environment EXPECT_DEBUG from the Robustness tests github workflow.

PR #16079, opened 2023-06-14
Robustness range request

serathius:
cc @ahrtr @ptabor @jmhbnz @chaochn47 @lavacat

### DIFF
--- a/tests/robustness/model/deterministic.go
+++ b/tests/robustness/model/deterministic.go
@@ -83,19 +83,19 @@ func initState(request EtcdRequest, response EtcdResponse) etcdState {
 		for i, op := range request.Txn.OperationsOnSuccess {
 			opResp := response.Txn.Results[i]
 			switch op.Type {
-			case Range:
+			case RangeOperation:
 				for _, kv := range opResp.KVs {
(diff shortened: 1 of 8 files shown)


tests/robustness/model/deterministic.go
```diff
@@ -281,14 +317,20 @@ type EtcdCondition struct {
...
+	PutOptions
 }
 
+type OperationType string
+
+const (
+	RangeOperation  OperationType = "range-operation"
+	PutOperation    OperationType = "put-operation"
+	DeleteOperation OperationType = "delete-operation"
+)
```
chaochn47, line 332: Can I think of the operation here is a similar alias of the following `RequestOp` used in the TxnRequest with one exception that `EtcdOperation` in robustness test was also used for robustness test watch event to patch the history and remove failed writes? 

https://github.com/etcd-io/etcd/blob/b4f49a55a5aba85a30cae5742a87989132a0f750/api/etcdserverpb/rpc.proto#L571-L580

serathius: Yes, those are sub-requests within a transaction request, in model I just called them operations. We are missing TxnRequest as model doesn't support subtransactions currently (think it's overkill).

PR #13348, opened 2021-09-14
Fix for v3.5 Ensure that cluster members stored in v2store and backend are in sync

serathius:
As discussed in https://github.com/etcd-io/etcd/issues/13196 there is a chance that between v3.1 and v3.4 state of bbolt and v2store diverged. This is only noticeable after we upgrade to v3.5 as the authoritative storage was changed from v2store to bbolt.

This PR is a direct fix this problem in v3.5. This is not cherry-pick from master as v3.6 plans to remove v2store totally. Current fix is designed to work assuming that users uses v3.5.1 before upgrading to v3.6. This approach was picked to avoid delaying removal of v2storage in v3.6 (discussed more on original issue).

What this PR does:
* AddMember/RemoveMember operation will execute on both storeV2 and backend before returning. If they have diverged, and storev2 has member that is not present in backend, re-adding this member will succeed (mirrored for removal). Adding an exiting member will only fail if it's present in both storages(mirrored for removal) . As writing to 2 different storages is not transactional, they might diverge during runtime. This should allow users to repeat the Add/Remove Member to ensure that storages are in sync. This should be enough to fix the problem as member changes are not done often.
* ~~When etcd bootstraps, RaftCluster Recover will sync contents of backend to match v2store. This operation is always correct and it should be cheap as number of cluster members should be around 3-5, and historically it should not grow beyond tens. This should fix problem of zombie members appearing after upgrade to v3.5.~~ Resigned from implementing this as during Recover it's expected that v2store and backend are out of sync. Reason is that Recover is called during member bootstrap before applying WAL, as storev2 and backend are persisted at different time (storev2 only after snapshot, backend every 5 seconds) .

cc @ptabor

Issue #13196, opened 2021-07-08
etcd 3.5.0 resurrects ancient (unremovable) members


anguslees:
Zombie members!

I [upgraded](https://github.com/anguslees/k8s-home/commit/b883c6d761f0a8a882a0b5ff1a5b46b46c5a0e36) my home k8s cluster to etcd 3.5.0 (from 3.4.14).  I noticed in the logs that it started complaining about the health of surprising IPs on my network, but it didn't seem to affect other functionality.  Error messages like:
```
{"level":"warn","ts":"2021-06-16T08:40:21.800Z","caller":"rafthttp/probing_status.go:68","msg":"prober detected unhealthy status","round-tripper-name":"ROUND_TRIPPER_SNAPSHOT","remote-peer-id":"8e9e05c52164694d","rtt":"0s","error":"dial tcp 192.168.0.9:2380: connect: no route to host"}
```
Where `192.168.0.9:2380` is _not_ one of my current etcd cluster members.  It's an ancient one from a _long_ time ago.  I did some git archaeology, and I think this member last existed back when I was using etcd 3.3.15 (or even earlier).

Now I note `etcdctl member list` includes all these ancient members, which it did not before the upgrade:
```console
roachbait:~% kubectl -n kube-system exec -ti etcd-2 -- etcdctl member list
407604e04d511f63, started, kmaster2, https://192.168.0.140:2380, https://192.168.0.140:2379, false
80df0211ad8a11c3, started, 0b5642a6cc18493d81a606483d9cbb7b, https://192.168.0.132:2380, https://192.168.0.132:2379, false
8e9e05c52164694d, started, default, https://192.168.0.9:2380, http://127.0.0.1:2379,https://192.168.0.9:2379, false
aaaacf0529a32f2f, started, default, https://192.168.0.9:2380, https://192.168.0.9:2379, false
b5ed3c441abd0677, unstarted, , https://192.168.0.40:2380, , false
c7a95206b6a22d21, started, fc4698cdc1184810a2c3447a7ee66689, https://192.168.0.129:2380, https://192.168.0.129:2379, false
c93ba62f815b42ed, started, 765885d83e774555bc7ee0f9c6fc1178, https://192.168.0.117:2380, https://192.168.0.117:2379, true
cc28792212a199be, started, 887f1b514ea54520a61643163d427d42, https://192.168.0.161:2380, https://192.168.0.161:2379, false
```
[...]

PR #16084, opened 2023-06-14
Validate conf change using v3

geetasg:
Eventually the tests will stop setting v2store.

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -303,9 +303,18 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 
 // ValidateConfigurationChange takes a proposed ConfChange and
 // ensures that it is still valid.
-func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
-	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
+	if c.be != nil && shouldApplyV3 {
+		membersMap, removedMap := c.be.MustReadMembersFromBackend()
+		if err := c.validateConfigurationChange(cc, membersMap, removedMap); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *RaftCluster) validateConfigurationChange(cc raftpb.ConfChange, membersMap map[types.ID]*Member, removedMap map[types.ID]bool) error {
+
 	id := types.ID(cc.NodeID)
 	if removedMap[id] {
 		return ErrIDRemoved
--- a/server/etcdserver/api/membership/cluster_test.go
+++ b/server/etcdserver/api/membership/cluster_test.go
@@ -278,6 +278,8 @@ func TestClusterValidateAndAssignIDs(t *testing.T) {
 
 func TestClusterValidateConfigurationChange(t *testing.T) {
 	cl := NewCluster(zaptest.NewLogger(t), WithMaxLearners(1))
+	be := newMembershipBackend()
+	cl.SetBackend(be)
 	cl.SetStore(v2store.New())
 	for i := 1; i <= 4; i++ {
 		var isLearner bool
@@ -455,7 +457,7 @@ func TestClusterValidateConfigurationChange(t *testing.T) {
 		},
 	}
 	for i, tt := range tests {
-		err := cl.ValidateConfigurationChange(tt.cc)
+		err := cl.ValidateConfigurationChange(tt.cc, true)
 		if err != tt.werr {
 			t.Errorf("#%d: validateConfigurationChange error = %v, want %v", i, err, tt.werr)
 		}
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1986,7 +1986,7 @@ func removeNeedlessRangeReqs(txn *pb.TxnRequest) {
 // applyConfChange applies a ConfChange to the server. It is only
 // invoked with a ConfChange that has already passed through Raft
 func (s *EtcdServer) applyConfChange(cc raftpb.ConfChange, confState *raftpb.ConfState, shouldApplyV3 membership.ShouldApplyV3) (bool, error) {
-	if err := s.cluster.ValidateConfigurationChange(cc); err != nil {
+	if err := s.cluster.ValidateConfigurationChange(cc, shouldApplyV3); err != nil {
 		cc.NodeID = raft.None
 		s.r.ApplyConfChange(cc)
 


geetasg (commented):
Following tests also need to be updated in similar manner - to not rely on v2store. 
I will update this CR if the refactoring shown in this version looks ok.


TestApplyConfigChangeUpdatesConsistIndex, 
TestApplyMultiConfChangeShouldStop, 
TestAddMember,
TestRemoveMember,
TestUpdateMember,

server/etcdserver/api/membership/cluster.go
```diff
@@ -305,7 +305,13 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
...
 func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
 	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+	var membersMap map[types.ID]*Member
+	var removedMap map[types.ID]bool
+	if c.v2store != nil {
+		membersMap, removedMap = membersFromStore(c.lg, c.v2store)
+	} else {
+		membersMap, removedMap = c.be.MustReadMembersFromBackend()
+	}
```
serathius, line 314: This doesn't seem like a test change but as a logic change to validate config change based on v3 store if v2store is not around. FYI since v3.5 v2 and v3 stores should include membership information, but the v3store membership should be authoritative. 

Please be clear on what this Pr is meant to do. To add tests or  to move validation to v3 store.

geetasg: As of now, this PR is meant for testing ConfChange entries with membership info from be. It does not change logic used by production.
In the code snip above, the else path is inactive in production since v2store is set. Ref:https://github.com/etcd-io/etcd/blob/main/server/etcdserver/bootstrap.go#L449

The test will take the else path and exercise ConfChange validation using info from be.

serathius: Please don't mix adding tests and changing logic. Fact that code you are adding will not change the fact that it adds logic.

geetasg: The test cannot be added without the change in cluster.go discussed above. Can you please clarify how might I just add the test in this case?

geetasg: @serathius can you please comment on this? should this test not be added at this time?

serathius: I don't understand the change proposed. We  should be already migrated to read membership from be as v3.5 was meant to do that.

server/etcdserver/api/membership/cluster.go
```diff
@@ -303,9 +303,27 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
...
+	var membersMap map[types.ID]*Member
+	var removedMap map[types.ID]bool
+	if c.v2store != nil {
+		membersMap, removedMap = membersFromStore(c.lg, c.v2store)
+		if err := c.validateConfigurationChange(cc, membersMap, removedMap); err != nil {
+			return err
+		}
+	}
+
+	if c.be != nil && shouldApplyV3 {
```
chaochn47, line 317: `shouldApplyV3` is set to `true` only when the applying entry's index is greater than consistent index for avoiding applying EntryConfChange entriy twice , say removing a etcd member twice on v3 backend which unnecessarily increases the index. ref. https://github.com/etcd-io/etcd/pull/12854

So, this configuration change validation does not need to be gated by `shouldApplyV3`. Did I understand it correctly? I think it's harmless but just not necessary.

geetasg:
why do we need shouldApplyV3 on Validation -
When recovering from wal without a snapshot, the wal replays. The backend can have future info in that case if ci is ahead and cannot be used for validation. This is similar to how AddMember skips backend addition if ci of backend is already ahead.

server/etcdserver/api/membership/cluster.go
serathius: True, should have no side effects. We will just validate v3 unnecessarily. 

On the other hand, I like the symmetry between validation other normal apply logic. It should make merging those paths easier when we will stop restoring from v2 snapshot.

server/etcdserver/server_test.go
```diff
@@ -514,8 +514,14 @@ func TestApplyRequestOnAdminMemberAttributes(t *testing.T) {
 }
 
 func TestApplyConfChangeError(t *testing.T) {
```
serathius, line 516: Don't understand the test changes. Please add a test that will confirm the change you introduced. Now we not only validate config change based on v2 state, but also from v3 state. I expect we should add a test that confirms that v3 is validated.

serathius:
For me this is a bug fix that should be backported to v3.5. Reason is that v3.5 release was meant to switch etcd source of truth for membership from v2 to v3 store. Looks like we missed this, meaning that v3.5 incorrectly validates only v2 store, instead of validating both.

serathius:
cc @ptabor @ahrtr

server/etcdserver/server_test.go
geetasg: TestClusterValidateConfigurationChange and TestApplyConfChangeError now set backend object for the RaftCluster. As a result, these tests will exercise the v3 validation from here - https://github.com/etcd-io/etcd/pull/16084/files#diff-09b9c924aaa34d5ad5abc7e40ed10a2dcc5257c62fda537a8671a7f206d03e6aR317

server/etcdserver/api/membership/cluster.go
geetasg: shouldApplyV3 is needed here because when cluster recovers without a snapshot and/or backend is ahead, the validation will be based on 'future' information and hence not correct. Similar to AddMember and related operations that make the change, the validation should also refer to backend information only if it is not in future.

geetasg:
I will prepare CR to backport to 3.5 after this review completes.

server/etcdserver/api/membership/cluster.go
ptabor: That's scary. @serathius 
I'm afraid we this change we are getting to be hit by linearizability problems. 

Let's consider situation when: 
  1. Cluster is doing some operations on members, where `foo` already: add(foo), remove(foo), 
  2. In healthy situation the add(foo) fails on the all members and remove(foo) succeeds on all the members. 
  3. Now one of this members crashes before snapshot. The member manages to commit remove(foo) to it's backend.  
  4. The member starts recover from snaoshit & reply scenario
  5. It will on the `add(foo)`:
  a) skip the validation (and even if not skipping, the validation would pass as foo is already removed in HEAD)
  b) apply the add(foo) modification to RAFT  -> (while the other members didn't applied it).  
And this leads to a state when one RAFT member is different than the others...

Now the fought experiment: 
  Why this is a problem for confChange and not for Txns that could also fail to compute precondition based on the HEAD ?  Because we skip the applied transaction entirely and trust the backend has proper state ? 

How we can solve this problem: 
 1. We should have @serathius linearability tests on config changes before we will modify the code, as it's very subtle. 
 2. The ultimate goal should be that we don't reply ANY log entries before the 'backend' state. 
 We should think that backend state is the snapshot we restore from, we bootstrap RAFT and membership store... and only apply the future logs.  
 3. In such situation we restore membership information and configure RAFT solely on the state of backend, so we don't need to apply any 'previous' conf changes onto RAFT at all. 

I'm afraid such change needs to happen atomically: 
 - either we restore from any ~V2 snapshot, and incrementally follow conf-changes based on V2 state. 
 - or we only restore from backend virtual snapshot (ignoring snapshots in the RAFT log ?) and configure RAFT based on the backend state. 

Potentially we might try to do some reconciliation:... buffer all the config changes during the reply without setting them in RAFT... and when we reach state of the 'backend', we would reconcile current config in RAFT with desired config in BACKEND. 
But it seems error prone and definitely would require a lot of robustness-testing.

geetasg: @ptabor how about this solution - now the PR has a replay store which should be consulted if backend cannot be consulted. Can you please take a look?

geetasg: pinging on this one. Once the changes to create v2 snapshot from v3 are done, I would like to remove the store from RaftCluster. This code path for validation will block that change. Please let me know if the proposed solution as seen in the diff looks ok. @ptabor @serathius @ahrtr

geetasg:
rebased and squashed commits

server/etcdserver/api/membership/cluster.go
ahrtr: Sorry for the late response. I just get time to look into this today.

> 
> Let's consider situation when:
> 
> 1. Cluster is doing some operations on members, where foo already: add(foo), remove(foo),
> 2. In healthy situation the add(foo) fails on the all members and remove(foo) succeeds on all the members.
> 3. Now one of this members crashes before snapshot. The member manages to commit remove(foo) to it's backend.
> 4. The member starts recover from snaoshit & reply scenario
> 5. It will on the add(foo):
>  a) skip the validation (and even if not skipping, the validation would pass as foo is already removed in HEAD)
>  b) apply the add(foo) modification to RAFT -> (while the other members didn't applied it).
>  And this leads to a state when one RAFT member is different than the others...

At step 5, if it replays the `add(foo)`, it will also replay `remove(foo)`; eventually the state is also correct. Note intermediate inconsistent isn't a problem, because the member can't serve linearilizablity request if its data isn't in sync with the leader. @ptabor 

> if c.be != nil && shouldApplyV3 {
>     ......
> } else {
>     // line 325 ~ 327
>     membersMap, removedMap = c.rs.Members()
>     c.validateConfigurationChange(...)
>     ...
> }
> 

The else branch (line 325 ~ 327) seems not correct. If I understand your change correctly, the replayStore (c.rs) only contains new configure changes after the member gets started. In other words, it may not contain the complete membership data. 

I also do not understand why do you add a replayStore.

ahrtr:
> For me this is a bug fix that should be backported to v3.5. Reason is that v3.5 release was meant to switch etcd source of truth for membership from v2 to v3 store. 

Agreed to backport the change to 3.5. If a member crashes before creating the v2 snapshot, when it starts again, it loads the stale membership data from v2 snapshot. Eventually may run into data inconsistency issue.

geetasg:
Thanks @ahrtr for the review. I prepared - https://docs.google.com/document/d/1RqYbNBrRyMOwbfHIfWa60Aa_PlWdl5sGZyCTuUKfaSk/edit#heading=h.5akmx4nmafa3 - to explain my current understanding of the issue and potential solution. The replay store is to protect against regression of making invalid conf change during wal replay. 

Please let me know your thoughts on - 
* Do we need the replay store?
* The replay store should should be synced up to contain full membership info once the applied index catches up to consistent index?

cc @serathius @ptabor

ahrtr:
Per my understanding, the validation on the conf cange can be as simple as something like below. The principle is simple: we only validate it only when we are applying the conf change. Note we don't need to validate v2store, because we will eventually remove it in 3.6. 

```
func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
	if shouldApplyV3 {
		membersMap, removedMap := c.be.MustReadMembersFromBackend()
		if err := c.validateConfigurationChange(cc, membersMap, removedMap); err != nil {
			return err
		}
	} 

	return nil
}

```

geetasg:
Updated the PR to not apply past entries to backend. /cc @ahrtr @serathius @ptabor

geetasg:
I will squash the two commits and potentially refactor the tests into a separate PR if the approach is agreed upon.

ahrtr:
High level looks good to me. Ping all maintainers to take a look.

geetasg:
@serathius @ptabor ptal. thanks

ahrtr:
@geetasg Please tidy up this PR, i.e.  removing the commented out source code block. Afterwards, pls mark this PR as "ready for review", then I will take a closer look later. thx

server/etcdserver/api/membership/cluster.go
serathius: After reading through all the etcd membership changes (don't ask how long it took) I think I understand the problem (somewhat). High level the problem is not necessarily with validation, but which parts of RaftCluster state we treat as the "current". Note there are 3 states: cache (fields in RaftCluster struct, V2 store and V3 store). I will resolve the comment as the discussion diverged from the original problem and leave new comments on the code directly.

server/etcdserver/api/membership/cluster.go
```diff
@@ -420,29 +438,29 @@ func (c *RaftCluster) RemoveMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
-			zap.Bool("removed-remote-peer-is-learner", m.IsLearner),
-		)
-	} else {
-		c.lg.Warn(
-			"skipped removing already removed member",
-			zap.String("cluster-id", c.cid.String()),
-			zap.String("local-member-id", c.localID.String()),
-			zap.String("removed-remote-peer-id", id.String()),
-		)
+		m, ok := c.members[id]
```
serathius, line 442: ditto

server/etcdserver/api/membership/cluster.go
```diff
@@ -491,13 +509,13 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
 	if c.be != nil && shouldApplyV3 {
 		c.be.MustSaveMemberToBackend(c.members[id])
-	}
 
-	c.lg.Info(
-		"promote member",
-		zap.String("cluster-id", c.cid.String()),
-		zap.String("local-member-id", c.localID.String()),
-	)
+		c.lg.Info(
```
serathius, line 513: ditto

server/etcdserver/api/membership/cluster.go
```diff
@@ -510,16 +528,16 @@ func (c *RaftCluster) UpdateRaftAttributes(id types.ID, raftAttr RaftAttributes,
...
 
-	c.lg.Info(
-		"updated member",
-		zap.String("cluster-id", c.cid.String()),
-		zap.String("local-member-id", c.localID.String()),
-		zap.String("updated-remote-peer-id", id.String()),
-		zap.Strings("updated-remote-peer-urls", raftAttr.PeerURLs),
-		zap.Bool("updated-remote-peer-is-learner", raftAttr.IsLearner),
-	)
+		c.lg.Info(
```
serathius, line 532: ditto

server/etcdserver/api/membership/cluster.go
```diff
@@ -303,9 +303,27 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 
 // ValidateConfigurationChange takes a proposed ConfChange and
 // ensures that it is still valid.
-func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
+func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
 	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+	var membersMap map[types.ID]*Member
+	var removedMap map[types.ID]bool
+	if c.v2store != nil {
```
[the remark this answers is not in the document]
ahrtr: Yes, it's true that we only need to verify v3 just as I mentioned in https://github.com/etcd-io/etcd/pull/16084#issuecomment-1698814517. But just as I mentioned in one of previous community meetings, we should backport the change to 3.5, so let's keep the v2 validation for now. After the backport is done, we can remove it later. Please also see https://github.com/etcd-io/etcd/pull/16084#issuecomment-1696362494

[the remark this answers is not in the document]
ptabor: I'm not againt fixing 3.5 with backport in general, but it doesn't change our assumptions... 
Someone can rollback from 3.6.x to 3.5.0 and experience consequences of V2 store being stale (but still considered as source of truth). Seems we would need another safeguard that invalidates Store V2 in etcd 3.6 such that rollback to version that 'does not understand' the invalidation is not possible...

server/etcdserver/api/membership/cluster.go
```diff
@@ -303,9 +303,27 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
...
+
+	if c.be != nil && shouldApplyV3 {
+		membersMap, removedMap = c.be.MustReadMembersFromBackend()
+		if err := c.validateConfigurationChange(cc, membersMap, removedMap); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
ptabor, line 325: Can: 
```
membersMap map[types.ID]*Member
removedMap map[types.ID]bool
```
be properties of RaftClusters... as mutable arguments to functions are not readable. 

Like: 
  `currentReplayV2MembersMap` & `currentReplayV2removedMap...`

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
ahrtr: >  experience consequences of V2 store being stale (but still considered as source of truth)

Should we change the source of truth to V3 store (bbolt) in 3.5? I prefer to YES.

server/etcdserver/api/membership/cluster.go
serathius: Not sure I understand, do you want to propose to also validate the RaftCluster.membersMap and RaftCluster.removedMap? Note; before this PR don't validate them, we only validate v2 store.

server/etcdserver/api/membership/cluster.go
```diff
@@ -395,19 +413,19 @@ func (c *RaftCluster) AddMember(m *Member, shouldApplyV3 ShouldApplyV3) {
 	}
 	if c.be != nil && shouldApplyV3 {
 		c.be.MustSaveMemberToBackend(m)
-	}
 
-	c.members[m.ID] = m
-	c.updateMembershipMetric(m.ID, true)
+		c.members[m.ID] = m
```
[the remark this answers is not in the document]
ahrtr: Note that the v3 store (bbolt) is the[ source of truth](https://github.com/etcd-io/etcd/blob/e1ebc260a9bb609ad4c07f7691d8517d1a6c3ad1/server/etcdserver/api/membership/cluster.go#L263) for membership data. If `shouldApplyV3` is false, then it means that the member has already been added into the `c.members`; In other words, etcd just replays a WAL entry which has already been applied to bboltDB.

I think we should add a verification here, to verify that the member indeed has been added into `c.members`. This comment applies to the following couple of "ditto"s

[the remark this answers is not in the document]
serathius: > I think we should add a verification here, to verify that the member indeed has been added into c.members when shouldApplyV3 is false. This comment applies to the following couple of "ditto"s

This verification doesn't make sense as you can have `add` followed by `remove`. Etcd could restore from snapshot  before `add` command, but the v3 already has `remove` applied. Meaning verification will fail.

serathius:
I think review is spinning around in circles, same problems repeated in different context, things misattributed. I see two ways out of it:
* Write a proper design document
* Refactor the code that change is easier to understand.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
serathius: > Note that the v3 store (bbolt) is the source of truth for membership data.

True, but this is unrelated to validation changes.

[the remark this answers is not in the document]
ahrtr: > as you can have `add` followed by `remove`.

The member should be in either `c.members` or `c.removed`. The verification still makes sense?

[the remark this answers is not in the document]
ahrtr: > True, but this is unrelated to validation changes.

Do not understand what's your point. I was just clarifying your comment https://github.com/etcd-io/etcd/pull/16084#discussion_r1333212602 isn't valid.

ahrtr:
> I think review is spinning around in circles, same problems repeated in different context, things misattributed.

I am strongly (but politely) against raising vague & abstract comments, especially after a long long period that there wasn't any review comments. Pls raise direct comment on code regarding any technical concern.

serathius:
> I am strongly (but politely) against raising vague & abstract comments, especially after a long long period that there wasn't any review comments. Pls raise direct comment on code regarding any technical concern.

Asking for design is not vague or abstract, and there is no rush in review. We should prioritize correctness and clarity when making decisions. I unfortunately make the mistake of proposing the current strategy of v2 storage removal without writing down the design. 

The mistake is on me, so I will write down the design ASAP. Hope it will break clarity that will convince everyone to one proper solution.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
serathius: Let's revisit this after I write down the design, I still stand by that this change is incorrect even though v3 store is source of truth.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
ahrtr: > but it doesn't change our assumptions...
> Someone can rollback from 3.6.x to 3.5.0 and experience consequences of V2 store being stale (but still considered as source of truth). 
> 

Yes, it's true that the v2 store of **3.5** might be stale right after bootstrapping (no matter it's downgraded from 3.6, or just 3.5 restarts), because 3.5 loads membership data from snapshot which might be stale; but I do not see any issue.
- Eventually the membership data in v2 store will be in sync with v3 store when the replaying WAL entries is done;
- Note etcd 3.5 isn't able to serve client requests until it [finishes publishing the server info to cluster through raft](https://github.com/etcd-io/etcd/blob/v3.5.9/server/embed/serve.go#L101), which must be applied after the replaying legacy WAL entries.

So 
- My previous comment https://github.com/etcd-io/etcd/pull/16084#issuecomment-1696362494 is also invalid.
- We don't have to change the [source of truth for membership data](https://github.com/etcd-io/etcd/blob/v3.5.9/server/etcdserver/api/membership/cluster.go#L259-L265) to V3 store (bbolt) in **3.5**.

@ptabor please kindly let me know if I misunderstood your point or you have any other concern.

ahrtr:
I don't think this PR is that complicated, although it took me some time to clarify some concerns/comments. Previously etcd (main branch) only validates v2 store when applying conf change, now it validates both v2 and v3 stores [Of course, eventually we will remove v2 validation in 3.6 after backporting the PR to 3.5]. Overall it looks good. But the applying process is the most important workflow, so any technical insights or comments is welcome; and we are happy & open to clarify & discuss.

If anyone has any comment on the overall v2 deprecation task, please let's discuss under the umbrella ticket https://github.com/etcd-io/etcd/issues/12913.

One minor improvement I can think of for now is adding the verification just as I mentioned in https://github.com/etcd-io/etcd/pull/16084#discussion_r1334300585. Any suggestion on how to improve the test & verification is welcome!

With regard the concern for **3.5** of bootstrapping on stale v2 store (no matter it's downgraded from 3.6, or just 3.5 restarts), please see my comment https://github.com/etcd-io/etcd/pull/16084#discussion_r1336087754.

geetasg:
squashed commits and updated to latest

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
serathius: @ahrtr you are missing one important consequence of changing validation. Normal raft proposal (like PUT) can be represented as a state machine without side effect. It means that we can care about only the end state, so assuming that we all members execute same WAL entries, we are good. This however is not true for membership.

Membership "state machine" has side effect in form of ConfState, which is state of truth for raft voting members. When applying the entry we run the validation function against the membership info to decide if it should have an effect on ConfState. This means that we not only care about the state, but also transitions. Every time we replay WAL we need to guarantee that validate behaves the same. That also applies to +1/-1 etcd versions to support upgrades.

serathius:
>  [Of course, eventually we will remove v2 validation in 3.6 after backporting the PR to 3.5] 

I'm against backporting to v3, as it doesn't bring any benefits just creates risk. We can remove v2 store in v3.6 without needing any changes to v3.5. Any design that would depend on backporting changes to v3.5 would be flawed, as for upgrade we cannot assume that users upgrade to latest v3.5 before upgrading to v3.6.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
ahrtr: In term of "side effect on ConfState", I assume you are talking about [server.go#L1989](https://github.com/etcd-io/etcd/blob/e85949db3a05dc7f746aa8c6c6416a9822d2a779/server/etcdserver/server.go#L1989) or [server.go#L2001](https://github.com/etcd-io/etcd/blob/e85949db3a05dc7f746aa8c6c6416a9822d2a779/server/etcdserver/server.go#L2001). Pls let me know if we are not on the same page. 

Note that I don't think this PR changes the validation logic on membership data. It just applies the validation on v3 store as well. If the validation on v2 store fails (or succeeds) [Of course, we will remove the validation on v2], then the validation on v3 should fails (or succeeds) as well. 

Pls be more specific if you still think there is any issue.

Again, just as I mentioned in https://github.com/etcd-io/etcd/pull/16084#issuecomment-1734036435, since this is the most important part, so any suggestion on how to improve the test cases is welcome.

ahrtr:
> Any design that would depend on backporting changes to v3.5 would be flawed, as for upgrade we cannot assume that users upgrade to latest v3.5 before upgrading to v3.6.

Note that the PR or design doesn't depend on backporting to 3.5 at all. They are totally independent. No matter we backport the PR or not, it will not block the task of v2 deprecation in 3.6.

> I'm against backporting to v3, as it doesn't bring any benefits just creates risk. We can remove v2 store in v3.6 without needing any changes to v3.5.

It's definitely nice to have to validate v3 store as well in 3.5 when applying conf change. Read https://github.com/etcd-io/etcd/pull/16084#discussion_r1336989792 as well. If the validation on v2 store succeeds (or fails), but the validation on v3 store fails (or succeeds), then it means the cluster has issue.

serathius:
Me and @ahrtr met and looked at the problem together. New thing we discovered was that there was already an attempt to migrate to v3 store, however it was unfinished https://github.com/etcd-io/etcd/pull/12914. This led to a big problem in v3.5.0 release https://github.com/etcd-io/etcd/issues/13196 and required dedicated patch https://github.com/etcd-io/etcd/pull/13348 that reverted the switch. Unfortunatelly the fix was not backported, so currently v3.6 is in half broken state. 

I think everyone agrees on the end goal, migrate etcd v3.6 to v3 store, however there are many ways of doing it. Like this PR that does only part of migration. It switches RaftCluster to v3, however WAL is still replayed from last snapshot.

My main learning from https://github.com/etcd-io/etcd/pull/12914 is fact that efforts can be de-prioritized, forgotten and abandoned. For efforts like migration to v3, we should ensure that any PR we merge is correct and complete on it's own and assume that it could be the last one. As so I would prefer to not change validation without migrating off v2 store. I will send a draft PR to show what I mean.

serathius:
https://github.com/etcd-io/etcd/pull/16656 the draft. It's goal is to move etcd bootstrap to be from v3 and not last snapshot. It doesn't touch RaftCluster apart of validation that needs to be switched to v3 now.

ahrtr:
Actually this PR includes a bug fix and a feature. We should breakdown this PR into two dedicated PRs, so that it's clearer.
- bug fix: please see https://github.com/etcd-io/etcd/pull/16658
- a feature: add validation on v3 store. Leave it to @geetasg to continue to work on this PR.

Regarding https://github.com/etcd-io/etcd/pull/16656, it's totally an another big topic, [although it's sort of related.], I suggest not to complicate the conversation and discussion for now. We can discuss it separately after we finish the v2 deprecation task.

For the concern of https://github.com/etcd-io/etcd/pull/13348 not being forward ported to 3.6 (main branch), please see https://github.com/etcd-io/etcd/pull/16655#issuecomment-1737200035.

I imagine @geetasg is going to finish the v2 deprecation task in the following 3~5 PRs. I agree that It would be better if there is a design documentation, so that everyone is on the same page and no any small task will be forgotten.

Issue #15913, opened 2023-05-16
High flakiness of Robustness nightly 


serathius:
### Which github workflows are flaking?

![image](https://github.com/etcd-io/etcd/assets/5873433/9fc4e259-3bd1-407f-aa55-0d549bcb3371)

High as we have at least 1 scenario failure in run, however it's not consistent.

![image](https://github.com/etcd-io/etcd/assets/5873433/0963408c-b822-4c0a-99c4-a9f45fa1fc57)

### Which tests are flaking?

Looks like linearizability times out. I expect it's due one of my recent changes, however wanted to put the issue for grabs as it's not super critical to fix. It's still doesn't impact presubmit tests.

fuweid:
If no one takes this, I would like to help. (trying to catch-up the recently changes) Thanks

ahrtr:
Thanks  @fuweid .  

Recently there are big & fast changes (too many PRs) on the robustness test. I don't think it's a good thing to make too many big changes too fast, even it's just test code. Since almost each PR is huge ( e.g. > 300 lines of change), it's really hard to carefully review each line of change.

Fast is definitely not equal to good, especially for etcd which is about 10 years old. 

I propose (not limited to the robustness test):
- Slowdown the speed. Can we submit at most one PR at a time for each code area each person? 
- Please try to deliver small PRs as much as possible. 
- Fixing issues (including product issue and test issues) should be always the top priority by default.

chaochn47:
Agreed. Sometimes I saw some PRs are reviewed and merged over the weekend that other etcd members did not get a chance to review, reason about / understand / catch up the changes. 

To share my experience, for example https://github.com/etcd-io/etcd/pull/15888 was raised in Saturday early morning 3 AM and get merged at Monday early morning 3 AM in PDT Time zone.

jmhbnz:
Thanks for suggesting a little slower pace @ahrtr.  I have been spending time every day to try and look at all the pr's and  I have found it tough to keep up and review as thoroughly as I would like to. I have been thinking along the same lines as your suggestions but didn't say anything because I figured it is just because I am newer to the project.

It's awesome that the robustness suite is getting so much improvement though and huge kudos to @serathius for all the work, maybe just a little slower, or leave some pr's as draft or open longer before they merge 🙏🏻

serathius:
> Thanks @fuweid .
> 
[...]

PR #16085, opened 2023-06-15
tests/robustness: Disable blackhole until snapshot for v3.5 and v3.4

serathius:
Part of https://github.com/etcd-io/etcd/issues/15913

### DIFF
--- a/tests/robustness/linearizability_test.go
+++ b/tests/robustness/linearizability_test.go
@@ -102,12 +102,14 @@ func TestRobustness(t *testing.T) {
 			e2e.WithClusterSize(1),
 		),
 	})
-	if v.Compare(version.V3_5) >= 0 {
+	// TODO: Deflake waiting for waiting until snapshot for etcd versions that don't support setting snapshot catchup entries.
+	if v.Compare(version.V3_6) >= 0 {
 		scenarios = append(scenarios, testScenario{
 			name:      "Issue15271",
 			failpoint: BlackholeUntilSnapshot,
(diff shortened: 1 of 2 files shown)


serathius:
cc @jmhbnz @ptabor @chaochn47 @ahrtr

tests/robustness/linearizability_test.go
```diff
@@ -102,12 +102,14 @@ func TestRobustness(t *testing.T) {
 			e2e.WithClusterSize(1),
 		),
 	})
-	if v.Compare(version.V3_5) >= 0 {
+	// TODO: Deflake waiting for waiting until snapshot for etcd versions that don't support setting snapshot catchup entries.
```
chaochn47, line 105: How about? 

```
// Disable BlackholeUntilSnapshot failpoint for etcd versions that don't support setting smaller snapshot catchup entries. Default is 5000 which is too long to wait in robustness test. 
```

The reason I propose removing TODO is I don't see how to address the TODO, either wait longer for 5000 entries or externalize snapshot catchup entries as etcd flags seems not actionable right now.

chaochn47 (approved):

chaochn47:
LGTM. Approved for deflaking https://github.com/etcd-io/etcd/issues/15913

tests/robustness/linearizability_test.go
serathius: Still we should address it, I would even consider backporting the flag (maybe hidden) just to be able to test for regressions.

PR #16086, opened 2023-06-16
etcdserver: add tests for apply_auth.go

CaojiamingAlan:
Achieved 100% test coverage for apply_auth.go.
image

The test cases focus on the authorization itself, can verifies operations can only be performed by proper users and roles. The correctness of the applying result is not checked by these tests.

### DIFF
--- a/server/etcdserver/apply/apply_auth_test.go
+++ b/server/etcdserver/apply/apply_auth_test.go
@@ -15,47 +15,791 @@
 package apply
 
 import (
-	"encoding/base64"
+	"context"
 	"testing"
 	"time"
 
-	"golang.org/x/crypto/bcrypt"
-
-	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
-
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
-
-	"go.etcd.io/etcd/server/v3/storage/schema"
+	"golang.org/x/crypto/bcrypt"
 
 	"go.etcd.io/etcd/api/v3/authpb"
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/membershippb"
+	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/auth"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/v3alarm"
+	"go.etcd.io/etcd/server/v3/etcdserver/cindex"
 	"go.etcd.io/etcd/server/v3/lease"
+	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
+	"go.etcd.io/etcd/server/v3/storage/schema"
 )
 
-func TestCheckLeasePutsKeys(t *testing.T) {
+func dummyIndexWaiter(_ uint64) <-chan struct{} {
+	ch := make(chan struct{}, 1)
+	ch <- struct{}{}
+	return ch
+}
+
+func dummyApplyFunc(_ context.Context, _ *pb.InternalRaftRequest, _ membership.ShouldApplyV3) *Result {
+	return &Result{}
+}
+
+type fakeRaftStatusGetter struct{}
+
+func (*fakeRaftStatusGetter) MemberId() types.ID {
+	return 0
+}
+func (*fakeRaftStatusGetter) Leader() types.ID {
+	return 0
+}
+func (*fakeRaftStatusGetter) CommittedIndex() uint64 {
+	return 0
+}
+func (*fakeRaftStatusGetter) AppliedIndex() uint64 {
+	return 0
+}
+func (*fakeRaftStatusGetter) Term() uint64 {
+	return 0
+}
+
+type fakeSnapshotServer struct{}
+
+func (*fakeSnapshotServer) ForceSnapshot() {}
+
+func defaultAuthApplierV3(t *testing.T) *authApplierV3 {
 	lg := zaptest.NewLogger(t)
+	be, _ := betesting.NewDefaultTmpBackend(t)
+	t.Cleanup(func() {
+		betesting.Close(t, be)
+	})
 
-	b, _ := betesting.NewDefaultTmpBackend(t)
-	defer betesting.Close(t, b)
+	cluster := membership.NewCluster(lg)
+	lessor := lease.NewLessor(lg, be, cluster, lease.LessorConfig{})
+	kv := mvcc.NewStore(lg, be, lessor, mvcc.StoreConfig{})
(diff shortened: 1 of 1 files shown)


CaojiamingAlan:
I would also like to write tests for other appliers, but most of the code of `applierV3backend` is  simply calling its backend and therefore not interesting :(    
For example:
https://github.com/etcd-io/etcd/blob/93e6bce01dd669da38f32299de4983db1dee8d03/server/etcdserver/apply/apply.go#L160-L162

However the test coverage for those dependencies is also poor. For example: [txn.Range](https://github.com/etcd-io/etcd/blob/93e6bce01dd669da38f32299de4983db1dee8d03/server/etcdserver/txn/txn.go#L117)

image

I think it is also good if we try to cover more branches of these hot paths.

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -0,0 +1,846 @@
...
+	"go.etcd.io/etcd/server/v3/lease"
+	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
+	"go.etcd.io/etcd/server/v3/storage/schema"
+)
+
+func dummyIndexWaiter(_ uint64) <-chan struct{} {
+	ch := make(chan struct{}, 1)
+	go func() {
+		ch <- struct{}{}
```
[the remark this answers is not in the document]
tjungblu: I think this is just a dummy. It's actually in several places of our testing code. I had to copy-paste this too:

https://github.com/etcd-io/etcd/commit/cb3730a30ff49bd5481d5306960e6c6d573661e2#diff-3736d70424fdcd4f4bc1f5a79b751d32c987bbf28fb93a6d90ab5de80ac35259R45-R51

serathius:
Please resolve conflict with https://github.com/etcd-io/etcd/pull/16005

server/etcdserver/apply/apply_auth_test.go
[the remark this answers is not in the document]
fuweid: Thanks for the information. I think we can just assign the item into channel without goroutine, which is easy to review. Nice to have.

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -15,47 +15,844 @@
...
+
+	authApplier := defaultAuthApplierV3(t)
+	MustCreateRolesAndEnableAuth(t, authApplier)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			SetAuthInfo(authApplier, tc.userName)
+			_, _, err := authApplier.Put(ctx, nil, tc.request)
+			if (err != nil) != tc.expectError {
```
tjungblu, line 459: you can simplify the two if's by just checking whether there was an error:

> require.Equal(t, tc.expectError, err)

it will handle nil's just fine, no need to dance around with the booleans. You can also equality check the error itself or leave it as nil when there's no error.

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -15,47 +15,844 @@
...
+	MustCreateRolesAndEnableAuth(t, authApplier)
+	dummyApplyFunc := func(ctx context.Context, r *pb.InternalRaftRequest, shouldApplyV3 membership.ShouldApplyV3) *Result {
+		return &Result{}
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			result := authApplier.Apply(ctx, tc.request, false, dummyApplyFunc)
+			if !reflect.DeepEqual(result, tc.expectResult) {
```
tjungblu, line 221: just double checked the code, the `require.Equal` will deepEqual internally as well, you can save yourself that if/fatal too

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -15,47 +15,844 @@
...
+				},
+				AuthEnable: &pb.AuthEnableRequest{},
+			},
+			expectResult: &Result{},
+		},
+	}
+
+	authApplier := defaultAuthApplierV3(t)
+	MustCreateRolesAndEnableAuth(t, authApplier)
+	dummyApplyFunc := func(ctx context.Context, r *pb.InternalRaftRequest, shouldApplyV3 membership.ShouldApplyV3) *Result {
```
tjungblu, line 213: I think you have this defined twice below, maybe extract this into a variable? like the dummy writer on top

tjungblu (changes requested):
test cases look great :+1:  I think you can simplify your assertions a little more, otherwise lgtm

CaojiamingAlan:
@fuweid @tjungblu Thank you for your comments. I have resolved them. @tjungblu No need to feel sorry about that, it was me who did not notice you had written a test XD

serathius (approved):
Wow, the most comprehensive auth test suite I have seen.

tjungblu (approved):
yep, well done! /lgtm (non-binding)

ahrtr:
Sorry for the late response. This PR is far behind the main branch.  Could you rebase this PR although github doesn't display any conflict?

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -15,47 +15,791 @@
...
-		go func() {
-			ch <- struct{}{}
-		}()
-		return ch
+	_, err = authApplier.AuthEnable()
+	require.NoError(t, err)
+}
+
+// SetAuthInfo manually sets the authInfo of the applier. In reality, authInfo is filled before Apply()
+func SetAuthInfo(authApplier *authApplierV3, userName string) {
```
ahrtr, line 166: Why it's an exported function?
```suggestion
func setAuthInfo(authApplier *authApplierV3, userName string) {
```

server/etcdserver/apply/apply_auth_test.go
```diff
@@ -15,47 +15,791 @@
...
+	rangeEnd        = "rangeEnd"
+	keyOutsideRange = "rangeEnd_outside"
+
+	LeaseId = 1
+)
+
+func mustCreateRolesAndEnableAuth(t *testing.T, authApplier *authApplierV3) {
+	_, err := authApplier.UserAdd(&pb.AuthUserAddRequest{Name: userRoot, Options: &authpb.UserAddOptions{NoPassword: true}})
+	require.NoError(t, err)
+	_, err = authApplier.RoleAdd(&pb.AuthRoleAddRequest{Name: userRoot})
```
ahrtr, line 130:
```suggestion
	_, err = authApplier.RoleAdd(&pb.AuthRoleAddRequest{Name: roleRoot})
```

ahrtr (approved):
LGTM

Thanks @CaojiamingAlan

jmhbnz (approved):
LGTM

@ahrtr I'm curious is `stage/merge-when-tests-green` something that should happen automatically or does the merge still need to be done manually?

serathius:
> @ahrtr I'm curious is stage/merge-when-tests-green something that should happen automatically or does the merge still need to be done manually?

It's mostly a reminder and a note to other maintainers

PR #16091, opened 2023-06-16
tests/robustness: Implement stale reads without validation

serathius:
For now we just validate stale read revision, but not response content. Reason is that etcd model only stores latest version of keys, and no history like real etcd.

Validating stale read contents needs to be done outside of model as storing whole history is just to costly for linearization validation.

### DIFF
--- a/tests/robustness/model/deterministic.go
+++ b/tests/robustness/model/deterministic.go
@@ -16,6 +16,7 @@ package model
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"hash/fnv"
 	"reflect"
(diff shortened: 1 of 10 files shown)


serathius:
ping @jmhbnz @ahrtr @chaochn47

tests/robustness/traffic/etcd.go
```diff
@@ -105,87 +110,120 @@ func (t etcdTraffic) Run(ctx context.Context, c *RecordingClient, limiter *rate.
...
-
-func (t etcdTraffic) RandomOperation(ctx context.Context, c *RecordingClient, limiter *rate.Limiter, key string, id identity.Provider, lm identity.LeaseIdStorage) error {
+func (t etcdTraffic) Request(ctx context.Context, c *RecordingClient, request etcdRequestType, limiter *rate.Limiter, key string, id identity.Provider, lm identity.LeaseIdStorage, lastRev int64) (rev int64, err error) {
 	opCtx, cancel := context.WithTimeout(ctx, RequestTimeout)
 
-	var err error
-	switch pickRandom(t.operations) {
+	switch request {
+	case StaleGet:
+		_, rev, err = c.Get(opCtx, key, lastRev)
```
fuweid, line 135: No sure that I understand it correctly: The `lastRev` is from last command. It seems like the `lastRev` is always latest one. Is it still considered as stale?

serathius: It's the latest one for the client, but with multiple clients it's always a little stale 1-5 revisions behind the latest request.

tests/robustness/model/deterministic.go
```diff
@@ -93,8 +94,15 @@ func (s etcdState) step(request EtcdRequest) (etcdState, MaybeEtcdResponse) {
...
 	switch request.Type {
 	case Range:
-		resp := s.getRange(request.Range.Key, request.Range.RangeOptions)
-		return s, MaybeEtcdResponse{EtcdResponse: EtcdResponse{Range: &resp, Revision: s.Revision}}
+		if request.Range.Revision == 0 || request.Range.Revision == s.Revision {
+			resp := s.getRange(request.Range.Key, request.Range.RangeOptions)
+			return s, MaybeEtcdResponse{EtcdResponse: EtcdResponse{Range: &resp, Revision: s.Revision}}
+		} else {
+			if request.Range.Revision > s.Revision {
+				return s, MaybeEtcdResponse{Err: EtcdFutureRevErr}
```
jmhbnz, line 102: Is this just a safety catch or are there some fully expected situations where this future rev is going to be hit?  Apologies if I am missing something obvious.

serathius: It's a dummy error that should not be hit in the current code, still in future we will want to generate traffic that might hit a future revision and test it.

jmhbnz (approved):
LGTM

PR #16093, opened 2023-06-16
Ensure goimports can be fixed individually

jmhbnz:
Sub task of https://github.com/etcd-io/etcd/issues/13896.

This pull request proposes a new `Makefile` target `fix-goimports` to ensure this can be run individually. As a side effect `scripts/fix.sh` becomes less monolithic which if I understand correctly is part of the intent of https://github.com/etcd-io/etcd/issues/13896.

Additionally I noticed a couple of `Makefile` `PHONY` target names did not match their child target names so I have aligned these under a second commit.


tjungblu (approved):
Thank you, I was fiddling with the `go run` command the whole time

/lgtm (non-binding)

serathius (approved):

serathius (approved):

PR #16096, opened 2023-06-16
tests/robustness: Limit model to start only from fresh state

serathius:
It is just to complicated to support starting from non-empty etcd. Existing implementation was very naive to assume that we can build full state from just one request. We might consider implementing validation of non-empty history in future, but for now settting this limit should clean up the code and speed up development.


chaochn47 (approved):
LGTM. Thanks @serathius!

PR #16100, opened 2023-06-18
Verify consistent index is latest at the time of snapshot

geetasg:

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1104,6 +1104,14 @@ func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
 	})
 }
 
+func verifyConsistentIndexIsLatest(lg *zap.Logger, snapshot raftpb.Snapshot, cindex uint64) {
+	verify.Verify(func() {
+		if cindex < snapshot.Metadata.Index {
+			lg.Panic(fmt.Sprintf("consistent_index(%d) is older than snapshot index (%d)", cindex, snapshot.Metadata.Index))
+		}
+	})
+}
+
 func (s *EtcdServer) applyEntries(ep *etcdProgress, apply *toApply) {
 	if len(apply.entries) == 0 {
 		return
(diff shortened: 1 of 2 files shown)


server/etcdserver/server.go
```diff
@@ -2082,6 +2090,9 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
 			}
 			lg.Panic("failed to create snapshot", zap.Error(err))
 		}
+
+		verifyConsistentIndexIsLatest(lg, snap, s.consistIndex.ConsistentIndex())
```
jmhbnz, line 2094: How does this new verify function call relate to the snippet below from earlier in the function?

```go
// commit kv to write metadata (for example: consistent index) to disk.
// This guarantees that Backend's consistent_index is >= index of last snapshot.
s.KV().Commit()
```

Are we saying that the commented guarantee is no longer accurate or is there some kind of timing issue meaning we need to reverify?

geetasg: This change is codifying/verifying the comment.

jmhbnz: Right - given the strong language in the comment I had assumed there was already some verification happening somewhere.  Thanks for raising this.

jmhbnz (approved):

chaochn47 (approved):

server/etcdserver/server.go
```diff
@@ -1104,6 +1104,14 @@ func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
 	})
 }
 
+func verifyConsistentIndexIsLatest(lg *zap.Logger, snapshot raftpb.Snapshot, cindex uint64) {
```
ahrtr, line 1107: Minor comment: we should consider to make this function has the same signature as `verifySnapshotIndex`.

server/etcdserver/server.go
ahrtr, line 2094: Since there isn't any real key/value data in v2 store anymore, probably it doesn't make much sense to perform the snapshot in a separate goroutine any more; then the cindex should be equal to snapshot index, accordingly we should use `verifySnapshotIndex` directly instead of adding a new function `verifyConsistentIndexIsLatest`

ahrtr (approved):
LGTM

Thanks @geetasg

ahrtr (approved):
LGTM

Thanks @geetasg 

We need to cleanup or simplify the `(*EtcdServer)snapshot` as tracked in https://github.com/etcd-io/etcd/issues/12913

PR #16115, opened 2023-06-20
tests/robustness: Tune Kubernetes tests to reduce number of delete requests

serathius:
Having too many delete requests is bad as they are not unique requests, so linearization is more prone to timeout on them.

Going from 30% to 10%.

cc @ahrtr @ptabor @fuweid @jmhbnz @chaochn47


jmhbnz (approved):
LGTM

fuweid (approved):
LGTM

PR #11838, opened 2020-05-02
clientv3: v3Manager#saveDB should always close db file

tedyu:
In v3Manager#saveDB, db file should always be closed upon return from the method.

Currently if db.Seek() call fails (around line 353), db is not closed.

PR #16120, opened 2023-06-21
etcdutl: fix db double closed

Tachone:
#11838 fix `db` may not be closed，but  in `768da4` missing line: `dbClosed = true`（`git bisect` is very useful ）

if `copyAndVerifyDB` return nil, `db` is closed twice.

### DIFF
--- a/etcdutl/snapshot/v3_snapshot.go
+++ b/etcdutl/snapshot/v3_snapshot.go
@@ -340,13 +340,8 @@ func (s *v3Manager) copyAndVerifyDB() error {
 	if dberr != nil {
 		return dberr
 	}
-	dbClosed := false
-	defer func() {
-		if !dbClosed {
-			db.Close()
-			dbClosed = true
-		}
-	}()
+	defer db.Close()
+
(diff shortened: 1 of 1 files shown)


etcdutl/snapshot/v3_snapshot.go
```diff
@@ -384,6 +384,8 @@ func (s *v3Manager) copyAndVerifyDB() error {
 
 	// db hash is OK, can now modify DB so it can be part of a new cluster
 	db.Close()
+	dbClosed = true
```
ahrtr, line 387: probably we can just remove line 386 (`db.Close()`) and `dbClosed` (line 343)

Tachone: yes, it's easier that way, updated.

ahrtr (approved):
LGTM

Thanks @Tachone

chaochn47 (approved):

ahrtr:
@Tachone could you backport this PR to 3.5?

Tachone:
> @Tachone could you backport this PR to 3.5?

Sure. #16201

PR #16122, opened 2023-06-21
tests/robustness: Add List and StaleList requests to etcd traffic

serathius:
cc @ahrtr @ptabor @jmhbnz @fuweid @chaochn47


chaochn47 (approved):
LGTM Thanks @serathius. 

IMHO, `List` concept is not widely adopted in etcd client but in kube-apiserver. `RangeWithPrefix` Maybe a better name in etcd traffic.

serathius:
Makes sense

PR #16132, opened 2023-06-23
Add a method to export membership info to v2 store from RaftCluster

geetasg:
Introducing method to create the snapshot in v2 format using v3 state. This is used by the backup command in this PR. Main snapshot path will be updated in subsequent PRs.

### DIFF
--- a/etcdutl/etcdutl/backup_command.go
+++ b/etcdutl/etcdutl/backup_command.go
@@ -28,9 +28,9 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/pkg/v3/idutil"
 	"go.etcd.io/etcd/pkg/v3/pbutil"
+	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
-	"go.etcd.io/etcd/server/v3/etcdserver/api/v2store"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	"go.etcd.io/etcd/server/v3/storage/datadir"
 	"go.etcd.io/etcd/server/v3/storage/schema"
@@ -167,7 +167,7 @@ func saveSnap(lg *zap.Logger, destSnap, srcSnap string, desired *desiredCluster)
 		walsnap.Index, walsnap.Term, walsnap.ConfState = snapshot.Metadata.Index, snapshot.Metadata.Term, &desired.confState
 		newss := snap.New(lg, destSnap)
 		snapshot.Metadata.ConfState = desired.confState
-		snapshot.Data = mustTranslateV2store(lg, snapshot.Data, desired)
+		snapshot.Data = mustTranslateV2store(lg, desired)
 		if err = newss.SaveSnap(*snapshot); err != nil {
 			lg.Fatal("saveSnap(Snapshoter.SaveSnap) failed", zap.Error(err))
 		}
@@ -175,24 +175,12 @@ func saveSnap(lg *zap.Logger, destSnap, srcSnap string, desired *desiredCluster)
 	return walsnap
 }
 
-// mustTranslateV2store processes storeData such that they match 'desiredCluster'.
-// In particular the method overrides membership information.
-func mustTranslateV2store(lg *zap.Logger, storeData []byte, desired *desiredCluster) []byte {
-	st := v2store.New()
-	if err := st.Recovery(storeData); err != nil {
-		lg.Panic("cannot translate v2store", zap.Error(err))
-	}
-
+// mustTranslateV2store returns membership info matching 'desiredCluster' in v2 format.
+func mustTranslateV2store(lg *zap.Logger, desired *desiredCluster) []byte {
 	raftCluster := membership.NewClusterFromMembers(lg, desired.clusterId, desired.members)
 	raftCluster.SetID(desired.nodeId, desired.clusterId)
-	raftCluster.SetStore(st)
-	raftCluster.PushMembershipToStorage()
-
-	outputData, err := st.Save()
-	if err != nil {
-		lg.Panic("cannot save v2store", zap.Error(err))
-	}
-	return outputData
+	d := etcdserver.GetMembershipInfoInV2Format(lg, raftCluster)
+	return d
 }
 
 func translateWAL(lg *zap.Logger, srcWAL string, walsnap walpb.Snapshot) (etcdserverpb.Metadata, raftpb.HardState, []raftpb.Entry) {
--- a/server/etcdserver/api/membership/cluster_test.go
+++ b/server/etcdserver/api/membership/cluster_test.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap/zaptest"
 
 	"go.etcd.io/etcd/client/pkg/v3/testutil"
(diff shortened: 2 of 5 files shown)


server/etcdserver/api/membership/cluster_test.go
```diff
@@ -975,3 +976,29 @@ func TestIsReadyToPromoteMember(t *testing.T) {
...
+	for _, tt := range tests {
+		c := newTestCluster(t, tt.mems)
+		c.removed = tt.removed
+
+		st := v2store.New("/0", "/1")
+		c.Store(st)
+		mst, rst := membersFromStore(c.lg, st)
+		assert.Equal(t, mst[types.ID(1)], tt.mems[0])
+		assert.Equal(t, rst, tt.removed)
+	}
```
ahrtr, line 1003: Please use go subtest.

ahrtr:
Overall looks good to me with a minor comment.

server/etcdserver/api/membership/storev2.go
```diff
@@ -94,7 +94,7 @@ func mustSaveMemberToStore(lg *zap.Logger, s v2store.Store, m *Member) {
 }
 
 func mustDeleteMemberFromStore(lg *zap.Logger, s v2store.Store, id types.ID) {
-	if _, err := s.Delete(MemberStoreKey(id), true, true); err != nil {
+	if _, err := s.Delete(MemberStoreKey(id), true, true); err != nil && !isKeyNotFound(err) {
```
serathius, line 97: Don't change this function. The function above `RemoveMember` assumes `isKeyNotFound` causes panic.

ahrtr: It isn't a big problem.

- Actually it doesn't matter. Even time when applying conf change, etcd will validate the conf change(search `ValidateConfigurationChange`). It will never run into panic. It's exactly the reason why we never see panic (bug) before
- We will remove the code of adding or removing members into v2store in following PRs soon.

ahrtr:
Please also rebase this PR since it's far behind the head of main.

server/etcdserver/api/membership/storev2.go
ahrtr: @geetasg Please revert this change, and raise a separate PR, so as not be blocked on this non-important thing.

geetasg:
@ahrtr @serathius thank you for the review. The change to not panic on keyNotFound is primarily driven by the new workflow to export membership to a new store (for snapshot). This store is empty and does not contain the members and will get a keyNotFound if a delete is attempted. The code is now updated to avoid changing mustDeletefromStore method. ptal. Thanks.

geetasg:
cc @ptabor

server/etcdserver/api/membership/storev2.go
```diff
@@ -78,6 +78,13 @@ func TrimMembershipFromV2Store(lg *zap.Logger, s v2store.Store) error {
 	return nil
 }
 
+func verifyNoMembersInStore(lg *zap.Logger, s v2store.Store) {
+	members, removed := membersFromStore(lg, s)
+	if len(members) != 0 || len(removed) != 0 {
+		lg.Panic("store has membership info")
+	}
+}
```
ahrtr, line 86: This verification makes sense, but should be in a following PR which removes all the code of adding/removing member in v2store.

etcdutl/etcdutl/backup_command.go
```diff
@@ -178,21 +178,10 @@ func saveSnap(lg *zap.Logger, destSnap, srcSnap string, desired *desiredCluster)
 // mustTranslateV2store processes storeData such that they match 'desiredCluster'.
```
serathius, line 178: You are no longer using `storeData`. Please update the comment and function arguments.

server/etcdserver/api/membership/storev2.go
geetasg: @ahrtr this is being called only from the RaftCluster Store function which must get an empty store and is only invoked from GetMembershipInfoInV2Format today. This verification is added because the code to assume that member deletion from the store is not needed - it directly adds the removed members to the RemovedMembers prefix in the store. 
I can move the verification to subsequent PR as per review feedback but wanted to clarify that it is protecting the correctness of logic to handle removed members. Please let me know if you would still like me to move this out.

etcdutl/etcdutl/backup_command.go
geetasg: updated

etcdutl/etcdutl/backup_command.go
```diff
@@ -167,32 +167,20 @@ func saveSnap(lg *zap.Logger, destSnap, srcSnap string, desired *desiredCluster)
 		walsnap.Index, walsnap.Term, walsnap.ConfState = snapshot.Metadata.Index, snapshot.Metadata.Term, &desired.confState
 		newss := snap.New(lg, destSnap)
 		snapshot.Metadata.ConfState = desired.confState
-		snapshot.Data = mustTranslateV2store(lg, snapshot.Data, desired)
+		snapshot.Data = mustTranslateV2store(lg, desired)
```
serathius, line 170: Why are we changing `etcdutl backup` command at all? It's a legacy deprecated way to create snapshot.

server/etcdserver/api/membership/storev2.go
ahrtr: > @ahrtr this is being called only from the RaftCluster Store function which must get an empty store

Yes, I agree. But the `(*RaftCluster) Store` will be called in multiple places in following PR(s). The verification will definitely fail **after** you add more calling of the Store and **before** you remove the code of adding/removing members data in v2store. Currently you only call the `Store` in the etcdutl backup command, so you do not see any verification failure.

A better way should be get it included in a follow PR as the first commit:
- add the verification. We should get a failed verification;
- remove all the code of adding/removing/updating in v2store in the applying code path. Then the verification should be all good.

Since we will resolve the big task "v2store deprecation" in lots of small PRs, let's try to get each PR as simple and standalone as possible.

geetasg: >>add the verification. We should get a failed verification;
Adding more clarification - The Store method will not get called directly. It will get called from GetMembershipInfoInV2Format  which should pass empty store as an argument to Store method. The verification should not fail.

etcdutl/etcdutl/backup_command.go
geetasg: This set of PRs is going towards removing the SetStore method of RaftCluster (along with usage of that store). The backup command is a consumer of SetStore so it needs to be updated. 

If the backup command needs to be removed from code base, I can open another issue to get that done and inform the user base that the backup command is going away in 3.6. Please let me know if the backup command should be removed. /cc @ahrtr

geetasg: created:https://github.com/etcd-io/etcd/issues/16321

server/etcdserver/api/membership/storev2.go
ahrtr: ack. I messed this v2store with [server.go#L252](https://github.com/etcd-io/etcd/blob/9f72c648d3829810f4256ed1868fcf1c7665aa44/server/etcdserver/server.go#L252)

etcdutl/etcdutl/backup_command.go
ahrtr: > This set of PRs is going towards removing the SetStore method of RaftCluster (along with usage of that store). The backup command is a consumer of SetStore so it needs to be updated.


Agreed. I need to update the `etcdutl backup` command, otherwise we have to keep some source code (e.g. SetStore) which are used by etcdutl only.

> Please let me know if the backup command should be removed. /cc @ahrtr

Let's discuss it separately. I tend to remove it.

server/etcdserver/api/membership/cluster_test.go
ahrtr: what about this comment?

geetasg: updated as per review feedback.

ahrtr (approved):
LGTM

thx @geetasg

geetasg:
can this PR be merged if there is no additional review feedback? I would like to submit next PR which will consume changes from this one.

ahrtr:
ping @ptabor @serathius @chaochn47 @fuweid

fuweid (approved):
LGTM

ahrtr:
cc @mitake and @wenjiaswe to take a look as well.

ahrtr:
It's unfortunate that there is no active review on a P0 item for 3.6.

If there is no any additional review feedback by the end of this week. I will merge this PR by then given it's a safe change.

wenjiaswe:
> It's unfortunate that there is no active review on a P0 item for 3.6.

Looks like there has been active review. 

Overall lgtm. @geetasg Thanks for working on this. 

I'll defer to @ahrtr and @serathius to decide on merging.

ahrtr:
thx @wenjiaswe for the review.

PR #16136, opened 2023-06-24
Robustness watch validation

serathius:


serathius:
cc @ahrtr @jmhbnz @chaochn47 @fuweid

tests/robustness/validate/watch.go
```diff
@@ -107,14 +110,61 @@ func validateAtomic(t *testing.T, report traffic.ClientReport) {
...
+		revision := firstRevision(op)
+		for index < len(events) && events[index].Revision < revision {
+			index++
+		}
+		if index == len(events) {
+			continue
+		}
+		for _, resp := range op.Responses {
+			for _, event := range resp.Events {
+				if events[index].Match(op.Request) && events[index] != event {
```
fuweid, line 125: Do we need to check the index's boundaries ?

fuweid (approved):
LGTM

tests/robustness/validate/watch.go
serathius: Technically no, events in report should be a subset of events. This is only test code, so even if someone breaks that assumption we should still be ok with panic.

PR #16142, opened 2023-06-26
tests/robustness: Test validate

serathius:


jmhbnz:
Discussed duiring sig-etcd triage meeting. @serathius can we close this now?

serathius:
I can close the PR as I don't have enough time to finish it, still the would be a nice improvement to robustness tests.

Robustness tests take a real etcd cluster run some traffic against it. They validate the operation history via a model and additional set of validators. Historically we made mistakes and introduced a bugs into the model or the additional validators. Such bugs are hard to debug because first you need to eliminate that this is an etcd bug.

This PR proposes to also add a "unit" test that would validate model against the validators (without real etcd) that should help us catch bugs earlier.

serathius:
Filled https://github.com/etcd-io/etcd/issues/17203

PR #16151, opened 2023-06-28
Revert backport of arm64 scheduled workflows

jmhbnz:
This reverts #16119.

Reverting changes as scheduled github actions apparently only work if they exist in default branch, refer: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule 

> Scheduled workflows run on the latest commit on the default or base branch. The shortest interval you can run scheduled workflows is once every 5 minutes.

So we need to take an alternative approach of having the workflows in `main` but adding parameter to checkout action so it checks out the `release-3.5` branch.


serathius (approved):

ahrtr (approved):
Thanks @jmhbnz

Issue #15832, opened 2023-05-05
Run arm64 integration and e2e workflows against a supported release branch


jmhbnz:
### What would you like to be added?

There have been recent efforts to resolve issues with arm64 tests and improve overall support for arm64, refer:
- https://github.com/etcd-io/etcd/pull/15829
- https://github.com/etcd-io/etcd/pull/15233
- https://github.com/etcd-io/etcd/pull/15230

These tests have been running more reliably now:
- https://github.com/etcd-io/etcd/actions/workflows/tests-arm64.yaml
- https://github.com/etcd-io/etcd/actions/workflows/e2e-arm64.yaml

With that in mind, should we run these workflows against `release-3.5` to get them running against a stable release branch?

cc @ahrtr, @serathius, @geetasg, @chaochn47 

### Why is this needed?

There have been several discussions and questions lately on improving tier of support for arm64. I believe this would be required step to progressing that?

jmhbnz:
I think https://github.com/etcd-io/etcd/issues/15647 is a blocking issue for this. Once that has been closed off I think this backport could proceed.

natusameer:
Hello @jmhbnz , I can help you with backporting the tests to release-3.5

jmhbnz:
Hey @natusameer - Thanks for volunteering to help with this!  As discussed, please start with a pull request for `release-3.5` initially.

Feel free to comment here or ping me on kubernetes slack if you have any questions along the way 🙏🏻

jmhbnz:
A backport to `release-3.5` has been raised here: https://github.com/etcd-io/etcd/pull/16119 

Once merged we will monitor scheduled runs to confirm tests passing as expected.

jmhbnz:
The backport needs to be reverted, explanation is in this pr: https://github.com/etcd-io/etcd/pull/16151.   Basically scheduled github actions only work if they exist in `default` branch.

I've raised https://github.com/etcd-io/etcd/pull/16151 to revert.

I will raise the pr to re-introduce the scheduled workflows under `main` with `release-3.5` as the ref arg to the checkout action.

PR #16152, opened 2023-06-28
Templated arm64 integration and e2e workflows for main and release-3.5

jmhbnz:
Follow-on from https://github.com/etcd-io/etcd/pull/16151 to setup scheduled workflows for `arm64` nightly integration and e2e suites against the `release-3.5` branch.

Arm64 integration and e2e suites now adopt a shared re-usable workflow template which is slightly customised for `main` or `release-3.5`.

Fixes https://github.com/etcd-io/etcd/issues/15832

### DIFF
--- a/.github/workflows/e2e-arm64-nightly.yaml
+++ b/.github/workflows/e2e-arm64-nightly.yaml
@@ -1,0 +1,19 @@
+---
+name: E2E Arm64 Nightly
+permissions: read-all
+on:
+  # schedules always run against the main branch, hence we have to create separate jobs
+  # with individual checkout actions for each of the active release branches
+  schedule:
+    - cron: '30 1 * * *' # runs daily at 1:30 am.
+jobs:
+  main-arm64:
+    uses: ./.github/workflows/e2e-arm64-template.yaml
+    with:
+      etcdBranch: main
+      e2eTestCmd: make test-e2e-release
+  release-35-arm64:
+    uses: ./.github/workflows/e2e-arm64-template.yaml
+    with:
+      etcdBranch: release-3.5
+      e2eTestCmd: PASSES='build e2e' COVER='false' ./test.sh
--- a/.github/workflows/e2e-arm64-template.yaml
+++ b/.github/workflows/e2e-arm64-template.yaml
@@ -1,8 +1,14 @@
 ---
-name: E2E-arm64
+name: Reusable Arm64 E2E Workflow
 on:
-  schedule:
-    - cron: '0 1 * * *' # runs daily at 1am.
+  workflow_call:
+    inputs:
+      etcdBranch:
+        required: true
+        type: string
+      e2eTestCmd:
+        required: false
+        type: string
 permissions: read-all
 jobs:
   test:
@@ -20,6 +26,8 @@ jobs:
           - linux-arm64-e2e
     steps:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: "${{ inputs.etcdBranch }}"
       # https://github.com/actions/checkout/issues/1169
       - run: git config --system --add safe.directory '*'
       - id: goversion
@@ -36,7 +44,7 @@ jobs:
           echo "${TARGET}"
           case "${TARGET}" in
             linux-arm64-e2e)
-              GOOS=linux GOARCH=arm64 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e-release
+              GOOS=linux GOARCH=arm64 CPU=4 EXPECT_DEBUG=true RACE=true ${{ inputs.e2eTestCmd }}
               ;;
             *)
               echo "Failed to find target"
--- a/.github/workflows/tests-arm64-nightly.yaml
+++ b/.github/workflows/tests-arm64-nightly.yaml
@@ -1,0 +1,22 @@
+---
+name: Integration Arm64 Nightly
+permissions: read-all
+on:
+  # schedules always run against the main branch, hence we have to create separate jobs
+  # with individual checkout actions for each of the active release branches
+  schedule:
+    - cron: '30 2 * * *' # runs daily at 2:30 am.
+jobs:
+  main-arm64:
+    uses: ./.github/workflows/tests-arm64-template.yaml
+    with:
+      etcdBranch: main
+      integrationTestCmd: make test-integration
+      unitTestCmd: GO_TEST_FLAGS='-p=2' make test-unit
+  release-35-arm64:
+    uses: ./.github/workflows/tests-arm64-template.yaml
+    with:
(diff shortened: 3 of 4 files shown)


jmhbnz:
Note for reviewers: Workflows have been tested `on: pull_request` and were green for both `main` and `release-3.5`:
- e2e: https://github.com/etcd-io/etcd/actions/runs/5409330146
- integration: https://github.com/etcd-io/etcd/actions/runs/5409330130

.github/workflows/tests-arm64-nightly.yaml
```diff
@@ -0,0 +1,22 @@
...
+    with:
+      etcdBranch: main
+      integrationTestCmd: make test-integration
+      unitTestCmd: GO_TEST_FLAGS='-p=2' make test-unit
+  release-35-arm64:
+    uses: ./.github/workflows/tests-arm64-template.yaml
+    with:
+      etcdBranch: release-3.5
+      integrationTestCmd: PASSES='integration' RACE='false' ./test.sh
+      unitTestCmd: PASSES='unit' CPU='4' ./test.sh -p=2
```
serathius, line 21: Could we maybe unify the commands to make it clearer?

.github/workflows/e2e-arm64-nightly.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
serathius, line 1: With addition of ARM nightly per branch I think we are getting the limit of clarity of `./github/workflows` directory.
Maybe we could consider creating subdirs and grouping workflows by purpose:
* `./github/workflows/periodics`
* `./github/workflows/presubmits`

Leaving the a idea for a followup.

.github/workflows/tests-arm64-nightly.yaml
jmhbnz: Can you please clarify.  Do you mean backporting the makefile changes so both `release-3.5` and `main` can use same `make` command?  Currently `release-3.5` seems to need `./test.sh`.

.github/workflows/tests-arm64-template.yaml
```diff
@@ -40,19 +56,19 @@ jobs:
           export JUNIT_REPORT_DIR=$(realpath ${TARGET})
           case "${TARGET}" in
             linux-arm64-integration-1-cpu)
-              make gofail-enable
-              GOOS=linux GOARCH=arm64 CPU=1 make test-integration
+              if [ "${{ inputs.gofailMake }}" == "yes" ]; then make gofail-enable; fi
```
serathius, line 59: I'm worried that such forced compatibility will make it harder to add/update tests on main branch and will not help at all with preventing a breaking change. 

I think I would just to prefer to keep the definition of  periodic and pull request workflows separate.

jmhbnz: These templates are only used for periodic.  We don't run arm64 workflows on pull request yet.

I do share your concern around this conditional hack though.

serathius:
tldr; we need to update the release process to incorporate those workflow failures as a signal.

I think we are stomping into new ground. Periodic workflows themselves are not enough to guarantee that next etcd release will work on ARM. They will run, fail but not result in any action if they fail. We need to make them somehow actionable.

To paint the picture. Example of what I mean is process that K8s has. First a dashboard that gives clear view of all periodic tests for all supported versions. Dashboard for v1.24 testing I own https://k8s-testgrid.appspot.com/sig-instrumentation-tests#kind-json-logging-1.24
If there is a repeated test failure, an owner of such test scenario will be notified by email. Independently the K8s release team is monitoring same test failures on their dashboard https://k8s-testgrid.appspot.com/sig-release-1.27-blocking#gce-cos-k8sstable1-default. If a particular test failure blocks a release, they will ping the test owner to fix it.
Before making a release they check that all runs are green.

I think to make the periodic workflows failures be a signal for release we need to adapt to a process similar to this. Of course we will not try to rollout TestGrid ourselves, but as etcd joins Kubernetes it would be nice to have some integration. We already have a release process and release owners defined, we just need to make results from those periodics a clear signal for our releases.

jmhbnz:
> I think to make the periodic workflows failures be a signal for release we need to adapt to a process similar to this. Of course we will not try to rollout TestGrid ourselves, but as etcd joins Kubernetes it would be nice to have some integration. We already have a release process and release owners defined, we just need to make results from those periodics a clear signal for our releases.

Completely agree.  Additionally I am intending to scale up our runners in future to have multiple on each box (we have tons of compute idle on each box) so that we can actually run some of these workflows on each pr also.

.github/workflows/tests-arm64-nightly.yaml
serathius: Yes, I mean backport the makefile changes to keep a single interface for all branches. It should the forkflows easier to maintain and simplify development of backports. cc @ahrtr

jmhbnz: I'm happy to create an issue and work on it if we have consensus from others.  I'm not sure I would want it to be a blocking issue for this pr though. Keen to see what overall consensus is 🙂

ahrtr: > Yes, I mean backport the makefile changes to keep a single interface for all branches.


Makes sense to me.

.github/workflows/e2e-arm64-template.yaml
```diff
@@ -1,8 +1,14 @@
 ---
-name: E2E-arm64
+name: Reusable Arm64 E2E Workflow
```
ahrtr, line 2: Probably I am missing something, the template is only used by e2e-arm64-nightly for now. In other words, the template has only one user, then why do you need to create a template? If the template can be reused by both ARM e2e test and NON-ARM e2e test, then it makes sense.

jmhbnz: We use a re-usable workflow so we can run the same template for `main` and `release-3.5` with different parameters. Similar to how robustness workflows run across different branches using the same workflow template.

ahrtr: thx for the clarification. It seems it would be better if there is a document for all the workflow tests.

.github/workflows/tests-arm64-nightly.yaml
jmhbnz: Created https://github.com/etcd-io/etcd/issues/16176 under our overall ci improvements plan to track.  I am keen to work on this once we have tier 1 arm64 announced 🙏🏻

jmhbnz:
Hey @serathius, @ahrtr - Can you please take another look at this? I'm keen to see how these tests will behave against 3.5.

ahrtr (approved):
LGTM

Thanks @jmhbnz 

I think we can discuss the adapting the similar K8s process in a separate session. Leave to @serathius to approve and merge this PR.

serathius (approved):

PR #16161, opened 2023-07-02
tests/robustness: Prevent to many concurrent non-unique writes which are causing linearization to timeout

serathius:
cc @ahrtr @ptabor @jmhbnz @chaochn47 @fuweid

### DIFF
--- a/tests/robustness/traffic/kubernetes.go
+++ b/tests/robustness/traffic/kubernetes.go
@@ -32,10 +32,11 @@ import (
 
 var (
 	KubernetesTraffic = Config{
-		Name:        "Kubernetes",
-		minimalQPS:  200,
-		maximalQPS:  1000,
-		clientCount: 12,
+		Name:                           "Kubernetes",
+		minimalQPS:                     200,
+		maximalQPS:                     1000,
+		clientCount:                    12,
+		maxNonUniqueRequestConcurrency: 3,
 		Traffic: kubernetesTraffic{
 			averageKeyCount: 10,
 			resource:        "pods",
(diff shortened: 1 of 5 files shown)


tests/robustness/traffic/limiter.go
```diff
@@ -0,0 +1,33 @@
+package traffic
```
jmhbnz, line 1: Just missing license header here.

serathius: Done, and added a test.

serathius:
ping  @ahrtr @ptabor @jmhbnz @chaochn47 @fuweid

tests/robustness/traffic/kubernetes.go
```diff
@@ -151,13 +152,19 @@ func (t kubernetesTraffic) Write(ctx context.Context, kc *kubernetesClient, ids
...
-			op := pickRandom(t.writeChoices)
+			choices := t.writeChoices
+			if !nonUniqueWriteLimiter.Take() {
+				choices = filterOutNonUniqueKuberntesWrites(t.writeChoices)
+			}
+			op := pickRandom(choices)
 			switch op {
 			case KubernetesDelete:
 				_, err = kc.OptimisticDelete(writeCtx, key, rev)
+				nonUniqueWriteLimiter.Return()
```
fuweid, line 167: Not sure I am understand the code correctly: If the number of client (8) is larger than non-unique write (3), if 5 clients don't get true from `Take()` at Line 160 at the same time, is it safe to call the `Return` here?

serathius: If they got `Take() = false`, then `op` should never be equal to `KubernetesDelete`.

fuweid (approved):
LGTM

PR #16166, opened 2023-07-03
etcdutl: update description for `--mark-compacted and` and `--bump-revision` flags in snapshot restore command

ahrtr:
It's really weird to have circular description between `--bump-revision` and `--mark-compacted`,
- description `required if --mark-compacted` for flag `--bump-revision`
- description `required if --bump-revision > 0` for flag `--mark-compacted`

Actually `--mark-compacted` is determined entirely by `--bump-revision`,
- It (--mark-compacted) isn't allowed if users do not bump revision
- It (--mark-compacted) must be specified if users bump the revision

So we only need to add the description(restrictions) to flag `--mark-compacted`. See also https://github.com/etcd-io/etcd/pull/16029#issuecomment-1612703238


serathius (approved):

jmhbnz (approved):
LGTM

PR #16137, opened 2023-06-25
*: should return exitCode even if cmd isn't nil

fuweid:
For the pkg/expect package, if the process has been stopped but there is no `Close()` call, the `ExitCode()` won't return exit code correctly. The `ExitCode()` should check `exitErr` and return exit code if cmd isn't nil.

And introduces `exitCode` to return correct exit code based on the process is signaled or exited.

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

-----

Context: https://github.com/etcd-io/etcd/pull/16094#discussion_r1233134482

cc @serathius @tjungblu 

fixes: #16134

PR #16173, opened 2023-07-04
pkg/expect: fix data race

fuweid:
Let's say there is command which outputs one line and exit with error.

There are three goroutines to acquire the lock:

1. ep.read()
2. ep.waitSaveExitErr()
3. ep.Expect()

When ep.read goroutine reads the log but it doesn't acquire the lock in time, the ep.waitSaveExitErr acquires the lock and updates the `exitErr`. And then ep.Expect acquires lock but it doesn't see any log yet and then returns err.

It's hard to reproduce it in local. Add the extra sleep can reproduce it.

```diff
diff --git a/pkg/expect/expect.go b/pkg/expect/expect.go
index a512a3ce4..602bea73f 100644
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -128,6 +128,7 @@ func (ep *ExpectProcess) tryReadNextLine(r *bufio.Reader) error {
        printDebugLines := os.Getenv("EXPECT_DEBUG") != ""
        l, err := r.ReadString('\n')

+       time.Sleep(10 * time.Millisecond)
        ep.mu.Lock()
        defer ep.mu.Unlock()
```

See it once in Github Action [[1]]. In order to fix it, the patch introduces `readCloseCh` to wait for ep.read to get all the data and retry it.

[1]: https://github.com/etcd-io/etcd/pull/16137#issuecomment-1605838518

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -44,6 +44,8 @@ type ExpectProcess struct {
 	fpty *os.File
 	wg   sync.WaitGroup
 
+	readCloseCh chan struct{} // close it if async read goroutine exits
+
 	mu       sync.Mutex // protects lines, count, cur, exitErr and exitCode
 	lines    []string
 	count    int   // increment whenever new line gets added
@@ -67,6 +69,7 @@ func NewExpectWithEnv(name string, args []string, env []string, serverProcessCon
 			args: args,
 			env:  env,
 		},
+		readCloseCh: make(chan struct{}),
 	}
 	ep.cmd = commandFromConfig(ep.cfg)
 
@@ -100,7 +103,10 @@ func (ep *ExpectProcess) Pid() int {
 }
 
 func (ep *ExpectProcess) read() {
-	defer ep.wg.Done()
+	defer func() {
+		ep.wg.Done()
+		close(ep.readCloseCh)
+	}()
 	defer func(fpty *os.File) {
 		err := fpty.Close()
 		if err != nil {
@@ -187,8 +193,24 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 		}
 	}
 
+	select {
+	// NOTE: we wait readCloseCh for ep.read() to complete draining the log before acquring the lock.
+	case <-ep.readCloseCh:
+	case <-ctx.Done():
+		return "", fmt.Errorf("failed to find match string: %w", ctx.Err())
+	}
+
 	ep.mu.Lock()
 	defer ep.mu.Unlock()
+
+	// retry it since we get all the log data
+	for i < len(ep.lines) {
+		line := ep.lines[i]
+		i++
+		if f(line) {
+			return line, nil
+		}
+	}
 
 	lastLinesIndex := len(ep.lines) - DEBUG_LINES_TAIL
 	if lastLinesIndex < 0 {
--- a/pkg/expect/expect_test.go
+++ b/pkg/expect/expect_test.go
@@ -210,3 +210,11 @@ func TestExitCodeAfterKill(t *testing.T) {
 	assert.Equal(t, 137, code)
 	assert.NoError(t, err)
 }
+
+func TestExpectForFailFastCommand(t *testing.T) {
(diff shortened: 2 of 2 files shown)


tjungblu:
wow, great catch. I wonder if we can simplify the whole expect a little by serialization? 
Let's say we would always wait for the process to exit first, then drain its pty output and then match for the function on that output.

I'm not entirely sure whether we have many cases that spawn a longer running task and require an "early" expect result for test runtime reasons. TestGrpcProxyAutoSync seems to be one such candidate.

fuweid:
> Let's say we would always wait for the process to exit first, then drain its pty output and then match for the function on that output.

I was thinking that case. But there are some cases to expect process is ready, like `EtcdServerProcess`. So we still need to check the ongoing log data.

tjungblu:
indeed, that's also what I found browsing for it. Alternatively we can also check the readiness probe instead? Anyway, I think that's a bigger change overall.

fuweid:
> Alternatively we can also check the readiness probe instead?  Anyway, I think that's a bigger change overall.

Agree. It's reasonable to use expect pkg for short-live process only. It needs ETCD server to expose more metrics in API to replace checking the logging. Maybe we can track this in issue for enhancement.

pkg/expect/expect.go
```diff
@@ -187,9 +193,26 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 		}
 	}
 
+	select {
+	// NOTE: The ep.read() goroutine needs the mutex so we have to drain the
+	// log without lock.
```
chaochn47, line 198: > so we have to drain the log without lock. 

Could you describe where we drain the log without lock? I thought ep.read() always drains the log with the mutex.

fuweid: Sorry for the unclear comment. 

https://github.com/etcd-io/etcd/blob/4364e7581b5f5fd664916fb6cb80e0a396b7e57d/pkg/expect/expect.go#L102-L138

If `ep.read()` just reads last line from pty, it will try to get lock and then append the log into `ep.lines`. So, if the `ep.Expect`  waits the `ep.readCloseCh` channel close with the lock, the `ep.read()` doesn't have a chance to get the lock and then the `ep.readCloseCh` won't be closed until `ep.Expect` releases the lock. It's deadlock.

chaochn47: Thanks for the clarification! 

```
// NOTE: The ep.read() goroutine needs the mutex so we have to drain the
// log without lock.
```

I guess `ep.ExpectFunc` does not drain the log which causes my confusion. 

How about updating to?
```
// NOTE: we wait readCloseCh for ep.read() to complete draining the log before acquring the lock. 
```

pkg/expect/expect.go
```diff
@@ -187,9 +193,26 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
...
+
 	ep.mu.Lock()
 	defer ep.mu.Unlock()
 
+	// retry it since we get all the log data
+	for i < len(ep.lines) {
+		line := ep.lines[i]
+		i++
+		if f(line) {
+			return line, nil
```
chaochn47, line 212: I don't think any consumer of `ExpectFunc` relies on the exit error, so return nil should be fine.

fuweid:
Thanks @chaochn47 ! updated

fuweid:
ping @ahrtr @serathius

pkg/expect/expect.go
```diff
@@ -69,6 +71,7 @@ func NewExpectWithEnv(name string, args []string, env []string, serverProcessCon
 		},
 	}
 	ep.cmd = commandFromConfig(ep.cfg)
+	ep.readCloseCh = make(chan struct{})
```
ahrtr, line 74: Minor comment, we can just put this line into the struct initialization,

```
	ep = &ExpectProcess{
		cfg: expectConfig{
			name: serverProcessConfigName,
			cmd:  name,
			args: args,
			env:  env,
		},
		readCloseCh: make(chan struct{}),
	}
```

ahrtr (approved):
LGTM with a very minor comment. 

Thanks @fuweid

PR #16189, opened 2023-07-06
Update community meeting frequency to fortnightly

jmhbnz:
As discussed at our recent community meeting we have run a brief survey to get consensus on how often the community would like to meet.

The [survey results](https://github.com/etcd-io/website/issues/703) are now in and the community would prefer to move to fortnightly meetings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ Now it's time to dig into the full etcd API and other guides.
 
 ### Community meetings
 
-etcd contributors and maintainers have monthly (every four weeks) meetings at 11:00 AM (USA Pacific) on Thursday.
+etcd contributors and maintainers meet every two weeks at 11:00 AM (USA Pacific) on Thursday.
 
 An initial agenda will be posted to the [shared Google docs][shared-meeting-notes] a day before each meeting, and everyone is welcome to suggest additional topics or other agendas.
 


README.md
```diff
@@ -140,7 +140,7 @@ Now it's time to dig into the full etcd API and other guides.
 
 ### Community meetings
 
-etcd contributors and maintainers have monthly (every four weeks) meetings at 11:00 AM (USA Pacific) on Thursday.
+etcd contributors and maintainers meet every two weeks at 11:00 AM (USA Pacific) on Thursday.
```
ahrtr, line 143: Can we have two times; One is the existing one, and the other one is an APAC or EMEA friendly time?
- 11:00 AM USA Pacific (keep the existing one as it's)
- Can we raise a discussion topic in etcd-dev and the etcd slack channel?

jmhbnz: If we want to do two different times for the meetings it might be difficult for @serathius who does the organising and video uploads to make it to both.  We would need to make sure someone could run each one, it makes things a bit more complicated maybe.

In terms of the feedback from the community I think the main thing was the community wanted to meet more often to allow more time to discuss agenda items. If we have split meeting times then we won't actually be achieving the community feedback of meeting more regularly as there will still only be one meeting per month for each timezone and it is unlikely people will be able to make both timezones?  We would probably need to make the meetings longer to achieve the goal in that case which isn't what the community wanted in the [survey results](https://github.com/etcd-io/website/issues/703#issuecomment-1621562421).

We would also need to think about how duplication of discussions between regions is handled. If something is discussed at one meeting but only few people can make it, does it have to be discussed all over again at the other meeting and how does that impact decision making?

I'm not opposed btw, just want to make sure we think it through and deliver on the feedback the community gave us in the survey.

ahrtr: > feedback the community gave us in the survey

What's the feedback on the timezone in the survey?

jmhbnz: Here are the 11 exact responses we received, some people did not enter anything for this question:

```text
PST/PDT
PST
India
GMT+2
PST
CEST
China Beijing
PDT
ET
NZST
EST
```

So the most common is PST/PDT with a scattering of others that make it impossible to have a time that works for everyone. The time we have currently can work for quite a few of us though:

image

ahrtr: thx for the info. It isn't a big problem to me, but let's see other's feedback.

spzala: I think good to merge this PR and continue checking if separate meeting is needed for the APAC / EMEA. If so, it will be good if we can identify someone from that region who can lead the call. Thanks!

ahrtr (approved):

README.md
wenjiaswe: Agree to submit the PR.

Also agree that we should have both AMER and APAC/EMEA friendly meetings. I can to run either or both meetings.

fuweid (approved):
LGTM

README.md
wenjiaswe: Was looking at the time table again, how about we have Wednesday 11pm PST once/4 weeks and Thursday 11am PST once/4 weeks?

image

serathius (approved):

README.md
serathius: I think it makes no sense to adjust the meeting based on guesses about user preference. Could we send a separate survey that will target users that cannot attend the meeting due to timezone. It should give us idea on how many they are and how to best accommodate them.

serathius:
Approving based on fact that we have maintainer quorum. We can iterate on the timezone.

PR #16200, opened 2023-07-09
clientv3: create keepAliveCtxCloser goroutine only if ctx can be canc…

kensou97:
We always create a `keepAliveCtxCloser` goroutine to eagerly update `keepAlive` when ctx is done.
But if ctx can never be canceled, there is no need to start a redundant goroutine.

### DIFF
--- a/client/v3/lease.go
+++ b/client/v3/lease.go
@@ -294,7 +294,9 @@ func (l *lessor) KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAl
 	}
 	l.mu.Unlock()
 
-	go l.keepAliveCtxCloser(ctx, id, ka.donec)
+	if ctx.Done() != nil {
+		go l.keepAliveCtxCloser(ctx, id, ka.donec)
+	}
 	l.firstKeepAliveOnce.Do(func() {
 		go l.recvKeepAliveLoop()
 		go l.deadlineLoop()


client/v3/lease.go
```diff
@@ -294,7 +294,9 @@ func (l *lessor) KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAl
 	}
 	l.mu.Unlock()
 
-	go l.keepAliveCtxCloser(ctx, id, ka.donec)
+	if ctx.Done() != nil {
+		go l.keepAliveCtxCloser(ctx, id, ka.donec)
```
fuweid, line 298: I think there is not just `ctx.Done()`, but also `ka.donec`. The `l.keepAliveCtxCloser` is also used to watch the event from `ka.donec`.

kensou97: > I think there is not just `ctx.Done()`, but also `ka.donec`. The `l.keepAliveCtxCloser` is also used to watch the event from `ka.donec`.

I think `keepAliveCtxCloser` is only used to watch user's active cancellation. It does watch `ka.donec`, but that is mainly used for exiting the goroutine to avoid leakage if `keepAlive` is closed for other reasons.

fuweid: Thanks for the comment. Change looks good to me.

fuweid (approved):
LGTM

ahrtr (approved):
LGTM

thx @kensou97 

Please also backport this PR to 3.5 and 3.4 if needed.

kensou97:
> LGTM
> 
> thx @kensou97
> 
> Please also backport this PR to 3.5 and 3.4 if needed.

ok

jmhbnz (approved):

chaochn47 (approved):
LGTM

PR #13334, opened 2021-09-09
fix IsOptsWithFromKey

LeoYang90:
fix https://github.com/etcd-io/etcd/issues/13332

PR #16224, opened 2023-07-11
expose op.isOptsWithFromKey and op.isOptsWithPrefix

### LINKED ISSUE #16217: Op.isOptsWithFromKey & Op.isOptsWithPrefix accessor

It would be nice if the access function of `Op.isOptsWithPrefix` and `Op.isOptsWithFromKey` were provided. Only the accessor to this field was missing, is there a history related to it?

```go
op := clientv3.OpDelete("mykey", clientv3.WithPrefix())
if op.IsOptsWithPrefix() {
  // ...
}
```

### Why is this needed?

For making detail logs in application.

CaojiamingAlan:
Fix #16217

### DIFF
--- a/client/v3/op.go
+++ b/client/v3/op.go
@@ -124,6 +124,10 @@ func (op Op) IsKeysOnly() bool { return op.keysOnly }
 // IsCountOnly returns whether countOnly is set.
 func (op Op) IsCountOnly() bool { return op.countOnly }
 
+func (op Op) IsOptsWithFromKey() bool { return op.isOptsWithFromKey }
+
+func (op Op) IsOptsWithPrefix() bool { return op.isOptsWithPrefix }
+
 // MinModRev returns the operation's minimum modify revision.
 func (op Op) MinModRev() int64 { return op.minModRev }
 
--- a/client/v3/op_test.go
+++ b/client/v3/op_test.go
@@ -78,24 +78,28 @@ func TestIsSortOptionValid(t *testing.T) {
 
 func TestIsOptsWithPrefix(t *testing.T) {
 	optswithprefix := []OpOption{WithPrefix()}
-	if !IsOptsWithPrefix(optswithprefix) {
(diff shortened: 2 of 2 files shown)


ahrtr:
They are already supported in https://github.com/etcd-io/etcd/pull/13334.

CaojiamingAlan:
> They are already supported in #13334.

The exported [methods](https://github.com/etcd-io/etcd/blob/882edb3d63cca13d4b2541e4e39e29c53f3814a6/client/v3/op.go#L572C1-L589) in #13334 takes a list of `OpOption` as parameter, which means the user needs to keep the options of the op if he or she wants to check the fields `isOptsWithFromKey` and `isOptsWithPrefix` later. what I am trying to do here is to expose those fields directly.

ahrtr:
> if he or she wants to check the fields `isOptsWithFromKey` and `isOptsWithPrefix` later

Yes, I understand that, but actually I don't see a real such use case.  After the `WithFromKey` or `WithPrefix` is evaluated, users only care about the field [end](https://github.com/etcd-io/etcd/blob/882edb3d63cca13d4b2541e4e39e29c53f3814a6/client/v3/op.go#L35). 

Of course, adding two methods (what this PR does) is harmless, but also unnecessary. I am neutral to this.

jmhbnz (approved):
This gets a tentative approval from me.  At the code level I see no issue, and we have had #16217 opened by a user asking for this for the purpose of making detailed logs in an application.

It still seems like a nice to have more than anything so I defer to maintainers to make final decision.   @changhoi please feel free to add any additional context here or in the issue in support of the pr.

changhoi:
As described in [PR](https://github.com/etcd-io/etcd/issues/16217), if the feature is added, our team will add more details to the log depending on whether Operation has a Prefix option.

Also, since there are other flag accessors, it seems natural to add them if there are no problems :)

ahrtr (approved):

Issue #16176, opened 2023-07-05
Streamline test commands between branches


jmhbnz:
### What would you like to be added?

Sub issue under https://github.com/etcd-io/etcd/issues/15951.

I would like to be able to run tests in github actions workflows across etcd branches using the same command.

Currently every branch is different, refer to end test suite as the example:

| Branch      | Command                                                                                                |
|-------------|--------------------------------------------------------------------------------------------------------|
| main        | `make test-e2e-release`                      |
| release-3.5 | `PASSES='build release e2e' COVER='false' ./test.sh` |
| release-3.4 | `PASSES='build release e2e' ./test`                                                       |

I would like to use this issue to explore if a minimal backport from `main` can be done to allow both `release-3.5` and perhaps also `release-3.4` to also use the same commands to run test suites.

### Why is this needed?

Currently etcd ci workflow test commands are different between branches, this creates confusion, cognitive burden and makes our github actions workflow templates brittle and over complicated.

serathius:
Note: release-3.5 should also run `PASSES='build release e2e' ./test`

jmhbnz:
> Note: release-3.5 should also run `PASSES='build release e2e' ./test` like release-3.4. Don't know why it's different.

Ahh I copied the command from the `arm64` files in my open pr rather than the release-3.5 existing branch workflow file.  I need to look at why none of our `arm64` workflows have it included.

PR #16230, opened 2023-07-13
Ensure release is run for arm64 e2e nightly tests

jmhbnz:
Recently in https://github.com/etcd-io/etcd/pull/16152 we extended our `arm64` nightly test coverage to include `release-3.5` as  a requirement for being able to announce tier1 support for `arm64`.

When these were merged the `PASSES` for the test command was `PASSES='build e2e'`.   To truly have alignment between what we run for `arm64` vs `amd64` we need to update that to `PASSES='build release e2e'`.

Now that the first iteration of the tests have been merged and are passing successfully overnight I would now like to fix the alignment so both architectures run the exact same command.

This is a pre step before I can start work on https://github.com/etcd-io/etcd/issues/16176.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -559,8 +559,30 @@ function dep_pass {
 
 function release_pass {
   rm -f ./bin/etcd-last-release
-  # to grab latest patch release; bump this up for every minor release
-  UPGRADE_VER=$(git tag -l --sort=-version:refname "v3.5.*" | head -1 | cut -d- -f1)
+
+  # Work out the previous release based on the version reported by etcd binary
+  binary_version=$(./bin/etcd --version | grep --only-matching --perl-regexp '(?<=etcd Version: )\d+\.\d+')
+  binary_major=$(echo "${binary_version}" | cut -d '.' -f 1)
+  binary_minor=$(echo "${binary_version}" | cut -d '.' -f 2)
+  previous_minor=$((binary_minor - 1))
+
+  # Handle the edge case where we go to a new major version
+  # When this happens we obtain latest minor release of previous major
+  if [ "${binary_minor}" -eq 0 ]; then
+    binary_major=$((binary_major - 1))
+    previous_minor=$(git ls-remote --tags https://github.com/etcd-io/etcd.git \
+    | grep --only-matching --perl-regexp "(?<=v)${binary_major}.\d.[\d]+?(?=[\^])" \
+    | sort --numeric-sort --key 1.3 | tail -1 | cut -d '.' -f 2)
+  fi
+  
+  # This gets a list of all remote tags for the release branch in regex
+  # Sort key is used to sort numerically by patch version
+  # Latest version is then stored for use below
+  UPGRADE_VER=$(git ls-remote --tags https://github.com/etcd-io/etcd.git \
+    | grep --only-matching --perl-regexp "(?<=v)${binary_major}.${previous_minor}.[\d]+?(?=[\^])" \
+    | sort --numeric-sort --key 1.5 | tail -1 | sed 's/^/v/')
+  log_callout "Found latest release: ${UPGRADE_VER}."
+
   if [ -n "${MANUAL_VER:-}" ]; then
     # in case, we need to test against different version
     UPGRADE_VER=$MANUAL_VER
--- a/.github/workflows/e2e-arm64-nightly.yaml
+++ b/.github/workflows/e2e-arm64-nightly.yaml
@@ -16,4 +16,4 @@ jobs:
     uses: ./.github/workflows/e2e-arm64-template.yaml
     with:
       etcdBranch: release-3.5
-      e2eTestCmd: PASSES='build e2e' COVER='false' ./test.sh
+      e2eTestCmd: PASSES='build release e2e' COVER='false' ./test.sh


jmhbnz (commented):
Note for reviewers - this small alignment has been tested `on: pull_request` and is passing as expected, refer: https://github.com/etcd-io/etcd/actions/runs/5540459409 

Now switched back to `on: schedule` for merge 🙏🏻

serathius:
Looks like release-35-arm64/test incorrectly fallbacks to downloading v3.3.0. https://screenshot.googleplex.com/Bq6FbvbfkeuvXLx

jmhbnz:
> Looks like release-35-arm64/test incorrectly fallbacks to downloading v3.3.0. https://screenshot.googleplex.com/Bq6FbvbfkeuvXLx
> 
> `release` pass downloads a older etcd binary. Either picked by providing a script, or tries to guess which to download. Some tests like downgrade/upgrade tests, depend on this binary being present and are skipped otherwise.
> 
> If we want to improve periodics with the `release` pass we should think about how to validate that each branch downloaded a proper etcd binary and run the conditional tests.

Great spotting. I was trying to avoid passing the `MANUAL_VER=<version>` parameter and instead rely on https://github.com/etcd-io/etcd/blob/release-3.5/test.sh#L639C1-L639C14.  Reviewing the shell script again I have a hunch what might be going wrong.  When I get back this weekend I will do some more testing and update this pr 🙏🏻

jmhbnz:
Hey @serathius I've resolved the release e2e issue. It wasn't limited to the `arm64` workflows either, the main e2e workflow was also hitting the fallback conditional, it just wasn't as problematic because that one was falling back to `3.5.0`.

So long story short we have this snippet in `test.sh`

```bash
  UPGRADE_VER=$(git tag -l --sort=-version:refname "v3.5.*" | head -1 | cut -d- -f1)
  if [ -n "${MANUAL_VER:-}" ]; then
    # in case, we need to test against different version
    UPGRADE_VER=$MANUAL_VER
  fi
  if [[ -z ${UPGRADE_VER} ]]; then
    UPGRADE_VER="v3.5.0"
    log_warning "fallback to" ${UPGRADE_VER}
  fi
```

Normally this would work great, however by default `actions/checkout` does a shallow clone so running `git tag` doesn't list anything.  We just need to pass the `fetch-depth` parameter to ensure we do a full clone and make tags available.

With this pr it's fixed now for both `arm64` nightly e2e for `main` and `release-3.5` and also standard e2e for `main`.

Note: For `releas-3.5` standard e2e we should probably update https://github.com/etcd-io/etcd/blob/release-3.5/.github/workflows/e2e.yaml#L28 to avoid passing a static `MANUAL_VER` and instead just do full clone so that version will always be dynamic for previous release. If we have agreement on this I will create a good first issue for it.

.github/workflows/e2e-arm64-template.yaml
```diff
@@ -28,6 +28,9 @@ jobs:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
           ref: "${{ inputs.etcdBranch }}"
+          # Workflows that run 'release e2e' tests need to avoid github default shallow clone
+          # Setting fetch-depth 0 clones full project so all tags are present
+          fetch-depth: 0
```
serathius, line 33: Don't like implicit dependency of `PASSES='build release e2e' COVER='false' ./test.sh` on local repo state. I think that `release` script should check the available releases from remote.

jmhbnz: Yeah fair point. I've switched to an approach based on `git ls-remote --tags`.  Please let me know what you think 🙂

jmhbnz: Here is the actions log for e2e showing the new approach in action:https://github.com/etcd-io/etcd/actions/runs/5585011211/jobs/10207184417?pr=16230#step:5:632

scripts/test.sh
```diff
@@ -559,8 +559,15 @@ function dep_pass {
...
 function release_pass {
   rm -f ./bin/etcd-last-release
-  # to grab latest patch release; bump this up for every minor release
-  UPGRADE_VER=$(git tag -l --sort=-version:refname "v3.5.*" | head -1 | cut -d- -f1)
+
+  # This gets a list of all remote tags for the release branch in regex
+  # Sort key is used to sort numerically by patch version
+  # Latest version is then stored for use below
+  UPGRADE_VER=$(git ls-remote --tags https://github.com/etcd-io/etcd.git \
+    | grep --only-matching --perl-regexp '(?<=v)3.5.[\d]+?(?=[\^])' \
```
serathius, line 567: For `main` branch we want to run test against v3.5 minor, but for `release-3.5` we should be picking from v3.4.

jmhbnz: Yeah I believe we just need to backport this change to `release-3.5` and change the regex to 3.4 once this merges.

jmhbnz: `make test-e2e-release` always runs `make build` first so how about something like this one liner, no params required, will always stay up to date based on minor version reported by the etcd binary:

```bash
echo $(( $(./bin/etcd --version | grep --only-matching --perl-regexp '(?<=etcd Version: 3.)\d') - 1 ))
```

serathius: That's a nice solution.

serathius (approved):

scripts/test.sh
jmhbnz: Cool ok - added a handler for the edge case of new major version, think it's all sorted now pending tests green ✔️

ahrtr (approved):
LGTM with a nit, pls feel free to keep it as it's.

thx @jmhbnz

scripts/test.sh
```diff
@@ -559,8 +559,30 @@ function dep_pass {
...
+  binary_minor=$(echo "${binary_version}" | cut -d '.' -f 2)
+  previous_minor=$((binary_minor - 1))
+
+  # Handle the edge case where we go to a new major version
+  # When this happens we obtain latest minor release of previous major
+  if [ "${binary_minor}" -eq 0 ]; then
+    binary_major=$((binary_major - 1))
+    previous_minor=$(git ls-remote --tags https://github.com/etcd-io/etcd.git \
+    | grep --only-matching --perl-regexp "(?<=v)${binary_major}.\d.[\d]+?(?=[\^])" \
+    | sort --numeric-sort --key 1.3 | tail -1 | sed 's/^/v/' | cut -d '.' -f 2)
```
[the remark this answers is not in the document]
fuweid: Yeah. We don't need the `v` in the result. anyway, `vx` is part one and we don't use it.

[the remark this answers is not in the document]
jmhbnz: Eek copy paste. Good spotting, fixed 👍🏻

fuweid (approved):
LGTM

Issue #16150, opened 2023-06-28
UserAdd and RoleAdd bypass --quota-backend-bytes


CaojiamingAlan:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request, support requests should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

When the db size hit the limit of --quota-backend-bytes, users can still add users and roles, causing the  db size continue to increase.

### What did you expect to happen?

user add/role add should also return ErrNoSpace when hitting the limit

### How can we reproduce it (as minimally and precisely as possible)?

1. Start etcd server with a low --quota-backend-bytes, for example 200kB: `etcd --quota-backend-bytes=$((200*1024))`

Check the db size with `etcdctl --write-out=table endpoint status`
The db's initial size is 98kB:
<img width="1008" alt="image" src="https://github.com/etcd-io/etcd/assets/35325198/89cee3a5-36b3-4a1b-9879-4fffdf70acf3">

2. Use scripts to add 1000 users:
```
package main

import (
	"context"
	"fmt"
	clientv3 "go.etcd.io/etcd/client/v3"
	"google.golang.org/grpc"
	"math/rand"
	"time"
)

var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
func RandStringRunes(n int) string {
	b := make([]rune, n)
	for i := range b {
		b[i] = letterRunes[rand.Intn(len(letterRunes))]
	}
	return string(b)
}

func main() {
	fmt.Println("start")
	cli, err := clientv3.New(clientv3.Config{
		Endpoints:   []string{"localhost:2379"},
		DialTimeout: time.Second * 5,
		DialOptions: []grpc.DialOption{grpc.WithBlock()},
	})
	if err != nil {
		fmt.Println(err.Error())
		return
	}
	for i := 0; i < 1000; i++ {
		_, err := cli.UserAdd(context.Background(), RandStringRunes(100), "password")
		if err != nil {
			fmt.Println(err.Error())
			return
		}
[...]

PR #16236, opened 2023-07-13
etcdserver: add kv tests to backend applier

iuriatan:
* add `put` tests on quota applier

Following discussions on  #16036 and #16150, this patch aims to improve appliers test coverage.

### DIFF
--- a/server/etcdserver/apply/apply_kv_test.go
+++ b/server/etcdserver/apply/apply_kv_test.go
@@ -1,0 +1,418 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apply
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+	"golang.org/x/crypto/bcrypt"
+
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/server/v3/auth"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/v3alarm"
+	"go.etcd.io/etcd/server/v3/etcdserver/cindex"
+	"go.etcd.io/etcd/server/v3/etcdserver/errors"
+	"go.etcd.io/etcd/server/v3/lease"
+	serverstorage "go.etcd.io/etcd/server/v3/storage"
+	"go.etcd.io/etcd/server/v3/storage/backend"
+	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
+	"go.etcd.io/etcd/server/v3/storage/schema"
+)
+
+const (
+	putInherentCost = 256
+	key2            = "key2"
+	key3            = "key3"
+	key4            = "key4"
+	value64         = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789?!"
+	value128        = value64 + value64
+)
+
+var (
+	fourKeys = []pb.PutRequest{
+		{Key: []byte(key), Value: []byte("key's value")},
+		{Key: []byte(key2), Value: []byte("key2's value")},
+		{Key: []byte(key3), Value: []byte("key3's value")},
+		{Key: []byte(key4), Value: []byte("key4's value")},
+	}
+	updatedValue  = []byte("up2date value")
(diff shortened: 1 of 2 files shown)


server/etcdserver/apply/apply_kv_test.go
```diff
@@ -0,0 +1,408 @@
...
+		Value: []byte(value128),
+	})
+	require.EqualValues(t, errors.ErrNoSpace, err)
+
+	// Putting 64B should go OK
+	result, _, err := qApp.Put(ctx, nil, &pb.PutRequest{
+		Key:   []byte(key),
+		Value: []byte(value64),
+	})
+	require.EqualValues(t, errors.ErrNoSpace, err)
```
CaojiamingAlan, line 406: This assertion fails the test.

CaojiamingAlan (changes requested):
Please make sure your PR passes all of the tests and checks. Please also run `make fix` to see if it fails any linter.

server/etcdserver/apply/apply_kv_test.go
iuriatan: The quota tests are sure to be reviewed. Beyond some leftovers, trusting exact DB startup sizes will fail in different platforms. I can fix it!

... and I couldn't `make verify` it here due to different linter version/behavior. I can do it now and get Static Analysis on track. 

Thanks a lot! 🙇🏼

iuriatan: Please resolve the conversation if you think this has been addressed 🙇🏼

server/etcdserver/apply/apply_kv_test.go
```diff
@@ -0,0 +1,414 @@
...
+		kv:                           kv,
+		alarmStore:                   alarmStore,
+		authStore:                    authStore,
+		lessor:                       lessor,
+		cluster:                      cluster,
+		raftStatus:                   &fakeRaftStatusGetter{},
+		snapshotServer:               &fakeSnapshotServer{},
+		consistentIndex:              consistentIndex,
+		txnModeWriteWithSharedBuffer: false,
+	}, be
```
fuweid, line 105: suggest to use `t.Cleanup` to tear down backend.

server/etcdserver/apply/apply_kv_test.go
```diff
@@ -0,0 +1,414 @@
...
+		{
+			name:      "Count keys greater then or equal to key2",
+			request:   &pb.RangeRequest{Key: []byte(key2), RangeEnd: zero, CountOnly: true},
+			expectedV: [][]byte{},
+			expectedC: 3,
+		},
+		// TODO: test Serializable
+	}
+
+	for _, tc := range tcs {
```
fuweid, line 263: please use `tc := tc` before `t.Run`, just in case that the previous var has been shadowed by next.

server/etcdserver/apply/apply_kv_test.go
```diff
@@ -0,0 +1,414 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			resp, err := appV3be.Range(ctx, nil, tc.request)
+			//fmt.Printf("*** %v\n", resp)
```
fuweid, line 266: please remove the debug line ~

server/etcdserver/apply/apply_kv_test.go
```diff
@@ -0,0 +1,414 @@
...
+		{Key: []byte(key), Value: updatedValue},  // rev 6
+		{Key: []byte(key), Value: updatedValue2}, // rev 7
+	}, appV3be)
+
+	tcs := []struct {
+		name       string
+		request    *pb.RangeRequest
+		expectedV  [][]byte
+		expectedC  int64
+		customTest func(r *pb.RangeResponse)
```
fuweid, line 213: Since you uses sub-test-runner, you should add `t *testing.T` to `cutomTest`.

server/etcdserver/apply/apply_kv_test.go
iuriatan: First of all, thank you for the nice catch!
Can you please increment the discussion further just for me to get it right?

🤯 ... As the `for` block declares an anonymous function capturing `tc` can't we assume the `for` bock captures `tc`? I mean, doesn't the `for` block hold it's own reference to `tc` value  (from the moment of the blocks creation)?

Are you using any static analysis to detect those potential flaws?

server/etcdserver/apply/apply_kv_test.go
```diff
@@ -0,0 +1,415 @@
...
+			require.Greater(t, resp.Header.Revision, prevRev)
+			prevRev = resp.Header.Revision
+		}
+	}
+}
+
+func defaultApplierV3Backend(t *testing.T) (*applierV3backend, backend.Backend) {
+	lg := zaptest.NewLogger(t)
+	be, _ := betesting.NewDefaultTmpBackend(t)
+	t.Cleanup(func() { be = nil })
```
fuweid, line 76: https://github.com/etcd-io/etcd/blob/af07f18dce96d4418bc1e89710675a44e3e5028b/server/storage/mvcc/kv_test.go#L758-L761

You can checkout this. We should call the `Close` to tear down the background goroutines.

server/etcdserver/apply/apply_kv_test.go
fuweid: Becasue the address of `tc` is the same to each loop. If there are two goroutines to read the same var, it's not expected. Based on your current code, the `t.Run` is running in sync. It can work well. Using `tc := tc` is to prevent from being shadowed if we enable `t.Parallel` in the sub-test.

> Are you using any static analysis to detect those potential flaws?

As far as I know, I don't see static analysis about this. :(

> NOTE: You can checkout this link https://github.com/golang/go/wiki/LoopvarExperiment. Hope it can help.

Issue #16242, opened 2023-07-14
Add the ability to do etcd snapshot to a pipe or a named pipe as a pseudo file


Ais8Ooz8:
### What would you like to be added?

An additional feature is needed within the ```etcdctl snapshot```. For example, ```save``` command saves to a file, ```pipe``` command sends to stdout.

### Why is this needed?

For security reasons, it may be preferable not to save etcd data to disk, but only to keep it in memory and pass it to a subsequent command.

sauerburger:
I agree with @jmhbnz . It would be beneficial to get the snapshot as a stream in memory and encrypt the content before it goes to a persistent disk.

evgkrsk:
It is not very convenient to write snapshot to disk just to get compressed version.

arichtman:
FWIW I took a crack at using the process's stdout file descriptor but it seems not to work since etcd wants to create the snapshot in parts. Symlinking the FD into another directory allowed the snapshot to succeed but it just overwrote the symlink.

<img width="1117" height="120" alt="Image" src="https://github.com/user-attachments/assets/46a067fb-25e1-49a5-a78a-ca0164021356" />

PR #16243, opened 2023-07-14
etcdctl: add etcdctl snapshot pipe command

Ais8Ooz8:
To improve the security of etcdctl. Added the ability to write snapshots to stdout without writing data to disk. Useful for read-only file systems.

Solves https://github.com/etcd-io/etcd/issues/16242

### DIFF
--- a/client/v3/snapshot/v3_snapshot.go
+++ b/client/v3/snapshot/v3_snapshot.go
@@ -97,7 +97,11 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 	defer os.RemoveAll(partPath)
 	defer f.Close()
 
-	version, _ := WriteSnapshotWithVersion(ctx, lg, cfg, f)
+	var version string
+	version, err = WriteSnapshotWithVersion(ctx, lg, cfg, f)
+	if err != nil {
+		return version, err
+	}
 	if err = os.Rename(partPath, dbPath); err != nil {
 		return version, fmt.Errorf("could not rename %s to %s (%v)", partPath, dbPath, err)
 	}


jmhbnz (commented):
Thanks for putting this proposal forward @Ais8Ooz8.

My initial feeling is that there is now unnecessary duplication here. Just as an idea we could maybe mostly refactor the `SaveWithVersion` and `PipeWithVersion` functions into a single new `SaveSnapshotWithVersion` function that accepts an additional io writer / file parameter?

This way, we could reuse most of the logic while providing flexibility in saving the snapshot to different destinations, such as an actual file or `os.Stdout`.

The implementations for Save vs Pipe would then become something roughly like this perhaps:

```go
func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, dbPath string) (string, error) {
	file, err := os.OpenFile(dbPath+".part", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, fileutil.PrivateFileMode)
	if err != nil {
		return "", fmt.Errorf("could not open %s (%v)", dbPath+".part", err)
	}
	defer os.RemoveAll(dbPath + ".part")
	defer file.Close()

	return SaveSnapshotWithVersion(ctx, lg, cfg, file)
}

func PipeWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config) (string, error) {
	return SaveSnapshotWithVersion(ctx, lg, cfg, os.Stdout)
}

```


Just an idea and not tested at all so feel free to hold off until more people review.

siyuanfoundation:
> This way, we could reuse most of the logic while providing flexibility in saving the snapshot to different destinations, such as an actual file or `os.Stdout`.

I agree with James. Much of the logic can be refactored by passing a file writer into `SaveWithVersion`.

@Ais8Ooz8 can you followup on this? Thanks!

client/v3/snapshot/v3_snapshot.go
```diff
@@ -37,15 +38,14 @@ func hasChecksum(n int64) bool {
...
 // to target path and returns server version. If the context "ctx" is canceled or timed out,
 // snapshot save stream will error out (e.g. context.Canceled,
 // context.DeadlineExceeded). Make sure to specify only one endpoint
 // in client configuration. Snapshot API must be requested to a
 // selected node, and saved snapshot is the point-in-time state of
-// the selected node.
-// Etcd <v3.6 will return "" as version.
-func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, dbPath string) (string, error) {
+// the selected node. Etcd < v3.6 will return "" as version.
+func GetSnapshotWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, f *os.File) (string, error) {
```
siyuanfoundation, line 48: I think a better name would be `WriteSnapshotWithVersion` because the snapshot is not in the return value.

siyuanfoundation:
@Ais8Ooz8 can you fix the fmt error? It is a nitty one about an empty line. Thanks!

etcdctl/README.md
```diff
@@ -1012,6 +1012,21 @@ DEFRAG returns a zero exit code only if it succeeded defragmenting all given end
...
+
+#### Output
+
+The backend snapshot is written to stdout.
+
+#### Example
+
+Write a snapshot to stdout:
+```
+./etcdctl snapshot pipe
```
jmhbnz, line 1027: Nit: For a safer example we could redirect the stdout to a file as dumping tons of stdout in a users terminal can sometimes have negative consequences.

```suggestion
./etcdctl snapshot pipe > snapshot.db
```

client/v3/snapshot/v3_snapshot.go
```diff
@@ -37,15 +38,14 @@ func hasChecksum(n int64) bool {
...
+// WriteSnapshotWithVersion fetches snapshot from remote etcd server, saves data
 // to target path and returns server version. If the context "ctx" is canceled or timed out,
 // snapshot save stream will error out (e.g. context.Canceled,
 // context.DeadlineExceeded). Make sure to specify only one endpoint
 // in client configuration. Snapshot API must be requested to a
 // selected node, and saved snapshot is the point-in-time state of
-// the selected node.
-// Etcd <v3.6 will return "" as version.
-func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, dbPath string) (string, error) {
+// the selected node. Etcd < v3.6 will return "" as version.
```
jmhbnz, line 47: Nit: etcd is always spelled lowercase.

```suggestion
// the selected node. etcd < v3.6 will return "" as version.
```

jmhbnz (approved):
Overall looks good to me - Thanks @Ais8Ooz8.  A couple small suggestions below if you have time.

Additionally - Given this is user facing new feature I would suggest we add an entry to https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#etcdctl-v3.

jmhbnz:
Discussed during sig-etcd triage - this is looking good, @ahrtr could you please take a look for final review?

client/v3/snapshot/v3_snapshot.go
```diff
@@ -94,9 +85,31 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
...
 	)
 
-	if err = os.Rename(partpath, dbPath); err != nil {
-		return resp.Version, fmt.Errorf("could not rename %s to %s (%v)", partpath, dbPath, err)
+	partPath := f.Name()
+	dbPath := strings.TrimSuffix(partPath, ".part")
+	if f != os.Stdout {
+		if err := os.Rename(partPath, dbPath); err != nil {
+			return resp.Version, fmt.Errorf("could not rename %s to %s (%v)", partPath, dbPath, err)
+		}
```
ahrtr, line 93: We should move these to `SaveWithVersion`. 

`WriteSnapshotWithVersion` is a generic function, it shouldn't care about such details which only applies to normal `snapshot save` functionality.

ahrtr:
Is there a real use case for the `etcdctl snapshot pipe` command? @Ais8Ooz8

serathius:
QQ, isn't the UNIX convention to use `-` as a file to mean writing to pipe?
So that would be `etcdctl snapshot safe -`

siyuanfoundation (approved):

ahrtr (approved):
LGTM

Thanks @Ais8Ooz8 for your first contribution!

ahrtr:
Since @serathius has a comment on adding a flag instead of a new command, so leave to @serathius to take a second look.

Either a flag or a new command works for me.  A new command is slight clearer, but not a big deal in this case.

siyuanfoundation:
@serathius This PR is still waiting for your review. Thanks!

PR #16245, opened 2023-07-15
Add `etcd_selector` into all dashboard queries 

v-zhuravlev:
This adds `etcd_selector` into all dashboard queries.

Otherwise common metrics like 'process_resident_memory_bytes' can return non etcd metrics (when `clusterLabel` is 'cluster' for example)

### DIFF
--- a/contrib/mixin/.lint
+++ b/contrib/mixin/.lint
@@ -1,0 +1,18 @@
+---
+exclusions:
+  template-instance-rule:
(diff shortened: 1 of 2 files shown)


jmhbnz:
Hey @v-zhuravlev - Suggest we close this one and focus on getting https://github.com/etcd-io/etcd/pull/16246 merged, does that sound ok?

v-zhuravlev:
Sure.

PR #16246, opened 2023-07-15
Refactor mixin with grafonnet

v-zhuravlev:
Uses new [grafonnet](https://grafana.github.io/grafonnet/) lib to declare dashboard. Generated dashboard has same layout, but now has timeseries panels instead of deprecated graphs

Also includes changes made in #16245.
image

It can serve as foundation for future dashboard declarative changes.


serathius:
cc @tomwilkie @lilic to help with review

lilic:
The https://github.com/etcd-io/etcd/pull/16245 change sounds goods to me. 👍 

> Uses new [grafonnet](https://grafana.github.io/grafonnet/) lib to declare dashboard.

Was there a need to move to use grafonnet to fix the "timeseries panels"? I want to understand the large change here.

cc @simonpasquier for your awareness.

v-zhuravlev:
Yes, an old grafonnet didn't support timeseries panels as it had been stuck in grafana-7 era.
New grafonnet supports timeseries, as well as always up to date with the latest grafana schema as it is autogenerated from it. So this shift can also become first step for other dashboard improvements.

lilic:
@v-zhuravlev Thanks for the answer! Will this new dashboard still work in the older Grafana versions, for example for companies that cannot use the newer Grafana versions due to the  AGPL-3.0 license change?

v-zhuravlev:
Grafana Dashboard Schema for versions 9.x-10.x changed in comparison to 7.x, and while generally newer schemas dashboards can be imported into previous versions of Grafana, this is not guaranteed that it would work 100%.

If this is critical, I think as an option old grafana mixin can be kept alongside this one.

jmhbnz (approved):
Thanks for completing this overhaul @v-zhuravlev.

I've tested the generated dashboard import on the following versions of grafana:

| Version | Outcome          |
|---------|------------------|
| 7.5.16  | Dashboard broken `Templating: Failed to upgrade legacy queries e.replace is not a function.` |
| 8.5.27  | Working 🎉         |
| 9.5.7   | Working 🎉          |
| 10.0.3   | Working 🎉 |

Given grafana 7.x active support ended 2 years ago (08 Jun 2021) and even security maintenance support has now ended for over a year personally I don't have any issue with breaking 7.x compatibility.

Note: I am keen to work on improvements to the documentation we provide in README.md and also the website. We have https://github.com/etcd-io/website/issues/690 open to track this.

jmhbnz:
Hey @v-zhuravlev - In the interest of getting these changes through review it might be more palatable if we retain the old version alongside.

We could have subdirectories under `contrib/mixin` for `grafana7x` and `grafana8x+`.   Keen to hear what others think of this approach.

lilic:
> We could have subdirectories under contrib/mixin for grafana7x and grafana8x+. Keen to hear what others think of this approach.

I think that's a great suggestion! This was it works for companies who cannot use Grafana due to the relicensing.

v-zhuravlev:
ok, I kept 7x dashboard under flag grafana7x. One can set `grafana7x: true` in config to generate 7x compatible dashboard instead of new one. WDYT?

v-zhuravlev:
> > ok, I kept 7x dashboard under flag grafana7x. One can set grafana7x: true in config to generate 7x compatible dashboard instead of new one. WDYT?
> 
> Sounds good. Lets make sure to document this. Otherwise lgtm.
> 
> cc @serathius this looks like a good comprise to me.

updated readme about this.

jmhbnz (approved):
Nice work @v-zhuravlev. Confirming the new conditional flag is working as expected at my end.  

I've tested generated etcd dashboard on grafana `7.x` with the flag and working.  Also repeated the json import on `10.x` without the flag set to false and also working.

I think this looks good! 🙏🏻

lilic:
> Can we even move the mixin into a separate repo and attract some dedicated people to maintain it?

Good idea! But I think this will break things for existing users?

cc @brancz for jsonnet review.

ahrtr:
> Good idea! But I think this will break things for existing users?

We can update the readmd.md to redirect users to the new location.

lilic:
> We can update the README.md to redirect users to the new location.

Sounds good. I would say we can move it and keep an old copy for one or two releases, and make changes only in the new repo.

This is how the imports paths work so users would need to change it. https://github.com/prometheus-operator/kube-prometheus/blob/18ed0ffb8754e6712dfa1fc063947848a5c1a6a8/jsonnet/kube-prometheus/jsonnetfile.json#L26

ahrtr:
@jmhbnz @serathius @wenjiaswe WDYT?

jmhbnz:
> Can we even move the mixin into a separate repo and attract some dedicated people to maintain it?

Not opposed to splitting it out, defer to the maintainer consensus on this.

FYI There appear to be 37 references in github to the current path: https://github.com/search?q=github.com%2Fetcd-io%2Fetcd%2Fcontrib%2Fmixin%2Fmixin&type=code.  I can take care of the OpenShift and prometheus operator entries in that list.  We would also need to update the website and as mentioned retain old copy at current path for a while.

wenjiaswe:
@CatherineF-dev Could you please help to take a look?

ahrtr:
> Can we even move the mixin into a separate repo and attract some dedicated people to maintain it?

This shouldn't be a hard decision. 
- What do you think? @mitake  @ptabor @serathius @spzala @wenjiaswe 
- Who is interested in maintaining the new repo `mixin` (or any new name?) after the separation? @jmhbnz @lilic @tomwilkie @brancz  

Firstly of all, could anyone review this PR so that we can get it merged firstly? @lilic @brancz

jmhbnz:
> Who is interested in maintaining the new repo mixin (or any new name?) after the separation? 

I am comfortable to maintain it separately as `etcd-io/mixin`, provided there is at least one other active maintainer I can work with on it.

serathius (approved):
Let's not waste great work and take the improvements to dashboards. It might help to attract more interest. 

Moving to separate repo is an option in the future, however I would not prioritize it as long there is no maintenance cost.

ahrtr:
> Let's not waste great work

Have you carefully reviewed this? :) I do not have full context, nor time to review this. 

Merging this PR because
- Verified by @jmhbnz (thx)
- No any objection so far
- It isn't a critical part of etcd

Issue #16255, opened 2023-07-17
pkg/v3/ioutil TestPageWriterRandom is flaking


Rajalakshmi-Girish:
### Which github workflows are flaking?

Jobs in our internal CI are flaking

### Which tests are flaking?

pkg/v3/ioutil TestPageWriterRandom

### Reason for failure (if possible)

The test is failing https://github.com/etcd-io/etcd/blob/main/pkg/ioutil/pagewriter_test.go#L43

We initially noticed it in `ppc64le` architecture that our CI jobs use.
Later when verified, this test is failing even on **x86_64**
```
[root@toad1 etcd]# go test -v go.etcd.io/etcd/pkg/v3/ioutil -c -count=1
go: downloading github.com/stretchr/testify v1.8.4
go: downloading go.uber.org/zap v1.24.0
go: downloading golang.org/x/sys v0.10.0
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading github.com/pmezard/go-difflib v1.0.0
go: downloading gopkg.in/yaml.v3 v3.0.1
go: downloading go.uber.org/atomic v1.7.0
go: downloading go.uber.org/multierr v1.11.0
[root@toad1 etcd]# ./ioutil.test -test.run TestPageWriterRandom
PASS
[root@toad1 etcd]# ./ioutil.test -test.run TestPageWriterRandom
PASS
[root@toad1 etcd]# ./ioutil.test -test.run TestPageWriterRandom
PASS
[root@toad1 etcd]# ./ioutil.test -test.run TestPageWriterRandom
--- FAIL: TestPageWriterRandom (0.00s)
    pagewriter_test.go:43: got 5016 bytes pending, expected less than 128 bytes
FAIL
[root@toad1 etcd]# ./ioutil.test -test.run TestPageWriterRandom
PASS
[root@toad1 etcd]# ./ioutil.test -test.run TestPageWriterRandom
PASS
[root@toad1 etcd]# uname -a
Linux toad1.fyre.ibm.com 4.18.0-358.el8.x86_64 #1 SMP Mon Jan 10 13:11:20 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
[root@toad1 etcd]# go version
go version go1.20.6 linux/amd64
[root@toad1 etcd]#
```

### Anything else we need to know?

Also flaking in `ppc64le` architecture:
```
[root@raji-workspace etcd]# ./ioutil.test -test.run TestPageWriterRandom
PASS
[...]

PR #16268, opened 2023-07-18
pkg/ioutil: deflake TestPageWriterRandom

fuweid:
The PageWriter has cache buffer so that it doesn't call the Writer until the cache is almost full. Since the data's length is random, the pending bytes should be always less than cache buffer size, instead of page size.

Fix: #16255


jmhbnz (approved):
LGTM - Thanks @fuweid

ahrtr (approved):

ahrtr:
Could anyone backport this PR to 3.4 as well? 

FYI. https://github.com/etcd-io/etcd/actions/runs/6123643643/job/16621962420?pr=16556

Issue #16159, opened 2023-06-30
Move revision related common functions to pkg


dusk125:
### What would you like to be added?

This is a followup to #16029; revision functionality was copied out of the [server package ](https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/revision.go) into the etcdutl package, this issue will have those duplicated functions moved to a common area to de-duplicate them.

### Why is this needed?

This is needed to remove dual maintenance of the revision related functions.

dusk125:
I can handle this issue, if someone can assign it to me, thanks.

PR #16269, opened 2023-07-18
Refactor common revision code to pkg

dusk125:
Addresses #16159.

Moves the revision structure definition and related functions from package mvcc in the server, to a subpackage in pkg.


server/storage/mvcc/kvstore.go
```diff
@@ -41,9 +43,8 @@ const (
 	// markedRevBytesLen is the byte length of marked revision.
```
fuweid, line 43: the comment needs to be changed for `markBytePosition`?

fuweid (commented):
The changes look good.

It might help to #15979.

server/storage/mvcc/kvstore.go
```diff
@@ -41,9 +43,8 @@ const (
...
 	// The first `revBytesLen` bytes represents a normal revision. The last
 	// one byte is the mark.
-	markedRevBytesLen      = revBytesLen + 1
-	markBytePosition       = markedRevBytesLen - 1
-	markTombstone     byte = 't'
+	markBytePosition      = revision.MarkedRevBytesLen - 1
```
serathius, line 46: Tombstone is part of revision.

serathius (changes requested):
This refactor incorrectly mixes revision as a transaction identifier and revision serialization used for mvcc. In the first case the revision has 17 bytes in second has 18 bytes as it includes tombstone.

The original code is no better, but at least it's private and separated, so changes in logic will not break etcdserver.

serathius:
I would propose to do the refactoring in different order. Proposed steps:
* Move `server/storage/mvcc/kvstore.go` revision/tombstone logic into `server/storage/mvcc/revision.go`. Functions like `isTombstone` and `appendMarkTombstone`.
* Create a `RevisionTombstone` struct that will include information about tombstone.
...

ahrtr:
1. We do need to make revision and related common method & functions public (under the `pkg/v3`), as we already explicitly document it in [The Implicit Kubernetes-ETCD Contract](https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit#heading=h.tlkin1a8b8bl). It can also reduce the previously copy-pasted duplicated code
    - The revision struct
    - Method `GreatThan`
    - Function `RevToBytes` and `BytesToRev`
2. Keep all others, including tomestome, unchanged. It's open to discussion on how to improve it, but they are not necessarily to be done in this PR.

serathius:
> We do need to make revision and related common method & functions public (under the `pkg/v3`), as we already explicitly document it in [The Implicit Kubernetes-ETCD Contract](https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit#heading=h.tlkin1a8b8bl). It can also reduce the previously copy-pasted duplicated code
> * The `Revision` struct and the `const RevBytesLen = 8 + 1 + 8`
> * Method `GreatThan`
> * Function `RevToBytes` and `BytesToRev`

We don't `need` to do anything if it increases risk of bug. Code duplication is not inherently bad, but exposing internals without proper care is. 

The The Implicit Kubernetes-ETCD Contract is totally unrelated. Please understand that revision as a concept for versioning changes is separate from how etcd stores revision on disk. Method `GreatThan` belongs to the first one, `RevBytesLen`, `RevToBytes` and `BytesToRev` belong to the second one.  Mixing them together will lead to subtle bug.

ahrtr:
> We don't `need` to do anything if it increases risk of bug. Code duplication is not inherently bad, but exposing internals without proper care is.
> 
> The The Implicit Kubernetes-ETCD Contract is totally unrelated. Please understand that revision as a concept for versioning changes is separate from how etcd stores revision on disk. Method `GreatThan` belongs to the first one, `RevBytesLen`, `RevToBytes` and `BytesToRev` belong to the second one. Mixing them together will lead to subtle bugs.

It partially makes sense to me. The revision storage format is just etcd's internal mechanism, and users do not care about it. So from this perspective, it doesn't make sense to expose them. But it's a common mechanism inside etcd, so a modification based on my above comment https://github.com/etcd-io/etcd/pull/16269#issuecomment-1641838402 is to get them included in `pkg/v3/internal/revision`.

serathius:
> It partially makes sense to me. The revision storage format is just etcd's internal mechanism, and users do not care about it. So from this perspective, it doesn't make sense to expose them. But it's a common mechanism inside etcd, so a modification based on my above comment https://github.com/etcd-io/etcd/pull/16269#issuecomment-1641838402 is to get them included in pkg/v3/internal/revision.

Right, I'm not against exposing the etcd storage mechanism. I'm postulating code should:
* Make clear separation between the concept revision for versioning and the revision storage representation. Both can be exposed, but each clearly distinguished. 
* Include the tombstone into the storage representation.

ahrtr:
Based on above discussion, the actions:
- Get the revision struct and related method & functions included in package `pkg/v3/internal/revision`;
- Get tombstone related definitions included in `pkg/v3/internal/revision` as well.
    - markedRevBytesLen
    - markBytePosition
    - markTombstone
    - IsTombstone()
    - AppendMarkTombstone()

fuweid:
> Include the tombstone into the storage representation. If we are exposing the etcd internal representation of revision, we need to remember that tombstone is part of this representation. You cannot read the bbolt Keys bucket and forget the tombstone.

Good point !

dusk125:
> I would propose to do the refactoring in different order. Proposed steps:
> 
> * Move `server/storage/mvcc/kvstore.go` revision/tombstone logic into `server/storage/mvcc/revision.go`. Functions like `isTombstone` and `appendMarkTombstone`.
> * Create a `RevisionTombstone` struct that will include information about tombstone.
> * Change `BytesToRev`, `NewRevBytes` and `RevToBytes` to use `RevisionTombstone`
> * And only then make the `revision` package public.

@serathius just so I make sure I'm understanding you properly. Are you suggesting to have both mvcc.RevisionTombstone (for internal storage usage in mvcc) and revision.Revision for external usage in things like etcdutl?

serathius:
Yes, etcdutl reads data stored in `Key` bucket that might include revisions with tombstone. It needs to properly read and store the tombstone.

server/storage/mvcc/revision.go
```diff
@@ -14,48 +14,84 @@
...
+	// one byte is the mark.
+	markedRevBytesLen      = revBytesLen + 1
+	markBytePosition       = markedRevBytesLen - 1
+	markTombstone     byte = 't'
+)
 
-	// sub is the sub revision of a change in a set of changes that happen
+// RevisionTombstone indicates modification of the key-value space.
+// The set of changes that share same main revision changes the key-value space atomically.
+type RevisionTombstone struct {
```
serathius, line 38: Note, when we transfer revision from or to bytes in MVCC, we are always doing operation on RevisionTombstone. However, when we compare a revision, we always operate on only the Revision.

Have you thought about creating a sub-struct for revision? For example:
```
type RevisionTombstone struct {
  Revision
  Tombstone bool
}
type Revision struct {
  Main int64
  Sub int64
}
```

server/storage/mvcc/revision.go
```diff
@@ -14,48 +14,84 @@
...
+	// markedRevBytesLen is the byte length of marked revision.
+	// The first `revBytesLen` bytes represents a normal revision. The last
+	// one byte is the mark.
+	markedRevBytesLen      = revBytesLen + 1
+	markBytePosition       = markedRevBytesLen - 1
+	markTombstone     byte = 't'
+)
 
-	// sub is the sub revision of a change in a set of changes that happen
+// RevisionTombstone indicates modification of the key-value space.
```
serathius, line 36: RevisionTombstone as a name describes that it holds, but it's hard to understand what's its purpose. 
The etcd MVCC managed they bbolt key bucket, which uses RevisionTombstone as a key. Have you thought about naming it based on it's purpose? For example `mvcc.Key` or `mvcc.BucketKey`. This

server/storage/mvcc/revision.go
```diff
@@ -14,48 +14,84 @@
...
+// The set of changes that share same main revision changes the key-value space atomically.
+type RevisionTombstone struct {
+	// Main is the main revision of a set of changes that happen atomically.
+	Main int64
+	// Sub is the sub revision of a change in a set of changes that happen
 	// atomically. Each change has different increasing sub revision in that
 	// set.
-	sub int64
+	Sub       int64
+	tombstone byte
```
serathius, line 45: Tombstone is represented as a byte 't', however it's just a true/false flag.

server/storage/mvcc/kvstore_txn.go
```diff
@@ -279,9 +279,9 @@ func (tw *storeTxnWrite) deleteRange(key, end []byte) int64 {
...
 
 func (tw *storeTxnWrite) delete(key []byte) {
-	ibytes := newRevBytes()
-	idxRev := revision{main: tw.beginRev + 1, sub: int64(len(tw.changes))}
-	revToBytes(idxRev, ibytes)
+	ibytes := NewRevBytes()
+	idxRev := RevisionTombstone{Main: tw.beginRev + 1, Sub: int64(len(tw.changes))}
+	RevToBytes(idxRev, ibytes)
 
 	ibytes = appendMarkTombstone(tw.storeTxnCommon.s.lg, ibytes)
```
serathius, line 287: What do you think about setting tombstone on RevisionTombstone struct, instead of setting it separetly via appendMarkTombstone?
```suggestion
	idxRev := RevisionTombstone{Main: tw.beginRev + 1, Sub: int64(len(tw.changes)), tombstone: true}
	RevToBytes(idxRev, ibytes)
```

server/storage/mvcc/revision.go
dusk125: Yeah I had considered it, but given the previous comments, I wasn't sure if that would be too tightly coupling the mvcc and versioning. Understanding that better now, I'll make this change

server/storage/mvcc/kvstore_txn.go
dusk125: Yeah I like that

server/storage/mvcc/hash.go
```diff
@@ -30,7 +30,7 @@ const (
 	hashStorageMaxSize = 10
 )
 
-func unsafeHashByRev(tx backend.UnsafeReader, compactRevision, revision int64, keep map[revision]struct{}) (KeyValueHash, error) {
+func unsafeHashByRev(tx backend.UnsafeReader, compactRevision, revision int64, keep map[BucketKey]struct{}) (KeyValueHash, error) {
```
serathius, line 33:
```suggestion
func unsafeHashByRev(tx backend.UnsafeReader, compactRevision, revision int64, keep map[Revision]struct{}) (KeyValueHash, error) {
```

server/storage/mvcc/hash.go
```diff
@@ -43,10 +43,10 @@ type kvHasher struct {
 	hash            hash.Hash32
 	compactRevision int64
 	revision        int64
-	keep            map[revision]struct{}
+	keep            map[BucketKey]struct{}
```
serathius, line 46:
```suggestion
	keep            map[Revision]struct{}
```

server/storage/mvcc/index.go
```diff
@@ -22,14 +22,14 @@ import (
 )
 
 type index interface {
```
serathius, line 24: This interface should be `Revision` based. Requests are about a specific revision, not key in bucket.

server/storage/mvcc/index_test.go
```diff
@@ -63,7 +63,7 @@ func TestIndexGet(t *testing.T) {
 
 func TestIndexRange(t *testing.T) {
```
serathius, line 64: This test is about revision, not bucket keys.

server/storage/mvcc/index_test.go
```diff
@@ -133,15 +133,15 @@ func TestIndexTombstone(t *testing.T) {
...
+	err = ti.Tombstone([]byte("foo"), newBucketKey(3, 0, false))
 	if err != ErrRevisionNotFound {
 		t.Errorf("tombstone error = %v, want %v", err, ErrRevisionNotFound)
 	}
 }
 
 func TestIndexRevision(t *testing.T) {
 	allKeys := [][]byte{[]byte("foo"), []byte("foo1"), []byte("foo2"), []byte("foo2"), []byte("foo1"), []byte("foo")}
-	allRevs := []revision{{main: 1}, {main: 2}, {main: 3}, {main: 4}, {main: 5}, {main: 6}}
+	allRevs := []BucketKey{newBucketKey(1, 0, false), newBucketKey(2, 0, false), newBucketKey(3, 0, false), newBucketKey(4, 0, false), newBucketKey(5, 0, false), newBucketKey(6, 0, false)}
```
serathius, line 144: ditto

server/storage/mvcc/key_index.go
```diff
@@ -73,21 +73,21 @@ var (
 //	{empty} -> key SHOULD be removed.
 type keyIndex struct {
 	key         []byte
-	modified    revision // the main rev of the last modification
+	modified    BucketKey // the main rev of the last modification
```
serathius, line 76: ditto

tools/.golangci.yaml
```diff
@@ -23,7 +23,7 @@ linters:
     - staticcheck
     - stylecheck
     - unused
-    - unconvert  # Remove unnecessary type conversions
+    - unconvert # Remove unnecessary type conversions
```
serathius, line 26: No need to touch this.

serathius:
Please note that mvcc is the package where `Revision` and `BucketKey` meet. Etcd storage will make requests about a specific revisions, so it should use `Revision` struct. However when mvcc writes to bbolt it writes it via `BucketKey`. So basically before writing a revision it needs to package the `Revision` struct into `BucketKey` and decide whether to set the tombstone.

tools/.golangci.yaml
dusk125: That's getting changed by `make fix`.

serathius: Hmm, @fuweid do you know why?
Can you send a separate PR?

fuweid: @dusk125 Thanks! Let me fix this

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -352,38 +352,38 @@ func (s *v3Manager) modifyLatestRevision(bumpAmount uint64) error {
 	return nil
 }
 
-func (s *v3Manager) unsafeBumpRevision(tx backend.UnsafeWriter, latest revision, amount int64) revision {
+func (s *v3Manager) unsafeBumpRevision(tx backend.UnsafeWriter, latest mvcc.BucketKey, amount int64) mvcc.BucketKey {
```
serathius, line 355: Would it make it more clear?
```suggestion
func (s *v3Manager) unsafeBumpRevisionInBucket(tx backend.UnsafeWriter, latest mvcc.BucketKey, amount int64) mvcc.BucketKey {
```

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -352,38 +352,38 @@ func (s *v3Manager) modifyLatestRevision(bumpAmount uint64) error {
...
+	latest.Sub = 0
+	k := mvcc.NewRevBytes()
+	k = mvcc.RevToBytes(latest, k)
 	tx.UnsafePut(schema.Key, k, []byte{})
 
 	return latest
 }
 
-func (s *v3Manager) unsafeMarkRevisionCompacted(tx backend.UnsafeWriter, latest revision) {
+func (s *v3Manager) unsafeMarkRevisionCompacted(tx backend.UnsafeWriter, latest mvcc.BucketKey) {
```
serathius, line 372: This should take Revision as argument

serathius, line 372:
```suggestion
func (s *v3Manager) unsafeMarkRevisionCompacted(tx backend.UnsafeWriter, latest mvcc.Revision) {
```

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -352,38 +352,38 @@ func (s *v3Manager) modifyLatestRevision(bumpAmount uint64) error {
...
+		zap.Int64("revision", latest.Main),
 	)
 
-	mvcc.UnsafeSetScheduledCompact(tx, latest.main)
+	mvcc.UnsafeSetScheduledCompact(tx, latest.Main)
 }
 
-func (s *v3Manager) unsafeGetLatestRevision(tx backend.UnsafeReader) (revision, error) {
-	var latest revision
+func (s *v3Manager) unsafeGetLatestRevision(tx backend.UnsafeReader) (mvcc.BucketKey, error) {
```
serathius, line 381:
```suggestion
func (s *v3Manager) unsafeGetLatestRevision(tx backend.UnsafeReader) (mvcc.Revision, error) {
```

etcdutl/snapshot/v3_snapshot.go
serathius, line 355:
```suggestion
func (s *v3Manager) unsafeBumpBucketsRevision(tx backend.UnsafeWriter, latest mvcc.BucketKey, amount int64) mvcc.BucketKey {
```

serathius (approved):
cc @ahrtr

server/storage/mvcc/revision.go
```diff
@@ -14,48 +14,89 @@
...
+	}
+}
+
+func NewRevBytes() []byte {
 	return make([]byte, revBytesLen, markedRevBytesLen)
 }
 
-func revToBytes(rev revision, bytes []byte) {
-	binary.BigEndian.PutUint64(bytes, uint64(rev.main))
+func RevToBytes(rev BucketKey, bytes []byte) []byte {
```
ahrtr, line 74:
```suggestion
func BucketKeyToBytes(rev BucketKey, bytes []byte) []byte {
```

Also add a separate `RevToBytes` something like below,
```
func RevToBytes(rev Revision, bytes []byte) []byte {
    return BucketKeyToBytes(newBucketKey(rev.Main, rev.Sub, false))
}
```

server/storage/mvcc/revision.go
```diff
@@ -14,48 +14,89 @@
...
+		}
+	}
+	return bytes
 }
 
-func bytesToRev(bytes []byte) revision {
-	return revision{
-		main: int64(binary.BigEndian.Uint64(bytes[0:8])),
-		sub:  int64(binary.BigEndian.Uint64(bytes[9:])),
+func BytesToRev(bytes []byte) BucketKey {
```
ahrtr, line 89:
```suggestion
func BytesToBucketKey(bytes []byte) BucketKey {
```

Also add a separate `BytesToRev`,
```
func BytesToRev(bytes []byte) Revision {
    return BytesToBucketKey(bytes).Revision
}
```

ahrtr:
I understand that the `BucketKey` is representing the format stored in disk, while `Revision` is representing the structure in memory.

But I find it's confusing when reviewing the PR, and sometimes there is no clear boundary when to use which one and which functions (`RevToBytes` vs `RevToBucketKey`, `BytesToRev` vs `BytesToBucketKey`). I believe other contributors may be confused as well when they read or update the related source code. 

Actually the [markTombstone](https://github.com/etcd-io/etcd/blob/de39c750530f3eea1682eefab2278c87552adece/server/storage/mvcc/kvstore.go#L46) is only used in a couple of places, and [appendMarkTombstone](https://github.com/etcd-io/etcd/blob/de39c750530f3eea1682eefab2278c87552adece/server/storage/mvcc/kvstore_txn.go#L286) is only used in `delete`.

So I suggest to remove `BucketKey` to get rid of the unnecessary confusion, and keep the original functions `isTombstone` and `appendMarkTombstone` from pragmatic perspective.

server/storage/mvcc/kvstore_txn.go
```diff
@@ -202,13 +202,13 @@ func (tw *storeTxnWrite) put(key, value []byte, leaseID lease.LeaseID) {
...
+		c = created.Main
 		oldLease = tw.s.le.GetLease(lease.LeaseItem{Key: string(key)})
 		tw.trace.Step("get key's previous created_revision and leaseID")
 	}
-	ibytes := newRevBytes()
-	idxRev := revision{main: rev, sub: int64(len(tw.changes))}
-	revToBytes(idxRev, ibytes)
+	ibytes := NewRevBytes()
+	idxRev := newBucketKey(rev, int64(len(tw.changes)), false)
+	ibytes = BucketKeyToBytes(idxRev, ibytes)
```
ahrtr, line 211:
```suggestion
	idxRev := Revision{Main: rev, Sub: int64(len(tw.changes))}
	ibytes = RevToBytes(idxRev, ibytes)
```

server/storage/mvcc/store.go
```diff
@@ -22,15 +22,15 @@ import (
 func UnsafeReadFinishedCompact(tx backend.UnsafeReader) (finishedComact int64, found bool) {
 	_, finishedCompactBytes := tx.UnsafeRange(schema.Meta, schema.FinishedCompactKeyName, nil, 0)
 	if len(finishedCompactBytes) != 0 {
-		return bytesToRev(finishedCompactBytes[0]).main, true
+		return BytesToBucketKey(finishedCompactBytes[0]).Main, true
```
ahrtr, line 25:
```suggestion
		return BytesToRev(finishedCompactBytes[0]).Main, true
```

server/storage/mvcc/store.go
```diff
@@ -22,15 +22,15 @@ import (
...
 	}
 	return 0, false
 }
 
 func UnsafeReadScheduledCompact(tx backend.UnsafeReader) (scheduledComact int64, found bool) {
 	_, scheduledCompactBytes := tx.UnsafeRange(schema.Meta, schema.ScheduledCompactKeyName, nil, 0)
 	if len(scheduledCompactBytes) != 0 {
-		return bytesToRev(scheduledCompactBytes[0]).main, true
+		return BytesToBucketKey(scheduledCompactBytes[0]).Main, true
```
ahrtr, line 33:
```suggestion
		return BytesToRev(scheduledCompactBytes[0]).Main, true
```

server/storage/mvcc/store.go
```diff
@@ -42,8 +42,8 @@ func SetScheduledCompact(tx backend.BatchTx, value int64) {
 }
 
 func UnsafeSetScheduledCompact(tx backend.UnsafeWriter, value int64) {
-	rbytes := newRevBytes()
-	revToBytes(revision{main: value}, rbytes)
+	rbytes := NewRevBytes()
+	rbytes = BucketKeyToBytes(newBucketKey(value, 0, false), rbytes)
```
ahrtr, line 46:
```suggestion
	rbytes = RevToBytes(Revision{Main: value, Sub: 0}, rbytes)
```

server/storage/mvcc/store.go
```diff
@@ -54,7 +54,7 @@ func SetFinishedCompact(tx backend.BatchTx, value int64) {
 }
 
 func UnsafeSetFinishedCompact(tx backend.UnsafeWriter, value int64) {
-	rbytes := newRevBytes()
-	revToBytes(revision{main: value}, rbytes)
+	rbytes := NewRevBytes()
+	rbytes = BucketKeyToBytes(newBucketKey(value, 0, false), rbytes)
```
ahrtr, line 58:
```suggestion
	rbytes = RevToBytes(Revision{Main: value, Sub: 0}, rbytes)
```

server/storage/mvcc/watchable_store.go
```diff
@@ -352,9 +352,9 @@ func (s *watchableStore) syncWatchers() int {
 	compactionRev := s.store.compactMainRev
 
 	wg, minRev := s.unsynced.choose(maxWatchersPerSync, curRev, compactionRev)
-	minBytes, maxBytes := newRevBytes(), newRevBytes()
-	revToBytes(revision{main: minRev}, minBytes)
-	revToBytes(revision{main: curRev + 1}, maxBytes)
+	minBytes, maxBytes := NewRevBytes(), NewRevBytes()
+	minBytes = BucketKeyToBytes(newBucketKey(minRev, 0, false), minBytes)
+	maxBytes = BucketKeyToBytes(newBucketKey(curRev+1, 0, false), maxBytes)
```
ahrtr, line 357:
```suggestion
	minBytes = RevToBytes(Revision{Main: minRev, Sub: 0}, minBytes)
	maxBytes = RevToBytes(Revision{Main: curRev+1, Sub: 0}, maxBytes)
```

server/storage/mvcc/watchable_store.go
```diff
@@ -428,7 +428,7 @@ func kvsToEvents(lg *zap.Logger, wg *watcherGroup, revs, vals [][]byte) (evs []m
 		if isTombstone(revs[i]) {
 			ty = mvccpb.DELETE
 			// patch in mod revision so watchers won't skip
-			kv.ModRevision = bytesToRev(revs[i]).main
+			kv.ModRevision = BytesToBucketKey(revs[i]).Main
```
ahrtr, line 431:
```suggestion
			kv.ModRevision = BytesToRev(revs[i]).Main
```

ahrtr:
Thanks @dusk125 for the quick update.

Now the `BucketKey` is only used in method `(*storeTxnWrite) delete`, so a more pragmatic approach is to remove `BucketKey` and keep the original `appendMarkTombstone` (Note I notice that the function `isTombstone` isn't removed :)).  But I won't insist on that if others think the current implementation is good. 

Overall looks good to me. Leave to @serathius to take another look.

ahrtr (approved):

PR #16272, opened 2023-07-19
pkg/flags: fix UniqueURLs'Set to remove duplicates in UniqueURLs'uss

gocurr:
From the name of func 'UniqueURLsFromFlag', we can tell that UniqueURLs'uss should not have duplicates. The current implemention of UniqueURLs'Set has a bug to make it unique. This PR fixes it.

Please take a look.

### DIFF
--- a/pkg/flags/unique_urls.go
+++ b/pkg/flags/unique_urls.go
@@ -50,7 +50,11 @@ func (us *UniqueURLs) Set(s string) error {
 	us.Values = make(map[string]struct{})
 	us.uss = make([]url.URL, 0)
 	for _, v := range ss {
-		us.Values[v.String()] = struct{}{}
+		x := v.String()
+		if _, exists := us.Values[x]; exists {
+			continue
+		}
+		us.Values[x] = struct{}{}
 		us.uss = append(us.uss, v)
 	}
 	return nil
--- a/pkg/flags/unique_urls_test.go
+++ b/pkg/flags/unique_urls_test.go
@@ -15,8 +15,11 @@
 package flags
 
 import (
-	"reflect"
+	"flag"
+	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestNewUniqueURLsWithExceptions(t *testing.T) {
@@ -83,11 +86,29 @@ func TestNewUniqueURLsWithExceptions(t *testing.T) {
 	}
 	for i := range tests {
 		uv := NewUniqueURLsWithExceptions(tests[i].s, tests[i].exception)
-		if !reflect.DeepEqual(tests[i].exp, uv.Values) {
-			t.Fatalf("#%d: expected %+v, got %+v", i, tests[i].exp, uv.Values)
-		}
-		if uv.String() != tests[i].rs {
-			t.Fatalf("#%d: expected %q, got %q", i, tests[i].rs, uv.String())
-		}
+		require.Equal(t, tests[i].exp, uv.Values)
+		require.Equal(t, tests[i].rs, uv.String())
 	}
+}
+
+func TestUniqueURLsFromFlag(t *testing.T) {
+	const name = "test"
+	urls := []string{
+		"https://1.2.3.4:1",
+		"https://1.2.3.4:2",
+		"https://1.2.3.4:3",
+		"https://1.2.3.4:1",
+	}
+	fs := flag.NewFlagSet(name, flag.ExitOnError)
+	u := NewUniqueURLsWithExceptions(strings.Join(urls, ","))
+	fs.Var(u, name, "usage")
+	uss := UniqueURLsFromFlag(fs, name)
+
+	require.Equal(t, len(u.Values), len(uss))
+
+	um := make(map[string]struct{})
+	for _, x := range uss {
+		um[x.String()] = struct{}{}
+	}
+	require.Equal(t, u.Values, um)
 }


gocurr:
By the way, I think the 'unique' methods are too complicated in this package. I understand people use map to remove dups, however it can be replaced with a sorted slice (using fast-slow pointers to remove dups), we can also use binary search to indicate an element exists.

pkg/flags/unique_urls.go
```diff
@@ -50,7 +50,11 @@ func (us *UniqueURLs) Set(s string) error {
 	us.Values = make(map[string]struct{})
 	us.uss = make([]url.URL, 0)
 	for _, v := range ss {
-		us.Values[v.String()] = struct{}{}
+		x := v.String()
+		if _, exists := us.Values[x]; exists {
+			continue
+		}
+		us.Values[x] = struct{}{}
```
ahrtr, line 57: If the items already exists, then it just overwrites it. 

This PR seems to be unnecessary.

gocurr: Thanks for your response.
```go
type UniqueURLs struct {
	Values  map[string]struct{}
	uss     []url.URL
	Allowed map[string]struct{}
}
```
UniqueURLs'uss is not documented that it is unique, but it's used in UniqueURLsFromFlag.
```go
// UniqueURLsFromFlag returns a slice from urls got from the flag.
func UniqueURLsFromFlag(fs *flag.FlagSet, urlsFlagName string) []url.URL {
	return (*fs.Lookup(urlsFlagName).Value.(*UniqueURLs)).uss
}
```
From the name of `UniqueURLsFromFlag`, people may think the returned []url.URL is unique, however it is undocumented either. If the returned []url.URL is unique, UniqueURLs'uss should also be unique.

My question is: should UniqueURLs'uss has the same length as UniqueURLs'Values?

Here's a test case for that.
```go
func TestUniqueURLsFromFlag(t *testing.T) {
	const name = "test"
	urls := []string{
		"https://1.2.3.4:1",
		"https://1.2.3.4:2",
		"https://1.2.3.4:3",
		"https://1.2.3.4:1",
	}
	fs := flag.NewFlagSet(name, flag.ExitOnError)
	u := NewUniqueURLsWithExceptions(strings.Join(urls, ","))
	fs.Var(u, name, "usage")
	uss := UniqueURLsFromFlag(fs, name)

	if len(uss) != len(u.Values) {
		t.Fatal("not unique")
	}
}
```

gocurr:
Looking forward to reply. Thanks.

pkg/flags/unique_urls.go
fuweid: Agree with @ahrtr. In your latest change, `continue` is the same to original code. Not follow your idea.

gocurr: Maybe, you misunderstood my meaning.
`Values` is certainty unique, however `uss`(a slice) may have duplicates.

Sorry, I have to question again: 
```text
Should `uss`(a slice) be unique?
```

If `uss`(a slice)  may have duplicates, please close this PR. Thanks.

ahrtr: Good catch! I think `uss` should be unique as well. Could you add a unit test to reproduce the issue (duplicated `uss`), and then confirm that your patch is able to fix the issue?

gocurr: Please take a look.

pkg/flags/unique_urls_test.go
```diff
@@ -91,3 +93,29 @@ func TestNewUniqueURLsWithExceptions(t *testing.T) {
...
+		"https://1.2.3.4:1",
+	}
+	fs := flag.NewFlagSet(name, flag.ExitOnError)
+	u := NewUniqueURLsWithExceptions(strings.Join(urls, ","))
+	fs.Var(u, name, "usage")
+	uss := UniqueURLsFromFlag(fs, name)
+
+	if len(uss) != len(u.Values) {
+		t.Fatal("not unique")
+	}
```
ahrtr, line 112:
```suggestion
	require.Equal(t, len(u.Values), len(uss))
```

pkg/flags/unique_urls_test.go
```diff
@@ -91,3 +93,29 @@ func TestNewUniqueURLsWithExceptions(t *testing.T) {
...
+
+	um := make(map[string]struct{})
+	for _, x := range uss {
+		um[x.String()] = struct{}{}
+	}
+	if !reflect.DeepEqual(um, u.Values) {
+		t.Fatal("not equal")
+	}
```
ahrtr, line 120:
```suggestion
	require.Equal(t, u.Values, um)
```

ahrtr: It's just a minor comment,` require.Equal` is recommended. If you look at the implementation, it actually calls `reflect.DeepEqual`, but it's simpler from user perspective.

gocurr: I understand. 
But if a file has two styles of tests, it's not very comfortable.

ahrtr: then pls change line 90 as well.

gocurr: Done.

ahrtr (approved):
LGTM

thx @gocurr 

Please backport this PR to both 3.5 and  3.4 if needed.

fuweid (approved):
LGTM

Thank you for explanation and patient.

gocurr:
> LGTM
> 
> thx @gocurr
> 
> Please backport this PR to both 3.5 and 3.4 if needed.

For now? Or after this is merged(if accepted)?

ahrtr:
This PR needs to be backported to both 3.5 and 3.4. Please raise an issue to track the effort. Otherwise it's hard to track the task after merging this PR.

ahrtr:
@gocurr Please signoff the commit

ahrtr:
See https://github.com/etcd-io/etcd/pull/16272/checks?check_run_id=15358965509

Screenshot 2023-07-26 at 13 27 58

gocurr:
Maybe it's the https://github.com/etcd-io/etcd/pull/16308, this pull uses wrong branch, and I closed it.
I don't know how to handle this.

gocurr:
Sorry, should it be abandoned? Sorry again.

gocurr:
Sorry, I'm not very familiar with github's workflow yet.
If there are any questions, please let me know.

ahrtr:
Please rebase this PR

pkg/flags/unique_urls_test.go
```diff
@@ -89,5 +93,29 @@ func TestNewUniqueURLsWithExceptions(t *testing.T) {
 		if uv.String() != tests[i].rs {
 			t.Fatalf("#%d: expected %q, got %q", i, tests[i].rs, uv.String())
 		}
+		require.Equal(t, tests[i].exp, uv.Values)
+		require.Equal(t, uv.String(), tests[i].rs)
```
ahrtr, line 97: With the addition of these two lines, you should remove line 89 ~ 95. Or remove these two lines and keep it as it's.

gocurr:
How I wish it can be cherry-picked.
In fact I really enjoy `Gerrit Code Review`.

Issue #16276, opened 2023-07-20
Introduce grpc health check in etcd client


chaochn47:
### Background

> gRPC Health checks are used to probe whether the server is able to handle RPCs. A server may choose to reply “unhealthy” because it is not ready to take requests, it is shutting down or some other reason. The client can act accordingly if the response is not received within some time window or the response says unhealthy in it.

ref. 
1. https://github.com/grpc/grpc/blob/master/doc/health-checking.md
2. https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md

https://github.com/etcd-io/etcd/pull/8121 added basic grpc health service only on server side since etcd v3.3. 
https://github.com/etcd-io/etcd/blob/622017468725647028e75aea2c971966ab70b3b5/server/etcdserver/api/v3rpc/grpc.go#L81-L86

### Problem

In a multi etcd server endpoints scenario, etcd client only fails over to the other endpoint when existing connection/channel is in not in `Ready` state. However, etcd client does not know about if etcd server can handle RPCs. 

For example
1. https://github.com/etcd-io/etcd/issues/9222
2. https://github.com/etcd-io/etcd/issues/14338

It needs a comprehensive design and testing. 

Placeholder google doc [etcd client grpc health check](https://docs.google.com/document/d/1x7guECGQI1wdAh2fBzPBRTva-sqGGLR8Bp0lrrmmDk0/edit#heading=h.euv7j5s250nm) copied from the [KEP template](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template). 

### Why is this needed?

Improve etcd availability by failing over to other healthy etcd endpoints

fuweid:
Hi @chaochn47 , based on your PR #16278, it seems like that the PR marks not_serving status because of defrag. It should be applied to any maintenance APIs, is it correct? My only concern is that based on that https://github.com/grpc/grpc-go/blob/d7f45cdf9ae720256d328fcdb6356ae58122a7f6/health/client.go#L104, it might impact the on-going requests(still in trasfering data) and force to recreate connection. (I haven't verified it yet,  update it later)

chaochn47:
Thanks @fuweid for the review!

> based on your PR https://github.com/etcd-io/etcd/pull/16278, it seems like that the PR marks not_serving status because of defrag. It should be applied to any maintenance APIs, is it correct?

[...]

PR #16278, opened 2023-07-20
gRPC health server sets serving status to NOT_SERVING on defrag

chaochn47:
Trying to fix https://github.com/kubernetes/kubernetes/issues/93280 on etcd side since etcd is part of the kubernetes sigs. 

Before failure rate is around 33%
```
v3_failover_test.go:101: etcd client failed to fail over, error (context deadline exceeded)
v3_failover_test.go:103: request failure rate is 33.59%, traffic volume success 172 requests, total 259 requests
```

After failure rate is 0%
```
v3_failover_test.go:103: request failure rate is 0.00%, traffic volume success 2116 requests, total 2116 requests
```

```
2023-07-20T22:18:20.243Z [[core] [Channel #19 SubChannel #20] Subchannel picks a new address "unix:localhost:m0" to connect]
2023-07-20T22:18:20.243Z [[balancer] base.baseBalancer: handle SubConn state change: 0xc0006aad68, CONNECTING]
2023-07-20T22:18:20.243Z [[core] [Channel #19 SubChannel #20] Subchannel Connectivity change to READY]
2023-07-20T22:18:20.243Z [[balancer] base.baseBalancer: handle SubConn state change: 0xc0006aad68, READY]
2023-07-20T22:18:20.243Z [[roundrobin] roundrobinPicker: Build called with info: {map[0xc0006aad68:{{
          "Addr": "unix:localhost:m0",
          "ServerName": "localhost",
          "Attributes": null,
          "BalancerAttributes": null,
          "Type": 0,
          "Metadata": null
        }} 0xc0006aadf8:{{
          "Addr": "unix:localhost:m2",
          "ServerName": "localhost",
          "Attributes": null,
          "BalancerAttributes": null,
          "Type": 0,
          "Metadata": null
        }}]}]


2023-07-20T22:18:20.743Z INFO	m0	starting defragment	{"member": "m0"}
2023-07-20T22:18:20.743Z [[core] [Channel #19 SubChannel #20] Subchannel Connectivity change to TRANSIENT_FAILURE]
2023-07-20T22:18:20.743Z [[balancer] base.baseBalancer: handle SubConn state change: 0xc0006aad68, TRANSIENT_FAILURE]
2023-07-20T22:18:20.743Z [[roundrobin] roundrobinPicker: Build called with info: {map[0xc0006aadb0:{{
          "Addr": "unix:localhost:m1",
          "ServerName": "localhost",
          "Attributes": null,
          "BalancerAttributes": null,
          "Type": 0,
          "Metadata": null
        }} 0xc0006aadf8:{{
          "Addr": "unix:localhost:m2",
          "ServerName": "localhost",
          "Attributes": null,
          "BalancerAttributes": null,
          "Type": 0,
          "Metadata": null
        }}]}]
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -192,6 +192,9 @@ type ServerConfig struct {
 	// a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
+	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
+	ExperimentalStopGRPCServiceOnDefrag bool `json:"experimental-stop-grpc-service-on-defrag"`
+
 	// ExperimentalBootstrapDefragThresholdMegabytes is the minimum number of megabytes needed to be freed for etcd server to
 	// consider running defrag during bootstrap. Needs to be set to non-zero value to take effect.
 	ExperimentalBootstrapDefragThresholdMegabytes uint `json:"experimental-bootstrap-defrag-threshold-megabytes"`
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -429,6 +429,9 @@ type Config struct {
 	// ExperimentalTxnModeWriteWithSharedBuffer enables write transaction to use a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
+	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
+	ExperimentalStopGRPCServiceOnDefrag bool `json:"experimental-stop-grpc-service-on-defrag"`
+
 	// V2Deprecation describes phase of API & Storage V2 support
 	V2Deprecation config.V2DeprecationEnum `json:"v2-deprecation"`
 }
@@ -529,6 +532,7 @@ func NewConfig() *Config {
 		ExperimentalDowngradeCheckTime:           DefaultDowngradeCheckTime,
 		ExperimentalMemoryMlock:                  false,
 		ExperimentalTxnModeWriteWithSharedBuffer: true,
+		ExperimentalStopGRPCServiceOnDefrag:      false,
 		ExperimentalMaxLearners:                  membership.DefaultMaxLearners,
 
 		ExperimentalCompactHashCheckEnabled: false,
@@ -725,6 +729,7 @@ func (cfg *Config) AddFlags(fs *flag.FlagSet) {
 	fs.DurationVar(&cfg.ExperimentalWarningUnaryRequestDuration, "experimental-warning-unary-request-duration", cfg.ExperimentalWarningUnaryRequestDuration, "Time duration after which a warning is generated if a unary request takes more time. It's deprecated, and will be decommissioned in v3.7. Use --warning-unary-request-duration instead.")
 	fs.BoolVar(&cfg.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
+	fs.BoolVar(&cfg.ExperimentalStopGRPCServiceOnDefrag, "experimental-stop-grpc-service-on-defrag", cfg.ExperimentalStopGRPCServiceOnDefrag, "Enable etcd gRPC service to stop serving client requests on defragmentation.")
 	fs.UintVar(&cfg.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
 	fs.IntVar(&cfg.ExperimentalMaxLearners, "experimental-max-learners", membership.DefaultMaxLearners, "Sets the maximum number of learners that can be available in the cluster membership.")
 	fs.Uint64Var(&cfg.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.SnapshotCatchUpEntries, "Number of entries for a slow follower to catch up after compacting the raft storage entries.")
--- a/server/etcdmain/help.go
+++ b/server/etcdmain/help.go
@@ -305,6 +305,8 @@ Experimental feature:
     Enable to enforce etcd pages (in particular bbolt) to stay in RAM.
   --experimental-snapshot-catchup-entries
     Number of entries for a slow follower to catch up after compacting the raft storage entries.
+  --experimental-stop-grpc-service-on-defrag
+    Enable etcd gRPC service to stop serving client requests on defragmentation.
 
 Unsafe feature:
   --force-new-cluster 'false'
--- a/server/etcdserver/api/v3rpc/health.go
+++ b/server/etcdserver/api/v3rpc/health.go
@@ -1,0 +1,68 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v3rpc
+
+import (
+	"go.uber.org/zap"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+const (
+	allGRPCServices = ""
+)
+
+type HealthNotifier interface {
+	StartServe()
+	StopServe(reason string)
+}
+
+func NewHealthNotifier(hs *health.Server, lg *zap.Logger) HealthNotifier {
+	if hs == nil {
+		panic("unexpected nil gRPC health server")
+	}
+	if lg == nil {
+		lg = zap.NewNop()
+	}
+	hc := &healthChecker{hs: hs, lg: lg}
+	// set grpc health server as serving status blindly since
+	// the grpc server will serve iff s.ReadyNotify() is closed.
+	hc.StartServe()
+	return hc
+}
+
+type healthChecker struct {
+	hs *health.Server
+	lg *zap.Logger
+}
+
+func (hc *healthChecker) StartServe() {
+	hc.lg.Info(
+		"grpc service status changed",
+		zap.String("service", allGRPCServices),
+		zap.String("status", healthpb.HealthCheckResponse_SERVING.String()),
+	)
+	hc.hs.SetServingStatus(allGRPCServices, healthpb.HealthCheckResponse_SERVING)
+}
+
+func (hc *healthChecker) StopServe(reason string) {
+	hc.lg.Warn(
+		"grpc service status changed",
+		zap.String("service", allGRPCServices),
+		zap.String("status", healthpb.HealthCheckResponse_NOT_SERVING.String()),
+		zap.String("reason", reason),
+	)
+	hc.hs.SetServingStatus(allGRPCServices, healthpb.HealthCheckResponse_NOT_SERVING)
+}
--- a/server/etcdserver/api/v3rpc/maintenance.go
+++ b/server/etcdserver/api/v3rpc/maintenance.go
@@ -74,10 +74,13 @@ type maintenanceServer struct {
 	cs     ClusterStatusGetter
 	d      Downgrader
 	vs     serverversion.Server
+	hn     HealthNotifier
+
+	stopServingOnDefrag bool
 }
 
-func NewMaintenanceServer(s *etcdserver.EtcdServer) pb.MaintenanceServer {
-	srv := &maintenanceServer{lg: s.Cfg.Logger, rg: s, hasher: s.KV().HashStorage(), bg: s, a: s, lt: s, hdr: newHeader(s), cs: s, d: s, vs: etcdserver.NewServerVersionAdapter(s)}
+func NewMaintenanceServer(s *etcdserver.EtcdServer, hn HealthNotifier) pb.MaintenanceServer {
+	srv := &maintenanceServer{lg: s.Cfg.Logger, rg: s, hasher: s.KV().HashStorage(), bg: s, a: s, lt: s, hdr: newHeader(s), cs: s, d: s, vs: etcdserver.NewServerVersionAdapter(s), hn: hn, stopServingOnDefrag: s.Cfg.ExperimentalStopGRPCServiceOnDefrag}
 	if srv.lg == nil {
 		srv.lg = zap.NewNop()
 	}
@@ -86,6 +89,10 @@ func NewMaintenanceServer(s *etcdserver.EtcdServer) pb.MaintenanceServer {
 
 func (ms *maintenanceServer) Defragment(ctx context.Context, sr *pb.DefragmentRequest) (*pb.DefragmentResponse, error) {
 	ms.lg.Info("starting defragment")
+	if ms.stopServingOnDefrag {
+		ms.hn.StopServe("defrag is active")
+		defer ms.hn.StartServe()
+	}
 	err := ms.bg.Backend().Defrag()
 	if err != nil {
 		ms.lg.Warn("failed to defragment", zap.Error(err))
--- a/tests/e2e/failover_test.go
+++ b/tests/e2e/failover_test.go
@@ -1,0 +1,163 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+	_ "google.golang.org/grpc/health"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+const (
+	// in sync with how kubernetes uses etcd
+	// https://github.com/kubernetes/kubernetes/blob/release-1.28/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L59-L71
+	keepaliveTime    = 30 * time.Second
+	keepaliveTimeout = 10 * time.Second
+	dialTimeout      = 20 * time.Second
+
+	clientRuntime  = 10 * time.Second
+	requestTimeout = 100 * time.Millisecond
+)
+
+func TestFailoverOnDefrag(t *testing.T) {
+	tcs := []struct {
+		name            string
+		clusterOptions  []e2e.EPClusterOption
+		gRPCDialOptions []grpc.DialOption
+
+		// common assertion
+		expectedMinTotalRequestsCount int
+		// happy case assertion
+		expectedMaxFailedRequestsCount int
+		// negative case assertion
+		expectedMinFailedRequestsCount int
+	}{
+		{
+			name: "defrag failover happy case",
+			clusterOptions: []e2e.EPClusterOption{
+				e2e.WithClusterSize(3),
+				e2e.WithExperimentalStopGRPCServiceOnDefrag(true),
+				e2e.WithGoFailEnabled(true),
+			},
+			gRPCDialOptions: []grpc.DialOption{
+				grpc.WithDisableServiceConfig(),
+				grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy": "round_robin", "healthCheckConfig": {"serviceName": ""}}`),
+			},
+			expectedMinTotalRequestsCount:  300,
+			expectedMaxFailedRequestsCount: 5,
+		},
+		{
+			name: "defrag blocks one-third of requests with stopGRPCServiceOnDefrag set to false",
+			clusterOptions: []e2e.EPClusterOption{
+				e2e.WithClusterSize(3),
+				e2e.WithExperimentalStopGRPCServiceOnDefrag(false),
+				e2e.WithGoFailEnabled(true),
+			},
+			gRPCDialOptions: []grpc.DialOption{
+				grpc.WithDisableServiceConfig(),
+				grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy": "round_robin", "healthCheckConfig": {"serviceName": ""}}`),
+			},
+			expectedMinTotalRequestsCount:  300,
+			expectedMinFailedRequestsCount: 90,
+		},
+		{
+			name: "defrag blocks one-third of requests with stopGRPCServiceOnDefrag set to true and client health check disabled",
+			clusterOptions: []e2e.EPClusterOption{
+				e2e.WithClusterSize(3),
+				e2e.WithExperimentalStopGRPCServiceOnDefrag(true),
+				e2e.WithGoFailEnabled(true),
+			},
+			expectedMinTotalRequestsCount:  300,
+			expectedMinFailedRequestsCount: 90,
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			e2e.BeforeTest(t)
+			clus, cerr := e2e.NewEtcdProcessCluster(context.TODO(), t, tc.clusterOptions...)
+			require.NoError(t, cerr)
+			t.Cleanup(func() { clus.Stop() })
+
+			endpoints := clus.EndpointsGRPC()
+
+			requestVolume, successfulRequestCount := 0, 0
+			g := new(errgroup.Group)
+			g.Go(func() (lastErr error) {
+				clusterClient, cerr := clientv3.New(clientv3.Config{
+					DialTimeout:          dialTimeout,
+					DialKeepAliveTime:    keepaliveTime,
+					DialKeepAliveTimeout: keepaliveTimeout,
+					Endpoints:            endpoints,
+					DialOptions:          tc.gRPCDialOptions,
+				})
+				if cerr != nil {
+					return cerr
+				}
+				defer clusterClient.Close()
+
+				timeout := time.After(clientRuntime)
+				for {
+					select {
+					case <-timeout:
+						return lastErr
+					default:
+					}
+					getContext, cancel := context.WithTimeout(context.Background(), requestTimeout)
+					_, err := clusterClient.Get(getContext, "health")
+					cancel()
+					requestVolume++
+					if err != nil {
+						lastErr = err
+						continue
+					}
+					successfulRequestCount++
+				}
+			})
+			triggerDefrag(t, clus.Procs[0])
+
+			err := g.Wait()
+			if err != nil {
+				t.Logf("etcd client failed to fail over, error (%v)", err)
+			}
+			t.Logf("request failure rate is %.2f%%, traffic volume successfulRequestCount %d requests, total %d requests", (1-float64(successfulRequestCount)/float64(requestVolume))*100, successfulRequestCount, requestVolume)
+
+			require.GreaterOrEqual(t, requestVolume, tc.expectedMinTotalRequestsCount)
+			failedRequestCount := requestVolume - successfulRequestCount
+			if tc.expectedMaxFailedRequestsCount != 0 {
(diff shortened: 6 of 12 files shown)


chaochn47:
Hi @ahrtr @serathius @jmhbnz @fuweid, could you please take a look this approach and share your thoughts about it? 

> 1. making defrag not blocking quorum reads
> 2. making etcd client's load balancer switch to a follower if the leader is doing defrag. (I believe today we only have round robin LB, am I correct?)

I was inspired by what https://github.com/kubernetes/kubernetes/issues/93280#issuecomment-664013380 proposed in the k8s issue. 

It will be helpful for eliminating online defrag impact. 

/cc @lavacat @dims

server/etcdserver/api/v3rpc/maintenance.go
```diff
@@ -87,6 +88,8 @@ func NewMaintenanceServer(s *etcdserver.EtcdServer) pb.MaintenanceServer {
 
 func (ms *maintenanceServer) Defragment(ctx context.Context, sr *pb.DefragmentRequest) (*pb.DefragmentResponse, error) {
 	ms.lg.Info("starting defragment")
+	ms.hn.StopServe("defrag is active")
+	defer ms.hn.StartServe()
```
chaochn47, line 92: There are chances that with a write-intensive workload, it will take some time for the local node to apply the backlogged in-flights entries. 

But this PR should make etcd availability much better when defrag is active.

lavacat:
> Please see [#16276 (comment)](https://github.com/etcd-io/etcd/issues/16276#issuecomment-1651845511)
@ahrtr are you referring to Livez/Readyz, 3.7 defrag plans or experimental dependency? 
 - Livez/Readyz: It will be a larger change and imo will build on this implementation. Also, current [design](https://docs.google.com/document/d/1PaUAp76j1X92h3jZF47m32oVlR8Y-p-arB5XOB7Nb6U/edit) doesn't address client side traffic failover during defrag.
 - 3.7 defrag: not clear if it will be back ported to 3.5 and will take longer time to implement.
 -  experimental dependency: I don't see this PR adding new dependencies. It does add `"healthCheckConfig": {"serviceName": ""}}` deepening current dependency (adds new [stream](https://github.com/grpc/grpc-go/blob/v1.58.2/health/client.go#L59C10-L59C10) behind the scenes).

IMO this PR is a bugfix for an existing problem with defrag. Fundamentally if we don't want to use https://pkg.go.dev/google.golang.org/grpc/health, we shouldn't merge this,  otherwise I think it's very useful fix.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,131 @@
...
+					close(errc)
+				}()
+				cc, cerr := clientv3.New(clientv3.Config{
+					DialTimeout:          dialTimeout,
+					DialKeepAliveTime:    keepaliveTime,
+					DialKeepAliveTimeout: keepaliveTimeout,
+					Endpoints:            endpoints,
+					DialOptions: []grpc.DialOption{
+						grpc.WithDisableServiceConfig(),
+						grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy": "round_robin", "healthCheckConfig": {"serviceName": ""}}`),
```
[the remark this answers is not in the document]
chaochn47: etcd client flag can be added separately as a follow up. 

I am trying to limit the required changes so
1. we could back port the change in 3.5 internally first
2. apiserver just need to add a patch without waiting for the etcd 3.5 release

fuweid:
@chaochn47 For the change, it looks good to me. It's helpful to the kube-apiserver which has multiple endpoints instead of one service. The kube-apiserver can forward the requests to available servers.  What do you think about introducing the flag to enable this feature? It might bring heavy traffic to other servers.

chaochn47:
Thanks @fuweid for the feedback! 

> What do you think about introducing the flag to enable this feature? It might bring heavy traffic to other servers.
Yeah, the heuristic is defrag won't take too long and once it's completed. New requests will be routed to this server and watch will break and has 1/3 chance to establish on this server. 

There is no built-in supported capacity aware load balancer in gRPC-go. I will explore and explain it in the [design doc](https://docs.google.com/document/d/1x7guECGQI1wdAh2fBzPBRTva-sqGGLR8Bp0lrrmmDk0/edit#heading=h.euv7j5s250nm) after livez/readyz.

lavacat (approved):

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,60 @@
...
+type healthChecker struct {
+	hs *health.Server
+	lg *zap.Logger
+}
+
+func (hc *healthChecker) StartServe() {
+	if hc.hs != nil {
+		if hc.lg != nil {
+			hc.lg.Info("start serving gRPC requests from client")
+		}
```
ahrtr, line 45: Let's ensure `lg` will never be `nil`. Same below.
```suggestion
		hc.lg.Info("start serving gRPC requests from client")
```

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,60 @@
...
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+type HealthNotifier interface {
+	StartServe()
+	StopServe(reason string)
+}
+
+func NewHealthNotifier(hs *health.Server, lg *zap.Logger) HealthNotifier {
+	hc := &healthChecker{hs: hs, lg: lg}
```
ahrtr, line 29:
```suggestion
        if lg == nil {
		lg = zap.NewNop()
	}
	hc := &healthChecker{hs: hs, lg: lg}
```

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+			t.Logf("request failure rate is %.2f%%, traffic volume success %d requests, total %d requests", (1-float64(success)/float64(cnt))*100, success, cnt)
+			// expect no more than 5 failed requests
+			require.InDelta(t, cnt, success, failedRequests)
+		})
+	}
+}
+
+func triggerDefrag(t *testing.T, clus *e2e.EtcdProcessCluster) {
+	err := clus.Procs[0].Failpoints().Setup(context.Background(), "defragBeforeCopy", `sleep(8000)`)
```
ahrtr, line 127:
```suggestion
	err := clus.Procs[0].Failpoints().Setup(context.Background(), "defragBeforeCopy", `sleep("8s")`)
```

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,131 @@
...
+		name            string
+		clusterOptions  []e2e.EPClusterOption
+		failureInjector func(t *testing.T, clus *e2e.EtcdProcessCluster)
+	}{
+		{
+			name:            "defrag",
+			clusterOptions:  []e2e.EPClusterOption{e2e.WithClusterSize(3), e2e.WithGoFailEnabled(true)},
+			failureInjector: triggerDefrag,
+		},
+	}
```
ahrtr, line 56: only one case in the table? Are you going to add more in followup PR?

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,131 @@
...
+					default:
+					}
+					cctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+					_, err := cc.Get(cctx, "health")
+					cancel()
+					cnt++
+					if err != nil {
+						lastErr = err
+						continue
+					}
```
ahrtr, line 107: In exactly what cases, the request may fail?

ahrtr:
Please also rebase the PR.

tests/e2e/failover_test.go
chaochn47: Potentially, I would imagine gRPC health server will re-use the readyz checks.

tests/e2e/failover_test.go
chaochn47: > Because of the inherently asynchronous nature of the network, whenever a backend transitions from healthy to unhealthy, it may still receive some number of RPCs that were already in flight from the client before the client received the notification that the backend is unhealthy. This race condition lasts approximately the network round trip time (it includes the one-way trip time for RPCs that were already sent by the client but not yet received by the server when the server goes unhealthy, plus the one-way trip time for RPCs sent by the client between when the server sends the unhealthy notification and when the client receives it).

https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md#race-conditions-and-subchannel-startup

RTT should be around 10ms if client and server are deployed in different host but in the same region.

chaochn47:
> Please also rebase the PR.

Done. 

---

IMHO, the PR itself is good to merge and safe to cherry picked to release branch since it's recommended to do defrag against one member at a time. 

One of the remaining risks to evaluate before expanding this to other readyz checks is client should fail open and at least keep one connection in the picker if non of the etcd health servers is SERVING.

tests/e2e/failover_test.go
ahrtr: Then pls leave a TODO comment here, thx

ahrtr:
I see that the e2e test case is very similar to the integration test case, is it possible to write a common test case?

chaochn47:
> I see that the e2e test case is very similar to the integration test case, is it possible to write a common test case?

Enabling failpoint both in e2e and integration test has not yet built into the common test framework. This can be a separate PR to follow up. 

> Another big question is should we add an experimental flag something like "--experimental-stop-grpc-service-on-defragmentation"? I think the answer should be YES. Any new feature might also bring new problem, at least we add an option for users to disable it. 

Okay.

chaochn47:
Ping @ahrtr

server/config/config.go
```diff
@@ -192,6 +192,9 @@ type ServerConfig struct {
 	// a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
+	// ExperimentalStopGRPCServiceOnDefrag enable
```
ahrtr, line 195:
```suggestion
	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
```

server/embed/config.go
```diff
@@ -429,6 +429,9 @@ type Config struct {
 	// ExperimentalTxnModeWriteWithSharedBuffer enables write transaction to use a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
+	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC health server sets serving status to NOT_SERVING on defragmentation.
```
ahrtr, line 432:
```suggestion
	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
```

server/embed/config.go
```diff
@@ -725,6 +729,7 @@ func (cfg *Config) AddFlags(fs *flag.FlagSet) {
 	fs.DurationVar(&cfg.ExperimentalWarningUnaryRequestDuration, "experimental-warning-unary-request-duration", cfg.ExperimentalWarningUnaryRequestDuration, "Time duration after which a warning is generated if a unary request takes more time. It's deprecated, and will be decommissioned in v3.7. Use --warning-unary-request-duration instead.")
 	fs.BoolVar(&cfg.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
+	fs.BoolVar(&cfg.ExperimentalStopGRPCServiceOnDefrag, "experimental-stop-grpc-service-on-defrag", cfg.ExperimentalStopGRPCServiceOnDefrag, "Enable etcd gRPC health server sets serving status to NOT_SERVING on defragmentation.")
```
ahrtr, line 732:
```suggestion
	fs.BoolVar(&cfg.ExperimentalStopGRPCServiceOnDefrag, "experimental-stop-grpc-service-on-defrag", cfg.ExperimentalStopGRPCServiceOnDefrag, "Enable etcd gRPC service to stop serving client requests on defragmentation.")
```

server/etcdmain/help.go
```diff
@@ -305,6 +305,8 @@ Experimental feature:
     Enable to enforce etcd pages (in particular bbolt) to stay in RAM.
   --experimental-snapshot-catchup-entries
     Number of entries for a slow follower to catch up after compacting the raft storage entries.
+  --experimental-stop-grpc-service-on-defrag
+    Enable etcd gRPC health server sets serving status to NOT_SERVING on defragmentation.
```
ahrtr, line 309:
```suggestion
    Enable etcd gRPC service to stop serving client requests on defragmentation.
```

tests/integration/v3_failover_test.go
```diff
@@ -172,3 +185,67 @@ func shouldRetry(err error) bool {
 	}
 	return false
 }
+
+func TestFailoverOnDefrag(t *testing.T) {
+	integration2.BeforeTest(t, integration2.WithFailpoint("defragBeforeCopy", `sleep(10000)`))
```
ahrtr, line 190:
```suggestion
	integration2.BeforeTest(t, integration2.WithFailpoint("defragBeforeCopy", `sleep(10s)`))
```

server/config/config.go
```diff
@@ -192,6 +192,9 @@ type ServerConfig struct {
 	// a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
+	// ExperimentalStopGRPCServiceOnDefrag enable enables etcd gRPC service to stop serving client requests on defragmentation.
```
ahrtr, line 195:
```suggestion
	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
```

ahrtr (approved):
LGTM

Good job! thx @chaochn47

chaochn47:
Ping @serathius

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,59 @@
...
+	return hc
+}
+
+type healthChecker struct {
+	hs *health.Server
+	lg *zap.Logger
+}
+
+func (hc *healthChecker) StartServe() {
+	if hc.hs != nil {
```
serathius, line 45: Why would hc.hs be ever nil? Let's avoid handling cases that should not happen during correct code execution.

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,59 @@
...
+		hc.lg.Info("start serving gRPC requests from client")
+		// server should register all the services manually
+		// use empty service name for all etcd services' health status,
+		// see https://github.com/grpc/grpc/blob/master/doc/health-checking.md for more
+		hc.hs.SetServingStatus("", healthpb.HealthCheckResponse_SERVING)
```
serathius, line 50: Can you obfuscate the value behind "". For example create const called `allGRPCServices` and use it here.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,134 @@
...
+				for {
+					select {
+					case <-timeout:
+						return
+					default:
+					}
+					cctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+					_, err := cc.Get(cctx, "health")
+					cancel()
+					cnt++
```
serathius, line 108: Please don't use short names.  Especially if the variable is long lived.

tests/integration/v3_failover_test.go
```diff
@@ -29,6 +30,18 @@ import (
...
+const (
+	// in sync with how kubernetes uses etcd
+	// https://github.com/kubernetes/kubernetes/blob/release-1.28/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L59-L71
+	keepaliveTime    = 30 * time.Second
+	keepaliveTimeout = 10 * time.Second
+	dialTimeout      = 20 * time.Second
+
+	clientRuntime = 10 * time.Second
+	// expect no more than 5 failed requests
+	failedRequests = 5
```
serathius, line 42: Please avoid using comments to suplement the name. Just call it `expectedMaxFailedRequests`.

tests/integration/v3_failover_test.go
```diff
@@ -172,3 +185,67 @@ func shouldRetry(err error) bool {
...
+		var lastErr error
+		var cc *clientv3.Client
+		defer func() {
+			if cc != nil {
+				cc.Close()
+			}
+			errc <- lastErr
+			close(donec)
+			close(errc)
+		}()
```
serathius, line 209: Please create a separate defer for each variable as they are used.

tests/integration/v3_failover_test.go
```diff
@@ -172,3 +185,67 @@ func shouldRetry(err error) bool {
...
 }
+
+func TestFailoverOnDefrag(t *testing.T) {
+	integration2.BeforeTest(t, integration2.WithFailpoint("defragBeforeCopy", `sleep("10s")`))
+	clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 3, ExperimentalStopGRPCServiceOnDefrag: true})
+	defer clus.Terminate(t)
+	endpoints := clus.Endpoints()
+
+	cnt, success := 0, 0
+	donec := make(chan struct{})
```
serathius, line 196: Please use `sync.WorkGroup`

tests/integration/v3_failover_test.go
```diff
@@ -29,6 +30,18 @@ import (
 	clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"
 )
 
+const (
```
serathius, line 33: No need to implement both integration and e2e tests. E2e tests is enough.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,134 @@
...
+	failedRequests = 5
+)
+
+func TestFailover(t *testing.T) {
+	tcs := []struct {
+		name           string
+		clusterOptions []e2e.EPClusterOption
+		injectFailure  func(t *testing.T, member e2e.EtcdProcess)
+	}{
+		{
```
serathius, line 51: Recommend always mirroring a test. Testing only feature enabled can be really fragile, we cannot distinquish whether reason for test passing is the new feature or there was some external change.

Recommend to add a test case with `WithExperimentalStopGRPCServiceOnDefrag(false)` that is expected to fail requests. Instead of checking maximum number of failed requests, there you can check the minimum number of failed requests.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+			endpoints := clus.EndpointsGRPC()
+
+			cnt, success := 0, 0
+			donec := make(chan struct{})
+			errc := make(chan error, 1)
+
+			go func() {
+				var lastErr error
+				var cc *clientv3.Client
```
serathius, line 77: Pre-defining a `cc` before it's instantiated and then checking for nil is not very readable.

Please create a separate defer for each variable as they are used.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,134 @@
...
+const (
+	// in sync with how kubernetes uses etcd
+	// https://github.com/kubernetes/kubernetes/blob/release-1.28/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L59-L71
+	keepaliveTime    = 30 * time.Second
+	keepaliveTimeout = 10 * time.Second
+	dialTimeout      = 20 * time.Second
+
+	clientRuntime = 10 * time.Second
+	// expect no more than 5 failed requests
+	failedRequests = 5
```
serathius, line 42: Please avoid using comments to suplement the name. Just call it `expectedMaxFailedRequests`.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,134 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			e2e.BeforeTest(t)
+			clus, cerr := e2e.NewEtcdProcessCluster(context.TODO(), t, tc.clusterOptions...)
+			require.NoError(t, cerr)
+			t.Cleanup(func() { clus.Stop() })
+
+			endpoints := clus.EndpointsGRPC()
+
+			cnt, success := 0, 0
+			donec := make(chan struct{})
```
serathius, line 72: Please use `sync.WaitGroup` to ensure goroutine exits.

serathius: Or even better `errgroup.Group` as it allows to return an error which would remove the need for `errc`.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,134 @@
...
+			tc.injectFailure(t, clus.Procs[0])
+
+			<-donec
+			err, ok := <-errc
+			if ok && err != nil {
+				t.Logf("etcd client failed to fail over, error (%v)", err)
+			}
+			t.Logf("request failure rate is %.2f%%, traffic volume success %d requests, total %d requests", (1-float64(success)/float64(cnt))*100, success, cnt)
+			// expect no more than 5 failed requests
+			require.InDelta(t, cnt, success, failedRequests)
```
serathius, line 126: Note, you should also check minimal number of requests executed. Environment could allow for only 5 requests passing. To make sure the test is working as intended we should require at least 20 or more requests.

server/etcdserver/api/v3rpc/health.go
chaochn47: okay. Switched to assert with panic result in that case.

server/etcdserver/api/v3rpc/health.go
chaochn47: Done

tests/e2e/failover_test.go
chaochn47: Done

tests/integration/v3_failover_test.go
chaochn47: Removed integration test.

tests/e2e/failover_test.go
chaochn47: Implemented negative tests.

tests/e2e/failover_test.go
chaochn47: Done

tests/e2e/failover_test.go
chaochn47: Done

tests/e2e/failover_test.go
chaochn47: Done

tests/e2e/failover_test.go
chaochn47: Done

chaochn47:
Thanks @serathius for the review. I have added aggressive test threshold for happy and negative test cases.

chaochn47:
E2E test failed in 386 arch. Reason is `gofail` is not enabled in CI. Test should be skipped after rebasing on top of the changes https://github.com/etcd-io/etcd/pull/16698 introduces.

```
2023-10-18T19:24:18.1851440Z     failover_test.go:145: request failure rate is 0.00%, traffic volume successfulRequestCount 15786 requests, total 15786 requests
2023-10-18T19:24:18.1852640Z     failover_test.go:169: 
2023-10-18T19:24:18.1853387Z         	Error Trace:	/__w/etcd/etcd/tests/e2e/failover_test.go:169
2023-10-18T19:24:18.1854532Z         	            				/__w/etcd/etcd/tests/e2e/failover_test.go:146
2023-10-18T19:24:18.1855364Z         	Error:      	"0" is not greater than or equal to "90"
2023-10-18T19:24:18.1857108Z         	Test:       	TestFailoverOnDefrag/defrag_blocks_one-third_of_requests_with_stopGRPCServiceOnDefrag_set_to_true_and_client_health_check_disabled
```

ref. https://github.com/etcd-io/etcd/actions/runs/6565503688/job/17834165379?pr=16278

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,65 @@
...
+	return hc
+}
+
+type healthChecker struct {
+	hs *health.Server
+	lg *zap.Logger
+}
+
+func (hc *healthChecker) StartServe() {
+	if hc.hs == nil {
```
serathius, line 49: If you are worried about hc.hs being nil, why not check in the `NewHealthNotifier` ?

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,170 @@
...
+	expectedMinFailedRequestsCount = 90
+)
+
+func TestFailoverOnDefrag(t *testing.T) {
+	tcs := []struct {
+		name           string
+		clusterOptions []e2e.EPClusterOption
+
+		disableClientHealthCheck bool
+		assertion                func(t *testing.T, requestVolume, successfulRequestCount int)
```
serathius, line 58: Please avoid a dedicated functions for assertions, instead have two variables, `expectedMinTotalRequestsCount`, `expectedMaxFailedRequestsCount` set on tcs and inline the assetions into TestFailoverOnDefrag function

```
	if tc.expectedMinTotalRequestsCount {
		require.GreaterOrEqual(t, requestVolume, tc.expectedMinTotalRequestsCount)
	}
	if tc.expectedMaxFailedRequestsCount {
		failedRequestCount := requestVolume - successfulRequestCount
		require.LessOrEqual(t, failedRequestCount, expectedMaxFailedRequestsCount)
	}
```

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,170 @@
...
+			requestVolume, successfulRequestCount := 0, 0
+			g := new(errgroup.Group)
+			g.Go(func() (lastErr error) {
+				dialOptions := make([]grpc.DialOption, 0)
+				if !tc.disableClientHealthCheck {
+					dialOptions = []grpc.DialOption{
+						grpc.WithDisableServiceConfig(),
+						grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy": "round_robin", "healthCheckConfig": {"serviceName": ""}}`),
+					}
+				}
```
serathius, line 108: Can you make `grpcDialOptions` a field on `tc`. Having it directly on test case will make it more readable when reading the test cases vs some custom variable that needs to be looked up.

chaochn47: Make sense to me. Done.

tests/e2e/failover_test.go
chaochn47: Done!

server/etcdserver/api/v3rpc/health.go
chaochn47: Good idea :p Done

chaochn47:
Kindly ping @serathius @wenjiaswe @ptabor @mitake @spzala @jmhbnz ~

chaochn47:
Hi @ahrtr @fuweid 

What's the suggested process to move progress on this PR? I think all the review comments are addressed.

ahrtr (approved):

ahrtr:
This is a nice to have feature, it can prevent client from being blocked on the member in progress of defragmentation.

We also have flag `--experimental-stop-grpc-service-on-defrag` (defaults to `false`). There is no any impact on users if they don't intentionally enable it. So it should be safe.

Leave to other maintainers to take a second look. 

Thanks @chaochn47

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,62 @@
...
+	if hs == nil {
+		panic("unexpected nil gRPC health server")
+	}
+	if lg == nil {
+		lg = zap.NewNop()
+	}
+	hc := &healthChecker{hs: hs, lg: lg}
+	// set grpc health server as serving status blindly since
+	// the grpc server will serve iff s.ReadyNotify() is closed.
+	hc.StartServe()
```
serathius, line 42: Nit: Calling `StartServe()` here means that etcd will emit `"start serving gRPC requests from client"` log which is incorrect. `NewHealthNotifier` doesn't control whether grpc is serving, this is done in different place. Health notifier only changes the serving status. I would change the log line to 
```
hc.lg.Info("grpc service status changed", zap.String("service", allGRPCServices), zap.String("status", "serving")) 
```

serathius (approved):

server/etcdserver/api/v3rpc/health.go
chaochn47: Makes sense. Updated the logging.

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,62 @@
...
+
+type healthChecker struct {
+	hs *health.Server
+	lg *zap.Logger
+}
+
+func (hc *healthChecker) StartServe() {
+	hc.lg.Info("start serving gRPC requests from client")
+	// server should register all the services manually
+	// use empty service name for all etcd services' health status,
```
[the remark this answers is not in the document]
chaochn47: Duplicated comments are removed now.

tests/e2e/failover_test.go
```diff
@@ -0,0 +1,163 @@
...
+		name            string
+		clusterOptions  []e2e.EPClusterOption
+		gRPCDialOptions []grpc.DialOption
+
+		// common assertion
+		expectedMinTotalRequestsCount int
+		// happy case assertion
+		expectedMaxFailedRequestsCount int
+		// negative case assertion
+		expectedMinFailedRequestsCount int
```
serathius, line 56: Idea for followup, better to assert on rate then on request count. Minimal request count is good, but what happens if one machines serves 300 requests, while other 3000? 90 failures out of 300 is 30% while only 3 % on 3000.

server/etcdserver/api/v3rpc/health.go
```diff
@@ -0,0 +1,68 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+const (
+	allGRPCServices = ""
+)
+
+type HealthNotifier interface {
```
serathius, line 27: Idea for followup. Instead of allowing defrag to change StartServe status as they will, better would be to just notify that defrag is happening and let health notify decide on health status. https://en.wikipedia.org/wiki/Law_of_Demeter, why defrag should know how about changing health.

serathius: Additional improvement would be that Maintenance server would no longer need to be aware of `stopServingOnDefrag` flag, instead configuration would be handled by health.

serathius:
I think this is good enough to merge, still please consider implementing followups.

server/etcdserver/api/v3rpc/health.go
chaochn47: Thanks! I like the idea of decoupling stop serving from the maintenance server. 

Eventually the health notifier may extend to other readyz checks (it need more in-depth thought though) implemented in the readyz handler.

tests/e2e/failover_test.go
chaochn47: Gotcha, something like the robustness TPS assertion.

PR #16280, opened 2023-07-21
Backport fix to e2e release version identifcation

jmhbnz:
Currently `release-3.5` e2e tests for `amd64` are relying on a hardcoded version, currently `v.3.4.7`. Refer: https://github.com/etcd-io/etcd/actions/runs/5603021362/job/15178839191#step:6:932 

```bash
'release' started at Wed Jul 19 19:00:50 UTC 2023
Downloading etcd-v3.4.7-linux-amd64.tar.gz
```

I believe this hard coded version was put in place because by default the github `checkout` action does a shallow clone, meaning our reliance on checking for local repository tags fails.

Recently we fixed this for `main` in  https://github.com/etcd-io/etcd/pull/16230 by instead updating `test.sh` to check the list of remote repository tags.

This looks to be working well after merge so I would like to suggest we backport it to `release-3.5`.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -25,7 +25,7 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-amd64-e2e)
-            PASSES='build release e2e' MANUAL_VER=v3.4.7 CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./test.sh
+            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./test.sh
             ;;
           linux-386-e2e)
             GOARCH=386 PASSES='build e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./test.sh
--- a/test.sh
+++ b/test.sh
@@ -635,8 +635,21 @@ function dep_pass {
 
 function release_pass {
   rm -f ./bin/etcd-last-release
-  # to grab latest patch release; bump this up for every minor release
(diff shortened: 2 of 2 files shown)


.github/workflows/e2e.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
         echo "${TARGET}"
         case "${TARGET}" in
           linux-amd64-e2e)
-            PASSES='build release e2e' MANUAL_VER=v3.4.7 CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./test.sh
+            PASSES='build release e2e' CPU='4' EXPECT_DEBUG='true' COVER='false' RACE='true' ./test.sh
```
ahrtr, line 28: Good catch!

test.sh
```diff
@@ -635,8 +635,21 @@ function dep_pass {
 
 function release_pass {
   rm -f ./bin/etcd-last-release
-  # to grab latest patch release; bump this up for every minor release
-  UPGRADE_VER=$(git tag -l --sort=-version:refname "v3.4.*" | head -1)
```
ahrtr, line 639: For the stable release (3.5 in this case), hard code isn't a problem. Its previous version must be 3.4.x.

jmhbnz: Hmm I do think it will be beneficial if our 3.5 release tests will run against the latest patch of `3.4.x`.  That way we get assurance our backports to 3.4 have not introduced an issue with the upgrade process.

ahrtr (approved):

fuweid (approved):
LGTM

serathius (approved):

PR #16310, opened 2023-07-26
tests/robustness: enhance compact failpoint

fuweid:
If the cluster serves requests slowly, the database has few revision number and then Compact won't trigger BatchCommit. Add a loop to check the last revision is big enough to trigger panic.

REF: https://github.com/etcd-io/etcd/pull/14691#issuecomment-1649473554

### DIFF
--- a/tests/robustness/failpoints.go
+++ b/tests/robustness/failpoints.go
@@ -54,8 +54,8 @@ var (
 	CompactAfterCommitScheduledCompactPanic  Failpoint = goPanicFailpoint{"compactAfterCommitScheduledCompact", triggerCompact{}, AnyMember}
 	CompactBeforeSetFinishedCompactPanic     Failpoint = goPanicFailpoint{"compactBeforeSetFinishedCompact", triggerCompact{}, AnyMember}
 	CompactAfterSetFinishedCompactPanic      Failpoint = goPanicFailpoint{"compactAfterSetFinishedCompact", triggerCompact{}, AnyMember}
-	CompactBeforeCommitBatchPanic            Failpoint = goPanicFailpoint{"compactBeforeCommitBatch", triggerCompact{}, AnyMember}
-	CompactAfterCommitBatchPanic             Failpoint = goPanicFailpoint{"compactAfterCommitBatch", triggerCompact{}, AnyMember}
+	CompactBeforeCommitBatchPanic            Failpoint = goPanicFailpoint{"compactBeforeCommitBatch", triggerCompact{multiBatchCompaction: true}, AnyMember}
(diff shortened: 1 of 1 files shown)


tests/robustness/failpoints.go
```diff
@@ -103,6 +103,18 @@ func validateFailpoint(clus *e2e.EtcdProcessCluster, failpoint Failpoint) error
 	return nil
 }
 
+func addMinimalRevForCompactFailpoint(failpoint Failpoint, batchNum int) Failpoint {
+	switch failpoint.Name() {
+	case "compactAfterCommitBatch", "compactBeforeCommitBatch":
```
serathius, line 108: Instead of using this decorator and doing name matching, can you add a flag?
```
var (
    CompactBeforeCommitBatchPanic            Failpoint = goPanicFailpoint{"compactBeforeCommitBatch", triggerCompact{waitForRevision: true}, AnyMember}
	CompactAfterCommitBatchPanic             Failpoint = goPanicFailpoint{"compactAfterCommitBatch", triggerCompact{waitForRevision: true}, AnyMember}
)

type triggerCompact struct {
 waitForRevision bool
}

	
func (t triggerCompact) Trigger(_ *testing.T, ctx context.Context, member e2e.EtcdProcess, _ *e2e.EtcdProcessCluster) error {
		Endpoints:            member.EndpointsGRPC(),
		DialKeepAliveTime:    10 * time.Second,
		DialKeepAliveTimeout: 100 * time.Millisecond,
	})
	if err != nil {
		return fmt.Errorf("failed creating client: %w", err)
	}
	defer cc.Close()
    if t.waitForRevision {
      	var rev int64
	  for {
		resp, err := cc.Get(ctx, "/")
		if err != nil {
			return err
		}
		if rev = resp.Header.Revision > member.Cfg.CompactionBatchLimit {
		    break
		}
		time.Sleep(50 * time.Millisecond)
	  }
    }

	_, err = cc.Compact(ctx, rev)
	if err != nil && !strings.Contains(err.Error(), "error reading from server: EOF") {
		return err
	}
	return nil
}
```

serathius:
Idea is good, just need to improve implementation.

serathius (approved):

tests/robustness/failpoints.go
```diff
@@ -338,9 +338,11 @@ func (t triggerDefrag) Available(e2e.EtcdProcessClusterConfig, e2e.EtcdProcess)
 	return true
 }
 
-type triggerCompact struct{}
+type triggerCompact struct {
+	waitForRevision bool
```
serathius, line 342: `waitForRevision` was a working name. Something like `multiBatchCompaction` would be better as we are basically waiting for revision to be big enough so compaction needs to be done in multiple batches.

fuweid: `multiBatchCompaction` looks better and more accurate. let me update it.

PR #16248, opened 2023-07-15
Replace unnecessary Lock()/Unlock()s with RLock()/RUnlock()s

CaojiamingAlan:
Follow up https://github.com/etcd-io/etcd/pull/16068#discussion_r1263664700

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #16312, opened 2023-07-26
server: Remove Lock/Unlock from ReadTx 

serathius:
Simple is better. Why have `Lock` on read transaction, why have `Rlock` on write transaction. It's bad Go practice to share interface between objects that interpret it differently.  `ReadTxn` and `BatchTxn` serve different purpose with different side effect, they should not share the same interface.

### DIFF
--- a/server/storage/backend/batch_tx.go
+++ b/server/storage/backend/batch_tx.go
@@ -45,6 +45,8 @@ type Bucket interface {
 
 type BatchTx interface {
 	ReadTx
+	Lock()
+	Unlock()
 	UnsafeCreateBucket(bucket Bucket)
 	UnsafeDeleteBucket(bucket Bucket)
 	UnsafePut(bucket Bucket, key []byte, value []byte)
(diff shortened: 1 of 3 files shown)


ahrtr (approved):
LGTM

- `readTx` implements `Lock()/Unlock()`, but the methods are not defined in the interface `ReadTx`; It can prevent similar wrong lock in future. Nice!
- This PR duplicates most of the change in https://github.com/etcd-io/etcd/pull/16248. The better way would be raising a comment in that PR to honor the contributor's work.

serathius:
> This PR duplicates most of the change in https://github.com/etcd-io/etcd/pull/16248. The better way would be raising a comment in that PR to honor the contributor's work.

I started from interface changes, so didn't know it duplicated PR. Let's merge the work.

serathius:
Rewrite to be even more stricter. Rlock is no longer available on BatchTx. 

Please take another look @ahrtr

ahrtr:
> Rewrite to be even more stricter. Rlock is no longer available on BatchTx.
> 
> Please take another look @ahrtr

I suggest to get the first commit included in a separate PR, and we can merge it right away because it's a simple & safe change, and it has obvious benefit. 

Afterwards, we can evaluate your second commit separately carefully.

serathius:
Ok, let me merge the first one.

ahrtr (approved):

PR #16313, opened 2023-07-26
checks the client supported versions based on the current version

### LINKED ISSUE #16273: Programmatically check the supported versions

Follow up https://github.com/etcd-io/etcd/pull/16270#discussion_r1267770143

Introduce a helper function in clientV3 that gets and checks the supported versions based on the [current version](https://github.com/etcd-io/etcd/blob/92de641a228d5be32177bf7eb9d3711f9453ef36/api/version/version.go#L29), and replace the [hardcoded](https://github.com/etcd-io/etcd/blob/92de641a228d5be32177bf7eb9d3711f9453ef36/client/v3/client.go#L512) version check. 
Only the current version and one previous version are supported.
i.e. Currently for v3.5 we support v3.4 and v3.5. When v3.6.0 is formally released, then v3.4 will be out of support.

### Why is this needed?

Improve codebase quality

CaojiamingAlan:
Fix #16273

### DIFF
--- a/api/version/version.go
+++ b/api/version/version.go
@@ -44,6 +44,9 @@ var (
 	V3_6 = semver.Version{Major: 3, Minor: 6}
 	V3_7 = semver.Version{Major: 3, Minor: 7}
 	V4_0 = semver.Version{Major: 4, Minor: 0}
+
+	// AllVersions keeps all the versions in ascending order.
+	AllVersions = []semver.Version{V3_0, V3_1, V3_2, V3_3, V3_4, V3_5, V3_6, V3_7, V4_0}
 )
 
 func init() {
--- a/client/v3/client.go
+++ b/client/v3/client.go
@@ -18,11 +18,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -32,6 +32,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	"go.etcd.io/etcd/api/v3/version"
 	"go.etcd.io/etcd/client/pkg/v3/logutil"
 	"go.etcd.io/etcd/client/v3/credentials"
 	"go.etcd.io/etcd/client/v3/internal/endpoint"
@@ -475,6 +476,22 @@ func (c *Client) roundRobinQuorumBackoff(waitBetween time.Duration, jitterFracti
 	}
 }
 
+// minSupportedVersion returns the minimum version supported, which is the previous minor release.
+func minSupportedVersion() *semver.Version {
+	ver := semver.Must(semver.NewVersion(version.Version))
+	// consider only major and minor version
+	ver = &semver.Version{Major: ver.Major, Minor: ver.Minor}
+	for i := range version.AllVersions {
+		if version.AllVersions[i].Equal(*ver) {
+			if i == 0 {
+				return ver
+			}
+			return &version.AllVersions[i-1]
+		}
+	}
+	panic("current version is not in the version list")
+}
+
 func (c *Client) checkVersion() (err error) {
 	var wg sync.WaitGroup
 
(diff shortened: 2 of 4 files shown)


serathius (approved):

client/v3/client.go
```diff
@@ -475,6 +476,14 @@ func (c *Client) roundRobinQuorumBackoff(waitBetween time.Duration, jitterFracti
 	}
 }
 
+func minSupportedVersion() *semver.Version {
+	ver := semver.Must(semver.NewVersion(version.Version))
+	if ver.Minor == 0 {
```
ahrtr, line 481: This is a breaking change. Applications based on `clientv3 SDK 4.0` can't talk to etcdserver of previous version (e.g. 3.7, assuming there is no 3.8) if `RejectOldCluster` is true. Obviously the versions (4.0 and 3.7) should be compatible.

Two proposals:
- Clearly document the behaviour change (at least in Changelog). Since `RejectOldCluster` is false by default, it might not be a big problem.
- Always figure out the correct previous minor release no matter current version is the first minor version (e.g. 4.0) or not, and allow client with current version to connect to it. Since we already define a list a version in [version.go#L38-L46](https://github.com/etcd-io/etcd/blob/af07f18dce96d4418bc1e89710675a44e3e5028b/api/version/version.go#L38-L46). We can add them into a slice, then that should be easy enough to do the trick. [**preferred**]

serathius: Not sure how you define a breaking change, but we didn't release v4.0 so don't think we are breaking anyone. When we release v4.0, we can decide whether and how many v3.X minor etcd version we want to support. Please remember that when we will cut v4.0, it will be using different code `client/v4` and not `client/v3` :P.

ahrtr: >  but we didn't release v4.0 so don't think we are breaking anyone.

- Usually we don't add breaking change in a patch release, so no breaking change are breaking anyone at current stage.
- It's a foreseeable behaviour change. Why don't we handle it now with just minor effort? We already handle similar case in [test.sh#L571-L576](https://github.com/etcd-io/etcd/blob/6981e4820fb9ab286d1cb3effc141987d88f8de9/scripts/test.sh#L571-L576) 
- Another approach is to just add a TODO, similar to [downgrade.go#L59](https://github.com/etcd-io/etcd/blob/6981e4820fb9ab286d1cb3effc141987d88f8de9/server/etcdserver/version/downgrade.go#L59). But I think we should update it in a similar way as I mentioned in [above](https://github.com/etcd-io/etcd/pull/16313#discussion_r1275941354).

Overall it isn't a big problem. Either follow my [comment above](https://github.com/etcd-io/etcd/pull/16313#discussion_r1275941354) to make the change, or raise a ticket to track all such changes in 4.0.

> we can decide whether and how many v3.X minor etcd version we want to support

We have already clearly called out that we only support two releases in https://etcd.io/docs/v3.6/op-guide/versioning/

> it will be using different code `client/v4` and not `client/v3`

It's true. But most likely we will just copy the files under client/v3 to client/v4 firstly, and will carry over all legacy potential issues.

ahrtr: @CaojiamingAlan Please raise a followup ticket as I mentioned [above](https://github.com/etcd-io/etcd/pull/16313#discussion_r1276190616) . Afterwards, I will approve & merge this PR.

api/version/version.go
```diff
@@ -44,6 +44,8 @@ var (
 	V3_6 = semver.Version{Major: 3, Minor: 6}
 	V3_7 = semver.Version{Major: 3, Minor: 7}
 	V4_0 = semver.Version{Major: 4, Minor: 0}
+
+	AllVersions = []semver.Version{V3_0, V3_1, V3_2, V3_3, V3_4, V3_5, V3_6, V3_7, V4_0}
```
chaochn47, line 48: How about add a comment, this must be sorted in ascending order? 

Alternately, sort this in `minSupportedVersion`?

client/v3/client.go
```diff
@@ -475,6 +476,22 @@ func (c *Client) roundRobinQuorumBackoff(waitBetween time.Duration, jitterFracti
 	}
 }
 
+// minSupportedVersion returns the minimum version supported, which is the previous minor release.
+func minSupportedVersion() *semver.Version {
```
chaochn47, line 480: I think return `semver.Version` will simplify the logic, what do you think? @CaojiamingAlan

CaojiamingAlan: I think so, but all other places return a pointer and I don't know why. I just want to make things consistent.

fuweid (approved):
LGTM! Thanks

jmhbnz (approved):
Nice work @CaojiamingAlan 👍🏻

chaochn47 (approved):
@CaojiamingAlan Thanks for the follow up PR!

ahrtr (approved):
LGTM

Nice work! thx

PR #16318, opened 2023-07-27
Simplify go data race detection controls

### LINKED ISSUE #16315: Enable race detection for all amd64 and arm64 tests

Sub issue under https://github.com/etcd-io/etcd/issues/15951.

Currently in `main` our `scripts/test.sh` logic disables golang test data race detector for all platforms except `amd64`, refer: https://github.com/etcd-io/etcd/blob/main/scripts/test.sh#L69-L78 

```go
# determine whether target supports race detection
if [ -z "${RACE:-}" ] ; then
  if [ "$GOARCH" == "amd64" ]; then
    RACE="--race"
  else
    RACE="--race=false"
  fi
else
  RACE="--race=${RACE:-true}"
fi
``` 

It's possible to override this with a command line param and our `arm64` nightly tests do this for one specific test, refer: https://github.com/etcd-io/etcd/blob/main/.github/workflows/tests-arm64-template.yaml#L70 

```bash
            linux-arm64-unit-4-cpu-race)
              GOOS=linux GOARCH=arm64 CPU=4 RACE=true ${{ inputs.unitTestCmd }}
            ;;
```

Where things get a little interesting is that for `amd64`, even though we enable race by default the way the workflow tests are defined it appears as though only one is intended to have race enabled? 

Example - workflow file https://github.com/etcd-io/etcd/blob/main/.github/workflows/tests.yaml defines the following targets:

```yaml
        target:
          - linux-amd64-integration-1-cpu
          - linux-amd64-integration-2-cpu
          - linux-amd64-integration-4-cpu
          - linux-amd64-unit-4-cpu-race
          - linux-386-unit-1-cpu
```

However if we look at any of the `amd64` targets that don't include `race` in the name, they still operate with `--race` enabled, for example: https://github.com/etcd-io/etcd/actions/runs/5674542350/job/15378271400#step:5:113

I'm wondering if there is an opportunity to simplify the test runs and just enable race for both `arm64` and `amd64` for all test jobs?

Alternatively we should ensure race is only enabled for the jobs that specifically have race in the name.

### Why is this needed?

Simplifying our use of `--race` for tests for `amd64` and `arm64` would simplify the code slightly.

Additionally doing this will help my cause in [streamlining test commands between branches](https://github.com/etcd-io/etcd/issues/16176).

Currently this situation is creating minor test behavior difference between `amd64` and `arm64` that I would like to resolve.

jmhbnz:
Use [data race detection](https://go.dev/doc/articles/race_detector) by default for all `amd64` and `arm64` tests.  This is already happening for `amd64` although may have been unexpected.  This pull request aligns both `arm64` so both platforms are treated the same.

Also removes now extraneous parameters in github workflows to centralise the control for race to the defaults defined in `scripts/test.sh`.

Fixes #16315

### DIFF
--- a/.github/workflows/e2e-arm64-template.yaml
+++ b/.github/workflows/e2e-arm64-template.yaml
@@ -44,7 +44,7 @@ jobs:
           echo "${TARGET}"
           case "${TARGET}" in
             linux-arm64-e2e)
-              GOOS=linux GOARCH=arm64 CPU=4 EXPECT_DEBUG=true RACE=true ${{ inputs.e2eTestCmd }}
+              GOOS=linux GOARCH=arm64 CPU=4 EXPECT_DEBUG=true ${{ inputs.e2eTestCmd }}
               ;;
             *)
               echo "Failed to find target"
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -27,10 +27,10 @@ jobs:
           echo "${TARGET}"
           case "${TARGET}" in
             linux-amd64-e2e)
-              VERBOSE=1 GOOS=linux GOARCH=amd64 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e-release
+              VERBOSE=1 GOOS=linux GOARCH=amd64 CPU=4 EXPECT_DEBUG=true make test-e2e-release
               ;;
             linux-386-e2e)
-              VERBOSE=1 GOOS=linux GOARCH=386 CPU=4 EXPECT_DEBUG=true RACE=true make test-e2e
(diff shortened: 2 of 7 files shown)


serathius (approved):

ahrtr (approved):
LGTM

thx @jmhbnz for the nice cleanup and improvement.

jmhbnz:
Just need to review any impacts to workflow execution time for the `arm64` nightlies once this merges.

jmhbnz:
Hey team - just to confirm, this is ready to merge from my perspective 🙏🏻 

I will review any impacts to the run duration of our `arm64` nightly jobs afterwards but I am not expecting any issues as the one `arm64` job we already have running with `race=true` completes in similar duration to the corresponding `amd64` job.

PR #16320, opened 2023-07-27
Remove RLock/RUnlock from BatchTx

serathius:

### DIFF
--- a/server/auth/store.go
+++ b/server/auth/store.go
@@ -206,7 +206,9 @@ type AuthBackend interface {
 }
 
 type AuthBatchTx interface {
-	AuthReadTx
+	Lock()
+	Unlock()
+	UnsafeAuthReader
 	UnsafeSaveAuthEnabled(enabled bool)
 	UnsafeSaveAuthRevision(rev uint64)
 	UnsafePutUser(*authpb.User)
@@ -216,14 +218,18 @@ type AuthBatchTx interface {
 }
 
 type AuthReadTx interface {
+	RLock()
+	RUnlock()
+	UnsafeAuthReader
+}
+
+type UnsafeAuthReader interface {
 	UnsafeReadAuthEnabled() bool
 	UnsafeReadAuthRevision() uint64
 	UnsafeGetUser(string) *authpb.User
 	UnsafeGetRole(string) *authpb.Role
 	UnsafeGetAllUsers() []*authpb.User
 	UnsafeGetAllRoles() []*authpb.Role
-	Lock()
-	Unlock()
 }
 
 type authStore struct {
@@ -354,8 +360,8 @@ func (as *authStore) CheckPassword(username, password string) (uint64, error) {
 	// to avoid putting it in the critical section of the tx lock.
 	revision, err := func() (uint64, error) {
 		tx := as.be.ReadTx()
-		tx.Lock()
-		defer tx.Unlock()
+		tx.RLock()
+		defer tx.RUnlock()
 
 		user = tx.UnsafeGetUser(username)
 		if user == nil {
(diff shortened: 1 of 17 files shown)


serathius:
cc @ahrtr @jame @CaojiamingAlan @fuweid

CaojiamingAlan (approved):
LGTM

server/auth/store.go
```diff
@@ -216,14 +218,18 @@ type AuthBatchTx interface {
 }
 
 type AuthReadTx interface {
+	RLock()
+	RUnlock()
+	UnsafeRead
+}
+
+type UnsafeRead interface {
```
ahrtr, line 226: Suggestion: rename to `UnsafeAuthReadTx` or `UnsafeReadTx`, because `UnsafeRead` looks like a method name

server/storage/backend/read_tx.go
```diff
@@ -28,7 +28,10 @@ import (
 type ReadTx interface {
 	RLock()
 	RUnlock()
+	UnsafeRead
+}
 
+type UnsafeRead interface {
```
ahrtr, line 34: ditto, rename to `UnsafeReadTx`

server/auth/store.go
serathius: Name intentionally skipped `Tx` in name.`UnsafeRead` is a group of methods for reading in unsafe way. I wanted to differentiate fact that is no longer a transaction, because it doesn't give access to `Lock`/`Unlock`. This means that consumer doesn't control other changes that can go the transaction, aka not a transaction.

ahrtr: > UnsafeRead is a group of methods for reading

Interface provide a way to specify the behavior of an object. It should be named in a declarative way (e.g. `UnsafeReadTx`), instead of imperative (e.g. `UnsafeRead`).

> because it doesn't give access to Lock/Unlock

It's why it's `Unsafe*Tx`

serathius: > It should be named in a declarative way (e.g. UnsafeReadTx), instead of imperative (e.g. UnsafeRead).

Don't think this has any relation to being declarative vs imperative. Both words are noun. "Read" is a noun, "Txn" is a noun. When you combine them "read txn", then "read becomes a adjective while "txn" remains a noun. Overall it's a correct english. I explicitly want to under underscore that read functions under this interface are just reads and not transactions. 

> It's why it's Unsafe*Tx

No, "unsafe" means that consumer needs to call `Lock` before.  You are mixing consumer of "UnsafeRead" interface, which is a function, and the caller of the `UnsafeX` function which can takes "UnsafeRead" as argument.

ahrtr: Read https://go.dev/doc/effective_go#interface-names

serathius: Ok, that's separate independent convention for Go. To use `agent noun` instead of just `noun`. I can update `UnsafeRead` to `UnsafeReader`.

ahrtr: Right, `UnsafeReader` is much better.

server/auth/store.go
```diff
@@ -216,14 +218,18 @@ type AuthBatchTx interface {
 }
 
 type AuthReadTx interface {
+	RLock()
+	RUnlock()
+	UnsafeReader
+}
+
+type UnsafeReader interface {
```
ahrtr, line 226:
```suggestion
type UnsafeAuthReader interface {
```

serathius: It's in package auth so `backend.UnsafeReader` and `auth.UnsafeReader` should suffice. If not I can link you this convention in https://go.dev/doc/effective_go

serathius: https://go.dev/blog/package-names

ahrtr: `AuthReadTx` and `AuthBatchTx` are all bad names, the point is to keep them consistent. Also when we search the code base using "UnsafeReader", so as to differentiate with the "UnsafeReader" in package backend.

I won't insist on this. Probably we should correct `AuthReadTx` and `AuthBatchTx` as well.

serathius: I can fix that, if you have review capacity.

ahrtr: We already have `AuthBackend`, `AuthBatchTx`, `AuthReadTx` and `UnsafeAuthReader` under package `auth`.   Although they are all bad names based on golang's convention, but it's clearer from pragmatic perspective because it's clearly differentiated from `Backend`, `BatchTx`, `ReadTx` and `UnsafeReader` under package backend.

serathius: Ok, don't care. Let's just merge it.

ahrtr (approved):

Issue #16290, opened 2023-07-24
Dependency: bump gRPC


ahrtr:
### What would you like to be added?

Currently etcd depends on gRPC 1.51, and it's far behind the latest gRPC version 1.56. But etcd has compatibility issue with gRPC 1.52. So we should resolve this.

### Why is this needed?

The newer gRPC versions contain lot of bug fixes.

ahrtr:
There is a pending PR https://github.com/etcd-io/etcd/pull/15131

Issue #13192, opened 2021-07-07
etcd client v3 v3.5.0 sends invalid `:authority` header field value


tatsuhiro-t:
Starting from etcd v3.5.0, etcd client v3 sends invalid `:authority` header field value which violates RFC.

Suppose we created a client like so:
```go
        c, err := clientv3.New(clientv3.Config{
                Endpoints: []string{
                        "https://example.com:2379",
                },
                ...
        }  
```

And started making a connection and tried to get some stuff from etcd server.

In the previous etcd client v3, it sends `:authority` header field value like this:

```
example.com:2379
```

But now with the v3.5.0, it sends like this:

```
#initially=[https://example.com:2379]
```

According to the RFC7540, `:authority` header field contains the authority portion of the target URI.
RFC3986 dictates that authority is:

```
authority   = [ userinfo "@" ] host [ ":" port ]
host        = IP-literal / IPv4address / reg-name
IP-literal = "[" ( IPv6address / IPvFuture  ) "]"
IPvFuture  = "v" 1*HEXDIG "." 1*( unreserved / sub-delims / ":" )
reg-name    = *( unreserved / pct-encoded / sub-delims )
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="
```

`#initially=[https://example.com:2379]` is not a valid authority.

It seems to me that there might be a bug in the code that authority is not parsed or extracted correctly.
The annoying fact of this issue is that it works if server does not validate the invalid field value.  But if you have a proxy between client and server which checks some non-compliant octets, then proxy refuses a request.

tatsuhiro-t:
Any news?  This issue prevents us from using a proxy between etcd client and server.

gbolo:
Please fix this issue, as linkerd-proxy (using the a rust h2 lib) prevents this traffic from passing due to this invalid authority header value!

harshavardhana:
Looks like an expected implementation 

https://github.com/etcd-io/etcd/blob/55b7b745891e816b257681a297a151ca14536fa1/CHANGELOG-3.5.md#package-clientv3

> Endpoints self identify now as etcd-endpoints://{id}/#initially={list of endpoints} e.g. etcd-endpoints://0xc0009d8540/#initially=[localhost:2079]


[...]

PR #16324, opened 2023-07-28
Fix http2 authority header in multiple endpoints scenario and bump up grpc from `v1.51.0` to `v1.52.0`

chaochn47:
Replace https://github.com/etcd-io/etcd/pull/15131 with [endpoints.Interpret returns Host:port as ServerName](https://github.com/etcd-io/etcd/commit/9aefd3452b905a0cef01ff596a208ed0d03ca69f). It should also help https://github.com/etcd-io/etcd/issues/16290

After this change, each h2 request will be sent with the proper authority header value instead of only the first one out of three dialing endpoints. This should resolve the issue mentioned in https://github.com/etcd-io/etcd/issues/13192#issuecomment-929229120. 

Note: client certificate validation on server name won't break because of https://github.com/grpc/grpc-go/pull/3082


chaochn47:
e2e test TestAuthority also needs update since it verifies the authority header value only matches the first endpoint.

I need to replace etcdctl with etcd client or somehow send at least 3 requests to round-robin it. 

TIL: run `stress` to validate the test is not flaky and its correctness. 

```
2023-07-28T00:17:39.5254127Z --- FAIL: TestAuthority (37.13s)
2023-07-28T00:17:39.5254617Z     --- PASS: TestAuthority/Size:_1,_Scenario:_"unix:path" (0.76s)
2023-07-28T00:17:39.5255084Z     --- PASS: TestAuthority/Size:_3,_Scenario:_"unix:path" (1.40s)
2023-07-28T00:17:39.5255597Z     --- PASS: TestAuthority/Size:_1,_Scenario:_"unix://absolute_path" (1.06s)
2023-07-28T00:17:39.5256115Z     --- PASS: TestAuthority/Size:_3,_Scenario:_"unix://absolute_path" (2.39s)
2023-07-28T00:17:39.5256615Z     --- PASS: TestAuthority/Size:_1,_Scenario:_"unixs:absolute_path" (0.77s)
2023-07-28T00:17:39.5257116Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"unixs:absolute_path" (4.05s)
2023-07-28T00:17:39.5257624Z     --- SKIP: TestAuthority/Size:_1,_Scenario:_"unixs://absolute_path" (0.05s)
2023-07-28T00:17:39.5258131Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"unixs://absolute_path" (4.55s)
2023-07-28T00:17:39.5258641Z     --- PASS: TestAuthority/Size:_1,_Scenario:_"http://domain[:port]" (0.36s)
2023-07-28T00:17:39.5259146Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"http://domain[:port]" (3.53s)
2023-07-28T00:17:39.5259719Z     --- SKIP: TestAuthority/Size:_1,_Scenario:_"http://address[:port]" (0.05s)
2023-07-28T00:17:39.5260217Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"http://address[:port]" (1.77s)
2023-07-28T00:17:39.5261198Z     --- SKIP: TestAuthority/Size:_1,_Scenario:_"https://domain[:port]_insecure" (0.05s)
2023-07-28T00:17:39.5421991Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"https://domain[:port]_insecure" (4.51s)
2023-07-28T00:17:39.5422977Z     --- PASS: TestAuthority/Size:_1,_Scenario:_"https://address[:port]_insecure" (1.12s)
2023-07-28T00:17:39.5424174Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"https://address[:port]_insecure" (4.44s)
2023-07-28T00:17:39.5424824Z     --- PASS: TestAuthority/Size:_1,_Scenario:_"https://domain[:port]" (0.87s)
2023-07-28T00:17:39.5425348Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"https://domain[:port]" (2.46s)
2023-07-28T00:17:39.5425890Z     --- SKIP: TestAuthority/Size:_1,_Scenario:_"https://address[:port]" (0.05s)
2023-07-28T00:17:39.5426435Z     --- FAIL: TestAuthority/Size:_3,_Scenario:_"https://address[:port]" (2.88s)
```

chaochn47:
```
$ stress -p=1 ./e2e.test -test.run TestAuthority

...
11m10s: 16 runs so far, 0 failures
```

chaochn47:
Ping @ahrtr @serathius @jmhbnz @fuweid

tests/e2e/ctl_v3_grpc_test.go
```diff
@@ -127,20 +127,20 @@ func TestAuthority(t *testing.T) {
...
 				defer epc.Close()
-				endpoints := templateEndpoints(t, tc.clientURLPattern, epc)
 
-				client, err := e2e.NewEtcdctl(cfg.Client, endpoints)
-				assert.NoError(t, err)
-				err = client.Put(ctx, "foo", "bar", config.PutOptions{})
-				if err != nil {
-					t.Fatal(err)
+				endpoints := templateEndpoints(t, tc.clientURLPattern, epc)
+				client := newClient(t, endpoints, cfg.Client)
```
ahrtr, line 132: Why change this? e2e test is supposed to use etcdctl to talk to etcdserver.

chaochn47: The possibility of using `etcdctl` to round-robin requests to different etcd servers is low. 
```
etcdctl --endpoints IP_1, IP_2, IP3 put key foo
etcdctl --endpoints IP_1, IP_2, IP3 put key foo
etcdctl --endpoints IP_1, IP_2, IP3 put key foo
```
is likely send all the requests to IP_1. 

It is because https://github.com/etcd-io/etcd/blob/c6fd719ccf2891b9e8c61dbb6bbc9ba63abbdd12/tests/framework/e2e/etcdctl.go#L154-L162 establish a short-lived connection and gRPC client usually picks the first IP to connect with. 

So I would like to use `clientv3.Client` to establish a long-lived connection and ensure client requests are evenly sent to multiple servers.

ahrtr:
We either:
1. set the [serverName](https://github.com/etcd-io/etcd/blob/44f6f4a4a272d561c987bf30d081fe35546f78ba/client/v3/internal/endpoint/endpoint.go#L136) to `Host:Port`, just as this PR does;
2. Or keep it's as it's (set ServerName to only Hostname), but update the test case, just as https://github.com/etcd-io/etcd/pull/15131 did.

Both solutions seem work, and I do not see it will break anything with either solution. But based on @dfawley's comment in https://github.com/grpc/grpc-go/pull/5748#issuecomment-1385964654, it seems that he likes option 2 above (setting `ServerName` to the host only) :)

ahrtr:
Also I suggest to bump gRPC to a more latest version, e.g. 1.56.2 or 1.57.0.

chaochn47:
No, authority header value should contains port. It is based on RFC. https://www.rfc-editor.org/rfc/rfc3986#section-3.2.3

tests/e2e/ctl_v3_grpc_test.go
serathius: This test was explicitly added to test `etcdctl` and not `clientv3`. There is matching integration tests for `clientv3`. Please keep the `etcdctl` here.

tests/integration/grpc_test.go
```diff
@@ -162,10 +164,11 @@ func templateAuthority(t *testing.T, pattern string, m *integration.Member) stri
 	return authority
 }
 
-func assertAuthority(t *testing.T, expectedAuthority string, clus *integration.Cluster, filterMethod string) {
+func assertAuthority(t *testing.T, expectedAuthorityPattern string, clus *integration.Cluster, filterMethod string) {
 	t.Helper()
 	requestsFound := 0
 	for _, m := range clus.Members {
+		expectedAuthority := templateAuthority(t, expectedAuthorityPattern, m)
```
fuweid, line 171: Since you introduce multiple calls in this patch, I think it makes sense to check `if requestsFound == 0` for each member. What do you think?

ahrtr:
> No, authority header value should contains port. It is based on RFC. https://www.rfc-editor.org/rfc/rfc3986#section-3.2.3

port is optional.

`authority   = [ userinfo "@" ] host [ ":" port ]`

tests/e2e/ctl_v3_grpc_test.go
```diff
@@ -156,16 +156,20 @@ func templateEndpoints(t *testing.T, pattern string, clus *e2e.EtcdProcessCluste
 	return endpoints
 }
 
-func assertAuthority(t *testing.T, expectAurhority string, clus *e2e.EtcdProcessCluster) {
-	var logs []e2e.LogsExpect
-	for _, proc := range clus.Procs {
-		logs = append(logs, proc.Logs())
+func assertAuthority(t *testing.T, expectAuthorityPattern string, clus *e2e.EtcdProcessCluster) {
+	for i := range clus.Procs {
+		line := firstMatch(t, `http2: decoded hpack field header field ":authority"`, clus.Procs[i].Logs())
```
fuweid, line 161: I think we can just use `clus.Procs[i].ExpectWithContext` here because in your patch, it doesn't merge member logs.

fuweid (approved):
It looks good to me.
Maybe we can consider changing the title with `Fix http2 authority header in multiple endpoints scenario`

fuweid:
> set the [serverName](https://github.com/etcd-io/etcd/blob/44f6f4a4a272d561c987bf30d081fe35546f78ba/client/v3/internal/endpoint/endpoint.go#L136) to Host:Port, just as this PR does;

I vote for this. If the endpoint is just ip:port, and according to RFC, it's valid to have port here. It doesn't look correct if we just have IP here.

dfawley:
> But based on @dfawley's comment in https://github.com/grpc/grpc-go/pull/5748#issuecomment-1385964654, it seems that he likes option 2 above (setting ServerName to the host only)

I'm not sure about the details in the test you're referring to, but by default, gRPC will use the full "endpoint" portion of the dial target as the authority ([ref](https://github.com/grpc/grpc-go/blob/b8d36caf8d036e1fb91898bf93ccdc7c4e5a40ea/clientconn.go#L1947C37-L1947C37)).  It's very common to dial `host:port`, so we will use a port quite often.

> Also I suggest to bump gRPC to a more latest version, e.g. 1.56.2 or 1.57.0.

This is important.  If you're using experimental APIs and your users upgrade to the latest version (or if one of their dependencies does), then you will break them if you are not keeping up with recent and upcoming changes, e.g. https://github.com/grpc/grpc-go/pull/6363, https://github.com/grpc/grpc-go/pull/6451, and the changes announced here https://github.com/grpc/grpc-go/issues/6472.  (I.e. https://github.com/etcd-io/etcd/issues/15145)

ahrtr:
thx @dfawley for the feedback. 

Then let's upgrade to 1.52 for now, and will take a closer look at the links you mentioned later.

chaochn47:
>  If you're using experimental APIs and your users upgrade to the latest version (or if one of their dependencies does), then you will break them if you are not keeping up with recent and upcoming changes

Thanks for calling it out. It's indeed a risk. ref. https://go.dev/ref/mod#minimal-version-selection

tests/e2e/ctl_v3_grpc_test.go
chaochn47: > There is matching integration tests for `clientv3`

okay. It's fair. I will try workaround based on `etcdctl`.

tests/integration/grpc_test.go
chaochn47: agreed. Thanks for calling it out.

tests/e2e/ctl_v3_grpc_test.go
chaochn47: Yeah, it can be simplified.

chaochn47:
> port is optional.
> authority   = [ userinfo "@" ] host [ ":" port ]

@ahrtr 

Just as the original 3.5.0 issue https://github.com/etcd-io/etcd/issues/13192 stated, the use case is putting etcd server behinds TCP proxy. 

With that in mind, etcd client (based on gRPC client [support on TCP proxy](https://github.com/grpc/proposal/blob/master/A1-http-connect-proxy-support.md)) first needs to establish connection with the proxy with CONNECT method. https://www.rfc-editor.org/rfc/rfc7540#section-8.3

The proxy server needs to know which etcd server it should connect to from the authority header. So that's why port should be included. 

However, there is an exception that if etcd server is configured to listen on default port of the scheme (:443 on https and :80 on http). It is descried in https://www.rfc-editor.org/rfc/rfc3986#section-3.2.3 and the proxy server should support it.

ahrtr (approved):

jmhbnz (approved):
LGTM

fuweid:
For e2e test case, the `etcdctl check perf` can sends request to all the endpoints. But it needs 60s to finish the performance at least. Maybe we can introduce `--timeout` for it and we don't need to care the error.

serathius (approved):

PR #16333, opened 2023-07-31
Introduce Kubernetes interface to etcd client

serathius:
Matching ticket on Kubernetes side https://github.com/kubernetes/kubernetes/pull/125258
cc @ahrtr @logicalhan

Design: https://docs.google.com/document/d/1-nIpoW87qqQ9FxINOzXPkhu7CD_FWW1piYqHZDr39nU/edit?usp=sharing

### DIFF
--- a/client/v3/client.go
+++ b/client/v3/client.go
@@ -154,7 +154,7 @@ func (c *Client) Close() error {
 		c.Lease.Close()
 	}
 	if c.conn != nil {
-		return toErr(c.ctx, c.conn.Close())
+		return ContextError(c.ctx, c.conn.Close())
 	}
 	return c.ctx.Err()
 }
@@ -598,7 +598,9 @@ func isUnavailableErr(ctx context.Context, err error) bool {
 	return false
 }
 
-func toErr(ctx context.Context, err error) error {
+// ContextError converts the error into an EtcdError if the error message matches one of
+// the defined messages; otherwise, it tries to retrieve the context error.
+func ContextError(ctx context.Context, err error) error {
 	if err == nil {
 		return nil
 	}
--- a/client/v3/kubernetes/client.go
+++ b/client/v3/kubernetes/client.go
@@ -1,0 +1,165 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kubernetes
+
+import (
+	"context"
+
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+// New creates Client from config.
+// Caller is responsible to call Close() to clean up client.
+func New(cfg clientv3.Config) (*Client, error) {
+	c, err := clientv3.New(cfg)
+	if err != nil {
+		return nil, err
+	}
+	kc := &Client{
+		Client: c,
+		kv:     clientv3.RetryKVClient(c),
+	}
+	kc.Kubernetes = kc
+	return kc, nil
+}
+
+type Client struct {
+	*clientv3.Client
+	Kubernetes Interface
+	kv         pb.KVClient
+}
+
+var _ Interface = (*Client)(nil)
+
+func (k Client) Get(ctx context.Context, key string, opts GetOptions) (resp GetResponse, err error) {
+	rangeResp, err := k.kv.Range(ctx, getRequest(key, opts.Revision))
+	if err != nil {
+		return resp, clientv3.ContextError(ctx, err)
+	}
+	resp.Revision = rangeResp.Header.Revision
+	if len(rangeResp.Kvs) == 1 {
+		resp.KV = rangeResp.Kvs[0]
+	}
+	return resp, nil
+}
+
+func (k Client) List(ctx context.Context, prefix string, opts ListOptions) (resp ListResponse, err error) {
+	rangeStart := prefix + opts.Continue
+	rangeEnd := clientv3.GetPrefixRangeEnd(prefix)
+
+	rangeResp, err := k.kv.Range(ctx, &pb.RangeRequest{
+		Key:      []byte(rangeStart),
+		RangeEnd: []byte(rangeEnd),
+		Limit:    opts.Limit,
+		Revision: opts.Revision,
+	})
+	if err != nil {
+		return resp, clientv3.ContextError(ctx, err)
+	}
+	resp.Kvs = rangeResp.Kvs
+	resp.Count = rangeResp.Count
+	resp.Revision = rangeResp.Header.Revision
+	return resp, nil
+}
+
+func (k Client) Count(ctx context.Context, prefix string, _ CountOptions) (int64, error) {
+	resp, err := k.kv.Range(ctx, &pb.RangeRequest{
+		Key:       []byte(prefix),
+		RangeEnd:  []byte(clientv3.GetPrefixRangeEnd(prefix)),
+		CountOnly: true,
+	})
+	if err != nil {
+		return 0, clientv3.ContextError(ctx, err)
+	}
+	return resp.Count, nil
+}
+
+func (k Client) OptimisticPut(ctx context.Context, key string, value []byte, expectedRevision int64, opts PutOptions) (resp PutResponse, err error) {
+	onSuccess := &pb.RequestOp{Request: &pb.RequestOp_RequestPut{RequestPut: &pb.PutRequest{Key: []byte(key), Value: value, Lease: int64(opts.LeaseID)}}}
+
+	var onFailure *pb.RequestOp
+	if opts.GetOnFailure {
+		onFailure = &pb.RequestOp{Request: &pb.RequestOp_RequestRange{RequestRange: getRequest(key, 0)}}
+	}
+
+	txnResp, err := k.optimisticTxn(ctx, key, expectedRevision, onSuccess, onFailure)
+	if err != nil {
+		return resp, clientv3.ContextError(ctx, err)
+	}
+	resp.Succeeded = txnResp.Succeeded
+	resp.Revision = txnResp.Header.Revision
+	if opts.GetOnFailure && !txnResp.Succeeded {
+		resp.KV = kvFromTxnResponse(txnResp.Responses[0])
+	}
+	return resp, nil
+}
+
+func (k Client) OptimisticDelete(ctx context.Context, key string, expectedRevision int64, opts DeleteOptions) (resp DeleteResponse, err error) {
+	onSuccess := &pb.RequestOp{Request: &pb.RequestOp_RequestDeleteRange{RequestDeleteRange: &pb.DeleteRangeRequest{Key: []byte(key)}}}
+
+	var onFailure *pb.RequestOp
+	if opts.GetOnFailure {
+		onFailure = &pb.RequestOp{Request: &pb.RequestOp_RequestRange{RequestRange: getRequest(key, 0)}}
+	}
+
+	txnResp, err := k.optimisticTxn(ctx, key, expectedRevision, onSuccess, onFailure)
+	if err != nil {
+		return resp, clientv3.ContextError(ctx, err)
+	}
+	resp.Succeeded = txnResp.Succeeded
+	resp.Revision = txnResp.Header.Revision
+	if opts.GetOnFailure && !txnResp.Succeeded {
+		resp.KV = kvFromTxnResponse(txnResp.Responses[0])
+	}
+	return resp, nil
+}
+
+func (k Client) optimisticTxn(ctx context.Context, key string, expectedRevision int64, onSuccess, onFailure *pb.RequestOp) (*pb.TxnResponse, error) {
+	txn := &pb.TxnRequest{
+		Compare: []*pb.Compare{
+			{
+				Result:      pb.Compare_EQUAL,
+				Target:      pb.Compare_MOD,
+				Key:         []byte(key),
+				TargetUnion: &pb.Compare_ModRevision{ModRevision: expectedRevision},
+			},
+		},
+	}
+	if onSuccess != nil {
+		txn.Success = []*pb.RequestOp{onSuccess}
+	}
+	if onFailure != nil {
+		txn.Failure = []*pb.RequestOp{onFailure}
+	}
+	return k.kv.Txn(ctx, txn)
+}
+
+func getRequest(key string, revision int64) *pb.RangeRequest {
+	return &pb.RangeRequest{
+		Key:      []byte(key),
+		Revision: revision,
+		Limit:    1,
+	}
+}
+
+func kvFromTxnResponse(resp *pb.ResponseOp) *mvccpb.KeyValue {
+	getResponse := resp.GetResponseRange()
+	if len(getResponse.Kvs) == 1 {
+		return getResponse.Kvs[0]
+	}
+	return nil
+}
--- a/client/v3/kubernetes/interface.go
+++ b/client/v3/kubernetes/interface.go
@@ -1,0 +1,140 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kubernetes
+
+import (
+	"context"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+// Interface defines the minimal client-side interface that Kubernetes requires
+// to interact with etcd. Methods below are standard etcd operations with
+// semantics adjusted to better suit Kubernetes' needs.
+type Interface interface {
+	// Get retrieves a single key-value pair from etcd.
+	//
+	// If opts.Revision is set to a non-zero value, the key-value pair is retrieved at the specified revision.
+	// If the required revision has been compacted, the request will fail with ErrCompacted.
+	Get(ctx context.Context, key string, opts GetOptions) (GetResponse, error)
+
+	// List retrieves key-value pairs with the specified prefix, ordered lexicographically by key.
+	//
+	// If opts.Revision is non-zero, the key-value pairs are retrieved at the specified revision.
+	// If the required revision has been compacted, the request will fail with ErrCompacted.
+	// If opts.Limit is greater than zero, the number of returned key-value pairs is bounded by the limit.
+	// If opts.Continue is not empty, the listing will start from the key immediately after the one specified by Continue.
+	// The Continue value should be the last key returned in a previous paginated ListResponse.
+	List(ctx context.Context, prefix string, opts ListOptions) (ListResponse, error)
+
+	// Count returns the number of keys with the specified prefix.
+	//
+	// Currently, there are no options for the Count operation. However, a placeholder options struct (CountOptions)
+	// is provided for future extensibility in case options become necessary.
+	Count(ctx context.Context, prefix string, opts CountOptions) (int64, error)
+
+	// OptimisticPut creates or updates a key-value pair if the key has not been modified or created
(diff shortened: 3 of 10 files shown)


logicalhan:
> Hello!
> 
> I'm not a developer of etcd or Kubernetes, asking just out of curiosity. From what I can see in the API you've provided, it appears that Kubernates never modifies multiple keys within a transaction. Is it true?

It's not supposed to, it breaks the watch protocol. [See this doc](https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit) for more details.

gritukan:
> It's not supposed to, it breaks the watch protocol. [See this doc](https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit) for more details.

I see, thank you!

I'm a little bit confused about it, since I've read multiple times that etcd is a bottleneck for Kubernates scalability. If there are no multi-keys transactions why isn't is possible to just partition all keys by hash for example, run multiple instances of etcd with each instance handling requests for a given subset of keys and get scalable Kubernates?

serathius:
> I'm a little bit confused about it, since I've read multiple times that etcd is a bottleneck for Kubernates scalability. If there are no multi-keys transactions why isn't is possible to just partition all keys by hash for example, run multiple instances of etcd with each instance handling requests for a given subset of keys and get scalable Kubernates?

It's not, K8s scalability limits are mostly hit by badly written operators that do not use proper API machinery. People are sharding etcd under Kubernetes (for example events are frequently separated due their different nature event vs state).

Note; This is an PR  and not a discussion about the K8s API design. Please leave any follow up questions on this [document](https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit). Any further questions will be removed to avoid cluttering the PR comments

serathius:
Doing a second iteration based on https://github.com/kubernetes/kubernetes/pull/125258

ping @ahrtr for feedback

serathius:
Please comment on https://docs.google.com/document/d/1-nIpoW87qqQ9FxINOzXPkhu7CD_FWW1piYqHZDr39nU/edit?usp=sharing, the document provides the motivation why SIG-etcd and SIG-api-machinery should collaborate on this.

If you don't remember, this was agreed upon a long time ago https://github.com/kubernetes/community/blob/master/sig-etcd/vision.md#etcd-is-a-standalone-solution-for-managing-infrastructure-configuration

serathius:
The interface was approved by @wojtek-t and @deads2k, we can start looking how we bring it in etcd.

cc @MadhavJivrajani @siyuanfoundation @fuweid

ahrtr:
> The interface was approved by @wojtek-t and @deads2k, we can start looking how we bring it in etcd.

Being approved by anyone != adding it into etcd. You can add it into K8s as well.

serathius:
We can start looking how we bring it in etcd, means we can start looking. It doesn't mean that it's approved by etcd.

client/v3/client.go
```diff
@@ -52,6 +52,7 @@ type Client struct {
 	Watcher
 	Auth
 	Maintenance
+	Kubernetes Kubernetes
```
ahrtr, line 55: etcd sdk is used by many users, including Kubernetes.  Some users may not care about the `Kubernetes` interface at all. It doesn't make much sense to expose `Kubernetes` interface to users other than Kubernetes or Controllers.

How about this? 
```
package kubernetes

func New(c *Client) Kubernetes {
	return &kubernetesClient{
		Client: c,
		kv: RetryKVClient(c)
	}
}

type kubernetesClient struct {
	Client
	kv pb.KVClient
}

......
```

api/v3rpc/rpctypes/error.go
```diff
@@ -277,3 +279,25 @@ func ErrorDesc(err error) string {
 	}
 	return err.Error()
 }
+
+func ContextError(ctx context.Context, err error) error {
```
serathius, line 283: Need to decide where to put this function, we either expose it from clientv3, or move to api as proposed here.

rpctypes package seems a good place as it already has functions to extract errors from etcd rpc.

client/v3/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,166 @@
...
+
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+// New creates KubernetesClient from config.
+// Caller is responsible to call Close() to clean up client.
+func New(cfg clientv3.Config) (*KubernetesClient, error) {
```
serathius, line 14: Depending on user it might be useful create KubernetesClient from config or existing client. I expect Kubernetes will prefer to use existing client as it would want avoid duplicating etcd connection. The downside of this flexibility

serathius:
With https://github.com/kubernetes/enhancements/pull/4744 almost approved we can move forward with the PR.

Still need to add tests, however the code is ready to be reviewed.

api/v3rpc/rpctypes/error.go
ahrtr, line 283:
```suggestion
// ConvertError converts the error into an EtcdError if the error message matches one of
// the defined messages; otherwise, it tries to retrieve the context error.
func ConvertError(ctx context.Context, err error) error {
```

ahrtr: It's only used by the client side, so suggest to put it in `clientv3`. 

If you want quick merge, suggest to do it in a separate PR.

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,62 @@
+package kubernetes
+
+import (
+	"context"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+type Interface interface {
```
jpbetz, line 10: Can the godoc for this interface be mostly copy-pasted from the existing interfaces and then augmented with k8s specifics?

serathius:
ping @ahrtr

client/v3/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,179 @@
...
+	return &KubernetesClient{
+		kv:    clientv3.RetryKVClient(c),
+		close: close,
+	}
+}
+
+type KubernetesClient struct {
+	kv    pb.KVClient
+	close func() error
+}
```
ahrtr, line 51: Are you sure that the user (Kubernetes) will never need to call any methods which have already been supported by `clientv3.Client`? If yes, then you should return the `Interface` (of course, you need to add the `Close` method into the interface) instead of `*KubernetesClient`.
```suggestion
func fromClient(c *clientv3.Client, close func() error) *KubernetesClient {
	return &KubernetesClient{
		Client: c,
		kv:    clientv3.RetryKVClient(c),
	}
}

type KubernetesClient struct {
	*clientv3.Client
	kv    pb.KVClient
}
```

client/v3/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return nil
+}
+
+func (k KubernetesClient) Close() error {
+	if k.close == nil {
+		return nil
+	}
+	return k.close()
+}
```
ahrtr, line 179: The `Client` methods (including `Close`) are automatically promoted into `KubernetesClient`
```suggestion
```

ahrtr: Refer to https://go.dev/ref/spec#Struct_types

client/v3/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return fromClient(c, c.Close), nil
+}
+
+// FromClient creates KubernetesClient from existing client.
+// Caller can still call Close() however, it's not needed as long caller should call Close() on the original client.
+func FromClient(c *clientv3.Client) *KubernetesClient {
+	return fromClient(c, nil)
```
ahrtr, line 38: Or you can remove the `fromClient`, and move the implementation into `FromClient`
```suggestion
	return fromClient(c)
```

ahrtr:
Also please add an unit test file (kubernetes_test.go), which can be done in a followup PR if you want.

client/v3/kubernetes/kubernetes.go
serathius: Done

client/v3/kubernetes/kubernetes.go
serathius: Done

client/v3/kubernetes/kubernetes.go
serathius: Done

client/v3/kubernetes/client.go
```diff
@@ -0,0 +1,165 @@
...
+		Client: c,
+		kv:     clientv3.RetryKVClient(c),
+	}
+	kc.Kubernetes = kc
+	return kc, nil
+}
+
+type Client struct {
+	*clientv3.Client
+	Kubernetes Interface
```
ahrtr, line 42: why add this?
```suggestion
```

serathius: Having an explicit interface is a feature of etcd client that Kubernetes uses to inject it's own hooks. For example request latency tracking.

serathius: See https://github.com/kubernetes/kubernetes/pull/125258/commits/0eb566c2ee115d025d4d931b9cc68be5caa459bd

ahrtr: Added a comment https://github.com/kubernetes/kubernetes/pull/125258#discussion_r1669135958

It isn't necessary to add such an additional indirect reference, because `kubernetes.Client` always implements `kubernetes.Interface`.

Actually an even better implementation is to rename `Get` to a different name (e.g. `KGet`) so as to be different from the `kv.Get`,https://github.com/etcd-io/etcd/blob/24e05998c68f481af2bd567ee104cf89879bbf89/client/v3/kv.go#L49

Then you don't need the `clientV3KVLatencyTracker` at all. You can use the `*kubernetes.Client` (returned by `kubernetes.New`) in either places directly. See also https://github.com/etcd-io/etcd/pull/16333#discussion_r1665819091

ahrtr: > Then you don't need the `clientV3KVLatencyTracker` at all. You can use the `*kubernetes.Client` (returned by `kubernetes.New`) in either places directly. See also [#16333 (comment)](https://github.com/etcd-io/etcd/pull/16333#discussion_r1665819091)

This might not be accurate, but the high level idea is that you don't need to differentiate an etcd client or a kubernete client,  a *kubernetes.Client (returned by kubernetes.New) can be used for both cases directly.

serathius: Not sure I understand, the Kubernetes currently wraps only `KV` interface that is part of client, not a full *clientv3.Client. With introduction of interface I plan to wrap `Kubernetes` interface and not the whole `kubernetes.Client`.

client/v3/kubernetes/client.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	resp.Succeeded = txnResp.Succeeded
+	resp.Revision = txnResp.Header.Revision
+	if opts.GetOnFailure && !txnResp.Succeeded {
+		resp.KV = kvFromTxnResponse(txnResp.Responses[0])
+	}
+	return resp, nil
+}
+
+func (k Client) optimisticTxn(ctx context.Context, key string, expectRevision int64, onSuccess, onFailure *pb.RequestOp) (*pb.TxnResponse, error) {
```
ahrtr, line 131:
```suggestion
func (k Client) optimisticTxn(ctx context.Context, key string, expectedRevision int64, onSuccess, onFailure *pb.RequestOp) (*pb.TxnResponse, error) {
```

ahrtr (approved):
LGTM with a minor comment.

cc @fuweid @ivanvc and @jmhbnz  to take a look as well.

serathius:
Would like to get review from SIG-api-machinery too.
cc @jpbetz @wojtek-t @logicalhan

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type GetOptions struct {
+	Revision int64
+}
+
+type ListOptions struct {
+	Revision int64
+	Limit    int64
+	Continue string
```
jpbetz, line 67: nit: These appear under-documented.

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type PutOptions struct {
+	GetOnFailure bool
```
jpbetz, line 71: Document what exactly constitutes a failure?

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+type ListResponse struct {
+	KVs      []*mvccpb.KeyValue
+	Count    int64
+	Revision int64
+}
+
+type PutResponse struct {
+	KV        *mvccpb.KeyValue
+	Succeeded bool
+	Revision  int64
```
jpbetz, line 95: Docs here too. Particularly around Succeeded=false.

client/v3/kubernetes/client.go
```diff
@@ -0,0 +1,165 @@
...
+
+type Client struct {
+	*clientv3.Client
+	Kubernetes Interface
+	kv         pb.KVClient
+}
+
+var _ Interface = (*Client)(nil)
+
+func (k Client) Get(ctx context.Context, key string, opts GetOptions) (resp GetResponse, err error) {
```
jpbetz, line 48: I was expecting corresponding tests for this interface implementation.

client/v3/kubernetes/client.go
```diff
@@ -0,0 +1,165 @@
...
+	rangeResp, err := k.kv.Range(ctx, &pb.RangeRequest{
+		Key:      []byte(rangeStart),
+		RangeEnd: []byte(rangeEnd),
+		Limit:    opts.Limit,
+		Revision: opts.Revision,
+	})
+	if err != nil {
+		return resp, clientv3.ContextError(ctx, err)
+	}
+	resp.KVs = rangeResp.Kvs
```
jpbetz, line 73: super nit:  Be consistent and use `Kvs`?
```suggestion
	resp.Kvs = rangeResp.Kvs
```

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+	// List retrieves key-value pairs with the specified prefix.
+	//
+	// If opts.Revision is non-zero, the key-value pairs are retrieved at the specified revision.
+	// If the required revision has been compacted, the request will fail with ErrCompacted.
+	// If opts.Limit is greater than zero, the number of returned key-value pairs is bounded by the limit.
+	// If opts.Continue is not empty, the listing will start from the key immediately after the one specified by Continue.
+	List(ctx context.Context, prefix string, opts ListOptions) (ListResponse, error)
+
+	// Count returns the number of keys with the specified prefix.
+	Count(ctx context.Context, prefix string) (int64, error)
```
jpbetz, line 43: The lack of options for Count caught my eye.  Would it be useful to have a placeholder options struct to make it easier to add any options in the future? Would support for revision be useful?

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// OptimisticPut creates or updates a key-value pair if the key has not been modified or created
+	// since the revision specified in expectedRevision. Otherwise, it updates the key-value pair
+	// only if it hasn't been modified since expectedRevision.
+	//
+	// If opts.GetOnFailure is true, the modified key-value pair will be returned if the put operation fails due to a revision mismatch.
+	// If opts.LeaseID is provided, it overrides the lease associated with the key. If not provided, the existing lease is cleared.
+	OptimisticPut(ctx context.Context, key string, value []byte, expectedRevision int64, opts PutOptions) (PutResponse, error)
```
jpbetz, line 51: Potential alternative designs:   
- Just use `Put` for this, but accept `expectedRevision` as an optional option in `Put`, maybe in a `precondition` section.
- Put expectedRevision in an OptimisticPutOptions here.  (OptimisticPutOptions could embed PutOptions and extend it...)

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// OptimisticDelete deletes the key-value pair if it hasn't been modified since the revision
+	// specified in expectedRevision.
+	//
+	// If opts.GetOnFailure is true, the modified key-value pair will be returned if the delete operation fails due to a revision mismatch.
+	OptimisticDelete(ctx context.Context, key string, expectedRevision int64, opts DeleteOptions) (DeleteResponse, error)
```
jpbetz, line 57: Same comments as with `OptimisticPut`

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+	// If the required revision has been compacted, the request will fail with ErrCompacted.
+	Get(ctx context.Context, key string, opts GetOptions) (GetResponse, error)
+
+	// List retrieves key-value pairs with the specified prefix.
+	//
+	// If opts.Revision is non-zero, the key-value pairs are retrieved at the specified revision.
+	// If the required revision has been compacted, the request will fail with ErrCompacted.
+	// If opts.Limit is greater than zero, the number of returned key-value pairs is bounded by the limit.
+	// If opts.Continue is not empty, the listing will start from the key immediately after the one specified by Continue.
+	List(ctx context.Context, prefix string, opts ListOptions) (ListResponse, error)
```
siyuanfoundation, line 40: Please specify if `opts.Limit` is >0, is order of the keys guaranteed, a.k.a. do you always get the same Limit results when the same List is called multiple times? Are the result KVs ordered by key value alphabetically or by rev?

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Count returns the number of keys with the specified prefix.
+	Count(ctx context.Context, prefix string) (int64, error)
+
+	// OptimisticPut creates or updates a key-value pair if the key has not been modified or created
+	// since the revision specified in expectedRevision. Otherwise, it updates the key-value pair
```
siyuanfoundation, line 46: Otherwise, some error would be returned?

client/v3/kubernetes/interface.go
siyuanfoundation: +1. It is unclear what the string value of `Continue` is supposed to be.

client/v3/kubernetes/interface.go
serathius: We had a short discussion with @wojtek-t that we never changed count function and we don't expect to expand it. Still maybe we should consider extensibility here too.

client/v3/kubernetes/interface.go
serathius: Kubernetes doesn't do Put without condition, and don't think it ever will. Don't understand why we would separate it.
Considered having expectedRevision in options, however I consider options, being optional while expectedRevision needs to be always provided because even zero has a semantic meaning.

client/v3/kubernetes/interface.go
serathius: Good point we should document that results are always ordered by key.

client/v3/kubernetes/interface.go
serathius: Done

client/v3/kubernetes/interface.go
serathius: Done

client/v3/kubernetes/client.go
serathius: I'm currently testing in the https://github.com/kubernetes/kubernetes/pull/125258, don't want to add some abstract tests that might diverge from K8s.

client/v3/kubernetes/interface.go
jpbetz: Yeah, We had some regret in client-go that we were not super consistent about having options defined uniformly for all methods and ended up breaking backward compatibility later to fix it.  Maybe just add an empty options type "just in case"?

serathius: Done

client/v3/kubernetes/interface.go
```diff
@@ -0,0 +1,141 @@
...
+
+type ListResponse struct {
+	// KVs is the list of key-value pairs retrieved from etcd, ordered lexicographically by key.
+	KVs []*mvccpb.KeyValue
+
+	// Count is the total number of keys with the specified prefix, even if not all were returned due to a limit.
+	Count int64
+
+	// Revision is the revision of the key-value store at the time of the List operation.
+	Revision int64
```
fuweid, line 116: Do we need to add `More` here? When `ListOptions.Limit` set and there are more recording than limit, the client will know that if it should continue or not.

jpbetz: Good question.  Being really clear on what responsibility this interface is taking on for pagination seems important.

serathius: Don't need more, More is redundant with count.

client/v3/kubernetes/client.go
serathius: Done

client/v3/kubernetes/client.go
serathius: Done

serathius:
Fixed the comments, please take another look @jpbetz @dead2k @wojtek-t

PR #16355, opened 2023-08-02
server: Separate internal txn functions for recursion and have public function create transaction and trace

serathius:
cc @ahrtr @fuweid @jmhbnz @wenjiaswe 

Functions `Range`, `Put`, `DeleteRange` and `Txn` are public function that can be called by both applier and recursively by `Txn`. The top call is responsible for opening/closing a transaction and trace. This complicated code as depending type of call it needed to manage transaction or not. This PR simplifies the code by separating the recursive vs external call, thus creating a begining point for traces and transactions.

### DIFF
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -31,9 +31,7 @@ import (
 	"go.etcd.io/etcd/server/v3/storage/mvcc"
 )
 
-func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, txnWrite mvcc.TxnWrite, p *pb.PutRequest) (resp *pb.PutResponse, trace *traceutil.Trace, err error) {
-	resp = &pb.PutResponse{}
-	resp.Header = &pb.ResponseHeader{}
+func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, p *pb.PutRequest) (resp *pb.PutResponse, trace *traceutil.Trace, err error) {
 	trace = traceutil.Get(ctx)
 	// create put tracing if the trace in context is empty
 	if trace.IsEmpty() {
@@ -42,17 +40,25 @@ func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, t
 			traceutil.Field{Key: "key", Value: string(p.Key)},
 			traceutil.Field{Key: "req_size", Value: p.Size()},
 		)
+		ctx = context.WithValue(ctx, traceutil.TraceKey, trace)
 	}
-	val, leaseID := p.Value, lease.LeaseID(p.Lease)
-	if txnWrite == nil {
-		if leaseID != lease.NoLease {
-			if l := lessor.Lookup(leaseID); l == nil {
-				return nil, nil, lease.ErrLeaseNotFound
-			}
+	leaseID := lease.LeaseID(p.Lease)
+	if leaseID != lease.NoLease {
+		if l := lessor.Lookup(leaseID); l == nil {
+			return nil, nil, lease.ErrLeaseNotFound
 		}
-		txnWrite = kv.Write(trace)
-		defer txnWrite.End()
 	}
+	txnWrite := kv.Write(trace)
+	defer txnWrite.End()
+	resp, err = put(ctx, txnWrite, p)
+	return resp, trace, err
+}
+
+func put(ctx context.Context, txnWrite mvcc.TxnWrite, p *pb.PutRequest) (resp *pb.PutResponse, err error) {
+	trace := traceutil.Get(ctx)
+	resp = &pb.PutResponse{}
+	resp.Header = &pb.ResponseHeader{}
+	val, leaseID := p.Value, lease.LeaseID(p.Lease)
 
 	var rr *mvcc.RangeResult
 	if p.IgnoreValue || p.IgnoreLease || p.PrevKv {
@@ -61,13 +67,13 @@ func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, t
 		}, "get previous kv pair")
 
 		if err != nil {
-			return nil, nil, err
+			return nil, err
 		}
 	}
 	if p.IgnoreValue || p.IgnoreLease {
 		if rr == nil || len(rr.KVs) == 0 {
 			// ignore_{lease,value} flag expects previous key-value pair
-			return nil, nil, errors.ErrKeyNotFound
+			return nil, errors.ErrKeyNotFound
 		}
 	}
 	if p.IgnoreValue {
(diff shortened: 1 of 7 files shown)


serathius:
> Need to address the panic :)

Ups, passed `nil` instead of `s.KV()` in one place.

jmhbnz (commented):
Reading through this simplification makes sense to me, though setting this as a comment rather than approval as want to ensure more eyes on this before it gets merged in case there is some wider implication I am not seeing.

server/etcdserver/txn/txn.go
```diff
@@ -235,8 +243,31 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
...
+	}
+	// When executing mutable txnWrite ops, etcd must hold the txnWrite lock so
+	// readers do not see any intermediate results. Since writes are
+	// serialized on the raft loop, the revision in the read view will
+	// be the revision of the write txnWrite.
+	if isWrite {
+		txnWrite.End()
+		txnWrite = kv.Write(trace)
+	}
+	txnResp, err := txn(ctx, lg, txnWrite, rt, isWrite, txnPath)
```
fuweid, line 259: Originally, (If I understand correctly) there are two phases to check lease ID is valid: `checkRequests`(compare txn) and `applyTxn -> Put`. In this change, we lost the lease ID checker in `applyTxn`. Since we open new transaction after `txnCompare`, it might cause race here? Just have one concern about this.

chaochn47: Looks like the `applyTxn -> put` always skips the step to check if the lease ID exists or not before or after the change because the write transaction is already opened after right after `checkRequests`. 

https://github.com/etcd-io/etcd/blob/9e1abbab6e4ba2886238f49b0d48fc19a546c7cf/server/etcdserver/txn/txn.go#L47-L55

So the race could be always there theoretically. The argument of the race does not exist is associating the nonexist lease could fail in [storeTxnWrite.put](https://github.com/etcd-io/etcd/blob/29769984e62ec9422ff0555ada919d913549d7c1/server/storage/mvcc/kvstore_txn.go#L258)

chaochn47 (approved):
LGTM. Thanks @serathius!

server/etcdserver/txn/txn.go
fuweid: oh, yeah. There is if-condition about the transaction.

fuweid (approved):
LGTM

server/etcdserver/txn/txn.go
```diff
@@ -43,16 +41,23 @@ func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, t
...
-		defer txnWrite.End()
 	}
+	txnWrite := kv.Write(trace)
+	defer txnWrite.End()
+	resp, err = put(ctx, txnWrite, p)
+	return resp, trace, err
+}
+
+func put(ctx context.Context, txnWrite mvcc.TxnWrite, p *pb.PutRequest) (resp *pb.PutResponse, err error) {
+	trace := traceutil.Get(ctx)
```
ahrtr, line 57: A separate trace as compared to the trace in `Put`? You may want to associate the trace with the ctx in `Put`

`ctx = context.WithValue(ctx, traceutil.TraceKey, trace)`

server/etcdserver/txn/txn.go
```diff
@@ -114,17 +121,19 @@ func DeleteRange(kv mvcc.KV, txnWrite mvcc.TxnWrite, dr *pb.DeleteRangeRequest)
...
 
-func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, txnRead mvcc.TxnRead, r *pb.RangeRequest) (*pb.RangeResponse, error) {
+func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, r *pb.RangeRequest) (*pb.RangeResponse, error) {
+	trace := traceutil.Get(ctx)
+	txnRead := kv.Read(mvcc.ConcurrentReadTxMode, trace)
+	defer txnRead.End()
+	return applyRange(ctx, lg, txnRead, r)
+}
+
+func applyRange(ctx context.Context, lg *zap.Logger, txnRead mvcc.TxnRead, r *pb.RangeRequest) (*pb.RangeResponse, error) {
```
ahrtr, line 131:
```suggestion
func range(ctx context.Context, lg *zap.Logger, txnRead mvcc.TxnRead, r *pb.RangeRequest) (*pb.RangeResponse, error) {
```

server/etcdserver/txn/txn.go
```diff
@@ -114,17 +121,19 @@ func DeleteRange(kv mvcc.KV, txnWrite mvcc.TxnWrite, dr *pb.DeleteRangeRequest)
 	return resp, nil
 }
 
-func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, txnRead mvcc.TxnRead, r *pb.RangeRequest) (*pb.RangeResponse, error) {
+func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, r *pb.RangeRequest) (*pb.RangeResponse, error) {
```
ahrtr, line 124: It seems that the `Range` will never be called. It isn't a big problem.

server/etcdserver/txn/txn.go
```diff
@@ -114,17 +121,19 @@ func DeleteRange(kv mvcc.KV, txnWrite mvcc.TxnWrite, dr *pb.DeleteRangeRequest)
...
+	trace := traceutil.Get(ctx)
```
ahrtr, line 125: ditto: associate the trace with the ctx if it's empty.

server/etcdserver/txn/txn.go
```diff
@@ -43,16 +41,23 @@ func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, t
...
-	val, leaseID := p.Value, lease.LeaseID(p.Lease)
-	if txnWrite == nil {
-		if leaseID != lease.NoLease {
-			if l := lessor.Lookup(leaseID); l == nil {
-				return nil, nil, lease.ErrLeaseNotFound
-			}
+	leaseID := lease.LeaseID(p.Lease)
+	if leaseID != lease.NoLease {
+		if l := lessor.Lookup(leaseID); l == nil {
+			return nil, nil, lease.ErrLeaseNotFound
```
ahrtr, line 47: Only perform this check in Put? Why not in `put`? Note that you will miss the check when put is called inside a txn.

Suggestion: move the check into function `put`

server/etcdserver/txn/txn.go
serathius: range is a keyword :P

server/etcdserver/txn/txn.go
serathius: What do you mean? Range is called in apply.go.

server/etcdserver/txn/txn.go
serathius: Done

server/etcdserver/txn/txn.go
serathius: Done

server/etcdserver/txn/txn.go
serathius: Please notice that I `put` function is only called via `applyTxn` and provided `writeTxn` is never nil. There is no need for `put` for checking lessor after separation from `Put`. 

To explain why it's done, please read the `Txn` code. Txn is executed in two phases, first with readTxn and second with writeTxn. It does all validation including checking the lease in the first phase, and only calls calls function `put` in second phase. It means that there is no need for `put` to validate the lease again.

fuweid:
> traffic.go:106: Requiring minimal 100.000000 qps for test results to be reliable, got 99.731074 qps

hmm. It's about `beforeWritebackBuf`. Tracked by https://github.com/etcd-io/etcd/issues/16370

server/etcdserver/txn/txn.go
ahrtr: There is no a standalone Range call coming from raft (apply.go) directly.

Previously the Range may be called in a TXN (applyTxn) or called from API directly. 

After this PR, the Range will only be called from API directly.

PR #16358, opened 2023-08-02
clientv3: remove the experimental gRPC API grpccredentials.Bundle

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/15145

The [credentials.Bundle](https://github.com/grpc/grpc-go/blob/8f496b2a9544fdceadb26b0d45392b1022d232d9/credentials/credentials.go#L180) is an experimental API. We should try to remove the dependency on such experimental gRPC API. 

Actually transport credential and per GRPC credentials are two different things, no need to get them bind together.

### DIFF
--- a/client/v3/client.go
+++ b/client/v3/client.go
@@ -69,7 +69,7 @@ type Client struct {
 	Username string
 	// Password is a password for authentication.
 	Password        string
-	authTokenBundle credentials.Bundle
+	authTokenBundle credentials.PerRPCCredentialsBundle
 
 	callOpts []grpc.CallOption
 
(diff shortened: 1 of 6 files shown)


jmhbnz (approved):
Makes sense, and good discovery @chaochn47.

Should this change be backported? Both prior release branches also use this api.

chaochn47 (approved):
LGTM. Thanks @ahrtr for cleaning it up.

ahrtr:
Right, usually we don't proactively bump dependencies for stable release (e.g. 3.5 and 3.4) unless there are major CVE or bugs.

fuweid (approved):
LGTM

PR #16371, opened 2023-08-04
server: Separate txnRead from txnWrite

serathius:
cc @ahrtr @ptabor @jmhbnz @chaochn47 @fuweid

PR #16372, opened 2023-08-04
server: Refactor checkTxn into single function handling all request types

serathius:
cc @ahrtr @ptabor @jmhbnz @chaochn47 @fuweid

### DIFF
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -267,11 +267,15 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
 		},
 		"compare",
 	)
-	err := checkTxn(ctx, txnRead, rt, isWrite, lessor, txnPath)
+	if isWrite {
+		trace.AddField(traceutil.Field{Key: "read_only", Value: false})
+	}
+	_, err := checkTxn(txnRead, rt, lessor, txnPath)
 	if err != nil {
 		txnRead.End()
 		return nil, nil, err
 	}
+	trace.Step("check requests")
 	// When executing mutable txnWrite ops, etcd must hold the txnWrite lock so
 	// readers do not see any intermediate results. Since writes are
 	// serialized on the raft loop, the revision in the read view will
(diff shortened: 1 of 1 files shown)


server/etcdserver/txn/txn.go
```diff
@@ -446,25 +420,32 @@ func checkRequestRange(rv mvcc.ReadView, reqOp *pb.RequestOp) error {
...
+			txns, err = checkTxn(rv, tv.RequestTxn, lessor, txnPath[1:])
+			txns += 1
+		default:
+			// empty union
 		}
-		if err := f(rv, req); err != nil {
+		if err != nil {
 			return 0, err
 		}
+		txnCount += txns
```
fuweid, line 447: just my two cents: the L447 and L448 should be moved into the case `RequestTxn`, which might be easy to read.

fuweid (commented):
LGTM

NOTE: It might be conflict with #16371. The change looks good!

server/etcdserver/txn/txn.go
serathius: Done

fuweid (approved):
LGTM

server/etcdserver/txn/txn.go
```diff
@@ -256,11 +256,15 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
...
-	err := checkTxn(ctx, txnRead, rt, isWrite, lessor, txnPath)
+	if isWrite {
+		trace.AddField(traceutil.Field{Key: "read_only", Value: false})
+	}
+	_, err := checkTxn(txnRead, rt, lessor, txnPath)
 	if err != nil {
 		txnRead.End()
 		return nil, nil, err
 	}
+	trace.Step("check requests")
```
ahrtr, line 267: Minor comment: put the trace into the function `checkTxn`. All the existing trace items are bound with the operations/functions.

jmhbnz (approved):
LGTM

server/etcdserver/txn/txn.go
serathius: Yes, but `checkTxn` is recursive, which means it would call `trace.Step("check requests")` for each subtransaction which is not what we want.

serathius:
> This looks like a bug fix instead of refactoring. For a non-readonly txn, it always checks each request in the TXN using checkRequestPut, which isn't correct.

Not sure I understand, please read the next line.

https://github.com/etcd-io/etcd/blob/2888a95d5d4f20e7e852214e888d79ead642300b/server/etcdserver/txn/txn.go#L305-L307

ahrtr:
Right, both `checkRequestRange`  and `checkRequestPut` may be tried for each request.

ahrtr (approved):

serathius:
Tested this PR with https://github.com/etcd-io/etcd/pull/16515

PR #16381, opened 2023-08-07
fix: bit operation

Jeremy-Run:
Change the remainder operation to bit operation to improve the calculation speed


ahrtr:
thx for the PR, but we are deprecating v2store. Please read https://github.com/etcd-io/etcd/issues/12913

Issue #16322, opened 2023-07-27
"github.com/gyuho/gocovmerge" in tools/mod/go.mod depencies is not a public respository


chaochn47:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request, support requests should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

The impact but minimal but annoying. IDE indexing failed.. 

### What did you expect to happen?

I expect this dependency should be under `etcd-io` org instead of located in a previous maintainer's repository. 

Somehow the action workflows have access to the repository but it's a potential risk. 

### How can we reproduce it (as minimally and precisely as possible)?

```
go list -modfile=tools/mod/go.mod -m -json -mod=mod all

go: github.com/gyuho/gocovmerge@v0.0.0-20171205171859-50c7e6afd535: invalid version: git ls-remote -q origin in /Users/chaochn/go/pkg/mod/cache/vcs/516c73b4fb2f2a3afac6e6b2b379be19f37566c31ceb2ace7992c16357e2c4f8: exit status 128:
	fatal: could not read Username for 'https://github.com': terminal prompts disabled
Confirm the import path was entered correctly.
If this is a private repository, see https://golang.org/doc/faq#git_https for additional information.
go: github.com/gyuho/gocovmerge@v0.0.0-20171205171859-50c7e6afd535: invalid version: git ls-remote -q origin in /Users/chaochn/go/pkg/mod/cache/vcs/516c73b4fb2f2a3afac6e6b2b379be19f37566c31ceb2ace7992c16357e2c4f8: exit status 128:
	fatal: could not read Username for 'https://github.com': terminal prompts disabled
Confirm the import path was entered correctly.
If this is a private repository, see https://golang.org/doc/faq#git_https for additional information.
```

### Etcd version (please run commands below)

It's on main branch. But I assume all the other release branch is also impacted. 

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
[...]

PR #16384, opened 2023-08-07
replace github.com/gyuho/gocovmerge with github.com/wadey/gocovmerge

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/16322

`github.com/gyuho/gocovmerge` has already been removed or turned into a private repo. I believe it's also forked from `github.com/wadey/gocovmerge`. So replace it with the original repo `github.com/wadey/gocovmerge`.


jmhbnz (approved):
LGTM

Note: Found a copy of the repo in waybackmachine: https://web.archive.org/web/20201023134702/https://github.com/gyuho/gocovmerge 

Looks like at that point in 2020 it was just two commits ahead of `wadey/gocovmerge`.

ahrtr:
> Edit - There are a number of forks of that original repo and this one https://github.com/alexfalkowski/gocovmerge does look to be getting more maintenance so we could consider that option also.

Good point! Updated.

PR #16385, opened 2023-08-07
client: Added new session option to set timeout for session creation

### LINKED ISSUE #14631: concurrency.NewSession hang after etcd server is killed with SIGSTOP(19)
### What happened?

[`concurrency.NewSession`](https://github.com/etcd-io/etcd/blob/55500416335e959e347d368c7f8a7a0229db3f6a/client/v3/concurrency/session.go#L38) hang after etcd server is kill by SIGSTOP(19)

`NewSession` can return error after server is killed.

1. start three or more etcd server nodes.
2. run `main` with following codes.
3. kill -19 `pidof etcd leader`

```
package main

import (
	"fmt"
	"time"

	"github.com/pingcap/log"
	"go.etcd.io/etcd/client/v3/concurrency"
	"go.uber.org/zap"
)

func initEtcdClient() *clientv3.Client {
	var client *clientv3.Client
	var err error
	endpoints := []string{"172.16.5.32:2379", "172.16.5.32:2382", "172.16.5.32:2384"}
	client, err = clientv3.New(clientv3.Config{
		Endpoints:            endpoints,
		DialTimeout:          5 * time.Second,
		DialKeepAliveTimeout: 5 * time.Second,
	})
	if err != nil {
		fmt.Printf("create client fail:%v\\n", err)
		log.Panic("create client fail", zap.Error(err))
	}
	return client
}

func main() {

	number := 0
	client := initEtcdClient()
	for {
		log.Info("create session begin.", zap.Int("time", number))
		s, err := concurrency.NewSession(client)
		if err != nil {
			log.Panic("create client fail", zap.Error(err))
		}
		log.Info("create session finish.", zap.Int("time", number))
		s.Close()
		number++
		time.Sleep(time.Second)
	}
}

``` 

If re-create etcd client after `kill -19`, it can return error. However, in our application, the client is created at the beginning and stored to use in the while lifecycle of the application.

<details>
go.etcd.io/etcd/client/v3 v3.5.5
```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

<details>

</details>

<details>

```console
# paste output here

# paste output here
```

</details>

### Relevant log output

_No response_

AngstyDuck:
When communicating via gRPC, etcd sets the gRPC option WaitForReady = true to minimize client request error responses due to transient failures. But because of this option, the creation of new sessions hang without errors when all nodes could not be contacted by the client. This new session option allows client to receive an error if the nodes cannot be contacted after the specified amount of time.

This seeks to resolve #14631 

### DIFF
--- a/client/v3/concurrency/session.go
+++ b/client/v3/concurrency/session.go
@@ -40,14 +40,24 @@ type Session struct {
 // NewSession gets the leased session for a client.
 func NewSession(client *v3.Client, opts ...SessionOption) (*Session, error) {
 	lg := client.GetLogger()
-	ops := &sessionOptions{ttl: defaultSessionTTL, ctx: client.Ctx()}
+	ops := &sessionOptions{
+		ttl: defaultSessionTTL,
+		ctx: client.Ctx(),
+	}
 	for _, opt := range opts {
 		opt(ops, lg)
 	}
 
+	sessionCreationCtx := ops.ctx
+	if ops.creationTimeout > 0 {
+		var creationCancel context.CancelFunc
+		sessionCreationCtx, creationCancel = context.WithTimeout(ops.ctx, ops.creationTimeout)
+		defer creationCancel()
+	}
+
 	id := ops.leaseID
 	if id == v3.NoLease {
-		resp, err := client.Grant(ops.ctx, int64(ops.ttl))
+		resp, err := client.Grant(sessionCreationCtx, int64(ops.ttl))
 		if err != nil {
 			return nil, err
 		}
@@ -115,9 +125,10 @@ func (s *Session) Close() error {
 }
 
 type sessionOptions struct {
-	ttl     int
-	leaseID v3.LeaseID
-	ctx     context.Context
+	ttl             int
+	leaseID         v3.LeaseID
+	ctx             context.Context
+	creationTimeout time.Duration
 }
 
 // SessionOption configures Session.
@@ -131,6 +142,19 @@ func WithTTL(ttl int) SessionOption {
 			so.ttl = ttl
 		} else {
 			lg.Warn("WithTTL(): TTL should be > 0, preserving current TTL", zap.Int64("current-session-ttl", int64(so.ttl)))
+		}
+	}
+}
+
+// WithCreationTimeout configures the timeout for creating a new session.
+// If timeout is <= 0, no timeout will be used, and the creating new session
+// will be blocked forever until the etcd server is reachable.
+func WithCreationTimeout(timeout time.Duration) SessionOption {
+	return func(so *sessionOptions, lg *zap.Logger) {
+		if timeout > 0 {
+			so.creationTimeout = timeout
+		} else {
+			lg.Warn("WithCreationTimeout(): timeout should be > 0, preserving current timeout", zap.Int64("current-session-timeout", int64(so.creationTimeout)))
 		}
 	}
 }
--- a/CHANGELOG/CHANGELOG-3.7.md
+++ b/CHANGELOG/CHANGELOG-3.7.md
@@ -12,6 +12,7 @@ Previous change logs can be found at [CHANGELOG-3.6](https://github.com/etcd-io/
 ### Package `clientv3`
 
 - [Make the etcd client creation non-blocking](https://github.com/etcd-io/etcd/pull/21942): etcd no longer honors the deprecated `grpc.WithBlock` dial option. To preserve the previous blocking behavior when needed, follow the guidance in grpc-go's [anti-patterns documentation](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#especially-bad-using-deprecated-dialoptions).
+- Add a timeout for concurrent session creation.
 
 ### etcd server
 
--- a/tests/integration/cluster_test.go
+++ b/tests/integration/cluster_test.go
@@ -26,11 +26,14 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/connectivity"
 
 	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/client/v3/concurrency"
 	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/integration"
+	clientv3test "go.etcd.io/etcd/tests/v3/integration/clientv3"
 )
 
 func init() {
@@ -525,4 +528,88 @@ func TestConcurrentMoveLeader(t *testing.T) {
 	_, err = c.Members[0].Client.MemberRemove(t.Context(), removeID)
 	require.NoError(t, err)
 	<-done
+}
+
+// TestCreationTimeout checks that the option WithCreationTimeout
+// sets a timeout for the creation of new sessions in case the cluster
+// shuts down
+func TestCreationTimeout(t *testing.T) {
+	integration.BeforeTest(t)
+
+	// create new cluster
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
+
+	// create new client
+	cli, err := integration.NewClient(t, clientv3.Config{Endpoints: clus.Client(0).Endpoints()})
+	require.NoError(t, err)
+	defer cli.Close()
+
+	// ensure the connection is established.
+	clientv3test.MustWaitPinReady(t, cli)
+
+	// terminating the cluster
+	clus.Terminate(t)
+
+	// NewSession must fail with context.DeadlineExceeded once the creation
+	// timeout elapses while the cluster is unreachable, instead of blocking
+	// forever due to gRPC's WaitForReady option.
+	_, err = concurrency.NewSession(cli, concurrency.WithCreationTimeout(3*time.Second))
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+
+	// the deadline must have been caused by loss of connectivity: the
+	// underlying gRPC connection should no longer report a Ready state.
+	require.NotEqual(t, connectivity.Ready, cli.ActiveConnection().GetState())
+}
+
+// TestTimeoutDoesntAffectSubsequentConnections checks that the option WithCreationTimeout
+// is only used when Session is created
+func TestTimeoutDoesntAffectSubsequentConnections(t *testing.T) {
+	integration.BeforeTest(t)
+
+	// create new cluster
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
+	clus.Members[0].KeepDataDirTerminate = true
+
+	// create new client
+	cli, err := integration.NewClient(t, clientv3.Config{Endpoints: clus.Client(0).Endpoints()})
+	require.NoError(t, err)
+	defer cli.Close()
+
+	// ensure the connection is established.
+	clientv3test.MustWaitPinReady(t, cli)
+
+	s, err := concurrency.NewSession(cli, concurrency.WithCreationTimeout(1*time.Second))
+	require.NoError(t, err)
+
+	// terminating the cluster
+	clus.Members[0].Terminate(t)
+
+	errorc := make(chan error)
+	defer close(errorc)
+	go func() {
+		_, err := cli.Put(s.Ctx(), "sample_key", "sample_value", clientv3.WithLease(s.Lease()))
+		errorc <- err
+	}()
+
+	select {
+	case err := <-errorc:
+		t.Fatalf("Operation put should be blocked forever when the server is unreachable: %v", err)
+	// if Put operation is blocked beyond the timeout specified using WithCreationTimeout,
+	// that timeout is not used by the Put operation
+	case <-time.After(2 * time.Second):
+	}
+
+	// restarting and ensuring that the Put operation will eventually succeed
+	clus.Members[0].Restart(t)
+	clus.Members[0].WaitOK(t)
+	select {
+	case err := <-errorc:
+		if err != nil {
+			t.Errorf("Put failed: %v", err)
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatal("Put function hung even after restarting cluster")
+	}
 }


AngstyDuck:
> What if the etcdserver shutdown after the connection has already established? Will the session timeout or be blocked waiting for the cluster to recover? The expected behaviour should be the latter. Can you add a test case to verify it?

Thanks for pointing this out, I've fixed the bug and added another test case to verify this. Do let me know what you think.

AngstyDuck:
Hey @ahrtr, any updates regarding the changes I've made? I'm not quite sure how to write a test that verifies that hanging would occur, so I have verified the context object instead.

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,76 @@ func TestSessionCtx(t *testing.T) {
 	}
 	assert.Equal(t, childCtx.Err(), context.Canceled)
 }
+
+// TestSessionCreationTimeout checks that the option WithSessionCreationTimeout
+// sets a timeout for the creation of new sessions
+func TestSessionCreationTimeout(t *testing.T) {
```
fuweid, line 118: I think this case covered the block case. 

If we need to confirm that the underlying grpc is retrying to get connection, as far as I know, we can use logger observer to expect the log content like this.

<details>

<summary> example </summary>

```diff
diff --git a/tests/framework/integration/testing.go b/tests/framework/integration/testing.go
index b0580dcd6..4a839668d 100644
--- a/tests/framework/integration/testing.go
+++ b/tests/framework/integration/testing.go
@@ -20,6 +20,7 @@ import (

        grpc_logsettable "github.com/grpc-ecosystem/go-grpc-middleware/logging/settable"
        "github.com/stretchr/testify/require"
+       "go.uber.org/zap"
        "go.uber.org/zap/zapcore"
        "go.uber.org/zap/zapgrpc"
        "go.uber.org/zap/zaptest"
@@ -30,6 +31,7 @@ import (
        "go.etcd.io/etcd/client/pkg/v3/verify"
        clientv3 "go.etcd.io/etcd/client/v3"
        "go.etcd.io/etcd/server/v3/embed"
+       "go.etcd.io/etcd/tests/v3/framework/testutils"
 )

 var grpc_logger grpc_logsettable.SettableLoggerV2
@@ -131,6 +133,26 @@ func BeforeTest(t testutil.TB, opts ...TestOption) {
        os.Chdir(t.TempDir())
 }

+func ClientGRPCLoggerObserver(t testutil.TB) *testutils.LogObserver {
+       level := zapcore.InfoLevel
+
+       obCore, logOb := testutils.NewLogObserver(level)
+
+       options := zaptest.WrapOptions(
+               zap.WrapCore(func(oldCore zapcore.Core) zapcore.Core {
+                       return zapcore.NewTee(oldCore, obCore)
+               }),
+       )
+
+       grpc_logger.Set(
+               zapgrpc.NewLogger(
+                       zaptest.NewLogger(t, zaptest.Level(level), options).
+                               Named("grpc-observer"),
+               ),
+       )
+       return logOb
+}
+
 func assertInTestContext(t testutil.TB) {
        if !insideTestContext {
                t.Errorf("the function can be called only in the test context. Was integration.BeforeTest() called ?")
diff --git a/tests/integration/clientv3/concurrency/session_test.go b/tests/integration/clientv3/concurrency/session_test.go
index 247c777df..b9410a93d 100644
--- a/tests/integration/clientv3/concurrency/session_test.go
+++ b/tests/integration/clientv3/concurrency/session_test.go
@@ -119,31 +119,30 @@ func TestSessionCreationTimeout(t *testing.T) {
        integration2.BeforeTest(t)

        // create new cluster
-       clus := integration2.NewCluster(t, &integration2.ClusterConfig{
-               Size: 3,
-       })
-
-       eps := []string{clus.Members[0].GRPCURL(), clus.Members[1].GRPCURL(), clus.Members[2].GRPCURL()}
-       lead := clus.WaitLeader(t)
+       clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 1})

        // create new client
-       cli, err := integration2.NewClient(t, clientv3.Config{Endpoints: []string{eps[lead]}})
+       cli, err := integration2.NewClient(t, clientv3.Config{Endpoints: []string{clus.Members[0].GRPCURL()}})
        if err != nil {
+               clus.Terminate(t)
                t.Fatal(err)
        }
        defer cli.Close()

-       // wait for eps[lead] to be pinned
+       // ensure the connection is established.
        clientv3test.MustWaitPinReady(t, cli)

-       // add all eps to list, so that when the original pined one fails
-       // the client can switch to other available eps
-       cli.SetEndpoints(eps...)
-
+       // terminating the cluster
        clus.Terminate(t)

-       _, err = concurrency.NewSession(cli, concurrency.WithSessionCreationTimeout(50))
+       // override the grpc logger
+       logOb := integration2.ClientGRPCLoggerObserver(t)
+
+       _, err = concurrency.NewSession(cli, concurrency.WithSessionCreationTimeout(3000))
        assert.Equal(t, err, context.DeadlineExceeded)
+
+       _, err = logOb.Expect(context.Background(), "Subchannel Connectivity change to TRANSIENT_FAILURE", 3)
+       assert.Nil(t, err)
 }

 // TestTimeoutDoesntAffectSubsequentConnections checks that the option WithSessionCreationTimeout
```

</details>

client/v3/concurrency/session.go
```diff
@@ -115,9 +126,10 @@ func (s *Session) Close() error {
 }
 
 type sessionOptions struct {
-	ttl     int
-	leaseID v3.LeaseID
-	ctx     context.Context
+	ttl                    int
+	leaseID                v3.LeaseID
+	ctx                    context.Context
+	sessionCreationTimeout int
```
fuweid, line 132: Suggest to remove session prefix because the struct is named by sessionOptions already.

client/v3/concurrency/session.go
```diff
@@ -135,6 +147,18 @@ func WithTTL(ttl int) SessionOption {
 	}
 }
 
+// WithSessionCreationTimeout configures the timeout for creating a new session
+// in milliseconds. If timeout is <= 0, no timeout will be used.
+func WithSessionCreationTimeout(timeout int) SessionOption {
```
fuweid, line 152: I think we can use time.Duration as type so that interface is more clear.

tests/integration/clientv3/concurrency/session_test.go
AngstyDuck: gotcha, applying diff and removing the subsequent test case :+1:

AngstyDuck:
Hey @fuweid, thanks for your review. I've addressed them accordingly, so do let me know what you think :pray:

fuweid (approved):
LGTM

client/v3/concurrency/session.go
```diff
@@ -135,6 +145,18 @@ func WithTTL(ttl int) SessionOption {
 	}
 }
 
+// WithCreationTimeout configures the timeout for creating a new session.
+// If timeout is <= 0, no timeout will be used.
```
ahrtr, line 149:
```suggestion
// WithCreationTimeout configures the timeout for creating a new session.
// If timeout is <= 0, no timeout will be used, and the creating new session
// will be blocked forever until the etcd server is reachable.
```

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,35 @@ func TestSessionCtx(t *testing.T) {
...
+	clus.Terminate(t)
+
+	// override the grpc logger
+	logOb := integration2.ClientGRPCLoggerObserver(t)
+
+	_, err = concurrency.NewSession(cli, concurrency.WithCreationTimeout(3000*time.Millisecond))
+	assert.Equal(t, err, context.DeadlineExceeded)
+
+	_, err = logOb.Expect(context.Background(), "Subchannel Connectivity change to TRANSIENT_FAILURE", 3)
+	assert.Nil(t, err)
```
ahrtr, line 145: Please add a case to verify that the default behaviour isn't affected. `When no CreationTimeout is specified, then the creating new session will be blocked forever until the etcd server is reachable.`

ahrtr:
Please also squash the commits, thx

AngstyDuck:
Hey @ahrtr, I've added a new test case where we verify that the default behavior isn't affected if an operation hangs beyond the timeout specified by the new option `CreationTimeout`. Do let me know if this works for you :pray:

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,74 @@ func TestSessionCtx(t *testing.T) {
...
+		_, _ = cli.Put(s.Ctx(), "sample_key", "sample_value", clientv3.WithLease(s.Lease()))
+	}()
+
+	select {
+	case <-donec:
+		t.Fatal("operation timed out using WithCreationTimeout")
+	// if Put operation is blocked beyond the timeout specified using WithCreationTimeout,
+	// that timeout is not used by the Put operation
+	case <-time.After(2 * time.Second):
+	}
```
ahrtr, line 184: Please start the etcdserver again, and verify that eventually the `cli.Put` will succeed.

AngstyDuck:
Hey @ahrtr, I've altered the test `TestTimeoutDoesntAffectSubsequentConnections` where I've restarted the test and check that the ongoing `Put` succeeds. Unfortunately this has made the test flaky (53.29% failing after 152 runs) and I'm unsure how to proceed. The log is as attached here: [log.txt](https://github.com/etcd-io/etcd/files/12540810/log.txt)

1. My first choice would be to omit the testing of this specific logic until we've resolved the flakiness issue. If we go with this, I'll be creating the issue and would like to look into this myself.
2. My second choice would be to resolve the flakiness before pushing this out to main, but the scope of this PR may be inflated.

How do you think we should proceed?

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,83 @@ func TestSessionCtx(t *testing.T) {
...
+
+	// ensure the connection is established.
+	clientv3test.MustWaitPinReady(t, cli)
+
+	s, err := concurrency.NewSession(cli, concurrency.WithCreationTimeout(1*time.Second))
+
+	// terminating the cluster
+	clus.Members[0].Terminate(t)
+
+	donec := make(chan struct{})
```
ahrtr, line 173:
```suggestion
	donec := make(chan struct{})
	errorc := make(chan error)
```

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,83 @@ func TestSessionCtx(t *testing.T) {
...
+	go func() {
+		defer close(donec)
+		_, _ = cli.Put(s.Ctx(), "sample_key", "sample_value", clientv3.WithLease(s.Lease()))
```
ahrtr, line 176:
```suggestion
		if _, err := cli.Put(s.Ctx(), "sample_key", "sample_value", clientv3.WithLease(s.Lease())); err != nil {
		    errorc <- err
		}    
```

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,83 @@ func TestSessionCtx(t *testing.T) {
...
+		t.Fatal("operation timed out using WithCreationTimeout")
+	// if Put operation is blocked beyond the timeout specified using WithCreationTimeout,
+	// that timeout is not used by the Put operation
+	case <-time.After(2 * time.Second):
+	}
+
+	// restarting and ensuring that the Put operation will eventually succeed
+	clus.Members[0].Restart(t)
+	select {
+	case <-donec:
```
ahrtr, line 190:
```suggestion
	case <-donec:
	case err := <- errorc:
	    t.Errorf("Put failed: %v", err)
```

AngstyDuck:
Hey @ahrtr, I've added the logging as suggested, and the log returned was `session_test.go:198: Put failed: etcdserver: requested lease not found`. This error is raised with or without the usage of the new option during session creation, so this should probably be a separate issue.

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,88 @@ func TestSessionCtx(t *testing.T) {
...
+	case err := <-errorc:
+		t.Errorf("Put failed: %v", err)
```
ahrtr, line 195: There is a potential issue here. When the Put finishes, both donec and errorc might be ready to be read, then it's uncertain which case is selected.

So I suggest to remove donec, and always feed the err (returned by Put) into errroc, even it's nil.

Sorry for the confusion.

AngstyDuck:
Hey @ahrtr no worries, I've removed donec as instructed. However I'm still facing the same error `session_test.go:199: Put failed: etcdserver: requested lease not found` from errorc, with or without the new session creation option. How should I proceed?

ahrtr:
> However I'm still facing the same error `session_test.go:199: Put failed: etcdserver: requested lease not found` from errorc, with or without the new session creation option. How should I proceed?

- I will take a closer look sometime next week.
- The workflow checks are all green. You only occasionally see the error?

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,88 @@ func TestSessionCtx(t *testing.T) {
...
+	errorc := make(chan error)
+	go func() {
+		defer close(donec)
+		_, err := cli.Put(s.Ctx(), "sample_key", "sample_value", clientv3.WithLease(s.Lease()))
+		errorc <- err
+	}()
+
+	select {
+	case <-donec:
+		t.Fatal("operation timed out using WithCreationTimeout")
```
ahrtr, line 183: I meant to remove the `donec` completely (remove line 173).
```suggestion
	case err := <-errorc:
		t.Fatalf("Operation put should be blocked forever when the server is unreachable: %v", err)
```

AngstyDuck: apologies for the confusion, I've removed it accordingly

AngstyDuck:
> The workflow checks are all green. You only occasionally see the error?

Hmm no, oddly this happens every time I've ran the test `TestTimeoutDoesntAffectSubsequentConnections`. I'll see if I can figure it out as well.

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,86 @@ func TestSessionCtx(t *testing.T) {
...
+}
+
+// TestTimeoutDoesntAffectSubsequentConnections checks that the option WithCreationTimeout
+// is only used when Session is created
+func TestTimeoutDoesntAffectSubsequentConnections(t *testing.T) {
+	integration2.BeforeTest(t)
+
+	// create new cluster
+	clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
```
ahrtr, line 155:
```suggestion
	defer clus.Terminate(t)
	clus.Members[0].KeepDataDirTerminate = true
```

tests/integration/clientv3/concurrency/session_test.go
```diff
@@ -111,3 +112,86 @@ func TestSessionCtx(t *testing.T) {
...
+	select {
+	case err := <-errorc:
+		t.Fatalf("Operation put should be blocked forever when the server is unreachable: %v", err)
+	// if Put operation is blocked beyond the timeout specified using WithCreationTimeout,
+	// that timeout is not used by the Put operation
+	case <-time.After(2 * time.Second):
+	}
+
+	// restarting and ensuring that the Put operation will eventually succeed
+	clus.Members[0].Restart(t)
```
ahrtr, line 188:
```suggestion
	clus.Members[0].Restart(t)
	clus.Members[0].WaitOK(t)
```

ahrtr:
The reason why the workflow is green is the test was skipped.
```
=== SKIP: integration/clientv3/concurrency TestTimeoutDoesntAffectSubsequentConnections (0.05s)
    leak.go:102: Found leaked goroutined BEFORE test appears to have leaked :
        github.com/soheilhy/cmux.muxListener.Accept(...)
        	/go/pkg/mod/github.com/soheilhy/cmux@v0.1.5/cmux.go:262
        net/http.(*Server).Serve(0xc00001c0f0, {0x1992220, 0xc0003b9820})
        	/__t/go/1.20.8/x64/src/net/http/server.go:3059 +0x5a7
        net/http/httptest.(*Server).goServe.func1()
        	/__t/go/1.20.8/x64/src/net/http/httptest/server.go:310 +0xbc
        created by net/http/httptest.(*Server).goServe
        	/__t/go/1.20.8/x64/src/net/http/httptest/server.go:308 +0xa7
```

When terminating the member, the data is removed automatically. You need to keep the data dir so that the lease is still valid after restarting the member. Please apply the suggested change, and try again.

AngstyDuck:
Hey @ahrtr, thanks for your guidance so far. I've done what you've suggested, and also moved the test cases to `tests/integration/cluster_test.go` instead. I believe the test cases are no longer skipped, and is running without issues. Do let me know if this works for you :+1:

ahrtr (approved):

fuweid (approved):
LGTM

jmhbnz (commented):
Hey @AngstyDuck - Can you please rebase this pr to address conflicts, thanks!

AngstyDuck:
Hey @jmhbnz, I've rebased the branch :+1:  thanks!

lhmzhou (approved):

jmhbnz:
Discussed during sig-etcd triage meeting, @AngstyDuck can you please rebase this pr once more?  Thanks!

tests/framework/integration/testing.go
```diff
@@ -29,6 +30,7 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/embed"
 	gofail "go.etcd.io/gofail/runtime"
```
jmhbnz, line 32: Looks like a minor issue here causing static analysis to fail, suggested fix below:

```diff
  	"go.etcd.io/etcd/client/pkg/v3/verify"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/embed"
-	gofail "go.etcd.io/gofail/runtime"
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
+	gofail "go.etcd.io/gofail/runtime"
```

AngstyDuck:
Hey @jmhbnz, the linting errors have been fixed, do let me know if you need anything else 👍

jmhbnz (approved):
LGTM - Thanks @AngstyDuck and apologies this pr has been in review for so long.

jmhbnz:
@AngstyDuck to be safe we should rebase one final time before we merge.

a-nych:
Any updates on this?

fuweid:
ping @a-nych sorry for late update. Please rebase it. Thanks!

AngstyDuck:
Hey all, sorry for the delay, the following changes are ready to be merged. @fuweid @ahrtr @jmhbnz

AngstyDuck:
Bumping this PR @fuweid @ahrtr @jmhbnz

client/v3/concurrency/session.go
```diff
@@ -45,9 +45,15 @@ func NewSession(client *v3.Client, opts ...SessionOption) (*Session, error) {
 		opt(ops, lg)
 	}
 
+	var cancel context.CancelFunc
+	sessionCreationCtx := ops.ctx
+	if ops.creationTimeout > 0 {
+		sessionCreationCtx, cancel = context.WithTimeout(ops.ctx, ops.creationTimeout)
+	}
```
ahrtr: The `cancel` is overwritten by the following the other `cancel`
```suggestion
	var creationCancel context.CancelFunc
	sessionCreationCtx := ops.ctx
	if ops.creationTimeout > 0 {
		sessionCreationCtx, creationCancel = context.WithTimeout(ops.ctx, ops.creationTimeout)
		defer creationCancel()
	}
```

ahrtr (changes requested):

ahrtr:
Pls also add a changelog item under https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.7.md#package-clientv3, something like below,
- Add a timeout for concurrent session creation

AngstyDuck:
Bumping this, will be working on it over the next few weeks, apologies for the delay 🙏

Issue #15956, opened 2023-05-25
Consider migrating all etcd unit, integration, e2e and robustness tests to containers


jmhbnz:
### What would you like to be added?

This is a sub issue under #15951.

Recent pull request https://github.com/etcd-io/etcd/pull/15927 introduced running our arm64 ci jobs in containers to prevent leaking test processes causing issues on our self hosted runners.

The github standard runners we use for non arm64 ci jobs do not have the cleanup issue as our self hosted ARM runners do so it isn't absolutely necessary to migrate all workflows into containers, however we should consider doing it on the basis that we have one unified way to run all ci workflows, and accordingly less cognitive burden in relation to these jobs.

This issue should be used to initially gather consensus from the community before proceeding with any changes. 

If we do decide to migrate all workflows, then eventually `robustness-template-arm64.yaml` and `robustness-template.yaml` should be combined back into one YAML file.

Additionally as part of this task we should ensure all ci workflows and our new devcontainer (https://github.com/etcd-io/etcd/blob/main/.devcontainer/devcontainer.json) are using a common base image and we have made a conscious decision about why that base image is preferred, taking into account factors like CVE data and update cadence.

### Why is this needed?

- Reduce maintenance and cognitive burden. 
- Have one consistent approach for isolation within all etcd test suites and platforms.
- Increase portability of test infrastructure.

tao12345666333:
Could you please assign this issue to me?

jmhbnz:
Hi @tao12345666333 - Thanks for offering to help! This issue needs to wait until pr https://github.com/etcd-io/etcd/pull/15927 has been merged and had some time to soak, and is also one of the higher risk items in the overall plan https://github.com/etcd-io/etcd/issues/15951.

Given this would be your first contribution to etcd and the above limitations I would suggest picking up https://github.com/etcd-io/etcd/issues/15955 first to get familiar with etcd ci workflows. If you're keen to pick it up just let me know with a comment on it and I will be most happy to assign it to you 🙏🏻

tao12345666333:
Don't worry. I will pick up #15955

jmhbnz:
Hey @tao12345666333 would you still be interested in helping with this?  Now that we have given https://github.com/etcd-io/etcd/pull/15927 some time to soak I think we could start proposing to migrate other test suite workflows to containers.

sharathsivakumar:
@jmhbnz I would like to work on this. 
[...]

PR #16392, opened 2023-08-08
Add lazyfs support for arm64 robustness workflow

jmhbnz:
Recently in https://github.com/etcd-io/etcd/pull/14691 we introduced support for lazyfs for nightly `amd64` robustness tests.  With `arm64` at tier1 we need to ensure `arm64` nightlies also support lazyfs.

This was a little more complicated as our `arm64` nightlies run in containers for better isolation on our self hosted runners. In order to mount the lazyfs fuse based filesystem inside a container we need to run a privileged container to ensure the fuse device is passed through and mountable. 

This pull request is a pre-requisite before wider adoption of containers for etcd workflows can be progressed under https://github.com/etcd-io/etcd/issues/15956.

The good news is once this pull request merges we will be able to combine `robustness-template-arm64.yaml` with `robustness-template.yaml` and go back down to one workflow template for all robustness workflows.


jmhbnz (commented):
Note for reviewers: I initially had this pr set to run robustness nightly jobs `on: pull_request` for verification.  You can find that actions run here https://github.com/etcd-io/etcd/actions/runs/5801718679.

From what I could see it looks like the lazyfs tests were behaving as expected.

ahrtr (approved):

.github/workflows/robustness-template-arm64.yaml
```diff
@@ -26,7 +26,9 @@ jobs:
   test:
     timeout-minutes: 210
     runs-on: ${{ fromJson(inputs.runs-on) }}
-    container: golang:1.19-bullseye
+    container:
+      image: golang:1.19-bullseye
+      options: --privileged --cap-add SYS_ADMIN --device /dev/fuse
```
fuweid, line 31: With privileged, if the engine is docker, by default the engine will mount all the device from host into container. And we don't need to add cap SYS_ADMIN because it's ready privileged.

jmhbnz: Great catch, and yes it's `docker` on the two self hosted runners at the moment. Updated thanks!

fuweid (approved):
LGTM

Issue #16380, opened 2023-08-07
`checkIntervals` will not validate correctly in certain nested txn scenarios


bsbds:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request, support requests should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

In `checkIntervals` there's seems to be bug in checking nested txns. In the following loop it check sub txn request and append the deletes to `dels` at the end.
https://github.com/etcd-io/etcd/blob/25315dd1f2271025e8f1c6a1e7167ee4fde50688/server/etcdserver/api/v3rpc/key.go#L208-L245
Consider the following senerio(assume all requests are in the success branch): If a txn `S` contains two sub txn `A` and `B`. `A` contains a put, `B` contains a delete, and the put overlaps with the delete. Then `S[A(put), B(del)]` will return a ok but `S[B(del), A(put)` will return an error.

The reason this inconsistency exist is that the sub txn requests are checked in a specific order, and append the delete to `dels` in this order. And the put checks `dels` for overlapping. So if the put is checked before the delete, it will not know that it overlaps with that delete.

### What did you expect to happen?

Both `S[A(put), B(del)]` and `S[B(del), A(put)` should return an error. The deletes should be append to `dels` first before check the intersection of puts. 

### How can we reproduce it (as minimally and precisely as possible)?

It should be reproduced by the following code
```go
func TestCheckIntervals(t *testing.T) {
  innerPut := pb.TxnRequest{
    Success: []*pb.RequestOp{
      {
        Request: &pb.RequestOp_RequestPut{RequestPut: &pb.PutRequest{
          Key: []byte("k1"),
        }},
      },
    },
  }
  innerDel := pb.TxnRequest{
    Success: []*pb.RequestOp{
      {
        Request: &pb.RequestOp_RequestDeleteRange{RequestDeleteRange: &pb.DeleteRangeRequest{
          Key:      []byte("k0"),
          RangeEnd: []byte("k3"),
        }},
      },
    },
  }
  req1 := pb.TxnRequest{
[...]

PR #16395, opened 2023-08-09
etcdserver: fix `checkIntervals` will not validate correctly

bsbds:
Fixes: #16380


bsbds:
In the original comment https://github.com/etcd-io/etcd/issues/16380#issuecomment-1668834426, I said we should union the all the deletes before check any put. However, it turns out that the success and failure branches can safely overlap. I refactored the code to check for each pair of sub-txns.

ahrtr:
https://github.com/etcd-io/etcd/issues/16380#issuecomment-1710087981

PR #16397, opened 2023-08-10
2023-08-10 bump up dependencies

chaochn47:
Supersedes https://github.com/etcd-io/etcd/pull/16382 and https://github.com/etcd-io/etcd/pull/16383. 

https://github.com/etcd-io/etcd/pull/16387 and https://github.com/etcd-io/etcd/pull/16386 is no-op and should be closed because they are the indirect dependency of the above two packages.

https://github.com/etcd-io/etcd/pull/16389 is purely an indirect dependency. Since we have used the new gomerge library, I think the right way to upgrade this dependency is to wait for a new release of the library.  


jmhbnz (changes requested):
Thanks for managing it this week @chaochn47 🙏🏻 

Can you please separate to individual commit per dependency as mentioned here: https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#steps-to-bump-a-dependency

chaochn47:
> When you bump multiple dependencies in one PR, it's recommended to create a separate commit for each dependency. But it isn't a must; for example, you can get all dependencies bumping for the module go.etcd.io/etcd/tools/v3 included in one commit.

Since there are only two dependencies, is it okay to be included in the same commit? 

Breaking them down needs to redo everything :p which I would like to avoid.

ahrtr:
It's OK to combine the two commits, as the two dependencies are very similar and [both are part of the Go project](https://pkg.go.dev/golang.org/x).

thx @chaochn47

ahrtr (approved):

PR #16419, opened 2023-08-15
Add example config for TLS protocol version

iiamabby:
As discussed recently in #13506  it would be good if an example of the TLS protocol configuration was included in the sample etcd conf file


ahrtr (approved):
LGTM

thx

PR #16421, opened 2023-08-15
test: increase the timeout for the flaky test `TestCtlV3AuthCertCNWithWithConcurrentOperation`

ahrtr:
Recently the e2e test is flaky. I double checked two workflow failures, and both of them were pointing to the test case `TestCtlV3AuthCertCNWithWithConcurrentOperation`.  The [creating & deleting users goroutine](https://github.com/etcd-io/etcd/blob/e3a1d7fb3c713371a6f535346d7cb695d21f9f2f/tests/e2e/ctl_v3_auth_no_proxy_test.go#L88-L104) somehow couldn't finish its task in 30 seconds, but it's very close to success (already finished about 94 out of the 100 users).  

### DIFF
--- a/tests/e2e/ctl_v3_auth_no_proxy_test.go
+++ b/tests/e2e/ctl_v3_auth_no_proxy_test.go
@@ -136,7 +136,7 @@ func TestCtlV3AuthCertCNWithWithConcurrentOperation(t *testing.T) {
 		t.Fatalf("Unexpected error: %v", err)
 	case <-donec:
 		t.Log("All done!")
-	case <-time.After(30 * time.Second):
-		t.Fatal("Test case timeout after 20 seconds")
+	case <-time.After(40 * time.Second):
+		t.Fatal("Test case timeout after 40 seconds")
 	}
 }


jmhbnz (approved):
Thanks for fixing this. That test in `linux-386-e2e` has been quite annoying.

ahrtr:
> Thanks for fixing this. That test in `linux-386-e2e` has been quite annoying.

When I was debugging the test failure. The metrics output (returned from the `curl /metrics` endpoint in `TestConnectionMultiplexing`) is really annoying, it's too large which leads to the log being truncated, so each time I have to download the complete log. It would be a little better if we can disable the metrics output in the case `TestConnectionMultiplexing` (or just print the first couple of lines).

jmhbnz:
> The metrics output (returned from the `curl /metrics` endpoint in `TestConnectionMultiplexing`) is really annoying, it's too large which leads to the log being truncated, so each time I have to download the complete log. It would be a little better if we can disable the metrics output in the case `TestConnectionMultiplexing` (or just print the first couple of lines).

Raised https://github.com/etcd-io/etcd/issues/16422.  Can give this a try later this week.

ahrtr:
thx @jmhbnz

ahrtr:
> Is there any link about TestCtlV3AuthCertCNWithWithConcurrentOperation failure?

https://github.com/etcd-io/etcd/actions/runs/5865180206/job/15901605507

fuweid:
amd64 takes 16.17s. i386 seems slow than amd64...

`2023-08-15T08:36:01.8727557Z --- PASS: TestCtlV3AuthCertCNWithWithConcurrentOperation (16.17s)`

PR #16433, opened 2023-08-17
Fix unit tests and make targets to not fail in some environments

shyamjvs:
Fixes for:

- Some bash versions can't take empty list variables and fail today with error like this:
```
PASSES="e2e" ./scripts/test.sh 
Running with --race=true --cpu=4
Starting at: Thu Aug 17 20:18:19 UTC 2023

'e2e' started at Thu Aug 17 20:18:19 UTC 2023
./scripts/test.sh: line 130: RUN_ARG[@]: unbound variable
make: *** [test-e2e] Error 1
```

- Unit tests which hardcode `tail` path fail with errors like this when it's present at different path:
```
--- FAIL: TestExpectFuncTimeout (0.51s)
    expect_test.go:75: 
        	Error Trace:	/home/$USER/workspace/go/src/github.com/shyamjvs/etcd/pkg/expect/expect_test.go:75
        	Error:      	Error "unexpected exit code [143] after running [/bin/tail -f /dev/null]" does not contain "unexpected exit code [143] after running [/usr/bin/tail -f /dev/null]"
        	Test:       	TestExpectFuncTimeout
```

### DIFF
--- a/pkg/expect/expect_test.go
+++ b/pkg/expect/expect_test.go
@@ -72,7 +72,7 @@ func TestExpectFuncTimeout(t *testing.T) {
 	}
 
 	err = ep.Close()
-	require.ErrorContains(t, err, "unexpected exit code [143] after running [/usr/bin/tail -f /dev/null]")
+	require.ErrorContains(t, err, "unexpected exit code [143]")
 	require.Equal(t, 143, ep.exitCode)
 }
 
@@ -89,7 +89,7 @@ func TestExpectFuncExitFailure(t *testing.T) {
 	_, err = ep.ExpectFunc(ctx, func(s string) bool {
 		return strings.Contains(s, "something entirely unexpected")
 	})
-	require.ErrorContains(t, err, "unexpected exit code [1] after running [/usr/bin/tail -x]")
+	require.ErrorContains(t, err, "unexpected exit code [1]")
 	require.Equal(t, 1, ep.exitCode)
 }
 
@@ -106,7 +106,7 @@ func TestExpectFuncExitFailureStop(t *testing.T) {
 	_, err = ep.ExpectFunc(ctx, func(s string) bool {
 		return strings.Contains(s, "something entirely unexpected")
 	})
-	require.ErrorContains(t, err, "unexpected exit code [1] after running [/usr/bin/tail -x]")
+	require.ErrorContains(t, err, "unexpected exit code [1]")
 	exitCode, err := ep.ExitCode()
 	require.Equal(t, 1, exitCode)
 	require.NoError(t, err)
@@ -115,7 +115,7 @@ func TestExpectFuncExitFailureStop(t *testing.T) {
 		t.Fatal(err)
 	}
 	err = ep.Close()
-	require.ErrorContains(t, err, "unexpected exit code [1] after running [/usr/bin/tail -x]")
+	require.ErrorContains(t, err, "unexpected exit code [1]")
 	exitCode, err = ep.ExitCode()
 	require.Equal(t, 1, exitCode)
 	require.NoError(t, err)
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -103,8 +103,8 @@ function build_pass {
 function run_unit_tests {
   local pkgs="${1:-./...}"
   shift 1
-  # shellcheck disable=SC2086
-  GOLANG_TEST_SHORT=true go_test "${pkgs}" "parallel" : -short -timeout="${TIMEOUT:-3m}" "${COMMON_TEST_FLAGS[@]}" "${RUN_ARG[@]}" "$@"
+  # shellcheck disable=SC2068 #For context see - https://github.com/etcd-io/etcd/pull/16433#issuecomment-1684312755
+  GOLANG_TEST_SHORT=true go_test "${pkgs}" "parallel" : -short -timeout="${TIMEOUT:-3m}" ${COMMON_TEST_FLAGS[@]:-} ${RUN_ARG[@]:-} "$@"
 }
 
 function unit_pass {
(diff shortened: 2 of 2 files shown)


shyamjvs:
cc @ahrtr - if you could kindly help review

jmhbnz (approved):
LGTM - Thanks for this first contribution @shyamjvs 🎉 

Minor suggestion would be to have one small helper function for the `tail` path lookup rather than repeat same code three times.

shyamjvs:
Changed it, thanks.

Also verified the tests pass on both linux/osx and bash/zsh:
```
$ make test
...
ok  	go.etcd.io/etcd/v3/contrib/raftexample	12.731s
ok  	go.etcd.io/etcd/v3/tools/etcd-dump-logs	1.453s
'unit' PASSED and completed at Thu Aug 17 14:57:26 PDT 2023
SUCCESS
```

logicalhan (commented):
/lgtm
/approve

pkg/expect/expect_test.go
```diff
@@ -46,6 +57,8 @@ func TestExpectFunc(t *testing.T) {
 }
 
 func TestExpectFuncTimeout(t *testing.T) {
+	tailpath := getTailPath(t)
+
 	ep, err := NewExpect("tail", "-f", "/dev/null")
```
fuweid, line 62: It's not block comment: why not use `tailpath` as first argument of `NewExpect`?

fuweid: or, just use `tail` instead of full path

ahrtr:
Please resolve the workflow failures, thx

pkg/expect/expect_test.go
```diff
@@ -72,11 +85,13 @@ func TestExpectFuncTimeout(t *testing.T) {
 	}
 
 	err = ep.Close()
-	require.ErrorContains(t, err, "unexpected exit code [143] after running [/usr/bin/tail -f /dev/null]")
+	require.ErrorContains(t, err, fmt.Sprintf("unexpected exit code [143] after running [%s -f /dev/null]", tailpath))
```
serathius, line 88: Could we just remove absolute paths?
```suggestion
	require.ErrorContains(t, err, "unexpected exit code [143] after running")
```

pkg/expect/expect_test.go
```diff
@@ -89,11 +104,13 @@ func TestExpectFuncExitFailure(t *testing.T) {
 	_, err = ep.ExpectFunc(ctx, func(s string) bool {
 		return strings.Contains(s, "something entirely unexpected")
 	})
-	require.ErrorContains(t, err, "unexpected exit code [1] after running [/usr/bin/tail -x]")
+	require.ErrorContains(t, err, fmt.Sprintf("unexpected exit code [1] after running [%s -x]", tailpath))
```
serathius, line 107:
```suggestion
	require.ErrorContains(t, err, "unexpected exit code [1] after running")
```

pkg/expect/expect_test.go
```diff
@@ -115,7 +132,7 @@ func TestExpectFuncExitFailureStop(t *testing.T) {
 		t.Fatal(err)
 	}
 	err = ep.Close()
-	require.ErrorContains(t, err, "unexpected exit code [1] after running [/usr/bin/tail -x]")
+	require.ErrorContains(t, err, fmt.Sprintf("unexpected exit code [1] after running [%s -x]", tailpath))
```
serathius, line 135:
```suggestion
	require.ErrorContains(t, err, "unexpected exit code [1]")
```

pkg/expect/expect_test.go
shyamjvs: It does make it simpler not printing verbose error. Changed it here and elsewhere, thanks.

shyamjvs:
@ahrtr Thanks for kicking off the CI tests. The jobs seem to fail because it's using a shell version that treats empty value for RUN_ARG (`''`) as an argument here:

```
% (cd api && 'env' 'ETCD_VERIFY=all' 'go' 'test' '-v' '-json' '-p=4' '-short' '-timeout=3m' '--race=false' '--cpu=1' '' './...')
stderr: no Go files in /home/runner/work/etcd/etcd/api
```

On the other hand changing `${RUN_ARG[@]:-}` back to `${RUN_ARG[@]}` would fail on older bash versions that treat an empty array as an unset variable and fail (due to `set -o unset`) - https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u.

To work with both, I've removed double quotes around the literal and tried both `make test-unit` and `CPU=4 RACE=true make test-unit` - they passed on both older/newer bash versions. Can you kick off CI again? Hopefully works now.

shyamjvs:
There is shellcheck directive that can disable specific checks where it makes sense ([xref](https://github.com/etcd-io/etcd/pull/16433#issuecomment-1684235851)). I'll add an exception for code `SC2068` before the lines I modified to make the script work for all.

pkg/expect/expect_test.go
```diff
@@ -72,7 +72,7 @@ func TestExpectFuncTimeout(t *testing.T) {
 	}
 
 	err = ep.Close()
-	require.ErrorContains(t, err, "unexpected exit code [143] after running [/usr/bin/tail -f /dev/null]")
+	require.ErrorContains(t, err, "unexpected exit code [143]")
```
ahrtr, line 75: minor comment:
```suggestion
	require.Regexp(t, "unexpected exit code .*143.* after running.*tail", err.Error())
```

shyamjvs: wondering if the current format is easier to understand and reason about? the pattern is used elsewhere too:
```
tests/e2e/ctl_v3_auth_test.go:                  require.ErrorContains(cx.t, err, "unexpected exit code")
tests/e2e/ctl_v3_elect_test.go:         require.ErrorContains(cx.t, err, "unexpected exit code")
tests/e2e/ctl_v3_lock_test.go:          require.ErrorContains(cx.t, err, "unexpected exit code")
tests/e2e/v3_curl_test.go:      require.ErrorContains(cx.t, err, "unexpected exit code")
```

But happy to change if you feel strongly about it.

shyamjvs:
The single test failure `linux-amd64-integration-4-cpu` looks like an unrelated flake. Can we re-kick that?

pkg/expect/expect_test.go
ahrtr: It isn't a big deal, feel free to keep it as it's.

ahrtr (approved):
LGTM

thx

shyamjvs:
Can one of the maintainers merge this change given tests passed?

serathius (approved):

Issue #16275, opened 2023-07-20
TestPageWriterRandom fails on ARM


zaitcev:
### Which github workflows are flaking?

The issue occurs in CI of Fedora Linux, not on Github, and only on architectures other than x86_64, such as aarch64.

### Which tests are flaking?

TestPageWriterRandom

### Reason for failure (if possible)

The TestPageWriterRandom has 2 problems.

Problem 1: it uses rand.Intn(), but forgets rand.Seed().
As a result, it passes or fails depending on how the platform
sets up the module. This is why it passes on x86_64: the platform
happens to seed the rand module in such a way that the sequence
makes the buggy test to work.

If I adds rand.Seed(time.Now().UnixNano()), the test begins to
fail on my laptop in the exact way it fails in Koji on aarch64.

Problem 2: The test is obviously incorrect, even though it
existed with no change since it was committed in 2016,
commit 2943bf908606ccbfaeda3bdf882a11a0138a0502.

The condition it tests obviously may fail. The buffer can
contain several pages (especially if they're this small).
If the very first write is longer several pages,
  if len(p)+pw.bufferedBytes <= pw.bufWatermarkBytes { }
triggers, and Write returns, while retaining everything
written in the buffer. At that point, the as-tested
condition is obviously violated, as cw.writeBytes is zero.
I'm too lazy to construct a long scenario, but it obviously
can happen on the last write too (for example, if the
previous 4045 writes reset the buffer to zero by accident).

The tested condition needs to change to a valid one.

fuweid:
@zaitcev Hi, I don't have arm64 linux env but I can try to reproduce it in my M1 macos. Wish me luck :)

zaitcev:
@fyweud It may be a misunderstanding. Your fix (to the arithmetic) looks good to me and it fixes the problem. However, adding a rand.Seed() that I propose, allows to reproduce the problem on x86_64. With that in place, there is no need to use PPC or ARM.

zaitcev:
I made a follow-up PR.

jmhbnz:
Closing following our migration to go 1.20 where global random is now seeded automatically, this flake should no longer be occurring.

zaitcev:
Sure, if that's the consensus. I'll make sure to withdraw the PR too unless it's already closed.

Issue #16428, opened 2023-08-16
all: stop using math/rand.Seed


gocurr:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [ ] This is not a support request, support requests should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

It's updated to Go 1.20 minor release, however math/rand.Seed is marked "Deprecated":

> // Prior to Go 1.20, the generator was seeded like Seed(1) at program startup.
// To force the old behavior, call Seed(1) at program startup.
// Alternately, set GODEBUG=randautoseed=0 in the environment
// before making any calls to functions in this package.
//
// Deprecated: As of Go 1.20 there is no reason to call Seed with
// a random value. Programs that call Seed with a known value to get
// a specific sequence of results should use New(NewSource(seed)) to
// obtain a local random generator.

### What did you expect to happen?

Stop using math/rand.Seed.

### How can we reproduce it (as minimally and precisely as possible)?

Use the new API as suggested.

### Etcd version (please run commands below)

Versions updated to Go 1.20.

jmhbnz:
Thanks for raising this @gocurr.  More context here for anyone interested: https://github.com/golang/go/issues/54880 

Looks like eight instances to consider updating:

```bash
 ➜ find . -name "*.go" -type f -print | xargs grep .Seed
./pkg/stringutil/rand.go:                                   rand.Seed(time.Now().UnixNano())
./pkg/adt/interval_tree_test.go:                            rand.Seed(time.Now().UnixNano())
./pkg/proxy/server.go:                                      mrand.Seed(int64(now.Nanosecond()))
./pkg/proxy/server_test.go:                                 rand.Seed(now)
./tests/integration/clientv3/snapshot/v3_snapshot_test.go:  rand.Seed(int64(time.Now().Nanosecond()))
./server/lease/lessor_bench_test.go:                        rand.Seed(time.Now().UTC().UnixNano())
./server/etcdserver/server.go:                              rand.Seed(time.Now().UnixNano())                                       
```

gocurr:
[...]

PR #16434, opened 2023-08-17
pkg/ioutil: make TestPageWriterRandom run the same on all platforms

zaitcev:
This fix came about because the test failed on alternative platforms, such as aarch64 and ppc64, but passed on x86_64. The following commit fixed the error in the test:
  fddd1add52b33649a99d7f756404924138344a10

However, the reason why this was not found on x86_64 is the incorrect use of time.random(). It has to be seeded, else it repeats the same fixed sequence. By a coincidence, x86_64 started its generator in such a way that made the test pass.

Fix: #16275

### DIFF
--- a/pkg/ioutil/pagewriter_test.go
+++ b/pkg/ioutil/pagewriter_test.go
@@ -17,11 +17,13 @@ package ioutil
 import (
 	"math/rand"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestPageWriterRandom(t *testing.T) {
+	rand.Seed(time.Now().UnixNano())
 	// smaller buffer for stress testing
 	defaultBufferBytes = 8 * 1024
 	pageBytes := 128


fuweid (approved):
LGTM

Maybe we should use crypto/rand

zaitcev:
I avoid using true randomness out of a reflex, although indeed it's not likely that fetching 4096 integers would exhaust a system's entropy and cause any hangs. So I don't have a strong preference for or against.

jmhbnz (commented):
Please refer to discussion here https://github.com/etcd-io/etcd/issues/16428.

Now that we have bumped to go 1.20 we need to re-evaulate any use of `rand.Seed` as it is now deprecated.

From 1.20 onwards if `Seed` is not called, the generator will be seeded randomly at program startup so the flake you are seeing may now be eliminated @zaitcev.  

Can you confirm if the flake is still present after our update to go 1.20?

ahrtr:
The rand.Seed has already been deprecated in go 1.20

Issue #16399, opened 2023-08-10
--socket-reuse-port and --socket-reuse-address These two flags have bugs


coderchuan:
### What happened?

--socket-reuse-port and --socket-reuse-address invalid in the configuration file, but are valid in the command flag

### What did you expect to happen?

--socket-reuse-port and --socket-reuse-address in the command flags should behave the same as in the configuration file

### Detail
**Following are the command flags，both node1 and node2 listening to the same port are running normally,  even though they use the same port, they work because of the --socket-reuse-port flag**
1.  etcd --name=node1 --force-new-cluster=false --initial-advertise-peer-urls=http://localhost:2380 --listen-peer-urls="http://localhost:2380" --advertise-client-urls="http://localhost:2379" --listen-client-urls="http://localhost:2379" --socket-reuse-port=true
2. etcd --name=node2 --force-new-cluster=false --initial-advertise-peer-urls=http://localhost:2380 --listen-peer-urls="http://localhost:2380" --advertise-client-urls="http://localhost:2379" --listen-client-urls="http://localhost:2379" --socket-reuse-port=true

**The following is the content of the configuration file in yaml format, running the first one is normal, running the second one will report address already in used, even with the socket-reuse-port flag enabled**
* node1, save as node1.yaml   ` etcd --config-file=node1.yaml`
  ```yaml
  name: node1
  force-new-cluster: false
  initial-advertise-peer-urls: http://localhost:2380
  listen-peer-urls: "http://localhost:2380"
  advertise-client-urls: "http://localhost:2379"
  listen-client-urls: "http://localhost:2379"
  socket-reuse-port: true
  ```
* node2,  save as node2.yaml    `etcd --config-file=node2.yaml`
  ```yaml
  name: node2
  force-new-cluster: false
  initial-advertise-peer-urls: http://localhost:2380
  listen-peer-urls: "http://localhost:2380"
  advertise-client-urls: "http://localhost:2379"
  listen-client-urls: "http://localhost:2379"
  socket-reuse-port: true
  ```

### Etcd version
```text
etcd Version: 3.5.9
Git SHA: bdbbde998
Go Version: go1.19.9
Go OS/Arch: linux/amd64
```

jmhbnz:
Thanks for raising this @coderchuan. 

After some quick investigation it looks as though the config file might work as follows:

```yaml
socket-options:
  reuse-port: true
```

[...]

PR #12702, opened 2021-02-19
*: add support for socket options

hexfusion:
This PR adds support for setting socket options `SO_REUSEADDR` and `SO_REUSEPORT` to etcd listeners via `ListenConfig`. These options give the flexibility to cluster admins who wish to more explicit control of these features. What we have found is during etcd process restart there can be a considerable time waiting for the port to release as it is held open by `TIME_WAIT` which on many systems is 60s.

```
--socket-reuse-port: enables SO_REUSEPORT [1]  allows rebind of a port already in use. 
--socket-reuse-address: enables SO_REUSEADDR [1]  allows binding to an address in `TIME_WAIT` state. 
```
```
$ ps aux | grep etcd | wc -l
0

$  ss --numeric -o state time-wait 
tcp    0       0                 10.0.138.12:34642            10.0.138.12:2380   timer:(timewait,50sec,0)                                                       
tcp    0       0                 10.0.138.12:34470            10.0.138.12:2380   timer:(timewait,35sec,0)                                                       
tcp    0       0                 10.0.138.12:34086            10.0.138.12:2380   timer:(timewait,10sec,0)                                                       
tcp    0       0                 10.0.138.12:34232            10.0.138.12:2380   timer:(timewait,20sec,0)                                                       
tcp    0       0                 10.0.138.12:34694            10.0.138.12:2380   timer:(timewait,55sec,0)                                                       
tcp    0       0                 10.0.138.12:34332            10.0.138.12:2380   timer:(timewait,25sec,0)                                                       
tcp    0       0                 10.0.138.12:33966            10.0.138.12:2380   timer:(timewait,845ms,0)                                                       
tcp    0       0                 10.0.138.12:34040            10.0.138.12:2380   timer:(timewait,5.845ms,0)                                                     
tcp    0       0                 10.0.138.12:34184            10.0.138.12:2380   timer:(timewait,15sec,0)                                                       
tcp    0       0                 10.0.138.12:34570            10.0.138.12:2380   timer:(timewait,45sec,0)                                                       
tcp    0       0                 10.0.138.12:34512            10.0.138.12:2380   timer:(timewait,40sec,0)                                                       
tcp    0       0                 10.0.138.12:34398            10.0.138.12:2380   timer:(timewait,30sec,0)  
```
So we can wait for many seconds even after etcd process is long dead for client and peer ports to become available.

```
2021-02-19 13:57:25.117257 C | etcdmain: listen tcp 10.0.138.12:2380: bind: address already in use
```

[...]

PR #16435, opened 2023-08-18
[3.5] Backport expose socket options

jmhbnz:
Currently in `main` it's possible to specify `--socket-reuse-port` and `--socket-reuse-address` in the config file. However, the same cannot be done for `release-3.5`.

An added complication is that the format for specifying these in the config file differs slightly from the command line parameter. Example below:

```yaml 
socket-options:
  reuse-port: true
```

This backport addresses the immediate bug of configuration via file for these options not being possible in `release-3.5`.

In future, we should also consider aligning  the way the option name should be used in both the configuration file and the command flags.

Relates to #16399.

### DIFF
--- a/client/pkg/transport/sockopt.go
+++ b/client/pkg/transport/sockopt.go
@@ -21,12 +21,12 @@ type SocketOpts struct {
 	// in which case lock on data file could result in unexpected
 	// condition. User should take caution to protect against lock race.
 	// [1] https://man7.org/linux/man-pages/man7/socket.7.html
-	ReusePort bool
+	ReusePort bool `json:"reuse-port"`
 	// ReuseAddress enables a socket option SO_REUSEADDR which allows
 	// binding to an address in `TIME_WAIT` state. Useful to improve MTTR
 	// in cases where etcd slow to restart due to excessive `TIME_WAIT`.
 	// [1] https://man7.org/linux/man-pages/man7/socket.7.html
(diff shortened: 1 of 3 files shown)


jmhbnz:
As another follow-up we should update our [sample etcd config file](https://github.com/etcd-io/etcd/blob/main/etcd.conf.yml.sample) to make it clear how these options should be specified.

fuweid (approved):
LGTM

serathius (approved):

ahrtr (approved):

ahrtr:
Could you also add a changelog item for 3.5? @jmhbnz

ahrtr:
Note: the option `SocketOpts` was introduced in 3.5 in https://github.com/etcd-io/etcd/pull/12702, so no need to backport this PR to 3.4.

jmhbnz:
> Note: the option `SocketOpts` was introduced in 3.5 in #12702, so no need to backport this PR to 3.4.

Hmm if the feature is not in `release-3.4` why would we backport a fix for the feature to that branch?  That would mean we need to backport the whole feature?

chaochn47:
I guess "no need" here means "there is no need" :) 

New 3.4 release won't consider backporting this PR.

jmhbnz:
> I guess "no need" here means "there is no need" :)

Whoops completely misread lol, thanks @chaochn47  🤦🏻‍♂️

PR #16438, opened 2023-08-18
Fix raftexample possible idempotent-problem

chenyang8094:
issue ref: https://github.com/etcd-io/etcd/discussions/16425

I know that the current kvstore does not have idempotence problems (even if the wal log is replayed multiple times), because it uses a map structure (key value). But I think raftexample should fully consider this scenario. Just like the etcd server, by maintaining a consistentIndex in the db to avoid the idempotence problem of replaying wal.

### DIFF
--- a/contrib/raftexample/kvstore.go
+++ b/contrib/raftexample/kvstore.go
@@ -39,15 +39,11 @@ type kv struct {
 	Val string
 }
 
-func newKVStore(snapshotter *snap.Snapshotter, proposeC chan<- string, commitC <-chan *commit, errorC <-chan error) *kvstore {
+func newKVStore(snapshotter *snap.Snapshotter, initialSnapshot *raftpb.Snapshot, proposeC chan<- string, commitC <-chan *commit, errorC <-chan error) *kvstore {
 	s := &kvstore{proposeC: proposeC, kvStore: make(map[string]string), snapshotter: snapshotter}
-	snapshot, err := s.loadSnapshot()
-	if err != nil {
-		log.Panic(err)
-	}
-	if snapshot != nil {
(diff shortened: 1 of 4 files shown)


ahrtr:
Thanks for the PR.
- raftexample is just an example, it doesn't have to be perfect.
- We are planning to remove raftexample from etcd repo, and implement a self-contained example in raft repo. Pls see https://github.com/etcd-io/raft/issues/2

So please do not spend too much effort to improve raftexample. Instead, if you have bandwidth and be interested, please help to resolve https://github.com/etcd-io/raft/issues/2

chenyang8094:
> Thanks for the PR.
> 
> * raftexample is just an example, it doesn't have to be perfect.
> * We are planning to remove raftexample from etcd repo, and implement a self-contained example in raft repo. Pls see [Implement a self-contained raft example raft#2](https://github.com/etcd-io/raft/issues/2)
> 
> So please do not spend too much effort to improve raftexample. Instead, if you have bandwidth and be interested, please help to resolve [etcd-io/raft#2](https://github.com/etcd-io/raft/issues/2)

I see. Although the raftexample is just one example, it may actually be more important than we think. In my work, I have seen many projects using raft based on raftexample. There are also countless articles on the Internet analyzing raftexample. The reason is that raftexample is indeed a simple and effective way to learn etcd/raft. Therefore, I think that raftexample should show the precautions when using etcd-raft library as much as possible, and provide more help to everyone who is looking at this project. I just took a little time to work out the little imperfection I got.

As for the issue https://github.com/etcd-io/raft/issues/2 , I will continue to pay attention, thank you.

jmhbnz:
Discussed during sig-etcd triage meeting. Given we are moving raftexample suggest we close this pr. @ahrtr do you agree?

PR #16445, opened 2023-08-19
Run integration, e2e & grpcproxy tests in container.

jmhbnz:
Recent pull request https://github.com/etcd-io/etcd/pull/15927 introduced running our `arm64` ci jobs in containers to prevent leaking test processes causing issues on our self hosted runners.

The github runners we use for non arm64 ci jobs do not have the cleanup issue as our self hosted runners do, so it isn't absolutely necessary to migrate all workflows into containers. However we should consider doing it on the basis that we have one unified way to run etcd testing workflows.

Fixes: https://github.com/etcd-io/etcd/issues/15956

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -5,6 +5,11 @@ permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
+    container:
+      image: golang:1.20-bullseye
+    defaults:
+      run:
+        shell: bash
     strategy:
       fail-fast: true
       matrix:
@@ -13,6 +18,8 @@ jobs:
           - linux-386-e2e
     steps:
       - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      # https://github.com/actions/checkout/issues/1169
+      - run: git config --system --add safe.directory '*'
       - id: goversion
         run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
       - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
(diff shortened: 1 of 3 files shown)


ahrtr (approved):
LGTM

Thanks

.github/workflows/e2e.yaml
```diff
@@ -5,6 +5,11 @@ permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
+    container:
+      image: golang:1.20-bullseye
```
serathius, line 9: Question, do we need to run in Go docker image? It undermines the work we put in to have single place that specifies Go version.  I would assume that block below should install a proper Go version.
```
      - id: goversion
        run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
      - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
        with:
          go-version: ${{ steps.goversion.outputs.goversion }}
```

serathius (approved):

.github/workflows/e2e.yaml
jmhbnz: It's a fair question, we don't need to.  I think originally I chose the golang based image so that if necessary I could "docker exec" into the container to test things and verify steps the ci would run were going to work as expected.

It gives us a way to test things without having to replicate what `actions/setup-go`  would do, and also doesn't impact/hinder the version of go installed by `actions/setup-go` so I think my preference would be to maintain current image rather than step back to pure debian or ubuntu image without golang present.

serathius: I expect that `actions/setup-go`  installs go version from binaries that is independent to one from `golang:1.20-bullseye`. Might be wrong though.

PR #16376, opened 2023-08-04
Update etcdutl restore to create v2 snapshot from v3

geetasg:
Related to https://github.com/etcd-io/etcd/issues/12913
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #16457, opened 2023-08-22
Add membership changes to downgrade tests

serathius:
cc @ahrtr @geetasg

### DIFF
--- a/tests/e2e/cluster_downgrade_test.go
+++ b/tests/e2e/cluster_downgrade_test.go
@@ -92,12 +92,25 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool) {
 			Storage: currentVersionStr,
 		})
 	}
+	cc := epc.Etcdctl()
 	t.Logf("Cluster created")
+	if len(epc.Procs) > 1 {
+		t.Log("Waiting health interval to required to make membership changes")
+		time.Sleep(etcdserver.HealthInterval)
+	}
+
+	t.Log("Adding member to test membership, but a learner avoid breaking quorum")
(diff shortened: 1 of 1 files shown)


tests/e2e/cluster_downgrade_test.go
```diff
@@ -92,12 +92,25 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool) {
...
+	}
+
+	t.Log("Adding member to test membership, but a learner avoid breaking quorum")
+	resp, err := cc.MemberAddAsLearner(context.Background(), "fake1", []string{"http://127.0.0.1:1001"})
+	require.NoError(t, err)
 	if triggerSnapshot {
 		t.Logf("Generating snapshot")
-		generateSnapshot(t, snapshotCount, epc)
+		generateSnapshot(t, snapshotCount, cc)
+		verifySnapshot(t, epc)
```
ahrtr, line 108: Suggest to verify the snapshot using `verifySnapshotMembers`. 

FYI. https://github.com/etcd-io/etcd/pull/16376

and also https://github.com/etcd-io/etcd/pull/16458

serathius: Don't think this is needed. Here this is a downgrade/upgrade test, so from downgrade perspective we just need to confirm that members don't change during downgrade/upgrade procedure. Checking exact list of members is not in scope of the test. Aka, I would treat members as black box instead of whitebox.

ahrtr (approved):

geetasg:
lgtm. investigating additional test cases for downgrade

geetasg:
Considering adding following scenarios - 
* Add a duplicate member after downgrade - should fail.
* Add a already removed member after downgrade - should fail.
* Same verifications after upgrade.
* Trying to find a way to trigger membership changes while snapshot is in progress

PR #16447, opened 2023-08-20
all: stop using math/rand.Seed

gocurr:
Fixes #16428.

PR #16459, opened 2023-08-22
Use crypto/rand.Read instead of deprecated math/rand.Read

jmhbnz:
We recently updated to golang 1.20.  This golang release deprecated `math/rand.Seed` (which in most cases can now be removed) and `math/rand.Read` in favor of `crypto/rand.Read`, refer: https://go.dev/doc/go1.20

In #16447 we removed instances of `math/rand.Seed`.

After scanning through the codebase this pull request updates the one instance I could find of us using `math/rand.Read` and replaces it with `crypto/rand.Read`.

### DIFF
--- a/server/storage/wal/wal_test.go
+++ b/server/storage/wal/wal_test.go
@@ -16,11 +16,11 @@ package wal
 
 import (
 	"bytes"
+	"crypto/rand"
 	"errors"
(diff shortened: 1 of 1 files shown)


ahrtr (approved):
LGTM

thx @jmhbnz

Issue #16034, opened 2023-06-08
Ensure etcd cli help flags cannot be missed


jmhbnz:
### What would you like to be added?

When we update etcd to introduce new command line parameters we currently don't have any tests or other automation to verify that our cli help stays in sync.

This situation played out recently where https://github.com/etcd-io/etcd/pull/14120 introduced new command line flags and reviewers did not catch the missing help content. This leads to followup pr's being required, example: https://github.com/etcd-io/etcd/pull/16031.

A potential approach we could try is introducing a test that will verify help content matches an expected template: https://github.com/kubernetes-sigs/metrics-server/pull/939.  Alternatively we could introduce some automation for the generation of the help content.

### Why is this needed?

- Ensure our help content is accurate and remains relevant. 
- Avoid user confusion and unnecessary support requests.
- Reduce burden on pull request reviewers.

charles-chenzz:
if I didn't understand it wrong, what we do here is add a file ``command-line-flags.txt`` and add command-line in ``etcd`` to the file, and refer the test-image.sh part of https://github.com/kubernetes-sigs/metrics-server/pull/939 

@jmhbnz

jmhbnz:
> if I didn't understand it wrong, what we do here is add a file `command-line-flags.txt` and add command-line in `etcd` to the file, and refer the test-image.sh part of [kubernetes-sigs/metrics-server#939](https://github.com/kubernetes-sigs/metrics-server/pull/939)

Hey @charles-chenzz - Yes I think we can re-use a similar approach. Rather than the bash script test though I'm thinking we could make it part of one the formal go test suites.

@serathius - If we go with implementing a test do you have any preference on if this is is a separate bash script approach, versus inclusion in one of the go test suites?

charles-chenzz:
if we re-use similar approach I think I can take it up

nitishchauhan0022:
The alternate approach of automated generation of help content sounds less error prone and will help the developers as well.

nitishchauhan0022:
hey @jmhbnz I am trying to understand the issue. One thing, i am not getting is if a developer forgets to add the help content of a new flag in server/etcdmain/help.go than how can verfiying it with a approach as mentioned in https://github.com/kubernetes-sigs/metrics-server/pull/939 can detect this since the help content here is not generating itself but it is using server/etcdmain/help.go file. Please correct me if I'm mistaken, but these are my preliminary understandings at the moment.

jmhbnz:
[...]

PR #16465, opened 2023-08-23
Unit test for CLI flags and add missing flags

riendeau:
An attempt to address https://github.com/etcd-io/etcd/issues/16034

Adds a unit test to validate that every (non-ignored) flag defined is present in the help text (either in the usage line or the flags line).

Also adds information on several flags that the unit test identified as missing to the usage text.


serathius (approved):

ahrtr (approved):
Nice test case!

Thanks @riendeau

PR #16472, opened 2023-08-25
test: support regular expression matching on the response

ahrtr:
Currently we only support exact string match on the response (e.g. curl response, and process output). But in some cases, the response may not be a fixed string. One example is discussed in https://github.com/etcd-io/etcd/pull/16454, there may be random spaces in the grpc-gateway v2 response.

So we need to support match the response based on a regular expression.

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 	"sync"
 	"syscall"
@@ -36,6 +37,11 @@ const DEBUG_LINES_TAIL = 40
 var (
 	ErrProcessRunning = fmt.Errorf("process is still running")
 )
+
+type ExpectedResponse struct {
+	Value         string
+	IsRegularExpr bool
+}
 
 type ExpectProcess struct {
 	cfg expectConfig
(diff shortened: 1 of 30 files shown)


ahrtr:
@fuweid @jmhbnz @serathius @spzala @wenjiaswe PTAL

jmhbnz (approved):
LGTM - Thanks @ahrtr.

Noting `expect.ExpectResponse` feels a bit verbose/duplicated naming  but I can't think of a better way of putting it in relation to the rest of the package.

pkg/expect/expect.go
```diff
@@ -223,14 +229,23 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 }
 
 // ExpectWithContext returns the first line containing the given string.
-func (ep *ExpectProcess) ExpectWithContext(ctx context.Context, s string) (string, error) {
-	return ep.ExpectFunc(ctx, func(txt string) bool { return strings.Contains(txt, s) })
+func (ep *ExpectProcess) ExpectWithContext(ctx context.Context, s ExpectedResponse) (string, error) {
+	return ep.ExpectFunc(ctx, func(txt string) bool {
+		if s.IsRegularExpr {
```
fuweid, line 234: I think we can Compile out of closure and return error as soon as possible. And then we don't need to Compile it all the time.

tests/e2e/ctl_v3_kv_test.go
```diff
@@ -354,25 +355,15 @@ func ctlV3Get(cx ctlCtx, args []string, kvs ...kv) error {
...
-	var lines []string
+	var lines []expect.ExpectedResponse
 	for _, elem := range kvs {
-		lines = append(lines, elem.key, elem.val)
+		lines = append(lines, expect.ExpectedResponse{Value: elem.key}, expect.ExpectedResponse{Value: elem.val})
 	}
 	return e2e.SpawnWithExpects(cmdArgs, cx.envMap, lines...)
 }
 
-// ctlV3GetWithErr runs "get" command expecting no output but error
```
fuweid, line 364: Good catch!

pkg/expect/expect.go
ahrtr: Makes sense & updated & thx

fuweid (approved):
LGTM

pkg/expect/expect.go
```diff
@@ -223,14 +229,29 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
...
+	return ep.ExpectFunc(ctx, func(txt string) bool {
+		if expr != nil {
+			return expr.MatchString(txt)
+		}
+		return strings.Contains(txt, s.Value)
+	})
 }
 
 // Expect returns the first line containing the given string.
 // Deprecated: please use ExpectWithContext instead.
```
wenjiaswe, line 252: This function is deprecated? Do we need to move it? Just want to mark, no need to remove in this PR,

pkg/expect/expect.go
```diff
@@ -37,6 +38,11 @@ var (
 	ErrProcessRunning = fmt.Errorf("process is still running")
 )
 
+type ExpectedResponse struct {
+	Value         string
+	IsRegularExpr bool
```
wenjiaswe, line 43: Do you want to add some regex tests?

pkg/expect/expect.go
ahrtr: We can cleanup the code in separate PR.

wenjiaswe: sgtm.

pkg/expect/expect.go
ahrtr: added, please take a look at the second commit. thx

wenjiaswe: Thanks.

wenjiaswe:
lgtm

ahrtr:
thx @wenjiaswe for the quick review. Also thx @jmhbnz and @fuweid

PR #16487, opened 2023-08-26
upgrade debian12 and bookworm

SagarRawat24:
signed-off-by: Sagar Rawat [sagarrawat24291@gmail.com](mailto:sagarrawat24291@gmail.com)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ARG ARCH=amd64
-FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian11
+FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian12
 
 ADD etcd /usr/local/bin/
 ADD etcdctl /usr/local/bin/
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
 	"name": "Go",
 	
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG ARCH=amd64
-FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian11
+FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian12
```
jmhbnz, line 2: As per the warning in the github issue, this should not be merged yet as https://console.cloud.google.com/gcr/images/distroless/GLOBAL/static-debian12 does not yet exist.

Refer: https://github.com/GoogleContainerTools/distroless/issues/1337

jmhbnz (commented):
Thanks for drafting this @SagarRawat24. We will need to hold off merging this until the Distroless image is ready.

Dockerfile
SagarRawat24: @jmhbnz  is it image does not exist globally or error in code

jmhbnz: Hey @SagarRawat24 - The image now exists globally, so these tests should now pass if you re trigger them as follows:

```bash
git reset --soft HEAD~1
git commit -s -a -m "Update to Debian 12 bookworm."
git push --force
```

jmhbnz (changes requested):
Hey @SagarRawat24 - It looks like there is one more file that needs to be updated, can you please check and update `.github/workflows/e2e-arm64-template.yaml`.

PR #16489, opened 2023-08-27
test: add v3 curl test to cover all maintenance REST API

ahrtr:
Linked to https://github.com/etcd-io/etcd/pull/16454

### DIFF
--- a/tests/e2e/v3_curl_test.go
+++ b/tests/e2e/v3_curl_test.go
@@ -370,21 +370,6 @@ func testV3CurlResignMissiongLeaderKey(cx ctlCtx) {
 	}
 }
 
-func TestV3CurlMaintenanceAlarmMissiongAlarm(t *testing.T) {
-	for _, p := range apiPrefix {
-		testCtl(t, testV3CurlMaintenanceAlarmMissiongAlarm, withApiPrefix(p), withCfg(*e2e.NewConfigNoTLS()))
-	}
-}
-
-func testV3CurlMaintenanceAlarmMissiongAlarm(cx ctlCtx) {
-	if err := e2e.CURLPost(cx.epc, e2e.CURLReq{
-		Endpoint: path.Join(cx.apiPrefix, "/maintenance/alarm"),
-		Value:    `{"action": "ACTIVATE"}`,
-	}); err != nil {
-		cx.t.Fatalf("failed post maintenance alarm (%s) (%v)", cx.apiPrefix, err)
-	}
-}
-
 // to manually decode; JSON marshals integer fields with
 // string types, so can't unmarshal with epb.CampaignResponse
 type campaignResponse struct {
--- a/tests/e2e/utils.go
+++ b/tests/e2e/utils.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
@@ -107,4 +108,19 @@ func curl(endpoint string, method string, curlReq e2e.CURLReq, connType e2e.Clie
 		return "", err
 	}
 	return strings.Join(lines, "\n"), nil
+}
+
+func runCommandAndReadJsonOutput(args []string) (map[string]interface{}, error) {
+	lines, err := e2e.RunUtilCompletion(args, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var resp map[string]interface{}
+	err = json.Unmarshal([]byte(strings.Join(lines, "\n")), &resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
 }
--- a/tests/e2e/v3_curl_maintenance_test.go
+++ b/tests/e2e/v3_curl_maintenance_test.go
@@ -1,0 +1,172 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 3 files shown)


tests/e2e/v3_curl_test.go
```diff
@@ -385,6 +386,107 @@ func testV3CurlMaintenanceAlarmMissiongAlarm(cx ctlCtx) {
...
+func TestV3CurlMaintenanceDefragment(t *testing.T) {
+	testCtl(t, testV3CurlMaintenanceDefragment, withCfg(*e2e.NewConfigNoTLS()))
+}
+
+func testV3CurlMaintenanceDefragment(cx ctlCtx) {
+	if err := e2e.CURLPost(cx.epc, e2e.CURLReq{
+		Endpoint: "/v3/maintenance/defragment",
+		Value:    "{}",
+		Expected: expect.ExpectedResponse{
+			Value: "{}",
```
jmhbnz, line 415: Just double checking this is intentional and there is no return we can validate here?

jmhbnz (approved):
LGTM - Thanks @ahrtr 👍🏻

tests/e2e/v3_curl_test.go
ahrtr: It's intentional. The response is just "{}".
```
$ curl -L http://127.0.0.1:2379/v3/maintenance/defragment -X POST
{}
```

fuweid (approved):
LGTM

tests/e2e/v3_curl_test.go
```diff
@@ -385,6 +386,107 @@ func testV3CurlMaintenanceAlarmMissiongAlarm(cx ctlCtx) {
...
+func TestV3CurlMaintenanceStatus(t *testing.T) {
+	testCtl(t, testV3CurlMaintenanceStatus, withCfg(*e2e.NewConfigNoTLS()))
+}
+
+func testV3CurlMaintenanceStatus(cx ctlCtx) {
+	if err := e2e.CURLPost(cx.epc, e2e.CURLReq{
+		Endpoint: "/v3/maintenance/status",
+		Value:    "{}",
+		Expected: expect.ExpectedResponse{
+			Value:         fmt.Sprintf(`.*"version":%q,"dbSize":.*,\s*"leader":.*,\s*"raftIndex":.*,\s*"raftTerm":.*,\s*"raftAppliedIndex":.*,\s*"dbSizeInUse":.*,\s*"storageVersion":.*`, version.Version),
```
serathius, line 398: I don't understand why you are testing json using regex? This creates a invalid dependency on order of keys.

Could you just marshall it to map and check keys?

ahrtr: > I don't understand why you are testing json using regex? 

Because we don't care the values, which are also uncertain.

> This creates a invalid dependency on order of keys.

The objects are marshalled in the order that fields are declared, why it is invalid? Note I don't think we will change the order of the fields frequently.

ahrtr: also read https://github.com/etcd-io/etcd/pull/16472

serathius: > The objects are marshalled in the order that fields are declared, why it is invalid? Note I don't think we will change the order of the fields frequently.

Order of the API fields is not something we guarantee and something we should test. I expect it just unexpected side effect in the current implementation of grpc gateway. We should not write tests based on unexpected side effects.

See https://github.com/etcd-io/etcd/blob/d09b8d474a01ab4ebf4a7bd877962c1036fb2b68/tests/e2e/cmux_test.go#L140-L171

serathius: I don't agree with https://github.com/etcd-io/etcd/pull/16472 it's unnesesery overkill that will lead to bad code patterns (aka, use regex for everything) and bloat (regex patterns gets so complicated we stop understanding them).

ahrtr: I agree the solution isn't perfect. But I'd like keep it as it's.

Only the order of map fields are clearly mentioned not guaranteed. Struct fields are stable for now, although I don't see clear document on that. 

Unmarshalling the response into a struct without error doesn't mean it's a valid or expected response. Json by default ignore the unknown fields. Even the server side gives your a "{}" response, You can also unmarshal it into most of the existing PB structs in etcd. You have to double check whether the values are expected. Comparing the string directly (exact matching or regular expression matching) doesn't have such issue.

Enhancing the test framework to support regular expression matching doesn't mean we will use it for everything. Also the existing regular expressions what we used is simple enough.

ahrtr: The duplicated message definitions seem also not good. https://github.com/etcd-io/etcd/blob/d09b8d474a01ab4ebf4a7bd877962c1036fb2b68/tests/e2e/cmux_test.go#L157-L168

serathius: > Json by default ignore the unknown fields.

We don't need to do that. 

> Even the server side gives your a "{}" response, You can also unmarshal it into most of the existing PB structs in etcd. You have to double check whether the values are expected. Comparing the string directly (exact matching or regular expression matching) doesn't have such issue.
 
 You can unmarshal into `map[string]interface{}` and check keys.

serathius: > The duplicated message definitions seem also not good.

yea, it's an overkill here.

ahrtr: > You can unmarshal into map[string]interface{} and check keys.

Actually I thought about this approach as well. The reason why I didn't do it is it complicates the test cases, especially when there are nested structs.

ahrtr: Updated anyway, PTAL.

ahrtr:
@serathius PTAL

serathius (approved):

PR #16491, opened 2023-08-28
 Use latest golang image to avoid confusing it's version with go-version

serathius:
Go version from docker image is not used as
https://github.com/actions/setup-go download its own Go binary.

cc @ahrtr @jmhbnz

### DIFF
--- a/.github/workflows/e2e-arm64-template.yaml
+++ b/.github/workflows/e2e-arm64-template.yaml
@@ -12,7 +12,8 @@ jobs:
     # this is to prevent the job to run at forked projects
     if: github.repository == 'etcd-io/etcd'
     runs-on: [self-hosted, Linux, ARM64]
-    container: golang:1.20-bullseye
+    container:
+      image: golang:bullseye@sha256:02f350d8452d3f9693a450586659ecdc6e40e9be8f8dfc6d402300d87223fdfa
     defaults:
       run:
         shell: bash
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -6,7 +6,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     container:
-      image: golang:1.20-bullseye
+      image: golang:bullseye@sha256:02f350d8452d3f9693a450586659ecdc6e40e9be8f8dfc6d402300d87223fdfa
     defaults:
       run:
         shell: bash
--- a/.github/workflows/grpcproxy.yaml
+++ b/.github/workflows/grpcproxy.yaml
@@ -5,7 +5,8 @@ permissions: read-all
 jobs:
   test:
     runs-on: ubuntu-latest
-    container: golang:1.20-bullseye
+    container:
+      image: golang:bullseye@sha256:02f350d8452d3f9693a450586659ecdc6e40e9be8f8dfc6d402300d87223fdfa
     defaults:
       run:
         shell: bash
--- a/.github/workflows/robustness-template.yaml
+++ b/.github/workflows/robustness-template.yaml
@@ -27,7 +27,7 @@ jobs:
     timeout-minutes: 210
     runs-on: ${{ fromJson(inputs.runs-on) }}
     container:
-      image: golang:1.20-bullseye
+      image: golang:bullseye@sha256:02f350d8452d3f9693a450586659ecdc6e40e9be8f8dfc6d402300d87223fdfa
       # Required for mounting fuse lazyfs
       options: --privileged
     defaults:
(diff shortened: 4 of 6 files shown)


jmhbnz (commented):
Hmm for performance and ease of debugging I would prefer not to go ahead with this change.

**Performance** - Without using the golang language image each ci job has to update `apt` cache and install some basics like `git` etc.  Apt cache update can sometimes take a while and it's a bit of an antipattern to use package managers during CI when we can bake things into runner image instead.

**Ease of debugging** - With the language image all our normal tools are ready to go, so outside github actions environment we can test things easily without having to manually install stuff.

We've proven that the two versions don't conflict, and in CI `setup-go` installation takes precedence.

If others feel strongly to proceed then happy to be out voted.

jmhbnz (approved):
LGTM - Happy with updated approach of non version pinned.  Part of me liked having it pinned to same version we use for etcd releases but I understand the drive to avoid any confusion with where etcd specifies golang version and it's easy enough to specify the version tag if doing any deep debugging locally.

serathius:
If we don't use Go from the image, the change has same level of version "pinning" by picking debian version.

jmhbnz:
> If we don't use Go from the image, the change has same "level of version pinning" by picking debian version.

We don't use go from image when we are in ci environment because the version installed by `setup-go` takes priority.  But if I wanted to debug something locally and manually start a container of `docker run golang:bullseye` it ships with `go version go1.21.0 linux/amd64` which is currently not the version of go used by the project.

It's not a big deal because I can just instead start the correct pinned version for debugging if I need to look into something on the `arm64` runners etc.

ahrtr:
Although we use the go version defined in setup-go, pining the image version also provides stable environment. Suggest to ping both go version and image version (in a centralized place as well).

serathius:
> pining the image version also provides stable environment

Let's clarify, neither `golang:1.20-bullseye` nor `golang:bullseye` pins the image. If we want to pin the image we would need to provide sha

.github/workflows/e2e-arm64-template.yaml
```diff
@@ -12,7 +12,8 @@ jobs:
     # this is to prevent the job to run at forked projects
     if: github.repository == 'etcd-io/etcd'
     runs-on: [self-hosted, Linux, ARM64]
-    container: golang:1.20-bullseye
+    container:
+      image: golang:bullseye@sha256:02f350d8452d3f9693a450586659ecdc6e40e9be8f8dfc6d402300d87223fdfa
```
ahrtr, line 16:
```suggestion
      image: golang:1.20.7-bullseye
```

ahrtr: also suggest to 
- get it defined in a centralised file, but can be in a separate PR.
- Keep consistent with the go version defined in `.go-version`.

serathius: ?? Do you understand that `golang:1.20.7-bullseye` is not a static image? It's a channel of latest debian bullseye release with golang 1.20.7 installed on top of it. We don't use Go version from it, only the tools that are taken from bullseye debian. Those tools can get patches so not specifying sha is not doing anything for us.

fuweid: For the image name, the name can be `repo:tag@digest`. When digest is in the name, the tag will be ignore during pulling. But the tag is alias for digest and it's easy to read. For the pin image, I think digest is good choice.

jmhbnz: > From our point of view (just using image tools) golang:bullseye and golang:1.20.7-bullseye are the same.

Except for the point I made above about debugging locally where it is convenient for go versions to align to the version the project uses.

I tend to agree with @ahrtr that we should aim to be consistent with our `.go-version`.  We can possibly even re-use that variable for this image definition.

serathius: This discussion is derailing, as mentioned above, the only reason we are using `golang` image is fact that it has tools like `git` pre installed. I strongly oppose a dependency on `.go-version`.

jmhbnz: Not intending to derail.  Just trying to explain that there there are situations I was in when completing our `arm64` work where having an image that also contained a golang install (and the same version etcd used) was invaluable. 

When CI was failing on `arm64` I could start a container manually on the `arm64` runner and step through commands CI was running to do additional debug on what was going wrong, without having to manually install a the required go version in the container first.

If we merge this, we can still do that debugging when needed we just need to know to `docker run ...` the right version of the image, instead of the image that will now be present in the github actions workflows.

As mentioned above, I have approved the changes and this isn't a blocker from my perspective. However if you are asking for my opinion, I would prefer to have consistent versions across our codebase and it would be one less thing to think about when doing debugging on `arm64` runners if we keep the versions aligned.

PR #16500, opened 2023-08-28
Enable reloading CA without a restart

hongbin:
* Add two options to server: "client-root-ca-reload" and "peer-root-ca-reload". By default, these options are false. Whenever the options are enabled, the server will dynamically load CA keys & certs.
* Provide implementation for "GetConfigForClient". This will allow server to load CA files on each TLS handshake.
* Provide implementation for "VerifyConnection". This will allow clients (for peer connection) to load CA files per request.

Note: this patch implements CA reloading without performance optimization. Optimization could be done in the future. Potential optimization is to avoid loading CA on each request. We could implement a background routine to periodically loading CA files instead.

Fixes https://github.com/etcd-io/etcd/issues/11555


vladorf:
hello, we are also in need of this feature

serathius:
Contributions are welcomed!

oblazek:
:wave: hey Marek, you mean we can contribute with the review? :)

serathius:
Please start from contributing to discussion https://github.com/etcd-io/etcd/issues/11555. 

Please note "I need it", is not a productive comment.

Issue #16462, opened 2023-08-23
Upgrade to debian 12 bookworm


jmhbnz:
### What would you like to be added?

Debian 12 "bookworm" was released in June 2023.  We should consider updating etcd project from debian 11  "bullseye".

The Kubernetes project has already made a start on this, refer: https://github.com/kubernetes/release/issues/3128.

For us this would involve:

- [x] Updating the `container:` used in `.github/workflows/` files.

```diff
-    container: golang:1.20-bullseye
+    container: golang:1.20-bookworm
```

- [x] Updating the `.devcontainer/devcontainer.json` used for etcd container based development environments.

```diff
-       "image": "mcr.microsoft.com/devcontainers/go:1.20-bullseye",
+       "image": "mcr.microsoft.com/devcontainers/go:1.20-bookworm",
```

- [x] Update the `FROM:` base image in our `Dockerfile`.   

```diff
-FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian11
+FROM --platform=linux/${ARCH} gcr.io/distroless/static-debian12
```

### Why is this needed?

Ensure etcd releases, testing workflows and container based development use a modern base image. 

Avoid any potential future technical debt and uncover any potential issues early.

jmhbnz:
The initial work for this has now been completed in https://github.com/etcd-io/etcd/pull/16516.

@ahrtr, @serathius what are your thoughts on backporting this? My thinking is we would only do so if for some reason `bullseye` was broken or out of support.  Keen to double check before I close this though.

jmhbnz:
Closing as the initial pr is complete and my instinct is we shouldn't backport this change.  

Happy to re-open if my assessment isn't right 🙏🏻

PR #16513, opened 2023-08-31
upgrade debian12 & bookworm (pull request 16487)

SagarRawat24:
signed-off-by: Sagar Rawat sagarrawat24291@gmail.com

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
 	"name": "Go",
 	
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-	"image": "mcr.microsoft.com/devcontainers/go:1.20-bullseye",
+	"image": "mcr.microsoft.com/devcontainers/go:1.20-bookworm",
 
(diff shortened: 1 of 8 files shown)


jmhbnz (approved):
LGTM Once tests green.  Thanks @SagarRawat24.

@ahrtr can you please kick these off?

Edit: Note fixes https://github.com/etcd-io/etcd/issues/16462

ahrtr:
This PR is breaking `FAIL: TestConnectionMultiplexing/ServerTLS/ClientTLS/curl/http1.0`. 

It looks like http 1.0 isn't supported any more in go:1.20-bookworm. Since HTTP 1.0 has already been obsoleted, it's OK to update the test case.

SagarRawat24:
@jmhbnz  should we update the test case from HTTP 1.0 to 1.1 or other version either we don't need to update

jmhbnz:
> @jmhbnz should we update the test case from HTTP 1.0 to 1.1 or other version either we don't need to update

Please remove the "1.0" entry in this line:

https://github.com/etcd-io/etcd/blob/e4e05c5713300d131e0c55af3fdb60d315c3b17f/tests/e2e/cmux_test.go#L124

PR #16515, opened 2023-09-01
server: Test txn checking

serathius:
cc @ahrtr @jmhbnz 

Tests for https://github.com/etcd-io/etcd/pull/16372

### DIFF
--- a/server/etcdserver/txn/txn_test.go
+++ b/server/etcdserver/txn/txn_test.go
@@ -22,6 +22,8 @@ import (
 
 	"go.uber.org/zap/zaptest"
 
+	"go.etcd.io/etcd/pkg/v3/traceutil"
+
 	"go.etcd.io/etcd/api/v3/authpb"
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/server/v3/auth"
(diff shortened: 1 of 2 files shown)


server/lease/lessor.go
```diff
@@ -816,13 +816,18 @@ func (le *lessor) initAndRecover() {
...
 
 func (fl *FakeLessor) SetCheckpointer(cp Checkpointer) {}
 
-func (fl *FakeLessor) Grant(id LeaseID, ttl int64) (*Lease, error) { return nil, nil }
+func (fl *FakeLessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
+	fl.LeaseSet[id] = struct{}{}
+	return nil, nil
+}
 
 func (fl *FakeLessor) Revoke(id LeaseID) error { return nil }
```
ahrtr, line 832: remove the lease from `LeaseSet`?

server/etcdserver/txn/txn_test.go
```diff
@@ -35,6 +37,241 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestCheckTxn(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	cancel()
```
ahrtr, line 42: Move this into the sub test body and also `defer cancel()`?

server/lease/lessor.go
serathius: The existing fake is not really fake, because it doesn't do anything. However, I don't plan implement full Interface if we don't need it.

server/etcdserver/txn/txn_test.go
serathius: Done

ahrtr (approved):

PR #16517, opened 2023-09-01
Use test cases from TestCheckTxn to test Put and Range

serathius:
cc @ahrtr @jmhbnz

### DIFF
--- a/server/etcdserver/txn/txn_test.go
+++ b/server/etcdserver/txn/txn_test.go
@@ -37,231 +37,198 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type testCase struct {
+	name        string
+	setup       testSetup
+	op          *pb.RequestOp
+	expectError string
+}
+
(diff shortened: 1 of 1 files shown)


server/etcdserver/txn/txn_test.go
```diff
@@ -37,231 +37,236 @@ import (
...
-							},
-						},
-					},
-				},
+			expectError: tc.expectError,
+		})
+	}
+	testCases = append(testCases, txnTestCase{
+		name:  "Invalid range on failed path should succeed",
+		setup: invalidRangeTestCase.setup,
```
ahrtr, line 194: pls make this case independent. In other words, pls do not cross link to/depend on other case. Also `invalidRangeTestCase.setup` isn't assigned a value at all.

serathius: DOne

ahrtr (approved):

PR #16519, opened 2023-09-01
Reuse check put

serathius:
cc @ahrtr @jmhbnz 
Built on https://github.com/etcd-io/etcd/pull/16517

### DIFF
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -43,55 +43,50 @@ func Put(ctx context.Context, lg *zap.Logger, lessor lease.Lessor, kv mvcc.KV, p
 		)
 		ctx = context.WithValue(ctx, traceutil.TraceKey{}, trace)
 	}
-	leaseID := lease.LeaseID(p.Lease)
-	if leaseID != lease.NoLease {
-		if l := lessor.Lookup(leaseID); l == nil {
-			return nil, nil, lease.ErrLeaseNotFound
-		}
-	}
 	txnWrite := kv.Write(trace)
 	defer txnWrite.End()
-	resp, err = put(ctx, txnWrite, p)
+	resp, err = put(ctx, txnWrite, lessor, p)
 	return resp, trace, err
 }
 
-func put(ctx context.Context, txnWrite mvcc.TxnWrite, p *pb.PutRequest) (resp *pb.PutResponse, err error) {
+func put(ctx context.Context, txnWrite mvcc.TxnWrite, lessor lease.Lessor, req *pb.PutRequest) (resp *pb.PutResponse, err error) {
 	trace := traceutil.Get(ctx)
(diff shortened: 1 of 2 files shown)


server/etcdserver/txn/txn.go
```diff
@@ -404,23 +375,33 @@ func executeTxn(ctx context.Context, lg *zap.Logger, txnWrite mvcc.TxnWrite, rt
...
 		}
-		if rr == nil || len(rr.KVs) == 0 {
-			return errors.ErrKeyNotFound
+		if req.IgnoreValue || req.IgnoreLease {
+			if resp == nil || len(resp.KVs) == 0 {
+				// ignore_{lease,value} flag expects previous key-value pair
+				return nil, nil, 0, errors.ErrKeyNotFound
+			}
 		}
+		prevKV = &resp.KVs[0]
```
iraj720, line 392: hi @serathius 
seems this line of code doesnt work just tested it

single put for a key that didnt exist with "req.PrevKv == true"
will be panic because of nil pointer dereference 

let me know if i can fix it after you merged with main
or you prefer to change it on your PR.

Thanks!

iraj720:
@serathius 
found some issue here : 
https://github.com/etcd-io/etcd/pull/16519/files#r1313451478

let me know if im wrong or i can fix it 
Thanks

iraj720:
saw the tests just now 
they are pointing to the same issue

server/etcdserver/txn/txn.go
serathius: Done

server/etcdserver/txn/txn.go
```diff
@@ -404,23 +373,36 @@ func executeTxn(ctx context.Context, lg *zap.Logger, txnWrite mvcc.TxnWrite, rt
...
+		val = prevKV.Value
 	}
-	if lease.LeaseID(req.Lease) != lease.NoLease {
-		if l := lessor.Lookup(lease.LeaseID(req.Lease)); l == nil {
-			return lease.ErrLeaseNotFound
+	if req.IgnoreLease {
+		leaseID = lease.LeaseID(prevKV.Lease)
+	} else if leaseID != lease.NoLease {
+		if l := lessor.Lookup(leaseID); l == nil {
+			return nil, nil, 0, lease.ErrLeaseNotFound
```
ahrtr, line 399: The following logic shouldn't be part of `checkPut`.

```
	if req.IgnoreValue {
		val = prevKV.Value
	}
	if req.IgnoreLease {
		leaseID = lease.LeaseID(prevKV.Lease)
	} 
	if !req.PrevKv {
		return val, nil, leaseID, nil
	}
```

serathius: I can implement it that way, but it results in weird duplication of logic to check whether prevKV is needed.

PR #16544, opened 2023-09-05
etcdserver: add skip param to RangeRequest

jjz921024:
Add skip param for skip the first N kvs returned

Current, range query only limit the number of returned. This feature allows user query in batches for a lot of special prefix key.

Signed-off-by: jjz921024  <hereisjjz@qq.com>


api/etcdserverpb/rpc.proto
```diff
@@ -494,6 +494,9 @@ message RangeRequest {
   // max_create_revision is the upper bound for returned key create revisions; all keys with
   // greater create revisions will be filtered away.
   int64 max_create_revision = 13 [(versionpb.etcd_version_field)="3.1"];
+
+  // skip is the skip the first N keys returned for the request.
+  int64 skip = 14;
```
fuweid, line 499: > . This feature allows user query in batches for a lot of special prefix key.

Can't follow your idea here. Would you try to use different range setting [key_start, key_end) to handle it?
If you know what the value of `skip` is, I think you can adjust the `key_start` to achieve it.

jjz921024: I know you meant. but sometime i can't get next `key_start`.

i try to implement redis  `scan` command base on etcd,  the `scan`  iterate data based on the cursor, 
for `scan 1000 count 10`, I don't know the `start_key` after the first 1000th element 
and i have to query the full amount of data to do this

fuweid: Not sure that I understand it correctly:

Assume that the target is [foo::000000, foo:999999). You can use Range API with limit 1000 and key-only option.
If there is mroe than 1000 key/values, the last key in the response is the target 1000th element you want.

Does it make sense to you?

jjz921024: I understand. This parameter is not necessary, but is useful for some scenarios.

If I want to implement a middleware that batches queries by cursor, and I don't want to maintain the cursor and key relationship.

The first query is for the first 10 key. It is ok, but I don't want to remember which key is the 10th. 
The second query I want to seek the 10th key by cursor. For example, the redis command `scan 10 limit 10`

fuweid: > I have to query the first 20 keys, then ignore the first 10 and return the query

If I understand correctly, the skip param is to let the server do that. :) 
Well, it might need to introduce new API for this if necessary.

Issue #16416, opened 2023-08-14
Flaking test case `TestV3WatchProgressOnMemberRestart`


ahrtr:
### Which github workflows are flaking?

test (linux-amd64-integration-1-cpu

### Which tests are flaking?

TestV3WatchProgressOnMemberRestart

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/5858992418/job/15884096085?pr=16414

### Reason for failure (if possible)

```
 v3_watch_test.go:965: Writing the same key 'foo' again after restarting the member
 v3_watch_test.go:969: Waiting for result
 v3_watch_test.go:976: Timed out waiting for the response
```

jmhbnz:
Quick note to confirm I was able to replicate this flake as follows, does look rare (1%):


1) Create individual test binary 

```bash
go test tests/integration/v3_watch_test.go -v -c -count 1 -run "^TestV3WatchProgressOnMemberRestart$"
```

2) Run test via stress

```bash
stress -p=8 ./integration.test

5s: 0 runs so far, 0 failures
10s: 0 runs so far, 0 failures
15s: 0 runs so far, 0 failures
20s: 0 runs so far, 0 failures
25s: 0 runs so far, 0 failures
30s: 0 runs so far, 0 failures
35s: 0 runs so far, 0 failures
40s: 0 runs so far, 0 failures
45s: 8 runs so far, 0 failures
50s: 8 runs so far, 0 failures
55s: 8 runs so far, 0 failures
1m0s: 8 runs so far, 0 failures
1m5s: 8 runs so far, 0 failures
1m10s: 8 runs so far, 0 failures
1m15s: 8 runs so far, 0 failures
1m20s: 9 runs so far, 0 failures
1m25s: 12 runs so far, 0 failures
1m30s: 16 runs so far, 0 failures
1m35s: 16 runs so far, 0 failures
1m40s: 16 runs so far, 0 failures
1m45s: 16 runs so far, 0 failures
1m50s: 16 runs so far, 0 failures
1m55s: 16 runs so far, 0 failures
2m0s: 17 runs so far, 0 failures
2m5s: 19 runs so far, 0 failures
2m10s: 24 runs so far, 0 failures
2m15s: 24 runs so far, 0 failures
2m20s: 24 runs so far, 0 failures
2m25s: 24 runs so far, 0 failures
2m30s: 24 runs so far, 0 failures
2m35s: 24 runs so far, 0 failures
[...]

PR #16547, opened 2023-09-05
test: de-flake test case TestV3WatchProgressOnMemberRestart

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/16416

The case may be blocked on sending progress notification, so may not be able to exit the goroutine.

### DIFF
--- a/tests/integration/v3_watch_test.go
+++ b/tests/integration/v3_watch_test.go
@@ -968,6 +968,8 @@ func TestV3WatchProgressOnMemberRestart(t *testing.T) {
 
 	t.Log("Waiting for result")
 	select {
+	case <-progressNotifyC:
+		t.Log("Progress notification received")
 	case err := <-errC:
 		t.Fatal(err)
 	case <-doneC:


tests/integration/v3_watch_test.go
```diff
@@ -968,6 +968,8 @@ func TestV3WatchProgressOnMemberRestart(t *testing.T) {
 
 	t.Log("Waiting for result")
 	select {
+	case <-progressNotifyC:
```
fuweid, line 971: the `lastWatchedModRevision` > 0 before restart member. It seems it's unlikely to send the item into `progressNotifyC`

fuweid: Added the sleep after restart and it can reproduce it.

```diff
--- a/tests/integration/v3_watch_test.go
+++ b/tests/integration/v3_watch_test.go
@@ -961,6 +961,7 @@ func TestV3WatchProgressOnMemberRestart(t *testing.T) {
        clus.Members[0].Restart(t)
        clus.Members[0].WaitOK(t)

+       time.Sleep(2 * time.Second)
        // write the same key again after the member restarted
        t.Log("Writing the same key 'foo' again after restarting the member")
        _, err = client.Put(ctx, "foo", "bar2")
```

ahrtr: yes, as long as the goroutine doesn't get blocked on sending progress notification, we are good.

ahrtr: let's keep the fix as simple as possible.

jmhbnz (approved):
LGTM

Issue #16266, opened 2023-07-18
server/v3/storage/mvcc TestHashKVWhenCompacting is flaking


Rajalakshmi-Girish:
### Which github workflows are flaking?

Job in our internal CI are flaking

### Which tests are flaking?

server/v3/storage/mvcc TestHashKVWhenCompacting

### Reason for failure (if possible)

The test is failing at https://github.com/etcd-io/etcd/blob/main/server/storage/mvcc/kvstore_test.go#L629
Seems like a test timeout issue.

This fails in `ppc64le` architecture.
```
[root@raji-workspace etcd]# go test -v go.etcd.io/etcd/server/v3/storage/mvcc -c -count=1
[root@raji-workspace etcd]#
[root@raji-workspace etcd]# /root/go/bin/stress ./mvcc.test -test.run TestHashKVWhenCompacting
5s: 0 runs so far, 0 failures
10s: 8 runs so far, 0 failures
15s: 8 runs so far, 0 failures

/tmp/go-stress-20230718T085107-1901309625
--- FAIL: TestHashKVWhenCompacting (10.38s)
    logger.go:130: 2023-07-18T08:51:15.864Z     INFO    kvstore restored        {"current-rev": 1}
    logger.go:130: 2023-07-18T08:51:16.040Z     INFO    compact tree index      {"revision": 9900}
    logger.go:130: 2023-07-18T08:51:17.078Z     INFO    finished scheduled compaction   {"compact-revision": 9900, "took": "1.038612654s", "hash": 2982592828}
    logger.go:130: 2023-07-18T08:51:17.079Z     INFO    storing new hash        {"hash": 2982592828, "revision": 9900, "compact-revision": -1}
    logger.go:130: 2023-07-18T08:51:17.193Z     INFO    compact tree index      {"revision": 9901}
    logger.go:130: 2023-07-18T08:51:17.193Z     INFO    finished scheduled compaction   {"compact-revision": 9901, "took": "439.747µs", "hash": 4092902393}
    logger.go:130: 2023-07-18T08:51:17.193Z     INFO    storing new hash        {"hash": 4092902393, "revision": 9901, "compact-revision": 9900}
    logger.go:130: 2023-07-18T08:51:17.308Z     INFO    compact tree index      {"revision": 9902}
    logger.go:130: 2023-07-18T08:51:17.308Z     INFO    finished scheduled compaction   {"compact-revision": 9902, "took": "451.774µs", "hash": 3442946395}
[...]

PR #16549, opened 2023-09-06
server/storage/mvcc: deflake TestHashKVWhenCompacting

fuweid:
The HashByRev-goroutines exit since receive `donec` notification. The Check-computed-hashes goroutine could not have chance to get the hash result and be stuck forever. We should add case for donec when we wait for hash result.

Closes: #16266


ahrtr (approved):
LGTM

thx @fuweid

PR #16558, opened 2023-09-08
[3.4] pkg/ioutil: deflake TestPageWriterRandom

fuweid:
The PageWriter has cache buffer so that it doesn't call the Writer until the cache is almost full. Since the data's length is random, the pending bytes should be always less than cache buffer size, instead of page size.

Fix: #16255

(cherry picked from commit fddd1add52b33649a99d7f756404924138344a10 #16268 )


ahrtr:
thx @fuweid for the quick fix.

ahrtr (approved):

PR #16559, opened 2023-09-08
Add diagrams for etcd internals and consistent read/write workflows

shyamjvs:
Brought this up on yesterday's community call and [here](https://kubernetes.slack.com/archives/C3HD8ARJ5/p1694111824706869).

As someone new to the etcd code base I've spend some time understanding its inner workings. Based on my understanding, prepared some diagrams explaining consistent read and write workflows (both for requests arriving at leader vs follower). It's a bit high-level (does not show the channels/structs/protobufs involved underneath) but has enough details to explain sequence of operations (memory/disk/network) and if they happen sequential-vs-parallel-vs-async. Goal is to help other new devs ramp up easily on the project with this understanding.

I'm also checking in the draw.io source code for the images so others can modify it later. Quick note on draw.io - it's a cross-platform graph drawing software developed in HTML5/JS that's free, open and commonly used for making flowcharts, process diagrams, UML diagrams, etc.

Happy to make any edits for correctness/clarity or move these to a better place.
As next step, I can also explain briefly these internals in a README file.

cc @ahrtr @serathius @wenjiaswe


Documentation/etcd-internals/diagrams/write_workflow_follower.png
ahrtr, line 1: Conceptually, it's the raft instead of etcdserver to forward the writing request to the leader. (The step 2).

shyamjvs: Thanks for pointing this. Fixed now, ptal.

ahrtr:
Overall looks good.
- Change "Inform replication is done" to "Notify etcdserver to apply committed entries";
- Change "Etcd (main)" to "etcdserver";
- Change "Applies the transaction to MVCC ..." to "Applies the entries to MVCC store and backend storage"; 
- Note that asyncing bbolt is independent to the application workflow.

shyamjvs:
> Applies the entries to MVCC store and backend storage

Here by backend storage do you mean v2 store or boltdb? Asklng because is it's the latter, that has already been represented as an async operation between mvcc and boltdb.

jmhbnz (commented):
Hey @shyamjvs - Great work putting these diagrams together!

I think these are a great resource and I would like to see us publish these a little broader.  I would like to propose we put together a short article and publish this with the diagrams on https://etcd.io/blog or under https://etcd.io/docs/v3.5/learning.

That would mean closing this pr and opening a pr under https://github.com/etcd-io/website but I do think it will be worthwhile to put these on the website and make them easy to refer and link to.

@ahrtr what do you think on moving this to the website?

ahrtr:
> Here by backend storage do you mean v2 store or boltdb? Asklng because if it's the latter, that has already been represented as an async operation between mvcc and boltdb. And iiuc we're in the process of deprecating v2 store - so it'll become obsolete soon?

I just mean the backend storage, which is independent to specific implementation (bbolt for now). Note that the async operation is just a performance adjustment for bbolt, and it has nothing to do with the core (application) workflow. Actually the application workflow has already written the data to backend storage (again bbolt for now), it just doesn't sync (commit) the data to disk due to performance concern.

> I would like to propose we put together a short article and publish this with the diagrams on https://etcd.io/blog or under https://etcd.io/docs/v3.5/learning.

Yes, I had the same thought when I replied it in the slack thread: `"We can work together to make it part of the official doc or blog post.`".  I can provide such a article, but I don't have that bandwidth in near future. So I don't mind to merge this PR firstly, and move then to the website (under the learning as you mentioned) later and with a article later in a separate PR. If anyone want to create an article(summary), please feel free to do it, and I am happy to review it.

shyamjvs:
Thanks for clarification @ahrtr - updated the PR with changes you suggested.

@jmhbnz I'd definitely love to port this over to blog. As @ahrtr was saying - I can add a readme with a short description explaining these workflows and we can then move everything together over to the other repo. Sgty?

Documentation/etcd-internals/diagrams/consistent_read_workflow.png
ahrtr, line 1: The protocol between leader and follower/candidate is complicated. Let's keep the read workflow as simple as possible.
- Client sends a read request to etcdserver;
- etcdserver ensures it's local data is in sync with the leader. Specifically, it ensures its applied Index is in sync with the leader's commit Index;
   - The node tries to get the commit index from the leader. [ If current node isn't the leader, it forwards the request to the leader];
   - If its applied index < leader's commitIndex; wait until it's in sync
- Respond to the client if it doesn't timed out.

shyamjvs: To clarify - did you want me to drop the details of communication between follower and leader? The reason I showed read request workflow from a follower node is because that is the more interesting case, and the purpose of showing those arrows to leader is to capture network roundtrips that happen in the sync path. Lemme know wyt.

ahrtr: Suggest to only display what I mentioned in above comment.
- MsgHeartbeatResp is just an acknowledgment of the heartbeat message. So I am afraid "`Follower confirms its committedIndex in heartbeat response`" isn't correct. Pls just simply the workflow as follower requesting & leader responding the commit Index;
- Remove the WAL from the diagram as it's unrelated details.

shyamjvs: Got it, thanks for clarifying. I've simplified it ptal now.

ahrtr: (step 6) raft doesn't wait for anything. Pasted my previous comment again: `etcdserver ensures it's local data is in sync with the leader. Specifically, it ensures its applied Index is in sync with the leader's commit Index;`

shyamjvs: > Note that etcd reads data from backend storage, which is already applied data(entries), and respond to clients.

So this part I understand (shown in step-8).

Maybe wait is the wrong word for step-6, but I'm trying to show that before step-7 raft needs to catch up with any committed entries till readIndex. Is there a better to show this on your mind?

shyamjvs: I discussed with @chaochn47 and he helped me understand what you're thinking. PTAL now.

ahrtr: - We should clearly & accurately differentiate two concepts: `committed` and `applied`. etcdserver only care about (and wait for) the appliedIndex when serving linearizability requests. 
- Writing the MVCC store and backend storage is part of the writing workflow, pls do not get it mixed with the read workflow.

shyamjvs: I see your point about not showing any write/apply operations that may need to happen amidst the read operation, given they're carried out in background. I've changed it a bit to only show read-specific path, ptal now.

ahrtr (approved):
Overall looks good to me.

Thanks.

jmhbnz (approved):
LGTM - Though as mentioned above, as a next step I would also like to have a pr for etcd-io/website to add these diagrams as a learning resource.

After some further conversations a blog post probably isn't the right approach, I think having these in a page under https://etcd.io/docs/v3.5/learning makes sense to make them more accessible and consumable.

PR #16572, opened 2023-09-11
test: simply the expected output for some curl test cases

ahrtr:
Link to https://github.com/etcd-io/etcd/pull/16454

Random spaces may be inserted between fields in a marshalled json response in grpc-gateway v2, so simply the expected output to make it work for both v1 and v2.

### DIFF
--- a/tests/e2e/v3_curl_auth_test.go
+++ b/tests/e2e/v3_curl_auth_test.go
@@ -121,7 +121,7 @@ func testCurlV3Auth(cx ctlCtx) {
 		if err = e2e.CURLPost(cx.epc, e2e.CURLReq{
 			Endpoint: "/v3/kv/put",
 			Value:    string(putreq),
-			Expected: expect.ExpectedResponse{Value: "error"},
+			Expected: expect.ExpectedResponse{Value: "etcdserver: user name is empty"},
 		}); err != nil {
 			cx.t.Fatalf("testCurlV3Auth failed to put without token (%v)", err)
 		}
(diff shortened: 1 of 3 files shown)


serathius:
Could we just use json parsing for JSON API? For example create something like EtcdctlV3  that has [spawnJsonCommand](https://github.com/etcd-io/etcd/blob/6544add9615107db55b7a57860c3ce2b345f6879/tests/framework/e2e/etcdctl.go#L663-L676) to parse JSON responses.

As we add more tests, the more complicated they are, and the more abstraction leak we get. Abstraction leak happens when abstractions you created doesn't correctly reflect the modeled problem. Like in those tests, we are using "find matching line" logic to analyse JSON responses, it doesn't work as JSON can have arbitrary whitespaces injected. 

Until we fix our abstraction we will continue to need to patch our tests like in this PR.

ahrtr:
> This PR lowers quality of testing by skipping testing some fields, which we are no longer able to cover using a naive "find matching line" validation.
- Partially agree with this.  Updated the verification on the response for `/v3/kv/txn`. 
- For other verification, we only need to verify what the etcdserver returns. Use "/v3/election/resign" as an example, etcdserver just returns an [ErrMissingLeaderKey](https://github.com/etcd-io/etcd/blob/6544add9615107db55b7a57860c3ce2b345f6879/server/etcdserver/api/v3election/election.go#L28), but gRPC adds some other info such as "code": 2. We only need to verify what the etceserver returns. 
- Note that we are NOT testing the full functionalities. We just need to ensure the RESTful APIs are not broken when we upgrade grpc-gateway from v1 to v2.

ahrtr:
> Until we fix our abstraction we will continue to need to patch our tests like in this PR.

Pls do not make it big on this PR. If you have any additional thoughts, please raise a separate ticket to discuss; and anyone is welcome to refactor or improve existing test cases.

jmhbnz (approved):
Agree it would be best to more formally parse the json responses.

With that said, provided we have a follow-up issue created to track the work I am comfortable for it to be addressed at a later date.

fuweid (approved):
LGTM

serathius (approved):

ahrtr:
thx all for the review!

Issue #11124, opened 2019-09-05
etcd 3.4.0 server expects client cert when client-cert-auth = false


ae6rt:
For a single node etcd 3.4.0 server:

## Server version

```
etcd Version: 3.4.0
Git SHA: 898bd1351
Go Version: go1.12.9
Go OS/Arch: linux/amd64
```

## Client version

```
etcdctl version: 3.4.0
API version: 3.4
```

## Expected result

When client-cert-auth is false, a client need not present a certificate to the server.

(N.B. The expected result is the observed result for etcd 3.3.15).

## Observed result

The server expects a client cert

```
etcdctl --cacert=cabundle.pem --endpoints=192.168.99.1:2379  member list
{"level":"warn","ts":"2019-09-05T09:01:47.701-0700","caller":"clientv3/retry_interceptor.go:61","msg":"retrying of unary invoker failed","target":"endpoint://client-f5825ff3-002f-4f77-976c-a506dfdeef15/192.168.99.1:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest connection error: connection error: desc = \"transport: authentication handshake failed: remote error: tls: bad certificate\""}
Error: context deadline exceeded
```

The server logs accordingly

```
{"level":"info","ts":"2019-09-05T15:32:25.817Z","caller":"embed/etcd.go:708","msg":"starting with client TLS","tls-info":"cert = /tls/server.pem, key = /tls/server-key.pem, trusted-ca = /tls/ca.pem, client-cert-auth = false, crl-file = ","cipher-suites":[]}
...
{"level":"warn","ts":"2019-09-05T16:01:45.237Z","caller":"embed/config_logging.go:279","msg":"rejected connection","remote-addr":"172.17.0.1:46882","server-name":"","error":"tls: client didn't provide a certificate"}
```

When the client presents a certificate, the operation succeeds

```
etcdctl  --cacert ~/lib/easycert-ca/cabundle.pem --endpoints=192.168.99.1:2379  --cert client-cert.pem --key client-key.pem  member list
521b45b49ab8a3b9, started, 49c0b6d3b201, https://49c0b6d3b201:2380, https://49c0b6d3b201:2379, false
```

## Server configuration

```
readonly HOSTN=$(hostname)

# Establish defaults
export ETCD_NAME=${ETCD_NAME:-${HOSTN}}
export ETCD_DATA_DIR=${ETCD_DATA_DIR:-/etc-data}
[...]

PR #16585, opened 2023-09-11
transport: separate implied client certificate usage with trusted ca

jmhbnz:
The `--trusted-ca-file` flag currently implies using client certificates. However, this is undocumented behavior and adding a CA may has led to some users unexpectedly having client cert auth enabled.

This pr revives https://github.com/etcd-io/etcd/pull/11703 to make `--client-cert-auth` a more explicit boolean flag and is intended to address https://github.com/etcd-io/etcd/issues/11124.

An alternative approach would be to update our documentation to make it clear the two flags are actually coupled.

### DIFF
--- a/client/pkg/transport/listener.go
+++ b/client/pkg/transport/listener.go
@@ -511,7 +511,7 @@ func (info TLSInfo) ServerConfig() (*tls.Config, error) {
 	}
 
 	cfg.ClientAuth = tls.NoClientCert
-	if info.TrustedCAFile != "" || info.ClientCertAuth {
+	if info.ClientCertAuth {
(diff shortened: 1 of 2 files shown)


ahrtr:
I think it's safer and more reasonable to require the client side to provide a valid certificate in HTTPS scheme. So suggest to update the doc, and not to change the behavior which has been there for a long time.

jmhbnz:
> I think it's safer and more reasonable to require the client side to provide a valid certificate in HTTPS scheme. So suggest to update the doc, and not to change the behavior which has been there for a long time.

Thanks for the feedback, it makes sense and I'm happy to change approach and handle it through a docs update.

Will give it a couple of days and see if anyone else has any thoughts before raising docs clarification.

Issue #15549, opened 2023-03-22
{markdown_you,govet_shadow,unparam,staticcheck,..}_pass fail in main branch


fuweid:
### Which github workflows are flaking?

Currently, there is no workflow to run `{markdown_you,govet_shadow,unparam,staticcheck,..}_pass`.

### Which tests are flaking?

{markdown_you,govet_shadow,unparam,staticcheck,..}_pass fail in main branch

For instance,

```bash
PASSES="nakedret" TIMEOUT=1m CPU=4 ./scripts/test.sh 2>&1
go: downloading golang.org/x/crypto v0.7.0
Running with --race --cpu=4
Starting at: Wed Mar 22 18:02:21 HKT 2023

'nakedret' started at Wed Mar 22 18:02:21 HKT 2023
% (cd tools/mod && 'go' 'install' 'github.com/alexkohler/nakedret')
% (cd api && '/home/fuwei/go/bin/nakedret' './...')
% (cd tools/mod && 'go' 'install' 'github.com/alexkohler/nakedret')
% (cd pkg && '/home/fuwei/go/bin/nakedret' './...')
% (cd tools/mod && 'go' 'install' 'github.com/alexkohler/nakedret')
% (cd client/pkg && '/home/fuwei/go/bin/nakedret' './...')
stderr: transport/keepalive_listener.go:107: Accept naked returns on 8 line function
stderr: transport/listener.go:113: newKeepAliveListener naked returns on 11 line function
stderr: transport/listener.go:216: SelfCert naked returns on 138 line function
stderr: transport/listener.go:226: SelfCert naked returns on 138 line function
stderr: transport/listener.go:231: SelfCert naked returns on 138 line function
stderr: transport/listener.go:235: SelfCert naked returns on 138 line function
stderr: transport/listener.go:245: SelfCert naked returns on 138 line function
stderr: transport/listener.go:257: SelfCert naked returns on 138 line function
stderr: transport/listener.go:295: SelfCert naked returns on 138 line function
stderr: transport/listener.go:306: SelfCert naked returns on 138 line function
stderr: transport/listener.go:316: SelfCert naked returns on 138 line function
stderr: transport/listener.go:326: SelfCert naked returns on 138 line function
stderr: transport/listener.go:337: SelfCert naked returns on 138 line function
```

```bash
PASSES="ineffassign" TIMEOUT=1m CPU=4 ./scripts/test.sh 2>&1
Running with --race --cpu=4
[...]

PR #16601, opened 2023-09-17
*: fix nakedret lint

fuweid:
Part of #15549

### DIFF
--- a/client/pkg/transport/keepalive_listener.go
+++ b/client/pkg/transport/keepalive_listener.go
@@ -101,10 +101,10 @@ type tlsKeepaliveListener struct {
 
 // Accept waits for and returns the next incoming TLS connection.
 // The returned connection c is a *tls.Conn.
-func (l *tlsKeepaliveListener) Accept() (c net.Conn, err error) {
-	c, err = l.Listener.Accept()
+func (l *tlsKeepaliveListener) Accept() (net.Conn, error) {
+	c, err := l.Listener.Accept()
 	if err != nil {
-		return
+		return nil, err
 	}
 
 	c = tls.Server(c, l.config)
--- a/server/etcdserver/bootstrap_test.go
+++ b/server/etcdserver/bootstrap_test.go
@@ -216,27 +216,32 @@ func TestBootstrapBackend(t *testing.T) {
 	}
 }
 
-func createDataDir(t *testing.T) (dataDir string, err error) {
+func createDataDir(t *testing.T) (string, error) {
+	var err error
+
 	// create the temporary data dir
-	dataDir = t.TempDir()
+	dataDir := t.TempDir()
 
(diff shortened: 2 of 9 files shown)


jmhbnz (approved):
LGTM - Thanks @fuweid 👍🏻

ahrtr:
Almost all the named return parameters are bad examples. The only exception is the one used in wal.go, it makes sense, because it makes the returned value more clear.

So suggest to convert all the named return parameter to naked return parameters except the wal.go. 

Reference: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

fuweid:
@ahrtr sorry for the late reply. The nakedret is used to detect the named return lint issue. The code change is caused by nakedret. I think we can keep the named return var for `retErr` check in defer or interface documentation. In this patch, you can see SelfCert is long function. Multiple naked return branches are far away from the declaration.  IMO, it's hard to review the code when there is any change. And even if err isn't nil, the function still return non-empty value. Even though the func-invoker should not use the non-empty value if the err isn't nil, it's still confuse.

Maybe we can improve the https://github.com/alexkohler/nakedret/blob/564f756361f1c6a648f759199ab7c03c59cae6dc/cmd/nakedret/main.go#L16 to config max function length which the nakedret tolerances. Or if it's not useful in etcd repo, we can consider to remove the lint.

cc @jmhbnz @serathius

client/pkg/transport/keepalive_listener.go
```diff
@@ -104,7 +104,7 @@ type tlsKeepaliveListener struct {
 func (l *tlsKeepaliveListener) Accept() (c net.Conn, err error) {
```
ahrtr, line 104:
```suggestion
func (l *tlsKeepaliveListener) Accept() (net.Conn, error) {
```

ahrtr:
To be clearer, the PR makes sense to me. I was requesting to make change something like https://github.com/etcd-io/etcd/pull/16601#discussion_r1328108604. The only exception is the named return parameter used in wal.go, it makes sense, so let's keep it as it's.

ahrtr:
Sorry for the confusion, I should be more clearer in the first place.

server/etcdserver/bootstrap_test.go
```diff
@@ -245,7 +250,7 @@ func prepareData(cfg config.ServerConfig) (err error) {
 func createWALFileWithSnapshotRecord(cfg config.ServerConfig, snapshotTerm, snapshotIndex uint64) (err error) {
```
fuweid, line 250: The `err` is used in defer so keep it here.

ahrtr: ack

ahrtr (approved):
LGTM

thx @fuweid

serathius (approved):

PR #16602, opened 2023-09-17
walk doc cleanup

redwrasse:
Improves documentation of `walk` method in `server/storage/mvcc/key_index.go`.


jmhbnz (approved):
LGTM

Note @redwrasse in my view this pr is marginal in terms of making improvement, and does distract / context switch reviewers away from other critical pr's that need a lot of attention. I suggest holding off raising future pr's that only minor tweak comment wording until we have some more capacity as a project.

Issue #16610, opened 2023-09-19
[lint] make verify-lint doesn't verify {api,server,client,...} modules 


fuweid:
### Which github workflows are flaking?

.github/workflows/static-analysis.yaml

### Which tests are flaking?

make verify-lint

### Reason for failure (if possible)

golanglint-ci doesn't support multiple modules in one repo. We need to chdir and run the lint for each module.

```bash
# in root
➜  etcd git:(main) make verify-lint
golangci-lint run --config tools/.golangci.yaml

# in server module
➜  server git:(main) golangci-lint run --config ../tools/.golangci.yaml
storage/datadir/datadir_test.go:20: File is not `goimports`-ed with -local go.etcd.io (goimports)
        "github.com/stretchr/testify/assert"
storage/backend/backend_bench_test.go:22: File is not `goimports`-ed with -local go.etcd.io (goimports)
        "github.com/stretchr/testify/assert"
storage/backend/backend_test.go:24: File is not `goimports`-ed with -local go.etcd.io (goimports)
        "github.com/stretchr/testify/assert"
storage/backend/hooks_test.go:22: File is not `goimports`-ed with -local go.etcd.io (goimports)
        "github.com/stretchr/testify/assert"
mock/mockwait/wait_recorder.go:37:33: unused-parameter: parameter 'id' seems to be unused, consider removing or renaming it as _ (revive)
func (w *waitRecorder) Register(id uint64) <-chan any {
                                ^
```

### Work items

- [x] goimports https://github.com/etcd-io/etcd/pull/16628
- [x] ineffassign https://github.com/etcd-io/etcd/pull/16605
- [x] revive https://github.com/etcd-io/etcd/pull/16634
- [x] staticcheck https://github.com/etcd-io/etcd/pull/16626
- [ ] stylecheck https://github.com/etcd-io/etcd/pull/16638
- [ ] unused https://github.com/etcd-io/etcd/pull/16638
- [ ] unconver https://github.com/etcd-io/etcd/pull/16638

PR #16622, opened 2023-09-20
*: fix unparam lint

fuweid:
Part of #16610 


tools/.golangci.yaml
```diff
@@ -24,6 +24,7 @@ linters:
     - stylecheck
     - unused
     - unconvert # Remove unnecessary type conversions
+    - unparam
```
serathius, line 27: Note, this is new check that was not enabled before, so this is not part of https://github.com/etcd-io/etcd/issues/16610 and should start from community discussion whether we want to adopt `unparam` check at all.

fuweid:
ping @ahrtr @serathius @jmhbnz

server/etcdmain/grpc_proxy.go
```diff
@@ -538,8 +538,7 @@ func mustHTTPListener(lg *zap.Logger, m cmux.CMux, tlsinfo *transport.TLSInfo, c
 func mustNewHTTPClient(lg *zap.Logger) *http.Client {
 	transport, err := newHTTPTransport(grpcProxyCA, grpcProxyCert, grpcProxyKey)
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
+		lg.Fatal("failed to new http transport", zap.Error(err))
```
serathius, line 541: This is not the same, it changes the exit code and the output.

client/internal/v2/client_test.go
```diff
@@ -499,7 +499,7 @@ func (f fakeCancelContext) Done() <-chan struct{} {
 func (f fakeCancelContext) Err() error        { return errFakeCancelContext }
 func (f fakeCancelContext) Value(key any) any { return 1 }
 
-func withTimeout(parent context.Context, timeout time.Duration) (
+func withTimeout(parent context.Context, _timeout time.Duration) (
```
ahrtr, line 502: a quick question, why add a prefix "_" for the parameter? Is it required by the linter?

tests/e2e/ctl_v3_snapshot_test.go
```diff
@@ -447,5 +447,6 @@ func hasKVs(t *testing.T, ctl *e2e.EtcdctlV3, kvs []testutils.KV, currentRev int
 		require.Equal(t, int64(baseRev+i), v.Kvs[0].CreateRevision)
 		require.Equal(t, int64(baseRev+i), v.Kvs[0].ModRevision)
 		require.Equal(t, int64(1), v.Kvs[0].Version)
+		require.True(t, int64(currentRev) >= v.Kvs[0].ModRevision)
```
ahrtr, line 450: why add this in this PR although it isn't a big problem?

tests/integration/revision_test.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 func TestRevisionMonotonicWithLeaderPartitions(t *testing.T) {
-	testRevisionMonotonicWithFailures(t, 11*time.Second, func(clus *integration.Cluster) {
+	testRevisionMonotonicWithFailures(t, 12*time.Second, func(clus *integration.Cluster) {
```
ahrtr, line 32: Why change 11 to 12? :)

server/etcdmain/grpc_proxy.go
fuweid: `lg.Fatal` exits with code 1. Yes, I should not update the output.
Updated with removing the `lg *zap.Logger` since no function uses `lg`.

client/internal/v2/client_test.go
fuweid: If the function has unused argument, the lint will blame it. Using `_X` is to ignore the lint error and it can be used in the future.

tests/e2e/ctl_v3_snapshot_test.go
fuweid: The `currentRev` as argument is not used by function. The lint blames this.

tests/integration/revision_test.go
fuweid: The `testRevisionMonotonicWithFailures` function is always called with `11*time.Second`. The lint blames this and it suggests us to use fixed number in the function instead of argument.

```
revision_test.go:77:54: `testRevisionMonotonicWithFailures` - `testDuration` always receives `11 * time.Second (11000000000)` (unparam)
func testRevisionMonotonicWithFailures(t *testing.T, testDuration time.Duration, injectFailures func(clus *integration.Cluster)) {
                                                     ^
```

Most of changes are related to `X` always receives `Y`.

ahrtr: thx & interesting

jmhbnz (approved):
LGTM - Great to tidy and tighten this up, thanks @fuweid.

fuweid:
Is it ready to move on?

fuweid:
ping @serathius @ahrtr

ahrtr:
> ping @serathius @ahrtr

Sorry for the delay. I already approved the PR, and leave it to @serathius to take another look and merge, since he raised comment/concern https://github.com/etcd-io/etcd/pull/16622#discussion_r1331276752 previously

fuweid:
Thanks for the comment.

As far as I know, the test script has been refactored several times and I believe we miss some of them. The unparam_pass isn't introduced in this pr. The pr just enables it. 

The lint did find some unparam issues. It has strict rule, especially when we write test code. But it could help us to find some issues in test code.

In my opinion, based on the fact that it's not new lint in the repo, we can enable it and evaluate it for a while. If it's annoying, we can consider to disable it.

fuweid:
ping @serathius ~

serathius (approved):

fuweid:
Thanks! @serathius

PR #16591, opened 2023-09-13
Update GRPC dependency

rsafonseca:
Fixes testing packages compatibility issues with grpc from 1.55.0 onwards
Resolves https://github.com/etcd-io/etcd/issues/15877

Issue #16216, opened 2023-07-11
Improving client 


CaojiamingAlan:
### 1. Expose more client call options
Most of the [gRPC call options](https://github.com/etcd-io/etcd/blob/c853b9c337c20505eff494ede2239caf334a4e40/client/v3/options.go#L24-L57) of clientV3 cannot be configured by users (while some of them can, like `MaxCallSendMsgSize`) , and results in issues like 
- #13344 
- #14631

We should provide users with more flexibility to make RPC calls.

### 2. Replace the customized retry interceptor with RetryPolicy
Old gRPC versions do not have retry options, and etcd client implemented retry through the [retry interceptor](https://github.com/etcd-io/etcd/blob/c853b9c337c20505eff494ede2239caf334a4e40/client/v3/retry_interceptor.go#L40). #12671 had replaced the balancer logic with the gRPC solution, and left a [TODO](https://github.com/etcd-io/etcd/blob/c853b9c337c20505eff494ede2239caf334a4e40/client/v3/client.go#L240) here. Maybe we can finish this replacement now.

### 3. (Maybe) Set `defaultWaitForReady` to True
I'm not sure about this. This is motivated by #14631. Actually, by my experiments, not only `NewSession()` but common usages like `Get()` may hang: whenever the call is directed to the `SIGSTOP`-ed server (this always happens since it uses round-robin), it hangs indefinitely waiting for the server to be ready. Neither did I find anywhere configures a default timeout.

It seems the motivation to set [WaitForReady](https://grpc.io/docs/guides/wait-for-ready/) to `True` is for 
1. preventing retry requests from overwhelming the network  #8122
2. minimizing client request error responses due to transient failures

(FYI `WaitForReady` is the opposite of `FailFast`, with the latter one being deprecated)

But if once we have the retry and WaitForReady configs exposed, we can change the default behavior to fail fast, and users may turn it on manually, or set a proper retry limit.

I do see [caveats](https://github.com/etcd-io/etcd/blob/c853b9c337c20505eff494ede2239caf334a4e40/client/internal/v2/README.md?plain=1#L110) of clientV2 but I think we may try to solve it in V3.

### Why is this needed?

To provide more flexibility and reduce code volume

serathius:
cc @ahrtr @serathius

ahrtr:
There are 3 different feature requests, let's breakdown them into three different tickets.

[...]

PR #15940, opened 2023-05-23
[release-3.5]pkg/types: Support Unix sockets in NewURLS

HubertZhang:
backport of #12469 to release 3.5, which enables using unix socket setup an embed server

PR #16625, opened 2023-09-20
[3.5] Bump grpc-go to1.47 (and fix the connection-string format)

chaochn47:
Split gRPC-go dependency update in release-3.5 into multiple PRs based on https://github.com/etcd-io/etcd/pull/16591#issuecomment-1724059411. 

Backport #14125

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -8,14 +8,14 @@ require (
 	github.com/golang/protobuf v1.5.2
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	google.golang.org/genproto v0.0.0-20210602131652-f16073e35f0c
-	google.golang.org/grpc v1.41.0
+	google.golang.org/grpc v1.47.0
 )
 
 require (
 	golang.org/x/net v0.17.0 // indirect
 	golang.org/x/sys v0.13.0 // indirect
 	golang.org/x/text v0.13.0 // indirect
-	google.golang.org/protobuf v1.26.0 // indirect
+	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -790,7 +790,7 @@ func (e *Etcd) grpcGatewayDial(splitHttp bool) (grpcDial func(ctx context.Contex
 	addr := sctx.addr
 	if network := sctx.network; network == "unix" {
 		// explicitly define unix network for gRPC socket support
-		addr = fmt.Sprintf("%s://%s", network, addr)
+		addr = fmt.Sprintf("%s:%s", network, addr)
 	}
 
 	opts := []grpc.DialOption{grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(math.MaxInt32))}
--- a/api/go.sum
+++ b/api/go.sum
@@ -7,7 +7,10 @@ github.com/cespare/xxhash/v2 v2.1.1/go.mod h1:VGX0DQ3Q6kWi7AoAeZDth3/j3BFtOZR5XL
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=
 github.com/cncf/udpa/go v0.0.0-20201120205902-5459f2c99403/go.mod h1:WmhPx2Nbnhtbo57+VJT5O0JRkEi1Wbu0z5j0R8u5Hbk=
-github.com/cncf/xds/go v0.0.0-20210805033703-aa0b78936158/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
+github.com/cncf/udpa/go v0.0.0-20210930031921-04548b0d99d4/go.mod h1:6pvJx4me5XPnfI9Z40ddWsdw2W/uZgQLFXToKeRcDiI=
+github.com/cncf/xds/go v0.0.0-20210922020428-25de7278fc84/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
+github.com/cncf/xds/go v0.0.0-20211001041855-01bcc9b48dfe/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
+github.com/cncf/xds/go v0.0.0-20211011173535-cb28da3451f1/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
 github.com/coreos/go-semver v0.3.0 h1:wkHLiw0WNATZnSG7epLsujiMCgPAc9xhjJ4tgnAxmfM=
 github.com/coreos/go-semver v0.3.0/go.mod h1:nnelYz7RCh+5ahJtPPxZlU+153eP4D4r3EedlOD2RNk=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
@@ -16,7 +19,7 @@ github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.m
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
 github.com/envoyproxy/go-control-plane v0.9.9-0.20201210154907-fd9021fe5dad/go.mod h1:cXg6YxExXjJnVBQHBLXeUAgxn2UodCpnH306RInaBQk=
 github.com/envoyproxy/go-control-plane v0.9.9-0.20210217033140-668b12f5399d/go.mod h1:cXg6YxExXjJnVBQHBLXeUAgxn2UodCpnH306RInaBQk=
-github.com/envoyproxy/go-control-plane v0.9.10-0.20210907150352-cf90f659a021/go.mod h1:AFq3mo9L8Lqqiid3OhADV3RfLJnjiw63cSpi+fDTRC0=
+github.com/envoyproxy/go-control-plane v0.10.2-0.20220325020618-49ff273808a1/go.mod h1:KJwIaB5Mv44NWtYuAOFCVOjcI94vtpEz2JU/D2v6IjE=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/ghodss/yaml v1.0.0/go.mod h1:4dBDuWmgqj2HViK6kFavaiC9ZROes6MMH2rRYeMEF04=
 github.com/gogo/protobuf v1.3.2 h1:Ov1cvc58UF3b5XjBnZv7+opcTcQFZebYjWzi34vdm4Q=
@@ -42,8 +45,9 @@ github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMyw
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
-github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
+github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/uuid v1.1.2/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/grpc-ecosystem/grpc-gateway v1.16.0 h1:gmcG1KaJ57LophUzW0Hy8NmPhnMZb4M0+kPpLofRdBo=
 github.com/grpc-ecosystem/grpc-gateway v1.16.0/go.mod h1:BDjrQk3hbvj6Nolgz8mAMFbcEtjT1g+wF4CSlocrBnw=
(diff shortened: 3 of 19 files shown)


chaochn47:
@ahrtr @serathius @jmhbnz @fuweid

fuweid (approved):
LGTM

serathius:
Is there any security issue we are trying to fix by backporting?

3 people complaining about dependency issue doesn't seem like a sensible reason for backport. https://github.com/etcd-io/etcd/issues/15877#issuecomment-1714158645

cc @liggitt

ahrtr:
> Is there any security issue we are trying to fix by backporting?

Just spending 2 minutes to search "CVE" or "security" in [grpc-go/releases](https://github.com/grpc/grpc-go/releases). You will get the "YES" answer.



> 3 people complaining about dependency issue doesn't seem like a sensible reason for backport. [#15877 (comment)](https://github.com/etcd-io/etcd/issues/15877#issuecomment-1714158645)

Why not? It's prevent some applications from integrating with etcd!

ahrtr (approved):
LGTM

thx @chaochn47

serathius (changes requested):
> > Is there any security issue we are trying to fix by backporting?
> 
> Just spending 2 minutes to search "CVE" or "security" in [grpc-go/releases](https://github.com/grpc/grpc-go/releases). You will get the "YES" answer.

Spent 10 seconds, didn't find anything worth backporting. Those "security" or "cve" are not related to grpc are unrelated to etcd security posture. For example a feature adding min/max TLS, or bump grpc dependency to address [CVE 2022-32149](https://www.cve.org/CVERecord?id=CVE-2022-32149).

> > 3 people complaining about dependency issue doesn't seem like a sensible reason for backport. [#15877 (comment)](https://github.com/etcd-io/etcd/issues/15877#issuecomment-1714158645)
> 
> Why not? It's prevent some applications from integrating with etcd!

What's a weak reason, you are comparing 3 applications vs all the remaining applications.

ahrtr:
> you are comparing 3 applications vs all the remaining applications.

It is not about 3 applications. It's about all applications which depends on a newer grpc version and etcd client SDK.

server/embed/etcd.go
```diff
@@ -790,7 +790,7 @@ func (e *Etcd) grpcGatewayDial(splitHttp bool) (grpcDial func(ctx context.Contex
 	addr := sctx.addr
 	if network := sctx.network; network == "unix" {
 		// explicitly define unix network for gRPC socket support
-		addr = fmt.Sprintf("%s://%s", network, addr)
```
liggitt, line 793: is this indicative of changes / breaks that will happen for etcd clients if this update merges?

liggitt:
from https://github.com/etcd-io/etcd/pull/16591#issuecomment-1728272209

> In order to use Outlier Detection, release-3.5 should be at least [gRPC-go v1.50.0](https://github.com/grpc/grpc-go/releases/tag/v1.50.0) onward. Working on it..

is it a goal to take new grpc features back to etcd release branches? propagating this many grpc updates to all etcd clients on patch updates is surprising to me

liggitt:
when is etcd 1.6.0 targeted? major bumps in grpc lib versions that pull in breaking changes like package renames seem better to push to consumers in a minor release

server/embed/etcd.go
ahrtr: No, it doesn't impact K8s at all; it doesn't impact any etcd clients at all. It just changes the way how the grpc-gateway connects to the backend etcdserver. The change is transparent to any clients.

ahrtr:
> when is etcd 1.6.0 targeted? major bumps in grpc lib versions that pull in breaking changes like package renames seem better to push to consumers in a minor release

Yes, that hit the point. Usually it isn't recommended to add any feature to a stable release (etcd 3.5 in this case), and users are supposed to integrate with etcd 3.6 to resolve such grpc impactable issue. But unfortunately, based on the [3.6 roadmap](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/roadmap.md#v360), I don't think we will release 3.6.0 in near future (e.g. at least > 0.5 year).

serathius:
> > when is etcd 1.6.0 targeted? major bumps in grpc lib versions that pull in breaking changes like package renames seem better to push to consumers in a minor release
> 
> Yes, that hit the point. Usually it isn't recommended to add any feature to a stable release (etcd 3.5 in this case), and users are supposed to integrate with etcd 3.6 to resolve such grpc incompatible issue. But unfortunately, based on the [3.6 roadmap](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/roadmap.md#v360), I don't think we will release 3.6.0 in near future (e.g. at least > 0.5 year).

Agree that it will be hard to release v3.6 in near future, however we should avoid the vicious cycle of backport anything because next release will not come.

ahrtr:
> is it a goal to take new grpc features back to etcd release branches?

The goal is to resolve users' pain points that any applications depending on a new grpc version are not able to integrate with etcd 3.5 sdk. gRPC is a very basic lib, so I imagine it will impact lots of applications which depend on etcd SDK as well.

> however we should avoid the vicious cycle of backport anything because next release will not come.

See my comment above. We should handle it from pragmatic propective and case by case.

ahrtr:
cc @dfawley

chaochn47:
> is it a goal to take new grpc features back to etcd release branches? propagating this many grpc updates to all etcd clients on patch updates is surprising to me. 

No, etcd clients does not use this gRPC feature while consumer of etcd client could use this experimental feature on demand.

rsafonseca:
@liggitt the goal is not to introduce new features, it's to maintain interoperability support for the latest stable branch (3.5)
There is no date for the release of 3.6 and this point has come up over and over again.

To be clear, the proposed changes should not be breaking, and the goal is not to add new features, simply to maintain the 3.5 branch which is the latest branch currently stable in ETCD, which is currently causing trouble to anyone who integrates with ETCD client and have other updated dependencies which don't use an old version of GRPC.

mhutchinson:
Hi all, we're actively looking for this fix to be merged for the google/trillian project. We use dependabot to ensure our dependencies are kept up to date. Until a tagged version of etcd is released that uses a newer version of grpc, many of our dependency updates are blocked. We're not looking for any of the newer features in grpc that this backport may introduce.

Trillian's dependency graph will continue to become increasingly out of date until this PR is merged or the 3.6 version of etcd is released. This is a cost for us as maintainers to manage dependabot (which is normal running is a very automatic process), and more importantly, it means that our project and any projects that depend on it (e.g. Certificate Transparency, Sigstore) will be stuck on old/obsolete/vulnerable versions of a growing number of libraries.

ahrtr:
@rsafonseca @mhutchinson Please feel free to bring the topic to tonight's etcd community meeting [I will not be able to join the meeting :(]. 

You need to subscribe etcd-dev@googlegroups.com (https://groups.google.com/g/etcd-dev) to gain permission to add topic to the google doc,

https://docs.google.com/document/d/16XEGyPBisZvmmoIHSZzv__LoyOeluC5a4x353CX0SIM/edit#heading=h.eu3cetgrd3ii

jmhbnz (approved):
On balance, I support this pull request merging, thanks @chaochn47 for raising.

As mentioned in our [dependency management guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#stable-branches) we generally don't bump dependencies for stable releases unless there is a security driver.

With that said, given how far away `3.6.0` is, and the input from the community on how not proceeding with the backport will negatively affect projects that depend on etcd stable releases (and how this will get worse over time).  Then on the understanding that this backport is non breaking / transparent to clients my vote would be that we proceed.

ahrtr:
Part of https://github.com/etcd-io/etcd/issues/16740

ahrtr:
@chaochn47 please rebase this PR.

ahrtr:
@chaochn47 can you please also bump grpc for 3.4?

chaochn47:
Security is the top priority, I will work on this today. Thanks for the heads up! @ahrtr

ahrtr:
> Just a reminder that this PR alone won't solve the dependency issues, it's just the first bump in a series that needs to happen to get to a version with the problem fixed (1.55+)

We are well aware of it, we will bump grpc-go step by step. FYI. https://github.com/etcd-io/etcd/issues/16740

api/go.mod
```diff
@@ -8,14 +8,14 @@ require (
...
-	google.golang.org/grpc v1.41.0
+	google.golang.org/grpc v1.47.0
 )
 
 require (
-	golang.org/x/net v0.17.0 // indirect
-	golang.org/x/sys v0.13.0 // indirect
-	golang.org/x/text v0.13.0 // indirect
-	google.golang.org/protobuf v1.26.0 // indirect
+	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4 // indirect
```
ahrtr, line 15: This isn't correct. You reverted the `golang.org/x/net` version.

chaochn47: It's a result of `go mod tidy` after `git rebase` that reverted this std lib version. 

```
dev-dsk-chaochn-2c-a26acd76 % go mod graph | grep "golang.org/x/tools@v0.1.2"
google.golang.org/genproto@v0.0.0-20210602131652-f16073e35f0c golang.org/x/tools@v0.1.2

dev-dsk-chaochn-2c-a26acd76 % go mod graph | grep "golang.org/x/net"
golang.org/x/tools@v0.1.2 golang.org/x/net@v0.0.0-20210405180319-a5a99cb37ef4
```

In conclusion, it was reverted because of `google.golang.org/genproto` was stick with old version.

chaochn47: I can manually bump up indirect dependency though.

server/embed/etcd.go
chaochn47: > Connecting to unix:localhost:port instead of never officially supported unix://localhost:port (https://grpc.github.io/grpc/cpp/md_doc_naming.html)

ahrtr:
Looks much better now. Should be ready to go. ping @serathius

ahrtr (approved):
LGTM

Thanks @chaochn47 for the quick response.

chaochn47:
> @chaochn47 can you please also bump grpc for 3.4?

I am on it. 3.4 `google.golang.org/grpc` is still on `v1.26.0` :p

ahrtr:
ping @wenjiaswe @serathius @fuweid @jmhbnz

fuweid (approved):
LGTM on new change

wenjiaswe:
Agree the security issue is super important for all, us included. But also want to be extra cautious given this is a stable version. I took a quick look at the diff between grpc-go, there are quite a bit new features, API changes, behavior changes and stuff. 

@chaochn47 would you please kindly check the diff between grpc-go 1.41 and 1.47, and make sure there won't be any breaking change? Let's just try to be extra careful, so many people are waiting for this patch, let's not break anyone.

chaochn47:
> @chaochn47 would you please kindly check the diff between grpc-go 1.41 and 1.47, and make sure there won't be any breaking changes? Let's just try to be extra careful, so many people are waiting for this patch, let's not break anyone.

Sure. 

After reviewing change log of [v1.42.0](https://github.com/grpc/grpc-go/releases/tag/v1.42.0) and https://github.com/grpc/grpc-go/compare/v1.41.0...v1.42.0, there are potentially three behavior changes. 
1. grpc: Dial("unix://relative-path") no longer works (https://github.com/grpc/grpc-go/pull/4817). Use "unix://absolute-path" or "unix:relative-path" instead in accordance with [our documentation](https://github.com/grpc/grpc/blob/master/doc/naming.md#name-syntax)
2. transport: server transport will reject requests with connection header (https://github.com/grpc/grpc-go/pull/4803). gRPC server transport will reject with END_STREAM when user uses HTTP1.1 request with connection header set dialing gRPC-gateway. This is not prohibited by https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.2 though. 
3. grpc: support grpc.WithAuthority when secure credentials are used (https://github.com/grpc/grpc-go/pull/4817) There is a new feature. Existing user of etcd 3.5 does not depend on this. 

In conclusion: we should be good upgrading to `v1.42.0`. I will continue with `v1.43.0` auditing..

server/embed/etcd.go
chaochn47: 3.5 server does not allow client dialing with UDS. See https://github.com/etcd-io/etcd/pull/15940

chaochn47:
Behavior Changes
- [ ] [gRPC v1.43.0 change log](https://github.com/grpc/grpc-go/releases/tag/v1.43.0)
    - [ ] status: support wrapped errors in FromContextError (https://github.com/grpc/grpc-go/pull/4977)
        - [ ] etcd does not uses this exported function. 
    - [ ] config: remove the environment variable to disable retry support (https://github.com/grpc/grpc-go/pull/4922)
        - [ ] etcd does not use gRPC retry option in the service config. We implement by the gRPC interceptor instead. So this is no-op for us. See more in #16216
- [ ] [gRPC v1.44.0 change log](https://github.com/grpc/grpc-go/releases/tag/v1.44.0)
    - [ ]  no behavior changes. But it reveals that we should replace all `grpc.WithInsecure()` with `grpc.WithTransportCredentials(insecure.NewCredentials())` based on https://github.com/grpc/grpc-go/pull/5068 in 3.4/3.5/main
- [ ] [gRPC v1.45.0 change log](https://github.com/grpc/grpc-go/releases/tag/v1.45.0)
    - [ ] server: convert context errors returned by service handlers to status with the correct status code (Canceled or DeadlineExceeded), instead of Unknown (https://github.com/grpc/grpc-go/pull/5156)
        - [ ] Somehow etcd server already relies on this behavior in [togRPCError](https://github.com/etcd-io/etcd/blob/release-3.5/server/etcdserver/api/v3rpc/util.go#L96-L99) "let gRPC server convert to codes.Canceled, codes.DeadlineExceeded"
- [ ] [gRPC v1.46.0 change log](https://github.com/grpc/grpc-go/releases/tag/v1.46.0)
    - [ ] server: Support setting TCP_USER_TIMEOUT on grpc.Server connections using keepalive.ServerParameters.Time (https://github.com/grpc/grpc-go/pull/5219) 
        - [ ] This will effectively change the default TCP user timeout as 20s in gRPC server transport layer. This is the same as the h2 keep alive ping timeout default value though. 

> I am currently of the opinion that if an application decides to change the timeout value from the default of 20 seconds, it is doing so knowingly and owns the responsibility of connections being dropped because of that.

If a server administrator configures a relatively smaller value of the `grpc-keepalive-timeout` depending on h2 ping is lightweight, but potentially a heavy range request takes longer than the configured TCP user timeout. 

Should we put a lower bound limit on this configuration? @fuweid 

- [ ] [gRPC v1.47.0 change log](https://github.com/grpc/grpc-go/releases/tag/v1.47.0)

chaochn47:
Done all the auditing.. No outstanding breaking changes to etcd AFAIK.

ahrtr:
> Done all the auditing.. No outstanding breaking changes to etcd AFAIK.

thx @chaochn47 for the investigation. It aligns with my understanding. There is no obvious differences between 3.5 and 3.6 in term of grpc usage. If a grpc version works for 3.6, there is no reason that it doesn't work for 3.5.

But 3.4 is another story.  see https://github.com/etcd-io/etcd/issues/16740#issuecomment-1760937656. We can have more discussion in your following PR of bumping grpc for 3.4.

wenjiaswe:
Thanks @chaochn47 !

> 1. grpc: Dial("unix://relative-path") no longer works (https://github.com/grpc/grpc-go/pull/4817). Use "unix://absolute-path" or "unix:relative-path" instead in accordance with [our documentation](https://github.com/grpc/grpc/blob/master/doc/naming.md#name-syntax). 3.5 etcd binary does not allow UDS in the listen client urls flag since https://github.com/etcd-io/etcd/pull/15940 has not yet merged.

Do we need to change https://github.com/etcd-io/etcd/pull/15940 accordingly? Have we changed https://github.com/etcd-io/etcd/pull/12469 since main branch already updated grpc-go? If so, could you please open an issue for tracking, to the next 3.5 cut is blocked by it?

>  2. transport: server transport will reject requests with connection header (https://github.com/grpc/grpc-go/pull/4803).
 gRPC server transport will reject with END_STREAM when user uses HTTP1.1 request with connection header set dialing gRPC-gateway. This behavior is aligned with https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.2 that prohibits such request with connection header. Existing user depends on the previous wrong behavior should change its usage.

This looks like a breaking change? @chaochn47 could you please make sure this is reflected in the CHANGELOG?

> [gRPC v1.44.0 change log](https://github.com/grpc/grpc-go/releases/tag/v1.44.0), no behavior changes. But it reveals that we should replace all grpc.WithInsecure() with grpc.WithTransportCredentials(insecure.NewCredentials()) based on https://github.com/grpc/grpc-go/pull/5068 in 3.4/3.5/main

Could you open a ticket as 3.5 release blocking issue for tracking?

cc @jmhbnz @serathius @ahrtr @fuweid could you please help to check if there is anything missing?

ahrtr:
> Do we need to change #15940 accordingly? Have we changed #12469 since main branch already updated grpc-go? If so, could you please open an issue for tracking, to the next 3.5 cut is blocked by it?

- I don't think 3.5 patch release is blocked by it;
- I don't see strong reason to reject the minor & safe feature for 3.5. Of course, since it's a feature (usually we don't backport feature to stable release), so I won't insist on it if others do not agree.

chaochn47:
@wenjiaswe 

> Do we need to change https://github.com/etcd-io/etcd/pull/15940 accordingly? Have we changed https://github.com/etcd-io/etcd/pull/12469 since main branch already updated grpc-go? If so, could you please open an issue for tracking, to the next 3.5 cut is blocked by it?

No, etcd 3.5 never supports unix socket domain in the `listen-client-urls`. So I think it's not a blocker of this PR to resolve CVE. Does that make sense to you? 

I just moved the discussion to the right place https://github.com/etcd-io/etcd/pull/15940#issuecomment-1763491517. 

> This looks like a breaking change? @chaochn47 could you please make sure this is reflected in the CHANGELOG?

https://github.com/etcd-io/etcd/pull/16762 is drafted since `CHANGELOG-3.5.md` is in main branch. 

> Could you open a ticket as 3.5 release blocking issue for tracking?

Deprecated does not mean removed. 3.5 release is not blocked by this issue. Tracked in https://github.com/etcd-io/etcd/issues/15145#issuecomment-1763492117 though.

chaochn47:
Hi @@wenjiaswe, @serathius, is it in a good shape to be merged and unblock the CVE fix?

wenjiaswe:
Looks like everyone approved. I will go ahead and merge.

PR #8901, opened 2017-11-21
auth, etcdserver: follow the correct usage of context

mitake:
The keys of context shouldn't be string. They should be a struct of
their own type.

Fix https://github.com/coreos/etcd/issues/8826

/cc @gyuho

PR #16626, opened 2023-09-21
*: fix staticcheck lint

fuweid:
Changed TraceKey/StartTimeKey/TokenFieldNameGRPCKey to struct{} to follow the correct usage of context. Similar patch to #8901.

~~And this patch also copies the tools/.golangci.yaml and run the linters for which we have already fixed. 
The temp .golangci.yaml will be removed when we fixes all the linters' issues.~~

Added `lint_pass` in scripts/test.sh to run the golangci linters for each modules. And disable no working linters in the global golangci.yaml and enable them by #16610.

Part of #16610

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ verify-dep:
 # still depends on legacy {ineffassign,nakedret,unparam,...}_pass. These X_pass
 # will be removed when the golangci-lint covers all the sub modules.
 .PHONY: verify-lint
-verify-lint: verify-ineffassign
+verify-lint: verify-golangcilint
 	golangci-lint run --config tools/.golangci.yaml
 
 .PHONY: fix-lint
@@ -150,9 +150,9 @@ verify-yamllint:
 verify-govet-shadow:
 	PASSES="govet_shadow" ./scripts/test.sh
 
-.PHONY: verify-ineffassign
-verify-ineffassign:
-	PASSES="ineffassign" ./scripts/test.sh
+.PHONY: verify-golangcilint
+verify-golangcilint:
+	PASSES="golangcilint" ./scripts/test.sh
 
 YAMLFMT_VERSION = $(shell cd tools/mod && go list -m -f '{{.Version}}' github.com/google/yamlfmt)
 
--- a/Makefile
+++ b/Makefile
@@ -85,14 +85,9 @@ fix-bom:
 verify-dep:
 	PASSES="dep" ./scripts/test.sh
 
-# TODO: https://github.com/etcd-io/etcd/issues/16610
-#
-# The golangci-lint doesn't verify sub modules. Before #16610 fixed, verify-lint
-# still depends on legacy {ineffassign,nakedret,unparam,...}_pass. These X_pass
-# will be removed when the golangci-lint covers all the sub modules.
 .PHONY: verify-lint
-verify-lint: verify-golangcilint
-	golangci-lint run --config tools/.golangci.yaml
+verify-lint:
+	PASSES="lint" ./scripts/test.sh
 
 .PHONY: fix-lint
 fix-lint:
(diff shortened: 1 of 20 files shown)


scripts/test.sh
```diff
@@ -425,11 +425,46 @@ function unparam_pass {
...
-  # versionpb/version.pb.go:69:15: proto.RegisterFile is deprecated: Use protoregistry.GlobalFiles.RegisterFile instead.  (SA1019)
-  run_for_modules generic_checker run_go_tool "honnef.co/go/tools/cmd/staticcheck"
+function golangcilint_pass {
+  local tmp_lintyaml
+  tmp_lintyaml=$(mktemp -t 'tmp_golangcilint_XXX.yaml')
+
+  # shellcheck disable=SC2064
+  trap "rm ${tmp_lintyaml}; trap - RETURN" RETURN
+
+  # TODO: Copy the tools/.golangci.yaml and run the linters for which we have
```
serathius, line 435: We should only have 1 source of lint configuration. Adding second one like this is bad for visibility.

Please just disable no working linters in the main golangci.yaml and add a TODO to reenable them.

Makefile
```diff
@@ -150,9 +150,9 @@ verify-yamllint:
 verify-govet-shadow:
 	PASSES="govet_shadow" ./scripts/test.sh
 
-.PHONY: verify-ineffassign
-verify-ineffassign:
-	PASSES="ineffassign" ./scripts/test.sh
+.PHONY: verify-golangcilint
```
serathius, line 153: Let's keep one target for linting.

scripts/test.sh
fuweid: Makes senses. Updated.

Makefile
fuweid: Updated.

Makefile
```diff
@@ -85,14 +85,9 @@ fix-bom:
...
-# TODO: https://github.com/etcd-io/etcd/issues/16610
-#
-# The golangci-lint doesn't verify sub modules. Before #16610 fixed, verify-lint
-# still depends on legacy {ineffassign,nakedret,unparam,...}_pass. These X_pass
-# will be removed when the golangci-lint covers all the sub modules.
 .PHONY: verify-lint
-verify-lint: verify-ineffassign
-	golangci-lint run --config tools/.golangci.yaml
+verify-lint:
+	PASSES="golangcilint" ./scripts/test.sh
```
serathius, line 90: Could we rename the pass to just `lint`? It would make it consistent with `verify-lint`.

serathius (approved):

tests/robustness/report/client.go
```diff
@@ -169,7 +169,6 @@ func loadKeyValueOperations(path string) (operations []porcupine.Operation, err
 func persistWatchOperations(t *testing.T, lg *zap.Logger, path string, responses []model.WatchOperation) {
 	lg.Info("Saving watch operations", zap.String("path", path))
 	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
-	defer file.Close()
```
ahrtr, line 172: Good catch!

client/v3/experimental/recipes/double_barrier.go
```diff
@@ -71,10 +71,9 @@ func (b *DoubleBarrier) Enter() error {
 			// delete itself now, otherwise other processes may need to wait
 			// until these keys are automatically deleted when the related
 			// lease expires.
-			if err = b.myKey.Delete(); err != nil {
-				// Nothing to do here. We have to wait for the key to be
-				// deleted when the lease expires.
-			}
+			b.myKey.Delete()
+			// Nothing to do here even if we run into error.
+			// We have to wait for the key to be deleted when the lease expires.
```
ahrtr, line 76: Suggest to keep this unchanged. Previously code explicitly commenting why the error is ignored. 

It can't even pass the static check, because all returned errors should be explicitly ignored (i.e. _ = xxxx) or be checked. The reason why all the workflows are green is probably because some checks/lints are disabled?

fuweid:
ping @ahrtr ~

client/v3/experimental/recipes/double_barrier.go
fuweid: > It can't even pass the static check, because all returned errors should be explicitly ignored (i.e. _ = xxxx) or be checked. The reason why all the workflows are green is probably because some checks/lints are disabled?

we don't enable `errcheck` before. So it can pass the check.

> Suggest to keep this unchanged. Previous code explicitly commenting why the error is ignored.
Sounds good. I was thinking that it's good enough to keep the comment. I updated with disable linter comment. Please take a look.

ahrtr (approved):
LGTM

nice work! thx @fuweid

PR #16651, opened 2023-09-25
Add livez and readyz for etcd

siyuanfoundation:
https://github.com/etcd-io/etcd/issues/16007

This is a prototype for adding livez/readyz support to etcd ([design doc](https://docs.google.com/document/d/1SkzmO4RT_GI9YhT0dw4a6nEwKVCciwrwbCDxK0D7ASM/edit?usp=sharing)). 
This pr is setting up the general structure for livez/ready checks, with only 2 simple checks implemented.

### DIFF
--- a/server/etcdserver/api/etcdhttp/health.go
+++ b/server/etcdserver/api/etcdhttp/health.go
@@ -15,10 +15,13 @@
 package etcdhttp
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"path"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
@@ -46,7 +49,7 @@ type ServerHealth interface {
 // HandleHealth registers metrics and health handlers. it checks health by using v3 range request
 // and its corresponding timeout.
 func HandleHealth(lg *zap.Logger, mux *http.ServeMux, srv ServerHealth) {
-	mux.Handle(PathHealth, NewHealthHandler(lg, func(ctx context.Context, excludedAlarms AlarmSet, serializable bool) Health {
+	mux.Handle(PathHealth, NewHealthHandler(lg, func(ctx context.Context, excludedAlarms StringSet, serializable bool) Health {
 		if h := checkAlarms(lg, srv, excludedAlarms); h.Health != "true" {
 			return h
 		}
@@ -55,10 +58,13 @@ func HandleHealth(lg *zap.Logger, mux *http.ServeMux, srv ServerHealth) {
 		}
 		return checkAPI(ctx, lg, srv, serializable)
 	}))
+
+	installLivezEndpoints(lg, mux, srv)
+	installReadyzEndpoints(lg, mux, srv)
 }
 
 // NewHealthHandler handles '/health' requests.
-func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAlarms AlarmSet, Serializable bool) Health) http.HandlerFunc {
+func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAlarms StringSet, Serializable bool) Health) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodGet {
 			w.Header().Set("Allow", http.MethodGet)
@@ -66,7 +72,7 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
 			lg.Warn("/health error", zap.Int("status-code", http.StatusMethodNotAllowed))
 			return
 		}
-		excludedAlarms := getExcludedAlarms(r)
+		excludedAlarms := getQuerySet(r, "exclude")
 		// Passing the query parameter "serializable=true" ensures that the
 		// health of the local etcd is checked vs the health of the cluster.
 		// This is useful for probes attempting to validate the liveness of
@@ -119,20 +125,18 @@ type Health struct {
 	Reason string `json:"reason"`
 }
 
-type AlarmSet map[string]struct{}
-
-func getExcludedAlarms(r *http.Request) (alarms AlarmSet) {
-	alarms = make(map[string]struct{}, 2)
-	alms, found := r.URL.Query()["exclude"]
+func getQuerySet(r *http.Request, query string) StringSet {
+	querySet := make(map[string]struct{})
+	qs, found := r.URL.Query()[query]
 	if found {
-		for _, alm := range alms {
-			if len(alm) == 0 {
+		for _, q := range qs {
+			if len(q) == 0 {
 				continue
 			}
-			alarms[alm] = struct{}{}
+			querySet[q] = struct{}{}
 		}
 	}
-	return alarms
+	return querySet
 }
 
 func getSerializableFlag(r *http.Request) bool {
@@ -141,7 +145,7 @@ func getSerializableFlag(r *http.Request) bool {
 
 // TODO: etcdserver.ErrNoLeader in health API
 
-func checkAlarms(lg *zap.Logger, srv ServerHealth, excludedAlarms AlarmSet) Health {
+func checkAlarms(lg *zap.Logger, srv ServerHealth, excludedAlarms StringSet) Health {
 	h := Health{Health: "true"}
 
 	for _, v := range srv.Alarms() {
@@ -192,4 +196,172 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
 	}
 	lg.Debug("serving /health true")
 	return h
+}
+
+type HealthCheck func(ctx context.Context) error
+
+type CheckRegistry struct {
+	path   string
+	checks map[string]HealthCheck
+}
+
+func installLivezEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
+	reg := CheckRegistry{path: "/livez", checks: make(map[string]HealthCheck)}
+	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.InstallHttpEndpoints(lg, mux)
+}
+
+func installReadyzEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
+	reg := CheckRegistry{path: "/readyz", checks: make(map[string]HealthCheck)}
+	reg.Register("data_corruption", activeAlarmCheck(server, pb.AlarmType_CORRUPT))
+	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.InstallHttpEndpoints(lg, mux)
+}
+
+func (reg *CheckRegistry) Register(name string, check HealthCheck) {
+	reg.checks[name] = check
+}
+
+func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
+	checkNames := make([]string, 0, len(reg.checks))
+	for k := range reg.checks {
+		checkNames = append(checkNames, k)
+	}
+
+	// installs the http handler for the root path.
+	reg.installRootHttpEndpoint(lg, mux, reg.path, checkNames...)
+	for _, checkName := range checkNames {
+		// installs the http handler for the individual check sub path.
+		subpath := path.Join(reg.path, checkName)
+		check := checkName
+		mux.Handle(subpath, newHealthHandler(subpath, lg, func(r *http.Request) Health {
+			return reg.runHealthChecks(r.Context(), check)
+		}))
+	}
+}
+
+func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...string) Health {
+	h := Health{Health: "true"}
+	var individualCheckOutput bytes.Buffer
+	for _, checkName := range checkNames {
+		check, found := reg.checks[checkName]
+		if !found {
+			panic(fmt.Errorf("Health check: %s not registered", checkName))
+		}
+		if err := check(ctx); err != nil {
+			fmt.Fprintf(&individualCheckOutput, "[-]%s failed: %v\n", checkName, err)
+			h.Health = "false"
+		} else {
+			fmt.Fprintf(&individualCheckOutput, "[+]%s ok\n", checkName)
+		}
+	}
+	h.Reason = individualCheckOutput.String()
+	return h
+}
+
+// installRootHttpEndpoint installs the http handler for the root path.
+func (reg *CheckRegistry) installRootHttpEndpoint(lg *zap.Logger, mux *http.ServeMux, path string, checks ...string) {
+	hfunc := func(r *http.Request) Health {
+		// extracts the health check names to be excludeList from the query param
+		excluded := getQuerySet(r, "exclude")
+
+		filteredCheckNames := filterCheckList(lg, listToStringSet(checks), excluded)
+		return reg.runHealthChecks(r.Context(), filteredCheckNames...)
+	}
+	mux.Handle(path, newHealthHandler(path, lg, hfunc))
+}
+
+// newHealthHandler generates a http HandlerFunc for a health check function hfunc.
+func newHealthHandler(path string, lg *zap.Logger, hfunc func(*http.Request) Health) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			w.Header().Set("Allow", http.MethodGet)
+			http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+			lg.Warn("Health request error", zap.String("path", path), zap.Int("status-code", http.StatusMethodNotAllowed))
+			return
+		}
+		h := hfunc(r)
+		// Always returns detailed reason for failed checks.
+		if h.Health != "true" {
+			http.Error(w, h.Reason, http.StatusServiceUnavailable)
+			lg.Error("Health check error", zap.String("path", path), zap.String("reason", h.Reason), zap.Int("status-code", http.StatusServiceUnavailable))
+			return
+		}
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		// Only writes detailed reason for verbose requests.
+		if _, found := r.URL.Query()["verbose"]; found {
+			fmt.Fprint(w, h.Reason)
+		}
+		fmt.Fprint(w, "ok\n")
+		lg.Debug("Health check OK", zap.String("path", path), zap.String("reason", h.Reason), zap.Int("status-code", http.StatusOK))
+	}
+}
+
+func filterCheckList(lg *zap.Logger, checks StringSet, excluded StringSet) []string {
+	filteredList := []string{}
+	for chk := range checks {
+		if _, found := excluded[chk]; found {
+			delete(excluded, chk)
+			continue
+		}
+		filteredList = append(filteredList, chk)
+	}
+	if len(excluded) > 0 {
+		// For version compatibility, excluding non-exist checks would not fail the request.
+		lg.Warn("some health checks cannot be excluded", zap.String("missing-health-checks", formatQuoted(excluded.List()...)))
+	}
+	return filteredList
+}
+
+// formatQuoted returns a formatted string of the health check names,
+// preserving the order passed in.
+func formatQuoted(names ...string) string {
+	quoted := make([]string, 0, len(names))
+	for _, name := range names {
+		quoted = append(quoted, fmt.Sprintf("%q", name))
+	}
+	return strings.Join(quoted, ",")
+}
+
+type StringSet map[string]struct{}
+
+func (s StringSet) List() []string {
+	keys := make([]string, 0, len(s))
+	for k := range s {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
+func listToStringSet(list []string) StringSet {
+	set := make(map[string]struct{})
+	for _, s := range list {
+		set[s] = struct{}{}
+	}
+	return set
+}
+
+// activeAlarmCheck checks if a specific alarm type is active in the server.
+func activeAlarmCheck(srv ServerHealth, at pb.AlarmType) func(context.Context) error {
+	return func(ctx context.Context) error {
+		as := srv.Alarms()
+		for _, v := range as {
+			if v.Alarm == at {
+				return fmt.Errorf("alarm activated: %s", at.String())
+			}
+		}
+		return nil
+	}
+}
+
+func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		ctx = srv.AuthStore().WithRoot(ctx)
+		_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: true})
+		if err != nil {
+			return fmt.Errorf("range error: %w", err)
+		}
+		return nil
+	}
 }
--- a/server/etcdserver/api/etcdhttp/health_test.go
+++ b/server/etcdserver/api/etcdhttp/health_test.go
@@ -16,11 +16,11 @@ package etcdhttp
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"go.uber.org/zap/zaptest"
@@ -32,16 +32,15 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/auth"
 	"go.etcd.io/etcd/server/v3/config"
-	"go.etcd.io/etcd/server/v3/etcdserver"
 	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
 	"go.etcd.io/etcd/server/v3/storage/schema"
 )
 
 type fakeHealthServer struct {
 	fakeServer
-	health    string
-	apiError  error
-	authStore auth.AuthStore
+	apiError      error
+	missingLeader bool
+	authStore     auth.AuthStore
 }
 
 func (s *fakeHealthServer) Range(_ context.Context, _ *pb.RangeRequest) (*pb.RangeResponse, error) {
@@ -53,87 +52,91 @@ func (s *fakeHealthServer) Config() config.ServerConfig {
 }
 
 func (s *fakeHealthServer) Leader() types.ID {
-	if s.health == "true" {
+	if !s.missingLeader {
 		return 1
 	}
 	return types.ID(raft.None)
 }
-func (s *fakeHealthServer) Do(_ context.Context, _ pb.Request) (etcdserver.Response, error) {
-	if s.health == "true" {
-		return etcdserver.Response{}, nil
-	}
-	return etcdserver.Response{}, fmt.Errorf("fail health check")
-}
-func (s *fakeHealthServer) AuthStore() auth.AuthStore   { return s.authStore }
+
+func (s *fakeHealthServer) AuthStore() auth.AuthStore { return s.authStore }
+
 func (s *fakeHealthServer) ClientCertAuthEnabled() bool { return false }
+
+type healthTestCase struct {
+	name             string
+	healthCheckURL   string
+	expectStatusCode int
+	inResult         []string
+	notInResult      []string
+
+	alarms        []*pb.AlarmMember
+	apiError      error
+	missingLeader bool
+}
 
 func TestHealthHandler(t *testing.T) {
 	// define the input and expected output
 	// input: alarms, and healthCheckURL
-	tests := []struct {
-		name           string
-		alarms         []*pb.AlarmMember
-		healthCheckURL string
-		apiError       error
-
-		expectStatusCode int
-		expectHealth     string
-	}{
+	tests := []healthTestCase{
 		{
 			name:             "Healthy if no alarm",
 			alarms:           []*pb.AlarmMember{},
 			healthCheckURL:   "/health",
 			expectStatusCode: http.StatusOK,
-			expectHealth:     "true",
 		},
 		{
 			name:             "Unhealthy if NOSPACE alarm is on",
 			alarms:           []*pb.AlarmMember{{MemberID: uint64(0), Alarm: pb.AlarmType_NOSPACE}},
 			healthCheckURL:   "/health",
 			expectStatusCode: http.StatusServiceUnavailable,
-			expectHealth:     "false",
 		},
 		{
 			name:             "Healthy if NOSPACE alarm is on and excluded",
 			alarms:           []*pb.AlarmMember{{MemberID: uint64(0), Alarm: pb.AlarmType_NOSPACE}},
 			healthCheckURL:   "/health?exclude=NOSPACE",
 			expectStatusCode: http.StatusOK,
-			expectHealth:     "true",
 		},
 		{
 			name:             "Healthy if NOSPACE alarm is excluded",
 			alarms:           []*pb.AlarmMember{},
 			healthCheckURL:   "/health?exclude=NOSPACE",
 			expectStatusCode: http.StatusOK,
-			expectHealth:     "true",
 		},
 		{
 			name:             "Healthy if multiple NOSPACE alarms are on and excluded",
 			alarms:           []*pb.AlarmMember{{MemberID: uint64(1), Alarm: pb.AlarmType_NOSPACE}, {MemberID: uint64(2), Alarm: pb.AlarmType_NOSPACE}, {MemberID: uint64(3), Alarm: pb.AlarmType_NOSPACE}},
 			healthCheckURL:   "/health?exclude=NOSPACE",
 			expectStatusCode: http.StatusOK,
-			expectHealth:     "true",
 		},
 		{
 			name:             "Unhealthy if NOSPACE alarms is excluded and CORRUPT is on",
 			alarms:           []*pb.AlarmMember{{MemberID: uint64(0), Alarm: pb.AlarmType_NOSPACE}, {MemberID: uint64(1), Alarm: pb.AlarmType_CORRUPT}},
 			healthCheckURL:   "/health?exclude=NOSPACE",
 			expectStatusCode: http.StatusServiceUnavailable,
-			expectHealth:     "false",
 		},
 		{
 			name:             "Unhealthy if both NOSPACE and CORRUPT are on and excluded",
 			alarms:           []*pb.AlarmMember{{MemberID: uint64(0), Alarm: pb.AlarmType_NOSPACE}, {MemberID: uint64(1), Alarm: pb.AlarmType_CORRUPT}},
 			healthCheckURL:   "/health?exclude=NOSPACE&exclude=CORRUPT",
 			expectStatusCode: http.StatusOK,
-			expectHealth:     "true",
 		},
 		{
 			name:             "Unhealthy if api is not available",
 			healthCheckURL:   "/health",
 			apiError:         fmt.Errorf("Unexpected error"),
 			expectStatusCode: http.StatusServiceUnavailable,
-			expectHealth:     "false",
+		},
+		{
+			name:             "Unhealthy if no leader",
+			healthCheckURL:   "/health",
+			expectStatusCode: http.StatusServiceUnavailable,
+			missingLeader:    true,
+		},
+		{
+			name:             "Healthy if no leader and serializable=true",
+			healthCheckURL:   "/health?serializable=true",
+			expectStatusCode: http.StatusOK,
+			missingLeader:    true,
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


siyuanfoundation:
Tested with local server
`curl '127.0.0.1:2379/readyz?verbose'`
`curl '127.0.0.1:2379/livez?verbose'`

`bin/etcdctl endpoint readyz --cluster`
`bin/etcdctl endpoint livez --cluster`

tjungblu:
Cool stuff and thanks for the PR, I was wondering why you're adding this to the GRPC API? We already have a health handler on the http server: https://github.com/etcd-io/etcd/blob/main/server/embed/etcd.go#L746

Just FYI, we had huge issues in the past with running etcdctl commands as a health probes in openshift with regards to zombie processes. We're much happier with just letting kubelet hit an http endpoint instead.

server/etcdserver/healthz_test.go
```diff
@@ -0,0 +1,326 @@
+// Copyright 2022 The etcd Authors
```
chaochn47, line 1: Need to change to the right calendar year 2023. 

Could you run `make fix` and `make verify` and install any lint tools if it is missing in your dev box?

server/etcdserver/healthz.go
```diff
@@ -0,0 +1,397 @@
...
+	s.healthHandler.installReadyz(lg, mux)
+	s.healthHandler.installHealthz(lg, mux)
+}
+
+func (s *EtcdServer) AddHealthCheck(check HealthChecker, isLivez bool, isReadyz bool) error {
+	return s.healthHandler.addHealthCheck(check, isLivez, isReadyz)
+}
+
+type HealthHandler struct {
+	server EtcdServerHealth
```
chaochn47: I feel like HealthHandler should be in the api layer and move to `server/etcdserver/api/`. If it is only used for etcdhttp and move it there.

chaochn47:
> We already have a health handler on the http server: https://github.com/etcd-io/etcd/blob/main/server/embed/etcd.go#L746

I think we can start from http endpoint. I don’t think backporting gRPC changes to release-3.5 is feasible and accepted. It will literally change the customer facing API.

server/etcdserver/healthz_test.go
siyuanfoundation: `make fix` does not detect the 2023 errors. 
`make verify` is broken due to the protogen error `sed: 1: "s#import \(#import \("g ...": RE error: parentheses not balanced`, which also broken for the main branch. I don't know where the error is.

server/etcdserver/healthz.go
siyuanfoundation: Even though the api change is not desirable in 3.5, are you sure we don't need to add it to gPRC in the future? Existing health check has different/repeated logic at different places, which is quite confusing.

siyuanfoundation:
> > We already have a health handler on the http server: https://github.com/etcd-io/etcd/blob/main/server/embed/etcd.go#L746
> 
> I think we can start from http endpoint. I don’t think backporting gRPC changes to release-3.5 is feasible and accepted. It will literally change the customer facing API even if it is just maintenance service.

Removed gPRC change.

serathius:
Are white spaces missing only for me or someone has the same problem?
image

server/etcdserver/healthz.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// HealthChecker is a named healthz checker.
+type HealthChecker interface {
+	Name() string
+	Check() error
+}
+
+type healthLogger interface {
+	Infof(template string, args ...interface{})
```
serathius, line 50: Why?

server/etcdserver/healthz.go
```diff
@@ -0,0 +1,397 @@
...
+)
+
+type EtcdServerHealth interface {
+	Alarms() []*etcdserverpb.AlarmMember
+	Range(context.Context, *etcdserverpb.RangeRequest) (*etcdserverpb.RangeResponse, error)
+}
+
+// HealthChecker is a named healthz checker.
+type HealthChecker interface {
+	Name() string
```
serathius, line 45: I would not put `Name()` as part of health check interface, but rather provide the name during registration.

server/embed/etcd.go
```diff
@@ -744,6 +744,7 @@ func (e *Etcd) serveClients() (err error) {
 	etcdhttp.HandleVersion(mux, e.Server)
 	etcdhttp.HandleMetrics(mux)
 	etcdhttp.HandleHealth(e.cfg.logger, mux, e.Server)
```
serathius, line 746: what about the old healthz code?

siyuanfoundation: I am keeping the old health endpoint, because some checks are not going to be included in the new healthz, such as all alarms, and it uses a different set of query params like "serializable=true".

server/etcdserver/healthz.go
siyuanfoundation: moved to server/etcdserver/api/v3health

wenjiaswe:
cc @ahrtr @mitake @jmhbnz @ptabor @lavacat @fuweid

server/embed/etcd.go
serathius: Oh, I would find having old "/health" and "/healthz" confusing. Can we skip "/healtz" in this PR?

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+type HealthChecker interface {
+	Check() error
+}
+
+type HealthStatus struct {
+	IsOk   bool   `json:"isOk"`
+	Reason string `json:"reason"`
+}
+
+type HealthHandler struct {
```
serathius, line 57: Instead of having one structure to handle all the checks with mirrored code for each type, why not have a generic struct that you can reuse?
```
func InstallHealthEndpoints(mux, server) {
  mux.Handle("/livez", LiveHandler(server))
  mux.Handle("/readyz", ReadyHandler(server))
}

type HealthCheck func() error

type CheckRegistry struct {
  checks map[string]HealthCheck
}

func  LivezHandler() http.Handler {
   reg := CheckRegistry{}
   reg.Register("live", checkAlarms)
   retrurn reg.Handler()
}

func  ReadyzHandler() http.Handler {
   reg := CheckRegistry{}
   reg.Register("alarms", checkAlarms)
   retrurn reg.Handler()
}


func (r *CheckRegister) Register(name string, check HealthCheck) {
  r.checks[name] = check
}

func (r *CheckRegister) Handler() http.Handler {
  return func(_, _) {
  
  }
}
```

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+	server EtcdServerHealth
+	// lock for health check related functions.
+	healthMux sync.Mutex
```
serathius, line 60: We don't need to handle concurrency here.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+
+func NewHealthHandler(s EtcdServerHealth) (handler *HealthHandler, err error) {
+	handler = &HealthHandler{
+		server:           s,
+		healthCheckStore: make(map[string]HealthChecker),
+		healthzChecks:    []string{},
+		livezChecks:      []string{},
+		readyzChecks:     []string{},
+	}
+	if err = handler.addDefaultHealthChecks(); err != nil {
```
serathius, line 83: I would suggest for installation of health checks to be an external function and not part of constructor.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+	Alarms() []*etcdserverpb.AlarmMember
+	Range(context.Context, *etcdserverpb.RangeRequest) (*etcdserverpb.RangeResponse, error)
+}
+
+// mux is an interface describing the methods InstallHandler requires.
+type mux interface {
+	Handle(pattern string, handler http.Handler)
+}
+
+type HealthChecker interface {
```
serathius, line 48: No need for healthChecker to be an interface. alias on function is enough.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+func NewHealthChecker(check func() error) HealthChecker {
+	return &healthzCheck{check}
+}
+
+func (h *HealthHandler) addDefaultHealthChecks() error {
+	// ----------------- checks that should be included both in livez and readyz -----------------
+	// Add a default simping ping check.
+	h.AddHealthCheck("ping", PingHealthz, true, true)
+	serializableReadCheck := NewHealthChecker(func() error {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
serathius, line 118: Context should be passed from the handler, so it should be an argument to the HealthChecker

chaochn47: I guess what you meant is pass the hander `http.Request.Context()` to each health checker so client cancellation can properly propagates.   

Looks like [existing health probe](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/etcdhttp/health.go#L181-L185) has the same problem.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+		}
+		fmt.Fprint(w, healthStatus.Reason)
+		fmt.Fprintf(w, "%s check passed\n", path)
+	}
+}
+
+// adaptCheckToHandler returns an http.HandlerFunc that serves the provided checks.
+func adaptCheckToHandler(c func() error) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		err := c()
```
chaochn47, line 324: `r.Context()` needs to pass to each check.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+	for _, name := range names {
+		quoted = append(quoted, fmt.Sprintf("%q", name))
+	}
+	return strings.Join(quoted, ",")
+}
+
+// checkAlarm checks if a specific alarm type is active in the server.
+func checkAlarm(srv EtcdServerHealth, at etcdserverpb.AlarmType) error {
+	as := srv.Alarms()
+	if len(as) > 0 {
```
chaochn47, line 346: if seems unnecessary.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+		readyzChecks:     []string{},
+	}
+	if err = handler.addDefaultHealthChecks(); err != nil {
+		return nil, err
+	}
+	return handler, nil
+}
+
+// PingHealthz returns true automatically when checked
+var PingHealthz HealthChecker = ping{}
```
chaochn47, line 90: I dont think we need a ping check since serializable read should cover that.

server/etcdserver/api/v3health/healthz.go
```diff
@@ -0,0 +1,378 @@
...
+// NewHealthChecker returns a healthz checker for the given function.
+func NewHealthChecker(check func() error) HealthChecker {
+	return &healthzCheck{check}
+}
+
+func (h *HealthHandler) addDefaultHealthChecks() error {
+	// ----------------- checks that should be included both in livez and readyz -----------------
+	// Add a default simping ping check.
+	h.AddHealthCheck("ping", PingHealthz, true, true)
+	serializableReadCheck := NewHealthChecker(func() error {
```
chaochn47, line 117: serializable read will be blocked by defrag. So health checks order is important. I suggest we can simply define all the health checkers first and then plumb through together in the future PRs.

server/etcdserver/healthz_test.go
chaochn47: > make fix does not detect the 2023 errors.

Feel free to create a issue. 

> make verify is broken due to the protogen error sed: 1: "s#import \(#import \("g ...": RE error: parentheses not balanced, which also broken for the main branch. I don't know where the error is.

The etcd scripts are not developed and tested in Mac so the `sed` version installed on Mac may not be the same. This should be reflected in the new contributor play book. 

Feel free to resolve the conversation if the comment is resolved.

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func (reg *CheckRegistry) InstallHttpEndpoints(mux *http.ServeMux, lg *zap.Logger) {
+	lg.Info(fmt.Sprintf("Installing %s http endpoint.", reg.path))
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	cancel()
+	mux.Handle(reg.path, reg.handleRootHealth(ctx, lg))
+	for checkName, check := range reg.checks {
+		mux.Handle(fmt.Sprintf("%s/%v", reg.path, checkName), adaptCheckToHandler(ctx, check))
+	}
```
chaochn47, line 72: It does not look correct. I think the health checker needs to respect client request context, instead of the hard-coded timeout configured by server. 

Please check  https://github.com/etcd-io/etcd/pull/16651#discussion_r1340716818 and https://github.com/etcd-io/etcd/pull/16651#discussion_r1340718539 and https://pkg.go.dev/net/http#Request.Context

Could you please write a integration test to validate it? Let's say a client disconnects right after sending the requests to server, will server waste the time to do health checks.

siyuanfoundation:
@chaochn47 @serathius Restructured the code  based on review feedbacks. Please re-review the code. Thanks!

server/etcdserver/api/v3health/healthz.go
siyuanfoundation: I don't think it is necessary to define all heath checkers. Refactored the code so that it is easier to track which order the checks are added.

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+			} else {
+				// no-op the check if we've specified we want to exclude the check
+				if _, found := excluded[checkName]; found {
+					delete(excluded, checkName)
+					fmt.Fprintf(&individualCheckOutput, "[+]%s excluded: ok\n", checkName)
+					continue
+				}
+			}
+			check := reg.checks[checkName]
+			if err := check(r.Context()); err != nil {
```
chaochn47, line 103: IMHO, code readability matters more than time complexity. So instead of mixing the exclude and check logic, I would suggest

1. get a list of checks given (reg.checkNames, excluded and included).
2. return http bad request to client if the intended (excluded/included) checks are not registered. So it forces administrator to configure it correctly because etcd will be in a crash back off state. 
3. run each check individually.

ahrtr:
Mark this PR as draft until the related design doc is well reviewed and approved.

ahrtr:
link to https://github.com/etcd-io/etcd/issues/16007

ahrtr (commented):
See also https://github.com/etcd-io/etcd/issues/16007#issuecomment-1740646481

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+		}
+		return nil
+	}
+}
+
+func checkSerializableRead(srv ServerHealth) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		_, err := srv.Range(ctx, &etcdserverpb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: true})
+		if err != nil && err != auth.ErrUserEmpty && err != auth.ErrPermissionDenied {
```
serathius, line 197: I think this check should work even if authorization is enabled.

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+			return fmt.Errorf("RANGE ERROR:%s", err)
```
serathius, line 198:
```suggestion
			return fmt.Errorf("RANGE ERROR: %s", err)
```

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+	return set
+}
+
+// checkAlarm checks if a specific alarm type is active in the server.
+func checkAlarm(srv ServerHealth, at etcdserverpb.AlarmType) func(context.Context) error {
+	return func(ctx context.Context) error {
+		as := srv.Alarms()
+		for _, v := range as {
+			if v.Alarm == at {
+				return fmt.Errorf("Alarm active:%s", at.String())
```
serathius, line 187:
```suggestion
				return fmt.Errorf("Alarm active: %s", at.String())
```

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
+// Copyright 2023 The etcd Authors
```
serathius, line 1: File name doesn't make sense.

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+	"net/http"
+	"strings"
+
+	"go.uber.org/zap"
+
+	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/server/v3/auth"
+)
+
+func InstallHealthEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
```
serathius, line 30: I think the `InstallHealthEndpoints` function is the only function that needs to be public. All other methods are not used externally.

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+func (reg *CheckRegistry) Register(name string, check HealthCheck) {
+	reg.checkNames = append(reg.checkNames, name)
+	reg.checks[name] = check
+}
+
+func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
+	lg.Info(fmt.Sprintf("Installing %s http endpoint.", reg.path))
+	mux.Handle(reg.path, reg.handleRootHealth(lg))
+	for checkName, check := range reg.checks {
+		mux.Handle(fmt.Sprintf("%s/%v", reg.path, checkName), adaptCheckToHandler(check))
```
serathius, line 67: `adaptCheckToHandler` function feels a little akward. Have you thought on instead having a `InstallCheck` method on `CheckRegistry`. It could look like 

```
func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
	lg.Info(fmt.Sprintf("Installing %s http endpoint.", reg.path))
   mux.Handle(reg.path, reg.handleRootHealth(lg))
	for checkName, check := range reg.checks {
	    reg.installCheckEndpoint(checkName, check, mux)
	}
}

func (reg *CheckRegistry) installCheckEndpoint(name string, check HealthCheck, mux *http.ServeMux) {
  mux.Handle(path.Join(reg.path, name), func(w http.ResponseWriter, r *http.Request) {
		err := check(r.Context())
		if err != nil {
			http.Error(w, fmt.Sprintf("internal server error: %v", err), http.StatusInternalServerError)
		} else {
			fmt.Fprint(w, "ok")
		}
	})
}
```

serathius, line 67: Use `path.Join(reg.path, checkName)` instead of `fmt.Sprintf("%s/%v", reg.path, checkName)`

server/etcdserver/api/etcdhttp/readiness.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/readiness.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+	reg.InstallHttpEndpoints(lg, mux)
+}
+
+func (reg *CheckRegistry) Register(name string, check HealthCheck) {
+	reg.checkNames = append(reg.checkNames, name)
+	reg.checks[name] = check
+}
+
+func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
+	lg.Info(fmt.Sprintf("Installing %s http endpoint.", reg.path))
```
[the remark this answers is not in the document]
siyuanfoundation: done

server/etcdserver/api/etcdhttp/readiness.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+	for checkName, check := range reg.checks {
+		mux.Handle(fmt.Sprintf("%s/%v", reg.path, checkName), adaptCheckToHandler(check))
+	}
+}
+
+// handleRootHealth returns an http.HandlerFunc that serves the provided checks.
+func (reg *CheckRegistry) handleRootHealth(lg *zap.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// extracts the health check names to be excludeList from the query param
+		excludeList, _ := r.URL.Query()["exclude"]
```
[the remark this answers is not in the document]
siyuanfoundation: done

server/etcdserver/api/etcdhttp/readiness.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/readiness.go
```diff
@@ -0,0 +1,202 @@
...
+// preserving the order passed in.
+func formatQuoted(names ...string) string {
+	quoted := make([]string, 0, len(names))
+	for _, name := range names {
+		quoted = append(quoted, fmt.Sprintf("%q", name))
+	}
+	return strings.Join(quoted, ",")
+}
+
+type stringSet map[string]struct{}
```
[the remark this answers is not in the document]
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -185,6 +190,156 @@ func TestHealthHandler(t *testing.T) {
...
+			defer ts.Close()
+			checkHttpResponse(t, ts, tt.healthCheckURL, tt.expectStatusCode, tt.inResult, tt.notInResult)
+		})
+	}
+}
+
+func checkHttpResponse(t *testing.T, ts *httptest.Server, url string, expectStatusCode int, inResult []string, notInResult []string) {
+	res, err := ts.Client().Do(&http.Request{Method: http.MethodGet, URL: testutil.MustNewURL(t, ts.URL+url)})
+	if err != nil {
+		t.Errorf("fail serve http request %s %v", url, err)
```
fuweid, line 313: suggest to use t.Fatal to abort.

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -185,6 +190,156 @@ func TestHealthHandler(t *testing.T) {
...
+	}
+	if res == nil {
```
fuweid, line 315: Ideally, the `res` is not nil if err = nil

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.
> REF: https://pkg.go.dev/net/http#Client.Do

And the defer to close body should be called before checking the status code.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -194,3 +201,183 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+		for _, v := range as {
+			if v.Alarm == at {
+				return fmt.Errorf("Alarm active: %s", at.String())
+			}
+		}
+		return nil
+	}
+}
+
+func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
```
fuweid, line 375: Follow-up: we can consider to merge the `checkAPI` with this function.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -15,10 +15,13 @@
 package etcdhttp
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"path"
+	"strings"
```
fuweid, line 24: Would you please rebase the main branch? we fix the linter issue.

At the Line 29, the `"go.etcd.io/etcd/api/v3/etcdserverpb"` is duplicated with `pb "go.etcd.io/etcd/api/v3/etcdserverpb"`. Not sure that the linter will be broken after merged. 
I think it's safe to rebase and rerun the linter. Thanks

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -194,3 +201,183 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+		excluded := getQuerySet(r, "exclude")
+		// extracts the health check names to be allowList from the query param
+		included := getQuerySet(r, "allowlist")
+
+		filteredCheckNames, err := filterCheckList(reg.checkNames, included, excluded)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		h, _ := reg.runHealthChecks(r.Context(), filteredCheckNames)
```
fuweid, line 278: Since the Line 278 doesn't check the error, the `h.Health` can be true I think.
I think it's unexpected.

server/embed/etcd.go
```diff
@@ -744,6 +744,7 @@ func (e *Etcd) serveClients() (err error) {
...
+	etcdhttp.InstallHealthEndpoints(e.cfg.logger, mux, e.Server)
```
serathius, line 747: Why not just use etcdhttp.HandleHealth?

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -182,14 +189,195 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+	reg.installHttpEndpoints(lg, mux)
+}
+
+func (reg *CheckRegistry) register(name string, check HealthCheck) {
+	reg.checkNames = append(reg.checkNames, name)
+	reg.checks[name] = check
+}
+
+func (reg *CheckRegistry) installHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
+	lg.Info(fmt.Sprintf("Installing %s http endpoint.", reg.path))
```
serathius, line 233:
```suggestion
	lg.Info("Installing http endpoint", "path", reg.path)
```

serathius: Note, we don't log registration of any other endpoint. I don't see a reason to make an exception add this logging just for health endpoints.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -118,78 +133,251 @@ type Health struct {
...
+	checkNames []string
+}
+
+func DefaultCheckRegistry(path string) *CheckRegistry {
+	reg := CheckRegistry{path: path, checks: make(map[string]HealthCheck), failCounter: healthFailed, successCounter: healthSuccess}
+	return &reg
+}
+
+// installHealthEndpoints installs endpoints for /health
+func installHealthEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
```
serathius, line 172: I see you did a full refactor of the existing health endpoint to the CheckRegistry. This is good change, however it makes the PR harder to review as reviewer not only needs to check the new health, but also validate if there were not unintentional changes to the old health. 

Could you consider moving the refactor to a separate PR?

server/etcdserver/api/etcdhttp/readiness.go
chaochn47: > Do you mean there is a way to override the permission or livez should fail if the requester is not authorized?

@siyuanfoundation I guess what @serathius meant is if authorization is enabled, the serializable check won't exercise the code path it wants to (read a key from mvcc) and return permission error early. 

And I have a idea to fix this existing implementation limitation. 

```
// WithRoot generates and installs a token that can be used as a root credential
```

`ctx = srv.AuthStore().WithRoot(ctx)`

It should generate a token valid for the range request to pass permission check. It has been used in [revokeExpiredLeases](https://github.com/etcd-io/etcd/blob/628b45c09921b11a4f61a9c5dd7b371e76cd5b59/server/etcdserver/server.go#L874). Ideally I want it to be read only root permissions to reduce security risks exposure since the livez check can be triggered from outside. 

However, I suggest we can keep the PR as it is and raise a separate issue to explain it and fix this.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +253,187 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+}
+
+func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
+	reg.installRootEndpoint(lg, mux)
+	for checkName, check := range reg.checks {
+		reg.installCheckEndpoint(checkName, check, mux)
+	}
+}
+
+func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames []string) (Health, error) {
```
chaochn47, line 294: Don't understand why `runHealthChecks` cannot reused in individual check handler. 

Otherwise, there is no metric emitted for individual check.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,164 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+		}
+		filteredList = append(filteredList, chk)
+	}
+	if len(excluded) > 0 {
+		return nil, fmt.Errorf("some health checks cannot be excluded: no matches for %s", formatQuoted(excluded.List()...))
+	}
+	return filteredList, nil
+}
+
+func filterIncludedChecks(checks []string, included StringSet) (filteredList []string, err error) {
```
logicalhan, line 332: why do we need included checks?

siyuanfoundation: Allowlist could be a nice option for users that only care about a few selected checks and don't want be affected by newly added checks.

server/etcdserver/api/etcdhttp/readiness.go
chaochn47: Sync'd up with @siyuanfoundation. 

“Failing health check if exclude query contains some non registered check” would be a blocker for us to remove a check. So we could switch to log an error with warn log level. Downside is admin needs to check the warn log to figure out if the prober setting is correct instead. 

Version compatibility outweighs forcing administrator to configure it correctly.. 

ref. https://github.com/kubernetes/kubernetes/pull/70676#discussion_r231610436

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: After checking the old discussions in https://github.com/kubernetes/kubernetes/pull/70676#discussion_r231291895 and discussing with Chao, we decided to follow Kubernetes convention and drop the allowlist, and make exclude non-exist check not a failure. Thank you @logicalhan for bringing the context to our attention!

server/etcdserver/api/etcdhttp/health_test.go
fuweid: Not sure that it's related to my chrome cache. The change seems be reverted before this comment.

logicalhan:
I think this isn't getting reviewed because the PR is in a draft state..

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,137 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+	for _, chk := range checks {
+		if _, found := excluded[chk]; found {
+			delete(excluded, chk)
+			continue
+		}
+		filteredList = append(filteredList, chk)
+	}
+	if len(excluded) > 0 {
+		return filteredList, fmt.Errorf("some health checks cannot be excluded: no matches for %s", formatQuoted(excluded.List()...))
+	}
```
chaochn47, line 314: It seems weird that the error was then logged in line 296, could you please log it here and update the function signature accordingly?

server/etcdserver/api/etcdhttp/health_test.go
chaochn47: @siyuanfoundation Could you please take a second look at this one? It may not be necessary but it's a good code hygiene to keep :)

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -49,136 +52,287 @@ func (s *fakeHealthServer) Config() config.ServerConfig {
...
+			apiError:         auth.ErrUserEmpty,
+			expectStatusCode: http.StatusServiceUnavailable,
+			inResult:         []string{"RANGE ERROR: auth: user name is empty"},
+		},
+		{
+			name:             "Not alive if authorization failed",
+			healthCheckURL:   "/livez",
+			apiError:         auth.ErrPermissionDenied,
+			expectStatusCode: http.StatusServiceUnavailable,
+		},
```
chaochn47, line 265: Those test cases are no longer valid IMO.

chaochn47 (approved):

chaochn47:
LGTM overall, nice work @siyuanfoundation

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health_test.go
siyuanfoundation: removed

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -49,136 +52,271 @@ func (s *fakeHealthServer) Config() config.ServerConfig {
...
 }
 
+func checkHttpResponse(t *testing.T, ts *httptest.Server, url string, expectStatusCode int, inResult []string, notInResult []string) {
+	res, err := ts.Client().Do(&http.Request{Method: http.MethodGet, URL: testutil.MustNewURL(t, ts.URL+url)})
+
+	if err != nil {
+		t.Errorf("fail serve http request %s %v", url, err)
+	}
+	health, err := parseHealthOutput(res.Body)
+	defer res.Body.Close()
```
fuweid, line 289: It seems that the original commit has been changed and it doesn't look good here.
The status code should be checked first and then read the body to get health info.
And the t.Errorf doesn't abort the progress and it will continue to work until return.

siyuanfoundation: @fuweid Thanks for reviewing. How does it look now?

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,135 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+		} else {
+			fmt.Fprintf(&individualCheckOutput, "[+]%s ok\n", checkName)
+		}
+	}
+	h.Reason = individualCheckOutput.String()
+	if len(failedChecks) > 0 {
+		h.Health = "false"
+	}
+	return h, nil
+}
```
serathius, line 286:
```suggestion
func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...string) (Health, error) {
	h := Health{Health: "true"}
	var individualCheckOutput bytes.Buffer
	for _, checkName := range checkNames {
		check, found := reg.checks[checkName]
		if !found {
			return Health{Health: "false"}, fmt.Errorf("Health check: %s not registered", checkName)
		}
		if err := check(ctx); err != nil {
			fmt.Fprintf(&individualCheckOutput, "[-]%s failed: %v\n", checkName, err)
			h.Health = "false"
		} else {
			fmt.Fprintf(&individualCheckOutput, "[+]%s ok\n", checkName)
		}
	}
	h.Reason = individualCheckOutput.String()
	return h, nil
}
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,135 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
 	lg.Debug("serving /health true")
 	return h
 }
+
+type HealthCheck func(ctx context.Context) error
```
serathius, line 228: High level note on the code. No need to address it now.

Code below has a mixed order, sometimes the calling function is below (installHttpEndpoint after runHealthChecks) and sometimes above (InstallHttpEndpoints before installHttpEndpoint).

Recommend use a single style. Either top to bottom or bottom to top. My personal preference is top to bottom, because we no longer are bounded by C language limits. It helps putting the top most important function on the top of the file, while the implementation details and helpers on the bottom.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,135 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+	reg.checkNames = append(reg.checkNames, name)
+	reg.checks[name] = check
+}
+
+func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
+	// installs the http handler for the root path.
+	reg.installHttpEndpoint(lg, mux, reg.path, reg.checkNames...)
+	for _, checkName := range reg.checkNames {
+		// installs the http handler for the individual check sub path.
+		reg.installHttpEndpoint(lg, mux, path.Join(reg.path, checkName), checkName)
```
serathius, line 261: Not sure I like that you are reusing `installHttpEndpoint` function to install checking of multiple and single checks. It unintentionally leaks `exclude` logic into the singular check logic. I have a strong preference for code to always be intentional and not have hidden side effects. After merging the code becomes the source of truth on how etcd should behave, obfuscating the intention.

I understand that this should not affect the calls if it's not used, but by adding to the API you make it a part of expected API behavior that adds unnecessary complexity. When someone will be refactoring this code in the future, they will need to understand if this is officially supported behavior. Some user could encounter it by mistake and start depending on this.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,135 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+		return nil
+	}
+}
+
+func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		ctx = srv.AuthStore().WithRoot(ctx)
+		_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: true})
+		if err != nil {
+			return fmt.Errorf("RANGE ERROR: %s", err)
```
serathius, line 354: Any reason for this to be capitalized?

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -49,136 +52,271 @@ func (s *fakeHealthServer) Config() config.ServerConfig {
...
+			alarms:           []*pb.AlarmMember{{MemberID: uint64(0), Alarm: pb.AlarmType_CORRUPT}, {MemberID: uint64(0), Alarm: pb.AlarmType_NOSPACE}},
+			healthCheckURL:   "/readyz/data_corruption",
+			expectStatusCode: http.StatusServiceUnavailable,
+		},
+		{
+			name:             "/ready/serializable_read ok if CORRUPT alarm is on",
+			alarms:           []*pb.AlarmMember{{MemberID: uint64(0), Alarm: pb.AlarmType_CORRUPT}, {MemberID: uint64(0), Alarm: pb.AlarmType_NOSPACE}},
+			healthCheckURL:   "/readyz/serializable_read",
+			expectStatusCode: http.StatusOK,
+		},
```
serathius, line 223: This case is misleading. I expect that in real etcd corrupt alarm will cause serializable read to fail. Not sure why we test serializable read in the `TestDataCorruptionCheck` test.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +224,135 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
 	return h
 }
+
+type HealthCheck func(ctx context.Context) error
+
+type CheckRegistry struct {
+	path string
+
+	checks map[string]HealthCheck
+	// the list of checkNames are needed to ensure the order of checks.
```
serathius, line 234: Why do we need to ensure order of checks?

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -91,6 +98,32 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(excludedAlarms AlarmSet, Serial
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			w.Header().Set("Allow", http.MethodGet)
+			http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+			lg.Warn("/health error", zap.Int("status-code", http.StatusMethodNotAllowed))
+			return
+		}
+		h, err := hfunc(r)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
serathius, line 112: Returning BadRequest here seems incorrect. See line 272. `runHealthChecks` might return `fmt.Errorf("Health check: %s not registered", checkName)` which is a programmer error not a client error like `StatusBadRequest` https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400.

`health check not registered` error can only happen if `CheckRegistry` has a bug that would cause `checks` and `checkNames` to be inconsistent. This is a invariant of `CheckRegistry` as so it should never normally happen, but when it does we are screwed. 

I would definitely not attribute this to client error, and I would consider `CheckRegistry` panic instead.

server/etcdserver/api/etcdhttp/health_test.go
fuweid: `defer res.Body.Close()` should be called after Line 287.
Line 286 should use `t.Fatal` to return fast.

server/etcdserver/api/etcdhttp/health.go
chaochn47: It was under the assumption if one of checks failed, handler return false early. This statement no longer stays so we should move it.

server/etcdserver/api/etcdhttp/health_test.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health_test.go
siyuanfoundation: Moved subpath check to a separate test.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: As @chaochn47 said, it was added due to an old assumption. Removed.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: changed to panic.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: ack

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -91,6 +98,31 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(excludedAlarms AlarmSet, Serial
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			w.Header().Set("Allow", http.MethodGet)
+			http.Error(w, "Method Not Allowed", http.StatusMethodNotAllowed)
+			lg.Warn("/health error", zap.Int("status-code", http.StatusMethodNotAllowed))
+			return
+		}
+		h, err := hfunc(r)
+		if err != nil {
+			panic(err)
```
serathius, line 112: From this place it's hard to know why error here should cause panic. I would recommend removing error from hfunc return, and panic on line 284 directly.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -91,6 +98,31 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(excludedAlarms AlarmSet, Serial
...
+		}
+		d, _ := json.Marshal(h)
```
serathius, line 114: Sorry for not noticing it earlier. The proposed API is a strange mix of etcd and Kubernetes one. A mix of JSON and custom text format. 

Just see the response. I will be 
```
{
 "health": "true",
 "reason": "[-]data_corruption failed: Alarm active: CORRUPT\n[+]serializable_read ok\n",
}
```

This mix is worse than using a single format, as it requires to use both json and custom parser for client to get a failed check name. We should stick to either JSON or to Kubernetes text format. 

How Kubernetes response should look like:
```
[-]data_corruption failed: Alarm active: CORRUPT
[+]serializable_read
```

How pure json response should look like:
```
{
 "health": "true",
 "checks": {"
   data_corruption": {
     "status": "failed",
      "error": "Alarm active: CORRUPT"
    }, 
    "serializable_read": {
      "status": "ok",
    }
}
```

My recommendation is to use Kubernetes. Reason is that Etcd became part of Kubernetes and having a consistent format should make it easier to integrate for users vs custom format. If you like to support JSON, we could use [content type negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation) as a way to serve Json to client that will request it.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +198,162 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+	checks map[string]HealthCheck
+}
+
+func installLivezEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
+	reg := CheckRegistry{path: "/livez", checks: make(map[string]HealthCheck)}
+	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.InstallHttpEndpoints(lg, mux)
+}
+
+func installReayzEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
```
logicalhan, line 215: `Readyz`

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +198,162 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+		fmt.Fprint(w, "ok\n")
+		lg.Debug("Health check OK", zap.String("path", path), zap.String("reason", h.Reason), zap.Int("status-code", http.StatusOK))
+	}
+}
+
+func filterCheckList(lg *zap.Logger, checks []string, excluded StringSet) []string {
+	filteredList := []string{}
+	for _, chk := range checks {
+		if _, found := excluded[chk]; found {
+			delete(excluded, chk)
```
fuweid, line 305: no need to remove the `excluded`. For the design, if the checks contains duplicate items, all the same items should be excluded as well instead of first one.

fuweid (approved):
LGTM

Nice work!

server/etcdserver/api/etcdhttp/health.go
ahrtr: Is this comment resolved?

ahrtr:
Please rebase this PR and resolve the left comment, thx

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +198,162 @@ func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
...
+			lg.Error("Health check error", zap.String("path", path), zap.String("reason", h.Reason), zap.Int("status-code", http.StatusServiceUnavailable))
+			return
+		}
+		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+		w.Header().Set("X-Content-Type-Options", "nosniff")
+		// Only writes detailed reason for verbose requests.
+		if _, found := r.URL.Query()["verbose"]; found {
+			fmt.Fprint(w, h.Reason)
+		}
+		fmt.Fprint(w, "ok\n")
```
serathius, line 296: Any reason to implement verbose for per check endpoint?

server/etcdserver/api/etcdhttp/health.go
serathius: No, checks should include duplicates, but for `filterCheckList` function to make sense `checks` should not have type `[]string`, but also `StringSet`.

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -49,144 +51,279 @@ func (s *fakeHealthServer) Config() config.ServerConfig {
...
-				return
-			}
-			if res.StatusCode != tt.expectStatusCode {
-				t.Errorf("want statusCode %d but got %d", tt.expectStatusCode, res.StatusCode)
+func TestHttpSubPath(t *testing.T) {
+	be, _ := betesting.NewDefaultTmpBackend(t)
+	defer betesting.Close(t, be)
+	tests := []healthTestCase{
+		{
+			name:             "/ready/data_corruption ok",
```
serathius, line 176:
```suggestion
			name:             "/readyz/data_corruption ok",
```

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -49,144 +51,279 @@ func (s *fakeHealthServer) Config() config.ServerConfig {
...
+			healthCheckURL:   "/readyz/data_corruption",
+			expectStatusCode: http.StatusOK,
+		},
+		{
+			name:             "/ready/serializable_read not ok with error",
```
serathius, line 181:
```suggestion
			name:             "/readyz/serializable_read not ok with error",
```

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: per check endpoint still print the detailed error message in verbose, and no details in non verbose. In theory, the user can get all the details from the root path verbose. But I think it still makes sense to follow the same paradigm.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: Updated signature of filterCheckList

siyuanfoundation:
@ahrtr Rebased the PR. Please review. Thanks!

logicalhan (approved):
/lgtm

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +197,170 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
...
+	}
+}
+
+func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...string) Health {
+	h := Health{Health: "true"}
+	var individualCheckOutput bytes.Buffer
+	for _, checkName := range checkNames {
+		check, found := reg.checks[checkName]
+		if !found {
+			panic(fmt.Errorf("Health check: %s not registered", checkName))
```
ahrtr, line 248: Why panic? Clients can easily put down etcdserver using a random checkname? it's a security hole to me.

serathius: Note, panic in http server is recovered.
Also note that this path cannot be reached by client. CheckRegistry only registers endpoints for existing checks, so client cannot provide a custom check here.

I proposed to have a panic here as this is unrecoverable error caused by programmer error. We could change it to `lg.Panic` to make it more consistent.

Unrecoverable because there is no good response we can give to user, something went wrong in etcd server that caused it to pick an unexisting check. 

It's a programmer error, and it works more like an assertion. Alternative would be a nil pointer exception if we didn't check this.

ahrtr: So users will get a 404 when accessing a non-exist check (path)?

siyuanfoundation: Yes, it would be a 404. Added a unit test for it.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +197,170 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
...
+	}
+
+	// installs the http handler for the root path.
+	reg.installRootHttpEndpoint(lg, mux, reg.path, checkNames...)
+	for _, checkName := range checkNames {
+		// installs the http handler for the individual check sub path.
+		subpath := path.Join(reg.path, checkName)
+		check := checkName
+		mux.Handle(subpath, newHealthHandler(subpath, lg,
+			func(r *http.Request) Health { return reg.runHealthChecks(r.Context(), check) }))
```
ahrtr, line 238:
```suggestion
		mux.Handle(subpath, newHealthHandler(subpath, lg, func(r *http.Request) Health { 
			return reg.runHealthChecks(r.Context(), check) 
		}))			
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +197,170 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
...
+	return set
+}
+
+// activeAlarmCheck checks if a specific alarm type is active in the server.
+func activeAlarmCheck(srv ServerHealth, at pb.AlarmType) func(context.Context) error {
+	return func(ctx context.Context) error {
+		as := srv.Alarms()
+		for _, v := range as {
+			if v.Alarm == at {
+				return fmt.Errorf("Alarm active: %s", at.String())
```
ahrtr, line 350:
```suggestion
				return fmt.Errorf("alarm activated: %s", at.String())
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +197,170 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
...
+		return nil
+	}
+}
+
+func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		ctx = srv.AuthStore().WithRoot(ctx)
+		_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: true})
+		if err != nil {
+			return fmt.Errorf("Range error: %s", err)
```
ahrtr, line 362:
```suggestion
			return fmt.Errorf("range error: %w", err)
```

ahrtr: FYI. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ahrtr (approved):
LGTM

Thanks for your first contribution, and great work!

Please also add a followup item to update the doc in [etcd-io/website](https://github.com/etcd-io/website).

serathius (approved):

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -193,3 +197,171 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
...
+type HealthCheck func(ctx context.Context) error
+
+type CheckRegistry struct {
+	path   string
+	checks map[string]HealthCheck
+}
+
+func installLivezEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
+	reg := CheckRegistry{path: "/livez", checks: make(map[string]HealthCheck)}
+	reg.Register("serializable_read", serializableReadCheck(server))
```
serathius, line 210: QQ, doesn't serializable read require store not to be corrupted?

ahrtr: It doesn't help to restart etcd on data corruption, so data corruption isn't a check for livez.

But if etcd (bbolt) crashes due to data corruption, then it's another story. It needs human intervention anyway in that case.

serathius: I'm not asking "Why there is no corrupt_alarm check?". I'm asking, if etcd cluster is corrupted, I expect every etcd member to return error on serializable read. Meaning that serializable_read is also checking corrupt alarm. 

I will double check the behavior.

serathius: Oh, wait. I was thinking about corrupt applier. Might not be the case as we read doesn't go through checking corrupt. Hmm, but this might also be a weird behavior.

serathius: Ok, wrote a PR that tests livez readyz and confirms that livez is not triggered by corruption. https://github.com/etcd-io/etcd/pull/16784

serathius:
Please note that this PR doesn't implement the full design as it was presented. Livez works correct, however we are missing checks for readyz making it not very useful. Please see https://github.com/etcd-io/etcd/pull/16784/commits/80ab2ad91cdc04a995e6dbf3315c0884b248c11e

I think this is ok to merge, however until we finish implementing readyz, we shouldn't backport nor document the new endpoints.

serathius (approved):

PR #16655, opened 2023-09-26
Fix RaftCluster boostrap being based on v3 store while apply is based on v2 store.

serathius:
Migration off v2 store was unfinished during v3.5 development https://github.com/etcd-io/etcd/pull/12914 leaving RaftCluster broken, half migrated. This was found in v3.5 and fixed in https://github.com/etcd-io/etcd/pull/13348, however this is was not patched on main branch.

This PR makes v3.6 RaftCluster consistent with v3.5 so we can cleanly migrate off v2 store.

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -260,12 +260,12 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 	c.Lock()
 	defer c.Unlock()
 
-	if c.be != nil {
-		c.version = c.be.ClusterVersionFromBackend()
-		c.members, c.removed = c.be.MustReadMembersFromBackend()
-	} else {
+	if c.v2store != nil {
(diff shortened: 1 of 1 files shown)


serathius:
cc @ahrtr 
This PR is to clarify issue with the current RaftCluster. Don't think we necessarily need to merge this PR.

ahrtr:
I do not see any need to forward port https://github.com/etcd-io/etcd/pull/13348 to 3.6(main branch), as long as we are going to remove v2store in 3.6 [of course, we also need to generate v2 snapshot using v3 store]. It's also the reason why https://github.com/etcd-io/etcd/pull/13348 was resolved in 3.5 directly instead of the main branch.

There is one possibility that the [Zombie members issue](https://github.com/etcd-io/etcd/issues/13196) might be carried over to 3.6, when users upgrade 3.4 to 3.5.0 (instead of 3.5.1+), and then to 3.6. But it is unlikely in practice just https://github.com/etcd-io/etcd/issues/13196#issuecomment-914472617 mentions.

PR #16656, opened 2023-09-26
Bootstrap from v3 store

serathius:
Very drafty draft for bootstraping etcd from v3 store.

Looks like there is no function in raft.MemoryStorage to start from ConsistentIndex + Confstate instead of snapshot.
So I forced it into a ApplySnapshot which I don't think it correct. Still it worked on trivial case for me locally.

I think there are still some edge cases to cover, for example:
* what happens if etcd exits before it flushes confstate and CI to db.

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -305,7 +305,7 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 // ensures that it is still valid.
 func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
 	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+	membersMap, removedMap := c.be.MustReadMembersFromBackend()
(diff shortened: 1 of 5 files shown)


geetasg:
Current source code uses the "shouldApplyV3" logic to avoid duplicate apply to backend for a normal apply request. Ref:https://github.com/etcd-io/etcd/blob/main/server/etcdserver/apply/uber_applier.go#L149
This logic of using shouldApplyV3 has been extended to membership as well Eg:https://github.com/etcd-io/etcd/blob/main/server/etcdserver/api/membership/cluster.go#L396 
Validation method has not switched over.

@serathius can you please comment if there is any potential correctness problem with using the shouldApplyV3 logic?

serathius:
> @serathius can you please comment if there is any potential correctness problem with using the shouldApplyV3 logic?

No, shouldApplyV3 logic is just result of checking of the index of an entry that is being applied greater then v3 store Index `e.Index > s.consistIndex.ConsistentIndex()` . If we bootstrap from v3, it just means `e.Index > s.consistIndex.ConsistentIndex()` will just always be true.

https://github.com/etcd-io/etcd/blob/main/server/etcdserver/server.go#L1835

PR #16657, opened 2023-09-27
pkg/stringutil: sort 'ss' in TestUniqueStrings

gocurr:
From the algorithm below, 'ss' should be sorted.

Also removes 'fmt.Println', because the idiomatic tests would not print.


jmhbnz (approved):
LGTM - Thanks for fixing this @gocurr

serathius (approved):

ahrtr (approved):

PR #16658, opened 2023-09-27
Fix memberList may return incorrect intermediate results right after bootstrap

ahrtr:

The first commit added a test case, basically which always fails,

```
2023-09-27T16:26:55.3567789Z     ctl_v3_member_test.go:128: 
2023-09-27T16:26:55.3568428Z         	Error Trace:	/__w/etcd/etcd/tests/e2e/ctl_v3_member_test.go:128
2023-09-27T16:26:56.3453793Z         	            				/__t/go/1.21.1/x64/src/runtime/asm_amd64.s:1650
2023-09-27T16:26:56.3455109Z         	Error:      	Not equal: 
2023-09-27T16:26:56.3456321Z         	            	expected: 1
2023-09-27T16:26:56.3457394Z         	            	actual  : 2
```

After applying the patch in the second commit, then the issue is resolved.

The third commit is fixing the broken unit test in server_test.go, and is coming from @geetasg 's PR https://github.com/etcd-io/etcd/pull/16084. I already set the author as @geetasg

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -395,19 +395,27 @@ func (c *RaftCluster) AddMember(m *Member, shouldApplyV3 ShouldApplyV3) {
 	}
 	if c.be != nil && shouldApplyV3 {
 		c.be.MustSaveMemberToBackend(m)
+
+		c.members[m.ID] = m
+		c.updateMembershipMetric(m.ID, true)
+
+		c.lg.Info(
+			"added member",
+			zap.String("cluster-id", c.cid.String()),
+			zap.String("local-member-id", c.localID.String()),
+			zap.String("added-peer-id", m.ID.String()),
+			zap.Strings("added-peer-peer-urls", m.PeerURLs),
+			zap.Bool("added-peer-is-learner", m.IsLearner),
+		)
+	} else {
+		c.lg.Info(
+			"ignore already added member",
+			zap.String("cluster-id", c.cid.String()),
+			zap.String("local-member-id", c.localID.String()),
+			zap.String("added-peer-id", m.ID.String()),
+			zap.Strings("added-peer-peer-urls", m.PeerURLs),
+			zap.Bool("added-peer-is-learner", m.IsLearner))
 	}
-
-	c.members[m.ID] = m
-	c.updateMembershipMetric(m.ID, true)
-
-	c.lg.Info(
-		"added member",
-		zap.String("cluster-id", c.cid.String()),
-		zap.String("local-member-id", c.localID.String()),
-		zap.String("added-peer-id", m.ID.String()),
-		zap.Strings("added-peer-peer-urls", m.PeerURLs),
-		zap.Bool("added-peer-is-learner", m.IsLearner),
-	)
 }
 
 // RemoveMember removes a member from the store.
@@ -420,25 +428,32 @@ func (c *RaftCluster) RemoveMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	}
 	if c.be != nil && shouldApplyV3 {
 		c.be.MustDeleteMemberFromBackend(id)
-	}
 
-	m, ok := c.members[id]
-	delete(c.members, id)
-	c.removed[id] = true
-	c.updateMembershipMetric(id, false)
+		m, ok := c.members[id]
+		delete(c.members, id)
+		c.removed[id] = true
+		c.updateMembershipMetric(id, false)
 
-	if ok {
-		c.lg.Info(
-			"removed member",
-			zap.String("cluster-id", c.cid.String()),
-			zap.String("local-member-id", c.localID.String()),
-			zap.String("removed-remote-peer-id", id.String()),
-			zap.Strings("removed-remote-peer-urls", m.PeerURLs),
-			zap.Bool("removed-remote-peer-is-learner", m.IsLearner),
-		)
+		if ok {
+			c.lg.Info(
+				"removed member",
+				zap.String("cluster-id", c.cid.String()),
+				zap.String("local-member-id", c.localID.String()),
+				zap.String("removed-remote-peer-id", id.String()),
+				zap.Strings("removed-remote-peer-urls", m.PeerURLs),
+				zap.Bool("removed-remote-peer-is-learner", m.IsLearner),
+			)
+		} else {
+			c.lg.Warn(
+				"skipped removing already removed member",
+				zap.String("cluster-id", c.cid.String()),
+				zap.String("local-member-id", c.localID.String()),
+				zap.String("removed-remote-peer-id", id.String()),
+			)
+		}
 	} else {
-		c.lg.Warn(
-			"skipped removing already removed member",
+		c.lg.Info(
+			"ignore already removed member",
 			zap.String("cluster-id", c.cid.String()),
 			zap.String("local-member-id", c.localID.String()),
 			zap.String("removed-remote-peer-id", id.String()),
@@ -484,42 +499,62 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	c.Lock()
 	defer c.Unlock()
 
-	c.members[id].RaftAttributes.IsLearner = false
-	c.updateMembershipMetric(id, true)
 	if c.v2store != nil {
-		mustUpdateMemberInStore(c.lg, c.v2store, c.members[id])
-	}
-	if c.be != nil && shouldApplyV3 {
-		c.be.MustSaveMemberToBackend(c.members[id])
+		m := *(c.members[id])
+		m.RaftAttributes.IsLearner = false
+		mustUpdateMemberInStore(c.lg, c.v2store, &m)
 	}
 
-	c.lg.Info(
-		"promote member",
-		zap.String("cluster-id", c.cid.String()),
-		zap.String("local-member-id", c.localID.String()),
-	)
+	if c.be != nil && shouldApplyV3 {
+		c.members[id].RaftAttributes.IsLearner = false
+		c.updateMembershipMetric(id, true)
+		c.be.MustSaveMemberToBackend(c.members[id])
+
+		c.lg.Info(
+			"promote member",
+			zap.String("cluster-id", c.cid.String()),
+			zap.String("local-member-id", c.localID.String()),
+		)
+	} else {
+		c.lg.Info(
+			"ignore already promoted member",
+			zap.String("cluster-id", c.cid.String()),
+			zap.String("local-member-id", c.localID.String()),
+		)
+	}
 }
 
 func (c *RaftCluster) UpdateRaftAttributes(id types.ID, raftAttr RaftAttributes, shouldApplyV3 ShouldApplyV3) {
 	c.Lock()
 	defer c.Unlock()
 
-	c.members[id].RaftAttributes = raftAttr
 	if c.v2store != nil {
-		mustUpdateMemberInStore(c.lg, c.v2store, c.members[id])
+		m := *(c.members[id])
+		m.RaftAttributes = raftAttr
+		mustUpdateMemberInStore(c.lg, c.v2store, &m)
 	}
 	if c.be != nil && shouldApplyV3 {
+		c.members[id].RaftAttributes = raftAttr
 		c.be.MustSaveMemberToBackend(c.members[id])
-	}
 
-	c.lg.Info(
-		"updated member",
-		zap.String("cluster-id", c.cid.String()),
-		zap.String("local-member-id", c.localID.String()),
-		zap.String("updated-remote-peer-id", id.String()),
-		zap.Strings("updated-remote-peer-urls", raftAttr.PeerURLs),
-		zap.Bool("updated-remote-peer-is-learner", raftAttr.IsLearner),
-	)
+		c.lg.Info(
+			"updated member",
+			zap.String("cluster-id", c.cid.String()),
+			zap.String("local-member-id", c.localID.String()),
+			zap.String("updated-remote-peer-id", id.String()),
+			zap.Strings("updated-remote-peer-urls", raftAttr.PeerURLs),
+			zap.Bool("updated-remote-peer-is-learner", raftAttr.IsLearner),
+		)
+	} else {
+		c.lg.Info(
+			"ignored already updated member",
+			zap.String("cluster-id", c.cid.String()),
+			zap.String("local-member-id", c.localID.String()),
+			zap.String("updated-remote-peer-id", id.String()),
+			zap.Strings("updated-remote-peer-urls", raftAttr.PeerURLs),
+			zap.Bool("updated-remote-peer-is-learner", raftAttr.IsLearner),
+		)
+	}
 }
 
 func (c *RaftCluster) Version() *semver.Version {
--- a/tests/e2e/ctl_v3_member_test.go
+++ b/tests/e2e/ctl_v3_member_test.go
@@ -15,12 +15,15 @@
 package e2e
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"reflect"
 	"strings"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
@@ -53,6 +56,85 @@ func TestCtlV3MemberUpdateClientAutoTLS(t *testing.T) {
 }
 func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
 	testCtl(t, memberUpdateTest, withCfg(*e2e.NewConfigPeerTLS()))
+}
+
+func TestCtlV3ConsistentMemberList(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t,
+		e2e.WithClusterSize(1),
+		e2e.WithWaitClusterReadyTimeout(0),
+		e2e.WithEnvVars(map[string]string{"GOFAIL_FAILPOINTS": `beforeApplyOneConfChange=sleep("2s")`}),
+	)
+	require.NoError(t, err, "failed to start etcd cluster: %v", err)
+	defer func() {
+		derr := epc.Close()
+		require.NoError(t, derr, "failed to close etcd cluster: %v", derr)
+	}()
+
+	t.Log("Adding and then removing a learner")
+	resp, err := epc.Etcdctl().MemberAddAsLearner(context.TODO(), "newLearner", []string{fmt.Sprintf("http://localhost:%d", e2e.EtcdProcessBasePort+11)})
+	require.NoError(t, err)
+	_, err = epc.Etcdctl().MemberRemove(context.TODO(), resp.Member.ID)
+	require.NoError(t, err)
+	t.Logf("Added and then removed a learner with ID: %x", resp.Member.ID)
+
+	t.Log("Restarting the etcd process to ensure all data is persisted")
+	err = epc.Procs[0].Restart(context.TODO())
+	require.NoError(t, err)
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	stopc := make(chan struct{}, 2)
+
+	t.Log("Starting a goroutine to repeatedly restart etcdserver")
+	go func() {
+		defer func() {
+			stopc <- struct{}{}
+			wg.Done()
+		}()
+		for i := 0; i < 3; i++ {
+			select {
+			case <-stopc:
+				return
+			default:
+			}
+
+			merr := epc.Procs[0].Restart(context.TODO())
+			require.NoError(t, merr)
+			epc.WaitLeader(t)
+
+			time.Sleep(100 * time.Millisecond)
+		}
+	}()
+
+	t.Log("Starting a goroutine to repeated check the member list")
+	count := 0
+	go func() {
+		defer func() {
+			stopc <- struct{}{}
+			wg.Done()
+		}()
+
+		for {
+			select {
+			case <-stopc:
+				return
+			default:
+			}
+
+			mresp, merr := epc.Etcdctl().MemberList(context.TODO(), true)
+			if merr != nil {
+				continue
+			}
+
+			count++
+			require.Equal(t, 1, len(mresp.Members))
+		}
+	}()
+
+	wg.Wait()
+	t.Logf("Checked the member list %d times", count)
 }
 
 func memberListTest(cx ctlCtx) {
--- a/tests/e2e/ctl_v3_member_test.go
+++ b/tests/e2e/ctl_v3_member_test.go
@@ -25,6 +25,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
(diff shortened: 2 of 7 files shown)


server/etcdserver/api/membership/cluster.go
```diff
@@ -484,42 +499,68 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
 
 func (c *RaftCluster) UpdateRaftAttributes(id types.ID, raftAttr RaftAttributes, shouldApplyV3 ShouldApplyV3) {
 	c.Lock()
 	defer c.Unlock()
 
-	c.members[id].RaftAttributes = raftAttr
+	if _, ok := c.members[id]; ok {
+		// TODO: move this line into the `if c.be != nil && shouldApplyV3 {`
+		// branch when v2store is removed.
+		c.members[id].RaftAttributes = raftAttr
```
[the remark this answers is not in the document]
ahrtr: Because the member data needs to be saved to both v2store and v3store, so we have to update the member info beforehand, so that both v2 and v3 save the latest data.

Once the v2store is removed, we can move it into the `if c.be != nil && shouldApplyV3 {` branch (line 544).

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +58,81 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
 	testCtl(t, memberUpdateTest, withCfg(*e2e.NewConfigPeerTLS()))
 }
 
+func TestCtlV3ConsistentMemberList(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t,
+		e2e.WithClusterSize(1),
+		e2e.WithWaitClusterReadyTimeout(0),
```
serathius, line 66: I think that setting cluster ready timeout to zero will make the issue appear even after switching to v3. Reason is that during bootstrap it is expected that both storages are stale. It's effects are more visible for v2 as it can be even 10`000 entries behind, but v3 store might also be couple of seconds behind after the last committed change. 

I expect if you added additional writes interleaved between member changes you will notice the same problem. As so this test is incorrect.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
serathius: Ok, I can understand that `mustUpdateMemberInStore` and `MustSaveMemberToBackend` requires a Member struct to be passed, however if you are switching apply to v3 storage, you should follow the pattern in other functions not update the `c.members[id]` if should shouldApplyV3 is false. You might need to copy Member data from c.members[id].

serathius:
Aside of those two comments, I think the change is correct because it doesn't touch validation.

tests/e2e/ctl_v3_member_test.go
ahrtr: > I think that setting cluster ready timeout to zero will make the issue appear even after switching to v3. Reason is that during bootstrap it is expected that both storages are stale. It's effects are more visible for v2 as it can be even 10`000 entries behind, but v3 store might also be couple of seconds behind after the last committed change.


Suggest to read https://github.com/etcd-io/etcd/pull/16084#discussion_r1307902400 again. v3store is always the source of truth, it will never be stale by itself. When we say a member's data is stale, it means it's data is behind of the leader's data; in that case, it can't serve lineariziable request.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
ahrtr: > You might need to copy Member data from c.members[id].

Yes, this is a valid point at the moment. Since we will remove v2store pretty soon, so I just leave a message here. If you  insist on it (I agree with your principle "`each PR should complete on it's own and assume that it could be the last one`." ), I will make the change.

tests/e2e/ctl_v3_member_test.go
serathius: >  it will never be stale by itself.
Oh, but it is. It's stale compared to the last WAL write. Imagine a situation where there was a lot of changes (including membership), all were marked as committed and applied to v3 store, however member crashed before flushing the data from disk. So when it starts back, even if it bootstraps from v3 store (much shorter number of entries that need to be replayed), it still needs to replay all the changes since last disk flush to v3store. 

I think it might be hard to reproduce as in ext4, fsync not only flushes the selected file descriptor, but all file descriptors https://lwn.net/Articles/842385/. I haven't look at this, but it's possible that WAL fsync is also fsyncing etcd db file.

server/etcdserver/api/membership/cluster.go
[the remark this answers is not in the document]
ahrtr: Just updated since it's just minor effort. Please take a look at the 4th commit https://github.com/etcd-io/etcd/pull/16658/commits/353b10f5962762f18f0dc902972714165f32b921

tests/e2e/ctl_v3_member_test.go
ahrtr: > all were marked as committed and applied to v3 store, however member crashed before flushing the data from disk

If the etcd crashes during flushing the data to disk, then the data isn't really applied. 

The backend commits the bbolt db periodically (100ms if I remember it correctly), if etcd crashes anytime in between, it will lose some "already applied" (note it's only in **memory**) entries, when etcd restars, it will replay the WAL files. It's a normal use case, and it's by design.  I do not see any issue.

serathius: > it will replay the WAL files. It's a normal use case, and it's by design. I do not see any issue.

Yes, it's not an issue. But I think it will be picked up by this test making it flaky.

ahrtr: Note I intentionally send a serializable MemberList request. (line 122 in the clt_v3_member_test.go). But it's just a one-member cluster, so the client shouldn't get stale data even it's serializable request.  So it's a bug.

ahrtr: > The test will pick up the lost "already applied" entries, making it flaky.

Still do not got your point how the test is flaky. Pls let me know if you have any confusion or need any clarification.

serathius: Ok, please clarify. If backend commits the bbolt db periodically (as you said above), what happens in this test if etcd crashes before it flushes bbolt. To clarify, before the crash etcd memory will include a new learner, after the crash etcd reads bbolt that doesn't include new learner. On start etcd will load bbolt that doesn't include learner, and will replay WAL entries to add the learner back.

In this test (with cluster ready timeout set to 0) does etcd serve the serializable membership information before it has replayed the WAL entries?

ahrtr: OK, eventually got your point, and make some sense to me. If etcd crashes during a test, I would say majority of the existing e2e test cases are flaky:).

Anyway, updated, please take a look at the last commit https://github.com/etcd-io/etcd/pull/16658/commits/7590dfe56a4144dd2d87389018a018a9c158898a

serathius: By crashing I didn't mean that test will flake due to random crash, I mean the goroutine "Starting a goroutine to repeatedly restart etcdserver" on lines 90-109.

ahrtr: pls see line 82 ~ 84. The membership change is guaranteed to be persisted.

serathius: Sure, but there is a difference between persisting WAL and persisting DB. Are you saying that you are expecting a graceful shutdown via SIGTERM (send by Restart) to result in DB being persisted?

Additional note, I'm having problems with reproducing the issue using the provided test. Will need to double check.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +58,85 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
 	testCtl(t, memberUpdateTest, withCfg(*e2e.NewConfigPeerTLS()))
 }
 
+func TestCtlV3ConsistentMemberList(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	epc, err := e2e.NewEtcdProcessCluster(context.TODO(), t,
```
serathius, line 64: Please create a shared `context.Background()` for all requests. There is no need to add more `context.TODO()`

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +58,85 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
...
+		}()
+
+		for {
+			select {
+			case <-stopc:
+				return
+			default:
+			}
+
+			mresp, merr := epc.Etcdctl().MemberList(context.TODO(), true)
```
serathius, line 126: Recommend using `clientv3` library as it's much faster. It should allow much higher chance of reproducing the issue.

tests/e2e/ctl_v3_member_test.go
serathius: This test, doesn't reproduce the issue for me. Any idea why it doesn't find the issue on my side?

ahrtr: > Are you saying that you are expecting a graceful shutdown via SIGTERM (send by Restart) to result in DB being persisted?

YES



> This test, doesn't reproduce the issue for me. Any idea why it doesn't find the issue on my side?

You need to remove the second commit and the third commit, and then run the test,

```
$ git reset --hard HEAD~2
$ make gofail-enable
$ cd tests/e2e/
$ go test -run TestCtlV3ConsistentMemberList -v
```

tests/e2e/ctl_v3_member_test.go
ahrtr: I am afraid I do not agree this.  An e2e test is supposed to use etcdctl to communicate with etcdserver. Pls follow https://github.com/etcd-io/etcd/pull/16658#discussion_r1341226598 to reproduce this issue locally.

tests/e2e/ctl_v3_member_test.go
ahrtr: done.

tests/e2e/ctl_v3_member_test.go
serathius: ok, it was obvious. I didn't run `make gofail-enable`. Not a fun of fact that the test doesn't do anything if we forget `make gofail-enable`. If we are adding requirement for e2e tests to use a binary with failpoints, we should validate this (or skip). Recommend this as a followup.

ahrtr: > Not a fun of fact that the test doesn't do anything if we forget `make gofail-enable`. If we are adding requirement for e2e tests to use a binary with failpoints, we should validate this (or skip)

Note that the failpoint is enabled statically instead of dynamically. To be clearer, I do not use `e2e.WithGoFailEnabled(true)`.  It will never be skipped in github workflow. But of course, developer might forget `make gofail-enable` when executing it locally. Added a comment.

tests/e2e/ctl_v3_member_test.go
serathius: No, when testing etcd standard features like Get request, e2e tests should test `etcdctl get`, however using `etcdctl` is not a rule for all tests. There is no benefit of using `etcdctl` for e2e tests, specifically for correctness testing.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +58,85 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
...
+			if merr != nil {
+				continue
+			}
+
+			count++
```
serathius, line 137: Please provide some minimal sensible value of count and assert. For example add `assert.Greater(t, count, 100)` after wg.Wait()

ahrtr: The two goroutine executes independently. You never know how many times the second goroutine have executed before the first one restarts the etcdserver in the first time. Adding the suggested assertion will make this test flaky.

It's for sure that the issue can be reproduce during the 3 restarts.

serathius: The problem is that there is no guarantee that `Etcdctl` will be executed at all and the test still will return success.
We need some sensible minimum, like 5, or 10, but be should not allow 0 runs.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +59,92 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
 	testCtl(t, memberUpdateTest, withCfg(*e2e.NewConfigPeerTLS()))
 }
 
+// TestCtlV3ConsistentMemberList requires the gofailpoint to be enabled.
+// If you execute this case locally, please do not forget to execute
+// `make gofail-enable`.
+func TestCtlV3ConsistentMemberList(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
serathius, line 68: I get timeout on 20 seconds. Can we just skip this timeout?

ahrtr: resolved & thx

serathius:
Managed to reproduce the issue on v3 storage, however I needed to SIGKILL the process instead of SIGTERM. Will double check if it reproduces on SIGTERM.

```
=== RUN   TestCtlV3ConsistentMemberList
    before.go:36: Changing working directory to: /tmp/TestCtlV3ConsistentMemberList1278668904/001
    logger.go:130: 2023-09-29T14:03:38.972+0200 INFO    starting server...      {"name": "TestCtlV3ConsistentMemberList-test-0"}
    logger.go:130: 2023-09-29T14:03:38.972+0200 INFO    spawning process        {"args": ["/usr/local/google/home/siarkowicz/src/go.etcd.io/etcd/bin/etcd", "--name", "TestCtlV3ConsistentMemberList-test-0", "--listen-client-urls", "http://localhost:20000", "--advertise-client-urls", "http://localhost:20000", "--listen-peer-urls", "http://localhost:20001", "--initial-advertise-peer-urls", "http://localhost:20001", "--initial-cluster-token", "new", "--data-dir", "/tmp/TestCtlV3ConsistentMemberList1278668904/002", "--snapshot-count", "10000", "--initial-cluster", "TestCtlV3ConsistentMemberList-test-0=http://localhost:20001", "--initial-cluster-state", "new"], "working-dir": "/tmp/TestCtlV3ConsistentMemberList1278668904/001", "name": "TestCtlV3ConsistentMemberList-test-0", "environment-variables": ["GOFAIL_FAILPOINTS=beforeApplyOneConfChange=sleep(\"2s\")", "PATH=/usr/local/google/home/siarkowicz/.gvm/gos/go1.21.1/bin:/usr/local/google/home/siarkowicz/.gvm/gos/go1.21.1/bin:/usr/local/google/home/siarkowicz/bin:/usr/local/google/home/siarkowicz/bin/:/usr/local/google/home/siarkowicz/.gvm/pkgsets/go1.20.7/global/bin:/usr/local/google/home/siarkowicz/.gvm/gos/go1.20.7/bin:/usr/local/google/home/siarkowicz/.gvm/pkgsets/go1.20.7/global/overlay/bin:/usr/local/google/home/siarkowicz/.gvm/bin:/usr/local/google/home/siarkowicz/.gvm/bin:/usr/local/google/home/siarkowicz/bin:/usr/lib/google-golang/bin:/usr/local/buildtools/java/jdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/google/home/siarkowicz/.local/bin:/usr/local/google/home/siarkowicz/bin", "ETCD_UNSUPPORTED_ARCH=amd64", "ETCD_VERIFY=all"]}
    logger.go:130: 2023-09-29T14:03:41.722+0200 INFO    started server. {"name": "TestCtlV3ConsistentMemberList-test-0", "pid": 3151124}
    ctl_v3_member_test.go:90: Adding and then removing a learner
    ctl_v3_member_test.go:102: Added and then removed a learner with ID: 2ddd8d996718531c
    ctl_v3_member_test.go:104: Restarting the etcd process to ensure all data is persisted
    ctl_v3_member_test.go:109: Starting a goroutine to repeatedly restart etcdserver
    ctl_v3_member_test.go:132: Starting a goroutine to repeated check the member list
    logger.go:130: 2023-09-29T14:03:45.759+0200 INFO    killing server...       {"name": "TestCtlV3ConsistentMemberList-test-0"}
    logger.go:130: 2023-09-29T14:03:45.761+0200 INFO    server exited   {"name": "TestCtlV3ConsistentMemberList-test-0", "code": 137}
    logger.go:130: 2023-09-29T14:03:45.761+0200 INFO    starting server...      {"name": "TestCtlV3ConsistentMemberList-test-0"}
    logger.go:130: 2023-09-29T14:03:45.761+0200 INFO    spawning process        {"args": ["/usr/local/google/home/siarkowicz/src/go.etcd.io/etcd/bin/etcd", "--name", "TestCtlV3ConsistentMemberList-test-0", "--listen-client-urls", "http://localhost:20000", "--advertise-client-urls", "http://localhost:20000", "--listen-peer-urls", "http://localhost:20001", "--initial-advertise-peer-urls", "http://localhost:20001", "--initial-cluster-token", "new", "--data-dir", "/tmp/TestCtlV3ConsistentMemberList1278668904/002", "--snapshot-count", "10000", "--initial-cluster", "TestCtlV3ConsistentMemberList-test-0=http://localhost:20001", "--initial-cluster-state", "new"], "working-dir": "/tmp/TestCtlV3ConsistentMemberList1278668904/001", "name": "TestCtlV3ConsistentMemberList-test-0", "environment-variables": ["GOFAIL_FAILPOINTS=beforeApplyOneConfChange=sleep(\"2s\")", "PATH=/usr/local/google/home/siarkowicz/.gvm/gos/go1.21.1/bin:/usr/local/google/home/siarkowicz/.gvm/gos/go1.21.1/bin:/usr/local/google/home/siarkowicz/bin:/usr/local/google/home/siarkowicz/bin/:/usr/local/google/home/siarkowicz/.gvm/pkgsets/go1.20.7/global/bin:/usr/local/google/home/siarkowicz/.gvm/gos/go1.20.7/bin:/usr/local/google/home/siarkowicz/.gvm/pkgsets/go1.20.7/global/overlay/bin:/usr/local/google/home/siarkowicz/.gvm/bin:/usr/local/google/home/siarkowicz/.gvm/bin:/usr/local/google/home/siarkowicz/bin:/usr/lib/google-golang/bin:/usr/local/buildtools/java/jdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/google/home/siarkowicz/.local/bin:/usr/local/google/home/siarkowicz/bin", "ETCD_UNSUPPORTED_ARCH=amd64", "ETCD_VERIFY=all"]}
    logger.go:130: 2023-09-29T14:03:50.819+0200 INFO    started server. {"name": "TestCtlV3ConsistentMemberList-test-0", "pid": 3151165}
    ctl_v3_member_test.go:155: 
                Error Trace:    /usr/local/google/home/siarkowicz/src/go.etcd.io/etcd/tests/e2e/ctl_v3_member_test.go:155
                                                        /usr/local/google/home/siarkowicz/.gvm/gos/go1.21.1/src/runtime/asm_amd64.s:1650
                Error:          Not equal: 
                                expected: 1
                                actual  : 2
                Test:           TestCtlV3ConsistentMemberList
    cluster.go:1059: members agree on a leader, members:map[14578408409545168728:0] , leader:14578408409545168728
    ctl_v3_member_test.go:161: Checked the member list 1 times
    logger.go:130: 2023-09-29T14:03:51.301+0200 INFO    closing test cluster...
    logger.go:130: 2023-09-29T14:03:51.301+0200 INFO    stopping server...      {"name": "TestCtlV3ConsistentMemberList-test-0"}
    logger.go:130: 2023-09-29T14:03:53.306+0200 INFO    stopped server. {"name": "TestCtlV3ConsistentMemberList-test-0"}
    logger.go:130: 2023-09-29T14:03:53.307+0200 INFO    closing server...       {"name": "TestCtlV3ConsistentMemberList-test-0"}
    logger.go:130: 2023-09-29T14:03:53.307+0200 INFO    removing directory      {"data-dir": "/tmp/TestCtlV3ConsistentMemberList1278668904/002"}
    logger.go:130: 2023-09-29T14:03:53.309+0200 INFO    closed test cluster.
--- FAIL: TestCtlV3ConsistentMemberList (14.34s)
FAIL
FAIL    go.etcd.io/etcd/tests/v3/e2e    95.326s
FAIL
tests $ ./bin/etcd --version
bash: ./bin/etcd: No such file or directory
tests $ ../bin/etcd --version
etcd Version: 3.6.0-alpha.0
Git SHA: 70a320550
Go Version: go1.21.1
Go OS/Arch: linux/amd64
```

serathius (approved):
Still have mixed feelings about the test. It depends on behavior that is not guaranteed even after the fix. Repro on SIGKILL could mean that it's dependant on race between shutdown and flushing bbolt. 

The good thing is that the Test is written in a way that minimizes the risk. So I think it's better to have a regression test, then not have one.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +59,91 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
 	testCtl(t, memberUpdateTest, withCfg(*e2e.NewConfigPeerTLS()))
 }
 
+// TestCtlV3ConsistentMemberList requires the gofailpoint to be enabled.
+// If you execute this case locally, please do not forget to execute
+// `make gofail-enable`.
```
ahrtr, line 64: From high level propestive, gracefully shutting down means letting the process to do some cleanup tasks,  including syncing the data. I think it's simple to understand.

From implementation perspective, it needs deep understanding on the repo. It's not about one line of code, it's about a complicated workflow.

For anyone reference: 
- When the etcd receives SIGTERM, it will trigger (*Etcd)Close https://github.com/etcd-io/etcd/blob/0a3dc1a8a8b6d06368ad13f9a8f13c038e7ca362/server/embed/etcd.go#L388

- Inside the Etcd.Close, it will execute EtcdServer.HardStop https://github.com/etcd-io/etcd/blob/0a3dc1a8a8b6d06368ad13f9a8f13c038e7ca362/server/etcdserver/server.go#L1244

- The main goroutine will exit at https://github.com/etcd-io/etcd/blob/0a3dc1a8a8b6d06368ad13f9a8f13c038e7ca362/server/etcdserver/server.go#L854

- Then the defered function will be executed, inside which s.Cleanup will be executed. https://github.com/etcd-io/etcd/blob/0a3dc1a8a8b6d06368ad13f9a8f13c038e7ca362/server/etcdserver/server.go#L829

- Inside the cleanup, backend.Close is executed. https://github.com/etcd-io/etcd/blob/0a3dc1a8a8b6d06368ad13f9a8f13c038e7ca362/server/etcdserver/server.go#L910

- Eventually it will execute bbolt.Commit https://github.com/etcd-io/etcd/blob/0a3dc1a8a8b6d06368ad13f9a8f13c038e7ca362/server/storage/backend/backend.go#L421

tests/e2e/ctl_v3_member_test.go
```diff
@@ -55,6 +59,91 @@ func TestCtlV3MemberUpdatePeerTLS(t *testing.T) {
...
+func TestCtlV3ConsistentMemberList(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	ctx := context.Background()
+
+	epc, err := e2e.NewEtcdProcessCluster(ctx, t,
+		e2e.WithClusterSize(1),
+		e2e.WithWaitClusterReadyTimeout(0),
```
serathius, line 72: Oh, don't know how I missed this. Setting `0` here means that flag is not passed by the e2e framework. The minimum you can set is `1`, which would be `1ns`.
cc @ahrtr

ahrtr: thx & fixed in https://github.com/etcd-io/etcd/pull/16671

The reason why I could still reproduce the issue is because there are 3 CONF changes, which will take 2 * 3 = 6s. The default wait-cluster-ready-timeout is 5s, so it's still easy to reproduce the issue.

PR #16661, opened 2023-09-28
all: use built-in min/max

gocurr:
Since go.mod is 1.21, we can simplify code by using these built-in functions.

### DIFF
--- a/pkg/report/timeseries.go
+++ b/pkg/report/timeseries.go
@@ -64,9 +64,9 @@ func (sp *secondPoints) Add(ts time.Time, lat time.Duration) {
 		sp.tm[tk] = secondPoint{minLatency: lat, maxLatency: lat, totalLatency: lat, count: 1}
 	} else {
 		if lat != time.Duration(0) {
-			v.minLatency = minDuration(v.minLatency, lat)
+			v.minLatency = min(v.minLatency, lat)
 		}
-		v.maxLatency = maxDuration(v.maxLatency, lat)
+		v.maxLatency = max(v.maxLatency, lat)
 		v.totalLatency += lat
 		v.count++
 		sp.tm[tk] = v
@@ -143,18 +143,4 @@ func (t TimeSeries) String() string {
 		log.Fatal(err)
 	}
 	return fmt.Sprintf("\nSample in one second (unix latency throughput):\n%s", buf.String())
-}
-
-func minDuration(a, b time.Duration) time.Duration {
-	if a < b {
-		return a
-	}
-	return b
-}
-
-func maxDuration(a, b time.Duration) time.Duration {
-	if a > b {
-		return a
-	}
-	return b
 }
--- a/tools/benchmark/cmd/put.go
+++ b/tools/benchmark/cmd/put.go
@@ -152,13 +152,6 @@ func compactKV(clients []*v3.Client) {
 	}
 }
 
-func max(n1, n2 int64) int64 {
-	if n1 > n2 {
-		return n1
-	}
-	return n2
-}
-
 func hashKV(cmd *cobra.Command, clients []*v3.Client) {
 	eps, err := cmd.Flags().GetStringSlice("endpoints")
 	if err != nil {


gocurr:
https://github.com/golangci/golangci-lint/blob/dafe1469eb4b35ca04af0b9725bed7d69f3c5918/go.mod#L3
The lint error is probably related to this.
The main branch uses Go 1.21, however, golangci-lint uses 1.20.
Should this PR be abandoned?

jmhbnz:
> https://github.com/golangci/golangci-lint/blob/dafe1469eb4b35ca04af0b9725bed7d69f3c5918/go.mod#L3 The lint error is probably related to this. The main branch uses Go 1.21, however, golangci-lint uses 1.20. Should this PR be abandoned?

Looks like golangci-lint supports go 1.21 fully from version [v1.54.1](https://github.com/golangci/golangci-lint/releases/tag/v1.54.1), refer https://github.com/golangci/golangci-lint/issues/3933.

We need to update our golangci-lint version here https://github.com/etcd-io/etcd/blob/main/.github/workflows/static-analysis.yaml#L18. 

Can you please add a second commit to this pr @gocurr  to bump that?  I believe it will resolve the issue you're seeing and allow us to merge this.

jmhbnz (approved):
LGTM - Thanks for cleaning this up @gocurr 👍🏻

ahrtr:
Users won't be able to build etcd (main) using golang 1.20.x any more if we merge this PR. Although we suggest to build the main branch using go 1.21, but technically we should prevent users from building using go 1.20.x.

So suggest on hold this PR until go 1.20.x is out of support.

.github/workflows/static-analysis.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
-          version: v1.53.3
+          version: v1.54.1
```
ahrtr, line 18: This can be merged separately.

ahrtr (changes requested):
Explicitly mark this as request changes. 

https://github.com/etcd-io/etcd/pull/16661#issuecomment-1740515027

.github/workflows/static-analysis.yaml
gocurr: I think it should be kept to prevent new contributors
from using the new features of 1.21.

serathius:
> Users won't be able to build etcd (main) using golang 1.20.x any more if we merge this PR. Although we suggest to build the main branch using go 1.21, technically we should NOT prevent users from building main using go 1.20.x for now.

Not sure I understand the motivation. We have updated the go version in go.mod to 1.21. This is a clear suggestion that libraries we publish are go1.21 compatible. As for 1.20 backward compatibility, I would agree it's important, however only for `release-3.4` and `release-3.5` branches as we promise backward compatibility. 

There is no reason to require go1.20 compatibility from `main` branch.

ahrtr:
> This is a clear suggestion that libraries we publish are go1.21 compatible.

Suggestion != mandatory

It forces the applications/users, which depend on/use etcd, to bump their golang version to 1.21. 

Overall it isn't a big deal, since 1.20 will be out of support about 4~5 months later. But let's try not to bring bad user experience.

jmhbnz:
For `main` I don't think we should be creating expectation that no new golang features will be adopted until the previous golang release is out of support.

It's a development branch with no guarantee around stability.

If there are exciting new features in golang these might come with a degree of complexity and sooner we can get those features into our CI system and start observing their behavior over time the better.

This pr is a trivial example in terms of new feature, but I'm glad it's forced this discussion.

I acknowledge that updating golang minor version locally can be annoying, but as developers working on a development branch I think it's to be expected from time to time.

serathius:
> > This is a clear suggestion that libraries we publish are go1.21 compatible.
> 
> Suggestion != mandatory

True, it's a suggestion because not every golang version introduces backward compatible changes which would make it mandatory.

serathius:
Asked @liggitt, he said that Kubernetes doesn't even do this. If you want to propose this, then we should discuss this.

serathius:
Still, I might agree that just the `min`, `max` function are not worth of breaking compatibility.

liggitt:
If you need to keep compatibility with older go versions, a possibility is to put the use of new language features in a go version-gated file and have an alternate implementation in a second file. Of course, that often negates the convenience of using the stdlib version in the first place.

ahrtr:
> a possibility is to put the use of new language features in a go version-gated file and have an alternate implementation in a second file.

thx for the suggestion. The change in this PR isn't worth of introducing additional complexity, nor breaking golang compatibility.

gocurr:
Abandoned this for 1.20 compatibility.

Issue #16666, opened 2023-09-29
--experimental-wait-cluster-ready-timeout causing stale response to linearizable read


serathius:
While working and discussing https://github.com/etcd-io/etcd/pull/16658 with @ahrtr I spotted one issue. Etcd going back in time during bootstrap. The good news is that the current reproduction limit the issue to v3.6 release. Impact on older releases is still under investigation.

As described in https://github.com/etcd-io/etcd/pull/16658#discussion_r1341346778 graceful shutdown via SIGTERM allows etcd flushing it's database to disk, however SIGKILL will mean that data on disk might be older than in memory state of etcd. While bootstrapping etcd will catch up on changes it has forgotten by replaying WAL. Etcd might go back in time if it started serving data before it caught up to state before the kill. But this doesn't happen right?

Unfortunately it's possible. In https://github.com/etcd-io/etcd/pull/13525 we have added a flag that skips the wait. Fortunately the change is only present on v3.6. `--experimental-wait-cluster-ready-timeout` goal was to skip the wait for quorum during member bootstrap allowing member to start serving serializable requests. This unfortunately also skips the wait for entries from before crash to be applied. The skip will happen by default, fortunately 5s wait is enough to prevent this happening in most cases.

After looking deeper I found more issues.
* [confirmed] --experimental-wait-cluster-ready-timeout causing member serializable responses to go back in time. Repro https://github.com/etcd-io/etcd/commit/51fb2d78ddd35c5aa08bd51466ecfebb73601293 It's done by reusing `TestCtlV3ConsistentMemberList` test introduced in https://github.com/etcd-io/etcd/pull/16658 and uses SIGKILL.
* [confirmed] --experimental-wait-cluster-ready-timeout breaking linearizability in single node cluster. In single node cluster, leader will not check readIndex the same way. It will trust it's local committed index, which is not guaranteed to be persisted before responding to user. Repro https://github.com/etcd-io/etcd/pull/16672
* Even without --experimental-wait-cluster-ready-timeout, clusterReady mechanism seems strange. It just waits for a joining member to be able to make a proposal (set it's own member attributes), however this has noting to do with replaying entries. I expect that with proper timings (injecting sleep into code) the etcd will start serving before all entries from before crash have been applied. I expect it can happen on both single and multi node cluster.

serathius:
cc @gyuho @jpbetz @philips @hexfusion @bdarnell

Would be great if to get your opinion.

ahrtr:
[...]

PR #16672, opened 2023-09-30
Reproduce --wait-cluster-ready-timeout flag causing linearizability issue

serathius:
Reproduces https://github.com/etcd-io/etcd/issues/16666 but finds that --wait-cluster-ready-timeout could even cause   linearizability issue in single node cluster.


serathius:
Selection_006

ahrtr:
Please integrate https://github.com/etcd-io/raft/pull/105 and try again.

serathius:
Will do,  however I think the https://github.com/etcd-io/raft/pull/105 is not a proper solution. I expect it will fix the issue, but I don't think the approach is correct.

PR #16678, opened 2023-10-02
test: add test cases to verify consistent reading right after writing

ahrtr:
It's super easy to reproduce the issue https://github.com/etcd-io/etcd/issues/16666

```
    ctl_v3_kv_test.go:102: 
        	Error Trace:	/Users/wachao/go/src/github.com/ahrtr/etcd/tests/e2e/ctl_v3_kv_test.go:102
        	            				/Users/wachao/software/go/src/runtime/asm_arm64.s:1197
        	Error:      	Not equal: 
        	            	expected: "bar2"
        	            	actual  : "bar"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-bar2
        	            	+bar
        	Test:       	TestCtlV3LinearizableRead
    ctl_v3_kv_test.go:115: Checked the key/value 1 times
```

The issue can be resolved by either of the following approaches,
- rollback https://github.com/etcd-io/etcd/pull/13525
- Remove `e2e.WithWaitClusterReadyTimeout(1*time.Nanosecond)` in the test in this PR. [It's still possible to reproduce this issue by injecting extremely bigger time sleep]

### DIFF
--- a/tests/e2e/corrupt_test.go
+++ b/tests/e2e/corrupt_test.go
@@ -364,3 +364,95 @@ func TestCompactHashCheckDetectCorruptionInterrupt(t *testing.T) {
 	}
 	t.Log("no corruption detected.")
 }
+
+func TestCtlV3SerializableRead(t *testing.T) {
+	testCtlV3ReadAfterWrite(t, clientv3.WithSerializable())
+}
+
+func TestCtlV3LinearizableRead(t *testing.T) {
+	testCtlV3ReadAfterWrite(t)
+}
+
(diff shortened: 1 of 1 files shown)


serathius:
I think the test doesn't work with cluster_proxy. Do you want to fix it or can add those tests to https://github.com/etcd-io/etcd/issues/16690

ahrtr:
The proxy misses some feature, such as the `Kill`, `Wait`, etc in `etcd_process.go` are not implemented at all in `cluster_proxy.go`. 

Just moved the test cases into corrupt_test.go. Let's revisit how to handle grpcProxy test later.

Issue #16693, opened 2023-10-05
Etcd revision decreased by 100, while the data stays the same


serathius:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [x] This is not a support request, support requests should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Robustness test failed https://github.com/etcd-io/etcd/actions/runs/6404137994/job/17384078003

### What did you expect to happen?

No failure

### How can we reproduce it (as minimally and precisely as possible)?

Commit hash: 6a0d2eb5ce8ce900b9f3be77662f5e5b1fb5254e
Environment: main-arm64
Failing test: TestRobustness/Kubernetes/HighTraffic/ClusterOfSize1
Failpoint: compactBeforeSetFinishedCompact

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```Shell
2023-10-04T09:46:24.8949231Z     watch.go:67: Broke watch guarantee: Ordered - events are ordered by revision; an event will never appear on a watch if it precedes an event in time that has already been posted, lastRevision: 251, currentRevision: 192, client: 9
2023-10-04T09:46:24.8952913Z     watch.go:91: Broke watch guarantee: Unique - an event will never appear on a watch twice, key: "/registry/pods/default/SfdQW", revision: 192, client: 9
2023-10-04T09:46:24.8956701Z     watch.go:91: Broke watch guarantee: Unique - an event will never appear on a watch twice, key: "/registry/pods/default/ObHFs", revision: 193, client: 9
[...]

PR #16695, opened 2023-10-05
Fix watch validation assuming that client requesting older watch revision

serathius:
cc @ahrtr 

Part of https://github.com/etcd-io/etcd/issues/16693

### DIFF
--- a/tests/robustness/validate/watch.go
+++ b/tests/robustness/validate/watch.go
@@ -43,8 +43,8 @@ func validateWatch(t *testing.T, cfg Config, reports []report.ClientReport) []mo
 }
 
 func validateBookmarkable(t *testing.T, report report.ClientReport) {
-	var lastProgressNotifyRevision int64
 	for _, op := range report.Watch {
+		var lastProgressNotifyRevision int64
 		for _, resp := range op.Responses {
 			for _, event := range resp.Events {
 				if event.Revision <= lastProgressNotifyRevision {
@@ -59,8 +59,8 @@ func validateBookmarkable(t *testing.T, report report.ClientReport) {
 }
 
 func validateOrdered(t *testing.T, report report.ClientReport) {
-	var lastEventRevision int64 = 1
 	for _, op := range report.Watch {
+		var lastEventRevision int64 = 1
 		for _, resp := range op.Responses {
 			for _, event := range resp.Events {
 				if event.Revision < lastEventRevision {
@@ -73,9 +73,8 @@ func validateOrdered(t *testing.T, report report.ClientReport) {
 }
 
 func validateUnique(t *testing.T, expectUniqueRevision bool, report report.ClientReport) {
-	uniqueOperations := map[any]struct{}{}
-
 	for _, op := range report.Watch {
+		uniqueOperations := map[any]struct{}{}
 		for _, resp := range op.Responses {
 			for _, event := range resp.Events {
 				var key any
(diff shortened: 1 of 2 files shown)


tests/robustness/validate/watch.go
```diff
@@ -59,8 +59,8 @@ func validateBookmarkable(t *testing.T, report report.ClientReport) {
 }
 
 func validateOrdered(t *testing.T, report report.ClientReport) {
-	var lastEventRevision int64 = 1
 	for _, op := range report.Watch {
+		var lastEventRevision int64 = 1
```
ahrtr, line 63: Not sure why change this. Aren't watch responses globally ordered for each client? In your test case `TestValidateWatch`, there is no clientId; do you intentionally verify that different clients may request an older revision?

serathius: True, we expect client to get increasing revisions due to either fact that watch doesn't break or that users will usually want to reestablish watches on the following revision. However, from etcd perspective, those are independent watch request each providing its own revision to start watching from. It's not invalid that single client can start watching from rev 200, and after that decide to establish new watch from rev 100. 

As in case https://github.com/etcd-io/etcd/issues/16693, for some unknown reason etcd went back from revision 301 to 192 in the KV store. So from clients perspective it behaved correctly, after watch was broken it established the new watch on revision 192, even though it has previously seen revision 301. 

Goal of this issue is to remove assumption about sensible client behavior (not going back on watch), and just validate the watch responses. This should increase readability of robustness test reports as client misbehavior caused by etcd linearizability issue will no longer also report invalid watch.

ahrtr: > etcd went back from revision 301 to 192 in the KV store. So from clients perspective it behaved correctly, after watch was broken it established the new watch on revision 192, even though it has previously seen revision 301

It's true. But in your test case, there is no watch establishment, so the revision shouldn't go back? 

The test change is OK. But I'd suggest you to have a deep dive to figure out why the revision go back. Let me know if you need my assistance or I misunderstood anything.

ahrtr (approved):
Let the PR go for now.

Please revisit https://github.com/etcd-io/etcd/pull/16695/files#r1347740431 later.

tests/robustness/validate/watch.go
ahrtr: https://github.com/etcd-io/etcd/blob/6a96ab7c09b4173406c8105c28d72d029ab50d4d/tests/robustness/traffic/client.go#L230

serathius: > It's true. But in your test case, there is no watch establishment, so the revision shouldn't go back?

There is, but only in Kubernetes traffic. It runs a ListWatch loop with 100 ms timeout.

Look consists of Read and Watch
https://github.com/etcd-io/etcd/blob/6a96ab7c09b4173406c8105c28d72d029ab50d4d/tests/robustness/traffic/kubernetes.go#L67-L82

And watch breaks every 100ms to simulate client loosing connection
https://github.com/etcd-io/etcd/blob/6a96ab7c09b4173406c8105c28d72d029ab50d4d/tests/robustness/traffic/kubernetes.go#L190-L197

chaochn47:
It is difficult for me to understand the title until fully read the PR comment.

The PR title can be updated to `Fix watch validation assuming that client not requesting older watch revision`. 

Fix should be followed by a problem and the problem is the validation assumes that client won't request watch with older revision.

serathius:
Sorry for that, I rewrote the title couple of times and it turned out not very clear.

chaochn47:
Thanks for updating!

PR #16696, opened 2023-10-05
Fix typo in ci infra admins table

jmhbnz:
As noticed in https://github.com/cncf/cluster/issues/242 we have a typo for a github handle in our infra admins table which caused some confusion so let's fix that.


chaochn47 (approved):

fuweid (approved):
LGTM

serathius (approved):

PR #16697, opened 2023-10-06
http health check bug fixes

chaochn47:
Part of #16007 

Discovered by reviewing https://github.com/etcd-io/etcd/pull/16651

@siyuanfoundation @serathius

### DIFF
--- a/server/etcdserver/api/etcdhttp/health.go
+++ b/server/etcdserver/api/etcdhttp/health.go
@@ -40,24 +40,25 @@ type ServerHealth interface {
 	Leader() types.ID
 	Range(context.Context, *pb.RangeRequest) (*pb.RangeResponse, error)
 	Config() config.ServerConfig
+	AuthStore() auth.AuthStore
 }
 
 // HandleHealth registers metrics and health handlers. it checks health by using v3 range request
 // and its corresponding timeout.
 func HandleHealth(lg *zap.Logger, mux *http.ServeMux, srv ServerHealth) {
-	mux.Handle(PathHealth, NewHealthHandler(lg, func(excludedAlarms AlarmSet, serializable bool) Health {
+	mux.Handle(PathHealth, NewHealthHandler(lg, func(ctx context.Context, excludedAlarms AlarmSet, serializable bool) Health {
 		if h := checkAlarms(lg, srv, excludedAlarms); h.Health != "true" {
 			return h
 		}
 		if h := checkLeader(lg, srv, serializable); h.Health != "true" {
 			return h
 		}
-		return checkAPI(lg, srv, serializable)
+		return checkAPI(ctx, lg, srv, serializable)
 	}))
 }
 
 // NewHealthHandler handles '/health' requests.
-func NewHealthHandler(lg *zap.Logger, hfunc func(excludedAlarms AlarmSet, Serializable bool) Health) http.HandlerFunc {
+func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAlarms AlarmSet, Serializable bool) Health) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodGet {
 			w.Header().Set("Allow", http.MethodGet)
@@ -71,7 +72,7 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(excludedAlarms AlarmSet, Serial
 		// This is useful for probes attempting to validate the liveness of
 		// the etcd process vs readiness of the cluster to serve requests.
 		serializableFlag := getSerializableFlag(r)
-		h := hfunc(excludedAlarms, serializableFlag)
+		h := hfunc(r.Context(), excludedAlarms, serializableFlag)
 		defer func() {
 			if h.Health == "true" {
 				healthSuccess.Inc()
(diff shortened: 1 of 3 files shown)


serathius:
Couple of high level questions

To clarify this PR makes etcd healthcheck work when authorization is enabled. Without it checking health is short circuited by permission denied error and doesn't validate reading db at all. Is this correct?

Do you plan to backport this change?

How have you tested the change? Have you considered testing it manually that everything works? Adding an e2e tests or verifying that this feature is covered by testing would also be appreciated.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -178,13 +180,14 @@ func checkLeader(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
 	return h
 }
 
-func checkAPI(lg *zap.Logger, srv ServerHealth, serializable bool) Health {
+func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializable bool) Health {
 	h := Health{Health: "true"}
 	cfg := srv.Config()
-	ctx, cancel := context.WithTimeout(context.Background(), cfg.ReqTimeout())
-	_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: serializable})
+	ctx = srv.AuthStore().WithRoot(ctx)
```
ahrtr, line 186: There is [another place of using WithRoot](https://github.com/etcd-io/etcd/blob/185ddb2e78a6f856c6151466d57ea6306ff17c57/server/etcdserver/server.go#L874), which makes sense, because it's completely triggered by etcdserver itself. But here, the health check is triggered by users of etcd. 

Shouldn't the credential be configured on the client side? But I don't see how to do it in K8s probes.

@mitake Do you have any concern on this?

serathius: Serializable read needs to pick a key to read, if authorization is enabled there is no easy way to know which keys users has access. Having check accept permission error seems like a big downgrade as it skips most of serializable read code. 

Having an inconsistent quality and responses of check depending on user is very weird. 

I also expect most health checkers will not get admin rights, so with auth enabled this check is pretty useless.

mitake: I think using WithRoot is fine here because the response isn't returned to a user. If a user can supply credential (and key name), just using Range for healthcheck purpose might also be possible? So I think the endpoint shouldn't require these parameters.

ahrtr: > I think using WithRoot is fine here because the response isn't returned to a user

Makes sense, although I do think the users (actually administrators of the control plane) have responsibility to get the credential configured.

chaochn47:
> To clarify this PR makes etcd healthcheck work when authorization is enabled. Without it checking health is short circuited by permission denied error and doesn't validate reading db at all. Is this correct?

Correct. 

> Do you plan to backport this change? Even though the check is not useful, I would prefer to avoid making a change in old releases. Better to ask users to migrate to new /health endpoint then just surprize them with the change.

No plan to backport this change. 

I agree we should ask users to migrate to `/livez` and `/readyz`. It forces them to test with their environment. This change is just for consistency between the serializable read checker in `/health` and `/livez`. 

> How have you tested the change? Have you considered testing it manually that everything works? Adding an e2e tests or verifying that this feature is covered by testing would also be appreciated.

Yeah, test is in progress.. refer #16698

ahrtr:
> Yeah, test is in progress.. refer #16698

My understanding is the test in the PR should all pass on top of this PR. Pls verify that and feedback the result, thx.

ahrtr (approved):

serathius (approved):

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -23,11 +23,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 
+	"go.etcd.io/raft/v3"
```
fuweid, line 26: the goimport check is not reliable. The `go.etcd.io` should be in one group. 

https://github.com/etcd-io/etcd/blob/185ddb2e78a6f856c6151466d57ea6306ff17c57/tools/.golangci.yaml#L29

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -149,10 +140,13 @@ func TestHealthHandler(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			mux := http.NewServeMux()
+			lg := zaptest.NewLogger(t)
+			be, _ := betesting.NewDefaultTmpBackend(t)
```
fuweid, line 144: suggest to add `defer betesting.Close(t, be)`

server/etcdserver/api/etcdhttp/health.go
chaochn47: Thanks! @fuweid 

It always bears me... After etcd and raft splits, Goland does not recognize raft as current project packages... 

FWIW, I have updated the code style perference and it is now align with the `.golangci` configuration!

image

server/etcdserver/api/etcdhttp/health_test.go
chaochn47: Okay. I will apply the suggested change. 

`NewDefaultTmpBackend` should auto close it after test is completed though.

fuweid:
Hi @chaochn47 the change looks good. but it seems that the change has been covered by https://github.com/etcd-io/etcd/pull/16651....

chaochn47:
> Hi @chaochn47 the change looks good. but it seems that the change has been covered by #16651....

Yeah, #16651 `/livez` and `/readyz`is the future. 

This PR is the bug fixes backported to `/health` discovered when reviewing #16651 as the top comment mentions~

fuweid (approved):
LGTM

ahrtr (approved):

Issue #15258, opened 2023-02-07
Robustness tests should validate if etcd binary supports gofail


serathius:
### What would you like to be added?

Etcd robustness tests scenarios that require gofail should validate that etcd binary was complied with gofail enabled and if it's not it should provide a verbose error with instructions how to compile etcd.

### Why is this needed?

It's not obvious for new users that linearizability tests require etcd build with gofail enabled. 
Error is not very verbose: `"connection refused" error from a goPanicFailpoint.`

https://github.com/etcd-io/etcd/pull/15237#issuecomment-1416150241

preetham:
@serathius I would like to pick this up, Can you please help me with any information required?

serathius:
Sure, feel free to reach me on [Slack](https://github.com/etcd-io/etcd#contact).

preetham:
@serathius Can you please check the PR: https://github.com/etcd-io/etcd/pull/15963? Does this line up with the expectation?

jmhbnz:
Discussed during sig etcd triage meeting. It's possible this is already resolved via #16698.

Assigned @moficodes to check and close if so.

moficodes:
@jmhbnz will take a look Thanks.

moficodes:
Tried with `gofail-disable` and build. 

Was hoping `test-robustness` would fail. It did not. Might be doing something wrong.

```bash
user@vs-code:~/documents/github.com/moficodes/etcd$ make gofail-disable && make build
```

```bash
cd tools/mod; go install go.etcd.io/gofail@v0.1.0
gofail disable server/etcdserver/ server/storage/backend/ server/storage/mvcc/ server/storage/wal/
cd ./server && go mod tidy
cd ./etcdutl && go mod tidy
cd ./etcdctl && go mod tidy
cd ./tests && go mod tidy
GO_BUILD_FLAGS=" -v -mod=readonly" ./scripts/build.sh
Running etcd_build
% 'rm' '-f' 'bin/etcd'
% (cd server && 'env' 'CGO_ENABLED=0' 'GO_BUILD_FLAGS= -v -mod=readonly' 'GOOS=linux' 'GOARCH=amd64' 'go' 'build' '-v' '-mod=readonly' '-trimpath' '-installsuffix=cgo' '-ldflags=-X=go.etcd.io/etcd/api/v3/version.GitSHA=904c0769e' '-o=../bin/etcd' '.')
% 'rm' '-f' 'bin/etcdutl'
[...]

PR #16698, opened 2023-10-06
add existing http health check handler e2e test

chaochn47:
Part of #16007

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1822,6 +1822,7 @@ func (s *EtcdServer) apply(
 			zap.Stringer("type", e.Type))
 		switch e.Type {
 		case raftpb.EntryNormal:
+			// gofail: var beforeApplyOneEntryNormal struct{}
 			s.applyEntryNormal(&e)
 			s.setAppliedIndex(e.Index)
 			s.setTerm(e.Term)
--- a/tests/e2e/http_health_check_test.go
+++ b/tests/e2e/http_health_check_test.go
@@ -1,0 +1,226 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+type healthCheckConfig struct {
+	url                  string
+	expectedStatusCode   int
+	expectedTimeoutError bool
+}
+
+func TestHTTPHealthHandler(t *testing.T) {
+	e2e.BeforeTest(t)
+	client := &http.Client{}
+	tcs := []struct {
+		name           string
+		injectFailure  func(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster)
+		clusterOptions []e2e.EPClusterOption
+		healthChecks   []healthCheckConfig
+	}{
+		{
+			name:           "no failures", // happy case
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(1)},
+			healthChecks: []healthCheckConfig{
+				{
+					url:                "/health",
+					expectedStatusCode: http.StatusOK,
+				},
+			},
+		},
+		{
+			name:           "activated no space alarm",
+			injectFailure:  triggerNoSpaceAlarm,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(1), e2e.WithQuotaBackendBytes(int64(13 * os.Getpagesize()))},
+			healthChecks: []healthCheckConfig{
+				{
+					url:                "/health",
+					expectedStatusCode: http.StatusServiceUnavailable,
+				},
+				{
+					url:                "/health?exclude=NOSPACE",
+					expectedStatusCode: http.StatusOK,
+				},
+			},
+		},
+		{
+			name:           "overloaded server slow apply",
+			injectFailure:  triggerSlowApply,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(3), e2e.WithGoFailEnabled(true)},
+			healthChecks: []healthCheckConfig{
+				{
+					url:                "/health?serializable=true",
+					expectedStatusCode: http.StatusOK,
+				},
+				{
+					url:                  "/health?serializable=false",
+					expectedTimeoutError: true,
+				},
+			},
+		},
+		{
+			name:           "network partitioned",
+			injectFailure:  blackhole,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(3), e2e.WithIsPeerTLS(true), e2e.WithPeerProxy(true)},
+			healthChecks: []healthCheckConfig{
+				{
+					url:                "/health?serializable=true",
+					expectedStatusCode: http.StatusOK,
+				},
+				{
+					url:                  "/health?serializable=false",
+					expectedTimeoutError: true,
+					expectedStatusCode:   http.StatusServiceUnavailable,
+					// old leader may return "etcdserver: leader changed" error with 503 in ReadIndex leaderChangedNotifier
+				},
+			},
+		},
+		{
+			name:           "raft loop deadlock",
+			injectFailure:  triggerRaftLoopDeadLock,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(1), e2e.WithGoFailEnabled(true)},
+			healthChecks: []healthCheckConfig{
+				{
+					// current kubeadm etcd liveness check failed to detect raft loop deadlock in steady state
+					// ref. https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/etcd/local.go#L225-L226
+					// current liveness probe depends on the etcd /health check has a flaw that new /livez check should resolve.
+					url:                "/health?serializable=true",
+					expectedStatusCode: http.StatusOK,
+				},
+				{
+					url:                  "/health?serializable=false",
+					expectedTimeoutError: true,
+				},
+			},
+		},
+		// verify that auth enabled serializable read must go through mvcc
+		{
+			name:           "slow buffer write back with auth enabled",
+			injectFailure:  triggerSlowBufferWriteBackWithAuth,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(1), e2e.WithGoFailEnabled(true)},
+			healthChecks: []healthCheckConfig{
+				{
+					url:                  "/health?serializable=true",
+					expectedTimeoutError: true,
+				},
+			},
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+			defer cancel()
+			clus, err := e2e.NewEtcdProcessCluster(ctx, t, tc.clusterOptions...)
+			require.NoError(t, err)
+			defer clus.Close()
+			testutils.ExecuteUntil(ctx, t, func() {
+				if tc.injectFailure != nil {
+					tc.injectFailure(ctx, t, clus)
+				}
+
+				for _, hc := range tc.healthChecks {
+					requestURL := clus.Procs[0].EndpointsHTTP()[0] + hc.url
+					t.Logf("health check URL is %s", requestURL)
+					doHealthCheckAndVerify(t, client, requestURL, hc.expectedStatusCode, hc.expectedTimeoutError)
+				}
+			})
+		})
+	}
+}
+
+func doHealthCheckAndVerify(t *testing.T, client *http.Client, url string, expectStatusCode int, expectTimeoutError bool) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	require.NoErrorf(t, err, "failed to creat request %+v", err)
+	resp, herr := client.Do(req)
+	cancel()
+	if expectTimeoutError {
+		if herr != nil && strings.Contains(herr.Error(), context.DeadlineExceeded.Error()) {
+			return
+		}
+	}
+	require.NoErrorf(t, herr, "failed to get response %+v", err)
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	resp.Body.Close()
+	require.NoErrorf(t, err, "failed to read response %+v", err)
+
+	t.Logf("health check response body is: %s", body)
+	require.Equal(t, expectStatusCode, resp.StatusCode)
+}
+
+func triggerNoSpaceAlarm(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	buf := strings.Repeat("b", os.Getpagesize())
+	etcdctl := clus.Etcdctl()
+	for {
+		if err := etcdctl.Put(ctx, "foo", buf, config.PutOptions{}); err != nil {
+			if !strings.Contains(err.Error(), "etcdserver: mvcc: database space exceeded") {
+				t.Fatal(err)
+			}
+			break
+		}
+	}
+}
+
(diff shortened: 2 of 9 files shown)


tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+			break
+		}
+	}
+}
+
+func triggerSlowApply(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	// the following proposal will be blocked at applying stage
+	// because when apply index < committed index, linearizable read would time out.
+	require.NoError(t, clus.Procs[0].Failpoints().Setup(ctx, "beforeApplyEntryNormal", `sleep(3000)`))
```
serathius, line 187: Did you mean 3 seconds?
```suggestion
	require.NoError(t, clus.Procs[0].Failpoints().Setup(ctx, "beforeApplyEntryNormal", `sleep(3s)`))
```

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(3), e2e.WithGoFailEnabled(true)},
+			healthChecks: []healthCheckConfig{
+				{
+					url:              "/health?serializable=true",
+					expectStatusCode: http.StatusOK,
+				},
+				{
+					url:                "/health?serializable=false",
+					expectTimeoutError: true,
+				},
```
serathius, line 89: Heh, I'm not sure why `slow apply` would interfere with linearizability. 

Ohh, maybe I know. Does it jus block a raft loop? Raft is used to acquire read index. It makes sense, however you could have just inserted failpoint to readIndex code for clarity. What do you think?

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+			},
+		},
+		{
+			name:           "network partitioned",
+			injectFailure:  blackhole,
```
serathius, line 94: Love this!

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+func blackhole(_ context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	member := clus.Procs[0]
+	proxy := member.PeerProxy()
+	t.Logf("Blackholing traffic from and to member %q", member.Config().Name)
+	proxy.BlackholeTx()
+	proxy.BlackholeRx()
+}
+
+func triggerRaftLoopDeadLock(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	err := clus.Procs[0].Failpoints().Setup(ctx, "raftBeforeSave", `sleep(3000)`)
```
serathius, line 200: ditto
```suggestion
	err := clus.Procs[0].Failpoints().Setup(ctx, "raftBeforeSave", `sleep(3s)`)
```

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+					url:                "/health?serializable=false",
+					expectTimeoutError: true,
+				},
+			},
+		},
+		{
+			name:           "raft loop deadlock auth is enabled",
+			injectFailure:  triggerRaftLoopDeadLock,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(3), e2e.WithGoFailEnabled(true)},
+			authEnabled:    true,
```
serathius, line 111: What has auth to do with this? Is this a test of healthchecks when auth is enabled or is it a test if raftLoopDeadlock is detected?

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+			clus, err := e2e.NewEtcdProcessCluster(ctx, t, tc.clusterOptions...)
+			require.NoError(t, err)
+			defer clus.Close()
+			testutils.ExecuteUntil(ctx, t, func() {
+				if tc.authEnabled {
+					setupAuth(ctx, t, clus.Etcdctl())
+				}
+				tc.injectFailure(ctx, t, clus)
```
serathius, line 137:
```suggestion
				if tc.injectFailure != nil {
					tc.injectFailure(ctx, t, clus)
				}
```

And `noop` is not needed

ahrtr: `noop` is a good practice of removing the dirty null check. 

Of course, it isn't a big deal though.

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+type healthCheckConfig struct {
+	url                string
+	expectStatusCode   int
+	expectTimeoutError bool
```
ahrtr, line 38: minor comment
```suggestion
	expectedStatusCode   int
	expectedTimeoutError bool
```

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,211 @@
...
+			},
+		},
+		{
+			name:           "raft loop deadlock auth is enabled",
+			injectFailure:  triggerRaftLoopDeadLock,
+			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(3), e2e.WithGoFailEnabled(true)},
+			authEnabled:    true,
+			healthChecks: []healthCheckConfig{
+				{
+					// current kubeadm etcd liveness check failed to detect raft loop deadlock in steady state
```
ahrtr, line 114: Can you clarify this comment?

tests/e2e/http_health_check_test.go
ahrtr: +1 seems not related to auth at all

tests/e2e/http_health_check_test.go
serathius: Hmm, for me checking a function is nil is not a dirty null check, but checking interface is. See https://yourbasic.org/golang/gotcha-why-nil-error-not-equal-nil/

I prefer my approach because `noop` function is just an useless artifact due to language conventions.

Don't have strong opinion though. Leaving it up to @chaochn47

tests/e2e/http_health_check_test.go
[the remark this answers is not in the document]
serathius: True

server/etcdserver/server.go
```diff
@@ -1822,6 +1823,7 @@ func (s *EtcdServer) apply(
 			zap.Stringer("type", e.Type))
 		switch e.Type {
 		case raftpb.EntryNormal:
+			// gofail: var beforeApplyEntryNormal struct{}
```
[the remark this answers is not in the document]
serathius: Agree, we already introduced this naming convention to other fail points in `apply` function.

.github/workflows/e2e.yaml
```diff
@@ -30,11 +30,11 @@ jobs:
         run: |
           set -euo pipefail
           go clean -testcache
+          make gofail-enable
 
           echo "${TARGET}"
           case "${TARGET}" in
             linux-amd64-e2e)
-              make gofail-enable
```
serathius, line 37: This change causes a big change in convention. I would recommend to not change this, but add a skip in tests if etcd binary was not complied with gofail enabled.

E2e tests should be runnable on any etcd binary, allowing us to do run them against old releases or the exact binaries we are releasing. This creates a problem, some tests like yours require gofail a custom build binaries. How do we just expect some e2e tests to fail now?

The approach I have previously proposed was to skip the test that cannot be executed for some reason. That way you can still run `make test-e2e` and get accurate results and CI can just compile in `gofail` to run all tests. I would suggest taking this approach.

The one remaining question is "If we are skipping some tests, how do we know we don't forget any tests?". If someone removes or breaks `make gofail-enable` then we are screwed. To mitigate this risk I think we should consider adding validation in CI that ensures that etcd binary tested has expected failpoints.

tests/e2e/http_health_check_test.go
chaochn47: It is left behind when validating without root permission, serializable read could return "false" healthy. I will clean it up and add a new test case just for auth.

tests/e2e/http_health_check_test.go
chaochn47: Thanks all! Both should work, I just look up the code base and decide to use nil check for consistency.

.github/workflows/e2e.yaml
chaochn47: > This change causes a big change in convention

Just to clarify the change you meant is moving `make gofail-enable` from line 37 to line 33 that causing both `amd64` and `386` will run e2e tests with gofail enabled and compiled etcd binary, correct? 

> E2e tests should be runnable on any etcd binary, allowing us to do run them against old releases or the exact binaries we are releasing. This creates a problem, some tests like yours require gofail a custom build binaries. How do we just expect some e2e tests to fail now?

> The approach I have previously proposed was to skip the test that cannot be executed for some reason. That way you can still run make test-e2e and get accurate results and CI can just compile in gofail to run all tests. I would suggest taking this approach.

Thanks for the detailed explanation!!

I did not think it through when making this change and hope the CI would pass because local run did not fail to set up the failpoint. 

I am thinking for each test, it could opt-in to create etcd process with gofail enabled and compiled binary. If the binary does not exist, then skip. By default e2e test will use the vanilla binary even in the CI. This should resolve the problem "test the exact binaries we are releasing". 

> To mitigate this risk I think we should consider adding validation in CI that ensures that etcd binary tested has expected failpoints.

Just some naive thought, if the test itself can tell it is running on github action or k8s CI framework with some environment variables, then it should fail to guarantee we don't forget any tests.

tests/e2e/http_health_check_test.go
chaochn47: I was referring to https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/etcd/local.go#L225-L226 to demonstrate current liveness check depends on the `/health` check has a flaw that new `livez` check should resolve.

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,227 @@
...
+				{
+					url:                  "/health?serializable=false",
+					expectedTimeoutError: true,
+				},
+			},
+		},
+		// verify that auth enabled serializable read must go through mvcc
+		// this test case is skipped until https://github.com/etcd-io/etcd/pull/16697 is merged.
+		{
+			name:           "slow buffer write back with auth enabled",
```
chaochn47, line 126: This will block opening backend read tx hence serializable check failed. 

To simulate real world failure scenario, we need 
1. `backendBatchLimit` set to 0
2. inject `beforeCommit`
3. serializable check an existing key to use boltdb tx. 

The 3rd requirement is not ready yet. Hence use the `beforeWritebackBuf` sleep to simulate stalled disk write.

chaochn47:
Ping @ahrtr @serathius, unless the last skipped test case, I think the others should be good to go.

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,237 @@
...
+func triggerRaftLoopDeadLock(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	err := clus.Procs[0].Failpoints().SetupHTTP(ctx, "raftBeforeSave", `sleep("3s")`)
+	if err != nil {
+		t.Skip("failpoint raftBeforeSave is not registered; please run `make gofail-enable` and compile etcd binary")
+	}
+	clus.Procs[0].Etcdctl().Put(context.Background(), "foo", "bar", config.PutOptions{})
+}
+
+func triggerSlowBufferWriteBackWithAuth(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	t.Skip("wait until https://github.com/etcd-io/etcd/pull/16697 merged that using root permissions doing serializable read")
```
ahrtr, line 224: https://github.com/etcd-io/etcd/pull/16697 has already been merged.

chaochn47: Removed the skip statement.

chaochn47:
Ping @serathius @ahrtr

/cc @siyuanfoundation since failure injections should be re-usable.

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+			break
+		}
+	}
+}
+
+func triggerSlowApply(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster) {
+	// the following proposal will be blocked at applying stage
+	// because when apply index < committed index, linearizable read would time out.
+	err := clus.Procs[0].Failpoints().SetupHTTP(ctx, "beforeApplyOneEntryNormal", `sleep("3s")`)
```
serathius, line 199:
```suggestion
	failpoint := "beforeApplyOneEntryNormal"
	if clus.Procs[0].Failpoints().Available(failpoint) {
		t.Skip("Failpoint not available")
	}
	err := clus.Procs[0].Failpoints().SetupHTTP(ctx, failpoint, `sleep("3s")`)
```

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		t.Skip("failpoint beforeApplyOneEntryNormal is not registered; please run `make gofail-enable` and compile etcd binary")
+	}
```
serathius, line 202:
```suggestion
	failpoint := "beforeApplyOneEntryNormal"
	if clus.Procs[0].Failpoints().Available(failpoint) {
		t.Skipf("failpoint %s is not registered; please run `make gofail-enable` and compile etcd binary", failpoint)
	}
	err := clus.Procs[0].Failpoints().SetupHTTP(ctx, failpoint, `sleep("3s")`)
	if err != nil {
		return err
	}
```

chaochn47: https://github.com/etcd-io/etcd/blob/f5e82260da1db26717721cfbf3a41657c238b134/tests/framework/e2e/etcd_process.go#L376-L386

Available will panic if gofail is not enabled and compiled into etcd binary. 

so that's why I use SetupHTTP.

chaochn47: @serathius ^

chaochn47: Added a Enabled() function to fix this.

chaochn47:
ping @ahrtr @serathius if this PR is good to merge. 

@siyuanfoundation and I are working on adding individual checks so test function in this PR can be re-used.

tests/e2e/http_health_check_test.go
```diff
@@ -0,0 +1,236 @@
...
+	require.NoError(t, etcdctl.AuthEnable(ctx))
+
+	require.NoError(t, clus.Procs[0].Failpoints().SetupHTTP(ctx, "beforeWritebackBuf", `sleep("3s")`))
+	clus.Procs[0].Etcdctl(e2e.WithAuth("root", "root")).Put(context.Background(), "foo", "bar", config.PutOptions{Timeout: 200 * time.Millisecond})
+}
+
+func checkFailPointEnabled(t *testing.T, member e2e.EtcdProcess) {
+	if !member.Failpoints().Enabled() {
+		t.Skip("please run `make gofail-enable` and compile etcd binary")
+	}
```
ahrtr, line 235: Can we perform the check automatically in e2e.BeforeTest? https://github.com/etcd-io/etcd/blob/262a3a1208a74e7dc0719c2c75096e934242546b/tests/framework/e2e/testing.go#L23

Similar to integration.BeforeTest,
https://github.com/etcd-io/etcd/blob/262a3a1208a74e7dc0719c2c75096e934242546b/tests/framework/integration/testing.go#L100-L108

chaochn47: We need to spin up gofail http server by `GOFAIL_HTTP="127.0.0.1:1234" ./etcd` first and send http request to determine if gofail is enabled. It sounds like over engineering to implement it in `e2e.BeforeTest`. 

I prefer keep the PR as it is unless you have better ideas. WDYT? @ahrtr

ahrtr: At least let's move this function into e2e framework as a common function?

The logic should be something like below,

```
if cfg.GoFailEnabled && !member.Failpoints().Enabled() {
    t.Skip("xxxxx")
}
```

jmhbnz: Note we have an open issue and stale pr for robustness that relates to this:
- https://github.com/etcd-io/etcd/issues/15258
- https://github.com/etcd-io/etcd/pull/15963

Would be great if we could close out that issue as part of this pr if we are already looking at this area.

chaochn47: > At least let's move this function into e2e framework as a common function?
> 
> The logic should be something like below,
> 
> ```
> if cfg.GoFailEnabled && !member.Failpoints().Enabled() {
>     t.Skip("xxxxx")
> }
> ```

Implemented.

tests/e2e/http_health_check_test.go
serathius: This should be a `TODO` left in code.

serathius (approved):

tests/framework/e2e/cluster.go
```diff
@@ -420,6 +420,13 @@ func StartEtcdProcessCluster(ctx context.Context, epc *EtcdProcessCluster, cfg *
 		}
 	}
 
+	for _, proc := range epc.Procs {
+		if cfg.GoFailEnabled && !proc.Failpoints().Enabled() {
+			epc.Close()
+			t.Skip("please run 'make gofail-enable && make build' before running the test")
+		}
```
ahrtr, line 427: Nice. It's much better now. Each use case doesn't need to bother to do this check.

ahrtr (approved):
LGTM

Issue #16673, opened 2023-09-30
Improvements to robustness tests to prevent issues like --experimental-wait-cluster-ready-timeout


serathius:
### What would you like to be added?

--experimental-wait-cluster-ready-timeout showed in https://github.com/etcd-io/etcd/pull/16672 that adding features to etcd is hard, especially if those are not well tested as they are optional. So testing just the default configuration is not enough.

Some ideas on preventing similar issues:
* Robustness tests should validate all possible configurations. To get good coverage we should look through etcd optional flags and pick a random subset of them.
* Testing etcd bootstrap is as important. We might extend failpoint injection from being HTTP only, to also environment variable based. 
* Sleep failpoints! We should implement them already
* Test adding and removing members. Especially in context of migration to v3 store.

### Why is this needed?

Correctnes

PR #16707, opened 2023-10-07
Dynamically generate flags passed to etcd binary

serathius:
Part of https://github.com/etcd-io/etcd/issues/16673 to randomize etcd configuration we need to expand coverage of flags that can be provided in e2e tests. Instead of manually transforming embed.Config struct fields, we can use flagset to automatically generate flags.

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -17,6 +17,7 @@ package e2e
 import (
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"net/url"
 	"path"
@@ -30,10 +31,8 @@ import (
 	"go.uber.org/zap/zaptest"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
-	"go.etcd.io/etcd/client/pkg/v3/logutil"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/pkg/v3/proxy"
-	config2 "go.etcd.io/etcd/server/v3/config"
 	"go.etcd.io/etcd/server/v3/embed"
 	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/tests/v3/framework/config"
@@ -538,9 +537,6 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 	if cfg.EnableV2 {
 		args = append(args, "--enable-v2")
 	}
-	if cfg.ServerConfig.ExperimentalInitialCorruptCheck {
-		args = append(args, "--experimental-initial-corrupt-check")
-	}
 	var murl string
 	if cfg.MetricsURLScheme != "" {
 		murl = (&url.URL{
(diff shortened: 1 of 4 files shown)


serathius:
cc @ahrtr

tests/framework/e2e/cluster.go
```diff
@@ -654,6 +602,20 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
...
+func values(cfg embed.Config) map[string]string {
+	fs := flag.NewFlagSet("etcd", flag.ContinueOnError)
+	cfg.AddFlags(fs)
+	values := map[string]string{}
+	fs.VisitAll(func(f *flag.Flag) {
+		var value string
+		if value != "false" && value != "0" {
+			value = f.Value.String()
+		}
+		values[f.Name] = value
```
ahrtr, line 614:
```suggestion
		value = f.Value.String()
		if value != "" && value != "false" && value != "0" {
			values[f.Name] = value
		}
```

serathius: I think the result is different, but I will double check if there impact on the results.

serathius: ok, makes sense.

ahrtr (approved):
LGTM

Nice simplification & refactoring! thx

serathius:
Flake

tests/framework/e2e/cluster.go
serathius: Needed to revert. The code change makes a difference for flags with default value true, that is set to false during the test.

serathius: If I have time I will add a unit test for this.

jmhbnz (approved):
LGTM - Nice work.

tests/framework/e2e/cluster.go
ahrtr: You define a variable at line 610 `var value string`, its default value is always ""; then you check its value at 611 `if value != "false" && value != "0"  { ... }`, the if condition will always be true. What's the point of the 2 lines of code?

serathius: Don't have time to deep dive here, the difference is one code works. Other code fails the tests. I will add unit test to show the difference.

serathius: And I responded, that after I applied you change the test started to consistently fail and it needs further investigation.

ahrtr (changes requested):
https://github.com/etcd-io/etcd/pull/16707#discussion_r1354478954

action needed:
- fix the simple golang programming problem (always checking a variable which is always "") as I mentioned.
- resolve related test failure

serathius:
Created the test

tests/framework/e2e/cluster.go
serathius: With https://github.com/etcd-io/etcd/pull/16747 I have confirmed that your change breaks logic:
```
--- FAIL: TestEtcdServerProcessConfig (0.00s)
    --- FAIL: TestEtcdServerProcessConfig/StrictReconfigCheck (0.00s)
        cluster_test.go:214: diff:   []string{
                ... // 11 identical elements
                "/tmp/fake/member-0",
                "--initial-cluster-token=new",
            +   "--strict-reconfig-check=false",
              }
FAIL
FAIL    go.etcd.io/etcd/tests/v3/framework/e2e  0.051s
FAIL
```

ahrtr: Please forget my original suggested change. Instead, change

```
		var value string
		if value != "false" && value != "0" {
			value = f.Value.String()
		}
```
 to

```
value := f.Value.String()
```

serathius: Changed updated it. PTAL. Now with unit test it should be clear what was the problem.

serathius:
ping @ahrtr

ahrtr (approved):

PR #16710, opened 2023-10-08
Add method `(*EtcdServer) IsRaftLoopBlocked` to support checking whether the raft loop is blocked

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/16007

### DIFF
--- a/server/etcdserver/raft.go
+++ b/server/etcdserver/raft.go
@@ -99,6 +99,9 @@ type raftNode struct {
 
 	stopped chan struct{}
 	done    chan struct{}
+
+	// used by liveness probe to check whether the raftloop is blocked.
+	dummyc chan struct{}
 }
 
 type raftNodeConfig struct {
@@ -142,6 +145,7 @@ func newRaftNode(cfg raftNodeConfig) *raftNode {
 		applyc:     make(chan toApply),
 		stopped:    make(chan struct{}),
 		done:       make(chan struct{}),
+		dummyc:     make(chan struct{}),
 	}
 	if r.heartbeat == 0 {
 		r.ticker = &time.Ticker{}
@@ -322,6 +326,8 @@ func (r *raftNode) start(rh *raftReadyHandler) {
 					// notify etcdserver that raft has already been notified or advanced.
 					raftAdvancedC <- struct{}{}
 				}
+			case <-r.dummyc:
+				r.lg.Debug("Received dummy event")
 			case <-r.stopped:
 				return
 			}
@@ -411,6 +417,17 @@ func (r *raftNode) onStop() {
 		r.lg.Panic("failed to close Raft storage", zap.Error(err))
 	}
 	close(r.done)
+}
+
+func (r *raftNode) trySendDummyEvent(timeout time.Duration) error {
+	select {
+	case r.dummyc <- struct{}{}:
+	case <-r.done:
+	case <-time.After(timeout):
+		return fmt.Errorf("failed to send dummy event in %s", timeout.String())
+	}
+
+	return nil
 }
 
 // for testing
--- a/server/etcdserver/raft_test.go
+++ b/server/etcdserver/raft_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap/zaptest"
 
 	"go.etcd.io/etcd/client/pkg/v3/types"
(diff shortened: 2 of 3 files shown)


ahrtr:
cc @serathius

chaochn47:
Copied from the design doc comment: 

> A counter could be added to the raftNode.tick function [1] and the prober could just look up the counter to decide if the raft loop is deadlocked or not.

> [1] [https://github.com/etcd-io/etcd/blob/aa97484166d2b3fb6afeb4390344e68b02afb566/server/etcdserver/raft.go#L155-L159](https://www.google.com/url?q=https://github.com/etcd-io/etcd/blob/aa97484166d2b3fb6afeb4390344e68b02afb566/server/etcdserver/raft.go%23L155-L159&sa=D&source=docs&ust=1696827572854607&usg=AOvVaw1vkCpb4CNijtUbcTChVpkm)

Since the raft loop deadlock will block the next select statement execution: 

I can see there are two approaches: 

1. prober sends a request to the etcd server and waits for a response back
2. prober queries the etcd server if in the past x second, there is at least select statement execution which is the raft tick timer. It immediately sends the response back. 

With the 1s k8s default prober timeout, looks like the 2nd approach fits this better, what do you think? @ahrtr

chaochn47:
cc @siyuanfoundation ^

ahrtr:
> 2\. prober queries the etcd server if in the past x second, there is at least one select statement execution which is the raft tick timer. It immediately sends the response back.

This means that you need to remember the timestamp of the last tick, and check it in the liveness probe something like `time.Sub(now - t) > timeout`. You will get immediate result, but It will be affected when clock drift. I would suggest to avoid this approach.

Note that this PR just provides a basic functionality for checking if the raftloop blocks. You can check it async.

chaochn47:
> This means that you need to remember the timestamp of the last tick, and check it in the liveness probe something like `time.Sub(now - t) > timeout`. You will get immediate result, but It will be affected when clock drift. I would suggest to avoid this approach.

It's not necessary, sent a drafted PR to demonstrate https://github.com/etcd-io/etcd/pull/16713. 

> Note that this PR just provides a basic functionality for checking if the raftloop blocks. You can check it async.

Yeah, assuming the "async" here means try send to the dummy channel and then go ahead to the remaining checks in the prober and then validate the previous sent has completed. It seems complicated compared with the counter approach. WDYT?

siyuanfoundation:
> Copied from the design doc comment:
> 
> > A counter could be added to the raftNode.tick function [1] and the prober could just look up the counter to decide if the raft loop is deadlocked or not.
> 
> > [1] [https://github.com/etcd-io/etcd/blob/aa97484166d2b3fb6afeb4390344e68b02afb566/server/etcdserver/raft.go#L155-L159](https://www.google.com/url?q=https://github.com/etcd-io/etcd/blob/aa97484166d2b3fb6afeb4390344e68b02afb566/server/etcdserver/raft.go%23L155-L159&sa=D&source=docs&ust=1696827572854607&usg=AOvVaw1vkCpb4CNijtUbcTChVpkm)
> 
> Since the raft loop deadlock will block the next select statement execution:
> 
> I can see there are two approaches:
> 
> 1. prober sends a request to the etcd server and waits for a response back with a configurable waiting timeout.
> 2. prober queries the etcd server if in the past x second, there is at least one select statement execution which is the raft tick timer. It immediately sends the response back.
> 
> With the goal of prober check fittng in the 1s timeout, looks like the 2nd approach is better, what do you think? @ahrtr



> Copied from the design doc comment:
> 
> > A counter could be added to the raftNode.tick function [1] and the prober could just look up the counter to decide if the raft loop is deadlocked or not.
> 
> > [1] [https://github.com/etcd-io/etcd/blob/aa97484166d2b3fb6afeb4390344e68b02afb566/server/etcdserver/raft.go#L155-L159](https://www.google.com/url?q=https://github.com/etcd-io/etcd/blob/aa97484166d2b3fb6afeb4390344e68b02afb566/server/etcdserver/raft.go%23L155-L159&sa=D&source=docs&ust=1696827572854607&usg=AOvVaw1vkCpb4CNijtUbcTChVpkm)
> 
> Since the raft loop deadlock will block the next select statement execution:
> 
> I can see there are two approaches:
> 
> 1. prober sends a request to the etcd server and waits for a response back with a configurable waiting timeout.
> 2. prober queries the etcd server if in the past x second, there is at least one select statement execution which is the raft tick timer. It immediately sends the response back.
> 
> With the goal of prober check fittng in the 1s timeout, looks like the 2nd approach is better, what do you think? @ahrtr

I think either way, the interval between checks could be too short to differentiate a slow ready process from a deadlock.  I suggest using another longer ticker to reset the count instead of resetting based on probes in https://github.com/etcd-io/etcd/pull/16713

chaochn47:
> I think either way, the interval between checks could be too short to differentiate a slow ready process from a deadlock. 

In the second approach, it's determined by the administrator case by case. 'prober interval * failure threshold' should be a sane value based on administrator judgement. e.g. if they are using a network based volume like EBS or physically attached SSD..

> I suggest using another longer ticker to reset the count instead of resetting based on probes in https://github.com/etcd-io/etcd/pull/16713

Adding another ticker may not be optimal, how will you plan set the new ticker interval, is it configurable? It may make etcd set up more complicated.

PR #16713, opened 2023-10-09
raft loop prober with counter

chaochn47:

### DIFF
--- a/server/etcdserver/raft.go
+++ b/server/etcdserver/raft.go
@@ -23,12 +23,13 @@ import (
 
 	"go.uber.org/zap"
 
+	"go.etcd.io/raft/v3"
+	"go.etcd.io/raft/v3/raftpb"
+
 	"go.etcd.io/etcd/client/pkg/v3/logutil"
 	"go.etcd.io/etcd/pkg/v3/contention"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
 	serverstorage "go.etcd.io/etcd/server/v3/storage"
-	"go.etcd.io/raft/v3"
-	"go.etcd.io/raft/v3/raftpb"
 )
 
 const (
@@ -82,6 +83,7 @@ type raftNode struct {
 
 	tickMu *sync.Mutex
 	raftNodeConfig
+	tickElapsed uint64
 
 	// a chan to send/receive snapshot
 	msgSnapC chan raftpb.Message
@@ -155,7 +157,14 @@ func newRaftNode(cfg raftNodeConfig) *raftNode {
 func (r *raftNode) tick() {
 	r.tickMu.Lock()
 	r.Tick()
+	r.tickElapsed++
 	r.tickMu.Unlock()
+}
+
+func (r *raftNode) safeReadTickElapsed() uint64 {
+	r.tickMu.Lock()
+	defer r.tickMu.Unlock()
+	return r.tickElapsed
 }
 
 // start prepares and starts raftNode in a new goroutine. It is no longer safe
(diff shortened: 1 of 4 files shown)


chaochn47:
cc @ahrtr

ahrtr:
This approach depends on the liveness probe's frequency / interval, the principle is `the instance is alive as long as there is at least one tick event since last liveness check`.

Overall looks good. Please feel free to reuse (might need minor update) the test case in https://github.com/etcd-io/etcd/pull/16710.

server/etcdserver/raft.go
```diff
@@ -155,9 +158,21 @@ func newRaftNode(cfg raftNodeConfig) *raftNode {
 func (r *raftNode) tick() {
 	r.tickMu.Lock()
 	r.Tick()
+	r.tickElapsed++
+	if r.tickElapsed == math.MaxUint64 {
```
serathius, line 162: Don't worry about int overflow. It will take a long time for etcd to overflow, I think there are many other things that will overflow first in etcd :p

server/etcdserver/server.go
```diff
@@ -1643,6 +1644,10 @@ func (s *EtcdServer) AppliedIndex() uint64 { return s.getAppliedIndex() }
 
 func (s *EtcdServer) Term() uint64 { return s.getTerm() }
 
+// ProbeRaftLoopProgress probes if the etcdserver raft loop is deadlocked
+// since last time the function is invoked.
+func (s *EtcdServer) ProbeRaftLoopProgress() bool { return s.r.safeReadTickElapsedAndClear() != 0 }
```
serathius, line 1649: Probe response should not be determined based on frequency of probing. If someone probes etcd more frequently then proposals are happening then this will return failure.

I still need to think more about the best approach (maybe a document that collects and compares the approaches would help?), but this could be improved by waiting for a tick instead of failing if there was no ticks since last probe.

chaochn47: > Probe response should not be determined based on frequency of probing. If someone probes etcd more frequently then proposals are happening then this will return failure.

Thanks, in that case maybe adding a simple throttler in etcdserver raft Node should help. 

> I still need to think more about the best approach (maybe a document that collects and compares the approaches would help?), but this could be improved by waiting for a tick instead of failing if there was no ticks since last probe.

Yeah, a small doc compares with pros and cons should help.

chaochn47: > I am also interested in how apiserver implements /livez and /readyz.

https://kubernetes.io/docs/reference/using-api/health-checks/

https://github.com/kubernetes/kubernetes/pull/70676#discussion_r232921395

@siyuanfoundation @logicalhan

siyuanfoundation: As far as I can tell, health probes in apiserver generally are not complicated. They mostly checks if something has started or initialized, such as https://github.com/kubernetes/kubernetes/blob/c7d270302c8de3afc9d7b01c70faf3a18407ce44/staging/src/k8s.io/client-go/tools/cache/shared_informer.go#L173

server/etcdserver/raft.go
chaochn47: Make sense, it will take 10^10 years with the default 100ms hearbeat timeout to overflow..

server/etcdserver/server.go
chaochn47: Addeded throttler to make raft loop prober check effective at most once every 5 seconds.

chaochn47: As promised, I wrote a small doc to compare the pros and cons. 

Could you please take a look? @ahrtr @serathius @siyuanfoundation Thanks~

ahrtr: Thanks. Overall this PR looks good to me, also added comments in the doc.

jmhbnz:
Discussed during sig-etcd triage meeting. This is still an important pr for etcd, @chaochn47 do you have some time to rebase and continue this effort?

chaochn47:
Yeah, I will take a look and drive to the resolution next week.

Issue #16712, opened 2023-10-09
Bump golang to 1.20.9 for 3.4/3.5 and to 1.21.2 for main


ahrtr:
### What would you like to be added?

Both 1.20.9 and 1.21.2 include a security fix.

### Why is this needed?

For better security.

ahrtr:
@0scvr Please feel free to deliver a PR, thx

ahrtr:
Will close this issue after the PR of updating changelog is raised & merged.
- 3.5: https://github.com/etcd-io/etcd/pull/16730
- 3.4: https://github.com/etcd-io/etcd/pull/16729

PR #16727, opened 2023-10-09
Upgrade go version to 1.21.2

0scvr:
To be up to date with the newest security fixes.

See #16712.


jmhbnz (commented):
Hey @0scvr - Thanks for raising this, the DCO check is failing as there is a minor difference between expected and actual, refer output below:

```diff
+ Expected - "Oscar 71343264+0scvr@users.noreply.github.com" 
- Actual - "0scvr 71343264+0scvr@users.noreply.github.com"
```

Looks like a case of a zero instead of a capital O.

jmhbnz (approved):
LGTM on workflow green.

ahrtr (approved):

PR #16737, opened 2023-10-10
Continue on e2e failure

serathius:
E2e tests has been flaky, with failfast and 2 scenarios tests are twice as fragile. Any early failure of one of the scenarios cases second to be canceled. Retryign always require running both scenarios.

Intead let's just wait as in other tests allowing us to retry just the scenario that failed

### DIFF
--- a/.github/workflows/e2e-arm64-template.yaml
+++ b/.github/workflows/e2e-arm64-template.yaml
@@ -17,7 +17,7 @@ jobs:
       run:
         shell: bash
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         target:
           - linux-arm64-e2e
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -11,7 +11,7 @@ jobs:
       run:
         shell: bash
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         target:
           - linux-amd64-e2e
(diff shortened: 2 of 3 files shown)


serathius:
cc @ahrtr @jmhbnz

jmhbnz (approved):
LGTM - However can we please update `e2e-arm64-template.yaml` to match? 

My initial reaction is these two files should stay in sync as much as possible given arm64 is tier 1 now.

https://github.com/etcd-io/etcd/blob/a8957107562ebb2b33cf463f46d182b9690b51c0/.github/workflows/e2e-arm64-template.yaml#L20

ahrtr:
> E2e tests has been flaky

This is actually what we should try to resolve. 

> Intead let's just wait as in other tests allowing us to retry just the scenario that failed

I am on the side of fast fail so that we can get & analyze & fix the error as soon as possible, instead of waiting for the whole e2e test suite finishes. But I won't insist on this non-critical thing.

serathius:
> I am on the side of fast fail so that we can get & analyze & fix the error as soon as possible, instead of waiting for the whole e2e test suite finishes. But I won't insist on this non-critical thing.

It doesn't work like you think. It's not about getting error, it is about having single workflow failed canceling all other workflows. The is no difference in when you get signal that a test failed. The difference is whether you need to rerun single, or all workloads after a failure. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

serathius:
ping @ahrtr

ahrtr (approved):

PR #16738, opened 2023-10-11
Robustness test: probabilistic packet loss

ZhouJianMS:
Use gofail to inject 50% possibility that sending messages not executed.


serathius:
Ambitious goal, have you considered using [PeerProxy](https://github.com/etcd-io/etcd/blob/main/tests/framework/e2e/etcd_process.go#L58C2-L58C26) instead of failpoint?

PR #16748, opened 2023-10-12
Test dynamic flags v2

serathius:
cc @ahrtr


tests/framework/e2e/cluster.go
```diff
@@ -509,27 +509,27 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
...
 	args := []string{
-		"--name", name,
-		"--listen-client-urls", strings.Join(curls, ","),
-		"--advertise-client-urls", strings.Join(curls, ","),
-		"--listen-peer-urls", peerListenUrl.String(),
-		"--initial-advertise-peer-urls", peerAdvertiseUrl.String(),
-		"--initial-cluster-token", cfg.ServerConfig.InitialClusterToken,
-		"--data-dir", dataDirPath,
-		"--snapshot-count", fmt.Sprintf("%d", cfg.ServerConfig.SnapshotCount),
+		"--name=" + name,
```
ahrtr, line 512: a quick question: why change this? do you see any issue?

ahrtr: it seems that you just want to make it easier to verify the args in UT?

ahrtr:
Please ping me when the test is green.

tests/framework/e2e/cluster.go
serathius: Yes, I want to be 100% sure that switching to dynamically generated flags doesn't make any changes in the provided flags.

serathius:
Ups, looks like a surprising bug. I was not able to switch `--data-dir` to use `=`. Don't know why, might require investigation as this looks like a bug in etcd server.

serathius:
ping @ahrtr

ahrtr:
> I was not able to switch `--data-dir` to use `=`. Don't know why, might require investigation as this looks like a bug in etcd server.

I assume you will create a followup.

Issue #9533, opened 2018-04-05
embed server never becomes ready / can't be stopped in certain start failure conditions


Quentin-M:
Hi guys,

Please find below a simple excerpt where we try to start an embed server, wait for it to be ready up to a certain point or stop it otherwise.
 
```
        // Assume .MemberAdd() was already called properly.

        c.server, err = embed.StartEtcd(etcdCfg)
	if err != nil {
		// embed.StartEtcd() will by itself call 
		//   `for _, sctx := range e.sctxs { close(sctx.serversC) }
		//   e.Close()
		// in case of an early failure.
		return fmt.Errorf("failed to start etcd: %s", err)
	}

	select {
	case <-c.server.Server.ReadyNotify():
		break
	case <-ctx.Done():
		//c.server.Server.HardStop()
		//c.server.Server = nil

		c.server.Close()

		return fmt.Errorf("server took too long to become ready")
	}
```

When we are setting up an embed server to join an existing cluster, it may take some time for the server to become ready as it needs to sync data. If anything happens during the transmission of the snapshot, or if the member is removed in the meantime, the etcdserver will notice it and rafthttp will stop the peers (see below) and idle. The allocated time for etcd to start runs out and the context cancels (or we read errors from the error channel). Therefore, we now try to Close() the server so etcd can be stopped and the ports freed - so we can retry or move on without having zombie etcd goroutines/servers around.

However, Close() is not too happy about it and stale. From my debugging, [this](https://github.com/coreos/etcd/blob/5d97bccff2be43a2baadc92663351358d0681ea8/embed/etcd.go#L228) line right here is blocking. I am not exactly sure of the lifecycle of this channel.

I would be happy to just attempt at closing manually the Peers / Listeners (as they are exported) (I haven't tried yet - and I am not sure that would be enough anyways), but at least, the Metrics server is not exported and thus could not be closed.

Do you have any suggestions?

```
I | embed: peerTLS: cert = /var/lib/eco/etcd/fixtures/peer/cert.pem, key = /var/lib/eco/etcd/fixtures/peer/key.pem, ca = , trusted-ca = , client-cert-auth = false, crl-file =
I | embed: listening for peers on https://172.16.4.239:2380
I | embed: listening for client requests on 172.16.4.239:2379
I | etcdserver: recovered store from snapshot at index 295
I | mvcc: restore compact to 239897
I | etcdserver: name = i-06429521c23f59cc7
[...]

PR #16754, opened 2023-10-13
Stop waiting for the ready notification if the server has stopped

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/9533


ahrtr:
Need to add a test.

server/embed/serve.go
```diff
@@ -99,7 +100,12 @@ func (sctx *serveCtx) serve(
 	splitHttp bool,
 	gopts ...grpc.ServerOption) (err error) {
 	logger := defaultLog.New(io.Discard, "etcdhttp", 0)
-	<-s.ReadyNotify()
+
+	select {
+	case <-s.ReadyNotify(): // wait for e.Server to join the cluster
+	case <-s.StopNotify(): // publish aborted from 'ErrStopped'
+		return errors.ErrStopped
```
Hendrik-H, line 107: before the return `close(sctx.serversC)` needs to be invoked.
in my local test that resolves the issue.

Hendrik-H:
I believe a test case could be added to `tests/integration/embed/embed_test.go` that works like this:
- create a cluster with 3 nodes
- stop one node
- remove the stopped node from the cluster
- restart the stopped node

The start should fail and `Close()` must not hang. I'll see what I can do ...

server/embed/serve.go
ahrtr: Good catch & updated.

ahrtr:
> restart the stopped node

We shouldn't restart stopped node, but etcd shouldn't hang indefinitely when users call `Close()`. So it's OK to add such case.

Another case is to only start one node in a 3-node cluster, then the node should be blocked on the [<-s.ReadyNotify()](https://github.com/etcd-io/etcd/blob/d0d0c3333847fe6c9372a897d3493c3ad2d80b2e/server/embed/serve.go#L102), in such case, when users call Close(), it shouldn't be blocked.

Hendrik-H:
The following is a possible test case that shows that the code change fixes the issue:
```
diff --git a/tests/integration/embed/embed_test.go b/tests/integration/embed/embed_test.go
index 7cdb42934..6b5e1720d 100644
--- a/tests/integration/embed/embed_test.go
+++ b/tests/integration/embed/embed_test.go
@@ -187,6 +187,77 @@ func testEmbedEtcdGracefulStop(t *testing.T, secure bool) {
 	}
 }
 
+func TestEmbedEtcdClusterGracefulStopSecure(t *testing.T) {
+	testEmbedEtcdClusterGracefulStop(t, true)
+}
+
+func TestEmbedEtcdClusterGracefulStopInsecure(t *testing.T) {
+	testEmbedEtcdClusterGracefulStop(t, false)
+}
+
+func testEmbedEtcdClusterGracefulStop(t *testing.T, secure bool) {
+	testutil.SkipTestIfShortMode(t, "Cannot start embedded cluster in --short tests")
+
+	nodes := 3 // need to be an odd number larger than 1
+
+	urls := newEmbedURLs(secure, nodes*2)
+
+	// start a cluster
+	var err error
+	cfg := make([]*embed.Config, nodes)
+	e := make([]*embed.Etcd, nodes)
+	for i := 0; i < nodes; i++ {
+		cfg[i] = setupEmbedClusterCfg(t, secure, urls[0:nodes], urls[nodes:], i)
+		e[i], err = embed.StartEtcd(cfg[i])
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// wait for all nodes to be ready
+	for i := 0; i < nodes; i++ {
+		select {
+		case <-e[i].Server.ReadyNotify(): // wait for e.Server to join the cluster
+		case <-time.After(10 * time.Second):
+			t.Fatalf("node %d did not join the cluster in time", i)
+		}
+	}
+
+	// stop one node
+	e[nodes-1].Close()
+
+	// remove the stopped node from the cluster
+	_, err = e[0].Server.RemoveMember(context.Background(), uint64(e[nodes-1].Server.MemberId()))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// start the node again -> fails as it was removed from the cluster
+	cfg[nodes-1] = setupEmbedClusterCfg(t, secure, urls[0:nodes], urls[nodes:], nodes-1)
+	e[nodes-1], err = embed.StartEtcd(cfg[nodes-1])
+	if err != nil {
+		t.Fatal(err)
+	}
+	select {
+	case <-e[nodes-1].Server.ReadyNotify():
+		t.Fatal("the start should have failed")
+	case <-e[nodes-1].Server.StopNotify():
+	}
+
+	go func() {
+		defer close(donec)
+		for i := 0; i < nodes; i++ {
+			e[i].Close()
+		}
+	}()
+	select {
+	case <-donec:
+	case <-time.After(2*time.Second + e[0].Server.Cfg.ReqTimeout()):
+		t.Fatalf("took too long to close servers")
+	}
+}
+
 func newEmbedURLs(secure bool, n int) (urls []url.URL) {
 	scheme := "unix"
 	if secure {
@@ -213,6 +284,30 @@ func setupEmbedCfg(cfg *embed.Config, curls []url.URL, purls []url.URL) {
 	cfg.InitialCluster = cfg.InitialCluster[1:]
 }
 
+func setupEmbedClusterCfg(t *testing.T, secure bool, curls []url.URL, purls []url.URL, n int) *embed.Config {
+	cfg := embed.NewConfig()
+	if secure {
+		cfg.ClientTLSInfo = testTLSInfo
+		cfg.PeerTLSInfo = testTLSInfo
+	}
+
+	cfg.Logger = "zap"
+	cfg.LogOutputs = []string{"/dev/null"}
+
+	cfg.Name = fmt.Sprintf("node-%d", n)
+	cfg.ClusterState = "new"
+	cfg.ListenClientUrls, cfg.AdvertiseClientUrls = []url.URL{curls[n]}, []url.URL{curls[n]}
+	cfg.ListenPeerUrls, cfg.AdvertisePeerUrls = []url.URL{purls[n]}, []url.URL{purls[n]}
+	cfg.InitialCluster = ""
+	for i := range purls {
+		cfg.InitialCluster += fmt.Sprintf(",node-%d=%s", i, purls[i].String())
+	}
+	cfg.InitialCluster = cfg.InitialCluster[1:]
+	cfg.Dir = filepath.Join(t.TempDir(), fmt.Sprintf("embed-etcd-%d", n))
+
+	return cfg
+}
+
 func TestEmbedEtcdAutoCompactionRetentionRetained(t *testing.T) {
 	cfg := embed.NewConfig()
 	urls := newEmbedURLs(false, 2)
```

I'm not sure why but the "Secure" version of the test does not work while the "Insecure" works just fine. Pretty much the identical code fix and test case also works for 3.5.9. Then only difference is that `Server.MemberId()` needs to be changed to `Server.ID()`.

ahrtr:
@Hendrik-H Please,
- add the second case (only start one node in a 3-nodes cluster) I mentioned above in a separate PR, it should be easier. The test should fail.
- cherry pick the commit in this PR into your PR. Then the test should pass.
- Feel free to continue to work on your test cases above in a followup PR.

Does it make sense to you?

Hendrik-H:
The test you suggested (starting only one node of a cluster) is actually a different issue. The `Close()` call does not result in the the server stopping and thus the goroutine still keeps hanging.
I'll try to open a new PR in a slightly different way:
- first commit my testcase
- second commit the cherry-pick of your change
- third commit the 1 node test case, which is still failing

Hendrik-H:
new PR is #16758

PR #16764, opened 2023-10-16
Limit a timeout in testing robustness validation

serathius:
As part of https://github.com/etcd-io/etcd/issues/15598 I was working on re-implementing the history patching to avoid too much space explosion. One thing I noticed that it's not easy to reproduce linearization timeout if patching is totally disabled.

To ensure that patching is working and helping reduce the runtime, I have generated and added a case to `TestValidation` that is expected to run over 5 minutes without any patching

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -94,7 +94,7 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, s testSce
 	watchProgressNotifyEnabled := report.Cluster.Cfg.ServerConfig.ExperimentalWatchProgressNotifyInterval != 0
 	validateGotAtLeastOneProgressNotify(t, report.Client, s.watch.requestProgress || watchProgressNotifyEnabled)
 	validateConfig := validate.Config{ExpectRevisionUnique: s.traffic.ExpectUniqueRevision()}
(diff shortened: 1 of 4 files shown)


serathius:
cc @ahrtr

serathius:
ping @ahrtr

ahrtr:
a quick question before I take a closer look, why add so many json files into the repo?

serathius:
Sorry for forgetting about this PR and your question.

> a quick question before I take a closer look, why add so many json files into the repo?

Those are report files from robustness tests, that I added to the repo for testing. They are used to ensure there is no breaking change nor performance regression on robustness testing.

 There is a testdata directory that you can just add a robustness report from CI and a test will run robustness validation on it. It's pretty useful to allow reproduction of bugs in robustness test framework.

serathius:
ping @ahrtr

ahrtr (approved):
lgtm

Please rebase this PR before merging. Previously I ran into a couple of cases which github did not identify the conflict.

serathius:
It will be more tricky than anticipated :(

serathius:
Ok, I run into a problem with creating a cpu bound test that will execute in CI in a predictable time. 
Locally on my machine the test runs 2s, while in CI it times out after 60s. 

Will remove adding the scenario for now.

PR #16766, opened 2023-10-16
Add member replace failpoint to robustness tests

serathius:

### DIFF
--- a/tests/robustness/failpoint/failpoint.go
+++ b/tests/robustness/failpoint/failpoint.go
@@ -46,6 +46,7 @@ var (
 		RaftBeforeFollowerSendPanic, RaftBeforeApplySnapPanic, RaftAfterApplySnapPanic, RaftAfterWALReleasePanic,
 		RaftBeforeSaveSnapPanic, RaftAfterSaveSnapPanic, BlackholeUntilSnapshot,
 		beforeApplyOneConfChangeSleep,
+		MemberReplace,
 	}
 )
 
--- a/tests/robustness/failpoint/cluster.go
+++ b/tests/robustness/failpoint/cluster.go
@@ -1,0 +1,165 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package failpoint
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"go.uber.org/zap"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/server/v3/etcdserver"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+var (
+	MemberReplace Failpoint = memberReplace{}
+)
+
+type memberReplace struct{}
+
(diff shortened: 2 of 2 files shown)


tests/robustness/failpoints.go
```diff
@@ -570,3 +574,123 @@ func (f killAndGofailSleep) Available(config e2e.EtcdProcessClusterConfig, membe
...
+		return err
+	}
+	for {
+		_, found, err := getID(ctx, cc, member.Config().Name)
+		if err != nil {
+			continue
+		}
+		if found {
+			break
+		}
```
ahrtr, line 662: if `found == false`, then it's a bug.

tests/robustness/failpoints.go
```diff
@@ -570,3 +574,123 @@ func (f killAndGofailSleep) Available(config e2e.EtcdProcessClusterConfig, membe
 	}
 	return memberFailpoints.Available(f.failpoint)
 }
+
+type memberReplace struct{}
+
+func (f memberReplace) Inject(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster) error {
+	memberId := rand.Int() % len(clus.Procs)
+	member := clus.Procs[memberId]
+	endpoints := []string{}
```
ahrtr, line 583:
```suggestion
	var endpoints []string
```

ahrtr (approved):
LGTM with two minor comments.

tests/robustness/failpoints.go
serathius: getId uses member name to identify member. Adding nor starting a member doesn't set a name. Only after member joins a cluster, it makes a proposal to set its name. So this is not a bug.

tests/robustness/failpoints.go
```diff
@@ -570,3 +574,123 @@ func (f killAndGofailSleep) Available(config e2e.EtcdProcessClusterConfig, membe
...
+	}
+	lg.Info("Starting member", zap.String("member", member.Config().Name))
+	err = member.Start(ctx)
+	if err != nil {
+		return err
+	}
+	for {
+		_, found, err := getID(ctx, cc, member.Config().Name)
+		if err != nil {
+			continue
```
serathius, line 658: Should check ctx Done here.

tests/robustness/failpoint/failpoint.go
```diff
@@ -570,3 +574,133 @@ func (f killAndGofailSleep) Available(config e2e.EtcdProcessClusterConfig, membe
...
+	defer cc.Close()
+	memberID, found, err := getID(ctx, cc, member.Config().Name)
+	if err != nil {
+		return err
+	}
+	if !found {
+		t.Fatal("Member not found")
+	}
+	// Need to wait health interval for cluster to accept member changes
+	time.Sleep(etcdserver.HealthInterval)
```
jmhbnz, line 605: Do we need to have some further verification here that the member is actually operating successfully?

jmhbnz (approved):
LGTM - Just one question below

tests/robustness/failpoint/failpoint.go
```diff
@@ -570,3 +574,133 @@ func (f killAndGofailSleep) Available(config e2e.EtcdProcessClusterConfig, membe
...
+	if err != nil {
+		return err
+	}
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+		_, found, err := getID(ctx, cc, member.Config().Name)
```
fuweid, line 666: Code changes looks good. Just have one question here: do we need to wait for the sync of snapshot or logs here?

tests/robustness/failpoint/failpoint.go
serathius: No, health verification is done before failpoint injection. Member is already healthy and serving requests. Membership changes just require this additional wait.

tests/robustness/failpoint/failpoint.go
serathius: If you are asking, are we sure that restarted member is serving requests, then there is no 100% percent chance that member. After a fail point injection, we check if member is healthy and we give it a second, but that's not enough to guarantee that we record any requests after.

fuweid: Thanks for the comment. LGTM

serathius: No problem, this issue has been on my mind for long time. It's a good candidate for followup.

PR #13255, opened 2021-07-29
Mixin: Support configuring cluster label

discordianfish:
I've used camel case to be in line with kube-mixins and node-exporter's mixins since this one seems to be an alpha version.
I can change it to snake case or make all other variables camel case if that's preferred.

PR #16778, opened 2023-10-16
Fix cluster variable in mixins

QuentinBisson:
Fix mixin cluster parameter otherwise generated mixins looks like this:

```json
{"type":"prometheus","uid":"${datasource}"},"label":"cluster","name":"job","query":"label_values(etcd_server_has_leader{job=~\".*etcd.*\"}, job)","refresh":2,"type":"query"}]},"time":{"from":"now-15m","to":"now"},"timezone": "`}}{{ .Values.grafana.defaultDashboardsTimezone }}{{`","title":"etcd","uid":"c2f4e12cdf69feb95caa41a5a1b423d9"}`}}
```

where name is job when the variable name used in dashboard queries is cluster.

Issue came up here: https://github.com/prometheus-community/helm-charts/pull/3880#issuecomment-1765190231

### DIFF
--- a/contrib/mixin/dashboards/variables.libsonnet
+++ b/contrib/mixin/dashboards/variables.libsonnet
@@ -9,7 +9,7 @@ function(config) {
     + var.datasource.generalOptions.withLabel('Data Source'),
 
   cluster:
-    var.query.new(config.clusterLabel)
+    var.query.new('cluster')
     + var.query.generalOptions.withLabel('cluster')
     + var.query.withDatasourceFromVariable(self.datasource)
     + { refresh: config.dashboard_var_refresh }


QuentinBisson:
cc @ahrtr as I see you refactored to grafonnet recently

jmhbnz (commented):
Thanks for raising @QuentinBisson.

For context here is the original pr that introduced configurable cluster label: https://github.com/etcd-io/etcd/pull/13255/files 

Potential regression when we refactored the mixin recently.

Can take a look at this tonight and provide some feedback.

cc @v-zhuravlev

QuentinBisson:
Thank you @jmhbnz.

As you can see in this line https://github.com/etcd-io/etcd/pull/13255/files#diff-007651fd1693f51e6c72b237590b65d2a1a72fa3bba2d24993c2bf417bb25792R1294 the variable name used to be `cluster` but it is now set to `config.clusterLabel` which is job by default as you can see in my PR :)

jmhbnz (approved):
LGTM - Thanks for fixing this @QuentinBisson.

ahrtr:
cc @v-zhuravlev to double confirm.

v-zhuravlev:
@ahrtr , @QuentinBisson ,
my bad, tested this with cluster value of clusterLabel, therefore, not realising issue with other labels (such as job which is default).
Fix in this PR looks correct to me.

ahrtr:
thx @v-zhuravlev for the quick confirmation.

ahrtr (approved):

PR #16785, opened 2023-10-17
Ensure that cluster is able to serve linearizable get request after failure injection

serathius:
cc @fuweid @ahrtr 

Follow up from https://github.com/etcd-io/etcd/pull/16766#discussion_r1361426618


fuweid (approved):
LGTM

PR #16787, opened 2023-10-17
osutil: remove unused Unsetenv function

evanj:
The osutil.Unsetenv function is not used. Today, os.Unsetenv exists in the standard library, and is already used elsewhere in etcd. Removing this function could break things that may be importing this package. According to pkg.go.dev, it seems like the only thing is etcd forks, so it may be worth getting rid of this code. See:

https://pkg.go.dev/github.com/coreos/etcd/pkg/osutil?tab=importedby

Signed-off-by: Evan Jones <ej@evanjones.ca>

### DIFF
--- a/pkg/osutil/osutil.go
+++ b/pkg/osutil/osutil.go
@@ -15,27 +15,7 @@
 // Package osutil implements operating system-related utility functions.
 package osutil
 
-import (
-	"os"
-	"strings"
-)
-
 var (
 	// support to override setting SIG_DFL so tests don't terminate early
 	setDflSignal = dflSignal
 )
-
-func Unsetenv(key string) error {
-	envs := os.Environ()
-	os.Clearenv()
-	for _, e := range envs {
-		strs := strings.SplitN(e, "=", 2)
-		if strs[0] == key {
-			continue
-		}
-		if err := os.Setenv(strs[0], strs[1]); err != nil {
-			return err
-		}
-	}
-	return nil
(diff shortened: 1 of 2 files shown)


evanj:
Feel free to reject this change. I noticed this while looking for things that may be calling os.Setenv due to https://github.com/golang/go/issues/63567 ; I figured removing this would probably make sense. This is just a clean up that shouldn't do anything.

fuweid (approved):
LGTM. Thanks

The go stdlib `os.Unsetenv` can cover this.

jmhbnz (approved):
LGTM - Thanks @evanj

ahrtr (approved):
It's a public function. It might be used by applications. 

But why does etcd have to implement such public function? Especially it has already been supported by https://pkg.go.dev/os#Unsetenv.

I agree to remove it anyway.

jmhbnz:
Hey @ahrtr - Had a quick scan on GitHub and I can't see any meaningful references to `osutil.unsetEnv`: https://github.com/search?q=osutil.unsetEnv&type=code

Of course someone could be renaming the import package but I think the risk would be low and the OP has also checked pkg.go.dev which only showed etcd forks.  With two reviewer approvals and one maintainer approval are we ok to merge this for `main`?

ahrtr:
> With two reviewer approvals and one maintainer approval are we ok to merge this for `main`?

OK, merging... thx

PR #16788, opened 2023-10-17
Implement terminate failpoint

serathius:


ahrtr (approved):

tests/robustness/failpoint/kill.go
```diff
@@ -60,10 +61,25 @@ func (f killFailpoint) Inject(ctx context.Context, t *testing.T, lg *zap.Logger,
...
-func (f killFailpoint) Available(e2e.EtcdProcessClusterConfig, e2e.EtcdProcess) bool {
+func (f sigkillFailpoint) Available(e2e.EtcdProcessClusterConfig, e2e.EtcdProcess) bool {
+	return true
+}
+
+type sigtermFailpoint struct{}
+
+func (f sigtermFailpoint) Inject(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster) error {
+	member := clus.Procs[rand.Int()%len(clus.Procs)]
+	return member.Restart(ctx)
```
jmhbnz, line 76: In `func (f sigkillFailpoint) Inject` we clear the cache for lazyfs when in use before member starts back up again.

Should we be doing that here also?

jmhbnz (approved):
LGTM - Though curious about handling of lazyfs for restarting member, question below.

tests/robustness/failpoint/kill.go
serathius: I decided against it due to different purpose of the failpoint.

 In kill case we are testing a forceful shutdown, a immediate process death that is meant to represent the machine etcd is running on, dying. It's sensible in that case to also assume that data might not have been persisted. 
 
 In terminate case we are testing the etcd graceful shutdown, to ensure that process shutdown doesn't allow serving any incorrect data.
 
Thanks for bringing this up, I can do better naming.

jmhbnz: Thanks for the clarification. Yeah the distinction of forceful vs graceful helps convey the intent 👍🏻

fuweid (approved):
LGTM

serathius:
Oops, looks like test don't pass after rebase.

serathius:
Very weird. I have re-implement graceful termination to do the exactly what forceful termination does, but use SIGTERM instead of SIGKILL, still the results are different. In 3 node cluster terminated cluster is unable to bootstrap raft, it hangs somewhere between connecting to other members and publishing member to cluster. Issue doesn't happen in single node cluster. 

Is there something wrong with e2e testing framework or etcd? cc @ahrtr 

Added a draft commit to on top for easier reproduction of the issue.

serathius:
I think it might be some issue with peer proxy in e2e framework. https://github.com/etcd-io/etcd/actions/runs/8425584127/job/23072006524?pr=16788

PR #16792, opened 2023-10-17
add readIndex check in readyz

siyuanfoundation:

cc @chaochn47 @serathius


server/etcdserver/v3_server.go
```diff
@@ -831,6 +827,59 @@ func (s *EtcdServer) linearizableReadLoop() {
...
+			nr.notify(err)
+			continue
+		}
+		nr.uint64Val = confirmedIndex
+		// unblock all read index requested at indices before confirmedIndex
+		nr.notify(nil)
+	}
+}
+
+func (s *EtcdServer) GetReadIndex(ctx context.Context) (uint64, error) {
```
chaochn47, line 862: What's the difference between `GetReadIndex` and `LinearizableReadNotify`?

serathius:
Please merge the e2e tests before this.

chaochn47:
@siyuanfoundation DCO check failed. Understood the corrupt alarm e2e test was authored by @serathius but you have to sign off this commit to pass the check.

chaochn47:
> Please merge the e2e tests before this.

@serathius Did you mean separating the e2e test to another PR for validating existing livez / readyz checks behaviors? 

ReadIndex check changes and related test case could stay in the current PR, correct? 

/cc @siyuanfoundation

chaochn47:
> > Please merge the e2e tests before this.
> 
> @serathius Did you mean separating the e2e test to another PR for validating existing livez / readyz checks behaviors?
> 
> ReadIndex check changes and related test case could stay in the current PR, correct?
> 
> /cc @siyuanfoundation

@siyuanfoundation Existing livez readyz e2e test cases are added. Could you please rebase on top of main and resolve the comments?

server/etcdserver/v3_server.go
chaochn47: @siyuanfoundation Is there any counter argument of this above comment? Could you please clarify?

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -295,6 +301,40 @@ func TestSerializableReadCheck(t *testing.T) {
...
 }
 
+func TestReadIndexCheck(t *testing.T) {
+	be, _ := betesting.NewDefaultTmpBackend(t)
+	defer betesting.Close(t, be)
+	tests := []healthTestCase{
+		{
+			name:             "Alive ok",
+			healthCheckURL:   "/livez",
+			readIndexError:   fmt.Errorf("failed to get read index"),
```
chaochn47, line 311: remove this error?

server/etcdserver/v3_server.go
chaochn47: oh, misread that the new `readIndexLoop` won't wait for the applied index catch up. 

Right after read index is received in `linearizableReadLoop`, `readIndexNotifier` would notify the caller. I think that simply the implementation, wdyt? 

We can reduce the background go routine, avoid `readMu` lock twice and don't need to add index into the notifier struct.

siyuanfoundation:
@chaochn47 
The problem with sharing the same loop is a slow apply loop would block readIndex request when there is a pending read, which makes it not much better than just using linearizableRead. So readIndex and linearizableRead has to be in separate loops.

server/etcdserver/v3_server.go
```diff
@@ -831,6 +827,59 @@ func (s *EtcdServer) linearizableReadLoop() {
...
+		select {
+		case <-leaderChangedNotifier:
+			continue
+		case <-s.readIndexWaitc:
+		case <-s.stopping:
+			return
+		}
+
+		nextnr := newNotifier()
+		s.readMu.Lock()
```
chaochn47, line 843: How about adding a separate lock dedicated for read index so adding this loop has no impact on the `linearizableReadLoop`?

server/etcdserver/v3_server.go
```diff
@@ -831,6 +827,59 @@ func (s *EtcdServer) linearizableReadLoop() {
...
+		nr.notify(nil)
+	}
+}
+
+func (s *EtcdServer) GetReadIndex(ctx context.Context) (uint64, error) {
+	s.readMu.RLock()
+	nc := s.readIndexNotifier
+	s.readMu.RUnlock()
+
+	// signal linearizable loop for current notify if it hasn't been already
```
chaochn47, line 867:
```suggestion
	// signal readIndexLoop for current notify if it hasn't been already
```

server/etcdserver/util.go
```diff
@@ -84,6 +84,9 @@ func longestConnected(tp rafthttp.Transporter, membs []types.ID) (types.ID, bool
 type notifier struct {
 	c   chan struct{}
 	err error
+
+	// pass some values in the notifier
+	uint64Val uint64
```
chaochn47, line 89:
```suggestion
	readIndex uint64
```

It's an unexported pkg local struct, I think explicitly naming it is better than generic.

chaochn47:
> @chaochn47 The problem with sharing the same loop is a slow apply loop would block readIndex request when there is a pending read, which makes it not much better than just using linearizableRead. 

Makes sense. 

> So readIndex and linearizableRead has to be in separate loops.

It may not be necessary if `requestCurrentIndex` can be used for read index check while the current server side 7s timeout is too long for the check. This function is likely to be refactored if we go down this route.

tests/e2e/http_health_check_test.go
```diff
@@ -232,16 +233,44 @@ func TestHTTPLivezReadyzHandler(t *testing.T) {
...
+					url:                "/livez",
+					expectedStatusCode: http.StatusOK,
+				},
+				{
+					url:                  "/readyz",
+					expectedTimeoutError: true,
+					expectedStatusCode:   http.StatusServiceUnavailable,
+					expectedRespSubStrings: []string{
+						`[+]serializable_read ok`,
+						`[+]data_corruption ok`,
```
serathius, line 247: We don't need to validate other checks, just `read_index`.

server/etcdserver/v3_server.go
```diff
@@ -799,8 +794,9 @@ func (s *EtcdServer) linearizableReadLoop() {
 		nr := s.readNotifier
 		s.readNotifier = nextnr
 		s.readMu.Unlock()
-
-		confirmedIndex, err := s.requestCurrentIndex(leaderChangedNotifier, requestId)
+		ctx, cancel := context.WithTimeout(context.Background(), s.Cfg.ReqTimeout())
```
serathius, line 797: Why add timeout here?

server/etcdserver/util.go
serathius: notifier was meant to be generic. Not sure if we should add fields here.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -365,3 +367,10 @@ func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
 		return nil
 	}
 }
+
+func readIndexCheck(srv ServerHealth) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		_, err := srv.GetReadIndex(ctx)
```
serathius, line 373: Changing a critical loop like `linearizableReadLoop` seems like a bad idea for backport. What's wrong with directly just calling `requestCurrentIndex`?

serathius: Just need to have a public function.
```
func (s *EtcdServer) RequestCurrentIndex() (uint64, error) {
  requestId := s.reqIDGen.Next()
  leaderChangedNotifier := s.leaderChanged.Receive()
  retun s.requestCurrentIndex(leaderChangedNotifier, requestId)
}
```

What do you think?

serathius: ping @siyuanfoundation

serathius: Oh, I see. `requestCurrentIndex` was designed to be only called from single routine like `linearizableReadLoop`. Nice catch!

Let me think about this more. Need to think on how to minimize the change to `linearizableReadLoop` for a safer backport.
cc @ahrtr for their ideas

PR #16797, opened 2023-10-18
etcdserver: add metric counters for livez/readyz health checks.

siyuanfoundation:

label field consistent with [K8s health metrics](https://github.com/kubernetes/kubernetes/blob/02b0b79eda5a3d41a07a6f2965ee97d492e8b6b2/staging/src/k8s.io/component-base/metrics/prometheus/slis/metrics.go#L35)

### DIFF
--- a/server/etcdserver/api/etcdhttp/health.go
+++ b/server/etcdserver/api/etcdhttp/health.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This file defines the http endpoints for etcd health checks.
+// The endpoints include /livez, /readyz and /health.
+
 package etcdhttp
 
 import (
@@ -34,8 +37,13 @@ import (
 )
 
 const (
-	PathHealth      = "/health"
-	PathProxyHealth = "/proxy/health"
+	PathHealth                 = "/health"
+	PathProxyHealth            = "/proxy/health"
+	HealthStatusSuccess string = "success"
+	HealthStatusError   string = "error"
+	checkTypeLivez             = "livez"
+	checkTypeReadyz            = "readyz"
+	checkTypeHealth            = "health"
 )
 
 type ServerHealth interface {
@@ -111,11 +119,29 @@ var (
 		Name:      "health_failures",
 		Help:      "The total number of failed health checks",
 	})
+	healthCheckGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthcheck",
+		Help:      "The result of each kind of healthcheck.",
+	},
+		[]string{"type", "name"},
+	)
+	healthCheckCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthchecks_total",
+		Help:      "The total number of each kind of healthcheck.",
+	},
+		[]string{"type", "name", "status"},
+	)
 )
 
 func init() {
 	prometheus.MustRegister(healthSuccess)
 	prometheus.MustRegister(healthFailed)
+	prometheus.MustRegister(healthCheckGauge)
+	prometheus.MustRegister(healthCheckCounter)
 }
 
 // Health defines etcd server health status.
@@ -123,6 +149,12 @@ func init() {
 type Health struct {
 	Health string `json:"health"`
 	Reason string `json:"reason"`
+}
+
+// HealthStatus is used in new /readyz or /livez health checks instead of the Health struct.
+type HealthStatus struct {
+	Reason string `json:"reason"`
+	Status string `json:"status"`
 }
 
 func getQuerySet(r *http.Request, query string) StringSet {
@@ -201,18 +233,18 @@ func checkAPI(ctx context.Context, lg *zap.Logger, srv ServerHealth, serializabl
 type HealthCheck func(ctx context.Context) error
 
 type CheckRegistry struct {
-	path   string
-	checks map[string]HealthCheck
+	checkType string
+	checks    map[string]HealthCheck
 }
 
 func installLivezEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
-	reg := CheckRegistry{path: "/livez", checks: make(map[string]HealthCheck)}
+	reg := CheckRegistry{checkType: checkTypeLivez, checks: make(map[string]HealthCheck)}
 	reg.Register("serializable_read", serializableReadCheck(server))
 	reg.InstallHttpEndpoints(lg, mux)
 }
 
 func installReadyzEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
-	reg := CheckRegistry{path: "/readyz", checks: make(map[string]HealthCheck)}
+	reg := CheckRegistry{checkType: checkTypeReadyz, checks: make(map[string]HealthCheck)}
 	reg.Register("data_corruption", activeAlarmCheck(server, pb.AlarmType_CORRUPT))
 	reg.Register("serializable_read", serializableReadCheck(server))
 	reg.InstallHttpEndpoints(lg, mux)
@@ -222,6 +254,10 @@ func (reg *CheckRegistry) Register(name string, check HealthCheck) {
 	reg.checks[name] = check
 }
 
+func (reg *CheckRegistry) RootPath() string {
+	return "/" + reg.checkType
+}
+
 func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMux) {
 	checkNames := make([]string, 0, len(reg.checks))
 	for k := range reg.checks {
@@ -229,19 +265,19 @@ func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMu
 	}
 
 	// installs the http handler for the root path.
-	reg.installRootHttpEndpoint(lg, mux, reg.path, checkNames...)
+	reg.installRootHttpEndpoint(lg, mux, checkNames...)
 	for _, checkName := range checkNames {
 		// installs the http handler for the individual check sub path.
-		subpath := path.Join(reg.path, checkName)
+		subpath := path.Join(reg.RootPath(), checkName)
 		check := checkName
-		mux.Handle(subpath, newHealthHandler(subpath, lg, func(r *http.Request) Health {
+		mux.Handle(subpath, newHealthHandler(subpath, lg, func(r *http.Request) HealthStatus {
 			return reg.runHealthChecks(r.Context(), check)
 		}))
 	}
 }
 
-func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...string) Health {
-	h := Health{Health: "true"}
+func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...string) HealthStatus {
+	h := HealthStatus{Status: HealthStatusSuccess}
 	var individualCheckOutput bytes.Buffer
 	for _, checkName := range checkNames {
 		check, found := reg.checks[checkName]
@@ -250,9 +286,11 @@ func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...str
 		}
 		if err := check(ctx); err != nil {
 			fmt.Fprintf(&individualCheckOutput, "[-]%s failed: %v\n", checkName, err)
-			h.Health = "false"
+			h.Status = HealthStatusError
+			recordMetrics(reg.checkType, checkName, HealthStatusError)
 		} else {
 			fmt.Fprintf(&individualCheckOutput, "[+]%s ok\n", checkName)
+			recordMetrics(reg.checkType, checkName, HealthStatusSuccess)
 		}
 	}
 	h.Reason = individualCheckOutput.String()
@@ -260,19 +298,20 @@ func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...str
 }
 
 // installRootHttpEndpoint installs the http handler for the root path.
-func (reg *CheckRegistry) installRootHttpEndpoint(lg *zap.Logger, mux *http.ServeMux, path string, checks ...string) {
-	hfunc := func(r *http.Request) Health {
+func (reg *CheckRegistry) installRootHttpEndpoint(lg *zap.Logger, mux *http.ServeMux, checks ...string) {
+	hfunc := func(r *http.Request) HealthStatus {
 		// extracts the health check names to be excludeList from the query param
 		excluded := getQuerySet(r, "exclude")
 
 		filteredCheckNames := filterCheckList(lg, listToStringSet(checks), excluded)
-		return reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		h := reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		return h
 	}
-	mux.Handle(path, newHealthHandler(path, lg, hfunc))
+	mux.Handle(reg.RootPath(), newHealthHandler(reg.RootPath(), lg, hfunc))
 }
 
 // newHealthHandler generates a http HandlerFunc for a health check function hfunc.
-func newHealthHandler(path string, lg *zap.Logger, hfunc func(*http.Request) Health) http.HandlerFunc {
+func newHealthHandler(path string, lg *zap.Logger, hfunc func(*http.Request) HealthStatus) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != http.MethodGet {
 			w.Header().Set("Allow", http.MethodGet)
@@ -282,7 +321,7 @@ func newHealthHandler(path string, lg *zap.Logger, hfunc func(*http.Request) Hea
 		}
 		h := hfunc(r)
 		// Always returns detailed reason for failed checks.
-		if h.Health != "true" {
+		if h.Status == HealthStatusError {
 			http.Error(w, h.Reason, http.StatusServiceUnavailable)
 			lg.Error("Health check error", zap.String("path", path), zap.String("reason", h.Reason), zap.Int("status-code", http.StatusServiceUnavailable))
 			return
@@ -340,6 +379,22 @@ func listToStringSet(list []string) StringSet {
 		set[s] = struct{}{}
 	}
 	return set
+}
+
+func recordMetrics(checkType, name string, status string) {
+	val := 0.0
+	if status == HealthStatusSuccess {
+		val = 1.0
+	}
+	healthCheckGauge.With(prometheus.Labels{
+		"type": checkType,
+		"name": name,
+	}).Set(val)
+	healthCheckCounter.With(prometheus.Labels{
+		"type":   checkType,
+		"name":   name,
+		"status": status,
+	}).Inc()
 }
 
 // activeAlarmCheck checks if a specific alarm type is active in the server.
--- a/server/etcdserver/api/etcdhttp/health_test.go
+++ b/server/etcdserver/api/etcdhttp/health_test.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap/zaptest"
 
 	"go.etcd.io/raft/v3"
@@ -193,6 +194,7 @@ func TestHttpSubPath(t *testing.T) {
 			ts := httptest.NewServer(mux)
 			defer ts.Close()
 			checkHttpResponse(t, ts, tt.healthCheckURL, tt.expectStatusCode, tt.inResult, tt.notInResult)
+			checkMetrics(t, tt.healthCheckURL, "", tt.expectStatusCode)
 		})
 	}
 }
@@ -291,6 +293,7 @@ func TestSerializableReadCheck(t *testing.T) {
 			ts := httptest.NewServer(mux)
 			defer ts.Close()
 			checkHttpResponse(t, ts, tt.healthCheckURL, tt.expectStatusCode, tt.inResult, tt.notInResult)
+			checkMetrics(t, tt.healthCheckURL, "serializable_read", tt.expectStatusCode)
 		})
 	}
 }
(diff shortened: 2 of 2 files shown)


server/etcdserver/api/etcdhttp/health.go
```diff
@@ -240,8 +281,36 @@ func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMu
...
 
+func (reg *CheckRegistry) recordMetrics(name string, status string) {
+	val := 0
+	if status == HealthStatusSuccess {
+		val = 1
+	}
+	// record the metrics for the root path.
+	if len(name) == 0 {
+		rootHealthCheckGauge.With(prometheus.Labels{
+			"type": reg.path,
```
chaochn47, line 292: How about stripping the leading forward slash for `type`? `livez` instead of `/livez`?

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -111,18 +113,57 @@ var (
 		Name:      "health_failures",
 		Help:      "The total number of failed health checks",
 	})
```
chaochn47, line 115: I feel people will be confused between `etcd_server_health_success` and `etcd_server_root_health`. 

How about putting deprecation notice on the former one?

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -111,18 +113,57 @@ var (
...
+	},
+		[]string{"type", "status"},
+	)
+	healthCheckGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthcheck",
+		Help:      "This metric records the result of a single healthcheck.",
+	},
+		[]string{"name", "type"},
```
chaochn47, line 138:
```suggestion
		[]string{"type", "name"},
```

`type` should come first and followed by `name`.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -111,18 +113,57 @@ var (
...
+	)
+	healthCheckCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthchecks_total",
+		Help:      "This metric records the total result counts of all healthchecks.",
+	},
+		[]string{"name", "type", "status"},
```
chaochn47, line 146:
```suggestion
		[]string{"type", "name", "status"},
```

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -323,3 +328,66 @@ func checkHttpResponse(t *testing.T, ts *httptest.Server, url string, expectStat
 		}
 	}
 }
+
+func checkMetrics(t *testing.T, path, checkName string, expectStatusCode int) {
+	rootGaugeVal := ptu.ToFloat64(rootHealthCheckGauge.With(prometheus.Labels{
+		"type": path,
+	}))
```
chaochn47, line 335: Is it possible to get the real value from `prometheus.DefaultRegisterer`?

For example: `prometheus.DefaultRegisterer.Gather()`

siyuanfoundation: what do you mean by real value? These are not mocks.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health_test.go
siyuanfoundation: updated.

siyuanfoundation:
cc @logicalhan @serathius

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,71 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
+		Help:      "The total number of failed health checks (Deprecated, use etcd_server_root_health_total instead)",
 	})
+	rootHealthCheckGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "root_health",
+		Help:      "This metric records the result of the root readyz/livez check.",
+	},
+		[]string{"type"},
+	)
```
logicalhan, line 127: Why do we need this root metric? Can't we just compute this from `healthCheckGauge`?

siyuanfoundation: The user would need to know that min of all healthCheckGauge gives you the overall health gauge. That might be inconvenient or error prone.

logicalhan: It's pretty easy to do that in a prometheus query.. That's exactly what labels are meant for.

siyuanfoundation: ok. Removed rootHealthCheckGauge.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,62 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
 	})
+	// Deprecated: use rootHealthCheckCounter instead.
 	healthFailed = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: "etcd",
 		Subsystem: "server",
 		Name:      "health_failures",
-		Help:      "The total number of failed health checks",
+		Help:      "The total number of failed health checks (Deprecated, use etcd_server_root_health_total instead)",
 	})
+	rootHealthCheckCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
```
logicalhan, line 120: same concern here.

siyuanfoundation: I think the accum counter can not be easily be derived from individual checks. The sum of all the checks depends on the number of checks there is.

dgrisonnet: I agree with Han, this is redundant with the other counter.

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -323,3 +327,52 @@ func checkHttpResponse(t *testing.T, ts *httptest.Server, url string, expectStat
...
+		}
+		if path != labelMap["type"] {
+			continue
+		}
+		if labelMap["name"] != checkName {
+			continue
+		}
+		if statusLabel, found := labelMap["status"]; found && statusLabel == HealthStatusError {
+			if val != expectedErrorCount {
+				t.Errorf("%s got errorCount %d, wanted %d\n", name, val, expectedErrorCount)
```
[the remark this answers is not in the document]
siyuanfoundation: done.

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -287,10 +288,13 @@ func TestSerializableReadCheck(t *testing.T) {
 				apiError:  tt.apiError,
 				authStore: auth.NewAuthStore(logger, schema.NewAuthBackend(logger, be), nil, 0),
 			}
+			resetCounters()
```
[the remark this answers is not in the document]
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: moved the root counts to healthCheckCounter with `name="/"`

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -291,6 +292,7 @@ func TestSerializableReadCheck(t *testing.T) {
 			ts := httptest.NewServer(mux)
 			defer ts.Close()
 			checkHttpResponse(t, ts, tt.healthCheckURL, tt.expectStatusCode, tt.inResult, tt.notInResult)
+			checkMetrics(t, tt.healthCheckURL, "serializable_read", tt.expectStatusCode)
```
logicalhan, line 295: should we add a test for any of the subpaths?

logicalhan (approved):
/lgtm
/approve

chaochn47:
```
=== FAIL: integration/clientv3/examples  (0.00s)
2023/10/19 17:53:43 Running tests (examples) in dir(/tmp/etcd-integration1751247847): ...
panic: test timed out after 15m0s
```

It's a known flake. Could you please rebase on top of main or `git commit --amend --no-edit and then push` ? 

@siyuanfoundation

serathius:
> It's a known flake. Could you please rebase on top of main or `git commit --amend --no-edit and then push` ?

Or ping @etcd-io/maintainers-etcd to rerun the test :P

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -241,7 +268,7 @@ func (reg *CheckRegistry) InstallHttpEndpoints(lg *zap.Logger, mux *http.ServeMu
 }
 
 func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...string) Health {
-	h := Health{Health: "true"}
+	h := Health{Status: HealthStatusSuccess}
```
ahrtr, line 271: Let's not to break the existing `/health` endpoint in 3.6.
```suggestion
	h := Health{Health: "true", Status: HealthStatusSuccess}
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -250,9 +277,11 @@ func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...str
 		}
 		if err := check(ctx); err != nil {
 			fmt.Fprintf(&individualCheckOutput, "[-]%s failed: %v\n", checkName, err)
-			h.Health = "false"
```
ahrtr, line 253: ditto, let's keep it so as not to break the existing `/health` in 3.6.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -266,7 +295,9 @@ func (reg *CheckRegistry) installRootHttpEndpoint(lg *zap.Logger, mux *http.Serv
 		excluded := getQuerySet(r, "exclude")
 
 		filteredCheckNames := filterCheckList(lg, listToStringSet(checks), excluded)
-		return reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		h := reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		recordMetrics(reg.path, "/", h.Status)
```
ahrtr, line 299: change the "/" to "default" or "root"? 

Also a related suggestion: can we define a list of enum for all checks? This can be discussed and addressed in a separate session/PR.

```suggestion
		recordMetrics(reg.path, "default", h.Status)
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -82,8 +84,10 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
 		defer func() {
 			if h.Health == "true" {
 				healthSuccess.Inc()
+				recordMetrics("/health", "/", HealthStatusSuccess)
```
ahrtr, line 87:
```suggestion
				recordMetrics("/health", "default", HealthStatusSuccess)
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -82,8 +84,10 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
 			} else {
 				healthFailed.Inc()
+				recordMetrics("/health", "/", HealthStatusError)
```
ahrtr, line 90:
```suggestion
				recordMetrics("/health", "default", HealthStatusError)
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,53 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
 		Subsystem: "server",
 		Name:      "health_failures",
-		Help:      "The total number of failed health checks",
+		Help:      "The total number of failed health checks (Deprecated, use etcd_server_healthchecks_total instead)",
 	})
+	healthCheckGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthcheck",
+		Help:      "This metric records the result of a single healthcheck.",
```
ahrtr, line 124:
```suggestion
		Help:      "The latest result of each kind of healthcheck. 1 is successfully, 0 otherwise",
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,53 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
+	},
+		[]string{"type", "name"},
+	)
+	healthCheckCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthchecks_total",
+		Help:      "This metric records the total result counts of all healthchecks.",
```
ahrtr, line 132:
```suggestion
		Help:      "The total number of each kind of healthcheck.",
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,53 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
 }
 
 // Health defines etcd server health status.
 // TODO: remove manual parsing in etcdctl cluster-health
 type Health struct {
+	// Health field is used in old /health check path.
 	Health string `json:"health"`
 	Reason string `json:"reason"`
+	// Status field is used in new /readyz or /livez health checks instead of the Health field.
+	Status string `json:"-"`
```
ahrtr, line 152: Did you intentionally use "-"?
```suggestion
	Status string `json:"status"`
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,53 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
-		Help:      "The total number of failed health checks",
+		Help:      "The total number of failed health checks (Deprecated, use etcd_server_healthchecks_total instead)",
 	})
+	healthCheckGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthcheck",
+		Help:      "This metric records the result of a single healthcheck.",
+	},
+		[]string{"type", "name"},
```
ahrtr, line 126:
```suggestion
		[]string{"path", "name"},
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +103,53 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
+	)
+	healthCheckCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "etcd",
+		Subsystem: "server",
+		Name:      "healthchecks_total",
+		Help:      "This metric records the total result counts of all healthchecks.",
+	},
+		[]string{"type", "name", "status"},
```
ahrtr, line 134:
```suggestion
		[]string{"path", "name", "status"},
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -342,6 +373,23 @@ func listToStringSet(list []string) StringSet {
 	return set
 }
 
+func recordMetrics(path, name string, status string) {
+	val := 0
+	if status == HealthStatusSuccess {
+		val = 1
+	}
+	path = strings.TrimPrefix(path, "/")
```
ahrtr, line 381:
```suggestion
```

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: this is not used in /health endpoint. It is only used for livez/healthz.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: this is not used in /health endpoint. It is only used for livez/healthz.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: This is ignored because I don't want to change the response of existing /health endpoint. If use `json:"status"`, the response would add an empty status field.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: I used `type` in order to be consistent with [K8s health metrics](https://github.com/kubernetes/kubernetes/blob/02b0b79eda5a3d41a07a6f2965ee97d492e8b6b2/staging/src/k8s.io/component-base/metrics/prometheus/slis/metrics.go#L35)

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: `path` usually starts with `/`, but it is suggested not have the `/` in the counter names.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: changed to root.

server/etcdserver/api/etcdhttp/health.go
ahrtr: Why introduce this field `Status`?  Why not to reuse the existing `Health` to check whether the health check is successful or failed?  If you are planning to remove the fields "Health" and "Reason" and the legacy health check endpoint /health in future release e.g. 4.0, please add a comment.

siyuanfoundation: We still would like to use the `Health` struct. But the `Health` string field is itself not very descriptive and not using constant strings. Added a comment about deprecation.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -99,30 +107,54 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
...
 	prometheus.MustRegister(healthFailed)
+	prometheus.MustRegister(healthCheckGauge)
+	prometheus.MustRegister(healthCheckCounter)
 }
 
 // Health defines etcd server health status.
 // TODO: remove manual parsing in etcdctl cluster-health
 type Health struct {
+	// Health field is used in old /health check path, and it would be deprecated along with the /health endpoint.
+	// To be deprecated in favor of the Status field
```
serathius, line 153: If we plan to drop `Health` then can we not use it in new failpoints?

server/etcdserver/api/etcdhttp/health.go
serathius: Can we have separate `Health` struct for new endpoints?
Old will use `Health` and `Reason` fields, while new will use `Reason` and `Status`.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -342,6 +378,23 @@ func listToStringSet(list []string) StringSet {
 	return set
 }
 
+func recordMetrics(path, name string, status string) {
+	val := 0
+	if status == HealthStatusSuccess {
+		val = 1
+	}
+	path = strings.TrimPrefix(path, "/")
```
serathius, line 386:
```suggestion
	type = strings.TrimPrefix(path, "/")
```

serathius: Can we change the `path` field in `CheckRegistry` into `type` to avoid trimming prefix?

server/etcdserver/api/etcdhttp/health.go
ahrtr: > Old will use `Health` and `Reason` fields, while new will use `Reason` and `Status`.

Sounds good to me.

siyuanfoundation: Added a new `HealthStatus` struct.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -342,6 +386,22 @@ func listToStringSet(list []string) StringSet {
...
 
+func recordMetrics(checkType, name string, status string) {
+	val := 0
+	if status == HealthStatusSuccess {
+		val = 1
+	}
+	healthCheckGauge.With(prometheus.Labels{
+		"type": checkType,
+		"name": name,
+	}).Set(float64(val))
```
serathius, line 397:
```suggestion
	val := 0.0
	if status == HealthStatusSuccess {
		val = 1.0
	}
	healthCheckGauge.With(prometheus.Labels{
		"type": checkType,
		"name": name,
	}).Set(val)
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -250,29 +290,33 @@ func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...str
...
-	hfunc := func(r *http.Request) Health {
+func (reg *CheckRegistry) installRootHttpEndpoint(lg *zap.Logger, mux *http.ServeMux, checks ...string) {
+	hfunc := func(r *http.Request) HealthStatus {
 		// extracts the health check names to be excludeList from the query param
 		excluded := getQuerySet(r, "exclude")
 
 		filteredCheckNames := filterCheckList(lg, listToStringSet(checks), excluded)
-		return reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		h := reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		recordMetrics(reg.checkType, "root", h.Status)
```
serathius, line 312: Why special "root" check name? It's just a sum of other checks, I also didn't find anything like this in Kubernetes code.

siyuanfoundation: It is true the root metrics do not exist in Kubernetes code. 
Imaging I just want to set up an alert based on the overall health status, in theory that could be computed based on the min of all of the other checks, but I could easily make a mistake of using `max` instead of `min`. I think it could be convenient to have a root metric.

serathius: But, if we are following Kubernetes design which expects users to utilize `min` function, why should we expect users to write different queries. Would it be more convenient for users to utilize same query across all components?

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -82,8 +90,10 @@ func NewHealthHandler(lg *zap.Logger, hfunc func(ctx context.Context, excludedAl
 		defer func() {
 			if h.Health == "true" {
 				healthSuccess.Inc()
+				recordMetrics(checkTypeHealth, "root", HealthStatusSuccess)
```
serathius, line 93: I'm not sure about support of old health endpoint.

/health endpoint supports parameters that influence the results. This can mean that if there is 2 probers using different parameters (one using health for liveness, second for readiness) it will lead to metric flipping between 0 and 1 without any reason. This is the flaw of the old health probe that we shouldn't inherit.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -250,29 +286,32 @@ func (reg *CheckRegistry) runHealthChecks(ctx context.Context, checkNames ...str
...
+		h := reg.runHealthChecks(r.Context(), filteredCheckNames...)
+		return h
 	}
-	mux.Handle(path, newHealthHandler(path, lg, hfunc))
+	mux.Handle(reg.RootPath(), newHealthHandler(reg.RootPath(), lg, hfunc))
 }
 
 // newHealthHandler generates a http HandlerFunc for a health check function hfunc.
-func newHealthHandler(path string, lg *zap.Logger, hfunc func(*http.Request) Health) http.HandlerFunc {
+func newHealthHandler(path string, lg *zap.Logger, hfunc func(*http.Request) HealthStatus) http.HandlerFunc {
```
serathius, line 314: Should `path` be replaced with `checkType` ?

serathius (approved):

server/etcdserver/api/etcdhttp/health.go
ahrtr: The `path` seems better here, because it may be 
- a root path something like `/livez` or `/readyz`;
- a subpath something like `/livez/serializable_read`.

ahrtr (approved):
lgtm

Thank you!

server/etcdserver/api/etcdhttp/health.go
serathius: Paths are useful for HTTP access level logging,

PR #16803, opened 2023-10-21
clientv3: allow setting JWT directly

mcrute:
etcd supports using signed JWTs in a verify-only mode where the server has access to only a public key and therefore can not create tokens but can validate them. For this to work a client must not call Authenticate and must instead submit a pre-signed JWT with their request. The server will validate this token, extract the username from it, and may allow the client access.

This change allows setting the JWT directly and not setting a username and password. If a JWT is provided the client will no longer call Authenticate, which would not work anyhow. It also provides a public method UpdateAuthToken to allow a user of the client to update their auth token, for example, if it expires.

In this flow all token lifecycle management is handled outside of the client as a concern of the client user.

### DIFF
--- a/client/v3/client.go
+++ b/client/v3/client.go
@@ -69,7 +69,10 @@ type Client struct {
 	// Username is a user name for authentication.
 	Username string
 	// Password is a password for authentication.
-	Password        string
+	Password string
+	// Token is a JWT used for authentication instead of a password.
+	Token string
+
 	authTokenBundle credentials.PerRPCCredentialsBundle
 
 	callOpts []grpc.CallOption
@@ -288,6 +291,11 @@ func (c *Client) Dial(ep string) (*grpc.ClientConn, error) {
 func (c *Client) getToken(ctx context.Context) error {
 	var err error // return last error in a case of fail
 
+	if c.Token != "" {
+		c.authTokenBundle.UpdateAuthToken(c.Token)
+		return nil
+	}
+
 	if c.Username == "" || c.Password == "" {
 		return nil
 	}
@@ -376,6 +384,10 @@ func newClient(cfg *Config) (*Client, error) {
 		creds = credentials.NewTransportCredential(cfg.TLS)
 	}
 
+	if cfg.Token != "" && (cfg.Username != "" || cfg.Password != "") {
+		return nil, errors.New("Username/Password and Token configurations are mutually exclusive")
+	}
+
 	// use a temporary skeleton client to bootstrap first connection
 	baseCtx := context.TODO()
 	if cfg.Context != nil {
@@ -414,6 +426,12 @@ func newClient(cfg *Config) (*Client, error) {
 		client.Password = cfg.Password
 		client.authTokenBundle = credentials.NewPerRPCCredentialBundle()
 	}
+
+	if cfg.Token != "" {
+		client.Token = cfg.Token
+		client.authTokenBundle = credentials.NewPerRPCCredentialBundle()
+	}
+
 	if cfg.MaxCallSendMsgSize > 0 || cfg.MaxCallRecvMsgSize > 0 {
 		if cfg.MaxCallRecvMsgSize > 0 && cfg.MaxCallSendMsgSize > cfg.MaxCallRecvMsgSize {
 			return nil, fmt.Errorf("gRPC message recv limit (%d bytes) must be greater than send limit (%d bytes)", cfg.MaxCallRecvMsgSize, cfg.MaxCallSendMsgSize)
--- a/client/v3/config.go
+++ b/client/v3/config.go
@@ -66,6 +66,9 @@ type Config struct {
 	// Password is a password for authentication.
 	Password string `json:"password"`
 
+	// Token is a JWT used for authentication instead of a password.
+	Token string `json:"token"`
+
 	// RejectOldCluster when set will refuse to create a client against an outdated cluster.
 	RejectOldCluster bool `json:"reject-old-cluster"`
 
@@ -130,6 +133,7 @@ type SecureConfig struct {
 type AuthConfig struct {
 	Username string `json:"username"`
 	Password string `json:"password"`
+	Token    string `json:"token"`
 }
 
 func (cs *ConfigSpec) Clone() *ConfigSpec {
@@ -157,7 +161,7 @@ func (cs *ConfigSpec) Clone() *ConfigSpec {
 }
 
 func (cfg AuthConfig) Empty() bool {
-	return cfg.Username == "" && cfg.Password == ""
+	return cfg.Username == "" && cfg.Password == "" && cfg.Token == ""
 }
 
 // NewClientConfig creates a Config based on the provided ConfigSpec.
@@ -180,6 +184,7 @@ func NewClientConfig(confSpec *ConfigSpec, lg *zap.Logger) (*Config, error) {
 	if confSpec.Auth != nil {
 		cfg.Username = confSpec.Auth.Username
 		cfg.Password = confSpec.Auth.Password
+		cfg.Token = confSpec.Auth.Token
 	}
 
 	return cfg, nil
--- a/etcdctl/ctlv3/ctl.go
+++ b/etcdctl/ctlv3/ctl.go
@@ -70,6 +70,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&globalFlags.TLS.CertFile, "cert", "", "identify secure client using this TLS certificate file")
 	rootCmd.PersistentFlags().StringVar(&globalFlags.TLS.KeyFile, "key", "", "identify secure client using this TLS key file")
 	rootCmd.PersistentFlags().StringVar(&globalFlags.TLS.TrustedCAFile, "cacert", "", "verify certificates of TLS-enabled secure servers using this CA bundle")
+	rootCmd.PersistentFlags().StringVar(&globalFlags.Token, "auth-jwt-token", "", "JWT token used for authentication (if this option is used, --user and --password should not be set)")
 	rootCmd.PersistentFlags().StringVar(&globalFlags.User, "user", "", "username[:password] for authentication (prompt if password is not supplied)")
 	rootCmd.PersistentFlags().StringVar(&globalFlags.Password, "password", "", "password for authentication (if this option is used, --user option shouldn't include password)")
 	rootCmd.PersistentFlags().StringVarP(&globalFlags.TLS.ServerName, "discovery-srv", "d", "", "domain name to query for SRV records describing cluster endpoints")
--- a/tests/common/auth_util.go
+++ b/tests/common/auth_util.go
@@ -17,8 +17,11 @@ package common
 import (
 	"context"
 	"fmt"
+	"os"
 	"testing"
+	"time"
 
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/stretchr/testify/require"
 
 	"go.etcd.io/etcd/api/v3/authpb"
@@ -93,6 +96,29 @@ func createUsers(c interfaces.Client, users []authUser) error {
 	return nil
 }
 
+func createSignedJWT(keyPath, alg, username string, authRevision uint64) (string, error) {
+	signMethod := jwt.GetSigningMethod(alg)
+
+	keyBytes, err := os.ReadFile(keyPath)
+	if err != nil {
+		return "", err
+	}
+
+	key, err := jwt.ParseRSAPrivateKeyFromPEM(keyBytes)
+	if err != nil {
+		return "", err
+	}
+
+	tk := jwt.NewWithClaims(signMethod,
+		jwt.MapClaims{
+			"username": username,
+			"revision": authRevision,
+			"exp":      time.Now().Add(time.Minute).Unix(),
+		})
+
+	return tk.SignedString(key)
+}
+
 func setupAuth(c interfaces.Client, roles []authRole, users []authUser) error {
 	// create roles
 	if err := createRoles(c, roles); err != nil {
@@ -105,6 +131,29 @@ func setupAuth(c interfaces.Client, roles []authRole, users []authUser) error {
 
 	// enable auth
 	return c.AuthEnable(context.TODO())
+}
+
+func setupAuthAndGetRevision(c interfaces.Client, roles []authRole, users []authUser) (uint64, error) {
+	// create roles
+	if err := createRoles(c, roles); err != nil {
+		return 0, err
+	}
+
+	if err := createUsers(c, users); err != nil {
+		return 0, err
+	}
+
+	// This needs to happen before enabling auth for the TestAuthJWTOnly
+	// test case because once auth is enabled we can no longer mint a valid
+	// auth token without the revision, which we won't be able to obtain
+	// without a valid auth token.
+	authrev, err := c.AuthStatus(context.TODO())
+	if err != nil {
+		return 0, err
+	}
+
+	// enable auth
+	return authrev.AuthRevision, c.AuthEnable(context.TODO())
 }
 
 func requireRolePermissionEqual(t *testing.T, expectRole authRole, actual []*authpb.Permission) {
--- a/client/v3/client_test.go
+++ b/client/v3/client_test.go
@@ -317,6 +317,75 @@ func TestAuthTokenBundleNoOverwrite(t *testing.T) {
 	}
 }
 
+func TestNewWithOnlyJWT(t *testing.T) {
+	// This call in particular changes working directory to the tmp dir of
+	// the test. The `etcd-auth-test:1` can be created in local directory,
+	// not exceeding the longest allowed path on OsX.
+	testutil.BeforeTest(t)
+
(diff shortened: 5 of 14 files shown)


mcrute:
I also added support for passing a token rather than a username and password in etcdctl. I can split this into a separate PR that depends on this one if needed; leaving it here for now though since it's a relatively small change.

mcrute:
The failing test here appears to be unrelated to this change:

```
2023-10-22T08:17:25.0015167Z /__w/etcd/etcd/bin/etcd (TestWatchQuorumLastVersion-test-2) (54187): {"level":"error","ts":"2023-10-22T08:17:24.98355Z","caller":"embed/etcd.go:855","msg":"setting up serving from embedded etcd failed.","error":"accept tcp 127.0.0.1:20011: use of closed network connection","stacktrace":"go.etcd.io/etcd/server/v3/embed.(*Etcd).errHandler\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:855\ngo.etcd.io/etcd/server/v3/embed.(*Etcd).servePeers.func3\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:607"}
```

jmhbnz (changes requested):
Hey @mcrute - Thanks for raising this idea.

Overall at first glance this seems reasonable.  Can you please add a test for this new direct jwt authentication so we can ensure it doesn't break in future?  Thanks!

mcrute:
Hi @jmhbnz, sure thing! I've added some tests for the functionality. Let me know if there's anything else needed to get this merged.

jmhbnz:
Thanks @mcrute - Quick heads up the etcd team are pretty busy at kubecon this week so there might be a minor delay getting this new feature reviewed. I will take another look when I get some quiet focus time hopefully end of the week 🙏🏻

mcrute:
Thanks @jmhbnz, I appreciate it. Would really like to get this merged so let me know if there's anything I can do to help move it along. Have a nice KubeCon!

jmhbnz (approved):
LGTM - Thanks @mcrute.

Would you mind please rebasing this pr onto latest changes in `main` as I notice `make verify` is failing on this branch (for an unrelated issue I believe which has since been fixed in `main`.).

mcrute:
Thanks for the review, I've rebased main under this so hopefully we'll get a green test run.

mcrute:
Looking at the failing test. I don't see how it's immediately related to this change but it's an auth test so I'm suspicious.

mcrute:
Found the issue. Unconditionally setting `authTokenBundle` to something non-nil breaks the no-auth tests. Revised to fix this.

jmhbnz:
Gentle ping @serathius & @ahrtr, I think this one is ready for maintainer review.

@mcrute If we go ahead with this we will probably want to update some documentation on github.com/etcd-io/website , perhaps under:
- https://etcd.io/docs/v3.5/op-guide/authentication/authentication
- https://etcd.io/docs/v3.5/learning/design-auth-v3

mcrute:
Once this is merged I'll create a PR for the docs. I learned a ton digging through this functionality and definitely would like to make sure others can use it with less effort required.

client/v3/client.go
```diff
@@ -68,7 +68,10 @@ type Client struct {
 	// Username is a user name for authentication.
 	Username string
 	// Password is a password for authentication.
-	Password        string
+	Password string
+	// Token is a JWT used for authentication instead of a password.
+	Token string
```
ahrtr, line 73: We should add a validation on the config, only the following three cases are allowed,
- Only username & password are specified;
- Only token is specified;
- Neither is specified.

If both username & password and token are specified, then we should raise an error or warning.

ahrtr:
I think we need to add an e2e test as well.

mcrute:
Have had some time constraints but I'll post an update soon. Adding validation + test case for credential mixes as well as an e2e test validating that standalone token auth works.

jmhbnz:
Discussed during sig-etcd triage meeting @mcrute do you have time to rebase this pr and add the e2e test? Would be great to get this merged. Thanks!

jmhbnz:
Discussed during sig-etcd triage meeting @mcrute do you have time to rebase this pr and add the e2e test? Would be great to get this merged. Thanks!

mcrute:
@jmhbnz I'm planning to pick this back up this week, have been distracted by other priorities but would very much like to get this finished and merged.

mcrute:
@jmhbnz sorry for the delay on this but I've now added an e2e test for this functionality. Let me know if there's anything else needed to get this merged. Thanks!

client/v3/client.go
mcrute: Added validation for this in the `New` path as well as a unit test.

mcrute:
Ping @serathius & @ahrtr. Do you have a moment to review (and hopefully merge) this PR? I think that it's complete and incorporates all of the feedback.

bill-of-materials.json
```diff
@@ -143,6 +143,15 @@
 			}
 		]
 	},
+	{
+		"project": "github.com/golang-jwt/jwt",
```
ahrtr, line 147: etcd already uses jwt/v4, please do not introduce out of date version.
https://github.com/etcd-io/etcd/blob/9d8000f374f68de6e4c6af97016be54b592f33dd/bill-of-materials.json#L147

client/v3/client.go
```diff
@@ -285,22 +288,34 @@ func (c *Client) Dial(ep string) (*grpc.ClientConn, error) {
 	return c.dial(creds, grpc.WithResolvers(resolver.New(ep)))
 }
 
+// UpdateAuthToken allows updating the JWT auth token held by the
+// client. It is safe to call this function concurrently with other
+// operations.
+func (c *Client) UpdateAuthToken(token string) {
+	c.authTokenBundle.UpdateAuthToken(token)
+}
```
ahrtr, line 296: I suggest not to add an exported/public method.
```suggestion
```

client/v3/client.go
```diff
@@ -376,6 +391,10 @@ func newClient(cfg *Config) (*Client, error) {
 		creds = credentials.NewTransportCredential(cfg.TLS)
 	}
 
+	if cfg.Token != "" && (cfg.Username != "" || cfg.Password != "") {
+		return nil, fmt.Errorf("Username/Password and Token configurations are mutually exclusive")
```
ahrtr, line 395:
```suggestion
		return nil, errors.New("Username/Password and Token configurations are mutually exclusive")
```

etcdctl/ctlv3/ctl.go
```diff
@@ -71,6 +71,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&globalFlags.TLS.KeyFile, "key", "", "identify secure client using this TLS key file")
 	rootCmd.PersistentFlags().StringVar(&globalFlags.TLS.TrustedCAFile, "cacert", "", "verify certificates of TLS-enabled secure servers using this CA bundle")
 	rootCmd.PersistentFlags().StringVar(&globalFlags.User, "user", "", "username[:password] for authentication (prompt if password is not supplied)")
+	rootCmd.PersistentFlags().StringVar(&globalFlags.Token, "auth-token", "", "JWT token used for authentication (if this option is used, --user and --password should not be set)")
```
ahrtr, line 74:
```suggestion
	rootCmd.PersistentFlags().StringVar(&globalFlags.Token, "auth-jwt-token", "", "JWT token used for authentication (if this option is used, --user and --password should not be set)")
```

ahrtr:
> For this to work a client must not call Authenticate

What will happen if such a client call Authenticate?

mcrute:
> > For this to work a client must not call Authenticate
> 
> What will happen if such a client call Authenticate?

There will be no change in behavior from the current state of the code. If the server has only a public key it will fail with an error because it will not be able to `assign` a token to the user. If it has a private key then it will return a new token to the user that they will need to assign to the client themselves.

Although my reading of the code is that this would be a pretty odd case for someone to do this since `Authenticate` is called automatically in the client `getToken` method. It is also not called if the client was configured with a `Token`.

bill-of-materials.json
mcrute: Oops. It looks like I had an incorrect import in one of my test cases. Resolved and rolled back the BOM changes.

client/v3/client.go
mcrute: Revised

etcdctl/ctlv3/ctl.go
mcrute: Revised.

client/v3/client.go
mcrute: I exposed this publicly because it's the only way to update an auth token directly in the client (since `authTokenBundle` is not visible to consumers of the client).

The use-case for exposing this is that a consumer of the client may be managing token lifecycle entirely outside of etcd and issue a replacement token for an expired token out-of-band from the client and the consumer of the client will need to call `UpdateAuthToken` to replace the expired token with the new one.

mcrute:
Rebasing this was a terrible idea 😭. Undoing that bad decision and hopefully these new, unrelated, tests failures disappear.

tests/common/auth_util.go
```diff
@@ -93,6 +96,29 @@ func createUsers(c interfaces.Client, users []authUser) error {
 	return nil
 }
 
+func createSignedJWT(keyPath, username string, authRevision uint64) (string, error) {
+	signMethod := jwt.GetSigningMethod("RS256")
```
ahrtr, line 100:
```suggestion
func createSignedJWT(keyPath, alg, username string, authRevision uint64) (string, error) {
	if alg == "" {
		alg = "RS256"
	}
	signMethod := jwt.GetSigningMethod(alg)
```

mcrute:
> > Rebasing this was a terrible idea
> 
> Please rebase this PR, because it is based on a commit of 3 weeks ago. We ran into conflict which wasn't detected by github.

Once I have a clean run of all the tests I'll rebase. The first rebase introduced a lot of seemingly spurious failures and I would like to make sure I've got completely working and tested code before digging into those.

> Two generic comments/questions,
> 
> * Do you have a real or strong production use case for this change?

Yes, we use some code similar to this for application configuration and secret bootstrapping. The application runner injects a pre-signed etcd token into the application environment when it starts then a library within the application discovers etcd (using DNS SRV records), connects with the injected token, and reads its configuration and secrets from etcd. There is a separate controller that handles token lifetimes by creating new tokens and passing them to applications over a pre-defined key in etcd so the application can update the token for their etcd connection before expiration (it's a little more complicated than that but that's the general workflow).

> * Users need to manage the token themselves. In that case, etcd (client sdk) should never automatically retry when it gets an invalid auth failure response. But unfortunately we don't support disabling the retry for now. Refer to the "Proposal" section in [etcd Go & Java client SDK's retry mechanism may break `Serializable` #18424](https://github.com/etcd-io/etcd/issues/18424).

At least for our use-case that should be okay. We swap tokens well in advance of their expiration. Although having this functionality would be a good safety measure.

mcrute:
Looks like the build was clean. Going to push the rebase now for testing.

mcrute:
@ahrtr all right, I've got all the requested changes incorporated and we've got a clean build with the rebase! Let me know if there's anything else you'd like to see or if we're good to merge this. Thanks.

ahrtr:
> There is a separate controller that handles token lifetimes by creating new tokens and passing them to applications over a pre-defined key in etcd so the application can update the token for their etcd connection before expiration

If the application doesn't update the token before expiration in time for whatever reason, then you won't have any chance to know the `authRevision` anymore (of course you can get it by hack way) because you don't have a valid token; but you won't be able to generate a new JWT token because you don't know the latest `authRevision`. You will run into a deadlock.

One possible solution is to loosen the access control on AuthStatus so that you can always get the latest authRevision no matter whether or not you have a valid token, https://github.com/etcd-io/etcd/blob/3aa56a76bd7cda7a573bdc858af705810466dfb6/client/v3/auth.go#L70

Refer to,
https://github.com/etcd-io/etcd/blob/3aa56a76bd7cda7a573bdc858af705810466dfb6/server/etcdserver/apply/apply_auth.go#L175-L176

Also we need to update the doc to clarify the use case/workflow for this change.

client/v3/client.go
ahrtr: > Given that use-case can we leave this exported?

OK.

mcrute:
Just checking back in on this. I'm open to removing loosening the access control on `AuthStatus` if you think that's a valid approach. I don't think it would reveal anything security sensitive other than possibly the timing of auth store changes. If you're +1 on this idea then I'll implement it as another commit on this PR.

I will also update the docs but since those are in a different repo I was planning to do that after this was merged. Do you want me to open a PR against the docs as well and cross-link them?

I'd really like to get this shipped in September, if possible, so we can stop vendoring a fork of this library.

ahrtr:
Overall this is a valid & minor change (only client side change) to me. Followups,

- Loosen the access control on AuthStatus as mentioned in https://github.com/etcd-io/etcd/pull/16803#issuecomment-2290808401. Personally I think that we should loosen it even without this PR. The purpose of `AuthStatus` is to check whether auth is enabled or not, it shouldn't require a valid token to access it, otherwise it's a chicken-egg issue.
- Once the JWT token is managed by users themselves, then it's out of etcd's control, so the client (SDK) should never automatically retry when it gets an invalid auth failure response as mentioned in https://github.com/etcd-io/etcd/pull/16803#issuecomment-2286260521. This can be resolved separately. We just need to expose an API to let users to disable the retry mechanism.
- Update doc (yes, please raise a separate ticket once we have a consensus)

It's hard to track all these tasks across PRs. Could you raise a ticket to track them?

cc @serathius @jmhbnz @ivanvc

client/v3/client.go
```diff
@@ -376,6 +384,10 @@ func newClient(cfg *Config) (*Client, error) {
 		creds = credentials.NewTransportCredential(cfg.TLS)
 	}
 
+	if cfg.Token != "" && (cfg.Username != "" || cfg.Password != "") {
+		return nil, errors.New("Username/Password and Token configurations are mutually exclusive")
```
ivanvc, line 388: Minor observation/question: thinking about the work we have been doing with errors, does setting this error as an exported variable make sense? Would it be valuable? WDYT @ahrtr?

client/v3/config.go
```diff
@@ -66,6 +66,9 @@ type Config struct {
 	// Password is a password for authentication.
 	Password string `json:"password"`
 
+	// Token is a JWT used for authentication instead of a password.
+	Token string `json:"token"`
```
ivanvc, line 70: Given that this will be exported and it may be difficult to change the name later on, I feel like "Token" may be too generic. Would "JWT" be a better name (more implementation specific)?

client/v3/client.go
ahrtr: Sounds like a good suggestion.

client/v3/config.go
ahrtr: Theoretically, it can be any token types that are supported by etcd. The etcdserver just delegates to the related token prover to parse the token based on the token type configured in `--auth-token`.

But in practice, it's JWT for now. So suggest to keep using `Token`

client/v3/client.go
ivanvc: Given the number of follow-ups we have for this pull request, and the time it has been open, I think it's okay to do this in a follow-up PR.

ivanvc (approved):
LGTM. Thanks, Mike.

mcrute:
@ivanvc I just rebased this and it's an approved PR, could you please merge this? Thanks!

mcrute:
Ping @jmhbnz as well since you approved this PR. Could you please merge this before it goes stale again? Thanks!

ahrtr:
Please raise an issue to track this feature so that the followups won't be forgotten. 

https://github.com/etcd-io/etcd/pull/16803#issuecomment-2359117749

mcrute:
@ahrtr done. I've opened feature requests for all of the main requests pointed out above. I'll start working through those once this is merged. Thanks!

ahrtr:
> @ahrtr done. I've opened feature requests issues for everything above. I'll start working through those once this is merged. Thanks!

Thanks. It'd to have a parent issue and link to all the child issues.

ahrtr (approved):
LGTM

thx

cc @fuweid @serathius

serathius:
Needs rebase

mcrute:
Failing test seems like a flake:

/test pull-etcd-integration-2-cpu-arm64

mcrute:
This test also appears to be a flake:

/test pull-etcd-e2e-amd64

mcrute:
@fuweid @serathius @ahrtr @jmhbnz pinging everyone that's mentioned with commit access. Is there some other magic prow incantation needed to merge this? Like `/lgtm`? It looks like we have all the approvers that have approved but it's still not being merged. Please help! Thanks.

serathius (approved):

client/v3/client.go
ahrtr: It seems that `(c *Client) UpdateAuthToken(token string)` was removed from this PR?

We should need it, otherwise there is no way for users to update the token. Without this method, users will have to create a new etcd client each time when they wanna update the token.

PR #12531, opened 2020-12-08
etcdserver: ignore old leader's request to revoke lease

tangcong:

Issue #16292, opened 2023-07-24
Evaluate and (Gradulate or deprecate/remove) experimental features


ahrtr:
### What would you like to be added?

There are lots of experimental features/flags, but they haven't been promoted to stable features, not deprecated for a long time. We should evaluate them one by one, and then (graduate or deprecate/remove) them.

### Why is this needed?

graduate or deprecate/remove experimental features

PR #16822, opened 2023-10-24
Ignore old leader's leases revoking request

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/15247.
- The first commit prevents an old leader from revoking any leases, similar to https://github.com/etcd-io/etcd/pull/12531.
- The second commit prevents the clients from getting stuck on an unhealthy member (i.e. an old leader)

Also refer to https://docs.google.com/document/d/1peLDjwebnSuNR69ombuUiJ5fkSw7Rd3Slv_mUpOLsmg/edit#heading=h.jbaw5ynuj9i7

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -394,8 +394,15 @@ func NewServer(cfg config.ServerConfig) (srv *EtcdServer, err error) {
 
 	if srv.Cfg.EnableLeaseCheckpoint {
 		// setting checkpointer enables lease checkpoint feature.
-		srv.lessor.SetCheckpointer(func(ctx context.Context, cp *pb.LeaseCheckpointRequest) {
+		srv.lessor.SetCheckpointer(func(ctx context.Context, cp *pb.LeaseCheckpointRequest) error {
+			if !srv.ensureLeadership() {
+				srv.lg.Warn("Ignore the checkpoint request because current member isn't a leader",
+					zap.Uint64("local-member-id", uint64(srv.MemberId())))
+				return lease.ErrNotPrimary
+			}
+
 			srv.raftRequestOnce(ctx, pb.InternalRaftRequest{LeaseCheckpoint: cp})
+			return nil
 		})
 	}
 
@@ -844,7 +851,19 @@ func (s *EtcdServer) run() {
 
 func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
 	s.GoAttach(func() {
+		// We shouldn't revoke any leases if current member isn't a leader,
+		// because the operation should only be performed by the leader. When
+		// the leader gets blocked on the raft loop, such as writing WAL entries,
+		// it can't process any events or messages from raft. It may think it
+		// is still the leader even the leader has already changed.
+		// Refer to https://github.com/etcd-io/etcd/issues/15247
 		lg := s.Logger()
+		if !s.ensureLeadership() {
+			lg.Warn("Ignore the lease revoking request because current member isn't a leader",
+				zap.Uint64("local-member-id", uint64(s.MemberId())))
+			return
+		}
+
 		// Increases throughput of expired leases deletion process through parallelization
 		c := make(chan struct{}, maxPendingRevokes)
 		for _, curLease := range leases {
@@ -875,6 +894,29 @@ func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
 			f(int64(curLease.ID))
 		}
 	})
+}
+
+// ensureLeadership checks whether current member is still the leader.
+func (s *EtcdServer) ensureLeadership() bool {
+	lg := s.Logger()
+
+	ctx, cancel := context.WithTimeout(s.ctx, s.Cfg.ReqTimeout())
+	defer cancel()
+	if err := s.linearizableReadNotify(ctx); err != nil {
+		lg.Warn("Failed to check current member's leadership",
+			zap.Error(err))
+		return false
+	}
+
+	newLeaderId := s.raftStatus().Lead
+	if newLeaderId != uint64(s.MemberId()) {
+		lg.Warn("Current member isn't a leader",
+			zap.Uint64("local-member-id", uint64(s.MemberId())),
+			zap.Uint64("new-lead", newLeaderId))
+		return false
+	}
+
+	return true
 }
 
 // Cleanup removes allocated objects by EtcdServer.NewServer in
--- a/tests/e2e/v3_lease_no_proxy_test.go
+++ b/tests/e2e/v3_lease_no_proxy_test.go
@@ -1,0 +1,160 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+// TestLeaseRevokeByOldLeader verifies that leases shouldn't be revoked by
+// old leader.
+// See the first case in the section "Root Cause" in
+// https://docs.google.com/document/d/1peLDjwebnSuNR69ombuUiJ5fkSw7Rd3Slv_mUpOLsmg/edit
+func TestLeaseRevokeByOldLeader(t *testing.T) {
+	testLeaseRevokeIssue(t, func(cluster *e2e.EtcdProcessCluster, leaderIndex int) []string {
+		// return the endpoint of a follower
+		return cluster.Procs[(leaderIndex+1)%len(cluster.Procs)].EndpointsGRPC()
+	})
+}
+
+// TestLeaseRevokeByOldLeader verifies that leases shouldn't be revoked by
+// new leader.
+// See the second case in the section "Root Cause" in
+// https://docs.google.com/document/d/1peLDjwebnSuNR69ombuUiJ5fkSw7Rd3Slv_mUpOLsmg/edit
+func TestLeaseRevokeByNewLeader(t *testing.T) {
+	testLeaseRevokeIssue(t, func(cluster *e2e.EtcdProcessCluster, leaderIndex int) []string {
+		// return all members' endpoints, so that the client can switch to
+		// other member in case the leader gets stuck on writing disk.
+		return cluster.EndpointsGRPC()
+	})
+}
+
+func testLeaseRevokeIssue(t *testing.T, epsFn func(cluster *e2e.EtcdProcessCluster, leaderIndex int) []string) {
+	e2e.BeforeTest(t)
+
+	ctx := context.Background()
+
+	t.Log("Starting a new etcd cluster")
+	epc, err := e2e.NewEtcdProcessCluster(ctx, t,
+		e2e.WithClusterSize(3),
+		e2e.WithGoFailEnabled(true),
+		e2e.WithGoFailClientTimeout(40*time.Second),
+	)
+	require.NoError(t, err)
+	defer func() {
+		if errC := epc.Close(); errC != nil {
+			t.Fatalf("error closing etcd processes (%v)", errC)
+		}
+	}()
+
+	leaderIdx := epc.WaitLeader(t)
+	t.Logf("Leader index: %d", leaderIdx)
+
+	epsForNormalOperations := epc.Procs[(leaderIdx+2)%3].EndpointsGRPC()
+	t.Logf("Creating a client for normal operations: %v", epsForNormalOperations)
+	client, err := clientv3.New(clientv3.Config{Endpoints: epsForNormalOperations, DialTimeout: 3 * time.Second})
+	require.NoError(t, err)
+	defer client.Close()
+
+	epsForLeaseKeepAlive := epsFn(epc, leaderIdx)
+	t.Logf("Creating a client for the leaseKeepAlive operation: %v", epsForLeaseKeepAlive)
+	clientForKeepAlive, err := clientv3.New(clientv3.Config{Endpoints: epsForLeaseKeepAlive, DialTimeout: 3 * time.Second})
+	require.NoError(t, err)
+	defer clientForKeepAlive.Close()
+
+	resp, err := client.Status(ctx, epsForNormalOperations[0])
+	require.NoError(t, err)
+	oldLeaderId := resp.Leader
+
+	t.Log("Creating a new lease")
+	leaseRsp, err := client.Grant(ctx, 20)
+	require.NoError(t, err)
+
+	t.Log("Starting a goroutine to keep alive the lease")
+	doneC := make(chan struct{})
+	stopC := make(chan struct{})
+	startC := make(chan struct{}, 1)
+	go func() {
+		defer close(doneC)
+
+		respC, kerr := clientForKeepAlive.KeepAlive(ctx, leaseRsp.ID)
+		require.NoError(t, kerr)
+		// ensure we have received the first response from the server
+		<-respC
+		startC <- struct{}{}
+
+		for {
+			select {
+			case <-stopC:
+				return
+			case <-respC:
+			}
+		}
+	}()
+
+	t.Log("Wait for the keepAlive goroutine to get started")
+	<-startC
+
+	t.Log("Trigger the failpoint to simulate stalled writing")
+	err = epc.Procs[leaderIdx].Failpoints().SetupHTTP(ctx, "raftBeforeSave", `sleep("30s")`)
+	require.NoError(t, err)
+
+	cctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	t.Logf("Waiting for a new leader to be elected, old leader index: %d, old leader ID: %d", leaderIdx, oldLeaderId)
+	testutils.ExecuteUntil(cctx, t, func() {
+		for {
+			resp, err = client.Status(ctx, epsForNormalOperations[0])
+			if err == nil && resp.Leader != oldLeaderId {
+				t.Logf("A new leader has already been elected, new leader index: %d", resp.Leader)
+				return
+			}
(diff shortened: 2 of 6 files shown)


server/etcdserver/server.go
```diff
@@ -858,8 +858,20 @@ func (s *EtcdServer) run() {
 }
 
 func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
+	lg := s.Logger()
+	newLeaderId := s.raftStatus().Lead
```
serathius, line 862: Is `s.raftStatus().Lead` updated if the raft loop is blocked? I was thinking that we need to make a quorum read here to confirm leadership.

ahrtr: > Is s.raftStatus().Lead updated if the raft loop is blocked? 

No, because it creates a separate channel to receive the response (raftStatus). FYI. https://github.com/etcd-io/raft/blob/7c3ed830bbb092753012c12fca1b7d4d29095552/node.go#L570

> I was thinking that we need to make a quorum read here to confirm leadership.

Already added `linearizableReadNotify` check. The new method `ensureLeadership` should be enough to guarantee the current member's leadership.

serathius: > Is s.raftStatus().Lead updated if the raft loop is blocked?
>
> No, because it creates a separate channel to receive the response (raftStatus). FYI. https://github.com/etcd-io/raft/blob/7c3ed830bbb092753012c12fca1b7d4d29095552/node.go#L570

Ok, raftStatus is not blocked. But will it serve latest raft state or will it be outdated?

ahrtr: > But will it serve latest raft state or will it be outdated?

Pls see line 906, `linearizableReadNotify` guarantees that the raft state is latest.

serathius: Ok, was just confirming how `raftStatus()` works on it's own. Possibly we should consider looking through all calls to `raftStatus` and double checking if they will benefit from `linearizableReadNotify` before it.

ahrtr: > Possibly we should consider looking through all calls to `raftStatus` and double checking if they will benefit from `linearizableReadNotify` before it.

It's a good point. Pls see https://github.com/etcd-io/etcd/issues/15944#issuecomment-1778860437

server/etcdserver/server.go
```diff
@@ -892,6 +895,31 @@ func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
...
+// performing the leases revoking operations. It can mitigate the issue
+// https://github.com/etcd-io/etcd/issues/15247, but it can't completely
+// resolve it because the leases may be wrongly revoked by the new leader.
+// Refer to the document in
+// https://github.com/etcd-io/etcd/issues/15247#issuecomment-1777862093.
+func (s *EtcdServer) ensureLeadership() bool {
+	lg := s.Logger()
+	if err := s.linearizableReadNotify(s.ctx); err != nil {
+		lg.Warn("Ignore leases revoking request due to failing to ensure current member's leadership",
+			zap.Error(err))
```
chaochn47, line 908: log is related to lease revoke but `ensureLeadership` can be used by any etcd internal process instead of just lease revoking.

server/etcdserver/server.go
chaochn47: > Ok, raftStatus is not blocked. But will it serve latest raft state or will it be outdated?

It could depend on in which failure mode. If it is just raft loop blocked while network is intact, the old leader will step down to follower after receiving a message higher than current term from new leader. Hence, raftStatus is updated while etcd server is not. 

If leader is both network partitioned and raft loop blocked, then raftStatus is outdated unless we separate the heartbeat ticker to a separate for loop.

server/etcdserver/server.go
ahrtr: Good point. Updated.

chaochn47:
```
2023-10-26T19:47:41.7815241Z     failover_test.go:161: 
2023-10-26T19:47:41.7816607Z         	Error Trace:	/home/runner/work/etcd/etcd/tests/e2e/failover_test.go:161
2023-10-26T19:47:41.7819776Z         	            				/home/runner/work/etcd/etcd/tests/e2e/failover_test.go:140
2023-10-26T19:47:41.7822841Z         	Error:      	Received unexpected error:
2023-10-26T19:47:41.7827565Z         	            	Put "http://127.0.0.1:12381/defragBeforeCopy": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
2023-10-26T19:47:41.7834278Z         	Test:       	TestFailoverOnDefrag/defrag_blocks_one-third_of_requests_with_stopGRPCServiceOnDefrag_set_to_false
```

Known flake.. We may possibly auth retry set up failpoint just like https://github.com/etcd-io/etcd/pull/16094

ahrtr:
@mitake @serathius @chaochn47 @fuweid @tjungblu Please take a look at this PR and also https://docs.google.com/document/d/1peLDjwebnSuNR69ombuUiJ5fkSw7Rd3Slv_mUpOLsmg/edit 

I will try to add e2e test cases after we get consensus.

ahrtr:
> Did you check this PR with [#15247 (comment)](https://github.com/etcd-io/etcd/issues/15247#issuecomment-1420285175) ?

Yes, confirmed that it works.

mitake:
@ahrtr I could also check that this PR can work with the failpoint based test, great!

I'm still feeling that radical long term solution might be something like disabling proposal forwarding (or adding term  information in proposal forwarding), e.g. https://github.com/etcd-io/etcd/pull/16285 IIUC, this PR might still have rare corner cases like: other processes on the server which runs etcd or other threads of etcd itself become very busy and execution stops between leader check and sending request (this line https://github.com/etcd-io/etcd/pull/16822/files#diff-164b95d19232df0c51cd511fe4ca61de738660fc11db6ddcdafa0ec5f5c80a35R405)

However I think this solution can work for many practical cases. Is my understanding aligned with you (according to the doc I think it's aligned)? Anyway I agree with the approach of this PR.

ahrtr:
@mitake thx for the feedback.



> I'm still feeling that radical long term solution might be something like disabling proposal forwarding (or adding term information in proposal forwarding), e.g. #16285

Please see https://github.com/etcd-io/raft/pull/88#issuecomment-1804172432. Of course, the long-term solution is still open to discussion. There is no rush to make any decision for now.



> this PR might still have rare corner cases like: other processes on the server which runs etcd or other threads of etcd itself become very busy and execution stops between leader check and sending request (this line https://github.com/etcd-io/etcd/pull/16822/files#diff-164b95d19232df0c51cd511fe4ca61de738660fc11db6ddcdafa0ec5f5c80a35R405)

Even it happens (It's unlikely in practice), it shouldn't be a big problem. The lease depends on physical wall clock, an error of tens of or even hundreds of milliseconds should be acceptable for lease. Note that the default checkpoint interval is [5 minutes](https://github.com/etcd-io/etcd/blob/6e30d9ed7d1d64db44a93d020e7bc57aed61c6b4/server/lease/lessor.go#L52); it means that etcd does't run checkpoint at all for leases <= 5minutes.

> Anyway I agree with the approach of this PR.

thx

server/etcdserver/server.go
```diff
@@ -859,7 +859,18 @@ func (s *EtcdServer) run() {
...
+		// We shouldn't revoke any leases if current member isn't a leader,
+		// because the operation should only be performed by the leader. When
+		// the leader gets blocked on the raft loop, such as writing WAL entries,
+		// it can't process any events or messages from raft. It may think it
+		// is still the leader even the leader has already changed.
+		// Refer to https://github.com/etcd-io/etcd/issues/15247
 		lg := s.Logger()
+		if !s.ensureLeadership() {
+			lg.Warn("Ignore the lease revoking request because current member isn't a leader")
+			return
```
serathius, line 879: What is expected time to retry a revoke? Can you add tests for this scenario? We should make sure that we don't forget any revokes.

serathius:
I agree with proposed approach, however to merge the PR we need proper e2e tests.

The change is also pretty risky for backport. Those problems persisted for so long because lease refresh is not as utilized as much as other parts of API. We should think about how we qualify this.

server/etcdserver/server.go
```diff
@@ -892,6 +903,26 @@ func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
 	})
 }
 
+// ensureLeadership checks whether current member is still the leader.
+func (s *EtcdServer) ensureLeadership() bool {
+	lg := s.Logger()
+	if err := s.linearizableReadNotify(s.ctx); err != nil {
```
serathius, line 917: `linearizableReadNotify` does two things:
* Get the readIndex
* Wait for readIndex to be applied to db

For this purpose we only need the first one and having the second part might be risky as it injects an arbitrary log wait. I see two solutions:
* Limit the wait by adding timeout to context.
* Separate readIndex notification like in https://github.com/etcd-io/etcd/pull/16792

ahrtr: added a timeout.  

We will backpprt the PR to 3.5 and 3.4, so let's try to keep the change as simple as possible.

server/etcdserver/server.go
ahrtr: The retry interval defaults to 7s.

https://github.com/etcd-io/etcd/blob/db4e95cb1f8d3a8eebfb89c880d5fbd8d6165292/server/etcdserver/server.go#L349

ahrtr:
The `pull-etcd-e2e-amd64` failure will be resolved once https://github.com/kubernetes/test-infra/pull/31359 gets merged.

tests/e2e/v3_lease_no_proxy_test.go
```diff
@@ -0,0 +1,212 @@
...
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/pkg/v3/expect"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+// TestLeaseRevokeByOldLeader verifies that leases shouldn't be revoked by
+// old leader.
+// See the first case in the section "Root Cause" in
+// https://docs.google.com/document/d/1peLDjwebnSuNR69ombuUiJ5fkSw7Rd3Slv_mUpOLsmg/edit
```
serathius, line 35: Please don't link to google docs, they are not permanent way to store project documentation. Please provide the nesesery context in the comment or move the conclusions to issue and link to it.

tests/e2e/v3_lease_no_proxy_test.go
```diff
@@ -0,0 +1,212 @@
...
+func TestLeaseRevokeByNewLeader(t *testing.T) {
+	testLeaseRevokeIssue(t, func(cluster *e2e.EtcdProcessCluster, leaderIndex int) []string {
+		// return all members' endpoints, so that the client can switch to
+		// other member in case the leader gets stuck on writing disk.
+		return cluster.EndpointsGRPC()
+	})
+}
+
+func testLeaseRevokeIssue(t *testing.T, epsFn func(cluster *e2e.EtcdProcessCluster, leaderIndex int) []string) {
```
serathius, line 55: Please don't use function as an argument. It externalizes important part of the test making it harder to track what exactly is tested. Instead you can just provide a bool to switch behavior within a test.

ahrtr: Resolved although do not fully agree. Not a big deal though.

tests/e2e/v3_lease_no_proxy_test.go
```diff
@@ -0,0 +1,212 @@
...
+	require.Equal(t, 1, len(leases.Leases))
+
+	t.Log("Waiting for the keepAlive goroutine to exit")
+	close(stopC)
+	<-doneC
+}
+
+// TestLeaseRetryRevoke verifies that etcd will retry revoking lease
+// if failed due to temporary network issue.
+func TestLeaseRetryRevoke(t *testing.T) {
```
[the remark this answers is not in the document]
ahrtr: The failpoint is just mocking temporary issue (e.g. network). Do not see any big problem to do it with e2e test.

Looks like there is no easy way to mock this in a unit test. Removed the test case `TestLeaseRetryRevoke` from this PR for now, will evaluate it in a separate PR, and decide whether doing it using a unit test or e2e test.

mitake:
> Thanks for the comment. I think we only need to call s.ensureLeadership() once instead of calling it for each leases. The purpose is to ensure the decision of revoking leases is correct. As long as the current member is still the leader, then it's correct to revoke all the received leases even leader changes during the process.

That makes sense, then I think keeping the current strategy is fine. Thanks for examining the idea.

mitake:
Thanks a lot @ahrtr , I tested this PR manually also run the 2 new test cases (w/ and w/o the changes in this PR) and it worked as expected. This PR LGTM but want to confirm a few additional points:
- I think it's better to check the overhead of readindex, especially for renew request (which can happen regularly if clients use the API). Adding readindex to all renew requests would introduce some overhead for the case of many leases. Benchmarking this scenario might be good. I also think adding overhead is better than causing renew from a stale leader though.
- checkpointing part isn't covered by the e2e tests. I think it's hard to test the path though... do you have ideas for this? I guess making a checkpoint interval very short might be good, but want to have your opinions.

If we can agree on these potentially remaining items, I'd like to approve from my side.

ahrtr:
> * I think it's better to check the overhead of readindex, especially for renew request (which can happen regularly if clients use the API). Adding readindex to all renew requests would introduce some overhead for the case of many leases. Benchmarking this scenario might be good. I also think adding overhead is better than causing renew from a stale leader though.

Indeed there is some performance reduce. I did simple performance comparison using [benchmark](https://github.com/etcd-io/etcd/tree/main/tools/benchmark) with command below,

```
./benchmark --clients 100 --conns 100 --endpoints <leader-endpoint> lease-keepalive --total 1000000
```

 The main branch can process 64191 `LeaseKeepAliveRequest`s per second, but this PR can process 50344 `LeaseKeepAliveRequest`s per second. 

See more details below,
<details>
<summary>main branch</summary>

```
Summary:
  Total:	15.5783 secs.
  Slowest:	0.0868 secs.
  Fastest:	0.0000 secs.
  Average:	0.0012 secs.
  Stddev:	0.0014 secs.
  Requests/sec:	64191.7447

Response time histogram:
  0.0000 [1]	|
  0.0087 [996951]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0174 [2133]	|
  0.0261 [496]	|
  0.0347 [258]	|
  0.0434 [71]	|
  0.0521 [15]	|
  0.0608 [46]	|
  0.0694 [26]	|
  0.0781 [2]	|
  0.0868 [1]	|

Latency distribution:
  10% in 0.0004 secs.
  25% in 0.0006 secs.
  50% in 0.0010 secs.
  75% in 0.0014 secs.
  90% in 0.0021 secs.
  95% in 0.0027 secs.
  99% in 0.0058 secs.
  99.9% in 0.0163 secs.
```
</details>

<details>
<summary>PR</summary>

```
Summary:
  Total:	19.8632 secs.
  Slowest:	0.0805 secs.
  Fastest:	0.0001 secs.
  Average:	0.0018 secs.
  Stddev:	0.0013 secs.
  Requests/sec:	50344.3480

Response time histogram:
  0.0001 [1]	|
  0.0081 [996620]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0161 [2790]	|
  0.0242 [331]	|
  0.0322 [142]	|
  0.0403 [43]	|
  0.0483 [10]	|
  0.0564 [20]	|
  0.0644 [39]	|
  0.0725 [1]	|
  0.0805 [3]	|

Latency distribution:
  10% in 0.0008 secs.
  25% in 0.0011 secs.
  50% in 0.0016 secs.
  75% in 0.0021 secs.
  90% in 0.0028 secs.
  95% in 0.0036 secs.
  99% in 0.0065 secs.
  99.9% in 0.0121 secs.
```
</details>

LeaseKeepalive performance isn't that critical because etcd client sends the `LeaseKeepAliveRequest` every `TTL/3` seconds,
https://github.com/etcd-io/etcd/blob/a49a867960034f61da4165f6873b2557392a2939/client/v3/lease.go#L531




> * checkpointing part isn't covered by the e2e tests. I think it's hard to test the path though...

It tries to prevent an old leader from sending checkpoint requests to other members to pollute other member's remainingTTL. I think the scenario has already been covered by the two test cases added in this PR.  As long as the leases will not be wrongly revoked by either the old leader or the new leader, then we are good.

A related comment, actually the checkpoint feature is still experimental (disabled by default) and there is no any existing e2e test case for checkpoint feature at all. We should consider to add some e2e test cases to qualify & graduate it or deprecate it. The effort is tracked in 3.6 roadmap using https://github.com/etcd-io/etcd/issues/16292.

mitake (approved):

serathius (approved):
LGTM,

 I'm little worried about backport. The issue of early lease revoked was present from at least v3.0 if not earlier, but itself is not impactful enough to be noticed and reported. Proposed solution makes pretty non-trivial changes to lease API that are risky to be introduced immediately to old etcd versions without previous soak.

ahrtr:
Just rebased this PR.



> Proposed solution makes pretty non-trivial changes to lease API that are risky to be introduced immediately to old etcd versions without previous soak.

Actually the added source code are all protection code. It should be fine. The only minor concern is about the performance impact on leaseKeepAlive, but it shouldn't be a big problem,
- There is no any impact on K8s, because K8s doesn't call `KeepAlive` or `KeepAliveOnce` at all;
- As I mentioned in https://github.com/etcd-io/etcd/pull/16822#issuecomment-1850490064, etcd client sends the LeaseKeepAliveRequest every TTL/3 seconds, so the lease keep alive's performance isn't that critical.

ivanvc:
I'll attempt first to implement `WaitLeader` in 3.5 as suggested by @ahrtr (https://github.com/etcd-io/etcd/pull/17240#issuecomment-1890694504)

ivanvc:
Hi @ahrtr, after backporting `ExecuteUntil` and `WaitLeader`, what's missing to finalize backporting the rest of the PR?

ahrtr:
> Hi @ahrtr, after backporting `ExecuteUntil` and `WaitLeader`, what's missing to finalize backporting the rest of the PR?

Please feel free to backport the PR to 3.5 and 3.4 and fix whatever is missing, thx

ivanvc:
@ahrtr, as I was working on backporting this into 3.5. I realized that another functionality needs to be added to the e2e tests. `BinaryFailpoints` receive a timeout in `main`, while it's not implemented on 3.5. Should I create a PR to implement this or add the implementation along with this backport?

ahrtr:
> as I was working on backporting this into 3.5. I realized that another functionality needs to be added to the e2e tests. `BinaryFailpoints` receive a timeout in `main`, while it's not implemented on 3.5. Should I create a PR to implement this or add the implementation along with this backport?

I would suggest the same principle as before:  try to backport it firstly; if it needs huge change or effort, then try to implement it in 3.5 & 3.4 directly. Thanks.

PR #16826, opened 2023-10-25
[3.4] Backport clientv3:get AuthToken gracefully without dialing gRPC with balancer API to get extra connection

chaochn47:
Part of https://github.com/etcd-io/etcd/issues/16740

Backport https://github.com/etcd-io/etcd/pull/12165 to release-3.4 as the first step to remove etcd dependence on gRPC old balancer API. 

https://github.com/etcd-io/etcd/pull/16800 should be merged first and this PR will be rebased on top of that. 

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Issue #16740, opened 2023-10-11
Resolve CVE-2023-44487


ahrtr:
### What would you like to be added?

Refer to https://github.com/advisories/GHSA-qppj-fm5r-hxr3

## Bump golang.org/x/net:
- [x] 0.17.0 for 3.4/3.5/main
  - [X] https://github.com/etcd-io/etcd/pull/16739
  - [x] https://github.com/etcd-io/etcd/pull/16743
  - [x] https://github.com/etcd-io/etcd/pull/16746

## Bump golang:
- [X] 1.21.3 for main: https://github.com/etcd-io/etcd/pull/16741
- [x] 1.20.10 for 3.4/3.5
  - [X] https://github.com/etcd-io/etcd/pull/16745
  - [x] https://github.com/etcd-io/etcd/pull/16744

## Bump grpc-go:
- [X] 1.58.3 for main: https://github.com/etcd-io/etcd/pull/16739
- [x] 1.56.3 or 1.57.1 or 1.58.3 for 3.4/3.5.
  - [x] https://github.com/etcd-io/etcd/pull/16625
  - [x] https://github.com/etcd-io/etcd/pull/16781
  - [x] https://github.com/etcd-io/etcd/pull/16790
  - [x] https://github.com/etcd-io/etcd/pull/16795
  - [x] https://github.com/etcd-io/etcd/pull/16800
  - [x] https://github.com/etcd-io/etcd/pull/16826
  - [x] https://github.com/etcd-io/etcd/pull/16827
    - [x] https://github.com/etcd-io/etcd/pull/16842
    - [x] https://github.com/etcd-io/etcd/pull/16844
    - [x] https://github.com/etcd-io/etcd/pull/16857
    - [x] https://github.com/etcd-io/etcd/pull/16988
  - [x] https://github.com/etcd-io/etcd/pull/16997
  - [x] https://github.com/etcd-io/etcd/pull/16999
## Clarification
- [x] https://github.com/grpc/grpc-go/pull/6706/files#r1354612788

### Why is this needed?

resolve CVE

tjungblu:
Just to tick all boxes on what we need to do here:

- bump go to 1.20
- bump x/net to 0.17
- bump grpc to [v1.58.3](https://github.com/grpc/grpc-go/releases/tag/v1.58.3) (?)

The fix in apiserver seems a little bit more involved than just bumping dependencies: 
https://github.com/kubernetes/kubernetes/pull/121120/files

[...]

PR #13359, opened 2021-09-17
Fix http2 authority header in single endpoint scenario

serathius:
Fixes https://github.com/etcd-io/etcd/issues/13192

This doesn't fully fix authority in multiple endpoint scenario, but
should at least fixes single endpoint scenario.

cc @ptabor  @tatsuhiro-t @menghanl

PR #13203, opened 2021-07-09
client: call .Endpoints() in dial() in client/v3/client.go instead of accessing cfg.Endpoints directly

yishuT:
https://github.com/etcd-io/etcd/blob/0cdd558361c6bdbbd9e4023558e2f6ece71c18ad/client/v3/client.go#L299 accesses
endpoints without acquiring lock. Fix it to call Endpoints()

Fix #13201

PR #16827, opened 2023-10-25
[3.4] Backport clientv3: Replace balancer with upstream grpc solution

chaochn47:
Part of https://github.com/etcd-io/etcd/issues/16740 

This PR fully removes etcd customized implementation of gRPC old balancer API without breaking changes. 

Backported 
- #11574
- #12671 
- #12706 (stripped grpc proxy changes)
- #13203 
- #13359 (only e2e test is implemented, corresponding integration test is not deserved to be back-ported given its risk caused by large number of test code changes) 

#16826 should be merged first and this PR will be rebased on top of that.

This should allow us bump gRPC to v1.51.0 easier in release-3.4 in the follow up PR. 


chaochn47:
All the tests including the new added test around resolver, endpoint and proper h2 authority header passed. 

Next step would be evaluating the proxy test coverage along with the client changes~

ahrtr:
@chaochn47 please rebase this PR. thx

chaochn47:
@ahrtr 

This PR has been rebased. PTAL, thanks~

ahrtr:
This is a huge PR. Could you break down this PR, such as backporting each PR in the list separately?

chaochn47:
> This is a huge PR. Could you break down this PR, such as backporting each PR in the list separately?

It's doable. 

Just want to confirm that even if the backported PR introduces bug, it's okay to backport the fix separately, correct?

ahrtr:
> Just want to confirm that even if the backported PR introduces bug, it's okay to backport the fix separately, correct?

I think so. The goal right now is to break down this huge PR to make the reviewing easier.

chaochn47:
> > Just want to confirm that even if the backported PR introduces bug, it's okay to backport the fix separately, correct?
> 
> I think so. The goal right now is to break down this huge PR to make the reviewing easier.

First one is out.

ahrtr:
@chaochn47 are you still working on backporting the last PR?

chaochn47:
Yeah, I will help my colleague @jasdel to finish the remaining gRPC version upgrade to resolve the CVE. 

Is there a deadline so we can adjust the priorities on our TODO list?

ahrtr:
It's a blocker for https://github.com/etcd-io/etcd/issues/16751. I expect it can be resolved asap to unblock the release of 3.4.28 (which includes fixes to two CVEs). Thanks.

PR #16836, opened 2023-10-26
v3rpc: online defrag toggles gRPC health server serving status

chaochn47:
Follow up of https://github.com/etcd-io/etcd/pull/16278

### DIFF
--- a/server/etcdserver/api/v3rpc/grpc.go
+++ b/server/etcdserver/api/v3rpc/grpc.go
@@ -77,8 +77,9 @@ func Server(s *etcdserver.EtcdServer, tls *tls.Config, interceptor grpc.UnarySer
 	pb.RegisterAuthServer(grpcServer, NewAuthServer(s))
 
 	hsrv := health.NewServer()
-	pb.RegisterMaintenanceServer(grpcServer, NewMaintenanceServer(s, NewHealthNotifier(hsrv, s.Logger())))
+	healthNotifier := newHealthNotifier(hsrv, s)
 	healthpb.RegisterHealthServer(grpcServer, hsrv)
+	pb.RegisterMaintenanceServer(grpcServer, NewMaintenanceServer(s, healthNotifier))
 
 	// set zero values for metrics registered for this grpc server
 	grpc_prometheus.Register(grpcServer)
--- a/server/etcdserver/api/v3rpc/health.go
+++ b/server/etcdserver/api/v3rpc/health.go
@@ -18,37 +18,47 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/grpc/health"
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+
+	"go.etcd.io/etcd/server/v3/etcdserver"
 )
 
 const (
 	allGRPCServices = ""
 )
 
-type HealthNotifier interface {
-	StartServe()
-	StopServe(reason string)
+type notifier interface {
+	defragStarted()
+	defragFinished()
 }
 
-func NewHealthNotifier(hs *health.Server, lg *zap.Logger) HealthNotifier {
+func newHealthNotifier(hs *health.Server, s *etcdserver.EtcdServer) notifier {
 	if hs == nil {
 		panic("unexpected nil gRPC health server")
 	}
-	if lg == nil {
-		lg = zap.NewNop()
-	}
-	hc := &healthChecker{hs: hs, lg: lg}
+	hc := &healthNotifier{hs: hs, lg: s.Logger(), stopGRPCServiceOnDefrag: s.Cfg.ExperimentalStopGRPCServiceOnDefrag}
 	// set grpc health server as serving status blindly since
 	// the grpc server will serve iff s.ReadyNotify() is closed.
-	hc.StartServe()
+	hc.startServe()
 	return hc
 }
 
-type healthChecker struct {
+type healthNotifier struct {
 	hs *health.Server
 	lg *zap.Logger
+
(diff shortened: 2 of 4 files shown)


chaochn47:
cc @serathius

server/etcdserver/api/v3rpc/maintenance.go
```diff
@@ -89,10 +95,10 @@ func NewMaintenanceServer(s *etcdserver.EtcdServer, hn HealthNotifier) pb.Mainte
...
 func (ms *maintenanceServer) Defragment(ctx context.Context, sr *pb.DefragmentRequest) (*pb.DefragmentResponse, error) {
 	ms.lg.Info("starting defragment")
-	if ms.stopServingOnDefrag {
-		ms.hn.StopServe("defrag is active")
-		defer ms.hn.StartServe()
-	}
+	ms.defragC <- defragStarted
+	defer func() {
+		ms.defragC <- defragFinished
+	}()
```
serathius, line 101: Don't think that maintenanceServer needs to be aware that signal is sent via channels.
```suggestion
	ms.notify.DefragStarted()
	defer ms.notify.DefragFinished()
```

server/etcdserver/api/v3rpc/grpc.go
```diff
@@ -75,9 +75,11 @@ func Server(s *etcdserver.EtcdServer, tls *tls.Config, interceptor grpc.UnarySer
 	pb.RegisterLeaseServer(grpcServer, NewQuotaLeaseServer(s))
 	pb.RegisterClusterServer(grpcServer, NewClusterServer(s))
 	pb.RegisterAuthServer(grpcServer, NewAuthServer(s))
+	maintenanceServer := NewMaintenanceServer(s)
+	pb.RegisterMaintenanceServer(grpcServer, maintenanceServer)
 
 	hsrv := health.NewServer()
-	pb.RegisterMaintenanceServer(grpcServer, NewMaintenanceServer(s, NewHealthNotifier(hsrv, s.Logger())))
+	NewHealthNotifier(hsrv, s, maintenanceServer.DefragNotify())
```
serathius, line 82: Can you reverse the dependency? It's MaintenanceServer that should take a Notify interface, not reverse. 

Having maintainenaceServer pushish events is a decent idea, however it doesn't work here. We don't plan to have multiple listeners and the current would not work with them.

server/etcdserver/api/v3rpc/health.go
```diff
@@ -18,37 +18,57 @@ import (
...
-		lg = zap.NewNop()
-	}
-	hc := &healthChecker{hs: hs, lg: lg}
+	hc := &healthNotifier{hs: hs, s: s, lg: s.Logger(), defragNotifier: defragNotifier, stopGRPCServiceOnDefrag: s.Cfg.ExperimentalStopGRPCServiceOnDefrag}
 	// set grpc health server as serving status blindly since
 	// the grpc server will serve iff s.ReadyNotify() is closed.
-	hc.StartServe()
-	return hc
+	hc.startServe()
+	go hc.watchDefragStatus()
```
serathius, line 41: I'm worried about livecycle of this gorutine. I would recommend not using goroutines here, but a simple variable with mutex.

chaochn47: okay. This will be more simple and safer.

server/etcdserver/api/v3rpc/maintenance.go
chaochn47: Done

server/etcdserver/api/v3rpc/grpc.go
chaochn47: Done

server/etcdserver/api/v3rpc/health.go
```diff
@@ -57,7 +61,10 @@ func (hc *healthChecker) StartServe() {
 	hc.hs.SetServingStatus(allGRPCServices, healthpb.HealthCheckResponse_SERVING)
 }
 
-func (hc *healthChecker) StopServe(reason string) {
+func (hc *healthNotifier) stopServe(reason string) {
+	if !hc.stopGRPCServiceOnDefrag {
```
serathius, line 65: This condition would make more sense if it was in `defragFinished`.

serathius (approved):

server/etcdserver/api/v3rpc/health.go
chaochn47: Done. I guess you meant `defragStarted` :p

server/etcdserver/api/v3rpc/health.go
```diff
@@ -18,37 +18,47 @@ import (
...
 )
 
 const (
 	allGRPCServices = ""
 )
 
-type HealthNotifier interface {
-	StartServe()
-	StopServe(reason string)
+type notifier interface {
```
ahrtr, line 29: Nit: The previous name `HealthNotifier` makes more sense than `notifier`, because it isn't defined inside package "`health`".

Overall this PR is a little over-engineering. Both `(*maintenanceServer) Defragment` and `(*healthChecker) StopServe` are in the same package `api/v3rpc`, we don't have to necessarily introduce or refine the interface for now. 

Anyway, not a big deal for now.

ahrtr (approved):

server/etcdserver/api/v3rpc/health.go
serathius: On the other side, we should avoid the other extreme. Fact that structs are in single package doesn't mean that they should access all internal methods. Having a internal interface doesn't complicate code too much, but it clarifies the interned usage.

PR #16156, opened 2023-06-29
clientv3: remove v3.WithFirstKey() in Barrier.Wait()

kensou97:
fix the unexpected blocking when using Barrier.Wait(), e.g. NewBarrier(client, "a").Wait() will block if key "a" is not existed but "a0" is existed, but it should return immediately.

PR #16843, opened 2023-10-27
[3.4] Backport clientv3: remove v3.WithFirstKey() in Barrier.Wait()

jmhbnz:
Backports https://github.com/etcd-io/etcd/pull/16156 to fix the unexpected blocking when using Barrier.Wait(), e.g. NewBarrier(client, a).Wait() will block if key a is not existed but a0 is existed, but it should return immediately.


chaochn47 (approved):
LGTM. Thanks @jmhbnz

ahrtr (approved):
LGTM

Thanks @jmhbnz

Issue #15944, opened 2023-05-24
Investigate processes executed by leader


serathius:
### What would you like to be added?

Etcd has multiple processes that executed by leader, however neither etcd nor raft guarantees that there is only 1 cluster member that identifies itself as a leader at one time. 

Raft only guarantees that commited entries will not be lost by ensuring that they are persisted on quorum of members and electing a leader requires a quorum. However, there are periods of time where two members can consider themselves a leader, even though in reality they no longer have quorum and cannot commit any entries. An example is when leader is disconnected from other members and they elect new leader, there will be a period of time when old and new leader will be  present. 

TODO:
* Identify all the etcd processes that depend on being executed by only the leader aka isLeader(). Example cluster version picking.
* Define list of invariants that those process should fulfil. For example: idempotent, transactional etc.
* Validate those invariants on all of those processes
* Codify those invariants so no future changes impact etcd correctness.

### Why is this needed?

To surface the issue with incorrect pattern used in etcd and improve the codebase.

ahrtr:
Another example, `s.raftStatus()`  might return stale info if current leader is in network partition status.
https://github.com/etcd-io/etcd/blob/3347568cc0684326a71cb110204a90827bf00399/server/etcdserver/server.go#L1491-L1500

Issue #13527, opened 2021-12-08
network partitioned leader was not able to step down to follower


chaochn47:
One of aws production 3-member etcd clusters had 2 leader at a time. During this event, the droplet (the Physical Host on which the EC2 instances run) behind the old leader (A) was unavailable and the associated EBS volume failed to serve WAL fsync. 

Ideally, the checkquorum raft message should be raised after the election timeout breaches the limit and `RecentActive` should be false for each peers if no recent hearbeat response or MsgAppResp reached leader. 

https://github.com/etcd-io/etcd/blob/161bf7e7be9a3e2bbb920568020d938f757adef5/etcdserver/raft.go#L170-L173

However, the ticker is not triggered due to the disk write stall in the raft output ready handling logic. This can be easier reproduced like the following

old Leader
```bash
# run A with the failpoint enabled
[root@ip-10-0-61-148 bin]# etcdctl -w table endpoint status
+----------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
|    ENDPOINT    |        ID        | VERSION | DB SIZE | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS |
+----------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
| 127.0.0.1:2379 | a320313566418492 |  3.4.18 |   20 kB |      true |      false |        35 |    1741005 |            1741005 |        |
+----------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
[root@ip-10-0-61-148 bin]# curl http://127.0.0.1:1234/go.etcd.io/etcd/etcdserver/raftAfterSave -XPUT -d'sleep(600000)'
[root@ip-10-0-61-148 bin]# iptables -A INPUT -s 10.0.123.82 -j DROP && iptables -A INPUT -s 10.0.171.218 -j DROP
[root@ip-10-0-61-148 bin]# curl -sL http://localhost:2379/metrics | grep "is_leader"
# HELP etcd_server_is_leader Whether or not this member is a leader. 1 if is, 0 otherwise.
# TYPE etcd_server_is_leader gauge
etcd_server_is_leader 1
[root@ip-10-0-61-148 bin]# etcdctl -w table endpoint status
+----------------+------------------+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
[...]

PR #16847, opened 2023-10-29
etcdserver/raft.go: separate raft tick and ready

chaochn47:
Related to https://github.com/etcd-io/etcd/issues/15944. 

etcd leader should not fail to send the heartbeat if disk is blocked. 

This should be helpful to cancel watch streams when disk is stalled. Refer to https://github.com/etcd-io/etcd/issues/13527. 

TODO:

1. add verification that watch will be cancelled if disk is stalled after this patch (also with negative test case). 
2. evaluate if all the `s.Leader()` should be replaced with `r.RaftStatus.Leader`


chaochn47:
If there are any committed membership reconfig changes, heartbeat to drive election and votes has a dependency that those config updates must be applied first. So this PR is incorrect and a no-go option. See more in https://github.com/etcd-io/etcd/pull/7706

PR #16850, opened 2023-10-30
add checks for available cipher before setting h2 proto

pschou:


jmhbnz (commented):
Hey @pschou - Thanks for raising this!

To save reviewers some time would you mind either sharing the reproduce steps / configuration you are using, or alternatively adding an test?

serathius:
> Without this, etcd crashes on startup.

If that's a case, please provide a test that confirms that and prevents invalid configuration in future. Best if that would be an e2e test. Please let me know if you need any help with writing it.

pschou:
Example:

./bin/etcd --tls-max-version=TLS1.2 --cipher-suites='TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384' --cert-file=/etc/pki/server.pem --key-file=/etc/pki/server.pem --trusted-ca-file=/etc/pki/ca.pem --client-cert-auth=true --advertise-client-urls=https://localhost:32379 --listen-client-urls=https://localhost:32379

Without the update:
...
{"level":"fatal","ts":"2023-11-06T11:34:00.039138-0500","caller":"etcdmain/etcd.go:196","msg":"listener failed","error":"http2: TLSConfig.CipherSuites is missing an HTTP/2-required AES_128_GCM_SHA256 cipher (need at least one of TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256)","stacktrace":"go.etcd.io/etcd/server/v3/etcdmain.startEtcdOrProxyV2\n\tgo.etcd.io/etcd/server/v3/etcdmain/etcd.go:196\ngo.etcd.io/etcd/server/v3/etcdmain.Main\n\tgo.etcd.io/etcd/server/v3/etcdmain/main.go:40\nmain.main\n\tgo.etcd.io/etcd/server/v3/main.go:31\nruntime.main\n\truntime/proc.go:250"}
...

With the update:
...
{"level":"info","ts":"2023-11-06T13:38:32.33227-0500","caller":"embed/serve.go:251","msg":"serving client traffic securely","traffic":"http","address":"127.0.0.1:32379"}
...

jmhbnz (commented):
Hey @pschou - Thanks for adding the example.  As suggested above, to make sure this behavior doesn't get broken again in future it would be ideal to have an e2e test covering this situation.

Please let us know if you need a hand adding this.

pschou:
> Hey @pschou - Thanks for adding the example.  As suggested above, to make sure this behavior doesn't get broken again in future it would be ideal to have an e2e test covering this situation.
> 
> Please let us know if you need a hand adding this.

Thank you for the review and I do think I will need help.  I have made an e2e attempt to see if this is a step in the right direction.  Any advice or help would be greatly appreciated.

client/pkg/transport/listener.go
```diff
@@ -527,8 +528,28 @@ func (info TLSInfo) ServerConfig() (*tls.Config, error) {
...
+	// list of ciphers may not allow HTTP2.  Test for this and disable if this is the case.
+	if len(cfg.CipherSuites) > 0 && cfg.MaxVersion > 0 && cfg.MaxVersion < tls.VersionTLS13 {
+
+		// Test for the cases where TLS1.3 is not allowed, and thus http2 may not
+		// be possible when select cipherSuites are defined.
+		var foundHTTP2Cipher bool
+		for _, c := range cfg.CipherSuites {
+			switch c {
+			// If one of these cipher suites are defined, then the h2 protocol is possible.
+			case tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256:
```
jmhbnz, line 541: Nit: There is a longstanding upstream golang issue and cl which proposes making these bad ciphers easier to test for so we can potentially simplify this in future once that is done.  Perhaps we should add a `// TODO:` in here mentioning the upstream issue so we can revisit in future: https://github.com/golang/go/issues/41068#issuecomment-722549561

jmhbnz (approved):
This aligns with http2 spec[1][2] so LGTM, thanks for refining this @pschou.

Just one suggestion about tracking a future improvement to how we implement this check.

1: https://httpwg.org/specs/rfc9113.html#tls12ciphers
2: https://httpwg.org/specs/rfc9113.html#BadCipherSuites

client/pkg/transport/listener.go
```diff
@@ -527,8 +528,28 @@ func (info TLSInfo) ServerConfig() (*tls.Config, error) {
...
+				foundHTTP2Cipher = true
+				break
+			}
+		}
+		info.DisableHTTP2 = !foundHTTP2Cipher
+	}
+
+	if !info.DisableHTTP2 {
```
serathius, line 549: Don't think it makes sense to allow running etcd without grpc which requires HTTP2. 

grpc is the main etcd API and all the clients implement it. We have alternative API via grpc-gateway, but it just translates request to GRPC API which must be running. User could potentially expose separate endpoint without TLS, but then why they would configure cyphers at all.

Would it be better to just abort etcd startup if user doesn't provide HTTP2 capable cypher?

jmhbnz: Agree it's an edge situation.  I'm happy with handling the case either via graceful shutdown or continued startup provided a clear log message explains the situation.

pschou: Ok, that makes sense.  Here is an update where it has reverted to a graceful shutdown (which was the issue causing this PR) and added a flag to disable h2.  The idea of adding a *disabling* option is to keep the invocation consistent with the current configurations already in production.

client/pkg/transport/listener.go
pschou: How about this... avoid TODO by letting the http2 ConfigureServer(s *http.Server, conf *Server) throw the built-in error when setting the cipher set.

The e2e test will see a different reply if this http2 cipher set changes, but a failing test can be adjusted when the test fails-- there's no real need to add a TODO in an e2e test as a failure will imply a TODO.

client/pkg/transport/listener.go
serathius: > The idea of adding a disabling option is to keep the invocation consistent with the current configurations already in production.

Not sure I understand. Didn't you present a case that you cannot start etcd if provided cypher suites don't support HTTP2? Fixing this should be our priority. Not sure about case of running etcd without HTTP2 yet.

pschou: They seem closely tied together-- in that choosing a set of select cipher suites will cause the server not to start, as http2 is enabled by default.  The cases for consideration are:

- HTTPS is not enabled - Cipher setting does not matter, and GRPC/HTTP2 can be enabled or disabled at will

- HTTPS is enabled, but HTTP2 ciphers are not allowed - then GRPC will not run, but the etcd service can still serve client requests (although without the HTTP2 protocol as an option)

- HTTPS is enabled, and HTTP2 ciphers are allowed - then GRPC can run (as HTTP2 is allowed), and any served connection can use HTTP or HTTP2 at will

The hard connection for consideration is GRPC relies on HTTP2.

serathius: My main concern for now is to avoid overcomplicating the etcd server initialization code. For now let's not  add any new behaviors/features (adding  --disable-http2 is a feature), just improve the user experience of misconfiguring the cyphers. This would be a simple improvement that we could consider backporting.

client/pkg/transport/listener.go
pschou: I stand corrected; a TODO is needed, as you suggested.  My apologies.  Please let me know if you have any further considerations.

serathius (approved):

serathius:
cc @ahrtr

serathius:
Please squash commits.

pschou:
Sorry about that last-minute thought.  I want to ensure that users can see an information line indicating what is happening so debugging can be more graceful.

client/pkg/transport/listener.go
```diff
@@ -527,8 +527,35 @@ func (info TLSInfo) ServerConfig() (*tls.Config, error) {
...
+			case tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256:
+				hasHTTP2cipher = true
+				break
+			}
+		}
+		if hasHTTP2cipher {
+			cfg.NextProtos = []string{"h2"}
+		} else {
+			// Display a warning that none of the h2 ciphers are available and likewise GRBC cannot be started.
+			info.Logger.Info("CipherSuites is missing an HTTP2 capable cipher (TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256). GRBC will not be available to start.")
```
serathius, line 553: Should we recommend users to either:
* Add a cipher
* Enable TLS 1.3

Also, would it be better to return error here so we exit here and not on panic?

pschou: Agreed... the scope of the Listener doesn't have the ability to check for this, so I'll move it over...

tests/e2e/etcd_config_test.go
```diff
@@ -452,3 +452,61 @@ func TestEtcdTLSVersion(t *testing.T) {
 	proc.Wait() // ensure the port has been released
 	proc.Close()
 }
+
+func TestEtcdNonH2CiphersWithGRPC(t *testing.T) {
+	e2e.SkipInShortMode(t)
```
ahrtr, line 457:
```suggestion
	e2e.BeforeTest(t)
```

tests/e2e/etcd_config_test.go
```diff
@@ -452,3 +452,61 @@ func TestEtcdTLSVersion(t *testing.T) {
...
+
+		"--tls-min-version", "TLS1.2",
+		"--tls-max-version", "TLS1.2",
+	}
+	err := e2e.SpawnWithExpect(argsWithGRPCandH2andWithoutH2Ciphers, expect.ExpectedResponse{Value: "http2: TLSConfig.CipherSuites is missing an HTTP/2-required AES_128_GCM_SHA256 cipher (need at least one of TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256)"})
+	require.ErrorContains(t, err, "http2: TLSConfig.CipherSuites is missing an HTTP/2-required AES_128_GCM_SHA256 cipher (need at least one of TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256)")
+}
+
+func TestEtcdNonH2CiphersWithoutGRPC(t *testing.T) {
+	e2e.SkipInShortMode(t)
```
ahrtr, line 484:
```suggestion
	e2e.BeforeTest(t)
```

pschou: changed

tests/e2e/etcd_config_test.go
pschou: changed

client/pkg/transport/listener.go
pschou: By "the ability to check for this", is the ability to verify that GRBC is being enabled.

pschou:
> Please squash commits.

done

client/pkg/transport/listener.go
```diff
@@ -527,8 +527,35 @@ func (info TLSInfo) ServerConfig() (*tls.Config, error) {
...
+			case tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256:
+				hasHTTP2cipher = true
+				break
+			}
+		}
+		if hasHTTP2cipher {
+			cfg.NextProtos = []string{"h2"}
+		} else {
+			// Display a warning that none of the h2 ciphers are available and likewise GRBC cannot be started.
+			info.Logger.Info("CipherSuites is missing an HTTP2 capable cipher (TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256).  Disabling HTTP2 protocol.")
```
ahrtr, line 553: We should return an error directly to fail the etcd's startup in such case.

The only concern is we also support httpOnly mode, in which case we shouldn't return error because it doesn't have to necessarily be over http2, https://github.com/etcd-io/etcd/blob/6db5e00103accde744c856be03f38f44569eca65/server/embed/etcd.go#L668

It's OK to have one listener for gRPC and the other for http, but it does't make sense if an etcd member only supports serving http requests. Probably we should add a validation to prevent such extreme scenario.

server/embed/serve.go
```diff
@@ -202,6 +203,19 @@ func (sctx *serveCtx) serve(
...
+			for _, p := range tlscfg.NextProtos {
+				if p == "h2" {
+					hasH2 = true
+				}
+			}
+			if !hasH2 {
+				grpcStartError := errors.New("GRPC cannot be started without HTTP2 protocol.  Please add at least one of the HTTP2 capable ciphers or set tls-max-version to TLS1.3.")
+				sctx.lg.Error("Configure https server failed", zap.Error(grpcStartError))
+				return grpcStartError
+			}
```
ahrtr, line 217: see https://github.com/etcd-io/etcd/pull/16850/files#r1405109505 . 

etcd should have already failed before entering into this method.
```suggestion
```

jmhbnz:
Discussed during sig-etcd triage meeting. Would be great to finish this off @pschou do you have time to rebase this and please address the two comments from @ahrtr?  Thanks!

PR #16858, opened 2023-10-31
etcdserver: make livez return ok when defrag is active.

siyuanfoundation:

cc @chaochn47 @serathius


tests/e2e/http_health_check_test.go
```diff
@@ -279,6 +280,27 @@ func TestHTTPLivezReadyzHandler(t *testing.T) {
...
+				},
+				{
+					url:                  "/readyz",
+					expectedTimeoutError: true,
+					expectedStatusCode:   http.StatusServiceUnavailable,
+					expectedRespSubStrings: []string{
+						`[-]serializable_read failed: defrag is active`,
+						`[+]data_corruption ok`,
+					},
+				},
```
chaochn47, line 301: Nice!

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -207,14 +208,14 @@ type CheckRegistry struct {
 
 func installLivezEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
 	reg := CheckRegistry{path: "/livez", checks: make(map[string]HealthCheck)}
-	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.Register("serializable_read", serializableReadCheck(server, true))
```
chaochn47, line 211:
```suggestion
	reg.Register("serializable_read", serializableReadCheck(server, true /* skipCheckDuringDefrag */))
```

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -355,13 +356,18 @@ func activeAlarmCheck(srv ServerHealth, at pb.AlarmType) func(context.Context) e
...
+		if srv.IsDefragActive() {
+			return fmt.Errorf("defrag is active")
+		}
 		ctx = srv.AuthStore().WithRoot(ctx)
 		_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: true})
-		if err != nil {
-			return fmt.Errorf("range error: %w", err)
-		}
-		return nil
+		return err
```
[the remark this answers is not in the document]
siyuanfoundation: ok.

server/storage/backend/backend.go
```diff
@@ -101,7 +102,8 @@ type backend struct {
 	// openReadTxN is the number of currently open read transactions in the backend
 	openReadTxN int64
 	// mlock prevents backend database file to be swapped
-	mlock bool
+	mlock          bool
+	isDefragActive bool
```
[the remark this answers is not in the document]
siyuanfoundation: good idea!

chaochn47 (approved):
LGTM, thanks @siyuanfoundation!

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -44,6 +44,7 @@ type ServerHealth interface {
 	Range(context.Context, *pb.RangeRequest) (*pb.RangeResponse, error)
 	Config() config.ServerConfig
 	AuthStore() auth.AuthStore
+	IsDefragActive() bool
```
serathius, line 47: Please utilize https://github.com/etcd-io/etcd/pull/16836 instead of writing a separate mechanism to track defrag.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -355,8 +356,16 @@ func activeAlarmCheck(srv ServerHealth, at pb.AlarmType) func(context.Context) e
 	}
 }
 
-func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
+func serializableReadCheck(srv ServerHealth, skipCheckDuringDefrag bool) func(ctx context.Context) error {
 	return func(ctx context.Context) error {
+		// skips the check if defrag is active.
+		if skipCheckDuringDefrag && srv.IsDefragActive() {
```
serathius, line 362: What if there is a deadlock on defrag? I think I missed this in the original design, but I don't think we should just skip checking the serializable read if defrag is active. Single liveness probe failure during defrag is not a bad thing. We expect that 3 failures are required to kill etcd.

The question is not "should defrag cause livez to fail?", but "should we kill etcd if defrag is taking 15 seconds?".
I think yes.

server/etcdserver/api/etcdhttp/health.go
chaochn47: I think it depends on if https://github.com/etcd-io/etcd/pull/16836 is backported to release-3.4 and release-3.5. Is there any objections on the backports? @ahrtr @serathius

ahrtr: No any objection from my side.

We need to backport https://github.com/etcd-io/etcd/pull/16278 firstly.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: I guess if the user expects the defrag latency to be long, they should probably configure the livez retry period to be long to match that.

serathius: > If we allow livez to fail during defrag, this etcd process will be in a crash loop state, right?

No crash loop as defrag operation will not persist between restarts. 

> In cloud environment, it is not uncommon that defrag could take more than 15 seconds with large number of key values and low throughput disk volume.

Asked @siyuanfoundation to take a look into defrag duration metrics. I agree that defrag can take couple of seconds, however I think 15s of API unavailability seems too long. I think we should assume some minimal IOPs from disk size. 

You are right it possible that on cloud users could provision minimal disk sizes (20GB) to just fit etcd db (6GB) and this will result with minimal IOPs, however I would argue that those are degenerate cases and those clusters would handle etcd benchmark. 

It's hard for me to judge what is expected defrag duration I don't think there is any official SLO or set of it. It  was not a problem for GKE so I haven't investigated further. I think the best way to get an agreement would be to share some internal data about defrag duration. 

@chaochn47 I know you were interested in it due to grpc probes. Could you share some data?

serathius: I double checked the defrag duration stats. I have confirmed that defrag time can be really long on large clusters, even with over provisioned disks it can grow easily over 15s. As long as defrag time is linear to size of DB, we won't be able to define a sensible maximum time that we can enforce.

With that I would agree that the proposal makes sense.

server/etcdserver/api/etcdhttp/health.go
serathius: No objections from my side.

server/etcdserver/v3_server.go
```diff
@@ -1028,3 +1028,13 @@ func (s *EtcdServer) downgradeCancel(ctx context.Context) (*pb.DowngradeResponse
 	resp := pb.DowngradeResponse{Version: version.Cluster(s.ClusterVersion().String())}
 	return &resp, nil
 }
+
+func (s *EtcdServer) SetDefragActive(active bool) {
```
serathius, line 1032: Please don't store more state on *EtcdServer. Please use `healthNotifier` to store state required to calculate whether etcd is healthy or not.

siyuanfoundation: done

ahrtr:
Please rebase this PR, I will take a look later.

server/etcdserver/api/v3rpc/maintenance.go
```diff
@@ -74,22 +74,19 @@ type maintenanceServer struct {
...
+	srv := &maintenanceServer{lg: s.Cfg.Logger, rg: s, hasher: s.KV().HashStorage(), bg: s, a: s, lt: s, hdr: newHeader(s), cs: s, d: s, vs: etcdserver.NewServerVersionAdapter(s)}
 	if srv.lg == nil {
 		srv.lg = zap.NewNop()
 	}
+	s.Backend().SubscribeDefragNotifier(healthNotifier)
 	return &authMaintenanceServer{srv, &AuthAdmin{s}}
 }
 
 func (ms *maintenanceServer) Defragment(ctx context.Context, sr *pb.DefragmentRequest) (*pb.DefragmentResponse, error) {
 	ms.lg.Info("starting defragment")
```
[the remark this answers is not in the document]
siyuanfoundation: The original notifier is called in grpc server. 

1. It is strange to have a http endpoint depend on grpc server calls. The cluster could be started with grpcEnabled=false.
2. There could be 2 grpc servers, 1 for insecure and 1 for secure. Which one should be used to determine the http endpoint state?
Based on these reasons, I think it is better to move the notifiers to the backend.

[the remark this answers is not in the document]
serathius: Ok, but this is unrelated an change. Still doesn't solve the issue that there might be multiple calls to Defrag, that will be blocked on somewhere in internals. This still leaves us open to concurrency issue. Let's just use the notifier as it is and do a followup issue to fix notifier.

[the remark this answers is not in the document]
siyuanfoundation: tbh, it is very gnarly to pass notifiers between http and grpc, something like https://github.com/etcd-io/etcd/compare/main...siyuanfoundation:etcd:defrag2?expand=1. 

How about I start a PR to move the notifiers to the backend first?

PR #16881, opened 2023-11-07
Add range response KV length as a metric

tjungblu:
This adds a metric to allow us to alert on large range responses described in #14809.

### DIFF
--- a/server/etcdserver/v3_server.go
+++ b/server/etcdserver/v3_server.go
@@ -116,6 +116,11 @@ func (s *EtcdServer) Range(ctx context.Context, r *pb.RangeRequest) (*pb.RangeRe
 				traceutil.Field{Key: "response_count", Value: len(resp.Kvs)},
 				traceutil.Field{Key: "response_revision", Value: resp.Header.Revision},
 			)
+
+			// collect the metric on very large responses only, because the key dimensionality might blow up Prometheus.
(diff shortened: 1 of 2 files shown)


server/etcdserver/v3_server.go
```diff
@@ -136,6 +136,9 @@ func (s *EtcdServer) Range(ctx context.Context, r *pb.RangeRequest) (*pb.RangeRe
 		err = serr
 		return nil, err
 	}
+
+	rangeResponseKvCount.WithLabelValues(string(r.Key)).Observe(float64(len(resp.Kvs)))
```
tjungblu, line 140: given the key label could cause a very large /metrics response, we might want to only emit this on n > 1000 or so. Also thought about adding this to the expensive trace request above. 

Let me know what you guys think.

jmhbnz: Hey @tjungblu - Thanks for raising this idea.  With the etcd project making tentative steps towards adopting the kubernetes enhancements process for new features or meaningful decisions this addition could be a potential item to first have a kep for `sig/etcd`.

Just an idea if you wanted a larger pool of eyes on the proposal. If you would rather just proceed here personally I am ok with it.

Is there any way you can provide some qualification of what impact this new metric would have on the response size? 🤔

Issue #16866, opened 2023-11-05
Add etcd_server_range_duration_seconds metrics


lavacat:
### What would you like to be added?

We already have [etcd_server_apply_duration_seconds](https://github.com/etcd-io/etcd/blob/d38f7dceb22c2aa66f749d7aa4d6a10d10bc8116/server/etcdserver/txn/metrics.go#L31) metrics.

But [Range](https://github.com/etcd-io/etcd/blob/d38f7dceb22c2aa66f749d7aa4d6a10d10bc8116/server/etcdserver/v3_server.go#L134) doesn't go through Applier. The only metrics we have about Range timing is `grpc_server_handled_total{grpc_method="Range"}` provided by grpc_middleware. Problem is that this grpc metric also includes timing of `linearizableReadNotify`.

Please add etcd_server_range_duration_seconds in [txn.Range](https://github.com/etcd-io/etcd/blob/d38f7dceb22c2aa66f749d7aa4d6a10d10bc8116/server/etcdserver/txn/txn.go#L135)

### Why is this needed?

better observability

RaviHari:
I would like to work on this

thedtripp:
Hello @jmhbnz. 
I'd like to take a crack at this one. I'm a new contributor looking for mentorship.

jmhbnz:
> Hello @jmhbnz. I'd like to take a crack at this one. I'm a new contributor looking for mentorship.

Thanks for volunteering @thedtripp! Assigned to you.  Please start by recreating the earlier proposed changes from #16902. Then we need a second commit to add a basic test verifying the new metric works. Let us know how you go and if you have any questions 🙏🏻 

/assign @thedtripp

thedtripp:
@jmhbnz 
I made a PR: https://github.com/etcd-io/etcd/pull/17983

thedtripp:
@jmhbnz @ivanvc @ahrtr Thank you for guiding me through that first PR. Any suggestions on where I should focus for my next contribution?

jmhbnz:
> @jmhbnz @ivanvc @ahrtr Thank you for guiding me through that first PR. Any suggestions on where I should focus for my next contribution?

Most welcome! For a much quicker next contribution we need to update our website community page https://github.com/etcd-io/website/issues/850.  If you're interested please feel free to leave a comment and I will assign it to you. Alternatively please take a look at our `help wanted` issues.

thedtripp:
[...]

PR #16902, opened 2023-11-09
feat: add etcd_server_range_duration_seconds metric

RaviHari:
fixes: https://github.com/etcd-io/etcd/issues/16866
add etcd_server_range_duration_seconds metric

### DIFF
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -144,11 +144,9 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, r *pb.RangeRequest) 
 		ctx = context.WithValue(ctx, traceutil.TraceKey{}, trace)
 	}
 	op := "range"
-	stringer := &pb.InternalRaftStringer{Request: &pb.InternalRaftRequest{Range: r}}
 	defer func(start time.Time) {
 		success := err == nil
-		RangeSecObserve(v3Version, op, success, time.Since(start))
-		WarnOfExpensiveRequest(lg, DefaultRangeWarningApplyDuration, start, stringer, nil, nil)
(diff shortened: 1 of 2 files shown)


jmhbnz (changes requested):
Thanks for raising this @RaviHari - Could you please take a look at Integration or e2e tests for this new metric so we don't accidentally break it in future?

server/etcdserver/txn/txn.go
```diff
@@ -138,6 +143,13 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, r *pb.RangeRequest)
 		trace = traceutil.New("range", lg)
 		ctx = context.WithValue(ctx, traceutil.TraceKey{}, trace)
 	}
+	op := "range"
+	stringer := &pb.InternalRaftStringer{Request: &pb.InternalRaftRequest{Range: r}}
+	defer func(start time.Time) {
+		success := err == nil
+		RangeSecObserve(v3Version, op, success, time.Since(start))
```
lavacat, line 150: I think it's better to remove "version" label, because it's always same value. The only benefit is that it follows `applySec` pattern.

lavacat (changes requested):

RaviHari:
> Thanks for raising this @RaviHari - Could you please take a look at Integration or e2e tests for this new metric so we don't accidentally break it in future?

Hi @jmhbnz I looked at Integration and e2e tests it does not look like this change will break any of those tests.

jmhbnz:
> Hi @jmhbnz I looked at Integration and e2e tests it does not look like this change will break any of those tests.

Apologies I should have been more clear, I was asking if you could please add a test to ensure this metric does not get broken in future.

lavacat:
@RaviHari do you need help with figuring out where to add test?

Issue #16893, opened 2023-11-08
Run `arm64` `e2e` tests for every pull request rather than on nightly schedule


jmhbnz:
### What would you like to be added?

Recently in https://github.com/etcd-io/etcd/pull/16801 we introduced on demand github actions runners for the `arm64` platform.

Having on demand runner infrastructure in place means we should now have enough capacity to begin running `arm64` tests for every pull request.

We should update our `.github/workflows/e2e-arm64-nightly.yaml` workflow to switch from scheduled nightly runs to instead running `on: [push, pull_request]`: https://github.com/etcd-io/etcd/blob/main/.github/workflows/e2e-arm64-nightly.yaml#L7-L8

### Why is this needed?

Gain greater confidence and samples of e2e tests running against `arm64` cpu architecture.

Hopefully detect issues before pull requests are merged to `main`.

mingli103:
Hi @jmhbnz please assign this issue to me.

jmhbnz:
One further thought for this issue @mingli103.  Now that we are moving to `on: [push, pull_request]` for the e2e arm64 workflow we can actually move away from having two separate files.

Currently we have:
- `.github/workflows/e2e-arm64-template.yaml` - Shared template
- `.github/workflows/e2e-arm64-nightly.yaml` - Runs template against both `release-3.5` and `main` branches nightly.

Moving forward we can just rename `.github/workflows/e2e-arm64-template.yaml` to `.github/workflows/e2e-arm64.yaml` and delete the other file.  We can then just make the template file a standard workflow that will run on pull request.

jmhbnz:
Closed as completed in #16950, thanks @mingli103 🙏🏻

PR #16801, opened 2023-10-20
Switch over to managed Arm64 hosts

alexellis:
## Description

Switch over to managed Arm64 hosts

The service is provided by actuated.dev, and **sponsored by both Ampere and the CNCF**.

## Details

This change switches over from 2x self-managed runners where side effects are possible between builds, to a pool of servers where each build runs in an isolated VM.

Benefits to the etcd maintainers and community:

* Securely isolated, immutable environments for every build
* No more server maintenance, reduced documentation and process
* Support from an externally contracted team, much more efficient use of CNCF credits.
* Potential to enable builds for each PR instead of just nightly
 
@dims and other etcd maintainers have been briefed on the discussion / plans via Slack.

Please feel free to ask questions here too.

PR #16912, opened 2023-11-10
etcd-e2d-test:rename e2e-arm64 file and runs it on every pull request

mingli103:
KubeconNA 2023 Contribfest issue  [#16893 ](https://github.com/etcd-io/etcd/issues/16893). Recently in #16801 we introduced on demand github actions runners for the arm64 platform. Having on demand runner infrastructure in place means we should now have enough capacity to begin running arm64 tests for every pull request. Currently we have:
.github/workflows/e2e-arm64-template.yaml - Shared template .github/workflows/e2e-arm64-nightly.yaml - Runs template against both release-3.5 and main branches nightly. Moving forward we can just rename .github/workflows/e2e-arm64-template.yaml to .github/workflows/e2e-arm64.yaml and delete the other file. We can then just make the template file a standard workflow that will run on pull request.


.github/workflows/e2e-arm64.yaml
```diff
@@ -20,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
-          ref: "${{ inputs.etcdBranch }}"
+          ref: "main"
```
jmhbnz, line 18: Rather than hardcoding `main` I suggest we remove the `with:` completely for `actions/checkout` so we use the default behaviour of only a single commit being fetched for the ref/SHA that triggered the workflow.

This will allow us to safely backport these changes to `release-3.5` branch.

jmhbnz (changes requested):
Thanks for working on this @mingli103 - A suggestion below about how we handle the branching 🙏🏻

.github/workflows/e2e-arm64.yaml
```diff
@@ -1,11 +1,6 @@
 ---
 name: Reusable Arm64 E2E Workflow
```
jmhbnz, line 2: Just noticed we should probably align this name with the existing e2e per pull request workflow which is `E2E`: https://github.com/etcd-io/etcd/blob/main/.github/workflows/e2e.yaml#L2

So we could make this `E2E-Arm64`.

mingli103: Make sense.

jmhbnz (approved):
LGTM - Thanks @mingli103

serathius (approved):

jmhbnz:
Hey @mingli103 - Can you please rebase and force push this branch?  We had some flaky tests recently and if you can rebase onto latest commits from `main` your branch will get some fixes so tests will hopefully pass.

serathius (approved):

jmhbnz:
Hey @mingli103 - It looks like this pr is rewriting history. Can you please retry the rebase so your pr only includes your single commit?

PR #16914, opened 2023-11-11
robustness test: add with functions of randomizable config params in robustness test

siyuanfoundation:
Compared with With functions in e2e like e2e.WithSnapshotCount, the new With functions takes a list of possible field values, and randomly pick one when running the test. This would make the test cluster dynamic from a set of exploration space.

Added a script to auto generate the with functions for server config fields.

### DIFF
--- a/tests/robustness/options/cluster_options.go
+++ b/tests/robustness/options/cluster_options.go
@@ -1,0 +1,50 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package options
+
+import (
+	"math/rand"
+	"time"
+
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+var internalRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+
+type ClusterOptions []e2e.EPClusterOption
+
+// WithClusterOptionGroups takes an array of EPClusterOption arrays, and randomly picks one EPClusterOption array when constructing the config.
+// This function is mainly used to group strongly coupled config options together, so that we can dynamically test different groups of options.
+func WithClusterOptionGroups(input ...ClusterOptions) e2e.EPClusterOption {
+	return func(c *e2e.EtcdProcessClusterConfig) {
+		optsPicked := input[internalRand.Intn(len(input))]
+		for _, opt := range optsPicked {
+			opt(c)
+		}
+	}
+}
+
+// WithSubsetOptions randomly select a subset of input options, and apply the subset to the cluster config.
+func WithSubsetOptions(input ...e2e.EPClusterOption) e2e.EPClusterOption {
+	return func(c *e2e.EtcdProcessClusterConfig) {
+		// selects random subsetLen (0 to len(input)) elements from the input array.
+		subsetLen := internalRand.Intn(len(input) + 1)
+		perm := internalRand.Perm(len(input))
+		for i := 0; i < subsetLen; i++ {
+			opt := input[perm[i]]
+			opt(c)
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


siyuanfoundation:
cc @serathius

tests/robustness/options/cluster_options_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"math/rand"
+	"testing"
+	"time"
+
+	"go.etcd.io/etcd/server/v3/embed"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestWithClusterOptionGroups(t *testing.T) {
+	Rand = rand.New(rand.NewSource(1))
```
serathius, line 27: Tests should cleanup after each other. Could you restore previous Rand value in defer?

tests/robustness/options/cluster_options.go
```diff
@@ -0,0 +1,50 @@
...
+package options
+
+import (
+	"math/rand"
+	"time"
+
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+var Rand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
serathius, line 24: Does Rand need to be public? I would be worried that it would be mistakenly imported. Maybe name `internalRand` would work better here?

serathius:
Goal of the task was to just add some randomization into configuration of robustness tests, but do we really need code generation to for util functions? I understand that it might be a little tedious to prepare functions to randomize configuration options, however it will be much more tedious to maintain codegen code. 

Reviewing, refactoring and fixing normal Go code is hard enough, I don't think we should complicate things more by requiring contributors to also learn code generation. Could you start from randomizing 1 or 2 options and see where it will take us?

siyuanfoundation:
> Goal of the task was to just add some randomization into configuration of robustness tests, but do we really need code generation to for util functions? I understand that it might be a little tedious to prepare functions to randomize configuration options, however it will be much more tedious to maintain codegen code.
> 
> Reviewing, refactoring and fixing normal Go code is hard enough, I don't think we should complicate things more by requiring contributors to also learn code generation. Could you start from randomizing 1 or 2 options and see where it will take us?

I am adding the code generation not just because it is tedious to prepare the functions. With the new With functions, when someone adds a new server config field, they would potentially need to add 3 different With functions to test it, `tests/framework/config/cluster.go`, `tests/framework/e2e/cluster.go`, and this new place. It would be confusing and adds more friction to development. 
I can start testing with 1 or 2 options. But if we were to adopt this solution, in the long run some automation would be needed.

serathius:
> I am adding the code generation not just because it is tedious to prepare the functions. With the new With functions, when someone adds a new server config field, they would potentially need to add 3 different With functions to test it, tests/framework/config/cluster.go, tests/framework/e2e/cluster.go, and this new place. It would be confusing and adds more friction to development.

You are right it might be tedious, but it's simple and understandable by everyone. I don't think code generation is necessary to replace writing 3 functions

tests/robustness/scenarios.go
```diff
@@ -65,10 +66,10 @@ func scenarios(t *testing.T) []testScenario {
 	}
 	enableLazyFS := e2e.BinPath.LazyFSAvailable()
 	baseOptions := []e2e.EPClusterOption{
-		e2e.WithSnapshotCount(100),
 		e2e.WithGoFailEnabled(true),
-		e2e.WithCompactionBatchLimit(100),
-		e2e.WithWatchProcessNotifyInterval(100 * time.Millisecond),
+		options.WithSnapshotCount(100),
+		options.WithExperimentalCompactionBatchLimit(100),
+		options.WithExperimentalWatchProgressNotifyInterval(100 * time.Millisecond),
```
serathius, line 72: This PR adds a lot of util code, but at the end doesn't implement any randomizing. Working in isolation can lead to design straying away from the goal. 

Could we start from making 1 or 2 parameters random before we add 250 lines of util functions and tests?

serathius (approved):

tests/robustness/options/cluster_options_test.go
```diff
@@ -0,0 +1,108 @@
...
+func resetRand() {
+	internalRand = rand.New(rand.NewSource(1))
+}
+
+func init() {
+	resetRand()
+}
+
+func TestWithClusterOptionGroups(t *testing.T) {
+	defer resetRand()
```
serathius, line 34:
```suggestion

func mockRand(source rand.Source) func() {
  tmpRand = internalRand
  internalRand = rand.New(source)
  return func() {
    internalRand = tmpRand
  }
}

func TestWithClusterOptionGroups(t *testing.T) {
	restore := mockRand(rand.NewSource(1))
	defer restore()
```

serathius: Let's do it as followup.

serathius:
Sent https://github.com/etcd-io/etcd/pull/16944

Issue #16865, opened 2023-11-05
Improve comments for clientv3test.TestUnresolvableOrderViolation


lavacat:
### What would you like to be added?

Please review the test and provide summary at the beginning of the test, so it's clear what and how we are testing.

### Why is this needed?

improving readability

shaoqin2:
From Kubecon contribfest: I can take a look at this one

jmhbnz:
Closed, was completed in #16920

PR #16920, opened 2023-11-11
clientv3test: add comments for clientv3test

shaoqin2:
Add tests for `clientv3test.TestUnresolvableOrderViolation` and `clientv3test.TestEndpointSwitchResolvesViolation`

Improve `TestEndpointSwitchResolvesViolation` with a partition resolved return no error case

Addresses https://github.com/etcd-io/etcd/issues/16865

### DIFF
--- a/tests/integration/clientv3/ordering_util_test.go
+++ b/tests/integration/clientv3/ordering_util_test.go
@@ -24,6 +24,9 @@ import (
 	integration2 "go.etcd.io/etcd/tests/v3/framework/integration"
 )
 
+// TestEndpointSwitchResolvesViolation ensures
+// - ErrNoGreaterRev error is returned from partitioned member when it has stale revision
+// - no more error after partition recovers
 func TestEndpointSwitchResolvesViolation(t *testing.T) {
 	integration2.BeforeTest(t)
 	clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 3})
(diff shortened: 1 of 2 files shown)


tests/integration/clientv3/ordering_util_test.go
```diff
@@ -78,8 +81,16 @@ func TestEndpointSwitchResolvesViolation(t *testing.T) {
 	if err != ordering.ErrNoGreaterRev {
 		t.Fatal("While speaking to partitioned leader, we should get ErrNoGreaterRev error")
 	}
+
+	clus.Members[2].RecoverPartition(t, clus.Members[:2]...)
+	time.Sleep(1 * time.Second) // give enough time for the operation
```
fuweid, line 86: NOTE: The `orderingKv.Get` doesn't have backoff try. It might make the test case flaky since the CI VM is unstable and the 1 second might be not enough for member[2] to get latest data.

> Sorry, I don't have good idea to ensure the data is synced from leader. Maybe we can remove `clientv3.WithSerializable()` option to force to use linear read.

shaoqin2: I think when creating this orderingKv object, we passed the retry function `NewOrderViolationSwitchEndpointClosure`
```
	orderingKv := ordering.NewKV(cli.KV, ordering.NewOrderViolationSwitchEndpointClosure(cli))
```
when the Get fails, we will enter the retry based on
```
		err = kv.orderViolationFunc(op, r, prevRev)
		if err != nil {
			return nil, err
		}
```
Am I understanding this correctly?

fuweid: Sorry for unclear comment. 

The `orderingKv` did retry immediately. 

However, the Github Action CI VM is unstable. When you recover member[2] from network partition and sleep 1s, it doesn't guarantee the member[2] can sync with leader during ordering's retry. 

The Line 87 `_, err = orderingKv.Get(ctx, "foo", clientv3.WithSerializable())` still might return error.

shaoqin2: Ah that make sense. The idea is with default linearizable it will be forced to talk to the other members to ensure the most recent data is returned when `Get` is called right?

Removed with `WithSerializable` option from the call. Let me know if looks good

fuweid (approved):
LGTM, Thanks!

ping @serathius @ahrtr to trigger the CI~

jmhbnz (approved):
LGTM - Thanks @shaoqin2

ahrtr (approved):

PR #16942, opened 2023-11-14
Update dependabot.yml

zhuxiaow0:
Enable Dependabot for Dockerfile and go.mod updates.
This will update images in Dockerfile FROM statements, and go modules.

This PR is created by a script. Please let me know if we should modify any values.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,21 +1,99 @@
 ---
 version: 2
 updates:
-  - package-ecosystem: github-actions
-    directory: /
-    schedule:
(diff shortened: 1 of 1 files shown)


jmhbnz:
Closing - Automated pr from script. Changes have not been discussed in an issue beforehand and `gomod` sections do not appear to be correct for etcd multi module dependency management.

In my view the addition of `docker` section is of minimal value given the type of tag we use and how infrequently we update distroless operating system version.

Happy to re-open this if a quorum of other reviewers have a different view.

PR #16949, opened 2023-11-15
tests: fix failed prow test by adding golangci-lint to Makefile

etsrpl:
Missing binary, fixes golangci-lint command not found on prow jobs

Attempt to fix https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/16944/pull-etcd-verify-lint/1724720730186190848


upodroid:
you need to issue /ok-to-test command

serathius (approved):

Makefile
```diff
@@ -152,6 +152,14 @@ endif
 
 # Tools
 
+GOLANGCI_LINT_VERSION = "v1.55.2"
```
ahrtr, line 155: We need to **manually** maintain the golangci version in **two** places,
https://github.com/etcd-io/etcd/blob/0c76f50dbce39591071c2c9df241115e5352e6dd/.github/workflows/static-analysis.yaml#L18

Propose to maintain the version in **one** place `tools/mod` so that it can be maintained **automatically** by `dependabot`, similar to `gofail` https://github.com/etcd-io/etcd/blob/0c76f50dbce39591071c2c9df241115e5352e6dd/tools/mod/go.mod#L15

serathius: If I understand correctly you propose to:
* Remove the golangci-lint step from https://github.com/etcd-io/etcd/blob/0c76f50dbce39591071c2c9df241115e5352e6dd/.github/workflows/static-analysis.yaml#L15-L19
* Change the installation method of golanci-lint to `go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2`  from the curl.
* Use `tools/mod` to store version golangci-lint

Overall SGTM, @etsrpl would you be able to make the changes?

ahrtr: Sorry for the confusion. I didn't propose to remove the golangci-lint step.

What I proposed are:
- Define the golangci-lint version in `tools/mod`, similar to what we have done for `gofail`, so that it can be automatically maintained by dependabot, https://github.com/etcd-io/etcd/blob/0c76f50dbce39591071c2c9df241115e5352e6dd/tools/mod/go.mod#L15
- get the version from `tools/mod` instead of hard code the version in the following two places,
  -  https://github.com/etcd-io/etcd/blob/0c76f50dbce39591071c2c9df241115e5352e6dd/.github/workflows/static-analysis.yaml#L18
    - here (line 155 in Makefile in this PR)
   
refer to https://github.com/etcd-io/etcd/blob/0c76f50dbce39591071c2c9df241115e5352e6dd/tests/robustness/makefile.mk#L35

serathius: > Sorry for the confusion. I didn't propose to remove the golangci-lint step.

Ok, but to maintain single source of truth for golangci-lint, without removing the workflow step we would need to implement templating. You didn't mention how version in `tools/mod` will be provided to workflow so I assumed you meant to delete it. 

Both options are ok, but I implementing workflow templating adds another layer of complexity that might be too much for a new contributor.

serathius: Our goal was to just fix the Kubernetes CI :P

ahrtr: OK, let's merge this for now, and fix the proposal separately.

ahrtr (approved):
lgtm

PR #16954, opened 2023-11-16
Get golangci-lint't version from `tools/mod` for `golangci/golangci-lint-action`

ahrtr:


.github/workflows/static-analysis.yaml
```diff
@@ -12,10 +12,12 @@ jobs:
       - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
         with:
           go-version: ${{ steps.goversion.outputs.goversion }}
+      - id: golangci-lint-version
+        run: echo "golangci-lint-version=$(cd tools/mod && go list -m -f {{.Version}} github.com/golangci/golangci-lint)" >> "$GITHUB_OUTPUT"
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
-          version: v1.55.2
+          version: ${{ steps.golangci-lint-version.outputs.golangci-lint-version }}
```
serathius, line 20: Would be surprised that dashes in the name work. I expect that this is implemented very naively via shell, which doesn't support it. Still let's see.

serathius: Nvm works https://screenshot.googleplex.com/7ihYdRwdXKE2CkE

ahrtr: FYI.
I also see example something like `SELECTED_COLOR: ${{ steps.color-selector.outputs.SELECTED_COLOR }}` in https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter

serathius: Ye, exactly. Even though workflow id uses a dash `color-selector`, when defining variable it changes it to be shell compatible `SELECTED_COLOR` by replacing it with an underscore.

ahrtr: all changed to underscore, thx

PR #16955, opened 2023-11-16
faulty message transition for robustness test

joshuazh-x:
etcd shall tolerate various message transition issues (caused by whatever reason) including message delay/drop/reordering/duplicate. So far we only inject delay in network TCP layer and it does not impact message orders and redundancy.

However it is not easy to implement such message level faults in proxy. So I make this change to inject such faults in encoder level. In such a way we can simulate various transition faults in message level.

### DIFF
--- a/server/etcdserver/api/rafthttp/faulty_encoder.go
+++ b/server/etcdserver/api/rafthttp/faulty_encoder.go
@@ -1,0 +1,635 @@
+// Copyright 2015 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


serathius:
Thanks for contributing to robustness tests!
I'm still little bit confused why implementing this on proxy side is harder. My plan was to keep the proxy generic and then at some point replace it with some existing solution so we don't need to build and maintain a low level network code.

Currently proxy implements both delay, and drop. Can you clarify why implementing reorder and duplicate would be hard?

joshuazh-x:
From my understanding, etcd sends/receives messages via http streaming. In proxy layer we can delay or drop packets but we cannot tell if that data is the whole part of a message, part of one or many messages.

serathius:
>  In proxy layer we can delay or drop packets but we cannot tell if that data is of the whole a message, partial of one or many messages.

True, but network doesn't work on messages, but on packets. If you are manipulating messages, you are not simulating a network failure. 

Not a network export, would want to get @aojea about best ways to simulate network failures.

joshuazh-x:
@serathius you're right, this is not a network fault. I tried to avoid using network failure in this test and use message transition instead. The purpose of this proposal is to ensure that etcd is implemented in a way that its correctness does not rely on how the messages are sent/received. 

I notice message transportation in etcd is implemented in strict ordering. Not sure if reordering is possible in current etcd. But I think this could be helpful but can stay in a low priority in case someday we may want to optimize transport logic.

joshuazh-x:
update PR description to avoid misleading.

PR #16959, opened 2023-11-16
etcdserver: move rpc defrag notifier into backend.

siyuanfoundation:

the problem with the current implementation is:
1. there could potentially be >1 rpc servers (1 for secure, 1 for insecure). 
2. each rpc server only sets its serve state based on defrag requests to its own server, unaware of each other. So one rpc server could still be serving traffic while the other requested defrag.

Based on these reasons, I think it is necessary to move the defrag notifiers further back to the backend.

### DIFF
--- a/server/storage/backend/backend.go
+++ b/server/storage/backend/backend.go
@@ -71,6 +71,8 @@ type Backend interface {
 
 	// SetTxPostLockInsideApplyHook sets a txPostLockInsideApplyHook.
 	SetTxPostLockInsideApplyHook(func())
+
+	SubscribeDefragNotifier(notifier DefragNotifier)
 }
 
 type Snapshot interface {
@@ -80,6 +82,11 @@ type Snapshot interface {
 	WriteTo(w io.Writer) (n int64, err error)
 	// Close closes the snapshot.
 	Close() error
+}
+
+type DefragNotifier interface {
+	DefragStarted()
+	DefragFinished()
 }
 
 type txReadBufferCache struct {
@@ -127,6 +134,9 @@ type backend struct {
 	txPostLockInsideApplyHook func()
 
 	lg *zap.Logger
+
+	defragNotifiers []DefragNotifier
+	notifierMu      sync.RWMutex
 }
 
 type BackendConfig struct {
@@ -445,6 +455,24 @@ func (b *backend) Defrag() error {
 	return b.defrag()
 }
 
+func (b *backend) SubscribeDefragNotifier(notifier DefragNotifier) {
+	b.notifierMu.Lock()
+	defer b.notifierMu.Unlock()
+	b.defragNotifiers = append(b.defragNotifiers, notifier)
+}
+
+func (b *backend) defragStarted() {
+	for _, notifier := range b.defragNotifiers {
+		notifier.DefragStarted()
+	}
+}
+
+func (b *backend) defragFinished() {
+	for _, notifier := range b.defragNotifiers {
+		notifier.DefragFinished()
+	}
+}
+
 func (b *backend) defrag() error {
 	now := time.Now()
 	isDefragActive.Set(1)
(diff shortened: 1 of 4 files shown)


siyuanfoundation:
cc @chaochn47 @serathius

server/storage/backend/backend.go
```diff
@@ -445,6 +455,27 @@ func (b *backend) Defrag() error {
 	return b.defrag()
 }
 
+func (b *backend) SubscribeDefragNotifier(notifier DefragNotifier) {
+	if notifier == nil {
+		return
+	}
```
serathius, line 461: Why this check? It doesn't protect against nil pointer exceptions.

It will work for 
```
b.SubscribeDefragNotifier(nil)
```
But not for 
```
var notifier *healthNotifier
b.SubscribeDefragNotifier(notifier)
```

Because `nil != (*healthNotifier)(nil)`

server/storage/backend/backend.go
```diff
@@ -459,6 +490,9 @@ func (b *backend) defrag() error {
 	// lock database after lock tx to avoid deadlock.
 	b.mu.Lock()
 	defer b.mu.Unlock()
+	// send notifications after acquiring the lock.
+	b.defragStarted()
```
serathius, line 494: Notifying listeners under lock is a simple and correct way to fix the race issue of two goroutines calling Defrag at once. However the current implementation comes with one flaw, it introduces an external call under a lock. 

Backend just calls and blocks on executing the notifiers, but doesn't really know what they are doing. We should be really careful about introducing a blocking call here. Do you know if call `.SetServingStatus(allGRPCServices, healthpb.HealthCheckResponse_SERVING)` is blocking?

siyuanfoundation: The `SetServingStatus ` call is not blocking. It sends update to a channel, but it clears the channel first. And the lock is used only in simple cases. 
```
// SetServingStatus is called when need to reset the serving status of a service
// or insert a new service entry into the statusMap.
func (s *Server) SetServingStatus(service string, servingStatus healthpb.HealthCheckResponse_ServingStatus) {
	s.mu.Lock()
	defer s.mu.Unlock()
	if s.shutdown {
		logger.Infof("health: status changing for %s to %v is ignored because health service is shutdown", service, servingStatus)
		return
	}

	s.setServingStatusLocked(service, servingStatus)
}

func (s *Server) setServingStatusLocked(service string, servingStatus healthpb.HealthCheckResponse_ServingStatus) {
	s.statusMap[service] = servingStatus
	for _, update := range s.updates[service] {
		// Clears previous updates, that are not sent to the client, from the channel.
		// This can happen if the client is not reading and the server gets flow control limited.
		select {
		case <-update:
		default:
		}
		// Puts the most recent update to the channel.
		update <- servingStatus
	}
}
```

Do you think we should make the `defragStarted ` call in a separate routine to remove such concerns?

server/storage/backend/backend.go
siyuanfoundation: this is to protect the line
```
mc := adapter.MaintenanceServerToMaintenanceClient(v3rpc.NewMaintenanceServer(s, nil))
```

Also, why would not it work for 
```
var notifier *healthNotifier
b.SubscribeDefragNotifier(notifier)
```

I tried 
```
type Person struct {
	Name string
	Age  int
}

var persons []*Person

func subscribePerson(p *Person) {
	if p == nil {
		fmt.Printf("p is nil\n")
	} else {
		fmt.Printf("p is not nil\n")
		persons = append(persons, p)
	}
}

func main() {
	var p1 *Person
	subscribePerson(p1)
	fmt.Printf("len(persons) = %d\n", len(persons))
}
```
It works with output `len(persons) = 0`

ahrtr:
close/reopen to re-trigger all arm64 workflows

siyuanfoundation:
closing this due to no real reported issue on this yet.

PR #16966, opened 2023-11-19
Remove v2 store apply logic

serathius:


serathius:
Validated that we cannot just remove v2 apply logic as v3.5 still uses v2 requests to set member attributes and cluster version.

PR #16984, opened 2023-11-20
etcdserver: add linearizable_read check to readyz.

siyuanfoundation:

Follow up of https://github.com/etcd-io/etcd/pull/16792:

Because `ReadIndex check` requires non-trivial changes, it would be difficult to backport. 
To unblock livez/readyz, a fallback to the old `linearizable_read` check would meet the MVP, the new `livez/readyz` checks would be no worse than the old `/heath` checks, and minimize the risk for backporting.

### DIFF
--- a/server/etcdserver/api/etcdhttp/health.go
+++ b/server/etcdserver/api/etcdhttp/health.go
@@ -239,14 +239,19 @@ type CheckRegistry struct {
 
 func installLivezEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
 	reg := CheckRegistry{checkType: checkTypeLivez, checks: make(map[string]HealthCheck)}
-	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.Register("serializable_read", readCheck(server, true /* serializable */))
 	reg.InstallHttpEndpoints(lg, mux)
 }
 
 func installReadyzEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
 	reg := CheckRegistry{checkType: checkTypeReadyz, checks: make(map[string]HealthCheck)}
 	reg.Register("data_corruption", activeAlarmCheck(server, pb.AlarmType_CORRUPT))
-	reg.Register("serializable_read", serializableReadCheck(server))
+	// serializable_read checks if local read is ok.
+	// linearizable_read checks if there is consensus in the cluster.
+	// Having both serializable_read and linearizable_read helps isolate the cause of problems if there is a read failure.
+	reg.Register("serializable_read", readCheck(server, true))
+	// linearizable_read check would be replaced by read_index check in 3.6
+	reg.Register("linearizable_read", readCheck(server, false))
 	reg.InstallHttpEndpoints(lg, mux)
 }
 
@@ -410,13 +415,10 @@ func activeAlarmCheck(srv ServerHealth, at pb.AlarmType) func(context.Context) e
 	}
 }
 
-func serializableReadCheck(srv ServerHealth) func(ctx context.Context) error {
+func readCheck(srv ServerHealth, serializable bool) func(ctx context.Context) error {
 	return func(ctx context.Context) error {
 		ctx = srv.AuthStore().WithRoot(ctx)
-		_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: true})
-		if err != nil {
-			return fmt.Errorf("range error: %w", err)
-		}
-		return nil
+		_, err := srv.Range(ctx, &pb.RangeRequest{KeysOnly: true, Limit: 1, Serializable: serializable})
+		return err
 	}
 }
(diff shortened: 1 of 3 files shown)


siyuanfoundation:
cc @chaochn47 @serathius @ahrtr

tests/e2e/http_health_check_test.go
```diff
@@ -226,22 +226,54 @@ func TestHTTPLivezReadyzHandler(t *testing.T) {
...
+						`[-]linearizable_read failed: etcdserver: leader changed`,
+					},
+				},
+			},
 		},
 		{
 			name:           "raft loop deadlock",
 			injectFailure:  triggerRaftLoopDeadLock,
 			clusterOptions: []e2e.EPClusterOption{e2e.WithClusterSize(1), e2e.WithGoFailEnabled(true)},
 			// TODO expected behavior of livez check should be 503 or timeout after RaftLoopDeadLock check is implemented.
```
serathius, line 265: TODO not needed as now we already return 503. We don't need to implement dedicated check for scenarios that are already covered by other checks.

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -239,14 +239,15 @@ type CheckRegistry struct {
...
-	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.Register("serializable_read", readCheck(server, true /* serializable */))
 	reg.InstallHttpEndpoints(lg, mux)
 }
 
 func installReadyzEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
 	reg := CheckRegistry{checkType: checkTypeReadyz, checks: make(map[string]HealthCheck)}
 	reg.Register("data_corruption", activeAlarmCheck(server, pb.AlarmType_CORRUPT))
-	reg.Register("serializable_read", serializableReadCheck(server))
+	reg.Register("serializable_read", readCheck(server, true /* serializable */))
```
serathius, line 249: Would be good to understand motivation of having `serializable_read` here as technically it's already covered by `linearizable_read`. Would also recommend leaving a comment with explanation for posterity.

tests/e2e/http_health_check_test.go
siyuanfoundation: this TODO is for livez check. There is not check to cover raft deadlock in livez yet.

server/etcdserver/api/etcdhttp/health.go
siyuanfoundation: done

tests/e2e/http_health_check_test.go
serathius: Oops, missed that.

serathius (approved):

server/etcdserver/api/etcdhttp/health_test.go
```diff
@@ -39,13 +39,17 @@ import (
...
+	serializableReadError error
+	linearizableReadError error
+	missingLeader         bool
+	authStore             auth.AuthStore
 }
 
-func (s *fakeHealthServer) Range(_ context.Context, _ *pb.RangeRequest) (*pb.RangeResponse, error) {
-	return nil, s.apiError
+func (s *fakeHealthServer) Range(_ context.Context, req *pb.RangeRequest) (*pb.RangeResponse, error) {
+	if req.Serializable || s.serializableReadError != nil {
```
fuweid, line 49: it should be like 

```
if req.Serializable {
   return nil, s.serializableReadError
}
```

Is it correct?

fuweid (approved):
LGTM with one question

server/etcdserver/api/etcdhttp/health.go
```diff
@@ -239,14 +239,19 @@ type CheckRegistry struct {
...
 func installReadyzEndpoints(lg *zap.Logger, mux *http.ServeMux, server ServerHealth) {
 	reg := CheckRegistry{checkType: checkTypeReadyz, checks: make(map[string]HealthCheck)}
 	reg.Register("data_corruption", activeAlarmCheck(server, pb.AlarmType_CORRUPT))
-	reg.Register("serializable_read", serializableReadCheck(server))
+	// serializable_read checks if local read is ok.
+	// linearizable_read checks if there is consensus in the cluster.
+	// Having both serializable_read and linearizable_read helps isolate the cause of problems if there is a read failure.
+	reg.Register("serializable_read", readCheck(server, true /* serializable */))
+	// linearizable_read check would be replaced by read_index check in 3.6
+	reg.Register("linearizable_read", readCheck(server, false /* serializable */))
```
ahrtr, line 254:
```suggestion
	reg.Register("linearizable_read", readCheck(server, false /* linearizable */))
```

siyuanfoundation: Removed.

server/etcdserver/api/etcdhttp/health_test.go
siyuanfoundation: My original thought is that if there is serializable error, linearizable read should fail as well. But that could be confusing. 
Changed for clarity.

ahrtr (approved):
lgtm

Issue #16969, opened 2023-11-20
Disable forwarding redirect for peer's HTTP communication


ahrtr:
### What would you like to be added?

It's possible that etcd server may run into SSRF (Server-side request forgery) when adding a new member. If users provide a malicious peer URL, the existing etcd members may be redirected to other unexpected internal URL when getting the new member's version,

https://github.com/etcd-io/etcd/blob/a9cf27b169b1fc89bc0189ffaea7d21c52856b24/server/etcdserver/cluster_util.go#L250

We should disable the forwarding redirect for the endpoint `/version`, as there is no need to support it.
```
client := &http.Client{
    Transport: rt,
    Timeout:   timeout,
    CheckRedirect: func(req *http.Request, via []*http.Request) error {
        return http.ErrUseLastResponse
    },
}
```

We should also evaluate all peer's HTTP communication, and disable the redirect as well if needed.
https://github.com/etcd-io/etcd/blob/d0114cf4c0ab8b491e7ef5c8bda8222983d05c93/server/etcdserver/api/etcdhttp/peer.go#L61-L75

### Why is this needed?

Improve security

jmhbnz:
Discussed during sig etcd triage meeting. Assigned to @ivanvc to investigate remaining etcd peer http interactions.

ivanvc:
Happy to help. Can you assign it to me? Thanks

lavacat:
Another area to check out are endpoints from `Transport.Hander`
https://github.com/etcd-io/etcd/blob/904c0769e9aaf4b676bda8ff2a866e558a40270b/server/etcdserver/api/rafthttp/transport.go#L157C1-L166C12

ivanvc:
If I understand correctly, we need to update the HTTP client from the following locations:
* Promote Member: https://github.com/etcd-io/etcd/blob/3b37afe/server/etcdserver/cluster_util.go#L292
* Lease
  * Renew: https://github.com/etcd-io/etcd/blob/3b37afe/server/lease/leasehttp/http.go#L153
  * TTL: https://github.com/etcd-io/etcd/blob/3b37afe/server/lease/leasehttp/http.go#L213
 * Downgrade Enabled: https://github.com/etcd-io/etcd/blob/3b37afe/server/etcdserver/cluster_util.go#L362-365
 * Hash KV: https://github.com/etcd-io/etcd/blob/3b37afe/server/etcdserver/corrupt.go#L470

[...]

PR #16986, opened 2023-11-20
Disable following redirects when checking peer urls

jmhbnz:
It's possible that etcd server may run into a server side request forgery situation when adding a new member. If users provide a malicious peer URL, the existing etcd members may be redirected to other unexpected internal URL when getting the new member's version.

One component of #16969


serathius (approved):

ahrtr (approved):
lgtm

Thanks

PR #16989, opened 2023-11-22
Remove v2 apply logic

serathius:
v2 store is no longer available in v3.6.
We can remove apply logic for it as they will never be used.

Only V2 PUT is needed as it applies to v3 storage and etcd v3.5 uses it for setting member attributes and cluster version.

I'm proposing etcd v3.6 to panic when encountering v2 store change. This can only happened if user has run v3.5 with `--enable-v2`. As documented in `--v2-deprecation=write-only` (default in etcd v3.6) will not allow custom v2 state. 
https://github.com/etcd-io/etcd/blob/fd0882b67e6f5238cbee9a75688537864c0deb5c/server/etcdmain/help.go#L169 

If etcd still has some v2 data, users are expected to either:
* Halt the upgrade and reconsider what to do with v2 state.
* Run etcd with `--v2-deprecation=write-only-drop-data` which is expected to delete the data. Implementation pending.

Note for reviewers: 
* We have dedicated tests for v2 cluster membership change and v2 cluster version setting. https://github.com/etcd-io/etcd/pull/16994
* We have dedicated tests for running hybrid clusters (mixed v3.6 and v3.5) https://github.com/etcd-io/etcd/blob/28d95649620a8961a61ec12de6365f036df14a1b/tests/common/e2e_test.go#L55-L73


fuweid (approved):
LGTM

fuweid:
```
FAIL: (code:1):
  % (cd server && 'golangci-lint' 'run' '--config' '/home/runner/work/etcd/etcd/tools/.golangci.yaml' './...')
Error: etcdserver/apply_v2.go:119:6: func `toResponse` is unused (unused)
func toResponse(ev *v2store.Event, err error) Response {
FAIL: 'run golangci-lint run --config /home/runner/work/etcd/etcd/tools/.golangci.yaml ./...' checking failed (!=0 return code)
```

Need to remove the unused function `toResponse`.

ahrtr:
Could you breakdown this PR into 2+ PRs? 
- This PR only focus on removing v2 apply logic (excluding Only V2 PUT, because it's still needed when in mix-version cluster). It should can be merged soon once the workflow is green.
- Keep all other changes in separate PRs, which need careful review.

server/etcdserver/apply_v2.go
```diff
@@ -121,6 +121,9 @@ func (a *applierV2store) Sync(r *RequestV2) Response {
 // applyV2Request interprets r as a call to v2store.X
 // and returns a Response interpreted from v2store.Event
 func (s *EtcdServer) applyV2Request(r *RequestV2, shouldApplyV3 membership.ShouldApplyV3) (resp Response) {
+	if r.Method != "PUT" && (storeMemberAttributeRegexp.MatchString(r.Path) || r.Path == membership.StoreClusterVersionKey()) {
+		s.lg.Panic("detected disallowed v2 WAL for stage --v2-deprecation=write-only", zap.String("method", r.Method))
+	}
```
chaochn47, line 126: From the description of this PR and understood the mix version of 3.5 and 3.6, this `v2PutMemberAttribute` is required to publish 3.5 server readiness, the following condition makes more sense, wdyt? 

```suggestion
	v2PutMemberAttribute = r.Method == "PUT" && (storeMemberAttributeRegexp.MatchString(r.Path) || r.Path == membership.StoreClusterVersionKey()) 
	if !v2PutMemberAttribute {
		s.lg.Panic("detected disallowed v2 WAL for stage --v2-deprecation=write-only", zap.String("method", r.Method))
	}
```

serathius: Right, make mistake while spitting the PR.

serathius:
All the subsequent PRs were merged. ping @ahrtr for review.

chaochn47 (approved):

ahrtr (approved):

tests/e2e/v2store_deprecation_test.go
```diff
@@ -58,6 +58,34 @@ func TestV2DeprecationNotYet(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestV2DeprecationWriteOnlyWAL(t *testing.T) {
```
ahrtr, line 61: Note: this case won't work anymore if we bootstrap etcd based on v3store. But that isn't finalized yet, and may not be done in 3.6.  But please let's add a comment/note.

serathius: We might need to update the test, however the test is still relevant.

PR #17015, opened 2023-11-24
Extract membership applier

serathius:
Cluster membership changes can be done via both `EntryNormal` (setting attribute on member) and `EntryConfChange` (adding member). This made it natural to keep them in separate code paths, but it resulted that changes to membership.Cluster can go through totally separate paths
* applyEntryNormal -> uberApplier -> corruptApplier -> authApplier ->  backendApplier -> membership.Cluster
* applyConfChange -> membership.Cluster

This has lead to very weird `shouldApplyV3` logic required to handle v2 and v3 synchronization.

My proposal would be to move the membershipChanges from `applyEntryNormal` and keep them next to `applyConfChange`. The long term goal would be to change `membership.Cluster` into a `membership.Applier`, that would be a single state machine handling all changes to cluster members.

 This PR makes the first step and illustrates how much `shouldApplyV3` is passed through the `v3applier` stack needelesly.

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1913,7 +1913,10 @@ func (s *EtcdServer) applyEntryNormal(e *raftpb.Entry, shouldApplyV3 membership.
 
 func (s *EtcdServer) applyInternalRaftRequest(r *pb.InternalRaftRequest, shouldApplyV3 membership.ShouldApplyV3) *apply.Result {
 	if r.ClusterVersionSet == nil && r.ClusterMemberAttrSet == nil && r.DowngradeInfoSet == nil {
-		return s.uberApply.Apply(r, shouldApplyV3)
+		if !shouldApplyV3 {
+			return nil
+		}
+		return s.uberApply.Apply(r)
 	}
 	membershipApplier := apply.NewApplierMembership(s.lg, s.cluster, s)
 	op := "unknown"
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -60,6 +60,7 @@ import (
 	"go.etcd.io/etcd/server/v3/etcdserver/apply"
 	"go.etcd.io/etcd/server/v3/etcdserver/cindex"
 	"go.etcd.io/etcd/server/v3/etcdserver/errors"
+	"go.etcd.io/etcd/server/v3/etcdserver/txn"
 	serverversion "go.etcd.io/etcd/server/v3/etcdserver/version"
 	"go.etcd.io/etcd/server/v3/lease"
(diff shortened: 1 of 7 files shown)


serathius:

server/etcdserver/server.go
```diff
@@ -1910,6 +1911,37 @@ func (s *EtcdServer) applyEntryNormal(e *raftpb.Entry, shouldApplyV3 membership.
...
+		membershipApplier.ClusterMemberAttrSet(r.ClusterMemberAttrSet, shouldApplyV3)
+	case r.DowngradeInfoSet != nil:
+		op = "DowngradeInfoSet" // Implemented in 3.5.x
+		membershipApplier.DowngradeInfoSet(r.DowngradeInfoSet, shouldApplyV3)
+	default:
+		s.lg.Panic("not implemented apply", zap.Stringer("raft-request", r))
+		return nil
+	}
+	return &apply.Result{}
+}
```
ahrtr, line 1943:
```suggestion
func (s *EtcdServer) applyInternalRaftRequest(r *pb.InternalRaftRequest, shouldApplyV3 membership.ShouldApplyV3) *apply.Result {
	membershipApplier := apply.NewApplierMembership(s.lg, s.cluster, s)
	op := "unknown"
	defer func(start time.Time) {
		txn.ApplySecObserve("v3", op, true, time.Since(start))
		txn.WarnOfExpensiveRequest(s.lg, s.Cfg.WarningApplyDuration, start, &pb.InternalRaftStringer{Request: r}, nil, nil)
	}(time.Now())
	switch {
	case r.ClusterVersionSet != nil:
		op = "ClusterVersionSet" // Implemented in 3.5.x
		membershipApplier.ClusterVersionSet(r.ClusterVersionSet, shouldApplyV3)
		return &apply.Result{}
	case r.ClusterMemberAttrSet != nil:
		op = "ClusterMemberAttrSet" // Implemented in 3.5.x
		membershipApplier.ClusterMemberAttrSet(r.ClusterMemberAttrSet, shouldApplyV3)
		return &apply.Result{}
	case r.DowngradeInfoSet != nil:
		op = "DowngradeInfoSet" // Implemented in 3.5.x
		membershipApplier.DowngradeInfoSet(r.DowngradeInfoSet, shouldApplyV3)
		return &apply.Result{}
	}

	if !shouldApplyV3 {
		return nil
	}
	return s.uberApply.Apply(r)
}
```

serathius: This is incorrect, you add defer call that would update metric to unknown operation.

ahrtr (approved):

PR #17016, opened 2023-11-24
Extract applyEntry from apply

serathius:

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1101,8 +1101,15 @@ func (s *EtcdServer) applyEntries(ep *etcdProgress, apply *toApply) {
 	if len(ents) == 0 {
 		return
 	}
-	var shouldstop bool
-	if ep.appliedt, ep.appliedi, shouldstop = s.apply(ents, &ep.confState, apply.raftAdvancedC); shouldstop {
+	s.lg.Debug("Applying entries", zap.Int("num-entries", len(ents)))
+	var shouldStop bool
+	for i := range ents {
+		e := ents[i]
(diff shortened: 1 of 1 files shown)


server/etcdserver/server.go
```diff
@@ -1771,50 +1771,56 @@ func (s *EtcdServer) apply(
...
-			)
-		}
+		removedSelf := s.applyEntry(&e, confState, raftAdvancedC)
+		shouldStop = shouldStop || removedSelf
 		appliedi, appliedt = e.Index, e.Term
 	}
 	return appliedt, appliedi, shouldStop
 }
 
+func (s *EtcdServer) applyEntry(e *raftpb.Entry, confState *raftpb.ConfState, raftAdvancedC <-chan struct{}) (shouldStop bool) {
```
fuweid, line 1781: IIUC, `shouldStop` should be `removedSelf`

fuweid (approved):
LGTM

server/etcdserver/server.go
serathius: done

ahrtr:
I do not see much value to simplify an already very simple method `apply` (only 50 lines of code, and the structure is also very clear). 

It also increases the depth of call stack.

ahrtr:
It is also weird that `applyEntry` returns a `removedSelf`

serathius:
Goal of this refactor is to extract applyEntry so we can unify the defer logic applying consistent index between two branches (applyNormalEntry and applyConfChange)

fuweid:
> It also increases the depth of call stack.

The go1.21 starts inline everything 😂

PR #17017, opened 2023-11-24
Switch to validating v3 when v2 and v3 are synchronized

serathius:
This PR is not very useful, but let's merge it to clarify how validation should look like as long as we bootstrap from v2 store 

Alternatives proposed before:
* https://github.com/etcd-io/etcd/pull/16084
* https://github.com/etcd-io/etcd/pull/17019

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -303,9 +303,16 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 
 // ValidateConfigurationChange takes a proposed ConfChange and
 // ensures that it is still valid.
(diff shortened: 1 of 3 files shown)


serathius:

ahrtr:
> Alternatives proposed before:
> 
> * [Validate conf change using v3 #16084](https://github.com/etcd-io/etcd/pull/16084)
> * [Stop writing v2store and validate confChange against v3store instead of v2store #17019](https://github.com/etcd-io/etcd/pull/17019)

To be clearer, It isn't an alternative solution. This PR is just a solution on top of https://github.com/etcd-io/etcd/pull/16084 with a further fix (thx for that). I would raise a comment on 16084 and honor the original author's work (@geetasg, thanks for your contribution & **patience**).

ahrtr (approved):

PR #17019, opened 2023-11-25
Stop writing v2store and validate confChange against v3store instead of v2store

ahrtr:
Part of https://github.com/etcd-io/etcd/issues/12913

- We already generate v2 snapshot using v3store. It makes no sense to write any data into v2store, because it isn't used at all.
- We should validate conf change against v3store, which is the source of truth in 3.6.


ahrtr:
https://github.com/etcd-io/etcd/pull/16084 is superseded by this PR.

server/etcdserver/api/membership/cluster.go
```diff
@@ -303,9 +300,13 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 
 // ValidateConfigurationChange takes a proposed ConfChange and
 // ensures that it is still valid.
-func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
-	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
+	// It makes no sense to validate a confChange if we don't apply it.
```
serathius, line 304: Not true, we are still applying confChange as long as we bootstrap from snapshot and not consistent index.

serathius:
Weird 33 tests have failed in prow tests, but none in github workflow tests. Would be good to investigate why there is a difference.

serathius:
> We should validate conf change against v3store, which is the source of truth in 3.6.

v3store will be a source of truth for membership, but it still is not for `confState`. `confState` is still loaded from snapshot and needs to be revalidate and reapplied using v2store. This validation cannot be skipped as it would lead to inconsistency with etcd v3.5. We need to migrate raft bootstrap to start from confstate in v3 store before we stop validating entries.

Let me add a test for this code too.

serathius:
Done, PTAL https://github.com/etcd-io/etcd/pull/17021

server/etcdserver/api/membership/cluster_test.go
```diff
@@ -517,52 +517,6 @@ func TestNodeToMemberBad(t *testing.T) {
 	}
 }
 
-func TestClusterAddMember(t *testing.T) {
```
serathius, line 520: Please migrate such tests v3 instead of just removing them

server/etcdserver/api/membership/cluster.go
ahrtr: Right, this is a longstanding design problem to me.  

# Current status summary
- in older etcd versions ( < 3.4?) we load all data from v2store, and replay WAL entries on top of the snapshot. All work well.
- Starting from 3.5, we load normal data from v3store and membership data still from v2store, and replay WAL entries on top of v2 snapshot.  It works well for conf change. A little ugly for normal key/value data, but it still works.
- Starting from 3.6 (`main` branch for now), we load all data from v3store on startup, but we still replay WAL entries on top of v2 snapshot. 
  - For normal key/value data, etcd skips already applied entries; it works although ugly. 
  - For conf change, the principle of "`skiping already applied entries`" doesn't work anymore. 
    - For etcd, it already loaded membership data from v3store, so etcd can just skip already applied entries. 
    - But raft doesn't know the latest membership data, we still depend on the raft loop to re-apply the conf change (with the old entry index < consistent-index), so as to notify raft all the membership data. So etcdd has to call raft's ApplyConfChange even etcd just skips it


# Solutions
## Short-term solutions
1. Validate the conf change against v2store if it is already in v3store (shouldApplyV3 == false)
2. apply conf change (read from v3store) to raft directly when restarting etcd members to bypass the raft loop? (need investigation)

## Long-term solutions
1. Bootstrap etcd from consistent_index.
   - It works for key/value data.
   - For conf change, it still needs to apply the conf change to raft directly when etcd restarting to bypass the raft loop.
2. Generate v3 snapshot periodically (similar to what etcd did on v2 snapshot), and bootstrap etcd from v3 snapshot, and then replay WAL entries on top of it.
  - need to evaluate the performance of generate v3snapshot on huge db file (e.g. 8GB)
  - The disk space usage will be multiple times bigger, depending on how many snapshot files to keep. 
  - Then why do we need the latest bbolt db file in such case?
4. Generate v3 snapshot periodically (similar to what etcd did on v2 snapshot), but bootstrap etcd from consistent_index. The v3 snapshot files are just for corruption recovery, especially in single-node edge case.

serathius: > Validate the conf change against v2store if it is already in v3store (shouldApplyV3 == false)

I know, https://github.com/etcd-io/etcd/pull/17017

serathius: > apply conf change (read from v3store) to raft directly when restarting etcd members to bypass the raft loop? (need investigation)

I have draft implementation for that.

serathius: > need to evaluate the performance of generate v3snapshot on huge db file (e.g. 8GB)

v3 db snapshot is not a thing. That's an incompatible concept. Recommend reading the raft paper on in-memory snapshotable storage (etcd v2) vs persistent one (etcd v3).

PR #17021, opened 2023-11-26
Test ApplyConfState after restart

serathius:
This test was written to show issue with https://github.com/etcd-io/etcd/pull/17019.
We cannot disable validation on entries before v3 state because we are still using them to update confState, which is still loaded from snapshot and not v3.

### DIFF
--- a/server/etcdserver/server_test.go
+++ b/server/etcdserver/server_test.go
@@ -198,6 +198,115 @@ func TestV2SetClusterVersion(t *testing.T) {
 	}
 }
 
+func TestApplyConfStateWithRestart(t *testing.T) {
+	n := newNodeRecorder()
+	srv := newServer(t, n)
+
+	assert.Equal(t, srv.consistIndex.ConsistentIndex(), uint64(0))
+
+	var nodeID uint64 = 1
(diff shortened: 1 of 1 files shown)


server/etcdserver/server_test.go
```diff
@@ -198,6 +198,115 @@ func TestV2SetClusterVersion(t *testing.T) {
...
+				NodeID: nodeID,
+			}),
+		},
+		{
+			Term:  1,
+			Index: 3,
+			Type:  raftpb.EntryConfChange,
+			Data: pbutil.MustMarshal(&raftpb.ConfChange{
+				Type:    raftpb.ConfChangeAddNode,
+				NodeID:  nodeID,
```
ahrtr, line 239: Note that etcd generates a new memberId each time when users add a new member. It isn't valid to use the same nodeID.

serathius: True, updated to use ConfChangeUpdateNode

serathius:

fuweid (approved):
LGTM

ahrtr (approved):

PR #17022, opened 2023-11-26
Bootstrap raft from 3 backend and not snapshot

serathius:

### DIFF
--- a/server/etcdserver/raftstorage.go
+++ b/server/etcdserver/raftstorage.go
@@ -1,0 +1,276 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package etcdserver
+
+import (
+	"fmt"
+	"sync"
+
+	"go.etcd.io/raft/v3"
+
+	pb "go.etcd.io/raft/v3/raftpb"
+)
+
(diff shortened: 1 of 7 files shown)


server/etcdserver/raftstorage.go
```diff
@@ -0,0 +1,276 @@
...
+	// goroutine.
+	sync.Mutex
+
+	hardState pb.HardState
+	snapshot  pb.Snapshot
+	// ents[i] has raft log position i+snapshot.Metadata.Index
+	ents []pb.Entry
+
+	callStats inMemStorageCallStats
+}
```
ahrtr, line 44: The struct and all its methods are copied from the raft lib. You only added stats on top of that, also added `ApplyConfState`. I am not sure what's the motivation.

ahrtr:
@serathius do you have bandwidth to continue to work on this PR? It's on the top of my review list now.

serathius:
> @serathius do you have bandwidth to continue to work on this PR? It's on the top of my review list now.

I have been doing it in my free hours, so not a lot of bandwidth, but constant work on weekends. The main problem is that I got stuck, after updating all the code I think is required it still fails some tests and I didn't have time to debug it. 

Would love to get your suggestion what I missed.

server/etcdserver/raftstorage.go
serathius: Yea, this is a temporary fork of the file to add the methods I needed to make raft properly bootstrap from v3. I did it so long time ago so I'm not 100% sure that we will need it, it was just simpler to implement it. 

After we make all the tests pass we can rethink if we really need to change raftstorage.

ahrtr:
> I have been doing it in my free hours, so not a lot of bandwidth, but constant work on weekends. The main problem is that I got stuck, after updating all the code I think is required it still fails some tests and I didn't have time to debug it.

Thanks for working on this. Tackling such a complex task requires strong commitment, sustained deep thinking and focus over a long period, and close collaboration from other maintainers. I think progress becomes difficult when any of these elements is missing.

Could you please at least rebase this PR so we can take a look at which tests are currently failing? If you don’t currently have enough time to continue, would you be open to someone else taking over the PR?

serathius:
> If you don’t currently have enough time to continue, would you be open to someone else taking over the PR?

Sure

Issue #17045, opened 2023-11-30
Tune memory allocation of arm64 github actions workflows


jmhbnz:
### What would you like to be added?

The etcd project currently uses managed `arm64` ci runners provided by actuated.dev.  This is defined in our github actions workflow files, for example for e2e:

https://github.com/etcd-io/etcd/blob/51eb29af360ef593fcd53488416473b7993c5c76/.github/workflows/e2e-arm64.yaml#L9-L14

In that example we are requesting 8cpu  and 32gb ram, but it's not clear that we currently need all that ram currently.

We can introduce https://github.com/catchpoint/workflow-telemetry-action to our arm64 workflows to gather simple charts of job memory consumption and then tune each workflow:

![image](https://github.com/etcd-io/etcd/assets/39425256/032aeac8-3f9f-4a65-92bc-fdae624bc6b6)

### Why is this needed?

If we can tune the amount of ram our jobs are requesting closer to what they actually need, we will be better citizens within the actuated managed runner ecosystem and potentially allow for more runner slots to be made available which will reduce how often our jobs get stuck in queues.

moficodes:
@jmhbnz would love to work on this!

fykaa:
can work on this too :)

jmhbnz:
Many thanks @moficodes and @fykaa for volunteering!

There are three files that need to be updated to include the new `workflow-telemetry-action`:
- [ ] https://github.com/etcd-io/etcd/blob/main/.github/workflows/tests-template.yaml
- [ ] https://github.com/etcd-io/etcd/blob/main/.github/workflows/robustness-template.yaml
- [ ] https://github.com/etcd-io/etcd/blob/main/.github/workflows/e2e-arm64.yaml

@moficodes if you can add the initial telemetry action then perhaps @fykaa could follow up with the second pull request after we get some profiling data to update the amount of ram requested for each workflow?

alexellis:
Hi folks,

We have our own variant of metering which has a bit less signal noise if you wanted to try it out for comparison:

https://x.com/alexellisuk/status/1732327834922766785?s=20

https://gist.github.com/alexellis/1f33e581c75e11e161fe613c46180771

@moficodes @jmhbnz did you notice anything telling about RAM usage from the community workflow action you added? We haven't seen much RAM consumed on the CNCF servers when the etcd-io jobs are running, but these two actions run inside the VM and can give a better picture.

Thanks,

Alex

alexellis:
[...]

PR #17046, opened 2023-11-30
add workflow telemetry to collect action metrics

moficodes:
part of #17045

### DIFF
--- a/.github/workflows/e2e-arm64.yaml
+++ b/.github/workflows/e2e-arm64.yaml
@@ -13,6 +13,8 @@ jobs:
         target:
           - linux-arm64-e2e
     steps:
+      - name: Collect Workflow Telemetry
+        uses: catchpoint/workflow-telemetry-action@v1
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - id: goversion
         run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/robustness-template.yaml
+++ b/.github/workflows/robustness-template.yaml
@@ -27,6 +27,8 @@ jobs:
     timeout-minutes: 210
     runs-on: ${{ fromJson(inputs.runs-on) }}
     steps:
+      - name: Collect Workflow Telemetry
+        uses: catchpoint/workflow-telemetry-action@v1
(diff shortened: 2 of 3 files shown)


moficodes:
hmm

The action post failed.

```
Error: [Workflow Telemetry] Unable to report process tracer result
Error: [Workflow Telemetry] Error
Error: Error: ENOENT: no such file or directory, open '/home/runner/actions-runner/_work/_actions/catchpoint/workflow-telemetry-action/v1/dist/proc-tracer/proc-trace.out'
[Workflow Telemetry] Reporting all content ...
Error: [Workflow Telemetry] Resource not accessible by integration
```

jmhbnz:
> @jmhbnz do you know how the metrics become available to view?

Hey @moficodes - I can see some results here for example: https://github.com/etcd-io/etcd/actions/runs/7051472834/attempts/1#summary-19194441512 

Can you check across all three (e2e, tests and robustness) and confirm we have results available for `arm64` runs?  If so I think we are looking ok to merge and collect some data for a few days before making a decision on reducing requested ram.

jmhbnz (approved):
Thanks @moficodes - I would like to collect some samples for a few days to make sure there are no outliers.

This pr LGTM so let's merge, collect some data for a few days and then @fykaa will raise the follow-up pr to reduce the ram requests and disable the telemetry.

jmhbnz:
> Or is it dynamically provisioned based on `actuated-arm64-8cpu-32gb` and we can instead do something like `actuated-arm64-4cpu-8gb` and it will just work.

Based on https://github.com/etcd-io/etcd/pull/16801#issuecomment-1773737853 I believe there is dynamic sizing/provisioning so we can just set the value and it will just work.

ahrtr:
Saw the following errors in "Post Collect Workflow Telemetry", but it seems the reports were already successfully uploaded,

```
Error: [Workflow Telemetry] Unable to report process tracer result
Error: [Workflow Telemetry] Error
Error: Error: ENOENT: no such file or directory, open '/home/runner/actions-runner/_work/_actions/catchpoint/workflow-telemetry-action/v1/dist/proc-tracer/proc-trace.out'
[Workflow Telemetry] Reporting all content ...
Error: [Workflow Telemetry] Resource not accessible by integration
```

PR #17047, opened 2023-12-01
Adding automated install of Protoc if not found on the system

iiamabby:
@jmhbnz and I are working on refining the test-infra prow workflow to harmonise with the legacy GitHub Actions workflow.

We noticed the [test-infra/config/jobs/etcd/etcd-presubmits.yaml](https://github.com/kubernetes/test-infra/blob/551ef90f7a3efa1560b593bf0c6eea7290bee516/config/jobs/etcd/etcd-presubmits.yaml#L69) is using `make verify-lint` whilst the [etcd/.github/workflows/static-anaylsis.yaml](https://github.com/etcd-io/etcd/blob/db4e95cb1f8d3a8eebfb89c880d5fbd8d6165292/.github/workflows/static-analysis.yaml#L37) is using `make verify`

We want to make a change from `make verify-lint` to `make verify` within the test-infra repo 

When testing the `make verify` command in the test-infra image: `gcr.io/k8s-staging-test-infra/kubekins-e2e:v20231122-5f461e0995-master` there was an error relating to `protoc-3.20.3` not been installed which immediately exists with code `255`

This proposed change adds the automation of installing `protoc-3.20.3` depending on the architecture of the system if its not found.

### DIFF
--- a/scripts/genproto.sh
+++ b/scripts/genproto.sh
@@ -25,8 +25,28 @@ fi
 source ./scripts/test_lib.sh
 
 if [[ $(protoc --version | cut -f2 -d' ') != "3.20.3" ]]; then
-  echo "could not find protoc 3.20.3, is it installed + in PATH?"
-  exit 255
+  echo "Could not find protoc 3.20.3, installing now..."
+
+  arch=$(go env GOARCH)
+
+  case ${arch} in
+    "amd64") file="x86_64" ;;
+    "arm64") file="aarch_64" ;;
+    *)
+      echo "Unsupported architecture: ${arch}"
+      exit 255
+      ;;
+  esac
+
+  download_url="https://github.com/protocolbuffers/protobuf/releases/download/v3.20.3/protoc-3.20.3-linux-${file}.zip"
+  echo "Running on ${arch}."
+  wget ${download_url} && unzip -p protoc-3.20.3-linux-${file}.zip bin/protoc > tmpFile && mv tmpFile bin/protoc 
+  rm protoc-3.20.3-linux-${file}.zip
+  chmod +x bin/protoc
+  PATH=$PATH:$(pwd)/bin
+  export PATH
+  echo "Now running: $(protoc --version)"
+
 fi
 
 GOFAST_BIN=$(tool_get_bin github.com/gogo/protobuf/protoc-gen-gofast)


scripts/genproto.sh
```diff
@@ -25,8 +25,31 @@ fi
...
+    "arm64") file="aarch_64" ;;
+    *)
+      echo "Unsupported architecture: ${arch}"
+      exit 255
+      ;;
+  esac
+
+  download_url="https://github.com/protocolbuffers/protobuf/releases/download/v3.20.3/protoc-3.20.3-linux-${file}.zip"
+
+  if [[ ${download_url} ]]; then
```
jmhbnz, line 43: Hey @iiamabby - It looks like this second if condition might be redundant because `download_url` is always set?

Is there some reason I'm missing why this second if condition is required?

iiamabby: no you're right, not needed here. Have removed :)

scripts/genproto.sh
```diff
@@ -25,8 +25,25 @@ fi
...
+    *)
+      echo "Unsupported architecture: ${arch}"
+      exit 255
+      ;;
+  esac
+
+  download_url="https://github.com/protocolbuffers/protobuf/releases/download/v3.20.3/protoc-3.20.3-linux-${file}.zip"
+  echo "Running on ${arch}."
+  wget ${download_url} && unzip protoc-3.20.3-linux-${file}.zip
+  mv bin/protoc /usr/local/bin/ && chmod +x /usr/local/bin/protoc
```
jmhbnz, line 44: Just want to note that this will fail without sudo. 

I think it would be less invasive to just `chmod` the file in the current working directory.

scripts/genproto.sh
```diff
@@ -25,8 +25,25 @@ fi
...
+  echo "export PATH=$PATH:/usr/local/bin" >> ~/.bashrc
```
jmhbnz, line 45: Also think updating `.bashrc` without explicit permission might be a bit invasive.  Let's just run the `export PATH` command directly based on file in working directory rather than adding it to `.bashrc`.

jmhbnz (changes requested):
Thanks @iiamabby - I think this is really close, just two more suggestions before I think this will be ready to merge.

scripts/genproto.sh
fuweid: +1

Or what about introduce gitignored temporary dir in the repo? Install in the temporary dir and export it in the PATH. No need to worry about .bashrc

scripts/genproto.sh
serathius: Agree, we shouldn't require sudo.

iiamabby: Updated thanks

scripts/genproto.sh
iiamabby: Updated thanks

serathius (approved):

scripts/genproto.sh
```diff
@@ -25,8 +25,27 @@ fi
...
+    "arm64") file="aarch_64" ;;
+    *)
+      echo "Unsupported architecture: ${arch}"
+      exit 255
+      ;;
+  esac
+
+  download_url="https://github.com/protocolbuffers/protobuf/releases/download/v3.20.3/protoc-3.20.3-linux-${file}.zip"
+  echo "Running on ${arch}."
+  wget ${download_url} && unzip protoc-3.20.3-linux-${file}.zip
```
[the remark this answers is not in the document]
jmhbnz: Yeah that's a good catch @fuweid, those extra files will be annoying.  The binary gets extracted to `bin/` in the repo which is `.gitignored` already but those other files are an issue.

How about we change line as follows:

```diff
- wget ${download_url} && unzip protoc-3.20.3-linux-${file}.zip
+ wget ${download_url} && unzip -p protoc-3.20.3-linux-${file}.zip bin/protoc > bin/protoc
+ rm protoc-3.20.3-linux-${file}.zip
```

That way we only extract the file we need from the zip (to a .gitignored location), and clean up the zip once done.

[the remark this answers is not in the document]
iiamabby: Thank you @jmhbnz and @fuweid I have updated. 

One thing to note though is that there was a ShellCheck error `SC2094` so have changed to match the preferred syntax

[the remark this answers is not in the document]
jmhbnz: Thanks Abby - nice work, I think this is all set to merge now 🎉

fuweid (approved):
LGTM

serathius:
cc @ahrtr

ahrtr (approved):

PR #17058, opened 2023-12-04
Rollback the pull/14080: which fixed the potential deadlock

ahrtr:
Rollback https://github.com/etcd-io/etcd/pull/14080

Please read https://github.com/etcd-io/etcd/pull/16035#discussion_r1413820384

### DIFF
--- a/server/lease/lessor.go
+++ b/server/lease/lessor.go
@@ -281,15 +281,15 @@ func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
 	// with longer TTL to reduce renew load.
 	l := NewLease(id, ttl)
 
-	if l.ttl < le.minLeaseTTL {
-		l.ttl = le.minLeaseTTL
-	}
-
 	le.mu.Lock()
 	defer le.mu.Unlock()
 
 	if _, ok := le.leaseMap[id]; ok {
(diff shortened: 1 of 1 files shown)


serathius:
Could you add a test to confirm the fix?

ahrtr:
> Could you add a test to confirm that there is no longer a leak and prevent future regressions?

This PR just rollback, can we do whatever other thing separately?

serathius:
It's a partial rollback, which means it should be treated as normal change.

ahrtr:
I don't understand why we are wasting time on such small thing. 

The part I did not rollback is a trivial change. If you are insist on, I can totally rollback the PR.

ahrtr:
Completely rollbacked, pls let me know if you still have any concern. @serathius

serathius:
> I don't understand why we are wasting time on such small thing.

Reverts are only trivial if they revert us to exactly the same stable point as before. Partial revert moves us to new unknown place. Possibly it is an improvement, but to verify we need to either review it thoroughly or have test shows the improvement.

The revert was motivated by issue that issue slipped our review, meaning it's as trivial as originally assumed. We didn't have any tests before, we should invest in them to reduce the dependency on review.

serathius (approved):

PR #17066, opened 2023-12-04
server: disable redirects in peer communication

### LINKED ISSUE #16969: Disable forwarding redirect for peer's HTTP communication

ivanvc:
Disable following redirects from peer HTTP communication on the client's side. Etcd server may run into SSRF (Server-side request forgery) when adding a new member. If users provide a malicious peer URL, the existing etcd members may be redirected to another unexpected internal URL when getting the new member's version.

Fixes #16969.

### DIFF
--- a/server/etcdserver/cluster_util.go
+++ b/server/etcdserver/cluster_util.go
@@ -72,6 +72,9 @@ func getClusterFromRemotePeers(lg *zap.Logger, urls []string, timeout time.Durat
 	cc := &http.Client{
 		Transport: rt,
 		Timeout:   timeout,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
 	}
 	for _, u := range urls {
 		addr := u + "/members"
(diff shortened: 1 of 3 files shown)


jmhbnz (approved):
LGTM - Thanks @ivanvc I think it's reasonable to disable following redirects for peer communication. To the best of my knowledge etcd peer communication should not require or expect redirects.

serathius (approved):

ahrtr (approved):

Issue #17069, opened 2023-12-05
Add a "memberReplace" e2e test to prevent regression


serathius:
### What would you like to be added?

https://github.com/etcd-io/etcd/commit/4fe46f92030e4381e6f9bf95adbb22a08282d297 introduced regression that was cought by robustness test https://github.com/etcd-io/etcd/issues/17052

Issue was related to how normal etcd operation, replacing a member, and was not specific to robustness testing.
We should have a normal e2e test to run this scenario.

### Why is this needed?

Prevent potential regressions when deprecating v2 store.

ZhouJianMS:
Would like to work on this

liangyuanpeng:
Seems like this is resolved and can be closed.

Issue #17052, opened 2023-12-01
Flaky robustness test


ahrtr:
### Which github workflows are flaking?

https://github.com/etcd-io/etcd/actions/runs/7061287598/job/19229219034?pr=17050

etcd somehow failed during boostraping, and the WAL directory wasn't created at all, so the verification failed with reason "`no such file or directory`"
```
2023-12-01T17:58:04.8181296Z /home/runner/work/etcd/etcd/bin/etcd (TestRobustnessKubernetesHighTrafficClusterOfSize3-test-0) (18330): {"level":"info","ts":"2023-12-01T17:58:04.802824Z","caller":"etcdmain/etcd.go:63","msg":"Running: ","args":["/home/runner/work/etcd/etcd/bin/etcd","--name=TestRobustnessKubernetesHighTrafficClusterOfSize3-test-0","--listen-client-urls=http://localhost:20000","--advertise-client-urls=http://localhost:20000","--listen-peer-urls=https://localhost:20001","--initial-advertise-peer-urls=https://localhost:20003","--initial-cluster-token=new","--data-dir","/tmp/TestRobustnessKubernetesHighTrafficClusterOfSize31653420488/001","--snapshot-count=1000","--peer-cert-file","/home/runner/work/etcd/etcd/tests/fixtures/server.crt","--peer-key-file","/home/runner/work/etcd/etcd/tests/fixtures/server.key.insecure","--peer-trusted-ca-file","/home/runner/work/etcd/etcd/tests/fixtures/ca.crt","--experimental-snapshot-catchup-entries=100","--initial-cluster-token=new","--heartbeat-interval=101","--election-timeout=521","--snapshot-count=1000","--experimental-compaction-batch-limit=100","--experimental-watch-progress-notify-interval=100ms","--initial-cluster=TestRobustnessKubernetesHighTrafficClusterOfSize3-test-0=https://localhost:20003,TestRobustnessKubernetesHighTrafficClusterOfSize3-test-1=https://localhost:20008,TestRobustnessKubernetesHighTrafficClusterOfSize3-test-2=https://localhost:20013","--initial-cluster-state=existing"]}
[...]

PR #17079, opened 2023-12-07
Add member replace e2e test

ZhouJianMS:
This PR address #17069

### DIFF
--- a/tests/e2e/ctl_v3_member_no_proxy_test.go
+++ b/tests/e2e/ctl_v3_member_no_proxy_test.go
@@ -1,0 +1,98 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"context"
+	"math/rand"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/server/v3/etcdserver"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
+)
+
+func TestMemberReplace(t *testing.T) {
+	e2e.BeforeTest(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+
+	epc, err := e2e.NewEtcdProcessCluster(ctx, t)
+	require.NoError(t, err)
+	defer epc.Close()
+
+	memberId := rand.Int() % len(epc.Procs)
+	member := epc.Procs[memberId]
+	memberName := member.Config().Name
+	var endpoints []string
+	for i := 1; i < len(epc.Procs); i++ {
+		endpoints = append(endpoints, epc.Procs[(memberId+i)%len(epc.Procs)].EndpointsGRPC()...)
+	}
+	cc, err := e2e.NewEtcdctl(epc.Cfg.Client, endpoints)
+	require.NoError(t, err)
+
+	c := epc.Etcdctl()
+	memberID, found, err := getMemberIdByName(ctx, c, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, true, "Member not found")
+
+	// Need to wait health interval for cluster to accept member changes
+	time.Sleep(etcdserver.HealthInterval)
+
+	t.Logf("Removing member %s", memberName)
+	_, err = c.MemberRemove(ctx, memberID)
+	require.NoError(t, err)
+	_, found, err = getMemberIdByName(ctx, c, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, false, "Expected member to be removed")
+	for member.IsRunning() {
+		err = member.Wait(ctx)
+		if err != nil && !strings.Contains(err.Error(), "unexpected exit code") {
+			t.Fatalf("member didn't exit as expected: %v", err)
+		}
+	}
+
+	t.Logf("Removing member %s data", memberName)
+	err = os.RemoveAll(member.Config().DataDirPath)
+	require.NoError(t, err)
+
+	t.Logf("Adding member %s back", memberName)
+	removedMemberPeerUrl := member.Config().PeerURL.String()
+	_, err = cc.MemberAdd(ctx, memberName, []string{removedMemberPeerUrl})
+	require.NoError(t, err)
+	err = patchArgs(member.Config().Args, "initial-cluster-state", "existing")
+	require.NoError(t, err)
+
+	t.Logf("Starting member %s", memberName)
+	err = member.Start(ctx)
+	require.NoError(t, err)
+	testutils.ExecuteUntil(ctx, t, func() {
+		for {
+			_, found, err := getMemberIdByName(ctx, c, memberName)
+			if err != nil || !found {
+				time.Sleep(10 * time.Millisecond)
+				continue
+			}
+			break
+		}
+	})
+}
--- a/tests/e2e/corrupt_test.go
+++ b/tests/e2e/corrupt_test.go
@@ -218,15 +218,10 @@ func TestPeriodicCheckDetectsCorruption(t *testing.T) {
 		assert.NoError(t, err, "error on put")
 	}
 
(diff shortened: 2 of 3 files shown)


serathius:
`grpcProxy-tests / test` failed. Looks like this test is incompatible with grpc-proxy. Recommend moving the test to separate file `ctl_v3_member_no_proxy_test.go` and adding  `//go:build !cluster_proxy` build tag at the top.

serathius:
`Static Analysis / run` failed. Please run `make fix-lint`

tests/e2e/ctl_v3_member_test.go
```diff
@@ -309,3 +312,104 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+	time.Sleep(etcdserver.HealthInterval)
+	_, err = c.MemberRemove(ctx, memberID)
+	assert.NoError(t, err)
+	_, found, err = getID(ctx, c, member.Config().Name)
+	assert.NoError(t, err)
+	assert.Equal(t, found, false, "Expected member to be removed")
+
+	t.Logf("Removing member %s", member.Config().Name)
+	for member.IsRunning() {
+		member.Kill()
```
serathius, line 348: Double checked the Kill here is not needed (mistake in the original test), the removed member should exit by themselves. You can just replace the whole block with `member.Wait(ctx)` to confirm that member exits.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -309,3 +312,104 @@ func TestRemoveNonExistingMember(t *testing.T) {
 	err = epc.Restart(ctx)
 	assert.NoError(t, err)
 }
+
+func TestMemberReplace(t *testing.T) {
+    e2e.BeforeTest(t)
+    ctx := context.Background()
```
serathius, line 318: Please provide some timeout for the context.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -309,3 +312,104 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+			t.Fail()
+		default:
+		}
+		reqCtx, cancel := context.WithTimeout(ctx, time.Second)
+		_, err = cc.MemberAdd(reqCtx, member.Config().Name, []string{removedMemberPeerUrl})
+		cancel()
+		if err == nil {
+			break
+		}
+	}
```
serathius, line 372: No need to retry here. Compared to robustness tests there is no other load that could cause request to fail.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -309,3 +312,104 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+	t.Logf("Adding member %s back", member.Config().Name)
+	removedMemberPeerUrl := member.Config().PeerURL.String()
+	for {
+		select {
+		case <-ctx.Done():
+			t.Fail()
+		default:
+		}
+		reqCtx, cancel := context.WithTimeout(ctx, time.Second)
+		_, err = cc.MemberAdd(reqCtx, member.Config().Name, []string{removedMemberPeerUrl})
```
serathius, line 367: Very interesting. I have used this test to validate https://github.com/etcd-io/etcd/pull/17056, looks like the issue is triggered by the immidiate `MemberAdd` request. When I removed the for loop and added a `time.Sleep(time.Second)` issue no longer reproduces.

This makes me think we have risk of potential race condition on membership changes that was is not covered by test at all. cc @ahrtr

serathius (approved):

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -0,0 +1,101 @@
...
+		default:
+		}
+		_, found, err := getID(ctx, c, memberName)
+		if err != nil {
+			continue
+		}
+		if found {
+			break
+		}
+	}
```
ahrtr, line 100:
```suggestion
	testutils.ExecuteUntil(ctx, t, func() {
	    for {
	         _, found, err := getMemberIdByName(ctx, c, memberName)
	         if err != nil || !found {
	             time.Sleep(10*time.Millisecond)
	             continue
	         }
	         break
	    }
	})
```

ahrtr:
The test is correct by itself. But I don't see how it can prevent https://github.com/etcd-io/etcd/pull/17017 from happening again.

Proposed actions:
- Merge this PR (after the comments are resolved);
- Create an e2e or integration test (or reuse any existing test) on top of https://github.com/etcd-io/etcd/pull/17017 to reproduce the issue https://github.com/etcd-io/etcd/issues/17052#issuecomment-1837518872.  Afterwards, I am happy to work with anyone to have a deep dive to figure out the root cause.

tests/e2e/ctl_v3_member_test.go
ZhouJianMS: Initially, I was using `assert.NoError()`, which allowed the test to continue executing even when errors occurred in the `MemberAdd` request. Upon logging the error message, I found that

>     ctl_v3_member_no_proxy_test.go:92: Adding member TestMemberReplace-test-2 back failed: match not found.  Set EXPECT_DEBUG for more info Errs: [unexpected exit code [1] after running [/home/zhoujian/etcd-test/etcd/bin/etcdctl --endpoints=http://localhost:20000,http://localhost:20005 member add TestMemberReplace-test-2 --peer-urls http://localhost:20011 -w json]], last lines:
>         {"level":"warn","ts":"2023-12-12T14:35:14.017056+0800","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0001c2780/localhost:20000","method":"/etcdserverpb.Cluster/MemberAdd","attempt":0,"error":"rpc error: code = FailedPrecondition desc = etcdserver: Peer URLs already exists"}
>         Error: etcdserver: Peer URLs already exists

In your case, a single `MemberAdd` request might fail, yet the test continues. Consequently, the cluster doesn't re-add the member, and it loops to locate the corresponding member ID until a timeout occurs. This divergence could arise from issue #17017, which might not get the updated `membersMap` timely to check peer urls.

ZhouJianMS:
> The test is correct by itself. But I don't see how it can prevent #17017 from happening again.
> 
> Proposed actions:
> 
> * Merge this PR (after the comments are resolved);
> * Create an e2e or integration test (or reuse any existing test) on top of [Switch to validating v3 when v2 and v3 are synchronized #17017](https://github.com/etcd-io/etcd/pull/17017) to reproduce the issue [Flaky robustness test #17052 (comment)](https://github.com/etcd-io/etcd/issues/17052#issuecomment-1837518872).  Afterwards, I am happy to work with anyone to have a deep dive to figure out the root cause.

Based on #17017, I have reproduced the issue with same test case. You can find it in draft PR #17100. But this kind of reproduce is quite unstable, over 10 times may only have 1 repro. But if removing loop for `MemberAdd` request as this PR does, the repro rate is quite high.  As I discussed here https://github.com/etcd-io/etcd/pull/17079#discussion_r1423631612, there may be some delay before `membersMap` get latest value from method `MustReadMembersFromBackend`. And it can lead to "failed to update; member unknown" as https://github.com/etcd-io/etcd/issues/17052#issuecomment-1837518872 said. 

This PR can prevent get the out-of-date value of `membersMap` for verification, thus preventing similar issue from happening again. If you have further ideas, I am happy to dig deeper for the root cause.

ahrtr (approved):
/lgtm

ahrtr:
> You can find it in draft PR #17100. But this kind of reproduce is quite unstable, over 10 times may only have 1 repro.

Thanks for the draft PR to reproduce the issue.  Could you rebase that PR? I will take a look later.

ahrtr:
> > You can find it in draft PR #17100. But this kind of reproduce is quite unstable, over 10 times may only have 1 repro.
> 
> Thanks for the draft PR to reproduce the issue. Could you rebase that PR? I will take a look later.

Please ping me after you finish rebasing that PR and can still reproduce the issue. Thanks.

ZhouJianMS:
> > > You can find it in draft PR #17100. But this kind of reproduce is quite unstable, over 10 times may only have 1 repro.
> > 
> > 
> > Thanks for the draft PR to reproduce the issue. Could you rebase that PR? I will take a look later.
> 
> Please ping me after you finish rebasing that PR and can still reproduce the issue. Thanks.

@ahrtr The branch `ZhouJianMS:repro17052` of PR #17100 is already based on the issue #17052. You can switch to that branch and run test `TestMemberReplaceMultiple` to reproduce the issue.

ahrtr:
> @ahrtr The branch `ZhouJianMS:repro17052` of PR #17100 is already based on the issue #17052. You can switch to that branch and run test `TestMemberReplaceMultiple` to reproduce the issue.

Thanks

ahrtr:
@ZhouJianMS could you backport the test cases to 3.5 and. 3.4 as well? I believe 3.5 and 3.4 have similar issue. I think we need to backport https://github.com/etcd-io/etcd/pull/17119 to 3.5 and 3.4 as well.

ZhouJianMS:
> @ZhouJianMS could you backport the test cases to 3.5 and. 3.4 as well? I believe 3.5 and 3.4 have similar issue. I think we need to backport #17119 to 3.5 and 3.4 as well.

Sure. Willing to do.

PR #17087, opened 2023-12-11
tests: add integration test for retry loop in keepAliveOnce

### LINKED ISSUE #16869: Add integration test for retry loop in lessor.KeepAliveOnce

[lessor.KeepAliveOnce ](https://github.com/etcd-io/etcd/blob/02794b00eb0e964943c91d283106072e18fdc092/client/v3/lease.go#L308)will retry on certain type of errors. For example ErrGRPCNoLeader.

This retry scenario isn't [covered](https://app.codecov.io/gh/etcd-io/etcd/blob/main/client%2Fv3%2Flease.go#L317) by tests.

It might be difficult to trigger retry. As a starting point check out TestKVPutWithRequireLeader for an example of NoLeader error.

### Why is this needed?

improve coverage

AvineshTripathi:
fixes #16869

The retry scenarion for lease keepAliveOnce function was missing integration test coverage. This commit checks for ErrNoLeader

### DIFF
--- a/tests/integration/clientv3/lease/lease_test.go
+++ b/tests/integration/clientv3/lease/lease_test.go
@@ -117,6 +117,21 @@ func TestLeaseKeepAliveOnce(t *testing.T) {
 	if err != rpctypes.ErrLeaseNotFound {
 		t.Errorf("expected %v, got %v", rpctypes.ErrLeaseNotFound, err)
 	}
+
+	clus.Members[1].Stop(t)
+	clus.Members[2].Stop(t)
+
+	// wait for election timeout, then member[0] will not have a leader.
+	var (
+		electionTicks = 10
+		tickDuration  = 10 * time.Millisecond
+	)
+	time.Sleep(time.Duration(3*electionTicks) * tickDuration)
+
+	_, err = lapi.KeepAliveOnce(context.Background(), resp.ID)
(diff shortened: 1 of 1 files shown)


tests/integration/clientv3/lease/lease_test.go
```diff
@@ -117,6 +117,21 @@ func TestLeaseKeepAliveOnce(t *testing.T) {
 	if err != rpctypes.ErrLeaseNotFound {
 		t.Errorf("expected %v, got %v", rpctypes.ErrLeaseNotFound, err)
 	}
+
+	clus.Members[1].Stop(t)
```
lavacat, line 121: nit: I think it's better to move this to a separate test.

AvineshTripathi: You mean another TestLeaseKeepAliveOnce func? Not sure if we shhould do that in integeration testing

siyuanfoundation:
@moficodes

tests/integration/clientv3/lease/lease_test.go
serathius: Copy the `TestLeaseKeepAliveOnce` test and rename it to `TestLeaseKeepAliveOnceNoLeader`

PR #17089, opened 2023-12-11
Adjusted RAM Requirements for arm64 Workflows

fykaa:
part of https://github.com/etcd-io/etcd/issues/17045
This PR updates the RAM requirements for several workflows based on the recent triage meeting discussions and analysis:

**E2E Tests:**

* Increased the RAM allocation for the actuated runner to 8GB.

**Reusable Tests Workflow:**

* Due to the workflow utilizing the `runs-on` input provided by the calling workflow, direct modification was not possible.
* Implemented a conditional statement to dynamically adjust the `runs-on` parameter based on the `arch` input variable:
    * For `arm64` architecture, the workflow now runs on `actuated-arm64-8cpu-8gb` with 8GB RAM.
    * Other architectures continue to use the RAM allocation specified by the calling workflow.

**Reusable Robustness Workflow:**

* The `runs-on` parameter is parsed at runtime using the `fromJson` function.
* We need to decide on the best approach to modify the RAM allocation:
  
    **Option 1: Update the Default Input:**
    
    ```yaml
    runs-on: "['ubuntu-latest', 'runs-on: actuated-arm64-8cpu-8gb']"
    ```
    
    * This pre-pends the `actuated-arm64-8cpu-8gb` runner to the default list, ensuring it's used for all workflow runs.
    * Simple and straightforward but applies to all architectures regardless of need.
    
    **Option 2: Modify Workflow Call Input:**
    
    ```yaml
    workflow_call:
      ...
      inputs:
        ...
        runs-on: actuated-arm64-8cpu-8gb
    ```

* This customizes the RAM allocation only for specific workflow calls.
* Provides flexibility but requires modification in each calling workflow.

Please provide feedback on which approach is preferred so I can make the appropriate changes.

### DIFF
--- a/.github/workflows/tests-template.yaml
+++ b/.github/workflows/tests-template.yaml
@@ -6,17 +6,16 @@ on:
       arch:
         required: true
         type: string
-      # runs-on:
-      #   required: true
-      #   type: string
+      runs-on:
+        required: true
+        type: string
 permissions: read-all
 
 jobs:
   test:
+    runs-on: ${{ inputs.runs-on }}
     # this is to prevent arm64 jobs from running at forked projects
-    if: inputs.arch == 'arm64' || github.repository == 'etcd-io/etcd'
-    # this is to tune memory allocation of arm64 architecture
-    runs-on: ${{ if inputs.arch == 'arm64' then 'actuated-arm64-8cpu-8gb' else inputs.runs-on end }}
+    if: inputs.arch == 'amd64' || github.repository == 'etcd-io/etcd'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/tests-template.yaml
+++ b/.github/workflows/tests-template.yaml
@@ -6,16 +6,17 @@ on:
       arch:
         required: true
         type: string
-      runs-on:
-        required: true
-        type: string
+      # runs-on:
+      #   required: true
+      #   type: string
 permissions: read-all
 
 jobs:
   test:
-    runs-on: ${{ inputs.runs-on }}
     # this is to prevent arm64 jobs from running at forked projects
-    if: inputs.arch == 'amd64' || github.repository == 'etcd-io/etcd'
+    if: inputs.arch == 'arm64' || github.repository == 'etcd-io/etcd'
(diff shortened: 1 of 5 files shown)


.github/workflows/tests-template.yaml
```diff
@@ -6,16 +6,17 @@ on:
...
 permissions: read-all
 
 jobs:
   test:
-    runs-on: ${{ inputs.runs-on }}
     # this is to prevent arm64 jobs from running at forked projects
-    if: inputs.arch == 'amd64' || github.repository == 'etcd-io/etcd'
+    if: inputs.arch == 'arm64' || github.repository == 'etcd-io/etcd'
+    # this is to tune memory allocation of arm64 architecture
+    runs-on: ${{ if inputs.arch == 'arm64' then 'actuated-arm64-8cpu-8gb' else inputs.runs-on end }}
```
jmhbnz, line 19: I don't believe this file needs to be touched by this pr.

My understanding is the files we need to update (in addition to the `e2e-arm64.yaml` completed above) are:

https://github.com/etcd-io/etcd/blob/a49a867960034f61da4165f6873b2557392a2939/.github/workflows/tests.yaml#L15

https://github.com/etcd-io/etcd/blob/a49a867960034f61da4165f6873b2557392a2939/.github/workflows/robustness.yaml#L21

https://github.com/etcd-io/etcd/blob/a49a867960034f61da4165f6873b2557392a2939/.github/workflows/robustness-nightly.yaml#L30

https://github.com/etcd-io/etcd/blob/a49a867960034f61da4165f6873b2557392a2939/.github/workflows/robustness-nightly.yaml#L46

The update should be simply replacing `32gb` with `8gb`.

jmhbnz (changes requested):
Hey @fykaa - Thanks for getting this started. Let's keep this pr simple by just updating the existing RAM values, see below.

jmhbnz (approved):
LGTM - I'm not seeing any degradation in workflow execution time with the reduce ram ceiling.  Thanks @fykaa.

ahrtr (approved):
/lgtm

Thanks both @jmhbnz and @fykaa

Issue #16687, opened 2023-10-03
`/members` peer handler is not guarded with linearizable check


chaochn47:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request, support requests should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Restart etcd server right after member reconfiguration and query the member list via HTTP /members, [its handler](https://github.com/etcd-io/etcd/blob/release-3.4/etcdserver/api/etcdhttp/peer.go#L107) will bypass the linearizable check. The member list response would be stale during bootstrap (restart) where the members are restored from v2 store and WAL is still replaying...

### What did you expect to happen?

When peer certificate is not used, `/members` should be protected by linearizable check. 

### How can we reproduce it (as minimally and precisely as possible)?

1. checkout code in https://github.com/chaochn47/etcd/tree/v3.4.20-eks.0-reproduce-member-name-mismatch
3. `cd integration && go test -v -run TestReproduceMemberNameMismatch`

### Anything else we need to know?

Relevant to discussion 

> > Restart etcd server right after member reconfiguration and query the member list via HTTP `/members`, [its handler](https://github.com/etcd-io/etcd/blob/release-3.4/etcdserver/api/etcdhttp/peer.go#L107) will bypass the linearizable check.
> 
> Actually the main also has this "issue", but this might not be a problem. Two reasons:
> 
>     * The `/members` is only supposed to accessed during bootstrap by a new etcd member who has just been added to the cluster;
> 
>     * The `/members` is protected by peer certificate; in other words, users shouldn't be able to access this endpoint at all.
> 
> 
> EDIT: but I still think it may be better to guard the `/members` by linearizable check. Please raise a separate issue to track it, and we can discuss it separately.

https://github.com/etcd-io/etcd/issues/16666#issuecomment-1744601073

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

[...]

PR #17100, opened 2023-12-12
Draft: Add member replace e2e test to repro issue 17052

ZhouJianMS:

### DIFF
--- a/tests/e2e/ctl_v3_member_no_proxy_test.go
+++ b/tests/e2e/ctl_v3_member_no_proxy_test.go
@@ -31,7 +31,28 @@ import (
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
 )
 
-func TestMemberReplace(t *testing.T) {
+func TestMemberReplaceMultiple(t *testing.T) {
+	// Loop for adding member is hard to repro
+	for i := 0; i < 10; i++ {
+		succeed := t.Run("LoopForAdd", func(t *testing.T) {
+			memberReplaceTest(t, true)
+		})
+		if !succeed {
+			t.Fatalf("LoopForAdd failed")
+		}
+	}
+	// Only call member add once is easy to repro
+	for i := 0; i < 10; i++ {
+		succeed := t.Run("OneStepAdd", func(t *testing.T) {
+			memberReplaceTest(t, false)
+		})
+		if !succeed {
+			t.Fatalf("LoopForAdd failed")
+		}
+	}
+}
+
+func memberReplaceTest(t *testing.T, loopForAdd bool) {
 	e2e.BeforeTest(t)
 	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
 	defer cancel()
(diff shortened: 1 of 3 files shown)


ahrtr:
I reproduced two issues based on this PR,
- "[member count is unequal](https://github.com/etcd-io/etcd/blob/dfdffe48f9e7c622f1b863d013754c2a824f6d35/server/etcdserver/api/membership/cluster.go#L741)".  It's caused by a known issue https://github.com/etcd-io/etcd/issues/16687. When starting the newly added member, it gets members from one of its peer, but the `/members` doesn't go through raft, so the peer might haven't finish applying the previous conf change.
  - Added more log message in https://github.com/etcd-io/etcd/pull/17113 to discover this issue.
- "etcdserver: Peer URLs already exists".  Still need investigation.

fuweid:
When I use `taskset` to use few cores during testing, the test run into the `etcdserver: Peer URLs already exists`. I think it might be related to txReadBuffer. That cache wasn't be updated. The `batchTxBuffer` only updates cache for `Put`. I think we can use `tx.Commit` during removing member.

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -0,0 +1,132 @@
...
+	c := epc.Etcdctl()
+	memberID, found, err := getMemberIdByName(ctx, c, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, true, "Member not found")
+
+	// Need to wait health interval for cluster to accept member changes
+	time.Sleep(etcdserver.HealthInterval)
+
+	t.Logf("Removing member %s", memberName)
+	_, err = c.MemberRemove(ctx, memberID)
```
fuweid, line 83: If the client selects that member which is going to be removed, it might return error because that server stops. 

client won't retry if the error is `code = Unavailable`

ZhouJianMS: I follow the guide in https://etcd.io/docs/v3.5/tutorials/how-to-deal-with-membership/ to remove member. Which saying is correct?

fuweid: I run into this issue in my local one time. https://github.com/etcd-io/etcd/blob/dfdffe48f9e7c622f1b863d013754c2a824f6d35/server/etcdserver/server.go#L1115

1 second might good enough to return response. never mind.

ahrtr: > If the client selects that member which is going to be removed, it might return error because that server stops.

Yes, actually there are two "issues" (one real issue, one improvement) in the test cases:
- [Issue] The client might connect to the member which will be removed, so the client might get an error response.
- [Improvement] We should use client sdk instead of etcdctl to communicate with etcdserver directly. Otherwise when the client failed, it can only see etcdctl 's log instead of etcdserver's log.

Actually I already updated the # 1 above.  Let me deliver a PR to fix it.

ahrtr:
> That cache wasn't be updated.

Exactly, thanks for the quick investigation. Let me fix it together with the PR mentioned in https://github.com/etcd-io/etcd/pull/17100#discussion_r1426677080

tests/e2e/ctl_v3_member_no_proxy_test.go
ahrtr: FYI. add "EXPECT_DEBUG=true" can get the detailed etcdserver's log, 

```
EXPECT_DEBUG=true go test -run TestMemberReplaceMultiple -v
```

ahrtr:
Resolved in https://github.com/etcd-io/etcd/pull/17119.

Thanks @ZhouJianMS and @fuweid

PR #17107, opened 2023-12-13
testutils: return early instead of first breaking in LogObserver.Expe…

redwrasse:
Return early instead of unnecessarily first breaking, and then returning under condition `len(res) >= count` in `LogObserver.ExpectFunc`, in `testutils` package. Control flow logic is (I think) unchanged.


serathius (approved):

jmhbnz (approved):
LGTM - Thanks for cleaning this up @redwrasse

fuweid (approved):
LGTM

ahrtr (approved):
lgtm

Thanks

PR #17109, opened 2023-12-13
Fix memory leak caused by for + time.After

silves-xiang:


client/v3/lease.go
```diff
@@ -549,9 +549,12 @@ func (l *lessor) recvKeepAlive(resp *pb.LeaseKeepAliveResponse) {
 // deadlineLoop reaps any keep alive channels that have not received a response
 // within the lease TTL
 func (l *lessor) deadlineLoop() {
+	timer := time.NewTimer(time.Second)
```
serathius, line 552: Note about using `time.After` from function docstring
```
// It is equivalent to NewTimer(d).C.
// The underlying Timer is not recovered by the garbage collector
// until the timer fires. If efficiency is a concern, use NewTimer
// instead and call Timer.Stop if the timer is no longer needed.
```
So this PR allows `Timer` to be garbage collected when function exits, instead of after 1 second.

serathius:
Please sign the DCO https://github.com/etcd-io/etcd/pull/17109/checks?check_run_id=19589099104

silves-xiang:
> Please sign the DCO https://github.com/etcd-io/etcd/pull/17109/checks?check_run_id=19589099104

i recommit this modify, please check it , thank you very much.

PR #17112, opened 2023-12-13
[3.4] Backport disable redirects in peer communication

ivanvc:
Disable following redirects from peer HTTP communication on the client's side. Etcd server may run into SSRF (Server-side request forgery) when adding a new member. If users provide a malicious peer URL, the existing etcd members may be redirected to another unexpected internal URL when getting the new member's version.

Related to #16969 

### DIFF
--- a/etcdserver/cluster_util.go
+++ b/etcdserver/cluster_util.go
@@ -66,6 +66,9 @@ func getClusterFromRemotePeers(lg *zap.Logger, urls []string, timeout time.Durat
 	cc := &http.Client{
 		Transport: rt,
 		Timeout:   timeout,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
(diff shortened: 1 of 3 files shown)


jmhbnz (approved):
LGTM - Thanks @ivanvc

ahrtr (approved):
lgtm

Could you add a chanagelog item for 3.4? Thanks

ivanvc:
> lgtm
> 
> Could you add a chanagelog item for 3.4? Thanks

Should I do the same for 3.5?

jmhbnz:
> Should I do the same for 3.5?

There is an existing 3.5 changelog https://github.com/etcd-io/etcd/commit/443af2497dd46aaec6b4832b11fe441ab7bdd376.  I do note we only called out the member addition as that was the perceived security risk. Let's add something similar for 3.4, can maybe make the message a bit broader.

PR #17117, opened 2023-12-14
etcdclient: Fix memory leak caused by for + time.After

silves-xiang:

### DIFF
--- a/client/v3/lease.go
+++ b/client/v3/lease.go
@@ -549,9 +549,12 @@ func (l *lessor) recvKeepAlive(resp *pb.LeaseKeepAliveResponse) {
 // deadlineLoop reaps any keep alive channels that have not received a response
 // within the lease TTL
 func (l *lessor) deadlineLoop() {
+	timer := time.NewTimer(time.Second)
+	defer timer.Stop()
 	for {
+		timer.Reset(time.Second)
 		select {
-		case <-time.After(time.Second):
+		case <-timer.C:
 		case <-l.donec:
 			return
 		}


client/v3/lease.go
```diff
@@ -549,9 +549,12 @@ func (l *lessor) recvKeepAlive(resp *pb.LeaseKeepAliveResponse) {
 // deadlineLoop reaps any keep alive channels that have not received a response
 // within the lease TTL
 func (l *lessor) deadlineLoop() {
+	timer := time.NewTimer(time.Second)
```
serathius, line 552: Copying comment from https://github.com/etcd-io/etcd/pull/17109

Note about using `time.After` from function docstring
```
// It is equivalent to NewTimer(d).C.
// The underlying Timer is not recovered by the garbage collector
// until the timer fires. If efficiency is a concern, use NewTimer
// instead and call Timer.Stop if the timer is no longer needed.
```
This PR allows `Timer` to be garbage collected immediately when function exits, instead of after 1 second. 

So this is less of a memory leak fix, and more of a proper way to use time package. Still, it's an improvement that we can merge.

serathius:
cc @ahrtr

fuweid (approved):
LGTM

client/v3/lease.go
ahrtr: nit: ticker is even better. At least, no need to call reset repeatedly.  Not a big deal though.

https://pkg.go.dev/time#example-NewTicker 

```
package main

import (
	"fmt"
	"time"
)

func main() {
	ticker := time.NewTicker(time.Second)
	defer ticker.Stop()
	done := make(chan bool)
	go func() {
		time.Sleep(10 * time.Second)
		done <- true
	}()
	for {
		select {
		case <-done:
			fmt.Println("Done!")
			return
		case t := <-ticker.C:
			fmt.Println("Current time: ", t)
		}
	}
}

```

ahrtr (approved):

serathius:
> nit: ticker is even better. At least, no need to call reset repeatedly. Not a big deal though.

Yes, but technically it's different behavior. Timer always waits 1 second no matter the execution time, while ticker waits up to 1 second. Difference is when loop execution time is substantial.

PR #17119, opened 2023-12-14
Commit transaction for each configuration change

ahrtr:
Current when etcd remove a member, it just removes it from bbolt db, but doesn't update the cache. On the other hands, etcd periodically commit each bbolt transaction. When etcd processes the next conf change request, it might not have finished committing the previous member remove request into bbolt. Accordingly it breaks the linerizability.

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1809,6 +1809,9 @@ func (s *EtcdServer) apply(
 			var cc raftpb.ConfChange
 			pbutil.MustUnmarshal(&cc, e.Data)
 			removedSelf, err := s.applyConfChange(cc, confState, shouldApplyV3)
+			if err != nil {
+				s.lg.Error("failed to apply conf change", zap.Bool("shouldApplyV3", bool(shouldApplyV3)), zap.Error(err))
+			}
 			s.setAppliedIndex(e.Index)
 			s.setTerm(e.Term)
 			shouldStop = shouldStop || removedSelf
--- a/server/storage/schema/membership.go
+++ b/server/storage/schema/membership.go
@@ -72,9 +72,28 @@ func (s *membershipBackend) MustDeleteMemberFromBackend(id types.ID) {
 
 	tx := s.be.BatchTx()
 	tx.LockInsideApply()
-	defer tx.Unlock()
 	tx.UnsafeDelete(Members, mkey)
 	tx.UnsafePut(MembersRemoved, mkey, []byte("removed"))
+	tx.Unlock()
+
+	// We need to forcibly commit the transaction, otherwise etcd might
+	// run into a situation that it haven't finished committing the data
+	// into backend storage (note: etcd periodically commits the bbolt
+	// transactions instead of on each request) when it receives next
+	// confChange request. Accordingly, etcd may still reads the stale
+	// data from bbolt when processing next confChange request. So it
+	// breaks linearizability.
+	//
+	// Note we don't need to forcibly commit the transaction for other
+	// kinds of request (e.g. normal key/value operations, member add
+	// or update requests), because there is a buffer on top of the bbolt.
+	// Each time when etcd reads data from backend storage, it will read
+	// data from both bbolt and the buffer. But there is no such a buffer
+	// for member delete requests.
+	//
+	// Please also refer to
+	// https://github.com/etcd-io/etcd/pull/17119#issuecomment-1857547158
+	s.be.ForceCommit()
 }
 
 func (s *membershipBackend) MustReadMembersFromBackend() (map[types.ID]*membership.Member, map[types.ID]bool) {
--- a/server/etcdserver/server_test.go
+++ b/server/etcdserver/server_test.go
@@ -298,6 +298,7 @@ func newServer(t *testing.T, recorder *nodeRecorder) *EtcdServer {
 		r:            *newRaftNode(raftNodeConfig{lg: lg, Node: recorder}),
 		cluster:      membership.NewCluster(lg),
 		consistIndex: cindex.NewConsistentIndex(be),
+		kv:           mvcc.New(zap.NewNop(), be, &lease.FakeLessor{}, mvcc.StoreConfig{}),
 	}
 	srv.cluster.SetBackend(schema.NewMembershipBackend(lg, be))
 	srv.cluster.SetStore(v2store.New())
@@ -475,6 +476,7 @@ func TestApplyConfigChangeUpdatesConsistIndex(t *testing.T) {
 		w:            wait.New(),
 		consistIndex: ci,
 		beHooks:      serverstorage.NewBackendHooks(lg, ci),
+		kv:           mvcc.New(zap.NewNop(), be, &lease.FakeLessor{}, mvcc.StoreConfig{}),
 	}
 
 	// create EntryConfChange entry
@@ -567,6 +569,7 @@ func TestApplyMultiConfChangeShouldStop(t *testing.T) {
 		w:            wait.New(),
 		consistIndex: ci,
 		beHooks:      serverstorage.NewBackendHooks(lg, ci),
+		kv:           mvcc.New(zap.NewNop(), be, &lease.FakeLessor{}, mvcc.StoreConfig{}),
 	}
 	var ents []raftpb.Entry
 	for i := 1; i <= 4; i++ {
@@ -876,6 +879,7 @@ func TestAddMember(t *testing.T) {
 		SyncTicker:   &time.Ticker{},
 		consistIndex: cindex.NewFakeConsistentIndex(0),
 		beHooks:      serverstorage.NewBackendHooks(lg, nil),
+		kv:           mvcc.New(zap.NewNop(), be, &lease.FakeLessor{}, mvcc.StoreConfig{}),
 	}
 	s.start()
 	m := membership.Member{ID: 1234, RaftAttributes: membership.RaftAttributes{PeerURLs: []string{"foo"}}}
@@ -981,6 +985,7 @@ func TestRemoveMember(t *testing.T) {
 		SyncTicker:   &time.Ticker{},
 		consistIndex: cindex.NewFakeConsistentIndex(0),
 		beHooks:      serverstorage.NewBackendHooks(lg, nil),
+		kv:           mvcc.New(zap.NewNop(), be, &lease.FakeLessor{}, mvcc.StoreConfig{}),
 	}
 	s.start()
 	_, err := s.RemoveMember(context.Background(), 1234)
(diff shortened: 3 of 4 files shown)


ahrtr:
Confirmed that this PR can fix https://github.com/etcd-io/etcd/pull/17100 and also https://github.com/etcd-io/etcd/issues/17052 

@serathius  Please feel free to resurrect https://github.com/etcd-io/etcd/pull/17017. From what we understood so far, https://github.com/etcd-io/etcd/pull/17017 does not introduce regression. Instead, it just discovered an existing issue.

ahrtr:
The workflow `pull-etcd-e2e-amd64` actually passed, but it is displayed as failed due to "Pod got deleted unexpectedly". It might be an issue of prow?  @jmhbnz

serathius:
Little confused about the naming and the proposed solution. Call  to `s.KV().Commit()` when applying membership change seems weirdly asymmetrical to normal change. Please elaborate why this needs to be flushed to disk.

> Current when etcd remove a member, it just removes it from bbolt db, but doesn't update the cache. On the other hands, etcd periodically commit each bbolt transaction. When etcd processes the next conf change request, it might not have finished committing the previous member remove request into bbolt. Accordingly it breaks the linerizability.

As for naming, what do you mean by cache in `when etcd remove a member, it just removes it from bbolt db, but doesn't update the cache`? Do you mean `RaftCluster` fields? In https://github.com/etcd-io/etcd/blob/2cf112f3b98021db0d017e5392c0ad63812d4894/server/etcdserver/api/membership/cluster.go#L421-L462 things seem ok.

 > When etcd processes the next conf change request, it might not have finished committing the previous member remove request into bbolt. Accordingly it breaks the linerizability.
 
 Shouldn't it be ok as long consistent index in bbolt is consistent with the conf change? When I recently surprised by the differences of the apply code for updating consistent index between apply normal https://github.com/etcd-io/etcd/blob/2cf112f3b98021db0d017e5392c0ad63812d4894/server/etcdserver/server.go#L1831-L1841 and apply config change https://github.com/etcd-io/etcd/blob/2cf112f3b98021db0d017e5392c0ad63812d4894/server/etcdserver/server.go#L1980-L1993
 
 I will need some time to deep dive and understand your finding. It might take me some time. Thanks for your understanding.

jmhbnz:
> The workflow pull-etcd-e2e-amd64 actually passed, but it is displayed as failed due to "Pod got deleted unexpectedly". It might be an issue of prow? @jmhbnz

I'm not certain why the pod was deleted. It may have been killed for a resource or preemption reason on the kubernetes infra cluster.

Below are the final logs for the prowjob:

```bash
 {"Time":"2023-12-14T15:42:36.887712595Z","Action":"output","Package":"go.etcd.io/etcd/tests/v3/common","Test":"TestUserAdd_Simple/PeerAutoTLS/no_password_without_noPassword_set","Output":"/home/prow/go/src/github.com/etcd-{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:173","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2023-12-14T15:42:38Z"}
++ early_exit_handler
++ '[' -n 17 ']'
++ kill -TERM 17
++ cleanup_dind
++ [[ false == \t\r\u\e ]]
+ EXIT_VALUE=143
+ set +o xtrace
make: *** [Makefile:53: test-e2e-release] Terminated 
```

For now I will retry it.  If we see this error again I will raise an issue in `k/test-infra`.

/retest

server/etcdserver/server.go
```diff
@@ -1809,6 +1809,11 @@ func (s *EtcdServer) apply(
 			var cc raftpb.ConfChange
 			pbutil.MustUnmarshal(&cc, e.Data)
 			removedSelf, err := s.applyConfChange(cc, confState, shouldApplyV3)
+			if err != nil {
+				s.lg.Error("failed to apply conf change", zap.Bool("shouldApplyV3", bool(shouldApplyV3)), zap.Error(err))
+			} else {
+				s.KV().Commit()
```
fuweid, line 1815: is it possible to call `Commit` in `applyConfChange`? 

And I think it's better to add comment on `MustSaveMemberToBackend`, `TrimClusterFromBackend` and `MustDeleteMemberFromBackend` and says that please ensure it should be synced with tx buffer. 

It can save time during debugging test cases 😂

ahrtr: Yes, makes sense. Suggestion accepted. Updated to get the `ForceCommit` included in `MustDeleteMemberFromBackend` so as to avoid introducing similar issue by other contributors.

server/storage/schema/membership.go
```diff
@@ -72,9 +72,25 @@ func (s *membershipBackend) MustDeleteMemberFromBackend(id types.ID) {
...
+	// run into a situation that it haven't finished committing the data
+	// into backend storage (note: etcd periodically commits the bbolt
+	// transactions instead of on each request) when it receives next
+	// confChange request. Accordingly, etcd may still reads the stale
+	// data from bbolt when processing next confChange request. So it
+	// breaks linearizability.
+	//
+	// Note we don't need to forcibly commit the transaction for other
+	// kinds of request (e.g. normal key/value operations, member add
+	// or update requests), because there is a buffer on top of the bbolt.
```
serathius, line 89: Can you be more specific about the buffer. Do you mean `batchTxBuffered`?

ahrtr:
Just as I mentioned in https://github.com/etcd-io/etcd/pull/17100#issuecomment-1854560938, there are two issues. Let me provide more context & details.

## `/members` requests do not go through raft
When starting the newly added member, it gets members from one of its peer, but the `/members` doesn't go through raft, so the peer might haven't finish applying the previous conf change. So the new member may run into issue "[member count is unequal](https://github.com/etcd-io/etcd/blob/dfdffe48f9e7c622f1b863d013754c2a824f6d35/server/etcdserver/api/membership/cluster.go#L741)".  It's already tracked https://github.com/etcd-io/etcd/issues/16687.

Note this isn't a critical issue, not even a major issue,
- Usually Conf changes (i.e. add/remove members) aren't frequent operations in production environment;
- It is low possibility to see this issue in production.
- In worst case, when it happens, just restarting the new member can resolve it.
- From operators' perspectively, it also easy to avoid this issue,  i.e. wait for the conf change be synced to all existing members before starting a new added member.

In our test case, I added a 100ms sleep before starting the new added member to bypass this issue.

## Issue "Peer URLs already exists"
Just copied my comment inside the method `MustDeleteMemberFromBackend` to explain the root cause,

	// We need to forcibly commit the transaction, otherwise etcd might
	// run into a situation that it haven't finished committing the data
	// into backend storage (note: etcd periodically commits the bbolt
	// transactions instead of on each request) when it receives next
	// confChange request. Accordingly, etcd may still reads the stale
	// data from bbolt when processing next confChange request. So it
	// breaks linearizability.
	//
	// Note we don't need to forcibly commit the transaction for other
	// kinds of request (e.g. normal key/value operations, member add
	// or update requests), because there is a buffer on top of the bbolt.
	// Each time when etcd reads data from backend storage, it will read
	// data from both bbolt and the buffer. But there is no such a buffer
	// for member delete requests.
	s.be.ForceCommit()

More context from code perspective, the buffer mentioned above are

write buffer
https://github.com/etcd-io/etcd/blob/2cf112f3b98021db0d017e5392c0ad63812d4894/server/storage/backend/batch_tx.go#L291

read buffer
https://github.com/etcd-io/etcd/blob/2cf112f3b98021db0d017e5392c0ad63812d4894/server/storage/backend/read_tx.go#L43

syncs data from write buffer to read buffer at the end of the etcd TXN (Note: it isn't bbolt TXN, FYI https://drive.google.com/file/d/1GAVg206rqrL4at14Dpr837SrtFCrJ-hJ/view?usp=drive_link)
https://github.com/etcd-io/etcd/blob/2cf112f3b98021db0d017e5392c0ad63812d4894/server/storage/backend/batch_tx.go#L310

## about Consistent index
Note that the issues mentioned above have nothing to do with consistent-index. If anyone has any concern or suggestion on consistent-index, Please let's discuss it in a separate session.

server/storage/schema/membership.go
ahrtr: pls read https://github.com/etcd-io/etcd/pull/17119#issuecomment-1857547158

serathius:
Thanks for explanation, for me this is more scary than just "Peer URLs already exists". For me this is etcd transaction buffers breaking "readCommited" guarantee when deleting keys. Impact is limited because etcd doesn't do a lot of key deletion, as KV adds a tombstone instead of deletion. However the problem is that it breaks basic assumptions that can pop out in other places. 

I have added a simple test to reproduce it https://github.com/etcd-io/etcd/pull/17124
I think we should fix the underlying issue (readCommited broken on delete) and not a side effect "Peer URLs already exists".

ahrtr:
> I think we should fix the underlying issue (readCommited broken on delete) and not a side effect "Peer URLs already exists".

This PR fixes the underlying issue!

ahrtr:
Please feel free to let me know if you can reproduce whatever other issue on top of this PR.

ahrtr:
It turns out to be more generic issue https://github.com/etcd-io/etcd/pull/17124#issuecomment-1857694297

ahrtr:
Extracted the change for the test case TestMemberReplace into a separate PR https://github.com/etcd-io/etcd/pull/17125, so that it can be reviewed & merged separately.

fuweid (approved):
LGTM

server/storage/schema/membership.go
```diff
@@ -72,9 +72,28 @@ func (s *membershipBackend) MustDeleteMemberFromBackend(id types.ID) {
...
+	// Each time when etcd reads data from backend storage, it will read
+	// data from both bbolt and the buffer. But there is no such a buffer
+	// for member delete requests.
+	//
+	// Please also refer to
+	// https://github.com/etcd-io/etcd/pull/17119#issuecomment-1857547158
+	s.be.ForceCommit()
```
jmhbnz, line 96: Should we be applying the same fix to `MustSaveMemberToBackend`?

ahrtr: No, actually the comment right above `s.be.ForceCommit()` already explains the reason. Please feel free to ping me if you need further clarification. 

Also note this PR only resolves the issue for conf change. We need to think about a more generic solution for all similar changes on bucket other than `key`, i.e auth, lease, etc.

jmhbnz (approved):
LGTM - Good find @ahrtr and thanks for the detailed explanation.

ahrtr:

PR #17123, opened 2023-12-15
[3.5] member replace e2e test

ZhouJianMS:
Follow up: https://github.com/etcd-io/etcd/pull/17079#issuecomment-1856089798

### DIFF
--- a/tests/e2e/ctl_v3_member_no_proxy_test.go
+++ b/tests/e2e/ctl_v3_member_no_proxy_test.go
@@ -1,0 +1,98 @@
+// Copyright 2023 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"context"
+	"math/rand"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/server/v3/etcdserver"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestMemberReplace(t *testing.T) {
+	e2e.BeforeTest(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+
+	epc, err := e2e.NewEtcdProcessCluster(t, &e2e.EtcdProcessClusterConfig{
+		ClusterSize:      3,
+		KeepDataDir:      true,
+		CorruptCheckTime: time.Second,
+	})
+	require.NoError(t, err)
+	defer epc.Close()
+
+	memberIdx := rand.Int() % len(epc.Procs)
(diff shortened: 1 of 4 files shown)


ahrtr:
can you please also backport the **change for test case TestMemberReplace** in https://github.com/etcd-io/etcd/pull/17119?

ZhouJianMS:
> can you please also backport the **change for test case TestMemberReplace** in #17119?

Sure, done.

serathius (approved):

ahrtr:
@ZhouJianMS Reran the e2e test 4 times, always failed. Can you please download the log and take a look what's the issue?

Screenshot 2023-12-15 at 16 10 21

ZhouJianMS:
> @ZhouJianMS Reran the e2e test 4 times, always failed. Can you please download the log and take a look what's the issue?
> 
> Screenshot 2023-12-15 at 16 10 21

The test is stuck in waiting member killed after executing member remove command. @serathius Does member in etcd 3.5 automatically kill themselves or not?

After adding `member.Kill()`, the member can be successfully stopped. But I noticed that adding member back failed due to error "the member has been permanently removed from the cluster". Investigating why re-added member got the same ID.

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -0,0 +1,97 @@
...
+	time.Sleep(etcdserver.HealthInterval)
+
+	t.Logf("Removing member %s", memberName)
+	_, err = cc.MemberRemove(memberID)
+	require.NoError(t, err)
+	_, found, err = getMemberIdByName(ctx, cc, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, false, "Expected member to be removed")
+	for member.IsRunning() {
+		time.Sleep(10 * time.Millisecond)
```
fuweid, line 68: For 3.5 release, there is no background goroutine as zombie process's reaper.
I think we can just call Stop here. Otherwise, zombie process is running and `member.IsRunning` is true until timeout.

https://github.com/etcd-io/etcd/blob/5b572f15162d9b61979fb5eed65e65b026917464/pkg/expect/expect.go#L88C8-L88C23

fuweid: We can also consider to backport https://github.com/etcd-io/etcd/pull/14672

tests/e2e/utils.go
```diff
@@ -114,3 +115,27 @@ func fillEtcdWithData(ctx context.Context, c *clientv3.Client, dbSize int) error
...
+func patchArgs(args []string, flag, newValue string) []string {
+	for i, arg := range args {
+		if strings.Contains(arg, flag) {
+			args[i] = fmt.Sprintf("--%s=%s", flag, newValue)
+			return args
+		}
+	}
+	args = append(args, fmt.Sprintf("--%s=%s", flag, newValue))
+	return args
+}
```
ahrtr, line 141: Why the implementation is different from the main branch?

https://github.com/etcd-io/etcd/blob/93530f6e0aeaebc0f00c98b8688579a0e576e71b/tests/e2e/utils.go#L139-L147

ZhouJianMS: 3.5 does not have "initial-cluster-state" as a default argument in e2e framework. So, we need to append it in `args`.

https://github.com/etcd-io/etcd/blob/93530f6e0aeaebc0f00c98b8688579a0e576e71b/tests/framework/e2e/cluster.go#L468-L473

First introduced in commit https://github.com/etcd-io/etcd/commit/6f63f4b676b775e7ce32bc9281d28c96f7165d9c

ahrtr: We should backport https://github.com/etcd-io/etcd/pull/16707 to 3.5 if possible.

Please add a comment to explain the difference for now.

ahrtr (approved):
lgtm

Thanks

PR #17124, opened 2023-12-15
Reproduce readTxn breaks read commited on delete

serathius:

### DIFF
--- a/server/storage/backend/batch_tx_test.go
+++ b/server/storage/backend/batch_tx_test.go
@@ -19,6 +19,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+
 	bolt "go.etcd.io/bbolt"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
@@ -204,4 +206,63 @@ func TestBatchTxBatchLimitCommit(t *testing.T) {
 		}
 		return nil
 	})
+}
+
+func TestRangeAfterDeleteMatch(t *testing.T) {
+	b, _ := betesting.NewTmpBackend(t, time.Hour, 10000)
+	defer betesting.Close(t, b)
+
+	tx := b.BatchTx()
+
+	tx.Lock()
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar"))
+	tx.Unlock()
+	tx.Commit()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0)
+
+	tx.Lock()
+	tx.UnsafeDelete(schema.Test, []byte("foo"))
+	tx.Unlock()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0)
+}
+
+func TestRangeAfterCommitMatch(t *testing.T) {
+	b, _ := betesting.NewTmpBackend(t, time.Hour, 10000)
+	defer betesting.Close(t, b)
+
+	tx := b.BatchTx()
+
+	tx.Lock()
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.Unlock()
+
+	tx.Lock()
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar"))
+	tx.Unlock()
+	tx.Commit()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), []byte("foo3"), 1)
+
+	tx.Lock()
+	tx.UnsafePut(schema.Test, []byte("foo2"), []byte("bar2"))
+	tx.Unlock()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), []byte("foo3"), 1)
+}
+
+func checkTxnResponseMatch(t *testing.T, tx backend.BatchTx, rtx backend.ReadTx, key, endKey []byte, limit int64) {
+	tx.Lock()
+	ks1, _ := tx.UnsafeRange(schema.Test, key, endKey, limit)
+	tx.Unlock()
+
+	rtx.RLock()
+	ks2, _ := rtx.UnsafeRange(schema.Test, key, endKey, limit)
+	rtx.RUnlock()
+
+	if diff := cmp.Diff(ks1, ks2); diff != "" {
+		t.Errorf("keys on read and batch transaction doesn't match, diff: %s", diff)
+	}
 }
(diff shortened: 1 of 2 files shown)


ahrtr:
Indeed almost all usage (e.g. auth, alarm, etc.) of `tx.UnsafeDelete` has similar issues. We have two choice,
- force commit for all such usage. It's easy but error prone (it may be easily broken by new contributors)
- add a buf for delete. One for all solution. But need super careful consideration.

ahrtr:
> Option 3: Mark that there was a delete operation and disable reading through buffer.

Sounds good. Rough idea is something like below,
1. mark that there was a deletion operation (add a bool variable).  Reset the flag when committing (either the periodical commit or forcibly commit) the bbolt TXN.
2. forcibly commit the bbolt txn when the flag is true, 
    - either at the end of current etcd TXN
    - or [`option 4`] at the begin of next etcd read TXN (**`lazy commit`**)

The lazy commit seems have a little better performance. But let's see what's the complexity in terms of implementation for each case. Readability overweights performance here. Better readability means easier to understand and lesser error prone.

serathius:
> The lazy commit seems have a little better performance. But let's see what's the complexity in terms of implementation for each case. Readability overweights performance here. Better readability means easier to understand and lesser error prone.

I would expect that marking that there was a deletion operation should be faster as it doesn't need to go to disk. The cost of reading from bbolt directly should be negligable assuming that we mark per bucket buffer as invalid and not a whole transactions. Having this per buffer should prevent performance degradation to `key` bucket which is most performance sensitive. I expect that the only  deletion in `key` bucket is done during compaction, which immediately does `ForceCommit` after.

serathius:
Oh, it's far worse than I expected. Looks like readTxn is totally broken. 

Logic that merges readTxn buffer and bbolt response for UnsafeRange doesn't make sense https://github.com/etcd-io/etcd/blob/51bd8bd4e81cee6711fac5e8361d11f6e6623b1f/server/storage/backend/read_tx.go#L78-L122

Added a second test.

server/storage/backend/batch_tx_test.go
```diff
@@ -205,3 +207,62 @@ func TestBatchTxBatchLimitCommit(t *testing.T) {
...
+	tx.Lock()
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar"))
+	tx.Unlock()
+	tx.Commit()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0)
+
+	tx.Lock()
+	tx.UnsafeDelete(schema.Test, []byte("foo"))
```
ahrtr, line 226: This is incorrect usage. `tx.UnsafeDelete` shouldn't be called directly to remove keys from the `key` bucket. Instead, etcd still puts a key/value with tombstone. The only exception is compaction.

ahrtr: Just ignore this comment, it makes sense to call UnsafeDelete on other buckets.

server/storage/backend/batch_tx_test.go
```diff
@@ -205,3 +207,62 @@ func TestBatchTxBatchLimitCommit(t *testing.T) {
...
+	tx.Unlock()
+	tx.Commit()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), []byte("foo3"), 1)
+
+	tx.Lock()
+	tx.UnsafePut(schema.Test, []byte("foo2"), []byte("bar2"))
+	tx.Unlock()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), []byte("foo3"), 1)
```
ahrtr, line 253: The main problem is it doesn't guarantee the order.  
- `batchTx.UnsafeRange` reads the bbolt directly, it's the correct results. 
- While `readTx.UnsafeRange` reads the buffer firstly, and then read from bbolt if the results returned from buffer doesn't meet the limit. Otherwise, if the data returned from buffer meets the limit, it will not continue to read from bbolt. But the results returned from buffer may not be correctly ordered against the data in bbolt.

In this case, if you inserts foo2 firstly, then foo, then the case will pass. 

My immediate thought on the correct solution should be:
- if `endKey == nil` or `limit == math.MaxInt64`, the existing logic is correct.
- Otherwise, we should read at most `limit` records from both buffer and bbolt respectively, merge the results and sort, finally return the correct `limit` number of records.

serathius: Right, what you proposed makes sense. What worries me is that this was totally unknown behavior which means huge gaps in our testing.

ahrtr: It might not be an issue. 

We do not have to guarantee the order here. If the client cares about the order of the result, then etcd will set the limit as 0, which means no limit.

https://github.com/etcd-io/etcd/blob/4e986363a371ad88479cb8f9b248d16ea83abc87/server/etcdserver/txn/txn.go#L153-L159

There are two special internal use cases. The range [key, endKey] must be only in bbolt for both cases. 
The first one is compaction.
https://github.com/etcd-io/etcd/blob/4e986363a371ad88479cb8f9b248d16ea83abc87/server/storage/mvcc/kvstore_compaction.go#L53

The second use case is recovering,
https://github.com/etcd-io/etcd/blob/4e986363a371ad88479cb8f9b248d16ea83abc87/server/storage/mvcc/kvstore.go#L341

siyuanfoundation: I don't think this is an issue for clients. `tx.UnsafeRange` with range limit is mainly used internally. For the client Range requests, it maps keys to revs through the keyIndex first
```
revpairs, total := tr.s.kvindex.Revisions(key, end, rev, int(ro.Limit))
```
then each rev key is queried individually 
```
for i, revpair := range revpairs[:len(kvs)] {
...
		revBytes = RevToBytes(revpair, revBytes)
                // UnsafeRange is called with nil endkey.
		_, vs := tr.tx.UnsafeRange(schema.Key, revBytes, nil, 0)
...
```
So the client range results are always ordered due to the btree in kvindex.

siyuanfoundation:
> Option 3: Mark that there was a delete operation and disable reading through buffer.

I am interested in working on this.

server/storage/backend/batch_tx_test.go
serathius: Just to make it clear, I agree it might not be an issue, however based on other discovery (deletion not being passed to read cache) I think we should double check. I think it's an unobvious behavior that:
* should be properly documented
* we should double check if there is any internal misuse
* we could consider reflecting in in function names.

ahrtr: Please raise an issue to track it.

ahrtr:
> > Option 3: Mark that there was a delete operation and disable reading through buffer.
> 
> I am interested in working on this.

Please feel free to work on https://github.com/etcd-io/etcd/issues/17158. Thanks. 

BTW: suggest to finish all the work for the livez/readyz firstly.

ahrtr:

PR #17125, opened 2023-12-15
Update test case TestMemberReplace to always connect to stable endpoints

ahrtr:


serathius (approved):
Note, sleep & refactors you proposed would be also useful to robustness test memberReplaceFailpoint. 

For such administrator operations like MemberReplace (documented as recovery when a corrupted member needs to be replaced) we should have matching e2e and robustness tests. Would be great to figure out how we could reuse the code.

ahrtr:
> Would be great to figure out how we could reuse the code.

could you raise an issue to track it so that other contributors can help?

PR #17127, opened 2023-12-15
Add a verification for inconsistent responses from readbuffer

serathius:
Should help us find all places impacted by readbuf inconsistency.

Goal is to collect all e2e tests that are triggered by this verification:
* `TestCtlV3Elect` triggers on `key` bucket


serathius:
Hmm, it might not be a good way to write such verification. There is no simple way to compare results from queries with and without read buffer enabled.

serathius:
Ok, I think this was a failed approach to do validation. Instead plugin into later between readBuf and bbolt, we should plug in before readBuf. Like wrap backend.ReadTx() and backend.BatchTx() and intercept UnsafeRead, UnsafePut calls. It would call real code and compare it with it's own internal state simulating real backend. Basically implement a model based testing but as a assertion.

PR #17131, opened 2023-12-17
Remove workflow telemetry github action

jmhbnz:
Recently in https://github.com/etcd-io/etcd/issues/17045 we introduced a new github action to etcd testing workflows that run on `arm64` so that we could tune the memory requests of those workflows in an evidence based way.

We have now completed that tuning, with workflows being reduced from `32GB` ram to `8GB` ram which also includes plenty of headroom based on current utilisation where workflows generally all sit below `5GB`:

image

With the tuning complete I propose we turn off collecting the detailed performance telemetry for every workflow as it creates quite noisy summaries and some error annotations on our workflows, for example: https://github.com/etcd-io/etcd/actions/runs/7231109865.   

We can always turn it back on sometime in future if performance profiling is required again.

### DIFF
--- a/.github/workflows/e2e-arm64.yaml
+++ b/.github/workflows/e2e-arm64.yaml
@@ -13,8 +13,6 @@ jobs:
         target:
           - linux-arm64-e2e
     steps:
-      - name: Collect Workflow Telemetry
-        uses: catchpoint/workflow-telemetry-action@v1
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - id: goversion
         run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/robustness-template.yaml
+++ b/.github/workflows/robustness-template.yaml
@@ -27,8 +27,6 @@ jobs:
     timeout-minutes: 210
     runs-on: ${{ fromJson(inputs.runs-on) }}
     steps:
-      - name: Collect Workflow Telemetry
-        uses: catchpoint/workflow-telemetry-action@v1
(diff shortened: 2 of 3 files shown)


ahrtr (approved):

.github/workflows/e2e-arm64.yaml
```diff
@@ -13,8 +13,6 @@ jobs:
         target:
           - linux-arm64-e2e
     steps:
-      - name: Collect Workflow Telemetry
-        uses: catchpoint/workflow-telemetry-action@v1
```
ivanvc, line 17: @jmhbnz, would it make sense to change it to run just when debug mode is active? i.e.,

```suggestion
      - name: Collect Workflow Telemetry
        uses: catchpoint/workflow-telemetry-action@v1
        if: runner.debug == '1'
```

Or is it not worth the effort/irrelevant?

fuweid (approved):
LGTM

.github/workflows/e2e-arm64.yaml
jmhbnz: This was an interesting feature to explore, I hadn't used it before.  I think for now we stick with removing them because it seems like the only way to trigger debug is to set `ACTIONS_STEP_DEBUG` to `true`  which sounds like a maintainer/org admin needs to be involved to set a variable or secret for the repo. 

If we could toggle the debug via k8s ci robot command I could see debug mode being quite handy but I don't think it can currently, we would need to write a new plugin.

ivanvc: You can enable it with the debug workflow/job button, still it does need to be triggered by a maintainer as you pointed it out

ivanvc: You can enable it with the debug workflow/job button, still it does need to be triggered by a maintainer as you pointed out.

Issue #17146, opened 2023-12-19
Etcd v3.4 broke `snapshot.Metadata.Index < db.consistentIndex` invariant


serathius:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

We had an overloaded etcd running in cluster. It started failing liveness probes and was killed. When it was restarted it immediately failed with `failed to recover v3 backend from snapshot` and started crashlooping.

```
WARNING 2023-12-16T11:46:09.613Z failed to find [SNAPSHOT-INDEX].snap.db {"error":"snap: snapshot file doesn't exist","snapshot-file-path":"/var/etcd/data/member/snap/0000000050e5a1a3.snap.db","snapshot-index":1357226403}
DEFAULT 2023-12-16T11:46:09.619706767Z panic: failed to recover v3 backend from snapshot
```

Warning `failed to find [SNAPSHOT-INDEX].snap.db` comes from https://github.com/etcd-io/etcd/blob/f7be2dfa17a04bd1eb2ffc02489b7518e7eb68ad/server/etcdserver/api/snap/db.go#L76-L95

It is executed when `snapshot.Metadata.Index < db.consistentIndex` invariant is broken causing etcd to assume that etcd was crashed when persisting snapshot downloaded from leader. https://github.com/etcd-io/etcd/blob/f7be2dfa17a04bd1eb2ffc02489b7518e7eb68ad/server/storage/backend.go#L98-L112

However, based on logs this was just a normal snapshot and there is no trace of logs that would point to downloading a snapshot. I expect that somehow, the db was not properly flushed to disk https://github.com/etcd-io/etcd/blob/f7be2dfa17a04bd1eb2ffc02489b7518e7eb68ad/server/etcdserver/server.go#L2104

### What did you expect to happen?

Etcd should:
* Provide more detailed logging to  track this invariant
* Double check if db is properly flushed before saving the snapshot
* Consider recovering even if invariant is broken.

[...]

PR #17152, opened 2023-12-20
Improve logs around recovering snapshot backend and add an e2e test 

serathius:
Did some digging and found that we never added any tests for this behavior.

Improves logs and add explicit tests and endpoint around recovering snapshot backend for https://github.com/etcd-io/etcd/issues/17146

cc @ahrtr

### DIFF
--- a/server/etcdserver/bootstrap.go
+++ b/server/etcdserver/bootstrap.go
@@ -227,6 +227,14 @@ func bootstrapBackend(cfg config.ServerConfig, haveWAL bool, st v2store.Store, s
 		}
 	}
 	if beExist {
+		s1, s2 := be.Size(), be.SizeInUse()
+		cfg.Logger.Info(
+			"recovered v3 backend",
+			zap.Int64("backend-size-bytes", s1),
+			zap.String("backend-size", humanize.Bytes(uint64(s1))),
+			zap.Int64("backend-size-in-use-bytes", s2),
+			zap.String("backend-size-in-use", humanize.Bytes(uint64(s2))),
+		)
 		if err = schema.Validate(cfg.Logger, be.ReadTx()); err != nil {
 			cfg.Logger.Error("Failed to validate schema", zap.Error(err))
 			return nil, err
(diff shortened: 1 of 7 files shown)


server/etcdserver/bootstrap.go
```diff
@@ -227,6 +227,14 @@ func bootstrapBackend(cfg config.ServerConfig, haveWAL bool, st v2store.Store, s
 		}
 	}
 	if beExist {
+		s1, s2 := be.Size(), be.SizeInUse()
```
serathius, line 230: Moved this log here as it's not specific to restoring from snapshot

server/etcdserver/server.go
```diff
@@ -1001,6 +1001,7 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, toApply *toApply) {
 	// wait for raftNode to persist snapshot onto the disk
 	<-toApply.notifyc
 
+	// gofail: var beforeOpenSnapshotBackend struct{}
```
ahrtr, line 1004: When I see the name, the immediately feeling is that it's used on bootstrap when opening a snapshot backend.

```suggestion
	// gofail: var beforeOpenSnapshotBackendInApply struct{}
```

tests/e2e/leader_snapshot_no_proxy_test.go
```diff
@@ -0,0 +1,95 @@
...
+	err = failpoint.Blackhole(ctx, t, blackholedMember, epc, true)
+	require.NoError(t, err)
+	err = blackholedMember.Wait(ctx)
+	require.NoError(t, err)
+	trafficCancel()
+	wg.Wait()
+	err = blackholedMember.Start(ctx)
+	require.NoError(t, err)
+	_, err = blackholedMember.Logs().ExpectWithContext(ctx, expect.ExpectedResponse{Value: "Recovering from snapshot backend"})
+	assert.NoError(t, err)
```
ahrtr, line 94: Should we also verify the `blackholedMember` can correctly receive & process client requests after restarting?

server/etcdserver/server.go
serathius: Should it be prefix or suffix based? Failpoints for raft have a nice `raft` prefix like `raftBeforeLeaderSend`. However it's less consistent for server/apply. For example `beforeCommit`, `beforeApplyOneConfigChange`.

Food for thought. For now I will pick your proposal as I cannot think of anything better.

tests/e2e/leader_snapshot_no_proxy_test.go
serathius: Possibly, I skipped this because also added robustness scenario. However, you are right, the test should be complete. Still hard to decide what exactly to verify. If we just want to check if server is healthy, possibly this should be already done in the `Start` code.

ahrtr: I think just a sanity test (e.g. client put a key/value) is fine here.

server/etcdserver/server.go
serathius: Done

tests/e2e/leader_snapshot_no_proxy_test.go
serathius: Done

ahrtr (approved):

serathius:
/retest

Not great looks like a third flake, every time different e2e test is failing.

PR #17157, opened 2023-12-21
Commit bbolt transaction if there is any pending deleting operations

ahrtr:
This PR is going to supersede https://github.com/etcd-io/etcd/pull/17119

Note: this is the simplest solution to resolve the issue I can think of. Readability (understandability) overweights performance in the core part, also performance isn't that critical for deleting operation. 
- etcd applies MVCC to normal key/values operation, so there is no deletion for normal key/value operations (compaction is a special case)
- Usually there isn't too much/frequent deletion operations (e.g. remove member, remove user/roles, etc)

### DIFF
--- a/server/storage/backend/batch_tx.go
+++ b/server/storage/backend/batch_tx.go
@@ -73,7 +73,8 @@ type batchTx struct {
 	tx      *bolt.Tx
 	backend *backend
 
-	pending int
+	pending                 int
+	pendingDeleteOperations int
 }
 
 // Lock is supposed to be called only by the unit test.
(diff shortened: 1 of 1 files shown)


siyuanfoundation:
Since delete only affects a few internal buckets, while most of the operations are in the key bucket. Would it be better for performance if we just skip reading from buffer if there is a delete for that bucket? Something like https://github.com/etcd-io/etcd/compare/main...siyuanfoundation:etcd:txBuf

ahrtr:
> Something like [main...siyuanfoundation:etcd:txBuf](https://github.com/etcd-io/etcd/compare/main...siyuanfoundation:etcd:txBuf)

It doesn't work. It's still possible to read stable data in your proposal. 

Note 
- The buf is dedicated for put operation. 
- The reason for stale read is due to the data not being committed to bbolt.

Please read the links included in the section "Root Cause" in https://github.com/etcd-io/etcd/issues/17158.

ahrtr:

Issue #17158, opened 2023-12-21
`(*batchTx) UnsafeDelete` may break the linearizability


ahrtr:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

## TL; DR
`(*batchTx) UnsafeDelete` may break the linearizability. etcd may still read stale data when during applying a request if previous applying process includes `(*batchTx) UnsafeDelete` request(s).

## Root cause
Please refer to
- https://github.com/etcd-io/etcd/pull/17157
- https://github.com/etcd-io/etcd/pull/17119#issuecomment-1857547158
- Test case `TestRangeAfterDeleteMatch` in https://github.com/etcd-io/etcd/pull/17124

## Solution
The PR should contains the following content/commits,
- Add a couple of new test case on removing users/roles to reproduce the issue
  - Let's use the auth users as an example. Firstly add an user, then remove it and immediately after add it again. The client may get an `ErrUserAlreadyExist` error.
- Cherry-pick the test case `TestRangeAfterDeleteMatch` in https://github.com/etcd-io/etcd/pull/17124;
- Cherry-pick https://github.com/etcd-io/etcd/pull/17157: (it is the simplest solution I can think of). Confirm that it can fix the test failures
  - Please fix the test failures
- Please feel free to suggest any other test case

## Backport
The fix should be backported to both 3.5 and. 3.4

### What did you expect to happen?

Ensure linearizability on any deletion operations.

### How can we reproduce it (as minimally and precisely as possible)?

See above

### Etcd version (please run commands below)

<details>
all releases

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

serathius:
[...]

PR #10283, opened 2018-11-26
*: add flags to setup backend related config

xiang90:
The default backend batch interval is 100ms, which is too long for a fast local ssd. It means that a SSD will be almost idle for 100ms for another fdatasync to happen. 

Making backend related configuration to be configurable through flags.

PR #17195, opened 2024-01-03
Fix delete inconsistencies in read buffer

siyuanfoundation:

The PR includes fix for 

- UnsafeDelete https://github.com/etcd-io/etcd/issues/17158

### DIFF
--- a/server/storage/backend/batch_tx.go
+++ b/server/storage/backend/batch_tx.go
@@ -288,7 +288,8 @@ func (t *batchTx) commit(stop bool) {
 
 type batchTxBuffered struct {
 	batchTx
-	buf txWriteBuffer
+	buf                     txWriteBuffer
+	pendingDeleteOperations int
 }
 
 func newBatchTxBuffered(backend *backend) *batchTxBuffered {
@@ -310,7 +311,27 @@ func (t *batchTxBuffered) Unlock() {
 		t.buf.writeback(&t.backend.readTx.buf)
 		// gofail: var afterWritebackBuf struct{}
 		t.backend.readTx.Unlock()
-		if t.pending >= t.backend.batchLimit {
+		// We commit the transaction when the number of pending operations
+		// reaches the configured limit(batchLimit) to prevent it from
+		// becoming excessively large.
+		//
+		// But we also need to commit the transaction immediately if there
+		// is any pending deleting operation, otherwise etcd might run into
+		// a situation that it haven't finished committing the data into backend
+		// storage (note: etcd periodically commits the bbolt transactions
+		// instead of on each request) when it applies next request. Accordingly,
+		// etcd may still read the stale data from bbolt when processing next
+		// request. So it breaks the linearizability.
+		//
+		// Note we don't need to commit the transaction for put requests if
+		// it doesn't exceed the batch limit, because there is a buffer on top
+		// of the bbolt. Each time when etcd reads data from backend storage,
+		// it will read data from both bbolt and the buffer. But there is no
+		// such a buffer for delete requests.
+		//
+		// Please also refer to
+		// https://github.com/etcd-io/etcd/pull/17119#issuecomment-1857547158
+		if t.pending >= t.backend.batchLimit || t.pendingDeleteOperations > 0 {
 			t.commit(false)
 		}
 	}
@@ -356,6 +377,7 @@ func (t *batchTxBuffered) unsafeCommit(stop bool) {
 	}
 
 	t.batchTx.commit(stop)
+	t.pendingDeleteOperations = 0
 
 	if !stop {
 		t.backend.readTx.tx = t.backend.begin(false)
@@ -370,4 +392,14 @@ func (t *batchTxBuffered) UnsafePut(bucket Bucket, key []byte, value []byte) {
 func (t *batchTxBuffered) UnsafeSeqPut(bucket Bucket, key []byte, value []byte) {
 	t.batchTx.UnsafeSeqPut(bucket, key, value)
 	t.buf.putSeq(bucket, key, value)
+}
+
+func (t *batchTxBuffered) UnsafeDelete(bucketType Bucket, key []byte) {
+	t.batchTx.UnsafeDelete(bucketType, key)
+	t.pendingDeleteOperations++
+}
+
+func (t *batchTxBuffered) UnsafeDeleteBucket(bucket Bucket) {
+	t.batchTx.UnsafeDeleteBucket(bucket)
+	t.pendingDeleteOperations++
 }
--- a/server/storage/backend/batch_tx_test.go
+++ b/server/storage/backend/batch_tx_test.go
@@ -19,6 +19,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+
 	bolt "go.etcd.io/bbolt"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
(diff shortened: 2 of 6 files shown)


siyuanfoundation:
cc @ahrtr @serathius

server/storage/backend/tx_buffer.go
```diff
@@ -83,6 +83,7 @@ func (txw *txWriteBuffer) writeback(txr *txReadBuffer) {
 		if !ok {
 			delete(txw.buckets, k)
 			txr.buckets[k] = wb
+			wb.dedupe()
```
fuweid, line 86: Nice catch!

server/storage/backend/batch_tx_test.go
```diff
@@ -205,3 +207,109 @@ func TestBatchTxBatchLimitCommit(t *testing.T) {
...
+
+	tx.Lock()
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar3"))
+	tx.Unlock()
+
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0)
+	checkTxnResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo1"), nil, 0)
+}
+
+func checkTxnResponseMatch(t *testing.T, tx backend.BatchTx, rtx backend.ReadTx, key, endKey []byte, limit int64) {
```
fuweid, line 299: like this idea. do we need to verify that result? I mean that just in case `batchTx` doesn't read bbolt directly.

serathius, line 299: This only checks the Range response, could we also test `ForEach`?

server/storage/backend/batch_tx.go
```diff
@@ -105,7 +106,23 @@ func (t *batchTx) LockOutsideApply() {
 }
 
 func (t *batchTx) Unlock() {
-	if t.pending >= t.backend.batchLimit {
+	// We need to commit the transaction if there is any pending deleting
```
serathius, line 109: I think this could make the purpose of this code clearer, my understanding is that:
* `t.pending >= t.backend.batchLimit` is for performance. We don't want buffer to grow too big as it has worse complexity.
* `t.pendingDeleteOperations > 0` is for correctness as buffer doesn't keep track of deletions.

server/storage/backend/batch_tx.go
```diff
@@ -133,6 +150,7 @@ func (t *batchTx) UnsafeDeleteBucket(bucket Bucket) {
 		)
 	}
 	t.pending++
+	t.pendingDeleteOperations++
```
serathius, line 153: From what I remember bucket operations are not buffered directly, but somewhat implicitly. Will just tracking pending deletions work for them? Have you tested this? Could you add a test for `UnsafeDeleteBucket`?

server/storage/backend/batch_tx.go
```diff
@@ -73,7 +73,8 @@ type batchTx struct {
 	tx      *bolt.Tx
 	backend *backend
 
-	pending int
+	pending                 int
+	pendingDeleteOperations int
```
serathius, line 77: It's not clear for me why we track deletions on the `batchTx` instead on the `batchTxBuffered`. The issue we originally discussed is about buffer in `batchTxBuffered` not representing deletes. Is there any logic in the `batchTx` I missed that would also require handling this?

server/storage/backend/batch_tx_test.go
siyuanfoundation: added.

siyuanfoundation: added key and value checks.

server/storage/backend/batch_tx.go
siyuanfoundation: done

server/storage/backend/batch_tx.go
siyuanfoundation: added a test for `UnsafeDeleteBucket`. It would fail without the fix.

ahrtr:
> The PR includes fix for
> 
> * UnsafeDelete [`(*batchTx) UnsafeDelete` may break the linearizability #17158](https://github.com/etcd-io/etcd/issues/17158)
> * I found another problem if there is duplicate keys in the same transaction (overwrite).

This PR is changing one of the critical parts of etcd, please let to breaking it into two separate PRs. Each only fixes one problem. For each PR, the first commit only includes test cases to reproduce the issue you are going to fix in the 2nd or 3rd commits. Ensure you can reproduce the issue before you adding the 2nd or 3rd commit/patch, and can't reproduce it anymore after adding the patch.

server/storage/backend/batch_tx.go
ahrtr: It's natural and easy to read to track two metrics (`pending***`) in one place. It isn't a big problem.

fuweid (approved):
LGTM

serathius:
> This PR is changing one of the critical parts of etcd, please let's breakdown it into two separate PRs. Each only fixes one problem. For each PR, the first commit only includes test cases to reproduce the issue you are going to fix in the 2nd or 3rd commits. Ensure you can reproduce the issue before you adding the 2nd or 3rd commit/patch, and can't reproduce it anymore after adding the patch.

Agree with that, we need to make sure that each of those issues are properly understood and tested. Having a single PR for both adds unnecessary cognitive overhead.

server/storage/backend/batch_tx.go
serathius: Don't agree, but give me some time to read the code again.

server/storage/backend/batch_tx.go
```diff
@@ -310,7 +314,14 @@ func (t *batchTxBuffered) Unlock() {
 		t.buf.writeback(&t.backend.readTx.buf)
 		// gofail: var afterWritebackBuf struct{}
 		t.backend.readTx.Unlock()
-		if t.pending >= t.backend.batchLimit {
+		// We need to commit the transaction if
+		// 1. t.pending >= t.backend.batchLimit for performance,
+		// 	  because We don't want buffer to grow too big as it has worse complexity.
+		// 2. there is any pending deleting for correctness,
+		//    because buffer doesn't keep track of deletions.
```
serathius, line 321:
```suggestion
		//    because batchTxBuffered doesn't keep track of deletions.
```

server/storage/backend/batch_tx.go
```diff
@@ -310,7 +314,14 @@ func (t *batchTxBuffered) Unlock() {
 		t.buf.writeback(&t.backend.readTx.buf)
 		// gofail: var afterWritebackBuf struct{}
 		t.backend.readTx.Unlock()
-		if t.pending >= t.backend.batchLimit {
+		// We need to commit the transaction if
+		// 1. t.pending >= t.backend.batchLimit for performance,
+		// 	  because We don't want buffer to grow too big as it has worse complexity.
```
serathius, line 319: That's my current understanding, however would be good to confirm. Could you check PRs when the batchlimit was introduced and see what was the motivation back then. Referencing PR would be also useful.

cc @ahrtr

server/storage/backend/batch_tx_test.go
```diff
@@ -205,3 +207,110 @@ func TestBatchTxBatchLimitCommit(t *testing.T) {
...
+	checkForEachResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, [][]byte{[]byte("foo")}, [][]byte{[]byte("bar")})
+
+	tx.Lock()
+	tx.UnsafeDelete(schema.Test, []byte("foo"))
+	tx.Unlock()
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0, nil, nil)
+	checkForEachResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, nil, nil)
+}
+
+func checkRangeResponseMatch(t *testing.T, tx backend.BatchTx, rtx backend.ReadTx, key, endKey []byte, limit int64, expectedKeys, expectedValues [][]byte) {
```
serathius, line 260: By adding validation of expected values you mixed assertion based testing with property based testing. Let's pick one of them.

Originally I implemented the test as property testing. The property I was testing was "ReadTx should always return the same response as BatchTX". This property should always hold, no matter what operations are executed before. I did it to allow make validation independent from test scenario and potentially randomize operations before validation.

Recent changes to the PR added `expectedKeys` and `expectedValues` testing, this means you hardeded the validation to specific cases. This is still correct, however the code can be much more simplified.

server/storage/backend/batch_tx.go
serathius: I worry about separation of concern in the batch_tx.go package. Fact that `batchTx` and `batchTxBuffered` are used together doesn't mean we should entangle them. For example for some reason precommit hooks are executed by of `batchTxBuffered` and not `batchTx`. Why? I don't know, maybe hook call location needed to be ordered in relation to buffer, however It's hard to know now without deep dive.

I agree that this isn't a big problem now, simplifying the transaction logic would be another large effort that we can postpone for later.

server/storage/backend/batch_tx.go
siyuanfoundation: this is one relevant pr https://github.com/etcd-io/etcd/pull/10283

tests/integration/clientv3/user_test.go
```diff
@@ -55,6 +55,66 @@ func TestUserError(t *testing.T) {
 	}
 }
 
+func TestAddUserAfterDelete(t *testing.T) {
```
ahrtr, line 58: Have you confirmed that the test case can reproduce the issue?

tests/integration/clientv3/user_test.go
```diff
@@ -55,6 +55,66 @@ func TestUserError(t *testing.T) {
...
+	cfg := clientv3.Config{
+		Endpoints:   authapi.Endpoints(),
+		DialTimeout: 5 * time.Second,
+		DialOptions: []grpc.DialOption{grpc.WithBlock()},
+	}
+	cfg.Username, cfg.Password = "root", "123"
+	authed, err := integration2.NewClient(t, cfg)
+	if err != nil {
+		t.Fatal(err)
+	}
```
ahrtr, line 75: Minor comment, please update all the following such usage for simplicity.
```suggestion
	require.NoError(t, err)
```

tests/integration/member_test.go
```diff
@@ -115,3 +119,52 @@ func TestSnapshotAndRestartMember(t *testing.T) {
 		}
 	}
 }
+
+func TestRemoveMember(t *testing.T) {
+	integration.BeforeTest(t)
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 3, UseBridge: true, BackendBatchInterval: 1000 * time.Second})
+	defer c.Terminate(t)
+	time.Sleep(etcdserver.HealthInterval)
```
ahrtr, line 127:
```suggestion
	c.WaitLeader(t)
```

tests/integration/member_test.go
```diff
@@ -115,3 +119,52 @@ func TestSnapshotAndRestartMember(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	_, err := c.Members[2].Client.MemberRemove(ctx, uint64(c.Members[0].Server.MemberId()))
+	cancel()
+
+	if err != nil {
+		t.Fatalf("should accept removing member: %s", err)
+	}
+	c.Members[0].Stop(t)
+	time.Sleep(etcdserver.HealthInterval)
```
ahrtr, line 137:
```suggestion
	c.WaitLeader(t)
```

tests/integration/member_test.go
```diff
@@ -115,3 +119,52 @@ func TestSnapshotAndRestartMember(t *testing.T) {
...
+
+	checkMemberCount(t, c.Members[1], 2)
+	checkMemberCount(t, c.Members[2], 2)
+}
+
+func checkMemberCount(t *testing.T, m *integration.Member, expectedMemberCount int) {
+	memberCount := mustReadMemberCountFromBackend(t, m.Server.Backend())
+	if memberCount != expectedMemberCount {
+		t.Errorf("Expect MemberCountFromBackend=%d, got %d", expectedMemberCount, memberCount)
+	}
```
ahrtr, line 147: This doesn't make much sense to me. bbolt not returning latest data doesn't mean etcd can't return the latest data. We are testing etcd instead of bbolt.

tests/integration/member_test.go
serathius: This is not about waiting for leader, membership changes additionally require cluster to be stable for `etcdserver.HealthInterval`. Please leave a comment to explain this.

tests/integration/clientv3/user_test.go
siyuanfoundation: Confirmed the test case if the [defensive commit](https://github.com/etcd-io/etcd/blob/f66b95d3b1374beb56dc772c587832f15bb6e697/server/storage/schema/auth.go#L114) is removed. Because auth force commits for every change, this is not actually a problem for auth right now. 

I am adding this test for completeness.

server/storage/backend/batch_tx_test.go
siyuanfoundation: In unit test, the test cases are hard coded. I think it is important to test the actual key and values in case ReadTx and BatchTX make the same mistake. 
Changed the code to just check the results in `checkForEach`

tests/integration/member_test.go
siyuanfoundation: This test is to test the assumption `MustReadMembersFromBackend` should return the correct members, like in https://github.com/etcd-io/etcd/issues/17052. 
Refactored to make it more clear.

tests/integration/clientv3/user_test.go
serathius: Wow, nice find. This defensive commit seems scary, similar to https://github.com/etcd-io/etcd/pull/17119 we noticed some inconsistency and didn't root cause it.

tests/integration/member_test.go
```diff
@@ -115,3 +118,31 @@ func TestSnapshotAndRestartMember(t *testing.T) {
...
+func TestRemoveMember(t *testing.T) {
+	integration.BeforeTest(t)
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 3, UseBridge: true, BackendBatchInterval: 1000 * time.Second})
+	defer c.Terminate(t)
+	// membership changes additionally require cluster to be stable for etcdserver.HealthInterval
+	time.Sleep(etcdserver.HealthInterval)
+
+	err := c.RemoveMember(t, c.Client(2), uint64(c.Members[0].ID()))
+	require.NoError(t, err)
+	// membership changes additionally require cluster to be stable for etcdserver.HealthInterval
```
serathius, line 131: Hmm there are no membership changes after, I think this is a linearizability issue.

siyuanfoundation: Removing this wait because `RemoveMember` already `WaitMembersMatch`

serathius (approved):

server/etcdserver/api/cluster.go
```diff
@@ -35,4 +35,6 @@ type Cluster interface {
 	Member(id types.ID) *membership.Member
 	// Version is the cluster-wide minimum major.minor version.
 	Version() *semver.Version
+
+	MustReadMembersFromBackend() (map[types.ID]*membership.Member, map[types.ID]bool)
```
ahrtr, line 39: `Cluster` is a client-facing interface, It's weird to expose the backend functionality to the cluster interface directly. If it's only used by the test, then you can update the test something like below,

```
	be := schema.NewMembershipBackend(m.Logger, m.Server.Backend())
	members, removedMember := be.MustReadMembersFromBackend()

```

siyuanfoundation: good idea!

ahrtr (approved):
lgtm

Could you backport the fix to 3.5 and 3.4? thx

PR #17198, opened 2024-01-04
receipt: dequeue not returning watch err

wxsms:
the error from watch is not returned, if the client token expired, the `ev` return by watch will be nil, therefore the Deque method will panic in such case.


serathius:
Can you add a test?

wxsms:
Hi there, Thanks for reply, I just added a test to simulate an error during watch by shutting down a cluster, since I'm very new to etcd, not sure if I make it correctly. The main propose of this PR is to make sure watch return a non-nil value or error. The testcase will failed as followed if queue code stay untouched:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xad0bb6]

goroutine 169 [running]:
go.etcd.io/etcd/client/v3/experimental/recipes.(*Queue).Dequeue(0xc00045b290)
        /workspaces/etcd/client/v3/experimental/recipes/queue.go:70 +0x1f6
go.etcd.io/etcd/tests/v3/integration/clientv3/experimental/recipes_test.TestQueueWatchError.func1()
        /workspaces/etcd/tests/integration/clientv3/experimental/recipes/v3_queue_test.go:92 +0x25
created by go.etcd.io/etcd/tests/v3/integration/clientv3/experimental/recipes_test.TestQueueWatchError in goroutine 64
        /workspaces/etcd/tests/integration/clientv3/experimental/recipes/v3_queue_test.go:91 +0x153
FAIL    go.etcd.io/etcd/tests/v3/integration/clientv3/experimental/recipes      1.107s
```

wxsms:
Sry but I have no idea why `integration/clientv3 TestMaintenanceSnapshotCancel` failing, every thing's fine while running tests  inside my codespace, including this one. Can anyone help to take a look?

siyuanfoundation:
cc @moficodes

Issue #17194, opened 2024-01-03
Etcd leaking snapshot files


serathius:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Etcd was leaking snapshot files, we found over 50`000 of them. Normally etcd should only keep last 5 snapshot files, but for some unknown reason it stopped purging them. 

I suspect etcd was unable to flock file for some reason and caused purgeFile function to exit early https://github.com/etcd-io/etcd/blob/85b640cee793e25f3837c47200089d14a8392dc7/pkg/fileutil/purge.go#L39C63-L98

### What did you expect to happen?

Snapshots are not flocked by etcd.T here is no reason for etcd to flock snapshots before deletion, this logic seems borrowed from WAL purging (reuse of same logic in https://github.com/etcd-io/etcd/blob/85b640cee793e25f3837c47200089d14a8392dc7/etcdserver/server.go#L837.

My recommendation Etcd should not flock snapshots ("snap" and "snap.db" files) and  before deletion.

### How can we reproduce it (as minimally and precisely as possible)?

n/a

### Etcd version (please run commands below)

v3.4.21

### Etcd configuration (command line flags or environment variables)

n/a

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

n/a

serathius:
Looks like https://github.com/etcd-io/etcd/pull/1927 unintentionally added flocking snapshot files before deletion.

serathius:
And https://github.com/etcd-io/etcd/pull/14576 added more logging.

fuweid:
If no process holds lock on snap dir, `ulimit -x` is ulimited? maybe the user holds too many flocks

ahrtr:
[...]

PR #1927, opened 2014-12-12
*: lock the in using files; do not purge locked the wal files

xiang90:
Fix #1659

PR #17199, opened 2024-01-04
Don't flock snapshot files

serathius:
Ref https://github.com/etcd-io/etcd/issues/17194

Flocking to snapshot files was added unintentionally in https://github.com/etcd-io/etcd/pull/1927.
Even at this time `PurgeFile` function was used for both wal and snapshot files https://github.com/etcd-io/etcd/blob/502396edd53140cd2a099fb5d7a9f59ec79f4475/etcdserver/server.go#L323-L339

Did minimal changes to avoid flocking files. Don't want to refactor it now to allow easy backport to older versions.

cc @ahrtr

### DIFF
--- a/client/pkg/fileutil/purge.go
+++ b/client/pkg/fileutil/purge.go
@@ -25,18 +25,24 @@ import (
 )
 
 func PurgeFile(lg *zap.Logger, dirname string, suffix string, max uint, interval time.Duration, stop <-chan struct{}) <-chan error {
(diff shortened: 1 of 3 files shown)


fuweid:
Not sure I understand correctly, if that server recovered from snapshot db, server will flock it by bbolt. If purge goroutine doesn't try lock, it will delete using data.

serathius:
ping @ahrtr

ahrtr (approved):

PR #17200, opened 2024-01-04
Automate installation of yamllint

iiamabby:
@jmhbnz and I are working on refining the test-infra prow workflow to harmonise with the legacy GitHub Actions workflow.

We noticed the [test-infra/config/jobs/etcd/etcd-presubmits.yaml](https://github.com/kubernetes/test-infra/blob/551ef90f7a3efa1560b593bf0c6eea7290bee516/config/jobs/etcd/etcd-presubmits.yaml#L69) is using `make verify-lint` whilst the [etcd/.github/workflows/static-anaylsis.yaml](https://github.com/etcd-io/etcd/blob/db4e95cb1f8d3a8eebfb89c880d5fbd8d6165292/.github/workflows/static-analysis.yaml#L37) is using `make verify`

We want to make a change from `make verify-lint` to `make verify` within the test-infra repo

When testing the `make verify` command in the test-infra image: `gcr.io/k8s-staging-test-infra/kubekins-e2e:v20231208-8b9fd88e88-master` there was an error relating to yamllint not being installed

This proposed change adds the automation of installing yamllint from pip which is the standard method of installing yamllint. 

Note: We also noticed a duplicate spellcheck command which has been removed.

Added a second commit to fix a minor bug with the previous protoc automation where `bin` directory may not yet exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ fuzz:
 # Static analysis
 
 verify: verify-gofmt verify-bom verify-lint verify-dep verify-shellcheck verify-goword \
-	verify-govet verify-license-header verify-receiver-name verify-mod-tidy verify-shellcheck \
+	verify-govet verify-license-header verify-receiver-name verify-mod-tidy \
 	verify-shellws verify-proto-annotations verify-genproto verify-yamllint \
 	verify-govet-shadow verify-markdown-marker
 fix: fix-bom fix-lint fix-yamllint
@@ -131,7 +131,15 @@ verify-genproto:
 
 .PHONY: verify-yamllint
 verify-yamllint:
+ifeq (, $(shell which yamllint))
+	@echo "Installing yamllint..."
+	python3 -m venv bin/python
+	bin/python/bin/python3 -m pip install yamllint
+	./bin/python/bin/yamllint --config-file tools/.yamllint .
+else
+	@echo "yamllint already installed..."
 	yamllint --config-file tools/.yamllint .
+endif
 
 .PHONY: verify-govet-shadow
 verify-govet-shadow:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -63,7 +63,7 @@ fuzz:
 # Static analysis
 
 verify: verify-gofmt verify-bom verify-lint verify-dep verify-shellcheck verify-goword \
-	verify-govet verify-license-header verify-receiver-name verify-mod-tidy verify-shellcheck \
+	verify-govet verify-license-header verify-receiver-name verify-mod-tidy \
```
serathius, line 66: Why remove `verify-shellcheck`?

Makefile
```diff
@@ -131,7 +131,15 @@ verify-genproto:
 
 .PHONY: verify-yamllint
 verify-yamllint:
+ifeq (, $(shell which yamllint))
+	@echo "Installing yamllint..."
+	python3 -m venv bin/python
```
serathius, line 136: Please update the `make clean` method to cover `bin/python`

serathius, line 136: Solution using local virtualenv is nice and thought out. Not a great fun of adding python dependency, still we don't have a choice as we decided to use a python tool for linting.

Would be interested if there is any other tool that has both fix (yamlfmt) and (yamllint) functionality. Maybe we should just use yamlfmt and implement our own validation based on diff like we did for gen proto. Don't want to block migration to test-infra on this, so we can use python for now.

serathius (approved):

Makefile
jmhbnz: Agreed - We should move away from `yamllint` in future to avoid bringing in python dependencies.

ahrtr (approved):
lgtm

Makefile
iiamabby: This was a duplicate, refer to line 65 :)

PR #17122, opened 2023-12-15
Add digest for base image and configure dependabot

wespanther:
This should not change the current behavior, but will cause dependabot to update the digest of the latest base image weekly. 

Related: https://github.com/etcd-io/etcd/issues/16987

Signed-off-by: Wes Panther <wpanther@google.com>

Issue #16987, opened 2023-11-22
Add digest to Dockerfile base image reference via dependabot


wespanther:
### What would you like to be added?

Hey folks,

This is more of a suggestion. I stumbled on https://github.com/etcd-io/etcd/pull/16942 and it made me think that it might be desirable to have a digest for the the etcd base image - `gcr.io/distroless/static-debian12` - in `/Dockerfile`. This could be done automatically by dependabot  by modifying the existing config. This would just entail adding a digest to that image and something like this to the existing `.github/dependabot.yml`:

```
- package-ecosystem: docker
  directory: /
  schedule:
    interval: weekly
```

I'm happy to make the change if it's agreeable.

### Why is this needed?

This would improve supply chain security (e.g. relying on the `latest` tag and make builds more easily reproducible). I wouldn't want someone to have to updated these digest manually, but if it's automated by dependabot I believe the benefits outweigh the slight additional overhead (a PR generated weekly, for example). We could change the schedule to whatever is acceptable.

jmhbnz:
Thanks for raising the idea @wespanther.  Agree this would be a valuable improvement to supply chain security.

I'm supportive of trialing it.  Let's see what others think over the coming week and if we have consensus please feel free to raise the pull request thanks 🙏🏻

jmhbnz:
Hey @wespanther - If you still have time please go ahead with raising this supply chain improvement 🙏🏻

jmhbnz:
Hey @wespanther - Many thanks for raising this improvement for our `main` development branch. With that now merged we should raise new pull requests to backport the change to our `release-3.4` and `release-3.5` stable release branches so etcd users can start benefiting from the additional release reproducibility.

Please raise these if you have time, if not all good just let us know and we can tackle it.

wespanther:
> Please raise these if you have time, if not all good just let us know and we can tackle it.

Sorry, I missed your message previously. Thanks for taking on the backports!

PR #17205, opened 2024-01-08
[3.5] Backport adding digest for etcd base image

jmhbnz:
This pull request proposes backporting https://github.com/etcd-io/etcd/pull/17122 to `release-3.5`.

Currently etcd official release image builds are not reproducible as we currently use the mutable `latest` tag for our distroless base image. With this backport they would become reproducible through a switch to immutable digests.

It needs to be called out there is a tradeoff that before each release we would ideally consider updating to the latest upstream image sha to ensure our reproducible builds are using the most current base image.  For `main` we have dependabot keeping on top of this.  For release branches I don't believe dependabot can be configured to analyse non default branch so we would need to do it manually.

Relates to https://github.com/etcd-io/etcd/issues/16987.


fuweid (approved):
LGTM

9be3fcc6abeaf985b5ecce59451acbcbb15e7be39472320c538d0d55a0834edc is for latest tag

serathius (approved):

ahrtr (approved):

PR #17219, opened 2024-01-09
Fix Struct Config has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation. 

devincd:
Struct Config has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation. 

In general, all methods on a given type should have either value or pointer receivers, but not a mixture of both.

https://go.dev/tour/methods/8

### DIFF
--- a/server/embed/config_logging.go
+++ b/server/embed/config_logging.go
@@ -34,7 +34,7 @@ import (
 )
 
 // GetLogger returns the logger.
-func (cfg Config) GetLogger() *zap.Logger {
+func (cfg *Config) GetLogger() *zap.Logger {
 	cfg.loggerMu.RLock()
 	l := cfg.logger
 	cfg.loggerMu.RUnlock()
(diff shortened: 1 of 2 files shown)


server/embed/config_logging.go
```diff
@@ -34,7 +34,7 @@ import (
 )
 
 // GetLogger returns the logger.
-func (cfg Config) GetLogger() *zap.Logger {
+func (cfg *Config) GetLogger() *zap.Logger {
 	cfg.loggerMu.RLock()
```
serathius, line 38: This looks even as a bug. Locking mutex on value receiver means we are locking a copy of lock.

serathius (approved):

ahrtr (approved):

jmhbnz (approved):
LGTM - Thanks @devincd for this great first contribution to etcd 🎉

server/embed/config_logging.go
ahrtr: It isn't a bug, the lock itself is a pointer.

PR #17233, opened 2024-01-12
Creation time is added to wal records

### LINKED ISSUE #16962: Point in time recovery in ETCD

So I recently started implementing [WAL_G](https://github.com/wal-g/wal-g.git) functionality for etcd and while doing it I noticed that most of the things are already implemented here. 

So you already Save WAL(state and entries) and snapshots to underlying stable storage. The only thing left behind is point in time recovery aka the ability for user to restore storage to a specific point in time using the closest snapshot and WAL. 

Right now you only provide restoration to the last entry in WAL using snapshot. 

### Why is this needed?

PTR can be useful in case we want to skip last entries (for instance in order to recover deleted data etc) and/or rewrite it

KosovGrigorii:

This PR resolves #16962


x4m (approved):

Denchick:
/cc @ahrtr

Hi!

We're currently working on the [SPQR project](https://github.com/pg-sharding/spqr), which is about PostgreSQL sharding. We store our metadata in etcd and already know how to restore shards to a specific point in time using [WAL-G](https://github.com/wal-g/wal-g/).

But there is no PITR support for etcd, so #17233 is very important to us. Could you please to approve us to run GitHub Workflows?

x4m:
Bump

x4m:
@serathius thank you for your inquiry.

When restoring a database from a backup to a point-in-time we do not need to coordinate nodes from a fault tolerance group. We need just one, restoring to specific point and leaving rest history aside.
In a sharded cluster point-in-time is trickier. E.g. in Greenplum we injection points-of-consistency into WAL where no multishard transactions are in-flight. In MongoDB we have a global history marks, so we translate given time into history mark on any shard. And then restore cluster to that mark.
I do not know which approach will be taken. But we did it, and did it many times.

> It never depends on absolute time

Indeed. Response time of almost a year is completely novel to us. Developers who were working on the project already changed a team. We will find someone, of course.

ahrtr:
- As mentioned above by @serathius , we don't save any timestamp anywhere. We don't see any value to etcd itself to add timestamp into WAL file.
- etcd stores data in [bbolt](https://github.com/etcd-io/bbolt), WAL data is just protection mechanism for potential data loss in abnormal cases, i.e. VM or process crashes. I don't see how to do point-in-time recovery using WAL files/data.

x4m:
@ahrtr consider symmetric point of view: etcd stores data in WAL, bbolt is just a cache for the case when no crashes happened.
To support this point of view I can name some invariants, that I think are uphold:
1. WAL is always flushed before bbolt is written
2. Each individual WAL record might be last during recovery, still leaving database in consistent state.

PITR is just a recovery constrained by some point in time. That's how PITR is working in Oracle, Postgres, MS-SQL, Cloudberry and many other databases.

serathius:
> PITR is just a recovery constrained by some point in time. That's how PITR is working in Oracle, Postgres, MS-SQL, Cloudberry and many other databases.

Those databases are all single master that decides the time. Etcd explicitly avoids trusting clock of any member and doesn't use absolute time. Not saying that you could not solve the problem, but it doesn't seem like something that could be easily integrated into etcd.

x4m:
@serathius it is not a problem for PITR at all, just use any source of time.

> Those databases are all single master that decides the time.

No, it's not true. We have different timelines, we have different primaries in different shards, and of course we have different primaries at different moments.

In Postgres HA installation you have a WAL time recordeded according to clocks of current Primary for every WAL record. But you easily can promote Standby at any given moment. And the history would be recorded according to clocks of current Primary.

In MySQL stuff is much more interesting, because you have a vector clock + GTID - every node writes it's own binlogs, and nodes slice logs differently. So we record history with the overlap. E.i. when primary switches over to secondary, we have some tail of committed transactions on old primary, that are available on secondary too, we deduplicate them using GTID and use earliest timestamp from different logs.

Leaderless replication is much easier wrt PITR. It is not even shardeded, so we do not have to deal with cross-shard consistency as we do in MongoDB, Greenplum and Cloudberry.

Please, describe any specific problem, not generic "Etcd explicitly avoids trusting clock". Because, for point-in-**time** you need time anyway.

Time is obviously non-monotonic and you cannot trust it for event ordering within database. But for matching events with real world - it's perfectly fine. You can trust input of your user like "yesterday my data was OK", or "I've dropped important information 1 hour ago". No matter what time source you use for determining where to top recovery.

x4m:
To prevent developers from misusing timestamps we can truncate it to, say, minutes. Anyone who will try to order something with this timestamp will quickly discovery that it's impossible. And for PITR it's OK if you can specify recovery minute, not a millisecond.

ahrtr:
> 2\. Each individual WAL record might be last during recovery, still leaving database in consistent state.

This might not be true. Only committed WAL records are guaranteed to be safe. For example, if you recover to a WAL record, which wasn't committed, then you might end up committing a failed client write.



> Do you mean "etcd project don't see any value in PITR"?

No any value to etcd itself to add timestamp into WAL file.

Backup & restore is important. Also I did not say PITR isn't important. I was saying I do not see how easily to do point-in-time recovery using WAL files/data.

Usually PITR depends on a base backup (full backup, i.e snapshots) + incremental logs (i.e. WAL logs).  But in etcd, the V2 snapshots files (*.snap files) do not contain any real k/v data, and we are deprecating it.  All data are stored in [bbolt db](https://github.com/etcd-io/bbolt), but the relationship between `bbolt and WAL files` isn't `snapshot + incremental logs` (usually a snapshot is static and immutable, but bbolt db is always dynamic). It's hard to recover to a point(data record) which is already in bbolt db.

It isn't clear how you will implement PITR. Are you going to do periodically backup for the bbolt db & WAL files, and select a right one which falls into the time range when doing PITR?

x4m:
@ahrtr currently our WIP state is described [here](https://github.com/wal-g/wal-g/blob/master/docs/ETCD.md#usage). We do basebackups and apply WAL that happened after snapshot. If we got something wrong - let's redesing, but so far we only hear iterations about properties of time that are very well known and do not actually create a problem.

Issue #17245, opened 2024-01-15
Etcd extentensivly logging error "bucket already exists"


serathius:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Looked at etcd logs in CI and found a lot of `Creating bucket "XXX" failed: bucket already exists` logs. This happens 1 time when starting from empty data dir and 11 times on etcd restart.

This happens on `main` branch, however with recent changes and backports to backend transactions, it might also appear in release-3.4 and release-3.5. Needs to be confirmed.

### What did you expect to happen?

I would expect that etcd doesn't write error logs during normal correct function, especially in such big number. 

### How can we reproduce it (as minimally and precisely as possible)?

`make && ./bin/etcd`

### Etcd version (please run commands below)

`main` branch

### Etcd configuration (command line flags or environment variables)

n/a

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

N/A

### Relevant log output

```Shell
{"level":"warn","ts":"2024-01-15T16:35:55.284938+0100","caller":"embed/config.go:893","msg":"Running http and grpc server on single port. This is not recommended for production."}
{"level":"warn","ts":"2024-01-15T16:35:55.285045+0100","caller":"embed/config.go:995","msg":"it isn't recommended to use default name, please set a value for --name. Note that etcd might run into issue when multiple members have the same default name","name":"default"}
{"level":"info","ts":"2024-01-15T16:35:55.28506+0100","caller":"etcdmain/etcd.go:63","msg":"Running: ","args":["./bin/etcd"]}
{"level":"warn","ts":"2024-01-15T16:35:55.285085+0100","caller":"etcdmain/etcd.go:95","msg":"'data-dir' was empty; using default","data-dir":"default.etcd"}
[...]

PR #17249, opened 2024-01-15
Replace `CreateBucket` with `CreateBucketIfNotExists`

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/17245

etcd creates the meta bucket twice on startup, 
- https://github.com/etcd-io/etcd/blob/4bc55e520861f6998d3e8b8b919c1b3abaeb487f/server/etcdserver/bootstrap.go#L212
- https://github.com/etcd-io/etcd/blob/4bc55e520861f6998d3e8b8b919c1b3abaeb487f/server/storage/mvcc/kvstore.go#L118

Overall, it isn't a big problem. But let's replace `CreateBucket` with `CreateBucketIfNotExists` to remove the error message, to avoid any confusion.

### DIFF
--- a/server/storage/backend/batch_tx.go
+++ b/server/storage/backend/batch_tx.go
@@ -112,8 +112,7 @@ func (t *batchTx) Unlock() {
 }
 
 func (t *batchTx) UnsafeCreateBucket(bucket Bucket) {
-	_, err := t.tx.CreateBucket(bucket.Name())
-	if err != nil && err != bolt.ErrBucketExists {
+	if _, err := t.tx.CreateBucketIfNotExists(bucket.Name()); err != nil {
 		t.backend.lg.Fatal(
 			"failed to create a bucket",
 			zap.Stringer("bucket-name", bucket),


jmhbnz (approved):
LGTM - Thanks @ahrtr

server/storage/backend/batch_tx.go
```diff
@@ -112,7 +112,7 @@ func (t *batchTx) Unlock() {
 }
 
 func (t *batchTx) UnsafeCreateBucket(bucket Bucket) {
-	_, err := t.tx.CreateBucket(bucket.Name())
+	_, err := t.tx.CreateBucketIfNotExists(bucket.Name())
 	if err != nil && err != bolt.ErrBucketExists {
```
serathius, line 116:
```suggestion
	if err != nil {
```

serathius (approved):

server/storage/backend/batch_tx.go
ahrtr: updated & thx

ahrtr:
> There is one optional followup. Etcd should not write error logs during normal function. We could add a e2e test that confirms that.

Agreed that it's nice to have.

Issue #17259, opened 2024-01-16
Increase number of exploratory runs in robustness tests


serathius:
### What would you like to be added?

Proposal:
* Separate exploratory and regression tests in robustness. One idea would be to add prefix to names:
  *  rename exploratory tests (`TestRobustness/Kubernetes/*`, `TestRobustness/Etcd/*`) -> `TestRobustness/Exploratory/Kubernetes/*` and `TestRobustness/ExploratoryEtcd/*`
    *  rename regression tests (`TestRobustness/Issues*`) -> `TestRobustness/Regression/Issue/*` 
* Run exploratory tests more often than regression tests. For example run only exploratory tests during nightly.

### Why is this needed?

We have limited number of runs we can execute. We should focus on exploratory tests to improve our chances of finding new issues. 

https://github.com/etcd-io/etcd/issues/17247#issuecomment-1893598759

PR #17261, opened 2024-01-16
Separate robustness test scenarios and increase number of times we run exploratory tests in nightly

serathius:
Not the best solution, but should be good enough without needing to wrap/customize golang testing.

For TestRobustness we run:
* 8x TestRobustnessExploratory
* 5x TestRobustnessRegression

Which is not as bad as I thought, still we can double number of times we do exploratory test that randomizes etcd configuration and failpoint.

Keeping regression test for CI.

cc @ahrtr @siyuanfoundation 

Fixes https://github.com/etcd-io/etcd/issues/17259


siyuanfoundation:
LGTM

jmhbnz (approved):
LGTM - Thanks @serathius

fuweid (approved):
LGTM

PR #17262, opened 2024-01-16
WIP: Testing ci

jmhbnz:
Do not merge this pr.

Just checking if the updates to `pull-etcd-verify` prow job worked.


jmhbnz:
Based on logs https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/17262/pull-etcd-verify/1747359198716366848 it looks like we need to enable strict mode, and also ensure the automated install of `golangci-lint` is working in the kubekins-e2e image when running in prow.  It did work locally when running inside that image so more investigation required.

cc @iiamabby - let's take a look at it during our pair session this week :)

Issue #17247, opened 2024-01-15
Duplicated watch event detected in robustness test


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

2 robustness test failures:
* https://github.com/etcd-io/etcd/actions/runs/7511619449
* https://github.com/etcd-io/etcd/actions/runs/7518298280

Common elements:
* Both failures have a duplicated watch event
* Both on Kubernetes traffic
* Both on 3 node cluster

Differences:
* Failpoints: `raftAfterSaveSnap=panic()` and `MemberReplace`

Both failures on have duplicated watch event:
```
2024-01-13T09:50:27.4782520Z     validate.go:41: Failed merging watch history to create event history, skipping further validation, err: events between clients 11 and 9 don't match, revision: 800, diff:   []model.WatchEvent{
2024-01-13T09:50:27.4786636Z           	{Event: {Type: "put-operation", Key: "/registry/pods/default/76Z51", Value: {Value: "1061"}}, Revision: 800},
2024-01-13T09:50:27.4788612Z         - 	{
2024-01-13T09:50:27.4789499Z         - 		Event: model.Event{
2024-01-13T09:50:27.4790382Z         - 			Type:  "put-operation",
2024-01-13T09:50:27.4791612Z         - 			Key:   "/registry/pods/default/76Z51",
2024-01-13T09:50:27.4792744Z         - 			Value: model.ValueOrHash{Value: "1061"},
2024-01-13T09:50:27.4793619Z         - 		},
2024-01-13T09:50:27.4794433Z         - 		Revision: 800,
2024-01-13T09:50:27.4795136Z         - 	},
2024-01-13T09:50:27.4795687Z           }
```
```
2024-01-14T10:24:58.3514126Z     validate.go:41: Failed merging watch history to create event history, skipping further validation, err: events between clients 10 and 7 don't match, revision: 1583, diff:   []model.WatchEvent{
[...]

PR #17228, opened 2024-01-11
Fix tx buffer inconsistency if there are duplicate keys in one tx.

siyuanfoundation:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

related to https://github.com/etcd-io/etcd/issues/17158

PR #17248, opened 2024-01-15
Try reproduce https://github.com/etcd-io/etcd/issues/17247

serathius:
Change-Id: Ieed5f16ded627ac27f6acae99fea23211e6f718d

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #17263, opened 2024-01-16
Fix tx buffer inconsistency if there are unordered key writes in one tx.

siyuanfoundation:
Recommit https://github.com/etcd-io/etcd/pull/17228 with buf fix. 

https://github.com/etcd-io/etcd/issues/17247 was fixed by reverting https://github.com/etcd-io/etcd/pull/17228. But the underlying problem with the buffer should still be fixed. 

No failure locally with 
```
EXPECT_DEBUG=true GO_TEST_FLAGS='-run=TestRobustness/Kubernetes/HighTraffic/ClusterOfSize3 --timeout=3000m --count=100 --failfast -v' RESULTS_DIR=./tmp/results make test-robustness
```
and just the `[RaftAfterSaveSnapPanic, MemberReplace]` Failpoints.

### DIFF
--- a/server/storage/backend/batch_tx_test.go
+++ b/server/storage/backend/batch_tx_test.go
@@ -243,24 +243,118 @@ func TestRangeAfterDeleteMatch(t *testing.T) {
 	tx.Unlock()
 	tx.Commit()
 
-	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0)
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), nil, 0)
 	checkForEach(t, b.BatchTx(), b.ReadTx(), [][]byte{[]byte("foo")}, [][]byte{[]byte("bar")})
 
 	tx.Lock()
 	tx.UnsafeDelete(schema.Test, []byte("foo"))
 	tx.Unlock()
 
-	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), []byte("foo"), nil, 0)
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), nil, 0)
 	checkForEach(t, b.BatchTx(), b.ReadTx(), nil, nil)
 }
 
-func checkRangeResponseMatch(t *testing.T, tx backend.BatchTx, rtx backend.ReadTx, key, endKey []byte, limit int64) {
+func TestRangeAfterUnorderedKeyWriteMatch(t *testing.T) {
+	b, _ := betesting.NewTmpBackend(t, time.Hour, 10000)
+	defer betesting.Close(t, b)
+
+	tx := b.BatchTx()
+
 	tx.Lock()
-	ks1, vs1 := tx.UnsafeRange(schema.Test, key, endKey, limit)
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafePut(schema.Test, []byte("foo5"), []byte("bar5"))
+	tx.UnsafePut(schema.Test, []byte("foo2"), []byte("bar2"))
+	tx.UnsafePut(schema.Test, []byte("foo1"), []byte("bar1"))
+	tx.UnsafePut(schema.Test, []byte("foo3"), []byte("bar3"))
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar"))
+	tx.UnsafePut(schema.Test, []byte("foo4"), []byte("bar4"))
+	tx.Unlock()
+
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), nil, 1)
+}
+
+func TestRangeAfterAlternatingBucketWriteMatch(t *testing.T) {
+	b, _ := betesting.NewTmpBackend(t, time.Hour, 10000)
+	defer betesting.Close(t, b)
+
+	tx := b.BatchTx()
+
+	tx.Lock()
+	tx.UnsafeCreateBucket(schema.Key)
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafeSeqPut(schema.Key, []byte("key1"), []byte("val1"))
+	tx.Unlock()
+
+	tx.Lock()
+	tx.UnsafeSeqPut(schema.Key, []byte("key2"), []byte("val2"))
+	tx.Unlock()
+	tx.Commit()
+	// only in the 2nd commit the schema.Key key is removed from the readBuffer.buckets.
+	// This makes sure to test the case when an empty writeBuffer.bucket
+	// is used to replace the read buffer bucket.
+	tx.Commit()
+
+	tx.Lock()
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar"))
+	tx.Unlock()
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Key, []byte("key"), []byte("key5"), 100)
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), []byte("foo3"), 1)
+}
+
+func TestRangeAfterOverwriteMatch(t *testing.T) {
+	b, _ := betesting.NewTmpBackend(t, time.Hour, 10000)
+	defer betesting.Close(t, b)
+
+	tx := b.BatchTx()
+
+	tx.Lock()
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar2"))
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar0"))
+	tx.UnsafePut(schema.Test, []byte("foo1"), []byte("bar10"))
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar1"))
+	tx.UnsafePut(schema.Test, []byte("foo1"), []byte("bar11"))
+	tx.Unlock()
+
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), []byte("foo3"), 1)
+	checkForEach(t, b.BatchTx(), b.ReadTx(), [][]byte{[]byte("foo"), []byte("foo1")}, [][]byte{[]byte("bar1"), []byte("bar11")})
+}
+
+func TestRangeAfterOverwriteAndDeleteMatch(t *testing.T) {
+	b, _ := betesting.NewTmpBackend(t, time.Hour, 10000)
+	defer betesting.Close(t, b)
+
+	tx := b.BatchTx()
+
+	tx.Lock()
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar2"))
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar0"))
+	tx.UnsafePut(schema.Test, []byte("foo1"), []byte("bar10"))
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar1"))
+	tx.UnsafePut(schema.Test, []byte("foo1"), []byte("bar11"))
+	tx.Unlock()
+
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), nil, 0)
+	checkForEach(t, b.BatchTx(), b.ReadTx(), [][]byte{[]byte("foo"), []byte("foo1")}, [][]byte{[]byte("bar1"), []byte("bar11")})
+
+	tx.Lock()
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar3"))
+	tx.UnsafeDelete(schema.Test, []byte("foo1"))
+	tx.Unlock()
+
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), nil, 0)
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo1"), nil, 0)
+	checkForEach(t, b.BatchTx(), b.ReadTx(), [][]byte{[]byte("foo")}, [][]byte{[]byte("bar3")})
+}
+
+func checkRangeResponseMatch(t *testing.T, tx backend.BatchTx, rtx backend.ReadTx, bucket backend.Bucket, key, endKey []byte, limit int64) {
+	tx.Lock()
+	ks1, vs1 := tx.UnsafeRange(bucket, key, endKey, limit)
 	tx.Unlock()
 
 	rtx.RLock()
-	ks2, vs2 := rtx.UnsafeRange(schema.Test, key, endKey, limit)
+	ks2, vs2 := rtx.UnsafeRange(bucket, key, endKey, limit)
 	rtx.RUnlock()
 
 	if diff := cmp.Diff(ks1, ks2); diff != "" {
--- a/server/storage/backend/batch_tx_test.go
+++ b/server/storage/backend/batch_tx_test.go
@@ -15,6 +15,8 @@
 package backend_test
 
 import (
+	"fmt"
+	"math/rand"
 	"reflect"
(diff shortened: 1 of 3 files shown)


fuweid:
Hi @siyuanfoundation you can try to limit cpu by the following command and use patch #17248

```
EXPECT_DEBUG=true GO_TEST_FLAGS=${GO_TEST_FLAGS} RESULTS_DIR=/tmp/results taskset -c 0,1,2 make test-robustness
```

Hope it can help.

siyuanfoundation:
> Hi @siyuanfoundation you can try to limit cpu by the following command and use patch #17248
> 
> ```
> EXPECT_DEBUG=true GO_TEST_FLAGS=${GO_TEST_FLAGS} RESULTS_DIR=/tmp/results taskset -c 0,1,2 make test-robustness
> ```
> 
> Hope it can help.

Thank you @fuweid, it did help!

siyuanfoundation:
cc @serathius @ahrtr @fuweid

server/storage/backend/tx_buffer.go
```diff
@@ -82,8 +82,10 @@ func (txw *txWriteBuffer) writeback(txr *txReadBuffer) {
 		rb, ok := txr.buckets[k]
 		if !ok {
 			delete(txw.buckets, k)
+			if seq, ok := txw.bucket2seq[k]; ok && !seq && wb.used > 1 {
+				wb.dedupe()
```
fuweid, line 86: If buffer.used is zero, after wb.dedupe, the `used` will be 1.
For `key` bucket, as far as I know, the put is seqPut. Not sure why it impacts the events.

And it seems `UnsafeRange` can return duplicate key/values. Not sure that it's related.

fuweid: I was thinking that


1. After `txWriterBuffer.reset()`, `txWriterBuffer.used` is zero. Let say the `txWriterBuffer.buf[0].key` = `{main: 910, sub: 0}`.
2. Backend commits, resets  `txReaderBuffer` and removes `txReaderBuffer.buf[`Key`]`. 
3. Next round write txn, there is no key/value commit, buf `txWriterBuffer` updates `txReaderBuffer.buf[`Key`]` with one item `{main: 910, sub: 0}`. However, the `{main:910, sub:0}` should be read from bbolt instead of cache.
4. WatchStore unsafe-ranges `txReaderBuffer.buf[`Key`]` and bbolt. It will return two same events `{main:910, sub:0}`.

So, I think we should change `dedupe` function as well. If `used` == 0, no-op.

siyuanfoundation: @fuweid This is great insight! And it did fix the failure on my local machine. 
But I still don't understand how it got into the case of `dedupe()` with `bb.used=0`. The scenario you described does not result in `dedupe()` with `bb.used=0` as in the test case `TestRangeAfterAlternatingBucketWriteMatch`. 

1. After `txWriterBuffer.reset()`, `txWriterBuffer.used=0`. Let say the `txWriterBuffer.buf[0].key = {main: 910, sub: 0}`, at this point `txReaderBuffer.used=1`
1. Backend commits, resets txReaderBuffer does not remove `txReaderBuffer.buf[Key]`, because `txReaderBuffer.used !=0`, it would just set `txReaderBufferr.buf[Key].used=0`. https://github.com/etcd-io/etcd/blob/e3c70c8a99401a417c7eb4ba4d805d6038b4707a/server/storage/backend/tx_buffer.go#L33C3-L33C3
1. Next round write txn, there is no key/value commit,  txReaderBuffer.buf[Key] and txWriterBuffer.buf[Key] both exist with `used=0`, which would not enter `dedupe`

The only place I have seen `buf.used=0` in the robustness test is around snapshot
```
{"level":"info","ts":"2024-01-17T19:38:29.573482Z","caller":"rafthttp/snapshot_sender.go:83","msg":"sending database snapshot","snapshot-index":1950,"remote-peer-id":"ac4ec652f10e5b49","bytes":177288,"size":"177 kB"}
sizhangDebug: bucket[1] buffer dedupe with bb.used = 0
{"level":"info","ts":"2024-01-17T19:38:29.585224Z","caller":"etcdserver/snapshot_merge.go:64","msg":"sent database snapshot to writer","bytes":167936,"size":"168 kB"}

```

server/storage/backend/batch_tx_test.go
```diff
@@ -241,17 +241,61 @@ func TestRangeAfterDeleteMatch(t *testing.T) {
...
+	tx.UnsafeCreateBucket(schema.Key)
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafeSeqPut(schema.Key, []byte("key1"), []byte("val1"))
+	tx.Unlock()
+
+	tx.Lock()
+	tx.UnsafeSeqPut(schema.Key, []byte("key2"), []byte("val2"))
+	tx.Unlock()
+	tx.Commit()
+	tx.Commit()
```
fuweid, line 290: Would you mind adding some comments for second commit? It's key to reproduce the issue I think.

fuweid (approved):
LGTM!

server/storage/backend/batch_tx_test.go
siyuanfoundation: done

ahrtr:
Please read https://github.com/etcd-io/etcd/issues/17247#issuecomment-1895725142

ahrtr (changes requested):
https://github.com/etcd-io/etcd/pull/17263#issuecomment-1898957545

siyuanfoundation:
cc @ahrtr @serathius

server/storage/backend/batch_tx_test.go
```diff
@@ -241,24 +241,118 @@ func TestRangeAfterDeleteMatch(t *testing.T) {
...
+	tx.UnsafeCreateBucket(schema.Test)
+	tx.UnsafePut(schema.Test, []byte("foo5"), []byte("bar5"))
+	tx.UnsafePut(schema.Test, []byte("foo2"), []byte("bar2"))
+	tx.UnsafePut(schema.Test, []byte("foo1"), []byte("bar1"))
+	tx.UnsafePut(schema.Test, []byte("foo3"), []byte("bar3"))
+	tx.UnsafePut(schema.Test, []byte("foo"), []byte("bar"))
+	tx.UnsafePut(schema.Test, []byte("foo4"), []byte("bar4"))
+	tx.Unlock()
+
+	checkRangeResponseMatch(t, b.BatchTx(), b.ReadTx(), schema.Test, []byte("foo"), nil, 1)
```
serathius, line 271: Double checking if I understood it correctly. This will only read key "foo", and not do prefix on "foo*"?

siyuanfoundation: yes, This will only read key "foo".

serathius:
The change looks good to me, it fixes a clear problem and adds regression tests, however https://github.com/etcd-io/etcd/pull/17228 did the same. Can we take some additional steps to ensure we prevent another regression?

I don't we can depend on reviewers trying to analyse and find edge cases as it failed before. One idea I have is to do comparison testing between buffered and unbuffered transaction. What do you think?

siyuanfoundation:
Added benchmark test for `writeback` function
Ran the test 3 times. The results are kind of noisy, but there is no clear indication of performance regression. 
Before the PR:
```
BenchmarkWritebackSeqBatches1BatchSize10000-8      	     100	 132904368, 137525118, 134134904  ns/op
BenchmarkWritebackSeqBatches10BatchSize1000-8      	     100	 135975057, 144034473, 138393072 ns/op
BenchmarkWritebackSeqBatches100BatchSize100-8      	     100	 145159646, 147396648, 148373301 ns/op
BenchmarkWritebackSeqBatches1000BatchSize10-8      	     100	 225271045, 225629845, 229453516 ns/op
BenchmarkWritebackNonSeqBatches1000BatchSize1-8    	     100	  29684442, 29051643, 31323272 ns/op
BenchmarkWritebackNonSeqBatches10000BatchSize1-8   	     100	1011180750, 1012368726, 1013996570 ns/op
BenchmarkWritebackNonSeqBatches100BatchSize10-8    	     100	  22771697, 22785210, 22529997 ns/op
BenchmarkWritebackNonSeqBatches1000BatchSize10-8   	     100	 233131758, 233625789, 238527133 ns/op
```
After the PR:
```
enchmarkWritebackSeqBatches1BatchSize10000-8      	     100	 136669713, 131847663, 130722632 ns/op
BenchmarkWritebackSeqBatches10BatchSize1000-8      	     100	 139880124, 139196486, 137805060 ns/op
BenchmarkWritebackSeqBatches100BatchSize100-8      	     100	 150674477, 146649422, 146416992 ns/op
BenchmarkWritebackSeqBatches1000BatchSize10-8      	     100	 226596374, 228097908, 223997282 ns/op
BenchmarkWritebackNonSeqBatches1000BatchSize1-8    	     100	  30098653, 30031651, 29341949 ns/op
BenchmarkWritebackNonSeqBatches10000BatchSize1-8   	     100	 990611851, 999573564, 994189677 ns/op
BenchmarkWritebackNonSeqBatches100BatchSize10-8    	     100	  22655668, 22641929, 21832902 ns/op
BenchmarkWritebackNonSeqBatches1000BatchSize10-8   	     100	 234623372, 233671979, 231825643 ns/op
```

siyuanfoundation:
> The change looks good to me, it fixes a clear problem and adds regression tests, however #17228 did the same. Can we take some additional steps to ensure we prevent another regression?
> 
> I don't we can depend on reviewers trying to analyse and find edge cases as it failed before. One idea I have is to do comparison testing between buffered and unbuffered transaction. What do you think?

This bug happened because of the coincidence of `Commit()` timing between two different buckets, like in the tests I had to do two consecutive `tx.Commit()` to reproduce the bug. It is very hard to ensure there is no regression deterministically. In this case, we just have to be careful with more unit tests and rely on non-deterministic robustness test to detect the regression.

siyuanfoundation:
> > Ran the test 3 times. The results are kind of noisy, but there is no clear indication of performance regression.
> > Before the PR:
> 
> Can you use benchstat ? https://sourcegraph.com/blog/go/gophercon-2019-optimizing-go-code-without-a-blindfold

Ok. I ran
```
go clean -testcache && go test -bench=. -count=10 -timeout 0 > benchmark_results.txt
benchstat benchmark_results.txt
```
Here are the results: (they are basically the same)
**Before**
```
WritebackSeqBatches1BatchSize10000-8               128.2m ± 1%
WritebackSeqBatches10BatchSize1000-8               130.7m ± 1%
WritebackSeqBatches100BatchSize100-8               139.5m ± 1%
WritebackSeqBatches1000BatchSize10-8               220.6m ± 4%
WritebackNonSeqBatches1000BatchSize1-8             28.34m ± 2%
WritebackNonSeqBatches10000BatchSize1-8             1.002 ± 1%
WritebackNonSeqBatches100BatchSize10-8             21.64m ± 2%
WritebackNonSeqBatches1000BatchSize10-8            226.3m ± 1%
```
**After**
```
WritebackSeqBatches1BatchSize10000-8              128.3m ± 1%
WritebackSeqBatches10BatchSize1000-8              130.8m ± 3%
WritebackSeqBatches100BatchSize100-8              140.0m ± 1%
WritebackSeqBatches1000BatchSize10-8              216.3m ± 4%
WritebackNonSeqBatches1000BatchSize1-8            27.49m ± 2%
WritebackNonSeqBatches10000BatchSize1-8           966.6m ± 1%
WritebackNonSeqBatches100BatchSize10-8            20.47m ± 2%
WritebackNonSeqBatches1000BatchSize10-8           221.3m ± 1%
```

ahrtr (approved):
/lgtm

Thanks

serathius:
> Here are the results: (they are basically the same)

Benchstat do not change the results, it just makes them readable and makes it easy to check if they are [statistically significant ](https://en.wikipedia.org/wiki/Statistical_significance)

If you provide both old and new benchmark results to benchstat, it will provide also the performance delta and [p-value](https://en.wikipedia.org/wiki/P-value)

server/storage/backend/batch_tx_test.go
```diff
@@ -294,3 +296,77 @@ func checkUnsafeForEach(t *testing.T, tx backend.UnsafeReader, expectedKeys, exp
...
+		var ks, vs []string
+		for j := i * batchSize; j < (i+1)*batchSize; j++ {
+			k := fmt.Sprintf("key%d", j)
+			v := fmt.Sprintf("val%d", j)
+			ks = append(ks, k)
+			vs = append(vs, v)
+		}
+		if !isSeq {
+			shuffleList(ks)
+		}
```
serathius, line 417: Nit: When benchmark needs some input data, it's best to avoid not include it in the measurements of the benchmark to avoid noise.

You can do that by preparing the input data first, and then calling `b.ResetTimer()` before running the code you want to benchmark.

server/storage/backend/batch_tx_test.go
serathius, line 1: It's great to have a regression tests, however I'm thinking that the scenarios we are covering are so intricate and complicated that it scares me. If there is another issue hiding in the transaction logic, we don't have any chance of finding it using this approach.

serathius (approved):

fuweid (approved):
LGTM

Thanks for adding benchmark test suites.

server/storage/backend/batch_tx_test.go
siyuanfoundation: Done.

siyuanfoundation:
> > Here are the results: (they are basically the same)
> 
> Benchstat do not change the results, it just makes them readable and makes it easy to check if they are [statistically significant ](https://en.wikipedia.org/wiki/Statistical_significance)
> 
> If you provide both old and new benchmark results to benchstat, it will provide also the performance delta and [p-value](https://en.wikipedia.org/wiki/P-value)

Reran benchstat twice with count=10,
run 1: 
```
                                        │ benchmark_before.txt │        benchmark_after.txt         │
                                        │        sec/op        │   sec/op     vs base               │
WritebackSeqBatches1BatchSize10000-8               127.7m ± 1%   127.2m ± 3%       ~ (p=0.796 n=10)
WritebackSeqBatches10BatchSize1000-8               129.9m ± 6%   129.9m ± 1%       ~ (p=0.971 n=10)
WritebackSeqBatches100BatchSize100-8               138.6m ± 1%   137.0m ± 1%  -1.17% (p=0.029 n=10)
WritebackSeqBatches1000BatchSize10-8               215.7m ± 1%   212.7m ± 4%       ~ (p=0.052 n=10)
WritebackNonSeqBatches1000BatchSize1-8             28.48m ± 2%   28.05m ± 2%  -1.48% (p=0.035 n=10)
WritebackNonSeqBatches10000BatchSize1-8            981.0m ± 1%   949.6m ± 0%  -3.20% (p=0.000 n=10)
WritebackNonSeqBatches100BatchSize10-8             20.34m ± 1%   21.21m ± 5%  +4.27% (p=0.000 n=10)
WritebackNonSeqBatches1000BatchSize10-8            216.3m ± 1%   216.4m ± 4%       ~ (p=0.796 n=10)
```
run 2:
```
                                        │ benchmark_before_0.txt │       benchmark_after_0.txt        │
                                        │         sec/op         │   sec/op     vs base               │
WritebackSeqBatches1BatchSize10000-8                 127.8m ± 3%   130.0m ± 2%  +1.73% (p=0.023 n=10)
WritebackSeqBatches10BatchSize1000-8                 127.2m ± 1%   130.9m ± 3%  +2.95% (p=0.001 n=10)
WritebackSeqBatches100BatchSize100-8                 135.4m ± 1%   139.2m ± 3%  +2.76% (p=0.001 n=10)
WritebackSeqBatches1000BatchSize10-8                 212.4m ± 2%   213.1m ± 0%       ~ (p=0.165 n=10)
WritebackNonSeqBatches1000BatchSize1-8               27.82m ± 2%   28.80m ± 3%  +3.51% (p=0.001 n=10)
WritebackNonSeqBatches10000BatchSize1-8              970.1m ± 2%   951.8m ± 1%  -1.89% (p=0.000 n=10)
WritebackNonSeqBatches100BatchSize10-8               20.23m ± 1%   20.39m ± 5%       ~ (p=0.165 n=10)
WritebackNonSeqBatches1000BatchSize10-8              217.1m ± 1%   216.6m ± 1%       ~ (p=0.529 n=10)
```

serathius:
So in the updated benchmark results you can see that it causes ~2% regression in performance. Still I think performance impact on overall etcd performance should be not noticable and we should pick correctness here.

chaochn47:
Any plan back-porting this correctness bug fix to release branches? @ahrtr @serathius @fuweid

cc @shyamjvs

PR #17280, opened 2024-01-19
Revert https://github.com/etcd-io/etcd/pull/17228

ahrtr:
Actually https://github.com/etcd-io/etcd/pull/17228 was fixing a non-exist issue, because there will never have duplicated keys. Please refer to
https://github.com/etcd-io/etcd/issues/17247#issuecomment-1895725142.

Let alone the PR (pull/17228) also introduces a bug (issues/17247).


ahrtr:
See the comment in the second commit in https://github.com/etcd-io/etcd/pull/17281.

The Meta bucket is the only case which may have duplicated keys in the buffer.

ahrtr:
Resolved in https://github.com/etcd-io/etcd/pull/17285

PR #17281, opened 2024-01-19
No need remove duplicate keys for the Key bucket and add a verification

ahrtr:
This PR is on top of https://github.com/etcd-io/etcd/pull/17280, which is where the first commit coming from.

So please only read the second commit. Since there will never be any duplicated keys in the bucket buffer, so there is no need to remove duplicated keys. Also added a verification to double check it. Note that it's an **invariant property** for etcd. 

Please also read https://github.com/etcd-io/etcd/issues/17247#issuecomment-1895725142

### DIFF
--- a/server/storage/backend/batch_tx.go
+++ b/server/storage/backend/batch_tx.go
@@ -28,6 +28,23 @@ import (
 
 type BucketID int
 
+const (
+	BucketIdKey     BucketID = 1
+	BucketIdMeta    BucketID = 2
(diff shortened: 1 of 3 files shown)


ahrtr:
Just rebased this PR, and refined the verification, which ensures that all the keys in the Key bucket are **monotonically increasing**.

It doesn't sort the keys in the Key bucket, nor to perform the operation of removing duplicated keys. So it should have a little better performance. Need to run benchmark to double confirm this.

PR #17284, opened 2024-01-19
Fix test issue in TestScheduleCompaction

ahrtr:
It always reuse the same memory space. Each next iteration will always overwrite previous data.

It took me about 2 hours to identify the root cause why there are duplicated keys in the buffer.


ahrtr:
The issue was discovered in https://github.com/etcd-io/etcd/pull/17281

fuweid (approved):
LGTM

PR #17286, opened 2024-01-19
Fix golangci revive error-strings issues

ivanvc:
I was looking at the disabled linter rules and addressed the `error-strings` warnings. The only offending files are in the deprecated v2 client test code.

After speaking with @jmhbnz, he told me that @fuweid was working recently with the linter rules. So, it would be great if you could take a look at this.


fuweid (approved):
LGTM

jmhbnz (approved):
LGTM - Thanks @ivanvc

ahrtr (approved):
lgtm

Issue #17121, opened 2023-12-14
Add memberReplaceWithLearner e2e test and scenario to robustness tests


serathius:
### What would you like to be added?

same as test ReplaceMember but replace it with learner and then promote it after some time 

### Why is this needed?

Improve testing coverage of membership information

nitishfy:
Hi, I'd like to take this up. Please assign this to me!

jmhbnz:
Thanks @nitishfy, assigned. I would start by taking a look at the existing ReplaceMember tests and feel free to ask any questions here as you go.

nitishfy:
**Update:** I've started working on this issue.

PR #17287, opened 2024-01-20
feat: add `memberReplaceLearnerTest` test 

nitishfy:
Fix: #17121

- Add `memberReplaceLearnerTest` to replace the member with Learner and later promote that member.


tests/e2e/ctl_v3_member_no_proxy_learner_test.go
```diff
@@ -0,0 +1,100 @@
...
+	require.NoError(t, err)
+	err = patchArgs(member.Config().Args, "initial-cluster-state", "existing")
+	require.NoError(t, err)
+
+	// Sleep 100ms to bypass the known issue https://github.com/etcd-io/etcd/issues/16687.
+	time.Sleep(100 * time.Millisecond)
+
+	// TODO
+	//_, err = cc.MemberPromote(ctx, memberID)
+	//require.NoError(t, err)
```
nitishfy, line 85:
```suggestion
	//require.NoError(t, err)
```

This is to be done yet. `cc.MemberPromote()` is leading to error with the following logs:

```
Received unexpected error: match not found.  Set EXPECT_DEBUG for more info Errs: [unexpected exit code [1] after running [/home/nitish/Documents/Work/Open-Source/etcd/bin/etcdctl --endpoints=http://localhost:20000,http://localhost:20005 member promote 72e86eb5e356a737 -w json]], last lines:
        	            	{"level":"warn","ts":"2024-01-20T18:57:56.348266+0530","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0001fc780/localhost:20000","method":"/etcdserverpb.Cluster/MemberPromote","attempt":0,"error":"rpc error: code = NotFound desc = etcdserver: member not found"}
        	            	Error: etcdserver: member not found
```

nitishfy (commented):
A small information described above ^^

tests/e2e/ctl_v3_member_no_proxy_learner_test.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2023 The etcd Authors
```
serathius, line 1: No need for a separate file, we can add this test to `ctl_v3_member_no_proxy_test.go`

tests/e2e/ctl_v3_member_no_proxy_learner_test.go
serathius: Also promote should be called after we started member on line 87-99.

tests/e2e/ctl_v3_member_no_proxy_learner_test.go
nitishfy: Gotcha

nitishfy:
**Update:** Fixed it, so we need to read the new ID for the learner once it has started running. I tried printing logs before running it and found that the value of the ID was 0. Due to this, it was unable to get the new member added as a learner. 

cc: @serathius

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -97,3 +97,79 @@ func TestMemberReplace(t *testing.T) {
...
+				time.Sleep(10 * time.Millisecond)
+				continue
+			}
+			break
+		}
+	})
+
+	learnMemberID, found, err := getMemberIdByName(ctx, cc, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, true, "Member not found")
```
serathius, line 171:
```suggestion
    var learnMemberID uint64
	testutils.ExecuteUntil(ctx, t, func() {
	    var found bool
	    var err error
		for {
			learnMemberID, found, err = getMemberIdByName(ctx, cc, memberName)
			if err != nil || !found {
				time.Sleep(10 * time.Millisecond)
				continue
			}
			break
		}
	})
```

serathius:
> Update: Fixed it, so we need to read the new ID for the learner once it has started running. I tried printing logs before running it and found that the value of the ID was 0. Due to this, it was unable to get the new member added as a learner.

Hmm, it's not something I would expect. Might need to investigate. cc @ahrtr

serathius:
Works on my side. `MemberAddAsLearner` returns `Member{ID:9082546050735520123 peerURLs:"http://localhost:20006" isLearner:true}`

serathius:
From static analysis 
```
Error: stderr: e2e/ctl_v3_member_no_proxy_test.go:160:7: declaration of "found" shadows declaration at line 120
Error: stderr: e2e/ctl_v3_member_no_proxy_test.go:161:7: declaration of "err" shadows declaration at line 106
```

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -97,3 +97,82 @@ func TestMemberReplace(t *testing.T) {
...
+	// Sleep 100ms to bypass the known issue https://github.com/etcd-io/etcd/issues/16687.
+	time.Sleep(100 * time.Millisecond)
+
+	t.Logf("Starting member %s", memberName)
+	err = member.Start(ctx)
+	require.NoError(t, err)
+	var learnMemberID uint64
+	testutils.ExecuteUntil(ctx, t, func() {
+		var found bool
+		var err error
```
serathius, line 161:
```suggestion
```

serathius (approved):

serathius:
cc @ahrtr

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -97,3 +97,80 @@ func TestMemberReplace(t *testing.T) {
...
+	var endpoints []string
+	for i := 1; i < len(epc.Procs); i++ {
+		endpoints = append(endpoints, epc.Procs[(memberIdx+i)%len(epc.Procs)].EndpointsGRPC()...)
+	}
+	cc, err := e2e.NewEtcdctl(epc.Cfg.Client, endpoints)
+	require.NoError(t, err)
+
+	memberID, found, err := getMemberIdByName(ctx, cc, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, true, "Member not found")
```
ahrtr, line 122: nit: Usually the second parameter is expected value, and the third one is actual value.
```suggestion
	require.Equal(t, true, found, "Member not found")
```

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -97,3 +97,80 @@ func TestMemberReplace(t *testing.T) {
...
+
+	// Need to wait health interval for cluster to accept member changes
+	time.Sleep(etcdserver.HealthInterval)
+
+	t.Logf("Removing member %s", memberName)
+	_, err = cc.MemberRemove(ctx, memberID)
+	require.NoError(t, err)
+	_, found, err = getMemberIdByName(ctx, cc, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, false, "Expected member to be removed")
```
ahrtr, line 132: ditto
```suggestion
	require.Equal(t, false, found, "Expected member to be removed")
```

tests/e2e/ctl_v3_member_no_proxy_test.go
```diff
@@ -97,3 +97,80 @@ func TestMemberReplace(t *testing.T) {
...
+				time.Sleep(10 * time.Millisecond)
+				continue
+			}
+			break
+		}
+	})
+
+	learnMemberID, found, err = getMemberIdByName(ctx, cc, memberName)
+	require.NoError(t, err)
+	require.Equal(t, found, true, "Member not found")
```
ahrtr, line 172: ditto
```suggestion
	require.Equal(t, true, found, "Member not found")
```

PR #17288, opened 2024-01-21
Added a error log when learner is not sync with etcd leader.

ishan16696:

Fix: https://github.com/kubernetes/kubeadm/issues/2997#issuecomment-1900069112

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -102,7 +102,9 @@ const (
 
 	recommendedMaxRequestBytes = 10 * 1024 * 1024
 
-	readyPercent = 0.9
+	// readyPercentThreshold is a threshold used to determine
+	// whether a learner is ready for a transition into a full voting member or not.
+	readyPercentThreshold = 0.9
 
 	DowngradeEnabledPath = "/downgrade/enabled"
 )
@@ -1491,8 +1493,7 @@ func (s *EtcdServer) promoteMember(ctx context.Context, id uint64) ([]*membershi
 
 func (s *EtcdServer) mayPromoteMember(id types.ID) error {
 	lg := s.Logger()
-	err := s.isLearnerReady(uint64(id))
-	if err != nil {
+	if err := s.isLearnerReady(lg, uint64(id)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


server/etcdserver/server.go
```diff
@@ -1548,6 +1548,10 @@ func (s *EtcdServer) isLearnerReady(id uint64) error {
 	leaderMatch := rs.Progress[leaderID].Match
 	// the learner's Match not caught up with leader yet
 	if float64(learnerMatch) < float64(leaderMatch)*readyPercent {
+		lg.Error(
+			"rejecting promote learner: learner is not ready",
+			zap.String("learner-ready-percent", fmt.Sprint(learnerMatch/leaderMatch*100)),
+		)
```
serathius, line 1554:
```suggestion
	learnerProgressPercent := float64(learnerMatch) / float64(leaderMatch)
	if learnerProgressPercent < readyPercent {
		lg.Error(
			"rejecting promote learner: learner is not ready",
			zap.Float64("learner-progress-percent", learnerProgressPercent),
			zap.Float64("learner-ready-percent", readyPercent),
		)
```

`learnerMatch/leaderMatch*100` without conversion to float64 will just be equal `0`.

ishan16696: > learnerMatch/leaderMatch*100 without conversion to float64 will just be equal 0.

yes, make sense. I missed that

ishan16696: > zap.Float64("learner-ready-percent", readyPercent),

this suggestion seems wrong to me as `readyPercent` is not a `learner-ready-percent` 🤔  but `readyPercent` is just a threshold to decide whether to allow a learner to be added or not in cluster.

serathius: Right, feel free to rename to make it clearer. For example rename `readyPercent` to `readyPercentThreshold` and  add field to log `ready-percent-threshold`.

I think it would be great to return the threshold in the log, so user doesn't need to read the etcd code to find it.

ishan16696: done

server/etcdserver/server.go
```diff
@@ -1545,9 +1546,20 @@ func (s *EtcdServer) isLearnerReady(id uint64) error {
 		return membership.ErrIDNotFound
 	}
 
+	var learnerReadyPercent float64
 	leaderMatch := rs.Progress[leaderID].Match
+
+	if leaderMatch > 0 {
```
serathius, line 1552: Why do you add this new condition? It will be much easier to review the PR if we limit the change to just adding a log.

ishan16696: I thought it will be better to have a safety for `leaderMatch`
```go 
learnerReadyPercent = float64(learnerMatch) / float64(leaderMatch)
```
If doesn't make sense then I can revert it back.

serathius (approved):

serathius:
cc @ahrtr

ahrtr (approved):
lgtm

Thank you!

pacoxu:
Thank you so much.

PR #17290, opened 2024-01-22
Add verification on keys: should be always moronically increasing

ahrtr:
Second solution to https://github.com/etcd-io/etcd/issues/17247.

This solution is much simpler. But it only verifies the write buffer. It doesn't verify the merged data (write and read buffer).

### DIFF
--- a/server/storage/backend/tx_buffer.go
+++ b/server/storage/backend/tx_buffer.go
@@ -16,6 +16,8 @@ package backend
 
 import (
 	"bytes"
+	"encoding/hex"
+	"fmt"
 	"sort"
 
 	"go.etcd.io/etcd/client/pkg/v3/verify"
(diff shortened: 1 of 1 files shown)


server/storage/backend/tx_buffer.go
```diff
@@ -52,7 +54,20 @@ func (txw *txWriteBuffer) put(bucket Bucket, k, v []byte) {
 }
 
 func (txw *txWriteBuffer) putSeq(bucket Bucket, k, v []byte) {
-	// TODO: Add (in tests?) verification whether k>b[len(b)]
+	// putSeq is only be called for the data in the Key bucket. The keys
```
serathius, line 57: Could we merge verify independent of the solution?

ahrtr: Raised PR https://github.com/etcd-io/etcd/pull/17325

serathius:
Can this be closed and work continued in https://github.com/etcd-io/etcd/pull/17263 ?

Issue #17241, opened 2024-01-14
[3.4] linux-amd64-e2e job takes more than 30min


fuweid:
### Which github workflows are flaking?

linux-amd64-e2e

### Which tests are flaking?

The e2e testcases need more than 30min but workflow timeout is 30min

### Github Action link

https://github.com/etcd-io/etcd/actions/runs/7505424685/job/20453159962?pr=17238

### Reason for failure (if possible)

Need to investigate e2e runtime.

ahrtr:
Recently we see lots of workflow timeout.

Could anyone take a look at this?

ahrtr:
It seems that it always fails when running the following two cases.
```
	TestWatchDelayForEvent (3s)
	TestWatchDelayForEvent/NoTLS (3s)
```

ahrtr:
Resolved in https://github.com/etcd-io/etcd/pull/17304

PR #17173, opened 2023-12-28
[3.4] member replace e2e test

ZhouJianMS:
Follow up: https://github.com/etcd-io/etcd/pull/17079#issuecomment-1856089798

PR #17144, opened 2023-12-18
[3.4] Backport e2e tests for livez/readyz. 

siyuanfoundation:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Similar to https://github.com/etcd-io/etcd/pull/17083

PR #16997, opened 2023-11-22
Release 3.4 upgrade grpc 1.52.0

chaochn47:
Part of https://github.com/etcd-io/etcd/issues/16740

Backported https://github.com/etcd-io/etcd/pull/14834 and https://github.com/etcd-io/etcd/pull/16324 just like the 3.5 one https://github.com/etcd-io/etcd/pull/16781

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #17304, opened 2024-01-23
[3.4] fix: linux-amd64-e2e job timeout issue

fuweid:
Assume #16916 as baseline. The E2E takes `1395.082s`.

* https://github.com/etcd-io/etcd/pull/16988

It introduced `TestAuthority` which takes `18.39s`. And after https://github.com/etcd-io/etcd/pull/16997, it takes `50.05s`.

* https://github.com/etcd-io/etcd/pull/16995

It introduced `TestInPlaceRecovery` which takes `17.37s`.

* https://github.com/etcd-io/etcd/pull/17144

  - New `TestHTTPHealthHandler` takes `29.9s`
  - New `TestHTTPLivezReadyzHandler` takes `35.20s`

* https://github.com/etcd-io/etcd/pull/17173

  - New `TestMemberReplace` takes `7.55s`.

Ideally, it should increase `140.07s`. It's not larger than `1800s` timeout value.

However, we run E2E cases 3 times. By default, we run E2E cases with `-cpu 1,2,4`. That means that we run 3 times.

```bash
$ go help testflag

 -count n
            Run each test, benchmark, and fuzz seed n times (default 1).

            If -cpu is set, run n times for each GOMAXPROCS value.
            Examples are always run once. -count does not apply to
            fuzz tests matched by -fuzz.
```

I don't think we should run E2E with different GOMAXPROCS value. All the `TestXYZ` are used to control etcd process and we don't set GOMAXPROCS env to etcd process. So, we don't need `-cpu` setting for E2E.

Closes: #17241


fuweid:
ping @ahrtr @serathius @jmhbnz

ahrtr:
Thanks for the analysis.

I suggest one of the following solutions,
- change the default value for `TEST_CPUS` to `"1,4"`.   https://github.com/etcd-io/etcd/blob/d441d74c0abdbee6daa1e8ca5c114571f7d2762a/test#L120
- or breakdown `linux-amd64-e2e` into two targets: `linux-amd64-e2e-1-cpu` and `linux-amd64-e2e-4-cpu`.
- [**preferred**] or just add an option `CPU=4` for the target `linux-amd64-e2e` in `tests.yaml` (consistent with the `main`` and release-3.5` branches)

ahrtr (approved):
lgtm

Thanks

jmhbnz (approved):
LGTM - Thanks for digging into this @fuweid

PR #17306, opened 2024-01-23
chore: Enable lint var name.

liangyuanpeng:


ivanvc:
Hi @liangyuanpeng, I see that you started working on addressing the warnings from this linter. However, the work required is rather large (I have addressed all of them, but after discussing this with the team on March 7th's community meeting, they decided to split addressing this into subtasks, which I'm already doing) and tracked in #17578.

However, would you be interested in working on enabling `unexported-return`?

liangyuanpeng:
@ivanvc  I'm not focusing on that right now, thank you for working on this and explain it.

Issue #17269, opened 2024-01-18
Bump golang version to 1.20.13 and 1.21.6


ahrtr:
### What would you like to be added?

- Bump golang version to 1.20.13 for 3.5 and 3.4
- Bump golang version to 1.21.6 for main

### Why is this needed?

To improve stability

jmhbnz:
Completion tracking below:

- [ ] `main`  
- [ ] `release-3.5`  (1.20.13) 
- [ ] `release-3.4`  (1.20.13) 
- [ ]  etcd-io `CHANGELOG` 
- [ ] `etcd-io/bbolt` `main`  
- [ ] `etcd-io/raft` `main`

ivanvc:
Can you assign this to me? Thanks

ahrtr:
All done. Thanks @ivanvc and @jmhbnz

PR #17309, opened 2024-01-23
changelog: add 3.4 and 3.5 note about go 1.20.13

ivanvc:
Updates 3.4 and 3.5 CHANGELOGs to add a line that the binaries are compiled with go version 1.20.13.

Related to #17269

### DIFF
--- a/CHANGELOG/CHANGELOG-3.5.md
+++ b/CHANGELOG/CHANGELOG-3.5.md
@@ -9,6 +9,9 @@ Previous change logs can be found at [CHANGELOG-3.4](https://github.com/etcd-io/
 ### etcd server
 - [Add livez/readyz HTTP endpoints](https://github.com/etcd-io/etcd/pull/17039)
 
+### Dependencies
+- Compile binaries using [go 1.20.13](https://github.com/etcd-io/etcd/pull/17275)
+
 ## v3.5.11 (2023-12-07)
 
 ### etcd server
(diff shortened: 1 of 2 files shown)


CHANGELOG/CHANGELOG-3.5.md
```diff
@@ -16,7 +16,7 @@ Previous change logs can be found at [CHANGELOG-3.4](https://github.com/etcd-io/
 - Fix [url redirects while checking peer urls during new member addition](https://github.com/etcd-io/etcd/pull/16986)
 
 ### Dependencies
-- Compile binaries using [go 1.20.12](https://github.com/etcd-io/etcd/pull/17077)
+- Compile binaries using [go 1.20.13](https://github.com/etcd-io/etcd/pull/17275)
```
jmhbnz, line 19: `3.5.11` has already been released so we need to leave this line as is and make a new entry for `### Dependencies` under `3.5.12`.

jmhbnz (commented):
Thanks for completing the upgrade @ivanvc - just a small tweak needed for 3.5 changelog then this will be good to merge.

CHANGELOG/CHANGELOG-3.5.md
serathius: Don't think we need separate lines each dependency update, at least historically we only kept the latest update of one kind.

ahrtr (approved):
lgtm

Thanks

serathius (approved):

PR #8263, opened 2017-07-14
api: hash by rev

fanminshi:
fixes https://github.com/coreos/etcd/issues/8016

PR #17313, opened 2024-01-24
test: fix `TestHashKVWhenCompacting`: ensure all goroutine finished

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/17232

- We should ensure all goroutines finished before exiting the test case
- Compaction 100 times may take longer than 10s, so reduce the count from 100 to 50;

### DIFF
--- a/server/storage/mvcc/kvstore_test.go
+++ b/server/storage/mvcc/kvstore_test.go
@@ -562,6 +562,7 @@ func TestHashKVWhenCompacting(t *testing.T) {
 	hashCompactc := make(chan hashKVResult, 1)
 	var wg sync.WaitGroup
 	donec := make(chan struct{})
+	stopc := make(chan struct{})
 
 	// Call HashByRev(10000) in multiple goroutines until donec is closed
 	for i := 0; i < 10; i++ {
@@ -574,6 +575,8 @@ func TestHashKVWhenCompacting(t *testing.T) {
 					t.Error(err)
 				}
 				select {
+				case <-stopc:
+					return
 				case <-donec:
 					return
 				case hashCompactc <- hashKVResult{hash.Hash, hash.CompactRevision}:
@@ -597,6 +600,8 @@ func TestHashKVWhenCompacting(t *testing.T) {
 				if r.hash != revHash[r.compactRev] {
 					t.Errorf("Hashes differ (current %v) != (saved %v)", r.hash, revHash[r.compactRev])
 				}
+			case <-stopc:
+				return
 			case <-donec:
 				return
 			}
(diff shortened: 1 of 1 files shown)


ahrtr:
Root cause:
When [test timed out](https://github.com/etcd-io/etcd/blob/15f95ec49b90fbfd09bc1315cf71d4c1167ee873/server/storage/mvcc/kvstore_test.go#L625), it fails the test right away, and [cleanup](https://github.com/etcd-io/etcd/blob/15f95ec49b90fbfd09bc1315cf71d4c1167ee873/server/storage/mvcc/kvstore_test.go#L555) will be executed before exiting the test gouroutine, and the backend storage will be closed. The other two gouroutines may not have finished yet, so eventually run into nic porinter panicking.
- https://github.com/etcd-io/etcd/blob/15f95ec49b90fbfd09bc1315cf71d4c1167ee873/server/storage/mvcc/kvstore_test.go#L569
- https://github.com/etcd-io/etcd/blob/15f95ec49b90fbfd09bc1315cf71d4c1167ee873/server/storage/mvcc/kvstore_test.go#L587

server/storage/mvcc/kvstore_test.go
```diff
@@ -603,10 +603,10 @@ func TestHashKVWhenCompacting(t *testing.T) {
 		}
 	}()
 
-	// Compact the store in a goroutine, using RevisionTombstone 9900 to 10000 and close donec when finished
+	// Compact the store in a goroutine, using RevisionTombstone 9950 to 10000 and close donec when finished
 	go func() {
 		defer close(donec)
-		for i := 100; i >= 0; i-- {
+		for i := 50; i >= 0; i-- {
```
jmhbnz, line 609: Curious do we know if there was there any rationale for `100` when this was originally written or was it just an arbitrary number?

I don't see any mention in https://github.com/etcd-io/etcd/commit/8609521ce29fd523d003c1fcb5f943ab9c8247b4 or https://github.com/etcd-io/etcd/pull/8263.

Might just be arbitrary to do that many times but I worry maybe reducing the count is going to mask performance regressions.

jmhbnz (commented):
Ensuring all goroutines finished is a good catch.

Have a question below about halving compaction count.

server/storage/mvcc/kvstore_test.go
ahrtr: It should be just an arbitrary number.

Reverted this change as it's already confirmed that it's an environment issue. https://github.com/etcd-io/etcd/issues/17232#issuecomment-1909437223

server/storage/mvcc/kvstore_test.go
```diff
@@ -620,10 +620,11 @@ func TestHashKVWhenCompacting(t *testing.T) {
 
 	select {
 	case <-donec:
-		wg.Wait()
 	case <-time.After(10 * time.Second):
 		testutil.FatalStack(t, "timeout")
```
fuweid, line 624: If it fatal, L627 is still working?

serathius (dismissed):

server/storage/mvcc/kvstore_test.go
ahrtr: Good catch. Updated.

The key point is to ensure all goroutines to finish before executing the deferred function.

fuweid (approved):
LGTM

ahrtr:
ping @serathius

PR #17314, opened 2024-01-24
Print error log when creating peer listener failed

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/17232

See https://github.com/etcd-io/etcd/issues/17232#issuecomment-1908150321


jmhbnz (approved):
LGTM

PR #13200, opened 2021-07-08
Implement schema migration and panic when trying to downgrade storage

serathius:
Part of milestone 1 of https://github.com/etcd-io/etcd/issues/13168
cc @ptabor

PR #11536, opened 2020-01-17
auth_command: create `etcdctl auth status` command

tarcinil:
This changes have started at etcdctl under auth.go, and make changes to stub out everything down into the internal raft.  Made changes to the .proto files and regenerated them so that the local version would build successfully.

fixes #11516

PR #17330, opened 2024-01-27
[3.4] allow downgrade from 3.5

siyuanfoundation:

Part of https://github.com/etcd-io/etcd/issues/15878#issuecomment-1910886682
Tested downgrade from 3.5 cluster locally:

1. Start 3.5 cluster
```
bin/etcd --name infra1 --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 --listen-peer-urls http://127.0.0.1:12380 --initial-advertise-peer-urls http://127.0.0.1:12380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new --enable-pprof --logger=zap --log-outputs=stderr

bin/etcd --name infra2 --listen-client-urls http://127.0.0.1:22379 --advertise-client-urls http://127.0.0.1:22379 --listen-peer-urls http://127.0.0.1:22380 --initial-advertise-peer-urls http://127.0.0.1:22380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new --enable-pprof --logger=zap --log-outputs=stderr

bin/etcd --name infra3 --listen-client-urls http://127.0.0.1:32379 --advertise-client-urls http://127.0.0.1:32379 --listen-peer-urls http://127.0.0.1:32380 --initial-advertise-peer-urls http://127.0.0.1:32380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new --enable-pprof --logger=zap --log-outputs=stderr
```
2. start putting data 
```
go run tools/benchmark/main.go put --total=10000000 --key-size=64 --val-size=512`

bin/etcdctl --endpoints=http://127.0.0.1:2379,http://127.0.0.1:22379,http://127.0.0.1:32379 endpoint status -w table
bin/etcdctl auth status
```
3. kill `infra3`
4. restart `infra3` with 3.4 binary
```
../etcd-3.4/bin/etcd --name infra3 --listen-client-urls http://127.0.0.1:32379 --advertise-client-urls http://127.0.0.1:32379 --listen-peer-urls http://127.0.0.1:32380 --initial-advertise-peer-urls http://127.0.0.1:32380 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state existing --enable-pprof --logger=zap --log-outputs=stderr --data-dir infra3.etcd --accept-next-storage-version

bin/etcdctl auth status
```
5. kill and restart `infra2` with 3.4 binary
```
../etcd-3.4/bin/etcd --name infra2 --listen-client-urls http://127.0.0.1:22379 --advertise-client-urls http://127.0.0.1:22379 --listen-peer-urls http://127.0.0.1:22380 --initial-advertise-peer-urls http://127.0.0.1:22380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state existing --enable-pprof --logger=zap --log-outputs=stderr --data-dir infra2.etcd --accept-next-storage-version
```
6. kill and restart `infra1` with 3.4 binary
```
../etcd-3.4/bin/etcd --name infra1 --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 --listen-peer-urls http://127.0.0.1:12380 --initial-advertise-peer-urls http://127.0.0.1:12380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state existing --enable-pprof --logger=zap --log-outputs=stderr --data-dir infra1.etcd --accept-next-storage-version
```


siyuanfoundation:
cc @serathius @ahrtr

serathius:
Not sure how this PR refers to dev work described in https://docs.google.com/document/d/1siSo-tA_xLRo5li5wK8KqWsOiq34PIgtmuSDcyxxwsY/edit?usp=sharing

In document you described that downgrade will gated behind "--allow-downgrade" flag and will include implementation of AuthStatusRequest handling.

mvcc/kvstore.go
```diff
@@ -379,6 +379,26 @@ func (s *store) restore() error {
 	tx := s.b.BatchTx()
 	tx.Lock()
 
+	// new keys introduced in 3.5.
```
serathius, line 382: You picked a pretty good place to add logic for handling db file. I think that I even missed in the original downgrade design that only migrates storage when bootstrapping etcd. Added https://github.com/etcd-io/etcd/issues/17361 to improve it.

However the validation you added would not work. What happens if during downgrade v3.4 member gets a snapshot from leader on v3.5? We need to reimplement parts of migrate code to detect storage version and downgrade it to v3.4, so we remove "term" and "confState" if encountered.

etcdserver/api/membership/cluster.go
```diff
@@ -787,10 +787,15 @@ func ValidateClusterAndAssignIDs(lg *zap.Logger, local *RaftCluster, existing *R
...
+	if cv == nil {
+		return
+	}
 	lv := semver.Must(semver.NewVersion(version.Version))
 	// only keep major.minor version for comparison against cluster version
 	lv = &semver.Version{Major: lv.Major, Minor: lv.Minor}
-	if cv != nil && lv.LessThan(*cv) {
+	// allow 3.4 server to join 3.5 cluster version, since the data is compatible after `etcdutil migrate` the data_dir.
+	allowedDowngradeVersion := &semver.Version{Major: cv.Major, Minor: cv.Minor - 1}
+	if lv.LessThan(*cv) && !lv.Equal(*allowedDowngradeVersion) {
```
serathius, line 798: Based on design I expected this will be version gated.

siyuanfoundation:
> Not sure how this PR refers to dev work described in https://docs.google.com/document/d/1siSo-tA_xLRo5li5wK8KqWsOiq34PIgtmuSDcyxxwsY/edit?usp=sharing
> 
> In document you described that downgrade will gated behind "--allow-downgrade" flag and will include implementation of AuthStatusRequest handling.

With the `etcdutil migrate` and online schema check, "--allow-downgrade" flag is kind of redundant. 
But I see your point about snapshot.

Added "--allow-downgrade" flag and implementation of AuthStatusRequest handling.

mvcc/kvstore.go
siyuanfoundation: done

siyuanfoundation:
cc @lavacat

embed/config.go
```diff
@@ -363,6 +363,8 @@ type Config struct {
 	// UnsafeNoFsync disables all uses of fsync.
 	// Setting this is unsafe and will cause data loss.
 	UnsafeNoFsync bool `json:"unsafe-no-fsync"`
+	// AcceptNextStorageVersion allows server to restore from 3.5 data, to allow downgrade from 3.5 to 3.4.
+	AcceptNextStorageVersion bool `json:"accept-next-storage-version"`
```
serathius, line 367: Users might not be aware of storage version, which was also introduced in v3.6.

Naming is hard, so let's iterate. Maybe something like "can-join-newer-cluster-version"?

etcdserver/server.go
```diff
@@ -379,6 +379,7 @@ func NewServer(cfg ServerConfig) (srv *EtcdServer, err error) {
 		cl.SetID(types.ID(0), existingCluster.ID())
 		cl.SetStore(st)
 		cl.SetBackend(be)
+		cl.SetAcceptNextStorageVersion(cfg.AcceptNextStorageVersion)
```
serathius, line 382: `cfg.AcceptNextStorageVersion` should never change during runtime. Setting it post construction has a risk that we miss it. Could we pass it in `RaftCluster` constructor?

mvcc/downgrade.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	confstate := unsafeConfStateFromBackend(lg, tx)
+	term := unsafeReadTerm(tx)
+	if confstate == nil && term == 0 {
+		// if both confstate and term are missing, assume it's v3.4
+		return version.V3_4, nil
+	} else if confstate == nil {
+		return v, fmt.Errorf("missing confstate information")
+	} else if len(confstate.Voters) == 0 && term == 0 {
```
serathius, line 127: Why distinguish between confstate set and empty? Don't think the code on main branch looks the same. We should provide more context in comments

serathius:
High level looks good, left couple of comments. Some issues possibly come from borrowing code on main branch which was written under different assumption or was just written incorrectly by me :(. Please make sure we fix those issues also on main branch.

embed/config.go
siyuanfoundation: 5 words is a mouthful. How about `forward-compatible`?

serathius:
ping @ahrtr

mvcc/kvstore.go
```diff
@@ -379,6 +380,19 @@ func (s *store) restore() error {
...
+	if !v.Equal(version.V3_4) && !v.Equal(version.V3_5) {
+		if s.lg != nil {
+			s.lg.Panic("unsupported storage version",
+				zap.String("storage-version", v.String()))
+		} else {
+			plog.Panicf("unsupported storage version: %s\n", v.String())
+		}
+	}
+	if v.Equal(version.V3_5) {
+		downgradeMetaBucket(s.lg, tx)
```
ahrtr, line 393: Suggest to gate this by the option `--next-cluster-version-compatible` as well.

etcdmain/config.go
```diff
@@ -194,6 +194,7 @@ func newConfig() *config {
 	fs.BoolVar(&cfg.ec.StrictReconfigCheck, "strict-reconfig-check", cfg.ec.StrictReconfigCheck, "Reject reconfiguration requests that would cause quorum loss.")
 	fs.BoolVar(&cfg.ec.EnableV2, "enable-v2", cfg.ec.EnableV2, "Accept etcd V2 client requests.")
 	fs.BoolVar(&cfg.ec.PreVote, "pre-vote", cfg.ec.PreVote, "Enable to run an additional Raft election phase.")
+	fs.BoolVar(&cfg.ec.NextClusterVersionCompatible, "next-cluster-version-compatible", false, "Allow server to join a cluster with cluster version 3.5.")
```
ahrtr, line 197:
```suggestion
	fs.BoolVar(&cfg.ec.NextClusterVersionCompatible, "next-cluster-version-compatible", false, "Enable 3.4 to be compatible with next version 3.5, to allow 3.4 server to join 3.5 cluster and start on 3.5 schema")
```

mvcc/kvstore.go
siyuanfoundation: the problem with gating this is it might break the upgrade process. 
If the 3.4 cluster is started without `--next-cluster-version-compatible`, when some node is upgraded to 3.5, and some 3.4 node is falling behind. When the 3.4 node receives a snapshot from 3.5 node, we do want to do `downgradeMetaBucket` to avoid any data problems when that node upgrade to 3.5. So this step is needed regardless of `--next-cluster-version-compatible`

ahrtr: I understood your point.

- This is a generic problem. I would suggest to fix it in main firstly, then 3.5 and 3.4.
- I think eventually we do need to gate this under a flag just in case it has any regression.

It can be a followup.

etcdserver/apply.go
```diff
@@ -143,6 +144,8 @@ func (a *applierV3backend) Apply(r *pb.InternalRaftRequest) *applyResult {
 		ar.resp, ar.err = a.s.applyV3.AuthEnable()
 	case r.AuthDisable != nil:
 		ar.resp, ar.err = a.s.applyV3.AuthDisable()
+	case r.AuthStatus != nil:
+		ar.resp, ar.err = a.s.applyV3.AuthStatus()
```
lavacat, line 148: My understanding is that we add AuthStatus only for compatibility, not back-porting full feature. https://github.com/etcd-io/etcd/pull/11536/files#diff-583048be54c04f9555a60191ec8767757fc7f0c477d5bc7f2191af8a3c1bf920

In this case, maybe it's simpler to `return nil` here and not backport applyV3.AuthStatus() and AuthStatusResponse.

etcdserver/verify/verify.go
```diff
@@ -0,0 +1,114 @@
...
+	}()
+
+	beConfig := backend.DefaultBackendConfig()
+	beConfig.Path = toBackendFileName(cfg.DataDir)
+	beConfig.Logger = cfg.Logger
+
+	be := backend.New(beConfig)
+	defer be.Close()
+
+	err = validateSchema(lg, be)
```
lavacat, line 78: This looks like a different implementation of `Verify` from main and `release-3.5` branch.
https://github.com/etcd-io/etcd/blob/main/server/verify/verify.go#L51
It's somewhat confusing for code readability across branches. And we are adding all this code to just check `confstate` and `term` buckets. 
Is this feature required? Looks like `nice-to-have`.

mvcc/kvstore.go
siyuanfoundation: added flag check.

etcdserver/verify/verify.go
siyuanfoundation: This is to ensure we did not miss any part that let the 3.5 schema through. Adding this is more ensuring for supporting something that 3.4 does not support before. The feature is mentioned in https://docs.google.com/document/d/1siSo-tA_xLRo5li5wK8KqWsOiq34PIgtmuSDcyxxwsY/edit#heading=h.efr9asc472y. 

In theory, this schema verification should be in 3.5 and main as well once downgrade feature is complete. I don't see this as different implementation that confuses readability.

etcdserver/apply.go
siyuanfoundation: This can be optional. But I think having the real response is
1. better for readability. 
2. 3.4 does not really know what the response could be used for in 3.5. It is probably safer to return the real response.

mvcc/downgrade.go
```diff
@@ -0,0 +1,126 @@
...
+			plog.Panicf("cannot unmarshal confState json retrieved from the backend, err: %s", err)
+		}
+	}
+	return &confState
+}
+
+// UnsafeDetectSchemaVersion returns version of storage schema. Returned value depends on etcd version that created the backend. For
+// * v3.6 and newer will return storage version.
+// * v3.5 will return it's version if it includes any storage fields added in v3.5.
+// * v3.4 will return it's version if it doesn't include any storage fields added in v3.5.
```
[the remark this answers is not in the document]
siyuanfoundation: Added a comment. As for test, I could not find any difference in the data schemas of 3.3 and 3.4, not sure about what to test.

etcdserver/verify/verify.go
lavacat: 1. I think original motivation for [Verify](https://github.com/etcd-io/etcd/commit/47b28b600adf210fe8e63c5484f63abea16d84fb) was ConsistencyIndex bug that was fixed later. IMO it's using env var as a guard because validateWal is expensive. There were no significant changes after original commit.
2. Your backport has completely different logic. Also removes `ExactIndex` in config.
3. At the moment there is no schema verification in 3.5/main Verify. But there is [schema.Validate](https://github.com/etcd-io/etcd/blob/main/server/etcdserver/bootstrap.go#L239) as part of NewServer logic. So, I'd add your validateSchema in server.NewServer after backend is initialized.

@ahrtr @serathius Please make a final call on this.

ahrtr:
Overall looks good. We need to add a couple of e2e test a well,
- The normal upgrade from 3.3 to 3.4. We need to ensure the normal upgrading from 3.3 to 3.4 isn't impacted by this PR.
- Downgrade from 3.5 to 3.4 
    - `next-cluster-version-compatible = false`
    - `next-cluster-version-compatible = true`

It can be addressed in a followup, but please manually verify the cases.

etcdserver/verify/verify.go
ahrtr: Ideally, the best approach should be always implement in `main` firstly, and then backport to stable releases (3.5 and then 3.4). For this case, we should ideally
- Backport the [verification](https://github.com/etcd-io/etcd/blob/main/server/verify/verify.go) from `main` to 3.4 (already backported to 3.5); This can be a separate PR. Note that we are adding more and more changes in 3.4, this is definitely nice to have to add more protection & verification.
- Add the `validateSchema` in `main`, then backport to 3.5 and 3.4.  This can be discussed separately. (The `schema.Validate` is a different story, it was added in https://github.com/etcd-io/etcd/pull/13200 and never be backported to 3.5). 

For now, I don't want to block this PR as long as they can be addressed in followup PRs. Also https://github.com/etcd-io/etcd/pull/17330#issuecomment-1944436289

etcdserver/apply.go
lavacat: unless you add [EtcdServer.AuthStatus ](https://github.com/etcd-io/etcd/pull/11536/files#diff-583048be54c04f9555a60191ec8767757fc7f0c477d5bc7f2191af8a3c1bf920) that registers request, response won't be used anywhere. This [Trigger](https://github.com/etcd-io/etcd/blob/release-3.4/etcdserver/server.go#L2275) in apply will be noop.

siyuanfoundation: set to nil

siyuanfoundation:
Tested 3.3 -> 3.4 upgrade manually with:

1. start 3.3 cluster
```
../etcd-3.3/etcd --name infra1 --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 --listen-peer-urls http://127.0.0.1:12380 --initial-advertise-peer-urls http://127.0.0.1:12380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new

../etcd-3.3/etcd --name infra2 --listen-client-urls http://127.0.0.1:22379 --advertise-client-urls http://127.0.0.1:22379 --listen-peer-urls http://127.0.0.1:22380 --initial-advertise-peer-urls http://127.0.0.1:22380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new

../etcd-3.3/etcd --name infra3 --listen-client-urls http://127.0.0.1:32379 --advertise-client-urls http://127.0.0.1:32379 --listen-peer-urls http://127.0.0.1:32380 --initial-advertise-peer-urls http://127.0.0.1:32380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state new
```
2. verify status
```
go run tools/benchmark/main.go put --total=10000000 --key-size=64 --val-size=512`

bin/etcdctl --endpoints=http://127.0.0.1:2379,http://127.0.0.1:22379,http://127.0.0.1:32379 endpoint status -w table
```
3. upgrade server one by one
```
../etcd-3.4/bin/etcd --name infra3 --listen-client-urls http://127.0.0.1:32379 --advertise-client-urls http://127.0.0.1:32379 --listen-peer-urls http://127.0.0.1:32380 --initial-advertise-peer-urls http://127.0.0.1:32380 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state existing --data-dir infra3.etcd 

../etcd-3.4/bin/etcd --name infra2 --listen-client-urls http://127.0.0.1:22379 --advertise-client-urls http://127.0.0.1:22379 --listen-peer-urls http://127.0.0.1:22380 --initial-advertise-peer-urls http://127.0.0.1:22380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state existing --data-dir infra2.etcd

../etcd-3.4/bin/etcd --name infra1 --listen-client-urls http://127.0.0.1:2379 --advertise-client-urls http://127.0.0.1:2379 --listen-peer-urls http://127.0.0.1:12380 --initial-advertise-peer-urls http://127.0.0.1:12380 --initial-cluster-token etcd-cluster-1 --initial-cluster 'infra1=http://127.0.0.1:12380,infra2=http://127.0.0.1:22380,infra3=http://127.0.0.1:32380' --initial-cluster-state existing --data-dir infra1.etcd
```
4. verify status
```
bin/etcdctl --endpoints=http://127.0.0.1:2379,http://127.0.0.1:22379,http://127.0.0.1:32379 endpoint status -w table
```

mvcc/downgrade.go
```diff
@@ -0,0 +1,128 @@
...
+	// meta bucket keys since 3.6
+	storageVersionKeyName = []byte("storageVersion")
+)
+
+// downgradeMetaBucket delete 3.5 specific keys to make backend fully compatible with 3.4
+func downgradeMetaBucket(lg *zap.Logger, tx backend.BatchTx) {
+	if lg != nil {
+		lg.Info(
+			"downgrade meta bucket",
+			zap.String("remove-keys", string(confStateKeyName)+","+string(termKeyName)),
```
serathius, line 41:
```suggestion
			zap.ByteStrings("remove-keys", [][]byte{confStateKeyName, termKeyName}),
```

serathius (approved):

ahrtr:
cc @fuweid to take a look. This PR should be also important to AKS.

fuweid (commented):
The change looks good. But it seems that this patch needs `go mod tidy`.

ahrtr:
> But it seems that this patch needs `go mod tidy`.

This PR doesn't change anything for go.mod and go.sum files. I do not see any issue when running "go mod tidy" command.  Please ensure you use exactly the same go version as https://github.com/etcd-io/etcd/blob/release-3.4/.go-version#L1

fuweid:
> This PR doesn't change anything for go.mod and go.sum files. I do not see any issue when running "go mod tidy" command.

If you pull this pull request in your local, you will see that. This pull request doesn't rebase so it shows diff.
It's fine. It's not blocking comment.

ahrtr:
> If you pull this pull request in your local, you will see that. This pull request doesn't rebase so it shows diff.

Yes, it's true. Curious why the workflow did not fail.

Anyway, please rebase this PR. @siyuanfoundation

siyuanfoundation:
rebased.  the workflow still passes.

ahrtr (approved):

Issue #17345, opened 2024-01-30
Bump golang.org/x/crypto to 0.17.0+ to address CVE-2023-48795


serathius:
### What would you like to be added?

Bump golang.org/x/crypto to 0.17.0+

### Why is this needed?

address CVE-2023-48795

serathius:
Fixed

PR #17346, opened 2024-01-30
Bump golang.org/x/crypto to v0.17.0 to address CVE-2023-48795

serathius:
Ref https://github.com/etcd-io/etcd/issues/17345


server/go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.20.0
 	go.uber.org/multierr v1.6.0
 	go.uber.org/zap v1.17.0
-	golang.org/x/crypto v0.14.0
+	golang.org/x/crypto v0.17.0
```
ahrtr, line 39: Please bump to 0.18.0 so that it's >= the version being used in 3.4, and also ensure all modules use the same version of each dependency (golang.org/x/crypto in this case)

serathius: SG

serathius: Managed to rework the https://github.com/etcd-io/etcd/pull/17347 managed to upgrade crypto to just 0.17.0 on release-3.4

ahrtr: > Managed to rework the #17347 managed to upgrade crypto to just 0.17.0 on release-3.4

Did you run into any issue when bumping to v0.18.0? A higher version is recommended.

serathius: Upgrading crypto to v0.18.0 required bumping grpc which I wanted to avoid. My goal was to do a minimal dependency bump that addresses the CVE. I would prefer to rather bump crypto to 0.17.0 (minimal required to address CVE) than bump crypto to 0.18.0 and grpc to 1.60.

ahrtr: OK

ahrtr (approved):

Issue #16838, opened 2023-10-27
Test PrevKey in robustness test


serathius:
### What would you like to be added?

Kubernetes establishes watches with PrevKey enabled.

TODO:
* Add WithPrevKey() to Kubernetes traffic https://github.com/etcd-io/etcd/blob/main/tests/robustness/traffic/kubernetes.go
* Add PrevKey validation to watch https://github.com/etcd-io/etcd/blob/main/tests/robustness/validate/watch.go

### Why is this needed?

To improve correctness of testing Kubernetes-etcd contract

MadhavJivrajani:
Hey @sharathsivakumar 👋🏼 

Wanted to check in and see if you're still working on this? I'd love to work on it or maybe collaborate if you're up for it. I have a rough PR that I can put up in a couple of days in case you haven't already started 😄

sharathsivakumar:
Hi @MadhavJivrajani I have started working on this and I will be turning in my PR sometime next week.

PR #17352, opened 2024-01-31
tests/robustness: add robustness test for watch with PrevKV()

MadhavJivrajani:
Kubernetes relies on the PrevKV() option in the watches it opens
against etcd. This commit adds a robustness test to validate the
same.

A watch response returned with PrevKV() is valid if:
The value in current event's prevKV matches the previous
event's value of the same key if this is not a create event.

There are cases where there can be a prevKV for a create event
as well, for example if a watch is opened after the key is creatd.
Since we don't simulate for that, we don't check for that.

Further, this adjusts revision numbers such that we can successfully create
a new replay. Needed now since we will have unit tests with
and without PrevKV co-existing and we requite creation of a
new replay everytime we validate PrevKV.

We also regenerate test data with so that prevKV exists in it

Fixes: https://github.com/etcd-io/etcd/issues/16838

/assign @serathius


tests/robustness/watch.go
```diff
@@ -70,7 +71,7 @@ func watchUntilRevision(ctx context.Context, t *testing.T, c *traffic.RecordingC
 	var lastRevision int64
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	watch := c.Watch(ctx, "", 1, true, true)
+	watch := c.Watch(ctx, "", 1, true, true, cfg.withPrevKV)
```
serathius, line 74: No, need to pass `withPrevKV` here, it's enough that Kubernetes Traffic uses PrevKV. We can pass `false` here.

tests/robustness/validate/validate.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // ValidateAndReturnVisualize returns visualize as porcupine.linearizationInfo used to generate visualization is private.
-func ValidateAndReturnVisualize(t *testing.T, lg *zap.Logger, cfg Config, reports []report.ClientReport) (visualize func(basepath string) error) {
+func ValidateAndReturnVisualize(t *testing.T, lg *zap.Logger, cfg Config, reports []report.ClientReport, withPrevKV bool) (visualize func(basepath string) error) {
```
serathius, line 31: Don't think we need to pass `withPrevKV` here. We can check if WatchRequest has `PrevKV` set.

tests/robustness/model/replay.go
```diff
@@ -99,13 +99,26 @@ func (r *EtcdReplay) next() (request EtcdRequest, revision int64, index int) {
...
 }
 
 type Event struct {
-	Type  OperationType
-	Key   string
-	Value ValueOrHash
+	Type      OperationType
+	Key       string
+	Value     ValueOrHash
+	PrevValue *ValueRevision
```
serathius, line 121: I would expect that `PrevValue` should be on `WatchEvent`. What was the reason to put it here?

tests/robustness/validate/patch_history.go
```diff
@@ -24,8 +24,8 @@ import (
 
 func patchedOperationHistory(reports []report.ClientReport) []porcupine.Operation {
 	allOperations := operations(reports)
-	uniqueEvents := uniqueWatchEvents(reports)
-	return patchOperationsWithWatchEvents(allOperations, uniqueEvents)
+	uniqueEvents, prevKVs := uniqueWatchEvents(reports)
+	return patchOperationsWithWatchEvents(allOperations, uniqueEvents, prevKVs)
```
serathius, line 28: Don't think we need to handle prevKV in the history patching at all.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +141,52 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+		if firstEvent != nil && !events[index].Equals(*firstEvent) {
+			t.Errorf("Resumable - A broken watch can be resumed by establishing a new watch starting after the last revision received in a watch event before the break, so long as the revision is in the history window, watch request: %+v, event missing: %+v, got: %+v", op.Request, events[index], *firstEvent)
+		}
+	}
+}
+
+// validatePrevKV ensures that a watch response (if configured with WithPrevKV()) returns
+// the appropriate response.
+func validatePrevKV(t *testing.T, report report.ClientReport) {
+	for _, op := range report.Watch {
```
serathius, line 153:
```suggestion
	for _, request := range report.Watch {
		if !request.WithPrevKV {
			continue
		}
```

tests/robustness/model/replay.go
serathius: Possibly we will need to split off another Event type.
* `Event` that has `Type`, `Key`, `Value`
* `PersistedEvent` that has `Event`, `Revision` and `IsCreate`.
* `WatchEvent` that has `PersistedEvent` and `PrevValue`

tests/robustness/validate/watch.go
```diff
@@ -138,8 +141,52 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+		keyTracker := map[string]model.ValueOrHash{}
+		for _, resp := range op.Responses {
+			for _, event := range resp.Events {
```
serathius, line 156: Don't think you can validate prevKV this way. 
Watch can be established after the key was created, and not observe the event of key creation even though it will report it in `PrevKey`. 

I see 2 ways to validate PrevKey:
* Take event history, read it and save history of each key in map[string][]WatchEvent. Then go through events in WatchRequest, get the history of changes of that key. Find the preceeding change and compare to prevKey.
* [preferred] Use the model replay. Go through events in WatchRequest, load model replay to revision before the event. Read the key from model and compare if it matches prevKey.

MadhavJivrajani: @serathius I think I understand what you mean here: can you please check this commit and see if this is accurate: https://github.com/etcd-io/etcd/pull/17352/commits/ea5dcde1aa218f21089c49012f406d244a414e0f

serathius: Looks good, lefts some comments.

tests/robustness/watch.go
MadhavJivrajani: @serathius not sure I fully understand here - the test runner calls into a common watch method for both Kubernetes and non kubernetes traffic, top of the call stach being here: 
https://github.com/etcd-io/etcd/blob/43d675997772bee9cf9d6bd4804ef05cc0cecb01/tests/robustness/main_test.go#L134

tests/robustness/validate/watch.go
```diff
@@ -138,8 +141,44 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				// It is not nescessary that PrevValue will always be nil for a current event.
+				// However, if PrevValue is nil provided a watch is opened with PrevKV configured,
+				// then that has to be a create event.
+				//
+				// Furthermore, this explicitly checks the condition that is used by Kubernetes
+				// while parsing events received from etcd:
+				// https://github.com/kubernetes/kubernetes/blob/a9e4f5b7862e84c4152eabe2e960f3f6fb9a4867/staging/src/k8s.io/apiserver/pkg/storage/etcd3/event.go#L59
+				if event.PrevValue == nil && !event.CreateEvent {
+					t.Errorf("PrevKV - got a create event with nil prevKV %+v", event)
+				}
```
MadhavJivrajani, line 172: @serathius thinking about this a little more, could use your input:

I'm not sure anymore if we should be checking this condition anymore here. Looks like in the face of compaction, prevKV can also be nil: 
https://github.com/etcd-io/etcd/blob/43d675997772bee9cf9d6bd4804ef05cc0cecb01/api/etcdserverpb/rpc.proto#L799-L801

tests/robustness/model/replay.go
MadhavJivrajani: Its probably better to split this off into another PR and then rebase that into this once merged. I'll send one in.

MadhavJivrajani:
Addressed reviews

MadhavJivrajani:
Hmm

```
=== Failed
=== FAIL: robustness/validate TestValidateWatch/earlier_event_after_bookmark_in_separate_request (0.00s)
    logger.go:130: 2024-02-05T10:04:25.751Z	INFO	Validating watch
    --- FAIL: TestValidateWatch/earlier_event_after_bookmark_in_separate_request (0.00s)
panic: Replay requires a complete event history [recovered]
	panic: Replay requires a complete event history
goroutine 79 [running]:
testing.tRunner.func1.2({0x1024500, 0x12e4220})
	/usr/local/go/src/testing/testing.go:1545 +0x3f7
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1548 +0x716
panic({0x1024500?, 0x12e4220?})
	/usr/local/go/src/runtime/panic.go:920 +0x270
go.etcd.io/etcd/tests/v3/robustness/model.NewReplay(...)
	/home/prow/go/src/github.com/etcd-io/etcd/tests/robustness/model/replay.go:26
go.etcd.io/etcd/tests/v3/robustness/validate.validatePrevKV(0xc0005ce4e0, {0x0, {0x0, 0x0, 0x0}, {0xc000036230, 0x2, 0x2}}, {0xc000137400, 0x1, ...})
	/home/prow/go/src/github.com/etcd-io/etcd/tests/robustness/validate/watch.go:151 +0x8be
go.etcd.io/etcd/tests/v3/robustness/validate.validateWatch(0x12f2470?, 0xc0005ce4e0?, {0x0?}, {0xc000980040, 0x1, 0x24?}, {0xc000137400, 0x1, 0x1})
	/home/prow/go/src/github.com/etcd-io/etcd/tests/robustness/validate/watch.go:38 +0x137
go.etcd.io/etcd/tests/v3/robustness/validate.TestValidateWatch.func1(0xc0005ce4e0)
	/home/prow/go/src/github.com/etcd-io/etcd/tests/robustness/validate/validate_test.go:231 +0x165
testing.tRunner(0xc0005ce4e0, 0xc0004f63e0)
	/usr/local/go/src/testing/testing.go:1595 +0x262
created by testing.(*T).Run in goroutine 78
	/usr/local/go/src/testing/testing.go:1648 +0x846
```

tests/robustness/validate/watch.go
serathius: It's good point, please skip it for now as we don't compact in robustness tests for now.  

I'm not sure documentation is correct as Kubernetes contract correctness depends on this field being available. Will investigate it further.

tests/robustness/validate/watch.go
```diff
@@ -117,8 +118,8 @@ func validateReliable(t *testing.T, events []model.WatchEvent, report report.Cli
 		}
 		for _, resp := range op.Responses {
 			for _, event := range resp.Events {
-				if events[index].Match(op.Request) && events[index] != event {
-					t.Errorf("Broke watch guarantee: Reliable - a sequence of events will never drop any subsequence of events; if there are events ordered in time as a < b < c, then if the watch receives events a and c, it is guaranteed to receive b, event missing: %+v", events[index])
+				if events[index].Match(op.Request) && !events[index].Equals(event) {
```
serathius, line 121:
```suggestion
				if events[index].Match(op.Request) && events[index].PersistedEvent != event.PersistedEvent {
```

We don't want/need t validate whole WatchEvent, just the peristedEvent within it.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,46 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
 		}
 		firstEvent := firstWatchEvent(op)
 		// If watch is resumable, first event it gets should the first event that happened after the requested revision.
-		if firstEvent != nil && events[index] != *firstEvent {
-			t.Errorf("Resumable - A broken watch can be resumed by establishing a new watch starting after the last revision received in a watch event before the break, so long as the revision is in the history window, watch request: %+v, event missing: %+v", op.Request, events[index])
+		if firstEvent != nil && !events[index].Equals(*firstEvent) {
```
serathius, line 142: ditto

tests/robustness/validate/watch.go
MadhavJivrajani: The discussion here is most relevant: https://github.com/kubernetes/kubernetes/issues/76624

serathius: Right.

serathius:
Ping @MadhavJivrajani, do you need any help?

tests/robustness/watch.go
serathius: Please address the comment

tests/robustness/validate/validate_test.go
```diff
@@ -37,6 +37,8 @@ func TestValidate(t *testing.T) {
 			path := filepath.Join(testdataPath, file.Name())
 			reports, err := report.LoadClientReports(path)
 			assert.NoError(t, err)
+			// TODO(MadhavJivrajani): update test here to include WithPrevKV()
+			// Will probably need to repopulate testdata.
```
MadhavJivrajani, line 41: @serathius apart from https://github.com/etcd-io/etcd/pull/17352#discussion_r1485965942, I could use your input on this ^

I'm still having some difficulty running a robustness test which is on a cluster of size 3, even with the `LowTraffic` profile, this is what I get:
```
    logger.go:130: 2024-02-12T15:42:14.448+0530	INFO	Recorded operations	{"operationCount": 190}
    logger.go:130: 2024-02-12T15:42:14.448+0530	INFO	Average traffic	{"qps": 53.50818790796354}
    traffic.go:105: Requiring minimal 100.000000 qps for test results to be reliable, got 53.508188 qps
```
It is worth noting that this is kinda flaky because sometimes it does end up working. What I am wondering is why that is the case, I don't see flakiness in the CI for this type of test and my system seems to have the right specs to run this as well
```
❯ system_profiler SPHardwareDataType
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Processor Name: 8-Core Intel Core i9
      Processor Speed: 2.4 GHz
      Number of Processors: 1
      Total Number of Cores: 8
      L2 Cache (per Core): 256 KB
      L3 Cache: 16 MB
      Hyper-Threading Technology: Enabled
      Memory: 32 GB
```

tests/robustness/watch.go
serathius: Even if you enable WithPrevKey on all watches, as we established in previous comments, etcd doesn't guarantee that it will provide it. So the issue with merging watch history could still happen.

The proper solution would be to not use `PrevValue` when merging watch history. It's redundant and not needed. I would recommend that `mergeWatchEventHistory` should not care about PreviousKV and returns `[]PersistedEvent` and not `[]WatchEvent`.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,46 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				}
+				var prevValue model.ValueRevision
+				if event.PrevValue != nil {
+					prevValue = *event.PrevValue
+				}
+				// A create event will not have an entry in our history.
+				if _, ok := state.KeyValues[event.Key]; !event.IsCreate && !ok {
+					t.Error("PrevKV - could not find event in past history")
+				}
+				if prevValue != state.KeyValues[event.Key] {
```
serathius, line 178: Would be good to comment that this should hold if there is no compaction.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,46 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				// https://github.com/kubernetes/kubernetes/blob/a9e4f5b7862e84c4152eabe2e960f3f6fb9a4867/staging/src/k8s.io/apiserver/pkg/storage/etcd3/event.go#L59
+				if event.PrevValue == nil && !event.IsCreate {
+					t.Errorf("PrevKV - got a create event with nil prevKV %+v", event)
+				}
+				var prevValue model.ValueRevision
+				if event.PrevValue != nil {
+					prevValue = *event.PrevValue
+				}
+				// A create event will not have an entry in our history.
+				if _, ok := state.KeyValues[event.Key]; !event.IsCreate && !ok {
```
serathius, line 175:
```suggestion
				if _, prevKeyExists := state.KeyValues[event.Key]; event.IsCreate == prevKeyExists {
```

Both should cause error:
*  `IsCreate=true` and `ok=true` 
* `IsCreate=false` and `ok=false`

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,46 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+					t.Error("PrevKV - could not find event in past history")
```
serathius, line 176: Can you provide more debug info?

tests/robustness/validate/validate_test.go
serathius: Not sure, we have the qps validation to make sure that tests serve their purpose and have enough QPS to trigger historical issues. 100 is pretty low qps to ensure it runs on CI. Don't think this is CPU limitation as I get 300 qps on my laptop 13 inch in power saver mode.

Don't know how to help you in this case, you can lower the qps min target locally, and we can validate the change fully on CI.

MadhavJivrajani: I regenerated the test data by changing the min qps to 50

tests/robustness/watch.go
```diff
@@ -62,6 +62,7 @@ func collectClusterWatchEvents(ctx context.Context, t *testing.T, clus *e2e.Etcd
 
 type watchConfig struct {
 	requestProgress bool
+	withPrevKV      bool
```
serathius, line 65: Not needed now

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,48 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				if event.PrevValue == nil && !event.IsCreate {
+					t.Errorf("PrevKV - got a create event with nil prevKV %+v", event)
+				}
+				var prevValue model.ValueRevision
+				if event.PrevValue != nil {
+					prevValue = *event.PrevValue
+				}
+				// A create event will not have an entry in our history.
+				if _, prevKeyExists := state.KeyValues[event.Key]; event.IsCreate == prevKeyExists {
+					t.Errorf("PrevKV - could not find event in past history, event: %+v", event)
```
serathius, line 176: Please update error message

tests/robustness/validate/validate.go
```diff
@@ -71,7 +71,7 @@ func mergeWatchEventHistory(reports []report.ClientReport) ([]model.WatchEvent,
 						if prev, found := revisionToEvents[lastRevision]; found {
 							// This assumes that there are txn that would be observed differently by two watches.
 							// TODO: Implement merging events from multiple watches about single revision based on operations.
-							if diff := cmp.Diff(prev.events, events); diff != "" {
+							if diff := cmp.Diff(prev.events, events, cmp.Comparer(compareWatchEvents)); diff != "" {
```
serathius, line 74: It would be be simpler and cleaner for this function to track and return `[]model.PersistedEvent` instead of rewriting comparer to skip everything apart of `PersistedEvent` field.

MadhavJivrajani: You're right - it would be cleaner. The reason for me doing this is: https://github.com/etcd-io/etcd/pull/17352#discussion_r1486491484

Would you prefer patching it later before it is passed into `validatePrevKV`?

tests/robustness/watch.go
serathius: Not sure I follow, output of `mergeWatchEventHistory` serves as authoritative list of events that happened. We don't need prevKV there. Not sure what you mean by `otherwise me would have to patch that before passing it into the validate funcs`. The prevKV in eventHistory is not needed for for validators. It just a list of operations in order we believe that were persisted by etcd. 

Changing it might require a bigger refactor, so we can skip it for this PR.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,48 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				// https://github.com/kubernetes/kubernetes/blob/a9e4f5b7862e84c4152eabe2e960f3f6fb9a4867/staging/src/k8s.io/apiserver/pkg/storage/etcd3/event.go#L59
+				if event.PrevValue == nil && !event.IsCreate {
+					t.Errorf("PrevKV - got a create event with nil prevKV %+v", event)
+				}
+				var prevValue model.ValueRevision
+				if event.PrevValue != nil {
+					prevValue = *event.PrevValue
+				}
+				// A create event will not have an entry in our history.
+				if _, prevKeyExists := state.KeyValues[event.Key]; event.IsCreate == prevKeyExists {
```
serathius, line 175: Note, event.IsCreate is set even if watch doesn't use WithPrevKey. This could be validated independent to prevKey.

tests/robustness/watch.go
MadhavJivrajani: Yeah that's what I meant - in order to refactor it, we would have to change what the model operates on as well. I can send a follow up after this one

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,48 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				// get state state just before the current event.
+				state, err := replay.StateForRevision(event.Revision - 1)
+				if err != nil {
+					t.Error(err)
+				}
+
+				// This explicitly checks the condition that is used by Kubernetes
+				// while parsing events received from etcd:
+				// https://github.com/kubernetes/kubernetes/blob/a9e4f5b7862e84c4152eabe2e960f3f6fb9a4867/staging/src/k8s.io/apiserver/pkg/storage/etcd3/event.go#L59
+				if event.PrevValue == nil && !event.IsCreate {
```
serathius, line 167: Note, same us check on line 180, this only holds true where there is no compaction.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,48 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+					t.Errorf("PrevKV - got a create event with nil prevKV %+v", event)
+				}
+				var prevValue model.ValueRevision
```
serathius, line 170: Nit, you declare and prepare variable `prevValue` on lines 170-173, but then do some other validation that doesn't require it, and only use it on line 180. It would be clearer to read (less context switches) if there was no interrupt, so creation and use of variable would be closer.

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,48 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+				if event.PrevValue != nil {
+					prevValue = *event.PrevValue
+				}
+				// A create event will not have an entry in our history.
+				if _, prevKeyExists := state.KeyValues[event.Key]; event.IsCreate == prevKeyExists {
+					t.Errorf("PrevKV - unexpected event ecountered, create event should not be in event history and update/delete event should be, event already exists: %t, is create event: %t, event: %+v", prevKeyExists, event.IsCreate, event)
+				}
+				// This holds true when there is no compaction.
```
serathius, line 178: I run the robustness tests with your changes and found an issue. I was wrong, robustness do run compaction during failpoint injection.

 https://github.com/etcd-io/etcd/blob/cb42e2264cd75dfb3e8fcdae7837561aa7b28596/tests/robustness/failpoint/gofail.go#L43-L48

Looks like we need to relax the validation in a way you proposed originally. Sorry for confusion.
In a followup PR we might consider checking whether compaction was executed in the test and requiring prevKV if there was none.

tests/robustness/validate/watch.go
MadhavJivrajani: That's a good point, its probably worth adding in another validate function that can test this, it would serve as a sanity check - I'll send in a follow up.

tests/robustness/validate/watch.go
MadhavJivrajani: This is very helpful - thank you. I'll make a note about sending a following up to see how we can do this.

MadhavJivrajani:
Follow ups to this PR so far:
- Validate `IsCreate` independent of `prevKV` (https://github.com/etcd-io/etcd/pull/17352#discussion_r1487455729)
- Refactor `mergeWatchEventHistory` to operate only on `PersistedEvent`s (https://github.com/etcd-io/etcd/pull/17352#discussion_r1472792763)
- See if we can check if compaction has happened already and based on that validate the presence of `prevKV` (https://github.com/etcd-io/etcd/pull/17352#discussion_r1487485666)

tests/robustness/validate/watch.go
MadhavJivrajani: I've relaxed the validation here

tests/robustness/validate/watch.go
```diff
@@ -138,8 +139,53 @@ func validateResumable(t *testing.T, events []model.WatchEvent, report report.Cl
...
+					t.Errorf("PrevKV - unexpected event ecountered, create event should not be in event history and update/delete event should be, event already exists: %t, is create event: %t, event: %+v", prevKeyExists, event.IsCreate, event)
+				}
+				var prevValue model.ValueRevision
+				if event.PrevValue != nil {
+					prevValue = *event.PrevValue
+				}
+				// This holds true when there is no compaction.
+				// Robustness tests currently do not do compaction.
+				if prevValue != state.KeyValues[event.Key] {
+					t.Errorf("PrevKV - PrevValue doesn't match previous value under the key %s, got: %+v, want: %+v", event.Key, prevValue, state.KeyValues[event.Key])
```
serathius, line 186:
```suggestion
				if event.PrevValue != nil && *event.PrevValue != state.KeyValues[event.Key] {
					t.Errorf("PrevKV - PrevValue doesn't match previous value under the key %s, got: %+v, want: %+v", event.Key, event.PrevValue, state.KeyValues[event.Key])
```

tests/robustness/validate/watch.go
```diff
@@ -176,14 +176,10 @@ func validatePrevKV(t *testing.T, report report.ClientReport, history []model.Wa
...
 				}
-				var prevValue model.ValueRevision
-				if event.PrevValue != nil {
-					prevValue = *event.PrevValue
-				}
-				// This holds true when there is no compaction.
-				// Robustness tests currently do not do compaction.
-				if prevValue != state.KeyValues[event.Key] {
-					t.Errorf("PrevKV - PrevValue doesn't match previous value under the key %s, got: %+v, want: %+v", event.Key, prevValue, state.KeyValues[event.Key])
+				// Perform the check only if there is no compaction.
```
serathius, line 179: Comment is not accurate. More accurate would be to say that we allow PrevValue to be nil as etcd doesn't guarantee its presence due to compaction.

serathius (approved):
Aside of one comment, LGTM.

cc @ahrtr

serathius:
Please squash commits.

MadhavJivrajani:
Done @serathius, thanks for all the help here!

serathius:
Let's merge and iterate on it.

Issue #17126, opened 2023-12-15
Revise backend transaction testing


serathius:
### What would you like to be added?

In https://github.com/etcd-io/etcd/pull/17119#issuecomment-1857623923 we discovered an issue with txn buffers. The writes are not propagated to the read transaction buffer. Repro https://github.com/etcd-io/etcd/pull/17124

Best would be to apply some black box approach that generates random combination of reads and writes and validates that results are correct. Similar to robustness testing we did for etcd API and bbolt.

### Why is this needed?

We need etcd backend code to be a solid foundation that we can trust.

PR #17359, opened 2024-02-02
Add VerifyTxConsistency to backend.

siyuanfoundation:
Followup test verification for https://github.com/etcd-io/etcd/issues/17158

Should help us find all places impacted by readbuf inconsistency.
Similar to https://github.com/etcd-io/etcd/pull/17127

### DIFF
--- a/server/storage/backend/verify.go
+++ b/server/storage/backend/verify.go
@@ -15,9 +15,11 @@
 package backend
 
 import (
+	"fmt"
 	"runtime/debug"
 	"strings"
 
+	"github.com/google/go-cmp/cmp"
 	"go.uber.org/zap"
 
 	"go.etcd.io/etcd/client/pkg/v3/verify"
@@ -66,4 +68,42 @@ func insideApply() bool {
 func insideUnittest() bool {
 	stackTraceStr := string(debug.Stack())
 	return strings.Contains(stackTraceStr, "_test.go") && !strings.Contains(stackTraceStr, "tests/")
+}
+
+// VerifyBackendConsistency verifies data in ReadTx and BatchTx are consistent.
+func VerifyBackendConsistency(b Backend, lg *zap.Logger, skipSafeRangeBucket bool, bucket ...Bucket) {
+	verify.Verify(func() {
+		if b == nil {
+			return
+		}
+		if lg != nil {
+			lg.Debug("verifyBackendConsistency", zap.Bool("skipSafeRangeBucket", skipSafeRangeBucket))
+		}
+		b.BatchTx().LockOutsideApply()
+		defer b.BatchTx().Unlock()
+		b.ReadTx().RLock()
+		defer b.ReadTx().RUnlock()
+		for _, bkt := range bucket {
+			if skipSafeRangeBucket && bkt.IsSafeRangeBucket() {
+				continue
+			}
+			unsafeVerifyTxConsistency(b, bkt)
+		}
+	})
+}
+
+func unsafeVerifyTxConsistency(b Backend, bucket Bucket) {
+	dataFromWriteTxn := map[string]string{}
+	b.BatchTx().UnsafeForEach(bucket, func(k, v []byte) error {
+		dataFromWriteTxn[string(k)] = string(v)
+		return nil
+	})
+	dataFromReadTxn := map[string]string{}
+	b.ReadTx().UnsafeForEach(bucket, func(k, v []byte) error {
+		dataFromReadTxn[string(k)] = string(v)
+		return nil
+	})
+	if diff := cmp.Diff(dataFromWriteTxn, dataFromReadTxn); diff != "" {
+		panic(fmt.Sprintf("bucket %s data mismatch\nwrite TXN: %v\nread TXN: %v\ndiff: %s", bucket.String(), dataFromWriteTxn, dataFromReadTxn, diff))
+	}
 }
--- a/etcdutl/go.mod
+++ b/etcdutl/go.mod
@@ -44,6 +44,7 @@ require (
 	github.com/golang-jwt/jwt/v4 v4.5.0 // indirect
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/google/btree v1.1.2 // indirect
+	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.19.1 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jonboulle/clockwork v0.4.0 // indirect
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -946,6 +946,7 @@ func (s *EtcdServer) Cleanup() {
 func (s *EtcdServer) applyAll(ep *etcdProgress, apply *toApply) {
 	s.applySnapshot(ep, apply)
 	s.applyEntries(ep, apply)
+	backend.VerifyBackendConsistency(s.Backend(), s.Logger(), true, schema.AllBuckets...)
 
 	proposalsApplied.Set(float64(ep.appliedi))
 	s.applyWait.Trigger(ep.appliedi)
@@ -2272,6 +2273,7 @@ func (s *EtcdServer) monitorKVHash() {
 			return
 		case <-checkTicker.C:
 		}
+		backend.VerifyBackendConsistency(s.be, lg, false, schema.AllBuckets...)
 		if !s.isLeader() {
 			continue
 		}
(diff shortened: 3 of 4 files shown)


siyuanfoundation:
cc @serathius @ahrtr

server/storage/backend/backend.go
```diff
@@ -71,6 +73,9 @@ type Backend interface {
 
 	// SetTxPostLockInsideApplyHook sets a txPostLockInsideApplyHook.
 	SetTxPostLockInsideApplyHook(func())
+	// VerifyTxConsistency verifies data in ReadTx and BatchTx are consistent.
+	VerifyTxConsistency(bucket Bucket)
+	UnsafeVerifyTxConsistency(bucket Bucket)
```
ahrtr, line 78: Let's try not change the important interface just for the test. Can you find an alternative way to do it?

ahrtr:
Also let's simplify the verification just at one place (e.g. **at the end of each etcd transaction**)

server/storage/backend/backend.go
siyuanfoundation: changed it to a verify util function.

siyuanfoundation:
> Also let's simplify the verification just at one place (e.g. **at the end of each etcd transaction**)

Doing it in applyAll is essentially **at the end of each etcd transaction**. 
Please see https://github.com/etcd-io/etcd/pull/17127 for the tx level approach and why it does not work.

server/storage/backend/verify.go
```diff
@@ -67,3 +69,41 @@ func insideUnittest() bool {
...
+				continue
+			}
+			unsafeVerifyTxConsistency(b, bkt)
+		}
+	})
+}
+
+func unsafeVerifyTxConsistency(b Backend, bucket Bucket) {
+	kvsFromRead := map[string]string{}
+	kvsFromBbolt := map[string]string{}
```
ahrtr, line 97:
```suggestion
	dataFromReadTxn := map[string]string{}
	dataFromWriteTxn := map[string]string{}
```

server/storage/backend/verify.go
```diff
@@ -67,3 +69,41 @@ func insideUnittest() bool {
...
+	b.BatchTx().UnsafeForEach(bucket, func(k, v []byte) error {
+		kvsFromBbolt[string(k)] = string(v)
+		return nil
+	})
+	b.ReadTx().UnsafeForEach(bucket, func(k, v []byte) error {
+		kvsFromRead[string(k)] = string(v)
+		return nil
+	})
+	if diff := cmp.Diff(kvsFromBbolt, kvsFromRead); diff != "" {
+		panic(fmt.Sprintf("bucket %s kvs mismatch\nbbolt: %v\nreadbuf: %v\ndiff: %s", bucket.String(), kvsFromBbolt, kvsFromRead, diff))
```
ahrtr, line 107:
```suggestion
		panic(fmt.Sprintf("bucket %s data mismatch\write TXN: %v\nread TXN: %v\ndiff: %s", bucket.String(), kvsFromBbolt, kvsFromRead, diff))
```

server/storage/backend/verify.go
```diff
@@ -67,3 +69,41 @@ func insideUnittest() bool {
...
+// VerifyBackendConsistency verifies data in ReadTx and BatchTx are consistent.
+func VerifyBackendConsistency(b Backend, lg *zap.Logger, skipSafeRangeBucket bool, bucket ...Bucket) {
+	verify.Verify(func() {
+		if b == nil {
+			return
+		}
+		if lg != nil {
+			lg.Debug("verifyBackendConsistency", zap.Bool("skipSafeRangeBucket", skipSafeRangeBucket))
+		}
+		b.BatchTx().LockOutsideApply()
```
serathius, line 82: Are we sure we should use LockOutsideApply? We call this function in applyAll

siyuanfoundation: yes, LockInsideApply would fail.

server/storage/backend/verify.go
```diff
@@ -67,3 +69,41 @@ func insideUnittest() bool {
...
+	b.BatchTx().UnsafeForEach(bucket, func(k, v []byte) error {
+		dataFromWriteTxn[string(k)] = string(v)
+		return nil
+	})
+	dataFromReadTxn := map[string]string{}
+	b.ReadTx().UnsafeForEach(bucket, func(k, v []byte) error {
+		dataFromReadTxn[string(k)] = string(v)
+		return nil
+	})
+	if diff := cmp.Diff(dataFromWriteTxn, dataFromReadTxn); diff != "" {
```
serathius, line 106: Have you confirmed that this would detect all the recently discovered problems?

When I proposed https://github.com/etcd-io/etcd/issues/17126 I was thinking that we should implement a stub for backend that would implement the same methods but without bbolt and keep the data only in memory. With validate enabled, we would send all request to bbolt to both backend and the stub and compare results. 

I think this would be a better long term approach, however your solution should be acceptable assuming it can reproduce previous issues.

siyuanfoundation: It can detect https://github.com/etcd-io/etcd/issues/17158. 
For https://github.com/etcd-io/etcd/pull/17263, since it is rare to write multiple keys in 1 tx, there is no existing issues to detect.

server/storage/backend/verify.go
ahrtr: We should use `LockOutsideApply`. This is just verification code, obviously we don't need to execute the [txPostLockInsideApplyHook](https://github.com/etcd-io/etcd/blob/cf7043391c624afb5be879592ce24a136840a1bd/server/storage/backend/backend.go#L73).

server/storage/backend/verify.go
```diff
@@ -67,3 +69,41 @@ func insideUnittest() bool {
...
+
+// VerifyBackendConsistency verifies data in ReadTx and BatchTx are consistent.
+func VerifyBackendConsistency(b Backend, lg *zap.Logger, skipSafeRangeBucket bool, bucket ...Bucket) {
+	verify.Verify(func() {
+		if b == nil {
+			return
+		}
+		if lg != nil {
+			lg.Debug("verifyBackendConsistency", zap.Bool("skipSafeRangeBucket", skipSafeRangeBucket))
+		}
```
ahrtr, line 81: Minor comment: suggest to remove the ugly nil check. We require the callers to pass in valid parameters.
```suggestion
		lg.Debug("verifyBackendConsistency", zap.Bool("skipSafeRangeBucket", skipSafeRangeBucket))
```

server/storage/backend/verify.go
ahrtr: > I was thinking that we should implement a stub for backend that would implement the same methods but without bbolt and keep the data only in memory

Sound like a good idea. We clear/reset the stub backend before each test, and compare the real backend with the stub at the end of each test. But when the member restarts, all the in-memory data will be lost, unless we also support persistent storage for the stub backend.  It's a separate topic (This PR just verifies the consistency between read TXN and write TXN.), please feel free to raise a separate ticket to track it if you want.

server/storage/backend/verify.go
ahrtr: > I don't want to fail those tests.

Which tests are "those tests"?

server/storage/backend/verify.go
serathius: > But when the member restarts, all the in-memory data will be lost, unless we also support persistent storage for the stub backend

You are right that it's impossible to validate cross restart without keeping own persistent storage, however we don't need to test durability, just the runtime behavior. If we validate the data before member is restarted, we should trust the data when backend is loaded back after restart.

server/storage/backend/verify.go
siyuanfoundation: `TestProcessDuplicatedAppRespMessage`

ahrtr: OK. Actually we shouldn't enable & run the verification in unit test at all. They are only supposed to be executed in e2e and integration tests.

We may want to remove the following setting.
https://github.com/etcd-io/etcd/blob/5d45a88ab73cc5461ccf23e4df70598d8d433ff0/scripts/test.sh#L48

Since it's unrelated to this PR, so I won't insist on it. It can be discussed separately.

ahrtr (approved):
lgtm

Thanks

cc @fuweid @jmhbnz @serathius

fuweid (commented):
I think we need to test that performance in the follow-up when uses enables `CorruptCheckTime`. It seems it's too expensive to read all the keys into memory.

ahrtr:
> I think we need to test that performance in the follow-up when uses enables `CorruptCheckTime`. It seems it's too expensive to read all the keys into memory.

The verification is only executed in test.

serathius (approved):

fuweid:
> The verification is only executed in test.

Thanks for the comment.

fuweid (approved):
LGTM

ahrtr:
I just merged another PR, which updated the go.mod files. Please rebase this PR. Sorry for the inconvenience.

siyuanfoundation:
rebased.

PR #17384, opened 2024-02-06
Remove string dependent error handling in watch and auth

mitake:

If we cannot finish https://github.com/etcd-io/etcd/pull/15253 before 3.6, it's better to drop the string dependent error handling from the main branch.
cc @ahrtr

### DIFF
--- a/client/v3/watch.go
+++ b/client/v3/watch.go
@@ -46,11 +46,6 @@ const (
 	InvalidWatchID = -1
 )
 
-var (
-	errMsgGRPCInvalidAuthToken = v3rpc.ErrGRPCInvalidAuthToken.Error()
-	errMsgGRPCAuthOldRevision  = v3rpc.ErrGRPCAuthOldRevision.Error()
-)
-
 type Event mvccpb.Event
 
 type WatchChan <-chan WatchResponse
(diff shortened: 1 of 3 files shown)


mitake:
It's better to avoid introducing the string dependent behavior in the main branch and 3.6, so reverting it now is safer.
Once people can allocate time, discussing again about https://github.com/etcd-io/etcd/pull/15253 and introducing the flag would be good.

ahrtr:
Thanks @mitake 

To be clearer, this PR is reverting https://github.com/etcd-io/etcd/pull/14935 and partially revert https://github.com/etcd-io/etcd/pull/14322 . 
- Since the server side change has already been backported to 3.5 (related to populating different string for `cancelReason` based on different errors), and it's OK to keep that part unchanged.
- Please also remove `AuthTokenTTL` from `tests/framework/integration/cluster.go`.

Regarding the next step, we have multiple choice,
- Keep it as it'is. In other words, do nothing (probably we need to update the doc). If applications see any issue, they just need to recreate the client.
  - Good side:
      - If Auth isn't enabled, then the issue https://github.com/etcd-io/etcd/issues/12385 doesn't exist. So it has no impact on K8s at all.
  - Bad side
      - It isn't reasonable to request applications to recreate the etcd client
- Adopt solution something like https://github.com/etcd-io/etcd/pull/15253
  - Good side
      - It can fix the issue completely
  - Bad side
      - It adds new field, and complicate the use case of downgrade (e.g. 3.5 -> 3.4)
      - It isn't flexible (as compared to the adding of reasonCode), but we don't see other valid cases in which the client sides will use the reasonCode.


Given our bandwidth and it isn't a critical issue (not even a major issue), prefer to the first solution (keep it as it's)? 
@mitake @serathius @lavacat @liggitt

ahrtr:
> * Please also remove `AuthTokenTTL` from `tests/framework/integration/cluster.go`.

Let me take care of this separately in a followup PR.

ahrtr (approved):
lgtm

cc. @serathius

PR #17426, opened 2024-02-14
tests/robustness: refactors, cleanups and followups

MadhavJivrajani:
This is mostly a follow up to https://github.com/etcd-io/etcd/pull/17352

This PR makes three main changes:
1. Refactors `mergeEventHistories` to operate on `[]PersistedEvents`
2. Adds a robustness test for validating create events that can be run on all traffic (not just K8s traffic)
3. [not a follow up to #17352] adds an option that allows persisting of reports in the case of a successful run (I needed that to regenerate test data in #17352) 

/assign @serathius 

Note: there is still one more follow up listed in https://github.com/etcd-io/etcd/pull/17352#issuecomment-1941122809 that is not part of this PR. I will send that in a separate PR (not fully sure yet what needs to be done).


MadhavJivrajani:
Can we merge this?

PR #14589, opened 2022-10-14
[test] Support dynamically adding & starting new member

ahrtr:
Continue to work on https://github.com/etcd-io/etcd/pull/14404, because there is no response from the original author for a long time. 

We need to resolve this because the other PR https://github.com/etcd-io/etcd/pull/14574 is blocked by this one

cc @spzala @mitake @serathius @ptabor 

cc @veshij @biosvs

PR #13621, opened 2022-01-18
Remove V2 API usage from Integration tests

serathius:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

cc @ptabor 
Required for https://github.com/etcd-io/etcd/pull/13612 and https://github.com/etcd-io/etcd/issues/12913

PR #17437, opened 2024-02-16
[3.5] backport support dynamically adding & starting new member

YaoC:
Backport of PR #14589

Align the `patchArgs` method with that in the main branch since the `initial-cluster-state` flag is added as a default argument. https://github.com/etcd-io/etcd/pull/17123#discussion_r1437513485


YaoC:
/cc @ahrtr @serathius 
PTAL, thanks!

ahrtr:
I am a little confused, the comment https://github.com/etcd-io/etcd/pull/17123#discussion_r1437546879 is suggesting to backport https://github.com/etcd-io/etcd/pull/16707, but this PR is backporting https://github.com/etcd-io/etcd/pull/14589. What did I miss?

YaoC:
> I am a little confused, the comment [#17123 (comment)](https://github.com/etcd-io/etcd/pull/17123#discussion_r1437546879) is suggesting to backport #16707, but this PR is backporting #14589. What did I miss?

Backport #16707 may not work here, since [func values](https://github.com/etcd-io/etcd/blob/main/tests/framework/e2e/cluster.go#L638) does not return the flag `--initial-cluster-state`, which is not in [func AddFlags](https://github.com/etcd-io/etcd/blob/main/server/embed/config.go#L561). While #14589 explicitly specifies the flag `--initial-cluster-state`.

tests/framework/e2e/cluster.go
```diff
@@ -253,173 +271,173 @@ func (cfg *EtcdProcessClusterConfig) PeerScheme() string {
...
-				panic("Can't use peer proxy without peer TLS as it can result in malformed packets")
-			}
-			peerAdvertiseUrl.Host = fmt.Sprintf("localhost:%d", peer2Port)
-			proxyCfg = &proxy.ServerConfig{
-				Logger: zap.NewNop(),
-				To:     purl,
-				From:   peerAdvertiseUrl,
-			}
-		}
+func (cfg *EtcdServerProcessConfig) SetInitialCluster(nodes []string, initialClusterState string) {
```
ahrtr, line 293: The implementation seems to be different from the main branch. There is no `SetInitialCluster`. Instead it's `SetInitialOrDiscovery`.

Also the implementation of `EtcdServerProcessConfig` is also a little different from the main branch.

YaoC: My understanding is that the PR #14589 is to extract the part originally in the for-loop as a new method `EtcdServerProcessConfig `. Then, using this method, we can create a config and dynamically add new members to the cluster. As for the implementation itself, it needs to backport some other PRs to align.

`SetInitialOrDiscovery` is also the same reason, `discovery` was introduced by https://github.com/etcd-io/etcd/pull/13621/commits/a53074542bdd62e139fc2c268e17c77caf16ef45, so I removed discovery from the method.

I can pause this PR first and align the content of `EtcdServerProcessConfig` and  `discovery`. What do you think?

ahrtr: > I can pause this PR first

OK



> align the content of `EtcdServerProcessConfig` and `discovery`.

Not fully understood your point. but happy to have more discuss under your new PR.

ahrtr:
Please feel free to raise a new PR if you have bandwidth and are still interested.

PR #17441, opened 2024-02-17
server/verify: use retErr to shadow local err

fuweid:
validateConsistentIndex function might return err but defer might still log it's successful.
Using `retErr` return value is to shadow local error so that `retErr` is always correct.


ahrtr:
> validateConsistentIndex function might return err but defer might still log it's successful.

Good catch!

ahrtr (approved):

PR #17450, opened 2024-02-19
tests/robustness: check for compaction before prevKV validation

MadhavJivrajani:
Follow up from https://github.com/etcd-io/etcd/pull/17352#issuecomment-1941122809

We can check for the condition that Kubernetes checks for, i.e. prevKV can be nil iff the event is not a create a event, only if we know whether compaction has occured or not. If compaction has occured, prevKV can be nil and that is completely valid.

This commit checks if compaction took place during the test run by querying the `/metrics` endpoint. Based on if compaction occured, we now check the Kubernetes condition in the prevKV robustness test.

/assign @serathius

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -16,6 +16,7 @@ package robustness
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 
@@ -23,6 +24,7 @@ import (
 	"go.uber.org/zap/zaptest"
 	"golang.org/x/sync/errgroup"
 
+	"go.etcd.io/etcd/pkg/v3/expect"
 	"go.etcd.io/etcd/tests/v3/framework"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 	"go.etcd.io/etcd/tests/v3/robustness/failpoint"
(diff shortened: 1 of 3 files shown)


MadhavJivrajani:
From a local run, if I change the failpoint to be one of the compaction failpoints:
```
/usr/bin/curl (curl_-L_http://localhost:20000/metrics) (84334): # HELP etcd_debugging_mvcc_db_compaction_keys_total Total number of db keys compacted.
/usr/bin/curl (curl_-L_http://localhost:20000/metrics) (84334): # TYPE etcd_debugging_mvcc_db_compaction_keys_total counter
/usr/bin/curl (curl_-L_http://localhost:20000/metrics) (84334): etcd_debugging_mvcc_db_compaction_keys_total 69
```

tests/robustness/main_test.go
```diff
@@ -158,3 +167,30 @@ func forcestopCluster(clus *e2e.EtcdProcessCluster) error {
...
+	}
+
+	err := e2e.CURLGet(clus, req)
+	if err == nil {
+		return true, nil
+	}
+
+	// If no match was found, compaction did not take place.
+	noCompaction := strings.Contains(err.Error(), "match not found.") ||
+		strings.Contains(err.Error(), "context deadline exceeded")
```
MadhavJivrajani, line 190: This feels odd to do - maybe its worth using `e2e.SpawnWithExpectsContext` and checking the len of the returned lines instead?

MadhavJivrajani:
Also I just realised that the Ci logs are going to be flooded with the output of `/metrics` now.
Another alternative could be to have something like the recording client but for the client being used for the `Trigger()` implementations?

serathius:
> Another alternative could be to have something like the recording client but for the client being used for the Trigger() implementations?

Yes, I was planning to suggest something like this. Add a `Compact` method to recording client, add Compact request type to etcdOperations, and scan for compact operations during the verify. This is the final approach we would like to use when we implement compaction in K8s client.

Still, I think that your solution is a good enough for now. What do you think?

tests/robustness/main_test.go
MadhavJivrajani: Ah that makes sense

However, one issue I see with this is if the value of the metric is 0 vs is curl didn't succeed, I can't really differentiate there. If I had the output of `SpawnWithExpectsContext`, I could maybe error out when it makes sense.

MadhavJivrajani:
> > Another alternative could be to have something like the recording client but for the client being used for the Trigger() implementations?
> 
> Yes, I was planning to suggest something like this. Add a `Compact` method to recording client, add Compact request type to etcdOperations, and scan for compact operations during the verify. This is the final approach we would like to use when we implement compaction in K8s client.
> 
> Still, I think that your solution is a good enough for now. What do you think?

That makes sense - I can work on this as we build up to having compaction as part of the client itself. We can go ahead with the current approach for now!

Only one consideration: https://github.com/etcd-io/etcd/pull/17450#discussion_r1494819397
I can work on this tomorrow if I didn't miss anything.

tests/robustness/main_test.go
serathius: That's a good point, there is a chance that we metric is changed and as result we run the validator. However, the resulting validation will be more strict as not detecting compaction will cause prevKV to be required. This leaves possibility of false positives, however I think this is acceptable for short term solution.

serathius (approved):

MadhavJivrajani:
Ah
I guess that makes sense - thank you for opening the revert @serathius!

I believe the increase in time is because `e2e.Curl` waits for a timeout to occur. I'll start working on a recording client here so that we can do it the right way.

Issue #17455, opened 2024-02-19
[Robustness] Etcd v3.4 required more than expected leader elections


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

We got 2 consecutive robustness test failures on v3.4

https://github.com/etcd-io/etcd/actions/runs/7928589548/job/21647130490
https://github.com/etcd-io/etcd/actions/runs/7940776245/job/21682419728

The cause was low average traffic:
```
2024-02-16T11:00:05.7694187Z     logger.go:130: 2024-02-16T11:00:05.557Z	INFO	Recorded operations	{"operationCount": 809}
2024-02-16T11:00:05.7695071Z     logger.go:130: 2024-02-16T11:00:05.557Z	INFO	Average traffic	{"qps": 59.03054745137809}
2024-02-16T11:00:05.7695949Z     traffic.go:105: Requiring minimal 100.000000 qps for test results to be reliable, got 59.030547 qps
```

The reason for traffic was that cluster was unavailable for longer then normally and required multiple leader elections. 

### What did you expect to happen?

Cluster should recover quickly after member crash and not require multiple leader elections.

### How can we reproduce it (as minimally and precisely as possible)?

No repro yet

### Anything else we need to know?

Expect that this might be due to v3.4 not using --pre-vote, which is default in v3.5.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```Shell
(member-0) (19805): 
(member-0) (19805): goroutine 204 [running]:
[...]

PR #17463, opened 2024-02-20
Force setting pre-vote flag even if it uses default value.

joshuazh-x:

Ref #17455


serathius:
Not sure about the proposed approach, we don't want to set `--pre-vote` everywhere, just in robustness test. Please consider adding it to `baseOptions` https://github.com/etcd-io/etcd/blob/5c311d5cabbbaf80c9449c11c79ffede8b1024cc/tests/robustness/scenarios.go#L75-L81

joshuazh-x:
> Not sure about the proposed approach, we don't want to set `--pre-vote` everywhere, just in robustness test. Please consider adding it to `baseOptions`
> 
> https://github.com/etcd-io/etcd/blob/5c311d5cabbbaf80c9449c11c79ffede8b1024cc/tests/robustness/scenarios.go#L75-L81

Simply adding an option to change prevote does not work, because any option same as the default one will be ignored. 
Set --pre-vote with default value is equivalent to skipping it so I believe it won't cause any problem.

tests/framework/e2e/cluster.go
```diff
@@ -571,6 +571,9 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 	}
 
 	defaultValues := values(*embed.NewConfig())
+	// The default pre-vote config in previous versions may not equal to that in current version.
+	// Clear prevote's default value so cluster always add pre-vote flag even if it uses default value.
+	defaultValues["pre-vote"] = ""
```
serathius, line 576: Could you add a check that just checks the etcd binary version and clears default just for v3.4?

joshuazh-x: Refine the fix to apply only to versions prior to 3.5. Also add an ExplicitFlags config in case we encounter similar issues in future.

serathius (approved):

joshuazh-x:
Fix goword check failure.

serathius:
cc @ahrtr

tests/framework/e2e/cluster.go
```diff
@@ -172,6 +172,9 @@ type EtcdProcessClusterConfig struct {
 	IsPeerAutoTLS      bool
 	CN                 bool
 
+	// ExplicitFlags contains flags to be explicitly set even if their values are same as default ones
+	ExplicitFlags map[string]struct{}
```
serathius, line 176: Long term we might consider hardcoding differences between defaults on previous releases, however this should be a good solution for now.

ahrtr:
Pls see my comment https://github.com/etcd-io/etcd/issues/17455#issuecomment-1956381800

Issue #16963, opened 2023-11-17
Test go.etcd.io/etcd/server/v3/storage/mvcc: TestHashKVWhenCompacting is flaking with timeout


srivastav-abhishek:
### Which github workflows are flaking?

[Failed-test-TestHashKVWhenCompacting-with-master-golang](https://prow.ppc64le-cloud.cis.ibm.net/view/s3/ppc64le-prow-logs/logs/postsubmit-master-golang-etcd-build-test-ppc64le/1725287890176446464)

### Which tests are flaking?

```go.etcd.io/etcd/server/v3/storage/mvcc: TestHashKVWhenCompacting ```

### Reason for failure (if possible)

It is failing due to timeout as ```compaction``` and ```storing new hash``` operations are not getting completed within 10 secs.

### Anything else we need to know?

This is observed on ```ppc64le``` arch.

ahrtr:
FYI. I recently updated this test case in https://github.com/etcd-io/etcd/pull/17313

Let's reopen this issue if we see this issue again in future

MadhavJivrajani:
Seeing this again: https://github.com/etcd-io/etcd/actions/runs/8002360176/job/21855397745?pr=17475

ahrtr:
We recently saw test timed out on arm many times, not only this case :(

srivastav-abhishek:
We are seeing test timed out on ppc arch as well,

https://prow.ppc64le-cloud.cis.ibm.net/view/gs/ppc64le-kubernetes/logs/periodic-etcd-test-ppc64le/1760756515901280256

ahrtr:
Thanks for all the feedback.

One of the reason should be that there were huge number of debug log messages. I just checked the logs, and the test case couldn't finish 100 times of compactions in 10s. I suggest to increase the timeout to `20s`. Could anyone raise a PR for that? thx

https://github.com/etcd-io/etcd/blob/8c7f911b3b8629d9c0311d2dca641e2243a6b45c/server/storage/mvcc/kvstore_test.go#L639

srivastav-abhishek:
@ahrtr I have associated this [PR](https://github.com/etcd-io/etcd/pull/17481) could you ptal ?

ahrtr:
> @ahrtr I have associated this [PR](https://github.com/etcd-io/etcd/pull/17481) could you ptal ?

thx, added a comment for that PR

PR #17475, opened 2024-02-22
tests/robustness: use WithRequireLeader in Kubernetes traffic

MadhavJivrajani:
Kubernetes uses WithRequireLeader to modify the context passed to the Watch() and RequestProgress() calls in order to ensure that a leader is present in the cluster before serving the request.

This commit mimics that behaviour in the Kubernetes traffic.

Towards: https://github.com/etcd-io/etcd/issues/15820

/assign @serathius


MadhavJivrajani:
Hmm - I'm also thinking about how we can verify the error returned to be `ErrNoLeader` in case of a partitioned node (this is probably something to test in k/k tho). I'm not sure we can test that in robustness tests because we mostly prune the errored responses iirc.

MadhavJivrajani:
The failure looks like https://github.com/etcd-io/etcd/issues/16963
/retest

serathius (approved):

Issue #17487, opened 2024-02-25
`TestPeriodicSkipRevNotChange` failed


ahrtr:
### Which Github Action / Prow Jobs are flaking?

pull-etcd-unit-test

### Which tests are flaking?

TestPeriodicSkipRevNotChange

### Github Action / Prow Job link

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/17486/pull-etcd-unit-test/1761691393031409664

### Reason for failure (if possible)

```
{Failed  === RUN   TestPeriodicSkipRevNotChange
    logger.go:130: 2024-02-25T09:58:05.625Z	INFO	starting auto periodic compaction	{"revision": 101, "compact-period": "5m0s"}
    logger.go:130: 2024-02-25T09:58:05.636Z	INFO	completed auto periodic compaction	{"revision": 101, "compact-period": "5m0s", "took": "30s"}
    logger.go:130: 2024-02-25T09:58:06.574Z	INFO	starting auto periodic compaction	{"revision": 102, "compact-period": "5m0s"}
    periodic_test.go:233: len(actions) = 0, expected >= 1
--- FAIL: TestPeriodicSkipRevNotChange (1.15s)
}
```

Rajalakshmi-Girish:
@ahrtr seems like related to https://github.com/etcd-io/etcd/issues/17054

ghouscht:
I assume this is not yet fixed, right? At least for me this test has a high fail percentage when I run the test with `stress` (~2.4% across 5k runs).

Increasing the timeout from:
https://github.com/etcd-io/etcd/blob/2c971109267d9d29ef7cb901148931bdbd55762a/server/etcdserver/api/v3compactor/periodic_test.go#L183

to the following values yields:

- 15ms - 0.1% failures (5k runs)
- 20ms - 0 failures (5k runs)

Anyway the 10ms timeout seems like an arbitrary value to me, or is there a reason it is set to 10ms? If it is an arbitrary value we could as well increase it to 20ms to make the test stable or remove the timeout?

ahrtr:
> * 20ms - 0 failures (5k runs)

Thanks. I do not get time digging into the test case. But I suppose that just changing the timeout from 10ms to 20ms won't change the logic of the test case. So no objection to increase the timeout to 20ms.

ghouscht:
> > * 20ms - 0 failures (5k runs)
> 
[...]

Issue #17054, opened 2023-12-03
Flaky `TestPeriodicHourly` and `TestPeriodicMinutes`


ahrtr:
### Which github workflows are flaking?

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/16822/pull-etcd-unit-test/1731327844841164800

```
{Failed  === RUN   TestPeriodicHourly
    logger.go:130: 2023-12-03T15:04:01.795Z	INFO	starting auto periodic compaction	{"revision": 1, "compact-period": "2h0m0s"}
    logger.go:130: 2023-12-03T15:04:01.806Z	INFO	completed auto periodic compaction	{"revision": 1, "compact-period": "2h0m0s", "took": "6m0s"}
    logger.go:130: 2023-12-03T15:04:01.975Z	INFO	starting auto periodic compaction	{"revision": 9, "compact-period": "2h0m0s"}
    logger.go:130: 2023-12-03T15:04:01.975Z	INFO	completed auto periodic compaction	{"revision": 9, "compact-period": "2h0m0s", "took": "0s"}
    periodic_test.go:77: compact request = revision:9 , want revision:10 
    logger.go:130: 2023-12-03T15:04:02.154Z	INFO	starting auto periodic compaction	{"revision": 19, "compact-period": "2h0m0s"}
    logger.go:130: 2023-12-03T15:04:02.154Z	INFO	completed auto periodic compaction	{"revision": 19, "compact-period": "2h0m0s", "took": "0s"}
    periodic_test.go:77: compact request = revision:19 , want revision:20 
    logger.go:130: 2023-12-03T15:04:02.375Z	INFO	starting auto periodic compaction	{"revision": 29, "compact-period": "2h0m0s"}
    logger.go:130: 2023-12-03T15:04:02.375Z	INFO	completed auto periodic compaction	{"revision": 29, "compact-period": "2h0m0s", "took": "0s"}
    periodic_test.go:77: compact request = revision:29 , want revision:30 
--- FAIL: TestPeriodicHourly (0.90s)
```

```
{Failed  === RUN   TestPeriodicMinutes
    logger.go:130: 2023-12-03T15:04:02.606Z	INFO	starting auto periodic compaction	{"revision": 1, "compact-period": "5m0s"}
    periodic_test.go:105: len(actions) = 0, expected >= 1
--- FAIL: TestPeriodicMinutes (0.27s)
```

### Which tests are flaking?

[...]

PR #17513, opened 2024-03-01
Use strict synchronization for revision getter to minimize flaky result caused by time racing.

joshuazh-x:

for issue #17054 and #17487

### DIFF
--- a/client/pkg/testutil/recorder.go
+++ b/client/pkg/testutil/recorder.go
@@ -17,6 +17,7 @@ package testutil
 import (
 	"errors"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 )
@@ -115,7 +116,10 @@ func (r *recorderStream) Chan() <-chan Action {
 
 func (r *recorderStream) Wait(n int) ([]Action, error) {
 	acts := make([]Action, n)
-	timeoutC := time.After(r.waitTimeout)
+	var timeoutC <-chan time.Time
+	if r.waitTimeout < math.MaxInt64 {
+		timeoutC = time.After(r.waitTimeout)
+	}
 	for i := 0; i < n; i++ {
 		select {
 		case acts[i] = <-r.ch:
--- a/server/etcdserver/api/v3compactor/periodic_test.go
+++ b/server/etcdserver/api/v3compactor/periodic_test.go
@@ -16,6 +16,7 @@ package v3compactor
 
 import (
 	"errors"
+	"math"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


server/etcdserver/api/v3compactor/periodic_test.go
```diff
@@ -189,7 +190,7 @@ func TestPeriodicSkipRevNotChange(t *testing.T) {
 	initialIntervals, intervalsPerPeriod := tb.getRetentions(), 10
 
 	// first compaction happens til 5 minutes elapsed
-	for i := 0; i < initialIntervals; i++ {
+	for i := 0; i < initialIntervals-1; i++ {
```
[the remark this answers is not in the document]
joshuazh-x: initialIntervals is 11 while 10 is enough to trigger initial compact. In original test, the last rg.Wait(1) in initial loop always timeout. I assume this is not needed so I removed it. Otherwise the test will stuck in the last rg.Wait(1) in initial loop with this fix.

server/etcdserver/api/v3compactor/periodic_test.go
```diff
@@ -43,7 +44,7 @@ func TestPeriodicHourly(t *testing.T) {
 	initialIntervals, intervalsPerPeriod := tb.getRetentions(), 10
 
 	// compaction doesn't happen til 2 hours elapse
-	for i := 0; i < initialIntervals; i++ {
+	for i := 0; i < initialIntervals-1; i++ {
 		rg.Wait(1)
```
ahrtr, line 48: I think we also need to check the length of the returned value of `rg.Wait(1)`. It's possible that it may return a slice with 2 elements. The number of loop should be equal to the sum of the number of returned elements.

The logic should be something like below,
```
cnt := 0
for cnt <= initialIntervals-1 {
     cnt += rg.Wait(1)
     ....
}

```

This comment applies to other places as well.

ahrtr: > Check every rg.Wait(1) to ensure it returns exact 1 action.

It seems `rg.Wait(1)` returning 2 isn't an issue. My previous suggestion `cnt += rg.Wait(1)` makes more sense?

ahrtr:
@joshuazh-x could you rebase this PR?

client/pkg/testutil/recorder.go
```diff
@@ -115,7 +116,10 @@ func (r *recorderStream) Chan() <-chan Action {
 
 func (r *recorderStream) Wait(n int) ([]Action, error) {
 	acts := make([]Action, n)
-	timeoutC := time.After(r.waitTimeout)
+	var timeoutC <-chan time.Time
+	if r.waitTimeout < math.MaxInt64 {
+		timeoutC = time.After(r.waitTimeout)
+	}
```
ahrtr, line 122: Suggest to follow the pattern "`A Timeout of zero means no timeout.`"

FYI.
https://github.com/golang/go/blob/c2c4a32f9e57ac9f7102deeba8273bcd2b205d3c/src/net/http/client.go#L96

srivastav-abhishek: @joshuazh-x  @ahrtr  I'd like to address this comment as we are seeing `TestPeriodic` tests flaking frequently in our CI. Please let me know if I can proceed or if there are any other plans to take it forward.

ahrtr: If there is no feedback from @joshuazh-x next week, please feel free to create a new PR. thx

srivastav-abhishek: @ahrtr  I am creating a new PR then with similar changes and review comments addressed. TY !

ahrtr: OK, thanks

srivastav-abhishek: I have cherry-picked the change and addressed review comments here, #18608 . Could you please take a look ?
thanks !

ahrtr:
Superseded by https://github.com/etcd-io/etcd/pull/18608

PR #17518, opened 2024-03-04
etcdserver: skip when detect a removed peer

### LINKED ISSUE #17514: When process messages to send to peers, leader should skip checking message type when detect removed peer.
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.

### What happened?

From reading code, in `processMessages`, if a peer has been removed, the leader should skip checking the message type.
https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/server/etcdserver/raft.go#L345-L350

Otherwise, when the message type is `raftpb.MsgHeartbeat`, registering `id = 0` in `r.td` is unintended, because `r.td` should only observe peer's ID greater than zero. This might lead to incorrectly increase `heartbeatSendFailures`.
https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/server/etcdserver/raft.go#L372-L375

`r.td` contains peer's ID greater than zero.

It might be hard to reproduce. It can happen when leader apply config change remove node and then process message detects node removed.

_No response_

main branch

<details>

</details>

<details>

```console
# paste output here

# paste output here
```

</details>

### Relevant log output

_No response_

datbeohbbh:
resolve #17514

### DIFF
--- a/server/etcdserver/raft.go
+++ b/server/etcdserver/raft.go
@@ -347,6 +347,7 @@ func (r *raftNode) processMessages(ms []raftpb.Message) []raftpb.Message {
 	for i := len(ms) - 1; i >= 0; i-- {
 		if r.isIDRemoved(ms[i].To) {
 			ms[i].To = 0
+			continue
 		}
 
 		if ms[i].Type == raftpb.MsgAppResp {


ahrtr (approved):

fuweid (approved):
LGTM

serathius:
Can we add a test? Any test.

ahrtr:
> Can we add a test? Any test.

It should be a very safe & very minor change change. We don't have to necessarily add a test for this PR.

Also we already have some membership related e2e and integration test cases. I think it's OK.

serathius:
For me every bug that is worth backporting is also worth testing against.

ahrtr:
I do not see how to reproduce this, so actually this PR fixed an non-exist issue. The fix just looks more natural (and of course safe and very minor) so we accept it.

There is no point to waste time to get stuck on this. If you insist on adding a test, then provide detailed guide on how to do it. Or just close the backporting PRs or even revert the fix on main branch. I am going to waste time on this, leave it to you to take care of it since you have concern on it.

ahrtr:
> I didn't expect that writing a unit test needs a detailed guideline. Please let me know what would make writing unit test harder in etcd so we need guideline for this?

Unit test doesn't always fit for each kind of fix. Understanding the root cause is the prerequisite, otherwise it doesn't make sense to request to add a test.

PR #17531, opened 2024-03-05
[3.5] backport mix version e2e test.

siyuanfoundation:

Part of https://github.com/etcd-io/etcd/issues/15878#issuecomment-1910886682

Original PR: https://github.com/etcd-io/etcd/pull/14707 and https://github.com/etcd-io/etcd/pull/15013

Slightly different from the tests in main in that we cannot verify a snapshot is sent because there is no `snapshot-catchup-entries` flag in 3.5.

### DIFF
--- a/tests/e2e/etcd_mix_versions_test.go
+++ b/tests/e2e/etcd_mix_versions_test.go
@@ -1,0 +1,194 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+type clusterTestCase struct {
+	name   string
+	config e2e.EtcdProcessClusterConfig
+}
+
+func clusterTestCases(size int) []clusterTestCase {
+	tcs := []clusterTestCase{
+		{
+			name:   "CurrentVersion",
+			config: e2e.EtcdProcessClusterConfig{ClusterSize: size},
+		},
+	}
+	if !fileutil.Exist(e2e.BinPathLastRelease) {
+		return tcs
+	}
+
+	tcs = append(tcs,
+		clusterTestCase{
+			name:   "LastVersion",
+			config: e2e.EtcdProcessClusterConfig{ClusterSize: size, Version: e2e.LastVersion},
+		},
+	)
+	if size > 2 {
+		tcs = append(tcs,
+			clusterTestCase{
+				name:   "MinorityLastVersion",
+				config: e2e.EtcdProcessClusterConfig{ClusterSize: size, Version: e2e.MinorityLastVersion},
+			}, clusterTestCase{
+				name:   "QuorumLastVersion",
+				config: e2e.EtcdProcessClusterConfig{ClusterSize: size, Version: e2e.QuorumLastVersion},
+			},
+		)
+	}
+	return tcs
+}
+
+// TestMixVersionsSnapshotByAddingMember tests the mix version send snapshots by adding member
+func TestMixVersionsSnapshotByAddingMember(t *testing.T) {
+	for _, tc := range clusterTestCases(1) {
+		t.Run(tc.name+"-adding-new-member-of-current-version", func(t *testing.T) {
+			mixVersionsSnapshotTestByAddingMember(t, tc.config, e2e.CurrentVersion)
+		})
+		// TODO: uncomment after v3.4.32 release.
+		// After v3.4.32, etcd can support adding a new member of 3.4 into
+		// a cluster with 3.5 if the 3.4 member is started with '--next-cluster-version-compatible'.
+
+		// t.Run(tc.name+"-adding-new-member-of-last-version", func(t *testing.T) {
+		// 	mixVersionsSnapshotTestByAddingMember(t, tc.config, e2e.LastVersion)
+		// })
+	}
+}
+
+func mixVersionsSnapshotTestByAddingMember(t *testing.T, cfg e2e.EtcdProcessClusterConfig, newInstanceVersion e2e.ClusterVersion) {
+	e2e.BeforeTest(t)
+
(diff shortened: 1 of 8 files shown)


siyuanfoundation:
cc @serathius

ahrtr:
Please link to the original PR next time when you backport any PR.

- https://github.com/etcd-io/etcd/pull/14707
- https://github.com/etcd-io/etcd/pull/15013

ahrtr (approved):
/lgtm


cc @fuweid @jmhbnz @serathius double check

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func mixVersionsSnapshotTestByAddingMember(t *testing.T, clusterVersion, newInstanceVersion e2e.ClusterVersion) {
+	e2e.BeforeTest(t)
+
+	if !fileutil.Exist(e2e.BinPathLastRelease) {
+		t.Skipf("%q does not exist", e2e.BinPathLastRelease)
+	}
+
+	e2e.BeforeTest(t)
```
fuweid, line 63: duplicate line with L57

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,214 @@
...
+	copiedCfg.Version = clusterVersion
+	copiedCfg.SnapshotCount = 10
+	copiedCfg.BasePeerScheme = "unix" // to avoid port conflict
+
+	epc, err := e2e.NewEtcdProcessCluster(t, copiedCfg)
+
+	require.NoError(t, err, "failed to start etcd cluster: %v", err)
+	defer func() {
+		derr := epc.Close()
+		require.NoError(t, derr, "failed to close etcd cluster: %v", derr)
```
fuweid, line 77: require.NoError(t, epx.Close(), "failed to close etcd cluster") might be cleaner.

tests/framework/e2e/config.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 The etcd Authors
```
fuweid, line 1: s/2022/2024/?

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+// TestMixVersionsSnapshotByAddingMember tests the mix version send snapshots by adding member
+func TestMixVersionsSnapshotByAddingMember(t *testing.T) {
+	cases := []struct {
+		name               string
+		clusterVersion     e2e.ClusterVersion
+		newInstanceVersion e2e.ClusterVersion
+	}{
+		// After v3.4.32, etcd can support adding a new member of 3.4 into
```
fuweid, line 35: we can add TODO here (IMO, it's easy to spot TODO:))

tests/e2e/etcd_mix_versions_test.go
siyuanfoundation: If using `require.NoError(t, epx.Close(), "failed to close etcd cluster")`, we would not know what error is returned.

fuweid: If epc.Close() is not nil, require.NoError will panic and show us.
Never mind. Just share that we don't need to pass error value in message.

fuweid (approved):
LGTM

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,213 @@
...
+		derr := epc.Close()
+		require.NoError(t, derr, "failed to close etcd cluster: %v", derr)
+	}()
+	toPartitionedMember := epc.Procs[mockPartitionNodeIndex]
+
+	// Stop and restart the partitioned member
+	err = toPartitionedMember.Stop()
+	require.NoError(t, err)
+
+	// Write more than SnapshotCount entries to trigger at least a snapshot.
```
serathius, line 168: Recommend to change comments to just more verbose logs.

serathius:
Testing snapshot exchange between members in mixed version cluster is super useful! Thanks for adding those tests.

My suggestions:
* Add scenario where there is a real network partition. Member downloading snapshot after being restarted is useful too, however I would prefer to also have a real network partition test. E2e test framework already supports separating member communication. You can check out robustness network blackhole failpoint.
* Those tests should also exist between v3.6 and v3.5, I would suggest to first add them to the main branch, then we could backport the e2e framework tools and only at the end backport the test to v3.5.

tests/e2e/etcd_mix_versions_test.go
serathius, line 1: Those tests scenarios are just testing behavior of snapshot sent between members. They would also work in homogeneous cluster, only difference is the cluster setup. 

Could rename the directory to etcd_snapshot_test.go and add also a scenario with cluster version being `CurrentVersion`?

See https://github.com/etcd-io/etcd/blob/547aa643e69d55328b203f40ccec50256b44158e/tests/common/e2e_test.go#L55-L73 for how I generalized it for any test.

tests/e2e/etcd_mix_versions_test.go
```diff
@@ -0,0 +1,213 @@
...
+	newCfg.Version = newInstanceVersion
+	t.Log("Starting a new etcd instance")
+	_, err = epc.StartNewProc(&newCfg, t)
+	require.NoError(t, err, "failed to start the new etcd instance: %v", err)
+	defer epc.Close()
+
+	// verify all nodes have exact same revision and hash
+	t.Log("Verify all nodes have exact same revision and hash")
+	assert.Eventually(t, func() bool {
+		hashKvs, err := e2e.NewEtcdctl(epc.EndpointsV3(), e2e.ClientNonTLS, false, false).HashKV(0)
```
serathius, line 101: I don't think this validation is very useful. I expect this check to pass even without fixes to migrating the snapshot. 

`HashKV` only checks the contents of Key bucket, ignoring all the other metadata. As the layout of key bucket didn't change at between v3.5 and v3.4, this doesn't really check if data was properly migrated. 

Options:
* Use `Hash` that returns hash of whole etcd storage, but skips some keys from metadata. Not sure it will work across etcd version.
* Check other property to verify that data was properly migrated.

siyuanfoundation: Added https://github.com/etcd-io/etcd/issues/17752 as followup. But I don't think that should necessarily block this PR.

tests/e2e/etcd_mix_versions_test.go
siyuanfoundation: See line 33 of tests/e2e/etcd_mix_versions_test.go

siyuanfoundation:
> Testing snapshot exchange between members in mixed version cluster is super useful! Thanks for adding those tests.
> 
> My suggestions:
> 
> * Add scenario where there is a real network partition. Member downloading snapshot after being restarted is useful too, however I would prefer to also have a real network partition test. E2e test framework already supports separating member communication. You can check out robustness network blackhole failpoint.
> * Those tests should also exist between v3.6 and v3.5, I would suggest to first add them to the main branch, then we could backport the e2e framework tools and only at the end backport the test to v3.5.

The blackhole failpoint is problematic right now https://github.com/etcd-io/etcd/issues/17737

tests/e2e/etcd_mix_versions_test.go
serathius: SG

serathius (approved):

PR #17543, opened 2024-03-06
[2024-03-04] Bump dependencies identified by dependabot

ivanvc:
This pull request completes this week's etcd dependency updates following our [dependency roster](https://docs.google.com/spreadsheets/d/1jodHIO7Dk2VWTs1IRnfMFaRktS9IH8XRyifOnPdSY8I/edit#gid=1394774387) and [dependency management instructions](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md).

Summary of actions:

- [x] google.golang.org/protobuf bump from 1.32.0 to 1.33.0, addresses CVE-2024-24786. Wasn't catch by dependabot, but by Go vulnerability CI check
- [x] https://github.com/etcd-io/etcd/pull/17527 - Bumped
- [x] https://github.com/etcd-io/etcd/pull/17519 - Bumped
- [x] https://github.com/etcd-io/etcd/pull/17525 - Bumped
- [ ] https://github.com/etcd-io/etcd/pull/17523 - Indirect
- [ ] https://github.com/etcd-io/etcd/pull/17522 - Indirect
- [ ] https://github.com/etcd-io/etcd/pull/17521 - Indirect
- [ ] https://github.com/etcd-io/etcd/pull/17520 - Indirect

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #17544, opened 2024-03-07
Tweak golang version policy to avoid unsupported minor versions

jmhbnz:
Our recent golang update pr's got me thinking about our version policy once again.

Re-reading our current agreed policy I noticed the wording made it possible for us to have periods of time when our stable release branches could potentially be promoting an unsupported version of go.

This pull requests proposes updating our policy so that rather than waiting to upgrade until the current version is no longer supported, we should instead get started on the update a bit sooner.

---

### Example

We currently promote golang version `1.21` as the minimum version to build our stable releases. The next upgrade for these branches will be to golang `1.22`. 

| Old policy         | New policy |
|--------------|:-----:|
| Wait until golang `1.23` is released before upgrading. Meaning we could have a period of time where our stable release branches are promoting an unsupported golang version (depending on how long it takes us to get around to do the update)|  Start the update before `1.23` releases (perhaps 1-3 months out) so we can ensure no period of promoting unsupported version |

### DIFF
--- a/Documentation/contributor-guide/dependency_management.md
+++ b/Documentation/contributor-guide/dependency_management.md
@@ -120,7 +120,7 @@ Suggested steps for performing a minor version upgrade for the etcd development 
 4. Run performance benchmarks locally to compare before and after.
 5. Raise a pull request for the changes, example <https://github.com/etcd-io/etcd/pull/16394>.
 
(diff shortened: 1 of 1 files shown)


fuweid (approved):
SGTM

ahrtr:
Either way works for me. 1 ~ 3 months delay on an old go version isn't a big deal to me as long as there is no any major CVE (>=7.5, actually we can even release a patch immediately in that case)

Once https://github.com/etcd-io/etcd/pull/17543 gets merged, then please rebase this PR. Thanks.

ahrtr (approved):

Issue #17529, opened 2024-03-04
etcd watch events starvation / lost multiplexed on a single watch stream


chaochn47:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### Watch events starvation

When there are hundreds of watch initialized on a key range with `prevKV` option and multiplexed on the same watch stream, the watch response could be delayed as long as several minutes. And the out of sync watcher was not cancelled by providing a [WatchResponse](https://github.com/etcd-io/etcd/blob/release-3.5/server/mvcc/watcher.go#L79) with `CompactRevision` not equal to 0 first. 

It was also observed `etcd_debugging_mvcc_events_total` on one etcd server were accumulated at 

`etcd_debugging_mvcc_pending_events_total <etcd-0> 8.740897e+06` and 

there were 722 slow watchers. 

`etcd_debugging_mvcc_slow_watcher_total <etcd-0> 722`

### Watch events lost
[moveVictims](https://github.com/etcd-io/etcd/blob/release-3.5/server/mvcc/watchable_store.go#L271) is to synchronize slow watchers to send out pending events in the `watchableStore`. The provider of victim watchers are either from synced watcher in [notify](https://github.com/etcd-io/etcd/blob/release-3.5/server/mvcc/watchable_store.go#L439) or [syncWatcher](https://github.com/etcd-io/etcd/blob/release-3.5/server/mvcc/watchable_store.go#L389) background routine due to the [<-sws.watchStream.Chan()](https://github.com/etcd-io/etcd/blob/release-3.5/server/etcdserver/api/v3rpc/watch.go#L412) is clogged and the pending events could not be sent out. The reason of this blocked channel could be the consumer of this channel is slow and cannot kept up with the events generator. 

[...]

PR #17555, opened 2024-03-08
Fix watch event loss

chaochn47:
Fix https://github.com/etcd-io/etcd/issues/17529

@fuweid @serathius @ahrtr 

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -39,6 +39,8 @@ var (
 	maxWatchersPerSync = 512
 )
 
+func ChanBufLen() int { return chanBufLen }
+
 type watchable interface {
 	watch(key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc)
 	progress(w *watcher)
@@ -370,6 +372,11 @@ func (s *watchableStore) syncWatchers() int {
 	victims := make(watcherBatch)
 	wb := newWatcherBatch(wg, evs)
 	for w := range wg.watchers {
+		if w.minRev < compactionRev {
+			// Skip the watcher that failed to send compacted watch response due to w.ch is full.
+			// Next retry of syncWatchers would try to resend the compacted watch response to w.ch
+			continue
+		}
 		w.minRev = curRev + 1
 
 		eb, ok := wb[w]
--- a/server/storage/mvcc/watchable_store_test.go
+++ b/server/storage/mvcc/watchable_store_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
@@ -247,6 +248,63 @@ func TestWatchCompacted(t *testing.T) {
 		}
 	case <-time.After(1 * time.Second):
 		t.Fatalf("failed to receive response (timeout)")
+	}
+}
+
+func TestWatchNoEventLossOnCompact(t *testing.T) {
+	oldChanBufLen, oldMaxWatchersPerSync := chanBufLen, maxWatchersPerSync
+
+	b, _ := betesting.NewDefaultTmpBackend(t)
+	lg := zaptest.NewLogger(t)
+	s := newWatchableStore(lg, b, &lease.FakeLessor{}, StoreConfig{})
+
+	defer func() {
+		cleanup(s, b)
+		chanBufLen, maxWatchersPerSync = oldChanBufLen, oldMaxWatchersPerSync
+	}()
+
+	chanBufLen, maxWatchersPerSync = 1, 4
+	testKey, testValue := []byte("foo"), []byte("bar")
+
+	maxRev := 10
+	compactRev := int64(5)
+	for i := 0; i < maxRev; i++ {
+		s.Put(testKey, testValue, lease.NoLease)
+	}
+	_, err := s.Compact(traceutil.TODO(), compactRev)
+	require.NoErrorf(t, err, "failed to compact kv (%v)", err)
+
+	w := s.NewWatchStream()
+	defer w.Close()
+
+	watchers := map[WatchID]int64{
+		0: 1,
+		1: 1, // create unsyncd watchers with startRev < compactRev
+		2: 6, // create unsyncd watchers with compactRev < startRev < currentRev
+	}
+	for id, startRev := range watchers {
+		_, err := w.Watch(id, testKey, nil, startRev)
+		require.NoError(t, err)
+	}
+	// fill up w.Chan() with 1 buf via 2 compacted watch response
+	s.syncWatchers()
+
+	for len(watchers) > 0 {
+		resp := <-w.Chan()
+		if resp.CompactRevision != 0 {
+			require.Equal(t, resp.CompactRevision, compactRev)
+			require.Contains(t, watchers, resp.WatchID)
+			delete(watchers, resp.WatchID)
+			continue
+		}
+		nextRev := watchers[resp.WatchID]
+		for _, ev := range resp.Events {
+			require.Equalf(t, nextRev, ev.Kv.ModRevision, "got event revision %d but want %d for watcher with watch ID %d", ev.Kv.ModRevision, nextRev, resp.WatchID)
+			nextRev++
+		}
+		if nextRev == s.rev()+1 {
+			delete(watchers, resp.WatchID)
+		}
 	}
 }
 
--- a/tests/integration/v3_watch_test.go
+++ b/tests/integration/v3_watch_test.go
@@ -17,6 +17,7 @@ package integration
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -24,13 +25,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc"
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
 	"go.etcd.io/etcd/tests/v3/framework/integration"
+	gofail "go.etcd.io/gofail/runtime"
 )
 
 // TestV3WatchFromCurrentRevision tests Watch APIs from current revision.
@@ -1511,4 +1516,57 @@ func TestV3WatchProgressWaitsForSyncNoEvents(t *testing.T) {
 		}
 	}
 	require.True(t, gotProgressNotification, "Expected to get progress notification")
+}
+
+// TestV3NoEventsLostOnCompact verifies that slow watchers exit with compacted watch response
+// if its next revision of events are compacted and no lost events sent to client.
+func TestV3NoEventsLostOnCompact(t *testing.T) {
+	if integration.ThroughProxy {
+		t.Skip("grpc proxy currently does not support requesting progress notifications")
+	}
+	integration.BeforeTest(t)
+	if len(gofail.List()) == 0 {
+		t.Skip("please run 'make gofail-enable' before running the test")
+	}
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
+
+	client := clus.RandClient()
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	// sendLoop throughput is rate-limited to 1 event per second
+	require.NoError(t, gofail.Enable("beforeSendWatchResponse", `sleep("1s")`))
+	wch := client.Watch(ctx, "foo")
+
+	var rev int64
+	writeCount := mvcc.ChanBufLen() * 11 / 10
+	for i := 0; i < writeCount; i++ {
+		resp, err := client.Put(ctx, "foo", "bar")
+		require.NoError(t, err)
+		rev = resp.Header.Revision
+	}
+	_, err := client.Compact(ctx, rev)
+	require.NoError(t, err)
+
+	time.Sleep(time.Second)
+	require.NoError(t, gofail.Disable("beforeSendWatchResponse"))
+
+	eventCount := 0
+	compacted := false
+	for resp := range wch {
+		err = resp.Err()
+		if err != nil {
+			if !errors.Is(err, rpctypes.ErrCompacted) {
+				t.Fatalf("want watch response err %v but got %v", rpctypes.ErrCompacted, err)
+			}
+			compacted = true
+			break
+		}
+		eventCount += len(resp.Events)
+		if eventCount == writeCount {
+			break
+		}
+	}
+	assert.Truef(t, compacted, "Expected stream to get compacted, instead we got %d events out of %d events", eventCount, writeCount)
 }
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -449,6 +449,7 @@ func (sws *serverWatchStream) sendLoop() {
 			sws.mu.RUnlock()
 
 			var serr error
+			// gofail: var beforeSendWatchResponse struct{}
 			if !fragmented && !ok {
 				serr = sws.gRPCStream.Send(wr)
 			} else {
--- a/tests/robustness/makefile.mk
+++ b/tests/robustness/makefile.mk
@@ -36,7 +36,7 @@ GOFAIL_VERSION = $(shell cd tools/mod && go list -m -f {{.Version}} go.etcd.io/g
 
 .PHONY: gofail-enable
 gofail-enable: install-gofail
-	gofail enable server/etcdserver/ server/storage/backend/ server/storage/mvcc/ server/storage/wal/
+	gofail enable server/etcdserver/ server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/
 	cd ./server && go get go.etcd.io/gofail@${GOFAIL_VERSION}
 	cd ./etcdutl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
 	cd ./etcdctl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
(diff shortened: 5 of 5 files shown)


chaochn47:
Filed flaky test report https://github.com/etcd-io/etcd/issues/17556, provided root cause and how to fix. 

Is there any way to re-test this specific test `E2E / test (linux-386)`? @jmhbnz

jmhbnz:
/retest

(Will only retest workflows that have failed by default)

tests/e2e/watch_delay_test.go
```diff
@@ -246,3 +253,241 @@ func continuouslyExecuteGetAll(ctx context.Context, t *testing.T, g *errgroup.Gr
...
+		gresp, err := c.Get(rootCtx, "foo")
+		if err != nil {
+			panic(err)
+		}
+		rev := gresp.Header.Revision
+
+		lastEventModifiedRevision := rev
+		watchCacheWatchOpts := append([]clientv3.OpOption{clientv3.WithCreatedNotify(), clientv3.WithRev(rev), clientv3.WithProgressNotify()}, commonWatchOpts...)
+		for wres := range c.Watch(rootCtx, watchKeyPrefix, watchCacheWatchOpts...) {
+			if wres.Err() != nil {
```
fuweid, line 292: maybe we can check the error and ensure it's ErrCompacted.

fuweid (approved):
LGTM

The change is small and it's easy to backport. Thanks

serathius:
Agree about the e2e tests with @ahrtr, also we should have a robustness test for this scenario, won't block the PR, but it's important. If I have time I will try to implement them myself.

server/storage/mvcc/watchable_store.go
```diff
@@ -370,6 +370,12 @@ func (s *watchableStore) syncWatchers() int {
 	victims := make(watcherBatch)
 	wb := newWatcherBatch(wg, evs)
 	for w := range wg.watchers {
+		if w.minRev < compactionRev {
+			// skip the watcher that failed to send compacted watch response due to w.ch is full
+			// next retry of syncWatchers would try to resend the compacted watch response to w.ch
+			// TODO prioritize sending compacted watch response over other watch responses with events.
+			continue
+		}
```
serathius, line 378: I think we make the fix better if we move it up higher. Let's group stages of this code:
* Line 354: Pick unsynched watchers that we want to sync, and their minRev which should be higher than compactionRev. It might also drop compacted  watchers.
* Lines 355-368:  Read events between minRev and curRev
* Line 371: Match watchers to events they should receive
* Lines 372+: Pass events to watchers

By implementing fix on line 373, you are skipping sync of watch that we already matched events too. 

Instead I recommend to change the Line 354 to not return compacted watches, even if `w.ch` is full. Don't know if you tried it already, but on first try it will not work properly, why? Because the `.choose` function is written badly. It sometimes mutates the original watchGroup, sometimes returns the same watch group, sometimes a copy. It's not great.

My recommended implementation of choose and chooseAll (also recommend rename of chooseAll to `minRev`) 
```
// choose selects watchers from the watcher group to update
func (wg *watcherGroup) choose(maxWatchers int, curRev, compactRev int64) (*watcherGroup, int64) {
	watcherNeedsCompaction := false
	for w := range wg.watchers {
		if w.minRev < compactRev {
			watcherNeedsCompaction = true
		}
	}
	if len(wg.watchers) < maxWatchers && !watcherNeedsCompaction {
		return wg, wg.chooseAll(curRev, compactRev)
	}
	ret := newWatcherGroup()
	for w := range wg.watchers {
		if maxWatchers <= 0 {
			break
		}
		maxWatchers--
		if w.minRev < compactRev {
			select {
			case w.ch <- WatchResponse{WatchID: w.id, CompactRevision: compactRev}:
				w.compacted = true
				wg.delete(w)
			default:
				// retry next time
			}
			continue
		}
		ret.add(w)
	}
	return &ret, ret.chooseAll(curRev, compactRev)
}

func (wg *watcherGroup) chooseAll(curRev, compactRev int64) int64 {
	minRev := int64(math.MaxInt64)
	for w := range wg.watchers {
		if w.minRev > curRev {
			// after network partition, possibly choosing future revision watcher from restore operation
			// with watch key "proxy-namespace__lostleader" and revision "math.MaxInt64 - 2"
			// do not panic when such watcher had been moved from "synced" watcher during restore operation
			if !w.restore {
				panic(fmt.Errorf("watcher minimum revision %d should not exceed current revision %d", w.minRev, curRev))
			}

			// mark 'restore' done, since it's chosen
			w.restore = false
		}
		if minRev > w.minRev && minRev > compactRev {
			minRev = w.minRev
		}
	}
	return minRev
}
```

I think it's cleaner, as for performance I think we need to test as there is a tradeoff between copying `watchableStore` or matching events.

tests/e2e/watch_delay_test.go
```diff
@@ -246,3 +253,241 @@ func continuouslyExecuteGetAll(ctx context.Context, t *testing.T, g *errgroup.Gr
 		return nil
 	})
 }
+
+// TestWatchOnStreamMultiplex ensures slow etcd watchers throws terminal ErrCompacted error if its
+// next batch of events to be sent are compacted.
+func TestWatchOnStreamMultiplex(t *testing.T) {
```
serathius, line 259: Good job on the test, I have confirmed that it reproduces the problem nicely. 

Is there any way to simplify the code?

server/storage/mvcc/watchable_store.go
chaochn47: Thanks for the feedback! 

> Line 354: Pick unsynched watchers that we want to sync, and their minRev which should be higher than compactionRev. It might also drop compacted watchers.

I agree it's a cleaner fix because we should not update the compacted watchers in the first place in Line 354. Currently, line 371 `wb := newWatcherBatch(wg, evs)` is a waste of space allocated up to N (compacted watchers) * `watchBatchMaxRevs` (1000) events and then garbage collected. 

> as for performance I think we need to test as there is a tradeoff between copying watchableStore or matching events. 

Don't understand this statement because I think we are already copying `watcherGroup` if # of unsync watchers is greater than 512. 

My question is: 

Is this recommended change acceptable for backporting 3.5 & 3.4 release branch comparing to original 2 lines of code change? Both proposed changes look correct to me.  @ahrtr @serathius @fuweid

chaochn47:
@ahrtr @serathius 

> it only removes the watcher from a temporary watcherGroup (see below). So it doesn't really remove the watcher? Could you double confirm this, e.g by writing a unit test? Thanks.

> Because the .choose function is written badly. It sometimes mutates the original watchGroup, sometimes returns the same watch group, sometimes a copy. It's not great.

Yeah, I have confirmed it would cause watcher group leak. 

```
func TestWatchGroupUpdate(t *testing.T) {
	ch := make(chan WatchResponse, chanBufLen)
	compactRev := int64(5)
	curRev := int64(12)
	wg := newWatcherGroup()
	wg.add(&watcher{
		key:    []byte("foo/"),
		end:    []byte("foo0"),
		minRev: 2,
		id:     0,
		ch:     ch,
	})
	wg.add(&watcher{
		key:    []byte("foo/"),
		end:    []byte("foo0"),
		minRev: 10,
		id:     1,
		ch:     ch,
	})

	wg.choose(1 /* maxWatchers */, curRev, compactRev)
	// we would expect watch id 0 is the compacted watcher and deleted from the watch group
	for w := range wg.watchers {
		t.Logf("compactRev is: %d; watcher %d with minRev %d still in the watcher group", compactRev, w.id, w.minRev)
	}
	require.Equal(t, 1, len(wg.watchers))
}
```

```
dev-dsk-chaochn-2c-a26acd76 % GOWORK=off go test -v -run TestWatchGroupUpdate
=== RUN   TestWatchGroupUpdate
    watcher_group_test.go:32: compactRev is: 5; watcher 1 with minRev 10 still in the watcher group
    watcher_group_test.go:32: compactRev is: 5; watcher 0 with minRev 2 still in the watcher group
    watcher_group_test.go:34:
        	Error Trace:	/home/chaochn/workplace/EKS-etcd/src/EKS-etcd/server/storage/mvcc/watcher_group_test.go:34
        	Error:      	Not equal:
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestWatchGroupUpdate
--- FAIL: TestWatchGroupUpdate (0.00s)
FAIL
exit status 1
FAIL	go.etcd.io/etcd/server/v3/storage/mvcc	0.010s
```

tests/e2e/watch_delay_test.go
chaochn47: Working on it.

chaochn47:
> Are you able to create a simple unit test to simulate the case that w.ch is full instead?

I would like to keep the existing e2e test to ensure no regression of etcd watch behavior after the fix gets in. I would simplify it by removing unnecessary code paths. 

Unit test is good to have and I am working on it.

server/storage/mvcc/watchable_store.go
serathius: I think this is a better fix as it is more readable, we can clearly separate responsibility of:
* Sending compacted response to watcher and removing it
* Selecting list of watchers we want to synchronize
* Picking the minimal revision between members we want to synchronize. 

Don't have anything against backporting this to older releases. I think it's better to backport a proper fix, than backporting just a shortcut, leaving a proper fix just on main branch, causing branches to diverge and further backports harder.

ahrtr:
Please ping me back when you resolve all comments.

ahrtr:
> I would like to keep the existing e2e test to ensure no regression of etcd watch behavior after the fix gets in. I would simplify it by removing unnecessary code paths.
> 
> Unit test is good to have and I am working on it.

OK

server/storage/mvcc/watchable_store.go
fuweid: just my two cents. even if we don't return watchers that they need to be compacted, we still need condition checker to make sure please do not touch their watchers' minRev. For the waiterGroup bug, we can have separate commits (in one pull request) to fix. The v3.4 release is still active and codebase layout is different from main and 3.5. multiple commits could help when we backport. Just don't want to involve refactor things in bugfix backports.

ahrtr: > Is this recommended change acceptable for backporting 3.5 & 3.4 release branch comparing to original 2 lines of code change? Both proposed changes look correct to me. @ahrtr @serathius @fuweid

Usually we do code refactoring on main branch only. But I am not against doing such kind of local/small area of code refactoring even in release-3.4/5, because we have to fix the second related issue as mentioned in https://github.com/etcd-io/etcd/pull/17555#issuecomment-1985854328.

Please deliver a PR and we will evaluate the impact. Thanks.

serathius:
I think we might need to do more changes in the watch code soon.

Let's go with a minimal fix proposed by the original PR to address the issue. 

@chaochn47 can you simplify the e2e test? I should be able to help with if needed.

tests/e2e/watch_delay_test.go
chaochn47: Done. Let me know if this is not simplified enough.

server/storage/mvcc/watchable_store.go
chaochn47: Thanks all, let's keep the current fix as it is and would refactor and fix the https://github.com/etcd-io/etcd/pull/17555#issuecomment-1985854328 in follow up PRs.

chaochn47:
Ping @ahrtr @serathius

serathius:
Have you considered using a gofailpoint to simulate watch stream being clogged? 

I managed to reproduce the issue using a following test:
```
func TestV3NoEventsLostOnCompact(t *testing.T) {
	if integration.ThroughProxy {
		t.Skip("grpc proxy currently does not support requesting progress notifications")
	}
	integration.BeforeTest(t)

	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
	defer clus.Terminate(t)

	client := clus.RandClient()
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	writeCount := mvcc.WatchStreamResponseBufferLen * 11 / 10

	wch := client.Watch(ctx, "foo")
	require.NoError(t, gofail.Enable("watchResponseSend", `sleep(1000)`))
	var rev int64 = 0
	for i := 0; i < writeCount; i++ {
		resp, err := client.Put(ctx, "foo", "bar")
		require.NoError(t, err)
		rev = resp.Header.Revision
	}
	_, err := client.Compact(ctx, rev)
	require.NoError(t, err)
	time.Sleep(time.Second)
	require.NoError(t, gofail.Disable("watchResponseSend"))

	event_count := 0
	compacted := false
	for resp := range wch {
		err = resp.Err()
		if err != nil {
			if !strings.Contains(err.Error(), "required revision has been compacted") {
				t.Fatal(err)
			}
			compacted = true
			break
		}
		event_count += len(resp.Events)
		if event_count == writeCount {
			break
		}
	}
	assert.Truef(t, compacted, "Expected stream to get compacted, instead we got %d events out of %d events", event_count, writeCount)
}
```
Where `watchResponseSend` is a new failpoint before grpc.Send

chaochn47:
Added unit test, integration test with failpoint and e2e test together and let you guys pick which one is the best.

server/storage/mvcc/watchable_store_test.go
```diff
@@ -250,6 +251,72 @@ func TestWatchCompacted(t *testing.T) {
...
+					zap.Int64("compact-revision", resp.CompactRevision))
+				compactedResponseCount++
+			}
+			for _, ev := range resp.Events {
+				if nextRev, ok := watchersNextRev[resp.WatchID]; ok {
+					if ev.Kv.ModRevision != nextRev {
+						t.Fatalf("got event revision %d but want %d for watcher with watch ID %d", ev.Kv.ModRevision, nextRev, resp.WatchID)
+					}
+					watchersNextRev[resp.WatchID]++
+				}
```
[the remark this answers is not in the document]
chaochn47: Makes sense.

tests/e2e/watch_delay_test.go
```diff
@@ -246,3 +252,191 @@ func continuouslyExecuteGetAll(ctx context.Context, t *testing.T, g *errgroup.Gr
...
+				return nil
+			}
+
+			select {
+			case <-ticker.C:
+			case <-timer.C:
+				triggered := eventsTriggered.Load()
+				received := watchCacheEventsReceived.Load()
+				return fmt.Errorf("5 minutes passed since load generation is done, watch cache lost event detected; "+
+					"watch evetns received %d, received %d", received, triggered)
```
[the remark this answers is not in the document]
chaochn47: Removed the test.

chaochn47:
> The e2e test is still a little over complicated to me. It will definitely be painful for other contributors to update/maintain such e2e test in future. I think it should be part of the robustness test, which already supports generating traffic, watching, compaction and verifying results (including watchResponses), why do you have to write similar complex test in e2e test suite? But I won't insist on that. 

Okay. Let's remove this e2e test from the PR and track this support in robustness or performance test suite. This test was created originally to simulate kubernetes traffic. With unit test and integration test, it's enough to capture regression.

server/storage/mvcc/watchable_store_test.go
```diff
@@ -250,6 +251,66 @@ func TestWatchCompacted(t *testing.T) {
...
+
+	maxRev := 10
+	compactRev := int64(5)
+	for i := 0; i < maxRev; i++ {
+		s.Put(testKey, testValue, lease.NoLease)
+	}
+	_, err := s.Compact(traceutil.TODO(), compactRev)
+	if err != nil {
+		t.Fatalf("failed to compact kv (%v)", err)
+	}
```
ahrtr, line 277: nit
```suggestion
	require.NoError(t, err)
```

server/storage/mvcc/watchable_store_test.go
```diff
@@ -250,6 +251,66 @@ func TestWatchCompacted(t *testing.T) {
...
+	w := s.NewWatchStream()
+	defer w.Close()
+
+	watchers := map[WatchID]int64{
+		0: 1,
+		1: 1, // create unsyncd watchers with startRev < compactRev
+		2: 6, // create unsyncd watchers with compactRev < startRev < currentRev
+	}
+	for id, startRev := range watchers {
+		w.Watch(id, testKey, nil, startRev)
```
ahrtr, line 288:
```suggestion
		_, werr := w.Watch(id, testKey, nil, startRev)
		require.NoError(t, werr)
```

ahrtr (approved):
LGTM with two very minor comments, which can be resolved separately.

We can let this PR in for now if there is no other comment. cc @serathius

server/storage/mvcc/watchable_store_test.go
```diff
@@ -250,6 +251,63 @@ func TestWatchCompacted(t *testing.T) {
 	}
 }
 
+func TestWatchNoEventLossOnCompact(t *testing.T) {
+	oldChanBufLen, oldMaxWatchersPerSync := chanBufLen, maxWatchersPerSync
```
serathius, line 255: Not sure how relevant is maxWatchersPerSync to the issue and this test as `len(watchers) < 4`. 

What about the case where `len(watchers) > maxWatchersPerSync` as pointed out in https://github.com/etcd-io/etcd/pull/17555#issuecomment-1985854328 ? I haven't verified it, but I expect that unremoved watcher from `s.unsynced` will cause `syncWatchers` to return != 0, causing function to be called earlier.

tests/integration/v3_watch_test.go
```diff
@@ -1512,3 +1517,56 @@ func TestV3WatchProgressWaitsForSyncNoEvents(t *testing.T) {
 	}
 	require.True(t, gotProgressNotification, "Expected to get progress notification")
 }
+
+// TestV3NoEventsLostOnCompact verifies that slow watchers exit with compacted watch response
+// if its next revision of events are compacted and no lost events sent to client.
+func TestV3NoEventsLostOnCompact(t *testing.T) {
+	if integration.ThroughProxy {
+		t.Skip("grpc proxy currently does not support requesting progress notifications")
```
serathius, line 1525: Don't think this comment is relevant.

serathius (approved):

server/storage/mvcc/watchable_store_test.go
ahrtr: > Not sure how relevant is maxWatchersPerSync to the issue and this test as `len(watchers) < 4`.

It seems to be true. Confirmed that it always runs into the if branch (see below), and confirmed that it has no any impact on the test case. 

But it isn't a big deal, and also from another prospective it should be OK to explicitly set a value to ensure `len(wg.watchers) < maxWatchers` although it's already `true` by default.

https://github.com/etcd-io/etcd/blob/63e394d090bb7821b6bf2c255d2a70e86e8fc693/server/storage/mvcc/watcher_group.go#L225-L227

> What about the case where len(watchers) > maxWatchersPerSync as pointed out in #17555 (comment) ?

Suggest to discuss & fix it separately. We may want to do minor local code refactor. @chaochn47 are you able to continue to work on this?

ahrtr: > confirmed that it has no any impact on the test case.

I mean that the even I don't change `maxWatchersPerSync` in the test case , the test case could also reproduce the issue without the fix and the issue disappeared after applying the patch in this PR.

serathius: Sounds good

ahrtr:
@chaochn47 do you have bandwidth to backport this to 3.5 and 3.4? We may need to release a 3.4 and 3.5 patch soon.

chaochn47:
> @chaochn47 do you have bandwidth to backport this to 3.5 and 3.4? We may need to release a 3.4 and 3.5 patch soon.

@ahrtr @jmhbnz sure. Just provided backports PRs. Could you please take a look? 
- https://github.com/etcd-io/etcd/pull/17612
- https://github.com/etcd-io/etcd/pull/17610

Issue #17507, opened 2024-02-28
Etcd not respond to progress notification on newly created watch RPC until there is a event in a watch


serathius:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

When trying make K8s depend on progress notification I found an broken edge case when working on https://github.com/kubernetes/kubernetes/pull/123513

What I found is that https://github.com/etcd-io/etcd/pull/15237 introduced a bug with `deferredProgress` logic.

If a progress request is send when watch [is not synced](https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/server/storage/mvcc/watchable_store.go#L491-L495) (like when it's just newly created watch), etcd [mark progress notification as deferred](https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/server/etcdserver/api/v3rpc/watch.go#L375), and [stop responding to progress notification request](https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/server/etcdserver/api/v3rpc/watch.go#L371). 

Progress notification can be [later restored](https://github.com/etcd-io/etcd/blob/e68afe7eb7ce3e5f83deb2a8ad58faa3b35494b9/server/etcdserver/api/v3rpc/watch.go#L486-L490) but only if there is an event to be sent. So if there is no event observed in the watch, etcd will never respond to progress notification requests.

### What did you expect to happen?

Don't expect event to arrive, send the progress notification when watchers become synced. 

### How can we reproduce it (as minimally and precisely as possible)?

Take PR.

Run make test WHAT=./vendor/k8s.io/apiserver/pkg/storage/... GOFLAGS="-v" KUBE_TEST_ARGS='--count 10 --run TestListWithListFromCache --failfast'

### Etcd version (please run commands below)

Reproduced on v3.5.10 embedded server

[...]

PR #17557, opened 2024-03-08
Fix progress notification for watch that doesn't get any events

serathius:
When implementing the fix for progress notifications (https://github.com/etcd-io/etcd/pull/15237) we made a incorrect assumption that that unsynched watches will always get at least one event.

Unsynched watches include not only slow watchers, but also newly created watches that requested old revision. In case that non of the events match watch filter, those newly created watches might become synched without any event going through.

Fixes https://github.com/etcd-io/etcd/issues/17507
/cc @scpmw @ahrtr

### DIFF
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -145,10 +145,6 @@ type serverWatchStream struct {
 	// records fragmented watch IDs
 	fragment map[mvcc.WatchID]bool
 
-	// indicates whether we have an outstanding global progress
-	// notification to send
-	deferredProgress bool
-
 	// closec indicates the stream is closed.
 	closec chan struct{}
 
@@ -177,8 +173,6 @@ func (ws *watchServer) Watch(stream pb.Watch_WatchServer) (err error) {
 		progress: make(map[mvcc.WatchID]bool),
 		prevKV:   make(map[mvcc.WatchID]bool),
 		fragment: make(map[mvcc.WatchID]bool),
-
-		deferredProgress: false,
 
 		closec: make(chan struct{}),
 	}
@@ -367,14 +361,7 @@ func (sws *serverWatchStream) recvLoop() error {
 		case *pb.WatchRequest_ProgressRequest:
 			if uv.ProgressRequest != nil {
 				sws.mu.Lock()
-				// Ignore if deferred progress notification is already in progress
-				if !sws.deferredProgress {
-					// Request progress for all watchers,
-					// force generation of a response
-					if !sws.watchStream.RequestProgressAll() {
-						sws.deferredProgress = true
-					}
-				}
+				sws.watchStream.RequestProgressAll()
 				sws.mu.Unlock()
 			}
 		default:
@@ -482,11 +469,6 @@ func (sws *serverWatchStream) sendLoop() {
 			if len(evs) > 0 && sws.progress[wresp.WatchID] {
 				// elide next progress update if sent a key update
 				sws.progress[wresp.WatchID] = false
-			}
-			if sws.deferredProgress {
-				if sws.watchStream.RequestProgressAll() {
-					sws.deferredProgress = false
-				}
 			}
 			sws.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


serathius:
ping @ahrtr

tests/integration/v3_watch_test.go
```diff
@@ -1467,3 +1468,48 @@ func TestV3WatchProgressWaitsForSync(t *testing.T) {
...
+	defer cancel()
+
+	resp, err := client.Put(ctx, "bar", "1")
+	require.NoError(t, err)
+
+	wch := client.Watch(ctx, "foo", clientv3.WithRev(resp.Header.Revision))
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
+
+	err = client.RequestProgress(ctx)
```
ahrtr, line 1492: If I understand it correctly, the `client.RequestProgress` won't get any progress notification response before etcd moves the watcher from unsynced list to synced list (see `(s *watchableStore) syncWatchers`), but it will set the `deferredProgress` to `true`.  But there isn't any data/events to send for the key "foo", so the `deferredProgress` will never to set to `false`, accordingly all the following `client.RequestProgress` in the loop won't get any progress notification without the fix in this PR.

Please add a comment in the test case.

tests/integration/v3_watch_test.go
```diff
@@ -1467,3 +1468,48 @@ func TestV3WatchProgressWaitsForSync(t *testing.T) {
...
+	if integration.ThroughProxy {
+		t.Skip("grpc proxy currently does not support requesting progress notifications")
+	}
+	integration.BeforeTest(t)
+
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer clus.Terminate(t)
+
+	client := clus.RandClient()
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
ahrtr, line 1482: The ctx is reused all through this test case, suggest to set a larger timeout value, e.g 5s. arm64 platform isn't stable, it may be flaky on arm64 platform.

tests/integration/v3_watch_test.go
```diff
@@ -1467,3 +1468,48 @@ func TestV3WatchProgressWaitsForSync(t *testing.T) {
...
+				gotProgressNotification = true
+			}
+		}
+		if gotProgressNotification {
+			break
+		}
+	}
+	if !gotProgressNotification {
+		t.Fatal("Expected to get progress notification")
+	}
```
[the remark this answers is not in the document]
serathius: Done

tests/integration/v3_watch_test.go
serathius: Added a comment why we send a progress notification here.

tests/integration/v3_watch_test.go
serathius: Right, lowered it for fast iteration when working on the fix, forgot to increase it.

p0lyn0mial:
/cc @wojtek-t @p0lyn0mial

Thanks @serathius for putting together this PR!

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -367,14 +361,7 @@ func (sws *serverWatchStream) recvLoop() error {
...
 				sws.mu.Lock()
-				// Ignore if deferred progress notification is already in progress
-				if !sws.deferredProgress {
-					// Request progress for all watchers,
-					// force generation of a response
-					if !sws.watchStream.RequestProgressAll() {
-						sws.deferredProgress = true
-					}
-				}
+				sws.watchStream.RequestProgressAll()
```
ahrtr, line 364: nit: print a debug log when it fails to send out the progress notification.

ahrtr (approved):
lgtm

server/etcdserver/api/v3rpc/watch.go
serathius: Let me address it in a separate PR, I think more places in watch code need debug logging.

Issue #16839, opened 2023-10-27
Watch starvation can cause OOMs


serathius:
### What would you like to be added?

Please read https://github.com/etcd-io/etcd/issues/16839#issuecomment-1989350422 for up to date information about the problem

serathius:
Profiles I promised. They are less extreme than onces we have seen originally. They are more recent, as we reduced the write throughput later. Still they show the problem:

32% of allocations done by Range to get PrevKey for Watch

![image](https://github.com/etcd-io/etcd/assets/5873433/3d97b39e-27dd-46ae-88e8-cf08861a353b)

57% of cpu used on GC
![image](https://github.com/etcd-io/etcd/assets/5873433/170de964-c7ce-4386-9cc3-f90ab94a304d)

serathius:
One clarification, access to prevKey in https://github.com/etcd-io/etcd/blob/a4f507c30a7b1335a240b9b9c263501c3d7262a1/server/storage/mvcc/kvstore_txn.go#L203 doesn't touch bbolt, only the in memory index.

chaochn47:
xref. https://github.com/etcd-io/etcd/issues/17529#issuecomment-1980258283.

chaochn47:
The bigger issue about `prevKV` is for a single write, we copy the event for the matching watchers and for each watcher, we copy the entire transaction buffer just to get prevKV and it would duplicate 100 time (let's say there are 100 watchers). 

Hence it is not efficient IMHO.

serathius:
Looked into performance of prevKV, it is abysmal, don't know how we released such feature and how Kubernetes adopted it without noticing. 

I have modified the [watch latency benchmark](https://github.com/etcd-io/etcd/blob/main/tools/benchmark/cmd/watch_latency.go) to be useful (will share code soon). 

For 100KB objects, 100 watchers, 100 put qps (similar to setup in https://github.com/kubernetes/kubernetes/issues/123448) I got:

<google-sheets-html-origin></style>
prevKV | etcd memory [GB] | etcd cpu [core] | 90%ile event latency[s] | 99%ile watch event latency[s] | 99.9%ile watch event latency[s]
-- | -- | -- | -- | -- | --
disabled | 1.1 | 0.76 | 0.2s | 0.73s | 1s
enabled | 8.6 | 1 | 16s | 18s | 25.7s

The @chaochn47 fix for compacted watchers will not help here, as there was no compaction done here. We need to address it even if it's a performance change.

[...]

PR #17563, opened 2024-03-10
Reuse events used for syncing watchers

serathius:
Part of https://github.com/etcd-io/etcd/issues/16839
Improve etcd behavior during watch congestion by reducing the memory needed. 
Congestion causes watch requests to backup and resync loop to allocate tons of memory. This patch has shown to reduce memory needed 5-10 times.

| Reuse events | object size [KB] | etcd memory[MB] | latency 50%ile | latency 90%ile | latency 99%ile |
| ------------ | ---------------- | --------------- | -------------- | -------------- | -------------- |
| FALSE        | 5                | 111.72          | 0.1005         | 0.1543         | 0.1856         |
| TRUE         | 5                | 109.096         | 0.1012         | 0.1556         | 0.1874         |
| FALSE        | 10               | 1103.924        | 2.7177         | 7.3661         | 11.9127        |
| TRUE         | 10               | 291.184         | 2.6623         | 7.687          | 12.5456        |
| FALSE        | 100              | 5138.712        | 14.0618        | 26.2721        | 29.1691        |
| TRUE         | 100              | 641.844         | 13.9372        | 25.8267        | 28.519         |

Tested using command `go run ./tools/benchmark/main.go watch-latency --watch-per-stream 1000 --streams 1 --put-total 200 --val-size 100000`. Parameters adjusted for total object size be a 20MB.

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -357,20 +357,7 @@ func (s *watchableStore) syncWatchers() int {
 	compactionRev := s.store.compactMainRev
 
 	wg, minRev := s.unsynced.choose(maxWatchersPerSync, curRev, compactionRev)
-	minBytes, maxBytes := NewRevBytes(), NewRevBytes()
-	minBytes = RevToBytes(Revision{Main: minRev}, minBytes)
-	maxBytes = RevToBytes(Revision{Main: curRev + 1}, maxBytes)
-
-	// UnsafeRange returns keys and values. And in boltdb, keys are revisions.
-	// values are actual key-value pairs in backend.
-	tx := s.store.b.ReadTx()
-	tx.RLock()
-	revs, vs := tx.UnsafeRange(schema.Key, minBytes, maxBytes, 0)
-	evs := kvsToEvents(s.store.lg, wg, revs, vs)
-	// Must unlock after kvsToEvents, because vs (come from boltdb memory) is not deep copy.
-	// We can only unlock after Unmarshal, which will do deep copy.
-	// Otherwise we will trigger SIGSEGV during boltdb re-mmap.
-	tx.RUnlock()
+	evs := rangeEvents(s.store.lg, s.store.b, minRev, curRev+1, wg)
 
 	victims := make(watcherBatch)
 	wb := newWatcherBatch(wg, evs)
@@ -422,15 +409,38 @@ func (s *watchableStore) syncWatchers() int {
 	return s.unsynced.size()
 }
 
+// rangeEvents returns events in range [minRev, maxRev).
+func rangeEvents(lg *zap.Logger, b backend.Backend, minRev, maxRev int64, c contains) []mvccpb.Event {
+	minBytes, maxBytes := NewRevBytes(), NewRevBytes()
+	minBytes = RevToBytes(Revision{Main: minRev}, minBytes)
+	maxBytes = RevToBytes(Revision{Main: maxRev}, maxBytes)
+
+	// UnsafeRange returns keys and values. And in boltdb, keys are revisions.
+	// values are actual key-value pairs in backend.
+	tx := b.ReadTx()
+	tx.RLock()
+	revs, vs := tx.UnsafeRange(schema.Key, minBytes, maxBytes, 0)
+	evs := kvsToEvents(lg, c, revs, vs)
+	// Must unlock after kvsToEvents, because vs (come from boltdb memory) is not deep copy.
+	// We can only unlock after Unmarshal, which will do deep copy.
+	// Otherwise we will trigger SIGSEGV during boltdb re-mmap.
+	tx.RUnlock()
+	return evs
+}
+
+type contains interface {
+	contains(string) bool
+}
+
 // kvsToEvents gets all events for the watchers from all key-value pairs
-func kvsToEvents(lg *zap.Logger, wg *watcherGroup, revs, vals [][]byte) (evs []mvccpb.Event) {
+func kvsToEvents(lg *zap.Logger, c contains, revs, vals [][]byte) (evs []mvccpb.Event) {
 	for i, v := range vals {
 		var kv mvccpb.KeyValue
 		if err := kv.Unmarshal(v); err != nil {
 			lg.Panic("failed to unmarshal mvccpb.KeyValue", zap.Error(err))
 		}
 
-		if !wg.contains(string(kv.Key)) {
+		if !c.contains(string(kv.Key)) {
 			continue
 		}
 
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -221,6 +221,7 @@ func (s *watchableStore) syncWatchersLoop() {
 	waitDuration := 100 * time.Millisecond
 	delayTicker := time.NewTicker(waitDuration)
 	defer delayTicker.Stop()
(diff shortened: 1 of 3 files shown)


chaochn47:
> Limit event prefix by slowest watch instead of compaction.

Do we expect sending events to compacted watchers (the slowest one) though? I think it should not. WDYT?

serathius:
> Do we expect sending events to compacted watchers (the slowest one) though? I think it should not. WDYT?

I expect that minRev and rev of victims is above compactedRev, if now we can just add a protection.

serathius:
ping @ahrtr

server/storage/mvcc/watchable_store.go
```diff
@@ -403,11 +404,39 @@ func (s *watchableStore) syncWatchers() int {
...
+	evs = evs[prefixIndex:]
+
+	if len(evs) == 0 {
+		return rangeEvents(lg, b, minRev, maxRev)
+	}
+	if evs[len(evs)-1].Kv.ModRevision+1 < maxRev {
+		evs = append(evs, rangeEvents(lg, b, evs[len(evs)-1].Kv.ModRevision+1, maxRev)...)
+	}
+	suffixIndex := len(evs) - 1
+	for suffixIndex >= 0 && evs[suffixIndex].Kv.ModRevision >= maxRev {
```
ahrtr, line 431: In which case `evs[suffixIndex].Kv.ModRevision >= maxRev` can be `true`?

serathius: From the calls in `syncWatchers` this should never happen as it would mean that curRev decreased, which should not be possible. However when writing the function I decided implement the function to handle this case for testing. 

It allows us to use the same tests for both `rangeEvents` and `rangeEventsWithReuse`, it's a very nice property that reduces the chance of error in test.

ahrtr: > that reduces the chance of error in test.

In that case, you need to fix the test instead of the production code.

> I decided implement the function to handle this case for testing.

On the contrary, I think it would be better to raise an error or even a panic in such case, because it means a major or critical bug in production.

serathius: Not every low level function needs to be aware of every assumption. Don't know if this rule has a name, but "high level functions should assume less and require more, however the lower the call stack you go you can assume more and require less". For me definition of function "rageRevisions" should handle arbitrary range. 

> In that case, you need to fix the test instead of the production code.

It's not issue with test. Tests that should be easy to understand and maintainable, this makes the code easier to test.

> > I decided implement the function to handle this case for testing.
>
> On the contrary, I think it would be better to raise an error or even a panic in such case, because it means a major or critical bug in production.

It's not responsibility of function for reading revisions to validate if the revision in the etcd is non-decreasing. If we couldn't implement this case then we should add a assertion to protect function returning bad response, however implementation is simple enough that we can handle it.

Even though we expect global revision to never decrease, don't know if this is necessarily true in watch serving code, especially during restore. We track revision progress per watcher, so if revision ever decreased we would just skip sending event. I think this case was exactly confirmed in https://github.com/etcd-io/etcd/issues/15271.

  Don't think we should introduce this assertion now.

ahrtr: Raised https://github.com/etcd-io/etcd/pull/18980 to verify this in a higher level function.

server/storage/mvcc/watchable_store_test.go
```diff
@@ -164,6 +164,77 @@ func TestSyncWatchers(t *testing.T) {
...
+	defer cleanup(s, b)
+
+	foo1 := []byte("foo1")
+	foo2 := []byte("foo2")
+	foo3 := []byte("foo3")
+	value := []byte("bar")
+	s.Put(foo1, value, lease.NoLease)
+	s.Put(foo2, value, lease.NoLease)
+	s.Put(foo3, value, lease.NoLease)
+	s.DeleteRange(foo1, foo3) // Deletes "foo1" and "foo2" generating 2 events
```
ahrtr, line 181: Can you add a comment something like below 

https://github.com/etcd-io/etcd/blob/afb5c1d85b2f0ec24852c4d409bca8401dbdea6b/server/storage/mvcc/index_test.go#L240-L256

or just a list of revisions 
```
rev: 1, key: foo1, PUT
rev: 2, key: foo2, PUT
rev: 3, key: foo3, PUT
rev: 4, key: foo1, DELETE
rev: 5, key: foo2, DELETE
```
 so that others can understand the following case more easily
?

server/storage/mvcc/watchable_store_test.go
```diff
@@ -164,6 +164,77 @@ func TestSyncWatchers(t *testing.T) {
...
+				return rangeEvents(lg, b, minRev, maxRev)
+			},
+		},
+		{
+			rangeFunc: func(minRev, maxRev int64) []mvccpb.Event {
+				return rangeEventsWithReuse(lg, b, evs, minRev, maxRev)
+			},
+		},
+	}
+	for _, tc := range tcs {
```
ahrtr, line 198: sub-test?

server/storage/mvcc/watchable_store_test.go
serathius: PTAL https://github.com/etcd-io/etcd/pull/18981

server/storage/mvcc/watchable_store_test.go
serathius: PTAL https://github.com/etcd-io/etcd/pull/18981

server/storage/mvcc/watchable_store_test.go
serathius: Done

server/storage/mvcc/watchable_store_test.go
serathius: Done

server/storage/mvcc/watchable_store_test.go
```diff
@@ -164,6 +164,101 @@ func TestSyncWatchers(t *testing.T) {
...
+		{minRev: 2, maxRev: 5, expectEvents: expectEvents[0:3]},
+		{minRev: 2, maxRev: 4, expectEvents: expectEvents[0:2]},
+		{minRev: 2, maxRev: 3, expectEvents: expectEvents[0:1]},
+		{minRev: 2, maxRev: 2, expectEvents: nil},
+		{minRev: 3, maxRev: 6, expectEvents: expectEvents[1:5]},
+		{minRev: 4, maxRev: 6, expectEvents: expectEvents[2:5]},
+		{minRev: 5, maxRev: 6, expectEvents: expectEvents[3:5]},
+		{minRev: 6, maxRev: 6, expectEvents: nil},
+		{minRev: 2, maxRev: 3, expectEvents: expectEvents[0:1]},
+		{minRev: 2, maxRev: 4, expectEvents: expectEvents[0:2]},
```
ahrtr, line 246: These two cases are duplicated? see line 238-239. Better to sort the cases by `minRev` or `maxRev`?

serathius: Yes, it's intentional. We group multiple cases into a scenarios. Added comments to make them easier to see. This is to better test the reuse, as the function call has side effects with modifying `evs`.

ahrtr (approved):
LGTM

It would be wonderful if we have more performance data (i.e. in K8s scale test) to showcase the benefit of such improvement.

PR #17567, opened 2024-03-11
[release-3.4] Fix progress notification for watch that doesn't get any events

serathius:
When implementing the fix for progress notifications (https://github.com/etcd-io/etcd/pull/15237) we made a incorrect assumption that that unsynched watches will always get at least one event.

Unsynched watches include not only slow watchers, but also newly created watches that requested current or older revision. In case that non of the events match watch filter, those newly created watches might become synched without any event going through.

Backport https://github.com/etcd-io/etcd/pull/17557


ahrtr (approved):
lgtm

Thanks

Issue #17578, opened 2024-03-12
Address var-name linter warnings


ivanvc:
### What would you like to be added?

As discussed at the March 7th etcd's community meeting, enabling the `var-name` revive linter involves about 1,000 lines of changes. To break this down into smaller subtasks, this would be small enough to be reviewable in pull requests while keeping the codebase functional.

- [x] `client/` directory - #17582
- [x] `api/` directory - Two public exported variables that need to be changed (`ErrGRPCClusterIdMismatch` and `ErrClusterIDMismatch`). Does it make sense to deprecate the variables? Or, given that it's the `main` branch, is it ok to rename them? - #17585
- [x] `pkg/` directory - #17584
- [x] `server/` directory - This contains about 500 instances that need to be changed
  - [x] `server/etcdserver` - #17614
  - [x] `server/storage` - #17660
  - [x] `server/config` - #17854
  - [x] `server/embed`
    - [x] #17674
    - [x] #17820
  - [x] `server/*` - #17624
- [x] `tests/` directory - This contains about 350 instances that need to be changed
  - [x] `tests/common`, `tests/integration` - #17635
  - [x] `tests/e2e` - #17636
  - [x] `tests/framework` - #17651
    - [x] removal of `(*Member) GRPCURL()` #17655
  - [x] `tests/robustness` - #17654
- [x] `etcdutl/`, `tools/`, `tests/common/` - #17863
- [x] Fix deprecation comments, see: https://github.com/etcd-io/etcd/pull/17674#discussion_r1554831812 - #17875
- [ ] Enable lint rule in `tools/.golangci.yaml` - #17921 

### Why is this needed?

To improve the quality of the code.

serathius:
Thanks @ivanvc for your hard work!

PR #17582, opened 2024-03-13
client: address golangci var-naming issues

ivanvc:
If we need to deprecate rather than rename an exported variable, we can add an ignore to the linter for that line.

Renamed exported variable:
* `logutil.JsonLogFormat` -> `logutil.JSONLogFormat`

Renamed exported variables from internal packages:
* `endpoint.CREDS_REQUIRE` -> `endpoint.CredsRequire`
* `endpoint.CREDS_DROP` -> `endpoint.CredsDrop`
* `endpoint.CREDS_OPTIONAL` -> `endpoint.CredsOptional`

Exported renamed variables to unexported as they weren't being used anywhere outside their package:
* `verify.ENV_VERIFY` -> `verify.envVerify`
* `verify.ENV_VERIFY_VALUE_ALL` -> `verify.envVerifyValueAll`
* `verify.ENV_VERIFY_VALUE_ASSERT` -> `verify.envVerifyValueAssert`

After addressing these, the var-naming rule passes in the `client` directory.

Relates to #17578

### DIFF
--- a/client/pkg/logutil/log_format.go
+++ b/client/pkg/logutil/log_format.go
@@ -17,19 +17,23 @@ package logutil
 import "fmt"
 
 const (
-	JsonLogFormat    = "json"
+	JSONLogFormat    = "json"
 	ConsoleLogFormat = "console"
+	//revive:disable:var-naming
+	// Deprecated: Please use JSONLogFormat.
+	JsonLogFormat = JSONLogFormat
+	//revive:enable:var-naming
 )
 
-var DefaultLogFormat = JsonLogFormat
+var DefaultLogFormat = JSONLogFormat
 
 // ConvertToZapFormat converts and validated log format string.
 func ConvertToZapFormat(format string) (string, error) {
 	switch format {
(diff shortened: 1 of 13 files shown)


ahrtr (approved):
lgtm

Thanks

jmhbnz (approved):
LGTM - Thanks for raising these as smaller pr's for each package 🙏🏻

client/pkg/logutil/log_format.go
```diff
@@ -17,19 +17,19 @@ package logutil
 import "fmt"
 
 const (
-	JsonLogFormat    = "json"
```
ahrtr, line 20: The only concern is this is a public/exported const. Probably we should 
- deprecate the existing name and plan to remove it in next major or minor release and skip the linter for this const;
- add a new const `JSONLogFormat`.

ivanvc: Done, I deprecated `JsonLogFormat`.

ahrtr (approved):

PR #17583, opened 2024-03-13
Increase sync watchers period to 1 second

serathius:
Ref https://github.com/etcd-io/etcd/issues/16839

Picked based on 
object size|streams| sync duration[s] | response throughput[MB/s] | etcd memory [GB] | 50%ile latency[s] | 90%ile latency[s] | 99%ile latency[s] | 99.9%ile latency[s] |
-----------|-------|------------------|---------------------------|------------------|-------------------|-------------------|-------------------|---------------------|
1MB        | 20    | 0.1              | 1579                      | 14.9             | 3.47              | 6.32              | 8.22              | 9.4                 |
30KB       | 1000  | 0.1              | 2006                      | 5.5              | 4.26              | 12.74             | 18.81             | 22.69               |
3KB        | 10000 | 0.1              | 2061                      | 1.8              | 4.78              | 12.82             | 19.33             | 23.68               |
1MB        | 20    | 1                | 1634                      | 9                | 3.45              | 5.63              | 7.98              | 9.34                |
30KB       | 1000  | 1                | 2257                      | 1.4              | 4.07              | 11.76             | 17.8              | 21.87               |
3KB        | 10000 | 1                | 2154                      | 1.1              | 4.29              | 11.74             | 18.01             | 22.36               |

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -37,6 +37,17 @@ var (
 
 	// maxWatchersPerSync is the number of watchers to sync in a single batch
 	maxWatchersPerSync = 512
+	// watchersSyncLoopPeriod is number od seconds we attempt to resync unsynched watchers.
+	// To frequent resyncing allocates a lot of memory. Value picked based on the following benchmark results:
+	// object size|streams| sync duration[s] | response throughput[MB/s] | etcd memory [GB] | 50%ile latency[s] | 90%ile latency[s] | 99%ile latency[s] | 99.9%ile latency[s] |
+	// -----------|-------|------------------|---------------------------|------------------|-------------------|-------------------|-------------------|---------------------|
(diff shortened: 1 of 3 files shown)


ahrtr:
Propose to add a config item, which defaults to 1s in main branch, and defaults to 100ms in the stable releases to keep the existing behaviour by default. 

https://github.com/etcd-io/etcd/issues/16839#issuecomment-1999336049

serathius:
Right, with K8s mitigation to stop sending watch to etcd, I don't think it's critical for now. Leaving it for someone to pick up.

serathius:
Updated the PR to provide context about how the variable was picked. I don't think there is a reason to expose is as configuration option, as benchmark results showed clear improvements.

ping @ahrtr for review

serathius:
Ok, test results showed the issue of many existing etcd client recepies depending on time to first event from watch created on concrete revision. Don't think we can move forward with this solution without separating newly created wachers from misbehaving ones.

serathius:
At that point it would be better to do https://github.com/etcd-io/etcd/pull/17563

PR #12667, opened 2021-02-04
Implement Update method for EndpointManager

lemonlinger:
This is a part of changes for https://github.com/etcd-io/etcd/issues/12652 including:
- Add testing for the resolver.
- Implement Update method for EndpointManager.

@peterbourgon

PR #17584, opened 2024-03-13
pkg: address golangci var-naming issues

ivanvc:
Addresses the `var-naming` issues in `pkg`.

* Renames the `grpc_testing` module to `grpctesting`
* Renames the exported `GrpcRecorder` struct to `GRPCRecorder`

Part of #17578

### DIFF
--- a/tests/framework/integration/cluster.go
+++ b/tests/framework/integration/cluster.go
@@ -50,7 +50,7 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	clientv3 "go.etcd.io/etcd/client/v3"
-	"go.etcd.io/etcd/pkg/v3/grpc_testing"
+	"go.etcd.io/etcd/pkg/v3/grpctesting"
 	"go.etcd.io/etcd/server/v3/config"
 	"go.etcd.io/etcd/server/v3/embed"
 	"go.etcd.io/etcd/server/v3/etcdserver"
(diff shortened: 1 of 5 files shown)


tests/framework/integration/cluster.go
```diff
@@ -50,7 +50,7 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	clientv3 "go.etcd.io/etcd/client/v3"
-	"go.etcd.io/etcd/pkg/v3/grpc_testing"
+	grpctesting "go.etcd.io/etcd/pkg/v3/grpc_testing"
```
jmhbnz, line 53: Should we consider changing this at the filesystem level and rename `etcd/pkg/grpc_testing` to `etcd/pkg/grpctesting`?

jmhbnz (approved):
LGTM - Just one question about which approach we use to fix `grpc_testing` below.

tests/framework/integration/cluster.go
ahrtr: Agree to rename `etcd/pkg/grpc_testing` to `etcd/pkg/grpctesting`. It's only used by test, so it should be fine to rename it although it's public.

The package was introduced in https://github.com/etcd-io/etcd/pull/12667. I guess the naming just follows the package used by grpc-go: `google.golang.org/grpc/test/grpc_testing`.

ivanvc: I'll rename the package. Reference to Go best practice: https://go.dev/blog/package-names, where revive complains about underscores in package names: https://github.com/mgechev/revive/blob/ef34f92/rule/var-naming.go#L66-L72. Unfortunately, revive doesn't provide a website with its rules.

ivanvc: Done

ahrtr (approved):
lgtm

cc @fuweid @serathius

ahrtr:
@ivanvc Please rebase this PR since we merged lots of PRs recently.

cc @serathius

serathius (approved):

PR #17585, opened 2024-03-14
api: address golangci var-naming issues

ivanvc:
Addresses the `var-naming` linter issues in `api`.

Renames the exported variables:

* `rpctypes.ErrGRPCClusterIdMismatch` -> `rpctypes.ErrGRPCClusterIDMismatch`
* `rpctypes.ErrClusterIdMismatch` -> `rpctypes.ErrClusterIDMismatch`

Should we need to deprecate rather than rename, I can add a linter exception.

Part of #17578.


ivanvc:
@ahrtr, let me know if I should deprecate the exported variables rather than renaming them as we agreed on `client/`'s

ahrtr:
> @ahrtr, let me know if I should deprecate the exported variables rather than renaming them as we agreed on `client/`'s

YES, I think so

ahrtr (approved):
lgtm

Thanks

serathius (approved):

serathius:
Nice, thanks for not breaking users and adding deprecations

PR #15006, opened 2022-12-16
Add trivy nightly scan for `release-3.5`

ArkaSaha30:
This PR will add trivy nightly scan for `release-3.5` to scan the latest release version `v3.5.6`.

Fixes: #15003

Signed-off-by: ArkaSaha30 <arkasaha30@gmail.com>

PR #17596, opened 2024-03-18
[WIP] Confirm trivy scanner workflow is working

jmhbnz:
Not for review - checking something.

### DIFF
--- a/.github/workflows/trivy-nightly-scan.yaml
+++ b/.github/workflows/trivy-nightly-scan.yaml
@@ -1,9 +1,10 @@
+---
 name: Trivy Nightly Scan
-on:
-  schedule:
-    - cron: '0 2 * * *' # run at 2 AM UTC
+on: [push, pull_request]
+  #schedule:
+  #  - cron: '0 2 * * *' # run at 2 AM UTC
 
-permissions: read-all 
+permissions: read-all
 jobs:
   nightly-scan:
     name: Trivy Scan nightly
@@ -12,7 +13,7 @@ jobs:
       matrix:
         # maintain the versions of etcd that need to be actively
         # security scanned
-        versions: [v3.5.6]
+        versions: [v3.5.12]
     permissions:
       security-events: write  # for github/codeql-action/upload-sarif to upload SARIF results
 


jmhbnz:
Hmm - @ahrtr, @serathius - In December 2022 in https://github.com/etcd-io/etcd/pull/15006 a workflow was introduced for `release-3.5` to scan a specified version of etcd image for vulnerabilities with Trivy.

This has basically never worked because it was set to a nightly schedule, yet merged into `release-3.5`.  In GitHub actions scheduled workflows can only exist in the `main` / primary branch otherwise they don't execute.

Testing in this pull request the workflow does appear to work when changed to not be on a schedule.  Do we want to remove these workflows (given we never used them), or do you want me to bring them into `main` and get them actually running?

Personally my vote would be to remove them - it was introduced ~1.5 years ago and never worked, yet we have managed to have enough security signals from elsewhere to keep up with CVE's.

ahrtr:
> Testing in this pull request the workflow does appear to work when changed to not be on a schedule. Do we want to remove these workflows (given we never used them), or do you want me to bring them into `main` and get them actually running?

Thanks for raising this. Actually I already noticed it, but did not get time to resolved it.  Propose to remove it from release-3.5 and 3.4, and get it added in main, which means we will cover all stable releases and main in the main's workflow. 

cc @ArkaSaha30

jmhbnz:
> Propose to remove it from release-3.5 and 3.4, and get it added in main, which means we will cover all stable releases and main in the main's workflow.

Sounds good to me - @ArkaSaha30 do you have capacity to work on this?

ArkaSaha30:
> > Propose to remove it from release-3.5 and 3.4, and get it added in main, which means we will cover all stable releases and main in the main's workflow.
> 
> 
> 
> Sounds good to me - @ArkaSaha30 do you have capacity to work on this?

Sure @jmhbnz , I can pick this up in the coming week.

serathius:
QQ: Are failures in trivy-scan blocking merging? Do we really want to block development of finding a CVE?

My thought is that CVE findings should be easily visible, but should not block merging. Maybe we could allow trivy-scan to fail? This would still give us information if CVE it was resolved by PR that updates dependencies, but not block PRs unrelated to CVE.

ahrtr:
> QQ: Are failures in trivy-scan blocking merging? Do we really want to block development of finding a CVE?
> 
> My thought is that CVE findings should be easily visible, but should not block merging. Maybe we could allow trivy-scan to fail? This would still give us information if CVE was resolved by dependency update, but not block unrelated PRs.

Overall not a big problem to me.  Either way works for me; but I prefer to keep following the principle of merge-when-all-test-green. It also pushes us to resolve the CVE asap, and most likely we also need to release patch releases soon. But I think we (etcd maintainers) still have some flexibility to merge unrelated PRs if it's really urgent.

Issue #17609, opened 2024-03-18
Plan to release etcd 3.4.31


jmhbnz:
### What would you like to be added?

The etcd [patch release criteria](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/release.md#patch-release-criteria) has been met for our `release-3.4`  stable release branch so we should release `v3.4.31`.

The list of commits included since the previous release is: https://github.com/etcd-io/etcd/compare/v3.4.30...release-3.4

Work in progress CHANGELOG is: https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.4.md#v3431-tbd

List of items we still need to backport from `main` to `release-3.4` before the patch release is issued:
- https://github.com/etcd-io/etcd/pull/17555
  - [ ] https://github.com/etcd-io/etcd/pull/17612
- https://github.com/etcd-io/etcd/pull/17600
  - [ ] https://github.com/etcd-io/etcd/pull/17613

### Why is this needed?

Regular patch releases are vital to ensure our users have bug free and secure software.

PR #17613, opened 2024-03-19
Let dapendabot to auto-update docker for release 3.4&3.5.

liangyuanpeng:

Now we will automatically upgrade the docker image in the main branch, let's treat release-3.4 and release-3.5 similarly

Releated: 
- https://github.com/etcd-io/etcd/issues/17609

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,3 +24,15 @@ updates:
     directory: /
     schedule:
       interval: weekly
+
+  - package-ecosystem: docker
+    directory: /
+    target-branch: "release-3.4"
+    schedule:
+      interval: monthly
+
+  - package-ecosystem: docker
+    directory: /
+    target-branch: "release-3.5"
+    schedule:
+      interval: monthly


jmhbnz (approved):
LGTM - Thanks @liangyuanpeng 

Based on my understanding of `target-branch` dependabot will scan `main`, detect an update is possible, and then raise a pull request against all three branches which seems like a good approach.

liangyuanpeng:
> If dependabot is scanning main but opening pr against release-3.5 I am not sure how it will figure out that there are actually 4x Dockerfiles in release-3.5 etc to update but lets give it a try and see what happens.

@jmhbnz  Thanks for your qulick reply.
I have verifyed in my fork repo and I think i forgot put the test link here, the dependabot will create the PR for `release-3.4` and `release-3.5`, 

check:
- https://github.com/liangyuanpeng/etcd/pull/17
- https://github.com/liangyuanpeng/etcd/pull/16

jmhbnz:
> I have verifyed in my fork repo and I think i forgot put the test link here, the dependabot will create the PR for `release-3.4` and `release-3.5`,

Brilliant, those pr's look perfect, thanks!

ahrtr:
I am not sure whether or not it's a good practice to always automatically bump to the latest version for the stable releases. Probably not. Usually we only bump dependencies for stable releases when there are any CVEs or critical/major bug fixes. The same for base image?

We can let this PR in, but I think we do need to figure out a best practice to bump the base image for the stable releases.

jmhbnz:
> I am not sure whether or not it's a good practice to always automatically bump to the latest version for the stable releases. Probably not. Usually we only bump dependencies for stable releases when there are any CVEs or critical/major bug fixes. The same for base image?

Good question - Historically we always used `latest` for image tag which meant our stable releases were always using latest image versions when releases were published. This pull request reduces manual effort for backports to keep using up to date base images, while still retaining reproducible builds via image digests so I think the pr is still a good idea.

ahrtr:
No strong opinion.

A temporary compromised solution I can think of is to bump the base image version for stable releases monthly instead of weekly, at least we have longer time to verify the base image in workflow before each patch releases.

liangyuanpeng:
> Usually we only bump dependencies for stable releases when there are any CVEs or critical/major bug fixes. The same for base image?

Make sense, this PR https://github.com/etcd-io/etcd/pull/17600 have not any CVE problem in the base image. 
If there is no CVE problem in the base image, then whether to upgrade the base image of the release branch depends on the release team. If the answer is no, then you only need to close the PR of dapadanbot is enough.

>  stable releases monthly instead of weekly,

Updated.

ahrtr (approved):
lgtm

Thanks

serathius (approved):

Issue #17615, opened 2024-03-19
Add metrics to record time costs of write system call in wal


Akiqqqqqqq:
### What would you like to be added?

We would like to have a metric to record time costs of **_write system call_** in wal in a single encoder.encode call

### Why is this needed?

Since our etcd cluster can go unstable with **constantly leader changes**, we tried to find the cause. 

But we can only observe that boltdb commits were slow(goes up to 4s at a time), other than that, any other metrics related to heartbeat loss, like wal fsync or cpu usage(goroutine cpu starvation) was normal.(fsync is up to 70ms at most)

We tried to look into it deeper, we use _pg_test_fsync_ and _biolalency_ , _ext4slower_ _ext4dist_ tools to locate the problem, at last we found the problem was at **_write system call_**, while the **_fsync system call_** was ok, **_write system call_** was so slow. So there was a kernel bug in **_write system call_**. After we upgraded kernel, the problem was solved.
ref: https://lore.kernel.org/all/20220817172139.3141101-1-shakeelb@google.com/

But still we cannot identify the direct cause of leader change at the beginning, only because there is only one metric recording the delay of **_fsync_** by wal, but the delay of **_write_** by wal is not counted. 

To summarize, we need a metric to record time costs of write system call in wal in a single encoder.encode call

tjungblu:
Sorry for being late to the party. Shall we also instrument the other parts of the encode block? I would also be interested in how long it would take to crc+marshal the data. Which should also help us to understand whether WAL writing is slow due to the CPU.

Also, please don't forget to update our docs at: https://etcd.io/docs/v3.5/metrics/#disk

Akiqqqqqqq:
> Sorry for being late to the party. Shall we also instrument the other parts of the encode block? I would also be interested in how long it would take to crc+marshal the data. Which should also help us to understand whether WAL writing is slow due to the CPU.
> 
> I reckon a metric for the full cycle of appending a record would also be useful.
> 
> Also, please don't forget to update our docs at: https://etcd.io/docs/v3.5/metrics/#disk edit: we could also think about adding an alert in https://github.com/etcd-io/etcd/tree/main/contrib/mixin/alerts (cc @Elbehery)

[...]

PR #17618, opened 2024-03-20
feat: add wal write system call metrics observation (main)

Akiqqqqqqq:
xref. #17615 

add new metrics: _**wal_write_duration_seconds**_

**_wal_write_duration_seconds_** represent total delay of: 
> syscall.write(padding_info + record_data_with_padding)

- delay of fsync is not included in _**wal_write_duration_seconds**_
- data: the raft log to write to the WAL on disk at one single raftNode loop

### DIFF
--- a/server/storage/wal/encoder.go
+++ b/server/storage/wal/encoder.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"os"
 	"sync"
+	"time"
 
 	"go.etcd.io/etcd/pkg/v3/crc"
 	"go.etcd.io/etcd/pkg/v3/ioutil"
@@ -84,17 +85,9 @@ func (e *encoder) encode(rec *walpb.Record) error {
 		data = e.buf[:n]
 	}
 
-	lenField, padBytes := encodeFrameSize(len(data))
-	if err = writeUint64(e.bw, lenField, e.uint64buf); err != nil {
-		return err
-	}
+	data, lenField := prepareDataWithPadding(data)
 
-	if padBytes != 0 {
-		data = append(data, make([]byte, padBytes)...)
-	}
-	n, err = e.bw.Write(data)
-	walWriteBytes.Add(float64(n))
-	return err
+	return write(e.bw, e.uint64buf, data, lenField)
 }
 
 func encodeFrameSize(dataBytes int) (lenField uint64, padBytes int) {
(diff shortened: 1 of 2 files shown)


jmhbnz (approved):
LGTM - Thanks @Akiqqqqqqq

Akiqqqqqqq:
@fuweid master, can help approve merge this PR?

server/storage/wal/metrics.go
```diff
@@ -28,6 +28,15 @@ var (
 		Buckets: prometheus.ExponentialBuckets(0.001, 2, 14),
 	})
 
+	walWriteSec = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: "etcd",
+		Subsystem: "disk",
```
serathius, line 33: why disk subsystem?

Akiqqqqqqq: not so sure, do you have any suggestion, since I saw `backend_commit_write_duration_seconds` in _server\storage\backend\metrics.go_ is also labeled as disk subsystem, which contains both write and fdatasync process

Akiqqqqqqq: to be precise, maybe we should use _file_system_ subsystem?

ahrtr: "disk" should be OK. It's consistent with the existing metrics `walWriteBytes`,

https://github.com/etcd-io/etcd/blob/1348062c23440a5f08b9834e427855a14af371e0/server/storage/wal/metrics.go#L31-L33

https://github.com/etcd-io/etcd/blob/1348062c23440a5f08b9834e427855a14af371e0/server/storage/wal/encoder.go#L95-L96

ahrtr:
Also update the metrics in `writeUint64`, just as what we already do for `walWriteBytes`?

https://github.com/etcd-io/etcd/blob/1348062c23440a5f08b9834e427855a14af371e0/server/storage/wal/encoder.go#L119-L120

Akiqqqqqqq:
> Also update the metrics in `writeUint64`, just as what we already do for `walWriteBytes`?
> 
> https://github.com/etcd-io/etcd/blob/1348062c23440a5f08b9834e427855a14af371e0/server/storage/wal/encoder.go#L119-L120



> Also update the metrics in `writeUint64`, just as what we already do for `walWriteBytes`?
> 
> https://github.com/etcd-io/etcd/blob/1348062c23440a5f08b9834e427855a14af371e0/server/storage/wal/encoder.go#L119-L120

will do

ahrtr (approved):
lgtm

Thanks

ahrtr:
cc @serathius

serathius (dismissed):

server/storage/wal/encoder.go
```diff
@@ -116,7 +119,9 @@ func (e *encoder) flush() error {
 func writeUint64(w io.Writer, n uint64, buf []byte) error {
 	// http://golang.org/src/encoding/binary/binary.go
 	binary.LittleEndian.PutUint64(buf, n)
+	start := time.Now()
 	nv, err := w.Write(buf)
+	walWriteSec.Observe(time.Since(start).Seconds())
```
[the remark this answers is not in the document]
serathius: I see you followed the same flow as for `walWriteBytes`, however the difference is that number of calls to cumulative `.Add` doesn't matter while `.Observe` on histogram does.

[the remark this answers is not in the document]
serathius: Observing a write of `writeUint64` as separate call would skew the histogram, as it's expected to write much smaller than size of data, then your 50%ile would be just `writeUint64`.

serathius (dismissed):
Please fix measurement of write so there is only 1 call to `.Observe`.

server/storage/wal/encoder.go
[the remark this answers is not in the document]
Akiqqqqqqq: I see, that make sense, will fix later

fuweid (approved):
LGTM

serathius (approved):

ahrtr:
Please also add a changelog for 3.6 if you don't backport the fix to 3.5; otherwise add a changelog for 3.5 after backporting.

Akiqqqqqqq:
> Please also add a changelog for 3.6 if you don't backport the fix to 3.5; otherwise add a changelog for 3.5 after backporting.

will do later

Akiqqqqqqq:
I backported to 3.5 and raise a changelog: https://github.com/etcd-io/etcd/pull/17728

Issue #17611, opened 2024-03-19
Watch API. blocked forever even with WithRequireLeader option


346999452:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

watch api blocked forever even with WithRequireLeader option

### What did you expect to happen?

no matter how the etcd server changes leader and restarts, as long as WithRequireLeader is set, it won't blocked

### How can we reproduce it (as minimally and precisely as possible)?

1. both 'Created' and 'Canceled' in pb.WatchResponse are True
2. the client receives an err immediately after receiving the response
3. err is HaltErr or ErrNoLeader

### Anything else we need to know?

clientv3/watch.go/run may 
1. blocked at w.wg.Wait() in defer
2. blocked at ws.outc not be closed

### Etcd version (please run commands below)

golang client version : v0.0.0-20200520232829-54ba9589114f
server version : etcd Version: 3.4.16

346999452:
Duplicated with this issue: https://github.com/etcd-io/etcd/issues/12534

PR #17622, opened 2024-03-21
client/v3: fix issue https://github.com/etcd-io/etcd/issues/17611

346999452:


ivanvc:
Hi, @346999452, thanks for your PR. It seems like your commit wasn't properly signed off. The proper syntax is

```
Signed-off-by: Name <email>
```

Could you please amend your commit and push it again? so the developer certificate of origin (DCO) check passes, i.e:


```bash
git commit --amend --signoff # Fix the signed off trailer
git push --force
```

serathius:
Can you provide a test to verify that the PR fixes the issue https://github.com/etcd-io/etcd/issues/17611?

346999452:
I simulated what I encountered:

`func (w *watchGRPCStream) serveWatchClient(wc pb.Watch_WatchClient) {
	for {
		resp, err := wc.Recv()
		if err != nil {
			select {
			case w.errc <- err:
			case <-w.donec:
			}
			return
		}
		if resp.Created {
			resp.Canceled = true
			resp.CancelReason = v3rpc.ErrGRPCPermissionDenied.Error()
		}
		select {
		case w.respc <- resp:
			if resp.Created && resp.Canceled && resp.CancelReason != "" {
				select {
				case w.errc <- errors.New("some err to test"):
				case <-w.donec:
				}
				return
			}
		case <-w.donec:
			return
		}
	}
}`

346999452:
> Can you provide a test to verify that the PR fixes the issue #17611?

The watch method is used as normal, it is difficult to reproduce the problem in real scenarios.
I simulated what I encountered and it could be easily reproduced:

The serveWatchClient method can be modified for easy reproduction

func (w *watchGRPCStream) serveWatchClient(wc pb.Watch_WatchClient) {
	for {
		resp, err := wc.Recv()
		if err != nil {
			select {
			case w.errc <- err:
			case <-w.donec:
			}
			return
		}
		if resp.Created {
			resp.Canceled = true
			resp.CancelReason = v3rpc.ErrGRPCPermissionDenied.Error()
		}
		select {
		case w.respc <- resp:
			if resp.Created && resp.Canceled && resp.CancelReason != "" {
				select {
				case w.errc <- errors.New("some err to test"):
				case <-w.donec:
				}
				return
			}
		case <-w.donec:
			return
		}
	}
}

346999452:
> @346999452, I think if you can provide (write) a unit or integration test for this, it would be the best.

```
func TestWatch(t *testing.T) {
	var client, err = clientv3.New(clientv3.Config{
		Endpoints:            []string{"127.0.0.1:2379", "127.0.0.1:3379", "127.0.0.1:4379"},
		DialTimeout:          1e11,
		Username:             "hello",
		Password:             "world",
		MaxCallSendMsgSize:   10485760,
		MaxCallRecvMsgSize:   107374182400,
		DialKeepAliveTime:    3e10,
		DialKeepAliveTimeout: 1e10,
		DialOptions:          []grpc.DialOption{grpc.WithBlock()},
	})
	if err != nil {
		panic(err)
	}
	var watcher = clientv3.NewWatcher(client)
	t.Log("to watch")
	var (
		resp        clientv3.WatchResponse
		watcherChan = watcher.Watch(context.Background(), "/hello/", clientv3.WithPrefix())
	)
	t.Log("watch success")
	for resp = range watcherChan {
		if resp.Err() != nil {
			panic(resp.Err())
		}
		t.Log("receive resp")
	}
}
```

There is nothing special about the unit testing, because the block is not caused by the way it is used, but by some extreme scenario
when the server is frequently faulty, a block may occur when the client rewatches

This code may not print 'watch success', or it may blocked while receiving from the watch channel

liangyuanpeng:
This will greatly increase the confidence of this PR if you can reproduce this problem in CI.

ivanvc:
I agree. Writing the test in `integration/clientv3/` would be better. Probably in `integration/clientv3/watch_test.go`.

346999452:
> This will greatly increase the confidence of this PR if you can reproduce this problem in CI.



> This will greatly increase the confidence of this PR if you can reproduce this problem in CI.

serathius:
Can you add a test?

PR #17624, opened 2024-03-21
server: address golangci var-naming issues

ivanvc:
This PR addresses minor var-naming violations from the server directory.

### DIFF
--- a/server/mock/mockstore/store_recorder.go
+++ b/server/mock/mockstore/store_recorder.go
@@ -119,7 +119,11 @@ func (s *storeRecorder) Clone() v2store.Store {
 	return s
 }
 
+//revive:disable:var-naming
 func (s *storeRecorder) JsonStats() []byte { return nil }
+
+//revive:enable:var-naming
+
 func (s *storeRecorder) DeleteExpiredKeys(cutoff time.Time) {
 	s.Record(testutil.Action{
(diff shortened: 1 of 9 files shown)


server/mock/mockstore/store_recorder.go
```diff
@@ -119,7 +119,11 @@ func (s *storeRecorder) Clone() v2store.Store {
 	return s
 }
 
+//revive:disable:var-naming
 func (s *storeRecorder) JsonStats() []byte { return nil }
+
+//revive:enable:var-naming
```
ivanvc, line 125: Added this ignores, as it comes from v2's `(*recorder) JsonStats().`

server/storage/backend/verify.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	ENV_VERIFY_VALUE_LOCK verify.VerificationType = "lock"
+	EnvVerifyValueLock verify.VerificationType = "lock"
```
ivanvc, line 29: Even though this is exported, I just renamed it as it is used only by the `backend_tests` package, which is in the same directory but in a different package.

ahrtr (approved):
lgtm

Thanks

jmhbnz (approved):
Nice work @ivanvc, another one tidied up!

PR #17634, opened 2024-03-21
all: call Close() after reading http.Response.Body

callthingsoff:
As the doc of http.Response.Body says:

// The http Client and Transport guarantee that Body is always
// non-nil, even on responses without a body or responses with
// a zero-length body. It is the caller's responsibility to
// close Body.

### DIFF
--- a/contrib/lock/client/client.go
+++ b/contrib/lock/client/client.go
@@ -69,6 +69,7 @@ func write(key string, value string, version int64) error {
 	if err != nil {
 		log.Fatalf("failed to read request body: %s", err)
 	}
+	httpResp.Body.Close()
 
 	resp := new(response)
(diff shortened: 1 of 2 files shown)


contrib/lock/client/client.go
```diff
@@ -66,6 +66,7 @@ func write(key string, value string, version int64) error {
 	}
 
 	respBytes, err := io.ReadAll(httpResp.Body)
+	httpResp.Body.Close()
```
ahrtr, line 69: Please put this line at line 73 (after the `if err != nil {} `).  When err != nil, usually means the response body has already closed.

This comment applies to following changes as well.

ahrtr (approved):

fuweid (approved):
LGTM

PR #17654, opened 2024-03-26
tests/robustness: address golangci var-naming issues

ivanvc:
Addresses the `var-naming` lint issues in `tests/robustness`.

In this case, too, there are some exported variables that were renamed, but as they're in the tests directory, I didn't think they would need to be deprecated rather than just renamed.

Part of #17578.


fuweid (approved):
LGTM

ahrtr (approved):
/lgtm


Thanks

serathius (approved):

PR #17655, opened 2024-03-26
tests/framework: address golangci var-naming issues [remove (*Member) GRPCURL()]

ivanvc:
Currently, (`test/framework/`'s) `integration.Member` exports `GrpcURL` already as a struct variable. However, when addressing the `var-naming` linting issues, renaming it from `GrpcURL` to `GRPCURL` clashes with the `GRPCURL()` function.

Given that it's already an exported variable, defining a getter function is unnecessary. This variable's use is also mixed, with calls to both the exported variable (`GrpcURL`) and the function (`GRPCURL()`).

If we prefer to keep the getter function `GRPCURL()`. Then, it would be better to unexport `GrpcURL` (`grpcURL`). But all of the other `GRPC*` variables are exported already. Therefore, it would make sense to unexport all of them, but would increase the rewrite effort.

Part of #17578.

### DIFF
--- a/tests/framework/integration/cluster.go
+++ b/tests/framework/integration/cluster.go
@@ -236,7 +236,7 @@ func (c *Cluster) Launch(t testutil.TB) {
 	c.WaitMembersMatch(t, c.ProtoMembers())
 	c.waitVersion()
 	for _, m := range c.Members {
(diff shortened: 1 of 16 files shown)


fuweid (approved):
LGTM

ahrtr (approved):

PR #17662, opened 2024-03-27
Add script and workflow to detect flaky tests in testgrid.

siyuanfoundation:

For https://github.com/etcd-io/etcd/issues/13167

Ideally this should live in the k8s test-infra or testgrid repo for it to be maintained long term. But that is a much bigger and longer effort that requires coordination with the sig-testing team. 

For now, I think it is beneficial to get it running while we seek long term solution with the sig-testing team. 

Example of auto created issue: https://github.com/siyuanfoundation/etcd/issues/6


siyuanfoundation:
cc @serathius @jmhbnz @ahrtr

scripts/testgrid_analysis/cmd/data.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2024 The etcd Authors
```
jmhbnz, line 1: Thanks for working on this @siyuanfoundation!

My instinct would be the golang utility would be better suited to live under `tools/`?

ahrtr: Suggest to put it under https://github.com/etcd-io/etcd/tree/main/tools

siyuanfoundation: moved to tools

tools/testgrid-analysis/cmd/github.go
```diff
@@ -0,0 +1,67 @@
...
+		createIssueIfNonExist(tab, t, openIssues, append(labels, "help wanted"))
+	}
+}
+
+func getOpenIssues(labels []string) []*github.Issue {
+	client := github.NewClient(nil).WithAuthToken(os.Getenv("GITHUB_TOKEN"))
+	ctx := context.Background()
+	// list open issues with label type/flake
+	issueOpt := &github.IssueListByRepoOptions{Labels: labels}
+	issues, _, err := client.Issues.ListByRepo(ctx, githubOwner, githubRepo, issueOpt)
```
ivanvc, line 38: Heads up that the GitHub API paginates the responses. The default number of items per page is 30 (https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues). Ideally, there should be at most 30 open issues for flaky tests, but if there are, it may create duplicate issues.

tools/testgrid-analysis/go.mod
```diff
@@ -0,0 +1,22 @@
+module go.etcd.io/etcd/tools/testgrid-analysis/v3
+
+go 1.22.1
```
ivanvc, line 3: I suggest not locking the minor as the rest of the `go.mod`s in the repository.

tools/testgrid-analysis/cmd/data.go
```diff
@@ -0,0 +1,146 @@
...
+func isParentTest(test string, allTests []string) bool {
+	for _, t := range allTests {
+		if t != test && strings.HasPrefix(t, test+"/") {
+			return true
+		}
+	}
+	return false
+}
+
+func fetchJson(url string) []byte {
```
ivanvc, line 123: I'm currently working on enabling the `var-naming` linter rule, and it will complain about this function. It should be named `fetchJSON(...)` :sweat_smile:

ivanvc (commented):
IMHO, the result is looking great :clap:
I left a couple of suggestions, and I hope it's ok :v:

tools/testgrid-analysis/cmd/github.go
siyuanfoundation: thanks for the pointer. Updated the logic to include pagination

ivanvc (approved):
👍 Thanks for addressing my comments

jmhbnz (approved):
Looks like a great first iteration - thanks @siyuanfoundation

ahrtr (approved):

serathius (approved):

PR #17674, opened 2024-03-30
server/embed: address golangci var-naming issues

ivanvc:
Addresses issues with `config.{TLSMinVersion,TLSMaxVersion,WALDir,MaxWALFiles}`.

This is part 1 of 2 from `server/embed`, as it's the package with the most changes.

Part of #17578.

### DIFF
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -149,8 +149,9 @@ func init() {
 
 // Config holds the arguments for configuring an etcd server.
 type Config struct {
-	Name   string `json:"name"`
-	Dir    string `json:"data-dir"`
+	Name string `json:"name"`
+	Dir  string `json:"data-dir"`
+	//revive:disable-next-line:var-naming
 	WalDir string `json:"wal-dir"`
 
 	SnapshotCount uint64 `json:"snapshot-count"`
@@ -163,7 +164,8 @@ type Config struct {
 	SnapshotCatchUpEntries uint64 `json:"experimental-snapshot-catch-up-entries"`
 
 	MaxSnapFiles uint `json:"max-snapshots"`
-	MaxWalFiles  uint `json:"max-wals"`
+	//revive:disable-next-line:var-naming
+	MaxWalFiles uint `json:"max-wals"`
 
 	// TickMs is the number of milliseconds between heartbeat ticks.
 	// TODO: decouple tickMs and heartbeat tick (current heartbeat tick = 1).
@@ -231,8 +233,11 @@ type Config struct {
 	CipherSuites []string `json:"cipher-suites"`
 
 	// TlsMinVersion is the minimum accepted TLS version between client/server and peers.
+	//revive:disable-next-line:var-naming
 	TlsMinVersion string `json:"tls-min-version"`
+
 	// TlsMaxVersion is the maximum accepted TLS version between client/server and peers.
+	//revive:disable-next-line:var-naming
 	TlsMaxVersion string `json:"tls-max-version"`
 
 	ClusterState          string `json:"initial-cluster-state"`
--- a/server/etcdmain/config_test.go
+++ b/server/etcdmain/config_test.go
@@ -57,7 +57,7 @@ func TestConfigFileMemberFields(t *testing.T) {
 	yc := struct {
 		Dir                    string `json:"data-dir"`
 		MaxSnapFiles           uint   `json:"max-snapshots"`
-		MaxWalFiles            uint   `json:"max-wals"`
+		MaxWALFiles            uint   `json:"max-wals"`
 		Name                   string `json:"name"`
 		SnapshotCount          uint64 `json:"snapshot-count"`
 		SnapshotCatchUpEntries uint64 `json:"experimental-snapshot-catch-up-entries"`
--- a/server/embed/util.go
+++ b/server/embed/util.go
@@ -21,9 +21,9 @@ import (
 )
 
 func isMemberInitialized(cfg *Config) bool {
-	waldir := cfg.WalDir
-	if waldir == "" {
-		waldir = filepath.Join(cfg.Dir, "member", "wal")
+	walDir := cfg.WalDir
+	if walDir == "" {
+		walDir = filepath.Join(cfg.Dir, "member", "wal")
 	}
-	return wal.Exist(waldir)
+	return wal.Exist(walDir)
 }


server/embed/config.go
```diff
@@ -151,7 +151,7 @@ func init() {
 type Config struct {
 	Name   string `json:"name"`
 	Dir    string `json:"data-dir"`
-	WalDir string `json:"wal-dir"`
+	WALDir string `json:"wal-dir"`
```
ahrtr, line 154: It will impact applications in which etcd is embedded.

PLease let's follow the same deprecation process to rename it. This comment applies to following changes as well.

ivanvc: @ahrtr, I don't think (or at least I can't think how) it's possible to deprecate a struct field; as Go doesn't support overriding the `=` operator, it would be impossible to know and reflect its value.

I can think of a couple of solutions:
1. Ignore these (`//revive:ignore ...`), and add a `TODO` to address them in the next major release
2. Deprecate, and everywhere the deprecated variable is referenced, call the non-deprecated version, i.e.
    ```diff
    -minVersion, err := tlsutil.GetTLSVersion(cfg.TLSMinVersion)
    +var tlsMinVersion string
    +if cfg.TlsMinVersion != "" {
    +  tlsMinVersion = cfg.TlsMinVersion
    +} else {
    +  tlsMinVersion = cfg.TLSMinVersion
    +}
    +minVersion, err := tlsutil.GetTLSVersion(tlsMinVersion)
    ```

However, as I mentioned, the fields won't be in sync. So, it adds the overhead of constantly checking both when referencing them. And could introduce bugs in external applications accessing these fields.

The problem with the first option is that it will likely not be addressed in the next major release.

Still, I prefer the first option. Thoughts?

ahrtr: We can add a method like below,
```
// getWALDir returns the `WALDir` if not empty, otherwise `WalDir`.
// TODO: remove this method once we remove the deprecated `WalDir`.
func (c *Config) getWALDir() string {
	if len(c.WALDir) > 0 {
		return c.WALDir
	}

	return c.WalDir
}
```

The whole diff for your reference,
```
$ git diff
diff --git a/server/embed/config.go b/server/embed/config.go
index 3bbdf0c9e..6696b7f33 100644
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -147,11 +147,24 @@ func init() {
        defaultHostname, defaultHostStatus = netutil.GetDefaultHost()
 }
 
+// getWALDir returns the `WALDir` if not empty, otherwise `WalDir`.
+// TODO: remove this method once we remove the deprecated `WalDir`.
+func (c *Config) getWALDir() string {
+       if len(c.WALDir) > 0 {
+               return c.WALDir
+       }
+
+       return c.WalDir
+}
+
 // Config holds the arguments for configuring an etcd server.
 type Config struct {
-       Name   string `json:"name"`
-       Dir    string `json:"data-dir"`
+       Name string `json:"name"`
+       Dir  string `json:"data-dir"`
+       // Deprecated: please use WALDir instead,
+       // to be decommissioned in 3.7
        WalDir string `json:"wal-dir"`
+       WALDir string `json:"wal-dir"`
 
        SnapshotCount uint64 `json:"snapshot-count"`
 
@@ -561,7 +574,7 @@ func NewConfig() *Config {
 func (cfg *Config) AddFlags(fs *flag.FlagSet) {
        // member
        fs.StringVar(&cfg.Dir, "data-dir", cfg.Dir, "Path to the data directory.")
-       fs.StringVar(&cfg.WalDir, "wal-dir", cfg.WalDir, "Path to the dedicated wal directory.")
+       fs.StringVar(&cfg.WALDir, "wal-dir", cfg.WALDir, "Path to the dedicated wal directory.")
        fs.Var(
                flags.NewUniqueURLsWithExceptions(DefaultListenPeerURLs, ""),
                "listen-peer-urls",
diff --git a/server/embed/etcd.go b/server/embed/etcd.go
index 42b9b78b9..2902f0ad9 100644
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -171,7 +171,7 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
                ClientURLs:                               cfg.AdvertiseClientUrls,
                PeerURLs:                                 cfg.AdvertisePeerUrls,
                DataDir:                                  cfg.Dir,
-               DedicatedWALDir:                          cfg.WalDir,
+               DedicatedWALDir:                          cfg.getWALDir(),
                SnapshotCount:                            cfg.SnapshotCount,
                SnapshotCatchUpEntries:                   cfg.SnapshotCatchUpEntries,
                MaxSnapFiles:                             cfg.MaxSnapFiles,
@@ -320,7 +320,7 @@ func print(lg *zap.Logger, ec Config, sc config.ServerConfig, memberInitialized
                zap.Bool("member-initialized", memberInitialized),
                zap.String("name", sc.Name),
                zap.String("data-dir", sc.DataDir),
-               zap.String("wal-dir", ec.WalDir),
+               zap.String("wal-dir", ec.getWALDir()),
                zap.String("wal-dir-dedicated", sc.DedicatedWALDir),
                zap.String("member-dir", sc.MemberDir()),
                zap.Bool("force-new-cluster", sc.ForceNewCluster),
diff --git a/server/embed/util.go b/server/embed/util.go
index 269fbc80b..d2ca91e70 100644
--- a/server/embed/util.go
+++ b/server/embed/util.go
@@ -21,7 +21,7 @@ import (
 )
 
 func isMemberInitialized(cfg *Config) bool {
-       waldir := cfg.WalDir
+       waldir := cfg.getWALDir()
        if waldir == "" {
                waldir = filepath.Join(cfg.Dir, "member", "wal")
        }
```

ivanvc: Done; the only issue is that Go doesn't allow setting the same JSON tag to two fields. So, I left the JSON tag set in the renamed (new) field. And because it's `go vet` that complains, I don't think I can add an ignore statement.

See:

```
% (cd server && 'go' 'vet' './...')
stderr: # go.etcd.io/etcd/server/v3/embed
stderr: # [go.etcd.io/etcd/server/v3/embed]
stderr: embed/config.go:158:2: struct field WALDir repeats json tag "wal-dir" also at config.go:156
stderr: embed/config.go:174:2: struct field MaxWALFiles repeats json tag "max-wals" also at config.go:172
stderr: embed/config.go:245:2: struct field TlsMinVersion repeats json tag "tls-min-version" also at config.go:242
stderr: embed/config.go:250:2: struct field TlsMaxVersion repeats json tag "tls-max-version" also at config.go:247
stderr: embed/config.go:455:2: struct field WALDir repeats json tag "wal-dir" also at config.go:156
stderr: embed/config.go:455:2: struct field MaxWALFiles repeats json tag "max-wals" also at config.go:172
stderr: embed/config.go:455:2: struct field TlsMinVersion repeats json tag "tls-min-version" also at config.go:242
stderr: embed/config.go:455:2: struct field TlsMaxVersion repeats json tag "tls-max-version" also at config.go:247

```

ivanvc: Done

ivanvc: Duh, right. What I was thinking is that when unmarshalling the field will only be at WALDir. Is that a use case? Or, it only marshals it?

ahrtr: JSON will marshal all exported fields using the fieldName by default, and overridden by the JSON tag if present. The same when unmarshalling it. I think we just need to explicitly clarify this.

ivanvc: Alright, I updated it. I hope that the message makes sense.

jmhbnz (approved):
LGTM - Thanks @ivanvc

serathius:
I don't think it's worth to be pedantic to complicate the config struct so much. If this is a blocker for enabling var-naming on whole repo, we should just exclude this file.

ahrtr:
> I don't think it's worth to be pedantic to complicate the config struct so much. If this is a blocker for enabling var-naming on whole repo, we should just exclude this file.

- Overall it isn't that complicated; this PR follows the same pattern to handle the 4 fields in Config struct. It is super clear and easy to understand.
- It's also backward compatible change.
- I won't insist on merging this PR is majority maintainers &members are against it, but I still think we should welcome & open to such change which improves readability, although it's minor.

serathius:
> but I still think we should welcome & open to such change which improves readability, although it's minor.

For me using linters like var-naming is not about readability, it's about code style consistency. Don't have strong preference about `WALDir` vs `WalDir`. However, duplicating fields definitely reduces readability.

ahrtr:
> For me using linters like var-naming is not about readability, it's about code style consistency.

Making code style consistent is part of readability. 



> However, duplicating fields definitely reduces readability.

It's a temporary solution to make the change backward compatible. We will remove one of them in 3.7. Also @ivanvc has been already following the same pattern in previous PRs.


Again, overall not a big problem, doesn't worth too much getting stuck on this. defer to majority maintainers & member's opinions.

ivanvc:
As the author of these Pull Requests, I left these changes as the last ones, as I felt they were the most controversial (and the decision agreed in this PR will impact follow-up PRs on other config fields with inconsistent naming). This time, getting to a position in which the change is backward compatibility was the most intricate. I initially suggested doing a linter ignore for these. Still, I think that if we don't deprecate the variable names now, it's improbable that we'll rename them later in another release without prior deprecation.

I think applying the linter rule helps with code style consistency and gives internal (and external) developers a better experience. For example, with the current state of the code (and inconsistencies with naming `Wal` and `WAL`), this could be potentially an end-user code that has an embedded etcd:

```golang
if cfg.MaxWalFiles != embed.DefaultMaxWALFiles {
  ...
}
```

Developers with an IDE may not have an issue with the naming, but it may not be very pleasant in other scenarios.

I'm advocating for enabling var-naming to have a more mature and formal naming for internal and external developers. It'd be good to hear from others and @fuweid, who worked to enable linter rules in the past.

server/embed/config.go
```diff
@@ -163,7 +170,14 @@ type Config struct {
 	SnapshotCatchUpEntries uint64 `json:"experimental-snapshot-catch-up-entries"`
 
 	MaxSnapFiles uint `json:"max-snapshots"`
-	MaxWalFiles  uint `json:"max-wals"`
+	// MaxWalFiles is the maximum number of write ahead log files stored in the wal directory.
```
fuweid, line 173: minor issue: add empty line before introducing `Deprecated`. For example, https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/io/ioutil/ioutil.go;l=87

fuweid (commented):
+1 for this change. We don't break command line interface and I believe it's good timing to do clean up. Yes, it needs some effort to maintain two fields. We can consider to add UT to cover that. 

> NOTE: When I re-enable lint, I just want to focus on existing lint and keep the change in small.

server/embed/config.go
ivanvc: Done. Thanks for the heads up. In previous PRs, I deprecated without adding that empty line. I added a task under #17578 to track this.

ivanvc:
@ahrtr, do we want any other team members to chime on this? Please cc who you think should have an opinion on this. Thanks.

ahrtr:
@serathius can we merge this PR?

serathius:
I don't like the change as it's the most controversial from all static analysis changes. It duplicates fields on user facing config which complicates the code and user interface. I want to discuss it further make sure we are not overzealous and just apply linters blindly. Making change now means we are committing to a new API, that we won't be able to clean up anytime soon. v3.7 will not come anytime soon. 

Nonetheless, the change is don't I will block the change if community supports it. @ahrtr feel free to merge if you think it's benefitial.

ahrtr:
> v3.7 will not come anytime soon.

Actually I am not insist on merging this PR. Due to bad maintenance history, each etcd minor release is actually a LTS version. If we have the same release cadence as K8s, e.g. a minor release each 3 ~ 4 months, then I will merge this PR. But the reality is we may release 3.7 at least 3+ years later based on the etcd release history.

@ivanvc So let's disable the linter for the fields in Config. Thanks for your work and sorry for the confusion.

server/etcdmain/config_test.go
```diff
@@ -57,7 +57,7 @@ func TestConfigFileMemberFields(t *testing.T) {
 	yc := struct {
 		Dir                    string `json:"data-dir"`
 		MaxSnapFiles           uint   `json:"max-snapshots"`
-		MaxWalFiles            uint   `json:"max-wals"`
+		MaxWALFiles            uint   `json:"max-wals"`
```
ivanvc, line 60: This is a struct exclusively used in this test case. It tests that marshaling the JSON fields works as expected, feeding it as a YAML config file. So, it should be fine to rename this field

ivanvc:
@serathius, @ahrtr PTAL 🙏

ahrtr (approved):
LGTM

Thank you !

server/etcdmain/config_test.go
serathius: Right

serathius (approved):

PR #17677, opened 2024-03-30
Fix release git url & document lessons learned

jmhbnz:
We recently released etcd `v3.5.13`. This pull request captures lessons from the release to ensure future releases can be smoother.

1. Versions specified on the release script must not be prefixed.
2. Git remote should be ssh format since pushing to github with https requires tokens now.
3. Network timeouts can occur pushing images, but it is safe to re-run the script.
4. Release script requires signing tags with gpg keys, ssh signing is not supported yet.

### DIFF
--- a/Documentation/contributor-guide/release.md
+++ b/Documentation/contributor-guide/release.md
@@ -72,10 +72,10 @@ which don't need to be executed before releasing each version.
 3. Clone the etcd repository and checkout the target branch,
    - `git clone git@github.com:etcd-io/etcd.git`
    - `git checkout release-3.X`
-4. Run the release script under the repository's root directory,
+4. Run the release script under the repository's root directory, replacing `${VERSION}` with a value without the `v` prefix, i.e. `3.5.13`.
    - `DRY_RUN=false ./scripts/release.sh ${VERSION}`
 
-   It generates all release binaries under directory ./release and images. Binaries are pushed to the google cloud bucket
+   It generates all release binaries under directory `./release` and images. Binaries are pushed to the google cloud bucket
    under project `etcd-development`, and images are pushed to `quay.io` and `gcr.io`.
 5. Publish release page in Github
    - Set release title as the version name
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -61,7 +61,7 @@ main() {
       REPOSITORY=$(pwd)
       BRANCH=$(git rev-parse --abbrev-ref HEAD)
   else
-      REPOSITORY=${REPOSITORY:-"https://github.com/etcd-io/etcd.git"}
+      REPOSITORY=${REPOSITORY:-"git@github.com:etcd-io/etcd.git"}
```
jmhbnz, line 64: If this change is accepted I will backport it to `release-3.4` and `release-3.5`.

Documentation/contributor-guide/release.md
```diff
@@ -100,3 +100,9 @@ etcd team
 7. Update the changelog to reflect the correct release date.
 8. Paste the release link to the issue raised at step 1 and close the issue.
 9. Crease a new stable branch through `git push origin release-${VERSION_MAJOR}.${VERSION_MINOR}` if this is a new major or minor stable release.
+
+#### Release known issues
+
+1. Network timeouts - It is rare, although possible for connection timeouts to occur when publishing etcd release images to `quay.io` or `gcr.io`. If this occurs, for `release-3.5` or newer it is known to be safe to rerun the release script command, and image uploads will gracefully resume.
```
ahrtr, line 106: I believe it's the same for 3.4. But it'd be better to double check this.
```suggestion
1. Network timeouts - It is rare, although possible for connection timeouts to occur when publishing etcd release images to `quay.io` or `gcr.io`. If this occurs,  rerun the release script command, and image uploads will gracefully resume.
```

We also need to clearly mention,
- If the release script fails during pushing image, rerun the script with flag `--no-upload`;
- If the release script fails during uploading binaries, then remove all binaries from google buckets, and rerun the script.

jmhbnz: > I believe it's the same for 3.4. But it'd be better to double check this.

Read through the script for `release-3.4`, yes it's the same behaviour which is great, updated.

> If the release script fails during pushing image, rerun the script with flag --no-upload

Good catch yes we used that flag, updated.

> If the release script fails during uploading binaries, then remove all binaries from google buckets, and rerun the script.

Added section. No need to remove binaries. I confirmed we don't use object versioning on our bucket and `gsutil cp` will overwrite existing file (unless you use `-n` flag which we don't) so there is no chance incorrect files can remain.

scripts/release.sh
ahrtr: > If this change is accepted I will backport it to `release-3.4` and `release-3.5`.

It's OK to backport the change to release-3.5. But let's keep 3.4 as it's. I haven't seen any issues so far when releasing 3.4, we don't really want to any surprise.

jmhbnz: Fair enough. We haven't seen the issue because we have been manually creating and pushing tag.

jmhbnz:
/retest

Flake already reported.

ahrtr (approved):
lgtm

Thanks

serathius (approved):

scripts/release.sh
ivanvc: The scheme/protocol for cloning a repository can be a personal preference. If you have properly configured Git for HTTP authentication, it should work fine. You could potentially use GitHub without an SSH key.

That's why I think the scheme could be opinionated. Maybe a good middle ground would be to receive an option `--use-https-repository-url` (or `--use-ssh-repository-url`), to switch from https and ssh, and keep the release scripts more in sync... Or we could also leave it as-is, according to each release manager preference :)

PR #17679, opened 2024-04-01
Add trivy scan action for current branch HEAD

ArkaSaha30:
This PR will add trivy scan action to run nightly on the current branch's HEAD(etcd image version), in this case-`main`.

Ref: https://github.com/etcd-io/etcd/pull/17596
Discussion: https://github.com/etcd-io/etcd/pull/17596#issuecomment-2004109139


ArkaSaha30:
cc @ahrtr @jmhbnz

ahrtr:
I am thinking probably we should follow the same way as [workflows/release.yaml](https://github.com/etcd-io/etcd/blob/main/.github/workflows/release.yaml), we just need to add one more step to scan the image (e.g. etcd:3.6.99) built in previous step in the workflow.

Next steps will be also doing the same thing for release-3.5 and release-3.4.

What do you think?

jmhbnz:
> I am thinking probably we should follow the same way as [workflows/release.yaml](https://github.com/etcd-io/etcd/blob/main/.github/workflows/release.yaml), we just need to add one more step to scan the image (e.g. etcd:3.6.99) built in previous step in the workflow.

> What do you think?

Agree we could make the scan part of release workflow to test images that are built rather than have separate workflow.

ArkaSaha30:
Apart from the scan being triggered via PR/push action, do we need it to be periodic as well? In that case we might need to have a separate workflow.

jmhbnz:
/ok-to-test

Please address minor static analysis failure coming from `make fix`. The diff it prints for me is:

```diff
-      security-events: write  # for github/codeql-action/upload-sarif to upload SARIF results
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results

```

ArkaSaha30:
The scan results are currently named `trivy-results-<platforms>.sarif`. Do we need to add any other information in the file name?

.github/workflows/release.yaml
```diff
@@ -32,3 +32,41 @@ jobs:
...
+        with:
+          name: etcd-img
+          path: /tmp/etcd-img.tar
+          retention-days: 1
+  trivy-scan:
+    needs: main
+    strategy:
+      fail-fast: false
+      matrix:
+        platforms: [amd64, arm64, ppc64le, s390x]
```
jmhbnz, line 49: Do we need multiple platforms here?  It looks like the same image would be loaded each time regardless?  I might be missing something though.

jmhbnz (commented):
> The scan results are currently named trivy-results-<platforms>.sarif. Do we need to add any other information in the file name?

Filename is ok to me. Question about platform inclusion below though.

.github/workflows/release.yaml
ahrtr: It's multi-arch image.  FYI. https://github.com/etcd-io/etcd/blob/547aa643e69d55328b203f40ccec50256b44158e/scripts/release.sh#L270-L276


Regarding whether to verify all platforms, it seems not necessary; I am not sure whether it's possible that some CVEs only exist for some platform/archs. But it won't do any harm to verify all archs.

.github/workflows/release.yaml
```diff
@@ -32,3 +32,41 @@ jobs:
...
           VERSION=3.6.99 ./scripts/test_images.sh
+      - name: save-image
+        run: |
+          docker image save -o /tmp/etcd-img.tar gcr.io/etcd-development/etcd
+      - name: upload-image
+        uses: actions/upload-artifact@v3
+        with:
+          name: etcd-img
+          path: /tmp/etcd-img.tar
+          retention-days: 1
```
ahrtr, line 43: Why do we need to save & upload the image? Can we just add one more step `scan-image` in the same way as previous step `test-image`?

ArkaSaha30: Since we are using matrix to run parallel jobs, if we add a step, the entire `main` job with all the steps will be replicated for all the archs. 
In my understanding, the step `release` which might be resource-demanding will be replicated for all the archs, each time doing the exact same thing.

ahrtr: Thanks for the clarification. 

Have you verified that `docker image save ...; docker load ...` can recover all images for different archs?

ahrtr: Confirmed that it indeed loads all images for different archs.

```
Run docker load < /tmp/etcd-img.tar
  docker load < /tmp/etcd-img.tar
  shell: /usr/bin/bash -e {0}
Loaded image: gcr.io/etcd-development/etcd:v[3](https://github.com/etcd-io/etcd/actions/runs/8523801808/job/23347153138?pr=17679#step:4:3).6.99-ppc64le
Loaded image: gcr.io/etcd-development/etcd:v3.6.99-s390x
Loaded image: gcr.io/etcd-development/etcd:v3.6.99-amd6[4](https://github.com/etcd-io/etcd/actions/runs/8523801808/job/23347153138?pr=17679#step:4:5)
Loaded image: gcr.io/etcd-development/etcd:v3.6.99-arm64
```

.github/workflows/release.yaml
```diff
@@ -32,3 +32,41 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: get-image
+        uses: actions/download-artifact@v3
+        with:
+          name: etcd-img
+          path: /tmp
+      - name: load-image
+        run: |
+          docker load < /tmp/etcd-img.tar
```
ahrtr, line 61: It seems that the workflow does the same steps multiple times, but actually we can only do it once. Not sure how to resolve this. But overall not a big problem.

jmhbnz (approved):
LGTM - Thanks @ArkaSaha30

ahrtr (approved):
Thanks @ArkaSaha30

ahrtr:
Please also backport (will be slightly different) the PR to 3.5 and 3.4 as well.

.github/workflows/release.yaml
ArkaSaha30: I agree, I tried to club common steps in `main` job as much as possible but there is no direct data retention/sharing across jobs.

ArkaSaha30:
> Please also backport (will be slightly different) the PR to 3.5 and 3.4 as well.

Sure, I will raise backport PRs for 3.4 and 3.5 respectively.

Issue #15480, opened 2023-03-14
Move grpc-proxy to separate repo


serathius:
### What would you like to be added?

Extract grpc-proxy from etcd and move it to separate project.

### Why is this needed?

Complexity of testing grpcproxy greatly reduced our ability to reduce flakiness of etcd tests.

liangyuanpeng:
Hi @serathius , Could you provide more detail for it? where can begin?

liangyuanpeng:
Try to disassemble this issue

- Migrate etcd-grpc-proxy to a separate repo, and migrate the logic related to etcd-grpc-proxy startup, so you need to create a repo first, I have created such a repo,  check https://github.com/liangyuanpeng/etcd-grpc-proxy
- Migrate test scripts from etcd repo, including integration tests and e2e tests. (https://github.com/etcd-io/etcd/blob/main/.github/workflows/grpcproxy.yaml)
- Delete etcd-grpc-proxy related code from etcd repo.

Two questions :
Does etcd still need to support the etcd grpx-proxy command, it's mean the built-in startup etcd-grpc-proxy with etcd.
What is the release process of etcd-grpc-proxy? Need to follow etcd version?

Remind me if i missed something,Thanks.

/cc @serathius

fuweid:
@liangyuanpeng I think you're right. If we decide to separate the proxy to another repo, it needs maintainer to init project at first :)

PR #17691, opened 2024-04-02
tests/e2e: ignore discovery tests when cluster_proxy tag is set

### LINKED ISSUE #16098: TestClusterOf1UsingDiscovery doesn't happy if there is last release of etcd
### Which github workflows are flaking?

no

### Which tests are flaking?

TestClusterOf1UsingDiscovery

### Github Action link

no

### Reason for failure (if possible)

https://github.com/etcd-io/etcd/blob/8de14bd36e846eef7bc0d66924a28e1352bb5a72/tests/e2e/discovery_test.go#L56

When we enable `cluster_proxy` as tags, the [ProxyV2](https://github.com/etcd-io/etcd/blob/8de14bd36e846eef7bc0d66924a28e1352bb5a72/tests/framework/e2e/cluster_proxy.go#L209) doesn't have a chance to start because we just define it. And in our testing workflow `test-grpcproxy-e2e`, we don't download last release binary so that the case never run. It only happens in local test.

If the ProxyV2 is not used any more, I think we should delete it and add the `!cluster_proxy` tag for the test.

ivanvc:
Ignore `e2e/discovery_test.go` when `cluster_proxy` tag is set.

Fixes #16098.

### DIFF
--- a/tests/e2e/discovery_test.go
+++ b/tests/e2e/discovery_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !cluster_proxy
+
 package e2e
 
 import (


jmhbnz (approved):
LGTM - Thanks @ivanvc

serathius (approved):
/hold
For @ahrtr

ahrtr:
Thanks for the fix.  Confirmed that the discovery test cases have never been run at all, because they are skipped due to `etcd-last-release` not exist.

As a long term goal, we should have a summary on all the existing test cases for grpc-proxy. Is it possible move all grpc-proxy test cases into a dedicated sub-folder? cc @jmhbnz @ivanvc

ahrtr (approved):

fuweid (approved):
LGTM

serathius:
> As a long term goal, we should have a summary on all the existing test cases for grpc-proxy. Is it possible move all grpc-proxy test cases into a dedicated sub-folder? cc @jmhbnz @ivanvc

We could have a prow scenario to run grpc-proxy test and have a dedicated dashboard like for e2e test? https://testgrid.k8s.io/sig-etcd-periodics#ci-etcd-e2e-amd64

ivanvc:
We could move the grpc-proxy tests to its own sub dir. There's also this issue open #15480, which seems to be low in priority, but if we're going to move the tests, would it make more sense just to move it all together into its own repository? However, the effort of the latter is definitely on the higher side.

PR #17710, opened 2024-04-04
tools/benchmark/cmd: don't panic with nil values in hashKV function

callthingsoff:
hashKV checks nilness of 'eh' and 'es', but doesn't use them. Instead
it panics with 'err' which is definitely nil.

This should be avoided, See:
https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/nilness#nilpanic

This patch renames 'eh' and 'es' to 'err' in order to correct the real
errors to panic.

### DIFF
--- a/tools/benchmark/cmd/put.go
+++ b/tools/benchmark/cmd/put.go
@@ -170,14 +170,14 @@ func hashKV(cmd *cobra.Command, clients []*v3.Client) {
 	host := eps[0]
 
 	st := time.Now()
-	rh, eh := clients[0].HashKV(context.Background(), host, 0)
-	if eh != nil {
-		fmt.Fprintf(os.Stderr, "Failed to get the hashkv of endpoint %s (%v)\n", host, eh)
+	rh, err := clients[0].HashKV(context.Background(), host, 0)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


ivanvc:
@henrybear327, there are many failures in your build. This could possibly be because of the update to Go 1.22.2. Can you rebase your branch with main?

henrybear327:
> @henrybear327, there are many failures in your build. This could possibly be because of the update to Go 1.22.2. Can you rebase your branch with main?

@ivanvc, sorry I don't get what you mean. Which branch are you referring to?

ivanvc:
Sorry @henrybear327, I meant @callthingsoff :sweat_smile:

tools/benchmark/cmd/put.go
```diff
@@ -173,12 +173,12 @@ func hashKV(cmd *cobra.Command, clients []*v3.Client) {
 	rh, eh := clients[0].HashKV(context.Background(), host, 0)
```
ivanvc, line 173: Another option would be to re-use `err` here and in line 178. And keep the panics as they were before.

serathius (approved):

callthingsoff:
Need a retest?

ahrtr (approved):

Issue #17717, opened 2024-04-04
etcd robustness test on Prow is flaky


siyuanfoundation:
### Which Github Action / Prow Jobs are flaking?

https://testgrid.k8s.io/sig-etcd-periodics#ci-etcd-robustness-amd64

### Which tests are flaking?

The flakiness is not specific to any particular test. 
For each prow job, there is always 1 test run failing. 
I don't think it is specific to Lazyfs. 

The majority of failures report error like `Requiring minimal 200.000000 qps for test results to be reliable, got 72.143568 qps`.

It is always the Run0 that fails. Not sure if prow just orders the results so that failed test ranks to the top or if there any warmup issue. 

### Github Action / Prow Job link

https://testgrid.k8s.io/sig-etcd-periodics#ci-etcd-robustness-amd64

siyuanfoundation:
cc @jmhbnz @BenTheElder @ArkaSaha30

BenTheElder:
Possible quick solution: Use a memory backed emptyDir for this tests's scratch space (configure this in the prowjob pod, make sure the test setup understands to use the mounted path for scratch space)

siyuanfoundation:
> Possible quick solution: Use a memory backed emptyDir for this tests's scratch space (configure this in the prowjob pod, make sure the test setup understands to use the mounted path for scratch space)

Thanks Ben for the suggestion. But for this test, I think we actually want to test with disk io.

BenTheElder:
At least on the GKE clusters, we're running on local SSD IIRC, but we can't make much more guarantee than that about I/O.

If this job is reduced to fit on n1-highmem-8 (and leave room for system-reserved) then it could run on the k8s infra prow builds GKE cluster and see if that's better.

I'm not sure what the disk config is on EKS, upodroid or @xmudrii would know better.

xmudrii:
@BenTheElder @siyuanfoundation The disk config on EKS is also local NVME SSDs, so disk performance shouldn't be a problem. However, EKS nodes are large which contributes to larger bin-packing. If your test lands on a node that already has other tests running, it might affect I/O performance.

I see two options:

- Try to reduce the job to fit on n1-highmem-8 on the GKE cluster
- Increase resources for the job so that we have only that one job running on the node in the EKS cluster

siyuanfoundation:
[...]

PR #17725, opened 2024-04-05
robustness: make qps below threshold retriable test.

siyuanfoundation:

For issue https://github.com/etcd-io/etcd/issues/17717

Errors like `Requiring minimal 200.000000 qps for test results to be reliable, got 72.143568 qps.` could be temporary. Proposing to make it retriable.

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -33,6 +33,10 @@ import (
 	"go.etcd.io/etcd/tests/v3/robustness/validate"
 )
 
+const (
+	robustnessRetries = 3
+)
+
 var testRunner = framework.E2eTestRunner
 
 func TestMain(m *testing.M) {
@@ -46,7 +50,7 @@ func TestRobustnessExploratory(t *testing.T) {
 			lg := zaptest.NewLogger(t)
 			scenario.cluster.Logger = lg
 			ctx := context.Background()
-			testRobustness(ctx, t, lg, scenario)
+			testRobustnessWithRetry(ctx, t, lg, scenario)
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


jmhbnz (approved):
I like the idea - nice work @siyuanfoundation

siyuanfoundation:
cc @serathius @ahrtr

serathius (changes requested):
Don't like the idea, if we can get the performance required on github action, we should be able to get it on Prow. If you look at the recent feailures in github actions they are not directly related to low qps, it's just consequence of other issue.

serathius:
> I still think it's better than current situation where one single performance flake throws off entire robustness run.

This is a separate, unrelated problem. Robustness test reports are written always to the same directory dependent on the name of the test. This means that we cannot have multiple failures in run as it would override the previous result. I'm open to changing it, if you think this is a problem, but I don't think that stopping on the first failure is bad. At least as long we maintain number of failures low, and there is no sense to have robustness test with high failure rate.

jmhbnz:
> Don't like the idea, if we can get the performance required on github action, we should be able to get it on Prow. If you look at the recent feailures in github actions they are not directly related to low qps, it's just consequence of other issue.

It's not necessarily that we can't get the qps, clearly later iterations on prow do. For some reason the first one doesn't.

Putting that aside I still think this feature makes sense. Why wouldn't we make the test suite more resilient to these annoying performance flakes?

serathius:
Same like we don't slap retries on every test, we need to understand the underlying problem. I would need to understand what is the problem with Prow first.

siyuanfoundation:
Here is my line of thought. If 1% of tests do not satisfy the qps min, is that really a problem? From the error message, it just means that test results is not reliable, not robustness failure. 

Since hundreds of tests are run for each day, very likely (99%^100=36%) some test is going to fail due to this reason. But we cannot just ignore them as "Flaky", because each run is with an exploratory config, 1 test run failure can mean a real issue. So  1% of tests not satisfying the qps min means there needs manual check almost everyday. Is it really worth it?

tests/robustness/main_test.go
```diff
@@ -124,7 +153,7 @@ func (s testScenario) run(ctx context.Context, t *testing.T, lg *zap.Logger, clu
 	maxRevisionChan := make(chan int64, 1)
 	g.Go(func() error {
 		defer close(maxRevisionChan)
-		operationReport = traffic.SimulateTraffic(ctx, t, lg, clus, s.profile, s.traffic, finishTraffic, baseTime, ids)
+		operationReport, retriableErr = traffic.SimulateTraffic(ctx, t, lg, clus, s.profile, s.traffic, finishTraffic, baseTime, ids)
```
MadhavJivrajani, line 156: nit: I think considering that this is the only place we are setting `retriableErr`, this should be fine, but in the future if we set it in any of the other goroutines, that might be a race.

How about returning the err instead of `nil` in this goroutine and catching it as part of `g.Wait()` below?

tests/robustness/main_test.go
```diff
@@ -114,8 +143,8 @@ func (s testScenario) run(ctx context.Context, t *testing.T, lg *zap.Logger, clu
 		defer close(finishTraffic)
 		err := failpoint.Inject(ctx, t, lg, clus, s.failpoint)
 		if err != nil {
-			t.Error(err)
 			cancel()
+			t.Fatal(err)
```
MadhavJivrajani, line 147: Just trying to understand this better, if we use `t.Fatal` here, we won't end up retrying the test right?
Referring to the comment here: https://github.com/etcd-io/etcd/pull/17725#issuecomment-2041203978

serathius:
> Here is my line of thought. If 1% of tests do not satisfy the qps min, is that really a problem?

No it's not as long we investigate the reason, track flakiness changes over time and not hide it behind retries. Retrying obscures the test results without understanding the underlying issue. Low qps not only is caused by infrastructure noise, but also by other underlying availability issues. Only by observing all the flakes, and categorizing them we can discover issues like https://github.com/etcd-io/etcd/issues/17455

Of cause categorizing issues in robustness test, brings additional toil. For now I'm happy to handle it, with plans to share the knowledge in the community via dedicated meetings, and potentially automating the process at some point by string matching.

PR #17729, opened 2024-04-07
fix: close profile in the mvccPutFunc func

testwill:


tools/benchmark/cmd/mvcc-put.go
```diff
@@ -94,6 +94,7 @@ func mvccPutFunc(_ *cobra.Command, _ []string) {
 		}
 
 		defer func() {
+			f.Close()
```
ahrtr, line 97: The following `pprof.WriteHeapProfile(f)` will run into issue because the file has already been closed? So suggest to add a `defer f.Close()` above (right after the `if err !=nil` check).

ahrtr (approved):
lgtm

Thanks

serathius (approved):

PR #17733, opened 2024-04-08
Document re-evaluating existing robustness test reports

serathius:
cc @ahrtr @siyuanfoundation @MadhavJivrajani

PR #17734, opened 2024-04-08
tests: add toolchain directive to go.mod

MadhavJivrajani:
Add toolchain directive in order to successfully run tests from root. Else, go will try and download a family of releases (of the form 1.x), which are not published binaries.

Ref: https://github.com/etcd-io/etcd/pull/17733#discussion_r1555501552


ahrtr (approved):

tests/go.mod
```diff
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/tests/v3
 
 go 1.22
 
+toolchain go1.22.2
```
jmhbnz, line 5: Thanks for raising this @MadhavJivrajani. Please see my suggestion here https://github.com/etcd-io/etcd/pull/17733#discussion_r1557058353.  Keen to see if we can avoid putting hardcoded go version into more files if possible as we update frequently and it might be easy to miss updating it in this lower version `go.mod`.

serathius:
Looks like I didn't keep up with latest Golang features. Don't know toolchain feature enough to make long term decision yet, however short term would be good to be consistent. The choice is either `.go-version` file with https://github.com/etcd-io/etcd/pull/17733#discussion_r1557058353, or migrate to `toolchain`.

I think until we decide to remove `.go-version` file, I would prefer it to be source of go version in the project.

ahrtr:
Agreed to use `.go-version` as the single source of truth of the go&toolchain version. Then I think we may want to update the build script to use the `.go-version` as the toolchain version as well.

serathius (approved):

MadhavJivrajani:
Hey all, getting back to this PR

@jmhbnz that's a good point. I've made the change that uses the `.go-version` file instead to set the `GOTOOLCHAIN` env var.

Also, as next steps, we could do what Kubernetes does: https://github.com/kubernetes/community/blob/aea3ba11fc4dee0ae21f20c7412c732b79c666eb/contributors/devel/development.md#building-kubernetes-using-a-specific-version-of-go

Maybe have a way of setting the `GOTOOLCHAIN` env var as a prereq to running tests in general.

tests/go.mod
MadhavJivrajani: Done, thanks!

ahrtr:
Please feel free to merge this PR if any maintainer wants. 

But we need to revisit the go version management. Please read https://github.com/etcd-io/etcd/issues/17857

serathius:
> Please feel free to merge this PR if any maintainer wants.
> 
> But we need to revisit the go version management. Please read #17857

Right, want to merge as the robustness command apparently doesn't work for @MadhavJivrajani

Issue #17754, opened 2024-04-08
Add Prow job to run periodic performance benchmarks


ivanvc:
### What would you like to be added?

As part of #16467, create a Prow job that nightly runs the rw-benchmarks. Storing the results as an artifact so results can be processed later and have an idea of how the development is affecting the project's performance.

### Why is this needed?

To start collecting performance benchmarks to establish a project perspective on performance.

jmhbnz:
Added to our tracking under https://github.com/etcd-io/etcd/issues/16467.

serathius:
Thanks for integrating benchmarks into the prow, however I was hoping that we will do the work in semi order describe in https://github.com/etcd-io/etcd/issues/16467#issuecomment-1691326907, so before we start running periodic benchmark we agree on what to measure first.

ivanvc:
This makes sense, @serathius; thanks for your input. @jmhbnz, should we regroup and formalize a proposal?

jmhbnz:
> This makes sense, @serathius; thanks for your input. @jmhbnz, should we regroup and formalize a proposal?

Certainly. I am currently taking some time off to move house but will be returning to open source work 6th May and very keen to regroup when I return. Feel free to update https://github.com/etcd-io/etcd/issues/16467 with any thoughts in the interim and thanks again for your work getting these initial tests of prow for heatmaps running @ivanvc

PR #17428, opened 2024-02-14
Reimplement tools/rw-heatmaps in go

ivanvc:
This PR translates Python's tools/rw-heatmaps to Go. It covers the use case in which the Python script takes a single input and plots a grid of heatmap (tripcolor) charts.

The Go implementation uses gonum.org/v1/plot, a plotting solution that could generate similar style charts while creating images. Unfortunately, it doesn't provide an equivalent to Python's tripcolor, so I'm doing a heatmap instead. While trying to keep the Go implementation similar to the original one, I used the closest color palette to the one that Python used (however, the color scale for Go's palette is inverted, so it needs to be inverted to have a similar color palette). 

I introduced a `go.mod` in the `tools/rw-heatmaps` directory. Because the `tools/` directory doesn't have a module defined. Therefore, any dependencies from this rewrite were getting into the top-level's `go.mod`. In the future, adding a module in `tools/` would be good if more tools are written in Go. Right now, the ones written in Go are not introducing any new dependencies, so the top level doesn't carry these dependencies.

### Python implementation charts

python_read
python_write

### Go implementation charts

go_read
go_write

### Used CSV

The CSV file used to generate the previous plots was this one: [result-202401270029.csv](https://github.com/etcd-io/etcd/files/14378480/result-202401270029.csv)

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #17755, opened 2024-04-09
tools/rw-heatmaps: allow overriding running parameters

ivanvc:
Currently, the running parameters in the `rw-benchmark.sh` script are hardcoded. Running this script end to end takes around 48 hours with 7 CPU cores. To split the load into smaller jobs, allow these variables to be provided at run time.

Part of #17754.


tools/rw-heatmaps/rw-benchmark.sh
```diff
@@ -2,7 +2,7 @@
 
 #set -x
 
-RATIO_LIST="1/128 1/8 1/4 1/2 2/1 4/1 8/1 128/1"
+RATIO_LIST="${RATIO_LIST:-1/128 1/8 1/4 1/2 2/1 4/1 8/1 128/1}"
```
jmhbnz, line 5: This is good thanks, suggest we extend it to other key dimensions below like `RUN_COUNT` and `KEY_SIZE` at the same time. Most of the script parameters should be tunable via emvironment vars imo.

serathius:
> Running this script end to end takes around 48 hours with 7 CPU cores. 

This is definitely not approachable. 

cc @jmhbnz 

When we propose list of etcd SLOs that we want to measure, we should also propose more sensible set of samples to test.

ivanvc:
@serathius, I agree. That's the reason behind parameterizing this script.

tools/rw-heatmaps/rw-benchmark.sh
ivanvc: Hi @jmhbnz, as I was working on enabling the rest of the arguments. I realized that the script allows setting some of the arguments (mostly not performance-related). My question is, do you prefer execution arguments or environment variables? I don't have a strong preference, although sometimes it's easier to use environment variables for automated jobs, and arguments are easier for end-user execution. We could support both but I'm not sure if that would be an overkill.

Ref: https://github.com/etcd-io/etcd/blob/b9b7c9328e6f37db1de9a214d414bd2aa02d391e/tools/rw-heatmaps/rw-benchmark.sh#L111-L137

jmhbnz: Lets implement optional environment variables for these uppercase variables as a basic starting point: https://github.com/etcd-io/etcd/blob/b9b7c9328e6f37db1de9a214d414bd2aa02d391e/tools/rw-heatmaps/rw-benchmark.sh#L5-L26.

Over time I am happy for more command line parameters to be added if we think it will improve user experience but it's not critical right now imo.

ivanvc: Agreed. Do you also want one for `CLIENT_PORT`? I initially left that one out.

jmhbnz: Yeah let's implement environment variable support for all those uppercase vars. I don't see why client port should be hard coded as etcd server allows it to be configurable.

ivanvc: Done. I also added a `.gitignore` to address the previous PR's last comment: https://github.com/etcd-io/etcd/pull/17428#pullrequestreview-1926814342.

jmhbnz (approved):
LGTM - Thanks @ivanvc

serathius (approved):

PR #17756, opened 2024-04-09
Don't cache test-robustness-reports

serathius:
cc @ahrtr @jmhbnz @MadhavJivrajani @siyuanfoundation

### DIFF
--- a/tests/robustness/makefile.mk
+++ b/tests/robustness/makefile.mk
@@ -1,6 +1,6 @@
 .PHONY: test-robustness-reports
 test-robustness-reports:
-	cd ./tests && go test ./robustness/validate -v --run TestDataReports
+	cd ./tests && go test ./robustness/validate -v --count 1 --run TestDataReports
 
 # Test previous release branches
 


tests/robustness/makefile.mk
```diff
@@ -1,6 +1,6 @@
 .PHONY: test-robustness-reports
 test-robustness-reports:
-	cd ./tests && go test ./robustness/validate -v --run TestDataReports
+	cd ./tests && go test ./robustness/validate -v --count 1 --run TestDataReports
```
ahrtr, line 3: Is 1 the default value for `--count`?

siyuanfoundation (approved):

tests/robustness/makefile.mk
serathius: No, https://github.com/golang/go/issues/24573

ahrtr: To be clearer, the default value for `--count` is 1 for sure. But there's a difference between providing it with the default value, or not providing it at all.  Cache will be disabled in the first case.

ahrtr (approved):

PR #17759, opened 2024-04-09
Add explicit checks for assumptions in robustness test validation

serathius:

Started using https://github.com/etcd-io/etcd/pull/17733 and imminently make a mistake of copying one report over another. This results in strange errors and linearization visualization. It took me more time than I want to admit to identify the problem.

Adding validation of assumptions that should always be true if report comes directly from robustness test run. This should re-evaluation of reports more user friendly by providing clear errors.

### DIFF
--- a/tests/robustness/validate/validate.go
+++ b/tests/robustness/validate/validate.go
@@ -30,6 +30,10 @@ import (
 
 // ValidateAndReturnVisualize returns visualize as porcupine.linearizationInfo used to generate visualization is private.
 func ValidateAndReturnVisualize(t *testing.T, lg *zap.Logger, cfg Config, reports []report.ClientReport, timeout time.Duration) (visualize func(basepath string) error) {
+	err := checkValidationAssumptions(reports)
+	if err != nil {
+		t.Fatalf("Broken validation assumptions: %s", err)
+	}
 	patchedOperations := patchedOperationHistory(reports)
 	linearizable, visualize := validateLinearizableOperationsAndVisualize(lg, patchedOperations, timeout)
 	if linearizable != porcupine.Ok {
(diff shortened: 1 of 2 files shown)


tests/robustness/validate/validate.go
```diff
@@ -52,6 +56,155 @@ type Config struct {
...
+			return fmt.Errorf("didn't observe revision %d in watch, required to patch operation and validate serializable requests", i)
+		}
+	}
+	return nil
+}
+
+func nonConcurrentClientRequests(reports []report.ClientReport) error {
+	lastClientRequestReturn := map[int]int64{}
+	for _, r := range reports {
+		for _, op := range r.KeyValue {
```
siyuanfoundation, line 194: question: could 2 different ClientReport contain ops from the same client? 
If not, why is the map needed? In that case, you just need to set lastRequest=0 at the beginning of each report. 
If yes, the time ordering between different reports are not guaranteed.

serathius: They should not, but it's possible if the report got corrupted. The goal is to detect such things early as linearization will return broken hard to decode results. 

> If yes, the time ordering between different reports are not guaranteed. 

Right, but it's ok to assume so.

siyuanfoundation: Thanks for the clarification. Can you add some comments regarding this? It might not be clear to future contributors digging into the code why the map is needed.

tests/robustness/validate/validate.go
```diff
@@ -52,6 +56,155 @@ type Config struct {
...
+				lastOperation = op
+			}
+		}
+	}
+	response := lastOperation.Output.(model.MaybeEtcdResponse)
+	if response.PartialResponse || response.Error != "" {
+		return fmt.Errorf("last operation %v failed, required to validate watch", lastOperation)
+	}
+	request := lastOperation.Input.(model.EtcdRequest)
+	if request.Type != model.Txn || len(request.Txn.Conditions) != 0 || len(request.Txn.OperationsOnSuccess) != 1 || request.Txn.OperationsOnSuccess[0].Type != model.PutOperation || request.Txn.OperationsOnSuccess[0].Put.Key != "tombstone" {
```
fuweid, line 140: For easy review, maybe we can use the following pattern or use function to check if it's not tombstone request.

```go
if condition_1 ||
   condition_2 ||
   condition_3 {
}
```

fuweid (commented):
LGTM

tests/robustness/validate/validate.go
serathius: Done

tests/robustness/validate/validate.go
```diff
@@ -52,6 +56,155 @@ type Config struct {
...
+	var lastOperation porcupine.Operation
+
+	for _, r := range reports {
+		for _, op := range r.KeyValue {
+			if op.Call > lastOperation.Call {
+				lastOperation = op
+			}
+		}
+	}
+	response := lastOperation.Output.(model.MaybeEtcdResponse)
```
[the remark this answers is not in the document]
serathius: Done

tests/robustness/validate/validate.go
```diff
@@ -52,6 +56,155 @@ type Config struct {
...
+	if response.PartialResponse || response.Error != "" {
+		return fmt.Errorf("last operation %v failed, required to validate watch", lastOperation)
+	}
+	request := lastOperation.Input.(model.EtcdRequest)
+	if request.Type != model.Txn || len(request.Txn.Conditions) != 0 || len(request.Txn.OperationsOnSuccess) != 1 || request.Txn.OperationsOnSuccess[0].Type != model.PutOperation || request.Txn.OperationsOnSuccess[0].Put.Key != "tombstone" {
+		return fmt.Errorf("last operation %+v didn't write `end` key, required to validate watch", request)
```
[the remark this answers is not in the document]
serathius: Done

ahrtr (approved):

PR #17768, opened 2024-04-10
[Robustness] Collect failed read operations to calculate request success rate

serathius:
cc @ahrtr @jmhbnz @MadhavJivrajani @siyuanfoundation 

Hope this will help debugging of arm64 failures. I expect the reason behind low QPS on arm is request timeout, currently set to 40ms. It's possible a lot of read request timeout due to lower single core performance. This PR will calculate a request success rate so we can confirm that.

### DIFF
--- a/tests/robustness/model/history.go
+++ b/tests/robustness/model/history.go
@@ -59,51 +59,33 @@ func (h *AppendableHistory) AppendRange(startKey, endKey string, revision, limit
 	if resp != nil && resp.Header != nil {
 		respRevision = resp.Header.Revision
 	}
-	h.appendSuccessful(porcupine.Operation{
-		ClientId: h.streamID,
-		Input:    staleRangeRequest(startKey, endKey, limit, revision),
-		Call:     start.Nanoseconds(),
-		Output:   rangeResponse(resp.Kvs, resp.Count, respRevision),
-		Return:   end.Nanoseconds(),
-	})
+	h.appendSuccessful(staleRangeRequest(startKey, endKey, limit, revision), start, end, rangeResponse(resp.Kvs, resp.Count, respRevision))
 }
 
 func (h *AppendableHistory) AppendPut(key, value string, start, end time.Duration, resp *clientv3.PutResponse, err error) {
 	request := putRequest(key, value)
 	if err != nil {
-		h.appendFailed(request, start.Nanoseconds(), err)
+		h.appendFailed(request, start, err)
 		return
 	}
 	var revision int64
 	if resp != nil && resp.Header != nil {
 		revision = resp.Header.Revision
 	}
-	h.appendSuccessful(porcupine.Operation{
-		ClientId: h.streamID,
-		Input:    request,
-		Call:     start.Nanoseconds(),
-		Output:   putResponse(revision),
-		Return:   end.Nanoseconds(),
-	})
+	h.appendSuccessful(request, start, end, putResponse(revision))
 }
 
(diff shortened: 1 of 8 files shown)


tests/robustness/traffic/traffic.go
```diff
@@ -93,14 +93,17 @@ func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
...
-	var operationCount int
+	var totalOperations int
+	var successfulOperations int
 	for _, r := range reports {
-		operationCount += len(r.KeyValue)
+		totalOperations += len(r.KeyValue)
+		successfulOperations += r.SuccessfulOperations()
 	}
-	lg.Info("Recorded operations", zap.Int("operationCount", operationCount))
+	lg.Info("Recorded operations", zap.Int("operations", totalOperations), zap.Float64("successRate", float64(successfulOperations)/float64(totalOperations)))
```
siyuanfoundation, line 102: Do you expect all the errors to be timeout errors? If not, would it make sense to collect failed counts by error types?

tests/robustness/model/history.go
```diff
@@ -493,54 +445,57 @@ func defragmentResponse(revision int64) MaybeEtcdResponse {
...
-		successful: make([]porcupine.Operation, 0, len(h.successful)+len(h2.successful)),
-		failed:     make([]porcupine.Operation, 0, len(h.failed)+len(h2.failed)),
+		operations: make([]porcupine.Operation, 0, len(h.operations)+len(h2.operations)),
 	}
-	result.successful = append(result.successful, h.successful...)
-	result.successful = append(result.successful, h2.successful...)
-	result.failed = append(result.failed, h.failed...)
-	result.failed = append(result.failed, h2.failed...)
+	result.operations = append(result.operations, h.operations...)
+	result.operations = append(result.operations, h2.operations...)
```
siyuanfoundation, line 456: check the order of and result[last] and h2[0] first.

tests/robustness/traffic/traffic.go
serathius: No, categorizing errors would be next step, but I don't think we will need it to confirm that there are more errors on arm vs amd.

tests/robustness/model/history.go
serathius: Oh, nice catch, however this function is no longer used so let's just remove it.

tests/robustness/model/history.go
```diff
@@ -195,35 +158,30 @@ func (h *AppendableHistory) AppendTxn(cmp []clientv3.Cmp, clientOnSuccessOps, cl
...
-func (h *AppendableHistory) appendSuccessful(op porcupine.Operation) {
+	}
 	if op.Call >= op.Return {
 		panic(fmt.Sprintf("Invalid operation, call(%d) >= return(%d)", op.Call, op.Return))
 	}
-	if len(h.successful) > 0 {
-		prevSuccessful := h.successful[len(h.successful)-1]
-		if op.Call <= prevSuccessful.Call {
-			panic(fmt.Sprintf("Out of order append, new.call(%d) <= prev.call(%d)", op.Call, prevSuccessful.Call))
+	if len(h.operations) > 0 {
```
siyuanfoundation, line 175: this section can be extracted to share with appendFailed.

serathius: Ups, during refactors I botched my git history and dropped some commits. Let me restore it.

jmhbnz (approved):
LGTM - Thanks for help looking into `arm64` @serathius

siyuanfoundation (approved):

ahrtr (approved):

fuweid (approved):
LGTM

PR #17770, opened 2024-04-10
Test: add tools/rw-heatmaps & tools/testgrid-analysis to module_dirs

### LINKED ISSUE #17760: Update scripts/test_lib.sh to include ./tools modules

As identified in #17751, `scripts/test_lib.sh` currently doesn't include go modules nested in `./tools`.

We need to update `module_dirs()` to include them.
 https://github.com/etcd-io/etcd/blob/fbda59186605b1c1a8fc06c6e65a33dcb98e2fbd/scripts/test_lib.sh#L168-L170

### Why is this needed?

Not having these nested modules in `test_lib.sh` makes them dodge some of our CI/CD workflows.

tico88612:
Resolve #17760

### DIFF
--- a/scripts/test_lib.sh
+++ b/scripts/test_lib.sh
@@ -166,7 +166,7 @@ function run_for_module {
 }
 
 function module_dirs() {
-  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests ."
+  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/rw-heatmaps tools/testgrid-analysis ."
 }
 
 # maybe_run [cmd...] runs given command depending on the DRY_RUN flag.
--- a/scripts/test_lib.sh
+++ b/scripts/test_lib.sh
@@ -178,6 +178,9 @@ function maybe_run() {
   fi
 }
 
+# modules
+# returns the list of all modules in the project, not including the tools,
+# as they are not considered to be added to the bill for materials.
 function modules() {
   modules=(
     "${ROOT_MODULE}/api/v3"
--- a/tools/testgrid-analysis/cmd/data.go
+++ b/tools/testgrid-analysis/cmd/data.go
@@ -95,9 +95,9 @@ func processRow(dashboard, tab string, row *apipb.ListRowsResponse_Row, allTests
 		if maxDays > 0 && header.Started.AsTime().Before(earliestTimeToConsider) {
 			continue
 		}
-		total += 1
+		total++
 		if _, ok := failureTestStatusesInt[cell.Result]; ok {
-			failed += 1
+			failed++
 			// markdown table format of | commit | log |
 			logs = append(logs, fmt.Sprintf("| %s | %s | https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/%s/%s |", strings.Join(header.Extra, ","), header.Started.AsTime().String(), tab, header.Build))
 		}
--- a/tools/testgrid-analysis/cmd/github.go
+++ b/tools/testgrid-analysis/cmd/github.go
@@ -26,7 +26,7 @@ import (
 func createIssues(tests []*TestResultSummary, labels []string) {
 	openIssues := getOpenIssues(labels)
 	for _, t := range tests {
-		createIssueIfNonExist(tab, t, openIssues, append(labels, "help wanted"))
+		createIssueIfNonExist(t, openIssues, append(labels, "help wanted"))
 	}
 }
 
@@ -54,7 +54,7 @@ func getOpenIssues(labels []string) []*github.Issue {
 	return allIssues
 }
 
-func createIssueIfNonExist(tab string, t *TestResultSummary, issues []*github.Issue, labels []string) {
+func createIssueIfNonExist(t *TestResultSummary, issues []*github.Issue, labels []string) {
 	// check if there is already an open issue regarding this test
 	for _, issue := range issues {
 		if strings.Contains(*issue.Title, t.Name) {


tico88612:
Should I fix lint for this PR?

https://github.com/etcd-io/etcd/actions/runs/8635285626/job/23677581680?pr=17770#step:10:58

ivanvc:
@tico88612, I feel like this is a reasonable thing to do, as I verified that after addressing the three (minor) issues that the linter shows right now, the rest of the linters run without issues.

@jmhbnz, correct me if I'm wrong, but I don't think we would want to merge the PR if it would break the CI.

jmhbnz:
> I don't think we would want to merge the PR if it would break the CI.

Absolutely - we generally don't merge unless tests are passing, including static analysis.

scripts/test_lib.sh
```diff
@@ -166,7 +166,7 @@ function run_for_module {
 }
 
 function module_dirs() {
-  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests ."
+  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/rw-heatmaps tools/testgrid-analysis ."
```
jmhbnz, line 169: There is a function below called `modules` on line `181`.   Do we also need to update that @ahrtr?

ahrtr: I think yes.

tico88612:
What kind of problem is this? I need some ideas to resolve this.

https://github.com/etcd-io/etcd/actions/runs/8655015103/job/23733280841?pr=17770#step:10:24

ivanvc:
> What kind of problem is this? I need some ideas to resolve this.
> 
> https://github.com/etcd-io/etcd/actions/runs/8655015103/job/23733280841?pr=17770#step:10:24

Hey @tico88612, running `make fix-bom` (and committing `bill-of-materials.json) should fix it.

scripts/test_lib.sh
ivanvc: @ahrtr, @jmhbnz, adding these to `modules`, breaks generating the BOM. This diff partially fixes the issue:

```diff
diff --git a/dummy.go b/dummy.go
index 48db9af93..ee4ff91e6 100644
--- a/dummy.go
+++ b/dummy.go
@@ -18,8 +18,10 @@ package main_test
 // module (e.g. for sake of 'bom' generation).
 // Thanks to this 'go mod tidy' is not removing that dependencies from go.mod.
 import (
-       _ "go.etcd.io/etcd/client/v2"                // keep
-       _ "go.etcd.io/etcd/etcdctl/v3/ctlv3/command" // keep
-       _ "go.etcd.io/etcd/etcdutl/v3/etcdutl"       // keep
-       _ "go.etcd.io/etcd/tests/v3/integration"     // keep
+       _ "go.etcd.io/etcd/client/v2"                      // keep
+       _ "go.etcd.io/etcd/etcdctl/v3/ctlv3/command"       // keep
+       _ "go.etcd.io/etcd/etcdutl/v3/etcdutl"             // keep
+       _ "go.etcd.io/etcd/tests/v3/integration"           // keep
+       _ "go.etcd.io/etcd/tools/rw-heatmaps/v3/cmd"       // keep
+       _ "go.etcd.io/etcd/tools/testgrid-analysis/v3/cmd" // keep
 )
diff --git a/go.mod b/go.mod
index 9d9aae976..ca4d6ebd4 100644
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,8 @@ replace (
        go.etcd.io/etcd/pkg/v3 => ./pkg
        go.etcd.io/etcd/server/v3 => ./server
        go.etcd.io/etcd/tests/v3 => ./tests
+       go.etcd.io/etcd/tools/rw-heatmaps/v3 => ./tools/rw-heatmaps
+       go.etcd.io/etcd/tools/testgrid-analysis/v3 => ./tools/testgrid-analysis
 )
 
 require (
@@ -33,6 +35,8 @@ require (
        go.etcd.io/etcd/pkg/v3 v3.6.0-alpha.0
        go.etcd.io/etcd/server/v3 v3.6.0-alpha.0
        go.etcd.io/etcd/tests/v3 v3.6.0-alpha.0
+       go.etcd.io/etcd/tools/rw-heatmaps/v3 v3.6.0-alpha.0
+       go.etcd.io/etcd/tools/testgrid-analysis/v3 v3.6.0-alpha.0
        go.etcd.io/raft/v3 v3.6.0-alpha.0
        go.uber.org/zap v1.27.0
        golang.org/x/time v0.5.0
@@ -41,21 +45,31 @@ require (
 )
```

Adding them to `dummy.go` so that they are added to `go.mod`. However, after doing that, the error from building the BOM looks random (fails at `pkg/expect/expect.go` and `tests/robustness/report/client.go`). I'm not sure if you have any hints of what's wrong:

```
2024/04/12 20:24:53 could not list go.etcd.io/etcd/api/v3/... go.etcd.io/etcd/pkg/v3/... go.etcd.io/etcd/client/pkg/v3/... go.etcd.io/etcd/client/v2/... go.etcd.io/etcd/client/v3/... go.etcd.io/etcd/server/v3/... go.etcd.io/etcd/etcdutl/v3/... go.etcd.io/etcd/etcdctl/v3/... go.etcd.io/etcd/tests/v3/... go.etcd.io/etcd/tools/rw-heatmaps/v3/
... go.etcd.io/etcd/tools/testgrid-analysis/v3/... go.etcd.io/etcd/v3/... dependencies: 'go list go.etcd.io/etcd/api/v3/... go.etcd.io/etcd/pkg/v3/... go.etcd.io/etcd/client/pkg/v3/... go.etcd.io/etcd/client/v2/... go.etcd.io/etcd/client/v3/... go.etcd.io/etcd/server/v3/... go.etcd.io/etcd/etcdutl/v3/... go.etcd.io/etcd/etcdctl/v3/... go.e
tcd.io/etcd/tests/v3/... go.etcd.io/etcd/tools/rw-heatmaps/v3/... go.etcd.io/etcd/tools/testgrid-analysis/v3/... go.etcd.io/etcd/v3/...' failed with:
pkg/expect/expect.go:32:2: missing go.sum entry for module providing package github.com/creack/pty (imported by go.etcd.io/etcd/pkg/v3/expect); to add:
        go get go.etcd.io/etcd/pkg/v3/expect@v3.6.0-alpha.0
tests/robustness/report/client.go:27:2: missing go.sum entry for module providing package github.com/anishathalye/porcupine (imported by go.etcd.io/etcd/tests/v3/robustness/model); to add:
        go get go.etcd.io/etcd/tests/v3/robustness/model@v3.6.0-alpha.0
```

ivanvc:
> Running `make fix-bom` doesn't work unless I delete those two lines.

See https://github.com/etcd-io/etcd/pull/17770#discussion_r1563642039

scripts/test_lib.sh
tico88612: Have any updates for this?

ahrtr: Sorry, did not get time to dig into this. But one thing for sure is that we don't want the `tools/*` to break any production packages.

For simplicity & safety, propose to take care of (e.g. `go mod tidy; gofmt, etc`) all the packages under the tools/* separately, for example, we can extend the existing `scripts/fix.sh`.

ahrtr: @ivanvc  could you guide the contributor to finish this? cc @ArkaSaha30

ivanvc: Hi @ahrtr, I meant to get back to this but was busy with other tasks. I'll take a stab at this.

ivanvc: By the way, @ahrtr, do we want the BOM to be updated with the dependencies from the tools sub-projects? I think that's what's problematic right now.

ahrtr: > By the way, @ahrtr, do we want the BOM to be updated with the dependencies from the tools sub-projects?

Currently the BOM doesn't include any dependencies in tools. So I think we just need to keep it as it's.

ivanvc: @ahrtr, I've carefully reviewed the bash scripts inside `scripts/`. Although we could create some new functions to declare the tool directories/modules and break down when we should run something on them and when not, I don't think it makes sense.

If we keep the tool modules out of the `modules` function/variable, then it means that the only tasks that won't run on them are:
1. BOM
2. Tests coverage

The rest of the tasks use `module_dirs`, which the contributor initially did. My suggestion would be to add a comment to the `modules` function specifying that the tools shouldn't be in there.

Do we care about test coverage (2.) for the tools?

cc. @ArkaSaha30

ahrtr: Thanks @ivanvc for the summary. 

- For BOM (bill-of-materials), as mentioned above, current the [BOM](https://github.com/etcd-io/etcd/blob/main/bill-of-materials.json) doesn't include any dependencies used by [tools](https://github.com/etcd-io/etcd/tree/main/tools). I was thinking to keep it as it's, because the dependences in tools are only supposed to be used in **dev** environment. In other words, any dependencies used by tools do not affect the security or license concerns of the etcd running in production environment. I suggest to keep it as it's unless someone (e.g. from the [K8s SRC](https://github.com/kubernetes/committee-security-response)) clearly requests/challenges it. (The other option is to create a separate list of BOM for the tools.) Could you raise a discussion in the sig-etcd channel to get feedback and CC the K8s SRC members?
- For the test coverage, it isn't a big deal.


Note the purpose of this PR is to ensure the scripts/fix.sh can also automatically run `go mod tidy` and `gofmt` for all the packages under tools, so that we don't need to manually take care of them separately.

ivanvc: I'll start a thread in #sig-etcd. `scripts/fix.sh` will run on the tools subdirectories, as it uses `module_dirs`.

@tico88612, sorry for the back-and-forth. But we're now aligned on how to solve the issue. 
1. Can you please revert commit 8d38d1f855fd8f892cbe7896fb0823cd6fb21413
2. Can you apply the following patch after reverting the previously mentioned commit?
   ```diff
   diff --git a/scripts/test_lib.sh b/scripts/test_lib.sh
   index 550c718d3..be5194fb1 100644
   --- a/scripts/test_lib.sh
   +++ b/scripts/test_lib.sh
   @@ -178,6 +178,9 @@ function maybe_run() {
      fi
    }
 
   +# modules
   +# returns the list of all modules in the project, not including the tools,
   +# as they are not considered to be added to the bill for materials.
    function modules() {
      modules=(
        "${ROOT_MODULE}/api/v3"
   ```

Please let me know if you have any follow-up questions. If you don't have the capacity to work on this anymore, I can start a new PR and co-author the commits you made. Thank you so much for your contribution.

ivanvc:
@ahrtr PTAL :)

scripts/test_lib.sh
ahrtr, line 169:
```suggestion
  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/rw-heatmaps tools/testgrid-analysis tools/mod ."
```

tico88612: @ahrtr CI test was failed.

ahrtr: Sorry, please revert the change. Let's take care of it separately.

ahrtr (approved):
LGTM

Thank you!

ivanvc (approved):
Thanks @tico88612! Giving my thumbs up. I'm not sure if James will have the capacity to review this.

ivanvc:
@siyuanfoundation, just a heads up of commit https://github.com/etcd-io/etcd/pull/17770/commits/e3c1812302f63f7ad2dfc7f04735600087e6e689. I hope you used that argument initially and then dropped it, so while technically, it won't affect the current behavior, maybe you wanted to use the tab somewhere in that method.

siyuanfoundation:
> @siyuanfoundation, just a heads up of commit [e3c1812](https://github.com/etcd-io/etcd/commit/e3c1812302f63f7ad2dfc7f04735600087e6e689). I hope you used that argument initially and then dropped it, so while technically, it won't affect the current behavior, maybe you wanted to use the tab somewhere in that method.

It was used initially but no longer needed. Thanks for the heads up!

siyuanfoundation (approved):

PR #17771, opened 2024-04-10
Validate bookmarkable checks the last event before progress notify

serathius:

Noticed that watch validation allows just an empty streams. The current validation is:
* `validateReliable` checks `if there are events ordered in time as a < b < c, then if the watch receives events a and c, it is guaranteed to receive b`.
* `validateResumable` checks watch request that should start from specified revision, this allows us to check that first event in the watch stream. The `a` in `a < b < c`.

As watch is eventually consistent it means that watch streams can just finish early and be empty. To make all the pieces we need something to guarantee the `c` in `a < b < c`.

Here is where `validateBookmarkable` could help. It's the only way to be sure that we got events up to some revision. Bookmarkable should guarantee that all events before progress notify were delivered. As we have `validateReliable` we don't really need to check if all events were delivered, just if the last event before progress notification was delivered, our `c` in `a < b < c`.

Marking as draft as code is too complicated to my liking and we need test. As long as validation is using `t.Error` we cannot test if for failures. Before sending PR officially to review I work on refactoring validation to make it more testable.

### DIFF
--- a/tests/robustness/validate/watch.go
+++ b/tests/robustness/validate/watch.go
@@ -17,6 +17,7 @@ package validate
 import (
 	"errors"
 
+	"github.com/google/go-cmp/cmp"
 	"go.uber.org/zap"
 
 	"go.etcd.io/etcd/tests/v3/robustness/model"
@@ -55,11 +56,11 @@ func validateWatch(lg *zap.Logger, cfg Config, reports []report.ClientReport, ev
 		if err != nil {
 			return err
 		}
-		err = validateBookmarkable(lg, r)
-		if err != nil {
-			return err
-		}
 		if eventHistory != nil {
+			err = validateBookmarkable(lg, eventHistory, r)
+			if err != nil {
+				return err
+			}
 			err = validateReliable(lg, eventHistory, r)
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


siyuanfoundation:
This PR looks correct.
But I have a question about `validateReliable`.  For this block https://github.com/etcd-io/etcd/blob/65ac859a1b613a3b1b509cd80400f9fcbeae97d6/tests/robustness/validate/watch.go#L120-L127
Should the `events` be filtered to matching events first? because `events` might contain events for other keys. 
For example, 
```
// persisted events: a1, b1, a2, b2, b3, a3
// watch stream events: a1, a2, a3
// index = 0, events[0]=a1, event = a1, match, continue to index++ and next watch event
// index = 1, events[1]=b1, event = a2, request do not match, continue to index++ and next watch event
// index = 2, events[2]=a2 , event = a3, request match, but events[2]=a2 != a3, report failure?
```

serathius:
Want to add tests via https://github.com/etcd-io/etcd/pull/17786 first.

serathius:
Ready to review
ping @MadhavJivrajani @ahrtr @siyuanfoundation

siyuanfoundation (approved):

tests/robustness/validate/watch.go
```diff
@@ -76,17 +77,37 @@ func validateWatch(lg *zap.Logger, cfg Config, reports []report.ClientReport, ev
...
-func validateBookmarkable(lg *zap.Logger, report report.ClientReport) (err error) {
+func validateBookmarkable(lg *zap.Logger, eventHistory []model.PersistedEvent, report report.ClientReport) (err error) {
 	for _, op := range report.Watch {
 		var lastProgressNotifyRevision int64
+		var gotEventBeforeProgressNotify *model.PersistedEvent
 		for _, resp := range op.Responses {
 			for _, event := range resp.Events {
 				if event.Revision <= lastProgressNotifyRevision {
 					lg.Error("Broke watch guarantee", zap.String("guarantee", "bookmarkable"), zap.Int("client", report.ClientID), zap.Int64("revision", event.Revision))
 					err = errBrokeBookmarkable
```
ahrtr, line 88: Should we return the error here?

ahrtr (approved):

tests/robustness/validate/watch.go
serathius: No, we want to log all the errors.

PR #17781, opened 2024-04-12
Remove limit from read requests after a failed write

serathius:
Limit can cause multiple request due to pagination. For reads after a failed write we would like to return to normal write request as soon as possible.

### DIFF
--- a/tests/robustness/traffic/kubernetes.go
+++ b/tests/robustness/traffic/kubernetes.go
@@ -63,6 +63,7 @@ func (t kubernetesTraffic) Run(ctx context.Context, c *RecordingClient, limiter 
 	s := newStorage()
 	keyPrefix := "/registry/" + t.resource + "/"
 	g := errgroup.Group{}
+	readLimit := t.averageKeyCount
 
 	g.Go(func() error {
 		for {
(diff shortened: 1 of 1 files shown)


tests/robustness/traffic/kubernetes.go
```diff
@@ -118,7 +118,7 @@ func (t kubernetesTraffic) Read(ctx context.Context, kc *kubernetesClient, s *st
 
 	for hasMore {
 		readCtx, cancel := context.WithTimeout(ctx, RequestTimeout)
-		resp, err := kc.Range(readCtx, rangeStart, rangeEnd, revision, limit)
+		resp, err := kc.Range(readCtx, rangeStart, rangeEnd, revision, int64(limit))
```
siyuanfoundation, line 121: No limit could cause the Range request taking longer too to pack all the KVs. Wouldn't it be faster to pass in a flag like ignoreMorePages and break in the for loop?

serathius: > No limit could cause the Range request taking longer too to pack all the KVs

Don't understand the point. There are only ~10 keys at one time. There is no performance impact of listing all keys. I just want to avoid multiple round trips with pagination.

siyuanfoundation: The comment is just in case there are a large range of keys, Range without limit could still take a long time, and one range request with limit is faster. 
If there are only 10 keys, then that should be fine.

siyuanfoundation (approved):

ahrtr (approved):

MadhavJivrajani (approved):

fuweid (approved):
LGTM

PR #17786, opened 2024-04-12
Add tests to watch validation

serathius:


serathius:

ready to review. This is super important as I found multiple a bug and couple of gaps in validation. Didn't change the validation logic, just added test scenarios and left TODO in tests to fix validation.

tests/robustness/validate/validate_test.go
```diff
@@ -54,250 +55,1364 @@ func TestDataReports(t *testing.T) {
...
+										putWatchEvent("b", "2", 3, true),
+										putWatchEvent("a", "1", 2, true),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectError: "broke Ordered - events are ordered by revision; an event will never appear on a watch if it precedes an event in time that has already been posted",
```
MadhavJivrajani, line 124: I think we should create new error types in `validate.go` like
```go
var brokenOrderErr = errors.New("...")
```
and use those here. It might get a little cumbersome to maintain the same text in more than one place.

serathius: Done

ahrtr (approved):

Issue #17780, opened 2024-04-12
Revision decreasing after panic during compaction


serathius:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Failure in https://github.com/etcd-io/etcd/actions/runs/8659974818

![image](https://github.com/etcd-io/etcd/assets/5873433/8508b160-c2f7-49f1-85b8-9930a1132826)

### What did you expect to happen?

Revision doesn't decrease

### How can we reproduce it (as minimally and precisely as possible)?

Follow https://github.com/etcd-io/etcd/tree/main/tests/robustness#re-evaluate-existing-report to validate report from https://github.com/etcd-io/etcd/actions/runs/8659974818

Run `TestRobustnessExploratory/Kubernetes/LowTraffic/ClusterOfSize1` with failpoint `compactBeforeSetFinishedCompact=panic()`

### Anything else we need to know?

TODO:
* [x] Try to reproduce the issue
* [x] Try to reproduce the issue on amd64

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

serathius:
Hmm, still no repro. Are we so unlucky that we hit a bit flip or broken machine?

siyuanfoundation:
I was not able to repro it on my mac (arm64) with 500 runs and failpoint `compactBeforeSetFinishedCompact=panic()`

chaochn47:
Not be able to download the log

[...]

PR #17792, opened 2024-04-15
Update the compaction log when bootstrap and update compact's signature

ahrtr:
Two changes:
- Print the "`resume scheduled compaction`" only when the `compactLockfree` is successful.
- Actually the compact() never return an error, so remove the second return parameter.

Link to https://github.com/etcd-io/etcd/issues/17780#issuecomment-2056558562


ahrtr:
@fuweid @serathius PTAL

ahrtr:
This is a straightforward change, PTAL

serathius (approved):

fuweid (approved):
LGTM

Issue #17752, opened 2024-04-08
Add e2e tests to verify snapshot data compatibility in a mixed version cluster


siyuanfoundation:
### What would you like to be added?

e2e tests that verify in a mixed version cluster (3.5+3.4, and 3.6+3.5), when a snapshot is sent from the higher version leader to the lower version follower, the snapshot file is properly migrated and compatible to the lower version. 

To achieve this, the following tasks need to be done:

- [x] [3.5->3.4] Add `--experimental-snapshot-catchup-entries` flag in 3.5 to make sure snapshot is sent in e2e test. https://github.com/etcd-io/etcd/pull/17808
- [ ] [3.5->3.4] Add tests to verify schema of db file using bbolt after snapshot is sent and applied in the mixed version cluster.
- [ ] [3.6->3.5] Add tests to verify schema of db file using bbolt after snapshot is sent and applied in the mixed version cluster, provided the online migration is implemented.

### Why is this needed?

During implementing 3.5->3.4 downgrade, we [caught an issue](https://github.com/etcd-io/etcd/pull/17330#discussion_r1476069034) that could be easily slipped in downgrade designs: when a lower version follower receives a snapshot from a higher version leader, the snapshot might not be conformant to the lower version schema. In addition to offline data migration during downgrade process, special attention needs to paid to cover online migration as well. 

This is important to guarantee data integrity and compatibility in a mixed version cluster. It is important not just to add the online migration, also to add proper tests to make sure of that.

siyuanfoundation:
cc @serathius

PR #17808, opened 2024-04-15
[3.5] add experimental-snapshot-catchup-entries flag

siyuanfoundation:

https://github.com/etcd-io/etcd/issues/17752

This is import prerequisite to verify that in a mixed version cluster, the follower receives a snapshot that has the correct schema for that version. 

Did not add the flag to help file so that it is a hidden flag that only is used for testing.

### DIFF
--- a/server/etcdmain/config.go
+++ b/server/etcdmain/config.go
@@ -302,6 +302,7 @@ func newConfig() *config {
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "Number of entries for a slow follower to catch up after compacting the raft storage entries. (WARNING: only use this flag for tests.)")
 
 	// unsafe
 	fs.BoolVar(&cfg.ec.UnsafeNoFsync, "unsafe-no-fsync", false, "Disables fsync, unsafe, will cause data loss.")
--- a/server/etcdmain/config.go
+++ b/server/etcdmain/config.go
@@ -302,7 +302,7 @@ func newConfig() *config {
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
-	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "Number of entries for a slow follower to catch up after compacting the raft storage entries. (WARNING: only use this flag for tests.)")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "(WARNING: Use this flag with caution!) Number of entries for a slow follower to catch up after compacting the raft storage entries.")
 
 	// unsafe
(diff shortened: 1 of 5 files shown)


siyuanfoundation:
@serathius @ahrtr

tests/framework/e2e/cluster.go
```diff
@@ -402,6 +403,10 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 		panic(fmt.Sprintf("Unknown cluster version %v", cfg.Version))
 	}
 
+	if cfg.SnapshotCatchUpEntries > 0 && execPath == BinPath {
```
serathius, line 406: Why check exec path?

server/etcdmain/config.go
```diff
@@ -302,6 +302,7 @@ func newConfig() *config {
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "Number of entries for a slow follower to catch up after compacting the raft storage entries. (WARNING: only use this flag for tests.)")
```
serathius, line 305: Put the warning at the beginning, sad we cannot use https://github.com/spf13/pflag?tab=readme-ov-file#hidden-flags

tests/framework/e2e/cluster.go
siyuanfoundation: This is to make sure we do not set the flag for 3.4 binary

server/etcdmain/config.go
siyuanfoundation: done.

tests/framework/e2e/cluster.go
serathius: Should we rather disable test that cannot use the flag? Assume that existing code already does this, Just a note to look into that in the future.

serathius (approved):

server/etcdmain/config.go
```diff
@@ -302,6 +302,7 @@ func newConfig() *config {
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "(WARNING: only use this flag for tests!) Number of entries for a slow follower to catch up after compacting the raft storage entries.")
```
ahrtr, line 305: It's a little strange to add a user-facing flag, but not allow users to use it. (but we don't technically prevent users from using it)
```suggestion
	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "(WARNING: Use this flag with caution!) Number of entries for a slow follower to catch up after compacting the raft storage entries.")
```

tests/framework/e2e/cluster.go
siyuanfoundation: Whether a test can use it or not depend on which version is the leader, not inherently to the test itself.

ahrtr: > This is to make sure we do not set the flag for 3.4 binary

Please add a comment to explain this.

server/etcdmain/config.go
```diff
@@ -302,6 +302,7 @@ func newConfig() *config {
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
+	fs.Uint64Var(&cfg.ec.SnapshotCatchUpEntries, "experimental-snapshot-catchup-entries", cfg.ec.SnapshotCatchUpEntries, "(WARNING: Use this flag with caution!) Number of entries for a slow follower to catch up after compacting the raft storage entries.")
```
ahrtr, line 305: I think we should add similar "WARNING: Use this flag with caution!" for the main branch as well.

ahrtr:
Please signoff the commit

server/etcdmain/config.go
serathius: Can you provide more context why you think that? The original PR added the warning because we are backporting a feature just for testing, and we didn't want to make it visible to users.

serathius: This is only a test only backport, but the original flag was introduced in https://github.com/etcd-io/etcd/pull/15033 with broader set of motivations.

tests/framework/e2e/cluster.go
siyuanfoundation: done

ahrtr (approved):

serathius (approved):

PR #17820, opened 2024-04-18
server/embed: address golangci var-naming issues

ivanvc:
Addresses issues in ListenPeerUrls, ListenClientUrls, ListenClientHttpUrls, AdvertisePeerUrls, AdvertiseClientUrls.

Following the same approach as in #17674, a linter ignore is added to these exported fields. However, non-exported functions and exported struct fields used only for JSON marshaling can be safely renamed.

Part of #17578.


server/embed/config.go
```diff
@@ -449,11 +452,11 @@ type configYAML struct {
...
-	ListenPeerUrls       string `json:"listen-peer-urls"`
-	ListenClientUrls     string `json:"listen-client-urls"`
-	ListenClientHttpUrls string `json:"listen-client-http-urls"`
-	AdvertisePeerUrls    string `json:"initial-advertise-peer-urls"`
-	AdvertiseClientUrls  string `json:"advertise-client-urls"`
+	ListenPeerURLs       string `json:"listen-peer-urls"`
+	ListenClientURLs     string `json:"listen-client-urls"`
+	ListenClientHTTPURLs string `json:"listen-client-http-urls"`
+	AdvertisePeerURLs    string `json:"initial-advertise-peer-urls"`
+	AdvertiseClientURLs  string `json:"advertise-client-urls"`
```
ivanvc, line 459: As stated in the PR description, these are exported fields from an unexported struct. It's internally used to marshal/unmarshal JSONs. So, it should be fine to rename.

server/etcdmain/config_test.go
```diff
@@ -61,10 +61,10 @@ func TestConfigFileMemberFields(t *testing.T) {
...
 		SnapshotCount          uint64 `json:"snapshot-count"`
 		SnapshotCatchUpEntries uint64 `json:"experimental-snapshot-catch-up-entries"`
-		ListenPeerUrls         string `json:"listen-peer-urls"`
-		ListenClientUrls       string `json:"listen-client-urls"`
-		ListenClientHttpUrls   string `json:"listen-client-http-urls"`
-		AdvertiseClientUrls    string `json:"advertise-client-urls"`
+		ListenPeerURLs         string `json:"listen-peer-urls"`
+		ListenClientURLs       string `json:"listen-client-urls"`
+		ListenClientHTTPURLs   string `json:"listen-client-http-urls"`
+		AdvertiseClientURLs    string `json:"advertise-client-urls"`
```
ivanvc, line 67: As stated in the PR description, these are used only for tests marshaling the config into JSON, so it should be safe to rename them.

server/embed/config.go
serathius: Right

serathius (approved):

ahrtr (approved):

Issue #17775, opened 2024-04-11
Make robustness qps requirements less fragile to CI performance


serathius:
### What would you like to be added?

One of the guiding principles of robustness tests was strictness, if we are expecting something to happen, we should validate that it really happens. If we require high qps for producing a bug we should validate that we achieve high qps. Running on lowered powered machines like arm doesn't mean that we can just drop the requirement as it undermines the quality of robustness tests. We need to go back to the original motivation of high qps and look for other ways to improve reliability of the robustness tests.

The qps requirements for robustness were driven by https://github.com/etcd-io/etcd/issues/13766, a data inconsistency bug requiring [high qps to even have a chance of reproduction ](https://github.com/etcd-io/etcd/issues/13766#issuecomment-1078897588). Reaching such qps required pretty advanced tricks to linearize the requests in finite time, however it was finally was delivered by https://github.com/etcd-io/etcd/pull/14682#issuecomment-1364664026.

Maintaining such high qps throughput the whole test is pretty hard, that's why even if we need 1000 qps for https://github.com/etcd-io/etcd/issues/13766, we allow it to fluctuate between 200-1000 from test to test. This is because the main part of the test is failure injection. Failure that can cause the etcd to become unavailable. This implicitly creates a dependency on etcd ability to recover from failure. As showed in https://github.com/etcd-io/etcd/issues/17455 this can vary from version to version, from configuration to configuration. It etcd takes too long to recover, the average qps suffers and the test fail.

So, how can me make the test feasible on lower powered machines while we cannot reduce the qps requirements? Simle, by asking whether we really need qps to be high throughput the whole test, or maybe a part of it. If we think about reproducing https://github.com/etcd-io/etcd/issues/13766 we only should care about qps just before we sent SIGKILL, not as much when etcd is down, nor after it recovers. It would be good to check if there are some requests after etcd is killed, but no high qps is required then.

Proposal:
* Track start and end time, member affected and type of failpoint injected. Add this information to into client report. In the future we could even add it to visualization. 
* Calculate QPS only for the time and requests finished before the failpoint injection.
* Add validation for number of requests that happened during and after failpoint injection. Expect that minimal number needs to be adjusted based on cluster configuration (3 vs 1 node) as some failpoints can cause total full downtime for 1 node cluster. 

cc @MadhavJivrajani @siyuanfoundation @ahrtr @jmhbnz 

### Why is this needed?

[...]

PR #17825, opened 2024-04-19
Don't require minimal for failpoint injection period

serathius:
Fixes https://github.com/etcd-io/etcd/issues/17775

cc @ahrtr @siyuanfoundation @fuweid @MadhavJivrajani

### DIFF
--- a/tests/robustness/failpoint/failpoint.go
+++ b/tests/robustness/failpoint/failpoint.go
@@ -29,10 +29,7 @@ import (
 )
 
 const (
-	triggerTimeout               = time.Minute
-	waitBetweenFailpointTriggers = time.Second
-	failpointInjectionsCount     = 1
-	failpointInjectionsRetries   = 3
+	triggerTimeout = time.Minute
 )
 
 var (
@@ -78,45 +75,37 @@ func Validate(clus *e2e.EtcdProcessCluster, failpoint Failpoint) error {
 	return nil
 }
 
-func Inject(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, failpoint Failpoint) error {
+func Inject(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, failpoint Failpoint, baseTime time.Time) (*InjectionReport, error) {
 	ctx, cancel := context.WithTimeout(ctx, triggerTimeout)
 	defer cancel()
 	var err error
-	successes := 0
-	failures := 0
-	for successes < failpointInjectionsCount && failures < failpointInjectionsRetries {
-		time.Sleep(waitBetweenFailpointTriggers)
 
-		lg.Info("Verifying cluster health before failpoint", zap.String("failpoint", failpoint.Name()))
-		if err = verifyClusterHealth(ctx, t, clus); err != nil {
-			return fmt.Errorf("failed to verify cluster health before failpoint injection, err: %v", err)
-		}
-
-		lg.Info("Triggering failpoint", zap.String("failpoint", failpoint.Name()))
-		err = failpoint.Inject(ctx, t, lg, clus)
-		if err != nil {
-			select {
-			case <-ctx.Done():
-				return fmt.Errorf("Triggering failpoints timed out, err: %v", ctx.Err())
-			default:
-			}
-			lg.Info("Failed to trigger failpoint", zap.String("failpoint", failpoint.Name()), zap.Error(err))
-			failures++
-			continue
-		}
-
-		lg.Info("Verifying cluster health after failpoint", zap.String("failpoint", failpoint.Name()))
-		if err = verifyClusterHealth(ctx, t, clus); err != nil {
-			return fmt.Errorf("failed to verify cluster health after failpoint injection, err: %v", err)
-		}
-
-		successes++
+	if err = verifyClusterHealth(ctx, t, clus); err != nil {
+		return nil, fmt.Errorf("failed to verify cluster health before failpoint injection, err: %v", err)
 	}
-	if successes < failpointInjectionsCount || failures >= failpointInjectionsRetries {
-		t.Errorf("failed to trigger failpoints enough times, err: %v", err)
+	lg.Info("Triggering failpoint", zap.String("failpoint", failpoint.Name()))
+	start := time.Since(baseTime)
+	err = failpoint.Inject(ctx, t, lg, clus)
+	if err != nil {
+		lg.Error("Failed to trigger failpoint", zap.String("failpoint", failpoint.Name()), zap.Error(err))
+		return nil, fmt.Errorf("failed triggering failpoint, err: %v", err)
 	}
+	if err = verifyClusterHealth(ctx, t, clus); err != nil {
+		return nil, fmt.Errorf("failed to verify cluster health after failpoint injection, err: %v", err)
+	}
+	lg.Info("Finished triggering failpoint", zap.String("failpoint", failpoint.Name()))
+	end := time.Since(baseTime)
 
-	return nil
+	return &InjectionReport{
+		Start: start,
+		End:   end,
+		Name:  failpoint.Name(),
+	}, nil
+}
+
+type InjectionReport struct {
+	Start, End time.Duration
+	Name       string
 }
 
 func verifyClusterHealth(ctx context.Context, _ *testing.T, clus *e2e.EtcdProcessCluster) error {
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -110,27 +110,32 @@ func (s testScenario) run(ctx context.Context, t *testing.T, lg *zap.Logger, clu
 	defer cancel()
 	g := errgroup.Group{}
 	var operationReport, watchReport []report.ClientReport
-	finishTraffic := make(chan struct{})
+	failpointInjected := make(chan failpoint.InjectionReport, 1)
 
 	// using baseTime time-measuring operation to get monotonic clock reading
 	// see https://github.com/golang/go/blob/master/src/time/time.go#L17
 	baseTime := time.Now()
 	ids := identity.NewIDProvider()
 	g.Go(func() error {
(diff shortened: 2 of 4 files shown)


tests/robustness/failpoint/failpoint.go
```diff
@@ -29,10 +29,7 @@ import (
 )
 
 const (
-	triggerTimeout               = time.Minute
-	waitBetweenFailpointTriggers = time.Second
-	failpointInjectionsCount     = 1
-	failpointInjectionsRetries   = 3
```
siyuanfoundation, line 35: why is the retries removed? was it ever useful?

tests/robustness/main_test.go
```diff
@@ -110,27 +110,30 @@ func (s testScenario) run(ctx context.Context, t *testing.T, lg *zap.Logger, clu
...
 
 	// using baseTime time-measuring operation to get monotonic clock reading
 	// see https://github.com/golang/go/blob/master/src/time/time.go#L17
 	baseTime := time.Now()
 	ids := identity.NewIDProvider()
 	g.Go(func() error {
-		defer close(finishTraffic)
-		err := failpoint.Inject(ctx, t, lg, clus, s.failpoint)
+		defer close(failpointInjected)
+		time.Sleep(time.Second)
```
MadhavJivrajani, line 121: Why are we sleeping here?

tests/robustness/main_test.go
```diff
@@ -110,27 +110,30 @@ func (s testScenario) run(ctx context.Context, t *testing.T, lg *zap.Logger, clu
...
 	g := errgroup.Group{}
 	var operationReport, watchReport []report.ClientReport
-	finishTraffic := make(chan struct{})
+	failpointInjected := make(chan failpoint.InjectionReport, 1)
 
 	// using baseTime time-measuring operation to get monotonic clock reading
 	// see https://github.com/golang/go/blob/master/src/time/time.go#L17
 	baseTime := time.Now()
 	ids := identity.NewIDProvider()
 	g.Go(func() error {
```
MadhavJivrajani, line 119: Also, not for this PR, but we should probably get rid of the errgroup.Group, probably a good idea to just use a wait groups and goroutines.

tests/robustness/traffic/traffic.go
```diff
@@ -98,23 +114,57 @@ func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
...
-	period := endTime.Sub(startTime)
-	qps := float64(successfulOperations) / period.Seconds()
-	lg.Info("Traffic from successful requests", zap.Float64("qps", qps), zap.Int("operations", successfulOperations), zap.Duration("period", period))
-	if qps < profile.MinimalQPS {
-		t.Errorf("Requiring minimal %f qps for test results to be reliable, got %f qps", profile.MinimalQPS, qps)
+	if beforeFailpointStats.QPS() < profile.MinimalQPS {
+		t.Errorf("Requiring minimal %f qps before failpoint injection for test results to be reliable, got %f qps", profile.MinimalQPS, beforeFailpointStats.QPS())
+	}
+	if afterFailpointStats.QPS() < profile.MinimalQPS {
+		t.Errorf("Requiring minimal %f qps after failpoint injection for test results to be reliable, got %f qps", profile.MinimalQPS, afterFailpointStats.QPS())
```
MadhavJivrajani, line 131: Thinking more here, do we want the tests to run if we deem it to be non-reliable? Is there value in failing fast here?

tests/robustness/main_test.go
fuweid: I think it's to postpone the failpoint if simulator hasn't started yet.

tests/robustness/failpoint/failpoint.go
```diff
@@ -78,45 +75,37 @@ func Validate(clus *e2e.EtcdProcessCluster, failpoint Failpoint) error {
...
-			return fmt.Errorf("failed to verify cluster health after failpoint injection, err: %v", err)
-		}
-
-		successes++
+	if err = verifyClusterHealth(ctx, t, clus); err != nil {
+		return nil, fmt.Errorf("failed to verify cluster health before failpoint injection, err: %v", err)
+	}
+	lg.Info("Triggering failpoint", zap.String("failpoint", failpoint.Name()))
+	start := time.Since(baseTime)
+	err = failpoint.Inject(ctx, t, lg, clus)
```
fuweid, line 88: I think we should keep the original retry here just in case that the failpoint http call timeouts

tests/robustness/failpoint/failpoint.go
serathius: It was introduced as a band aid solution to flakes. It reduced problem a little, but there were still some issues with failpoint that retries didn't help. Over time we improved, especially after it we discovered issue with process execution. 

Hard to say if removing it will expose some issues, we should see them in CI.

tests/robustness/main_test.go
serathius: Yes, it's a wait before and after failpoint injection to ensure traffic has time to wind up.

serathius: Added comments.

tests/robustness/main_test.go
serathius: Why? This code was changed to use errgroup to make error handling cleaner.

tests/robustness/traffic/traffic.go
serathius: Usually errors are found when cluster transitions from unhealthy to healthy status. You want to ensure there is proper coverage of qps when failpoint is injected (See https://github.com/etcd-io/etcd/issues/17775) and after cluster recovers. For example when member is killed, you want to see how cluster reacts to member rejoining it. 

This validation is to ensure that the test is strictly validating it's assumptions, we want to make sure that period after failpoint injection is also covered by requests. If we didn't overtime we might introduce regression. For example we could remove health check after failpoint, which could allow clusters to be left broken and never recover after failpoint injection. We need to make sure that all failpoints we introduce restore cluster to health to make sure we not only find issues during the failpoint being run (for example network split), but also consequences of it (when separated member rejoins cluster).

serathius: I want to make the validation even more stricter here, to ensure not only total qps, but guarantee per member qps. This way we detect cases where member didn't recover from failpoint, but cluster is serving data due to quorum availability.

serathius:
Hmm, looks like validation of traffic after failpoint injection might not work as well:
```
2024-04-19T20:41:33.6949208Z     logger.go:146: 2024-04-19T20:41:33.230Z	INFO	Reporting complete traffic	{"successes": 323, "failures": 465, "successRate": 0.4098984771573604, "period": "2.804586175s", "qps": 115.16850609876518}
2024-04-19T20:41:33.6951530Z     logger.go:146: 2024-04-19T20:41:33.230Z	INFO	Reporting traffic before failure injection	{"successes": 210, "failures": 115, "successRate": 0.6461538461538462, "period": "977.189618ms", "qps": 214.9019966358259}
2024-04-19T20:41:33.6953790Z     logger.go:146: 2024-04-19T20:41:33.230Z	INFO	Reporting traffic during failure injection	{"successes": 8, "failures": 108, "successRate": 0.06896551724137931, "period": "386.950683ms", "qps": 20.674469257882148}
2024-04-19T20:41:33.6955721Z     logger.go:146: 2024-04-19T20:41:33.230Z	INFO	Reporting traffic after failure injection	{"successes": 105, "failures": 242, "successRate": 0.3025936599423631, "period": "1.440445874s", "qps": 72.89409612346185}
2024-04-19T20:41:33.6957188Z     traffic.go:131: Requiring minimal 100.000000 qps after failpoint injection for test results to be reliable, got 72.894096 qps
```

serathius:
Or maybe it just show that qps was head heavy. I mean it was mainly driven by requests before failpoint injection. 
Logs show that only 30% of requests succeed after failpoint injection. That might imply that latency increases after failpoint injection. But also the success rate before failpoint was also pretty low 65%. Which implies that request timeout is too low.

tests/robustness/traffic/traffic.go
serathius: Heh, needed to drop the validation for now. I was wrong to assume this validation will pass. Leaving a TODO for the future.

serathius:
> I think we might need to increase request timeout enough so the success rate before failpoint 100%.

Increasing the request timeout 5 times to 200ms doesn't help. At that point increasing it gives reverse effect of reducing qps. For now will drop QPS validation post failpoint injection.

ahrtr:
defer to @MadhavJivrajani @siyuanfoundation @ArkaSaha30 @fuweid to review robustness test PRs.

tests/robustness/failpoint/failpoint.go
fuweid: Thanks for the comment. It sounds good to me.

fuweid:
> For now will drop QPS validation post failpoint injection.

Not sure that I understand it correct. 
For the QPS profile, it seems to make sure that the failpoint is triggered under that the certain request pressure.
If so, the change looks good~.

serathius:
> > For now will drop QPS validation post failpoint injection.
> 
> Not sure that I understand it correct. For the QPS profile, it seems to make sure that the failpoint is triggered under that the certain request pressure. If so, the change looks good~.

Yes, I tried to also have a QPS validation after failpoint is injected, but it causes too much flakiness.

tests/robustness/main_test.go
MadhavJivrajani: Right, but iiuc errgroups are useful only if you return an error, we always return nil afaict.

serathius: Ok, makes sense.

fuweid (approved):
LGTM

siyuanfoundation (approved):

tests/robustness/failpoint/failpoint.go
serathius: Checked logs form one of robustness test runs. Didn't find any "Failed to trigger failpoint" logs.

PR #17833, opened 2024-04-21
Implement compaction support in robustness test

serathius:


serathius:
Yey, revision returned by compaction is not linearizable. 

Screenshot from 2024-05-21 21-00-38

https://github.com/etcd-io/etcd/actions/runs/9179653925?pr=17833

cc @MadhavJivrajani @fuweid @siyuanfoundation

siyuanfoundation:
> https://github.com/etcd-io/etcd/actions/runs/9179653925?pr=17833

I don't understand why the line is drawn at the end of the compact op. 
There is a valid path from `Compact(314), rev: 318` -> `Compact(297), rev: 319` -> `put("key0", "308"), rev: 319` -> ...

serathius:
Assuming that compaction returns linearizable revision (that might not be necessarily a property we want to provide), operations `Compact(314), rev: 318` and `Compact(297), rev: 319` should not coexist. Operations execute order should follow returned revision, and it should not be possible to get both success in sequence `Compact(314)` and `Compact(297)`, because it doesn't make sense to compact revision `297` after `314`.

As mentioned above, it's not necessarily a bug, but at least a undocumented behavior. It stems from fact that etcd puts revision in the same field of response, independent of its meaning. Depending on the request the revision can be interpreted in a different way. Some examples:
* For PUT/TXN operations, response revision is the revision of the operation. Linearizable
* For Read request with revision=0, the response revision is the read. Linearizable
* For Read request with revision!=0, the response revision is the latest revision in cluster. Linearizable
* For Watch request, for response with events it's the revision of local member, for bookmarks it's revision of the watch stream.Non-linearizable.
* For other request I expect two cases:
  * Request goes through raft, revision is cluster revision. Linearizable. I expected Compaction is here, but it might not be.
  * Request doesn't require raft, revision is local. Non-linearizable.

serathius:
Here is even clearer example of this
image

serathius:
Ready for review, finally.
cc @siyuanfoundation @fuweid @MadhavJivrajani

tests/robustness/failpoint/trigger.go
```diff
@@ -78,11 +78,21 @@ func (t triggerCompact) Trigger(ctx context.Context, _ *testing.T, member e2e.Et
...
-		return nil, err
+		return nil, fmt.Errorf("failed to compact: %w", err)
 	}
 	return []report.ClientReport{cc.Report()}, nil
 }
 
-func (t triggerCompact) Available(e2e.EtcdProcessClusterConfig, e2e.EtcdProcess) bool {
+func (t triggerCompact) Available(config e2e.EtcdProcessClusterConfig, _ e2e.EtcdProcess) bool {
+	// Since introduction of compaction into traffic, injecting compaction failpoints started interfeering with peer proxy.
+	// TODO: Re-enable the peer proxy for compact failpoints when we confirm the root cause.
```
siyuanfoundation, line 88: Is there an issue tracking this?

serathius: Not yet, I think https://github.com/etcd-io/etcd/pull/16788 might be related. 

I was also hoping that maybe this will be auto resolved with https://github.com/etcd-io/etcd/pull/17938 as major rewrite might change.

siyuanfoundation (approved):

serathius:
ping @fuweid

fuweid (approved):

PR #17821, opened 2024-04-18
[3.4] Allow updateClusterVersion when downgrading from 3.5.

siyuanfoundation:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

https://github.com/etcd-io/etcd/issues/17818

Tested with 1 server cluster downgrade

```bash
bin/etcd
curl http://127.0.0.1:2379/version
>> {"etcdserver":"3.5.13","etcdcluster":"3.5.0"}

# stop 3.5 server and restart with 3.4 binary
../etcd-3.4/bin/etcd --next-cluster-version-compatible
>> ...
>> mvcc: downgrade meta bucket: remove keys [confState, term]
>> ...
>> etcdserver/membership: updated the cluster version from 3.5 to 3.4

curl http://127.0.0.1:2379/version
>> {"etcdserver":"3.4.31","etcdcluster":"3.4.0"}
```

PR #17847, opened 2024-04-22
refactor IsValidVersionChange.

siyuanfoundation:

refactor the function based on the feedbacks of https://github.com/etcd-io/etcd/pull/17821

### DIFF
--- a/server/etcdserver/api/capability.go
+++ b/server/etcdserver/api/capability.go
@@ -64,7 +64,7 @@ func UpdateCapability(lg *zap.Logger, v *semver.Version) {
 		return
 	}
 	enableMapMu.Lock()
-	if curVersion != nil && !serverversion.IsValidVersionChange(v, curVersion) {
+	if curVersion != nil && !serverversion.IsValidClusterVersionChange(v, curVersion) {
 		enableMapMu.Unlock()
 		return
 	}
--- a/server/etcdserver/version/downgrade.go
+++ b/server/etcdserver/version/downgrade.go
@@ -59,17 +59,17 @@ func allowedDowngradeVersion(ver *semver.Version) *semver.Version {
 	return &semver.Version{Major: ver.Major, Minor: ver.Minor - 1}
 }
 
-// IsValidVersionChange checks the two scenario when version is valid to change:
+// IsValidClusterVersionChange checks the two scenario when version is valid to change:
 // 1. Downgrade: cluster version is 1 minor version higher than local version,
 // cluster version should change.
 // 2. Cluster start: when not all members version are available, cluster version
 // is set to MinVersion(3.0), when all members are at higher version, cluster version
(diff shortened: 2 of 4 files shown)


siyuanfoundation:
From this refactor, it seems obvious the version order of `IsValidClusterVersionChange` in `UpdateCapability` is reversed. 
Should I fix it? @ahrtr @serathius

ahrtr:
> From this refactor, it seems obvious the version order of `IsValidClusterVersionChange` in `UpdateCapability` is reversed.
> Should I fix it?

Do you mean the change like below? Strange we did not see any issue in the test so far. It seems that we need to evaluate the downgrade's code. Please raise a separate issue to discuss this so as not to block your current PR.

```
$ git diff
diff --git a/server/etcdserver/api/capability.go b/server/etcdserver/api/capability.go
index 4cd9aca84..12dead393 100644
--- a/server/etcdserver/api/capability.go
+++ b/server/etcdserver/api/capability.go
@@ -64,7 +64,7 @@ func UpdateCapability(lg *zap.Logger, v *semver.Version) {
                return
        }
        enableMapMu.Lock()
-       if curVersion != nil && !serverversion.IsValidVersionChange(v, curVersion) {
+       if curVersion != nil && !serverversion.IsValidVersionChange(curVersion, v) {
                enableMapMu.Unlock()
                return
        }
```

siyuanfoundation:
> > From this refactor, it seems obvious the version order of `IsValidClusterVersionChange` in `UpdateCapability` is reversed.
> > Should I fix it?
> 
> Do you mean the change like below? Strange we did not see any issue in the test so far. It seems that we need to evaluate the downgrade's code. Please raise a separate issue to discuss this so as not to block your current PR.
> 
> ```
> $ git diff
> diff --git a/server/etcdserver/api/capability.go b/server/etcdserver/api/capability.go
> index 4cd9aca84..12dead393 100644
> --- a/server/etcdserver/api/capability.go
> +++ b/server/etcdserver/api/capability.go
> @@ -64,7 +64,7 @@ func UpdateCapability(lg *zap.Logger, v *semver.Version) {
>                 return
>         }
>         enableMapMu.Lock()
> -       if curVersion != nil && !serverversion.IsValidVersionChange(v, curVersion) {
> +       if curVersion != nil && !serverversion.IsValidVersionChange(curVersion, v) {
>                 enableMapMu.Unlock()
>                 return
>         }
> ```

Yes, that's the change I meant. 
The reason we did not see any issue is that the capability does not change over the 3.x versions https://github.com/etcd-io/etcd/blob/dd4e35a58597e66b74a84a5a99ad42a7b7723415/server/etcdserver/api/capability.go#L36-L43. 
It would be a no-op change. But I will raise a separate PR for it.

ahrtr (approved):
LGTM

Thanks

serathius (approved):

PR #17858, opened 2024-04-23
tests/robustness: cleanup errgroup usage

MadhavJivrajani:
We don't really need to use errgroups. In most places we return nil from the errgroup goroutines, those can be replaced by goroutines + waitgroups.

There was one place where we returned ctx.Err() from the errgroup routine but never caught that error from Wait(). It is okay to not return an error here as well since we catch the context cancellation error higher up in the call stack anyway.

cc @serathius

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -16,12 +16,12 @@ package robustness
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
-	"golang.org/x/sync/errgroup"
 
 	"go.etcd.io/etcd/tests/v3/framework"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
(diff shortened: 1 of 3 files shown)


tests/robustness/traffic/kubernetes.go
```diff
@@ -62,16 +61,18 @@ func (t kubernetesTraffic) Run(ctx context.Context, c *RecordingClient, limiter
...
 
-	g.Go(func() error {
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
 		for {
 			select {
 			case <-ctx.Done():
-				return ctx.Err()
+				return
```
MadhavJivrajani, line 73: Not returning an error here since we check for it anyway higher up in the call stack: 
https://github.com/etcd-io/etcd/blob/94b155a1c922c301119eb1b3d3e950355a5225d7/tests/robustness/traffic/traffic.go#L101

Also, this way it is more consistent with how the etcdTraffic implementation handles this.

tests/robustness/main_test.go
```diff
@@ -108,15 +108,17 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, s testSce
 func (s testScenario) run(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster) (reports []report.ClientReport) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
-	g := errgroup.Group{}
```
serathius, line 111: Maybe we don't use errors, but usage of errgroup is more concise.

PR #17860, opened 2024-04-23
fix version order when calling IsValidClusterVersionChange in UpdateCapability

siyuanfoundation:
`v` is the version the code want to change the `curVersion` so it should be the `verTo` when calling `IsValidClusterVersionChange`.

This should be a no-op change because Capability does not really change in 3.x. 


siyuanfoundation:
retest

siyuanfoundation:
retest

serathius (approved):

ahrtr (approved):
LGTM

Good catch.

PR #17815, opened 2024-04-17
[RFC] fix revision loss issue caused by compaction - 17780

fuweid:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #17862, opened 2024-04-24
server/storage/mvcc: should update currentRev in revMu

fuweid:
Follow-up: #17815


ahrtr (approved):
LGTM

Good catch, we should update the `currentRev` under the protection of `s.revMu`.

serathius (approved):

PR #17865, opened 2024-04-24
[3.5] fix revision loss issue caused by compaction - 17780

fuweid:
Backport:

* https://github.com/etcd-io/etcd/pull/17815
* https://github.com/etcd-io/etcd/pull/17862

### DIFF
--- a/tests/e2e/reproduce_17780_test.go
+++ b/tests/e2e/reproduce_17780_test.go
@@ -1,0 +1,115 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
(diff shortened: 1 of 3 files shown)


tests/e2e/reproduce_17780_test.go
```diff
@@ -0,0 +1,115 @@
...
+	require.NoError(t, clus.Procs[targetIdx].Failpoints().SetupHTTP(ctx, "compactBeforeSetFinishedCompact", `panic`))
+
+	_, err := cli.Compact(ctx, 11, clientv3.WithCompactPhysical())
+	require.Error(t, err)
+
+	require.Error(t, clus.Procs[targetIdx].Stop())
+	// NOTE: The proc panics and exit code is 2. It's impossible to restart
+	// that etcd proc because last exit code is 2 and Restart() refuses to
+	// start new one. Using IsRunning() function is to cleanup status.
+	require.False(t, clus.Procs[targetIdx].IsRunning())
```
fuweid, line 86: Side note: 
```
// The proc panics and exit code is 2. It's impossible to restart
// that etcd proc because last exit code is 2 and Restart() refuses to
// start new one. Using IsRunning() function is to cleanup status.
```

serathius: I think we might be missing some improvements to `expect` on release-3.4 branch. As a testing library change, we could consider backporting them avoid need for branch specific test code code causing branches to diverge.

serathius (approved):

tests/e2e/reproduce_17780_test.go
ahrtr: +1 to make the test as consistent as possible. @fuweid can you please raise a followup to track it? thanks

fuweid: Filed https://github.com/etcd-io/etcd/issues/17869

Issue #17857, opened 2024-04-23
Setup a consistent way to manage go version


ahrtr:
### What would you like to be added?

Currently there are multiple places setting the go versions, for examples,
- go.mod files, including both `go` and `toolchain` directives;
- .go-version file;
- environment variable, such as https://github.com/etcd-io/etcd/pull/17734

We should set up a consistent way to manage go version. There should be only one source of truth for the go&toolchain version.

One problem is that if a dependency sets a higher go or toolchain version, then when running scripts/fix.sh will automatically update the go & toolchain version in our go.mod files.  Refer to https://github.com/etcd-io/etcd/pull/17841/files#r1575344566

References:
- https://go.dev/doc/toolchain
- https://github.com/kubernetes/community/blob/master/contributors/devel/development.md#building-kubernetes-using-a-specific-version-of-go

cc @fuweid  @henrybear327  @ivanvc @jmhbnz  @MadhavJivrajani @serathius 

### Why is this needed?

Setup consistent way to manage go version

ahrtr:
> In Kubernetes we added a pre-submit check to catch such accidental bumps if they occur: [kubernetes/kube-openapi#467](https://github.com/kubernetes/kube-openapi/pull/467)

Thanks @MadhavJivrajani 

I think we should do similar check for all etcd branches: including `release-3.4`, `release-3.4` and `main`.

The other potential enhancement is to automatically populate the go versions for the `go` and `toolchain` lines in all go.mod (and go.sum) files based on the `.go-version` file.

MadhavJivrajani:
I have some ideas on how we can go about doing this. Will create a PR and we can discuss there!

ahrtr:
Followups:
- Update document: https://github.com/etcd-io/etcd/pull/17876#discussion_r1587360706
- Refactor the ./scripts/fix.sh as mentioned in https://github.com/etcd-io/etcd/pull/17876#discussion_r1593660044

ahrtr:
I think we also need to backport the changes to both 3.5 and 3.4

ahrtr:
We should follow the same rule to manage go version for all repositories under etcd-io. Note we only require users to depend on a minor version supported by golang team, i.e. **1.21**, but we don't force users to depend on a patch version, i.e. **1.21.10**.

jmhbnz:
[...]

PR #17876, opened 2024-04-25
[RFC] Setup a way to consistently manage go versions across scripts and go.mods

MadhavJivrajani:
This PR does the following:

### scripts: default to using .go-version's version for tests and builds

Additionally, provide ability to opt-out of the .go-version and use a
custom one via env vars: FORCE_HOST_GO and GO_VERSION.

### .*: sync go toolchain version and add ability to verify versions

This commit adds a script to sync the version present in .go-version
across all go.mod files as the toolchain directive. As part of that,
this commit also modifies go.mod files that did not have synced toolchain
directives.

Additionally, this also adds a script to verify all toolchain and go
directives against the version present in .go-version as follows:
(1) The go directive <= version in .go-version
(2) The toolchain directive == version in .go-version

The make target for this script is run as part of `make verify` which will automatically be run as a presubmit.

/assign @ahrtr @jmhbnz @serathius

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ fuzz:
 verify: verify-gofmt verify-bom verify-lint verify-dep verify-shellcheck verify-goword \
 	verify-govet verify-license-header verify-receiver-name verify-mod-tidy \
 	verify-shellws verify-proto-annotations verify-genproto verify-yamllint \
-	verify-govet-shadow verify-markdown-marker
+	verify-govet-shadow verify-markdown-marker verify-go-versions
 
 .PHONY: fix
 fix: fix-bom fix-lint fix-yamllint
@@ -187,3 +187,11 @@ clean:
 	rm -rf ./tests/e2e/default.proxy
 	rm -rf ./bin/shellcheck*
 	find ./ -name "127.0.0.1:*" -o -name "localhost:*" -o -name "*.log" -o -name "agent-*" -o -name "*.coverprofile" -o -name "testname-proxy-*" -delete
+
+.PHONY: verify-go-versions
+verify-go-versions:
+	./scripts/verify_go_versions.sh
+
+.PHONY: sync-toolchain-directive
+sync-toolchain-directive:
+	./scripts/sync_go_toolchain_directive.sh
--- a/scripts/fix.sh
+++ b/scripts/fix.sh
@@ -36,5 +36,7 @@ run_for_modules run ${GO_CMD} fmt || exit 2
 run_for_module tests bom_fix || exit 2
 bash_ws_fix || exit 2
 
+log_callout "Syncing go toolchain directives"
+run ./scripts/sync_go_toolchain_directive.sh || exit 2
 
 log_success -e "\\nSUCCESS: etcd code is fixed :)"
--- a/scripts/sync_go_toolchain_directive.sh
+++ b/scripts/sync_go_toolchain_directive.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+# This script looks at the version present in the .go-version file and treats
+# that to be the value of the toolchain directive that go should use. It then
+# updates the toolchain directives of all go.mod files to reflect this version.
+#
+# We do this to ensure that .go-version acts as the source of truth for go versions.
+
+set -euo pipefail
+
+source ./scripts/test_lib.sh
+
+TARGET_GO_VERSION="${TARGET_GO_VERSION:-"$(cat "${ETCD_ROOT_DIR}/.go-version")"}"
+find . -name 'go.mod' -exec go mod edit -toolchain=go"${TARGET_GO_VERSION}" {} \;
--- a/api/go.mod
+++ b/api/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/api/v3
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	github.com/coreos/go-semver v0.3.1
 	github.com/gogo/protobuf v1.3.2
--- a/client/internal/v2/go.mod
+++ b/client/internal/v2/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/client/v2
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	go.etcd.io/etcd/api/v3 v3.6.0-alpha.0
 	go.etcd.io/etcd/client/pkg/v3 v3.6.0-alpha.0
--- a/client/pkg/go.mod
+++ b/client/pkg/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/client/pkg/v3
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	github.com/coreos/go-systemd/v22 v22.5.0
 	github.com/stretchr/testify v1.9.0
--- a/client/v3/go.mod
+++ b/client/v3/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/client/v3
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	github.com/coreos/go-semver v0.3.1
 	github.com/dustin/go-humanize v1.0.1
--- a/etcdctl/go.mod
+++ b/etcdctl/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/etcdctl/v3
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	github.com/bgentry/speakeasy v0.1.0
 	github.com/cheggaaa/pb/v3 v3.1.5
--- a/etcdutl/go.mod
+++ b/etcdutl/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/etcdutl/v3
 
 go 1.22
 
+toolchain go1.22.2
+
 replace (
 	go.etcd.io/etcd/api/v3 => ../api
 	go.etcd.io/etcd/client/pkg/v3 => ../client/pkg
--- a/pkg/go.mod
+++ b/pkg/go.mod
@@ -2,6 +2,8 @@ module go.etcd.io/etcd/pkg/v3
 
 go 1.22
 
+toolchain go1.22.2
+
 require (
 	github.com/creack/pty v1.1.18
 	github.com/dustin/go-humanize v1.0.1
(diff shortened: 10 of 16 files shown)


scripts/sync_go_toolchain_directive.sh
```diff
@@ -0,0 +1,14 @@
...
+# updates the toolchain directives of all go.mod files to reflect this version.
+#
+# We do this to ensure that .go-version acts as the source of truth for go versions.
+
+set -euo pipefail
+
+source ./scripts/test_lib.sh
+
+TARGET_GO_VERSION="${TARGET_GO_VERSION:-"$(cat "${ETCD_ROOT_DIR}/.go-version")"}"
+find . -name 'go.mod' -exec go mod edit -toolchain=go${TARGET_GO_VERSION} {} \;
```
ahrtr, line 14: There are three options,
- only sync go line, e.g. `go 1.22.2`;
- only sync toolchain line, e.g. `toolchain go1.22.2`;
- sync both go and toolchain lines.

No matter which way we select, it will always use the highest go version by default.

For simplicity, can we go with option 1?

MadhavJivrajani: @ahrtr, if I understand correctly, the issue I see with option 1 and 3 is that whenever we allow bumping the `go` directive, we are also forcing developers of etcd and other projects that import etcd modules to bump their versions. From https://go.dev/doc/toolchain:
> the go line sets the minimum required Go version necessary to use a module or workspace.

scripts/verify_toolchain_directive.sh
```diff
@@ -0,0 +1,28 @@
...
+log_info
+
+out_of_sync="false"
+for mod in $(find . -name 'go.mod'); do
+    toolchain_version="$(go mod edit -json ${mod} | jq -r .Toolchain)"
+    if [[ "go${target_go_version}" != "${toolchain_version}" ]]; then
+        log_error "go toolchain directive out of sync for ${mod}, got: ${toolchain_version}"
+        out_of_sync="true"
+    fi
+done
```
ahrtr, line 22: The `go` lines in all etcd's `go.mod` files are also automatically bumped if any etcd's dependency bumped its go version. Should we also verify that the version in `go lines` <= `.go-version`?

scripts/sync_go_toolchain_directive.sh
ahrtr: Yes, it's true. It's OK to go for option 2. But please also see my next related comment https://github.com/etcd-io/etcd/pull/17876#discussion_r1582930923

ahrtr:
Please also resolve the workflow failure: `# shellcheck disable=SC2086`

We also need to clearly document the behaviour.

scripts/verify_toolchain_directive.sh
MadhavJivrajani: @ahrtr that's a good point, however I'm not sure what the course of action is once the dependency bump PRs are merged. If a `go line` bump occurred and the PR is merged, do we just update the `.go-version` file then? 

My main concern with allowing dependency bumps to also bump the go line is that it can cause an issue with projects like Kubernetes that import `go.etcd.io...` modules. If the go lines of these projects are behind those of the etcd modules, then they will be either forced to bump or they cannot update the dependency till they catch up (https://github.com/kubernetes/kubernetes/issues/123744)

This behaviour is mitigated in go1.23, but till that is released and we can adopt that, it is risky allowing the go line to be bumped at all, especially from a perspective of Kubernetes consuming etcd. For bumps related to security fixes or bugs that we see in etcd, we should probably discuss what the course of action should be, but we should probably we a little more aware for usual maintenance bumps.

ahrtr: >  they cannot update the dependency till they catch up 

I think we should follow this. 

For main branch, it isn't a problem because usually we keep using the latest go version (e.g. 1.22.2 for now).

For stable releases (3.5 and 3.4 for now), usually we keep using the latest patch of previous minor version (e.g. 1.21.9 for now). Usually, we don't proactively bump dependencies for etcd stable releases unless there are any CVEs or bugs that affect etcd. If bumping a dependency for a stable release leads to the go version being automatically bumped, 
- check whether there is an alternative. For example, if both 4.20.6 and 4.21.0 of a dependency resolve a CVE, but 4.20.6 requires go1.21.9 and 4.21.0 requires go 1.22.2, then we bump 4.20.6 for etcd stable releases (3.4/3.5).
- if there is no alternative
  - if we have to bump the dependency, then we bump go version for etcd stable releases (3.4/3.5) as well.
  - if we don't have to bump the dependency (_it's unlikely running into this case_), we can hold off for now until we bump go version.


We need to clearly document the behaviour in https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#golang-versions

MadhavJivrajani: @ahrtr that sounds good, I've added the extra check.

MadhavJivrajani: Also, ack on documenting this.

MadhavJivrajani:
@ahrtr on thinking about this a little more:
Whenever a dep bump changes the toolchain version, the version will be changed to whatever is the version of the local toolchain, from https://go.dev/doc/toolchain:
> For [repeatability](https://research.swtch.com/vgo-principles#repeatability), any command that updates the go line also updates the toolchain line to record its own toolchain name. 

This means that automated PRs, if it runs `./scripts/test_lib.sh`, will never accidentally change the toolchain line because we keep that in sync by setting the `GOTOOLCHAIN` env var as the version in .go-version. However, we still need to check it in case of manual bumps.

I've also changed the workflow to a presubmit instead of a periodic because now either case should not be allowed.

Makefile
```diff
@@ -196,3 +196,11 @@ clean:
 	rm -rf ./tests/e2e/default.proxy
 	rm -rf ./bin/shellcheck*
 	find ./ -name "127.0.0.1:*" -o -name "localhost:*" -o -name "*.log" -o -name "agent-*" -o -name "*.coverprofile" -o -name "testname-proxy-*" -delete
+
+.PHONY: verify-go-versions
+verify-go-versions:
```
ahrtr, line 201: Please get `verify-go-versions` included in the target `verify`, and you don't add a separate workflow check because there is an existing workflow check `static-analysis.yaml`
https://github.com/etcd-io/etcd/blob/3d4d4a0053915f701e7d1b75834b309364d65109/Makefile#L55

Makefile
```diff
@@ -196,3 +196,11 @@ clean:
...
+	./scripts/verify_go_versions.sh
+
+.PHONY: sync-toolchain-directive
+sync-toolchain-directive:
+	./scripts/sync_go_toolchain_directive.sh
```
ahrtr, line 206: Please also run this script in `./script/fix.sh`, so that contributors don't need to run it separately,.

Makefile
MadhavJivrajani: Done

Makefile
MadhavJivrajani: Done

MadhavJivrajani:
@ahrtr can you PTAL? Thank you!

Makefile
ahrtr: Probably I did not comment it clearly, sorry for the confusion. I wasn't saying to run `./scripts/fix.sh` in the target `sync-toolchain-directive`; instead, I was saying to run `./scripts/sync_go_toolchain_directive.sh` inside `./scripts/fix.sh`.

The `./scripts/fix.sh` is the single point of entry to fix everything for etcd.

MadhavJivrajani: Ah I see, I misunderstood you there, sorry. I've made the change @ahrtr, thank you,

ahrtr:
Overall looks good. Could you rebase this PR although github doesn't show any conflict? I see that it's still based on a commit of 2 weeks ago.

ahrtr (approved):
Great work!

Thanks @MadhavJivrajani 

Please also consider to update the doc as mentioned in https://github.com/etcd-io/etcd/pull/17876#discussion_r1587360706. It can be resolved in a separate PR.

ahrtr:
cc @fuweid @ivanvc  @jmhbnz @serathius

scripts/fix.sh
```diff
@@ -36,5 +36,7 @@ run_for_modules run ${GO_CMD} fmt || exit 2
 run_for_module tests bom_fix || exit 2
 bash_ws_fix || exit 2
 
+log_callout "Syncing go toolchain directives"
+run ./scripts/sync_go_toolchain_directive.sh || exit 2
```
serathius, line 40: Would prefer to move subroutines from `./scripts/fix.sh` to makefile targets. As you already have a makefile target `sync-toolchain-directive` why not add it as subtarget to `fix`?

serathius (approved):

scripts/fix.sh
ahrtr: > why not add it as subtarget to `fix`?

Yes, it's another option. Either way works for me.

We can even breakdown the existing ./script/fix.sh into a couple of Makefile targets
- fix-go-dependency (in which we will run `mod_tidy_fix` and the `sync_go_toolchain_directive.sh`)
- fix-gofmt
- fix-script-whitespace

It can be discussed & resolved in a separate PR.

fuweid:
Hi @ahrtr @serathius @MadhavJivrajani , do we consider to apply this to library project, like bbolt and raft? since we have discussion about this in bbolt project two days ago. 

REF: https://github.com/etcd-io/bbolt/pull/747

ahrtr:
> do we consider to apply this to library project, like bbolt and raft?

I think YES.  Please see https://github.com/etcd-io/etcd/issues/17857#issuecomment-2100965096, and please feel free to add comment in the issue.

MadhavJivrajani:
Coming back to the discussions here this week 👍🏼

ivanvc:
This implementation is great and simplifies what would be tedious if done by hand. We recently had the Go version updated by a new contributor, and I believe it wasn't very clear to him how to update the toolchain in the different `go.mod` files (refer to: https://github.com/etcd-io/etcd/pull/17975#issuecomment-2103331884). I think we would benefit from documenting this somewhere in the repository.

A potential improvement regarding documentation would be to document how to update the Go version. In the main branch with this new script, I think it's as easy as updating the value in `.go-version` and running `make fix` after :tada:

jmhbnz:
> A potential improvement regarding documentation would be to document how to update the Go version. In the main branch with this new script, I think it's as easy as updating the value in `.go-version` and running `make fix` after 🎉

Should we update that into this docs page https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#golang-versions?

Issue #17872, opened 2024-04-24
Add better verbosity to the etcdctl endpoint status output that includes percentage of DB used (to illustrate Defrag rate)


Scotchman0:
### What would you like to be added?

I have opened a PR with the specific changes: https://github.com/etcd-io/etcd/pull/17871 

### Why is this needed?

Adding the output from `OMC` which grants visibility of the size of the DB, how much is in use, and then providing the difference as a percentage for quick visual on storage impact on the cluster. A minor change but one that makes it much easier to illustrate where storage is going.

proposed change:
`$ etcdctl endpoint status`

~~~
+-----------------------------+------------------+---------+----------------+----------+-----------+------------+-----------+------------+--------------------+--------+
|          ENDPOINT           |        ID        | VERSION | DB SIZE/IN USE | NOT USED | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS |
+-----------------------------+------------------+---------+----------------+----------+-----------+------------+-----------+------------+--------------------+--------+
| https://xxx.xxx.xxx:2379  | 1fc63d8d0eb5fa7b | 3.5.9   | 6.7 GB/5.6 GB  | 18%      | false     | false      |      5853 | 4073918326 |         4073918326 |        |
| https://xxx.xxx.xxx:2379 | 3bd2d35c028a6a2f | 3.5.9   | 6.8 GB/5.6 GB  | 19%      | true      | false      |      5853 | 4073918327 |         4073918327 |        |
| https://xxx.xxx.xxx:2379 | 41cbbc084d60007d | 3.5.9   | 6.8 GB/5.6 GB  | 18%      | false     | false      |      5853 | 4073918330 |         4073918330 |        |
+-----------------------------+------------------+---------+----------------+----------+-----------+------------+-----------+------------+--------------------+--------+
~~~

This output is available in OMC (used for analysis of existing etcd databases running on openshift presently) and visual representation is not available to customers as a default, when it could be.

ivanvc:
@Scotchman0, it may be helpful if you can add the current output to compare to your suggestion.

/assign @Scotchman0

ivanvc:
I like the idea, but after seeing the implementation, I wonder if it would be better to add more columns rather than reusing/multi-purposing the current column. I defer to @ahrtr / @serathius  with their point of view.

Scotchman0:
[...]

PR #17877, opened 2024-04-25
add quota to endpoint status response

tjungblu:
This adds the configured backend quota bytes to the endpoint status response. related discussion in #17821


tjungblu:
seems the robustness is failing on cert issues from https://github.com/alexellis/arkade/issues/1056

alexellis:
If the test is failing, could it be due to trying to download vmmeter or something else via arkade?

The `arkade-get` action can be used instead of curl | sh..

ahrtr:
Overall looks good to me. The change is clear and straightforward!

Thanks @tjungblu

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/17872

ahrtr (approved):

alexellis:
From (uses curl/sh and the utility domain)

`      - uses: alexellis/setup-arkade@master` 

=>

To (Downloads directly from GitHub's releases page)

```yaml
      - uses: alexellis/arkade-get@master
        with:
           vmmeter: latest
```

fuweid (approved):
LGTM

ahrtr:
We also need to change the client side, including `etcdctl`. It can be in a separate PR.

tjungblu:
I think Scott had this in #17871 

Shall we also backport this?

ahrtr:
> Shall we also backport this?

Usually we don't backport new feature, but I am not against it if other members/contributors agree to backport it. 

We need to ensure it doesn't break any existing client, for example an application which depends on the old etcd client SDK can still work with a new etcdserver.

tjungblu:
it is not critical for us to have in 3.5, but it's also not a big and invasive change. 

>  for example an application which depends on the old etcd client SDK can still work with a new etcdserver.

Wouldn't protobuf just flag this as an unknown field?

ahrtr:
> Wouldn't protobuf just flag this as an unknown field?

For this PR, it's safe. My point is that we don't have such mix-version test cases,
- new client SDK version vs old etcdserver
- old client SDK version vs new etcdserver

It can be discussed separately.

ahrtr:
cc @fuweid @jmhbnz @serathius

ahrtr:
@tjungblu Please update 3.6 changelog for now. thx. If we decide to backport the PR to 3.5 later, we can update the changelog for 3.5 (revert the changelog for 3.6) by then.

Issue #17887, opened 2024-04-26
grpc-proxy has resolver-ttl broken since 3.4.28


RodionGork:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

When using several instances of `etcd grpc-proxy start ...` - if one of them goes down, this is supposed to be reflected in `etcdctl member list` after interval specified by `--resolver-ttl` elapses. This was broken with changes introduced in `v3.4.28` (but worked before quite well).

Please kindly see reproducing steps below, with the attached docker-compose file it should be easy.

**Cause**

My colleague Peter Zhu identified the problem:

Here in `cluster.go` we monitor additions and deletions of members, particularly here is deletion  
https://github.com/etcd-io/etcd/blob/v3.4.28/proxy/grpcproxy/cluster.go#L110  
this function have slightly changed since v3.4.27, but general sense is the same

Real culprit seemingly is where the event is issued:  
https://github.com/etcd-io/etcd/blob/v3.4.28/clientv3/naming/endpoints/endpoints_impl.go#L155

you see, `iup.Addr` is not set here, though used few lines below, unless we are mistaken.

### What did you expect to happen?

after TTL is elapsed, all other proxies should show updated list, without the node which was shut down.

### How can we reproduce it (as minimally and precisely as possible)?

Consider the attached file which contains simple docker-based setup to reproduce the issue:

- unpack zip file - so here are Dockerfile and compose.yaml
- run `docker-compose up` - it will launch `etcd-main` container and two proxies `etcd-proxy-a` and `etcd-proxy-b`

Now enter one of the proxies and list advertised members:

```sh
docker exec -it etcd-etcd-proxy-a-1 /bin/bash
root@6b1e5e88e4b9:/# /etcdctl member list

0, started, 72c909135c09, , http://etcd-proxy-b:2379, false
0, started, 6b1e5e88e4b9, , http://etcd-proxy-a:2379, false
```

[...]

PR #15907, opened 2023-05-16
[release-3.5]grpcproxy: fix memberlist results not update when proxy node down

yellowzf:
Backport of https://github.com/etcd-io/etcd/pull/15835
cc @lavacat @ahrtr

PR #17896, opened 2024-04-29
[3.4] fix grpc proxy member list not updated with node deletion

ivanvc:
If grpc starts with `--resolver-prefix`, its memberlist will return all alive proxy nodes. When one GRPC proxy node goes down, it shouldn't return that node. However, it's returned.

Backport of PR #15907 / commit ecfed91e5050431385961573e5c09658a66a59d5.

Part to #17887.


ahrtr (approved):
LGTM

Thanks @ivanvc

Issue #17657, opened 2024-03-27
Support Feature Gates


ahrtr:
### What would you like to be added?

Currently we add any new feature as an experimental feature with a configuration flag prefixed with "`experimental`", e.g. `--experimental-feature-name`. When we decide to graduate the experimental feature, we remove the prefix `experimental`.  Refer to [contributor-guide/features.md](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/features.md). Obviously it's a breaking change when we rename any configuration flag, so it isn't good.

So propose to add feature gates in the similar way as Kubernetes does. We can turn on or off any feature using the `--feature-gates` flag.

Reference:
- https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

I expect any new etcd features should follow the feature gates process once the feature gates feature is ready,
- added as an alpha feature
- disabled by default in the beginning

cc @fuweid @jmhbnz @serathius @siyuanfoundation @neolit123 @logicalhan @ArkaSaha30

### Why is this needed?

Support a better & non-breaking way to manage new features

neolit123:
> Obviously it's a breaking change when we rename any configuration flag, so it isn't good.

i'm a proponent to not prefix experimental features (flags, k8s labels, etc) with "experimental" because it's a breaking change, yes.
a flag can be just --foo instead of --experimental-foo and when used it can throw a warning and also be documented in its description as "experimental". if the user does not read the description, i would say it's their fault.

in terms of using feature gates in etcd, i don't see an issue. it's a smoother op-in for users when a beta FG enables a feature by default and has the opt-out.

logicalhan:
+100 from me as well! Really happy to see this initiative being driven forward!

logicalhan:
+100 from me as well! Really happy to see this initiative being driven forward!

(GitHub is doing weird things in their mobile UI)

tjungblu:
Sounds good, what's the plan for the current experimental features?

ahrtr:
> Expect we alias them to prevent breaking changes at least for 1 release, have both `--experimental-feature` and `--feature` flip the same state 

Agreed with this.

If we decide to graduate or deprecate an experimental feature, we should keep the experimental flag `--experimental-feature` (and co-exist with the flag `--feature` for the graduation case) for at least one major/minor release , and print warning if users still use it.

> some integration to work the same way after introduction of --feature-gates.

No strong opinion on this. 

[...]

PR #17914, opened 2024-04-30
[3.5] gRPC health server sets serving status to NOT_SERVING on defrag

tjungblu:
gRPC health server sets serving status to NOT_SERVING on defrag.

Backported from 3.6 in #16278

---

would be awesome to have this available in 3.5 already, we ran into this recently with a performance test when running etcd with 38 gigs.

### DIFF
--- a/server/etcdmain/config.go
+++ b/server/etcdmain/config.go
@@ -301,6 +301,7 @@ func newConfig() *config {
 	fs.DurationVar(&cfg.ec.ExperimentalWarningApplyDuration, "experimental-warning-apply-duration", cfg.ec.ExperimentalWarningApplyDuration, "Time duration after which a warning is generated if request takes more time.")
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
+	fs.BoolVar(&cfg.ec.ExperimentalStopGRPCServiceOnDefrag, "experimental-stop-grpc-service-on-defrag", cfg.ec.ExperimentalStopGRPCServiceOnDefrag, "Enable etcd gRPC service to stop serving client requests on defragmentation.")
 	fs.UintVar(&cfg.ec.ExperimentalBootstrapDefragThresholdMegabytes, "experimental-bootstrap-defrag-threshold-megabytes", 0, "Enable the defrag during etcd server bootstrap on condition that it will free at least the provided threshold of disk space. Needs to be set to non-zero value to take effect.")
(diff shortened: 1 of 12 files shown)


tjungblu:
the grpcProxy failure seems to be #16475 which wasn't backported to 3.5

chaochn47 (approved):
Thanks @tjungblu, the backport looks clean.

chaochn47:
Click approve too soon: there is a follow up PR https://github.com/etcd-io/etcd/pull/16836/ which improves code readability. Could you please also take that one? 

There is another in flight PR fixing the insecure / secure RPC servers set up https://github.com/etcd-io/etcd/pull/16959. 

Although this is not a blocker of this PR since this is protected by an experimental feature flag and still improves the availability, we should document that in the help messages clearly.

tjungblu:
Thanks @chaochn47, I'll pick the other PR too and let you know

tjungblu:
@chaochn47 please take a look again with your refactoring applied

otherwise @ahrtr and @serathius PTAL

server/etcdmain/config.go
```diff
@@ -301,6 +301,7 @@ func newConfig() *config {
 	fs.DurationVar(&cfg.ec.ExperimentalWarningApplyDuration, "experimental-warning-apply-duration", cfg.ec.ExperimentalWarningApplyDuration, "Time duration after which a warning is generated if request takes more time.")
 	fs.BoolVar(&cfg.ec.ExperimentalMemoryMlock, "experimental-memory-mlock", cfg.ec.ExperimentalMemoryMlock, "Enable to enforce etcd pages (in particular bbolt) to stay in RAM.")
 	fs.BoolVar(&cfg.ec.ExperimentalTxnModeWriteWithSharedBuffer, "experimental-txn-mode-write-with-shared-buffer", true, "Enable the write transaction to use a shared buffer in its readonly check operations.")
+	fs.BoolVar(&cfg.ec.ExperimentalStopGRPCServiceOnDefrag, "experimental-stop-grpc-service-on-defrag", cfg.ec.ExperimentalStopGRPCServiceOnDefrag, "Enable etcd gRPC service to stop serving client requests on defragmentation.")
```
ahrtr, line 304: Adding `experimental` prefix and removing it in future is a breaking change as we discussed in https://github.com/etcd-io/etcd/issues/17657. Also see https://github.com/etcd-io/etcd/issues/17657

Are we still happy to add prefix `experimental` for an flag?

I know this PR just backports changes from main to 3.5, so the suggestion (adding `experimental` in description only) is actually for the main branch (update main firstly). Do we want to follow the suggestion for now? Or do we have an agreement on https://github.com/etcd-io/etcd/issues/17657 and https://github.com/etcd-io/etcd/issues/17657 before we add any new flags? @fuweid @jmhbnz @serathius @siyuanfoundation @spzala 

```suggestion
	fs.BoolVar(&cfg.ec.ExperimentalStopGRPCServiceOnDefrag, "stop-grpc-service-on-defrag", cfg.ec.ExperimentalStopGRPCServiceOnDefrag, "(Experimental) Enable etcd gRPC service to stop serving client requests on defragmentation.")
```

serathius: As part of https://github.com/etcd-io/etcd/issues/17657 we will still need to support migration from the old flag naming scheme (with prefix) to the new one. One more flag should not change much as we already have other flags to migrate. So I don't think we need to block on this.

jmhbnz: I'm happy for this to proceed using older approach. We are still working on the KEP for feature flags so it will be a while off and agree with @serathius above we don't want to block this work.

ahrtr (approved):

chaochn47:
> @chaochn47 please take a look again with your refactoring applied

LGTM, thanks @tjungblu !!

serathius (approved):

PR #17921, opened 2024-05-01
Enable var-naming linter rule

### LINKED ISSUE #17578: Address var-name linter warnings

ivanvc:
These are the last two remaining tasks to enable the var-naming golangci lint rule.

* Enable the `var-naming` linter rule in `tools/.golangci.yaml`
* Address a warning from `tools/testgrid-analysis`

Fixes #17578.

### DIFF
--- a/tools/.golangci.yaml
+++ b/tools/.golangci.yaml
@@ -87,9 +87,12 @@ linters-settings: # please keep this alphabetized
       - name: error-strings
         severity: error
         disabled: false
-      # TODO: enable the following rules
       - name: var-naming
-        disabled: true
+        disabled: false
+        arguments:
+          - []
+          - ["GRPC", "WAL"]
+      # TODO: enable the following rules
       - name: exported
         disabled: true
       - name: unexported-return
--- a/tools/testgrid-analysis/cmd/data.go
+++ b/tools/testgrid-analysis/cmd/data.go
@@ -110,9 +110,9 @@ func processRow(dashboard, tab string, row *apipb.ListRowsResponse_Row, allTests
 	t.FailureLogs = logs
 	t.FailureRate = float32(failed) / float32(total)
 	if t.FailedRuns > 0 {
-		dashboardUrl := fmt.Sprintf("[%s](https://testgrid.k8s.io/%s#%s)", tab, dashboard, tab)
+		dashboardURL := fmt.Sprintf("[%s](https://testgrid.k8s.io/%s#%s)", tab, dashboard, tab)
 		t.IssueBody = fmt.Sprintf("## %s Test: %s \nTest failed %.1f%% (%d/%d) of the time\n\nfailure logs are:\n| commit | started | log |\n| --- | --- | --- |\n%s\n",
-			dashboardUrl, t.FullName, t.FailureRate*100, t.FailedRuns, t.TotalRuns, strings.Join(t.FailureLogs, "\n"))
+			dashboardURL, t.FullName, t.FailureRate*100, t.FailedRuns, t.TotalRuns, strings.Join(t.FailureLogs, "\n"))
 		t.IssueBody += "\nPlease follow the [instructions in the contributing guide](https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#check-for-flaky-tests) to reproduce the issue.\n"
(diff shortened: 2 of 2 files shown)


tools/.golangci.yaml
```diff
@@ -87,9 +87,12 @@ linters-settings: # please keep this alphabetized
       - name: error-strings
         severity: error
         disabled: false
-      # TODO: enable the following rules
       - name: var-naming
-        disabled: true
+        disabled: false
+        arguments:
+          - []
+          - ["GRPC", "WAL"]
```
ahrtr, line 94: Question: why add the two items in the block list? Please add a comment to clarify it.

tools/.golangci.yaml
```diff
@@ -87,9 +87,12 @@ linters-settings: # please keep this alphabetized
       - name: error-strings
         severity: error
         disabled: false
-      # TODO: enable the following rules
       - name: var-naming
-        disabled: true
+        disabled: false
+        arguments:
+          - [] # AllowList: leave it empty to go with defaults, not relaxing the rule.
+          - ["GRPC", "WAL"] # DenyList: Add GRPC and WAL to strict the rule not allowing instances like Wal or Grpc.
```
ahrtr, line 94: We already added some skipping directive in the repo for some specific cases. Can we just remove both list so that we just use the default settings?

ivanvc: > We already added some skipping directive in the repo for some specific cases.

Is there another place in which we have skipped directives?

> Can we just remove both list so that we just use the default settings?

As I was working on enabling the rule I found inconsistent use of WAL/Wal and GRPC/Grpc. If we decide to remove it from the deny list, the linter won't complain about the mixed usage of these keywords.

ivanvc: > > We already added some skipping directive in the repo for some specific cases.
> 
> Is there another place in which we have skipped directives?

Ah, got it. Yes, we have them inline in the code. However, new instances (i.e., new variables, new struct fields), won't be evaluated against this rule, so it will allow new instances of Wal and Grpc.

ahrtr: I am curious what's the default settings/values for the allowlist and blocklist? If we leave both slices as empty, does it have the same effect as disabling the linter?

ivanvc: The AllowList is empty as far as I know. The DenyList has the following items: https://github.com/mgechev/revive/blob/v1.3.7/lint/utils.go#L93-L133

I had to dive into the code to get these values, as they're not publicly documented (it's an unexported variable). As you can see, it doesn't include GRCP, and (of course) WAL.

ahrtr: Thanks for the clarification. Can you please add a comment to link to it so that all contributors are well aware of the. default values for blockList?

ivanvc: Done. PTAL :)

ahrtr (approved):
Perfect!

Thanks.

jmhbnz (approved):
LGTM - Thanks @ivanvc

Issue #17773, opened 2024-04-11
Measure testgrid flakinesss didn't detect flakes that happend


serathius:
### What would you like to be added?

https://github.com/etcd-io/etcd/actions/runs/8584759086/job/23525383783
![image](https://github.com/etcd-io/etcd/assets/5873433/fa08e2f8-f182-4c7f-ba43-78b8be02a3fa)

![image](https://github.com/etcd-io/etcd/assets/5873433/103fed09-0985-4602-bcbd-9b8526f971d7)
cc @siyuanfoundation 

### Why is this needed?

Last run https://github.com/etcd-io/etcd/actions/runs/8584759086/job/23525383783 on April 7th, didn't detect a flake on April 4th.

siyuanfoundation:
The flaky detection is meant to detect tests that fails sometimes, not one-off failures. 
this test fails about 2% of the time. @serathius  Do you think 2% threshold is reasonable?

serathius:
Hmm, not sure. The 16 % flakiness on main branch seems not great https://github.com/etcd-io/etcd/actions/runs/8584643093/job/23525115058.

siyuanfoundation:
I think the 16 % flakiness on main branch includes all the workflows on a PR. I am seeing a lot of flakiness wrt arm64.

jmhbnz:
> I think the 16 % flakiness on main branch includes all the workflows on a PR. I am seeing a lot of flakiness wrt arm64.

Raised a flake issue for `TestMemberAdd` e2e on `arm64` and `amd64`. I have seen it fail a few times in GitHub actions for `arm64` and there are also instances in TestGrid for `amd64` on prow.

https://github.com/etcd-io/etcd/issues/17778

serathius:
> The flaky detection is meant to detect tests that fails sometimes, not one-off failures.
this test fails about 2% of the time. @serathius Do you think 2% threshold is reasonable?

Maybe we could improve on visibility. What was surprising for me was fact that the tool didn't mention any flakes. Could we maybe log flakes below 2%, with note that it's too low to file an issue?

serathius:
The reports are very nice.
![image](https://github.com/etcd-io/etcd/assets/5873433/2deecbf1-5215-4055-8b84-e45ddd0cac34)

My suggestions:
* Make them easier to find, maybe use https://github.com/marketplace/actions/publish-test-report  to pushish a report in summary
[...]

PR #17924, opened 2024-05-01
testgrid-analysis: create issues based on test set flakiness instead of individual tests.

siyuanfoundation:

https://github.com/etcd-io/etcd/issues/17773#issuecomment-2084928796

Tested locally, example issue: https://github.com/siyuanfoundation/etcd/issues/53


tools/testgrid-analysis/cmd/data.go
```diff
@@ -150,6 +202,24 @@ func intStatusSet(statuses []statuspb.TestStatus) map[int32]struct{} {
...
-	return parts[len(parts)-1]
+	parts := strings.Split(fullname, "/")
+	// keep the package name of the test.
+	if len(parts) < 2 || strings.Contains(parts[len(parts)-1], ".") {
+		return parts[len(parts)-1]
+	}
+	return parts[len(parts)-2]
+}
+
+func mergeSet(from, to map[string]struct{}) {
```
ivanvc, line 213: Maybe `mergeMaps(...)` is a better name for the function, as technically, they are maps in Go.

ivanvc:
This report looks useful. However, I'm not sure if having it as a GitHub issue makes sense. I feel like we'll just need to close the report issues, which may add to the work of managing issues.

That being said, I agree that there's great value in having the reports more visible rather than nested in a GitHub action run. My suggestion would be to explore [Job Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/). Also, something that would be useful if we decide to use Job Summaries, would be to provide the link to the GitHub action run in the generated issue.

cc. @serathius, as you suggested more visibility of the reports.

jmhbnz:
How spammy do we expect this new issue opening to be? We already have quite a few open flake issues https://github.com/etcd-io/etcd/issues?q=is%3Aopen+is%3Aissue+label%3Atype%2Fflake

We might end up having to continually close issues as duplicates until some existing known flakes are resolved which could be annoying.

Some description of what impact this pr is expected to have would help make a decision on if it will be helpful or harmful.

siyuanfoundation:
I will re-evaluate if this PR is necessary. Mark it as draft for now.

PR #17891, opened 2024-04-28
[RFC] Attempt to make Blockhole blocking on L7

henrybear327:
Based on [the idea from @fuweid](https://github.com/etcd-io/etcd/issues/17737#issuecomment-2053608094), this PR attempts to employ blocking on L7 but without using external tools.

# Main ideas
The main idea is
1) utilize the `X-PeerURLs` field from the header, as we know all traffic to peer will contain this header field.
2) As we also know all nodes will create direct connections with their peers, so the traffic blocking will have to happen at all nodes' proxies, contrary to the current design, where only the proxy of the peer is being blackholed.

Based on the main ideas, we introduce an SSL termination proxy so we can obtain the `X-PeerURLs` field from the header.

# PoC Implementation Issues

There are 2 known issues with this approach
1) still leaking some traffic. But the leaked traffic (as discussed later, it won't affect the blackhole idea that we would like to achieve) as stream and pipeline traffic between raft nodes are now properly terminated
2) we would need to employ an SSL termination proxy, which might lead to a small performance hit

For 1), this way of blocking (by utilizing `X-PeerURLs` from the header) will miss certain types of traffic to certain endpoints, as the traffic to some endpoints doesn't have the `X-PeerURLs` field. Currently, the known ones are: `/members`, `/version`, and `/raft/probing`.
As you can see from the log, its header doesn't contain the `X-PeerURLs` field, but only the following fields:
- `map[Accept-Encoding:[gzip] User-Agent:[Go-http-client/1.1]] /members`
- `map[Accept-Encoding:[gzip] User-Agent:[Go-http-client/1.1]] /version`
- `map[Accept-Encoding:[gzip] User-Agent:[Go-http-client/1.1]] /raft/probing`

For 2), in order to read out `X-PeerURLs` from the header, we need to terminate the SSL connection, as we can't drop cleartext traffic (ref [1]). Thus, a new option `e2e.WithSSLTerminationProxy(true)` is introduced, which will change the network flow into
```
A -- B's SSL termination proxy - B's transparent proxy - B
       ^ newly introduced                ^ in the original codebase
```

# Known improvements required before turning RFC into PR

The prototype needs to be further improved for code review after fixing the following issues:
- blocking only RX or TX traffic (as currently a call to `blackholeTX` or `blackholeRX` will black both TX and RX traffic instead of just the specified one.
- slowness when performing test cleanup (I think this is related to the SSL timeout setting, but I haven't looked into it yet)
- coding style improvements

References:
[...]

Issue #17737, opened 2024-04-08
Blackhole failpoint in the proxy does not block all updates


siyuanfoundation:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

When mocking a network partition in e2e test with the `proxy.BlackholeTx()` and `proxy.BlackholeRx()`, the partitioned follower node can still received all the write updates happening during that blackhole period. 

When the partitioned node was the original leader, new write updates are not applied to the partitioned node. 

This bug makes the reliability of existing tests depending on this failpoint questionable. 

### What did you expect to happen?

The blackhole failpoint should drop all packets sent to the partitioned node, and it should not receive any write updates happening during that blackhole period. 

### How can we reproduce it (as minimally and precisely as possible)?

https://github.com/etcd-io/etcd/pull/17736

```bash
cd tests/e2e
go test -run TestBlackholeByMockingPartitionFollower -v
```

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here
etcd Version: 3.6.0-alpha.0
Git SHA: 733aa6bd8
Go Version: go1.22.0
Go OS/Arch: linux/amd64

$ etcdctl version
# paste output here
etcdctl version: 3.6.0-alpha.0
API version: 3.6
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

siyuanfoundation:
> Can I attempt this issue :) @siyuanfoundation

Of course :) Thank you for volunteering!

siyuanfoundation:
cc @serathius @ahrtr

serathius:
[...]

PR #17661, opened 2024-03-27
Support setting LocalAddr in peer communication - with e2e tests

flawedmatrix:
This PR builds upon #17085 in addressing issue #17068.

The e2e test added includes a positive test that --set-member-localaddr is needed for an etcd node to come up, and a test that without it, etcd will fail to connect to peers due to the certificate being issued for a different IP.

I did have to add in some cert generation to create certificates for the host IP where the test is running.

There is an implementation detail here that `--set-member-localaddr` will set the LocalAddr only if there is a specified non-loopback IP address in the `--initial-advertise-peer-urls` list. But I don't know if this is too restrictive.

PR #17938, opened 2024-05-05

henrybear327:
Based on Fu Wei's idea discussed in the issue [1], we employ the network traffic blocking on L7, using a forward proxy, without the need to use external tools.

[Background]

A peer will
(a) receive traffic from its peers
(b) initiate connections to its peers (via stream and pipeline).

Thus, the current mechanism of only blocking peer traffic via the peer's existing reverse proxy is insufficient, since only scenario (a) is handled, and scenario (b) is not blocked at all.

[Proposed solution]

We introduce a forward proxy for each peer, which will be proxying all the connections initiated from a peer to its peers.

We will remove the current use of the reverse proxy, as the forward proxy holds the information of the destination, we can block all in and out traffic that is initiated from a peer to others, without having to resort to external tools, such as iptables.

The modified architecture will look something like this:
```
A --- A's forward proxy ----- B
   ^ newly introduced
```

It's verified that the blocking of traffic is complete, compared to previous solutions attempted in PRs [2][3].

[Implementation]

The main subtasks are
- redesigned as an L7 forward proxy
- Unix socket support is dropped: e2e test supports unix sockets for peer communication, but only several e2e test cases use Unix sockets as majority of e2e test cases use HTTP/HTTPS
- introduce a new environment variable `E2E_TEST_FORWARD_PROXY_IP`
- implement L7 forward proxy by drastically reducing the existing proxy server code and design to use blocking traffic

Known limitations are
- Doesn't support unix socket (L7 HTTP transport proxy only supports HTTP/HTTPS/and socks5)
- It's L7 so we need to send a perfectly crafted HTTP request
-Doesn’t support reordering, dropping, etc. packets on-the-fly

[Testing]
- `make gofail-enable && make build && make gofail-disable && go test -timeout 60s -run ^TestBlackholeByMockingPartitionLeader$ go.etcd.io/etcd/tests/v3/e2e -v -count=1`
- `make gofail-enable && make build && make gofail-disable && go test -timeout 60s -run ^TestBlackholeByMockingPartitionFollower$ go.etcd.io/etcd/tests/v3/e2e -v -count=1`
- `go test -timeout 30s -run ^TestServer_ go.etcd.io/etcd/pkg/v3/proxy -v -failfast`

[References]
[1] issue https://github.com/etcd-io/etcd/issues/17737
[2] PR (V1) https://github.com/henrybear327/etcd/tree/fix/e2e_blackhole
[3] PR (V2) https://github.com/etcd-io/etcd/pull/17891
[4] https://github.com/etcd-io/etcd/pull/17938#discussion_r1615622709
[5] https://github.com/etcd-io/etcd/pull/17985#discussion_r1598020110

### DIFF
--- a/client/pkg/transport/transport.go
+++ b/client/pkg/transport/transport.go
@@ -18,11 +18,28 @@ import (
 	"context"
 	"net"
 	"net/http"
+	"net/url"
+	"os"
 	"strings"
 	"time"
 )
 
 type unixTransport struct{ *http.Transport }
+
+var httpTransportProxyParsingFunc = determineHTTPTransportProxyParsingFunc
+
+func determineHTTPTransportProxyParsingFunc() func(req *http.Request) (*url.URL, error) {
+	// according to the comment of http.ProxyFromEnvironment: if the proxy URL is "localhost"
+	// (with or without a port number), then a nil URL and nil error will be returned.
+	// Thus, we workaround this limitation by manually setting an ENV named E2E_TEST_FORWARD_PROXY_IP
+	// and parse the URL (which is a localhost in our case)
+	if forwardProxy, exists := os.LookupEnv("E2E_TEST_FORWARD_PROXY_IP"); exists {
+		return func(req *http.Request) (*url.URL, error) {
+			return url.Parse(forwardProxy)
+		}
+	}
+	return http.ProxyFromEnvironment
+}
 
 func NewTransport(info TLSInfo, dialtimeoutd time.Duration) (*http.Transport, error) {
 	cfg, err := info.ClientConfig()
@@ -39,7 +56,7 @@ func NewTransport(info TLSInfo, dialtimeoutd time.Duration) (*http.Transport, er
 	}
 
 	t := &http.Transport{
-		Proxy: http.ProxyFromEnvironment,
+		Proxy: httpTransportProxyParsingFunc(),
 		DialContext: (&net.Dialer{
 			Timeout:   dialtimeoutd,
 			LocalAddr: ipAddr,
@@ -60,7 +77,7 @@ func NewTransport(info TLSInfo, dialtimeoutd time.Duration) (*http.Transport, er
 		return dialer.DialContext(ctx, "unix", addr)
 	}
 	tu := &http.Transport{
-		Proxy:               http.ProxyFromEnvironment,
+		Proxy:               httpTransportProxyParsingFunc(),
 		DialContext:         dialContext,
 		TLSHandshakeTimeout: 10 * time.Second,
 		TLSClientConfig:     cfg,
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -19,6 +19,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
+	"math/bits"
 	mrand "math/rand"
 	"net"
 	"net/http"
@@ -44,11 +46,17 @@ var (
 // latency spikes and packet drop or corruption. The proxy overhead is very
 // small overhead (<500μs per request). Please run tests to compute actual
 // overhead.
+//
+// Note that the current implementation is a forward proxy, thus, unix socket
+// is not supported, due to the forwarding is done in L7, which requires
+// properly constructed HTTP header and body
+//
+// Also, because we are forced to use TLS to communicate with the proxy server
+// and using well-formed header to talk to the destination server,
+// we can't do random modification on the data on-the-fly anymore.
 type Server interface {
-	// From returns proxy source address in "scheme://host:port" format.
-	From() string
-	// To returns proxy destination address in "scheme://host:port" format.
-	To() string
+	// Listen returns proxy listen address in "scheme://host:port" format.
+	Listen() string
 
 	// Ready returns when proxy is ready to serve.
 	Ready() <-chan struct{}
@@ -115,6 +123,16 @@ type Server interface {
 	// UnblackholeRx removes blackhole operation on "receiving".
 	UnblackholeRx()
 
+	// BlackholePeerTx drops all outgoing traffic of a peer.
+	BlackholePeerTx(peer url.URL)
+	// UnblackholePeerTx removes blackhole operation on "sending".
+	UnblackholePeerTx(peer url.URL)
+
+	// BlackholePeerTx drops all incoming traffic of a peer.
+	BlackholePeerRx(peer url.URL)
+	// UnblackholePeerRx removes blackhole operation on "receiving".
+	UnblackholePeerRx(peer url.URL)
+
 	// PauseTx stops "forwarding" packets; "outgoing" traffic blocks.
 	PauseTx()
 	// UnpauseTx removes "forwarding" pause operation.
@@ -124,29 +142,29 @@ type Server interface {
 	PauseRx()
 	// UnpauseRx removes "receiving" pause operation.
 	UnpauseRx()
-
-	// ResetListener closes and restarts listener.
-	ResetListener() error
 }
 
 // ServerConfig defines proxy server configuration.
 type ServerConfig struct {
 	Logger        *zap.Logger
-	From          url.URL
-	To            url.URL
+	Listen        url.URL
 	TLSInfo       transport.TLSInfo
 	DialTimeout   time.Duration
 	BufferSize    int
 	RetryInterval time.Duration
 }
 
+const (
+	blackholePeerTypeNone uint8 = iota
+	blackholePeerTypeTx
+	blackholePeerTypeRx
+)
+
 type server struct {
 	lg *zap.Logger
 
-	from     url.URL
-	fromPort int
-	to       url.URL
-	toPort   int
+	listen     url.URL
+	listenPort int
 
 	tlsInfo     transport.TLSInfo
 	dialTimeout time.Duration
@@ -158,11 +176,12 @@ type server struct {
 	donec  chan struct{}
 	errc   chan error
 
-	closeOnce sync.Once
-	closeWg   sync.WaitGroup
+	closeOnce         sync.Once
+	closeWg           sync.WaitGroup
+	closeHijackedConn sync.WaitGroup
 
 	listenerMu sync.RWMutex
-	listener   net.Listener
+	listener   *customListener
 
 	pauseAcceptMu sync.Mutex
 	pauseAcceptc  chan struct{}
@@ -187,6 +206,11 @@ type server struct {
 
 	latencyRxMu sync.RWMutex
 	latencyRx   time.Duration
+
+	blackholePeerMap   map[int]uint8 // port number, blackhole type
+	blackholePeerMapMu sync.RWMutex
+
+	httpServer *http.Server
 }
 
 // NewServer returns a proxy implementation with no iptables/tc dependencies.
@@ -195,8 +219,7 @@ func NewServer(cfg ServerConfig) Server {
 	s := &server{
 		lg: cfg.Logger,
 
-		from: cfg.From,
-		to:   cfg.To,
+		listen: cfg.Listen,
 
 		tlsInfo:     cfg.TLSInfo,
 		dialTimeout: cfg.DialTimeout,
@@ -211,17 +234,12 @@ func NewServer(cfg ServerConfig) Server {
 		pauseAcceptc: make(chan struct{}),
 		pauseTxc:     make(chan struct{}),
 		pauseRxc:     make(chan struct{}),
+
+		blackholePeerMap: make(map[int]uint8),
 	}
 
-	_, fromPort, err := net.SplitHostPort(cfg.From.Host)
-	if err == nil {
-		s.fromPort, _ = strconv.Atoi(fromPort)
-	}
-	var toPort string
-	_, toPort, err = net.SplitHostPort(cfg.To.Host)
-	if err == nil {
-		s.toPort, _ = strconv.Atoi(toPort)
-	}
+	var err error
+	var fromPort string
 
 	if s.dialTimeout == 0 {
 		s.dialTimeout = defaultDialTimeout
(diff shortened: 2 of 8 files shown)


ivanvc:
/retitle Fix Blackhole implementation for e2e tests

pkg/proxy/server.go
```diff
@@ -353,6 +367,40 @@ func (s *server) listenAndServe() {
 			continue
 		}
 
+		parseHeaderForDestination := func() string {
+			// the first request should always contain a CONNECT header field
+			// since we set the transport to forward the traffic to the proxy
+			buf := make([]byte, s.bufferSize)
+			var data []byte
+			if nr1, err := in.Read(buf); err != nil {
+				if err == io.EOF {
```
fuweid, line 376: NOTE: During teardown, the connection will be closed and `in.Read` returns EOF. It will panic.

henrybear327: Thanks for confirming! I suspected this to be the cause but I will have to check it in the ioCopy function too!

pkg/proxy/server.go
```diff
@@ -353,6 +367,40 @@ func (s *server) listenAndServe() {
...
+				data = buf[:nr1]
+			}
+
+			// attempt to parse for the HOST from the CONNECT request
+			var req *http.Request
+			if req, err = http.ReadRequest(bufio.NewReader(bytes.NewReader(data))); err != nil {
+				panic("Failed to parse header in forward proxy")
+			}
+
+			if req.Method == http.MethodConnect {
```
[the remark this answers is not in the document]
henrybear327: hmm... I am not sure what you mean by this 
I will sync with you offline and post the result here!

pkg/proxy/server.go
fuweid: I think you can just return error and allow caller to close connection because it's closed by peer.

pkg/proxy/server.go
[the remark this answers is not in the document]
henrybear327: Reference: https://github.com/golang/go/blob/master/src/vendor/golang.org/x/net/http/httpproxy/proxy.go#L115, https://github.com/etcd-io/etcd/pull/17938/files#diff-110c492a19c9cff872b93ff979483976e6c752900453b2b9b83627bdcdfa0026R37

The thing is that both HTTP_PROXY and HTTPS_PROXY won't work if we use localhost as its proxy URL! So I have to set the transport proxy parsing myself!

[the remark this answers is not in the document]
henrybear327: May I ask what you mean by `what about introducing http proxy as new peer proxy plugin?` :) 

The current PR basically sets up a forward proxy per peer such that all the connections initiated from a peer will go through its own forward proxy. Thus, we can blackhole the traffic as we do with the existing proxy.

client/pkg/transport/transport.go
```diff
@@ -31,7 +33,17 @@ func NewTransport(info TLSInfo, dialtimeoutd time.Duration) (*http.Transport, er
 	}
 
 	t := &http.Transport{
-		Proxy: http.ProxyFromEnvironment,
+		Proxy: func(req *http.Request) (*url.URL, error) {
```
serathius, line 36: Why not just use the standard `HTTP_PROXY` environment variable? For localhost it you need to just to explicitly set Proxy to `ProxyFromEnvironment`

pkg/proxy/server.go
```diff
@@ -130,18 +132,21 @@ type Server interface {
...
-	BufferSize    int
-	RetryInterval time.Duration
+	Logger         *zap.Logger
+	From           url.URL
+	To             url.URL
+	TLSInfo        transport.TLSInfo
+	DialTimeout    time.Duration
+	BufferSize     int
+	RetryInterval  time.Duration
+	IsForwardProxy bool
```
serathius, line 142: Why do we need to distinguish forward proxy? 

Isn't forward proxy doing exactly the same? Maybe I missed something, but my understanding of the proposed solution was to just add additional hop.

client/pkg/transport/transport.go
henrybear327: Because of this (comment taken from Golang codebase) 
https://github.com/etcd-io/etcd/pull/17938/files#diff-110c492a19c9cff872b93ff979483976e6c752900453b2b9b83627bdcdfa0026R37

Basically, localhost and 127.0.0.1 will be ignored by Golang when setting those 2 ENV variables (tried and verified that it's indeed not working)

@fuweid and I synced offline and I will put up a fix to further optimize the proxy setup!

serathius:
cc @aojea

pkg/proxy/server.go
henrybear327: The original problem that the traffic is leaking is because the connection initiated from a peer will go through the destination peer's proxy then to the destination peer, and these will not be blackholed since we only drop traffic on the sender peer's proxy. 

Thus, this PR intends to fix this by adding a forward proxy on each peer. By doing so, we can perform traffic dropping just on the sender peer.

serathius: Ok, I thought about this. As you mentioned at the beginning one proxy functions as reverse-proxy other as forward proxy. They use somewhat different code.

henrybear327: I will put up another final solution to reduce the total proxy count to only 1, and see if that will work

client/pkg/transport/transport.go
serathius: Nice, sorry for asking without reading the comment, looks like you did your research. Great work!

henrybear327: No worries! I should have done a better job explaining all these in the PR description in the first place 🙃

Notes taken!

pkg/proxy/server.go
henrybear327: @serathius this PR https://github.com/etcd-io/etcd/pull/17950 is what @fuweid and I discussed offline - trying to use just 1 HTTP proxy to fulfill the blackhole requirement.

pkg/proxy/server.go
```diff
@@ -216,10 +223,12 @@ func NewServer(cfg ServerConfig) Server {
...
-	var toPort string
-	_, toPort, err = net.SplitHostPort(cfg.To.Host)
-	if err == nil {
-		s.toPort, _ = strconv.Atoi(toPort)
+	if !s.isHTTPProxy {
+		var toPort string
+		_, toPort, err = net.SplitHostPort(cfg.To.Host)
+		if err == nil {
+			s.toPort, _ = strconv.Atoi(toPort)
+		}
```
ivanvc, line 231: nit: As `toPort` and `err` are only used in this scope (correct me if I'm wrong), this can be simplified to:

```suggestion		
		if _, toPort, err := net.SplitHostPort(cfg.To.Host); err == nil {
			s.toPort, _ = strconv.Atoi(toPort)
		}
```

I know you're not introducing this code, but it may be a good opportunity for a clean-up.

serathius:
Talked with @aojea, proposed L7 forward proxy is an improvement over the previous faulty approach, however it might not be necessarily what we want. We want to simulate a realistic network disruptions, and there is no way to do that using a L7 proxy. This is currently impossible with current setup where etcd processes communicate via localhost, to modify packets we need them to go through kernel network stack, and that's not possible if they share the same network stack. 

Unfortunately setting dedicated network stack this is pretty complicated, @aojea suggest using docker which sets it up for each container, however I'm worried about how much using docker will complicate etcd e2e testing. I would prefer to set up network ourselves, however this makes it less portable. Main benefit of using docker is portability across platforms.

As we won't be able to build such solution anytime soon, I'm ok with still using L7 http proxy in the meantime.

client/pkg/transport/transport.go
```diff
@@ -31,7 +33,17 @@ func NewTransport(info TLSInfo, dialtimeoutd time.Duration) (*http.Transport, er
...
+			// according to the comment of http.ProxyFromEnvironment: if the
+			// proxy URL is "localhost" (with or without a port number),
+			// then a nil URL and nil error will be returned.
+			// Thus, we need to workaround this by manually setting an
+			// ENV named FORWARD_PROXY and parse the URL (which is a localhost in our case)
+			if httpProxy, exists := os.LookupEnv("FORWARD_PROXY"); exists {
```
serathius, line 42: Don't know if we want to support a new environment variable `FORWARD_PROXY` for proxying.

I know that `http.ProxyFromEnvironment(req)` will ignore `HTTP_PROXY` when it URL targets localhost, but what about just changing the `LookupEnv` to check for `HTTP_PROXY` and return early?


cc @ahrtr

tests/framework/e2e/etcd_process.go
```diff
@@ -72,6 +73,7 @@ type EtcdServerProcess struct {
 	cfg        *EtcdServerProcessConfig
 	proc       *expect.ExpectProcess
 	proxy      proxy.Server
+	httpProxy  proxy.Server
```
serathius, line 76: Don't know if I like the name `httpProxy`. Preferred the previous naming.
Why not have `reverseProxy` and `forwardProxy`?

henrybear327:
> Talked with @aojea, proposed L7 forward proxy is an improvement over the previous faulty approach, however it might not be necessarily what we want. We want to simulate a realistic network disruptions, and there is no way to do that using a L7 proxy. This is currently impossible with current setup where etcd processes communicate via localhost, to modify packets we need them to go through kernel network stack, and that's not possible if they share the same network stack.
> 
> Unfortunately setting dedicated network stack this is pretty complicated, @aojea suggest using docker which sets it up for each container, however I'm worried about how much using docker will complicate etcd e2e testing. I would prefer to set up network ourselves, however this makes it less portable. Main benefit of using docker is portability across platforms.
> 
> As we won't be able to build such solution anytime soon, I'm ok with still using L7 http proxy in the meantime.

Hey @serathius, thanks for the detailed write-up! I agree that what we want to do eventually, like dropping HTTP packets instead of having to resort to HTTPS, reordering of the packets, etc., isn't doable on L7. I will finish this PR properly, and then research the potential setup using docker and submit a PoC afterward!  

@fuweid and I have discussed my other PR changes (https://github.com/etcd-io/etcd/pull/17950#discussion_r1592403773), and I am currently working on implementing it (PR should be updated in a couple of hours!). By doing it this way, we will not need to have both forward and reverse proxies, just 1 forward proxy per peer, and we should be good to go!

WDYT? :)

client/pkg/transport/transport.go
henrybear327: The reason I did it this way is that I don't want to interfere with the default Golang behavior, since this code change will be reflected in the normally built binaries and I am not sure if this will break some other parts of the system or existing user assumption/usage. (my main concern is that `HTTP_PROXY` or `HTTPS_PROXY` parsing behavior will change if we use `LookupEnv` to intercept the `HTTP_PROXY` or `HTTPS_PROXY` and if it's set with localhost or 127.0.0.1)

I am too new to judge so I need @ahrtr 's help to confirm this!

tests/framework/e2e/etcd_process.go
henrybear327: I will make the naming consistent throughout the PR!

client/pkg/transport/transport.go
ahrtr: I suggest that we do not change this very basic common function. Instead, we can extend its functionality on top of it in **test** only using the [Decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern).

Specifically, you can add a similar function `NewTransport` in test framework, it calls this function, but it also lookup the environment variable "`HTTP_PROXY`" or "`HTTPS_PROXY`", and set the proxy no matter it's localhost or not.

ahrtr: The other option is to enhance the test framework to enable contributors to use a non localhost IP, so that you can use environment variable "HTTP_PROXY" or "HTTPS_PROXY" to set a proxy directly, and no need to change this common function `NewTransport`, similar to what https://github.com/etcd-io/etcd/pull/17661 does.

ahrtr: Clarify on my previous comment https://github.com/etcd-io/etcd/pull/17938#discussion_r1595329790, 

We can enable users to inject a customized `NewTransport`, similar to 
https://github.com/etcd-io/etcd/blob/e1244f19d66aff5714913efe2ddfc6a971a8f784/client/pkg/srv/srv.go#L29-L30

serathius:
This is awesome! Thanks so much for taking the lead on this. Networking can be tough, but you and @fuweid are totally crushing it!

henrybear327:
> This is awesome! Thanks so much for taking the lead on this. Networking can be tough, but you and @fuweid are totally crushing it!

I really have to say thank you to @fuweid for his guidance and PR reviews in the past weeks, and @siyuanfoundation too for reviewing my PRs!

tests/framework/e2e/etcd_process.go
henrybear327: The naming convention is now aligned! 
Necessary blackhole-related failpoint changes are also made!

client/pkg/transport/transport.go
henrybear327: Thanks @ahrtr for the code snippets! 

After looking into the suggestions, I think it's hard to obtain a non-localhost IP and use it as a proxy during e2e testing, so I went on to look into injecting a customized `NewTransport`.

My understanding is that the e2e test is not like a unit test where we can change the function pointers during the execution of the unit test dynamically (like https://github.com/etcd-io/etcd/pull/17938#discussion_r1595352530) because they are in the same module. Instead, the client code is compiled before the e2e test is run. Therefore, some sort of communication channel is required for the e2e test to indicate its need to change the code path during testing (not as easy as swapping `lookupSRV`), so I use the ENV `E2E_TEST_FORWARD_PROXY_IP` as a way to pass in the proxy address. 

My current attempt is https://github.com/etcd-io/etcd/pull/17938/commits/c93e3fc05d8d23f210a6b624514eeae5e95a7042. The idea is kind of borrowed from https://github.com/etcd-io/etcd/pull/17938#discussion_r1595352530, where the HTTP proxy parsing function in use will be determined during the first `Transport` initialization. I currently can't come up with a cleaner way without having to use a custom ENV though... :(  I am not using `HTTP_PROXY` or `HTTPS_PROXY` as we would like to not affect the default behavior of `http.ProxyFromEnvironment` (not parse `localhost` addresses), and only use a custom `NewTransport` during e2e testing only when `E2E_TEST_FORWARD_PROXY_IP` is set. 

Not sure if this makes sense to you! Happy to know if there is a better way of implementing this. :)

pkg/proxy/server.go
henrybear327: You are right! Applied the suggestion! Thanks @ivanvc :)

henrybear327:
> > Talked with @aojea, proposed L7 forward proxy is an improvement over the previous faulty approach, however it might not be necessarily what we want. We want to simulate a realistic network disruptions, and there is no way to do that using a L7 proxy. This is currently impossible with current setup where etcd processes communicate via localhost, to modify packets we need them to go through kernel network stack, and that's not possible if they share the same network stack.
> > Unfortunately setting dedicated network stack this is pretty complicated, @aojea suggest using docker which sets it up for each container, however I'm worried about how much using docker will complicate etcd e2e testing. I would prefer to set up network ourselves, however this makes it less portable. Main benefit of using docker is portability across platforms.
> > As we won't be able to build such solution anytime soon, I'm ok with still using L7 http proxy in the meantime.
> 
> Hey @serathius, thanks for the detailed write-up! I agree that what we want to do eventually, like dropping HTTP packets without having to resort to HTTPS, reordering of the packets, etc., isn't doable on L7. I will finish this PR properly, and then research the potential setup using docker and submit a PoC afterward!
> 
> @fuweid and I have discussed my other PR changes [#17950 (comment)](https://github.com/etcd-io/etcd/pull/17950#discussion_r1592403773), and I am currently working on implementing it. By doing it this way, we will not need to have both forward and reverse proxies, just 1 forward proxy per peer, and we should be good to go!
> 
> WDYT? :)

This PR is forward + reverse proxy, and https://github.com/etcd-io/etcd/pull/17985 (a PoC) is keeping only the forward proxy. 

This current PR is easier to blackhole traffic, since for the peer we would like to blackhole, we just need to drop all the traffic to and from its forward and reverse proxies, whereas if we only have the reverse proxy, we need to configure proxies from all peers to perform traffic filtering. But if we only keep the forward proxy, we reduce the hop count on our e2e testing network.

henrybear327:
@fuweid Based on the discussion [here](https://github.com/etcd-io/etcd/pull/17985#discussion_r1598020110), I implemented the L7 proxy via ServeHTTP function, and now only keeping 1 forward proxy per peer.

I did not have a unified implementation for this PR because the current proxy implementation has "pause/delay accepting connections" features built-in, and these can only be done at L4 instead of L7.

I checked the codebase and "pause accepting connections" isn't used, and "delay accepting connections" is only used during the proxy server's unit test, so in a sense, we can remove the L4 server altogether if we remove those 2 features. But if we ever need those 2 features again, we will need to reimplement the L4 proxy...

serathius:
Any progress?

henrybear327:
> Any progress? We don't need a perfect solution, just need something working.

Yes @serathius, @fuweid and I came up with a working fix a while ago already! 

Please take a look at https://github.com/etcd-io/etcd/pull/17938#discussion_r1615622709 and let us know what your thoughts are :) The PR can be prepared and ready for review within a day of work if the direction is set!

Thank you!

serathius:
I still don't know what the status is. Please understand how open source review works.

Maintainers review tens of PRs and issues daily, sometimes some comment notifications get lost sometimes we don't have time to deep dive on the whole discussion thread. 

For me PR is stuck in draft and I didn't see a reason to review it until:
* Some other community reviewer gives a LGTM
* PR is marked as ready to review by the author

If you need any answers from maintainers please post them like:

```
Need to make a decision, can X and Y take a look and help us answer the questions:
1. Question 1
2. Question 2
```

pkg/proxy/server.go
```diff
@@ -123,25 +134,31 @@ type Server interface {
 	PauseRx()
 	// UnpauseRx removes "receiving" pause operation.
 	UnpauseRx()
-
-	// ResetListener closes and restarts listener.
-	ResetListener() error
 }
 
 // ServerConfig defines proxy server configuration.
 type ServerConfig struct {
```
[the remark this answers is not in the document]
serathius: Don't know I'm not TCP expert, if it works then it works. Maintainers cannot research every topic, please list possible options, their pros and cons and your suggestion.

henrybear327:
As discussed with Marek during the robustness meeting, we will keep only the minimum functionality (just blocking traffic and nothing else), and move the features such as network packet ordering, etc. to a new proper fix, potentially using containers.

henrybear327:
Hey @fuweid @serathius @ahrtr @ivanvc, sorry for the long wait... but this PR is finally in shape for another pass! :) 

Thank you!

serathius:
This PR is too big to be reviewed in the current state, I see you have already nicely separate the commits, maybe we could take those commits and merge them independently. For example:
* Remove pose
* Remove packet modification
* Improve naming
* Remove latency accept

Removal of unused L4 proxy features seems ok to be done even before we fix the blackhole. 

Please feel free to create a PR for each commit and assign them for me to review. It should be very quick to merge.

henrybear327:
> This PR is too big to be reviewed in the current state, I see you have already nicely separate the commits, maybe we could take those commits and merge them independently. For example:
> 
> * Remove pose
> * Remove packet modification
> * Improve naming
> * Remove latency accept
> 
> Removal of unused L4 proxy features seems ok to be done even before we fix the blackhole.
> 
> Please feel free to create a PR for each commit and assign them for me to review. It should be very quick to merge.

Thanks Marek for the ideas! :)

I will do that tonight!

Issue #17964, opened 2024-05-07
Bump go 1.21.10 and 1.22.3


ahrtr:
### What would you like to be added?

Both 1.21.10 and 1.22.3 include security fixes

### Why is this needed?

fix CVE

ivanvc:
Completion tracking below:

- [ ] `main`: go v1.22.3
- [ ] `release-3.5`: go v1.21.10
- [ ] `release-3.4`: go v1.21.10
- [ ] `CHANGELOG`
- [ ] `etcd-io/bbolt` `main`: go v1.22.3
- [ ] `etcd-io/bbolt` `release-1.3`: v1.21.10
- [ ] `etcd-io/raft` `main`: go v1.22.3

Refer to previous PRs as a reference, i.e., https://github.com/etcd-io/etcd/issues/17269

lavishpal:
@ahrtr could you assign this issue to me ?

ahrtr:
> @ahrtr, do we want to update bbolt and raft? Based on recent conversations I'm unsure if that's the intention.

I think the answer is YES. We follow the same rule as documented in [dependency_management.md#golang-versions](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#golang-versions) for all repos, and also https://github.com/etcd-io/etcd/pull/17876

ahrtr:
Can we submit the PRs of bumping golang version for etcd (including main, release-3.5 and release-3.4) this week? Otherwise we will keep seeing failed workflow checks.

henrybear327:
> Can we submit the PRs of bumping golang version for etcd (including main, release-3.5 and release-3.4) this week? Otherwise we will keep seeing failed workflow checks.
> 
> We also need to release new patches for 3.5 and 3.4 soon.

It would be nice to have it done ASAP since this is blocking https://github.com/etcd-io/etcd/pull/17973 as the CI will not pass (and we would not like this to spillover to next week).

@ahrtr Maybe I can take over the etcd main branch update so I can proceed with the dependency update normally, while in the meantime @lavishpal can take his/her time working on the rest of the branches? :)

ivanvc:
Ping @lavishpal. Would you work on this this week? Otherwise, we may need to reassign to some collaborator who can help with it, as it is making it fail our CI jobs. Thanks.

lavishpal:
I will complete this within 2 days .

jmhbnz:
[...]

PR #17965, opened 2024-05-08
Fix caching by not depending on PHONY target in non-PHONY target

serathius:
cc @MadhavJivrajani @fuweid @siyuanfoundation


jmhbnz (commented):
Do we have an issue for this or some more context on what is happening with cache?

serathius:
If you run `make test-robustness-issue14370` twice, it will download and build etcd-v3.5.4 twice, when it doesn't need to.

ivanvc:
Would it make sense to define the `install-gofail` phony target as dependent on the actual `gofail` binary path? This would work, too, without having to change the target dependencies. i.e., something like:

```diff
@@ -58,13 +56,14 @@ gofail-disable:
        cd ./tests && go mod tidy
 
 .PHONY: install-gofail
-install-gofail:
+install-gofail: $(GOPATH)/bin/gofail
+
+$(GOPATH)/bin/gofail:
        go install go.etcd.io/gofail@${GOFAIL_VERSION}
```

serathius:
> Would it make sense to define the `install-gofail` phony target as dependent on the actual `gofail` binary path? This would work, too, without having to change the target dependencies. i.e., something like:
> 
> ```diff
> @@ -58,13 +56,14 @@ gofail-disable:
>         cd ./tests && go mod tidy
>  
>  .PHONY: install-gofail
> -install-gofail:
> +install-gofail: $(GOPATH)/bin/gofail
> +
> +$(GOPATH)/bin/gofail:
>         go install go.etcd.io/gofail@${GOFAIL_VERSION}
> ```

Right, but what happens when `GOFAIL_VERSION` is changed?

ivanvc:
> Right, but what happens when `GOFAIL_VERSION` is changed?

Good point, it won't work. Although we could add the version to the path, it won't be a clean solution.

serathius:
Figured out we can do it by depending on go.mod, PTAL @ivanvc

fuweid (approved):
LGTM

We need to bump go version first to make CI happy

ivanvc:
> We need to bump go version first to make CI happy

@fuweid, we're tracking that in #17964.

serathius:
> LGTM
> 
> We need to bump go version first to make CI happy

Or I can just merge :P

PR #17966, opened 2024-05-08
Relax assumptions about all client request persisted in WAL to only require first and last request to be persisted

serathius:
This assumption is not true during durability issues like #14370. In reality we want to avoid situations where WAL is was truncated, for that it's enough that we ensure that first and last operations are present.

Found it when running `make test-robustness-issue14370` and instead of getting `Model is not linearizable` I got that assumptions were broken, which is harder to understand.

cc @siyuanfoundation @MadhavJivrajani @fuweid

### DIFF
--- a/tests/robustness/validate/validate.go
+++ b/tests/robustness/validate/validate.go
@@ -149,16 +149,37 @@ func validatePersistedRequestMatchClientRequests(reports []report.ClientReport, 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


serathius:
ping @siyuanfoundation @MadhavJivrajani @fuweid

fuweid (approved):
LGTM

siyuanfoundation (approved):

Issue #17970, opened 2024-05-08
Auto label PRs on Github


serathius:
### What would you like to be added?

Would be nice for PR discovery to have them automatically labeled based on directory they touch.
For example if the test touches `./tests/robustness` we can always label it with `area/robustness-testing`.

I think K8s OWNERS support this https://go.k8s.io/owners

cc @jmhbnz 

### Why is this needed?

Better discovery of PRs for people interested in reviewing robustness test changes.

jmhbnz:
Agree this would be great to start bringing online. Thanks @liangyuanpeng for drafting the initial pr.

For this to be helpful we do need to add nested `OWNERS` files with the label stanza, for example: https://github.com/kubernetes/kubernetes/blob/master/cluster/images/OWNERS

Nested `OWNERS` we have already discussed and have consensus to progress so I will write an initial suggestion of labels here shortly so we can raise the pr.

fuweid:
Hi @jmhbnz , as far as I know, `/lgtm` command doesn't work. Is there any plan to use OWNERS file and enable `/lgtm` as review process?

jmhbnz:
> Hi @jmhbnz , as far as I know, `/lgtm` command doesn't work. Is there any plan to use OWNERS file and enable `/lgtm` as review process?

Yes. But we need to work carefully and set up the nested `OWNERS` files that it would depend on. It's a gradual journey adopting the k/k prow features, want to make sure we don't try to do too much at once and cause problems.  This auto labeling is a good driver for us to start creating nested `OWNERS`.

ah8ad3:
Very nice feature to have let me know if i can help with anything.

serathius:
@jmhbnz was https://github.com/etcd-io/etcd/pull/17979 enough to have auto labeling working? Have you tested it?

jmhbnz:
> @jmhbnz was #17979 enough to have auto labeling working? Have you tested it?

Yes it is working now, example https://github.com/etcd-io/etcd/pull/17983:

![image](https://github.com/etcd-io/etcd/assets/39425256/c6716410-96e4-4022-92f8-8f186c071081)

I still need to check nesting for label hierarchy but basics are working :)

PR #17979, opened 2024-05-10
Add nested owners files to automate labeling pull requests

jmhbnz:
This pull request proposes adding an initial set of nested `OWNERS` files with `labels:` populated in order to automate labeling for pull requests now that we have enabled the `owners-label` prow plugin in https://github.com/kubernetes/test-infra/pull/32589.

I've structured the `OWNERS` files with the assumption that inheritance works, we need to test/verify this.  For example, a pull request for `tools/rw-heatmaps` should in theory be allocated both the `area/tooling` and `area/performance` labels based on the inheritance.

Moving forward we can also use the nested `OWNERS` files to further delegate approvers and reviewers for specific directories of `etcd-io/etcd` code.

Fixes: https://github.com/etcd-io/etcd/issues/17970

### DIFF
--- a/Documentation/OWNERS
+++ b/Documentation/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/documentation
--- a/tools/etcd-dump-logs/OWNERS
+++ b/tools/etcd-dump-logs/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/debugging
--- a/.github/workflows/OWNERS
+++ b/.github/workflows/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - github_actions
--- a/client/v3/OWNERS
+++ b/client/v3/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/clientv3
--- a/contrib/OWNERS
+++ b/contrib/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/contrib
--- a/contrib/mixin/OWNERS
+++ b/contrib/mixin/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/observability
--- a/etcdctl/OWNERS
+++ b/etcdctl/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/etcdctl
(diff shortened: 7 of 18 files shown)


Documentation/OWNERS
```diff
@@ -0,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/documentation
```
serathius, line 4: The documentation directory is no longer used in the original way. Originally it included https://github.com/etcd-io/website. Now the remaining documentation is more of contributor documentation, using "area/documentation" label for this directory might be misleading.

serathius (approved):

Documentation/OWNERS
jmhbnz: Yes that directory is rather awkward. I think we have a larger decision to make on if we will shift to serving our contributor facing documentation to a new area of `etcd-io/website`.

To be transparent I would be in favor of this. It's something k8s has been doing for a while https://kubernetes.io/docs/contribute/docs.

I don't want to force that decision in this pr though, happy to come back to it later, perhaps as a topic for one of our upcoming community calls.

liangyuanpeng (commented):
LGTM

Documentation/OWNERS
jmhbnz: Based on when the linked pr was merged I would guess the unmaintained issue was a reflection of the state the project was in at that point in time around 2022? We are [in a better position now](https://etcd.devstats.cncf.io/d/74/contributions-chart?orgId=1&var-period=m&var-metric=contributions&var-repogroup_name=etcd-io%2Fwebsite&var-country_name=All&var-company_name=All&var-company=all&from=1638270000000&to=now-2d) in my view with more contributors available for website maintenance.

I like the idea of having better testing for documentation, and actually this isn't limited to contributor docs, I think many of the docs pages for our user facing documentation should have automated testing to ensure they are correct. In the ~2 years since that pr you linked was merged I don't believe we added unit tests for any of the files under our `Documentation/` repo. However if we want to do this there is no reason we can't have scheduled prow job for `etcd-io/website` to let us know if something is broken.

Let's shift this discussion to a community meeting sometime and properly explore some pros/cons. At the end of the day I want our documentation to be the best it can be, irrespective of which git repo it is stored in.

tools/etcd-dump-logs/OWNERS
```diff
@@ -0,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+labels:
+  - area/observability
```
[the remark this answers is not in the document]
ahrtr: Also add an OWNERs file under tools/etcd-dump-db with the same albel "area/debugging"

liangyuanpeng (commented):
how about add an OWNERs file  for `.github/workflows`, and label can be `github_actions`,it is also using by dependabot .

tools/etcd-dump-logs/OWNERS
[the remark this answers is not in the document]
jmhbnz: Updated 👍🏻

ahrtr (approved):
LGTM

Thanks

Issue #17870, opened 2024-04-24
fatal error: fuse3/fuse.h: No such file or directory


ahrtr:
### Which Github Action / Prow Jobs are flaking?

```
-- Build files have been written to: /tmp/lazyfs/lazyfs/build
gmake[1]: Entering directory '/tmp/lazyfs/lazyfs/build'
gmake[2]: Entering directory '/tmp/lazyfs/lazyfs/build'
gmake[3]: Entering directory '/tmp/lazyfs/lazyfs/build'
gmake[3]: Leaving directory '/tmp/lazyfs/lazyfs/build'
gmake[3]: Entering directory '/tmp/lazyfs/lazyfs/build'
[ 25%] Building CXX object CMakeFiles/lazyfs.dir/include/lazyfs/fusepp/Fuse.cpp.o
In file included from /tmp/lazyfs/lazyfs/include/lazyfs/fusepp/Fuse.cpp:26:
/tmp/lazyfs/lazyfs/include/lazyfs/fusepp/Fuse.h:36:10: fatal error: fuse3/fuse.h: No such file or directory
   36 | #include <fuse3/fuse.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
gmake[3]: *** [CMakeFiles/lazyfs.dir/build.make:76: CMakeFiles/lazyfs.dir/include/lazyfs/fusepp/Fuse.cpp.o] Error 1
gmake[3]: Leaving directory '/tmp/lazyfs/lazyfs/build'
gmake[2]: *** [CMakeFiles/Makefile2:83: CMakeFiles/lazyfs.dir/all] Error 2
gmake[2]: Leaving directory '/tmp/lazyfs/lazyfs/build'
gmake[1]: *** [Makefile:91: all] Error 2
gmake[1]: Leaving directory '/tmp/lazyfs/lazyfs/build'
make: *** [Makefile:179: bin/lazyfs] Error 2

```

### Which tests are flaking?

Robustness test

### Github Action / Prow Job link

https://github.com/etcd-io/etcd/actions/runs/8815562613/job/24197769264?pr=17868

serathius:
Github is down 
![image](https://github.com/etcd-io/etcd/assets/5873433/a3ad13b5-370f-4251-a1b0-b9ffb9021a9a)

```
Run sudo apt update && sudo apt-get --yes install cmake libfuse3-dev libfuse3-3 fuse3

[...]

PR #17984, opened 2024-05-11
github workflow: fix all apt-install  error return code

jimmy-bro:
This PR is create for #17870

### DIFF
--- a/.github/workflows/robustness-template.yaml
+++ b/.github/workflows/robustness-template.yaml
@@ -53,7 +53,7 @@ jobs:
       - name: install-lazyfs
         if: ${{ inputs.lazyfsEnabled }}
         run: |
-          sudo apt update && sudo apt-get --yes install cmake libfuse3-dev libfuse3-3 fuse3
+          sudo apt-get -o APT::Update::Error-Mode=any update && sudo apt-get --yes install cmake libfuse3-dev libfuse3-3 fuse3
           sudo sed -i 's/#user_allow_other/user_allow_other/g' /etc/fuse.conf
           make install-lazyfs
       - name: test-robustness


jimmy-bro:
There is only 1  `apt update`  in Github Action / Prow Job now，so just need to modify 1 place

ahrtr (approved):

.github/workflows/robustness-template.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: install-lazyfs
         if: ${{ inputs.lazyfsEnabled }}
         run: |
-          sudo apt update && sudo apt-get --yes install cmake libfuse3-dev libfuse3-3 fuse3
+          sudo apt -o APT::Update::Error-Mode=any update && sudo apt-get --yes install cmake libfuse3-dev libfuse3-3 fuse3
```
ivanvc, line 56: Suggestion to use `apt-get` instead of `apt`, as `apt` shows warning: 
> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

And quoting [`apt`'s man page](https://manpages.ubuntu.com/manpages/xenial/man8/apt.8.html#description):
> apt provides a high-level commandline interface for the package management system. It is
intended as an end user interface and enables some options better suited for interactive
usage by default compared to more specialized APT tools like apt-get(8) and apt-cache(8).

ivanvc (commented):
I left a suggestion. Thanks, @jimmy-bro.

.github/workflows/robustness-template.yaml
jimmy-bro: Thanks for your mentioning @ivanvc , I also think  `apt-get` is better than `apt `

ivanvc (approved):
LGTM. Thanks, @jimmy-bro

jmhbnz (approved):
LGTM - Thanks @jimmy-bro

PR #17985, opened 2024-05-12

henrybear327:

For each proxy, we remove the reverse proxy and keep only the forward proxy.

PoC-level code quality for now...

### DIFF
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -219,14 +219,11 @@ func NewServer(cfg ServerConfig) Server {
 		pauseRxc:     make(chan struct{}),
 	}
 
-	_, fromPort, err := net.SplitHostPort(cfg.From.Host)
-	if err == nil {
+	if _, fromPort, err := net.SplitHostPort(cfg.From.Host); err == nil {
 		s.fromPort, _ = strconv.Atoi(fromPort)
 	}
 	if !s.isForwardProxy {
-		var toPort string
-		_, toPort, err = net.SplitHostPort(cfg.To.Host)
-		if err == nil {
+		if _, toPort, err := net.SplitHostPort(cfg.To.Host); err == nil {
 			s.toPort, _ = strconv.Atoi(toPort)
 		}
 	}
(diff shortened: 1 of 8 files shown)


henrybear327:
TODO: 
- fix all the blackhole failpoints to use the proper blackhole function

pkg/proxy/server.go
```diff
@@ -353,6 +387,44 @@ func (s *server) listenAndServe() {
 			continue
 		}
 
+		parseHeaderForDestination := func() *string {
+			// the first request should always contain a CONNECT header field
+			// since we set the transport to forward the traffic to the proxy
+			buf := make([]byte, s.bufferSize)
```
fuweid, line 393: If this proposal is going to use http proxy, suggest to use `ServeHTTP` http handler to handle this, instead of building http function on tcp connection. It's hard to maintain that code (at least to me). The go std package provides `Hijacker` to convert connection to TCP for us.

For example,

```go
type serverHandler struct{}

func (serverHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
       // for https
       if req.Method == "CONNECT" {
              hijacker, _ := resp.(http.Hijacker)
              conn, _, err := hijacker.Hijack()
              if err != nil {
                    // write error back to conn
                    return
               }
  
               conn.Write([]byte("HTTP/1.0 200 Connection established\r\n\r\n"))
               
               // dial to target host
               targetConn, err := net.Dial("tcp", req.URL.Host)
               if err != nil {
                       // write error back to conn
                      return
                //
                go io.Copy(targetConn, conn)
                go io.Copy(conn, targetConn)

       } else { // for plain http
               // just act as agent to forward request to target host
       }
}

http.ListenAndServe(":8080", &serverHandler{})
```

fuweid: Suggest to file the change to original one and make all the CI happy because all the context is in one pull request.
You don't have to force push and `append commits` can save history. So, you don't need to worry if new proposal has been rejected.

henrybear327: Thanks for the suggestion @fuweid! 

I am in the process of refactoring/rewriting the `server.go` code. As this approach will be changing a bit how the listener is set up :)

PR #17986, opened 2024-05-12
Fix the auto-compaction-retention logging in the startup

### LINKED ISSUE #17822: auto-compaction-retention in logs displays nanoseconds when auto-compaction-mode set to revision
### Bug report criteria

### What happened?

For one of the ETCD clusters that we maintain, we've switched the `auto-compaction-mode` to revision based compaction. The `auto-compaction-retention` is set to `5000`:

```
$ grep ^auto-compaction /etc/etcd/etcd.conf.yml 
auto-compaction-mode: "revision"
auto-compaction-retention: "5000"```
```

When restarting, the etcd log shows the following:
```
{"level":"info","ts":"2024-04-18T15:27:31.278121+0200","caller":"embed/etcd.go:308","msg":"starting an etcd server","etcd-version":"3.5.12","git-sha":"e7b3bb6cc","go-version":"go1.20.13","go-os":"linux","go-arch":"amd64","max-cpu-set":2,"max-cpu-available":2,"member-initialized":true,"name":"10.10.10.10","data-dir":"/var/lib/etcd","wal-dir":"","wal-dir-dedicated":"","member-dir":"/var/lib/etcd/member","force-new-cluster":false,"heartbeat-interval":"100ms","election-timeout":"1s","initial-election-tick-advance":true,"snapshot-count":100000,"max-wals":5,"max-snapshots":5,"snapshot-catchup-entries":5000,"initial-advertise-peer-urls":["https://10.10.10.10:2380"],"listen-peer-urls":["https://10.10.10.10:2380"],"advertise-client-urls":["https://10.10.10.10:2379"],"listen-client-urls":["https://10.10.10.10:2379"],"listen-metrics-urls":["https://10.10.10.10:4001"],"cors":["*"],"host-whitelist":["*"],"initial-cluster":"","initial-cluster-state":"new","initial-cluster-token":"","quota-backend-bytes":8589934592,"max-request-bytes":1572864,"max-concurrent-streams":4294967295,"pre-vote":true,"initial-corrupt-check":false,"corrupt-check-time-interval":"0s","compact-check-time-enabled":false,"compact-check-time-interval":"1m0s","auto-compaction-mode":"revision","auto-compaction-retention":"1µs","auto-compaction-interval":"1µs","discovery-url":"","discovery-proxy":"","downgrade-check-interval":"5s"}
```

In this log, I would've expected to see `"auto-compaction-retention":"5000"`, instead it shows `"auto-compaction-retention":"1µs"`. This gives you the idea you made mistakes in the configuration.

Also, `"auto-compaction-interval":"1µs"` also seems wrong, since the interval is 5 minutes for revision based compaction (from what I saw, hardcoded in the source code)

The logs should show `auto-compaction-retention` set to `5000` instead of `1µs`

Change `auto-compaction-mode` to revision, restart and check the logs.

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.5.12
Git SHA: e7b3bb6cc
Go Version: go1.20.13
Go OS/Arch: linux/amd64

$ etcdctl version
etcdctl version: 3.5.12
API version: 3.5
```

</details>

_No response_

_No response_

### Relevant log output

_No response_

akin-tekeoglu:
Fixes #17822

It seemed to me that `auto-compaction-interval` was useless so I nuked it

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -34,6 +34,7 @@ import (
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/soheilhy/cmux"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/keepalive"
@@ -48,6 +49,7 @@ import (
 	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/etcdhttp"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
+	compactor "go.etcd.io/etcd/server/v3/etcdserver/api/v3compactor"
 	"go.etcd.io/etcd/server/v3/storage"
 	"go.etcd.io/etcd/server/v3/verify"
 )
@@ -307,6 +309,12 @@ func print(lg *zap.Logger, ec Config, sc config.ServerConfig, memberInitialized 
 	if quota == 0 {
 		quota = storage.DefaultQuotaBytes
 	}
+	var autoCompactRetention zapcore.Field
+	if sc.AutoCompactionMode == compactor.ModePeriodic {
+		autoCompactRetention = zap.Duration("auto-compaction-retention", sc.AutoCompactionRetention)
+	} else {
+		autoCompactRetention = zap.Int64("auto-compaction-retention", int64(sc.AutoCompactionRetention))
+	}
 
 	lg.Info(
 		"starting an etcd server",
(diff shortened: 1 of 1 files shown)


akin-tekeoglu:
I see your name in the owner's file @fuweid. Can you take a look?

server/embed/etcd.go
```diff
@@ -307,6 +309,12 @@ func print(lg *zap.Logger, ec Config, sc config.ServerConfig, memberInitialized
 	if quota == 0 {
 		quota = storage.DefaultQuotaBytes
 	}
+	var autoCompactRetention zapcore.Field
+	if sc.AutoCompactionMode == compactor.ModePeriodic {
+		autoCompactRetention = zap.Duration("auto-compaction-retention", sc.AutoCompactionRetention)
+	} else {
+		autoCompactRetention = zap.Int64("auto-compaction-retention", int64(sc.AutoCompactionRetention))
```
siyuanfoundation, line 316: I like this simplification for less confusion. But should we keep the name the name not to breaking existing users?

akin-tekeoglu: @siyuanfoundation I made the changes. Can you take a look again?

server/embed/etcd.go
```diff
@@ -307,6 +309,12 @@ func print(lg *zap.Logger, ec Config, sc config.ServerConfig, memberInitialized
 	if quota == 0 {
 		quota = storage.DefaultQuotaBytes
 	}
+	var autoCompactionRetention zapcore.Field
+	if sc.AutoCompactionMode == compactor.ModePeriodic {
```
fuweid, line 313: There is type alias for `compactor.ModePeriodc` in server/embed: `CompactorModePeriodic`.

Please consider to use `switch` branch because that mode can be invalid, like empty.

PR #17967, opened 2024-05-08
Update the robustness README and fix the #14370 reproduction case

serathius:
cc @siyuanfoundation @MadhavJivrajani @fuweid

PR #17989, opened 2024-05-13
Prevent picking a failpoint that waiting till snapshot that doesn't support lower snapshot catchup entries but allow reproducing issue #15271

serathius:
Overdid it in https://github.com/etcd-io/etcd/pull/17967

cc @MadhavJivrajani  for fast review.

### DIFF
--- a/tests/robustness/failpoint/network.go
+++ b/tests/robustness/failpoint/network.go
@@ -57,6 +57,11 @@ func (tb triggerBlackhole) Trigger(ctx context.Context, t *testing.T, member e2e
 }
 
 func (tb triggerBlackhole) Available(config e2e.EtcdProcessClusterConfig, process e2e.EtcdProcess) bool {
+	// Avoid triggering failpoint if waiting for failpoint would take too long to fit into timeout.
+	// Number of required entries for snapshot depends on etcd configuration.
+	if tb.waitTillSnapshot && entriesToGuaranteeSnapshot(config) > 200 {
+		return false
+	}
 	return config.ClusterSize > 1 && process.PeerProxy() != nil
 }
 
@@ -127,14 +132,18 @@ func waitTillSnapshot(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCl
 		}
 		t.Logf("clusterRevision: %d, blackholedMemberRevision: %d", clusterRevision, blackholedMemberRevision)
 		// Blackholed member has to be sufficiently behind to trigger snapshot transfer.
-		// Need to make sure leader compacted latest revBlackholedMem inside EtcdServer.snapshot.
-		// That's why we wait for clus.Cfg.SnapshotCount (to trigger snapshot) + clus.Cfg.SnapshotCatchUpEntries (EtcdServer.snapshot compaction offset)
-		if clusterRevision-blackholedMemberRevision > int64(clus.Cfg.ServerConfig.SnapshotCount+clus.Cfg.ServerConfig.SnapshotCatchUpEntries) {
+		if clusterRevision-blackholedMemberRevision > int64(entriesToGuaranteeSnapshot(*clus.Cfg)) {
 			break
 		}
 		time.Sleep(100 * time.Millisecond)
 	}
 	return nil
+}
+
+func entriesToGuaranteeSnapshot(config e2e.EtcdProcessClusterConfig) uint64 {
+	// Need to make sure leader compacted latest revBlackholedMem inside EtcdServer.snapshot.
+	// That's why we wait for clus.Cfg.SnapshotCount (to trigger snapshot) + clus.Cfg.SnapshotCatchUpEntries (EtcdServer.snapshot compaction offset)
+	return config.ServerConfig.SnapshotCount + config.ServerConfig.SnapshotCatchUpEntries
 }
 
 // latestRevisionForEndpoint gets latest revision of the first endpoint in Client.Endpoints list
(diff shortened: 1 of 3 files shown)


tests/robustness/failpoint/network.go
```diff
@@ -57,6 +57,9 @@ func (tb triggerBlackhole) Trigger(ctx context.Context, t *testing.T, member e2e
 }
 
 func (tb triggerBlackhole) Available(config e2e.EtcdProcessClusterConfig, process e2e.EtcdProcess) bool {
+	if tb.waitTillSnapshot && config.ServerConfig.SnapshotCatchUpEntries > 100 {
```
serathius, line 60: Test will not pass if SnapshotCatchUpEntries is not set to lower values, default in older releases was 10`000, which is much too much. We cannot allow failpoint randomiser to pick it.

tests/robustness/main_test.go
```diff
@@ -78,10 +78,6 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, s testSce
 			t.Fatal(err)
 		}
 	}
-	err = failpoint.Validate(r.Cluster, s.failpoint)
```
serathius, line 81: Removing validation as `PickRandom` should already limit failpoint picked to only ones passing validation. This leaves us with only failpoints that were hardcoded, which is ok as we only hardcode the reproduction cases like `make test-robustness-issue15271`.

tests/robustness/failpoint/network.go
MadhavJivrajani: Is the intention here to only manually hardcode a failpoint with a `triggerBlackhole` trigger and never to allow `PickRandom` to pick it? Because I see only this where we configure `SnapshotCatchupEntried`:
https://github.com/etcd-io/etcd/blob/333bd7bf7612a9f83cc94e8334565012e84f354f/tests/robustness/scenarios.go#L175

This sets it to `100`. And I also see `waitTillSnapshot` is always set to `true` in whatever failpoint that is defined with a `triggerBlackhole` trigger.

If this is the intention can you add a comment? Also, is it okay disallowing picking a failpoint with `triggerBlackhole` for all failpoints?
Maybe as a followup, we can disallow a particular scenario from being considered by `PickRandom`.

serathius: Intention is to revert change from https://github.com/etcd-io/etcd/pull/17967/files where this was already present.

MadhavJivrajani: Yeah, my bad for not clarifying. I meant to ask this irrespective of the revert.

serathius: Added a comment

tests/robustness/main_test.go
MadhavJivrajani: Question: Is it ever possible that a failpoint we hardcode is not `Available` for any of the cluster procs?

tests/robustness/failpoint/network.go
MadhavJivrajani: Also, the default for `SnapshotCatchupEntries` always seems to be `5000` and for `DefaultSnapshotCount` is `10000`. I'm guessing `5000` is too high as well?

serathius: Right, however in robustness test we lower snapshot count to 50-1000.

tests/robustness/main_test.go
serathius: Yes, for `make test-robustness-issue15271` it reproduces #15271 on older etcd version that didn't yet have  `SnapshotCatchupEntries` flag.

tests/robustness/failpoint/network.go
```diff
@@ -57,6 +57,10 @@ func (tb triggerBlackhole) Trigger(ctx context.Context, t *testing.T, member e2e
 }
 
 func (tb triggerBlackhole) Available(config e2e.EtcdProcessClusterConfig, process e2e.EtcdProcess) bool {
+	// Waiting for snapshot with default snapshot-catchup-entries (10`000) takes too long.
```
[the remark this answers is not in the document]
serathius: Rewrote the comment.

tests/robustness/failpoint/network.go
```diff
@@ -57,6 +57,11 @@ func (tb triggerBlackhole) Trigger(ctx context.Context, t *testing.T, member e2e
 }
 
 func (tb triggerBlackhole) Available(config e2e.EtcdProcessClusterConfig, process e2e.EtcdProcess) bool {
+	// Avoid triggering failpoint if waiting for failpoint would take too long to fit into timeout.
+	// Number of required entries for snapshot depends on etcd configuration.
+	if tb.waitTillSnapshot && entriesToGuaranteeSnapshot(config) > 200 {
```
MadhavJivrajani, line 62: If we want to completely avoid this, should this be 200?
I see we are randomly picking `SnapshotCount` from 50, 100 and 1000. If 50n is picked and `SnapshotCatchupEntries` is 100 then this branch will not be taken if I understand correctly.

MadhavJivrajani: Nvm, I think I understand now. 50 is within the timeout.

serathius: Possible steps of `SnaphostCount` + `SnapshotCatchupEntries`:
* `150`
* `200`
* `1100`
* `5050`
* `5100`
* `6000`

`200` takes around 10 seconds on my machine which is ok.

The next jump would be to `1100` which is too big.

MadhavJivrajani (approved):
Thanks for the responses!

Issue #17874, opened 2024-04-24
add etcd_debugging_lease_total gauge metric


gojoy:
### What would you like to be added?

we already have  [etcd_debugging_lease_granted_total](https://github.com/etcd-io/etcd/blob/d38f7dceb22c2aa66f749d7aa4d6a10d10bc8116/server/lease/metrics.go#L22) metric with  type counter, but we cannot observe the actual lease num in etcd by this metric.

Please add etcd_debugging_lease_total  with type gauge metric

### Why is this needed?

better observability. Monitor lease num  in etcd cluster

jimmy-bro:
Hi, @jmhbnz  ,could you please assgin this to me?

PR #17983, opened 2024-05-11
etcdserver: add server range duration metrics

thedtripp:
This PR is intended to address this issue:
https://github.com/etcd-io/etcd/issues/16866

PR #17991, opened 2024-05-13
feat: add etcd_debugging_lease_total gauge metric

jimmy-bro:
The feature was discussed in #17874


server/lease/metrics.go
```diff
@@ -49,11 +49,19 @@ var (
 			// 1 second -> 3 months
 			Buckets: prometheus.ExponentialBuckets(1, 2, 24),
 		})
+
+	leaseTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "total",
```
jimmy-bro, line 56: I‘m wondering if `active_total`  would be a more suitable name for this metric. 
Cause i feel like this description is more precise.

jimmy-bro:
Should I write unit tests? I'm willing to write unit tests for all the metrics in https://github.com/etcd-io/etcd/blob/main/server/lease/metrics.go

server/lease/lessor.go
```diff
@@ -303,6 +303,7 @@ func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
 
 	leaseTotalTTLs.Observe(float64(l.ttl))
 	leaseGranted.Inc()
+	leaseTotal.Inc()
```
gojoy, line 306: should leaseTotal Inc when etcdserver restart ? the lease may recover from backend in [initAndRecover](https://github.com/etcd-io/etcd/blob/main/server/lease/lessor.go#L790)

jimmy-bro: Thanks for your mentioning @gojoy ,I already updated the commit .
When it recover from backend, the metric will reset the value through leaseTotal.Set(float64(len(le.leaseMap))),
I used to considered doing leaseTotal.Inc inside the recover for-loop, but that would likely have a significant impact.

gojoy: LGTM

jmhbnz:
> Should I write unit tests? I'm willing to write unit tests for all the metrics in https://github.com/etcd-io/etcd/blob/main/server/lease/metrics.go

Yes please include tests, refer to https://github.com/etcd-io/etcd/pull/17983 for examples of unit and integration metrics tests.

PR #18009, opened 2024-05-14
all: fix setupEmbedCfg functions to set cfg.InitialCluster more safely

callthingsoff:
When the slice is empty, it can panic with "index 1 out of bounds". This patch rewrites these functions to set cfg.InitialCluster more safely.

### DIFF
--- a/tests/integration/embed/embed_test.go
+++ b/tests/integration/embed/embed_test.go
@@ -206,11 +206,16 @@ func setupEmbedCfg(cfg *embed.Config, curls []url.URL, purls []url.URL) {
 	cfg.ClusterState = "new"
 	cfg.ListenClientUrls, cfg.AdvertiseClientUrls = curls, curls
 	cfg.ListenPeerUrls, cfg.AdvertisePeerUrls = purls, purls
-	cfg.InitialCluster = ""
-	for i := range purls {
-		cfg.InitialCluster += ",default=" + purls[i].String()
+
+	cfg.InitialCluster = joinURLs(purls)
+}
+
+func joinURLs(purls []url.URL) string {
+	initialCluster := make([]string, len(purls))
+	for i, purl := range purls {
+		initialCluster[i] = fmt.Sprintf("default=%s", purl.String())
 	}
-	cfg.InitialCluster = cfg.InitialCluster[1:]
+	return strings.Join(initialCluster, ",")
 }
 
 func TestEmbedEtcdAutoCompactionRetentionRetained(t *testing.T) {
--- a/tests/integration/embed/join_urls_test.go
+++ b/tests/integration/embed/join_urls_test.go
@@ -1,0 +1,66 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package embed_test
+
+import (
+	"net/url"
+	"testing"
+)
+
+func parseURL(s string) url.URL {
+	x, err := url.Parse(s)
+	if err != nil {
+		panic(err)
+	}
+	return *x
+}
+
+func TestJoinURLs(t *testing.T) {
+	tests := []struct {
+		urls []url.URL
+		want string
+	}{
+		{
+			urls: []url.URL{},
+			want: "",
+		},
+		{
+			urls: []url.URL{
+				parseURL("http://example.com/#x"),
+			},
+			want: "default=http://example.com/#x",
+		},
+		{
+			urls: []url.URL{
+				parseURL("http://example.com/#x"),
+				parseURL("http://example.com/#y"),
+			},
+			want: "default=http://example.com/#x,default=http://example.com/#y",
+		},
+		{
+			urls: []url.URL{
+				parseURL("http://example.com/#x"),
+				parseURL("http://example.com/#y"),
+				parseURL("http://example.com/#z"),
+			},
+			want: "default=http://example.com/#x,default=http://example.com/#y,default=http://example.com/#z",
+		},
+	}
+	for i, tt := range tests {
+		if got := joinURLs(tt.urls); got != tt.want {
+			t.Errorf("#%d want: %s, got: %s", i, tt.want, got)
+		}
+	}
+}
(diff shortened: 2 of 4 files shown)


tests/integration/embed/embed_test.go
```diff
@@ -206,11 +206,15 @@ func setupEmbedCfg(cfg *embed.Config, curls []url.URL, purls []url.URL) {
...
 	cfg.ListenClientUrls, cfg.AdvertiseClientUrls = curls, curls
 	cfg.ListenPeerUrls, cfg.AdvertisePeerUrls = purls, purls
-	cfg.InitialCluster = ""
-	for i := range purls {
-		cfg.InitialCluster += ",default=" + purls[i].String()
+
+	if len(purls) > 0 {
+		builder := new(strings.Builder)
+		builder.WriteString(fmt.Sprintf("default=%s", purls[0].String()))
+		for _, purl := range purls[1:] {
```
siyuanfoundation, line 213: how about using `"default=" + strings.Join(purls, ",default=")`

callthingsoff: `purls []url.URL` is not `[]string`
```go
go doc strings.Join

package strings // import "strings"

func Join(elems []string, sep string) string
    Join concatenates the elements of its first argument to create a single
    string. The separator string sep is placed between elements in the resulting
    string.
```

siyuanfoundation: oh, I see

tests/integration/embed/embed_test.go
```diff
@@ -206,11 +206,15 @@ func setupEmbedCfg(cfg *embed.Config, curls []url.URL, purls []url.URL) {
 	cfg.ClusterState = "new"
 	cfg.ListenClientUrls, cfg.AdvertiseClientUrls = curls, curls
 	cfg.ListenPeerUrls, cfg.AdvertisePeerUrls = purls, purls
-	cfg.InitialCluster = ""
-	for i := range purls {
-		cfg.InitialCluster += ",default=" + purls[i].String()
+
```
siyuanfoundation, line 209:
```suggestion
builder := new(strings.Builder)
for i, purl := range(purls) {
    if i == 0 {
        builder.WriteString(fmt.Sprintf("default=%s",purl.String()))
    } else {
        builder.WriteString(fmt.Sprintf(",default=%s",purl.String()))
    }
}
cfg.InitialCluster = builder.String()
```

callthingsoff: Thanks, personally I prefer the previous version of mine.
It's consistent with other code and faster.

callthingsoff:
cc @ahrtr @serathius @jmhbnz

tests/integration/embed/embed_test.go
```diff
@@ -206,11 +206,15 @@ func setupEmbedCfg(cfg *embed.Config, curls []url.URL, purls []url.URL) {
...
+	if len(purls) > 0 {
+		builder := new(strings.Builder)
+		builder.WriteString(fmt.Sprintf("default=%s", purls[0].String()))
```
ivanvc, line 212: I don't think the `url.URL` needs explicit casting into a string. The following should have the same result.

```golang
fmt.Sprintf("default=%s", purls[0])
```

tools/etcd-dump-metrics/etcd.go
```diff
@@ -53,11 +53,14 @@ func setupEmbedCfg(cfg *embed.Config, curls, purls, ics []url.URL) {
 	cfg.ListenClientUrls, cfg.AdvertiseClientUrls = curls, curls
 	cfg.ListenPeerUrls, cfg.AdvertisePeerUrls = purls, purls
 
-	cfg.InitialCluster = ""
-	for i := range ics {
-		cfg.InitialCluster += fmt.Sprintf(",%d=%s", i, ics[i].String())
+	if len(ics) > 0 {
+		builder := new(strings.Builder)
+		builder.WriteString(fmt.Sprintf("%d=%s", 0, ics[0].String()))
```
ivanvc, line 58: See my previous comment. And, in this case, it could also be simplified to:

```go
fmt.Sprintf("0=%s", ics[0])
```

tests/integration/embed/embed_test.go
```diff
@@ -206,11 +206,15 @@ func setupEmbedCfg(cfg *embed.Config, curls []url.URL, purls []url.URL) {
...
+		for _, purl := range purls[1:] {
+			builder.WriteString(fmt.Sprintf(",default=%s", purl.String()))
```
ivanvc, line 214: Ditto

tools/etcd-dump-metrics/etcd.go
```diff
@@ -53,11 +53,14 @@ func setupEmbedCfg(cfg *embed.Config, curls, purls, ics []url.URL) {
...
+		for i, ic := range ics[1:] {
+			builder.WriteString(fmt.Sprintf(",%d=%s", i, ic.String()))
```
ivanvc, line 60: Ditto.

ivanvc (commented):
Thanks for your PR @callthingsoff. I left a possible improvement for your code :v:

serathius:
QQ, if you are touching this code. Have you double checked if it has any tests? If not please start from adding tests.

tests/integration/embed/embed_test.go
callthingsoff:
```go
fmt.Sprintf("default=%s", &purls[0]
```
I think you mean `&purls[0]`
```go
go doc url.URL.String
package url // import "net/url"

func (u *URL) String() string
```
Thanks, but I think it's just fine to call String() here and below.

tools/etcd-dump-metrics/etcd.go
callthingsoff:
```go
fmt.Sprintf("default=%s", &purls[0]
```
See my comment above.

tests/integration/embed/embed_test.go
callthingsoff: See my comment above.

tools/etcd-dump-metrics/etcd.go
callthingsoff: See my comment above.

tests/integration/embed/embed_test.go
ivanvc: Good point; I didn't check the type of `purls`.

tests/integration/embed/join_urls_test.go
```diff
@@ -0,0 +1,66 @@
...
+			urls: []url.URL{
+				parseURL("http://example.com/#x"),
+				parseURL("http://example.com/#y"),
+				parseURL("http://example.com/#z"),
+			},
+			want: "default=http://example.com/#x,default=http://example.com/#y,default=http://example.com/#z",
+		},
+	}
+	for i, tt := range tests {
+		if got := joinURLs(tt.urls); got != tt.want {
```
ahrtr, line 62: `joinURLs` is already part of the test, we are adding test for test. Not necessary to me.

tools/etcd-dump-metrics/etcd.go
```diff
@@ -53,11 +53,15 @@ func setupEmbedCfg(cfg *embed.Config, curls, purls, ics []url.URL) {
 	cfg.ListenClientUrls, cfg.AdvertiseClientUrls = curls, curls
 	cfg.ListenPeerUrls, cfg.AdvertisePeerUrls = purls, purls
 
-	cfg.InitialCluster = ""
-	for i := range ics {
-		cfg.InitialCluster += fmt.Sprintf(",%d=%s", i, ics[i].String())
```
ahrtr, line 58: It seems an unnecessary change,
- The function `setupEmbedCfg` is already very simple. 
- The new function `joinURLs` isn't used by multiple users (it has only one user).

callthingsoff:
Thanks, closing this.

Issue #17227, opened 2024-01-11
Etcdutl migrate should read WAL from last snapshot


serathius:
### What would you like to be added?

When reading `etcdutl migrate` code I noticed that I made a mistake (not a serious one :P).

In code
https://github.com/etcd-io/etcd/blob/b3bf59a3557ce81fb9d4615df667b4cacc6a4f6d/etcdutl/etcdutl/migrate_command.go#L98-L101

I pass `walpb.Snapshot{}` as a reference to snapshot from which to start reading WAL. Empty struct means that fields `Index` and `Term` are set to `0`, which I think is interpreted as read from beginning. 

Instead of reading from the beginning we should be reading from the last snapshot, as this is the part of WAL we expect to be read by etcd when started.  

The important part is to update the tests to ensure that checking WAL entries work, we need:
* Test etcdutl migrate with and without snapshot
* Test etcdutl migrate detects a newer WAL entry and will abort
* Test etcdutl migrate passes if there was a snapshot after the newer WAL entry.

Current tests: https://github.com/etcd-io/etcd/blob/b3bf59a3557ce81fb9d4615df667b4cacc6a4f6d/tests/e2e/utl_migrate_test.go

### Why is this needed?

Improve correctness of etcdutl migrate command

serathius:
cc @siyuanfoundation

fykaa:
I would like to work on this issue, @serathius. Can you assign it to me?

ahrtr:
Overall it's a good idea so that we don't lose any data, which exist in WAL but not in bbolt db. 

ping @fykaa are you still working on this?

ArkaSaha30:
@fykaa I can take this up if you do not have the bandwidth

fykaa:
Hi @ArkaSaha30, I've started working on this. Forgot updating in the comments

serathius:
Ping @fykaa, do you mind if we reassign the issue so other contributors can help.

ahrtr:
It's similar to what the etcdserver bootstrap does.

https://github.com/etcd-io/etcd/blob/f128d89cea0b8d029cdb93d672099ab598c13e20/server/etcdserver/bootstrap.go#L391-L397

@joshuazh-x do you have bandwidth to take care of this ticket?

ahrtr:
I will take over this if nobody work on it this week.

PR #18010, opened 2024-05-15
Fix: reading last snapshot in etcdutl migrate command

fykaa:
Fixes a part of #17227 

In the `etcdutl/migrate_command.go` file, I've rectified a minor oversight in the logic for reading the last snapshot during the WAL migration process. Previously, an empty `walpb.Snapshot{}` was passed as a reference, that would result in reading of the WAL from the beginning.

I will raise a follow-up PR that should also cover the addition of test cases, completing the solution for this issue


siyuanfoundation:
Thanks @fykaa for the pr. Can you fix the tests?

fykaa:
Hi @siyuanfoundation, yes, I am on it. Fixing some test bugs that occured for me.

etcdutl/etcdutl/migrate_command.go
```diff
@@ -95,7 +95,11 @@ func (o *migrateOptions) Config() (*migrateConfig, error) {
 	c.be = backend.NewDefaultBackend(GetLogger(), dbPath)
 
 	walPath := datadir.ToWALDir(o.dataDir)
-	w, err := wal.OpenForRead(c.lg, walPath, walpb.Snapshot{})
+	lastSnapshot, err := getLastSnapshotIndex(walPath)
```
MadhavJivrajani, line 98: Can we propagate `c.lg` into `getSnapshotIndex` and eventually to `ValidSnapshotEntries`? It'll be good to log errors out that happen in those calls.

etcdutl/etcdutl/migrate_command.go
```diff
@@ -95,7 +95,11 @@ func (o *migrateOptions) Config() (*migrateConfig, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf(`failed to find last snapshot: %v`, err)
+	}
```
MadhavJivrajani, line 101: Considering the past behaviour was to read from the start, should we fallback to `Index=0` in case we cannot find snapshot at a certain index?

etcdutl/etcdutl/migrate_command.go
```diff
@@ -95,7 +95,11 @@ func (o *migrateOptions) Config() (*migrateConfig, error) {
...
+	w, err := wal.OpenForRead(c.lg, walPath, walpb.Snapshot{Index: lastSnapshot})
```
MadhavJivrajani, line 102: e2e failure logs: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/18010/pull-etcd-e2e-amd64/1794986793938980864 

The error seems to be coming from:
https://github.com/etcd-io/etcd/blob/bb701b9265f31d61db5906325e0a7e2abf7d3627/etcdutl/etcdutl/migrate_command.go#L105

Error excerpt: 
```
        	Error Trace:	/home/prow/go/src/github.com/etcd-io/etcd/tests/e2e/utl_migrate_test.go:162
        	Error:      	Error "[/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdutl migrate --data-dir /tmp/TestEtctlutlMigrateMigrate_v3.5_to_v3.5_is_no-op2858091090/002/member-0 --target-version 3.5] match not found.  Set EXPECT_DEBUG for more info Errs: [unexpected exit code [128] after running [/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdutl migrate --data-dir /tmp/TestEtctlutlMigrateMigrate_v3.5_to_v3.5_is_no-op2858091090/002/member-0 --target-version 3.5]], last lines:\n2024-05-27T07:18:46Z\tinfo\tbbolt\tbackend/backend.go:197\tOpening db file (/tmp/TestEtctlutlMigrateMigrate_v3.5_to_v3.5_is_no-op2858091090/002/member-0/member/snap/db) with mode 2d72772d2d2d2d2d2d2d and with options: {Timeout: 0s, NoGrowSync: false, NoFreelistSync: true, PreLoadFreelist: false, FreelistType: , ReadOnly: false, MmapFlags: 8000, InitialMmapSize: 10737418240, PageSize: 0, NoSync: false, OpenFile: 0x0, Mlock: false, Logger: 0xc000130238}\r\n2024-05-27T07:18:46Z\tinfo\tbbolt\tbbolt@v1.4.0-alpha.1/db.go:321\tOpening bbolt db (/tmp/TestEtctlutlMigrateMigrate_v3.5_to_v3.5_is_no-op2858091090/002/member-0/member/snap/db) successfully\r\nError: failed to read wal: wal: snapshot mismatch\r\n (expected \"storage version up-to-date\\t{\\\"storage-version\\\": \\\"3.5\\\"}\", got []). Try EXPECT_DEBUG=TRUE" does not contain "storage version up-to-date\t{\"storage-version\": \"3.5\"}"
```
The error shown:
```
wal: snapshot mismatch
```
is returned only when indices match but Terms don't afaict, this is the error type:
https://github.com/etcd-io/etcd/blob/bb701b9265f31d61db5906325e0a7e2abf7d3627/server/storage/wal/wal.go#L60

and that comes from here because we end up calling `ReadAll` as part of `ReadWALVersion`:
https://github.com/etcd-io/etcd/blob/bb701b9265f31d61db5906325e0a7e2abf7d3627/server/storage/wal/wal.go#L496-L499

`w.start` here is ultimately the snapshot object that we pass into `OpenForRead`:
https://github.com/etcd-io/etcd/blob/bb701b9265f31d61db5906325e0a7e2abf7d3627/server/storage/wal/wal.go#L353

I think we should be passing in the `Term` here as well as part of `walpb.Snapshot`

fykaa: Thank you for pointing out the issue with the snapshot mismatch in the `wal` package.

I've made some adjustments to implement your suggestion. The `getLastSnapshot` function now returns the `walpb.Snapshot` object, which includes full snapshot information (the `Index` and `Term`). This snapshot object is then passed directly to `wal.OpenForRead`. I believe this should help mitigate the `ErrSnapshotMismatch` issue.

siyuanfoundation:
if you print out the WAL version after 
```
c.walVersion, err = wal.ReadWALVersion(w)
	if err != nil {
		return nil, fmt.Errorf(`failed to read wal: %v`, err)
	}
	fmt.Printf("got walVersion = %v\n", c.walVersion.MinimalEtcdVersion().String())
```
for the test `Downgrade v3.6 to v3.5 with force should work`, the main branch `walVersion=3.6.0`, with this change `walVersion=3.0.0`. So it is not reading the correct snapshot. The test got stuck in `e2e.SpawnWithExpect` forever because it could not find the expected log message.

serathius:
ping any updates?

fykaa:
Sorry for not being able to respond earlier. My focus has shifted in the meantime, and as the related issue is already resolved, I’m closing this now

PR #18017, opened 2024-05-15
[release-3.5] Setup a way to consistently manage go versions across scripts and go.mods

MadhavJivrajani:
Replicates https://github.com/etcd-io/etcd/pull/17876

This PR does the following:

scripts: default to using .go-version's version for tests and builds
Additionally, provide ability to opt-out of the .go-version and use a
custom one via env vars: FORCE_HOST_GO and GO_VERSION.

.*: sync go toolchain version and add ability to verify versions
This commit adds a script to sync the version present in .go-version
across all go.mod files as the toolchain directive. As part of that,
this commit also modifies go.mod files that did not have synced toolchain
directives.

Additionally, this also adds a script to verify all toolchain and go
directives against the version present in .go-version as follows:
(1) The go directive <= version in .go-version
(2) The toolchain directive == version in .go-version

The make target for this script is run as part of make verify which will automatically be run as a presubmit.

/assign @ahrtr

### DIFF
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -1,0 +1,32 @@
+---
+name: Static Analysis
+on: [push, pull_request]
+permissions: read-all
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b # v4.1.5
+      - id: goversion
+        run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+      - uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+        with:
+          go-version: ${{ steps.goversion.outputs.goversion }}
+      - run: |
+          set -euo pipefail
+
+          make verify
+      - run: |
+          set -euo pipefail
+
+          make fix
+
+          DIFF=$(git status --porcelain)
+
+          if [ -n "$DIFF" ]; then
+            echo "These files were modified:"
+            echo
+            echo "$DIFF"
+            echo
+            exit 1
+          fi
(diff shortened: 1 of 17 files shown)


serathius (approved):

Makefile
```diff
@@ -580,3 +580,17 @@ gofail-enable: install-gofail
...
+
+.PHONY: verify
+verify: verify-go-versions
+
+.PHONY: verify-go-versions
+verify-go-versions:
+	./scripts/verify_go_versions.sh
+
+.PHONY: fix
+fix: sync-toolchain-directive
```
ahrtr, line 592: In 3.5 (and probably 3.4), we didn't have Makefile targets `fix` and `verify`. Now you add them in this PR, but each of they has only one dependency. 

It means contributors need to run two commands when they bump dependencies for 3.5/3.4:
- ./scripts/fix.sh
- make fix

We should only run one command when bumping dependencies for 3.5/3.4,
 - either `make fix`: you need to execute ./scripts/fix.sh in target `fix`
 - or `./scripts/fix.sh`: you need to execute the `./scripts/sync_go_toolchain_directive.sh` inside `./scripts/fix.sh`
 
Refer to https://github.com/etcd-io/etcd/pull/17876#discussion_r1590724913

ahrtr, line 592:
```suggestion
fix: sync-toolchain-directive
     ./scripts/fix.sh
```

MadhavJivrajani:
@ahrtr can you PTAL? Thank you!

.github/workflows/static-analysis.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - id: goversion
+        run: echo "goversion=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+      - uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
+        with:
+          go-version: ${{ steps.goversion.outputs.goversion }}
+      - run: |
+          set -euo pipefail
+          make verify
+      - run: |
+          set -euo pipefail
```
jmhbnz, line 19: Nit: Line breaks in this file don't match one for `release-3.4` or `main`. Could we please make it the same so the files are easy to diff in future if needed.

MadhavJivrajani: Good catch @jmhbnz thanks, I probably missed something while copying over the content. 
Done!

Issue #16291, opened 2023-07-24
Integrate raft's new feature (async write) into etcd


ahrtr:
### What would you like to be added?

raft introduced a new feature [async write](https://github.com/etcd-io/raft/pull/8), etcd should try to integrate the new feature into etcd.

### Why is this needed?

Improve performance

yashkundu:
Hi @ahrtr, @jmhbnz  I really want to work on this feature and contribute, can you please assign this issue to me?

serathius:
Draft that I implemented some time ago https://github.com/etcd-io/etcd/pull/18027

mjlshen:
I took an initial stab at this in #22223 and have a couple questions:

1. Should this be behind a feature flag such as `RaftAsyncStorageWrites` (I'm leaning towards yes)?
2. I am not sure if I'm benchmarking properly. I can get `tools/benchmark` working against a three-node etcd cluster running in Docker Desktop on my Macbook, but that's probably not ideal. Is there recommended hardware to run the benchmark on? I want to make sure I'm providing relevant data when comparing performance.

serathius:
Ad 1, yes
Ad 2, I would want to try to develop https://perf-dash.k8s.io/#/?jobname=etcd&metriccategoryname=EtcdAPIPerformance&metricname=Benchmark%20&Operation=PUT to have the benchmark results publicly verifiable. To test it we would need to:
1. Tune the arguments passed in benchmark put to be more aggressive https://github.com/kubernetes/test-infra/blob/d9b63ee450b630ce64602d10303f6279729f42dc/config/jobs/etcd/etcd-benchmarks-periodic.yaml#L4
2. Copy the job to have optional presubmit that you could run on PR.
3. Consider if we need more benchmarks to cover all dimensions that async raft impacts.

PR #18027, opened 2024-05-19
Async raft

serathius:
Having fun rewriting the raft handling code in etcd to use raft async feature.

Ref https://github.com/etcd-io/etcd/issues/16291


serathius:
Draft for https://github.com/etcd-io/etcd/issues/16291

PR #18058, opened 2024-05-22
Complete migration of verify workflow to prow

jmhbnz:
After recently addressing some issues impacting our `pull-etcd-verify` presubmit job in prow I believe the job is now performing as expected and we are now ready to proceed with completing the prow migration for this workflow by deleting the workflow file. This is process we will need to gradually repeat for other workflows over time.

Refer the job summary in prow & testgrid:
- https://testgrid.k8s.io/sig-etcd-presubmits#pull-etcd-verify
- https://prow.k8s.io/?repo=etcd-io%2Fetcd&job=pull-etcd-verify

Configuration for the prow job is here: https://github.com/kubernetes/test-infra/blob/master/config/jobs/etcd/etcd-presubmits.yaml#L53-L87

Note - As the first workflow to be fully migrated to prow this pr creates a new dependence on our use of `/ok-to-test`. I have updated some documentation to make this requirement clear but ultimately all @etcd-io/maintainers-etcd need to be vigilant that prow jobs have been triggered before merging a pr.

Progress towards https://github.com/kubernetes/k8s.io/issues/6102


ahrtr (approved):

serathius (approved):

ivanvc (approved):
LGTM. We identified that the only check that's not working on the Prow infra is `marker`. I'll open an issue to find an alternative to it soon.

PR #18067, opened 2024-05-25
feat: add etcd_debugging_lease_total gauge metric

jimmy-bro:
The feature was discussed in https://github.com/etcd-io/etcd/issues/17874 .


server/lease/metrics.go
```diff
@@ -49,11 +49,19 @@ var (
 			// 1 second -> 3 months
 			Buckets: prometheus.ExponentialBuckets(1, 2, 24),
 		})
+
+	leaseTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "total",
```
serathius, line 56:
```suggestion
		Name:      "active",
```
Looking at the naming scheme above we have "granted_total", "revoked_total", "renewed_total", so if in the Help you called them "active leases" then let's put it in the name of the metric.

In prometheus conventions https://prometheus.io/docs/practices/naming/, suffix total is reserved for accumulating metrics like counters, for gauges we should skip it.

chaochn47:
@jimmy-bro could you please rebase and trigger the test again?

jimmy-bro:
Thanks for your mentiong @chaochn47 ,  I've already rebased the PR on git

jmhbnz (commented):
Thanks for raising this @jimmy-bro - Overall it's looking good but can you please add some tests so we would be able to merge it?

Refer to https://github.com/etcd-io/etcd/pull/17983 for examples of unit and integration metrics tests.

PR #18083, opened 2024-05-28
Add build subtargets based on ARCH

ArkaSaha30:
This PR will group similar targets as per architecture into subtargets to build. It also adds a `build-all` target to build all OS-ARCH combinations.

Conversation: https://github.com/kubernetes/test-infra/pull/32672#discussion_r1616735510
Fixes: https://github.com/kubernetes/test-infra/pull/32672

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,18 @@ include tests/robustness/makefile.mk
 build:
 	GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh
 
+PLATFORMS=linux-amd64 linux-386 linux-arm linux-arm64 linux-ppc64le linux-s390x darwin-amd64 darwin-arm64 windows-amd64 windows-arm64
+
+.PHONY: build-all
+build-all:
+	@for platform in $(PLATFORMS); do \
+		$(MAKE) build-$${platform}; \
+	done
+
+.PHONY: build-%
+build-%:
+	GOOS=$$(echo $* | cut -d- -f 1) GOARCH=$$(echo $* | cut -d- -f 2) GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh
+
 .PHONY: tools
 tools:
 	GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build_tools.sh


scripts/build_subtargets.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
jmhbnz, line 1: Keen to check consensus/views on separate script vs just updating `build.sh` and having the one script. What were your thoughts on this @ArkaSaha30?

My initial reaction would be just to add this logic into `build.sh` so only one script is involved but keen to hear what others think and happy to go with overall consensus.

ArkaSaha30: > My initial reaction would be just to add this logic into build.sh so only one script is involved

I agree on this, I was wondering the same but thought of keeping the `build.sh` untouched since it is an alpha/beta level change 🤔

scripts/build_subtargets.sh
```diff
@@ -0,0 +1,45 @@
...
+  *)
+    echo "Failed to find subtarget"
+    exit 2
+    ;;
+esac
+
+for subtarget in "${build_subtargets[@]}"
+do
+  GOOS="${subtarget%-*}" GOARCH="${subtarget#*-}" GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh
+done
```
henrybear327, line 45: nit: have a newline at the end of the file :)

ahrtr:
We have two options,
- [**preferred**] add Makefile targets `build-limux-amd64`, `build-linux-arm64`, etc. and `build-all`. We don't need to add or change any script files.
```
build-linux-amd64:
    GOOS=linux GOARCH=amd64 GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh

......

build-all: build-limux-amd64 build-linux-arm64 ......

```
- This PR

Makefile
```diff
@@ -6,6 +6,47 @@ include tests/robustness/makefile.mk
...
+	GOOS=linux GOARCH=ppc64le GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh
+
+.PHONY: build-linux-s390x
+build-linux-s390x:
+	GOOS=linux GOARCH=s390x GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh
+
+
+.PHONY: build-all
+build-all: build-linux-amd64 build-linux-386 build-darwin-amd64 build-darwin-arm64 build-windows-amd64 \
+		build-linux-arm build-linux-arm64 build-linux-ppc64le build-linux-s390x
```
serathius, line 48: PLATFORMS=linux-amd64 linux-386 linux-arm linux-arm64 linux-ppc64le linux-s390x darwin-amd64 darwin-arm64 windows-amd64 windows-arm64

.PHONY: build-all
build-all:
	@for platform in $(PLATFORMS); do \
		$(MAKE) build-$${platform}; \
	done

.PHONY: build-%
build-%:
	GOOS=$$(echo $* | cut -d- -f 1) GOARCH=$$(echo $* | cut -d- -f 2) GO_BUILD_FLAGS="${GO_BUILD_FLAGS} -v -mod=readonly" ./scripts/build.sh

BenTheElder: The approach with explicit targets could use make -j,

the base approach should probably consider something like https://github.com/kubernetes-sigs/kind/blob/bbcc02f84c8efd68d20c3a6585c2f282d7c1b8c9/hack/release/build/cross.sh#L37 to get parallelism

ArkaSaha30: I am not sure if we want parallelism...context: https://github.com/kubernetes/test-infra/pull/32672#discussion_r1616735750

BenTheElder: Serial can be a special case of setting N=1, but local users will probably want an option and I think we'll find this is a lot slower. The matrix on github actions would've been parallel and the prow equivilant would be creating a CI job for each.

jmhbnz (approved):
LGTM - Thanks @serathius and @ArkaSaha30.

ivanvc:
Can we have an agreement on https://github.com/kubernetes/test-infra/pull/32672#issuecomment-2135840239 + https://github.com/etcd-io/etcd/pull/18083#discussion_r1617546349 before merging this PR? :sweat_smile:

Makefile
ArkaSaha30: ^^ I agree, it makes sense to have parallelism. 
WDYT @jmhbnz @ahrtr @serathius

Makefile
```diff
@@ -2,10 +2,22 @@
 all: build
 include tests/robustness/makefile.mk
 
+PLATFORMS=linux-amd64 linux-386 linux-arm linux-arm64 linux-ppc64le linux-s390x darwin-amd64 darwin-arm64 windows-amd64 windows-arm64
```
ahrtr, line 5: nit: please put `PLATFORMS` right above the target `build-all`

Makefile
ahrtr: Overall not a big deal, suggest to follow the `for loop` solution
- building with a loop is much simpler, so it has a little better readability.
- Serial building just takes ~2 minutes, which is far less than other workflow check; so it has no impact on the overall workflow time.

Makefile
ArkaSaha30: Sure, thank you 👍

ahrtr (approved):

serathius (approved):

Issue #18085, opened 2024-05-28
Memory leak


TheRealMal:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

file is being opened here
https://github.com/etcd-io/etcd/blob/9914047acb5c679531f644d9c81718fcd85a3108/client/v3/snapshot/v3_snapshot.go#L63

but being closed only here
https://github.com/etcd-io/etcd/blob/9914047acb5c679531f644d9c81718fcd85a3108/client/v3/snapshot/v3_snapshot.go#L87

Between these lines errors can be occured and file won't be closed

### What did you expect to happen?

There should close function invocation via defer

### How can we reproduce it (as minimally and precisely as possible)?

Check code

### Anything else we need to know?

-

### Etcd version (please run commands below)

latest

### Etcd configuration (command line flags or environment variables)

-

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

-

### Relevant log output

```Shell
-
```

vivekpatani:
@ahrtr @TheRealMal https://github.com/etcd-io/etcd/pull/18200 - does this make sense?

ahrtr:
Resolved in https://github.com/etcd-io/etcd/pull/19093

We need to backport the fix to 3.5 as well.

fuweid:
https://github.com/golang/go/issues/41505

The go runtime has GC on that leak file descriptor. it could be closed on next GC cycle.
just wondering that why it failed on Close.

Issue #18084, opened 2024-05-28
Missing nil check


TheRealMal:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

There is no nil check here
https://github.com/etcd-io/etcd/blob/9914047acb5c679531f644d9c81718fcd85a3108/client/pkg/transport/listener.go#L316

That can lead to dereference of nil variable

### What did you expect to happen?

There should be `if info.Logger != nil` check like everywhere

### How can we reproduce it (as minimally and precisely as possible)?

Just check code

### Anything else we need to know?

-

### Etcd version (please run commands below)

latest

### Etcd configuration (command line flags or environment variables)

-

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

-

ahrtr:
It doesn't need the `nil` check; it requires the callers must pass in a valid `lg`.
https://github.com/etcd-io/etcd/blob/9914047acb5c679531f644d9c81718fcd85a3108/client/pkg/transport/listener.go#L210

We can remove all nil check in that function, please feel free to deliver a PR. thx

PR #18086, opened 2024-05-28
Bugs fixes

TheRealMal:
This PR fixes 2 bugs:

Fixes etcd-io/etcd#18085 `Memory leak`
Fixes etcd-io/etcd#18084 `Dereference of nil`

### DIFF
--- a/client/v3/snapshot/v3_snapshot.go
+++ b/client/v3/snapshot/v3_snapshot.go
@@ -64,6 +64,7 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 	if err != nil {
 		return "", fmt.Errorf("could not open %s (%v)", partpath, err)
 	}
+	defer f.Close()
 	lg.Info("created temporary db file", zap.String("path", partpath))
 
 	start := time.Now()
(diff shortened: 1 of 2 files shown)


client/v3/snapshot/v3_snapshot.go
```diff
@@ -64,6 +64,7 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 	if err != nil {
 		return "", fmt.Errorf("could not open %s (%v)", partpath, err)
 	}
+	defer f.Close()
```
tjungblu, line 67: it sounds like an easy fix, but how would the rename operation below work on an open file?

jmhbnz (commented):
Hi @TheRealMal  - Thanks for raising these fixes. It looks like your two commits are not signed. Can you please take a look and ensure all commits are signed so the developer certificate of origin check passes? i.e.

```bash
git rebase --signoff HEAD~2
```

client/v3/snapshot/v3_snapshot.go
henrybear327: Joining the errors should work :) 

Ref: `https://pkg.go.dev/errors#Join`

jmhbnz:
Discussed during sig-etcd triage meeting - hey @TheRealMal do you have capacity to continue this pr?

PR #18103, opened 2024-05-30
Compaction hash

serathius:
Found a bug in compaction hash consistency check when there are multiple concurrent compaction it discards  hashes from them.

Need to add a test.


chaochn47:
> Found a bug in compaction hash consistency check when there are multiple concurrent compaction it discards hashes from them.

Maybe I lack context on this, but isn't compaction is schedued in the store fifo queue so there are no concurrent compactions in theory. 

https://github.com/etcd-io/etcd/blob/eb9bfaa983554748841bc06fd1f6ae14e87d99e4/server/mvcc/kvstore.go#L263

Is it a correct understanding?

serathius:
> Is it a correct understanding?

From backend perspective yes, I meant from client perspective they can schedule multiple requests. They will be put into job queue and executed consecutively.

PR #18107, opened 2024-06-01
Improve e2e error reporting

serathius:
cc @fuweid @jmhbnz @MadhavJivrajani @siyuanfoundation

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -193,7 +193,7 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


pkg/expect/expect.go
```diff
@@ -193,7 +193,7 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 
 		select {
 		case <-ctx.Done():
-			return "", fmt.Errorf("failed to find match string: %w", ctx.Err())
+			return "", fmt.Errorf("context done before to found matching log")
```
jmhbnz, line 196: Sentence is a bit hard to understand, perhaps:
```suggestion
			return "", fmt.Errorf("context done before matching log found")
```

serathius:
Fixed the comments

Issue #18131, opened 2024-06-05
Worrying trend with test coverage


serathius:
### What would you like to be added?

Code coverage value doesn't imply muchy, however we can infer some information from its trends.

Within last 12 months the etcd code coverage fell by 6%. 
![image](https://github.com/etcd-io/etcd/assets/5873433/243ead87-88df-4cf4-bb1b-eb9c0128bb1f)

Link: https://app.codecov.io/gh/etcd-io/etcd/tree/main?search=&displayType=tree&trend=12%20months

I encourage all contributors, reviewers and approvers to ensure that we not continue this trend. As described in https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/features.md all features should have a proper testing. We should ensure that all PRs, both features and bugfixes, come with basic set of unit test that ensure that feature/bugfix does what it claims to be.

If you can please recall any PRs that you have reviewed and would benefit from more testing. We should ensure there are a proper followups to add those tests.

cc @jmhbnz @ahrtr @wenjiaswe 

### Why is this needed?

Testing is the foundation of project maintainability. Without them, we cannot add any features or fix any bugs without knowing that something else will break

ahrtr:
I recall we have a coverage workflow running for each PR, but got removed for some reasons I don't remember anymore. Can we start with adding it back?

henrybear327:
/assign @henrybear327 
I can give this a try!

serathius:
Don't know if this would be visible on a single PR bases. I don't like a per PR information about coverage because it encourages a bad habit of writing low quality tests just for coverage sake. 

We should think holistically which areas are important to keep high coverage, which might not, and encourage contributors to add tests where and when they are needed.

ahrtr:
> and I don't like per-PR coverage information because it encourages writing low-quality tests just for coverage

In my opinion, this is a viewpoint with obvious personal subjective bias. Adding a coverage workflow is a ver feasible way to keep the test coverage.  With respect to "low quality tests", it's the scope of review.

jmhbnz:
If we are concerned pr's are not hitting requirements for testing and coverage is reducing I completely agree with @ahrtr that the way to address this would be to surface coverage information directly within pull requests.

It's incumbent on reviewers of pull requests to provide feedback on the quality of tests written and ensure tests add value before hitting merge.

wenjiaswe:
Thanks @serathius for opening this issue! 6% decrease is not a good trend, let's avoid that.

[...]

PR #18143, opened 2024-06-07
Update codecov config to support commenting in PR

henrybear327:
- add missing directory fixing go.etcd.io/etcd/etcdutl/v3/::etcdutl/
- update codecov config to support commenting in PR

Reference: 
- https://github.com/etcd-io/etcd/issues/18131
- https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,9 +1,11 @@
 ---
 name: Coverage
-on: [push]
+on: [push, pull_request]
 permissions: read-all
 jobs:
   coverage:
+    # this is to prevent the job to run at forked projects
+    if: github.repository == 'etcd-io/etcd'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,12 +1,17 @@
 ---
+# https://docs.codecov.com/docs/codecovyml-reference
 codecov:
   token: 6040de41-c073-4d6f-bbf8-d89256ef31e1
   disable_default_path_fixes: true
+  require_ci_to_pass: false
+  notify:
+    wait_for_ci: false
 fixes:
   - go.etcd.io/etcd/api/v3/::api/
   - go.etcd.io/etcd/client/v3/::client/v3/
   - go.etcd.io/etcd/client/v2/::client/v2/
   - go.etcd.io/etcd/etcdctl/v3/::etcdctl/
+  - go.etcd.io/etcd/etcdutl/v3/::etcdutl/
   - go.etcd.io/etcd/pkg/v3/::pkg/
   - go.etcd.io/etcd/server/v3/::server/
 ignore:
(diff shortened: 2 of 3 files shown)


henrybear327:
@serathius I was checking [other project's codecov configuration file](https://github.com/kubernetes/dashboard/blob/f5bd92216e3783aeab0d266ae973c655c30ab0f6/.github/workflows/ci.yml#L139) and [the official documentation](https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage#github-actions), we should consider putting codecov token in Github Secret (maybe we need to regenerate a new token first, but we need to have access to the codecov account, which I don't know who to ping). 

I can't do this on my own since I don't have Github Repo Setting access, so this part I might need your help to look into it!

Thanks

jmhbnz (approved):
LGTM - Thanks @henrybear327, the preview above in this pr is looking good.

henrybear327:
> LGTM - Thanks @henrybear327, the preview above in this pr is looking good.

Thanks for reviewing @jmhbnz! :) 

Currently, the Codecov comment is set to update the existing code coverage comment when a new commit is pushed (default behavior), instead of deleting the existing comment and creating a new one. 

Using the default behavior means that the code coverage information will likely be at the top of the comment, so PRs with long discussion threads will require some scrolling to view the code coverage information. Deleting the old one and adding a new comment with each commit will generate more email notifications though.

In order to avoid email spamming, I think it's better that we use the default behavior. 

What do you think? :)

ahrtr:
> see 20 files with indirect coverage changes

Thanks @henrybear327 for the PR. But I am curious why the coverage decreased for many go files even this PR doesn't change any go source code?

jmhbnz:
> In order to avoid email spamming, I think it's better that we use the default behavior.

Agreed. Let's keep it to one coverage comment per pr that is always up to date.

henrybear327:
> > see 20 files with indirect coverage changes
> 
> Thanks @henrybear327 for the PR. But I am curious why the coverage decreased for many go files even this PR doesn't change any go source code?

@ahrtr I suspect it's due to rebasing. I will look into this!

henrybear327:
> > > see 20 files with indirect coverage changes
> > 
> > 
> > Thanks @henrybear327 for the PR. But I am curious why the coverage decreased for many go files even this PR doesn't change any go source code?
> 
> @ahrtr I suspect it's due to rebasing. I will look into this!

This is due to [indirect changes](https://app.codecov.io/gh/etcd-io/etcd/pull/18143/indirect-changes) [1]. 

I suspect that during test coverage execution, we have some of the tests written in a way that is not deterministic. Because when I check the "indirect changes" of PRs from this [list](https://app.codecov.io/gh/etcd-io/etcd/pulls), hit/miss differences come from things like select cases (e.g. error handling, timeout, etc.), cache eviction, etc. 

Screenshot 2024-06-12 at 8 07 18 AM
Screenshot 2024-06-12 at 8 07 00 AM

So our test coverage percentage will have a small variation across runs, as you can see [here](https://app.codecov.io/gh/etcd-io/etcd/commits?branch=main).

Do you think this make sense @ahrtr?

Reference:
[1] https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes

ahrtr:
> [1] https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes

Thanks for the clarification, which basically makes sense. Can you pls add a comment to include the link for reference?

henrybear327:
> > [1] https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes
> 
> Thanks for the clarification, which basically makes sense. Can you pls add a comment to include the link for reference?

Thanks @ahrtr! I have
- updated the PR description 
- updated the commit message
- added link to the discussion comment in the issue

Not sure if this is what you expected :)

henrybear327:
> https://docs.codecov.com/docs/unexpected-coverage-changes#reasons-for-indirect-changes

No worries :) 

Done as requested! Thanks

ahrtr (approved):
LGTM

Thanks @henrybear327

PR #18158, opened 2024-06-11
tests/robustness: add in compaction checks for prevKV

MadhavJivrajani:
Kubernetes issues watch to etcd using PrevKV() and with that, it relies on the property that create events always have prevKV=nil. This commit incorporates that explicit check in the robustness tests.

Further, etcd also can have prevKV=nil for any event type in the case of compaction taking place. This commit also adds in that check to ensure we are not changing these invariants.

We can make this change now because compaction support exists in robustness tests: https://github.com/etcd-io/etcd/pull/17833

Towards: https://github.com/etcd-io/etcd/issues/15820 

/assign @serathius @siyuanfoundation

### DIFF
--- a/tests/robustness/validate/watch.go
+++ b/tests/robustness/validate/watch.go
@@ -245,26 +245,31 @@ func validatePrevKV(lg *zap.Logger, replay *model.EtcdReplay, report report.Clie
 		}
 		for _, resp := range op.Responses {
 			for _, event := range resp.Events {
-				// Get state state just before the current event.
-				state, err2 := replay.StateForRevision(event.Revision - 1)
+				// Get state just before the current event.
+				prevState, err2 := replay.StateForRevision(event.Revision - 1)
 				if err2 != nil {
 					panic(err2)
 				}
-				// TODO(MadhavJivrajani): check if compaction has been run as part
-				// of failpoint injection. If compaction has run, prevKV can be nil
-				// even if it is not a create event.
-				//
-				// Considering that Kubernetes opens watches to etcd using WithPrevKV()
-				// option, ideally we would want to explicitly check the condition that
-				// Kubernetes does while parsing events received from etcd:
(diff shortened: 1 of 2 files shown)


MadhavJivrajani:
oh wait I see my error - will fix in a little bit.

serathius:
Please ping me when it's ready for review.

tests/robustness/validate/watch.go
```diff
@@ -245,26 +245,31 @@ func validatePrevKV(lg *zap.Logger, replay *model.EtcdReplay, report report.Clie
...
-				// Kubernetes does while parsing events received from etcd:
+
+				// Considering that Kubernetes opens watches to etcd using WithPrevKV() option, we explicitly
+				// check the condition that Kubernetes does while parsing events received from etcd:
 				// https://github.com/kubernetes/kubernetes/blob/a9e4f5b7862e84c4152eabe2e960f3f6fb9a4867/staging/src/k8s.io/apiserver/pkg/storage/etcd3/event.go#L59
-				// i.e. prevKV is nil iff the event is a create event, we cannot reliably
-				// check that without knowing if compaction has run.
+				// i.e. for create events, prevKV is nil.
+				if event.IsCreate && event.PrevValue != nil {
+					lg.Error("prevValue field shold be nil in create event", zap.Int("client", report.ClientID), zap.Any("event", event), zap.Any("previousValue", prevState.KeyValues[event.Key]))
```
serathius, line 259:
```suggestion
					lg.Error("prevValue field should be nil in create event", zap.Int("client", report.ClientID), zap.Any("event", event), zap.Any("previousValue", prevState.KeyValues[event.Key]))
```

tests/robustness/validate/watch.go
```diff
@@ -245,26 +245,31 @@ func validatePrevKV(lg *zap.Logger, replay *model.EtcdReplay, report report.Clie
...
+					err = errBrokePrevKV
+				}
+
+				// If compaction took place in the previous state, prevKV should be nil for all types of events in the current state:
```
serathius, line 263: This is not correct. Event generation and compaction are independent processes. They are not serialized by WAL. 

Your idea is correct, but it cannot be implemented this way.

PR #18162, opened 2024-06-12
Always keep the compact revision, and delay compact tombstone revision

ahrtr:
Always keep the compact revision, no matter it's a tomestone revision or not. If it's a tombstone revision, then cleanup it in next compact operation.

PoC of option 2 in https://docs.google.com/document/d/1APJE38DxENsRLLVapRz1CQ6UD4-uYaFutO12Y01VcNw/edit

cc @fuweid 

### DIFF
--- a/server/storage/mvcc/key_index.go
+++ b/server/storage/mvcc/key_index.go
@@ -233,7 +233,14 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
 		return
 	}
 
-	ki.doCompact(atRev, available)
+	genIdx, revIndex := ki.doCompact(atRev, available)
+	g := &ki.generations[genIdx]
+	if !g.isEmpty() {
+		// remove any tombstone
+		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
+			delete(available, g.revs[revIndex])
+		}
+	}
 }
 
 func (ki *keyIndex) doCompact(atRev int64, available map[Revision]struct{}) (genIdx int, revIndex int) {
--- a/server/storage/mvcc/key_index.go
+++ b/server/storage/mvcc/key_index.go
@@ -221,11 +221,6 @@ func (ki *keyIndex) compact(lg *zap.Logger, atRev int64, available map[Revision]
 		if revIndex != -1 {
 			g.revs = g.revs[revIndex:]
 		}
-		// remove any tombstone
-		if len(g.revs) == 1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[0])
-			genIdx++
-		}
 	}
 
 	// remove the previous generations.
@@ -238,14 +233,7 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
 		return
 	}
 
-	genIdx, revIndex := ki.doCompact(atRev, available)
-	g := &ki.generations[genIdx]
-	if !g.isEmpty() {
-		// remove any tombstone
-		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[revIndex])
-		}
-	}
+	ki.doCompact(atRev, available)
 }
 
 func (ki *keyIndex) doCompact(atRev int64, available map[Revision]struct{}) (genIdx int, revIndex int) {
@@ -262,8 +250,15 @@ func (ki *keyIndex) doCompact(atRev int64, available map[Revision]struct{}) (gen
 	genIdx, g := 0, &ki.generations[0]
 	// find first generation includes atRev or created after atRev
 	for genIdx < len(ki.generations)-1 {
-		if tomb := g.revs[len(g.revs)-1].Main; tomb > atRev {
-			break
+		// if len(g.revs) == 1, then it means that the generation has
+		// only one tombstone revision, which was kept in last compact
+		// operation. We need to clean it up in this round of compact
+		// operation. Refer to discussion in
+		// https://github.com/etcd-io/etcd/issues/18089.
+		if len(g.revs) > 1 {
+			if tomb := g.revs[len(g.revs)-1].Main; tomb >= atRev {
+				break
+			}
 		}
 		genIdx++
 		g = &ki.generations[genIdx]


ahrtr:
@fuweid Please feel free to continue to work on the issue https://github.com/etcd-io/etcd/issues/18089 on top of this PR if it makes sense to you.  Pls feel free to ping me on slack if you want discussion.

server/storage/mvcc/key_index.go
```diff
@@ -238,14 +233,7 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
...
 
-	genIdx, revIndex := ki.doCompact(atRev, available)
-	g := &ki.generations[genIdx]
-	if !g.isEmpty() {
-		// remove any tombstone
-		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[revIndex])
-		}
-	}
+	ki.doCompact(atRev, available)
```
ahrtr, line 236: This change may impact the `HashByRev`. We may want to revert this change.

ahrtr:
These are all the test cases I can think of, please feel free to add more. Most likely some of them have already been covered by existing test cases.

I also added the test cases in the [doc](https://docs.google.com/document/d/1APJE38DxENsRLLVapRz1CQ6UD4-uYaFutO12Y01VcNw/edit) (see option 2)

cc @fuweid @serathius 

# Test cases

## unit test on keyIndex
**case 1**: compact a tombstone revision X, it shouldn't be compacted. Next compaction should cleanup the revision X for both cases below,
- 1.1 The revision X is the last revision for the key
- 1.2 The revision X isn't the last revision for the key

**case 2**: compact a normal revision X, it shouldn't be compacted. In next compaction,
- 2.1 If the revision X is the last revision for the key, it shouldn't be compacted.
- 2.2 If the revision X isn't the last revision for the key, it should be compacted. 


## e2e test on watch
**case 1**: compact tomestone revisions shouldn't impact watch
    rough steps:
    1) start a watch client to watch on a key
    2) start a goroutine to continuous put or delete a key
    3) perform compaction against the different tombstone revisions multiple times, verify that no revisions are dropped.    

**case 2**: compact normal revisions shouldn't impact watch
    similar steps as above, but we only execute put operations

**case 3**: a new watch starting on a compacted tombstone revision should receive the delete event
    Refer to the reproduction steps in the issue


## e2e test on hash values
**case 1**: All members should have the same hash value up to the compacted revision after compacting a tombstone revision
    rough steps:
    1) start a 3 member cluster;
    2) execute some put and delete operations;
    3) execute compaction on revision X;
    4) get hashKV up to revision X from all members, verify the hash values are consistent

- 1.1 all members have the same etcd version
- 1.2 mix versions, i.e. main vs 3.5.14 or release-3.5 vs 3.5.14

**case 2**: all member should have the same hash value up to the compacted revision after compacting a non-tombstone revision
    similar steps as above, but execute compact on a non-tombstone revision    

- 2.1 all members have the same etcd version
- 2.2 mix versions, i.e. main vs 3.5.14 or release-3.5 vs 3.5.14

**case 3**: All members should have the same hash value up to a non-compacted revision after compacting a tombstone revision
    similar steps as case 1, but get hashKv up to a non-compacted revision

- 3.1 all members have the same etcd version
- 3.2 mix versions, i.e. main vs 3.5.14 or release-3.5 vs 3.5.14

**case 4**: All members should have the same hash value up to a non-compacted revision after compacting a non-tombstone revision
    similar steps as case 2, but get hashKv up to a non-compacted revision

-  4.1 all members have the same etcd version
-  4.2 mix versions, i.e. main vs 3.5.14 or release-3.5 vs 3.5.14

## robustness test
Support compaction, and verify the impact on hash values and watch.

ahrtr:
@fuweid are you still working on https://github.com/etcd-io/etcd/issues/18089 based on this PR?

serathius:
Can we start implementing those tests? 

The bigger problem then the bug existing is fact that we don't have those tests already. Even before we have a fix, we can just add the test to codify the current behavior. It should give us better visibility into impact of when we implement the fix.

serathius:
cc @MadhavJivrajani @siyuanfoundation @fuweid @jmhbnz @henrybear327 @ah8ad3

server/storage/mvcc/key_index.go
```diff
@@ -221,11 +221,6 @@ func (ki *keyIndex) compact(lg *zap.Logger, atRev int64, available map[Revision]
 		if revIndex != -1 {
 			g.revs = g.revs[revIndex:]
 		}
-		// remove any tombstone
-		if len(g.revs) == 1 && genIdx != len(ki.generations)-1 {
```
fuweid, line 225: Hi @ahrtr I'm still working on this.

I handle this in a different way. However, I still need more time to add more test cases for that.

```diff
diff --git a/server/storage/mvcc/key_index.go b/server/storage/mvcc/key_index.go
index e19068f9c..e8223380e 100644
--- a/server/storage/mvcc/key_index.go
+++ b/server/storage/mvcc/key_index.go
@@ -223,8 +223,10 @@ func (ki *keyIndex) compact(lg *zap.Logger, atRev int64, available map[Revision]
 		}
 		// remove any tombstone
 		if len(g.revs) == 1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[0])
-			genIdx++
+			if g.revs[0].Main < atRev {
+				delete(available, g.revs[0])
+				genIdx++
+			}
 		}
 	}

@@ -243,7 +245,9 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
 	if !g.isEmpty() {
 		// remove any tombstone
 		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[revIndex])
+			if g.revs[revIndex].Main < atRev {
+				delete(available, g.revs[revIndex])
+			}
 		}
 	}
 }
```

ahrtr: > ```diff
> @@ -223,8 +223,10 @@ func (ki *keyIndex) compact(lg *zap.Logger, atRev int64, available map[Revision]
>  		}
>  		// remove any tombstone
>  		if len(g.revs) == 1 && genIdx != len(ki.generations)-1 {
> -			delete(available, g.revs[0])
> -			genIdx++
> +			if g.revs[0].Main < atRev {
> +				delete(available, g.revs[0])
> +				genIdx++
> +			}
>  		}
>  	}
> ```

It doesn't make sense, because `if g.revs[0].Main < atRev` will be always `false`, so it has the same effect as my change, which is much simpler.

fuweid: Thanks for pointing it out. Yes, when there is only one record and it's not the last generation, we should always keep it.

server/storage/mvcc/key_index.go
```diff
@@ -262,8 +257,15 @@ func (ki *keyIndex) doCompact(atRev int64, available map[Revision]struct{}) (gen
...
 	// find first generation includes atRev or created after atRev
 	for genIdx < len(ki.generations)-1 {
-		if tomb := g.revs[len(g.revs)-1].Main; tomb > atRev {
-			break
+		// if len(g.revs) == 1, then it means that the generation has
+		// only one tombstone revision, which was kept in last compact
+		// operation. We need to clean it up in this round of compact
+		// operation. Refer to discussion in
+		// https://github.com/etcd-io/etcd/issues/18089.
+		if len(g.revs) > 1 {
```
fuweid, line 265: I think we don't need this if here. Just let it check the following condition. It's more easy to understand and reason

ahrtr: It seems right.  If `len(g.revs)==1` is `true`, it means that lastCompactionRev must be equal to the tombstone rev. In this round of compaction, the tombstone revision must be < compaction rev.

So we can revert this change.

ah8ad3:
> Can we start implementing those tests?

>The bigger problem then the bug existing is fact that we don't have those tests already. Even before we have a fix, we can just add the test to codify the current behavior. It should give us better visibility into impact of when we implement the fix.

>/cc @MadhavJivrajani @siyuanfoundation @fuweid @jmhbnz @henrybear327 @ah8ad3


I would like to start working on those test but i am a little bit lost of track if anybody started it or not is there any update?

PR #18169, opened 2024-06-13
Ensure the right leader is picked in watcher results for observe.

### LINKED ISSUE #18163: Election.Observe may report the wrong leader
### Bug report criteria

### What happened?

The implementation of observe does a get on the election key (prefix)
  https://github.com/etcd-io/etcd/blob/5c3190903fa1115b083e22fe55f86e3a78d82f36/client/v3/concurrency/election.go#L178
and if that get returns zero kvs, it does a watch:
  https://github.com/etcd-io/etcd/blob/5c3190903fa1115b083e22fe55f86e3a78d82f36/client/v3/concurrency/election.go#L190

The watch may return multiple PUT events, in which case the first one is returned (note the break statement in the if body matching a PUT event type):
  https://github.com/etcd-io/etcd/blob/5c3190903fa1115b083e22fe55f86e3a78d82f36/client/v3/concurrency/election.go#L201

It seems to me returning the first PUT event can be wrong if there are multiple PUTs in the returned `wr.Events`, unless I am misunderstanding this (which is possible, since there is a fair amount of complexity here). So please bear with me, and help me understand if this is not broken.

I think instead of returning the first one, the PUT event with the lowest create revision should be returned, otherwise we risk returning the wrong leader.  Returning the event with the lowest create revision would be consistent with what is done if the initial get is not empty.  When the initial get is not empty, the first KV in the get range is returned
  https://github.com/etcd-io/etcd/blob/5c3190903fa1115b083e22fe55f86e3a78d82f36/client/v3/concurrency/election.go#L210

since the get was done with options for ascending order of creation
  https://github.com/etcd-io/etcd/blob/5c3190903fa1115b083e22fe55f86e3a78d82f36/client/v3/concurrency/election.go#L178
  https://github.com/etcd-io/etcd/blob/5c3190903fa1115b083e22fe55f86e3a78d82f36/client/v3/op.go#L462

The only way I could see ignoring any other PUTs but the first not being an issue is if there was a guarantee that there will never be more than one PUT event in the returned `wr.Events` at that point in the code.  I don't believe that is guaranteed, since watches are eventual consistent and there can be a network issue between the first get and this subsequent watch; if at the time of the first get there is no election leader (I don't mean etcd cluster leader, I mean elected candidate in the election API), and then a transient network error occurs, a retry of the watch may succeed at a time when there are several candidates already registered for the election, which I believe would make the watch return multiple PUTs.

N/A

N/A

N/A

<details>
3.5.12
</details>

N/A

N/A

### Relevant log output

_No response_

jcferretti:
Fixes #18163

### DIFF
--- a/client/v3/concurrency/election.go
+++ b/client/v3/concurrency/election.go
@@ -73,7 +73,7 @@ func (e *Election) Campaign(ctx context.Context, val string) error {
 	k := fmt.Sprintf("%s%x", e.keyPrefix, s.Lease())
 	txn := client.Txn(ctx).If(v3.Compare(v3.CreateRevision(k), "=", 0))
 	txn = txn.Then(v3.OpPut(k, val, v3.WithLease(s.Lease())))
-	txn = txn.Else(v3.OpGet(k))
+	txn = txn.Else(v3.OpGet(k, v3.WithFirstCreate()...))
 	resp, err := txn.Commit()
 	if err != nil {
 		return err
@@ -195,13 +195,15 @@ func (e *Election) observe(ctx context.Context, ch chan<- v3.GetResponse) {
 					return
 				}
 				// only accept puts; a delete will make observe() spin
+				// find the smaller create revision among the puts
 				for _, ev := range wr.Events {
 					if ev.Type == mvccpb.PUT {
-						hdr, kv = &wr.Header, ev.Kv
-						// may have multiple revs; hdr.rev = the last rev
-						// set to kv's rev in case batch has multiple Puts
-						hdr.Revision = kv.ModRevision
-						break
+						if kv == nil || ev.Kv.CreateRevision < kv.CreateRevision {
+							hdr, kv = &wr.Header, ev.Kv
+							// may have multiple revs; hdr.rev = the last rev
+							// set to kv's rev in case batch has multiple Puts
+							hdr.Revision = kv.ModRevision
+						}
 					}
 				}
 			}
--- a/client/v3/concurrency/election.go
+++ b/client/v3/concurrency/election.go
@@ -73,7 +73,7 @@ func (e *Election) Campaign(ctx context.Context, val string) error {
 	k := fmt.Sprintf("%s%x", e.keyPrefix, s.Lease())
 	txn := client.Txn(ctx).If(v3.Compare(v3.CreateRevision(k), "=", 0))
 	txn = txn.Then(v3.OpPut(k, val, v3.WithLease(s.Lease())))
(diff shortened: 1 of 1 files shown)


ivanvc:
Hi @jcferretti, thanks for your contribution. Could you please sign your commit? So the developer certificate of origin (DCO) check passes, i.e:

```
git rebase HEAD~1 --signoff
git push --force
```

Thanks!

Ref: https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#commit-your-change

jcferretti:
> Hi @jcferretti, thanks for your contribution. Could you please sign your commit? So the developer certificate of origin (DCO) check passes, i.e:
> 
> ```
> git rebase HEAD~1 --signoff
> git push --force
> ```
> 
> Thanks!
> 
> Ref: https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#commit-your-change

Done, thank you, sorry about missing that.

ahrtr:
Please add a test to reproduce the issue which you are trying fix firstly.

jcferretti:
> Please add a test to reproduce the issue which you are trying to fix firstly.

Can you please comment on the issue report regarding the analysis there?  Thanks.
https://github.com/etcd-io/etcd/issues/18163

client/v3/concurrency/election.go
```diff
@@ -195,13 +195,15 @@ func (e *Election) observe(ctx context.Context, ch chan<- v3.GetResponse) {
...
-						// may have multiple revs; hdr.rev = the last rev
-						// set to kv's rev in case batch has multiple Puts
-						hdr.Revision = kv.ModRevision
-						break
+						if kv == nil || ev.Kv.CreateRevision < kv.CreateRevision {
+							hdr, kv = &wr.Header, ev.Kv
+							// may have multiple revs; hdr.rev = the last rev
+							// set to kv's rev in case batch has multiple Puts
+							hdr.Revision = kv.ModRevision
+						}
```
ahrtr, line 206: I do not see any issue here. 

- The `for` condition already guarantees that the `kv == nil`, 
https://github.com/etcd-io/etcd/blob/debc8fb00bd935249256e397fc0e001bc2e87abb/client/v3/concurrency/election.go#L191

- Also please read https://etcd.io/docs/v3.5/learning/api_guarantees/#watch-apis

jcferretti: The issue as I see it is not kv being null.  The issue as I see it is the possibility that the watch returns more than one PUT, and the current code ignores any other PUT aside from the first.

Can you please make explicit what specific watch guarantee ensures we can't get more than one PUT?

PR #18171, opened 2024-06-13
Fix govulncheck CI check on main branch

henrybear327:
This commit fixed the Go Vulnerability Checker CI job, which isn't scanning for all go.mod files within the project.

Reference:
- https://github.com/etcd-io/etcd/discussions/18168

### DIFF
--- a/.github/workflows/govuln.yaml
+++ b/.github/workflows/govuln.yaml
@@ -16,4 +16,6 @@ jobs:
       - run: |
           set -euo pipefail
 
-          go install golang.org/x/vuln/cmd/govulncheck@latest && govulncheck ./...
+          go install golang.org/x/vuln/cmd/govulncheck@latest
+
+          find -name go.mod -exec /bin/bash -c 'echo scanning $(dirname {}); govulncheck -C $(dirname {}) -show verbose ./...' \;


.github/workflows/govuln.yaml
```diff
@@ -16,4 +16,18 @@ jobs:
...
           set -euo pipefail
 
-          go install golang.org/x/vuln/cmd/govulncheck@latest && govulncheck ./...
+          go install golang.org/x/vuln/cmd/govulncheck@latest
+
+          # Find all go.mod files recursively, as we have multiple of them within the project
+          gomod_files=$(find . -name go.mod)
+
+          # Loop through each go.mod file path
+          for filepath in $gomod_files; do
```
ivanvc, line 25: Usually looping over `find` results is not the best idea, see: https://www.shellcheck.net/wiki/SC2044

An alternative would be to do:

```sh
find -name go.mod -exec /bin/bash -c 'echo scanning $(dirname {}); govulncheck -C $(dirname {}) -show verbose ./...' \;
```

ivanvc (commented):
Thanks, @henrybear327. This option looks good as a short-term solution. However, probably a more sturdy solution would be to add a new `Makefile` target, so we can also port this action into prow more smoothly.

.github/workflows/govuln.yaml
henrybear327: Thank you for the suggestion! :) Patched as suggested!

ivanvc (approved):
LGTM. Thanks, @henrybear327.

ahrtr (approved):
LGTM

Thanks

PR #18172, opened 2024-06-13
Fix govulncheck CI check on release-3.4

henrybear327:
This commit fixed the Go Vulnerability Checker CI job, which isn't scanning for all go.mod files within the project.

Reference:
- https://github.com/etcd-io/etcd/discussions/18168

### DIFF
--- a/.github/workflows/govuln.yaml
+++ b/.github/workflows/govuln.yaml
@@ -16,4 +16,6 @@ jobs:
       - run: |
           set -euo pipefail
 
-          go install golang.org/x/vuln/cmd/govulncheck@latest && govulncheck ./...
+          go install golang.org/x/vuln/cmd/govulncheck@latest
+
+          find -name go.mod -exec /bin/bash -c 'echo scanning $(dirname {}); govulncheck -C $(dirname {}) -show verbose ./...' \;


.github/workflows/govuln.yaml
```diff
@@ -16,4 +16,18 @@ jobs:
...
           set -euo pipefail
 
-          go install golang.org/x/vuln/cmd/govulncheck@latest && govulncheck ./...
+          go install golang.org/x/vuln/cmd/govulncheck@latest
+
+          # Find all go.mod files recursively, as we have multiple of them within the project
+          gomod_files=$(find . -name go.mod)
+
+          # Loop through each go.mod file path
+          for filepath in $gomod_files; do
```
ivanvc, line 25: Please refer to https://github.com/etcd-io/etcd/pull/18171#discussion_r1639012673 :sweat_smile:

henrybear327: Thank you for the suggestion! :) Patched as suggested!

ahrtr:
3.4 has only one go module.

ivanvc:
> 3.4 has only one go module. 

Good point. We don't need this change then.

henrybear327:
My idea was that we update in case in the future we need this, but after second thought, we would never do active development here anymore, so I will close this one for now!

PR #18164, opened 2024-06-13
bugfix: register of walWriteSec

andyxning:
bugfix: https://github.com/etcd-io/etcd/pull/17618.

`walWriteSec` should be registered in prometheus.

/cc @Akiqqqqqqq

PR #18174, opened 2024-06-14
bugfix: register of walWriteSec

andyxning:

cherry-pick https://github.com/etcd-io/etcd/pull/18164 to release-3.5 

/cc @ivanvc

### DIFF
--- a/server/wal/metrics.go
+++ b/server/wal/metrics.go
@@ -47,5 +47,6 @@ var (
 
 func init() {
 	prometheus.MustRegister(walFsyncSec)
+	prometheus.MustRegister(walWriteSec)
 	prometheus.MustRegister(walWriteBytes)
 }


ahrtr (approved):
Good catch!

Thank you.

ivanvc (approved):
LGTM. Thanks, @andyxning. Could you open a PR to update 3.5's CHANGELOG with this bug fix as pointed out by Benjamin? Thanks again.

andyxning:
@ivanvc This is a bugfix to adding support for `walWriteSec` and `walWriteSec` was added in v3.5. There is no need to update in the changelog, imho.

ivanvc:
> @ivanvc This is a bugfix to adding support for `walWriteSec` and `walWriteSec` was added in v3.5. There is no need to update in the changelog, imho. 

Yes, but at least 3.5.14 was released with this bug. I think that's why we need the release note, and why Benjamin requested it.

PR #18176, opened 2024-06-14
feat(server/embed): mkdir on log outputs

gyuho:
I notice that zap logger syncer assumes its parent directory already exists, as a result, --log-outputs values can cause server start failures:

> "caller":"etcdmain/etcd.go:75","msg":"failed to verify flags","error":"couldn't open sink \"/var/folders/z

It's fairly standard to mkdir parent dir in kubernetes artifacts as well. etcd can follow the same pattern.

e.g., https://github.com/kubernetes/kubernetes/blob/6ac60160c5729ade462b041b170ec8ac0f1eb3bc/staging/src/k8s.io/client-go/util/cert/io.go#L64-L69

```go
func WriteCert(certPath string, data []byte) error {
	if err := os.MkdirAll(filepath.Dir(certPath), os.FileMode(0755)); err != nil {
		return err
	}
	return os.WriteFile(certPath, data, os.FileMode(0644))
}
```

### DIFF
--- a/server/embed/config_logging.go
+++ b/server/embed/config_logging.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"path/filepath"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -66,6 +67,7 @@ func (cfg *Config) setupLogging() error {
 		}
 
 		outputPaths, errOutputPaths := make([]string, 0), make([]string, 0)
+		logDirs := make([]string, 0)
 		isJournal := false
 		for _, v := range cfg.LogOutputs {
 			switch v {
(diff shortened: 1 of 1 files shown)


serathius:
Not convinced about this time, having etcd autocreate directories for itself opens a can of worms related to default permissions and ownership. I think there is a distinction between directories where application store data which are usually exclusive to them and directories for logging which are usually shared. I don't think the example from k8s you provided is good. I'm not up to date with K8s logging, but I don't remember the klog (K8s logging library) creating directory for itself.

gyuho:
@serathius Fair enough. Let's revisit in the future :)

tjungblu:
I wouldn't entirely dismiss it, we currently have to carry a silly init container around to do that:
https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/pod.yaml#L16-L26

What about having another argument that can create the log dir? `--logger-dir=/var/log/whatever` / `--logger-dir-perms=0700`

serathius:
It's not silly. That's why Kubernetes supports init containers. Also this only hit you only because you are writing logs against the K8s best practices of writing to stdout and stderr.

tjungblu:
@serathius we do not supply any configuration besides `etcd --logger=zap --log-level=${VERBOSITY}` - so the defaults are not compliant with K8s best practises.

serathius:
@tjungblu Let's then change it. Could you be more specific what defaults you mean that are not compliant to best practices?

Issue #17777, opened 2024-04-11
Organize etcdctl commands and add more documents


ah8ad3:
### What would you like to be added?

I want commands of etcdctl to be organized and documented in a better way.

1. For organizing i suggest to remove nested part of commands for example instead of having this:
```
        user add                Adds a new user
        user delete             Deletes a user
        user get                Gets detailed information of a user
        user grant-role         Grants a role to a user
        user list               Lists all users
        user passwd             Changes password of user
        user revoke-role        Revokes a role from a user
```

We can do something like this:
```
      Basic KV commands:
        del                  Removes the specified key or range of keys [key, range_end)
        get                  Gets the key or a range of keys
        put                  Puts the given key into the store

      Management Commands:
        user                  User related commands
        role                  Role related commands
```
And rest of this for other commands. 

2. For documenting i suggest to make the documents and help texts for commands better and more helpful throw some examples and define better.
For example when we invoke `etcdctl leahe -h` we get 
```
NAME:
        lease - Lease related commands

USAGE:
        etcdctl lease <subcommand> [flags]
```

Maybe we can do something like this:
```
NAME:
        lease - Lease related commands

Description:
        Leases in etcd act as timers for data, automatically deleting keys when their associated lease expires.

USAGE:
        etcdctl lease <subcommand> [flags]

        etcdctl lease grant 60   # Creates a lease that lasts 60 seconds

        etcdctl lease keep-alive <lease_id> --once # Resets the keep-alive time to its original value and cobrautl.Exits immediately

```
Add a one line description so new users can understand what it is without going and searching for what is the lease, it's a tradeoff we don't want to throw the whole document here just a little sip of what is it. 
And put some popular usages in the USAGE section would be nice too.

### Why is this needed?

As a newcomer to etcdctl it was hard for me to understand what it does at first, i was constantly searching the websites for commands to understand some of it. To enhance user experience, I propose implementing measures to clarify command functionalities upfront, easing the learning curve.

Also i think the first view of `etcdctl -h` is a little bit crowded and i believe with removing some subcommands that are not necessarry to show at first and style it a little bit this way users are greeted with a more intuitive interface.

ah8ad3:
I want to get more involve in etcd community and i think with this issue i can help to make `etcdctl` better and also understand the basics and hopefully start working on core of etcd. If team approves, i would like to define some tasks and start working on this issue.

ivanvc:
[...]

PR #18183, opened 2024-06-16
Add examples to `etcdctl snapshot` command's help

ah8ad3:
Part of https://github.com/etcd-io/etcd/issues/17777.
Add example section into `etcdctl` by first adding it to helper function of cobra.
Create a `util` module into `etcdctl` move helper there.
Add a normalizer function for exmaples.
Change indent of `etcdctl` from `"\t`" to `"  "`.

I tried to keep this PR simple and add only functionality of examples to helper and add examples of snapshot so it can be easy to review. 
/cc @ivanvc @jmhbnz

### DIFF
--- a/etcdctl/util/normalizer.go
+++ b/etcdctl/util/normalizer.go
@@ -1,0 +1,49 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import "strings"
+
+const indentation = "  "
+
+// Normalize normalizes a string:
+//  1. trim the leading and trailing space
+//  2. add an indentation before each line
+func Normalize(s string) string {
+	if len(s) == 0 {
+		return s
+	}
+	return normalizer{s}.trim().indent().string
+}
+
+type normalizer struct {
+	string
+}
+
+func (n normalizer) trim() normalizer {
+	n.string = strings.TrimSpace(n.string)
+	return n
+}
+
+func (n normalizer) indent() normalizer {
+	indentedLines := []string{}
+	for _, line := range strings.Split(n.string, "\n") {
+		trimmed := strings.TrimSpace(line)
+		indented := indentation + trimmed
+		indentedLines = append(indentedLines, indented)
+	}
+	n.string = strings.Join(indentedLines, "\n")
+	return n
+}
--- a/pkg/cobrautl/help.go
+++ b/pkg/cobrautl/help.go
@@ -43,6 +43,10 @@ var (
 			}
 			return strings.Join(parts, " ")
 		},
+		"indent": func(s string) string {
+			pad := strings.Repeat(" ", 2)
+			return pad + strings.Replace(s, "\n", "\n"+pad, -1)
+		},
 	}
 )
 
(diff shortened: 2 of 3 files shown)


ah8ad3:
Im not sure about this test https://github.com/etcd-io/etcd/actions/runs/9534220896/job/26278395488?pr=18183

ah8ad3:
> I think you need to run a make fix and commit the changes to the go.mods, because you moved a file from pkg to etcdctl and didn't update the go.mod.

I see what you mean, thanks that makes sense. I updated the code hopefully that will fix it.

ivanvc:
Speaking of moving the file from `pkg/cobrautil` to `etcdctl/util`, why did you decide to do so? I'm not persuaded by this change, as if at some point we want to improve the help from other commands like `etcdutl`, it would make sense to move it back. Or maybe there's some reason why you moved it that is not clear to me.

ivanvc:
/retitle Add examples to `etcdctl snapshot` command's help

ah8ad3:
>  Speaking of moving the file from pkg/cobrautil to etcdctl/util, why did you decide to do so? I'm not persuaded by this change, as if at some point we want to improve the help from other commands like etcdutl, it would make sense to move it back. Or maybe there's some reason why you moved it that is not clear to me.

As i understand this help file is only for `etcdctl` and not `etcdutl` and `etcd` commands, So it would make sense to move it to the `etcdctl` package. 
Also this helper is not well written, or maybe it was few years ago, we should change it to be more readable but i don't want to do that in this PR.

etcdctl/util/help.go
```diff
@@ -14,7 +14,7 @@
 
 // copied from https://github.com/rkt/rkt/blob/master/rkt/help.go
 
-package cobrautl
+package util
```
jmhbnz, line 17: Hey @ah8ad3 - Thanks for raising this - In order to ensure this pr is not blocked on wider consensus I would suggest not moving from `pkg` and keep this pr to only adding examples.  If we want to move it imo we should do that as a separate pr.

ah8ad3:
Sorry, always happens to me when i rebase with wrong repository(forked one), tried to fix it. 
Let me know if should i do something.
PTAL @jmhbnz

etcdctl/util/help.go
ah8ad3: I'm pretty sure that this file is safe to move here, and we should do that, because the only use of this file is `etcdctl`, but i move it back. PTAL
I will raise another PR for that.

etcdctl/util/normalizer.go
```diff
@@ -0,0 +1,47 @@
...
+func (s normalizer) indent() normalizer {
+	indentedLines := []string{}
+	for _, line := range strings.Split(s.string, "\n") {
+		trimmed := strings.TrimSpace(line)
+		indented := Indentation + trimmed
+		indentedLines = append(indentedLines, indented)
+	}
+	s.string = strings.Join(indentedLines, "\n")
+	return s
+}
```
ivanvc, line 47: I'm not entirely sold on introducing this struct. However, I see why you did it. As I was reviewing, some comments/observations came to my mind:

1. I usually see the receiver variable named after its struct. In this case, I'd suggest using `n` rather than `s` (https://google.github.io/styleguide/go/decisions#receiver-names).
2. I think that modifying the held string's value is not transparent.
3. I don't know if this struct will be reused or if it is going to be used just for this context.

Because of these reasons, I wonder if it would be easier (and simpler) to define a single function that does the normalization without adding a struct that does it.

ivanvc (commented):
Thanks for addressing James' comment. I think this is going in the right direction. I left a couple of comments.

pkg/cobrautl/help.go
```diff
@@ -43,6 +43,10 @@ var (
 			}
 			return strings.Join(parts, " ")
 		},
+		"indent": func(s string) string {
+			pad := strings.Repeat(" ", 2)
```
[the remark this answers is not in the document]
ah8ad3: Thanks for the review, TBH using tab is not common in cli tools(at least tools that i worked with), most of tools are using two spaces example are etcd, etcdutl, kubectl and many more examples. IMO it is gaining better user experience this way, but i'm not that biased to it, i'm suggesting using this way if we don't want to i can rollback.

etcdctl/util/normalizer.go
```diff
@@ -0,0 +1,47 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import "strings"
+
+const Indentation = "  "
```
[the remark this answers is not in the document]
ah8ad3: Nice catch, thanks.

etcdctl/util/normalizer.go
ah8ad3: Thanks, you are right i used pretty old code structure here, i will update them. 

For number 3 we don't need it to be structure now, but in future this function can evolve fast and add more thing to it if we want to extend normalizer function to other parts when we want to use exports, add more attributes to the `etcdctl` cobra that need some kind of normalizer.
And it is simpler/clearer to me/personally to use struct than using an function because i can see the function's parent and i can extend it's functionality easier. But if you think it is not needed i can replace it with a simple function right now!

ah8ad3: Can you please define this phrase? 
> I think that modifying the held string's value is not transparent.

ivanvc: It could be biased or a personal preference, but in my opinion, returning a modified value rather than altering the property from the structure is less error-prone. I think the pattern of returning the struct to then run another function on it is more common in NodeJS than in Go.

So, in a nutshell, I think an implementation like the following would be cleaner:

```golang
func (s normalizer) indent() normalizer {
	...
	return normalizer{strings.Join(indentedLines, "\n")}
}
```

But, again, the pattern of returning the `normalizer` so it can be chained with other functions is something that I haven't seen in Go a lot. So, this could be cleaner:

```golang
func (s normalizer) indent() string {
	...
	return strings.Join(indentedLines, "\n")
}
```

That's why I question the use of this structure, as it seems like it could be simplified to a simple, unexported function.

With the caveats, I may be missing the bigger picture if you think this could be reused in later implementations as part of your bigger effort to have more examples across the CLI tools.

pkg/cobrautl/help.go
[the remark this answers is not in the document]
ivanvc: Let's keep the two spaces, and you have a valid point. I wasn't aware of this and wanted to understand why you changed it.

etcdctl/util/normalizer.go
ah8ad3: Thanks for explaining i see what you mean, TBH currently the implementation is just about using it in the function here, with those concerns i can do two things
1. replacing structure with a simple function and for later if we needed we can do a structure
2. altering the current behavior from returning structure to returning string to be more go style

Your'e call i'm ok either way.

ivanvc: @ahrtr, do you have an opinion on this discussion?

ahrtr: Overall not a big deal to me. Since it's just internal implementation, we can update it later when needed.

The `normalizer{s}.trim().indent().string` is a good pattern to me. It reflects the ideas of functional programming and also (although compromised) object-oriented programming. 

But on the other hand, the implementation is pretty simple, so we can also just get it included in one function. In other words, no need to add `normalizer` struct at all.

Again, as mentioned above, it isn't a big deal. This is solely for the purpose of a purely technical discussion. 

Let's just keep the current implementation as it's if there isn't no strong objection.

etcdctl/util/normalizer.go
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+package util
+
+import "strings"
+
+const indentation = "  "
+
+// Examples normalizes a command's examples to follow the conventions.
+func Examples(s string) string {
```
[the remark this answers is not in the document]
ah8ad3: Done

ahrtr (approved):
LGTM

Thanks @ah8ad3

jmhbnz (approved):
LGTM - New help examples look great, thanks @ah8ad3

ivanvc (approved):
LGTM. Thanks, @ah8ad3 :)

Issue #8060, opened 2017-06-08
Expose /metrics an a separate port


fabxc:
In many scenarios etcd nodes run on designated nodes that are very restrictive for outside traffic.

If we want to monitor etcd, we don't need access to any critical APIs but just `/metrics`. It's not generally feasible to run Prometheus on the same nodes as etcd though. Having etcd expose `/metrics` on a designated port that can be exposed relatively safely would solve those scenarios.

It could be an optional flag,that exposes an additonal `/metrics` endpoint on another address if set.

@xiang90 @brancz

heyitsanthony:
@fabxc would etcd get an allocation on https://github.com/prometheus/prometheus/wiki/Default-port-allocations or is that something different?

brancz:
That would generally make sense @heyitsanthony. Next one up is usually marked. It is `9267` right now. This should of course be configurable if a user desires to bind it differently.

fabxc:
TBH, I'm not too much of a fan of dedicated Prometheus ports beyond our
core components. It's just not practical to maintain those global
registries.
As an etcd user, I'd probably prefer the default metric port to be default
API port + 1 or similar, so it's easier to remember.

brancz:
FWIW certainly not all exporters and applications oblige to it, so etcd being a relatively prominent application I think it's valid to choose whatever we like best. The wiki page is also open and I don't think anyone monitors the changes much, so the reliability is questionable either way (I've brought this up before but it was decided to keep as guidance for now).

xiang90:
@heyitsanthony 

do you think this is a reasonable enhancement we could do in 3.3?

gyuho:
@xiang90 @heyitsanthony Do we want to *move* `/metrics` endpoints to a separate port, or *duplicate* the handler in the separate port?

heyitsanthony:
Duplicate the handler. There could be clients that have access to the internal port that expect /metrics to be available.

brancz:
Awesome! Thanks for the collaboration everyone!

seh:
Is it correct that even if one activates additional URLs with distinct ports for _etcd_ serving metrics (per #8242), the server still demands the same authentication credentials as the other _etcd_ ports (2379, 2380)? It is challenging to arrange for Prometheus to have the client key and certificate available to authenticate to _etcd_—as mentioned in kubernetes/kubernetes#53405.

[...]

PR #18186, opened 2024-06-17
fix(server): enforce listen-metrics-urls client TLS info when its scheme is https/unixs

gyuho:
Otherwise, it will fail with

> cannot listen on TLS for 127.0.0.1:8080: KeyFile and CertFile are not presented

We should instead explicitly fail fast, with a clear error message.

Also, adding some documentation how to configure TLS for metrics URLs.

c.f., https://github.com/etcd-io/etcd/issues/8060

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -822,6 +822,24 @@ func (e *Etcd) pickGRPCGatewayServeContext(splitHTTP bool) *serveCtx {
 	panic("Expect at least one context able to serve grpc")
 }
 
+var errMissingClientTLSInfoForMetricsURL = errors.New("client TLS key/cert (--cert-file, --key-file) must be provided for metrics url")
+
+func (e *Etcd) createMetricsListener(murl url.URL) (net.Listener, error) {
+	tlsInfo := &e.cfg.ClientTLSInfo
+	switch murl.Scheme {
+	case "http":
+		tlsInfo = nil
+	case "https", "unixs":
+		if e.cfg.ClientTLSInfo.Empty() {
+			return nil, errMissingClientTLSInfoForMetricsURL
+		}
+	}
+	return transport.NewListenerWithOpts(murl.Host, murl.Scheme,
+		transport.WithTLSInfo(tlsInfo),
+		transport.WithSocketOpts(&e.cfg.SocketOpts),
+	)
+}
+
 func (e *Etcd) serveMetrics() (err error) {
 	if e.cfg.Metrics == "extensive" {
 		grpc_prometheus.EnableHandlingTimeHistogram()
@@ -833,14 +851,7 @@ func (e *Etcd) serveMetrics() (err error) {
 		etcdhttp.HandleHealth(e.cfg.logger, metricsMux, e.Server)
 
 		for _, murl := range e.cfg.ListenMetricsUrls {
-			tlsInfo := &e.cfg.ClientTLSInfo
-			if murl.Scheme == "http" {
-				tlsInfo = nil
-			}
-			ml, err := transport.NewListenerWithOpts(murl.Host, murl.Scheme,
-				transport.WithTLSInfo(tlsInfo),
-				transport.WithSocketOpts(&e.cfg.SocketOpts),
-			)
+			ml, err := e.createMetricsListener(murl)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


server/embed/etcd.go
```diff
@@ -822,6 +822,24 @@ func (e *Etcd) pickGRPCGatewayServeContext(splitHTTP bool) *serveCtx {
 	panic("Expect at least one context able to serve grpc")
 }
 
+var errMissingClientTLSInfoForMetricsURL = errors.New("client TLS key/cert (--cert-file, --key-file) must be provided for metrics url")
+
+func (e *Etcd) createMetricsListener(murl url.URL) (net.Listener, error) {
```
serathius, line 827: Please add an e2e test for this.

server/embed/etcd_test.go
```diff
@@ -0,0 +1,24 @@
+package embed
```
henrybear327, line 1: The license header is missing in this new file!

serathius: Why is header validation passing?

serathius (approved):

tests/e2e/etcd_config_test.go
```diff
@@ -120,6 +121,60 @@ func TestEtcdUnixPeers(t *testing.T) {
...
+	}
+	defer func() {
+		if err := proc.Stop(); err != nil {
+			t.Error(err)
+		}
+		proc.Wait() // ensure the port has been released
+		_ = proc.Close()
+	}()
+
+	if err := e2e.WaitReadyExpectProc(context.TODO(), proc, []string{embed.ErrMissingClientTLSInfoForMetricsURL.Error()}); err != nil {
```
serathius, line 173: No need to fix it now as there are more config e2e tests with this problem, but it would be a nice followup.

In case that etcd doesn't produce the log, would be nice to test to exit after couple of seconds instead of waiting infinitely.

tests/e2e/etcd_config_test.go
```diff
@@ -120,6 +121,60 @@ func TestEtcdUnixPeers(t *testing.T) {
...
+
+	proc, err := e2e.SpawnCmd(commonArgs, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := proc.Stop(); err != nil {
+			t.Error(err)
+		}
+		proc.Wait() // ensure the port has been released
```
serathius, line 169: Looked at other test, not all of them do this. Would be nice to have a consistent approach to shutting down proc.

serathius:
cc @ahrtr

tests/e2e/etcd_config_test.go
gyuho: Haha, yes. Some archaic code base here :)

serathius: Heh, we have been trying to clean it slowly up but we run out of steam.

https://github.com/etcd-io/etcd/issues/13637

tests/e2e/etcd_config_test.go
```diff
@@ -120,6 +121,60 @@ func TestEtcdUnixPeers(t *testing.T) {
...
+		_ = proc.Close()
```
ahrtr, line 170: Minor comment: `proc.Wait()` seems to be unnecessary, `proc.Close()` already calls `ep.wg.Wait()`.

https://github.com/etcd-io/etcd/blob/c70e0e4f558961b7c113540050df2ea8b8a4fc9e/pkg/expect/expect.go#L356-L362

ahrtr (approved):
LGTM with a minor comment.

You might want to backport the change to 3.5.

server/embed/etcd.go
```diff
@@ -822,6 +822,24 @@ func (e *Etcd) pickGRPCGatewayServeContext(splitHTTP bool) *serveCtx {
 	panic("Expect at least one context able to serve grpc")
 }
 
+var ErrMissingClientTLSInfoForMetricsURL = errors.New("client TLS key/cert (--cert-file, --key-file) must be provided for metrics url")
```
ahrtr, line 825:
```suggestion
var ErrMissingClientTLSInfoForMetricsURL = errors.New("client TLS key/cert (--cert-file, --key-file) must be provided for metrics secure url")
```

ahrtr (approved):

server/embed/etcd.go
ahrtr: Let me deliver a quick followup PR to fix this.

PR #18200, opened 2024-06-18
client/v3: fix open file memory leak

vivekpatani:
- reference: https://github.com/etcd-io/etcd/issues/18085
- defer close and catch errors if unsuccessful

### DIFF
--- a/client/v3/snapshot/v3_snapshot.go
+++ b/client/v3/snapshot/v3_snapshot.go
@@ -54,13 +54,29 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 	if err != nil {
 		return "", err
 	}
-	defer cli.Close()
+	defer func(cli *clientv3.Client) {
+		err = cli.Close()
+		if err != nil {
+			lg.Error("failed to close etcd client", zap.Error(err))
+		}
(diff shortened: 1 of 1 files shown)


client/v3/snapshot/v3_snapshot.go
```diff
@@ -54,13 +54,29 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
...
+		err = os.RemoveAll(path)
+		if err != nil {
+			lg.Error("failed to remove path", zap.String("path", path), zap.Error(err))
+		}
+	}(partpath)
 
 	var f *os.File
 	f, err = os.OpenFile(partpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, fileutil.PrivateFileMode)
+	defer func(f *os.File) {
+		err = f.Close()
```
fuweid, line 75: The `f` can be nil here. defer should be called only if the err is nil.

tjungblu: and it suffers from the same issue on Windows as I've described here:
https://github.com/etcd-io/etcd/pull/18086#discussion_r1617253049

vivekpatani:
clean up, duplicate.

PR #18201, opened 2024-06-18
tests/e2e: add e2e test to reproduce issue 18089

MadhavJivrajani:
e2e for https://github.com/etcd-io/etcd/issues/18089 

The goal is to reproduce a DELETE event being dropped in a watch after a compaction occurs on the revision where the deletion took place. In order to reproduce this, we perform the following sequence (steps for reproduction thanks to @ahrtr):
  - PUT k v2 (assume returned revision = r2)
  - PUT k v3 (assume returned revision = r3)
  - PUT k v4 (assume returned revision = r4)
  - DELETE k (assume returned revision = r5)
  - PUT k v6 (assume returned revision = r6)
  - COMPACT r5
  - WATCH rev=r5

We should get the DELETE event (r5) followed by the PUT event (r6). However, currently we only get the PUT event with returned revision of r6 (key=k, val=v6).

/assign @ahrtr @serathius @siyuanfoundation

### DIFF
--- a/tests/e2e/watch_test.go
+++ b/tests/e2e/watch_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"golang.org/x/sync/errgroup"
 
+	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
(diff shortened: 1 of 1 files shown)


siyuanfoundation:
Very nice! Thank you @MadhavJivrajani

siyuanfoundation (approved):

serathius (approved):

serathius:
cc @ahrtr

serathius:
Please fix `pull-etcd-verify`

tests/e2e/delete_event_drop_issue18089_test.go
```diff
@@ -0,0 +1,108 @@
...
+// This is an e2e test to reproduce the issue reported in: https://github.com/etcd-io/etcd/issues/18089
+//
+// The goal is to reproduce a DELETE event being dropped in a watch after a compaction
+// occurs on the revision where the deletion took place. In order to reproduce this, we
+// perform the following sequence (steps for reproduction thanks to @ahrtr):
+//   - PUT k v2 (assume returned revision = r2)
+//   - PUT k v3 (assume returned revision = r3)
+//   - PUT k v4 (assume returned revision = r4)
+//   - DELETE k (assume returned revision = r5)
+//   - PUT k v6 (assume returned revision = r6)
```
fuweid, line 37: Side note: Even if there is not new revision for K, the DELETE event is missing as well.

tests/e2e/delete_event_drop_issue18089_test.go
```diff
@@ -0,0 +1,108 @@
...
+	t.Logf("DELTE key=%s", key)
+	deleteResp, err := c.KV.Delete(ctx, key)
+	require.NoError(t, err)
+
+	t.Logf("PUT key=%s, val=%s", key, v6)
+	_, err = c.KV.Put(ctx, key, v6)
+	require.NoError(t, err)
+
+	t.Logf("COMPACT rev=%d", deleteResp.Header.Revision)
+	_, err = c.KV.Compact(ctx, deleteResp.Header.Revision)
```
fuweid, line 85: Suggest to use --physical to block until compact finished

tests/e2e/delete_event_drop_issue18089_test.go
serathius: Thinking if we could use a modelReplay from robustness to provide a clearer explanation of expectations for the results.

tests/e2e/delete_event_drop_issue18089_test.go
MadhavJivrajani: Done!

tests/e2e/delete_event_drop_issue18089_test.go
MadhavJivrajani: @serathius could you elaborate?

tests/e2e/delete_event_drop_issue18089_test.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		if watchResp.Events[0].Type != mvccpb.PUT {
+			t.Fatalf("unexpected event type, expected: %s, got: %s", mvccpb.PUT, watchResp.Events[0].Type)
+		}
+		if string(watchResp.Events[0].Kv.Key) != key {
+			t.Fatalf("unexpected key, expected: %s, got: %s", key, string(watchResp.Events[0].Kv.Key))
+		}
+		if string(watchResp.Events[0].Kv.Value) != v6 {
+			t.Fatalf("unexpected valye, expected: %s, got: %s", v6, string(watchResp.Events[0].Kv.Value))
+		}
```
ahrtr, line 103: The first event should be `DELETE`, are you planning to update these checks after the issue is resolved? If yes, please add a TODO.

tests/e2e/delete_event_drop_issue18089_test.go
ahrtr: > Thinking if we could use a modelReplay from robustness to provide a clearer explanation of expectations for the results.

NO. 

Robustness testing is characterized by randomness and is often conducted under high traffic conditions. But this is just a very corner case, which should be covered by e2e or integration test.

ahrtr:
Can we add the test info `watch_test.go`?  I see that there is an existing `watch_delay_test.go`, I think we can rename it to `watch_test.go` if there is no strong objection.

tests/e2e/delete_event_drop_issue18089_test.go
serathius: Heh, I said modelReplay, which is deterministic stub of etcd.

tests/e2e/delete_event_drop_issue18089_test.go
MadhavJivrajani: Yes, the hope is to try and have this in till we fix the actual issue.

Added a TODO, thanks!

MadhavJivrajani:
@ahrtr can you PTAL again?

tests/e2e/delete_event_drop_issue18089_test.go
ahrtr: > I said modelReplay, which is deterministic stub of etcd.

Not sure we are talking about the same thing.

My understanding is the existing check on watch API guarantee should be already good enough, otherwise this issue could't be reproduced by robustness test.

Any specific additional check for such corner case isn't reasonable to be added into robustness test.

ahrtr (approved):
LGTM, nice work!

Thanks @MadhavJivrajani

PR #18204, opened 2024-06-18
Use $(MAKE) instead of make

henrybear327:
Recursive `make` commands should always use the variable `MAKE`, as the value of this variable is the file name with which make was invoked

References:
- https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html
- https://github.com/etcd-io/gofail/pull/69#discussion_r1644163380


ivanvc (approved):
LGTM. Thanks, @henrybear327.

/cc @serathius
/retest

ahrtr (approved):

serathius (approved):

PR #18209, opened 2024-06-20
fix typo in mvcc: store.go

Willhow-Gao:
fix typo in mvcc module,store.go


ivanvc (commented):
I think an even better fix, would be to remove the named return values, as they're not used in the implementation.

ivanvc:
@Willhow-Gao, can you squash your commits into a single one, and sign your commits so the developer certificate of origin (DCO) check passes?

You can squash your three commits and sign off by running:

```
git reset --soft HEAD~3
git commit --signoff -m <commit message>
```

Issue #17472, opened 2024-02-21
[3.4] Linters fail locally, but pass on CI


ivanvc:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

There are some issues regarding the state of the linters in the CI.
* The shellcheck and revive linters fail when locally running the `fmt` `PASS`. But they are not checked on the CI. 
  * Shellcheck errors:
    ```
    'shellcheck' started at Wed Feb 21 11:13:45 AM CST 2024                                                                                                                                                 
    shellcheck checking failed:                                                                                                                                                                             
    test:168:64: note: Double quote to prevent globbing and word splitting. [SC2086]                                                                                                                        
    test:168:92: note: Double quote to prevent globbing and word splitting. [SC2086]                                                                                                                        
    test:205:25: note: Double quote to prevent globbing and word splitting. [SC2086]                                                                                                                        
    test:206:26: note: Double quote to prevent globbing and word splitting. [SC2086]                    
    test:208:25: note: Double quote to prevent globbing and word splitting. [SC2086]                                                                                                                        
    test:209:25: note: Double quote to prevent globbing and word splitting. [SC2086]                    
    test:210:25: note: Double quote to prevent globbing and word splitting. [SC2086]                    
    test:344:59: note: Double quote to prevent globbing and word splitting. [SC2086]                    
    test:360:26: note: Double quote to prevent globbing and word splitting. [SC2086]                    
    test:361:26: note: Double quote to prevent globbing and word splitting. [SC2086]
    scripts/install-marker.sh:13:6: note: Double quote to prevent globbing and word splitting. [SC2086]
    ```
  * Revive errors:
    ```
    revive checking failed:
    clientv3/naming/endpoints/internal/update.go:15:1: should have a package comment
    clientv3/internal/resolver/resolver.go:15:1: should have a package comment
    wal/walpb/record.go:15:1: should have a package comment
    pkg/grpc_testing/stub_server.go:15:1: should have a package comment
[...]

PR #18216, opened 2024-06-23
etcd 3.4: install shellcheck and report if not installed.

thedtripp:
Looking into https://github.com/etcd-io/etcd/issues/17472, it appears that shellcheck is not installed. I added some code to install it, though I'm not sure if this is the correct place to do this. 

At the very least, I think we should add an else block to inform that shellcheck isn't installed:

	if command -v shellcheck >/dev/null; then
		shellcheckResult=$(shellcheck -fgcc build test scripts/*.sh 2>&1 || true)
		if [ -n "${shellcheckResult}" ]; then
			echo -e "shellcheck checking failed:\\n${shellcheckResult}"
			exit 255
		fi
	else
		echo "Skipping shellcheck..."
	fi

### DIFF
--- a/test
+++ b/test
@@ -397,12 +397,30 @@ function release_pass {
 }
 
 function shellcheck_pass {
+	if ! command -v shellcheck >/dev/null; then
+        echo "shellcheck: command not found. Installing ShellCheck..."
+        # Determine the platform and install shellcheck accordingly
+        if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+            if command -v apt-get >/dev/null; then
+                sudo apt-get update
+                sudo apt-get install -y shellcheck
+            else
+                echo "Unsupported Linux distribution. Please install ShellCheck manually."
+                return 1
+            fi
+		else
+            echo "Unsupported operating system. Please install ShellCheck manually."
+            return 1
+        fi
+	fi
 	if command -v shellcheck >/dev/null; then
 		shellcheckResult=$(shellcheck -fgcc build test scripts/*.sh 2>&1 || true)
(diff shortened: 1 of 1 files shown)


jmhbnz (changes requested):
Hey @thedtripp - Thanks for proposing this, agree `shellcheck` should be automatically installed if not already present.

We do already have this functionality in `main`, refer: https://github.com/etcd-io/etcd/pull/14872.

We should backport the approach from that pull request to `release-3.5` then `release-3.4` so all branches managed `shellcheck` the same way.

idnandre:
Hi @thedtripp @jmhbnz.
In "Tests / test (linux-amd64-fmt)" workflows there is still exists skipping revive message. Are we need to install revive also?

thedtripp:
> Hey @thedtripp - Thanks for proposing this, agree `shellcheck` should be automatically installed if not already present.
> 
> We do already have this functionality in `main`, refer: #14872.
> 
> We should backport the approach from that pull request to `release-3.5` then `release-3.4` so all branches managed `shellcheck` the same way.

Hey @jmhbnz. I submitted another PR for 3.5 here: https://github.com/etcd-io/etcd/pull/18229
If it looks good, I will repeat for 3.4. Thanks.

thedtripp:
> Hi @thedtripp @jmhbnz. In "Tests / test (linux-amd64-fmt)" workflows there is still exists skipping revive message. Are we need to install revive also?

Hey @idnandre,
Thanks for your review. I plan to address revive in a separate PR.

jmhbnz:
Closing in favor of #18248

Issue #18090, opened 2024-05-29
Panic:   index out of range[0] with length 0 at v3.5.11 go.etcd.io/etcd/client/v3/client.go:302


hanxuejian:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

We have a test scenario to reset all etcd services at the same time. In this test scenario, there is a low probability that the panic stack occurs on the etcd client.

The panic stack is as follows:

panic: runtiome error: index out of range[0] with length 0
goroutine 195602 [running]:
go.etcd.io/etcd/client/v3.(*Client).dial(***)
go.etcd.io/etcd/client/v3/client.go:302

### What did you expect to happen?

Panic should not occur. The code can be hardened. The null check may be performed on the input parameter before SetEndpoints is executed.

### How can we reproduce it (as minimally and precisely as possible)?

I do not know the specific scenario where the panic occurs. However, according to the code analysis, when the autoSync method in etcd/client/v3/client.go is executed periodically, if MemberList returns a null value, Endpoints may be set to null.

### Etcd version (please run commands below)

v3.5.11

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

ahrtr:
> We have a test scenario to reset all etcd services at the same time

Can you explain how do you "reset all etcd services"?

ahrtr:
> In my test environment, there are three etcd nodes. I use the kill -9 command to kill three etcd processes at the same time. The three processes can be restarted normally and form a cluster. However, there is a low probability that the client panics.

[...]

PR #18220, opened 2024-06-24
Add verification on `client.Sync` to ensure the returned endpoints never empty

ahrtr:
It should be helpful to detect issue like https://github.com/etcd-io/etcd/issues/18090


serathius (approved):
LGTM for the change, should we consider returning error early to also avoid setting empty endpoints?

ahrtr:
> should we consider returning error early to also avoid setting empty endpoints?

I tend to keep it as it's. As long as `MemberList` returns successfully, the endpoints should never be empty. Returning an error means that we need to add test to test/verify it, but we can't think of a single case to trigger this scenario.

Issue #18023, opened 2024-05-17
Add server level feature gate


siyuanfoundation:
### What would you like to be added?

Add the Kubernetes style feature gate framework into etcd to gate future feature enhancement.
Users would be able to turn features on or off for the single server and query feature enablement in a consistent way. 

This issue is for tracking [KEP-4578](https://github.com/kubernetes/enhancements/pull/4610).
Tasks in this KEP include

* Milestone 1
- [x] server feature gate interface defined, and code copied from K8s into `pkg/`: https://github.com/etcd-io/etcd/pull/18062.
- [x] `--feature-gates` flag added: https://github.com/etcd-io/etcd/pull/18279, https://github.com/etcd-io/etcd/pull/18365.
- [x] feature gate added to the server code, and used by a server level experimental feature: https://github.com/etcd-io/etcd/pull/18359
- [x] feature metrics added. https://github.com/etcd-io/etcd/issues/19324

* Milestone 2
- [x] e2e tests added for the feature gate equivalent of the selected experimental feature.
- [x] robustness test scenarios added for the selected experimental feature(s).
- [x] documentation to track all feature gates added to [etcd-io/website]: https://github.com/etcd-io/website/issues/940.

* Milestone 3
- [x] migrate all existing `--experimental` server level features to feature gates without removing the old flags.
List of features to migrate:
  - [x] `experimental-stop-grpc-service-on-defrag`: https://github.com/etcd-io/etcd/pull/18359
  - [x] `experimental-enable-distributed-tracing`: https://github.com/etcd-io/etcd/issues/19050
  - [x] `experimental-initial-corrupt-check`: https://github.com/etcd-io/etcd/pull/18478
  - [x] `experimental-compact-hash-check-enabled`: https://github.com/etcd-io/etcd/pull/19053
  - [x] `experimental-enable-lease-checkpoint` and `experimental-enable-lease-checkpoint-persist`: https://github.com/etcd-io/etcd/pull/19356
  - [x] `experimental-memory-mlock`: https://github.com/etcd-io/etcd/issues/19061
  - [x] `experimental-txn-mode-write-with-shared-buffer`: https://github.com/etcd-io/etcd/issues/19063

### Why is this needed?

To make it easier to add and use new features in etcd.
Please refer to https://github.com/kubernetes/enhancements/issues/4578 for more details.

siyuanfoundation:
[...]

PR #18234, opened 2024-06-26
Add "feature-gates" flag and field in config file and add feature gate for StopGRPCServiceOnDefrag

siyuanfoundation:
part of https://github.com/etcd-io/etcd/issues/18023

This PR includes changes to
* add "feature-gates" flag and field in config file
* add feature gate for StopGRPCServiceOnDefrag
* verify that feature gate flag is not used at the same time as the --experimental flag.

### DIFF
--- a/etcd.conf.yml.sample
+++ b/etcd.conf.yml.sample
@@ -155,3 +155,6 @@ cipher-suites: [
 # Limit etcd to specific TLS protocol versions 
 tls-min-version: 'TLS1.2'
 tls-max-version: 'TLS1.3'
+
+# Comma-separated list of server feature gate enablement in the format of feature=true|false
+server-feature-gates: StopGRPCServiceOnDefrag=false,DistributedTracing=false
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -42,6 +42,7 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/pkg/v3/featuregate"
 	"go.etcd.io/etcd/pkg/v3/flags"
 	"go.etcd.io/etcd/pkg/v3/netutil"
 	"go.etcd.io/etcd/server/v3/config"
@@ -50,6 +51,7 @@ import (
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3compactor"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3discovery"
+	"go.etcd.io/etcd/server/v3/features"
 )
 
 const (
@@ -108,6 +110,8 @@ const (
 	maxElectionMs = 50000
 	// backend freelist map type
 	freelistArrayType = "array"
+
+	ServerFeatureGateFlagName = "server-feature-gates"
 )
 
 var (
@@ -455,6 +459,9 @@ type Config struct {
 
 	// V2Deprecation describes phase of API & Storage V2 support
 	V2Deprecation config.V2DeprecationEnum `json:"v2-deprecation"`
+
+	// ServerFeatureGate is a server level feature gate
+	ServerFeatureGate featuregate.FeatureGate
 }
 
 // configYAML holds the config suitable for yaml parsing
@@ -476,6 +483,8 @@ type configJSON struct {
 
 	ClientSecurityJSON securityConfig `json:"client-transport-security"`
 	PeerSecurityJSON   securityConfig `json:"peer-transport-security"`
+
+	ServerFeatureGatesJSON string `json:"server-feature-gates"`
 }
 
 type securityConfig struct {
@@ -576,6 +585,7 @@ func NewConfig() *Config {
 		},
 
 		AutoCompactionMode: DefaultAutoCompactionMode,
+		ServerFeatureGate:  features.NewDefaultServerFeatureGate(DefaultName, nil),
 	}
 	cfg.InitialCluster = cfg.InitialClusterFromName(cfg.Name)
 	return cfg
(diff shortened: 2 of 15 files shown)


etcd.conf.yml.sample
```diff
@@ -155,3 +155,6 @@ cipher-suites: [
 # Limit etcd to specific TLS protocol versions 
 tls-min-version: 'TLS1.2'
 tls-max-version: 'TLS1.3'
+
+# Comma-separated list of feature gate enablement in the format of feature=true|false
+feature-gates: StopGRPCServiceOnDefrag=true,DistributedTracing=false
```
liangyuanpeng, line 160: `StopGRPCServiceOnDefrag` is false by default in DefaultEtcdServerFeatureGates, we should keep the same value.

server/embed/config.go
```diff
@@ -877,6 +908,25 @@ func (cfg *configYAML) configFromFile(path string) error {
 	return cfg.Validate()
 }
 
+// SetFeatureGatesFromExperimentalFlags sets the feature gate values if their corresponding experimental flags are
+// explicitly set.
+func (cfg *Config) SetFeatureGatesFromExperimentalFlags(isExperimentalFlagSet func(string) bool, featureGatesVal string) error {
+	// verify that the feature gate and its experimental flag are not both set at the same time.
+	for expFlagName, featureName := range features.ExperimentalFlagToFeatureMap {
+		if isExperimentalFlagSet(expFlagName) && strings.Contains(featureGatesVal, string(featureName)) {
+			return fmt.Errorf("cannot specified both flags: --%s=(true|false) and --%s=%s=(true|false) at the same time, please just use --%s=%s=(true|false)",
```
henrybear327, line 917: cannot ~specified~ specify both flags

server/embed/config.go
```diff
@@ -877,6 +908,25 @@ func (cfg *configYAML) configFromFile(path string) error {
...
+	for expFlagName, featureName := range features.ExperimentalFlagToFeatureMap {
+		if isExperimentalFlagSet(expFlagName) && strings.Contains(featureGatesVal, string(featureName)) {
+			return fmt.Errorf("cannot specify both flags: --%s=(true|false) and --%s=%s=(true|false) at the same time, please just use --%s=%s=(true|false)",
+				expFlagName, featuregate.FlagName, featureName, featuregate.FlagName, featureName)
+		}
+	}
+
+	m := make(map[string]bool)
+	defaultEc := NewConfig()
+	if cfg.ExperimentalStopGRPCServiceOnDefrag != defaultEc.ExperimentalStopGRPCServiceOnDefrag {
```
jmhbnz, line 924: Question here - For a function that sounds generic we are only internally only focusing on `ExperimentalStopGRPCServiceOnDefrag` can you please reply with some some context on the intention for this function now and longer term to help me understand?

jmhbnz (commented):
Nice work @siyuanfoundation - Overall looking good, one main question below about ExperimentalFlags vs FeatureGate handling below.

server/embed/config.go
```diff
@@ -907,6 +957,7 @@ func (cfg *Config) Validate() error {
 	if err := cfg.setupLogging(); err != nil {
 		return err
 	}
+	cfg.ServerFeatureGate.(featuregate.MutableFeatureGate).SetLogger(cfg.logger)
```
serathius, line 960: Why this cannot be done in constructor?

server/embed/config_test.go
```diff
@@ -89,6 +91,117 @@ func TestConfigFileOtherFields(t *testing.T) {
...
+			expectedFeatures: map[featuregate.Feature]bool{
+				features.StopGRPCServiceOnDefrag: false,
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var b []byte
+			var err error
+			if tc.setExperimentalStopGRPCServiceOnDefrag {
```
serathius, line 162: Don't understand why `setExperimentalStopGRPCServiceOnDefrag` exist, or what it does.

server/embed/config_test.go
```diff
@@ -89,6 +91,117 @@ func TestConfigFileOtherFields(t *testing.T) {
...
+				b, err = yaml.Marshal(&yc)
+			} else {
+				yc := struct {
+					FeatureGatesJSON string `json:"feature-gates"`
+				}{
+
+					tc.featureGatesJSON,
+				}
+				b, err = yaml.Marshal(&yc)
+			}
```
serathius, line 179:
```suggestion
			data := map[string]string{}
			if tc.featureGatesJSON != "" {
						data["feature-gates"] = tc.featureGatesJSON
			}
			if tc.experimentalStopGRPCServiceOnDefrag != "" {
						data["experimental-stop-grpc-service-on-defrag"] = tc.experimentalStopGRPCServiceOnDefrag
			}
			b, err = yaml.Marshal(&yc)
```

server/etcdmain/help.go
```diff
@@ -103,6 +103,8 @@ Member:
     Read timeout set on each rafthttp connection
   --raft-write-timeout '` + rafthttp.DefaultConnWriteTimeout.String() + `'
     Write timeout set on each rafthttp connection
+  --feature-gates ''
+    A set of key=value pairs that describe feature gates for alpha/experimental features.
```
serathius, line 107: Can we list available feature flags?

serathius:
High level looks good, left some nits.
cc @ahrtr

server/embed/config.go
siyuanfoundation: because the server logger is not set up before this point. And the ServerFeatureGate could be needed before that, for example there could be a feature gate to change the logger.

server/etcdmain/help.go
siyuanfoundation: done.

server/embed/config.go
siyuanfoundation: Added some comments in the function. Does that make sense? 
`ExperimentalStopGRPCServiceOnDefrag` is just what we are migrating in this PR. For any experimental flag we are migrating in the future, this function needs to be updated.

server/embed/config_test.go
siyuanfoundation: Removed. I cannot use the `map[string]string` because yaml will marshal the `bool` as `string` which would cause parsing error later, so have to use the 2 struct approach.

server/embed/config.go
jmhbnz: Thanks for adding the explanation, sgtm.

ahrtr:
Please try not to raise huge PR.
Can you try to breakdown this PR into a couple of smaller PRs?

siyuanfoundation:
> part of #18023

Opened https://github.com/etcd-io/etcd/pull/18279.
Closing this one.

PR #18236, opened 2024-06-26
increase/commit auth revision when auth enable

andyxning:

Since all the auth store update function should update auth revision, i.e., calling `commitRevision`. 
`AuthEnable` has already called `commitRevision`, `AuthDisable` also should call `commitRevision`.


serathius:
If you think the current behavior is incorrect, please add tests showing (both unit and e2e)

PR #6424, opened 2016-09-13
etcdserver: range queries with min/max create revision

heyitsanthony:
More API updates pursuant to #5412 / reducing sychronization message overhead.

PR #6411, opened 2016-09-12
etcdserver: Range with min/max mod revision

heyitsanthony:
Partially addresses #5412. Mainly in response to #6408 since the current Mutex code has some gross hacks to work around the lack of min/max mod rev. RWMutex would need to inherit those hacks to work, so might as well update the API so it can be done cleanly.

PR #18268, opened 2024-07-03
Ensure Count in GET response is consistent with key filtering by revision

### LINKED ISSUE #18267: Filtering a GET request via With{Min,Max}{Create,Mod}Rev results in wrong result Count
### Bug report criteria

### What happened?

```
root@cfs-campaign-1 18:58:58 ~
# etcdctl --write-out=fields get --prefix /k     # Show no keys with a "/k" prefix exist
"ClusterID" : 6736027128771979048
"MemberID" : 5334263768432527757
"Revision" : 1453
"RaftTerm" : 4
"More" : false
"Count" : 0

root@cfs-campaign-1 18:59:08 ~
# etcdctl --write-out=fields put /k1 v1        # Add /k1=v1 at revision 1454
"ClusterID" : 6736027128771979048
"MemberID" : 5334263768432527757
"Revision" : 1454
"RaftTerm" : 4

root@cfs-campaign-1 18:59:14 ~
# etcdctl --write-out=fields put /k2 v2        # Add /k2=v2 at revision 1455
"ClusterID" : 6736027128771979048
"MemberID" : 5334263768432527757
"Revision" : 1455
"RaftTerm" : 4

root@cfs-campaign-1 18:59:16 ~
# etcdctl --write-out=fields put /k3 v3        # Add /k3=v3 at revision 1456
"ClusterID" : 6736027128771979048
"MemberID" : 5334263768432527757
"Revision" : 1456
"RaftTerm" : 4

root@cfs-campaign-1 18:59:19 ~
# etcdctl --write-out=fields get --prefix --max-create-rev 1454 /k           # The count is wrong, should be 1.
"ClusterID" : 6736027128771979048
"MemberID" : 5334263768432527757
"Revision" : 1456
"RaftTerm" : 4
"Key" : "/k1"
"CreateRevision" : 1454
"ModRevision" : 1454
"Version" : 1
"Value" : "v1"
"Lease" : 0
"More" : false
"Count" : 3
```

`Count` in the GET request's result should be consistent with the size of the kvs array returned.

See "what happened" above.

The bug seems to be in `etc/server/etcdserver/txn/txn.go`.  When the filtering is applied:

  https://github.com/etcd-io/etcd/blob/37cbd6cd97f6ed6f7f6335bd661da051ae470089/server/etcdserver/txn/txn.go#L181

the implementation of `pruneKvs` does update the resulting array of kvs

  https://github.com/etcd-io/etcd/blob/37cbd6cd97f6ed6f7f6335bd661da051ae470089/server/etcdserver/txn/txn.go#L483

but the count is set without taking into account the actual returned kvs that may have been filtered:

  https://github.com/etcd-io/etcd/blob/37cbd6cd97f6ed6f7f6335bd661da051ae470089/server/etcdserver/txn/txn.go#L240

<details>

```console
# etcd --version
etcd Version: 3.5.12
Git SHA: e7b3bb6cc
Go Version: go1.20.13
Go OS/Arch: linux/amd64

# etcdctl version
API version: 3.6
```

</details>

<details>

```
name: etcd-1

data-dir: /var/lib/etcd/dh/cbd0f3c32

listen-client-urls: https://10.128.1.160:2379,https://localhost:2379
listen-peer-urls: https://10.128.1.160:2380,https://localhost:2380

advertise-client-urls: https://10.128.1.160:2379
initial-advertise-peer-urls: https://10.128.1.160:2380

initial-cluster-state: new
initial-cluster-token: cbd0f3c32
initial-cluster: etcd-1=https://10.128.1.160:2380

strict-reconfig-check: true
enable-v2: false

# * DHE worker processes start on demand and sometimes en-masse.
# * A newly elected etcd leader after failover needs to validate lots
#   of credentials for all the new clients showing up.
# In either of these circumstances, too high a cost to bcrypt
# (used during role based authentication) can starve a machine of CPU.
# Reduce the default of 10 to the min of 4.
bcrypt-cost: 4

peer-transport-security:
  # the peer certs need to have CN=peer.etcd.deephaven.local
  cert-allowed-cn: peer.etcd.deephaven.local
  client-cert-auth: true
  trusted-ca-file: /etc/etcd/dh/cbd0f3c32/ssl/peer/ca.crt
  cert-file: /etc/etcd/dh/cbd0f3c32/ssl/peer/etcd-1.public.crt
  key-file: /etc/etcd/dh/cbd0f3c32/ssl/peer/etcd-1.private.key

client-transport-security:
  # clients are encrypted, but not authenticated - we rely on username / password for auth
  client-cert-auth: false
  # don't need to set CA since we aren't authenticating clients
  # trusted-ca-file: /etc/etcd/dh/cbd0f3c32/ssl/server/ca.crt
  cert-file: /etc/etcd/dh/cbd0f3c32/ssl/server/etcd-1.public.crt
  key-file: /etc/etcd/dh/cbd0f3c32/ssl/server/etcd-1.private.key

auto-compaction-mode: periodic
auto-compaction-retention: "168"
```

</details>

<details>

```console
# etcdctl member list -w table
+------------------+---------+--------+---------------------------+---------------------------+------------+
|        ID        | STATUS  |  NAME  |        PEER ADDRS         |       CLIENT ADDRS        | IS LEARNER |
+------------------+---------+--------+---------------------------+---------------------------+------------+
| 4a071ca29fa8498d | started | etcd-1 | https://10.128.1.160:2380 | https://10.128.1.160:2379 |      false |
+------------------+---------+--------+---------------------------+---------------------------+------------+

# etcdctl endpoint status -w table
+---------------------------+------------------+---------+-----------------+---------+--------+-----------------------+-------+-----------+------------+-----------+------------+--------------------+--------+
|         ENDPOINT          |        ID        | VERSION | STORAGE VERSION | DB SIZE | IN USE | PERCENTAGE NOT IN USE | QUOTA | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS |
+---------------------------+------------------+---------+-----------------+---------+--------+-----------------------+-------+-----------+------------+-----------+------------+--------------------+--------+
| https://10.128.1.160:2379 | 4a071ca29fa8498d |  3.5.12 |                 |  6.1 MB | 6.0 MB |                    1% |   0 B |      true |      false |         4 |       2325 |               2325 |        |
+---------------------------+------------------+---------+-----------------+---------+--------+-----------------------+-------+-----------+------------+-----------+------------+--------------------+--------+
```

</details>

### Relevant log output

_No response_

jcferretti:
Fixes #18267

### DIFF
--- a/etcdctl/ctlv3/ctl.go
+++ b/etcdctl/ctlv3/ctl.go
@@ -52,7 +52,8 @@ func init() {
 	rootCmd.PersistentFlags().StringSliceVar(&globalFlags.Endpoints, "endpoints", []string{"127.0.0.1:2379"}, "gRPC endpoints")
 	rootCmd.PersistentFlags().BoolVar(&globalFlags.Debug, "debug", false, "enable client-side debug logging")
 
-	rootCmd.PersistentFlags().StringVarP(&globalFlags.OutputFormat, "write-out", "w", "simple", "set the output format (fields, json, protobuf, simple, table)")
+	rootCmd.PersistentFlags().StringVarP(&globalFlags.OutputFormat, "write-out", "w", "simple",
+		"set the output format (fields, json, protobuf, simple, table); note json encodes kvs as base64")
 	rootCmd.PersistentFlags().BoolVar(&globalFlags.IsHex, "hex", false, "print byte strings as hex encoded strings")
 	rootCmd.RegisterFlagCompletionFunc("write-out", func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
 		return []string{"fields", "json", "protobuf", "simple", "table"}, cobra.ShellCompDirectiveDefault
--- a/etcdutl/ctl.go
+++ b/etcdutl/ctl.go
@@ -35,7 +35,8 @@ var (
 )
 
 func init() {
-	rootCmd.PersistentFlags().StringVarP(&etcdutl.OutputFormat, "write-out", "w", "simple", "set the output format (fields, json, protobuf, simple, table)")
+	rootCmd.PersistentFlags().StringVarP(&etcdutl.OutputFormat, "write-out", "w", "simple",
+		"set the output format (fields, json, protobuf, simple, table); note json encodes kvs as base64")
 	rootCmd.RegisterFlagCompletionFunc("write-out", func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
 		return []string{"fields", "json", "protobuf", "simple", "table"}, cobra.ShellCompDirectiveDefault
 	})
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -152,9 +152,6 @@ func Range(ctx context.Context, lg *zap.Logger, kv mvcc.KV, r *pb.RangeRequest) 
 func executeRange(ctx context.Context, lg *zap.Logger, txnRead mvcc.TxnRead, r *pb.RangeRequest) (*pb.RangeResponse, error) {
 	trace := traceutil.Get(ctx)
 
-	resp := &pb.RangeResponse{}
-	resp.Header = &pb.ResponseHeader{}
-
 	limit := r.Limit
 	if r.SortOrder != pb.RangeRequest_NONE ||
 		r.MinModRevision != 0 || r.MaxModRevision != 0 ||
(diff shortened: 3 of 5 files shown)


jcferretti:
There have been no comments in this PR for two months, and on a separate PR a vague [comment](https://github.com/etcd-io/etcd/issues/18486#issuecomment-2328065736) was made about the appropriateness of the tests and that being the reason for it being held.

It would be good to hear, directly here, concrete points about what is believed to be lacking in the existing tests, and what else specifically would be important to have to move this PR forward.

Thanks.

jmhbnz:
Discussed during sig-etcd triage meeting, @ahrtr could you please take a look at this pr and provide some guidance on next steps, thanks!

ahrtr:
Thanks for the PR! It's a bit hard to follow, and it's also missing comprehensive e2e and integration tests.  But see more important comments below.

I think we have two approaches to fix it:
- Do not change the golang source code. Instead, we just need to update the document to clarify that `Count` reflects the number of keys within the range, without considering the filters (Min or Max, Create or Modify, Revisions).
https://github.com/etcd-io/etcd/blob/68e712276bb8a8f35f7f077add143b28422e8d66/api/etcdserverpb/rpc.proto#L510-L511
- If we really want the `Count` also reflects the filters, then we should get the filters (Min or Max, Create or Modify, Revisions) included in `RangeOptions`, see https://github.com/etcd-io/etcd/blob/68e712276bb8a8f35f7f077add143b28422e8d66/server/storage/mvcc/kv.go#L26-L30

    So the returned data already is already the filtered data. 
https://github.com/etcd-io/etcd/blob/68e712276bb8a8f35f7f077add143b28422e8d66/server/etcdserver/txn/txn.go#L176

    Obviously it has a little big impact on the most important part of etcdserver. We may even need to update the API `Revisions`. So it definitely needs super careful review. https://github.com/etcd-io/etcd/blob/68e712276bb8a8f35f7f077add143b28422e8d66/server/storage/mvcc/index.go#L27

@serathius What do you think?

BTW, FYI the filters were introduced in https://github.com/etcd-io/etcd/pull/6411 and https://github.com/etcd-io/etcd/pull/6424

jcferretti:
From my perspective I would be ok (and I think in favor) of just updating the documentation to be explicit on what count is returned in this case and then not changing the code (meaning closing this PR without implementing it).

All the options and what any particular combination means can get pretty complicated from a semantics perspective.  I think the API is already breaking the principle of least surprise.  I think that results from the fact that combinations do not really compose, each option I imagine ws incrementally added to do its own thing and over time we get weird combinations meaning things that are not used combined that frequently but if they are when both are provided what the result would be is not obvious.  Obvious I don't think changing the API is a reasonable option at this point, so I think a doc warning of "here be dragons" is good enough and we can all move on.

Thanks

jmhbnz:
Discussed during sig-etcd triage meeting, it looks like we don't have consensus to proceed with the code change and a docs change would be a safer way forward. Closing in favor of a docs update we can progress under the open issue.

PR #18213, opened 2024-06-22
Disable robustness test detection of #18089 to allow detecting other issues

serathius:
/cc @siyuanfoundation @MadhavJivrajani @fuweid @fykaa @henrybear327

PR #2894, opened 2015-05-31
Storage initial compaction

xiang90:
1. refactor key-index to support reversion (main, sub)

2. initial compaction support

PR #18274, opened 2024-07-03
*: keep tombstone if revision == compactAtRev

fuweid:
Before this patch, the tombstone can be deleted if its revision is equal compacted revision. It causes that the watch subscriber won't get this DELETE event. Based on Compact API [[1]], we should keep tombstone revision if it's not less than the compaction revision.

> CompactionRequest compacts the key-value store up to a given revision.
> All superseded keys with a revision less than the compaction revision
> will be removed.

[1]: https://etcd.io/docs/latest/dev-guide/api_reference_v3/

------------

To Reviewers:

During writing tests, I found that the compact doesn't always remove smaller revision data.
For example,

```
➜  bin/etcdctl put hello world -w json
{"header":{"cluster_id":14841639068965178418,"member_id":10276657743932975437,"revision":2,"raft_term":2}}

➜  bin/etcdctl put hello-v2 world -w json
{"header":{"cluster_id":14841639068965178418,"member_id":10276657743932975437,"revision":3,"raft_term":2}}

➜  bin/etcdctl put hello world-v2 -w json
{"header":{"cluster_id":14841639068965178418,"member_id":10276657743932975437,"revision":4,"raft_term":2}}

➜  bin/etcdctl get --rev 2 hello
hello
world

➜  bin/etcdctl compact 3 --physical=true
compacted revision 3

➜  bin/etcdctl get --rev 2 hello
{"level":"warn","ts":"2024-07-03T20:12:39.086991+0800","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0002105a0/127.0.0.1:2379","method":"/etcdserverpb.KV/Range","attempt":0,"error":"rpc error: code = OutOfRange desc = etcdserver: mvcc: required revision has been compacted"}
Error: etcdserver: mvcc: required revision has been compacted

➜  bin/tools/etcd-dump-db iterate-bucket ./default.etcd/member/snap/db key --decode
rev={Revision:{Main:4 Sub:0} tombstone:false}, value=[key "hello" | val "world-v2" | created 2 | mod 4 | ver 2]
rev={Revision:{Main:3 Sub:0} tombstone:false}, value=[key "hello-v2" | val "world" | created 3 | mod 3 | ver 1]
rev={Revision:{Main:2 Sub:0} tombstone:false}, value=[key "hello" | val "world" | created 2 | mod 2 | ver 1]
```

Not sure we should fix it in the follow-up. Just want to highlight it.

### DIFF
--- a/server/storage/mvcc/hash.go
+++ b/server/storage/mvcc/hash.go
@@ -63,6 +63,9 @@ func (h *kvHasher) WriteKeyValue(k, v []byte) {
 	if !upper.GreaterThan(kr) {
 		return
 	}
+
+	isTombstone := BytesToBucketKey(k).tombstone
+
 	lower := Revision{Main: h.compactRevision + 1}
 	// skip revisions that are scheduled for deletion
 	// due to compacting; don't skip if there isn't one.
@@ -71,6 +74,17 @@ func (h *kvHasher) WriteKeyValue(k, v []byte) {
 			return
 		}
 	}
+
+	// When performing compaction, if the compacted revision is a
+	// tombstone, older versions (<= 3.5.15 or <= 3.4.33) will delete
+	// the tombstone. But newer versions (> 3.5.15 or > 3.4.33) won't
+	// delete it. So we should skip the tombstone in such cases when
+	// computing the hash to ensure that both older and newer versions
+	// can always generate the same hash values.
+	if kr.Main == h.compactRevision && isTombstone {
+		return
+	}
+
 	h.hash.Write(k)
 	h.hash.Write(v)
 }
--- a/server/storage/mvcc/index_test.go
+++ b/server/storage/mvcc/index_test.go
@@ -18,7 +18,7 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/google/btree"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
 )
 
@@ -235,101 +235,445 @@ func TestIndexRevision(t *testing.T) {
 
 func TestIndexCompactAndKeep(t *testing.T) {
 	maxRev := int64(20)
-	tests := []struct {
-		key     []byte
-		remove  bool
-		rev     Revision
-		created Revision
-		ver     int64
-	}{
-		{[]byte("foo"), false, Revision{Main: 1}, Revision{Main: 1}, 1},
-		{[]byte("foo1"), false, Revision{Main: 2}, Revision{Main: 2}, 1},
-		{[]byte("foo2"), false, Revision{Main: 3}, Revision{Main: 3}, 1},
-		{[]byte("foo2"), false, Revision{Main: 4}, Revision{Main: 3}, 2},
-		{[]byte("foo"), false, Revision{Main: 5}, Revision{Main: 1}, 2},
-		{[]byte("foo1"), false, Revision{Main: 6}, Revision{Main: 2}, 2},
-		{[]byte("foo1"), true, Revision{Main: 7}, Revision{}, 0},
-		{[]byte("foo2"), true, Revision{Main: 8}, Revision{}, 0},
-		{[]byte("foo"), true, Revision{Main: 9}, Revision{}, 0},
-		{[]byte("foo"), false, Revision{Main: 10}, Revision{Main: 10}, 1},
-		{[]byte("foo1"), false, Revision{Main: 10, Sub: 1}, Revision{Main: 10, Sub: 1}, 1},
+
+	// key: "foo"
+	// modified: 10
+	// generations:
+	//	{{10, 0}}
+	//	{{1, 0}, {5, 0}, {9, 0}(t)}
+	//
+	// key: "foo1"
+	// modified: 10, 1
+	// generations:
+	//	{{10, 1}}
+	//	{{2, 0}, {6, 0}, {7, 0}(t)}
+	//
+	// key: "foo2"
+	// modified: 8
+	// generations:
+	//	{empty}
+	//	{{3, 0}, {4, 0}, {8, 0}(t)}
+	//
+	buildTreeIndex := func() index {
+		ti := newTreeIndex(zaptest.NewLogger(t))
+
+		ti.Put([]byte("foo"), Revision{Main: 1})
+		ti.Put([]byte("foo1"), Revision{Main: 2})
+		ti.Put([]byte("foo2"), Revision{Main: 3})
+		ti.Put([]byte("foo2"), Revision{Main: 4})
+		ti.Put([]byte("foo"), Revision{Main: 5})
+		ti.Put([]byte("foo1"), Revision{Main: 6})
+		require.NoError(t, ti.Tombstone([]byte("foo1"), Revision{Main: 7}))
+		require.NoError(t, ti.Tombstone([]byte("foo2"), Revision{Main: 8}))
+		require.NoError(t, ti.Tombstone([]byte("foo"), Revision{Main: 9}))
+		ti.Put([]byte("foo"), Revision{Main: 10})
+		ti.Put([]byte("foo1"), Revision{Main: 10, Sub: 1})
+		return ti
 	}
 
-	// Continuous Compact and Keep
-	ti := newTreeIndex(zaptest.NewLogger(t))
-	for _, tt := range tests {
-		if tt.remove {
-			ti.Tombstone(tt.key, tt.rev)
-		} else {
-			ti.Put(tt.key, tt.rev)
-		}
+	afterCompacts := []struct {
+		atRev      int
+		keyIndexes []keyIndex
+		keep       map[Revision]struct{}
+		compacted  map[Revision]struct{}
+	}{
+		{
+			atRev: 1,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 1}, Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 2}, Revision{Main: 6}, Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 3}, Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+			},
+		},
+		{
+			atRev: 2,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 1}, Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 2}, Revision{Main: 6}, Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 3}, Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+				Revision{Main: 2}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+				Revision{Main: 2}: {},
+			},
+		},
+		{
+			atRev: 3,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 1}, Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 2}, Revision{Main: 6}, Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 3}, Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+				Revision{Main: 2}: {},
+				Revision{Main: 3}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+				Revision{Main: 2}: {},
+				Revision{Main: 3}: {},
+			},
+		},
+		{
+			atRev: 4,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 1}, Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 2}, Revision{Main: 6}, Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+				Revision{Main: 2}: {},
+				Revision{Main: 4}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 1}: {},
+				Revision{Main: 2}: {},
+				Revision{Main: 4}: {},
+			},
+		},
+		{
+			atRev: 5,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 2}, Revision{Main: 6}, Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 2}: {},
+				Revision{Main: 4}: {},
+				Revision{Main: 5}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 2}: {},
+				Revision{Main: 4}: {},
+				Revision{Main: 5}: {},
+			},
+		},
+		{
+			atRev: 6,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 6}, Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 6}: {},
+				Revision{Main: 4}: {},
+				Revision{Main: 5}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 6}: {},
+				Revision{Main: 4}: {},
+				Revision{Main: 5}: {},
+			},
+		},
+		{
+			atRev: 7,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 2}, revs: []Revision{Revision{Main: 7}}},
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 4}, Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 4}: {},
+				Revision{Main: 5}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 7}: {},
+				Revision{Main: 4}: {},
+				Revision{Main: 5}: {},
+			},
+		},
+		{
+			atRev: 8,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 5}, Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+				{
+					key:      []byte("foo2"),
+					modified: Revision{Main: 8},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 3}, revs: []Revision{Revision{Main: 8}}},
+						{},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 5}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 8}: {},
+				Revision{Main: 5}: {},
+			},
+		},
+		{
+			atRev: 9,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 3, created: Revision{Main: 1}, revs: []Revision{Revision{Main: 9}}},
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 9}: {},
+			},
+		},
+		{
+			atRev: 10,
+			keyIndexes: []keyIndex{
+				{
+					key:      []byte("foo"),
+					modified: Revision{Main: 10},
+					generations: []generation{
+						{ver: 1, created: Revision{Main: 10}, revs: []Revision{Revision{Main: 10}}},
+					},
+				},
+				{
+					key:      []byte("foo1"),
+					modified: Revision{Main: 10, Sub: 1},
+					generations: []generation{
+						{ver: 1, created: Revision{Main: 10, Sub: 1}, revs: []Revision{Revision{Main: 10, Sub: 1}}},
+					},
+				},
+			},
+			keep: map[Revision]struct{}{
+				Revision{Main: 10}:         {},
+				Revision{Main: 10, Sub: 1}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 10}:         {},
+				Revision{Main: 10, Sub: 1}: {},
+			},
+		},
 	}
+
+	ti := buildTreeIndex()
+	// Continuous Compact and Keep
 	for i := int64(1); i < maxRev; i++ {
+		j := i - 1
+		if i >= int64(len(afterCompacts)) {
+			j = int64(len(afterCompacts)) - 1
+		}
+
 		am := ti.Compact(i)
+		require.Equal(t, afterCompacts[j].compacted, am, "#%d: compact(%d) != expected", i, i)
+
 		keep := ti.Keep(i)
-		if !(reflect.DeepEqual(am, keep)) {
-			t.Errorf("#%d: compact keep %v != Keep keep %v", i, am, keep)
+		require.Equal(t, afterCompacts[j].keep, keep, "#%d: keep(%d) != expected", i, i)
+
+		nti := newTreeIndex(zaptest.NewLogger(t)).(*treeIndex)
+		for k := range afterCompacts[j].keyIndexes {
+			ki := afterCompacts[j].keyIndexes[k]
+			nti.tree.ReplaceOrInsert(&ki)
 		}
-		wti := &treeIndex{tree: btree.NewG(32, func(aki *keyIndex, bki *keyIndex) bool {
-			return aki.Less(bki)
-		})}
-		for _, tt := range tests {
-			if _, ok := am[tt.rev]; ok || tt.rev.GreaterThan(Revision{Main: i}) {
-				if tt.remove {
-					wti.Tombstone(tt.key, tt.rev)
-				} else {
-					restore(wti, tt.key, tt.created, tt.rev, tt.ver)
-				}
-			}
-		}
-		if !ti.Equal(wti) {
-			t.Errorf("#%d: not equal ti", i)
-		}
+		require.True(t, ti.Equal(nti), "#%d: not equal ti", i)
 	}
 
 	// Once Compact and Keep
 	for i := int64(1); i < maxRev; i++ {
-		ti := newTreeIndex(zaptest.NewLogger(t))
-		for _, tt := range tests {
-			if tt.remove {
-				ti.Tombstone(tt.key, tt.rev)
-			} else {
-				ti.Put(tt.key, tt.rev)
-			}
+		ti := buildTreeIndex()
+
+		j := i - 1
+		if i >= int64(len(afterCompacts)) {
+			j = int64(len(afterCompacts)) - 1
 		}
+
 		am := ti.Compact(i)
+		require.Equal(t, afterCompacts[j].compacted, am, "#%d: compact(%d) != expected", i, i)
+
 		keep := ti.Keep(i)
-		if !(reflect.DeepEqual(am, keep)) {
-			t.Errorf("#%d: compact keep %v != Keep keep %v", i, am, keep)
-		}
-		wti := &treeIndex{tree: btree.NewG(32, func(aki *keyIndex, bki *keyIndex) bool {
-			return aki.Less(bki)
-		})}
-		for _, tt := range tests {
-			if _, ok := am[tt.rev]; ok || tt.rev.GreaterThan(Revision{Main: i}) {
-				if tt.remove {
-					wti.Tombstone(tt.key, tt.rev)
-				} else {
-					restore(wti, tt.key, tt.created, tt.rev, tt.ver)
-				}
-			}
-		}
-		if !ti.Equal(wti) {
-			t.Errorf("#%d: not equal ti", i)
-		}
-	}
-}
+		require.Equal(t, afterCompacts[j].keep, keep, "#%d: keep(%d) != expected", i, i)
 
-func restore(ti *treeIndex, key []byte, created, modified Revision, ver int64) {
-	keyi := &keyIndex{key: key}
+		nti := newTreeIndex(zaptest.NewLogger(t)).(*treeIndex)
+		for k := range afterCompacts[j].keyIndexes {
+			ki := afterCompacts[j].keyIndexes[k]
+			nti.tree.ReplaceOrInsert(&ki)
+		}
 
-	ti.Lock()
-	defer ti.Unlock()
-	okeyi, _ := ti.tree.Get(keyi)
-	if okeyi == nil {
-		keyi.restore(ti.lg, created, modified, ver)
-		ti.tree.ReplaceOrInsert(keyi)
-		return
+		require.True(t, ti.Equal(nti), "#%d: not equal ti", i)
 	}
-	okeyi.put(ti.lg, modified.Main, modified.Sub)
 }
--- a/server/storage/mvcc/key_index.go
+++ b/server/storage/mvcc/key_index.go
@@ -65,7 +65,8 @@ var (
 // compact(5):
 // generations:
 //
-//	{empty} -> key SHOULD be removed.
+//	{empty}
+//	{5.0(t)}
 //
 // compact(6):
 // generations:
@@ -202,8 +203,7 @@ func (ki *keyIndex) since(lg *zap.Logger, rev int64) []Revision {
 }
 
 // compact compacts a keyIndex by removing the versions with smaller or equal
-// revision than the given atRev except the largest one (If the largest one is
-// a tombstone, it will not be kept).
+// revision than the given atRev except the largest one.
 // If a generation becomes empty during compaction, it will be removed.
 func (ki *keyIndex) compact(lg *zap.Logger, atRev int64, available map[Revision]struct{}) {
 	if ki.isEmpty() {
@@ -221,11 +221,6 @@ func (ki *keyIndex) compact(lg *zap.Logger, atRev int64, available map[Revision]
 		if revIndex != -1 {
 			g.revs = g.revs[revIndex:]
 		}
-		// remove any tombstone
-		if len(g.revs) == 1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[0])
-			genIdx++
-		}
 	}
 
 	// remove the previous generations.
@@ -241,7 +236,12 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
 	genIdx, revIndex := ki.doCompact(atRev, available)
 	g := &ki.generations[genIdx]
 	if !g.isEmpty() {
-		// remove any tombstone
+		// If the given `atRev` is a tombstone, we need to skip it.
+		//
+		// Note that this s different from the `compact` function which
+		// keeps tombstone in such case. We need to stay consistent with
+		// existing versions, ensuring they always generate the same hash
+		// values.
 		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
 			delete(available, g.revs[revIndex])
 		}
@@ -262,7 +262,7 @@ func (ki *keyIndex) doCompact(atRev int64, available map[Revision]struct{}) (gen
 	genIdx, g := 0, &ki.generations[0]
 	// find first generation includes atRev or created after atRev
 	for genIdx < len(ki.generations)-1 {
-		if tomb := g.revs[len(g.revs)-1].Main; tomb > atRev {
+		if tomb := g.revs[len(g.revs)-1].Main; tomb >= atRev {
 			break
 		}
 		genIdx++
--- a/server/storage/mvcc/key_index_test.go
+++ b/server/storage/mvcc/key_index_test.go
@@ -18,6 +18,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 )
@@ -308,12 +309,15 @@ func TestKeyIndexCompactAndKeep(t *testing.T) {
 				key:      []byte("foo"),
 				modified: Revision{Main: 16},
 				generations: []generation{
+					{created: Revision{Main: 2}, ver: 3, revs: []Revision{Revision{Main: 6}}},
 					{created: Revision{Main: 8}, ver: 3, revs: []Revision{Revision{Main: 8}, Revision{Main: 10}, Revision{Main: 12}}},
 					{created: Revision{Main: 14}, ver: 3, revs: []Revision{Revision{Main: 14}, Revision{Main: 15, Sub: 1}, Revision{Main: 16}}},
 					{},
 				},
 			},
-			map[Revision]struct{}{},
+			map[Revision]struct{}{
+				Revision{Main: 6}: {},
+			},
 		},
 		{
 			7,
@@ -394,11 +398,14 @@ func TestKeyIndexCompactAndKeep(t *testing.T) {
 				key:      []byte("foo"),
 				modified: Revision{Main: 16},
 				generations: []generation{
+					{created: Revision{Main: 8}, ver: 3, revs: []Revision{Revision{Main: 12}}},
 					{created: Revision{Main: 14}, ver: 3, revs: []Revision{Revision{Main: 14}, Revision{Main: 15, Sub: 1}, Revision{Main: 16}}},
 					{},
 				},
 			},
-			map[Revision]struct{}{},
+			map[Revision]struct{}{
+				Revision{Main: 12}: {},
+			},
 		},
 		{
 			13,
@@ -446,6 +453,20 @@ func TestKeyIndexCompactAndKeep(t *testing.T) {
 				key:      []byte("foo"),
 				modified: Revision{Main: 16},
 				generations: []generation{
+					{created: Revision{Main: 14}, ver: 3, revs: []Revision{Revision{Main: 16}}},
+					{},
+				},
+			},
+			map[Revision]struct{}{
+				Revision{Main: 16}: {},
+			},
+		},
+		{
+			17,
+			&keyIndex{
+				key:      []byte("foo"),
+				modified: Revision{Main: 16},
+				generations: []generation{
 					{},
 				},
 			},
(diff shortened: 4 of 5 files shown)


fuweid:
ping @ahrtr @serathius @siyuanfoundation @chaochn47 ~ PTAL, thanks

ahrtr:
> ```
> ➜  bin/tools/etcd-dump-db iterate-bucket ./default.etcd/member/snap/db key --decode
> rev={Revision:{Main:4 Sub:0} tombstone:false}, value=[key "hello" | val "world-v2" | created 2 | mod 4 | ver 2]
> rev={Revision:{Main:3 Sub:0} tombstone:false}, value=[key "hello-v2" | val "world" | created 3 | mod 3 | ver 1]
> rev={Revision:{Main:2 Sub:0} tombstone:false}, value=[key "hello" | val "world" | created 2 | mod 2 | ver 1]
> ```
> 
> 
>     
>   
> 
> Not sure we should fix it in the follow-up. Just want to highlight it.

I think it's expected behavior. Compaction always keeps the latest revision K, which is <= compaction_rev.

ahrtr:
The e2e test cases, especially the mix-version test cases, as mentioned in https://github.com/etcd-io/etcd/pull/18162#issuecomment-2168322174 are important. What's the plan to implement the e2e test cases?

fuweid:
> I think it's expected behavior. Compaction always keeps the latest revision K, which is <= compaction_rev.

ETCD server prevents requests from fetching revisions which are smaller than compacted revision. It looks fine to me.
However, based on the compact API, if the revision is not latest and smaller than the compaction revision, that revision should be deleted.

> The e2e test cases, especially the mix-version test cases, as mentioned in https://github.com/etcd-io/etcd/pull/18162#issuecomment-2168322174 are important.

For the mix-version part, it's hard to finish it in one request. I'm trying to use patch mode in this pull request, like https://github.com/etcd-io/etcd/blob/main/tests/robustness/patches/beforeSendWatchResponse/watch.patch.

ahrtr:
> ETCD server prevents requests from fetching revisions which are smaller than compacted revision. It looks fine to me.
> However, based on the compact API, if the revision is not latest and smaller than the compaction revision, that revision should be deleted

It sounds reasonable, but we still have two options, either keep it as it's or fix it.

The rev, which is smaller than compaction rev, isn't accessible via the get/range API, but it's still in the db file. I agree it's an issue, but it seems harmless. Fixing it may break user experience as well. 

If we decide to fix it (let's go for this direction if there is objection), 
- we should fix it in one PR instead of two.
- evaluate the effort & impact.



> For the mix-version part, it's hard to finish it in one request.

It's OK to do it in followup PR, but we should manually verify it before we merge this PR. Please also feedback if you need help from others.

server/storage/mvcc/key_index.go
```diff
@@ -205,6 +206,18 @@ func (ki *keyIndex) since(lg *zap.Logger, rev int64) []Revision {
 // revision than the given atRev except the largest one (If the largest one is
 // a tombstone, it will not be kept).
 // If a generation becomes empty during compaction, it will be removed.
+//
+// FIXME(fuweid): There is an exception that compact function won't remove
```
siyuanfoundation, line 210: why is this an exception? `removing the versions with smaller or equal
revision than the given atRev except the largest one` means  revision{1,0} would not be deleted because it is the largest one <= compactRev

server/storage/mvcc/key_index.go
```diff
@@ -205,6 +206,18 @@ func (ki *keyIndex) since(lg *zap.Logger, rev int64) []Revision {
 // revision than the given atRev except the largest one (If the largest one is
 // a tombstone, it will not be kept).
```
siyuanfoundation, line 207: Remove the comment of `(If the largest one is
// a tombstone, it will not be kept).`

server/storage/mvcc/key_index.go
```diff
@@ -238,20 +246,15 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
...
-	if !g.isEmpty() {
-		// remove any tombstone
-		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[revIndex])
-		}
-	}
+	ki.doCompact(atRev, available)
 }
 
 func (ki *keyIndex) doCompact(atRev int64, available map[Revision]struct{}) (genIdx int, revIndex int) {
```
siyuanfoundation, line 252: nit: can you add documentation for this `doCompact` function about what it returns since we are here?

server/storage/mvcc/index_test.go
siyuanfoundation, line 1: Can you add the tests in a 1 first commit so that it is easier to see what the tombstone change does to the final results?

fuweid: Sure. Updated

server/storage/mvcc/key_index.go
fuweid: I left it empty because the `compact` function has comment. :)

server/storage/mvcc/key_index.go
fuweid: Done.

fuweid:
ping @ahrtr I updated the pull request with new e2e cases. The `TestVerifyHashKVAfterCompactMixVersion` is to test this change in a mix-version cluster (1 x main process + 2 x v3.5.14 process). One case will cause inconsistent hash value and I left comment in the case. Please take a look when you have time. Thanks

client/pkg/testutil/leak.go
```diff
@@ -155,6 +155,7 @@ func interestingGoroutines() (gs []string) {
 				"rcrypto/internal/boring.(*PublicKeyRSA).finalize",
 				"net.(*netFD).Close(",
 				"testing.(*T).Run",
+				"crypto/tls.(*certCache).evict",
```
fuweid, line 158:
```
=== RUN   TestVerifyHashKVAfterCompactMixVersion/#0
    leak.go:103: Found leaked goroutined BEFORE test appears to have leaked :
        sync.(*Map).LoadAndDelete(0xc00031e180, {0xe07320, 0xc00009fde0})
                /usr/local/go/src/sync/map.go:272 +0x192
        sync.(*Map).Delete(...)
                /usr/local/go/src/sync/map.go:297
        crypto/tls.(*certCache).evict(...)
                /usr/local/go/src/crypto/tls/cache.go:73
        crypto/tls.(*certCache).active.func1(0x0?)
                /usr/local/go/src/crypto/tls/cache.go:65 +0x67
```

The GC invokes the deletion. We should skip it.

ahrtr: Please resolve this in a separate PR.

server/storage/mvcc/key_index.go
```diff
@@ -238,20 +233,15 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
...
 	// walk until reaching the first revision smaller or equal to "atRev",
 	// and add the revision to the available map
 	f := func(rev Revision) bool {
+		// NOTE: Duplicate keys in one txn isn't allowed. So, there is
+		// no need to check Sub revision here.
```
ahrtr, line 244: This comment is confusing. Suggest to remove it to avoid confusion.

server/storage/mvcc/key_index.go
```diff
@@ -238,20 +233,15 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
...
 	}
 
-	genIdx, revIndex := ki.doCompact(atRev, available)
-	g := &ki.generations[genIdx]
-	if !g.isEmpty() {
-		// remove any tombstone
-		if revIndex == len(g.revs)-1 && genIdx != len(ki.generations)-1 {
-			delete(available, g.revs[revIndex])
-		}
-	}
```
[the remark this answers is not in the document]
fuweid: The behaviour was introduced in https://github.com/etcd-io/etcd/pull/2894/files#diff-150706d0e34ff6ac7c79d761ab4317aea899365f194b90f1667e1e009bb38e23. The tombstone will be deleted during compact.

I added test case for mix-version cluster. Found one issue if there is only one key. Please checkout the  `TestVerifyHashKVAfterCompactMixVersion` in this pull request.

server/storage/mvcc/key_index.go
fuweid: Updated.

server/storage/mvcc/key_index_test.go
```diff
@@ -687,6 +707,10 @@ func TestGenerationWalk(t *testing.T) {
 	}
 }
 
+// TODO(fuweid):
+//
+// Replace {14, 1} with {15, 1} because ETCD server doesn't allow duplicate
+// keys in one txn.
```
ahrtr, line 713: Can you fix this firstly so that we have meaningful test cases?
Please let me know if you need any help on this.

ahrtr: Please do it in a separate PR.

tests/e2e/watch_test.go
```diff
@@ -311,24 +315,370 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+
+func TestVerifyHashKVAfterCompactMixVersion(t *testing.T) {
+	if !fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
+		t.Skipf("%q does not exist", e2e.BinPath.EtcdLastRelease)
+	}
+
+	tests := []struct {
+		compactedOnTombstoneRev bool
+		hashKVOnCompactedRev    bool
+		consistentHash          bool
```
ahrtr, line 402: It's confusing, why do we need this field? All etcd members should always return the same hashKV on the same rev, no matter it's mix-version or not.

ahrtr:
Please put all hashKV related e2e cases into a separate file, e.g. `hashkv_test.go`

tests/e2e/watch_test.go
ahrtr: Please ping me on slack if you want discussion.

ahrtr:
Discussed with @fuweid today. The existing implementation of HashByRev and Compaction are tightly coupled, and also they don't meet the expected behavior (see below). 

The draft expected behavior is recorded in the [doc](https://docs.google.com/document/d/1APJE38DxENsRLLVapRz1CQ6UD4-uYaFutO12Y01VcNw/edit#heading=h.n6mard91upel). See summary below as well. Please comment in the [doc](https://docs.google.com/document/d/1APJE38DxENsRLLVapRz1CQ6UD4-uYaFutO12Y01VcNw/edit#heading=h.n6mard91upel) to ensure we have consensus.


 ## HashByRev (rev)

- All revisions up to $(rev) should be included/calculated. For example, HashByRev(9), all revisions, which <=9, should be included/calculated.
- All members should always return the same hash,  no matter if it's a mix-version cluster or not.


**Concern**
Compatibility is a big concern.  it's hard to ensure all members always return the same hash in a mix-version cluster.

**Solution**: use cluster-level feature-gate to resolve this concern. The fix/change will not be applied until all members have enabled the feature.

## Compaction
When compacting revision X, 

- compact all history up to X (exclusive), but we always keep the revision X.
- For all revisions which are < X, always keep the latest non-tombstone revision for each key.

It means that users can start to read Revision X and watch starting from X after the compaction. Reading or Watching X-1 will get an `ErrCompacted` error.

For example, compact (9),
- Revision 9 will be kept;
- For all revisions, which < 9
  - If it’s the latest non-tombstone revision for a key, then keep it;
  - Otherwise, compact the revisions

## Next step
- Resolve https://github.com/etcd-io/etcd/pull/18274#discussion_r1672070530 in a separate PR;
- Ensure we have consensus on the expected behavior. Please comment on the [doc](https://docs.google.com/document/d/1APJE38DxENsRLLVapRz1CQ6UD4-uYaFutO12Y01VcNw/edit#bookmark=id.rluiv8kyg3nh);
- Make change per the agreed expected behavior, but it should can be enabled or disabled by a flag. Once the cluster-level feature is ready, migrate the flag to the cluster-level feature.

fuweid:
ping @ahrtr @serathius @siyuanfoundation @chaochn47 

Updated the pull request to make sure that it's compatible with existing releases for HashKV API.
It's ready to review. PTAL. Thanks

serathius:
Can we merge the e2e test before we make changes? This should give us more confidence about the behavior not changing.

ahrtr:
> Can we merge the e2e test before we make changes? This should give us more confidence about the behavior not changing.

hash related e2e test should be OK to be merged firstly as long as the behavior doesn't change. But the watch related e2e cases will fail without the fix.

Since we need to backport the PR to 3.5 and 3.4, so we have to keep the behavior consistent and minimize the impact. But in order to improve the long-term understandability & maintainability, we still need to have a clear documented expected behavior, and accordingly might need some sort of refactoring, roughly as https://github.com/etcd-io/etcd/pull/18274#issuecomment-2223188506 summarises.

serathius:
Can you remove usage of https://github.com/etcd-io/etcd/blob/4488f2c9b6aa4f3f97cd231f3378e53f777d3212/tests/robustness/validate/watch.go#L345-L372 from robustness test to confirm that PR resolves the correctness issue?

fuweid:
In my local, I run into error. checking

```
raftAfterSave=panic()

    logger.go:146: 2024-08-01T21:32:57.617+0800 ERROR   Broke watch guarantee   {"guarantee": "reliable", "client": 2, "diff": "  []model.PersistedEvent{\n  \t... // 584 identical elements\n  \t{Event: {Type: \"put-operation\", Key: \"key0\", Value: {Value: \"508\"}}, Revision: 528},\n  \t{Event: {Type: \"put-operation\", Key: \"key1\", Value: {Value: \"509\"}}, Revision: 529},\n- \t{Event: model.Event{Type: \"delete-operation\", Key: \"key7\"}, Revision: 530},\n  \t{Event: {Type: \"put-operation\", Key: \"key1\", Value: {Value: \"510\"}}, Revision: 530},\n  \t{Event: {Type: \"put-operation\", Key: \"key4\", Value: {Value: \"511\"}}, Revision: 530},\n  \t... // 348 identical elements\n  }\n"}
=== NAME  TestRobustnessExploratory/EtcdHighTraffic/ClusterOfSize1/raftAfterSave=panic()
    validate.go:49: Failed validating watch history, err: broke Reliable - a sequence of events will never drop any subsequence of events; if there are events ordered in time as a < b < c, then if the watch receives events a and c, it is guaranteed to receive b
=== NAME  TestRobustnessExploratory/EtcdHighTraffic/ClusterOfSize1
```

serathius:
@fuweid Can you provide a report and command you used to get failure? I'm trying to reproduce it.

fuweid:
> @fuweid Can you provide a report and command or changed you used to get the failure? I'm failing to reproduce it.

Sorry for late reply. I just run the command like this. I was thinking I can check the result today. But it failed in few minutes.

```
$ EXPECT_DEBUG=true GO_TEST_FLAGS="-v --count 200 --timeout 10h --failfast" RESULTS_DIR=/tmp/results make test-robustness-main
```

serathius:
Hmm, please don't run `test-robustness-main` as it includes testing hybrid cluster versions like v3.5. The issue might not come from your change, but from still unpatched release-3.5 branch. 

cc @jmhbnz @siyuanfoundation @MadhavJivrajani 
We might need to consider making it clearer which robustness test command should be used when. Documenting and renaming to be more informative should help.

fuweid:
> Hmm, please don't run test-robustness-main as it includes testing hybrid cluster versions like v3.5. The issue might not come from your change, but from still unpatched release-3.5 branch.

OK. I started `EXPECT_DEBUG=true GO_TEST_FLAGS="-v --count 200 --timeout 10h --failfast" RESULTS_DIR=/tmp/results make test-robustness` in my local. Will share the result later.

serathius:
> OK. I started EXPECT_DEBUG=true GO_TEST_FLAGS="-v --count 200 --timeout 10h --failfast" RESULTS_DIR=/tmp/results make test-robustness in my local. Will share the result later.

Please consider estimating the rate of reproduction of the issue via reverting https://github.com/etcd-io/etcd/pull/18213 with and without your change. This should give you a quicker answer then running the tests for 10 hours :P

fuweid:
I run 200 times without errors.

```
--- PASS: TestRobustnessRegression (22.69s)
    --- SKIP: TestRobustnessRegression/Issue14370 (0.86s)
    --- SKIP: TestRobustnessRegression/Issue14685 (0.85s)
    --- PASS: TestRobustnessRegression/Issue13766 (6.97s)
    --- PASS: TestRobustnessRegression/Issue15220 (4.29s)
    --- SKIP: TestRobustnessRegression/Issue17529 (0.85s)
    --- PASS: TestRobustnessRegression/Issue15271 (8.85s)
PASS
ok      go.etcd.io/etcd/tests/v3/robustness     7812.456s
'robustness' PASSED and completed at Fri Aug  2 09:22:04 PM CST 2024
SUCCESS
➜  etcd git:(fix-event-loss-caused-by-compact) EXPECT_DEBUG=true GO_TEST_FLAGS="-v -count 200 -timeout 10h -failfast" RESULTS_DIR=/tmp/results make test-robustness
```

I revert the commit #18213 in main branch. But I didn't reproduce the issue 😂

serathius:
Interesting try config from https://github.com/etcd-io/etcd/pull/18145 and only run ClusterOfSize1 with EtcdTrafficDeleteLeases traffic

fuweid:
I reproduced this issue without this patch and #18213.

```
  logger.go:146: 2024-08-04T17:56:06.865+0800 INFO    Finished traffic
    logger.go:146: 2024-08-04T17:56:07.871+0800 INFO    Reporting complete traffic      {"successes": 616, "failures": 22, "successRate": 0.9655172413793104, "period": "2.151034471s", "qps": 286.3738393339769}
    logger.go:146: 2024-08-04T17:56:07.871+0800 INFO    Reporting traffic before failure injection      {"successes": 401, "failures": 2, "successRate": 0.9950372208436724, "period": "999.359697ms", "qps": 401.2569260134972}
    logger.go:146: 2024-08-04T17:56:07.871+0800 INFO    Reporting traffic during failure injection      {"successes": 4, "failures": 5, "successRate": 0.4444444444444444, "period": "149.615927ms", "qps": 26.73512158902708}
    logger.go:146: 2024-08-04T17:56:07.871+0800 INFO    Reporting traffic after failure injection       {"successes": 211, "failures": 15, "successRate": 0.9336283185840708, "period": "1.002058847s", "qps": 210.56647584291025}
    logger.go:146: 2024-08-04T17:56:07.871+0800 INFO    Finished simulating traffic     {"max-revision": 214}
    logger.go:146: 2024-08-04T17:56:07.878+0800 INFO    Validating linearizable operations      {"timeout": "5m0s"}
    logger.go:146: 2024-08-04T17:56:08.013+0800 INFO    Linearization success   {"duration": "134.264413ms"}
    logger.go:146: 2024-08-04T17:56:08.013+0800 INFO    Validating watch
    logger.go:146: 2024-08-04T17:56:08.015+0800 ERROR   Broke watch guarantee   {"guarantee": "reliable", "client": 2, "diff": "  []model.PersistedEvent{\n  \t... // 160 identical elements\n  \t{Event: {Type: \"put-operation\", Key: \"key7\", Value: {Value: \"125\"}}, Revision: 141},\n  \t{Event: {Type: \"delete-operation\", Key: \"key4\"}, Revision: 142},\n- \t{Event: model.Event{Type: \"delete-operation\", Key: \"key7\"}, Revision: 143},\n  \t{Event: {Type: \"put-operation\", Key: \"key4\", Value: {Value: \"128\"}}, Revision: 144, ...},\n  \t{Event: {Type: \"put-operation\", Key: \"key7\", Value: {Value: \"129\"}}, Revision: 145, ...},\n  \t... // 75 identical elements\n  }\n"}
=== NAME  TestRobustnessExploratory/EtcdTrafficDeleteLeases/ClusterOfSize1/raftAfterSave=panic()
    validate.go:49: Failed validating watch history, err: broke Reliable - a sequence of events will never drop any subsequence of events; if there are events ordered in time as a < b < c, then if the watch receives events a and c, it is guaranteed to receive b
=== NAME  TestRobustnessExploratory/EtcdTrafficDeleteLeases/ClusterOfSize1
    logger.go:146: 2024-08-04T17:56:08.015+0800 INFO    Validating serializable operations
    logger.go:146: 2024-08-04T17:56:08.017+0800 INFO    Saving robustness test report   {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665"}
    logger.go:146: 2024-08-04T17:56:08.017+0800 INFO    Saving member data dir  {"member": "TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0", "path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/server-TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0"}
    logger.go:146: 2024-08-04T17:56:08.017+0800 INFO    no watch operations for client, skip persisting {"client-id": 1}
    logger.go:146: 2024-08-04T17:56:08.017+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-1/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.018+0800 INFO    Saving watch operations {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-2/watch.json"}
    logger.go:146: 2024-08-04T17:56:08.018+0800 INFO    no KV operations for client, skip persisting    {"client-id": 2}
    logger.go:146: 2024-08-04T17:56:08.018+0800 INFO    no watch operations for client, skip persisting {"client-id": 3}
    logger.go:146: 2024-08-04T17:56:08.018+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-3/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.018+0800 INFO    no watch operations for client, skip persisting {"client-id": 4}
    logger.go:146: 2024-08-04T17:56:08.018+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-4/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.019+0800 INFO    no watch operations for client, skip persisting {"client-id": 5}
    logger.go:146: 2024-08-04T17:56:08.019+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-5/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.019+0800 INFO    no watch operations for client, skip persisting {"client-id": 6}
    logger.go:146: 2024-08-04T17:56:08.019+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-6/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.019+0800 INFO    no watch operations for client, skip persisting {"client-id": 7}
    logger.go:146: 2024-08-04T17:56:08.019+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-7/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    no watch operations for client, skip persisting {"client-id": 8}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-8/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    no watch operations for client, skip persisting {"client-id": 9}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-9/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    no watch operations for client, skip persisting {"client-id": 10}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    Saving operation history        {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/client-10/operations.json"}
    logger.go:146: 2024-08-04T17:56:08.020+0800 INFO    Saving visualization    {"path": "/tmp/results/TestRobustnessExploratory_EtcdTrafficDeleteLeases_ClusterOfSize1_raftAfterSave=panic()/1722765368017800665/history.html"}
    logger.go:146: 2024-08-04T17:56:08.040+0800 INFO    killing server...       {"name": "TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0"}
    logger.go:146: 2024-08-04T17:56:08.040+0800 INFO    stopping server...      {"name": "TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0"}
    logger.go:146: 2024-08-04T17:56:08.043+0800 INFO    stopped server. {"name": "TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0"}
--- FAIL: TestRobustnessExploratory (4.00s)
    --- FAIL: TestRobustnessExploratory/EtcdTrafficDeleteLeases/ClusterOfSize1 (3.99s)
        --- FAIL: TestRobustnessExploratory/EtcdTrafficDeleteLeases/ClusterOfSize1/raftAfterSave=panic() (3.33s)
FAIL
FAIL    go.etcd.io/etcd/tests/v3/robustness     29.317s
FAIL
FAIL: (code:1):
  % (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' 'go.etcd.io/etcd/tests/v3/robustness' '-timeout=30m' '-v' '-count' '200' '--run' 'TestRobustnessExploratory/EtcdTrafficDeleteLeases/ClusterOfSize1' '--failfast')
ERROR: Tests for following packages failed:
   go.etcd.io/etcd/tests/v3/robustness
FAIL: 'robustness' FAILED at Sun Aug  4 05:56:08 PM CST 2024
make: *** [Makefile:59: test-robustness] Error 255
➜  etcd git:(fuweid/revert-18213) ✗
```



    -\t{Event: model.Event{Type: \"delete-operation\", Key: \"key7\"}, Revision: 143},

The etcd deleted key7 during panic.

```
{
  "ClientId": 4,
  "Input": {
    "Type": "txn",
    "LeaseGrant": null,
    "LeaseRevoke": null,
    "Range": null,
    "Txn": {
      "Conditions": null,
      "OperationsOnSuccess": [
        {
          "Type": "delete-operation",
          "Range": {
            "Start": "",
            "End": "",
            "Limit": 0
          },
          "Put": {
            "Key": "",
            "Value": {
              "Value": "",
              "Hash": 0
            },
            "LeaseID": 0
          },
          "Delete": {
            "Key": "key7"
          }
        }
      ],
      "OperationsOnFailure": null
    },
    "Defragment": null,
    "Compact": null
  },
  "Call": 1005589898,
  "Output": {
    "Txn": null,
    "Range": null,
    "LeaseGrant": null,
    "LeaseRevoke": null,
    "Defragment": null,
    "Compact": null,
    "ClientError": "",
    "Revision": 0,
    "Persisted": false,
    "PersistedRevision": 0,
    "Error": "rpc error: code = Unavailable desc = error reading from server: EOF"
  },
  "Return": 3152879375
}
```

There was compaction scheduled on revision 143. Even if the delete operation was failed caused by panic, it did create tombstone for key7 on revision 143. So, the etcd deleted that revision and then we lost the event.

```
    logger.go:146: 2024-08-04T17:56:05.862+0800 INFO    started server. {"name": "TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0", "pid": 1486109}
    logger.go:146: 2024-08-04T17:56:05.863+0800 INFO    Finished triggering failpoint   {"failpoint": "raftAfterSave=panic()"}
/home/fuweid/go/src/github.com/etcd-io/etcd/bin/etcd (TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0) (1486109): {"level":"info","ts":"2024-08-04T17:56:06.738386+0800","caller":"mvcc/index.go:194","msg":"compact tree index","revision":143}
/home/fuweid/go/src/github.com/etcd-io/etcd/bin/etcd (TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0) (1486109): {"level":"info","ts":"2024-08-04T17:56:06.742785+0800","caller":"mvcc/kvstore_compaction.go:71","msg":"finished scheduled compaction","compact-revision":143,"took":"4.330324ms","hash":3508007622,"current-db-size-bytes":503808,"current-db-
size":"504 kB","current-db-size-in-use-bytes":65536,"current-db-size-in-use":"66 kB"}
/home/fuweid/go/src/github.com/etcd-io/etcd/bin/etcd (TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize1-test-0) (1486109): {"level":"info","ts":"2024-08-04T17:56:06.742814+0800","caller":"mvcc/hash.go:143","msg":"storing new hash","hash":3508007622,"revision":143,"compact-revision":137}

```

cc @ahrtr  @serathius

ahrtr:
> I reproduced this issue without this patch and #18213.

Which dev branch are you testing against? What's the exactly command you ran?

fuweid:
> Which dev branch are you testing against? What's the exactly command you ran?

```
$ git checkout main
$ git fetch upstream # my remote is pointing to upstream
$ git rebase upstream/main
$ git checkout -b revert-18213
$ make clean
$ git revert 4fe227c46c
$ make gofail-enable
$ make
$ EXPECT_DEBUG=true GO_TEST_FLAGS="-v -count 100 --run TestRobustnessExploratory/EtcdTrafficDeleteLeases/ClusterOfSize1 --failfast" RESULTS_DIR=/tmp/results make test-robustness
```

@ahrtr  It needs luck to reproduce it because compaction should be applied on tombstone revision :)

serathius:
From my side I had problem with reproducing the issue on any other commit than https://github.com/etcd-io/etcd/pull/18213, I tried rebasing the it on newer branch it still didn't work, and even if it worked it was rare (<1/100). I even modified traffic to remove reads leaving only PUT, DELETE, COMPACTION. Still reproducibility was low. 

I trust that testing added by @fuweid should confirm the fix, however I think we should also improve robustness test to be able to get reproducability of at least 1/50, so we can properly confirm resolution from robustness perspective. I will look into adding write only traffic to increase the chance of compaction on tombstone revision, however I'm open to any ideas you have @fuweid @MadhavJivrajani @ahrtr

ahrtr:
I reproduced the issue as well in my local environment with the commands mentioned above https://github.com/etcd-io/etcd/pull/18274#issuecomment-2268004184.

I believe that the issue of "`broke Reliable`" is exactly the issue that we have already figured out the root cause, and also what this PR is trying to fix. 

The key point to reproduce this issue is compaction right after a delete, and with a panic in between. E2e tests instead of robustness test are the best fit to verify such a corner case.

If you want to improve the rebustness test, probably you can add operations such as "`panic right after deletion`", "`compaction right after panic`" and "`compaction after deletion`".  Also, let all such scenarios be randomly composed. But on other hand, we should always consider to make the test simpler and easy to understand; otherwise we may end up spending huge effort debug the test itself. It's a tradeoff.

fuweid:
> I even modified traffic to remove reads leaving only PUT, DELETE, COMPACTION. Still reproducibility was low.

YES. I didn't reproduce this issue in 4 hours, when I used `taskset -c 0,1,2` to run the cases with 3 cores (GITHUB action was using 3 cores). After I switch to 16 cores, I can reproduce it with `1/500` possibility in 30 minutes. It's pretty low. 

```
$ (my local) lscpu
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          48 bits physical, 48 bits virtual
  Byte Order:             Little Endian
CPU(s):                   16
  On-line CPU(s) list:    0-15
Vendor ID:                AuthenticAMD
  Model name:             AMD Ryzen 7 5800H with Radeon Graphics
    CPU family:           25
    Model:                80
    Thread(s) per core:   2
...
```
 

> however I think we should also improve robustness test to be able to get reproducability of at least 1/50, so we can properly confirm resolution from robustness perspective.

Agree. Like what @ahrtr said, we can add new cases with `panic after deletion`, `compaction on deletion`. The cases can cover deletion and compaction scenarios.

server/storage/mvcc/hash.go
```diff
@@ -71,6 +74,16 @@ func (h *kvHasher) WriteKeyValue(k, v []byte) {
 			return
 		}
 	}
+
+	// NOTE: If the previous compacted revision was tombstone, existing
+	// releases already deleted that key(s). However, new releases won't
+	// delete the key(s) on compacted revision, no matter what type key it
+	// is. So, we should skip previous compacted revision to keep consistent
+	// with existing releases.
```
ahrtr, line 82:
```suggestion	
	// When performing compaction, if the compacted revision is a
	// tombstone, older versions (<= 3.5.15 or <= 3.4.33) will delete
	// the tombstone. But newer versions (> 3.5.15 or > 3.4.33) won't
	// delete it. So we should skip the tombstone in such cases when
	// computing the hash to ensure that both older and newer versions
	// can always generate the same hash values.	
```

server/storage/mvcc/key_index.go
```diff
@@ -242,6 +237,10 @@ func (ki *keyIndex) keep(atRev int64, available map[Revision]struct{}) {
 	g := &ki.generations[genIdx]
 	if !g.isEmpty() {
 		// remove any tombstone
+		//
+		// NOTE: It's different from compact function which keeps
+		// tombstone in available, because we need to keep consistent
+		// with existing releases.
```
ahrtr, line 243:
```suggestion		
		// If the given `atRev` is a tombstone, we need to skip it.
		//
		// Note that this s different from the `compact` function which
		// keeps tombstone in such case. We need to stay consistent with
		// existing versions, ensuring they always generate the same hash
		// values.		
```

ahrtr:
Please also put all robustness test related change into a separate PR to make the review easier.

server/storage/mvcc/key_index.go
fuweid: Updated.

server/storage/mvcc/hash.go
fuweid: Updated.

fuweid:
> Please also put all robustness test related change into a separate PR to make the review easier.

Dropped the last commit. PTAL. Thanks

server/storage/mvcc/key_index_test.go
```diff
@@ -453,6 +474,12 @@ func TestKeyIndexCompactAndKeep(t *testing.T) {
 		},
 	}
 
+	tombstones := map[int64]struct{}{
+		6:  {},
+		12: {},
+		16: {},
+	}
```
ahrtr, line 481: You should be able to check whether a compacted revision is a tombstone programatically.
```suggestion
```

ahrtr: Hard coding the tombstone is weird and not friendly to other contributors, and according hard to maintain.

server/storage/mvcc/index_test.go
```diff
@@ -235,101 +235,444 @@ func TestIndexRevision(t *testing.T) {
...
+				Revision{Main: 10, Sub: 1}: {},
+			},
+			compacted: map[Revision]struct{}{
+				Revision{Main: 10}:         {},
+				Revision{Main: 10, Sub: 1}: {},
+			},
+		},
 	}
+
+	ti := buildTreeIndex()
```
ahrtr, line 634: Please keep the original comment
```suggestion
	ti := buildTreeIndex()
	// Continuous Compact and Keep
```

ahrtr: Also put the `ti := buildTreeIndex()` inside the `for` loop?

server/storage/mvcc/key_index_test.go
```diff
@@ -462,9 +489,14 @@ func TestKeyIndexCompactAndKeep(t *testing.T) {
 		if !reflect.DeepEqual(ki, kiclone) {
 			t.Errorf("#%d: ki = %+v, want %+v", i, ki, kiclone)
 		}
-		if !reflect.DeepEqual(am, tt.wam) {
-			t.Errorf("#%d: am = %+v, want %+v", i, am, tt.wam)
+
+		if _, ok := tombstones[tt.compact]; ok {
+			assert.Equal(t, len(am), 0, "#%d: ki = %d, keep result wants empty because tombstone", i, ki)
```
ahrtr, line 494: Also as mentioned above, you should be able to programmatically check whether `tt.compact` is a tombstone
```suggestion
			assert.Equal(t, 0, len(am), "#%d: ki = %d, keep result wants empty because tombstone", i, ki)
```

server/storage/mvcc/key_index_test.go
```diff
@@ -508,9 +540,14 @@ func TestKeyIndexCompactAndKeep(t *testing.T) {
 		if !reflect.DeepEqual(ki, kiClone) {
 			t.Errorf("#%d: ki = %+v, want %+v", i, ki, kiClone)
 		}
-		if !reflect.DeepEqual(am, tt.wam) {
-			t.Errorf("#%d: am = %+v, want %+v", i, am, tt.wam)
+
+		if _, ok := tombstones[tt.compact]; ok {
+			assert.Equal(t, len(am), 0, "#%d: ki = %d, keep result wants empty because tombstone", i, ki)
```
ahrtr, line 545:
```suggestion
			assert.Equal(t, 0, len(am), "#%d: ki = %d, keep result wants empty because tombstone", i, ki)
```

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,175 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+		}
+	}()
+
+	receivedEvents := make([]*clientv3.Event, 0)
+
+	fromCompactedRev := false
+	for fromRev := range revisionChan {
+		watchChan := c.Watch(ctx, key, clientv3.WithRev(fromRev))
+
+		oldEl := len(receivedEvents)
```
ahrtr, line 420:
```suggestion
		prevEventCount := len(receivedEvents)
```

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,175 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+
+		// firstReceived represents this is first watch response.
+		// Just in case that ETCD sends event one by one.
+		firstReceived := true
+
+		t.Logf("Start to watch key %s starting from revision %d", key, fromRev)
+	watchLoop:
+		for {
+			curEl := len(receivedEvents)
```
ahrtr, line 429:
```suggestion
			currentEventCount := len(receivedEvents)
```

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,175 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+	type valueEvent struct {
+		value string
+		typ   mvccpb.Event_EventType
+	}
+	values := make([]valueEvent, 0)
+
+	// revisionChan sends compacted revisions from which watchers should start.
+	revisionChan := make(chan int64)
+	// compactedAfterChanges means that client would compact after 7 put or delete requests.
+	compactedAfterChanges := 7
```
ahrtr, line 368:
```suggestion	
	var (
		requestedValues   = make([]valueEvent, 0) // record all requested changes
		compactionRevChan = make(chan int64)      // pass each compacted revision via this channel
		compactionStep    = 7                     // perform a compaction on every 7 operations
	)	
```

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,175 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+			select {
+			case watchResp := <-watchChan:
+				t.Logf("Receive the number of events: %d", len(watchResp.Events))
+				for i := range watchResp.Events {
+					ev := watchResp.Events[i]
+
+					// If the $fromRev is the compacted revision,
+					// the first event should be the same as the last event receives in last watch response.
+					if i == 0 && firstReceived && fromCompactedRev {
+						last := receivedEvents[oldEl-1]
```
ahrtr, line 443:
```suggestion
					if firstReceived && fromCompactedRev {
						firstReceived = false
						last := receivedEvents[oldEl-1]
```

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,175 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+					}
+					receivedEvents = append(receivedEvents, ev)
+				}
+
+				if len(watchResp.Events) == 0 {
+					require.Equal(t, v3rpc.ErrCompacted, watchResp.Err())
+					break watchLoop
+				}
+
+				firstReceived = false
```
ahrtr, line 461:
```suggestion
```

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,175 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+					// the first event should be the same as the last event receives in last watch response.
+					if i == 0 && firstReceived && fromCompactedRev {
+						last := receivedEvents[oldEl-1]
+
+						assert.Equal(t, last.Type, ev.Type,
+							"Last received event type %s, but got event type %s", last.Type, ev.Type)
+						assert.Equal(t, string(last.Kv.Key), string(ev.Kv.Key),
+							"Last received event key %s, but got event key %s", string(last.Kv.Key), string(ev.Kv.Key))
+						assert.Equal(t, string(last.Kv.Value), string(ev.Kv.Value),
+							"Last received event value %s, but got event value %s", string(last.Kv.Value), string(ev.Kv.Value))
```
ahrtr, line 450: nit: Lower case for the first letter of the message

ahrtr:
Overall looks good to me. 

@ivanvc @jmhbnz @serathius @siyuanfoundation Please take a look.

server/storage/mvcc/key_index_test.go
fuweid: Updated.

server/storage/mvcc/index_test.go
fuweid: Updated.

> Also put the ti := buildTreeIndex() inside the for loop?

The case keeps performing compaction on one `ki`. So we can't put it inside the loop.

server/storage/mvcc/key_index_test.go
fuweid: Updated.

server/storage/mvcc/key_index_test.go
fuweid: Updated.

tests/e2e/watch_test.go
fuweid: Updated.

tests/e2e/watch_test.go
fuweid: Updated.

tests/e2e/watch_test.go
fuweid: Updated.

tests/e2e/watch_test.go
fuweid: Thanks. Updated.

tests/e2e/watch_test.go
fuweid: Updated.

serathius:
Can we close/mark as addressed all the comments opened about the code?

fuweid:
> Can we close/mark as addressed all the comments opened about the code?

Done. PTAL，thanks

serathius (approved):

serathius (approved):

tests/e2e/watch_test.go
```diff
@@ -311,24 +313,178 @@ func TestDeleteEventDrop_Issue18089(t *testing.T) {
...
+func TestStartWatcherFromCompactedRevision(t *testing.T) {
+	t.Run("compaction on tombstone revision", func(t *testing.T) {
+		testStartWatcherFromCompactedRevision(t, true)
+	})
+	t.Run("compaction on normal revision", func(t *testing.T) {
+		testStartWatcherFromCompactedRevision(t, false)
+	})
+}
+
+func testStartWatcherFromCompactedRevision(t *testing.T, performCompactOnTombstone bool) {
```
serathius, line 344: This function would benefit from splitting it up and cleanup, but let's not block fix on test.

ahrtr (approved):
LGTM

Nice work! @fuweid

ahrtr:
@fuweid could you backport https://github.com/etcd-io/etcd/pull/18369 and this PR to 3.5 and 3.4? thx

fuweid:
> @fuweid could you backport #18369 and this PR to 3.5 and 3.4? thx

@ahrtr Yes, I will do the backport. Thanks for the review! @ahrtr @serathius

ahrtr:
I see the failure below in
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19077/pull-etcd-e2e-amd64/1868930443156393984 .

But unfortunately, I couldn't reproduce it locally. I also manually verified the steps, and couldn't reproduce it.

```
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-2) (61865): {"level":"info","ts":"2024-12-17T08:23:04.293653Z","caller":"embed/serve.go:258","msg":"serving client traffic securely","traffic":"grpc+http","address":"127.0.0.1:20010"}
    hashkv_test.go:210: Writing keys: [key0], value: 1
    hashkv_test.go:210: Writing keys: [key0], value: 2
    hashkv_test.go:210: Writing keys: [key0], value: 3
    hashkv_test.go:210: Writing keys: [key0], value: 4
    hashkv_test.go:210: Writing keys: [key0], value: 5
    hashkv_test.go:210: Writing keys: [key0], value: 6
    hashkv_test.go:210: Writing keys: [key0], value: 7
    hashkv_test.go:210: Writing keys: [key0], value: 8
    hashkv_test.go:210: Writing keys: [key0], value: 9
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-2) (61865): {"level":"info","ts":"2024-12-17T08:23:04.324070Z","caller":"etcdserver/server.go:518","msg":"initialized peer connections; fast-forwarding election ticks","local-member-id":"72e86eb5e356a737","forward-ticks":8,"forward-duration":"800ms","election-ticks":10,"election-timeout":"1s","active-remote-members":2}
    hashkv_test.go:196: Deleting key=key0
    hashkv_test.go:210: Writing keys: [key0], value: 11
    hashkv_test.go:210: Writing keys: [key0], value: 12
    hashkv_test.go:210: Writing keys: [key0], value: 13
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd-last-release (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-1) (61866): {"level":"info","ts":"2024-12-17T08:23:04.332Z","caller":"etcdserver/server.go:767","msg":"initialized peer connections; fast-forwarding election ticks","local-member-id":"5ac8aa22f1eb4c8f","forward-ticks":8,"forward-duration":"800ms","election-ticks":10,"election-timeout":"1s","active-remote-members":2}
    hashkv_test.go:210: Writing keys: [key0], value: 14
    hashkv_test.go:210: Writing keys: [key0], value: 15
    hashkv_test.go:210: Writing keys: [key0], value: 16
    hashkv_test.go:210: Writing keys: [key0], value: 17
    hashkv_test.go:210: Writing keys: [key0], value: 18
    hashkv_test.go:210: Writing keys: [key0], value: 19
    hashkv_test.go:196: Deleting key=key0
    hashkv_test.go:210: Writing keys: [key0], value: 21
    hashkv_test.go:210: Writing keys: [key0], value: 22
    hashkv_test.go:210: Writing keys: [key0], value: 23
    hashkv_test.go:210: Writing keys: [key0], value: 24
    hashkv_test.go:210: Writing keys: [key0], value: 25
    hashkv_test.go:210: Writing keys: [key0], value: 26
    hashkv_test.go:210: Writing keys: [key0], value: 27
    hashkv_test.go:210: Writing keys: [key0], value: 28
    hashkv_test.go:210: Writing keys: [key0], value: 29
    hashkv_test.go:196: Deleting key=key0
    hashkv_test.go:210: Writing keys: [key0], value: 31
    hashkv_test.go:210: Writing keys: [key0], value: 32
    hashkv_test.go:210: Writing keys: [key0], value: 33
    hashkv_test.go:210: Writing keys: [key0], value: 34
    hashkv_test.go:210: Writing keys: [key0], value: 35
    hashkv_test.go:210: Writing keys: [key0], value: 36
    hashkv_test.go:210: Writing keys: [key0], value: 37
    hashkv_test.go:210: Writing keys: [key0], value: 38
    hashkv_test.go:210: Writing keys: [key0], value: 39
    hashkv_test.go:196: Deleting key=key0
    hashkv_test.go:91: COMPACT on rev=16
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd-last-release (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-0) (61864): {"level":"info","ts":"2024-12-17T08:23:04.411Z","caller":"mvcc/index.go:214","msg":"compact tree index","revision":16}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd-last-release (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-0) (61864): {"level":"info","ts":"2024-12-17T08:23:04.411Z","caller":"mvcc/kvstore_compaction.go:57","msg":"finished scheduled compaction","compact-revision":16,"took":"66.101µs"}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd-last-release (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-1) (61866): {"level":"info","ts":"2024-12-17T08:23:04.411Z","caller":"mvcc/index.go:214","msg":"compact tree index","revision":16}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd-last-release (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-1) (61866): {"level":"info","ts":"2024-12-17T08:23:04.411Z","caller":"mvcc/kvstore_compaction.go:57","msg":"finished scheduled compaction","compact-revision":16,"took":"59.131µs"}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-2) (61865): {"level":"info","ts":"2024-12-17T08:23:04.411560Z","caller":"mvcc/index.go:194","msg":"compact tree index","revision":16}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-2) (61865): {"level":"info","ts":"2024-12-17T08:23:04.412101Z","caller":"mvcc/kvstore_compaction.go:71","msg":"finished scheduled compaction","compact-revision":16,"took":"454.515µs","hash":2489252793,"current-db-size-bytes":24576,"current-db-size":"25 kB","current-db-size-in-use-bytes":24576,"current-db-size-in-use":"25 kB"}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd (TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-2) (61865): {"level":"info","ts":"2024-12-17T08:23:04.412168Z","caller":"mvcc/hash.go:157","msg":"storing new hash","hash":2489252793,"revision":16,"compact-revision":-1}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_compact_16_--physical) (61917): compacted revision 16
    hashkv_test.go:223: HashKV on rev=16
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_endpoint_hashkv_--rev_16_-w_json) (61928): {"level":"warn","ts":"2024-12-17T08:23:04.442409Z","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc000344780/localhost:20000","method":"/etcdserverpb.Maintenance/HashKV","attempt":0,"error":"rpc error: code = OutOfRange desc = etcdserver: mvcc: required revision has been compacted"}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_endpoint_hashkv_--rev_16_-w_json) (61928): Failed to get the hash of endpoint https://localhost:20000 (etcdserver: mvcc: required revision has been compacted)
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_endpoint_hashkv_--rev_16_-w_json) (61928): {"level":"warn","ts":"2024-12-17T08:23:04.458105Z","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc0003c01e0/localhost:20005","method":"/etcdserverpb.Maintenance/HashKV","attempt":0,"error":"rpc error: code = OutOfRange desc = etcdserver: mvcc: required revision has been compacted"}
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_endpoint_hashkv_--rev_16_-w_json) (61928): Failed to get the hash of endpoint https://localhost:20005 (etcdserver: mvcc: required revision has been compacted)
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_endpoint_hashkv_--rev_16_-w_json) (61928): [{"Endpoint":"https://localhost:20010","HashKV":{"header":{"cluster_id":16074511605466142234,"member_id":8279989642402899767,"revision":41,"raft_term":2},"hash":2489252793,"compact_revision":-1,"hash_revision":16}}]
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl (/home/prow/go/src/github.com/etcd-io/etcd/bin/etcdctl_--cacert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/ca.crt_--cert=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.crt_--key=/home/prow/go/src/github.com/etcd-io/etcd/tests/fixtures/server.key.insecure_--endpoints=https://localhost:20000,https://localhost:20005,https://localhost:20010_endpoint_hashkv_--rev_16_-w_json) (61928): Error: etcdserver: mvcc: required revision has been compacted
    hashkv_test.go:227: 
        	Error Trace:	/home/prow/go/src/github.com/etcd-io/etcd/tests/e2e/hashkv_test.go:227
        	            				/home/prow/go/src/github.com/etcd-io/etcd/tests/e2e/hashkv_test.go:96
        	Error:      	"1" is not greater than "1"
        	Test:       	TestVerifyHashKVAfterCompact/compactedOnTombstone=false_-_quorum-last-version_-_Keys=[key0]
    logger.go:146: 2024-12-17T08:23:04.480Z	INFO	closing test cluster...
    logger.go:146: 2024-12-17T08:23:04.480Z	INFO	stopping server...	{"name": "TestVerifyHashKVAfterCompactcompactedOnTombstonefalse_-_quorum-last-version_-_Keyskey0-test-0"}
```

fuweid:
ping @ahrtr 

I think it's related to github service. The test was using v3.5.0 which is not expected.

```
'release' started at Tue Dec 17 08:08:05 UTC 2024
fatal: unable to access 'https://github.com/etcd-io/etcd.git/': The requested URL returned error: 503
Found latest release: .
              
fallback to v3.5.0
Downloading etcd-v3.5.0-linux-amd64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

 76 18.4M   76 14.1M    0     0  22.1M      0 --:--:-- --:--:-- --:--:-- 22.1M
100 18.4M  100 18.4M    0     0  27.1M      0 --:--:-- --:--:-- --:--:--  101M
etcd-v3.5.0-linux-amd64/Documentation/
etcd-v3.5.0-linux-amd64/Documentation/dev-guide/
etcd-v3.5.0-linux-amd64/Documentation/dev-guide/apispec/
etcd-v3.5.0-linux-amd64/Documentation/dev-guide/apispec/swagger/
etcd-v3.5.0-linux-amd64/Documentation/dev-guide/apispec/swagger/v3lock.swagger.json
etcd-v3.5.0-linux-amd64/Documentation/dev-guide/apispec/swagger/v3election.swagger.json
etcd-v3.5.0-linux-amd64/Documentation/dev-guide/apispec/swagger/rpc.swagger.json
etcd-v3.5.0-linux-amd64/Documentation/README.md
etcd-v3.5.0-linux-amd64/README-etcdutl.md
etcd-v3.5.0-linux-amd64/READMEv2-etcdctl.md
etcd-v3.5.0-linux-amd64/README-etcdctl.md
etcd-v3.5.0-linux-amd64/README.md
etcd-v3.5.0-linux-amd64/etcdutl
etcd-v3.5.0-linux-amd64/etcdctl
etcd-v3.5.0-linux-amd64/etcd
'release' PASSED and completed at Tue Dec 17 08:08:07 UTC 2024
```

ahrtr:
> I think it's related to github service. The test was using v3.5.0 which is not expected.

script issue?
https://github.com/etcd-io/etcd/blob/b4450d510ddf0f011c874c28e95175b4bac7f6c8/scripts/test.sh#L598-L599

fuweid:
It's github service issue.

```
'release' started at Tue Dec 17 08:08:05 UTC 2024
fatal: unable to access 'https://github.com/etcd-io/etcd.git/': The requested URL returned error: 503
Found latest release: .
```

The git service returns 503 so that we fallback to v3.5.0

fuweid:
+1 to abort test if there is something wrong during fetching latest version

PR #18279, opened 2024-07-03
Add "server-feature-gates" flag.

siyuanfoundation:
part of https://github.com/etcd-io/etcd/issues/18023

### DIFF
--- a/pkg/featuregate/feature_gate.go
+++ b/pkg/featuregate/feature_gate.go
@@ -16,6 +16,7 @@
 package featuregate
 
 import (
+	"flag"
 	"fmt"
 	"sort"
 	"strconv"
@@ -30,7 +31,7 @@ import (
 type Feature string
 
 const (
-	flagName = "feature-gates"
+	defaultFlagName = "feature-gates"
 
 	// allAlphaGate is a global toggle for alpha features. Per-feature key
 	// values override the default set by allAlphaGate. Examples:
@@ -98,7 +99,7 @@ type MutableFeatureGate interface {
 	FeatureGate
 
 	// AddFlag adds a flag for setting global feature gates to the specified FlagSet.
-	AddFlag(fs *pflag.FlagSet)
+	AddFlag(fs *flag.FlagSet, flagName string)
 	// Set parses and stores flag gates for known features
 	// from a string like feature1=true,feature2=false,...
 	Set(value string) error
@@ -165,6 +166,9 @@ func setUnsetBetaGates(known map[Feature]FeatureSpec, enabled map[Feature]bool, 
 var _ pflag.Value = &featureGate{}
 
 func New(name string, lg *zap.Logger) *featureGate {
+	if lg == nil {
+		lg = zap.NewNop()
+	}
 	known := map[Feature]FeatureSpec{}
 	for k, v := range defaultFeatures {
 		known[k] = v
@@ -349,7 +353,10 @@ func (f *featureGate) Enabled(key Feature) bool {
 }
 
 // AddFlag adds a flag for setting global feature gates to the specified FlagSet.
-func (f *featureGate) AddFlag(fs *pflag.FlagSet) {
+func (f *featureGate) AddFlag(fs *flag.FlagSet, flagName string) {
+	if flagName == "" {
+		flagName = defaultFlagName
+	}
 	f.lock.Lock()
 	// TODO(mtaufen): Shouldn't we just close it on the first Set/SetFromMap instead?
 	// Not all components expose a feature gates flag using this AddFlag method, and
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -42,6 +42,7 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/pkg/v3/featuregate"
 	"go.etcd.io/etcd/pkg/v3/flags"
 	"go.etcd.io/etcd/pkg/v3/netutil"
 	"go.etcd.io/etcd/server/v3/config"
@@ -50,6 +51,7 @@ import (
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3compactor"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3discovery"
+	"go.etcd.io/etcd/server/v3/features"
 )
 
 const (
@@ -108,6 +110,8 @@ const (
 	maxElectionMs = 50000
 	// backend freelist map type
 	freelistArrayType = "array"
+
+	ServerFeatureGateFlagName = "feature-gates"
 )
 
 var (
@@ -455,6 +459,9 @@ type Config struct {
 
 	// V2Deprecation describes phase of API & Storage V2 support
 	V2Deprecation config.V2DeprecationEnum `json:"v2-deprecation"`
+
+	// ServerFeatureGate is a server level feature gate
+	ServerFeatureGate featuregate.FeatureGate
 }
 
 // configYAML holds the config suitable for yaml parsing
@@ -576,6 +583,7 @@ func NewConfig() *Config {
 		},
 
 		AutoCompactionMode: DefaultAutoCompactionMode,
+		ServerFeatureGate:  features.NewDefaultServerFeatureGate(DefaultName, nil),
 	}
 	cfg.InitialCluster = cfg.InitialClusterFromName(cfg.Name)
 	return cfg
@@ -762,6 +770,9 @@ func (cfg *Config) AddFlags(fs *flag.FlagSet) {
 	// unsafe
 	fs.BoolVar(&cfg.UnsafeNoFsync, "unsafe-no-fsync", false, "Disables fsync, unsafe, will cause data loss.")
 	fs.BoolVar(&cfg.ForceNewCluster, "force-new-cluster", false, "Force to create a new one member cluster.")
+
+	// featuregate
+	cfg.ServerFeatureGate.(featuregate.MutableFeatureGate).AddFlag(fs, ServerFeatureGateFlagName)
 }
 
 func ConfigFromFile(path string) (*Config, error) {
--- a/server/etcdmain/config_test.go
+++ b/server/etcdmain/config_test.go
@@ -25,8 +25,10 @@ import (
 
 	"sigs.k8s.io/yaml"
 
+	"go.etcd.io/etcd/pkg/v3/featuregate"
 	"go.etcd.io/etcd/pkg/v3/flags"
 	"go.etcd.io/etcd/server/v3/embed"
+	"go.etcd.io/etcd/server/v3/features"
 )
 
 func TestConfigParsingMemberFlags(t *testing.T) {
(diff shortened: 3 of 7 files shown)


server/features/etcd_features.go
```diff
@@ -0,0 +1,61 @@
...
+	// MyFeature featuregate.Feature = "MyFeature"
+	//
+	// Feature gates should be listed in alphabetical, case-sensitive
+	// (upper before any lower case character) order. This reduces the risk
+	// of code conflicts because changes are more likely to be scattered
+	// across the file.
+
+	// DistributedTracing enables experimental distributed  tracing using OpenTelemetry Tracing.
+	// alpha: v3.5
+	// issue: https://github.com/etcd-io/etcd/issues/12460
```
ahrtr, line 40: How does Kubernetes manage this?

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

server/embed/config.go
```diff
@@ -762,6 +770,9 @@ func (cfg *Config) AddFlags(fs *flag.FlagSet) {
 	// unsafe
 	fs.BoolVar(&cfg.UnsafeNoFsync, "unsafe-no-fsync", false, "Disables fsync, unsafe, will cause data loss.")
 	fs.BoolVar(&cfg.ForceNewCluster, "force-new-cluster", false, "Force to create a new one member cluster.")
+
+	// featuregate
+	cfg.ServerFeatureGate.(featuregate.MutableFeatureGate).AddFlag(fs, ServerFeatureGateFlagName)
```
ahrtr, line 775: We haven't finalized how to proceed with the so-called "cluster-level" feature.
```suggestion
	cfg.ServerFeatureGate.(featuregate.MutableFeatureGate).AddFlag(fs, "feature-gates")
```

server/embed/config.go
```diff
@@ -108,6 +110,8 @@ const (
 	maxElectionMs = 50000
 	// backend freelist map type
 	freelistArrayType = "array"
+
+	ServerFeatureGateFlagName = "server-feature-gates"
```
ahrtr, line 114:
```suggestion
```

pkg/featuregate/feature_gate.go
```diff
@@ -121,6 +122,8 @@ type MutableFeatureGate interface {
 	// overriding its default to true for a limited number of components without simultaneously
 	// changing its default for all consuming components.
 	OverrideDefault(name Feature, override bool) error
+	// SetLogger replaces the logger with the provided logger.
+	SetLogger(lg *zap.Logger)
```
ahrtr, line 126: It's a little weird to add `SetLogger` into the interface. No any public interface methods depend on the zap.Logger.
Can we try to remove it?
```suggestion
```

server/features/etcd_features.go
siyuanfoundation: what do you mean by "manage this"? The docs are created manually and enforced as part of release process.

server/embed/config.go
siyuanfoundation: `ServerFeatureGateFlagName` is referenced in another place as well. We would want different flags for `server-feature-gates` and `cluster-feature-gates`.

pkg/featuregate/feature_gate.go
siyuanfoundation: removed.

server/embed/config.go
serathius: We need to clarify semantics of feature gates to ensure that flag names properly reflect that. Please see my comment. https://github.com/kubernetes/enhancements/pull/4662#discussion_r1662333970

 Still until we cut v3.6 we can always change the name. Let's base our current naming on fact that we are only implementing server level flags and that KEP for cluster level features is not yet landed.

siyuanfoundation: ok. Keeping the flag name `feature-gates` for now.

pkg/featuregate/feature_gate.go
```diff
@@ -98,7 +99,7 @@ type MutableFeatureGate interface {
 	FeatureGate
 
 	// AddFlag adds a flag for setting global feature gates to the specified FlagSet.
-	AddFlag(fs *pflag.FlagSet)
+	AddFlag(fs *flag.FlagSet, flagName string)
```
ivanvc, line 102: Is there a particular reason why you replaced `pflag` with `flag`?

server/features/etcd_features.go
tjungblu: it's basically the same:
https://github.com/kubernetes/kubernetes/blob/3125877de0aefbb642833beaa3a4c189d833e797/pkg/features/kube_features.go

ahrtr: How can you make sure the comments, source code and user-facing doc are in sync? You don't have to do it in this PR, but please think about it.

server/embed/config.go
```diff
@@ -455,6 +459,9 @@ type Config struct {
 
 	// V2Deprecation describes phase of API & Storage V2 support
 	V2Deprecation config.V2DeprecationEnum `json:"v2-deprecation"`
+
+	// ServerFeatureGate is a server level feature gate
+	ServerFeatureGate featuregate.FeatureGate
```
ahrtr, line 464:
```suggestion
	ServerFeatureGate featuregate.MutableFeatureGate
```

server/embed/config.go
ahrtr, line 775: If you follow https://github.com/etcd-io/etcd/pull/18279/files#r1670113740, then no need to do the type cast.
```suggestion
	cfg.ServerFeatureGate.AddFlag(fs, ServerFeatureGateFlagName)
```

pkg/featuregate/feature_gate.go
ahrtr: Removing `SetLogger` is good, but on the other hand, it seems that you can't pass a valid `lg` to `featuregates.New`. Not a big problem though.

pkg/featuregate/feature_gate.go
siyuanfoundation: all the other flags in `server/embed/config.go` uses `flag` instead of `pflag`

server/embed/config.go
siyuanfoundation: I think it is safer to keep the general interface immutable, and explicitly cast to `MutableFeatureGate` when it is needed in a limited few places. This way, we can prevent the feature gate from being accidentally changed.

pkg/featuregate/feature_gate.go
ivanvc: That is good to know. Thanks for your answer. I would think that using `pflag` is better as it is POSIX-compliant, but I do see that across the project, we're using `flag` for flagsets. We still parse flags using Cobra, which makes them POSIX-compliant.

server/features/etcd_features.go
```diff
@@ -0,0 +1,61 @@
...
+	DistributedTracing featuregate.Feature = "DistributedTracing"
+	// StopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
+	// owner: @chaochn47
+	// alpha: v3.6
+	StopGRPCServiceOnDefrag featuregate.Feature = "StopGRPCServiceOnDefrag"
```
ivanvc, line 45: Does adding the suggested `owner:` to `DistributedTracing` and `issue:` to `StopGRPCServiceOnDefrag` make sense? Given that this pull request will set the example for other feature flags, I think it would be important to either add them or specify in the documentation (and template) above what to do in such cases when there is no owner or issue to tag.

siyuanfoundation: makes sense. Added more info for these 2 features.

server/etcdmain/help.go
```diff
@@ -103,6 +105,8 @@ Member:
     Read timeout set on each rafthttp connection
   --raft-write-timeout '` + rafthttp.DefaultConnWriteTimeout.String() + `'
     Write timeout set on each rafthttp connection
+  --feature-gates ''
```
jmhbnz, line 108: Noting we will need to raise corresponding pr for etcd website after merging to add user facing docs.

server/etcdmain/config_test.go
```diff
@@ -395,6 +397,55 @@ func TestFlagsPresentInHelp(t *testing.T) {
 	})
 }
 
+func TestParseFeatureGateFlags(t *testing.T) {
```
jmhbnz, line 400: Should we also update `tests/e2e/etcd_config_test.go` to add a feature gate related case? Might be nice to verify something like feature gate in config file.

jmhbnz (commented):
Thanks @siyuanfoundation - Overall this is looking good to me. Two minor items below.

server/etcdmain/help.go
siyuanfoundation: ack. This flag name might still change before we finalize the design of the cluster feature gate, so I will add the doc once that has been decided.

server/etcdmain/config_test.go
siyuanfoundation: That is a good suggestion. But right now, the features are not actually used, so the e2e would have no visible effect. In the followup PR that uses `StopGRPCServiceOnDefrag` we will have a e2e test added. Is that ok?

jmhbnz: No problem, just want to make sure e2e suite will also have coverage to exercise feature flag related paths in future.  I'm happy to defer it to next planned pr.

jmhbnz (approved):
LGTM - Many thanks for the effort @siyuanfoundation

server/features/etcd_features.go
siyuanfoundation: The comments in the source code will have to be manually reviewed. For the user facing doc, I have a [POC](https://github.com/kubernetes/kubernetes/pull/125830/files#diff-5caebbd07d40571513547d4992511a518eae2f3427abcdb0459f75e4a6e6dd90) in k8s to auto generate the feature list with lifecycles with static analysis. We can add similar scripts to do that

server/embed/config.go
siyuanfoundation: similar to https://github.com/etcd-io/etcd/pull/18279#discussion_r1670846400

server/embed/config.go
siyuanfoundation: Please see https://github.com/etcd-io/etcd/pull/18279#discussion_r1668945176

ivanvc (approved):
LGTM. Thanks for addresing my comments, @siyuanfoundation. Great job 😀

fuweid (approved):
LGTM

server/etcdmain/help.go
```diff
@@ -103,6 +105,8 @@ Member:
...
+    A set of key=value pairs that describe server level feature gates for alpha/experimental features. Options are:'` + strings.Join(features.NewDefaultServerFeatureGate("", nil).KnownFeatures(), ",") + `'
```
ahrtr, line 109: This is what I got. Please format the output, print each feature in a separate line.
```
  --feature-gates ''
    A set of key=value pairs that describe server level feature gates for alpha/experimental features. Options are:'AllAlpha=true|false (ALPHA - default=false),AllBeta=true|false (BETA - default=false),DistributedTracing=true|false (ALPHA - default=false),StopGRPCServiceOnDefrag=true|false (ALPHA - default=false)'
```

Please let me know if you want to fix it in this PR or a separate PR.

ahrtr, line 109:
```suggestion
    A set of key=value pairs that describe server level feature gates for alpha/experimental features. Options are:\n` + strings.Join(features.NewDefaultServerFeatureGate("", nil).KnownFeatures(), "\n") + `
```

siyuanfoundation: updated

server/features/etcd_features.go
```diff
@@ -0,0 +1,63 @@
...
+	// owner: @chaochn47
+	// alpha: v3.6
+	// main PR: https://github.com/etcd-io/etcd/pull/18279
+	StopGRPCServiceOnDefrag featuregate.Feature = "StopGRPCServiceOnDefrag"
+)
+
+var (
+	DefaultEtcdServerFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+		DistributedTracing:      {Default: false, PreRelease: featuregate.Alpha},
+		StopGRPCServiceOnDefrag: {Default: false, PreRelease: featuregate.Alpha},
```
ahrtr, line 53: You just added these two features, but they are still controlled by the legacy flags, are you going to migrate them to feature gate in a followup PR?

ahrtr (approved):
Overall looks good to me with a question. 

Thanks @siyuanfoundation 

cc @serathius

server/features/etcd_features.go
siyuanfoundation: Yes. They will be migrated in the next few followup PRs.

serathius (approved):

PR #18313, opened 2024-07-12
Differentiate the warning message for rejected client and peer connections

ahrtr:
etcdserver generates the same warning message "`rejected connection`" when it rejects a connection from the client and from the peer. It isn't good for debugging some connection issue. This PR is trying to generate a different warning message to differentiate client connections and peer connections.

### DIFF
--- a/server/embed/config_logging.go
+++ b/server/embed/config_logging.go
@@ -165,35 +165,38 @@ func (cfg *Config) setupLogging() error {
 			return err
 		}
 
-		logTLSHandshakeFailure := func(conn *tls.Conn, err error) {
-			state := conn.ConnectionState()
-			remoteAddr := conn.RemoteAddr().String()
-			serverName := state.ServerName
-			if len(state.PeerCertificates) > 0 {
-				cert := state.PeerCertificates[0]
-				ips := make([]string, len(cert.IPAddresses))
-				for i := range cert.IPAddresses {
-					ips[i] = cert.IPAddresses[i].String()
+		logTLSHandshakeFailureFunc := func(msg string) func(conn *tls.Conn, err error) {
+			return func(conn *tls.Conn, err error) {
+				state := conn.ConnectionState()
+				remoteAddr := conn.RemoteAddr().String()
+				serverName := state.ServerName
+				if len(state.PeerCertificates) > 0 {
+					cert := state.PeerCertificates[0]
+					ips := make([]string, len(cert.IPAddresses))
+					for i := range cert.IPAddresses {
+						ips[i] = cert.IPAddresses[i].String()
+					}
+					cfg.logger.Warn(
+						msg,
+						zap.String("remote-addr", remoteAddr),
+						zap.String("server-name", serverName),
+						zap.Strings("ip-addresses", ips),
+						zap.Strings("dns-names", cert.DNSNames),
+						zap.Error(err),
+					)
+				} else {
+					cfg.logger.Warn(
+						msg,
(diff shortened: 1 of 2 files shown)


tests/e2e/zap_logging_test.go
```diff
@@ -73,3 +75,84 @@ type logEntry struct {
...
+			// wait for the goroutine to get started
+			<-startedCh
+
+			t.Log("Running curl command to trigger the corresponding warning message.")
+			curlCmdArgs := []string{"curl", "--connect-timeout", "1", "-k", tc.url}
+			curlCmd, err := e2e.SpawnCmd(curlCmdArgs, nil)
+			require.NoError(t, err)
+
+			defer func() {
+				curlCmd.Stop()
```
serathius, line 146: Don't think you need stop when you use close.

tests/e2e/zap_logging_test.go
```diff
@@ -73,3 +75,84 @@ type logEntry struct {
...
+		expectedErrMsg string
+	}{
+		{
+			name:           "reject client connection",
+			url:            "https://127.0.0.1:2379/version",
+			expectedErrMsg: "rejected connection from client",
+		},
+		{
+			name:           "reject peer connection",
+			url:            "https://127.0.0.1:2380/members",
```
serathius, line 94: We cannot 100% know that only peers use the peer endpoint. Maybe message should be more like `rejected connection on peer url"

ahrtr: updated

tests/e2e/zap_logging_test.go
ahrtr: `Stop` is used to stop the process, and `Close` is to wait for the process to exit, both of them are needed.

Note this is `ExpectProcess` instead of `EtcdProcessCluster`.  Indeed, `EtcdProcessCluster.Stop` isn't needed when you use `EtcdProcessCluster.Close`. You might have confused the two.
https://github.com/etcd-io/etcd/blob/eeeb7e05b32466d13a25b3de0cbb03884319c1a9/tests/framework/e2e/cluster.go#L981-L983

serathius (approved):

server/embed/config_logging.go
```diff
@@ -165,35 +165,38 @@ func (cfg *Config) setupLogging() error {
...
-				ips := make([]string, len(cert.IPAddresses))
-				for i := range cert.IPAddresses {
-					ips[i] = cert.IPAddresses[i].String()
+		logTLSHandshakeFailureFunc := func(msg string) func(conn *tls.Conn, err error) {
+			return func(conn *tls.Conn, err error) {
+				state := conn.ConnectionState()
+				remoteAddr := conn.RemoteAddr().String()
+				serverName := state.ServerName
+				if len(state.PeerCertificates) > 0 {
+					cert := state.PeerCertificates[0]
```
jmhbnz, line 174: Could we ever be in a state where `PeerCertificates` other than `[0]` should be inspected?

ahrtr: I didn't change this. I just set different warning message on client and peer endpoints in this PR.

Going back to your question, `state.PeerCertificates[0]` is the leaf certificate which we should check/verify against, so I think it's expected.

jmhbnz: Yes not a change introduced by this pr, it was just a question that crossed my mind as I was reviewing overall.  Thanks for clarifying it.

jmhbnz (approved):
LGTM

PR #18250, opened 2024-06-29
Remove gsutil acl command for bucket permissions from release.sh

jmhbnz:
As outlined in https://github.com/etcd-io/etcd/issues/18013#issuecomment-2140934998 and https://github.com/etcd-io/etcd/issues/18013#issuecomment-2197759674 we have verified our `etcd-development/etcd` bucket in use for `gcr.io` is already fully public and manually verified that new objects pushed to the bucket automatically become public without needing `gsutil acl ...` command run.

This pull request will dramatically reduce output spam from this script when publishing a release which can make it frustrating to scrollback and review output.

We should merge and backport this before `v3.5.15` is released.

Fixes #18013

PR #18345, opened 2024-07-19
 [3.4] Remove gsutil acl command for bucket permissions

jmhbnz:
With verification completed during release `v3.5.15` we are certain this improvement is safe so we should now also backport https://github.com/etcd-io/etcd/pull/18250/commits/f0246a9f84f7839825d68511ac65eb52eb24e81d to `release-3.4`.

> The bucket is already fully public so new releases don't need to run this command.
> 
> This will prevent significant terminal spam during the release process as there are thousands of objects in the bucket.


jmhbnz:
cc @ahrtr, @ivanvc

ivanvc (approved):
LGTM. Thanks, @jmhbnz.

ahrtr (approved):
LGTM

Thanks

PR #18356, opened 2024-07-23
etcdctl: move cobra helper of etcdctl from main pkg to etcdctl pkg

ah8ad3:
This PR moves cobra helper of `etcdctl` from main pkg to `etcdctl` pkg. This will remove cobra dependency from main pkg which wasn't necessary.

This PR was created after suggestion in this comment https://github.com/etcd-io/etcd/pull/18183#discussion_r1687171454
Part of https://github.com/etcd-io/etcd/issues/17777

### DIFF
--- a/etcdctl/ctlv3/ctl.go
+++ b/etcdctl/ctlv3/ctl.go
@@ -23,6 +23,7 @@ import (
 
 	"go.etcd.io/etcd/api/v3/version"
 	"go.etcd.io/etcd/etcdctl/v3/ctlv3/command"
+	"go.etcd.io/etcd/etcdctl/v3/util"
 	"go.etcd.io/etcd/pkg/v3/cobrautl"
 )
 
@@ -102,7 +103,7 @@ func init() {
 }
 
 func usageFunc(c *cobra.Command) error {
-	return cobrautl.UsageFunc(c, version.Version, version.APIVersion)
+	return util.UsageFunc(c, version.Version, version.APIVersion)
 }
 
 func Start() error {
--- a/etcdctl/util/help.go
+++ b/etcdctl/util/help.go
@@ -1,4 +1,4 @@
-// Copyright 2015 The etcd Authors
+// Copyright 2025 The etcd Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@
 
 // copied from https://github.com/rkt/rkt/blob/master/rkt/help.go
 
-package cobrautl
+package util
 
 import (
 	"fmt"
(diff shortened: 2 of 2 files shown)


ivanvc:
Thanks for the pull request, @ah8ad3. IMO, this change makes more sense in its own context than in the other pull request.

Looking at the test failure, it seems like you need to rebase your branch. Can you do it so we can start moving forward? Thanks again.

ivanvc (approved):
LGTM. Thanks, @ah8ad3.
cc. @jmhbnz, @ahrtr

jmhbnz (approved):
LGTM

ahrtr:
The only concern is that `UsageFunc` is an exported function, which means it might be used by any application, although it is unlikely in practice.

https://github.com/etcd-io/etcd/blob/ad942b7118b0b2b640ea7bba2f1c45c39211f4c1/pkg/cobrautl/help.go#L154

Please let's 
- either don't change it( close this PR);
- or add a deprecation note, do not remove it until 3.7.

ah8ad3:
> The only concern is that UsageFunc is an exported function, which means it might be used by any application, although it is unlikely in practice.

That is super unlikely for someone to use this, but you are right to ensure we don't break anybody we should do this. 
Personally i think removing `cobra` from there is needed because both the main `pkg` become lighter and no other module needs it there.

I will try to add deprecation note.

ivanvc:
Covered during our triage meeting. @ah8ad3, are you looking into doing Benjamin's suggestion? Or should we close this pull request?

ah8ad3:
Thanks for looking into this @ivanvc , and sorry for delaying this task.
I will take care of the comment and will add the note.

ah8ad3:
I can not understand what the problem is with verify failing, am i missing some sort of convention in making a function deprecated?
Can you please help? @ivanvc

pkg/cobrautl/help.go
```diff
@@ -151,6 +151,7 @@ func getSubCommands(cmd *cobra.Command) []*cobra.Command {
 	return subCommands
 }
 
+// Deprecated: Please use go.etcd.io/etcd/etcdctl/v3/util instead.
```
ahrtr, line 154: The linter error is complaining that the comment isn't starting with `UsageFunc`
```suggestion
// UsageFunc is the usage function for the cobra command.
// Deprecated: Please use the UsageFunc in go.etcd.io/etcd/etcdctl/v3/util instead.
```

ah8ad3: Oh, I see thanks.

ahrtr (approved):

ahrtr:
can you also add a changelog deprecation item for 3.7 (please add a CHANGELOG-3.7.md file), similar to [test_20250216](https://github.com/ahrtr/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#deprecations)? thx

ah8ad3:
> can you also add a changelog deprecation item for 3.7 (please add a CHANGELOG-3.7.md file), similar to [CHANGELOG-3.6.md#deprecations](https://github.com/ahrtr/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#deprecations)? thx

Of course i will do that as well @ahrtr

PR #18359, opened 2024-07-23
migrate experimental-stop-grpc-service-on-defrag flag to feature gate.

siyuanfoundation:

part of https://github.com/etcd-io/etcd/issues/18023, an example of basic wiring from an existing experimental feature flag to the new feature gate.

final step of of breaking up https://github.com/etcd-io/etcd/pull/18234/ into smaller PRs.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -193,9 +193,6 @@ type ServerConfig struct {
 	// a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
-	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
-	ExperimentalStopGRPCServiceOnDefrag bool `json:"experimental-stop-grpc-service-on-defrag"`
-
 	// ExperimentalBootstrapDefragThresholdMegabytes is the minimum number of megabytes needed to be freed for etcd server to
 	// consider running defrag during bootstrap. Needs to be set to non-zero value to take effect.
 	ExperimentalBootstrapDefragThresholdMegabytes uint `json:"experimental-bootstrap-defrag-threshold-megabytes"`
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -805,6 +805,25 @@ func (cfg *configYAML) configFromFile(path string) error {
 		}
 	}
 
+	// parses the yaml bytes to raw map first, then getBoolFlagVal can get the top level bool flag value.
+	var cfgMap map[string]interface{}
+	err = yaml.Unmarshal(b, &cfgMap)
+	if err != nil {
+		return err
+	}
+	getBoolFlagVal := func(flagName string) *bool {
+		flagVal, ok := cfgMap[flagName]
+		if !ok {
+			return nil
+		}
+		boolVal := flagVal.(bool)
+		return &boolVal
+	}
+	err = SetFeatureGatesFromExperimentalFlags(cfg.ServerFeatureGate, getBoolFlagVal, cfg.configJSON.ServerFeatureGatesJSON)
+	if err != nil {
+		return err
+	}
+
 	if cfg.configJSON.ListenPeerURLs != "" {
 		u, err := types.NewURLs(strings.Split(cfg.configJSON.ListenPeerURLs, ","))
 		if err != nil {
@@ -895,6 +914,36 @@ func (cfg *configYAML) configFromFile(path string) error {
 		cfg.SelfSignedCertValidity = 1
 	}
 	return cfg.Validate()
+}
+
+// SetFeatureGatesFromExperimentalFlags sets the feature gate values if the feature gate is not explicitly set
+// while their corresponding experimental flags are explicitly set, for all the features in ExperimentalFlagToFeatureMap.
+// TODO: remove after all experimental flags are deprecated.
+func SetFeatureGatesFromExperimentalFlags(fg featuregate.FeatureGate, getExperimentalFlagVal func(string) *bool, featureGatesVal string) error {
+	m := make(map[featuregate.Feature]bool)
+	// verify that the feature gate and its experimental flag are not both set at the same time.
+	for expFlagName, featureName := range features.ExperimentalFlagToFeatureMap {
+		flagVal := getExperimentalFlagVal(expFlagName)
+		if flagVal == nil {
+			continue
+		}
+		if strings.Contains(featureGatesVal, string(featureName)) {
+			return fmt.Errorf("cannot specify both flags: --%s=%v and --%s=%s=%v at the same time, please just use --%s=%s=%v",
+				expFlagName, *flagVal, ServerFeatureGateFlagName, featureName, fg.Enabled(featureName), ServerFeatureGateFlagName, featureName, fg.Enabled(featureName))
+		}
+		m[featureName] = *flagVal
+	}
+
+	// filter out unknown features for fg, because we could use SetFeatureGatesFromExperimentalFlags both for
+	// server and cluster level feature gates.
+	allFeatures := fg.(featuregate.MutableFeatureGate).GetAll()
+	mFiltered := make(map[string]bool)
+	for k, v := range m {
+		if _, ok := allFeatures[k]; ok {
+			mFiltered[string(k)] = v
+		}
+	}
+	return fg.(featuregate.MutableFeatureGate).SetFromMap(mFiltered)
 }
 
 func updateCipherSuites(tls *transport.TLSInfo, ss []string) error {
--- a/server/embed/config_test.go
+++ b/server/embed/config_test.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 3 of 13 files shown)


server/embed/config.go
```diff
@@ -888,6 +888,36 @@ func (cfg *configYAML) configFromFile(path string) error {
 	return cfg.Validate()
 }
 
+// SetFeatureGatesFromExperimentalFlags sets the feature gate values if the feature gate is not explicitly set
+// while their corresponding experimental flags are explicitly set.
+// TODO: remove after all experimental flags are deprecated.
+func SetFeatureGatesFromExperimentalFlags(fg featuregate.FeatureGate, getExperimentalFlagVal func(string) (bool, bool), featureGatesFlagName, featureGatesVal string) error {
```
serathius, line 894: Mark as deprecated?

server/etcdmain/help.go
```diff
@@ -312,7 +312,7 @@ Experimental feature:
   --experimental-snapshot-catchup-entries
     Number of entries for a slow follower to catch up after compacting the raft storage entries.
   --experimental-stop-grpc-service-on-defrag
-    Enable etcd gRPC service to stop serving client requests on defragmentation.
+    Enable etcd gRPC service to stop serving client requests on defragmentation. TO BE DEPRECATED, use '--feature-gates=StopGRPCServiceOnDefrag=true' instead.
```
serathius, line 315: To be deprecated when? and removed when?

server/etcdmain/config.go
```diff
@@ -238,6 +239,30 @@ func (cfg *config) configFromCmdLine() error {
 		cfg.ec.InitialCluster = ""
 	}
 
+	// getBoolFlagVal checks if the a given flag is explicitly set in the cmd line arguments.
+	getBoolFlagVal := func(flagName string) (bool, bool) {
```
serathius, line 243:
```suggestion
	getBoolFlagVal := func(flagName string) *bool {
```

Three value logic (true, false, unset) here would better map to pointer instead of two unnamed bool

serathius, line 243: Why not normal function?

server/config/config.go
```diff
@@ -193,9 +193,6 @@ type ServerConfig struct {
 	// a shared buffer in its readonly check operations.
 	ExperimentalTxnModeWriteWithSharedBuffer bool `json:"experimental-txn-mode-write-with-shared-buffer"`
 
-	// ExperimentalStopGRPCServiceOnDefrag enables etcd gRPC service to stop serving client requests on defragmentation.
-	ExperimentalStopGRPCServiceOnDefrag bool `json:"experimental-stop-grpc-service-on-defrag"`
```
serathius, line 197: Would removing this be a breaking change for json config or embeded cluster? (just checking).

server/etcdmain/config_test.go
```diff
@@ -412,16 +412,60 @@ func TestParseFeatureGateFlags(t *testing.T) {
...
+			name: "can set feature gate to true from feature gate flag",
+			args: []string{
+				"--feature-gates=StopGRPCServiceOnDefrag=true",
+			},
+			expectedFeatures: map[featuregate.Feature]bool{
 				features.StopGRPCServiceOnDefrag: true,
 			},
 		},
+		{
+			name: "can set feature gate to false from feature gate flag",
```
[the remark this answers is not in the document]
siyuanfoundation: done

server/embed/config.go
siyuanfoundation: Marking `ExperimentalFlagToFeatureMap` as deprecated, because that's where the actual experimental features are added.

server/etcdmain/config.go
```diff
@@ -238,14 +244,44 @@ func (cfg *config) configFromCmdLine() error {
...
+	})
+	if !explicitlySet {
+		return nil
+	}
+	flagVal, parseErr := strconv.ParseBool(cfg.cf.flagSet.Lookup(flagName).Value.String())
+	if parseErr != nil {
+		panic(parseErr)
+	}
+	return &flagVal
+}
```
ahrtr, line 274: We can move this into https://github.com/etcd-io/etcd/blob/main/pkg/flags/flag.go as a generic function something like below,

```
func GetBoolFlagVal(fs *flag.FlagSet, flagName string) *bool {
    ......
}
```

server/etcdmain/config.go
```diff
@@ -76,8 +78,12 @@ type configFlags struct {
 }
 
 func newConfig() *config {
+	return newConfigFromEmbedConfig(embed.NewConfig())
+}
+
+func newConfigFromEmbedConfig(ec *embed.Config) *config {
```
ahrtr, line 84: The implementation of `newConfig` is simple enough, let's try not to do such unnecessary change.

server/etcdmain/config.go
```diff
@@ -238,14 +244,44 @@ func (cfg *config) configFromCmdLine() error {
...
+	}
+	return &flagVal
+}
+
 func (cfg *config) configFromFile(path string) error {
 	eCfg, err := embed.ConfigFromFile(path)
 	if err != nil {
 		return err
 	}
+	// TODO: remove after adding feature-gates field in the config file.
```
ahrtr, line 281: This was missed in your previous PR, I would suggest to get this resolved firstly. 

This PR is already a followup PR, let's try not add too much followup PR in the chain.

siyuanfoundation: ok. Let's do https://github.com/etcd-io/etcd/pull/18365 first.

server/etcdmain/config.go
siyuanfoundation: done.

server/etcdmain/config.go
siyuanfoundation: done.

server/embed/config.go
```diff
@@ -897,6 +916,36 @@ func (cfg *configYAML) configFromFile(path string) error {
 	return cfg.Validate()
 }
 
+// SetFeatureGatesFromExperimentalFlags sets the feature gate values if the feature gate is not explicitly set
+// while their corresponding experimental flags are explicitly set, for all the features in ExperimentalFlagToFeatureMap.
+// TODO: remove after all experimental flags are deprecated.
+func SetFeatureGatesFromExperimentalFlags(fg featuregate.FeatureGate, getExperimentalFlagVal func(string) *bool, featureGatesFlagName, featureGatesVal string) error {
```
ahrtr, line 922: You should be able to read `embed.ServerFeatureGateFlagName` directly.
```suggestion
func SetFeatureGatesFromExperimentalFlags(fg featuregate.FeatureGate, getExperimentalFlagVal func(string) *bool, featureGatesVal string) error {
```

siyuanfoundation: this is passed as a param because we might want to reuse this function for `cluster-feature-gates`.

ahrtr: We haven't finalized the cluster-level feature yet. We can make change later once it's finalized.

ahrtr:
cc @fuweid @ivanvc  @jmhbnz @serathius

server/embed/config.go
siyuanfoundation: ok, removed.

fuweid (approved):
LGTM

server/embed/config_test.go
```diff
@@ -743,3 +778,99 @@ func TestUndefinedAutoCompactionModeValidate(t *testing.T) {
...
+			err = SetFeatureGatesFromExperimentalFlags(fg, getExperimentalFlagVal, tc.featureGatesFlag)
+			if tc.expectErr {
+				if err == nil {
+					t.Fatal("expect error")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatal(err)
+			}
```
ahrtr, line 868:
```suggestion			
			if tc.experimentalStopGRPCServiceOnDefrag != "" {
				// overwrite the value for `features.StopGRPCServiceOnDefrag`
				// based on flag `--experimental-stop-grpc-service-on-defrag`.
				getExperimentalFlagVal := func(flagName string) *bool {
					if flagName != "experimental-stop-grpc-service-on-defrag" {
						return nil
					}
					flagVal, parseErr := strconv.ParseBool(tc.experimentalStopGRPCServiceOnDefrag)
					require.NoError(t, parseErr)
					return &flagVal
				}

				err = SetFeatureGatesFromExperimentalFlags(fg, getExperimentalFlagVal, tc.featureGatesFlag)
				if tc.expectErr {
					require.True(t, err != nil, "expect error")
					return
				}
				require.NoError(t, err)
			}			
```

ahrtr: > This misses out the test case when the flag is not set.

You can add a separate case for that. Always remember to improve the readability and keep the test case easy to understand

siyuanfoundation: personally I don't think putting the function this test is mainly about inside a if block is improving the readability. 
Added some comments, hopefully that helps.

ahrtr: 1. It isn't about a if block, it's about not coupling a test driving condition `tc.experimentalStopGRPCServiceOnDefrag != ""` with a hard code condition `flagName != "experimental-stop-grpc-service-on-defrag"`. Putting the former in a outer / upper layer gives a clear hint what it's for, so easier to understand.
2. The special missed test case you mentioned above deserves a dedicated test case; do not implicitly test it in this test, otherwise other contributors may update this test in future and miss it again.
3. Use require.NoError to get rid of too many `if err != nil` or `if err == nil`

siyuanfoundation: updated code to `require.NoError`

ahrtr (approved):
LGTM

Thanks

ahrtr:
cc  @jmhbnz @serathius

serathius (approved):

PR #18360, opened 2024-07-24
Fix passing default grpc call options in Kubernetes client

serathius:
Totally slipped past me, only found this because K8s has integration tests that expect client side grpc error for size of request. I didn't catch it as they are not located in `apiserver` package, but in `controlplane`. When iterating I assumed that all relevant tests are in apiserver package.

cc @ahrtr

### DIFF
--- a/client/v3/kubernetes/client.go
+++ b/client/v3/kubernetes/client.go
@@ -16,6 +16,7 @@ package kubernetes
 
 import (
 	"context"
+	"fmt"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
@@ -31,7 +32,6 @@ func New(cfg clientv3.Config) (*Client, error) {
 	}
 	kc := &Client{
 		Client: c,
-		kv:     clientv3.RetryKVClient(c),
 	}
 	kc.Kubernetes = kc
 	return kc, nil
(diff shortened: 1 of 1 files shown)


client/v3/kubernetes/client.go
```diff
@@ -31,24 +31,20 @@ func New(cfg clientv3.Config) (*Client, error) {
...
 		Client: c,
-		kv:     clientv3.RetryKVClient(c),
 	}
-	kc.Kubernetes = kc
 	return kc, nil
 }
 
 type Client struct {
 	*clientv3.Client
-	Kubernetes Interface
```
ahrtr, line 42: Why you remove `Kubernetes` now? If you remove it, then you need to provide another way to return a `Kubernetes` interface to client.

client/v3/kubernetes/client.go
```diff
@@ -31,24 +31,20 @@ func New(cfg clientv3.Config) (*Client, error) {
...
-	kv         pb.KVClient
 }
 
 var _ Interface = (*Client)(nil)
 
 func (k Client) Get(ctx context.Context, key string, opts GetOptions) (resp GetResponse, err error) {
-	rangeResp, err := k.kv.Range(ctx, getRequest(key, opts.Revision))
+	rangeResp, err := k.KV.Get(ctx, key, clientv3.WithRev(opts.Revision), clientv3.WithLimit(1))
```
ahrtr, line 45: minor comment, either way works, but `k.Get` is a little simpler. This comment applies to some other similar places as well. 
```suggestion
	rangeResp, err := k.Get(ctx, key, clientv3.WithRev(opts.Revision), clientv3.WithLimit(1))
```

serathius: That leads to infinite recursion.

ahrtr (approved):
LGTM

It's much simpler now! Thanks

PR #18369, opened 2024-07-26
tests/e2e: add test cases related to HashKV

fuweid:
Related to https://github.com/etcd-io/etcd/pull/18274

### DIFF
--- a/tests/e2e/hashkv_test.go
+++ b/tests/e2e/hashkv_test.go
@@ -1,0 +1,234 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestVerifyHashKVAfterCompact(t *testing.T) {
+	scenarios := []struct {
+		clusterVersion e2e.ClusterVersion
+		keys           []string // used for data generators
+	}{
+		{
+			clusterVersion: e2e.CurrentVersion,
+			keys:           []string{"key0"},
+		},
+		{
+			clusterVersion: e2e.CurrentVersion,
+			keys:           []string{"key0", "key1"},
+		},
+		{
+			clusterVersion: e2e.QuorumLastVersion,
+			keys:           []string{"key0"},
+		},
+		{
+			clusterVersion: e2e.QuorumLastVersion,
+			keys:           []string{"key0", "key1"},
+		},
+	}
+
+	for _, compactedOnTombstoneRev := range []bool{false, true} {
+		for _, scenario := range scenarios {
+			t.Run(fmt.Sprintf("compactedOnTombstone=%v - %s - Keys=%v", compactedOnTombstoneRev, scenario.clusterVersion, scenario.keys), func(t *testing.T) {
+				e2e.BeforeTest(t)
+
+				if scenario.clusterVersion != e2e.CurrentVersion {
+					if !fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
+						t.Skipf("%q does not exist", e2e.BinPath.EtcdLastRelease)
+					}
+				}
+
+				ctx := context.Background()
+
+				cfg := e2e.NewConfigClientTLS()
+				clus, err := e2e.NewEtcdProcessCluster(ctx, t,
+					e2e.WithConfig(cfg),
+					e2e.WithClusterSize(3),
+					e2e.WithVersion(scenario.clusterVersion))
+				require.NoError(t, err)
+
+				t.Cleanup(func() { clus.Close() })
+
+				tombstoneRevs, lastestRev := populateDataForHashKV(t, clus, cfg.Client, scenario.keys)
+
+				compactedOnRev := tombstoneRevs[0]
+
+				// If compaction revision isn't a tombstone, select a revision in the middle of two tombstones.
+				if !compactedOnTombstoneRev {
+					compactedOnRev = (tombstoneRevs[0] + tombstoneRevs[1]) / 2
+					require.Greater(t, tombstoneRevs[1], compactedOnRev)
+				}
+
+				cli, err := e2e.NewEtcdctl(cfg.Client, clus.EndpointsGRPC())
+				require.NoError(t, err)
+
+				t.Logf("COMPACT on rev=%d", compactedOnRev)
+				_, err = cli.Compact(ctx, compactedOnRev, config.CompactOption{Physical: true})
+				require.NoError(t, err)
+
+				for rev := compactedOnRev; rev <= lastestRev; rev++ {
+					verifyConsistentHashKVAcrossAllMembers(t, cli, rev)
+				}
+			})
+		}
+	}
+}
+
+func TestVerifyHashKVAfterTwoCompactionsOnTombstone_MixVersions(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	if !fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
+		t.Skipf("%q does not exist", e2e.BinPath.EtcdLastRelease)
+	}
+
+	ctx := context.Background()
+
+	cfg := e2e.NewConfigClientTLS()
+	clus, err := e2e.NewEtcdProcessCluster(ctx, t,
+		e2e.WithConfig(cfg),
+		e2e.WithClusterSize(3),
+		e2e.WithVersion(e2e.QuorumLastVersion))
+	require.NoError(t, err)
+	t.Cleanup(func() { clus.Close() })
+
+	tombstoneRevs, lastestRev := populateDataForHashKV(t, clus, cfg.Client, []string{"key0"})
+
+	cli, err := e2e.NewEtcdctl(cfg.Client, clus.EndpointsGRPC())
+	require.NoError(t, err)
+
+	firstCompactOnRev := tombstoneRevs[0]
+	t.Logf("COMPACT rev=%d", firstCompactOnRev)
+	_, err = cli.Compact(ctx, firstCompactOnRev, config.CompactOption{Physical: true})
+	require.NoError(t, err)
+
+	secondCompactOnRev := tombstoneRevs[1]
+	t.Logf("COMPACT rev=%d", secondCompactOnRev)
+	_, err = cli.Compact(ctx, secondCompactOnRev, config.CompactOption{Physical: true})
+	require.NoError(t, err)
+
+	for rev := secondCompactOnRev; rev <= lastestRev; rev++ {
+		verifyConsistentHashKVAcrossAllMembers(t, cli, rev)
+	}
+}
+
+func TestVerifyHashKVAfterCompactionOnLastTombstone_MixVersions(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	if !fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
+		t.Skipf("%q does not exist", e2e.BinPath.EtcdLastRelease)
+	}
+
+	for _, keys := range [][]string{
+		[]string{"key0"},
+		[]string{"key0", "key1"},
+	} {
+		t.Run(fmt.Sprintf("#%v", keys), func(t *testing.T) {
+			ctx := context.Background()
+
+			cfg := e2e.NewConfigClientTLS()
+			clus, err := e2e.NewEtcdProcessCluster(ctx, t,
+				e2e.WithConfig(cfg),
+				e2e.WithClusterSize(3),
+				e2e.WithVersion(e2e.QuorumLastVersion))
+			require.NoError(t, err)
+			t.Cleanup(func() { clus.Close() })
+
+			tombstoneRevs, lastestRev := populateDataForHashKV(t, clus, cfg.Client, keys)
+
+			cli, err := e2e.NewEtcdctl(cfg.Client, clus.EndpointsGRPC())
+			require.NoError(t, err)
+
+			compactOnRev := tombstoneRevs[len(tombstoneRevs)-1]
+			t.Logf("COMPACT rev=%d", compactOnRev)
+			_, err = cli.Compact(ctx, compactOnRev, config.CompactOption{Physical: true})
+			require.NoError(t, err)
+
+			for rev := compactOnRev; rev <= lastestRev; rev++ {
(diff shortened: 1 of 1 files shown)


tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+		resp, terr := c.Txn(ctx).Then(ops...).Commit()
+		require.NoError(t, terr)
+		require.True(t, resp.Succeeded)
+		require.Len(t, resp.Responses, len(ops))
+		lastRevision = resp.Header.Revision
+	}
+}
+
+func TestVerifyHashKVAfterTwoCompactions_MixVersions(t *testing.T) {
```
fuweid, line 181: Side note: If https://github.com/etcd-io/etcd/pull/18274 is accepted, the main branch always keep compacted revision. However, existing releases delete that compacted revisions if they're tombstone.  The #18274 should skip previous compacted revision which is tombstone. So that HashKV result can be the same in mix versions, especially it won't file data corrupt alert when cluster is updating to new release.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,314 @@
...
+	fe2e "go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestVerifyHashKVAfterCompact(t *testing.T) {
+	tests := []struct {
+		compactedOnRev int64
+		hashKVOnRev    int64
+	}{
+		{
+			compactedOnRev: 33, // tombstone
```
serathius, line 38: Please don't use comments to imply some relation of number here and consts in `newTestKeySetInCluster`. Passing an explicit variable should always be better.

fuweid: Rollback to original one. Please take a look. If it's accepted, I will squash the commits. Thanks

fuweid:
ping @ahrtr

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+		},
+		{
+			compactedOnTombstoneRev: false,
+			hashKVOnTombstoneRev:    false,
+		},
+	}
+
+	scenarios := []struct {
+		ClusterVersion fe2e.ClusterVersion
+		OnlyOneKey     bool
```
serathius, line 57: Can you provide more context what `OnlyOneKey` means?

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+func newClusterForHashKV(t *testing.T, clusVersion fe2e.ClusterVersion) (*fe2e.EtcdProcessCluster, fe2e.ClientConfig) {
+	if clusVersion != fe2e.CurrentVersion {
+		if !fileutil.Exist(fe2e.BinPath.EtcdLastRelease) {
+			t.Skipf("%q does not exist", fe2e.BinPath.EtcdLastRelease)
+		}
+	}
+
+	ctx := context.Background()
+
+	fe2e.BeforeTest(t)
```
ahrtr, line 347: These should be in the test case directly instead of a helper function.

I don't think we need `newClusterForHashKV` at all.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+				dataset := newTestDatasetInCluster(t, clus, cliCfg, scenario.OnlyOneKey)
+
+				// ensure we can pick up the right revision
+				require.True(t, len(dataset.tombstones) > 2)
+				require.True(t, dataset.tombstones[0]+3 < dataset.tombstones[1])
+				require.True(t, dataset.tombstones[1]+3 < dataset.tombstones[2])
+
+				compactedOnRev := dataset.tombstones[0]
+				if !tt.compactedOnTombstoneRev {
+					compactedOnRev += 3
```
serathius, line 93: What's the significance of number `3` here? Can we create a local const? Like `numberOfRevisions` or `emptyWrites`.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+				require.Equal(t, resp[0].Hash, resp[1].Hash)
+				require.Equal(t, resp[1].Hash, resp[2].Hash)
+			})
+		}
+	}
+}
+
+type datasetInfo struct {
+	keys map[string]struct{}
```
ahrtr, line 124: You just populate the data for `keys`, but it isn't used at all.

We can remove `datasetInfo`, and just return `tombstones` and `latestRev` in the functions of populating data.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+				t.Logf("HashKV on rev=%d", hashKVOnRev)
+				resp, err := cli.HashKV(ctx, hashKVOnRev)
+				require.NoError(t, err)
+
+				require.Len(t, resp, 3)
+				require.True(t, resp[0].Hash != 0)
+				t.Logf("One Hash value is %d", resp[0].Hash)
+
+				require.Equal(t, resp[0].Hash, resp[1].Hash)
+				require.Equal(t, resp[1].Hash, resp[2].Hash)
```
ahrtr, line 117: minor comment, we can add a helper function something `verifyConsistentHashAcrossAllMembers`

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	fe2e "go.etcd.io/etcd/tests/v3/framework/e2e"
```
serathius, line 29: Why not just e2e? All other test just import this as e2e.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+type datasetInfo struct {
+	keys map[string]struct{}
+
+	tombstones      []int64
+	lastestRevision int64
+}
+
+// key: "foo"
```
serathius, line 130: Thanks for crafting such a meticulous testing scenario, this seems very throughout, however for reviewer it's or any future contributor it might be hard to understand why we picked this exact data shape and no other. 

My question would be, how hard would it be to generalize? We know one an exact case that we know that we could break hashKV, but how sure we are that we don't miss another case? Hardcoding a exact scenario protects us against this exact mistake, but can we be sure there will not be more such cases in the future? Can we be sure that HashKV, is resilient enough?

Have you heard about property testing? https://www.mayhem.security/blog/what-is-property-based-testing. For me the etcd should always hold property that no matter the type of write, no matter the compaction revision, no matter the etcd version, no matter the revision, HashKV should always be the same on all cluster members.

I think it would be ok to merge those exact scenarios, but I also we should think how we can generalize it. For example, instead of checking hashKV on one revision, test all of them in for loop.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,359 @@
...
+// generations:
+//
+//	{23, 1}
+//
+// key: "foo-3"
+// modified: 34
+// generations:
+//
+//	{34, 1}
+func newTestDatasetInCluster(t *testing.T, clus *fe2e.EtcdProcessCluster, cliCfg fe2e.ClientConfig, onlyOneKey bool) *datasetInfo {
```
ahrtr, line 158: It's unnecessary complicated.

Please consider a function something like below,
```
// populateData populates some sample data, and return a slice of tombstone revisions and the latest revision
func populateData(t *testing.T, clus *fe2e.EtcdProcessCluster, clientCfg fe2e.ClientConfig, keys []string) ([]int64, int64) {
	c := newClient(t, clus.EndpointsGRPC(), clientCfg)

	ctx := context.Background()
	totalOperations := 40

	var (
		tombStoneRevs []int64
		latestRev     int64
	)

	deleteStep := 10 // submit a delete operation on every 10 operations
	for i := 1; i <= totalOperations; i++ {
		if i%deleteStep == 0 {
			t.Logf("Deleting key=%s", keys[0]) // Only delete the first key for simplicity
			resp, derr := c.Delete(ctx, keys[0])
			require.NoError(t, derr)
			latestRev = resp.Header.Revision
			tombStoneRevs = append(tombStoneRevs, resp.Header.Revision)
			continue
		}

		value := fmt.Sprintf("%d", i)
		var ops []clientv3.Op
		for _, key := range keys {
			ops = append(ops, clientv3.OpPut(key, value))
		}

		t.Logf("Writing keys: %v, value: %s", keys, value)
		resp, terr := c.Txn(ctx).Then(ops...).Commit()
		require.NoError(t, terr)
		require.True(t, resp.Succeeded)
		require.Len(t, resp.Responses, len(ops))
		latestRev = resp.Header.Revision
	}

	return tombStoneRevs, latestRev
}
```

tests/e2e/hashkv_test.go
fuweid: > instead of checking hashKV on one revision, test all of them in for loop.

Good point. Will simplify data generator.

tests/e2e/hashkv_test.go
fuweid: OK. I removed it.

tests/e2e/hashkv_test.go
fuweid: Thanks. I used your code in latest commit. Please take a look.

tests/e2e/hashkv_test.go
fuweid: Done.

tests/e2e/hashkv_test.go
fuweid: I follow @ahrtr suggestion with using `populateData`. Please take a look.

tests/e2e/hashkv_test.go
fuweid: Updated.

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,195 @@
...
+	t.Logf("HashKV on rev=%d", hashKVOnRev)
+	resp, err := cli.HashKV(ctx, hashKVOnRev)
+	require.NoError(t, err)
+
+	require.Len(t, resp, 3)
+	require.True(t, resp[0].Hash != 0)
+	t.Logf("One Hash value is %d", resp[0].Hash)
+
+	require.Equal(t, resp[0].Hash, resp[1].Hash)
+	require.Equal(t, resp[1].Hash, resp[2].Hash)
```
ahrtr, line 194:
```suggestion	
	require.Greater(t, len(resp), 1)
	require.True(t, resp[0].Hash != 0)
	t.Logf("One Hash value is %d", resp[0].Hash)

	for i := 1; i < len(resp); i++ {
		require.Equal(t, resp[0].Hash, resp[i].Hash)
	}	
```

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+				compactedOnRev := tombstoneRevs[0]
+
+				// If compaction revision is not tombstone, select revision after 3 writes from first tombstone.
+				// And ensure it's not the following tombstone.
+				const afterWriters = int64(3)
+				if !compactedOnTombstoneRev {
+					compactedOnRev += afterWriters
+					require.True(t, tombstoneRevs[1] > compactedOnRev)
+				}
```
ahrtr, line 87: The `3` may cause unnecessary confusion. We don't care about the number at all, we just need to ensure the compaction revision isn't a tombstone.
```suggestion
				// If compaction revision isn't a tombstone, select a revision in the middle of two tombstones.
				if !compactedOnTombstoneRev {
					compactedOnRev = (tombstoneRevs[0] + tombstoneRevs[1]) / 2
					require.True(t, tombstoneRevs[0] < compactedOnRev && compactedOnRev < tombstoneRevs[1])
				}				
```

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,195 @@
...
+					e2e.WithConfig(cfg),
+					e2e.WithClusterSize(3),
+					e2e.WithVersion(scenario.clusterVersion))
+				require.NoError(t, err)
+
+				t.Cleanup(func() { clus.Close() })
+
+				tombstoneRevs, lastestRev := populateDataForHashKV(t, clus, cfg.Client, scenario.keys)
+
+				compactedOnRev := tombstoneRevs[0]
```
ahrtr, line 79: Can we also add a scenario which compact the last tombstone `tombstoneRevs[len(tombstoneRevs) - 1]`?

tests/e2e/hashkv_test.go
fuweid: Updated.

tests/e2e/hashkv_test.go
fuweid: Added `TestVerifyHashKVAfterCompactionOnLastTombstone_MixVersions`

tests/e2e/hashkv_test.go
fuweid: thanks for the suggestion. updated

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+				// If compaction revision is not tombstone, select revision after 3 writes from first tombstone.
+				// And ensure it's not the following tombstone.
```
ahrtr, line 82: minor comment
```suggestion
				// If compaction revision isn't a tombstone, select a revision in the middle of two tombstones.
```

fuweid: my mistake. Updated

ahrtr:
Ideally, we should consolidate the two cases, `TestVerifyHashKVAfterCompact` and `TestVerifyHashKVAfterCompactionOnLastTombstone_MixVersions`, into one. But I won't insist on that because the PR is already good enough.

ahrtr (approved):
LGTM

Nice work, thanks @fuweid

tests/e2e/hashkv_test.go
```diff
@@ -0,0 +1,234 @@
...
+				t.Cleanup(func() { clus.Close() })
+
+				tombstoneRevs, lastestRev := populateDataForHashKV(t, clus, cfg.Client, scenario.keys)
+
+				compactedOnRev := tombstoneRevs[0]
+
+				// If compaction revision isn't a tombstone, select a revision in the middle of two tombstones.
+				if !compactedOnTombstoneRev {
+					compactedOnRev = (tombstoneRevs[0] + tombstoneRevs[1]) / 2
+					require.Greater(t, tombstoneRevs[1], compactedOnRev)
```
serathius, line 84:
```suggestion
					require.Greater(t, tombstoneRevs[1], tombstoneRevs[0])
```

serathius: And move this up.

serathius (approved):
Looks good to give us confidence in fixing the tombstone compaction bug. Thanks @fuweid awesome work.

I think we should also consider generalizing it by adding hashKV checking to robustness tests.

tests/e2e/hashkv_test.go
ahrtr, line 84: nit: It seems we partially missed the comment https://github.com/etcd-io/etcd/pull/18369#discussion_r1695462341


```
require.True(t, tombstoneRevs[0] < compactedOnRev && compactedOnRev < tombstoneRevs[1])
```

fuweid:
Thanks @ahrtr @serathius for the review. I file pr https://github.com/etcd-io/etcd/pull/18387 to fix the comment https://github.com/etcd-io/etcd/pull/18369#discussion_r1698144827.

fuweid:
ping @ahrtr could we remove label `backport/v3.4`? The #18476 covers mixed cluster versions [3.4 and 3.5].

ahrtr:
> ping @ahrtr could we remove label `backport/v3.4`? The #18476 covers mixed cluster versions [3.4 and 3.5].

Removed.

I was thinking the e2e test cases should also support different patch versions, e.g 3.4.33 vs 3.4.34+, or 3.5.15 vs 3.5.16+. But I do not see a simple/maintainable way to do it.

PR #18373, opened 2024-07-27
Fix wait time docs, tests to indicate trigger deadline geq to wait deadline.

redwrasse:
Wait time implementation in fact supports a trigger deadline that is greater than or equal to wait deadlines, rather than greater than wait deadlines. Updates docs and a couple tests accordingly.

### DIFF
--- a/pkg/wait/wait_time_test.go
+++ b/pkg/wait/wait_time_test.go
@@ -57,7 +57,7 @@ func TestWaitTestStress(t *testing.T) {
 	for i := 0; i < 10000; i++ {
 		chs = append(chs, wt.Wait(uint64(i)))
 	}
-	wt.Trigger(10000 + 1)
+	wt.Trigger(10000 - 1)
 
 	for _, ch := range chs {
 		select {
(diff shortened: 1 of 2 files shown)


pkg/wait/wait_time_test.go
```diff
@@ -57,7 +57,7 @@ func TestWaitTestStress(t *testing.T) {
 	for i := 0; i < 10000; i++ {
 		chs = append(chs, wt.Wait(uint64(i)))
 	}
-	wt.Trigger(10000 + 1)
+	wt.Trigger(10000 - 1)
```
jmhbnz, line 60: Do we need the math operation here?  Could we just change the value to `9999`? Not a blocker it just took me a moment to review what the test was trying to do.

ahrtr: something like below better?
```
	for i := 0; i <= 9999; i++ {
		chs = append(chs, wt.Wait(uint64(i)))
	}
	wt.Trigger(9999)
```

redwrasse: Agreed this is clearest, I'll update the MR.

ahrtr (approved):

pkg/wait/wait_time_test.go
redwrasse: Updated indices in tests.

jmhbnz (approved):
LGTM - Thanks for tidying this up @redwrasse 🙏🏻

Issue #18240, opened 2024-06-27
Failpoint `raftAfterSave=sleep(1s)` is flaking in robustness test


serathius:
### Which Github Action / Prow Jobs are flaking?

https://testgrid.k8s.io/sig-etcd-robustness#ci-etcd-robustness-amd64

### Which tests are flaking?

TestRobustnessExploratoryKubernetesHighTrafficClusterOfSize3

### Github Action / Prow Job link

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-etcd-robustness-main-amd64/1806029033377370112

### Reason for failure (if possible)

```
    logger.go:146: 2024-06-26T19:45:53.423Z	INFO	goFailpoint deactivate failed	{"failpoint": "raftAfterSave=sleep(1s)", "error": "Delete \"http://127.0.0.1:12381/raftAfterSave\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"}
    logger.go:146: 2024-06-26T19:45:53.423Z	ERROR	Failed to trigger failpoint	{"failpoint": "raftAfterSave=sleep(1s)", "error": "goFailpoint raftAfterSave=sleep(1s) deactivate failed, err: Delete \"http://127.0.0.1:12381/raftAfterSave\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"}
    main_test.go:122: failed triggering failpoint, err: goFailpoint raftAfterSave=sleep(1s) deactivate failed, err: Delete "http://127.0.0.1:12381/raftAfterSave": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

MadhavJivrajani:
```
validate.go:34: Broken validation assumptions: non empty database at start or first write didn't succeed, required by model implementation
```
@serathius I also see this in the logs ^

MadhavJivrajani:
For the failpoint flake, we have the option of increasing the client timeout for that particular failpoint, but I'm still not entirely sure why its timing out to begin with.

serathius:
@henrybear327 when you will be able to send the PR? Just the timeout increase should be 1 line change.

henrybear327:
> @henrybear327 when you will be able to send the PR? Just the timeout increase should be 1 line change.

Done just now!

I will still continue my investigation on the gofail library to see if any issues are coming from the gofail v0.2.0 changes that I made.

henrybear327:
[...]

PR #18390, opened 2024-08-01
Increase raftAfterSave to 3 seconds

henrybear327:
Reference:
- https://github.com/etcd-io/etcd/issues/18240


serathius:
How did you test it? This increases the time failpoint is active, not the request timeout. Is this really fixing the issue?

henrybear327:
> How did you test it? This increases the time failpoint is active, not the request timeout. Is this really fixing the issue?

Oops I misread the requirement of increasing the `client timeout` instead of the `failpoint time`. Sorry about this. I will make the change accordingly ...

henrybear327:
With https://github.com/etcd-io/etcd/pull/18393, I guess we will not need to increase the client timeout in the meantime right @serathius?

PR #18392, opened 2024-08-01
Remove flake caused failpoint in watch disrupting progress notifies

serathius:
cc @ahrtr @siyuanfoundation @fuweid @MadhavJivrajani

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -96,8 +96,11 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, s testSce
 		t.Fatal(err)
 	}
 
-	watchProgressNotifyEnabled := c.Cfg.ServerConfig.ExperimentalWatchProgressNotifyInterval != 0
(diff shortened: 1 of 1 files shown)


ahrtr (approved):

tests/robustness/main_test.go
```diff
@@ -96,8 +96,11 @@ func testRobustness(ctx context.Context, t *testing.T, lg *zap.Logger, s testSce
 		t.Fatal(err)
 	}
 
-	watchProgressNotifyEnabled := c.Cfg.ServerConfig.ExperimentalWatchProgressNotifyInterval != 0
-	validateGotAtLeastOneProgressNotify(t, r.Client, s.watch.requestProgress || watchProgressNotifyEnabled)
+	failpointImpactingWatch := s.failpoint == failpoint.SleepBeforeSendWatchResponse
```
siyuanfoundation, line 99: this seems to be a important failpoint to test progress notify. Is there plan to investigate further?

serathius: No, with progress notifies not being guaranteed in every situation, it's hard to make a strong requirement on them when we use failpoints that directly disrupt their delivery. We can consider revisiting it only after we improve etcd progress guarantees.

Issue #18069, opened 2024-05-27
Performance Issue: CPU usage of the leader node increased by 20%


qixiaoyang0:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

In the ETCD cluster I tested, the lease renew rate is 1500/s, arm64 CPU, and the CPU usage of the leader node increased by 20%.

This problem is because of this modification
https://github.com/etcd-io/etcd/blob/bb701b9265f31d61db5906325e0a7e2abf7d3627/server/etcdserver/v3_server.go#L288

We conducted rigorous tests, including checking pprof, collecting metrics, deleting that line of code and recompiling etcd, and confirmed that this was the reason for the CPU increase.

### What did you expect to happen?
It may not be necessary to add leader confirmation in lease renew.
Or using a more performant method.

### How can we reproduce it (as minimally and precisely as possible)?

Sending many lease update requests to the ETCD cluster, v3.5.11 version CPU will increase a lot.

### Etcd version (please run commands below)

v3.5.11

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

jmhbnz:
Hi @qixiaoyang0 - Thanks for highlighting this. Adding some context, this performance impact was mentioned as part of the pr that introduced the line you mention. Refer https://github.com/etcd-io/etcd/pull/16822#issuecomment-1850490064.

The pr was required to fix https://github.com/etcd-io/etcd/issues/15247.

I am not sure if there will be much we can do to alleviate this currently, defer to @ahrtr and @serathius on any ideas for an alternative approach (if any).

ahrtr:
[...]

PR #18428, opened 2024-08-09
Skip leadership check if the etcd instance is active processing heartbeats

ahrtr:
Resolve https://github.com/etcd-io/etcd/issues/18069

Will post benchmark result later.

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -904,9 +904,25 @@ func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
 	})
 }
 
+// isActive checks if the etcd instance is still actively processing the
+// heartbeat message (ticks). It returns false if no heartbeat has been
+// received within 3 * tickMs.
+func (s *EtcdServer) isActive() bool {
+	latestTickTs := s.r.getLatestTickTs()
+	threshold := 3 * time.Duration(s.Cfg.TickMs) * time.Millisecond
+	return latestTickTs.Add(threshold).After(time.Now())
+}
+
 // ensureLeadership checks whether current member is still the leader.
 func (s *EtcdServer) ensureLeadership() bool {
 	lg := s.Logger()
+
+	if s.isActive() {
+		lg.Debug("The member is active, skip checking leadership",
+			zap.Time("latestTickTs", s.r.getLatestTickTs()),
+			zap.Time("now", time.Now()))
+		return true
+	}
 
 	ctx, cancel := context.WithTimeout(s.ctx, s.Cfg.ReqTimeout())
 	defer cancel()
--- a/server/etcdserver/raft.go
+++ b/server/etcdserver/raft.go
@@ -80,7 +80,9 @@ type toApply struct {
 type raftNode struct {
 	lg *zap.Logger
 
-	tickMu *sync.Mutex
+	tickMu *sync.RWMutex
+	// timestamp of the latest tick
+	latestTickTs time.Time
 	raftNodeConfig
 
 	// a chan to send/receive snapshot
@@ -132,8 +134,9 @@ func newRaftNode(cfg raftNodeConfig) *raftNode {
 	raft.SetLogger(lg)
 	r := &raftNode{
 		lg:             cfg.lg,
-		tickMu:         new(sync.Mutex),
+		tickMu:         new(sync.RWMutex),
 		raftNodeConfig: cfg,
+		latestTickTs:   time.Now(),
 		// set up contention detectors for raft heartbeat message.
 		// expect to send a heartbeat within 2 heartbeat intervals.
 		td:         contention.NewTimeoutDetector(2 * cfg.heartbeat),
@@ -155,7 +158,14 @@ func newRaftNode(cfg raftNodeConfig) *raftNode {
 func (r *raftNode) tick() {
 	r.tickMu.Lock()
 	r.Tick()
+	r.latestTickTs = time.Now()
 	r.tickMu.Unlock()
+}
+
+func (r *raftNode) getLatestTickTs() time.Time {
+	r.tickMu.RLock()
+	defer r.tickMu.RUnlock()
+	return r.latestTickTs
 }
 
 // start prepares and starts raftNode in a new goroutine. It is no longer safe
--- a/server/etcdserver/server_test.go
+++ b/server/etcdserver/server_test.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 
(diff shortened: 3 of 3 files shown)


ahrtr:
## Summary

This PR resolves the performance issue https://github.com/etcd-io/etcd/issues/18069

## Main branch

```
Summary:
  Total:	25.8324 secs.
  Slowest:	0.1114 secs.
  Fastest:	0.0001 secs.
  Average:	0.0024 secs.
  Stddev:	0.0030 secs.
  Requests/sec:	38711.1170

Response time histogram:
  0.0001 [1]	|
  0.0112 [986525]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0223 [10104]	|
  0.0335 [1905]	|
  0.0446 [712]	|
  0.0557 [247]	|
  0.0669 [243]	|
  0.0780 [234]	|
  0.0891 [21]	|
  0.1003 [5]	|
  0.1114 [3]	|

Latency distribution:
  10% in 0.0008 secs.
  25% in 0.0012 secs.
  50% in 0.0017 secs.
  75% in 0.0027 secs.
  90% in 0.0043 secs.
  95% in 0.0060 secs.
  99% in 0.0132 secs.
  99.9% in 0.0406 secs.
```

## This PR

```
Summary:
  Total:	14.8739 secs.
  Slowest:	0.0824 secs.
  Fastest:	0.0000 secs.
  Average:	0.0012 secs.
  Stddev:	0.0014 secs.
  Requests/sec:	67231.7847

Response time histogram:
  0.0000 [1]	|
  0.0083 [996078]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0165 [2981]	|
  0.0247 [548]	|
  0.0330 [185]	|
  0.0412 [99]	|
  0.0494 [26]	|
  0.0577 [41]	|
  0.0659 [39]	|
  0.0741 [0]	|
  0.0824 [2]	|

Latency distribution:
  10% in 0.0004 secs.
  25% in 0.0006 secs.
  50% in 0.0009 secs.
  75% in 0.0014 secs.
  90% in 0.0021 secs.
  95% in 0.0028 secs.
  99% in 0.0063 secs.
  99.9% in 0.0160 secs.
```

## Revert the https://github.com/etcd-io/etcd/pull/16822 (checked out `2f03bbc5dd34b4250c9baece58754e8180786bd4`)

```
Summary:
  Total:	14.6155 secs.
  Slowest:	0.0637 secs.
  Fastest:	0.0000 secs.
  Average:	0.0011 secs.
  Stddev:	0.0012 secs.
  Requests/sec:	68420.5991

Response time histogram:
  0.0000 [1]	|
  0.0064 [993920]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.0128 [4897]	|
  0.0191 [705]	|
  0.0255 [216]	|
  0.0319 [158]	|
  0.0383 [55]	|
  0.0446 [20]	|
  0.0510 [7]	|
  0.0574 [16]	|
  0.0637 [5]	|

Latency distribution:
  10% in 0.0004 secs.
  25% in 0.0006 secs.
  50% in 0.0009 secs.
  75% in 0.0013 secs.
  90% in 0.0019 secs.
  95% in 0.0025 secs.
  99% in 0.0054 secs.
  99.9% in 0.0136 secs.
```

ahrtr:
cc @fuweid @jmhbnz @serathius @qixiaoyang0 @ranandfigma @caesarxuchao

This PR resolves the performance issue with a minimum change.

ah8ad3 (commented):
Thanks for implementation looks clean. LGTM
I have one question: Imagine a scenario where leadership is lost due to two leaders fighting over with similar terms and equal votes.
Does that mean after merging this the new vote will be start with more delay? And possibly the cluster will be down?

I know this scenario is hard to produce i'm asking to find out the behavior.

ahrtr:
> I have one question: Imagine a scenario where leadership is lost due to two candidates fighting over with similar terms and equal votes.
> Does that mean after merging this the new vote will be start with more delay? And possibly the cluster will be down?

This PR has nothing to do with any raft's protocol. It just tries to avoid unnecessary `ensureLeadership` check, so as to avoid unnecessary performance reduce.

caesarxuchao (approved):
LGTM. Thanks for the quick fix!

server/etcdserver/server.go
```diff
@@ -904,10 +904,26 @@ func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
 	})
 }
 
+// isActive checks if the etcd instance is still actively processing the
+// heartbeat message (ticks). It returns false if no heartbeat has been
+// received within 3 * tickMs.
```
jmhbnz, line 909: Just out of curiosity is there a particular reason for `3`, or was it simply a reasonable duration?

ahrtr: It should be a reasonable value derived from experience, e.g. Keepalive interval & Keepalive retry. There are also some route protocol use similar retry value.

jmhbnz (approved):
LGTM - Nice work @ahrtr

ivanvc (approved):
LGTM. Thanks, Benjamin.

ahrtr:
Thanks all for the review.

server/etcdserver/server.go
serathius, line 909: Why 3 election ticks? I think the correct value is somewhere between `TickMs` (default 100ms) and `ElectionMs` (default 1s), meaning the 3 makes sense for default config, but I would argue it can be problematic for cases where `ElectionMS` < 3* `TicksMs`

server/etcdserver/server.go
```diff
@@ -904,10 +904,26 @@ func (s *EtcdServer) revokeExpiredLeases(leases []*lease.Lease) {
...
+	latestTickTs := s.r.getLatestTickTs()
+	threshold := 3 * time.Duration(s.Cfg.TickMs) * time.Millisecond
+	return latestTickTs.Add(threshold).After(time.Now())
+}
+
 // ensureLeadership checks whether current member is still the leader.
 func (s *EtcdServer) ensureLeadership() bool {
 	lg := s.Logger()
 
+	if s.isActive() {
```
serathius, line 920: This is incorrect assumption for 5 node clusters, getting a tick from 1 member is not effective way to confirm no other leader was elected. Imagine 3/2 network split with leader on side of 2. With your change the leader can continue think it's active by receiving ticks from 1 member, where on the other side of network 3 members have already elected new leader.

serathius:
The proposed solution is not compatible with etcd configuration options and doesn't work for 5 node clusters. Please consider reverting and redesign.

server/etcdserver/server.go
ahrtr: > Why 3 election ticks?

Read https://github.com/etcd-io/etcd/pull/18428#discussion_r1712624482.

> I would argue it can be problematic for cases where `ElectionMS` < 3* `TicksMs`

I am not worry about this.
- checking activity should be only related to `tickMs`.
- also if `ElectionMS` < 3* `TicksMs`, it means it's a very inappropriate election value.

server/etcdserver/server.go
ahrtr: It has nothing to do with how many nodes the cluster has. The intention is to guard the case the node being stuck by itself, e.g stall write.



> Imagine 3/2 network split with leader on side of 2. With your change the leader can continue think it's active by receiving ticks from 1 member

It's active, but it won't be a leader any more.  It will step down automatically in such case due to losing quorum.

ahrtr:
> The proposed solution is not compatible with etcd configuration options and doesn't work for 5 node clusters. Please consider reverting and redesign.

Don't agree. The only minor possible improvement is to guard the threshold < electionTimeout, but I am not worry about it as mentioned in https://github.com/etcd-io/etcd/pull/18428#discussion_r1718609232

server/etcdserver/server.go
ahrtr: Note that the raft protocol handle the network partition case perfectly.

server/etcdserver/server.go
serathius: > also if ElectionMS < 3* TicksMs, it means it's a very inappropriate election value.

Then let's codify that is is inappropriate. Either working or failing validation.

server/etcdserver/server.go
serathius: > It's active, but it won't be a leader any more.

We are talking about 1 second before network partition, until heath probes fail for ElectionMs, the old leader will still think it's a leader before they resign.

PR #18442, opened 2024-08-14
Reduce number of concurrent clients to 8 and compactions to avoid flakes

serathius:
Not hitting minimal QPS is expected to be caused by introduction of
compation. Let's avoid it for high throughput test cases.

Reducing number of clients to avoid linearization timeouts.

cc @ah8ad3 @fuweid @MadhavJivrajani @siyuanfoundation


ah8ad3 (commented):
/lgtm
Thanks Marek.

ah8ad3:
Another hit 
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-etcd-robustness-amd64/1824780028328546304

PR #18457, opened 2024-08-17
[3.5] tests/e2e: add e2e test to reproduce issue 18089

fuweid:
The goal is to reproduce a DELETE event being dropped in a watch after a compaction occurs on the revision where the deletion took place. In order to reproduce this, we perform the following sequence (steps for reproduction thanks to @ahrtr):
  - PUT k v2 (assume returned revision = r2)
  - PUT k v3 (assume returned revision = r3)
  - PUT k v4 (assume returned revision = r4)
  - DELETE k (assume returned revision = r5)
  - PUT k v6 (assume returned revision = r6)
  - COMPACT r5
  - WATCH rev=r5

We should get the DELETE event (r5) followed by the PUT event (r6). However, currently we only get the PUT event with returned revision of r6 (key=k, val=v6).

(cherry picked from commit ebf2cac6bda38892121585d2467982da54e60ef1)

Part of https://github.com/etcd-io/etcd/pull/18274

### DIFF
--- a/tests/e2e/watch_test.go
+++ b/tests/e2e/watch_test.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
+	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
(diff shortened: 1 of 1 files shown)


ahrtr (approved):
LGTM

serathius (approved):

Issue #17098, opened 2023-12-11
Reduce memory usage of  etcd member catchup mechanism


serathius:
### What would you like to be added?

All requests made to etcd are serialized into raft entry proto and persisted on disk WAL. That's good, but to allow slow/disconnected members to catchup etcd also stores last 10`000 entries in raft.InMemoryStorage, all loaded into memory. In some cases this can cause huge memory bloat of etcd. Imagine you have a sequence of large put requests (for example 1MB configmaps in Kubernetes). etcd will keep all 10GB in memory, doing nothing. 

This can be reproduced by running `./bin/tools/benchmark put --total=1000 --val-size=1000000` and collecting inuse_space heap profile.

![Selection_012](https://github.com/etcd-io/etcd/assets/5873433/99ac472c-0c59-4292-a3c5-ae385c30701f)

The mechanism is really dump and could benefit from following improvements:
* Compact raft.InMemoryStorage after every apply instead of once every snapshot (10`000 entries is default snapshot frequency). With removal of v2storage we can switch to using applied index instead of snapshot index. As apply index is updated more frequently we can execute Compact more frequently, possibly after every apply assuming that it's not too costly,
* Compact raft.InMemoryStorage based state of the slowest member. Why keep 5`000 entries (default catchup entries) in 1 node cluster? or there all members are up to date? We could read the state of the slowest member and Compact based on that
* Tune the default snapshot catchup entries (5`000 entries). Current is based on 1ms latency and 10k throughput https://github.com/etcd-io/etcd/pull/2403/files. Would be good to revisit this and tune it. For example compare catchup times and availability for different sizes of WAL entries, DB file, latency, network throughput etc.
* Change the semantic of catchup-entries from "we always store at least X entries" to "we store entries only for members that are behind by X entries max". If member is behind more then X entries, as it no longer makes sense to use raft entries to sync it. In this case we will use snapshot. So why keep those entries?

### Why is this needed?

Prevent etcd memory bloating and make memory usage more predictable.

serathius:
cc @ahrtr

moficodes:
I am interested in helping. But I am not sure I know exactly what needs to be done. Could I shadow someone or get some guidance if I were to attempt this? @serathius

serathius:
There are 4 different changes proposed. Looking at the first one `Compact raft.InMemoryStorage after every apply instead of once every snapshot.`.  High level what we would want to do is:
[...]

PR #18459, opened 2024-08-17
etcdserver: separate raft log compact from snapshot 

clement2026:
Part of  #17098

## Key Changes

* Added a new function `compactRaftLog` for raft log compaction.
* Introduce 2 variables: `RaftLogCompactionStep` and `DefaultRaftLogCompactionStep` to manage how often the raft log is compacted.
* Refactor some existing tests. These tests previously expected a compaction immediately after a snapshot to ensure a snapshot was sent to followers. This PR broke those tests, so I modified them to ensure they still work.
* Add unit and integration tests for this PR.

- [ ] Blocked by #18494

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -58,6 +58,11 @@ type ServerConfig struct {
 	// follower to catch up.
 	SnapshotCatchUpEntries uint64
 
+	// RaftLogCompactionStep ensures raft log is compacted whenever
+	// applied index is a multiple of RaftLogCompactionStep.
+	// The minimum value is 1, meaning compacting raft log each time applied index increases.
+	RaftLogCompactionStep uint64
+
 	MaxSnapFiles uint
 	MaxWALFiles  uint
 
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -86,7 +86,7 @@ const (
 	// DefaultRaftLogCompactionStep ensures raft log is compacted whenever
 	// applied index is a multiple of DefaultRaftLogCompactionStep.
 	// The minimum value is 1, meaning compacting raft log each time applied index increases.
-	DefaultRaftLogCompactionStep uint64 = 100
+	DefaultRaftLogCompactionStep uint64 = 10
 
 	StoreClusterPrefix = "/0"
 	StoreKeysPrefix    = "/1"
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -83,6 +83,11 @@ const (
 	// follower to catch up.
 	DefaultSnapshotCatchUpEntries uint64 = 5000
 
+	// DefaultRaftLogCompactionStep ensures raft log is compacted whenever
+	// applied index is a multiple of DefaultRaftLogCompactionStep.
+	// The minimum value is 1, meaning compacting raft log each time applied index increases.
+	DefaultRaftLogCompactionStep uint64 = 100
+
 	StoreClusterPrefix = "/0"
 	StoreKeysPrefix    = "/1"
 
@@ -568,6 +573,14 @@ func (s *EtcdServer) start() {
 		)
 		s.Cfg.SnapshotCatchUpEntries = DefaultSnapshotCatchUpEntries
 	}
+	if s.Cfg.RaftLogCompactionStep == 0 {
+		lg.Info(
+			"updating raft-log-compaction-step to default",
+			zap.Uint64("given-raft-log-compaction-step", s.Cfg.RaftLogCompactionStep),
+			zap.Uint64("updated-raft-log-compaction-step", DefaultRaftLogCompactionStep),
+		)
+		s.Cfg.RaftLogCompactionStep = DefaultRaftLogCompactionStep
+	}
 
 	s.w = wait.New()
 	s.applyWait = wait.NewTimeList()
@@ -979,6 +992,7 @@ func (s *EtcdServer) applyAll(ep *etcdProgress, apply *toApply) {
 	<-apply.notifyc
 
 	s.triggerSnapshot(ep)
+	s.compactRaftLog(ep.appliedi)
 	select {
 	// snapshot requested via send()
 	case m := <-s.r.msgSnapC:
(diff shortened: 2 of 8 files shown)


clement2026:
# Benchmark Summary

Here are the benchmark results for compacting the raft log every time `applied index` increases by 1, 10, 100, and 1000 (let's call this "step").

| Branch            |       | Throughput       |
|-------------------|-------|------------------|
| main(as the base) | read  | -                |
|                   | write | -                |
| step_1            | read  | [2.86%, 17.00%]  |
|                   | write | [3.02%, 16.75%]  |
| step_10           | read  | [2.82%, 16.53%]  |
|                   | write | [2.98%, 16.95%]  |
| step_100          | read  | [-2.09%, 9.77%]  |
|                   | write | [-1.00%, 10.93%] |
| step_1000         | read  | [-4.13%, 7.27%]  |
|                   | write | [-4.91%, 7.65%]  |
| main again        | read  | [-4.19%, 7.39%]  |
|                   | write | [-4.41%, 9.03%]  |

Throughput performs best when the step is 1 or 10. There isn’t much difference between the two. I've set the default step value to 10 in the code.

clement2026:
## Benchmark Details

Here's how I ran the benchmark:
1. Run benchmark on `main`, `step_1`, `step_10`, `step_100`, `step_1000` and `main` branch in that order. 
2. Reboot the OS between each run.
3. Collect .csv files and generate the heat maps.

### step_1 vs main
Compact raft log every time `applied index` increases by 1.

step_1_read

### step_10  vs main
Compact raft log every time `applied index` increases by 10.

step_10_read

### step_100  vs main
Compact raft log every time `applied index` increases by 100.

step_100_read

### step_1000  vs main
Compact raft log every time `applied index` increases by 1000.

step_1000_read

### main vs main
Re-run benchmarks on the main branch to ensure results are consistent

main_2_read

### step_1 vs step_10
Since the benchmark results for step_1 and step_10 are pretty close, here I compare them to see the difference:

step_1-vs-step_10_read


---

GitHub workflow: https://github.com/clement2026/etcd-benchmark-action/actions/runs/10424407957

The archive containing .csv files and the script: [etcd-benchmark-20240817-06-00-29.zip](https://github.com/user-attachments/files/16644019/etcd-benchmark-20240817-06-00-29.zip)

The benchmark was run on a cloud VM with:
* 16GB RAM
* 8 vCPUs
* 150GB NVMe SSD
* Ubuntu 24.04 LTS x64

clement2026:
# Reason for the failure
The robustness test failed at `panic("need non-empty snapshot")` in the raft library. 

This happens before the leader creates any snapshot, and `raftStorage.ents` has already been compacted and lacks the entries a slow follower needs. So, the slow follower requests a snapshot instead, causing the leader to panic.

https://github.com/etcd-io/raft/blob/ba8d7e340d530ae106118f059d3da4d89f933d23/raft.go#L662C1-L689
```go
// maybeSendSnapshot fetches a snapshot from Storage, and sends it to the given
// node. Returns true iff the snapshot message has been emitted successfully.
func (r *raft) maybeSendSnapshot(to uint64, pr *tracker.Progress) bool {
	if !pr.RecentActive {
		r.logger.Debugf("ignore sending snapshot to %x since it is not recently active", to)
		return false
	}


	snapshot, err := r.raftLog.snapshot()
	if err != nil {
		if err == ErrSnapshotTemporarilyUnavailable {
			r.logger.Debugf("%x failed to send snapshot to %x because snapshot is temporarily unavailable", r.id, to)
			return false
		}
		panic(err) // TODO(bdarnell)
	}
	if IsEmptySnap(snapshot) {
		panic("need non-empty snapshot")
	}
	sindex, sterm := snapshot.Metadata.Index, snapshot.Metadata.Term
	r.logger.Debugf("%x [firstindex: %d, commit: %d] sent snapshot[index: %d, term: %d] to %x [%s]",
		r.id, r.raftLog.firstIndex(), r.raftLog.committed, sindex, sterm, to, pr)
	pr.BecomeSnapshot(sindex)
	r.logger.Debugf("%x paused sending replication messages to %x [%s]", r.id, to, pr)


	r.send(pb.Message{To: to, Type: pb.MsgSnap, Snapshot: &snapshot})
	return true
}
```

# Visualization
Here are some images to help us understand the situation better.

### Main branch
Compaction only happens right after a snapshot is created. 

Untitled-2024-08-16-18062

### This PR
Compaction can happen before a snapshot is created. 

Untitled-2024-08-16-1806 3

# Solutions
* Modify the implementation of `func (r *raft) maybeSendSnapshot(to uint64, pr *tracker.Progress) bool` to send an empty snapshot instead of causing a panic. However, I’m not sure about the expected behavior or potential risks.
* Ensure a snapshot is created if it doesn’t exist when etcd starts up. This might sound silly, I know 😂

I'm stumped right now. Any ideas would be awesome!

serathius:
> Ensure a snapshot is created if it doesn’t exist when etcd starts up. This might sound silly, I know

Sounds reasonable. We are separating mechanisms for raft log compaction and snapshots, meaning that now we need to guarantee that snapshot is always available. Creating empty snapshot at the etcd start sounds like a good first iteration.

clement2026:
> Sounds reasonable. We are separating mechanisms for raft log compaction and snapshots, meaning that now we need to guarantee that snapshot is always available. Creating empty snapshot at the etcd start sounds like a good first iteration.

Thank you for the info! I'll give this a shot and open a new PR soon.

ahrtr:
> Compaction can happen before a snapshot is created.
> 
> Untitled-2024-08-16-1806 3

leader sending a snapshot means that the leader sends the whole bbolt db snapshot. It doesn't care about what's the current applied index the follower is.

clement2026:
@ahrtr the image isn't loading; It's not set to public: https://private-user-images.githubusercontent.com/27894831/358932866-df2dbdb7-2165-492c-93b8-907e09651fd4.png


> leader sending a snapshot means that the leader sends the whole bbolt db snapshot. It doesn't care about what's the current applied index the follower is.

Though the leader doesn't care about the follower, the leader itself can panic before sending out the snapshot.

Before sending a snapshot to a follower,  the leader runs `maybeSendSnapshot()` and may panic here:

https://github.com/etcd-io/raft/blob/ba8d7e340d530ae106118f059d3da4d89f933d23/raft.go#L678-L680

```go
	if IsEmptySnap(snapshot) {
		panic("need non-empty snapshot")
	}
```

All the leader wants is actually a bbolt db snapshot, not a raft log snapshot. But it seems I can't walk around `maybeSendSnapshot()` as it's part of the raft mechanism.

ahrtr:
> Though the leader doesn't care about the follower, the leader itself can panic before sending out the snapshot.
> 
> Before sending a snapshot to a follower, the leader runs `maybeSendSnapshot()` and may panic here:
> 
> https://github.com/etcd-io/raft/blob/ba8d7e340d530ae106118f059d3da4d89f933d23/raft.go#L678-L680
> 
> ```go
> 	if IsEmptySnap(snapshot) {
> 		panic("need non-empty snapshot")
> 	}
> ```

It's concerning we mix a potential bug with enhancement. Can you raise a separate issue to clearly clarify how to reproduce the issue?

clement2026:
> It's concerning we mix a potential bug with enhancement. Can you raise a separate issue to clearly clarify how to reproduce the issue?

Sure, I'll create the issue a bit later.

clement2026:
After chatting with serathius, I've realized this PR isn’t needed anymore. I'll keep working on https://github.com/etcd-io/etcd/issues/17098 in a new PR.
Closing this.

PR #18462, opened 2024-08-19
update broken link in mixin readme

Lusitaniae:
old link doesn't exist

according to https://github.com/monitoring-mixins/docs/blob/master/README.md the pdf is related to the google doc file


spzala (commented):
@Lusitaniae thanks for the PR but please sign your commit to pass the DCO for further review.

Issue #18471, opened 2024-08-19
Revision auto compaction should have configurable timer instead of fixed 5 minute timer


jrcichra:
### What would you like to be added?

Today the revision auto compaction has a [constant 5 minute compaction](https://github.com/etcd-io/etcd/blob/f402c2ae2b32c450bbc2cb2a3610bcd1c0ece1b6/server/etcdserver/api/v3compactor/revision.go#L61). Change this to be a command-line / environment variable configurable option.

### Why is this needed?

With enough key revisions per second, the constant 5 minute interval on revision compaction isn't fast enough to keep up with growth. Consider making the interval configurable, which will override the default behavior of 5 minutes.

parthlaw:
I would like to work on this issue.

jrcichra:
> I would like to work on this issue.

Hey @parthlaw , I have an implementation here I made along with the issue: https://github.com/etcd-io/etcd/pull/18472. Let me know what you think!

parthlaw:
Thanks for the update @jrcichra . I am looking into the implementation. This is a great learning opportunity for me—Thanks for the guidance!

ivanvc:
Discussed during the fortnightly triage meeting. @moficodes to review.

PR #18472, opened 2024-08-19
etcdserver: add auto compaction interval option

jrcichra:
With enough key revisions per second, the constant 5 minute interval on revision compaction isn't fast enough to keep up with growth. Make the interval configurable, which will override the default behavior.

Also pass the configuration through to periodic compaction for finer control of when compaction happens based on time.

Fixes:
https://github.com/etcd-io/etcd/issues/18471


jrcichra:
My implementation here tries to support both `periodic` and `revision` but my original focus was solving the problem for just `revision`. It may not make as much sense for `periodic`.

jrcichra:
I'm passing all the tests now. @jmhbnz how should I proceed? I see there's a codecov report but there's a large section of code I didn't touch regarding failed auto compaction.

jrcichra:
Rebase is complete 🎉

jrcichra:
Hey @moficodes have you had a chance to look at this change?

jrcichra:
Hey @jmhbnz / @moficodes I've rebased and all the tests are passing. Could I get a review please? Thanks!

jmhbnz (commented):
Hi @jrcichra - Thanks for your effort working on this feature, it seems reasonable to allow the existing constant top be configurable. To speed up getting this accepted I think it would be best to include an e2e test.  There are some compaction related e2e tests in `tests/e2e/watch_test.go` which may provide some inspiration on potential approaches.

PR #18494, opened 2024-08-25
etcdserver: creates a non-empty raft log snapshot on server startup

clement2026:
#18459 requires that a non-empty raft log snapshot is always available. This PR creates a non-empty raft log snapshot on server startup.

## Key changes

* Alter `shouldSnapshot` function
* Added integration tests in `tests/integration/raft_log_snapshot_test.go`
* Fixed failing tests

**Blocked by**
- [ ] #18582 
- [ ] #18583 
- [ ] #18584

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1212,7 +1212,7 @@ func (s *EtcdServer) triggerSnapshot(ep *etcdProgress) {
 func (s *EtcdServer) shouldSnapshot(ep *etcdProgress) bool {
 	return (s.forceSnapshot && ep.appliedi != ep.snapi) ||
 		(ep.appliedi-ep.snapi > s.Cfg.SnapshotCount) ||
-		// ensures a non-empty snapshot always exists
+		// make sure a non-empty snapshot always exists
 		(ep.snapi == 0 && ep.appliedi > ep.snapi)
 }
 
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1210,7 +1210,10 @@ func (s *EtcdServer) triggerSnapshot(ep *etcdProgress) {
 }
 
 func (s *EtcdServer) shouldSnapshot(ep *etcdProgress) bool {
-	return (s.forceSnapshot && ep.appliedi != ep.snapi) || (ep.appliedi-ep.snapi > s.Cfg.SnapshotCount)
+	return (s.forceSnapshot && ep.appliedi != ep.snapi) ||
+		(ep.appliedi-ep.snapi > s.Cfg.SnapshotCount) ||
+		// ensures a non-empty snapshot always exists
+		(ep.snapi == 0 && ep.appliedi > ep.snapi)
 }
 
 func (s *EtcdServer) hasMultipleVotingMembers() bool {
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -2184,24 +2184,24 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
 	}
 
 	// keep some in memory log entries for slow followers.
-	compacti := uint64(1)
 	if snapi > s.Cfg.SnapshotCatchUpEntries {
-		compacti = snapi - s.Cfg.SnapshotCatchUpEntries
-	}
-
-	err = s.r.raftStorage.Compact(compacti)
-	if err != nil {
-		// the compaction was done asynchronously with the progress of raft.
-		// raft log might already been compact.
-		if err == raft.ErrCompacted {
-			return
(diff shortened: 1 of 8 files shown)


server/etcdserver/server.go
```diff
@@ -2181,24 +2184,24 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
 	}
 
 	// keep some in memory log entries for slow followers.
-	compacti := uint64(1)
```
clement2026, line 2184: This could cause a panic if the applied index is also 1

server/etcdserver/server_test.go
```diff
@@ -99,8 +99,12 @@ func TestApplyRepeat(t *testing.T) {
 		SyncTicker:   &time.Ticker{},
 		consistIndex: cindex.NewFakeConsistentIndex(0),
 		uberApply:    uberApplierMock{},
+		kv:           mvcc.New(zaptest.NewLogger(t), be, &lease.FakeLessor{}, mvcc.StoreConfig{}),
```
clement2026, line 102: Set `kv` to avoid a nil pointer panic:

https://github.com/etcd-io/etcd/blob/4bb9392289623e8bf05916281bd696fbbe42fd4d/server/etcdserver/server.go#L2132-L2149

server/etcdserver/server_test.go
```diff
@@ -99,8 +99,12 @@ func TestApplyRepeat(t *testing.T) {
...
 	}
 	s.start()
+
+	n.readyc <- newDummyPutReqReady()
+
```
clement2026, line 107: The test case wasn’t appending raft log entries correctly when the applied index increases, leading to an “slice bounds out of range” panic during snapshot creation.

tests/framework/integration/cluster.go
```diff
@@ -1693,5 +1693,30 @@ func (c *Cluster) MustNewMember(t testutil.TB, resp *clientv3.MemberAddResponse)
...
+		default:
+			err = fmt.Errorf("unsupported scheme: %s", url.Scheme)
+		}
+
+		if err != nil {
+			t.Fatal("failed to listen on %v: %v", url, err)
+		}
+		listeners = append(listeners, l)
+	}
+	m.PeerListeners = listeners
```
clement2026, line 1719: The test case added a learner member with peer URL `http://127.0.0.1:1234` to the cluster, but the learner member didn't listen to the host and couldn’t receive a snapshot from the leader.

clement2026:
Hey @serathius, it would be great if you could check this out when you have a moment. Thanks!

server/etcdserver/server_test.go
serathius, line 102: I see that we might need to fix couple of things in tests before we can merge the logic change. How about creating a separate PR to fix the tests first? That would give us confidence that there is no interdependence.

server/etcdserver/server_test.go
serathius: Can you move it to a separate PR as proposed above?

tests/e2e/v2store_deprecation_test.go
```diff
@@ -138,7 +139,32 @@ func TestV2DeprecationSnapshotMatches(t *testing.T) {
 	members2 := addAndRemoveKeysAndMembers(ctx, t, cc2, snapshotCount)
 	assert.NoError(t, epc.Close())
 
-	assertSnapshotsMatch(t, oldMemberDataDir, newMemberDataDir, func(data []byte) []byte {
+	lastVer, err := e2e.GetVersionFromBinary(e2e.BinPath.EtcdLastRelease)
```
serathius, line 142: Not sure why this is needed, please all test changes to separate PR.

tests/framework/integration/cluster.go
serathius: Please move to a separate PR so we can merge this faster.

server/etcdserver/server_test.go
clement2026: It’s good practice to avoid interdependence. Thanks for the insight! I’ll create new PRs to fix the tests whenever they can be separated.

ahrtr:
> #18459 requires that a non-empty raft log snapshot is always available.

Can you explain this? Why a non-empty snapshot is always required? We don't have such restriction before.

clement2026:
> Can you explain this? Why a non-empty snapshot is always required? We don't have such restriction before.

@ahrtr sure, check out this comment for more details https://github.com/etcd-io/etcd/pull/18459#issuecomment-2295388421

clement2026:
After chatting with serathius, I've realized this PR isn’t needed anymore. I'll keep working on https://github.com/etcd-io/etcd/issues/17098 in a new PR.
Closing this.

Issue #18370, opened 2024-07-26
Enable `unexported-return` linter rule


ivanvc:
### What would you like to be added?

There are two remaining linter rules that we haven't enabled but are left as TODO (`./tools/.golangci.yaml`): `exported` and `unexported-return`. The former will have significant changes to exported functions and interfaces. The latter, even though it impacts exported structs, it's for the better (it exposes them without breaking client implementations).

I suggest breaking the pull requests into small tasks like we did on #17578.

The following modules have warnings to be addressed:

- [x] `api` - #18645
- [ ] `pkg`
- [ ] `client/pkg`
- [ ] `client/v3`
- [ ] `server` - #19052 
- [x] `tests`
  - #19105 

Refer to the [attached log](https://github.com/user-attachments/files/16395433/unexported-return.log) file with the result from `KEEP_GOING_MODULE=true make verify-lint`.

### Why is this needed?

To improve the quality of the code.

thedtripp:
I'd like to work on this issue.

jmhbnz:
> @jmhbnz, could this be a bug? @thedtripp is already part of our org 🤔
> 
> Edit: I think it's because he doesn't belong to any teams. Should he open a pull request to add himself to the members? (https://github.com/kubernetes/org/blob/main/config/etcd-io/sig-etcd/teams.yaml#L77)

Good spotting - Yes this should be tidied up, @thedtripp please feel free to raise a `kubernetes/org` pr to add yourself to the `members` team  🙏🏻

thedtripp:
> > @jmhbnz, could this be a bug? @thedtripp is already part of our org 🤔
> > Edit: I think it's because he doesn't belong to any teams. Should he open a pull request to add himself to the members? (https://github.com/kubernetes/org/blob/main/config/etcd-io/sig-etcd/teams.yaml#L77)
> 
> Good spotting - Yes this should be tidied up, @thedtripp please feel free to raise a `kubernetes/org` pr to add yourself to the `members` team 🙏🏻

I'll raise that PR in the next few days. Thanks!

jmhbnz:
Discussed during sig-etcd triage meeting. There are a number of modules that need updating, so we could break this up into smaller chunks and do it one module at a team.

@thedtripp let us know which module you will focus on and we can assign out the others.

ivanvc:
I'll do the first one (`api`) to set an example of what should be done with this issue.

thedtripp:
[...]

PR #18539, opened 2024-09-04
Support scan-keys command to scan the key space starting a specific revision

ahrtr:
The `etcd-dump-db scan-keys ...` is useful to analyze issue something like https://github.com/etcd-io/bbolt/issues/778.  It supports scanning the key space even the db is corrupted.

### DIFF
--- a/tools/etcd-dump-db/main.go
+++ b/tools/etcd-dump-db/main.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -40,6 +41,11 @@ var (
 		Short: "iterate-bucket lists key-value pairs in reverse order.",
 		Run:   iterateBucketCommandFunc,
 	}
+	scanKeySpaceCommand = &cobra.Command{
+		Use:   "scan-keys [data dir or db file path] [start revision]",
+		Short: "scan-keys scans all the key-value pairs starting from a specific revision in the key space.",
+		Run:   scanKeysCommandFunc,
+	}
 	getHashCommand = &cobra.Command{
 		Use:   "hash [data dir or db file path]",
 		Short: "hash computes the hash of db file.",
@@ -58,6 +64,7 @@ func init() {
 
 	rootCommand.AddCommand(listBucketCommand)
 	rootCommand.AddCommand(iterateBucketCommand)
+	rootCommand.AddCommand(scanKeySpaceCommand)
 	rootCommand.AddCommand(getHashCommand)
 }
 
@@ -102,6 +109,27 @@ func iterateBucketCommandFunc(_ *cobra.Command, args []string) {
 	}
 	bucket := args[1]
 	err := iterateBucket(dp, bucket, iterateBucketLimit, iterateBucketDecode)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func scanKeysCommandFunc(_ *cobra.Command, args []string) {
+	if len(args) != 2 {
+		log.Fatalf("Must provide 2 arguments (got %v)", args)
+	}
+	dp := args[0]
+	if !strings.HasSuffix(dp, "db") {
+		dp = filepath.Join(snapDir(dp), "db")
+	}
+	if !existFileOrDir(dp) {
+		log.Fatalf("%q does not exist", dp)
+	}
+	startRev, err := strconv.ParseInt(args[1], 10, 64)
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = scanKeys(dp, startRev)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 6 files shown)


tools/etcd-dump-db/main.go
```diff
@@ -107,6 +114,27 @@ func iterateBucketCommandFunc(_ *cobra.Command, args []string) {
 	}
 }
 
+func scanKeysCommandFunc(_ *cobra.Command, args []string) {
+	if len(args) != 2 {
+		log.Fatalf("Must provide 2 arguments (got %v)", args)
+	}
```
ivanvc, line 120: Suggestion to use the `Args` field of `cobra.Command` (in line 44)

```
Args:  cobra.ExactArgs(2),
```

However, I see that other command functions have the validation defined this way.

tools/etcd-dump-db/page.go
```diff
@@ -0,0 +1,68 @@
...
+func (p *page) isLeafPage() bool {
+	return p.flags == leafPageFlag
+}
+
+func loadPage(buf []byte) *page {
+	return (*page)(unsafe.Pointer(&buf[0]))
+}
+
+// LeafPageElement retrieves the leaf node by index
+func (p *page) LeafPageElement(index uint16) *leafPageElement {
```
ivanvc, line 43: Suggestion to export the `leafPageElement` struct. Ref: #18370 / golang/lint#210

tools/etcd-dump-db/main.go
```diff
@@ -40,6 +41,11 @@ var (
 		Short: "iterate-bucket lists key-value pairs in reverse order.",
 		Run:   iterateBucketCommandFunc,
 	}
+	scanKeySpaceCommand = &cobra.Command{
+		Use:   "scan-keys [data dir or db file path] [start revision]",
+		Short: "scan-keys scans all the key-value pairs starting from a specific revision in the key space.",
```
jmhbnz, line 46: Suggest we also update documentation to explain the new command in https://github.com/etcd-io/etcd/blob/main/tools/etcd-dump-db/README.md

tools/etcd-dump-db/meta.go
```diff
@@ -0,0 +1,40 @@
...
+import "unsafe"
+
+const magic uint32 = 0xED0CDAED
+
+type inBucket struct {
+	root     uint64 // page id of the bucket's root-level page
+	sequence uint64 // monotonically incrementing, used by NextSequence()
+}
+
+type Meta struct {
```
jmhbnz, line 26: Just an idea but in future could we pull some of these values from `bbolt` rather than duplicating them here? i.e. https://github.com/etcd-io/bbolt/blob/0cecda66e016f212c9d13aa74cfd93cb07948ff6/internal/common/meta.go#L12

Also https://github.com/etcd-io/bbolt/blob/0cecda66e016f212c9d13aa74cfd93cb07948ff6/internal/common/types.go#L16

tools/etcd-dump-db/scan.go
```diff
@@ -0,0 +1,150 @@
...
+	"io"
+	"os"
+
+	"go.etcd.io/etcd/server/v3/storage/mvcc"
+)
+
+func scanKeys(dbPath string, startRev int64) error {
+	pgSize, hwm, err := readPageAndHWMSize(dbPath)
+	if err != nil {
+		return fmt.Errorf("failed to read page and HWM size: %w", err)
```
jmhbnz, line 28: Nitpick: Is it intentional that some error messages start lowercase and some start uppercase?  Would be nice if we could be consistent where possible.

tools/etcd-dump-db/page.go
ahrtr: Forgot to change `LeafPageElement` to `leafPageElement`, done & thx

tools/etcd-dump-db/scan.go
ahrtr: Yes, it's intentional, please read https://go.dev/wiki/CodeReviewComments#error-strings

tools/etcd-dump-db/meta.go
ahrtr: I thought about it, but I intentionally did not do it because,
- to avoid circuarly dependency
- the Meta/Page are included in internal package, so they can't be used by other repo

tools/etcd-dump-db/main.go
ahrtr: Good catch, will update later.

tools/etcd-dump-db/main.go
ahrtr: Yes, I intentionally kept it consistent with other commands. Please feel free to update them separately. Thanks.

tools/etcd-dump-db/main.go
ahrtr: Done.

tools/etcd-dump-db/scan.go
jmhbnz: Thanks for clarification!

jmhbnz (approved):
LGTM - Thanks @ahrtr

ivanvc (approved):
LGTM. Thanks, Benjamin.

PR #18546, opened 2024-09-05
Optimize for-loop iteration: exit early when httpOnly is true

cfanbo:
Optimize for-loop iteration: exit early when httpOnly is true

```
	for _, sctx := range e.sctxs {
		if sctx.httpOnly {
			splitHTTP = true
			break
		}
	}
```


cfanbo:
Why hasn't it been approved？

PR #18557, opened 2024-09-09
use uint64 for priority variable

piggito:
Changing priority variable from uint16 to uint64 allows for a wider range of values. For instance, we have a use case where we need unix timestamp in nanoseconds granularity which is not supported by uint16 but requires uint64.

### DIFF
--- a/client/v3/experimental/recipes/priority_queue.go
+++ b/client/v3/experimental/recipes/priority_queue.go
@@ -41,6 +41,13 @@ func (q *PriorityQueue) Enqueue(val string, pr uint16) error {
 	return err
 }
 
+// EnqueueUint64 puts a value into a queue with a given uint64 priority.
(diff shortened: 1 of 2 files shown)


ivanvc:
Hi @piggito, thanks for your pull request. Unfortunately, this change would break the public API. I think you will have better odds at merging if you add a new function, `EnqueueUint64()` instead (along with its tests).

mickymiek:
Hi, is there something blocking this from getting merged?

jmhbnz:
> Hi, is there something blocking this from getting merged?

@piggito can you please rebase this pr and then review test output, thanks 🙏🏻

piggito:
@jmhbnz I rebased and fixed tests, please can you review.

ivanvc (approved):
LGTM. Thanks, @piggito.

/cc @jmhbnz @ahrtr @serathius @fuweid

PR #18582, opened 2024-09-13
tests: fix server_test.go

clement2026:
#18494 creates a non-empty raft log snapshot on server startup, which requires:

1. There are  available entries in raft log.
2. A non-nil `kv` is set :
https://github.com/etcd-io/etcd/blob/981061a4958e8d386b025dbecd43af9b974709e2/server/etcdserver/server.go#L2132-L2143

---

This PR fixes the failed tests in `server_test.go` by :
1. Appending raft log entries by sending a put request to raft.Node's readyc
2. Setting a non-nil `kv` for `EtcdServer`

---

### DIFF
--- a/server/etcdserver/server_test.go
+++ b/server/etcdserver/server_test.go
@@ -1598,11 +1598,14 @@ func TestIsActive(t *testing.T) {
 	}
 }
 
-// newDummyPutReqReady creates a dummy raft.Ready.
+// newDummyPutReqReady creates a raft.Ready.
 //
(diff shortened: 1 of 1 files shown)


serathius (approved):

serathius:
cc @ahrtr

ahrtr:
The change is simple and straightforward, but let's get https://github.com/etcd-io/etcd/pull/18494#issuecomment-2348945021 clarified firstly.

PR #18575, opened 2024-09-11
[2024-09-11] Bump dependency updates identified by dependabot

henrybear327:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

This pull request completes this week's etcd dependency updates following our [dependency roster](https://docs.google.com/spreadsheets/d/1jodHIO7Dk2VWTs1IRnfMFaRktS9IH8XRyifOnPdSY8I/edit#gid=1394774387) and [dependency management instructions](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md).

# Summary of actions

## Bumped
- https://github.com/etcd-io/etcd/pull/18566
- https://github.com/etcd-io/etcd/pull/18565
- https://github.com/etcd-io/etcd/pull/18562 + https://github.com/etcd-io/etcd/pull/18567
- https://github.com/etcd-io/etcd/pull/18564 + https://github.com/etcd-io/etcd/pull/18559
- https://github.com/etcd-io/etcd/pull/18558 + https://github.com/etcd-io/etcd/pull/18563
## Fully indirect
- https://github.com/etcd-io/etcd/pull/18560
- https://github.com/etcd-io/etcd/pull/18561

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #18590, opened 2024-09-14
Add tools/mod to module_dirs

henrybear327:
As `tools/mod` also contains the `go.mod` file, we should look into adding it to the `module_dirs` variable, so that when executing `./scripts/fix.sh`, the proper checks and fixes can be applied.

To address the issue of broken unit tests and code coverage due to `tools/mod` directory's lack of Go code, we've introduced a new `doc.go` file. This file acts as a placeholder, enabling tools like `golangci-lint` and `go test` to function correctly.

Discovered when working on https://github.com/etcd-io/etcd/pull/18575, where running `./scripts/fix.sh` missed fixing the go mod file in `tools/mod`, causing the CI pipeline to fail.

### DIFF
--- a/tools/mod/doc.go
+++ b/tools/mod/doc.go
@@ -1,4 +1,4 @@
-// Copyright 2016 The etcd Authors
+// Copyright 2024 The etcd Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,14 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build libs
+// As this directory implements the pattern for tracking tool dependencies as documented here:
+// https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module, it doesn't
+// contain any valid go source code in the directory directly. This would break scripts for
+// unit testing, golangci-lint, and coverage calculation.
+//
+// Thus, to ensure tools to run normally, we've added this empty file.
 
-// This file implements that pattern:
-// https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
-// for etcd. Thanks to this file 'go mod tidy' does not removes dependencies.
-
-package libs
-
-import (
-	_ "github.com/gogo/protobuf/proto"
-)
+package mod
--- a/scripts/test_lib.sh
+++ b/scripts/test_lib.sh
@@ -166,7 +166,7 @@ function run_for_module {
 }
 
 function module_dirs() {
-  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/rw-heatmaps tools/testgrid-analysis ."
+  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/mod tools/rw-heatmaps tools/testgrid-analysis ."
 }
 
 # maybe_run [cmd...] runs given command depending on the DRY_RUN flag.
--- a/tools/mod/tools.go
+++ b/tools/mod/tools.go
@@ -15,7 +15,7 @@
 //go:build tools
 
 // This file implements that pattern:
-// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+// https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
 // for etcd. Thanks to this file 'go mod tidy' does not removes dependencies.
 
 package tools


henrybear327:
@ivanvc , CI will pass when rebasing on main after https://github.com/etcd-io/etcd/pull/18575 is merged

tools/mod/mod.go
ivanvc, line 1: Suggestion to rename the file to `tools/mod/doc.go` to follow the same pattern across the repository.

tools/mod/mod.go
```diff
@@ -0,0 +1,23 @@
...
+
+package mod
+
+// As this directory implements the pattern for tracking tool dependencies as documented here:
+// https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+// it doesn't contain any valid Go code directly.
+
+// This would break scripts for unit testing and coverage calculation. However,
+// to ensure tools like golangci-lint and go test run normally and perform go mod checks,
+// we've added this empty file.
```
ivanvc, line 23: On that same line, these lines should be before the package definition.

ivanvc (commented):
Thanks for catching and addressing this, Henry! I left a suggestion, but if we don't want to block the PR, I'm okay with a follow-up to address them.

tools/mod/mod.go
henrybear327: Will do! Thanks for flagging this :) @ivanvc

tools/mod/mod.go
henrybear327: Will also do! Thanks @ivanvc

henrybear327:
Let’s make sure this PR is perfect before merging! Doing a follow-up will make the codebase commit history ugly :(

ivanvc (approved):
LGTM. Thanks, Henry.

jmhbnz (approved):
LGTM

cc @serathius, @ahrtr in case there is some historical reason why `tools/mod` was intentionally excluded that I am not aware of.

ivanvc:
Trying to shed more context. We're currently excluding it from `modules` (ref: #17770), as we don't want to include its dependencies in the bill of materials. But I think it should be fine (and sound) to add them to `module_dirs` so they get checked by `Makefile`'s `verify` target.

tools/mod/doc.go
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// As this directory implements the pattern for tracking tool dependencies as documented here:
+// https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+// it doesn't contain any valid Go code directly.
+//
+// This would break scripts for unit testing and coverage calculation. However,
+// to ensure tools like golangci-lint and go test run normally and perform go mod checks,
+// we've added this empty file.
```
ahrtr, line 21: Can you explain this? Why it would break scripts for unit testing and coverage calculation? also why it has some impact on golangci-lint and go test?

ahrtr:
> as we don't want to include its dependencies in the bill of materials.

I might lose some context. You added the tools/mod into the list, but why BOM wasn't changed?

tools/mod/doc.go
henrybear327: Yes definitely @ahrtr!

Please see https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/18590/pull-etcd-unit-test-386/1835058579317460992

As you can see, the main issue is 
```
% (cd tools/mod && 'env' 'ETCD_VERIFY=all' 'go' 'test' '-v' '-json' '-p=4' '-short' '-timeout=3m' '--race=false' '--cpu=1' './...')
stderr: go: warning: "./..." matched no packages
stderr: no packages to test
```

As we don't have any source file in the directory, `./...` will cause the tools to report errors, thus, causing the CI to fail 

What do you think? :)

henrybear327:
> > as we don't want to include its dependencies in the bill of materials.
> 
> I might have lost some context. You added the tools/mod into the list, but why BOM wasn't changed?

This I actually don't know. @ivanvc do you have some idea? :)

tools/mod/doc.go
ahrtr: Thanks for the clarification. It seems the reason is the module name `go.etcd.io/etcd/tools/v3` is different from the directory name `mod`?

ahrtr: Please add this into the comment to avoid any confusion for now. 

Also why not to make them consistent to avoid such unnecessary confusing comment? It's OK to resolve it separately.

ivanvc:
> This I actually don't know. @ivanvc do you have some idea? :)

It is definitely not straightforward; I hope that with the Go workspace, we can clean up/simplify some of these workflows.

I'll break down the execution stack step by step to try to make it clear for everyone.

1. Calling `make fix` includes updating the bill of materials: https://github.com/etcd-io/etcd/blob/043096067f2368e173c928581257e1ada305b593/Makefile#L72-L74
2. We'll focus on `fix-bom` target. It executes `scripts/updatebom.sh`: https://github.com/etcd-io/etcd/blob/043096067f2368e173c928581257e1ada305b593/Makefile#L84-L86
3. Executing `scripts/updatebom.sh` runs the `bom_fix` function defined in that same file: https://github.com/etcd-io/etcd/blob/c4499adfd5b42a839dbe60c5d1a01c075152e3ac/scripts/updatebom.sh#L39-L42
   1. Note that this script sources `scripts/tests_lib.sh`: https://github.com/etcd-io/etcd/blob/c4499adfd5b42a839dbe60c5d1a01c075152e3ac/scripts/updatebom.sh#L5
4. The `bom_fix` function calls the `bom_fixlet` function defined in that same file: https://github.com/etcd-io/etcd/blob/c4499adfd5b42a839dbe60c5d1a01c075152e3ac/scripts/updatebom.sh#L32-L37
5. `bom_fixlet` will iterate over `modules_exp` defined in `scripts/test_lib.sh`: https://github.com/etcd-io/etcd/blob/c4499adfd5b42a839dbe60c5d1a01c075152e3ac/scripts/updatebom.sh#L7-L19
6. Finally, `modules_exp` has the following definition (please make sure to scroll down in the code snippet area to reach `modules_exp`): https://github.com/etcd-io/etcd/blob/e094139b05244c5eab0e62172f92ac5e90feb73b/scripts/test_lib.sh#L181-L203

Therefore, running the BOM update script will not run over any of the `tools` modules.

ahrtr:
Thanks @ivanvc for the clarification.  So we have a dedicated modules (the modules_exp) for BOM. In that case, can we rename `modules_exp` to something like `modules_for_bom`?

ivanvc:
> Thanks @ivanvc for the clarification. So we have a dedicated modules (the modules_exp) for BOM. In that case, can we rename `modules_exp` to something like `modules_for_bom`?

I just verified that `modules_exp` is just called by the BOM tasks. So, this is a great suggestion. Would you suggest doing it in a follow-up?

ahrtr:
> create a `bom.sh` file that contains all the functions

Seems like a good idea, but let's do it separately.

Two followups,
- try to get rid of the confusing comment (and optionally remove the unnecessary doc.go) as mentioned in https://github.com/etcd-io/etcd/pull/18590#discussion_r1764770158
- rename `modules_exp` to something like `modules_for_bom` as mentioned in https://github.com/etcd-io/etcd/pull/18590#issuecomment-2359057943

When above two followups are done, I am open to do minor refactor as mentioned above https://github.com/etcd-io/etcd/pull/18590#issuecomment-2359066567

ahrtr (approved):
LGTM

Thanks

ahrtr:
Could anyone raise a ticket to track https://github.com/etcd-io/etcd/pull/18590#issuecomment-2359085904 before we merge this PR? Thanks

ivanvc:
> Could anyone raise a ticket to track [#18590 (comment)](https://github.com/etcd-io/etcd/pull/18590#issuecomment-2359085904) before we merge this PR? Thanks

I opened #18602 to track the rename from `modules_exp` to `modules_for_bom`.

@henrybear327, do you want to do the following in this same PR, or should we do it in a follow-up?

> try to get rid of the confusing comment https://github.com/etcd-io/etcd/pull/18590#discussion_r1764770158

Regarding:

> and optionally remove the unnecessary doc.go

@ahrtr, I think due to `golanci-lint`, we need to add the documentation for the package, that's why I suggested naming it `doc.go`. But I'm also okay with implementing something similar to the root's [`dummy.go`](https://github.com/etcd-io/etcd/blob/4f9cd336a64e0b709b2f2ef37cf3fff21a86ba39/dummy.go), and add an exception if the linter fails.

ahrtr:
> @ahrtr, I think due to `golanci-lint`, we need to add the documentation for the package, that's why I suggested naming it `doc.go`. But I'm also okay with implementing something similar to the root's [`dummy.go`](https://github.com/etcd-io/etcd/blob/4f9cd336a64e0b709b2f2ef37cf3fff21a86ba39/dummy.go), and add an exception if the linter fails. Basically rename the file from `doc.`

OK. https://github.com/etcd-io/etcd/pull/18607#issuecomment-2360534933

PR #18601, opened 2024-09-18
Support using mysql as a new storage engine for etcd.

ngaut:


hpatro:
@ngaut Interesting PR.

Curious to understand, why mysql and how does it fare against existing raft implementation in terms of performance. 
Also, does RDBMS like `mysql` provide the same guarantees like `raft` around serializability of each operation or are there any limitation to the approach ?

PR #18574, opened 2024-09-11
Ensure consistent file permissions on broken WAL files

lucasrod16:
Fixes #18571

PR #18605, opened 2024-09-19
Add function to create WAL files

lucasrod16:
Relates to https://github.com/etcd-io/etcd/pull/18574#discussion_r1758781591

This PR adds a new function to consolidate WAL file creation into a single function to help reduce errors and make the process more consistent.

### DIFF
--- a/server/storage/wal/wal.go
+++ b/server/storage/wal/wal.go
@@ -126,7 +126,7 @@ func Create(lg *zap.Logger, dirpath string, metadata []byte) (*WAL, error) {
 	}
 
 	p := filepath.Join(tmpdirpath, walName(0, 0))
-	f, err := fileutil.LockFile(p, os.O_WRONLY|os.O_CREATE, fileutil.PrivateFileMode)
+	f, err := createNewWALFile[*fileutil.LockedFile](p, false)
 	if err != nil {
 		lg.Warn(
 			"failed to flock an initial WAL file",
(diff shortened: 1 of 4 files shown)


server/storage/wal/wal_test.go
```diff
@@ -96,6 +96,60 @@ func TestNew(t *testing.T) {
...
+		{
+			name:     "creating standard file with truncate should succeed",
+			fileType: &os.File{},
+			truncate: true,
+		},
+		{
+			name:     "creating locked file with truncate should succeed",
+			fileType: &fileutil.LockedFile{},
+			truncate: true,
+		},
```
ahrtr, line 122: can you also add two cases to verify the following two cases?
- truncate:`true`, it should truncate the file content if present
- truncate: `false`: it shouldn't truncate the file content if present

server/storage/wal/wal.go
```diff
@@ -233,6 +233,31 @@ func Create(lg *zap.Logger, dirpath string, metadata []byte) (*WAL, error) {
 	return w, nil
 }
 
+// createNewWALFile creates a WAL file.
+// To create a locked file, use *fileutil.LockedFile type parameter.
+// To create a standard file, use *os.File type parameter.
+// If truncate is true, the file will be truncated if it already exists.
+func createNewWALFile[T *os.File | *fileutil.LockedFile](path string, truncate bool) (T, error) {
```
ahrtr, line 240: `truncate` is imperative name, while `forceNew` is declarative name.
```suggestion
func createNewWALFile[T *os.File | *fileutil.LockedFile](path string, forceNew bool) (T, error) {
```

lucasrod16: Updated to use declarative name `forceNew`

server/storage/wal/wal_test.go
lucasrod16: Added test cases to verify truncate behavior

serathius (approved):
Awesome work, nice use of generics.

server/storage/wal/wal_test.go
```diff
@@ -96,6 +96,73 @@ func TestNew(t *testing.T) {
...
+			require.NoError(t, err)
+
+			var f interface{}
+			switch tt.fileType.(type) {
+			case *os.File:
+				f, err = createNewWALFile[*os.File](p, tt.forceNew)
+				require.IsType(t, &os.File{}, f)
+			case *fileutil.LockedFile:
+				f, err = createNewWALFile[*fileutil.LockedFile](p, tt.forceNew)
+				require.IsType(t, &fileutil.LockedFile{}, f)
```
ahrtr, line 142:
```suggestion
			case *os.File:
				f, err = createNewWALFile[*os.File](p, tt.forceNew)
				require.IsType(t, &os.File{}, f)
			case *fileutil.LockedFile:
				f, err = createNewWALFile[*fileutil.LockedFile](p, tt.forceNew)
				require.IsType(t, &fileutil.LockedFile{}, f)
			default:	
				panic("unknown file type")					
```

lucasrod16: Added a default case to handle unknown file types. Good catch on handling that scenario

ahrtr (approved):
LGTM

Nice work!

PR #18606, opened 2024-09-19
etcdserver: fix panic when checking IsLearner of removed member

jscissr:
Previously, calling s.IsLearner() when the local node is no longer a member panics. There was an attempt to fix this by first checking IsMemberExist(), but this is not a correct fix because the member could be removed between the two calls. Instead of panicking when the member was removed, IsLearner() should return false. A node which is not a member is also not a learner.

There was a similar concurrency bug when accessing the IsLearner property of a member, which will panic with a nil pointer access error if the member is removed between the IsMemberExist() and Member() calls.

I did not add a unit test because it's basically impossible to test for such concurrency bugs.

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -817,8 +817,10 @@ func (c *RaftCluster) SetDowngradeInfo(d *serverversion.DowngradeInfo, shouldApp
 // IsMemberExist returns if the member with the given id exists in cluster.
 func (c *RaftCluster) IsMemberExist(id types.ID) bool {
 	c.Lock()
-	defer c.Unlock()
 	_, ok := c.members[id]
+	c.Unlock()
+
+	// gofail: var afterIsMemberExist struct{}
 	return ok
 }
 
--- a/tests/integration/cluster_test.go
+++ b/tests/integration/cluster_test.go
@@ -518,3 +518,51 @@ func TestSpeedyTerminate(t *testing.T) {
 	case <-donec:
 	}
 }
+
+// TestConcurrentRemoveMember demonstrated a panic in mayRemoveMember with
+// concurrent calls to MemberRemove. To reliably reproduce the panic, a delay
+// needed to be injected in IsMemberExist, which is done using a failpoint.
+// After fixing the bug, IsMemberExist is no longer called by mayRemoveMember.
+func TestConcurrentRemoveMember(t *testing.T) {
+	integration.BeforeTest(t, integration.WithFailpoint("afterIsMemberExist", `sleep("1s")`))
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer c.Terminate(t)
+
+	addResp, err := c.Members[0].Client.MemberAddAsLearner(context.Background(), []string{"http://localhost:123"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	removeID := addResp.Member.ID
+	done := make(chan struct{})
+	go func() {
+		time.Sleep(time.Second / 2)
+		c.Members[0].Client.MemberRemove(context.Background(), removeID)
+		close(done)
+	}()
+	if _, err := c.Members[0].Client.MemberRemove(context.Background(), removeID); err != nil {
+		t.Fatal(err)
+	}
+	<-done
+}
+
+func TestConcurrentMoveLeader(t *testing.T) {
+	integration.BeforeTest(t, integration.WithFailpoint("afterIsMemberExist", `sleep("1s")`))
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	defer c.Terminate(t)
+
+	addResp, err := c.Members[0].Client.MemberAddAsLearner(context.Background(), []string{"http://localhost:123"})
+	if err != nil {
+		t.Fatal(err)
+	}
+	removeID := addResp.Member.ID
+	done := make(chan struct{})
+	go func() {
+		time.Sleep(time.Second / 2)
+		c.Members[0].Client.MoveLeader(context.Background(), removeID)
+		close(done)
+	}()
+	if _, err := c.Members[0].Client.MemberRemove(context.Background(), removeID); err != nil {
+		t.Fatal(err)
+	}
+	<-done
+}
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1224,7 +1224,8 @@ func (s *EtcdServer) isLeader() bool {
 
 // MoveLeader transfers the leader to the given transferee.
 func (s *EtcdServer) MoveLeader(ctx context.Context, lead, transferee uint64) error {
-	if !s.cluster.IsMemberExist(types.ID(transferee)) || s.cluster.Member(types.ID(transferee)).IsLearner {
+	member := s.cluster.Member(types.ID(transferee))
+	if member == nil || member.IsLearner {
 		return errors.ErrBadLeaderTransferee
 	}
 
@@ -1593,9 +1594,9 @@ func (s *EtcdServer) mayRemoveMember(id types.ID) error {
 	}
 
 	lg := s.Logger()
-	isLearner := s.cluster.IsMemberExist(id) && s.cluster.Member(id).IsLearner
+	member := s.cluster.Member(id)
 	// no need to check quorum when removing non-voting member
-	if isLearner {
+	if member != nil && member.IsLearner {
 		return nil
 	}
 
--- a/tests/robustness/makefile.mk
+++ b/tests/robustness/makefile.mk
@@ -54,7 +54,7 @@ install-gofail: $(GOPATH)/bin/gofail
 
 .PHONY: gofail-enable
 gofail-enable: $(GOPATH)/bin/gofail
-	$(GOPATH)/bin/gofail enable server/etcdserver/ server/lease/leasehttp server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/
+	$(GOPATH)/bin/gofail enable server/etcdserver/ server/lease/leasehttp server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/ server/etcdserver/api/membership/
 	cd ./server && go get go.etcd.io/gofail@${GOFAIL_VERSION}
 	cd ./etcdutl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
 	cd ./etcdctl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
@@ -62,7 +62,7 @@ gofail-enable: $(GOPATH)/bin/gofail
 
 .PHONY: gofail-disable
 gofail-disable: $(GOPATH)/bin/gofail
-	$(GOPATH)/bin/gofail disable server/etcdserver/ server/lease/leasehttp server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/
+	$(GOPATH)/bin/gofail disable server/etcdserver/ server/lease/leasehttp server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/ server/etcdserver/api/membership/
 	cd ./server && go mod tidy
 	cd ./etcdutl && go mod tidy
 	cd ./etcdctl && go mod tidy


ahrtr:
> Previously, calling s.IsLearner() when the local node is no longer a member panics.

Did you see a case where panic happened or can you create an e2e or integration test case to make it happen?

jscissr:
> Did you see a case where panic happened or can you create an e2e or integration test case to make it happen?

Yes I can, here are integration tests which demonstrate the panic. I need to add an artificial delay in IsMemberExist to reliably show the panic.

```patch
diff --git a/server/etcdserver/api/membership/cluster.go b/server/etcdserver/api/membership/cluster.go
index 6becdfd62..4b6dbda64 100644
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -816,6 +816,7 @@ func (c *RaftCluster) SetDowngradeInfo(d *serverversion.DowngradeInfo, shouldApp
 
 // IsMemberExist returns if the member with the given id exists in cluster.
 func (c *RaftCluster) IsMemberExist(id types.ID) bool {
+	defer time.Sleep(time.Second)
 	c.Lock()
 	defer c.Unlock()
 	_, ok := c.members[id]
diff --git a/tests/integration/cluster_test.go b/tests/integration/cluster_test.go
index 29f8ae8dd..852a11e85 100644
--- a/tests/integration/cluster_test.go
+++ b/tests/integration/cluster_test.go
@@ -201,6 +201,56 @@ func TestAddMemberAfterClusterFullRotation(t *testing.T) {
 	clusterMustProgress(t, c.Members)
 }
 
+func TestConcurrentRemoveMember(t *testing.T) {
+	integration.BeforeTest(t)
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 2})
+	defer c.Terminate(t)
+
+	time.Sleep(time.Second)
+	removeID := uint64(c.Members[1].Server.MemberID())
+	go func() {
+		time.Sleep(time.Second / 2)
+		c.Members[0].Client.MemberRemove(context.Background(), removeID)
+	}()
+	if _, err := c.Members[0].Client.MemberRemove(context.Background(), removeID); err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(time.Second)
+}
+
+func TestConcurrentMoveLeader(t *testing.T) {
+	integration.BeforeTest(t)
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 2})
+	defer c.Terminate(t)
+
+	time.Sleep(time.Second)
+	removeID := uint64(c.Members[1].Server.MemberID())
+	go func() {
+		time.Sleep(time.Second / 2)
+		c.Members[0].Client.MoveLeader(context.Background(), removeID)
+	}()
+	if _, err := c.Members[0].Client.MemberRemove(context.Background(), removeID); err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(time.Second)
+}
+
+func TestConcurrentUnary(t *testing.T) {
+	integration.BeforeTest(t)
+	c := integration.NewCluster(t, &integration.ClusterConfig{Size: 2})
+	defer c.Terminate(t)
+
+	time.Sleep(2 * time.Second)
+	go func() {
+		time.Sleep(time.Second + time.Second/2)
+		c.Members[0].Client.Get(context.Background(), "key")
+	}()
+	if _, err := c.Members[0].Client.MemberRemove(context.Background(), uint64(c.Members[0].Server.MemberID())); err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(time.Second)
+}
+
 // TestIssue2681 ensures we can remove a member then add a new one back immediately.
 func TestIssue2681(t *testing.T) {
 	integration.BeforeTest(t)
```

Here are the stack traces when running these tests:

```
% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' './integration/...' '-timeout=15m' '--race' '-run=TestConcurrentRemoveMember' '-p=2')
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x145df4d]

goroutine 218 [running]:
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).mayRemoveMember(0xc0003d3808, 0xa5a138a8b8d2b107)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/server.go:1596 +0xed
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).RemoveMember(0xc0003d3808, {0x1cf91e0, 0xc001dbe540}, 0xa5a138a8b8d2b107)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/server.go:1428 +0x85
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*ClusterServer).MemberRemove(0xc000012ae0, {0x1cf91e0, 0xc001dbe540}, 0xc001dbe570)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/api/v3rpc/member.go:71 +0xad


% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' './integration/...' '-timeout=15m' '--race' '-run=TestConcurrentMoveLeader' '-p=2')
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1459bed]

goroutine 207 [running]:
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).MoveLeader(0xc0003fb508, {0x1cf91e0, 0xc001e33bc0}, 0x984e59084a1f2c8b, 0xb3a11ebbc585e63a)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/server.go:1227 +0xcd
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*maintenanceServer).MoveLeader(0xc00041f790, {0x1cf91e0, 0xc001e33bc0}, 0xc001e33bf0)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/api/v3rpc/maintenance.go:282 +0x116
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.(*authMaintenanceServer).MoveLeader(0xc000600ef0, {0x1cf91e0, 0xc001e33bc0}, 0xc001e33bf0)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/api/v3rpc/maintenance.go:347 +0xb1


% (cd tests && 'env' 'ETCD_VERIFY=all' 'go' 'test' './integration/...' '-timeout=15m' '--race' '-run=TestConcurrentUnary' '-p=2')
panic: failed to find local ID in cluster members

goroutine 365 [running]:
go.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x2, 0xc000fa95f0, {0xc000229a00?, 0x2?, 0x2?})
        /home/jan/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:196 +0x9f
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc000fa95f0, {0xc000229980, 0x2, 0x2})
        /home/jan/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:262 +0x3ad
go.uber.org/zap.(*Logger).Panic(0xc000084a00, {0x1aa86bb, 0x2a}, {0xc000229980, 0x2, 0x2})
        /home/jan/go/pkg/mod/go.uber.org/zap@v1.27.0/logger.go:285 +0x68
go.etcd.io/etcd/server/v3/etcdserver/api/membership.(*RaftCluster).IsLocalMemberLearner(0xc0003a04d0)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/api/membership/cluster.go:786 +0x35b
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).IsLearner(0xc000295508)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/server.go:2471 +0x45
go.etcd.io/etcd/server/v3/etcdserver/api/v3rpc.Server.newUnaryInterceptor.func4({0x1cf91e0, 0xc001ed27b0}, {0x1a3ec60, 0xc000002e10}, 0xc001ea4320?, 0xc001eab540)
        /home/jan/Documents/git-thirdparty/etcd/server/etcdserver/api/v3rpc/interceptor.go:52 +0x9a
```

ahrtr:
Thanks for the integration test cases.

I agree that it may happen theoretically. Did you ever see the panic in production or your test environment with the official etcd releases? I believe not. So overall minor issues to me.

The change to `server/etcdserver/server.go` is straightforward and safe, please also add test cases `TestConcurrentMoveLeader` and `TestConcurrentRemoveMember` into this PR. You can use the [gofailpoint](https://github.com/etcd-io/gofail) to trigger the `sleep` failpoint. See an example below 
https://github.com/etcd-io/etcd/blob/59cfd7a61d3431d25657c2dc4625fe4923113dd3/tests/integration/v3_lease_test.go#L1086

For the change to `server/etcdserver/api/v3rpc/interceptor.go`, I prefer to keep it unchanged,
- It changes the behaviour/protocal between etcdserver and client.
- Also it makes more sense to return an error something "member not found" instead of `ErrGRPCNotSupportedForLearner`;
- When the local member is removed from the cluster, it will eventually stop automatically. A panic right before stopping might not be too serious.

jscissr:
I found the bug by reading the code, and have indeed not observed it happen without the added delay.

I have added the `TestConcurrentMoveLeader` and `TestConcurrentRemoveMember` tests to the PR, with failpoint and reliability improvements. Though, I'm not sure that adding these tests to the codebase has much value. The only way they could catch a bug is if someone adds back calls to `IsMemberExist` in these functions.

For `server/etcdserver/api/v3rpc/interceptor.go`, I don't understand your concern. The existing behavior is left unchanged by my changes, they only reduce the set of possible behaviors by removing the possibility of panicking. I changed `IsLocalMemberLearner` to return false when the local node is not a member. This means that `s.IsMemberExist(s.MemberID()) && s.IsLearner()` is equivalent to `s.IsLearner()`.

server/etcdserver/api/membership/cluster.go
```diff
@@ -783,11 +783,7 @@ func (c *RaftCluster) IsLocalMemberLearner() bool {
 	defer c.Unlock()
 	localMember, ok := c.members[c.localID]
 	if !ok {
-		c.lg.Panic(
-			"failed to find local ID in cluster members",
-			zap.String("cluster-id", c.cid.String()),
-			zap.String("local-member-id", c.localID.String()),
-		)
```
ahrtr, line 790: Suggest not to change this. I don't think it will happen in production or test environment. If it happens, it means something critical occurs.

ahrtr:
> For `server/etcdserver/api/v3rpc/interceptor.go`, I don't understand your concern. The existing behavior is left unchanged by my changes, they only reduce the set of possible behaviors by removing the possibility of panicking.

Please read https://github.com/etcd-io/etcd/pull/18606#discussion_r1775665002, and also comments below,
- Previously etcdserver panics in such case, but now it returns a `rpctypes.ErrGRPCNotSupportedForLearner` error. So it changes the behaviour, and it may mislead the users.
- Also an error something like "member not found" is more proper in such case.

Also as mentioned previously, when the local member is removed from the cluster, it will eventually stop automatically. A panic right before stopping might not be too serious.

So I suggest not to change
- server/etcdserver/api/v3rpc/interceptor.go
- server/etcdserver/api/membership/cluster.go

server/etcdserver/api/membership/cluster.go
```diff
@@ -816,6 +812,9 @@ func (c *RaftCluster) SetDowngradeInfo(d *serverversion.DowngradeInfo, shouldApp
 
 // IsMemberExist returns if the member with the given id exists in cluster.
 func (c *RaftCluster) IsMemberExist(id types.ID) bool {
+	// gofail: var sleepAfterIsMemberExist struct{}
+	// defer time.Sleep(time.Second)
+
```
ahrtr, line 817: You can add a failpoint right above the line 821 "`return ok`",

```
// gofail: var sleepAfterIsMemberExist struct{}
```

and inject `sleep("1s")` to it during test

jscissr:
I removed the changes which you dislike, and adjusted the failpoint.

ahrtr (approved):
LGTM

Thanks

tests/integration/cluster_test.go
```diff
@@ -518,3 +518,51 @@ func TestSpeedyTerminate(t *testing.T) {
 	case <-donec:
 	}
 }
+
+// TestConcurrentRemoveMember demonstrated a panic in mayRemoveMember with
+// concurrent calls to MemberRemove. To reliably reproduce the panic, a delay
```
serathius, line 523: There should never be a concurrent call to `MemberRemove`. Cluster state changes should be all done via apply loop, which guarantees that there is only one concurrent change to cluster state. 

The reason for locking in cluster struct (and any other internal structs) is to prevent reader+writer races. There should never be writer+writer race.

ahrtr: This PR has nothing to do with the apply loop/workflow. It's fixing a panic in the API layer. Refer to https://github.com/etcd-io/etcd/pull/18606#issuecomment-2367970699

serathius: Oh, I didn't notice that this is integration test. I'm just little confused why MemberRemove call by client is executed concurrently. Should it be linearized?

ahrtr: > I'm just little confused why MemberRemove call by client is executed concurrently.

Right, usually we don't expect users to do this. But etcdserver shouldn't panic due to whatever users' inappropriate behaviour. Note that the panicking happens at API layer instead of the apply loop/workflow.

serathius: Sorry, didn't mean "why?", but "how?". Agree that this is a problem, but I'm trying to understand how but in cluster struct surfaces to API.

serathius (approved):
Independent of discussion, this change is good to merge.

PR #18607, opened 2024-09-19
change package name for tools and remove unnecessary doc.go

ahrtr:


ahrtr:
The libs.go and tools.go have build constraint like `//go:build tools`, so the pull-etcd-verify still complains `context loading failed: no go files to analyze`.

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/18607/pull-etcd-verify/1836702520009822208

So we can't remove the doc.go file.

PR #18622, opened 2024-09-23
etcdserver: compact raft log entries more frequently

clement2026:
Part of  #17098 

## Goal
Compact raft log more frequently to save memory.

## Key Changes

* Added a new function `maybeCompactRaftLog` to handle raft log compaction.
* Introduce 2 variables: `CompactRaftLogEveryNApplies` and `DefaultCompactRaftLogEveryNApplies` to control how frequently the raft log is compacted.
* Fix a test

- [ ] Add tests for the changes
- [ ] Run benchmarks


server/etcdserver/server.go
```diff
@@ -2170,6 +2183,18 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
...
+func (s *EtcdServer) maybeCompactRaftLog(ep *etcdProgress) {
+	lg := s.Logger()
+
+	// Keep some in memory log entries for slow followers, while keeping the entries up to snapshot index.
+	// Only compact raft log once every N applies
+	if ep.appliedi <= ep.snapi+s.Cfg.SnapshotCatchUpEntries || ep.appliedi%s.Cfg.CompactRaftLogEveryNApplies != 0 {
+		return
+	}
+
+	compacti := ep.appliedi - s.Cfg.SnapshotCatchUpEntries
```
serathius, line 2197:
```suggestion
	compacti := ep.snapi - s.Cfg.SnapshotCatchUpEntries
```

serathius, line 2197:
```suggestion
	// Retain all log entries up to the latest snapshot index to ensure any member can recover from that snapshot. 
    // Beyond the snapshot index, preserve the most recent s.Cfg.SnapshotCatchUpEntries entries in memory. 
    // This allows slow followers to catch up by synchronizing entries instead of requiring a full snapshot transfer.
	// Only compact raft log once every N applies
	if ep.snapi > s.Cfg.SnapshotCatchUpEntries  {
		return
	}

	compacti := ep.snapi - s.Cfg.SnapshotCatchUpEntries
```

In the first PR, let's keep the same number of entries, just make the compaction independent from snapshot.

serathius:
This change does to many things, there are two changes here that should be independent:
* Separate `maybeCompactRaftLog` function allowing us to compact log independently from snapshots.
* Run compaction more frequently. As currently compacti is dependent on snapshot, we would first need to confirm that we are ok with lower  number of entries.

clement2026:
Sure. I'll break down this PR.

PR #18635, opened 2024-09-24
separate maybeCompactRaftLog function to compact raft log independent…

clement2026:
Part of  #17098 

Goal: separate maybeCompactRaftLog function to compact raft log independently from snapshots.

## TODO
- [ ] Add tests for the changes

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -980,6 +980,7 @@ func (s *EtcdServer) applyAll(ep *etcdProgress, apply *toApply) {
 	<-apply.notifyc
 
 	s.triggerSnapshot(ep)
+	s.maybeCompactRaftLog(ep)
 	select {
 	// snapshot requested via send()
 	case m := <-s.r.msgSnapC:
@@ -2170,6 +2171,21 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
 		"saved snapshot",
 		zap.Uint64("snapshot-index", snap.Metadata.Index),
 	)
+}
+
+func (s *EtcdServer) maybeCompactRaftLog(ep *etcdProgress) {
+	// Retain all log entries up to the latest snapshot index to ensure any member can recover from that snapshot.
+	// Beyond the snapshot index, preserve the most recent s.Cfg.SnapshotCatchUpEntries entries in memory.
+	// This allows slow followers to catch up by synchronizing entries instead of requiring a full snapshot transfer.
+	// Only compact raft log once every N applies
+	if ep.snapi <= s.Cfg.SnapshotCatchUpEntries  {
+		return
+	}
+
+	// make sure compacti >= 0, because s.r.raftStorage.Compact(compacti) always returns ErrCompacted if compacti <= 0
+	compacti := ep.snapi - s.Cfg.SnapshotCatchUpEntries
+
+	lg := s.Logger()
 
 	// When sending a snapshot, etcd will pause compaction.
 	// After receives a snapshot, the slow follower needs to get all the entries right after
@@ -2181,13 +2197,7 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
 		return
 	}
 
-	// keep some in memory log entries for slow followers.
-	compacti := uint64(1)
-	if snapi > s.Cfg.SnapshotCatchUpEntries {
-		compacti = snapi - s.Cfg.SnapshotCatchUpEntries
-	}
-
-	err = s.r.raftStorage.Compact(compacti)
+	err := s.r.raftStorage.Compact(compacti)
 	if err != nil {
 		// the compaction was done asynchronously with the progress of raft.
 		// raft log might already been compact.


clement2026:
@serathius Please take a look. This PR focuses on one task only.

server/etcdserver/server.go
```diff
@@ -2181,13 +2195,7 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
...
 	}
 
-	// keep some in memory log entries for slow followers.
-	compacti := uint64(1)
-	if snapi > s.Cfg.SnapshotCatchUpEntries {
-		compacti = snapi - s.Cfg.SnapshotCatchUpEntries
-	}
-
-	err = s.r.raftStorage.Compact(compacti)
+	err := s.r.raftStorage.Compact(compacti)
```
serathius, line 2198: To avoid calling Compact multiple times with the same value, maybe we can add `compacti` field to `etcdProgress`? It's default value can be taken from `s.r.raftStorage.FirstIndex() - 1`.

It would give us a clearer view on raft log compaction.

clement2026: @serathius This approach is more efficient. I’ve implemented your suggestion and added a test case.

clement2026:
I can't test `ep.compacti` directly because it's not accessible. However, I checked the logs and confirmed it’s working as expected.

## Add log statements
```go
	ep := etcdProgress{
		confState: sn.Metadata.ConfState,
		snapi:     sn.Metadata.Index,
		appliedt:  sn.Metadata.Term,
		appliedi:  sn.Metadata.Index,
		compacti:  fi - 1,
	}
	s.Logger().Info("initial ep.compacti", zap.Uint64("ep.compacti", ep.compacti),zap.Uint64("ep.snapi", ep.snapi))
```


```go
	compacti := ep.snapi - s.Cfg.SnapshotCatchUpEntries
	if compacti <= ep.compacti {
		return
	}
	s.Logger().Info("compacti > ep.compacti", zap.Uint64("compacti", compacti), zap.Uint64("ep.compacti", ep.compacti))

```
## Start a new etcd instance and make 110 put requests

```bash
➜  rm -rf default.etcd;  bin/etcd --experimental-snapshot-catchup-entries=5 --snapshot-count=10  2>&1 | grep 'ep.compacti'

➜ benchmark put --key-space-size=9999999 --val-size=100 --total=110
```

Logs:
```
{"level":"info","ts":"2024-09-24T23:38:10.660332+0800","caller":"etcdserver/server.go:823","msg":"initial ep.compacti","ep.compacti":0,"ep.snapi":0}
{"level":"info","ts":"2024-09-24T23:38:16.890460+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":6,"ep.compacti":0}
{"level":"info","ts":"2024-09-24T23:38:16.959513+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":17,"ep.compacti":6}
{"level":"info","ts":"2024-09-24T23:38:17.017577+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":28,"ep.compacti":17}
{"level":"info","ts":"2024-09-24T23:38:17.096548+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":39,"ep.compacti":28}
{"level":"info","ts":"2024-09-24T23:38:17.159693+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":50,"ep.compacti":39}
{"level":"info","ts":"2024-09-24T23:38:17.238754+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":61,"ep.compacti":50}
{"level":"info","ts":"2024-09-24T23:38:17.316651+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":72,"ep.compacti":61}
{"level":"info","ts":"2024-09-24T23:38:17.391588+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":83,"ep.compacti":72}
{"level":"info","ts":"2024-09-24T23:38:17.460776+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":94,"ep.compacti":83}
{"level":"info","ts":"2024-09-24T23:38:17.539552+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":105,"ep.compacti":94}
^C
```

## Restart the etcd instance and make 50 put requests

```bash
➜  bin/etcd --experimental-snapshot-catchup-entries=5 --snapshot-count=10  2>&1 | grep 'ep.compacti'

➜ benchmark put --key-space-size=9999999 --val-size=100 --total=50
```

Logs:
```
{"level":"info","ts":"2024-09-24T23:38:24.076007+0800","caller":"etcdserver/server.go:823","msg":"initial ep.compacti","ep.compacti":110,"ep.snapi":110}
{"level":"info","ts":"2024-09-24T23:39:16.807112+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":116,"ep.compacti":110}
{"level":"info","ts":"2024-09-24T23:39:16.880017+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":127,"ep.compacti":116}
{"level":"info","ts":"2024-09-24T23:39:16.941133+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":138,"ep.compacti":127}
{"level":"info","ts":"2024-09-24T23:39:17.012076+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":149,"ep.compacti":138}
{"level":"info","ts":"2024-09-24T23:39:17.075207+0800","caller":"etcdserver/server.go:2195","msg":"compacti > ep.compacti","compacti":160,"ep.compacti":149}
```

server/etcdserver/server.go
```diff
@@ -764,6 +765,10 @@ func (s *EtcdServer) run() {
 	if err != nil {
 		lg.Panic("failed to get snapshot from Raft storage", zap.Error(err))
 	}
+	fi, err := s.r.raftStorage.FirstIndex()
```
serathius, line 768: Please don't use short obfuscated variable name for variables spanning hundred lines of code. Either:
* Move the call to FirstIndex next to creation of etcdProgress
* Use proper variable name like `firstRaftIndex`.

server/etcdserver/server.go
```diff
@@ -813,6 +818,7 @@ func (s *EtcdServer) run() {
 		snapi:     sn.Metadata.Index,
 		appliedt:  sn.Metadata.Term,
 		appliedi:  sn.Metadata.Index,
+		compacti:  fi - 1,
```
serathius, line 821: Would be good to add a comment why we substract one. Compact index should be the last index that raftStorage.Compact was called. Whether we should substract depend on whether the Compact operation is inclusive or exclusive. Need to confirm this.

server/etcdserver/server.go
```diff
@@ -2181,13 +2204,8 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
...
 
-	// keep some in memory log entries for slow followers.
-	compacti := uint64(1)
-	if snapi > s.Cfg.SnapshotCatchUpEntries {
-		compacti = snapi - s.Cfg.SnapshotCatchUpEntries
-	}
-
-	err = s.r.raftStorage.Compact(compacti)
+	err := s.r.raftStorage.Compact(compacti)
+	ep.compacti = compacti
```
serathius, line 2208: I think we should update compact id if there was no error.

server/etcdserver/memory_storage_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// after compacting, entries should be:
+	// [ {Index: 3}, {Index: 4}, {Index: 5} ]
+	err = raftStorage.Compact(3)
+	assert.NoError(t, err)
+
+
+	firstIndex, err = raftStorage.FirstIndex()
+	assert.NoError(t, err)
+	assert.Equal(t, uint64(3+1), firstIndex)
```
serathius, line 66:
```suggestion
	compacti := 3
	err = raftStorage.Compact(compacti)
	assert.NoError(t, err)


	firstIndex, err = raftStorage.FirstIndex()
	assert.NoError(t, err)
	assert.Equal(t, uint64(compacti+1), firstIndex)
```

server/etcdserver/memory_storage_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"go.etcd.io/raft/v3"
+	"go.etcd.io/raft/v3/raftpb"
+)
+
+// TestMemoryStorageCompaction tests that after calling raftStorage.Compact(compacti)
+// without errors, the dummy entry becomes {Index: compacti} and
+// raftStorage.FirstIndex() returns (compacti+1, nil).
+func TestMemoryStorageCompaction(t *testing.T) {
```
serathius, line 29: Thanks for the test, it's really great to protect our assumption via test.

serathius, line 29:
```suggestion
// TestMemoryStorageCompactInclusive tests that compacting index is inclusive, meaning that first index after compaction is larger by one than compacted index.
func TestMemoryStorageCompactInclusive(t *testing.T) {
```

server/etcdserver/server.go
```diff
@@ -818,7 +818,13 @@ func (s *EtcdServer) run() {
 		snapi:     sn.Metadata.Index,
 		appliedt:  sn.Metadata.Term,
 		appliedi:  sn.Metadata.Index,
-		compacti:  fi - 1,
+		// compacti is the index from the last time raftStorage.Compact was called
+		// without errors.
+		//
+		// After calling raftStorage.Compact(compacti) without errors, the dummy entry of
+		// raftStorage becomes {Index: compacti}, and raftStorage.FirstIndex() returns
+		// (compacti+1, nil). This is validated by TestMemoryStorageCompaction.
```
serathius, line 826:
```suggestion
		// Compaction is inclusive, meaning compact index should be lower by one than the first index after compaction. 
		// This is validated by TestMemoryStorageCompaction.
```
Don't describe the variables in comments, if you want to make `compacti` more readable, then make the variable names more descriptive, but let's not address it now.

server/etcdserver/memory_storage_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"go.etcd.io/raft/v3"
+	"go.etcd.io/raft/v3/raftpb"
+)
+
+// TestMemoryStorageCompactInclusive tests that compacting index is inclusive, meaning that first index after compaction is larger by one than compacted index.
```
serathius, line 26:
```suggestion
// TestMemoryStorageCompactInclusive tests that compacting is inclusive, meaning that first index after compaction is larger by one than compacted index.
```

serathius: Sorry my mistake.

server/etcdserver/server.go
serathius: Awesome work.

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,95 @@
...
+		Size:                   1,
+		SnapshotCount:          10,
+		SnapshotCatchUpEntries: 5,
+	})
+	defer clus.Terminate(t)
+
+	mem := clus.Members[0]
+	kvc := integration.ToGRPC(mem.Client).KV
+
+	// When starting a new cluster with 1 member, the member will have an index of 4.
```
serathius, line 41: All the cluster metadata needs to be established via raft, this means that during bootstrap of cluster and on each member start there is a set of proposals that always goes through, even if nothing is happening in the cluster.

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// TODO: Can someone explain this?
+	// Currently, if `ep.appliedi-ep.snapi > s.Cfg.SnapshotCount`,
+	// a raft log snapshot is created, and raft log entries are compacted.
+	// In this case, it triggers when the index is a multiple of 11.
+	appliedi := 4
```
serathius, line 46: Let's not hardcode base index. How many entries are written during cluster boostrap is an implementation detail that we should not care about. Can you write the test in a way that doesn't need to hardcode number `4`?

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,95 @@
...
+	expectMemberLog(t, mem, time.Second, "compacted Raft logs", 100)
+	expectMemberLog(t, mem, time.Second, "\"compact-index\": 1095", 1)
+
+	// No more snapshot and compaction should happen.
+	expectMemberLogTimeout(t, mem, 5*time.Second, "saved snapshot", 101)
+	expectMemberLogTimeout(t, mem, time.Second, "compacted Raft logs", 101)
+}
+
+// expectMemberLogTimeout ensures that the log has fewer than `count` occurrences of `s` before timing out
+func expectMemberLogTimeout(t *testing.T, m *integration.Member, timeout time.Duration, s string, count int) {
```
serathius, line 85:
```suggestion
func logOccuredLessThanNTimes(t *testing.T, m *integration.Member, timeout time.Duration, s string, count int) {
```

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// In this case, it triggers when the index is a multiple of 11.
+	appliedi := 4
+	for ; appliedi <= 10; appliedi++ {
+		_, err := kvc.Put(context.TODO(), &pb.PutRequest{Key: []byte("foo"), Value: []byte("bar")})
+		if err != nil {
+			t.Errorf("#%d: couldn't put key (%v)", appliedi, err)
+		}
+	}
+	// The first snapshot and compaction shouldn't happen because the index is less than 11
+	expectMemberLogTimeout(t, mem, 5*time.Second, "saved snapshot", 1)
```
serathius, line 54: Checking for less than `1` occurrence seems not very clear. Have you considered testing for at most N occurrences instead of less than N? Then checking that "saved snapshot" occured at most zero times better matches the comment above.

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,95 @@
...
+	expectMemberLogTimeout(t, mem, time.Second, "compacted Raft logs", 1)
+
+	for ; appliedi <= 11; appliedi++ {
+		_, err := kvc.Put(context.TODO(), &pb.PutRequest{Key: []byte("foo"), Value: []byte("bar")})
+		if err != nil {
+			t.Errorf("#%d: couldn't put key (%v)", appliedi, err)
+		}
+	}
+	// The first snapshot and compaction should happen because the index is 11
+	expectMemberLog(t, mem, 5*time.Second, "saved snapshot", 1)
```
serathius, line 64: This is exactly same validation as above, but comment expect compaction has happened. What gives?

tests/integration/raft_log_test.go
clement2026: Sure, it seems I can get the index from `mem.Client.Status` like this:
```
status,_ := mem.Client.Status(context.TODO(),"endpoint")
status.RaftAppliedIndex
```

serathius: Good idea!

tests/integration/raft_log_test.go
clement2026: The above validation ensures a snapshot mustn't happen.
This line ensures a snapshot must happen.

tests/integration/raft_log_test.go
clement2026: This sounds good! I'll apply the suggestion.

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,114 @@
...
+	//
+	// increase applied index to 10
+	for ; appliedi < 10; appliedi++ {
+		_, err := kvc.Put(context.TODO(), &pb.PutRequest{Key: []byte("foo"), Value: []byte("bar")})
+		if err != nil {
+			t.Errorf("#%d: couldn't put key (%v)", appliedi, err)
+		}
+	}
+	// The first snapshot and compaction shouldn't happen because applied index is less than 11
+	logOccurredAtMostNTimes(t, mem, 5*time.Second, "saved snapshot", 0)
```
serathius, line 65: Those timeouts are very high, can we not have as big timeout? Snapshot and compaction should happen synchronously to operations, meaning logs should be available immidietly.

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,114 @@
...
+	logOccurredAtMostNTimes(t, mem, time.Second, "compacted Raft logs", 0)
```
serathius, line 66: Is there a reason sometimes we wait for 1 second, sometimes for 5 seconds?

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// increase applied index to 11
+	for ; appliedi < 11; appliedi++ {
+		_, err := kvc.Put(context.TODO(), &pb.PutRequest{Key: []byte("foo"), Value: []byte("bar")})
+		if err != nil {
+			t.Errorf("#%d: couldn't put key (%v)", appliedi, err)
+		}
+	}
+	// The first snapshot and compaction should happen because applied index is 11
+	logOccurredAtMostNTimes(t, mem, 5*time.Second, "saved snapshot", 1)
```
serathius, line 76: Should we wait for the log to happen at least once instead of at most once?

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+	// With applied index at 1100, snapshot and compaction should happen 100 times.
+	logOccurredAtMostNTimes(t, mem, 5*time.Second, "saved snapshot", 100)
+	logOccurredAtMostNTimes(t, mem, time.Second, "compacted Raft logs", 100)
+	expectMemberLog(t, mem, time.Second, "\"compact-index\": 1095", 1)
+}
+
+// logOccurredAtMostNTimes ensures that the log has exactly `count` occurrences of `s` before timing out, no more, no less.
+func logOccurredAtMostNTimes(t *testing.T, m *integration.Member, timeout time.Duration, s string, count int) {
```
serathius, line 94:
```suggestion

func logOccurredExactlyNTimes(t *testing.T, m *integration.Member, timeout time.Duration, s string, count int) {
```

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,114 @@
...
+	defer cancel()
+
+	// The log must have `count` occurrences before timeout
+	_, err := m.LogObserver.Expect(ctx, s, count)
+	if err != nil {
+		t.Fatalf("failed to expect(log:%s, count:%d): %v", s, count, err)
+	}
+
+	// The log mustn't have `count+1` occurrences before timeout
+	lines, err := m.LogObserver.Expect(ctx, s, count+1)
```
serathius, line 105: Maybe instead of calling `Expect` twice, you could add this function to logObserver?
* Rename `Expect` to `ExpectAtLeastNOccurances`
* Create ExpectExactlyNOccurances`

tests/integration/raft_log_test.go
clement2026: The first `logOccurredAtMostNTimes` waits for 5 seconds. Once it's done, we can assume the log is synced up, so the second `logOccurredAtMostNTimes` doesn't need to wait that long.

tests/integration/raft_log_test.go
clement2026: Yes, I’d love to. That will make things much clearer.

tests/integration/raft_log_test.go
clement2026: logOccurredAtMostNTimes should be changed to logOccurredExactlyNTimes.

tests/integration/raft_log_test.go
clement2026: Alright, I'll reduce the timeout to 1 or 2 seconds second and see how it works out.

tests/integration/raft_log_test.go
serathius: We should have times in matter of tens of miliseconds not multiple seconds. The whole test should take seconds.

server/etcdserver/server.go
```diff
@@ -2170,6 +2180,22 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
 		"saved snapshot",
 		zap.Uint64("snapshot-index", snap.Metadata.Index),
 	)
+}
+
+func (s *EtcdServer) maybeCompactRaftLog(ep *etcdProgress) {
+	// Retain all log entries up to the latest snapshot index to ensure any member can recover from that snapshot.
+	// Beyond the snapshot index, preserve the most recent s.Cfg.SnapshotCatchUpEntries entries in memory.
+	// This allows slow followers to catch up by synchronizing entries instead of requiring a full snapshot transfer.
+	if ep.snapi <= s.Cfg.SnapshotCatchUpEntries {
```
ahrtr, line 2189:
```suggestion
	if ep.appliedi <= s.Cfg.SnapshotCatchUpEntries {
```

server/etcdserver/server.go
```diff
@@ -2170,6 +2180,22 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
...
+		return
+	}
+
+	compacti := ep.snapi - s.Cfg.SnapshotCatchUpEntries
```
ahrtr, line 2193:
```suggestion
	compacti := ep.appliedi - s.Cfg.SnapshotCatchUpEntries
```

ahrtr: What's the purpose of the change?
If you use `ep.snapi`, then the behaviour is exactly the same as existing behaviour, because `ep.snapi` only gets updated each time after creating the (v2) snapshot.

ahrtr: Please see https://github.com/etcd-io/etcd/issues/17098#issuecomment-2374327464. Can we get that confirmed firstly?

server/etcdserver/server.go
```diff
@@ -980,6 +989,7 @@ func (s *EtcdServer) applyAll(ep *etcdProgress, apply *toApply) {
 	<-apply.notifyc
 
 	s.triggerSnapshot(ep)
+	s.maybeCompactRaftLog(ep)
```
ahrtr, line 992: If you really want to get this small merged firstly, then please ensure it's as independent as possible.  Currently the `s.snapshot` performs both snapshot and compaction operations. It makes sense to extract the compaction operation as an independent function/method, but let's call the method inside `s.triggerSnapshot`,

```
func (s *EtcdServer) triggerSnapshot(ep *etcdProgress) {
	if !s.shouldSnapshot(ep) {
		return
	}
	lg := s.Logger()
	lg.Info(
		"triggering snapshot",
		zap.String("local-member-id", s.MemberID().String()),
		zap.Uint64("local-member-applied-index", ep.appliedi),
		zap.Uint64("local-member-snapshot-index", ep.snapi),
		zap.Uint64("local-member-snapshot-count", s.Cfg.SnapshotCount),
		zap.Bool("snapshot-forced", s.forceSnapshot),
	)
	s.forceSnapshot = false

	s.snapshot(ep.appliedi, ep.confState)
	ep.snapi = ep.appliedi
	s.compact(xxxxx)   // call the new method here, so we still do it each time after creating a snapshot.

}
```

clement2026: No rush on merging this PR. If we do merge it, we need to ensure etcd actually benefits from it. Let's resolve https://github.com/etcd-io/etcd/issues/17098#issuecomment-2374327464 first.

server/etcdserver/server.go
serathius: snapi is more correct here, the motivation is described in the comment.

server/etcdserver/server.go
serathius: This change doesn't change when compaction is run or how many times it's executed. I was aware of https://github.com/etcd-io/raft/blob/5d6eb55c4e6929e461997c9113aba99a5148e921/storage.go#L266-L269 code, that's why I was proposing compacting only ever X applies.

server/etcdserver/server.go
serathius: The goal of the PR is to make compaction independent from snapshot. Not just refactor the function.

tests/integration/raft_log_test.go
```diff
@@ -0,0 +1,114 @@
...
+	for ; appliedi < 11; appliedi++ {
+		_, err := kvc.Put(context.TODO(), &pb.PutRequest{Key: []byte("foo"), Value: []byte("bar")})
+		if err != nil {
+			t.Errorf("#%d: couldn't put key (%v)", appliedi, err)
+		}
+	}
+	// The first snapshot and compaction should happen because applied index is 11
+	logOccurredAtMostNTimes(t, mem, 5*time.Second, "saved snapshot", 1)
+	logOccurredAtMostNTimes(t, mem, time.Second, "compacted Raft logs", 1)
+	expectMemberLog(t, mem, time.Second, "\"compact-index\": 6", 1)
```
serathius, line 78: Do we really need to check if `compacted Raft log`  occured at most N times? This is hard to check, why checking if `"compact-index": X` is not enough ?

server/etcdserver/server.go
ahrtr: > Not just refactor the function.

Just refactoring the function (extract the compact into a separate method) is an independent and safe change, accordingly can be merged soon.

> The goal of the PR is to make compaction independent from snapshot

It modifies the logic/semantics, so it's no longer an independent change.

ahrtr: To be clearer about https://github.com/etcd-io/etcd/pull/18635#discussion_r1775414327, I am proposing an independent & safe minor refactoring below as the very first step

https://github.com/ahrtr/etcd/commit/efae0d29f7ffef1ef743fb11b8936f3f69497cac

serathius:
Closed with https://github.com/etcd-io/etcd/pull/18825

PR #18637, opened 2024-09-25
testutils: add ExpectExactNTimes function to LogObserver

clement2026:
Context https://github.com/etcd-io/etcd/pull/18635#discussion_r1775086281

#18635 needs this.

### DIFF
--- a/tests/framework/testutils/log_observer.go
+++ b/tests/framework/testutils/log_observer.go
@@ -46,8 +46,8 @@ func NewLogObserver(level zapcore.LevelEnabler) (zapcore.Core, *LogObserver) {
 	}
 }
 
-// Expect returns the first N lines containing the given string.
-func (logOb *LogObserver) Expect(ctx context.Context, s string, count int) ([]string, error) {
+// ExpectAtLeastNTimes returns the first N lines containing the given string.
+func (logOb *LogObserver) ExpectAtLeastNTimes(ctx context.Context, s string, count int) ([]string, error) {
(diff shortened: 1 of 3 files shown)


tests/framework/testutils/log_observer.go
```diff
@@ -46,8 +46,8 @@ func NewLogObserver(level zapcore.LevelEnabler) (zapcore.Core, *LogObserver) {
 	}
 }
 
-// Expect returns the first N lines containing the given string.
-func (logOb *LogObserver) Expect(ctx context.Context, s string, count int) ([]string, error) {
+// ExpectAtLeastNTimes returns the first N lines containing the given string.
+func (logOb *LogObserver) ExpectAtLeastNTimes(ctx context.Context, s string, count int) ([]string, error) {
```
serathius, line 50: Please update the Expect function instead. 
Current implementation of expect should be renamed to `ExpectNMatches`

tests/framework/testutils/log_observer.go
```diff
@@ -46,8 +46,8 @@ func NewLogObserver(level zapcore.LevelEnabler) (zapcore.Core, *LogObserver) {
...
 	return logOb.ExpectFunc(ctx, func(log string) bool { return strings.Contains(log, s) }, count)
```
serathius, line 51: Please update the Expect function instead. 
Current implementation of expect should be renamed to `ExpectNMatches`

serathius:
Do we really need to check if `compacted Raft log`  occured at most N times? This is hard to check, why checking if `"compact-index": X` is not enough ?

PR #18640, opened 2024-09-25
e2e test reverse proxy unused functions cleanup

henrybear327:
Part of the patches to fix https://github.com/etcd-io/etcd/issues/17737. 

This PR aims to simplify the review load of https://github.com/etcd-io/etcd/pull/18641, by separating all the code cleanup parts first.

During the development of https://github.com/etcd-io/etcd/pull/17938, we agreed that during the transition to an L7 forward proxy, unused features and features targeting the L4 reverse proxy will be dropped.

The functions dropped are
- PauseAccept
- DelayAccept
- LatencyAccept
- PauseTX and PauseRX

### DIFF
--- a/pkg/proxy/server.go
+++ b/pkg/proxy/server.go
@@ -59,11 +59,6 @@ type Server interface {
 	// Close closes listener and transport.
 	Close() error
 
-	// PauseAccept stops accepting new connections.
-	PauseAccept()
-	// UnpauseAccept removes pause operation on accepting new connections.
-	UnpauseAccept()
-
 	// DelayAccept adds latency ± random variable to accepting
 	// new incoming connections.
 	DelayAccept(latency, rv time.Duration)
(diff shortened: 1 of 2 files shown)


tests/robustness/failpoint/network.go
```diff
@@ -193,21 +193,22 @@ func (f dropPeerNetworkFailpoint) Inject(ctx context.Context, t *testing.T, lg *
...
 	proxy := member.PeerProxy()
 
-	proxy.ModifyRx(f.modifyPacket)
-	proxy.ModifyTx(f.modifyPacket)
+	if !f.shouldDropPacket() {
+		return nil, nil
+	}
+
+	proxy.BlackholeRx()
+	proxy.BlackholeTx()
```
serathius, line 201: This is incorrect change, previously it allowed to drop random packets between two members, how it either drops all traffic or doesn't do anything. Looks like MotifyRx/ModifyTx was used after all. 

I think we should leave the L4 proxy afterall.

henrybear327: Good catch @serathius! Maybe I am doing the cleanup a bit too aggressively :( 

I will roll back the changes!

henrybear327: Rollback performed!

serathius:
As mentioned in above comment, instead of gutting L4 proxy to make it L7 proxy can you instead fork it?

henrybear327:
> As mentioned in above comment, instead of gutting L4 proxy to make it L7 proxy can you instead fork it?

May I clarify what you mean by "fork it"? Do you mean that we maintain both L4 and L7 proxies in the codebase? :)

ahrtr:
Thanks @henrybear327 for the cleanup.  We created so many unused API/methods! 

The only concern is that they are exported API/Methods, it means that they may be used by other applications (which is unlikely in practice?).

Issue #18604, opened 2024-09-19
Use GitHub's CLI `gh` to publish releases


ivanvc:
### What would you like to be added?

In an effort to improve automation in the release process, I want to open the discussion of automating step 7 from the [release guide](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/release.md#release-steps). In my experience by participating in the release process, this is one of the most involved steps.

I think we could automate it by using [`gh release create`](https://cli.github.com/manual/gh_release_create). We could start by drafting a new release, eliminating the step of copying and manually editing the release text, and uploading the resulting artifacts. Then, we could keep a manual step of publishing the release while verifying that the draft looks correct.

### Why is this needed?

Work towards improving the release process. Although I would like to have way more automation, I feel it's a good compromise to start by taking small steps toward full automation.

ivanvc:
cc. @jmhbnz @ahrtr @serathius. Please let me know your thoughts.

serathius:
https://github.com/etcd-io/etcd/issues/13980 as reference for previous work in this area.

ahrtr:
Thanks for the proposal and sounds good to me. Do you have a PoC PR?

ivanvc:
> #13980 as reference for previous work in this area.

@serathius, I love the idea of having it automated after we push a tag (as I think we also discussed during one release). But I also see that as far-fetched and still needs some more investigation. As we're now part of the Kubernetes organization, there are some challenges that we've been seeing with automated jobs that need to publish stuff to GitHub.
That all to say that I'm aligned with your thinking :)

> Thanks for the proposal and sounds good to me. Do you have a PoC PR?

@ahrtr, not yet. I wanted to hear your opinions before implementing it.

jmhbnz:
Thanks for creating the feature issue @ivanvc.  I am completely supportive.  

In future we will move away from manually running scripts to having prow jobs (as GitHub Actions token permissions are too restrictive under k8s enterprise org).

In the short term further automation for the manual steps we currently perform is a great idea and will help the transition to prow jobs long term.

ivanvc:
/assign

As we're all aligned, I'll go ahead and draft a PoC PR.

bavarianbidi:
Sorry for being late on this discussion :see_no_evil: 

On `auger` i've created a PR to use `goreleaser` for creating a release (https://github.com/etcd-io/auger/pull/136).
Probably goreleaser for doing a release might be also an alternative?

[...]

PR #18649, opened 2024-09-27
release: use GitHub's gh to create GitHub release

ivanvc:
Implements #18604.

While we verify that this works as expected, it will create the release as a draft. Then, we can verify that it looks correct and manually publish them.

I tested this feature in my fork. Refer to the latest [release](https://github.com/ivanvc/etcd/releases/tag/untagged-2fbc011075d562e8f480) (and its assets).

I introduced a `scripts/release_notes.tpl.txt`. Because I don't have access to edit a previous release and copy the Markdown, I manually crafted the Markdown based on what I saw and improved how we reference the version for the Docker instructions. But, because it was manual, I may have an error.

While working on importing the release notes, I realized that we had been copy-pasting the release text with a broken link.

> For installation guides, please check out play.etcd.io and operating etcd. Latest support status for common architectures and operating systems can be found at supported platforms.

* "operating etcd" points to https://github.com/etcd-io/etcd/tree/master/Documentation#operating-etcd-clusters. I updated it to https://etcd.io/docs/v3.6/op-guide
* "supported platforms" points to https://github.com/etcd-io/website/blob/main/content/en/docs/v3.5/op-guide/supported-platform.md. Updated to https://etcd.io/docs/v3.6/op-guide/supported-platform

### DIFF
--- a/Documentation/contributor-guide/release.md
+++ b/Documentation/contributor-guide/release.md
@@ -62,6 +62,8 @@ which don't need to be executed before releasing each version.
 4. Authenticate the image registry, refer to [Authentication methods](https://cloud.google.com/container-registry/docs/advanced-authentication).
    - `gcloud auth login`
    - `gcloud auth configure-docker`
+5. Install gh, refer to [GitHub's documentation](https://github.com/cli/cli#installation). Ensure that running
+   `gh auth login` succeeds for the GitHub account you use to contribute to etcd.
 
 ### Release steps
 
@@ -85,13 +87,8 @@ which don't need to be executed before releasing each version.
    It generates all release binaries under the directory `/tmp/etcd-release-${VERSION}/etcd/release/` and images. Binaries are pushed to the Google Cloud bucket
    under project `etcd-development`, and images are pushed to `quay.io` and `gcr.io`.
 7. Publish the release page on GitHub
-   - Set the release title as the version name
-   - Choose the correct release tag (generated from step #4)
-   - Follow the format of previous release pages
-   - Attach the generated binaries and signature file
-   - Verify the historical binary size for each architecture. If there's a big difference, verify that it works for that architecture
-   - Select whether it's a pre-release
-   - Publish the release
+   - Open the **draft** release URL shown by the release script
+   - Review that it looks correct, then publish the release
 8. Announce to the etcd-dev googlegroup
 
    Follow the format of previous release emails sent to etcd-dev@googlegroups.com, see an example below. After sending out the email, ask one of the mailing list maintainers to approve the email from the pending list. Additionally, label the release email as `Release`.
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -33,15 +33,15 @@ help() {
   echo "WARNING: This does not perform the 'Add API capabilities', 'Performance testing' "
   echo "         or 'Documentation' steps. These steps must be performed manually BEFORE running this tool."
   echo ""
-  echo "WARNING: This script does not sign releases, publish releases to github or sent announcement"
-  echo "         emails. These steps must be performed manually AFTER running this tool."
+  echo "WARNING: This script does not send announcement emails. This step must be performed manually AFTER running this tool."
   echo ""
   echo "  args:"
   echo "    version: version of etcd to release, e.g. 'v3.2.18'"
   echo "  flags:"
-  echo "    --no-upload: skip gs://etcd binary artifact uploads."
-  echo "    --no-docker-push: skip docker image pushes."
   echo "    --in-place: build binaries using current branch."
+  echo "    --no-docker-push: skip docker image pushes."
+  echo "    --no-gh-release: skip creating the GitHub release using gh."
+  echo "    --no-upload: skip gs://etcd binary artifact uploads."
   echo ""
   echo "One can perform a (dry-run) test release from any (uncommitted) branch using:"
   echo "  DRY_RUN=true REPOSITORY=\`pwd\` BRANCH='local-branch-name' ./scripts/release 3.5.0-foobar.2"
@@ -117,6 +117,21 @@ main() {
   if [[ "${current_go_version}" != "${go_version}" ]]; then
     log_error "Current go version is ${current_go_version}, but etcd ${RELEASE_VERSION} requires ${go_version} (see .go-version)."
     exit 1
+  fi
+
+  if [ "${NO_GH_RELEASE}" == 1 ]; then
+    log_callout "Skipping gh verification, --no-gh-release is set"
+  else
+    # Check that gh is installed and logged in.
+    log_callout "Check gh installation"
+    if ! command -v gh >/dev/null; then
+      log_error "Cannot find gh. Please follow the installation instructions at https://github.com/cli/cli#installation"
+      exit 1
+    fi
+    if ! gh auth status &>/dev/null; then
+      log_error "GitHub authentication failed for gh. Please run gh auth login."
+      exit 1
+    fi
   fi
 
   # If the release tag does not already exist remotely, create it.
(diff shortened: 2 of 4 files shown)


ivanvc:
@ahrtr, @jmhbnz, @serathius, PTAL :)

Documentation/contributor-guide/release.md
```diff
@@ -62,6 +62,7 @@ which don't need to be executed before releasing each version.
 4. Authenticate the image registry, refer to [Authentication methods](https://cloud.google.com/container-registry/docs/advanced-authentication).
    - `gcloud auth login`
    - `gcloud auth configure-docker`
+5. Install gh, refer to [GitHub's documentation](https://github.com/cli/cli#installation). Ensure that running `gh auth login` succeeds.
```
jmhbnz, line 65: Might be over the top  but some people have a work GitHub account, let's prompt them to think about which account they are logging in with.

```suggestion
5. Install gh, refer to [GitHub's documentation](https://github.com/cli/cli#installation). Ensure that running `gh auth login` succeeds for the GitHub account you use to contribute to etcd.
```

scripts/release.sh
```diff
@@ -119,6 +118,23 @@ main() {
...
+  # Check that gh is installed and logged in.
+  log_callout "Check gh installation"
+  if ! command -v gh >/dev/null; then
+    log_error "Cannot find gh. Please follow the installation instructions at https://github.com/cli/cli#installation"
+    exit 1
+  fi
+  local gh_version
+  gh_version=$(gh --version | head -n1 | awk 'match($0,/([0-9]+\.){2}[0-9]+/) {print substr($0,RSTART,RLENGTH)}')
+  if [ "$(cut -d. -f1 <(echo "${gh_version}"))" -lt "2" ] || [ "$(cut -d. -f2 <(echo "${gh_version}"))" -lt "57" ]; then
+    log_error "The minimum version required for gh is 2.57.0, got ${gh_version}"
```
jmhbnz, line 130: Curious - Why do we have a dependency on this version? Is it purely from a standpoint of what we have tested?

scripts/release.sh
```diff
@@ -119,6 +118,23 @@ main() {
...
+    exit 1
+  fi
+  if [ "${DRY_RUN}" != "true" ] && ! gh auth status &>/dev/null; then
```
jmhbnz, line 133: Should we validate `gh auth status` irrespective of `DRY_RUN`?  If I was using `DRY_RUN=true` I would still want to know that I had a problem with `gh`.

scripts/release.sh
```diff
@@ -314,9 +330,41 @@ main() {
...
+  log_callout "Creating GitHub release for ${RELEASE_VERSION} on ${gh_repo}"
+
+  # Disable sellcheck SC2016, the single quoted syntax for sed is intentional.
+  # shellcheck disable=SC2016
+  local release_notes_temp_file=$(mktemp)
+  sed 's/${RELEASE_VERSION}/'"${RELEASE_VERSION}"'/' ./scripts/release_notes.tpl.txt > "${release_notes_temp_file}"
+  maybe_run gh release create "${RELEASE_VERSION}" \
+      --repo "${gh_repo}" \
+      --draft \
+      --prerelease \
```
jmhbnz, line 350: Do we want to be marking as pre-release?  I think this would depend on which branch we were creating a release for?  I.e. if we were publishing an alpha/rc for 3.6.0 vs publishing a patch release for 3.5 or 3.4.

scripts/release_notes.tpl.txt
```diff
@@ -0,0 +1,91 @@
...
+  --logger zap \
+  --log-outputs stderr
+
+docker exec etcd-gcr-${ETCD_VER}/usr/local/bin/etcd --version
+docker exec etcd-gcr-${ETCD_VER}/usr/local/bin/etcdctl version
+docker exec etcd-gcr-${ETCD_VER}/usr/local/bin/etcdutl version
+docker exec etcd-gcr-${ETCD_VER}/usr/local/bin/etcdctl endpoint health
+docker exec etcd-gcr-${ETCD_VER}/usr/local/bin/etcdctl put foo bar
+docker exec etcd-gcr-${ETCD_VER}/usr/local/bin/etcdctl get foo
+```
```
jmhbnz, line 91: Here is the markdown for a recent release if you wanted to diff it:
[release.txt](https://github.com/user-attachments/files/17174255/release.txt)

jmhbnz (commented):
Great work @ivanvc - A few questions but this is an awesome start.

Documentation/contributor-guide/release.md
ivanvc: Sounds good. Thanks, James.

scripts/release.sh
ivanvc: I was having issues with `gh auth status`'s exit code (the exit code was 0 even when it failed). I swear I saw a pull request/a release note for this behavior in 2.57.0 (which is also the latest version). But I just tested again with multiple versions and went back all the way to 2.37.0, and it works as expected.

So, I'm fine removing this check.

scripts/release.sh
ivanvc: I avoided the check because it would fail when running on the CI. One option would be to add a new execution flag (i.e., `--no-github-release`) so we can set that in the release test and still keep it green. WDYT?

scripts/release.sh
ivanvc: For `main`, it should be `--prerelease`, for `release-3.5`, it should be `--latest` and for `release-3.4`, it should not have an additional option.

I'm open to documenting it or adding a branch check. Let me know your thoughts.

scripts/release_notes.tpl.txt
ivanvc: Thanks, it looks the same. I guess, maybe it would be a good idea to update the second paragraph (at least in the two latest release [3.5.16 and 3.4.34]) to the following (to address the broken links):

```md
For installation guides, please check out [play.etcd.io](http://play.etcd.io) and [operating etcd](https://etcd.io/docs/v3.6/op-guide). Latest support status for common architectures and operating systems can be found at [supported platforms](https://github.com/etcd-io/website/blob/main/content/en/docs/v3.5/op-guide/supported-platform.md).
```

scripts/release_notes.tpl.txt
```diff
@@ -0,0 +1,91 @@
+Please check out [CHANGELOG](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.5.md) for a full list of changes. And make sure to read [upgrade guide](https://github.com/etcd-io/website/blob/main/content/en/docs/v3.6/upgrades/upgrade_3_6.md) before upgrading etcd (there may be breaking changes).
```
serathius, line 1: Note; we wanted to simplify the release notes. https://github.com/etcd-io/etcd/issues/15185

ivanvc: I agree with that issue/improvement, but I'd like to avoid scope creeping this pull request and implementing it by only creating the release with `gh`. I think it would be a good idea to create an umbrella issue for the improvements we want to make to the release process. I'll note this and create it when I close some of the tasks I currently have open.

scripts/release.sh
jmhbnz: Makes sense, having a new flag we can use in CI sounds good to me.

scripts/release.sh
jmhbnz: I think we need to have some conditional logic in the script to provide the appropriate parameter based on which branch we are performing release steps for.  Otherwise we would need to edit the instructions to make sure we update draft release manually?

ivanvc: Sounds like a good idea, I'll implement it.

scripts/release.sh
```diff
@@ -314,10 +329,59 @@ main() {
...
+        --repo "${gh_repo}" \
+        --draft \
+        --title "${RELEASE_VERSION}" \
+        --notes-file "${release_notes_temp_file}" \
+        "${gh_release_args[@]}" \
+        ./release/SHA256SUMS
+
+    # Upload files one by one, as gh doesn't support passing globs as input.
+    maybe_run find ./release '(' -name '*.tar.gz' -o -name '*.zip' ')' -exec \
+      gh --repo "${gh_repo}" release upload "${RELEASE_VERSION}" {} \;
```
ahrtr, line 376: We also need to upload `SHA256SUMS`

scripts/release_notes.tpl.txt
ahrtr: We can do it separately. This PR only focuses on automating the creating github release.

Should we create separate templates for each release (including 3.5.x, 3.4.x and 3.6.x)? @ivanvc Since this PR is for the main branch, we can resolve it in followup PR if you want.

ivanvc: Ah, good point. I didn't realize that it has mixed links from 3.5 and 3.6 in the release notes. I'll work on it in this PR.

scripts/release.sh
ivanvc: This is uploaded in line https://github.com/etcd-io/etcd/pull/18649/files#diff-1bba00e5aca52286a667c09eff92ba2ca8e3ca77e0d31b0599cc829ab0173389R372.

However, your comment made me realize that these steps I wrote are not stateless, and if we re-run the script after a failure, it won't be able to recover/continue drafting the GitHub release. I'll address that issue in this PR.

ivanvc: Done

ivanvc:
I updated the script. It's a bit difficult to test it out locally. We could bring more improvements later on.

A sample release is the following: https://github.com/ivanvc/etcd/releases/tag/untagged-1dc492de524109e605b4

scripts/release.sh
```diff
@@ -119,6 +119,21 @@ main() {
     exit 1
   fi
 
+  if [ "${NO_GH_RELEASE}" == 1 ]; then
+    log_callout "Skipping gh verification, --no-gh-release is set"
```
ahrtr, line 123:
```suggestion
  if [ "${DRY_RUN}" == "true" ] || [ "${NO_GH_RELEASE}" == 1 ];
    log_callout "Skipping gh verification, --no-gh-release or DRY_RUN=true is set"
```

scripts/release.sh
```diff
@@ -314,10 +329,70 @@ main() {
...
+          --repo "${gh_repo}" \
+          --draft \
+          --title "${RELEASE_VERSION}" \
+          --notes-file "${release_notes_temp_file}" \
+          "${gh_release_args[@]}"
+    fi
+
+    # Upload files one by one, as gh doesn't support passing globs as input.
+    maybe_run find ./release '(' -name '*.tar.gz' -o -name '*.zip' ')' -exec \
+      gh --repo "${gh_repo}" release upload "${RELEASE_VERSION}" {} \;
```
ahrtr, line 386: Just to double check, is this command idempotent? If we run this command multiple times, will it just replace the files?

scripts/release.sh
```diff
@@ -314,10 +329,70 @@ main() {
...
+    local gh_release_args=()
+
+    case "${BRANCH}" in
+      "main")
+        gh_release_args=(--prerelease)
+        ;;
+      "release-3.5")
+        gh_release_args=(--latest)
+        ;;
+    esac
```
ahrtr, line 350: 3.4.x isn't supported yet?

scripts/release.sh
ivanvc: Hi @ahrtr, this is intentional, as suggested by @jmhbnz (ref: https://github.com/etcd-io/etcd/pull/18649#discussion_r1779470541). I initially had it the way you suggested, but I agree with James that verifying the auth status would be reasonable when doing it in DRY mode.

scripts/release.sh
ivanvc: Thanks for the call out. I just tested and need to add an argument to allow overwriting the file. I'll update the PR shortly.

scripts/release.sh
ivanvc: If the branch is:
* `main`: we want to mark it as a prerelease
* `release-3.5`: we want to mark it as latest
* `release-3.4`: we want to leave it as is (therefore, it doesn't need an additional argument)

Let me document it to ensure everything is clear.

ahrtr: ack

scripts/release.sh
ahrtr: OK, thx for the clarification.

scripts/release.sh
ivanvc: Done in 88d2d2472c32e4cfa0e7223fa0b173865f0ddf3c

scripts/release.sh
ivanvc: Done in 88d2d2472c32e4cfa0e7223fa0b173865f0ddf3c

ahrtr (approved):
LGTM

Thanks @ivanvc for improving the release work. Nice work!

scripts/release.sh
```diff
@@ -314,10 +329,70 @@ main() {
...
+    local release_notes_temp_file
+    local release_url
+    local gh_release_args=()
+
+    # For the main branch (v3.6), we should mark the release as a prerelease.
+    # The release-3.5 (v3.5) branch, should be marked as latest. And release-3.4 (v3.4)
+    # should be left without any additional mark (therefore, it doesn't need a special argument).
+    if [ "${BRANCH}" = "main" ]; then
+      gh_release_args=(--prerelease)
+    elif [ "${BRANCH}" = "release-3.5" ]; then
```
serathius, line 348: What happens when we release 3.6, this will not update? Maybe we should assert that branch name and fail that if new branch is cut we need to update which is the latest?

scripts/release.sh
```diff
@@ -314,10 +329,70 @@ main() {
...
+      gh_release_args=(--latest)
+    fi
+
+    if [ "${REPOSITORY}" = "$(pwd)" ]; then
+      gh_repo=$(git remote get-url origin)
```
serathius, line 353: This assumes that people cloned etcd main repo. If they cloned fork the `origin` would point to it. Is there any reason we cannot just skip it?

scripts/release.sh
```diff
@@ -314,10 +329,70 @@ main() {
...
+    else
+      gh_repo="${REPOSITORY}"
+    fi
+
+    gh_repo=$(echo "${gh_repo}" | sed 's/^[^@]\+@//' | sed 's/https\?:\/\///' | sed 's/\.git$//' | tr ':' '/')
+    log_callout "Creating GitHub release for ${RELEASE_VERSION} on ${gh_repo}"
```
serathius, line 359: Do we really want to support pushing to other repositories?

serathius (approved):
Couple of nits

scripts/release.sh
ivanvc: Re: this comment and https://github.com/etcd-io/etcd/pull/18649#discussion_r1814589059. I understand that we may want to drop supporting other repositories to simplify the process. However, the only reason I'd like to keep it, is that it's the only way to test locally without doing a release in etcd-io's repository (i.e., https://github.com/ivanvc/etcd/releases).

WDYT @jmhbnz, @ahrtr?

jmhbnz: I like the ability to test in forks, let's keep it.

ahrtr: > I like the ability to test in forks, let's keep it imo.

+1

Also it's just creating a draft release, and isn't published yet.  It needs the release person to manually publish it.

Issue #18412, opened 2024-08-06
Detected disallowed custom content in v2store


ahrtr:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

The e2e test case `TestV2DeprecationWriteOnlySnapshot` failed on `main` branch, see

```
[...]

PR #18664, opened 2024-10-01
fix(tests/e2e): ensure spawned cmds are closed

ghouscht:
This should fix https://github.com/etcd-io/etcd/issues/18412.

After the test case is done Go's testing pkg tries to cleanup the temporary directories that were created by calls to `TempDir()`. This cleanup sometimes failed with errors like:

```
testing.go:1231: TempDir RemoveAll cleanup: unlinkat /tmp/TestV2DeprecationWriteOnlySnapshot1900751190/002/member-0/member/snap: directory not empty
```

I belive the cleanup sometimes happened before the etcd process was stopped which lead to the above error. Adding the defer statement to make sure the process is done before the test case returns seems to fix the error for me.

Before my changes I was able to reliably reproduce the issue with
```
go test -v -count 1000 -run "^TestV2DeprecationWriteOnlySnapshot$"
```
and with the changes from this PR I'm no longer able to reproduce this.

### DIFF
--- a/tests/e2e/v2store_deprecation_test.go
+++ b/tests/e2e/v2store_deprecation_test.go
@@ -111,6 +111,7 @@ func TestV2DeprecationWriteOnlySnapshot(t *testing.T) {
 	t.Log("Verify its infeasible to start etcd with --v2-deprecation=write-only mode")
 	proc, err := e2e.SpawnCmd([]string{e2e.BinPath.Etcd, "--v2-deprecation=write-only", "--data-dir=" + memberDataDir}, nil)
 	assert.NoError(t, err)
+	defer proc.Close()
 
 	_, err = proc.Expect("detected disallowed custom content in v2store for stage --v2-deprecation=write-only")
 	assert.NoError(t, err)


ahrtr (approved):
LGTM

Thank you @ghouscht

ahrtr:
@ghouscht could you please also backport the fix to 3.5 and 3.4 if needed?

ivanvc:
Quick (and maybe naive) question. As I was checking if we needed to backport, I realized that the test doesn't exist for the `release-3.5` branch. However, I also noticed that we're also calling an etcd spawn in other places in the tests. i.e., refer to: https://github.com/etcd-io/etcd/blob/ddf0ac2dc270d6ef78bf1c4069279522069ae152/tests/e2e/v2store_deprecation_test.go#L82

Do we need to close the command in those instances, too? Or is it only relevant to this test because we're creating a snapshot?

ghouscht:
> Quick (and maybe naive) question. As I was checking if we needed to backport, I realized that the test doesn't exist for the `release-3.5` branch. However, I also noticed that we're also calling an etcd spawn in other places in the tests. i.e., refer to:
> 
> https://github.com/etcd-io/etcd/blob/ddf0ac2dc270d6ef78bf1c4069279522069ae152/tests/e2e/v2store_deprecation_test.go#L82
> 
> Do we need to close the command in those instances, too? Or is it only relevant to this test because we're creating a snapshot?

It seems to not matter for the other tests in this file. Most likely because they seem to test certain flag values that aren't supported so nothing is written to the filesystem. But as a rule of thumb I'd say whenever `SpawnCmd` is called also `Close` should be called. If it doesn't help it also doesn't hurt, no? I can open another PR for that if you want me to.

ahrtr:
> refer to:
> 
> https://github.com/etcd-io/etcd/blob/ddf0ac2dc270d6ef78bf1c4069279522069ae152/tests/e2e/v2store_deprecation_test.go#L82
> 
> Do we need to close the command in those instances, too?

In that case, etcdserver will [panic](https://github.com/etcd-io/etcd/blob/2c971109267d9d29ef7cb901148931bdbd55762a/server/etcdserver/apply_v2.go#L30) and exit automatically. But It will not do any harm to call `proc.Close()`. Please feel free to raise a separate PR to fix it if anyone wants, thx

ivanvc:
> In that case, etcdserver will [panic](https://github.com/etcd-io/etcd/blob/2c971109267d9d29ef7cb901148931bdbd55762a/server/etcdserver/apply_v2.go#L30) and exit automatically. But It will not do any harm to call `proc.Close()`. Please feel free to raise a separate PR to fix it if anyone wants, thx

Thanks for the explanation. I wasn't sure if that was required. I'm fine whether we address it in a follow-up or not (as it shouldn't cause an issue).

As for `release-3.4` and `release-3.5`, I think we don't have that test case. But feel free to double-check, @ghouscht.

ivanvc (approved):
LGTM. Thanks so much for your work toward addressing flaky tests, Thomas.

Issue #18667, opened 2024-10-02
Specifying a revision for a range request in a transaction may cause data inconsistency


ahrtr:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Specifying a revision for a range request in a transaction may cause data inconsistency. 

The root cause is that etcd server removes range requests from the TXN for endpoints the client isn’t connected to. 

https://github.com/etcd-io/etcd/blob/2c971109267d9d29ef7cb901148931bdbd55762a/server/etcdserver/server.go#L1967-L1971

For example, if the client connects to member 1, then etcdserver removes the range request (`get k1 --rev=10` in this example) from the TXN in member 2 and 3. Accordingly, member 1 applies failed due to checkRange's failures, but member 2 & 3 apply the TXN successfully because the range request was removed. Eventually it leads to the situation that different members have different data.

https://github.com/etcd-io/etcd/blob/2c971109267d9d29ef7cb901148931bdbd55762a/server/etcdserver/txn/txn.go#L431-L441

### What did you expect to happen?

etcd should never run into data inconsistency.

### How can we reproduce it (as minimally and precisely as possible)?

- Step 1:  start a brand new 3 member cluster
- Step 2: Execute `for i in {1..20}; do ./etcdctl put k$i v$i; done`
- Step 3: Execute `./etcdctl compact 21`
- Step 4: Execute `./etcdctl txn --interactive`
```
compares:
value("k1") = "v1"

success requests (get, put, del):
put k2 foo
get k1 --rev=10

failure requests (get, put, del):
```

The client will get a `**etcdserver: mvcc: required revision has been compacted**` error.

- Step 5: Execute `./etcdctl get k2` against different endpoints, you will get different values,
```
$ ./etcdctl --endpoints=127.0.0.1:2379 get k2
k2
v2
$ ./etcdctl --endpoints=127.0.0.1:22379 get k2
k2
foo
[...]

PR #18674, opened 2024-10-03
Add a caution note about making apply context bounded

shyamjvs:
Following up from https://github.com/etcd-io/etcd/issues/18667#issuecomment-2392286839.

/cc @ahrtr @serathius

### DIFF
--- a/server/etcdserver/apply/uber_applier.go
+++ b/server/etcdserver/apply/uber_applier.go
@@ -115,6 +115,11 @@ func (a *uberApplier) Apply(r *pb.InternalRaftRequest) *Result {
 	// then dispatch() unpacks the request to a specific method (like Put),
 	// that gets executed down the hierarchy again:
 	// i.e. CorruptApplier.Put(CappedApplier.Put(...(BackendApplier.Put(...)))).
+	//
+	// CAUTION: The context below should NOT be changed to a bounded value without
+	// first addressing the risk of a transaction's operations being only partially
+	// applied when some operations timeout. More details here:
+	// https://github.com/etcd-io/etcd/issues/18667#issuecomment-2392286839
 	return a.applyV3.Apply(context.TODO(), r, a.dispatch)
 }
 


server/etcdserver/apply/uber_applier.go
```diff
@@ -115,6 +115,11 @@ func (a *uberApplier) Apply(r *pb.InternalRaftRequest) *Result {
 	// then dispatch() unpacks the request to a specific method (like Put),
 	// that gets executed down the hierarchy again:
 	// i.e. CorruptApplier.Put(CappedApplier.Put(...(BackendApplier.Put(...)))).
+	//
+	// CAUTION: The context below should NOT be changed to a bounded value without
+	// first addressing the risk of a transaction's operations being only partially
+	// applied when some operations timeout. More details here:
+	// https://github.com/etcd-io/etcd/issues/18667#issuecomment-2392286839
```
ahrtr, line 122:
```suggestion
	//
	// CAUTION: Do NOT change the context below to have a timeout (i.e., bounded value).
	// The apply workflow may be intentionally interrupted for expected reasons, but it
	// should never fail due to non-deterministic factors such as a context timeout. If
	// the workflow is interrupted by such factors, it can lead to a scenario where some
	// members apply changes successfully while others fail, potentially causing data
	// inconsistency issues like https://github.com/etcd-io/etcd/issues/18667.
	// See also https://github.com/etcd-io/etcd/issues/18667#issuecomment-2392286839.		
```

ahrtr:
Please also signoff the commit, please read https://github.com/etcd-io/etcd/pull/18674/checks?check_run_id=31053782708

server/etcdserver/apply/uber_applier.go
serathius: I don't like the idea that adding a comment is meant to protect us from disaster. Context timeout can be assigned at any lower level of the apply loop where is comment is not present, with same consequences.

Would prefer a slow methodical removal of context. First PR can just remove ctx from `Apply` method that we know uses context.TODO().

ahrtr: > I don't like the idea that adding a comment is meant to protect us from disaster. Context timeout can be assigned at any lower level of the apply loop where is comment is not present, with same consequences.

It's just an easy & safe & temporary improvement for the existing situation. Overall not a big deal to me, so doesn't deserve too much discussing this. Either quickly approve & merge this PR or just reject it.



> First PR can just remove ctx from `Apply` method that we know uses context.TODO().

I'd suggest to evaluate the effort & impact. Afterwards, we can breakdown it into PRs.

serathius: > I'd suggest to evaluate the effort & impact. Afterwards, we can breakdown it into PRs.

We already did in https://github.com/etcd-io/etcd/issues/18667#issuecomment-2393130782, as I mentioned that context is just used in 2 ways.

ahrtr: > We already did in [#18667 (comment)](https://github.com/etcd-io/etcd/issues/18667#issuecomment-2393130782), as I mentioned that context is just used in 2 ways, tracing and authorization metadata.

I don't think a simple comment is enough. I expect a doc or summary to clarify why (of course it's already clear) and "how" you will resolve it, and the "impact" on the etcdserver. It would be perfect if we can have PoC PRs.

PR #18675, opened 2024-10-04
Remove context from top level apply

serathius:
cc @ahrtr 

https://github.com/etcd-io/etcd/issues/18667#issuecomment-2393130782

### DIFF
--- a/server/etcdserver/apply/apply.go
+++ b/server/etcdserver/apply/apply.go
@@ -68,7 +68,7 @@ type applyFunc func(ctx context.Context, r *pb.InternalRaftRequest) *Result
 type applierV3 interface {
 	// Apply executes the generic portion of application logic for the current applier, but
 	// delegates the actual execution to the applyFunc method.
-	Apply(ctx context.Context, r *pb.InternalRaftRequest, applyFunc applyFunc) *Result
+	Apply(r *pb.InternalRaftRequest, applyFunc applyFunc) *Result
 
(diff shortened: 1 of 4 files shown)


server/etcdserver/apply/apply.go
```diff
@@ -146,8 +146,8 @@ func newApplierV3Backend(
 		txnModeWriteWithSharedBuffer: txnModeWriteWithSharedBuffer}
 }
 
-func (a *applierV3backend) Apply(ctx context.Context, r *pb.InternalRaftRequest, applyFunc applyFunc) *Result {
-	return applyFunc(ctx, r)
+func (a *applierV3backend) Apply(r *pb.InternalRaftRequest, applyFunc applyFunc) *Result {
+	return applyFunc(context.TODO(), r)
```
ahrtr, line 150: Suggest to add similar comment as in https://github.com/etcd-io/etcd/pull/18674 right above this line to make this PR completely independent. In the future, when anyone continue to completely get rid of the context, then we can remove the comment by then.

ahrtr (approved):
LGTM with a minor comment.

Thanks for the change which looks clean.

server/etcdserver/apply/apply.go
serathius: I have all commit ready. Just going one by one.

serathius: https://github.com/etcd-io/etcd/pull/18676 is next

PR #18678, opened 2024-10-04
Remove context from mvcc public interface

serathius:
cc @ahrtr 

https://github.com/etcd-io/etcd/issues/18667#issuecomment-2393130782

### DIFF
--- a/server/etcdserver/txn/txn_test.go
+++ b/server/etcdserver/txn/txn_test.go
@@ -15,8 +15,6 @@
 package txn
 
 import (
-	"context"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


server/etcdserver/txn/txn_test.go
```diff
@@ -304,72 +296,6 @@ func setup(t *testing.T, setup testSetup) (mvcc.KV, lease.Lessor) {
 	return s, lessor
 }
 
-func TestReadonlyTxnError(t *testing.T) {
```
serathius, line 307: Need to double check the original motivation for the test and if we need to rewrite it. Marking the PR as draft until then.

serathius: Ok, I think that enough context removal. Based on https://github.com/etcd-io/etcd/pull/14149 this Test validates that Readonly TXN pass context and don't panic on it. This is the exact place where we want to allow canceling Read transactions as it's safe and a feature to allow clients abort requests.

At some point would be good to design how context should be handled in etcd codebase. Non-cancel-able metadata for apply loop that executes writes, that can also be cancel-able for Read request. Both paths meet in the transaction code layer. Would be nice to be able to statically validate at compilation time whether transaction is read only. 

cc @ahrtr @shyamjvs

PR #18685, opened 2024-10-05
Fix robustness test logs by identifing failpoint availability on temporary cluster before creating subtest

serathius:
cc @ahrtr @fuweid @henrybear327 @siyuanfoundation @MadhavJivrajani

### DIFF
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -44,24 +44,37 @@ func TestMain(m *testing.M) {
 func TestRobustnessExploratory(t *testing.T) {
 	testRunner.BeforeTest(t)
 	for _, s := range exploratoryScenarios(t) {
-		t.Run(s.name, func(t *testing.T) {
+		ctx := context.Background()
+		s.failpoint = randomFailpointForConfig(ctx, t, s.profile, s.cluster)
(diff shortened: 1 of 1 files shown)


tests/robustness/main_test.go
```diff
@@ -44,26 +44,39 @@ func TestMain(m *testing.M) {
...
+	config.Logger = zap.NewNop()
+	c, err := e2e.NewEtcdProcessCluster(ctx, t, e2e.WithConfig(&config))
+	if err != nil {
+		t.Fatal(err)
+	}
+	f, err := failpoint.PickRandom(c, profile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = forcestopCluster(c)
```
siyuanfoundation, line 73: why not return the cluster and reuse it?

serathius: Because logs will be directed to the different log name. During cluster creation we provide the logger, to correlate logs to tests, we use `zaptest.New` which takes the `*testing.T` as argument. If we reuse the cluster they will not be correlated to logs and clusters from all tests scenarios will be mixed.

fuweid (approved):
LGTM

serathius:
This fails in CI for some reason. Need to debug more.

PR #18686, opened 2024-10-05
fix: enable bool-compare rule from testifylint

mmorel-35:
This fixes [bool-compare](https://github.com/Antonboom/testifylint?tab=readme-ov-file#bool-compare) rule from [testifylint](https://github.com/Antonboom/testifylint).

It also replaces `testutil.AssertFalse` and `testutil.AssertTrue` by `assert.False` and `assert.True` as they are identical (t.Helper() called in testify's functions)


ahrtr (approved):

jmhbnz (approved):
LGTM - Thanks @mmorel-35

jmhbnz:
> What’s left in the process before merging ?

Requesting review from @ivanvc who has been involved in many linting improvements recently, primarily for awareness.

ivanvc (approved):
This looks great. Thanks, Matthieu. I think after merging this pull request, we can:

1. Replace `testutil.AssertNil` and `testutil.AssertNotNil` with their assert package equivalent
2. Enable `nil-compare`
3. Remove the stray `client/pkg/testutil/assert.go`, as it won't be used anymore by any test cases (technically, right now, we could delete the functions `testutil.AssertTrue`/`testutil.AssertFalse`, but I'm okay with doing this later)

@jmhbnz, should we create a parent issue? I'd like to do it only if we have contributors interested in working on this.

jmhbnz:
> This looks great. Thanks, Matthieu. I think after merging this pull request, we can:
> 
>     1. Replace `testutil.AssertNil` and `testutil.AssertNotNil` with their assert package equivalent
> 
>     2. Enable `nil-compare`
> 
>     3. Remove the stray `client/pkg/testutil/assert.go`, as it won't be used anymore by any test cases (technically, right now, we could delete the functions `testutil.AssertTrue`/`testutil.AssertFalse`, but I'm okay with doing this later)
> 
> 
> @jmhbnz, should we create a parent issue? I'd like to do it only if we have contributors interested in working on this.

1 and 2 sound like reasonable next steps. An issue would be a good way to check if there is consensus for 3.  Let's go ahead and merge this initial improvement.

ivanvc:
Thanks, @jmhbnz. I missed your last reply. I think something reasonable would be to deprecate and open a pull request regarding 3. There, we can see if we have a consensus (x-posting https://github.com/etcd-io/etcd/pull/18689#issuecomment-2397656875)

ivanvc:
Link to #18719

PR #18689, opened 2024-10-07
fix: enable nil-compare rule from testifylint

mmorel-35:
This fixes [nil-compare](https://github.com/Antonboom/testifylint?tab=readme-ov-file#nil-compare) rule from [testifylint](https://github.com/Antonboom/testifylint).

It also replaces `testutil.AssertNil` and `testutil.AssertNotNil` by `assert.Nil` and `require.NotNil`

The require package is used for NotNil replacement because of the Fatal method being used in testutil version of the method.

### DIFF
--- a/pkg/ioutil/pagewriter_test.go
+++ b/pkg/ioutil/pagewriter_test.go
@@ -146,7 +146,7 @@ func TestPageWriterPageBytes(t *testing.T) {
 				}, "expected panic when pageBytes is %d", tc.pageBytes)
 			} else {
 				pw := NewPageWriter(cw, tc.pageBytes, 0)
-				assert.NotEqual(t, pw, nil)
+				assert.NotNil(t, pw)
 			}
 		})
 	}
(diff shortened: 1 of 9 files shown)


ahrtr (approved):

ahrtr:
Thank you!

ivanvc (approved):
LGTM. Thanks, Matthieu. Can you delete `client/pkg/testutil/assert.go` in a follow-up pull request?

ivanvc:
> Are you sure that's not used by any project that depends on etcd.

Good point. Let's deprecate.

ivanvc:
> OK, I updated the comments on the `client/pkg/testutil/assert.go` functions

Sorry, @mmorel-35. I think it'd be better in another PR, to open it up for discussion.

ivanvc:
> Alright, it's reverted

Thank you so much.

/retest

ivanvc:
@jmhbnz, Matthieu removed the deprecation notice from `client/pkg/testutil/assert.go`. The pull request is now back in the state it was in when Benjamin and I approved it. Could you PTAL at it? Thanks.

ivanvc:
Link to https://github.com/etcd-io/etcd/issues/18719

PR #18711, opened 2024-10-09
server,tests: add additional lease metrics and test

vivekpatani:
- metrics to capture leases attached and detached
- metrics to capture duration to grant, revoke, and renew leases
- metric to capture initial lease count at startup

**Help**
- Primarily I'd like to get feedback, I think these metrics can be useful, especially under heavy load.
- Need help with the last part of the testing, where I try to capture the count the count at initial startup, please let me know if my understanding is wanting in this case. More specifically - [this](https://github.com/etcd-io/etcd/pull/18711/files#diff-2d7b95e0dbdfa5eee525091e31daea409a4d08800053c3554d41daf9ffe0a81fR170-R193).
  - Currently that part of the test does not pass.
  - My understanding is that when the cluster recovers the lease should exist in the database and should reflect in terms of metric and `LeaseLeases` response.

### DIFF
--- a/server/lease/lessor.go
+++ b/server/lease/lessor.go
@@ -280,10 +280,12 @@ func (le *lessor) SetCheckpointer(cp Checkpointer) {
 
 func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
 	if id == NoLease {
+		leaseGrantError.WithLabelValues(ErrLeaseNotFound.Error()).Inc()
 		return nil, ErrLeaseNotFound
 	}
 
 	if ttl > MaxLeaseTTL {
+		leaseGrantError.WithLabelValues(ErrLeaseExists.Error()).Inc()
 		return nil, ErrLeaseTTLTooLarge
 	}
 
@@ -329,6 +331,7 @@ func (le *lessor) Revoke(id LeaseID) error {
 	l := le.leaseMap[id]
 	if l == nil {
 		le.mu.Unlock()
+		leaseRevokeError.WithLabelValues(ErrLeaseNotFound.Error()).Inc()
 		return ErrLeaseNotFound
 	}
 
@@ -418,12 +421,15 @@ func (le *lessor) Renew(id LeaseID) (int64, error) {
 		// quorum to be revoked. To be accurate, renew request must wait for the
 		// deletion to complete.
 		case <-l.revokec:
+			leaseRenewError.WithLabelValues(ErrLeaseNotFound.Error()).Inc()
 			return -1, ErrLeaseNotFound
 		// The expired lease might fail to be revoked if the primary changes.
 		// The caller will retry on ErrNotPrimary.
 		case <-demotec:
+			leaseRenewError.WithLabelValues(ErrNotPrimary.Error()).Inc()
 			return -1, ErrNotPrimary
 		case <-le.stopC:
+			leaseRenewError.WithLabelValues(ErrNotPrimary.Error()).Inc()
 			return -1, ErrNotPrimary
 		}
 	}
@@ -433,6 +439,7 @@ func (le *lessor) Renew(id LeaseID) (int64, error) {
 	// of RAFT entries written per lease to a max of 2 per checkpoint interval.
 	if clearRemainingTTL {
 		if err := le.cp(context.Background(), &pb.LeaseCheckpointRequest{Checkpoints: []*pb.LeaseCheckpoint{{ID: int64(l.ID), Remaining_TTL: 0}}}); err != nil {
+			leaseRenewError.WithLabelValues(err.Error()).Inc()
 			return -1, err
 		}
 	}
@@ -555,6 +562,7 @@ func (le *lessor) Attach(id LeaseID, items []LeaseItem) error {
 
 	l.mu.Lock()
 	for _, it := range items {
+		leaseAttached.Inc()
 		l.itemSet[it] = struct{}{}
 		le.itemMap[it] = id
 	}
@@ -582,6 +590,7 @@ func (le *lessor) Detach(id LeaseID, items []LeaseItem) error {
 
 	l.mu.Lock()
 	for _, it := range items {
+		leaseDetached.Inc()
 		delete(l.itemSet, it)
 		delete(le.itemMap, it)
 	}
@@ -821,6 +830,7 @@ func (le *lessor) initAndRecover() {
 		}
 	}
 	le.leaseExpiredNotifier.Init()
+	initLeaseCount.Add(float64(len(lpbs)))
 	heap.Init(&le.leaseCheckpointHeap)
 
 	le.b.ForceCommit()
--- a/server/lease/metrics.go
+++ b/server/lease/metrics.go
@@ -49,6 +49,48 @@ var (
 			// 1 second -> 3 months
 			Buckets: prometheus.ExponentialBuckets(1, 2, 24),
 		})
+
+	leaseAttached = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "attach_total",
+		Help:      "The number of leases that are attached to a lease item.",
+	})
+
+	leaseDetached = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "detach_total",
+		Help:      "The number of leases that are detached from a lease item.",
+	})
+
+	initLeaseCount = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "initial_lease_count",
+		Help:      "Reports an initial lease count.",
+	})
+
+	leaseGrantError = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "grant_errors",
+		Help:      "Error count by type to count for lease grants.",
+	}, []string{"error"})
+
+	leaseRevokeError = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "revoke_errors",
+		Help:      "Error count by type to count for lease revokes.",
+	}, []string{"error"})
+
+	leaseRenewError = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "renew_errors",
+		Help:      "Error count by type to count for lease renewals.",
+	}, []string{"error"})
 )
 
 func init() {
@@ -56,4 +98,9 @@ func init() {
 	prometheus.MustRegister(leaseRevoked)
 	prometheus.MustRegister(leaseRenewed)
 	prometheus.MustRegister(leaseTotalTTLs)
+	prometheus.MustRegister(leaseAttached)
+	prometheus.MustRegister(leaseDetached)
+	prometheus.MustRegister(leaseGrantError)
+	prometheus.MustRegister(leaseRevokeError)
+	prometheus.MustRegister(leaseRenewError)
 }
--- a/tests/integration/v3_lease_test.go
+++ b/tests/integration/v3_lease_test.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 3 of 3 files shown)


server/lease/metrics.go
```diff
@@ -49,11 +49,88 @@ var (
...
+		Name:      "attach_total",
+		Help:      "The number of leases that are attached to a lease item.",
+	})
+
+	leaseDetached = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "detach_total",
+		Help:      "The number of leases that are detached from a lease item.",
+	})
```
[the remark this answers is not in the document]
vivekpatani: I don't have enough background about cardinality but my understanding is that, adding a key multiplies the cardinalities. But if not, then the idea sounds good. @ahrtr thoughts?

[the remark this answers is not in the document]
ahrtr: > adding a key multiplies the cardinalities

Yes, it's true. Indeed, It isn't a good idea to add label "lease_id". The prometheus official document clearly clarifies it https://prometheus.io/docs/practices/naming/#labels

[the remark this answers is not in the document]
ahrtr: So
- either keep it as it's (two metrics, leaseAttached and leaseDetached), but shouldn't them counter instead of gauge?
- or consolidate them into one as mentioned above, but without the label "lease_id". In this case, it should be Gauge for sure.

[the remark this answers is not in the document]
serathius: I have background in metrics cardinality. `label_id` is a big no-no. Please note that not everything can be expressed in metrics, they are aggregations that allow you to observe state of program. Some things, especially for debugging purposes are better expressed as events. Aka logs.

server/lease/lessor.go
```diff
@@ -279,11 +279,16 @@ func (le *lessor) SetCheckpointer(cp Checkpointer) {
 }
 
 func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
+	defer func(start time.Time) {
+		leaseGrantDurationSeconds.Observe(time.Since(start).Seconds())
```
serathius, line 283: How this differs than grpc latency metrics for "leaseGrant" request?

server/lease/lessor.go
```diff
@@ -324,11 +329,15 @@ func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
 }
 
 func (le *lessor) Revoke(id LeaseID) error {
+	defer func(start time.Time) {
+		leaseRevokeDurationSeconds.Observe(time.Since(start).Seconds())
```
serathius, line 333: How this differs than grpc latency metrics for "leaseRevoke" request?

server/lease/lessor.go
```diff
@@ -394,6 +403,9 @@ func greaterOrEqual(first, second semver.Version) bool {
 // Renew renews an existing lease. If the given lease does not exist or
 // has expired, an error will be returned.
 func (le *lessor) Renew(id LeaseID) (int64, error) {
+	defer func(start time.Time) {
+		leaseRenewDurationSeconds.Observe(time.Since(start).Seconds())
```
serathius, line 407: How this differs than grpc latency metrics for "leaseRenew" request?

server/lease/metrics.go
```diff
@@ -49,11 +49,88 @@ var (
 			// 1 second -> 3 months
 			Buckets: prometheus.ExponentialBuckets(1, 2, 24),
 		})
+
+	leaseAttached = prometheus.NewCounter(prometheus.CounterOpts{
```
serathius, line 53: What's the purpose of metrics about lease attachements? What you are trying to measure? number of keys per lease?

server/lease/lessor.go
vivekpatani: Ack, don't need this in favor of `grpc_server_handling_seconds_bucket{grpc_method="LeaseGrant"...}`. Will update PR.

server/lease/lessor.go
vivekpatani: Ack, don't need this in favor of `grpc_server_handling_seconds_bucket{grpc_method="LeaseRevoke"...}`. Will update PR.

server/lease/lessor.go
vivekpatani: Ack, don't need this in favor of `grpc_server_handling_seconds_bucket{grpc_method="LeaseRenew"...}`. Will update PR.

server/lease/metrics.go
vivekpatani: > number of keys per lease?

Precisely, yes.

server/lease/metrics.go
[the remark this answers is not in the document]
vivekpatani: > or consolidate them into one as mentioned above, but without the label "lease_id". In this case, it should be Gauge for sure.

This seems like a good idea.

> Some things, especially for debugging purposes are better expressed as events. So write logs

Logs are great but sometimes it's difficult to miss a trend, when there's log. Makes it easier to setup a trigger based on logs, than the trend of metrics, if that makes sense.

server/lease/metrics.go
vivekpatani: @serathius ping^

server/lease/metrics.go
```diff
@@ -49,11 +49,58 @@ var (
...
+
+	leaseDetached = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "detach_total",
+		Help:      "The number of leases that are detached from a lease item.",
+	})
+
+	initLeaseCount = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
```
serathius, line 68: Please add experimental prefix for new metrics, we don't want to give users notion that those metrics are stable.

server/lease/metrics.go
```diff
@@ -49,11 +49,58 @@ var (
...
+		Subsystem: "lease",
+		Name:      "initial_lease_count",
+		Help:      "Reports an initial lease count.",
+	})
+
+	leaseGrantError = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
serathius, line 74: This should be covered by RPC metrics

server/lease/metrics.go
```diff
@@ -49,11 +49,58 @@ var (
...
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "grant_errors",
+		Help:      "Error count by type to count for lease grants.",
+	}, []string{"error"})
+
+	leaseRevokeError = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
serathius, line 81: This should be covered by RPC metrics

server/lease/metrics.go
```diff
@@ -49,11 +49,58 @@ var (
...
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "revoke_errors",
+		Help:      "Error count by type to count for lease revokes.",
+	}, []string{"error"})
+
+	leaseRenewError = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
serathius, line 88: This should be covered by RPC metrics

server/lease/metrics.go
```diff
@@ -49,11 +49,58 @@ var (
...
+	})
+
+	leaseDetached = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "etcd_debugging",
+		Subsystem: "lease",
+		Name:      "detach_total",
+		Help:      "The number of leases that are detached from a lease item.",
+	})
+
+	initLeaseCount = prometheus.NewCounter(prometheus.CounterOpts{
```
serathius, line 67: Why we need this metric?

server/lease/lessor.go
```diff
@@ -821,6 +830,7 @@ func (le *lessor) initAndRecover() {
 		}
 	}
 	le.leaseExpiredNotifier.Init()
+	initLeaseCount.Add(float64(len(lpbs)))
```
serathius, line 833: I think the init can be called multiple times during server lifecycle.

serathius:
I'm really confused about the motivation behind those metrics. Instead of trying to add a metric for everything just in case, can you describe what do you want to achieve? We can discuss a design a metric per each use case.

server/lease/metrics.go
vivekpatani: Could you perhaps show me an example of how to get specific errors for a metric like this? @serathius

server/lease/metrics.go
vivekpatani: I'll defer to [one thread above](https://github.com/etcd-io/etcd/pull/18711#discussion_r1853085860), since the metrics are pretty much similar in terms of what getting count by error type.

server/lease/lessor.go
vivekpatani: My understanding was that in reverse calling stack, this is only ever initiated once per etcd member, but I might've missed something. `NewServer` -> `NewLessor` -> `newLessor` -> `l.initAndRecover`?

The intuition behind this metric is to see how many leases are available at the start. An issue recently occurred where we accumulated a lot of leases, and we wanted to get a number, so that it gives us an approximate idea on how long it will take for these leases to be revoked.

@serathius

server/lease/metrics.go
vivekpatani: Replied [here](https://github.com/etcd-io/etcd/pull/18711/files/ba35d471dfae1de5fd002069c4d055afb230f142#r1853090002) but we could also roll this up into another metric but I also don't want to affect the behavior of existing metrics, hence created a separate metric.

server/lease/metrics.go
vivekpatani: I'll defer to [one thread above](https://github.com/etcd-io/etcd/pull/18711#discussion_r1853085860), since the metrics are pretty much similar in terms of what getting count by error type.

vivekpatani:
Bump @serathius or @ahrtr, thanks.^

server/lease/lessor.go
```diff
@@ -280,10 +280,12 @@ func (le *lessor) SetCheckpointer(cp Checkpointer) {
 
 func (le *lessor) Grant(id LeaseID, ttl int64) (*Lease, error) {
 	if id == NoLease {
+		leaseGrantError.WithLabelValues(ErrLeaseNotFound.Error()).Inc()
```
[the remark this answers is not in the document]
ahrtr: FYI.
https://github.com/etcd-io/etcd/blob/c1171dac78f09151602fa834b2922da30d12fb5a/api/v3rpc/rpctypes/error.go#L181

```
grpc_server_handled_total{grpc_code="Aborted",grpc_method="LeaseGrant",grpc_service="etcdserverpb.Lease",grpc_type="unary"} 0
grpc_server_handled_total{grpc_code="Aborted",grpc_method="LeaseRevoke",grpc_service="etcdserverpb.Lease",grpc_type="unary"} 0
grpc_server_handled_total{grpc_code="Aborted",grpc_method="LeaseKeepAlive",grpc_service="etcdserverpb.Lease",grpc_type="bidi_stream"} 0
```

server/lease/metrics.go
ahrtr: https://github.com/etcd-io/etcd/pull/18711#discussion_r1871425664

server/lease/metrics.go
ahrtr: > number of keys per lease?

It can only get the total number of keys attached by all leases because obviously there isn't a `lease_id` label.

plsnotracking:
@serathius 

> How would you know that by metric?

I think there are two ways to go about it:

1. What @ahrtr suggested [here](https://github.com/etcd-io/etcd/pull/18711#discussion_r1806668438)
2. OR we can just have a simpler metric that shows us how many leases are attached/detached as suggested [here](https://github.com/etcd-io/etcd/pull/18711#discussion_r1871463027)

vivekpatani:
@serathius

How would you know that by metric?
I think there are two ways to go about it:

1. What @ahrtr suggested[ here](https://github.com/etcd-io/etcd/pull/18711#discussion_r1806668438)
2. OR we can just have a simpler metric that shows us how many leases are attached/detached as suggested [here](https://github.com/etcd-io/etcd/pull/18711#discussion_r1871463027)

server/lease/lessor.go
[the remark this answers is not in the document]
vivekpatani: Thank you, let me try that out. @ahrtr

server/lease/lessor.go
vivekpatani: @serathius any feedback here as well?^

server/lease/metrics.go
vivekpatani: resolving for now.

vivekpatani:
Clean up, seems like not needed/did not hear back.

PR #18723, opened 2024-10-14
Update `epHealthCommandFunc` to reuse `clientConfigFromCmd`

ahrtr:
Minor refactoring... try to reuse the common function so as to remove duplication.


client/v3/config.go
```diff
@@ -130,6 +130,30 @@ type AuthConfig struct {
 	Password string `json:"password"`
 }
 
+func (cs *ConfigSpec) Clone() *ConfigSpec {
```
serathius, line 133: Maybe add a test that ensures that Clone doesn't get outdated if any new field is added?

ahrtr: Fair point. Added a test case. Thanks

serathius (approved):

Issue #18719, opened 2024-10-11
Enable testifylint linter rules


ivanvc:
### What would you like to be added?

This issue formalizes @mmorel-35's work (thanks!) to enable the linter rules from the `testifylint` linter.

The completion track is as follows:
- [x] bool-compare
  - [x] #18686 
  - [x] #18717
- [x] compares - #18715
- [x] empty - #18712
- [x] error-is-as - #18740
- [x] error-nil - #18716
- [x] expected-actual - #18720
- [x] float-compare - #18776
- [x] formatter - #18741
- [ ] go-require - #18827
- [x] len - #18712
- [x] negative-positive - #18715 
- [x] nil-compare
  - [x] #18689 
  - [x] #18717
- [x] require-error
  - [x] #18800
  - [x] #18814

### Why is this needed?

To improve the code quality.

ghouscht:
Hey @ivanvc, do you need some help here? I could take on a few items on the list if you want me to. Just let me know which are the preferred ones.
Thank you

ivanvc:
Hi @ghouscht, thanks for offering your help with this task. Although it's unassigned (because I can't assign it to people outside of the etcd-io organization if they haven't participated in the issue), @mmorel-35 is doing it.

Therefore, I'm going to redirect the question. @mmorel-35, is there anything that @ghouscht could do to help with this task? Thanks.

ghouscht:
Thanky you for the response @mmorel-35, I found some time today and opened two PRs for the `error-is-as` and `formatter` rules:

- https://github.com/etcd-io/etcd/pull/18740
- https://github.com/etcd-io/etcd/pull/18741

I also tagged you on the PRs (sorry for the spam 🙂).

jmhbnz:
Just curious - we should compare the `make verify` runtime before enabling these checks and now to make sure we aren't increasing test runtimes too significantly.

@mmorel-35 - As you have enabled these linter rules have you noticed any meaningful increase in test runtimes in CI or locally?

ghouscht:
Running `make verify-lint` yields the following runtime on my machine

with a clean golangci-lint cache:
- with testifylint `21.941s` 
- without `20.632s` 

with golangci-lint cache present:
- with testifylint `4.926s` 
- without `4.947s`

I'd say that is neglible 🙂

PR #18740, opened 2024-10-15
fix: enable error-is-as rule from testifylint

ghouscht:
This PR enables the `error-is-as` testifylint rule as part of https://github.com/etcd-io/etcd/issues/18719

### DIFF
--- a/pkg/expect/expect.go
+++ b/pkg/expect/expect.go
@@ -193,7 +193,7 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 
 		select {
 		case <-ctx.Done():
-			return "", fmt.Errorf("context done before matching log found")
+			return "", fmt.Errorf("context done before matching log found: %w", ctx.Err())
 		case <-time.After(time.Millisecond * 10):
 			// continue loop
 		}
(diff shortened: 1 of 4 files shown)


pkg/expect/expect.go
```diff
@@ -193,7 +193,7 @@ func (ep *ExpectProcess) ExpectFunc(ctx context.Context, f func(string) bool) (s
 
 		select {
 		case <-ctx.Done():
-			return "", fmt.Errorf("context done before matching log found")
+			return "", fmt.Errorf("no matching log found: %w", ctx.Err())
```
ghouscht, line 196: I'm not entirely sure if this change is ok. Previsouly the `context.DeadlineExceeded` error was "swallowed" here. 

This was not discovered until now because the test for this wrongly used `require.ErrorAs` (not appropriate for a sentinel error value like `context.DeadlineExceeded`) instead of `require.ErrorIs`

ghouscht:
/assign mmorel-35

Hope it is ok if I assign this to you.

pkg/expect/expect.go
ahrtr: I do not see any issue to wrap the err as long as all test can pass. But I suggest not to change the error message.

ghouscht: 👍🏻 I updated the PR and left the message unchanged but wrapped the error. Tests seem to pass.

ahrtr (approved):
LGTM

Thanks

serathius (approved):

ivanvc (approved):
LGTM. Thanks, Thomas.

PR #18741, opened 2024-10-15
fix: enable formatter rule from testifylint

ghouscht:
This PR enables the `formatter` testifylint rule as part of https://github.com/etcd-io/etcd/issues/18719

### DIFF
--- a/client/pkg/testutil/assert.go
+++ b/client/pkg/testutil/assert.go
@@ -40,12 +40,12 @@ func AssertNotNil(t *testing.T, v any) {
 // Deprecated: use github.com/stretchr/testify/assert.True instead.
 func AssertTrue(t *testing.T, v bool, msg ...string) {
 	t.Helper()
-	assert.True(t, v, msg)
+	assert.True(t, v, msg) //nolint:testifylint
 }
 
 // AssertFalse
 // Deprecated: use github.com/stretchr/testify/assert.False instead.
 func AssertFalse(t *testing.T, v bool, msg ...string) {
 	t.Helper()
-	assert.False(t, v, msg)
+	assert.False(t, v, msg) //nolint:testifylint
 }
--- a/server/etcdserver/api/v2store/stats_test.go
+++ b/server/etcdserver/api/v2store/stats_test.go
@@ -26,7 +26,7 @@ func TestStoreStatsGetSuccess(t *testing.T) {
 	s := newStore()
 	s.Create("/foo", false, "bar", false, TTLOptionSet{ExpireTime: Permanent})
 	s.Get("/foo", false, false)
-	assert.Equal(t, uint64(1), s.Stats.GetSuccess, "")
+	assert.Equal(t, uint64(1), s.Stats.GetSuccess)
 }
 
 // TestStoreStatsGetFail ensures that a failed Get is recorded in the stats.
@@ -34,14 +34,14 @@ func TestStoreStatsGetFail(t *testing.T) {
 	s := newStore()
 	s.Create("/foo", false, "bar", false, TTLOptionSet{ExpireTime: Permanent})
 	s.Get("/no_such_key", false, false)
-	assert.Equal(t, uint64(1), s.Stats.GetFail, "")
+	assert.Equal(t, uint64(1), s.Stats.GetFail)
 }
 
 // TestStoreStatsCreateSuccess ensures that a successful Create is recorded in the stats.
 func TestStoreStatsCreateSuccess(t *testing.T) {
 	s := newStore()
 	s.Create("/foo", false, "bar", false, TTLOptionSet{ExpireTime: Permanent})
-	assert.Equal(t, uint64(1), s.Stats.CreateSuccess, "")
+	assert.Equal(t, uint64(1), s.Stats.CreateSuccess)
 }
 
 // TestStoreStatsCreateFail ensures that a failed Create is recorded in the stats.
@@ -49,7 +49,7 @@ func TestStoreStatsCreateFail(t *testing.T) {
 	s := newStore()
 	s.Create("/foo", true, "", false, TTLOptionSet{ExpireTime: Permanent})
 	s.Create("/foo", false, "bar", false, TTLOptionSet{ExpireTime: Permanent})
-	assert.Equal(t, uint64(1), s.Stats.CreateFail, "")
+	assert.Equal(t, uint64(1), s.Stats.CreateFail)
 }
 
 // TestStoreStatsUpdateSuccess ensures that a successful Update is recorded in the stats.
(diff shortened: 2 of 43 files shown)


client/pkg/testutil/assert.go
```diff
@@ -40,12 +40,12 @@ func AssertNotNil(t *testing.T, v any) {
 // Deprecated: use github.com/stretchr/testify/assert.True instead.
 func AssertTrue(t *testing.T, v bool, msg ...string) {
 	t.Helper()
-	assert.True(t, v, msg)
+	assert.True(t, v, msg) //nolint:testifylint
```
ghouscht, line 43: Both the `AssertTrue` and `AssertFalse` functions are part of the public API of the `testutils` pkg. It looks like they are no longer used within the etcd codebase but they could be used outside of the etcd codebase. I thought it is best to keep them as they are, thus the `nolint` directive. Let me know if you think that is ok.

ghouscht:
FYI: @mmorel-35 here is the second PR for the `formatter`rule. From my point of view it is ready but I marked it as draft to keep the number of open and ready testifylint PRs small.

server/etcdserver/api/v2store/stats_test.go
```diff
@@ -26,76 +26,76 @@ func TestStoreStatsGetSuccess(t *testing.T) {
 	s := newStore()
 	s.Create("/foo", false, "bar", false, TTLOptionSet{ExpireTime: Permanent})
 	s.Get("/foo", false, false)
-	assert.Equal(t, uint64(1), s.Stats.GetSuccess, "")
+	assert.Equal(t, uint64(1), s.Stats.GetSuccess)
```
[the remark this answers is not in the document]
ghouscht: I fixed it manually. testifylint replaced these with the f functions but obviously that is useless

ghouscht:
> There are many places where the expected or unexpected error is is printed with `: %v", err`
> 
> I'm wondering if it's not redondant with the behavior of testify which will print it

Yes, that is true. I can clean them up if you and the maintainers agree on that.

tests/e2e/corrupt_test.go
```diff
@@ -339,14 +339,14 @@ func TestCompactHashCheckDetectCorruptionInterrupt(t *testing.T) {
 
 	// Wait until the node finished compaction and the leader finished compaction hash check
 	_, err = epc.Procs[slowCompactionNodeIndex].Logs().ExpectWithContext(ctx, expect.ExpectedResponse{Value: "finished scheduled compaction"})
-	require.NoError(t, err, "can't get log indicating finished scheduled compaction")
+	require.NoErrorf(t, err, "can't get log indicating finished scheduled compaction")
 
 	leaderIndex := epc.WaitLeader(t)
 	_, err = epc.Procs[leaderIndex].Logs().ExpectWithContext(ctx, expect.ExpectedResponse{Value: "finished compaction hash check"})
-	require.NoError(t, err, "can't get log indicating finished compaction hash check")
+	require.NoErrorf(t, err, "can't get log indicating finished compaction hash check")
```
ahrtr, line 346: Question: there is no variable-length variable in this case, shouldn't `require.NoError` be better than `require.NoErrorf`?

ahrtr: Note this is a generic question.

ahrtr:
> There are many places where the expected or unexpected error is is printed with `: %v", err`
> 
> I'm wondering if it's not redondant with the behavior of testify which will print it

Yes, testify already prints the error, so we don't need to print it again. But not a big deal.

tests/e2e/corrupt_test.go
ahrtr: We should always use `*f` functions when `require-f-funcs: true` no matter there is variable-length variables or not?

tools/.golangci.yaml
```diff
@@ -115,7 +115,8 @@ linters-settings: # please keep this alphabetized
       - error-nil
       - expected-actual
       - float-compare
-      - formatter
       - go-require
       - require-error
     enable-all: true
+    formatter:
+      require-f-funcs: true
```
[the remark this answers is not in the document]
ghouscht: https://golangci-lint.run/usage/linters/#testifylint states:
> To require f-assertions if format string is used.

I'll add a comment 👍🏻

server/etcdserver/api/v2store/stats_test.go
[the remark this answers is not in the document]
ghouscht: raised: https://github.com/Antonboom/testifylint/issues/194

ghouscht:
> > There are many places where the expected or unexpected error is is printed with `: %v", err`
> > I'm wondering if it's not redondant with the behavior of testify which will print it
> 
> Yes, testify already prints the error, so we don't need to print it again. But not a big deal.

👍🏻  I cleaned up the messages and removed redundant information.

tools/.golangci.yaml
```diff
@@ -115,7 +115,9 @@ linters-settings: # please keep this alphabetized
       - error-nil
       - expected-actual
       - float-compare
-      - formatter
       - go-require
       - require-error
     enable-all: true
+    formatter:
+      # Require f-assertions (e.g. assert.Equalf) if a format string is used.
```
ivanvc, line 122: It could be just a nitpick on the comment, but after reading
* this conversation https://github.com/etcd-io/etcd/pull/18741#discussion_r1802658930
* and the references https://github.com/Antonboom/testifylint?tab=readme-ov-file#formatter, https://github.com/Antonboom/testifylint?tab=readme-ov-file#historical-reference.

I think this comment's "format string" part is inaccurate. The suggestion is to use the `*f` functions regardless of whether the message string has formatting or not (if I understood correctly).

ghouscht: You're right, even in the documentation from golangci-lint it is not very accurate. I think it something like
> # Require f-assertions (e.g. assert.Equalf) if a message is passed to the assertion.
might be better.

ahrtr: How about 


> Require f-assertions (e.g. `assert.Equalf`) if a message is passed to the assertion, even there is no variable-length variables, i.e. require `require.NoErrorf` for both cases below,
> `require.NoErrorf(t, err, "whatever message")`
> `require.NoErrorf(t, err, "whatever message: %v", v)`

ghouscht:
/test pull-etcd-integration-1-cpu-amd64

One last try, seems there is always some kind of timeout in this test.

ahrtr:
From golang programming perspective, I think we still prefer non-f-functions (i.e. `fmt.Print`) to f-functions (i.e. `fmt.Printf`) when there is no variable-length parameters.

For example, we prefer to use `Print` instead of `Printf` when no variable-length parameters.

in `fmt` package
```
func Print(a ...any) (n int, err error)
func Printf(format string, a ...any) (n int, err error)
```

in `log` package
```
func (l *Logger) Print(v ...any)
func (l *Logger) Printf(format string, v ...any)
```

But in [stretchr/testify](https://github.com/stretchr/testify), each f-function is actually a syntax sugar of corresponding non-f-function. For example, `Containsf` is just a simple encapsulation of `Contains`. also refer to [here](https://github.com/stretchr/testify/blob/a012e45d185128fe5bf3370e42b0854856e0998e/assert/assertion_format.go#L25-L30).

So there is no reason to keep both. But they can't simplify remove anyone otherwise it will break lots of application's tests.

```
func Containsf(t TestingT, s interface{}, contains interface{}, msg string, args ...interface{}) bool {
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	return Contains(t, s, contains, append([]interface{}{msg}, args...)...)
}
```

So it's accepted to always require `f-functions` for `stretchr/testify`, but not for golang standard lib.

ahrtr (approved):
LGTM

Thanks

ivanvc (approved):
To @ahrtr's point, I agree it may not be straight forward if you're not familiar with `stretchr/testify`. When I initially saw the pull request I had to check the documentation to understand the issue and the linter rule.

But with that aside, it LGTM. Thanks, @ghouscht.

ahrtr:
Raised a followup https://github.com/etcd-io/etcd/pull/18766

Antonboom:
Hi, guys

@ghouscht appeciated for using of `testifylint`. @mmorel-35, thank you for support.

@ahrtr, @ivanvc I improved the linter description based on your threads:

1. https://github.com/Antonboom/testifylint?tab=readme-ov-file#3
2. https://github.com/golangci/golangci-lint/pull/5131

Also, I reviewed the `testify` code deeper and was surprised, but you can use non-string arg in asserts if this is the single argument, like
```
assert.Equal(a, b, new(time.Time))
```

It leads to the following points:

- https://github.com/Antonboom/testifylint/pull/200
- https://github.com/stretchr/testify/issues/1679
- https://github.com/Antonboom/testifylint/issues/201
- https://github.com/Antonboom/testifylint/issues/202

Thanks! 😭

PR #18748, opened 2024-10-17
fix: enable gofumpt instead of gofmt linter in client

mmorel-35:
enables and fixes [gofumpt](https://golangci-lint.run/usage/linters/#gofumpt) linter issues in client directory

### DIFF
--- a/client/internal/v2/auth_user.go
+++ b/client/internal/v2/auth_user.go
@@ -23,9 +23,7 @@ import (
 	"path"
 )
 
-var (
-	defaultV2AuthPrefix = "/v2/auth"
-)
+var defaultV2AuthPrefix = "/v2/auth"
 
 type User struct {
 	User     string   `json:"user"`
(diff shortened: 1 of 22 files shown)


ivanvc:
@ahrtr, @serathius, thoughts on replacing `gofmt` with [`gofumpt`](https://github.com/mvdan/gofumpt)?

ahrtr:
A quick question, are all the changes in this PR detected/suggested by `gofumpt`? 

It seems a little better to apply `gofumpt` based on the changes in this PR.

ahrtr (approved):
LGTM

Thanks

ivanvc (approved):
LGTM. Should we add a note in the contributor guide about using `gofumpt` instead of `gofmt`? I believe the latter is more widely used by editors, IDEs, etc. (To be done with the pull request when you enable this linter).

ahrtr:
To be clearer, we enable both `gofmt` via [verify-gofmt](https://github.com/etcd-io/etcd/blob/a594a62eab8846294e0bd8a5b26ec40c1e0de9fc/Makefile#L77-L78) and [gofumpt](https://github.com/mvdan/gofumpt) via the [verify-lint](https://github.com/etcd-io/etcd/blob/a594a62eab8846294e0bd8a5b26ec40c1e0de9fc/Makefile#L93-L94). 

> Enforce a stricter format than gofmt, while being backwards compatible. That is, gofumpt is happy with a subset of the formats that gofmt is happy with.

Issue #18679, opened 2024-10-05
Write txn shouldn't End() on a failure


shyamjvs:
### What happened?

Offshoot of https://github.com/etcd-io/etcd/issues/18667#issuecomment-2394848548

When two nodes both execute the same write txn (i.e validation step already passed on both), but one ends up in failed execution and the other in success, we shouldn't let the failed node increment its CI and move on - but explicitly fail and crash at the old CI - because we [never expect a write txn to fail](https://github.com/etcd-io/etcd/blob/c1976a67172fff1e63ee1b0117d34e8be8c35b45/server/etcdserver/txn/txn.go#L312-L313). Otherwise it would cause asymmetric side-effect across nodes and lead to data inconsistency.

Here's where I think our code may be potentially violating that:

https://github.com/etcd-io/etcd/blob/c1976a67172fff1e63ee1b0117d34e8be8c35b45/server/etcdserver/txn/txn.go#L307-L314

If I'm reading that correctly, when we run into a txn failure (which may contain some partially executed writes) the txn is ended **before** triggering server panic - which means there could be a KV rev and CI bump with some incorrect changes. So if/when the server restarts, it will assume it's caught up (till that badly applied txn) and move on to the next record leading to inconsistent state. Reproducing such failure is going to be hard iiuc but it seems like we should panic without calling txn.End here?

### What did you expect to happen?

The etcd node on which write txn execution fails should crash without trying to commit the failed transaction (or other side effects like incrementing KV revision or CI).

### Anything else we need to know?

There was no observed failure or a confirmed repro I have for this issue, but bringing it up as a potential risk in our current code.

Please share any thoughts about making the above change/trying to repro or if you think this is a non-issue.
/cc @serathius @ahrtr

serathius:
Context why this was introduced https://github.com/etcd-io/etcd/pull/14149

shyamjvs:
As I read through that change, it seems the goal was to make timeouts of read-only txns not cause panic. But the change also makes undesirable side-effect of letting the partial-write end before panic.

shyamjvs:
This comment from @ptabor was quite relevant actually (the miss seems to be to not panic right away):

> I have small preference towards apply workflow being extremely strict and deterministic and any unexpected error causing premature exit from function (even in RO) code is something that should lead to investigation and fixing.

@ahrtr - please share any add'l context you might have from that PR - or I can make a change to remove txn.End and test

ahrtr:
[...]

PR #14178, opened 2022-06-29
[3.5] server: don't panic in readonly serializable txn

lavacat:
Problem: We pass grpc context down to applier in readonly serializable txn.
This context can be cancelled for example due to timeout.
This will trigger panic inside applyTxn

Solution: Only panic for transactions with write operations

fixes https://github.com/etcd-io/etcd/issues/14110
backported from main https://github.com/etcd-io/etcd/pull/14149

Signed-off-by: Bogdan Kanivets <bkanivets@apple.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #18749, opened 2024-10-17
Fix risk of a partial write txn being applied

shyamjvs:
Fixes https://github.com/etcd-io/etcd/issues/18679

I was able to confirm this risk exists based on the unit test failing with txnWrite.End() still around. I did have to add a few milliseconds of wait time between the txn End and panic to reliably reproduce it because it depends on whether backend commit happens in that window or not.

/cc @serathius @ahrtr

### DIFF
--- a/server/etcdserver/txn/txn_test.go
+++ b/server/etcdserver/txn/txn_test.go
@@ -16,6 +16,9 @@ package txn
 
 import (
 	"context"
+	"crypto/sha256"
+	"io"
+	"os"
 	"strings"
 	"testing"
 	"time"
@@ -336,8 +339,8 @@ func TestReadonlyTxnError(t *testing.T) {
 	}
 }
 
-func TestWriteTxnPanic(t *testing.T) {
-	b, _ := betesting.NewDefaultTmpBackend(t)
+func TestWriteTxnPanicWithoutApply(t *testing.T) {
+	b, bePath := betesting.NewDefaultTmpBackend(t)
 	defer betesting.Close(t, b)
 	s := mvcc.NewStore(zaptest.NewLogger(t), b, &lease.FakeLessor{}, mvcc.StoreConfig{})
 	defer s.Close()
@@ -366,6 +369,26 @@ func TestWriteTxnPanic(t *testing.T) {
 			},
 		},
 	}
+
+	// compute DB file hash before applying the txn
+	dbHashBefore, err := computeFileHash(bePath)
+	if err != nil {
+		t.Fatalf("failed to compute DB file hash before txn: %v", err)
+	}
+
+	// we verify the following things below:
+	// - server panics after a write txn aply fails (invariant being tested: server should never try to move on from a failed write)
+	// - no writes from the txn are applied to the backend (invariant being tested: failed write should have no side-effect on DB state besides panic)
+	defer func() {
+		dbHashAfter, err := computeFileHash(bePath)
+		if err != nil {
+			t.Fatalf("failed to compute DB file hash after txn: %v", err)
+		}
+
+		if dbHashAfter != dbHashBefore {
+			t.Fatalf("expected DB hash after failed txn: %s, got: %s", dbHashBefore, dbHashAfter)
+		}
+	}()
 
 	assert.Panics(t, func() { Txn(ctx, zaptest.NewLogger(t), txn, false, s, &lease.FakeLessor{}) }, "Expected panic in Txn with writes")
 }
@@ -564,6 +587,20 @@ func setupAuth(t *testing.T, be backend.Backend) auth.AuthStore {
 	require.NoError(t, err)
 
 	return as
+}
+
+func computeFileHash(filePath string) (string, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, file); err != nil {
+		return "", err
+	}
+	return string(h.Sum(nil)), nil
 }
 
 // CheckTxnAuth variables setup.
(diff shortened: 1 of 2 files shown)


server/etcdserver/txn/txn_test.go
```diff
@@ -367,6 +370,26 @@ func TestWriteTxnPanic(t *testing.T) {
...
+	// compute DB file hash before applying the txn
+	dbHashBefore, err := computeFileHash(bePath)
+	if err != nil {
+		t.Fatalf("failed to compute DB file hash before txn: %v", err)
+	}
+
+	// we verify the following things below:
+	// - server panics after a write txn aply fails (invariant being tested: server should never try to move on from a failed write)
+	// - no writes from the txn are applied to the backend (invariant being tested: failed write should have no side-effect on DB state besides panic)
+	defer func() {
```
serathius, line 382: Don't think there is a need to defer this? `assert.Panics` should handle panic?

server/etcdserver/txn/txn_test.go
```diff
@@ -367,6 +370,26 @@ func TestWriteTxnPanic(t *testing.T) {
 		},
 	}
 
+	// compute DB file hash before applying the txn
+	dbHashBefore, err := computeFileHash(bePath)
+	if err != nil {
+		t.Fatalf("failed to compute DB file hash before txn: %v", err)
+	}
```
ahrtr, line 377: Please use `require.NoErrorf`.

cc @mmorel-35

server/etcdserver/txn/txn_test.go
```diff
@@ -367,6 +370,26 @@ func TestWriteTxnPanic(t *testing.T) {
...
+
+	// we verify the following things below:
+	// - server panics after a write txn aply fails (invariant being tested: server should never try to move on from a failed write)
+	// - no writes from the txn are applied to the backend (invariant being tested: failed write should have no side-effect on DB state besides panic)
+	defer func() {
+		dbHashAfter, err := computeFileHash(bePath)
+		if err != nil {
+			t.Fatalf("failed to compute DB file hash after txn: %v", err)
+		}
+
```
ahrtr, line 387: ditto, pls use `require.NoErrorf`

server/etcdserver/txn/txn_test.go
```diff
@@ -367,6 +370,26 @@ func TestWriteTxnPanic(t *testing.T) {
...
+		if dbHashAfter != dbHashBefore {
+			t.Fatalf("expected DB hash after failed txn: %s, got: %s", dbHashBefore, dbHashAfter)
+		}
```
ahrtr, line 390: What's the recommendation here? `require.Equalf` ? cc @mmorel-35

server/etcdserver/txn/txn_test.go
ahrtr: Did not follow @serathius 's comment above. 

- The purpose here is to verify the hash of the db did not change after the panicking. It's good!
- When panicking happens, any functions [deferred](https://go.dev/ref/spec#Defer_statements) are then executed as usual

shyamjvs: that's correct, test function stops executing after assert.panic - so further verification needs to happen either inside a defer or a t.Cleanup()

server/etcdserver/txn/txn_test.go
shyamjvs: thanks for this suggestion - made it much cleaner

server/etcdserver/txn/txn_test.go
fuweid: I think what @serathius said is that the `assert.Panics` is capable to recover the panic. there is no need to use defer.
We should check the hash after `assert.Panics`.

fuweid: FYI https://github.com/stretchr/testify/blob/bb548d0473d4e1c9b7bbfd6602c7bf12f7a84dd2/assert/assertions.go#L1190-L1205

chaochn47: Shyam - if the initial assertion requires opening a bbolt file, it is protected by the advisory flock so it appears that so further verification could be added after `assert.Panics`. 

However, this advisory lock does not block you from opening the file and perform checksum.

shyamjvs: sorry I stand corrected - the test indeed continues after the assert (the defer pattern was a remnant from earlier when I was calling Txn() directly without an assert) 

however, the test still got stuck after adding the assert because of the `defer betesting.Close(t, b)` at the start of the function, because it tries to write to an unbuffered channel [here](https://github.com/etcd-io/etcd/blob/3869f0e77dce84fc3da8d9385927fcdff6c87169/server/storage/backend/backend.go#L446) that is no longer read from anywhere

fixed both now, thanks!

ahrtr (approved):
LGTM

Thanks @shyamjvs 

Ideally it would be great if we could create an e2e or integration test to reproduce the partially committed/persisted issue.

ahrtr:
@shyamjvs Please signoff the commit. Please read https://github.com/etcd-io/etcd/pull/18749/checks?check_run_id=31776671478

server/etcdserver/txn/txn_test.go
ahrtr: > I think what @serathius said is that the `assert.Panics` is capable to recover the panic. there is no need to use defer.

thx for the clarification. `defer` is still a good pattern for such case, but not a big deal though.

shyamjvs:
done - thanks @ahrtr

serathius (approved):

ahrtr:
@shyamjvs do you have bandwidth to backport the PR to 3.5?

FYI. The PR (issue) was also backported to 3.5 in https://github.com/etcd-io/etcd/pull/14178

shyamjvs:
Sorry for delay, sent out the backport for review.

PR #18750, opened 2024-10-18
embed: fix typo client-listen-x to listen-client-x

mingqing:
When I was separating GRPC and HTTP ports, the following error occurred, causing ETCD to not start properly and causing confusion for me.

error log:

```text
cannot bind both --client-listen-urls and --client-listen-http-urls on the same url https://192.168.31.23:2379
```

key config:

```text
...
    --listen-client-urls=https://127.0.0.1:2379,https://192.168.31.23:2379 \
    --listen-client-http-urls=https://192.168.31.23:2379 \
...
```

 It should be a spelling error in the error log output, client-listen-* will be listen-client-*. I hope it can help others.


ivanvc (approved):
LGTM. Thanks, @mingqing.

ahrtr (approved):
Good catch, Thanks.

PR #18752, opened 2024-10-18
fix: use require.NoError instead of t.Fatal(err) in contrib and tools packages

mmorel-35:
There is no linter for this.
This uses testify instead of testing for
With 
```go
require.NoError(t, err)

require.Error(t, err)
```
instead of

```go
if err != nil {
    t.Fatal(err)
}


if err == nil {
    t.Fatal(err)
}
```

### DIFF
--- a/contrib/raftexample/kvstore_test.go
+++ b/contrib/raftexample/kvstore_test.go
@@ -17,6 +17,8 @@ package main
 import (
 	"reflect"
 	"testing"
+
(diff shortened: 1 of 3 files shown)


ahrtr (approved):
LGTM

Note the we are going to remove `raftexample` once https://github.com/etcd-io/raft/issues/2 is resovled.

fuweid (approved):
LGTM

PR #18753, opened 2024-10-19
fix: use require.NoError instead of t.Fatal(err) in tests package (part 1)

mmorel-35:
There is no linter for this.
This uses testify instead of testing for
With 
```go
require.NoError(t, err)

require.Error(t, err)
```
instead of

```go
if err != nil {
    t.Fatal(err)
}


if err == nil {
    t.Fatal(err)
}
```

### DIFF
--- a/tests/common/alarm_test.go
+++ b/tests/common/alarm_test.go
@@ -21,6 +21,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/config"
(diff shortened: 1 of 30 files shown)


ahrtr:
@mmorel-35 Thanks for you constant effort, but please try to make each each such PR as small as possible, i.e < 100 lines of code change, so that each one can be approved & merged soon.

serathius (approved):

serathius:
Read through it, still +1 to what @ahrtr said. Don't optimize for number of lines changed, optimize for time it takes to review. That will help you land your work faster.

serathius:
Please fix
```
robustness/model/deterministic_test.go:39:6: formatter: use require.NoErrorf (testifylint)
					require.NoError(t, err, "Failed to load state")
```

PR #18754, opened 2024-10-19
fix: enable gofumpt instead of gofmt linter in contrib, etcdctl and etcdutl

mmorel-35:
enables and fixes [gofumpt](https://golangci-lint.run/usage/linters/#gofumpt) linter issues in contrib, etcdctl and etcdutl directories


ahrtr (approved):

ivanvc (approved):
LGTM. Thanks, @mmorel-35

PR #18771, opened 2024-10-23
Ensure compaction is executed at even intervals to avoid overload

serathius:
Should reduce number of flakes due to etcd being overloaded by too many compaction at once

### DIFF
--- a/tests/robustness/traffic/traffic.go
+++ b/tests/robustness/traffic/traffic.go
@@ -27,6 +27,7 @@ import (
 	"go.etcd.io/etcd/tests/v3/robustness/client"
 	"go.etcd.io/etcd/tests/v3/robustness/identity"
 	"go.etcd.io/etcd/tests/v3/robustness/model"
+	"go.etcd.io/etcd/tests/v3/robustness/random"
 	"go.etcd.io/etcd/tests/v3/robustness/report"
 )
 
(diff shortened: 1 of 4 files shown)


serathius:
cc @jmhbnz

tests/robustness/traffic/traffic.go
```diff
@@ -181,5 +197,33 @@ func (p Profile) WithoutCompaction() Profile {
 type Traffic interface {
 	Run(ctx context.Context, c *client.RecordingClient, qpsLimiter *rate.Limiter, ids identity.Provider, lm identity.LeaseIDStorage, nonUniqueWriteLimiter ConcurrencyLimiter, finish <-chan struct{})
 	ExpectUniqueRevision() bool
-	WithoutCompact() Traffic
+}
+
+func RunCompactLoop(ctx context.Context, c *client.RecordingClient, period time.Duration, finish <-chan struct{}) {
+	var lastRev int64 = 2
```
jmhbnz, line 203: Why is `lastRev` always set to 2 here?

serathius: It's initial value.

jmhbnz: Ahh there is only one possible compaction loop per simulation. I understand now thanks.

jmhbnz (approved):
LGTM

PR #18772, opened 2024-10-23
Allow duplicated puts in robustness 

serathius:
Rewrite patching history, so instead of patching only  puts by assuming all of them are unique, rather count client calls and only allow patching for writes that were called once. 

This is needed to implement Kubernetes like compaction that reuses same key.

cc @ah8ad3 @ahrtr @MadhavJivrajani @henrybear327


serathius:
ping @ahrtr

serathius:
cc @siyuanfoundation

serathius:
cc @jmhbnz

serathius:
Let's split it into smaller PRs

PR #18774, opened 2024-10-24
fix: enable tenv linter

mmorel-35:
Tenv is an analyzer that detects using os.Setenv instead of t.Setenv since Go1.17.

enables and fixes [tenv](https://golangci-lint.run/usage/linters/#tenv) linter issues


serathius:
This is awesome improvement.

PR #18775, opened 2024-10-24
fix: enable thelper linter in pkg and server directories (part 1)

mmorel-35:
enables and fixes [thelper](https://golangci-lint.run/usage/linters/#thelper) linter issues in pkg and server directories

### DIFF
--- a/pkg/proxy/server_test.go
+++ b/pkg/proxy/server_test.go
@@ -47,6 +47,7 @@ func TestServer_Unix_Secure_DelayTx(t *testing.T)   { testServer(t, "unix", true
 func TestServer_TCP_Secure_DelayTx(t *testing.T)    { testServer(t, "tcp", true, true) }
 
 func testServer(t *testing.T, scheme string, secure bool, delayTx bool) {
+	t.Helper()
 	lg := zaptest.NewLogger(t)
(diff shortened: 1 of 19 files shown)


pkg/proxy/server_test.go
```diff
@@ -486,6 +488,7 @@ func TestServerHTTP_Secure_DelayTx(t *testing.T)   { testServerHTTP(t, true, tru
 func TestServerHTTP_Insecure_DelayRx(t *testing.T) { testServerHTTP(t, false, false) }
 func TestServerHTTP_Secure_DelayRx(t *testing.T)   { testServerHTTP(t, true, false) }
 func testServerHTTP(t *testing.T, secure, delayTx bool) {
+	t.Helper()
```
serathius, line 491: Don't agree with that change. From what I know calling `t.Helper()` will change which line failure is attributed? https://pkg.go.dev/testing#T.Helper.

I'm pretty sure this function is not a helper.

This kind of flake make me think we should skip this linter.

serathius (changes requested):
Etcd code passes *testing.T a lot because it uses multiple advanced testing approaches that require non trivial tooling. I don't see any benefits of using t.Helper here, so I'm against using this linter.

PR #18779, opened 2024-10-24
Define a `DLocker` interface for distributed locker

ahrtr:
Also updated the [contrib/lock](https://github.com/etcd-io/etcd/tree/main/contrib/lock) example to demo how to use the `DLocker`. 

For more background, please refer to [the article by Martin Kleppmann](https://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html).

- It should be an improvement to help users better use etcd as a distributed lock.
- It's also backward compatible. It won't break any existing applications


ivanvc:
/test pull-etcd-verify

I want to check if it was a flake in `govet`.

PR #18782, opened 2024-10-24
fix: use require instead of t.Fatal(err) in tests/robustness package

mmorel-35:
There is no linter for this.


ahrtr (approved):

ivanvc (approved):
LGTM

PR #18799, opened 2024-10-28
[3.5] manual: Fix risk of a partial write txn being applied

shyamjvs:
Manually backported https://github.com/etcd-io/etcd/pull/18749 given conflicting changes.

/assign @ahrtr


ahrtr (approved):
LGTM

The workflows on arm64 are blocked because we don't have arm64 runners. Once the workflows are migrated to prow, we will remove arm64 workflow on github.

serathius (approved):

PR #18800, opened 2024-10-28
fix: enable require-error rule from testifylint in client, pkg and server packages

mmorel-35:
This fixes [require-error](https://github.com/Antonboom/testifylint?tab=readme-ov-file#require-error) rule from [testifylint](https://github.com/Antonboom/testifylint) in client, pkg and server packages

Related to #18719


ivanvc (approved):
LGTM. Thanks, @mmorel-35.

ahrtr (approved):

PR #18803, opened 2024-10-28
Handle read-only write txn failures gracefully

shyamjvs:
Following up from https://github.com/etcd-io/etcd/issues/18667#issuecomment-2397913066.

The idea is simple:
1. We have two kinds of txns - read-only and write
2. Read-only txns do not have any write operations inside their success/failure operation blocks - these are already bypassing apply layer and served directly via linearizable read loop ([xref](https://github.com/etcd-io/etcd/blob/448fb7e36c11971d610bef241f2ffb7ddf39788d/server/etcdserver/v3_server.go#L161-L192))
3. Write txns have at least 1 write operation inside their success/failure operation blocks - however during txn execution (apply time), the actual operation paths chosen based on "compare" outputs could effectively render them as read-only txns

This change improves the behavior for 3 by avoiding an unnecessary panic when such read-only txns see an error - instead return the error back to server handler and move on with applying subsequent entries. Panic + server restart seems unnecessary overhead in such cases. This also paves a way to bring back context-based timeouts for such txns. One case where that could help is during high write contention scenarios where CAS operations often end up going to "failure" block and triggers a (potentially slow) range request.

/cc @serathius @ahrtr

### DIFF
--- a/server/etcdserver/txn/txn.go
+++ b/server/etcdserver/txn/txn.go
@@ -275,7 +275,11 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
 	if isWrite {
 		trace.AddField(traceutil.Field{Key: "read_only", Value: false})
 	}
-	_, err := checkTxn(txnRead, rt, lessor, txnPath)
+	// checkTxn performs pre-flight checks on the txn to confirm that it's valid (e.g range operation is not asking for
+	// a future/compacted revision, put operations doesn't supply a lease which doesn't exist, etc) or returns an error.
+	// performsWrite says if the txn actually performs any write operations based on the compare choices we land with
+	// (i.e. txnPath) - which has been pre-computed. So even if isWrite was true for the txn, performsWrite may be false.
+	_, performsWrite, err := checkTxn(txnRead, rt, lessor, txnPath)
 	if err != nil {
 		txnRead.End()
 		return nil, nil, err
@@ -286,13 +290,13 @@ func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWit
 	// serialized on the raft loop, the revision in the read view will
 	// be the revision of the write txnWrite.
 	var txnWrite mvcc.TxnWrite
-	if isWrite {
+	if performsWrite {
 		txnRead.End()
 		txnWrite = kv.Write(trace)
 	} else {
 		txnWrite = mvcc.NewReadOnlyTxnWrite(txnRead)
 	}
-	txnResp, err := txn(ctx, lg, txnWrite, rt, isWrite, txnPath)
+	txnResp, err := txn(ctx, lg, txnWrite, rt, performsWrite, txnPath)
 	txnWrite.End()
 
 	trace.AddField(
@@ -314,6 +318,8 @@ func txn(ctx context.Context, lg *zap.Logger, txnWrite mvcc.TxnWrite, rt *pb.Txn
 			// - data inconsistency across different etcd members if they applied the txn asymmetrically
 			lg.Panic("unexpected error during txn with writes", zap.Error(err))
 		} else {
+			// Given the txn performs only read operations, it does not cause any apply layer side-effects.
+			// Therefore, it's safe for the server to return an error here (without panic) and continue applying subsequent records.
 			lg.Error("unexpected error during readonly txn", zap.Error(err))
 		}
 	}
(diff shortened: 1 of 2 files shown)


serathius:
Don't like the change and don't think it is correct. My high level understanding of the motivation for `isWrite` is whether TXn is executed via [apply](https://github.com/etcd-io/etcd/blob/3696b3241194ead696d47cb1093e9df855813d86/server/etcdserver/apply/apply.go#L166) or [locally](https://github.com/etcd-io/etcd/blob/3696b3241194ead696d47cb1093e9df855813d86/server/etcdserver/v3_server.go#L161). Even though this is not expressed in code directly, it's reflected in:
* Type of transaction we use (shared vs concurrent). We want to used shared only in apply.
* Closing and reopening transaction. This cannot I mean really cannot be done outside of apply loop.
* Decision whether we panic or return error (we need to panic in apply)

I don't see any benefits of what you are proposing, and I'm against about making `isWrite` inconsistent between those checks. 

What I can recommend is replacing `isWrite` check with something more explicitly showing differences between how TXN should be executed in Apply vs locally.

ahrtr:
I understand the motivation of this PR to remove unnecessary panic when the execution branch/path is readonly, but obviously it complicates the TXN applying workflow and without big benefit. So I suggest that we don't change it unless there is a strong justification.



> My high level understanding of the motivation for `isWrite` is whether TXn is executed via [apply](https://github.com/etcd-io/etcd/blob/3696b3241194ead696d47cb1093e9df855813d86/server/etcdserver/apply/apply.go#L166) or [locally](https://github.com/etcd-io/etcd/blob/3696b3241194ead696d47cb1093e9df855813d86/server/etcdserver/v3_server.go#L161).

That's the original only use case. The https://github.com/etcd-io/etcd/pull/14149 also removed the unnecessary panic in applying workflow; which I think is also valid use case, although it introduced a potential issues as @shyamjvs pointed out and fixed in https://github.com/etcd-io/etcd/pull/18749

shyamjvs:
Thanks for those comments @serathius @ahrtr. Few clarifications:

- This change does not modify the behavior of read-only txns being executed locally. Similarly, write txns (whether they end up actually doing a write or not) still go through apply layer. In fact, they **have** to go through apply layer (otherwise compare operations cannot be serializably evaluated)
- Reg mvcc txn handling, there are two cases here:
  - read-only txn that's handled outside of apply loop - this **does not** close + reopen the txn and the same txnRead we create for check phase is continued in use for actual txn execution but upgraded as `mvcc.NewReadOnlyTxnWrite` (this behavior is unchanged)
  - write txn that's handled inside the apply loop which actually performs write - we end the txnRead used for check phase and open a txnWrite for execution phase, which is safe because all write txns are serialized in apply so another write won't interfere (this behavior is also unchanged)
  - now for write txn inside apply loop that **does not** actually perform writes - this PR is changing the behavior to continue using the txnRead from check phase also for execution phase. IMHO this is actually safer than using a write txn for an operation that **we know** should be read-only. Because if the txn accidentally tries to perform any writes (despite check phase determining it to be read-only), server will now explicitly panic thanks to using [mvcc.ReadOnlyTxnWrite](https://github.com/etcd-io/etcd/blob/3696b3241194ead696d47cb1093e9df855813d86/server/storage/mvcc/kv.go#L92)
- Reg "should we panic on read-only write txn error" - this is orthogonal decision from above and we can choose to go either way for both. I'm not too confident about the benefit of panic here though - as it's going to be costly recovery operation, request fails anyway but we lose opportunity to surface the error back to client with panic. FWIW this rationale already lives [here](https://github.com/etcd-io/etcd/blob/3696b3241194ead696d47cb1093e9df855813d86/server/etcdserver/txn/txn.go#L316-L318) today
- The change is relatively small (mainly changing checkTxn part), but I admit it is additional complexity. If we think it isn't worth the tradeoff, I'm happy to close the PR

ahrtr:
> * The change is relatively small (mainly changing checkTxn part), but I admit it is additional complexity. If we think it isn't worth the tradeoff, I'm happy to close the PR

Suggest to close it. Readability & understandability is more important in the critical applying workflow, so let's keep it as simple as possible.

shyamjvs:
Ack - will let @serathius also chime in with any final thoughts before closing.

serathius:
> Suggest to close it. Readability & understandability is more important in the critical applying workflow, so let's keep it as simple as possible.

+1 to that. 

> The change is relatively small (mainly changing checkTxn part), but I admit it is additional complexity. If we think it isn't worth the tradeoff, I'm happy to close the PR

We are especially careful about small changes. I don't think there is a tradeoff as we are just sacrificing readability to solve a problem that no one reported. I would like to see some evidence of this being a problem.

ahrtr:
Closing...

@shyamjvs Please feel free to reach out if you still have any concern or comment. Thank you!

shyamjvs:
Sounds good, thanks for discussing!

Issue #17329, opened 2024-01-26
Add an e2e tests that detects error logs during etcd bootstrap


serathius:
### What would you like to be added?

Etcd should not write any error logs during normal operation. We could add a test that starts etcd cluster, does some simple operations and validates if there is any error level log written.

This could be further improved by enabling this check in all e2e tests when cluster/server is being closed. As some test intentionally inject errors, we could make it configurable option that is enabled by default and disabled in all tests we expect error to be written.

### Why is this needed?

Logs are used to provide useful debug information to users, however if mislabeled they could cause user to lose trust and ignore errors. For error logs to be useful they should only be emitted where there is something wrong.

We should prevent issues like https://github.com/etcd-io/etcd/issues/17245 where a change in code started generating a large number of error logs.

Proposed in https://github.com/etcd-io/etcd/pull/17249#issuecomment-1893399469 with agreement between maintainers.

nitishfy:
Hi, I'd like to take this up!

serathius:
Thanks, feel free to reach out to me on [slack](https://kubernetes.slack.com/archives/C3HD8ARJ5) ([get an invite](http://slack.kubernetes.io/)) if you need any help 
/assign @nitishfy

serathius:
https://github.com/etcd-io/etcd/pull/17423 covered a single node cluster, we should also cover:
* 3 node cluster
* different TLS configurations.

nitishfy:
> #17423 covered a single node cluster, we should also cover:
> 
> * 3 node cluster
> * different TLS configurations.

we can raise a new PR for it. WDYT?

serathius:
SG, thanks

serathius:
Also, when we are done, we should consider backporting the test.

serathius:
Still remaining to be completed:
* Extend the [existing test](https://github.com/etcd-io/etcd/pull/17423) to also validate 3 node cluster. 
* Backport the test to release-3.5 and release-3.4 branches to prevent regressions there.

ghouscht:
I think I can take this.

/assign

ghouscht:
done 🙂 I think this can be closed now.

serathius:
Awesome, thanks @ghouscht !

ahrtr:
It looks like the test doesn't enable TLS by default. Should we also extend the test to support the case of enabling TLS?

serathius:
[...]

PR #18806, opened 2024-10-29
chore(e2e): extend logging test to 3 node cluster

ghouscht:
See also https://github.com/etcd-io/etcd/issues/17329

This is how I would extend the logging test to include both a single node and a three node cluster. If it is ok like that I'd submit PRs for backporting as written in the linked issue

### DIFF
--- a/tests/e2e/logging_test.go
+++ b/tests/e2e/logging_test.go
@@ -26,39 +26,66 @@ import (
 )
 
 func TestNoErrorLogsDuringNormalOperations(t *testing.T) {
-	e2e.BeforeTest(t)
-	ctx := context.TODO()
-
-	epc, err := e2e.NewEtcdProcessCluster(ctx, t,
-		e2e.WithClusterSize(1),
-		e2e.WithLogLevel("debug"),
-	)
-	require.NoError(t, err)
-	defer epc.Close()
-
-	logs := epc.Procs[0].Logs()
-	time.Sleep(time.Second)
-	if err = epc.Close(); err != nil {
-		t.Fatalf("error closing etcd processes (%v)", err)
-	}
-	var entry logEntry
-	lines := logs.Lines()
-	if len(lines) == 0 {
-		t.Errorf("Expected at least one log line")
+	tests := []struct {
+		name          string
+		clusterSize   int
+		allowedErrors map[string]bool
+	}{
+		{
+			name:          "single node cluster",
+			clusterSize:   1,
+			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
+		},
+		{
+			name:          "three node cluster",
+			clusterSize:   3,
+			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
+		},
 	}
 
-	allowedErrors := map[string]bool{"setting up serving from embedded etcd failed.": true}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			e2e.BeforeTest(t)
+			ctx := context.TODO()
 
-	for _, line := range lines {
-		err := json.Unmarshal([]byte(line), &entry)
-		if err != nil {
-			t.Errorf("Failed to parse log line as json, err: %q, line: %s", err, line)
-			continue
-		}
-		if allowedErrors[entry.Message] {
-			continue
-		}
+			epc, err := e2e.NewEtcdProcessCluster(ctx, t,
+				e2e.WithClusterSize(tc.clusterSize),
+				e2e.WithLogLevel("debug"),
+			)
+			require.NoError(t, err)
+			defer epc.Close()
 
-		require.NotEqual(t, "error", entry.Level)
+			require.Lenf(t, epc.Procs, tc.clusterSize, "embedded etcd cluster process count is not as expected")
+
+			// Collect the handle of logs before closing the processes.
+			var logHandles []e2e.LogsExpect
+			for i := range tc.clusterSize {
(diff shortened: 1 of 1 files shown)


tests/e2e/logging_test.go
```diff
@@ -26,39 +26,60 @@ import (
...
+				e2e.WithLogLevel("debug"),
+			)
+			require.NoError(t, err)
+			defer epc.Close()
+
+			require.Lenf(t, epc.Procs, tc.clusterSize, "embedded etcd cluster process count is not as expected")
+
+			var lines []string
+			for i := range tc.clusterSize {
+				lines = append(lines, epc.Procs[i].Logs().Lines()...)
```
serathius, line 62: Think we need to collect log lines after we close the server.

tests/e2e/logging_test.go
```diff
@@ -26,39 +26,60 @@ import (
...
-	ctx := context.TODO()
+	tests := []struct {
+		name          string
+		clusterSize   int
+		allowedErrors map[string]bool
+	}{
+		{
+			name:          "single node cluster",
+			clusterSize:   1,
+			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
```
serathius, line 37: Would be good to followup why we get this log.

serathius: cc @jmhbnz @ahrtr

tests/e2e/logging_test.go
ghouscht: I did encounter the allowed error but apparently not all the time. It indeed does make a difference if `.Lines()` is called after closing. By doing so the resulting log lines seem to be constant. Will update the PR.

tests/e2e/logging_test.go
ghouscht: Yes, saw the comment in the issue. From what I saw in `release-3.5` branch this does not happen. I'll have a look where exactly it comes from.

serathius: Interesting, thanks @ghouscht !

ahrtr: The error logs are generated when closing/shutting down etcd. The listeners are closed firstly, then the accepter will return errors. They should be expected. One possible additional test case is to verify that there is no any error logs before closing etcd, which of course can be discussed & implemented separately.

```
{"level":"info","ts":"2024-10-30T08:25:35.363864Z","caller":"osutil/interrupt_unix.go:65","msg":"received signal; shutting down","signal":"interrupt"}
{"level":"info","ts":"2024-10-30T08:25:35.363960Z","caller":"embed/etcd.go:398","msg":"closing etcd server","name":"default","data-dir":"default.etcd","advertise-peer-urls":["http://localhost:2380"],"advertise-client-urls":["http://localhost:2379"]}
......
{"level":"error","ts":"2024-10-30T08:25:35.364157Z","caller":"embed/etcd.go:874","msg":"setting up serving from embedded etcd failed.","error":"http: Server closed","stacktrace":"go.etcd.io/etcd/server/v3/embed.(*Etcd).errHandler\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:874\ngo.etcd.io/etcd/server/v3/embed.(*serveCtx).serve.func3\n\tgo.etcd.io/etcd/server/v3/embed/serve.go:176"}
{"level":"error","ts":"2024-10-30T08:25:35.364250Z","caller":"embed/etcd.go:874","msg":"setting up serving from embedded etcd failed.","error":"accept tcp 127.0.0.1:2379: use of closed network connection","stacktrace":"go.etcd.io/etcd/server/v3/embed.(*Etcd).errHandler\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:874\ngo.etcd.io/etcd/server/v3/embed.(*Etcd).serveClients.func1\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:776"}
...
{"level":"error","ts":"2024-10-30T08:25:35.394822Z","caller":"embed/etcd.go:874","msg":"setting up serving from embedded etcd failed.","error":"accept tcp 127.0.0.1:2380: use of closed network connection","stacktrace":"go.etcd.io/etcd/server/v3/embed.(*Etcd).errHandler\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:874\ngo.etcd.io/etcd/server/v3/embed.(*Etcd).servePeers.func3\n\tgo.etcd.io/etcd/server/v3/embed/etcd.go:614"}
```

ghouscht: Yeah this is what I figured out as well. In that case these errors are expected and we could simply adapt the [`errorHandler`](https://github.com/ghouscht/etcd/blob/issue-17329/server/embed/etcd.go#L873-L875) function to ignore them, e.g. like this:
```go
	if err != nil &&
		!errors.Is(err, http.ErrServerClosed) &&
		!errors.Is(err, net.ErrClosed) {
		e.GetLogger().Error("setting up serving from embedded etcd failed.", zap.Error(err))
	}
```

ahrtr: It's difficult to be certain whether `http.ErrServerClose` or `net.ErrClosed` is expected by just one error itself. Suggest not to change the production code for now.

ahrtr (approved):
LGTM

Thanks

tests/e2e/logging_test.go
serathius: But I think it is. https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/net/http/server.go;l=3286-3288;drc=773767def0e0f29584a69bd760430167b7479d7d 
http.ErrServerClose means.

PR #18812, opened 2024-10-30
release: properly change working dir if tmp location already exists

ivanvc:
I identified this issue while working on #18649. Currently, if we follow the practice of testing the release script in the first step when doing a release with `DRY=true` and then rerun the script, the script won't change directories into the freshly cloned repository. Instead, it will use the current directory, which may be prone to unwanted changes accidentally being pushed as part of the release.

With this approach, running it after a DRY run will yield the error:

```
In workspace '/tmp/etcd-release-3.5.124/etcd' the branch: release-3.5 diverges from the origin.
Consider cleaning up / renaming this directory or (cd /tmp/etcd-release-3.5.124/etcd && git reset --hard origin/release-3.5)
```

However, I feel the best experience would be to replace:

```
    run git checkout "${BRANCH}" || exit 2
    run git pull origin
```

With a fetch and the hard reset to avoid manual intervention. But, I didn't want to implement it without alignment on this.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -86,10 +86,12 @@ main() {
   # Set up release directory.
   local reldir="/tmp/etcd-release-${VERSION}"
   log_callout "Preparing temporary directory: ${reldir}"
-  if [ ! -d "${reldir}/etcd" ] && [ "${IN_PLACE}" == 0 ]; then
-    mkdir -p "${reldir}"
-    cd "${reldir}"
-    run git clone "${REPOSITORY}" --branch "${BRANCH}" --depth 1
+  if [ "${IN_PLACE}" == 0 ]; then
+    if [ ! -d "${reldir}/etcd" ]; then
+      mkdir -p "${reldir}"
+      cd "${reldir}"
+      run git clone "${REPOSITORY}" --branch "${BRANCH}" --depth 1
+    fi
     run cd "${reldir}/etcd" || exit 2
     run git checkout "${BRANCH}" || exit 2
     run git pull origin


ahrtr:
> With this approach, running it after a DRY run will yield the error:
> 
> ```
> In workspace '/tmp/etcd-release-3.5.124/etcd' the branch: release-3.5 diverges from the origin.
> Consider cleaning up / renaming this directory or (cd /tmp/etcd-release-3.5.124/etcd && git reset --hard origin/release-3.5)
> ```

Probably I was missing something. Why we did not see this issue previously when we were releasing 3.5.x together?



> However, I feel the best experience would be to replace:
> 
> ```
>     run git checkout "${BRANCH}" || exit 2
>     run git pull origin
> ```
> 
> With a fetch and the hard reset to avoid manual intervention

Sounds good to me. But leave it to you and James to decide it.

ivanvc:
> Probably I was missing something. Why we did not see this issue previously when we were releasing 3.5.x together?

The issue is that instead of erroring out, it silently doesn't change the directory. The only way to catch it is by reading the log lines and noticing that it skips changing the directory. Given the number of log lines the release script generates, it is difficult to catch.

The temp directory it creates follows the pattern: `/tmp/etcd-{release_version}`. If we first run the release script with `DRY=true,` but a fictitious version (i.e., v3.5.99) as we've been doing, then running again with `DRY=false` and the actual version won't present the issue. It could be that maybe the team was aware of this limitation, and we were running it with an unreal version. cc. @jmhbnz.

/retest

ahrtr:
> The issue is that instead of erroring out, it silently doesn't change the directory. The only way to catch it is by reading the log lines and noticing that it skips changing the directory. Given the number of log lines the release script generates, it is difficult to catch.

Usually there isn't any issue as long as the tag was successfully created, because the release script will checkout the tag anyway (see code link below). But theoretically there is a potential issue if the `/tmp/etcd-release-${VERSION} ` was created but the tag wasn't created (unlikely in practice), and the release process was somehow interrupted, then when running it again, it might update the `version.go` and create tag on the wrong branch.

Thanks for the good catch!

https://github.com/etcd-io/etcd/blob/98e53e5aaaf66a718de94660266daf639d704db5/scripts/release.sh#L107-L110

ahrtr (approved):
LGTM

Thank you!

Issue #18576, opened 2024-09-11
Use `errors.Is` for error equality checks


ianlancetaylor:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

When using the current Go HEAD which will be the future Go 1.24 release, in the directory etcd-io/etcd/server, running `go test ./embed` fails.  The test error messages are:

```
--- FAIL: TestLogRotation (0.00s)
    --- FAIL: TestLogRotation/invalid_logger_config (0.00s)
        config_test.go:736: test "invalid logger config", expected error: invalid log rotation config: json: cannot unmarshal bool into Go struct field logRotationConfig.maxsize of type int, got: invalid log rotation config: json: cannot unmarshal bool into Go struct field logRotationConfig.Logger.maxsize of type int
```

### What did you expect to happen?

I expected the tests to pass.

### How can we reproduce it (as minimally and precisely as possible)?

Build Go from HEAD as outlined at https://go.dev/doc/install/source.  Use the newly built Go to run the etcd-io/server tests.

### Anything else we need to know?

In general we recommend against testing for exact error strings.  The Go standard library reserves the right to change error strings.  It's better to simply test whether an error occurred or not.  If it's necessary to check for a particular error, it's better to use `errors.Is` or `errors.As` to check the error.  If that is infeasible, it's better to use `strings.Contains` to check that the error message contains some identifying string.  Thanks.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

linux/amd64

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

ivanvc:
[...]

PR #18813, opened 2024-10-31
fix: enable errorlint in etcdctl and etcdutl directories

mmorel-35:
This enable errorlint in  etcdctl and etcdutl directories

Related to #18576


jmhbnz (approved):
LGTM - Thanks @mmorel-35

ahrtr (approved):
Thank you

Issue #18810, opened 2024-10-30
Etcd crashes when running out of space during defrag


serathius:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Etcd crashes with stacktrace
```
{"level":"info","ts":"2024-10-30T20:53:52.006800+0100","caller":"v3rpc/maintenance.go:110","msg":"starting defragment"}
{"level":"info","ts":"2024-10-30T20:53:52.008359+0100","caller":"backend/backend.go:509","msg":"defragmenting","path":"limited_disk_64MB/member/snap/db","current-db-size-bytes":2334720,"current-db-size":"2.3 MB","current-db-size-in-use-bytes":2236416,"current-db-size-in-use":"2.2 MB"}
{"level":"warn","ts":"2024-10-30T20:53:52.024275+0100","caller":"v3rpc/maintenance.go:115","msg":"failed to defragment","error":"write limited_disk_64MB/member/snap/db.tmp.172848138: no space left on device"}
{"level":"info","ts":"2024-10-30T20:53:52.024402+0100","caller":"v3rpc/health.go:63","msg":"grpc service status changed","service":"","status":"SERVING"}
{"level":"warn","ts":"2024-10-30T20:53:52.025202+0100","logger":"etcd-client","caller":"v3@v3.6.0-alpha.0/retry_interceptor.go:65","msg":"retrying of unary invoker failed","target":"etcd-endpoints://0xc00022c5a0/127.0.0.1:2379","method":"/etcdserverpb.Maintenance/Defragment","attempt":0,"error":"rpc error: code = Unknown desc = write limited_disk_64MB/member/snap/db.tmp.172848138: no space left on device"}
Failed to defragment etcd member[127.0.0.1:2379]. took 26.406294ms. (rpc error: code = Unknown desc = write limited_disk_64MB/member/snap/db.tmp.172848138: no space left on device)

[...]

PR #18822, opened 2024-11-01
fix(defrag): handle no space left error

ghouscht:
PR contains an e2e test, gofailpoint and a fix for the issue described in https://github.com/etcd-io/etcd/issues/18810. 

Without the fix the test triggers a nil ptr panic in etcd as described in the linked issue:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: execute job failed
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x787ad8]

goroutine 136 [running]:
go.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x80?, 0x2?, {0xf?, 0x0?, 0x0?})
	go.uber.org/zap@v1.27.0/zapcore/entry.go:196 +0x78
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0x40005d81a0, {0x400058a780, 0x2, 0x2})
	go.uber.org/zap@v1.27.0/zapcore/entry.go:262 +0x1c4
go.uber.org/zap.(*Logger).Panic(0xcd8746?, {0xce7440?, 0xb54f60?}, {0x400058a780, 0x2, 0x2})
	go.uber.org/zap@v1.27.0/logger.go:285 +0x54
go.etcd.io/etcd/pkg/v3/schedule.(*fifo).executeJob.func1()
	go.etcd.io/etcd/pkg/v3@v3.6.0-alpha.0/schedule/schedule.go:202 +0x24c
panic({0xb54f60?, 0x1681ec0?})
	runtime/panic.go:785 +0x124
go.etcd.io/bbolt.(*Tx).Bucket(...)
	go.etcd.io/bbolt@v1.4.0-alpha.1/tx.go:112
go.etcd.io/etcd/server/v3/storage/backend.unsafeForEach(0x0, {0xe9b988?, 0x1695e60?}, 0x4000580460)
	go.etcd.io/etcd/server/v3/storage/backend/batch_tx.go:235 +0x38
go.etcd.io/etcd/server/v3/storage/backend.(*batchTx).UnsafeForEach(...)
	go.etcd.io/etcd/server/v3/storage/backend/batch_tx.go:231
go.etcd.io/etcd/server/v3/storage/backend.unsafeVerifyTxConsistency({0xea55e0, 0x40000bafc0}, {0xe9b988, 0x1695e60})
	go.etcd.io/etcd/server/v3/storage/backend/verify.go:97 +0xa4
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).applyAll.VerifyBackendConsistency.func1()
	go.etcd.io/etcd/server/v3/storage/backend/verify.go:90 +0x244
go.etcd.io/etcd/client/pkg/v3/verify.Verify(0x40000efdf8)
	go.etcd.io/etcd/client/pkg/v3@v3.6.0-alpha.0/verify/verify.go:71 +0x3c
go.etcd.io/etcd/server/v3/storage/backend.VerifyBackendConsistency(...)
	go.etcd.io/etcd/server/v3/storage/backend/verify.go:75
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).applyAll(0x40003a3c08, 0x4000172180, 0x40005da000)
	go.etcd.io/etcd/server/v3/etcdserver/server.go:972 +0xcc
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).run.func6({0x4000196850?, 0x4000494a80?})
	go.etcd.io/etcd/server/v3/etcdserver/server.go:847 +0x28
go.etcd.io/etcd/pkg/v3/schedule.job.Do(...)
	go.etcd.io/etcd/pkg/v3@v3.6.0-alpha.0/schedule/schedule.go:41
go.etcd.io/etcd/pkg/v3/schedule.(*fifo).executeJob(0x40000eff70?, {0xe943f8?, 0x40005b6330?}, 0x0?)
	go.etcd.io/etcd/pkg/v3@v3.6.0-alpha.0/schedule/schedule.go:206 +0x78
go.etcd.io/etcd/pkg/v3/schedule.(*fifo).run(0x400039a0e0)
	go.etcd.io/etcd/pkg/v3@v3.6.0-alpha.0/schedule/schedule.go:187 +0x15c
created by go.etcd.io/etcd/pkg/v3/schedule.NewFIFOScheduler in goroutine 155
	go.etcd.io/etcd/pkg/v3@v3.6.0-alpha.0/schedule/schedule.go:101 +0x178
```

~~I think from here on we can discuss potential solutions for the problem. @ahrtr already suggested two possible options in the linked issue.~~

As mentioned in https://github.com/etcd-io/etcd/pull/18822#issuecomment-2453561758 the PR now restores the environment and lets etcd continue to run.

### DIFF
--- a/server/storage/backend/backend.go
+++ b/server/storage/backend/backend.go
@@ -477,10 +477,6 @@ func (b *backend) defrag() error {
 	b.readTx.Lock()
 	defer b.readTx.Unlock()
 
-	b.batchTx.unsafeCommit(true)
-
-	b.batchTx.tx = nil
-
 	// Create a temporary file to ensure we start with a clean slate.
 	// Snapshotter.cleanupSnapdir cleans up any of these that are found during startup.
 	dir := filepath.Dir(b.db.Path())
@@ -488,11 +484,14 @@ func (b *backend) defrag() error {
 	if err != nil {
 		return err
 	}
+
 	options := bolt.Options{}
 	if boltOpenOptions != nil {
 		options = *boltOpenOptions
 	}
 	options.OpenFile = func(_ string, _ int, _ os.FileMode) (file *os.File, err error) {
+		// gofail: var defragNoSpace string
+		// return nil, fmt.Errorf(defragNoSpace)
 		return temp, nil
 	}
 	// Don't load tmp db into memory regardless of opening options
@@ -515,6 +514,11 @@ func (b *backend) defrag() error {
 			zap.String("current-db-size-in-use", humanize.Bytes(uint64(sizeInUse1))),
 		)
 	}
+
+	// Commit/stop and then reset current transactions (including the readTx)
+	b.batchTx.unsafeCommit(true)
+	b.batchTx.tx = nil
+
 	// gofail: var defragBeforeCopy struct{}
 	err = defragdb(b.db, tmpdb, defragLimit)
 	if err != nil {
--- a/server/storage/backend/backend.go
+++ b/server/storage/backend/backend.go
@@ -490,8 +490,8 @@ func (b *backend) defrag() error {
 		options = *boltOpenOptions
 	}
 	options.OpenFile = func(_ string, _ int, _ os.FileMode) (file *os.File, err error) {
-		// gofail: var defragNoSpace string
-		// return nil, fmt.Errorf(defragNoSpace)
+		// gofail: var defragOpenFileError string
+		// return nil, fmt.Errorf(defragOpenFileError)
 		return temp, nil
 	}
 	// Don't load tmp db into memory regardless of opening options
@@ -526,6 +526,11 @@ func (b *backend) defrag() error {
 		if rmErr := os.RemoveAll(tmpdb.Path()); rmErr != nil {
 			b.lg.Error("failed to remove db.tmp after defragmentation completed", zap.Error(rmErr))
 		}
+
+		// restore the bbolt transactions if defragmentation fails
+		b.batchTx.tx = b.unsafeBegin(true)
+		b.readTx.tx = b.unsafeBegin(false)
+
 		return err
 	}
 
@@ -578,6 +583,9 @@ func (b *backend) defrag() error {
 }
 
 func defragdb(odb, tmpdb *bolt.DB, limit int) error {
+	// gofail: var defragdbFail string
+	// return fmt.Errorf(defragdbFail)
+
 	// open a tx on tmpdb for writes
 	tmptx, err := tmpdb.Begin(true)
 	if err != nil {
--- a/tests/e2e/defrag_no_space_test.go
+++ b/tests/e2e/defrag_no_space_test.go
@@ -1,0 +1,49 @@
+// Copyright 2024 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/tests/v3/framework/config"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+func TestDefragNoSpace(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	clus, err := e2e.NewEtcdProcessCluster(context.TODO(), t,
+		e2e.WithClusterSize(1),
+		e2e.WithGoFailEnabled(true),
+	)
+	require.NoError(t, err)
+	t.Cleanup(func() { clus.Stop() })
+
+	member := clus.Procs[0]
+
+	require.NoError(t, member.Failpoints().SetupHTTP(context.Background(), "defragNoSpace", `return("no space")`))
+	require.ErrorContains(t, member.Etcdctl().Defragment(context.Background(), config.DefragOption{Timeout: time.Minute}), "no space")
+
+	// Make sure etcd continues to run even after the failed defrag attempt
+	require.NoError(t, member.Etcdctl().Put(context.Background(), "foo", "bar", config.PutOptions{}))
+	value, err := member.Etcdctl().Get(context.Background(), "foo", config.GetOptions{})
+	require.NoError(t, err)
+	require.Len(t, value.Kvs, 1)
+	require.Equal(t, "bar", string(value.Kvs[0].Value))
+}
--- a/tests/e2e/defrag_no_space_test.go
+++ b/tests/e2e/defrag_no_space_test.go
@@ -16,6 +16,7 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


ahrtr:
The e2e test looks good.

The proposed solution is to restore the environment (i.e. reopen the bbolt) when defragmentation somehow fails and panicking if the restoring fails again. If the bbols fails to be opened, then etcdserver can't serve any requests, so it makes sense to panic it. cc @fuweid  @ivanvc  @jmhbnz  @serathius @tjungblu

ghouscht:
> The e2e test looks good.
> 
> The proposed solution is to restore the environment (i.e. reopen the bbolt) when defragmentation somehow fails and panicking if the restoring fails again. If the bbols fails to be opened, then etcdserver can't serve any requests, so it makes sense to panic it. cc @fuweid @ivanvc @jmhbnz @serathius @tjungblu

I added a second commit that contains a working implementation of a possible restore operation. I did some manual testing with the failpoint and the e2e test and it seems to work. However this opens up a whole lot of other possible problems. I highlighted some of them with `TODO` in the code - feedback appreciated 🙂

server/storage/backend/backend.go
```diff
@@ -455,7 +456,61 @@ func (b *backend) Commits() int64 {
 }
 
 func (b *backend) Defrag() error {
-	return b.defrag()
+	err := b.defrag()
+	if err != nil {
```
ghouscht, line 460: Instead of doing the restore here we could probably place it inside a `defer` in `defrag()` function? However, if we do so we need to be a bit careful about releasing the locks -> order of defered function execution is important in that case.

server/storage/backend/backend.go
```diff
@@ -455,7 +456,61 @@ func (b *backend) Commits() int64 {
...
+		// lock database after lock tx to avoid deadlock.
+		b.mu.Lock()
+		defer b.mu.Unlock()
+
+		// block concurrent read requests while resetting tx
+		b.readTx.Lock()
+		defer b.readTx.Unlock()
+
+		// TODO: Is there a way to figure out if the bbolt db is closed or open?
+		// Haven't found one so far, thus I'm using b.db.Begin() and then checking the error it returns.
```
ghouscht, line 481: With the current failpoint `defragNoSpace` the bbolt db will never be closed. Because all other filesystem operations (e.g. the rename of the temp to new db) seem to call a `Fatal` log function it should be safe to assume that the db is still open if we try to restore? If so this would simplify the restore operation a lot and we could avoid trying to figure out if the db is still open or not.

server/storage/backend/backend.go
serathius: Not sure why do generic error handling here.
Why not identify where `b.defrag()` generates error and then improve error handler there?

ghouscht: Sure that is an option, I just thought you're looking for a generic error handling.

I already know where the error happens so we can improve that. I'll add another commit.

server/storage/backend/backend.go
ghouscht: no longer relevant

server/storage/backend/backend.go
ghouscht: Added 1fb20649b2eace00c9f17f712d5019b5917805de which addresses your comment and only handles the specific error. The e2e test is adapted as well. Manual testing and the e2e test seem to confirm that this is working.

serathius: Looks good, can you confirm that the test reproduces the bug on the old version?

server/storage/backend/backend.go
```diff
@@ -477,29 +476,38 @@ func (b *backend) defrag() error {
...
 	dir := filepath.Dir(b.db.Path())
 	temp, err := os.CreateTemp(dir, "db.tmp.*")
 	if err != nil {
 		return err
 	}
+	defer temp.Close()
+
+	// Commit/stop and then reset current transactions (including the readTx)
+	b.batchTx.unsafeCommit(true)
+	b.batchTx.tx = nil
```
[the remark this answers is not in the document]
serathius: Makes sense

serathius (approved):

server/storage/backend/backend.go
ghouscht: Yes it does, the test triggers a panic in etcd I just verified that.

server/storage/backend/backend.go
```diff
@@ -477,29 +477,38 @@ func (b *backend) defrag() error {
...
-	b.batchTx.tx = nil
-
 	// Create a temporary file to ensure we start with a clean slate.
 	// Snapshotter.cleanupSnapdir cleans up any of these that are found during startup.
 	dir := filepath.Dir(b.db.Path())
 	temp, err := os.CreateTemp(dir, "db.tmp.*")
 	if err != nil {
 		return err
 	}
+	defer temp.Close()
```
ahrtr, line 487: Note that the temp file will be renamed after the defragmentation operation. I think we should close the temp file right before renaming. On Linux, it isn't a problem to rename a file when it's still open; but it's a problem on some other platform, i.e. Windows (although it's tier 3 for now).

https://github.com/etcd-io/etcd/blob/bd8896364a1d9adb554b538232004e5b2d6c850a/server/storage/backend/backend.go#L537

I agree that this `defer temp.Close()` still makes sense, but we need to guarantee it's only called once using `sync.Once`, although it isn't a big deal to call it multiple times.

So suggest to revert this change and resolve it in a separate PR.

server/storage/backend/backend.go
```diff
@@ -477,29 +477,38 @@ func (b *backend) defrag() error {
...
+
+	// Commit/stop and then reset current transactions (including the readTx)
+	b.batchTx.unsafeCommit(true)
+	b.batchTx.tx = nil
```
ahrtr, line 491: We can even move these two lines to right before `defragdb`. i.e. if we fail to open the temp bbolt db, we don't need to reopen the existing working bbolt db.

https://github.com/etcd-io/etcd/blob/bd8896364a1d9adb554b538232004e5b2d6c850a/server/storage/backend/backend.go#L519

ghouscht: Of course, yes 🙂, will move it.

server/storage/backend/backend.go
ghouscht: Oh good catch, I think that slipped in from my testing, I didn't want to commit this honestly.

server/storage/backend/backend.go
ghouscht: Now in case the `defragdb` fails and returns an error we could potentially endup in the same state as described in the issue (`b.batchTx.tx == nil` and nil ptr panic). Should this error be handled or is this something were etcd should stop e.g. panic?

ahrtr:
I think we still need to handle the error if any during `defragdb`, something like below, also add one more failpoint and a new [sub]test case.

```
$ git diff -l10
diff --git a/server/storage/backend/backend.go b/server/storage/backend/backend.go
index 95f5cf96f..5a9361ae8 100644
--- a/server/storage/backend/backend.go
+++ b/server/storage/backend/backend.go
@@ -522,6 +522,11 @@ func (b *backend) defrag() error {
                if rmErr := os.RemoveAll(tmpdb.Path()); rmErr != nil {
                        b.lg.Error("failed to remove db.tmp after defragmentation completed", zap.Error(rmErr))
                }
+
+               // restore the bbolt transactions if defragmentation fails
+               b.batchTx.tx = b.unsafeBegin(true)
+               b.readTx.tx = b.unsafeBegin(false)
+
                return err
        }
```

https://github.com/etcd-io/etcd/blob/bd8896364a1d9adb554b538232004e5b2d6c850a/server/storage/backend/backend.go#L520-L526

server/storage/backend/backend.go
ahrtr: Please see https://github.com/etcd-io/etcd/pull/18822#issuecomment-2459209181

ahrtr:
Note we need to resolve https://github.com/etcd-io/etcd/pull/18822#discussion_r1830692111 in a separate PR. Could you please raise a new issue to track it? Thanks.

ghouscht:
> Note we need to resolve [#18822 (comment)](https://github.com/etcd-io/etcd/pull/18822#discussion_r1830692111) in a separate PR. Could you please raise a new issue to track it? Thanks.

https://github.com/etcd-io/etcd/issues/18841

server/storage/backend/backend.go
ghouscht: I added a second commit for that, please let me know if that is ok.

server/storage/backend/backend.go
```diff
@@ -477,22 +477,21 @@ func (b *backend) defrag() error {
...
 		return err
 	}
+
 	options := bolt.Options{}
 	if boltOpenOptions != nil {
 		options = *boltOpenOptions
 	}
 	options.OpenFile = func(_ string, _ int, _ os.FileMode) (file *os.File, err error) {
+		// gofail: var defragNoSpace string
+		// return nil, fmt.Errorf(defragNoSpace)
```
ahrtr, line 494:
```suggestion
		// gofail: var defragOpenFileError string
		// return nil, fmt.Errorf(defragOpenFileError)
```

ahrtr:
Overall looks good now. Please signoff the second commit. Refer to https://github.com/etcd-io/etcd/pull/18822/checks?check_run_id=32589384806

ahrtr (approved):
LGTM

Thank you!

ahrtr:
@ghouscht can you please backport this PR to 3.5 and 3.4?

ahrtr:
@ghouscht Please also update the changelog for [3.5](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.5.md#v3517-tbc) and [3.4](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.4.md#v3435-tbc). Thanks

server/storage/backend/backend.go
ahrtr: My previous comment isn't correct, bbolt will close the file when closing bbolt instance. The file should have already been closed before renaming the file.

https://github.com/etcd-io/bbolt/blob/c2e4c77280fc2f47870995aadd6e15c6422c5af0/db.go#L712

PR #18823, opened 2024-11-01
fix: use require instead of t.Fatal(err) in tests/common package

mmorel-35:


ahrtr (approved):
LGTM

Thanks

ivanvc (approved):
LGTM. Thanks, @mmorel-35.

PR #18825, opened 2024-11-02
Run a separate in memory snapshot to reduce number of entries stored in raft memory storage

serathius:

Alternative to https://github.com/etcd-io/etcd/pull/18635

Goal: Reduce number of raft entries stored in memory

Context:
* Etcd maintains 2 different independently updates storages. V2 - stored as snapshots every X entries (100`000 by default), V3 - memory mapped bbolt storage, flushed every 5 seconds to disk
* Etcd maintains guarantee for the data stored on disk. V3 storage needs to be always fresher than V2 snapshot. Meaning V3 consistent index is >= index of last snapshot.
* This guarantee requires that before writing any V2 snapshots etcd needs to flush V3 storage which is costly.
* In etcd v3.6 no longer uses v2 storage as source of truth, however it maintains periodic snapshot generation for backward compatibility.
* Raft in memory store needs to store last snapshot and all entries since. However, this snapshot doesn't need to be persisted to disk.

Proposal:
* Separate disk and memory snapshotting.
* Disk snapshotting will be used for backward compatibility. Done every --snapshot-count. 
* Memory snapshots would no longer be bound by disk overhead.

Benchmark results `./bin/tools/benchmark put --total=15000 --val-size=100000`

| Scenario           | Snapshot count | Memory snapshot count | CPU[%] | Memory[MB] | PUTs per second |
| ------------------ | -------------- | --------------------- | ------ | ---------- | --------------- |
| Base               | 10000          | n/a                   | 49.4   | 1954.5     | 1029.2          |
| Snapshot in memory | 10000          | 1                     | 51.1   | 1330.4     | 1006.1          |
| Snapshot in memory | 10000          | 10                    | 48.9   | 1264.2     | 1046.3          |
| Snapshot in memory | 10000          | 100                   | 48.1   | 1210.4     | 1051.8          |
| Snapshot in memory | 10000          | 1000                  | 48.2   | 1360.0     | 1047.2          |
| Snapshot in memory | 10000          | 10000                 | 49.4   | 2040.3     | 1027.4          |

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -899,7 +899,7 @@ func (c *RaftCluster) Store(store v2store.Store) {
 		if m.ClientURLs != nil {
 			mustUpdateMemberAttrInStore(c.lg, store, m)
 		}
-		c.lg.Info(
+		c.lg.Debug(
 			"snapshot storing member",
 			zap.String("id", m.ID.String()),
 			zap.Strings("peer-urls", m.PeerURLs),
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -109,6 +109,7 @@ const (
 	readyPercentThreshold = 0.9
 
 	DowngradeEnabledPath = "/downgrade/enabled"
+	memorySnapshotCount  = 100
 )
 
 var (
@@ -293,6 +294,7 @@ type EtcdServer struct {
 	*AccessController
 	// forceDiskSnapshot can force snapshot be triggered after apply, independent of the snapshotCount.
 	// Should only be set within apply code path. Used to force snapshot after cluster version downgrade.
+	// TODO: Replace with flush db in v3.7 assuming v3.6 bootstraps from db file.
 	forceDiskSnapshot bool
 	corruptionChecker CorruptionChecker
 }
@@ -1195,15 +1197,22 @@ func (s *EtcdServer) ForceSnapshot() {
 }
 
 func (s *EtcdServer) snapshotIfNeededAndCompactRaftLog(ep *etcdProgress) {
-	if !s.shouldSnapshot(ep) {
+	//TODO: Remove disk snapshot in v3.7
+	shouldSnapshotToDisk := s.shouldSnapshotToDisk(ep)
+	shouldSnapshotToMemory := s.shouldSnapshotToMemory(ep)
+	if !shouldSnapshotToDisk && !shouldSnapshotToMemory {
 		return
 	}
-	s.snapshot(ep)
+	s.snapshot(ep, shouldSnapshotToDisk)
 	s.compactRaftLog(ep.appliedi)
 }
 
-func (s *EtcdServer) shouldSnapshot(ep *etcdProgress) bool {
+func (s *EtcdServer) shouldSnapshotToDisk(ep *etcdProgress) bool {
 	return (s.forceDiskSnapshot && ep.appliedi != ep.diskSnapshotIndex) || (ep.appliedi-ep.diskSnapshotIndex > s.Cfg.SnapshotCount)
+}
+
+func (s *EtcdServer) shouldSnapshotToMemory(ep *etcdProgress) bool {
+	return ep.appliedi > ep.memorySnapshotIndex+memorySnapshotCount
 }
 
 func (s *EtcdServer) hasMultipleVotingMembers() bool {
@@ -2119,28 +2128,30 @@ func (s *EtcdServer) applyConfChange(cc raftpb.ConfChange, confState *raftpb.Con
 }
 
 // TODO: non-blocking snapshot
-func (s *EtcdServer) snapshot(ep *etcdProgress) {
+func (s *EtcdServer) snapshot(ep *etcdProgress, toDisk bool) {
 	lg := s.Logger()
-	lg.Info(
-		"triggering snapshot",
-		zap.String("local-member-id", s.MemberID().String()),
-		zap.Uint64("local-member-applied-index", ep.appliedi),
-		zap.Uint64("local-member-snapshot-index", ep.diskSnapshotIndex),
-		zap.Uint64("local-member-snapshot-count", s.Cfg.SnapshotCount),
-		zap.Bool("snapshot-forced", s.forceDiskSnapshot),
-	)
-	s.forceDiskSnapshot = false
-	d := GetMembershipInfoInV2Format(s.Logger(), s.cluster)
-	// commit kv to write metadata (for example: consistent index) to disk.
-	//
-	// This guarantees that Backend's consistent_index is >= index of last snapshot.
-	//
-	// KV().commit() updates the consistent index in backend.
-	// All operations that update consistent index must be called sequentially
-	// from applyAll function.
-	// So KV().Commit() cannot run in parallel with toApply. It has to be called outside
-	// the go routine created below.
-	s.KV().Commit()
+	d := GetMembershipInfoInV2Format(lg, s.cluster)
+	if toDisk {
+		s.Logger().Info(
+			"triggering snapshot",
+			zap.String("local-member-id", s.MemberID().String()),
+			zap.Uint64("local-member-applied-index", ep.appliedi),
+			zap.Uint64("local-member-snapshot-index", ep.diskSnapshotIndex),
+			zap.Uint64("local-member-snapshot-count", s.Cfg.SnapshotCount),
+			zap.Bool("snapshot-forced", s.forceDiskSnapshot),
+		)
+		s.forceDiskSnapshot = false
+		// commit kv to write metadata (for example: consistent index) to disk.
+		//
+		// This guarantees that Backend's consistent_index is >= index of last snapshot.
+		//
+		// KV().commit() updates the consistent index in backend.
+		// All operations that update consistent index must be called sequentially
+		// from applyAll function.
+		// So KV().Commit() cannot run in parallel with toApply. It has to be called outside
+		// the go routine created below.
+		s.KV().Commit()
+	}
 
 	// For backward compatibility, generate v2 snapshot from v3 state.
 	snap, err := s.r.raftStorage.CreateSnapshot(ep.appliedi, &ep.confState, d)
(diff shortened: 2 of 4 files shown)


server/etcdserver/api/membership/cluster.go
```diff
@@ -899,12 +899,6 @@ func (c *RaftCluster) Store(store v2store.Store) {
 		if m.ClientURLs != nil {
 			mustUpdateMemberAttrInStore(c.lg, store, m)
 		}
-		c.lg.Info(
-			"snapshot storing member",
-			zap.String("id", m.ID.String()),
-			zap.Strings("peer-urls", m.PeerURLs),
-			zap.Bool("is-learner", m.IsLearner),
-		)
```
ahrtr, line 907: why remove this log ?

server/etcdserver/server.go
```diff
@@ -2202,10 +2231,6 @@ func (s *EtcdServer) compactRaftLog(snapi uint64) {
 		}
 		lg.Panic("failed to compact", zap.Error(err))
 	}
-	lg.Info(
-		"compacted Raft logs",
-		zap.Uint64("compact-index", compacti),
-	)
```
ahrtr, line 2208: why remove this log?

server/etcdserver/server.go
```diff
@@ -2188,11 +2218,10 @@ func (s *EtcdServer) compactRaftLog(snapi uint64) {
 	}
 
 	// keep some in memory log entries for slow followers.
-	compacti := uint64(1)
-	if snapi > s.Cfg.SnapshotCatchUpEntries {
-		compacti = snapi - s.Cfg.SnapshotCatchUpEntries
+	if snapi <= s.Cfg.SnapshotCatchUpEntries {
+		return
 	}
-
```
ahrtr, line 2195: This seems to be unrelated to this PR, can we do it in a separate small PR?

server/etcdserver/api/membership/cluster.go
serathius: How happens too frequently and spams logs

server/etcdserver/server.go
serathius: Now happens too frequently and spams logs

server/etcdserver/server.go
serathius: Hmm, compactLog was causing some issue with when called too frequently. Need to double check.

serathius:
ping @ahrtr

server/etcdserver/server.go
```diff
@@ -109,6 +109,7 @@ const (
 	readyPercentThreshold = 0.9
 
 	DowngradeEnabledPath = "/downgrade/enabled"
+	memorySnapshotCount  = 100
```
ahrtr, line 112: Should we add a config item (i.e. `--compact-raft-log-interval-indices`) for this?  We don't have a feature flag for this PR. If users set the the same value as `--snapshot-count`, then it's similar to disable this feature.

server/etcdserver/server.go
```diff
@@ -1188,31 +1193,37 @@ func (s *EtcdServer) applyEntries(ep *etcdProgress, apply *toApply) {
...
+	s.forceDiskSnapshot = true
 }
 
 func (s *EtcdServer) triggerSnapshot(ep *etcdProgress) {
-	if !s.shouldSnapshot(ep) {
+	if !s.shouldSnapshotToDisk(ep) {
+		if ep.appliedi > ep.memorySnapshotIndex+memorySnapshotCount {
+			s.snapshotToMemory(ep.appliedi, ep.confState)
+			s.compactRaftLog(ep.appliedi)
+			ep.memorySnapshotIndex = ep.appliedi
```
ahrtr, line 1204: We should update the `ep.memorySnapshotIndex` when creating the disk v2 snapshot file as well.

The following three operations are supposed to be always executed together, so it might be better to get them included in one method/function,
- create raft snapshot
- compact raft log
- update `ep.memorySnapshotIndex`

ahrtr:
Overall looks good to me. 

We don't keep up to 100K (--snapshot-count) + 5K (catch-up-entries) raft log entries any more, instead, we only keep at most 100 (--compact-raft-log-interval-indices) + 5K.

server/etcdserver/server.go
serathius: I think it's good to discuss whether a configuration option would be useful for user, however we cannot expose each numeric const as option, we don't want etcd options to feel like atomic plant control panel. 

For each option we should configure whether and how user will use it and whether it makes sense to support it long term. In this case I don't see reason to expose it as there is no tradeoff happening here.  It's not like snapshot count, which I expect dates to etcd v2 where you trade off performance vs time to recovery. Here it's always performance as time to recovery will be based on last snapshot.

ahrtr: Actually I am not too much concerned about this option. I was thinking memory saving vs CPU increasing/throughput reducing might be a tradeoff, but based on your test result, it seems that there isn't too much impact on CPU & throughput.

The main point is that we changed the internal behaviour of etcdserver, but there is no way to disable/revert the behaviour in case of any regression.

Anyway, I am OK to keep it as it's for now.

server/etcdserver/server.go
serathius: Done

server/etcdserver/api/membership/cluster.go
serathius: Done

server/etcdserver/server.go
serathius: > The main point is that we changed the internal behaviour of etcdserver, but there is no way to disable/revert the behaviour in case of any regression.

For such cases we should consider using feature gates, the downside is that it will require keeping old code for multiple releases. I would consider taking an exception for v2storage clenaup, which I consider this change as it based on removal of snapshots. 

Post v3.7 we might consider feature gating such changes.

ahrtr: > I would consider taking an exception for v2storage clenaup, which I consider this change as it based on removal of snapshots.

OK.

server/etcdserver/server.go
serathius: Removed it, it was causing problems during benchmarks when I set the memorySnapshotCount <5, let's see if it passes tests now.

server/etcdserver/server.go
```diff
@@ -2128,23 +2131,32 @@ func (s *EtcdServer) applyConfChange(cc raftpb.ConfChange, confState *raftpb.Con
...
 	lg := s.Logger()
+	d := GetMembershipInfoInV2Format(lg, s.cluster)
+	if toDisk {
+		s.Logger().Info(
+			"triggering snapshot",
+			zap.String("local-member-id", s.MemberID().String()),
+			zap.Uint64("local-member-applied-index", ep.appliedi),
+			zap.Uint64("local-member-snapshot-index", ep.diskSnapshotIndex),
+			zap.Uint64("local-member-snapshot-count", s.Cfg.SnapshotCount),
+			zap.Bool("snapshot-forced", s.forceDiskSnapshot),
```
ahrtr, line 2144: You already set `s.forceDiskSnapshot` to `false` at line 1207, so it will always print `false` here.  So set `s.forceDiskSnapshot` to `false` after printing this log?

server/etcdserver/server.go
```diff
@@ -2128,23 +2131,32 @@ func (s *EtcdServer) applyConfChange(cc raftpb.ConfChange, confState *raftpb.Con
...
+		//
+		// This guarantees that Backend's consistent_index is >= index of last snapshot.
+		//
+		// KV().commit() updates the consistent index in backend.
+		// All operations that update consistent index must be called sequentially
+		// from applyAll function.
+		// So KV().Commit() cannot run in parallel with toApply. It has to be called outside
+		// the go routine created below.
+		s.KV().Commit()
+	}
```
ahrtr, line 2156: Minor comment
```suggestion
	} else {
	    // generating a debug log for triggering memory snapshot?
	}
```

server/etcdserver/server.go
```diff
@@ -2153,21 +2165,25 @@ func (s *EtcdServer) snapshot(snapi uint64, confState raftpb.ConfState) {
...
 
-	lg.Info(
-		"saved snapshot",
-		zap.Uint64("snapshot-index", snap.Metadata.Index),
-	)
+		lg.Info(
+			"saved snapshot to disk",
+			zap.Uint64("snapshot-index", snap.Metadata.Index),
+		)
+	}
```
ahrtr, line 2186: Minor comment
```suggestion
	} else {
	    // generating a debug log for triggering memory snapshot?
	}	
```

server/etcdserver/server_test.go
```diff
@@ -679,12 +679,76 @@ func TestSnapshot(t *testing.T) {
...
 	<-ch
 	if len(st.Action()) != 0 {
 		t.Errorf("no action expected on v2store. Got %d actions", len(st.Action()))
 	}
+	if ep.diskSnapshotIndex != 0 {
+		t.Errorf("ep.diskSnapshotIndex = %d, want 0", ep.diskSnapshotIndex)
+	}
+	if ep.memorySnapshotIndex != 1 {
+		t.Errorf("ep.memorySnapshotIndex = %d, want 1", ep.memorySnapshotIndex)
+	}
```
ahrtr, line 751: This seems will break the linter ? @mmorel-35 is it recommended to use `require` package?

server/etcdserver/server_test.go
```diff
@@ -679,12 +679,76 @@ func TestSnapshot(t *testing.T) {
 			t.Errorf("action = %s, want Release", gaction[1])
 		}
 	}()
+	ep := etcdProgress{appliedi: 1, confState: raftpb.ConfState{Voters: []uint64{1}}}
+	srv.snapshot(&ep, true)
+	<-ch
+	if len(st.Action()) != 0 {
+		t.Errorf("no action expected on v2store. Got %d actions", len(st.Action()))
```
ahrtr, line 686: cc @mmorel-35

use `require`?

server/etcdserver/server_test.go
ahrtr: I vaguely recall there is a rule that it doesn't make sense to continue the test if already failed. So require is recommended instead of assert?

server/etcdserver/server.go
serathius: Done

server/etcdserver/server.go
serathius: Memory snapshot already generates debug log, don't see reason to add more logging here.

server/etcdserver/server.go
serathius: Don't se a reason to add more logging.

ahrtr (approved):

serathius:
image
Could care less about coverage going down by 0.08%. @jmhbnz can we disable this?

jmhbnz:
> Could care less about coverage going down by 0.08%. @jmhbnz can we disable this?

We can tune it yeah, opened https://github.com/etcd-io/etcd/pull/18964

Issue #18841, opened 2024-11-06
make sure temp file(s) are closed during defrag


ghouscht:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Some temp file(s) (https://github.com/etcd-io/etcd/blob/bd8896364a1d9adb554b538232004e5b2d6c850a/server/storage/backend/backend.go#L487) might not be closed during defrag. See discussion https://github.com/etcd-io/etcd/pull/18822#discussion_r1830692111 for more details.

ghouscht:
/assign

Will take care of this.

ghouscht:
Apparently if everything works as expected the temp file is closed:

- `tmpdb.Close` (*bbolt.DB) is called here: https://github.com/etcd-io/etcd/blob/main/server/storage/backend/backend.go#L532 (which also closes the file descriptor of the db file)
- because of https://github.com/etcd-io/etcd/blob/main/server/storage/backend/backend.go#L495-L497 the temporary file isn't opened again, so calling `tmpdb.Close` also closes the `temp` file descriptor

Probably in one of the error cases this doesn't happen. Need to check.

ghouscht:
This is done. All PRs, including the ones for the backports to 3.4 and 3.5 have been merged.

PR #18851, opened 2024-11-07
fix(defrag): close temp file in case of error

ghouscht:
See issue https://github.com/etcd-io/etcd/issues/18841 and the discussion https://github.com/etcd-io/etcd/pull/18822#discussion_r1830692111 for details.

As written in the issue the temp file is closed if the defragmentation works as expected. The only case where it is left open is when the `bolt.Open` call fails. I think it is enough to simply include the close there, in my opinion there is no need to complicate this more.

### DIFF
--- a/server/storage/backend/backend.go
+++ b/server/storage/backend/backend.go
@@ -499,6 +499,15 @@ func (b *backend) defrag() error {
 	tdbp := temp.Name()
 	tmpdb, err := bolt.Open(tdbp, 0600, &options)
 	if err != nil {
+		temp.Close()
+		if err := os.Remove(temp.Name()); err != nil && b.lg != nil {
+			b.lg.Error(
+				"failed to remove temporary file",
+				zap.String("path", temp.Name()),
+				zap.Error(err),
+			)
+		}
+
 		return err
 	}
 


server/storage/backend/backend.go
```diff
@@ -499,6 +499,7 @@ func (b *backend) defrag() error {
 	tdbp := temp.Name()
 	tmpdb, err := bolt.Open(tdbp, 0600, &options)
 	if err != nil {
+		temp.Close()
```
ahrtr, line 502: we also need to remove the temp file, refer to below https://github.com/etcd-io/etcd/blob/4ad92611942a842c06324ae2da74ead48db79ee9/server/storage/backend/backend.go#L526

ghouscht: Is this really needed? etcd anyway removes the `db.tmp.*` files at startup.

ahrtr: It's the tmp file instead of db.tmp.*

ghouscht: So should I simply add a `os.Remove(temp.Name())` after the call to `Close`?

ahrtr: Yes, I think so. Please also log an error message if the removing fails

ghouscht: I added something, let me know what you think. Is there also a need to write a unit test for this?

In my opinion the error log doesn't really provide any value to an etcd user so personally I would omit it. As a user I do not really care if there is an old temp file lying around (that will be deleted on next etcd restart anyway) that is why I think it is not needed. However, I'm fine with adding it if you think it is needed 🙂

fuweid: Is it possible to use defer here for cleanup? I saw the line 525 does the same thing. If there are multiple places to handle, maybe we can consider to use defer here.

ahrtr: > Is it possible to use defer here for cleanup? 

We don't need to close & remove it in normal case. The tmp will be renamed to the bbolt db file.




> In my opinion the error log doesn't really provide any value to an etcd user so personally I would omit it. 

It's for debug purpose. It should be very rare. Overall not a big problem, and doesn't deserve too much time to debate on this.

ahrtr (approved):

server/storage/backend/backend.go
fuweid: > We don't need to close & remove it in normal case. The tmp will be renamed to the bbolt db file.

I mean, 

```
defer func() {
    if returnErr != nil {
          dosomething
    }
}()
```

It's just an option.

fuweid (approved):

ahrtr:
could you backport this PR to 3.5 and 3.4?

PR #18860, opened 2024-11-09
fix: use testify instead of t.Fatal in tests/common package (part 2)

mmorel-35:
#### Description

Related to #18972


ahrtr:

fuweid (approved):
LGTM

PR #18886, opened 2024-11-12
release: release (undraft) GitHub release from release script

ivanvc:
Release (updraft) GitHub release from the release script after creating it, avoiding the last manual step of this process. However, if the user answers no to the prompt, the script won't exit with a failure code, as it is still possible to manually release it.

I tested this [in my fork](https://github.com/ivanvc/etcd/releases), and it works as expected.

Part of #18604

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -343,13 +343,17 @@ main() {
     local release_url
     local gh_release_args=()
 
-    # For the main branch (v3.6), we should mark the release as a prerelease.
-    # The release-3.5 (v3.5) branch, should be marked as latest. And release-3.4 (v3.4)
-    # should be left without any additional mark (therefore, it doesn't need a special argument).
-    if [ "${BRANCH}" = "main" ]; then
-      gh_release_args=(--prerelease)
-    elif [ "${BRANCH}" = "release-3.5" ]; then
-      gh_release_args=(--latest)
+    local prompt
+    log_warning "WARNING: Only releases from the latest stable branch should be marked as the latest release"
+    read -p "Mark this as a the latest release [y/N]? " -r prompt
+    if [[ "${prompt,,}" == "y" ]]; then
+      gh_release_args=(--latest=true)
+    else
+      log_warning "WARNING: Only releases from the latest unstable branch should be marked as pre-releases (i.e., alpha, beta, or release candidates)"
+      read -p "Mark this as a pre-release [y/N]? " -r prompt
+      if [[ "${prompt,,}" == "y" ]]; then
+        gh_release_args=(--prerelease=true)
+      fi
     fi
 
     if [ "${REPOSITORY}" = "$(pwd)" ]; then
@@ -392,9 +396,18 @@ main() {
 
     release_url=$(gh --repo "${gh_repo}" release view "${RELEASE_VERSION}" --json url --jq '.url')
 
-    log_warning ""
-    log_warning "WARNING: The GitHub release for ${RELEASE_VERSION} has been created as a draft, please go to ${release_url} and release it."
-    log_warning ""
+    # Give a 10 minute timeout to the user to confirm the release.
+    read -p "Release GitHub release for ${RELEASE_VERSION} [y/N]? " -t 600 -r confirm
+    if [[ "${confirm,,}" == "y" ]]; then
+      maybe_run gh release edit "${RELEASE_VERSION}" \
+          --repo "${gh_repo}" \
+          --draft=false \
+          "${gh_release_args[@]}"
+    else
+      log_warning ""
+      log_warning "WARNING: The GitHub release for ${RELEASE_VERSION} has been created as a draft, please review it at ${release_url}."
+      log_warning ""
+    fi
   fi
 
   log_success "Success."


scripts/release.sh
```diff
@@ -391,8 +398,19 @@ main() {
     release_url=$(gh --repo "${gh_repo}" release view "${RELEASE_VERSION}" --json url --jq '.url')
 
     log_warning ""
-    log_warning "WARNING: The GitHub release for ${RELEASE_VERSION} has been created as a draft, please go to ${release_url} and release it."
+    log_warning "WARNING: The GitHub release for ${RELEASE_VERSION} has been created as a draft, please review it at ${release_url}."
     log_warning ""
+
+    # Give a 10 minute timeout to the user to confirm the release.
+    read -p "Release GitHub release for ${RELEASE_VERSION} [y/N]? " -t 600 -r confirm
```
jmhbnz, line 405: I think with a bit more testing / confidence we can simply drop the publishing of the release in draft mode and 10 minute review prompt and simply publish the full release. 

It was a really great milestone to see the script automatically creating the GitHub release for us today so let's just get a bit more confidence in the next round of releases and then revisit it :)

scripts/release.sh
```diff
@@ -337,19 +337,26 @@ main() {
...
-      gh_release_args=(--prerelease)
-    elif [ "${BRANCH}" = "release-3.5" ]; then
-      gh_release_args=(--latest)
-    fi
+    # Explicitly set release arguments per branch.
+    case "${BRANCH}" in
+      "main")
+        gh_release_args=(--latest=false --prerelease=true)
+        ;;
+      "release-3.5")
```
jmhbnz, line 353: We might want to start capturing a list of tasks in an issue and adding it to the https://github.com/etcd-io/etcd/milestone/38 in order to capture the trivial release related tasks we would need to perform in order to safely cut new `release-3.6` branch. For example, updating this script to add the `release-3.6` logic.

jmhbnz (approved):
LGTM - Thanks @ivanvc

scripts/release.sh
ivanvc: I agree. I was hesitant to add this change, but no matter what arguments I'd pass to `gh create release`, the draft release in the edit UI always has the "set as latest release" checkbox ticked. So, the workaround was to publish the release by editing it, while passing all the possible arguments.

ivanvc: I'm going to draft this PR then. We can revisit it when we feel confident about automating the release process.

scripts/release.sh
```diff
@@ -342,13 +342,15 @@ main() {
...
 
-    # For the main branch (v3.6), we should mark the release as a prerelease.
-    # The release-3.5 (v3.5) branch, should be marked as latest. And release-3.4 (v3.4)
-    # should be left without any additional mark (therefore, it doesn't need a special argument).
-    if [ "${BRANCH}" = "main" ]; then
-      gh_release_args=(--prerelease)
-    elif [ "${BRANCH}" = "release-3.5" ]; then
-      gh_release_args=(--latest)
+    local prompt
+    read -p "Mark this as a the latest release [y/N]? " -r prompt
```
serathius, line 346: Can you leave a note that we expect that latest releases come from latest branch and not from patches to old releases?

scripts/release.sh
```diff
@@ -342,13 +342,15 @@ main() {
...
+    if [[ "${prompt,,}" == "y" ]]; then
+      gh_release_args=(--latest=true)
+    else
+      read -p "Mark this as a pre-release [y/N]? " -r prompt
```
serathius, line 350: Same as above, would be good to inform that is expectation about what is and what shouldn't be a pre-release.

scripts/release.sh
ivanvc: Ah, thanks for the review and comment, @serathius. You caught me before I undrafted this pull request :smile:. I'll be sure to apply the suggestion, I'm still working on this.

ivanvc: Done

scripts/release.sh
ivanvc: Done

scripts/release.sh
ivanvc: I left this in place, as it follows the approach of the other steps.

scripts/release.sh
ivanvc: With the new approach, asking if it should be marked as the latest or as a pre-release is no longer relevant. Resolving.

ivanvc:
I'm parking this pull request again. I'm thinking that we can keep the draft release as we have now and release it using a GitHub action (after the released artifacts tests pass). Based on `etcd-io/auger` I know that we have the permissions to publish the release from an action.

ivanvc:
Note: I'll need to automate checking if it's a pre-release or if we should mark it as a release. I think it is doable with the following logic:

#### Pre-release

If there's a pre-release identifier (i.e., `-alpha.0` from `v3.6.0-alpha.0`), mark it as a pre-release.

#### Latest version

1. If the patch ends with `0` (i.e., `3.6.0`), mark it as the latest release.
2. If not,
   1. Get the latest release `gh release list --json isLatest,tagName --jq '.[] | select(.isLatest) | .tagName'`
   2. If the combination of the major and minor is the same as the release in question (i.e., `v3.6`), mark it as the latest release.

PR #18482, opened 2024-08-21
CHANGELOG: update changes for issue 18089

fuweid:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #18893, opened 2024-11-14
fix runtime error: comparing uncomparable type

ktalg:
Fixes a runtime panic that occurs when KeepAlive is called with a Context implemented by an uncomparable type, which is later canceled. The panic message is:

```
panic: runtime error: comparing uncomparable type lease_test.uncomparableCtx

goroutine 380 [running]:
go.etcd.io/etcd/client/v3.(*lessor).keepAliveCtxCloser(0xc000327720, {0x13416c8, 0xc00011b9b0}, 0x14c4932ade455905, 0xc002291420)
    /home/kevin/IdeaProjects/etcd/client/v3/lease.go:351 +0x1d6
created by go.etcd.io/etcd/client/v3.(*lessor).KeepAlive in goroutine 156
    /home/kevin/IdeaProjects/etcd/client/v3/lease.go:299 +0x554
```

To reproduce the issue, the existing test has been updated to include a custom uncomparable Context implementation. Previously, this test consistently caused a panic due to the uncomparable Context type. The updated implementation modifies keepAliveCtxCloser to properly handle such cases.

### DIFF
--- a/client/v3/lease.go
+++ b/client/v3/lease.go
@@ -264,6 +264,12 @@ func (l *lessor) Leases(ctx context.Context) (*LeaseLeasesResponse, error) {
 	return nil, ContextError(ctx, err)
 }
 
+// To identify the context passed to `KeepAlive`, a key/value pair is
+// attached to the context. The key is a `keepAliveCtxKey` object, and
+// the value is the pointer to the context object itself, ensuring
+// uniqueness as each context has a unique memory address.
+type keepAliveCtxKey struct{}
+
 func (l *lessor) KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAliveResponse, error) {
 	ch := make(chan *LeaseKeepAliveResponse, LeaseResponseChSize)
 
@@ -278,6 +284,10 @@ func (l *lessor) KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAl
 	default:
 	}
 	ka, ok := l.keepAlives[id]
+
+	if ctx.Done() != nil {
+		ctx = context.WithValue(ctx, keepAliveCtxKey{}, &ctx)
+	}
 	if !ok {
 		// create fresh keep alive
 		ka = &keepAlive{
@@ -348,7 +358,7 @@ func (l *lessor) keepAliveCtxCloser(ctx context.Context, id LeaseID, donec <-cha
 
 	// close channel and remove context if still associated with keep alive
 	for i, c := range ka.ctxs {
-		if c == ctx {
+		if c.Value(keepAliveCtxKey{}) == ctx.Value(keepAliveCtxKey{}) {
 			close(ka.chs[i])
 			ka.ctxs = append(ka.ctxs[:i], ka.ctxs[i+1:]...)
 			ka.chs = append(ka.chs[:i], ka.chs[i+1:]...)
--- a/tests/integration/clientv3/lease/lease_test.go
+++ b/tests/integration/clientv3/lease/lease_test.go
@@ -136,7 +136,14 @@ func TestLeaseKeepAlive(t *testing.T) {
 		t.Errorf("failed to create lease %v", err)
 	}
 
-	rc, kerr := lapi.KeepAlive(context.Background(), resp.ID)
+	type uncomparableCtx struct {
+		context.Context
+		_ func()
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	rc, kerr := lapi.KeepAlive(uncomparableCtx{Context: ctx}, resp.ID)
 	if kerr != nil {
 		t.Errorf("failed to keepalive lease %v", kerr)
 	}
@@ -152,6 +159,26 @@ func TestLeaseKeepAlive(t *testing.T) {
 
 	if kresp.ID != resp.ID {
 		t.Errorf("ID = %x, want %x", kresp.ID, resp.ID)
+	}
+
+	ctx2, cancel2 := context.WithCancel(context.Background())
+	rc2, kerr2 := lapi.KeepAlive(uncomparableCtx{Context: ctx2}, resp.ID)
+	if kerr2 != nil {
+		t.Errorf("failed to keepalive lease %v", kerr2)
+	}
+
+	cancel2()
+
+	_, ok = <-rc2
+	if ok {
+		t.Errorf("chan is not closed, want cancel stop keepalive")
+	}
+
+	select {
+	case <-rc:
+		// cancel2() should not affect first keepalive
+		t.Errorf("chan is closed, want keepalive continue")
+	default:
 	}
 
 	lapi.Close()


ktalg:
Could you kindly review this? @serathius

ktalg:
Could you kindly review this? @ahrtr

tests/integration/clientv3/lease/lease_test.go
```diff
@@ -136,7 +136,13 @@ func TestLeaseKeepAlive(t *testing.T) {
...
 	}
 
-	rc, kerr := lapi.KeepAlive(context.Background(), resp.ID)
+	type uncomparableCtx struct {
+		context.Context
+		_ func()
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	rc, kerr := lapi.KeepAlive(uncomparableCtx{Context: ctx}, resp.ID)
```
ahrtr, line 145: The easiest workaround is to pass in a pointer `&uncomparableCtx{Context: ctx}`.

ahrtr:
Thanks for raising this PR. The root cause is that the etcd client SDK tries to compare two context intances, which might not be comparable.

https://github.com/etcd-io/etcd/blob/41862836f9bdf35e7c3ab06b31510e846aa76ee7/client/v3/lease.go#L351

Is this your a real use case in your production code? The easiest workaround is to pass in a pointer of your customized context intance as mentioned in https://github.com/etcd-io/etcd/pull/18893#discussion_r1848648057.

We can attach an unique ID for each context passed to KeepAlive something like below, and compare the IDs when we need to identify the context,

```
const uniqueContextKey = "uniqueContextKey"
func withUniqueID(ctx context.Context, id string) context.Context {
	return context.WithValue(ctx, uniqueContextKey, id)
}

func getUniqueID(ctx context.Context) (string, bool) {
	id, ok := ctx.Value(uniqueContextKey).(string)
	return id, ok
}
```

We also need to review the source code to check if there are other places which compare the context as well.

ahrtr:
Or we can add a function something like below,
```
func CompareContexts(ctx1, ctx2 context.Context, key string) bool {
    return ctx1.Value(key) == ctx2.Value(key)
}
```

I am not sure whether there is an existing linter which can detect direct context comparison like below case. Probably we can add a such linter and integrate into golangci-lint if possible? @ivanvc @mmorel-35 @ldez

https://github.com/etcd-io/etcd/blob/41862836f9bdf35e7c3ab06b31510e846aa76ee7/client/v3/lease.go#L351

ldez:
Hello,

> I am not sure whether there is an existing linter which can detect direct context comparison like below case.

There is no linter on this specific case.

> Probably we can add a such linter and integrate into golangci-lint if possible?

The contexts are just structs so they are technically comparable.

https://go.dev/play/p/dyfw2rkHfTi

I understand your current issue, I need to investigate to kown if there are "valid" use cases behind this kind of comparison.

ahrtr:
> There is no linter on this specific case.

Thanks for the confirmation.



> The contexts are just structs so they are technically comparable.

Not really. 

Firstly the applications (including etcd) interact with with interface `context.Context`
https://github.com/golang/go/blob/ed07b321aef7632f956ce991dd10fdd7e1abd827/src/context/context.go#L68

Secondly, struct types are comparable if all their field types are comparable. In other words, if any field isn't comparable, then the struct isn't comparable.

Usually the default implementations of contexts included golang std lib are comparable. But users' customized implementation might not be comparable.

ldez:
FYI, I created a quick linter about context comparison to evaluate the relevance.

I analyzed several large projects, and I found no context comparison.

But I found 2 comparisons inside Go itself:
- [net/http/h2_bundle.go](https://github.com/golang/go/blob/5fdadffe3d20be9de491835c4351dee5dc81c42c/src/net/http/h2_bundle.go#L1022)
- [net/http/socks_bundle.go](https://github.com/golang/go/blob/e82308c198fbc26800fd1780cbd9f29935b210c5/src/net/http/socks_bundle.go#L36)

And as you said, struct comparison is not an anti-pattern.

For now, I think it's not worth adding a linter for this.

ahrtr:
Thanks @ldez 

@ktalg Could you please confirm if this is a real use case from your production environment or a hypothetical scenario?

The workaround is to pass in a pointer as mentioned in https://github.com/etcd-io/etcd/pull/18893#discussion_r1848648057

ktalg:
@ahrtr 

Thank you very much for reviewing this PR and providing detailed feedback!

> Could you please confirm if this is a real use case from your production environment or a hypothetical scenario?

 I encountered this issue while working on production-grade project code. The problem arose not from creating a new Context implementation but from passing a struct embedding a Context into the KeepAlive function, as illustrated in the test case I added.

This situation highlights two key points:

1. When a struct embeds an interface type, it automatically becomes an implementation of that interface. https://go.dev/doc/effective_go#embedding
2. The == operator is not a guaranteed valid operation for interface types; its behavior depends on the specific implementation. https://go.dev/ref/spec#Comparison_operators

I didn't anticipate that KeepAlive would internally use == to compare the type I passed in. My perspective is that robust code should rely solely on the declared contract of an interface, not on implicit assumptions. In this case, it seems to "assume" that the interface implementation is always comparable, which isn't guaranteed.

I believe this change improves the robustness of the code, ensuring it adheres to these principles.

client/v3/lease.go
```diff
@@ -348,7 +353,7 @@ func (l *lessor) keepAliveCtxCloser(ctx context.Context, id LeaseID, donec <-cha
 
 	// close channel and remove context if still associated with keep alive
 	for i, c := range ka.ctxs {
-		if c == ctx {
+		if val := c.Value(ctx); val != nil {
```
ahrtr, line 356: I am confused about this change. 

Have you considered https://github.com/etcd-io/etcd/pull/18893#issuecomment-2486198060 and also https://github.com/etcd-io/etcd/pull/18893#issuecomment-2486156823?

client/v3/lease.go
```diff
@@ -264,6 +264,9 @@ func (l *lessor) Leases(ctx context.Context) (*LeaseLeasesResponse, error) {
 	return nil, ContextError(ctx, err)
 }
 
+// for marking context itself
```
ahrtr, line 267:
```suggestion
// To identify the context passed to `KeepAlive`, a key/value pair is
// attached to the context. The key is a `keepAliveCtxKey` object, and
// the value is the pointer to the context object itself, ensuring
// uniqueness as each context has a unique memory address.
```

client/v3/lease.go
```diff
@@ -278,6 +281,11 @@ func (l *lessor) KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAl
 	default:
 	}
 	ka, ok := l.keepAlives[id]
+
+	if ctx.Done() != nil {
+		// mark this context for later retrieval
```
ahrtr, line 286:
```suggestion
```

client/v3/lease.go
```diff
@@ -348,7 +356,8 @@ func (l *lessor) keepAliveCtxCloser(ctx context.Context, id LeaseID, donec <-cha
 
 	// close channel and remove context if still associated with keep alive
 	for i, c := range ka.ctxs {
-		if c == ctx {
+		// the target ctx should have the same value as the current ctx
```
ahrtr, line 359:
```suggestion
```

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/18935

Please also squash the commits. Also do you have time to backport this to 3.5 and 3.4?

ktalg:
@ahrtr It seems I made a mistake when syncing with the upstream repository. This has now been fixed—thank you for pointing it out!

ahrtr (approved):
LGTM

Thanks

fuweid (approved):
LGTM

ahrtr:
@ktalg can you update the changelog for both 3.5 and 3.4? See example below,

https://github.com/etcd-io/etcd/pull/18482

ahrtr:
> should I also update the changelog for version 3.6?

No, 3.6 changelog only is supposed to only include new features in 3.6.

PR #18903, opened 2024-11-16
Robustness refactor patch operations

serathius:
First stage for https://github.com/etcd-io/etcd/pull/18772

We clean up the patchOperations functions to be simpler and use switch allowing support for other commands.

cc @ah8ad3 @ahrtr @MadhavJivrajani @henrybear327 @siyuanfoundation @jmhbnz


jmhbnz (approved):
LGTM - Many thanks for breaking this into both smaller commits and separate pr's @serathius it was much easier to review compared to the full initial pr we stepped through at KubeCon.

PR #18904, opened 2024-11-16
Robustness duplicated puts 2

serathius:
Spitted from https://github.com/etcd-io/etcd/pull/18772

cc @ah8ad3 @ahrtr @MadhavJivrajani @henrybear327 @siyuanfoundation @jmhbnz

### DIFF
--- a/tests/robustness/validate/patch_history.go
+++ b/tests/robustness/validate/patch_history.go
@@ -19,16 +19,16 @@ import (
 
 	"github.com/anishathalye/porcupine"
 
-	"go.etcd.io/etcd/tests/v3/robustness/client"
 	"go.etcd.io/etcd/tests/v3/robustness/model"
 	"go.etcd.io/etcd/tests/v3/robustness/report"
 )
 
 func patchLinearizableOperations(reports []report.ClientReport, persistedRequests []model.EtcdRequest) []porcupine.Operation {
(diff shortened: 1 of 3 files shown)


tests/robustness/validate/patch_history.go
```diff
@@ -46,25 +46,53 @@ func relevantOperations(reports []report.ClientReport) []porcupine.Operation {
...
+	requestRevision := map[keyValue]int64{}
+	for _, client := range reports {
+		for _, watch := range client.Watch {
+			for _, resp := range watch.Responses {
+				for _, event := range resp.Events {
+					switch event.Type {
+					case model.RangeOperation:
+					case model.PutOperation:
+						kv := keyValue{Key: event.Key, Value: event.Value}
+						requestRevision[kv] = event.Revision
```
fuweid, line 83: Is it required to check revision should not be decreased here for same key/value? The `validateOrdered` already covers that. It will be good to add some comments here.

serathius: There can only be one revision assigned to a unique write, we only patch unique writes like put, case you described should never happen. We do that for return time because its just estimation it based on different sources, like client response time, watch events and following write.

serathius:
ping @fuweid

fuweid (approved):

PR #18962, opened 2024-11-29
Fix benchmark indexing

serathius:
cc @jmhbnz @ahrtr

### DIFF
--- a/tools/benchmark/cmd/watch_latency.go
+++ b/tools/benchmark/cmd/watch_latency.go
@@ -145,7 +145,7 @@ func setupWatchChannels(key string) []clientv3.WatchChan {
 	wchs := make([]clientv3.WatchChan, len(streams)*watchLWatchersPerStream)
 	for i := 0; i < len(streams); i++ {
 		for j := 0; j < watchLWatchersPerStream; j++ {
-			wchs[i*len(streams)+j] = streams[i].Watch(context.TODO(), key, opts...)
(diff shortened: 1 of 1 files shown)


tools/benchmark/cmd/watch_latency.go
```diff
@@ -145,7 +145,7 @@ func setupWatchChannels(key string) []clientv3.WatchChan {
 	wchs := make([]clientv3.WatchChan, len(streams)*watchLWatchersPerStream)
 	for i := 0; i < len(streams); i++ {
 		for j := 0; j < watchLWatchersPerStream; j++ {
-			wchs[i*len(streams)+j] = streams[i].Watch(context.TODO(), key, opts...)
+			wchs[j*len(streams)+i] = streams[i].Watch(context.TODO(), key, opts...)
```
ahrtr, line 148: It's interesting. 

If we regard the `wchs` as a `len(streams)` * `watchLWatchersPerStream` matrix (specifically, it has `len(streams)` rows, and each row has `watchLWatchersPerStream` columns). 

The original implementation is saving the watchers row by row, while your implementation is column by column. Both are correct to me.

serathius: The code is not correct as multiplier is wrong. So things break when the number or row doesn't match number of columns.

ahrtr (approved):

PR #18964, opened 2024-11-30
Add a threshold for codecov failure

jmhbnz:
Now that we have enabled proper codecov integration in https://github.com/kubernetes/org/issues/5231 we have noticed pr's being failed for coverage reductions that are trivial, i.e. < .1%.

Let's try and introduce some configuration to make codecov only fail when a meaningful reduction in coverage has occurred.

cc @serathius

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -23,6 +23,13 @@ coverage:
   range: 60..80
   round: down
   precision: 2
+  status:
+    project:
+      default:
+        target: auto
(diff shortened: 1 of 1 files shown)


ahrtr (approved):

serathius (approved):
1% of whole codebase might make it never trigger. On the other hand I don't think kind of check could be ever automated. Would rather depend on approvers to see that there is a new feature with low covereage.

PR #18968, opened 2024-11-30
fix: use testify instead of t.Fatal or t.Error in contrib package

mmorel-35:
#### Description

This uses testify instead of testing for t.Fatal or t.Error calls in contrib package

Related to #18972


ahrtr (approved):

ivanvc (approved):
LGTM. Thanks, @mmorel-35.

PR #18969, opened 2024-11-30
fix: use testify instead of t.Fatal or t.Error in etcdctl packages

mmorel-35:
#### Description

This uses testify instead of testing for t.Fatal or t.Error calls in client package

Related to #18972


jmhbnz (approved):
LGTM - Thanks @mmorel-35

ahrtr (approved):

PR #18970, opened 2024-11-30
fix: use testify instead of t.Fatal or t.Error in pkg package (part 1)

mmorel-35:
#### Description

This uses testify instead of testing for t.Fatal or t.Error calls in pkg package

Related to #18972


ahrtr (approved):

jmhbnz (approved):
LGTM - Thanks @mmorel-35

Issue #18965, opened 2024-11-30
Make fragment in client watch request default


serathius:
### What would you like to be added?

Fragment is a client controller Watch feature introduced in etcd v3.4. [Proto](https://github.com/etcd-io/etcd/blob/278a9a2682c24366afcadbed420db40d810cb9bc/api/etcdserverpb/rpc.proto#L811)

If configured by client, it allows server to chunk Watch response by maximum size. 
When response is fragmented instead of sending one big response, etcd will send subsets of events with responses marked as fragmented. Only for the last response it will unmark fragmented, letting client know that this response was finished. 

From user perspective enabling fragmenting doesn't cause any changes, all the fragmenting is handled within client code, which will combine fragmented responses into single one. It's a protocol optimization, not user visible.

Marking responses as fragmented seems to be always good, the only exception would be if there are multiple events sharing revision from single response. When fragmenting is critical to maintain [atomic watch property](https://etcd.io/docs/v3.5/learning/api_guarantees/#watch-apis). 

As part of this issue I would like to recommend K8s to use it.

### Why is this needed?

Make etcd more reliable during watch conjestion.

In tests done by running `go run ./tools/benchmark/main.go watch-latency --watch-per-stream 1000 --streams 1 --put-total 200 --val-size 100000`, we notice around 20% reduction in memory used and almost 2x improvement in latency while watch becomes congested.

| object size [KB] | Events/s    | Fragment | etcd memory[MB] | 50%ile  | 90%ile  | 99%ile  |
| ---------------- | ----------- | -------- | --------------- | ------- | ------- | ------- |
| 5                | 100646.0428 | FALSE    | 111.72          | 0.1005  | 0.1543  | 0.1856  |
| 5                | 100887.1057 | TRUE     | 110.44          | 0.1005  | 0.1547  | 0.1866  |
| 10               | 78963.1236  | FALSE    | 1103.924        | 2.7177  | 7.3661  | 11.9127 |
| 10               | 87156.2031  | TRUE     | 783.276         | 1.9748  | 5.2995  | 9.2281  |
| 100              | 6576.9363   | FALSE    | 5138.712        | 14.0618 | 26.2721 | 29.1691 |
| 100              | 10619.2212  | TRUE     | 4144.832        | 8.5963  | 15.969  | 17.8447 |

serathius:
cc @ahrtr @dead2k @jpbetz @wojtek-t

PR #18975, opened 2024-12-01
Use size of events to batch responses

serathius:
Instead of batching by number of revisions (do not understand it's purpose), do batching by size of events to avoid resync generating huge responses. For example Watch opening on very old revision (like 5 minutes behind within but still compaction window), resync will load the whole history and try to push it in one response. If watch is congested or client is not reading this can cause a huge spikes in memory. 

Implements https://github.com/etcd-io/etcd/issues/18965 without needing to reconfigure clients.

This is basically implementing fragment watch feature, without needing to configure it on client side. We still need to maintain [watch Atomic property](https://etcd.io/docs/v3.5/learning/api_guarantees/#watch-apis), so we cannot split revision between responses. This is simple, if we start an a revision we continue it even if we are over the limit. 

Proposing to use the same response size as used by fragment, the size of request + grpc overhead (2MB). As we cannot split revisions, without fragmenting the maximum response size is 2x that (4MB). 

Possible downsides of batching early in pipeline:
* Batches generated early might be working against existing fragment batching. 
* Limits maximum throughput per watch, resync every 100ms means we limit watch throughput per watcher to 20MB/s. Which is both good and bad, depending on whether we over or undershoot existing limit. Needs to be tested.
* Will cause more frequent resync calls, which should be mitigated by https://github.com/etcd-io/etcd/pull/17563


serathius:
cc @ahrtr @jmhbnz

serathius:
Spitted https://github.com/etcd-io/etcd/pull/18976 out

Issue #18925, opened 2024-11-20
Improve dependency management


ivanvc:
### What would you like to be added?

The current dependency management process involves a lot of manual intervention. There have been some attempts at improving it, including updating the `update_deps.sh` script, trying out different dependabot configurations, and evaluating other tools besides dependabot.

I want to formalize the work and have a central place to discuss alternatives, pros, and cons.

### Possible approaches

1. Do nothing / keep dependency management as it is right now. As stated before, the current process requires a lot of manual intervention, and the toil is high. The time to create the weekly PR and merge it keeps sliding into later in the week. There are times in which by human error, the commits are not properly bumping the dependencies.
2. Improve dependabot configuration (so far, I've tested the following options):
    1. Use the [_new_ `directories` property](https://github.blog/changelog/2024-06-25-simplified-dependabot-yml-configuration-with-multi-directory-key-directories-and-wildcard-glob-support/): The result of this is multiple pull requests for the same dependency across all of the submodules where it is present. Although we (@jmhbnz and I) thought [this](https://github.com/dependabot/dependabot-core/issues/7547) was going to be the silver bullet, in the end, it doesn't work as we expected. Enabling this will just add noise to our repository, as we still need to create a PR to bump dependencies manually. See the following example of `go.opentelemetry.io/otel/trace` in my fork: https://github.com/ivanvc/etcd/pull/410 https://github.com/ivanvc/etcd/pull/406 https://github.com/ivanvc/etcd/pull/390 https://github.com/ivanvc/etcd/pull/386.
    2. Use the `directories` property and [`groups`](https://github.blog/changelog/2023-08-24-grouped-version-updates-for-dependabot-are-generally-available/). Although this looked promising, as it creates a single pull request with all the weekly dependencies, the issues that I find are: 1. it creates a single commit for the whole group (rather than one per dependency bumped), 2. it doesn't run `go mod tidy`, and it breaks our CI. See an example in my fork: https://github.com/ivanvc/etcd/pull/265 (in the files changed and any `go.sum` file, it's clear that it is not tidying them).
3. Replace dependabot with renovatebot. I noticed that several projects on board for dependabot's `directories` beta feature moved to renovate. The immediate issue is that we would need a new GitHub application installed for our GitHub organization. This is not the silver bullet but may help decrease the effort. I tested some configurations in my fork:
[...]

PR #18992, opened 2024-12-03
chore: group tools/mod updates and update only direct dependencies

mmorel-35:
Concerning tools update it seems more appropriate to update only direct dependencies.
They can also be grouped in one PR update.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,7 +18,7 @@ updates:
     schedule:
       interval: weekly
     allow:
-      - dependency-type: all
+      - dependency-type: direct
 
   - package-ecosystem: docker
     directory: /


.github/dependabot.yml
```diff
@@ -18,7 +18,11 @@ updates:
     schedule:
       interval: weekly
     allow:
-      - dependency-type: all
+      - dependency-type: direct
+    groups:
+      tools-mod:
+        patterns:
+          - "*"
```
jmhbnz, line 25: This seems reasonable to me, most of our purely indirect dependencies come from tools so we end up closing them. Thanks for proposing it @mmorel-35, it should reduce weekly spam pr's.   What do you think @ivanvc?

ivanvc:
Link to #18925.

.github/dependabot.yml
ivanvc: I think this PR has some overlap with the work from issue https://github.com/etcd-io/etcd/issues/18925.

That aside, I tested it in my fork, and it seems like dependabot is tidying the module correctly (https://github.com/ivanvc/etcd/pull/450), as there is only a single directory specified (it looks like the tidying bug is only present when there is more than one directory specified).

The only potential risk I see is that if we have a dependency in `/tools/mod` that is also used in other submodule (i.e., `github.com/gogo/protobuf`, `go.etcd.io/raft/v3`, `github.com/grpc-ecosystem/grpc-gateway/v2`), our CI check `pull-etcd-verify` (`verify-dep`), will fail because the dependencies won't be at the same version.

ivanvc: An alternative would be first to merge the manual updates (including updating `/tools/mod` dependencies that are used in other modules). Then, wait for dependabot to update the grouped pull request. Finally, we could merge the grouped `/tools/mod` PR.

Even though it can be an alternative, it may not be viable (as just writing the steps in the previous paragraph took me a while to ensure I got it right). This alternative could also be potentially confusing, and instead of decreasing the toil, it could possibly increase it :sweat_smile:... But I'm trying to think of alternatives to make this work.

ivanvc: So, I'm in favor of keeping `dependency-type: direct` but dropping the group. What do you think, @jmhbnz?

jmhbnz: Sounds good to me, let's go with just the `dependency-type: direct` for this pr @mmorel-35.

ivanvc (approved):
LGTM. Thanks, @mmorel-35.

ivanvc:
/retitle chore: update only tools/mod direct dependencies

Issue #18994, opened 2024-12-03
Don't allow users to change the value for `--v2-deprecation`


ahrtr:
### What would you like to be added?

part of https://github.com/etcd-io/etcd/issues/12913

If the v2store still contains custom content, 3.6 should panic on bootstrap.  But if users change `--v2-deprecation` to [V2Depr0NotYet](https://github.com/etcd-io/etcd/blob/2abb334103a327270e4812fa0da21e941125440f/server/config/v2_deprecation.go#L21), then it won't panic anymore even if v2store contains user data (custom content). It may lead to data inconsistency in such case.

To ensure a smoother upgrade from 3.5 to 3.6, we still keep the option, but ignore users' provided value; and print a warning if present.

### Why is this needed?

To avoid potential data inconsistency due to users misconfiguration.

ahrtr:
Note [V2Depr0NotYet](https://github.com/etcd-io/etcd/blob/2abb334103a327270e4812fa0da21e941125440f/server/config/v2_deprecation.go#L21) isn't a valid option any more starting from v3.6, see link below. But it still doesn't make sense to let users to configure the flag. We only need to proceed with deprecating v2store as scheduled; users don’t need to worry about opting in.

Also there is no any different between these valid options in 3.6, because `V2Depr1WriteOnlyDrop` (`--write-only-drop-data`) isn't implemented in 3.6.

https://github.com/etcd-io/etcd/blob/7e04bd2de9c41f6ba2d6ea8a4571f78f928af3b3/server/etcdmain/config.go#L98-L101

PR #19006, opened 2024-12-04
Deprecate `--v2-deprecation` and schedule to remove it in 3.8

ahrtr:
Resolve https://github.com/etcd-io/etcd/issues/18994

### DIFF
--- a/server/config/v2_deprecation.go
+++ b/server/config/v2_deprecation.go
@@ -17,47 +17,50 @@ package config
 type V2DeprecationEnum string
 
 const (
-	// No longer supported in v3.6
+	// V2Depr0NotYet means v2store isn't deprecated yet.
+	// Default in v3.5, and no longer supported in v3.6.
 	V2Depr0NotYet = V2DeprecationEnum("not-yet")
-	// No longer supported in v3.6
-	//
-	// Deprecated: Please use V2Depr0NotYet.
+
+	// Deprecated: to be decommissioned in 3.7. Please use V2Depr0NotYet.
+	// TODO: remove in 3.7
 	//revive:disable-next-line:var-naming
 	V2_DEPR_0_NOT_YET = V2Depr0NotYet
-	// Default in v3.6.  Meaningful v2 state is not allowed.
-	// The V2 files are maintained for v3.5 rollback.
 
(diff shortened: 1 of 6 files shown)


fuweid (approved):

ahrtr:
ping @serathius

server/config/v2_deprecation.go
```diff
@@ -17,47 +17,50 @@ package config
...
-	// Deprecated: Please use V2Depr2Gone
+
+	// Deprecated: to be decommissioned in 3.7. Please use V2Depr2Gone.
+	// TODO: remove in 3.7
 	//revive:disable-next-line:var-naming
 	V2_DEPR_2_GONE = V2Depr2Gone
 
-	// Default deprecation level.
+	// V2DeprDefault is the default deprecation level.
 	V2DeprDefault = V2Depr1WriteOnly
```
serathius, line 60: Is `V2Depr1WriteOnly` the correct v2 default config for v3.6? From above 
```
	// V2Depr1WriteOnly means only writing v2store is allowed.
	// Default in v3.6.  Meaningful v2 state is not allowed.
	// The V2 files are maintained for v3.5 rollback.
```

However we plan to panic on any v2 state. Should we reconcile this?

ahrtr: > Is `V2Depr1WriteOnly` the correct v2 default config for v3.6?

YES.



> However we plan to panic on any v2 state.

In 3.6, we only allow membership data in v2state. We will keep it as it's in 3.6. In 3.7, the default level will be `V2Depr1WriteOnlyDrop`, and we will remove any v2 snapshot files if found.

https://github.com/etcd-io/etcd/blob/fbac2a064c381d6f8c74c4d9d01fde816e200eef/server/etcdserver/bootstrap.go#L407-L409

serathius: I thought `V2DeprecationEnum` was about key-value data stored in v2 store, not about membership data. 

To clarify, enums are not accurate to the current state, as the plan has changed in https://github.com/etcd-io/etcd/issues/12913#issuecomment-1531941532.

serathius (approved):

ahrtr:
Thanks both @fuweid and @serathius !

Issue #19010, opened 2024-12-05
Release script doesn't work for v3.6.0 (main branch) releases


ivanvc:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

While working on using `gh` to automate creating the GitHub releases, I noticed that the release script failed when doing v3.6.0 prereleases.

### What did you expect to happen?

It should work but fails because it expects a branch `release-3.6` to exist.

### How can we reproduce it (as minimally and precisely as possible)?

Running the release script with a v3.6.0 version, i.e.:

```bash
$ REPOSITORY=git@github.com:ivanvc/etcd.git DRY_RUN=true ./scripts/release.sh v3.6.0-alpha.2
DRY_RUN=true
RELEASE_VERSION=vv3.6.0-alpha.2
MINOR_VERSION=v3.6
BRANCH=release-v3.6
REPOSITORY=git@github.com:ivanvc/etcd.git

Preparing temporary directory: /tmp/etcd-release-v3.6.0-alpha.2
% (cd ../../../../../../tmp/etcd-release-v3.6.0-alpha.2 && 'git' 'clone' 'git@github.com:ivanvc/etcd.git' '--branch' 'release-v3.6' '--depth' '1')
stderr: Cloning into 'etcd'...
stderr: warning: Could not find remote branch release-v3.6 to clone.
stderr: fatal: Remote branch release-v3.6 not found in upstream origin
```

### Etcd version (please run commands below)

N/A

### Etcd configuration (command line flags or environment variables)

N/A

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

N/A

ivanvc:
Link to #13538

PR #19011, opened 2024-12-05
release: Allow "v" prefix when specifying release version

ivanvc:
This pull requests:

1. Allows to specify the release version with either the "v" prefix or not (i.e., `v3.6.0` or `3.6.0`)
~2. Sets the `BRANCH` to main if the minor version is `3.6`.~

~Technically, this is not needed if we pass the environment variable `BRANCH` set to `main` with every pre-release. So, another "fix" would be to document this in the release guide. I'm open to reverting and doing the documentation this way.~

Part of #19010.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -48,8 +48,9 @@ help() {
 }
 
 main() {
-  VERSION=$1
-  if [[ ! "${VERSION}" =~ [0-9]+.[0-9]+.[0-9]+ ]]; then
+  # Allow to receive the version with the "v" prefix, i.e. v3.6.0.
+  VERSION=${1#v}
+  if [[ ! "${VERSION}" =~ ^[0-9]+.[0-9]+.[0-9]+ ]]; then
     log_error "Expected 'version' param of the form '<major-version>.<minor-version>.<patch-version>' but got '${VERSION}'"
     exit 1
   fi


scripts/release.sh
```diff
@@ -62,7 +62,14 @@ main() {
       BRANCH=$(git rev-parse --abbrev-ref HEAD)
   else
       REPOSITORY=${REPOSITORY:-"git@github.com:etcd-io/etcd.git"}
-      BRANCH=${BRANCH:-"release-${MINOR_VERSION}"}
+      if [ "${MINOR_VERSION}" == "3.6" ]; then
```
serathius, line 65: For non-released branches I expect that script will expect caller will be picking the branch manually.

scripts/release.sh
```diff
@@ -49,11 +49,11 @@ help() {
 
 main() {
   VERSION=$1
-  if [[ ! "${VERSION}" =~ [0-9]+.[0-9]+.[0-9]+ ]]; then
+  if [[ ! "${VERSION}" =~ ^v?[0-9]+.[0-9]+.[0-9]+ ]]; then
     log_error "Expected 'version' param of the form '<major-version>.<minor-version>.<patch-version>' but got '${VERSION}'"
     exit 1
   fi
-  RELEASE_VERSION="v${VERSION}"
+  RELEASE_VERSION="v${VERSION#v}"
```
ahrtr, line 56: I think we can merge this enhancement separately.

scripts/release.sh
ivanvc: That sounds good; let's go with updating the documentation, then.

scripts/release.sh
ivanvc: I'll repurpose this pull request only to have this enhancement. And will open up another one to update the release documentation.

ivanvc: Done.

jmhbnz (approved):
LGTM - Thanks @ivanvc

ahrtr (approved):

PR #18973, opened 2024-11-30
fix: enable gosimple linter

mmorel-35:

PR #19013, opened 2024-12-05
fix: golangci-lint yamllint

mmorel-35:
#### Description 

Golangci-lint config file is not well formatted. This fixes it

### DIFF
--- a/tools/.golangci.yaml
+++ b/tools/.golangci.yaml
@@ -8,7 +8,7 @@ issues:
     # exclude ineffassing linter for generated files for conversion
     - path: conversion\.go
       linters: [ineffassign]
-    - text: "S1000"  # TODO: Fix me
+    - text: "S1000" # TODO: Fix me
       linters:
         - gosimple
   exclude-files:


jmhbnz (approved):

tools/.golangci.yaml
```diff
@@ -8,7 +8,7 @@ issues:
     # exclude ineffassing linter for generated files for conversion
     - path: conversion\.go
       linters: [ineffassign]
-    - text: "S1000"  # TODO: Fix me
+    - text: "S1000" # TODO: Fix me
```
ahrtr, line 11: We forgot to add the "ok-to-test" label, so it was NOT detected in previous PR https://github.com/etcd-io/etcd/pull/18973

Can we enforce the "ok-to-test" label before merging any PR technically?

jmhbnz: I think for this we need to enable `tide` prow plugin and delegate the merging of pr's to tide so we no longer do it manually.  Usually the tide criteria is combination of `approved` and `lgtm` labels I think but we could probably require `ok-to-test` as well.

I discussed enabling tide with @serathius in Salt Lake City KubeCon but I have not had bandwidth to plan the change yet.

ahrtr: OK. The immediate solution that I can think of is to fail the [Approve GitHub Workflows](https://github.com/etcd-io/etcd/blob/main/.github/workflows/gh-workflow-approve.yaml) if `ok-to-test` label isn't present. Accordingly it breaks the `merge-when-workflow-green` rule.

ahrtr (approved):

PR #19030, opened 2024-12-07
Add rw benchmark line chart

ivanvc:
I've been working on this on and off, and I wanted to get it moving before the end of this year.

It implements line charts for the "rw-heatmaps." I want to get some feedback on the resulting charts. We should also rename `tools/rw-heatmaps` to something that suits both heatmaps and line charts.

I'll squash commits as we get closer to merging this.

Supersedes #15060.

/cc @ahrtr @jmhbnz

### DIFF
--- a/tools/rw-heatmaps/cmd/root.go
+++ b/tools/rw-heatmaps/cmd/root.go
@@ -33,6 +33,8 @@ var (
 	ErrMissingInputFileArg = fmt.Errorf("missing input file argument")
 	// ErrInvalidOutputFormat is returned when the output format is invalid.
 	ErrInvalidOutputFormat = fmt.Errorf("invalid output format, must be one of png, jpg, jpeg, tiff")
+	// ErrInvalidChartTYpe is returned when the chart type is invalid.
+	ErrInvalidChartType = fmt.Errorf("invalid chart type, must be one of line, heatmap")
 )
 
 // NewRootCommand returns the root command for the rw-heatmaps tool.
@@ -56,6 +58,10 @@ func NewRootCommand() *cobra.Command {
 				}
 			}
 
+			if o.chartType == "line" {
+				return chart.PlotLineCharts(datasets, o.title, o.outputImageFile, o.outputFormat)
+			}
+
 			return chart.PlotHeatMaps(datasets, o.title, o.outputImageFile, o.outputFormat, o.zeroCentered)
 		},
 	}
@@ -70,6 +76,7 @@ type options struct {
 	outputImageFile string
 	outputFormat    string
 	zeroCentered    bool
+	chartType       string
 }
 
 // newOptions returns a new options for the command with the default values applied.
@@ -77,6 +84,7 @@ func newOptions() options {
 	return options{
 		outputFormat: "jpg",
 		zeroCentered: true,
+		chartType:    "heatmap",
 	}
 }
 
@@ -86,6 +94,7 @@ func (o *options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&o.outputImageFile, "output-image-file", "o", o.outputImageFile, "output image filename (required)")
 	fs.StringVarP(&o.outputFormat, "output-format", "f", o.outputFormat, "output image file format")
 	fs.BoolVar(&o.zeroCentered, "zero-centered", o.zeroCentered, "plot the improvement graph with white color represents 0.0")
+	fs.StringVarP(&o.chartType, "chart-type", "c", o.chartType, `type of chart to plot ["line", or "heatmap"]`)
 }
 
 // Validate returns an error if the options are invalid.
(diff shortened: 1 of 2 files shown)


ivanvc:
Given the following files: 

* [result-202402281701.csv](https://github.com/user-attachments/files/18046507/result-202402281701.csv)
* [result-202401270029.csv](https://github.com/user-attachments/files/18046508/result-202401270029.csv)

Running it with:

```bash
go run . result-20240* -t test-line -o test-line -c line
```

Plots:
test-line_readwrite

ivanvc:
Running it for a single dataset, i.e., [result-202402281701.csv](https://github.com/user-attachments/files/18047655/result-202402281701.csv)

```bash
go run . result-202402281701.csv -t test-line -o test-line -c line
```

Generates:

test-line_readwrite

ivanvc:
Another alternative is to have the value size represent the line width, and set dashes for the second dataset, which looks like this for an individual dataset:

test-line_readwrite

And like this for multiple datasets:

test-line_readwrite

ahrtr:
Thanks @ivanvc for the progress.

As compared to my previous solution in https://github.com/etcd-io/etcd/pull/15060, the good side / bad side of this PR is exactly the bad side / good side of my previous solution.

This PR,
- good side: get everything included in one jpg file. So easy to share in PR discussion.
-  bad side: the data density is too big and accordingly less clearer.


Probably we can find a compromise or combine the best aspects of both.
- We still present different R/W ratio in different line charts. It's the same as what this PR does.
- For each specific RW ratio, 
  - There are 4 kinds of color, we can simplify it to just 2 colors; i.e. red --> write, and blue --> read. If there are two datasets, then present the second on in dot dash line, and with the same color as the first one.  The key is to create a simple rule to remember: 
    - red for write and blue for read, dot dash for the second dataset; 
      - dataset 1 write: solid red line
      - dataset 1 read: solid blue line
      - dataset 2 write: dot dash red line
      - dataset 2 read: dot dash blue line 
    - or solid line for write and dot dash line for read, and red for first data set and blue for the second dataset.
      - dataset 1 write: solid red line
      - dataset 1 read: dot dash red line
      - dataset 2 write: solid blue line
      - dataset 2 read: dot dash blue line 
     
    This should be easy to do.  But I agree that this may involve some degree of personal subjectivity

  - present the diagram(line chart) in html page instead of jpg file, so that we can highlight the data of a specific value size (or hide the data of some value sizes). For example, we still present the result of all value sizes (i.e. including 256, 512, 1024, 2048) in one line chart, but users can hide the result for value sizes 512, 1024 and 2048, so only 256 is displayed, so as to make it clearer and more readable. Note this is one of the key good sides of my previous solution. This might be a little hard based on the library this PR selects.

ivanvc:
I agree that the chart is too dense. We would benefit from an interactive format (i.e., your solution/d3js, etc.). The problem with an HTML file is that it's less portable (and embeddable) than an image. If we go that route, we should host it somewhere (i.e., Netlify), but it adds to the complexity of the implementation.

Implementing your suggestions:

test-line_readwrite

or

test-line_readwrite

ahrtr:
> The problem with an HTML file is that it's less portable (and embeddable) than an image. If we go that route, we should host it somewhere (i.e., Netlify), but it adds to the complexity of the implementation.

You can export each line chart into a .PNG file, refer to https://github.com/etcd-io/etcd/pull/15060#issuecomment-1385398082.

You can also create a screen shot for all line charts, see the picture in https://github.com/etcd-io/etcd/pull/15060#issuecomment-1379928791.

> Implementing your suggestions:

It's a little better now, but it's really hard to differentiate the lines for different value sizes.

ahrtr:
> red for write and blue for read, dot dash for the second dataset;
> * dataset 1 write: solid red line
> * dataset 1 read: solid blue line
> * dataset 2 write: dot dash red line
> * dataset 2 read: dot dash blue line

I suggest to get this merged firstly, and consider to improve it later.

ahrtr:
@ivanvc can we try to get this merged? thx

ivanvc:
Yes, this still needs some code improvements. It was a PoC, and I didn't finish it due to other priorities.

ahrtr:
We need this PR to visualize the benchmark result. It's in the top of my review list. thx

ivanvc:
Alright, I'll clean up the code and will have it ready soon.

ivanvc:
Oops, I clicked ready for review too soon. I had changes on my laptop, which I forgot to push. The code stands with the second option we discussed.

It also makes sense to rename it from `rw-heatmaps` to `rw-bench`. I'm open to doing that in this pull request or as a follow-up.

ahrtr (approved):

jmhbnz (approved):
LGTM - Thanks for your work on this @ivanvc.  I think we'll still iterate on the presentation to try and make the graphs clearer but it's a solid first iteration and great to have this option available in the tool.

Issue #18983, opened 2024-12-02
Reduce the time e2e tests take


serathius:
### What would you like to be added?

E2e tests are taking over 40 minutes now, we should keep them unver 30 minutes or even lower.

For the top test candidates to optimize use https://testgrid.k8s.io/sig-etcd-periodics#ci-etcd-e2e-amd64&graph-metrics=test-duration-minutes

Current state 
* go.etcd.io/etcd/tests/v3/common.TestMemberAdd > 5 minutes
* go.etcd.io/etcd/tests/v3/common.TestMemberRemove > 2 minutes
* go.etcd.io/etcd/tests/v3/e2e.TestAuthority > 1.5 minutes
* go.etcd.io/etcd/tests/v3/e2e.TestRuntimeReconfigGrowClusterSize > 1 minute
* go.etcd.io/etcd/tests/v3/common.TestUserAdd_Simple > 1 minute
* go.etcd.io/etcd/tests/v3/common.TestLeaseGrantAndList ~ 50 seconds
* go.etcd.io/etcd/tests/v3/e2e.TestVerifyHashKVAfterCompact ~50 seconds
* go.etcd.io/etcd/tests/v3/e2e.TestMixVersionsSnapshotByMockingPartition ~ 40 seconds
* go.etcd.io/etcd/tests/v3/common.TestKVDelete ~ 40 seconds
* go.etcd.io/etcd/tests/v3/common.TestLeaseGrantTimeToLiveExpired ~ 35 seconds
* go.etcd.io/etcd/tests/v3/e2e.TestWatchDelayForManualProgressNotification ~ 35 seconds
...

Looking for someone to take a look at the top tests and propose improvements to reduce runtime

![image](https://github.com/user-attachments/assets/18681f39-28e1-4447-a27b-2d682c626e4d)

### Why is this needed?

We should keep the test execution time low to ensure good contributor expirience.

manthanguptaa:
@serathius I can take a look at this

ivanvc:
Thanks, @manthanguptaa!

/assign @manthanguptaa

manthanguptaa:
Hey @ivanvc! I really appreciate you checking in, and I’m so sorry for not getting this to you on time. Things have been quite hectic on my end with work, and unfortunately, I won’t be able to take this up right now. I truly appreciate your patience and understanding. I hope to take up something in the future to work

ivanvc:
[...]

PR #19032, opened 2024-12-09
Update TestMemberAdd's context timeout by adding 5 more seconds

ahrtr:
The test case will wait 5 seconds if waitForQuorum is `true`

https://github.com/etcd-io/etcd/blob/801dfc3964749a741aa3c6b7c5be07c3ef8a931c/tests/common/member_test.go#L127-L129

The context timeout is 10 seconds in total, but sometimes it may take a long time for the cluster to be ready, and the remaining time to run the test may < 5 seconds. In that case, the test will timeout for sure.

```
execute.go:47: ---> Test failed: test timed out after 4.252903074s, err: context deadline exceeded
```

### DIFF
--- a/tests/common/member_test.go
+++ b/tests/common/member_test.go
@@ -113,7 +113,11 @@ func TestMemberAdd(t *testing.T) {
 		for _, quorumTc := range quorumTcs {
 			for _, clusterTc := range clusterTestCases() {
 				t.Run(learnerTc.name+"/"+quorumTc.name+"/"+clusterTc.name, func(t *testing.T) {
-					ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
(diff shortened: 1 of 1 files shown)


serathius:
Please note that TestMemberAdd is the slowest test we have https://github.com/etcd-io/etcd/issues/18983 . Thanks for removing flakes, but maybe you have idea how to speed up the test?

ahrtr:
> TestMemberAdd is the slowest test

It's because it has too many sub cases: 5 * 4 * 2 = 40, and in some cases there is a 5 seconds sleep.

https://github.com/etcd-io/etcd/blob/801dfc3964749a741aa3c6b7c5be07c3ef8a931c/tests/common/member_test.go#L112-L114

serathius:
Right, we should think about how to remove the sleep.

PR #8289, opened 2017-07-20
clientv3, etcdmain, proxy: support authed RPCs with grpcproxy

mitake:
This commit lets grpcproxy support authed RPCs. Auth tokens supplied
by clients are now forwarded to etcdserver by grpcproxy.

Fix https://github.com/coreos/etcd/issues/8247

/cc @zhuqunzhou @heyitsanthony 

This PR is still WIP because I'm considering how to protect proxy based watch. It isn't simple like other RPCs because of its streaming and broadcast nature.

@zhuqunzhou could you try it for your purpose? Authed KV requests with proxy should work now.

PR #19033, opened 2024-12-09
grpcproxy: use metadata instead of context withvalue in with client auth token

krijohs:
Change to use metadata instead of `context.WithValue` to ensure each proxy watcher client has a new stream created with its token. 
Previously context.WithValue resulted in `streamKeyFromCtx` returning an empty string in the clientv3 watcher, causing stream reuse.
When new clients connected to proxy after the token expired (token for the initial client which connected) the reused stream's context would still contain the expired token.                                                                                                                           This caused auth failures when `isWatchPermitted` on cluster checked the stream's context resulting in hanging proxy watcher clients.

Issue can be reproduced by setting a low `--auth-token-ttl` on cluster and connect 1 client first to proxy and then connect a second one after token expired.

### DIFF
--- a/server/proxy/grpcproxy/util.go
+++ b/server/proxy/grpcproxy/util.go
@@ -37,7 +37,8 @@ func getAuthTokenFromClient(ctx context.Context) string {
 func withClientAuthToken(ctx, ctxWithToken context.Context) context.Context {
 	token := getAuthTokenFromClient(ctxWithToken)
 	if token != "" {
-		ctx = context.WithValue(ctx, rpctypes.TokenFieldNameGRPCKey{}, token)
+		md := metadata.Pairs(rpctypes.TokenFieldNameGRPC, token)
+		return metadata.NewOutgoingContext(ctx, md)
 	}
 	return ctx
 }
@@ -66,9 +67,9 @@ func AuthUnaryClientInterceptor(ctx context.Context, method string, req, reply a
 }
 
 func AuthStreamClientInterceptor(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
-	tokenif := ctx.Value(rpctypes.TokenFieldNameGRPCKey{})
-	if tokenif != nil {
-		tokenCred := &proxyTokenCredential{tokenif.(string)}
+	token := getAuthTokenFromClient(ctx)
+	if token != "" {
+		tokenCred := &proxyTokenCredential{token}
 		opts = append(opts, grpc.PerRPCCredentials(tokenCred))
 	}
 	return streamer(ctx, desc, cc, method, opts...)
--- a/tests/e2e/etcd_grpcproxy_test.go
+++ b/tests/e2e/etcd_grpcproxy_test.go
@@ -17,6 +17,8 @@ package e2e
 import (
 	"context"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -26,6 +28,7 @@ import (
 	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/pkg/v3/expect"
+	v3electionpb "go.etcd.io/etcd/server/v3/etcdserver/api/v3election/v3electionpb"
 	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 	"go.etcd.io/etcd/tests/v3/framework/testutils"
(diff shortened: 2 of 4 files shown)


krijohs:
Added test case which reproduces the issue, with included fix it passes

krijohs:
Have anyone had the chance to have a look at this? pinging from reveiwers in owners file, @fuweid @ivanvc

ivanvc:
Hi @krijohs, thanks for your pull request. Ideally, we would want to discuss the issue and possible solutions before a pull request. Could you please open an issue so other members with more expertise in this area can jump in?

Thanks again.

krijohs:
Hello @ivanvc ok, got it will open an issue so possible solutions can be discussed, thanks.

ivanvc:
@krijohs, could you please rebase your branch with the latest upstream `main` branch? I'm having issues running the tests because the base is outdated. Thanks :)

krijohs:
@ivanvc sure no problem, just rebased and pushed

ivanvc (approved):
Although not an expert on this part of the code, I ran the tests without the fix to `util.go`, and they failed.

/cc @ahrtr @serathius

tests/e2e/etcd_grpcproxy_test.go
```diff
@@ -142,3 +144,97 @@ func waitForEndpointInLog(ctx context.Context, proxyProc *expect.ExpectProcess,
...
+	go func() {
+		defer wg.Done()
+		handler(events)
+	}()
+
+	clusterCli := cluster.Etcdctl(withAuth)
+	require.NoError(t, clusterCli.Put(ctx, "/test/1", "test", config.PutOptions{}))
+	require.NoError(t, err)
+
+	time.Sleep(time.Second * 2)
```
ahrtr, line 218: This seems unnecessary.
```suggestion
```

server/proxy/grpcproxy/util.go
```diff
@@ -37,7 +37,8 @@ func getAuthTokenFromClient(ctx context.Context) string {
 func withClientAuthToken(ctx, ctxWithToken context.Context) context.Context {
 	token := getAuthTokenFromClient(ctxWithToken)
 	if token != "" {
-		ctx = context.WithValue(ctx, rpctypes.TokenFieldNameGRPCKey{}, token)
+		md := metadata.Pairs(rpctypes.TokenFieldNameGRPC, token)
+		return metadata.NewOutgoingContext(ctx, md)
```
ahrtr, line 41: The change looks good.

It look like the previous implementation is wrong. All the usage on `rpctypes.TokenFieldNameGRPC` should be via the package `google.golang.org/grpc/metadata` (e.g,`metadata.FromIncomingContext`, `metadata.NewOutgoingContext`). However, grpcproxy gets the client auto token using `grpc/metadata` package, but re-add it using raw context WithValue method. 

It seems that the grpcprocy couldn't pass the client auth token to the backend etcdserver at all. Could anyone double confirm this? thx

tests/e2e/etcd_grpcproxy_test.go
krijohs: Updated, was long time ago i submitted PR, but this sleep is to ensure the token has expired when the subsequent watchers connect, to be able to reproduce issue in the e2e test

server/proxy/grpcproxy/util.go
krijohs: From testing i noticed that the proxy did pass auth tokens to the etcd server successfully for the initial client that connected. That is, the first client connecting to the proxy worked fine and could authenticate and establish the watch. But after the initial client's token expired subsequent clients that tried to connect could not.
This can be reproduced with the e2e test

krijohs:
Hi @ivanvc just checking if anyone have had time to look at this PR?

server/proxy/grpcproxy/util.go
ahrtr: You might want to change the following place to as well. It won't work anymore to get the token from context after the change in this PR.

https://github.com/etcd-io/etcd/blob/bc12c9498b824c4ddf69b489ad69176dda521c04/server/proxy/grpcproxy/util.go#L69


It was introduced in https://github.com/etcd-io/etcd/pull/8289 in 2017, and etcdserver has changed a lot, but that part of grpcproxy never be updated.

krijohs: Thanks for having a look, updated it that place as well

ahrtr:
Pls squash the commit, thx

krijohs:
> Pls squash the commit, thx

sure, just squashed and pushed

ahrtr (approved):
LGTM

Thanks

ahrtr:
@mitake would you be able to take a look at this PR as well?  Thanks.  It's related to your PR https://github.com/etcd-io/etcd/pull/8289.

ahrtr:
@krijohs can you update the e2e test to verify the stream request (e.g. watch) via grpcproxy also work when auth is enabled?

krijohs:
> @krijohs can you update the e2e test to verify the stream request (e.g. watch) via grpcproxy also work when auth is enabled?

If im understanding you correctly.
The added e2e test `TestGRPCProxyWatchersAfterTokenExpiry` in this PR already uses authenticated watch streams through the grpc-proxy, all three watches use `WithAuth("root", "rootPassword")` and `WithEndpoints(proxyClientURL)` while auth is enabled on the server.

It verifies the proxy forwards auth for streaming requests, if you like i can add a comment to make it more obvious.

ahrtr:
> The added e2e test `TestGRPCProxyWatchersAfterTokenExpiry` in this PR already uses authenticated watch streams through

why the test did not see any issue before you resolve https://github.com/etcd-io/etcd/pull/19033#discussion_r2541813945?

krijohs:
> > The added e2e test `TestGRPCProxyWatchersAfterTokenExpiry` in this PR already uses authenticated watch streams through
> 
> why the test did not see any issue before you resolve [#19033 (comment)](https://github.com/etcd-io/etcd/pull/19033#discussion_r2541813945)?

When grpc proxy sets up a new watch broadcast it calls `withClientAuthToken` in `newWatchBroadcast` so the `AuthStreamClientInterceptor` is not used for watchers from what i can see.
I can try and create e2e test that verifies the latest changes made to `AuthStreamClientInterceptor`

ahrtr:
I think if the token expires, then the test should fail (before you resolved my comment). It should work now as you have already resolved it. Can you double check this using a test case or manually? 

Refer to https://github.com/etcd-io/etcd/issues/11954

krijohs:
> I think if the token expires, then the test should fail (before you resolved my comment). It should work now as you have already resolved it. Can you double check this using a test case or manually?
> 
> Refer to #11954

Added a separate test (it can not be tested with watchers as it calls `withClientAuthToken` explicitly) that proves `AuthStreamClientInterceptor` fixes token forwarding for streaming RPCs (with the old `ctx.Value` approach it fails).

PR #19035, opened 2024-12-10
Run Member tests in parallel

AwesomePatrol:
On my local machine it brings down execution time from 5 minutes to 32s.

I added a port allocator which allocates ports in 11000-19000 range.

Issue: https://github.com/etcd-io/etcd/issues/18983

### DIFF
--- a/tests/framework/e2e/e2e.go
+++ b/tests/framework/e2e/e2e.go
@@ -83,6 +83,9 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, opts ...config.
 	default:
 		t.Fatalf("PeerTLS config %q not supported", cfg.PeerTLS)
 	}
+	if cfg.UniquePortAlloc {
+		e2eConfig.BasePort = -1
+	}
 	epc, err := NewEtcdProcessCluster(ctx, t, WithConfig(e2eConfig))
 	if err != nil {
 		t.Fatalf("could not start etcd integrationCluster: %s", err)
--- a/tests/framework/integration/cluster.go
+++ b/tests/framework/integration/cluster.go
@@ -260,13 +260,12 @@ func (c *Cluster) ProtoMembers() []*pb.Member {
 }
 
 func (c *Cluster) mustNewMember(t testutil.TB) *Member {
-	memberNumber := c.LastMemberNum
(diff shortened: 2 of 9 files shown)


tests/framework/e2e/e2e.go
```diff
@@ -83,6 +86,7 @@ func (e e2eRunner) NewCluster(ctx context.Context, t testing.TB, opts ...config.
 	default:
 		t.Fatalf("PeerTLS config %q not supported", cfg.PeerTLS)
 	}
+	e2eConfig.BasePort = 11000 + int(atomic.AddInt32(&uniqueCount, int32(cfg.ClusterSize+2)*5)%8000)
```
serathius, line 89: Can you encapsulate the port generation? Something like:
```
uniquePort := uniquePortProvider{
  min: 11000,
  max: 20000,
}

ports := uniquePort.GetBatch(5)
port.Get()
port.Get()
port.Get()
port.Get()
port.Get()
port.Get() <- expect panic
```

serathius: Batching is not even needed, as there is no requirement of ports being continuous.

serathius: Example: https://github.com/etcd-io/etcd/blob/main/tests/robustness/identity/id.go

AwesomePatrol: Added port allocator that will keep track of free ports.

tests/framework/integration/cluster.go
```diff
@@ -260,13 +260,12 @@ func (c *Cluster) ProtoMembers() []*pb.Member {
 }
 
 func (c *Cluster) mustNewMember(t testutil.TB) *Member {
-	memberNumber := c.LastMemberNum
+	uniqueNumber := atomic.AddInt32(&UniqueNumber, 1)*10 + int32(c.LastMemberNum)
```
serathius, line 263: Is this change needed?

AwesomePatrol: Yes, otherwise the names of unix sockets will collide in some cases

AwesomePatrol:
> In addition to this, have you considered using [paralleltest](https://golangci-lint.run/usage/linters/#paralleltest) ?

Thanks for your suggestion. Running tests in parallel requires more changes than just adding `t.Parallel()` calls. I would prefer to keep it optional until it becomes that simple (requires refactoring a few more tests/frameworks).

Issue #18978, opened 2024-12-01
Etcd writing error in tests about missing confstate


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Periodic failures in https://testgrid.k8s.io/sig-etcd-periodics#ci-etcd-e2e-amd64
`TestNoErrorLogsDuringNormalOperations/three_node_cluster_with_auto_tls_(peers)` is failing with:

```
Messages:   	error level log message found: {"level":"error","ts":"2024-11-16T20:58:00.500222Z","caller":"version/monitor.go:120","msg":"failed to update storage version","cluster-version":"3.6.0","error":"cannot detect storage schema version: missing confstate information","stacktrace":"go.etcd.io/etcd/server/v3/etcdserver/version.(*Monitor).UpdateStorageVersionIfNeeded\n\tgo.etcd.io/etcd/server/v3/etcdserver/version/monitor.go:120\ngo.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).monitorStorageVersion\n\tgo.etcd.io/etcd/server/v3/etcdserver/server.go:2286\ngo.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).GoAttach.func1\n\tgo.etcd.io/etcd/server/v3/etcdserver/server.go:2467"}
```

Started at least before 16 Nov based on testgrid history
Example: 
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-etcd-e2e-amd64/1857887040968855552

### What did you expect to happen?

Etcd should not write errors during startup

### How can we reproduce it (as minimally and precisely as possible)?

Return test

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

[...]

PR #19040, opened 2024-12-11
Delay the monitorStorageVersion goroutine until the server is fully ready

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/18978

```
	// The field "storageVersion" in Meta bucket was introduced in 3.6.
	// It doesn't exist in 3.5 and older versions. We depend on fields
	// "confState" and "term" to identify 3.5, as the two fields were
	// introduced in 3.5.
	// But the field "confState" is only guaranteed to be populated
	// after the member fully bootstraps itself. So we need to wait
	// for the etcdserver ready for serve client requests here.
```

### DIFF
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -16,6 +16,7 @@ package embed
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	defaultLog "log"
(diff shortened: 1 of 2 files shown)


ahrtr:
The change of `server/embed/serve.go` has been extracted into another PR https://github.com/etcd-io/etcd/pull/19041. 

For the change of `server/etcdserver/server.go`, since the error "`cannot detect storage schema version: missing confstate information`" doesn't do any harm, so let's just ignore the error in the test `TestNoErrorLogsDuringNormalOperations` instead of updating the production code. But let's also add a TODO comment to remove it in 3.7.

https://github.com/etcd-io/etcd/blob/f03dee90668d4765cdd1b14c8bd0d753ce39f8f3/server/storage/schema/schema.go#L100

Note that we don't need to depend on the field "`confstate`" and "`term`" to identify the storage version in 3.7 anymore. So we can update `UnsafeDetectSchemaVersion` to remove the operations of reading `confState` and `term` in 3.7. Refer to https://github.com/etcd-io/etcd/pull/19047

https://github.com/etcd-io/etcd/blob/f03dee90668d4765cdd1b14c8bd0d753ce39f8f3/server/storage/schema/schema.go#L93-L107

PR #19045, opened 2024-12-12
scripts: fix shellws pass

ivanvc:
The repository scripts now always have the 'sh' extension. Therefore, the find over '*.sh' matches all scripts without the need to append `./scripts/build-binary.sh`, `./scripts/build-docker.sh`, and `./scripts/release.sh`.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -340,24 +340,15 @@ function shellcheck_pass {
 }
 
 function shellws_pass {
-  TAB=$'\t'
   log_callout "Ensuring no tab-based indention in shell scripts"
   local files
-  files=$(find ./ -name '*.sh' -print0 | xargs -0 )
-  # shellcheck disable=SC2206
-  files=( ${files[@]} "./scripts/build-binary.sh" "./scripts/build-docker.sh" "./scripts/release.sh" )
-  log_cmd "grep -E -n $'^ *${TAB}' ${files[*]}"
-  # shellcheck disable=SC2086
-  if grep -E -n $'^ *${TAB}' "${files[@]}" | sed $'s|${TAB}|[\\\\tab]|g'; then
-    log_error "FAIL: found tab-based indention in bash scripts. Use '  ' (double space)."
-    local files_with_tabs
-    files_with_tabs=$(grep -E -l $'^ *\\t' "${files[@]}")
-    log_warning "Try: sed -i 's|\\t|  |g' $files_with_tabs"
-    return 1
-  else
-    log_success "SUCCESS: no tabulators found."
-    return 0
+  if files=$(find . -name '*.sh' -print0 | xargs -0 grep -E -n $'^\s*\t'); then
+    log_error "FAIL: found tab-based indention in the following bash scripts. Use '  ' (double space):"
+    log_error "${files}"
+    log_warning "Suggestion: run \"make fix\" to address the issue."
+    return 255
   fi
+  log_success "SUCCESS: no tabulators found."
 }
 
 function markdown_marker_pass {


scripts/test.sh
```diff
@@ -342,12 +342,10 @@ function shellcheck_pass {
 function shellws_pass {
   TAB=$'\t'
   log_callout "Ensuring no tab-based indention in shell scripts"
-  local files
-  files=$(find ./ -name '*.sh' -print0 | xargs -0 )
-  # shellcheck disable=SC2206
-  files=( ${files[@]} "./scripts/build-binary.sh" "./scripts/build-docker.sh" "./scripts/release.sh" )
```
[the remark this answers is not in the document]
ivanvc: First thought: We could, but we would also need to change lines 349 and 351, as they assume that files are an array.

After trying to find a more concise way of generating the list of files, I just realized that the check is not working. The check still passes if I add a tab to a shell script on the main branch. I'm working on rewriting the implementation.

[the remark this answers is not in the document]
ahrtr: > After trying to find a more concise way of generating the list of files, I just realized that the check is not working.

Do you mean the existing script doesn't work either?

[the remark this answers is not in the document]
ivanvc: This check is not working. Refer to pull request: https://github.com/etcd-io/etcd/pull/19049

And the false positive result from prow: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19049/pull-etcd-verify/1867302228360957952#1:build-log.txt%3A504-511

However, in the end, the `pull-etcd-verify` job fails due to the `fix` step, which replaces tabs with whitespace, and it leaves modified files in the git repository (https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19049/pull-etcd-verify/1867302228360957952#1:build-log.txt%3A919-925)

@ahrtr, WDYT, should we fix the check or delete it and keep it covered by `fix`? I don't have a strong opinion, but I lean toward deleting the check to avoid duplication.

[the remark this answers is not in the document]
ahrtr: Thanks for the confirmation. I think that we still need to fix it. `verify` (including `verify-shellws`) is used to verification, and `fix` is used to fix the issues detected by `verify`. Both of them should work.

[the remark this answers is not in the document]
ivanvc: Done. Expected failure with the revised code: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19049/pull-etcd-verify/1867363984722628608

scripts/test.sh
```diff
@@ -340,24 +340,14 @@ function shellcheck_pass {
...
+  if files=$(find . -name '*.sh' -print0 | xargs -0 grep -E -n $'^\s*\t'); then
     log_error "FAIL: found tab-based indention in bash scripts. Use '  ' (double space)."
-    local files_with_tabs
-    files_with_tabs=$(grep -E -l $'^ *\\t' "${files[@]}")
-    log_warning "Try: sed -i 's|\\t|  |g' $files_with_tabs"
-    return 1
-  else
-    log_success "SUCCESS: no tabulators found."
-    return 0
+    log_warning "Try: sed -i 's|\\t|  |g' $(echo "${files}" | cut -d: -f1 | paste -sd ' ')"
```
ahrtr, line 347: It has two minor issues:
- It may print duplicate files name if there are TABs in multiple lines in one file. For example,
  ```
  Try: sed -i 's|\t|  |g' ./scripts/release.sh ./scripts/release.sh
  ```
- It doesn't work on Mac OS, it should be caused by `paste`'s different behaviour.  Replacing `paste -sd ' '` with `tr '\n' ' '` can resolve this.

ivanvc: Good point. I'll always make sure to test the scripts in MacOS to ensure BSD compatibility.

However, running `make fix` also fixes this issue. IMHO, it would be more valuable to show the files and the lines with the error as output and suggest running `make fix`. I aimed to keep the same behavior, but I think this would be better.

ivanvc: i.e.,

```
'shellws' started at Fri Dec 13 11:52:19 AM PST 2024
Ensuring no tab-based indention in shell scripts
FAIL: found tab-based indention in the following bash scripts. Use '  ' (double space):
./scripts/test.sh:345:          if files=$(find . -name '*.sh' -print0 | xargs -0 grep -E -n $'^\s*\t'); then
./scripts/test.sh:351:          log_success "SUCCESS: no tabulators found."
Run make fix to address the issue.
FAIL: 'shellws' FAILED at Fri Dec 13 11:52:19 AM PST 2024
```

ahrtr: > it would be more valuable to show the files and the lines with the error as output and suggest running `make fix`

Sounds good to me.

ivanvc: Done.

scripts/test.sh
```diff
@@ -340,24 +340,15 @@ function shellcheck_pass {
...
-    files_with_tabs=$(grep -E -l $'^ *\\t' "${files[@]}")
-    log_warning "Try: sed -i 's|\\t|  |g' $files_with_tabs"
-    return 1
-  else
-    log_success "SUCCESS: no tabulators found."
-    return 0
+  if files=$(find . -name '*.sh' -print0 | xargs -0 grep -E -n $'^\s*\t'); then
+    log_error "FAIL: found tab-based indention in the following bash scripts. Use '  ' (double space):"
+    log_error "${files}"
+    log_warning "Suggestion: running \"make fix\" addresses the issue."
```
ahrtr, line 348:
```suggestion
    log_warning "Suggestion: running \"make fix\" to address the issue."
```

ahrtr (approved):
LGTM with minor comment.

Thanks

PR #19046, opened 2024-12-12
Set a 80% threshold for the codecov status.patch

ahrtr:
Basically we depend on maintainers to decide whether we should add test to improve the coverage for each PR, although we set a 1% threshold for now.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -30,6 +30,10 @@ coverage:
         # allow some coverage reductions within a threshold
         # this allows a 1% drop from the previous base commit coverage
         threshold: 1%
+    patch:
+      default:
+        target: auto
+        threshold: 80%
 comment:
   layout: "header, files, diff, footer"
   behavior: default # default: update, if exists. Otherwise post new; new: delete old and post new


ahrtr:
cc @ivanvc @jmhbnz @serathius

ahrtr:
cc @fuweid

fuweid (approved):

ahrtr:
The PR doesn't fix the failure of `codecov/patch` in https://github.com/etcd-io/etcd/pull/19041. @ivanvc @jmhbnz do you know how to adjust the parameter to fix it? Or we  can just disable the [patch status](https://docs.codecov.com/docs/commit-status#disabling-a-status).

ivanvc:
> The PR doesn't fix the failure of `codecov/patch` in #19041. @ivanvc @jmhbnz do you know how to adjust the parameter to fix it? Or we can just disable the [patch status](https://docs.codecov.com/docs/commit-status#disabling-a-status).

@ahrtr, after reading the documentation, I think the behavior that we want is to set the threshold to 99% rather than 1%. Can you try that on your other pull request to see if there's a failure?

ahrtr:
> @ahrtr, after reading the documentation, I think the behavior that we want is to set the threshold to 99% rather than 1%. Can you try that on your other pull request to see if there's a failure?

Thanks. Just updated https://github.com/etcd-io/etcd/pull/19041, Let's see what's the result. Most likely it should work. Probably we should just disable it.

ahrtr:
Just changed to 80%, at least it's better than off/disabled. As mentioned above, we still need to depend on Maintainers to whether we should add test to improve the coverage for each PR.

ahrtr:
@ivanvc yes, it works in https://github.com/etcd-io/etcd/pull/19041. 

Please take a look at this PR, let's get it merged firstly. thx

ivanvc (approved):
LGTM. Thanks, @ahrtr.

PR #19057, opened 2024-12-13
Remove duplicated <-s.ReadyNotify()

ahrtr:
Follow up to https://github.com/etcd-io/etcd/pull/19041

It was a mistake that I made when debugging the `codecov/patch` failure in https://github.com/etcd-io/etcd/pull/19041.  Sorry. Also refer to https://github.com/etcd-io/etcd/pull/19046#issuecomment-2539373718

### DIFF
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -101,7 +101,6 @@ func (sctx *serveCtx) serve(
 	gopts ...grpc.ServerOption,
 ) (err error) {
 	logger := defaultLog.New(io.Discard, "etcdhttp", 0)
-	<-s.ReadyNotify()
 
 	select {
 	case <-s.StoppingNotify():


serathius (approved):

serathius:
Should we add a test for shutdown during initialization?

ahrtr:
> Should we add a test for shutdown during initialization?

It's good to have, but hard to test. The problem is that etcd doesn't wait for the following goroutine to complete or exit when it's stopping. So from product functionality perspective, we don't see any issue. This PR (and previous one) just avoids some unnecessary operations when the server is stopping.

https://github.com/etcd-io/etcd/blob/026bbecd25e0f9312b99175940a6761580cf5643/server/embed/etcd.go#L777-L779

Even the goroutine leakage detection can't detect it either, because a golang program execution does not wait for the invoked function to complete.

https://github.com/etcd-io/etcd/blob/026bbecd25e0f9312b99175940a6761580cf5643/client/pkg/testutil/leak.go#L101-L109

ahrtr:
An related but different case: I recall that user applications might run into a deadlock in embeded use case if etcd is being stopped during bootstrap. Raised https://github.com/etcd-io/etcd/issues/19058

PR #19060, opened 2024-12-13
fix(e2e): ignore error log about failed storage update

ghouscht:
Should fix https://github.com/etcd-io/etcd/issues/18978 as suggested by @ahrtr

### DIFF
--- a/tests/e2e/logging_test.go
+++ b/tests/e2e/logging_test.go
@@ -37,7 +37,12 @@ func TestNoErrorLogsDuringNormalOperations(t *testing.T) {
 				e2e.WithClusterSize(1),
 				e2e.WithLogLevel("debug"),
 			},
-			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
+			allowedErrors: map[string]bool{
+				"setting up serving from embedded etcd failed.": true,
+				// See https://github.com/etcd-io/etcd/pull/19040#issuecomment-2539173800
+				// TODO: Remove with etcd 3.7
+				"cannot detect storage schema version: missing confstate information": true,
+			},
 		},
 		{
 			name: "three node cluster",
@@ -45,7 +50,12 @@ func TestNoErrorLogsDuringNormalOperations(t *testing.T) {
 				e2e.WithClusterSize(3),
 				e2e.WithLogLevel("debug"),
 			},
-			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
+			allowedErrors: map[string]bool{
+				"setting up serving from embedded etcd failed.": true,
+				// See https://github.com/etcd-io/etcd/pull/19040#issuecomment-2539173800
+				// TODO: Remove with etcd 3.7
+				"cannot detect storage schema version: missing confstate information": true,
+			},
 		},
 		{
 			name: "three node cluster with auto tls (all)",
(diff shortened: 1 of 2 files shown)


tests/e2e/logging_test.go
```diff
@@ -37,15 +37,22 @@ func TestNoErrorLogsDuringNormalOperations(t *testing.T) {
 				e2e.WithClusterSize(1),
 				e2e.WithLogLevel("debug"),
 			},
-			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
+			allowedErrors: map[string]bool{
+				"setting up serving from embedded etcd failed.": true,
+			},
```
ahrtr, line 42: Why don't we ignore the error in such case? Suggest to add it into this case as well.

tests/e2e/logging_test.go
```diff
@@ -37,15 +37,22 @@ func TestNoErrorLogsDuringNormalOperations(t *testing.T) {
...
 			options: []e2e.EPClusterOption{
 				e2e.WithClusterSize(3),
 				e2e.WithLogLevel("debug"),
 			},
-			allowedErrors: map[string]bool{"setting up serving from embedded etcd failed.": true},
+			allowedErrors: map[string]bool{
+				"setting up serving from embedded etcd failed.": true,
+				// See https://github.com/etcd-io/etcd/pull/19040#issuecomment-2539173800
+				// TODO: Remove with etcd 3.7
+				"failed to update storage version": true,
```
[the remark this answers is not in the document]
ghouscht: The test seems to check the `msg` not `error` object in the logged JSON: https://github.com/ghouscht/etcd/blob/ignore-missing-err-confstate/tests/e2e/logging_test.go#L136-L139

tests/e2e/logging_test.go
ghouscht: We can ignore it here as well, but according to testgrid there was no failure for this test. Only the three node cases were affected from what I can see.

tests/e2e/logging_test.go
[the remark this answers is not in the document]
ahrtr: I think we need to check both the msg and error. In this case, it makes more sense to ignore the specific error message `cannot detect storage schema version: missing confstate information`.

We can add one more field "`Error`" into `logEntry`.

```
type logEntry struct {
	Level     string `json:"level"`
	Timestamp string `json:"ts"`
	Caller    string `json:"caller"`
	Message   string `json:"msg"`
}
```

tests/e2e/logging_test.go
ahrtr: In theory, one member may still run into the issue, but just with a low possibility.

ghouscht: Ok, I added the exclusion for this test as well.

tests/e2e/logging_test.go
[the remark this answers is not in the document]
ghouscht: Adapted the code to check the Message and Error for a match. If one of the two matches exactly the error is ignored. Hope that is ok

ahrtr (approved):
LGTM

Thank you!

ahrtr:
cc @ivanvc @jmhbnz @serathius

ivanvc (approved):
I'm not an expert on this, but it LGTM. Anything that helps with flakes is greatly appreciated. Thanks, @ghouscht.

jmhbnz (approved):
LGTM - Many thanks for working on this @ghouscht 🙏🏻

PR #19077, opened 2024-12-16
backend: Remove logger nil checks

### LINKED ISSUE #19015: Possible missing nil checks
### Bug report criteria

### What happened?

After having been compared to a nil value at backend.go:515, pointer 'b.lg' is passed as implicit 'this' parameter in call to function 'Error' at backend.go:545 or pointer 'b.lg' is passed as implicit 'this' parameter in call to function 'Fatal' at backend.go:532. 
```
if b.lg != nil {
		b.lg.Info(
			"defragmenting",
			zap.String("path", dbp),
			zap.Int64("current-db-size-bytes", size1),
			zap.String("current-db-size", humanize.Bytes(uint64(size1))),
			zap.Int64("current-db-size-in-use-bytes", sizeInUse1),
			zap.String("current-db-size-in-use", humanize.Bytes(uint64(sizeInUse1))),
		)
	}

	defer func() {
		// NOTE: We should exit as soon as possible because that tx
		// might be closed. The inflight request might use invalid
		// tx and then panic as well. The real panic reason might be
		// shadowed by new panic. So, we should fatal here with lock.
		if rerr := recover(); rerr != nil {
			b.lg.Fatal("unexpected panic during defrag", zap.Any("panic", rerr))
		}
	}()

	// Commit/stop and then reset current transactions (including the readTx)
	b.batchTx.unsafeCommit(true)
	b.batchTx.tx = nil

	// gofail: var defragBeforeCopy struct{}
	err = defragdb(b.db, tmpdb, defragLimit)
	if err != nil {
		tmpdb.Close()
		if rmErr := os.RemoveAll(tmpdb.Path()); rmErr != nil {
			b.lg.Error("failed to remove db.tmp after defragmentation completed", zap.Error(rmErr))
		}

		// restore the bbolt transactions if defragmentation fails
		b.batchTx.tx = b.unsafeBegin(true)
		b.readTx.tx = b.unsafeBegin(false)

		return err
	}
```
Please clarify whether verification is required in this case? Do you need to make any changes/corrections to the code?

It might be worth adding additional checks for b.lg, if required.

Checking code

_No response_

Latest

<details>

</details>

<details>

</details>

### Relevant log output

_No response_

srujangit123:
To remove the ugly nil checks by instantiating the logger with a no-op logger if it is nil.

Fixes #19015

### DIFF
--- a/server/storage/backend/backend.go
+++ b/server/storage/backend/backend.go
@@ -28,6 +28,7 @@ import (
 	"go.uber.org/zap"
 
 	bolt "go.etcd.io/bbolt"
+	"go.etcd.io/etcd/client/pkg/v3/verify"
 )
 
 var (
@@ -187,6 +188,11 @@ func newBackend(bcfg BackendConfig) *backend {
 	if boltOpenOptions != nil {
 		*bopts = *boltOpenOptions
 	}
+
+	if bcfg.Logger == nil {
+		bcfg.Logger = zap.NewNop()
+	}
+
 	bopts.InitialMmapSize = bcfg.mmapSize()
 	bopts.FreelistType = bcfg.BackendFreelistType
 	bopts.NoSync = bcfg.UnsafeNoFsync
@@ -458,6 +464,7 @@ func (b *backend) Defrag() error {
 }
 
 func (b *backend) defrag() error {
+	verify.Assert(b.lg != nil, "the logger should not be nil")
 	now := time.Now()
 	isDefragActive.Set(1)
 	defer isDefragActive.Set(0)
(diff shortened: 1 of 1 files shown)


ahrtr:
Please add an assert at the first line of `(b *backend) defrag()`, something like below,

https://github.com/etcd-io/etcd/blob/b4450d510ddf0f011c874c28e95175b4bac7f6c8/pkg/osutil/interrupt_unix.go#L51

ahrtr:
cc @fuweid  @ivanvc @jmhbnz @serathius

server/storage/backend/backend.go
```diff
@@ -199,6 +200,10 @@ func newBackend(bcfg BackendConfig) *backend {
 		bcfg.Logger.Panic("failed to open database", zap.String("path", bcfg.Path), zap.Error(err))
 	}
 
+	if bcfg.Logger == nil {
```
fuweid, line 203: I don't think it works. Both the `newBoltLoggerZap` call and Line 200 `panic` already use `bcfg.Logger`.
That assignment should be placed at the beginning of the `newBackend`

ahrtr: Good catch!

ahrtr (changes requested):
https://github.com/etcd-io/etcd/pull/19077/files#r1889186695

ahrtr (approved):

ivanvc (approved):
LGTM. Thanks, @srujangit123.

serathius (approved):

PR #13405, opened 2021-10-08
Implement single node downgrades

serathius:
This PR implement single downgrades as proposed in https://docs.google.com/document/d/1yD0GDkxqWBPAax6jLZ97clwAz2Gp0Gux6xaTrtJ6wHE/edit?usp=sharing with the goal of introducing e2e tests that can confirm that storage versioning properly validates WAL entries during downgrade. 

This doesn't mean that with this PR etcd supports downgrades, there are still a lot of testing, small problems that we need to fix before  we can say that downgrades are safe. This is meant to allow us to expand testing of downgrades with different scenarios to confirm it's reliability.

Problem detected during implementation that will need to be fixed:
* As Etcd v3.5 imminently panics on SetClusterVersion with version set "3.6" entry in WAL I added it to MinEtcdVersion logic. We should consider adding logic to `etcdutl migrate` to drop this entry. 
* Sometimes setting ClusterVersion after upgrade timeouts, we should debug why.
* I didn't implement snapshoting WAL after lowering cluster version (required to remove non-backward compatible entries). To work around I use very low snapshot count in tests.

cc @ptabor @lilic

PR #19081, opened 2024-12-18
Add more info in the error message in downgrade detection

ahrtr:
Add more info in the error message. 

Also read https://github.com/etcd-io/etcd/pull/13405#discussion_r1890378185


serathius (approved):

serathius:
Right, maybe we could make it a warning as the procedure is repeating and inform user that it might happen and we will try again.

ahrtr:
> maybe we could make it a warning as the procedure

Makes sense to change this specific case as a warning. But there is a long call stack between [generating the error/warning message](https://github.com/etcd-io/etcd/blob/e4d1da73e17d142ced6d4955f49a64367324e232/server/storage/schema/schema.go#L76) and [printing it as an error log](https://github.com/etcd-io/etcd/blob/e4d1da73e17d142ced6d4955f49a64367324e232/server/etcdserver/version/monitor.go#L120). There are multiple reasons for the error log, we can't just change it as a warning log.

PR #19082, opened 2024-12-18
[tools/etcd-dump-logs] Fix the usage of --start-index in etcd-dump-logs

ahrtr:
Currently if users explicitly set --start-index=0, it will be ignored. It isn't expected, we should respect it as long as users explicitly set it.


ahrtr:
@fuweid @ivanvc @jmhbnz @serathius

fuweid (approved):

PR #19092, opened 2024-12-19
Fix lease expiration check

### LINKED ISSUE #19091: Client lease library does not check for expiration
### Bug report criteria

### What happened?

The `leasing` library fails a new test case that does 
```
client1.Put("k", "abc");
client1.Get("k"); // to add it to client1's cache
client2.Put("k", "def");
assert(client1.Get("k") == "def") // sometimes fails, seeing the stale "abc"
```
with some delays and failures added in between.
This is because the client `Lease` and `Session` libraries do not provide a method to check if a lease is currently unexpired, and so the `leasing` package does not correctly check that its lease is valid.

Based on `leasing` documentation, reads should be linearizable, which means the above sequence of operations should cause the final `assert` to succeed.
https://github.com/etcd-io/etcd/blob/9fa35e53f429ca8f21b0d6b26f24e1848f2652a6/client/v3/leasing/doc.go#L15

The new failing test is in forked [tests/integration/clientv3/lease/leasing_test.go](https://github.com/upamanyus/etcd/blob/d065a4a9eeff4343bb02546c3d09fd9318c8f670/tests/integration/clientv3/lease/leasing_test.go#L419-L491):
```
func TestLeasingGetChecksForExpiration(t *testing.T) {
	integration2.BeforeTest(t)

	ttl := 6

	// There's no way to partition a client from the server, so use multiple
	// servers and shut down a single server to partition the client from the
	// system.
	clus := integration2.NewCluster(t, &integration2.ClusterConfig{Size: 3, UseBridge: true})
	defer clus.Terminate(t)

	// Try to make sure server 0 is not the leader
	if clus.Members[0].Server.Leader() == clus.Members[0].Server.MemberID() {
		err := clus.Members[0].Server.MoveLeader(context.TODO(),
			clus.Members[0].Server.Lead(),
			uint64(clus.Members[1].Server.MemberID()))
		if err != nil {
			panic(err)
		}
	}
	time.Sleep(2 * time.Second)

	leaderID := clus.Members[0].Server.Leader()
	if leaderID == clus.Members[0].Server.MemberID() {
		panic("test wants 0 to not be leader")
	}

	// This client will get partitioned away from the system (by killing the
	// node it's connected to).
	lkv0, closeLKV0, err := leasing.NewKV(clus.Client(0), "pfx/", concurrency.WithTTL(ttl))
	require.NoError(t, err)
	defer closeLKV0()

	lkv1, closeLKV1, err := leasing.NewKV(clus.Client(1), "pfx/")
	require.NoError(t, err)
	defer closeLKV1()

	if _, err = lkv0.Put(context.TODO(), "k", "abc"); err != nil {
		t.Fatal(err)
	}
	if _, err = lkv0.Get(context.TODO(), "k"); err != nil {
		t.Fatal(err)
	}

	time.Sleep(6500 * time.Millisecond)
	clus.Members[0].Stop(t)

	if _, err = lkv1.Put(context.TODO(), "k", "def"); err != nil {
		t.Fatal(err)
	}

	resp, err := lkv1.Get(context.TODO(), "k")
	if err != nil {
		t.Fatal(err)
	}
	if len(resp.Kvs) != 1 || string(resp.Kvs[0].Value) != "def" {
		t.Fatalf(`expected "k"->"def" from lkv1, got response %+v`, resp)
	}

	go func() {
		// Eventually bring back the server so the disconnected client can
		// finish its last `Get()`.
		time.Sleep(1 * time.Second)
		clus.Members[0].Restart(t)
	}()
	cachedResp, err := lkv0.Get(context.TODO(), "k")
	if err != nil {
		t.Fatal(err)
	}
	if len(cachedResp.Kvs) != 1 || string(cachedResp.Kvs[0].Value) != "def" {
		t.Fatalf(`expected "k"->"def", got response %+v`, cachedResp)
	}
}
```

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.6.0-alpha.0
Git SHA: d065a4a9e
Go Version: go1.23.4
Go OS/Arch: linux/amd64

$ etcdctl version
API version: 3.6
```

</details>

_No response_

_No response_

### Relevant log output

_No response_

upamanyus:
Fixes #19091.

This PR adds and uses a client-side `Unexpired()` method that determines if a lease is valid by checking if its expiration time is after the current time. This helps fix a bug in the client `leasing` library, and provides a general API to check if a lease is valid, which was not previously exposed to users.

This adds a new test for `leasing` which does a few Puts and Gets with delays and with a client that gets network partitioned. The test fails often on the old `leasing` implementation. The bug within `leasing` was here:
https://github.com/etcd-io/etcd/blob/9fa35e53f429ca8f21b0d6b26f24e1848f2652a6/client/v3/leasing/kv.go#L469-L473
This code tries to determine if a session is "ready" by checking if an non-blocking receive on `lkv.session.Done()` fails. However, failing to immediately receive on a `chan` does not guarantee anything: the other side's send or `close` may just have been delayed by a bit.

Fortunately for reproducing the problem, there's a delay already present in the code: the `Done()` channel is closed by a background loop that sleeps and periodically checks whether the lease is expired. If the lease expires while this loop is sleeping, there is a delay between the true expiration time and when the `Done()` channel is closed, which seems to make the test fail relatively reliably (it still sometimes takes a few tries to get a failure).

Using `Unexpired()`, the test always appears to pass.

I couldn't find other places using etcd leases with the problematic pattern from the `leasing` library, but I'm admittedly not sure how the `Lease` library is used by others.

### DIFF
--- a/client/v3/lease.go
+++ b/client/v3/lease.go
@@ -136,6 +136,10 @@ type Lease interface {
 	// (see https://github.com/etcd-io/etcd/pull/7866)
 	KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAliveResponse, error)
 
+	// Unexpired returns true iff the lease is unexpired (more precisely: iff the
+	// lease was unexpired during the execution of the Unexpired() call).
+	Unexpired(id LeaseID) bool
+
 	// KeepAliveOnce renews the lease once. The response corresponds to the
 	// first message from calling KeepAlive. If the response has a recoverable
 	// error, KeepAliveOnce will retry the RPC with a new keep alive message.
@@ -623,4 +627,14 @@ func (ka *keepAlive) close() {
 	for _, ch := range ka.chs {
 		close(ch)
 	}
+}
+
+func (l *lessor) Unexpired(id LeaseID) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	ka, ok := l.keepAlives[id]
+	if !ok {
+		return false
+	}
+	return ka.deadline.After(time.Now())
 }
--- a/client/v3/lease.go
+++ b/client/v3/lease.go
@@ -136,9 +136,9 @@ type Lease interface {
 	// (see https://github.com/etcd-io/etcd/pull/7866)
 	KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAliveResponse, error)
 
-	// Unexpired returns true iff the lease is unexpired (more precisely: iff the
-	// lease was unexpired during the execution of the Unexpired() call).
-	Unexpired(id LeaseID) bool
+	// Expired returns true iff the lease is expired (more precisely: iff the
+	// lease was expired during the execution of the Expired() call).
+	Expired(id LeaseID) bool
 
 	// KeepAliveOnce renews the lease once. The response corresponds to the
 	// first message from calling KeepAlive. If the response has a recoverable
@@ -629,12 +629,12 @@ func (ka *keepAlive) close() {
 	}
 }
 
-func (l *lessor) Unexpired(id LeaseID) bool {
+func (l *lessor) Expired(id LeaseID) bool {
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	ka, ok := l.keepAlives[id]
 	if !ok {
-		return false
+		return true
 	}
-	return ka.deadline.After(time.Now())
+	return ka.deadline.Before(time.Now())
 }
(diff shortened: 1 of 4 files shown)


client/v3/lease.go
```diff
@@ -136,6 +136,10 @@ type Lease interface {
 	// (see https://github.com/etcd-io/etcd/pull/7866)
 	KeepAlive(ctx context.Context, id LeaseID) (<-chan *LeaseKeepAliveResponse, error)
 
+	// Unexpired returns true iff the lease is unexpired (more precisely: iff the
+	// lease was unexpired during the execution of the Unexpired() call).
+	Unexpired(id LeaseID) bool
```
serathius, line 141: Let's avoid negative logic in the interface. `!Expired(id)` should be simpler.

PR #19093, opened 2024-12-20
Prevent leaking file descriptor during snapshotting and provide better logging of errors

serathius:
cc @ahrtr @fuweid @ivanvc 

Found case where a leaked file descriptor prevents data being deleted from disk and filling up the disk.

Would like to backport it to all v3.5 and v3.4

### DIFF
--- a/client/v3/snapshot/v3_snapshot.go
+++ b/client/v3/snapshot/v3_snapshot.go
@@ -17,6 +17,7 @@ package snapshot
 import (
 	"context"
 	"crypto/sha256"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


ahrtr:
We need to close all previous similar PRs, one of which is https://github.com/etcd-io/etcd/pull/18200

client/v3/snapshot/v3_snapshot.go
```diff
@@ -54,38 +55,58 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
...
 
-	var f *os.File
-	f, err = os.OpenFile(partpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, fileutil.PrivateFileMode)
+	f, err := os.OpenFile(partpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, fileutil.PrivateFileMode)
 	if err != nil {
 		return "", fmt.Errorf("could not open %s (%w)", partpath, err)
 	}
+	defer func() {
+		err = f.Close()
+		if err != nil && !errors.Is(err, os.ErrClosed) {
```
ahrtr, line 79: It works. But I think a better and more explicit way is to use `sync.Once` to ensure f.Close is only executed once. So that it's well aware that we need to close the file before renaming, but also need to ensure it's closed in case early return due to any error.

fuweid: Checked https://pkg.go.dev/os#File.Close and the api can guarantee it can be executed only once. `sync.Once` is not neccesary here.

client/v3/snapshot/v3_snapshot.go
```diff
@@ -98,5 +119,5 @@ func SaveWithVersion(ctx context.Context, lg *zap.Logger, cfg clientv3.Config, d
 		return resp.Version, fmt.Errorf("could not rename %s to %s (%w)", partpath, dbPath, err)
 	}
 	lg.Info("saved", zap.String("path", dbPath))
-	return resp.Version, nil
+	return resp.Version, err
```
fuweid, line 122: just curious about using `err` here. `nil` is more accurate

fuweid (approved):
LGTM

client/v3/snapshot/v3_snapshot.go
ahrtr: Not a big deal, please feel free to merge this PR and backport to 3.5 and 3.4.

client/v3/snapshot/v3_snapshot.go
serathius: Done

client/v3/snapshot/v3_snapshot.go
serathius: SGTM

ahrtr (approved):

ahrtr:
The PR fixed https://github.com/etcd-io/etcd/issues/18085

PR #19038, opened 2024-12-10
Add MemberDowngrade failpoint

siyuanfoundation:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

https://github.com/etcd-io/etcd/issues/17118

PR #19095, opened 2024-12-21
Still return continuous WAL entries when running into ErrSliceOutOfRange

ahrtr:
Please read https://github.com/etcd-io/etcd/pull/19038#issuecomment-2555157934 and  https://github.com/etcd-io/etcd/pull/19038#issuecomment-2557414292

cc @serathius @siyuanfoundation


ahrtr:
Confirmed that this PR can fix the error in https://github.com/etcd-io/etcd/pull/19038#issuecomment-2532919530. @siyuanfoundation please let me know if you can still reproduce it in your environment.

ahrtr:
> I can confirm this fixes the failure in #19038. Thank you @ahrtr !

Thanks for the confirmation.

Can we get this merged firstly? PTAL cc @serathius

ahrtr:
cc @fuweid  @ivanvc @jmhbnz

siyuanfoundation:
I just found the robustness test still fails even with this fix. Because `validatePersistedRequestMatchClientRequests` requires the `lastOp` to be persisted, partial WAL entries would not work for this check. 
I got the error of:
```
last succesful client write {"Type":"txn","LeaseGrant":null,"LeaseRevoke":null,"Range":null,"Txn":{"Conditions":null,"OperationsOnSuccess":[{"Type":"put-operation","Range":{"Start":"","End":"","Limit":0},"Put":{"Key":"tombstone","Value":{"Value":"true","Hash":0},"LeaseID":0},"Delete":{"Key":""}}],"OperationsOnFailure":null},"Defragment":null,"Compact":null} was not persisted, required to validate

```

ahrtr:
> After syncing my repo, I just found the robustness test still fails even with this fix. Because `validatePersistedRequestMatchClientRequests` requires the `lastOp` to be persisted, partial WAL entries would not work for this check. I got the error of:
> 
> ```
> last succesful client write {"Type":"txn","LeaseGrant":null,"LeaseRevoke":null,"Range":null,"Txn":{"Conditions":null,"OperationsOnSuccess":[{"Type":"put-operation","Range":{"Start":"","End":"","Limit":0},"Put":{"Key":"tombstone","Value":{"Value":"true","Hash":0},"LeaseID":0},"Delete":{"Key":""}}],"OperationsOnFailure":null},"Defragment":null,"Compact":null} was not persisted, required to validate
> ```

@siyuanfoundation how often did you see this error? Or in other words, is it easy to reproduce this error?

If I understood it correctly, the robustness test error means that the last client write which already got successful response, but it wasn't persisted in WAL file. Please let me know if I misunderstood it.

Each time when we see an issue, the first thing is to figure out whether it's a real issue from end user perspective. can you double check whether the last successful client write was persisted in bbolt db?

Also I see that robustness test might not process the WAL records correctly, the longest one might not be he correct one. As long as the WAL records were not committed yet, they may be overwritten by following WAL records.
https://github.com/etcd-io/etcd/blob/fce823ac2830033270f8fe03fa1b56e62bf882b8/tests/robustness/report/wal.go#L78-L79

I regard it as a test issue for now, please raise a separate issue to track it. Thanks.

serathius:
> Also I see that robustness test might not process the WAL records correctly, the longest one might not be he correct one. As long as the WAL records were not committed yet, they may be overwritten by following WAL records.

You are right that normally longest WAL is not necessarily include the longest commit sequence, however in robustness test we explicitly make a single additional transaction after the test is finished, this should ensure that there are no any other uncommitted transactions. We require the transaction to succeed and later use it to assert that WAL is complete.

serathius (approved):
Change looks good, however I haven't validated how it works with repair.

ahrtr:
@siyuanfoundation I am a little confused, probably I did not say it clearly.

There are two errors. One is https://github.com/etcd-io/etcd/pull/19038#issuecomment-2532919530, and it's already confirmed that this PR can fix it. Please let me know if can still see the error with the patch included in this PR.

The second error is https://github.com/etcd-io/etcd/pull/19095#issuecomment-2569911236. A successful client write **must** have been persisted in the WAL files at least majority of the members, and **probably** also in bbolt DB. This is exactly I was requesting to double confirm manually as mentioned in https://github.com/etcd-io/etcd/pull/19095#issuecomment-2571273156.

Also

> You are right that normally longest WAL is not necessarily include the longest commit sequence, however in robustness test we explicitly make a single additional transaction after the test is finished, this should ensure that there are no any other uncommitted transactions. We require the transaction to succeed and later use it to assert that WAL is complete.

@serathius thx for the clarification. But theoretically it's still possible that the longest one isn't the correct one. The single additional successful transaction you mentioned only guarantees that majority members have the correct WAL data. Also the last successful client write was a `LeaseGrant` mentioned in https://github.com/etcd-io/etcd/pull/19095#issuecomment-2569911236 instead of the final additional transaction you mentioned (see below). Did I miss anything?

https://github.com/etcd-io/etcd/blob/70a172696d8e44a20fded6074cb9e9682bcfe628/tests/robustness/traffic/traffic.go#L119-L121

ahrtr:
> Also the last successful client write was a `LeaseGrant` mentioned in [#19095 (comment)](https://github.com/etcd-io/etcd/pull/19095#issuecomment-2569911236) instead of the final additional transaction you mentioned (see below).

OOPS, It seems that it's indeed the final additional transaction.

`"Put":{"Key":"tombstone","Value":{"Value":"true","Hash":0},"LeaseID":0}`

ahrtr:
Let me merge this PR firstly, since it's already confirmed that it can resolve the first error. Regarding the second error, it should be a test issue.

siyuanfoundation:
> @siyuanfoundation I am a little confused, probably I did not say it clearly.
> 
> There are two errors. One is [#19038 (comment)](https://github.com/etcd-io/etcd/pull/19038#issuecomment-2532919530), and it's already confirmed that this PR can fix it. Please let me know if can still see the error with the patch included in this PR.
> 

The statement `it's already confirmed that this PR can fix it` is no longer true. Before applying this PR, I am seeing the WAL error, with this PR, the same downgrade robustness test no longer passes like I tested before, but changed the error msg to not finding the last commit. The two errors are the same under the hood because of missing WAL entries in the persisted file.

ahrtr:
> but changed the error msg to not finding the last commit. The two errors are the same under the hood because of missing WAL entries in the persisted file.

They are not the same error. Even without this PR, the robustness test still has the second error. The reason why you did not see it before is that it's hidden by the first error. If you really understood my previous comment, the robustness's way of reading WAL is wrong.

ahrtr:
Just raised https://github.com/etcd-io/etcd/issues/19147

Issue #18993, opened 2024-12-03
[3.5] Add online & offline commands to check whether the v2store contains custom content


ahrtr:
### What would you like to be added?

Part of https://github.com/etcd-io/etcd/issues/12913

etcd 3.6 will panic if the v2store contains custom content. Before users upgrade from 3.5 to 3.6, they must ensure there is no any custom content in v2store; and both the online and offline commands need to check it. For the offline command, it also needs to check the WAL records after the latest v2 snapshot file.

The proposed commands:
```
etcdctl check v2store
etcdutl check v2store
```

EDIT: We only need to add the offline tool. For online tool, Users should be able to use the existing etcdctl v2 command `ETCDCTL_API=2 ./bin/etcdctl ls` directly, refer to https://github.com/etcd-io/etcd/pull/19114#issuecomment-2564688453

~~Also `--enable-v2` and `--experimental-enable-v2v3` have already been removed i 3.6, so users must ensure that they do not configure these two flags before upgrading to 3.6. Only the online command checks it.~~

### Why is this needed?

To ensure smooth upgrade from 3.5 to 3.6.

ahrtr:
For the command-line flags, let's take care of it in https://github.com/etcd-io/etcd/pull/18999. This issue focuses on checking the v2store.

serathius:
Would it be better to integrate it into v3.6? Before starting etcd v3.6 we can scan for v2state and abort startup. Would prefer to minimize the number of steps needed to upgrade etcd.

ahrtr:
> Would it be better to integrate it into v3.6?

This tool is only targeted for 3.5. Users need to check the v2store before upgrading to 3.6 instead of after. But users don't have to run this tool if they have already removed or disabled the `--enable-v2` for a long time.

In 3.6, etcd already has the function to check the v2store and will fail the start if v2store has custom content.
https://github.com/etcd-io/etcd/blob/fbac2a064c381d6f8c74c4d9d01fde816e200eef/server/etcdserver/bootstrap.go#L407-L409

ahrtr:
@fuweid any progress on this feature?

fuweid:
> any progress on this feature?

@ahrtr I'm still working on it. Will file pull request once it's ready.

ahrtr:
Done in https://github.com/etcd-io/etcd/pull/19113

PR #19114, opened 2024-12-29
[3.5] *: support custom content check online in v2store

fuweid:
It's to use v2 API to list all keys in v2store. If there is any active key in v2store, `etcdctl check v2store` returns error. However, the last snapshot might still contain custom content. The end-user still needs to check that content offline by `etcdutl check v2store`.

Part of #18993

### DIFF
--- a/etcdctl/ctlv3/command/check.go
+++ b/etcdctl/ctlv3/command/check.go
@@ -113,6 +113,7 @@ func NewCheckCommand() *cobra.Command {
 
 	cc.AddCommand(NewCheckPerfCommand())
 	cc.AddCommand(NewCheckDatascaleCommand())
+	cc.AddCommand(NewCheckV2StoreCommand())
 
 	return cc
 }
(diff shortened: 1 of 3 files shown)


ahrtr:
Users should be able to use the existing etcdctl v2 command `ETCDCTL_API=2 ./bin/etcdctl ls` directly,

https://github.com/etcd-io/etcd/blob/762e93874f23213c3b3d918ba82ce2803bee3d46/etcdctl/ctlv2/ctl.go#L63

Firstly start etcd with v2 enabled,
```
$ ./bin/etcd  --enable-v2
```

Then run commands below,
```
$ ETCDCTL_API=2 ./bin/etcdctl ls
//Nothing returned

$ ETCDCTL_API=2 ./bin/etcdctl set k1 v1
v1

$ ETCDCTL_API=2 ./bin/etcdctl ls
/k1

$ ETCDCTL_API=2 ./bin/etcdctl set k2 v2
v2

$ ETCDCTL_API=2 ./bin/etcdctl ls
/k2
/k1
```

If v2 isn't enabled at all, 
```
$ ETCDCTL_API=2 ./bin/etcdctl ls
Error:  client: response is invalid json. The endpoint is probably not valid etcd cluster endpoint
```

fuweid:
> Users should be able to use the existing etcdctl v2 command ETCDCTL_API=2 ./bin/etcdctl ls directly,

Yes. The v2 api is that only way to check v2store online. Let me close it. Thanks for the comment @ahrtr

Issue #18340, opened 2024-07-17
Etcd send a corrupt snapshot or missing hash snapshot to a snapshot api call which causes the restoration to fail.


ishan16696:
### Bug report criteria

- [X] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [X] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [X] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [X] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

It has been observed that during the restoration of etcd cluster from the etcd snapshot (taken via [snapshot](https://github.com/etcd-io/etcd/blob/e7f572914d79f0705b3dc8ca28d9a14b0f854d49/client/v3/maintenance.go#L80) api call) that snapshot was missing the hash value or snapshot was corrupted, which caused the restoration to fail.

```
FATA[0002] Failed to restore snapshot: failed to restore from the base snapshot: snapshot missing hash but --skip-hash-check=false
```

### What did you expect to happen?

Is there a way to detect corruption or missing hash of snapshot  (taken via [snapshot](https://github.com/etcd-io/etcd/blob/e7f572914d79f0705b3dc8ca28d9a14b0f854d49/client/v3/maintenance.go#L80)) early rather than waiting till restoration ?

I have following methods in my mind:
1. Take a snapshot via snapshot api call, start a embedded etcd with different data-dir path and try to restore from this latest snapshot, if restoration fails then take a snapshot again (retry till restoration succeeds) else it's fine(snapshot isn't corrupted).

This method will work but starting a embedded etcd and wait for restoration to complete can be time taken and costly process.

2. Is there a way by which we can just calculate the hash of db till `x`  revision and compare it with hash of snapshot (removing the appended hash). If it matched then our snapshot integrity is intact else re-try to take the snapshot again till hash matches.

But, I'm not sure about how to calculate the hash of db till `x` revision ? Is there any api call available for that ?

[...]

PR #19116, opened 2024-12-30
Fix the calculation of `remaining bytes` while sending snapshot to client.

ishan16696:
Currently, `remaining bytes` is only considering the size of database, not considering the `sha256 size` as it's also required to send. Please refer here for more info: https://github.com/etcd-io/etcd/issues/18340#issuecomment-2242358927

This PR fixes the calculation of `remaining bytes` while sending snapshot to client. Moreover, it also improves the log to give more information.

### DIFF
--- a/server/etcdserver/api/v3rpc/maintenance.go
+++ b/server/etcdserver/api/v3rpc/maintenance.go
@@ -147,12 +147,13 @@ func (ms *maintenanceServer) Snapshot(sr *pb.SnapshotRequest, srv pb.Maintenance
 
 	sent := int64(0)
 	total := snap.Size()
+	totalByteToSend := total + int64(h.Size())
 	size := humanize.Bytes(uint64(total))
 
 	start := time.Now()
 	ms.lg.Info("sending database snapshot to client",
-		zap.Int64("total-bytes", total),
-		zap.String("size", size),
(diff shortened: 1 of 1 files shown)


server/etcdserver/api/v3rpc/maintenance.go
```diff
@@ -178,7 +179,7 @@ func (ms *maintenanceServer) Snapshot(sr *pb.SnapshotRequest, srv pb.Maintenance
 		// No, the client will still receive non-nil response
 		// until server closes the stream with EOF
 		resp := &pb.SnapshotResponse{
-			RemainingBytes: uint64(total - sent),
+			RemainingBytes: uint64(totalByteToSend - sent),
```
serathius, line 182: Maybe add a test?

ishan16696: ok, Can you please tell me where should I add the unit test for this as I don't see a test file `maintenance_test.go` ?

ahrtr: We can add an e2e test in [ctl_v3_snapshot_test.go](https://github.com/etcd-io/etcd/blob/main/tests/e2e/ctl_v3_snapshot_test.go)
- `len(recevied bytes) + RemainingBytes` should be always equal to the `totalByteToSend`

PR #19140, opened 2025-01-07
Fix mixVersion test case: ensure a snapshot to be sent out

ahrtr:
The test case `TestMixVersionsSnapshotByMockingPartition` is flaky. Sometimes, when the partitioned member gets restarted, its consistentIndex might be 8 or 9. If it's 9, it's exactly the latestSnapshotIndex - `SnapshotCatchUpEntries`, accordingly the leader won't send a snapshot.


ahrtr:
@jmhbnz @ivanvc @fuweid @siyuanfoundation @serathius

serathius (approved):

PR #19152, opened 2025-01-09
chore: enable early-return and superfluous-else from revive

mmorel-35:
#### Description 

[early-return](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#early-return): In Go it is idiomatic to minimize nesting statements, a typical example is to avoid if-then-else constructions.

[superfluous-else](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#superfluous-else): To improve the readability of code, it is recommended to reduce the indentation as much as possible. This rule highlights redundant else-blocks that can be eliminated from the code.

### DIFF
--- a/etcdctl/ctlv3/ctl.go
+++ b/etcdctl/ctlv3/ctl.go
@@ -116,9 +116,8 @@ func MustStart() {
 	if err := Start(); err != nil {
 		if rootCmd.SilenceErrors {
 			cobrautl.ExitWithError(cobrautl.ExitError, err)
-		} else {
-			os.Exit(cobrautl.ExitError)
 		}
+		os.Exit(cobrautl.ExitError)
 	}
 }
 
--- a/contrib/lock/client/client.go
+++ b/contrib/lock/client/client.go
@@ -135,11 +135,10 @@ func main() {
 
 	err = write("key0", fmt.Sprintf("value from client %x", mode), int64(version))
 	if err != nil {
-		if mode == 1 {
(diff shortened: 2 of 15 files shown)


ahrtr (approved):

etcdctl/ctlv3/ctl.go
```diff
@@ -114,11 +114,10 @@ func Start() error {
 
 func MustStart() {
 	if err := Start(); err != nil {
-		if rootCmd.SilenceErrors {
-			cobrautl.ExitWithError(cobrautl.ExitError, err)
-		} else {
+		if !rootCmd.SilenceErrors {
 			os.Exit(cobrautl.ExitError)
 		}
+		cobrautl.ExitWithError(cobrautl.ExitError, err)
```
ivanvc, line 120: I think this can also be simplified as just removing the else condition, as [`cobrautl.ExitWithError()`](https://github.com/etcd-io/etcd/blob/507c0de87bd5034e3de4ab76ebf96b54dae0cd52/pkg/cobrautl/error.go#L37-L40) already does an `os.Exit()`.

```suggestion
		if rootCmd.SilenceErrors {
			cobrautl.ExitWithError(cobrautl.ExitError, err)
        }
		os.Exit(cobrautl.ExitError)
```

jmhbnz (approved):
LGTM - Thanks @mmorel-35

PR #19155, opened 2025-01-09
chore: enable unnecessary-stmt rule from revive

mmorel-35:
#### Description 

[unnecessary-stmt](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#unnecessary-stmt): This rule suggests to remove redundant statements like a break at the end of a case block, for improving the code's readability.


jmhbnz (approved):
LGTM - Thanks @mmorel-35

Issue #19147, opened 2025-01-08
Workaround the flaky downgrade robustness test due to WAL records missing in all members


ahrtr:
## Background
Currently robustness test leverages the WAL records to rebuild the etcdserver's real history to check correctness
- check whether it matches what the client side receives.

So it requires that at least one member has complete WAL records.

## Issue
Based on the discussion in https://github.com/etcd-io/etcd/pull/19095 and https://github.com/etcd-io/etcd/pull/19038, a member might fail to flush some WAL records to disk when being stopped (and restarted again later). Accordingly, it causes the error `failed to read WAL, cannot be repaired, err: wal: slice bounds out of range`, refer to https://github.com/etcd-io/etcd/pull/19038#issuecomment-2532919530; the issue was fixed in https://github.com/etcd-io/etcd/pull/19095.

Usually it isn't a problem, because usually only one member gets stopped or killed in robustness test, only one member missing WAL record isn't a problem.
- Note there is NO any issue from users perspective. A member missing some WAL records isn't a problem, because it can get a snapshot from the leader when it gets started again if it's lag far behind the leader.

But in downgrade test, we need to stop & restart all the members one by one, so it's possible that each member has some missing WAL records. So it might be impossible to read the complete WAL records; accordingly it causes the error `last succesful client write .... was not persisted, required to validate`, refer to https://github.com/etcd-io/etcd/pull/19095#issuecomment-2569911236
- Again, there is NO any issue from users perspective.

## Proposed solution
Based on my previous test, usually the issue (WAL records fail to be flushed to disk) only happens in high traffic scenario, so one workaround solution is that we only play very low traffic when doing downgrade case in robustness test.

Also currently robustness test reads the longest WAL records. But the longest one may not be the correct one. We should ensure at least majorities members have the same longest WAL records. Refer to https://github.com/etcd-io/etcd/pull/19095#issuecomment-2577383006

cc @siyuanfoundation @serathius 

## Related discussion 

This is for other contributors reference. The goal of robustness test is to verify correctness of etcd, so ideally it should NOT depend on the any data (including WAL files) generated by etcd; it should fully regard etcd as a black box.  

But it's hard and super cost for robustness to build all exponencial possibilities (when a client gets a failure response, the server side may fail or success;  when there are multiple failed client requests, then the possibilities increase exponencially). So a practical way is to use WAL records to build the real history from the server side.

### Why is this needed?

Workaround the issue of the downgrade robustness test

[...]

PR #19160, opened 2025-01-10
robustness: only run MemberDowngrade test for high SnapshotCatchUpEntries

siyuanfoundation:
This should prevent snapshot from being sent in the test.
Tested locally for 100 times, did not see any flakiness.

https://github.com/etcd-io/etcd/issues/19147 

### DIFF
--- a/tests/robustness/failpoint/cluster.go
+++ b/tests/robustness/failpoint/cluster.go
@@ -206,6 +206,11 @@ func (f memberDowngrade) Available(config e2e.EtcdProcessClusterConfig, member e
 	if !fileutil.Exist(e2e.BinPath.EtcdLastRelease) {
 		return false
 	}
+	// only run memberDowngrade test if no snapshot would be sent between members.
+	// see https://github.com/etcd-io/etcd/issues/19147 for context.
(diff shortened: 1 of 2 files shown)


serathius (approved):

serathius:
Awesome finding! Thanks @siyuanfoundation

ahrtr:
Nice, it's a better solution. Each member will have complete WAL records.

But theoretically, it's still possible that a follower may lag behind the leader more than the `DefaultSnapshotCatchUpEntries` in high traffic case, so the leader will still send a snapshot, although it's low possibility. 

> Tested locally for 100 times, did not see any flakiness.

The other reason why you do not see the error is 50% of downgrade cases out of the 100 tries were skipped due to the catchUpEntries too small.

serathius:
> The other reason why you do not see the error is 50% of downgrade cases out of the 100 tries were skipped due to the catchUpEntries too small.

My assumption it was tested without the skip

ahrtr:
> My assumption it was 100 runs without the skip

In this PR, it set two snapCatchUpEntries (`100, etcdserver.DefaultSnapshotCatchUpEntries`), while robustness randomly selects one of them, so it's 50% possibilities? Once 100 is selected, then the downgrade test is skipped. 

Anyway, it's already good enough. Up to your guys to decide whether & how to enhance it.

Issue #17911, opened 2024-04-29
add snapshot step in `etcdutl migrate`


siyuanfoundation:
### What would you like to be added?

Currently during downgrade, if there is a WAL entry with fields from the higher version, the offline migration tool `etcdutl migrate` would abort because of the [version check](https://github.com/etcd-io/etcd/blob/6a1e28a519fde2cd24948770d73fb22b58e4b9e9/server/storage/schema/schema.go#L75). This would block the offline downgrade migration even if the entries of higher version have been committed to the backend. 

On the contrary, in the online migration, after enabling downgrade, [`ClusterVersionSet`](https://github.com/etcd-io/etcd/blob/2b1914c262749441f3657be6f1aa1b9a03ab11d3/server/etcdserver/apply/apply.go#L412) would downgrade the cluster version, and force snapshot, so that old WAL entries would be compacted and not blocking the migration. 

We need to make the offline migration consistent with the online migration flow, by adding a snapshot step before calling [`schema.Migrate`](https://github.com/etcd-io/etcd/blob/2b1914c262749441f3657be6f1aa1b9a03ab11d3/etcdutl/etcdutl/migrate_command.go#L131)

### Why is this needed?

To make offline downgrade consistent with online downgrade. 

https://github.com/etcd-io/etcd/issues/13168 Milestone 3

siyuanfoundation:
cc @serathius @henrybear327

ah8ad3:
If you think this task is suitable for me i would like to give it a try.

siyuanfoundation:
@ah8ad3 yes, it would be great if you can give it a try. You can ping me on slack if you have any questions. Thank you!

ah8ad3:
Sorry do we have any implementation of getting snapshot without need of server? i couldn't find any, all i find needs `Endpoint` to get snapshot from server. 
If i'm right we need to create an offline `snapshotter`.
Am i right? @siyuanfoundation @serathius

siyuanfoundation:
`snapshot.SaveWithVersion` fetches snapshot from remote etcd server, and when the data is connected to an ectd server, it could add new entries, which is not desirable. So it would need to be an offline `snapshotter` or just something to clean up the old WAL entries. @ah8ad3 can you dive into the details of [`EtcdServer.snapshot`](https://github.com/etcd-io/etcd/blob/2eb01b17ee1105869e9a5fc906fc2e1edbbad244/server/etcdserver/server.go#L2110) to see if it is possible?

[...]

PR #19168, opened 2025-01-10
Create a v2 snapshot when running etcdutl migrate command

ahrtr:
Refer to https://github.com/etcd-io/etcd/issues/17911#issuecomment-2539224498

This PR will make the `etcdutl migrate` command fully functional.
- It creates a v2snapshot from the v3store.
  You will never see error below anymore when executing `etcdutl migrate` command,

  ```
  Error: cannot downgrade storage, WAL contains newer entries, as the target version (3.5.0) is lower than the version (3.6.0) detected from WAL logs
  ```

   After executing the migrate command for all members, you just need to directly replace the binary of each member, then all done for the offline downgrade. Of course, it's still recommended to follow/perform the online downgrade process, as it doesn't break the workload. cc @ivanvc @jmhbnz 

- It also adds a separate `etcdutl v2snapshot create` command
   It's  just a manual last to resort solution for any potential issue. Usually we don't need it.

I need to add e2e test. I may also break down it into smaller PRs.

### DIFF
--- a/etcdutl/etcdutl/common.go
+++ b/etcdutl/etcdutl/common.go
@@ -16,14 +16,19 @@ package etcdutl
 
 import (
 	"errors"
+	"fmt"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
 	"go.etcd.io/etcd/client/pkg/v3/logutil"
 	"go.etcd.io/etcd/pkg/v3/cobrautl"
+	"go.etcd.io/etcd/server/v3/etcdserver"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
+	"go.etcd.io/etcd/server/v3/storage/backend"
 	"go.etcd.io/etcd/server/v3/storage/datadir"
+	"go.etcd.io/etcd/server/v3/storage/schema"
 	"go.etcd.io/etcd/server/v3/storage/wal"
 	"go.etcd.io/etcd/server/v3/storage/wal/walpb"
 	"go.etcd.io/raft/v3/raftpb"
@@ -67,4 +72,92 @@ func getLatestV2Snapshot(lg *zap.Logger, dataDir string) (*raftpb.Snapshot, erro
 	}
 
 	return snapshot, nil
+}
+
+func createV2SnapshotFromV3Store(dataDir string, be backend.Backend) error {
+	var (
+		lg = GetLogger()
+
+		snapDir = datadir.ToSnapDir(dataDir)
+		walDir  = datadir.ToWALDir(dataDir)
+	)
+
+	ci, term := schema.ReadConsistentIndex(be.ReadTx())
+
+	cl := membership.NewCluster(lg)
+	cl.SetBackend(schema.NewMembershipBackend(lg, be))
+	cl.UnsafeLoad()
+
+	latestWALSnap, err := getLatestWALSnap(lg, dataDir)
+	if err != nil {
+		return err
+	}
+
+	// Each time before creating the v2 snapshot, etcdserve always flush
+	// the backend storage (bbolt db), so the consistent index should never
+	// less than the Index or term of the latest snapshot.
+	if ci < latestWALSnap.Index || term < latestWALSnap.Term {
+		// This should never happen
+		return fmt.Errorf("consistent_index [Index: %d, Term: %d] is less than the latest snapshot [Index: %d, Term: %d]", ci, term, latestWALSnap.Index, latestWALSnap.Term)
+	}
+
+	voters, learners := getVotersAndLearners(cl)
+	confState := raftpb.ConfState{
+		Voters:   voters,
+		Learners: learners,
+	}
+
+	// create the v2 snaspshot file
+	raftSnap := raftpb.Snapshot{
+		Data: etcdserver.GetMembershipInfoInV2Format(lg, cl),
+		Metadata: raftpb.SnapshotMetadata{
+			Index:     ci,
+			Term:      term,
+			ConfState: confState,
+		},
+	}
+	sn := snap.New(lg, snapDir)
+	if err = sn.SaveSnap(raftSnap); err != nil {
+		return err
+	}
+
+	// save WAL snapshot record
+	w, err := wal.Open(lg, walDir, latestWALSnap)
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+	// We must read all records to locate the tail of the last valid WAL file.
+	_, st, _, err := w.ReadAll()
+	if err != nil {
+		return err
+	}
+
+	if err := w.SaveSnapshot(walpb.Snapshot{Index: ci, Term: term, ConfState: &confState}); err != nil {
+		return err
+	}
+
+	if term >= st.Term && ci > st.Commit {
+		if err := w.Save(raftpb.HardState{Term: term, Commit: ci, Vote: st.Vote}, nil); err != nil {
+			return err
+		}
+	}
+	return w.Sync()
+}
+
+func getVotersAndLearners(cl *membership.RaftCluster) ([]uint64, []uint64) {
+	var (
+		voters   []uint64
+		learners []uint64
+	)
+	for _, m := range cl.Members() {
+		if m.IsLearner {
+			learners = append(learners, uint64(m.ID))
+			continue
+		}
+
+		voters = append(voters, uint64(m.ID))
+	}
+
+	return voters, learners
 }
--- a/etcdutl/etcdutl/common_test.go
+++ b/etcdutl/etcdutl/common_test.go
@@ -15,16 +15,26 @@
 package etcdutl
 
 import (
+	"path"
+	"path/filepath"
 	"testing"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/fileutil"
+	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/pkg/v3/pbutil"
+	"go.etcd.io/etcd/server/v3/etcdserver"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/snap"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/v2store"
+	"go.etcd.io/etcd/server/v3/etcdserver/cindex"
+	"go.etcd.io/etcd/server/v3/storage/backend"
 	"go.etcd.io/etcd/server/v3/storage/datadir"
+	"go.etcd.io/etcd/server/v3/storage/schema"
 	"go.etcd.io/etcd/server/v3/storage/wal"
 	"go.etcd.io/etcd/server/v3/storage/wal/walpb"
 	"go.etcd.io/raft/v3/raftpb"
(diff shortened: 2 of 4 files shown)


ahrtr:
This is a huge PR, let me breakdown it into small PRs to make the review easier.

serathius:
So many codecov warnings, is there any way to hide them?

ahrtr:
> So many codecov warnings, is there any way to hide them?

Added an unit test, PTAL.

cc @fuweid @serathius  @siyuanfoundation

ahrtr:
Note that previously I was thinking creating a v2 snapshot file is just an optional step when executing `etcdutl migarte`. 

But it would definitely be better to make it as a required step, otherwise we will always see the [cannot downgrade storage, WAL contains newer entries](https://github.com/etcd-io/etcd/blob/5ccbeec769685826424aa147aed452ea4ec52673/server/storage/schema/schema.go#L78) error message when executing `etcdutl migrate` on a newly created cluster, due to the clusterVer with a new version (e.g 3.6, when migrating from 3.6 to 3.5) .

See also https://github.com/etcd-io/etcd/pull/13405#discussion_r1890378185

cc @fuweid @siyuanfoundation @serathius PTAL

serathius:
What about invariant `snapshot.Metadata.Index < db.consistentIndex`?  https://github.com/etcd-io/etcd/blob/0dcd015147fa7dd758b2a71b02cb8618e53ae418/server/storage/backend.go#L98-L101

We cannot just snapshot without updating db. That's not easy thing to change.

For `cannot downgrade storage, WAL contains newer entries, as the target version (3.5.0) is lower than the version (3.6.0) detected from WAL logs` error we might consider not considering SetClusterVersion("3.6") as an entry that makes wal version equal to 3.6.

ahrtr:
> What about invariant `snapshot.Metadata.Index < db.consistentIndex`?
> 
> https://github.com/etcd-io/etcd/blob/0dcd015147fa7dd758b2a71b02cb8618e53ae418/server/storage/backend.go#L98-L101
> 
> We cannot just snapshot without updating db. That's not easy thing to change.

Not sure I got your point. The `etcdutl migrate` is just an offline tool, and we are going to creating a v2snapshot file based on the v3store(bbolt db). Recovering the DB from a snapshot in case etcd crashes isn't etcdutl's responsiblity.



> For `cannot downgrade storage, WAL contains newer entries, as the target version (3.5.0) is lower than the version (3.6.0) detected from WAL logs` error we might consider not considering SetClusterVersion("3.6") as an entry that makes wal version equal to 3.6.

I thought about it before, but the answer is NO. Reasons,
- 3.5 doesn't have `SetClusterVersion` entries in WAL file. If we see such entries, it's definitely generated by 3.6 or above version.
  Example for 3.6,

  ```
  $ ./etcd-dump-logs ../../default.etcd/
  Snapshot:
  empty
  Start dumping log entries from snapshot.
  WAL metadata:
  nodeID=8e9e05c52164694d clusterID=cdf818194e3a8c32 term=2 commitIndex=5 vote=8e9e05c52164694d
  WAL entries: 5
  lastIndex=5
  term	     index	type	data
     1	         1	conf	method=ConfChangeAddNode id=8e9e05c52164694d
     2	         2	norm	
     2	         3	norm	header:<ID:7587884260861681666 > cluster_member_attr_set:<member_ID:10276657743932975437 member_attributes:<name:"default" client_urls:"http://localhost:2379" > > 
     2	         4	norm	header:<ID:7587884260861681668 > cluster_version_set:<ver:"3.6.0" > 
     2	         5	norm	header:<ID:7587884260861681669 > put:<key:"k1" value:"v1" > 
  
  Entry types (Normal,ConfigChange) count is : 5
  ```

  Example for 3.5,

  ```
  $ ./etcd-dump-logs ../../default.etcd/
  Snapshot:
  empty
  Start dumping log entries from snapshot.
  WAL metadata:
  nodeID=8e9e05c52164694d clusterID=cdf818194e3a8c32 term=2 commitIndex=5 vote=8e9e05c52164694d
  WAL entries: 5
  lastIndex=5
  term	     index	type	data
     1	         1	conf	method=ConfChangeAddNode id=8e9e05c52164694d
     2	         2	norm	
     2	         3	norm	method=PUT path="/0/members/8e9e05c52164694d/attributes" val="{\"name\":\"default\",\"clientURLs\":[\"http://localhost:2379\"]}"
     2	         4	norm	method=PUT path="/0/version" val="3.5.0"
     2	         5	norm	header:<ID:7587884260899009541 > put:<key:"k1" value:"v1" > 
  
  Entry types (Normal,ConfigChange) count is : 5
  ```


- We can revisit this in 3.7, as it won't be needed anymore.

etcdutl/etcdutl/common_test.go
```diff
@@ -141,3 +151,184 @@ func TestGetLatestWalSnap(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			dataDir := t.TempDir()
+			lg := zap.NewNop()
+
+			require.NoError(t, fileutil.TouchDirAll(lg, datadir.ToMemberDir(dataDir)))
+			require.NoError(t, fileutil.TouchDirAll(lg, datadir.ToWALDir(dataDir)))
+			require.NoError(t, fileutil.TouchDirAll(lg, datadir.ToSnapDir(dataDir)))
+
+			t.Log("Populate the wal file")
+			w, err := wal.Create(lg, datadir.ToWALDir(dataDir), pbutil.MustMarshal(
```
fuweid, line 231: Is it possible to have comment to describe what the initial state looks like?

ahrtr: Updated.

etcdutl/etcdutl/common.go
```diff
@@ -68,3 +73,88 @@ func getLatestV2Snapshot(lg *zap.Logger, dataDir string) (*raftpb.Snapshot, erro
...
+	// We must read all records to locate the tail of the last valid WAL file.
+	_, st, _, err := w.ReadAll()
+	if err != nil {
+		return err
+	}
+
+	if err := w.SaveSnapshot(walpb.Snapshot{Index: ci, Term: term, ConfState: &confState}); err != nil {
+		return err
+	}
+	if err := w.Save(raftpb.HardState{Term: term, Commit: ci, Vote: st.Vote}, nil); err != nil {
```
serathius, line 139: Not sure about adding a HardState here, what's the purpose? We using CI read from db file, means the index here are for sure committed, the commit index might be might further than db which is only flushed once every 5 seconds. I'm worried that it might cause problems with by breaking monotonicity of commit index.

ahrtr: > Not sure about adding a HardState here, what's the purpose? 

To ensure the snapshot index is committed, otherwise the snapshot may be filtered by etcdserver on bootstrap. Of course, I agree that we need to ensure the commitIndex never decrease. Will update later.

> We using CI read from db file, means the index here are for sure committed

It is NOT guaranteed, because etcd applies the entries and wal sync hardstate async. It's easy to verify, 

Firstly, make change something like below,
```
$ git diff
diff --git a/server/storage/wal/wal.go b/server/storage/wal/wal.go
index f3d7bc5f4..9597a0ccd 100644
--- a/server/storage/wal/wal.go
+++ b/server/storage/wal/wal.go
@@ -961,6 +961,13 @@ func (w *WAL) Save(st raftpb.HardState, ents []raftpb.Entry) error {
                return nil
        }
 
+       if st.Commit > 5 {
+               w.lg.Info("########### Sleeping 10 seconds", zap.Uint64("Index", st.Commit))
+               time.Sleep(10 * time.Second)
+               w.lg.Info("########### Panicking after 10 seconds")
+               panic("non empty hard state")
+       }
+
        mustSync := raft.MustSync(st, w.state, len(ents))
 
        // TODO(xiangli): no more reference operator
```

Execute a couple of `etcdctl put` command after starting etcd, then etcdserve will panic. Then you will find that consistentIndex is greater than commitIndex.

```
$ ./etcd-dump-db iterate-bucket ../../default.etcd/member/snap/db meta --decode
key="term", value=2
key="storageVersion", value="3.6.0"
key="consistent_index", value=6
key="confState", value="{\"voters\":[10276657743932975437],\"auto_leave\":false}"

$ ./etcd-dump-logs ../../default.etcd/
Snapshot:
empty
Start dumping log entries from snapshot.
WAL metadata:
nodeID=8e9e05c52164694d clusterID=cdf818194e3a8c32 term=2 commitIndex=5 vote=8e9e05c52164694d
WAL entries: 6
lastIndex=6
term	     index	type	data
   1	         1	conf	method=ConfChangeAddNode id=8e9e05c52164694d
   2	         2	norm	
   2	         3	norm	header:<ID:7587884288152690690 > cluster_member_attr_set:<member_ID:10276657743932975437 member_attributes:<name:"default" client_urls:"http://localhost:2379" > > 
   2	         4	norm	header:<ID:7587884288152690692 > cluster_version_set:<ver:"3.6.0" > 
   2	         5	norm	header:<ID:7587884288152690694 > put:<key:"k1" value:"v1" > 
   2	         6	norm	header:<ID:7587884288152690695 > put:<key:"k2" value:"v2" > 

Entry types (Normal,ConfigChange) count is : 6
```

ahrtr: Updated.

serathius: Good observation, you are right, there is nothing beneficial from fsyncing HardState to WAL. Still I don't think there is anything beneficial from commiting the Snapshot.

ahrtr: Probably I did not explain it clearly. 

Firstly the consistent_index may be greater than the commit Index (already confirmed based on my above comment).

Secondly, when etcdserver bootstraps, it only loads v2 snapshot with Index <= commit Index (see below),

https://github.com/etcd-io/etcd/blob/363b166319cc20579bc0233bea01009a8ee8609b/server/storage/wal/wal.go#L645-L652

So we need to commit the snapshot index, of course only when it's greater than the existing commitIndex.

fuweid (approved):

ahrtr:
cc @serathius

ah8ad3 (commented):
Thanks for implementing this, it gave me a good insight.

serathius:
> 3.5 doesn't have SetClusterVersion entries in WAL file. If we see such entries, it's definitely generated by 3.6 or above version.

ClusterVersionSet was introduced in v3.5 and is annotated as such https://github.com/etcd-io/etcd/blob/32cfd45182aa0ce06f336eb62127b3302c937439/api/membershippb/membership.proto#L40-L44.

The reason why v3.5 etcd still uses old `method=PUT path="/0/version" val="3.5.0"` is backward compatibility. The versioned WAL entries were introduced just in v3.6, so before that etcd maintainers added new proto across multiple releases to ensure compatibility. `SetClusterVersion` was added in v3.5, but not used to ensure it doesn't break v3.4 during upgrade. `SetClusterVersion` started being used in v3.6, but is v3.5 compatible because v3.5 understands it.

etcdutl/etcdutl/common.go
serathius: Ok, I think I now understand the motivation about putting HardState, however I see another problem. If the reason to add generation of snapshot is to prevent etcd v3.5 from going back in WAL and reading `SetClusterVersion("3.6")`, then what about situations where `SetClusterVersion` has raft index after CI?

serathius:
cc @siyuanfoundation can you also take a look as you worked on downgrades.

serathius:
> We can revisit this in 3.7, as it won't be needed anymore.

I think if we change how `SetClusterVersion` is handled in WAL version detection than we would not need this PR at all. I think that would be a simpler solution.

ahrtr:
> I think if we change how `SetClusterVersion` is handled in WAL version detection than we would not need this PR at all.

It means that we need to partially rollback https://github.com/etcd-io/etcd/pull/13405, which is what I had been trying avoid doing in the last minute before we release 3.6.0.

Anyway, let me see what's the effort and what's the impact.

etcdutl/etcdutl/common.go
ahrtr: > then what about situations where `SetClusterVersion` has raft index after CI?

Theoretically it's possible, but in practice it's unlikely. Creating a v2 snapshot is just a best effort action.



> PS: Please don't resolve conversations on discussion that didn't get an answer that was accepted by both sides. It slows down review

The motivation is that resolving the conversation gives reviewers an impression it's ready for further round of review, instead of still getting stuck on the existing comment.

Also based on previous experience, usually there is no response for a comment for a long time, sometimes no any followup comments at all (it's not aim at you, just a generic comment on review process). So I tend to resolve it if I think everything has resolved.

Improvement:
- I will try to keep it for a little longer next time even if I think everything is resolved.
- Please also feel free to unresolved it if you or anyone still have comment in a conversation.

ahrtr:
Just raised https://github.com/etcd-io/etcd/pull/19263, PTAL

siyuanfoundation (approved):

ahrtr:
We don't need this PR anymore. We will go for https://github.com/etcd-io/etcd/pull/19263

thx all

PR #19170, opened 2025-01-11
client/pkg/logutil: simplify MergeOutputPaths function

callthingsoff:
This patch follows #19181.

### DIFF
--- a/client/pkg/logutil/zap.go
+++ b/client/pkg/logutil/zap.go
@@ -15,7 +15,7 @@
 package logutil
 
 import (
-	"sort"
+	"slices"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


serathius:
Please consider first sending a PR to add a test before refactor. Separating tests would give more trust into changes because we would now nothing changes. Now there might be a bug/change in test that might slip.

callthingsoff:
> Please consider first sending a PR to add a test before refactor. Separating tests would give more trust into changes because we would now nothing changes. Now there might be a bug/change in test that might slip.

This patch follows https://github.com/etcd-io/etcd/pull/19181, PTAL.

callthingsoff:
> > Please consider first sending a PR to add a test before refactor. Separating tests would give more trust into changes because we would now nothing changes. Now there might be a bug/change in test that might slip.
> 
> This patch follows #19181, PTAL.

I have rebased onto main, PTAL. Thanks.

client/pkg/logutil/zap.go
```diff
@@ -72,37 +72,21 @@ var DefaultZapLoggerConfig = zap.Config{
...
-		outputs[v] = struct{}{}
-	}
-	outputSlice := make([]string, 0)
-	if _, ok := outputs["/dev/null"]; ok {
-		// "/dev/null" to discard all
-		outputSlice = []string{"/dev/null"}
-	} else {
-		for k := range outputs {
-			outputSlice = append(outputSlice, k)
+	mergePaths := func(old []string) []string {
```
serathius, line 75: Any reason for anonymous function? Would a normal named function be more readable?

client/pkg/logutil/zap.go
```diff
@@ -72,37 +72,22 @@ var DefaultZapLoggerConfig = zap.Config{
...
+		return []string{"/dev/null"}
 	}
-	cfg.ErrorOutputPaths = errOutputSlice
-	sort.Strings(cfg.ErrorOutputPaths)
-
-	return cfg
+	// clone a new one; don't modify the original, in case it matters.
+	dup := slices.Clone(old)
+	slices.Sort(dup)
+	return slices.Compact(dup)
```
serathius, line 92: This changes the complexity, however I don't expect it number of outputs to be large.

ahrtr (approved):

PR #19174, opened 2025-01-11
Enhance test case TestEtctlutlMigrate to support multiple member cluster

ahrtr:
This PR is extracted from https://github.com/etcd-io/etcd/pull/19168

cc @fuweid @serathius @ivanvc @jmhbnz @siyuanfoundation

### DIFF
--- a/tests/e2e/utl_migrate_test.go
+++ b/tests/e2e/utl_migrate_test.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !cluster_proxy
+
 package e2e
 
 import (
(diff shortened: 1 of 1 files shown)


serathius:
What's the benefit of running command on 3 member cluster if we just run it 3 times and never start the cluster?

ahrtr:
> What's the benefit of running command on 3 member cluster if we just run it 3 times and never start the cluster?

It will be useful to the following PR. It's just to make reviewing the following PR easier.

I will try to start the members again after the migrating. For example, after we migrate from 3.6 to 3.5, we should be able to start the members with 3.5 binary.

serathius (approved):

PR #19176, opened 2025-01-11
tests: use high burst value in limiter.

jmao-dd:
Use the highest MaximalQPS of all traffic profiles as burst otherwise actual traffic may be accidentally limited. For example, if burst is set to 200 it is unlikely traffic can achieve higher than 200.


serathius:
Thanks for the improvement.

Issue #19179, opened 2025-01-13
Missing delete event on watch opened on same revision as compaction request


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

**Starting from 9 January we started getting failures on presubmit tests**.

Presubmit history goes up to December 31, with failures only starting on implying the issue is new.
![image](https://github.com/user-attachments/assets/e6f630f0-8dce-4179-a8e7-c4df4746a957)

Failues are due to resumable guarantee being broken
```
 logger.go:146: 2025-01-10T22:33:35.465Z	ERROR	Broke watch guarantee	{"guarantee": "resumable", "client": 4, "request": {"Key":"/registry/pods/","Revision":409,"WithPrefix":true,"WithProgressNotify":true,"WithPrevKV":true}, "got-event": {"Type":"delete-operation","Key":"/registry/pods/default/jCocA","Value":{"Value":"","Hash":0},"Revision":410,"IsCreate":false,"PrevValue":{"Value":{"Value":"143","Hash":0},"ModRevision":146}}, "want-event": {"Type":"delete-operation","Key":"/registry/pods/default/OL767","Value":{"Value":"","Hash":0},"Revision":409,"IsCreate":false}}
    validate.go:48: Failed validating watch history, err: broke Resumable - A broken watch can be resumed by establishing a new watch starting after the last revision received in a watch event before the break, so long as the revision is in the history window
```

From history visualizations I have seen it follows pattern:
* Delete operation on rev X
* Compect on Rev X
* Etcd crashes on Rev X
* Watch opened on Rev X

![image](https://github.com/user-attachments/assets/9e921eea-495c-4ccd-b6a6-d20e2fb01f6c)

### What did you expect to happen?

Resumable guarantee should not be broken. 

### How can we reproduce it (as minimally and precisely as possible)?

Didn't yet managed to reproduce it locally.

### Anything else we need to know?

[...]

PR #19188, opened 2025-01-14
mvcc: restore tombstone index if it's first revision

fuweid:
The tombstone could be the only one available revision in database. It happens when all historical revisions have been deleted in previous compactions. Since tombstone revision is still in database, we should restore it as valid key index. Otherwise, we lost that deletion event if we try to compact on that revision.

REF: #19179 

### DIFF
--- a/server/storage/mvcc/key_index.go
+++ b/server/storage/mvcc/key_index.go
@@ -116,6 +116,15 @@ func (ki *keyIndex) restore(lg *zap.Logger, created, modified Revision, ver int6
 	keysGauge.Inc()
 }
 
+// restoreTombstone is used to restore a tombstone revision, which is the only
+// revision so far for a key. We don't know the creating revision (i.e. already
+// compacted) of the key, so set it empty.
+func (ki *keyIndex) restoreTombstone(lg *zap.Logger, main, sub int64) {
+	ki.restore(lg, Revision{}, Revision{main, sub}, 1)
+	ki.generations = append(ki.generations, generation{})
+	keysGauge.Dec()
+}
+
 // tombstone puts a revision, pointing to a tombstone, to the keyIndex.
 // It also creates a new empty generation in the keyIndex.
 // It returns ErrRevisionNotFound when tombstone on an empty generation.
--- a/server/storage/mvcc/kvstore.go
+++ b/server/storage/mvcc/kvstore.go
@@ -473,8 +473,12 @@ func restoreIntoIndex(lg *zap.Logger, idx index) (chan<- revKeyValue, <-chan int
 					continue
 				}
 				ki.put(lg, rev.Main, rev.Sub)
-			} else if !isTombstone(rkv.key) {
-				ki.restore(lg, Revision{Main: rkv.kv.CreateRevision}, rev, rkv.kv.Version)
+			} else {
+				if isTombstone(rkv.key) {
+					ki.restoreTombstone(lg, rev.Main, rev.Sub)
+				} else {
+					ki.restore(lg, Revision{Main: rkv.kv.CreateRevision}, rev, rkv.kv.Version)
+				}
 				idx.Insert(ki)
 				kiCache[rkv.kstr] = ki
 			}
--- a/tests/e2e/watch_test.go
+++ b/tests/e2e/watch_test.go
@@ -488,3 +488,77 @@ func testStartWatcherFromCompactedRevision(t *testing.T, performCompactOnTombsto
 		}
 	}
 }
+
+// TestResumeCompactionOnTombstone verifies whether a deletion event is preserved
+// when etcd restarts and resumes compaction on a key that only has a tombstone revision.
+func TestResumeCompactionOnTombstone(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	ctx := context.Background()
+	compactBatchLimit := 5
+
+	cfg := e2e.DefaultConfig()
+	clus, err := e2e.NewEtcdProcessCluster(context.Background(),
+		t,
+		e2e.WithConfig(cfg),
+		e2e.WithClusterSize(1),
+		e2e.WithCompactionBatchLimit(compactBatchLimit),
+		e2e.WithGoFailEnabled(true),
+		e2e.WithWatchProcessNotifyInterval(100*time.Millisecond),
+	)
+	require.NoError(t, err)
+	defer clus.Close()
+
+	c1 := newClient(t, clus.EndpointsGRPC(), cfg.Client)
+	defer c1.Close()
+
+	keyPrefix := "/key-"
+	for i := 0; i < compactBatchLimit; i++ {
+		key := fmt.Sprintf("%s%d", keyPrefix, i)
+		value := fmt.Sprintf("%d", i)
+
+		t.Logf("PUT key=%s, val=%s", key, value)
+		_, err = c1.KV.Put(ctx, key, value)
+		require.NoError(t, err)
+	}
+
+	firstKey := keyPrefix + "0"
+	t.Logf("DELETE key=%s", firstKey)
+	deleteResp, err := c1.KV.Delete(ctx, firstKey)
+	require.NoError(t, err)
+
+	var deleteEvent *clientv3.Event
+	select {
+	case watchResp := <-c1.Watch(ctx, firstKey, clientv3.WithRev(deleteResp.Header.Revision)):
+		require.Len(t, watchResp.Events, 1)
+
+		require.Equal(t, mvccpb.DELETE, watchResp.Events[0].Type)
+		deletedKey := string(watchResp.Events[0].Kv.Key)
+		require.Equal(t, firstKey, deletedKey)
+
+		deleteEvent = watchResp.Events[0]
+	case <-time.After(100 * time.Millisecond):
+		t.Fatal("timed out getting watch response")
+	}
+
+	require.NoError(t, clus.Procs[0].Failpoints().SetupHTTP(ctx, "compactBeforeSetFinishedCompact", `panic`))
+
+	t.Logf("COMPACT rev=%d", deleteResp.Header.Revision)
+	_, err = c1.KV.Compact(ctx, deleteResp.Header.Revision, clientv3.WithCompactPhysical())
+	require.Error(t, err)
+
+	require.NoError(t, clus.Restart(ctx))
+
+	c2 := newClient(t, clus.EndpointsGRPC(), cfg.Client)
+	defer c2.Close()
+
+	watchChan := c2.Watch(ctx, firstKey, clientv3.WithRev(deleteResp.Header.Revision))
+	select {
+	case watchResp := <-watchChan:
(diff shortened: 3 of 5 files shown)


server/storage/mvcc/kvstore.go
```diff
@@ -477,6 +477,11 @@ func restoreIntoIndex(lg *zap.Logger, idx index) (chan<- revKeyValue, <-chan int
 				ki.restore(lg, Revision{Main: rkv.kv.CreateRevision}, rev, rkv.kv.Version)
 				idx.Insert(ki)
 				kiCache[rkv.kstr] = ki
+			} else {
+				lg.Warn("historical revision(s) has been compacted", zap.Int64("main", rev.Main), zap.Int64("sub", rev.Sub))
```
serathius, line 481: Why warning? Is the fact that historical revisions have been compacted a problem that that user should know about?

server/storage/mvcc/key_index.go
```diff
@@ -116,6 +116,29 @@ func (ki *keyIndex) restore(lg *zap.Logger, created, modified Revision, ver int6
 	keysGauge.Inc()
 }
 
+// restoreTombstone is only used to build index when there is no previous revision.
+// Since there is no historical information, CreateRevision always is empty.
+func (ki *keyIndex) restoreTombstone(lg *zap.Logger, main, sub int64) {
```
serathius, line 121: Apart of not increasing `keyGauge` this function doesn't differs from `restore`. Maybe we could combine them as they represent the same process just with different parameters? 

Think it should be equivalent to `restore(lg, 0, Revision{main, sub}, 1)`

tests/e2e/watch_test.go
```diff
@@ -488,3 +488,65 @@ func testStartWatcherFromCompactedRevision(t *testing.T, performCompactOnTombsto
...
+		e2e.WithWatchProcessNotifyInterval(100*time.Millisecond),
+	)
+	require.NoError(t, err)
+	defer clus.Close()
+
+	c1 := newClient(t, clus.EndpointsGRPC(), cfg.Client)
+	defer c1.Close()
+
+	keyPrefix := "/key-"
+	for idx := 1; idx <= compactBatchLimit; idx++ {
```
serathius, line 514: Name `idx` implies index, however you are not indexing anything, just iterating. Maybe `i` is enough here?

serathius:
How much this differs from https://github.com/etcd-io/etcd/issues/18089 ?

server/storage/mvcc/kvstore.go
ahrtr: Right, let's remove this warning. It's expected behaviour. Also, you will see lots of such warning message when there are lots of deletion operations.

ahrtr:
> How much this differs from #18089 ?

As mentioned in https://github.com/etcd-io/etcd/issues/19179#issuecomment-2589616092, it's an edge case that we missed when fixing #18089.  etcdserver shouldn't silently drop any deletion event, it should either deliver the event to client or respond with an ErrCompact error.  

But since we missed this edge case (etcdserver crash right **after** finishing the compaction but **before** setting the `finishedCompactRev`), the etcdserver might silently drops deletion events.

server/storage/mvcc/key_index.go
fuweid: hi @serathius @ahrtr do you mean by this?

```go
func (ki *keyIndex) restoreTombstone(lg *zap.Logger, main, sub int64) {
     ki.restore(lg, Revision{}, Revsion{main, sub}, 1)
     ki.generations = append(ki.generations, generation{})
     keyGauge.Dec()
}
```

I was thinking about combining the following functions instead of introducing new function. What do you think about it?

```go
ki.restore(lg, Revision{}, Revision{main-1, sub}, 0) just introduces previous revision and then compact it
ki.tombstone(lg, main, sub)
ki.compact(lg, main, map[Revision]struct{}{})
```

tests/e2e/watch_test.go
fuweid: Yeah. Updated~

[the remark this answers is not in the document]
fuweid: Sure. Updated to use `0`

fuweid:
> How much this differs from https://github.com/etcd-io/etcd/issues/18089 ?

I think it's part of #18089 which I miss in the fix patch. 😂 
Just wondering that why robustness test just catches it after two months 🤔 The recent changes look irrelative to it.

server/storage/mvcc/key_index.go
ahrtr: > hi @serathius @ahrtr do you mean by this?
> 
> ```go
> func (ki *keyIndex) restoreTombstone(lg *zap.Logger, main, sub int64) {
>      ki.restore(lg, Revision{}, Revsion{main, sub}, 1)
>      ki.generations = append(ki.generations, generation{})
>      keyGauge.Dec()
> }
> ```

It's also an valid option.

serathius:
> Just wondering that why robustness test just catches it after two months 🤔 The recent changes look irrelative to it.

Don't know :(
Looks like it was causing flakes in presubmits so we never looked at them during robustness meeting because they are assumed to fail due to issues with PRs https://testgrid.k8s.io/sig-etcd-robustness#pull-etcd-robustness-arm64&include-filter-by-regex=Issue17780 . I randomly looked at one PR that had this failure. Filled https://github.com/etcd-io/etcd/issues/19189 to discuss and take lessons from this.

server/storage/mvcc/key_index.go
```diff
@@ -116,6 +116,14 @@ func (ki *keyIndex) restore(lg *zap.Logger, created, modified Revision, ver int6
 	keysGauge.Inc()
 }
 
+// restoreTombstone is only used to build index when there is no previous revision.
+// Since there is no historical information, CreateRevision always is empty.
```
ahrtr, line 120:
```suggestion
// restoreTombstone is used to restore a tombstone revision, which is the only
// revision so far for a key. We don't know the creating revision (i.e. already
// compacted) of the key, so set it empty.
```

server/storage/mvcc/kvstore.go
```diff
@@ -477,6 +477,10 @@ func restoreIntoIndex(lg *zap.Logger, idx index) (chan<- revKeyValue, <-chan int
 				ki.restore(lg, Revision{Main: rkv.kv.CreateRevision}, rev, rkv.kv.Version)
 				idx.Insert(ki)
 				kiCache[rkv.kstr] = ki
+			} else {
+				ki.restoreTombstone(lg, rev.Main, rev.Sub)
+				idx.Insert(ki)
+				kiCache[rkv.kstr] = ki
```
ahrtr, line 483: Minor comment, you can combine them 

```
			} else if !isTombstone(rkv.key) {
				ki.restore(lg, Revision{Main: rkv.kv.CreateRevision}, rev, rkv.kv.Version)
				idx.Insert(ki)
				kiCache[rkv.kstr] = ki
			} else {
				ki.restoreTombstone(lg, rev.Main, rev.Sub)
				idx.Insert(ki)
				kiCache[rkv.kstr] = ki
			}
```

into something like below,

```
			} else {
				if isTombstone(rkv.key) {
					ki.restoreTombstone(lg, rev.Main, rev.Sub)
				} else {
					ki.restore(lg, Revision{Main: rkv.kv.CreateRevision}, rev, rkv.kv.Version)
				}
				idx.Insert(ki)
				kiCache[rkv.kstr] = ki
			}
```

ahrtr:
Overall looks good, with only two minor comments. Thanks @fuweid

tests/e2e/watch_test.go
```diff
@@ -488,3 +488,67 @@ func testStartWatcherFromCompactedRevision(t *testing.T, performCompactOnTombsto
...
+		e2e.WithConfig(cfg),
+		e2e.WithClusterSize(1),
+		e2e.WithCompactionBatchLimit(compactBatchLimit),
+		e2e.WithGoFailEnabled(true),
+		e2e.WithWatchProcessNotifyInterval(100*time.Millisecond),
+	)
+	require.NoError(t, err)
+	defer clus.Close()
+
+	c1 := newClient(t, clus.EndpointsGRPC(), cfg.Client)
```
serathius, line 512: Why new client?

tests/e2e/watch_test.go
```diff
@@ -488,3 +488,67 @@ func testStartWatcherFromCompactedRevision(t *testing.T, performCompactOnTombsto
...
+
+	c2 := newClient(t, clus.EndpointsGRPC(), cfg.Client)
+	defer c2.Close()
+
+	watchChan := c2.Watch(ctx, firstKey, clientv3.WithRev(deleteResp.Header.Revision))
+	select {
+	case watchResp := <-watchChan:
+		require.Len(t, watchResp.Events, 1)
+
+		require.Equal(t, mvccpb.DELETE, watchResp.Events[0].Type)
```
serathius, line 546: I was a little surprised that we are able to restore tombstone without any information about key creation, but apparently fields like `CreateRevision`, `Version` are set to 0 on delete. Still I think it would be nice if it was clear from the test.

Instead of comparing subset of fields to some hardcoded values, can we compare whole event to matching event from before the compaction? 
```
	var deleteEvent *clientv3.Event
	select {
	case watchResp := <-c1.Watch(ctx, firstKey, clientv3.WithRev(deleteResp.Header.Revision)):
		require.Len(t, watchResp.Events, 1)
		deleteEvent = watchResp.Events[0]
	case <-time.After(100 * time.Millisecond):
		// we care only about the first response, but have an
		// escape hatch in case the watch response is delayed.
		t.Fatal("timed out getting watch response")
	}

	require.NoError(t, clus.Procs[0].Failpoints().SetupHTTP(ctx, "compactBeforeSetFinishedCompact", `panic`))

	t.Logf("COMPACT rev=%d", deleteResp.Header.Revision)
	_, err = c1.KV.Compact(ctx, deleteResp.Header.Revision, clientv3.WithCompactPhysical())
	require.Error(t, err)

	require.NoError(t, clus.Restart(ctx))

	c2 := newClient(t, clus.EndpointsGRPC(), cfg.Client)
	defer c2.Close()

	watchChan := c2.Watch(ctx, firstKey, clientv3.WithRev(deleteResp.Header.Revision))
	select {
	case watchResp := <-watchChan:
		require.Len(t, watchResp.Events, 1)
		require.Equal(t, deleteEvent, watchResp.Events[0])
	case <-time.After(100 * time.Millisecond):
		// we care only about the first response, but have an
		// escape hatch in case the watch response is delayed.
		t.Fatal("timed out getting watch response")
	}
```

fuweid: Good point. Updated.

tests/e2e/watch_test.go
fuweid: I want to use a brand-new client after restarting, just to ensure there are no connection issues with the new server.

serathius (approved):

ahrtr (approved):
LGTM

Thanks @fuweid 

Could you please backport this PR to 3.5 and 3.4? thx

PR #19191, opened 2025-01-14
Try to reproduce #19179

serathius:
#19179

### DIFF
--- a/tests/robustness/scenarios/scenarios.go
+++ b/tests/robustness/scenarios/scenarios.go
@@ -15,20 +15,11 @@
 package scenarios
 
 import (
-	"path/filepath"
 	"testing"
-	"time"
 
-	"github.com/stretchr/testify/require"
-
-	"go.etcd.io/etcd/api/v3/version"
-	"go.etcd.io/etcd/client/pkg/v3/fileutil"
-	"go.etcd.io/etcd/server/v3/etcdserver"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 	"go.etcd.io/etcd/tests/v3/robustness/client"
 	"go.etcd.io/etcd/tests/v3/robustness/failpoint"
-	"go.etcd.io/etcd/tests/v3/robustness/options"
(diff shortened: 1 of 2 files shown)


serathius:
cc @fuweid any recommendation on how to reproduce the https://github.com/etcd-io/etcd/issues/19179

serathius:
> There were only two kinds of requests before compaction panic. One is creation and other is deletion.

I think I can guess the reason, in Kubernetes traffic operations are based on local state that is feed from watch. It tries to balance number of objects within the storage to keep the average. If the watch was very delayed there would be a long time where the traffic would execute only creates as it would not be aware of any objects. When watch would caught up, the traffic would skip random operations and immediately go for deletions, as there were too many objects in storage.

Still no progress on reproduction.

fuweid:
> It tries to balance number of objects within the storage to keep the average. If the watch was very delayed there would be a long time where the traffic would execute only creates as it would not be aware of any objects. 

I checked the json data from operation.json file and range-request returned empty. It looks weird. 

```json
{
  "ClientId": 2,
  "Input": {
    "Type": "range",
    "LeaseGrant": null,
    "LeaseRevoke": null,
    "Range": {
      "Start": "/registry/pods/",
      "End": "",
      "Limit": 0,
      "Revision": 0
    },
    "Txn": null,
    "Defragment": null,
    "Compact": null
  },
  "Call": 7733479,
  "Output": {
    "Txn": null,
    "Range": {
      "KVs": [],
      "Count": 0
    },
    "LeaseGrant": null,
    "LeaseRevoke": null,
    "Defragment": null,
    "Compact": null,
    "ClientError": "",
    "Revision": 17,
    "Persisted": false,
    "PersistedRevision": 0,
    "Error": ""
  },
  "Return": 8828359
}
```

> Still no progress on reproduction.

yeah... +1

fuweid:
ping @serathius https://github.com/etcd-io/etcd/pull/19353 has been merged.

Issue #19141, opened 2025-01-07
Migrate non-boolean experimental flags in 3.6


siyuanfoundation:
### What would you like to be added?

For the 3.6 release, we would like to have the list of non-boolean `--experimental` flags migrated to flags without the `--experimental` prefix.

- [x] experimental-corrupt-check-time: https://github.com/etcd-io/etcd/pull/19156
- [x] experimental-compaction-batch-limit: https://github.com/etcd-io/etcd/pull/19196
- [x] experimental-watch-progress-notify-interval: https://github.com/etcd-io/etcd/pull/19248
- [x] experimental-warning-apply-duration: https://github.com/etcd-io/etcd/pull/19264
- [x] experimental-bootstrap-defrag-threshold-megabytes: https://github.com/etcd-io/etcd/pull/19267
- [x] ~~experimental-warning-unary-request-duration~~
- [x] experimental-max-learners: https://github.com/etcd-io/etcd/pull/19298
- [x] experimental-snapshot-catch-up-entries: https://github.com/etcd-io/etcd/pull/19352
- [x] experimental-compaction-sleep-interval: https://github.com/etcd-io/etcd/pull/19351
- [x] experimental-downgrade-check-time: https://github.com/etcd-io/etcd/pull/19328
- [x] experimental-set-member-localaddr https://github.com/etcd-io/etcd/pull/19413
- [x] experimental-peer-skip-client-san-verification: https://github.com/etcd-io/etcd/issues/19062

In 3.6, we would like to have both the flags with and without the `--experimental` prefix working. [`experimentalNonBoolFlagMigrationMap`](https://github.com/etcd-io/etcd/blob/aac7ef6bcc78c6aa3b558fdb2d330f3c782652da/server/embed/config.go#L135) should be used to achieve that. https://github.com/etcd-io/etcd/pull/19156 is an example.

### Why is this needed?

flag clean up for release 3.6

siyuanfoundation:
@ahrtr @serathius Please double check the flag list and see if any flag needs to remain `--experimental` or missing any.

siyuanfoundation:
/reopen 
because there are still other flags to migrate.

ahrtr:
More contributors are welcome to work on this to accelerate it.

ajaysundark:
I can take a look at some -

randomly self-assigning half of the set

```
experimental-snapshot-catch-up-entries
experimental-compaction-sleep-interval
experimental-downgrade-check-time
[...]

PR #19196, opened 2025-01-15
migrate flag experimental-compaction-batch-limit to use compaction-batch-limit

gangli113:
Fixes one flag described in: https://github.com/etcd-io/etcd/issues/19141

The experimental flags are marked as deprecated and will be decommissioned in v3.7.

### DIFF
--- a/server/etcdmain/config_test.go
+++ b/server/etcdmain/config_test.go
@@ -485,7 +485,6 @@ func TestCompactHashCheckTimeFlagMigration(t *testing.T) {
 		name                             string
 		compactHashCheckTime             string
 		experimentalCompactHashCheckTime string
-		useConfigFile                    bool
 		expectErr                        bool
 		expectedCompactHashCheckTime     time.Duration
 	}{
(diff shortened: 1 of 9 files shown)


siyuanfoundation (approved):

server/etcdmain/config_test.go
```diff
@@ -659,6 +633,92 @@ func TestCorruptCheckTimeFlagMigration(t *testing.T) {
...
+
+			if tc.experimentalCompactionBatchLimit != 0 {
+				cmdLineArgs = append(cmdLineArgs, fmt.Sprintf("--experimental-compaction-batch-limit=%d", tc.experimentalCompactionBatchLimit))
+				yc.ExperimentalCompactionBatchLimit = tc.experimentalCompactionBatchLimit
+			}
+
+			cfgFromCmdLine, errFromCmdLine, cfgFromFile, errFromFile := generateCfgsFromFileAndCmdLine(t, yc, cmdLineArgs)
+
+			if tc.expectErr {
+				if errFromCmdLine == nil || errFromFile == nil {
```
ahrtr, line 685:
```suggestion
				if errFromCmdLine == nil && errFromFile == nil {
```

gangli113: Hi @ahrtr, it should be better to keep `||`. Because we want both errFromCmdLine and errFromFile to have values. If one err is nil, something is wrong and test case should catch it.

ahrtr: OK

ahrtr (approved):

Issue #7942, opened 2017-05-17
tcpproxy: SRV discovery differently broken in 3.1.7 and master


maikzumstrull:
The issues are in `etcdmain/gateway.go:startGateway`, which was touched between latest release and master.

In latest release (3.1.7 at time of writing), insecure SRV discovery does not work at all. As is relatively easy to tell from the code, the discovered endpoints are simply discarded in the insecure code path.

The rewrite of this code in master has accidentally fixed this (the goal of the rewrite was something unrelated), but introduces a minor cosmetic issue: the discovered endpoints are now logged as hex dumps, e.g.
```
proxy/tcpproxy: ready to proxy client requests to [0xc42023a400 0xc42023a420 0xc42023a440]
```
which is not that useful.

Mentioning @heyitsanthony who last touched this code.

PR #19198, opened 2025-01-15
server/proxy/tcpproxy: reuse addr in TCPProxy.Run

callthingsoff:
"addr" is the result of net.JoinHostPort which supports IPv6 address.

Rather than ranging "tp.Endpoints" again, it's better to reuse "addr" in the first loop to collect "eps".

See #7942 and #7943 for context.


ahrtr (approved):

ivanvc (approved):
LGTM. Thanks, @callthingsoff.

Issue #19172, opened 2025-01-11
Race condition when closing the embedded etcd


ahrtr:
### Which Github Action / Prow Jobs are flaking?

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19168/pull-etcd-integration-4-cpu-amd64/1878113519820345344
```
=== FAIL: integration/embed TestEmbedEtcd (2.17s)
==================
WARNING: DATA RACE
Write at 0x00c000346760 by goroutine 132:
  runtime.racewrite()
      <autogenerated>:1 +0x1e
  go.etcd.io/etcd/server/v3/embed.(*Etcd).Close()
      /home/prow/go/src/github.com/etcd-io/etcd/server/embed/etcd.go:460 +0xddc
  go.etcd.io/etcd/tests/v3/integration/embed_test.TestEmbedEtcd()
      /home/prow/go/src/github.com/etcd-io/etcd/tests/integration/embed/embed_test.go:120 +0x14d1
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:[169](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19168/pull-etcd-integration-4-cpu-amd64/1878113519820345344#1:build-log.txt%3A169)0 +0x226
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1743 +0x44
Previous read at 0x00c000346760 by goroutine 282:
  runtime.raceread()
      <autogenerated>:1 +0x1e
  go.etcd.io/etcd/server/v3/embed.(*Etcd).errHandler()
      /home/prow/go/src/github.com/etcd-io/etcd/server/embed/etcd.go:875 +0x6a
  go.etcd.io/etcd/server/v3/embed.(*Etcd).servePeers.func3()
25 skipped lines...
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:[174](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19168/pull-etcd-integration-4-cpu-amd64/1878113519820345344#1:build-log.txt%3A174)3 +0x44
==================
    testing.go:1399: race detected during execution of test
```

### Which tests are flaking?

.

joshuazh-x:
[...]

PR #19206, opened 2025-01-16
Avoid racing in closing Etcd.errc channel

joshuazh-x:

Another fix for #19172. It may be an overkill as the odds for having writes to errc after stopc is closed is pretty low.


ahrtr:
It's too complicated.

PR #19218, opened 2025-01-17
 Fix passing compaction-batch-limit to etcd v3.4 and v3.5

serathius:
PR https://github.com/etcd-io/etcd/pull/19196 broke robustness test periodics.

/cc @ahrtr @siyuanfoundation @gangli113

### DIFF
--- a/tests/robustness/failpoint/trigger.go
+++ b/tests/robustness/failpoint/trigger.go
@@ -75,7 +75,7 @@ func (t triggerCompact) Trigger(ctx context.Context, _ *testing.T, member e2e.Et
 		}
 		rev = resp.Header.Revision
 
-		if !t.multiBatchCompaction || rev > int64(clus.Cfg.ServerConfig.CompactionBatchLimit) {
+		if !t.multiBatchCompaction || rev > int64(clus.Cfg.ServerConfig.ExperimentalCompactionBatchLimit) {
 			break
 		}
 		time.Sleep(50 * time.Millisecond)
@@ -99,7 +99,7 @@ func (t triggerCompact) Available(config e2e.EtcdProcessClusterConfig, _ e2e.Etc
 	// For multiBatchCompaction we need to guarantee that there are enough revisions between two compaction requests.
 	// With addition of compaction requests to traffic this might be hard if experimental-compaction-batch-limit is too high.
 	if t.multiBatchCompaction {
-		return config.ServerConfig.CompactionBatchLimit <= 10
+		return config.ServerConfig.ExperimentalCompactionBatchLimit <= 10
 	}
 	return true
 }
--- a/tests/robustness/options/server_config_options.go
+++ b/tests/robustness/options/server_config_options.go
@@ -26,9 +26,9 @@ func WithSnapshotCount(input ...uint64) e2e.EPClusterOption {
 	}
 }
 
-func WithCompactionBatchLimit(input ...int) e2e.EPClusterOption {
+func WithExperimentalCompactionBatchLimit(input ...int) e2e.EPClusterOption {
 	return func(c *e2e.EtcdProcessClusterConfig) {
-		c.ServerConfig.CompactionBatchLimit = input[internalRand.Intn(len(input))]
+		c.ServerConfig.ExperimentalCompactionBatchLimit = input[internalRand.Intn(len(input))]
 	}
 }
 
--- a/tests/robustness/scenarios/scenarios.go
+++ b/tests/robustness/scenarios/scenarios.go
@@ -97,7 +97,7 @@ func Exploratory(_ *testing.T) []TestScenario {
 		options.WithSubsetOptions(randomizableOptions...),
 		e2e.WithGoFailEnabled(true),
 		// Set low minimal compaction batch limit to allow for triggering multi batch compaction failpoints.
-		options.WithCompactionBatchLimit(10, 100, 1000),
+		options.WithExperimentalCompactionBatchLimit(10, 100, 1000),
 		e2e.WithWatchProcessNotifyInterval(100 * time.Millisecond),
 	}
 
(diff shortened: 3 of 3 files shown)


server/embed/config.go
```diff
@@ -941,6 +941,13 @@ func (cfg *configYAML) configFromFile(path string) error {
 		cfg.ClusterState = ClusterStateFlagNew
 	}
 
+	if cfg.ExperimentalCompactionBatchLimit != 0 {
+		if cfg.CompactionBatchLimit != 0 {
+			return fmt.Errorf(`cannot set "experimentental-compaction-batch-limit" and "compaction-batch-limit" at the same time, please use "compaction-batch-limit" only`)
+		}
```
ahrtr, line 947: It's already been covered below,

https://github.com/etcd-io/etcd/blob/9da01a8275c4b8e1b3838ef92cf74d00e41b2edf/server/embed/config.go#L1023-L1026

server/embed/config.go
```diff
@@ -517,6 +513,10 @@ type configJSON struct {
 	PeerSecurityJSON   securityConfig `json:"peer-transport-security"`
 
 	ServerFeatureGatesJSON string `json:"feature-gates"`
+	// ExperimentalCompactionBatchLimit Sets the maximum revisions deleted in each compaction batch.
+	// Deprecated in v3.6 and will be decommissioned in v3.7.
+	// TODO: Delete in v3.7
+	ExperimentalCompactionBatchLimit int `json:"experimental-compaction-batch-limit"`
```
ahrtr, line 519: You just moved them from line 391-395 to here?

serathius: I moved it from `Config` to `configJSON` to support JSON config. Let me update PR name

server/embed/config.go
serathius: This is for flags, not for JSON config.

tests/framework/e2e/cluster.go
```diff
@@ -640,6 +636,12 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 		if flag == "experimental-snapshot-catchup-entries" && !CouldSetSnapshotCatchupEntries(execPath) {
 			continue
 		}
+		if flag == "experimental-compaction-batch-limit" {
+			continue
+		}
```
ahrtr, line 641: Has this already been covered in the above case `if defaultValue := defaultValues[flag]; value == "" || value == defaultValue` at line 633?

tests/framework/e2e/cluster.go
```diff
@@ -640,6 +636,12 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
...
+		if flag == "compaction-batch-limit" {
+			flag = "experimental-compaction-batch-limit"
+		}
```
ahrtr, line 644:
```suggestion
		// To be compatible with 3.4 & 3.5, as they don't have flag `--compaction-batch-limit`.
		if flag == "compaction-batch-limit" {
			flag = "experimental-compaction-batch-limit"
		}
```

But it's still a little tricky. Starting from 3.7, we will remove the experimental flags. So 
- 3.4 & 3.5 only have the experimental flags
- 3.6 have both flags
- 3.7 will only have the non-experimental flags.

So we may need to check the binary version and decide to use which flag. It's better to get them wrapped in a separate function.

tests/framework/e2e/cluster.go
```diff
@@ -881,16 +883,13 @@ func (epc *EtcdProcessCluster) StartNewProcFromConfig(ctx context.Context, tb te
...
 // to the cluster.
-func (epc *EtcdProcessCluster) UpdateProcOptions(i int, tb testing.TB, opts ...EPClusterOption) error {
+func (epc *EtcdProcessCluster) UpdateProcOptions(i int, tb testing.TB, newflags ...string) error {
 	if epc.Procs[i].IsRunning() {
 		return fmt.Errorf("process %d is still running, please close it before updating its options", i)
 	}
 	cfg := *epc.Cfg
-	for _, opt := range opts {
-		opt(&cfg)
-	}
```
ahrtr, line 891: are you sure this won't cause any issue?

tests/framework/e2e/cluster.go
serathius: The upgrade/downgrade robustness test can set this config, but will not update the flags before restart. We would need to rewrite the e2e framework to reload flags from config on every restart.

tests/framework/e2e/cluster.go
serathius: Don't know, but the idea of `UpdateProcOptions` function is incorrect. We should not reload config like this.

tests/framework/e2e/cluster.go
ahrtr: > We would need to rewrite the e2e framework to reload flags from config on every restart.

It seems that we need to reconcile all the config items once we update the binary (i.e. from 3.6 to 3.5, or from 3.5 to 3.6)

siyuanfoundation: use the `experimentalNonBoolFlagMigrationMap` to map the flags in migration for now? It should work before we rewrite the framework functions.

serathius: With this PR I will try to address the main issue of robustness test failing.

ahrtr: The only concern is that `--compaction-batch-limit` won't be covered by the e2e test anymore, because you always replace it with the `--experimental-compaction-batch-limit`.

Are you going to resolve it in a separate followup PR?

ahrtr: It was and it is, but it won't if we merge this PR.

ahrtr: I will let this PR in if you are aware of it and are planing to resolve it in a followup PR.

serathius: Let me simplify the it further to not require a followup.

ahrtr (approved):

ahrtr:
@serathius Note that https://github.com/etcd-io/etcd/pull/19221 fixes the workflow failure.

PR #19139, opened 2025-01-07
Avoid deadlock in etcd.Close when stopping during bootstrapping

joshuazh-x:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Fix issue described in #19058. Also refer to #10600. 

**Root cause**
`etcd.Close()` will read remaining items in channel `sctx.serversC` which was supposed to be closed when exiting `serveCtx.serve`. Stopping during bootstrapping may skip closing `sctx.serversC` hence cause `etcd.Close()` stuck.

Skip closing `sctx.serversC`
https://github.com/etcd-io/etcd/blob/aac7ef6bcc78c6aa3b558fdb2d330f3c782652da/server/embed/serve.go#L105-L109

https://github.com/etcd-io/etcd/blob/aac7ef6bcc78c6aa3b558fdb2d330f3c782652da/server/embed/serve.go#L124-L125

`etcd.Close()` reads closed channel
https://github.com/etcd-io/etcd/blob/aac7ef6bcc78c6aa3b558fdb2d330f3c782652da/server/embed/serve.go#L124-L125

PR #19221, opened 2025-01-17
Fix race condition (also a regression of the PR 19139)

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/19172

Please review this PR commit by commit.

Two high level thoughts,
- There are multiple levels of goroutines. The grandparent ( [StartEtcd](https://github.com/etcd-io/etcd/blob/5d47d7fc2e6bed14c6d3cf71ed3e6a939dd76bd5/server/embed/etcd.go#L99) ) creates multiple child goroutines ( [client listeners](https://github.com/etcd-io/etcd/blob/5d47d7fc2e6bed14c6d3cf71ed3e6a939dd76bd5/server/embed/etcd.go#L776-L780), peer listeners and metrics listeners). The [client listeners](https://github.com/etcd-io/etcd/blob/5d47d7fc2e6bed14c6d3cf71ed3e6a939dd76bd5/server/embed/etcd.go#L776-L780) creates some grandson goroutines (see the first commit).  Each one should only manage their immediate children.
- For `sync.WaitGroup`, we should always call `wg.Add` and `wg.Wait` in the same goroutine. 

cc @serathius @fuweid @ivanvc @jmhbnz  @joshuazh-x

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -79,12 +79,23 @@ type Etcd struct {
 
 	Server *etcdserver.EtcdServer
 
-	cfg   Config
-	stopc chan struct{}
-	errc  chan error
+	cfg Config
 
+	// closeOnce is to ensure `stopc` is closed only once, no matter
+	// how many times the Close() method is called.
 	closeOnce sync.Once
-	wg        sync.WaitGroup
+	// stopc is used to notify the sub goroutines not to send
+	// any errors to `errc`.
+	stopc chan struct{}
+	// errc is used to receive error from sub goroutines (including
+	// client handler, peer handler and metrics handler). It's closed
+	// after all these sub goroutines exit (checked via `wg`). Writers
+	// should avoid writing after `stopc` is closed by selecting on
+	// reading from `stopc`.
+	errc chan error
+
+	// wg is used to track the lifecycle of all sub goroutines created by `StartEtcd`.
+	wg sync.WaitGroup
 }
 
 type peerListener struct {
@@ -388,6 +399,24 @@ func (e *Etcd) Config() Config {
 // Close gracefully shuts down all servers/listeners.
 // Client requests will be terminated with request timeout.
 // After timeout, enforce remaning requests be closed immediately.
+//
+// The rough workflow to shut down etcd:
+//  1. close the `stopc` channel, so that all error handlers (child
+//     goroutines) won't send back any errors anymore;
+//  2. stop the http and grpc servers gracefully, within request timeout;
+//  3. close all client and metrics listeners, so that etcd server
+//     stops receiving any new connection;
+//  4. call the cancel function to close the gateway context, so that
+//     all gateway connections are closed.
+//  5. stop etcd server gracefully, and ensure the main raft loop
+//     goroutine is stopped;
+//  6. stop all peer listeners, so that it stops receiving peer connections
+//     and messages (wait up to 1-second);
+//  7. wait for all child goroutines (i.e. client handlers, peer handlers
+//     and metrics handlers) to exit;
+//  8. close the `errc` channel to release the resource. Note that it's only
+//     safe to close the `errc` after step 7 above is done, otherwise the
+//     child goroutines may send errors back to already closed `errc` channel.
 func (e *Etcd) Close() {
 	fields := []zap.Field{
 		zap.String("name", e.cfg.Name),
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -607,7 +607,9 @@ func (e *Etcd) servePeers() {
 
 	// start peer servers in a goroutine
 	for _, pl := range e.Peers {
+		e.wg.Add(1)
 		go func(l *peerListener) {
+			defer e.wg.Done()
 			u := l.Addr().String()
 			e.cfg.logger.Info(
 				"serving peer traffic",
@@ -774,7 +776,9 @@ func (e *Etcd) serveClients() {
 
 	// start client servers in each goroutine
 	for _, sctx := range e.sctxs {
+		e.wg.Add(1)
 		go func(s *serveCtx) {
+			defer e.wg.Done()
 			e.errHandler(s.serve(e.Server, &e.cfg.ClientTLSInfo, mux, e.errHandler, e.grpcGatewayDial(splitHTTP), splitHTTP, gopts...))
 		}(sctx)
 	}
@@ -859,7 +863,9 @@ func (e *Etcd) serveMetrics() (err error) {
 				return err
 			}
 			e.metricsListeners = append(e.metricsListeners, ml)
+			e.wg.Add(1)
 			go func(u url.URL, ln net.Listener) {
+				defer e.wg.Done()
 				e.cfg.logger.Info(
 					"serving metrics",
 					zap.String("address", u.String()),
@@ -872,9 +878,6 @@ func (e *Etcd) serveMetrics() (err error) {
 }
 
 func (e *Etcd) errHandler(err error) {
-	e.wg.Add(1)
-	defer e.wg.Done()
-
 	if err != nil {
 		e.GetLogger().Error("setting up serving from embedded etcd failed.", zap.Error(err))
 	}
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -61,14 +61,22 @@ type serveCtx struct {
 	insecure bool
 	httpOnly bool
 
+	// ctx is used to control the grpc gateway. Terminate the grpc gateway
+	// by calling `cancel` when shutting down the etcd.
 	ctx    context.Context
 	cancel context.CancelFunc
 
 	userHandlers    map[string]http.Handler
 	serviceRegister func(*grpc.Server)
-	serversC        chan *servers
-	closeOnce       sync.Once
 
+	// serversC is used to receive the http and grpc server objects (created
+	// in `serve`), both of which will be closed when shutting down the etcd.
+	// Close it when `serve` returns or when etcd fails to bootstrap.
+	serversC chan *servers
+	// closeOnce is to ensure `serversC` is closed only once.
+	closeOnce sync.Once
+
+	// wg is used to track the lifecycle of all sub goroutines created by `serve`.
 	wg sync.WaitGroup
 }
 
--- a/server/embed/serve.go
+++ b/server/embed/serve.go
@@ -68,6 +68,8 @@ type serveCtx struct {
 	serviceRegister func(*grpc.Server)
 	serversC        chan *servers
 	closeOnce       sync.Once
+
+	wg sync.WaitGroup
 }
 
 type servers struct {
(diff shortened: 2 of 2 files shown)


ahrtr:
@fuweid @ivanvc @jmhbnz @serathius 

This PR fixed a regression caused by https://github.com/etcd-io/etcd/pull/19139. So let's get this merged and backport to 3.5 and probably 3.4. We need to get it included in 3.5.18

server/embed/etcd.go
```diff
@@ -411,6 +411,16 @@ func (e *Etcd) Close() {
 		close(e.stopc)
 	})
 
+	for i := range e.Clients {
+		if e.Clients[i] != nil {
```
fuweid, line 415: The Clients here is about net listener. If we close it at the beginning, all the connections will be closed, and it seems like we are unable to drain all the inflight requests gracefully. The `stopServers` function is used to gracefully shutdown.

ahrtr: > The Clients here is about net listener. If we close it at the beginning, all the connections will be closed,

It's intentional.  
- We should stop accepting any new connections immediately.
- It won't close the already established connections

ahrtr:
> @ahrtr could you add a comment describing the shutdown protocol you have in mind? It should make it easier to review and be useful for the future.

Done. Please see the last commit. cc @fuweid @ivanvc @jmhbnz @serathius

server/embed/etcd.go
```diff
@@ -388,6 +388,24 @@ func (e *Etcd) Config() Config {
 // Close gracefully shuts down all servers/listeners.
 // Client requests will be terminated with request timeout.
 // After timeout, enforce remaning requests be closed immediately.
+//
+// The rough workflow to shut down etcd:
```
serathius, line 392: Thanks for comments, however it's not what I had in mind. Sorry for not being clearer, having a comments for Close function is good, however I'm more worried about interactions of multiple concurrency primitives like `errc  chan error`, `serversC chan *servers`, `closeOnce sync.Once`, `	wg sync.WaitGroup`, `stopc chan struct{}`. There is zero documentation for those variables. It's not even clear for me what they are responsible, how they are uses by sub what are different states?

Some things that I noticed:
* All sub-goroutines are expected to add to `wg sync.WaitGroup` and write errors to `errc  chan error`.
* Before closing the `errc  chan error` channel, we need to wait for `wg sync.WaitGroup` to ensure all goroutines that write to `errc` has exited (writing to closed channel causes panic).
* To avoid deadlock on subroutines writing errors to `errc` they will give up as writing error when `stopc chan struct{}` is closed. 

If we tracked all of those dependencies, maybe we can untangle the web.

Maybe we can add this information to those variables.
```
type Etcd struct {
    ...
	stopc chan struct{}
	
    // errc includes error from subroutines. 
    // Closed after all subroutines exit (checked via `wg`).
    // Writers should avoid writing after `stopc` is closed by select on reading from `stopc`.
	errc  chan error

	closeOnce sync.Once
    // wg tracks lifecycle of all subroutines. 
	wg        sync.WaitGroup
}
```

ahrtr: Indeed, it's a little complicated. It took me some time to think about similar situation on multiple projects (including etcd). Essentially I think that it's caused by the intrinsic complexity of the software. But it's that much scary if we sort out the workflow/details.

> * To avoid deadlock on subroutines writing errors to `errc` they will give up as writing error when `stopc chan struct{}` is closed.

It will panic when writing to a closed channel. I was thinking not to fix it, as it's very low possibility to run into such situation, also panicing right before shutting down might not be a big problem. But it might give user an impression that etcd isn't stable or have issues on shutting down, so I decided to fix it.





> Maybe we can add this information to those variables.
> 
> ```
> type Etcd struct {
>     ...
> 	stopc chan struct{}
> 	
>     // errc includes error from subroutines. 
>     // Closed after all subroutines exit (checked via `wg`).
>     // Writers should avoid writing after `stopc` is closed by select on reading from `stopc`.
> 	errc  chan error
> 
> 	closeOnce sync.Once
>     // wg tracks lifecycle of all subroutines. 
> 	wg        sync.WaitGroup
> }
> ```

Updated & thx

server/embed/etcd.go
```diff
@@ -81,10 +81,16 @@ type Etcd struct {
 
 	cfg   Config
 	stopc chan struct{}
```
serathius, line 83: Can you add a similar explanation to stopc, closeOnce and sctxs[].serversC?

ahrtr: Done.

server/embed/etcd.go
```diff
@@ -388,6 +399,24 @@ func (e *Etcd) Config() Config {
...
 // Client requests will be terminated with request timeout.
 // After timeout, enforce remaning requests be closed immediately.
+//
+// The rough workflow to shut down etcd:
+//  1. close the `stopc` channel, so that all error handlers (child
+//     goroutines) won't send back any errors anymore;
+//  2. close all client and metrics listeners, so that etcd server
+//     stops receiving any new connection immediately;
+//  3. stop the http and grpc servers gracefully, within request timeout;
+//  4. call the cancel function to close the gateway context, so that
```
serathius, line 409: Should we close gateway before we close grpc servers? Correct me if I'm wrong, gateway exposes HTTP API and translates HTTP requests to grpc API by connecting to grpc server.

server/embed/etcd.go
```diff
@@ -388,6 +399,24 @@ func (e *Etcd) Config() Config {
 // Close gracefully shuts down all servers/listeners.
 // Client requests will be terminated with request timeout.
 // After timeout, enforce remaning requests be closed immediately.
+//
+// The rough workflow to shut down etcd:
+//  1. close the `stopc` channel, so that all error handlers (child
+//     goroutines) won't send back any errors anymore;
+//  2. close all client and metrics listeners, so that etcd server
```
serathius, line 406: Do we expect closing clients, listiners, servers is blocking? How we know we can immediately proceed?

ahrtr: Closing listeners is only to stop receiving new connections, it won't close the already established connection. We should gracefully shutdown the etcdserver.

server/embed/etcd.go
ahrtr: > I would prefer to revert the #19139 and not backport more changes

#19139 fixed the deadlock issue, and that part makes sense.  But the problem is that it introduced an regression due the `wg`.  We just need to fix it. 



> It's too risky break etcd just because of it.

We previously underestimated this issue, and there was also negligence when reviewing #19139. This is why I have spent so much time personally addressing the problem.

server/embed/etcd.go
ahrtr: Reverted this change, to reduce the review burden.

fuweid (approved):
LGTM

server/embed/etcd.go
ahrtr: The description has already been updated. Where did you get the description?

fuweid: In the pull request's description

ahrtr: Oh, updated. thx

ahrtr:
cc @serathius do you have any further comment?

There are three commits in this PR. Note the third commit only adds some comment. The first and second commits are straight forward, please see the [description](https://github.com/etcd-io/etcd/pull/19221#issue-2795718242) of this PR.

server/embed/etcd.go
```diff
@@ -79,12 +79,23 @@ type Etcd struct {
...
-	wg        sync.WaitGroup
+	// stopc is used to notify the sub goroutines not to send
+	// any errors to `errc`.
+	stopc chan struct{}
+	// errc is used to receive error from sub goroutines (including
+	// client handler, peer handler and metrics handler). It's closed
+	// after all these sub goroutines exit (checked via `wg`). Writers
+	// should avoid writing after `stopc` is closed by selecting on
+	// reading from `stopc`.
+	errc chan error
```
serathius, line 95: Using wg to drain errc is a new concept introduced in previous PR, looks like previously we dependent on us correctly predicting needed capacity https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/etcd.go#L257
Possibly issues stem from the fact that this logic got outdated, I think now we can have even 4 writers to errc per sctx.

If that's true then we should safe with setting capacity of errc channel to 0.

server/embed/etcd.go
```diff
@@ -774,7 +805,9 @@ func (e *Etcd) serveClients() {
 
 	// start client servers in each goroutine
 	for _, sctx := range e.sctxs {
+		e.wg.Add(1)
```
serathius, line 808: I would recommend wrapping the logic into a function to schedule a subroutine:

```
func (e *Etcd) runSubroutine(subroutine func() err) {
  e.wg.Add(1)
  go func() {
    defer e.wg.Done()
    e.errHandler(subroutine())
  }()
}
```

serathius: If you plan to backport the PR it should make it much safer to replace `errHandler` calls as it would prevent us from missing any caller.

server/embed/etcd.go
ahrtr: I admit that the calculation of the errc capacity might not be accurate, but in practice it's already good enough. Also it's a separate topic.

server/embed/etcd.go
ahrtr: On the contrary, I tend not to do the refactoring because we need to backport the PR to 3.5.

- From the top level, it's clear that the [StartEtcd](https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/etcd.go#L99C6-L99C15) only creates three kinds of subroutines, which need to send back error to `errc`. So I don't see we miss any call for now. If you think we might miss it in future when we (especially new contributors) make new changes, then let's discuss it separately and do whatever refactor on **main only**.
https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/etcd.go#L274-L278

-  Also note that as mentioned in the [description](https://github.com/etcd-io/etcd/pull/19221#issue-2795718242) of this PR, the [errHandler](https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/etcd.go#L874) is also used to track the grandson goroutine. Making the refactoring as you suggested may make it even harder to backport the PR.

ahrtr: Please see https://github.com/etcd-io/etcd/pull/19256

serathius (approved):

ahrtr:
@fuweid @serathius then let's merge this PR and backport it to 3.5. Afterwards, we can continue to do the refactoring in main only in https://github.com/etcd-io/etcd/pull/19257. We can have more discussion under that PR.

Please let me know your thoughts before I merge this PR. thx

serathius:
https://github.com/etcd-io/etcd/pull/19139 Was backported to v3.4 so we also need to backport there.

ahrtr:
> #19139 Was backported to v3.4 so we also need to backport there.

#19139 was only partially backported to 3.4, and already confirmed that 3.4 doesn't have this regression due to big code difference. So this PR only needs to be backported to 3.5.

fuweid:
> let's merge this PR and backport it to 3.5. Afterwards, we can continue to do the refactoring in main only in https://github.com/etcd-io/etcd/pull/19257. We can have more discussion under that PR.

+1. Agree

PR #19231, opened 2025-01-19
Fix race condition (also a regression of the PR 19139)

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/19172, which is a regression caused by https://github.com/etcd-io/etcd/pull/19139

I decided to partially rollback 19139 to fix https://github.com/etcd-io/etcd/issues/19172. We need to backport the fix to 3.5, so let's keep it as simple as possible.

We can do any further enhance in 3.6 only.

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -84,7 +84,6 @@ type Etcd struct {
 	errc  chan error
 
 	closeOnce sync.Once
-	wg        sync.WaitGroup
 }
 
 type peerListener struct {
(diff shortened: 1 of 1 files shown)


ahrtr:
We can't remove the [waitgroup](https://github.com/etcd-io/etcd/blob/88b39386b43823b9f662829d588a684a8073363b/server/embed/etcd.go#L87). So we have to follow https://github.com/etcd-io/etcd/pull/19221

PR #19233, opened 2025-01-20
Add compaction to kubernetes robustness tests

AwesomePatrol:
Previously proposed in https://github.com/etcd-io/etcd/pull/18773

### DIFF
--- a/tests/robustness/traffic/etcd.go
+++ b/tests/robustness/traffic/etcd.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"time"
 
 	"golang.org/x/time/rate"
 
@@ -108,7 +109,7 @@ func (t etcdTraffic) Name() string {
 	return "Etcd"
 }
 
-func (t etcdTraffic) Run(ctx context.Context, c *client.RecordingClient, limiter *rate.Limiter, ids identity.Provider, lm identity.LeaseIDStorage, nonUniqueWriteLimiter ConcurrencyLimiter, finish <-chan struct{}) {
+func (t etcdTraffic) RunTrafficLoop(ctx context.Context, c *client.RecordingClient, limiter *rate.Limiter, ids identity.Provider, lm identity.LeaseIDStorage, nonUniqueWriteLimiter ConcurrencyLimiter, finish <-chan struct{}) {
 	lastOperationSucceeded := true
 	var lastRev int64
 	var requestType etcdRequestType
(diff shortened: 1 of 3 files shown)


tests/robustness/traffic/kubernetes.go
```diff
@@ -322,13 +321,8 @@ func (s *storage) Count() int {
 func (s *storage) PickRandom() (key string, rev int64) {
 	s.mux.RLock()
 	defer s.mux.RUnlock()
-	n := rand.Intn(len(s.keyRevision))
-	i := 0
 	for k, v := range s.keyRevision {
```
serathius, line 322: Let's keep the `PickRandom` function random and fair. Picking first element from map, might seem random, but in reality it will return the same element everytime if map was not changed. I think it's important to have this function to have some equal distribution.

serathius:
Please consider splitting PRs.

tests/robustness/traffic/kubernetes.go
AwesomePatrol: Looks random to me: https://go.dev/play/p/HjoOidSuBOq

https://go.dev/ref/spec#For_range:
> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

You are right. It is not guaranteed to change. I will remove it when splitting the changes.

tests/robustness/traffic/etcd.go
```diff
@@ -155,6 +156,35 @@ func (t etcdTraffic) Run(ctx context.Context, c *client.RecordingClient, limiter
 	}
 }
 
+func (t etcdTraffic) RunCompactLoop(ctx context.Context, c *client.RecordingClient, period time.Duration, finish <-chan struct{}) {
+	var lastRev int64 = 2
+	timer := time.NewTimer(period)
```
serathius, line 161:
```suggestion
	timer := time.NewTimer(period)
	defer timer.Stop()
```

serathius: Or use time.After like in Kubernetes

AwesomePatrol: I think `time.Ticker` is more suitable here.

At first I also didn't want to change this code as it is just move from `tests/robustness/traffic/traffic.go` to here, but it is as good time as any to fix it.

tests/robustness/traffic/kubernetes.go
```diff
@@ -205,6 +207,62 @@ func (t kubernetesTraffic) generateKey() string {
...
+	for {
+		select {
+		case <-time.After(interval):
+		case <-ctx.Done():
+			return
+		case <-finish:
+			return
+		}
+
+		compactTime, rev, err = compact(ctx, c, compactTime, rev)
```
serathius, line 224: Noticed we don't set timeout here. Don't know if that an issue, but it might need a followup. As a separate process that doesn't have a failure condition, it would be good to put some external validation. For example validate that at least 1 compaction request was recorded.

serathius (approved):

PR #19195, opened 2025-01-14
Switch from grpc-ecosystem/go-grpc-prometheus to grpc-ecosystem/go-grpc-middleware/providers/prometheus

dims:
Reviving previous effort from: https://github.com/etcd-io/etcd/pull/17974

xref: https://github.com/kubernetes/kubernetes/issues/128583

Added a new test to make sure we are not missing any expected metrics.

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #19242, opened 2025-01-21
[WIP] Test for etcd metrics

dims:
The test being added here will fail until https://github.com/etcd-io/etcd/pull/19195 lands


serathius:
Can we first merge the test?

tests/integration/clientv3/metrics_test.go
```diff
@@ -168,11 +172,198 @@ func getHTTPBodyAsLines(t *testing.T, url string) []string {
...
+}
+
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
```
ahrtr, line 369: You should be able to use [slices.Contains](https://pkg.go.dev/slices#Contains) directly.

ahrtr:
> The test being added here will fail until #19195 lands


The test should be an independent change. We should finish & merge the test firstly. Afterwards, you need to rebase #19195.

I think you just need to start an etcd instance (and probably write a couple of key/values), then get all the metrics using /metrics endpoint. Please refer to https://github.com/ahrtr/etcd-diagnosis/blob/main/plugins/metrics/plugin.go

tests/integration/clientv3/metrics_test.go
dims: Done

dims:
@serathius @ahrtr updated this PR, this can go first. 3 metrics are not captured currently in this test (see notes in PR text). When i update the other PR will get added.

tests/integration/clientv3/metrics_test.go
```diff
@@ -168,11 +170,186 @@ func getHTTPBodyAsLines(t *testing.T, url string) []string {
...
+	go func() {
+		defer close(donec)
+
+		serr := srv.Serve(ln)
+		if serr != nil && !transport.IsClosedConnError(serr) {
+			t.Errorf("Err serving http requests: %v", serr)
+		}
+	}()
+
+	url := "unix://" + addr + "/metrics"
```
serathius, line 215: Thinking whether e2e test here would better, more accurate and would not require this custom code. But don't want to block on it.

tests/integration/clientv3/metrics_test.go
```diff
@@ -168,11 +170,186 @@ func getHTTPBodyAsLines(t *testing.T, url string) []string {
...
+	// Get the list of generated metrics
+	generatedMetrics := getMetricsList(t, url)
+	for _, metric := range expectedMetrics {
+		if !slices.Contains(generatedMetrics, metric) {
+			t.Errorf("Expected metric %s not found in generated metrics", metric)
+		}
+	}
+}
+
+func getMetricsList(t *testing.T, url string) []string {
```
serathius, line 338: FYI "github.com/prometheus/common/expfmt" package has a parser for metrics. Example usage https://github.com/kubernetes-sigs/metrics-server/blob/0dcee07ea614e8354ec740dfc3f528a572cacec4/test/e2e_test.go#L315-L329

ahrtr:
This PR follows the existing integration test cases.

I am not sure how other maintainers think, but personally I don't think the existing integration tests on metrics (see below) are appropriate, 
- https://github.com/etcd-io/etcd/blob/main/tests/integration/metrics_test.go
- https://github.com/etcd-io/etcd/blob/main/tests/integration/clientv3/metrics_test.go

For integration test, I think it would be better to simulate the embedded use case (start etcd by calling `StartEtcd`); but the current integration test creates the EtcdServer, provision all listeners, etc. directly, it does too much hack on etcd itself. It's hard to maintain, nor hard to ensure it's testing exactly the same implementation as production environment.  Even in the test, it also does too much hack. (i.e. register/listen prometheus metrics, etc.)

https://github.com/etcd-io/etcd/blob/2471e419c2c993f3b47f9f89b8dbb46644650e2e/server/embed/etcd.go#L110

Ideally, we should test etcd as a black box, so e2e test is much better.  If we test it in integrate test, we should start etcd via `StartEtcd`.

Of course, it's not this PR's problem. So I am OK to let this PR in and we can discuss it separately and probably add an e2e test separately.

tests/integration/clientv3/metrics_test.go
dims: @serathius I will update the implementation of this test as suggested in the other PR when i rebase!

tests/integration/clientv3/metrics_test.go
dims: 🙏🏾

dims:
> So I am OK to let this PR in and we can discuss it separately and probably add an e2e test separately.

thanks @ahrtr !

ahrtr (approved):

PR #19251, opened 2025-01-21
[3.4] mvcc: restore tombstone index if it's first revision

fuweid:
The tombstone could be the only one available revision in database. It happens when all historical revisions have been deleted in previous compactions. Since tombstone revision is still in database, we should restore it as valid key index. Otherwise, we lost that event.

(cherry picked from commit d8b41925749ec85b0320839968a2218c4f9dc73f) [#19188](https://github.com/etcd-io/etcd/pull/19188)

### DIFF
--- a/mvcc/key_index.go
+++ b/mvcc/key_index.go
@@ -127,6 +127,15 @@ func (ki *keyIndex) restore(lg *zap.Logger, created, modified revision, ver int6
 	keysGauge.Inc()
 }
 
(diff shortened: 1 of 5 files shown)


ahrtr (approved):
LGTM

Thanks @fuweid 

cc @serathius @ivanvc @jmhbnz

serathius (approved):

PR #19244, opened 2025-01-21
Add downgrade cancellation e2e tests

henrybear327:
Part 1 of the e2e test - downgrade cancellation is called before any of the node is actually downgraded

A README fix was also associated with this PR, as downgrade command doesn't seem to take version as a parameter. 

Reference: https://github.com/etcd-io/etcd/issues/17976

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Issue #17976, opened 2024-05-09
Add downgrade cancellation e2e tests


henrybear327:
### What would you like to be added?

Add e2e tests for downgrade cancellation.

### Why is this needed?

Improves test coverage for downgrade procedure which is critical for implementing downgrade support.

ahrtr:
@henrybear327 are you still working on this?

henrybear327:
> @henrybear327 are you still working on this?

Yes! I have a branch that I never fully finish, sorry. I will polish it and submit it for review ASAP.

ahrtr:
Discussed with @serathius, the intention is to ensure the cluster can be reverted to the original state in case any error happens during the downgrade.

henrybear327:
> [@henrybear327](https://github.com/henrybear327) any update on this?

Had a call with @siyuanfoundation today and discussed. PR will be up this weekend.

henrybear327:
Hey @ahrtr @siyuanfoundation, 

First part of the e2e test is delivered here https://github.com/etcd-io/etcd/pull/19244. 

The partial downgraded cluster cancellation PR will be delivered by the end of today, as I have a bug with it. Sorry that I probably should have broken it down when I was debugging so it won't look like there is no progress made so far...

henrybear327:
The partial downgraded cluster cancellation PR is delivered here https://github.com/etcd-io/etcd/pull/19252

henrybear327:
@siyuanfoundation @ahrtr I think we can close this issue after we either close or merge https://github.com/etcd-io/etcd/pull/19399 :)

Thank you!

PR #19252, opened 2025-01-21
Add downgrade cancellation e2e tests

henrybear327:
Part 2 of the e2e test - downgrade cancellation is called after the cluster has been partially downgraded (refer to part 1: https://github.com/etcd-io/etcd/pull/19244)

Reference: https://github.com/etcd-io/etcd/issues/17976

### DIFF
--- a/tests/e2e/cluster_downgrade_test.go
+++ b/tests/e2e/cluster_downgrade_test.go
@@ -17,12 +17,14 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"testing"
 	"time"
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
 
 	"go.etcd.io/etcd/api/v3/version"
 	"go.etcd.io/etcd/client/pkg/v3/fileutil"
@@ -43,41 +45,50 @@ const (
 	noCancellation CancellationState = iota
 	cancelRightBeforeEnable
 	cancelRightAfterEnable
+	cancelAfterDowngrading
 )
 
 func TestDowngradeUpgradeClusterOf1(t *testing.T) {
-	testDowngradeUpgrade(t, 1, false, noCancellation)
+	testDowngradeUpgrade(t, 1, 1, false, noCancellation)
 }
 
 func TestDowngradeUpgradeClusterOf3(t *testing.T) {
-	testDowngradeUpgrade(t, 3, false, noCancellation)
+	testDowngradeUpgrade(t, 3, 3, false, noCancellation)
 }
 
 func TestDowngradeUpgradeClusterOf1WithSnapshot(t *testing.T) {
-	testDowngradeUpgrade(t, 1, true, noCancellation)
+	testDowngradeUpgrade(t, 1, 1, true, noCancellation)
 }
 
 func TestDowngradeUpgradeClusterOf3WithSnapshot(t *testing.T) {
-	testDowngradeUpgrade(t, 3, true, noCancellation)
+	testDowngradeUpgrade(t, 3, 3, true, noCancellation)
 }
 
 func TestDowngradeCancellationWithoutEnablingClusterOf1(t *testing.T) {
-	testDowngradeUpgrade(t, 1, false, cancelRightBeforeEnable)
+	testDowngradeUpgrade(t, 0, 1, false, cancelRightBeforeEnable)
 }
 
 func TestDowngradeCancellationRightAfterEnablingClusterOf1(t *testing.T) {
-	testDowngradeUpgrade(t, 1, false, cancelRightAfterEnable)
+	testDowngradeUpgrade(t, 0, 1, false, cancelRightAfterEnable)
 }
 
 func TestDowngradeCancellationWithoutEnablingClusterOf3(t *testing.T) {
-	testDowngradeUpgrade(t, 3, false, cancelRightBeforeEnable)
+	testDowngradeUpgrade(t, 0, 3, false, cancelRightBeforeEnable)
 }
 
 func TestDowngradeCancellationRightAfterEnablingClusterOf3(t *testing.T) {
-	testDowngradeUpgrade(t, 3, false, cancelRightAfterEnable)
+	testDowngradeUpgrade(t, 0, 3, false, cancelRightAfterEnable)
 }
 
-func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool, triggerCancellation CancellationState) {
+func TestDowngradeCancellationAfterDowngrading1InClusterOf3(t *testing.T) {
+	testDowngradeUpgrade(t, 1, 3, false, cancelAfterDowngrading)
+}
+
+func TestDowngradeCancellationAfterDowngrading2InClusterOf3(t *testing.T) {
+	testDowngradeUpgrade(t, 2, 3, false, cancelAfterDowngrading)
+}
+
+func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterSize int, triggerSnapshot bool, triggerCancellation CancellationState) {
 	currentEtcdBinary := e2e.BinPath.Etcd
 	lastReleaseBinary := e2e.BinPath.EtcdLastRelease
 	if !fileutil.Exist(lastReleaseBinary) {
@@ -135,7 +146,7 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool, t
 		t.Logf("Cancelling downgrade before enabling")
 		e2e.DowngradeCancel(t, epc)
 		t.Log("Downgrade cancelled, validating if cluster is in the right state")
-		e2e.ValidateMemberVersions(t, epc, generateIdenticalVersions(clusterSize, currentVersionStr))
+		e2e.ValidateMemberVersions(t, epc, generateIdenticalVersions(clusterSize, currentVersion))
 
 		return // No need to perform downgrading, end the test here
 	}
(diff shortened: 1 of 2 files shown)


tests/e2e/cluster_downgrade_test.go
```diff
@@ -233,3 +291,35 @@ func getMembersAndKeys(t *testing.T, cc *e2e.EtcdctlV3) (*clientv3.MemberListRes
...
+}
+
+func generatePartialCancellationVersions(clusterSize int, membersToChange []int, ver *semver.Version) []*version.Versions {
+	ret := make([]*version.Versions, clusterSize)
+
+	for i := range clusterSize {
+		ret[i] = &version.Versions{
+			Cluster: ver.String(),
+			Server:  e2e.OffsetMinor(ver, 1).String(),
+			Storage: "",
```
henrybear327, line 316: Will continue to investigate this tomorrow

henrybear327: Thanks to @siyuanfoundation, it's intended behavior for node version 3.5. Refer to the following comment https://github.com/etcd-io/etcd/blob/291353bc09e84e901409b4087fc50c93def04ed7/server/etcdserver/server.go#L193-L194

henrybear327:
Rebased to main as https://github.com/etcd-io/etcd/pull/19244 is merged

ahrtr:
@siyuanfoundation PTAL, thx

tests/e2e/cluster_downgrade_test.go
```diff
@@ -43,41 +45,50 @@ const (
...
-func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool, triggerCancellation CancellationState) {
+func TestDowngradeCancellationAfterDowngrading1InClusterOf3(t *testing.T) {
+	testDowngradeUpgrade(t, 1, 3, false, cancelAfterDowngrading)
+}
+
+func TestDowngradeCancellationAfterDowngrading2InClusterOf3(t *testing.T) {
+	testDowngradeUpgrade(t, 2, 3, false, cancelAfterDowngrading)
+}
+
+func testDowngradeUpgrade(t *testing.T, cancellationSize int, clusterSize int, triggerSnapshot bool, triggerCancellation CancellationState) {
```
siyuanfoundation, line 91: change `cancellationSize` to `numberOfMembersToDowngrade`, and set it 0 for cases like cancelRightAfterEnable.

tests/framework/e2e/downgrade.go
```diff
@@ -65,6 +65,13 @@ func DowngradeCancel(t *testing.T, epc *EtcdProcessCluster) {
 }
 
 func DowngradeUpgradeMembers(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, numberOfMembersToChange int, currentVersion, targetVersion *semver.Version) error {
+	membersToChange := rand.Perm(len(clus.Procs))[:numberOfMembersToChange]
+	t.Logf(fmt.Sprintln("Elect members for operations"), zap.Any("members", membersToChange))
+
+	return DowngradeUpgradeMembersByID(t, lg, clus, membersToChange, currentVersion, targetVersion)
+}
+
+func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, membersToChange []int, currentVersion, targetVersion *semver.Version) error {
```
siyuanfoundation, line 74: don't have to create a new function for this. just return the ids of members changed
```go
func DowngradeUpgradeMembers(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, numberOfMembersToChange int, currentVersion, targetVersion *semver.Version) (membersChanged []int, err error)
```

tests/e2e/cluster_downgrade_test.go
henrybear327: Done as requested! Thanks!

tests/framework/e2e/downgrade.go
henrybear327: The main reason for adding this parameter is for this https://github.com/etcd-io/etcd/pull/19252/files#diff-b8efecadc80e6f11d87d4a3b3d59327048683b02a36b4c8e214e9ceec1222bb0R202

Since we are checking if the binary to be upgraded was indeed downgraded in the first place, I can't just pass in # of nodes in the cluster as it will fail. Thus, I modified it to take the set of nodes that are supposed to be upgraded.

Does this make sense to you? :)

tests/e2e/cluster_downgrade_test.go
```diff
@@ -176,7 +199,7 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool, t
 	beforeMembers, beforeKV = getMembersAndKeys(t, cc)
 
 	t.Logf("Starting upgrade process to %q", currentVersionStr)
-	err = e2e.DowngradeUpgradeMembers(t, nil, epc, len(epc.Procs), lastClusterVersion, currentVersion)
+	err = e2e.DowngradeUpgradeMembersByID(t, nil, epc, membersToChange, lastClusterVersion, currentVersion)
```
henrybear327, line 202: @siyuanfoundation The main reason for adding `membersToChange` parameter is because of this line here. 

Since we are checking if the node to be upgraded was indeed downgraded in the first place, I can't just pass in total number of nodes in the cluster as it will fail. Thus, I modified it to take the set of nodes that are supposed to be upgraded.

Does this make sense to you? :)

tests/framework/e2e/downgrade.go
henrybear327: Please see https://github.com/etcd-io/etcd/pull/19252/files#r1933609935

tests/e2e/cluster_downgrade_test.go
```diff
@@ -135,7 +162,7 @@ func testDowngradeUpgrade(t *testing.T, clusterSize int, triggerSnapshot bool, t
 		t.Logf("Cancelling downgrade before enabling")
 		e2e.DowngradeCancel(t, epc)
```
henrybear327, line 163: When I was running e2e tests locally, the `DowngradeCancel` call times out (`Error: etcdserver: request timed out`) quite often, esp. with more nodes downgraded in the cluster.

tests/e2e/cluster_downgrade_test.go
siyuanfoundation: I see. makes sense. Thanks!

henrybear327:
I will look into `pull-etcd-release-tests` when all other tests have passed. This is probably not related to this PR.

ahrtr:
> I will look into `pull-etcd-release-tests` when all other tests have passed. This is probably not related to this PR.

https://github.com/kubernetes/test-infra/pull/34236 should resolve the failure. It will be merged a couple of minutes later.

henrybear327:
Thanks @siyuanfoundation for putting up https://github.com/etcd-io/etcd/pull/19317. I have rebased this PR on top of this and let's see how much of an improvement on CI side we would have. 

I am still looking into other ways to reduce the flakiness in the meantime.

henrybear327:
The more the number of nodes in the cluster is downgraded, the easier it is to cause the downgrade cancellation to timeout. 

I don't think I have a better idea of reducing flakiness so far, so I added this [retry mechanism](https://github.com/etcd-io/etcd/pull/19252/commits/5f6fd5401c55c62c8325a354f3461c215399b91f). Might not be elegant, but at least the CI tests should be somewhat stable.

Tested with this command `PASSES="release e2e" PKG=./tests/e2e TESTCASE="TestDowngradeCancellationAfterDowngrading4InClusterOf5" TIMEOUT=40m ./scripts/test.sh -v -count=50 -failfast`, as this test usually times out on my machine every 5 runs or so.

tests/framework/e2e/downgrade.go
```diff
@@ -55,11 +55,38 @@ func DowngradeEnable(t *testing.T, epc *EtcdProcessCluster, ver *semver.Version)
...
 	c := epc.Etcdctl()
-	testutils.ExecuteWithTimeout(t, 20*time.Second, func() {
-		err := c.DowngradeCancel(context.TODO())
-		require.NoError(t, err)
+
+	// Due to the fact that cancellation request can timeout quite easily, we retry it for a bounded number of times before failing the test to reduce flakiness
+	var err error
+	testutils.ExecuteWithTimeout(t, 2*time.Minute, func() {
+		maxRetryCount := 5
+		for retryCount := 1; retryCount <= maxRetryCount; retryCount++ {
```
siyuanfoundation, line 63: How about just `for` for keep retrying until 2min is over?

tests/framework/e2e/downgrade.go
```diff
@@ -55,11 +55,38 @@ func DowngradeEnable(t *testing.T, epc *EtcdProcessCluster, ver *semver.Version)
...
+					t.Logf("etcdctl downgrade cancel request timed out, retrying again (attempt %v)", retryCount)
+					continue
+				}
+
+				if retryCount > 1 && strings.Contains(err.Error(), "no inflight downgrade job") {
+					// if the cancellation request times out, the cancellation can still be performed successfully
+					t.Logf("etcdctl downgrade cancel executed successfully though timed out in the previous call (attempt %v)", retryCount)
+					err = nil
+					break
+				}
```
siyuanfoundation, line 80:
```
if strings.Contains(err.Error(), "no inflight downgrade job") {
	// no need for cancel
	t.Log(err)
	err = nil
	break
}
t.Logf("etcdctl downgrade error: %w, retrying again (attempt %v)", err, retryCount)
```

tests/framework/e2e/downgrade.go
henrybear327: Original code as a reference

```go
if strings.Contains(err.Error(), "context deadline exceeded") {
	t.Logf("etcdctl downgrade cancel context deadline exceeded, retrying again (attempt %v)", retryCount)
	continue
} else if strings.Contains(err.Error(), "request timed out") {
	t.Logf("etcdctl downgrade cancel request timed out, retrying again (attempt %v)", retryCount)
	continue
}

if retryCount > 1 && strings.Contains(err.Error(), "no inflight downgrade job") {
	// if the cancellation request times out, the cancellation can still be performed successfully
	t.Logf("etcdctl downgrade cancel executed successfully though timed out in the previous call (attempt %v)", retryCount)
	err = nil
	break
}
```

siyuanfoundation (approved):

henrybear327:
> Thanks for working on this. Overall looks good to me.
> 
> 
> 
> After the downgrade is cancelled, users should be able to restore the cluster to its original state. They just need to replace the binary to previous version for each downgraded member. I think we should cover it as well. Of course, it can be addressed in followup PRs.

I can quickly work on this. 

Could you elaborate regarding the "replace the binary to the previous version for each downgraded member"? 

Thank you @ahrtr!

ahrtr:
> Could you elaborate regarding the "replace the binary to the previous version for each downgraded member"?

For example, for a 3 member cluster of 3.6. You downloaded 2 of them to 3.5, then cancelled the downgrade. The cluster version should be 3.5.

The recovery process is straightforward, just replace binary for the 2 downloaded member one by one,
- stop the member;
- replace binary (using 3.6 binary in this example)
- start the member

When it's done, the cluster version should be 3.6, and the cluster should be able to serve client requests correctly.

henrybear327:
> > Could you elaborate regarding the "replace the binary to the previous version for each downgraded member"?
> 
> For example, for a 3 member cluster of 3.6. You downloaded 2 of them to 3.5, then cancelled the downgrade. The cluster version should be 3.5.
> 
> The recovery process is straightforward, just replace binary for the 2 downloaded member one by one,
> 
> * stop the member;
> * replace binary (using 3.6 binary in this example)
> * start the member
> 
> When it's done, the cluster version should be 3.6, and the cluster should be able to serve client requests correctly.

Maybe I am mistaken. Isn't this what we are doing in `DowngradeUpgradeMembers` now? :)

ahrtr:
You can reuse `DowngradeUpgradeMembers`. The goal is to extend the e2e test case to verify that users are able to recover the partially downgraded cluster to its original state.

henrybear327:
> You can reuse `DowngradeUpgradeMembers`. The goal is to extend the e2e test case to verify that users are able to recover the partially downgraded cluster to its original state.

I will submit a follow-up PR and see if it fits your expectation! :)

PR #19254, opened 2025-01-22
Align NewConfig with AddFlags defaults

serathius:
cc @ahrtr @fuweid @AwesomePatrol @siyuanfoundation

### DIFF
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -71,13 +71,18 @@ const (
 	DefaultGRPCKeepAliveTimeout        = 20 * time.Second
 	DefaultDowngradeCheckTime          = 5 * time.Second
 	DefaultAutoCompactionMode          = "periodic"
+	DefaultAutoCompactionRetention     = "0"
 	DefaultAuthToken                   = "simple"
 	DefaultCompactHashCheckTime        = time.Minute
+	DefaultLoggingFormat               = "json"
 
-	DefaultDiscoveryDialTimeout      = 2 * time.Second
-	DefaultDiscoveryRequestTimeOut   = 5 * time.Second
-	DefaultDiscoveryKeepAliveTime    = 2 * time.Second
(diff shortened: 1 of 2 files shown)


server/embed/config.go
```diff
@@ -578,25 +578,31 @@ func NewConfig() *Config {
 		CORS:          map[string]struct{}{"*": {}},
 		HostWhitelist: map[string]struct{}{"*": {}},
 
-		AuthToken:    DefaultAuthToken,
-		BcryptCost:   uint(bcrypt.DefaultCost),
-		AuthTokenTTL: 300,
+		AuthToken:              DefaultAuthToken,
+		BcryptCost:             uint(bcrypt.DefaultCost),
+		AuthTokenTTL:           300,
+		SelfSignedCertValidity: 1,
```
ahrtr, line 584: We need to update the default value in `AddFlags`: change the 1 with `cfg.SelfSignedCertValidity`.  This comment applies other fields as well.

https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/config.go#L734

serathius: Done

ahrtr: I was saying to change

```
fs.UintVar(&cfg.SelfSignedCertValidity, "self-signed-cert-validity", 1, "The validity period of the client and peer certificates, unit is year") 
```

to 

```
fs.UintVar(&cfg.SelfSignedCertValidity, "self-signed-cert-validity", cfg.SelfSignedCertValidity, "The validity period of the client and peer certificates, unit is year") 
```

Also need to do similar changes for other fields you changed in this PR.

serathius: Oh, sorry I misread you. I agree with the direction however I worried about changing defaults by mistake that is easy to miss. So I proposed to make a minimal step that is easily verifiable by test.

ahrtr (approved):
LGTM with a minor comment, thx

PR #19255, opened 2025-01-22
Add Version field to the robustness model

AwesomePatrol:
It helps perform Txn before Compact in the same way as it is done in Kubernetes.

/cc @serathius


tests/robustness/model/history.go
```diff
@@ -183,12 +183,16 @@ func toEtcdCondition(cmp clientv3.Cmp) (cond EtcdCondition) {
 	switch {
 	case cmp.Result == etcdserverpb.Compare_EQUAL && cmp.Target == etcdserverpb.Compare_MOD:
 		cond.Key = string(cmp.KeyBytes())
+		cond.ExpectedRevision = cmp.TargetUnion.(*etcdserverpb.Compare_ModRevision).ModRevision
 	case cmp.Result == etcdserverpb.Compare_EQUAL && cmp.Target == etcdserverpb.Compare_CREATE:
 		cond.Key = string(cmp.KeyBytes())
+		cond.ExpectedRevision = cmp.TargetUnion.(*etcdserverpb.Compare_ModRevision).ModRevision
```
serathius, line 189: Don't know why we added parsing `Compare_CREATE` but we should not mix `CreateRevision` with `ModRevision`.

serathius: Could you leave a TODO and/or send a followup PR for that?

tests/robustness/report/wal.go
```diff
@@ -199,6 +199,13 @@ func parseEntryNormal(ent raftpb.Entry) (*model.EtcdRequest, error) {
 			OperationsOnFailure: []model.EtcdOperation{},
 		}
 		for _, cmp := range raftReq.Txn.Compare {
+			if ver := cmp.GetVersion(); ver > 0 {
```
serathius, line 202: Can you create a switch based on cmp.Target and cmp.Result?

serathius (approved):
LGTM, just couple of nits

tests/robustness/model/history.go
AwesomePatrol: Removed. It doesn't seem to be used anywhere and it will be easy to add when needed

tests/robustness/report/wal.go
AwesomePatrol: Done

tests/robustness/report/wal.go
```diff
@@ -199,10 +199,18 @@ func parseEntryNormal(ent raftpb.Entry) (*model.EtcdRequest, error) {
...
+				txn.Conditions = append(txn.Conditions, model.EtcdCondition{
+					Key:             string(cmp.Key),
+					ExpectedVersion: cmp.GetVersion(),
+				})
+			case cmp.Result == pb.Compare_EQUAL && cmp.Target == pb.Compare_MOD:
+				txn.Conditions = append(txn.Conditions, model.EtcdCondition{
+					Key:              string(cmp.Key),
+					ExpectedRevision: cmp.GetModRevision(),
+				})
+			}
```
serathius, line 213: We should panic if there is a unsupported case.

PR #19256, opened 2025-01-22
[Problematic new solution] Fix race condition (also a regression of the PR 19139) 

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/19172

Note this solution has a potential race condition, because the grandson routine (created in [serve](https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/serve.go#L96) ) call `wg.Add(1)` async, so it may happen after `wg.Wait()`.

Original solution: https://github.com/etcd-io/etcd/pull/19221

cc @serathius @fuweid

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -79,12 +79,24 @@ type Etcd struct {
 
 	Server *etcdserver.EtcdServer
 
-	cfg   Config
-	stopc chan struct{}
-	errc  chan error
+	cfg Config
 
+	// closeOnce is to ensure `stopc` is closed only once, no matter
+	// how many times the Close() method is called.
 	closeOnce sync.Once
-	wg        sync.WaitGroup
+	// stopc is used to notify the sub goroutines not to send
+	// any errors to `errc`.
+	stopc chan struct{}
+	// errc is used to receive error from sub goroutines (including
+	// client handler, peer handler and metrics handler). It's closed
+	// after all these sub goroutines exit (checked via `wg`). Writers
+	// should avoid writing after `stopc` is closed by selecting on
+	// reading from `stopc`.
+	errc chan error
+
+	// wg is used to track the lifecycle of all sub goroutines which
+	// need to send error back to the `errc`.
+	wg sync.WaitGroup
 }
 
 type peerListener struct {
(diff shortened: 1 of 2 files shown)


serathius:
> Note this solution has a potential race condition, because the grandson routine (created in [serve](https://github.com/etcd-io/etcd/blob/c9045d650e6c1cb2e5361723adf3bb8c6ea22ee0/server/embed/serve.go#L96) ) call wg.Add(1) async, so it may happen after wg.Wait().

But `serve` is also called in `errHandler`, so Wait will not exit until the `serve` exits. I think this pattern is composable.

ahrtr:
> But `serve` is also called in `errHandler`, so Wait will not exit until the `serve` exits.

It should be not correct. `sync.WaitGroup.Add(1)` is only safe iff  it's guaranteed to be called before `sync.WaitGroup.Wait()`.  Once it's async, then it isn't guaranteed any more. 

We have two correct solutions:
- https://github.com/etcd-io/etcd/pull/19221 
  It's straight forward.
- https://github.com/etcd-io/etcd/pull/19257
  There is a little refactoring, which might be a little better. If there is no strong objection, then let's go for this one. cc @fuweid @serathius

serathius:
Not sure I understand, are you saying that the following code is not concurrency safe?

https://go.dev/play/p/-y3KuHTE2WU

ahrtr:
> Not sure I understand, are you saying that the following code is not concurrency safe?
> 
> https://go.dev/play/p/-y3KuHTE2WU

You are right. It's safe both in your example and this PR.  Calling `WaitGroup.Add(1)` is safe as long as the counter is greater than 0 when it being called.

https://github.com/golang/go/blob/194de8fbfaf4c3ed54e1a3c1b14fc67a830b8d95/src/sync/waitgroup.go#L36-L39

But I still prefer https://github.com/etcd-io/etcd/pull/19257 to this one. 
- #19257 is clearer, while this one is somewhat like performing acrobatics and the readability is slightly poor.
- We also need to wait for the grpc gaateway goroutine to exit, https://github.com/etcd-io/etcd/blob/b74bc0792b000d64ba475eee43f0478aea2ebb33/server/embed/serve.go#L357-L368

PR #19263, opened 2025-01-23
Remove the code of checking ClusterVersionSet in MinimalEtcdVersion

ahrtr:
Link to https://github.com/etcd-io/etcd/pull/19168

### DIFF
--- a/server/storage/schema/schema_test.go
+++ b/server/storage/schema/schema_test.go
@@ -24,7 +24,6 @@ import (
 	"go.uber.org/zap"
 
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
-	"go.etcd.io/etcd/api/v3/membershippb"
 	"go.etcd.io/etcd/api/v3/version"
 	"go.etcd.io/etcd/server/v3/storage/backend"
 	betesting "go.etcd.io/etcd/server/v3/storage/backend/testing"
@@ -181,17 +180,17 @@ func TestMigrate(t *testing.T) {
 			},
 			expectVersion: nil,
 		},
+		/* TODO: add a dedicated request for testing only to cover such case
 		{
-			name:          "Downgrading v3.6 to v3.5 fails if there are newer WAL entries",
+			name:          "Downgrading v3.6 to v3.5 works even if there is ClusterVersionSetRequest WAL entries with 3.6 clusterVersion",
 			version:       version.V3_6,
 			targetVersion: version.V3_5,
 			walEntries: []etcdserverpb.InternalRaftRequest{
 				{ClusterVersionSet: &membershippb.ClusterVersionSetRequest{Ver: "3.6.0"}},
 			},
-			expectVersion:  &version.V3_6,
-			expectError:    true,
-			expectErrorMsg: "cannot downgrade storage, WAL contains newer entries",
-		},
+			expectVersion: nil, // 3.5 doesn't have field `storageVersion`, so it should be nil.
+			expectError:   false,
+		},*/
 		{
 			name:           "Downgrading v3.5 to v3.4 is not supported as schema was introduced in v3.6",
 			version:        version.V3_5,
--- a/server/storage/wal/version.go
+++ b/server/storage/wal/version.go
@@ -112,16 +112,6 @@ func visitEntryData(entryType raftpb.EntryType, data []byte, visitor Visitor) er
 			break
 		}
 		msg = proto.MessageReflect(&raftReq)
-		if raftReq.ClusterVersionSet != nil {
-			ver, err := semver.NewVersion(raftReq.ClusterVersionSet.Ver)
-			if err != nil {
-				return err
-			}
-			err = visitor(msg.Descriptor().FullName(), ver)
-			if err != nil {
-				return err
-			}
-		}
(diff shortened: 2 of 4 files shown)


ahrtr:
The test failures are expected. I will fix them later.

@serathius PTAL

ahrtr:
cc @fuweid @siyuanfoundation

siyuanfoundation (approved):

siyuanfoundation:
I think this makes sense because 3.5 already understands `ClusterVersionSet` and there is no reason to prevent the downgrade if no new data field is used except a higher cluster version.

siyuanfoundation:
This would probably not impact 3.6->3.5. But for 3.7->3.6 downgrade, if there is an entry of `ClusterVersionSet` to be applied after the downgrade, would that mean the cluster version be set to 3.7 when the old `ClusterVersionSet` entry is applied?

fuweid (approved):

ahrtr:
> This would probably not impact 3.6->3.5.

This PR resolves the offline migration (3.6 -> 3.5) issue as mentioned in https://github.com/etcd-io/etcd/pull/19168#issuecomment-2603053579.



> would that mean the cluster version be set to 3.7 when the old `ClusterVersionSet` entry is applied?

Nope. The change in this PR only means the minimum etcd version to understand `ClusterVersionSet` is 3.5.

cc @serathius PTAL in case you have any further comment.

server/storage/wal/version.go
```diff
@@ -113,11 +113,12 @@ func visitEntryData(entryType raftpb.EntryType, data []byte, visitor Visitor) er
 		}
 		msg = proto.MessageReflect(&raftReq)
 		if raftReq.ClusterVersionSet != nil {
```
serathius, line 115: I think you can remove the whole block for `ClusterVersionSet` and just move the call `return visitMessage(msg, visitor)` from line 145 here.

server/storage/wal/version_test.go
```diff
@@ -67,7 +67,7 @@ func TestEtcdVersionFromEntry(t *testing.T) {
 				Type:  raftpb.EntryNormal,
 				Data:  clusterVersionV3_6Data,
 			},
-			expect: &version.V3_6,
+			expect: &version.V3_5,
```
serathius, line 70: Thik this test case can be removed.

server/storage/schema/schema_test.go
```diff
@@ -182,15 +182,14 @@ func TestMigrate(t *testing.T) {
...
 			targetVersion: version.V3_5,
 			walEntries: []etcdserverpb.InternalRaftRequest{
 				{ClusterVersionSet: &membershippb.ClusterVersionSetRequest{Ver: "3.6.0"}},
 			},
-			expectVersion:  &version.V3_6,
-			expectError:    true,
-			expectErrorMsg: "cannot downgrade storage, WAL contains newer entries",
+			expectVersion: nil, // 3.5 doesn't have field `storageVersion`, so it should be nil.
+			expectError:   false,
 		},
```
serathius, line 193: I think we should have a test case that confirms that downgrade cannot happen if there is newer WAL entry. This might be tricky for v3.6 -> v3.5 as we don't have such entry. My original idea was to add a fake entry to wal proto definition that can be used for testing.

server/storage/wal/version.go
ahrtr: Updated. 

Note we can't move `return visitMessage(msg, visitor)` from line 145 here, otherwise we will have to add a meaningless "`return nil`" at the end of the function.

server/storage/wal/version_test.go
ahrtr: removed

server/storage/schema/schema_test.go
ahrtr: added a TODO, which will be resolved in a followup PR. thx

ahrtr:
@serathius PTAL, thx

tests/integration/utl_wal_version_test.go
```diff
@@ -85,7 +68,7 @@ func TestEtcdVersionFromWAL(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	assert.Equal(t, &semver.Version{Major: 3, Minor: 6}, walVersion.MinimalEtcdVersion())
+	assert.Equal(t, &semver.Version{Major: 3, Minor: 5}, walVersion.MinimalEtcdVersion())
```
serathius, line 71: I think it's ok to leave this test and have it return v3.4 as minimal version.

server/storage/schema/schema_test.go
serathius: SG

server/storage/wal/version.go
serathius: OK

tests/integration/utl_wal_version_test.go
ahrtr: I will update the proto (add a new message/field) for testing only in a followup PR, and then I will rewrite this test by then.

ahrtr: Updated.

serathius (approved):

PR #19274, opened 2025-01-25
[Solution 1] Add `DowngradeVersionTest` for Downgrade or migration test only

ahrtr:
This PR seems to be huge, but actually most of the code is generated by protoc. 

Please review this PR commit by commit, it should be super clear and straightforward.


ahrtr:
## Solution 1 (this PR)
- Good side:
We can verify the WAL version check in both unit test and e2e & integration test

- Bad side:
This PR changes the protoc between etcdclient and etcdserver, this might be a concern.  

## Solution 2: https://github.com/etcd-io/etcd/pull/19276
- Good side:
There is no any change between client and serve protoc

- Bad side:
We only verify the WAL version check in unit test

ahrtr:
cc @fuweid @serathius @siyuanfoundation

serathius:
Can you highlight the differences between solution 1 and 2? It's not clear from PRs themselves.

api/etcdserverpb/raft_internal.proto
```diff
@@ -71,6 +71,8 @@ message InternalRaftRequest {
   membershippb.ClusterVersionSetRequest cluster_version_set = 1300 [(versionpb.etcd_version_field) = "3.5"];
   membershippb.ClusterMemberAttrSetRequest cluster_member_attr_set = 1301 [(versionpb.etcd_version_field) = "3.5"];
   membershippb.DowngradeInfoSetRequest  downgrade_info_set = 1302 [(versionpb.etcd_version_field) = "3.5"];
+
+  DowngradeVersionTestRequest downgrade_version_test = 1400 [(versionpb.etcd_version_field) = "3.6"];
```
serathius, line 75: Recommending replacing `1400` with some very high number to ensure it's separated from normal fields.

ahrtr:
> Can you highlight the differences between solution 1 and 2? It's not clear from PRs themselves.

Solution 1 adds a rpc `DowngradeVersionTest` in `service Maintenance`, so it updates the protoc between client and server. Of course It also adds a message `DowngradeVersionTestRequest`, which is also added as a field into `InternalRaftRequest`.  

Solution 2 only adds a `DowngradeVersionTestRequest` message, which is also added a field into `InternalRaftRequest`. So we can only verify the `cannot downgrade storage, WAL contains newer entries` in unit test cases.

api/etcdserverpb/raft_internal.proto
ahrtr: Updated both this PR and https://github.com/etcd-io/etcd/pull/19276

ahrtr:
> Can you highlight the differences between solution 1 and 2? It's not clear from PRs themselves.

Updated my previous comment, https://github.com/etcd-io/etcd/pull/19274#issuecomment-2614293384

ahrtr:
> Sorry I missed the context here. Why is this test needed if the proto is never going to be in real WAL files? Can you add more context here @ahrtr ? Thanks!

I am going to close this PR. We prefer to https://github.com/etcd-io/etcd/pull/19276

When WAL files contain new entries with higher version, then it isn't safe to downgrade or migration. But actually there isn't any protoc change in 3.6.0, so there is no way to test it.  So we intentionally add one such message for test purpose only.

PR #19276, opened 2025-01-26
[Solution 2] Add `DowngradeVersionTestRequest` for Downgrade or migration test only 

ahrtr:
Link to https://github.com/etcd-io/etcd/pull/19274

Please review this PR commit by commit, it should be super clear and straightforward.


ahrtr:
cc @fuweid @serathius @siyuanfoundation This PR might be a little better than https://github.com/etcd-io/etcd/pull/19274

ahrtr:
The solution 1 https://github.com/etcd-io/etcd/pull/19274 changes the protoc between client and server, so I prefer to this PR (solution 2). See my summary https://github.com/etcd-io/etcd/pull/19274

cc @fuweid @siyuanfoundation @serathius

api/etcdserverpb/rpc.proto
```diff
@@ -1158,6 +1158,16 @@ message DowngradeResponse {
   string version = 2;
 }
 
+// DowngradeVersionTestRequest is used for test only. The version in
+// this request will be read as the WAL record version.If the downgrade
+// target version is less than this version, then the downgrade(online)
+// or migration(offline) isn't safe, so shouldn't be allowed.
+message DowngradeVersionTestRequest {
```
fuweid, line 1165: Is it like dry-run change?

ahrtr:
@fuweid @siyuanfoundation @serathius this is one of the blockers to cut release-3.6, PTAL

api/etcdserverpb/rpc.proto
fuweid: Thanks for the information.

If I understand correctly, solution 2 is better than solution 1 because it’s difficult to ensure there are no misuse calls.  If it's dry-run thing, like pre-validation call, solution 1 is better. I vote for this change.

Just thinking that `DowngradeVersionTestRequest` name is kind of confusing to me. It's not about downgrade, it's just to inject a WAL record to impact the mininum version result. What about `MigrationTestRequest`?

fuweid (approved):
LGTM

api/etcdserverpb/rpc.proto
ahrtr: Resolve this comment as we discussed on the call.

siyuanfoundation (approved):

ahrtr:
Thanks both @fuweid and @siyuanfoundation for the quick review!

PR #19272, opened 2025-01-25
tests: Add retry if promote member fails.

jmao-dd:
To handle the case of learner not being ready when being promoted, add retry with some delay to improve the stability of e2e tests.

This is for https://github.com/etcd-io/etcd/issues/19216

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #19123, opened 2025-01-03
Robustness: Improve the documentation

amentee:
Included the correct path of the makefile to build the failpoints targets

Signed-off-by: Sachin Kumar <amentee@github.com>

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #19279, opened 2025-01-26
Fix addMemberAsLearnerAndPromote to avoid error 'etcdserver: can only…

ahrtr:
… promote a learner member'

Fix a test regression caused by https://github.com/etcd-io/etcd/pull/19272

This is another example of regression caused by "ok-to-test" label wasn't added.

### DIFF
--- a/tests/e2e/runtime_reconfiguration_test.go
+++ b/tests/e2e/runtime_reconfiguration_test.go
@@ -178,7 +178,6 @@ func addMemberAsLearnerAndPromote(ctx context.Context, t *testing.T, epc *e2e.Et
 
 	id, err := epc.StartNewProc(ctx, nil, t, true /* addAsLearner */)
 	require.NoError(t, err)
-	_, err = epc.Etcdctl(e2e.WithEndpoints(endpoints)).MemberPromote(ctx, id)
 
 	attempt := 0
 	for attempt < 3 {


ahrtr:
> This is another example of regression caused by "ok-to-test" label wasn't added.

@ivanvc @jmhbnz Is it possible to prevent this next time? For example, add a workflow to check the "ok-to-test" label, if not set, then fails the workflow?

ahrtr:
See https://github.com/etcd-io/etcd/pull/19272#discussion_r1929839547

serathius (approved):

ivanvc:
> > This is another example of regression caused by "ok-to-test" label wasn't added.
> 
> @ivanvc @jmhbnz Is it possible to prevent this next time? For example, add a workflow to check the "ok-to-test" label, if not set, then fails the workflow?

I'm unsure if it will be possible to add a workflow, as workflows still need approval to run. I think a better approach would be prioritizing migrating all remaining workflows to prow jobs and eventually enabling tide.

ahrtr:
> I'm unsure if it will be possible to add a workflow, as workflows still need approval to run.

Something like below?
```
name: Check ok-to-test Label

on:
  pull_request:
    types:
      - opened
      - edited
      - synchronize
      - labeled
      - unlabeled

jobs:
  check-label:
    runs-on: ubuntu-latest
    steps:
      # Step 1: Checkout the code (required for most workflows)
      - name: Checkout code
        uses: actions/checkout@v3

      # Step 2: Fetch PR information
      - name: Get PR labels
        id: get-labels
        uses: actions/github-script@v6
        with:
          script: |
            const labels = context.payload.pull_request.labels.map(label => label.name);
            core.setOutput("labels", labels);

      # Step 3: Verify if the label 'ok-to-test' exists
      - name: Check for 'ok-to-test' label
        run: |
          echo "Labels found: ${{ steps.get-labels.outputs.labels }}"
          if [[ "${{ steps.get-labels.outputs.labels }}" != *"ok-to-test"* ]]; then
            echo "Error: 'ok-to-test' label is missing."
            exit 1
          fi

```

ivanvc:
@ahrtr, yes, but the issue is that we'll need someone to approve the workflows manually (which is what `gh-workflow-approve.yaml`/`/ok-to-test` is doing):
https://github.com/etcd-io/etcd/blob/2895b19f536ce145c3ac84a88ee4afbca8b03265/.github/workflows/gh-workflow-approve.yaml#L39-L44

For Prow, having the `ok-to-test` label is enough to kick off the jobs, but we need that workflow for GitHub.

So, if we add a workflow as you suggest, we will still need to approve the workflow somehow. Therefore, having a workflow like that sounds redundant if it can't run unless someone approves it.

ahrtr:
> So, if we add a workflow as you suggest, we will still need to approve the workflow somehow.

Probably I missed something. If we just run the workflow (Check ok-to-test Label) in github instead of prow, then it should work?

ivanvc:
Please look at this pull request from a first-time contributor: https://github.com/etcd-io/etcd/pull/19123. It doesn't run any GitHub workflows unless someone approves them. Therefore, a workflow to check if the `ok-to-test` label is present won't work.

ivanvc:
However, for regular contributors (still not team members), it will still trigger the workflows, i.e., https://github.com/etcd-io/etcd/pull/18711. I guess it would still be an improvement. Would you like to pursue this workflow?

ahrtr:
> i.e., #18711. I guess it would still be an improvement. Would you like to pursue this workflow?

YES, it's exactly the motivation of the suggestion.

PR #19281, opened 2025-01-27
Fix documentation of List method of Kubernetes interface

AwesomePatrol:
It wrongly asked the API user to use the last key of previous response when they should append "\x00" to exclude the last key from the contents of the next requested page.

### DIFF
--- a/client/v3/kubernetes/interface.go
+++ b/client/v3/kubernetes/interface.go
@@ -36,8 +36,9 @@ type Interface interface {
 	// If opts.Revision is non-zero, the key-value pairs are retrieved at the specified revision.
 	// If the required revision has been compacted, the request will fail with ErrCompacted.
 	// If opts.Limit is greater than zero, the number of returned key-value pairs is bounded by the limit.
-	// If opts.Continue is not empty, the listing will start from the key immediately after the one specified by Continue.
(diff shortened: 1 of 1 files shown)


fuweid (approved):

client/v3/kubernetes/interface.go
```diff
@@ -36,8 +36,9 @@ type Interface interface {
 	// If opts.Revision is non-zero, the key-value pairs are retrieved at the specified revision.
 	// If the required revision has been compacted, the request will fail with ErrCompacted.
 	// If opts.Limit is greater than zero, the number of returned key-value pairs is bounded by the limit.
-	// If opts.Continue is not empty, the listing will start from the key immediately after the one specified by Continue.
-	// The Continue value should be the last key returned in a previous paginated ListResponse.
+	// If opts.Continue is not empty, the listing will start from the key specified by it.
+	// The Continue value should be the last key returned in a previous
```
serathius, line 40: When Paginating the Continue should be set to last observed key with "\x00" appended to it.

serathius, line 40:
```suggestion
	// When Paginating the Continue should be set to last observed key with "\x00" appended to it.
```

serathius (approved):

PR #19297, opened 2025-01-28
Mark etcd-local-tester as deprecated so we can remove it in future

jmhbnz:
While looking at something else in our `tools/` directory the other day I stumbled into the `local-tester` directory and taking a brief look it seemed to me like it was something we no longer used as we have the massively more comprehensive robustness testing suite now.  The last meaningful update to the tool looked to be about eight years ago.

I propose we mark this old utility as deprecated and remove it in future so we don't confuse anyone with what tool they should be using to validate etcd behaviour under failure conditions.

cc @serathius, @ahrtr

### DIFF
--- a/tools/local-tester/README.md
+++ b/tools/local-tester/README.md
@@ -1,5 +1,8 @@
 # etcd local-tester
 
+> [!WARNING]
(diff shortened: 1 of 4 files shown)


serathius (approved):
Seems like `bridge` is s separate implementation of network failure injection that is already implemented by `proxy` package used by robustness tests.

ahrtr (approved):

PR #19305, opened 2025-01-29
scripts/test_images.sh: refactor to import job into the prow infrastructure

ivanvc:
While importing the release tests into the Prow infrastructure, after some exploration, I found that the current script didn't work correctly because it allows an input environment variable "IMAGE," which in the Prow infra points to the image from the container. This made the `scripts/test_images.sh` script fail. So, I took an opportunity to rewrite the script in a more modular way while fixing this issue.

* Wrap script in functions, add a main function, and call them from it.
* Make environment variables local inside the function scope to avoid unexpected behaviors by the running environment-supplied variables.
* Fix the typo in the "REPOSITORY" environment variable.
* Rename the input "IMAGE" to "TEST_IMAGE" to avoid clashing with the environment variable from the Prow infrastructure.

### DIFF
--- a/scripts/test_images.sh
+++ b/scripts/test_images.sh
@@ -7,88 +7,132 @@ IFS=$'\n\t'
 source ./scripts/test_lib.sh
 source ./scripts/build_lib.sh
 
-function startContainer {
+# Can't run darwin binaries in linux containers.
+if [[ $(go env GOOS) == "darwin" ]]; then
+  log_error "Error: Please use a Linux machine to test release images builds."
+  exit 1
+fi
+
+# Can't proceed without Docker.
+if ! command -v docker >/dev/null; then
+  log_error "Error: Cannot find docker. Please follow the installation instructions at: https://docs.docker.com/get-docker/"
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


ivanvc:
Let me know if you want me to start a fresh PR. I reused the one I used to test that the prow job worked.

scripts/test_images.sh
```diff
@@ -7,88 +7,132 @@ IFS=$'\n\t'
...
+  }
+
+# Run a version check for the given Docker image.
+function run_version_check {
+  local output
+  local found_version
+  local image=$1
+  shift
+  local expected_version=$1
+  shift
```
ahrtr, line 49:
```suggestion
  local expected_version=$2
```

ahrtr (approved):
LGTM with a minor comment.

scripts/test_images.sh
ivanvc: I need to shift because of line 51:

```bash
  output=$(docker run --rm "${image}" "${@}")
```

ahrtr: ack

Issue #19228, opened 2025-01-17
Flaky `TestSnapshotStatus `


ahrtr:
### Which Github Action / Prow Jobs are flaking?

```
=== FAIL: snapshot TestSnapshotStatus (0.12s)
    v3_snapshot_test.go:47: 
        	Error Trace:	/home/prow/go/src/github.com/etcd-io/etcd/etcdutl/snapshot/v3_snapshot_test.go:47
        	Error:      	Not equal: 
        	            	expected: 0x62132b4d
        	            	actual  : 0x6bf9[632](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19192/pull-etcd-unit-test-386/1880292144506212352#1:build-log.txt%3A632)4
        	Test:       	TestSnapshotStatus
```

Refer to https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19192/pull-etcd-unit-test-386/1880292144506212352

### Which tests are flaking?

.

PR #19313, opened 2025-01-31
deflaky: TestSnapshotStatus

fuweid:
## Root Cause

The backend commits pending changes into boltdb per 100 ms by default. When brand-new member starts, that member will save member into backend by `MustSaveMemberToBackend`. However, it doesn't force change into boltdb. It depends on backend's auto-commit per 100ms.

If auto-commit happens after `UpdateStorageVersion`, the `UpdateStorageVersion`
will fail on `missing confstate information`. The confstate information is handled by commit-prehook. Once `UpdateStorageVersion` fails, `TestSnapshotStatus` won't have another chance to do that because that
server has been closed. So without storage version information, hash of boltdb data is not expected.

## How to reproduce it?

```diff
diff --git a/etcdutl/snapshot/v3_snapshot_test.go b/etcdutl/snapshot/v3_snapshot_test.go
index c2b3d5202..cd88ff995 100644
--- a/etcdutl/snapshot/v3_snapshot_test.go
+++ b/etcdutl/snapshot/v3_snapshot_test.go
@@ -141,6 +141,7 @@ func createDB(t *testing.T, generateContent func(*etcdserver.EtcdServer)) string
        t.Helper()

        cfg := embed.NewConfig()
+       cfg.BackendBatchInterval = 5 * time.Second
        cfg.LogLevel = "fatal"
        cfg.Dir = t.TempDir()
```

```bash
➜  snapshot git:(deflaky-TestSnapshotStatus) ✗ go test -v -run "^TestSnapshotStatus$" -count 1 ./
=== RUN   TestSnapshotStatus
    v3_snapshot_test.go:47:
                Error Trace:    /home/weifu/workspace/etcd/etcdutl/snapshot/v3_snapshot_test.go:47
                Error:          Not equal:
                                expected: 0x62132b4d
                                actual  : 0x6bf96324
                Test:           TestSnapshotStatus
--- FAIL: TestSnapshotStatus (0.14s)
FAIL
FAIL    go.etcd.io/etcd/etcdutl/v3/snapshot     0.154s
FAIL
```

## How to fix?

Change `BackendBatchLimit` to `1` so that we can force-commit during `MustSaveMemberToBackend`.

Fixes: #19228

cc @ahrtr @serathius 

### DIFF
--- a/etcdutl/snapshot/v3_snapshot_test.go
+++ b/etcdutl/snapshot/v3_snapshot_test.go
@@ -141,6 +141,7 @@ func createDB(t *testing.T, generateContent func(*etcdserver.EtcdServer)) string
 	t.Helper()
 
 	cfg := embed.NewConfig()
+	cfg.BackendBatchLimit = 1
 	cfg.LogLevel = "fatal"
 	cfg.Dir = t.TempDir()
 


ahrtr:
> When brand-new member starts, that member will save member into backend by `MustSaveMemberToBackend`. However, it doesn't force change into boltdb. It depends on backend's auto-commit per 100ms.

The analysis basically makes sense. The reason is `UpdateStorageVersion` might fail due to Conf State not being persisted in time. What's the relationship with `MustSaveMemberToBackend`?

ahrtr:
> What's the relationship with `MustSaveMemberToBackend`?

OK, the confState only gets updated only when `applyConfChange`.

ahrtr (approved):
LGTM

Thanks for the nice catch!

PR #19314, opened 2025-02-01
Enable Go workspace

### LINKED ISSUE #18409: Introduce a Go workspace

After discussing this in last week's community meeting and based on feedback from the Go team (https://github.com/golang/go/issues/68254) due to the vulnerability (GHSA-5x4g-q5rc-36jp / golang/vulndb#2952), there are several benefits to introducing a Go workspace in the project, and one of the biggest motivations is to simplify the test scripts.

I have a branch with the `go.workspace` ([diff](https://github.com/etcd-io/etcd/compare/main...ivanvc:etcd:use-a-go-workspace?expand=1)). It requires changes in the build scripts and the test libraries. It still doesn't work, but there's some progress.

I mostly based it on how kubernetes/kubernetes defines the Go workspace. However, from golang/go#68254 and the motivation to have `govulncheck` spot vulnerabilities within our modules, their suggestion is to remove `replaces` pointing to local code in `go.mod`s, but k/k still has these replaces.

I wanted to open up the discussion to get feedback and/or implementation ideas.

### Progress track

- [x] Workspace activation: #19423 
- [ ] verify targets
  - [x] lint: #20788
  - [x] gofmt: #20814
  - [x] bom: #20829
  - [x] dep: #20848
  - [x] ~shellcheck~
  - [x] goword: #20815
    - [x] We can remove `go_srcs_in_module` after merging #20815 and #20812
  - [x] govet: #20810
  - [x] license-header: #20812 / #20818
  - [x] mod-tidy: #20828
  - [x] ~shellws~
  - [x] ~proto-annotations~
  - [x] ~genproto~
  - [x] ~yamllint~
  - [x] govet-shadow: #20810
  - [x] ~markdown-marker~
  - [ ] go-versions: Somewhat blocked by https://github.com/golang/go/issues/75941
  - [x] gomodguard: #20849
  - [ ] `scripts/verify_grpc_experimental.sh`: #21900 [^1]
- [x] fix targets
  - [x] bom: #20829
  - [x] lint: #20788
  - [x] yamllint: #20850 / #20857
  - [x] ~sync-toolchain-directive~
  - [x] scripts/fix.sh
    - [x] mod-tidy: #20851
    - [x] go fmt: #20858
    - [x] bashws: #20860
- [x] test targets
  - [x] unit: #20872
  - [x] integration: #20909
  - [x] e2e: #20925
  - [x] grpcproxy-integration: #20954
  - [x] grpcproxy-e2e: #20954
  - [x] ~e2e-release~: Not needed, it runs e2e and release tests. 
  - [x] ~release~: Not needed, it doesn't use any module functions.
  - [x] robustness: #20962
  - [x] coverage: #21145 
  - [x] build: #21824
  - [x] clean up old functions: #21826
- [x] build targets
  - [x] release: #21825
- [x] Miscellaneous scripts
  - [x] `scripts/update_dep.sh`: #21792
  - [x] run-govulncheck: #21823

### Why is this needed?

To improve the code quality and to spot vulnerabilities in our code firsthand.

[^1]: Needs 3.7 backport.

ivanvc:
I know that this may be like dropping a bomb, and I also know that the pull request has many commits and may not be easy to review.

However, after speaking with @thockin at KubeCon Salt Lake City and based on his experience doing this for the k/k repository, I don't think there's an easy way to break this pull request into smaller ones, as it would leave the repository in a broken state. Thanks for sending the first draft and simplifying how to use/update the forbidden dependencies.

However, I've been opening several pull requests targeting the fixes for findings I've found while editing the files, from removal of linters by duplications to addressing linting issues that were dodged in our previous approach.

I've been maintaining this parallel branch for at least a month now. It took me a while to write the commit messages correctly. So, it is easy to keep it up to the main branch; it just requires editing the first commit, where I introduce the `go.work` and `go.sum`.

This pull request introduces two new functionalities:

* Auto-generating the `go.work` file based on the submodules in the repository. This may sound like overkill with the small number of submodules we have, but it future-proofs the project if we need to add or delete any of them.
* Regenerate the `go.work.sum` based on the submodules `go.mod`/`go.sum`.

Both of these new additions are executed by `make fix`, so this will become a must when doing the dependency update. However, a new verify check will fail if the file `go.work.sum` is out of date.

Fixes #18409.


ivanvc:
There's a failure in the BOM check. I'll draft the pull request after it finishes running all of the checks. I want to see if anything else fails.

ivanvc:
Noting that the coverage output is also wrong, making the CodeCov check fail.

scripts/fix.sh
```diff
@@ -31,8 +35,8 @@ function bash_ws_fix {
 
 log_callout -e "\\nFixing etcd code for you...\n"
 
-run_for_modules mod_tidy_fix || exit 2
-run_for_modules run ${GO_CMD} fmt || exit 2
+run_for_all_modules mod_tidy_fix || exit 2
+run run gofmt -w . || exit 2
```
serathius, line 39: run run?

tests/robustness/makefile.mk
```diff
@@ -111,7 +111,7 @@ $(GOPATH)/bin/gofail: tools/mod/go.mod tools/mod/go.sum
 	  (cd server; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
 	  (cd etcdctl; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
 	  (cd etcdutl; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
-	  (cd tools/mod; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
+	  (go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
```
serathius, line 114: Please don't change how older etcd versions are built.

scripts/build_lib.sh
```diff
@@ -26,36 +26,33 @@ etcd_build() {
 
   run rm -f "${out}/etcd"
   (
-    cd ./server
```
serathius, line 29: Is this change required for workspace to work or is it a cleanup thanks to usage of workspace? I would recommend to defer changes that are not critical for workspace to work.

api/go.mod
```diff
@@ -31,9 +31,9 @@ require (
 // Bad imports are sometimes causing attempts to pull that code.
 // This makes the error more explicit.
 replace (
-	go.etcd.io/etcd => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/api/v3 => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/pkg/v3 => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/tests/v3 => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/v3 => ./FORBIDDEN_DEPENDENCY
+	go.etcd.io/etcd => ../FORBIDDEN_DEPENDENCY
+	go.etcd.io/etcd/api/v3 v3.0.0 => ../FORBIDDEN_DEPENDENCY
```
ahrtr, line 35: why v3.0.0?

go.work.sum
```diff
@@ -0,0 +1,986 @@
+bitbucket.org/creachadair/shell v0.0.7 h1:Z96pB6DkSb7F3Y3BBnJeOZH2gazyMTWlvecSD4vDqfk=
```
ahrtr, line 1: This file isn't up to date. If you remove this file and re-generate it (run `make build`), then you will see that lots of entries will be cleaned up.

api/go.mod
ahrtr: Also I don't see a need to change `./FORBIDDEN_DEPENDENCY` to `../FORBIDDEN_DEPENDENCY`, although not very important.

scripts/fix.sh
```diff
@@ -15,8 +15,12 @@ source ./scripts/updatebom.sh
...
 
 function mod_tidy_fix {
-  run rm ./go.sum
-  run ${GO_CMD} mod tidy || return 2
+  local modules=("$@")
+  for module in "${modules[@]}"; do
+    local module_dir
+    module_dir="${module%/...}"
+    run_for_module "${module_dir}" run "${GO_CMD}" mod tidy
+  done
```
ahrtr, line 23: `mod_tidy_fix` is supposed to run on a single module, and the loop is the responsibility of the caller.

Do not get time to read through the whole PR yet, is it possible not to change such logic to make the review easier? If this is a change unrelated to go workspace, then we should do it separately.

ahrtr:
It might be a little risky to merge this into main branch right before we release 3.6. We may want to only merge it into 3.7.

Also please only focus this PR only the necessary change related to go workspace. If it isn't necessary or any improvement, let's do it separately in follow PR.

scripts/build_lib.sh
ivanvc: It's a clean up. I'll revert it.

scripts/fix.sh
ivanvc: My bad.

api/go.mod
ivanvc: > why v3.0.0?

For the workspace to work, dependencies need to be consistent. The problem is that we have the same dependency used and also in a replacement for the forbidden dependencies, then go complains with the following error:

```
conflicting replacements found for go.etcd.io/etcd/api/v3@v3.6.0-alpha.0 in workspace modules defined by [...]/etcd/etcd/go.mod and [...]/etcd/etcd/api/go.mod
```

It could technically be anything, but if we set it to a version like 3.6.0, we would need to change that when we release the 3.6.0 version.

> Also I don't see a need to change `./FORBIDDEN_DEPENDENCY` to `../FORBIDDEN_DEPENDENCY`, although not very important.

In my first attempt to enable the Go workspace, I couldn't figure out why the forbidden dependency trick wasn't working. I removed it and implemented how `k/k` does the forbidden dependencies. However, Tim pointed out that it works if you point this virtual dependency to the exact location across the different submodules. So, the top-level points them to `./FORBIDDEN_DEPENDENCY,` for the first-level (i.e., `./api`), it should point to `../FORBIDDEN_DEPENDENCY` and so on.

ivanvc:
> It might be a little risky to merge this into main branch right before we release 3.6. We may want to only merge it into 3.7.

I agree. I was trying to rush it to get it for 3.6.0.

> Also please only focus this PR only the necessary change related to go workspace. If it isn't necessary or any improvement, let's do it separately in follow PR.

I have an idea of what I could do. This pull request was:

1. Enabling the Go workspace.
2. Untangling the build and test scripts to use the workspace without changing directories (as now commands work from the top level).

If we go only with the first option, would we consider enabling the workspace for 3.6?

tests/robustness/makefile.mk
ivanvc: Good point. My bad.

api/go.mod
ahrtr: > I removed it and implemented how `k/k` does the forbidden dependencies.

How does K8s implement the forbidden dependencies? I don't see any forbidden directive in K8s's go.mod files at all.



> It could technically be anything, but if we set it to a version like 3.6.0, we would need to change that when we release the 3.6.0 version.

Even if we use "v3.0.0", we also need to manually update it when we release new major release, i.e. v4.0.0, but it's less frequent.

Can we raise a question in golang community to ask for best practice?

ahrtr:
> If we go only with the first option, would we consider enabling the workspace for 3.6?

Everything is working well without go workspace, so I suggest not to rush to integration workspace in such a short period before releasing v3.6.0.

go.work.sum
ivanvc: This is because while running `make build`, the `tools/` submodules are not touched. Then, it deletes those dependencies. Running `make update-go-workspace` reconciles the changes.

api/go.mod
ivanvc: > How does K8s implement the forbidden dependencies? I don't see any forbidden directive in K8s's go.mod files at all.

They use an internal tool [import-boss](https://github.com/kubernetes/kubernetes/tree/e90cfb83ed0cfed05f63ecf453fa21e36ae270d0/cmd/import-boss), which works with [.import-restrictions](https://github.com/kubernetes/kubernetes/blob/82e3a671e79d1740ab9a3b3fac8a3bb7d065a6fb/pkg/.import-restrictions) files.

ivanvc:
I'll close this pull request and re-work this feature by targeting the minimal code needed to enable the workspace without adding improvements.

Issue #19292, opened 2025-01-27
[robustness] kubernetes traffic is unable to send List request


fuweid:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Since https://github.com/etcd-io/etcd/pull/19137, the kubernetes traffic is unable to send list request because `end` field is reset.

https://github.com/etcd-io/etcd/blob/2895b19f536ce145c3ac84a88ee4afbca8b03265/tests/robustness/client/client.go#L90-L92

### What did you expect to happen?

kubernetes traffic api should send list request during robustness test

### How can we reproduce it (as minimally and precisely as possible)?

Run any robustness tests with kubernetes traffic

### Anything else we need to know?

client/v3 package doesn't export `limit` so it's hard to get all the information from `Op` provided by kubernetes.List.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```Shell

```

serathius:
cc @AwesomePatrol

serathius:
@AwesomePatrol Can you take a look and confirm if this is an issue?

PR #19137, opened 2025-01-07
Use new interface in robustness tests

AwesomePatrol:
Robustness tests now use `kubernetes.Interface` (where possible) replacing use of `kubernetesClient` (deleted)

PR #19319, opened 2025-02-03
*: introduce (*Op) Limit() interface for robustness

fuweid:
Since #19137, kubernetes traffic profile is unable to send List request with page size, because limit in option is not accessable. To fix it, this fix is to introduce Limit() interface.

Fixes: #19292


fuweid:
cc @serathius @ahrtr 

This is an API change, and I’m not sure if it’s acceptable. An alternative would be to revert #19137. I’d appreciate your feedback. Thanks!

serathius (approved):
cc @AwesomePatrol

ahrtr (approved):

PR #19325, opened 2025-02-03
fix test.sh to download correct binary for mac os

gangli113:
When testing on mac os, the script downloads incorrect binary, `ELF 64-bit LSB executable, x86-64, version 1 (SYSV)`, which is not executable on MAC OS. 

With this change, the script downloads executable binary, `Mach-O 64-bit executable x86_64`

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -576,7 +576,13 @@ function release_pass {
     log_warning "fallback to" ${UPGRADE_VER}
   fi
 
-  local file="etcd-$UPGRADE_VER-linux-$GOARCH.tar.gz"
+  local file
+  if [[ "$(uname -s)" == 'Darwin' ]]; then
+    file="etcd-$UPGRADE_VER-darwin-$GOARCH.zip"
+  else
+    file="etcd-$UPGRADE_VER-linux-$GOARCH.tar.gz"
+  fi
+
   log_callout "Downloading $file"
 
   set +e


scripts/test.sh
```diff
@@ -576,7 +576,13 @@ function release_pass {
     log_warning "fallback to" ${UPGRADE_VER}
   fi
 
-  local file="etcd-$UPGRADE_VER-linux-$GOARCH.tar.gz"
+  local file
+  if [[ "$OSTYPE" == "darwin"* ]]; then
```
ahrtr, line 580: I don't see where is this env variable set. Are you going to set & use it in a separate PR? Or it's just for contributor to use locally?

ahrtr:
cc @ivanvc

scripts/test.sh
ahrtr: It looks like most OS already supports this env variable.

ahrtr (approved):

scripts/test.sh
ivanvc: Would it be better to use "$(uname --kernel-name)"?

gangli113: I borrowed the example in https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script. Seems $OSTYPE is an environment variable set by Bash. https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

gangli113: Sorry didn't see your following comments before my last reply.

@ivanvc Thank you for the comment. Do you mean to get the OS type by running uname cmd? I tried the command on MAC os. the command 'uname --kernel-name' doesn't work. But `uname` without option works. Seems the options of uname command are different in different operating system.

```
$uname --kernel-name
uname: illegal option -- -
usage: uname [-amnoprsv]
$ uname -kernel-name
uname: illegal option -- k
usage: uname [-amnoprsv]
$ uname
Darwin
```

ivanvc: Sorry, I meant `uname -s`.

ivanvc (commented):
Thanks for the pull request, @gangli113. I left a comment on the possible usage of `uname`. I'm fine with us including that in a follow-up PR.

scripts/test.sh
gangli113: I see. Thank you!

I changed the code to use uname -s now.

However I cannot simplify the code to `"etcd-$UPGRADE_VER-$(uname -s | tr '[:upper:]' '[:lower:]')-$GOARCH.tar.gz"`, because on Mac OS, the file is a `zip` file and on Linux it's `tar.gz`

scripts/test.sh
```diff
@@ -576,7 +576,13 @@ function release_pass {
...
+    file="etcd-$UPGRADE_VER-darwin-$GOARCH.zip"
+  else
+    file="etcd-$UPGRADE_VER-linux-$GOARCH.tar.gz"
+  fi
+
```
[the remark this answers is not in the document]
gangli113: Seems not able to simplify it, because the file type are different.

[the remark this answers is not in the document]
ivanvc: Good point about the zip vs tarball. I don't think I can resolve this conversation, though :smile:

ivanvc (approved):
LGTM on green. Thanks, @gangli113.

ivanvc:
Ah, one thing, though, @gangli113. Can you please squash your commits? Thanks.

gangli113:
> Ah, one thing, though, @gangli113. Can you please squash your commits? Thanks.

@ivanvc Done. Thank you!

ivanvc (approved):
Thanks, @gangli113

ahrtr (approved):

PR #19326, opened 2025-02-03
fileutil: removes redundant continue statement in RemoveMatchFile

redwrasse:
Looks to my eyes that the `continue` statement in `RemoveMatchFile` is redundant? I see no logic below it in the enclosing for loop.


ivanvc (approved):
LGTM. Thanks, @redwrasse.

ahrtr (approved):

PR #19052, opened 2024-12-12
server: fix unexported-return lint issue

aladesawe:
PR to fix `unexported-return` lint issues under the `server` folder as mentioned in [Issue 18370](https://github.com/etcd-io/etcd/issues/18370). 

This PR:
- uses existing interfaces to fix

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #19343, opened 2025-02-05
Create an MembershipBackend interface to fix unexported-return

aladesawe:


aladesawe:
Pulled from [PR](https://github.com/etcd-io/etcd/pull/19052)

server/storage/schema/membership.go
```diff
@@ -32,12 +32,26 @@ const (
...
+	TrimClusterFromBackend() error
+	MustDeleteMemberFromBackend(id types.ID)
+	MustReadMembersFromBackend() (map[types.ID]*membership.Member, map[types.ID]bool)
+	TrimMembershipFromBackend() error
+	MustSaveClusterVersionToBackend(ver *semver.Version)
+	MustSaveDowngradeToBackend(downgrade *version.DowngradeInfo)
+	MustCreateBackendBuckets()
+	ClusterVersionFromBackend() *semver.Version
+	DowngradeInfoFromBackend() *version.DowngradeInfo
+}
```
ahrtr, line 47: It's duplicated to 

https://github.com/etcd-io/etcd/blob/f1d09c0e6badd3b8ce5c7aef685d23f706457be9/server/etcdserver/api/membership/store.go#L27-L32

Again, the simplest solution is to move it into storage/schema

aladesawe: @ahrtr do we want to import into schema versus moving it to schema? There's been no exported interface definition in schema

ahrtr: > @ahrtr Moving can't be done without a cyclic import; please refer to [commit](https://github.com/etcd-io/etcd/commit/8dc493e9b72a109a9262934a74700589ddd24e8a). tldr MembershipBackend is still referenced within the membership package

thx for pointing this out. Unfortunately, this existing design/implementation has some serious flaws. The `server/etcdserver/api/*` and `server/storage/*` are referencing (depending on) each other.
- Ideally, they shouldn't depend on each other, instead both of them should only depend on the same set of interfaces (abstract).
- The less optimised solution is to apply only one way reference (dependence)

Let me see what I can do.

ahrtr: Unfortunately it needs big refactor to resolve this. So I suggest that we only add `// revive:disable-next-line:unexported-return` to ignore it for now. In future (i.e 3.8 or 4.0), when we get rid of v2snapshot, we can revisit it.

server/storage/schema/membership.go
```diff
@@ -37,7 +37,7 @@ type membershipBackend struct {
 	be backend.Backend
 }
 
-func NewMembershipBackend(lg *zap.Logger, be backend.Backend) *membershipBackend {
+func NewMembershipBackend(lg *zap.Logger, be backend.Backend) membership.MembershipBackend {
```
ahrtr, line 40:
```suggestion
// revive:disable-next-line:unexported-return
func NewMembershipBackend(lg *zap.Logger, be backend.Backend) *membershipBackend {
```

ahrtr (approved):
LGTM

Thanks

ivanvc (approved):
LGTM. Thanks, @aladesawe.

PR #19344, opened 2025-02-05
etcdutl: use map to count unique user keys in snapshot status

### LINKED ISSUE #19253: etcdutl Snapshot Status: TotalKey Calculation Inquiry
### Discussed in https://github.com/etcd-io/etcd/discussions/19012

<div type='discussions-op-text'>

<sup>Originally posted by **jxustc** December  5, 2024</sup>
## Problem Description
When using the `etcdutl snapshot status` command, I've observed the following characteristics of TotalKey calculation:
- Current implementation counts keys across all buckets in the bbolt database
- Includes keys beyond visible etcd key-value data
- Key counts in other buckets may vary with different etcd versions

## Case
image

image

During  testing, I discovered that the TotalKey calculation includes keys from various internal buckets beyond the main key-value store:
### Identified Included Keys
- In `meta` bucket:
  - `term`
  - `scheduledCompactRev`
  - `confState`
  - ...
- In `auth` bucket:
  - `authRevision`
- Keys from other internal buckets

## Specific Questions
1. What is the precise definition of TotalKey?
2. Is the current calculation method appropriate?
3. Should the count be limited to etcd key-value entries?

## Potential Implementation Approach

```golang
func (s *v3Manager) Status(dbPath string) (ds Status, err error) {
	...
	if err = db.View(func(tx *bolt.Tx) error {
		...
		c := tx.Cursor()
		for next, _ := c.First(); next != nil; next, _ = c.Next() {
			b := tx.Bucket(next)
			if b == nil {
				return fmt.Errorf("nil bucket: %q", string(next))
			}
			_, err = h.Write(next)
			if err != nil {
				return fmt.Errorf("cannot hash bucket name: %q err: %w", string(next), err)
			}

			iskeyb := (bytes.Equal(next, schema.Key.Name()))
			if err = b.ForEach(func(k, v []byte) error {
				_, err = h.Write(k)
				if err != nil {
					return fmt.Errorf("cannot hash bucket key: %q err: %w", k, err)
				}
				_, err = h.Write(v)
				if err != nil {
					return fmt.Errorf("cannot hash bucket key: %q value: %q err: %w", k, v, err)
				}
				if iskeyb {
					var rev mvcc.Revision
					rev, err = bytesToRev(k)
					if err != nil {
						return fmt.Errorf("cannot parse revision key: %q err: %w", k, err)
					}
					ds.Revision = rev.Main

					var kv mvccpb.KeyValue
					err = kv.Unmarshal(v)
					if err != nil {
						return fmt.Errorf("cannot unmarshal value, key: %q value: %q err: %w", k, v, err)
					}
                                         // move to here 
                                         ds.TotalKey++
				}
                                 // remove 
				ds.TotalKey++
				return nil
			}); err != nil {
				return fmt.Errorf("error during bucket key iteration, name: %q err: %w", string(next), err)
			}
		}
		return nil
	}); err != nil {
		return ds, err
	}
...
}
```

**Besides this solution, could we possibly add a field to count the number of non-built-in key-value pairs in etcd?**

## Proposed Improvement

If appropriate, I am willing to submit a PR to implement this change. </div>

jxustc:
The new implementation:
- Uses map to track unique keys for accurate counting
- Excludes internal built-in keys from total count
- Improves code maintainability

Although this approach uses additional memory for the map, the trade-off is acceptable since:
- Status() is not in hot path
- Correctness takes priority over performance optimization
- Simpler code is easier to maintain

Fixes #19253

### DIFF
--- a/etcdutl/snapshot/v3_snapshot.go
+++ b/etcdutl/snapshot/v3_snapshot.go
@@ -126,6 +126,7 @@ func (s *v3Manager) Status(dbPath string) (ds Status, err error) {
 	defer db.Close()
 
 	h := crc32.New(crc32.MakeTable(crc32.Castagnoli))
+	seenKeys := make(map[string]struct{})
 
 	if err = db.View(func(tx *bolt.Tx) error {
 		// check snapshot file integrity first
@@ -175,8 +176,14 @@ func (s *v3Manager) Status(dbPath string) (ds Status, err error) {
 					if err != nil {
 						return fmt.Errorf("cannot unmarshal value, key: %q value: %q err: %w", k, v, err)
 					}
+					key := string(kv.Key)
+					// refer to https://etcd.io/docs/v3.5/learning/data_model/
+					if !mvcc.IsTombstone(k) {
+						seenKeys[key] = struct{}{}
+					} else {
+						delete(seenKeys, key)
+					}
 				}
-				ds.TotalKey++
 				return nil
 			}); err != nil {
 				return fmt.Errorf("error during bucket key iteration, name: %q err: %w", string(next), err)
@@ -187,6 +194,7 @@ func (s *v3Manager) Status(dbPath string) (ds Status, err error) {
 		return ds, err
 	}
 
+	ds.TotalKey = len(seenKeys)
 	ds.Hash = h.Sum32()
 	return ds, nil
 }
--- a/etcdutl/snapshot/v3_snapshot_test.go
+++ b/etcdutl/snapshot/v3_snapshot_test.go
@@ -118,6 +118,106 @@ func TestSnapshotStatusNegativeRevisionSub(t *testing.T) {
 	require.ErrorContains(t, err, "negative revision")
 }
 
+// TestSnapshotStatusTotalKey tests if snapshot status command correctly reports total number of valid keys.
+func TestSnapshotStatusTotalKey(t *testing.T) {
+	cases := []struct {
+		name     string
+		prepare  func(srv *etcdserver.EtcdServer)
+		expected int
+	}{
+		{
+			name: "duplicate keys",
+			prepare: func(srv *etcdserver.EtcdServer) {
+				keys := []string{"key1", "key2", "key1"}
+				val := make([]byte, len(keys))
+				for _, key := range keys {
+					for i := 0; i < 3; i++ {
+						req := etcdserverpb.PutRequest{
+							Key:   []byte(key),
+							Value: val,
+						}
+						_, err := srv.Put(context.TODO(), &req)
+						require.NoError(t, err)
+					}
+				}
+			},
+			expected: 2,
+		},
+		{
+			name: "mixed revisions",
+			prepare: func(srv *etcdserver.EtcdServer) {
+				// key1: create -> put -> delete
+				key := []byte("key1")
+				for i := 0; i < 3; i++ {
+					if i < 2 {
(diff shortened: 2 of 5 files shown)


ahrtr:
Thanks for the PR, but I am afraid that the fix isn't accurate. When a key has already been removed from user perspective, but it is still in the db file with tombstone flag set. We shouldn't count such keys.

Refer to https://etcd.io/docs/v3.5/learning/data_model/

jxustc:
> Thanks for the PR, but I am afraid that the fix isn't accurate. When a key has already been removed from user perspective, but it is still in the db file with tombstone flag set. We shouldn't count such keys.
> 
> Refer to https://etcd.io/docs/v3.5/learning/data_model/

Thanks for the feedback. I haven‘t notice that yet. I will fix it

etcdutl/snapshot/v3_snapshot.go
```diff
@@ -175,8 +179,17 @@ func (s *v3Manager) Status(dbPath string) (ds Status, err error) {
...
 					}
+					key := string(kv.Key)
+					// Version == 0 indicates logical deletion.
+					// refer to https://etcd.io/docs/v3.5/learning/data_model/
+					isActive := kv.Version > 0
+					if curr, exists := keyStates[key]; !exists || rev.Main > curr.rev {
+						keyStates[key] = struct {
+							rev    int64
+							active bool
+						}{rev.Main, isActive}
```
ahrtr, line 190: - Pls use `isTombstone` to check whether it's a tombstone. We can expose the function. https://github.com/etcd-io/etcd/blob/c6b9e9e9277121b347f63eb82d28c2c79611557e/server/storage/mvcc/revision.go#L120
- No need to compare revision. They are guaranteed to be increasing.

etcdutl/snapshot/v3_snapshot_test.go
```diff
@@ -118,6 +118,106 @@ func TestSnapshotStatusNegativeRevisionSub(t *testing.T) {
 	require.ErrorContains(t, err, "negative revision")
 }
 
+// TestSnapshotStatusDuplicateKeys tests if TotalKey in status remains unchanged when inserting the same key multiple times
+func TestSnapshotStatusDuplicateKeys(t *testing.T) {
```
ahrtr, line 122: All the test cases should be able to be consolidated into one table-driven sub tests. But not a big problem.

jxustc:
Thanks for the review. You make a good point about table-driven tests. I'm happy to refactor them. I also noticed some e2e test failures due to the changed TotalKey semantics - I'll fix those as well.

ahrtr (approved):
LGTM

Please also update 3.6 changelog  https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#etcdutl-v3

thx

ahrtr:
> Please also update 3.6 changelog https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#etcdutl-v3

It could be in a in a separate PR

ahrtr:
cc @fuweid @serathius

ahrtr:
> Please also update 3.6 changelog https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#etcdutl-v3

Done in https://github.com/etcd-io/etcd/pull/19361

jxustc:
> > Please also update 3.6 changelog https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.6.md#etcdutl-v3
> 
> Done in #19361

Thx

PR #19358, opened 2025-02-07
github workflows: remove release tests

ivanvc:
This workflow has already been migrated to the prow infrastructure as a presubmit job, and it's stable.

The current GitHub workflow also has a second part, which uses Trivy to check for image vulnerabilities in the generated images. However, these results overlap (or duplicate) what we obtain with `govulncheck`, as our images are based on a distroless static Debian image. Therefore, it only checks the etcd binaries. For example, when we had the report of GO-2024-2527, it never failed.

So, I think we can delete it. Or, maybe another approach would be to move it to a periodic job.


ahrtr:
> However, these results overlap (or duplicate) what we obtain with `govulncheck`, as our images are based on a distroless static Debian image

`govulncheck` is scanning the golang source code, while trivy is scanning the image. They are different. I think we need both.



> For example, when we had the report of GO-2024-2527, it never failed.

Do not get time to dig into the details. It shouldn't be a reason to remove trivy.

> maybe another approach would be to move it to a periodic job.

I think It's better to run trivy on each PR instead of periodically. It doesn't take too much time or resource.

ivanvc:
As this may be a longer conversation, I opened #19363.

ivanvc:
/cc @ahrtr, @serathius 

With an agreement on #19363. We can now remove the Trivy scan from this job.

ahrtr (approved):

serathius (approved):

PR #19365, opened 2025-02-09
Add a new cancelAfterDowngradingAndUpgrading scenario for downgrade e2e test

henrybear327:
Address https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634778319 by adding another test case where we
- select a set of nodes from the cluster
- stop those selected nodes, downgrade and start them
- stop those selected nodes, upgrade and start them 
- perform downgrade cancellation

Tested with the following command when developing locally
`clear; PASSES="release e2e" PKG=./tests/e2e TESTCASE="TestDowngradeCancellation" TIMEOUT=20m ./scripts/test.sh -v -failfast`

Tracking issue: https://github.com/etcd-io/etcd/issues/17976

### DIFF
--- a/tests/e2e/cluster_downgrade_test.go
+++ b/tests/e2e/cluster_downgrade_test.go
@@ -46,6 +46,7 @@ const (
 	cancelRightBeforeEnable
 	cancelRightAfterEnable
 	cancelAfterDowngrading
+	cancelAfterDowngradingAndUpgrading
 )
 
 func TestDowngradeUpgradeClusterOf1(t *testing.T) {
@@ -86,6 +87,14 @@ func TestDowngradeCancellationAfterDowngrading1InClusterOf3(t *testing.T) {
 
 func TestDowngradeCancellationAfterDowngrading2InClusterOf3(t *testing.T) {
 	testDowngradeUpgrade(t, 2, 3, false, cancelAfterDowngrading)
+}
+
+func TestDowngradeCancellationAfterDowngradingAndUpgrading1InClusterOf3(t *testing.T) {
+	testDowngradeUpgrade(t, 1, 3, false, cancelAfterDowngradingAndUpgrading)
+}
+
+func TestDowngradeCancellationAfterDowngradingAndUpgrading2InClusterOf3(t *testing.T) {
+	testDowngradeUpgrade(t, 2, 3, false, cancelAfterDowngradingAndUpgrading)
 }
 
 func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterSize int, triggerSnapshot bool, triggerCancellation CancellationState) {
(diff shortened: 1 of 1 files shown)


henrybear327:
@ahrtr hopefully this is what you were expecting for the additional test case! If not, I am happy to add more that aligns to your expectation. 

Nevertheless, this currently added test case should be tested IMHO anyways.

Thank you!

henrybear327:
`TestDowngradeCancellationAfterDowngradingAndUpgrading1InClusterOf3` is failing consistently. 

I am looking into the root cause of it

ahrtr:
I think there is a misunderstanding. 

The goal is to verify that users are able to restore a partially (including the case of completed downgraded) downgraded cluster to its original state as mentioned in https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634873365. The high level steps is mentioned in https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634778319.

There is no need to add another enum of `CancellationState`. You just need to try to restore the cluster to it's original state at the end of each downgrade test case, and verify below items,
- they have expected server versions and cluster version;
- the cluster is able to serve client requests.

henrybear327:
> I think there is a misunderstanding.
> 
> The goal is to verify that users are able to restore a partially (including the case of completed downgraded) downgraded cluster to its original state as mentioned in [#19252 (comment)](https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634873365). The high level steps is mentioned in [#19252 (comment)](https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634778319).
> 
> There is no need to add another enum of `CancellationState`. You just need to try to restore the cluster to it's original state at the end of each downgrade test case, and verify below items,
> 
> * they have expected server versions and cluster version;
> * the cluster is able to serve client requests.

I will ping you on slack for clarification!

henrybear327:
> > I think there is a misunderstanding.
> > The goal is to verify that users are able to restore a partially (including the case of completed downgraded) downgraded cluster to its original state as mentioned in [#19252 (comment)](https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634873365). The high level steps is mentioned in [#19252 (comment)](https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634778319).
> > There is no need to add another enum of `CancellationState`. You just need to try to restore the cluster to it's original state at the end of each downgrade test case, and verify below items,
> > 
> > * they have expected server versions and cluster version;
> > * the cluster is able to serve client requests.
> 
> I will ping you on slack for clarification!

Message sent. 

The test case I added is another scenario where x nodes were downgraded and then upgraded. Then the downgrade cancellation is called (basically adding a new downgrade cancellation call timing, which I think is better to be done in the robustness test since we can't cover all potential timing cases in e2e tests).

ahrtr:
I just realised that my comment https://github.com/etcd-io/etcd/pull/19252#issuecomment-2634873365 has already been supported in the existing test case, see below, 

https://github.com/etcd-io/etcd/blob/fa2926adb02b1f59d9ed26c38871335ad642f2aa/tests/e2e/cluster_downgrade_test.go#L200-L205




> Then the downgrade cancellation is called (basically adding a new downgrade cancellation call timing,

It seems unnecessary.

Probably we should verify that the downgrade can be cancelled automatically when the downgrade is completed. But I regard it as optional.

We should be all good now.

Thanks for all the effort.

ahrtr:
> Probably we should verify that the downgrade can be cancelled automatically when the downgrade is completed. But I regard it as optional.

FYI.  https://github.com/etcd-io/etcd/blob/1ea4f435cc449216fbaa9aec16c408a734fc43e9/server/etcdserver/server.go#L2410

henrybear327:
> > Probably we should verify that the downgrade can be cancelled automatically when the downgrade is completed. But I regard it as optional.
> 
> FYI.
> 
> https://github.com/etcd-io/etcd/blob/1ea4f435cc449216fbaa9aec16c408a734fc43e9/server/etcdserver/server.go#L2410

Thanks for the pointer!

I will close this PR and submit a new one according to the discussion we have here!

PR #19366, opened 2025-02-09
Increase the timeout value for e2e downgrade tests

henrybear327:
Since the e2e downgrade tests exhibits timeout on the CI more frequently than expected, this PR attempts to increase timeout and see if it will alleviate the situation.

See one of the timeout being flagged [here](https://kubernetes.slack.com/archives/C3HD8ARJ5/p1738927033745799)


tests/framework/e2e/downgrade.go
```diff
@@ -56,7 +56,7 @@ func DowngradeCancel(t *testing.T, epc *EtcdProcessCluster) {
 	c := epc.Etcdctl()
 
 	var err error
-	testutils.ExecuteWithTimeout(t, 1*time.Minute, func() {
+	testutils.ExecuteWithTimeout(t, 2*time.Minute, func() {
```
ahrtr, line 59: Can you elaborate why it takes so long to cancel the downgrade? @henrybear327 @siyuanfoundation

tests/framework/e2e/downgrade.go
```diff
@@ -143,7 +143,7 @@ func ValidateMemberVersions(t *testing.T, epc *EtcdProcessCluster, expect []*ver
 }
 
 func ValidateVersion(t *testing.T, cfg *EtcdProcessClusterConfig, member EtcdProcess, expect version.Versions) {
-	testutils.ExecuteWithTimeout(t, 1*time.Minute, func() {
+	testutils.ExecuteWithTimeout(t, 2*time.Minute, func() {
```
ahrtr, line 146: same question why it takes so long?

tests/framework/e2e/downgrade.go
henrybear327: I haven't had a solid conclusion so far. I will have to investigate deeper to be able to answer
1. How often are we getting the time out if we use 1 min
2. What are the usual errors

tests/framework/e2e/downgrade.go
henrybear327: ditto (https://github.com/etcd-io/etcd/pull/19366#discussion_r1949843196)

tests/framework/e2e/downgrade.go
ahrtr: @siyuanfoundation @henrybear327 can you take this as a high priority? I see multiple times of timeout of the downgrade e2e cases.

60s should be already long enough. We need a clearer understanding why it takes so long. 

also cc @fuweid

fuweid: I am checking this flaky case in my local. I think it maybe related to case. So, I think increasing timeout isn't working

ahrtr: @fuweid Thanks for looking into this. Just assigned https://github.com/etcd-io/etcd/issues/19391 to you. Feel free to let me know if you need discussion tomorrow.

henrybear327: I will keep looking into this tomorrow morning as my first task.

ahrtr:
See fix in https://github.com/etcd-io/etcd/pull/19398 and discussion in https://github.com/etcd-io/etcd/issues/19391

PR #19399, opened 2025-02-12
Add e2e downgrade automatic cancellation test

henrybear327:
Please see https://github.com/etcd-io/etcd/pull/19365#issuecomment-2649299898


ahrtr (approved):
LGTM


cc @fuweid @siyuanfoundation @serathius

ahrtr:
@henrybear327 can you please rebase this PR ? I just merged https://github.com/etcd-io/etcd/pull/19398

tests/e2e/cluster_downgrade_test.go
```diff
@@ -298,14 +380,23 @@ func getMembersAndKeys(t *testing.T, cc *e2e.EtcdctlV3) (*clientv3.MemberListRes
 	return members, kvs
 }
 
-func generateIdenticalVersions(clusterSize int, ver *semver.Version) []*version.Versions {
+func generateIdenticalVersions(t *testing.T, clusterSize int, ver *semver.Version) []*version.Versions {
 	ret := make([]*version.Versions, clusterSize)
 
+	// storage version string is non-empty starting from 3.6.0
+	ver3_6_0, err := semver.NewVersion("3.6.0")
```
ahrtr, line 387: can you reuse V3_6? https://github.com/etcd-io/etcd/blob/f30cbaac111aa01a310fe75360c759cdd4d9cd14/api/version/version.go#L44

tests/e2e/cluster_downgrade_test.go
```diff
@@ -207,6 +207,88 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
...
+	epc := newCluster(t, clusterSize, snapshotCount)
+	for i := 0; i < len(epc.Procs); i++ {
+		e2e.ValidateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{
+			Cluster: currentVersionStr,
+			Server:  version.Version,
+			Storage: currentVersionStr,
+		})
+	}
+	cc := epc.Etcdctl()
+	t.Logf("Cluster created")
```
ahrtr, line 247: nit
```suggestion
	t.Log("Cluster created")
```

tests/e2e/cluster_downgrade_test.go
```diff
@@ -207,6 +207,88 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
...
+	require.NoError(t, err)
+	beforeMembers, beforeKV := getMembersAndKeys(t, cc)
+
+	e2e.DowngradeEnable(t, epc, lastVersion)
+
+	t.Logf("Starting downgrade process for node 0 to %q", lastVersionStr)
+	err = e2e.DowngradeUpgradeMembersByID(t, nil, epc, []int{0}, currentVersion, lastClusterVersion)
+	require.NoError(t, err)
+
+	e2e.DowngradeCancel(t, epc)
```
ahrtr, line 267: This case is to test auto downgrade cancel as mentioned in https://github.com/etcd-io/etcd/pull/19365#issuecomment-2650053733, then we don't need to manually cancel it.
```suggestion
```

tests/e2e/cluster_downgrade_test.go
henrybear327: Ahh yes that would be better! Thanks!

tests/e2e/cluster_downgrade_test.go
henrybear327: I thought the idea is that
- we start the downgrade
- half way through (e.g x nodes got downgraded), we cancel the downgrade
- we continue on the downgrade
- we finish the downgrade process
- we check that indeed, there isn't anymore downgrade inflight, so the cancellation will fail with error "no inflight cancellation"

Does this make sense to you?

We can add a flag to test both the case you mentioned and the existing PR's case! What do you think? :)

tests/e2e/cluster_downgrade_test.go
henrybear327: Addressed all similar types of issues

tests/e2e/cluster_downgrade_test.go
ahrtr: > half way through (e.g x nodes got downgraded), we cancel the downgrade

it makes this new e2e test case pointless.

tests/e2e/cluster_downgrade_test.go
```diff
@@ -207,6 +207,88 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
...
+	afterMembers, afterKV := getMembersAndKeys(t, cc)
+	assert.Equal(t, beforeKV.Kvs, afterKV.Kvs)
+	assert.Equal(t, beforeMembers.Members, afterMembers.Members)
+
+	if len(epc.Procs) > 1 {
+		t.Log("Waiting health interval to required to make membership changes")
+		time.Sleep(etcdserver.HealthInterval)
+	}
+
+	e2e.DowngradeAutoCancelCheck(t, epc)
```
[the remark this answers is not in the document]
henrybear327: So it's better to 
- monitor/wait for the log entry "the cluster has been downgraded"
- call `downgradeCancellation` to verify that it's indeed cancelled

Right?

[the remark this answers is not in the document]
ahrtr: The first item is required, the second one is nice to have (optional).

tests/e2e/cluster_downgrade_test.go
henrybear327: I think it's not pointless! 

My take is that if the downgrade cancellation is triggered, we can still finish the downgrade if we proceed to do so! This e2e test verifies this is actually the case.

WDYT?

tests/e2e/cluster_downgrade_test.go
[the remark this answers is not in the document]
henrybear327: Got it! Will impl. in a bit

tests/e2e/cluster_downgrade_test.go
ahrtr: > My take is that if the downgrade cancellation is triggered, we can still finish the downgrade

It's true.

But this test case is to verify the downgrade can be automatically cancelled. If you manually cancel it, the final validation `DowngradeAutoCancelCheck` will always pass for sure. what's the point?

ahrtr: > > My take is that if the downgrade cancellation is triggered, we can still finish the downgrade
> 
> It's true.

Also FYI. it works in the case of downgrade from 3.6 --> 3.5. But it's problematic if there is any protoc change; it's dangerous to continue to downgrade if the schema hasn't been migrated due to the downgrade cancellation.

ahrtr: Suggest you to do some archeology (try to find out the original downgrade issue / PRs/ docs), to have a better understanding on the downgrade feature.

ahrtr (changes requested):
Please resolve the comments

tests/e2e/cluster_downgrade_test.go
```diff
@@ -301,11 +383,17 @@ func getMembersAndKeys(t *testing.T, cc *e2e.EtcdctlV3) (*clientv3.MemberListRes
 func generateIdenticalVersions(clusterSize int, ver *semver.Version) []*version.Versions {
 	ret := make([]*version.Versions, clusterSize)
 
+	// storage version string is non-empty starting from 3.6.0
+	storageStr := ver.String()
+	if ver.Compare(version.V3_6) == -1 {
+		storageStr = ""
+	}
```
ahrtr, line 390:
```suggestion
	if ver.LessThan(version.V3_6) {
		storageStr = ""
	}
```

tests/framework/e2e/downgrade.go
```diff
@@ -117,6 +131,14 @@ func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcess
 			return err
 		}
 	}
+
+	if opString == "downgrading" && len(membersToChange) == len(clus.Procs) {
+		lg.Info("Waiting for downgrade completion log line")
+		leader := clus.WaitLeader(t)
+		_, err := clus.Procs[leader].Logs().ExpectWithContext(context.TODO(), expect.ExpectedResponse{Value: "the cluster has been downgraded"})
```
henrybear327, line 138: @ahrtr I tried to implement [this comment](https://github.com/etcd-io/etcd/pull/19399/files#r1952973254) by monitoring the log for the leader. But I am not sure if this is the best way to go about it (e.g., the leader is re-elected during the time I am monitoring the log, etc.)

ahrtr, line 138: Please set a timeout, i.e. 30s.

henrybear327: Done as requested!

Do you see a risk with this approach of monitoring the leader's log @ahrtr? :)

tests/e2e/cluster_downgrade_test.go
```diff
@@ -207,6 +207,77 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
 	assert.Equal(t, beforeMembers.Members, afterMembers.Members)
 }
 
+func TestAutomaticDowngradeCancellationAfterCompletingDowngradingInClusterOf3(t *testing.T) {
```
ahrtr, line 210: The name is too long.
```suggestion
func TestDowngradeAutoCancelAfterCompletion(t *testing.T) {
```

tests/framework/e2e/downgrade.go
```diff
@@ -82,6 +83,19 @@ func DowngradeCancel(t *testing.T, epc *EtcdProcessCluster) {
...
 }
 
+func DowngradeAutoCancelCheck(t *testing.T, epc *EtcdProcessCluster) {
+	c := epc.Etcdctl()
+
+	var err error
+	testutils.ExecuteWithTimeout(t, 1*time.Minute, func() {
+		t.Log("etcdctl downgrade cancel")
+		err = c.DowngradeCancel(context.TODO())
+		require.Errorf(t, err, "no inflight downgrade job")
```
ahrtr, line 93:
```suggestion
		require.ErrorContains(t, err, "no inflight downgrade job")
```

tests/e2e/cluster_downgrade_test.go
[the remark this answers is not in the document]
ahrtr: Theoretically, when you see the log message "`the cluster has been downgraded`", the leader hasn't send out the downgradeCancel request yet. So when you try to send downgradeCancel yourself in `e2e.DowngradeAutoCancelCheck`, it may succeed. I think it's easy to verify if you add a sleep in between.

But in practice, it's unlikely because there is a 5s (`etcdserver.HealthInterval`) sleep after the downgrade completion.

So either just fix comment https://github.com/etcd-io/etcd/pull/19399/files#r1957283593 or just remove `e2e.DowngradeAutoCancelCheck` (preferred)

tests/framework/e2e/downgrade.go
ahrtr: The only concern is it will fail if the leader somehow changes.

So suggest to add a retry mechanism (i.e. up to 3 times), and let's wait 15s instead of 30s in each retry.

henrybear327:
Is this PR still relevant as we are doing https://github.com/etcd-io/etcd/issues/19439?

ahrtr:
> Is this PR still relevant as we are doing #19439?

Yes, I think so. The auto downgrade cancel should still work.

tests/e2e/cluster_downgrade_test.go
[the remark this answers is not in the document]
siyuanfoundation: I suggest add a TODO item to followup once the downgrade query API is implemented. https://github.com/etcd-io/etcd/issues/19439#issuecomment-2666755025

ahrtr:
FYI. https://github.com/etcd-io/etcd/pull/19451#discussion_r1965899599


Thanks for the effort anyway. @henrybear327

ahrtr:
> FYI. [#19451 (comment)](https://github.com/etcd-io/etcd/pull/19451#discussion_r1965899599)

3.5 has the auto downgrade cancellation as well, so this test case still makes sense. @fuweid @henrybear327 

https://github.com/etcd-io/etcd/blob/release-3.5/server/etcdserver/server.go#L2724

ahrtr:
> > FYI. [#19451 (comment)](https://github.com/etcd-io/etcd/pull/19451#discussion_r1965899599)
> 
> 3.5 has the auto downgrade cancellation as well, so this test case still makes sense. @fuweid @henrybear327
> 
> https://github.com/etcd-io/etcd/blob/release-3.5/server/etcdserver/server.go#L2724

But 3.5 doesn't have the downgrade status query API. So this PR should be independent to https://github.com/etcd-io/etcd/pull/19451. So there is no need to wait for that PR.  @fuweid @henrybear327 @siyuanfoundation

henrybear327:
> > > FYI. [#19451 (comment)](https://github.com/etcd-io/etcd/pull/19451#discussion_r1965899599)
> > 
> > 
> > 3.5 has the auto downgrade cancellation as well, so this test case still makes sense. @fuweid @henrybear327
> > https://github.com/etcd-io/etcd/blob/release-3.5/server/etcdserver/server.go#L2724
> 
> But 3.5 doesn't have the downgrade status query API. So this PR should be independent to #19451. So there is no need to wait for that PR. @fuweid @henrybear327 @siyuanfoundation

So ... is the conclusion to still merge this PR or no? Sorry that I am pretty lost during these exchanges.

ahrtr:
> So ... is the conclusion to still merge this PR or no?

YES. It's nice to merge this PR before next Tuesday. But it isn't a blocker for 3.6.0-rc.1.

henrybear327:
> > So ... is the conclusion to still merge this PR or no?
> 
> YES. It's nice to merge this PR before next Tuesday. But it isn't a blocker for 3.6.0-rc.1.

Got it! I have rebased it now, but I will likely complete addressing all comments and incorporating changes regarding cancellation status by Monday noon (as I am out on both days on the weekend).

tests/framework/e2e/downgrade.go
henrybear327: Done as suggested

tests/e2e/cluster_downgrade_test.go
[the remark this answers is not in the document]
henrybear327: `e2e.DowngradeAutoCancelCheck(t, epc)` removed.

henrybear327:
Ready for a final pass!

tests/e2e/cluster_downgrade_test.go
```diff
@@ -228,6 +228,85 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
...
+	t.Log("Downgrade should be disabled")
+	e2e.ValidateDowngradeInfo(t, epc, &pb.DowngradeInfo{Enabled: false})
+
+	t.Log("Adding member to test membership, but a learner avoid breaking quorum")
+	resp, err := cc.MemberAddAsLearner(context.Background(), "fake1", []string{"http://127.0.0.1:1001"})
+	require.NoError(t, err)
+	t.Log("Removing learner to test membership")
+	_, err = cc.MemberRemove(context.Background(), resp.Member.ID)
+	require.NoError(t, err)
+	beforeMembers, beforeKV := getMembersAndKeys(t, cc)
```
siyuanfoundation, line 283: I think we should add `writeKVs()` before this, and in the other tests too.

tests/e2e/cluster_downgrade_test.go
```diff
@@ -228,6 +228,85 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
...
+	assert.Equal(t, beforeKV.Kvs, afterKV.Kvs)
+	assert.Equal(t, beforeMembers.Members, afterMembers.Members)
+
+	if len(epc.Procs) > 1 {
+		t.Log("Waiting health interval to required to make membership changes")
+		time.Sleep(etcdserver.HealthInterval)
+	}
+
+	t.Log("Downgrade should be disabled")
+	e2e.ValidateDowngradeInfo(t, epc, &pb.DowngradeInfo{Enabled: false})
```
siyuanfoundation, line 304: This validation does not seem very useful since the `DowngradeInfo` would always be false in 3.5. Which means we probably do not need this test ...

PR #11034, opened 2019-08-14
mvcc: Optimize compaction for short commit pauses

jpbetz:
Reduce compaction intervals to 1k and duration between compactions to 10ms.

Also forces immediate commit of the write buffer after each compaction (thanks @jingyih for this).

I'll also make this configurable via a flag.

Supersedes #11021

PR #11021, opened 2019-08-12
mvcc/backend: Optimize etcd compaction

jpbetz:
etcd's current compaction batch size is 10k and compactions are executed via mvcc backend write transaction. This can result in severely degraded p99 latency under heavy load each time a compaction occurs, due, in large part, to the batch size.

This PR reduces the batch size to 1k.

Also, the compaction writes are currently accumulated in the mvcc backend write buffer which must later perform a large write while holding the mvcc backend's main read/write lock, preventing new reads from initializing (i.e. ongoing reads may progress concurrently, but new ones cannot begin). But since compaction process removes only boltdb records for revisions that are no longer visible to readers. The main benefit of writing via a mvcc backend write transaction--buffering of writes in a way that ensure they remain visible to readers--is not needed. So this PR change performs compaction writes directly against boltdb. Avoiding lock contention in the mvcc backend and skipping the write buffer. This improves performance significantly.  The kube sig-scalability team tested this against 5k kubemark clusters and validated that it largely eliminates degraded p99 latency.

Even though this is simpler and more efficient. The boltdb writes that perform compaction block the mvcc backend write buffer from performing writeback. When this happens, the writeback will hold the main mvcc backend read/write lock, preventing new reads from initializing. Reducing the batch size to 1k helps minimize this (if the batch size is too low, we risk losing throughput).

Potential future work:
- Find a more optimal batch size)
- Improve how the mvcc backend manages locks to only acquire the main mvcc backend read/write lock once it has acquired the boltdb write lock, to increase concurrency.

There is a potential edge case where writes in the write buffer that are not yet flushed to boltdb become eligible for compaction. By deleting compacted revisions directly from boltdb, these revisions would not be compacted like they should. But this is not a problem because they are not visible to readers, and the next compaction operation after the write buffer is flushed will delete them from boltdb.

PR #19405, opened 2025-02-12
[release-3.5] Fix a performance regression due to uncertain compaction sleep interval

miancheng7:
The compaction behavior [here](https://github.com/etcd-io/etcd/blob/v3.5.16/server/mvcc/kvstore_compaction.go#L93) is changed in commit
[02635](https://github.com/etcd-io/etcd/commit/0263597ba84f65047948141696be926e53aa2429) and introduces a latency issue. To be more speicific, the `ticker.C`  acts as a fixed timer that triggers every 10ms, regardless of how long each batch of compaction takes. This means that if a previous compaction batch takes longer than 10ms, the next batch starts immediately, making compaction a blocking operation for etcd.

To fix the issue, this commit revert the compaction to the previous behavior which ensures a 10ms delay between each batch of compaction, allowing other read and write operations to proceed smoothly.

issue https://github.com/etcd-io/etcd/issues/19406

### DIFF
--- a/server/mvcc/kvstore_compaction.go
+++ b/server/mvcc/kvstore_compaction.go
@@ -39,8 +39,6 @@ func (s *store) scheduleCompaction(compactMainRev, prevCompactRev int64) (KeyVal
 	binary.BigEndian.PutUint64(end, uint64(compactMainRev+1))
 
 	batchNum := s.cfg.CompactionBatchLimit
-	batchTicker := time.NewTicker(s.cfg.CompactionSleepInterval)
-	defer batchTicker.Stop()
 	h := newKVHasher(prevCompactRev, compactMainRev, keep)
 	last := make([]byte, 8+1+8)
 
@@ -90,7 +88,7 @@ func (s *store) scheduleCompaction(compactMainRev, prevCompactRev int64) (KeyVal
 		dbCompactionPauseMs.Observe(float64(time.Since(start) / time.Millisecond))
 
 		select {
-		case <-batchTicker.C:
+		case <-time.After(s.cfg.CompactionSleepInterval):
 		case <-s.stopc:
 			return KeyValueHash{}, fmt.Errorf("interrupted due to stop signal")
 		}


miancheng7:
cc @chaochn47 for visibility

server/mvcc/kvstore_compaction.go
```diff
@@ -90,7 +88,7 @@ func (s *store) scheduleCompaction(compactMainRev, prevCompactRev int64) (KeyVal
 		dbCompactionPauseMs.Observe(float64(time.Since(start) / time.Millisecond))
 
 		select {
-		case <-batchTicker.C:
+		case <-time.After(s.cfg.CompactionSleepInterval):
```
hakuna-matatah, line 91: - Curious why its defaulted to `10ms` today in the first place   ?

- As I understand it, compaction is taking longer than `10ms` in clusters with large data sets, leading to noticeable latency increases on client side (that's how this issue is surfaced iiuc). 
So, I was wondering if we can adjust `sleep interval dynamically`  based on `pending apply requests` instead of creating a room for `10ms` ?. The more apply requests pending, the longer the sleep interval, thus reducing client latency(for reads/writes) during high load scenarios and could be beneficial in large clusters. 
wdyt ?

JalinWang: For q1, 10ms is introduced by https://github.com/etcd-io/etcd/pull/11034, and it seems like an empirical param https://github.com/etcd-io/etcd/pull/11021#issuecomment-520789359 (not the PR author's comment)

JalinWang: For question 2, regardless of the duration of a compaction batch, the current implementation does not guarantee a full CompactionSleepInterval sleep. Given that CompactionSleepInterval is already configurable, I think we could  address this issue first and then introduce an adaptive sleep mode in a separate pull request :)

serathius: For backport please stick to minimal change that will address the issue

serathius:
Please start from sending PR to the `main` branch, get it merged and then backport to release-3.6 and release3.5 branches.

server/mvcc/kvstore_compaction.go
ahrtr: +1 to keep the minimal change in this PR.

For any improvement, let's address & discuss it separately.

ahrtr (approved):
LGTM

Thank you!

ahrtr:
Let me send a PR to main now.

ahrtr:
Link to the PR on main https://github.com/etcd-io/etcd/pull/19410

ahrtr:
cc @fuweid @serathius

fuweid (approved):
Changes looks good. 

Just thinking how to handle backport in the future. Since we have merged the change into main branch, should we cherry-pick it or just merge this one? My suggestion is that we should address issue in main branch and then backport if applicable.

However, this change came first. We can merge this.

ahrtr:
> Just thinking how to handle backport in the future. Since we have merged the change into main branch, should we cherry-pick it or just merge this one?

Usually fix goes into main firstly, then backport to stable releases. But not a big problem. It's OK as long as we ensure the fix is merged on higher versions firstly.

server/mvcc/kvstore_compaction.go
hakuna-matatah: Yes, I am totally all for merging this for sure to fix the problem at hand. 

> let's address & discuss it separately.

I just left the community to understand what community thinks about this to see if it makes sense to go in that direction ?   I can probably cut a separate issue to channel this discussion.

ahrtr: Sorry, do not get time to dig into your proposal yet. Please feel free to raise a separate ticket to present your thought/idea if you want. But you need to convince us that it's indeed a real issue and demo the real benefit to go in that direction. Overall I do not see a high priority to change it for now.

ivanvc:
/retitle [3.5] Fix a performance regression due to uncertain compaction sleep interval

wenjiaswe:
> Just thinking how to handle backport in the future. Since we have merged the change into main branch, should we cherry-pick it or just merge this one? My suggestion is that we should address issue in main branch and then backport if applicable.

Agree to @fuweid  we should address issue in main branch and then backport if applicable. This is generally the best practice for several reasons: centralized Fixes, reduced Risk, clear commit History.

PR #19423, opened 2025-02-15
[do not review] Go workspace

ivanvc:
Minimal changes to introduce a Go workspace.

Do not review WIP.


scripts/test.sh
```diff
@@ -508,7 +508,7 @@ function bom_pass {
 
 function dump_deps_of_module() {
   local module
-  if ! module=$(run go list -m); then
+  if ! module=$(run go mod edit -json | jq -r .Module.Path); then
```
serathius, line 511: Could we merge it independent to migration to workspaces as a separate PR? Having less changes in the main PR will increase confidence during review.

scripts/updatebom.sh
```diff
@@ -14,7 +14,7 @@ function bom_fixlet {
   # shellcheck disable=SC2207
   modules=($(modules_for_bom))
 
-  if GOFLAGS=-mod=mod run_go_tool "github.com/appscodelabs/license-bill-of-materials" \
```
serathius, line 17: `GOFLAGS=-mod=mod` is this even a correct flag? Could we remove it regardless of workspace migration?

scripts/release_mod.sh
```diff
@@ -31,7 +31,7 @@ function update_module_version() {
   local v2version="${2}"
   local modules
   run go mod tidy
-  modules=$(run go list -f '{{if not .Main}}{{if not .Indirect}}{{.Path}}{{end}}{{end}}' -m all)
+  modules=$(go mod edit -json | jq -r '.Require[] | select(.Indirect == null) | .Path')
```
serathius, line 34: When I run the two commands now I get different results, is that different when we enable workspaces?
```
$ go list -f '{{if not .Main}}{{if not .Indirect}}{{.Path}}{{end}}{{end}}' -m all | sort
github.com/anishathalye/porcupine
github.com/coreos/go-semver
github.com/golang/protobuf
github.com/google/go-cmp
github.com/grpc-ecosystem/go-grpc-middleware
github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus
github.com/grpc-ecosystem/grpc-gateway/v2
github.com/prometheus/client_golang
github.com/prometheus/client_model
github.com/prometheus/common
github.com/soheilhy/cmux
github.com/stretchr/testify
go.etcd.io/etcd/api/v3
go.etcd.io/etcd/client/pkg/v3
go.etcd.io/etcd/client/v2
go.etcd.io/etcd/client/v3
go.etcd.io/etcd/etcdctl/v3
go.etcd.io/etcd/etcdutl/v3
go.etcd.io/etcd/pkg/v3
go.etcd.io/etcd/server/v3
go.etcd.io/gofail
go.etcd.io/raft/v3
golang.org/x/crypto
golang.org/x/sync
golang.org/x/time
google.golang.org/grpc
google.golang.org/protobuf
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
go.opentelemetry.io/otel
go.opentelemetry.io/otel/sdk
go.opentelemetry.io/otel/trace
go.opentelemetry.io/proto/otlp
go.uber.org/zap
```
```
$ go mod edit -json | jq -r '.Require[] | select(.Inderect == null) | .Path' | sort
github.com/anishathalye/porcupine
github.com/beorn7/perks
github.com/bgentry/speakeasy
github.com/cenkalti/backoff/v4
github.com/cespare/xxhash/v2
github.com/cheggaaa/pb/v3
github.com/coreos/go-semver
github.com/coreos/go-systemd/v22
github.com/creack/pty
github.com/davecgh/go-spew
github.com/dustin/go-humanize
github.com/fatih/color
github.com/gogo/protobuf
github.com/golang/groupcache
github.com/golang-jwt/jwt/v5
github.com/golang/protobuf
github.com/go-logr/logr
github.com/go-logr/stdr
github.com/google/btree
github.com/google/go-cmp
github.com/google/uuid
github.com/gorilla/websocket
github.com/grpc-ecosystem/go-grpc-middleware
github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus
github.com/grpc-ecosystem/go-grpc-middleware/v2
github.com/grpc-ecosystem/grpc-gateway/v2
github.com/inconshreveable/mousetrap
github.com/jonboulle/clockwork
github.com/klauspost/compress
github.com/mattn/go-colorable
github.com/mattn/go-isatty
github.com/mattn/go-runewidth
github.com/munnerz/goautoneg
github.com/olekukonko/tablewriter
github.com/pmezard/go-difflib
github.com/prometheus/client_golang
github.com/prometheus/client_model
github.com/prometheus/common
github.com/prometheus/procfs
github.com/rivo/uniseg
github.com/sirupsen/logrus
github.com/soheilhy/cmux
github.com/spf13/cobra
github.com/spf13/pflag
github.com/stretchr/testify
github.com/tmc/grpc-websocket-proxy
github.com/VividCortex/ewma
github.com/xiang90/probing
go.etcd.io/bbolt
go.etcd.io/etcd/api/v3
go.etcd.io/etcd/client/pkg/v3
go.etcd.io/etcd/client/v2
go.etcd.io/etcd/client/v3
go.etcd.io/etcd/etcdctl/v3
go.etcd.io/etcd/etcdutl/v3
go.etcd.io/etcd/pkg/v3
go.etcd.io/etcd/server/v3
go.etcd.io/gofail
go.etcd.io/raft/v3
golang.org/x/crypto
golang.org/x/net
golang.org/x/sync
golang.org/x/sys
golang.org/x/text
golang.org/x/time
google.golang.org/genproto/googleapis/api
google.golang.org/genproto/googleapis/rpc
google.golang.org/grpc
google.golang.org/protobuf
go.opentelemetry.io/auto/sdk
go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
go.opentelemetry.io/otel
go.opentelemetry.io/otel/exporters/otlp/otlptrace
go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc
go.opentelemetry.io/otel/metric
go.opentelemetry.io/otel/sdk
go.opentelemetry.io/otel/trace
go.opentelemetry.io/proto/otlp
gopkg.in/natefinch/lumberjack.v2
gopkg.in/yaml.v3
go.uber.org/multierr
go.uber.org/zap
sigs.k8s.io/json
sigs.k8s.io/yaml
```

ahrtr:
I will review this PR after we officially release v3.6.0, thx

scripts/test.sh
ivanvc: Great idea. Thanks, Marek :bow:. I'll break down this pull request even further, as most of the commits can coexist with the current codebase.

scripts/updatebom.sh
ivanvc: Yes, I think so. This should predate Go 1.11.

scripts/test.sh
ivanvc: I opened #19463.

scripts/release_mod.sh
ivanvc: This observation led me down a rabbit hole. It took me a while to review why running `go list—m all` differs from the output of doing `go mod edit—json`. 

It made me realize I had an issue with my `jq` expression. So, I revised the expression. The amended version is: `.Require[] | select(.Indirect | not) | .Path`. And I wrote the following to double-check that the output matches:

```bash
$ find . -name go.mod -exec /bin/bash -c 'echo evaluating $(dirname {}); cd $(dirname {}); diff <(go list -f "{{if not .Main}}{{if not .Indirect}}{{.Path}}{{end}}{{end}}" -m all | sort) <(go mod edit -json | jq -r ".Require[] | select(.Indirect | not) | .Path" | sort)' \;
evaluating ./client/pkg
evaluating ./client/v3
evaluating ./client/internal/v2
evaluating ./etcdctl
evaluating ./etcdutl
evaluating ./pkg
evaluating ./server
evaluating ./tests
evaluating ./tools/mod
evaluating ./tools/rw-heatmaps
evaluating ./tools/testgrid-analysis
evaluating ./api
evaluating .
```

So, there are no differences.

My rabbit hole, investigation, solution, and verification may have been an overkill :sweat_smile:, as in the context of the original function, it filters the output only to keep `go.etcd.io/etcd/*/v{2,3}` modules (to do the version bump), which are always in the `go.mod`. But thanks for the good catch :) I'll update the code and open a pull request targeting this change, as it can coexist with the current code without the workspace.

scripts/updatebom.sh
ivanvc: Ah, never mind. Actually, I believe, due to the note in `bom_pass`, we may need to leave the `-mod=mod` flag in the tests. But you're right. There's no need for it to be here, as `test_lib.sh` doesn't define a `GOFLAGS=-mod` that we need to override. I'll open a pull request for this change.

scripts/release_mod.sh
serathius: Thanks for looking into that.

ivanvc: Opened #19546.

ivanvc:
@ahrtr, with 3.6.0 out the door, can we revisit this pull request now?

scripts/updatebom.sh
ivanvc: I attempted to make this change stand-alone in #19467 but failed to achieve it. Refer to https://github.com/etcd-io/etcd/pull/19467#issuecomment-2689953166. I'll mark this comment as resolved for now.

serathius (approved):

ahrtr:
Will take a look at this PR next week, thx @ivanvc

api/go.mod
```diff
@@ -31,9 +31,9 @@ require (
...
 // This makes the error more explicit.
 replace (
-	go.etcd.io/etcd => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/api/v3 => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/pkg/v3 => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/tests/v3 => ./FORBIDDEN_DEPENDENCY
-	go.etcd.io/etcd/v3 => ./FORBIDDEN_DEPENDENCY
+	go.etcd.io/etcd => ../FORBIDDEN_DEPENDENCY
+	go.etcd.io/etcd/api/v3 v3.999.999 => ../FORBIDDEN_DEPENDENCY
+	go.etcd.io/etcd/pkg/v3 v3.999.999 => ../FORBIDDEN_DEPENDENCY
```
ahrtr, line 36: Why it's v3.999.999? In this example, we should prevent module `api` from depending on `pkg` of all versions, not just specific to a fake v3.999.999.

ahrtr:
I assume that you will remove all permitted replace directive in a followup PR, see one example below, correct?

https://github.com/etcd-io/etcd/blob/ae01ad849a1bd9deb2ca282220b7d98e13900dfc/etcdctl/go.mod#L49-L54

api/go.mod
ivanvc: Refer to our conversation from https://github.com/etcd-io/etcd/pull/19314#discussion_r1939585736

scripts/update_go_workspace.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+#
+# Copyright 2024 The etcd Authors
```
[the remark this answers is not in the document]
ivanvc: Done. Thanks for the catch.

api/go.mod
ivanvc: Because Go requires specific versions, it would be for anything in the "3" major version. To the best of my knowledge, there is no way to do this across multiple major versions. If you specify a different version (other than what's defined in the rest of the project) for a module like `go.etcd.io/etcd/api/v3`, Go will complain with the error:

```
conflicting replacements found for go.etcd.io/etcd/api/v3@v3.6.0 in workspace modules
defined by [...]/etcd/etcd/go.mod and [...]/etcd/etcd/api/go.mod
```

For more context for the 3.999.999 fictitious version, refer to b5cf2ecef316b5cec4e6635a030a3cc9b099e878 commit message
> Point all forbidden dependencies to a common virtual place (top-level
FORBIDDEN_DEPENDENCY). For the workspace to function we need all
dependencies to be consistent across the repository. That means that the
current approach with FORBIDDEN_DEPENDENCY doesn't work, because they
are pointing to different directories. Set the fictional v3.999.999
version for these dependencies, so they are consitent, and they don't
clash with the actual depdendencies references in other modules.

ivanvc:
> I am aware of it, and have no any comment on that. I am just double confirming the next step. Even in your previous PR #19314, I do not see you made the change as mentioned in [#19423 (comment)](https://github.com/etcd-io/etcd/pull/19423#issuecomment-2912641428)

If I remember correctly, we can't remove the replaces from the top-level `go.mod`, but we can from the submodules (but my memory may not be serving me well). I'll work on that after, and will get a definitive answer on what we need to do moving forward.

ivanvc:
Rebased/force pushed due to #20039.

api/go.mod
ahrtr: Probably I did not say it clearly. 

There are two things.

## Top level `FORBIDDEN_DEPENDENCY` is good
All pointing to the same top level `FORBIDDEN_DEPENDENCY` is good. I have no comment on this.

## version `v3.999.999` might not correct
For the version `v3.999.999`, I am afraid it isn't correct. As I mentioned above, it only prevents us from importing that specific version. Let's work with an example, the pkg can't depend on api, because we have line below,

https://github.com/etcd-io/etcd/blob/b3b9e986e037727a53ea08e5d7f8672cd5397fc7/pkg/go.mod#L43

When you execute command `go get go.etcd.io/etcd/api/v3@v3.6.0` under the pkg module, you will get error below,

```
$ go get go.etcd.io/etcd/api/v3@v3.6.0
go: go.etcd.io/etcd/api/v3@v3.6.0: replacement directory ./FORBIDDEN_DEPENDENCY does not exist
```

However if you add a version like below,
```
go.etcd.io/etcd/api/v3 v3.999.999 => ./FORBIDDEN_DEPENDENCY
```

then you can add api into pkg's go.mod because it only prevents you from importing v3.999.999. see below,

```
$ go get go.etcd.io/etcd/api/v3@v3.6.0
go: added go.etcd.io/etcd/api/v3 v3.6.0

$ git diff
diff --git a/pkg/go.mod b/pkg/go.mod
index ed27778fb..3f6d63aef 100644
--- a/pkg/go.mod
+++ b/pkg/go.mod
@@ -21,6 +21,7 @@ require (
        github.com/google/go-cmp v0.7.0 // indirect
        github.com/inconshreveable/mousetrap v1.1.0 // indirect
        github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+       go.etcd.io/etcd/api/v3 v3.6.0 // indirect
        go.opentelemetry.io/otel v1.36.0 // indirect
        go.opentelemetry.io/otel/sdk v1.36.0 // indirect
        go.uber.org/multierr v1.11.0 // indirect
@@ -40,7 +41,7 @@ replace go.etcd.io/etcd/client/pkg/v3 => ../client/pkg
 // Shouldn't import unnecessary dependencies
 replace (
        go.etcd.io/etcd => ./FORBIDDEN_DEPENDENCY
-       go.etcd.io/etcd/api/v3 => ./FORBIDDEN_DEPENDENCY
+       go.etcd.io/etcd/api/v3 v3.999.999 => ./FORBIDDEN_DEPENDENCY
        go.etcd.io/etcd/tests/v3 => ./FORBIDDEN_DEPENDENCY
        go.etcd.io/etcd/v3 => ./FORBIDDEN_DEPENDENCY
 )
diff --git a/pkg/go.sum b/pkg/go.sum
index a74117318..3a7e61cf4 100644
--- a/pkg/go.sum
+++ b/pkg/go.sum
@@ -35,6 +35,8 @@ github.com/spf13/pflag v1.0.6 h1:jFzHGLGAlb3ruxLB8MhbI6A8+AQX/2eW4qeyNZXNp2o=
 github.com/spf13/pflag v1.0.6/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An2Bg=
 github.com/stretchr/testify v1.10.0 h1:Xv5erBjTwe/5IxqUQTdXv5kgmIvbHo3QQyRwhJsOfJA=
 github.com/stretchr/testify v1.10.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
+go.etcd.io/etcd/api/v3 v3.6.0 h1:vdbkcUBGLf1vfopoGE/uS3Nv0KPyIpUV/HM6w9yx2kM=
+go.etcd.io/etcd/api/v3 v3.6.0/go.mod h1:Wt5yZqEmxgTNJGHob7mTVBJDZNXiHPtXTcPab37iFOw=
 go.opentelemetry.io/auto/sdk v1.1.0 h1:cH53jehLUN6UFLY71z+NDOiNJqDdPRaXzTel0sJySYA=
 go.opentelemetry.io/auto/sdk v1.1.0/go.mod h1:3wSPjt5PWp2RhlCcmmOial7AvC4DQqZb7a7wCow3W8A=
 go.opentelemetry.io/otel v1.36.0 h1:UumtzIklRBY6cI/lllNZlALOF5nNIzJVb16APdvgTXg=

```

If you change the definition above to `go.etcd.io/etcd/api/v3 v3.6.0 => ./FORBIDDEN_DEPENDENCY`, then the command `go get go.etcd.io/etcd/api/v3@v3.6.0` will definitely fail. 

## Conclusion & Proposal?

**Conclusions**:
- The change in this PR breaks the function of `FORBIDDEN_DEPENDENCY`, because it can't technically prevents us from importing any version other than v3.999.999.
- Keeping v3.999.999 is better not keeping it, at least it has document purpose.

**Proposal**: 
Can we raise a question on the [golang-nuts](https://groups.google.com/g/golang-nuts) to get more professional / official advice?

ivanvc: Thank you for the clear explanation and for helping me with my misunderstanding.

I see and understand the point. Before reaching out to the Go mailing list, I propose an alternative to using the fictitious `3.999.999` version: keep these versions in sync, pointing to the latest released version (i.e., as part of our release script).

We could also reach out to Tim Hockin, as he proposed this solution after I attempted to use k/k's import-boss to solve this issue.

serathius: cc @thockin

ahrtr: > keep these versions in sync, pointing to the latest released version (i.e., as part of our release script).

Yes, I had the same thought. It's much better although It still has similar flaw, which can also be detected by linter check (inconsistent dependencies). Let me provide an example to avoid misunderstanding (please also double confirm this),
- Change the definition to `go.etcd.io/etcd/api/v3 v3.6.0 => ../FORBIDDEN_DEPENDENCY`
- So you can't import module `api/v3@v3.6.0`; but you can import other version, i.e. `api/v3@v3.5.21`
- But the dependencies will be inconsistent (`api/v3@v3.5.21` vs `api/v3@v3.6.0`, assuming all other permitted modules depend on `api/v3@v3.6.0`), so the linter workflow check will fail.

Please try this out.

ivanvc: Sorry for the late reply. I couldn't get back to this until now.

@thockin, we're discussing setting a fictional version for the forbidden dependencies in the go workspace. And how it won't work, as it doesn't forbid using the released version.

> Change the definition to go.etcd.io/etcd/api/v3 v3.6.0 => ../FORBIDDEN_DEPENDENCY
@ahrtr, we can't do this. We would have the following error:

```
conflicting replacements found for go.etcd.io/etcd/api/v3@v3.6.0-alpha.0 in workspace modules defined by ...
```

The issue is that versions need to be consistent both in the `required` section and in `replace`. Therefore, I actually retract what I said before:

> keep these versions in sync, pointing to the latest released version (i.e., as part of our release script).

This won't work, as it will generate conflicting replacements.

At this point, I'm running out of ideas on how to do it within `go.mod`. I'm leaning towards building a poorman's approach to `import-boss`. (i.e., adding a new `.forbidden-dependencies` file and a shell script that ensures those dependencies are not in the `go.mod`). I searched golang's GitHub and the go-nuts mailing list for similar requests/issues, even though I found some related https://github.com/golang/go/issues/32780 / https://github.com/golang/go/issues/25588, they were closed a long time ago. I'll reach out to the mailing list to see if someone has an idea on how to do this, but at least my searches haven't yielded any conclusive results.

ivanvc: After a lot of investigation, I landed on using the `depguard` linter. By using it, we can remove this `go.mod` hack and let `golangci-lint` fail instead, if a module requires a forbidden dependency. Refer to commit c7d7744ef3c6caa9ddaf88aa54fc5b36103eb7ac.

With this intentional diff:

```diff
diff --git a/api/go.mod b/api/go.mod
index 2a825b489..d2d5f0d62 100644
--- a/api/go.mod
+++ b/api/go.mod
@@ -10,6 +10,7 @@ require (
        github.com/golang/protobuf v1.5.4
        github.com/grpc-ecosystem/grpc-gateway/v2 v2.27.2
        github.com/stretchr/testify v1.11.1
+       go.etcd.io/etcd/pkg/v3 v3.6.5
        google.golang.org/genproto/googleapis/api v0.0.0-20250825161204-c5933d9347a5
        google.golang.org/grpc v1.75.1
        google.golang.org/protobuf v1.36.8
diff --git a/api/go.sum b/api/go.sum
index f14bec486..0c03985b4 100644
--- a/api/go.sum
+++ b/api/go.sum
@@ -31,6 +31,8 @@ github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
 github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
+go.etcd.io/etcd/pkg/v3 v3.6.5 h1:byxWB4AqIKI4SBmquZUG1WGtvMfMaorXFoCcFbVeoxM=
+go.etcd.io/etcd/pkg/v3 v3.6.5/go.mod h1:uqrXrzmMIJDEy5j00bCqhVLzR5jEJIwDp5wTlLwPGOU=
 go.opentelemetry.io/auto/sdk v1.1.0 h1:cH53jehLUN6UFLY71z+NDOiNJqDdPRaXzTel0sJySYA=
 go.opentelemetry.io/auto/sdk v1.1.0/go.mod h1:3wSPjt5PWp2RhlCcmmOial7AvC4DQqZb7a7wCow3W8A=
 go.opentelemetry.io/otel v1.37.0 h1:9zhNfelUvx0KBfu/gb+ZgeAfAgtWrfHJZcAqFC228wQ=
diff --git a/api/version/version.go b/api/version/version.go
index 31ca2f14d..f469a01fa 100644
--- a/api/version/version.go
+++ b/api/version/version.go
@@ -21,6 +21,8 @@ import (
        "strings"
 
        "github.com/coreos/go-semver/semver"
+
+       "go.etcd.io/etcd/pkg/v3/httputil"
 )
 
 var (
@@ -66,6 +68,7 @@ type Versions struct {
 
 // Cluster only keeps the major.minor.
 func Cluster(v string) string {
+       httputil.GracefulClose(nil)
        vs := strings.Split(v, ".")
        if len(vs) <= 2 {
                return v
```

Running the linter fails with:

```
golangci-lint version: v2.4.0
PASSES="lint" ./scripts/test.sh
Running with --race
Starting at: Wed Sep 24 12:07:33 AM PDT 2025
'lint' started at Wed Sep 24 12:07:33 AM PDT 2025
% (cd api && 'golangci-lint' 'run' '--config' '/var/home/ivan/Code/Personal/etcd/etcd/tools/.golangci.yaml' '--show-stats=false' './...')
FAIL: (code:1):
  % (cd api && 'golangci-lint' 'run' '--config' '/var/home/ivan/Code/Personal/etcd/etcd/tools/.golangci.yaml' '--show-stats=false' './...')
../api/version/version.go:25:2: import 'go.etcd.io/etcd/pkg/v3/httputil' is not allowed from list 'api' (depguard)
        "go.etcd.io/etcd/pkg/v3/httputil"
        ^
FAIL: 'run golangci-lint run --config /var/home/ivan/Code/Personal/etcd/etcd/tools/.golangci.yaml --show-stats=false ./...' checking failed (!=0 return code)
There was a Failure in module api, aborting...
FAIL: 'lint' FAILED at Wed Sep 24 12:07:33 AM PDT 2025
make: *** [Makefile:126: verify-lint] Error 255
```

ivanvc:
`pull-etcd-verify` is failing due to the bom check. Before I managed to make it fail consistently in a just cloned repository. I have a workaround for this. I think it may be a combination of Go 1.25 and the workspace that is making it fail. I'll address it soon.

serathius:
Using depguard looks great. Could we separate PR to migrate to depguard? That would allow us to make smaller, more intentional steps and main PR smaller.

ivanvc:
> Using depguard looks great. Could we separate PR to migrate to depguard? That would allow us to make smaller, more intentional steps and main PR smaller.

Great idea! I'll open a PR shortly :)

api/go.mod
```diff
@@ -1,6 +1,6 @@
 module go.etcd.io/etcd/api/v3
 
-go 1.25
+go 1.25.0
```
serathius, line 3: Why set to 1.25.0 while toolchain is on 1.25.1?

ivanvc: ~~Human error due to rebasing 😅😆~~

Nevermind, this is the minimum Go version. We usually leave w/o the patch until a dependency forces us to set it to the "0" patch (1.25.0 in this case).

ivanvc:
The @k8s-ci-robot reports that `pull-etcd-govulncheck-main` failed. But this is an old test that we don't have anymore. We replaced it with `pull-etcd-govulncheck`, and the run is successful: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19423/pull-etcd-govulncheck/1974246855806029824.

ahrtr (approved):
LGTM

Thanks for the great work!

serathius:
This is awesome!

PR #19424, opened 2025-02-15
[WIP] github/workflows: remove coverage workflow

ivanvc:
:warning: **WIP: I'm testing if the upload to CodeCov still works from the Prow job** :warning: 

This workflow has already been migrated to the Prow infrastructure.


scripts/codecov_upload.sh
```diff
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
 
-# Script used to collect and upload test coverage (mostly by travis).
-# Usage ./test_coverage_upload.sh [log_file]
+# Script used to collect and upload test coverage.
 
 set -o pipefail
 
 # We try to upload whatever we have:
-bash <(curl -s https://codecov.io/bash) -f ./covdir/all.coverprofile -cF all
```
ahrtr, line 9: FYI. I saw an error related to this script in another PR,

```
'cov' PASSED and completed at Sat Feb 22 17:08:13 UTC 2025
              
SUCCESS
./scripts/codecov_upload.sh
/dev/fd/63: line 2: syntax error near unexpected token `<'
/dev/fd/63: line 2: `<html><head>'
make: *** [Makefile:67: upload-coverage-report] Error 2
+ EXIT_VALUE=2
+ set +o xtrace
```

Refer to https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19426/pull-etcd-coverage-report/1893341510284349440

ivanvc: It is highly likely that Codecov answered with an HTTP error. We could try to add retries or just run a `/retest` (or vendorize the script, which I don't think is a good idea). The previous behavior was not to do any retrying, but it should be easy to implement this. Let me know your thoughts.

ahrtr: > or vendorize the script,

I don't think it's a good idea either. We don't want to maintain/update it periodically.




> The previous behavior was not to do any retrying, but it should be easy to implement this. Let me know your thoughts.

I think we should at least to avoid misleading error message. The error should be something like "failed to download the codecov script" instead of "syntax error near unexpected token `<'".  Can we breakdown the one line script into multiple lines something like below,

```
curl -s https://codecov.io/bash -o codecov.sh
bash codecov.sh -f ./covdir/all.coverprofile -cF all
```

ivanvc: I added a `-f` to `curl`. So, it will make it fail if it returns something other than 200.

ahrtr (approved):

ivanvc:
@ahrtr, could we get this merged? Or do we want/need a second reviewer?

scripts/codecov_upload.sh
```diff
@@ -1,9 +1,18 @@
...
+curl -sf -o ./bin/codecov.sh https://codecov.io/bash
+
+bash ./bin/codecov.sh -f "${COVERDIR}/all.coverprofile" \
+  -cF all \
+  -C "${PULL_PULL_SHA}" \
+  -r "${REPO_OWNER}/${REPO_NAME}" \
+  -P "${PULL_NUMBER}" \
+  -b "${BUILD_ID}" \
+  -B "${PULL_BASE_REF}" \
+  -N "${PULL_BASE_SHA}"
```
ahrtr, line 18: This seems to be an unrelated enhancement. It would be better to do it separately. But I won't insist on it.

ivanvc: It's not unrelated because we will publish the reports from the Prow job; we need to specify these variables/arguments. In the GitHub actions environment, CodeCov would assume these values.

PR #19425, opened 2025-02-16
Tooling for markdown linting

### LINKED ISSUE #18059: Find a replacement for marker (markdown linter)

Right now we're using [marker](https://github.com/crawford/marker) as the markdown linter. However, I see two issues:

1. Its functionality is limited, from its `README`
    > This is a tool for finding issues in [CommonMark](http://commonmark.org/) documentation. Right now, it only identifies broken links (malformed URLs, non-existent paths, etc.).
2. It's written in Rust, and while dropping our GitHub action for static checking (#18058). We wont run this linter in the Prow infrastructure as we don't have Rust/Cargo available (refer to https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/etcd-io_etcd/18058/pull-etcd-verify/1793352130300481536#1:build-log.txt%3A659-662).

While trying to fix 2, I opened a PR crawford/marker#32 to be able to use a published binary, rather than building the project itself.

However, it would be better to use a Go-written Markdown linter as alternative.

### Why is this needed?

To keep linting markdown files.

Zanda256:
With this pull request, we add a bash script to run linting check on modified or newly added markdown files. We then add a makefile target to run the script using make. 

Partially resolves #18059

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,3 +218,8 @@ verify-go-versions:
 .PHONY: sync-toolchain-directive
 sync-toolchain-directive:
 	./scripts/sync_go_toolchain_directive.sh
+
+.PHONY: markdown-diff-lint
+markdown-diff-lint:
+	./scripts/markdown_diff_lint.sh
+	
--- a/scripts/markdown_diff_lint.sh
+++ b/scripts/markdown_diff_lint.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+# This script runs markdownlint-cli2 on changed files.
+# Usage: ./markdown_lint.sh
+
+
+# We source ./scripts/test_lib.sh, it sets the log functions and color variables.
+source ./scripts/test_lib.sh
+
+# When we source ./scripts/test_lib.sh, it has the line set -u which treats unset variables as errors.
+# We need to unset the variable to avoid the error.
+set +u -eo pipefail
+
+if ! command markdownlint-cli2 dummy.md &>/dev/null; then
+  log_error "markdownlint-cli2 needs to be installed."
+  log_error "Please refer to https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#install for installation instructions."
+  exit 1
+fi
+
+
+if [ -z "${PULL_BASE_SHA}" ]; then
+  echo "Empty base reference (\$PULL_BASE_SHA), assuming: main"
+  PULL_BASE_SHA=main
+fi
+
+if [ -z "${PULL_PULL_SHA}" ]; then
+  PULL_PULL_SHA="$(git rev-parse HEAD)"
+  echo "Empty pull reference (\$PULL_PULL_SHA), assuming: ${PULL_PULL_SHA}"
+fi
+
+MD_LINT_URL_PREFIX="https://github.com/DavidAnson/markdownlint/blob/main/doc/"
+
+mapfile -t changed_files < <(git diff "${PULL_BASE_SHA}" --name-only)
+declare -A files_with_failures start_ranges end_ranges
+
+for file in "${changed_files[@]}"; do
+  if ! [[ "$file" =~ .md$ ]]; then
+    continue
+  fi
+
+  # Find start and end ranges from changed files.
+  start_ranges=()
+  end_ranges=()
+  # From https://github.com/paleite/eslint-plugin-diff/blob/46c5bcf296e9928db19333288457bf2805aad3b9/src/git.ts#L8-L27
+  ranges=$(git diff "${PULL_BASE_SHA}" \
+           --diff-algorithm=histogram \
+           --diff-filter=ACM \
+           --find-renames=100% \
+           --no-ext-diff \
+           --relative \
+           --unified=0 -- "${file}" | \
+    gawk 'match($0, /^@@\s-[0-9,]+\s\+([0-9]+)(,([0-9]+))?/, m) { \
+               print m[1] ":" m[1] + ((m[3] == "") ? "0" : m[3]) }')
+  i=0
+  for range in ${ranges}; do
+    start_ranges["${i}"]=$(echo "${range}" | awk -F: '{print $1}')
+    end_ranges["${i}"]=$(echo "${range}" | awk -F: '{print $2}')
+    i=$((1 + i))
+  done
+  if [ -z "${ranges}" ]; then
+    start_ranges[0]=0
+    end_ranges[0]=0
+  fi
+
+  i=0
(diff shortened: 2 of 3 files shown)


joshjms:
#18059

Zanda256:
@ivanvc  could we please give the failing tests a re-run?

ivanvc:
> @ivanvc could we please give the failing tests a re-run?

Yes, you can also do it by yourself by commenting:

/retest

Note: I haven't forgotten about this pull request, but I haven't been able to review it as I have other things in my backlog. I'll return to this soon :v:

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,99 @@
...
+COLOR_ORANGE='\033[0;33m'
+COLOR_NONE='\033[0m' # No Color
+
+function log_error {
+  echo -n -e "${COLOR_BOLD}${COLOR_RED}$*${COLOR_NONE}\n"
+}
+
+function log_warning {
+  echo -n -e "${COLOR_ORANGE}$*${COLOR_NONE}\n"
+}
```
ivanvc, line 24: We can benefit from sourcing `scripts/test_lib.sh`, which defines these functions and color codes. Therefore, we can remove these definitions.

Makefile
```diff
@@ -218,3 +218,7 @@ verify-go-versions:
 .PHONY: sync-toolchain-directive
 sync-toolchain-directive:
 	./scripts/sync_go_toolchain_directive.sh
+
+.PHONY: markdown-diff-lint
+markdown-diff-lint:
+	./scripts/markdown_diff_lint.sh
```
ivanvc, line 224: nit: Can we keep the linebreak in the last line of the file?

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,99 @@
...
+done
+
+echo "Finished linting"
+
+for file in "${!files_with_failures[@]}"; do
+  log_error "${file} has linting issues"
+done
+if [ "${#files_with_failures[@]}" -gt "0" ]; then
+  exit 1
+fi
```
ivanvc, line 99: nit: Can we keep the linebreak in the last line of the file?

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
+# This script runs markdownlint-cli2 on changed files.
+# Usage: ./markdown_lint.sh
+
+set -eo pipefail
+
+if ! command markdownlint-cli2 dummy.md &>/dev/null; then
+  echo "markdownlint-cli2 needs to be installed."
+  echo "Please refer to https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#install for installation instructions."
```
ivanvc, line 9: If we source `test_lib.sh`. These two lines can be changed to `log_errror ...`.

ivanvc (commented):
Thanks for the pull request, @Zanda256. I'm sorry for not getting back to you sooner. I left a couple of comments/suggestions. Thanks again :)

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,95 @@
...
+    i=$((1 + i))
+  done
+  if [ -z "${ranges}" ]; then
+    start_ranges[0]=0
+    end_ranges[0]=0
+  fi
+
+  i=0
+  # Run markdownlint-cli2 with the changed file and print only the summary (stdout).
+  markdownlint-cli2 "${file}" 2>/dev/null || true
```
ivanvc, line 66: Can you please add `--config` here, so we can replace the default location of the configuration file?

```suggestion
markdownlint-cli2 "${file}" --config "${ETCD_ROOT_DIR}/tools/.markdownlint.jsonc" 2>/dev/null || true
```

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,95 @@
...
+
+    if [ "${i}" -lt "${#start_ranges[@]}" ] && [ "${line_number}" -ge "${start_ranges["${i}"]}" ] && [ "${line_number}" -le "${end_ranges["${i}"]}" ]; then
+      # Inside range with changes, raise an error.
+      log_error "${lint_error}"
+      files_with_failures["${file}"]=1
+    else
+      # Outside of range, raise a warning.
+      log_warning "${lint_error}"
+    fi
+  done < <(markdownlint-cli2 "${file}" 2>&1 >/dev/null || true)
```
ivanvc, line 83: Same here:

```suggestion
  done < <(markdownlint-cli2 "${file}" --config "${ETCD_ROOT_DIR}/tools/.markdownlint.jsonc" 2>/dev/null || true)
```

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,95 @@
...
+echo "Finished linting"
+
+for file in "${!files_with_failures[@]}"; do
+  log_error "${file} has linting issues"
+done
+if [ "${#files_with_failures[@]}" -gt "0" ]; then
+  exit 1
+fi
+
+set -u
```
ivanvc, line 95: Why did you add this `set -u` at the end of the script?

ivanvc (commented):
Thanks for the changes, @Zanda256. Could you:

1. Please add a new file `tools/.markdownlint.jsonc`? The contents of the file should be `{}`
2. Squash your commits into a single one?

This is looking good. I tested locally and it works as expected, we just need these latest changes. Thanks again!

scripts/markdown_diff_lint.sh
Zanda256: We source testlib.sh which has sets the flag to treat unset variables as errors. In the markdown script we have some unset variables. I left a comment about it [here](https://github.com/Zanda256/etcd/blob/c869284a05f4dab05f133874fba6998a253e8ca8/scripts/markdown_diff_lint.sh#L16)

ivanvc: Ah, I missed those lines (https://github.com/etcd-io/etcd/pull/19425/files#diff-1c685ee93aa9be97d91c96d5815453479d9203aee1552c824b567cc0360c27e7R16-R18)

Can you move those lines to the top of the file, along with the `set—eo pipefail`? And because this script is intended to be executed and not sourced, I don't think we need `set -u` here.

Zanda256: Okay great. Let me do that and squash the commits.

ivanvc (approved):
LGTM. Thanks, @Zanda256.

/cc @jmhbnz

Issue #19306, opened 2025-01-29
DowngradeUpgrade failpoint timeout


siyuanfoundation:
### Which Github Action / Prow Jobs are flaking?

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-etcd-robustness-main-amd64/1884310668266442752

### Which tests are flaking?

DowngradeUpgrade

### Reason for failure (if possible)

The DowngradeUpgrade failpoint stops and restarts up etcd servers up to 6 times. It could take a long time for a new server to join a cluster.
So it is likely to see the test fail from time to time because Failpoints are expected to finish within 60s. 

We should 
- [ ] make Failpoint respect context
- [ ] reduce the time it takes for DowngradeUpgrade or 
- [ ] make it possible increase the timeout for specific failpoints

serathius:
I don't see any significant difference from increasing the timeout. Interesting the biggest change in test results happened between Feb 2-3. 

![Image](https://github.com/user-attachments/assets/07c1559b-59f4-40d0-8207-b54638bfd0c3)


No difference on Feb 6-7
![Image](https://github.com/user-attachments/assets/507f6b6e-0647-4c94-9266-be87051b4523)

Main failures come from `ClusterOfSize1`. Do we have any idea why this could be problematic?

fuweid:
After https://github.com/etcd-io/etcd/pull/19451 merged, the robustness CI has been green for 2 days. Closing it.

cc @serathius

serathius:
Awesome work @fuweid !

PR #19427, opened 2025-02-16
*: deflake MemberDowngradeUpdate robustness case

fuweid:
### 1. Update DowngradeUpgradeMembersByID Interface

By default, the leader cancels the downgrade process once all members reach the target version. However, this behavior is unpredictable and may cause unexpected errors.

Example 1: Single-Member Cluster

  T0: Set up a single-member cluster.
  T1: Enable the downgrade process.
  T2: Downgrade the member from v3.6.0 to v3.5.0.
  T3: Upgrade the member from v3.5.0 to v3.6.0.

If the leader cancels the downgrade process between T2 and T3, the cluster should end up at v3.6.0; otherwise, it remains at v3.5.0.

Example 2: Three-Member Cluster

  T0: Set up a three-member cluster.
  T1: Enable the downgrade process.
  T2: Downgrade two members from v3.6.0 to v3.5.0.
  T3: Upgrade those two members back to v3.6.0.

Since the downgrade process is incomplete, after T3, the cluster should remain at v3.5.0 instead of v3.6.0.

Key Takeaways

  * The binary version alone should not determine the cluster version.
  * The caller must explicitly track and manage the expected version.

### 2. Manually cancels the downgrade process after a successful downgrade in MemberDowngradeUpdate failpoint

As demonstrated in Example 1, canceling downgrade process can reduce flakiness.

### 3. Remove AssertProcessLogs from DowngradeEnable

The log message "The server is ready to downgrade" appears only when the storage version monitor detects a mismatch between the cluster and storage versions.

If traffic is insufficient to trigger a commit or if an auto-commit occurs right after reading the storage version, the monitor may fail to update it, leading to errors like:

```bash
"msg":"failed to update storage version","cluster-version":"3.6.0",
"error":"cannot detect storage schema version: missing confstate information"
```

Given this, we should remove the AssertProcessLogs statement because we won't receive log 'The server is ready to downgrade'.

Similar to #19313

cc @serathius @ahrtr @henrybear327 @siyuanfoundation 

Fixes: https://github.com/etcd-io/etcd/issues/19306

### DIFF
--- a/tests/framework/e2e/downgrade.go
+++ b/tests/framework/e2e/downgrade.go
@@ -46,7 +46,6 @@ func DowngradeEnable(t *testing.T, epc *EtcdProcessCluster, ver *semver.Version)
 			Server:  OffsetMinor(ver, 1).String(),
 			Storage: ver.String(),
 		})
-		AssertProcessLogs(t, epc.Procs[i], "The server is ready to downgrade")
 	}
 
 	t.Log("Cluster is ready for downgrade")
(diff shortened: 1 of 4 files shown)


ahrtr:
> If the leader cancels the downgrade process between T2 and T3, the cluster should end up at v3.6.0; otherwise, it remains at v3.5.0.

It's expected behaviour.  If the users executed upgrade right after downgrade, then they need to manually cancel the downgrade at the end to ensure the cluster is in the expected state. 

The Auto downgrade cancel is just to help the users to do something automatically, but it isn't the AUTO downgrade cancel's fault to leading to the situation you mentioned, it's the users' fault, so they need to manually cancel the downgrade at the end. We need to clearly document this. 


> Since the downgrade process is incomplete, after T3, the cluster should remain at v3.5.0 instead of v3.6.0.

It's also expected behaviour. As mentioned above, users need to cancel the downgrade themselves if you downgrade and upgrade back and forth.



> As demonstrated in Example 1, disabling the downgrade monitor reduces test flakiness.

I think we need to update the test case. Users need to cancel the downgrade after upgrading the member again.

fuweid:
Hi @ahrtr Yes, all the behavior is expected. For automation test cases, enabling auto-downgrade without checking the etcd member log to confirm its readiness for cancellation may increase flakiness. This patch aims to deflake existing test cases.

> > As demonstrated in Example 1, disabling the downgrade monitor reduces test flakiness.
>
> I think we need to update the test case. Users need to cancel the downgrade after upgrading the member again.

In robustness testing, at this moment disabling auto-cancel-downgrade in MemberDowngradeUpdate may make maintenance easier.
That said, we should define our goals for such cases. If we want to have both the older and newer versions running simultaneously, disabling auto-cancel-downgrade can make maintenance easier. However, if the goal is to test reliability during downgrading and upgrading, we should consider adding a delay after downgrade before upgrading. If that's the case, I can handle it in a follow-up.

tests/robustness/failpoint/cluster.go
```diff
@@ -174,7 +171,10 @@ func (f memberDowngrade) Inject(ctx context.Context, t *testing.T, lg *zap.Logge
 	time.Sleep(etcdserver.HealthInterval)
 	e2e.DowngradeEnable(t, clus, lastVersion)
 
-	err = e2e.DowngradeUpgradeMembers(t, lg, clus, numberOfMembersToDowngrade, currentVersion, lastVersion)
+	err = e2e.DowngradeUpgradeMembers(t, lg, clus, numberOfMembersToDowngrade, e2e.BinPath.EtcdLastRelease, version.Versions{
```
siyuanfoundation, line 174: Should we add `time.Sleep(etcdserver.HealthInterval)` before starting the process since the assertion is removed?

fuweid: Do you mean this assertion `AssertProcessLogs(t, epc.Procs[i], "The server is ready to downgrade")`?

We already have code to check cluster/binary/storage versions. There is 1 minute as timeout. I think it's good enough.

```go

ValidateVersion(t, epc.Cfg, epc.Procs[i], version.Versions{
			Cluster: ver.String(),
			Server:  OffsetMinor(ver, 1).String(),
			Storage: ver.String(),
		})
```

So I think we don't need to add that  `time.Sleep(etcdserver.HealthInterval)`.

ahrtr:
@fuweid should we finish the downgrade query API firstly? Afterwards, then update this test later?

tests/framework/e2e/downgrade.go
```diff
@@ -82,48 +81,55 @@ func DowngradeCancel(t *testing.T, epc *EtcdProcessCluster) {
 	t.Log("Cluster downgrade cancellation is completed")
 }
 
-func DowngradeUpgradeMembers(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, numberOfMembersToChange int, currentVersion, targetVersion *semver.Version) error {
+func DowngradeUpgradeMembers(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, numberOfMembersToChange int, newExecPath string, targetVersion version.Versions) error {
```
ahrtr, line 84: Do we still need to update the signature of this function after introducing the downgrade query API?

fuweid: I was thinking it's better to pass expected version by caller instead of inferring by number of members.
Let me think over it and update it later.

Issue #19417, opened 2025-02-14
Update  to Go 1.24 / 1.23


ivanvc:
### What would you like to be added?

Go 1.24.0 was released the day before yesterday. Therefore, Go 1.22 will be out of support when 1.25.0 is released. We should update the main branches to Go 1.24 and the release branches to 1.23.

However, based on previous experience, adopting a new Go minor could take time. So, first, we must ensure that the main branches work fine with 1.24.0. We should also keep an eye on reported issues in the [golang/go](https://github.com/golang/go/issues?q=is%3Aissue%20state%3Aopen%201.24) repository.

Progress track:

- [ ] [etcd](https://github.com/etcd-io/etcd)
  - [ ] [`main`](https://github.com/etcd-io/etcd/tree/main): go v1.24.0 - #19440
  - [x] ~[`release-3.6`](https://github.com/etcd-io/etcd/tree/release-3.6): go v1.23.6~ - already on 1.23.6
  - [x] [`release-3.5`](https://github.com/etcd-io/etcd/tree/release-3.5): go v1.23.6 - #19430
  - [x] [`release-3.4`](https://github.com/etcd-io/etcd/tree/release-3.4): go v1.23.6 - #19429
  - [x] [`CHANGELOG`](https://github.com/etcd-io/etcd/tree/main/CHANGELOG) - #19431
- [ ] [bbolt](https://github.com/etcd-io/bbolt)
  - [ ] [`main`](https://github.com/etcd-io/bbolt/tree/main): go v1.24.0 - etcd-io/bbolt#907
  - [x] [`release-1.3`](https://github.com/etcd-io/bbolt/tree/release-1.3): v1.23.6 - etcd-io/bbolt#908
  - [x] ~[`release-1.4`](https://github.com/etcd-io/bbolt/tree/release-1.4): v1.23.6~ - already on 1.23.6
- [x] [raft](https://github.com/etcd-io/raft)
  - [x] [`main`](https://github.com/etcd-io/raft/tree/main): go v1.24.0 - etcd-io/raft#268
  - [x] ~[`release-3.6`](https://github.com/etcd-io/raft/tree/release-3.6): go v1.23.6~ - already on 1.23.6
- [x] [`etcd-io/gofail` `main`](https://github.com/etcd-io/gofail/tree/main): go v1.23.6 - etcd-io/gofail#102
[...]

PR #19429, opened 2025-02-17
[3.4] Update golang toolchain to 1.23.6

joshjms:
* Update `.go-version` to 1.23.6
* Update Go version in `go.mod` files to `1.23.0`
* Update toolchain to `1.23.6`

Subtask of #19417


go.mod
```diff
@@ -1,8 +1,8 @@
 module go.etcd.io/etcd
 
-go 1.22
+go 1.23
```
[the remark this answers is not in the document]
joshjms: Got it, I referred to the `go.mod` file in the main branch. Will update. Thanks!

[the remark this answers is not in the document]
joshjms: I see, thanks a lot! I'll update in my PR for `release-3.5` as well.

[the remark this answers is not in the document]
Elbehery: iiuc, we don't use `patch` in go.mod 

only the toolchain 

cc @ivanvc @henrybear327

go.mod
```diff
@@ -1,8 +1,8 @@
 module go.etcd.io/etcd
 
-go 1.22
+go 1.23.0
```
ahrtr, line 3: can we use "1.23"?

Same comment as https://github.com/etcd-io/etcd/pull/19430#discussion_r1957978582

joshjms: Sure, I'll change it. 

If I'm not mistaken, the `go` directive is just there to enforce the minimum version to build, therefore 1.23 and 1.23.0 should be the same.

ahrtr: > therefore 1.23 and 1.23.0 should be the same.

It's true, but "1.23" is preferred, 
- It's more explicit that we don't require/force any applications (which depend on etcd) to use any go patch version.

joshjms: Ah, I'm under the impression that 1.23 = 1.23.x (compatible with any x). So in the case where the minimum version is 1.23, it will take 1.23.0 (as it is the minimum of all 1.23.x). Thanks for the insights.

joshjms: @ahrtr done

ahrtr:
please squash the commit

joshjms:
i squashed the commit

ivanvc:
@ahrtr / @jmhbnz, can you manually approve the workflows? Neither `/ok-to-test` nor closing and reopening triggered the tests. Thanks.

ahrtr (approved):
LGTM

Thank you!

ivanvc (approved):
LGTM. Thanks, @joshjms.

PR #19440, opened 2025-02-18
Bump go toolchain to 1.24.1

ivanvc:
Bumps Go to 1.24.

Part of #19417 


ivanvc:
The `usetesting` linter shows a lot of violations. I addressed them, but this pull request would be huge with all the changes:

```
146 files changed, 1157 insertions(+), 1212 deletions(-)
```

The problem is that it's suggesting to use functions that are not in Go 1.23. Therefore, I will disable the linter so we can merge this. And I will open multiple pull requests targeting main. Once I finish with them, I'll reenable the linter.

server/embed/config_test.go
```diff
@@ -843,7 +843,7 @@ func TestLogRotation(t *testing.T) {
 			logOutputs:        []string{"/tmp/path"},
 			logRotationConfig: `{"maxsize": true}`,
 			wantErr:           true,
-			wantErrMsg:        errors.New("invalid log rotation config: json: cannot unmarshal bool into Go struct field logRotationConfig.maxsize of type int"),
+			wantErrMsg:        errors.New("invalid log rotation config: json: cannot unmarshal bool into Go struct field logRotationConfig.Logger.maxsize of type int"),
```
ahrtr, line 846: question: is this a side effect to bump the go version?

tools/rw-heatmaps/rw-benchmark.sh
```diff
@@ -170,6 +170,8 @@ for RATIO_STR in ${RATIO_LIST}; do
       CURRENT_ETCD_PID=$!
       sleep 5
 
+      "${ETCD_BIN_DIR}"/resource_monitor.sh --pid "${CURRENT_ETCD_PID}" --output-file "${OUTPUT_FILE/%.csv/}-resources-ratio-${RATIO}-conn-${CONN_CLI_COUNT}-value-${VALUE_SIZE}.csv" &
```
ahrtr, line 173: question: is this reusing https://github.com/ahrtr/resource-monitor/blob/main/resource_monitor.sh directly? If yes, does it mean that we need to manually that it's present under the /bin directory?

ivanvc: My bad, this change shouldn't be here. I'll remove it from the commit.

server/embed/config_test.go
ivanvc: Yes, the error message is different now, it also shows the embedded structure.

tools/rw-heatmaps/rw-benchmark.sh
ivanvc: When rebasing I added a commit that didn't belong to this branch. I removed it.

server/embed/config_test.go
ivanvc: I added more context in 31650ab0c8df43af05fc4c13b48ffee59271eec7's commit message:

>     The embedded structure is now also part of the log message. The failure
>     from the tests is the following:
>    
>     --- FAIL: TestLogRotation (0.00s)
>         --- FAIL: TestLogRotation/invalid_logger_config (0.00s)
>             config_test.go:868: test "invalid logger config", expected
>             error: invalid log rotation config: json: cannot unmarshal
>                    bool into Go struct field logRotationConfig.maxsize
>                    of type int,
>             got: invalid log rotation config: json: cannot unmarshal
>                  bool into Go struct field logRotationConfig.Logger.maxsize
>                  of type int
>     FAIL

ahrtr (approved):

PR #19432, opened 2025-02-17
perf: use RLock in Demoted method for read-only access to expiry

1911860538:
Use RLock in Demoted method to improve read-only access to expiry with better concurrency.

PR #19445, opened 2025-02-18
[3.5]: use RLock in Demoted method for read-only access to expiry

1911860538:
#19432

### DIFF
--- a/server/lease/lessor.go
+++ b/server/lease/lessor.go
@@ -26,11 +26,12 @@ import (
 	"time"
 
 	"github.com/coreos/go-semver/semver"
+	"go.uber.org/zap"
+
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/server/v3/lease/leasepb"
 	"go.etcd.io/etcd/server/v3/mvcc/backend"
 	"go.etcd.io/etcd/server/v3/mvcc/buckets"
-	"go.uber.org/zap"
 )
 
 // NoLease is a special LeaseID representing the absence of a lease.
@@ -912,8 +913,8 @@ func (l *Lease) forever() {
 
 // Demoted returns true if the lease's expiry has been reset to forever.
 func (l *Lease) Demoted() bool {
-	l.expiryMu.Lock()
-	defer l.expiryMu.Unlock()
+	l.expiryMu.RLock()
+	defer l.expiryMu.RUnlock()
 	return l.expiry == forever
 }
 


ahrtr (approved):

1911860538:
> Please don't backport performance changes, etcd v3.6 is soon to be released. The argument that we need to backport to make it available should no longer be relevant. We should focus on releasing v3.6 and its adoption.

@ahrtr @serathius Acknowledged, I am going to close this pull request.

ahrtr:
> Please don't backport performance changes, etcd v3.6 is soon to be released. The argument that we need to backport to make it available should no longer be relevant. We should focus on releasing v3.6 and its adoption.


This is not a performance fix but a correction of a misuse of sync.RWMutex, which is why backporting makes sense. That said, it’s not a critical issue, and I feel the concern here is somewhat overstated.

While 3.6 is the priority, that doesn’t mean we can’t make small, low-effort fixes along the way.

ahrtr (approved):
cc @fuweid @ivanvc @jmhbnz

fuweid (approved):
LGTM

serathius:
> This is not a performance fix but a correction of a misuse of sync.RWMutex, which is why backporting makes sense. That said, it’s not a critical issue, and I feel the concern here is somewhat overstated.

Switching from Lock to RLock, doesn't fix any concurrency issue, just improves performance.

ahrtr:
> > This is not a performance fix but a correction of a misuse of sync.RWMutex, which is why backporting makes sense. That said, it’s not a critical issue, and I feel the concern here is somewhat overstated.
> 
> Switching from Lock to RLock, doesn't fix any concurrency issue, just improves performance.

It's unfortunate that we're spending time debating such a trivial fix.

fuweid:
This change is valid since RLock is read-only check. is there any concern for this change?

ahrtr:
Let's do not waste more time on this. Let's see what other active maintainers/reviewers think. @ivanvc @jmhbnz 

If majority approve, then just merge it; otherwise, just close it. Please respond on both https://github.com/etcd-io/etcd/pull/19443 and this PR. Thanks.

ivanvc:
/retitle [3.5]: use RLock in Demoted method for read-only access to expiry

ivanvc:
I'm on the fence on this one. I agree with both parties. It looks like it's not a bug fix, so that's why it shouldn't be backported. But, if it's a misuse, it makes sense to backport.

I lean towards a no, but that will make this a tie. Then, we'll need to hear from @jmhbnz.

jmhbnz (approved):
This is a straightforward minor fix, given we're so close to `v3.6.0` I would lean towards just releasing it there as is already planned.  

With that said I don't have any personal objections to it being backported to `release-3.5` as it is a valid fix, and the work has already been done to raise it thanks to @1911860538

ivanvc:
/retitle [release-3.5] use RLock in Demoted method for read-only access to expiry

PR #19450, opened 2025-02-19
Simplify release permissions process

jmhbnz:
As discussed in last two release meetings this pull request simplifies our release permissions process by moving away from manual changes to branch restrictions and instead making the permissions based on membership of the release GitHub team.

Historically for every release we would:
1) Raise a `k/org` pr in advance to give release team elevated permissions by changing the permission levels for groups.
2) Once merged, manually edit our branch restrictions in GitHub so that the release team member running the release script could push the release commit directly to the stable release branch without a pull request.
3) After release raise a `k/org` pr to revoke release team elevated permissions by changing the permission levels for groups.

In future we will instead:
1) Raise a `k/org` pr in advance of each release to add release team members to the GitHub team
2) After release raise a `k/org` pr to revoke release team elevated permissions by removing members from release group.

By adding @etcd-io/release-etcd as an entity that can bypass the stable release branch restrictions and just managing the membership of that group we have a simpler, less manual way of carrying out future releases.

cc @ahrtr, @serathius, @ivanvc

### DIFF
--- a/Documentation/contributor-guide/release.md
+++ b/Documentation/contributor-guide/release.md
@@ -15,6 +15,7 @@ for ensuring the stability of the release branch.
 - Marek Siarkowicz [@serathius](https://github.com/serathius)
 - Sahdev Zala [@spzala](https://github.com/spzala)
 - Wenjia Zhang [@wenjiaswe](https://github.com/wenjiaswe)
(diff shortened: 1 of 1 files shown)


ahrtr (approved):

ivanvc (approved):
LGTM. Thanks, @jmhbnz.

One less manual step :tada:

PR #19451, opened 2025-02-19
*: support DowngradeInfo field in maintenence.Status API

fuweid:
This patch has three updates:

### 1. maintence.Status API

Add `DowngradeInfo` in the response. We could consider backport this to v3.5 release.

### 2. Add two columns in `etcdctl endpoint status` output

```bash
bin/etcdctl  endpoint status -w table
+----------------+------------------+---------------+-----------------+---------+--------+-----------------------+-------+-----------+------------+-----------+------------+--------------------+--------+--------------------------+-------------------+
|    ENDPOINT    |        ID        |    VERSION    | STORAGE VERSION | DB SIZE | IN USE | PERCENTAGE NOT IN USE | QUOTA | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS | DOWNGRADE TARGET VERSION | DOWNGRADE ENABLED |
+----------------+------------------+---------------+-----------------+---------+--------+-----------------------+-------+-----------+------------+-----------+------------+--------------------+--------+--------------------------+-------------------+
| 127.0.0.1:2379 | 8e9e05c52164694d | 3.6.0-alpha.0 |           3.5.0 |   20 kB |  16 kB |                   20% |   0 B |      true |      false |         2 |          6 |                  6 |        |                    3.5.0 |              true |
+----------------+------------------+---------------+-----------------+---------+--------+-----------------------+-------+-----------+------------+-----------+------------+--------------------+--------+--------------------------+-------------------+
```

#### 1. Update DowngradeUpgradeMembersByID Interface

If it is a downgrade process, the desired cluster version should be the target version.
If it is an upgrade process, the desired cluster version should be determined by the minimum binary version of the members.

#### 2. Remove AssertProcessLogs from DowngradeEnable

The log message "The server is ready to downgrade" appears only when the storage
version monitor detects a mismatch between the cluster and storage versions.

If traffic is insufficient to trigger a commit or if an auto-commit occurs right
after reading the storage version, the monitor may fail to update it, leading
to errors like:

```bash
"msg":"failed to update storage version","cluster-version":"3.6.0",
"error":"cannot detect storage schema version: missing confstate information"
```

Given this, we should remove the AssertProcessLogs statement.

Similar to https://github.com/etcd-io/etcd/pull/19313


fuweid:
ping @ahrtr @siyuanfoundation @serathius

ahrtr:
Thanks @fuweid for the quick PR.

Two comments:
- If we extend the response of the existing `etcdctl endpoint status`, then we don't need to add a separate command `etcdctl downgrade status`;
- Could you add or update e2e test cases?
  - By default, when there is no downgrade enabled, the "downgrade target version" should be empty, and "downgrade enabled" should be false. We can just update the existing e2e test case.
  - Partially downgraded. I think we can just update the existing e2e test case.
  - Completed downgraded. We can use @henrybear327 's PR https://github.com/etcd-io/etcd/pull/19399

fuweid:
> Partially downgraded. I think we can just update the existing e2e test case.

For this change, I need the interface change in https://github.com/etcd-io/etcd/pull/19427. If there is no objection, I can merge https://github.com/etcd-io/etcd/pull/19427 into this pull request.

ahrtr:
> For this change, I need the interface change in #19427. If there is no objection, I can merge #19427 into this pull request.

OK, please feel free to add the test case changes into this PR.

ahrtr:
@ivanvc @jmhbnz  @siyuanfoundation @serathius PTAL, thx

This PR looks huge, but actually most of the code is auto generated. It's pretty simple & straightforward. High level looks good to me. Will take a closer look tomorrow my time.

etcdctl/ctlv3/command/printer_fields.go
```diff
@@ -203,6 +203,8 @@ func (p *fieldsPrinter) EndpointStatus(eps []epStatus) {
 		fmt.Println(`"RaftAppliedIndex" :`, ep.Resp.RaftAppliedIndex)
 		fmt.Println(`"Errors" :`, ep.Resp.Errors)
 		fmt.Printf("\"Endpoint\" : %q\n", ep.Ep)
+		fmt.Printf("\"DowngradeTargetVersion\" : %q\n", ep.Resp.DowngradeInfo.GetTargetVersion())
+		fmt.Println(`"DowngradeEnable" :`, ep.Resp.DowngradeInfo.GetEnabled())
```
ahrtr, line 207:
```suggestion
		fmt.Println(`"DowngradeEnabled" :`, ep.Resp.DowngradeInfo.GetEnabled())
```

tests/e2e/cluster_downgrade_test.go
```diff
@@ -150,6 +168,21 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
...
+		mc := member.Etcdctl()
+		mName := member.Config().Name
+
+		statuses, serr := mc.Status(context.Background())
+		require.NoErrorf(t, serr, "member %s", mName)
+		require.Lenf(t, statuses, 1, "member %s", mName)
+		st := (*pb.StatusResponse)(statuses[0])
+		require.Truef(t, st.GetDowngradeInfo().GetEnabled(), "member %s", mName)
+		require.Equalf(t, lastClusterVersion.String(), st.GetDowngradeInfo().GetTargetVersion(), "member %s", mName)
+	}
```
ahrtr, line 184: This might be flaky. The status API isn't guarded by linearizable check. So let's issue a read request after calling `DowngradeEnable` and before checking the downgrade status?

tests/framework/e2e/etcd_process.go
```diff
@@ -329,6 +331,27 @@ func AssertProcessLogs(t *testing.T, ep EtcdProcess, expectLog string) {
 	}
 }
 
+func GetLeader(t *testing.T, epc *EtcdProcessCluster) EtcdProcess {
```
ahrtr, line 334: I recall that we have an existing function with name like `WithLeader`?

tests/robustness/failpoint/cluster.go
```diff
@@ -228,13 +229,32 @@ func (f memberDowngradeUpgrade) Inject(ctx context.Context, t *testing.T, lg *za
...
+	})
 	if err != nil {
 		return nil, err
 	}
+
+	// NOTE: By default, the leader can cancel the downgrade once all members
+	// have reached the target version. However, determining the final stable
+	// cluster version after an upgrade can be challenging. To ensure stability,
+	// we should wait for leader to cancel downgrade process.
+	e2e.AssertProcessLogs(t, e2e.GetLeader(t, clus), "the cluster has been downgraded")
```
ahrtr, line 244: can we check the downgrade status as well?

fuweid: v3.5 doesn't have this API change so it's unlikely to check status here.

ahrtr: OK, got it.

tests/framework/e2e/etcd_process.go
fuweid: correct. I updated to use `epc.WaitLeader()`.

etcdctl/ctlv3/command/printer_fields.go
fuweid: Updated.

tests/e2e/cluster_downgrade_test.go
fuweid: Good point. I updated to use loop for check instead of one-time check.

etcdctl/ctlv3/command/printer.go
```diff
@@ -238,6 +238,8 @@ func makeEndpointStatusTable(statusList []epStatus) (hdr []string, rows [][]stri
 			fmt.Sprint(status.Resp.RaftIndex),
 			fmt.Sprint(status.Resp.RaftAppliedIndex),
 			fmt.Sprint(strings.Join(status.Resp.Errors, ", ")),
+			fmt.Sprint(status.Resp.DowngradeInfo.GetTargetVersion()),
+			fmt.Sprintf("%v", status.Resp.DowngradeInfo.GetEnabled()),
```
ivanvc, line 242: I know this probably follows the previous lines. But I don't think we need to do a `fmt.Sprintf(...)` on the string values. Or is there a reason to do that? (Sorry, I'm not very familiar with `etcdctl`).

I mean:

```suggestion
			status.Resp.DowngradeInfo.GetTargetVersion(),
			strconv.FormatBool(status.Resp.DowngradeInfo.GetEnabled()),
```

(The bool conversion raises a [`perfsprint`](https://github.com/catenacyber/perfsprint) linter warning).

ahrtr: I agree that `strconv.FormatBool` is better, because obvious it has a little performance. I think it's OK to update it separately,
- It isn't critical
- To keep consistent with other bool (i.e. `fmt.Sprint(status.Resp.IsLearner)`) , you can just use `fmt.Sprint(status.Resp.DowngradeInfo.GetEnabled())`.

tests/framework/e2e/downgrade.go
```diff
@@ -82,48 +82,100 @@ func DowngradeCancel(t *testing.T, epc *EtcdProcessCluster) {
...
+func DowngradeUpgradeMembers(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, numberOfMembersToChange int, newExecPath string, targetVersion version.Versions) error {
 	membersToChange := rand.Perm(len(clus.Procs))[:numberOfMembersToChange]
 	t.Logf("Elect members for operations on members: %v", membersToChange)
 
-	return DowngradeUpgradeMembersByID(t, lg, clus, membersToChange, currentVersion, targetVersion)
+	return DowngradeUpgradeMembersByID(t, lg, clus, membersToChange, newExecPath, targetVersion)
 }
 
-func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, membersToChange []int, currentVersion, targetVersion *semver.Version) error {
+func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcessCluster, membersToChange []int, newExecPath string, targetVersion version.Versions) error {
```
ahrtr, line 137: I feel like the existing implementation of this function is simpler (so better). i.e. You don't need to check upgrading or downgrading for each member.

fuweid: I added some comments about this change. The major reason is that we can't imply desire version based on target version. Even if we have DowngradeInfo from member status, the caller still needs to ensure what the downgrade info is. If we want to keep the original interface, and there is any flaky happening, it's time consuming task to analyze why it doesn't get into desire state. IMO, original interface is trying to maintain logic of downgrade, similar to how the server does. The test case should be simple as performing one action and expecting the result. I'm not object to revert it. Just share my thoughts.

ahrtr: Let's have a quick call. I will be available around 30 minutes later.

etcdctl/ctlv3/command/printer.go
fuweid: @ivanvc thanks! Updated

tests/framework/e2e/downgrade.go
fuweid: Updated. Please take a look.

tests/framework/e2e/downgrade.go
```diff
@@ -117,11 +161,16 @@ func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcess
 			return err
 		}
 	}
+
+	clusterVersion := targetVersion.String()
+	if !isDowngrade && len(membersToChange) != len(clus.Procs) {
+		clusterVersion = currentVersion.String()
+	}
```
ahrtr, line 168: Can we revert this change as well to make this PR as simple as possible? It unnecessarily complicates the test case. 

When upgrading, the current test case ensures that all members will be on the target version.

```suggestion
```

tests/framework/e2e/downgrade.go
```diff
@@ -117,11 +161,16 @@ func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcess
...
 	lg.Info("Validating versions")
 	for _, memberID := range membersToChange {
 		member := clus.Procs[memberID]
 		ValidateVersion(t, clus.Cfg, member, version.Versions{
-			Cluster: targetVersion.String(),
+			Cluster: clusterVersion,
```
ahrtr, line 173: revert this change as well.
```suggestion
			Cluster: targetVersion.String(),
```

tests/framework/e2e/downgrade.go
fuweid: But it's bug in this function....

The robustness test failure is caused by this.
If we remove this, the robustness test always fails when the failpoint is related to downgrade.

tests/e2e/cluster_downgrade_test.go
```diff
@@ -165,7 +177,7 @@ func testDowngradeUpgrade(t *testing.T, numberOfMembersToDowngrade int, clusterS
 	err = e2e.DowngradeUpgradeMembersByID(t, nil, epc, membersToChange, currentVersion, lastClusterVersion)
 	require.NoError(t, err)
 	if len(membersToChange) == len(epc.Procs) {
-		e2e.AssertProcessLogs(t, leader(t, epc), "the cluster has been downgraded")
+		e2e.AssertProcessLogs(t, epc.Procs[epc.WaitLeader(t)], "the cluster has been downgraded")
```
ahrtr, line 180: Can we also call `e2e.ValidateDowngradeInfo` to check the downgrade info, to ensure that the downgrade can be automatically cancelled?

fuweid: release 3.5 doesn't have this change yet. so we can't do that

ahrtr: If this is resolved, then we might not need https://github.com/etcd-io/etcd/pull/19399 any more

ahrtr: > then we might not need #19399 any more

It seems that https://github.com/etcd-io/etcd/pull/19399 doesn't make any sense?

fuweid: I think so.

tests/framework/e2e/downgrade.go
ahrtr: > When upgrading, the current test case ensures that all members will be on the target version.

This isn't true in robustness test?

tests/e2e/cluster_downgrade_test.go
ahrtr: FYI. @henrybear327 then let's close https://github.com/etcd-io/etcd/pull/19399

ahrtr (approved):
LGTM

Thanks for the quick PR.

cc @ivanvc @serathius @siyuanfoundation PTAL

tests/e2e/cluster_downgrade_test.go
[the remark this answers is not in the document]
fuweid: We can't replace log check here because v3.5 doesn't have this api change. 

So you mean, add ValidateDowngradeInfo check in the end of test?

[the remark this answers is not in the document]
siyuanfoundation: > We can't replace log check here because v3.5 doesn't have this api change.
> 
> So you mean, add ValidateDowngradeInfo check in the end of test?

oh, right. 
Just ValidateDowngradeInfo check in the end of test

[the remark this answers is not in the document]
fuweid: Updated @siyuanfoundation  PTAL when you have time. Thanks

ahrtr (approved):

Issue #19453, opened 2025-02-20
Makefile: Not choosing `SHELL` in reliable way


kavirajk:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

When I wanted to run `robustness` tests, [I followed the README](https://github.com/etcd-io/etcd/tree/8866fce73f3cb79451d3db2d94c5b3a92103e0be/tests/robustness#running-locally) and it says three steps to build etcd with failpoints.

```bash
    make gofail-enable
    make build
    make gofail-disable
```

But running `make gofail-enable` failed with following error.

```
-bash-5.2$ make gofail-enable
/bin/sh: line 0: cd: tools/mod: No such file or directory
go install go.etcd.io/gofail@
go: go.etcd.io/gofail@: version must not be empty
make: *** [/Users/kavirajkanagaraj/bin/gofail] Error 1
```

### What did you expect to happen?

I expect `make gofail-enable` to work correctly to run robustness tests.

### How can we reproduce it (as minimally and precisely as possible)?

I found this error coming [from this line ](https://github.com/etcd-io/etcd/blob/8866fce/tests/robustness/makefile.mk#L65)when trying to get `GOFAIL_VERSION`.

The underlying issue seems to be "type of SHELL" `make` command chooses to run. Example

If I run that same line from repo root using my default shell (bash), it works.
```
-bash-5.2$ pwd
/Users/kavirajkanagaraj/src/etcd
-bash-5.2$ cd tools/mod && go list -m -f {{.Version}} go.etcd.io/gofail
v0.2.0
-bash-5.2$
```

But if run same with `sh` shell, it fails with exact error that makefile failed with in the first place.
```
-bash-5.2$ pwd
/Users/kavirajkanagaraj/src/etcd
-bash-5.2$ sh -c "cd tools/mod && go list -m -f {{.Version}} go.etcd.io/gofail"
[...]

PR #19454, opened 2025-02-20
makefile: Make `bash` as default shell for all make targets

kavirajk:
fixes: https://github.com/etcd-io/etcd/issues/19453


kavirajk:
closing as per discussion[ here.](https://github.com/etcd-io/etcd/issues/19453#issuecomment-2673954456)

Issue #18902, opened 2024-11-16
Generate modern sbom formats when releasing etcd


idunbarh:
### What would you like to be added?

I'd like to contribute SBOM generation to the release process of this project in both cyclonedx and spdx formats.

I'm part of https://github.com/CISA-SBOM-Community/SBOM-Generation thats building reference implementations for "good" SBOM generation and we thought etcd would be a great candidate.

### Why is this needed?

SBOMs are becoming a common part of software releases because they provide insight into what dependencies are used in a project. This allows better vulnerability management.

jmhbnz:
Hi @idunbarh many thanks for raising this and volunteering. For background we last touched the SBOM process for etcd in https://github.com/etcd-io/etcd/issues/15747 and the process essentially relies on a very antiquated approach to generate the json file https://github.com/etcd-io/etcd/blob/main/bill-of-materials.json.

Adopting new cyclonedx and spdx format support is definitely something I would support.  We would need to still retain the old json format sbom process in the stable release branches until the next minor release at least in order to not break any existing processes but I am very keen to forge ahead implementing this in `main` initially.

cc @ivanvc, @ahrtr, @serathius  for any further thoughts on this.

/assign @idunbarh

jmhbnz:
/retitle Generation modern sbom formats when releasing etcd

ivanvc:
I think this is a great idea. I spoke with @puerco (sig release) regarding this at KubeCon.

ahrtr:
Thanks for raising this discussion. The json format of [SPDX](https://spdx.dev/) seems a better choice to me. The only minor concern is that [SBOM-Generation](https://github.com/CISA-SBOM-Community/SBOM-Generation) is implemented with Python. It means that we have to get python installed in contributors' dev environment to update SBOM.

Alternatives:
- https://github.com/anchore/syft
- generate a tool based on https://github.com/spdx/tools-golang

ahrtr:
> I'll use syft for this implementation.

Do you mean https://github.com/anchore/syft? If yes, then looks good to me.

jmhbnz:
Discussed during sig-etcd triage meeting. Checking in, @idunbarh are you still keen to work on this?

jmhbnz:
Discussed during sig-etcd triage meeting, opening this up for other contributors given the time passed and adding help wanted.

vivekpatani:
@jmhbnz can be assigned to me.

jmhbnz:
[...]

PR #19467, opened 2025-02-22
scripts/updatebom: remove unnecessary mod override

ivanvc:
The `GOFLAGS=-mod=mod` environment variable is not required, as that's the default mode. This code was likely copied from the tests, where `GOFLAGS` is set to `-mod=readonly`. So, executing the tool needs to override `mod`, as per the documentation, it tends to modify the `go.mod` file.

Spun off from: #19423.

### DIFF
--- a/scripts/updatebom.sh
+++ b/scripts/updatebom.sh
@@ -14,7 +14,7 @@ function bom_fixlet {
   # shellcheck disable=SC2207
   modules=($(modules_for_bom))
 
-  if GOFLAGS=-mod=mod run_go_tool "github.com/appscodelabs/license-bill-of-materials" \
+  if run_go_tool "github.com/appscodelabs/license-bill-of-materials" \
       --override-file ./bill-of-materials.override.json \
       "${modules[@]}" > ./bill-of-materials.json.tmp; then
     cp ./bill-of-materials.json.tmp ./bill-of-materials.json


serathius (approved):

ahrtr:
please rebase this PR and mark it ready to review, thx

ivanvc:
There are a couple of obscure failures in the BOM generation. Finding a way to reproduce it was difficult, but I found an issue running this tool in a clean environment. Running the following command from the main branch, makes the bom generation fail:

```
docker run --rm -w /workspace -v $PWD:/workspace --entrypoint /bin/bash gcr.io/k8s-staging-test-infra/kubekins-e2e:v20250212-16f67660c2-master -c 'git config --global --add safe.directory /workspace && make verify-bom'
```

This simulates running `verify-bom` (and `scripts/updatebom.sh`) in a clean environment. The workaround I used when implementing the Go workspace was to run `go mod download` across all the submodules to avoid downloading dependencies when running the check.

I opened and immediately drafted this pull request because there's another obscure error that I haven't been able to reproduce in the `verify` job: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19467/pull-etcd-verify/1893448496992751616.

I believe the error here may be something similar, but haven't had the chance to verify it.

In theory, removing `GOFLAGS=-mod=mod` shouldn't affect, but it's making the script fail.

IMHO, ideally, we should stop building the BOM with this tool and switch to a modern SBOM generation tool (#18902).

PR #19469, opened 2025-02-23
etcdctl: fix json output

### LINKED ISSUE #19262: Enhance  `etcdctl member` commands to support printing clusterID and memberID in hex
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).

### What happened?

I ran `etcdctl member add -w json --hex=true`, with the aim of extracting the newly added member ID for use with `etcdctl member promote later`. However, the member ID was printed as decimal:

```
> etcdctl member add test-node --peer-urls=https://10.16.36.111:2380 --learner=true -w json --hex=true
{"header":{"cluster_id":8614491861582198852,"member_id":9202843606982955899,"raft_term":2},"member":{"ID":12981492951526344977,"peerURLs":["https://10.16.36.111:2380"],"isLearner":true},"members":[{"ID":677637297205855628,"name":"etcd-i-0287724fa565aba47","peerURLs":["https://10.16.43.142:2380"],"clientURLs":["https://10.16.43.142:2379"]},{"ID":9202843606982955899,"name":"etcd-i-0a11bcfc6be104246","peerURLs":["https://10.16.33.50:2380"],"clientURLs":["https://10.16.33.50:2379"]},{"ID":12981492951526344977,"peerURLs":["https://10.16.36.111:2380"],"isLearner":true},{"ID":16455373321881049425,"name":"etcd-i-0e6aa2e540fd3c248","peerURLs":["https://10.16.36.53:2380"],"clientURLs":["https://10.16.36.53:2379"]}]
```

This does not match the behaviour of `etcdctl member list -w json --hex=true`, which prints the member IDs in hexadecimal, compatible for use with other `etcdctl` commands.

I expect `etcdctl member add ... -w json --hex=true` to print the IDs in hexadecimal, just like `etcdctl member list` does.

Run the following command (the new member doesn't actually ever have to join the cluster, it can be immediately deleted afterwards):

```bash
etcdctl member add test-node --peer-urls=https://10.16.36.111:2380 --learner=true -w json --hex=true
```

_No response_

<details>

```console
$ etcd --version
etcd Version: 3.5.15
Git SHA: 9a5533382
Go Version: go1.21.12
Go OS/Arch: linux/arm64

$ etcdctl version
etcdctl version: 3.5.15
API version: 3.5
```

</details>

<details>

</details>

<details>

```console
# paste output here

# paste output here
```

</details>

### Relevant log output

```Shell

```

## Summary
- `etcdctl member add`: https://github.com/etcd-io/etcd/pull/19469
- `etcdctl member list`:  https://github.com/etcd-io/etcd/pull/20264

kstrifonoff:
Fix #19262 

Example output of `etcdctl member add test-node --learner=true --peer-urls=http://127.0.0.1:22380 --write-out json --hex` command:
### Before:
```json
{
  "header": {
    "cluster_id": 14841639068965178418,
    "member_id": 10276657743932975437,
    "raft_term": 2
  },
  "member": {
    "ID": 17436648322565278587,
    "peerURLs": ["http://127.0.0.1:22380"],
    "isLearner": true
  },
  "members": [
    {
      "ID": 10276657743932975437,
      "name": "default",
      "peerURLs": ["http://localhost:2380"],
      "clientURLs": ["http://localhost:2379"]
    },
    {
      "ID": 17436648322565278587,
      "peerURLs": ["http://127.0.0.1:22380"],
      "isLearner": true
    }
  ]
}
```
### After
```json
{
  "header": {
    "cluster_id": "cdf818194e3a8c32",
    "member_id": "8e9e05c52164694d",
    "raft_term": 2
  },
  "member": {
    "ID": "5bdeb7bdb679126b",
    "peerURLs": ["http://127.0.0.1:22380"],
    "isLearner": true
  },
  "members": [
    {
      "ID": "5bdeb7bdb679126b",
      "peerURLs": ["http://127.0.0.1:22380"],
      "isLearner": true
    },
    {
      "ID": "8e9e05c52164694d",
      "name": "default",
      "peerURLs": ["http://localhost:2380"],
      "clientURLs": ["http://localhost:2379"]
    }
  ]
}
```


ivanvc:
Hi @kstrifonoff, thanks for the pull request. I'm respectful to the original contributor who asked to work on this. Our guideline is to re-assign after a week of no answer when asking a contributor for updates.

A couple of suggestions by quickly looking at the code.

1. The imports contain a lint error. Running `make verify-lint` locally will show you the error.
2. The code doesn't have unit tests for this, but there's an e2e test, `TestCtlV3MemberListWithHex`. I think the test should pass, but adding your new cases would be good.
3. I suggest reviewing the code from `printKV`. I think it would be better to follow that implementation approach, i.e., there would be less repetition with the if conditions.

Thanks again for your pull request :)

kstrifonoff:
Hi @ivanvc , 

I’ve made some changes to my initial suggestions based on your feedback. 

Now, the “JSON with Hex” implementation is just a slight extension of the default “printJSON” implementation. I think it should probably be moved into the “utils” package. 

Here are a few things I noticed: 

1. The implementation assumes marshalling back and forth the default JSON string, which can be inefficient, but I assume it’s still okay for the CLI. 
2. Also, the current approach makes it easy to add “json with hex” output for all the rest of commands that support JSON output. 
3. I’m not sure what kind of end-to-end test should be done here. What should it cover? It’s also not clear to me what the current end-to-end test actually checks.

ivanvc:
> I’m not sure what kind of end-to-end test should be done here. What should it cover? It’s also not clear to me what the current end-to-end test actually checks.

The e2e tests already exist, my suggestion is to expand them to cover the new cases you added: https://github.com/etcd-io/etcd/blob/9f1709e015640838269bfbecae9ee2be41b47939/tests/e2e/ctl_v3_member_test.go#L206

kstrifonoff:
hi @ivanvc 
Could you give it another look? I’ve made a few changes since we last spoke. The initial solution had a bug, so I’ve fixed it.

I’ve also fixed all the remaining `member` and existing `endpoint` commands. Now, you can also use them with `json` and `hex` output.
I’ve also extended the e2e tests as suggested. 

Let me know if you have any questions!

kstrifonoff:
@ivanvc just a friendly reminder: could you please review this once you have some time?

kstrifonoff:
/cc @fuweid 

Could you please review this PR?

etcdctl/ctlv3/command/printer_json.go
```diff
@@ -36,65 +34,187 @@ func newJSONPrinter(isHex bool) printer {
 	}
 }
 
-func (p *jsonPrinter) EndpointHealth(r []epHealth) { printJSON(r) }
-func (p *jsonPrinter) EndpointStatus(r []epStatus) { printJSON(r) }
-func (p *jsonPrinter) EndpointHashKV(r []epHashKV) { printJSON(r) }
+type (
+	HexStatusResponse pb.StatusResponse
```
fuweid, line 38: Thanks for filing this refactor change.
Would you please add one method `MemberAdd` to `jsonPrinter` to fix that issue first?
Basically, we don't involve refactor change in bugfix. It's easy to review and maintain.
Thanks!

kstrifonoff: @fuweid it makes sense. I’ve removed the extra code. Could you please check it again?

If it looks good, let’s talk about how to refactor the rest of the code.

ahrtr:
thx for the contribution. But it seems that you introduced an inconsistent way as one existing implementation,

https://github.com/etcd-io/etcd/blob/b3b9e986e037727a53ea08e5d7f8672cd5397fc7/etcdctl/ctlv3/command/printer_json.go#L43-L60

Also this is a generic problem, other commands have similar problem, but you only fixed `MemberAddResponse`. 

It'd be good that we have a overall plan on how to resolve the json output problem for all commands in a consistent way.

ahrtr:
thx for the clarification. 

Breaking down changes into smaller PRs is good. Please clarify your plan in https://github.com/etcd-io/etcd/issues/19262 ( I just reopened it) and get approval first (please bring it to our community meeting. Note: I am supportive on it). Afterwards,  then resolve other commands one by one per the plan.

ahrtr:
> Another comment is that there is no test. Could you also add a unit test for each command you updated?

Also please provide example outputs before and after the change for reference/comparison in each PR (including this one).

etcdctl/ctlv3/command/printer_json.go
```diff
@@ -47,6 +79,7 @@ func (p *jsonPrinter) MemberList(r clientv3.MemberListResponse) {
 		printJSON(r)
 	}
 }
+func (p *jsonPrinter) MemberAdd(r clientv3.MemberAddResponse) { p.printJSON(r) }
```
ahrtr, line 85: I think we can construct a fake `MemberAddResponse` object, and call this method directly to ensure that it generates the expected output with and without `--hex`.

We also need to make similar change for some other commands. So let's make this PR as a good example for following PRs to follow.

ahrtr:
Overall looks good with a comment (adding unit test) https://github.com/etcd-io/etcd/pull/19469#discussion_r2144585903

Please also list all the commands which we need to make similar change in the issue, so that we don't miss any one of them.

etcdctl/ctlv3/command/printer_json.go
kstrifonoff: hi @ahrtr 
just to confirm: this comment is your suggestion how could I implement the unit test; you are good with current implementation and you don't want me to change it. is this correct?

ahrtr: It would be better to implement the unit test in this PR, so that any following PRs can just follow this PR as mentioned above.

You can add a field nested `io.Writer` into `jsonPrinter`, and set it to `os.Stdout` by default. In your unit test, you can change it to `bytes.Buffer`, so that it's easier to verify the output.

kstrifonoff: I understand your point.

However, it also requires modifying the rest of the code to explicitly use that writer instead of `os.Stdout`.

For instance, the signature of `printJSON` method should be changed, therefore I'll also need to change other lines of code where it is used:
```go
func (p *jsonPrinter) EndpointHealth(r []epHealth) { printJSON(r) }
func printJSON(v any) {
	b, err := json.Marshal(v)
	if err != nil {
		fmt.Fprintf(os.Stderr, "%v\n", err)
		return
	}
	fmt.Println(string(b))
}
```
->
```go
func (p *jsonPrinter) EndpointHealth(r []epHealth) { printJSON(p.w, r) }
func printJSON(w io.Writer, v any) {
	b, err := json.Marshal(v)
	if err != nil {
		fmt.Fprintf(os.Stderr, "%v\n", err)
		return
	}
	fmt.Fprintln(w, string(b))
}
```


Should I create a kind of "pre-requisite" PR where I’ll simply add and use the writer explicitly? This way, the changes will remain small and manageable.

ahrtr: > Should I create a kind of "pre-requisite" PR where I’ll simply add and use the writer explicitly?

Sounds good to me. Pls feel free to deliver a separate PR for that. thx

ahrtr:
pls rebase this PR instead of merging main into your dev branch. Also pls consider adding an unit test as mentioned in https://github.com/etcd-io/etcd/pull/19469#discussion_r2144585903

kstrifonoff:
> pls rebase this PR instead of merging main into your dev branch. 

@ahrtr Thanks for the comment. In any case, I intend to squash all the commits in this branch once I complete the tasks before submitting the PR for re-approval. In such a scenario, I believe it doesn’t matter whether I rebase or merge.

However, I believe it might be worthwhile to convert this to a draft until then.

ahrtr (approved):
LGTM & thx

Elbehery:
@ivanvc did @kstrifonoff address your review ?

PR #19470, opened 2025-02-23
refactor: use a more straightforward return value

beforetech:
use a more straightforward return value


beforetech:
Thanks for your review.

serathius (approved):

Issue #19354, opened 2025-02-06
Migrate experimental flags in robustness tests


siyuanfoundation:
### What would you like to be added?

With https://github.com/etcd-io/etcd/issues/19141 and https://github.com/etcd-io/etcd/issues/18023, we will be removing all experimental flags in 3.7. 

The robustness tests are used by different binary versions, for the same parameter, we would need to use the experimental flag in 3.5 and 3.4, while using the non-experimental flag in 3.6. 

The affected flags include:
- [ ] experimental-snapshot-catchup-entries
- [ ] experimental-compaction-batchLimit
- [ ] experimental-watch-process-notify-interval

### Why is this needed?

The robustness test would fail when we delete the deprecated flags if we don't fix this.

kavirajk:
@siyuanfoundation I would like to pick this up if that's ok.

Also quick clarification. I see few more experimental flags are to be removed in v1.37.

```bash
-bash-5.2$ rg "TODO: Delete in v3.7" -A 3 | rg "Experimental" | rg "json" | awk '{print$2}'
ExperimentalTxnModeWriteWithSharedBuffer
ExperimentalEnableLeaseCheckpointPersist
ExperimentalCompactionBatchLimit
ExperimentalCompactionSleepInterval
ExperimentalWatchProgressNotifyInterval
ExperimentalWarningApplyDuration
ExperimentalBootstrapDefragThresholdMegabytes
ExperimentalWarningUnaryRequestDuration
ExperimentalMaxLearners
ExperimentalPeerSkipClientSanVerification
ExperimentalMemoryMlock
ExperimentalTxnModeWriteWithSharedBuffer
ExperimentalStopGRPCServiceOnDefrag
```

Those things can also cause failure in robustness test IIUC? no? But I see you mentioned only subset of it.

siyuanfoundation:
> [@siyuanfoundation[<img alt="" width="15" height="15" src="chrome-extension://dlebflppeeemcdpidccbiblndppbmjmh/ospo-chrome-ext-logo.png">](https://teams.googleplex.com/sizhang@google.com)](https://github.com/siyuanfoundation) I would like to pick this up if that's ok.
> 
> Also quick clarification. I see few more experimental flags are to be removed in v1.37.
> 
> -bash-5.2$ rg "TODO: Delete in v3.7" -A 3 | rg "Experimental" | rg "json" | awk '{print$2}'
> ExperimentalTxnModeWriteWithSharedBuffer
[...]

PR #19506, opened 2025-03-03
robustness-test: Migrate `experimental-compaction-batch-limit` flag

kavirajk:
Related: https://github.com/etcd-io/etcd/issues/19354

Robustness tests run different versions of binaries. This flag is experimental in v3.4 and v3.5. But non-experimental in v3.6+. This PR adds ability to run different binaries with it's canonical flag name.

In robustness tests we pass these config via `embed.ServerConfig` struct and get converted into CLI flags before spawning the etcd server binary. This PR maps into right canonical flag based on versions. So even when the experimental field is removed from the struct (in v3.7), the flags will be correctly mapped.

### DIFF
--- a/tests/framework/e2e/cluster.go
+++ b/tests/framework/e2e/cluster.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/coreos/go-semver/semver"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 
@@ -645,8 +646,11 @@ func (cfg *EtcdProcessClusterConfig) EtcdServerProcessConfig(tb testing.TB, i in
 		}
 	}
 
-	defaultValues := values(*embed.NewConfig())
-	overrideValues := values(cfg.ServerConfig)
+	version, err := GetVersionFromBinary(execPath)
+	require.NoError(tb, err)
+
+	defaultValues := values(version, *embed.NewConfig())
+	overrideValues := values(version, cfg.ServerConfig)
 	for flag, value := range overrideValues {
 		if defaultValue := defaultValues[flag]; value == "" || value == defaultValue {
 			continue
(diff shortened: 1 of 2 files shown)


tests/framework/e2e/cluster.go
```diff
@@ -727,16 +731,30 @@ func (epc *EtcdProcessCluster) MinServerVersion() (*semver.Version, error) {
...
+	return name
+}
+
+func values(version *semver.Version, cfg embed.Config) map[string]string {
 	fs := flag.NewFlagSet("etcd", flag.ContinueOnError)
+
 	cfg.AddFlags(fs)
 	values := map[string]string{}
 	fs.VisitAll(func(f *flag.Flag) {
+		name := canonicalFlagName(version, f.Name)
```
serathius, line 752: This approach will not work because rendering flags happens before we run the cluster and we never re-render them. There are tests that upgrade/downgrade binary while using same args. The proper solution would be to re-render the flags every time we start binary.

For that to work we we would want to rewrite `EtcdServerProcessConfig` to not include `Args` field, but `Args` method, however this is pretty complicated rewrite.

kavirajk: Thanks @serathius for catching that :) 

I'm assuming you are talking [this code](https://github.com/kavirajk/etcd/blob/kavirajk%2Fmigrate-experimental-compaction-batch-limit-robustness-test/tests/framework/e2e/downgrade.go#L142-L164) about upgrade/downgrade the etcd process with newPath/newVersion by using same args. correct?

I like your idea of making `EtcdServerProcessConfig`'s `Args` more dynamic by making it as method instead (keeping the filed `args` as private). So [during actual spawning of the process](https://github.com/kavirajk/etcd/blob/kavirajk%2Fmigrate-experimental-compaction-batch-limit-robustness-test/tests/framework/e2e/etcd_process.go#L171), it will call `EtcdServerPRocessConfig.Args()` and that will return canonical flags based on version derived from it's `EtcdServerProcessConfig.ExecPath`. Do that sounds good to you? wdyt?

> however this is pretty complicated rewrite.

Your concern here is I believe is mostly extra work needed to update it in every places. Or am I missing anything?

One more clarification. Any efficient way to run `e2e-tests` in local? Every time I run ` make test-e2e` it takes for ever even to start (and I think we need approval to trigger from CI)

kavirajk: Thinking bit more about this. 

On the other hand, this is only needed for specific cases in Robustness tests. Making it part of original `EtcdServerProcessConfig` seems bit irrelevant. Plus `Args()` api then need to know the version to return correct Args and in order to extract it from binary path introduce a failure case, that this api has to handle. All feel bit un-necessary to put it in `EtcdServerProcessConfig`. 

Because by general definition of this config, once the it is created, it's Args are fixed for a single binary. It's our use case of downgrade and upgrade in robustness tests makes this problematic. So we can have a `helper(version, args) args` and use it just in robustness tests before spawning the server. Just like the way we do with [newExecPath](https://github.com/kavirajk/etcd/blob/kavirajk%2Fmigrate-experimental-compaction-batch-limit-robustness-test/tests/framework/e2e/downgrade.go#L158) in those tests.

I don't know. Just thinking out loud. curious to know what you think :)

PR #19536, opened 2025-03-05
perf(mvcc): optimize Hashes method using copy instead of append

1911860538:


serathius:
Not a big fan of micro optimizations without any benchmark nor evidence to back them up.

Issue #19557, opened 2025-03-08
[release-3.5] Learner promotion isn't persisted to v3store (bbolt)


ahrtr:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

## What's the symptom
When upgrading etcd from v3.5.x to v3.6.0-rc.x, the upgrade may fail due to "`membership: too many learner members in cluster`", because etcd allows only 1 learner at most by default.

Thanks @neolit123 for uploading the initial log.

## Which versions are impacted
Note only release-3.5 has this issue. The issue was introduced in `v3.5.1` in https://github.com/etcd-io/etcd/pull/13348. All etcd patch versions in `v3.5.1 - v3.5.19` are affected.

If you ever added & promoted learner(s) in `v3.5.1 - v3.5.19` and try to upgrade from 3.5.1+ to v3.6.0-rc.x, then you will see this issue,
- If you only added & promoted one learner, you will see that the member become a learner again after upgrading to `v3.6.0-rc.x`;
- if you added & promoted multiple learner (>=2), then the upgrade will fail, because the etcdserver will crash on bootstrap due to `membership: too many learner members in cluster`.

## What's the root cause
When promoting a learner, the change is only persisted in v2store, not in v3store. The reason is simple, because etcd returns `errMemberAlreadyExist`. Clearly, the member (learner) ID has already existed. See 3.5 code below,

https://github.com/etcd-io/etcd/blob/1810af31b22c0784c85073cddb09a9b8737ce6a2/server/etcdserver/api/membership/store.go#L57-L59

So the membership data will be **inconsistent** between v2store and v3store in such case.

## Why we only see this issue when upgrading from 3.5 to 3.6?

In 3.5, the v2store is the source of truth for the membership data. In 3.6, v3store (bbolt) is the source of truth for the membership data. When upgrading from 3.5.x to 3.6, the source of truth changes.

### Manual steps
[...]

PR #19563, opened 2025-03-10
[release-3.5] Fix the learner promotion changes not being persisted into v3store (bbolt)

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/19557

Also added a generic verification on ensure the membership data in v3store is always in sync with v2store.

### DIFF
--- a/server/etcdserver/api/membership/cluster_test.go
+++ b/server/etcdserver/api/membership/cluster_test.go
@@ -1049,12 +1049,6 @@ func TestAddMemberSyncsBackendAndStoreV2(t *testing.T) {
 			backendMembers: []*Member{alice},
 		},
 		{
-			name:           "Adding member should fail if it exists in both",
-			storeV2Members: []*Member{alice},
-			backendMembers: []*Member{alice},
-			expectPanics:   true,
-		},
-		{
 			name:           "Adding member should fail if it exists in storeV2 and backend is nil",
 			storeV2Members: []*Member{alice},
 			backendNil:     true,
--- a/server/etcdserver/api/membership/cluster_test.go
+++ b/server/etcdserver/api/membership/cluster_test.go
@@ -1049,6 +1049,11 @@ func TestAddMemberSyncsBackendAndStoreV2(t *testing.T) {
 			backendMembers: []*Member{alice},
 		},
 		{
+			name:           "Adding member should success if it exists in both",
+			storeV2Members: []*Member{alice},
+			backendMembers: []*Member{alice},
+		},
+		{
 			name:           "Adding member should fail if it exists in storeV2 and backend is nil",
 			storeV2Members: []*Member{alice},
 			backendNil:     true,
--- a/server/etcdserver/api/membership/store.go
+++ b/server/etcdserver/api/membership/store.go
@@ -54,9 +54,6 @@ func unsafeSaveMemberToBackend(lg *zap.Logger, be backend.Backend, m *Member) er
 	tx := be.BatchTx()
 	tx.LockInsideApply()
 	defer tx.Unlock()
-	if unsafeMemberExists(tx, mkey) {
-		return errMemberAlreadyExist
-	}
 	tx.UnsafePut(buckets.Members, mkey, mvalue)
 	return nil
 }
--- a/server/etcdserver/server_test.go
+++ b/server/etcdserver/server_test.go
@@ -655,6 +655,7 @@ func TestApplyConfigChangeUpdatesConsistIndex(t *testing.T) {
 
 	be, _ := betesting.NewDefaultTmpBackend(t)
 	defer betesting.Close(t, be)
+	cl.SetBackend(be)
 	cindex.CreateMetaBucket(be.BatchTx())
 
 	cl.AddMember(&membership.Member{ID: types.ID(1)}, true)
--- a/tests/e2e/ctl_v2_test.go
+++ b/tests/e2e/ctl_v2_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"go.etcd.io/etcd/server/v3/verify"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
 
@@ -228,6 +229,10 @@ func TestUtlCtlV2Backup(t *testing.T) {
 
 func testUtlCtlV2Backup(t *testing.T, snapCount int, v3 bool, utl bool) {
 	BeforeTestV2(t)
+
+	// disable the verification because this case updated the db offline
+	revertFunc := verify.DisableVerification()
+	defer revertFunc()
 
 	backupDir, err := ioutil.TempDir(t.TempDir(), "testbackup0.etcd")
 	if err != nil {
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -696,6 +696,10 @@ func (c *RaftCluster) IsReadyToPromoteMember(id uint64) bool {
 	return true
 }
 
+func (c *RaftCluster) MembersFromStore() (map[types.ID]*Member, map[types.ID]bool) {
+	return membersFromStore(c.lg, c.v2store)
+}
+
 func membersFromStore(lg *zap.Logger, st v2store.Store) (map[types.ID]*Member, map[types.ID]bool) {
 	members := make(map[types.ID]*Member)
 	removed := make(map[types.ID]bool)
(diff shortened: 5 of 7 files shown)


ahrtr:
cc @fuweid @ivanvc @serathius @jmhbnz @siyuanfoundation PTAL

This PR is the first item of the proposed actions mentioned in https://github.com/etcd-io/etcd/issues/19557#issuecomment-2710148232

Once this gets merged, I will work on the second item "auto sync v3store based on v2store on bootstrap" in release-3.5 instead of release-3.6.

serathius:
This PR seems to fix the bug in leader promotion, but doesn't patches store for clusters that already encountered the bug. Meaning that they will still need to run the `etcdutl check members` && `etcdutl sync members` commands, making it critical command before upgrade to v3.6 or risking cluster shuttdown.

ahrtr:
> This PR seems to fix the bug in leader promotion

"leader promotion" --> "learner promotion".

Yes, exactly.



> but doesn't patches store for clusters that already encountered the bug.

It will be covered in a followup PR, also mentioned this multiple times. Please see the second item as mentioned in https://github.com/etcd-io/etcd/issues/19557#issuecomment-2710148232

> Meaning that they will still need to run the `etcdutl check members` && `etcdutl sync members` commands, making it critical command before upgrade to v3.6 or risking cluster shuttdown.

Already clarified this several times, please see the proposed action and also https://github.com/etcd-io/etcd/issues/19557#issuecomment-2710714977

tests/e2e/ctl_v2_test.go
```diff
@@ -229,6 +230,9 @@ func TestUtlCtlV2Backup(t *testing.T) {
 func testUtlCtlV2Backup(t *testing.T, snapCount int, v3 bool, utl bool) {
 	BeforeTestV2(t)
 
+	revertFunc := verify.DisableVerification()
```
serathius, line 233: Can you provide context why we disable verification here?

ahrtr: sure, it's because the case offline updated the bbolt db file. Will add a comment.

server/etcdserver/api/membership/store.go
```diff
@@ -220,8 +220,23 @@ func unsafeSaveMemberToStore(lg *zap.Logger, s v2store.Store, m *Member) error {
 		lg.Panic("failed to marshal raftAttributes", zap.Error(err))
 	}
 	p := path.Join(MemberStoreKey(m.ID), raftAttributesSuffix)
-	_, err = s.Create(p, false, string(b), false, v2store.TTLOptionSet{ExpireTime: v2store.Permanent})
-	return err
+	if _, err = s.Create(p, false, string(b), false, v2store.TTLOptionSet{ExpireTime: v2store.Permanent}); err != nil {
```
serathius, line 220: This seems like a separate issue. Thanks for adding validation! However I'm worried about trying to fix two issues in one PR thus not putting enough attention into understanding each of them and introducing proper testing.

server/etcdserver/server_test.go
```diff
@@ -599,6 +599,11 @@ func TestApplyConfChangeError(t *testing.T) {
 func TestApplyConfChangeShouldStop(t *testing.T) {
 	cl := membership.NewCluster(zaptest.NewLogger(t))
 	cl.SetStore(v2store.New())
+
+	be, _ := betesting.NewDefaultTmpBackend(t)
```
serathius, line 603: Could we merge this commit (at least the addition of Backends in tests) as a separate PR?

tests/e2e/v2store_deprecation_test.go
```diff
@@ -56,7 +56,7 @@ func assertVerifyCanStartV2deprecationNotYet(t testing.TB, dataDirPath string) {
 
 	if err := e2e.CURLGet(epc, e2e.CURLReq{
 		Endpoint: "/v2/keys/foo",
-		Expected: `{"action":"get","node":{"key":"/foo","value":"bar9","modifiedIndex":13,"createdIndex":13}}`}); err != nil {
+		Expected: `{"action":"get","node":{"key":"/foo","value":"bar9","modifiedIndex":14,"createdIndex":14}}`}); err != nil {
```
serathius, line 59: Would like to better understand the source of the changes in index. If possible separate changes in logic from  changes to tests.

server/etcdserver/api/membership/store.go
ahrtr: > This seems like a separate issue.

Yes, it's a separate (but related) minor issue. When adding a new member, 
- v3store persists the whole member, 
- but v2store only persists the raftAttributes (peerURL, isLearner); so it causes a temprary inconsistency between v2 and v3store, and it depends on the following publish logic to update the attributes (name, clientURL), so it converge soon.

If we don't fix it in this PR, then the verification will fail, otherwise I will have to update the verification to ignore the attribute and only verify the raftAttributes(including peerURLs, isLearner).

Since from users perspective, this isn't an issue. We can also leave it as it's and only verify raftAttribute.

tests/e2e/v2store_deprecation_test.go
ahrtr: caused by https://github.com/etcd-io/etcd/pull/19563#discussion_r1988974045

server/etcdserver/server_test.go
ahrtr: just raised https://github.com/etcd-io/etcd/pull/19573 PTAL

server/etcdserver/api/membership/store.go
serathius: > If we don't fix it in this PR, then the verification will fail, otherwise I will have to update the verification to ignore the attribute and only verify the raftAttributes(including peerURLs, isLearner).

That's why I'm suggesting to send PR to just fix this so we can merge it fast. This could also be backported to v3.4.

ahrtr: > Since this isn't an issue from user perspective, so suggest to keep it as it's

I mean that I will revert the change.

ahrtr: reverted the change.

server/etcdserver/api/membership/cluster_test.go
```diff
@@ -1048,12 +1048,6 @@ func TestAddMemberSyncsBackendAndStoreV2(t *testing.T) {
 			name:           "Adding member should succeed if it was only in backend",
 			backendMembers: []*Member{alice},
 		},
-		{
```
serathius, line 1051: Reason for this test was to confirm that we will still return error when we try to add an existing member to cluster. The bug came from fact that we forgot about learners and setting attributes, however I think we should have an equivalent test.

ahrtr: updated

ahrtr:
cc @jmhbnz @spzala @serathius @wenjiaswe 

cc @ivanvc @fuweid @siyuanfoundation

tests/e2e/ctl_v2_test.go
```diff
@@ -229,6 +230,10 @@ func TestUtlCtlV2Backup(t *testing.T) {
 func testUtlCtlV2Backup(t *testing.T, snapCount int, v3 bool, utl bool) {
 	BeforeTestV2(t)
 
+	// disable the verification because this case offline updated the db
```
siyuanfoundation, line 233:
```suggestion
	// disable the verification because this case updated the bbolt db offline
```

siyuanfoundation (approved):

tests/e2e/ctl_v2_test.go
ahrtr: updated.

ahrtr:
cc @jmhbnz @spzala @serathius @wenjiaswe

server/etcdserver/api/membership/cluster_test.go
```diff
@@ -1049,10 +1049,9 @@ func TestAddMemberSyncsBackendAndStoreV2(t *testing.T) {
 			backendMembers: []*Member{alice},
 		},
 		{
-			name:           "Adding member should fail if it exists in both",
+			name:           "Adding member should success if it exists in both",
```
serathius, line 1052: This test case was meant to prevent same member being added twice, while we are fixing the issue of v3store rejecting learner being promoted. Should the test be updated to separate those two cases?

ahrtr: For v3store (bbolt), when we add a member or promote a learner, etcd just marshals the member and write the data into bbolt, if it already exists, just overwrites it. This is the behaviour in both release-3.4, release-3.6 and main. But In release-3.5, it returned an error already exists, it's a bug and now we fix it in this PR.

Also FYI.
- When we add a member again, the `ValidateConfigurationChange` will fail.
- We also have separate unit test case to verify the v2store will return error when adding a member again. (see line 1057 in cluster_test.go).

ahrtr: > Should the test be updated to separate those two cases?

No need. Please see my response above.

serathius (approved):

Issue #19524, opened 2025-03-04
Bump Go to 1.24.1 / 1.23.7


ivanvc:
### What would you like to be added?

Golang 1.24.1 and 1.23.7 were released today. According to our [Dependency management documentation](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#golang-versions), we want to stay on the latest patch version. This means updating the `release-3.4`, `release-3.5` and `release-3.6` branches to 1.23.7 and `main` to 1.24.1.

This patch includes the security fix for CVE-2025-22870.

Progress track:

- [x] [etcd](https://github.com/etcd-io/etcd)
  - [x] [`main`](https://github.com/etcd-io/etcd/tree/main): go v1.24.1 - #19440
  - [x] [`release-3.6`](https://github.com/etcd-io/etcd/tree/release-3.6): go v1.23.7 - #19527
  - [x] [`release-3.5`](https://github.com/etcd-io/etcd/tree/release-3.5): go v1.23.7 - #19528
  - [x] [`release-3.4`](https://github.com/etcd-io/etcd/tree/release-3.4): go v1.23.7 - #19533
  - [x] [`CHANGELOG`](https://github.com/etcd-io/etcd/tree/main/CHANGELOG) - #19532 
- [x] [bbolt](https://github.com/etcd-io/bbolt)
  - [x] [`main`](https://github.com/etcd-io/bbolt/tree/main): go v1.24.1 - etcd-io/bbolt#925
  - [x] [`release-1.3`](https://github.com/etcd-io/bbolt/tree/release-1.3): v1.23.7 - etcd-io/bbolt#920
  - [x] [`release-1.4`](https://github.com/etcd-io/bbolt/tree/release-1.4): v1.23.7 - etcd-io/bbolt#921
- [x] [raft](https://github.com/etcd-io/raft)
  - [x] [`main`](https://github.com/etcd-io/raft/tree/main): go v1.24.1 - etcd-io/raft#275
  - [x] [`release-3.6`](https://github.com/etcd-io/raft/tree/release-3.6): go v1.23.7 - etcd-io/raft#276
- [x] [`etcd-io/gofail` `main`](https://github.com/etcd-io/gofail/tree/main): go v1.23.7 - etcd-io/gofail#107
- [x] [`etcd-io/auger` `main`](https://github.com/etcd-io/auger/tree/main): go v1.24.1 - etcd-io/auger#222
[...]

PR #19582, opened 2025-03-12
Bump devcontainer version to 1.24

henrybear327:
I think this is the last missing piece in https://github.com/etcd-io/etcd/pull/19440 :)

Reference:
- https://github.com/etcd-io/etcd/issues/19524

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,7 @@
 {
 	"name": "Go",
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-	"image": "mcr.microsoft.com/devcontainers/go:1.23-bookworm",
+	"image": "mcr.microsoft.com/devcontainers/go:dev-1.24-bookworm",
 	// Features to add to the dev container. More info: https://containers.dev/features.
 	"features": {
 		"ghcr.io/devcontainers/features/docker-in-docker:2": {},


ivanvc (approved):
Good catch. LGTM. Thanks, @henrybear327.

henrybear327:
@ivanvc for some reason, according to the available image tags [here](https://mcr.microsoft.com/v2/devcontainers/go/tags/list), they only offer the `dev-` prefixed images for 1.24 now. 

I have tested the changes on GitHub codespaces (as this is my main development environment now) which works for now, but we might need to keep an eye out for the images tag when it's updated.

ivanvc:
> @ivanvc for some reason, according to the available image tags [here](https://mcr.microsoft.com/v2/devcontainers/go/tags/list), they only offer the `dev-` prefixed images for 1.24 now.

Shouldn't we prefer the `dev-` ones? I'm assuming that they may have some development libraries installed.

ahrtr (approved):
thx

ivanvc:
@henrybear327, just a heads up that 1.24 is now listed on the API you provided. So, if the `dev-` prefix is actually a prefix Microsoft uses for experimental versions, we can drop it and bump it to `1.24-bookworm`.

henrybear327:
> @henrybear327, just a heads up that 1.24 is now listed on the API you provided. So, if the `dev-` prefix is actually a prefix Microsoft uses for experimental versions, we can drop it and bump it to `1.24-bookworm`.

Thanks @ivanvc

PR #19586, opened 2025-03-13
[release-3.5] [Solution 1] Auto sync members in v3store if Islearner is the only field that differs between v2store and v3store

ahrtr:
Provide a fix for the users who have already been affected by https://github.com/etcd-io/etcd/issues/19557

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -299,6 +299,9 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
 		zap.Strings("listen-metrics-urls", e.cfg.getMetricsURLs()),
 	)
 	serving = true
+
+	e.Server.SyncLearnerPromotionIfNeeded()
+
 	return e, nil
 }
 
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"reflect"
 	"sort"
 	"strings"
 	"sync"
@@ -540,6 +541,53 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 		zap.String("local-member-id", c.localID.String()),
 		zap.String("promoted-member-id", id.String()),
 	)
+}
+
+func (c *RaftCluster) SyncLearnerPromotionIfNeeded() {
+	c.Lock()
+	defer c.Unlock()
+
+	v2Members, _ := membersFromStore(c.lg, c.v2store)
+	v3Members, _ := membersFromBackend(c.lg, c.be)
+
+	for id, v2Member := range v2Members {
+		v3Member, ok := v3Members[id]
+
+		if !ok {
+			c.lg.Error("Detected member only in v2store but missing in v3store", zap.String("member", fmt.Sprintf("%+v", *v2Member)))
+			continue
+		}
+
+		// A peerURL list is considered the same regardless of order.
+		clonedV2Member := v2Member.Clone()
+		sort.Strings(clonedV2Member.PeerURLs)
+
+		clonedV3Member := v3Member.Clone()
+		sort.Strings(clonedV3Member.PeerURLs)
+
+		if reflect.DeepEqual(clonedV2Member.RaftAttributes, clonedV3Member.RaftAttributes) {
+			c.lg.Info("Member's RaftAttributes is consistent between v2store and v3store", zap.String("member", fmt.Sprintf("%+v", *v2Member)))
+			continue
+		}
+
+		// Sync member iff both conditions below are true,
+		//  1. IsLearner is the only field in RaftAttributes that differs between v2store and v3store.
+		//  2. v2store.IsLearner == false && v3store.IsLearner == true.
+		clonedV3Member.IsLearner = false
+		if reflect.DeepEqual(clonedV2Member.RaftAttributes, clonedV3Member.RaftAttributes) {
+			syncedV3Member := v3Member.Clone()
+			syncedV3Member.IsLearner = false
+			c.lg.Warn("Syncing member in v3store", zap.String("member", fmt.Sprintf("%+v", *syncedV3Member)))
+			unsafeHackySaveMemberToBackend(c.lg, c.be, syncedV3Member)
+			c.be.ForceCommit()
+			continue
+		}
+
+		c.lg.Error("Cannot sync member in v3store due to IsLearner not being the only field that differs between v2store amd v3store",
+			zap.String("v2member", fmt.Sprintf("%+v", *v2Member)),
+			zap.String("v3member", fmt.Sprintf("%+v", *v3Member)),
+		)
+	}
 }
 
 func (c *RaftCluster) UpdateRaftAttributes(id types.ID, raftAttr RaftAttributes, shouldApplyV3 ShouldApplyV3) {
--- a/server/etcdserver/api/membership/cluster_test.go
+++ b/server/etcdserver/api/membership/cluster_test.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/stretchr/testify/assert"
-	betesting "go.etcd.io/etcd/server/v3/mvcc/backend/testing"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 
@@ -33,6 +33,7 @@ import (
 	"go.etcd.io/etcd/raft/v3/raftpb"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v2store"
 	"go.etcd.io/etcd/server/v3/mock/mockstore"
+	betesting "go.etcd.io/etcd/server/v3/mvcc/backend/testing"
 )
 
 func TestClusterMember(t *testing.T) {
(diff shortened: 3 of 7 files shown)


ahrtr:
@ivanvc once this PRs get merged and a couple of days workflow/pipeline soak, then we are ready to release 3.5.20.

@fuweid the new Kubeadm style e2e test should pass once 3.5.20 is out.

server/etcdserver/api/membership/cluster_test.go
```diff
@@ -1210,3 +1211,130 @@ func TestRemoveMemberSyncsBackendAndStoreV2(t *testing.T) {
...
+						PeerURLs:  []string{"http://10.0.0.9:2380"},
+						IsLearner: true,
+					},
+				},
+			},
+			expectV3Members: map[types.ID]*Member{
+				100: {
+					ID: 100,
+					RaftAttributes: RaftAttributes{
+						PeerURLs:  []string{"http://10.0.0.9:2380"},
```
siyuanfoundation, line 1308: nit: is it possible for the `PeerURLs` to be the same but ordered differently in v2store and v3 store?

ahrtr: > is it possible for the `PeerURLs` to be the same but ordered differently in v2store and v3 store?

It's not possible. The whole `RaftAttributes` is always json.Marshalled or json.Unmarshalled into/from both v2store and v3store. release-3.5 is a stable release, we will never change that logic / order.

ahrtr: Updated even it's not possible. The implementation should be correct by itself

ahrtr:
Hi folks, please take a look when you get time, thx

server/embed/etcd.go
```diff
@@ -299,6 +299,9 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
 		zap.Strings("listen-metrics-urls", e.cfg.getMetricsURLs()),
 	)
 	serving = true
+
+	e.Server.SyncLearnerPromotionIfNeeded()
```
serathius, line 303: Changing cluster membership after the apply loop has been started seems like a bad idea. My expectation was that we shouldn't modify cluster state outside apply loop. Any reason to do that?

server/etcdserver/api/membership/cluster.go
```diff
@@ -542,6 +543,53 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
 }
 
+func (c *RaftCluster) SyncLearnerPromotionIfNeeded() {
+	c.Lock()
+	defer c.Unlock()
+
+	v2Members, _ := membersFromStore(c.lg, c.v2store)
+	v3Members, _ := membersFromBackend(c.lg, c.be)
+
+	for id, v2Member := range v2Members {
```
serathius, line 553: For v3.6 we care about v3membership information. If for some reason v2 and v3 has different members (this is what happen in previous attempt to make v3 authoritative), it's safer to check v3 members.

server/etcdserver/api/membership/cluster.go
```diff
@@ -542,6 +543,53 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
+		}
+
+		// A peerURL list is considered the same regardless of order.
+		clonedV2Member := v2Member.Clone()
+		sort.Strings(clonedV2Member.PeerURLs)
+
+		clonedV3Member := v3Member.Clone()
+		sort.Strings(clonedV3Member.PeerURLs)
+
+		if reflect.DeepEqual(clonedV2Member.RaftAttributes, clonedV3Member.RaftAttributes) {
```
serathius, line 568: Could you just compare IsLearner? We don't care about other information

server/etcdserver/api/membership/cluster.go
```diff
@@ -542,6 +543,53 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
+
+		// Sync member iff both conditions below are true,
+		//  1. IsLearner is the only field in RaftAttributes that differs between v2store and v3store.
+		//  2. v2store.IsLearner == false && v3store.IsLearner == true.
+		clonedV3Member.IsLearner = false
+		if reflect.DeepEqual(clonedV2Member.RaftAttributes, clonedV3Member.RaftAttributes) {
+			syncedV3Member := v3Member.Clone()
+			syncedV3Member.IsLearner = false
+			c.lg.Warn("Syncing member in v3store", zap.String("member", fmt.Sprintf("%+v", *syncedV3Member)))
+			unsafeSaveMemberToBackend(c.lg, c.be, syncedV3Member)
```
serathius, line 581: `unsafeSaveMemberToBackend` uses `LockInsideApply` this means two thing:
* This would cause CI to be incorrectly flushed (see my recommendation above about doing changes to members from apply loop).
* The validation for `LockInsideApply` didn't catch this locking. We should double check why.

server/embed/etcd.go
ahrtr: Expectation is good, but not feasible.

I already clarified it in the comment for the method. We must wait for etcd finishes replaying the WAL records.

server/etcdserver/api/membership/cluster.go
ahrtr: This PR is for 3.5 instead of 3.6. In 3.5, v2store is the source of truth. Also we are syncing from v2store to v3store.

Also note that we need to check both v2 and v3store, the member ID must match. Not sure what's your concern here.

server/etcdserver/api/membership/cluster.go
ahrtr: > * This would cause CI to be incorrectly flushed (see my recommendation above about doing changes to members from apply loop).

thx for pointing this out. Will update to use the LockOutsideApply.



> * The validation for `LockInsideApply` didn't catch this locking. We should double check why.

because there isn't an e2e test to trigger it being executed. 

PTAL https://github.com/etcd-io/etcd/pull/19602, once it gets merged, will continue to update that e2e test in this PR.

server/etcdserver/api/membership/cluster.go
ahrtr: RaftAttribute includes both peerURLs and IsLearner, both of them should always match between v2 and v3store. I think it's more prudent to double check peerURL matches.

server/embed/etcd.go
serathius: > We must wait for etcd finishes replaying the WAL records.

Yes, we can also do that in apply loop.

ahrtr: > Yes, we can also do that in apply loop.

If we intentionally issue a member promotion request, then only one member should do it. Then we must wait for leader gets elected?

I really hate the leader's privilege mode; it causes some uncertainties, i.e. leader change etc.  I agree that doing this outside of apply loop isn't ideal, but actually it's safe for this specific case. And we would never let such code/implementation go into main branch.

serathius: I'm not proposing to do a member promotion request.

Just a local patching of v3store:
* When bootstrapping read v2 and v3, if there is a unpromoted member in v3 that is promoted in v2, we can mark this member in v3 as promoted. This is correct because it member can only be promoted once and v2 has always older data than v3.
* To handle Promote Member in WAL, depending on etcd version:
  * For v3.5, it should just work with fixes for PromoteMember. This should work as v2 store is the source of truth.
  * For v3.6, when bootstrapping we read WAL since v2 snapshot, if we find any MemberPromote, just mark it as promoted in v3. This needs to be done before we start raft as v3 store should be source of truth.

Proposed actions are based on assumption (would need to confirm as I might remember it wrongly), that in v3.6 the v3 store is source of truth of membership even when raft is just started and even during replaying of WAL entries between v2 snapshot and v3 consistent index.

ahrtr: > * For v3.5, it should just work with fixes for PromoteMember. This should work as v2 store is the source of truth.

Normally it won't work because the index might be < consistent_index. It means that we need to make an exception for learner promotion request, apply it for v3store anyway no matter `shouldApplyV3` is true or not. 

More important is that it may cause data inconsistence in mix version case, i.e. 3.5.19 vs 3.5.20. In 3.5.19 it won't apply, but 3.5.20 apply, then they may end of having different consisitent index.



> in v3.6 the v3 store is source of truth of membership even when raft is just started and even during replaying of WAL entries between v2 snapshot and v3 consistent index.

It's true. But still I don't think we should do any such hacky workaround in 3.6.

serathius: > Normally it won't work because the index might be < consistent_index. It means that we need to make an exception for learner promotion request, apply it for v3store anyway no matter shouldApplyV3 is true or not.

Right, forgot about it.

> More important is that it may cause data inconsistence in mix version case, i.e. 3.5.19 vs 3.5.20. In 3.5.19 it won't apply, but 3.5.20 apply, then they may end of having different consisitent index.

Not sure how. We just override the field without bumping CI.

serathius: > It's true. But still I don't think we should do any such hacky workaround in 3.6.

Heh, would love to have robustness tests work for membership information. It's just hard to test non linearizable requests and dynamic clusters.

ahrtr: > Not sure how. We just override the field without bumping CI.

You are also creating the second big exception: use `LookOutsideApply` in apply. Also you won't advance the consistent_index even in normal case (promoting a new learner which is added in 3.5.20).

ahrtr: As mentioned in https://github.com/etcd-io/etcd/issues/19557#issuecomment-2713409509, doing such hacky auto-sync in 3.6 is almost infeasible.

I am updating the e2e test right now, still following the existing approach...

server/etcdserver/api/membership/cluster.go
ahrtr: can we resolve this comment?

server/etcdserver/api/membership/cluster.go
ahrtr: can we resolve this comment?

server/etcdserver/api/membership/cluster.go
ahrtr: updated.

ahrtr:
e2e test updated, PTAL, thx

server/embed/etcd.go
ahrtr: Just raised another PR (solution 2) https://github.com/etcd-io/etcd/pull/19606

ahrtr: Hash only is calcuated based on key space. 

pls review https://github.com/etcd-io/etcd/pull/19606 instead, let's not waste time

ahrtr: also data inconsistence has nothing to do with this PR (and https://github.com/etcd-io/etcd/pull/19606). 

mix-version will lead to member ship data inconsistence, but it will converge once upgrading to 3.5.20+

ahrtr: I pinged your guys in slack, also clearly mentioned that preferred to https://github.com/etcd-io/etcd/pull/19606. It's really confused why you still review this PR? 

If you insist on reviewing this PR, as least you should state why you prefer to this PR firstly, instead of wasting time to still review it.

siyuanfoundation: If you had closed this PR, then that would not have result in the confusion.

ahrtr: I already clarified in above comment. What confusion do you still have? 

again, let's focus on https://github.com/etcd-io/etcd/pull/19606

PR #19594, opened 2025-03-13
test/robustness: rename makefile.mk to Makefile

ivanvc:
This idea came from a robustness meeting where @serathius explained how the robustness tests `Makefile` was structured. Renaming the file to `Makefile` allows users to execute `make` commands from the `tests/robustness` directory. However, to also be able to import it from the top-level `Makefile`, we need to ensure that we are not using relative directories. This is simple to achieve by introducing variables to the repository's top-level, and executing the top-level `make`/`Makefile` from the robustness by using make's `--directory` argument.

/cc @serathius

### DIFF
--- a/tests/robustness/Makefile
+++ b/tests/robustness/Makefile
@@ -1,68 +1,76 @@
+REPOSITORY_ROOT := $(shell git rev-parse --show-toplevel)
+
 .PHONY: test-robustness-reports
-test-robustness-reports: export GOTOOLCHAIN := go$(shell cat .go-version)
+test-robustness-reports: export GOTOOLCHAIN := go$(shell cat $(REPOSITORY_ROOT)/.go-version)
 test-robustness-reports:
-	cd ./tests && go test ./robustness/validate -v --count 1 --run TestDataReports
+	cd $(REPOSITORY_ROOT)/tests && go test ./robustness/validate -v --count 1 --run TestDataReports
 
 # Test main and previous release branches
 
+# Note that executing make at the top-level repository needs a change in the
+# directory. So, instead of calling just $(MAKE) or make, use this
+# $(TOPLEVEL_MAKE) variable.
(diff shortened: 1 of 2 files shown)


tests/robustness/Makefile
```diff
@@ -113,10 +121,10 @@ $(GOPATH)/bin/gofail: tools/mod/go.mod tools/mod/go.sum
 	cd /tmp/etcd-v3.5.$*-failpoints/; \
 	  git clone --depth 1 --branch v3.5.$* https://github.com/etcd-io/etcd.git .; \
 	  go get go.etcd.io/gofail@${GOFAIL_VERSION}; \
-	  (cd server; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
-	  (cd etcdctl; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
-	  (cd etcdutl; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
-	  (cd tools/mod; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
+	  (cd $(REPOSITORY_ROOT)/server; go get go.etcd.io/gofail@${GOFAIL_VERSION}); \
```
serathius, line 124: Don't think those packages should be changed. They are relative to the `cd /tmp/etcd-v3.5.$*-failpoints/`. This applies to all binaries built in `/tmp/etcd-*`

ivanvc: Ah, right. Makes sense. Reverting these.

Maybe a dumb question, but why are they using the `gofail` version specified in the parent repository `tools/mod`? Shouldn't they be using the version from the cloned repository?

ivanvc: Makes sense, thanks for the explanation.

I'm just making a mental note for myself: We assume the latest version is always backward compatible.

serathius (approved):

serathius:
This is really awesome!

PR #19599, opened 2025-03-14
development: fix 'make verify' errors in OSX

Dogacel:
Hi folks 👋

I am a newbie and I was excited about contributing to etcd. Currently local development workflow suggests running 'make verify' to verify all checks pass before committing. However, currently this check fails in the main branch. 

Two errors have been faced and fixed. 
- Firstly, updated check for "github.com/prometheus/procfs" in "bill-of-materials.json" to be skipped in OSX.
- Secondly, updated "scripts/test.sh" to download tools "crawford/marker" and "koalaman/shellcheck" for the host OS and architecture. Previously ARM MacOS devices were having errors while running make verify because downloaded version of shellcheck was for linux x86_64 architecture.

For reference, here are the errors faced during local setup.

1. Error before updating `bill-if-materials.json`,

```
381,389d380
<               "project": "github.com/prometheus/procfs",
<               "licenses": [
<                       {
<                               "type": "Apache License 2.0",
<                               "confidence": 1
<                       }
<               ]
<       },
<       {
modularized licenses do not match given bill of materials
FAIL: 'bom' FAILED at Fri Mar 14 00:02:01 CDT 2025
```

2. Error before updating shellcheck download,

```
stderr: ./scripts/test_lib.sh: line 148: ./bin/shellcheck: cannot execute binary file: Exec format error
FAIL: (code:126):
  % './bin/shellcheck' '-fgcc' 'scripts/build-binary.sh' 'scripts/build-docker.sh' 'scripts/build-release.sh' 'scripts/build.sh' 'scripts/build_lib.sh' 'scripts/build_tools.sh' 'scripts/codecov_upload.sh' 'scripts/fix.sh' 'scripts/fuzzing.sh' 'scripts/genproto.sh' 'scripts/markdown_diff_lint.sh' 'scripts/measure-testgrid-flakiness.sh' 'scripts/release.sh' 'scripts/release_mod.sh' 'scripts/sync_go_toolchain_directive.sh' 'scripts/test.sh' 'scripts/test_images.sh' 'scripts/test_lib.sh' 'scripts/update_dep.sh' 'scripts/update_proto_annotations.sh' 'scripts/updatebom.sh' 'scripts/verify_genproto.sh' 'scripts/verify_go_versions.sh' 'scripts/verify_proto_annotations.sh'

FAIL: 'run ./bin/shellcheck -fgcc scripts/build-binary.sh scripts/build-docker.sh scripts/build-release.sh scripts/build.sh scripts/build_lib.sh scripts/build_tools.sh scripts/codecov_upload.sh scripts/fix.sh scripts/fuzzing.sh scripts/genproto.sh scripts/markdown_diff_lint.sh scripts/measure-testgrid-flakiness.sh scripts/release.sh scripts/release_mod.sh scripts/sync_go_toolchain_directive.sh scripts/test.sh scripts/test_images.sh scripts/test_lib.sh scripts/update_dep.sh scripts/update_proto_annotations.sh scripts/updatebom.sh scripts/verify_genproto.sh scripts/verify_go_versions.sh scripts/verify_proto_annotations.sh' checking failed (!=0 return code)
FAIL: 'shellcheck' FAILED at Fri Mar 14 00:03:03 CDT 2025
```

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -67,6 +67,18 @@ if [ -z "${GOARCH:-}" ]; then
   GOARCH=$(go env GOARCH);
 fi
 
+if [ -z "${OS:-}" ]; then
+  OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+fi
+
+if [ -z "${ARCH:-}" ]; then
+  ARCH=$(uname -m)
+
+  if [ "$ARCH" = "arm64" ]; then
+    ARCH="aarch64"
+  fi
+fi
+
 # determine whether target supports race detection
 if [ -z "${RACE:-}" ] ; then
   if [ "$GOARCH" == "amd64" ] || [ "$GOARCH" == "arm64" ]; then
@@ -331,7 +343,7 @@ function shellcheck_pass {
   SHELLCHECK=shellcheck
   if ! tool_exists "shellcheck" "https://github.com/koalaman/shellcheck#installing"; then
     log_callout "Installing shellcheck $SHELLCHECK_VERSION"
-    wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar -xJv -C /tmp/ --strip-components=1
+    wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.${OS}.${ARCH}.tar.xz" | tar -xJv -C /tmp/ --strip-components=1
     mkdir -p ./bin
     mv /tmp/shellcheck ./bin/
     SHELLCHECK=./bin/shellcheck
@@ -356,7 +368,14 @@ function markdown_marker_pass {
   # TODO: check other markdown files when marker handles headers with '[]'
   if ! tool_exists "$marker" "https://crates.io/crates/marker"; then
     log_callout "Installing markdown marker $MARKDOWN_MARKER_VERSION"
-    wget -qO- "https://github.com/crawford/marker/releases/download/${MARKDOWN_MARKER_VERSION}/marker-${MARKDOWN_MARKER_VERSION}-x86_64-unknown-linux-musl.tar.gz" | tar -xzv -C /tmp/ --strip-components=1 >/dev/null
+    MARKER_OS=$OS
+    if [ "$OS" = "darwin" ]; then
+      MARKER_OS="apple-darwin"
+    elif [ "$OS" = "linux" ]; then
+      MARKER_OS="unknown-linux-musl"
+    fi
+
+    wget -qO- "https://github.com/crawford/marker/releases/download/${MARKDOWN_MARKER_VERSION}/marker-${MARKDOWN_MARKER_VERSION}-${ARCH}-${MARKER_OS}.tar.gz" | tar -xzv -C /tmp/ --strip-components=1 >/dev/null
     mkdir -p ./bin
     mv /tmp/marker ./bin/
     marker=./bin/marker


ivanvc:
Hi, @Dogacel. Thanks for the pull request. The changes to run in Darwin look reasonable. I'm unsure about the change to the `bill-of-materials.json`, as the CI would complain if it's invalid. Let's see what it says.

/ok-to-test

ivanvc:
@Dogacel, `github.com/prometheus/procfs` needs to be in the bom (https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19599/pull-etcd-verify/1900667187287298048):

```
go.mod
77:     github.com/prometheus/procfs v0.15.1 // indirect
client/v3/go.mod
34:     github.com/prometheus/procfs v0.15.1 // indirect
server/go.mod
66:     github.com/prometheus/procfs v0.15.1 // indirect
tests/go.mod
84:     github.com/prometheus/procfs v0.15.1 // indirect
etcdutl/go.mod
67:     github.com/prometheus/procfs v0.15.1 // indirect
```

Could it be that you have made changes in your environment and removed the indirect dependency? Could you please revert the change to `bill-of-materials.json`? Again, the rest of the changes look like something reasonable to add.

Dogacel:
> @Dogacel, `github.com/prometheus/procfs` needs to be in the bom (https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19599/pull-etcd-verify/1900667187287298048):
> 
> ```
> go.mod
> 77:     github.com/prometheus/procfs v0.15.1 // indirect
> client/v3/go.mod
> 34:     github.com/prometheus/procfs v0.15.1 // indirect
> server/go.mod
> 66:     github.com/prometheus/procfs v0.15.1 // indirect
> tests/go.mod
> 84:     github.com/prometheus/procfs v0.15.1 // indirect
> etcdutl/go.mod
> 67:     github.com/prometheus/procfs v0.15.1 // indirect
> ```
> 
> Could it be that you have made changes in your environment and removed the indirect dependency? Could you please revert the change to `bill-of-materials.json`? Again, the rest of the changes look like something reasonable to add.

@ivanvc I have just cloned the repository and did no changes before running `make verify`. After doing some investigation, I have realized OSX doesn't have `procfs` support, I suspect if somehow it doesn't get included in my BOM due to that.

```console
~/workdir/etcd dogac/fix-local-setup* 22s ❯ go mod why -m github.com/prometheus/procfs                                                                                                                                                                 22:42:57
# github.com/prometheus/procfs
go.etcd.io/etcd/v3/contrib/raftexample
go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp
github.com/prometheus/client_golang/prometheus
github.com/prometheus/procfs
```

So I went ahead and run the following commands,

```
GOOS=darwin go build -x ./...;
GOOS=linux go build -x ./...;
```

First one actually did not include any logs that include `procfs`, however the second one included some logs regarding `procfs`. So I feel like my theory is right in some sense? I have added a small conditional block to make that verification check pass.

Please let me know whether doing active development in `OSX` really makes sense. I guess many people would do their development in OSX but maybe they have some other settings in place (or they never run make verify?).

scripts/test.sh
```diff
@@ -488,6 +507,19 @@ function bom_pass {
...
+    output=$(echo "$output" | jq --tab '. + [{
+      "project": "github.com/prometheus/procfs",
+      "licenses": [
+        {
+          "type": "Apache License 2.0",
+          "confidence": 1
+        }
+      ]
+    }] | sort_by(.project)')
+  fi
```
ivanvc, line 521: Thanks for the changes, @Dogacel. I don't think this is a clean solution, as we'll also need to implement something similar for the `fix` target. Could you explore an alternative by adding this license to the `bill-of-materials-override.json`, and removing this workaround?

Dogacel: Unfortunately I couldn't make `override` work as `procfs` should be excluded in the BOM file rather than being overwritten. There is no way to delete entries using override file [(Source)](https://github.com/appscodelabs/license-bill-of-materials/blob/6018e0c5287c58796f780a14f5ee3b13c4e0d0fe/license-bill-of-materials.go#L634-L643). Also running `fix` just removes `procfs` from the BOM 😢

ivanvc: :face_with_head_bandage: I suggest removing the `bom` part from your pull request. If you remove that, I think we should be able to merge it, and there will be less friction in merging as the change is relatively small.

Then, in a follow-up pull request, we can address the `bom` issue. A possible solution would be to skip verifying and updating the `bom` if the architecture is other than amd64 (I wonder if Linux ARM64 would fail, too).

Dogacel: Makes sense, I also wonder if we can generate the BOM targeting the amd64 architecture in Go compiler using `GOARCH` params, maybe we can work with the BOM generator library to support that.

As you said, I'll separate that part and open another PR.

Dogacel: Reverted the BOM workaround 👍 PTAL 🙂

ivanvc (commented):
Thanks for the changes, @Dogacel. I left a minor comment. Can you also please squash your commits? Thanks again :)

scripts/test.sh
```diff
@@ -67,6 +67,18 @@ if [ -z "${GOARCH:-}" ]; then
   GOARCH=$(go env GOARCH);
 fi
 
+if [ -z "${OS:-}" ]; then
```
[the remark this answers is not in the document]
Dogacel: We default it in other places as well, without is `set -o nounset` setting throw an error.

> ./scripts/test.sh: line 70: OS: unbound variable

Dogacel:
> Thanks for the changes, @Dogacel. I left a minor comment. Can you also please squash your commits? Thanks again :)

Squashed 👍 PTAL 🙂

Dogacel:
It seems like my next contribution could be just fixing those broken tests 😄

henrybear327:
/retest

> It seems like my next contribution could be just fixing those broken tests 😄

That's where we all started with :)

ivanvc:
Thanks for squashing, @Dogacel. Sorry to bother you again, but can you un-coauthor your commit with Benjamin? I don't think he had any input in your pull request :sweat_smile:

Dogacel:
> Thanks for squashing, @Dogacel. Sorry to bother you again, but can you un-coauthor your commit with Benjamin? I don't think he had any input in your pull request 😅

Should be fixed now 👍 Thanks!

ivanvc (approved):
LGTM. Thanks for your first contribution, @Dogacel :)

ahrtr (approved):

ivanvc:
Link to #19665

Issue #19577, opened 2025-03-12
mvcc: negative watcher count gauges


kjgorman:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

We observe a slow negative trend on the `etcd_debugging_mvcc_watcher_total` gauge—implying there are unpaired increment/decrement operations:

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/63a6476f-f791-4c49-a04e-0a46e49e6283" />

### What did you expect to happen?

Logically the active watcher gauge should never be negative.

### How can we reproduce it (as minimally and precisely as possible)?

I think this issue is very closely related to [this PR][0] which previously attempted to address negative watch counts from double decrements—but covers specifically the case when `wa.compacted` is true, and there is a `cancel/close` race. It's not so easy to reproduce as it is a race condition I believe, but I think you can somewhat provoke the scenario by contriving a unit test. 

Consider for example adding this test to `mvcc/watcher_test.go`:

```go
func TestDoubledDecrementWithCompactedWatcher(t *testing.T) {
	b, _ := betesting.NewDefaultTmpBackend(t)
	s := New(zaptest.NewLogger(t), b, &lease.FakeLessor{}, StoreConfig{})
	defer cleanup(s, b)

	w := s.NewWatchStream()

	wid, err := w.Watch(0, []byte("foo"), nil, 0)
	if err != nil {
		t.Fatalf("failed to watch: %v", err)
	}

	// pretend this is now compacted
	for id, wa := range w.(*watchStream).watchers {
		if wid == id {
			wa.compacted = true
		}
	}

	wg := sync.WaitGroup{}
	wg.Add(2)
	go func() {
		w.Cancel(wid)
		wg.Done()
	}()

	time.Sleep(time.Millisecond * 50)

	go func() {
		w.Close()
		wg.Done()
	}()

	wg.Wait()

	t.Fatalf("print output")
}
```

[...]

PR #19600, opened 2025-03-14
mvcc: avoid double decrement of watcher gauge on close/cancel race

kjgorman:
This occurs specifically when the watch is for a compacted revision, as there's a possible interleaving of cancel/close that invokes the `cancelWatch` function twice.

It's fairly difficult to provoke the race condition but it is possible to observe on `main` the racing test can fail with a negative gauge:

```
$ go test ./...  -run TestNewWatcherCountGauge/compacted_watch,_close/cancel_race
--- FAIL: TestNewWatcherCountGauge (0.34s)
    watchable_store_test.go:86:  # HELP etcd_debugging_mvcc_watcher_total Total number of watchers.
         # TYPE etcd_debugging_mvcc_watcher_total gauge
        -etcd_debugging_mvcc_watcher_total -1
        +etcd_debugging_mvcc_watcher_total 0

FAIL
FAIL    go.etcd.io/etcd/server/v3/storage/mvcc  0.830s
?       go.etcd.io/etcd/server/v3/storage/mvcc/testutil [no test files]
FAIL
```

It seems as though it is partially expected for the cancel function to be invoked multiple times and to handle that safely (i.e., the existing `ch == nil` check) - the bug here is that in the `if/else if` branches it comes "too late", and multiple invocations where `wa.compacted` is true will both decrement the counter. Shifting the case up one ensures that we can't follow that decrement branch multiple times.

In fact, it seems logically more sensible to put this `wa.ch == nil` case _first_, as a guard for the function being invoked multiple times, but moving i before the sync/unsynced watch set delete functions could have a greater inadvertent functional impact (i.e., if we never deleted cancelled watches from these sets it would presumably introduce a leak), so from an abundance of caution I've made the smallest change I think will fix my issue.

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -162,11 +162,11 @@ func (s *watchableStore) cancelWatcher(wa *watcher) {
 		} else if s.synced.delete(wa) {
 			watcherGauge.Dec()
 			break
-		} else if wa.compacted {
-			watcherGauge.Dec()
-			break
 		} else if wa.ch == nil {
 			// already canceled (e.g., cancel/close race)
+			break
+		} else if wa.compacted {
+			watcherGauge.Dec()
 			break
 		}
 
--- a/server/storage/mvcc/watchable_store_test.go
+++ b/server/storage/mvcc/watchable_store_test.go
@@ -17,10 +17,12 @@ package mvcc
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap/zaptest"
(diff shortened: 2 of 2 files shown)


ivanvc:
I verified your test case against the main branch, with `stress` (refer to our [deflaking documentation](https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#check-for-flaky-tests)). And I confirm that it does have an issue, without the change:

```
15m0s: 33798 runs so far, 84 failures (0.25%), 8 active
```

With the change:

```
30m0s: 67547 runs so far, 0 failures, 8 active
```

/ok-to-test

ivanvc (commented):
This solution looks good to me, but it's not my area of expertise.

/cc @fuweid @ahrtr @serathius

serathius (approved):

serathius:
Sorry for missing this one. cc @ahrtr

ahrtr:
This PR was merged when I was reviewing it. I think we still need follow the discipline of merging a PR with two approvals in the community. We can relax the restriction with less impactful PR, i.e test, workflow checks, etc.  cc @ivanvc @jmhbnz @fuweid

@kjgorman Thank you for fixing the issue, which is a valid fix. Really sorry for the late response. We are really busy with the release of v3.6.0 in the past several months. I think we shouldn't depend on only two tech-leads to review all PRs, it would great if all maintainers/reviewers and contributors can help us review PRs (CVE & bug fixes take priority)

We are releasing v3.6.0 tomorrow, so I'd like to freeze the code change in release-3.6 unless it's a critical or major bug fix. We really don't want to see any surprise. Although this PR looks safe, it changes the core mvcc code. I think we can get it included in 3.6.1.

ahrtr:
> $ go test ./...  -run TestNewWatcherCountGauge/compacted_watch,_close/cancel_race
> --- FAIL: TestNewWatcherCountGauge (0.34s)
>     watchable_store_test.go:86:  # HELP etcd_debugging_mvcc_watcher_total Total number of watchers.
>          # TYPE etcd_debugging_mvcc_watcher_total gauge
>         -etcd_debugging_mvcc_watcher_total -1
>         +etcd_debugging_mvcc_watcher_total 0
> 
> FAIL
> FAIL    go.etcd.io/etcd/server/v3/storage/mvcc  0.830s
> ?       go.etcd.io/etcd/server/v3/storage/mvcc/testutil [no test files]
> FAIL

Confirmed that I reproduced this failure by reverting the fix.  But it's really low possibility, I ran command below multiple times, and only reproduce it once. When I changed to `-count 50`, couldn't reproduce it with multiple retries. A more reliable way to reproduce it is to inject a [failpoint](https://github.com/etcd-io/gofail). 
```
go test -run TestNewWatcherCountGauge/compacted_watch,_close/cancel_race -v --failfast -count 100
```

You can inject a sleep (i.e. 1s) failpoint at line 157 below, right after cancel() but before removing it from the map.

https://github.com/etcd-io/etcd/blob/094a4adfd660baa8e02103a6f08a5bcf0ad43f69/server/storage/mvcc/watcher.go#L156-L158

kjgorman:
Thanks for the review both 🙏 

@.ahtr the failpoint approach is a nice one - when I was investigating this issue (e.g., [see description of the original issue I raised](https://github.com/etcd-io/etcd/issues/19577)), I did literally add a `time.Sleep(time.Second)` in those lines to try and provoke the race condition 😅

ahrtr:
@kjgorman are you still working on the test case? I am OK to backport the PR as it's for now, and enhance the test case separately if you have bandwidth.

Can you please manually backport this PR to previous stable releases (3.6, 3.5 and 3.4)? thx

PR #19606, opened 2025-03-14
[release-3.5] [Solution 2] Auto sync members in v3store if Islearner is the only field that differs between v2store and v3store

ahrtr:
Second solution to provide a fix for the users who have already been affected by https://github.com/etcd-io/etcd/issues/19557. It's based on the comment https://github.com/etcd-io/etcd/pull/19586#discussion_r1995637427,
- good side
  - Sync the v3store before the raft loop getting started, and also in appy workflow. So no any potential conflict
- bad side
  - created two exceptions
    - Forcibly apply member promotion to v3store even when `shouldApplyV3` is `false`
    - Use `LookOutsideApply` inside the apply workflow, so we have to disable the verification in e2e test case.
  - updated the apply workflow... 

The first solution is https://github.com/etcd-io/etcd/pull/19586
- good side
  - did not change the apply workflow
  - it's only executed on boostrap, it's an one-time operation
- bad side
  - It updates the backend outside apply workflow, so has potential conflict. But actually it's protected by lock, so it's still safe.

Both solutions work. If there is no strong objection or preference, let's go for solution 2 (this PR)

### DIFF
--- a/server/embed/etcd.go
+++ b/server/embed/etcd.go
@@ -277,6 +277,9 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
 			return e, err
 		}
 	}
+
+	e.Server.SyncLearnerPromotionIfNeeded()
+
 	e.Server.Start()
 
 	if err = e.servePeers(); err != nil {
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"reflect"
 	"sort"
 	"strings"
 	"sync"
@@ -530,8 +531,22 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	if c.v2store != nil {
 		mustUpdateMemberInStore(c.lg, c.v2store, c.members[id])
 	}
-	if c.be != nil && shouldApplyV3 {
-		unsafeSaveMemberToBackend(c.lg, c.be, c.members[id])
+	if c.be != nil {
+		if shouldApplyV3 {
+			unsafeSaveMemberToBackend(c.lg, c.be, c.members[id])
+		} else {
+			// Workaround https://github.com/etcd-io/etcd/issues/19557 for the case that
+			// learner promotion hasn't been applied.
+			v3Members, _ := membersFromBackend(c.lg, c.be)
+			if m, ok := v3Members[id]; ok {
+				if m.IsLearner {
+					c.lg.Info("Forcibly apply member promotion", zap.String("member", fmt.Sprintf("%+v", *c.members[id])))
+					unsafeHackySaveMemberToBackend(c.lg, c.be, c.members[id])
+				} else {
+					c.lg.Info("Member promotion had already been correctly applied", zap.String("member", fmt.Sprintf("%+v", *c.members[id])))
+				}
+			}
+		}
 	}
 
 	c.lg.Info(
@@ -540,6 +555,53 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 		zap.String("local-member-id", c.localID.String()),
 		zap.String("promoted-member-id", id.String()),
 	)
+}
+
+func (c *RaftCluster) SyncLearnerPromotionIfNeeded() {
+	c.Lock()
+	defer c.Unlock()
+
+	v2Members, _ := membersFromStore(c.lg, c.v2store)
+	v3Members, _ := membersFromBackend(c.lg, c.be)
+
+	for id, v2Member := range v2Members {
+		v3Member, ok := v3Members[id]
+
+		if !ok {
+			c.lg.Error("Detected member only in v2store but missing in v3store", zap.String("member", fmt.Sprintf("%+v", *v2Member)))
+			continue
+		}
+
+		// A peerURL list is considered the same regardless of order.
+		clonedV2Member := v2Member.Clone()
+		sort.Strings(clonedV2Member.PeerURLs)
+
+		clonedV3Member := v3Member.Clone()
+		sort.Strings(clonedV3Member.PeerURLs)
+
+		if reflect.DeepEqual(clonedV2Member.RaftAttributes, clonedV3Member.RaftAttributes) {
+			c.lg.Info("Member's RaftAttributes is consistent between v2store and v3store", zap.String("member", fmt.Sprintf("%+v", *v2Member)))
+			continue
+		}
+
+		// Sync member iff both conditions below are true,
+		//  1. IsLearner is the only field in RaftAttributes that differs between v2store and v3store.
+		//  2. v2store.IsLearner == false && v3store.IsLearner == true.
+		clonedV3Member.IsLearner = false
+		if reflect.DeepEqual(clonedV2Member.RaftAttributes, clonedV3Member.RaftAttributes) {
+			syncedV3Member := v3Member.Clone()
+			syncedV3Member.IsLearner = false
+			c.lg.Warn("Syncing member in v3store", zap.String("member", fmt.Sprintf("%+v", *syncedV3Member)))
+			unsafeHackySaveMemberToBackend(c.lg, c.be, syncedV3Member)
+			c.be.ForceCommit()
+			continue
+		}
+
+		c.lg.Error("Cannot sync member in v3store due to IsLearner not being the only field that differs between v2store amd v3store",
+			zap.String("v2member", fmt.Sprintf("%+v", *v2Member)),
+			zap.String("v3member", fmt.Sprintf("%+v", *v3Member)),
+		)
+	}
 }
 
 func (c *RaftCluster) UpdateRaftAttributes(id types.ID, raftAttr RaftAttributes, shouldApplyV3 ShouldApplyV3) {
--- a/server/etcdserver/api/membership/store.go
+++ b/server/etcdserver/api/membership/store.go
@@ -58,6 +58,22 @@ func unsafeSaveMemberToBackend(lg *zap.Logger, be backend.Backend, m *Member) er
 	return nil
 }
 
+// unsafeHackySaveMemberToBackend updates the member in a hacky way.
+// It's only used to workaround https://github.com/etcd-io/etcd/issues/19557.
+func unsafeHackySaveMemberToBackend(lg *zap.Logger, be backend.Backend, m *Member) error {
+	mkey := backendMemberKey(m.ID)
+	mvalue, err := json.Marshal(m)
+	if err != nil {
+		lg.Panic("failed to marshal member", zap.Error(err))
+	}
+
+	tx := be.BatchTx()
+	tx.LockOutsideApply()
+	defer tx.Unlock()
+	tx.UnsafePut(buckets.Members, mkey, mvalue)
+	return nil
+}
+
 // TrimClusterFromBackend removes all information about cluster (versions)
 // from the v3 backend.
 func TrimClusterFromBackend(be backend.Backend) error {
--- a/tests/e2e/ctl_v3_member_test.go
+++ b/tests/e2e/ctl_v3_member_test.go
@@ -24,7 +24,12 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"go.etcd.io/bbolt"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/client/pkg/v3/types"
+	"go.etcd.io/etcd/server/v3/datadir"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
+	"go.etcd.io/etcd/server/v3/mvcc/buckets"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
 )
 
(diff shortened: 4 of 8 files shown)


ahrtr:

Hi folks, can you spend at least 1 hour each day to move this forward?

server/etcdserver/api/membership/cluster.go
```diff
@@ -530,8 +531,13 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
 		mustUpdateMemberInStore(c.lg, c.v2store, c.members[id])
 	}
-	if c.be != nil && shouldApplyV3 {
-		unsafeSaveMemberToBackend(c.lg, c.be, c.members[id])
+	if c.be != nil {
+		if shouldApplyV3 {
+			unsafeSaveMemberToBackend(c.lg, c.be, c.members[id])
+		} else {
+			c.lg.Info("Forcibly apply member promotion", zap.String("member", fmt.Sprintf("%+v", *c.members[id])))
+			unsafeHackySaveMemberToBackend(c.lg, c.be, c.members[id])
```
siyuanfoundation, line 539: can you also do the check of the 2 conditions in `SyncLearnerPromotionIfNeeded ` to minimize the cases where it is actually called?
1. IsLearner is the only field in RaftAttributes that differs between v2store and v3store.
2. v2store.IsLearner == false && v3store.IsLearner == true.

ahrtr, line 539: We can't do that. When `shouldApplyV3` is `false`, it means v2store is lag behind of v3store, so we should NOT compare v2store and v3store.

Note learner can only be promoted once, and the operation can't be reverted, so forcibly promotion should be safe.

ahrtr: Just noticed that you removed your comment, to be clearer, we don't need to do the same check of the 2 conditions in `SyncLearnerPromotionIfNeeded`

siyuanfoundation, line 539: in theory, we only need to sync during bootstrap. for future maintenance, can you add some comments why this is needed for regular PromoteMember & CI index is ahead?

server/etcdserver/api/membership/store.go
```diff
@@ -58,6 +58,22 @@ func unsafeSaveMemberToBackend(lg *zap.Logger, be backend.Backend, m *Member) er
...
+// It's only used to workaround https://github.com/etcd-io/etcd/issues/19557.
+func unsafeHackySaveMemberToBackend(lg *zap.Logger, be backend.Backend, m *Member) error {
+	mkey := backendMemberKey(m.ID)
+	mvalue, err := json.Marshal(m)
+	if err != nil {
+		lg.Panic("failed to marshal member", zap.Error(err))
+	}
+
+	tx := be.BatchTx()
+	tx.LockOutsideApply()
```
fuweid, line 71: I checked the code and found that `PromoteMember` is only called in `applyEntries`. I tested it in my local with verify env. It panics as expected.

```bash
panic: Called inside of APPLY!

goroutine 193 [running]:
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc0003aa000, {0xc0004b2440, 0x1, 0x1})
        go.uber.org/zap@v1.17.0/zapcore/entry.go:234 +0x2dc
go.uber.org/zap.(*Logger).Panic(0x2?, {0x1156d3a?, 0x73?}, {0xc0004b2440, 0x1, 0x1})
        go.uber.org/zap@v1.17.0/logger.go:227 +0x51
go.etcd.io/etcd/server/v3/mvcc/backend.ValidateCalledOutSideApply(0xc0000149b0)
        go.etcd.io/etcd/server/v3/mvcc/backend/verify.go:45 +0x145
go.etcd.io/etcd/server/v3/mvcc/backend.(*batchTx).LockOutsideApply(0xc00034b200)
        go.etcd.io/etcd/server/v3/mvcc/backend/batch_tx.go:92 +0x26
go.etcd.io/etcd/server/v3/etcdserver/api/membership.unsafeHackySaveMemberToBackend(0xc0000149b0, {0x12f8748, 0xc000180480}, 0xc0004b02a0)
        go.etcd.io/etcd/server/v3/etcdserver/api/membership/store.go:71 +0x211
go.etcd.io/etcd/server/v3/etcdserver/api/membership.(*RaftCluster).PromoteMember(0xc000301320, 0xb5a6d2146af3b32f, 0x0)
        go.etcd.io/etcd/server/v3/etcdserver/api/membership/cluster.go:539 +0x36b
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).applyConfChange(0xc000582a08, {0x0, 0xb5a6d2146af3b32f, {0xc000046ec0, 0x3c, 0x40}, 0x6a6c95a6401bfe0c}, 0xc000136980, 0x0)
        go.etcd.io/etcd/server/v3/etcdserver/server.go:2411 +0x835
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).apply(0xc000582a08, {0xc0001a5cb0, 0x7, 0x0?}, 0xc000136980)
        go.etcd.io/etcd/server/v3/etcdserver/server.go:2251 +0x527
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).applyEntries(0xc000582a08, 0xc000136980, 0xc0000fed10?)
        go.etcd.io/etcd/server/v3/etcdserver/server.go:1465 +0x25c
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).applyAll(0xc000582a08, 0xc000136980, 0xc0000fed10)
        go.etcd.io/etcd/server/v3/etcdserver/server.go:1279 +0x57
go.etcd.io/etcd/server/v3/etcdserver.(*EtcdServer).run.func8({0x0?, 0x0?})
        go.etcd.io/etcd/server/v3/etcdserver/server.go:1151 +0x1f
go.etcd.io/etcd/pkg/v3/schedule.(*fifo).run(0xc000200000)
        go.etcd.io/etcd/pkg/v3@v3.5.19/schedule/schedule.go:157 +0xfc
created by go.etcd.io/etcd/pkg/v3/schedule.NewFIFOScheduler in goroutine 173
        go.etcd.io/etcd/pkg/v3@v3.5.19/schedule/schedule.go:70 +0x156

```

server/etcdserver/api/membership/cluster.go
fuweid: Based on my local test, we should not call this function. https://github.com/etcd-io/etcd/pull/19606/files#r1999777081

server/etcdserver/api/membership/store.go
ahrtr: > I checked the code and found that `PromoteMember` is only called in `applyEntries`. I tested it in my local with verify env. It panics as expected.

It's the reason why we need to disable the verification in `TestCtlV3PromotingLearnerOnApply`, see https://github.com/ahrtr/etcd/blob/3c9928088c5e8ed78d4110beb5d861fd2f3a3a8d/tests/e2e/ctl_v3_member_test.go#L247-L249

Which test did you see this panic? We need to disable it as well.

server/etcdserver/api/membership/cluster.go
ahrtr: I assumed that you guys already understood the intention, but obvious I was wrong.

What if v2store is lad behind, and the learner promotion hasn't been applied yet? Only syncing during bootstrap can't resolve it in that case.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -230,13 +236,29 @@ func ctlV3MemberUpdate(cx ctlCtx, memberID, peerURL string) error {
...
+func TestCtlV3PromotingLearnerOnBootstrap(t *testing.T) {
 	e2e.BeforeTest(t)
 
+	testCtlV3PromotingLearner(t, 10)
+}
+
+func TestCtlV3PromotingLearnerOnApply(t *testing.T) {
+	e2e.BeforeTest(t)
+	// We call `LockOutsideApply` in apply, so we have to disable the verification.
+	revertFunc := verify.DisableVerification()
```
serathius, line 248: Could we maybe just give an exception for stacktraces with `unsafeHackySaveMemberToBackend` ?

server/etcdserver/api/membership/cluster.go
```diff
@@ -530,8 +531,13 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	if c.v2store != nil {
 		mustUpdateMemberInStore(c.lg, c.v2store, c.members[id])
 	}
-	if c.be != nil && shouldApplyV3 {
-		unsafeSaveMemberToBackend(c.lg, c.be, c.members[id])
+	if c.be != nil {
+		if shouldApplyV3 {
+			unsafeSaveMemberToBackend(c.lg, c.be, c.members[id])
+		} else {
+			c.lg.Info("Forcibly apply member promotion", zap.String("member", fmt.Sprintf("%+v", *c.members[id])))
```
serathius, line 538: Could we only force promotion when it's needed? If `shouldApplyV3` is false and `Member.Learner` is true, means this member was not promoted in v3 and we fixing it here. Also please leave a comment explaining it.

server/embed/etcd.go
```diff
@@ -277,6 +277,9 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
 			return e, err
 		}
 	}
+
+	e.Server.SyncLearnerPromotionIfNeeded()
```
serathius, line 281: Not 100% sure about consequences of calling sync here, by proposing to run it during bootstrap I meant within `bootstrap.go`

tests/e2e/ctl_v3_member_test.go
ahrtr: good call & updated

server/etcdserver/api/membership/store.go
ahrtr: added a comment

server/embed/etcd.go
serathius: How do you think this change will interact with corrupt check? Even if we do the sync before corrupt check on one member, next member will start with corrupt check before sync.

ahrtr: > I don't know if we really don't run any goroutines in background at this point, and proving so might be very hard.

We call `SyncLearnerPromotionIfNeeded` before starting the raft loop, also before starting to serve client & peer requests. It's safe. What's your concern here?



> How do you think this change will interact with corrupt check?

If there is data corrupt, we should leave it as it's.

server/etcdserver/api/membership/store.go
ahrtr: let me know if you still can reproduce/see the panic

server/embed/etcd.go
fuweid: I am thinking if we still need this migration.

I found there is publish API call during starting. Since we already allow to override member information, this publish can help us migration (if I understand it correctly).

https://github.com/etcd-io/etcd/blob/3c65dfa9d3f85fbfff19f5f0e0f57a67ecefa48d/server/etcdserver/server.go#L2132

* If there is snapshot, the v2store contains valid data. UpdateAtribute will override existing invalid one.
* If there is not snapshot, WAL replay can fix this issue as well.

ahrtr: @serathius please let me know if you still have any concern on my previous comment. I think we should be all good now.



> I am thinking if we still need this migration.
> 
> I found there is publish API call during starting. Since we already allow to override member information, this publish can help us migration (if I understand it correctly).

@fuweid I don't think we are on the same page. The publish logic is about attribute, what we are resolving/talking about in this PR (and issue 19557) is about IsLearner, which is part of RaftAttribute.

ahrtr: > override member information

Yes, this is true. thx for pointing out this. 

Although the publish logic is only supposed to update attribute, but the implementation just overwrites the whole member info into v3store (bbolt), so it can automatically fix the issues which are affected by https://github.com/etcd-io/etcd/issues/19557 on top of the patch https://github.com/etcd-io/etcd/pull/19563 

Just raised another PR with only e2e test cases https://github.com/etcd-io/etcd/pull/19629, which I believe is better than https://github.com/etcd-io/etcd/pull/19627

ahrtr: For release-3.6, we don't need any further validation (i.e. requiring 3.5.20+ before upgrading to 3.6) any more, reasons:
- If there are 2 or more (already promoted) learners, then etcd will crash anyway.
- If there is only 1 (already promoted) learner, then etcd will automatically fix it due to the same reason as mentioned above. 

So for release-3.6, we only need to add two e2e test cases:
- the kubeadm style upgrade case
- verify etcd is able to automatically fix the issue when upgrading to 3.6 for the one learner case, similar to https://github.com/etcd-io/etcd/pull/19629

tests/e2e/ctl_v3_member_test.go
```diff
@@ -231,30 +236,145 @@ func ctlV3MemberUpdate(cx ctlCtx, memberID, peerURL string) error {
...
-	etcdctl := epc.Procs[0].Etcdctl(e2e.ClientNonTLS, false, false)
-	err = etcdctl.Put("foo", "bar")
+		return b.Put(mkey, mvalue)
+	})
 	require.NoError(t, err)
+}
 
-	t.Logf("Promoting the learner %x", learnerID)
-	_, err = etcdctl.MemberPromote(learnerID)
+func ensureAllMembersAreVotingMembers(t *testing.T, etcdctl *e2e.Etcdctl) {
```
siyuanfoundation, line 374: can you verify AllMembersAreVotingMembers in both v2store and v3 directly from the backend?

tests/e2e/ctl_v3_member_test.go
```diff
@@ -231,30 +236,145 @@ func ctlV3MemberUpdate(cx ctlCtx, memberID, peerURL string) error {
...
+			snapshotCount: 10,
+		},
+		{
+			name:          "during apply",
+			snapshotCount: 0,
+		},
+		{
+			name:             "during apply",
+			snapshotCount:    0,
+			promotionSuccess: true,
```
siyuanfoundation, line 255: rename to `promotionWrittenToV3`?

tests/e2e/ctl_v3_member_test.go
```diff
@@ -231,30 +236,145 @@ func ctlV3MemberUpdate(cx ctlCtx, memberID, peerURL string) error {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			t.Log("Create a single node etcd cluster")
+			cfg := e2e.NewConfigNoTLS()
+			cfg.BasePeerScheme = "unix"
+			cfg.ClusterSize = 1
```
siyuanfoundation, line 265: can you set `cfg.InitialCorruptCheck = true`?

Issue #18972, opened 2024-11-30
Use testify instead of native testing library


mmorel-35:
### What would you like to be added?

Testify is more concise and have a lot functions that holds default messages that needs to be explicitly wrote with native testing package.

The completion track is as follows:

- [ ] #18971
- [x] ~~#18970~~
- [x] ~~#18969~~
- [x] ~~#18968~~
- [x] ~~#18967~~
- [x] ~~#18966~~
- [ ] #18860 
- [x] ~~#18823~~
- [x] ~~#18821~~
- [ ] #18820
- [x] ~~#18782~~
- [x] ~~#18753~~
- [x] ~~#18752~~
- [x] ~~#18705~~

ahrtr:
All done. Thanks @mmorel-35

PR #19612, opened 2025-03-15
chore: use testify instead of testing in server/etcdmain

mmorel-35:
#### Description

This uses testify instead of testing for t.Fatal or t.Error calls in server/etcdmain package

Related to #18972


jmhbnz (approved):
LGTM - Thanks @mmorel-35

Issue #19579, opened 2025-03-12
Robustness history sets incorrect return time for a failed TXN (compaction)


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

During robustness test meeting we investigated a linearization failure https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-etcd-robustness-main-amd64/1898807228299743232

We found that:
* There is failed TXN for K8s compaction that doesn't linearize
* Watch observed that this TXN has revision 2830
* If patching is enabled the return time of TXN is much earlier than any rev 2830.
* If patching is disabled TXN is used much earlier and there is a gap in revisions between 2829 and 2831 that is not filled.
* When we investigated the TXN return time we noticed that is set in the report to a pretty low value.
* When the TXN return time was increased the linearization started passing.

Context
* For failed TXN robustness test don't know when the transaction takes place, it could be after the response was received. Imagine a client connection error, at the beggining of the transaction. Client might receive it much before transaction is commited. 
* To handle such cases robustness tests fake return time by setting it around the end of the test, however for the TXN it seemed it was much earlier. Like it was based on client receiving error.
* Robustness test preserves return time for Read operations just for visualization, however we haven't seen why this would happen here.  

TODO:
* Investigate why the return time was set so low for this transcation.
* Propose fix for appending failed request to history. 
    * Existing logic is pretty complicated. For simplicity we might just want to just set it to MaxInt if it works and doesn't disrupt visualization.

### What did you expect to happen?

No flakes in robustness test

### How can we reproduce it (as minimally and precisely as possible)?

Use report from https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-etcd-robustness-main-amd64/1898807228299743232

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

[...]

PR #19626, opened 2025-03-18
Set return time to max value of int64 for failed requests during robustness test validation phase

henrybear327:
The failed requests will now look something like the following, extending till the end of the right.

Screenshot 2025-03-19 at 12 21 01 AM

Reference:
- https://github.com/etcd-io/etcd/issues/19579


serathius:
That's not what I had in mind, we should do it during history collection, not history patching.

Issue #19604, opened 2025-03-14
Workflow `pull-etcd-markdown-lint` failed


ahrtr:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

```
make: go: No such file or directory
./scripts/markdown_diff_lint.sh
make: go: No such file or directory
./scripts/test_lib.sh: line 7: go: command not found
must be run from 'go.etcd.io/etcd/v3' module directory
make: *** [Makefile:224: markdown-diff-lint] Error 2[55](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19603/pull-etcd-markdown-lint/1900534475683532800#1:build-log.txt%3A55)
```

Refer to https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19603/pull-etcd-markdown-lint/1900534475683532800

### What did you expect to happen?

No any error.

Also why we do not see the error on main post merging PRs? I expect that we can discover such error immediately instead of until next PR.

### How can we reproduce it (as minimally and precisely as possible)?

.

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```Shell

```

ivanvc:
We're still working on making this work. It should be set as optional and not with always run. Let me check.

/assign

joshjms:
[...]

PR #19631, opened 2025-03-19
Split `scripts/test_lib.sh`

joshjms:
From #19604 , the `markdown_diff_lint.sh` fails because of:

```bash
if [[ "$(go list)" != "${ROOT_MODULE}/v3" ]]; then
  echo "must be run from '${ROOT_MODULE}/v3' module directory"
  exit 255
fi
```

While the `markdown_diff_lint.sh` does not require `go`. The idea is to break `test_lib.sh` into another file that only contains basic functionalities like logging and getting relative path (i.e. `test_utils.sh`). Both `test_lib.sh` and `markdown_diff_lint.sh` will `source` from `test_utils.sh`. 

Solves #18059 

### DIFF
--- a/scripts/markdown_diff_lint.sh
+++ b/scripts/markdown_diff_lint.sh
@@ -3,10 +3,10 @@
 # Usage: ./markdown_lint.sh
 
 
-# We source ./scripts/test_lib.sh, it sets the log functions and color variables.
-source ./scripts/test_lib.sh
+# We source ./scripts/test_utils.sh, it sets the log functions and color variables.
+source ./scripts/test_utils.sh
 
-# When we source ./scripts/test_lib.sh, it has the line set -u which treats unset variables as errors.
+# When we source ./scripts/test_utils.sh, it has the line set -u which treats unset variables as errors.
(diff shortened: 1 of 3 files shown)


ivanvc:
Thanks, @joshjms. This looks good, but I'm not entirely sold on the name of the file `test_utils.sh`. I'm trying to think of a better name that is more specific to what it does, but I'm drawing a blank. Do you have any suggestions?

joshjms:
Hmm, I am honestly quite conflicted as well. Since `markdown-lint.sh` only uses the logs, maybe we can move the relative path function back to test_lib.sh and name the file test_logging.sh or something. Otherwise do you think `test_commons.sh` work? I'm not good at deciding file names, I'll trust your judgement.

ahrtr (approved):

ahrtr:
cc @fuweid @serathius

ivanvc (approved):
LGTM. We can rename the file later to something else, maybe `test_logging.sh`. Thanks, @joshjms.

fuweid (approved):

PR #19633, opened 2025-03-20
Patch the return time during validation with MaxInt64 in robustness test

henrybear327:
Instead of marking the return time with -1 during operation data collection phase, we patch the failed request's response time during validation so we get to preserve the original data, and we get to avoid the problem we see mentioned in issue #19579.

Reference: https://github.com/etcd-io/etcd/issues/19579

### DIFF
--- a/tests/robustness/model/history.go
+++ b/tests/robustness/model/history.go
@@ -290,8 +290,6 @@ func (h *AppendableHistory) appendFailed(request EtcdRequest, start, end time.Du
 	}
 	isRead := request.IsRead()
 	if !isRead {
-		// Failed writes can still be persisted, setting -1 for now as don't know when request has took effect.
-		op.Return = -1
 		// Operations of single client needs to be sequential.
 		// As we don't know return time of failed operations, all new writes need to be done with new stream id.
 		h.streamID = h.idProvider.NewStreamID()
@@ -300,7 +298,7 @@ func (h *AppendableHistory) appendFailed(request EtcdRequest, start, end time.Du
 }
 
 func (h *AppendableHistory) append(op porcupine.Operation) {
-	if op.Return != -1 && op.Call >= op.Return {
+	if op.Call >= op.Return {
 		panic(fmt.Sprintf("Invalid operation, call(%d) >= return(%d)", op.Call, op.Return))
 	}
 	if len(h.operations) > 0 {
@@ -488,34 +486,8 @@ func (h History) Len() int {
 
 func (h History) Operations() []porcupine.Operation {
 	operations := make([]porcupine.Operation, 0, len(h.operations))
-	maxTime := h.lastObservedTime()
-	for _, op := range h.operations {
-		// Failed requests don't have a known return time.
-		if op.Return == -1 {
-			// Simulate Infinity by using last observed time.
-			op.Return = maxTime + time.Second.Nanoseconds()
-		}
-		operations = append(operations, op)
-	}
+	operations = append(operations, h.operations...)
 	return operations
-}
-
-func (h History) lastObservedTime() int64 {
-	var maxTime int64
-	for _, op := range h.operations {
-		if op.Return == -1 {
-			// Collect call time from failed operations
-			if op.Call > maxTime {
-				maxTime = op.Call
-			}
-		} else {
-			// Collect return time from successful operations
-			if op.Return > maxTime {
-				maxTime = op.Return
-			}
-		}
-	}
-	return maxTime
 }
 
 func (h History) MaxRevision() int64 {
--- a/tests/robustness/validate/operations.go
+++ b/tests/robustness/validate/operations.go
@@ -17,6 +17,7 @@ package validate
 import (
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/anishathalye/porcupine"
@@ -53,6 +54,8 @@ func validateLinearizableOperationsAndVisualize(
 	operations []porcupine.Operation,
 	timeout time.Duration,
 ) (results Results) {
+	operations = patchResponseTimeOfFailedOperations(operations)
+
 	lg.Info("Validating linearizable operations", zap.Duration("timeout", timeout))
 	start := time.Now()
 	result, info := porcupine.CheckOperationsVerbose(model.NonDeterministicModel, operations, timeout)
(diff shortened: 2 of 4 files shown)


tests/robustness/model/history.go
```diff
@@ -488,36 +489,17 @@ func (h History) Len() int {
 
 func (h History) Operations() []porcupine.Operation {
 	operations := make([]porcupine.Operation, 0, len(h.operations))
-	maxTime := h.lastObservedTime()
 	for _, op := range h.operations {
 		// Failed requests don't have a known return time.
 		if op.Return == -1 {
-			// Simulate Infinity by using last observed time.
-			op.Return = maxTime + time.Second.Nanoseconds()
+			// TODO(henrybear327): add reasoning here
```
serathius, line 495: You can just leave previous comment `// Simulate infinity for failed requests`

henrybear327: Done as requested!

henrybear327:
@serathius for the record, the robustness test that I ran locally over lunch passed. 

As discussed on Slack, I will now:
- select some of the past issues, e.g. `test-robustness-issue14370` to validate if it also still works
- if the visualization looks ok by force-generating it

If all is good, we modify the logic to not set the return value to `-1`, but during validation when there is an error contained in the operation, we modify it to max int so we can preserve the original response time.

henrybear327:
`test-robustness-issue18089` and `test-robustness-issue19179` both are successfully reproduced.

henrybear327:
One force-generated report

[TestRobustnessExploratory_EtcdHighTraffic_ClusterOfSize3_blackholePeerNetwork.zip](https://github.com/user-attachments/files/19374807/TestRobustnessExploratory_EtcdHighTraffic_ClusterOfSize3_blackholePeerNetwork.zip)

tests/robustness/validate/patch_history.go
```diff
@@ -239,6 +241,17 @@ func uniquePutReturnTime(allOperations []porcupine.Operation, reports []report.C
 	return earliestReturnTime
 }
 
+// Failed writes can still be persisted, but we don't know when request has took effect, so we override it with MaxInt64
+func patchFailedOperations(operations *[]porcupine.Operation) {
```
serathius, line 245: Please return copy.

tests/robustness/validate/patch_history.go
```diff
@@ -29,7 +30,9 @@ func patchLinearizableOperations(reports []report.ClientReport, persistedRequest
 	persistedPutCount := countPersistedPuts(persistedRequests)
 	clientPutCount := countClientPuts(reports)
 	putReturnTime := uniquePutReturnTime(allOperations, reports, persistedRequests, clientPutCount)
-	return patchOperations(allOperations, putRevision, putReturnTime, clientPutCount, persistedPutCount)
+	patchedOperations := patchOperations(allOperations, putRevision, putReturnTime, clientPutCount, persistedPutCount)
```
serathius, line 33: Please don't do that for patching, but for validation. This is required part of validation due to how pourcupine work, while patching is just optimization. I would want to have it separate process before patching as part of validation.

henrybear327: Moved into the validation function

tests/robustness/model/history.go
```diff
@@ -488,36 +486,10 @@ func (h History) Len() int {
 
 func (h History) Operations() []porcupine.Operation {
 	operations := make([]porcupine.Operation, 0, len(h.operations))
-	maxTime := h.lastObservedTime()
```
henrybear327, line 491: @serathius Because of this removal, the unit tests would have to be updated, due to [this pre-condition](https://github.com/etcd-io/etcd/blob/main/tests/robustness/model/history.go#L312) failing.

But if I am not mistaken, each client is sending the requests sequentially, so the test should be ok to be updated this way, right?

tests/robustness/validate/patch_history_test.go
```diff
@@ -152,37 +152,37 @@ func TestPatchHistory(t *testing.T) {
...
-			},
-			persistedRequest: []model.EtcdRequest{
-				putRequestWithLease("key", "value", 123),
-			},
-			watchOperations: watchResponse(3, putEvent("key", "value", 2)),
-			expectedRemainingOperations: []porcupine.Operation{
-				{Return: 3, Output: model.MaybeEtcdResponse{Persisted: true, PersistedRevision: 2}},
-			},
-		},
+		// {
```
henrybear327, line 173: I am still trying to understand why this test is failing, so I commented this out.

henrybear327: See https://github.com/etcd-io/etcd/pull/19633#discussion_r2007754157

tests/robustness/model/history.go
serathius: Yes, updating unit tests here makes sense.

tests/robustness/validate/operations.go
```diff
@@ -53,6 +54,8 @@ func validateLinearizableOperationsAndVisualize(
 	operations []porcupine.Operation,
 	timeout time.Duration,
 ) (results Results) {
+	operations = patchResponseTimeOfFailedOperations(operations)
```
serathius, line 57: This should happen before we patch operations, because then patching will not have opportunity to modify the return time.

tests/robustness/validate/patch_history.go
```diff
@@ -28,7 +28,7 @@ func patchLinearizableOperations(reports []report.ClientReport, persistedRequest
 	putRevision := putRevision(reports)
 	persistedPutCount := countPersistedPuts(persistedRequests)
 	clientPutCount := countClientPuts(reports)
-	putReturnTime := uniquePutReturnTime(allOperations, reports, persistedRequests, clientPutCount)
+	putReturnTime := uniquePutReturnTime(allOperations, reports, clientPutCount)
```
serathius, line 31: Please revert

tests/robustness/validate/patch_history.go
```diff
@@ -170,10 +172,14 @@ func uniquePutReturnTime(allOperations []porcupine.Operation, reports []report.C
 				if count := clientPutCount[kv]; count > 1 {
 					continue
 				}
+				if resp.Error != "" {
+					failedRequest[kv] = true
+				}
 				if returnTime, ok := earliestReturnTime[kv]; !ok || returnTime > op.Return {
+					// The time that the response is received
+					// This doesn't mean anything for the failed request, as the request might have actually succeeded (e.g. connection dropped, etc.)
 					earliestReturnTime[kv] = op.Return
```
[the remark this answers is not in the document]
serathius: Not following. Please move unrelated changes to separate PR.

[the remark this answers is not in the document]
henrybear327: My bad, my local branch is screwed up when rebasing. Sorry about this mess...

henrybear327:
Discussed offline, we will start a new PR with minimal changes and improved append validation rules.

PR #19602, opened 2025-03-14
[release-3.5] Fix the issue that learner promotion command doesn't support json output

ahrtr:
cc @fuweid @ivanvc @jmhbnz @serathius @siyuanfoundation 

Note main & release-3.6 do not have this issue.

PR #19636, opened 2025-03-21
[release-3.6] Auto sync members in v3store is IsLearner differs between v2 and v3 store

ahrtr:
release-3.6 isn't able to automatically fix the issues which have already been affected by https://github.com/etcd-io/etcd/issues/19557.

Also adding the validation has flaws as mentioned in https://github.com/etcd-io/etcd/issues/19557#issuecomment-2741425899

Eventually I tried to add similar auto sync in 3.6 as what we did for 3.5 in https://github.com/etcd-io/etcd/pull/19606 

This is the only way to avoid all confusion and reduce of user experience.

cc @fuweid @serathius @siyuanfoundation

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -524,23 +524,69 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 		isLearner.Set(0)
 	}
 
-	if c.be != nil && shouldApplyV3 {
-		c.members[id].RaftAttributes.IsLearner = false
-		c.updateMembershipMetric(id, true)
-		c.be.MustSaveMemberToBackend(c.members[id])
+	if c.be != nil {
+		m := c.members[id]
+		if shouldApplyV3 {
+			m.RaftAttributes.IsLearner = false
+			c.updateMembershipMetric(id, true)
+			c.be.MustSaveMemberToBackend(m)
 
-		c.lg.Info(
-			"promote member",
-			zap.String("cluster-id", c.cid.String()),
-			zap.String("local-member-id", c.localID.String()),
-			zap.String("promoted-member-id", id.String()),
-		)
+			c.lg.Info(
+				"promote member",
+				zap.String("cluster-id", c.cid.String()),
+				zap.String("local-member-id", c.localID.String()),
+				zap.String("promoted-member-id", id.String()),
+			)
+		} else {
+			// Workaround the issues which have already been affected by
+			// https://github.com/etcd-io/etcd/issues/19557. The learner
+			// promotion request had been applied to v3store, but not saved
+			// to v2snapshot yet when in 3.5. Once upgrading to 3.6, the
+			// patch here ensure the issue can be automatically fixed.
+			if m.IsLearner {
+				m.RaftAttributes.IsLearner = false
+				c.lg.Info("Forcibly apply member promotion request", zap.String("member", fmt.Sprintf("%+v", *m)))
+				c.be.MustHackySaveMemberToBackend(m)
+			} else {
+				c.lg.Info(
+					"ignore already promoted member",
+					zap.String("cluster-id", c.cid.String()),
+					zap.String("local-member-id", c.localID.String()),
+				)
+			}
+		}
 	} else {
 		c.lg.Info(
-			"ignore already promoted member",
+			"ignore already promoted member due to backend being nil",
 			zap.String("cluster-id", c.cid.String()),
 			zap.String("local-member-id", c.localID.String()),
 		)
+	}
+}
+
+// SyncLearnerPromotionIfNeeded provides a workaround solution to fix the issues
+// which have already been affected by https://github.com/etcd-io/etcd/issues/19557.
+func (c *RaftCluster) SyncLearnerPromotionIfNeeded() {
+	c.Lock()
+	defer c.Unlock()
+
+	v2Members, _ := membersFromStore(c.lg, c.v2store)
+	v3Members, _ := c.be.MustReadMembersFromBackend()
+
+	for id, v3Member := range v3Members {
+		v2Member, ok := v2Members[id]
+		if !ok {
+			// This isn't an error. The conf change on the member hasn't been saved to the v2 snapshot yet.
+			c.lg.Info("Detected member only in v3store but missing in v2store", zap.String("member", fmt.Sprintf("%+v", *v3Member)))
+			continue
+		}
+
+		if !v2Member.IsLearner && v3Member.IsLearner {
+			syncedV3Member := v3Member.Clone()
+			syncedV3Member.IsLearner = false
+			c.lg.Warn("Syncing member in v3store", zap.String("member", fmt.Sprintf("%+v", *syncedV3Member)))
+			c.be.MustHackySaveMemberToBackend(syncedV3Member)
+		}
 	}
 }
 
--- a/server/etcdserver/bootstrap.go
+++ b/server/etcdserver/bootstrap.go
@@ -448,6 +448,11 @@ func (c *bootstrappedCluster) Finalize(cfg config.ServerConfig, s *bootstrappedS
 	}
 	c.cl.SetStore(s.st)
 	c.cl.SetBackend(schema.NewMembershipBackend(cfg.Logger, s.backend.be))
+
+	// Workaround the issues which have already been affected
+	// by https://github.com/etcd-io/etcd/issues/19557.
+	c.cl.SyncLearnerPromotionIfNeeded()
+
 	if s.wal.haveWAL {
 		c.cl.Recover(api.UpdateCapability)
 		if c.databaseFileMissing(s) {
--- a/tests/e2e/ctl_v3_member_test.go
+++ b/tests/e2e/ctl_v3_member_test.go
@@ -29,9 +29,17 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"go.etcd.io/bbolt"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/client/pkg/v3/fileutil"
+	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/pkg/v3/expect"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
+	"go.etcd.io/etcd/server/v3/storage/datadir"
+	"go.etcd.io/etcd/server/v3/storage/schema"
+	"go.etcd.io/etcd/tests/v3/framework/config"
 	"go.etcd.io/etcd/tests/v3/framework/e2e"
+	"go.etcd.io/etcd/tests/v3/framework/testutils"
 )
 
 func TestCtlV3MemberList(t *testing.T)        { testCtl(t, memberListTest) }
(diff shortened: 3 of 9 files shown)


tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,230 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
 }
+
+// TestCtlV3PromotingLearner tests whether etcd can automatically fix
+// the issue caused by https://github.com/etcd-io/etcd/issues/19557
+// when upgrading from 3.5 to 3.6.
+func TestCtlV3PromotingLearner(t *testing.T) {
+	testCases := []struct {
+		name                  string
+		snapshotCount         uint64
+		writeToV3StoreSuccess bool
```
fuweid, line 323: I think we don't need this field.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,230 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+
+			if tc.snapshotCount != 0 {
+				t.Logf("Write %d keys to trigger a snapshot", tc.snapshotCount)
+				for i := 0; i < int(tc.snapshotCount); i++ {
+					err := epc.Etcdctl().Put(ctx, fmt.Sprintf("key_%d", i), fmt.Sprintf("value_%d", i), config.PutOptions{})
+					require.NoError(t, err)
+				}
+			}
+
+			if tc.writeToV3StoreSuccess {
```
fuweid, line 365: remove this

tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,230 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+		}
+	}
+
+	t.Log("Ensure all members are voting members from user perspective")
+	ensureAllMembersAreVotingMembers(t, epc)
+
+	return epc, promotedMembers
+}
+
+func mustSaveMemberIntoBbolt(t *testing.T, dataDir string, protoMember *etcdserverpb.Member) {
```
fuweid, line 475:
```diff
+func mustTweakMemberInfoByBbolt(t *testing.T, dataDir string, memberID uint64, tweakFn func(*membership.Member)) {
+       dbPath := datadir.ToBackendFileName(dataDir)
+       db, err := bbolt.Open(dbPath, 0666, nil)
+       require.NoError(t, err)
+       defer func() {
+               require.NoError(t, db.Close())
+       }()
+
+       err = db.Update(func(tx *bbolt.Tx) error {
+               mKey := []byte(types.ID(memberID).String())
+
+               b := tx.Bucket(schema.Members.Name())
+               mValue := b.Get(mKey)
+               if mValue == nil {
+                       return fmt.Errorf("member(%x) not found", memberID)
+               }
+
+               m := &membership.Member{}
+               if err := json.Unmarshal(mValue, m); err != nil {
+                       return fmt.Errorf("failed to unmarshal (%s) into member: %w", string(mValue), err)
+               }
+
+               tweakFn(m)
+
+               mValue, err = json.Marshal(m)
+               if err != nil {
+                       return fmt.Errorf("failed to marshal into json: %w", err)
+               }
+               return b.Put(mKey, mValue)
+       })
+       require.NoError(t, err)
+}
```

I am thinking introduce tweakFn to apply patch instead of passing whole object.

tests/e2e/ctl_v3_member_test.go
ahrtr: This is to verify the normal case. See line 390

```
if tc.writeToV3StoreSuccess {
	e2e.AssertProcessLogs(t, proc, "ignore already promoted member")
}
```

tests/e2e/ctl_v3_member_test.go
ahrtr: Actually there isn't much difference. The current implementation reads the member (promoted members) outside the function. 

Also I am not expecting this will be a common function to be reused.

tests/e2e/ctl_v3_member_test.go
ahrtr: pls see my below comment. If there is no further comment, let's resolve this and next one.

server/etcdserver/api/membership/cluster.go
```diff
@@ -524,26 +524,72 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
+
+	for id, v3Member := range v3Members {
+		v2Member, ok := v2Members[id]
+		if !ok {
+			// This isn't an error. The conf change on the member hasn't been saved to the v2 snapshot yet.
+			c.lg.Info("Detected member only in v3store but missing in v2store", zap.String("member", fmt.Sprintf("%+v", *v3Member)))
+			continue
+		}
+
+		if !v2Member.IsLearner && v3Member.IsLearner {
```
siyuanfoundation, line 584: can you check that `IsLearner` is the only different field in `RaftAttributes`?

tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,229 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+	}{
+		{
+			name:          "create snapshot after learner promotion which is not saved to v3store",
+			snapshotCount: 10,
+		},
+		{
+			name:          "no snapshot and learner promotion is not saved to v3store",
+			snapshotCount: 0,
+		},
+		/*{
```
siyuanfoundation, line 333: why is this commented out?

ahrtr: because 3.5.20 isn't released yet, the case will fail for now. Will uncomment it once 3.5.20 is out

ahrtr: uncommented

server/etcdserver/api/membership/cluster.go
ahrtr: We don't care other fields.

tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,229 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+				require.NoError(t, epc.Stop())
+
+				t.Log("Manually changing the already promoted member to a learner again in the v3store of all members to simulate the issue https://github.com/etcd-io/etcd/issues/19557")
+				promotedMembers[0].IsLearner = true
+				for i := 0; i < len(epc.Procs); i++ {
+					mustSaveMemberIntoBbolt(t, epc.Procs[i].Config().DataDirPath, promotedMembers[0])
+				}
+			}
+
+			t.Log("Upgrading the cluster")
```
fuweid, line 378: we can move this assignment to the beginning of the test 

```go
func TestCtlV3PromotingLearner(t *testing.T) {
  -- here
```

tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,229 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+		})
+	}
+}
+
+func mustCreateNewClusterByPromotingMembers(t *testing.T, clusterVersion e2e.ClusterVersion, clusterSize int, opts ...e2e.EPClusterOption) (*e2e.EtcdProcessCluster, []*etcdserverpb.Member) {
+	require.GreaterOrEqualf(t, clusterSize, 1, "clusterSize must be at least 1")
+
+	ctx := context.Background()
+
+	t.Logf("Creating new etcd cluster - version: %s, clusterSize: %v", clusterVersion, clusterSize)
```
fuweid, line 423: the `clusterVersion` is empty when it's current version. we can remove version information from log

fuweid (approved):
LGTM

tests/e2e/ctl_v3_member_test.go
ahrtr: yes, it's true. It's fine although not perfect. We know it's current version when it's empty string.

We can try to resolve it separately, i.e. https://github.com/etcd-io/etcd/pull/19640

tests/e2e/ctl_v3_member_test.go
ahrtr: not quite get the point. I intentionally print the log right before we upgrade the cluster.

Changed the test name to `TestClusterUpgradeWithPromotingLearner`

tests/e2e/ctl_v3_member_test.go
ahrtr: > we can remove version information from log

I tend to keep it as it's important info for debug.

tests/e2e/ctl_v3_member_test.go
fuweid: sorry I comment on wrong line. I mean these four lines

```go
	                currentVersion, err := e2e.GetVersionFromBinary(e2e.BinPath.Etcd)
			require.NoError(t, err)
			lastVersion, err := e2e.GetVersionFromBinary(e2e.BinPath.EtcdLastRelease)
			require.NoError(t, err)
```

ahrtr: I intentionally added them right before where they being used.

But overall seems not a big deal.

ahrtr:
cc @siyuanfoundation @serathius

serathius:
> release-3.6 isn't able to automatically fix the issues which have already been affected by https://github.com/etcd-io/etcd/issues/19557.

What do you mean by that?

server/etcdserver/api/membership/cluster.go
```diff
@@ -524,26 +524,72 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
...
+		} else {
+			// Workaround the issues which have already been affected by
+			// https://github.com/etcd-io/etcd/issues/19557. The learner
+			// promotion request had been applied to v3store, but not saved
+			// to v2snapshot yet when in 3.5. Once upgrading to 3.6, the
+			// patch here ensure the issue can be automatically fixed.
+			if m.IsLearner {
+				m.RaftAttributes.IsLearner = false
+				c.lg.Info("Forcibly apply member promotion request", zap.String("member", fmt.Sprintf("%+v", *m)))
+				c.be.MustSaveMemberToBackend(m)
```
serathius, line 549: MustSaveMemberToBackend will use `LockInsideApply` which will flush the CI. I don't remember what is state of CI when `shouldApplyV3`, but I assume it's pointing to the raftindex of the promote request. 

I think this is the place to call `MustHackySaveMemberToBackend`

server/etcdserver/bootstrap.go
```diff
@@ -448,6 +448,11 @@ func (c *bootstrappedCluster) Finalize(cfg config.ServerConfig, s *bootstrappedS
 	}
 	c.cl.SetStore(s.st)
 	c.cl.SetBackend(schema.NewMembershipBackend(cfg.Logger, s.backend.be))
+
+	// Workaround the issues which have already been affected
+	// by https://github.com/etcd-io/etcd/issues/19557.
+	c.cl.SyncLearnerPromotionIfNeeded()
```
serathius, line 454: Like this place

tests/e2e/ctl_v3_member_test.go
```diff
@@ -304,3 +312,230 @@ func TestRemoveNonExistingMember(t *testing.T) {
...
+			name:          "create snapshot after learner promotion which is not saved to v3store",
+			snapshotCount: 10,
+		},
+		{
+			name:          "no snapshot and learner promotion is not saved to v3store",
+			snapshotCount: 0,
+		},
+		{
+			name:                  "no snapshot and learner promotion is saved to v3store",
+			snapshotCount:         0,
```
serathius, line 335: Why not test with `shapshotCount: 10` and `writeToV3StoreSuccess: true,`

serathius:
Left couple of comments apart of that LGTM.

server/etcdserver/api/membership/cluster.go
ahrtr: Conceptually, you are right, we should call `MustHackySaveMemberToBackend` here. But actually the result is the same. When `shouldApplyV3` is false, even calling `LockInsideApply` won't advance the CI.

Anyway, changed to `MustHackySaveMemberToBackend`, thx.

tests/e2e/ctl_v3_member_test.go
ahrtr: We were even planning to remove `writeToV3StoreSuccess`, the only reason is to test the normal case, ensure it goes into "`ignore already promoted member`" branch.

ahrtr: There is nothing to expect when `shapshotCount: 10` and `writeToV3StoreSuccess: true`.

ahrtr:
> > release-3.6 isn't able to automatically fix the issues which have already been affected by #19557.
> 
> What do you mean by that?

release-3.5 is able to automatically fix the already affected issues after https://github.com/etcd-io/etcd/pull/19602. But release-3.6 isn't automaticaly fix it, see https://github.com/etcd-io/etcd/issues/19557#issuecomment-2741425899

It's exactly the reason why we need this PR.

server/etcdserver/api/membership/cluster.go
serathius: Thanks for confirming and updating code.

tests/e2e/ctl_v3_member_test.go
serathius: Aaa, ok I mistook the conditions. Thought it's reversed. Now it makes sense. Thanks.

Issue #19635, opened 2025-03-21
dropped internal Raft message since sending buffer is full (overloaded network)


rahulbapumore:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

[logs.txt](https://github.com/user-attachments/files/19391626/logs.txt)

We are facing below error messages and etcd is restarting because it could not maintain quorum.
`2025-03-18T12:21:38.350+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.451+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.551+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.651+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.751+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.850+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.951+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:38.972+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:39.041+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:39.151+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:39.251+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:39.351+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:39.451+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
2025-03-18T12:21:39.550+01:00 dropped internal Raft message since sending buffer is full (overloaded network)
[...]

PR #19663, opened 2025-03-24
Dynamically assign pipeline buffer size

kumarlokesh:
Attempts to resolve issue https://github.com/etcd-io/etcd/issues/19635.

Dynamically set the configuration `pipeline-buffer-size`, instead of using
hard-coded value 64.
Fallback to default value of 64 if configuration isn't provided.

### DIFF
--- a/server/embed/config.go
+++ b/server/embed/config.go
@@ -577,6 +577,11 @@ type Config struct {
 	// be refined to mlock in-use area of bbolt only.
 	MemoryMlock bool `json:"memory-mlock"`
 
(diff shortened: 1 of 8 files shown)


jmhbnz:
Discussed during sig-etcd triage meeting - Hey @kumarlokesh thanks for raising this idea, can you please provide some performance benchmarks outlining the value of having this field dynamically configurable?  This default has been in place for a long time without concern so we need to understand if this pr will make any meaningful improvement before exposing another user configurable field.

jmhbnz:
Closing pr, underlying issue has been closed

Issue #19670, opened 2025-03-26
[release-3.6] Hiding client/v2 causes dependency problem for applications which are depending on `go.etcd.io/etcd/server/v3`


ahrtr:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

In release-3.6, the client/v2 was changed into an internal package in https://github.com/etcd-io/etcd/pull/15591. But the problem is that client/v2 is still used by v2discovery in etcd/server. So it's causing the dependency problem (see below) for any applications which are using `go.etcd.io/etcd/server/v3`

```
$ go get go.etcd.io/etcd/server/v3@v3.6.0-rc.2
go: downloading go.etcd.io/etcd/server/v3 v3.6.0-rc.2
go: go.etcd.io/etcd/server/v3@v3.6.0-rc.2 requires go.etcd.io/etcd/client/v2@v2.306.0-rc.2: missing go.etcd.io/etcd/client/go.mod and .../v2/go.mod at revision client/v2.306.0-rc.2
```

Two possible solutions that I can think of for now:

- [preferred] either revert the change (do not hide client/v2 in 3.6);
- decommission the v2discovery in release-3.6. But it may break the existing users which are still using the deprecated v2discovery.

cc @liggitt @fuweid @serathius @ivanvc 

### What did you expect to happen?

It shouldn't break any user applications

### How can we reproduce it (as minimally and precisely as possible)?

see above

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

[...]

PR #19671, opened 2025-03-26
[release-3.6] Revert client/internal/v2 to client/v2

ahrtr:
The client/v2 is still used by v2discovery in etcdserver. If we only change client/v2 into an internal package, then all applications which depend on etcdserve module will have dependency issue.

Please see the issue 19670 to get more details.

Note that we don't need make this change in main, because we will remove all v2 stuff in main/3.7 soon.

cc @fuweid @ivanvc @serathius @liggitt


ahrtr:
Link to https://github.com/etcd-io/etcd/issues/19670

serathius (approved):

fuweid (approved):

serathius:
Thanks for fast update!

PR #19673, opened 2025-03-26
[release-3.6] Move client/internal/v2 into server/internal/clientv2

ahrtr:
Second solution for https://github.com/etcd-io/etcd/issues/19670

Previously we moved client/v2 into client/internal/v2, but it caused dependency problem, see issue 19670. Since client/v2 is only used by v2discovery in etcdserver, so moving it into the server module.

Link to https://github.com/etcd-io/etcd/issues/19670


ahrtr:
cc @fuweid @ivanvc @serathius @liggitt

serathius (approved):
Not huge fan of removing e2e test for feature we will still support in v3.6, even if deprecated. But, that's acceptable to move forward with release.

ahrtr:
> Not huge fan of removing e2e test for feature we will still support in v3.6, even if deprecated. But, that's acceptable to move forward with release.

The test fails. It (test) can't import an internal package

tests/go.mod
```diff
@@ -7,7 +7,6 @@ toolchain go1.23.7
 replace (
 	go.etcd.io/etcd/api/v3 => ../api
 	go.etcd.io/etcd/client/pkg/v3 => ../client/pkg
-	go.etcd.io/etcd/client/v2 => ./../client/internal/v2
```
liggitt, line 10: you could lock this to v2.305.0 to keep the test

liggitt:
> The test fails. It (test) can't import an internal package

you might be able to lock the test go.mod to client/v2@v2.305.0 to keep the test

tests/go.mod
liggitt: something like `go.etcd.io/etcd/client/v2 => go.etcd.io/etcd/client/v2 v2.305.0`

ahrtr: It's a good suggestion, but we need to fix & merge & backport https://github.com/etcd-io/etcd/pull/19675 firstly :)

ahrtr: resolved in this PR & thx

ahrtr:
The e2e test is recovered.

@fuweid @serathius PTAL

@ivanvc please also take a look, ensure you are happy with the change to `release_mod.sh`

ivanvc (approved):
The changes to the release script LGTM. Thanks, Benjamin.

fuweid (approved):

PR #19675, opened 2025-03-26
Remove v2 from the release script

ahrtr:
v2 is gone. Let's remove v2 from the release script.

cc @fuweid @serathius @ivanvc


ahrtr:
v2 isn't gone yet in main ( we will cleanup all v2 related stuff later separately), we have to add the commit into https://github.com/etcd-io/etcd/pull/19673

PR #19669, opened 2025-03-26
[Robustness test] Preserve return time for failed requests

henrybear327:
In the current implementation, we patch the return time of failed requests to infinite and save it in the report, so we lose the original return time. 

With the change in this PR, we would now save the return time as-is, and during validation, we patch it to infinite. 

Reference:
- https://github.com/etcd-io/etcd/issues/19579
- https://github.com/etcd-io/etcd/pull/19651#discussion_r2011549432

Closes https://github.com/etcd-io/etcd/issues/19579

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #19679, opened 2025-03-27
Pass a non-infinite returnTime in history patching tests

serathius:
Since we use MaxInt64 to represent infinite for failed transactions, there is no need to provide some infinite ourselves. We can just pass returnTime observed by client and expect it will be set to infinite during preparation step.

/assign @henrybear327

### DIFF
--- a/tests/robustness/validate/patch_history_test.go
+++ b/tests/robustness/validate/patch_history_test.go
@@ -65,7 +65,7 @@ func TestPatchHistory(t *testing.T) {
 		{
 			name: "failed put remains if there is a matching event, return time untouched",
 			historyFunc: func(h *model.AppendableHistory) {
-				h.AppendPut("key", "value", 100, infinite, nil, errors.New("failed"))
+				h.AppendPut("key", "value", 100, 200, nil, errors.New("failed"))
 			},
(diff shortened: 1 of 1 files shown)


henrybear327 (approved):

henrybear327:
Applied the changes to https://github.com/etcd-io/etcd/pull/19669 and the tests have passed.

ahrtr (approved):

ahrtr:
> I would say yes, assuming that we will maintain the same merge practices, use `/hold` command for critical command to ensure everyone had chance to review. Sounds ok?

Basically sounds good. But we may forget to add `/hold` sometimes for critical change:( But anyway, leave it to @ivanvc and @jmhbnz to drive this.

ivanvc:
> > I would say yes, assuming that we will maintain the same merge practices, use `/hold` command for critical command to ensure everyone had chance to review. Sounds ok?
> 
> Basically sounds good. But we may forget to add `/hold` sometimes for critical change:( But anyway, leave it to @ivanvc and @jmhbnz to drive this.

James was working on this. IIRC, there were challenges with our DCO check in his exploration. We agreed to test the workflow on the website.

PR #19692, opened 2025-03-29
Update golangci-lint to v2 and refine linter configurations

mmorel-35:
#### Description

This updates golangci-lint to v2 and adapt associated tools/mod and configurations


ivanvc:
@mmorel-35, can golangci-lint v2 run/work with a v1 configuration YAML file? I'm thinking if we can split up the single commit into multiple.

server/etcdserver/api/v2store/store_ttl_test.go
```diff
@@ -132,7 +132,7 @@ func TestStoreUpdateDirTTL(t *testing.T) {
 	assert.False(t, e.Node.Dir)
 	assert.Equal(t, eidx, e.EtcdIndex)
 	e, _ = s.Get("/foo/bar", false, false)
-	assert.Equal(t, "", *e.Node.Value)
+	assert.Empty(t, *e.Node.Value)
```
ivanvc, line 135: Could you please open a PR with this and `tests/e2e/v2store_deprecation_test.go`'s changes? I believe we can merge them into main, and the linters still pass (I tested it out). Thanks!

ivanvc:
@mmorel-35, with #19757 merged, can you rebase this pull request? Thanks.

ivanvc:
@mmorel-35, sorry for the back and forth regarding the number of commits. But, because we need the new YAML format to use golangci-lint v2, I think it makes more sense to have this in a single commit. The body of the commit can explain the two changes introduced by the single commit in detail.

After that, I think we'll be ready to merge :) Thanks again for the work, @mmorel-35!

PR #19694, opened 2025-03-30
Switch to validating v3 when v2 and v3 are synchronized

serathius:
cc @ahrtr @siyuanfoundation 
Was ready the membership code and noticed that it's still validated with v2. Is this intentional?

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -313,9 +313,16 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 
 // ValidateConfigurationChange takes a proposed ConfChange and
 // ensures that it is still valid.
-func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange) error {
-	// TODO: this must be switched to backend as well.
-	membersMap, removedMap := membersFromStore(c.lg, c.v2store)
+func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
+	var membersMap map[types.ID]*Member
+	var removedMap map[types.ID]bool
+
+	if shouldApplyV3 {
+		membersMap, removedMap = c.be.MustReadMembersFromBackend()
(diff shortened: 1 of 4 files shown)


serathius:
Looks like we planned to implement it for v3.6 but it was reverted https://github.com/etcd-io/etcd/pull/17056

serathius:
Looks like we were planning to change this for v3.6 but it was reverted due to bug in implementation that showed up as flakes in robustness tests https://github.com/etcd-io/etcd/pull/17056

ahrtr:
I think it's safe to keep it as it's in release-3.6 ( I mean it's safe to not change anything in release-3.6), but on the other hand,  if we keep it as it's, we are just postponing the "problem" to release-3.7, eventually we will face & resolve it (`validate on v2store` vs `validate on v3store` in mix version case).

A couple of proposals,
- We need to add the similar verification (added in https://github.com/etcd-io/etcd/pull/19563) in release-3.6 as well. Although v2store has deprecated and we generate v2store using v3store, but we still need to guarantee they are consistent, because we need to guarantee the `ValidateConfigurationChange` has consistent behaviour on v2store vs v3store.
- Previously the failure might be related to the [upgrade issue](https://etcd.io/blog/2025/upgrade_from_3.5_to_3.6_issue/), so I suggest to resurrect https://github.com/etcd-io/etcd/pull/17017 (seems you already did) and test it again.

serathius:
Thinking about the current weird state, please correct me if I missed something:
* Release v3.6 starts from last snapshot, reads v2 state from snapshot.
* During bootstrap it validates membership changes in WAL requests using v2 state and applies them to just v2 state.
* When etcd is bootstrapped, it will continue to validate membership using v2 state, but after it replays WAL till CI it will also update v3 state. This is currently proposed to change in this PR.
* Any new snapshot will be generated not from v2, but from v3 state. This affects snapshots written to WAL and requested from other member.

Potential problem I noticed:
* Can etcd write a new snapshot when replaying WAL? Let's say between restarts user changes --snapshot-count to lower value, will we write a snapshot with CI from v3 and Confstate from v2?

ahrtr (approved):

PR #19695, opened 2025-03-30
Assume v2 store is always not nil

serathius:
I didn't find know any case where during server lifecycle v2store would be nil. I have a sad suspicion that this condition was introduced just for testing.

Would like to introduce hard assumption to v2 not being nil to ensure that it is properly tested and we are 100% sure that when we move to v3 storage the v2 is not used.

### DIFF
--- a/server/etcdserver/api/membership/membership_test.go
+++ b/server/etcdserver/api/membership/membership_test.go
@@ -34,10 +34,6 @@ func TestAddRemoveMember(t *testing.T) {
 	c.AddMember(newTestMember(18, nil, "node18", nil), true)
 	c.RemoveMember(18, true)
 
-	// Skipping removal of already removed member
-	c.RemoveMember(17, true)
-	c.RemoveMember(18, true)
-
 	c.AddMember(newTestMember(19, nil, "node19", nil), true)
 
 	// Recover from backend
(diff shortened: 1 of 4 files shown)


server/etcdserver/api/membership/membership_test.go
```diff
@@ -34,10 +34,6 @@ func TestAddRemoveMember(t *testing.T) {
 	c.AddMember(newTestMember(18, nil, "node18", nil), true)
 	c.RemoveMember(18, true)
 
-	// Skipping removal of already removed member
```
serathius, line 37: Looks like this test was totally incorrectly written?! It dependent on fact that v2store was not set at all.

server/etcdserver/api/membership/storev2.go
```diff
@@ -183,18 +183,3 @@ func MemberStoreKey(id types.ID) string {
 func MemberAttributesStorePath(id types.ID) string {
 	return path.Join(MemberStoreKey(id), attributesSuffix)
 }
-
-func clusterVersionFromStore(lg *zap.Logger, st v2store.Store) *semver.Version {
```
serathius, line 187: No longer used, linting requires removal.

ahrtr:
For release-3.6, I'd like to keep it as it's. For main branch, I suggest to cleanup v2store after 3.6.0 is released, to make the backport (main -> release-3.6) easier.

serathius:
SG

serathius:
@ahrtr any progress with v3.6 release?

serathius:
ping @ahrtr

ahrtr (approved):

Issue #19587, opened 2025-03-13
Use bot to remove stage/triaged tag if no activity for a while


siyuanfoundation:
### What would you like to be added?

Use a bot to remove stage/triaged tag if no activity in a PR/issue for a while (2 month?)

### Why is this needed?

It is easier to rely on the `stage/triaged` tag filter to triage issues and PRs, but we need to watch out for abandoned issue/PR.

/cc @ivanvc @Arka

IAMRogerXi:
@siyuanfoundation 
i'd like to work on this one. thanks.

siyuanfoundation:
/assign @IAMRogerXi 

Thank you!

IAMRogerXi:
i will start the work on this item this week...

IAMRogerXi:
PR: https://github.com/etcd-io/etcd/pull/19716

IAMRogerXi:
thanks for the explanation. i will abandon #19716 and push a new one with prow command in comment.

IAMRogerXi:
i don't have time to push a PR last weekend. will try my best to do so this week...

lavishpal:
Hi @IAMRogerXi , are you still working on this issue?

IAMRogerXi:
a bit busy on work recently... will try to catch this up next week

lavishpal:
Hi @ivanvc , @siyuanfoundation 
By reading the above comments, I noticed that similar jobs are defined in the existing [yaml](https://github.com/kubernetes/test-infra/blob/9c8103abd65925a9e075c01908d40c31d00c1bbf/config/jobs/kubernetes/sig-k8s-infra/trusted/sig-contribex-k8s-triage-robot.yaml#L605).
Would it be okay to add the etcd-specific job there, or would it be better to create a new file like `etcd-triage-bot.yaml` under the same directory?
Just want to confirm the preferred approach before opening a PR.

ivanvc:
Hi @lavishpal, thanks for your interest in working on this. But, @IAMRogerXi is assigned and said he'll get back at it next week. If he doesn't show progress next week, we'll re-assign the task. But as long as it's assigned to another one, let's try not to work on it. Thanks.

IAMRogerXi:
looks like there is a limitation in Prow. it only supports the following labels....
https://github.com/kubernetes/test-infra/blob/master/label_sync/labels.md#why-these-labels


triage/accepted | Indicates an issue or PR is ready to be actively worked on. | org members | label
[...]

PR #19716, opened 2025-04-07
add bot to remove tag for stale issues...

IAMRogerXi:
Add a GitHub bot basing on the requirement in https://github.com/etcd-io/etcd/issues/19587.
- scan all issues with tag 'stage/triaged'.
- if it is not touched for two (2) months, remove the tag 'stage/triaged'
- run the job on the first day of each month.


IAMRogerXi:
close this PR for now. i will push a new one with better approach...

PR #19768, opened 2025-04-20
adt: split interval tree by right endpoint on matched left endpoints

redwrasse:
Splits interval trees by right endpoint on matched left endpoints, to improve exact interval matching performance. 
This addresses the issue: https://github.com/etcd-io/etcd/issues/19769

Additionally I've attached below code snippets and outputs for non-rigorous (Go benchmarks on before/after performance of `Find()` and `Insert()` on randomly built interval trees with multiple occurrences of given left/right endpoint values, which is the regime where I think a performance difference may appear. Run on my laptop, there appears to be no change to the `Insert()` op performance, but a roughly 88% at p-value of 0 (according to benchstat output below) performance increase in `Find()`, on such interval trees.

```
// Benchmark of the Insert op on random interval trees of up to 128 nodes with endpoint values in 0..9
func BenchmarkIntervalTreeRandomInsert(b *testing.B) {
	ivs := make(map[xy]struct{})
	ivt := NewIntervalTree()
	maxv := 128

	xmax := 10
	ymax := 10

	for b.Loop() {
		// Setup random tree.
		for i := rand.Intn(maxv) + 1; i != 0; i-- {
			x, y := int64(rand.Intn(xmax)), int64(rand.Intn(ymax))
			if x > y {
				t := x
				x = y
				y = t
			} else if x == y {
				y++
			}
			iv := xy{x, y}
			if _, ok := ivs[iv]; ok {
				// don't double insert
				continue
			}
			ivt.Insert(NewInt64Interval(x, y), 123)
			ivs[iv] = struct{}{}
		}
	}

}

// Benchmark of the Find op on random interval trees of up to 128 nodes with endpoint values in 0..9
func BenchmarkIntervalTreeRandomFind(b *testing.B) {
	ivs := make(map[xy]struct{})
	ivt := NewIntervalTree()
	maxv := 128

	xmax := 10
	ymax := 10

	// Setup random tree.
	for i := rand.Intn(maxv) + 1; i != 0; i-- {
		x, y := int64(rand.Intn(xmax)), int64(rand.Intn(ymax))
		if x > y {
			t := x
			x = y
			y = t
		} else if x == y {
			y++
		}
		iv := xy{x, y}
		if _, ok := ivs[iv]; ok {
			// don't double insert
			continue
		}
		ivt.Insert(NewInt64Interval(x, y), 123)
		ivs[iv] = struct{}{}
	}

	// Benchmark Find() operation time.
	for b.Loop() {
		for ab := range ivs {
			_ = ivt.Find(NewInt64Interval(ab.x, ab.y))
		}
	}
}


```

Benchmark results for before (count of 10 in each case)

```

goos: darwin
goarch: arm64
pkg: go.etcd.io/etcd/pkg/v3/adt
cpu: Apple M1 Pro
BenchmarkIntervalTreeRandomFind-10    	   79048	     15180 ns/op
BenchmarkIntervalTreeRandomFind-10    	   67766	     17690 ns/op
BenchmarkIntervalTreeRandomFind-10    	   94506	     12688 ns/op
BenchmarkIntervalTreeRandomFind-10    	   77904	     15427 ns/op
BenchmarkIntervalTreeRandomFind-10    	  157444	      7627 ns/op
BenchmarkIntervalTreeRandomFind-10    	  247746	      4840 ns/op
BenchmarkIntervalTreeRandomFind-10    	   72440	     16566 ns/op
BenchmarkIntervalTreeRandomFind-10    	   98607	     12179 ns/op
BenchmarkIntervalTreeRandomFind-10    	   76416	     15652 ns/op
BenchmarkIntervalTreeRandomFind-10    	   67321	     17829 ns/op
PASS
ok  	go.etcd.io/etcd/pkg/v3/adt	12.270s


goos: darwin
goarch: arm64
pkg: go.etcd.io/etcd/pkg/v3/adt
cpu: Apple M1 Pro
BenchmarkIntervalTreeRandomInsert-10    	  531319	      2243 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  555193	      2137 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  555756	      2158 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  554901	      2136 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  542766	      2194 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  546540	      2212 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  555234	      2195 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  549458	      2154 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  543097	      2240 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  535682	      2254 ns/op
PASS
ok  	go.etcd.io/etcd/pkg/v3/adt	12.278s

```

Benchmark results for after:

```
goos: darwin
goarch: arm64
pkg: go.etcd.io/etcd/pkg/v3/adt
cpu: Apple M1 Pro
BenchmarkIntervalTreeRandomFind-10    	  698035	      1720 ns/op
BenchmarkIntervalTreeRandomFind-10    	 2167977	       553.4 ns/op
BenchmarkIntervalTreeRandomFind-10    	  692677	      1734 ns/op
BenchmarkIntervalTreeRandomFind-10    	 1000000	      1142 ns/op
BenchmarkIntervalTreeRandomFind-10    	  667176	      1797 ns/op
BenchmarkIntervalTreeRandomFind-10    	  714930	      1679 ns/op
BenchmarkIntervalTreeRandomFind-10    	  778672	      1543 ns/op
BenchmarkIntervalTreeRandomFind-10    	  687740	      1745 ns/op
BenchmarkIntervalTreeRandomFind-10    	  659038	      1823 ns/op
BenchmarkIntervalTreeRandomFind-10    	  643330	      1859 ns/op
PASS
ok  	go.etcd.io/etcd/pkg/v3/adt	12.244s

goos: darwin
goarch: arm64
pkg: go.etcd.io/etcd/pkg/v3/adt
cpu: Apple M1 Pro
BenchmarkIntervalTreeRandomInsert-10    	  553756	      2149 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  540813	      2224 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  546811	      2181 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  525974	      2237 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  542532	      2218 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  514786	      2328 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  538801	      2233 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  532995	      2262 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  526350	      2265 ns/op
BenchmarkIntervalTreeRandomInsert-10    	  559215	      2183 ns/op
PASS
ok  	go.etcd.io/etcd/pkg/v3/adt	12.296s

```

Benchstat (https://pkg.go.dev/golang.org/x/perf/cmd/benchstat#hdr-Example) summary:

```
goos: darwin
goarch: arm64
pkg: go.etcd.io/etcd/pkg/v3/adt
cpu: Apple M1 Pro
                          │ old_find.txt  │             new_find.txt             │
                          │    sec/op     │    sec/op     vs base                │
IntervalTreeRandomFind-10   15.304µ ± 50%   1.727µ ± 34%  -88.71% (p=0.000 n=10)



goos: darwin
goarch: arm64
pkg: go.etcd.io/etcd/pkg/v3/adt
cpu: Apple M1 Pro
                            │ old_insert.txt │        new_insert.txt         │
                            │     sec/op     │   sec/op     vs base          │
IntervalTreeRandomInsert-10      2.195µ ± 3%   2.228µ ± 2%  ~ (p=0.190 n=10)

```

which I believe says, for the above benchmark cases, 88% speedup with p-value of 0 for `Find`, and no noticeable difference for `Insert`.

### DIFF
--- a/pkg/adt/interval_tree.go
+++ b/pkg/adt/interval_tree.go
@@ -472,8 +472,16 @@ func (ivt *intervalTree) Insert(ivl Interval, val any) {
 	x := ivt.root
 	for x != ivt.sentinel {
 		y = x
-		if z.iv.Ivl.Begin.Compare(x.iv.Ivl.Begin) < 0 {
+		// Split on left endpoint. If left endpoints match, instead split on right endpoint.
+		beginCompare := z.iv.Ivl.Begin.Compare(x.iv.Ivl.Begin)
+		if beginCompare < 0 {
 			x = x.left
+		} else if beginCompare == 0 {
+			if z.iv.Ivl.End.Compare(x.iv.Ivl.End) < 0 {
+				x = x.left
+			} else {
+				x = x.right
+			}
 		} else {
 			x = x.right
 		}
@@ -483,8 +491,15 @@ func (ivt *intervalTree) Insert(ivl Interval, val any) {
 	if y == ivt.sentinel {
 		ivt.root = z
 	} else {
-		if z.iv.Ivl.Begin.Compare(y.iv.Ivl.Begin) < 0 {
+		beginCompare := z.iv.Ivl.Begin.Compare(y.iv.Ivl.Begin)
+		if beginCompare < 0 {
 			y.left = z
+		} else if beginCompare == 0 {
+			if z.iv.Ivl.End.Compare(y.iv.Ivl.End) < 0 {
+				y.left = z
+			} else {
+				y.right = z
+			}
 		} else {
 			y.right = z
 		}
@@ -701,16 +716,29 @@ func (ivt *intervalTree) Visit(ivl Interval, ivv IntervalVisitor) {
 
 // find the exact node for a given interval
 func (ivt *intervalTree) find(ivl Interval) *intervalNode {
-	ret := ivt.sentinel
-	f := func(n *intervalNode) bool {
-		if n.iv.Ivl != ivl {
-			return true
+	x := ivt.root
+	// Search until hit sentinel or exact match.
+	for x != ivt.sentinel {
+		beginCompare := ivl.Begin.Compare(x.iv.Ivl.Begin)
+		endCompare := ivl.End.Compare(x.iv.Ivl.End)
+		if beginCompare == 0 && endCompare == 0 {
+			return x
 		}
-		ret = n
-		return false
+		// Split on left endpoint. If left endpoints match,
+		// instead split on right endpoints.
+		if beginCompare < 0 {
+			x = x.left
+		} else if beginCompare == 0 {
+			if endCompare < 0 {
+				x = x.left
+			} else {
+				x = x.right
+			}
+		} else {
+			x = x.right
+		}
 	}
-	ivt.root.visit(&ivl, ivt.sentinel, f)
-	return ret
+	return x
 }
 
 // Find gets the IntervalValue for the node matching the given interval


pkg/adt/interval_tree.go
```diff
@@ -701,16 +716,29 @@ func (ivt *intervalTree) Visit(ivl Interval, ivv IntervalVisitor) {
 
 // find the exact node for a given interval
 func (ivt *intervalTree) find(ivl Interval) *intervalNode {
```
serathius, line 718: I was not able to find a proper tests for the `Find` method. Would you mind preparing some tests before we change the logic?

serathius:
Please include benchmark from https://github.com/etcd-io/etcd/issues/19769 in the PR.

pkg/adt/interval_tree.go
redwrasse: Yes,I'll add tests. I was thinking of just adding assertions on the `Find` operation to the existing `TestIntervalTreeRandom` [here](https://github.com/etcd-io/etcd/blob/main/pkg/adt/interval_tree_test.go#L344) as say
```
	for ab := range ivs {
		for xy := range ivs {
			v := xy.x + int64(rand.Intn(int(xy.y-xy.x)))
			require.NotEmptyf(t, ivt.Stab(NewInt64Point(v)), "expected %v stab non-zero for [%+v)", v, xy)
			require.Truef(t, ivt.Intersects(NewInt64Point(v)), "did not get %d as expected for [%+v)", v, xy)
		}
		iv := ivt.Find(NewInt64Interval(ab.x, ab.y))
		assert.NotNil(t, iv)
		assert.Equal(t, NewInt64Interval(ab.x, ab.y), iv.Ivl)
		assert.Truef(t, ivt.Delete(NewInt64Interval(ab.x, ab.y)), "did not delete %v as expected", ab)
		delete(ivs, ab)
	}
```

In addition I can add a separate dedicated unit test for the `Find` op to test edge cases if that's warranted.

redwrasse:
> Please include benchmark from #19769 in the PR.

Will do, I'll investigate using benchstats.

pkg/adt/interval_tree.go
serathius: Thanks, please send a separate PR. It should be much easier to review and merge a new tests that don't change the logic.

redwrasse: @serathius opened up a separate MR with `IntervalTree.Find()` tests: https://github.com/etcd-io/etcd/pull/19801

redwrasse:
> > Please include benchmark from #19769 in the PR.
> 
> Will do, I'll investigate using benchstats.

Will try to get benchstats added in the next few days.

redwrasse:
> > > Please include benchmark from #19769 in the PR.
> > 
> > 
> > Will do, I'll investigate using benchstats.
> 
> Will try to get benchstats added in the next few days.

Added benchmarks to this PR description, using benchstat. cc @serathius

serathius:
Can you add the benchmarks to the PR so we can use them to evaluate performance in the future? We could also consider merging them first in separate PR.

serathius:
If there is no performance change for `Insert` maybe let's drop the changes there from the PR. Reducing the scope will make the PR easier to review. Complicating `Insert` code without improving performance doesn't seem like a good tradeoff.

redwrasse:
If I'm understanding , are you suggesting removing the code changes to `Insert` but keeping the ones in `Find`? I'm not sure how that would work- the interval tree structure changes need to be reflected in both methods I believe.

I can definitely add in the benchmarks from the description, if we feel they are appropriate.

serathius:
cc @ahrtr

redwrasse:
Cormen "Introduction to Algorithms", Chapter 14 Exercise 14.3.5 is I think relevant:

14.3-5
Suggest modifications to the interval-tree procedures to support the new operation `INTERVAL-SEARCH-EXACTLY(T,i)`, where `T` is an interval tree and `i` is an interval. The operation should return a pointer to a node `x` in `T` such that `x.int.low = i.low` and `x.int.high = i.high`, or `T.nil` if `T` contains no such node. All operations, including `INTERVAL-SEARCH-EXACTLY`, should run in `O(lg n)` time on an `n`-node interval tree.

serathius (approved):
Change makes sense to me, but I'm not an expert on this. 

Will still check on coverage whether the tests in https://github.com/etcd-io/etcd/pull/19801  are enough.

pkg/adt/interval_tree.go
```diff
@@ -472,8 +472,16 @@ func (ivt *intervalTree) Insert(ivl Interval, val any) {
 	x := ivt.root
 	for x != ivt.sentinel {
 		y = x
-		if z.iv.Ivl.Begin.Compare(x.iv.Ivl.Begin) < 0 {
+		// Split on left endpoint. If left endpoints match, instead split on right endpoint.
```
serathius, line 475: Can you update the function docstring?

serathius:
Coverage looks pretty good https://app.codecov.io/gh/etcd-io/etcd/pull/19768?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=etcd-io

pkg/adt/interval_tree.go
```diff
@@ -701,16 +716,29 @@ func (ivt *intervalTree) Visit(ivl Interval, ivv IntervalVisitor) {
...
+		if beginCompare < 0 {
+			x = x.left
+		} else if beginCompare == 0 {
+			if endCompare < 0 {
+				x = x.left
+			} else {
+				x = x.right
+			}
+		} else {
+			x = x.right
```
ahrtr, line 738: Pls revert the change to the `Insert` method
```suggestion
		if beginCompare < 0 {
			x = x.left
		} else {
			x = x.right
```

pkg/adt/interval_tree.go
```diff
@@ -472,8 +472,16 @@ func (ivt *intervalTree) Insert(ivl Interval, val any) {
...
+		// Split on left endpoint. If left endpoints match, instead split on right endpoint.
+		beginCompare := z.iv.Ivl.Begin.Compare(x.iv.Ivl.Begin)
+		if beginCompare < 0 {
 			x = x.left
+		} else if beginCompare == 0 {
+			if z.iv.Ivl.End.Compare(x.iv.Ivl.End) < 0 {
+				x = x.left
+			} else {
+				x = x.right
+			}
```
ahrtr, line 484: **You changed the behaviour, so it's a breaking change.**

Previously it only checks the interval's `Begin`; an intervalTree inserts a new node to the left if` newNode.Ivl.Begin < x.Ivl.Begin`, otherwise inserts it to the right.

Now you not only checks the `Begin`, but also checks the `End`, and inserts the new node to the left if Begin matches and newNode's End is less.

This PR's purpose is to optimize the exact search (find), I don't see a reason why update the Insert. So please consider to revert the change. Pls also see comment for the `find` method.

redwrasse: @ahrtr thanks very much for reviewing. 

So now I'm confused. The goal I had with this MR / issue is indeed to speed up `Find` to logarithmic time (as, for example, the Cormen 14.3-5 excercise addresses), instead of the existing visitor implementation. To do this, in say a textbook implementation, I thought requires updating both Find and Insert operations to further split on right endpoint if the left endpoints are matched?

redwrasse: I think you're proposing: keep the existing interval tree structure of splitting left if less than left endpoint, else split right. Update the `Find` implementation to follow this split logic, hence an optimization in the sense that `Find` will never be checking both left and right subtrees.

With this logic I see the existing etcd interval tree tests fail. The issue is I think this approach doesn't preserve red-black tree rotation invariance. 

(code changes for below snippets on this branch https://github.com/etcd-io/etcd/commit/84120212bd98b8f909cbba0dcadf7bbfd45c1da7)

Here is updated `find`, using the proposed logic of split left if less than left endpoint, else split right:

```
// find the exact node for a given interval
func (ivt *intervalTree) find(ivl Interval) *intervalNode {
	x := ivt.root
	// Search until hit sentinel or exact match.
	for x != ivt.sentinel {
		beginCompare := ivl.Begin.Compare(x.iv.Ivl.Begin)
		endCompare := ivl.End.Compare(x.iv.Ivl.End)
		if beginCompare == 0 && endCompare == 0 {
			// Found a match.
			return x
		}
		// Split on left endpoint: if less than, go left, else go right.
		if beginCompare < 0 {
			x = x.left
		} else {
			x = x.right
		}
	}
	return x
}
```

and a corresponding unit test (which fails) I think illustrating the point about rotations:

```
func TestProposedFindExample(t *testing.T) {

	//won't work  because won't satisfy tree rotation invariance property required of red-black trees, eg. if [2,7] is written then can't asssme all subsequently written [2,*] entities will remain in right subtree of [2,7]- won't, because tree rotations are possible.

	// OTOH with the proposed approach, all subsequent [2, x] writes with x > 7 will land in right/after of [2,7], and those with x < 7 will land to left /before. Under tree rotation ordering is preserved.

	ivt := NewIntervalTree()

	lEndp := int64(2)
	rEndps := []int64{7, 3, 9}
	val := 123

	for _, re := range rEndps {
		ivt.Insert(NewInt64Interval(lEndp, re), val)
	}

	// What we would expect from 'insert left if less than left endpoint, else insert right' without tree rotations:
	// (we can generate this tree by commenting out the `ivt.insertFixup(z)` line in the `Insert` op)
	//  Insert [2, 7), then Insert [2, 3), then Insert [2, 9) becomes:
	//   	[2, 7)
	//	        \
	//  		[2, 3)
	//				\
	//			  [2, 9)

	// Instead, due to rotations (rb-fixup), we get:
	//      [2, 3)
	//     /     \
	//  [2, 7)   [2, 9)

	// Find fails because it assumes the former tree structure, eg. thinks it can always search right in the above example:
	for _, re := range rEndps {
		ivl := NewInt64Interval(lEndp, re)
		assert.NotNil(t, ivt.Find(ivl))
		assert.Equal(t, ivl, ivt.Find(ivl).Ivl)
	}
}

```

ahrtr: I see. I was thinking the `find` should work as long as it matches logic (search path) as `Insert`, but actually it might not match due to `insertFixup`.

The question for now is will `insertFixup` cause the same problem for this PR?

In this PR, you updated both `find` and `Insert`, and follow the same logic (search path): splitting both left(`Begin`) and right(`End`) endpoints. Due to `insertFixup`, is it possible that it may also cause the `find` fail?

ahrtr: > Any thoughts for how to further test/guarantee correctness?

Good question. 

- I think both `rotateLeft` and `rotateRight` always keep the invariable property: `x.left < x < x.right`

rotateLeft (a):
```
       a                                                          b
         \                                                     /      \
           b                     --->                   a          d
        /      \                                               \
    c          d                                              c
```  

rotateRight (a):
```
       a                                                      b
      /                                                     /      \
    b                     --->                       c          a
  /      \                                                        /
 c          d                                              d
```  

- I ran `go test -run TestIntervalTreeRandom -v -count 200 -failfast` multiple times, and always passed.

ahrtr: As mentioned in https://github.com/etcd-io/etcd/pull/19768#discussion_r2095649617, it changes the behaviour, but I think we are good as long as we don't break the `IntervalTree` API

ahrtr (approved):
LGTM

Thanks for the optimisation.

ahrtr:
We should add a performance item in 3.7 changelog, let me take care of together with another change.

pkg/adt/interval_tree.go
redwrasse: Sounds good, thanks for reviewing and investigating @ahrtr!

pkg/adt/interval_tree.go
ahrtr: Sorry, missed this comment.  The comment for the Insert method isn't accurate anymore, we need to update it. We also need to add comment for the `find` method. @redwrasse 

https://github.com/etcd-io/etcd/blob/c849507a3895a93742d83845635e17a069a8ee2e/pkg/adt/interval_tree.go#L437-L468

redwrasse: @serathius @ahrtr I'll open an MR for updating the comments.

redwrasse: Created a pull request with docstring updates: https://github.com/etcd-io/etcd/pull/20015

Issue #19700, opened 2025-04-01
panic: runtime error: comparing uncomparable type map[string]interface {}


cruvie:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

use grpc client made by  etcd resolver with `google.golang.org/grpc v1.71.0` cause panic 

### What did you expect to happen?

call rpc method successfully

### How can we reproduce it (as minimally and precisely as possible)?

get a grpc client like this

```go

import (
	"go.etcd.io/etcd/client/v3/naming/resolver"
	"google.golang.org/grpc"
)

// GetGrpcClient get grpc client for serverName
//
// don't forget to close conn conn.Close()
func GetGrpcClient[T any](serverName string,
	clientBuilder func(grpcConn grpc.ClientConnInterface) (client T),
	opts ...grpc.DialOption) (conn *grpc.ClientConn, client T, err error) {

	//refers to https://etcd.io/docs/v3.5/dev-guide/grpc_naming/
	etcdResolver, err := resolver.NewBuilder(GetClient())
	if err != nil {
		return nil, client, err
	}

	options := []grpc.DialOption{
		grpc.WithResolvers(etcdResolver),
		grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy":"round_robin"}`),
	}
	options = append(options, opts...)

	conn, err = grpc.NewClient("etcd:///grpc/"+serverName,
		options...,
	)
	client = clientBuilder(conn)
	return conn, client, err
}
```

and use the client call a method

### Anything else we need to know?

go 1.24
google.golang.org/grpc v1.70.0 works fine
create a grpc client manually  works fine

```go

func TestPanic(t *testing.T) {
	etcdCloseFunc := global_config.InitEtcdClient(false)
	defer etcdCloseFunc()
	_, client, err := GetClient()
	if err != nil {
		t.Error(err)
	} else {
[...]

PR #19780, opened 2025-04-22
Replace `resolver.State.Addresses` with `resolver.State.Endpoint.Addresses`

ahrtr:
grpc-go is going to remove `resolver.State.Addresses`, and it will be fully replaced by `resolver.State.Endpoint`.

References:
- https://github.com/etcd-io/etcd/issues/19700#issuecomment-2819173767
- https://github.com/grpc/grpc-go/blob/v1.71.1/resolver/resolver.go#L196-L205

cc @dfawley @fuweid @amosehiguese


ahrtr:
also cc @ivanvc and @serathius

dfawley (approved):

PR #12671, opened 2021-02-07
clientv3: Replace balancer with upstream grpc solution

ptabor:
Replace custom balancer logic in clientv3 with grpc infrastructure: 
 - custom resolver + round_robin load balancer. 

This commit significantly reduces volume of custom code
in etcd client v3, while preserving full existing functionality.

PR #12614, opened 2021-01-12
Replacement API for client/v3/naming package to be compatible with new GRPC1.30+ resolver API.

ptabor:
This is not yet implementation, just API definitions.

Please review whether the API is good / can be improved. The goal is to reach agreement about
desired layout of the naming/resolver packages in etcd-3.5+, before investement in the implementation of the API. 

We propose here 3 packages:
  - clientv3/naming/endpoints ->
    That is abstraction layer over etcd that allows to write, read &
watch Endpoints information. It's independent from GRPC API. It hides
the storage details.

  - clientv3/naming/endpoints/internal ->
    That contains the grpc's compatible Update class to preserve the
internal JSON mashalling format.

  - clientv3/naming/resolver ->
    That implements the GRPC resolver API, such that etcd can be
    used for connection.Dial in grpc.

Please see the [grpc_naming.md](TLDR: Please take a look at the proposed modified 3.5 [grpc_naming user guide](https://github.com/etcd-io/etcd/pull/12614/files?short_path=5329e8f#diff-5329e8f4ea030d3052a34c8831bfc948bc5100e935c4cff4175fe75cbd9ee877) document changes & grpcproxy/cluster.go new integration, to see how the new abstractions work.

Note: I'm trying to reach agreement on the vision. I would appreciate (when/if accepted) contributions covering the implementation aspect. 

@skyao @mcluseau @scDisorder @xiang90

PR #6632, opened 2016-10-12
clientv3/naming: support resolving to multiple hosts

heyitsanthony:
Previous implementation watches a single key so there's no way to have separate hosts associate with separate keys for a single grpc target. Instead, accept all keys on a prefix.

Also fixes first the Next() to read current name data from etcd instead of waiting for the next event on a synced watcher.

PR #5799, opened 2016-06-28
clientv3: add grpc naming resolver

xiang90:
/cc @iamqizhao

PR #19785, opened 2025-04-23
feat(client/v3/naming): add Attributes into Endpoint and deprecate Metadata

### LINKED ISSUE #19706: Deprecate usage of `gresolver.Address.Metadata`

Deprecate usage of `gresolver.Address.Metadata` in [client/v3/naming/resolver/resolver.go](https://github.com/etcd-io/etcd/blob/d6416b8773b2bfef2d788027a60e1f2de8d7e4ae/client/v3/naming/resolver/resolver.go#L114), and add `Attributes` field.

### Why is this needed?

`gresolver.Address.Metadata` in grpc is deprecated, and should use `Attributes` instead.
https://github.com/grpc/grpc-go/blob/51d6a43ec59753d42ccd02bb12d2e9e40c164f0f/resolver/resolver.go#L117-L121

amosehiguese:
This implementation adds the new Attributes *attributes.Attributes into Endpoint and deprecate Metadata accordingly. The metadata field is still supported for backward compatibility

Fixes #19706 

cc: @ahrtr @siyuanfoundation

### DIFF
--- a/client/v3/naming/endpoints/endpoints.go
+++ b/client/v3/naming/endpoints/endpoints.go
@@ -32,6 +32,8 @@ type Endpoint struct {
 	// Metadata is the information associated with Addr, which may be used
 	// to make load balancing decision.
 	// Since etcd 3.1
+	//
+	// Deprecated: The field is deprecated and should not be used.
 	Metadata any
 }
 
--- a/client/v3/naming/resolver/resolver.go
+++ b/client/v3/naming/resolver/resolver.go
@@ -100,22 +100,26 @@ func (r *resolver) watch() {
 				}
 			}
 
-			addrs := convertToGRPCAddress(allUps)
-			r.cc.UpdateState(gresolver.State{Addresses: addrs})
+			eps := convertToGRPCEndpoint(allUps)
+			r.cc.UpdateState(gresolver.State{Endpoints: eps})
 		}
 	}
 }
 
-func convertToGRPCAddress(ups map[string]*endpoints.Update) []gresolver.Address {
-	var addrs []gresolver.Address
+func convertToGRPCEndpoint(ups map[string]*endpoints.Update) []gresolver.Endpoint {
+	var eps []gresolver.Endpoint
 	for _, up := range ups {
-		addr := gresolver.Address{
-			Addr:     up.Endpoint.Addr,
-			Metadata: up.Endpoint.Metadata,
+		ep := gresolver.Endpoint{
+			Addresses: []gresolver.Address{
+				{
+					Addr:     up.Endpoint.Addr,
+					Metadata: up.Endpoint.Metadata,
+				},
+			},
 		}
-		addrs = append(addrs, addr)
+		eps = append(eps, ep)
 	}
-	return addrs
+	return eps
 }
 
 // ResolveNow is a no-op here.
--- a/client/v3/internal/resolver/resolver.go
+++ b/client/v3/internal/resolver/resolver.go
@@ -61,13 +61,15 @@ func (r *EtcdManualResolver) SetEndpoints(endpoints []string) {
 
 func (r EtcdManualResolver) updateState() {
 	if r.CC != nil {
-		addresses := make([]resolver.Address, len(r.endpoints))
+		eps := make([]resolver.Endpoint, len(r.endpoints))
 		for i, ep := range r.endpoints {
 			addr, serverName := endpoint.Interpret(ep)
-			addresses[i] = resolver.Address{Addr: addr, ServerName: serverName}
+			eps[i] = resolver.Endpoint{Addresses: []resolver.Address{
+				{Addr: addr, ServerName: serverName},
+			}}
 		}
 		state := resolver.State{
-			Addresses:     addresses,
+			Endpoints:     eps,
 			ServiceConfig: r.serviceConfig,
 		}
 		r.UpdateState(state)


client/v3/naming/resolver/resolver.go
```diff
@@ -100,22 +102,30 @@ func (r *resolver) watch() {
...
 
-			eps := convertToGRPCEndpoint(allUps)
+			eps := convertToGRPCEndpoint(lg, allUps)
 			r.cc.UpdateState(gresolver.State{Endpoints: eps})
 		}
 	}
 }
 
-func convertToGRPCEndpoint(ups map[string]*endpoints.Update) []gresolver.Endpoint {
+func convertToGRPCEndpoint(lg *zap.Logger, ups map[string]*endpoints.Update) []gresolver.Endpoint {
```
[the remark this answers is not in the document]
amosehiguese: I was thinking since it is a private function and only called within the watch function it wouldn't have much impact.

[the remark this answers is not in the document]
amosehiguese: I will revert back then. Thank you for the feedback

ahrtr:
please squash the commits. 

We also need to update the doc https://etcd.io/docs/v3.5/dev-guide/grpc_naming/
- the `Metadata` has deprecated, use `Attributes` instead
- Mark the gRPC naming and discovery feature as experimental, depending on how the resolver package in grpc-go evolves.

amosehiguese:
Alright. I'll do just that.

amosehiguese:
> We also need to update the doc

I believe this should be done from the etcd’s website repo, right?

ahrtr:
> * the `Metadata` has deprecated, use `Attributes` instead

let's update this first together with this PR.

- after this PR gets merged, backport it to 3.6 and 3.5.
- raise a PR in etcd-io/website to update the doc



> * Mark the gRPC naming and discovery feature as experimental, depending on how the resolver package in grpc-go evolves.

we will discuss & resolve this separately.


cc @dfawley @fuweid @serathius @ivanvc

client/v3/naming/resolver/resolver.go
```diff
@@ -112,8 +112,9 @@ func convertToGRPCEndpoint(ups map[string]*endpoints.Update) []gresolver.Endpoin
 		ep := gresolver.Endpoint{
 			Addresses: []gresolver.Address{
 				{
-					Addr:     up.Endpoint.Addr,
-					Metadata: up.Endpoint.Metadata,
+					Addr:       up.Endpoint.Addr,
+					Metadata:   up.Endpoint.Metadata,
+					Attributes: up.Endpoint.Attributes,
```
ahrtr, line 117: @dfawley 

Probably eventually we might want to use [Endpoint.Attributes](https://github.com/grpc/grpc-go/blob/ec2d624ac9eebaeacb5ff3be128b3d6e0c4f9fec/resolver/resolver.go#L191) instead of  [Address.Attributes](https://github.com/grpc/grpc-go/blob/ec2d624ac9eebaeacb5ff3be128b3d6e0c4f9fec/resolver/resolver.go#L107), as we only care about the LB policy?

Currently we assume each endpoint has only one address, but actually it may contain multiple addresses (we won't change the behavior for now).

ahrtr:
You also need to persist the new added attributes (line 41 in internal/update.go), and read it back in List,

https://github.com/etcd-io/etcd/blob/8f933a5b5867d078c714fd6a9584aa47f450d8d0/client/v3/naming/endpoints/endpoints_impl.go#L175-L192

amosehiguese:
> let's update this first together with this PR.
Alright. 

> after this PR gets merged, backport it to 3.6 and 3.5.
I'm on standby.

> raise a PR in etcd-io/website to update the doc
I will hold off on this, till others has been done.

client/v3/naming/resolver/resolver.go
amosehiguese: Yeah, I saw it and then realised that the attributes field in the Endpoint was specific to LB-Policy which is the usecase for etcd and wondered why that wasn't used instead. But I now have clarity on that now. Thanks

amosehiguese:
> You also need to persist the new added attributes (line 41 in internal/update.go), and read it back in List,
> 
> https://github.com/etcd-io/etcd/blob/8f933a5b5867d078c714fd6a9584aa47f450d8d0/client/v3/naming/endpoints/endpoints_impl.go#L175-L192

Got it.

amosehiguese:
> You might also need to update grpcproxy

Pretty much adjust every other usage of the endpoint struct to account for the new Attributes field. 

Will do that.

amosehiguese:
Two files may also need adjustments but I thought to run it by you first.

👉 endpoints_test.go [integration tests]

https://github.com/etcd-io/etcd/blob/8f933a5b5867d078c714fd6a9584aa47f450d8d0/tests/integration/clientv3/naming/endpoints_test.go#L46

👉 cluster.go [grpc-proxy]
In MembersList, If proxies are registered, it will lists them using their metadata and endpoint addresses.

https://github.com/etcd-io/etcd/blob/8f933a5b5867d078c714fd6a9584aa47f450d8d0/server/proxy/grpcproxy/cluster.go#L146


What would you have me do to these files?

ahrtr:
I think 
- we need update (or add) the test to cover the new field `Attributes`
- we shouldn't break any usage on the existing field `Metadata`.

amosehiguese:
> I think 
> - we need update (or add) the test to cover the new field `Attributes`
> - we shouldn't break any usage on the existing field `Metadata`.

Alright then.

ahrtr:
I just mark this as a **priority/important-soon** task. 

As mentioned above https://github.com/etcd-io/etcd/pull/19785#issuecomment-2824785129, I think we need to backport this PR to release-3.6 and release-3.5, of course we will evaluate the impact.  The goal is to get it included in 3.6.0, so that we can remove it in 3.7; otherwise, we will have to completely get rid of the deprecated Metadata field in 3.8 (~6 years later based on current cadence)

ahrtr:
also cc grpc-go tech-lead @dfawley, thx for all the help so far.

amosehiguese:
@ahrtr
Quick question. Are flaky tests expected in the end-to-end test suite? I’ve been running make test-e2e and sometimes it passes, but other times it fails without me changing anything. Just want to be sure if that’s expected or if something else is going on, because it’s been slowing down my push.

amosehiguese:
> I just mark this as a priority/important-soon task.

I will done with this from my end once I'm able to resolve whether or not flaky tests are expected in the e2e suite

amosehiguese:
> > Quick question. Are flaky tests expected in the end-to-end test suite?
> 
> You need to ensure the test failures are not caused by your changes.

I think I’ll have to run it against the main branch a couple of times to be sure.

amosehiguese:
erro
Same error occured on the main branch after running successfully in the previous run. Seems to look like a connection timeout.

What do you reckon I do? push?

cc: @ahrtr

ahrtr:
> Same error occured on the main branch after running successfully in the previous run. Seems to look like a connection timeout.

Windows is only tier 3 support, please try to test it on a linux platform.



> What do you reckon I do? push?

Pushing to your dev branch won't break anything. Please review your PR yourself first to ensure it's the best you can deliver.

amosehiguese:
Oh, I am actually working on linux machine running on Azure cloud. This is just my backup pc, so I am forced to do everything on the cloud. 

>Pushing to your dev branch won't break anything. Please review your PR yourself first to ensure it's the best you can deliver.
<

I definitely will. Thanks for the feedback

amosehiguese:
@dfawley

Just a bit of clarification in your docs. This line to be precise. Is this meant to be, "Is it impossible..." or "It is impossible"

https://github.com/grpc/grpc-go/blob/030938e543b4a721cd426d15611d50ecf097dcf3/attributes/attributes.go#L137

amosehiguese:
> I think 
> - we need update (or add) the test to cover the new field `Attributes`
> - we shouldn't break any usage on the existing field `Metadata`.

Attempting to do this would break these test cases. 

👇 https://github.com/etcd-io/etcd/blob/f5dbde1d19644352283f6a32d81e6174a04479c3/tests/integration/clientv3/naming/endpoints_test.go#L29

👇
https://github.com/etcd-io/etcd/blob/f5dbde1d19644352283f6a32d81e6174a04479c3/tests/integration/clientv3/naming/endpoints_test.go#L120


Mostly because of the design decision by grpc on the Attributes type to make it impossible to  unmarshal attributes from a JSON representation

client/v3/naming/endpoints/endpoints.go
```diff
@@ -32,7 +34,14 @@ type Endpoint struct {
...
 	// to make load balancing decision.
 	// Since etcd 3.1
+	//
+	// Deprecated: use Attributes instead
 	Metadata any
+
+	// Attributes contains arbitrary data about this address intended for
+	// consumption by the SubConn.
+	// Since etcd 3.5
+	Attributes *attributes.Attributes
```
ahrtr, line 44: We can assign this map to attributes.Attributes using the utilities `New` and `WithValue`

https://github.com/grpc/grpc-go/blob/4cedec40eb2ccfbe3f56bb15e894903111ada2d2/attributes/attributes.go#L44-L62 

```suggestion
	Attributes map[string]any
```

ahrtr: Unfortunately the the grpc attributes package is also `Experimental`

amosehiguese: This approach will solve the JSON problem for sure. I will adjust it now. Meanwhile, you also mentioned the grpc attributes package is also `Experimental`. I mean we've already decided to take this direction so, things like this should be expected. Except you have something else in mind, of course. Are you having second thoughts? 

cc: @ahrtr

ahrtr: > Are you having second thoughts?

The alternative is to support generic type (a comparable type), so that the key of the map doesn't have to be a string. But let's keep it simple for now, as we are playing with grpc-go's experimental feature/packages.

client/v3/naming/resolver/resolver.go
ahrtr: > Probably eventually we might want to use [Endpoint.Attributes](https://github.com/grpc/grpc-go/blob/ec2d624ac9eebaeacb5ff3be128b3d6e0c4f9fec/resolver/resolver.go#L191) instead of [Address.Attributes](https://github.com/grpc/grpc-go/blob/ec2d624ac9eebaeacb5ff3be128b3d6e0c4f9fec/resolver/resolver.go#L107), as we only care about the LB policy?

It seems that this comment hasn't been resolved yet.

client/v3/naming/resolver/resolver.go
```diff
@@ -130,3 +132,17 @@ func (r *resolver) Close() {
...
 }
+
+func convertToGRPCAttibutes(attrMap map[string]any) *attributes.Attributes {
+	var attr *attributes.Attributes
+	if len(attrMap) > 0 {
+		for k, v := range attrMap {
+			if attr != nil {
+				attr = attr.WithValue(k, v)
+			} else {
+				attr = attributes.New(k, v)
```
dfawley, line 143: `WithValue` already does this, so you can simplify by deleting the `else` here.

client/v3/naming/endpoints/internal/update.go
```diff
@@ -49,4 +49,11 @@ type Update struct {
 	// Metadata is not required for a custom naming implementation.
 	// Since etcd 3.1.
 	Metadata any
+
+	// Attributes is for storing and retrieving generic key/value pairs.
+	// Keys must be hashable, and users should define their own types for keys.
```
dfawley, line 54: > Keys must be hashable

This needs to be updated with the change to `map`.

client/v3/naming/endpoints/endpoints.go
dfawley: +1 you should not use any experimental APIs in your own APIs, unless you want to declare them as experimental and your users are aware that you reserve the right to break them in the future.

You also need that disclaimer on any package that internally uses experimental APIs.  But if you refrain from exposing it in your own APIs, then there should be some path forward that doesn't require changes to the user's code, too.

client/v3/naming/resolver/resolver.go
amosehiguese: We'll have to wait on @dfawley, I guess

dfawley: The attributes on the address are passed to our code that creates and manages the connection to that address.  This means if there are two `Addresses` between name resolver updates with the same IP address but non-equal attributes, we will consider them different and reconnect.

The attributes on the endpoint are for anything else you might want to do.  gRPC won't look at your attributes.

One important thing that is still not clear to me: what is your use case for attributes?  Why are you adding them into gRPC endpoints/addresses?

ahrtr: > One important thing that is still not clear to me: what is your use case for attributes? Why are you adding them into gRPC endpoints/addresses?

This is a good question, which I am not 100% sure either. But I believe the concept should be tightly coupled with gprc-go.

In the first place, we (previous maintainer) added Metadata (see below), which was passed to grpc-go's [Address.Metadata](https://github.com/grpc/grpc-go/blob/ec2d624ac9eebaeacb5ff3be128b3d6e0c4f9fec/resolver/resolver.go#L117-L121). Based on the comment, it's used for load balancing decision.
https://github.com/etcd-io/etcd/blob/f5dbde1d19644352283f6a32d81e6174a04479c3/client/v3/naming/endpoints/endpoints.go#L32-L35







> This means if there are two `Addresses` between name resolver updates with the same IP address but non-equal attributes, we will consider them different and reconnect.

It seems the only usage of the Address.Attribute is to differentiate two addresses with the same IP address? I originally thought grpc-go would transparently pass through any data users attach—Attributes—to the server. 

Actually based on current implementation, each endpoint has only one address inside.

> The attributes on the endpoint are for anything else you might want to do. gRPC won't look at your attributes.

In that case, then I don't think we should use it if users do not use it either.  Would grpc-go transparently pass through the endpoint Attributes—to the server?

dfawley: > In the first place, we (previous maintainer) added Metadata (see below), which was passed to grpc-go's Address.Metadata. Based on the comment, it's used for load balancing decision.

Except...you don't have a custom load balancer, so there's no decisions you could be making based on it.  Unless your users are also injecting their own custom LB policy.

> It seems the only usage of the Address.Attribute is to differentiate two addresses with the same IP address? I originally thought grpc-go would transparently pass through any data users attach—Attributes—to the server.

Address attributes are mostly used to configure how grpc connects to the remote server.  We currently have two or three attributes that we use for this, but they're actually all currently private.  As an example: one attribute allows us to specify an HTTP CONNECT proxy to use when creating the connection.

One thing we have mentioned (unfortunately & many years ago) that `Metadata` could be used for, is to allow the application to create multiple versions of the same address to force the channel to connect to it multiple times in the round robin LB policy.  Address attributes would allow you to do the same thing, since you'd be able to "trick" the LB policy into thinking the addresses are distinct, even though the IP address is the same.  The majority of users have no need for this.  We really want a better solution for this class of problem that will work cross-language, but it's a hard problem.  See https://github.com/grpc/grpc/issues/21386 for more details.

> In that case, then I don't think we should use it if users do not use it either. Would grpc-go transparently pass through the endpoint Attributes—to the server?

gRPC has no way to serialize these attributes and send them to the server.  gRPC doesn't actually do _any_ communication between client and server besides RPCs.  So it would also have to be part of the RPCs somehow.

client/v3/naming/resolver/resolver.go
amosehiguese: Thank you for the review @dfawley. I will make the necessary adjustments.

client/v3/naming/resolver/resolver.go
ahrtr: thx for the feedback. It seems that we have two options:
1. Add Attributes and pass it to the grpc-go's [Address.Attributes](https://github.com/grpc/grpc-go/blob/ec2d624ac9eebaeacb5ff3be128b3d6e0c4f9fec/resolver/resolver.go#L107)
    - The problem is that we don't know how to properly document it (how & why should users use it?). Is there a link from grpc-go side so that we can just point users to?
2. Do not add the new field `Attributes` at all
    - The original thought is to add a replacement for the deprecated `Metadata`. Do we have to necessarily do it? 
    - If we add the new field `Attributes`, it must be used by grpc-go or users. Based on your feedback, it seems neither need it (for grpc-go, users can't set it due to they being private in grpc-go)?

dfawley: > The problem is that we don't know how to properly document it (**how & why should users use it?**). Is there a link from grpc-go side so that we can just point users to?

Sorry, I don't think so.  Address attributes are mostly internal implementation details, not things we expect users to be doing.  Since you are our user, we were hoping you would know.  But it looks like only _your_ users are (or might be?) using it.  So nobody know how or even if it's being used.  Do you have some way to contact people who might be using it to see what they say?  Or can we go back to the original PR that added this functionality and see if there are breadcrumbs or contact the people involved back then, to see what the motivation was?

> The original thought is to add a replacement for the deprecated Metadata. Do we have to necessarily do it?

It depends on the purpose of the use.  `Metadata` had multiple uses.  The partial replacement of `Endpoint.Attributes` is used for LB policy decision-making purposes.  The partial replacement of `Address.Attributes` is mainly used for gRPC internal usage.

> If we add the new field Attributes, it must be used by grpc-go or users. Based on your feedback, it seems neither need it (for grpc-go, users can't set it due to they being private in grpc-go)?

Correct.  Currently `Address.Attributes` cannot be set to any value that gRPC will use.  `Endpoint.Attributes` is intended for users of custom LB policies to use, but if you are using the built-in round robin, then the only use of that would be to make the same address appear multiple times and not get deduplicated.

dfawley: I'm looking a little into `etcd`'s history and it looks like there used to _also_ be a `balancer` package.  Maybe the `Metadata` is just a relic of some of that code?

ahrtr: > Do you have some way to contact people who might be using it to see what they say?

I am afraid there is no one to ask, but cc one of previous maintainers @ptabor anyway.


> I'm looking a little into `etcd`'s history and it looks like there used to _also_ be a `balancer` package. Maybe the `Metadata` is just a relic of some of that code?


Based on the original two PRs (6 years ago, see below), etcd didn't define `Metadata` at all, instead it just used the grpc-go's [naming.Update](https://github.com/grpc/grpc-go/blob/v1.0.4/naming/naming.go#L50-L58) (which includes a field `Metadata`) directly,
- https://github.com/etcd-io/etcd/pull/5799
- https://github.com/etcd-io/etcd/pull/6632

Previously etcd implemented a custom load balancer, but was replaced with custom resolver + round_robin load balancer, see below PR (4 years ago). **It's exactly when etcd started to depend on grpc-go's experimental resolver package**,
- https://github.com/etcd-io/etcd/pull/12671

Note that from etcd perspective, the [etcd's naming package](https://github.com/etcd-io/etcd/tree/main/client/v3/naming) (which includes `endpoints` and `resolver` packages) is just a utility library (which is based on both etcd's K/V store & watch and grpc-go's resolver) for developers. etcd itself doesn't care about the `Metadata` at all, nor the `Attributes` (added in this PR). As mentioned earlier, all the concepts discussed here (Metadata, Attributes) should be solely coming from grpc-go.

See also https://github.com/etcd-io/etcd/blob/9d246198b5dc5e1b20156de63f26b3eb814aef02/client/v3/naming/endpoints/internal/update.go#L29-L52


The etcd side `Endpoint` (including a `Metadata` field) was added in below PR (4 years ago). The `Endpoint` of etcd side was mapped to grpc-go's `Addresses` (now grpc-go's `Endpoint` after https://github.com/etcd-io/etcd/pull/19780)
- https://github.com/etcd-io/etcd/pull/12614



> Address attributes are mostly internal implementation details, not things we expect users to be doing.

I am a little confused. If they are only internal implementation details, then why does grpc-go expose them? (Based on previous discussion, the only possible users facing usage is to differentiate two addresses with the same IP address). 

The same confusion for `Endpoint.Attributes` (of grpc-go side), why does grpc-go expose it if users don't need to care about it?

Since grpc-go exposes both Metadata and Attributes, does grpc-go have official guide on why & how should users use them? Since Metadata has deprecated, so we can ignore it.

dfawley: > I am a little confused. If they are only internal implementation details, then why does grpc-go expose them? (Based on previous discussion, the only possible users facing usage is to differentiate two addresses with the same IP address).

Hiding this field would be fairly difficult.  But also, it's there for future usage.  A user of the name resolver API might want to add the setting to an address that makes us use a proxy when connecting.  So in theory that API should be experimental and public, too, but our current philosophy is that if nobody asks for a thing, then we generally don't export it.  And nobody has asked yet.

> Since grpc-go exposes both Metadata and Attributes, does grpc-go have official guide on why & how should users use them? Since Metadata has deprecated, so we can ignore it.

Sorry, there are no docs that I know of that go into this, because it is an experimental API and we don't expect many users of it.  But the high-level is:

- `Address.Attributes` are for things that the transport (which we own as an internal implementation detail) or transport credentials can use for anything they want when creating/managing the connection to that address.
- `Endpoint.Attributes` and `resolver.State.Attributes` are for the name resolver to communicate _any_ necessary information to the LB policy (`balancer`).

Internally, gRPC uses all these things for various purposes.  Externally, very very few users should need them.  But  high-level, what it does is allow plugins to communicate with other plugins in different parts of the system.

ahrtr: thx for the explanation.

- The [Address.Metadata](https://github.com/grpc/grpc-go/blob/4cedec40eb2ccfbe3f56bb15e894903111ada2d2/resolver/resolver.go#L117-L121) has deprecated, so etcd should also deprecate it in both 3.5 and 3.6. let's do it in main first, and backport to release-3.6 and release-3.5 later. Eventually let's remove it in 3.7 (main branch).
- Let's do not add Attributes for now, reasons,
  - `Externally, very very few users should need them` as @dfawley mentioned above
  - From etcd side, it isn't clear why & how users should use them. We can add it in the future when it's clear.

amosehiguese: > Let's do not add Attributes for now... 
>

Given that the current relevance of this field is unclear ATM. Not adding it is the only reasonable call to make. 

Just as @dfawley said, "our current philosophy is that if nobody asks for a thing, then we generally don't export it." 

So, since etcd is reliant on grpc, taking a similar stance would be advised. 

That said, I will kindly leave the deprecated metadata field as is but revert all attempts to introduce the Attributes fields.

ahrtr: Yes, this PR should be as simple as just adding a deprecation notice for the Metadata field.

amosehiguese: I love the sound of that. On it!

dfawley: Is there a path to deleting it from etcd, though?  Marking it as deprecated is good, but we ultimately want to delete it.  It would be great if there was some way to find out who might be using it, if anyone, and talk with them.

I guess as an intermediate step for gRPC, we could stop reading the field entirely (e.g. clear it as soon as it comes out of the resolver)?  Maybe with an environment variable to re-enable the behavior in case of emergency?

ahrtr: We will remove the field in 3.7 (main branch). After we backport the change to 3.6 and 3.5, we will remove it from main branch.



> I guess as an intermediate step for gRPC, we could stop reading the field entirely (e.g. clear it as soon as it comes out of the resolver)? Maybe with an environment variable to re-enable the behavior in case of emergency?

I think we are good as long as it doesn't cause any compiling error.

PR #19789, opened 2025-04-23
Update changelog to cover the deprecating resolver.State.Addresses

ahrtr:
Link to
- https://github.com/etcd-io/etcd/pull/19782
- https://github.com/etcd-io/etcd/pull/19783

PR #19791, opened 2025-04-23
mardownlint: allow hr inline tags

ivanvc:
The CHANGELOG uses `<hr />` instead of `---` historically. Adding an exception to `MD033` to allow this HTML tag.

Link to https://github.com/etcd-io/etcd/pull/19789#discussion_r2056719094, #18059.

/cc @ahrtr 

### DIFF
--- a/tools/.markdownlint.jsonc
+++ b/tools/.markdownlint.jsonc
@@ -148,7 +148,7 @@
     // MD033/no-inline-html : Inline HTML : https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md033.md
     "MD033": {
       // Allowed elements
-      "allowed_elements": []
+      "allowed_elements": ["hr"]
     },
   
     // MD034/no-bare-urls : Bare URL used : https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md034.md


tools/.markdownlint.jsonc
```diff
@@ -148,7 +148,7 @@
     // MD033/no-inline-html : Inline HTML : https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md033.md
     "MD033": {
       // Allowed elements
-      "allowed_elements": []
+      "allowed_elements": ["hr"]
```
jmhbnz, line 151: Rather than turning off linting for it I would be in favor of just fixing the root cause with

```bash
for file in $(grep -Ril '<hr>'); do sed -i 's/<hr>/---/g' $file; done
```

Let's use native Markdown where it makes sense and avoid embedding html tags imo.

ahrtr: I suggest to merge this PR for now, so as not to block any update on changelog. 

Please feel free to do whatever bigger change separately. You never know what other existing behaviors may be broken until you try to update it as https://github.com/etcd-io/etcd/pull/19789. Leave it to @ivanvc and @joshjms to followup any following changes.

ahrtr (approved):
cc @fuweid @serathius

tools/.markdownlint.jsonc
jmhbnz: Why make extra work?  We can just fix it with a trivial one line command I have already provided. It would only take a second to run it against the `CHANGELOG` directory and actually fix the problem rather than just add a workaround which we then have to fix later anyway. Let's please just fix it now.

ahrtr: > Why make extra work?

Why did we even add such restriction in the first place? We don't see any problem so far.



> We can just fix it with a trivial one line command I have already provided. It would only take a second to run it against the `CHANGELOG` directory and actually fix the problem rather than just add a workaround which we then have to fix later anyway. Let's please just fix it now.

I am OK as long as we can fix it now (today or this week).

jmhbnz: For motivations refer to discussion in:
- https://github.com/etcd-io/website/issues/857
- https://github.com/etcd-io/etcd/issues/18059#issuecomment-2436062946.  

Long story short last year we added linting for Markdown files for our `etcd-io/website` documentation, and since that has been running successfully for a while we have now extended the approach to any Markdown documentation we maintain within our `etcd-io/etcd` main repository.

The new linting only applies to Markdown files when they are edited, so as to reduce impact where possible. The intent is to gradually improve, standardise and follow best practices with our documentation files over time.

> I am OK as long as we can fix it now (today or this week).

Thanks - yes it should be very quick to update this pr and merge the root cause fix.

joshjms: Got it, to make it quick, I'll just make a PR for this

ahrtr: > Got it, to make it quick, I'll just make a PR for this (since i'm already at my keyboard). I'll just do it for all markdown files ok?

It should be OK. Thanks for the good collaboration.

ahrtr: If you look at the linter logs, there are lots of violations

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19789/pull-etcd-markdown-lint/1915117651537956864

I suggest that we just fix the problem (replace `<hr />` with `---`) discussed above for now. For all other violation, please feel free to followup separately. thx

jmhbnz:
Closing in favor of https://github.com/etcd-io/etcd/pull/19794

PR #19808, opened 2025-04-26
Split executeRange function

serathius:


fuweid (approved):
LGTM on green

ahrtr:
I'm not a big fan of this refactoring style. Breaking one method into many small ones often hurts readability, as it forces readers to jump around the codebase more. 

This small refactor seems harmless, but another similar bad example is the refactoring on [NewServer](https://github.com/etcd-io/etcd/blob/29e53f5e8e3745ee8498916d815987cd47a2c858/server/etcdserver/server.go#L334) , which splits one big (but easy to understand) method into too many small methods/functions. Based on my years of comparing the source code between 3.5 and 3.6, the NewServer of 3.5 is much much easier to understand than 3.6.

Also any change on the core applying workflow should allow more time to review!

serathius:
> Breaking one method into many small ones often hurts readability, as it forces readers to jump around the codebase more.

The goal isn't just more functions, but better abstraction. Well-defined, descriptively named sub-functions can clarify higher-level logic, allowing readers to understand the main flow without needing to read every underlying implementation detail immediately. Poor abstractions can indeed cause the issue you described.

Regarding NewServer (etcd-io/etcd/pull/13230): while a large constructor might seem familiar, refactoring it, even if the current abstractions aren't perfect, helped uncover issues like circular dependencies and improved testability. This paves the way for further simplification. The executeRange refactor is a much smaller, more contained case, making it easier to get the abstractions right.

For executeRange, splitting the ~100-line function into clear steps (calculate limit, fetch, filter, sort, prepare response) aims to make the overall process more evident at a glance. Each sub-function has a single, clear responsibility.

serathius:
> Also any change on the core applying workflow should allow more time to review!

Code changes were reviewed by @fuweid, one of the benefit of simple refactors like this is that they don't change any business logic thus are much safer and can be reviewed faster.

fuweid:
Today I reviewed several pull requests in the core package. Many of them focus on cleanup—reducing arguments, splitting logic into different files, and so on. From a reviewer’s perspective, these changes seem relatively safe since we have tests to cover them.

Whether to keep a function large or split it into smaller ones is always a tradeoff and depends on the context. A large function might span more than one screen, which still requires scrolling or jumping. On the other hand, small, well-named functions can help by clearly indicating each step. For me, this `executeRange` function has five distinct stages:

* Check limitations
* Perform RangeKeys
* Filter results
* Sort if needed
* Return the final result

The flow is clear and understandable. If a larger function lacks this kind of clear structure, then breaking it into smaller functions might not help readability.

Anyway, in terms of code style, I believe keeping the exposed API stable is more important than internal refactoring. Internal code can always be adapted later to support new features when needed.

:)

ahrtr:
> For executeRange, splitting the ~100-line function into clear steps (calculate limit, fetch, filter, sort, prepare response) aims to make the overall process more evident at a glance. Each sub-function has a single, clear responsibility.

It's a little over-engineering. I am not against encapsulation or abstraction, we should avoid too small granularity encapsulation.

This PR isn't that obvious, but we should try to avoid such trend.

siyuanfoundation:
I agree we should avoid too small granularity encapsulation. A good encapsulation should be functionally complete and independent with descriptive names, which is well represented in this PR. 

Breaking down a large function helps people, especially new contributors to understand the purposes and boundaries of different sections. Overall I think it improves readability.

PR #19813, opened 2025-04-26
Simplify apply by handling membership apply through the same applyV3 stack

serathius:

### DIFF
--- a/server/etcdserver/apply/uber_applier.go
+++ b/server/etcdserver/apply/uber_applier.go
@@ -21,13 +21,14 @@ import (
 	"go.uber.org/zap"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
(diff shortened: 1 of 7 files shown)


serathius:
PTAL @fuweid

server/etcdserver/apply/uber_applier.go
```diff
@@ -133,6 +133,31 @@ func (a *uberApplier) dispatch(r *pb.InternalRaftRequest) *Result {
...
+	case r.DowngradeVersionTest != nil:
+		op = "DowngradeVersionTest" // Implemented in 3.6 for test only
+		// do nothing, we are just to ensure etcdserver don't panic in case
+		// users(test cases) intentionally inject DowngradeVersionTestRequest
+		// into the WAL files.
+		return ar
+	default:
+	}
+	if !shouldApplyV3 {
+		return ar
```
fuweid, line 158: to align with orginal `applyInternalRaftRequest` function, we should return nil here

PR #19815, opened 2025-04-28
Add test-release-tests Makefile target

joshjms:

Moving scripts from Prow job (https://github.com/kubernetes/test-infra/blob/master/config/jobs/etcd/etcd-presubmits.yaml) to `test.sh`. 

This is for ease of migrating `release-3.4` and `release-3.5` Github Actions workflows to Prow.

ref: https://github.com/kubernetes/test-infra/issues/32754

/cc @ivanvc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,10 @@ test-grpcproxy-e2e: build
 test-e2e-release: build
 	PASSES="release e2e" ./scripts/test.sh $(GO_TEST_FLAGS)
 
+.PHONY: test-release
+test-release:
+	PASSES="release_tests" CI=$$CI ./scripts/test.sh $(GO_TEST_FLAGS)
+
 .PHONY: test-robustness
 test-robustness:
 	PASSES="robustness" ./scripts/test.sh $(GO_TEST_FLAGS)
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -623,6 +623,31 @@ function release_pass {
   mv /tmp/etcd ./bin/etcd-last-release
 }
 
+function release_tests_pass {
+  VERSION=$(go list -m go.etcd.io/etcd/api/v3 2>/dev/null | awk '{split(substr($2,2), a, "."); print a[1]"."a[2]".99"}')
+
+  if [ -n "${CI:-}" ]; then
+    git config user.email "prow@etcd.io"
+    git config user.name "Prow"
+
+    gpg --batch --gen-key <<EOF
+%no-protection
+Key-Type: 1
+Key-Length: 2048
+Subkey-Type: 1
+Subkey-Length: 2048
+Name-Real: Prow
+Name-Email: prow@etcd.io
+Expire-Date: 0
+EOF
+
+    git remote add origin https://github.com/etcd-io/etcd.git
+  fi
+
+  DRY_RUN=true run "${ETCD_ROOT_DIR}/scripts/release.sh" --no-upload --no-docker-push --no-gh-release --in-place "${VERSION}"
+  VERSION="${VERSION}" run "${ETCD_ROOT_DIR}/scripts/test_images.sh"
+}
+
 function mod_tidy_for_module {
   run go mod tidy -diff
 }


Makefile
```diff
@@ -56,6 +56,10 @@ test-grpcproxy-e2e: build
 test-e2e-release: build
 	PASSES="release e2e" ./scripts/test.sh $(GO_TEST_FLAGS)
 
+.PHONY: test-release-tests
+test-release-tests:
```
serathius, line 60:
```suggestion
.PHONY: test-release
test-release:
```

serathius (dismissed):

Makefile
joshjms: Won't it be too similar to `test-e2e-release`?

```sh
.PHONY: test-e2e-release
test-e2e-release: build
	PASSES="release e2e" ./scripts/test.sh $(GO_TEST_FLAGS)
```

Considering both have completely different workflows (and purpose), I thought it might (?) be misleading?

I'd be happy to change it though.

joshjms:
Wait let me test it first on kind before I remove the draft

Makefile
serathius: I think it's ok as naming is hierarchical, so `e2e` should be treated as much important than `release`.

joshjms: Ah I understand, will change

joshjms: Done!

scripts/test.sh
```diff
@@ -623,6 +623,26 @@ function release_pass {
...
+Key-Type: 1
+Key-Length: 2048
+Subkey-Type: 1
+Subkey-Length: 2048
+Name-Real: Prow
+Name-Email: prow@etcd.io
+Expire-Date: 0
+EOF
+
+  git remote add origin https://github.com/etcd-io/etcd.git
```
ivanvc, line 641: The issue I see here is that running these tests locally will mess with the local developer's Git setup. Would it make sense to wrap these lines in an if statement checking if `${CI}` is set?

scripts/test.sh
```diff
@@ -623,6 +623,26 @@ function release_pass {
...
+  DRY_RUN=true run "${ETCD_ROOT_DIR}/scripts/release.sh" --no-upload --no-docker-push --no-gh-release --in-place 3.6.99
+  VERSION=3.6.99 run "${ETCD_ROOT_DIR}/scripts/test_images.sh"
```
ivanvc, line 643: Can we read the current minor from `./api/version/version.go` and set the patch to 99? That way, they can cleanly backport to 3.6 and 3.5 [3.4 has a different file structure, so maybe it makes sense to fix it to 3.4.99]).

It may sound like an overkill, but tests will fail once we release 3.6 and start working on 3.7 on the main branch.

ivanvc (commented):
Thanks, @joshjms. Please take a look at my comments :)

scripts/test.sh
joshjms: Right that makes sense. Thanks for pointing that out.

scripts/test.sh
joshjms: Will do!

scripts/test.sh
joshjms: Done, I also set the export CI=true in the Prow job.

joshjms:
Still some work needs to be done for release-3.4 in the `test.sh` (only changing from `scripts/test.sh` to `test` actually) but generally it works fine for `release-3.6` and `release-3.5`.

scripts/test.sh
serathius: Ups, I missed that.

ivanvc (commented):
Thanks for the updates, @joshjms. I left some comments based on your updates.

scripts/test.sh
```diff
@@ -623,6 +623,35 @@ function release_pass {
   mv /tmp/etcd ./bin/etcd-last-release
 }
 
+function release_tests_pass {
+  VERSION=3.4.99
+
+  if [ "${1-}" ]; then
+    VERSION=$1
+  fi
```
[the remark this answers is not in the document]
joshjms: Got it!

scripts/test.sh
```diff
@@ -623,6 +623,35 @@ function release_pass {
...
+
+  if [ "${CI}" ]; then
```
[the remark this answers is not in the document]
joshjms: Right!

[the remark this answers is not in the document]
joshjms: fixed!

scripts/test.sh
```diff
@@ -623,6 +623,35 @@ function release_pass {
   mv /tmp/etcd ./bin/etcd-last-release
 }
 
+function release_tests_pass {
+  VERSION=3.4.99
```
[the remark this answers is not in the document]
joshjms: In the case of release-3.4, `go list -m go.etcd.io/etcd/api/v3` will not work, so I misunderstood your earlier review

> 3.4 has a different file structure, so maybe it makes sense to fix it to 3.4.99

Now, I'll just let it be that (I moved the Makefile line 61 to test.sh) and for release-3.4 I will adjust the test.sh accordingly (make it hardcoded at 3.4.99)

joshjms:
@ivanvc I addressed all your reviews, ptal and let me know if I should fix anything else. Thanks for the reviews!

Makefile
```diff
@@ -56,6 +56,10 @@ test-grpcproxy-e2e: build
 test-e2e-release: build
 	PASSES="release e2e" ./scripts/test.sh $(GO_TEST_FLAGS)
 
+.PHONY: test-release
+test-release:
+	PASSES="release_tests" CI=$$CI ./scripts/test.sh $(GO_TEST_FLAGS)
```
ivanvc, line 61: There's no need to downstream the `CI` environment variable.

scripts/test.sh
```diff
@@ -623,6 +623,31 @@ function release_pass {
   mv /tmp/etcd ./bin/etcd-last-release
 }
 
+function release_tests_pass {
+  VERSION=$(go list -m go.etcd.io/etcd/api/v3 2>/dev/null | awk '{split(substr($2,2), a, "."); print a[1]"."a[2]".99"}')
```
ivanvc, line 627: I would either create a local variable for this or support overriding the `VERSION` variable (I would go with the latter for cases such as 3.4).

```sh
  local version
  version=$(go list -m go.etcd.io/etcd/api/v3 2>/dev/null | awk '{split(substr($2,2), a, "."); print a[1]"."a[2]".99"}')
```

```sh
  if [ -z "${VERSION:-}" ]; then
    VERSION=$(go list -m go.etcd.io/etcd/api/v3 2>/dev/null | awk '{split(substr($2,2), a, "."); print a[1]"."a[2]".99"}')
  fi
```

ivanvc (approved):
Given https://github.com/kubernetes/test-infra/pull/34740, I suggest we merge it and address improvements in a follow-up PR. Otherwise, the CI will start failing for all new pull requests. Thanks, @joshjms.

/cc @jmhbnz

ivanvc:
Note that we must backport this to `release-3.6`; otherwise, the CI will fail for that branch.

scripts/test.sh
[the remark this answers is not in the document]
ivanvc: Refer to https://github.com/etcd-io/etcd/pull/19815#discussion_r2069437868

jmhbnz (approved):
Thanks @joshjms 

Let's address the final tweaks in a follow-up pr as I missed adding a `/hold` on the associated `k/test-infra` pr and we don't want tests to start failing!

joshjms:
Thanks @ivanvc @jmhbnz , I will address the new reviews.

scripts/test.sh
joshjms: Ah got it, I agree to support overriding the `VERSION` variable for release-3.4

Issue #19050, opened 2024-12-12
migrate `--experimental-enable-distributed-tracing` flag to feature gate


siyuanfoundation:
### What would you like to be added?

Migrate `--experimental-enable-distributed-tracing` flag to use feature gate

Example for migrating an experimental flag: https://github.com/etcd-io/etcd/pull/19053
Also need to create new flags for this feature like `--experimental-distributed-tracing-address` -> `--distributed-tracing-address` and mark the old flag deprecated

### Why is this needed?

part of https://github.com/etcd-io/etcd/issues/18023

serathius:
For distributed tracing I don't think we want to add a feature gate as the feature should stay optional in the future. I think we just want to migrate the flag of the `--experimental` prefix.

omerap12:
> For distributed tracing I don't think we want to add a feature gate as the feature should stay optional in the future. I think we just want to migrate the flag of the `--experimental` prefix.

Understood.

PR #19843, opened 2025-04-30
remove DistributedTracing from feature gate list.

siyuanfoundation:
Because the feature is not feature gated.

See https://github.com/etcd-io/etcd/issues/19050#issuecomment-2553103000 for context 

### DIFF
--- a/server/embed/config_test.go
+++ b/server/embed/config_test.go
@@ -111,7 +111,6 @@ func TestConfigFileFeatureGates(t *testing.T) {
 		{
 			name: "default",
 			expectedFeatures: map[featuregate.Feature]bool{
-				features.DistributedTracing:           false,
 				features.StopGRPCServiceOnDefrag:      false,
 				features.InitialCorruptCheck:          false,
 				features.TxnModeWriteWithSharedBuffer: true,
@@ -139,12 +138,12 @@ func TestConfigFileFeatureGates(t *testing.T) {
 		},
 		{
 			name:                                "ok to set different experimental flag and feature gate flag",
-			serverFeatureGatesJSON:              "DistributedTracing=true",
+			serverFeatureGatesJSON:              "InitialCorruptCheck=true",
 			experimentalStopGRPCServiceOnDefrag: "true",
 			expectedFeatures: map[featuregate.Feature]bool{
-				features.DistributedTracing:           true,
 				features.StopGRPCServiceOnDefrag:      true,
 				features.TxnModeWriteWithSharedBuffer: true,
+				features.InitialCorruptCheck:          true,
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


ahrtr:
> Because the feature is not feature gated.
> 
> See [#19050 (comment)](https://github.com/etcd-io/etcd/issues/19050#issuecomment-2553103000) for context

So the intention is not to feature gate a feature if it should be still optional (users can still enable or disable it) after GA? In that case, How should we manage the lifecycle of such feature in future? Note that this is a generic question, not specific to the DistributedTracing feature.

It seems that the only way is to add a flag without `--experimental` flag, and add comment above it to describe the phase (i.e. alpha, beta, GA, etc)?

ahrtr:
@ivanvc we need to backport this PR to 3.6 as well.

@siyuanfoundation can you review all such features, so that we don't miss any other similar cases? thx

siyuanfoundation:
> > Because the feature is not feature gated.
> > See [#19050 (comment)](https://github.com/etcd-io/etcd/issues/19050#issuecomment-2553103000) for context
> 
> So the intention is not to feature gate a feature if it should be still optional (users can still enable or disable it) after GA? In that case, How should we manage the lifecycle of such feature in future? Note that this is a generic question, not specific to the DistributedTracing feature.
> 
> It seems that the only way is to add a flag without `--experimental` flag, and add comment above it to describe the phase (i.e. alpha, beta, GA, etc)?

Yes, the intention is not to feature gate a feature if it should be still optional (users can still enable or disable it) after GA. The idea is having the boolean flag in the experimental section, and move it out of the section once its ready to GA
/cc @serathius

siyuanfoundation:
> @ivanvc we need to backport this PR to 3.6 as well.
> 
> @siyuanfoundation can you review all such features, so that we don't miss any other similar cases? thx

`DistributedTracing` is the only feature that was not migrated as feature gate, but the feature gate was created for testing. 

For other features, the decision is not irreversible either. If we decide we need to provide an option to disable InitialCorruptCheck after GA, we can add a boolean flag as well when it GAs.

ahrtr:
> The idea is having the boolean flag in the experimental section, and move it out of the section once its ready to GA

This is also a breaking change.

ivanvc (approved):
LGTM. Thanks, @siyuanfoundation.

@ahrtr, 3.6's CHANGELOG does not mention the removal of `ExperimentalEnableDistributedTracing` in favor of `EnableDistributedTracing`. Do you think we should add a note?

ahrtr:
> @ahrtr, 3.6's CHANGELOG does not mention the removal of `ExperimentalEnableDistributedTracing` in favor of `EnableDistributedTracing`. Do you think we should add a note?

When we add feature gate for `EnableDistributedTracing` , we did not add changelog. Now we reverted the change, probably we shouldn't add changelog either.

I migrated several experimental flags to either feature gate or flags (without "experimental" prefix), I think it'd better to update changelog to cover all such changes.  Probably we can have a combined/unified changelog when we release the final v3.6.0? It's up to you and @jmhbnz

PR #19852, opened 2025-05-02
Use robustness traffic generation

### LINKED ISSUE #19839: [Antithesis] Use robustness traffic generation

Replace the existing code in [delete-keys](https://github.com/etcd-io/etcd/blob/main/tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go) to use robustness code for generating traffic.

Traffic generator https://github.com/etcd-io/etcd/blob/main/tests/robustness/traffic/etcd.go

Initial configuration:
```
	EtcdPut Traffic = etcdTraffic{
		keyCount:     10,
		// Please keep the sum of weights equal 100.
		requests: []random.ChoiceWeight[etcdRequestType]{
			{Choice: Get, Weight: 50},
			{Choice: Put, Weight: 50},
		},
	}
```

To generate traffic you can check the code:
https://github.com/etcd-io/etcd/blob/c590dd6c58bc10ea1e03b54683c12861bf4b114c/tests/robustness/traffic/traffic.go#L77-L90

Let's use 3 concurrent clients.

Should run for random time 5s-60s

### Why is this needed?

Next step in robustness integration

nwnt:
Fix #19839 

This currently includes exporting etcdTraffic so it can be reused by the antithesis robustness tests.

### DIFF
--- a/tests/robustness/traffic/etcd.go
+++ b/tests/robustness/traffic/etcd.go
@@ -75,6 +75,16 @@ var (
 			{Choice: Delete, Weight: 50},
 		},
 	}
+	EtcdAntithesis Traffic = etcdTraffic{
+		keyCount:     10,
+		largePutSize: 32769,
+		leaseTTL:     DefaultLeaseTTL,
+		// Please keep the sum of weights equal 100.
+		requests: []random.ChoiceWeight[etcdRequestType]{
+			{Choice: Get, Weight: 50},
+			{Choice: Put, Weight: 50},
+		},
+	}
 )
 
 type etcdTraffic struct {
--- a/tests/antithesis/test-template/robustness/main.go
+++ b/tests/antithesis/test-template/robustness/main.go
@@ -18,13 +18,12 @@ package main
 
 import (
 	"context"
-	"log"
+	"fmt"
 	"os"
 	"sync"
 	"time"
 
 	"github.com/antithesishq/antithesis-sdk-go/assert"
-	"github.com/antithesishq/antithesis-sdk-go/random"
 	"golang.org/x/time/rate"
 
 	"go.etcd.io/etcd/tests/v3/robustness/client"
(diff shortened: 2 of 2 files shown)


tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go
```diff
@@ -45,77 +49,64 @@ func Connect() *client.RecordingClient {
...
+		wg.Add(1)
+		c := Connect()
+		go func(c *client.RecordingClient) {
+			defer wg.Done()
+			defer c.Close()
+
+			etcdTraffic.RunTrafficLoop(ctx, c, limiter,
+				identity.NewIDProvider(),
+				identity.NewLeaseIDStorage(),
+				traffic.NewConcurrencyLimiter(profile.MaxNonUniqueRequestConcurrency),
```
serathius, line 100: Those three should be shared by all clients. There should be one source of identity for used by all clients, same for storage and concurrency.

tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go
```diff
@@ -45,77 +49,64 @@ func Connect() *client.RecordingClient {
...
-		_, err := cli.Delete(ctx, k)
-		// Antithesis Assertion: sometimes delete requests are successful. A failed request is OK since we expect them to happen.
-		assert.Sometimes(err == nil, "Client can make successful delete requests", map[string]any{"error": err})
-		if err != nil {
-			log.Printf("Failed to delete key %s: %v", k, err)
-		} else {
-			log.Printf("Successfully deleted key %v", k)
-			deletedKeys = append(deletedKeys, k)
+// wrap converts a receive-only channel to receive-only struct{} channel
+func wrap[T any](from <-chan T) <-chan struct{} {
```
serathius, line 70: Note on style. Not blocking this PR. As we are not writing C, there is no limit on order of functions. Best to write them in order of reading. From most important to least important. From top function in hierarchy to function at the bottom. So `wrap` should be after `testRobustness` and `testRobustness` should be after `main`.

tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go
```diff
@@ -45,77 +49,64 @@ func Connect() *client.RecordingClient {
...
+				finish,
+			)
+			mux.Lock()
+			reports = append(reports, c.Report())
+			mux.Unlock()
+		}(c)
 	}
-	cli.Close()
-
-	assert.Reachable("Completion of a key deleting check", nil)
```
serathius, line 115: Can you keep a assert like this, just change the text. Like `Completion of robustness traffic generation`.

serathius:
Looks good, nice work.

Left couple of comments mostly nits.

tests/robustness/traffic/etcd.go
```diff
@@ -65,56 +65,56 @@ var (
...
 	}
 )
 
-type etcdTraffic struct {
-	keyCount     int
-	requests     []random.ChoiceWeight[etcdRequestType]
-	leaseTTL     int64
-	largePutSize int
+type EtcdTraffic struct {
+	KeyCount     int
```
serathius, line 81: Making the struct public make sense, however we should consider if it's convenient and safe to be used by users outside this package. It's good to consider that users might not know some internal assumptions about the fields.

Instead of making all fields in struct public we could create a function working as a constructor. This is safer as it allows us to add some logic on construction like validation for fields that would prevent panics. Maybe in the future would be good to check keyCount, LargePutSize and leaseTTL are not negative etcd.

tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go
nwnt: done.

tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go
nwnt: Roger that.  Will keep that in mind for future PRs.

tests/antithesis/test-template/go-delete-keys/serial_driver_delete_keys.go
nwnt: Done.

tests/robustness/traffic/etcd.go
nwnt: Done, I added `NewEtcdTraffic` in adding to exporting `EtcdTraffic`.  I still have the fields public for now.  Let me know if you want me to turn them back to private.

nwnt:
@serathius thank you for the review comments, they should have all been fixed.  it's ready for another review :)

serathius:
Applied some small patches to minimize back and forth due to timezones. Some changes:
* baseTime also needs to be reused between clients
* we should avoid global variables other than configuration
* we shouldn't use generic functions if not needed. `wrap` is an overkill for one case where we need to wait some time and close channel.
* Making all fields in traffic was too much, we just need a separate traffic shape.

serathius (approved):

PR #19856, opened 2025-05-04
Refactor: etcdserver-move rafthttp.Transport initialization to bootstrap

carpewang:
### Descriptions:
* This PR addresses the existing TODO comment in the NewServer function:`// TODO: move transport initialization near the definition of remote.`
* This PR refactors the server initialization process by decoupling the Raft consensus component from the initial bootstrap phase. Previously, Raft was initialized within the bootstrap function, which tightly coupled cluster setup with consensus mechanics. 
### Changes Include:
* Moved the initialization and startup of rafthttp.Transport into the bootstrap() function.

* The transport instance is now part of the returned bootstrappedServer struct.

* Refactored NewServer() to simply assign the bootstrapped transport and wire it up.

### DIFF
--- a/server/etcdserver/bootstrap.go
+++ b/server/etcdserver/bootstrap.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	stats "go.etcd.io/etcd/server/v3/etcdserver/api/v2stats"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


carpewang:
Hi @spzala, could you please review this PR and comment `/ok-to-test` if it's reasonable to test?

I’ve already verified that it passes all tests locally.

Also, if everything looks good, please consider approving with `/approve`.

Thanks!

PR #19878, opened 2025-05-07
Use first_ command type to ensure antithesis runs robustness tests only once

serathius:


serathius:
Merging as this is trivial change to antithesis and we are still iterating on config.

PR #19892, opened 2025-05-08
Allow retryable check of database empty and enable failure injection in Antithesis

serathius:
In Antithesis we don't have control of failure injection. So we don't have guarantee that the first request will pass.

Changing first request to read fixes the issue of first request not being retryable. Changing the test name prefix to "singleton_driver_" enables antithesis fault injection.

/cc @siyuanfoundation @fuweid @nwnt @henrybear327


siyuanfoundation (approved):

serathius:
Needed to fix tests

PR #19889, opened 2025-05-08
[Antithesis] Get access to etcd data directory

nwnt:
Fix #19882 

This will likely have a conflict with #19887.  Once that is merged, I can rebase this PR and resolve the conflict.

PR #19897, opened 2025-05-09
[Antithesis] store validation on host volume

### LINKED ISSUE #19883: [Antithesis] Store the validation visualization in the host volume

We can volume mount in client container to host.  For example `artifacts` directory (local to docker-compose).

Visualization should write to this directory by changing https://github.com/etcd-io/etcd/blob/297e9d1d586a12388c8ec49d3bf5e36803064ae0/tests/antithesis/test-template/robustness/main.go#L63

### Why is this needed?

Have report uploaded

nwnt:
Fix #19883 

Will very likely have a conflict with #19889, but should be easily rebased on top of it.

### DIFF
--- a/tests/antithesis/test-template/robustness/main.go
+++ b/tests/antithesis/test-template/robustness/main.go
@@ -51,6 +51,8 @@ const (
 	localetcd0 = "127.0.0.1:12379"
 	localetcd1 = "127.0.0.1:22379"
 	localetcd2 = "127.0.0.1:32379"
+
+	reportPath = "/var/report/histor.html"
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


tests/antithesis/test-template/robustness/main.go
```diff
@@ -51,6 +51,8 @@ const (
 	localetcd0 = "127.0.0.1:12379"
 	localetcd1 = "127.0.0.1:22379"
 	localetcd2 = "127.0.0.1:32379"
+
+	reportPath = "/var/report/histor.html"
```
serathius, line 55:
```suggestion
	reportPath = "/var/report/history.html"
```

serathius, line 55: @marcus-hodgson-antithesis I know you did some work with uploading report and it's already working. Will this break the integration you did?

marcus-hodgson-antithesis: @serathius yes. The current logic is to extract "/build/tests/history.html". If you'd like me to change that to "/var/report/history.html" or that entire directory, we can easily change it on our side.

marcus-hodgson-antithesis: We can compress the entire "reports" directory (if there will be multiple history.html files) and provide that as an artifact, or just the "/var/report/history.html". Let me know when to switch this over, but for now I will keep it as "/build/tests/history.html" until this is merged

serathius:
Added to https://github.com/etcd-io/etcd/pull/19889 to fix issue with saving history.

PR #19902, opened 2025-05-09
Reduce runtime of antithesis tests to 15s

serathius:


fivemtebex:
Very clean code with good test coverage. This is a great example of best practices.

fivemhostingserver:
Interesting solution to this problem! I've been working on similar challenges recently.

Issue #19884, opened 2025-05-07
Investigate robustness test flake


serathius:
### Which Github Action / Prow Jobs are flaking?

periodic robustness test, discovered during robustness test meeting.

### Which tests are flaking?

Robustness test

### Github Action / Prow Job link

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-etcd-robustness-main-amd64/1919826806836301824 

### Reason for failure (if possible)

Broken validation assumptions

### Anything else we need to know?

/cc @henrybear327 @nwnt

carsontham:
hi, adding on that [similar flaky test](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19913/ci-etcd-robustness-release36-amd64/1926313175934832640) were observed on TestRobustnessExploratory, triggered from this [PR here. ](https://github.com/etcd-io/etcd/pull/19913)

siyuanfoundation:
@GunaKKIBM do you still plan to look into this?

serathius:
We no longer have data on the flake :(

PR #19913, opened 2025-05-10
client/pkg: fix revive unexported-return issue

carsontham:
Fixes part of #18370

This PR updates the `client/pkg` to fix unexported-return linter issue.

In this pull request, I introduced a `UnsafeSet` interface that contains the original `Set` interface, along with a method ContainsAll(). This was to avoid making changes to this unsafeSet unit-test [here](https://github.com/etcd-io/etcd/blob/main/client/pkg/types/set_test.go#L161). 

The changes are:
- NewUnsafeSet() now returns an `UnsafeSet` interface, instead of `*unsafeSet` struct
- NewThreadsafeSet now returns a `Set` interface, instead of `*tsafeSet` struct

### DIFF
--- a/client/pkg/types/set.go
+++ b/client/pkg/types/set.go
@@ -31,7 +31,12 @@ type Set interface {
 	Sub(Set) Set
 }
 
-func NewUnsafeSet(values ...string) *unsafeSet {
+type UnsafeSet interface {
+	Set
+	ContainsAll(values []string) bool
+}
+
+func NewUnsafeSet(values ...string) UnsafeSet {
 	set := &unsafeSet{make(map[string]struct{})}
 	for _, v := range values {
 		set.Add(v)
@@ -39,10 +44,12 @@ func NewUnsafeSet(values ...string) *unsafeSet {
 	return set
 }
 
-func NewThreadsafeSet(values ...string) *tsafeSet {
+func NewThreadsafeSet(values ...string) Set {
 	us := NewUnsafeSet(values...)
 	return &tsafeSet{us, sync.RWMutex{}}
 }
+
+var _ UnsafeSet = (*unsafeSet)(nil)
 
 type unsafeSet struct {
 	d map[string]struct{}
@@ -122,8 +129,10 @@ func (us *unsafeSet) Sub(other Set) Set {
 	return result
 }
 
+var _ Set = (*tsafeSet)(nil)
+
 type tsafeSet struct {
-	us *unsafeSet
+	us Set
 	m  sync.RWMutex
 }
 


carsontham:
hi @jmhbnz, I linked the flaky test triggered in this PR to one of the [existing open issues ](https://github.com/etcd-io/etcd/issues/19884) as per the etcd contribution guide. could we rerun the test for this PR? thanks!

ivanvc (approved):
LGTM. I'm sorry for not getting back to you sooner. Thanks, @carsontham.

/cc @ahrtr

client/pkg/types/set.go
```diff
@@ -31,19 +31,26 @@ type Set interface {
 	Sub(Set) Set
 }
 
-func NewUnsafeSet(values ...string) *unsafeSet {
+type UnsafeSet interface {
+	Set
+	ContainsAll(values []string) bool
```
ahrtr: `ContainsAll` should be part of `Set`. But a better choice is to change `Contains(string) bool` to `Contains(val ...string) bool`,
- It's backward compatible.
- It also covers the functionality of `ContainsAll`

But in order to be backward compatible, we need to keep `ContainsAll` (add it into `Set`), but mark it as deprecated, and recommend users to use `Contains(val ...string) bool` instead.

carsontham: hi @ahrtr, i have updated the change. please take a look, thanks!

client/pkg/types/set.go
```diff
@@ -23,27 +23,31 @@ import (
...
 	Add(string)
 	Remove(string)
-	Contains(string) bool
+	Contains(val ...string) bool
 	Equals(Set) bool
 	Length() int
 	Values() []string
 	Copy() Set
 	Sub(Set) Set
+	// ContainsAll is Deprecated. Use Contains(val ...string) bool instead.
```
ahrtr: Please follow example below to add deprecated,

https://github.com/etcd-io/etcd/blob/effaa795086bd7a09203447fc55e3821e93f7181/server/embed/config.go#L406C5-L406C15

ahrtr (commented):
pls also squash the commits

client/pkg/types/set.go
```diff
@@ -59,12 +63,17 @@ func (us *unsafeSet) Remove(value string) {
...
+	for _, value := range values {
+		if _, exists := us.d[value]; exists {
+			return true
+		}
+	}
+	return false
 }
 
 // ContainsAll returns whether the set contains all given values
+// ContainsAll is deprecated. Use Contains(val ...string) bool instead.
```
ahrtr: ditto

client/pkg/types/set.go
carsontham: i have followed the example and fixed it. thanks

client/pkg/types/set.go
carsontham: fixed here too.

client/pkg/types/set.go
```diff
@@ -139,10 +152,21 @@ func (ts *tsafeSet) Remove(value string) {
...
+}
+
+// ContainsAll returns whether the set contains all given values
+// Deprecated: Use Contains instead.
+func (ts *tsafeSet) ContainsAll(values []string) bool {
+	for _, s := range values {
+		if !ts.Contains(s) {
+			return false
+		}
+	}
```
ahrtr:
```suggestion
	ts.m.RLock()
	defer ts.m.RUnlock()
	return ts.us.ContainsAll()
```

client/pkg/types/set.go
```diff
@@ -122,8 +133,10 @@ func (us *unsafeSet) Sub(other Set) Set {
 	return result
 }
 
+var _ Set = (*tsafeSet)(nil)
+
 type tsafeSet struct {
-	us *unsafeSet
+	us Set
```
ahrtr, line 143: pls revert this change. It should be explicitly `unsafeSet` instead of Set, otherwise you might run into indefinitely loop if it's a tsafeSet instance.

carsontham: hi @ahrtr, for unsafeSet, they are created in the method [NewUnsafeSet()](https://github.com/etcd-io/etcd/blob/main/client/pkg/types/set.go#L34-L40.) But this method is the one that is causing the unexported-return linter issue.

What do you think if I introduce an UnsafeSet interface and update the tsafe struct? that way we can still explicitly keep a unsafeSet while resolving the linter issue.

```
type UnsafeSet interface {
	Set
}

type tsafeSet struct {
	us UnsafeSet
	m  sync.RWMutex
}

func NewUnsafeSet(values ...string) UnsafeSet {
         ....
}
```

ahrtr: The problem for now is that the users (callers) always get a `Set` interface, so they don't know whether or not it's threadsafe.
 
I suggest that we define a `ThreadsafeSet`, so we will have two interfaces:
- The existing Set: which isn't threadsafe
- The new `ThreadsafeSet`, which is threadSafe.

carsontham: hi @ahrtr, i have added the ThreadsafeSet interface, pls take another look. thanks!

client/pkg/types/set.go
```diff
@@ -59,12 +69,17 @@ func (us *unsafeSet) Remove(value string) {
 }
 
 // Contains returns whether the set contains the given value
-func (us *unsafeSet) Contains(value string) (exists bool) {
-	_, exists = us.d[value]
-	return exists
+func (us *unsafeSet) Contains(values ...string) (exists bool) {
+	for _, value := range values {
+		if _, exists := us.d[value]; exists {
+			return true
```
ahrtr: This isn't correct, it should contains all rather than any one item.

client/pkg/types/set.go
```diff
@@ -122,8 +137,10 @@ func (us *unsafeSet) Sub(other Set) Set {
 	return result
 }
 
+var _ ThreadsafeSet = (*tsafeSet)(nil)
+
 type tsafeSet struct {
-	us *unsafeSet
+	us ThreadsafeSet
```
ahrtr: The comment applies here as well. 

Please use `*unsafeSet` or `Set`

client/pkg/types/set.go
carsontham: i have fixed in latest commit

client/pkg/types/set.go
carsontham: fixed in latest commit

siyuanfoundation:
@ahrtr Can you review again? Thanks!

client/pkg/types/set.go
ahrtr: This suggestion is to suggest you replace the existing implementation with the proposed code. You just added the proposed code, but did not remove the old implementation.

ahrtr:
The change looks good now. Thanks

Please squash the commits and rebase this PR.

ahrtr:
Please signoff the commit, thx
```
git rebase HEAD~1 --signoff
git push --force-with-lease origin ct/fix-client-pkg-unexported-return
```

ahrtr:
Probably we need a changelog item for this change cc @carsontham @ivanvc

PR #19915, opened 2025-05-11
[Antithesis] dump result to robustness report

### LINKED ISSUE #19899: [Antithesis] Implement dumping the whole robustness report

Currently the antithesis tests just write the visualization https://github.com/etcd-io/etcd/blob/49767381e803cc45c308fabea3c636ebb20f1d2f/tests/antithesis/test-template/robustness/main.go#L81

However this is not enough to re-evaluate issues locally the way that robustness tests allow https://github.com/etcd-io/etcd/tree/main/tests/robustness#re-evaluate-existing-report

To get that we would need to create a `TestReport` and use it to safe all the request history.
https://github.com/etcd-io/etcd/blob/49767381e803cc45c308fabea3c636ebb20f1d2f/tests/robustness/main_test.go#L87-L111

The `TestReport` assumes access to cluster created by e2e tests to get access to etcd data directories, we might want to disable it for now and follow it after https://github.com/etcd-io/etcd/issues/19883

### Why is this needed?

Be able to re-evaluate reports locally

/cc @henrybear327 @nwnt

nwnt:
Fix #19899

### DIFF
--- a/tests/antithesis/test-template/robustness/main.go
+++ b/tests/antithesis/test-template/robustness/main.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"maps"
 	"os"
+	"slices"
 	"sync"
 	"time"
 
@@ -86,34 +88,32 @@ func main() {
 		fmt.Sprintf(etcdDataPath, 2),
 	}
 
-	if err := os.RemoveAll(reportPath); err != nil {
-		panic(err)
-	}
-	if err := os.MkdirAll(reportPath, 0o700); err != nil {
-		panic(err)
-	}
-
 	ctx := context.Background()
 	baseTime := time.Now()
 	duration := time.Duration(robustnessrand.RandRange(5, 15) * int64(time.Second))
-	testRobustness(ctx, hosts, persistedRequestdirs, reportPath, baseTime, duration)
-}
 
-func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, reportPath string, baseTime time.Time, duration time.Duration) {
 	lg, err := zap.NewProduction()
 	if err != nil {
 		panic(err)
 	}
-	serverDataPaths := make(map[string]string)
+	r := report.TestReport{Logger: lg, ServersDataPath: map[string]string{}}
 	for i := range len(persistedRequestdirs) {
-		serverDataPaths[fmt.Sprintf("etcd%d", i)] = persistedRequestdirs[i]
+		r.ServersDataPath[fmt.Sprintf("etcd%d", i)] = persistedRequestdirs[i]
 	}
+	defer func() {
+		if err := r.Report(reportPath); err != nil {
+			lg.Error("Failed to save validation report", zap.Error(err))
+		}
+	}()
+	testRobustness(ctx, lg, &r, hosts, baseTime, duration)
+}
+
+func testRobustness(ctx context.Context, lg *zap.Logger, r *report.TestReport, hosts []string, baseTime time.Time, duration time.Duration) {
 	lg.Info("Start traffic generation", zap.Duration("duration", duration))
-	r := report.TestReport{Logger: lg}
 	r.Client = runTraffic(ctx, lg, hosts, baseTime, duration)
 	lg.Info("Completed traffic generation")
 
-	etcdDataDirs, err := report.PersistedRequestsDirs(lg, persistedRequestdirs)
+	etcdDataDirs, err := report.PersistedRequestsDirs(lg, slices.Collect(maps.Values(r.ServersDataPath)))
 	if err != nil {
 		lg.Error("Failed to create PersistedRequestdirs to access etcd WAL files")
 		panic(err)
@@ -131,9 +131,6 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 		assert.Always(result.Linearization.Linearizable == porcupine.Ok, "Linearization validation passes", nil)
 	}
 	r.Visualize = result.Linearization.Visualize
-	if err := r.Report(reportPath); err != nil {
-		lg.Error("Failed to save validation report", zap.Error(err))
-	}
 	assert.Always(result.WatchError == nil, "Watch validation passes", map[string]any{"error": result.WatchError})
 	assert.Always(result.SerializableError == nil, "Serializable validation passes", map[string]any{"error": result.WatchError})
 	lg.Info("Completed robustness validation")
--- a/tests/antithesis/test-template/robustness/main.go
+++ b/tests/antithesis/test-template/robustness/main.go
@@ -59,8 +59,8 @@ const (
 	defaultLocalEtcdDataPath = "/tmp/etcddata%d"
 	localEtcdDataPathEnv     = "ETCD_ROBUSTNESS_DATA_PATH"
 
-	defaultReportPath = "/var/report/history.html"
-	localReportPath   = "history.html"
+	defaultReportPath = "/var/report/"
+	localReportPath   = "report"
 )
 
 func main() {
@@ -86,6 +86,13 @@ func main() {
 		fmt.Sprintf(etcdDataPath, 2),
 	}
 
+	if err := os.RemoveAll(reportPath); err != nil {
+		panic(err)
+	}
+	if err := os.MkdirAll(reportPath, 0o700); err != nil {
+		panic(err)
+	}
+
 	ctx := context.Background()
 	baseTime := time.Now()
 	duration := time.Duration(robustnessrand.RandRange(5, 15) * int64(time.Second))
@@ -97,8 +104,13 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 	if err != nil {
 		panic(err)
 	}
+	serverDataPaths := make(map[string]string)
+	for i := range len(persistedRequestdirs) {
+		serverDataPaths[fmt.Sprintf("etcd%d", i)] = persistedRequestdirs[i]
+	}
 	lg.Info("Start traffic generation", zap.Duration("duration", duration))
-	reports := runTraffic(ctx, lg, hosts, baseTime, duration)
+	r := report.TestReport{Logger: lg}
+	r.Client = runTraffic(ctx, lg, hosts, baseTime, duration)
 	lg.Info("Completed traffic generation")
 
 	etcdDataDirs, err := report.PersistedRequestsDirs(lg, persistedRequestdirs)
@@ -107,7 +119,7 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 		panic(err)
 	}
 	validateConfig := validate.Config{ExpectRevisionUnique: traffic.EtcdAntithesis.ExpectUniqueRevision()}
-	result, err := validate.ValidateAndReturnVisualize(lg, validateConfig, reports, etcdDataDirs, 5*time.Minute)
+	result, err := validate.ValidateAndReturnVisualize(lg, validateConfig, r.Client, etcdDataDirs, 5*time.Minute)
 	if err != nil {
 		lg.Info("Validation error", zap.Error(err))
 		assert.Unreachable("Validation error", map[string]any{"error": err})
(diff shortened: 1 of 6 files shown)


tests/antithesis/Makefile
```diff
@@ -16,7 +16,7 @@ antithesis-run-container-tests:
 
 .PHONY: antithesis-run-local-tests
 antithesis-run-local-tests:
-	go run ./test-template/robustness/main.go --local
```
serathius, line 19: Need to confirm why we need a race at all. @marcus-hogson-antithesis

tests/antithesis/test-template/robustness/main_test.go
```diff
@@ -88,16 +88,23 @@ func main() {
...
+func testRobustness(ctx context.Context, t *testing.T, hosts, persistedRequestdirs []string, reportPath string, baseTime time.Time, duration time.Duration) {
 	lg, err := zap.NewProduction()
 	if err != nil {
 		panic(err)
 	}
+
+	r := report.TestReport{Logger: lg}
+	panicked := true
+	defer func() {
+		r.Report(t, panicked)
```
serathius, line 103: Instead to changing the binary to be a test, can we just remove `t` in r.Report and instead just return error?

nwnt: Not totally sure I got it right.  TestReport.Report needs a *testing.T to be called.  I have removed that call for now as we have not set it up correctly to persist with the (optional?) environment variable.  Let me know if this is off..

nwnt: pushed the new commit in.  making it to nil for now.

tests/antithesis/Makefile
marcus-hodgson-antithesis: @serathius This is building the go executable with race detection. This allows us to find DATA RACES. Not sure how necessary it would be for this test script, but definitely more important imo to build the etcd server containers with the data race flag.

CHANGELOG/CHANGELOG-3.6.md
```diff
@@ -97,6 +97,19 @@ See [code changes](https://github.com/etcd-io/etcd/compare/v3.5.0...v3.6.0).
 - `etcdctl` will sleep(2s) in case of range delete without `--range` flag. See [pull/13747](https://github.com/etcd-io/etcd/pull/13747)
 - Applications which depend on etcd v3.6 packages must be built with go version >= v1.18.
 
+#### Flags Removed
```
serathius, line 100: Unrelated, think you made mistake during rebase.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -62,10 +62,9 @@ const (
 	localReportPath   = "history.html"
 )
 
-func main() {
-	local := flag.Bool("local", false, "run tests locally and connect to etcd instances via localhost")
-	flag.Parse()
+var local = flag.Bool("local", false, "run tests locally and connect to etcd instances via localhost")
```
serathius, line 65: This change is not related.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -96,8 +95,15 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 	if err != nil {
 		panic(err)
 	}
+
+	r := report.TestReport{Logger: lg}
+	panicked := true
+	defer func() {
+		r.Report(nil, panicked)
```
serathius, line 102: You cannot pass nil here. You need to modify the Report function

nwnt:
@serathius thanks for the patience with me going the wrong way here.  I think this latest commit should be closer to what you actually needed.

Removing `*testing.T` involved quite a bit of refactoring here.  Let me know if you want me to split that into a different PR instead.

tests/robustness/report/report.go
```diff
@@ -35,51 +30,33 @@ type TestReport struct {
 	Visualize func(lg *zap.Logger, path string) error
 }
 
-func testResultsDirectory(t *testing.T) string {
-	resultsDirectory, ok := os.LookupEnv("RESULTS_DIR")
-	if !ok {
-		resultsDirectory = "/tmp/"
+func (r *TestReport) Report(failed, force, persistResults bool, path string) error {
```
serathius, line 33: Those three variables work identically, making them really function as single one. And the if statement line below just works as a turn off switch. Based on that I think we can just remove all of them and move the condition outside function. Like

```
shouldReport := failed || force || persistResults
if shouldReport {
  t.Report(path)
}
```

tests/robustness/report/client.go
```diff
@@ -169,36 +174,36 @@ func loadKeyValueOperations(path string) (operations []porcupine.Operation, err
...
 }
 
-func persistWatchOperations(t *testing.T, lg *zap.Logger, path string, responses []model.WatchOperation) {
+func persistWatchOperations(lg *zap.Logger, path string, responses []model.WatchOperation) error {
 	lg.Info("Saving watch operations", zap.String("path", path))
 	file, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o755)
 	if err != nil {
-		t.Errorf("Failed to save watch operations: %v", err)
-		return
+		return fmt.Errorf("Failed to save watch operations: %v", err)
```
serathius, line 181: This makes sense, however we need to lowercase the error message as the static validation will requires it.

tests/robustness/main_test.go
```diff
@@ -83,15 +87,40 @@ func TestRobustnessRegression(t *testing.T) {
 	}
 }
 
+func testResultsDirectory(t *testing.T) string {
```
serathius, line 90: Move this function to the bottom of file. It's much less important function, so it should be lower. Code should be read from top to the bottom. We shouldn't put low importantance function on the top.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -98,7 +97,10 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 		panic(err)
 	}
 	lg.Info("Start traffic generation", zap.Duration("duration", duration))
-	reports := runTraffic(ctx, lg, hosts, baseTime, duration)
+	r := report.TestReport{Logger: lg}
+
+	lg.Info("Start traffic generation")
```
serathius, line 102: Duplicate with 3 lines above. Looks like a bad rebase.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -118,9 +120,9 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 	} else {
 		assert.Always(result.Linearization.Linearizable == porcupine.Ok, "Linearization validation passes", nil)
 	}
-	err = result.Linearization.Visualize(lg, reportPath)
-	if err != nil {
-		lg.Error("Failed to save visualization", zap.Error(err))
+	r.Visualize = result.Linearization.Visualize
+	if err := r.Report(true, false, true, reportPath); err != nil {
```
serathius, line 124:
```suggestion
	if err := r.Report(reportPath); err != nil {
```

serathius, line 124: Could not write a comment on line 62, oh GitHub.

Think the report path no longer makes sense
```
	defaultReportPath = "/var/report/history.html"
	localReportPath   = "history.html"
```
We should modify it to make sense both outside and inside.

```
	defaultReportPath = "/var/report/"
	localReportPath   = "report"
```

For container this will dump report in mounted directory, and for local run it will create a `report` directory.
The only one problem I see is conflicts when report is written on directory that already contains it. This mangles the data. Let's leave the cleanup up to the `make antithesis-clean` for now.

tests/robustness/main_test.go
```diff
@@ -83,15 +87,40 @@ func TestRobustnessRegression(t *testing.T) {
...
 	// Refer to: https://github.com/golang/go/issues/49929
 	panicked := true
 	defer func() {
-		r.Report(t, panicked)
+		path := testResultsDirectory(t)
+		failed := t.Failed()
+		_, persistResults := os.LookupEnv("PERSIST_RESULTS")
+		if err := r.Report(failed, panicked, persistResults, path); err != nil {
+			t.Error(err)
+		}
```
serathius, line 121:
```suggestion
		if failed || panicked || persistResults {
			err := r.Report(failed, panicked, persistResults, path)
			if err != nil {
				t.Error(err)
			}
		}

```

tests/robustness/main_test.go
```diff
@@ -83,15 +87,40 @@ func TestRobustnessRegression(t *testing.T) {
...
 	}()
+
```
serathius, line 123: Not needed.

tests/robustness/report/report.go
```diff
@@ -35,51 +30,33 @@ type TestReport struct {
...
-}
-
-func (r *TestReport) Report(t *testing.T, force bool) {
-	_, persistResultsEnvSet := os.LookupEnv("PERSIST_RESULTS")
-	if !t.Failed() && !force && !persistResultsEnvSet {
-		return
-	}
-	path := testResultsDirectory(t)
 	r.Logger.Info("Saving robustness test report", zap.String("path", path))
 	for _, member := range r.Cluster.Procs {
```
serathius, line 38: Think this will panic as r.Cluster is empty in antithesis tests. To make data directories work with antithesis we need to change `TestReport` to use list of directories instead of cluster object.

For example:
```
type TestReport struct {
	Logger    *zap.Logger
	// map from server name to data dir path
	ServersDataPath map[string]string
	Client    []ClientReport
	Visualize func(lg *zap.Logger, path string) error
}

// Used by robustness
func serverDataPaths(c *e2e.EtcdProcessCluster) map[string]string {
	serverDataPaths := map[string]string{}
	for _, member := range c.Procs {
		serverDataPaths[member.Config().Name] = memberDataDir(member)
	}
}

// Used by antithesis
var serverDataPaths = map[string]string{
	"etcd0": "/var/etcddata0",
	"etcd1": "/var/etcddata1",
	"etcd2": "/var/etcddata2",
}

func (r *TestReport) Report(path string) error {
	for server, dataPath := range r.ServersDataPath {
		serverReportPath := filepath.Join(path, fmt.Sprintf("server-%s", server))
		err := os.Rename(dataPath, serverReportPath)
		...
	}
}
```

serathius:
> Removing *testing.T involved quite a bit of refactoring here. Let me know if you want me to split that into a different PR instead.

Don't think there is a need for a separate PR, at least for now. The switch from `testing.T` to returning error is needed just for this work and is pretty trivial change.

tests/antithesis/Makefile
serathius: Make sense.

tests/antithesis/test-template/robustness/main.go
nwnt: Yep, it was from the rebase :( sorry about that.

tests/antithesis/test-template/robustness/main.go
nwnt: I have added a step to `init` to clean up before `chown` the folder the user.  This should help with the clean up.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -86,6 +86,13 @@ func main() {
 		fmt.Sprintf(etcdDataPath, 2),
 	}
 
+	if err := os.RemoveAll(reportPath); err != nil {
```
serathius, line 89: Do we need delete in both here and init container?

tests/antithesis/test-template/robustness/main.go
```diff
@@ -97,8 +104,13 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
 	if err != nil {
 		panic(err)
 	}
+	serverDataPaths := make(map[string]string)
+	for i := range len(persistedRequestdirs) {
+		serverDataPaths[fmt.Sprintf("etcd%d", i)] = persistedRequestdirs[i]
+	}
```
serathius, line 110: Please move it to separate function. We should try to simplify the robustness as possible.

serathius (approved):

PR #19919, opened 2025-05-11
Use btree to avoid expensive copy and sorting response

serathius:
```
                                                      │  new3.bench  │            new4-4.bench             │
                                                      │    sec/op    │   sec/op     vs base                │
ValidateLinearizableOperations/Successes-14              3.639m ± 3%   3.508m ± 3%   -3.60% (p=0.003 n=10)
ValidateLinearizableOperations/AllFailures-14            1.538m ± 1%   1.459m ± 1%   -5.18% (p=0.000 n=10)
ValidateLinearizableOperations/PutFailuresWithRead-14   1052.6µ ± 1%   845.4µ ± 2%  -19.68% (p=0.000 n=10)
geomean                                                  1.806m        1.629m        -9.79%
```

### DIFF
--- a/tests/robustness/model/deterministic.go
+++ b/tests/robustness/model/deterministic.go
@@ -21,9 +21,9 @@ import (
 	"hash/fnv"
 	"maps"
 	"reflect"
-	"sort"
 
 	"github.com/anishathalye/porcupine"
+	"github.com/google/btree"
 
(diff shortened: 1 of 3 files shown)


serathius:
Thinking we might not even need a dynamic data structure. Technically we should know all keys used in linearization based on history. We could just allocate slice and assign keys to index.

PR #19931, opened 2025-05-12
Move linearization timeout and other validations to separate assertions

serathius:
/cc @marcus-hodgson-antithesis

### DIFF
--- a/tests/antithesis/test-template/robustness/main.go
+++ b/tests/antithesis/test-template/robustness/main.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/anishathalye/porcupine"
 	"github.com/antithesishq/antithesis-sdk-go/assert"
 	"go.uber.org/zap"
 	"golang.org/x/time/rate"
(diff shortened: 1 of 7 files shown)


tests/antithesis/test-template/robustness/main.go
```diff
@@ -106,13 +107,24 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
...
 	validateConfig := validate.Config{ExpectRevisionUnique: traffic.EtcdAntithesis.ExpectUniqueRevision()}
-	result := validate.ValidateAndReturnVisualize(lg, validateConfig, reports, etcdDataDirs, 5*time.Minute)
+	result, err := validate.ValidateAndReturnVisualize(lg, validateConfig, reports, etcdDataDirs, 5*time.Minute)
+	if err != nil {
+		lg.Info("Validation error", zap.Error(err))
+		assert.Unreachable("Validation error", map[string]any{"error": err})
+		return
+	}
+	if result.Linearization.Linearizable == porcupine.Unknown {
+		assert.Unreachable("Linearization timeout", nil)
```
serathius, line 117: @marcus-hodgson-antithesis

tests/antithesis/test-template/robustness/main.go
```diff
@@ -106,13 +107,24 @@ func testRobustness(ctx context.Context, hosts, persistedRequestdirs []string, r
...
+	} else {
+		assert.Always(result.Linearization.Linearizable == porcupine.Ok, "Linearization validation passes", nil)
+	}
 	err = result.Linearization.Visualize(lg, reportPath)
 	if err != nil {
 		lg.Error("Failed to save visualization", zap.Error(err))
 	}
-	assert.Always(result.Error == nil, "Robustness validation passes", map[string]any{"error": result.Error})
-	lg.Info("Completed robustness validation", zap.Error(result.Error))
+	assert.Always(result.WatchError == nil, "Watch validation passes", map[string]any{"error": result.WatchError})
```
serathius, line 125: Moved other validations to separate asserts so we can track it more easily.

tests/antithesis/test-template/robustness/main.go
marcus-hodgson-antithesis: @serathius for the "validation error" unreachable assertion, what has happened in that conditional if err != nil? If we see a linearization timeout will that err != nil?

Similar question for the "Linearization timeout" unreachable assertion, will porcupine.Unknown always indicate a timeout?

The logic looks good if these conditionals flow correctly!

serathius: > @serathius for the "validation error" unreachable assertion, what has happened in that conditional if err != nil? If we see a linearization timeout will that err != nil?

This is error in validation setup. I moved it as separate error because if it happens no validation has taken place.

> Similar question for the "Linearization timeout" unreachable assertion, will porcupine.Unknown always indicate a timeout?

Yes, `Unknown` means only timeout. `Ok` means success, `Illegal` means linearization failure

PR #19932, opened 2025-05-13
remove stale probot, and replace it with stale action

hwdef:
The stale bot currently used by etcd is no longer maintained and needs to be migrated to the new stale action.

https://github.com/probot/stale

https://github.com/apps/stale

PS:
- The k8s bot is also an optional choice, but I think it is heavy. https://github.com/k8s-triage-robot 

PPS:
- In fact, I also doubt the necessity of the stale action. It doesn't truly solve the problem, it just closes the issue. Perhaps we could directly remove the stale bot? >.<

/cc @siyuanfoundation @ahrtr @ivanvc


serathius (approved):
For now migrating off deprecated bot to a supported workflow, makes sense to me

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
```
jmhbnz, line 16: Nit: Suggest we pin version by hash as we do for other github actions for security reasons.

hwdef: Done.

jmhbnz (approved):
LGTM - Thanks @hwdef

ahrtr (approved):
LGTM

Thanks for the cleanup!!

ivanvc:
@jmhbnz, does exempting issues with the "help wanted" label sound like a good idea?

jmhbnz:
> @jmhbnz, does exempting issues with the "help wanted" label sound like a good idea?

Potentially yes. I think there is a clear difference between an issue that has been triaged and determined beneficial and the project is seeking help with, versus an issue that has not been triaged to that level for whatever reason.

PR #19935, opened 2025-05-13
client: Better error handling in Watcher.Close()

Jille:
The previous implementation might've ignored a real error if the last error happened to be context.Cancelled.

While we're at it, might as well use the new errors.Join() to actually return all errors.


Jille:
Bump

PR #19937, opened 2025-05-14
[Antithesis] Remove Go version from Dockerfile

nwnt:
This ensures we use the latest, same version of Go as the etcd project

cc @ivanvc


serathius (approved):

serathius:
Ups, this PR broke antithesis integration as it doesn't use makefile, but uses dockerfile directly. 

cc @henrybear327 we need those tests

PR #19946, opened 2025-05-15
[Antithesis] Add support for watch requests

### LINKED ISSUE #19940: [Antithesis] Add support for watch requests

Adding a concurrent watch requests to the traffic. Like in robustness tests:
https://github.com/etcd-io/etcd/blob/0c43c4b8688664532800f3b15ebed18f4a3b3c3c/tests/robustness/main_test.go#L166-L169

### Why is this needed?

Cover larger part of etcd API

nwnt:
Fix #19940

### DIFF
--- a/tests/antithesis/test-template/robustness/main.go
+++ b/tests/antithesis/test-template/robustness/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/anishathalye/porcupine"
 	"github.com/antithesishq/antithesis-sdk-go/assert"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/time/rate"
 
 	"go.etcd.io/etcd/tests/v3/robustness/client"
@@ -110,7 +111,12 @@ func main() {
 
 func testRobustness(ctx context.Context, lg *zap.Logger, r *report.TestReport, hosts []string, baseTime time.Time, duration time.Duration) {
 	lg.Info("Start traffic generation", zap.Duration("duration", duration))
-	r.Client = runTraffic(ctx, lg, hosts, baseTime, duration)
+	var err error
+	r.Client, err = runTraffic(ctx, lg, hosts, baseTime, duration)
+	if err != nil {
+		lg.Error("Failed to generate traffic")
+		panic(err)
+	}
 	lg.Info("Completed traffic generation")
 
 	etcdDataDirs, err := report.PersistedRequestsDirs(lg, slices.Collect(maps.Values(r.ServersDataPath)))
@@ -136,21 +142,47 @@ func testRobustness(ctx context.Context, lg *zap.Logger, r *report.TestReport, h
 	lg.Info("Completed robustness validation")
 }
 
-func runTraffic(ctx context.Context, lg *zap.Logger, hosts []string, baseTime time.Time, duration time.Duration) []report.ClientReport {
-	limiter := rate.NewLimiter(rate.Limit(profile.MaximalQPS), profile.BurstableQPS)
+func runTraffic(ctx context.Context, lg *zap.Logger, hosts []string, baseTime time.Time, duration time.Duration) ([]report.ClientReport, error) {
 	ids := identity.NewIDProvider()
-	storage := identity.NewLeaseIDStorage()
-	concurrencyLimiter := traffic.NewConcurrencyLimiter(profile.MaxNonUniqueRequestConcurrency)
-
 	r, err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, ids, baseTime)
 	if err != nil {
 		lg.Fatal("Failed empty database at start check", zap.Error(err))
 	}
-
-	finish := closeAfter(ctx, duration)
 	reports := []report.ClientReport{r}
+	watchReport := []report.ClientReport{}
+	maxRevisionChan := make(chan int64, 1)
+	watchConfig := client.WatchConfig{
+		RequestProgress: true,
+	}
+	g := errgroup.Group{}
+	g.Go(func() error {
+		defer close(maxRevisionChan)
+		reports := slices.Concat(reports, simulateTraffic(ctx, hosts, ids, baseTime, duration))
+		maxRevision := report.OperationsMaxRevision(reports)
+		maxRevisionChan <- maxRevision
+		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
+		return nil
+	})
+	g.Go(func() error {
+		var watchErr error
+		watchReport, _, watchErr = client.CollectClusterWatchEvents(ctx, lg, hosts, maxRevisionChan, watchConfig, baseTime, ids)
+		return watchErr
+	})
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+
+	return slices.Concat(reports, watchReport), nil
+}
+
+func simulateTraffic(ctx context.Context, hosts []string, ids identity.Provider, baseTime time.Time, duration time.Duration) []report.ClientReport {
 	var mux sync.Mutex
 	var wg sync.WaitGroup
+	storage := identity.NewLeaseIDStorage()
+	limiter := rate.NewLimiter(rate.Limit(profile.MaximalQPS), profile.BurstableQPS)
+	concurrencyLimiter := traffic.NewConcurrencyLimiter(profile.MaxNonUniqueRequestConcurrency)
+	finish := closeAfter(ctx, duration)
+	reports := []report.ClientReport{}
 	for i := 0; i < profile.ClientCount; i++ {
 		c := connect([]string{hosts[i%len(hosts)]}, ids, baseTime)
 		defer c.Close()
--- a/tests/robustness/client/watch.go
+++ b/tests/robustness/client/watch.go
@@ -20,29 +20,32 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
 
-	"go.etcd.io/etcd/tests/v3/framework/e2e"
 	"go.etcd.io/etcd/tests/v3/robustness/identity"
 	"go.etcd.io/etcd/tests/v3/robustness/report"
 )
 
-func CollectClusterWatchEvents(ctx context.Context, t *testing.T, clus *e2e.EtcdProcessCluster, maxRevisionChan <-chan int64, cfg WatchConfig, baseTime time.Time, ids identity.Provider) []report.ClientReport {
+func CollectClusterWatchEvents(ctx context.Context, lg *zap.Logger, endpoints []string, maxRevisionChan <-chan int64, cfg WatchConfig, baseTime time.Time, ids identity.Provider) ([]report.ClientReport, bool, error) {
 	mux := sync.Mutex{}
 	var wg sync.WaitGroup
-	reports := make([]report.ClientReport, len(clus.Procs))
-	memberMaxRevisionChans := make([]chan int64, len(clus.Procs))
-	for i, member := range clus.Procs {
-		c, err := NewRecordingClient(member.EndpointsGRPC(), ids, baseTime)
-		require.NoError(t, err)
+	neverFailed := true
+	reports := make([]report.ClientReport, len(endpoints))
+	memberMaxRevisionChans := make([]chan int64, len(endpoints))
+	for i, endpoint := range endpoints {
+		c, err := NewRecordingClient([]string{endpoint}, ids, baseTime)
+		if err != nil {
+			return nil, false, err
+		}
 		memberMaxRevisionChan := make(chan int64, 1)
 		memberMaxRevisionChans[i] = memberMaxRevisionChan
 		wg.Add(1)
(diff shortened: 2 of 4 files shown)


nwnt:
@serathius I think I still missed something, but I want to get your early feedback if I got the correct idea about doing or not.

tests/robustness/main_test.go
```diff
@@ -219,3 +214,13 @@ func testResultsDirectory(t *testing.T) string {
 	require.NoError(t, err)
 	return path
 }
+
+func processEndpoints(clus *e2e.EtcdProcessCluster) [][]string {
+	endpoints := make([][]string, len(clus.Procs))
+	for _, proc := range clus.Procs {
+		for i, grpcEndpoint := range proc.EndpointsGRPC() {
+			endpoints[i] = append(endpoints[i], grpcEndpoint)
```
serathius, line 222: `EndpointsGRPC()` returns all endpoints for process, however in this case one endpoint per process should be enough.

tests/robustness/client/watch.go
```diff
@@ -76,12 +83,14 @@ resetWatch:
 			select {
 			case <-ctx.Done():
 				if maxRevision == 0 {
-					t.Errorf("Client didn't collect all events, max revision not set")
+					lg.Error("Client didn't collect all events, max revision not set")
+					failed = true
 				}
 				if lastRevision < maxRevision {
-					t.Errorf("Client didn't collect all events, revision got %d, expected: %d", lastRevision, maxRevision)
+					lg.Sugar().Errorf("Client didn't collect all events, revision got %d, expected: %d", lastRevision, maxRevision)
```
serathius, line 90:
```suggestion
					lg.Error("Client didn't collect all events", zap.Int64("revision-got", lastRevision), zap.Int64("revision-expected", maxRevision))
```

tests/robustness/client/watch.go
```diff
@@ -56,17 +62,18 @@ func CollectClusterWatchEvents(ctx context.Context, t *testing.T, clus *e2e.Etcd
...
+	return reports, everFailed, nil
 }
 
 type WatchConfig struct {
 	RequestProgress bool
 }
 
 // watchUntilRevision watches all changes until context is cancelled, it has observed revision provided via maxRevisionChan or maxRevisionChan was closed.
-func watchUntilRevision(ctx context.Context, t *testing.T, c *RecordingClient, maxRevisionChan <-chan int64, cfg WatchConfig) {
+func watchUntilRevision(ctx context.Context, lg *zap.Logger, c *RecordingClient, maxRevisionChan <-chan int64, cfg WatchConfig) bool {
```
serathius, line 73: Code is good and correct. Just a nit: Returning bool for failure doesn't seem to be very Go like. 

Couple of possible improvements, implementing any of them should be ok:
* Avoid negative logic, it's starts getting confusing when you start negating or doing logic on it. Prefer returning `success` over `failed`.
* Return just error, you can both log and return error.
```
err = errors.New("client didn't collect all events")
lg.Error("Failed watch until revision", zap.Error(err))
```

tests/robustness/client/watch.go
```diff
@@ -20,28 +20,34 @@ import (
...
 		wg.Add(1)
 		go func(i int, c *RecordingClient) {
 			defer wg.Done()
 			defer c.Close()
-			watchUntilRevision(ctx, t, c, memberMaxRevisionChan, cfg)
+
+			watchFailed := watchUntilRevision(ctx, lg, c, memberMaxRevisionChan, cfg)
+			if watchFailed && !everFailed {
+				everFailed = true
+			}
```
serathius, line 50: This code is executed in separate goroutine, it needs to be synchronized via mutex, same like collecting reports.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -149,29 +149,51 @@ func runTraffic(ctx context.Context, lg *zap.Logger, hosts []string, baseTime ti
...
+					ids,
+					storage,
+					concurrencyLimiter,
+					finish,
+				)
+				mux.Lock()
+				reports = append(reports, c.Report())
+				mux.Unlock()
+			}(c)
+		}
```
serathius, line 176: Move this code to subfunction, same like in robustness tests.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -149,29 +149,51 @@ func runTraffic(ctx context.Context, lg *zap.Logger, hosts []string, baseTime ti
...
+		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
+	}()
+	go func() {
+		endpoints := make([][]string, len(hosts))
+		for _, host := range hosts {
+			endpoints = append(endpoints, []string{host})
+		}
+		watchConfig := client.WatchConfig{
+			RequestProgress: true,
+		}
```
serathius, line 189: Move this code outside goroutine. Goroutines should have minimal code needed for coordination between them.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -149,29 +149,51 @@ func runTraffic(ctx context.Context, lg *zap.Logger, hosts []string, baseTime ti
...
+		var watchErr error
+		watchReport, _, watchErr = client.CollectClusterWatchEvents(ctx, lg, endpoints, maxRevisionChan, watchConfig, time.Now(), ids)
+		if watchErr != nil {
+			panic(watchErr)
```
serathius, line 193: Use `errgroup.Group{}` to pass the errors to caller. Same like robustness tests.

tests/robustness/client/watch.go
nwnt: Thank you. I went with the first option as the existing robustness test still uses it to indicate whether the test case fails.

nwnt:
Resolved all the comments.  Ready for another review.

tests/robustness/client/watch.go
```diff
@@ -20,29 +20,34 @@ import (
...
-	reports := make([]report.ClientReport, len(clus.Procs))
-	memberMaxRevisionChans := make([]chan int64, len(clus.Procs))
-	for i, member := range clus.Procs {
-		c, err := NewRecordingClient(member.EndpointsGRPC(), ids, baseTime)
-		require.NoError(t, err)
+	neverFailed := true
+	reports := make([]report.ClientReport, len(etcdEndpoints))
+	memberMaxRevisionChans := make([]chan int64, len(etcdEndpoints))
+	for i, endpoint := range etcdEndpoints {
+		c, err := NewRecordingClient(endpoint, ids, baseTime)
```
serathius, line 36: Like mentioned before we don't need to connect to all endpoints from a member. One endpoint per client is enough.

tests/antithesis/test-template/robustness/main.go
```diff
@@ -137,20 +138,49 @@ func testRobustness(ctx context.Context, lg *zap.Logger, r *report.TestReport, h
...
 	}
-
-	finish := closeAfter(ctx, duration)
 	reports := []report.ClientReport{r}
+	watchReport := []report.ClientReport{}
+	maxRevisionChan := make(chan int64, 1)
+	endpoints := make([][]string, len(hosts))
+	for _, host := range hosts {
+		endpoints = append(endpoints, []string{host})
+	}
```
serathius, line 152: Let's simplify this function. This loop should not be needed if we only passed hosts to `CollectClusterWatchEvents`. This function doesn't need to accept `etcdEndpoints [][]string`, just `hosts []string` would be enough

tests/antithesis/test-template/robustness/main.go
```diff
@@ -137,20 +138,49 @@ func testRobustness(ctx context.Context, lg *zap.Logger, r *report.TestReport, h
...
+		defer close(maxRevisionChan)
+		reports := slices.Concat(reports, simulateTraffic(ctx, hosts, baseTime, duration))
+		maxRevision := report.OperationsMaxRevision(reports)
+		maxRevisionChan <- maxRevision
+		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
+		return nil
+	})
+	g.Go(func() error {
+		var watchErr error
+		watchReport, _, watchErr = client.CollectClusterWatchEvents(ctx, lg, endpoints, maxRevisionChan, watchConfig, time.Now(), ids)
```
serathius, line 167:
```suggestion
		watchReport, _, watchErr = client.CollectClusterWatchEvents(ctx, lg, endpoints, maxRevisionChan, watchConfig, baseTime, ids)
```

serathius: There needs to be one shared time.

tests/robustness/client/watch.go
```diff
@@ -20,29 +20,34 @@ import (
...
+		if err != nil {
+			return nil, true, err
```
serathius, line 38:
```suggestion
			return nil, false, err
```

tests/robustness/client/watch.go
```diff
@@ -20,29 +20,34 @@ import (
...
 		wg.Add(1)
 		go func(i int, c *RecordingClient) {
 			defer wg.Done()
 			defer c.Close()
-			watchUntilRevision(ctx, t, c, memberMaxRevisionChan, cfg)
+			watchSuccessful := watchUntilRevision(ctx, lg, c, memberMaxRevisionChan, cfg)
 			mux.Lock()
+			if !watchSuccessful && neverFailed {
+				neverFailed = false
+			}
```
serathius, line 50:
```suggestion
			neverFailed &=watchSuccessful
```

tests/antithesis/test-template/robustness/main.go
```diff
@@ -137,20 +138,49 @@ func testRobustness(ctx context.Context, lg *zap.Logger, r *report.TestReport, h
...
+		return watchErr
+	})
+	g.Wait()
```
serathius, line 170:
```suggestion
	err := g.Wait()
	if err != nil {
		return nil, err
	}
```

tests/antithesis/test-template/robustness/main.go
serathius, line 167: What's the purpose of returning bool if you are ignoring it?

tests/robustness/main_test.go
```diff
@@ -158,13 +157,22 @@ func runScenario(ctx context.Context, t *testing.T, s scenarios.TestScenario, lg
...
 		return nil
 	})
 	g.Go(func() error {
-		watchReport = client.CollectClusterWatchEvents(ctx, t, clus, maxRevisionChan, s.Watch, baseTime, ids)
+		var failed bool
+		var err error
+		endpoints := processEndpoints(clus)
+		watchReport, failed, err = client.CollectClusterWatchEvents(ctx, lg, endpoints, maxRevisionChan, s.Watch, baseTime, ids)
+		if err != nil {
+			return err
```
serathius, line 171:
```suggestion
			t.Error(err)
```

tests/robustness/client/watch.go
nwnt: I love this trick.  It goes very nicely with your prior suggestion about returning true.

nwnt: Hmm I can only make it `neverFailed = neverFailed && watchSuccessful` though.  `&=` seems only for bitwise operations.

tests/antithesis/test-template/robustness/main.go
nwnt: For the existing compatibility with the existing robustness test.  In main, it uses [`t.Errorf`](https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/tests/robustness/client/watch.go#L79) to indicate that the test case has failed.  Changing it to zap log's Error doesn't make the test case fail, that's why I'm returning the boolean so that the existing robustness test can fail it.

Alternatively, I can pass the fail function in or just ignore it altogether.  Let me know which one is more preferred.

tests/robustness/main_test.go
nwnt: Actually, I'll replace this with `require.NoError`.  This error is from [here](https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/tests/robustness/client/watch.go#L37), which stops the test execution when an error occurs.

nwnt: This piece is related to the above question about the unused bool.  It's to differentiate between this fatal error and cases the tests should be marked as failed, but still continue the execution.

nwnt:
Thanks again for the patience while I'm still wrapping my head around endpoints and clients.  Fixed the issues and left some comments about the questions you had.  Ready for another round of review.

tests/robustness/main_test.go
```diff
@@ -219,3 +212,11 @@ func testResultsDirectory(t *testing.T) string {
 	require.NoError(t, err)
 	return path
 }
+
+func processEndpoints(clus *e2e.EtcdProcessCluster) []string {
```
serathius, line 216: `clusterEndpoints` would be simpler name, but let's leave that for a followup.

serathius (approved):

serathius:
Heh, didn't noticed that the tests were not run :( 
https://github.com/etcd-io/etcd/pull/19968

PR #19951, opened 2025-05-15
[release-3.6] Update release script to exactly match the target tag

ahrtr:
When we release etcd v3.6.0, we ran into an issue. The command `git tag --list | grep -c "${RELEASE_VERSION}")" -gt 0` returned 64, so it thought the tag v3.6.0 already existed, but actually not. 

So updating the script to exactly match the target tag.

Normally we fix in main, and backport to stable releases, given the tight schedule, so we fix it release-3.6 for now, and backport and forward port to other branches.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -184,7 +184,7 @@ main() {
     fi
 
     # Tag release.
-    if [ "$(git tag --list | grep -c "${RELEASE_VERSION}")" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


jmhbnz (approved):
LGTM - We validated this during the release call. Thanks @ahrtr

ahrtr:
To be clearer, the command below returned 64, so it's greater than 0

```
git tag --list | grep -c "${RELEASE_VERSION}")"
```

PR #19956, opened 2025-05-16
Build dedicated etcd image for antithesis

### LINKED ISSUE #19943: [Antithesis] Build dedicated etcd image for antithesis testing

For now we are just using v3.5.21 etcd official image in antithesis.

Instead of using release image would want to start building etcd from release-3.5 branch and push a image with it.

### Why is this needed?

Test latest code on a branch

nwnt:
Fix #19943

### DIFF
--- a/.github/workflows/antithesis-test.yml
+++ b/.github/workflows/antithesis-test.yml
@@ -82,6 +82,16 @@ jobs:
             ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-client:latest,
             ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-client:${{ github.sha }}
 
+      - name: Build and push etcd image
+        run: |
+          make antithesis-build-etcd-image
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-server:latest"
+          docker tag etcd-server:latest $TAG
+          docker push $TAG
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-server:${{ github.sha }}"
+          docker tag etcd-server:latest $TAG
+          docker push $TAG
+
       - name: Run Antithesis Tests
         uses: antithesishq/antithesis-trigger-action@b7d0c9d1d9316bd4de73a44144c56636ea3a64ba # main commit on Mar 13, 2025
         with:
--- a/tests/antithesis/Makefile
+++ b/tests/antithesis/Makefile
@@ -1,10 +1,21 @@
 REPOSITORY_ROOT := $(shell git rev-parse --show-toplevel)
 USER_ID := $(shell id -u)
 GROUP_ID := $(shell id -g)
+ARCH ?= $(shell go env GOARCH)
 
 .PHONY: antithesis-build-client-docker-image
 antithesis-build-client-docker-image:
 	docker build --build-arg GO_VERSION=$(shell cat $(REPOSITORY_ROOT)/.go-version) -f $(REPOSITORY_ROOT)/tests/antithesis/test-template/Dockerfile $(REPOSITORY_ROOT) -t etcd-client:latest
+
+.PHONY: antithesis-build-etcd-image
+antithesis-build-etcd-image:
+	git worktree add /tmp/etcd3.5 origin/release-3.5
+	cd /tmp/etcd3.5 && make build
+	cd /tmp/etcd3.5 && cp ./Dockerfile-release.$(ARCH) ./bin/Dockerfile
+	cd /tmp/etcd3.5 && docker build \
+		--tag etcd-server:latest \
+		./bin
+	git worktree remove -f /tmp/etcd3.5
 
 .PHONY: antithesis-docker-compose-up
 antithesis-docker-compose-up:
--- a/tests/antithesis/README.md
+++ b/tests/antithesis/README.md
@@ -8,6 +8,7 @@ Run this command from the `antithesis/test-template` directory:
 
 ```bash
 make antithesis-build-client-docker-image
+make antithesis-build-etcd-image
 ```
 
 ### 2. (Optional) Check the Image Locally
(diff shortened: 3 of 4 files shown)


tests/antithesis/Makefile
```diff
@@ -1,11 +1,24 @@
...
 GROUP_ID := $(shell id -g)
+ARCH ?= $(shell go env GOARCH)
 
 .PHONY: antithesis-build-client-docker-image
 antithesis-build-client-docker-image:
 	docker build --build-arg GO_VERSION=$(shell cat $(REPOSITORY_ROOT)/.go-version) -f $(REPOSITORY_ROOT)/tests/antithesis/test-template/Dockerfile $(REPOSITORY_ROOT) -t etcd-client:latest
 
+.PHONY: antithesis-build-etcd-image
+antithesis-build-etcd-image:
+	git worktree add /tmp/etcd3.5 origin/release-3.5
```
serathius, line 12: I'm not familiar with using git worktree this way. Can you explain what it's purpose? cc @ivanvc

tests/antithesis/Makefile
```diff
@@ -1,11 +1,24 @@
...
+	cd /tmp/etcd3.5 && make build
+	cd /tmp/etcd3.5 && cp ./Dockerfile-release.$(ARCH) ./bin/Dockerfile-release.$(ARCH)
+	cd /tmp/etcd3.5 && docker build \
+		--network=host \
+		--tag etcd-server:latest \
+	  --file ./Dockerfile-release.$(ARCH) \
```
serathius, line 18:
```suggestion
		--file ./Dockerfile-release.$(ARCH) \
```

serathius, line 18:
```suggestion
	cd /tmp/etcd3.5 && cp ./Dockerfile-release.$(ARCH) ./bin/Dockerfile
	cd /tmp/etcd3.5 && docker build \
		--network=host \
		--tag etcd-server:latest \
```

tests/antithesis/README.md
```diff
@@ -8,6 +8,7 @@ Run this command from the `antithesis/test-template` directory:
 
 ```bash
 make antithesis-build-client-docker-image
```
serathius, line 10: We need to change how the image is build in antithesis workflow.

https://github.com/etcd-io/etcd/blob/f206a8a579eb121d17e314637cb3fd90b8e30cf1/.github/workflows/antithesis-test.yml#L75-L83

We could replace the docker/build-push-action with just invoking makefile, however we would also need to retag and push the images to `us-central1-docker.pkg.dev/molten-verve-216720/linuxfoundation-repository/etcd-server:latest`

tests/antithesis/Makefile
nwnt: Sure.  To build the image from `release-3.5`, we need to check out the whole branch and can't retrieve just the dockerfile file as the binaries it needs have to be built from the code on the branch.  `git checkout origin/release-3.5` to build the image would fail when you have uncommitted changes, or if something fails during the build then git would still be on origin/release-3.5's commit.  So I think it's safer to do a temporary worktree and build there instead.

I don't think this is the best way to do it though.  I'm open to any suggestions or better ways to do it.

tests/antithesis/README.md
nwnt: Thank you.  Sorry for some reasons it was still not registered in my brain that we have the workflow files too... Should be the last time I forget about it.

tests/antithesis/Makefile
serathius: I was not aware of this, if you look at https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/tests/robustness/Makefile#L101-L197 my approach was very naive.

.github/workflows/antithesis-test.yml
```diff
@@ -82,6 +82,13 @@ jobs:
             ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-client:latest,
             ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-client:${{ github.sha }}
 
+      - name: Build and push etcd image
+        run: |
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-server:latest"
+          make antithesis-tag-and-push-etcd-image
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-server:${{ github.sha }}"
+          make antithesis-tag-and-push-etcd-image
```
serathius, line 90:
```suggestion
          make antithesis-build-etcd-image
          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-server:latest"
          docker tag etcd-server:latest $TAG
          docker push $TAG
          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-server:${{ github.sha }}"
          docker tag etcd-server:latest $TAG
          docker push $TAG
```

tests/antithesis/Makefile
```diff
@@ -1,11 +1,27 @@
...
+	git worktree add /tmp/etcd3.5 origin/release-3.5
+	cd /tmp/etcd3.5 && make build
+	cd /tmp/etcd3.5 && cp ./Dockerfile-release.$(ARCH) ./bin/Dockerfile
+	cd /tmp/etcd3.5 && docker build \
+		--tag etcd-server:latest \
+		./bin
+	git worktree remove -f /tmp/etcd3.5
+
+.PHONY: antithesis-tag-etcd-image
+antithesis-tag-and-push-etcd-image: antithesis-build-etcd-image
```
serathius, line 21: Don't think we need a target for docker tag and push

tests/antithesis/Makefile
nwnt: got it.  do you prefer that way more?

serathius: Don't know what would be the difference.

nwnt: technically, using worktree might be a bit faster, since it doesn't do another clone but uses the locally fetched commits.

serathius: Make sense to me. Feel free to improve https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/tests/robustness/Makefile#L101-L197 too

serathius (approved):

PR #19959, opened 2025-05-16
Cleanup the deprecated experimental flags

ahrtr:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #19961, opened 2025-05-16
Bump version to 3.7

ahrtr:


ahrtr:
Several issues:
- We need to resolve all test failures. I will take care of this
- @ivanvc @jmhbnz @joshjms  we need to disable the release workflow first. After I finish this PR, we can change the `3.6.99 `to `3.7.99` in the release workflow test.

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/19961/pull-etcd-release-tests/1923398518836301824

```
stderr: [0;31mWrong etcd minor version in api/version/version.go. Expected 3.6 but got 3.7. Aborting.
[0;31mFAIL: (code:1):
  % '/home/prow/go/src/github.com/etcd-io/etcd/scripts/release.sh' '--no-upload' '--no-docker-push' '--no-gh-release' '--in-place' '3.6.99'
% '/home/prow/go/src/github.com/etcd-io/etcd/scripts/test_images.sh'
```

- We need to populate the storageVersion during bootstrap in greenfield installation case.  cc @fuweid @serathius 
 
  - Currently, the storageVersion is always read as V3.5, and then migrated to v3.7 later.
https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/server/storage/schema/schema.go#L111

  - We can populate the storageVersion with the binary version ([version.Version](https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/api/version/version.go#L29)) during bootstrap if it's a fresh storage,
https://github.com/etcd-io/etcd/blob/b605213c058c7ce4efda748ccb771ebee596b1af/server/etcdserver/bootstrap.go#L221-L228

We can change the above code to something like below,

```
	if haveWAL {
		if err = fileutil.IsDirWriteable(cfg.WALDir()); err != nil {
			return nil, fmt.Errorf("cannot write to WAL directory: %w", err)
		}
		bwal = bootstrapWALFromSnapshot(cfg, backend.snapshot)
	} else {
		// Initialize the schema
		populate StorageVersion with version.Version
		// anything else
	}

```


Note that https://github.com/etcd-io/etcd/pull/19959 is also blocked by this one, because the mix version tests failed.

server/storage/schema/migration_test.go
```diff
@@ -51,11 +51,9 @@ func TestNewPlan(t *testing.T) {
...
 		},
 		{
-			name:           "Upgrade v3.6 to v3.7 should fail as v3.7 is unknown",
-			current:        version.V3_6,
-			target:         version.V3_7,
-			expectError:    true,
-			expectErrorMsg: `version "3.7.0" is not supported`,
+			name:    "Upgrade v3.6 to v3.7 should work",
+			current: version.V3_6,
+			target:  version.V3_7,
```
serathius, line 56: Maybe we should add a new test for v3.8 to make sure this test is updated when a cut new release?

ahrtr: SG, updated & thx

serathius:
LGTM, stil we need to decide how to handle v3.8. We either add new sets of tests for v3.8, or we rewrite them.

ahrtr:
> LGTM, stil we need to decide how to handle v3.8. We either add new sets of tests for v3.8, or we rewrite them.

We are working on 3.7, so I don't think we need to worry about 3.8 for now.

For the test cases, I agree that we should try to make some of them more flexible so that we don't need to update them each time when we cut major or minor release. We can discuss & address it separately.

serathius:
> We are working on 3.7, so I don't think we need to worry about 3.8 for now.

That's why we should just copy the test for now and tackle the problem if needed. Validation for v3.8 release, might seem irrelevant for now, but also easy to forget later. I think we should prefer safe now, than sorry later.

server/storage/schema/schema_test.go
```diff
@@ -73,10 +73,8 @@ func TestValidate(t *testing.T) {
 			version: version.V3_6,
 		},
 		{
-			name:           `V3.7 schema is unknown and should return error`,
```
siyuanfoundation, line 76: I think it would be easier just to add a new test case for 3.8. Otherwise, we might lose track of which 3.8 validation tests are needed.

ahrtr: added & thx

siyuanfoundation (approved):

PR #19972, opened 2025-05-18
Add StaleGet traffic for antithesis

### LINKED ISSUE #19941: [Antithesis] Cover larger part of etcd API

Bring other requests types to EtcdAntithesis from other traffic types

https://github.com/etcd-io/etcd/blob/0c43c4b8688664532800f3b15ebed18f4a3b3c3c/tests/robustness/traffic/etcd.go#L34-L88

We would want to add each type one by one, after each step run the antithesis for 12h to see the results.

### Why is this needed?

Cover whole API
/cc @nwnt @henrybear327 @joshjms

nwnt:
Fix #19941

### DIFF
--- a/tests/robustness/traffic/etcd.go
+++ b/tests/robustness/traffic/etcd.go
@@ -81,8 +81,9 @@ var (
 		leaseTTL:     DefaultLeaseTTL,
 		// Please keep the sum of weights equal 100.
 		requests: []random.ChoiceWeight[etcdRequestType]{
-			{Choice: Get, Weight: 50},
-			{Choice: Put, Weight: 50},
+			{Choice: Get, Weight: 45},
+			{Choice: Put, Weight: 45},
+			{Choice: StaleGet, Weight: 10},
 		},
 	}
 )


tests/robustness/traffic/etcd.go
```diff
@@ -81,8 +81,9 @@ var (
 		leaseTTL:     DefaultLeaseTTL,
 		// Please keep the sum of weights equal 100.
 		requests: []random.ChoiceWeight[etcdRequestType]{
-			{Choice: Get, Weight: 50},
-			{Choice: Put, Weight: 50},
+			{Choice: Get, Weight: 45},
+			{Choice: Put, Weight: 45},
+			{Choice: Put, Weight: 10},
```
serathius, line 86: This doesn't look correct

nwnt: Sorry I was half dazed when I was doing this...  Corrected the new Put to StaleGet now.

serathius (approved):

PR #19995, opened 2025-05-21
Measure traffic during test

### LINKED ISSUE #19871: [Antithesis] Measure traffic during the test

Report traffic generated in antithesis test like in robustness test.

Code analyse traffic https://github.com/etcd-io/etcd/blob/aedae3a4fa83d1d1ca2bbae3db9599a5171b3768/tests/robustness/traffic/traffic.go#L136. Reporting just complete traffic should be enough.

We can make public:  `CalculateStats` function, `trafficStats` struct and all it's fields.

This should be added by the end of the traffic generation in antithesis tests. https://github.com/etcd-io/etcd/blob/aedae3a4fa83d1d1ca2bbae3db9599a5171b3768/tests/antithesis/test-template/robustness/main.go#L81

### Why is this needed?

Give us some idea of QPS generated in antithesis platform.

nwnt:
Fix #19871

### DIFF
--- a/tests/antithesis/test-template/robustness/traffic/main.go
+++ b/tests/antithesis/test-template/robustness/traffic/main.go
@@ -75,7 +75,6 @@ func main() {
 		lg.Error("Failed to generate traffic")
 		panic(err)
 	}
-	lg.Info("Completed traffic generation")
 }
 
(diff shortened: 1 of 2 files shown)


serathius (approved):
Good job

PR #19997, opened 2025-05-21
mvcc: switch to fail point in test

kjgorman:
This test is hard to provoke without inserting a sleep at the right place... fortunately this is what failpoints can do for us.

Note that to my understanding, the user still needs to run `gofail enable` etc in order to run this unit test with this delay inserted (i.e., I don't see that we are using failpoints already as part of the unit testing flow for anything in this package).

### DIFF
--- a/server/go.mod
+++ b/server/go.mod
@@ -31,6 +31,7 @@ require (
 	go.etcd.io/etcd/client/v2 v2.306.0-alpha.0
 	go.etcd.io/etcd/client/v3 v3.6.0-alpha.0
 	go.etcd.io/etcd/pkg/v3 v3.6.0-alpha.0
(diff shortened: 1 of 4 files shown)


server/go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	go.etcd.io/etcd/client/v2 v2.306.0-alpha.0
 	go.etcd.io/etcd/client/v3 v3.6.0-alpha.0
 	go.etcd.io/etcd/pkg/v3 v3.6.0-alpha.0
+	go.etcd.io/gofail v0.2.0
```
kjgorman, line 34: I don't have an intuition for how undesirable it might be to add this dependency to a core package like `server`

PR #20001, opened 2025-05-21
Converts between experimental and non-experimental flags in downgrade & upgrade robustness & e2e test

ahrtr:
followup to https://github.com/etcd-io/etcd/pull/19959#issuecomment-2897914417

### DIFF
--- a/tests/framework/e2e/config.go
+++ b/tests/framework/e2e/config.go
@@ -14,6 +14,15 @@
 
 package e2e
 
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-semver/semver"
+
+	"go.etcd.io/etcd/api/v3/version"
+)
+
 type ClusterVersion string
 
 const (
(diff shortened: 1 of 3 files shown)


tests/framework/e2e/config.go
```diff
@@ -33,3 +41,33 @@ func (cv ClusterVersion) String() string {
...
+		name == "enable-distributed-tracing" ||
+		name == "distributed-tracing-address" ||
+		name == "distributed-tracing-service-name" ||
+		name == "distributed-tracing-instance-id" ||
+		name == "distributed-tracing-sampling-rate" {
+		return fn()
+	}
+
+	return fmt.Sprintf("--%s=%s", name, value)
+}
```
serathius, line 73:
```suggestion
	if ver != nil || ver.GreatherOrEqual(version.V3_6) {
		return fmt.Sprintf("--%s=%s", name, value)
	}	
	switch name {
	case  "compact-hash-check-time", "corrupt-check-time", "corrupt-check-time", "compaction-batch-limit", "watch-progress-notify-interval",  "warning-apply-duration", "bootstrap-defrag-threshold-megabytes",  "memory-mlock", "snapshot-catchup-entries", "compaction-sleep-interval", "downgrade-check-time", "peer-skip-client-san-verification", "enable-distributed-tracing", "distributed-tracing-address", "distributed-tracing-service-name", "distributed-tracing-instance-id", "distributed-tracing-sampling-rate":
		return fmt.Sprintf("--experimental-%s=%s", name, value)
	default:
		return fmt.Sprintf("--%s=%s", name, value)
	}
```

tests/framework/e2e/downgrade.go
```diff
@@ -157,6 +157,13 @@ func DowngradeUpgradeMembersByID(t *testing.T, lg *zap.Logger, clus *EtcdProcess
 		if err := member.Stop(); err != nil {
 			return err
 		}
+
+		// When we downgrade or upgrade a member, we need to re-generate the flags, to convert some non-experimental
+		// flags to experimental flags, or vice verse.
+		if err := clus.UpdateProcOptions(memberID, t); err != nil {
```
serathius, line 163: This will not be called in robustness tests

ahrtr: My understanding is both robustness test and e2e test reuse the same `DowngradeUpgradeMembersByID` to perform downgrade & upgrade.

If it doesn't, then robustness test will fail.

But the robustness test on older version is periodically test, can't be verified immediately. So I raised https://github.com/kubernetes/test-infra/pull/34834

serathius: ack

ahrtr:
> FYI: I'm waiting for CI to pass before reviewing it again.

I will rebase this PR once https://github.com/etcd-io/etcd/pull/20016 gets merged.

Also we can see e2e workflow result in this PR before being rebased.

ahrtr:
ping @serathius  It should be working now.

serathius (approved):

ahrtr:
Since the https://github.com/kubernetes/test-infra/pull/34834 hasn't been merged, so the robustness test against previous versions can't be verified in this PR. If there is still any issues after merging this PR, we can fix in followup PR(s).

serathius:
> Since the https://github.com/kubernetes/test-infra/pull/34834 hasn't been merged, so the robustness test against previous versions can't be verified in this PR. If there is still any issues after merging this PR, we can fix in followup PR(s).

Have you run the test locally?

ahrtr:
> Have you run the test locally?

3.6 works.

3.5/3.4 have issues, but unrelated to flags.

test-robustness-release-3.5:
```
        logger.go:146: 2025-05-23T14:25:27.534+0100	ERROR	Failed to trigger failpoint	{"failpoint": "beforeSendWatchResponse=sleep", "error": "goFailpoint beforeSendWatchResponse=sleep setup failed, err:bad status code: 400, err: fail to set failpoint: failpoint: failpoint does not exist\n"}
        main_test.go:144: failed triggering failpoint, err: goFailpoint beforeSendWatchResponse=sleep setup failed, err:bad status code: 400, err: fail to set failpoint: failpoint: failpoint does not exist
```

So we should be good to merge this PR, and investigate the failure separately.

serathius:
> So we should be good to merge this PR, and investigate the failure separately.

SG

Issue #19984, opened 2025-05-19
Moving window of keys used by robustness


serathius:
### What would you like to be added?

Currently robustness tests just use a static range of keys `key0`-`key9`. During the test we execute different types of writing operations like PUT and DELETE. To make linearization faster we use the fact that PUTs are unique due to unique saved value, however this doesn't work for DELETE. To make DELETE unique at least partially we could make the key window slowly shift. For example, when during the traffic generation delete `key0`, we could shift the window up to be `key1` - `key10`. This way we could increase the chance that key will be deleted only once.

### Why is this needed?

Make deletes unique, removing a need to treat them differently than put.

joshjms:
I'd love to take this

/assign

PR #20002, opened 2025-05-21
Add key window to achieve unique keys for delete operations

joshjms:

Adds the `keyStore` into `etcdTraffic` which maintains a consistent pool of keys (size 10) such that used keys will be swapped out and replaced with a new key. 

```bash
PERSIST_RESULTS=true EXPECT_DEBUG=true GO_TEST_FLAGS='-v' make test-robustness
```

The porcupine results shows that the uniqueness of keys in early stages for EtcdTrafficDeleteLeases:

image

and in the later stages:

image

ref: https://github.com/etcd-io/etcd/issues/19984
/cc @serathius

### DIFF
--- a/tests/robustness/traffic/etcd.go
+++ b/tests/robustness/traffic/etcd.go
@@ -121,13 +121,13 @@ func (t etcdTraffic) Name() string {
 	return "Etcd"
 }
 
-func (t etcdTraffic) RunTrafficLoop(ctx context.Context, c *client.RecordingClient, limiter *rate.Limiter, ids identity.Provider, lm identity.LeaseIDStorage, nonUniqueWriteLimiter ConcurrencyLimiter, finish <-chan struct{}) {
+func (t etcdTraffic) RunTrafficLoop(ctx context.Context, c *client.RecordingClient, limiter *rate.Limiter, ids identity.Provider, lm identity.LeaseIDStorage, nonUniqueWriteLimiter ConcurrencyLimiter, keyStore *keyStore, finish <-chan struct{}) {
 	lastOperationSucceeded := true
 	var lastRev int64
 	var requestType etcdRequestType
 	client := etcdTrafficClient{
 		etcdTraffic:  t,
-		keyPrefix:    "key",
+		keyStore:     keyStore,
 		client:       c,
 		limiter:      limiter,
 		idProvider:   ids,
@@ -208,7 +208,7 @@ func filterOutNonUniqueEtcdWrites(choices []random.ChoiceWeight[etcdRequestType]
 
 type etcdTrafficClient struct {
 	etcdTraffic
-	keyPrefix    string
+	keyStore     *keyStore
 	client       *client.RecordingClient
 	limiter      *rate.Limiter
 	idProvider   identity.Provider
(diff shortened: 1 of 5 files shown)


tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+	key := randKey
+
+	return fmt.Sprintf("%s%d", k.keyPrefix, key)
+}
+
+func (k *KeyWindow) useWindow() bool {
+	randVal := rand.Intn(100)
+	if randVal < 10 {
+		return false
+	}
```
nwnt, line 59:
```suggestion
return rand.Intn(100) >= 10
```

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// This will use the default key0 to key9
+	randKey := rand.Intn(k.WindowSize)
+	key := randKey
+
+	return fmt.Sprintf("%s%d", k.keyPrefix, key)
+}
+
+func (k *KeyWindow) useWindow() bool {
+	randVal := rand.Intn(100)
```
nwnt, line 56: This function can be simplified to be something like
```suggestion
return !rand.Intn(100) < 10
```

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+	Counter    int
+	WindowSize int
+	Keys       []int
+	keyPrefix  string
+}
+
+func NewKeyWindow(size int, keyPrefix string) *KeyWindow {
+	k := &KeyWindow{
+		Counter:    0,
+		WindowSize: size,
```
nwnt, line 21: can we instead use len(Keys) when we need the size?

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"fmt"
+	"math/rand"
+	"sync"
+)
+
+type KeyWindow struct {
+	sync.Mutex
+
+	Counter    int
```
nwnt, line 12: if we don't use this anywhere, I think let's not make it public.

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+		randKey := rand.Intn(k.WindowSize)
+		key := k.Keys[randKey]
+		k.Keys[randKey] = k.Counter
+		k.Counter++
+
+		return fmt.Sprintf("%s%d", k.keyPrefix, key)
+	}
+
+	// This will use the default key0 to key9
+	randKey := rand.Intn(k.WindowSize)
```
nwnt, line 49: Since `randkey` doesn't seem to be used anywhere, we can combine line 49 and 50 to be:
```go
key := rand.Intn(k.WindowSize)
```

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+		k.Counter++
+	}
+
+	return k
+}
+
+func (k *KeyWindow) GetKey() string {
+	k.Lock()
+	defer k.Unlock()
+
```
nwnt, line 37: I think we can structure this function by making it like this:
```go
key := rand.Intn(k.WindowSize)
if k.useWindow() {
  randkey := key
  key = k.Keys[randKey]
  k.Keys[randKey] = k.Counter
  k.Counter++
}
return fmt.Sprintf("%s%d", k.keyPrefix, key)
```

nwnt: no hold on one sec on this one.

nwnt: nvm, I think I kinda got it right, confused myself about WindowSize and the random value in useWindow().

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
+package traffic
+
+import (
+	"fmt"
+	"math/rand"
+	"sync"
+)
+
+type KeyWindow struct {
```
serathius, line 9: Think we should pick a better name. The `window` in name comes from idea for algorithm where we have a moving window of key range. However, we should rather name it based on it's purpose. Like `keyStore`, `keyGenerator`, `keyProvider`

tests/robustness/traffic/key_window.go
serathius: All those fields should be private and we should fill them through constructor.

tests/robustness/traffic/key_window.go
serathius: +1

tests/robustness/traffic/key_window.go
```diff
@@ -0,0 +1,62 @@
...
+	sync.Mutex
+
+	Counter    int
+	WindowSize int
+	Keys       []int
```
serathius, line 14: Maybe Key field should include key names and not numbers used to generated keys.

joshjms:
Thanks for the reviews guys, I will address them as soon as possible. Thanks for bearing with me :smile:

tests/robustness/traffic/key_window.go
joshjms: sliding window but nothing is sliding :frowning_face: ok will change to keyStore

joshjms:
I addressed all the reviews you guys gave, please let me know if there's anything I missed. Please look at the updated screenshots for the new evidence.

tests/robustness/traffic/etcd.go
```diff
@@ -215,21 +215,21 @@ type etcdTrafficClient struct {
 	leaseStorage identity.LeaseIDStorage
 }
 
-func (c etcdTrafficClient) Request(ctx context.Context, request etcdRequestType, lastRev int64) (rev int64, err error) {
+func (c etcdTrafficClient) Request(ctx context.Context, request etcdRequestType, lastRev int64, keyStore *keyStore) (rev int64, err error) {
```
serathius, line 218: You can put the keyStore on `etcdTrafficClient` struct.

tests/robustness/traffic/key_store.go
```diff
@@ -0,0 +1,105 @@
...
+	k.mu.Lock()
+	defer k.mu.Unlock()
+
+	useKeyIndex := rand.Intn(len(k.keys))
+	useKey := k.keys[useKeyIndex]
+
+	if rand.Intn(100) < 90 {
+		k.keys[useKeyIndex] = fmt.Sprintf("%s%d", k.keyPrefix, k.counter)
+		k.counter++
+	}
```
serathius, line 68:
```suggestion
	if rand.Intn(100) < 90 {
		k.keys[useKeyIndex] = fmt.Sprintf("%s%d", k.keyPrefix, k.counter)
		k.counter++
	}
```
Move this to a separate function`replaceKey`

tests/robustness/traffic/key_store.go
```diff
@@ -0,0 +1,105 @@
...
+		panic("GetKeysForMultiTxnOps: number of operations is more than the key pool size")
+	}
+
+	keys := make([]string, numOps)
+
+	permutedKeyIndexes := rand.Perm(len(k.keys))
+	for i, op := range ops {
+		keys[i] = k.keys[permutedKeyIndexes[i]]
+
+		if op == model.DeleteOperation && rand.Intn(100) < 90 {
```
serathius, line 89: This repeats rolling 90%

joshjms: I completely missed removing the roll inside, thanks for noticing

PR #17006, opened 2023-11-23
Remove v2 sync

serathius:

PR #7347, opened 2017-02-17
*: add 'staticcheck' to 'test'

gyuho:
Fix https://github.com/coreos/etcd/issues/7298.

We will still have some failures until we migrate to Go 1.8.
So this just mask those errors to not fail the whole test suite.

```
skipping error:
contrib/raftexample/raftexample_test.go:105:3: the goroutine calls T.Fatalf, which must be called in the same goroutine as the test (SA2002)
skipping error:
e2e/ctl_v3_watch_test.go:72:3: the goroutine calls T.Fatalf, which must be called in the same goroutine as the test (SA2002)
e2e/etcd_release_upgrade_test.go:156:3: the goroutine calls T.Fatalf, which must be called in the same goroutine as the test (SA2002)
skipping error:
etcdctl/ctlv3/command/elect_command.go:76:36: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (SA1016)
etcdctl/ctlv3/command/elect_command.go:110:36: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (SA1016)
etcdctl/ctlv3/command/lock_command.go:60:36: os.Kill cannot be trapped (did you mean syscall.SIGTERM?) (SA1016)
etcdctl/ctlv3/command/snapshot_command.go:313:36: os.SEEK_END is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
etcdctl/ctlv3/command/snapshot_command.go:320:25: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
etcdctl/ctlv3/command/snapshot_command.go:338:26: os.SEEK_END is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
etcdctl/ctlv3/command/snapshot_command.go:356:27: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
skipping error:
etcdmain/etcd.go:338:27: prometheus.Handler is deprecated: Please note the issues described in the doc comment of InstrumentHandler. You might want to consider using promhttp.Handler instead (which is non instrumented).  (SA1019)
etcdmain/grpc_proxy.go:131:29: prometheus.Handler is deprecated: Please note the issues described in the doc comment of InstrumentHandler. You might want to consider using promhttp.Handler instead (which is non instrumented).  (SA1019)
skipping error:
[...]

PR #20010, opened 2025-05-22
remove deprecated SyncTicker in EtcdServer

silentred:
It seems [`SyncTicker`](https://github.com/etcd-io/etcd/blob/c849507a3895a93742d83845635e17a069a8ee2e/server/etcdserver/server.go#L269) in `EtcdServer` is no longer used. Should we remove this field?  @serathius

### DIFF
--- a/server/etcdserver/raft_test.go
+++ b/server/etcdserver/raft_test.go
@@ -278,11 +278,10 @@ func TestProcessDuplicatedAppRespMessage(t *testing.T) {
 	})
 
 	s := &EtcdServer{
-		lgMu:       new(sync.RWMutex),
-		lg:         zaptest.NewLogger(t),
-		r:          *r,
-		cluster:    cl,
-		SyncTicker: &time.Ticker{},
+		lgMu:    new(sync.RWMutex),
+		lg:      zaptest.NewLogger(t),
+		r:       *r,
+		cluster: cl,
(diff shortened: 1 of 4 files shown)


silentred:
> Would be good to confirm what it was used for before and when we stopped using it. Maybe fact that it's not used is a bug. Can you look into that? If we confirm that we can just remove it.

sure, I will look into it.

silentred:
@serathius `SycnTicker` was used to sync TTL for v2 store. The code is referenced as below. 
```
func (s *EtcdServer) run() {
    for {
		select {
		case ap := <-s.r.apply():
		// getSyncC() returns SyncTicker.C
		case <-getSyncC():
		    if s.v2store.HasTTLKeys() {
		        s.sync(s.Cfg.ReqTimeout())
		    }
	}
}
```
The `getSyncC` case was removed in the [PR 17006](https://github.com/etcd-io/etcd/pull/17006)
I think `SycnTicker` is no longer needed.

serathius:
Please sign the DCO https://github.com/etcd-io/etcd/pull/20010/checks?check_run_id=42704086653

ahrtr (approved):
Thanks for the cleanup.

It was added 8 years ago in https://github.com/etcd-io/etcd/pull/7347

PR #19911, opened 2025-05-10
[release-3.4] Update make verify for prow job

joshjms:
Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Add relevant checks to `make verify` for the `pull-etcd-verify` prow job. 

<details>
<summary>Output using <code>pj-on-kind</code> (expand for issues) </summary>
<pre>
➜  1921752994244726784 cat build-log.txt                
./scripts/verify_go_versions.sh
go: downloading github.com/soheilhy/cmux v0.1.5
go: downloading github.com/coreos/go-systemd v0.0.0-20180511133405-39ca1b05acc7
go: downloading github.com/spf13/cobra v0.0.3
go: downloading github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
go: downloading github.com/coreos/pkg v0.0.0-20160727233714-3ac0863d7acf
go: downloading go.uber.org/zap v1.10.0
go: downloading golang.org/x/crypto v0.35.0
go: downloading golang.org/x/net v0.36.0
go: downloading google.golang.org/grpc v1.58.3
go: downloading sigs.k8s.io/yaml v1.1.0
go: downloading github.com/jonboulle/clockwork v0.1.0
go: downloading github.com/grpc-ecosystem/grpc-gateway v1.11.0
go: downloading github.com/tmc/grpc-websocket-proxy v0.0.0-20200427203606-3cfed13b9966
go: downloading go.etcd.io/bbolt v1.3.11
go: downloading github.com/gogo/protobuf v1.3.2
go: downloading github.com/golang/protobuf v1.5.4
go: downloading github.com/prometheus/client_golang v1.11.1
go: downloading github.com/spf13/pflag v1.0.1
go: downloading github.com/coreos/go-semver v0.2.0
go: downloading github.com/dustin/go-humanize v0.0.0-20171111073723-bb3d318650d4
go: downloading golang.org/x/sys v0.30.0
go: downloading golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2
go: downloading github.com/json-iterator/go v1.1.11
go: downloading github.com/modern-go/reflect2 v1.0.1
go: downloading gopkg.in/yaml.v2 v2.4.0
go: downloading github.com/gorilla/websocket v1.4.2
go: downloading github.com/sirupsen/logrus v1.9.3
go: downloading github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2
[...]

PR #20011, opened 2025-05-22
[release-3.4] Add verify-mod-tidy Makefile target

joshjms:

Adds `verify-mod-tidy` to `make verify`. 

Ref: https://github.com/kubernetes/test-infra/issues/32754

/cc @ivanvc @abdurrehman107

### DIFF
--- a/build
+++ b/build
@@ -105,8 +105,6 @@ tools_build() {
 	done
 }
 
-toggle_failpoints_default
-
 if [[ "${ETCD_SETUP_GOPATH:-}" == "1" ]]; then
 	etcd_setup_gopath
 fi
--- a/test
+++ b/test
@@ -426,6 +426,26 @@ EOF
 	VERSION="${VERSION}" "./scripts/test_images.sh"
 }
 
+function mod_tidy_pass {
+	if ! output=$(go mod tidy -diff); then
+		echo -e "go mod tidy checking failed for go.mod (!=0 return code):\\n${output}"
+		exit 255
+	fi
+	if [ -n "${output}" ]; then
+		echo -e "go mod tidy checking failed for go.mod (printed output):\\n${output}"
+		exit 255
+	fi
+
+	if ! output=$(cd ./tools/mod && go mod tidy -diff); then
+		echo -e "go mod tidy checking failed for tools/mod/go.mod (!=0 return code):\\n${output}"
+		exit 255
+	fi
+	if [ -n "${output}" ]; then
+		echo -e "go mod tidy checking failed for tools/mod/go.mod (printed output):\\n${output}"
+		exit 255
+	fi
+}
+
 function shellcheck_pass {
 	SHELLCHECK=shellcheck
 
--- a/Makefile
+++ b/Makefile
@@ -555,7 +555,7 @@ gofail-disable: install-gofail
 	PASSES="toggle_failpoints" ./test
 
 .PHONY: verify
-verify: verify-go-versions verify-gofmt verify-bom verify-dep verify-shellcheck verify-goword verify-govet verify-revive verify-license-header verify-receiver-name verify-markdown-you verify-markdown-marker
+verify: verify-go-versions verify-gofmt verify-bom verify-dep verify-shellcheck verify-goword verify-govet verify-revive verify-license-header verify-mod-tidy verify-receiver-name verify-markdown-you verify-markdown-marker
 
 .PHONY: verify-shellcheck
 verify-shellcheck:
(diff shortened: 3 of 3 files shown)


joshjms:
Another ref: https://github.com/etcd-io/etcd/pull/19911

test
```diff
@@ -426,6 +426,10 @@ EOF
 	VERSION="${VERSION}" "./scripts/test_images.sh"
 }
 
+function mod_tidy_pass {
+	go mod tidy -diff
```
ivanvc, line 430: This is partially correct because of:

1. In the main branch, we execute this check by wrapping it with a call to [`generic_checker`](https://github.com/etcd-io/etcd/blob/0167cb307510f58129ff365fa12f145c42b10104/scripts/test.sh#L163-L178), which checks that the stdout is empty after it runs.
2. We have two modules in 3.4 that we want to check (the top level) `go.mod`, and `tools/mod/go.mod`.

ivanvc (commented):
Thanks for the pull request, @joshjms. Please take a look at my comment.

test
joshjms: Ah, I see

build
```diff
@@ -105,8 +105,6 @@ tools_build() {
 	done
 }
 
-toggle_failpoints_default
```
ivanvc, line 108: Why should we disable failpoints? I don't understand this change.

joshjms: Because running `toggle_failpoints_default` automatically runs `go mod tidy`. But I suppose it's not a good solution, I'll take time to understand the script first.

test
```diff
@@ -426,6 +426,26 @@ EOF
...
 
+function mod_tidy_pass {
+	if ! output=$(go mod tidy -diff); then
+		echo -e "go mod tidy checking failed for go.mod:\\n${output}"
+		exit 255
+	fi
+	if [ -n "${output}" ]; then
+		echo -e "go mod tidy checking failed for go.mod:\\n${output}"
+		exit 255
+	fi
```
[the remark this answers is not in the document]
joshjms: Sorry, in the main branch, they are separated for logging reasons (!=0 return code and printed output respectively), I will update the error logs.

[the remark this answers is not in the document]
joshjms: Would you rather for it to be consolidated or like this:
```bash
if ! output=$(go mod tidy -diff); then
	echo -e "go mod tidy checking failed for go.mod (!=0 return code):\\n${output}"
	exit 255
fi
if [ -n "${output}" ]; then
	echo -e "go mod tidy checking failed for go.mod (printed output):\\n${output}"
	exit 255
fi
```

build
joshjms: The function in question:

```bash
toggle_failpoints() {
	mode="$1"
	if command -v gofail >/dev/null 2>&1; then
		gofail "$mode" etcdserver/ lease/leasehttp/ mvcc/ mvcc/backend/ wal/
		# shellcheck disable=SC2086
		if [[ "$mode" == "enable" ]]; then
			go get go.etcd.io/gofail@${GOFAIL_VERSION}
			cd ./etcdserver && go get go.etcd.io/gofail@${GOFAIL_VERSION}
			cd ../etcdctl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
			cd ../tests && go get go.etcd.io/gofail@${GOFAIL_VERSION}
			cd ../
		else
			go mod tidy
			cd ./etcdserver && go mod tidy
			cd ../etcdctl && go mod tidy
			cd ../tests && go mod tidy
			cd ../
		fi
	elif [[ "$mode" != "disable" ]]; then
		echo "FAILPOINTS set but gofail not found"
		exit 1
	fi
}
```

joshjms: By the way, just to clarify, from my understanding (I might be wrong), running `toggle_failpoints_default` doesn't disable failpoints. It just runs `go mod tidy` when the FAILPOINTS env variable is empty (i.e. when failpoints are set to be disabled). Which is why I don't know why we should run it by default since other commands which need it already have that (so it's run twice). When I removed it, there also seems to be no problem with the current CI checks.

ivanvc: Alright, thanks for the clarification. After doing some research myself, I also found that this function call is unnecessary. The same function is called from `etcd_build()`:

https://github.com/etcd-io/etcd/blob/74fdeaae188c2cf707758179dd366d1f77b074a9/build#L68-L71

The issue is that if this script is sourced, it will run the `toggle_failpoints_default` function.

I suggest making an individual commit for this change and documenting the issue in the body of the commit. That way, if we ever need to go back and understand why it was there and later removed, we can do so by leaving a good paper trail.

test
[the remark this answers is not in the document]
joshjms: TIL that WET is the opposite of DRY. It took me a while to get it.

jmhbnz (approved):
LGTM - Thanks team

PR #20020, opened 2025-05-24
Handle duplicate key delete in one TXN

serathius:
The PR https://github.com/etcd-io/etcd/pull/20002 added TXN with multiple delete operations on single key. Didn't know that was even possible as I don't think it's allowed for Puts.

Still, the logic for generating events for such TXN operation is wrong.

/cc @joshjms @henrybear327 @nwnt @siyuanfoundation @fuweid

### DIFF
--- a/tests/robustness/model/replay.go
+++ b/tests/robustness/model/replay.go
@@ -74,7 +74,7 @@ func toWatchEvents(prevState *EtcdState, request EtcdRequest, response MaybeEtcd
 		} else {
 			ops = request.Txn.OperationsOnSuccess
 		}
-		for _, op := range ops {
+		for i, op := range ops {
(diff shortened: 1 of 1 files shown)


joshjms:
I think the change looks good to me for handling duplicate delete operations. Although the issue with PR #20002 is my fault as we're not supposed to even have duplicate keys in a single txn, as you already mentioned. 

> Txn processes multiple requests in a single transaction. A txn request increments the revision of the key-value store and generates events with the same revision for every completed request. It is not allowed to modify the same key several times within one txn.

serathius:
I don't think the problem is in your PR, rather with etcd.

joshjms (approved):

PR #20022, opened 2025-05-25
concurrency: add timeout for LeaseGrant in NewSession

### LINKED ISSUE #19778: Missing deadline for LeaseGrant when creating session
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).

### What happened?

In the `NewSession` function within `client/v3/concurrency/session.go`, the RPC call `LeaseGrant` is invoked without a deadline, as shown below.

```go
resp, err := client.Grant(ops.ctx, int64(ops.ttl))
```

However, `LeaseGrant` should be called with a deadline set to the `ttl` value, similar to how the `LeaseRevoke` RPC is handled when closing the session. Otherwise it may stuck and block the `NewSession` function endlessly.

```go
ctx, cancel := context.WithTimeout(s.opts.ctx, time.Duration(s.opts.ttl)*time.Second)
_, err := s.client.Revoke(ctx, s.id)
```

RPC `LeaseGrant` can return `DeadlineExceeded` to avoid endless stuck.

`NewSession` can be called in the distributed database using etcd client.

For example, execute following statements in TiDB.

```sql
create table ddl__column_modify2.test (a int auto_increment primary key, b int);
insert ddl__column_modify2.test values (1, 1);
update ddl__column_modify2.test set b = b + 1 where a = 1;
insert into ddl__column_modify2.test values (2, 2);
insert into ddl__column_modify2.test (a) values (3);
insert into ddl__column_modify2.test values (4, 4);
insert into ddl__column_modify2.test (a) values (5);
insert ddl__column_modify2.test values (6, 6);
alter table ddl__column_modify2.test add index idx1 (a, b);
```

_No response_

<details>

```console
$ etcd --version
v3.5.12

$ etcdctl version
# paste output here
```

</details>

<details>

</details>

<details>

```console
# paste output here

# paste output here
```

</details>

### Relevant log output

```Shell

```

Antraxmin:
## What this PR does
This PR adds a timeout to the `LeaseGrant` RPC call in the `NewSession` function to prevent infinite blocking.

- Fixes #19778 

## Why this change is needed
Currently, `LeaseGrant` in `NewSession` is called without a deadline, which can cause the function to block indefinitely if the RPC call hangs. This is inconsistent with the `Close()` method, which properly uses `context.WithTimeout` for the `LeaseRevoke` call.

## Changes made
- Modified `NewSession` function in `client/v3/concurrency/session.go` to use `context.WithTimeout` for `LeaseGrant` call
- The timeout is set to `ttl` seconds, matching the pattern used in `Session.Close()`
- Added integration tests to verify timeout behavior

## Testing
- All existing integration tests pass
- Added new tests to verify timeout behavior works correctly
- Verified normal session creation still works as expected

### DIFF
--- a/client/v3/concurrency/session.go
+++ b/client/v3/concurrency/session.go
@@ -47,7 +47,9 @@ func NewSession(client *v3.Client, opts ...SessionOption) (*Session, error) {
 
 	id := ops.leaseID
 	if id == v3.NoLease {
-		resp, err := client.Grant(ops.ctx, int64(ops.ttl))
+		ctx, cancel := context.WithTimeout(ops.ctx, time.Duration(ops.ttl)*time.Second)
+		resp, err := client.Grant(ctx, int64(ops.ttl))
+		cancel()
 		if err != nil {
 			return nil, err
 		}
--- a/tests/integration/clientv3/concurrency/session_timeout_test.go
+++ b/tests/integration/clientv3/concurrency/session_timeout_test.go
@@ -1,0 +1,218 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package concurrency_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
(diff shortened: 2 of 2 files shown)


siyuanfoundation:
@Antraxmin Can you fix your DCO first? Thanks!

client/v3/concurrency/session.go
```diff
@@ -47,7 +47,9 @@ func NewSession(client *v3.Client, opts ...SessionOption) (*Session, error) {
 
 	id := ops.leaseID
 	if id == v3.NoLease {
-		resp, err := client.Grant(ops.ctx, int64(ops.ttl))
+		ctx, cancel := context.WithTimeout(ops.ctx, time.Duration(ops.ttl)*time.Second)
```
silentred, line 50: It is not appropriate to use `ttl` as timeout. Maybe we could add a default timeout constant. For example, `etcdctl` has a command timeout `defaultCommandTimeOut   = 5 * time.Second`

PR #20037, opened 2025-05-27
Increase the number of clients in antithesis tests to match robustness

serathius:
Think we need to increase number of clients to improve quality of space exploration.

/hold


serathius:
/hold

Would like to qualify all etcd branches before we make another change in antithesis testing.

serathius:
Too many timeouts to increase client count.

PR #20046, opened 2025-05-28
[release-3.5] makefile: Split fmt into separate verify commands

abdurrehman107:
Manual backport of https://github.com/etcd-io/etcd/pull/14481

cc @ivanvc @joshjms


ivanvc:
Link to https://github.com/kubernetes/test-infra/issues/32754

joshjms:
release-3.5 doesn't use golangci-lint so there will not be a verify-lint target. We already use revive for this branch so I don't think we should introduce golangci-lint.

jmhbnz (approved):
LGTM - Thanks for tidying this up team 👍🏻

PR #20048, opened 2025-05-28
Simplify Makefile by removing set-version-* targets

serathius:
/cc @fuweid @nwnt @henrybear327

### DIFF
--- a/tests/antithesis/Makefile
+++ b/tests/antithesis/Makefile
@@ -13,16 +13,20 @@ antithesis-build-etcd-image:
 	docker build --build-arg GO_VERSION=$(shell cat $(REPOSITORY_ROOT)/.go-version) --build-arg REF=${REF} $(REPOSITORY_ROOT)/tests/antithesis/server/ -t etcd-server:latest
 
 .PHONY: antithesis-build-etcd-image-release-3.4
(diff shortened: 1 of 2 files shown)


tests/antithesis/Makefile
```diff
@@ -13,16 +13,20 @@ antithesis-build-etcd-image:
...
+antithesis-build-etcd-image-release-3.5: antithesis-build-etcd-image
 
 .PHONY: antithesis-build-etcd-image-release-3.6
-antithesis-build-etcd-image-release-3.6: set-version-3.6 antithesis-build-etcd-image
+antithesis-build-etcd-image-release-3.6: REF=release-3.6
+antithesis-build-etcd-image-release-3.6: antithesis-build-etcd-image
 
 .PHONY: antithesis-build-etcd-image-main
-antithesis-build-etcd-image-main: set-version-main antithesis-build-etcd-image
+antithesis-build-etcd-image-main: REF=main
```
ivanvc, line 28: `REF` is set in line 5, so it has a default value set to `main`. So, setting it here may be unnecessary.

serathius: This is meant to maintain same name pattern for branches like `release-3.5` and `main`. Current default is `main` but we might change it in the future.

PR #20049, opened 2025-05-29
Fix multi-version concurrent tests

### LINKED ISSUE #20036: [Antithesis] Concurrent tests cause incorrect server image being used

With a ability of running tests on different etcd branches (main, release-3.6, release-3.5, release-3.4), I started 4 jobs for each of branches. Unfortunately I discovered in logs that only 2 branches were tested (main and release-3.6). 

As antithesis reuse same docker tag `etcd-server:latest` concurrent tests can override the image impacting execution of other tests.

Fix would require us to use unique image tag per branch. To implement it we would need to:
* Tag docker images based on GIT_REF they were built from
* Template the docker-compose file injected into config image to include a tag based on GIT_REF
* Move the images input from github workflow to be generated based on GIT_REF.

/cc @nwnt

### Why is this needed?

Ensure tests don't impact each other

nwnt:
Fix #20036


serathius:
> @serathius do we need to also add the built image tag into the list of images in inputs.images or is this something you set when you run the job?

Yes we should add it. List of images being a input parameter is a leftover from original config, that doesn't make sense to me.

serathius:
> I was thinking we can try to automate that by using something like this: 

We could just not make it a input.

tests/antithesis/docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
         chown -R ${USER_ID:-root}:${GROUP_ID:-root} /var/etcddata0 /var/etcddata1 /var/etcddata2 /var/report
 
   etcd0:
-    image: 'etcd-server:latest'
+    image: 'etcd-server:${REF:-latest}'
```
serathius, line 24: This will not work because the Antithesis doesn't template the file. We would also need to render this templating into a config file.

serathius:
This PR is missing steps described in issue https://github.com/etcd-io/etcd/issues/20036

tests/antithesis/docker-compose.yml
nwnt: Sorry about the confusion.  I went to read Antithesis's documentation and now understand what you mean.  This newly pushed one should do what is required... I think.

.github/workflows/antithesis-test.yml
```diff
@@ -19,11 +19,6 @@ on:
         required: true
         default: us-central1-docker.pkg.dev/molten-verve-216720/linuxfoundation-repository/etcd-config:latest
```
serathius, line 20: Same issue for config, it needs to be uniquely tagged.

tests/antithesis/Makefile
```diff
@@ -28,9 +28,17 @@ antithesis-build-etcd-image-release-3.6: antithesis-build-etcd-image
 antithesis-build-etcd-image-main: REF=main
 antithesis-build-etcd-image-main: antithesis-build-etcd-image
 
+.PHONY: antithesis-templatize-docker-compose
+antithesis-templatize-docker-compose:
+	export REF=$(REF) && cat ./docker-compose.yml | envsubst > docker-compose-config.yml
```
serathius, line 33: Maybe we could render the docker-file when building docker image?

serathius: Like 
```
FROM docker.io/library/ubuntu:latest as build
RUN \
  apt-get update \
  && apt-get -y install gettext-base \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
ARG REF
COPY docker-compose.yml /docker-compose.yml.template
RUN cat /docker-compose.yml.template | envsubst > /docker-compose.yml
FROM scratch
COPY --from=build /docker-compose.yml /docker-compose.yml
```

nwnt: Yes, this is a better way.  Thanks for the feedback Marek.

tests/antithesis/docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
         chown -R ${USER_ID:-root}:${GROUP_ID:-root} /var/etcddata0 /var/etcddata1 /var/etcddata2 /var/report
 
   etcd0:
-    image: 'etcd-server:latest'
+    image: 'etcd-server:${REF}'
```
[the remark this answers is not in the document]
serathius: It would allow us to skip passing  `export REF=$(REF)`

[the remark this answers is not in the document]
nwnt: Done.  I had to change the util though as the original `envsubst` tool doesn't support substituting variables with a default value.

.github/workflows/antithesis-test.yml
```diff
@@ -68,14 +58,15 @@ jobs:
...
-          tags: |
-            ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:latest,
-            ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:${{ github.sha }}
+        working-directory: ./tests/antithesis
+        run: |
+          make antithesis-build-antithesis-config-image REF=${{ inputs.etcd_ref }}
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:${{ inputs.etcd_ref }}_${{ github.sha }}"
+          docker tag etcd-config:latest $TAG
+          docker push $TAG
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:${{ github.sha }}"
```
serathius, line 67: Nit, don't think we need more tags.

.github/workflows/antithesis-test.yml
```diff
@@ -68,14 +58,12 @@ jobs:
...
-            ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:latest,
-            ${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:${{ github.sha }}
+        working-directory: ./tests/antithesis
+        run: |
+          make antithesis-build-antithesis-config-image REF=${{ inputs.etcd_ref }}
+          export TAG="${{ env.REGISTRY }}/${{ env.REPOSITORY }}/etcd-config:${{ inputs.etcd_ref }}_${{ github.sha }}"
+          docker tag etcd-config:latest $TAG
+          docker push $TAG
 
       - name: Build and push client image
```
serathius, line 68: For future, maybe we use git_sha based tag too.

tests/antithesis/Dockerfile.config
```diff
@@ -1,3 +1,11 @@
-from scratch
+ARG GO_VERSION=1.24.3
 
-COPY docker-compose.yml /docker-compose.yml
+FROM golang:$GO_VERSION AS build
+RUN go install github.com/a8m/envsubst/cmd/envsubst@latest
+
+ARG REF
+COPY docker-compose.yml /docker-compose.yml.template
+RUN export REF=${REF} && cat /docker-compose.yml.template | envsubst > /docker-compose.yml
```
serathius, line 8:
```suggestion
RUN REF=${REF} cat /docker-compose.yml.template | envsubst > /docker-compose.yml
```

tests/antithesis/Makefile
```diff
@@ -28,9 +28,13 @@ antithesis-build-etcd-image-release-3.6: antithesis-build-etcd-image
...
 antithesis-build-etcd-image-main: antithesis-build-etcd-image
 
+.PHONY: antithesis-build-config-image
+antithesis-build-config-image:
+	docker build -f ./Dockerfile.config . -t etcd-config:latest -t etcd-config:$(REF) --build-arg REF=$(REF)
+
 .PHONY: antithesis-docker-compose-up
 antithesis-docker-compose-up:
-	export USER_ID=$(USER_ID) && export GROUP_ID=$(GROUP_ID) && docker-compose up
+	export USER_ID=$(USER_ID) && export GROUP_ID=$(GROUP_ID) && export REF=$(REF) && docker-compose up
```
serathius, line 37: Default REF is `main`, instead maybe we could not template it and depend latest being default?

nwnt:
this should take care of them.  let me know if you have more comments.

serathius (approved):

PR #20054, opened 2025-05-29
Set antithesis source to keep separate failure history for runs from different branches

serathius:
/cc @fuweid @henrybear327 @joshjms @nwnt


marcus-hodgson-antithesis:
LGTM!

henrybear327 (approved):

PR #20055, opened 2025-05-29
Fix the issue that `--force-new-cluster` can't remove all other members in a corner case

ahrtr:
See discussion in https://github.com/etcd-io/raft/pull/300. 

`--force-new-cluster` will automatically remove all other members on bootstrap, but it won't work in a corner case (HardState isn't synched to disk, and then etcd crashes). Reproduced the issue in the first commit, and fixed it in the second commit,

```
    force_new_cluster_test.go:98: Online checking the member count
    force_new_cluster_test.go:101: 
        	Error Trace:	/Users/wachao/go/src/github.com/ahrtr/etcd/tests/e2e/force_new_cluster_test.go:101
        	Error:      	"[ID:4075498731181859066 name:"TestForceNewCluster_MemberCount-test-1" peerURLs:"http://localhost:20006" clientURLs:"http://localhost:20005"  ID:14578408409545168728 name:"TestForceNewCluster_MemberCount-test-0" peerURLs:"http://localhost:20001" clientURLs:"http://localhost:20000" ]" should have 1 item(s), but has 2
        	Test:       	TestForceNewCluster_MemberCount
--- FAIL: TestForceNewCluster_MemberCount (10.14s)
```

### DIFF
--- a/server/etcdserver/bootstrap.go
+++ b/server/etcdserver/bootstrap.go
@@ -87,7 +87,8 @@ func bootstrap(cfg config.ServerConfig) (b *bootstrappedServer, err error) {
 		if err = fileutil.IsDirWriteable(cfg.WALDir()); err != nil {
 			return nil, fmt.Errorf("cannot write to WAL directory: %w", err)
 		}
-		bwal = bootstrapWALFromSnapshot(cfg, backend.snapshot)
+		cfg.Logger.Info("Bootstrapping WAL from snapshot")
+		bwal = bootstrapWALFromSnapshot(cfg, backend.snapshot, backend.ci)
 	}
 
 	cfg.Logger.Info("bootstrapping cluster")
@@ -551,7 +552,7 @@ func (b *bootstrappedRaft) newRaftNode(ss *snap.Snapshotter, wal *wal.WAL, cl *m
 	)
 }
 
-func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot) *bootstrappedWAL {
+func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot, ci cindex.ConsistentIndexer) *bootstrappedWAL {
 	wal, st, ents, snap, meta := openWALFromSnapshot(cfg, snapshot)
 	bwal := &bootstrappedWAL{
 		lg:       cfg.Logger,
@@ -564,6 +565,19 @@ func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot
 	}
 
 	if cfg.ForceNewCluster {
+		consistentIndex := ci.ConsistentIndex()
+		oldCommitIndex := bwal.st.Commit
+		// If only `HardState.Commit` increases, HardState won't be persisted
+		// to disk, even though the committed entries might have already been
+		// applied. This can result in consistent_index > CommitIndex.
+		//
+		// When restarting etcd with `--force-new-cluster`, all uncommitted
+		// entries are dropped. To avoid losing entries that were actually
+		// committed, we reset Commit to max(HardState.Commit, consistent_index).
+		//
+		// See: https://github.com/etcd-io/raft/pull/300 for more details.
+		bwal.st.Commit = max(oldCommitIndex, consistentIndex)
+
 		// discard the previously uncommitted entries
 		bwal.ents = bwal.CommitedEntries()
 		entries := bwal.NewConfigChangeEntries()
@@ -573,6 +587,7 @@ func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot
 			"forcing restart member",
 			zap.String("cluster-id", meta.clusterID.String()),
 			zap.String("local-member-id", meta.nodeID.String()),
+			zap.Uint64("wal-commit-index", oldCommitIndex),
 			zap.Uint64("commit-index", bwal.st.Commit),
 		)
 	} else {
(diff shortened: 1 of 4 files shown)


server/storage/wal/wal.go
```diff
@@ -962,6 +962,11 @@ func (w *WAL) Save(st raftpb.HardState, ents []raftpb.Entry) error {
 	}
 
 	mustSync := raft.MustSync(st, w.state, len(ents))
+	// We should sync the HardState to disk as long as it differs from previous value.
+	// See discussion in https://github.com/etcd-io/raft/pull/300.
+	if st != w.state {
```
serathius, line 967: I'm worried about overhead of forcing sync for every entry just to fix `--force-new-cluster`. From what I understand this will introduce syncing every time commit index increases.

Have you tried suggestion by @pav-kv https://github.com/etcd-io/raft/pull/300#issuecomment-2910081042?

ahrtr: Updated

ahrtr:
@fuweid @ivanvc @jmhbnz @serathius PTAL

Once this PR gets merged & backported, then we are ready to release v3.6.1.

server/etcdserver/bootstrap.go
```diff
@@ -573,6 +578,7 @@ func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot
 			"forcing restart member",
 			zap.String("cluster-id", meta.clusterID.String()),
 			zap.String("local-member-id", meta.nodeID.String()),
+			zap.Uint64("old-commit-index", oldCommitIndex),
```
serathius, line 581:
```suggestion
			zap.Uint64("wal-commit-index", oldCommitIndex),
```

server/etcdserver/bootstrap.go
```diff
@@ -564,6 +565,10 @@ func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot
 	}
 
 	if cfg.ForceNewCluster {
+		consistentIndex := ci.ConsistentIndex()
+		oldCommitIndex := bwal.st.Commit
+		bwal.st.Commit = max(oldCommitIndex, consistentIndex)
```
serathius, line 570: Can you leave a comment explaining why we take max?

server/etcdserver/bootstrap.go
```diff
@@ -564,6 +565,19 @@ func bootstrapWALFromSnapshot(cfg config.ServerConfig, snapshot *raftpb.Snapshot
...
 
 	if cfg.ForceNewCluster {
+		consistentIndex := ci.ConsistentIndex()
+		oldCommitIndex := bwal.st.Commit
+		// If only `HardState.Commit` increases, HardState won't be persisted
+		// to disk, even though the committed entries might have already been
+		// applied. This can result in consistent_index > CommitIndex.
+		//
+		// When restarting etcd with `--force-new-cluster`, all uncommitted
+		// entries are dropped. To avoid losing entries that were actually
```
pav-kv, line 575: Are you leaving any log in place, or the index/term of the entry at which the log is truncated? One caveat to mention is, depending how synchronization between the log and the state machine works in your case, you may end up not just with a higher applied index, but also one for each the pre-image entry does not exist in the log. But you might need to detect this case.

Scenario:
1. Log is `[1][1][1][2][2][2]`, with indices 1..6
2. The node receives an overriding append / `MsgApp` from a new leader, which also says that all these entries are committed: log is now `[1][1][3][3]`, with commit index = 4. Sends this mutation to log storage.
3. Entries `[1][1][3][3]` are sent for state machine application.

If steps (2) and (3) are parallel, it's possible that on crash/restart you end up with:
- state machine applied `[1][1][3][3]`
- the log is `[1][1][1][2][2][2]`, i.e. it forgot the last append

In this case, the state machine's last entry at index 4 (known to be committed) does not match the log. Which is fine: it just means that the state machine completely obsoletes the log, and the log should be removed.

ahrtr: > Scenario:
> 
> 1. Log is `[1][1][1][2][2][2]`, with indices 1..6
> 2. The node receives an overriding append / `MsgApp` from a new leader, which also says that all these entries are committed: log is now `[1][1][3][3]`, with commit index = 4. Sends this mutation to log storage.
> 3. Entries `[1][1][3][3]` are sent for state machine application.

In etcd's case, when restarting with `--force-new-cluster`, it will forcibly create an one-member cluster.  So it won't have this problem.



> If steps (2) and (3) are parallel, it's possible that on crash/restart you end up with:
> 
> * state machine applied `[1][1][3][3]`
> * the log is `[1][1][1][2][2][2]`, i.e. it forgot the last append
> 
> In this case, the state machine's last entry at index 4 (known to be committed) does not match the log. Which is fine: it just means that the state machine completely obsoletes the log, and the log should be removed.

I don't think it's true. Raft won't commit any entries before them being replicated (& synced to WAL). Pls refer to https://github.com/etcd-io/etcd/issues/14370#issuecomment-1254645353

ahrtr: > Are you leaving any log in place, or the index/term of the entry at which the log is truncated?

Note that this PR only changes the logic with `--force-new-cluster`. etcd discards all uncommitted entries (which `index > max(commitIndex, applieIndex)`).  All entries <= `max(commitIndex, applieIndex)` will keep it as it's. 



> but also one for each the pre-image entry does not exist in the log.

> In this case, the state machine's last entry at index 4 (known to be committed) does not match the log.

This isn't a problem in the  `--force-new-cluster` use case, even if etcd leaves some unmatched logs in the WAL file. That said, such unmatched logs are unlikely to occur in production environments, since etcd periodically (asynchronously) commits applied data while writes WAL entries synchronously. In any case, those entries will either be ignored or overwritten.
- etcd won't re-apply any already applied entries.
- any committed but not yet applied entries will be applied.

Issue #19371, opened 2025-02-10
Develop a caching library for etcd


serathius:
Submitted as project as part of Google Summer of Code with @MadhavJivrajani as second mentor.

While etcd is a powerful distributed key-value store, building scalable infrastructure management systems directly on top of it can be challenging.  Kubernetes has demonstrated the effectiveness of the reconciliation pattern for managing complex deployments, and its watch cache plays a crucial role in achieving scalability.  However, this crucial caching mechanism is tightly coupled with Kubernetes and not readily available for general etcd usage.  Projects like Cilium and Calico Typha, while successfully using etcd for control planes, have had to implement custom solutions to address this gap.

This project addresses the need for a standardized, performant caching solution for etcd, enabling easier adoption of the reconciliation pattern and simplifying the development of scalable etcd-based systems.  By providing a generic watch cache implementation, we aim to lower the barrier to entry for building robust and efficient infrastructure management tools on etcd.

Goals:
* Develop a generic proxy that provides feature parity to K8s watch cache
* Enable possibility of integrating into K8s and Cilium.

Milestones:
* Cache for watch requests, stores history of watch events and demultiplexes requests. Combining multiple local Watch requests into single watch to etcd. 
* Cache for non-consistent list requests, stores latest state of etcd in Btree cache. Cache is fed by Range response, that is later updated by subscribing to updates from Watch Cache.
* Handling requests during cache initialization and re-intialization. 
* Testing, including e2e and robustness tests
* Metrics for cache size, cache latency etc
* Benchmarks for watch and read throughput. 
* Support for custom encoder/decoder
* Support for custom indexing
* Support for consistent reads
* Support for exact stale reads, by storing snapshots of btree.

I'm proposing to locate the project within the etcd mono repo, but as a separate package, that will not be released/tagged until it's ready. Proposed package name: go.etcd.io/cache. Client library would be developed under go.etcd.io/cache/client.
 

/cc @fuweid @MadhavJivrajani @ahrtr @henrybear327

ahrtr:
High level I agree with the improvement & direction, as performance should be one of the key areas that we should spend more effort on. It will definitely ensure the long-term success of etcd.

fuweid:
> We need to develop a generic cache for etcd, that allows users to easily addopt multi layered caching architecture similar to K8s. Having an official library would allow us to properly test it ensuring it's correctness and performance.

Sounds great. It could be more efficient to make and evaluate changes as an official library. +1 for help, if need.

[...]

PR #20056, opened 2025-05-29
Setup directory for cache

serathius:
Needed for https://github.com/etcd-io/etcd/issues/19371

/cc @ahrtr @fuweid @siyuanfoundation @jmhbnz @ivanvc

### DIFF
--- a/cache/LICENSE
+++ b/cache/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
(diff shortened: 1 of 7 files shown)


cache/LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
```
ahrtr, line 5: I think we can reuse the existing LICENSE file?

cache/go.mod
```diff
@@ -0,0 +1,5 @@
+module go.etcd.io/etcd/cache
```
ahrtr, line 1: Can it just be a package under the [client/v3](https://github.com/etcd-io/etcd/tree/main/client/v3) module? If not, what's the relationship between client/v3 and the cache?

Will the cache only be used by applications which talk to etcdserver?

serathius: > Can it just be a package under the [client/v3](https://github.com/etcd-io/etcd/tree/main/client/v3) module?

I proposed to have it as separate package as cache will have it's own set of unique dependencies. For performance, cache will be using data structures like btree from https://github.com/google/btree and possibly many more. I don't think we should force those dependencies of users that just want to talk to etcd.

> what's the relationship between client/v3 and the cache?

Cache will import client/v3 and implement interfaces from it like clientv3.KV or clientv3.Watch that would be served from cache instead of etcd.

 > Will the cache only be used by applications which talk to etcdserver?
  
 Yes, what other cases you are thinking about?

ahrtr: > Cache will import client/v3 and implement interfaces from it like clientv3.KV or clientv3.Watch that would be served from cache instead of etcd.

You should have more knowledge on the cache given it's the similar implementation coming from the Kubernetes repo. But I think we still need some discussion on the user interface, which is hard to change once decided & published. I was thinking from users perspective, what they see is only the client/v3 interface, the cache is just an internal capability that users can enable on demand. I will take a look at your original issue and feedback later. If you have more doc or examples, please share. 

cc others as well @fuweid @ivanvc @jmhbnz

fuweid: Based on my current understanding, I don't think this should be part of the client/v3 package. This package is built on top of the client, similar to how kube-apiserver builds on client/v3. It really depends on the long-term vision for client/v3. That said, it’s still very early in the development phase, and until it reaches v1.0.0, we can always make adjustments as needed.

fuweid (approved):
LGTM

ivanvc:
There's a consideration we need to clear to enable the module in CI scripts (considering we still don't have the Go workspace :sweat_smile:):

First, we need to add the directory to the list of directories from `module_dirs`:
https://github.com/etcd-io/etcd/blob/c904edc1747f8cc73b36c946ee3192b12dc1a8ff/scripts/test_lib.sh#L80-L82

Then, we need to define whether we want their dependencies to be included in the bill of materials. If we decide to add them, we also need to add it here (in the package module format): https://github.com/etcd-io/etcd/blob/c904edc1747f8cc73b36c946ee3192b12dc1a8ff/scripts/test_lib.sh#L96-L109

cache/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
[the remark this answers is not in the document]
serathius: Possibly :P
We are missing it in the whole project.

cache/go.mod
ahrtr: Just because Kubernetes does it that way doesn’t mean we have to follow the exact same path. We're trying to build a more generic module or library, so it's a different story now. That said, I agree it’s easier to start off by doing it the same way, and we can always refactor or improve things later.

Since it’s just a module or layer on top of `client/v3`, and etcd itself doesn’t depend on it at all, I think it makes more sense to put it in a separate repo — that way we don’t clutter the main one. Plus, we can move faster with development if it's kept separate.

cache/LICENSE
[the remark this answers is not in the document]
serathius: Fix https://github.com/etcd-io/etcd/pull/20068

cache/go.mod
serathius: > It feels a bit odd to add a module to the main etcd repo when etcd itself doesn't depend on it at all — just for the sake of reusing the robustness test. That would unnecessarily clutter the repo with more dependencies, a larger BOM, etc.
> 
> I think we should aim to move components out (similar to an out-of-tree plugin approach), rather than bringing more into the core repo.


I don't think it's odd. It think it's a reflection of how tight is the coupling and how stable is the API. Over time modules become mature and stable, not needing the tight coupling, allowing us to move them out of the mono repo. However, this doesn't mean we should throw everything out and not accept any new projects. There is a lot of benefit of being able to develop things together like in main etcd repo.

Moving components out should not be a goal by itself. Some things make sense to be kept together. See the Kubernetes project, while vendor specific things were moved to out-of-tree plugins, the repo has more submodules than ever. 

I failed to mention other reason to keep the cache in the main repo. As part of cache development I expect we will identify possible improvements to the etcd API. For example an API for cache consistency checking. I think it would be beneficial to develop and test those improvements together, etcd and cache.

ahrtr: > It will, however to test correctness you need more advanced tools like model for linearizability testing and watch validation that was developed by robustness tests.

A better approach is to make the rebustness test reusable so that other projects can make use of it out of the box instead of adding them into etcd.



> However, this doesn't mean we should throw everything out and not accept anything new. There is a lot of benefit of being able to develop things together like in main etcd repo.

Apart from the concerns mentioned above, adding something new tends to attract more interest and attention — people are naturally get excited about new things. But unfortunately, no one wants to work on the technical debts (i.e. v1 protobuf, experimental grpc package, non-lineraizable lease issue, poor readability of the source code i.e. watch etc....., also 3.7 priorities), which is actually one of the main things holding back etcd’s long-term growth.

Also to clarify, we are not rejecting new things. I think it's better to experiment new things in a sandbox (a separate repo) first.


> I failed to mention other reason to keep the cache in the main repo. As part of cache development I expect we will identify possible improvements to the etcd API. For example an API for cache consistency checking.

I think the client API is and should be stable. 

That said, I won't block this if you insist on adding it into etcd main repo.

scripts/test_lib.sh
```diff
@@ -78,7 +78,7 @@ function run_for_module {
 }
 
 function module_dirs() {
-  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/mod tools/rw-heatmaps tools/testgrid-analysis ."
+  echo "api pkg client/pkg client/internal/v2 client/v3 server etcdutl etcdctl tests tools/mod tools/rw-heatmaps tools/testgrid-analysis . cache"
```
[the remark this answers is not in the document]
serathius: Done

ivanvc:
@serathius, do you want help adding the module to the test scripts?

ivanvc:
> > @serathius, do you want help adding the module to the test scripts?
> 
> Sure

I noticed that you've already added the directory to `module_dirs()`. That should be enough for now. I believe that once the subproject reaches a more stable state, we can include its dependencies in the BOM.

PR #20068, opened 2025-05-30
Fix LICENSE file

serathius:
Provide date based on file creation date

/cc @ahrtr @ivanvc @jmhbnz @fuweid

### DIFF
--- a/bill-of-materials.json
+++ b/bill-of-materials.json
@@ -472,7 +472,7 @@
 		"licenses": [
 			{
 				"type": "Apache License 2.0",
-				"confidence": 1
+				"confidence": 0.9988925802879292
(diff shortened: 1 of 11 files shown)


bill-of-materials.json
```diff
@@ -472,7 +472,7 @@
 		"licenses": [
 			{
 				"type": "Apache License 2.0",
-				"confidence": 1
+				"confidence": 0.9988925802879292
```
ahrtr, line 475: question: is this automatically updated by the script?

serathius: Yes, verify failed. I assume it calculates distance to license with those values unfilled.

PR #20109, opened 2025-06-03
Remove v2discovery

ahrtr:
v2discovery was deprecated in 3.6, and remove it 3.7.


server/config/config.go
```diff
@@ -315,7 +313,12 @@ func (c *ServerConfig) WALDir() string {
 func (c *ServerConfig) SnapDir() string { return filepath.Join(c.MemberDir(), "snap") }
 
 func (c *ServerConfig) ShouldDiscover() bool {
-	return c.DiscoveryURL != "" || len(c.DiscoveryCfg.Endpoints) > 0
+	for _, ep := range c.DiscoveryCfg.Endpoints {
+		if ep != "" {
```
serathius: Why we would allow empty endpoint?

ahrtr: `ShouldDiscover` returns true as long as there is at least one non-empty endpoint.

serathius: I understand what the code is doing, but why there would be an empty endpoint? Should we validate that all endpoints are not empty?

server/config/config_test.go
```diff
@@ -33,10 +35,14 @@ func mustNewURLs(t *testing.T, urls []string) []url.URL {
...
 
-func TestConfigVerifyBootstrapWithoutClusterAndDiscoveryURLFail(t *testing.T) {
+func TestConfigVerifyBootstrapWithoutClusterFail(t *testing.T) {
 	c := &ServerConfig{
-		Name:               "node1",
-		DiscoveryURL:       "",
+		Name: "node1",
+		DiscoveryCfg: v3discovery.DiscoveryConfig{
+			ConfigSpec: clientv3.ConfigSpec{
+				Endpoints: []string{""},
```
serathius: This test is incorrect. `DiscoveryURL:       "",` should be replaced by empty slice `Endpoints: []string{},` and not a slice with one empty element.

ahrtr: good catch & updated

server/config/config.go
ahrtr: updated. also added a unit test `TestDiscoveryCfg`

ahrtr:
This is another major step to make the main branch much cleaner.

PTAL @fuweid @ivanvc @jmhbnz @serathius

serathius (approved):

Issue #20118, opened 2025-06-05
3.6.0: If ETCDCTL_CACERT, ETCDCTL_CERT, ETCDCTL_KEY or ETCDCTL_ENDPOINTS are set, 'etcdctl endpoint health' will fail


Timothy-Dougherty:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

After upgrading to 3.6.0 if you run:

```
etcdctl endpoint health
```
**_NOTE: no command line flags are provided_** 

it will fail with

```
[...]

PR #20120, opened 2025-06-05
Update 'endpoint health' command to avoid calling `flags.SetPflagsFromEnv` multiple times

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/20118

Note both `etcdctl endpoint health`  and `etcdctl endpoint health --cluster` have this issue.

`flags.SetPflagsFromEnv` should only be called once.  It has already been called in `clientConfigFromCmd`, so there is no need to call it again in `epHealthCommandFunc`.

Also reverted https://github.com/etcd-io/etcd/pull/18742 to fix the `etcdctl endpoint health --cluster` issue


ahrtr:

Let's get the fix included in v3.6.1

fuweid (approved):

Issue #18707, opened 2024-10-09
Document how to debug performance of prow job


serathius:
### What would you like to be added?

In robustness tests we noticed a series of flakes and we wanted to eliminate performance as a factor. @ah8ad3 has reached out to Kubernetes prow and found interesting dashboards:
https://monitoring-gke.prow.k8s.io/d/96Q8oOOZk/builds?orgId=1&refresh=30s&var-org=etcd-io&var-repo=etcd&var-job=ci-etcd-robustness-main-amd64&var-build=All&from=now-7d&to=now

We should preserve this knowledge and document for future how etcd contributors can navigate performance of prow jobs. Possibly a new section in etcd contributor documentation about prow that either links to the dashboard or some prow documentation.

### Why is this needed?

Make it easier for others to debug etcd job performance issues.

serathius:
cc @ah8ad3

ivanvc:
I agree. There's also an EKS dashboard, but they are not documented anywhere in the infra testing repo, I also found about it by asking on Slack

CFly17:
I am interested in learning more about and contributing to etcd!

Are you thinking of a new doc (like `etcd/Documentation/contributor-guide/prow_jobs.md`) with sections like:

- Introduction to Prow
- How Prow is Used for etcd Testing
- Creating/Navigating Performance Dashboard (Grafana)
- Interpreting Metrics (link to [prow documentation](https://github.com/kubernetes-sigs/prow/blob/main/site/content/en/docs/metrics/_index.md))
- Troubleshooting Common Issues (e.g.?)
- etc.

Please let me know if I can help! :)

ivanvc:
Hi, @CFly17, thanks for volunteering!

> Are you thinking of a new doc (like etcd/Documentation/contributor-guide/prow_jobs.md) with sections like:

Yes, we need a new document, and your suggestion sounds good. I think we can start with:

* Introduction to Prow 
* How Prow is Used for etcd Testing - And where the jobs are located, i.e., test-infra repository, in the `config/jobs/etcd directory`
* Navigating Performance Dashboard (Grafana) (just navigating. I don't think we have an intention to create dashboards at this moment
* Selecting our jobs, explaining the different kinds of workloads/jobs, i.e., robustness, integration, static check, etc. (at a high level).
* Interpreting Metrics

[...]

PR #20132, opened 2025-06-06
Add documentation for Prow jobs

joshjms:
Ref: https://github.com/etcd-io/etcd/issues/18707

/cc @ivanvc 

### DIFF
--- a/Documentation/contributor-guide/prow.md
+++ b/Documentation/contributor-guide/prow.md
@@ -1,0 +1,77 @@
+# etcd Prow Documentation
+
+Prow is a Kubernetes based CI/CD system developed to serve the Kubernetes community. 
+-- https://github.com/kubernetes-sigs/prow
+
+etcd uses Prow to perform various CI tests to maintain the guarantees that etcd promises.
+
+## List of Workflows
+
+### Presubmits
+
+Presubmits are run after a PR is created, given an `ok-to-test`, and before the PR is merged.
+
+#### `main`
+
+| Name | Link | Description |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


serathius:
List of every dashboard in prow will quickly make it outdated.

joshjms:
> List of every dashboard in prow will quickly make it outdated.

Hmm, that's true. Should I just make a description for each tests we have (e.g. instead of test-robustness-amd64 and test-robustness-arm64, we can just have them under "Robustness Tests" and a brief description of what they do)?

serathius:
> Hmm, that's true. Should I just make a description for each tests we have (e.g. instead of test-robustness-amd64 and test-robustness-arm64, we can just have them under "Robustness Tests" and a brief description of what they do)?

Sounds good, however for robustness tests we already have documentation https://github.com/etcd-io/etcd/blob/main/tests/robustness/README.md. Maybe instead of writing another one, just link to it. We don't have documentation for other types of tests, mostly because it's spread around the codebase, like unit tests. E2e and integration tests have their own directories so maybe we could add documentation there.

I think documenting types of tests, their goals and their method would be great.

joshjms:
Right, definitely we can link to it. I'll write the documentation for the tests as well as how do we use Prow in the same documentation then. I'll write any questions I have here :)

ivanvc:
Hi, @joshjms. Please also take a look at my comment on the parent issue (https://github.com/etcd-io/etcd/issues/18707#issuecomment-2434174108). I believe the proposed breakdown it's a bit different from what this pull request has. Thanks.

joshjms:
> Hi, @joshjms. Please also take a look at my comment on the parent issue (https://github.com/etcd-io/etcd/issues/18707#issuecomment-2434174108). I believe the proposed breakdown it's a bit different from what this pull request has. Thanks.

Ah okay noted, I was thinking of a good way to provide a high level overview for the Prow jobs that we have. I think making brief descriptions and linking it to their respective documentations (if any) would be good. Will work on the other sections too of course.

PR #20139, opened 2025-06-07
Remove v2 store

serathius:
/cc @ahrtr @fuweid

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -46,8 +46,7 @@ type RaftCluster struct {
 	localID types.ID
 	cid     types.ID
 
-	v2store v2store.Store
-	be      MembershipBackend
+	be MembershipBackend
 
 	sync.Mutex // guards the fields below
 	version    *semver.Version
@@ -116,7 +115,6 @@ func NewCluster(lg *zap.Logger, opts ...ClusterOption) *RaftCluster {
 		removed:       make(map[types.ID]bool),
 		downgradeInfo: &serverversion.DowngradeInfo{Enabled: false},
 		maxLearners:   clOpts.maxLearners,
-		v2store:       v2store.New(),
 	}
 }
 
@@ -246,8 +244,6 @@ func (c *RaftCluster) SetID(localID, cid types.ID) {
 	c.buildMembershipMetric()
 }
 
-func (c *RaftCluster) SetStore(st v2store.Store) { c.v2store = st }
-
 func (c *RaftCluster) SetBackend(be MembershipBackend) {
 	c.be = be
 	c.be.MustCreateBackendBuckets()
@@ -307,14 +303,11 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
 // ValidateConfigurationChange takes a proposed ConfChange and
 // ensures that it is still valid.
 func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
-	var membersMap map[types.ID]*Member
-	var removedMap map[types.ID]bool
-
-	if shouldApplyV3 {
-		membersMap, removedMap = c.be.MustReadMembersFromBackend()
-	} else {
-		membersMap, removedMap = membersFromStore(c.lg, c.v2store)
+	if !shouldApplyV3 {
+		return nil
 	}
+
+	membersMap, removedMap := c.be.MustReadMembersFromBackend()
 
 	id := types.ID(cc.NodeID)
 	if removedMap[id] {
@@ -400,7 +393,6 @@ func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldAp
 func (c *RaftCluster) AddMember(m *Member, shouldApplyV3 ShouldApplyV3) {
 	c.Lock()
 	defer c.Unlock()
-	mustSaveMemberToStore(c.lg, c.v2store, m)
 
 	if m.ID == c.localID {
 		setIsLearnerMetric(m)
@@ -436,7 +428,6 @@ func (c *RaftCluster) AddMember(m *Member, shouldApplyV3 ShouldApplyV3) {
 func (c *RaftCluster) RemoveMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	c.Lock()
 	defer c.Unlock()
-	mustDeleteMemberFromStore(c.lg, c.v2store, id)
 	if shouldApplyV3 {
 		c.be.MustDeleteMemberFromBackend(id)
 
@@ -478,7 +469,6 @@ func (c *RaftCluster) UpdateAttributes(id types.ID, attr Attributes, shouldApply
 
 	if m, ok := c.members[id]; ok {
 		m.Attributes = attr
-		mustUpdateMemberAttrInStore(c.lg, c.v2store, m)
 		if shouldApplyV3 {
 			c.be.MustSaveMemberToBackend(m)
 		}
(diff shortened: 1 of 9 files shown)


ahrtr:
I think before we remove v2store, we need to be clear how to generate a valid v2snapshot when downgrading from 3.7 to 3.6.

ahrtr:
I have been working on https://github.com/etcd-io/etcd/issues/20187 in the past two days, I realized that this PR can greatly simplify the code base, and also simplify the task [20187](https://github.com/etcd-io/etcd/issues/20187).

@serathius do you have bandwidth to finish this PR this week? It would be great to get this done soon so we can unblock https://github.com/etcd-io/etcd/issues/20187. If you don't have bandwidth, I am happy to take over.

serathius:
We need to rewrite the tests first to support v3 storage, https://github.com/etcd-io/etcd/pull/21092 would be a prerequisite.

serathius:
Getting test failure on unit test. Looks harder to debug, might not have time to fix it. 

Possibly something with bootstrap as it shows different output when applying entries second time.
```
=== RUN   TestApplyConfStateWithRestart
    /src/go.etcd.io/etcd/server/etcdserver/server_test.go:314: Applying entries for the first time
    /src/go.etcd.io/etcd/server/etcdserver/server_test.go:320: Reapplying same entries after restart
    /src/go.etcd.io/etcd/server/etcdserver/server_test.go:324: actions don't match:   []testutil.Action{
                {Name: "ApplyConfChange", Params: {raftpb.ConfChange{NodeID: 1, Context: `{"id":1,"peerURLs":[""]}`}}},
                {
                        Name: "ApplyConfChange",
                        Params: []any{
                                raftpb.ConfChange{
                                        Type:    s"ConfChangeRemoveNode",
        -                               NodeID:  0,
        +                               NodeID:  1,
                                        Context: nil,
                                        ID:      0,
                                },
                        },
                },
                {Name: "ApplyConfChange", Params: {raftpb.ConfChange{Type: s"ConfChangeUpdateNode", Context: `{"id":1,"peerURLs":[""]}`}}},
          }
--- FAIL: TestApplyConfStateWithRestart (0.02s)
FAIL
FAIL    go.etcd.io/etcd/server/v3/etcdserver    0.053s
```

server/etcdserver/api/membership/cluster.go
```diff
@@ -312,8 +308,6 @@ func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldAp
 
 	if shouldApplyV3 {
 		membersMap, removedMap = c.be.MustReadMembersFromBackend()
-	} else {
-		membersMap, removedMap = membersFromStore(c.lg, c.v2store)
```
ahrtr, line 316: This is the root cause that causes the unit test failure.

When you reapply the entries, `shouldApplyV3` is false, so it doesn't load memberMap and removedMap at all. So it fails to apply the second `ConfChangeRemoveNode`, so node is set to None (0).


Proposal, we don't call `ValidateConfigurationChange` at all if `shouldApplyV3` is false. We already removed v2store, we don't need to worry any inconsistence between v3store and v2store.

ahrtr:
> Getting test failure on unit test. Looks harder to debug, might not have time to fix it.
> 
> Possibly something with bootstrap as it shows different output when applying entries second time.
> 
> ```
> === RUN   TestApplyConfStateWithRestart
>     /src/go.etcd.io/etcd/server/etcdserver/server_test.go:314: Applying entries for the first time
>     /src/go.etcd.io/etcd/server/etcdserver/server_test.go:320: Reapplying same entries after restart
>     /src/go.etcd.io/etcd/server/etcdserver/server_test.go:324: actions don't match:   []testutil.Action{
>                 {Name: "ApplyConfChange", Params: {raftpb.ConfChange{NodeID: 1, Context: `{"id":1,"peerURLs":[""]}`}}},
>                 {
>                         Name: "ApplyConfChange",
>                         Params: []any{
>                                 raftpb.ConfChange{
>                                         Type:    s"ConfChangeRemoveNode",
>         -                               NodeID:  0,
>         +                               NodeID:  1,
>                                         Context: nil,
>                                         ID:      0,
>                                 },
>                         },
>                 },
>                 {Name: "ApplyConfChange", Params: {raftpb.ConfChange{Type: s"ConfChangeUpdateNode", Context: `{"id":1,"peerURLs":[""]}`}}},
>           }
> --- FAIL: TestApplyConfStateWithRestart (0.02s)
> FAIL
> FAIL    go.etcd.io/etcd/server/v3/etcdserver    0.053s
> ```

Pls refer to https://github.com/etcd-io/etcd/pull/20139#discussion_r2671782536?

ahrtr:
> Getting test failure on unit test. Looks harder to debug, might not have time to fix it.

Do you want me to fix it now?

serathius:
> Do you want me to fix it now?

Sent updated file PTAL

ahrtr:
We need to update the test case.

When applying the same entries, it should have a different `want`, because etcd won't validate any entries, so the last `ConfChangeUpdateNode` still has nodeID 1.

```
	want := []testutil.Action{
		{
			Name: "ApplyConfChange",
			Params: []any{raftpb.ConfChange{
				Type:    raftpb.ConfChangeAddNode,
				NodeID:  nodeID,
				Context: memberData,
			}},
		},
		{
			Name: "ApplyConfChange",
			Params: []any{raftpb.ConfChange{
				Type:   raftpb.ConfChangeRemoveNode,
				NodeID: nodeID,
			}},
		},
		// This action is expected to fail validation, thus NodeID is set to 0
		{
			Name: "ApplyConfChange",
			Params: []any{raftpb.ConfChange{
				Type:    raftpb.ConfChangeUpdateNode,
				Context: memberData,
				NodeID:  1,
			}},
		},
	}
```

serathius:
> When applying the same entries, it should have a different want,

The test is build on idea that replaying same entries gives same response. Don't know if test makes sense anymore.

ahrtr:
> The test is build on idea that replaying same entries gives same response. Don't know if test makes sense anymore.

I think we can update the goal of the test case to verify the different behaviour of `ValidateConfigurationChange` (shouldApplyV3 or not).

server/etcdserver/server_test.go
```diff
@@ -322,8 +315,15 @@ func TestApplyConfStateWithRestart(t *testing.T) {
...
-	srv.cluster.SetStore(v2store.New())
+	t.Log("Simulating etcd restart by clearing v3 store")
+	be, _ := betesting.NewDefaultTmpBackend(t)
+	t.Cleanup(func() {
+		betesting.Close(t, be)
+	})
+	lg := zaptest.NewLogger(t)
+	srv.cluster.SetBackend(schema.NewMembershipBackend(lg, be))
+	srv.beHooks = serverstorage.NewBackendHooks(lg, srv.consistIndex)
+	srv.consistIndex.SetBackend(be)
```
ahrtr, line 326: The previous case for v2store makes sense, because etcd might not generate v2 snapshot at all before restarting.

But It doesn't make sense for v3store. If data in v3store gets lost after restart, it's a disaster for etcd. etcd won't bootstrap successfully in e2e test.

For this case, we just need to define two different `want` for two applies as mentioned in https://github.com/etcd-io/etcd/pull/20139#issuecomment-3723331402

serathius: Etcd flushes v3 store every 100ms, if etcd crashed before flushing data it would result in exactly same scenario

ahrtr: > Etcd flushes v3 store every 100ms

The v3store may just lose 100ms data, but now you clear (recreate) the v3store. When you reapply the same entries, essentially it works exactly the same as when applying the first time because the v3store is completely new created. The `ValidateConfigurationChange` run with exactly the same workflow in both applying (shouldApplyV3 always true). So the change loses the capability to verify different behaviour of `ValidateConfigurationChange`

ahrtr: > but now you clear (recreate) the v3store.

To be clearer, normally the case would never happen in production environment. If it happens, then it needs human intervention or orchestration platform to fix it.

So It doesn't make sense to clear v3store before restarting etcd. I don't want to block this PR too long. If you don't have bandwidth to take care of it for now, I am OK to merge it first.

ahrtr: Note that in theory the original implementation (clearing v2store) doesn't make sense either. But in practice usually there isn't any v2 snapshot being generated at all (snapshot-count is 10K by default) in a test case, so restarting etcd will result in an empty v2store.

server/etcdserver/api/membership/cluster.go
```diff
@@ -307,15 +303,12 @@ func (c *RaftCluster) Recover(onSet func(*zap.Logger, *semver.Version)) {
...
 func (c *RaftCluster) ValidateConfigurationChange(cc raftpb.ConfChange, shouldApplyV3 ShouldApplyV3) error {
-	var membersMap map[types.ID]*Member
-	var removedMap map[types.ID]bool
-
-	if shouldApplyV3 {
-		membersMap, removedMap = c.be.MustReadMembersFromBackend()
-	} else {
-		membersMap, removedMap = membersFromStore(c.lg, c.v2store)
+	if !shouldApplyV3 {
+		return nil
```
ahrtr, line 307: This creates a hole that on bootstrap etcd may blindly apply some confChange to raft. I need to think this through in https://github.com/etcd-io/etcd/issues/20187

Issue #20165, opened 2025-06-12
etcd stopped accepts new connections on port 2380 after a network temporary error


zwtop:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via [etcd maintainers](mailto:etcd-maintainers@googlegroups.com).
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

etcd stopped accepts new connections on port 2380 after a network temporary error.

### What did you expect to happen?

etcd should always accepts new connections on port 2380 after a network temporary error, or log an error and exit.

### How can we reproduce it (as minimally and precisely as possible)?

1. Run etcd with a small nofile limit
```
$ prlimit --nofile=160:160 -- ./etcd-v3.6.1-linux-amd64/etcd --peer-auto-tls --listen-peer-urls=https://localhost:2380
```

2. Generate a large number of connections to port 2380
```
$ cat > main.go << EOF
package main

import "net"

func main() {
        for n := 0; n < 200; n++ {
                net.Dial("tcp", "127.0.0.1:2380")
        }
}
EOF

$ go run main.go
```

After the above steps, etcd stopped accepts new connections on port 2380.

The Recv-Q on port 2380 would never handled by etcd.

```
$ netstat -ltn4
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
tcp        0      0 127.0.0.1:2379          0.0.0.0:*               LISTEN
tcp       53      0 127.0.0.1:2380          0.0.0.0:*               LISTEN
```

### Anything else we need to know?

I think the problem happens here:

1. `tlsListener` return on any accept error:
https://github.com/etcd-io/etcd/blob/release-3.6/client/pkg/transport/listener_tls.go#L126-L130

2. `cMux` continue accept on the `Temporary` error:
https://github.com/soheilhy/cmux/blob/v0.1.5/cmux.go#L248-L250

[...]

PR #20167, opened 2025-06-12
client: donot return on tlsListener receive net temporary error

zwtop:
Fix etcd-io/etcd#20165.

### DIFF
--- a/client/pkg/transport/listener_tls.go
+++ b/client/pkg/transport/listener_tls.go
@@ -125,6 +125,9 @@ func (l *tlsListener) acceptLoop() {
 	for {
 		conn, err := l.Listener.Accept()
 		if err != nil {
+			if ne, ok := err.(net.Error); ok && ne.Temporary() {
+				continue
+			}
 			l.err = err
 			return
 		}


ahrtr:
Is it possible to add an unit test to verify the `acceptLoop` still works under temporary error?

zwtop:
> Is it possible to add an unit test to verify the `acceptLoop` still works under temporary error?

It's hard to generate network temporary errors in a unit test.
Maybe I can use gomonkey to generate an error, or write a fake listener can return a temporary error.

client/pkg/transport/listener_tls.go
```diff
@@ -125,6 +125,9 @@ func (l *tlsListener) acceptLoop() {
 	for {
 		conn, err := l.Listener.Accept()
 		if err != nil {
+			if ne, ok := err.(net.Error); ok && ne.Temporary() {
+				continue
+			}
```
ahrtr, line 130: The `Temporary` has already been deprecated. It seems that you should use `Timeout()` instead

https://github.com/golang/go/blob/53af292aed21c3f6ea75d16e5b25f660b2c028fb/src/net/net.go#L424-L433

zwtop: Yes. This fix plan seems problematic because it assumes specific caller behavior.
We should address the root cause: The `tlsListener`'s error handling design unconditionally exits and returns the latest error, which callers cannot handle this behavior. 
I think the `tlsListener` should keep `acceptLoop` until `Close` is called, or return a specific error on exit.

I will close this PR, and maybe open a new one later.

Issue #20126, opened 2025-06-05
Bump Go to 1.24.4 / 1.23.10


ivanvc:
### What would you like to be added?

Golang 1.24.4 and 1.23.10 were released today. According to our [Dependency management documentation](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/dependency_management.md#golang-versions), we want to stay on the latest patch version. This means updating the `release-3.4`, `release-3.5`, and `release-3.6` branches to 1.23.10 and `main` to 1.24.4.

This patch includes the security fixes for CVE-2025-4673, CVE-2025-0913, and CVE-2025-22874.

Progress track:

- [x] [etcd](https://github.com/etcd-io/etcd)
  - [x] [`main`](https://github.com/etcd-io/etcd/tree/main): go v1.24.4 - #20163
  - [x] [`release-3.6`](https://github.com/etcd-io/etcd/tree/release-3.6): go v1.23.10 - #20128
  - [x] [`release-3.5`](https://github.com/etcd-io/etcd/tree/release-3.5): go v1.23.10 - #20164
  - [x] [`release-3.4`](https://github.com/etcd-io/etcd/tree/release-3.4): go v1.23.10 - #20166
  - [x] [`CHANGELOG`](https://github.com/etcd-io/etcd/tree/main/CHANGELOG)
    - [x] 3.6 - #20134
    - [x] 3.4, 3.5 - #20174
- [x] [bbolt](https://github.com/etcd-io/bbolt)
  - [x] [`main`](https://github.com/etcd-io/bbolt/tree/main): go v1.24.4 - etcd-io/bbolt#983
  - [x] [`release-1.3`](https://github.com/etcd-io/bbolt/tree/release-1.3): v1.23.10 - etcd-io/bbolt#984
  - [x] [`release-1.4`](https://github.com/etcd-io/bbolt/tree/release-1.4): v1.23.10 - etcd-io/bbolt#985
- [x] [raft](https://github.com/etcd-io/raft)
  - [x] [`main`](https://github.com/etcd-io/raft/tree/main): go v1.24.4 - etcd-io/raft#304
  - [x] [`release-3.6`](https://github.com/etcd-io/raft/tree/release-3.6): go v1.23.10 - etcd-io/raft#305
- [x] [`etcd-io/gofail` `main`](https://github.com/etcd-io/gofail/tree/main): go v1.23.10 - etcd-io/gofail#116
[...]

PR #20168, opened 2025-06-12
[release-3.5] Add test-release makefile target 

abdurrehman107:
This PR adds `test-release` makefile target for release-3.5. 

ref: [kubernetes/test-infra#32754](https://github.com/kubernetes/test-infra/issues/32754#issuecomment-2942746295)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,6 +179,10 @@ test-e2e:
 test-e2e-release:
 	PASSES="build release e2e" ./test.sh $(GO_TEST_FLAGS)
 
+.PHONY: test-release
+test-release:
+	PASSES="release_tests" ./test.sh $(GO_TEST_FLAGS)
+
 ensure-docker-test-image-exists:
 	make pull-docker-test || ( echo "WARNING: Container Image not found in registry, building locally"; make build-docker-test )
 
--- a/test.sh
+++ b/test.sh
@@ -687,6 +687,34 @@ function release_pass {
   mv /tmp/etcd ./bin/etcd-last-release
 }
 
+function release_tests_pass {
+  if [ -z "${VERSION:-}" ]; then
+    VERSION=$(go list -m go.etcd.io/etcd/api/v3 2>/dev/null | \
+     awk '{split(substr($2,2), a, "."); print a[1]"."a[2]".99"}')
+  fi
+
+  if [ -n "${CI:-}" ]; then
+    git config user.email "prow@etcd.io"
+    git config user.name "Prow"
+
+    gpg --batch --gen-key <<EOF
+%no-protection
+Key-Type: 1
+Key-Length: 2048
+Subkey-Type: 1
+Subkey-Length: 2048
+Name-Real: Prow
+Name-Email: prow@etcd.io
+Expire-Date: 0
+EOF
+
+    git remote add origin https://github.com/etcd-io/etcd.git
+  fi
+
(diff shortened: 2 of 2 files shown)


joshjms:
I think the vulnerability check failure is related to [this](https://github.com/etcd-io/etcd/issues/20126). Let's wait for the go version bump. 

Edit: watching https://github.com/etcd-io/etcd/pull/20164

ivanvc:
Thanks for the pull request, @abdurrehman107. Please rebase your branch due to #20164.

Makefile
```diff
@@ -179,6 +179,10 @@ test-e2e:
 test-e2e-release:
 	PASSES="build release e2e" ./test.sh $(GO_TEST_FLAGS)
 
+.PHONY: test-release
+test-release:
+	PASSES="release_tests" VERSION="3.7.99" ./test.sh $(GO_TEST_FLAGS)
```
ivanvc: This doesn't work locally because of this environment variable. We need to drop it.

```suggestion
	PASSES="release_tests" ./test.sh $(GO_TEST_FLAGS)
```

ivanvc (commented):
Thanks for the pull request, @abdurrehman107. Please review my comment. Also, I'd like to suggest running the make targets you're adding locally to ensure that they work (as we haven't enabled them in the CI yet).

joshjms:
The integration test [results](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/20168/pull-etcd-integration-1-cpu-amd64/1934864552973832192) show a race condition. In the Github Actions workflow, we run the integration test with the race flag disabled. 

```
linux-integration-1-cpu)
  make install-gofail
  GOARCH=${{ inputs.arch }} CPU=1 RACE='false' FAILPOINTS='true' make test-integration
  ;;
```

However, in the Prow job, we also check for the race conditions. Would it be better to disable the race flag for release-3.5 to keep the original behavior?

/cc @serathius @jmhbnz

serathius:
Failure is in `./contrib/raftexample`, `test-integration` should not run those tests. All integration tests are located in `./tests/integration`

ivanvc:
@serathius, do you suggest, then, deleting this function altogether?

https://github.com/etcd-io/etcd/blob/75357ec9f18ca8b464f75b1e470f7a89c6e0b0f5/test.sh#L106-L113

Which is called from `integration_pass`:

https://github.com/etcd-io/etcd/blob/75357ec9f18ca8b464f75b1e470f7a89c6e0b0f5/test.sh#L115-L119

serathius:
> @serathius, do you suggest, then, deleting this function altogether?

Hmm, we dropped running raftexample in integration testing in https://github.com/etcd-io/etcd/pull/14881
I think the reasoning was that we were moving raft to separate repo and we expected that raftexample will be moved there too. Unfortunately that didn't happen due to lack of contributions.  

I think it's ok to remove it, because we don't test it on release-3.6 nor main branch and we don't have plan to fix race condition in raftexample.

ahrtr (approved):

ahrtr:
Once https://github.com/etcd-io/etcd/pull/20198 gets merged, please rebase this PR, thx

ivanvc:
@abdurrehman107, could you please rebase this pull request? Thanks.

ivanvc (approved):
LGTM. Thanks, @abdurrehman107.

abdurrehman107:
rebased. Thank you guys. 
cc @ivanvc

Issue #19893, opened 2025-05-08
[Antithesis] Refactor recording client management


serathius:
### What would you like to be added?

Current recording client management has a problem. It's easy to create a client and then forget to collect report from it. 

Would be better if we had some higher level structure for creating clients that collects all clients and is able to return report for all clients constructed.

 

### Why is this needed?

Make it harder to forget collecting report from client.

nwnt:
Not sure if you want to discuss this in the weekly robustness call, but I can have a look at this in the meantime too.

serathius:
Goal is to replace something like:
```
reports := []report.ClientReport{}
var mux sync.Mutex
var wg sync.WaitGroup
for i:=0;i<6;i++{
  c := client.NewRecordingClient(hosts[i%len(hosts)], ids, baseTime)
  defer c.Close()
  wg.Add(1)
  go func() {
    defer wg.Done()
    traffic.RunTraffic(c...)
    mux.Lock()
    reports = append(reports, c.Report())
    mux.Unlock()
  }()
}
wg.Wait()
return reports
```

With
```
clientSet := client.NewSet(ids, baseTime)
defer clientSet.Close()
var wg sync.WaitGroup
for i:=0;i<6;i++{
  c := clientSet.NewClient(hosts[i%len(hosts)])
  wg.Add(1)
  go func() {
    defer wg.Done()
    traffic.RunTraffic(c...)
  }()
}
wg.Wait()
return clientSet.Reports()
```

serathius:
So:
* Add a ClientSet struct that manages livecycle of clients
* Move common arguments like ids, baseTime to ClientSet
* All clients created are recorded within clientSet, allowing as to collect reports via clientset for all clients created by it.
* [optional] Allow for closing clients directly or via clientset.
* [optional] Safeguard collecting reports before all clients are closed. Prevent Creation of new clients after clientset is closed. 

Note, RecordingClient is used around the code, but we don't need to refactor everything at once. Let's do it in smaller steps.

PR #20173, opened 2025-06-14
Refactor and add connect all client

### LINKED ISSUE #20052: [Antithesis+Robustness] Add a client that connects to all members

Currently we configure clients to connect to a single etcd member. This ensures that member will always receive traffic after it goes down. If clients were created to all of etcd they would migrate out of it and not get back due to stickiness. 

Still I think it would be good to have 1 client that loadbalances. We can split `ClientCount` available into `MemberClientCount` and `ClusterClientCount`, where member connects to single member and cluster to all members in cluster.

/cc @joshjms @henrybear327 @nwnt 

### Why is this needed?

Reason is finding bugs in retry logic in cases where request failed to first member and then loadbalancer moved it to other member for a retry.

nwnt:
Fix #20052

This also contains quite extensive refactoring for `simulateTraffic`.


tests/robustness/traffic/traffic.go
```diff
@@ -43,14 +43,14 @@ var (
...
-		ClientCount:                    8,
+		MemberClientCount:              8,
 		MaxNonUniqueRequestConcurrency: 3,
 	}
 	HighTrafficProfile = Profile{
 		MinimalQPS:                     100,
 		MaximalQPS:                     1000,
 		BurstableQPS:                   1000,
-		ClientCount:                    8,
+		MemberClientCount:              8,
```
serathius: Can you add a one cluster client here too?

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -137,58 +138,119 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
 	return reports
 }
 
+type simulator struct {
+	hosts              []string
+	tf                 traffic.Traffic
+	reporting          reporter
+	ids                identity.Provider
+	baseTime           time.Time
+	finish             <-chan struct{}
```
serathius: You should watch out when creating struct based on collection of argument variables, not all variables are equal. 
Don't think `finish` make sense here. It's more like context, and should be passed like it.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -137,58 +138,119 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
+	storage            identity.LeaseIDStorage
+	limiter            *rate.Limiter
+	concurrencyLimiter traffic.ConcurrencyLimiter
+	keyStore           *traffic.KeyStore
+}
+
+func (s *simulator) runSingleMemberTraffic(ctx context.Context) {
+	for i := range profile.MemberClientCount {
+		c := connect([]string{s.hosts[i%len(s.hosts)]}, s.ids, s.baseTime)
+		go s.reporting.makeReportOf(trafficLoad(ctx, s.tf, s.limiter, s.ids, s.storage, s.concurrencyLimiter, s.keyStore, s.finish)).usingClient(c)
```
serathius: This one line is overloaded. Shorter doesn't mean better,

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -137,58 +138,119 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
+	}
+}
+
+func (s *simulator) runAllMemberTraffic(ctx context.Context) {
```
serathius: There is no reason to have separate function for running all member traffic vs single member traffic. It's almost the same code structure. Separating things like this implies that they are less related causing potential confusion in reader.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -43,7 +43,8 @@ var (
 		MinimalQPS:                     100,
 		MaximalQPS:                     1000,
 		BurstableQPS:                   1000,
-		ClientCount:                    3,
+		MemberClientCount:              3,
```
serathius, line 46: I don't recommend mixing refactors and feature changes. Can you send a separate PR?

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -137,58 +138,119 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
+
+func (s *simulator) runDefragment(ctx context.Context) {
+	defragPeriod := traffic.DefaultCompactionPeriod * time.Duration(len(s.hosts))
+	for _, h := range s.hosts {
+		c := connect([]string{h}, s.ids, s.baseTime)
+		go s.reporting.makeReportOf(defragment(ctx, defragPeriod, s.finish)).usingClient(c)
+	}
+}
+
+type reporter struct {
```
serathius: This struct is confusing for me, I don't understand what's doing. Can you instead adopt more https://github.com/etcd-io/etcd/issues/19893 like approach?

tests/robustness/traffic/traffic.go
serathius: Maybe 6 per member and 2 Cluster.

tests/antithesis/test-template/robustness/traffic/main.go
nwnt: With #19893 merged, I reverted the changes here and added the client.  Hopefully this looks better now.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -43,7 +43,8 @@ var (
 		MinimalQPS:                     100,
 		MaximalQPS:                     1000,
 		BurstableQPS:                   1000,
-		ClientCount:                    3,
+		MemberClientCount:              3,
+		ClusterClientCount:             3,
```
serathius: Please just have 1 cluster client. Doubling the number of clients is an overkill.

nwnt: Adjusted to 1.

serathius (approved):

PR #20175, opened 2025-06-15
Refactor recording client management

### LINKED ISSUE #19893: [Antithesis] Refactor recording client management

nwnt:
Fix #19893

### DIFF
--- a/tests/antithesis/test-template/robustness/traffic/main.go
+++ b/tests/antithesis/test-template/robustness/traffic/main.go
@@ -93,12 +93,12 @@ func main() {
 
 func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts []string, baseTime time.Time, duration time.Duration) ([]report.ClientReport, error) {
 	ids := identity.NewIDProvider()
-	r, err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, ids, baseTime)
+	trafficSet := client.NewSet(ids, baseTime)
+	defer trafficSet.Close()
+	err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, trafficSet)
 	if err != nil {
 		lg.Fatal("Failed empty database at start check", zap.Error(err))
 	}
-	trafficReports := []report.ClientReport{r}
-	watchReport := []report.ClientReport{}
 	maxRevisionChan := make(chan int64, 1)
 	watchConfig := client.WatchConfig{
 		RequestProgress: true,
@@ -107,22 +107,23 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
 	startTime := time.Since(baseTime)
 	g.Go(func() error {
 		defer close(maxRevisionChan)
-		trafficReports = slices.Concat(trafficReports, simulateTraffic(ctx, tf, hosts, ids, baseTime, duration))
-		maxRevision := report.OperationsMaxRevision(trafficReports)
+		simulateTraffic(ctx, tf, hosts, trafficSet, duration)
+		maxRevision := report.OperationsMaxRevision(trafficSet.Reports())
 		maxRevisionChan <- maxRevision
 		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
 		return nil
 	})
+	watchSet := client.NewSet(ids, baseTime)
+	defer watchSet.Close()
 	g.Go(func() error {
-		var err error
-		watchReport, err = client.CollectClusterWatchEvents(ctx, lg, hosts, maxRevisionChan, watchConfig, baseTime, ids)
+		err := client.CollectClusterWatchEvents(ctx, lg, hosts, maxRevisionChan, watchConfig, watchSet)
 		return err
 	})
 	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 	endTime := time.Since(baseTime)
-	reports := slices.Concat(trafficReports, watchReport)
+	reports := slices.Concat(trafficSet.Reports(), watchSet.Reports())
 	totalStats := traffic.CalculateStats(reports, startTime, endTime)
 	lg.Info("Completed traffic generation",
 		zap.Int("successes", totalStats.Successes),
@@ -134,59 +135,46 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
 	return reports, nil
 }
 
-func simulateTraffic(ctx context.Context, tf traffic.Traffic, hosts []string, ids identity.Provider, baseTime time.Time, duration time.Duration) []report.ClientReport {
-	var mux sync.Mutex
+func simulateTraffic(ctx context.Context, tf traffic.Traffic, hosts []string, clientSet *client.ClientSet, duration time.Duration) {
 	var wg sync.WaitGroup
 	storage := identity.NewLeaseIDStorage()
 	limiter := rate.NewLimiter(rate.Limit(profile.MaximalQPS), profile.BurstableQPS)
 	concurrencyLimiter := traffic.NewConcurrencyLimiter(profile.MaxNonUniqueRequestConcurrency)
 	finish := closeAfter(ctx, duration)
-	reports := []report.ClientReport{}
 	keyStore := traffic.NewKeyStore(10, "key")
-	for i := 0; i < profile.ClientCount; i++ {
-		c := connect([]string{hosts[i%len(hosts)]}, ids, baseTime)
+	for i := range profile.ClientCount {
+		c := connect(clientSet, []string{hosts[i%len(hosts)]})
 		wg.Add(1)
 		go func(c *client.RecordingClient) {
 			defer wg.Done()
 			defer c.Close()
-
 			tf.RunTrafficLoop(ctx, c, limiter,
-				ids,
+				clientSet.IdentityProvider(),
 				storage,
 				concurrencyLimiter,
 				keyStore,
 				finish,
 			)
-			mux.Lock()
-			reports = append(reports, c.Report())
-			mux.Unlock()
 		}(c)
 	}
 	wg.Add(1)
-	compactClient := connect(hosts, ids, baseTime)
+	compactClient := connect(clientSet, hosts)
 	go func(c *client.RecordingClient) {
 		defer wg.Done()
 		defer c.Close()
 		tf.RunCompactLoop(ctx, c, traffic.DefaultCompactionPeriod, finish)
-		mux.Lock()
-		reports = append(reports, c.Report())
-		mux.Unlock()
 	}(compactClient)
 	defragPeriod := traffic.DefaultCompactionPeriod * time.Duration(len(hosts))
 	for _, h := range hosts {
-		c := connect([]string{h}, ids, baseTime)
+		c := connect(clientSet, []string{h})
 		wg.Add(1)
 		go func(c *client.RecordingClient) {
 			defer wg.Done()
 			defer c.Close()
 			runDefragLoop(ctx, c, defragPeriod, finish)
-			mux.Lock()
-			reports = append(reports, c.Report())
-			mux.Unlock()
 		}(c)
 	}
 	wg.Wait()
-	return reports
 }
 
 func runDefragLoop(ctx context.Context, c *client.RecordingClient, period time.Duration, finish <-chan struct{}) {
@@ -210,8 +198,8 @@ func runDefragLoop(ctx context.Context, c *client.RecordingClient, period time.D
 	}
 }
 
-func connect(endpoints []string, ids identity.Provider, baseTime time.Time) *client.RecordingClient {
-	cli, err := client.NewRecordingClient(endpoints, ids, baseTime)
+func connect(cs *client.ClientSet, endpoints []string) *client.RecordingClient {
+	cli, err := cs.NewClient(endpoints)
 	if err != nil {
 		// Antithesis Assertion: client should always be able to connect to an etcd host
 		assert.Unreachable("Client failed to connect to an etcd host", map[string]any{"endpoints": endpoints, "error": err})
--- a/tests/robustness/client/client.go
+++ b/tests/robustness/client/client.go
@@ -16,6 +16,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -370,4 +371,82 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
 		panic(fmt.Sprintf("Unexpected event type: %s", event.Type))
 	}
 	return watch
+}
+
+type ClientSet struct {
+	idProvider identity.Provider
+	baseTime   time.Time
+
+	mux     sync.Mutex
+	closed  bool
+	clients []*RecordingClient
+	reports []report.ClientReport
+}
+
+func NewSet(ids identity.Provider, baseTime time.Time) *ClientSet {
+	return &ClientSet{
+		idProvider: ids,
+		baseTime:   baseTime,
+
+		clients: []*RecordingClient{},
+	}
+}
+
+func (cs *ClientSet) NewClient(endpoints []string) (*RecordingClient, error) {
+	cs.mux.Lock()
+	defer cs.mux.Unlock()
+	if cs.closed {
+		return nil, errors.New("the clientset is already closed")
+	}
+	cli, err := NewRecordingClient(endpoints, cs.idProvider, cs.baseTime)
+	if err != nil {
+		return nil, err
+	}
+	cs.clients = append(cs.clients, cli)
+	return cli, nil
+}
+
+func (cs *ClientSet) Reports() []report.ClientReport {
+	cs.mux.Lock()
+	defer cs.mux.Unlock()
+	if !cs.closed {
+		cs.close()
+	}
+	if cs.reports == nil {
+		reports := cs.generateReports()
+		cs.reports = reports
+	}
+	return cs.reports
+}
+
+func (cs *ClientSet) Close() {
+	cs.mux.Lock()
+	defer cs.mux.Unlock()
+	cs.close()
+}
+
+func (cs *ClientSet) close() {
+	if cs.closed {
+		return
+	}
+	for _, c := range cs.clients {
+		c.Close()
+	}
+	cs.closed = true
+}
+
+func (cs *ClientSet) generateReports() []report.ClientReport {
+	reports := make([]report.ClientReport, 0, len(cs.clients))
+	for _, c := range cs.clients {
+		reports = append(reports, c.Report())
+	}
+	return reports
+}
+
+func (cs *ClientSet) IdentityProvider() identity.Provider {
+	return cs.idProvider
+}
+
+func (cs *ClientSet) BaseTime() time.Time {
+	return cs.baseTime
 }
--- a/tests/robustness/main_test.go
+++ b/tests/robustness/main_test.go
@@ -20,6 +20,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 	"time"
@@ -128,7 +129,7 @@ func runScenario(ctx context.Context, t *testing.T, s scenarios.TestScenario, lg
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	g := errgroup.Group{}
-	var operationReport, watchReport, failpointClientReport []report.ClientReport
+	var failpointClientReport []report.ClientReport
 	failpointInjected := make(chan report.FailpointInjection, 1)
 
 	// using baseTime time-measuring operation to get monotonic clock reading
@@ -152,19 +153,22 @@ func runScenario(ctx context.Context, t *testing.T, s scenarios.TestScenario, lg
 		}
 		return nil
 	})
+	trafficSet := client.NewSet(ids, baseTime)
+	defer trafficSet.Close()
 	maxRevisionChan := make(chan int64, 1)
 	g.Go(func() error {
 		defer close(maxRevisionChan)
-		operationReport = traffic.SimulateTraffic(ctx, t, lg, clus, s.Profile, s.Traffic, failpointInjected, baseTime, ids)
+		operationReport := traffic.SimulateTraffic(ctx, t, lg, clus, s.Profile, s.Traffic, failpointInjected, trafficSet)
 		maxRevision := report.OperationsMaxRevision(operationReport)
 		maxRevisionChan <- maxRevision
 		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
 		return nil
 	})
+	watchSet := client.NewSet(ids, baseTime)
+	defer watchSet.Close()
 	g.Go(func() error {
-		var err error
 		endpoints := processEndpoints(clus)
-		watchReport, err = client.CollectClusterWatchEvents(ctx, lg, endpoints, maxRevisionChan, s.Watch, baseTime, ids)
+		err := client.CollectClusterWatchEvents(ctx, lg, endpoints, maxRevisionChan, s.Watch, watchSet)
 		return err
 	})
 	err := g.Wait()
@@ -176,7 +180,7 @@ func runScenario(ctx context.Context, t *testing.T, s scenarios.TestScenario, lg
 	if err != nil {
 		t.Error(err)
 	}
-	return append(operationReport, append(failpointClientReport, watchReport...)...)
+	return slices.Concat(trafficSet.Reports(), watchSet.Reports(), failpointClientReport)
 }
 
 func randomizeTime(base time.Duration, jitter time.Duration) time.Duration {
--- a/tests/robustness/traffic/traffic.go
+++ b/tests/robustness/traffic/traffic.go
@@ -55,17 +55,15 @@ var (
 	}
 )
 
-func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, profile Profile, traffic Traffic, failpointInjected <-chan report.FailpointInjection, baseTime time.Time, ids identity.Provider) []report.ClientReport {
-	mux := sync.Mutex{}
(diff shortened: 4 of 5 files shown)


tests/robustness/client/client.go
```diff
@@ -371,3 +372,49 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+	cli, err := NewRecordingClient(endpoints, cs.idProvider, cs.baseTime)
+	if err != nil {
+		return nil, err
+	}
+	cs.clients = append(cs.clients, cli)
+	return cli, nil
+}
+
+func (cs *ClientSet) Reports() []report.ClientReport {
+	cs.reports = make([]report.ClientReport, 0, len(cs.clients))
```
serathius: Why this sets reports but not uses it?

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -135,58 +135,45 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
 }
 
 func simulateTraffic(ctx context.Context, tf traffic.Traffic, hosts []string, ids identity.Provider, baseTime time.Time, duration time.Duration) []report.ClientReport {
```
serathius: Use it in robustness tests too?

tests/robustness/client/client.go
nwnt: This currently collects the reports from all the clients and returns it - kinda following what's proposed on the issue.  The returned client reports are going to be combined with the watch and database empty check reports in `runTraffic`.  Do you expect the reports to be used differently than just being returned?  If so, let me know and I can add in.

tests/antithesis/test-template/robustness/traffic/main.go
nwnt: I can start looking at it.  Do you want that to be in a separate PR instead of this one?

nwnt: never mind, it's actually simper than I thought...

serathius: Right, I would want to avoid one asymmetrically applying improvements to framework. It would make the codebase harder to maintain. As long as change is not huge we should apply it to both.

tests/robustness/traffic/traffic.go
```diff
@@ -73,25 +72,26 @@ func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
 
 	keyStore := NewKeyStore(10, "key")
 
+	clientSet := client.NewSet(ids, baseTime)
```
serathius: Can you also pass the clientSet to `CheckEmptyDatabaseAtStart` above

tests/robustness/client/client.go
serathius: I'm saying that saving to variable cs.report will just leak memory and duplicate reports upon second call. 

The intention of saving reports initially was to have 2 different modes of clientSet: 
* `open`, that allows users to create new clients using `NewClient`, and make requests. When all traffic is generated user should then generate reports
* `closed`, all clients are closed so they cannot make requests and change data in reports. User can only generate reports using `Reports`. As reports should not change we can compute reports once and cache.

This is however not how the current implementation of `Reports` works as it's not combined with `open`.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,49 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
 	}
 	return watch
 }
+
+type ClientSet struct {
+	idProvider identity.Provider
+	baseTime   time.Time
+
+	open    bool
```
serathius: `closed` is better than `open` variable here.

tests/robustness/client/client.go
nwnt: Ok, I made `Reports()` close the client set if it's not closed and return the reports.  The reports are collected when `Close()` is called.  Let me know if this is good.

tests/robustness/traffic/traffic.go
nwnt: Done.  However, it doesn't look as clean on the antithesis side.   I put a TODO comment there as I think further refactoring there might be a bit complicated due to the early collection of the reports for max revision used by the watch operation.  I will have a look at that later and have a follow up PR for it.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,56 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+}
+
+func (cs *ClientSet) Reports() []report.ClientReport {
+	if !cs.closed {
+		cs.Close()
+	}
+	return cs.reports
+}
+
+func (cs *ClientSet) Close() {
```
serathius, line 413: This is incorrect, usually close is called after `Reports` as `Close` is deferred.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,56 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+	if cs.closed {
+		return
+	}
+	cs.closed = true
+	cs.reports = make([]report.ClientReport, 0, len(cs.clients))
```
serathius: It is incorrect to calculate reports only on `Close` as it's always called after `Reports`. Reports is called within function while Close is deferred. 

Solution: Reports should also close.

nwnt: that's already the case, right?  In `Reports` I have:

```go
func (cs *ClientSet) Reports() []report.ClientReport {
	if !cs.closed {
		cs.Close()
	}
```

serathius: Ups, missed that. nvm

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -93,11 +93,13 @@ func main() {
...
 func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts []string, baseTime time.Time, duration time.Duration) ([]report.ClientReport, error) {
 	ids := identity.NewIDProvider()
-	r, err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, ids, baseTime)
+	clientSet := client.NewSet(ids, baseTime)
+	err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, clientSet)
 	if err != nil {
 		lg.Fatal("Failed empty database at start check", zap.Error(err))
 	}
-	trafficReports := []report.ClientReport{r}
+	// TODO use clientSet in all operations requiring c client creation
```
serathius: Can you implement this?

tests/robustness/client/client.go
```diff
@@ -371,3 +372,56 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+	for _, c := range cs.clients {
+		cs.reports = append(cs.reports, c.Report())
+		c.Close()
+	}
+}
+
+func (cs *ClientSet) IdentityProvider() identity.Provider {
```
serathius, line 446: Is this method needed? Don't see any calls to IdentityProvider

tests/robustness/client/client.go
```diff
@@ -371,3 +372,56 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+}
+
+func (cs *ClientSet) Close() {
+	if cs.closed {
+		return
+	}
+	cs.closed = true
+	cs.reports = make([]report.ClientReport, 0, len(cs.clients))
+	for _, c := range cs.clients {
+		cs.reports = append(cs.reports, c.Report())
```
serathius: First close, and then collect report.

tests/robustness/client/client.go
serathius: Maybe what was unclear for me that you called `Close` from reports directly. For readability would be nice to have a separate function called `generateReports`.

For example: 
```
func (cs *ClientSet) Reports() []report.ClientReport {
	if !cs.closed {
		cs.Close()
	}
	if cs.reports == nil {
		cs.reports = cs.generateReports()
	}
	return cs.reports
}

func (cs *ClientSet) Close() {
	if cs.closed {
		return
	}
	for _, c := range cs.clients {
		err := c.Close()
	}
	cs.closed = true
}

func (cs *ClientSet) generateReports() []report.ClientReport {
	reports : = make([]report.ClientReport, 0, len(cs.clients))
	for _, c := range cs.clients {
		reports = append(cs.reports, c.Report())
	}
	return reports
}
```

serathius: This should make the code cleared because it would keep Closing focused on closing, while Report on generating report.

nwnt: Done.

tests/antithesis/test-template/robustness/traffic/main.go
nwnt: I had to introduce another struct called `MultClientSet` on top of `ClientSet`.  The reason for that is because there's a cross-channel dependency between the traffic simulation and watch operation where the latter needs the max revision from the traffic simulation.  Hope that's ok.

tests/robustness/client/client.go
nwnt: Now it's needed so that we can just pass in the clientset to the functions that also use the identity provider.

nwnt:
@serathius sorry was away for a few days.  How does the latest commit look?

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -93,12 +92,11 @@ func main() {
 
 func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts []string, baseTime time.Time, duration time.Duration) ([]report.ClientReport, error) {
 	ids := identity.NewIDProvider()
-	r, err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, ids, baseTime)
+	multiClientSet := client.NewMultiSet(ids, baseTime)
```
serathius: Why multi set? I don't follow the reason we need a multiple sets, I should suffice.

tests/robustness/client/client.go
serathius: OK

tests/antithesis/test-template/robustness/traffic/main.go
nwnt: During traffic simulation, the reports have to be collected early to find the max revision, which will then be fed to the channel for the watch operation.  If we have only one set, collecting the report would also close the set, correct?
```go 
		trafficReports = slices.Concat(trafficReports, simulateTraffic(ctx, tf, hosts, ids, baseTime, duration))
		maxRevision := report.OperationsMaxRevision(trafficReports)
		maxRevisionChan <- maxRevision
		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
```

serathius: Ok, that makes sense. The problem is that we want to first finish traffic generation and get the report, but still wait for watch requests. Using a single ClientSet would result in traffic generation closing the set and impacting watch.

serathius: While MultiSet tackles the generic issue, I think we should avoid over-designing here. 
Could you just have one Set for traffic + CheckEmptyDatabaseAtStart and have a separate set for watch?

nwnt: Done!

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +133,45 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
-			defer c.Close()
 			runDefragLoop(ctx, c, defragPeriod, finish)
-			mux.Lock()
-			reports = append(reports, c.Report())
-			mux.Unlock()
 		}(c)
 	}
 	wg.Wait()
-	return reports
+	return clientSet.Reports()
```
serathius: Nit: We no longer need to return the reports.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,73 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+
+type ClientSet struct {
+	mux        sync.Mutex
+	idProvider identity.Provider
+	baseTime   time.Time
+
+	closed  bool
+	clients []*RecordingClient
+	reports []report.ClientReport
+}
```
serathius:
```suggestion
type ClientSet struct {
	idProvider identity.Provider
	baseTime   time.Time

	mux        sync.Mutex
	closed  bool
	clients []*RecordingClient
	reports []report.ClientReport
}
```

It good to have mutex over the variables it's meant to protect. We don't need to protect idProvider or baseTime as they are never changed.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,73 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+	return &ClientSet{
+		idProvider: ids,
+		baseTime:   baseTime,
+
+		clients: []*RecordingClient{},
+	}
+}
+
+func (cs *ClientSet) NewClient(endpoints []string) (*RecordingClient, error) {
+	if cs.closed {
```
serathius, line 398: Checking `cs.closed` should be under mutex.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,73 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+		return nil, err
+	}
+	cs.mux.Lock()
+	cs.clients = append(cs.clients, cli)
+	cs.mux.Unlock()
+	return cli, nil
+}
+
+func (cs *ClientSet) Reports() []report.ClientReport {
+	if !cs.closed {
```
serathius: Checking closed and and closing should be under mutex.

tests/robustness/client/client.go
```diff
@@ -371,3 +372,73 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+		cs.Close()
+	}
+	if cs.reports == nil {
+		reports := cs.generateReports()
+		cs.reports = reports
+	}
+	return cs.reports
+}
+
+func (cs *ClientSet) Close() {
```
serathius, line 422: Should be under mutex

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -93,12 +93,11 @@ func main() {
 
 func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts []string, baseTime time.Time, duration time.Duration) ([]report.ClientReport, error) {
 	ids := identity.NewIDProvider()
-	r, err := traffic.CheckEmptyDatabaseAtStart(ctx, lg, hosts, ids, baseTime)
+	trafficSet := client.NewSet(ids, baseTime)
```
serathius, line 96: You should always defer closing clientSet.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -107,22 +106,22 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
 	g.Go(func() error {
 		defer close(maxRevisionChan)
-		trafficReports = slices.Concat(trafficReports, simulateTraffic(ctx, tf, hosts, ids, baseTime, duration))
+		trafficReports := simulateTraffic(ctx, tf, hosts, trafficSet, duration)
 		maxRevision := report.OperationsMaxRevision(trafficReports)
 		maxRevisionChan <- maxRevision
 		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
 		return nil
 	})
+	watchSet := client.NewSet(ids, baseTime)
```
serathius, line 116: ditto

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +133,45 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
-		mux.Unlock()
 	}(compactClient)
 	defragPeriod := traffic.DefaultCompactionPeriod * time.Duration(len(hosts))
 	for _, h := range hosts {
-		c := connect([]string{h}, ids, baseTime)
+		c := connect(clientSet, []string{h})
 		wg.Add(1)
 		go func(c *client.RecordingClient) {
 			defer wg.Done()
-			defer c.Close()
```
serathius, line 181: Don't remove close.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +133,45 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
-	reports := []report.ClientReport{}
 	keyStore := traffic.NewKeyStore(10, "key")
-	for i := 0; i < profile.ClientCount; i++ {
-		c := connect([]string{hosts[i%len(hosts)]}, ids, baseTime)
+	for i := range profile.ClientCount {
+		c := connect(clientSet, []string{hosts[i%len(hosts)]})
 		wg.Add(1)
 		go func(c *client.RecordingClient) {
 			defer wg.Done()
-			defer c.Close()
```
serathius, line 151: Don't remove close.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -210,8 +195,8 @@ func runDefragLoop(ctx context.Context, c *client.RecordingClient, period time.D
 	}
 }
 
-func connect(endpoints []string, ids identity.Provider, baseTime time.Time) *client.RecordingClient {
```
serathius, line 213: For followup, we should remove this function.

tests/robustness/main_test.go
```diff
@@ -152,19 +153,20 @@ func runScenario(ctx context.Context, t *testing.T, s scenarios.TestScenario, lg
 		}
 		return nil
 	})
+	trafficSet := client.NewSet(ids, baseTime)
```
serathius, line 156: ditto

tests/robustness/main_test.go
```diff
@@ -152,19 +153,20 @@ func runScenario(ctx context.Context, t *testing.T, s scenarios.TestScenario, lg
...
 	g.Go(func() error {
 		defer close(maxRevisionChan)
-		operationReport = traffic.SimulateTraffic(ctx, t, lg, clus, s.Profile, s.Traffic, failpointInjected, baseTime, ids)
+		operationReport := traffic.SimulateTraffic(ctx, t, lg, clus, s.Profile, s.Traffic, failpointInjected, trafficSet)
 		maxRevision := report.OperationsMaxRevision(operationReport)
 		maxRevisionChan <- maxRevision
 		lg.Info("Finished simulating Traffic", zap.Int64("max-revision", maxRevision))
 		return nil
 	})
+	watchSet := client.NewSet(ids, baseTime)
```
serathius, line 167: ditto

tests/robustness/traffic/traffic.go
```diff
@@ -55,17 +55,16 @@ var (
...
-	mux := sync.Mutex{}
+func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, profile Profile, traffic Traffic, failpointInjected <-chan report.FailpointInjection, clientSet *client.ClientSet) []report.ClientReport {
 	endpoints := clus.EndpointsGRPC()
 
 	lm := identity.NewLeaseIDStorage()
 	// Use the highest MaximalQPS of all traffic profiles as burst otherwise actual traffic may be accidentally limited
 	limiter := rate.NewLimiter(rate.Limit(profile.MaximalQPS), profile.BurstableQPS)
 
-	r, err := CheckEmptyDatabaseAtStart(ctx, lg, endpoints, ids, baseTime)
+	defer clientSet.Close()
```
serathius: Defer close should be called next to the construction.

tests/robustness/traffic/traffic.go
```diff
@@ -121,16 +113,15 @@ func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
...
 	lg.Info("Finished traffic")
-	endTime := time.Since(baseTime)
+	endTime := time.Since(clientSet.BaseTime())
 
 	time.Sleep(time.Second)
 	// Ensure that last operation succeeds
-	cc, err := client.NewRecordingClient(endpoints, ids, baseTime)
+	cc, err := clientSet.NewClient(endpoints)
 	require.NoError(t, err)
-	defer cc.Close()
```
serathius, line 130: Don't remove the close.

tests/robustness/traffic/traffic.go
nwnt: I have removed this one instead since it's now constructed outside of this func

tests/robustness/client/client.go
nwnt: Instead of this check, I made the method always call `Close`.  The same check is done there.  If we leave the function here and cover it with mutex, it will run into a deadlock from acquiring lock at `Reports` and acquiring lock again at `Close`.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +135,46 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
 	return reports, nil
 }
 
-func simulateTraffic(ctx context.Context, tf traffic.Traffic, hosts []string, ids identity.Provider, baseTime time.Time, duration time.Duration) []report.ClientReport {
-	var mux sync.Mutex
+func simulateTraffic(ctx context.Context, tf traffic.Traffic, hosts []string, clientSet *client.ClientSet, duration time.Duration) {
 	var wg sync.WaitGroup
+	defer clientSet.Close()
```
serathius: simulateTraffic function didn't create the clientSet thus it should not be responsible for closing it.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +135,46 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
 	concurrencyLimiter := traffic.NewConcurrencyLimiter(profile.MaxNonUniqueRequestConcurrency)
 	finish := closeAfter(ctx, duration)
-	reports := []report.ClientReport{}
 	keyStore := traffic.NewKeyStore(10, "key")
-	for i := 0; i < profile.ClientCount; i++ {
-		c := connect([]string{hosts[i%len(hosts)]}, ids, baseTime)
+	for i := range profile.ClientCount {
+		c := connect(clientSet, []string{hosts[i%len(hosts)]})
 		wg.Add(1)
+		defer c.Close()
```
serathius: Defer was inside the goroutine before because we don't want to defer in a loop.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +135,46 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
-			reports = append(reports, c.Report())
-			mux.Unlock()
 		}(c)
 	}
 	wg.Add(1)
-	compactClient := connect(hosts, ids, baseTime)
+	compactClient := connect(clientSet, hosts)
 	go func(c *client.RecordingClient) {
 		defer wg.Done()
-		defer c.Close()
```
serathius, line 169: Restore defer here.

tests/antithesis/test-template/robustness/traffic/main.go
```diff
@@ -134,59 +135,46 @@ func runTraffic(ctx context.Context, lg *zap.Logger, tf traffic.Traffic, hosts [
...
-		mux.Lock()
-		reports = append(reports, c.Report())
-		mux.Unlock()
 	}(compactClient)
 	defragPeriod := traffic.DefaultCompactionPeriod * time.Duration(len(hosts))
 	for _, h := range hosts {
-		c := connect([]string{h}, ids, baseTime)
+		c := connect(clientSet, []string{h})
 		wg.Add(1)
+		defer c.Close()
```
serathius: Defer in the goroutine

tests/robustness/client/client.go
```diff
@@ -371,3 +372,75 @@ func toWatchEvent(event clientv3.Event) (watch model.WatchEvent) {
...
+	cli, err := NewRecordingClient(endpoints, cs.idProvider, cs.baseTime)
+	if err != nil {
+		return nil, err
+	}
+	cs.clients = append(cs.clients, cli)
+	return cli, nil
+}
+
+func (cs *ClientSet) Reports() []report.ClientReport {
+	cs.Close()
```
serathius: You could avoid using mutex twice by having 2 functions `Close` that takes mutex and calls `close` that just closes assuming that mutex is already taken. That would allow `Reports` to take mutex once can just call `close`.

tests/robustness/traffic/traffic.go
```diff
@@ -74,40 +72,33 @@ func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
...
+	startTime := time.Since(clientSet.BaseTime())
+	for i := range profile.ClientCount {
 		wg.Add(1)
-		c, nerr := client.NewRecordingClient([]string{endpoints[i%len(endpoints)]}, ids, baseTime)
+
+		c, nerr := clientSet.NewClient([]string{endpoints[i%len(endpoints)]})
 		require.NoError(t, nerr)
 		go func(c *client.RecordingClient) {
 			defer wg.Done()
-			defer c.Close()
```
serathius, line 84: Keep the defer here.

tests/robustness/traffic/traffic.go
```diff
@@ -74,40 +72,33 @@ func SimulateTraffic(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2
...
 	if !profile.ForbidCompaction {
 		wg.Add(1)
-		c, nerr := client.NewRecordingClient(endpoints, ids, baseTime)
+		c, nerr := clientSet.NewClient(endpoints)
 		if nerr != nil {
 			t.Fatal(nerr)
 		}
 		go func(c *client.RecordingClient) {
 			defer wg.Done()
-			defer c.Close()
```
serathius: Keep the defer here.

nwnt:
This should fix all of them.  Can I bother you to review again @serathius?

serathius (approved):

PR #20177, opened 2025-06-15
etcdctl: refactor json printing to support custom writers

kstrifonoff:
A necessary "pre-change" for #19469.

The `jsonPrinter` struct now includes an `io.Writer` field, enabling unit testing of this module.


ahrtr (approved):
LGTM & thx

PR #20196, opened 2025-06-20
Split model types into types.go file

serathius:
/cc @joshjms @nwnt @fuweid @siyuanfoundation


serathius:
> do we normally put constants in the same file as the type definitions or do we put them in a separate file like consts.go?

I usually prefer to be flexible based on size and relation within code. I don't think single const value justifies a whole `consts.go` file. 

We don't have project wide guidelines about package structure. I think we should do what makes sense locally.

serathius:
PTAL @fuweid @siyuanfoundation @ahrtr

fuweid (approved):

PR #20219, opened 2025-06-25
[Antithesis] Fix building the client image

joshjms:
Ref: https://github.com/etcd-io/etcd/pull/20216 https://github.com/etcd-io/etcd/issues/20112

/cc @serathius 

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

PR #20226, opened 2025-06-26
Fix working dir for building antithesis client image

serathius:
Missed different working dir when reviewing https://github.com/etcd-io/etcd/pull/20219

/cc @joshjms @ahrtr @fuweid @siyuanfoundation @nwnt


joshjms:
Oh no I missed that too. Thanks for that.

PR #20227, opened 2025-06-26
refactor: move out common channels in contrib/raftexample

### LINKED ISSUE #20213: Circular Initialization of kv store in contrib/raftexample

In `contrib/raftexample/main.go` there is definition of `kvstore` and `raftNode`: 

	// raft provides a commit stream for the proposals from the http api
	var kvs *kvstore
	getSnapshot := func() ([]byte, error) { return kvs.getSnapshot() }
	commitC, errorC, snapshotterReady := newRaftNode(*id, strings.Split(*cluster, ","), *join, getSnapshot, proposeC, confChangeC)

	kvs = newKVStore(<-snapshotterReady, proposeC, commitC, errorC)

	// the key-value http handler will propose updates to raft
	serveHTTPKVAPI(kvs, *kvport, confChangeC, errorC)

I see some circularity in these definitions: we use `getSnapshot` which is defined via (empty) `kvs` store in order to create `raftNode` and to use this raft node in order to initialize `kvs` store. It does not seem to be good design.

I don't have the solution ATM since I'm just getting into codebase, but I would like to start some discussion on that one.

Side question: Wouldn't it be better to rename `kvstore` to `kvStore`? 

### Why is this needed?

Separation of concerns. Code readability would be improved.

DeamonDev:
This closes #20213 

I moved out common channels in order to avoid circularity in bootstrap process mentioned in issue.


serathius:
Please sign DCO https://github.com/etcd-io/etcd/pull/20227/checks?check_run_id=44830463262

DeamonDev:
Ok, I've signed the commit.

DeamonDev:
So @serathius can we proceed it further?

moficodes:
It looks like we are swapping one problematic code for another. Now we are doing side effects to set the values passed to the `newRaftNode`. I am not sure if this is easier to understand or maintain than the original.

Issue #20221, opened 2025-06-25
Watch on future revision might receive old events or notifications, happens when client connects to multiple members.


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Just yesterday we merged https://github.com/etcd-io/etcd/pull/20173 that added to both Robustness tests and Antithesis tests new type of client, one that connects to all members.

Within 1 day both frameworks found a new issue with `watch: broke Bookmarkable - Progress notification events guarantee that all events up to a revision have been already delivered`. Failures:
* Antithesis: https://linuxfoundation.antithesis.com/report/UZjUP_KGxboJepL7k1q_8pa4/ZqL0Vt9a7YESiiBmGecPMkBP8YgM1IwlTZJ4dcYjmZ8.html?auth=v2.public.eyJuYmYiOiIyMDI1LTA2LTI1VDAzOjE4OjIzLjM4MDU2MDQwMFoiLCJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoiWnFMMFZ0OWE3WUVTaWlCbUdlY1BNa0JQOFlnTTFJd2xUWko0ZGNZam1aOC5odG1sIiwicmVwb3J0X2lkIjoiVVpqVVBfS0d4Ym9KZXBMN2sxcV84cGE0In19feIAsYO4-UIigcL4eMu7QUqA6XFbCU3Hnw7BeyZW06o9x11mFqleHbSbRWdIcLdTH2Xzx42DXNB7dBqYq25Ujg4
* https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-etcd-robustness-main-amd64/1937584192116232192

Example request for revision 148 returns event with revision 67.
```
{
  "Request": {
    "Key": "/registry/pods/",
    "Revision": 148,
    "WithPrefix": true,
    "WithProgressNotify": true,
    "WithPrevKV": true
  },
  "Responses": [
    {
[...]

PR #20225, opened 2025-06-25
Add verification to verify the watch response have a bigger revision than minRev

ahrtr:
Link to https://github.com/etcd-io/etcd/issues/20221

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

cc @fuweid @serathius

PR #20229, opened 2025-06-26
Skip sending progress notification for watch with starting revision in the future

serathius:
Ref https://github.com/etcd-io/etcd/issues/20221

### DIFF
--- a/server/etcdserver/api/v3rpc/watch.go
+++ b/server/etcdserver/api/v3rpc/watch.go
@@ -305,12 +305,7 @@ func (sws *serverWatchStream) recvLoop() error {
 
 			filters := FiltersFromRequest(creq)
 
-			wsrev := sws.watchStream.Rev()
-			rev := creq.StartRevision
-			if rev == 0 {
-				rev = wsrev + 1
-			}
-			id, err := sws.watchStream.Watch(mvcc.WatchID(creq.WatchId), creq.Key, creq.RangeEnd, rev, filters...)
+			id, err := sws.watchStream.Watch(mvcc.WatchID(creq.WatchId), creq.Key, creq.RangeEnd, creq.StartRevision, filters...)
 			if err == nil {
 				sws.mu.Lock()
 				if creq.ProgressNotify {
@@ -328,7 +323,7 @@ func (sws *serverWatchStream) recvLoop() error {
 			}
 
 			wr := &pb.WatchResponse{
-				Header:   sws.newResponseHeader(wsrev),
+				Header:   sws.newResponseHeader(sws.watchStream.Rev()),
 				WatchId:  int64(id),
 				Created:  true,
 				Canceled: err != nil,
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -122,12 +122,13 @@ func (s *watchableStore) NewWatchStream() WatchStream {
 
 func (s *watchableStore) watch(key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc) {
 	wa := &watcher{
-		key:    key,
-		end:    end,
-		minRev: startRev,
-		id:     id,
-		ch:     ch,
-		fcs:    fcs,
+		key:      key,
+		end:      end,
+		startRev: startRev,
+		minRev:   startRev,
+		id:       id,
+		ch:       ch,
+		fcs:      fcs,
 	}
 
 	s.mu.Lock()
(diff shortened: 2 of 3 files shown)


server/storage/mvcc/watchable_store.go
```diff
@@ -572,6 +577,7 @@ type watcher struct {
 	// except when the watcher were to be moved from "synced" watcher group
 	restore bool
 
+	startRev int64
```
serathius, line 580: Needed to add a separate field for progress notifications, this is because minRev points to next revision which is cannot distinguish case of starting_rev=current_rev and starting_rev=current_rev+1.

For example if etcd is on rev 2:
* Watch opened on revision 2 will have minRev equal to 3. Next event will have rev 3, accepts progress notification with rev 2.
* Watch opened on revision 3 will have minRev equal to 3. Next event also will have rev 3, but should not get progress notification rev 2.

server/storage/mvcc/watchable_store.go
```diff
@@ -122,12 +122,13 @@ func (s *watchableStore) NewWatchStream() WatchStream {
...
 	wa := &watcher{
-		key:    key,
-		end:    end,
-		minRev: startRev,
-		id:     id,
-		ch:     ch,
-		fcs:    fcs,
+		key:      key,
+		end:      end,
+		startRev: startRev,
```
ahrtr, line 127: The `startRev` may be 0, in that case etcd will automatically assign `currentRev + 1`, which is what we need to save to `startRev` in such case.

server/storage/mvcc/watchable_store.go
ahrtr: > because minRev points to next revision

It's exactly the reason why https://github.com/etcd-io/etcd/pull/20225 fails.

server/storage/mvcc/watchable_store.go
serathius: In that case the check:
```
		if rev < w.startRev {
			return false
		}
```
Will be always false, so things will work.

For simplicity I'm thinking to keep the startRev from the request, not one that is resolved based on local rev.

server/storage/mvcc/watchable_store.go
```diff
@@ -532,11 +533,15 @@ func (s *watchableStore) progressIfSync(watchers map[WatchID]*watcher, responseW
...
 
+	rev := s.rev()
 	// Any watcher unsynced?
 	for _, w := range watchers {
 		if _, ok := s.synced.watchers[w]; !ok {
 			return false
 		}
+		if rev < w.startRev {
+			return false
+		}
```
ahrtr, line 544: We only need to skip this specific watcher, but the implementation skips all watchers?

server/storage/mvcc/watchable_store.go
serathius: Tests integration failed, so looks like  I was wrong. Will double check.

server/storage/mvcc/watchable_store.go
ahrtr: It should be OK, since all watchers share the same underlying stream

serathius: Yes, this is because there are 2 types of progress notifications:
* For specific WatchID. 
* For all watches in stream (WatchID=-1). 

In the second case progress notification will be received by all watchers in the stream, so either all of them get progress notification or none of them. Even if 1 watcher is unsynced, or in this case, in the future, we cannot send progress notification.

This is why progress notification in etcd are not guaranteed, this is a long standing issue, however not prioritized because Kubernetes with watch cache establishes only one watch per stream.

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -305,12 +305,7 @@ func (sws *serverWatchStream) recvLoop() error {
 
 			filters := FiltersFromRequest(creq)
 
-			wsrev := sws.watchStream.Rev()
-			rev := creq.StartRevision
-			if rev == 0 {
-				rev = wsrev + 1
```
serathius, line 311: This is weird, we have this logic duplicated in watchable_store.

server/storage/mvcc/watcher_test.go
```diff
@@ -289,48 +289,102 @@ func TestWatchStreamCancelWatcherByID(t *testing.T) {
...
+						t.Fatalf("got %+v, expect %+v", resp, wrs)
+					}
+				} else {
+					t.Fatalf("unexpected response %+v", resp)
+				}
+			default:
+				if tc.expectProgressBeforeSync {
+					t.Fatalf("failed to receive progress")
+				}
+			}
```
ahrtr: It's identical to line 371-386. Can we have a small function or method to be reused?

serathius: Right

serathius:
Verified with running robustness tests locally for over 1 hours with no reproductions, while without this fix issue occurs usually before 5 minutes of testing.

ahrtr (approved):
LGTM & thx

serathius:
Ok, let's hunt for the second issue that breaks `Resumable`.

ahrtr:
reopen to track the changelog effort, pls add a changelog, thx

PR #20238, opened 2025-06-27
chore(release): fix release script for MacOS

Elbehery:
Follow up on https://github.com/etcd-io/bbolt/pull/1009/files#r2172475625

cc @ahrtr @ivanvc

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -153,7 +153,11 @@ main() {
       TARGET_VERSION="v${VERSION}" update_versions_cmd
 
       log_callout "Updating version from ${source_version} to ${VERSION} in api/version/version.go"
-      sed -i "s/${source_version}/${VERSION}/g" api/version/version.go
+      if [[ "$OSTYPE" == "darwin"* ]]; then
+         sed -i '' "s/${source_version}/${VERSION}/g" api/version/version.go
+      else
+         sed -i "s/${source_version}/${VERSION}/g" api/version/version.go
+      fi
     fi
 
 


scripts/release.sh
```diff
@@ -153,7 +153,11 @@ main() {
       TARGET_VERSION="v${VERSION}" update_versions_cmd
 
       log_callout "Updating version from ${source_version} to ${VERSION} in api/version/version.go"
-      sed -i "s/${source_version}/${VERSION}/g" api/version/version.go
+      if [[ "$OSTYPE" == "darwin"* ]]; then
```
jmhbnz, line 156: We only release etcd from linux machines and I don't believe there is any intent to change that.  Can you please provide some context on why this pr is required?

Elbehery: hello ✋🏽 

the issue was raised in https://github.com/etcd-io/bbolt/pull/1009#discussion_r2172475625 

thats why i raised this PR

jmhbnz: Ok thanks for this proposal.  Currently our release procedures for etcd state: https://github.com/etcd-io/etcd/blob/33a458068d27ba9da2c38898277b16a891757f5a/Documentation/contributor-guide/release.md?plain=1#L59-L62

Unless we are making a change to this procedure to add macOS as a supported release platform I would prefer not to proceed with this pr which adds more complexity to our already complex release scripting.

I would prefer we invest effort in further automation to execute the release script through automated pipelines in a declaritive environment i.e. Linux AMD64.

ivanvc: I agree, the complexity of releasing etcd is greater than bbolt. To fully support macOS, we may need to do additional work. We should drop this pull request. Thanks for the initiative, @Elbehery.

Issue #20137, opened 2025-06-07
[Robustness] Test growing cluster from 1 to 3 members


serathius:
### What would you like to be added?

Might need two version:
* Adding normal members
* Adding a members and promoting them.

/cc @nwnt @henrybear327 @joshjms

### Why is this needed?

Test cluster membership changes

nwnt:
We're still working on this.

PR #20239, opened 2025-06-28
Add growing member test scenario

nwnt:
For #20137

### DIFF
--- a/tests/robustness/failpoint/cluster.go
+++ b/tests/robustness/failpoint/cluster.go
@@ -40,6 +40,9 @@ var (
 	MemberReplace          Failpoint = memberReplace{}
 	MemberDowngrade        Failpoint = memberDowngrade{}
 	MemberDowngradeUpgrade Failpoint = memberDowngradeUpgrade{}
+	ClusterScaleOutAndIn   Failpoint = &scalingMember{
+		scalingTarget: 2,
+	}
 )
 
 type memberReplace struct{}
@@ -303,4 +306,40 @@ func patchArgs(args []string, flag, newValue string) error {
 		}
 	}
 	return fmt.Errorf("--%s flag not found", flag)
+}
+
+type scalingMember struct {
+	scalingTarget int
+}
+
+func (f *scalingMember) Inject(ctx context.Context, t *testing.T, lg *zap.Logger, clus *e2e.EtcdProcessCluster, baseTime time.Time, ids identity.Provider) ([]report.ClientReport, error) {
+	// scaling in
+	cfgs := make([]*e2e.EtcdServerProcessConfig, 0, f.scalingTarget)
+	for i := range f.scalingTarget {
+		proc := clus.Procs[i]
+		cfgs = append(cfgs, proc.Config())
+		err := clus.CloseProc(ctx, func(p e2e.EtcdProcess) bool {
+			return p.Config().ClientURL == proc.Config().ClientURL
+		})
+		if err != nil {
+			return nil, err
+		}
+		time.Sleep(etcdserver.HealthInterval)
+	}
+	for i := range f.scalingTarget {
+		err := clus.StartNewProcFromConfig(ctx, t, cfgs[i])
+		if err != nil {
+			return nil, err
+		}
+		time.Sleep(etcdserver.HealthInterval)
+	}
+	return nil, nil
+}
+
+func (f *scalingMember) Name() string {
+	return "ScalingMember"
+}
+
+func (f *scalingMember) Available(config e2e.EtcdProcessClusterConfig, member e2e.EtcdProcess, profile traffic.Profile) bool {
+	return config.ClusterSize >= 3
 }
--- a/tests/robustness/scenarios/scenarios.go
+++ b/tests/robustness/scenarios/scenarios.go
@@ -109,12 +109,14 @@ func Exploratory(_ *testing.T) []TestScenario {
 		name := filepath.Join(tp.Name, "ClusterOfSize1")
 		clusterOfSize1Options := baseOptions
 		clusterOfSize1Options = append(clusterOfSize1Options, e2e.WithClusterSize(1))
-		scenarios = append(scenarios, TestScenario{
-			Name:    name,
-			Traffic: tp.Traffic,
-			Profile: tp.Profile,
-			Cluster: *e2e.NewConfig(clusterOfSize1Options...),
-		})
+		scenarios = append(scenarios,
+			TestScenario{
+				Name:    name,
+				Traffic: tp.Traffic,
+				Profile: tp.Profile,
+				Cluster: *e2e.NewConfig(clusterOfSize1Options...),
+			},
+		)
 	}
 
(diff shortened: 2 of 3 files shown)


nwnt:
@serathius Not sure if this is the same as what you have in mind.  I probably missed a lot of things so definitely would love your feedback about it.

tests/robustness/failpoint/cluster.go
```diff
@@ -40,6 +40,10 @@ var (
 	MemberReplace          Failpoint = memberReplace{}
 	MemberDowngrade        Failpoint = memberDowngrade{}
 	MemberDowngradeUpgrade Failpoint = memberDowngradeUpgrade{}
+	GrowMember             Failpoint = &growMember{
```
serathius: We are not growing members here :P 

Usually talk about "cluster" and not "members". And we call the process "scale up". Maybe `ClusterScaleUp`?

tests/robustness/failpoint/cluster.go
```diff
@@ -304,3 +308,28 @@ func patchArgs(args []string, flag, newValue string) error {
 	}
 	return fmt.Errorf("--%s flag not found", flag)
 }
+
+type growMember struct {
+	currentSize int
```
serathius: Which "current" you mean? Maybe `initialSize`.

tests/robustness/failpoint/cluster.go
```diff
@@ -304,3 +308,28 @@ func patchArgs(args []string, flag, newValue string) error {
...
+	}
+	return nil, nil
+}
+
+func (f *growMember) Name() string {
+	return "GrowCluster"
+}
+
+func (f *growMember) Available(config e2e.EtcdProcessClusterConfig, member e2e.EtcdProcess, profile traffic.Profile) bool {
+	return f.currentSize < f.targetSize
```
serathius: Here is the tricky part. If you start from cluster of size 1, then the traffic will talk to the 1 initial member. I would recommend instead to start to from 3 members, remove 2 and add back 2. This is still tricky because `StartNewProc` will allocate unique port numbers so we will need to work around that.

tests/robustness/scenarios/scenarios.go
```diff
@@ -109,12 +109,20 @@ func Exploratory(_ *testing.T) []TestScenario {
...
-			Cluster: *e2e.NewConfig(clusterOfSize1Options...),
-		})
+		scenarios = append(scenarios,
+			TestScenario{
+				Name:    name,
+				Traffic: tp.Traffic,
+				Profile: tp.Profile,
+				Cluster: *e2e.NewConfig(clusterOfSize1Options...),
+			},
+			TestScenario{
```
serathius: No need for this. As long the failpoint is registered in `allFailpoints` it will be used. For testing I recommend just temporarily disabling other failpoints,.

serathius:
Good work on the implementation, implementing a new failpoint is pretty tricky.

Got an error `main_test.go:114: failed to read WAL, err: wal: slice bounds out of range, snapshot[Index: 0, Term: 0], current entry[Index: 4093, Term: 3], len(ents): 0`

Problem we are hitting with `PersistedRequestsCluster` not being able to read WAL files from added members. Current implementation of `PersistedRequestsCluster` assumes that quorum members have all entries in the WAL. In this case only the initial member has the full WAL, so it fails. I think we could rewrite it to consider each index individually and accepting it as long as there is no conflict. 

Unfortunately to merge this feature we will need to rewrite `PersistedRequestsCluster` first.

serathius:
With https://github.com/etcd-io/etcd/pull/20435 we might be able test it again. Can you rebase and see what errors we get? I assume we might need to change how errors are handled when reading WAL.

tests/robustness/failpoint/cluster.go
nwnt: How about `ClusterScaleInAndOut` because we're doing both? :)

tests/robustness/failpoint/cluster.go
nwnt: I no longer need this initialSize.  It's now `scalingTarget`.

tests/robustness/failpoint/cluster.go
nwnt: Yep.  How about the current one?

tests/robustness/scenarios/scenarios.go
nwnt: Removed.

serathius:
From `pull-etcd-robustness-arm64` might need to debug watch.
```
    traffic.go:126: Traffic finished before failure was injected: context canceled
    main_test.go:175: Client didn't collect all events, max revision not set
    main_test.go:180: context canceled
    main_test.go:107: stat .: no such file or directory
```

serathius:
From `pull-etcd-integration-1-cpu-amd64` looks like a flake

serathius:
Interesting no flakes on postsubmit https://testgrid.k8s.io/sig-etcd-postsubmits#post-etcd-integration-4-cpu-arm64&include-filter-by-regex=integration.TestCacheWatchOldRevisionCompacted

nwnt:
From https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/20239/pull-etcd-robustness-arm64/1957745366933704704
```
{"level":"warn","ts":"2025-08-19T10:08:20.850392Z","caller":"etcdserver/server.go:854","msg":"server error","error":"the member has been permanently removed from the cluster"}
{"level":"warn","ts":"2025-08-19T10:08:20.850415Z","caller":"etcdserver/server.go:855","msg":"data-dir used by this member must be removed"}
```

Should I remove each process's data directory after scaling down during `Inject`?

nwnt:
From https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_etcd/20239/pull-etcd-robustness-amd64/1957745366849818624
```
/home/prow/go/src/github.com/etcd-io/etcd/bin/etcd (TestRobustnessExploratoryEtcdTrafficDeleteLeasesClusterOfSize3-test-2) (24040): {"level":"warn","ts":"2025-08-19T10:20:00.008061Z","caller":"etcdserver/server.go:1639","msg":"rejecting member remove request; local member has not been connected to all peers, reconfigure breaks active quorum","local-member-id":"bf19ae4419db00dc","requested-member-remove":"eabdbb777cf498cb","active-peers":1,"error":"etcdserver: unhealthy cluster"}
```
This looks like the condition I should add to `Available`.  What do you think @serathius?

serathius:
> Should I remove each process's data directory after scaling down during Inject?

No, please don't read random logs.

> This looks like the condition I should add to Available. What do you think @serathius?

No, this log is related to fact that membership are rejected if there is suspicion that cluster member are unstable. It requires at least 5 seconds of being connected to all members. This could be mitigated by adding 5 seconds sleep before membership changes.

serathius:
Please read MemberReplace failpoint, seems there is a some overlap with your case.

PR #20240, opened 2025-06-30
Disable progress notify validation until we can guarantee response

serathius:
After https://github.com/etcd-io/etcd/pull/20229 restriction in progress notification has started causing flakes in robustness tests 

image

/cc @ahrtr @siyuanfoundation @fuweid @nwnt @henrybear327


fuweid:
maybe we can re-enable progress notify validation?

PR #20263, opened 2025-07-01
[DO_NOT_MERGE] Reproduce 20221

fuweid:


server/storage/mvcc/watchable_store.go
```diff
@@ -604,16 +602,6 @@ func (w *watcher) send(wr WatchResponse) bool {
...
-	verify.Verify(func() {
-		if w.startRev > 0 {
-			for _, ev := range wr.Events {
-				if ev.Kv.ModRevision < w.startRev {
-					panic(fmt.Sprintf("Event.ModRevision(%d) is less than the w.startRev(%d) for watchID: %d", ev.Kv.ModRevision, w.startRev, w.id))
-				}
-			}
-		}
-	})
-
```
ahrtr: I strongly suggest to keep this commit in your test. So that we can exclude the server side issue. You also need to keep the field `startRev` in `watcher`

client/v3/watch.go
```diff
@@ -855,6 +855,9 @@ func (w *watchGRPCStream) serveSubstream(ws *watcherStream, resumec chan struct{
 				nextRev = wr.Events[len(wr.Events)-1].Kv.ModRevision + 1
 			}
 
+			if nextRev < ws.initReq.rev {
+				fmt.Println("oops --------> initReq.rev(", ws.initReq.rev, ") vs nextRev(", nextRev, ")")
```
ahrtr: Please also print whether there is any Events in the watch response. If yes, print the biggest `ModRevision`?

ahrtr:
Just to double check, you can only reproduce it after reverting the fix https://github.com/etcd-io/etcd/pull/20229? In other words, can you reproduce it without reverting the fix? The fix might be exactly the fix to both issues.

```
oops --------> initReq.rev( 1000 ) vs nextRev( 12 )
```

fuweid:
> Just to double check, you can only reproduce it after reverting the fix https://github.com/etcd-io/etcd/pull/20229? In other words, can you reproduce it without reverting the fix? The fix might be exactly the fix to both issues.

Yes. With #20229, there is no progress notification received so that initReq.rev won't be decreased.

client/v3/watch.go
fuweid: Will update it.

server/storage/mvcc/watchable_store.go
fuweid: agree

fuweid:
cc @ahrtr @serathius seeking for comments if it's good move to panic as receiving decreased revision. thanks

client/v3/watch.go
```diff
@@ -848,11 +848,20 @@ func (w *watchGRPCStream) serveSubstream(ws *watcherStream, resumec chan struct{
 				}
 			} else {
 				// current progress of watch; <= store revision
-				nextRev = wr.Header.Revision + 1
+				storeRev := wr.Header.Revision + 1
+				if wr.Err() == nil && storeRev < ws.initReq.rev {
+					panic(fmt.Sprintf("received watch response header with revision %d less than expected next revision %d: %v", storeRev, nextRev, wr))
```
serathius, line 853: Introducing panic in the critical path should be considered a change of behavior that many projects will not accept. 

Definitely not K8s which has a rule exactly against introducing a panic in the code.

This change should definitely not be introduced as enabled by default. We could consider breaking the watch stream when we are 100% sure this is incorrect, and leave panic as optional `Strict` mode.

client/v3/watch.go
```diff
@@ -848,11 +848,20 @@ func (w *watchGRPCStream) serveSubstream(ws *watcherStream, resumec chan struct{
...
+				}
+				nextRev = storeRev
 			}
 
 			if len(wr.Events) > 0 {
-				nextRev = wr.Events[len(wr.Events)-1].Kv.ModRevision + 1
+				firstRev := wr.Events[0].Kv.ModRevision
+				lastRev := wr.Events[len(wr.Events)-1].Kv.ModRevision + 1
+				if wr.Err() == nil && (lastRev < ws.initReq.rev || firstRev < ws.initReq.rev) {
+					panic(fmt.Sprintf("received watch events [%d, %d) less than expected next revision %d", firstRev, lastRev, ws.initReq.rev))
```
serathius, line 862: FYI I would like to use the same exact places to introduce etcd verify statements and utilize it in Antithesis testing https://github.com/etcd-io/etcd/issues/20283

fuweid: does it dymatically add assertation during executation? or need to write code for it?
If use gofail, it sounds good to me.

client/v3/watch.go
fuweid: Other option is like, we can just add warning log for it. as least, during the test, we can tell that server was sending descreased revision.

client/v3/watch.go
serathius: I will want to make it work similar to gofail. 

For example have
```
import "go.etcd.io/etcd/pkg/v3/verify"

....
  verify.Verify(event.Kv.ModRev >= rev, "Watch revision never decreases")
....
```
And have Antithesis replace it before compilcation
```
import "github.com/antithesishq/antithesis-sdk-go/assert"

....
  assert.Always(event.Kv.ModRev >= rev, "Watch revision never decreases")
....
```

client/v3/watch.go
serathius: In tests we can enable panic, no-one will read logs from tests to find the failure.

This is how the existing verify package works. https://github.com/etcd-io/etcd/blob/main/client/pkg/verify/verify.go

Issue #20171, opened 2025-06-13
Add unit tests for etcdctl


siyuanfoundation:
### What would you like to be added?

Add unit tests for etcdctl. 
Tasks include:

- [ ] Add a fake client
- [ ] Add unit tests for most files in etcdctl/ctlv3/command

### Why is this needed?

Right now all the changes to `etcdctl` can only manually tested, which is hard to verify for reviewers. To improve reliability, we need to have proper testing.

hwdef:
I want to do this.
Please assign it to me

Elbehery:
@siyuanfoundation Could this be split among many contributors ? 

cc @ahrtr @ivanvc

siyuanfoundation:
> @siyuanfoundation Could this be split among many contributors ? 
> 
> cc @ahrtr @ivanvc 

Yes, I will break it into subtasks once the FakeClient fixture is added.

Elbehery:
> > [@siyuanfoundation](https://github.com/siyuanfoundation) Could this be split among many contributors ?
> > cc [@ahrtr](https://github.com/ahrtr) [@ivanvc](https://github.com/ivanvc)
> 
> Yes, I will break it into subtasks once the FakeClient fixture is added.

thx for your response 👍🏽 

Anyone is working on the `FakeClient` already ?

hwdef:
@Elbehery 
If no one else does it, I'm willing to give it a try.

Elbehery:
i assigned you last week already 🙂🙂

hwdef:
Thanks! I'll submit PR in a few days.

hwdef:
still work in progress

hwdef:
still work in progress

akstron:
Hi @siyuanfoundation , bringing up your attention on this: https://github.com/etcd-io/etcd/pull/21230#issuecomment-5099226471 . Let me know if this issue is not prioritised or some other decision was made?

PR #20270, opened 2025-07-03
etcdctl: add fake client for UT

hwdef:
Part of: #20171 

/cc @siyuanfoundation @Elbehery

### DIFF
--- a/etcdctl/ctlv3/command/alarm_command_test.go
+++ b/etcdctl/ctlv3/command/alarm_command_test.go
@@ -1,0 +1,83 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package command
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/spf13/cobra"
+
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/etcdctl/v3/ctlv3/command/fakeclient"
+)
+
+func newTestRoot() *cobra.Command {
+	root := &cobra.Command{Use: "etcdctl"}
+	root.AddGroup(NewKVGroup(), NewClusterMaintenanceGroup(), NewConcurrencyGroup(), NewAuthenticationGroup(), NewUtilityGroup())
+	RegisterGlobalFlags(root)
+	return root
+}
+
+func TestAlarmList_PrintsAlarms(t *testing.T) {
+	var gotReq int
+	fc := &fakeclient.Client{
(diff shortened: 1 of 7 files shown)


etcdctl/ctlv3/fakeclient/fake_client.go
```diff
@@ -0,0 +1,267 @@
...
+	c.Maintenance = fakeMaint
+	c.Cluster = fakeCluster
+	c.Watcher = &watcherStub{}
+
+	return c
+}
+
+// KV
+
+type kvStub struct{}
```
siyuanfoundation: we need to think about how to pass in the mock response. 
Like a fake dict of kvs, a list of alarms.

Can you try to use the fakeclient to test alarm_command as an example of how it should work?

hwdef:
@siyuanfoundation @Elbehery @serathius
PTAL. This PR is ready for review :)

etcdctl/ctlv3/command/alarm_command_test.go
```diff
@@ -0,0 +1,134 @@
...
+	newClientFunc = func(cfg clientv3.Config) (*clientv3.Client, error) {
+		return fakeclient.WrapAsClientV3(fc), nil
+	}
+	return func() { newClientFunc = oldNew }
+}
+
+func newTestRoot() *cobra.Command {
+	root := &cobra.Command{Use: "etcdctl"}
+	root.AddGroup(NewKVGroup(), NewClusterMaintenanceGroup(), NewConcurrencyGroup(), NewAuthenticationGroup(), NewUtilityGroup())
+	root.PersistentFlags().StringSlice("endpoints", []string{"127.0.0.1:2379"}, "")
```
siyuanfoundation: abstract to a function shared with `init()`

etcdctl/ctlv3/command/alarm_command_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/etcdctl/v3/ctlv3/command/fakeclient"
+)
+
+func withStdoutCapture(t *testing.T, fn func()) string {
```
siyuanfoundation: move this and `setupFake` to a separate file like test_utils.go

siyuanfoundation:
Overall this approach works. 
But if the tests run in parallel, there will be race condition on the `var newClientFunc`. To ensure only 1 test runs at a time, we will need to add a new test suite in `test.sh` with `go test -p 1`. 

Another option would be refactor all the functions like `NewAlarmCommand` to take a client argument. Personally I would prefer this. 

@ahrtr would really appreciate your input. Thanks!

hwdef:
PTAL @ahrtr @Elbehery

etcdctl/ctlv3/command/global.go
```diff
@@ -30,39 +31,11 @@ import (
...
 	"go.etcd.io/etcd/client/pkg/v3/srv"
-	"go.etcd.io/etcd/client/pkg/v3/transport"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/pkg/v3/cobrautl"
 	"go.etcd.io/etcd/pkg/v3/flags"
 )
 
-// GlobalFlags are flags that defined globally
-// and are inherited to all sub-commands.
-type GlobalFlags struct {
```
siyuanfoundation, line 41: It seems this struct is not very useful to begin with. How about moving the refactor to a separate PR?

hwdef: check unused flag.

PR #20281, opened 2025-07-04
Avoid lowering revision of watchers in the future after restore

serathius:
Fixes https://github.com/etcd-io/etcd/issues/20221

/cc @ahrtr @fuweid 

The bug is in resync setting minRev. It assumes that watcher can only be behind, so after synchronizing it sets revision to current. However restore moves all watcher to unsynced, even those that are in the future. Those watchers don't get any events from resync, but their `minRev` is overriden.

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -39,6 +39,9 @@ var (
 
 	// maxWatchersPerSync is the number of watchers to sync in a single batch
 	maxWatchersPerSync = 512
+
+	// maxResyncPeriod is the period of executing resync.
+	watchResyncPeriod = 100 * time.Millisecond
 )
 
 func ChanBufLen() int { return chanBufLen }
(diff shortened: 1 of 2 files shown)


server/storage/mvcc/watchable_store.go
```diff
@@ -370,7 +372,7 @@ func (s *watchableStore) syncWatchers(evs []mvccpb.Event) (int, []mvccpb.Event)
 			// Next retry of syncWatchers would try to resend the compacted watch response to w.ch
 			continue
 		}
-		w.minRev = curRev + 1
```
serathius, line 373: The bug

ahrtr: It only happens when watching a future revision?

ahrtr: We might want to add such protection for all places which update `w.minRev`.

```
func (w *watcher) updateMinRev(rev) {
    if rev > w.minRev {
        w.minRev = rev
   }
}
```

serathius: Applying this everywhere might hide other kinds of bugs. I would prefer to cover the code with tests and then rewrite it.

ahrtr (approved):
LGTM

Thanks for the great work!

server/storage/mvcc/watchable_store.go
ahrtr: OK

fuweid (commented):
LGTM

Good catch!

serathius:
Manual backport https://github.com/etcd-io/etcd/pull/20290

PR #20282, opened 2025-07-04
Make WatchID unique per member

serathius:
One of the issues with debugging https://github.com/etcd-io/etcd/issues/20221 was problems with identifying watch requests. Currently watch id is unique per watch stream, meaning it's reset after each connection is re-established, but makes it very hard to debug. With watch id not being unique, storage cannot use it for identify clients, and uses pointers to watcher structs. This makes it hard to debug. 

I don't see any reason we cannot make the watch ids unique per member to make debugging easier. The only consequence of it is fact that watch ids no longer will reset, however I don't think there is a risk of hitting MaxInt64, and even if we do I don't think rotating is a big problem. 

For now just proposing a draft to get people thoughts. 
/cc @ahrtr @fuweid

### DIFF
--- a/server/storage/mvcc/watcher.go
+++ b/server/storage/mvcc/watcher.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"errors"
 	"sync"
+	"sync/atomic"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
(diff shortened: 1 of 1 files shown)


server/storage/mvcc/watcher.go
```diff
@@ -123,11 +124,7 @@ func (ws *watchStream) Watch(id WatchID, key, end []byte, startRev int64, fcs ..
 	}
 
 	if id == clientv3.AutoWatchID {
-		for ws.watchers[ws.nextID] != nil {
```
fuweid, line 126: sounds good to use unique ID across all the watch streams.
it's draft version. just want to mention that it still needs to avoid allocate that ID which has been occupied by user request.

serathius: Right!

PR #20289, opened 2025-07-06
Allow partial entries for PersistedRequestsCluster

nwnt:
Make `PersistedRequestsCluster` retry with the WAL file starting index to deal with the error found in #20239.


nwnt:
@serathius this is still a WIP but want to get your early feedback about it.  Do you think this approach is ok?

nwnt:
will close this one as WAL logic needs a rework.

PR #20303, opened 2025-07-07
scripts: modernize SBOM generation and workflow

vivekpatani:
- Generation: syft (pinned v1.44.0 via go install) emits both formats in one pass.
- Augmentation: `jq` transforms add document metadata and fix UNKNOWN versions for local modules.
- Determinism: byte-stable output (strips timestamps + UUIDs).
- Workflow (mirrors `bom_pass`): make `verify-modern-sbom` diffs against committed files; make `fix-modern-sbom` regenerates.


vivekpatani:
Sample output from a local run

```shell
> PASSES="modern_sbom" ./scripts/test.sh
Running with --race
Starting at: Mon Jun 30 13:20:59 PDT 2025

'modern_sbom' started at Mon Jun 30 13:20:59 PDT 2025
validating modern SBOM generation...
syft is already installed: 1.27.1
sbomasm is already installed locally
generating modern SBOM files for version dev
generating SPDX SBOM...
 ✔ Indexed file system                                                                                                                                   . 
 ✔ Cataloged contents                                                                     19be9340167c940e8dc3cbae0b85682c19973b77c6f3a30def20c6976b478d0b 
   ├── ✔ Packages                        [912 packages]  
   ├── ✔ Executables                     [4 executables]  
   ├── ✔ File metadata                   [22 locations]  
   └── ✔ File digests                    [22 files]  
SPDX SBOM generated: /var/folders/6h/5jv5cfs54v72cvvc_pvysdqw0000gn/T/tmp.EUNXl5rXEh/sbom.spdx.json
generating CycloneDX SBOM...
 ✔ Indexed file system                                                                                                                                   . 
 ✔ Cataloged contents                                                                     19be9340167c940e8dc3cbae0b85682c19973b77c6f3a30def20c6976b478d0b 
   ├── ✔ Packages                        [912 packages]  
   ├── ✔ Executables                     [4 executables]  
   ├── ✔ File metadata                   [22 locations]  
   └── ✔ File digests                    [22 files]  
CycloneDX SBOM generated: /var/folders/6h/5jv5cfs54v72cvvc_pvysdqw0000gn/T/tmp.EUNXl5rXEh/sbom.cdx.json
augmenting SBOM files with enhanced metadata...
augmenting SPDX SBOM...
{"level":"info","ts":1751314862.2400372,"caller":"edit/spdx_edit.go:82","msg":"SPDX error updating supplier: not supported"}
{"level":"info","ts":1751314862.2400792,"caller":"edit/spdx_edit.go:82","msg":"SPDX error updating repository: not supported"}
SPDX SBOM augmented successfully
augmenting CycloneDX SBOM...
CycloneDX SBOM augmented successfully
SBOM augmentation completed successfully
modern SBOM files generated successfully
files created:
  - /var/folders/6h/5jv5cfs54v72cvvc_pvysdqw0000gn/T/tmp.EUNXl5rXEh/sbom.spdx.json
  - /var/folders/6h/5jv5cfs54v72cvvc_pvysdqw0000gn/T/tmp.EUNXl5rXEh/sbom.cdx.json
validating generated SBOM files...
Validated: /var/folders/6h/5jv5cfs54v72cvvc_pvysdqw0000gn/T/tmp.EUNXl5rXEh/sbom.spdx.json
Validated: /var/folders/6h/5jv5cfs54v72cvvc_pvysdqw0000gn/T/tmp.EUNXl5rXEh/sbom.cdx.json
all SBOM files validated successfully
modern SBOM generation validation PASSED
'modern_sbom' PASSED and completed at Mon Jun 30 13:21:02 PDT 2025
SUCCESS
```

scripts/test.sh
```diff
@@ -57,7 +57,7 @@ if [ -n "${OUTPUT_FILE}" ]; then
   exec > >(tee -a "${OUTPUT_FILE}") 2>&1
 fi
 
-PASSES=${PASSES:-"gofmt bom dep build unit"}
+PASSES=${PASSES:-"gofmt bom modern_sbom dep build unit"}
```
serathius: Please don't integrate here, but create a makefile target.

vivekpatani: Ack.

Updated PR. Added the following to `Makefile`:

```makefile
# Static analysis
.PHONY: verify
verify: verify-gofmt verify-bom verify-modern-sbom verify-lint verify-dep verify-shellcheck verify-goword \
	verify-govet verify-license-header verify-mod-tidy \
	verify-shellws verify-proto-annotations verify-genproto verify-yamllint \
	verify-govet-shadow verify-markdown-marker verify-go-versions

.PHONY: verify-modern-sbom
verify-modern-sbom:
	PASSES="modern_sbom" ./scripts/test.sh
```

@serathius

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,196 @@
...
+    # augment SPDX SBOM with etcd-specific metadata
+    log_callout "augmenting SPDX SBOM..."
+    if "${sbomasm_cmd}" edit --append --subject Document \
+        --author 'etcd maintainers (https://github.com/etcd-io/etcd/blob/main/MAINTAINERS)' \
+        --supplier 'etcd project (https://etcd.io/)' \
+        --repository 'https://github.com/etcd-io/etcd' \
+        --license 'Apache-2.0 (https://raw.githubusercontent.com/etcd-io/etcd/main/LICENSE)' \
+        "${output_dir}/sbom.spdx.json" > "${output_dir}/sbom.spdx.json.tmp"; then
+        mv "${output_dir}/sbom.spdx.json.tmp" "${output_dir}/sbom.spdx.json"
+        log_success "SPDX SBOM augmented successfully"
```
jmhbnz: When running `make verify` locally on this pr I see:

```bash
augmenting SBOM files with enhanced metadata...                                                                                                               
augmenting SPDX SBOM...                                    
{"level":"info","ts":1752054550.5096443,"caller":"edit/spdx_edit.go:82","msg":"SPDX error updating supplier: not supported"}
{"level":"info","ts":1752054550.5097733,"caller":"edit/spdx_edit.go:82","msg":"SPDX error updating repository: not supported"}
```

Are the errors on updating repository and supplier expected?

jmhbnz (commented):
Awesome progress @vivekpatani - thanks for drafting this!

scripts/generate-modern-sbom.sh
vivekpatani: > Are the errors on updating repository and supplier expected?

My understanding is the `sbomasm` tool intentionally does not support `--supplier` and `--repository` flags for SPDX format in v1.0.4 (which is the current latest). However, the newest `main` has some diff in the supplier version which addresses the issue. I'd prefer sticking to a release version, rather than build main manually.

https://github.com/interlynk-io/sbomasm/blob/a0d5adcfacfb1a822642edfc143d54c3e0cc4b4a/pkg/edit/spdx_edit.go#L127C1-L150C2 vs https://github.com/interlynk-io/sbomasm/blob/a23a22f1438c75c972902f428e66c52b6667f363/pkg/edit/spdx_edit.go#L144-L186


Obviously, I'm open to feedback, and can adjust.

@jmhbnz thoughts?

vivekpatani:
Hello team, gentle bump @jmhbnz @ivanvc @ahrtr @serathius

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,196 @@
...
+    local output_dir="$2"
+    
+    log_callout "augmenting SBOM files with enhanced metadata..."
+    
+    local sbomasm_cmd="${SBOMASM_BIN:-sbomasm}"
+    
+    # augment SPDX SBOM with etcd-specific metadata
+    log_callout "augmenting SPDX SBOM..."
+    if "${sbomasm_cmd}" edit --append --subject Document \
+        --author 'etcd maintainers (https://github.com/etcd-io/etcd/blob/main/MAINTAINERS)' \
```
ivanvc: This should point to the `OWNERS` file:

```suggestion
        --author 'etcd maintainers (https://github.com/etcd-io/etcd/blob/main/OWNERS)' \
```

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,196 @@
...
+
+VERSION=${1:-"dev"}
+OUTPUT_DIR=${2:-"."}
+
+function install_syft() {
+    local syft_bin="./bin/syft"
+    if ! command -v syft >/dev/null && [ ! -f "${syft_bin}" ]; then
+        log_callout "installing syft to local bin directory..."
+        mkdir -p ./bin
+        curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b ./bin
```
ivanvc: I think that because Syft and SBOMAS are both Go projects, we could add their dependencies to `tools/mod` and maintain the version through Dependabot, rather than relying on their installation scripts or pulling the latest release. For example, if there's a breaking change, we could lock to a specific version while we update the code to make it compatible with the latest.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,196 @@
...
+set -euo pipefail
+
+source ./scripts/test_lib.sh
+
+VERSION=${1:-"dev"}
+OUTPUT_DIR=${2:-"."}
+
+function install_syft() {
+    local syft_bin="./bin/syft"
+    if ! command -v syft >/dev/null && [ ! -f "${syft_bin}" ]; then
```
ivanvc: I would simplify this and use only our provided binaries to avoid mismatches if the user is running a different version locally.

scripts/release.sh
ivanvc, line 1: I'm not a specialist in SBOMs. Is this the correct way to distribute them (by placing them as released files, rather than in the repository)?

FWIW, I like the idea.

ivanvc (commented):
Thanks for putting this together, @vivekpatani, great job :)

I ran `scripts/generate-modern-sbom.sh v3.6.1` locally, but the output file shows UNKNOWN for our local packages both in `sbom.cdx.json` and `sbom.spdx.json`. And in `sbom.spdx.json`, all the dependencies show `NOASSERTION` for `licenseDeclared` and `copyrightText` (among other fields). I assume this is not the expected behavior, right? Unless I executed them incorrectly.

scripts/generate-modern-sbom.sh
ivanvc: In this case, I would hold the pull request until they release the new version (assuming we have verified that this addresses the issue).

viveksahu26: Hey, basically SPDX at document level doesn't support these fields `supplier` and `repository`. Instead it has [author](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#68-creator-field) which could be a person or organization.  However these fields `supplier` and `repository` are supported at package level, which are referenced via [PackageSupplier](https://spdx.github.io/spdx-spec/v2.3/package-information/#75-package-supplier-field) and [PackageDownloadLocation](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field). 

That's why sbomam throws an error that SPDX doesn't support these fields, whereas CycloneDX do support these fields at document level. If you try the same same using CycloneDX SBOM it would work fine. 

One thing we can do here, under **Author** as a `person`, add this part: `'etcd maintainers (https://github.com/etcd-io/etcd/blob/main/MAINTAINERS)'`, which represents list of all authors as a person, whereas this part: `'etcd project (https://etcd.io/)'`, which represents "organization" and it's "website" can be added under "Author" as a `organization`, instead of supplier.

Currently, sbomasm do support adding of author as person, but not for organization. The [issue](https://github.com/interlynk-io/sbomasm/issues/178) is already there for the same to add support for adding suthor as an organization, hopefully in next release will fix it.

scripts/generate-modern-sbom.sh
vivekpatani: now references OWNERS. @ivanvc

scripts/generate-modern-sbom.sh
vivekpatani: - Instead, the script pins `SYFT_VERSION="v1.44.0"` and installs via `go install @version`. Version bumps are a one-line change. sbomasm has been dropped entirely, so only one tool to manage.
- Happy to move to `tools/mod` if the team prefers Dependabot tracking despite the dependency footprint.

Thoughts? @ivanvc

scripts/generate-modern-sbom.sh
vivekpatani: Done. The script always runs `go install github.com/anchore/syft/cmd/syft@${SYFT_VERSION}` with a pinned version.

scripts/release.sh
vivekpatani: Kube includes SBOMs as release artifacts since 1.26, containerd does the same. SBOMs are generated at release time from the source tree and uploaded alongside binaries and SHA256SUMS, not committed to the repo.

scripts/generate-modern-sbom.sh
vivekpatani: Dropped sbomasm entirely. As @viveksahu26  correctly identified, the SPDX spec doesn't support supplier and repository at document level that's a spec limitation, not a tool bug that will be fixed.

The new approach uses jq for metadata augmentation, which is already available.

vivekpatani:
@ivanvc not inline but the response to the above comment:
- `UNKNOWN` versions: These were the local workspace modules (go.etcd.io/etcd/api/v3, client/v3, etc.) that use replace directives pointing to local paths. Syft can't resolve their versions from source. The script now post-processes with jq to replace all UNKNOWN versions, I think seems like a decent approach?
- `NOASSERTION` for licenseDeclared: This is correct SPDX behavior. licenseDeclared means "what the package author explicitly declared", syft can't determine that from source scanning. The actual license detection is in licenseConcluded, where syft reports licenses for ~609/923 packages.

Sorry it took so long @ahrtr @ivanvc feedback, would be appreciated.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 The etcd Authors
```
ahrtr:
```suggestion
# Copyright 2026 The etcd Authors
```

scripts/release.sh
```diff
@@ -252,11 +252,20 @@ main() {
...
+  # generate modern SBOM files
+  log_callout "generating modern SBOM files..."
+  if ./scripts/generate-modern-sbom.sh "${RELEASE_VERSION}" "./release"; then
+    log_success "modern SBOM files generated successfully"
+  else
+    log_error "failed to generate modern SBOM files. Aborting."
+    exit 1
+  fi
+
+  # Generate SHA256SUMS (including SBOM files)
```
ahrtr: not sure why update the release script.  Currently the bom file is only updated each time when we bump dependencies. Normally the generated bom file should be stable as long as we don't bump any dependencies. Can we follow the same pattern?  Or did I miss anything?

ahrtr:
Can we generate a bom file and add it into the repo (this PR)? Is there any difference with the existing bom file?

scripts/release.sh
ivanvc: Replying to this and this comment:

> Can we generate a bom file and add it into the repo (this PR)? Is there any difference with the existing bom file?

The idea with this BOM is that we distribute it as part of the release, rather than placing it in the repository. It's how other projects distribute the SBOM, same as the binary checksums, it's considered a release artifact. For example, k8s publishes it in sbom.k8s.io (https://kubernetes.io/releases/download/#container-image-signatures).

ahrtr: > we distribute it as part of the release

I have no any objection on this.



> rather than placing it in the repository.

I am not sure why we don't place it in the repository if it's relative stable. One of the benefit is that we can check the latest BOM anytime we want, even before we publish next version. But I agree this isn't a big problem.

ivanvc: As a first step, and looking to replace our JSON file, I believe it makes sense to check in the new file. And in the long run, we should aim to move it completely to the release artifacts.

scripts/generate-modern-sbom.sh
vivekpatani: done.

scripts/release.sh
vivekpatani: followed the existing `bom_pass` pattern:
- Reverted scripts/release.sh changes
- SBOMs committed at repo root (sbom.spdx.json, sbom.cdx.json)
- make `verify-modern-sbom` diffs against committed files
- make `fix-modern-sbom` regenerates them when deps change

As an aside: verified output is byte-deterministic across runs (stripped 4 non-deterministic fields: SPDX created set to epoch, CDX serialNumber/timestamp/tools deleted).

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,133 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+source ./scripts/test_lib.sh
+
```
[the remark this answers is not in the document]
vivekpatani: - `release.sh` integration was reverted in prior round
- etcd convention is to treat `jq` as baseline (7 scripts use it without checking, including bom_pass)
- bash's `jq: command not found` IS actionable.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,133 @@
...
+  local version="$1"
+  local output_dir="$2"
+
+  mkdir -p "${output_dir}"
+  log_callout "generating modern SBOM files for version ${version}"
+
+  log_callout "ensuring syft ${SYFT_VERSION} is installed..."
+  GOARCH="" GOOS="" run go install "github.com/anchore/syft/cmd/syft@${SYFT_VERSION}" || return 2
+  local syft_bin
+  syft_bin="$(go env GOPATH)/bin/syft"
```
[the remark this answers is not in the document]
vivekpatani: good catch, addressed.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,133 @@
...
+  log_callout "  - ${output_dir}/sbom.spdx.json"
+  log_callout "  - ${output_dir}/sbom.cdx.json"
+}
+
+function augment_sbom() {
+  local version="$1"
+  local output_dir="$2"
+
+  log_callout "augmenting SBOMs with project metadata..."
+
```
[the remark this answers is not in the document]
vivekpatani: Will update description.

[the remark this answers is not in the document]
vivekpatani: done.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,133 @@
...
+      if (.name | startswith("go.etcd.io/etcd")) and .versionInfo == "UNKNOWN"
+      then .versionInfo = $ver
+      else . end
+    ] |
+    .name = "etcd-\($ver)" |
+    .documentNamespace = "https://github.com/etcd-io/etcd/releases/tag/\($ver)/sbom.spdx.json" |
+    .creationInfo.creators = [
+      "Organization: etcd project (https://etcd.io/)",
+      "Tool: syft"
+    ]
```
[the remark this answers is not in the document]
vivekpatani: good catch, addressed.

vivekpatani:
@ahrtr added it to the commit. Thank you.

vivekpatani:
@ahrtr @ivanvc one question as a follow up. The previous sbom process excluded tools, do we want to continue to exclude it or do we want to include it? The current implementation doesn't exclude it.

Should I address `Phase 2` in this PR or as separate follow-ups?
- Phase 2a - wire SBOM generation into `release.sh` for artifact distribution (gs://etcd + GitHub releases).
- Phase 2b - remove `bill-of-materials.json` once the new pipeline is stable.

  Happy either way, just want to confirm the scope before expanding this PR vs. shipping the foundation first.

ivanvc:
> Should I address Phase 2 in this PR or as separate follow-ups?

We discussed this PR at today's triage meeting. Yes, let's do this later.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,151 @@
...
+  # byte-stable output we rely on for diff-based verification).
+  jq --arg ver "${version}" --arg syftver "${SYFT_VERSION}" '
+    .packages = [.packages[] |
+      if (.name | startswith("go.etcd.io/etcd")) and .versionInfo == "UNKNOWN"
+      then .versionInfo = $ver
+      else . end
+    ] |
+    .name = "etcd-\($ver)" |
+    .documentNamespace = "https://github.com/etcd-io/etcd/releases/tag/\($ver)/sbom.spdx.json" |
+    .creationInfo.created = "1970-01-01T00:00:00Z" |
```
ivanvc: Do we need to specify this field? I understand that this may not help to detect drift, but looking at the result in the `sbom.spdx.json` with the `0` epoch date doesn't look correct.

Could we skip it instead?

vivekpatani: Thanks for the feedback. Addressed, `creationInfo.created` is no longer overwritten. The committed SBOM keeps syft's real timestamp; non-deterministic fields (SPDX created, CDX serialNumber/timestamp/tools) are stripped at diff time in `modern_sbom_pass` instead. Validated that drift detection still catches real dependency changes.

Does that make sense? @ivanvc

scripts/test.sh
```diff
@@ -497,6 +497,46 @@ function gomodguard_pass {
   run_for_workspace_modules module_gomodguard "${tool_bin}"
 }
 
+function modern_sbom_pass {
```
serathius: Please don't use `_pass` hack in makefile. Just dedicated script.

vivekpatani: Addressed, moved to dedicated script. @serathius

scripts/generate-modern-sbom.sh
ivanvc: Thanks, it does.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,153 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+source ./scripts/test_lib.sh
+
+SYFT_VERSION="v1.44.0"
```
ivanvc: This may become out of date quickly. I suggest we add the entry to `tools/mod`, and we keep it up to date with Dependabot.

We can later then install it using `run_go_tool` instead of line 33.

ivanvc (approved):
This LGTM. I suggest we address my comment in a follow-up pull request, as adding a tool may add many changes, and this pull request is small enough to be merged soon.

scripts/generate-modern-sbom.sh
```diff
@@ -0,0 +1,153 @@
...
+
+  mkdir -p "${output_dir}"
+  log_callout "generating modern SBOM files for version ${version}"
+
+  log_callout "ensuring syft ${SYFT_VERSION} is installed..."
+  GOARCH="" GOOS="" run go install "github.com/anchore/syft/cmd/syft@${SYFT_VERSION}" || return 2
+  local syft_bin gobin
+  gobin="$(go env GOBIN)"
+  gobin="${gobin:-$(go env GOPATH)/bin}"
+  syft_bin="${gobin}/syft"
```
serathius: Please add syft to `tools/mod` and use go install like any other tool in etcd.

ivanvc:
@vivekpatani, we discussed this PR during today's community meeting. We agreed that it would be a good idea to somehow have a comparison of the old `bill-of-materials.json` and the new `sbom.{spdx,cdx}.json‎` files.

Given the nature (and length) of the files, we think it may be a good idea to have AI do a comparison.

scripts/generate-modern-sbom.sh
vivekpatani: Moved, let me know if that's the correct way to do it. Thanks.

scripts/generate-modern-sbom.sh
vivekpatani: Addressed, let me know if that's the correct way to do it. Thanks.

vivekpatani:
> we discussed this PR during today's community meeting. We agreed that it would be a good idea to somehow have a comparison of the old bill-of-materials.json and the new sbom.{spdx,cdx}.json‎ files.
Given the nature (and length) of the files, we think it may be a good idea to have AI do a comparison.

@ivanvc https://gist.github.com/vivekpatani/8075f444870569c8182c1be78a994b7a, does this make sense

serathius:
> Given the nature (and length) of the files, we think it may be a good idea to have AI do a comparison.

Please provide raw versions too :P

vivekpatani:
Added the raw diff as bom-vs-sbom.diff.

vivekpatani:
Hello,

IIUC, putting syft in `tools/mod` is what's breaking the verify job. `syft` drags in way too much, and one of its deps collides with the version the rest of etcd uses. fix is to give syft its own module that's outside the workspace.

`pull-etcd-verify` runs `make verify`, and it dies on the `verify-dep` step. that step has one rule: every dependency has to be the same version across all modules in the workspace. syft v1.45.1 pulls in containerd, which wants a newer protobuf than we pin:

- tools/mod (via syft): google.golang.org/protobuf v1.36.12-0.2026...
- everything else: google.golang.org/protobuf v1.36.11

two protobuf versions in one workspace → `verify-dep` says no → job fails.

syft isn't a normal tool. our other tools (`golangci-lint`, `cfssl`, etc.) happen to use deps that line up with ours. syft doesn't, it has ~463 deps and needs newer versions than we pin. and because `tools/mod` is part of the go workspace, it's forced to agree with everything else. it can't.

I think the fix would be(?):

- new `tools/sbom/go.mod` with just `syft`
- keep it out of `go.work` (and out of `update_go_workspace.sh`, otherwise make fix quietly re-adds it and we're back here)
- install it via go install from that module
- add `/tools/sbom` to `dependabot`


As an aside, the govulncheck failure isn't from this PR. govulncheck is installed @latest (scripts/test.sh), and it rolled from v1.3.0 (x-tools v0.44.0) to v1.4.0 (x-tools v0.46.0), which panics on a generic in `github.com/olekukonko/ll` (pulled in via `tablewriter`, used by etcdctl/etcdutl/tests). #21971 passed govulncheck on Jun 16 with v1.3.0 and the same olekukonko/ll v0.1.6; this PR hit v1.4.0 on Jun 17 and panicked. it'll hit main and other PRs once they're retested, so we should pin govulncheck instead of @latest?

@ivanvc @serathius @ahrtr

ahrtr:
Pls resolve the workflow pull-etcd-verify failure, thx

vivekpatani:
@ivanvc @ahrtr @serathius I tried to resolve `etcd-pull-verify` however it's a bit more nuanced.

I tried this exactly as suggested first: `syft` was added to `tools/mod` and installed via `run_go_tool`, like every other tool (that's what was in the commit that ran CI). It's what surfaced the `pull-etcd-verify` failure, and digging in, `syft` turns out to be a special case.

`tools/mod` is part of `go.work`, and `verify-dep` (`dep_pass` in scripts/test.sh) enforces a single version per dependency across all workspace modules. `syft v1.45.1` pulls ~280 transitive modules, and several of them require versions newer than the rest of etcd pins:
- google.golang.org/protobuf `v1.36.12-0.2026...` vs our `v1.36.11`
- go.opentelemetry.io/otel (+ `/metric`, `/sdk`, `/sdk/metric`, `/trace`)
- github.com/prometheus/procfs, golang.org/x/exp

Two versions of the same dep in one workspace makes `verify-dep` fail, by design. This is unique to syft: `golangci-lint`, `cfssl`, `gomodguard`, etc. happen to align with our pinned versions, but syft is a full application (containerd, AWS SDK, the charm TUI stack, …) on a fast-moving dependency tree, so it can't be held to the workspace's versions.

The fix keeps the intent of your guidance (install via go install, version tracked by Dependabot) but moves `syft` to its own module, `tools/sbom`, kept out of `go.work`. Because `verify-dep` only inspects workspace modules, syft's deps no longer collide, and the rest of the workspace stays byte-identical to main (zero delta). I also excluded `tools/sbom` from `update_go_workspace.sh` so make fix won't silently re-add it, and added a `/tools/sbom` Dependabot entry.

So: agreed on `go install` + `Dependabot`, the only deviation is `workspace` membership, which is forced by syft's dependency footprint. Open to alternatives if you'd prefer (e.g. pinning syft to a version whose deps align), but that seems more fragile across future bumps.

tools/sbom/tools.go
```diff
@@ -0,0 +1,29 @@
...
+//
+// syft lives in its own module, deliberately outside go.work, because it
+// pulls in a large dependency set (containerd, newer protobuf, etc.) that
+// conflicts with the versions pinned across the etcd workspace. Keeping it
+// here isolates those dependencies from the workspace 'verify-dep' check.
+
+package tools
+
+import (
+	_ "github.com/anchore/syft/cmd/syft"
```
ahrtr, line 28: I am curious why not to add this into `tools/mod/tools.go`? why add a new mod and file?

.github/dependabot.yml
```diff
@@ -20,6 +20,13 @@ updates:
     allow:
       - dependency-type: direct
 
+  - package-ecosystem: gomod
+    directory: /tools/sbom # syft, kept outside the go workspace
```
ahrtr, line 24: same question as https://github.com/etcd-io/etcd/pull/20303#discussion_r3465878828

scripts/update_go_workspace.sh
```diff
@@ -49,7 +49,13 @@ sed "${sed_no_backup[@]}" '1i\
 # Include all submodules from the repository.
 # Use while-read loop for portability (dirname -z is GNU-specific, not available on macOS)
 git ls-files -z ':(glob)**/go.mod' | while IFS= read -r -d '' modfile; do
-    go work edit -use "$(dirname "$modfile")"
+    moddir=$(dirname "$modfile")
+    # tools/sbom is intentionally outside the workspace: syft pulls in deps
+    # that conflict with the workspace-wide versions enforced by verify-dep.
```
ahrtr, line 54: This isn't specific to syft, all tools defined in tools/mod have the same "problem". So we may still want to add syft in tools/mod

ahrtr:
> `tools/mod` is part of `go.work`, and `verify-dep` (`dep_pass` in scripts/test.sh) enforces a single version per dependency across all workspace modules. `syft v1.45.1` pulls ~280 transitive modules, and several of them require versions newer than the rest of etcd pins:

We should manage all tools in a consistent way. I think we have two options:
- move `tools/mod` out of `go.work`, and add syft into `tools/mod`
- add syft into `tools/mod`, and ensure we have consistent version for all deps

either way, we need to add syft into `tools/mod`. cc @ivanvc @vivekpatani

ahrtr:
I was curious why the two BOM files (`sbom.spdx.json` and `sbom.cdx.json`) are so huge files. After looking into the contents of the two files, I understand that for each dependency syft generate an item for each etcd module. 

For example, `github.com/spf13/cobra` is used by 6 etcd modules (not including the modules under `tools`), so there are 6 items for such single dependency in both files. Since etcd already guarantees that all etcd modules have consistent dependency versions, so I think it doesn't make sense to have duplicated items for each dependency.

Also for github action components such as `actions/checkout`, there are also multiple entries (such as 6 entries for this example) in both files, because it's used in 6 github workflow yaml files. Again, it doesn't make sense to have duplicated items.

@vivekpatani can you please resolve this to remove the duplicated items for each dependency?

vivekpatani:
> We should manage all tools in a consistent way. I think we have two options:
move tools/mod out of go.work, and add syft into tools/mod
add syft into tools/mod, and ensure we have consistent version for all deps
either way, we need to add syft into tools/mod

Agreed on managing tools consistently. Quick take on the two options:
- Option 2 (add to `tools/mod` + force consistent versions): I'd push back on this one. I tested it. Making `verify-dep` pass requires `go work sync`, which rewrites `server/go.mod` and the root module onto syft's pre-release google.golang.org/protobuf v1.36.12-0.2026 (we're on the released v1.36.11). That ties the server's protobuf version to an SBOM tool, and every syft bump could churn production deps. To reproduce:

```bash
( cd tools/mod && go get github.com/anchore/syft@v1.45.1 ) && go work edit -go 1.26.3 && go work sync
grep protobuf server/go.mod   # now v1.36.12-pre; 
```
- Option 1 (move `tools/mod` out of `go.work` + add syft): viable, but a wider blast radius. It does fix the conflict (verify-dep stops governing `tools/mod`, and the go 1.26.3 vs go.work go 1.26 issue goes away), and it would let us install syft via `run_go_tool` like the other tools, which addresses @ivanvc earlier point. But it changes dependency governance for every tool in tools/mod, likely needs `run_go_tool`/`tool_get_bin` to run with `GOWORK=off`, and could shift the other tools' shared deps via MVS inside tools/mod. I'd want to validate those ripples first.

For context, tools/sbom as it stands is basically Option 1 scoped to just syft: out of the `workspace`, `GOWORK=off`, no impact on the other tools, and it still gets Dependabot tracking through its own stanza.

I'm happy either way. If you'd prefer a single tools module, I'll take Option 1 and validate the tool-resolution and MVS ripples. If the scoped module is acceptable, I'll keep it. What's your preference? @ahrtr @ivanvc

vivekpatani:
> Q (@ahrtr): "syft generates an item for each dependency per etcd module (cobra has 6 items), and multiple entries for github actions like actions/checkout (6 entries). Since etcd guarantees consistent versions, duplicates don't make sense. Can you remove the duplicated items?"

Good catch, confirmed it. The dir scan catalogs each module separately, so a dep shared by N modules shows up N times. SPDX is 628 packages but only 115 unique name@version (82% redundant), CDX is the same (627 to 114). cobra is 6x, actions/checkout is 6x, and the test-only deps (testify, go-spew, yaml.v3) are 10x since they sit in all 10 modules.

Two ways to fix it:
1. `jq` dedup on the current source scan: collapse to one entry per name@version and rewrite the SPDX relationships / CDX dependencies to the canonical refs. Keeps the existing deterministic pipeline, just need to avoid orphaning any relationship refs.
2. Scan the built binaries instead of dir:.: this dedups naturally and drops the test-only deps that shouldn't be in a release SBOM anyway. Cleaner, but a bigger change, and I'd re-confirm determinism holds for binary scans.

I'm leaning toward (1) for this PR since it's the smaller change and preserves determinism, with (2) as a possible follow-up. Sound good? @ahrtr

ahrtr:
> Option 1 (move tools/mod out of go.work + add syft): viable, but a wider blast radius.

Can you explain why there is a wider blast radius?



> I'm happy either way. If you'd prefer a single tools module, I'll take Option 1 and validate the tool-resolution and MVS ripples. If the scoped module is acceptable, I'll keep it. What's your preference? @ahrtr @ivanvc

I'd defer to you and @ivanvc to work out the best solution. We need to setup the requirement first,
- we should have consistent way to mange all tool
- We should ensure all etcd production modules have consistent versions of deps. 
- Do we need to ensure the deps used by tools have consistent version as production modules? It seems that we can simplify thing if we don't require this?
- should we get all dependencies used by tool included in the BOM files? How does K8s handle it?




> Two ways to fix it:
> 
> 1. `jq` dedup on the current source scan: collapse to one entry per name@version and rewrite the SPDX relationships / CDX dependencies to the canonical refs. Keeps the existing deterministic pipeline, just need to avoid orphaning any relationship refs.
> 2. Scan the built binaries instead of dir:.: this dedups naturally and drops the test-only deps that shouldn't be in a release SBOM anyway. Cleaner, but a bigger change, and I'd re-confirm determinism holds for binary scans.
> 
> I'm leaning toward (1) for this PR since it's the smaller change and preserves determinism, with (2) as a possible follow-up. Sound good? @ahrtr

These seem workaround solutions. Does syft have native way to handle this?

vivekpatani:
> Can you explain why there is a wider blast radius?

It is the proto toolchain. `protoc-gen-go` ships in the same module as `google.golang.org/protobuf`, and it is the binary `genproto.sh` builds via `tool_get_bin` (scripts/genproto.sh:80). Adding syft to `tools/mod` forces that module's protobuf to syft's pre-release `v1.36.12`, so `protoc-gen-go`'s stamp flips from `v1.36.11` to `v1.36.11-devel`, and regen would restamp all 12 committed `*.pb.go` (verify-genproto fails). Repro:

  ```bash
  WT=$(mktemp -d); git worktree add -q --detach "$WT" HEAD
  GOWORK=off go -C "$WT/tools/mod" get github.com/anchore/syft@v1.45.1
  GOBIN="$WT/bin" GOWORK=off go -C "$WT/tools/mod" install google.golang.org/protobuf/cmd/protoc-gen-go
  "$WT/bin/protoc-gen-go" --version   # v1.36.11-devel  (repo's *.pb.go are stamped v1.36.11)
  git worktree remove --force "$WT"
  ```

> Do we need to ensure the deps used by tools have consistent version as production modules? It seems that we can simplify things if we don't require this?

No, we do not need to. Tools are build-time only and never ship, so their deps do not have to match the production modules. That is exactly what makes isolating syft acceptable: its deps conflict with the workspace, but since tool deps are not required to be consistent with production, keeping it out of `go.work` is correct, not a workaround. Tool management stays consistent (go install + Dependabot, the etcd convention); only the module location differs.

> should we get all dependencies used by tool included in the BOM files? How does K8s handle it?

No. The BOM should describe what ships, not the build tools, so tool deps are excluded (the script already excludes `./tools/**`). k8s does the same: it uses its own tool `bom` (`sigs.k8s.io/bom`, not syft) to scan the source repo for the release SBOM, and does not put build-tool deps in the product SBOM.

> These seem workaround solutions. Does syft have native way to handle this?

Not via a flag. syft has no dedup/merge option, and `syft edit` is still an open feature request (anchore/syft#3485,anchore/syft#4461); its only built-in dedup is binary-vs-OS overlap, which does not apply to a Go source tree. The native way to avoid the per-module duplication is to scan the built binaries instead of `dir:.` (one module graph per binary, deduplicated by construction):

| scan | total | unique | duplicates |
|---|---|---|---|
| `dir:.` (current) | 628 | 115 | 513 |
| binaries (etcd 60, etcdctl 38, etcdutl 57) | 153 | 78 | 0 per-binary |

The binary scan still captures every shipped dep (grpc, protobuf, bbolt, raft, zap, otel) and drops the test-only and CI deps that do not ship.

ahrtr:
> That is exactly what makes isolating syft acceptable

OK.

> The native way to avoid the per-module duplication is to scan the built binaries

Wouldn't that have the platform-dependent problem introduced at build time due to build tag such as `//go:build cgo && amd64`, as mentioned in https://github.com/etcd-io/etcd/issues/18902#issuecomment-4353763662? Also if we scan binaries, does it mean we ignore all dependencies used by e2e test?

> ([anchore/syft#3485](https://github.com/anchore/syft/issues/3485),[anchore/syft#4461](https://github.com/anchore/syft/issues/4461))

I see jq is recommended in https://github.com/anchore/syft/issues/4324#issuecomment-3470514039, so I am not against your previous proposal of using jq.

vivekpatani:
> Wouldn't that have the platform-dependent problem introduced at build time due to build tag such as //go:build cgo && amd64, as mentioned in https://github.com/etcd-io/etcd/issues/18902#issuecomment-4353763662? Also if we scan binaries, does it mean we ignore all dependencies used by e2e test?

about the platform bit, pinning the build (`CGO_ENABLED=0 GOOS=linux GOARCH=amd64 -trimpath`) makes it reproducible on any host, so the build-tag concern is avoidable.

But your second point is a valid concern, a binary scan drops the test/e2e deps (`testify`, `antithesis-sdk-go`, `gofail`,...) that `bill-of-materials.json` tracks, so the scan stays `dir:..` I've implemented the `jq` dedup on the `dir:.` output: SPDX 628 -> 114 unique, CycloneDX 627 -> 114, GitHub Actions collapsed too, 0 duplicates. 

Comparison and raw list are in the [gist](https://gist.github.com/vivekpatani/8075f444870569c8182c1be78a994b7a).

deps.txt
ivanvc, line 1: Do we need to keep this file? I don't see it used by anything else.

vivekpatani: Sorry, my bad, I generated that for the gist. Forgot to delete it before updating PR, removed.

ivanvc:
> Good catch, confirmed it. The dir scan catalogs each module separately, so a dep shared by N modules shows up N times. SPDX is 628 packages but only 115 unique name@version (82% redundant), CDX is the same (627 to 114). cobra is 6x, actions/checkout is 6x, and the test-only deps (testify, go-spew, yaml.v3) are 10x since they sit in all 10 modules.

Is this dedup process something that other projects do?

---

Regarding the issue with the tools directory. I suggested removing `tools/mod` from the `verify-dep` check before.

I agree that we should have a better way of managing the tools. But maybe rather than introducing a new module in this PR. Could we just accept installing the tool with a fixed version from the script? Adding a new tool pattern may make it more difficult to remove it later.

Issue #12460, opened 2020-11-06
Add Distributed Tracing using OpenTelemetry


dashpole:
Prior issues/PRs related to distributed tracing: https://github.com/etcd-io/etcd/issues/11166, https://github.com/etcd-io/etcd/issues/9242, https://github.com/etcd-io/etcd/pull/5425.

In the kubernetes instrumentation sig, we are adding working to add Distributed Tracing support using OpenTelemetry to the kube-apiserver: [kubernetes/enhancements/keps/sig-instrumentation/647-apiserver-tracing](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/647-apiserver-tracing).  We plan to have Alpha (disabled by default) support for this feature in kubernetes v1.22. Progress can be tracked in https://github.com/kubernetes/enhancements/issues/647.  This will generate spans for incoming and outgoing requests, including requests to etcd.  It will also propagate the [w3c trace context](https://www.w3.org/TR/trace-context/) and [w3c baggage](https://github.com/w3c/baggage) along with outgoing requests.

This presents an opportunity to make Etcd easier to debug when using kubernetes.  Since distributed tracing provides a unified view of a single request across multiple components, a trace including the API Server and Etcd would show both the API call to the APIServer and the resulting Etcd call, providing better context on the call to Etcd.  Based on https://github.com/etcd-io/etcd/pull/5425#issuecomment-226627966, there is already interest in distributed tracing, so I won't elaborate further.

Previous attempts at distributed tracing were [looking into](https://github.com/etcd-io/etcd/pull/5425#issuecomment-360867285) OpenTracing, OpenCensus, and jaeger formats and libraries.  Thankfully, these have since unified behind the OpenTelemetry project, which makes a potential evaluation considerably easier.  OpenTelemetry will be GA soon, so now seems like the right time to open a discussion around adding distributed tracing to Etcd.

ccing people who have been involved with or interested in this topic.
@logicalhan @tedsuo @ehashman @brancz @bg451 @bhs @gyuho @jingyih @serathius

jingyih:
Today in etcd, tracing is implemented in a similar fashion as in kubernetes (https://github.com/kubernetes/utils/tree/master/trace).

/cc original author for etcd tracing: @YoyinZyc

lilic:
Hey, @dashpole 👋  I am always on board when it comes to adding more observability signals into any project and tracing is something that is very useful! With that being said, I have two concerns with adding OpenTelemetry:
[...]

PR #20307, opened 2025-07-08
Replace returned Trace with Context argument in internal Applier interface

AwesomePatrol:
Part of https://github.com/etcd-io/etcd/issues/12460.

`context.Context` argument is added to replace returning `traceutil.Trace`. This solves the problem of "merging" two traces and makes it easy to transition to nested spans possible in OpenTelemetry tracing.

Some occurrences of `context.TODO()` were also replaced with actual context.

### DIFF
--- a/server/etcdserver/apply/uber_applier.go
+++ b/server/etcdserver/apply/uber_applier.go
@@ -15,12 +15,14 @@
 package apply
 
 import (
+	"context"
 	"errors"
 	"time"
 
 	"go.uber.org/zap"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/pkg/v3/traceutil"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/v3alarm"
 	"go.etcd.io/etcd/server/v3/etcdserver/txn"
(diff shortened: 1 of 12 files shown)


server/etcdserver/v3_server.go
```diff
@@ -198,18 +196,8 @@ func (s *EtcdServer) Txn(ctx context.Context, r *pb.TxnRequest) (*pb.TxnResponse
 }
 
 func (s *EtcdServer) Compact(ctx context.Context, r *pb.CompactionRequest) (*pb.CompactionResponse, error) {
-	startTime := time.Now()
-	result, err := s.processInternalRaftRequestOnce(ctx, pb.InternalRaftRequest{Compaction: r})
 	ctx, trace := traceutil.EnsureTrace(ctx, s.Logger(), "compact")
-	if result != nil && result.Trace != nil {
```
serathius: Can you provide more context why you remove it? Don't have enough context about the code, while most code is trivial refactor, this pops out as a removal.

server/etcdserver/v3_server.go
```diff
@@ -117,7 +117,7 @@ func (s *EtcdServer) Range(ctx context.Context, r *pb.RangeRequest) (*pb.RangeRe
 				traceutil.Field{Key: "response_revision", Value: resp.Header.Revision},
 			)
 		}
-		trace.LogIfLong(traceThreshold)
+		trace.LogAllStepsIfLong(traceThreshold)
```
serathius: Ok, but why we would want to log more steps?

server/etcdserver/v3_server.go
AwesomePatrol: Re-added to simplify PR

server/etcdserver/v3_server.go
AwesomePatrol: Re-added to simplify PR

server/etcdserver/apply/uber_applier.go
```diff
@@ -130,19 +132,29 @@ func (a *uberApplier) dispatch(r *pb.InternalRaftRequest, shouldApplyV3 membersh
 	switch {
 	case r.Range != nil:
 		op = "Range"
-		ar.Resp, ar.Trace, ar.Err = a.applyV3.Range(r.Range)
+		ctx := context.Background()
```
serathius: Do you need to have separate context constructor in each switch ? Moving it outside of switch should not be a problem and will be more readable.

server/etcdserver/apply/uber_applier.go
```diff
@@ -130,19 +132,29 @@ func (a *uberApplier) dispatch(r *pb.InternalRaftRequest, shouldApplyV3 membersh
...
+		ar.Resp, ar.Err = a.applyV3.Range(ctx, r.Range)
+		ar.Trace = traceutil.Get(ctx)
```
serathius: This context will not have trace attached. Attaching a value to context creates a sub context and doesn't affect the  original context.

server/etcdserver/apply/uber_applier.go
AwesomePatrol: `EnsureTrace` takes `op` as an argument; the only other option would be adding a wrapper, but it would also add ~2 lines

server/etcdserver/apply/uber_applier.go
```diff
@@ -130,19 +132,29 @@ func (a *uberApplier) dispatch(r *pb.InternalRaftRequest, shouldApplyV3 membersh
 	switch {
 	case r.Range != nil:
 		op = "Range"
-		ar.Resp, ar.Trace, ar.Err = a.applyV3.Range(r.Range)
+		ctx, trace := traceutil.EnsureTrace(context.Background(), a.lg, op)
```
serathius, line 135: Please move `context.Background()` to outside of switch to encourage other branches to also use it.

serathius, line 135: Calling `EnsureTrace` here means the calls in apply will be skipped. It means this changes names of operations and removes fields that are passed to `EnsureTrace`.

PR #20310, opened 2025-07-08
cache: fix waitGroup goroutine registration

apullo777:

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -31,13 +31,14 @@ var ErrUnsupportedWatch = errors.New("cache: unsupported watch parameters")
 
 // Cache buffers a single etcd Watch for a given key‐prefix and fan‑outs local watchers.
 type Cache struct {
-	prefix    string // prefix is the key-prefix this shard is responsible for ("" = root).
-	cfg       Config // immutable runtime configuration
-	watcher   clientv3.Watcher
-	demux     *demux // demux fans incoming events out to active watchers and manages resync.
-	ready     chan struct{}
-	stop      context.CancelFunc
-	waitGroup sync.WaitGroup
+	prefix      string // prefix is the key-prefix this shard is responsible for ("" = root).
+	cfg         Config // immutable runtime configuration
+	watcher     clientv3.Watcher
+	demux       *demux // demux fans incoming events out to active watchers and manages resync.
+	ready       chan struct{}
+	stop        context.CancelFunc
+	waitGroup   sync.WaitGroup
+	internalCtx context.Context
 }
 
 // watchCtx collects all the knobs that both serveWatchEvents and watchRetryLoop need.
@@ -63,11 +64,12 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
 	internalCtx, cancel := context.WithCancel(context.Background())
 
 	cache := &Cache{
-		prefix:  prefix,
-		cfg:     cfg,
-		watcher: watcher,
-		ready:   make(chan struct{}),
-		stop:    cancel,
+		prefix:      prefix,
+		cfg:         cfg,
+		watcher:     watcher,
+		ready:       make(chan struct{}),
+		stop:        cancel,
+		internalCtx: internalCtx,
 	}
 
 	cache.demux = NewDemux(internalCtx, &cache.waitGroup, cfg.HistoryWindowSize, cfg.ResyncInterval)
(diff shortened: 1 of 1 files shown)


bill-of-materials.json
MadhavJivrajani, line 1: Let's split this in a separate PR

cache/cache.go
```diff
@@ -159,6 +166,7 @@ func (c *Cache) WaitReady(ctx context.Context) error {
 
 // Close cancels the private context and blocks until all goroutines return.
 func (c *Cache) Close() {
+	c.demux.Purge()
```
MadhavJivrajani: Any reason to do this here?

apullo777: I try to use demux.Purge() to stop every watcher as I noticed after the fix there is a deadlock for Close(). I guess I need to find the root cause of this deadlock and fix it.

serathius: Please remove.

apullo777: There was a deadlock where `Close()` was waiting on watcher goroutines whose contexts were never cancelled. A solution in my mind is to keep a keep a list of all per-watcher cancel functions (and a mutex) in `Cache` and call those cancel functions all at the start of `Close()`. Do you see issue with this approach or is there a more idiomatic way to orchestrate shutdown?

serathius: The Watch goroutine shutdown is very strange. It's spaghetti. I think you should fix that instead of adding purge or managing cancel functions.

Instead of having this:
```
func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption) clientv3.WatchChan {
	select {
	case <-c.ready:
	case <-ctx.Done():
		emptyWatchChan := make(chan clientv3.WatchResponse)
		close(emptyWatchChan)
		return emptyWatchChan
	}

	op := clientv3.OpGet(key, opts...)
	requestedRev := op.Rev()

	// build the internal Watch event stream (for historic reply + live)
	stream, w, err := c.newWatchEventStream(ctx, key, requestedRev, opts)
	if err != nil {
		// terminal error → one-shot reply, needs a single buffer slot
		responseChan := make(chan clientv3.WatchResponse, 1)
		if errors.Is(err, rpctypes.ErrCompacted) {
			compactRev := c.demux.PeekOldest()
			responseChan <- clientv3.WatchResponse{CompactRevision: compactRev}
		} else {
			responseChan <- clientv3.WatchResponse{Canceled: true}
		}
		close(responseChan)
		return responseChan
	}

	go func() { <-w.Done(); cancel() }()

	responseChan := make(chan clientv3.WatchResponse)
	go func() {
		defer cancel()
		defer close(responseChan)

		for event := range stream {
			select {
			case <-ctx.Done():
				return
			case responseChan <- clientv3.WatchResponse{Events: []*clientv3.Event{event}}:
			}
		}
	}()
	return responseChan
}
func (c *Cache) newWatchEventStream(
	ctx context.Context,
	key string,
	startRev int64,
	opts []clientv3.OpOption,
) (<-chan *clientv3.Event, *watcher, error) {
	// TODO: Support watch on subprefix and single key & arbitrary startRev support once we guarantee gap-free replay.
	if key != c.prefix || !clientv3.IsOptsWithPrefix(opts) || startRev != 0 {
		return nil, nil, ErrUnsupportedWatch
	}

	pred := func(k []byte) bool { return strings.HasPrefix(string(k), key) }

	w := newWatcher(c.cfg.PerWatcherBufferSize, pred)
	c.demux.Register(w, startRev)

	go func() {
		<-ctx.Done()
		c.demux.Unregister(w)
	}()

	return w.eventQueue, w, nil
}

```

It should be something like:

```
func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption) clientv3.WatchChan {
	select {
	case <-c.ready:
	case <-ctx.Done():
		emptyWatchChan := make(chan clientv3.WatchResponse)
		close(emptyWatchChan)
		return emptyWatchChan
	}

	op := clientv3.OpGet(key, opts...)
	startRev := op.Rev()

	if key != c.prefix || !clientv3.IsOptsWithPrefix(opts) || startRev != 0 {
		responseChan := make(chan clientv3.WatchResponse, 1)
		responseChan <- clientv3.WatchResponse{Canceled: true}
		close(responseChan)
		return responseChan
	}

	w := newWatcher(c.cfg.PerWatcherBufferSize, func(k []byte) bool { return strings.HasPrefix(string(k), key) })
	c.demux.Register(w, startRev)

	responseChan := make(chan clientv3.WatchResponse)
	go func() {
		defer close(responseChan)
		defer c.demux.Unregister(w)
		for {
			select {
			case <-ctx.Done():
				return
			case <-c.internalCtx.Done():
				return
			case event := <-w.eventQueue:
				select {
				case <-ctx.Done():
					return
				case <-c.internalCtx.Done():
					return
				case responseChan <- clientv3.WatchResponse{Events: []*clientv3.Event{event}}:
				}
			}
		}
	}()
	return responseChan
}

```
Basically:
* Watch cancelation shutdown should be handled in the same goroutine to make process clearer
* I passed internalCtx to pass the signal directly. We could consider using Purge in Close too, however this creates ownership problem. Who is responsible for unregistering goroutine?

apullo777: Thanks for the guidance and explanation. With this new approach I notice that we will need to: 

1. Read directly from w.eventQueue, which yields a []*clientv3.Event batch instead of single *Event.
2. Check the “ok” flag on events, ok := <-w.eventQueue so we can detect when the channel is closed and exit cleanly. 

So that goroutine will be adapted into: 

```
    c.waitGroup.Add(1)
    go func() {
        defer c.waitGroup.Done()
        defer close(responseChan)
        defer c.demux.Unregister(w)
        for {
            select {
            case <-ctx.Done():
                return
            case <-c.internalCtx.Done():
                return
            case events, ok := <-w.eventQueue:
                if !ok {
                    return
                }
                select {
                case <-ctx.Done():
                    return
                case <-c.internalCtx.Done():
                    return
                case responseChan <- clientv3.WatchResponse{Events: events}:
                }
            }
        }
    }()
```

Does that match what you had in mind?

bill-of-materials.json
apullo777: I am curious about why `make fix` suggest we remove the antithesis-sdk-go license, does that mean antithesis-sdk-go is no longer a root dependency (but was a root dependency before)?

cache/cache.go
serathius: Right, makes sense.

serathius (approved):

PR #20312, opened 2025-07-09
Inject antithesis assertion in place of etcd verify.Verify

### LINKED ISSUE #20283: [Anithesis] Inject assertions in place etcd verify.

Antithesis assertions can really help with finding bugs using their platform.
Etcd has it's own verify assertions, it would be great if we could add Antithesis assertions there too.

Problem is that we don't want etcd server to take Antithesis as a direct dependency. So similarly to gofail we would like to have some option to on demand be able to rewrite etcd verify to antithesis assertions. 

/cc @henrybear327 @joshjms @nwnt @ahrtr 

### Why is this needed?

Adding Antithesis assertions for watch in https://github.com/etcd-io/etcd/pull/20281 has improved reproducability of the  issue and surfaced other issues like panics.

This could drive adding more verify statements into etcd code.

nwnt:
Fix #20283 

Background: we want to be able to dynamically inject Antithesis tests without explicitly make etcd depend on Antithesis's library.

Proposed solution: the current `client`'s `verify.Verify` is adjusted so that:
- it eventually calls the function that will be replaced by Antithesis's `assert.Always` during Antithesis test build time.
- in other cases, the function falls back to `panic` which is the existing behavior of all references of `verify.Verify` and `verify.Assert`
- it retains the behavior of `verify.Verify` where no work is done when `verify.Verify` is not enabled - no work required for the evaluation will be done.
- All references to the existing `verify.Assert` are converted to `verify.Verify`.  `verify.Assert` is then removed.

### DIFF
--- a/client/pkg/verify/verify.go
+++ b/client/pkg/verify/verify.go
@@ -66,9 +66,18 @@ func DisableVerifications() func() {
 
 // Verify performs verification if the assertions are enabled.
 // In the default setup running in tests and skipped in the production code.
-func Verify(f func()) {
+func Verify(msg string, f VerifyFunc) {
 	if IsVerificationEnabled(envVerifyValueAssert) {
-		f()
+		ok, details := f()
+		verifier(ok, msg, details)
+	}
+}
+
+type VerifyFunc func() (condition bool, details map[string]any)
+
+func verifier(condition bool, msg string, details map[string]any) {
+	if !condition {
+		panic(fmt.Sprintf("%s. details: %v.", msg, details))
 	}
 }
 
--- a/client/v3/client.go
+++ b/client/v3/client.go
@@ -201,10 +201,8 @@ func (c *Client) Sync(ctx context.Context) error {
 	}
 	// The linearizable `MemberList` returned successfully, so the
 	// endpoints shouldn't be empty.
-	verify.Verify(func() {
-		if len(eps) == 0 {
-			panic("empty endpoints returned from etcd cluster")
-		}
+	verify.Verify("empty endpoints returned from etcd cluster", func() (bool, map[string]any) {
+		return len(eps) > 0, nil
 	})
 	c.SetEndpoints(eps...)
 	c.lg.Debug("set etcd endpoints by autoSync", zap.Strings("endpoints", eps))
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1163,18 +1163,22 @@ func (s *EtcdServer) NewUberApplier() apply.UberApplier {
 }
 
 func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
-	verify.Verify(func() {
-		if cindex != snapshot.Metadata.Index {
-			panic(fmt.Sprintf("consistent_index(%d) isn't equal to snapshot index (%d)", cindex, snapshot.Metadata.Index))
-		}
+	verify.Verify("consistent_index isn't equal to snapshot index", func() (bool, map[string]any) {
+		return cindex == snapshot.Metadata.Index,
+			map[string]any{
+				"consistent_index": cindex,
+				"snapshot_index":   snapshot.Metadata.Index,
+			}
 	})
 }
 
-func verifyConsistentIndexIsLatest(lg *zap.Logger, snapshot raftpb.Snapshot, cindex uint64) {
-	verify.Verify(func() {
-		if cindex < snapshot.Metadata.Index {
-			lg.Panic(fmt.Sprintf("consistent_index(%d) is older than snapshot index (%d)", cindex, snapshot.Metadata.Index))
-		}
+func verifyConsistentIndexIsLatest(snapshot raftpb.Snapshot, cindex uint64) {
+	verify.Verify("consistent_index is older than snapshot_index", func() (bool, map[string]any) {
+		return cindex >= snapshot.Metadata.Index,
+			map[string]any{
+				"consistent_index": cindex,
+				"snapshot_index":   snapshot.Metadata.Index,
+			}
 	})
 }
 
@@ -2147,7 +2151,7 @@ func (s *EtcdServer) snapshot(ep *etcdProgress, toDisk bool) {
 	}
 	ep.memorySnapshotIndex = ep.appliedi
 
-	verifyConsistentIndexIsLatest(lg, snap, s.consistIndex.ConsistentIndex())
+	verifyConsistentIndexIsLatest(snap, s.consistIndex.ConsistentIndex())
 
 	if toDisk {
 		// SaveSnap saves the snapshot to file and appends the corresponding WAL entry.
--- a/server/storage/backend/verify.go
+++ b/server/storage/backend/verify.go
@@ -15,7 +15,6 @@
 package backend
 
 import (
-	"fmt"
 	"runtime/debug"
 	"strings"
 
@@ -72,9 +71,9 @@ func insideUnittest() bool {
 
 // VerifyBackendConsistency verifies data in ReadTx and BatchTx are consistent.
 func VerifyBackendConsistency(b Backend, lg *zap.Logger, skipSafeRangeBucket bool, bucket ...Bucket) {
-	verify.Verify(func() {
+	verify.Verify("bucket data mismatch", func() (bool, map[string]any) {
 		if b == nil {
-			return
+			return true, nil
 		}
 		if lg != nil {
 			lg.Debug("verifyBackendConsistency", zap.Bool("skipSafeRangeBucket", skipSafeRangeBucket))
@@ -87,12 +86,15 @@ func VerifyBackendConsistency(b Backend, lg *zap.Logger, skipSafeRangeBucket boo
 			if skipSafeRangeBucket && bkt.IsSafeRangeBucket() {
 				continue
 			}
-			unsafeVerifyTxConsistency(b, bkt)
+			if ok, details := unsafeVerifyTxConsistency(b, bkt); !ok {
+				return false, details
+			}
 		}
+		return true, nil
 	})
 }
 
-func unsafeVerifyTxConsistency(b Backend, bucket Bucket) {
+func unsafeVerifyTxConsistency(b Backend, bucket Bucket) (bool, map[string]any) {
 	dataFromWriteTxn := map[string]string{}
 	b.BatchTx().UnsafeForEach(bucket, func(k, v []byte) error {
 		dataFromWriteTxn[string(k)] = string(v)
@@ -104,6 +106,12 @@ func unsafeVerifyTxConsistency(b Backend, bucket Bucket) {
 		return nil
 	})
 	if diff := cmp.Diff(dataFromWriteTxn, dataFromReadTxn); diff != "" {
-		panic(fmt.Sprintf("bucket %s data mismatch\nwrite TXN: %v\nread TXN: %v\ndiff: %s", bucket.String(), dataFromWriteTxn, dataFromReadTxn, diff))
+		return false, map[string]any{
+			"bucket":    bucket.String(),
+			"write TXN": dataFromWriteTxn,
+			"read TXN":  dataFromReadTxn,
+			"diff":      diff,
+		}
 	}
+	return true, nil
 }
--- a/server/storage/backend/tx_buffer.go
+++ b/server/storage/backend/tx_buffer.go
@@ -17,7 +17,6 @@ package backend
 import (
 	"bytes"
 	"encoding/hex"
-	"fmt"
 	"sort"
 
 	"go.etcd.io/etcd/client/pkg/v3/verify"
@@ -56,17 +55,19 @@ func (txw *txWriteBuffer) put(bucket Bucket, k, v []byte) {
 func (txw *txWriteBuffer) putSeq(bucket Bucket, k, v []byte) {
 	// putSeq is only be called for the data in the Key bucket. The keys
 	// in the Key bucket should be monotonically increasing revisions.
-	verify.Verify(func() {
+	verify.Verify("Broke the rule of monotonically increasing", func() (bool, map[string]any) {
 		b, ok := txw.buckets[bucket.ID()]
 		if !ok || b.used == 0 {
-			return
+			return true, nil
 		}
-
 		existingMaxKey := b.buf[b.used-1].key
 		if bytes.Compare(k, existingMaxKey) <= 0 {
-			panic(fmt.Sprintf("Broke the rule of monotonically increasing, existingMaxKey: %s, currentKey: %s",
-				hex.EncodeToString(existingMaxKey), hex.EncodeToString(k)))
+			return false, map[string]any{
+				"existingMaxKey": hex.EncodeToString(existingMaxKey),
+				"currentKey":     hex.EncodeToString(k),
+			}
 		}
+		return true, nil
 	})
 	txw.putInternal(bucket, k, v)
 }
--- a/server/storage/mvcc/kvstore.go
+++ b/server/storage/mvcc/kvstore.go
@@ -459,9 +459,10 @@ func restoreIntoIndex(lg *zap.Logger, idx index) (chan<- revKeyValue, <-chan int
 			}
 
 			rev := BytesToRev(rkv.key)
-			verify.Verify(func() {
-				if rev.Main < currentRev {
-					panic(fmt.Errorf("revision %d shouldn't be less than the previous revision %d", rev.Main, currentRev))
+			verify.Verify("revision shouldn't be less than the previous revision", func() (bool, map[string]any) {
+				return rev.Main >= currentRev, map[string]any{
+					"revision":          rev.Main,
+					"previous revision": currentRev,
 				}
 			})
 			currentRev = rev.Main
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -15,7 +15,6 @@
 package mvcc
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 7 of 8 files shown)


server/etcdserver/server.go
```diff
@@ -1163,11 +1163,9 @@ func (s *EtcdServer) NewUberApplier() apply.UberApplier {
...
 
 func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
-	verify.Verify(func() {
-		if cindex != snapshot.Metadata.Index {
-			panic(fmt.Sprintf("consistent_index(%d) isn't equal to snapshot index (%d)", cindex, snapshot.Metadata.Index))
-		}
-	})
+	verify.VerifyPrototype(func() bool {
+		return cindex != snapshot.Metadata.Index
+	}, fmt.Sprintf("consistent_index(%d) isn't equal to snapshot index (%d)", cindex, snapshot.Metadata.Index), nil)
```
serathius: We don't want to render `fmt.Sprintf` not construct `details map[string]any` if `IsVerificationEnabled` is false.

serathius:
This is a good direction, we just need to decide on interface for verify.

server/etcdserver/server.go
nwnt: How about this newly pushed one?  By returning them in the func, the string should not be rendered and the map won't be constructed prematurely.

client/pkg/verify/verify.go
```diff
@@ -78,3 +79,16 @@ func Assert(condition bool, msg string, v ...any) {
 		panic(fmt.Sprintf("assertion failed: "+msg, v...))
 	}
 }
+
+func VerifyPrototype(p func() (bool, string, map[string]any)) {
```
serathius: Create a type for function `p`, like:
```
type VerifyFunc func() (ok bool, name string, details map[string]any)
```

server/etcdserver/server.go
```diff
@@ -1163,10 +1163,8 @@ func (s *EtcdServer) NewUberApplier() apply.UberApplier {
 }
 
 func verifySnapshotIndex(snapshot raftpb.Snapshot, cindex uint64) {
-	verify.Verify(func() {
-		if cindex != snapshot.Metadata.Index {
-			panic(fmt.Sprintf("consistent_index(%d) isn't equal to snapshot index (%d)", cindex, snapshot.Metadata.Index))
-		}
+	verify.VerifyPrototype(func() (bool, string, map[string]any) {
+		return cindex != snapshot.Metadata.Index, fmt.Sprintf("consistent_index(%d) isn't equal to snapshot index (%d)", cindex, snapshot.Metadata.Index), nil
```
serathius: I'm thinking that it would be easier to use static string and only allow passing dynamic values via details.

serathius: This way it would be:
```
	verify.VerifyPrototype("consistent_index should equal snapshot_index", func() (bool, map[string]any) {
		return cindex != snapshot.Metadata.Index, map[string]any{
			"consistent_index": cindex,
			"snapshot_index": snapshot.Metadata.Index,
		}
	})
```

nwnt:
Ready for the full review now @serathius

server/etcdserver/server.go
```diff
@@ -1163,18 +1163,22 @@ func (s *EtcdServer) NewUberApplier() apply.UberApplier {
...
+				"consistent_index": cindex,
+				"snapshot_index":   snapshot.Metadata.Index,
+			}
 	})
 }
 
 func verifyConsistentIndexIsLatest(lg *zap.Logger, snapshot raftpb.Snapshot, cindex uint64) {
-	verify.Verify(func() {
-		if cindex < snapshot.Metadata.Index {
-			lg.Panic(fmt.Sprintf("consistent_index(%d) is older than snapshot index (%d)", cindex, snapshot.Metadata.Index))
```
nwnt, line 1176: Looks like this is strictly required by robustness test.  Removing this is causing the test failure.  

Should we find a way to add this back or adjust the robustness test @serathius?

client/pkg/verify/verify.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	// import_placeholder
```
serathius: Please make it more clear reason to have placeholder and why it should be removed.
Like 
```
// NO NOT REMOVE; used to replace panic with Antithesis assertions to validate etcd on their platform.
// antithesis_import_assert_placeholder
```

client/pkg/verify/verify.go
```diff
@@ -66,9 +67,10 @@ func DisableVerifications() func() {
 
 // Verify performs verification if the assertions are enabled.
 // In the default setup running in tests and skipped in the production code.
-func Verify(f func()) {
+func Verify(msg string, f VerifyFunc) {
 	if IsVerificationEnabled(envVerifyValueAssert) {
-		f()
+		ok, details := f()
+		verifier(ok, msg, details)
```
serathius, line 72: Please leave a comment  that would document that `verifier` name should not be changed as it is replaced with Antithesis assertions when validating etcd on their platform.

tests/antithesis/server/Dockerfile
```diff
@@ -6,6 +6,15 @@ FROM golang:$GO_VERSION
 # cloning etcd
 ARG REF=main
 RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etcd
+
+# replace verify with antithesis
+WORKDIR /etcd/client/pkg/verify
+RUN if [ -f verify.go ]; then \
+        sed -i 's|// import_placeholder|"github.com/antithesishq/antithesis-sdk-go/assert"|' ./verify.go; \
```
serathius: Please make sure that `sed` fails if the replacement fails.

client/pkg/verify/verify.go
```diff
@@ -64,17 +67,36 @@ func DisableVerifications() func() {
 	}
 }
 
+// Assert will panic with a given formatted message if the given condition is false.
+func Assert(condition bool, msg string, v ...any) {
```
serathius: We should replace all calls to `Assert` with verify and remove it.

client/pkg/verify/verify.go
```diff
@@ -64,17 +67,36 @@ func DisableVerifications() func() {
...
+
+func verifier(condition bool, msg string, details map[string]any) {
+	f := func() { panic(fmt.Sprintf("%s. details: %v.", msg, details)) }
+	if postAction, ok := details[PostActionFunc]; ok {
+		if action, ok := postAction.(func()); ok {
+			f = action
+		}
+	}
+	if condition {
+		f()
```
serathius, line 100: Please revert.

client/pkg/verify/verify.go
```diff
@@ -64,17 +67,36 @@ func DisableVerifications() func() {
...
+
 // Verify performs verification if the assertions are enabled.
 // In the default setup running in tests and skipped in the production code.
-func Verify(f func()) {
+func Verify(msg string, f VerifyFunc) {
 	if IsVerificationEnabled(envVerifyValueAssert) {
-		f()
+		ok, details := f()
+		// the function name should not be changed. It will be replaced with Antithesis assertions
+		// when validating etcd on their platform.
```
serathius:
```suggestion
		// DO NOT CHANGE; the line below is replaced with Antithesis assertions  when validating etcd on their platform.
		// antithesis_verifier_placeholder
```

client/pkg/verify/verify.go
```diff
@@ -18,6 +18,9 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	// DO NOT REMOVE; will be replaced by the Antithesis assertion package when
+	// building etcd-server for running on the Antithesis platform.
```
serathius:
```suggestion
	// DO NOT REMOVE; will be replaced by the Antithesis assertion package when building etcd-server for running on the Antithesis platform.
```

client/pkg/verify/verify.go
serathius: Please revert to what was in previous commit.

nwnt: Yep, one moment.  I thought the `lg.Panic` was required but that seems not to be the case.  Will be reverting it.

tests/antithesis/server/Dockerfile
```diff
@@ -6,6 +6,15 @@ FROM golang:$GO_VERSION
 # cloning etcd
 ARG REF=main
 RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etcd
+
+# replace verify with antithesis
+WORKDIR /etcd/client/pkg/verify
+RUN if [ -f verify.go ]; then \
+        sed -i 's|// antithesis_import_assert_placeholder|"github.com/antithesishq/antithesis-sdk-go/assert"|' ./verify.go && \
+        sed -i '0,/verifier/{s/verifier/assert.Always/}' ./verify.go; \
```
serathius: Please include replacing placeholder. We don't want to replace any other random `verifier` string.

nwnt:
This is not working yet.  Will tag you once it's ready.

client/pkg/verify/verify.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	// DO NOT REMOVE; will be replaced by the Antithesis assertion package when building etcd-server for running on the Antithesis platform.
```
serathius:
```suggestion
	// DO NOT CHANGE LINE BELOW; it is replaced with Antithesis assertions when validating etcd on their platform
```

client/pkg/verify/verify.go
```diff
@@ -66,15 +68,19 @@ func DisableVerifications() func() {
 
 // Verify performs verification if the assertions are enabled.
 // In the default setup running in tests and skipped in the production code.
-func Verify(f func()) {
+func Verify(msg string, f VerifyFunc) {
 	if IsVerificationEnabled(envVerifyValueAssert) {
-		f()
+		ok, details := f()
+		// DO NOT CHANGE; the line below is replaced with Antithesis assertions when validating etcd on their platform.
```
serathius:
```suggestion
		// DO NOT CHANGE LINES BELOW; they are replaced with Antithesis assertions when validating etcd on their platform
```

client/pkg/verify/verify.go
```diff
@@ -66,15 +68,19 @@ func DisableVerifications() func() {
...
+		// antithesis_verifier_placeholder
+		verifier(ok, msg, details)
```
serathius:
```suggestion
		// antithesis_verifier_placeholder_start
		verifier(ok, msg, details)
		// antithesis_verifier_placeholder_end
```
Maybe wrap the replaced lines to make it clear.

client/pkg/verify/verify.go
```diff
@@ -66,15 +68,20 @@ func DisableVerifications() func() {
 
 // Verify performs verification if the assertions are enabled.
 // In the default setup running in tests and skipped in the production code.
-func Verify(f func()) {
+func Verify(msg string, f VerifyFunc) {
 	if IsVerificationEnabled(envVerifyValueAssert) {
-		f()
+		ok, details := f()
+		// DO NOT CHANGE LINE BELOW; it is replaced with Antithesis assertions when validating etcd on their platform.
```
serathius:
```suggestion
		// DO NOT CHANGE LINES BELOW; it is replaced with Antithesis assertions when validating etcd on their platform.
```

nwnt:
@serathius should be ready now.  the tests were failing because `verify.Assert`'s condition is the negate of `verify.Verify`.

serathius:
> should be ready now. the tests were failing because verify.Assert's condition is the negate of verify.Verify.

We should not negate the condition and the tests should pass.

nwnt:
> > should be ready now. the tests were failing because verify.Assert's condition is the negate of verify.Verify.
> 
> We should not negate the condition and the tests should pass.

Assert actually does a negation on the condition:
```go
// Assert will panic with a given formatted message if the given condition is false.
func Assert(condition bool, msg string, v ...any) {
	if !condition {
		panic(fmt.Sprintf("assertion failed: "+msg, v...))
	}
}
```
We need the negation on the original condition to work with the new verifier:
```go
func verifier(condition bool, msg string, details map[string]any) {
	if condition {
		panic(fmt.Sprintf("%s. details: %v.", msg, details))
	}
}
```

nwnt:
Without the negation, the tests would fail: https://github.com/etcd-io/etcd/commit/a66e99bc4b31ae05a9f4d42d12add9acbaad1358

client/pkg/verify/verify.go
```diff
@@ -66,15 +68,20 @@ func DisableVerifications() func() {
...
 }
 
-// Assert will panic with a given formatted message if the given condition is false.
-func Assert(condition bool, msg string, v ...any) {
-	if !condition {
-		panic(fmt.Sprintf("assertion failed: "+msg, v...))
+type VerifyFunc func() (condition bool, details map[string]any)
+
+func verifier(condition bool, msg string, details map[string]any) {
+	if condition {
```
serathius:
```suggestion
	if !condition {
```

client/pkg/fileutil/fileutil.go
```diff
@@ -47,7 +47,7 @@ func IsDirWriteable(dir string) error {
 // TouchDirAll is similar to os.MkdirAll. It creates directories with 0700 permission if any directory
 // does not exists. TouchDirAll also ensures the given directory is writable.
 func TouchDirAll(lg *zap.Logger, dir string) error {
-	verify.Assert(lg != nil, "nil log isn't allowed")
+	verify.Verify("nil log isn't allowed", func() (bool, map[string]any) { return lg == nil, nil })
```
serathius, line 50: When you assert or verify, you `assert true`.

serathius:
```suggestion
	verify.Verify("nil log isn't allowed", func() (bool, map[string]any) { return lg != nil, nil })
```
When you assert or verify, you assert true.

client/pkg/verify/verify.go
serathius: When you assert or verify, you assert true. so you should panic when false.

nwnt:
I have made `Verify` to be like `Asssert` now.

serathius:
Please fix tests.

client/v3/client.go
```diff
@@ -201,10 +201,8 @@ func (c *Client) Sync(ctx context.Context) error {
 	}
 	// The linearizable `MemberList` returned successfully, so the
 	// endpoints shouldn't be empty.
-	verify.Verify(func() {
-		if len(eps) == 0 {
-			panic("empty endpoints returned from etcd cluster")
-		}
+	verify.Verify("empty endpoints returned from etcd cluster", func() (bool, map[string]any) {
```
serathius, line 204:
```suggestion
	verify.Verify("etcd cluster endpoints are not empty", func() (bool, map[string]any) {
```

nwnt:
Yep, fixing now.  Sorry, I committed and pushed just before going to bed..

server/storage/backend/verify.go
```diff
@@ -87,12 +86,15 @@ func VerifyBackendConsistency(b Backend, lg *zap.Logger, skipSafeRangeBucket boo
 			if skipSafeRangeBucket && bkt.IsSafeRangeBucket() {
 				continue
 			}
-			unsafeVerifyTxConsistency(b, bkt)
+			if ok, details := unsafeVerifyTxConsistency(b, bkt); !ok {
+				return ok, details
```
serathius:
```suggestion
				return false, details
```

ahrtr:
- The existing `verify.Assert` is used in production code. It means it has impact on users or production. But `verify.Verify` is only enabled/used in test; its purpose is to identify any potential issues. I suggest we don't change this. But for each case (whether to use`verify.Assert` or `verify.Verify`), it's open to discussion.
- For Antithesis test, based on the change in `tests/antithesis/server/Dockerfile`, you just need to call `assert.Always(...)`, I don't see any technical blocker to reuse the existing `verify.Verify`.
  - I am not sure whether it's appropriate to depend on a commercial solution in etcd's production code, although the sdk is MIT license.  It might be OK to add Antithesis specific code & implementation under [tests/antithesis](https://github.com/etcd-io/etcd/tree/main/tests/antithesis)
  - Reference: https://www.linuxfoundation.org/legal/antitrust-policy

serathius:
> * The existing `verify.Assert` is used in production code. It means it has impact on users or production. But `verify.Verify` is only enabled/used in test; its purpose is to identify any potential issues. I suggest we don't change this. But for each case (whether to use`verify.Assert` or `verify.Verify`), it's open to discussion.

Ups I might have missed that. I think we should target `verify.Verify` for the Antithesis assertions. @nwnt can you revert changes to `verify.Assert`?

Considering how trivial `verify.Assert` function is I think we should consider replacing it with panics in the code, instead of having a dedicated library function. I don't think that mixing unconditional and conditional verification is good for clarify. But that's a separate discussion. Thanks for pointing it out!

serathius:
> I am not sure whether it's appropriate to depend on a commercial solution in etcd's production code, although the sdk is MIT license. It might be OK to add Antithesis specific code & implementation under [tests/antithesis](https://github.com/etcd-io/etcd/tree/main/tests/antithesis)

I might be too paranoid, but usage of verify is throughout the etcd code, I don't want to add another dependency to our users just for our testing.

nwnt:
Reverted all the changes related to verify.Assert.

client/pkg/verify/verify.go
```diff
@@ -64,17 +66,29 @@ func DisableVerifications() func() {
 	}
 }
 
+// Assert will panic with a given formatted message if the given condition is false.
```
serathius: Revert the change in order of functions so there is no changes to Assert function.

nwnt:
done!

fuweid (approved):

serathius:
> I am not sure whether it's appropriate to depend on a commercial solution in etcd's production code, although the sdk is MIT license. It might be OK to add Antithesis specific code & implementation under [tests/antithesis](https://github.com/etcd-io/etcd/tree/main/tests/antithesis)
Reference: https://www.linuxfoundation.org/legal/antitrust-policy

Import and usage of Antithesis code is limited to `tests/antithesis` and I would prefer it doesn't touches etcdserver. That's why I proposed to dynamically replace assertions just in Antithesis Dockerfile.

serathius:
Discussed things with @ahrtr, we would like to avoid any placeholders in etcd code. That should not be a problem if we change our approach. Instead of using `sed` on placeholders, we can use `diff` and `patch`.

@nwnt Can you keep the changes to `verify.Verify` interface but remove all the placeholders and Antithesis changes in this PR? Let's just merge updated interface.

We can implement the antithesis patching in a followup PR.

ahrtr:
> we can use `diff` and `patch`.

Yes, It's better. Pls also put the `patch` under `test/antithesis`

nwnt:
done!

serathius:
Awesome work @nwnt !

ahrtr (approved):
LGTM & thx

Issue #20007, opened 2025-05-21
[Antithesis] Introduce etcd config changes similar to robustness


serathius:
### What would you like to be added?

Robustness randomizes some config like `--snapshot-count` flag from `100`, `1000`, `10'000` (guessing from memory). 
Best case we would do randomization in antithesis too, but for now let's just pick lower values for Antithesis default config.

### Why is this needed?

Make reproducing issues involving unusual configurations easier.

Brahmjot-singh1313:
Title: Use Lower Default --snapshot-count in Antithesis Config to Improve Reproducibility

Description:

What is being changed?
This PR updates the default Antithesis configuration to use lower values for the --snapshot-count flag. Instead of relying on higher default or randomized values (as seen in robustness testing), we now set this to a smaller, fixed value (e.g., 100, 1,000, or 10,000).

Why is this change needed?
Reducing the --snapshot-count value helps surface issues that typically arise under less common configurations. This change makes it easier to:

Reproduce edge cases or rare bugs locally

Avoid relying solely on robustness randomization for test coverage

Debug and iterate faster during development and CI

While full config randomization in Antithesis would be ideal long-term, this serves as a practical interim step toward better test coverage and reproducibility.

joshjms:
The robustness tests use `--snapshot-count` values with 50, 100, or 1000.

```go
// tests/robustness/scenarios/scenarios.go
baseOptions := []e2e.EPClusterOption{
	options.WithSnapshotCount(50, 100, 1000), // this
	options.WithSubsetOptions(randomizableOptions...),
	e2e.WithGoFailEnabled(true),
	// Set low minimal compaction batch limit to allow for triggering multi batch compaction failpoints.
	options.WithCompactionBatchLimit(10, 100, 1000),
	e2e.WithWatchProcessNotifyInterval(100 * time.Millisecond),
}
```

Should we use these values for Antithesis as well?

serathius:
Yes.

ivanvc:
Can we close this issue, as #20317 is merged? Or are we missing something else?

PR #20317, opened 2025-07-09
[Antithesis] Add randomization of etcd server configurations

joshjms:
Added the following randomization for Antithesis tests:
* `ETCD_SNAPSHOT_COUNT`: 50, 100, or 1000
* `ETCD_SNAPSHOT_CATCHUP_ENTRIES`: 50, 100, or 5000
* `ETCD_COMPACTION_BATCH_LIMIT`: 10, 100, or 1000

```
make antithesis-docker-compose-up
```

or to specify value:

```
CFG_ETCD_SNAPSHOT_COUNT=167 make antithesis-docker-compose-up
```

/cc @serathius 

### DIFF
--- a/tests/antithesis/config/docker-compose-1-node.yml
+++ b/tests/antithesis/config/docker-compose-1-node.yml
@@ -32,6 +32,9 @@ services:
       ETCD_INITIAL_CLUSTER: "etcd0=http://etcd0:2380"
       ETCD_INITIAL_CLUSTER_STATE: "new"
       ETCD_DATA_DIR: "/var/etcd/data"
+      ETCD_SNAPSHOT_CATCHUP_ENTRIES: 100
+      ETCD_SNAPSHOT_COUNT: 50
+      ETCD_COMPACTION_BATCH_LIMIT: 10
     user: "${USER_ID:-root}:${GROUP_ID:-root}"
     depends_on:
       init:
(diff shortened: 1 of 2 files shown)


nwnt:
LGTM

tests/antithesis/Makefile
```diff
@@ -42,7 +54,11 @@ antithesis-build-config-image: validate-node-count
 
 .PHONY: antithesis-docker-compose-up
 antithesis-docker-compose-up: validate-node-count
```
serathius: This target is just for local testing. For antithesis only the results of `antithesis-build-config-image` target matters.

tests/antithesis/Makefile
```diff
@@ -7,6 +7,18 @@ IMAGE_TAG = latest
 
 CFG_NODE_COUNT ?= 3
 
+ETCD_SNAPSHOT_COUNT_OPTIONS := 50 100 1000
```
serathius: Antithesis only accepts static docker-compose file. That's why in https://github.com/etcd-io/etcd/issues/20007#issue-3081157091 I didn't suggest randomizing values, just picking the lower ones.

I would drop this randomization to avoid implying that this randomization is executed by Antithesis.

joshjms: Hmm, I see

Issue #20182, opened 2025-06-17
A method to track ways Kubernetes utilizes Etcd APIs


AwesomePatrol:
### What would you like to be added?

Test that analyzes Kubernetes traffic to drive work on robustness tests.

Example test: https://github.com/etcd-io/etcd/pull/20181

### Why is this needed?

Given the large API surface of Etcd I would like to introduce a method to track what is currently used by Kubernetes. This may help improve tests by better simulating actual traffic, alert about novel use cases, and in some cases highlight the potential to reduce the used API surface.

Related to https://github.com/etcd-io/etcd/issues/15820

Doc with more information and discussion: https://docs.google.com/document/d/1dd5dRk53VN89nHTEapMYNrbhIxK9wB52vsnuqwQi8LI/edit?usp=sharing

### Work plan

- [x] Make long-running goroutines export `Trace`s from Apiserver (currently they only create `Span`s in Etcd)
- [x] Improve tracing in Etcd (continue migration to OTEL, tracked as [\#12460](https://github.com/etcd-io/etcd/issues/12460))
- [x] Improve tracing in Kubernetes (apply the contract decorator and deprecate bespoke tracing solution, tracked in [KEP-647](https://github.com/kubernetes/enhancements/issues/647))
- [x] Automate:
  - [x] Set up the cluster (similar to [the integration test](https://github.com/kubernetes/kubernetes/tree/v1.33.0/test/integration/apiserver/tracing))
   - [x] Download and analyze traces (using Jaeger/[OTEL library](https://opentelemetry.io/docs/languages/go/))
- [ ] Run more complex setup and/or explore ways to exercise more code paths
   - [ ] Toggle experimental features
- [ ] Expand Etcd robustness tests to better simulate Kubernetes’ traffic

/cc @serathius

serathius:
High level I like the idea of semi automatic way to track K8s etcd contract, to make sure we don't miss anything and be up to date. Will review the document.

/cc @ahrtr @fuweid

serathius:
Presented the issue during robustness test meeting asking for comment and feedback. Don't think there was any voices against the proposal, still feedback is encouraged.

ahrtr:
> Given the large API surface of Etcd I would like to introduce a method to track what is currently used by Kubernetes.

Improving etcd's tracing makes sense to me. Thanks for working on this.

serathius:
[...]

PR #20325, opened 2025-07-10
Add OpenTelemetry Spans to Range and Txn

AwesomePatrol:

This adds new spans in `Range` and `Txn` methods. This makes it easier to distinguish actual execution time from the overhead of grpc and other pre/post processing.

Attributes added:
1. `Range`:
   * `range_begin` (`key` field in proto), `range_end` - used to better understand key transformations from other systems (like Kubernetes)
   * `rev` - revision used to distinguish between requests for old or latest data
   * `limit` - lists vs gets
   * `count_only` (boolean)
2. `Txn`:
   * `first_compare_key` - first non-empty key from `compare`s
   * `compare_len` - the number of compare operation in the request
   * `failure_len`, `success_len` - counts for `failure` and `success` operations in the request
   * `read_only` (boolean) - if any of the operations can modify data

Example trace from Jeager with Apiserver tracing also configured:
image

### DIFF
--- a/pkg/traceutil/trace.go
+++ b/pkg/traceutil/trace.go
@@ -22,8 +22,18 @@ import (
 	"strings"
 	"time"
 
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
 	"go.uber.org/zap"
 )
+
+const instrumentationScope = "go.etcd.io/etcd"
+
+var Tracer trace.Tracer = noop.NewTracerProvider().Tracer(instrumentationScope)
+
+func Init(tp trace.TracerProvider) {
+	Tracer = tp.Tracer(instrumentationScope)
+}
 
 // TraceKey is used as a key of context for Trace.
 type TraceKey struct{}
--- a/server/etcdserver/v3_server.go
+++ b/server/etcdserver/v3_server.go
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 	"golang.org/x/crypto/bcrypt"
 
@@ -102,6 +104,16 @@ type Authenticator interface {
 }
 
 func (s *EtcdServer) Range(ctx context.Context, r *pb.RangeRequest) (*pb.RangeResponse, error) {
+	var span trace.Span
+	ctx, span = traceutil.Tracer.Start(ctx, "range", trace.WithAttributes(
+		attribute.String("range_begin", string(r.GetKey())),
+		attribute.String("range_end", string(r.GetRangeEnd())),
+		attribute.Int64("rev", r.GetRevision()),
+		attribute.Int64("limit", r.GetLimit()),
+		attribute.Bool("count_only", r.GetCountOnly()),
+	))
+	defer span.End()
+
 	ctx, trace := traceutil.EnsureTrace(ctx, s.Logger(), "range",
 		traceutil.Field{Key: "range_begin", Value: string(r.Key)},
 		traceutil.Field{Key: "range_end", Value: string(r.RangeEnd)},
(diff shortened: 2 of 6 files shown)


server/etcdserver/v3_server.go
```diff
@@ -156,8 +168,30 @@ func (s *EtcdServer) DeleteRange(ctx context.Context, r *pb.DeleteRangeRequest)
...
+	}
+	return ""
+}
+
 func (s *EtcdServer) Txn(ctx context.Context, r *pb.TxnRequest) (*pb.TxnResponse, error) {
 	readOnly := txn.IsTxnReadonly(r)
+
+	var span trace.Span
+	ctx, span = traceutil.Tracer.Start(ctx, "txn", trace.WithAttributes(
+		attribute.String("first_compare_key", firstCompareKey(r.GetCompare())),
```
serathius: Maybe `compare_first_key` to have consistent naming scheme `compare_*`?

server/etcdserver/v3_server.go
```diff
@@ -156,8 +168,30 @@ func (s *EtcdServer) DeleteRange(ctx context.Context, r *pb.DeleteRangeRequest)
...
+		attribute.Int("compare_len", len(r.GetCompare())),
```
serathius, line 188: Maybe add a type of comparison and value we compare?
Like `attribute.Int("compare_first_op", "equal")`, `attribute.Int("compare_first_source", "mod_rev")`, `attribute.Int("compare_first_value", "42")`,

server/etcdserver/v3_server.go
```diff
@@ -156,8 +168,30 @@ func (s *EtcdServer) DeleteRange(ctx context.Context, r *pb.DeleteRangeRequest)
...
+		attribute.Int("success_len", len(r.GetSuccess())),
```
serathius, line 189: Maybe add `attribute.Int("success_first_key", "key")`, `attribute.Int("success_first_op", "put")`

tests/integration/tracing_test.go
```diff
@@ -50,6 +50,28 @@ func TestTracing(t *testing.T) {
...
+			// make a request with the instrumented client
+			resp, err := cli.Get(t.Context(), "key", clientv3.WithCountOnly())
+			require.NoError(t, err)
+			require.Empty(t, resp.Kvs)
+			return nil
+		})
+	})
+
+	t.Run("UnaryTxnRPC", func(t *testing.T) {
+		testRPCTracing(t, "UnaryTxnRPC", containsTxnSpan, func(cli *clientv3.Client) error {
```
serathius, line 64: I find it hard to follow the test, as `containsTxnSpan` is a separate function, and handles all validation.


Could we maybe have function to find the relevant span, but the validation of the span contents could be written here, as part of the test code.

server/etcdserver/v3_server.go
AwesomePatrol: Renamed

server/etcdserver/v3_server.go
AwesomePatrol: There are two problems:
1. `Result` defaults to `equal`, `Target` defaults to `version` making it slightly harder to implement as we need to ensure attributes are not added when there is no `compare` (is it required?)
2. `TargetUnion` is `oneof` and with the currently generated code, there is no nice way to get a string out of it; again, not impossible to overcome, but it could be better suited for another PR

server/etcdserver/v3_server.go
AwesomePatrol: `success_first_key` is possible, `success_first_op` refers to `oneof` field making it non-trivial to implement

tests/integration/tracing_test.go
AwesomePatrol: I followed existing pattern. I will reorganize the test to make it easier to read

server/etcdserver/v3_server.go
serathius: Ok, we can move it to followup.

server/etcdserver/v3_server.go
serathius: Ok, we can move it to followup.

tests/integration/tracing_test.go
serathius: ok

serathius (approved):

serathius (approved):

pkg/traceutil/trace.go
```diff
@@ -22,9 +22,19 @@ import (
...
+	"go.opentelemetry.io/otel/trace/noop"
 	"go.uber.org/zap"
 )
 
+const instrumentationScope = "go.etcd.io/etcd"
+
+var Tracer trace.Tracer = noop.NewTracerProvider().Tracer(instrumentationScope)
+
+func Init(tp trace.TracerProvider) {
+	Tracer = tp.Tracer(instrumentationScope)
```
fuweid, line 35: maybe we should introduce syncOnce for this tracer. just a case that it could be race-condition. it's not blocker

tests/integration/tracing_test.go
fuweid: it looks like we can use table-style here to organize the test.

fuweid (approved):

pkg/traceutil/trace.go
AwesomePatrol: I tried, but unfortunately it breaks tracing integration tests (`Tracer` needs to be overwritten for each test case)

tests/integration/tracing_test.go
AwesomePatrol: Addressed in https://github.com/etcd-io/etcd/pull/20338

server/etcdserver/v3_server.go
AwesomePatrol: Followed up in https://github.com/etcd-io/etcd/pull/20511

PR #20326, opened 2025-07-10
cache: enable range, prefix, fromKey key filtering in cache watch

apullo777:

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -15,10 +15,10 @@
 package cache
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
@@ -105,15 +105,15 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	op := clientv3.OpGet(key, opts...)
 	startRev := op.Rev()
 
-	// TODO: Support watch on subprefix and single key & arbitrary startRev support once we guarantee gap-free replay.
-	if key != c.prefix || !clientv3.IsOptsWithPrefix(opts) || startRev != 0 {
-		responseChan := make(chan clientv3.WatchResponse, 1)
-		responseChan <- clientv3.WatchResponse{Canceled: true}
-		close(responseChan)
-		return responseChan
+	pred, err := c.validateWatch(key, opts...)
+	if err != nil {
+		ch := make(chan clientv3.WatchResponse, 1)
+		ch <- clientv3.WatchResponse{Canceled: true}
+		close(ch)
+		return ch
 	}
 
-	w := newWatcher(c.cfg.PerWatcherBufferSize, func(k []byte) bool { return strings.HasPrefix(string(k), key) })
+	w := newWatcher(c.cfg.PerWatcherBufferSize, pred)
 	c.demux.Register(w, startRev)
 
 	responseChan := make(chan clientv3.WatchResponse)
@@ -225,4 +225,58 @@ func readWatchChannel(
 		demux.Broadcast(resp.Events)
 	}
 	return nil
+}
+
+func (c *Cache) validateWatch(key string, opts ...clientv3.OpOption) (pred KeyPredicate, err error) {
+	op := clientv3.OpGet(key, opts...)
+	startRev := op.Rev()
+	// TODO: Support watch on arbitrary startRev support once we guarantee gap-free replay.
+	if startRev != 0 {
+		return nil, ErrUnsupportedWatch
+	}
+
+	startKey := []byte(key)
+	endKey := op.RangeBytes() // nil = single key, {0}=FromKey, else explicit range
+
+	if err := c.validateWatchRange(startKey, endKey); err != nil {
+		return nil, err
+	}
+	return KeyPredForRange(startKey, endKey), nil
+}
+
+func (c *Cache) validateWatchRange(startKey, endKey []byte) error {
+	prefixStart := []byte(c.prefix)
+	prefixEnd := []byte(clientv3.GetPrefixRangeEnd(c.prefix))
+
+	isSingleKey := len(endKey) == 0
+	isFromKey := len(endKey) == 1 && endKey[0] == 0
+
+	switch {
+	case isSingleKey:
+		if c.prefix == "" {
+			return nil
+		}
+		if bytes.Compare(startKey, prefixStart) < 0 || bytes.Compare(startKey, prefixEnd) >= 0 {
+			return ErrUnsupportedWatch
+		}
+		return nil
+
+	case isFromKey:
+		if c.prefix != "" {
+			return ErrUnsupportedWatch
+		}
+		return nil
+
+	default:
+		if bytes.Compare(endKey, startKey) <= 0 {
+			return ErrUnsupportedWatch
+		}
+		if c.prefix == "" {
+			return nil
+		}
+		if bytes.Compare(startKey, prefixStart) < 0 || bytes.Compare(endKey, prefixEnd) > 0 {
+			return ErrUnsupportedWatch
+		}
+		return nil
+	}
 }
--- a/cache/cache_test.go
+++ b/cache/cache_test.go
@@ -1,0 +1,190 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"testing"
+
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+func TestValidateWatchRange(t *testing.T) {
+	type tc struct {
+		name        string
+		watchKey    string
+		opts        []clientv3.OpOption
+		cachePrefix string
+		wantErr     bool
+	}
+
+	tests := []tc{
+		{
+			name:        "single key",
+			watchKey:    "/a",
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "prefix single key",
+			watchKey:    "/foo/a",
+			cachePrefix: "/foo",
+			wantErr:     false,
+		},
+		{
+			name:        "single key outside prefix returns error",
+			watchKey:    "/z",
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "explicit range",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/b")},
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "exact prefix range",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/b")},
+			cachePrefix: "/a",
+			wantErr:     false,
+		},
+		{
+			name:        "prefix subrange",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/foo/a")},
+			cachePrefix: "/foo",
+			wantErr:     false,
+		},
+		{
+			name:        "reverse range returns error",
+			watchKey:    "/b",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/a")},
+			cachePrefix: "",
+			wantErr:     true,
+		},
+		{
+			name:        "empty range returns error",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/foo")},
+			cachePrefix: "",
+			wantErr:     true,
+		},
+		{
+			name:        "range starting below cache prefix returns error",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/foo")},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "range encompassing cache prefix returns error",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/z")},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "range crossing prefixEnd returns error",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/z")},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "empty prefix",
+			watchKey:    "",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "empty prefix with cachePrefix returns error",
+			watchKey:    "",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "prefix watch matches cachePrefix exactly",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "/foo",
+			wantErr:     false,
+		},
+		{
+			name:        "prefix watch inside cachePrefix",
+			watchKey:    "/foo/bar",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "/foo",
+			wantErr:     false,
+		},
+		{
+			name:        "prefix starting below cachePrefix returns error",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "prefix starting above shard prefixEnd returns error",
+			watchKey:    "/fop",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "fromKey open‑ended",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithFromKey()},
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "fromKey starting at prefix start",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithFromKey()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "fromKey starting below prefixEnd",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithFromKey()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "fromKey starting above prefixEnd returns error",
+			watchKey:    "/fop",
+			opts:        []clientv3.OpOption{clientv3.WithFromKey()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+	}
+
+	for _, c := range tests {
+		t.Run(c.name, func(t *testing.T) {
+			dummyCache := &Cache{prefix: c.cachePrefix}
+			op := clientv3.OpGet(c.watchKey, c.opts...)
+			err := dummyCache.validateWatchRange([]byte(c.watchKey), op.RangeBytes())
+			if gotErr := err != nil; gotErr != c.wantErr {
+				t.Fatalf("validateWatchRange(%q, %q, %v) err=%v, wantErr=%v",
+					c.cachePrefix, c.watchKey, c.opts, err, c.wantErr)
+			}
+		})
+	}
+}
--- a/cache/predicate.go
+++ b/cache/predicate.go
@@ -14,6 +14,8 @@
 
 package cache
 
+import "bytes"
+
 type Prefix string
 
 func (prefix Prefix) Match(key []byte) bool {
@@ -22,4 +24,29 @@ func (prefix Prefix) Match(key []byte) bool {
 	}
 	prefixLen := len(prefix)
 	return len(key) >= prefixLen && string(key[:prefixLen]) == string(prefix)
+}
+
+func ExactKey(key []byte) KeyPredicate {
+	return func(k []byte) bool { return bytes.Equal(k, key) }
+}
+
+func FromKey(start []byte) KeyPredicate {
+	return func(k []byte) bool { return bytes.Compare(k, start) >= 0 }
+}
+
+func Range(start, end []byte) KeyPredicate {
+	return func(k []byte) bool {
+		return bytes.Compare(k, start) >= 0 &&
+			bytes.Compare(k, end) < 0
+	}
+}
+
+func KeyPredForRange(start, end []byte) KeyPredicate {
+	if len(end) == 0 {
+		return ExactKey(start)
+	}
+	if len(end) == 1 && end[0] == 0 {
+		return FromKey(start)
+	}
+	return Range(start, end)
 }
--- a/tests/integration/cache_test.go
+++ b/tests/integration/cache_test.go
@@ -34,7 +34,7 @@ func TestCacheWatch(t *testing.T) {
 	t.Cleanup(func() { clus.Terminate(t) })
 	client := clus.Client(0)
 
-	c, err := cache.New(client, "/", cache.WithHistoryWindowSize(32))
+	c, err := cache.New(client, "/foo", cache.WithHistoryWindowSize(32))
 	if err != nil {
 		t.Fatalf("New(...): %v", err)
 	}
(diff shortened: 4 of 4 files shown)


tests/integration/cache_test.go
```diff
@@ -45,6 +45,110 @@ func TestCacheWatch(t *testing.T) {
 	testWatch(t, client.KV, c)
 }
 
+func TestCacheWatchSubRange(t *testing.T) {
```
serathius, line 48: Please merge it into TestWatch

tests/integration/cache_test.go
```diff
@@ -93,44 +197,143 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
 		Type: clientv3.EventTypeDelete,
 		Kv: &mvccpb.KeyValue{
 			Key:         []byte("/b"),
 			ModRevision: 5,
 		},
 	}
+	event6PutFoo := &clientv3.Event{
+		Type: clientv3.EventTypePut,
+		Kv: &mvccpb.KeyValue{
+			Key:            []byte("/foo"),
```
serathius: No need to have separate events for testing subranges, you can edit the existing `event1PutA` to make the test work with subranges.

tests/integration/cache_test.go
```diff
@@ -93,44 +197,143 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
+			wantEvents: nil,
+		},
+		{
+			name:       "Watch range empty",
+			key:        "",
+			opts:       []clientv3.OpOption{clientv3.WithRange("")},
+			wantEvents: nil,
+		},
+		{
+			name: "Watch range open-end",
```
serathius: Isn't this just `fromKey`? No need to manually put the range.

tests/integration/cache_test.go
```diff
@@ -93,44 +197,143 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
+			wantEvents: nil,
+		},
+		{
+			name:       "Watch invalid range [/b, /a)",
+			key:        "/b",
+			opts:       []clientv3.OpOption{clientv3.WithRange("/a")},
+			wantEvents: nil,
+		},
+		{
+			name: "Watch from key /a",
```
serathius: Range from `/a` covers all keys, maybe instead test from `/b`?

tests/integration/cache_test.go
```diff
@@ -93,44 +197,143 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
+			key:        "/foo",
+			opts:       []clientv3.OpOption{clientv3.WithFromKey()},
+			wantEvents: []*clientv3.Event{event6PutFoo, event7PutFooBar, event8PutFooBaz},
+		},
+		{
+			name: "Watch from empty key",
+			key:  "",
+			opts: []clientv3.OpOption{clientv3.WithFromKey()},
+			wantEvents: []*clientv3.Event{event1PutA, event2PutB, event3DeleteA, event4PutA, event5DeleteB,
+				event6PutFoo, event7PutFooBar, event8PutFooBaz},
```
serathius: Can you add cases to test `WithPrefix`?

tests/integration/cache_test.go
```diff
@@ -45,6 +45,110 @@ func TestCacheWatch(t *testing.T) {
...
+		want []*clientv3.Event
+	}{
+		{
+			name: "sub-range fully inside cache [/foo/a, /foo/c)",
+			key:  "/foo/a",
+			opts: []clientv3.OpOption{clientv3.WithRange("/foo/c")},
+			want: []*clientv3.Event{event1PutFooA, event2PutFooB, event3DelFooA},
+		},
+		{
+			name: "sub-range partially outside cache [/foo/b, /z)",
```
serathius: We should reject any range that is outside the cached values.

tests/integration/cache_test.go
```diff
@@ -45,6 +45,110 @@ func TestCacheWatch(t *testing.T) {
...
+			key:  "/foo/b",
+			opts: []clientv3.OpOption{clientv3.WithRange("/z")},
+			want: []*clientv3.Event{event2PutFooB, event4PutFooC},
+		},
```
serathius: Please add many more cases, like test WithPrefix and FromKey.

tests/integration/cache_test.go
serathius: I recommend to merge this test into the `TestWatch` and handle it:
* For cases where Watch range is within cache range, response from cache and direct watch should match
* For cases where Watch range is outside cache range, cache should return error and direct watch can skip the test.

You can handle skipping by:
* Add `cacheEnabled` argument to `testWatch` function so it caller can make decision based on it
* For cases where Watch range is outside cache range, we can add a field `skipIfCacheEnabled`
* When executing the test cases we can skip cases when `cacheEnabled && tc.skipIfCacheEnabled`

tests/integration/cache_test.go
serathius: No, we should expect empty response/cancel.

tests/integration/cache_test.go
```diff
@@ -51,91 +53,228 @@ func TestWatch(t *testing.T) {
...
+			key:        "/foo/b",
+			opts:       []clientv3.OpOption{clientv3.WithRange("/foo/c")},
+			wantEvents: []*clientv3.Event{event2PutFooB, event5DeleteFooB, event7PutFooBar, event8PutFooBaz},
+		},
+		{
+			name:               "Watch range [/foo/c, /z)",
+			key:                "/foo/c",
+			opts:               []clientv3.OpOption{clientv3.WithRange("/z")},
+			wantEvents:         []*clientv3.Event{event6PutFooC, event9PutYoo},
+			skipIfCacheEnabled: true,
```
serathius: We should still test if cache rejects response.

tests/integration/cache_test.go
```diff
@@ -29,20 +29,22 @@ import (
 )
 
 func TestCacheWatch(t *testing.T) {
+	t.Skip("skip until we support cache watch on range")
```
serathius: ???

serathius: We cannot just merge commit that disables tests.

apullo777: Ah, I got it. Just to check I’ve got this right. Since `newWatchEventStream` today returns `ErrUnsupportedWatch` for any sub‐prefix/range, should we update our tests so that any case with Watch on range asserts that the watch is canceled (e.g. resp.Canceled == true) rather than emitting normal events? So perhaps to add a field like `expectCanceled`.

serathius: There is no value of merging code just to test watch. The goal of preparing those tests was to guide your implementation.

tests/integration/cache_test.go
apullo777: Ok. I will modify `skipIfCacheEnabled` to be `expectCanceled` and change the validate logic like below:

```
	t.Log("Validate")
	for i, tc := range tcs {
		i, tc := i, tc
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			resp, ok := <-watches[i]
			if !ok {
				t.Fatalf("unexpected closed watch")
			}
			if resp.Canceled {
				if tc.expectCanceled {
					// invalid range or cache-only out of range key
					return
				}
				t.Fatalf("unexpected canceled watch: %+v", resp)
			}
			// We ignore the first WatchResponse which is the CreatedNotify (len(resp.Events) == 0) and read the real payload that follows.
			events, _ := readEvents(watches[i])
			if diff := cmp.Diff(tc.wantEvents, events); diff != "" {
				t.Errorf("unexpected events (-want +got):\n%s", diff)
			}
		})
	}
```
and then also add `clientv3.WithCreatedNotify()` so that the initial `<-watches[i] ` unblock deterministically:

```
		watches[i] = watcher.Watch(
			ctx,
			tc.key,
			append(tc.opts, clientv3.WithCreatedNotify())...,
		)
```

tests/integration/cache_test.go
apullo777: Sorry. Totally misunderstood the purpose of this PR. I will continue work on this PR to do the implementation using these tests.

serathius:
Please fix conflict.

cache/cache.go
```diff
@@ -103,6 +103,23 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
 	requestedRev := op.Rev()
 
+	startKey := []byte(key)
+	endKey := op.RangeBytes() // nil → single key, {0} → fromKey
+
+	prefix := []byte(c.prefix)
+	startIn := bytes.HasPrefix(startKey, prefix)
+	endIn := len(endKey) == 0 || bytes.HasPrefix(endKey, prefix)
+	openEnded := len(endKey) == 1 && endKey[0] == 0
+	invalid := len(endKey) > 0 && !openEnded && bytes.Compare(startKey, endKey) >= 0
```
serathius: I said to not use `prefix(a)` semantic, but use `range(a,b)`

cache/cache.go
```diff
@@ -105,8 +105,35 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	op := clientv3.OpGet(key, opts...)
 	startRev := op.Rev()
 
-	// TODO: Support watch on subprefix and single key & arbitrary startRev support once we guarantee gap-free replay.
-	if key != c.prefix || !clientv3.IsOptsWithPrefix(opts) || startRev != 0 {
+	// TODO: Support arbitrary startRev support once we guarantee gap-free replay.
+	if  startRev != 0 {
```
serathius: This function is too long, move watch validation to separate function.

cache/cache.go
```diff
@@ -226,3 +221,29 @@ func readWatchChannel(
 	}
 	return nil
 }
+
+func (c *Cache) validateWatch(key string, opts ...clientv3.OpOption) (cancelChan clientv3.WatchChan, pred KeyPredicate) {
```
serathius: `validateWatch` should validate watch and return error if failed, not return `(cancelChan clientv3.WatchChan, pred KeyPredicate)`

cache/predicate.go
```diff
@@ -23,3 +25,16 @@ func (prefix Prefix) Match(key []byte) bool {
 	prefixLen := len(prefix)
 	return len(key) >= prefixLen && string(key[:prefixLen]) == string(prefix)
 }
+
+func rangePred(start, end []byte) func(k []byte) bool {
+	return func(k []byte) bool {
+		if bytes.Compare(k, start) < 0 {
+			return false
+		}
+		// no end, or the sentinel {0}  => accept everything ≥ start
```
serathius: Don't use sentinel, create separate predicate.

cache/cache.go
```diff
@@ -226,3 +221,29 @@ func readWatchChannel(
...
+
+	switch {
+	case len(endKey) == 0:
+		pred = func(k []byte) bool { return bytes.Equal(k, startKey) }
+	case openEnd:
+		pred = func(k []byte) bool { return bytes.Compare(k, startKey) >= 0 }
+	default:
+		pred = rangePred(startKey, endKey)
+	}
+	return nil, pred
```
serathius: Move code to create predicate to dedicated function in `cache/predicate.go`

cache/cache.go
```diff
@@ -102,26 +102,21 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
-		return responseChan
+	cancelChan, pred := c.validateWatch(key, opts...)
+	if cancelChan != nil {
+		return cancelChan
 	}
 
-	w := newWatcher(c.cfg.PerWatcherBufferSize, func(k []byte) bool { return strings.HasPrefix(string(k), key) })
-	c.demux.Register(w, startRev)
+	w := newWatcher(c.cfg.PerWatcherBufferSize, pred)
+	c.demux.Register(w, 0)
```
```suggestion
	startRev := 0
	c.demux.Register(w, startRev)
```

serathius: Setting here a zero obfuscates the argument.

cache/cache.go
MadhavJivrajani: +1
Interpret the error at the call sight and then close the channel like you do currently

tests/integration/cache_test.go
apullo777: I found that TestCacheLaggingWatcher/all_event_fit was failing with:

```
=== RUN   TestCacheLaggingWatcher/all_event_fit
    cache_test.go:367: gotEvents=0, wantEvents=9
    cache_test.go:373: closed=true, wantClosed=false
```
because the unbuffered channel blocked on the {Created:true} send, and never read real events.

I noticed that client/v3/watch.go do  `make(chan WatchResponse, 1)` before sending the Created notification. So I tried to follow this pattern to switched the watch’s response channel to a one-slot buffer:  
`responseChan := make(chan clientv3.WatchResponse, 1)`

We could also drop the “created” notification entirely, but I need to figure out how to do that.

cache/cache.go
```diff
@@ -105,23 +104,24 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
 	c.demux.Register(w, startRev)
 
-	responseChan := make(chan clientv3.WatchResponse)
+	responseChan := make(chan clientv3.WatchResponse, 1)
 	c.waitGroup.Add(1)
 	go func() {
 		defer c.waitGroup.Done()
 		defer close(responseChan)
 		defer c.demux.Unregister(w)
+		responseChan <- clientv3.WatchResponse{Created: true}
```
serathius: Create response should only be send if client requests it.

cache/cache.go
```diff
@@ -175,7 +175,7 @@ func serveWatchEvents(ctx context.Context, watchCtx *watchCtx) {
 	backoff := watchCtx.backoffStart
 	for {
 		opts := []clientv3.OpOption{
-			clientv3.WithPrefix(),
+			clientv3.WithFromKey(),
```
serathius: This should not be changed.

apullo777: Ok. I will add a helper so that testWatch can use it to filter wantEvents by cache's prefix.

serathius: I don't understand.

apullo777: Sorry, I did not explain it well. We have a generic table-driven helper `testWatch` that runs the same scenarios against both a raw client watch and our cache-backed watch, and in the table we includes an event on "/yoo" (event 9) that is outside cache's prefix "/foo" (we called `cache.New(cli, "/foo", …)` in `TestCacheWatch`).  I just found that I did not design a mechanism in `testWatch` to filter for test-cases that originally expected to see the “/yoo” event for `TestWatch` so the cache tests will constantly fail with “missing event” with `WithPrefix()`.

serathius: Ok, please be clear next time to avoid confusion. No need to be so detailed. 

"Ok, will also need to update tests scenarios because they depend on this" is enough.

cache/cache.go
```diff
@@ -226,3 +230,37 @@ func readWatchChannel(
...
+
+	if isRange && bytes.Compare(endKey, startKey) <= 0 {
+		return nil, ErrUnsupportedWatch
+	}
+
+	if c.prefix != "" && isRange {
+		prefixStart := []byte(c.prefix)
+		prefixEndStr := clientv3.GetPrefixRangeEnd(c.prefix)
+		prefixEnd := []byte(prefixEndStr)
+		endBound := endKey
```
```suggestion
```

cache/cache.go
```diff
@@ -226,3 +230,37 @@ func readWatchChannel(
...
+
+		if bytes.Compare(startKey, prefixStart) < 0 ||
+			bytes.Compare(startKey, prefixEnd) >= 0 ||
+			bytes.Compare(endBound, prefixStart) <= 0 ||
+			bytes.Compare(endBound, prefixEnd) > 0 {
+			return nil, ErrUnsupportedWatch
+		}
+	}
```
serathius: Validating ranges should be a separate function and have dedicated unit tests.

tests/integration/cache_test.go
```diff
@@ -51,93 +52,258 @@ func TestWatch(t *testing.T) {
...
+			key:        "/foo/b",
+			opts:       []clientv3.OpOption{clientv3.WithRange("/foo/c")},
+			wantEvents: []*clientv3.Event{event2PutFooB, event5DeleteFooB, event7PutFooBar, event8PutFooBaz},
+		},
+		{
+			name:               "Watch valid range [/foo/c, /z)",
+			key:                "/foo/c",
+			opts:               []clientv3.OpOption{clientv3.WithRange("/z")},
+			wantEvents:         []*clientv3.Event{event6PutFooC, event9PutYoo},
+			skipIfCacheEnabled: true,
```
serathius: Instead of testing direct watch and skipping testing cache, you should do reverse. Skip testing watch, and test that cache returns error for cache outside of cache prefix.

tests/integration/cache_test.go
```diff
@@ -51,93 +52,258 @@ func TestWatch(t *testing.T) {
...
+				return
+			}
 			events, _ := readEvents(watches[i])
-			if diff := cmp.Diff(tc.wantEvents, events); diff != "" {
+			filteredWant := tc.wantEvents
+			if cp, ok := watcher.(interface{ Prefix() string }); ok {
+				if p := cp.Prefix(); p != "" {
+					filteredWant = filterByPrefix(filteredWant, p)
+				}
+			}
```
serathius: There should be no events filtered. 

Don't know what it does, but it's definitely wrong. Please remove.

tests/integration/cache_test.go
```diff
@@ -144,6 +252,73 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 	}
 }
 
+func TestCacheRejectsInvalidWatchOptions(t *testing.T) {
```
serathius: This test should be merged with `TestCacheRejectsUnsupportedWatch`. We should have single test to track what options are supported/work.

cache/cache.go
```diff
@@ -226,3 +230,59 @@ func readWatchChannel(
...
+			bytes.Compare(endKey, prefixStart) <= 0 ||
+			bytes.Compare(endKey, prefixEnd) > 0 {
+			return ErrUnsupportedWatch
+		}
+	}
+
+	// single-key watch must lie within [prefixStart, prefixEnd)
+	if bytes.Compare(startKey, prefixStart) < 0 || bytes.Compare(startKey, prefixEnd) >= 0 {
+		return ErrUnsupportedWatch
+	}
```
serathius: This should be switch because cases are mutually exclusive

cache/cache.go
```diff
@@ -165,6 +165,10 @@ func (c *Cache) WaitReady(ctx context.Context) error {
 	}
 }
 
+func (c *Cache) Prefix() string {
```
serathius: This is no longer used.

cache/cache.go
```diff
@@ -226,3 +230,59 @@ func readWatchChannel(
...
+
+	if err := c.validateWatchRange(startKey, endKey); err != nil {
+		return nil, err
+	}
+	return KeyPredForRange(startKey, endKey), nil
+}
+
+func (c *Cache) validateWatchRange(startKey, endKey []byte) error {
+	isFrom := len(endKey) == 1 && endKey[0] == 0
+	isRange := endKey != nil && !isFrom
```
serathius: What about case where `endKey` is empty?

cache/cache.go
```diff
@@ -226,3 +230,59 @@ func readWatchChannel(
...
+	prefixStart := []byte(c.prefix)
+	prefixEnd := []byte(clientv3.GetPrefixRangeEnd(c.prefix))
+
+	if isRange && bytes.Compare(endKey, startKey) <= 0 {
```
serathius: This code for testing range should be next to line 273 where there is other statement to test range.

cache/cache.go
```diff
@@ -226,3 +230,59 @@ func readWatchChannel(
...
+		return ErrUnsupportedWatch
+	}
+
+	if c.prefix == "" {
```
serathius: Don't like that you are treating `c.prefix == ""` as special case. Will rest of the validation fail?

cache/cache_test.go
```diff
@@ -0,0 +1,199 @@
...
+			end:     []byte{0},
+			wantErr: false,
+		},
+		{
+			name:    "prefix m: fromKey above prefix end error",
+			prefix:  "m",
+			start:   []byte("z"),
+			end:     []byte{0},
+			wantErr: true,
+		},
```
serathius: hardcoding cases based on byte representation of client options like `fromKey` makes the test less readable. The test case should be:

```
		{
			name:    "prefix m: fromKey above prefix end error",
			cachePrefix:  "m",
			watchKey: "z",
			option: clientv3.WithFromKey(),
			wantErr: true,
		},
```

And the test then would be:
```
			op := clientv3.OpGet(c.key, c.opt)
			err := dummyCache.validateWatchRange([]byte(c.key), op.RangeBytes())
			if gotErr := err != nil; gotErr != c.wantErr {
				t.Fatalf("validateWatchRange(%q, %v) = %v, wantErr=%v",
					c.prefix, c.start, err, c.wantErr)
			}
```

cache/cache_test.go
```diff
@@ -0,0 +1,199 @@
...
+		{
+			name:    "root: reverse range error",
+			prefix:  "",
+			start:   []byte("b"),
+			end:     []byte("a"),
+			wantErr: true,
+		},
+		{
+			name:    "prefix m: range below start error",
```
serathius: `prefix m:` doesn't inform anything about the test. Also client we would not normally use `m`, tests should cover plausible arguments.


See example:
```
		{
			name:    "range encompassing cache prefix returns error",
			key:   []byte("/a"),
			opt:   clientv3.WithRange("/z")
			cachePrefix:  "/foo",
			wantErr: true,
		},
```

cache/cache.go
apullo777: I found it difficult to handle the empty string prefix case, which will get us a non-nil, zero-length []byte{""} (and then `GetPrefixRangeEnd("")` give us a `prefixEnd` with []byte{0}). An option I have in mind is to translate its `prefixStart` & `prefixEnd` to represent open-ended. Or we can hide this special case to be a check before we do prefix validation like below:

```
	prefixStart := []byte(c.prefix)
	prefixEnd := []byte(clientv3.GetPrefixRangeEnd(c.prefix))

	switch {
	case len(endKey) == 0: // single key
		if c.prefix != "" &&
			(bytes.Compare(startKey, prefixStart) < 0 ||
				bytes.Compare(startKey, prefixEnd) >= 0) {
			return ErrUnsupportedWatch
		}

	case len(endKey) == 1 && endKey[0] == 0: //fromKey
		if c.prefix != "" &&
			bytes.Compare(startKey, prefixEnd) >= 0 {
			return ErrUnsupportedWatch
		}

	default: // range
		if bytes.Compare(endKey, startKey) <= 0 {
			return ErrUnsupportedWatch
		}
		if c.prefix != "" &&
			(bytes.Compare(startKey, prefixStart) < 0 ||
				bytes.Compare(startKey, prefixEnd) >= 0 ||
				bytes.Compare(endKey, prefixStart) <= 0 ||
				bytes.Compare(endKey, prefixEnd) > 0) {
			return ErrUnsupportedWatch
		}
	}
```

serathius: That's better

cache/cache_test.go
```diff
@@ -0,0 +1,203 @@
...
+		},
+		{
+			name:        "range crossing prefixEnd returns error",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/z")},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "range crossing prefixEnd returns error",
```
serathius: This test is duplicated

cache/cache_test.go
```diff
@@ -0,0 +1,203 @@
...
+		},
+		{
+			name:        "empty prefix",
+			watchKey:    "",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "empty prefix with cachePrefix",
```
serathius: This should return error, cache will return only objects with "/foo" prefix, while the watcher wants all objects.

cache/cache_test.go
```diff
@@ -0,0 +1,203 @@
...
+		},
+		{
+			name:        "fromKey open‑ended",
+			watchKey:    "/a",
+			opts:        []clientv3.OpOption{clientv3.WithFromKey()},
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "fromKey starting at prefix start",
```
serathius, line 157: This should return error. Watch from "/foo" allows for keys like "/z" that are not within prefix.

cache/cache_test.go
```diff
@@ -0,0 +1,203 @@
...
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithFromKey()},
+			cachePrefix: "/foo",
+			wantErr:     false,
+		},
+		{
+			name:        "fromKey starting below prefixEnd",
```
serathius, line 164: This should return error.

cache/cache_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	for _, c := range tests {
+		t.Run(c.name, func(t *testing.T) {
+			dummyCache := &Cache{prefix: c.cachePrefix}
+			var op clientv3.Op
+			if len(c.opts) == 0 {
+				op = clientv3.OpGet(c.watchKey)
+			} else {
+				op = clientv3.OpGet(c.watchKey, c.opts...)
+			}
```
serathius:
```suggestion
			op := clientv3.OpGet(c.watchKey, c.opts...)
```

cache/cache_test.go
apullo777: Thanks! Just want to check if I got this right:

- A root cache (prefix=="") will accept any valid etcd watch.
- A shard cache (e.g. prefix="/foo") can only accept watches whose entire [startKey,endKey) lies within "/foo"…"/fop",. And with `WithFromKey()` we should reject a watch starting anywhere below "/foo" even though that will also return events within "/foo"…"/fop".

serathius: Yes, there should not be any events missing in watch, as so watch can only be opened on cache that encompasses its range. For WithFromKey() it means it only works for cache with prefix "".

apullo777: Ok. Then I should also modify integration tests to make sure cache rejecting WithFromKey & empty WithPrefix.

serathius: Sounds good

cache/cache.go
```diff
@@ -226,3 +226,54 @@ func readWatchChannel(
...
+	endKey := op.RangeBytes() // nil = single key, {0}=FromKey, else explicit range
+
+	if err := c.validateWatchRange(startKey, endKey); err != nil {
+		return nil, err
+	}
+	return KeyPredForRange(startKey, endKey), nil
+}
+
+func (c *Cache) validateWatchRange(startKey, endKey []byte) error {
+	shard := c.prefix != ""
```
serathius: What is shard? There is no such concept in etcd. Not it was ever mentioned in our discussions. Please revert to previous code.

cache/cache.go
```diff
@@ -226,3 +226,54 @@ func readWatchChannel(
...
+	isSingleKey := len(endKey) == 0
+	isFromKey := len(endKey) == 1 && endKey[0] == 0
+
+	switch {
+	case isSingleKey:
+		if shard &&
+			(bytes.Compare(startKey, prefixStart) < 0 ||
+				bytes.Compare(startKey, prefixEnd) >= 0) {
+			return ErrUnsupportedWatch
+		}
```
```suggestion
		if c.prefix == "" {
			return nil
		}
		if bytes.Compare(startKey, prefixStart) < 0 || bytes.Compare(startKey, prefixEnd) >= 0 {
			return ErrUnsupportedWatch
		}
		return nil
```

cache/cache.go
```diff
@@ -226,3 +226,54 @@ func readWatchChannel(
...
+
+	case isFromKey:
+		if shard {
+			return ErrUnsupportedWatch
+		}
```
```suggestion
		if c.prefix != "" {
			return ErrUnsupportedWatch
		}
		return nil
```

cache/cache.go
```diff
@@ -226,3 +226,54 @@ func readWatchChannel(
...
+	default:
+		if bytes.Compare(endKey, startKey) <= 0 {
+			return ErrUnsupportedWatch
+		}
+		if shard &&
+			(bytes.Compare(startKey, prefixStart) < 0 ||
+				bytes.Compare(endKey, prefixEnd) > 0) {
+			return ErrUnsupportedWatch
+		}
+	}
```
```suggestion
		if bytes.Compare(endKey, startKey) <= 0 {
			return ErrUnsupportedWatch
		}
		if c.prefix == "" {
			return nil
		}
		if bytes.Compare(startKey, prefixStart) < 0 || bytes.Compare(endKey, prefixEnd) > 0 {
			return ErrUnsupportedWatch
		}
		return nil
	}
```

cache/cache_test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	tests := []tc{
+		{
+			name:        "single key",
+			watchKey:    "/a",
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
```
serathius, line 39:
```suggestion
		{
			name:        "prefix single key",
			watchKey:    "/foo/a",
			cachePrefix: "/foo",
			wantErr:     false,
		},
		{
```

cache/cache_test.go
```diff
@@ -0,0 +1,191 @@
...
+			name:        "empty range returns error",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/foo")},
+			cachePrefix: "",
+			wantErr:     true,
+		},
+		{
+			name:        "range strating below cache prefix returns error",
```
serathius:
```suggestion
			name:        "range starting below cache prefix returns error",
```

cache/cache_test.go
```diff
@@ -0,0 +1,191 @@
...
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "range ending above cache prefix returns error",
+			watchKey:    "/foo",
+			opts:        []clientv3.OpOption{clientv3.WithRange("/z")},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
```
serathius:
```suggestion
```

This test duplicates `range crossing prefixEnd returns error`

cache/cache_test.go
```diff
@@ -0,0 +1,191 @@
...
+		{
+			name:        "empty prefix",
+			watchKey:    "",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "",
+			wantErr:     false,
+		},
+		{
+			name:        "empty prefix with cachePrefix",
```
serathius:
```suggestion
			name:        "empty prefix with cachePrefix returns error",
```

cache/cache_test.go
```diff
@@ -0,0 +1,191 @@
...
+			watchKey:    "",
+			opts:        []clientv3.OpOption{clientv3.WithPrefix()},
+			cachePrefix: "/foo",
+			wantErr:     true,
+		},
+		{
+			name:        "prefix watch matches cachePrefi xexactly",
```
serathius:
```suggestion
			name:        "prefix watch matches cachePrefix exactly",
```

tests/integration/cache_test.go
```diff
@@ -34,7 +34,7 @@ func TestCacheWatch(t *testing.T) {
 	t.Cleanup(func() { clus.Terminate(t) })
 	client := clus.Client(0)
 
-	c, err := cache.New(client, "/", cache.WithHistoryWindowSize(32))
+	c, err := cache.New(client, "", cache.WithHistoryWindowSize(32))
```
serathius: We wanted to test cache opened on prefix "/foo". Why change that?

tests/integration/cache_test.go
```diff
@@ -56,84 +56,192 @@ func TestWatch(t *testing.T) {
...
+		Type: clientv3.EventTypePut,
+		Kv: &mvccpb.KeyValue{
+			Key:            []byte("/foo/baz"),
+			Value:          []byte("z"),
+			CreateRevision: 8,
+			ModRevision:    8,
+			Version:        1,
+		},
+	}
+	event9PutFooYoo := &clientv3.Event{
```
serathius, line 133: Can you add one write with key without `/foo` prefix?

tests/integration/cache_test.go
apullo777: I removed it because now our cache will reject events outside cache prefix, and testEatch will compare their results end to end. Perhaps I should find a way to handle these different expectations.

serathius: I think we need separate tests when cache prefix is "" or "/foo". However this is not trivial to implement. Let's skip it for now and just test "/foo".

serathius (approved):

Issue #20224, opened 2025-06-25
Update docs of --quota-backend-bytes


derselbst:
### What would you like to be added?

The documentation of `--quota-backend-bytes` currently read as this:

> Raise alarms when backend size exceeds the given quota (0 defaults to low space quota).

For me as a newbie, this means an alarm will be raised but etcd will continue to work fine. After running into this issue, I found that this is obviously not the case. If the quota is exceeded an alarm is raised and etcd is set to read-only mode, causing every Kubernetes cluster to fail hazardously.

The docs should clearly point out, that this sets the maximum storage that the database may ever consume. When exceeded, etcd is put to RO mode. The hint about "0 defaults to low space quota" is not helpful either. I need to dig into the source code to find that it's only 2Gi. This should be explicitly mentioned as well.

[This blog](https://etcd.io/blog/2023/how_to_debug_large_db_size_issue/) entry has it correct. But a user normally doesn't search through blogs when trying to understand the commandline options.

### Why is this needed?

Because the current doc is misleading and wrong.

ahrtr:
@derselbst pls feel free to raise a PR to update it. thx

zechariahkasina:
@ahrtr Please review the PR https://github.com/etcd-io/etcd/pull/20328

ahrtr:
Reopen to track the backport effort, otherwise it's easy to be forgotten.

We need someone to us manually backport the change to 3.5 and 3.4.  It's good first issue.

3.6 is already present https://github.com/etcd-io/etcd/pull/20352

ahrtr:
Done

PR #20328, opened 2025-07-10
Updated help for --quota-backend-bytes

zechariahks:
Fixes the issue# [20224](https://github.com/etcd-io/etcd/issues/20224)


ahrtr:
Please signoff the commit, read https://github.com/etcd-io/etcd/pull/20328/checks?check_run_id=45763400272

thx

server/etcdmain/help.go
```diff
@@ -78,7 +78,7 @@ Member:
   --memory-mlock
     Enable to enforce etcd pages (in particular bbolt) to stay in RAM.
   --quota-backend-bytes '0'
-    Raise alarms when backend size exceeds the given quota (0 defaults to low space quota).
+    Sets the maximum size (in bytes) that the etcd backend database may consume. If the quota is exceeded, etcd raises an alarm and transitions into read-only mode, preventing further writes. A value of 0 applies the default low space quota of 2 GiB.
```
serathius: > Sets the maximum size (in bytes) that the etcd backend database may consume

I would be careful about saying that it sets a maximum. This is not a accurate hard max, it's based on heuristic executed that etcd tries to guess that write will go over the maximum. The real database size doesn't grow linearly, but is executed as we run out of free pages. I expect that bbolt does some overallocations for future, but I haven't read code recently to say how much.

cc @ahrtr

derselbst: > This is not a accurate hard max, it's based on heuristic executed that etcd tries to guess that write will go over the maximum.

IMO, this is an implementation detail, which is not of relevance here. As the one who opened the original issue, I find this explanation sufficiently accurate. Thanks @zechariahks for taking care of this!

To the maintainers: Pls. ensure this documentation update will be properly reflected in the v3.5 and v3.6 docs.

derselbst (approved):

server/etcdmain/help.go
ahrtr: > I would be careful about saying that it sets a maximum. This is not a accurate hard max

I think it's accurate. The only thing that we might need to clarify is that the last write right before exceeding the quota is still persisted, but users may get an `ErrNoSpace`. So the users may get confused. I recall that we discussed this before, but I couldn't find it now. We can raise a separate issue to track it (update the doc).



> The real database size doesn't grow linearly, but is executed as we run out of free pages. I expect that bbolt does some overallocations for future

It's a different topic, it's exactly the reason why we need defragmentation to reclaim the unused space. But from etcd & users perspective, they don't need to care about the details. They only need to care about `the whole db size + current size of request < quota`.

zechariahkasina: Thank you @derselbst @ahrtr for the detailed explanation. I just added the sign off to the commit.

ahrtr: Please also update https://github.com/etcd-io/etcd/blob/cfd17b506088a1d1365e5d3100c657450bf65686/server/embed/config.go#L626
```suggestion
    Sets the maximum size (in bytes) that the etcd backend database may consume. Exceeding this triggers an alarm and puts etcd in read-only mode. Set to 0 to use the default 2GiB limit.
```

zechariahkasina: @ahrtr I have updated both the files. Can you please look into this now.

ahrtr:
pls squash the commits, thx

zechariahkasina:
> pls squash the commits, thx

@ahrtr Squashed the commits. Please check. Thanks again.

ahrtr (approved):
LGTM & thx

zechariahkasina:
@ahrtr Will this be merged automatically? or do I need to tag someone here? Thank you.

ahrtr:

ahrtr:
We also need to backport this PR to 3.5 and 3.4. Contribution is welcome! thx

4rivappa:
Hi @ahrtr 

I have raised backport PRs for `release-3.4` and `release-3.5`  🎉

ahrtr:
> Hi @ahrtr
> 
> I have raised backport PRs for `release-3.4` and `release-3.5` 🎉

Thanks!

@4rivappa @ivanvc we may also want to update the changelog items for 3.4/3.5/3.6.

4rivappa:
Hi @ahrtr @ivanvc 

I have raised a PR to update changelog - #20381 
Please review it.

PR #20344, opened 2025-07-16
Remove verify.Assert

serathius:
The verify package is meant to build test only verification during runtime. Having a trivial function that always panics causes confusion.

There is no reason to have a library function just to replace a simple if with panic statement.

/cc @fuweid @ahrtr

### DIFF
--- a/pkg/ioutil/pagewriter.go
+++ b/pkg/ioutil/pagewriter.go
@@ -16,8 +16,6 @@ package ioutil
 
 import (
 	"io"
-
-	"go.etcd.io/etcd/client/pkg/v3/verify"
 )
 
 var defaultBufferBytes = 128 * 1024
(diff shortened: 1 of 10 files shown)


ahrtr:
Pls read https://github.com/etcd-io/etcd/pull/13947

The original motivation to add `verify.Assert` is explicit, I think it still stands.

serathius:
Explicit was used when comparing to `lg.Debug`. 

The proposed assertion in https://github.com/etcd-io/etcd/pull/13947#issuecomment-1108246247 was based on private function, not a public library function.

pkg/ioutil/pagewriter.go
```diff
@@ -43,7 +41,9 @@ type PageWriter struct {
 // NewPageWriter creates a new PageWriter. pageBytes is the number of bytes
 // to write per page. pageOffset is the starting offset of io.Writer.
 func NewPageWriter(w io.Writer, pageBytes, pageOffset int) *PageWriter {
-	verify.Assert(pageBytes > 0, "invalid pageBytes (%d) value, it must be greater than 0", pageBytes)
+	if pageBytes <= 0 {
+		panic("invalid pageBytes value, it must be greater than 0")
```
fuweid, line 45:
```suggestion
		panic(fmt.Sprinf("invalid pageBytes (%d) value, it must be greater than 0", pageBytes))
```

fuweid (commented):
I don't have a strong opinion on this. Both assert and panic work for me.

Assert might be slightly better than panic because it wraps `fmt.Sprintf`, which provides more convenient string formatting. That's about the only advantage I see.

PR #20345, opened 2025-07-16
cache: refactor tests into predix vs. no-prefix cases

apullo777:

@serathius


tests/integration/cache_test.go
```diff
@@ -250,71 +274,61 @@ func TestCacheRejectsInvalidWatch(t *testing.T) {
...
+		name           string
+		key            string
+		opts           []clientv3.OpOption
+		expectCanceled bool
 	}{
 		{
-			name:        "non‑zero start revision",
-			cachePrefix: "",
-			key:         "",
-			opts:        []clientv3.OpOption{clientv3.WithPrefix(), clientv3.WithRev(123)},
```
serathius, line 262: You can keep the test case with revision.

PR #20347, opened 2025-07-16
cache: enable Watch on arbitrary start_revision

apullo777:

@serathius @MadhavJivrajani

### DIFF
--- a/tests/integration/cache_test.go
+++ b/tests/integration/cache_test.go
@@ -16,6 +16,7 @@ package integration
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
@@ -23,6 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	cache "go.etcd.io/etcd/cache/v3"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/tests/v3/framework/integration"
@@ -56,7 +58,7 @@ func TestWatch(t *testing.T) {
 
 func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 	ctx := t.Context()
-	event1PutFooA := &clientv3.Event{
+	rev2PutFooA := &clientv3.Event{
 		Type: clientv3.EventTypePut,
 		Kv: &mvccpb.KeyValue{
 			Key:            []byte("/foo/a"),
@@ -66,7 +68,7 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 			Version:        1,
 		},
 	}
-	event2PutFooB := &clientv3.Event{
+	rev3PutFooB := &clientv3.Event{
 		Type: clientv3.EventTypePut,
 		Kv: &mvccpb.KeyValue{
 			Key:            []byte("/foo/b"),
@@ -76,14 +78,14 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 			Version:        1,
 		},
 	}
-	event3DeleteFooA := &clientv3.Event{
+	rev4DeleteFooA := &clientv3.Event{
 		Type: clientv3.EventTypeDelete,
 		Kv: &mvccpb.KeyValue{
 			Key:         []byte("/foo/a"),
 			ModRevision: 4,
 		},
 	}
-	event4PutFooA := &clientv3.Event{
+	rev5PutFooA := &clientv3.Event{
 		Type: clientv3.EventTypePut,
 		Kv: &mvccpb.KeyValue{
 			Key:            []byte("/foo/a"),
@@ -93,14 +95,14 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 			Version:        1,
 		},
 	}
-	event5DeleteFooB := &clientv3.Event{
+	rev5DeleteFooB := &clientv3.Event{
 		Type: clientv3.EventTypeDelete,
 		Kv: &mvccpb.KeyValue{
 			Key:         []byte("/foo/b"),
 			ModRevision: 5,
 		},
 	}
-	event6PutFooC := &clientv3.Event{
+	rev6PutFooC := &clientv3.Event{
 		Type: clientv3.EventTypePut,
 		Kv: &mvccpb.KeyValue{
 			Key:            []byte("/foo/c"),
@@ -110,7 +112,7 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 			Version:        1,
 		},
 	}
-	event7PutFooBar := &clientv3.Event{
+	rev7PutFooBar := &clientv3.Event{
 		Type: clientv3.EventTypePut,
 		Kv: &mvccpb.KeyValue{
 			Key:            []byte("/foo/bar"),
@@ -120,7 +122,7 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 			Version:        1,
 		},
 	}
-	event8PutFooBaz := &clientv3.Event{
+	rev8PutFooBaz := &clientv3.Event{
 		Type: clientv3.EventTypePut,
 		Kv: &mvccpb.KeyValue{
 			Key:            []byte("/foo/baz"),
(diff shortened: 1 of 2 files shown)


tests/integration/cache_test.go
```diff
@@ -197,61 +237,144 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
+			opts:       []clientv3.OpOption{clientv3.WithRev(12), clientv3.WithPrefix()},
+			wantEvents: []*clientv3.Event{rev12PutFooTx1, rev12DeleteFooFuture, rev12PutFooTx2},
 		},
 	}
 
-	t.Log("Open test watchers")
+	t.Log("Open watchers (prewrite = false)")
 	watches := make([]clientv3.WatchChan, len(tcs))
 	for i, tc := range tcs {
+		if tc.prewrite {
```
serathius: What is prewrite? You never mentioned it in discussions. Field nor test doesn't explain why it's beneficial to split watchers. I don't see a how this improves the test. Please provide motivation and explanation or remove.

apullo777: Sorry for not communicating about `prewrite`. I was thinking that perhaps we should also test scenarios where watchers are being opened after the cache has already buffered initial batch of events. So I wanted to use a flag to control when we open the watch relative to the initial batch of writes (revisions 2–10) vs. the later batch (revisions 11–12). I will remove the logic if it does not work or is unnecessary.

serathius: I would just move that all watch requests are opened after the initial writes and update the test cases.
Now the base test case is not one that opens with `rev=0`, but one that opens with `rev=2` to guarantee it gets all events.

apullo777: Got it! I will remove the `prewrite` flag and restructure the test in this order (1) write the first batch of events, (2) open all watches with `WithRev(2)`, and then (3) write the second batch of events.

serathius: (2) Please also open watches without `WithRev`.

apullo777: Ah, I finally got it. Then I will open watches without `WithRev()` and add that explicit `WithRev(2)` to test cases that need to see revisions 2–10. Thanks!

tests/integration/cache_test.go
```diff
@@ -161,7 +200,7 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 			name:       "Watch single key existing /foo/c",
 			key:        "/foo/c",
 			opts:       nil,
```
serathius:
```suggestion
			opts:       []clientv3.OpOption{clientv3.WithRev(2)},,
```

tests/integration/cache_test.go
```diff
@@ -197,61 +236,130 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
 	}
-	if _, err := kv.Put(ctx, string(event10PutZoo.Kv.Key), string(event10PutZoo.Kv.Value)); err != nil {
+	if _, err := kv.Put(ctx, string(rev10PutZoo.Kv.Key), string(rev10PutZoo.Kv.Value)); err != nil {
 		t.Fatalf("Put: %v", err)
 	}
+
+	t.Log("Open watches (WithRev(2))")
+	watches := make([]clientv3.WatchChan, len(tcs))
+	for i, tc := range tcs {
+		opts := append([]clientv3.OpOption{clientv3.WithRev(2)}, tc.opts...)
```
serathius, line 347: Don't change arguments!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

serathius:
```suggestion
```
Don't change arguments!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

tests/integration/cache_test.go
```diff
@@ -160,98 +199,160 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
+			key:  "/foo/a",
+			opts: []clientv3.OpOption{clientv3.WithRange("/foo/c"), clientv3.WithRev(4)},
+			wantEvents: []*clientv3.Event{
+				rev4DeleteFooA, rev5PutFooA, rev5DeleteFooB, rev7PutFooBar, rev8PutFooBaz,
+			},
+		},
+		{
+			name:       "Watch from rev 11 with single key /foo/future",
+			key:        "/foo/future",
+			opts:       []clientv3.OpOption{clientv3.WithRev(11)},
```
serathius:
```suggestion
			name:       "Watch with single key /foo/future",
			key:        "/foo/future",
			opts:       nil,
```
For watcher on revisions written after we open a watch please drop WithRev

tests/integration/cache_test.go
```diff
@@ -160,98 +199,160 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
+			wantEvents: []*clientv3.Event{rev11PutFooFuture, rev12DeleteFooFuture},
+		},
+		{
+			name:       "Watch from rev 12 with txn prefix /foo",
+			key:        "/foo",
+			opts:       []clientv3.OpOption{clientv3.WithRev(12), clientv3.WithPrefix()},
```
serathius, line 310:
```suggestion
			opts:       []clientv3.OpOption{clientv3.WithPrefix()},
```
For watcher on revisions written after we open a watch please drop WithRev

apullo777: This test was intended to wait until the future moment when rev == 12 before it starts receiving events, so it shouldn’t receive rev == 11. Maybe I should drop the `WithRev(11)` for the other test ("Watch from rev 11 with single key /foo/future") and rename it, and keep `WithRev(12)` for this one?

serathius: Duplicate scenarios, have one that watches for future revision (`WithRev(11)`, `WithRev(12)`) and one that watches for latest revision where the next revision is `rev==11`.

tests/integration/cache_test.go
```diff
@@ -160,98 +199,160 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
...
 	}
-	if _, err := kv.Put(ctx, string(event10PutZoo.Kv.Key), string(event10PutZoo.Kv.Value)); err != nil {
+	if _, err := kv.Put(ctx, string(rev10PutZoo.Kv.Key), string(rev10PutZoo.Kv.Value)); err != nil {
 		t.Fatalf("Put: %v", err)
 	}
+
+	t.Log("Open watches")
+	watches := make([]clientv3.WatchChan, len(tcs))
+	for i, tc := range tcs {
+		watches[i] = watcher.Watch(ctx, tc.key, tc.opts...)
```
serathius, line 347: FYI, you might need to add a sleep here to ensure that next event for watches opened on `rev:0` will be `rev:11`.

tests/integration/cache_test.go
apullo777: Ok. Just double checking if I got this right. I will modify to have below two distinct tests:

(1) Latest‑revision watcher: without WithRev & first incoming event should be revision 11.
(2) Future‑revision watcher: watching at revision 12  (`WithRev(12)`) which hasn’t happened yet. That one should stay idle until revision 12 shows up.

serathius: Sure, we can also have watch on `WithRev(11)`.

tests/integration/cache_test.go
serathius: Maybe put the sleep after the for loop.

apullo777:
Updated the PR:

-  What's changed: in `Watch()` I added a check on `startRev != 0 && startRev < demux.PeekOldest()` to reject watch that requests out-of-window revisions, and remove the old guard & test for rejecting non-zero revision.
-  TODO: handling non‑zero `startRev` on an empty cache to guarantee gap‑free replay. After that, KeyPredicate in Filter?

serathius:
> handling non‑zero startRev on an empty cache to guarantee gap‑free replay. After that, KeyPredicate in Filter?

Don't understand what you mean.

apullo777:
> > handling non‑zero startRev on an empty cache to guarantee gap‑free replay. After that, KeyPredicate in Filter?
> 
> 
> 
> Don't understand what you mean.

It was from our previous discussions regarding what if our cache just start and history is empty when a watcher requests a non-zero starRev.

PR #20354, opened 2025-07-17
Draft implementation of testing cache in robustness tests

serathius:
For now just to show how the integration could look like, a more permanent solution would have a separate test scenario.

I think the tests deadlocks because cache doesn't support progress notifications.

/cc @apullo777 @MadhavJivrajani


serathius:
Extracted https://github.com/etcd-io/etcd/pull/20355

PR #20359, opened 2025-07-18
Inject antithesis at verify.Verify

nwnt:
As a follow up of #20312.  This replaces `verify.Verify` with Antithesis's `assert.Always`.

Note, I had to use `git apply` instead of `patch` because patch does not exist on the image.

cc @serathius @ahrtr

### DIFF
--- a/tests/antithesis/server/Dockerfile
+++ b/tests/antithesis/server/Dockerfile
@@ -6,6 +6,16 @@ FROM golang:$GO_VERSION
 # cloning etcd
 ARG REF=main
 RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etcd
+
+# replace verify with antithesis
+WORKDIR /etcd/client/pkg/verify
+COPY inject/verify.patch verify.patch
+RUN if [ "${REF}" == "main" ]; then \
+      git apply verify.patch; \
+    fi
+RUN rm verify.patch
+WORKDIR /etcd/client/pkg
+RUN go mod tidy
 WORKDIR /etcd
 
 # setup go mod
--- a/tests/antithesis/server/inject/verify.patch
+++ b/tests/antithesis/server/inject/verify.patch
@@ -1,0 +1,22 @@
+diff --git a/client/pkg/verify/verify.go b/client/pkg/verify/verify.go
+index cb48d8ff0..095f890ec 100644
+--- a/client/pkg/verify/verify.go
++++ b/client/pkg/verify/verify.go
+@@ -18,6 +18,8 @@ import (
+ 	"fmt"
+ 	"os"
+ 	"strings"
++
++	"github.com/antithesishq/antithesis-sdk-go/assert"
+ )
+ 
+ const envVerify = "ETCD_VERIFY"
+@@ -69,7 +71,7 @@ func DisableVerifications() func() {
+ func Verify(msg string, f VerifyFunc) {
+ 	if IsVerificationEnabled(envVerifyValueAssert) {
+ 		ok, details := f()
+-		verifier(ok, msg, details)
++		assert.Always(ok, msg, details)
+ 	}
+ }
+ 


ahrtr:
I am OK with the change. Leave it to @serathius to approve & merge this PR.

thx

tests/antithesis/server/Dockerfile
```diff
@@ -6,6 +6,16 @@ FROM golang:$GO_VERSION
 # cloning etcd
 ARG REF=main
 RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etcd
+
+# replace verify with antithesis
+WORKDIR /etcd/client/pkg/verify
+COPY inject/verify.patch verify.patch
+RUN if [ -f verify.go ]; then \
+      git apply verify.patch; \
```
serathius, line 14: Do we need to run this through git? you should be able to just run `patch` command or not?

serathius: See https://github.com/etcd-io/etcd/blob/0888376faba99e93a5789dd67339342a913c8eee/tests/robustness/Makefile#L148-L149

nwnt: `patch` is not available on the go image.

serathius: ok

nwnt: image

serathius: LGTM, have you tested this on all etcd stable branches?

nwnt: I tested in `main` and `release-3.4` and `release-3.5`.  Things break for `release-3.6` because client `verify.Verify` has a different func signature due to the change we recently merged.  The question is do we want to skip the patch for `release-3.6` or do we want to cherry pick the change we made to `verify.Verify` to the branch.  I'd recommend the latter because it'd enable antithesis tests for it.

nwnt: Or maybe both.  I can make it always continue so that it doesn't break when it fails to apply.  what do you think @serathius.

serathius: I think script should fail if apply fail so we detect breaking changes, but let's disable it for other branches.

nwnt: sorry went ill during the weekend.  just pushed a new commit to enable this injection only for main.  Tested against `release-3.4`, `release-3.5` and ` release-3.6` and they were built successfully @serathius.

tests/antithesis/server/Dockerfile
```diff
@@ -6,6 +6,16 @@ FROM golang:$GO_VERSION
 # cloning etcd
 ARG REF=main
 RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etcd
+
+# replace verify with antithesis
+WORKDIR /etcd/client/pkg/verify
+COPY inject/verify.patch verify.patch
+RUN if [ $(git branch --show-current) == "main" && -f verify.go ]; then \
```
serathius:
```suggestion
RUN if [ "${REF}" == "main" ]; then \
```

nwnt: Done.  Note that in the future we will need to bring back `-f verify.go` once we support other branches since the file doesn't exist on etcd 3.4 and 3.5.

serathius: We only should add other branches when they start supporting verify and verify.go will be present.

serathius:
This doesn't work 
```
#9 [ 5/25] RUN if [ "main" == "main" ]; then       git apply verify.patch;     fi
#9 0.117 /bin/sh: 1: [: main: unexpected operator
#9 DONE 0.1s
```

Link, https://github.com/etcd-io/etcd/actions/runs/16478596663/job/46586694433

Issue #20342, opened 2025-07-15
--watch-progress-notify-interval doesn't work through JSON config


Jille:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

I got this error from etcd when setting `"watch-progress-notify-interval": "1m",` in a JSON config. I assume that passing it as a flag from the commandline does work.

```
[etcdmain/etcd.go:66] failed to verify flags
error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field configYAML.watch-progress-notify-interval of type time.Duration
```

I suspect this is because `time.Duration` doesn't have an `UnmarshalJSON` method, so there's nothing doing the duration parsing. It can be given as an integer, but then it's interpreted as nanoseconds.

Both `watch-progress-notify-interval` and `experimental-watch-progress-notify-interval` exhibit this behavior.

### What did you expect to happen?

It to work the same as passing `--watch-progress-notify-interval=1m`.

### How can we reproduce it (as minimally and precisely as possible)?

Add `"experimental-watch-progress-notify-interval": "1m",` to JSON.

### Etcd version (please run commands below)

<details>

```console
etcd Version: 3.6.1
Git SHA: a4708be
Go Version: go1.23.10
Go OS/Arch: linux/amd64
```

</details>

BBQing:
I was able to reproduce it for other configuration options, that are durations - basically it fails if the configuration contains anything that is not integer (mostly it matters for duration strings

What is interesting in cli it works conversely - it fails with integer, but works with duration string

BBQing:
@Jille  So the issue is caused by the fact, that time.Duration is fancy way to present int64 - therefore it can be unmarshalled as integer in case it is provided from yaml configuration  - if it is provided from cli as flag it is processed by flag.DurationVar, which is fancy and more useful way of presenting the time.ParseDuration which accepts strings.

This is not consistent behaviour and good luck using nanoseconds in your yaml config :).

[...]

PR #20360, opened 2025-07-18
Fixed issue with duration strings used in yaml config file

BBQing:
Relates: #20342

This is a proposal on how to fix the issue. the problem is, that for the parameter --watch-progress-notify-interval, there is a difference in the behavior. While it is accepted as duration string from CLI, when provided in config_file, it has to be a number, that then gets converted into nanoseconds so for example for 1s you have to use 1000000000. Although it works well for machines it's bad for humans.

I have wrapped the time.duration, so that I can implement the UnmarshalJSON interface and also I have implemented flag.Value interface. And applied it where necessary.
I have also fixed some tests.

The fix supports both numbers and duration string from config file and in the future there can be issued warning, in case it will be decided to abandon the possibility to use numbers in the configuration.  

It might be necessary to do this for every duration parameter in the future.


server/embed/config.go
```diff
@@ -156,6 +156,47 @@ func init() {
 	defaultHostname, defaultHostStatus = netutil.GetDefaultHost()
 }
 
+type Duration struct {
```
BBQing: Should it be here in this file?

server/embed/config.go
```diff
@@ -156,6 +156,47 @@ func init() {
...
+	Duration time.Duration
+}
+
+func FromTimeDuration(duration time.Duration) Duration {
```
BBQing: Is this name expressive enough? The function is used as a "glue" in various tests.

BBQing:
@ivanvc could you check it?

ivanvc:
Please refer to my comment in https://github.com/etcd-io/etcd/issues/20342#issuecomment-3115403173.

server/embed/config.go
```diff
@@ -156,6 +156,47 @@ func init() {
...
+
+func (d *Duration) UnmarshalJSON(b []byte) error {
+	var v any
+	var err error
+	if err = yaml.Unmarshal(b, &v); err != nil {
+		return err
+	}
+	switch value := v.(type) {
+	case float64:
+		fmt.Println("float64", value)
```
ivanvc: Please remove this line.

```suggestion
```

server/embed/config.go
```diff
@@ -156,6 +156,46 @@ func init() {
 	defaultHostname, defaultHostStatus = netutil.GetDefaultHost()
 }
 
+type Duration struct {
+	Duration time.Duration
```
ivanvc: I left a comment here before, but I don't know what happened as it seems like GitHub dropped it :thinking:.

I suggested changing this to an anonymous field. I believe it would be transparent in cases like:

```
		WatchProgressNotifyInterval:       cfg.WatchProgressNotifyInterval.Duration,
```

We could revert this change.

```suggestion
	time.Duration
```

server/embed/etcd.go
```diff
@@ -224,7 +224,7 @@ func StartEtcd(inCfg *Config) (e *Etcd, err error) {
 		UnsafeNoFsync:                     cfg.UnsafeNoFsync,
 		CompactionBatchLimit:              cfg.CompactionBatchLimit,
 		CompactionSleepInterval:           cfg.CompactionSleepInterval,
-		WatchProgressNotifyInterval:       cfg.WatchProgressNotifyInterval,
+		WatchProgressNotifyInterval:       cfg.WatchProgressNotifyInterval.Duration,
```
ivanvc: With the anonymous `time.Duration` field, I believe this change is no longer necessary.

server/embed/config.go
```diff
@@ -156,6 +156,46 @@ func init() {
...
+		d.Duration, err = time.ParseDuration(value)
+
+		return err
+	}
+	return err
+}
+
+func (d Duration) String() string {
+	return d.Duration.String()
+}
```
ivanvc: And this is not required, either. As it would call `String()` from `d.Duration`.

server/embed/config.go
BBQing: I initially had it anonymous, but then I have decided to be more explicit - I have reverted the change

server/embed/etcd.go
BBQing: I have tried to use embed.Duration(current location) instead of time.Duration in the ServerConfig and it failed on the circular imports - I am thinking if the Duration should be placed in the pkg directory in its own package (or in some already existing package).
I'll continue with changes tomorrow. But thank you for the feedback - I am learning a lot.

ivanvc: Thanks, @BBQing. I'll keep an eye on this pull request.

server/embed/config.go
```diff
@@ -904,6 +905,10 @@ func (cfg *configYAML) configFromFile(path string) error {
 	if cfg.SelfSignedCertValidity == 0 {
 		cfg.SelfSignedCertValidity = 1
 	}
+
+	if cfg.WatchProgressNotifyIntervalJSON.Duration != 0 {
+		cfg.WatchProgressNotifyInterval = cfg.WatchProgressNotifyIntervalJSON.Duration
+	}
```
ivanvc: Suggestion to drop the `.Duration`

```suggestion
	if cfg.WatchProgressNotifyIntervalJSON != 0 {
		cfg.WatchProgressNotifyInterval = cfg.WatchProgressNotifyIntervalJSON
	}
```

tests/framework/e2e/cluster.go
```diff
@@ -359,7 +359,9 @@ func WithCompactionSleepInterval(time time.Duration) EPClusterOption {
 }
 
 func WithWatchProcessNotifyInterval(interval time.Duration) EPClusterOption {
-	return func(c *EtcdProcessClusterConfig) { c.ServerConfig.WatchProgressNotifyInterval = interval }
+	return func(c *EtcdProcessClusterConfig) {
+		c.ServerConfig.WatchProgressNotifyInterval = interval
+	}
```
ivanvc: Could you please revert this change?

server/embed/config_duration.go
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+package embed
+
+import (
+	"encoding/json"
+	"time"
+)
+
+type Duration struct {
```
ivanvc: This should be named `ConfigDuration` if the filename is `config_duration.go`.

server/embed/config.go
```diff
@@ -366,7 +366,8 @@ type Config struct {
 	// CompactionSleepInterval is the sleep interval between every etcd compaction loop.
 	CompactionSleepInterval time.Duration `json:"compaction-sleep-interval"`
 	// WatchProgressNotifyInterval is the time duration of periodic watch progress notifications.
-	WatchProgressNotifyInterval time.Duration `json:"watch-progress-notify-interval"`
+	WatchProgressNotifyInterval     time.Duration
+	WatchProgressNotifyIntervalJSON Duration `json:"watch-progress-notify-interval"`
```
ivanvc: What was the issue of keeping a single field?

ivanvc (commented):
Thanks for the updates, @BBQing. Please take a look at my comments :)

server/embed/config_duration.go
BBQing: I'll change that

tests/framework/e2e/cluster.go
BBQing: fixed

server/embed/config.go
BBQing: The problem for me was, that in the downstream the value is compared to some time.Duration constant (minimal watch progress interval - I could have retyped the constant into the new Duration type, but I think since the functionality of the wrapper type is related to configuration, it should remain there and not leak outside. So I have decided to follow approach used for different configuration fields, that also require some special treatment. beyond the standard unmarshalling

ivanvc: Makes sense, thanks for the clarification. Let's see what other contributors think :)

server/embed/config_duration.go
ivanvc: Thanks!

tests/framework/e2e/cluster.go
ivanvc: Thanks :)

server/embed/config.go
ivanvc: Thanks, I missed the casting. Thanks for adding that.

ivanvc:
This looks good to me. It fixes the issue for `watch-progress-notify-interval`. Ideally, after this pull request, we should have a follow-up to address the remaining duration fields.

However, I'm not an expert on etcdserver. @ahrtr, @fuweid, would you be able to please take a look at this? Thanks.

BBQing:
@ahrtr @fuweid Could you please check this PR? Thank you

ahrtr:
Thanks for the contribution, but please see my comment https://github.com/etcd-io/etcd/issues/20342#issuecomment-3172583838

PR #20369, opened 2025-07-21
[Antithesis] Enable Antithesis nightly workflow

joshjms:
Enable Antithesis Github workflow to run on PR and periodically.

/cc @serathius @henrybear327 @nwnt 

### DIFF
--- a/.github/workflows/antithesis-test.yml
+++ b/.github/workflows/antithesis-test.yml
@@ -2,43 +2,36 @@
 name: Build and trigger Antithesis exploration
 
 on:
-  # Disabled as discussed in https://github.com/etcd-io/etcd/pull/19750#issuecomment-2809840402
   # pull_request:
   #   branches: [main]
-  # schedule:
-  #   - cron: "0 0 * * *" # run every day at midnight
+  schedule:
+    - cron: "0 0 * * *" # run every day at midnight
   workflow_dispatch:
     inputs:
       test:
         description: 'Test name'
         required: false
-        default: 'etcd nightly antithesis run'
         type: string
       duration:
         description: 'Duration (exploration hours)'
         required: true
         type: int
-        default: 12
       description:
         description: 'Description (avoid quotes, please!)'
         required: true
         type: string
-        default: "etcd nightly antithesis run"
       etcd_ref:
         description: 'etcd version to build etcd-server from'
         required: false
         type: string
-        default: 'release-3.5'
       email:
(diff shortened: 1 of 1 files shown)


joshjms:
Btw, how can I test the workflow? @serathius

.github/workflows/antithesis-test.yml
```diff
@@ -2,11 +2,10 @@
 name: Build and trigger Antithesis exploration
 
 on:
-  # Disabled as discussed in https://github.com/etcd-io/etcd/pull/19750#issuecomment-2809840402
-  # pull_request:
-  #   branches: [main]
-  # schedule:
-  #   - cron: "0 0 * * *" # run every day at midnight
+  pull_request:
+    branches: [main, release-3.6, release-3.5, release-3.4]
```
serathius: Don't run on PR.

.github/workflows/antithesis-test.yml
```diff
@@ -52,7 +51,7 @@ jobs:
     runs-on: ubuntu-latest
     environment: Antithesis
     env:
-      CFG_NODE_COUNT: ${{ inputs.cfg_node_count }}
+      CFG_NODE_COUNT: ${{ inputs.cfg_node_count || 3 }}
```
serathius, line 48: No need to repeat defaults, they should come from `inputs`

joshjms: The inputs should only be accessible when triggered using workflow dispatch. I don't think it's accessible for `schedule` tasks like the cron jobs, so I had to make the defaults. CMIIW.

serathius: Oh Github, please remove defaults in input, having two sources of default values could easily lead to them getting desynchronized.

joshjms: Understood, I removed the default fields under `inputs`

serathius:
Ups I missed a mistake, there is no `latest` branch. The test is failing:
```
 > [ 2/25] RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=latest /etcd:
0.127 Cloning into '/etcd'...
0.480 warning: Could not find remote branch latest to clone.
0.480 fatal: Remote branch latest not found in upstream origin
```

Link: https://github.com/etcd-io/etcd/actions/runs/16510470974/job/46691249070

joshjms:
Oh, I didn't notice that. Will fix.

PR #20372, opened 2025-07-21
cache: implement storage for last‐observed state

apullo777:
@serathius @MadhavJivrajani

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -22,7 +22,7 @@ import (
 	"sync"
 	"time"
 
-	rpctypes "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
@@ -39,24 +39,18 @@ type Cache struct {
 	prefix      string // prefix is the key-prefix this shard is responsible for ("" = root).
 	cfg         Config // immutable runtime configuration
 	watcher     clientv3.Watcher
+	kv          clientv3.KV
 	demux       *demux // demux fans incoming events out to active watchers and manages resync.
+	store       *store // last‑observed snapshot
 	ready       chan struct{}
 	stop        context.CancelFunc
 	waitGroup   sync.WaitGroup
 	internalCtx context.Context
 }
 
-// watchCtx collects all the knobs that both serveWatchEvents and watchRetryLoop need.
-type watchCtx struct {
-	cache           *Cache
-	backoffStart    time.Duration
-	backoffMax      time.Duration
-	onFirstResponse func() // callback to fire once on first upstream response
-}
-
 // New builds a cache shard that watches only the requested prefix.
 // For the root cache pass "".
-func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error) {
+func New(client *clientv3.Client, prefix string, opts ...Option) (*Cache, error) {
 	cfg := defaultConfig()
 	for _, opt := range opts {
 		opt(&cfg)
@@ -71,7 +65,9 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
 	cache := &Cache{
 		prefix:      prefix,
 		cfg:         cfg,
-		watcher:     watcher,
+		watcher:     client.Watcher,
+		kv:          client.KV,
+		store:       newStore(),
 		ready:       make(chan struct{}),
 		stop:        cancel,
 		internalCtx: internalCtx,
@@ -82,15 +78,7 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
 	cache.waitGroup.Add(1)
 	go func() {
 		defer cache.waitGroup.Done()
-		readyOnce := sync.Once{}
-
-		watchCtx := &watchCtx{
-			cache:           cache,
-			backoffStart:    cfg.InitialBackoff,
-			backoffMax:      cfg.MaxBackoff,
-			onFirstResponse: func() { readyOnce.Do(func() { close(cache.ready) }) },
-		}
-		serveWatchEvents(internalCtx, watchCtx)
+		cache.getWatchLoop(internalCtx)
 	}()
 
 	return cache, nil
@@ -99,9 +87,7 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
 // Watch registers a cache-backed watcher for a given key or prefix.
 // It returns a WatchChan that streams WatchResponses containing events.
 func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption) clientv3.WatchChan {
-	select {
-	case <-c.ready:
-	case <-ctx.Done():
+	if err := c.WaitReady(ctx); err != nil {
 		emptyWatchChan := make(chan clientv3.WatchResponse)
 		close(emptyWatchChan)
 		return emptyWatchChan
@@ -162,6 +148,30 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	return responseChan
 }
 
+func (c *Cache) Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.GetResponse, error) {
+	if err := c.WaitReady(ctx); err != nil {
+		return nil, err
+	}
+	op := clientv3.OpGet(key, opts...)
+
+	if _, err := c.validateGet(key, op); err != nil {
+		return nil, err
+	}
+
+	startKey := []byte(key)
+	endKey := op.RangeBytes()
+	kvs, rev, err := c.store.Get(startKey, endKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return &clientv3.GetResponse{
+		Header: &pb.ResponseHeader{Revision: rev},
+		Kvs:    kvs,
+		Count:  int64(len(kvs)),
+	}, nil
+}
+
 // Ready reports whether the cache has finished its initial load.
 func (c *Cache) Ready() bool {
 	select {
@@ -176,9 +186,22 @@ func (c *Cache) Ready() bool {
 func (c *Cache) WaitReady(ctx context.Context) error {
 	select {
 	case <-c.ready:
-		return nil
+		return ctx.Err()
 	case <-ctx.Done():
 		return ctx.Err()
+	}
+}
+
+func (c *Cache) WaitForRevision(ctx context.Context, rev int64) error {
+	for {
+		if c.store.LatestRev() >= rev {
+			return nil
+		}
+		select {
+		case <-time.After(10 * time.Millisecond):
+		case <-ctx.Done():
+			return ctx.Err()
+		}
 	}
 }
 
@@ -188,60 +211,83 @@ func (c *Cache) Close() {
 	c.waitGroup.Wait()
 }
 
-func serveWatchEvents(ctx context.Context, watchCtx *watchCtx) {
-	backoff := watchCtx.backoffStart
+func (c *Cache) getWatchLoop(ctx context.Context) {
+	cfg := defaultConfig()
+	backoff := cfg.InitialBackoff
 	for {
-		opts := []clientv3.OpOption{
-			clientv3.WithPrefix(),
-			clientv3.WithProgressNotify(),
-			clientv3.WithCreatedNotify(),
-		}
-		if oldestRev := watchCtx.cache.demux.PeekOldest(); oldestRev != 0 {
-			opts = append(opts,
-				clientv3.WithRev(oldestRev+1))
-		}
-		watchCh := watchCtx.cache.watcher.Watch(ctx, watchCtx.cache.prefix, opts...)
-
-		if err := readWatchChannel(watchCh, watchCtx.cache, watchCtx.cache.demux, watchCtx.onFirstResponse); err == nil {
+		if err := ctx.Err(); err != nil {
 			return
 		}
-
+		if err := c.getWatch(ctx); err != nil {
+			fmt.Printf("getWatch failed, will retry after %v: %v\n", backoff, err)
+		}
 		select {
 		case <-ctx.Done():
 			return
 		case <-time.After(backoff):
 		}
-		if backoff < watchCtx.backoffMax {
-			backoff *= 2
+	}
+}
+
+func (c *Cache) getWatch(ctx context.Context) error {
+	getResp, err := c.get(ctx)
+	if err != nil {
+		return err
+	}
+	return c.watch(ctx, getResp.Header.Revision+1)
+}
+
+func (c *Cache) get(ctx context.Context) (*clientv3.GetResponse, error) {
+	resp, err := c.kv.Get(ctx, c.prefix, clientv3.WithPrefix())
+	if err != nil {
+		return nil, err
+	}
+	c.store.Restore(resp.Kvs, resp.Header.Revision)
+	return resp, nil
+}
+
+func (c *Cache) watch(ctx context.Context, rev int64) error {
+	readyOnce := sync.Once{}
+	for {
+		watchCh := c.watcher.Watch(
+			ctx,
+			c.prefix,
+			clientv3.WithPrefix(),
+			clientv3.WithRev(rev),
+			clientv3.WithProgressNotify(),
+			clientv3.WithCreatedNotify(),
+		)
+
+		for resp := range watchCh {
+			readyOnce.Do(func() { close(c.ready) })
+			if err := resp.Err(); err != nil {
+				c.demux.Purge()
+				c.resetReady()
+				c.store.Reset()
+				return err
+			}
+
+			if err := c.store.Apply(resp.Events); err != nil {
+				c.demux.Purge()
+				c.resetReady()
+				c.store.Reset()
+				return err
+			}
+			c.demux.Broadcast(resp.Events)
+		}
+
+		if ctx.Err() != nil {
+			return ctx.Err()
 		}
 	}
 }
 
-// readWatchChannel reads an etcd Watch stream into History and enqueueCh, returning nil on cancel or first error.
-func readWatchChannel(
-	watchChan clientv3.WatchChan,
-	cache *Cache,
-	demux *demux,
-	onFirstResponse func(),
-) error {
-	for resp := range watchChan {
-		onFirstResponse()
-
-		if err := resp.Err(); err != nil {
-			if errors.Is(err, rpctypes.ErrCompacted) {
-				select {
-				case <-cache.ready:
-					// TODO: Reinitialize cache.ready safely; current direct channel assignment can race with concurrent watchers
-					cache.ready = make(chan struct{})
-				default:
-				}
-				demux.Purge()
-			}
-			return err
-		}
-		demux.Broadcast(resp.Events)
+func (c *Cache) resetReady() {
+	select {
+	case <-c.ready:
+		c.ready = make(chan struct{})
+	default:
 	}
-	return nil
 }
 
 func (c *Cache) validateWatch(key string, op clientv3.Op) (pred KeyPredicate, err error) {
@@ -266,6 +312,41 @@ func (c *Cache) validateWatch(key string, op clientv3.Op) (pred KeyPredicate, er
 	if err := c.validateRange(startKey, endKey); err != nil {
 		return nil, err
 	}
+	return KeyPredForRange(startKey, endKey), nil
+}
+
+func (c *Cache) validateGet(key string, op clientv3.Op) (KeyPredicate, error) {
+	switch {
+	case op.IsCountOnly():
+		return nil, fmt.Errorf("%w: CountOnly not supported", ErrUnsupportedRequest)
+	case op.IsPrevKV():
+		return nil, fmt.Errorf("%w: PrevKV not supported", ErrUnsupportedRequest)
+	case op.IsSortSet():
+		return nil, fmt.Errorf("%w: SortSet not supported", ErrUnsupportedRequest)
+	case op.Limit() != 0:
+		return nil, fmt.Errorf("%w: Limit(%d) not supported", ErrUnsupportedRequest, op.Limit())
+	case op.MinModRev() != 0:
+		return nil, fmt.Errorf("%w: MinModRev(%d) not supported", ErrUnsupportedRequest, op.MinModRev())
+	case op.MaxModRev() != 0:
+		return nil, fmt.Errorf("%w: MaxModRev(%d) not supported", ErrUnsupportedRequest, op.MaxModRev())
+	case op.MinCreateRev() != 0:
+		return nil, fmt.Errorf("%w: MinCreateRev(%d) not supported", ErrUnsupportedRequest, op.MinCreateRev())
+	case op.MaxCreateRev() != 0:
+		return nil, fmt.Errorf("%w: MaxCreateRev(%d) not supported", ErrUnsupportedRequest, op.MaxCreateRev())
+	// cache now only serves serializable reads of the latest revision (rev == 0).
+	case op.Rev() != 0:
+		return nil, fmt.Errorf("%w: Rev(%d) not supported", ErrUnsupportedRequest, op.Rev())
+	case !op.IsSerializable():
+		return nil, fmt.Errorf("%w: non-serializable request", ErrUnsupportedRequest)
+	}
+
+	startKey := []byte(key)
+	endKey := op.RangeBytes()
+
+	if err := c.validateRange(startKey, endKey); err != nil {
+		return nil, err
+	}
+
 	return KeyPredForRange(startKey, endKey), nil
 }
 
--- a/cache/cache_test.go
+++ b/cache/cache_test.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	mvccpb "go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
@@ -188,7 +189,14 @@ func TestCacheWatchAtomicOrderedDelivery(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			mw := newMockWatcher(16)
-			cache, err := New(mw, "")
+			fakeClient := &clientv3.Client{
+				Watcher: mw,
+				KV:      &kvStub{},
+			}
+			cache, err := New(fakeClient, "")
+			if err != nil {
+				t.Fatalf("New cache: %v", err)
+			}
 			if err != nil {
 				t.Fatalf("New cache: %v", err)
 			}
@@ -395,6 +403,32 @@ type mockWatcher struct {
 	registered chan struct{}
 }
 
+type kvStub struct{}
+
+func (kvStub) Get(ctx context.Context, key string, _ ...clientv3.OpOption) (*clientv3.GetResponse, error) {
+	return &clientv3.GetResponse{Header: &pb.ResponseHeader{Revision: 0}}, nil
+}
+
+func (kvStub) Put(ctx context.Context, key, val string, _ ...clientv3.OpOption) (*clientv3.PutResponse, error) {
+	return nil, nil
+}
+
+func (kvStub) Delete(ctx context.Context, key string, _ ...clientv3.OpOption) (*clientv3.DeleteResponse, error) {
+	return nil, nil
+}
+
+func (kvStub) Compact(ctx context.Context, rev int64, _ ...clientv3.CompactOption) (*clientv3.CompactResponse, error) {
+	return nil, nil
+}
+
+func (kvStub) Do(ctx context.Context, op clientv3.Op) (clientv3.OpResponse, error) {
+	return clientv3.OpResponse{}, nil
+}
+
+func (kvStub) Txn(ctx context.Context) clientv3.Txn {
+	return nil
+}
+
 func newMockWatcher(buf int) *mockWatcher {
 	return &mockWatcher{
 		responses:  make(chan clientv3.WatchResponse, buf),
@@ -403,7 +437,11 @@ func newMockWatcher(buf int) *mockWatcher {
 }
 
 func (m *mockWatcher) Watch(_ context.Context, _ string, _ ...clientv3.OpOption) clientv3.WatchChan {
-	close(m.registered)
+	select { // prevent “close of closed channel” on re-watch
+	case <-m.registered:
+	default:
+		close(m.registered)
+	}
 	return m.responses
 }
 
--- a/cache/config.go
+++ b/cache/config.go
@@ -29,6 +29,8 @@ type Config struct {
 	InitialBackoff time.Duration
 	// MaxBackoff caps the exponential back-off between successive upstream watch retries.
 	MaxBackoff time.Duration
+	// GetTimeout is the timeout applied to the first Get() used to bootstrap the cache.
+	GetTimeout time.Duration
 }
 
 // TODO: tune via performance/load tests.
@@ -39,6 +41,7 @@ func defaultConfig() Config {
 		ResyncInterval:       50 * time.Millisecond,
 		InitialBackoff:       50 * time.Millisecond,
 		MaxBackoff:           2 * time.Second,
+		GetTimeout:           5 * time.Second,
 	}
 }
 
@@ -62,4 +65,8 @@ func WithInitialBackoff(d time.Duration) Option {
 
 func WithMaxBackoff(d time.Duration) Option {
 	return func(c *Config) { c.MaxBackoff = d }
+}
+
+func WithGetTimeout(d time.Duration) Option {
+	return func(c *Config) { c.GetTimeout = d }
 }
--- a/cache/store.go
+++ b/cache/store.go
@@ -1,0 +1,145 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"bytes"
+	"fmt"
+	"sort"
+	"sync"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+var ErrNotInitialized = fmt.Errorf("cache: store not initialized")
+
+type store struct {
+	mu        sync.RWMutex
+	kvs       map[string]*mvccpb.KeyValue
+	latestRev int64
+}
+
+func newStore() *store {
+	return &store{kvs: make(map[string]*mvccpb.KeyValue)}
+}
+
+func (s *store) Get(startKey, endKey []byte) ([]*mvccpb.KeyValue, int64, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	if s.latestRev == 0 {
+		return nil, 0, ErrNotInitialized
+	}
+
+	var out []*mvccpb.KeyValue
+	switch {
+	case len(endKey) == 0:
+		out = s.getSingle(startKey)
+	case isPrefixScan(endKey):
+		out = s.scanPrefix(startKey)
+	default:
+		out = s.scanRange(startKey, endKey)
+	}
+
+	sort.Slice(out, func(i, j int) bool {
+		return bytes.Compare(out[i].Key, out[j].Key) < 0 // default: lexicographical, ascending‐by‐key sort
+	})
+	return out, s.latestRev, nil
+}
+
+func (s *store) Restore(kvs []*mvccpb.KeyValue, rev int64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.kvs = make(map[string]*mvccpb.KeyValue, len(kvs))
+	for _, kv := range kvs {
+		s.kvs[string(kv.Key)] = kv
+	}
+	s.latestRev = rev
+}
+
+// Reset purges all in-memory state when the upstream watch stream reports compaction.
+func (s *store) Reset() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.kvs = make(map[string]*mvccpb.KeyValue)
+	s.latestRev = 0
+}
+
+func (s *store) Apply(events []*clientv3.Event) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for _, ev := range events {
+		if ev.Kv.ModRevision < s.latestRev {
+			return fmt.Errorf("cache: stale event batch (rev %d < latest %d)", ev.Kv.ModRevision, s.latestRev)
+		}
+	}
+
+	for _, ev := range events {
+		switch ev.Type {
+		case clientv3.EventTypeDelete:
+			delete(s.kvs, string(ev.Kv.Key))
+		case clientv3.EventTypePut:
+			s.kvs[string(ev.Kv.Key)] = ev.Kv
+		}
+		if ev.Kv.ModRevision > s.latestRev {
+			s.latestRev = ev.Kv.ModRevision
+		}
+	}
+	return nil
+}
+
+func (s *store) LatestRev() int64 {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.latestRev
+}
+
+// getSingle fetches one key or nil
+func (s *store) getSingle(key []byte) []*mvccpb.KeyValue {
+	if kv, ok := s.kvs[string(key)]; ok {
+		return []*mvccpb.KeyValue{kv}
+	}
+	return nil
+}
+
+// scanPrefix returns all keys >= startKey
+func (s *store) scanPrefix(startKey []byte) []*mvccpb.KeyValue {
+	var res []*mvccpb.KeyValue
+	for _, kv := range s.kvs {
+		if bytes.Compare(kv.Key, startKey) >= 0 {
+			res = append(res, kv)
+		}
+	}
+	return res
+}
+
+// scanRange returns all keys in [startKey, endKey)
+func (s *store) scanRange(startKey, endKey []byte) []*mvccpb.KeyValue {
+	var res []*mvccpb.KeyValue
+	for _, kv := range s.kvs {
+		if bytes.Compare(kv.Key, startKey) >= 0 && bytes.Compare(kv.Key, endKey) < 0 {
+			res = append(res, kv)
+		}
+	}
+	return res
+}
+
+// isPrefixScan detects endKey=={0} semantics
+func isPrefixScan(endKey []byte) bool {
+	return len(endKey) == 1 && endKey[0] == 0
+}
--- a/tests/integration/cache_test.go
+++ b/tests/integration/cache_test.go
@@ -374,7 +374,7 @@ func testWatch(t *testing.T, kv clientv3.KV, watcher Watcher) {
 	}
 }
 
-func TestCacheWithPrefix(t *testing.T) {
+func TestCacheWithPrefixWatch(t *testing.T) {
 	integration.BeforeTest(t)
 	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
 	t.Cleanup(func() { clus.Terminate(t) })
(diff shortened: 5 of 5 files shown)


tests/integration/cache_test.go
```diff
@@ -485,6 +485,122 @@ func TestCacheWithPrefix(t *testing.T) {
...
+		wantEvents map[string]string // expected last observed state {key:val}
+	}{
+		{
+			name: "single put",
+			events: []struct{ key, val string }{
+				{"/foo/a", "v1"},
+			},
+			key:        "/foo",
+			opts:       []clientv3.OpOption{clientv3.WithPrefix()},
+			wantEvents: map[string]string{"/foo/a": "v1"},
```
serathius: testGet should just test Get. Remove `events` and `wantEvents`. 
Please compare `Kvs` and `header.revision` from response.

apullo777: Ok. To double check, I should drop the `events` & `wantEvents` as they are specific to Watch semantics, and to have fields that Get care about like below: 

```
type tests struct {
    name        string
    key         string
    opts        []clientv3.OpOption
    puts        []struct{ key, val string } 
    wantKVs     map[string]string  
}
```
and we do assertions like below:

```
got := map[string]string{}
for _, kv := range resp.Kvs {
    got[string(kv.Key)] = string(kv.Value)
}
if diff := cmp.Diff(tc.wantKVs, got); diff != "" {
    t.Fatalf("unexpected KVs (-want +got):\n%s", diff)
}
if resp.Header.Revision != lastRev {
    t.Fatalf("revision: got %d, want %d", resp.Header.Revision, lastRev)
}
```

serathius: `puts` should not be test specific, but a shared between all test cases.

serathius: Also add `wantRevision`

cache/cache.go
```diff
@@ -26,8 +26,11 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
-// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
-var ErrUnsupportedWatch = errors.New("cache: unsupported watch parameters")
+var (
+	// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
+	ErrUnsupportedWatch = errors.New("cache: unsupported watch parameters")
+	ErrNotImplemented   = errors.New("cache.Get: not implemented yet")
```
serathius: Don't know what's the difference between the two errors. Any reason to have Get and Watch specific errors?

tests/integration/cache_test.go
```diff
@@ -485,6 +485,196 @@ func TestCacheWithPrefix(t *testing.T) {
...
+		},
+		{
+			name:         "prefix /foo",
+			key:          "/foo",
+			opts:         []clientv3.OpOption{clientv3.WithPrefix()},
+			wantKVs:      map[string]string{"/foo/a": "a3", "/foo/b": "b2", "/foo/c": "c1"},
+			wantRevision: 8,
+		},
+		{
+			name:         "range [/foo/a, /foo/b)",
```
serathius:
```suggestion
			name:         "range [/foo/a, /foo/c)",
```

tests/integration/cache_test.go
```diff
@@ -485,6 +485,196 @@ func TestCacheWithPrefix(t *testing.T) {
...
+
+	tests := []struct {
+		name         string
+		key          string
+		opts         []clientv3.OpOption
+		wantKVs      map[string]string
+		wantRevision int64
+	}{
+		{
+			name:         "single key /foo/a",
```
serathius, line 605: Please test Get on non existing key

tests/integration/cache_test.go
```diff
@@ -485,6 +485,196 @@ func TestCacheWithPrefix(t *testing.T) {
...
+			ModRevision:    8,
+			Version:        3,
+		},
+	}
+
+	tests := []struct {
+		name         string
+		key          string
+		opts         []clientv3.OpOption
+		wantKVs      map[string]string
```
serathius, line 600: Please not only test `keyValue`, but the whole `KeyValue` including fields like `CreateRevision`, `ModRevision` and `Version`.

tests/integration/cache_test.go
```diff
@@ -485,6 +485,196 @@ func TestCacheWithPrefix(t *testing.T) {
...
+		wantRevision int64
+	}{
+		{
+			name:         "single key /foo/a",
+			key:          "/foo/a",
+			opts:         nil,
+			wantKVs:      map[string]string{"/foo/a": "a3"},
```
serathius:
```suggestion
			wantKVs:      []*mvccpb.KeyValue{rev2PutFooA.Kv}
```

tests/integration/cache_test.go
serathius:
```suggestion
		wantKVs      []*mvccpb.KeyValue
```
Please not only test `keyValue`, but the whole `KeyValue` including fields like `CreateRevision`, `ModRevision` and `Version`.

tests/integration/cache_test.go
```diff
@@ -485,6 +485,196 @@ func TestCacheWithPrefix(t *testing.T) {
...
+func TestGet(t *testing.T) {
+	integration.BeforeTest(t)
+	clus := integration.NewCluster(t, &integration.ClusterConfig{Size: 1})
+	t.Cleanup(func() { clus.Terminate(t) })
+
+	client := clus.Client(0)
+	testGet(t, client.KV, client.KV)
+}
+
+func testGet(t *testing.T, kv clientv3.KV, reader Getter) {
```
serathius, line 517: Please consider adding also a test for unsupported arguments of Get, similar to what you did for watch.
And a test for Get where the cache has prefix, also same like what you did for watch.

cache/cache.go
apullo777: Ok. Maybe just use one single error named ErrUnsupportedRequest

serathius: Sounds good. Please keep in mind there is a differences between unsupported and OutOfBounds (name could be improved) request for cache. A request for option that is not supoorted like WithPrevKV should return unsupported. A request for "/foo/a" key when cache monitors "/bar" prefix is OutOfBounds.

apullo777: Got it! Then perhaps we also have an error named `ErrKeyOutOfRange` or `ErrKeyRangeInvalid` to cover cases for both "outside cache prefix key range" & "empty/reversed ranges" like range [/foo, /foo) or [/foo/b, /foo/a). Like below:

```
// Returned when an option combination isn’t yet handled by the cache (e.g. WithPrevKV, WithProgressNotify for Watch(), WithCountOnly for Get()).
ErrUnsupportedRequest = errors.New("cache: unsupported request parameters")

// Returned when the requested key or key‑range is invalid (empty or reversed) or lies outside c.prefix.
ErrKeyRangeInvalid = errors.New("cache: invalid or out‑of‑range key range")
```

cache/cache.go
```diff
@@ -26,8 +26,13 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
-// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
-var ErrUnsupportedWatch = errors.New("cache: unsupported watch parameters")
+var (
+	// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
```
serathius: Don't think this TODO is relevant anymore

tests/integration/cache_test.go
```diff
@@ -486,6 +486,291 @@ func TestCacheWithPrefix(t *testing.T) {
...
+	t.Cleanup(func() { clus.Terminate(t) })
+	client := clus.Client(0)
+
+	ctx := t.Context()
+
+	tests := []struct {
+		name        string
+		key         string
+		opts        []clientv3.OpOption
+		expectError bool
```
serathius, line 703: With errors for invalid prefix defined maybe would be good to check the type of error not only fact that error exist. This could be done in followup.

cache/cache.go
```diff
@@ -271,25 +280,25 @@ func (c *Cache) validateWatchRange(startKey, endKey []byte) error {
 			return nil
 		}
 		if bytes.Compare(startKey, prefixStart) < 0 || bytes.Compare(startKey, prefixEnd) >= 0 {
-			return ErrUnsupportedWatch
+			return ErrKeyRangeInvalid
```
serathius: Please consider splitting this changes to separate PR that can be merged earlier.

tests/integration/cache_test.go
```diff
@@ -605,9 +890,50 @@ func TestCacheUnsupportedWatchOptions(t *testing.T) {
...
+	defer c.Close()
+	if err := c.WaitReady(t.Context()); err != nil {
+		t.Fatalf("cache not ready: %v", err)
+	}
+
+	unsupported := []struct {
+		name string
+		opt  clientv3.OpOption
+	}{
+		{"WithSerializable", clientv3.WithSerializable()},
```
serathius: Technically the cache now doesn't follow the same default as client. By default client requests are Linearizable meaning newest globally (approximate definition), which is reverse of Serializable meaning newest locally (approximate definition). While cache only serves newest locally to cache. 

Could we make that tests in cache always passes WithSerializable and it's a required setting?

apullo777: Got it. I will remove the "WithSerializable" rejection test here, and in all test cases we add `clientv3.WithSerializable()` in opts field (or do something like `opts = append(opts, clientv3.WithSerializable())`)

MadhavJivrajani: +1 to what @serathius said

I was thinking more about this earlier, and I see that the existing grpcproxy serves a response from its cache if WithSerializable() is set
https://github.com/etcd-io/etcd/blob/f79fc2c4d6d9500e175c70520b50793dd8063ffb/server/proxy/grpcproxy/kv.go#L42

For the cache here, do we want to reject requests that does _not_ have this set for now? And maybe in a future iteration, make a call into etcd?

serathius: We temporarily will reject those requests, until we implement consistent read.

cache/cache.go
```diff
@@ -27,7 +27,6 @@ import (
 )
 
 var (
-	// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
```
serathius: Please remove unrelated changes from this PR.

cache/cache.go
```diff
@@ -72,13 +75,27 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
+		store:       newStore(),
 		ready:       make(chan struct{}),
 		stop:        cancel,
 		internalCtx: internalCtx,
 	}
 
 	cache.demux = NewDemux(internalCtx, &cache.waitGroup, cfg.HistoryWindowSize, cfg.ResyncInterval)
 
+	var startRev int64
+	if kv, ok := watcher.(clientv3.KV); ok {
```
apullo777: Since in `cache.New` we don’t just call `Watch` anymore, but also need to call `Get` to bootstrap the snapshot. Do we need to introduce something like a `KVWatcher` interface like below?

```
type KVWatcher interface {
    clientv3.KV
    clientv3.Watcher
}
```
so that we can change `cache.New`'s  ignature like below:

`func New(kvWatcher KVWatcher, prefix string, opts ...Option) (*Cache, error)`

and to rewrite this section like this:
```
var startRev int64
getCtx, cancelGet := context.WithTimeout(internalCtx, 5*time.Second)
defer cancelGet()
resp, err := kvWatcher.Get(getCtx, prefix, clientv3.WithPrefix(), clientv3.WithSerializable())
if err != nil {
    cancel()
    return nil, err
}
cache.store.initializeStoreState(resp.Kvs, resp.Header.Revision)
startRev = resp.Header.Revision + 1
```

cache/cache.go
```diff
@@ -72,13 +75,27 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
+		getCtx, cancelGet := context.WithTimeout(internalCtx, 5*time.Second)
+		defer cancelGet()
+		resp, err := kv.Get(getCtx, prefix, clientv3.WithPrefix(), clientv3.WithSerializable())
+		if err != nil {
+			cancel()
+			return nil, err
+		}
+		cache.store.initializeStoreState(resp.Kvs, resp.Header.Revision)
```
apullo777: I noticed that our current tests do not cover a scenario where we preload keys before starting the cache to verify that the bootstrap Get was correctly loaded into our store.

cache/cache.go
```diff
@@ -22,12 +22,13 @@ import (
 	"sync"
 	"time"
 
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	rpctypes "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
+// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
```
MadhavJivrajani: @apullo777 see if you can revert this comment change. Ideally it shouldn't show up as a diff here.

cache/cache.go
```diff
@@ -52,6 +54,7 @@ type watchCtx struct {
 	backoffStart    time.Duration
 	backoffMax      time.Duration
 	onFirstResponse func() // callback to fire once on first upstream response
+	startRev        int64  // first revision to watch from (resp.Header.Revision+1)
```
MadhavJivrajani: Let's not add impl. details as part of the comment, makes it hard to follow along. Either remove or just `first revision to watch from` is fine

cache/cache.go
```diff
@@ -72,13 +75,27 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
 		ready:       make(chan struct{}),
 		stop:        cancel,
 		internalCtx: internalCtx,
 	}
 
 	cache.demux = NewDemux(internalCtx, &cache.waitGroup, cfg.HistoryWindowSize, cfg.ResyncInterval)
 
+	var startRev int64
+	if kv, ok := watcher.(clientv3.KV); ok {
+		getCtx, cancelGet := context.WithTimeout(internalCtx, 5*time.Second)
```
MadhavJivrajani: How have you chosen 5 seconds here?

cache/cache.go
```diff
@@ -257,13 +299,38 @@ func (c *Cache) validateWatch(key string, op clientv3.Op) (pred KeyPredicate, er
...
+		op.Limit() != 0 ||
+		op.MinModRev() != 0 ||
+		op.MaxModRev() != 0 ||
+		op.MinCreateRev() != 0 ||
+		op.MaxCreateRev() != 0 ||
+		// cache now only serves serializable reads of the latest revision (rev == 0).
+		op.Rev() != 0 ||
+		!op.IsSerializable() {
+		return nil, ErrUnsupportedRequest
+	}
```
MadhavJivrajani: This check if getting heavy. In a follow up, let's refactor this to something like:
```go
switch {
    case op.IsCountOnly():
        return nil, fmt.Errorf("%w: CountOnly() not supported", ErrUnsupportedRequest)
    case op.IsPrevKV():
        return nil, fmt.Errorf("%w: PrevKV() not supported", ErrUnsupportedRequest)
    ...
}
```
We should get some insight into what option was set and not supported.

cache/cache.go
```diff
@@ -72,13 +75,27 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
 		internalCtx: internalCtx,
 	}
 
 	cache.demux = NewDemux(internalCtx, &cache.waitGroup, cfg.HistoryWindowSize, cfg.ResyncInterval)
 
+	var startRev int64
+	if kv, ok := watcher.(clientv3.KV); ok {
+		getCtx, cancelGet := context.WithTimeout(internalCtx, 5*time.Second)
+		defer cancelGet()
+		resp, err := kv.Get(getCtx, prefix, clientv3.WithPrefix(), clientv3.WithSerializable())
```
MadhavJivrajani: We should be a little careful here. The prefix might be a one with a large number of keys in etcd, `GET` on the key and using it to initialize the store can be expensive to do. We can maybe do a `Get()` with a [`Limit=1`](https://pkg.go.dev/go.etcd.io/etcd/client/v3#WithLimit) because what's important is getting the rev from this response and starting a watch from there to populate the store. 

Wdyt?

cache/cache.go
MadhavJivrajani: I think we can change the signature to take in `client clientv1.Client` instead and use client.KV or client.Watcher wherever needed

cache/store.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"bytes"
+	"sort"
+	"sync"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+type store struct {
```
MadhavJivrajani, line 29: How is the store being updated from the watch? Or is that yet to be impl.?

cache/cache.go
```diff
@@ -162,6 +180,28 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
+
+	pred, err := c.validateGet(key, op)
+	if err != nil {
+		return nil, err
+	}
+
+	kvs, rev := c.store.selectKVs(pred)
+
+	return &clientv3.GetResponse{
+		Header: &pb.ResponseHeader{Revision: rev},
```
MadhavJivrajani, line 169: As the store gets updated from the watch, we should try and reuse the `ResponseHeader` from the watch response.

apullo777: Thanks! That makes sense. In that way we can preserve the full metadata in every cached response, and only overwriting its `Revision` field with the cache’s current revision

cache/store.go
apullo777: That happens in `readWatchChannel` like below:

```
for resp := range watchChan {
    …
    cache.store.apply(resp.Events)
    demux.Broadcast(resp.Events)
}
```

cache/cache.go
apullo777: just picked 5 seconds as a safe number. Would you recommend pull that value from existing client timeout, or make it an explicit knob like `InitialLoadTimeout` in the config?

cache/cache.go
apullo777: Thanks for flagging this. Perhaps we could split into two phases? (1) do a Get with `WithLimit(1)` to fetch the latest revision quickly, and then (2) start watching from that revision to capture all updates.

cache/cache.go
serathius: Good point! I would recommend to keep the same scenarios to avoid duplicates, but cache in 3 separate scenarios:
* From empty etcd
* From partial state
* From complete state.

cache/cache.go
serathius: > just picked 5 seconds as a safe number. 

It's not `safe number`, it's an `arbitrary number` without any data behind it. Please move it to config.

cache/cache.go
serathius: Recommend to remove this

cache/cache.go
serathius: This block should not be called once on cache start, but every time watch is broken and

cache/cache.go
serathius: Using serializable here is definitely not correct

cache/cache.go
```diff
@@ -162,6 +180,28 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
+		return nil, ctx.Err()
+	}
+	op := clientv3.OpGet(key, opts...)
+
+	pred, err := c.validateGet(key, op)
+	if err != nil {
+		return nil, err
+	}
+
+	kvs, rev := c.store.selectKVs(pred)
```
```suggestion
	kvs, rev := c.store.Get(pred)
```

cache/cache.go
```diff
@@ -197,8 +237,9 @@ func serveWatchEvents(ctx context.Context, watchCtx *watchCtx) {
 			clientv3.WithCreatedNotify(),
 		}
 		if oldestRev := watchCtx.cache.demux.PeekOldest(); oldestRev != 0 {
-			opts = append(opts,
-				clientv3.WithRev(oldestRev+1))
+			opts = append(opts, clientv3.WithRev(oldestRev+1))
+		} else if watchCtx.startRev != 0 {
+			opts = append(opts, clientv3.WithRev(watchCtx.startRev))
```
serathius, line 242: This is wrong you first should Get, and then Watch, so the demux should not matter here anymore.

cache/cache.go
serathius: No, we definitely should not. Header includes member information that we don't want to be provided with random values depending on where the event comes from.

cache/cache.go
serathius: No, we should have 1 Get for all objects. We can implement pagination later.

cache/cache.go
```diff
@@ -239,6 +280,7 @@ func readWatchChannel(
 			}
 			return err
 		}
+		cache.store.apply(resp.Events)
```
serathius: store should be subscribed to demux as any other watcher.

cache/store.go
```diff
@@ -0,0 +1,72 @@
...
+	latestRev int64
+}
+
+func newStore() *store {
+	return &store{kvs: make(map[string]*mvccpb.KeyValue)}
+}
+
+func (s *store) initializeStoreState(kvs []*mvccpb.KeyValue, rev int64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
serathius, line 77: First purge store before writing new keys.

cache/store.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	s.latestRev = rev
+}
+
+func (s *store) apply(events []*clientv3.Event) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for _, ev := range events {
+		if ev.Type == clientv3.EventTypeDelete {
```
serathius: make it a switch statement.

cache/store.go
```diff
@@ -0,0 +1,72 @@
...
+			delete(s.kvs, string(ev.Kv.Key))
+		} else {
+			s.kvs[string(ev.Kv.Key)] = ev.Kv
+		}
+		if ev.Kv.ModRevision > s.latestRev {
```
serathius, line 99: If event mod revision is less than latestRev we are in very bad situation. We should rather reject whole batch of events than apply them

cache/store.go
```diff
@@ -0,0 +1,72 @@
...
+			s.latestRev = ev.Kv.ModRevision
+		}
+	}
+}
+
+func (s *store) selectKVs(pred KeyPredicate) ([]*mvccpb.KeyValue, int64) {
```
serathius:
```suggestion
func (s *store) Get(pred KeyPredicate) ([]*mvccpb.KeyValue, int64) {
```

serathius: They key predicate as argument is a bad idea. Predicate is not a first type citizen. Just use `Get(startKey, endKey string)`

cache/store.go
```diff
@@ -0,0 +1,72 @@
...
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	var out []*mvccpb.KeyValue
+	for _, kv := range s.kvs {
```
serathius, line 123: What happens when someone just wants to get single key, you iterate all keys to just find one instead of map lookup?

cache/cache.go
apullo777: Got it. Thanks! just to double check do you mean we shouldn’t reuse the watch’s ResponseHeader at all (and instead always construct a fresh header with only the fields we control)?

serathius: Yes, we can discuss this later, but we need to analyse each field in ResponseHeader and make dedicated decision for each. For example we could return cluster id, but we that would require us to detect if it changes.

cache/cache.go
```diff
@@ -191,14 +235,27 @@ func (c *Cache) Close() {
...
+		if kv, ok := watchCtx.cache.watcher.(clientv3.KV); ok {
+			if resp, err := kv.Get(ctx, watchCtx.cache.prefix, clientv3.WithPrefix()); err == nil {
+				events := make([]*clientv3.Event, 0, len(resp.Kvs))
+				for _, kv := range resp.Kvs {
+					events = append(events, &clientv3.Event{
+						Type: clientv3.EventTypePut,
+						Kv:   kv,
+					})
+				}
+				watchCtx.cache.demux.Broadcast(events)
```
apullo777: Curious if there will be a race when a caller do `Cache.Get()` immediately after `Ready(),` and demux goroutine hasn’t dequeued/applied those events yet so store.kvs would still be empty.

apullo777:
Flake in integration test:
```
=== FAIL: integration TestCacheWithPrefixGet/single_key_within_prefix (0.00s)
    cache_test.go:768: Get "/foo/a" returned no KVs, expected at least one
    --- FAIL: TestCacheWithPrefixGet/single_key_within_prefix (0.00s)
```
Might be that `WaitReady` returns as soon as the initial `CreatedNotify` arrives, but before the actual PUT event has been applied. 

Maybe we should: (1) make the test wait for the cache to actually apply the PUT (2) change `WaitReady` to only close after the first real watch event so it truly reflects up-to-date.

cache/cache.go
apullo777: Still trying to figure out how to write these tests. Since we only want to test Cache with these scenarios so perhaps a new test called `TestCacheBootstrapScenarios` that use the existing `testGet` with test table like below?

```
{
			name: "complete‑state",
			preload: func(ctx context.Context, kv clientv3.KV) {
				kv.Put(ctx, "/foo/a", "1")
				kv.Put(ctx, "/foo/b", "2")
				kv.Put(ctx, "/foo/c", "3")
			},
			wantKVs: []*mvccpb.KeyValue{...},
		},
```

cache/store.go
```diff
@@ -0,0 +1,98 @@
...
+	"sync"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+type store struct {
+	mu        sync.RWMutex
+	kvs       map[string]*mvccpb.KeyValue
+	latestRev int64
```
apullo777, line 32: With the current integration results seems like stale reads happen as `latestRev` lags behind writes under its lock. Should we switch `latestRev` to `atomic.Int64 `that writer update after applying a batch?

cache/cache.go
```diff
@@ -72,13 +74,32 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
 		watcher:     watcher,
+		store:       newStore(),
 		ready:       make(chan struct{}),
 		stop:        cancel,
 		internalCtx: internalCtx,
 	}
 
 	cache.demux = NewDemux(internalCtx, &cache.waitGroup, cfg.HistoryWindowSize, cfg.ResyncInterval)
 
+	storeW := newWatcher(cfg.HistoryWindowSize, nil) // nil predicate == match all
```
serathius: Move to separate function. Please try to reuse code from Watch if possible.

cache/cache.go
```diff
@@ -72,13 +74,32 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
+	cache.demux.Register(storeW, 0)
+	cache.waitGroup.Add(1)
+	go func() {
+		defer cache.waitGroup.Done()
+		for {
+			select {
+			case <-cache.internalCtx.Done():
+				return
+			case evs, ok := <-storeW.eventQueue:
+				if !ok { // compaction purge closed the queue
```
serathius: What happens after compaction purge? You need to reestablish the watch.

cache/cache.go
```diff
@@ -191,14 +235,27 @@ func (c *Cache) Close() {
 func serveWatchEvents(ctx context.Context, watchCtx *watchCtx) {
 	backoff := watchCtx.backoffStart
 	for {
+		var startRev int64
+		if kv, ok := watchCtx.cache.watcher.(clientv3.KV); ok {
```
serathius: This is not correct way to use interface. Please don't assert `clientv3.KV` interface here, but require it at the client input.

cache/cache.go
serathius, line 248: What is this? this is not how eventually consistent cache works. You don't convert state into diff.

```suggestion
```
What is this? this is not how eventually consistent cache works. You don't convert state into diff.

cache/cache.go
```diff
@@ -245,25 +303,66 @@ func readWatchChannel(
...
 
 func (c *Cache) validateWatch(key string, op clientv3.Op) (pred KeyPredicate, err error) {
-	if op.IsPrevKV() ||
-		op.IsFragment() ||
-		op.IsProgressNotify() ||
-		op.IsCreatedNotify() ||
-		op.IsFilterPut() ||
-		op.IsFilterDelete() {
-		return nil, ErrUnsupportedRequest
+	switch {
```
serathius: This should be moved to separate PR.

cache/config.go
```diff
@@ -29,6 +29,8 @@ type Config struct {
 	InitialBackoff time.Duration
 	// MaxBackoff caps the exponential back-off between successive upstream watch retries.
 	MaxBackoff time.Duration
+	// InitialLoadTimeout is the timeout applied to the first Get() used to bootstrap the cache.
+	InitialLoadTimeout time.Duration
```
serathius:
```suggestion
	// GetTimeout is the timeout applied to the Get() used to bootstrap the cache.
	GetTimeout time.Duration
```
Don't invent new terms.

cache/store.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (s *store) Get(startKey, endKey []byte) ([]*mvccpb.KeyValue, int64) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	if len(endKey) == 0 { // fast path: single key lookup
```
serathius: Don't use comments to separate code blocks, move blocks to separate functions

cache/store.go
serathius: ???

serathius:
> Maybe we should: (1) make the test wait for the cache to actually apply the PUT (2) change WaitReady to only close after the first real watch event so it truly reflects up-to-date.

No, best would be to wait until cache reaches certain revision.

serathius:
Please rebase

cache/cache.go
```diff
@@ -72,13 +74,32 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
+		store:       newStore(),
 		ready:       make(chan struct{}),
 		stop:        cancel,
 		internalCtx: internalCtx,
 	}
 
 	cache.demux = NewDemux(internalCtx, &cache.waitGroup, cfg.HistoryWindowSize, cfg.ResyncInterval)
 
+	storeW := newWatcher(cfg.HistoryWindowSize, nil) // nil predicate == match all
+	cache.demux.Register(storeW, 0)
```
serathius: You cannot register watch for storage on `0` revision because any state. You need to register it on revision returned by get plus one.

cache/cache.go
serathius: I don't understand the comment as the whole code is incorrectly written with regards to revision.

cache/cache.go
```diff
@@ -72,13 +74,32 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
...
+		defer cache.waitGroup.Done()
+		for {
+			select {
+			case <-cache.internalCtx.Done():
+				return
+			case evs, ok := <-storeW.eventQueue:
+				if !ok { // compaction purge closed the queue
+					return
+				}
+				_ = cache.store.apply(evs)
```
MadhavJivrajani: Any reason to ignore the error here?

cache/cache.go
```diff
@@ -162,6 +183,29 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	return responseChan
 }
 
+func (c *Cache) Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.GetResponse, error) {
+	select {
+	case <-c.ready:
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
```
MadhavJivrajani: Something to note here (and even in Watch): if the cache is ready but the Done channel is closed, because the parent context somehow is cancelled, then this can execute either branch in a random order. We don't want to end up in a situation where we skip the Done() channel being closed.

We could do something like:
```go
<- c.ready
select {
case <-ctx.Done():
    return nil, ctx.Err()
default:
}
```

We should try and reuse the `Ready()` function somewhere here since we have it.

cache/store.go
```diff
@@ -0,0 +1,98 @@
...
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+type store struct {
+	mu        sync.RWMutex
+	kvs       map[string]*mvccpb.KeyValue
+	latestRev int64
+}
+
+func (s *store) Get(startKey, endKey []byte) ([]*mvccpb.KeyValue, int64) {
```
MadhavJivrajani: nit: let's make all functions of store private.

cache/cache.go
MadhavJivrajani: Let's remove this part. The `Get()` should only be used to populate the store.
Maybe we can do that as part of the constructor itself. And then use the rev from Get to register the watch on the demux.

cache/store.go
```diff
@@ -0,0 +1,98 @@
...
+			return []*mvccpb.KeyValue{kv}, s.latestRev
+		}
+		return nil, s.latestRev
+	}
+
+	var out []*mvccpb.KeyValue
+	for _, kv := range s.kvs { // range / prefix scan
+		k := kv.Key
+		switch {
+		case len(endKey) == 1 && endKey[0] == 0: // prefix & fromKey
```
MadhavJivrajani: Can we modify the signature and pass in `op` and just check `op.IsWithPrefix()`?

cache/cache.go
serathius: We cannot wait for ready indefinitely, but you are right that this code might continue even context is closed.

cache/cache.go
apullo777: Sorry I mixed things up badly here. After the initial Get, we should instead update store and register the store watcher here with the returned rev+1 like below:

```
		var startRev int64
		if resp, err := watchCtx.cache.kv.Get(ctx, watchCtx.cache.prefix, clientv3.WithPrefix()); err == nil {
			watchCtx.cache.store.set(resp.Kvs, resp.Header.Revision)
			watchCtx.cache.demux.Register(watchCtx.cache.storeWatcher, resp.Header.Revision+1)
			watchCtx.onFirstResponse()
			startRev = resp.Header.Revision + 1
		}
```

Since in `New` we are also handling the store watcher, so at high level I am thinking maybe it is better keep a `storeWatcher *watcher` field in `Cache` struct that drives two loops: 
- one in `New` reads its event queue to do `store.apply(events)` (and after compaction errors rebuilds itself)
- the other (in `serveWatchEvents`) does the initial Get, registers that same `storeWatcher` with the demux

So that we can avoid doing that `store.apply` loop here within the `serveWatchEvents` and keep that in `New`, while both loops can handle the same `storeWatcher`

So in `New` we do `store.apply` loop like below:

```
	cache.storeWatcher = newWatcher(cfg.HistoryWindowSize, nil)
	cache.waitGroup.Add(1)
	go func(w *watcher) {
		defer cache.waitGroup.Done()
		for {
			select {
			case <-cache.internalCtx.Done():
				return
			case evs, ok := <-w.eventQueue:
				if !ok { // compaction purge closed the queue, re‑register watcher
					w = newWatcher(cache.cfg.HistoryWindowSize, nil)
					cache.storeWatcher = w
					continue
				}
				if err := cache.store.apply(evs); err != nil { 
				    // unreliable stale state, guarantee broken, panic...
			}
		}
	}(cache.storeWatcher)
```

But I am not very sure if this is the best way because now the compaction is handled across two separate go-routines.

cache/cache.go
apullo777: Do we modify the signature as Madhav suggested before to accept `clientv3.Client` like below (use pointer so that we do not need to modify integration tests too much, but still need to modify mocked watcher in unit tests)? 

```
func New(client *clientv3.Client, prefix string, opts ...Option) (*Cache, error) {
        ...
	cache := &Cache{
		prefix:      prefix,
		cfg:         cfg,
		watcher:     client.Watcher,
		kv:          client.KV,
```

cache/cache.go
serathius: > I am thinking maybe it is better keep a storeWatcher *watcher field in Cache struct that drives two loops:

Why?

serathius: I don't think there should be two loops.

cache/store.go
serathius: store is already private, there is no benefit of making all methods private. I like the idea of having capitalization to mark which methods should be called by other structs vs methods internal to store.

cache/cache.go
apullo777: Ok! So I should better move that  `store.apply `go-routine inside `serveWatchEvents` to keep only 1 retry loop.

apullo777:
Still trying to figure out how to write tests to cover scenarios where we preload keys before starting the cache. To reuse scenarios and avoid duplicate, I was thinking to change `testGet`’s signature to` func testGet(t *testing.T, kv clientv3.KV, reader Getter, skipCount int) `, write a helper `preloadEvents`, and modify our `TestCacheWithoutPrefixGet` like below. 

```
func TestCacheWithoutPrefixGet(t *testing.T) {
  cases := []struct {
    name    string
    preload int // how many events to write before cache starts
  }{
    {"empty",   0},
    {"partial", 4},
    {"complete",8},
  }

  for _, c := range cases {
      ...
      preloadEvents(t, client.KV, c.preload) // preload first N events before cache.New()
      ...
      testGet(t, client.KV, cache, c.preload)
```
But this approach will still duplicate event puts in `preloadEvents`. I will keep finding a better solution.

cache/cache.go
apullo777: Thanks for flagging! We do have a `Ready()` & `WaitReady()` in cache, and I was thinking perhaps we can modify `WaitReady` to reuse it:


```
func (c *Cache) WaitReady(ctx context.Context) error {
    select {
    case <-c.ready:
        return ctx.Err() // it used to return nil but now we let cancellation win
    case <-ctx.Done():
        return ctx.Err()
    }
}
```
and then we can use the modified `WaitReady` in `Get` (and `Watch` as well):

```
func (c *Cache) Get(ctx context.Context, …) (…) {
    if err := c.WaitReady(ctx); err != nil {
        return nil, err
    }
    ...
```
```
func (c *Cache) Watch(ctx context.Context, …) clientv3.WatchChan {
    if err := c.WaitReady(ctx); err != nil {
        emptyWatchChan := make(chan WatchResponse)
        close(emptyWatchChan)
        return emptyWatchChan
    }
    …
}
```

apullo777:
CI started failing `TestCacheWithoutPrefixWatch/Latest-revision_watcher_for_/foo` and several Get-based cases. 

I noticed that our `Watch` is still using the peek history then start at 0 if ring buffer is empty logic (in `demux.Register` as shown below) instead of the new get-and-watch approach, looks like it sometimes kicks off before any events arrive in the ring buffer. 

```
        latestRev := d.history.PeekLatest()
	// Special case: 0 means “newest”.
	if startingRev == 0 {
		if latestRev == 0 {
			d.activeWatchers[w] = 0
			return
		}
		startingRev = latestRev + 1
	}
```

Does this indicates that we also need to compute `startRev` from the latest snapshot revision + 1,  so that `WithRev(0)` path truly means “from now on”. Maybe to solve this by having an accessor like  `store.LatestRev()` and do in `Watch` like below?

```
    op := clientv3.OpWatch(key, opts...)
    startRev := op.Rev()
    
    if startRev == 0 {
        startRev = c.store.LatestRev() + 1
    }
```

cache/cache.go
serathius: Looks good to me

cache/cache.go
```diff
@@ -188,17 +216,44 @@ func (c *Cache) Close() {
 	c.waitGroup.Wait()
 }
 
-func serveWatchEvents(ctx context.Context, watchCtx *watchCtx) {
+func getAndWatchRetry(ctx context.Context, watchCtx *watchCtx) {
 	backoff := watchCtx.backoffStart
 	for {
+		var startRev int64
+		if resp, err := watchCtx.cache.kv.Get(ctx, watchCtx.cache.prefix, clientv3.WithPrefix()); err == nil {
```
serathius: You cannot move forward if err is not empty, that was pointed out in previews comments. Please don't close comments without confirming they are addressed. Please add retry logic.

cache/cache.go
```diff
@@ -188,17 +216,44 @@ func (c *Cache) Close() {
...
+			watchCtx.cache.store.Set(resp.Kvs, resp.Header.Revision)
```
serathius: This method should not be `Set`. `Set` implies single key. Please use name like `Reset`.

cache/cache.go
```diff
@@ -188,17 +216,44 @@ func (c *Cache) Close() {
...
+
+			storeW := newWatcher(watchCtx.cache.cfg.HistoryWindowSize, nil)
+
+			watchCtx.cache.waitGroup.Add(1)
+			go func() {
```
serathius: Again move to separate function. This code is too long to be embeded.

cache/cache.go
```diff
@@ -188,17 +216,44 @@ func (c *Cache) Close() {
...
+				defer watchCtx.cache.waitGroup.Done()
+				defer watchCtx.cache.demux.Unregister(storeW)
+				for {
+					select {
+					case events, ok := <-storeW.eventQueue:
+						if !ok { // compaction, will rebuild watcher at next retry iteration
```
serathius: I don't see any retry for this

cache/cache.go
```diff
@@ -188,17 +216,44 @@ func (c *Cache) Close() {
...
+							return
+						}
+						if err := watchCtx.cache.store.Apply(events); err != nil {
+							// seeing rev older than last-observed -> breaks the “monotonic revision” contract
+							panic(err)
```
serathius: Don't panic, just reset.

cache/cache.go
```diff
@@ -188,17 +216,44 @@ func (c *Cache) Close() {
...
+						}
+					case <-ctx.Done():
+						return
+					}
+				}
+			}()
+			watchCtx.cache.demux.Register(storeW, resp.Header.Revision+1)
+			watchCtx.onFirstResponse()
```
serathius: Don't know what `onFirstResponse` is doing here.

cache/store.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func newStore() *store {
+	return &store{kvs: make(map[string]*mvccpb.KeyValue)}
+}
+
+func (s *store) Get(startKey, endKey []byte) ([]*mvccpb.KeyValue, int64) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
```
serathius, line 42: store should reject requests if it's not initialized.

cache/cache.go
```diff
@@ -162,6 +167,29 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	return responseChan
 }
 
+func (c *Cache) Get(ctx context.Context, key string, opts ...clientv3.OpOption) (*clientv3.GetResponse, error) {
+	select {
+	case <-c.ready:
```
serathius: This should wait until store is initialized.

cache/cache.go
serathius: This should be renamed to `markReady`

cache/store.go
```diff
@@ -0,0 +1,134 @@
...
+	if len(endKey) == 0 {
+		return s.getSingle(startKey), s.latestRev
+	}
+
+	var out []*mvccpb.KeyValue
+	if isPrefixScan(endKey) {
+		out = s.scanPrefix(startKey)
+	} else {
+		out = s.scanRange(startKey, endKey)
+	}
```
serathius:
```suggestion
	var out []*mvccpb.KeyValue
	switch {
	case len(endKey) == 0:
		out = s.getSingle(startKey)
	case isPrefixScan(endKey):
		out = s.scanPrefix(startKey)
	default:
		out = s.scanRange(startKey, endKey)
	}
```

cache/store.go
```diff
@@ -0,0 +1,134 @@
...
+		return bytes.Compare(out[i].Key, out[j].Key) < 0 // default: lexicographical, ascending‐by‐key sort
+	})
+	return out, s.latestRev
+}
+
+func (s *store) Set(kvs []*mvccpb.KeyValue, rev int64) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	s.kvs = make(map[string]*mvccpb.KeyValue, len(kvs))
```
serathius, line 62: Reset before adding new keys.

cache/cache.go
apullo777: Sorry I missed the issue. Yes. We cannot proceed without successfully obtaining the Get result, and should do retry with backoff like below:
               
```
        backoff := watchCtx.backoffStart
	for {
                var startRev int64
		resp, err := watchCtx.cache.kv.Get(ctx, watchCtx.cache.prefix, clientv3.WithPrefix())
		if err != nil {
			select {
			case <-ctx.Done():
				return
			case <-time.After(backoff):
			}
			if backoff < watchCtx.backoffMax {
				backoff *= 2
			}
			continue
		}
		backoff = watchCtx.backoffStart // successful snapshot: reset back‑off
		watchCtx.cache.store.ResetSnapshot(resp.Kvs, resp.Header.Revision)
```

Since we have similar code block at the end, maybe we can move the shared block into a helper?

```
func backoffRetry(ctx context.Context, backoff *time.Duration, max time.Duration) bool {
    select {
    case <-ctx.Done():
        return false
    case <-time.After(*backoff):
    }
    if *backoff < max {
        *backoff *= 2
    }
    return true
}
```

cache/cache.go
apullo777: Ok. I will rename `store.Set` as `store.ResetSnapshot`, and to avoid confusion rename the original `store.Reset()` as `store.Purge`.

cache/store.go
apullo777: Ok! In rejecting should we return an explicit `ErrNotInitialized` and change `store.Get`'s signature to (..., error)?

cache/cache.go
apullo777: Just want to make sure if my understanding is correct: 

1.  When `readWatchChannel` sees a compaction error it resets the cache’s ready signal, purges the demux and store (which in turn stops the store watcher)
2. Reacting to the closed channel, we return from this `store.Apply` go-routine, trigger the two defers (`demux.Unregister` + `waitGroup.Done`)
3.  Then the outer `getAndWatchRetry` loop back‑offs, retry a new iteration of `Get` + populates, registering a new store watcher, and `Watch` from the new revision

So I was thinking it is `readWatchChannel` which will do the purge, and this `store.Apply` go-routine is reacting to it, and then `getAndWatchRetry` will do the retry. 

Do you have something else in mind? Appreciate any pointer so I can find the issue more quickly!

cache/cache.go
apullo777: Now I see that I used to `markReady()` immediately after `demux.Register(storeW, rev+1)`, which was too early. This `markReady()`  should be removed.

I am thinking to move that `markReady()` to the `store.Apply()` go-routine. So that we wait until we actually see traffic from the etcd watch (either in `readWatchChannel` or in `store.Apply()` go-routine)

```
	go func() {
		defer cache.waitGroup.Done()
		defer cache.demux.Unregister(w)

		first := true
		for {
			select {
			case events, ok := <-w.eventQueue:
				...
				if err := cache.store.Apply(events); err != nil {
					...
				}
				if first {
					first = false
					markReady()
				}
```

cache/cache.go
serathius: Keep it separate

cache/cache.go
serathius: Or `Restore`

cache/cache.go
serathius: I think it make sense, the code is too complicated. Code should be simple:
```
func (c *cache) getWatchLoop() {
  for {
    c.getWatch()
  }
}
func (c *cache) getWatch() {
  getResp, err := c.get()
  if err != nil {
    return
  }
  c.watch(getResp.Revision+1)
  return
}
func (c *cache) get() (resp, error) {
  resp, err := c.client.get()
  if err != nil {
    return nil, err
  } 
  c.store.Restore(resp)
  return resp, nil
} 

func (c *cache) watch(revision) {
  for {
    watch := c.client.watch(revision)
    for resp := range watch {
      revision = lastRevision(resp)
      c.store.Apply(resp.Events)
    } 
  }
} 
```

apullo777: Thanks! I totally agree. It does feel more complex than it needs to be. I will try to take some time to think how I can simplify the logic.

cache/store.go
serathius: Yes, if you can reset the store and make it empty, you cannot casually serve empty response.

cache/cache.go
apullo777: With this new retry logic, I noticed that we will have double close issue in unit tests. I was thinking to modify the Watch as below to adapt:

```
func (m *mockWatcher) Watch(_ context.Context, _ string, _ ...clientv3.OpOption) clientv3.WatchChan {
	select {
	case <-m.registered:
	default:
		close(m.registered)
	}
	return m.responses
}
```

apullo777:
> > Maybe we should: (1) make the test wait for the cache to actually apply the PUT (2) change WaitReady to only close after the first real watch event so it truly reflects up-to-date.
> 
> No, best would be to wait until cache reaches certain revision.

To fix this flake, I am thinking to expose a helper named `WaitForRevision` so that our test will wait until the PUT is done:

```
func (c *Cache) WaitForRevision(ctx context.Context, rev int64) error {
  for {
    if c.store.LatestRev() >= rev {
      return nil
    }
    select {
    case <-time.After(10 * time.Millisecond):
    case <-ctx.Done():
      return ctx.Err()
    }
  }
}
```
in store.go:
```
func (s *store) LatestRev() int64 {
  s.mu.RLock()
  defer s.mu.RUnlock()
  return s.latestRev
}

```
and use the helper in GET related cache tests:

```
putResp, _ := client.Put(ctx, testKey, "val")
if err := c.WaitForRevision(ctx, putResp.Header.Revision); err != nil {
    t.Fatalf("cache never caught up to rev %d: %v", putResp.Header.Revision, err)
}
```

cache/cache.go
```diff
@@ -188,60 +197,76 @@ func (c *Cache) Close() {
 	c.waitGroup.Wait()
 }
 
-func serveWatchEvents(ctx context.Context, watchCtx *watchCtx) {
-	backoff := watchCtx.backoffStart
+func (c *Cache) getWatchLoop(ctx context.Context) {
```
apullo777, line 214: Wondering if we need to do backoff logic here like below. If so, would also need to modify `getWatch`'s & `watch`'s  signatures to let them surface error so that we backoff for certain issues like network issues and failed `get`.

```
func (c *Cache) getWatchLoop(ctx context.Context) {
	cfg := defaultConfig()
	backoff := cfg.InitialBackoff
	for {
		select {
		case <-ctx.Done():
			return
		default:
			if err := c.getWatch(ctx); err == nil {
				backoff = cfg.InitialBackoff
			} else { // failure → back-off
				select {
				case <-ctx.Done():
					return
				case <-time.After(backoff):
				}
				if backoff < cfg.MaxBackoff {
					backoff *= 2
				}
			}
		}
	}
}
```

cache/cache.go
```diff
@@ -99,9 +86,7 @@ func New(watcher clientv3.Watcher, prefix string, opts ...Option) (*Cache, error
 // Watch registers a cache-backed watcher for a given key or prefix.
 // It returns a WatchChan that streams WatchResponses containing events.
 func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption) clientv3.WatchChan {
-	select {
-	case <-c.ready:
-	case <-ctx.Done():
+	if err := c.WaitReady(ctx); err != nil {
```
serathius, line 89: Please move to separate PR

cache/cache.go
```diff
@@ -188,60 +197,76 @@ func (c *Cache) Close() {
...
-		}
-		if oldestRev := watchCtx.cache.demux.PeekOldest(); oldestRev != 0 {
-			opts = append(opts,
-				clientv3.WithRev(oldestRev+1))
-		}
-		watchCh := watchCtx.cache.watcher.Watch(ctx, watchCtx.cache.prefix, opts...)
+		)
+
+		for resp := range watchCh {
+			readyOnce.Do(func() { close(c.ready) })
```
serathius, line 241: Create a dedicated function to mark ready.

cache/cache.go
```diff
@@ -188,60 +197,76 @@ func (c *Cache) Close() {
...
+			if err := resp.Err(); err != nil { // compaction
+				c.demux.Purge()
+				c.resetReady()
+				c.store.Reset()
+				return
+			}
 
-		if err := readWatchChannel(watchCh, watchCtx.cache, watchCtx.cache.demux, watchCtx.onFirstResponse); err == nil {
-			return
+			if err := c.store.Apply(resp.Events); err != nil { // statle state
```
serathius: Stop using comments! They don't provide any information and  will get outdated. Fact that error in store.Apply causes reset of cache makes sense, `// stale cache` doesn't. This leaks internal information about when error can be returned that will get outdated.

cache/cache.go
serathius: Sounds good

cache/cache_test.go
```diff
@@ -188,7 +189,14 @@ func TestCacheWatchAtomicOrderedDelivery(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			mw := newMockWatcher(16)
-			cache, err := New(mw, "")
+			fakeClient := &clientv3.Client{
```
serathius, line 192: Note, we might also want a unit test that injects a compaction

serathius:
> To fix this flake, I am thinking to expose a helper named WaitForRevision so that our test will wait until the PUT is done:

Ok

tests/integration/cache_test.go
```diff
@@ -486,6 +486,301 @@ func TestCacheWithPrefix(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("cache.New: %v", err)
+	}
+	defer c.Close()
+	if err := c.WaitReady(ctx); err != nil {
+		t.Fatalf("cache.WaitReady: %v", err)
+	}
+
+	const testKey = "/foo/a"
+	putResp, _ := client.Put(ctx, testKey, "val")
```
serathius: Please check error

cache/cache.go
```diff
@@ -162,6 +148,30 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
+	op := clientv3.OpGet(key, opts...)
+
+	if _, err := c.validateGet(key, op); err != nil {
+		return nil, err
+	}
+
+	startKey := []byte(key)
+	endKey := op.RangeBytes()
+	kvs, rev, err := c.store.Get(startKey, endKey)
+	if err != nil { // ErrNotInitialized
```
```suggestion
	if err != nil {
```

cache/store.go
```diff
@@ -0,0 +1,145 @@
...
+	"bytes"
+	"fmt"
+	"sort"
+	"sync"
+
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+var ErrNotInitialized = fmt.Errorf("cache: store not initialized")
```
serathius, line 27: Name of error is inaccurate. You return when the store is not ready, which by default will be when it's not initialized. Fixing this would require a rewrite of ready status, but let's not focus on it now.

cache/cache.go
```diff
@@ -176,72 +186,114 @@ func (c *Cache) Ready() bool {
...
-	backoff := watchCtx.backoffStart
+func (c *Cache) getWatchLoop(ctx context.Context) {
+	cfg := defaultConfig()
+	backoff := cfg.InitialBackoff
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			if err := c.getWatch(ctx); err == nil {
```
serathius: I think you always want to backoff as lack of error still means the watch exited which is bad.

cache/cache.go
```diff
@@ -176,72 +186,114 @@ func (c *Cache) Ready() bool {
...
+				backoff = cfg.InitialBackoff
+			} else {
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(backoff):
+				}
+				if backoff < cfg.MaxBackoff {
+					backoff *= 2
```
serathius: There is no logic for reseting the backoff on success, so I'm not sure if we should do exponential. Over time you will always get some errors (like 1 error per 24h) and the backoff will grow even when errors are very rare. Increasing backoff is done to avoid too frequent retries, not to punish rare retries.

serathius: In best case you would reset backoff after watch is successfully created.

apullo777: Thanks for the feedback! You are right under our current `getWatch` it won’t return nil during normal operation. I was thinking two options:

1. Split `getWatch` into two phases: connect & stream (like `startWatch` and `runWatch`) and have it return nil when we successfully create the upstream watch
2. pass a `resetBackoff` callback from the outer `getWatchLoop` down into `watch`, so that we reset our delay back to InitialBackoff on the first successful created notification like below:

```
    resetBackoff := func() {
        backoff = cfg.InitialBackoff
    }
```

modify `getWatch`:
```
func (c *Cache) getWatch(ctx context.Context, resetBackoff func()) error {
    resp, err := c.kv.Get(ctx, c.prefix, clientv3.WithPrefix())
    if err != nil {
        return err
    }
    return c.watch(ctx, resp.Header.Revision+1, resetBackoff)
}
```

and in watch loop do this:

```
          for resp := range watchCh {
			readyOnce.Do(func() {
				close(c.ready)
				resetBackoff()
			})
```

cache/cache.go
apullo777: Maybe I should fix it like this:

```
func (c *Cache) getWatchLoop(ctx context.Context) {
     cfg := defaultConfig()
     backoff := cfg.InitialBackoff

     for {
         select {
         case <-ctx.Done():
             return
         default:
            err := c.getWatch(ctx)

            select {
            case <-ctx.Done():
                return
            case <-time.After(backoff):
            }

            if err != nil {
                if backoff < cfg.MaxBackoff {
                    backoff *= 2
                }
            } else {
                backoff = cfg.InitialBackoff
            }
         }
     }
 }
```

cache/cache.go
```diff
@@ -176,72 +186,108 @@ func (c *Cache) Ready() bool {
...
+			clientv3.WithProgressNotify(),
+			clientv3.WithCreatedNotify(),
+		)
+
+		for resp := range watchCh {
+			readyOnce.Do(func() { close(c.ready) })
+			if err := resp.Err(); err != nil {
+				c.demux.Purge()
+				c.resetReady()
+				c.store.Reset()
```
serathius, line 266: Why in this order? I would expect `c.resetReady()` should be first.

cache/cache.go
```diff
@@ -176,72 +186,108 @@ func (c *Cache) Ready() bool {
...
+				return err
+			}
+
+			if err := c.store.Apply(resp.Events); err != nil {
+				c.demux.Purge()
```
serathius, line 271: ditto

serathius (approved):

cache/cache.go
apullo777: My initial thinking was that readiness depended on the store being initialized first, so should make sure it was rest before we reset the store. But you are absolutely right that we should always call resetReady() before any purge/reset. I will update the order in a follow-up PR (or include it in the readiness-fix PR).

Issue #20340, opened 2025-07-15
ETCD fails to start, with panic on 'failed to nodeToMember'


shanduur:
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [x] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [x] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [x] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

During preparation for migration of default ETCD version in Talos, we prepared a patch with new version, to run against our test suite. The test suite _sometimes_ fail during the full cluster reboot, when one of the ETCD becomes corrupt. It fails with a panic (`failed to nodeToMember`, full stack trace and logs below).

> [!WARNING]
> Compressed databases from all ETCD nodes: [db.tgz](https://github.com/user-attachments/files/21234529/db.tgz)

### What did you expect to happen?

No panic

### How can we reproduce it (as minimally and precisely as possible)?

This guide assumes Linux host, with qemu, docker and make. 

```shell
git clone https://github.com/shanduur/talos.git
cd talos && git checkout test/etcd-3.6.x

# prepare environment
docker buildx create --driver docker-container  --driver-opt network=host --name local1 --buildkitd-flags '--allow-insecure-entitlement security.insecure' --use
docker run -d -p 5005:5000 \
    --restart always \
    --name local registry:2

# build necessary components
make installer-base IMAGE_REGISTRY=127.0.0.1:5005 PUSH=true
make imager IMAGE_REGISTRY=127.0.0.1:5005 PUSH=true
make installer IMAGE_REGISTRY=127.0.0.1:5005
# Record installer image tag for the installation step
make initramfs kernel talosctl

# start talos cluster
bash hack/start-registry-proxies.sh
sudo --preserve-env=HOME _out/talosctl-linux-amd64 cluster create \
    --provisioner=qemu \
    --cidr=172.20.0.0/24 \
    --registry-mirror docker.io=http://172.20.0.1:5000 \
    --registry-mirror registry.k8s.io=http://172.20.0.1:5001  \
    --registry-mirror gcr.io=http://172.20.0.1:5003 \
[...]

PR #20387, opened 2025-07-22
[release-3.6] Check promotion against v2store when updating v2store

ahrtr:
fix https://github.com/etcd-io/etcd/issues/20340

Backport https://github.com/etcd-io/etcd/pull/20391 to 3.6.

### DIFF
--- a/server/etcdserver/api/membership/cluster.go
+++ b/server/etcdserver/api/membership/cluster.go
@@ -522,8 +522,9 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	defer c.Unlock()
 
 	if c.v2store != nil {
-		if _, ok := c.members[id]; ok {
-			m := *(c.members[id])
+		membersMap, _ := membersFromStore(c.lg, c.v2store)
+		if _, ok := membersMap[id]; ok {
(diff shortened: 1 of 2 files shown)


server/etcdserver/api/membership/cluster.go
```diff
@@ -522,8 +522,9 @@ func (c *RaftCluster) PromoteMember(id types.ID, shouldApplyV3 ShouldApplyV3) {
 	defer c.Unlock()
 
 	if c.v2store != nil {
-		if _, ok := c.members[id]; ok {
-			m := *(c.members[id])
+		membersMap, _ := membersFromStore(c.lg, c.v2store)
```
fuweid, line 525:
```
Start single node cluster
add learner etcd-2, and promote it;
add learner etcd-3, and promote it;
remove etcd-2;
add learner etcd-2 again, and promote it;
reboot any member will trigger the issues:
Error: membership: too many learner members in cluster
Panicking
```

I try to write e2e code to reproduce that issue without your patch. I can't reproduce it right now.
sometimes, if we join member back with same name fast, it will show member is removed.
Just share my testing result.

ahrtr: Try the e2e test in https://github.com/etcd-io/etcd/pull/20391

PR #20388, opened 2025-07-23
etcdctl: refactor `printer_json` files

kstrifonoff:
This PR includes refactoring changes only. No new functionality is introduced.

- Updated printer_json.go to follow consistent import naming conventions.
- Reduced code duplication in printer_json_test.go by extracting common setup logic into helper functions.

These changes improve code readability and maintainability without altering behavior.

### DIFF
--- a/etcdctl/ctlv3/command/printer_json_test.go
+++ b/etcdctl/ctlv3/command/printer_json_test.go
@@ -63,10 +63,55 @@ func assertHexFieldEqual(t *testing.T, obj map[string]any, key string, want stri
 	assert.Equalf(t, want, str, "unexpected value for hex field %q", key)
 }
 
+func assertHeader(t *testing.T, testGroup *testScenario, tt *testCase, got map[string]any) {
+	rawHeader, ok := got[keyHeader]
+	require.Truef(t, ok, "output does not contain %q field: %v", keyHeader, got)
+	header, ok := rawHeader.(map[string]any)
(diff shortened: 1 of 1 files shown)


kstrifonoff:
/cc @ahrtr 
Could you please review this PR? It's a small update intended to simplify upcoming code changes related to #19262

etcdctl/ctlv3/command/printer_json.go
```diff
@@ -21,7 +21,7 @@ import (
 	"os"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
-	clientv3 "go.etcd.io/etcd/client/v3"
+	v3 "go.etcd.io/etcd/client/v3"
```
ahrtr: Personally I think `clientv3` is accurate & better. (Actually we don't need to specify package name here, because it's `clientv3` by default.

https://github.com/etcd-io/etcd/blob/df6b790d7c5bac171bd18f605de352126e01518a/client/v3/compare.go#L15

So I suggest to keep this file unchanged.

kstrifonoff: ok, it makes sense. reverted this
however, I did notice a few other places where the `v3` naming is still used, so it seems there is some inconsistency. that said, I agree that the majority of files use the `clientv3` naming instead.

ahrtr (approved):
LGTM & thx

PR #20389, opened 2025-07-23
Add OpenTelemetry Spans to Put, DeleteRange, Compact

AwesomePatrol:

Attributes added:
1. `Put`: method not used by Kubernetes outside integration tests
   * `key`
2. `DeleteRange`: method not used by Kubernetes outside integration tests
   * `range_begin`
   * `range_end`
3. `Compact`: method not part of k8s-etcd interface, but added for completeness
   * `rev`
4. `Range`:
    * `keys_only` - used by Kubernetes 1.34 since https://github.com/kubernetes/kubernetes/pull/132355 (depending on feature gate)

Events added:
1. `Send raft proposal`
2. `Receive raft result`

### DIFF
--- a/server/etcdserver/v3_server.go
+++ b/server/etcdserver/v3_server.go
@@ -152,6 +152,12 @@ func (s *EtcdServer) Range(ctx context.Context, r *pb.RangeRequest) (*pb.RangeRe
 }
 
 func (s *EtcdServer) Put(ctx context.Context, r *pb.PutRequest) (*pb.PutResponse, error) {
+	var span trace.Span
+	ctx, span = traceutil.Tracer.Start(ctx, "put", trace.WithAttributes(
+		attribute.String("key", string(r.GetKey())),
+	))
+	defer span.End()
+
 	ctx = context.WithValue(ctx, traceutil.StartTimeKey{}, time.Now())
 	resp, err := s.raftRequest(ctx, pb.InternalRaftRequest{Put: r})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


serathius (approved):

server/etcdserver/v3_server.go
```diff
@@ -721,6 +742,7 @@ func (s *EtcdServer) RoleDelete(ctx context.Context, r *pb.AuthRoleDeleteRequest
 func (s *EtcdServer) raftRequestOnce(ctx context.Context, r pb.InternalRaftRequest) (proto.Message, error) {
 	result, err := s.processInternalRaftRequestOnce(ctx, r)
 	if err != nil {
+		trace.SpanFromContext(ctx).RecordError(err)
```
fuweid, line 745: do we need to trace error if `result.Err` is not empty?

AwesomePatrol: We can, but don't need to. The traceutil tracing doesn't seem to care about errors (only how long the operations take). I see three options:
1. Drop `RecordError` from this PR and add it after we migrate fully to OTEL
2. Add `RecordError` to more places where it could make sense (like `if result.Err != nil`)
3. Keep this one example as is and see how useful it is

Which one do you prefer?

fuweid (approved):

server/etcdserver/v3_server.go
fuweid: vote for #3

PR #20391, opened 2025-07-23
Check promotion against v2store when updating v2store

ahrtr:
Fix https://github.com/etcd-io/etcd/issues/20340

Refer to https://github.com/etcd-io/etcd/issues/20340#issuecomment-3105037914

### DIFF
--- a/tests/e2e/member_no_proxy_test.go
+++ b/tests/e2e/member_no_proxy_test.go
@@ -1,0 +1,61 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !cluster_proxy
+
+package e2e
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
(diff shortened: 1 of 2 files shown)


ivanvc (approved):
LGTM. Verified that the panic doesn't exist after testing with the script from https://github.com/etcd-io/etcd/issues/20340#issuecomment-3105037914. Thanks, @ahrtr.

tests/e2e/member_no_proxy_test.go
```diff
@@ -0,0 +1,61 @@
...
+	epc, members := mustCreateNewClusterByPromotingMembers(t, e2e.CurrentVersion, 3)
+	defer func() {
+		require.NoError(t, epc.Close())
+	}()
+
+	t.Logf("Removing the second member (%x)", members[0].ID)
+	etcdctl := epc.Procs[0].Etcdctl()
+	testutils.ExecuteWithTimeout(t, 10*time.Second, func() {
+		for {
+			_, merr := etcdctl.MemberRemove(ctx, members[0].ID)
```
fuweid, line 45: OK. it makes sense. If we add it back, we won't reproduce this issue.

tests/e2e/member_no_proxy_test.go
```diff
@@ -0,0 +1,61 @@
...
+					time.Sleep(1 * time.Second)
+					continue
+				}
+				t.Fatalf("failed to remove member: %s", merr.Error())
+			}
+			break
+		}
+	})
+
+	epc.Procs = append(epc.Procs[:1], epc.Procs[2:]...)
```
fuweid, line 57: It looks like we don't need to remove second proc.

fuweid (approved):
LGTM

tests/e2e/member_no_proxy_test.go
ahrtr: We can also reproduce the issue even we add it back. Note that when we add it back, it's a different member (with different member ID).

I just wanted to keep the test as simple as possible, so I didn't add it back.

tests/e2e/member_no_proxy_test.go
ahrtr: When we remove the member, it will automatically stop, so the final cluster cleanup might fail. Not a big problem though.

PR #20399, opened 2025-07-24
cache: rename validateWatchRange as validateRequestRange

apullo777:

validateWatchRange is renamed as validateRequestRange so that it can be reused by other function in future like validateGet
@serathius @MadhavJivrajani

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -252,13 +252,13 @@ func (c *Cache) validateWatch(key string, op clientv3.Op) (pred KeyPredicate, er
 	startKey := []byte(key)
 	endKey := op.RangeBytes() // nil = single key, {0}=FromKey, else explicit range
 
-	if err := c.validateWatchRange(startKey, endKey); err != nil {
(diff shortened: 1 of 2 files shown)


cache/cache.go
```diff
@@ -252,13 +252,13 @@ func (c *Cache) validateWatch(key string, op clientv3.Op) (pred KeyPredicate, er
...
 
-	if err := c.validateWatchRange(startKey, endKey); err != nil {
+	if err := c.validateRequestRange(startKey, endKey); err != nil {
 		return nil, err
 	}
 	return KeyPredForRange(startKey, endKey), nil
 }
 
-func (c *Cache) validateWatchRange(startKey, endKey []byte) error {
+func (c *Cache) validateRequestRange(startKey, endKey []byte) error {
```
```suggestion
func (c *Cache) validateRange(startKey, endKey []byte) error {
```

serathius: I don't see `Request` in the arguments.

PR #20404, opened 2025-07-26
robustness: update README.md

yagikota:
Fix invalid URL


joshjms:
Hi, thanks for your PR. Change seems valid to me. Can we exclude changes to lines other than 256 from the commit since I think they are unrelated to the commit. Thanks!

Only keep this:

```
This is consistent with the root cause of [#15271] where the member reconnecting to cluster will resend revisions.
```

yagikota:
@joshjms 

>Hi, thanks for your PR. Change seems valid to me. Can we exclude changes to lines other than 256 from the commit since I think they are unrelated to the commit. Thanks!

Updated

joshjms:
Looks good @yagikota , please squash the commits, thanks! That should also fix the DCO issue (since your revert commit isn't signed).

yagikota:
@joshjms 

Updated.

serathius (approved):

PR #20405, opened 2025-07-26
Fix Antithesis makefile if statement

serathius:
/cc @fuweid


tests/antithesis/server/Dockerfile
```diff
@@ -10,10 +10,7 @@ RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etc
 # replace verify with antithesis
 WORKDIR /etcd/client/pkg/verify
 COPY inject/verify.patch verify.patch
-RUN if [ "${REF}" == "main" ]; then \
```
serathius, line 13: '==' is a bash thing, while we only have access to /bin/sh here.

fuweid (approved):

PR #20406, opened 2025-07-26
Antithesis build

serathius:
/cc @fuweid @nwnt


serathius:
Ups not docker daemon available.

Makefile
```diff
@@ -15,6 +15,7 @@ build-all:
 	@for platform in $(PLATFORMS); do \
 		$(MAKE) build-$${platform}; \
 	done
+	$(MAKE) -C tests/antithesis antithesis-build-etcd-image-local antithesis-build-client-image antithesis-build-config-image
```
ahrtr, line 18: I suggest that all Antithesis-related code, config, or scripts be scoped strictly within `tests/antithesis/`. Since Antithesis is a commercial platform and this integration is optional, it's important we keep the core repo clean and avoid coupling external services into the main codebase.

Adding workflows is fine,
https://github.com/etcd-io/etcd/blob/d02a151e08dd00155598f15c5e165a1d63dcce89/.github/workflows/antithesis-test.yml#L68-L71

serathius: Sounds good

PR #20419, opened 2025-07-29
cache: refactor validateWatch to use a switch for unsupported watch ops

apullo777:

@serathius @MadhavJivrajani


serathius (approved):
We should not spend a lot of time on this code as we will just implement the methods in the future.

PR #20424, opened 2025-07-30
etcdserver: add validation of remoteServer

ProjectMutilation:
The remoteServer was used unsafely after nil check
Found by PostgresPro with the Svace Static Analyzer


ProjectMutilation:
@ahrtr hi, can you give me any feedback please?

PR #20428, opened 2025-07-31
Refactor robustness WAL reading logic to merge unparsed entries

serathius:
In the future merging for unparsed raft entries will allow us to glue raft histories using raft index.

/cc @fuweid @siyuanfoundation @nwnt @henrybear327 @joshjms


tests/robustness/report/wal.go
```diff
@@ -109,23 +125,23 @@ func persistedRequests(lg *zap.Logger, dataDirs []string, reader persistedReques
 		}
 	}
 	// Select longest history that has votes from quorum.
-	longestHistory := []model.EtcdRequest{}
-	quorum := len(dataDirs)/2 + 1
+	longestHistory := []raftpb.Entry{}
+	quorum := len(memberEntries)/2 + 1
 	foundQuorum := false
-	for i := 0; i < len(memberRequestHistories); i++ {
+	for i := 0; i < len(memberEntries); i++ {
```
nwnt, line 128: nit
```suggestion
	for i := range len(memberEntries) {
```

serathius: This would be unrelated change that makes PR harder to read.

serathius:
Hmm, not sure how this change could increase runtime of robustness tests.

serathius:
Ok, confirmed that this change increases robustness time execution by 60%.

serathius:
Ok, found the culprit. `cmp.Diff` took 200 times longer. Don't know why, but after replacing it with `reflect.DeepEqual` the runtime dropped back.

fuweid (approved):
LGTM

PR #20432, opened 2025-08-02
Inject asserts in place of gofail

### LINKED ISSUE #20283: [Anithesis] Inject assertions in place etcd verify.

nwnt:
Fix #20283 

Note that this doesn't handle the gofail with control label and only replaces `// gofail: var VAR_NAME struct{}`.  

@serathius

### DIFF
--- a/tests/antithesis/server/Dockerfile
+++ b/tests/antithesis/server/Dockerfile
@@ -7,6 +7,13 @@ FROM golang:$GO_VERSION
 ARG REF=main
 RUN git clone --depth=1 https://github.com/etcd-io/etcd.git --branch=${REF} /etcd
 
+# inject assertions in place of gofail
+WORKDIR /etcd/server
+RUN go install golang.org/x/tools/cmd/goimports@latest
+RUN go get github.com/antithesishq/antithesis-sdk-go@v0.4.4
+RUN for file in $(grep -rl '// gofail'); do sed -i 's|\/\/ gofail.*var \([[:alnum:]]*\) .*|assert\.Reachable("\1", nil)|' $file; goimports -w $file; done
+RUN go mod tidy
+
 # replace verify with antithesis
 WORKDIR /etcd/client/pkg/verify
 COPY inject/verify.patch verify.patch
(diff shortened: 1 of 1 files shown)


tests/antithesis/server/Dockerfile
```diff
@@ -19,7 +26,7 @@ WORKDIR /etcd
 RUN go mod download
 
 # install instrumentor
-RUN go get github.com/antithesishq/antithesis-sdk-go@a802e8810442e01d16b3e9df77d7ce3875e36e55 # v0.4.3
+RUN go get github.com/antithesishq/antithesis-sdk-go@v0.4.4
 RUN go install github.com/antithesishq/antithesis-sdk-go/tools/antithesis-go-instrumentor@a802e8810442e01d16b3e9df77d7ce3875e36e55 # v0.4.3
```
serathius, line 30: Should this version match?

serathius (approved):

serathius:
Awesome work!

serathius:
Results from Antithesis confirm that this works as intended. We hit all assertions in 15 minute run which is much better than I anticipated.

image


https://linuxfoundation.antithesis.com/report/w4PCj2bCrBErCMKdT21PaMOTw4rCqcKsSFg/qQMdfUA6Q3fplhmYfK38yY7HkB0j4D3mrTqTALAvF_U.html?auth=v2.public.eyJuYmYiOiIyMDI1LTA4LTA1VDA3OjA0OjI4LjkyMDEyNzUxNFoiLCJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoicVFNZGZVQTZRM2ZwbGhtWWZLMzh5WTdIa0IwajREM21yVHFUQUxBdkZfVS5odG1sIiwicmVwb3J0X2lkIjoidzRQQ2oyYkNyQkVyQ01LZFQyMVBhTU9UdzRyQ3FjS3NTRmcifX19uWKTuHDBBeJUjq0lYrrpI8oAyGp4eRb4WyIw1mC3HQhKeSxH2FAed6RkQXuwXeS-kTmD60_oSDIeMZZL2FOmAw

PR #20433, opened 2025-08-03
Refactor-e2e-txn-test

yagikota:
## What?
* Refactor `AddTxnResponse` function from a public package-level function to a private method `addTxnResponse` on the EtcdctlV3 struct
* Remove the corresponding unit test file `e2e_test.go` that was testing the now-private function

## Why?
- Improved encapsulation: Converting the public function to a private method better encapsulates the transaction response handling logic within the EtcdctlV3 struct, following object-oriented design principles
- Consistency with existing codebase: The etcd framework doesn't typically include unit tests for internal e2e framework helper functions, so removing the test aligns with the existing testing approach and architectural patterns


tests/framework/e2e/e2e_test.go
yagikota, line 1: Do you think we need to keep this test file as it is?
If so we should change the file name from `e2e_test.go` to `etcdctl_test.go` because the target function(`addTxnResponse`) is implemented in `etcdctl.go`

serathius (changes requested):
`addTxnResponse` doesn't use EtcdctlV3, there is no reason to make it a method.

yagikota:
>addTxnResponse doesn't use EtcdctlV3, there is no reason to make it a method.

Good point — you're right, there's no need to make it a method in that case.
But I think making it private would be better like this:

``` diff
- func AddTxnResponse(resp *clientv3.TxnResponse, jsonData string) {
+ func addTxnResponse(resp *clientv3.TxnResponse, jsonData string) {
...
}
```

tests/framework/e2e/e2e_test.go
yagikota: @serathius 
What do you think about it?

serathius:
> But I think making it private would be better like this:

Sure, if a **test package** function is not used publicly we can make it private.

Please don't remove the test. It's just not worth time to review whether it's safe.

PR #20447, opened 2025-08-06
cache: add unit test injecting mid-way compaction

apullo777:
@serathius @MadhavJivrajani


cache/cache_test.go
```diff
@@ -398,35 +399,84 @@ func TestValidateWatchRange(t *testing.T) {
...
-func (kvStub) Do(ctx context.Context, op clientv3.Op) (clientv3.OpResponse, error) {
-	return clientv3.OpResponse{}, nil
+func waitForResetReady(t *testing.T, c *Cache, timeout time.Duration) {
+	deadline := time.Now().Add(timeout)
+	for c.Ready() && time.Now().Before(deadline) {
+		time.Sleep(5 * time.Millisecond)
+	}
+	if c.Ready() {
+		t.Fatal("cache.Ready() never flipped to false after compaction")
+	}
```
serathius, line 470: Why not use `WaitReady`?

cache/cache_test.go
```diff
@@ -398,35 +399,84 @@ func TestValidateWatchRange(t *testing.T) {
...
+		t.Fatalf("initial WaitReady: %v", err)
+	}
 
-func (kvStub) Get(ctx context.Context, key string, _ ...clientv3.OpOption) (*clientv3.GetResponse, error) {
-	return &clientv3.GetResponse{Header: &pb.ResponseHeader{Revision: 0}}, nil
-}
+	t.Log("Phase 2: simulate compaction")
+	mw.errorCompacted(10)
+	waitForResetReady(t, cache, 100*time.Millisecond)
+	if gotRev := cache.store.LatestRev(); gotRev != 0 {
```
serathius: Can you test Get request instead of inspecting internals? We should expect that Get should timeout waiting for ready.

cache/cache_test.go
```diff
@@ -398,35 +399,84 @@ func TestValidateWatchRange(t *testing.T) {
...
+	if err = cache.WaitReady(t.Context()); err != nil {
+		t.Fatalf("initial WaitReady: %v", err)
+	}
 
-func (kvStub) Get(ctx context.Context, key string, _ ...clientv3.OpOption) (*clientv3.GetResponse, error) {
-	return &clientv3.GetResponse{Header: &pb.ResponseHeader{Revision: 0}}, nil
-}
+	t.Log("Phase 2: simulate compaction")
+	mw.errorCompacted(10)
+	waitForResetReady(t, cache, 100*time.Millisecond)
```
serathius: This 100ms has hidden dependency on fact that initialbackoff is much larger. This might also flake in overloaded environment

apullo777: Thanks for the feedback! Maybe should modify the logic to keep checking on every poll interval like below:   

```
	mw.errorCompacted(10)
	waitUntil(t, time.Second, 10*time.Millisecond, func() bool { return !cache.Ready() })
```
using this helper:
```
func waitUntil(t *testing.T, timeout, poll time.Duration, cond func() bool) {
	deadline := time.Now().Add(timeout)
	for time.Now().Before(deadline) {
		if cond() {
			return
		}
		time.Sleep(poll)
	}
	t.Fatalf("condition not satisfied within %s", timeout)
}
```

serathius: This is not what I meant. The cache should become unready only for 1 second, when it will retry and become healthy again.

serathius: In the future we should rewrite ready to lock and signals to allow subscribing.

apullo777: > The cache should become unready only for 1 second, when it will retry and become healthy again.

Just want to make sure if I've captured your intent correctly before I push any changes. So for now we wanted to decouple the test from the `initialbackoff` (currently 50 ms). And for now we could set 1 second to be the upper-bound for the whole "compaction->backoff->get-watch latency" unready window.

If we want to keep `waitUntil` helper, maybe we should (1) adjust the timeout to be a period of time < 1s, (2) pushe the `CreatedNotify` right away, and (3) wraps `WaitReady` in 1 second context (like below). 

Since we already verify the cache is un-ready with `waitUntil(… !cache.Ready())`, verifying with `Get` failure in between becomes redundant?

```
    mw.errorCompacted(10)
	waitUntil(t, 500*time.Millisecond, 10*time.Millisecond, func() bool { return !cache.Ready() })
	
	// drop the Get() timeout assertion
	
	mw.triggerCreatedNotify()
	ctxReady, cancelReady := context.WithTimeout(t.Context(), time.Second)
	defer cancelReady()
	if err := cache.WaitReady(ctxReady); err != nil {
	...
```

cache/cache_test.go
```diff
@@ -398,35 +400,95 @@ func TestValidateWatchRange(t *testing.T) {
...
+		t.Errorf("Watch started at rev=%d; want %d", gotWatchStart, expectedWatchStart)
+	}
+
+	ctx, cancel := context.WithTimeout(t.Context(), time.Second)
+	defer cancel()
+	if err = cache.WaitForRevision(ctx, expectSnapshotRev); err != nil {
+		t.Fatalf("cache never reached rev=%d: %v", expectSnapshotRev, err)
+	}
+
+	gotSnapshot, err := cache.Get(t.Context(), "foo", clientv3.WithSerializable())
```
serathius, line 465: You should also test the content so we ensure that cache was properly cleaned up.

cache/cache_test.go
```diff
@@ -398,35 +400,95 @@ func TestValidateWatchRange(t *testing.T) {
...
-}
+	waitUntil(t, time.Second, 10*time.Millisecond, func() bool { return !cache.Ready() })
+	start := time.Now()
 
-func (kvStub) Delete(ctx context.Context, key string, _ ...clientv3.OpOption) (*clientv3.DeleteResponse, error) {
-	return nil, nil
-}
+	ctxGet, cancelGet := context.WithTimeout(t.Context(), 100*time.Millisecond)
+	defer cancelGet()
+	_, err = cache.Get(ctxGet, "foo", clientv3.WithSerializable())
```
serathius, line 438: Note for the future. If the underlying watch is broken/compacted, we could still serve serializable requests for data we have cached. Dropping is just needed before we start applying new data.

cache/cache_test.go
serathius: We can do that in followup

PR #20453, opened 2025-08-07
cache: refactor ready status using isReady with readyChan and mutex to prevent races

apullo777:

@serathius @MadhavJivrajani


cache/cache.go
```diff
@@ -40,9 +40,11 @@ type Cache struct {
...
 	watcher     clientv3.Watcher
 	kv          clientv3.KV
-	demux       *demux // demux fans incoming events out to active watchers and manages resync.
-	store       *store // last‑observed snapshot
-	ready       chan struct{}
+	demux       *demux        // demux fans incoming events out to active watchers and manages resync.
+	store       *store        // last‑observed snapshot
+	mu          sync.Mutex    // protects isReady & readyChan
+	isReady     bool          // true once snapshot loaded & watch confirmed
+	readyChan   chan struct{} // closed exactly once per ready transition
```
serathius: Please move it to separate structure like in https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/cacher/ready.go

Fact that you have comments to explain fields just means that there are too many fields and you need to build hierarchy.

PR #18804, opened 2024-10-29
Update 3.5 changelog to cover two major fixes

ahrtr:
Update changelog to cover the following two fixes:
- https://github.com/etcd-io/etcd/pull/18784
- https://github.com/etcd-io/etcd/pull/18799

PR #20454, opened 2025-08-08
add e2e to reproduce issue 18667

silentred:
Add an e2e test for #18667

### DIFF
--- a/tests/e2e/reproduce_18667_test.go
+++ b/tests/e2e/reproduce_18667_test.go
@@ -1,0 +1,85 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/tests/v3/framework/e2e"
+)
+
+// TestReproduce18667 reproduces the issue: https://github.com/etcd-io/etcd/issues/18667.
+func TestReproduce18667(t *testing.T) {
+	e2e.BeforeTest(t)
+
+	ctx := t.Context()
+	clus, cerr := e2e.NewEtcdProcessCluster(ctx, t,
+		e2e.WithClusterSize(3),
+		e2e.WithSnapshotCount(1000),
+	)
+	require.NoError(t, cerr)
+
+	t.Cleanup(func() { clus.Stop() })
+
+	targetIdx := rand.Intn(3)
+	cli := newClient(t, clus.Procs[targetIdx].EndpointsGRPC(), e2e.ClientConfig{})
+
+	t.Log("Put key-value [k1...k20]")
+	var revision int64
+	for i := 1; i <= 20; i++ {
+		resp, err := cli.Put(ctx, fmt.Sprintf("k%d", i), fmt.Sprintf("v%d", i))
+		require.NoError(t, err)
+		revision = resp.Header.Revision
+	}
+
+	t.Logf("Compact on the latest revision %d", revision)
+	_, err := cli.Compact(ctx, revision)
+	require.NoError(t, err)
+
+	t.Log("Send txn to make data inconsistent among etcdservers")
+	cmp := clientv3.Compare(clientv3.Value("k1"), "=", "v1")
+	then := []clientv3.Op{
+		clientv3.OpPut("k2", "foo"),
+		clientv3.OpGet("k1", clientv3.WithRev(revision-1)),
+	}
+	_, err = cli.Txn(ctx).If(cmp).Then(then...).Commit()
+	require.Error(t, err)
+	require.Contains(t, err.Error(), rpctypes.ErrCompacted.Error())
+
+	t.Log("Verify all members have consistent data on key 'k2'")
+	for i := 0; i < clus.Cfg.ClusterSize; i++ {
+		idx := (targetIdx + i) % clus.Cfg.ClusterSize
+		cli = newClient(t, clus.Procs[idx].EndpointsGRPC(), e2e.ClientConfig{})
+		resp, err := cli.Get(ctx, "k2")
+		require.NoError(t, err)
+		// TODO: All members are supposed to have the same key/value pair k2:v2.
+		// However, the issue https://github.com/etcd-io/etcd/issues/18667
+		// hasn't been resolved yet, so some members hold the wrong data
+		// "k2:foo". Once the issue is fixed, we should remove the if-else
+		// branch below, and all member should have consistent data k2:v2.
+		if i == 0 {
+			assert.Equal(t, "v2", string(resp.Kvs[0].Value))
+		} else {
+			assert.Equal(t, "foo", string(resp.Kvs[0].Value))
+		}
+	}
+}


tests/e2e/reproduce_18667_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Put key-value [k1...k20], revision = 21
+	n := 20
+	for i := 1; i <= n; i++ {
+		_, err := cli.Put(ctx, fmt.Sprintf("k%d", i), fmt.Sprintf("v%d", i))
+		require.NoError(t, err)
+	}
+
+	// Compact revision 21
+	_, err := cli.Compact(ctx, int64(n+1))
```
serathius:
```suggestion
    var revision int64
	for i := 1; i <= 20; i++ {
		resp, err := cli.Put(ctx, fmt.Sprintf("k%d", i), fmt.Sprintf("v%d", i))
		require.NoError(t, err)
		revision = resp.Header.Revision
	}

	// Compact revision 21
	_, err := cli.Compact(ctx, revision)
```

tests/e2e/reproduce_18667_test.go
```diff
@@ -0,0 +1,79 @@
...
+		e2e.WithSnapshotCount(1000),
+	)
+	require.NoError(t, cerr)
+
+	t.Cleanup(func() { clus.Stop() })
+
+	targetIdx := rand.Intn(3)
+	cli := newClient(t, clus.Procs[targetIdx].EndpointsGRPC(), e2e.ClientConfig{})
+
+	// Put key-value [k1...k20]
```
serathius:
```suggestion
	t.Log("Put key-value [k1...k20]")
```
Recommend using logs instead of comments to mark sections of the test.

serathius:
Have you tested it against old etcd release to confirm reproduction? If so, please you paste logs to confirm.

silentred:
> Have you tested it against old etcd release to confirm reproduction? If so, please you paste logs to confirm.

Is it enough to test `release-3.4`, `release-3.5` and `release-3.6`?

```
# for release-3.4
$bin/etcd --version
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

etcd Version: 3.4.37
Git SHA: 79dc9a1
Go Version: go1.23.11
Go OS/Arch: linux/amd64

$go test -timeout 30s -run ^TestReproduce18667$ go.etcd.io/etcd/tests/v3/e2e
ok  	go.etcd.io/etcd/tests/v3/e2e	4.444s

# for release-3.5
$bin/etcd --version
etcd Version: 3.5.17
Git SHA: 8f6c30d
Go Version: go1.22.9
Go OS/Arch: linux/amd64

$go test -timeout 30s -run ^TestReproduce18667$ go.etcd.io/etcd/tests/v3/e2e
ok  	go.etcd.io/etcd/tests/v3/e2e	4.492s

# for release-3.6
$bin/etcd --version
etcd Version: 3.6.4
Git SHA: 2cf8fb0
Go Version: go1.23.11
Go OS/Arch: linux/amd64

$go test -timeout 30s -run ^TestReproduce18667$ go.etcd.io/etcd/tests/v3/e2e
ok  	go.etcd.io/etcd/tests/v3/e2e	2.402s
```

serathius:
The bug was fixed in v3.5.17 https://github.com/etcd-io/etcd/pull/18804/files

So we should run this test o v3.5.16 and confirm it fails.

ahrtr:
> The bug was fixed in v3.5.17 https://github.com/etcd-io/etcd/pull/18804/files
> 
> So we should run this test o v3.5.16 and confirm it fails.

I don't think this issue has been resolved. Since K8s isn't affected, (and likely normally users don't specify revision in range request in TXN), so it isn't a priority, at least not a critical issue.

I do not get time to look into this PR, but curiously why we do not see workflow failure because as mentioned above the issue hasn't been resolved yet.

serathius:
> I do not get time to look into this PR, but curiously why we do not see workflow failure because as mentioned above the issue hasn't been resolved yet.

Think the order of operations in TXN should be reversed. Read with revision should be before PUT.

silentred:
> The bug was fixed in v3.5.17 https://github.com/etcd-io/etcd/pull/18804/files

I think this bug has not been fixed yet. The `ErrCompacted` in this test case is thrown by `checkTxn`, before `txn` which is modified in PR https://github.com/etcd-io/etcd/pull/18804. The code is as below.
```
func Txn(ctx context.Context, lg *zap.Logger, rt *pb.TxnRequest, txnModeWriteWithSharedBuffer bool, kv mvcc.KV, lessor lease.Lessor) (txnResp *pb.TxnResponse, trace *traceutil.Trace, err error) {
    // .......
    // checkTxn throw errors out if there is invalid range in txn.Success or txn.Failure
    _, err = checkTxn(trace, txnRead, rt, lessor, txnPath)
	if err != nil {
		txnRead.End()
		return nil, nil, err
	}
	trace.Step("check requests")
    // ......
	txnResp, err = txn(ctx, lg, txnWrite, rt, isWrite, txnPath)
	txnWrite.End()
}
```

> I do not get time to look into this PR, but curiously why we do not see workflow failure because as mentioned above the issue hasn't been resolved yet.

This test case asserts the data inconsistent among etcdservers. So the test will pass.
```
        t.Log("Verify: k2 equals to v2 in the etcdserver which the Txn was sent to")
	kv2, err := cli.Get(ctx, "k2")
	require.NoError(t, err)
	assert.Equal(t, "v2", string(kv2.Kvs[0].Value))

	t.Log("Verify: k2 equals to foo in other etcdservers")
	targetIdx = (targetIdx + 1) % clus.Cfg.ClusterSize
	cli = newClient(t, clus.Procs[targetIdx].EndpointsGRPC(), e2e.ClientConfig{})
	followerK2, err := cli.Get(ctx, "k2")
	require.NoError(t, err)
	assert.Equal(t, "foo", string(followerK2.Kvs[0].Value))
```


> Think the order of operations in TXN should be reversed. Read with revision should be before PUT.

The order should be irrelevant as I explained above. The root cause is that the etcdserver which is receiving the Txn, changes the Txn request then sends to other members.

cc @ahrtr @serathius

serathius:
Make sense, thanks for detailed explanation.

```
	t.Log("Verify: k2 equals to v2 in the etcdserver which the Txn was sent to")
	kv2, err := cli.Get(ctx, "k2")
	require.NoError(t, err)
	assert.Equal(t, "v2", string(kv2.Kvs[0].Value))

	t.Log("Verify: k2 equals to foo in other etcdservers")
```

This is a test that confirms a bug. Can you add a TODO to update the test when we fix the bug?

serathius (approved):

tests/e2e/reproduce_18667_test.go
```diff
@@ -0,0 +1,79 @@
...
+	kv2, err := cli.Get(ctx, "k2")
+	require.NoError(t, err)
+	assert.Equal(t, "v2", string(kv2.Kvs[0].Value))
+
+	t.Log("Verify: k2 equals to foo in other etcdservers")
+	targetIdx = (targetIdx + 1) % clus.Cfg.ClusterSize
+	cli = newClient(t, clus.Procs[targetIdx].EndpointsGRPC(), e2e.ClientConfig{})
+	followerK2, err := cli.Get(ctx, "k2")
+	require.NoError(t, err)
+	assert.Equal(t, "foo", string(followerK2.Kvs[0].Value))
```
ahrtr:
```suggestion
	t.Log("Verify all members have consistent data on key 'k2'")
	for i := 0; i < clus.Cfg.ClusterSize; i++ {
		idx := (targetIdx + i) % clus.Cfg.ClusterSize
		cli = newClient(t, clus.Procs[idx].EndpointsGRPC(), e2e.ClientConfig{})
		resp, err := cli.Get(ctx, "k2")
		require.NoError(t, err)
		// All members are supposed to have the same key/value pair k2:v2.
		// However, the issue https://github.com/etcd-io/etcd/issues/18667
		// hasn't been resolved yet, so some members hold the wrong data
		// "k2:foo". Once the issue is fixed, we should remove the if-else
		// branch below, and all member should have consistent data k2:v2.
		if i == 0 {
			assert.Equal(t, "v2", string(resp.Kvs[0].Value))
		} else {
			assert.Equal(t, "foo", string(resp.Kvs[0].Value))
		}
	}		
```

tests/e2e/reproduce_18667_test.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2024 The etcd Authors
```
ahrtr:
```suggestion
// Copyright 2025 The etcd Authors
```

ahrtr (approved):
LGTM with a minor comment, thx

PR #20457, opened 2025-08-09
Fix broken failpoint labelRaftDropHeartbeat

### LINKED ISSUE #20455: [Robustness Tests] Broken failpoint - labelRaftDropHeartbeat
### Bug report criteria

- [x] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.

### What happened?

When executing `make gofail-enable`, `labelRaftDropHeartbeat` doesn't work as expected.  Need to investigate why that is the case.

`make gofail-enable` and then `git diff` should show that the failpoint `labelRaftDropHeartbeat` has been altered in a way that the marker comment becomes a failpoint.

```
make gofail-enable
git diff
```
Then search for `labelRaftDropHeartbeat`

_No response_

N/A

N/A

N/A

### Relevant log output

```Shell

```

nwnt:
Fix #20455 

@serathius

### DIFF
--- a/server/etcdserver/api/rafthttp/stream.go
+++ b/server/etcdserver/api/rafthttp/stream.go
@@ -498,7 +498,7 @@ func (cr *streamReader) decodeLoop(rc io.ReadCloser, t streamType) error {
 			return err
 		}
 
-		// gofail-go: var raftDropHeartbeat struct{}
+		// gofail: var raftDropHeartbeat struct{}
 		// continue labelRaftDropHeartbeat
 		receivedBytes.WithLabelValues(types.ID(m.From).String()).Add(float64(m.Size()))
 
--- a/tests/robustness/Makefile
+++ b/tests/robustness/Makefile
@@ -86,7 +86,7 @@ install-gofail: $(GOPATH)/bin/gofail
 
 .PHONY: gofail-enable
 gofail-enable: $(GOPATH)/bin/gofail
-	$(GOPATH)/bin/gofail enable server/etcdserver/ server/lease/leasehttp server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/ server/etcdserver/api/membership/
+	$(GOPATH)/bin/gofail enable server/etcdserver/ server/lease/leasehttp server/storage/backend/ server/storage/mvcc/ server/storage/wal/ server/etcdserver/api/v3rpc/ server/etcdserver/api/membership/ server/etcdserver/api/rafthttp/
 	cd $(REPOSITORY_ROOT)/server && go get go.etcd.io/gofail@${GOFAIL_VERSION}
 	cd $(REPOSITORY_ROOT)/etcdutl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
 	cd $(REPOSITORY_ROOT)/etcdctl && go get go.etcd.io/gofail@${GOFAIL_VERSION}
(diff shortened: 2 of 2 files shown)


serathius:
Have you looked into background why the failpoint was broken?

nwnt:
Can you elaborate more about the background? My understanding was the tag wasn't correct and the folder path was not included in the make target, which may not be accurate.

nwnt:
@serathius sorry this took me some time.  I still don't understand the history completely, but I'm more confident now that this fix for the broken failpoint should be ok.  Based on this [issue](https://github.com/etcd-io/gofail/issues/14) in `gofail`, `gofail-go` was introduced in this [PR](https://github.com/etcd-io/gofail/pull/15) for releasing the lock in the failpoint injected in a loop without using defer because defer may never release the lock.  Now that @ahrtr [removed locking](https://github.com/etcd-io/gofail/issues/14#issuecomment-1326847226) in a failpoint, we do not need this special syntax anymore and a typical `gofail` should suffice.  

Let me know if I missed anything.

server/etcdserver/api/rafthttp/stream.go
```diff
@@ -498,7 +498,7 @@ func (cr *streamReader) decodeLoop(rc io.ReadCloser, t streamType) error {
 			return err
 		}
 
-		// gofail-go: var raftDropHeartbeat struct{}
+		// gofail: var raftDropHeartbeat struct{}
 		// continue labelRaftDropHeartbeat
```
serathius, line 502: Is the line `continue labelRaftDropHeartbeat` needed?

nwnt: Yes, that is needed.  Here is the diff after running go-fail enable:
```
diff --git a/server/etcdserver/api/rafthttp/stream.go b/server/etcdserver/api/rafthttp/stream.go
index cd4e11e1f..e4073da12 100644
--- a/server/etcdserver/api/rafthttp/stream.go
+++ b/server/etcdserver/api/rafthttp/stream.go
@@ -488,7 +488,8 @@ func (cr *streamReader) decodeLoop(rc io.ReadCloser, t streamType) error {
        }
        cr.mu.Unlock()

-       // gofail: labelRaftDropHeartbeat:
+       /* gofail-label */
+labelRaftDropHeartbeat:
        for {
                m, err := dec.decode()
                if err != nil {
@@ -498,8 +499,17 @@ func (cr *streamReader) decodeLoop(rc io.ReadCloser, t streamType) error {
                        return err
                }

-               // gofail: var raftDropHeartbeat struct{}
-               // continue labelRaftDropHeartbeat
+               if vraftDropHeartbeat, __fpErr := __fp_raftDropHeartbeat.Acquire(); __fpErr == nil {
+                       _, __fpTypeOK := vraftDropHeartbeat.(struct{})
+                       if !__fpTypeOK {
+                               goto __badTyperaftDropHeartbeat
+                       }
+                       continue labelRaftDropHeartbeat
+                       goto __nomockraftDropHeartbeat
+               __badTyperaftDropHeartbeat:
+                       __fp_raftDropHeartbeat.BadType(vraftDropHeartbeat, "struct{}")
+               __nomockraftDropHeartbeat:
+               }
                receivedBytes.WithLabelValues(types.ID(m.From).String()).Add(float64(m.Size()))

                cr.mu.Lock()
```

serathius:
We might also want, to register the failpoint in robustness tests. https://github.com/etcd-io/etcd/blob/main/tests/robustness/failpoint/failpoint.go

PR #20465, opened 2025-08-11
cache: generalize ready channel to signal any state change instead of ready-only

apullo777:

This PR generalized the `readyChan` into a generic state change notification (now `stateChan`) by closing and replacing the channel on any state transition, so that we can re-use the same mechanism to wait for both ready and not ready.

@serathius @MadhavJivrajani

### DIFF
--- a/cache/ready.go
+++ b/cache/ready.go
@@ -19,16 +19,15 @@ import (
 	"sync"
 )
 
-// ready tracks the transition from “not ready” → “ready” and allows waiters to block
-// until the first transition has happened, and to be reset on error.
+// ready tracks readiness state changes and allows callers to wait for a target state
 type ready struct {
-	mu        sync.Mutex
-	isReady   bool
-	readyChan chan struct{} // closed exactly once per ready transition
+	mu      sync.Mutex
+	isReady bool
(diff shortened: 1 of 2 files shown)


cache/cache.go
```diff
@@ -182,6 +182,11 @@ func (c *Cache) WaitReady(ctx context.Context) error {
 	return c.ready.WaitReady(ctx)
 }
 
+// WaitNotReady blocks until the cache becomes not ready or the ctx is cancelled..
+func (c *Cache) WaitNotReady(ctx context.Context) error {
```
serathius: Do we want to make it public? I assumed we need it for unit testing. I don't see any callers yet.

serathius:
Maybe we could add some unit tests? 

/ok-to-test

cache/cache.go
apullo777: Yes, it is now only needed for the unit test in #20447. Should I make it `waitNotReady`, or  let the test call `ready.WaitNotReady` directly?

apullo777:
> Maybe we could add some unit tests?

Ok! Will add some unit tests in this PR!

PR #20475, opened 2025-08-12
Add OpenTelemetry Spans to Watch

AwesomePatrol:

Attributes added:
* `key`, `range_end`
* `start_rev`
* `prev_kv`, `fragment`, `progress_notify`

As `Watch` is a long-running operation I did not add any events as it would risk significantly increasing memory consumption.

In order to associate `Span` from gRPC call to `Watch` with actual `watch` `Span`s, I added `ctx` argument to `watch` method. It caused significant diff in tests.

Example in Jaeger:
image

### DIFF
--- a/server/storage/mvcc/watchable_store.go
+++ b/server/storage/mvcc/watchable_store.go
@@ -15,11 +15,9 @@
 package mvcc
 
 import (
-	"context"
 	"sync"
 	"time"
 
-	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	"go.etcd.io/etcd/api/v3/mvccpb"
@@ -48,7 +46,7 @@ var (
 func ChanBufLen() int { return chanBufLen }
 
 type watchable interface {
-	watch(ctx context.Context, key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc)
+	watch(key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc)
 	progress(w *watcher)
 	progressAll(watchers map[WatchID]*watcher) bool
 	rev() int64
@@ -126,7 +124,7 @@ func (s *watchableStore) NewWatchStream() WatchStream {
 	}
 }
 
-func (s *watchableStore) watch(ctx context.Context, key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc) {
+func (s *watchableStore) watch(key, end []byte, startRev int64, id WatchID, ch chan<- WatchResponse, fcs ...FilterFunc) (*watcher, cancelFunc) {
 	wa := &watcher{
 		key:      key,
 		end:      end,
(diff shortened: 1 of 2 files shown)


server/storage/mvcc/watchable_store.go
```diff
@@ -199,6 +202,7 @@ func (s *watchableStore) cancelWatcher(wa *watcher) {
 		time.Sleep(time.Millisecond)
 	}
 
+	wa.span.End()
```
serathius: We should be careful to make sure that we don't leak any spans. Are you sure all watchers opened will be closed through `cancelWatcher`?

serathius (approved):

server/etcdserver/api/v3rpc/watch.go
```diff
@@ -304,8 +307,16 @@ func (sws *serverWatchStream) recvLoop() error {
 			}
 
 			filters := FiltersFromRequest(creq)
-
-			id, err := sws.watchStream.Watch(mvcc.WatchID(creq.WatchId), creq.Key, creq.RangeEnd, creq.StartRevision, filters...)
+			ctx, _ := traceutil.Tracer.Start(sws.gRPCStream.Context(), "watch", trace.WithAttributes(
```
serathius, line 310: Shouldn't  we add a test for watch like for other methods?

server/storage/mvcc/watchable_store.go
AwesomePatrol: Yes, `Watch` always calls `sws.close()` before returning which will go through all the `cancel` functions and invoke them: https://github.com/etcd-io/etcd/blob/v3.6.4/server/etcdserver/api/v3rpc/watch.go#L226

Alternatively, we can store `span` reference alongside `cancels` for the same effect (and maybe avoid modifying `watch`)

server/etcdserver/api/v3rpc/watch.go
AwesomePatrol: Added to integration tests

serathius (approved):

server/storage/mvcc/watchable_store.go
fuweid: it looks like we do it in this way.

```
return wa, func() {
       defer span.End()
       s.cancelWatcher(wa)
}
```

AwesomePatrol: I moved it one-layer up, so it will be:
```
	ws.cancels[id] = func() {
		defer trace.SpanFromContext(ctx).End()
		c()
	}
```
Which one do you prefer?

There is also an option to keep both `trace.Start` and `span.End` on the same abstraction layer, but then the attributes (`prev_kv`, `fragment`, `progress_notify`) couldn't be used for filtering

server/storage/mvcc/watcher.go
```diff
@@ -134,7 +137,10 @@ func (ws *watchStream) Watch(id WatchID, key, end []byte, startRev int64, fcs ..
 
 	w, c := ws.watchable.watch(key, end, startRev, id, ws.ch, fcs...)
 
-	ws.cancels[id] = c
+	ws.cancels[id] = func() {
+		defer trace.SpanFromContext(ctx).End()
```
fuweid: I think we should call trace.SpanFromContext(ctx) out of func() {} first.

AwesomePatrol: Changed to:
```go
	span := trace.SpanFromContext(ctx)
	ws.cancels[id] = func() {
		defer span.End()
		c()
	}
```

PR #20476, opened 2025-08-12
cache: preserve cached snapshot after watch errors

apullo777:
@serathius @MadhavJivrajani 

I open this PR to make the cache preserve cached snapshot after watch error and only only reset it on new snapshot (so `cache.Get` is now only blocked on cold start). That includes:
- Removed` store.Reset` as now only `store.Restore` is needed for resetting latest-observed snapshot.
- Modified `TestCacheCompactionResync` to also verify both the snapshot content and whether cached data can be served during compaction.


cache/cache_test.go
```diff
@@ -426,6 +431,10 @@ func TestCacheCompactionResync(t *testing.T) {
 	if err = cache.WaitReady(t.Context()); err != nil {
 		t.Fatalf("initial WaitReady: %v", err)
 	}
+	verifySnapshot(t, cache, map[string]string{
+		"foo": "old_value",
+		"bar": "old_bar",
+	})
```
serathius: Please test whole response, not just key-value, but also modrevision, version etc

cache/cache_test.go
```diff
@@ -655,3 +674,19 @@ func collectAndAssertAtomicEvents(ctx context.Context, t *testing.T, watchCh cli
...
 	}
 }
+
+func verifySnapshot(t *testing.T, cache *Cache, want map[string]string) {
+	allKeys, err := cache.Get(t.Context(), "", clientv3.WithPrefix(), clientv3.WithSerializable())
+	if err != nil {
+		t.Fatalf("Get all keys: %v", err)
+	}
+
+	got := make(map[string]string)
```
serathius: Please don't use map, order of kvs in response matters.

serathius (approved):

PR #20477, opened 2025-08-12
cache: make ring buffer generic for events and snapshots

apullo777:
@serathius @MadhavJivrajani 

Opened this PR to makes ring buffer generic (not just for events but for snapshots in future). This includes:
- renamed `Filter` as `GetFromRevision` and apply binary search instead of full scan
- added a few new functions like `FindAtOrBefore` which might be useful for snapshots

For now I use a `RevisionOf[T]` function to extract revisions from any item type, maybe should later switch to a wrapper that contains the revision directly

### DIFF
--- a/cache/demux.go
+++ b/cache/demux.go
@@ -27,7 +27,7 @@ type demux struct {
 	// activeWatchers & laggingWatchers hold the first revision the watcher still needs (nextRev).
 	activeWatchers  map[*watcher]int64
 	laggingWatchers map[*watcher]int64
-	history         *ringBuffer
+	history         ringBuffer[*clientv3.Event]
 	resyncInterval  time.Duration
 }
 
@@ -45,7 +45,7 @@ func newDemux(historyWindowSize int, resyncInterval time.Duration) *demux {
 	return &demux{
 		activeWatchers:  make(map[*watcher]int64),
 		laggingWatchers: make(map[*watcher]int64),
-		history:         newRingBuffer(historyWindowSize),
+		history:         *newRingBuffer(historyWindowSize, func(ev *clientv3.Event) int64 { return ev.Kv.ModRevision }),
 		resyncInterval:  resyncInterval,
 	}
 }
--- a/cache/ringbuffer.go
+++ b/cache/ringbuffer.go
@@ -14,56 +14,55 @@
 
 package cache
 
-import (
-	"fmt"
+import "fmt"
 
-	clientv3 "go.etcd.io/etcd/client/v3"
-)
-
-type ringBuffer struct {
-	buffer []batch
+type ringBuffer[T any] struct {
+	buffer []entry[T]
 	// head is the index immediately after the last non-empty entry in the buffer (i.e., the next write position).
 	head, tail, size int
+	revisionOf       RevisionOf[T]
 }
 
-// batch groups all events that share one ModRevision.
-type batch struct {
-	rev    int64
-	events []*clientv3.Event
+type entry[T any] struct {
+	revision int64
+	items    []T
 }
 
-type KeyPredicate = func([]byte) bool
+type (
+	KeyPredicate      = func([]byte) bool
+	RevisionOf[T any] func(T) int64
+	IterFunc[T any]   func(rev int64, items []T) bool
+)
 
-type IterFunc func(rev int64, events []*clientv3.Event) bool
-
-func newRingBuffer(capacity int) *ringBuffer {
+func newRingBuffer[T any](capacity int, revisionOf RevisionOf[T]) *ringBuffer[T] {
 	// assume capacity > 0 – validated by Cache
-	return &ringBuffer{
-		buffer: make([]batch, capacity),
+	return &ringBuffer[T]{
+		buffer:     make([]entry[T], capacity),
+		revisionOf: revisionOf,
 	}
 }
 
-func (r *ringBuffer) Append(events []*clientv3.Event) {
+func (r *ringBuffer[T]) Append(items []T) {
 	start := 0
-	for end := 1; end < len(events); end++ {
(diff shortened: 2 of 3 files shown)


cache/ringbuffer.go
```diff
@@ -14,102 +14,164 @@
...
-		buffer: make([]batch, capacity),
+	return &ringBuffer[T]{
+		buffer: make([]entry[T], capacity),
 	}
 }
 
-func (r *ringBuffer) Append(events []*clientv3.Event) {
+type RevisionOf[T any] func(T) int64
+
+func (r *ringBuffer[T]) Append(items []T, revOf RevisionOf[T]) {
```
serathius: This function `revOf RevisionOf[T]` should be part of ringBuffer configuration during construction instead of passed on every `Append` call

cache/ringbuffer.go
```diff
@@ -14,102 +14,164 @@
...
-	clientv3 "go.etcd.io/etcd/client/v3"
-)
-
-type ringBuffer struct {
-	buffer []batch
+type ringBuffer[T any] struct {
+	buffer []entry[T]
 	// head is the index immediately after the last non-empty entry in the buffer (i.e., the next write position).
-	head, tail, size int
+	head, tail, entryCount int
```
serathius: `size` still makes sense. No need to change it.

cache/ringbuffer.go
```diff
@@ -14,102 +14,164 @@
...
+	r.buffer[r.head] = entry
 	r.head = (r.head + 1) % len(r.buffer)
 }
 
-// Filter returns all events in the buffer whose ModRevision is >= minRev.
-// TODO: use binary search on the ring buffer to locate the first entry >= nextRev instead of a full scan
-func (r *ringBuffer) Filter(minRev int64) (eventBatches [][]*clientv3.Event) {
-	eventBatches = make([][]*clientv3.Event, 0, r.size)
+// GetFromRev returns all entries with revision >= minRev
+func (r *ringBuffer[T]) GetFromRevision(minRev int64) [][]T {
```
serathius: Maybe we could use naming from btree implementation, for this function we could use name `AscendGreatherOrEqual`

https://github.com/google/btree/blob/aeba20f7a1e1315badec4eca4fdc9f754f5f880a/btree_generic.go#L766

serathius: To make ringBuffer interface cleaner, maybe we could flatten `[][]T` to `[]T` and expect caller to handle batching. We can do that in followup to avoid complicating this PR.

serathius: Or maybe instead accept iterator like btree https://github.com/google/btree/blob/aeba20f7a1e1315badec4eca4fdc9f754f5f880a/btree_generic.go#L127

cache/ringbuffer.go
```diff
@@ -14,102 +14,164 @@
...
-	for n, i := 0, r.tail; n < r.size; n, i = n+1, (i+1)%len(r.buffer) {
-		eventBatch := r.buffer[i]
-		if eventBatch.events == nil {
-			panic(fmt.Sprintf("ringBuffer.Filter: unexpected nil eventBatch at index %d", i))
-		}
-		if eventBatch.rev >= minRev {
-			eventBatches = append(eventBatches, eventBatch.events)
-		}
+// FindAtOrBefore finds the entry with the highest revision that is <= targetRev
+func (r *ringBuffer[T]) FindAtOrBefore(targetRev int64) (entry[T], bool) {
```
serathius: If we changed interface to iterators like in btree we could implement `DescendLessOrEqual` method and allow client to stop iteration after first result.

https://github.com/google/btree/blob/aeba20f7a1e1315badec4eca4fdc9f754f5f880a/btree_generic.go#L1040

cache/ringbuffer.go
```diff
@@ -14,102 +14,164 @@
...
+	head, tail, entryCount int
 }
 
-// batch groups all events that share one ModRevision.
-type batch struct {
-	rev    int64
-	events []*clientv3.Event
+type entry[T any] struct {
+	revision int64
+	items    []T
```
serathius, line 28: Maybe instead:
* entry having `items []T`
* demux storing `history *ringBuffer[*clientv3.Event]`

We could implement:
* entry having `item T`
* demux storing `history *ringBuffer[[]*clientv3.Event]`

cache/demux.go
```diff
@@ -27,7 +27,7 @@ type demux struct {
 	// activeWatchers & laggingWatchers hold the first revision the watcher still needs (nextRev).
 	activeWatchers  map[*watcher]int64
 	laggingWatchers map[*watcher]int64
-	history         *ringBuffer
```
serathius, line 30: FYI, you don't need a `history *ringBuffer`, `history ringBuffer` would be enough and would be cleaner and avoid additional dereference.

cache/ringbuffer.go
apullo777: Now with the iterator approach adopted, maybe we could still move the batching logic to demux to make ringbuffer interface more generic (with entry having `item T`) in a follow-up PR.

cache/ringbuffer.go
```diff
@@ -14,181 +14,240 @@
...
-		if eventBatch.events == nil {
-			panic(fmt.Sprintf("ringBuffer.AscendGreaterOrEqual: unexpected nil at %d", i))
-		}
-
-		if eventBatch.rev < pivot {
-			continue
-		}
-
-		if !iter(eventBatch.rev, eventBatch.events) {
+	iterStartIndex := r.findFirstGreaterOrEqualIndex(pivot)
```
serathius: Don't know why did you extracted this logic. Please don't mix changes in single PR.

serathius:
> added a few new functions like FindAtOrBefore which might be useful for snapshots

Please move to separate PR.

/ok-to-test

PR #20483, opened 2025-08-13
cache: make store a normal demux watcher to reduce lock contention

apullo777:
@serathius  @MadhavJivrajani

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -78,7 +78,7 @@ func New(client *clientv3.Client, prefix string, opts ...Option) (*Cache, error)
 	cache.waitGroup.Add(1)
 	go func() {
 		defer cache.waitGroup.Done()
-		cache.getWatchLoop(internalCtx)
+		cache.getWatchLoop()
 	}()
 
 	return cache, nil
@@ -203,14 +203,15 @@ func (c *Cache) Close() {
 	c.waitGroup.Wait()
 }
 
-func (c *Cache) getWatchLoop(ctx context.Context) {
+func (c *Cache) getWatchLoop() {
 	cfg := defaultConfig()
+	ctx := c.internalCtx
 	backoff := cfg.InitialBackoff
 	for {
 		if err := ctx.Err(); err != nil {
 			return
 		}
-		if err := c.getWatch(ctx); err != nil {
+		if err := c.getWatch(); err != nil {
 			fmt.Printf("getWatch failed, will retry after %v: %v\n", backoff, err)
 		}
 		select {
@@ -221,12 +222,12 @@ func (c *Cache) getWatchLoop(ctx context.Context) {
 	}
 }
 
-func (c *Cache) getWatch(ctx context.Context) error {
-	getResp, err := c.get(ctx)
+func (c *Cache) getWatch() error {
+	getResp, err := c.get(c.internalCtx)
 	if err != nil {
 		return err
 	}
-	return c.watch(ctx, getResp.Header.Revision+1)
+	return c.watch(getResp.Header.Revision + 1)
 }
 
 func (c *Cache) get(ctx context.Context) (*clientv3.GetResponse, error) {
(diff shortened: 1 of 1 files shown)


cache/cache.go
```diff
@@ -241,6 +241,12 @@ func (c *Cache) get(ctx context.Context) (*clientv3.GetResponse, error) {
 func (c *Cache) watch(ctx context.Context, rev int64) error {
 	readyOnce := sync.Once{}
 	for {
+		storeW := newWatcher(c.cfg.PerWatcherBufferSize*4, nil)
```
serathius: Why `*4`?

cache/cache.go
```diff
@@ -250,24 +256,61 @@ func (c *Cache) watch(ctx context.Context, rev int64) error {
...
-				return err
+		err := c.watchEvents(ctx, watchCh, applyErr, &readyOnce)
+		c.demux.Unregister(storeW)
+
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		if err != nil {
+			return err
+		}
```
serathius: If context is done, wouldn't `c.watchEvents` exit and return the same error as `ctx.Err()`?

cache/cache.go
```diff
@@ -241,6 +241,12 @@ func (c *Cache) get(ctx context.Context) (*clientv3.GetResponse, error) {
 func (c *Cache) watch(ctx context.Context, rev int64) error {
 	readyOnce := sync.Once{}
 	for {
+		storeW := newWatcher(c.cfg.PerWatcherBufferSize*4, nil)
+		c.demux.Register(storeW, rev)
+		applyErr := make(chan error, 1)
+		c.waitGroup.Add(1)
+		go c.applyStorage(ctx, storeW, applyErr)
```
```suggestion
		c.waitGroup.Add(1)
		go func() {
				defer c.waitGroup.Done()
				go c.applyStorage(ctx, storeW, applyErr)
		}()

```

serathius: Add + Done should be next to each other so it's easy to verify their paring.

cache/cache.go
```diff
@@ -250,24 +256,61 @@ func (c *Cache) watch(ctx context.Context, rev int64) error {
...
+			return err
+		}
+	}
+}
+
+func (c *Cache) applyStorage(ctx context.Context, storeW *watcher, applyErr chan<- error) {
+	defer c.waitGroup.Done()
+	for {
+		select {
+		case <-ctx.Done():
```
serathius: Isn't ctx passed here is the same context as c.internalCtx. Recommend to remove ctx from arguments if you plan to have it as method on Cache.

cache/cache.go
```diff
@@ -250,24 +256,61 @@ func (c *Cache) watch(ctx context.Context, rev int64) error {
...
+			return
+		case events, ok := <-storeW.eventQueue:
+			if !ok {
+				return
+			}
+			if err := c.store.Apply(events); err != nil {
+				select {
+				case applyErr <- err:
+				default:
+				}
```
```suggestion
				applyErr <- err
```
I don't think you should skip error. Even if this should not happen it's bad to imply that sending error is optional.

serathius: Or better would be to have the `applyStorage` function return the error, and do putting to channel next to channel creation where you see that there is only 1 writer to channel of size 1.

cache/cache.go
```diff
@@ -250,24 +256,61 @@ func (c *Cache) watch(ctx context.Context, rev int64) error {
...
 			}
+		}
+	}
+}
 
-			if err := c.store.Apply(resp.Events); err != nil {
+func (c *Cache) watchEvents(ctx context.Context, watchCh clientv3.WatchChan, applyErr <-chan error, readyOnce *sync.Once) error {
+	for {
+		select {
+		case <-ctx.Done():
```
serathius: Decide on one option, do we select on `ctx` passed as argument or `c.internalCtx`.

serathius: Maybe we could resign from using context for internal cancelation and instead use stopChan to differentiate.

cache/cache.go
apullo777: I was thinking that perhaps the store watcher needs a larger buffer to handle the full event stream, while normal watcher will normally handle filtered subset. Should we make this configurable like to have a separate `StoreWatcherBufferSize`? Or just using `PerWatcherBufferSize` is enough?

serathius: Don't do magic numbers, or guess numbers. If the default is good enough users, it should be good enough for store.

cache/cache.go
apullo777: Since we always call `getWatchLoop` with `internalCtx`, maybe we can remove the `ctx` parameter from those internal methods for clarity (`getWatchLoop`, `getWatch`, `watch`, `applyStorage`, and `watchEvents`; `get` still need `ctx` to pass on to `Get`).

`stopChan` sounds great! But I will need to see if the change will affect `WaitReady`. Will do that in a follow-up PR.

serathius:
We got a flake after merging this PR https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/directory/pull-etcd-integration-1-cpu-arm64/1957629391110410240 

Can you take a look?

Issue #20488, opened 2025-08-14
Client Cache breaks Resumable guarantee


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Cache is a experimental library https://github.com/etcd-io/etcd/tree/main/cache developed as part of GSoC https://github.com/etcd-io/etcd/issues/19371

Run robustness test on latest implementation of cache https://github.com/etcd-io/etcd/pull/20354

If found the issue
```
    logger.go:146: 2025-08-14T11:30:16.618+0200 ERROR   Broke watch guarantee   
{
  "guarantee": "resumable",
  "client": 3,
  "request": {
    "Key": "",
    "Revision": 989,
    "WithPrefix": true,
    "WithProgressNotify": false,
    "WithPrevKV": false
  },
  "got-event": {
    "Type": "put-operation",
    "Key": "key42",
    "Value": {
      "Value": "1556"
    },
    "Revision": 1540
  },
  "want-event": {
    "Type": "put-operation",
    "Key": "key13",
    "Value": {
      "Value": "1004"
    },
    "Revision": 989
  }
```

### What did you expect to happen?

Cache should provide the same guarantees as etcd.

### How can we reproduce it (as minimally and precisely as possible)?

Pull the https://github.com/etcd-io/etcd/pull/20354

And run:
```
make gofail-enable
make 
make gofail-disable
GO_TEST_FLAGS='--count=1 -v --run TestRobustnessExploratory  --failfast' make test-robustness
```
/cc @MadhavJivrajani @apullo777 

/assign @apullo777 

### Etcd version (please run commands below)

<details>

```console
$ etcd --version
# paste output here

$ etcdctl version
# paste output here
```

</details>

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

[...]

PR #20491, opened 2025-08-14
cache: preserve resumable guarantee with empty history watchers

apullo777:

I open this PR to address the issue #20488

The root issue is that a cache can be ready but still have no history. So we need to ensure proper watcher and gap handling when history is empty. That includes:

- When history is empty, a watcher with a non-zero `startRev` should be registered as lagging. 
- `resyncLaggingWatchers` will not promote watcher to active while history is empty.
- `Broadcast` could defensively detect a gap (`firstRev` > `nextRev`) and move such watchers back to lagging instead of streaming past the gap.

/cc @serathius @MadhavJivrajani

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -27,7 +27,6 @@ import (
 )
 
 var (
-	// TODO: add gap-free replay for arbitrary startRevs and drop this guard.
 	// Returned when an option combination isn’t yet handled by the cache (e.g. WithPrevKV, WithProgressNotify for Watch(), WithCountOnly for Get()).
 	ErrUnsupportedRequest = errors.New("cache: unsupported request parameters")
 	// Returned when the requested key or key‑range is invalid (empty or reversed) or lies outside c.prefix.
--- a/cache/demux.go
+++ b/cache/demux.go
@@ -71,13 +71,17 @@ func (d *demux) Register(w *watcher, startingRev int64) {
 	defer d.mu.Unlock()
 
 	latestRev := d.history.PeekLatest()
+	if latestRev == 0 {
+		if startingRev == 0 {
+			d.activeWatchers[w] = 0
+		} else {
+			d.laggingWatchers[w] = startingRev
+		}
+		return
+	}
 
 	// Special case: 0 means “newest”.
 	if startingRev == 0 {
-		if latestRev == 0 {
-			d.activeWatchers[w] = 0
-			return
-		}
 		startingRev = latestRev + 1
 	}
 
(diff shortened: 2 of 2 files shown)


cache/cache.go
```diff
@@ -97,7 +96,16 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	startRev := op.Rev()
 
 	if startRev != 0 {
-		if oldest := c.demux.PeekOldest(); oldest != 0 && startRev < oldest {
+		oldestHistoryRev := c.demux.PeekOldest()
+		var tooOld bool
+		if oldestHistoryRev != 0 {
+			tooOld = startRev < oldestHistoryRev
+		} else {
+			storeFloor := c.store.LatestRev() + 1
```
serathius: Don't think watch should care about storeFloor.

apullo777: Yes. `storeFloor` is unrelated here.
If we want to decide compaction when history is empty, maybe I should move the compaction check into `demux.Register` and return `(compacted, compactRev)` so the decision and watcher registration can happen under the same lock:

```
func (d *demux) Register(w *watcher, startingRev int64) (compacted bool, compactRev int64) {
	d.mu.Lock()
	defer d.mu.Unlock()

	oldestRev := d.history.PeekOldest()
	if startingRev != 0 && oldestRev != 0 && startingRev < oldestRev {
		return true, oldestRev - 1
	}
	
	latestRev := d.history.PeekLatest()
	if latestRev == 0 {
	// watcher registration
	...
```

and in `Cache.Watch` we use that result to return canceled response:

```
	op := clientv3.OpWatch(key, opts...)
	startRev := op.Rev()

	pred, err := c.validateWatch(key, op)
    ...

	w := newWatcher(c.cfg.PerWatcherBufferSize, pred)
	compacted, compactRev := c.demux.Register(w, startRev)
	if compacted {
		ch := make(chan clientv3.WatchResponse, 1)
		ch <- clientv3.WatchResponse{
			Canceled:        true,
			CompactRevision: compactRev,
		}
		close(ch)
		return ch
	}
```

serathius: I think we should remove all this logic, and just register watch and depend on resync to kick it out. Reading demux state from under lock, can become very stale before the watch is started. Is a premature optimization that adds number of problematic edgecases.

apullo777: Removing the early check in `Cache.Watch` makes sense. For now `resyncLaggingWatchers` just silently stops watcher on compaction, but I think it should also return cancel response. However, that will require some work to give `watcher` a way to carry cancel/termination response so that the goroutine in `Cache.Watch` can forward the response before exiting. Maybe I should just revert this part and do the removal in a follow-up PR?

serathius: > I think it should also return cancel response

+1

> Maybe I should just revert this part and do the removal in a follow-up PR?

Sounds good

serathius (approved):

Issue #20396, opened 2025-07-24
Proposal: protection strategy during overload conditions


silentred:
### What would you like to be added?

x

### Why is this needed?

x

tjungblu:
reading through #18175, maybe we can simply lift the limit by default? 5000 seems very smallish for the amount of throughput etcd can handle nowadays, especially when writing many small key/values.

silentred:
> reading through [#18175](https://github.com/etcd-io/etcd/issues/18175), maybe we can simply lift the limit by default? 5000 seems very smallish for the amount of throughput etcd can handle nowadays, especially when writing many small key/values.

Thanks for your reply. It seems dificault to determine a proper value for all kinds of workload. A larger value could let etcd server have more time to consume the committed log before rejecting all incoming requests. However if etcd server cannot speed up applying, for example, APIServers are listing all objects, the server will eventually crash due to no chance to do LeaseRevoke.

silentred:
@ahrtr @serathius any suggestions?

ahrtr:
> 2\. Within the protection logic, reserve some queue space specifically for critical requests, so that essential system requests (`LeaseRevoke`, `Compact`) would have an chance to be executed under system pressure, preventing system crashes. Demo codes as below

It seems a valid point to me, also a low-hang fruit. `LeaseRevoke` should be fine, but not sure about `Compact`. Note that etcdserver can't process any other requests during compaction.

silentred:
> > 2. Within the protection logic, reserve some queue space specifically for critical requests, so that essential system requests (`LeaseRevoke`, `Compact`) would have an chance to be executed under system pressure, preventing system crashes. Demo codes as below
> 
> It seems a valid point to me, also a low-hang fruit. `LeaseRevoke` should be fine, but not sure about `Compact`. Note that etcdserver can't process any other requests during compaction.

Thanks @ahrtr . I am not sure about `Compact` either. Imaging etcd server missed serveral compactions, the outcomes may be 1. the db size quickly growing, which is not harmful
2. the treeIndex has many obsolete keyIndex, which may cuase OOM if the situation lasts for a long time.

silentred:
@ahrtr  Should I submit a PR for this? If so, I guess it should be put in the FeatureGate.
And what do you think about configurable `maxGapBetweenApplyAndCommitIndex`?

ahrtr:
Pls bring both topic to our community meeting if possible.

[...]

PR #20492, opened 2025-08-15
stability enhancement during overload conditions

silentred:
The PR is for the proposal #20396 
Based on the previous discussion, we agreed that `LeaseRevoke` should have higher priority, and we remain cautious about `Compact`. Therefore, I've prepared a table comparing the potential impacts of applying and not-applying `Compact`.

| Request | Effect of Apply | Effect of No Apply | Is Critical | Typical Scenario |
|--------|--------|--------|--------|--------|
| LeaseRevoke | Cleaning a few keys, which is supposed to happen | Keys cannot stop growing, maybe finally trigger a crash of server. It takes very long time to reboot because of large db size. Upstream watcher may be oom-killed due to writing too many resources to cache. | YES | K8S apiserver recording events |
| Compact | Clean obsolete KVs in index and db, but make the applying slower | treeIndex and db size growing, may reach disk quota. Requires SREs to recover the service. | TBD | default periodical compaction  | 

I would like to have some advices that should I add a unit test or an e2e test? I am not quite sure how this feature should be tested.

### DIFF
--- a/server/etcdserver/util.go
+++ b/server/etcdserver/util.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"time"
 
+	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/client/pkg/v3/types"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/membership"
 	"go.etcd.io/etcd/server/v3/etcdserver/api/rafthttp"
@@ -40,6 +41,26 @@ func isConnectedSince(transport rafthttp.Transporter, since time.Time, remote ty
 // members in the cluster since the given time.
 func isConnectedFullySince(transport rafthttp.Transporter, since time.Time, self types.ID, members []*membership.Member) bool {
 	return numConnectedSince(transport, since, self, members) == len(members)
+}
+
+// isTooManyRequest checks whether the gap between applied index and committed
+// index is too large. r *pb.InternalRaftRequest should read-only.
+func isTooManyRequest(ai, ci uint64, r *pb.InternalRaftRequest) bool {
+	// Critical request kinds include LeaseRevoke
+	isCritical := r != nil && r.LeaseRevoke != nil
+
+	// If the gap is larger than maxGapBetweenApplyAndCommitIndex, reject all normal request
+	if !isCritical && ci > ai+maxGapBetweenApplyAndCommitIndex {
+		return true
+	}
+
+	// Make an extra 10% gap buffer for critical requests, so thay can be applied
+	// during overload conditions.
+	if isCritical && ci-ai > maxGapBetweenApplyAndCommitIndex+500 {
+		return true
+	}
+
+	return false
 }
 
 // numConnectedSince counts how many members are connected to the local member
--- a/server/etcdserver/v3_server.go
+++ b/server/etcdserver/v3_server.go
@@ -792,7 +792,8 @@ func (s *EtcdServer) doSerialize(ctx context.Context, chk func(*auth.AuthInfo) e
 func (s *EtcdServer) processInternalRaftRequestOnce(ctx context.Context, r pb.InternalRaftRequest) (*apply2.Result, error) {
 	ai := s.getAppliedIndex()
 	ci := s.getCommittedIndex()
-	if ci > ai+maxGapBetweenApplyAndCommitIndex {
+
+	if isTooManyRequest(ai, ci, &r) {
 		return nil, errors.ErrTooManyRequests
 	}
 


silentred:
@ahrtr would you mind to take a look?

server/etcdserver/v3_server.go
```diff
@@ -792,7 +792,8 @@ func (s *EtcdServer) doSerialize(ctx context.Context, chk func(*auth.AuthInfo) e
 func (s *EtcdServer) processInternalRaftRequestOnce(ctx context.Context, r pb.InternalRaftRequest) (*apply2.Result, error) {
 	ai := s.getAppliedIndex()
 	ci := s.getCommittedIndex()
-	if ci > ai+maxGapBetweenApplyAndCommitIndex {
+
+	if isTooManyRequest(ai, ci, &r) {
```
serathius: Can you add unit tests?

server/etcdserver/util_test.go
```diff
@@ -110,3 +112,54 @@ type testStringerFunc func() string
...
 	return s()
 }
+
+func TestIsTooManyRequest(t *testing.T) {
+	tests := []struct {
+		ci     uint64
+		ai     uint64
+		expect bool
+		req    *pb.InternalRaftRequest
+		desc   string
```
ahrtr:
```suggestion
		name string
		ci     uint64
		ai     uint64
		expectedResult bool
		req    *pb.InternalRaftRequest
```

server/etcdserver/util_test.go
```diff
@@ -110,3 +112,54 @@ type testStringerFunc func() string
...
+			expect: true,
+			req:    &pb.InternalRaftRequest{LeaseRevoke: &pb.LeaseRevokeRequest{}},
+			desc:   "Test critical request and gap is larger than 110% maxGapBetweenApplyAndCommitIndex",
+		},
+	}
+
+	for _, test := range tests {
+		t.Log(test.desc)
+		assert.Equal(t, test.expect, isTooManyRequest(test.ai, test.ci, test.req))
+	}
```
ahrtr: pls use golang sub test
```suggestion
	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			assert.Equal(t, test.expect, isTooManyRequest(test.ai, test.ci, test.req))
		}
	}
```

server/etcdserver/util.go
```diff
@@ -42,6 +43,26 @@ func isConnectedFullySince(transport rafthttp.Transporter, since time.Time, self
...
+	}
+
+	// Make an extra 10% gap buffer for critical requests, so thay can be applied
+	// during overload conditions.
+	if isCritical && ci-ai > maxGapBetweenApplyAndCommitIndex+500 {
+		return true
+	}
+
+	return false
+}
```
ahrtr:
```suggestion
// exceedsRequestLimit checks if the committed index is too far ahead of
// the applied index. Priority requests are allowed up to 10% extra gap.
func exceedsRequestLimit(ai, ci uint64, r *pb.InternalRaftRequest) bool {
	if ci <= ai+maxGapBetweenApplyAndCommitIndex {
		return false
	}

	// allow up to 10% extra gap for priority requests
	if ci <= ai+maxGapBetweenApplyAndCommitIndex*110/100 {
		if isPriorityRequest(r) {
			return false
		}
	}

	return true
}

func isPriorityRequest(r *pb.InternalRaftRequest) bool {
	return r != nil && r.LeaseRevoke != nil
}

```

ahrtr:
Did this actually happen in production or in dev? Have you tested the patch and seen the expected improvement?

silentred:
> Did this actually happen in production or in dev? Have you tested the patch and seen the expected improvement?

It happened several times in production. One of cases is that K8S event keys growed to 50Mi, because leases cannot be revoked successfully. This patch has not been deployed to our production yet.
I've made a simulation of slow applying with follwing changes. https://github.com/etcd-io/etcd/commit/dcc092cc9939b39ee1b593aa726a2d1a908f0210
```
# keep sending PUT request with lease
bin/benchmark put --endpoints=http://xxx:3379 --clients=200 --conns=50 --rate=2000 --total=10000000 --key-space-size=10000000 --lease-reuse
```
This results in leases living longer than it should be.
image

If we patch this PR, then everything works fine.

ahrtr:
thx for the feedback, please resolve the review comments

serathius:
> One of cases is that K8S event keys growed to 50Mi, because leases cannot be revoked successfully. This patch has not been deployed to our production yet.

Oh, that's a lot. I have never seen etcd cluster be stable after crossing 2M. Have you verified that this change really would fix your production issue? It might move the needle, but we are talking here about 25x improvement.

silentred:
> Oh, that's a lot. I have never seen etcd cluster be stable after crossing 2M. 

etcdserver was not working properly in that situation. The worst case we've been through is that we have to clear all data and re-setup a new empty cluster for K8S event.

> Have you verified that this change really would fix your production issue?

This patch has not been deployed to production, it is not a common issue, maybe once half a year. But I think my simulation test above could verify this patch works.

serathius:
> The worst case we've been through is that we have to clear all data and re-setup a new empty cluster for K8S event.

That's recommended practice. Her my old talk about this issue https://www.youtube.com/watch?v=aJVMWcVZOPQ&t=4m9s

> But I think my simulation test above could verify this patch works.

What test?

silentred:
> > The worst case we've been through is that we have to clear all data and re-setup a new empty cluster for K8S event.
> 
> That's recommended practice. Her my old talk about this issue https://www.youtube.com/watch?v=aJVMWcVZOPQ&t=4m9s
> 
> > But I think my simulation test above could verify this patch works.
> 
> What test?

Thanks, I will check it.
Please take a look at this comment https://github.com/etcd-io/etcd/pull/20492#issuecomment-3216807625

ahrtr:
The PR looks good.

The only comment is that there isn't a real performance results comparison.

silentred:
> The PR looks good.
> 
> The only comment is that there isn't a real performance results comparison.

I am planning to add a FeatureGate for this. Also, I will make a benchmark comparison with master branch and post it here. Thanks.

serathius:
Did some basic testing and I'm not sure this addresses the main problem. Simple scenario, grant leases with TTL of 5 seconds for 1 minute and see if there are properly revoked. The results:
* Above 100 QPS of grant leases the etcd is unable to keep up with revoking the leases. 
* Only after passing 1000 QPS of grant leases returning `too many requests` errors.
* Prioritizing LeaseRevokes improves the revoke rate 20-40% but only after passing 1000 qps

While I think prioritizing LeaseGrant help somewhat, I'm not sure it's a worth doing a special case that doesn't address the underlying issue. Lease revokation rate is too slow.

| Scenario                  | QPS Sent | Leases revoked | Improvement | Max active leases | Handled QPS | Leases granted | `too many requests` count |
| ------------------------- | -------- | -------------- | ----------- | ----------------- | ----------- | -------------- | ------ |
| Base                      | 2000     | 18.22%         |             | 95687             | 1703.3747   | 106184         | 6499   |
| Base                      | 1000     | 24.84%         |             | 46674             | 887.155     | 55902          | 1980   |
| Base                      | 500      | 41.43%         |             | 20072             | 465.3443    | 30000          | 0      |
| Base                      | 250      | 73.63%         |             | 5205              | 241.6778    | 15000          | 0      |
| Base                      | 100      | 99.15%         |             | 551               | 100.0116    | 6000           | 0      |
| WithPrioritizeLeaseRevoke | 2000     | 25.50%         | 40%         | 54945             | 990.5109    | 66328          | 48786  |
| WithPrioritizeLeaseRevoke | 1000     | 29.79%         | 20%         | 38117             | 737.9423    | 48527          | 9848   |
| WithPrioritizeLeaseRevoke | 500      | 41.58%         | 0%          | 20025             | 462.8422    | 30000          | 0      |
| WithPrioritizeLeaseRevoke | 250      | 73.45%         | 0%          | 5233              | 241.5084    | 15000          | 0      |
| WithPrioritizeLeaseRevoke | 100      | 99.17%         | 0%          | 550               | 100.0115    | 6000           | 0   |

Code 
```
func leaseGrantFunc(cmd *cobra.Command, _ []string) {
	clients := mustCreateClients(totalClients, totalConns)

	r := newReport(cmd.Name())
	qps := 100
	takeN := int(math.Ceil(float64(qps) / 500))
	limiter := rate.NewLimiter(rate.Limit(qps), takeN)

	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
	defer cancel()
	for i := range totalClients {
		c := clients[i]
		wg.Go(func() {
			for {
				select {
				case <-ctx.Done():
					return
				default:
				}
				err := limiter.WaitN(ctx, takeN)
				if err != nil {
					return
				}
				wg.Go(func() {
					for range takeN {
						select {
						case <-ctx.Done():
							return
						default:
						}
						start := time.Now()
						_, err = c.Grant(context.Background(), 5)
						end := time.Now()
						if err != nil {
							fmt.Printf("%v\n", err)
						}
						r.Results() <- report.Result{Start: start, End: end, Err: err}
					}
				})
			}
		})
	}

	rc := r.Run()
	wg.Wait()
	close(r.Results())
	fmt.Printf("%s", <-rc)
}

```

silentred:
Hi @serathius , Thanks for your feedback. I think you are addressing a different problem, and your test is proving this PR having positive effects in some way. Let me explain.

Q: Why `Above 100 QPS of grant leases the etcd is unable to keep up with revoking the leases.`
A: etcdserver has parallelization control of expired lease revocations by `maxPendingRevokes`=16. 100 QPS of `LeaseRevoke` is enough for lease creation in most use cases. K8S reuses a lease for all events created in one minute, so its rate is 1/min. Any usage of above 100 QPS of `LeaseRevok` should be considered as a misuse of etcdserver from my perspective. 


Q: Why there is no "too many request" under 500 QPS ?
A: Because `maxGapBetweenApplyAndCommitIndex` is playing its role. `Lease` requests are waiting to be applied, meanwhile the un-applied logs length are not reaching `maxGapBetweenApplyAndCommitIndex`. 


Q: Why `but only after passing 1000 qps` ?
A: Because this PR only aims to ease the problem of flooding "too many request" errors reducing the probability of `LeaseRevoke` to be applied. "too many request" error is a necessary condition. 


Q: Why this is a different problem?
A: You mentioned the problem is "Lease revokation rate is too slow", when `LesaeGrant` above 100 QPS, which in my case is 1 QPM. And in my case, the problem is flooding Txn causing "too many request" error, causing drops of `LeaseRevoke` due to `maxGapBetweenApplyAndCommitIndex`.

serathius:
@silentred Makes sense. Could you propose a benchmark that would better match the setup you are targeting? I think Kubernetes is allocating up to 5k keys per lease. Would a mix of lease and TXN requests in that proportion work?

silentred:
I have some new findings by benchmark test. I added following code before `if exceedsRequestLimit(ai, ci, &r, true) `
```
func (s *EtcdServer) processInternalRaftRequestOnce(ctx context.Context, r pb.InternalRaftRequest) (*applyResult, error) {
// to check the value of "ci - ai"
if isPriorityRequest(&r) {
  s.lg.Info("call revoke lease",
    zap.Uint64("ci - ai", ci-ai),
    zap.Bool("exceed", exceedsRequestLimit(ai, ci, &r, s.FeatureEnabled(features.PriorityRequest))),
  )
}
// ...
}
```
the value of `ci - ai` could be bigger than 10000 under high QPS condition, and it rarely be in range of (5000, 5500]. So the factor in `ci <= ai+maxGapBetweenApplyAndCommitIndex*110/100` matters a lot. The following benchmark simulates k8s event workload with lease TTL=15s, reuse duration=1s, which means the test allocates one lease per second. 

benchmark PR: https://github.com/etcd-io/etcd/pull/20819

The result compares LeareRevoke results under different QPS limit in 1min.
- Allocated Lease: the total number of leases allocated in 1 min.
- Left Lease: the number of leases in the end of the test, fetched by LeaseLeases API.
- Invalid Lease: the number of leases, of which TTL is less than 0, fetched by LesaeTimeToLive API.
- Revoke Failure: the number of lines in the etcdserver log, including "failed to revoke lease", which is logged if LeaseRevoke failed.

Less `Invalid Lease` is better. The result shows obvious improvement when factor=2.0 .

| Branch      | Priority Gate  | Factor | QPS Limit | Handled QPS | Allocated Lease | Left Lease | Invalid Lease | Revoke Failure |
|-------------|----------------|--------|-----------|-------------|-----------------|------------|---------------|----------------|
| master      | off            |        | 7500      | 7397        | 60              | 15         | 0             | 0              |
|             | off            |        | 8500      | 6849        | 60              | 18         | 3             | 3              |
|             | off            |        | 10000     | 5393        | 52              | 15         | 4             | 10             |
|             | off            |        | 12000     | 5229        | 45              | 20         | 10            | 14             |
| master      | on             | 1.1    | 7500      | 7345        | 60              | 15         | 0             | 0              |
|             | on             | 1.1    | 8500      | 6778        | 60              | 15         | 0             | 0              |
|             | on             | 1.1    | 10000     | 6239        | 58              | 21         | 7             | 7              |
|             | on             | 1.1    | 12000     | 5725        | 49              | 21         | 13            | 19             |
| master      | on             | 2.0    | 8500      | 5940        | 58              | 15         | 0             | 0              |
|             | on             | 2.0    | 10000     | 6246        | 59              | 15         | 0             | 0              |
|             | on             | 2.0    | 12000     | 5977        | 52              | 14         | 4             | 4              |
| release-3.4 | on             | 1.1    | 10000     | 2954        | 37              | 16         | 12            | 12             |
|             | on             | 2.0    | 10000     | 2991        | 38              | 15         | 2             | 2              |

serathius:
Please send PR to add a benchmark so we can review it

server/etcdserver/util.go
```diff
@@ -41,6 +42,25 @@ func isConnectedFullySince(transport rafthttp.Transporter, since time.Time, self
 	return numConnectedSince(transport, since, self, members) == len(members)
 }
 
+// exceedsRequestLimit checks if the committed index is too far ahead of
+// the applied index. Priority requests are allowed up to 10% extra gap.
```
serathius: This comment about 10% is inconsistent with one 5 lines below. Recommend to remove the comments and not comment on values of const values. Overall rule: Don't comment on "what?" the code is doing, that should be explained by code, comment on "why?" things are done certain way. For example why we need this feature. Comment explaining that we prioritize leaseRevoke because we don't want to leak leases.

silentred: Thanks a lot. Comment updated.

server/etcdserver/util.go
```diff
@@ -41,6 +42,26 @@ func isConnectedFullySince(transport rafthttp.Transporter, since time.Time, self
 	return numConnectedSince(transport, since, self, members) == len(members)
 }
 
+// exceedsRequestLimit checks if the committed index is too far ahead of the applied index.
+// LeaseRevoke requests are prioritized to ensure timely lease expiration,
+// which helps mitigate pressure on the cluster.
+func exceedsRequestLimit(ai, ci uint64, r *pb.InternalRaftRequest, enablePriority bool) bool {
```
serathius: Please use proper name for `ai` and `ci` instead of 2 letter names. Previously the name didn't need to be short because it was used across 3 lines of code, 2 of which was reading using server methods that mentioned full name. As argument names they are confusing.

server/etcdserver/util.go
```diff
@@ -41,6 +42,26 @@ func isConnectedFullySince(transport rafthttp.Transporter, since time.Time, self
...
+	if ci <= ai+maxGapBetweenApplyAndCommitIndex {
+		return false
+	}
+	// allow up to 100% extra gap for priority requests
```
serathius: Mentioned before to remove such comments.

server/etcdserver/util.go
```diff
@@ -41,6 +42,26 @@ func isConnectedFullySince(transport rafthttp.Transporter, since time.Time, self
...
+	if enablePriority && isPriorityRequest(r) {
+		if ci <= ai+maxGapBetweenApplyAndCommitIndex*200/100 {
```
serathius: Now we want 2 const variables:
* `maxNormalGapBetweenApplyAndCommitIndex`
* `maxPriorityGapBetweenApplyAndCommitIndex = 2*maxNormalGapBetweenApplyAndCommitIndex`

Names could be shortened to just `normalGap` and `maxPriorityGap`.

serathius:
cc @fuweid @ahrtr

server/features/etcd_features.go
```diff
@@ -79,6 +79,11 @@ const (
 	// beta: v3.7
 	// main PR: https://github.com/etcd-io/etcd/pull/20589
 	FastLeaseKeepAlive featuregate.Feature = "FastLeaseKeepAlive"
+	// PriorityRequest enables certain kinds of requests(e.g. LreaeRevoke) having higher priority to be applied under overload conditions.
```
ahrtr:
```suggestion
	// PriorityRequest enables certain kinds of requests(e.g. LeaseRevoke) having higher priority to be applied under overload conditions.
```

ahrtr:
can you also update the 3.7 changelog ?  It can be in this PR or in a separate PR? thx

https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.7.md#etcd-server

PR #20494, opened 2025-08-15
Remove checking `confState` during schema migration

ahrtr:
The field `confState` was added since etcd v3.5. If an etcd cluster was created since v3.4, then it wouldn't have this field. In 3.5+, the field will only be automatically populated during conf change (i.e. add/remove/update member). So when upgrading from 3.4 to 3.5, and from 3.5 to 3.6, the field will never be populated unless users add or remove or update members.

Note if you can't reproduce this issue, it should be because that there is no v2 snapshot, and etcdserver will replay all WAL records (including add member records) on bootstrap.

See also https://github.com/etcd-io/etcd/discussions/20446#discussioncomment-14114666

### DIFF
--- a/server/storage/schema/schema.go
+++ b/server/storage/schema/schema.go
@@ -95,15 +95,11 @@ func UnsafeDetectSchemaVersion(lg *zap.Logger, tx backend.UnsafeReader) (v semve
 		return *vp, nil
 	}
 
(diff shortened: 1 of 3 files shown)


ahrtr:
Fix https://github.com/etcd-io/etcd/issues/20495

ahrtr:

PR #20499, opened 2025-08-15
cache: remove early compaction check from Cache.Watch

apullo777:

To address the issue #20488
#20491 follow-up: we want to:

- remove early compaction check from `Cache.Watch` (therefore `demux.PeekOldest` is not needed)
- rely on `demux.resyncLaggingWatchers` to handle compacted watchers
- make `watcher` carry cancel response so that `Cache.Watch` goroutine can forward it

@serathius @MadhavJivrajani

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	pb "go.etcd.io/etcd/api/v3/etcdserverpb"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
@@ -95,18 +96,6 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 	op := clientv3.OpWatch(key, opts...)
 	startRev := op.Rev()
 
-	if startRev != 0 {
-		if oldest := c.demux.PeekOldest(); oldest != 0 && startRev < oldest {
-			ch := make(chan clientv3.WatchResponse, 1)
-			ch <- clientv3.WatchResponse{
-				Canceled:        true,
-				CompactRevision: startRev,
-			}
-			close(ch)
-			return ch
-		}
-	}
-
 	pred, err := c.validateWatch(key, op)
 	if err != nil {
 		ch := make(chan clientv3.WatchResponse, 1)
@@ -132,6 +121,13 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
 				return
 			case events, ok := <-w.eventQueue:
 				if !ok {
+					if w.cancelResp != nil {
+						select {
+						case <-ctx.Done():
+						case <-c.internalCtx.Done():
+						case responseChan <- *w.cancelResp:
+						}
+					}
 					return
 				}
 				select {
@@ -299,7 +295,11 @@ func (c *Cache) watchEvents(watchCh clientv3.WatchChan, applyErr <-chan error, r
 			readyOnce.Do(func() { c.ready.Set() })
 			if err := resp.Err(); err != nil {
 				c.ready.Reset()
-				c.demux.Purge()
+				if errors.Is(err, rpctypes.ErrCompacted) || resp.CompactRevision > 0 {
+					c.demux.Compact(resp.CompactRevision)
+				} else {
+					c.demux.Purge()
+				}
 				return err
 			}
 			c.demux.Broadcast(resp.Events)
--- a/cache/demux.go
+++ b/cache/demux.go
@@ -151,7 +151,7 @@ func (d *demux) Broadcast(events []*clientv3.Event) {
 	}
 }
 
-// Purge is called when etcd compaction invalidates our cached history, so clients should resubscribe.
+// Purge stops all watchers and rebase history on watch errors
 func (d *demux) Purge() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
@@ -166,6 +166,21 @@ func (d *demux) Purge() {
 	d.laggingWatchers = make(map[*watcher]int64)
 }
 
+// Compact is called when etcd reports a compaction at compactRev to rebase history;
+// it keeps provably-too-old watchers for later cancellation, stops others, and clients should resubscribe.
+func (d *demux) Compact(compactRev int64) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.history.RebaseHistory()
+
+	for w, next := range d.activeWatchers {
+		if next != 0 && next <= compactRev {
+			delete(d.activeWatchers, w)
+			d.laggingWatchers[w] = next
+		}
+	}
+}
+
 func (d *demux) resyncLaggingWatchers() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
@@ -177,7 +192,7 @@ func (d *demux) resyncLaggingWatchers() {
 
 	for w, nextRev := range d.laggingWatchers {
 		if nextRev < oldestRev {
-			w.Stop()
+			w.Compact(nextRev)
 			delete(d.laggingWatchers, w)
 			continue
 		}
@@ -199,10 +214,4 @@ func (d *demux) resyncLaggingWatchers() {
 			d.laggingWatchers[w] = nextRev
 		}
 	}
-}
-
-func (d *demux) PeekOldest() int64 {
-	d.mu.RLock()
-	defer d.mu.RUnlock()
-	return d.history.PeekOldest()
 }
--- a/cache/watcher.go
+++ b/cache/watcher.go
@@ -15,24 +15,24 @@
 package cache
 
 import (
-	"sync/atomic"
+	"sync"
 
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
 // watcher holds one client’s buffered stream of events.
 type watcher struct {
 	eventQueue chan []*clientv3.Event
+	cancelResp *clientv3.WatchResponse
 	keyPred    KeyPredicate
-	stopped    int32
-	done       chan struct{} // closed together with Stop()
+	stopOnce   sync.Once
 }
 
 func newWatcher(bufSize int, pred KeyPredicate) *watcher {
 	return &watcher{
 		eventQueue: make(chan []*clientv3.Event, bufSize),
 		keyPred:    pred,
-		done:       make(chan struct{}),
 	}
 }
 
@@ -59,12 +59,21 @@ func (w *watcher) enqueueEvent(eventBatch []*clientv3.Event) bool {
 	}
 }
 
-// Stop closes the event channel atomically.
-func (w *watcher) Stop() {
-	if atomic.CompareAndSwapInt32(&w.stopped, 0, 1) {
-		close(w.eventQueue)
-		close(w.done)
+func (w *watcher) Compact(compactRev int64) {
+	resp := &clientv3.WatchResponse{
+		Canceled:        true,
+		CompactRevision: compactRev,
+		CancelReason:    rpctypes.ErrCompacted.Error(),
 	}
+	w.stopOnce.Do(func() {
+		w.cancelResp = resp
+		close(w.eventQueue)
+	})
 }
 
-func (w *watcher) Done() <-chan struct{} { return w.done }
+// Stop closes the event channel atomically.
+func (w *watcher) Stop() {
+	w.stopOnce.Do(func() {
+		close(w.eventQueue)
+	})
+}


cache/watcher.go
```diff
@@ -26,6 +28,8 @@ type watcher struct {
 	keyPred    KeyPredicate
 	stopped    int32
 	done       chan struct{} // closed together with Stop()
+	cancelMu   sync.RWMutex
+	cancelResp *clientv3.WatchResponse
```
serathius, line 27: Why keep it here, maybe change eventQueue to response channel and write response there?

apullo777:
I tried with replacing `eventQueue` with `respQueue`, and keep a lock for `watcher` to synchronize `notifyCompacted` and `Stop`. 

But was a bit struggling with dealing with a scenario where if a client requests a too-old revision and registers as a lagger at the exact moment we detect an upstream compaction, seems like it will be non-deterministic for the already-registered laggers. If `Purge` wins, the lagging watcher that `resync` intended to cancel will just be silently closed by `w.Stop()` before `notifyCompacted` can run. Maybe I should try to find a way to solve this without further complicating the code

cache/demux.go
```diff
@@ -157,16 +157,24 @@ func (d *demux) Purge() {
 
 func (d *demux) resyncLaggingWatchers() {
 	d.mu.Lock()
-	defer d.mu.Unlock()
```
MadhavJivrajani, line 171: Any reason to remove defer here?

cache/watcher.go
```diff
@@ -52,17 +55,35 @@ func (w *watcher) enqueueEvent(eventBatch []*clientv3.Event) bool {
...
+	defer w.mu.Unlock()
+	if atomic.LoadInt32(&w.stopped) != 0 {
+		return
+	}
+	w.respQueue <- clientv3.WatchResponse{
+		Canceled:        true,
+		CompactRevision: compactRev,
+		CancelReason:    rpctypes.ErrCompacted.Error(),
+	}
+	atomic.StoreInt32(&w.stopped, 1)
```
MadhavJivrajani: Something to maybe cleanup in the future (not needed in this PR): we don't need atomics when we're reading and writing to `w.stopped` under a lock anyway.

MadhavJivrajani:
> I tried with replacing eventQueue with respQueue, and keep a lock for watcher to synchronize notifyCompacted and Stop.

This still might not be enough since you are calling `notifyCompacted` in a loop - lock will get released between iterations.

MadhavJivrajani:
> But was a bit struggling with dealing with a scenario where if a client requests a too-old revision and registers as a lagger at the exact moment we detect an upstream compaction, seems like it will be non-deterministic for the already-registered laggers. If Purge wins, the lagging watcher that resync intended to cancel will just be silently closed by w.Stop() before notifyCompacted can run. Maybe I should try to find a way to solve this without further complicating the code

Thinking about this a bit more - it might help to make `Purge()` a little more aware of the reason it is being called.
Currently, we purge on the following situations:
1. Error in watch response
2. Error applying events to the store (internal error)

We can try interpreting the error in `Purge()` and instead of calling `Stop()`, send a compaction notification if needed with the required information depending on the error occured. This way either Purge executes sending out the appropriate notification if needed, or the resync loop can proceed with kicking out watchers as needed and sending compaction notifications. 

wdyt?

apullo777:
> it might help to make `Purge()` a little more aware of the reason it is being called. Currently, we purge on the following situations:
> 
> 1. Error in watch response
> 2. Error applying events to the store (internal error)
> 
> We can try interpreting the error in `Purge()` and instead of calling `Stop()`, send a compaction notification if needed with the required information depending on the error occured. This way either Purge executes sending out the appropriate notification if needed, or the resync loop can proceed with kicking out watchers as needed and sending compaction notifications.

Thanks! I think that makes sense.

Maybe we can have a `PurgeWithCompaction` which distinguishes upstream compaction (via the error) and sends compaction cancels to the affected watchers (whose `next <= compactRev`) first and only afterward stop the rest.

The resync path mirrors this by compact notifying laggers that fell behind the retained window. 

So even If both happen, the shared lock + one shot classification makes sure the lagging watcher with too-old rev is decided only once and always receives the compaction cancel before any `Stop`.

cache/demux.go
apullo777: I want to unlock `d.mu` before calling `notifyCompacted` so we do not hold the demux lock while doing a blocking send to a watcher’s channel

cache/demux.go
```diff
@@ -141,32 +141,63 @@ func (d *demux) Broadcast(events []*clientv3.Event) {
...
 	d.laggingWatchers = make(map[*watcher]int64)
+
+	d.mu.Unlock()
+
+	for _, w := range compactList {
+		w.notifyCompacted(compactRev)
+	}
+	for _, w := range stopList {
+		w.Stop()
+	}
```
serathius: Don't like the idea of doing it outside of lock. This makes it easy to introduce races in code. Like this code racing against resync. 

Have you considered separating identifying which watchers to compact notifying them later? We could send compaction in `resyncLaggingWatchers`, this way all the logic is executed by one loop.

cache/demux.go
```diff
@@ -188,10 +219,8 @@ func (d *demux) resyncLaggingWatchers() {
...
 	}
-}
-
-func (d *demux) PeekOldest() int64 {
-	d.mu.RLock()
-	defer d.mu.RUnlock()
-	return d.history.PeekOldest()
+	d.mu.Unlock()
+	for _, c := range compacted {
+		c.w.notifyCompacted(c.rev)
```
serathius: This is a blocking operation. While moving it outside of lock helps unlock other operation, but means that resync loop would be stuck deadlocked if any watch would not read response.

apullo777: Maybe in `resyncLaggingWatchers` we should try a non-blocking `notifyCompacted`. If the watcher’s `respQueue` cannot read the cancel, we immediately `Stop()` it instead of blocking the resync loop. But that also means the channel is just closed silently and the client will not see a final cancel response

cache/demux.go
```diff
@@ -152,18 +152,33 @@ func (d *demux) Broadcast(events []*clientv3.Event) {
...
+		} else {
+			w.Stop()
+			delete(d.activeWatchers, w)
+		}
 	}
-	for w := range d.laggingWatchers {
-		w.Stop()
+	for w, next := range d.laggingWatchers {
+		if isCompaction && next != 0 && next <= compactRev {
+			d.laggingWatchers[w] = next
```
serathius: Don't think this is needed

cache/demux.go
```diff
@@ -152,18 +152,33 @@ func (d *demux) Broadcast(events []*clientv3.Event) {
 }
 
 // Purge is called when etcd compaction invalidates our cached history, so clients should resubscribe.
-func (d *demux) Purge() {
+func (d *demux) Purge() { d.purge(false, 0) }
+
+// PurgeWithCompaction sends CompactRevision cancels to watchers with too-old next revisions, then stops others to avoid a resync race.
+func (d *demux) PurgeWithCompaction(compactRev int64) { d.purge(true, compactRev) }
```
serathius:
```suggestion
func (d *demux) Compact(compactRev int64) { d.purge(true, compactRev) }
```

cache/demux.go
```diff
@@ -152,18 +152,33 @@ func (d *demux) Broadcast(events []*clientv3.Event) {
...
+
+func (d *demux) purge(isCompaction bool, compactRev int64) {
```
serathius: Don't think I like the idea of reusing `purge` function for purging all watchers and compacting watchers. Can you just implement it as separate functions?

cache/demux.go
```diff
@@ -177,7 +192,9 @@ func (d *demux) resyncLaggingWatchers() {
 
 	for w, nextRev := range d.laggingWatchers {
 		if nextRev < oldestRev {
-			w.Stop()
+			if !w.notifyCompacted(nextRev) {
+				w.Stop()
```
serathius:
```suggestion
				continue
```

cache/watcher.go
```diff
@@ -15,30 +15,33 @@
...
 type watcher struct {
-	eventQueue chan []*clientv3.Event
-	keyPred    KeyPredicate
-	stopped    int32
-	done       chan struct{} // closed together with Stop()
+	respQueue chan clientv3.WatchResponse
+	keyPred   KeyPredicate
+	stopped   int32
+	done      chan struct{} // closed together with Stop()
+	mu        sync.Mutex
```
serathius: What does mux protect?

Please move mutex to separate section of variables and put variables protected by mutext under mutex.

cache/watcher.go
```diff
@@ -61,8 +87,10 @@ func (w *watcher) enqueueEvent(eventBatch []*clientv3.Event) bool {
 
 // Stop closes the event channel atomically.
 func (w *watcher) Stop() {
+	w.mu.Lock()
+	defer w.mu.Unlock()
 	if atomic.CompareAndSwapInt32(&w.stopped, 0, 1) {
```
serathius: If you are using mutex, no need to use atomic.

cache/watcher.go
```diff
@@ -15,30 +15,33 @@
...
 
 // watcher holds one client’s buffered stream of events.
 type watcher struct {
-	eventQueue chan []*clientv3.Event
-	keyPred    KeyPredicate
-	stopped    int32
-	done       chan struct{} // closed together with Stop()
+	respQueue chan clientv3.WatchResponse
+	keyPred   KeyPredicate
+	stopped   int32
```
serathius: Reading the code again I'm missing why do we even need separate `stopped` variable and separate done channel.

cache/demux.go
serathius: Don't think we should just close if we cannot inform about being compacted. We can retry in next iteration.

cache/watcher.go
serathius: `Done` is not called by anyone so we don't need a separate `done` channel.
I don't see any reason we would call `Close` twice, so I don't know if we need `stopped`.

cache/demux.go
apullo777: Thanks! Just realized that I should do something like this

```
	for w, nextRev := range d.laggingWatchers {
		if nextRev < oldestRev {
			if !w.notifyCompacted(nextRev) {
				continue
			}
			delete(d.laggingWatchers, w)
			continue
		}
```

cache/watcher.go
apullo777: Maybe we still need `stopped` because both `Stop() `(via `demux.Purge`) and `notifyCompacted()` (via resync) can try to close the same `respQueue`?

Two options in mind: 

1. keep a `stopped` bool on watcher and set/close it only while holding `d.mu`
2. make demux fully own closing: `notifyCompacted` returns a cancel flag and demux under `d.mu` removes the watcher and close the channel

serathius: Make sense. Option 1 sounds good, don't understand option 2

serathius:
Please note `TestCacheWatchOldRevisionCompacted` still flakes on this PR.

cache/demux.go
apullo777: maybe just to keep those too old watchers for later cancellation and stop other?

```
	for w, next := range d.laggingWatchers {
		if next != 0 && next > compactRev {
			continue
		} else {
			w.Stop()
			delete(d.laggingWatchers, w)
		}
```

apullo777:
> Please note `TestCacheWatchOldRevisionCompacted` still flakes on this PR.

Now I am a bit more confident that the flake is rooted in resync's `notifyCompacted` non-blocking send being unreliable for this PR.

Experimenting with `ResyncInterval=1h` (so `resyncLoop` doesn't run in `TestCacheWatchOldRevisionCompacted`), the old code (post-#20483) almost never fails thanks to its synchronous early check in `Watch`; the new code fails 100% because we fully rely on resync’s non-blocking `notifyCompacted` send.

I update the PR and attempt to fix this by separating data from control. Now each watcher has two channels: 
1. `eventsCh`  for normal events, configurable size
2. `cancelCh` that the forwarder always prioritizes, size=1  

Cancel is only handled in resync and sent on `cancelCh` while holding the demux lock and the watcher is removed, so the cancel can't be blocked by `eventsCh` and will always go through. And since now we do not double close the same channel, we no longer need `stopped` and `done` but perhaps need a  `stopOnce sync.Once`.

I also temporarily added `TestCacheWatchOldRevisionCompacted` back on this PR so that we can test against it. Did stress on different setting and no failures.

cache/demux.go
```diff
@@ -151,7 +151,7 @@ func (d *demux) Broadcast(events []*clientv3.Event) {
 	}
 }
 
-// Purge is called when etcd compaction invalidates our cached history, so clients should resubscribe.
```
apullo777, line 154: Do we also want to send some response to stopped watchers to notify them they need to resubscribe after upstream watch compaction/error?

serathius: ? Where this concept came from? There is no such mechanism in etcd.

cache/demux.go
```diff
@@ -166,6 +166,31 @@ func (d *demux) Purge() {
...
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.history.RebaseHistory()
+
+	for w, next := range d.activeWatchers {
+		if next != 0 && next <= compactRev {
+			delete(d.activeWatchers, w)
+			d.laggingWatchers[w] = next
+		} else {
+			w.Stop()
```
serathius: If there is a watch on `next=0` I don't see a reason to close it on compaction.

cache/demux.go
```diff
@@ -166,6 +166,31 @@ func (d *demux) Purge() {
...
+			delete(d.activeWatchers, w)
+		}
+	}
+	for w, next := range d.laggingWatchers {
+		if next != 0 && next > compactRev {
+			continue
+		}
+		w.Stop()
+		delete(d.laggingWatchers, w)
```
serathius: Lagging watchers should also be informed about being compacted. Why do we just drop them?

cache/cache.go
```diff
@@ -130,7 +119,17 @@ func (c *Cache) Watch(ctx context.Context, key string, opts ...clientv3.OpOption
...
+				if !ok {
+					return
+				}
+				select {
+				case <-ctx.Done():
+				case <-c.internalCtx.Done():
+				case responseChan <- resp:
+				}
+				return
+			case resp, ok := <-w.eventsCh:
```
serathius: There is no deterministic order when selecting on multiple channels. Close on eventsCh can come before event on cancelCh.

cache/demux.go
```diff
@@ -177,7 +202,9 @@ func (d *demux) resyncLaggingWatchers() {
 
 	for w, nextRev := range d.laggingWatchers {
 		if nextRev < oldestRev {
-			w.Stop()
+			if !w.notifyCompacted(nextRev) {
```
serathius: Now notifyCompacted doesn't close.

cache/watcher.go
```diff
@@ -15,30 +15,31 @@
...
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
 // watcher holds one client’s buffered stream of events.
 type watcher struct {
-	eventQueue chan []*clientv3.Event
-	keyPred    KeyPredicate
-	stopped    int32
-	done       chan struct{} // closed together with Stop()
+	eventsCh chan clientv3.WatchResponse
```
serathius: My recommended solution for compaction:
1. Watcher should stop receiving new events
2. Watcher should check if there is a cancel reason/response.
3. Watcher should write the cancel response
4. Watcher should close write channel

So I expect that compaction from demux perspective should work as follows:
1. We set the cancel reason
1. We close the events channel

This means we need 3 variables:
* `respCh chan clientv3.WatchResponse`; note this could be still `chan []*clientv3.Event` and refactor to `chan clientv3.WatchResponse` could be moved to separate PR.
* `cancelResp` *clientv3.WatchResponse.
* `stopOnce sync.Once`

We don't need mutex or channel, as long we guarantee that we set the `cancelResp` before we close `respCh` and do that under `stopOnce` to avoid races with `Stop`.

 

```
func (w *watcher) Stop() {
	w.stopOnce.Do(func() {
		close(w.eventsCh)
	})
}

func (w *watcher) Compact(compactRev int64) bool {
	resp := clientv3.WatchResponse{
		Canceled:        true,
		CompactRevision: compactRev,
		CancelReason:    rpctypes.ErrCompacted.Error(),
	}
	w.stopOnce.Do(func() {
		w.cancelResp = resp
		close(w.eventsCh)
	})
}
```
Watch select is simple:
```
			case resp, ok := <-w.eventsCh:
				if !ok {
					if w.cancelResp != nil {
						select {
							case <-ctx.Done():
							case <-c.internalCtx.Done():
							case responseChan <- w.cancelResp:
						}
					}
					return
				}
```

tests/integration/cache_test.go
```diff
@@ -960,6 +961,56 @@ func TestCacheUnsupportedGetOptions(t *testing.T) {
 	}
 }
 
+func TestCacheWatchOldRevisionCompacted(t *testing.T) {
```
serathius: Please send this as separate PR with the test working.

cache/demux.go
apullo777: Just want to double check, a watcher is only stored with `next==0` when history is empty and `startingRev==0`; otherwise `startingRev==0` gets normalized to `latestRev+1` in `Register`, so here we also want to keep those watcher for both `next==0` and `latestRev+1`(the latter needs to do `PeekLatest`)

tests/integration/cache_test.go
apullo777: Ok! Will add the test back in follow-up PR after we fix the issue

cache/demux.go
apullo777: Yes I think I mixed things up. We want to keep `next <= compctRev` for later cancellation

cache/demux.go
serathius: Don't understand. Compaction should remove compacted watchers. So those who are behind. Don't think watcher with `next=0` is behind, neither one with `next>compactedRev`. Not sure about `next=compactedRev`.

PR #20500, opened 2025-08-15
chore: Replace which with command -v

mrueg:
command -v is a shell built-in, while which requires a separate binary to be around.


ivanvc:
👏👏👏 thanks, @mrueg. Installing the dependency tools is something that definitely needs to be improved, I have it on my to do list 😅 

/ok-to-test

ahrtr (approved):
LGTM & thx

PR #20507, opened 2025-08-18
cache: migrate storage layer to B-tree

apullo777:
@serathius @MadhavJivrajani


cache/store.go
```diff
@@ -28,12 +28,29 @@ var ErrNotReady = fmt.Errorf("cache: store not ready")
...
 	mu        sync.RWMutex
-	kvs       map[string]*mvccpb.KeyValue
+	tree      *btree.BTree
+	degree    int
 	latestRev int64
 }
 
-func newStore() *store {
-	return &store{kvs: make(map[string]*mvccpb.KeyValue)}
+func newStore(degree int) *store {
```
serathius, line 36: Do we have enough tests for store to replace implementation now?

apullo777: I will write some unit tests for store!

cache/store_test.go
```diff
@@ -0,0 +1,445 @@
...
+		{
+			name:       "apply_with_same_revision_multiple_events",
+			initialKVs: []*mvccpb.KeyValue{makeKV("/a", "oldA", 10), makeKV("/b", "oldB", 10)},
+			initialRev: 10,
+			eventBatches: [][]*clientv3.Event{
+				{makePutEvent("/a", "newA", 10), makePutEvent("/b", "newB", 10)},
+			},
+			expectedLatestRev: 10,
+			expectedSnapshot:  []*mvccpb.KeyValue{makeKV("/a", "newA", 10), makeKV("/b", "newB", 10)},
+		},
```
serathius: Note, events with single revision should be applied at once, within single lock, same like etcd. If not, cache will expose partially applied revision which is invalid. So this test case is incorrect, also please make sure we correctly apply revision. We could add validation in storage that prevents same revision being applied twice to prevent this kind of bug.

cache/store_test.go
```diff
@@ -0,0 +1,445 @@
...
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+
+	mvccpb "go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+func TestStoreGet(t *testing.T) {
```
serathius: Thanks for extensive tests, however I would recommend to send them in separate PR that we merge first. This way we know that both previous and the current implementation passes them.

cache/store_test.go
```diff
@@ -0,0 +1,445 @@
...
+			initialKVs: []*mvccpb.KeyValue{makeKV("/a", "X", 10)},
+			initialRev: 10,
+			eventBatches: [][]*clientv3.Event{
+				{makeDelEvent("/a", 11)},
+			},
+			expectedLatestRev: 11,
+			expectedSnapshot:  nil,
+		},
+		{
+			name:              "delete_nonexistent",
```
serathius: This is sign of potential inconsistency, we should consider rejecting such write and purging cache. This is however something for a followup.

serathius (approved):

PR #20510, opened 2025-08-19
Add OpenTelemetry Spans to Lease methods

AwesomePatrol:

It adds `lease_grant`, `lease_revoke`, `lease_renew` spans to the corresponding `LeaseGrant`, `LeaseRevoke`, `LeaseRenew` methods.

The only two attributes are:
1. `id` for all methods
2. `ttl` for `LeaseGrant` only

/cc @serathius


tests/integration/tracing_test.go
```diff
@@ -130,6 +131,26 @@ func TestTracing(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "UnaryLeaseRevoke",
```
serathius, line 187: Why cover only this Method and not others?

AwesomePatrol: `LeaseGrant` will allocate random ID making it difficult to write test for. I can add `LeaseRenew`, but it doesn't provide any more information than this one

serathius: > I can add LeaseRenew, but it doesn't provide any more information than this one

It's a good regression test.

serathius: For `LeaseGrant` I assumed you can still provide id based on comment `// no id given? choose one`

AwesomePatrol: Maybe I am using too high interface, because `lease.Grant` doesn't allow setting `ID`: https://github.com/etcd-io/etcd/blob/v3.6.4/client/v3/lease.go#L216

serathius (approved):

PR #20511, opened 2025-08-19
Add more attributes to OpenTelemetry Txn spans

AwesomePatrol:

It adds the following attributes to `txn` spans:
1. `success_first_key` - `key` of the first success operation
2. `success_first_type` - either "put", "range", "delete_range", "txn"
3. `success_first_lease` - lease ID or `-1` if no `put` operation in the transaction or `0` if not set

Example: 
image

/cc @serathius

### DIFF
--- a/server/etcdserver/tracing.go
+++ b/server/etcdserver/tracing.go
@@ -1,0 +1,74 @@
+// Copyright 2025 The etcd Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


server/etcdserver/tracing.go
```diff
@@ -0,0 +1,77 @@
...
+		var key string
+		switch op := operation.GetRequest().(type) {
+		case *pb.RequestOp_RequestPut:
+			key = string(op.RequestPut.GetKey())
+		case *pb.RequestOp_RequestRange:
+			key = string(op.RequestRange.GetKey())
+		case *pb.RequestOp_RequestDeleteRange:
+			key = string(op.RequestDeleteRange.GetKey())
+		case *pb.RequestOp_RequestTxn:
+			key = "nested: " + firstOpKey(op.RequestTxn.GetSuccess())
```
serathius: Let's avoid recursion and just return empty.

serathius:
/ok-to-test

Can you squash commits?

PR #20521, opened 2025-08-20
Cache drop test

serathius:
/assign @ahrtr


ahrtr:
I didn't realize that it's related to cache. thx for dropping the test.

An unrelated comment: the interface of cache and relationship between cache and etcd client SDK seems a little confusing.  In the test cases, users can use the client SDK to operate k/v, but use the cache to watch. It complicates the usage. That's why I think it's better to get it clarified in the first place.

Anyway, it can be discussed separately.

PR #20524, opened 2025-08-20
Sort trace analysis table

AwesomePatrol:
Part of https://github.com/etcd-io/etcd/issues/20182.

The table is sorted by dictionary order of `pattern` then dictionary order of `method` then dictionary order of `args` (columns to the right of `pattern`).

`Spans` are analyzed instead of `ResourceSpans` what fixes the problem when multiple gRPC calls were made to Etcd in the context of a single trace.

/cc @serathius

PR #20533, opened 2025-08-21
Improve pattern matching in trace analysis

AwesomePatrol:
Follow up to https://github.com/etcd-io/etcd/pull/20524

Moved pattern matching and sorting to separate files with their own unit tests.

/cc @serathius


tests/robustness/coverage/pattern_test.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+		{
+			name: "core_namespace_list",
+			key:  "/registry/controllerrevisions/kube-system/",
+			want: "/registry/{resource}/{namespace}/",
+		},
+		{
+			name: "core_subresource_namespace_get",
+			key:  "/registry/services/specs/default/kubernetes",
+			want: "/registry/{resource}/{namespace}/{name}",
```
AwesomePatrol: Do you prefer to treat `services/specs` and `services/endpoints` as a regular `resource` or a special case?

tests/robustness/coverage/coverage_test.go
```diff
@@ -57,8 +55,8 @@ type refOp struct {
 	keyAttrName string
 }
 
-type row struct {
-	method, pattern, args string
+type Row struct {
```
serathius, line 58: Why make it public?

tests/robustness/coverage/pattern_test.go
```diff
@@ -0,0 +1,178 @@
+// Copyright 2025 The etcd Authors
```
serathius: There is no concept of pattern in etcd. Maybe call it `key_pattern.go`

tests/robustness/coverage/sort_test.go
```diff
@@ -0,0 +1,66 @@
...
+		{Method: "List", Pattern: "/registry/masterleases/", Args: ""},
+		{Method: "Get", Pattern: "/registry/masterleases/{name}", Args: ""},
+		{Method: "List", Pattern: "/registry/{api-group}/{resource}/", Args: "XX"},
+		{Method: "List", Pattern: "/registry/{api-group}/{resource}/", Args: "XXX"},
+		{Method: "Get", Pattern: "/registry/{api-group}/{resource}/{name}", Args: ""},
+		{Method: "Get", Pattern: "/registry/{resource}", Args: ""},
+		{Method: "List", Pattern: "/registry/{resource}/", Args: ""},
+		{Method: "Get", Pattern: "/registry/{resource}/{namespace}/{name}", Args: ""},
+		{Method: "Get", Pattern: "/registry/{resource}/{name}", Args: ""},
+		{Method: "Compaction", Pattern: "compact_rev_key", Args: ""},
```
serathius:
```suggestion
		{Pattern: "/registry/events/", Method: "", Args: ""},
		{Pattern: "/registry/events/{namespace}/", Method: "List", Args: ""},
		{Pattern: "/registry/events/{namespace}/{name}", Method: "Get", Args: ""},
		{Pattern: "/registry/health", Method: "Healthcheck", Args: ""},
		{Pattern: "/registry/masterleases/", Method: "List", Args: ""},
		{Pattern: "/registry/masterleases/{name}", Method: "Get", Args: ""},
		{Pattern: "/registry/{api-group}/{resource}/", Method: "List", Args: "XX"},
		{Pattern: "/registry/{api-group}/{resource}/", Method: "List", Args: "XXX"},
		{Pattern: "/registry/{api-group}/{resource}/{name}", Method: "Get", Args: ""},
		{Pattern: "/registry/{resource}", Method: "Get", Args: ""},
		{Pattern: "/registry/{resource}/", Method: "List", Args: ""},
		{Pattern: "/registry/{resource}/{namespace}/{name}", Method: "Get", Args: ""},
		{Pattern: "/registry/{resource}/{name}", Method: "Get", Args: ""},
		{Pattern: "compact_rev_key", Method: "Compaction", Args: ""},
```

tests/robustness/coverage/pattern_test.go
```diff
@@ -0,0 +1,178 @@
...
+// limitations under the License.
+
+package coverage_test
+
+import (
+	"strings"
+	"testing"
+)
+
+func normalize(k string) ([]string, string) {
```
serathius: Please define a struct to represent normalized key.

tests/robustness/coverage/pattern_test.go
```diff
@@ -0,0 +1,178 @@
...
+		// masterleases are a special case as they are not a regular (namespaced) resource.
+		return "masterleases", 1
+	case "events":
+		// events are a special case as they have watch cache disabled by default.
+		return "events", 1
+	}
+	return "{resource}", 1
+}
+
+func pattern(k string) (string, bool) {
```
serathius: Most important, top in call stack, function should be at the top of the file.

tests/robustness/coverage/coverage_test.go
AwesomePatrol: Otherwise `cmp.Diff` did not want to cooperate with me

tests/robustness/coverage/pattern_test.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+		{
+			name: "crd_no_namespace_list",
+			key:  "/registry/apiregistration.k8s.io/apiservices/",
+			want: "/registry/{api-group}/{resource}/",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got, valid := pattern(tc.key)
+			if tc.want == "" != !valid {
```
serathius: Please separate `valid` verification instead of assuming empty string is invalid.

tests/robustness/coverage/pattern_test.go
AwesomePatrol: Renamed to `key_pattern_test.go`

tests/robustness/coverage/pattern_test.go
AwesomePatrol: Split into two sets of table tests

tests/robustness/coverage/pattern_test.go
AwesomePatrol: Changed the order to start with this function

tests/robustness/coverage/pattern_test.go
AwesomePatrol: It is processed in only one place and it is extremely simple. Removed it instead

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		{
+			name: "core_namespace_list",
+			key:  "/registry/controllerrevisions/kube-system/",
+			want: "/registry/{resource}/{namespace}/",
+		},
+		{
+			name: "core_subresource_namespace_get",
+			key:  "/registry/services/specs/default/kubernetes",
+			want: "/registry/{resource}/{namespace}/{name}",
```
serathius: Where is subresource here?

AwesomePatrol: That was my question in https://github.com/etcd-io/etcd/pull/20533#discussion_r2290788028:
> Do you prefer to treat services/specs and services/endpoints as a regular resource or a special case?

There is only one such case (`services/{endpoints,specs}`) and I don't think it provides any insight (as opposed to `events` or `masterleases`). Moreover, it won't change due to compatibility promise

serathius: I'm not following. If the key includes subresource I expect that matching pattern includes subresource.

We should not create some new arbitrary grouping. Detecting patterns should identify patterns existing in apiserver.

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		{
+			name: "core_subresource_rev",
+			key:  "/registry/services/endpoints",
+			want: "/registry/{resource}",
```
serathius:
```suggestion
			want: "/registry/{resource}/{subresource}",
```

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		{
+			name: "core_subresource_namespace_list",
+			key:  "/registry/services/specs/default/",
+			want: "/registry/{resource}/{namespace}/",
```
serathius:
```suggestion
			want: "/registry/{resource}/{subresource}/{namespace}/",
```

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,189 @@
...
+			key:  "compact_rev_key",
+			want: "compact_rev_key",
+		},
+		{
+			name: "special_healthcheck",
+			key:  "/registry/health",
+			want: "/registry/health",
+		},
+		{
+			name: "core_namespace_get",
```
serathius: Why call it "core_namespace_get"  and not "core_namespace_name" ? We want to recognize key pattern, not a method.

Also `core` is K8s concept that might not be known for etcd users. Maybe call it `coreAPI_resource_namespace_name`

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,189 @@
...
+			key:  "/registry/cilium.io/ciliumendpoints/kube-system/",
+			want: "/registry/{api-group}/{resource}/{namespace}/",
+		},
+		{
+			name: "crd_no_namespace_get",
+			key:  "/registry/apiregistration.k8s.io/apiservices/v1alpha1.storage.k8s.io",
+			want: "/registry/{api-group}/{resource}/{name}",
+		},
+		{
+			name: "crd_no_namespace_rev",
```
serathius: Maybe let's describe what it included and not what's is missing. Like "crd_apigroup_resource"

serathius:
Sorting part looks ok, could you send a separate PR for sorting so we can merge it?

/ok-to-test

tests/robustness/coverage/coverage_test.go
serathius: Ok, makes sense

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Any grouping we will do in this test will be arbitrary. Added `services/{subresource}` group

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,189 @@
...
+			key:  "/registry/priorityclasses",
+			want: "/registry/{resource}",
+		},
+		{
+			name: "core_no_namespace_list",
+			key:  "/registry/priorityclasses/",
+			want: "/registry/{resource}/",
+		},
+		{
+			name: "crd_namespace_get",
```
serathius: There are non-core API that follow this pattern and are not CRDs. Like `apps` with `Deployment` resource.

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Not sure core is the correct name anyways. `Deployments` which have `apiVersion: apps/v1` still use `/registry/deployments` key. `nonCRD` doesn't sound good either.

> Also core is K8s concept

The entire trace analysis relies on k8s concepts. I don't think we should limit test case names this much (the test is simple enough to be easy to understand regardless what the name is).

> Maybe call it coreAPI_namespace_name

What do you think about: `core` -> `root`; `crd` -> `nested`?

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: I don't think this is correct. Here is dump of example keys:
```
/registry/deployments/deployment-1590/test-rolling-update-deployment
/registry/deployments/field-validation-3363/deployment-shared-map-item-removal
/registry/deployments/deployment-1482/test-deployment
/registry/deployments/kubectl-4245/frontend
/registry/deployments/kube-system/coredns
/registry/deployments/field-validation-7339/deployment-shared-unset
/registry/deployments/kubectl-4245/agnhost-primary
/registry/deployments
/registry/deployments/field-validation-7599/deployment
/registry/deployments/webhook-2741/sample-webhook-deployment
/registry/deployments/deployment-6209/test-rollover-deployment
```

tests/robustness/coverage/key_pattern_test.go
serathius: Talked offline, let's stay with `core` and `crd`

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Removed test case names in favor of descriptive values of keys

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+	slash := ""
+	if strings.HasSuffix(k, "/") {
+		slash = "/"
+	}
+	s := strings.Split(strings.Trim(k, "/"), "/")
+	if len(s[0]) == 0 {
+		return k, false
+	}
+
+	resource, shift := resourceOrCRD(s[0])
```
serathius: Isn't `resource` really a `apiResource` where api can be empty for non core?

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+	if len(s) == 2 && slash == "" {
+		// Get on a namespaced resource.
+		return "/registry/" + resource + "/{namespace}/{name}", true
+	}
+
+	return k, false
+}
+
+func resourceOrCRD(s string) (string, int) {
+	if strings.Contains(s, ".") {
```
serathius, line 64: Can you provide reference to K8s code confirming "." means CRD?

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+		return "/registry/" + resource + slash, true
+	}
+	if len(s) == 1 && slash == "" {
+		// Get on a non-namespaced resource.
+		return "/registry/" + resource + "/{name}", true
+	}
+	if len(s) == 1 {
+		// Listing all resources in a namespace.
+		return "/registry/" + resource + "/{namespace}" + slash, true
+	}
```
serathius:
```suggestion
	if len(s) == 1 {
		if slash == "" {
			// Get on a non-namespaced resource.
			return "/registry/" + resource + "/{name}", true
		}
		// Listing all resources in a namespace.
		return "/registry/" + resource + "/{namespace}" + slash, true
	}
```

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+	s := strings.Split(strings.Trim(k, "/"), "/")
+	if len(s[0]) == 0 {
+		return k, false
+	}
+
+	resource, shift := resourceOrCRD(s[0])
+	s = s[shift:]
+
+	if len(s) == 0 {
```
serathius: I think it would be better to go back to switch statement on `len(s)`

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+	if strings.HasSuffix(k, "/") {
+		slash = "/"
+	}
+	s := strings.Split(strings.Trim(k, "/"), "/")
+	if len(s[0]) == 0 {
+		return k, false
+	}
+
+	resource, shift := resourceOrCRD(s[0])
+	s = s[shift:]
```
serathius: Maybe set the name based on what it can contain. I think `namespaceName` is more clear and accurate than `s`, because `s` here was changed.

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+// limitations under the License.
+
+package coverage_test
+
+import (
+	"strings"
+	"testing"
+)
+
+func pattern(k string) (string, bool) {
```
serathius:
```suggestion
func pattern(key string) (string, bool) {
```
What's with the single letter variables?

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	k, valid := strings.CutPrefix(k, "/registry")
+	if !valid {
+		return k, false
+	}
+	slash := ""
+	if strings.HasSuffix(k, "/") {
+		slash = "/"
+	}
+	s := strings.Split(strings.Trim(k, "/"), "/")
```
serathius: Where variable name `s` comes from? It's common to call the results of split as `parts`.

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Renamed to `apiResource`

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: This is not part of the schema so it is not an enforced requirement. However, it still needs to be a DNS name, so it is unlikely that a name without a single `.` will be used.

AwesomePatrol: Added comment to explain what are the requirements and how unlikely it is that a CRD won't have a dot

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Changed. I still think the previous one was more readable

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: There is no guarantee that there will be a namespace name in the remainder part

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Left-over from refactor. Previously `k` was modified version of `key` so it avoided confusion. Updated

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Renamed to `parts`

tests/robustness/coverage/key_pattern_test.go
AwesomePatrol: Done

tests/robustness/coverage/key_pattern_test.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	return key, false
+}
+
+func resourceOrCRD(s string) (string, int) {
+	if strings.Contains(s, ".") {
+		// Group name from the Custom Resource Definition.
+		// Names without dots are technically valid DNS name, but against best practices:
```
serathius, line 66: So are you saying there is no guarantee that there is a "." in the CRD and you might incorrectly classify a pattern?

AwesomePatrol: This is correct, it is possible that some non-namespaced CRD resources will be classified as `/registry/{resource}/{namespace}/{name}` instead of `/registry/{api-group}/{resource}/{name}`, but the other cases should show up in the list as a new (unexpected) pattern

serathius: Makes sense, thanks for confirming.

PR #20543, opened 2025-08-22
cache: add unit tests and validation for store

apullo777:
@serathius @MadhavJivrajani

### DIFF
--- a/cache/store.go
+++ b/cache/store.go
@@ -75,15 +75,16 @@ func (s *store) Apply(events []*clientv3.Event) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	for _, ev := range events {
-		if ev.Kv.ModRevision < s.latestRev {
-			return fmt.Errorf("cache: stale event batch (rev %d < latest %d)", ev.Kv.ModRevision, s.latestRev)
-		}
+	if err := validateRevisions(events, s.latestRev); err != nil {
+		return err
 	}
 
 	for _, ev := range events {
 		switch ev.Type {
 		case clientv3.EventTypeDelete:
+			if _, ok := s.kvs[string(ev.Kv.Key)]; !ok {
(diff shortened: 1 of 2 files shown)


cache/store.go
```diff
@@ -135,3 +136,18 @@ func (s *store) scanRange(startKey, endKey []byte) []*mvccpb.KeyValue {
...
+
+func validateRevisions(events []*clientv3.Event, latestRev int64) error {
+	for _, ev := range events {
+		if ev.Kv.ModRevision < latestRev {
+			return fmt.Errorf("cache: stale event batch (rev %d < latest %d)", ev.Kv.ModRevision, latestRev)
+		}
+		if ev.Kv.ModRevision == latestRev {
+			return fmt.Errorf("cache: duplicate revision batch (rev %d == latest %d)", ev.Kv.ModRevision, latestRev)
+		}
+		if ev.Kv.ModRevision > latestRev+1 {
```
serathius: You don't have guarantee of revision continuity. Are you sure you have proper test coverage if you didn't notice that?

cache/store_test.go
```diff
@@ -0,0 +1,436 @@
...
+					t.Fatalf("Get after restore to rev=0 err=%v; want %v", err, ErrNotReady)
+				}
+				return
+			}
+			verifyStoreSnapshot(t, s, tt.expectedSnap, tt.expectedRev)
+		})
+	}
+}
+
+func TestStoreConcurrentGetApply(t *testing.T) {
```
serathius: I don't think this is a good test, you don't do test for concurrency. You can just run race detection.

cache/store.go
apullo777: Thanks. I totally missed that! I will spend some to see if we have enough test cases add more cases like Apply accepts a single watch response with multiple revisions, or a gap test (e.g. latest=20 with arriving event at 22).

serathius (approved):

PR #20503, opened 2025-08-18
DONOTMERGE: reproduce #20271

fuweid:
REF: #20271

Please read https://github.com/etcd-io/etcd/blob/main/CONTRIBUTING.md#contribution-flow.

Issue #20271, opened 2025-07-03
panic: assertion failed: Page expected to be: 5


serathius:
### Bug report criteria

- [ ] This bug report is not security related, security issues should be disclosed privately via security@etcd.io.
- [ ] This is not a support request or question, support requests or questions should be raised in the etcd [discussion forums](https://github.com/etcd-io/etcd/discussions).
- [ ] You have read the etcd [bug reporting guidelines](https://github.com/etcd-io/etcd/blob/main/Documentation/contributor-guide/reporting_bugs.md).
- [ ] Existing open issues along with etcd [frequently asked questions](https://etcd.io/docs/latest/faq) have been checked and this is not a duplicate.

### What happened?

Panic discovered by Antithesis testing

https://linuxfoundation.antithesis.com/report/G-9rIjiZJiwodTEN5avQ7wgK/u_uFsWOwZSxS5mOmbEprwMUijNhsWdV6mfde_CT-y4k.html?auth=v2.public.eyJuYmYiOiIyMDI1LTA3LTAzVDA3OjMxOjUyLjQzMzQ2ODk3NFoiLCJzY29wZSI6eyJSZXBvcnRTY29wZVYxIjp7ImFzc2V0IjoidV91RnNXT3daU3hTNW1PbWJFcHJ3TVVpak5oc1dkVjZtZmRlX0NULXk0ay5odG1sIiwicmVwb3J0X2lkIjoiRy05cklqaVpKaXdvZFRFTjVhdlE3d2dLIn19fUHU0wnVRoDtfilwOCROUiDTtcOlIZkrVaddCqjorH3utgcIEPIzlsrMAJGXFC6NTZMneLqAWWU_lq-9prD_tQc

/cc @ahrtr @fuweid 

### What did you expect to happen?

No panic

### How can we reproduce it (as minimally and precisely as possible)?

Contact @marcus-hodgson-antithesis to help reproduce it on Antithesis 

### Etcd version (please run commands below)

main branch

### Etcd configuration (command line flags or environment variables)

<details>

# paste your configuration here

</details>

### Etcd debug information (please run commands below, feel free to obfuscate the IP address or FQDN in the output)

<details>

```console
$ etcdctl member list -w table
# paste output here

$ etcdctl --endpoints=<member list> endpoint status -w table
# paste output here
```

</details>

### Relevant log output

```Shell
[...]

PR #20553, opened 2025-08-24
[NotMerge] simple fix for 20271

silentred:

### DIFF
--- a/server/etcdserver/server.go
+++ b/server/etcdserver/server.go
@@ -1029,6 +1029,9 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, toApply *toApply) {
 	// wait for raftNode to persist snapshot onto the disk
 	<-toApply.notifyc
 
+	// protect the backend from concurrent defrag
+	s.bemu.Lock()
+
 	// gofail: var applyBeforeOpenSnapshot struct{}
 	newbe, err := serverstorage.OpenSnapshotBackend(s.Cfg, s.snapshotter, toApply.snapshot, s.beHooks)
 	if err != nil {
@@ -1066,7 +1069,7 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, toApply *toApply) {
 	// Closing old backend might block until all the txns
 	// on the backend are finished.
 	// We do not want to wait on closing the old backend.
-	s.bemu.Lock()
+	// s.bemu.Lock()
 	oldbe := s.be
 	go func() {
 		lg.Info("closing old backend file")


ahrtr:
This isn't correct. It can't prevent defragmentation from renaming the db file concurrently.

ahrtr:
The defragmentation also needs to acquire this lock.

https://github.com/etcd-io/etcd/blob/84ac605338e4f90ed892702de7bb723f1c57f788/server/etcdserver/api/v3rpc/maintenance.go#L109-L120



The `(s *EtcdServer) Backend()` is problematic, it just protects getting the backend instance; it can't protect the following operation (i.e. defragmentation).
https://github.com/etcd-io/etcd/blob/84ac605338e4f90ed892702de7bb723f1c57f788/server/etcdserver/server.go#L2427-L2431

silentred:
> it just protects getting the backend instance

You are right. 

I was thinking `bemu` should  protect not only changes of variable `be` but also underlying resources of the backend, e.g. db file, readTx, batchTx. So if we use a lock to protect 
- from `OpenSnapshotBackend` to the end of `applySnapshot`
- and whole `defrag` process

is that supposed to work? 
Please correct me if I am wrong, thank you.

ahrtr:
> * and whole `defrag` process

I think we can add a `Defragment` method for `Etcdserver` and acquire the lock `bemu`, the rough implementation is something like below,
```
func (s *Etcdserver) Defragment(...) {
     	s.bemu.RLock() 
 	defer s.bemu.RUnlock() 
 	return s.Backend().Defrag() 
```

 The `(ms *maintenanceServer) Defragment` should cal `(s *Etcdserver) Defragment` instead of the `Backend().Defrag()`.

also cc @fuweid

server/etcdserver/server.go
```diff
@@ -1029,6 +1029,9 @@ func (s *EtcdServer) applySnapshot(ep *etcdProgress, toApply *toApply) {
 	// wait for raftNode to persist snapshot onto the disk
 	<-toApply.notifyc
 
+	// protect the backend from concurrent defrag
+	s.bemu.Lock()
```
fuweid, line 1039: Just have concern about increasing that scope of critical section. Yeah. it's simple fix. Just in case we run into dead lock. We probably need e2e test to cover that.

silentred:
Hi @ahrtr , thanks for your detailed explanation https://github.com/etcd-io/etcd/pull/20585#issuecomment-3242035232 . 
According to your comment, I think you might prefer this PR's solution. I have already modified the code in the direction of this suggestion https://github.com/etcd-io/etcd/pull/20553#issuecomment-3227608313 .

I think we still need to align our understandings.

> The problem is that during the above period, there are two different backend instances.

Yes, there are 2 backend instances`newbe` and `s.be`. 

> but both of which point to the same db file.

I think they have different inode .`s.be` points to old db file, `newbe` points to new db file, renamed from `xxx.snap.db`

> Note that normally the old backend instance still reads from the old db file even after the renaming, until you restart etcdserver or perform a defragment right after the renaming or setting the new backend (s.be = newbe).

I think after `s.kv.Restore(newbe)`, etcdserver starts reading from `newbe` to serve client's range requests.

> We should block any write to both backends during the above process

Execpt `defrag`, I am not seeing any writes during `applySnapshot`, because it is in the apply loop, in which there is no concurrent user writes.


https://github.com/etcd-io/etcd/pull/20585 this PR attempts to rename snap.db file in the backend, so there is only one backedn instance. It locks the whole process of renaming `xxx.snap.db` to `db`, prevents concurrent `defrag`.

server/etcdserver/server.go
```diff
@@ -968,6 +968,13 @@ func (s *EtcdServer) Cleanup() {
 	}
 }
 
+func (s *EtcdServer) Defragment() (err error) {
+	s.bemu.Lock()
+	defer s.bemu.Unlock()
+	err = s.be.Defrag()
+	return
+}
```
ahrtr:
```suggestion
func (s *EtcdServer) Defragment() error {
	s.bemu.Lock()
	defer s.bemu.Unlock()
	return s.be.Defrag()
}
```

ahrtr:
> Hi @ahrtr , thanks for your detailed explanation [#20585 (comment)](https://github.com/etcd-io/etcd/pull/20585#issuecomment-3242035232) .
> According to your comment, I think you might prefer this PR's solution. I have already modified the code in the direction of this suggestion [#20553 (comment)](https://github.com/etcd-io/etcd/pull/20553#issuecomment-3227608313) .

I didn't realize that you raised two separate PRs, pls close https://github.com/etcd-io/etcd/pull/20585 to avoid confusion.



> I think they have different inode .`s.be` points to old db file, `newbe` points to new db file, renamed from `xxx.snap.db`

It's the reason why I said  `Note that normally the old backend instance still reads from the old db file even after the renaming, until you restart etcdserver or perform a defragment right after the renaming or setting the new backend (s.be = newbe)`.



> Execpt `defrag`, I am not seeing any writes during `applySnapshot`, because it is in the apply loop, in which there is no concurrent user writes.

it isn't true. There is also API layer write, i.e. Storage version update. 

https://github.com/etcd-io/etcd/blob/963a72e241043b14680224c55b68abee623622e3/server/etcdserver/version/monitor.go#L107

silentred:
> it isn't true. There is also API layer write, i.e. Storage version update.

Thanks. I didn't notice.


I will squash this PR's commits when it is ready to merge.

server/etcdserver/server.go
silentred: I am not sure what this E2E test should do. My thought is this test might include:
- your e2e https://github.com/etcd-io/etcd/pull/20503 , which verifies concurrent defrag
- verify other concurrent API layer write, like @ahrtr mentioned in https://github.com/etcd-io/etcd/pull/20553#issuecomment-3242920816

Any advice?  Big thanks.

BTW, can I ask why should we not use SIGSTOP to pause the etcd process?

ahrtr:
-  I see the PR is based on a very old commit. Please rebase this PR although there isn't conflict. 
- Please do the minor refactoring as https://github.com/ahrtr/etcd/commit/7f61069016e6d63b4fee17c7ad6110183607c720, there are two benefits,
  - to ensure the bemu is always unlocked eventually, even panicking in between.
  - Move the closing of the old backend instance out of the critical section. Note deferred functions won't be executed if another goroutine panicking.
- Please add an e2e test similar to https://github.com/etcd-io/etcd/pull/20503 to verify that when a follower is applying a snapshot, a concurrent defragmentation won't cause any problem.

silentred:
> * I see the PR is based on a very old commit. Please rebase this PR although there isn't conflict.
> * Please do the minor refactoring as [ahrtr@7f61069](https://github.com/ahrtr/etcd/commit/7f61069016e6d63b4fee17c7ad6110183607c720), there are two benefits,
>   
>   * to ensure the bemu is always unlocked eventually, even panicking in between.
>   * Move the closing of the old backend instance out of the critical section. Note deferred functions won't be executed if another goroutine panicking.
> * Please add an e2e test similar to [DONOTMERGE: reproduce #20271 #20503](https://github.com/etcd-io/etcd/pull/20503) to verify that when a follower is applying a snapshot, a concurrent defragmentation won't cause any problem.

Done.

tests/e2e/reproduce_20271_test.go
```diff
@@ -0,0 +1,112 @@
...
+	t.Log("Step 3: Delete some key values.")
+	for i := 0; i < snapCount+20; i++ {
+		_, err = epc.Procs[0].Etcdctl().Delete(ctx, fmt.Sprintf("foo%d", i), config.DeleteOptions{})
+		require.NoError(t, err)
+	}
+
+	t.Log("Step 4: Restarting the first two members to change term and trigger new leader to send snapshot file to the third member.")
+	for _, proc := range epc.Procs[:2] {
+		require.NoError(t, proc.Restart(ctx))
+	}
```
ahrtr, line 76: This should be unnecessary. 
```suggestion
```

tests/e2e/reproduce_20271_test.go
```diff
@@ -0,0 +1,112 @@
...
+	require.Equal(t, 1, leaderIndex)
+
+	t.Log("Step 9: The third member will commit them into deleted boltdb file. The boltdb file will be different from deleted one.")
+	for i := 0; i < snapCount+50; i++ {
+		err = epc.Procs[2].Etcdctl().Put(ctx, fmt.Sprintf("foo%d", i), "bar", config.PutOptions{})
+		require.NoError(t, err)
+	}
+
+	t.Log("Step 10: Restart the first member and it should be panic")
+	require.NoError(t, epc.Procs[0].Start(ctx))
```
ahrtr: It's a little complicated and hard to maintain. I think we just need to write some data, and finally check all members have consistent data (hash).

tests/framework/e2e/etcd_process.go
```diff
@@ -275,6 +275,16 @@ func (ep *EtcdServerProcess) Kill() error {
...
 
+func (ep *EtcdServerProcess) Pause() error {
+	ep.cfg.lg.Info("Pausing server...", zap.String("name", ep.cfg.Name))
+	return ep.proc.Signal(syscall.SIGSTOP)
+}
+
+func (ep *EtcdServerProcess) Resume() error {
+	ep.cfg.lg.Info("Resuming server...", zap.String("name", ep.cfg.Name))
+	return ep.proc.Signal(syscall.SIGCONT)
+}
```
ahrtr, line 288: https://github.com/etcd-io/etcd/pull/20503#discussion_r2321526911

tests/e2e/reproduce_20271_test.go
```diff
@@ -0,0 +1,112 @@
...
+		require.NoError(t, epc.Procs[0].Etcdctl().Put(ctx,
+			fmt.Sprintf("foo%d", i),
+			strings.Repeat("Oops", 1024),
+			config.PutOptions{}))
+	}
+
+	t.Log(`Step 2: Currently, it's unable to use PeerProxy to block specific\n
+data path from one follower to leader. So here, I use SIGSTOP to pause the third member.`)
+	require.NoError(t, epc.Procs[2].Failpoints().SetupHTTP(ctx, "applyAfterOpenSnapshot", `sleep("15s")`))
+	epc.Procs[2].(*e2e.EtcdServerProcess).Pause()
```
ahrtr: After https://github.com/etcd-io/etcd/pull/20553/files#r2321527783 is resolved, then follow suggestion below
```suggestion
	epc.Procs[2].Pause()
```

tests/e2e/reproduce_20271_test.go
```diff
@@ -0,0 +1,112 @@
...
+	for i := 0; i < snapCount+20; i++ {
+		_, err = epc.Procs[0].Etcdctl().Delete(ctx, fmt.Sprintf("foo%d", i), config.DeleteOptions{})
+		require.NoError(t, err)
+	}
+
+	t.Log("Step 4: Restarting the first two members to change term and trigger new leader to send snapshot file to the third member.")
+	for _, proc := range epc.Procs[:2] {
+		require.NoError(t, proc.Restart(ctx))
+	}
+	epc.Procs[2].(*e2e.EtcdServerProcess).Resume()
```
ahrtr:
```suggestion
	epc.Procs[2].Resume()
```

tests/e2e/reproduce_20271_test.go
siyuanfoundation: +1, you can use `assertKVHash`

ahrtr:
Note this PR is a blocker for our next 3.6 & 3.5 patch releases.

server/etcdserver/server.go
siyuanfoundation: It is weird that the be file is replaced inside `OpenSnapshotBackend`, and the be is replaced outside. 
I would suggest moving `os.Rename(snapPath, cfg.BackendPath())` to a new func like
```
func (s *EtcdServer) swapBackend(newbe *backend.Backend) {
s.bemu.Lock()
defer s.bemu.Unlock()
s.be = newbe
os.Rename(snapPath, cfg.BackendPath())
}
```

silentred: > moving `os.Rename(snapPath, cfg.BackendPath())` to a new func

Considering the perspective of upper-level services (kvstore, consistIndex),  replacing backend instance is not sufficient for them to be ready to serve. Resoring each of them is necessary. If we add a new `swapBackend` func, then I think we have to also move restoring process to this function.

tests/e2e/reproduce_20271_test.go
silentred: Thanks for the advice. I am on vacation right now, coming back tomorrow. I will fix this ASAP.

tests/e2e/reproduce_20271_test.go
fuweid: SIGSTOP is used to stop process but TCP layer is still functioning.
Restarting existing running etcd members can re-connect and drop pending message in kernel side.
And it also increases term, which could be easier to trigger snapshot. (changing term is not required here, sorry for that confusing).

tests/e2e/reproduce_20271_test.go
fuweid: That origin case was trying to follow that issue #20271.
That KVHash is still reading opened backend instead of opening new file. 
The restarting here is to force member to receive snapshot.
Would like to see how to use KVHash here.

tests/e2e/reproduce_20271_test.go
ahrtr: > and drop pending message in kernel side.

OK, it might make some sense.

server/etcdserver/server.go
ahrtr: Let keep it as it's and try not to make too much refactoring for now. Any refactoring should be discussed separately on main branch only. Note this PR needs to backported.

ahrtr:
The PR overall looks good. Please simplify the logs message in the e2e test case.

silentred:
> The PR overall looks good. Please simplify the logs message in the e2e test case.

done

ahrtr:
It'd be better to squash the commits.

ahrtr:
@silentred could you manually backport this PR to 3.5 and 3.4? And also please update the changelog for 3.4, 3.5 and 3.6.

tests/e2e/reproduce_20271_test.go
```diff
@@ -0,0 +1,98 @@
...
+	require.NoError(t, epc.Procs[2].Failpoints().SetupHTTP(ctx, "applyAfterOpenSnapshot", `sleep("15s")`))
+	epc.Procs[2].Pause()
+
+	t.Log("Step 3: Delete some key values to trigger new snapshot on the first two members")
+	for i := 0; i < snapCount+20; i++ {
+		_, err = epc.Procs[0].Etcdctl().Delete(ctx, fmt.Sprintf("foo%d", i), config.DeleteOptions{})
+		require.NoError(t, err)
+	}
+
+	t.Log("Step 4: Restarting the first two members to change term and trigger new leader to send snapshot file to the third member.")
```
fuweid, line 73: this comment should be updated. It is to reconnect to paused one.

silentred:
> @silentred could you manually backport this PR to 3.5 and 3.4? And also please update the changelog for 3.4, 3.5 and 3.6.

no problem.

tests/e2e/reproduce_20271_test.go
silentred: "Step 4: Restarting the first two members to re-connect to the paused member, so the inflight messages will be dropped. This will trigger new leader to send snapshot file to the third member."
Does this comment look good to you? If so, I will change this comment in the backport PR and raise another PR for main branch. Thanks.

fuweid: Yes. It looks good to me. Thanks

